PR #6461, opened 2019-12-29
Introduce internal jobs for functions

### LINKED ISSUE #6053: Inconsistent job ids
I must admit that the job handling is very weird in fish, since the job id seems to be pretty random. Quick example, where I start different tools in the prescribed order and send them to background (`ccmake, nano, man gcc, vim, vi` where vi is an alias to `nvim -O`). The output of jobs is extremely unexpected, since `man gcc` will receive jobid 5 instead of 3. `vim` started as fourth tool will receive 4. The `nvim` alias started last will receive 7.

```
jobs
7       30462   0%      stopped nvim  $argv
4       30393   0%      stopped vim
5       30348   0%      stopped command man $argv
2       30308   0%      stopped nano
1       30273   0%      stopped ccmake .
```
Or simply starting `man gcc` and then `vim` results in job 1 and 2 not even being present
```
4       1716    0%      stopped nvim  $argv
3       1660    0%      stopped command man $argv
```
In short, the job ids are not consecutive, which is weird, but even worse they are not even monotonically increasing. This effectively makes the use of commands using these ids such as `fg/bg %id` trial and error and significantly less useful compared to most other shells.

fish version: 3.0.2
Sys: Ubuntu 18.04

danzimm:
## Motivation

This PR is aimed at improving how job ids are assigned. In particular,
previous to this commit, a job id would be consumed by functions (and
thus aliases). Since it's usual to use functions as command wrappers
this results in awkward job id assignments.

For example if the user is like me and just made the jump from vim -> neovim
then the user might create the following alias:
```
alias vim=nvim
```
Previous to this commit if the user ran `vim` after setting up this
alias, backgrounded (^Z) and ran `jobs` then the output might be:
```
Job	Group	State	Command
2	60267	stopped	nvim  $argv
```
If the user subsequently opened another vim (nvim) session, backgrounded
and ran jobs then they might see what follows:
```
Job	Group	State	Command
4	70542	stopped	nvim  $argv
2	60267	stopped	nvim  $argv
```
These job ids feel unnatural, especially when transitioning away from
e.g. bash where job ids are sequentially incremented (and aliases/functions
don't consume a job id).

See #6053 for more details.

## Implementation Details

As @ridiculousfish pointed out in https://github.com/fish-shell/fish-shell/issues/6053#issuecomment-559899400,
we want to elide a job's job id if it corresponds to a single function in the
foreground. This translates to the following prerequisites:

- A job must correspond to a single process (i.e. the job continuation
    must be empty)
- A job must be in the foreground (i.e. `&` wasn't appended)
- The job's single process must resolve to a function invocation

If all of these conditions are true then we should mark a job as
"internal" and somehow remove it from consideration when any
infrastructure tries to interact with jobs / job ids.

I saw two paths to implement these requirements:

- At the time of job creation calculate whether or not a job is
  "internal" and use a separate list of job ids to track their ids.
  Additionally introduce a new flag denoting that a job is internal so
  that e.g. `jobs` doesn't list internal jobs
  - I started implementing this route but quickly realized I was
    computing the same information that would be computed later on (e.g.
    "is this job a single process" and "is this jobs statement a
    function"). Specifically I was computing data that populate_job_process
    would end up computing later anyway. Additionally this added some
    weird complexities to the job system (after the change there were two
    job id lists AND an additional flag that had to be taken into
    consideration)
- Once a function is about to be executed we release the current jobs
  job id if the prerequisites are satisfied (which at this point have
  been fully computed).
  - I opted for this solution since it seems cleaner. In this
  implementation "releasing a job id" is done by both calling
  `release_job_id` and by marking the internal job_id member variable to
  -1. The former operation allows subsequent child jobs to reuse that
  same job id (so e.g. the situation described in Motivation doesn't
  occur), and the latter ensures that no other job / job id
  infrastructure will interact with these jobs because valid jobs have
  positive job ids. The second operation causes job_id to become
  non-const which leads to the list of code changes outside of `exec.c`
  (i.e. a codemod from `job_t::job_id` -> `job_t::job_id()` and moving the
   old member variable to a non-const private `job_t::job_id_`)

Note: Its very possible I missed something and setting the job id to -1
will break some other infrastructure, please let me know if so!

## Misc Comments

I tried to run `make/ninja lint`, but a bunch of non-relevant issues
appeared (e.g. `fatal error: 'config.h' file not found`). I did
successfully clang-format (`git clang-format -f`) and run tests, though.
This PR closes #6053.

As an ad-hoc test I went through the situation described in the motivation section and observed that vim only took up a single job id:
```
nad||~/Codes/fish-shell/build||0||seqjob                                                             Nu.fios-router.home.ttys002||Sun Dec 29 11:28:33 EST 2019
|> ./fish
nad||~/Codes/fish-shell/build||0||seqjob                                                             Nu.fios-router.home.ttys002||Sun Dec 29 11:31:43 EST 2019
|> type vim
vim is a function with definition
# Defined in - @ line 1
function vim --wraps=nvim --description 'alias vim=nvim'
    nvim  $argv;
end
nad||~/Codes/fish-shell/build||0||seqjob
|> vim
nad||~/Codes/fish-shell/build||0||seqjob
|> jobs
Job	Group	State	Command
1	73795	stopped	nvim  $argv
```
---

## TODOs:

### DIFF
--- a/src/exec.cpp
+++ b/src/exec.cpp
@@ -542,7 +542,7 @@ static bool handle_builtin_output(parser_t &parser, const std::shared_ptr<job_t>
         p->completed = true;
         if (p->is_last_in_job) {
             FLOGF(exec_job_status, L"Set status of job %d (%ls) to %d using short circuit",
-                  j->job_id, j->preview().c_str(), p->status);
+                  j->job_id(), j->preview().c_str(), p->status);
             parser.set_last_statuses(j->get_statuses());
         }
         return true;
@@ -718,7 +718,7 @@ using proc_performer_t = std::function<proc_status_t(parser_t &parser)>;
 // exists. This is just a dumb artifact of the fact that we only capture the functions name, not its
 // properties, when creating the job; thus a race could delete the function before we fetch its
 // properties.
-static proc_performer_t get_performer_for_process(process_t *p, const job_t *job,
+static proc_performer_t get_performer_for_process(process_t *p, job_t *job,
                                                   const io_chain_t &io_chain) {
     assert((p->type == process_type_t::function || p->type == process_type_t::block_node) &&
            "Unexpected process type");
@@ -746,6 +746,13 @@ static proc_performer_t get_performer_for_process(process_t *p, const job_t *job
         };
     } else {
         assert(p->type == process_type_t::function);
+        // If this function is the only process in the current job
+        // and that job is in the foreground then mark this job as internal
+        // so it doesn't increment the job id for any jobs created within this
+        // function.
+        if (p->is_first_in_job && p->is_last_in_job && job->flags().foreground) {
+            job->mark_internal();
+        }
         auto props = function_get_properties(p->argv0());
         if (!props) {
             FLOGF(error, _(L"Unknown function '%ls'"), p->argv0());
@@ -1131,7 +1138,7 @@ bool exec_job(parser_t &parser, shared_ptr<job_t> j, const job_lineage_t &lineag
         }
     }
 
-    FLOGF(exec_job_exec, L"Executed job %d from command '%ls' with pgrp %d", j->job_id,
+    FLOGF(exec_job_exec, L"Executed job %d from command '%ls' with pgrp %d", j->job_id(),
           j->command_wcstr(), j->pgid);
 
     j->mark_constructed();
--- a/src/builtin_bg.cpp
+++ b/src/builtin_bg.cpp
@@ -22,12 +22,12 @@ static int send_to_bg(parser_t &parser, io_streams_t &streams, job_t *j) {
     if (!j->wants_job_control()) {
         wcstring error_message = format_string(
             _(L"%ls: Can't put job %d, '%ls' to background because it is not under job control\n"),
-            L"bg", j->job_id, j->command_wcstr());
+            L"bg", j->job_id(), j->command_wcstr());
         builtin_print_help(parser, streams, L"bg", &error_message);
         return STATUS_CMD_ERROR;
     }
 
-    streams.err.append_format(_(L"Send job %d '%ls' to background\n"), j->job_id,
+    streams.err.append_format(_(L"Send job %d '%ls' to background\n"), j->job_id(),
                               j->command_wcstr());
     parser.job_promote(j);
     j->mut_flags().foreground = false;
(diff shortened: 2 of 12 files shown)


danzimm:
Should we add an additional test for this? Or is the ad-hoc testing ok? This isn't really fixing a "regression" so I'm not sure what the standard is here.

src/exec.cpp
```diff
@@ -746,6 +746,13 @@ static proc_performer_t get_performer_for_process(process_t *p, const job_t *job
         };
     } else {
         assert(p->type == process_type_t::function);
+        // If this function is the only process in the current job
```
danzimm: This is the only real change- everything else is a consequence of moving job_t::job_id to a private member variable and introducing a public getter.

ridiculousfish:
Seems reasonable to me. This should be testable, I'll write one and share it here.

ridiculousfish:
Testing revealed that while job IDs are consecutive, the order is still funny - investigating.

danzimm:
@ridiculousfish While working on this I noticed that a lot of seemingly internal processes get their own job (and this job id). I think generally speaking all non-command (or command like) jobs shouldn’t get a job id, but right now because we don’t “resolve” a job (i.e. we don’t realize the AST into any IR) this is a difficult computation to make at job creation time.

Unfortunately I’m mobile otherwise I could give more concrete examples about what I mean from the above paragraph. Fwiw it was useful to trace job creation (I just put printf’s with a bunch of state when a new job was constructed)

danzimm:
It might be worthwhile to reinvestigate when job ids are assigned. Right now it’s at job creation time, but it might be beneficial to populate a job with a job id after we’ve fully transformed an AST node into a job

ridiculousfish:
Bleh, there's a bug in `job_promote` - it's using `std::rotate` wrongly. I must have read that code a million times, how did I miss that.

ridiculousfish:
I merged this as 8e17d29e04eeef27471eb682a06f8d9116cb26ca (thanks!!) but there's *still* some weird ordering bug, investigating.

ridiculousfish:
Looks like this is in 3.1 after all

ridiculousfish:
Ah hit's not an ordering bug, it's that this fix needs to be extended to block jobs (not just functions).

danzimm:
Thank you!!

danzimm:
@ridiculousfish do you want me to take that on? I can try to come up with some tests too. I have time tomorrow, so just let me know!

ridiculousfish:
Thanks for offering! But it's an easy fix.

ridiculousfish:
It took a couple of iterations but [here's the test I ended up with](https://github.com/fish-shell/fish-shell/blob/91404f17629e8e5e7360d92a6c555383669af31d/tests/checks/job-ids.fish#L1), as an illustration.

danzimm:
Whoa! That’s a pretty sweet test! I didn’t know fish had filecheck tests- that’s so cool!

PR #6486, opened 2020-01-09
Clarify how to redirect stderr into stdout in tutorial

gladhorn:
People get confused about the redirection (there are several questions on stack overflow for example) and re-reading the tutorial section didn't make it clear to me immediately, so add one example to help everyone.


faho:
Merged, thanks!

PR #6501, opened 2020-01-14
Add completions for loginctl and resolvectl

239:
## Description

What would be the best practice for (optional) arguments like `--host=[USER@]HOST` or `--raw[=payload|packet]`?

## TODOs:


faho:
>What would be the best practice for (optional) arguments like --host=[USER@]HOST or --raw[=payload|packet]?

So you mean `--raw` by itself is okay? In that case don't use `-x`, which is `-r` (`--require-parameter`) and `-f` (`--no-files`) in one. So you'll just want `-f` if it can't take files.

`--host` is also in machinectl and systemctl, and there I've `-x`d them and used our `__fish_print_hostnames` function to print a host. I did't add the "USER@" part, that would require `__fish_complete_user_at_host`. I don't think just `--host` by itself is valid? At least I don't see what it should specify.

239:
I added both options for `--host`  without any completion functions for now.
Not sure how to deal with commands like `resolvectl dns [LINK [SERVER...]]` but I added the commands as simple lists for rudimentary completion.

share/completions/loginctl.fish
```diff
@@ -0,0 +1,26 @@
...
+set -l output cat export json json-pretty json-seq json-sse short short-full short-iso short-iso-precise short-monotonic short-precise short-unix verbose with-unit
+
+#commands
+complete -c loginctl -x -n "not $seen $commands" -a "$commands"
+
+#options
+complete -c loginctl -x -n "not $seen $commands" -l all -s a -d "Show all properties, including empty ones"
+complete -c loginctl -x -n "not $seen $commands" -l full -s l -d "Do not ellipsize output"
+complete -c loginctl -x -n "not $seen $commands" -l help -s h -d "Show this help"
+complete -c loginctl -x -n "not $seen $commands" -l host -a "HOST USER@HOST" -s H -d "Operate on remote host"
```
faho: Please don't add empty arguments like this - they're annoying to use.

faho:
Merged, thanks!

PR #6511, opened 2020-01-19
Replace debug() with flog

faho:
## Description

This turns all debug() calls into `FLOG` or `FLOGF` as needed.

In doing so it increases the number of flog categories from 17 to 29.

It doesn't yet remove the debug *code*, in particular both `fish` and `fish_indent` (which has one total warning) read numbers for `--debug`, but then don't print anything, and the debug macro and implementation still exist.

Very few debug messages have been removed without replacement, these are ones I judged to be of no use, especially if they were at debug level > 2, because I recall that being super spammy. If we ever need them we can add them back.

I would like input regarding the categories and their naming. This is a user interface, and naming categories in a way that users have an idea which they should enable is good, even if the users end up being just the fish-shell members.

Also: Should we add this for 3.1 still, or wait for 3.2?

## TODOs:

### DIFF
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -184,7 +184,7 @@ bool is_windows_subsystem_for_linux() {
                 // is bypassed. We intentionally do not include this in the error message because
                 // it'll only allow fish to run but not to actually work. Here be dragons!
                 if (getenv("FISH_NO_WSL_CHECK") == nullptr) {
-                    debug(0,
+                    FLOGF(error,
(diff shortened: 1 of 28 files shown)


zanchey:
> Also: Should we add this for 3.1 still, or wait for 3.2?

Please wait!

faho:
Just checking if you're paying attention :smiling_imp:

faho:
Okay, let's merge this now (to master, I'm not a monster!) because it's very likely to go stale fast.

PR #6513, opened 2020-01-20
treewide: replace deprecated C headers

neheb:
Found with clang-tidy's modernize-deprecated-headers.

### DIFF
--- a/src/builtin_function.cpp
+++ b/src/builtin_function.cpp
@@ -22,7 +22,7 @@
 #include "parser.h"
 #include "parser_keywords.h"
 #include "proc.h"
-#include "signal.h"
+#include "signal.h" // NOLINT
 #include "wgetopt.h"
 #include "wutil.h"  // IWYU pragma: keep
 
--- a/src/builtin_functions.cpp
+++ b/src/builtin_functions.cpp
@@ -26,7 +26,7 @@
 #include "parser.h"
 #include "parser_keywords.h"
 #include "proc.h"
-#include "signal.h"
+#include "signal.h" // NOLINT
 #include "wcstringutil.h"
 #include "wgetopt.h"
 #include "wutil.h"  // IWYU pragma: keep
--- a/src/builtin_wait.cpp
+++ b/src/builtin_wait.cpp
@@ -10,7 +10,7 @@
 #include "common.h"
 #include "parser.h"
 #include "proc.h"
-#include "signal.h"
+#include "signal.h" // NOLINT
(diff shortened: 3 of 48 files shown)


floam:
We ought to be using the std:: namespace if using the `cfoo` headers.

floam:
e.g. what I did in https://github.com/fish-shell/fish-shell/commit/aaacdb89b628eb37bb2e7c461662a677baf9b23e#diff-eebc687857eb591b1beb2a6c87379378

zanchey:
And there's symbols on some platforms that aren't in `std::`; given the probable fallout is there significant benefit to this kind of cleanup?

floam:
> And there's symbols on some platforms that aren't in std::
Still the case since the C++11 switch? Since that change we require a newer Clang/GCC than we had in the past. I'm surprised a libc++ or libstdc++ with C++11 officially implemented would still have that happen.

zanchey:
3b0f642de9fef3bac3ef98d8bcecdbdfb81e945b and 1293cd8b6a are the commits that spring to mind; I don't mind testing this on RHEL 6 at some stage as that's probably the oldest platform we support, but I'm not clear on what the actual benefit is.

krobelus:
Regarding a switch to `std::`: if this would affect many lines I would only do it given a good reason as to not unnecessarily clutter `git blame` output.

neheb:
@zanchey the former (which I authored) deals with the libc not supporting _GLIBCXX_USE_C99_MATH_TR1. Fish doesn't actually compile with that libc because of missing locale_t.

In any case, using namespace std; is found nowhere in the codebase. As the C++ headers include the C headers, which is a relatively moot point.

zanchey:
Works fine on RHEL 6; @Mno-hime, are you able to test on OpenIndiana?

Mno-hime:
@zanchey Works me with GCC 6 & 8.

zanchey:
OK, that's useful to know.

I'm still ambivalent about merging; I realise that this is unhelpful "stop energy" but I don't understand what the benefit is. Am I missing something, or is it more than cleanup of technically-deprecated but unlikely to ever stop working headers?

ridiculousfish:
Let's not do this, it does not appear to provide any practical benefit. It's not clear to me which functions would get put in `std` with this change anyways - I think it's C functions but not Unix ones? 

(Incidentally `using namespace std;` would be a grave mistake).

Thank you for opening this anyways - hygiene is important!

PR #6516, opened 2020-01-21
Remove `dropbox` completions

ammgws:
## Description

These completions look to be for the [dropbox.py script](https://www.dropbox.com/download?dl=packages/dropbox.py), however as far as I know this script isn't necessarily found on $PATH as `dropbox`. In fact, the Dropbox [site](https://www.dropbox.com/install-linux) says:

> Download this Python script to control Dropbox from the command line. For easy access, put a symlink to the script anywhere in your PATH.

On my Arch Linux system the script is available on the AUR as `dropbox-cli` and on path also as `dropbox-cli`, and the desktop app/daemon is found as `dropbox`. Thus this completion script unexpectedly causes dropbox to open when the completion script is run:

image

It looks like on Ubuntu this script is available as `dropbox` according to [this](http://manpages.ubuntu.com/manpages/bionic/man1/dropbox.1.html).

I guess we could check the shebang of `dropbox` and if it is python then it is most likely the dropbox.py script, but it seemed kinda kludgey so my thought is to just remove this completion altogether.

## TODOs:


faho:
Yeah, seems like a good idea, merged.

PR #6517, opened 2020-01-21
Fix encodings completion script for FreeBSD/Cygwin

ammgws:
## Description

`iconv --list` output format is different between the GNU and FreeBSD/Cygwin verisons.

On GNU (when stdout is redirected): prints encodings separated by `//`

On Cygwin/FreeBSD: prints encodings separated by '\n', however multiple encodings can appear on the same line. 
It seems to be grouping aliases together, e.g. IBM-1097, IBM1097 CSIIBM1097, CP1097 appear on the same line which breaks the completion script.

Fixes issue #

## TODOs:


faho:
Thanks, merged.

@zanchey This seems to me like it would be okay for 3.1.0, but we could also wait for e.g. a 3.1.1 release. How do you want this handled? Personally I'd rather not delay 3.1.0 any longer, so if that takes not adding anything else then so be it.

zanchey:
Go ahead if you would like to push it to Integration_3.1.0

Issue #243, opened 2012-07-23
Variables can not be used as command substitution indices


maxfl:
I have found that the following code doesn't work:

``` sh
set n 1
echo (seq 10)[$n]
```

but the following code works ok:

``` sh
set n 1
echo $PATH[$n]
```

While it is a rarely used case, I think that it's very confusing, that square brackets are expanded in a different way in these cases.

krader1961:
Relabeling as an enhancement because the current behavior is not obviously wrong or in contradiction of the documentation. The proposed change, however, is desirable.

mqudsi:
The problem is that fish does not recursively evaluate indices, the literal contents are checked for an integer value.

Same thing with process substitution:

```fish
mqudsi@ZBook ~/r/fish-shell> echo (seq 10)[(seq 1)]
fish: Invalid index value
echo (seq 10)[(seq 1)]
              ^
```

santpent:
Is this the relevant line in the documentation?

> Note that variables can be used as indices for expansion of variables, but not command substitution.

https://fishshell.com/docs/current/#expand-index-range

Just wanted to add a common (for me) use case which causes the same error:
```
set idx 2
echo (echo 'a b c d' | string split ' ' -- )[$idx]
```

mqudsi:
@santpent Yes, that is an example of command substitution indexing.

PR #6518, opened 2020-01-21
Stringify gpg completions

ammgws:
## Description

Use builtin string function instead of sed.

P.S.: Is it possible to one-liner this using `string`?

## TODOs:

### DIFF
--- a/share/functions/__fish_complete_gpg_key_id.fish
+++ b/share/functions/__fish_complete_gpg_key_id.fish
@@ -5,13 +5,15 @@ function __fish_complete_gpg_key_id -d 'Complete using gpg key ids' -a __fish_co
     set -l keyid
     $__fish_complete_gpg_command --list-keys --with-colons | while read garbage
         switch $garbage
-            # Extract user ids
+            # Extract user ids (note: gpg escapes colons as '\x3a')
             case "uid*"
-                echo $garbage | cut -d ":" -f 10 | sed -e "s/\\\x3a/:/g" | read uid
+                set -l __uid (string split ":" -- $garbage)
+                set uid (string replace -a '\x3a' ':' -- $__uuid[10])
                 printf "%s\t%s\n" $keyid $uid
-                # Extract key fingerprints (no subkeys)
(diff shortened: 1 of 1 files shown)


share/functions/__fish_complete_gpg_key_id.fish
```diff
@@ -5,13 +5,15 @@ function __fish_complete_gpg_key_id -d 'Complete using gpg key ids' -a __fish_co
     set -l keyid
     $__fish_complete_gpg_command --list-keys --with-colons | while read garbage
         switch $garbage
-            # Extract user ids
+            # Extract user ids (note: gpg escapes colons as '\x3a')
             case "uid*"
-                echo $garbage | cut -d ":" -f 10 | sed -e "s/\\\x3a/:/g" | read uid
+                set -l __uid (echo $garbage | string split ":")
```
faho: String doesn't need to read from stdin. Use `string split : -- $garbage`.

faho:
>P.S.: Is it possible to one-liner this using string?

Technically anything's possible with regex. You can match something like `([^:]*:){10}` to get the tenth field (or possibly the eleventh, I've not tried this), and that replaces that `cut`.

Easily? No, we lack the field selection for `string split`. I could swear there's an issue open about it, but I can't find it. It would be useful.

share/functions/__fish_complete_gpg_key_id.fish
ammgws: Oopsie. Cheers!

ammgws:
> Easily? No, we lack the field selection for `string split`. I could swear there's an issue open about it, but I can't find it. It would be useful.

I guess #243?

faho:
Yeah, that works. Merged, thanks!

ohk2kt3t4:
> Easily? No, we lack the field selection for `string split`. I could swear there's an issue open about it, but I can't find it. It would be useful.

However at least this seems to work in my terminal:

`set uid (string replace -a '\x3a' ':' -- (string split ":" -- $garbage)[10])`

Am I overlooking something?

PR #6519, opened 2020-01-21
Stringify netctl completions

ammgws:
## Description

Use string built in over sed

Fixes issue #

## TODOs:

### DIFF
--- a/share/completions/netctl.fish
+++ b/share/completions/netctl.fish
@@ -1,7 +1,7 @@
 set -l cmds list store restore stop-all start stop restart switch-to status enable disable reenable is-enabled edit
 
 function __fish_netctl_get_profiles
-    command netctl list | sed -e 's/^[ \t*]*//'
+    command netctl list | string trim -l -c ' *'\t
 end
 
 complete -f -c netctl -l help -d 'Display help'


share/completions/netctl.fish
```diff
@@ -1,7 +1,7 @@
 set -l cmds list store restore stop-all start stop restart switch-to status enable disable reenable is-enabled edit
 
 function __fish_netctl_get_profiles
-    command netctl list | sed -e 's/^[ \t*]*//'
+    command netctl list | string replace -r '^[ \t*]*' ''
```
faho: I'm not entirely certain that `*` in the brackets does the right thing.

Do you happen to know what `netctl list` output looks like?

In this `string` it would interpret the `*` as another character in the class to replace, in the `sed` it doesn't seem to, so I'm not quite sure what it was supposed to do there.

ammgws: netctl list output is prepended with *, space, or tab:
```
* ethernet-dhcp
  ethernet-custom
```    

Both the sed and string versions return:
```
ethernet-dhcp
ethernet-custom
```

faho: Ah, okay. In that case `string trim` would probably express the intent more clearly:

```fish
command netctl list | string trim -l -c ' *'\t
```

trims the characters space, "*" and tab (outside of the quotes so it's replaced with a tab character then passed to string) from the left of the string.

faho:
Merged, thanks!

PR #6523, opened 2020-01-21
Get the list of vmctl VMs in a less fragile way.

qbit:
## Description

Currently the code that parses the list of VMs gets things a bit incorrect as a column was added.

This makes the vmctl completion always grab the last column for each vm line.


faho:
Beautiful!

Issue #6520, opened 2020-01-21
fishify functions/completions


ammgws:
The following functions/completions use `sed` etc in some way that may be able to be sensibly replaced by fish builtins such as `string`. Opening this issue so can track which ones have been completed. 

`sed`:
- [x] completions/abook.fish (eba099bf775df8e52eec3dc9f5af7762369ca33a)
- [x] completions/apropos.fish (69afb1b560b1c15f4830a599ca83c785f45103b4)
- [ ] completions/bundle.fish
- [ ] completions/emerge.fish
- [ ] completions/eselect.fish
- [ ] completions/highlight.fish
- [ ] completions/killall.fish
- [ ] completions/mplayer.fish
- [ ] completions/mvn.fish
- [x] completions/netctl.fish 
- [x] completions/open.fish (b1bbb58af6f04aa13462e386d11373cde9c4430e)
- [ ] completions/pactl.fish
- [ ] completions/perl.fish
- [ ] completions/setxkbmap.fish
- [x] completions/ssh.fish
- [x] completions/vared.fish (7d89cadd583ef36b21ae6c9dc9247fc83cf1b0cc)
- [x] functions/__fish_complete_gpg_key_id.fish
- [x] functions/__fish_make_completion_signals.fish 
- [ ] functions/__fish_print_modules.fish
- [ ] functions/__fish_print_packages.fish
- [x] functions/__fish_pwd.fish
- [ ] functions/fish_svn_prompt.fish

`awk`:
- [ ] completions/adb.fish
- [ ] completions/cwebp.fish
- [ ] completions/ezjail-admin.fish
- [ ] completions/find.fish
- [ ] completions/grunt.fish
- [ ] completions/helm.fish
- [ ] completions/heroku.fish
- [ ] completions/lsusb.fish
- [ ] completions/make.fish
- [ ] completions/p4.fish
- [ ] completions/psql.fish
- [x] completions/rmmod.fish (6004d04da68b139c69724b1b924192baeaaca27b)
- [ ] completions/vim-addons.fish
- [ ] functions/__fish_complete_convert_options.fish
- [ ] functions/__fish_complete_man.fish
- [ ] functions/__fish_describe_command.fish
- [ ] functions/__fish_print_addresses.fish
- [ ] functions/__fish_print_packages.fish

`tr`:
- [ ] completions/aura.fish
- [ ] completions/grunt.fish
- [ ] completions/pactree.fish
- [ ] completions/pkgfile.fish
[...]

PR #6526, opened 2020-01-22
Stringify version parsing in help.fish

ammgws:
## Description

One of many tasks in #6520. Naturally we should parse the fish version using fish functions ;)

## TODOs:


zanchey:
Im

zanchey:
I don't think this is clearer. `cut` has an easy-to-understand interface, and the regular expression is reasonably impenetrable. This is not performance sensitive and I don't think there's a need to avoid external commands.

ridiculousfish:
I think we'll pass on this; thank you for the PR nonetheless!

Issue #5474, opened 2019-01-03
`string replace` in regex-mode requires waaaay too many backslashes for the replacement


faho:
(I could swear we already discussed this somewhere, but I can't find it. Anyway I'm pretty sure it's been closed, and I want to rerun it)

`string replace -r` currently does an additional unescaping pass on the replacement (see the `interpret_escapes` function in builtin_string.cpp), and just the replacement.

That means  if you e.g. want to add a backslash to any a or b, you'll have to do this:

```fish
string replace -ra '([ab])' '\\\\\\\$1' a
```

with one or two backslashes, it just prints "a", with three to six backslashes, it prints `$1`. With seven to ten, it prints the desired `\a`.

If we removed `interpret_escapes` instead:

- one or two backslashes prints `$1`

- three or four prints `\a`

- five or six prints `\$1`

- seven or eight prints `\\a`

This is explainable - `\` is an escape character in fish even in single-quotes, so `\\` is reduced to just `\` before it reaches fish. `\$` isn't reduced, because `$` doesn't need to be escaped. So both `\$` and `\\$` result in `\$` being passed to fish.

Similarly, `\\\$` and `\\\\$` reduce to `\\$`, which is taken by string to mean "a literal backslash, and then the dollar replacement".

The major issue with this is backwards-compatibility. There is, as far as I can see, no way to make a common subset to easily write expressions that work with both interpretations.

So we'd have to do something:

-  `string replace --real-regex` or `string replace --regex --replacement-without-annoying-backslashes`
- or a feature flag
- or break stuff and tell people to check $version

I still feel like this is worth it, simply because the current behavior is ridiculous.

faho:
I'm pretty sure we'd have to fix the paste escaping in fish_clipboard_paste.

Also of note is that the tests currently do _not_ cover this.

zanchey:
> `string replace -r` currently does an additional unescaping pass on the replacement

Any idea why?

faho:
It's been that way since string was added, so I'm assuming it's pure oversight.

There's no logic to it that I can see.

floam:
https://github.com/fish-shell/fish-shell/issues/3381 I think talked about the second unescape.

faho:
Ah, indeed, it did.

And... well, there's no good reason why we do the second unescape, and there's mentions of backwards compatibility, and of breakage. So basically what I expected.

[...]

PR #6527, opened 2020-01-22
Stringify __fish_pwd.fish

ammgws:
## Description

One of many tasks from #6520 

## TODOs:

### DIFF
--- a/share/functions/__fish_pwd.fish
+++ b/share/functions/__fish_pwd.fish
@@ -1,7 +1,12 @@
 switch (uname)
     case 'CYGWIN_*'
         function __fish_pwd --description "Show current path"
-            pwd | sed -e 's-^/cygdrive/\(.\)/\?-\u\1:/-'
+            if status test-feature regex-easyesc
+                pwd | string replace -r '^/cygdrive/(.)?' '\U$1:'
+            else
+                # TODO: Remove this once regex-easyesc becomes the default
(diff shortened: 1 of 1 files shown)


share/functions/__fish_pwd.fish
```diff
@@ -1,10 +1,10 @@
 switch (uname)
     case 'CYGWIN_*'
         function __fish_pwd --description "Show current path"
-            pwd | sed -e 's-^/cygdrive/\(.\)/\?-\u\1:/-'
+            pwd | string replace -r '^/cygdrive/(.)?' '\\\U$1:'
```
faho: Unfortunately this is affected by #5474 - we run an additional round of unescaping on the replacement argument. I felt it was too risky to change, so I implemented the fix behind a feature flag (to be clear this means it *will* become the default and eventually only behavior - it's not a configuration option you can choose to taste). For the time being this needs to be

```fish
if status test-feature regex-easyesc
    pwd | string replace -r '^/cygdrive/(.)?' '\U$1:'
else
    pwd | string replace -r '^/cygdrive/(.)?' '\\\U$1:'
end
```

ammgws: Ahhhh... I had a hard time figuring that one out.

Thanks for the explanation.

faho:
Merged, thanks!

PR #6529, opened 2020-01-22
Stringify ssh completions

ammgws:
## Description

Fixes part of issue #6520

## TODOs:

### DIFF
--- a/share/completions/ssh.fish
+++ b/share/completions/ssh.fish
@@ -17,8 +17,12 @@ complete -c ssh -n 'test (__fish_number_of_cmd_args_wo_opts) -ge 2' -d "Command 
 
 complete -c ssh -s a -d "Disables forwarding of the authentication agent"
 complete -c ssh -s A -d "Enables forwarding of the authentication agent"
-# TODO: Improve this since /proc/net/arp is not POSIX compliant.
-complete -x -c ssh -s b -d "Interface to transmit from" -a "(string match -r '^(\d+.\d+.\d+.\d+)' < /proc/net/arp)"
+
+if type -q jq
+    ip -json -family inet address show | jq -r '.[] | select(.ifname != "lo") | .addr_info[0] | "\(.local) \(.label)"' | while read -l ipaddr intf
+        complete -x -c ssh -s b -d "Bind to $intf" -a "$ipaddr"
+    end
+end
 
 complete -x -c ssh -s e -d "Escape character" -a "\^ none"
 complete -c ssh -s f -d "Go to background"
--- a/share/completions/ssh.fish
+++ b/share/completions/ssh.fish
@@ -18,11 +18,7 @@ complete -c ssh -n 'test (__fish_number_of_cmd_args_wo_opts) -ge 2' -d "Command 
 complete -c ssh -s a -d "Disables forwarding of the authentication agent"
 complete -c ssh -s A -d "Enables forwarding of the authentication agent"
 
-if type -q jq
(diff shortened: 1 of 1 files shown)


share/completions/ssh.fish
```diff
@@ -18,9 +18,7 @@ complete -c ssh -n 'test (__fish_number_of_cmd_args_wo_opts) -ge 2' -d "Command
 complete -c ssh -s a -d "Disables forwarding of the authentication agent"
 complete -c ssh -s A -d "Enables forwarding of the authentication agent"
 # TODO: Improve this since /proc/net/arp is not POSIX compliant.
-complete -x -c ssh -s b -d "Interface to transmit from" -a "
-(cut -d ' ' -f 1 /proc/net/arp 2>/dev/null | string match -r -v '^IP')
-"
+complete -x -c ssh -s b -d "Interface to transmit from" -a "(string match -r '^(\d+.\d+.\d+.\d+)' < /proc/net/arp)"
```
faho: This errors out if /proc/net/arp doesn't exist, and I'm pretty sure that's the case on most, maybe even all, non-linux platforms.

That was what redirecting stderr was about - it would error about not finding the file and not print anything, harmlessly.

---

Also: The parens around the ip here cause it to double-print the result.

(Also I can confirm /proc/net/arp does not contain ipv6 addresses, so this appears to *work* for ipv4. We'd have to find something better for this to add ipv6 support)

ammgws: Ah so that is what the POSIX comment was referring to.

ammgws: @faho
 Does this completion even make sense in the first place? 

From what I understand you would use the -b option when your local machine has multiple network interfaces and you want to specify which one should be used for the ssh connection.

However I checked the contents of /proc/net/arp and it contains the addresses for remote machines on my network, so it doesn't make sense to me to be using them as completions for -b.

Have I misunderstood something?

zanchey: Yeah, I don't think this is right. I'm on mobile but I think the output of `ip` needs examining rather than spelunking through `/proc` or `/sys`.

ammgws: `ip` has a nice JSON output mode so if `jq` is acceptable then we can get the IPs in a predictable manner like this:

` ip -json -family inet address show | jq '.[] | select(.ifname != "lo") | .addr_info[0] | .local'`

This returns the IPv4 address for all non-loopback network interfaces on the local machine.

It would be nice if the network interface label (e.g. enp3s0, eth0, etc) could be used for `complete`'s description field, but as far as I can see that's not possible.

faho: The description is weird - `ssh -b` takes an *address*.

I think this might actually just need `__fish_print_addresses`.

>ip has a nice JSON output mode so if jq is acceptable then we can get the IPs in a predictable manner like this:

It is not.

It's quite slow and not typically installed, and there is *absolutely* no need to involve json here. 
 `ip`s oneline mode is entirely sufficient, and easy to parse using builtins.

(the places where we've needed json so far we've fallen back on using python - it's actually *faster* than jq and it's already an optional dependency)

faho: Also there is no need to limit this to ipv4.

ammgws: __fish_print_addresses was exactly what was needed.

Do you think it is worthwhile to add the interface label to each description as in the jq example above? Since unless you have a memorised/static IP you might not know which interface the IP is for.

faho: Sure. Just add them to each line after a tab - the format is `completion\tdescription\n`.

Not necessary for this PR tho, so I've merged it now. Thanks!

zanchey:
Cherry-picked to 3.1.1 as 5f6e43df6

PR #6530, opened 2020-01-22
chore: Rename functions to match others

ammgws:
## Description

Fixes:
- systemctl function only had one underscore
- portage print functions were the odd one out regarding the word order

## TODOs:


faho:
Yeah, _fish_systemctl was meant to usher in a bold new era of us saving on one useless underscore.

It's never happened, mostly because that kind of mass-renaming isn't all that useful.

So, yeah, let's give up.

Issue #6124, opened 2019-09-19
A better test builtin: is


zanchey:
`test` sucks (#1337, #5186, #5947, #6114). Plus I always have to look up the manpage because I can never remember what the options do. Its one saving grace is its compliance with POSIX, so rather than enhancing it beyond that, can we do better? Bash/zsh has `[[`, which is mostly backward-compatible, but I think we should take a different approach.

I propose a new builtin, which I will call `is` for now.

Some examples:

```fish
is file $somefilename; or touch $somefilename

is exists /etc/debian_release; or rm -rf /

if is sgid /usr/bin/screen ...

if is 1 greaterthan 2 ... # or gt for shorthand?

if is 1 = 1 ...

if is $FISH_VERSION versionlessthan 3.0.0
    echo go fishing!
end

```

Replaces #6065. Should include #3589.

Absolutely must not include combining operators.

Should not include operations available elsewhere (eg `isatty`, `string`, `contains`); I'd be tempted to elide string operations entirely in favour of `string match --entire` and `string length`, though perhaps some shorthand could be added.

One problem then is that `is` is then  numeric comparisons and file operations, which are not at all related; perhaps it could be broken out into `isnum` and `ispath`?

Name open to debate (though `is` doesn't collide with any binaries in Debian, at least).

faho:
We need to be careful to not redo `test`s argument parsing bogosities.

You've already left out the combiners, and we should also leave out the no-operator form (`test somestring` being the same as `test -n somestring` is the single worst bit about test).

I've had some ideas over the years, including always requiring three arguments:

- `is a = b` tests string equality
- No `test -n` or `test -z`, instead you'll have to use `is a = ''` and `not is a = ''`

But that causes problems with e.g. unquoted variables - what would `is $a = ''` do if $a is list-empty?

It could error out, or since there *is* a valid operator it could assume the other argument is empty. Erroring seems a bit excessive.

Or we could require operators first, like you've done here.

Then there's the question of what happens if an argument looks like an operator. If we used `test`s `-eq` etc numeric operators (unfortunately the intuitive `<` et al would require more than a builtin), we could get something like `is $a -eq 4`, and then $a could start with a "-" like `-4`, or it could even be `-eq`.

I don't think that's a huge problem - if you forget to add an operator *and* the variable looks like one you're gonna have a bad time.

[...]

PR #6531, opened 2020-01-22
Use `string length -q` over `test -n`

ammgws:
## Description

Hopefully this isn't controversial.

Reason for changing:
- promote "fishy" ways by using it in our own code (assuming this change is more fishy than current)
- test -n $var returns true if $var is list-empty (or currently undefined)
- current docs for `string length` mention the commands are "equivalent"

## TODOs:

### DIFF
--- a/share/completions/adb.fish
+++ b/share/completions/adb.fish
@@ -24,7 +24,7 @@ function __fish_adb_run_command -d 'Runs adb with any -s parameters already give
     set -l cmd (commandline -poc)
     set -e cmd[1]
     for i in $cmd
-        if test -n "$sopt_is_next"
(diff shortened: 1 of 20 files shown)


faho:
To be honest, I'm not enthusiastic enough about this that I'd want to merge it.

For one, as you can see from the test failures replacing `test -n "$var"` with `string length -q "$var"` isn't actually enough. It would have to be `string length -q -- "$var"`.

And even that has little gain, but some risk associated with the change. I don't think the churn is worth it.

Let's implement something like #6124, and then we can make this nicer.

Hopefully that makes sense to you, sorry!

ammgws:
That's totally fine :)

PR #6535, opened 2020-01-25
Add interface label/name to the completions for `ssh -b`

ammgws:
image

### DIFF
--- a/share/completions/ssh.fish
+++ b/share/completions/ssh.fish
@@ -18,7 +18,19 @@ complete -c ssh -n 'test (__fish_number_of_cmd_args_wo_opts) -ge 2' -d "Command 
 complete -c ssh -s a -d "Disables forwarding of the authentication agent"
 complete -c ssh -s A -d "Enables forwarding of the authentication agent"
 
-complete -x -c ssh -s b -d "Local address to bind to" -a "(__fish_print_addresses)"
+
+function __ssh_print_local_addresses_with_labels
+    if command -sq ip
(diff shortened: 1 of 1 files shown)


share/completions/ssh.fish
```diff
@@ -18,7 +18,10 @@ complete -c ssh -n 'test (__fish_number_of_cmd_args_wo_opts) -ge 2' -d "Command
 complete -c ssh -s a -d "Disables forwarding of the authentication agent"
 complete -c ssh -s A -d "Enables forwarding of the authentication agent"
 
-complete -x -c ssh -s b -d "Local address to bind to" -a "(__fish_print_addresses)"
+function __ssh_print_local_addresses_with_labels
+  ip --oneline address | string replace -r '\d+:\s+(\S+)\s+\S+\s+(.+)/.*' '$2\t$1'
```
faho: And OSes without `ip`? That includes not just macOS and the BSDs, but also old and stripped down linux.

ammgws: > And OSes without `ip`? That includes not just macOS and the BSDs, but also old and stripped down linux.

Sorry about that. I will implement just IP address support for non-Linux for now until I can get access to a BSD machine to work further on this.

floam: Should be able to use grep/string replace on `ifconfig` output, but be warned the output differs on different OSes.

ammgws: Yeah it's probably not worth the effort at the moment.

faho:
Yeah, merged this for now. Thanks!

PR #6140, opened 2019-09-25
Add default \es keybinding for "prepend sudo"

gpanders:
## Description

This creates a new function `__fish_prepend_sudo` and a new default keybinding `Alt+S` that prepends `sudo ` to the beginning of your current commandline.

Example:
```fish
$ vim /etc/hosts
```
pressing `Alt+S` yields
```fish
$ sudo vim /etc/hosts
```

This does not move the cursor.

This was inspired by the comment [here](https://github.com/fish-shell/fish-shell/issues/288#issuecomment-16552504).

Personally, I think this might allay some of the agitation around the lack of the `!!` operator in fish, since `sudo !!` can now effectively be done with `Ctrl+p Alt+s`, which is even easier!

PR #6542, opened 2020-01-27
__fish_prepend_sudo: Don't move the relative position of the cursor

sjohannes:
## Description

At the moment the "prepend sudo" (Alt-S) functionality always sets the cursor to the end of the line. This changes it to restore the relative position of the cursor.

The functionality was added in https://github.com/fish-shell/fish-shell/pull/6140 (0f802ea). The PR description says "This does not move the cursor", which suggests that the current behaviour is a mistake.

For example, let's say the user entered `ls /root` and then moved the cursor to, e.g.

    ls |/root

where | is the cursor position. Running the function currently turns it to

    sudo ls /root|

whereas after this change it is

    sudo ls |/root

## TODOs:


krobelus:
Yes, that's definitely an improvement. thank you!

PR #6547, opened 2020-01-28
Added Debian installation manual

opensource-assist:


zanchey:
So the reason we don't currently include the instruction to "just `apt-get install fish`" is that it will get you a very out of date version on most platforms, hence the use of our own packages. Perhaps instead we could change the text to read "to get the latest version of fish".

ridiculousfish:
Closing, thanks for opening this PR nonetheless

Issue #6209, opened 2019-10-15
Completions for apt install do not support filenames


int19h:
(fish 3.0.2 on Debian 10.1, no customizations)

`apt install` supports filenames as targets, not just package names. If the name can be interpreted as a package name, it is, even if it's also a valid filename. But when it cannot be a package name, e.g. because it contains slashes, it's treated as a filename. Thus, one can say `apt install ./foo.deb`.

Fish completion logic for `apt` only completes available package names, however.

PR #6552, opened 2020-01-29
apt: Add support for filename completions

ammgws:
Attempt at solving #6209 

Possible issue with this implementation: the filename completion adds all subdirectories of $PWD to the completion list which may not be desirable? Also was hoping to list .deb files in the current directory before the generated list of packages, however they are intermingled at the moment.

### DIFF
--- a/share/completions/apt.fish
+++ b/share/completions/apt.fish
@@ -3,8 +3,9 @@
 set -l all_subcmds update upgrade full-upgrade search list install show remove edit-sources purge changelog autoremove depends rdepends
 set -l pkg_subcmds install upgrade full-upgrade show search purge changelog policy depends rdepends
 set -l installed_pkg_subcmds remove
+set -l handle_file_pkg_subcmds install
 
-function __fish_apt_subcommand
+function __fish_apt_subcommand --no-scope-shadowing
     set subcommand $argv[1]
     set -e argv[1]
     complete -f -c apt -n "not __fish_seen_subcommand_from $all_subcmds" -a $subcommand $argv
@@ -16,17 +17,20 @@ function __fish_apt_option
     complete -f -c apt -n "__fish_seen_subcommand_from $subcommand" $argv
 end
 
-complete -c apt -n "__fish_seen_subcommand_from $pkg_subcmds" -a '(__fish_print_packages | head -n 250)'
+complete -c apt -f
+
+complete -k -c apt -n "__fish_seen_subcommand_from $pkg_subcmds" -a '(__fish_print_packages | head -n 250 | sort)'
 complete -c apt -n "__fish_seen_subcommand_from $installed_pkg_subcmds" -a '(__fish_print_packages --installed | string match -re -- "(?:\\b|_)"(commandline -ct | string escape --style=regex) | head -n 250)' -d 'Package'
+complete -k -c apt -n "__fish_seen_subcommand_from $handle_file_pkg_subcmds" -a '(__fish_complete_suffix .deb)'
 
 # Support flags
-complete -x -f -c apt -s h -l help -d 'Display help'
-complete -x -f -c apt -s v -l version -d 'Display version and exit'
+complete -f -c apt -s h -l help -d 'Display help'
+complete -f -c apt -s v -l version -d 'Display version and exit'
 
 # General options
-complete -f -c apt -s o -l option -d 'Set a configuration option'
-complete -f -c apt -s c -l config-file -d 'Configuration file'
-complete -f -c apt -s t -d 'Target release'
+complete -x -c apt -s o -l option -d 'Set a configuration option'
+complete -r -c apt -s c -l config-file -d 'Configuration file'
+complete -x -c apt -s t -d 'Target release'
 
 # List
 __fish_apt_subcommand list -d 'List packages'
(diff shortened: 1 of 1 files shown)


zanchey:
> filename completion adds all subdirectories of $PWD to the completion list which may not be desirable

This is by design - `__fish_complete_suffix` does not descend into subdirectories to check for files which match the suffix, listing them as completions instead.

faho:
> This is by design - __fish_complete_suffix does not descend into subdirectories to check for files which match the suffix, listing them as completions instead.

Indeed. The reason being that we can't afford to recurse, because that *will* take minutes in some very common cases (people doing `apt install /<TAB>`), and we can't just leave all directories out because that would make it hard to get to files you know are there.

>Also was hoping to list .deb files in the current directory before the generated list of packages, however they are intermingled at the moment.

This seems like an issue with "--keep-order". I think it's only keeping the grouping for things that are marked "--keep-order", intermingling those elements that aren't.

As it stands, I think the correct solution is to remove the "-k" flag, because we do want the debs to be sorted - we just don't want them to be grouped, and that might necessitate another flag. We'll have to think some about this.

ohk2kt3t4:
image

This is what it looks like. (on WSL)

mqudsi:
You can keep the `-k` but sort your completions before giving them back to fish.

ohk2kt3t4:
In this case the output of `__fish_complete_suffix .deb` is already sorted on my system, if that's what you meant.

From manpages for `complete`:
> -k or --keep-order preserves the order of the OPTION_ARGUMENTS specified via -a or --arguments instead of sorting alphabetically. Multiple complete calls with -k result in arguments of the later ones displayed first.
My understanding is that with the following two commands in the order below will result in the file completions showing up first, but it isn't happening.
```complete -c apt -n "__fish_seen_subcommand_from $pkg_subcmds" -a '(__fish_print_packages | head -n 250)'
complete -k -c apt -n "__fish_seen_subcommand_from $handle_file_pkg_subcmds" -a '(__fish_complete_suffix .deb)'
```

Also tried putting it all in one command but it still ends up mixed together.
`complete -k -c apt -n "__fish_seen_subcommand_from $install_pkg_subcmds" -a '(__fish_complete_suffix .deb | sort; and __fish_print_packages | head -n 250 | sort)'`

mqudsi:
I was saying you also need to add -k to the other completion.

ohk2kt3t4:
Output is still mixed together:
```
complete -k -c apt -n "__fish_seen_subcommand_from $pkg_subcmds" -a '(__fish_print_packages | head -n 250 | sort)'
complete -k -c apt -n "__fish_seen_subcommand_from $handle_file_pkg_subcmds" -a '(__fish_complete_suffix .deb | sort)'
```
image

Surely the single line command in my last comment should work too, but somehow it doesn't...

ohk2kt3t4:
Actually they are sorted correctly when testing using `--do-complete` as shown below. Seems like whatever function that prints the completions is sorting them.

complete -C"apt install "
```
rdepends        List package reverse dependencies
depends List package dependencies
policy  Display source or package priorities
autoremove      Remove packages no longer needed as dependencies
changelog       Download and display package changelog
purge   Remove packages and delete their config files
full-upgrade    Upgrade packages, removing others when needed
upgrade Upgrade packages
update  Update package list
edit-sources    Edit sources list
remove  Remove packages
install Install packages
show    Show package information
search  Search for packages
list    List packages
3D Objects/
AppData/
Cache/
Contacts/
Desktop/
Documents/
Downloads/
Favorites/
IntelGraphicsProfiles/
Links/
MicrosoftEdgeBackups/
Music/
Pictures/
Saved Games/
Searches/
Videos/
VirtualBox VMs/
apt-completions-test.deb
done/
fish-shell/
scoop/
wsl_dev/
accountsservice query and manipulate user account information
acct    GNU Accounting utilities for process and login accounting
acl     access control list - utilities
acpi-support    scripts for handling many ACPI events
acpid   Advanced Configuration and Power Interface event daemon
adduser add and remove users and groups
...
```

mqudsi:
That looks like a regression....

ridiculousfish:
Opened #6568 to investigate if it's a regression.

ohk2kt3t4:
Sorry for the fuss, adding `-k` to [this](https://github.com/fish-shell/fish-shell/blob/8f4797a38bf154f03f3ea5ca858d99cdee306442/share/completions/apt.fish#L10) line solved the issue regarding ordering.

image

ohk2kt3t4:
File completions for the install command are now working well:
image

I also fixed the issue of apt subcommands being falsely given as completions. 
This is before it was fixed:
image
and
image

One last thing I want to try improve is that pressing tab after `apt list` etc gives file completions, but I think it would be better if it just didn't suggest anything at all. Is there a way to do this with the complete command? All the complete commands have -f (--no-files) but it still ends up doing file completions...
image

mqudsi:
> Is there a way to do this with the complete command?

You would add a predicate to your `complete` instance that generates the filename completions with `-n 'not seen_subcommand_from list'`

ohk2kt3t4:
Currently for the `list` subcommand:
```
__fish_apt_subcommand list -d 'List packages'
complete -f -k -c apt -n "not __fish_seen_subcommand_from $__fish_apt_all_subcmds" -a $subcommand $argv
```
The `complete` is used with `-f` which I thought means fish won't offer any filename completions after `apt list`, however it does.

krobelus:
The problem was that none of the the options with `-f` was enabled (w.r.t. their `-n`-conditions) when a subcommand was given. Fixed now.

ohk2kt3t4:
That did the trick! Thanks!

krobelus:
Merged with some minor modifications, thanks!

PR #6560, opened 2020-01-31
Fix broken links to index.html in cmd docs

ammgws:
Everything linking back to index.html in the parent directory is broken because it gets resolved to cmds/index.html instead of just index.html.

Reproduced on master/3.1b on two different systems (Arch and WSL Ubuntu).

On phone so can't produce a better screenshot but here is an example of a broken link:
Screenshot_20200131-113748

### DIFF
--- a/sphinx_doc_src/cmds/and.rst
+++ b/sphinx_doc_src/cmds/and.rst
@@ -18,7 +18,7 @@ Description
 
 ``and`` statements may be used as part of the condition in an :ref:`while <cmd-while>` or :ref:`if <cmd-if>` block.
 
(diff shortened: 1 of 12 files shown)


ammgws:
Have tested this fix on the WSL system, so it should be good to go.

ammgws:
There might be some links to tutorial.html etc as well, I'll get onto that in a bit.

zanchey:
The second commit wasn't quite right, so I've fixed it up and pushed to master and Integration_3.1.0 as follows - thanks!

   578cde6ec..9b4cb28c8  Integration_3.1.0 -> Integration_3.1.0
   3f8907d7c..ce61c745a  master -> master

Issue #6270, opened 2019-10-31
macOS 10.15: slow command autocompletion


dimaqq:
Ever since Catalina update, I experience ~10s delays every now and again (not often, but typically at least once after boot and maybe less than once an hour).

For example, I have `poetry` installed ( https://github.com/sdispater/poetry ) the point is only that it installs something "executable".

When I type `poe` in the shell, fish tries to show dark gray autocomplete, in my case typically `poetry shell`. Usually it's instantaneous, but sometimes it takes 10s during which I cannot type anything... or rather I can type by terminal is not updated.

My setup is `iTerm2` + `fish` from pkg installer + usual dev things, git, yarn, npm, python3.8, poetry, vim, etc.

```
> fish --version
fish, version 3.0.2
```

I'm not actually entirely sure what causes the sudden lag and I don't know if I can in fact blame it on `fish`, but I'd love to figure that out :)

When I've upgraded to Catalina, I had to manually "allow access" for iterm2 to Downloads, Documents, etc... I wonder if that's connected?

Alternatively, could there be something with `hostname` occasionally lagging?

krobelus:
If you can reproduce this with fish launched as `fish --profile my-profile-file`, you can look at the breakdown in `my-profile-file` to see what is taking so much time.

If you use the command `hostname` you can try the variable `$hostname` instead but that should not make a difference.

Fingel:
I believe I have the same issue. After upgrading to Catalina, I noticed completions became very slow. Today I decided to look into it. I launched fish with `fish -d5`. I noticed anytime I tried to tab-complete something, I'd get this output:

```
<4> fish: select_try on fd 7
<4> fish: select_try hit timeout
<4> fish: select_try: no fds returned valid data within the timeout
```
Repeated like 50 times. Whether or not that is related to the next issue, I'm not sure.

I did a little more digging and discovered that the output of the `whatis` and `apropos` commands contained a lot of useless garbage:

```
$ whatis man
hugo-gen-man(1)          - Generate man pages for the Hugo CLI
groff_man(7)             - groff `man' macros to support generation of man pages
groffer(1)               - display groff files and man~pages on X and tty
man(1)                   - format and display the on-line manual pages
man.conf(5)              - configuration data for man
zshall(1)                - the Z shell meta-man page
xml2man(1)               - MPGL to mdoc (man page) translator
makewhatis: /usr/lib/./libgutenprint.2.dylib: No such file or directory
[...]

PR #6562, opened 2020-01-31
Avoid apropos on macOS 10.15 with man completion

geoff-nixon:
Same issue with `apropos` and `makewhatis` as described in #6270 occurs with man completions on macOS 10.15. I've just copied the workaround merged in 611a658 for `__fish_describe_command.fish` verbatim here for `__fish_complete_man.fish` (except for the function name, of course).

### DIFF
--- a/share/functions/__fish_complete_man.fish
+++ b/share/functions/__fish_complete_man.fish
@@ -1,3 +1,20 @@
+# macOS 10.15 "Catalina" has some major issues.
+# The whatis database is non-existent, so apropos tries (and fails) to create it every time,
+# which takes about half a second.
+#
+# So we disable this entirely in that case.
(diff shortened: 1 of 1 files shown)


zanchey:
We probably need to take this for 3.1.0, right?

krobelus:
It may be better to provide functions for `apropos`/`whatis`, as in https://github.com/fish-shell/fish-shell/issues/6270#issuecomment-570946361 that would also fix all other uses. If we don't do that, taking this instead seems good.

zanchey:
Merged with rebase:

   ee3f0b19b..6866f8e6b  Integration_3.1.0 -> Integration_3.1.0
   43edbf4a9..f7edfba5d  master -> master

Thanks!

Issue #5869, opened 2019-05-10
Brace expansion should skip braces with no commas


lilyball:
At some point we changed brace expansion to not trigger if the braces were empty, e.g. `foo{}` expands to `foo{}`. We should extend this to not trigger if the braces have no (unescaped) comma. This would match Bash/Zsh behavior.

In particular, I want this because of how much Git uses braces (without commas) as special syntax. For years now I've had a global variable `u` bound to the string `@{u}` so I can write e.g. `master$u` to refer to master's upstream, but that doesn't help with any of the other braced expressions, such as `@{1}`.

More generally, I can't think of any good reason to perform brace expansion when there are no (unescaped) commas, because all that does is delete the braces without adding anything.

faho:
What about `{$var}`?

Lily Ballard <notifications@github.com> schrieb am Fr., 10. Mai 2019, 20:23:

lilyball:
You mean if `$var` contains commas? That doesn't trigger brace expansion.

```fish
> set foo one,two,three
> echo bar{$foo}
barone,two,three
```

faho:
>You mean if $var contains commas? That doesn't trigger brace expansion.

It does. That *is* brace expansion. If it wasn't, the braces would still be there in the output.

What I'm getting at is this:

It's possible that we can come up with a rule that allows `HEAD@{0}`, but it would have to treat `{$var}` as it is currently, or we've got a *big* backwards-incompatibility on our hands. And the rule would have to be simple enough to explain, or it might also not be worth it.

lilyball:
Oh I see, I completely forgot that Fish uses `{$foo}` instead of `${foo}` in order to terminate the variable reference when it's immediately followed by identifier characters. That's frustrating.

We could say "if the braced text contains no commas or variable references, the braces are left intact". That's a fairly simple rule. It solves the `HEAD@{0}` case while still allowing `{$var}foo`.

0ion9:
The only case I can think of where this would be annoying is a 'fill this in later' type scenario

`for x in {foo}.png; stat -c %Y $x; end | sort -rn| head -1`

where you are expecting to fill in additional items after foo as you refine your idea.

`for x in {foo,bar,baz}.png; stat -c %Y $x; end|sort -rn | head -1`

special casing 'no commas' would mean the 'single literal value' cannot be written with the implication of being intended to become an array, as the expansion of `{foo}` is then not a subset of the expansion of `{foo,bar,baz}`

[...]

Issue #5880, opened 2019-05-19
Parser should accept `echo {foo`


lilyball:
Current `master` no longer expands braces that do not contain a comma or variable reference (see 967c1d51eec0f840273c20a326152ea72f920d3d). This means that e.g. `echo {foo}` echoes `{foo}`.

As such, we should probably allow for expressions like `echo {foo`, which should echo `{foo`, as the brace is not considered special. The expression `echo {foo,` would continue to error out, as this is an attempted parameter expansion.

According to [this comment](https://github.com/fish-shell/fish-shell/issues/5869#issuecomment-493741430) the issue with accepting `echo {foo` is that the tokenizer checks matching braces (and it references #2485).

mqudsi:
I guess it depends on whether you consider a sole opening brace to be special or not. How often do you have to type a literal `{` but not a literal `}` at any point thereafter? It seems like it would be an incredibly small number of commands that would benefit by this, to the extent that they may be overshadowed by typos where brace expansion was intended?

lilyball:
The point here is that as of 967c1d51eec0f840273c20a326152ea72f920d3d `echo {foo}` no longer treats the braces as special, and since they're not special, then `echo {foo` should also work. Braces only become special once they contain a comma or variable reference, so `echo {foo,` would become an error, but up until you type the comma the brace should be fine.

mqudsi:
Yes, I understand that. @faho's 967c1d5 was really more of special-casing an empty `{}` to work around a common user error rather than actually changing fish's syntax. My earlier reply was questioning the utility of taking that further.

lilyball:
We special-cased `{}` a while ago. 967c1d51eec0f840273c20a326152ea72f920d3d is actually a deliberate change to make braces non-special unless they contain commas or variables, largely motivated by the fact that Git uses braces for some of its syntax (e.g. `master@{upstream}` to reference the upstream branch for master, or `HEAD@{2}` to reference something in the reflog, or `v0.3.2^{commit}` to convert a tag into the pointed-to commit, etc). Also see #5869 for discussion.

Phidica:
I agree with @mqudsi that allowing a sole `{` to be treated literally would surely have little practical benefit, and could more often trip people up by *not* throwing a syntax error, because most likely they have forgotten the closing brace. 

[...]

PR #6565, opened 2020-02-03
Retain leading spaces in non-expanding braces

### LINKED ISSUE #6564: 3.1b1: The consumption of whitespace inside braces has changed weirdly since 3.0
Tried using tagged release 3.1b1 as well as current master (1101cff56) built from source on Fedora 30, running in Gnome Terminal.

I just noticed this while responding to #5880. 

In 3.0, extra whitespace inside of braces was consumed from both sides:
```fish
$ echo .{  foo  bar  }.
.foo  bar.
```

In 3.1b1, (perhaps due to the changes from #5869?) we trim whitespace but *only from the front*:
```fish
$ echo .{  foo  bar  }.
.{foo  bar  }.
```

Honestly, I'm not sure what should be happening. Do we want to preserve *everything* inside of braces that don't contain a comma/variable? Or, do we still trim spaces from the front and back like we do when there *is* a comma/variable? Consider that
```fish
$ echo .{  foo  ,  bar  }.
.foo. .bar.
```
still comes out symmetric in 3.1b1.

ridiculousfish:
This is a fix for #6564 which looks important for 3.1 release. I'd like review because it touches on how brace expansion works.

This makes two changes:

1. Remove the 'brace_text_start' idea. The idea of 'brace_text_start' was to prevent emitting `BRACE_SPACE` at the beginning or end of an item. But we later strip these off anyways, so there is no apparent benefit. If we are not doing brace expansion, this prevented emitting whitespace at the beginning or end of an item, leading to #6564.

2. When performing brace expansion, only stomp the space character with `BRACE_SPACE`; do not stomp newlines and tabs. This is because the fix in #5869 "un-converted" `BRACE_SPACE` back to space, but if the `BRACE_SPACE` came from a newline or tab literal, then we would have effectively replaced a newline or tab with a space, so this is important for #6564 as well. Moreover, it is not easy to place a literal newline or tab inside a brace expansion, and users who do probably do not mean for it to be stripped, so I believe this is a good change in general.

It also migrates some relevant '.in' tests to littlecheck.

Fixes #6564

### DIFF
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -1435,7 +1435,6 @@ static bool unescape_string_internal(const wchar_t *const input, const size_t in
     // The positions of variable expansions or brace ","s.
     // We only read braces as expanders if there's a variable expansion or "," in them.
     std::vector<size_t> vars_or_seps;
-    bool brace_text_start = false;
     int brace_count = 0;
 
     bool errored = false;
@@ -1531,7 +1530,6 @@ static bool unescape_string_internal(const wchar_t *const input, const size_t in
                         // assert(brace_count > 0 && "imbalanced brackets are a tokenizer error, we
                         // shouldn't be able to get here");
                         brace_count--;
-                        brace_text_start = brace_text_start && brace_count > 0;
                         to_append_or_none = BRACE_END;
                         if (!braces.empty()) {
                             // If we didn't have a var or separator since the last '{',
@@ -1559,17 +1557,13 @@ static bool unescape_string_internal(const wchar_t *const input, const size_t in
                 case L',': {
                     if (unescape_special && brace_count > 0) {
                         to_append_or_none = BRACE_SEP;
-                        brace_text_start = false;
                         vars_or_seps.push_back(input_position);
                     }
                     break;
                 }
-                case L'\n':
-                case L'\t':
                 case L' ': {
                     if (unescape_special && brace_count > 0) {
-                        to_append_or_none =
-                            brace_text_start ? maybe_t<wchar_t>(BRACE_SPACE) : none();
+                        to_append_or_none = BRACE_SPACE;
                     }
                     break;
                 }
@@ -1586,9 +1580,6 @@ static bool unescape_string_internal(const wchar_t *const input, const size_t in
                     break;
                 }
                 default: {
-                    if (unescape_special && brace_count > 0) {
-                        brace_text_start = true;
-                    }
                     break;
                 }
             }
--- a/tests/checks/braces.fish
+++ b/tests/checks/braces.fish
@@ -1,0 +1,50 @@
+#RUN: %fish %s
+
+echo x-{1}
+#CHECK: x-{1}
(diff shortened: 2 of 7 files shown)


mqudsi:
For the record:

**Behavior prior to this patch:**

```
> printf "\"%s\"\n" {  line\n 1\n}
"{line
 1
}"
```

**Behavior with this patch:**
```
> printf "\"%s\"\n" {  line\n 1\n}
"{  line
 1
}"
```

This patch makes the handling of both leading and trailing spaces consistent with one another, but the bigger question regarding tokenization isn't addressed.

@ridiculousfish I don't know if you saw [my comment on 6564](https://github.com/fish-shell/fish-shell/issues/6564#issuecomment-581185676)?

Braces are now equivalent to a new form of quoting, except the quote is preserved and emitted in the output. Unless you try to use a variable within them. Or go back and try to add a comma. *This is weird and complicated*.

e.g. consider @lilyball's proposal in #5880 (which I am against because it complicates things even further), how are you going to tokenize `{  hello` (that's two spaces between the `{` and the `h`)? 

If it's really so horrible that braces are special characters then get rid of brace expansion altogether and force people to use `(printf "foo" "bar")` everywhere they were using `{foo, bar}` for cartesian expansion because as hard as that would be, it's still easier than trying to understand this new model.

ridiculousfish:
Ok, fair point. We need to decide whether e.g.:

    count .{  foo  bar  }.

should be 1 or 4.

krobelus:
This PR looks good, it only improves consistency.
> ```
> count .{  foo  bar  }.
> ```
It's probably better to make this 4, for the reasons @mqudsi has given.
IMO after 3.1.0 is fine for that - I'd be surprised if someone relied on the current quoting behavior of `{ foo }`..

ridiculousfish:
I will evaluate how difficult / risky the tokenization fix would be 3.1. It would be unfortunate if we had to churn this again for 3.2

ridiculousfish:
I think the tokenization fix would be too difficult. Somewhere inside our tokenizer is a real recursive descent parser struggling to get out and that's how this can be fixed.

I feel better about the churn because unbalanced brackets are still an error:

```
echo abc{def
fish: Unexpected end of string, incomplete parameter expansion
```

so if we allow this in 3.2 it cannot break anyone. I'm going to merge this.

ridiculousfish:
bd06a9aa6ccdbf72e5d36c3c5c3d2a830c186541 in master, eaecb817ca9cffd5bdf91710d0363efb9d5d4791 in Integration_3.1.0

mqudsi:
I'm fine with this PR shipping in 3.1 and the functionality being corrected in 3.2, but disagree that it can't break anything as code can be written in 3.1 relying on the new tokenization behavior (e.g. `set json { \"hello\"\, \"world\" }`) that will no longer be valid (or at least not retain the same behavior) in 3.2; but that's being extremely pedantic.

PR #6566, opened 2020-02-03
prompts: fix pipestatus for jobs prefixed with "not"

krobelus:
6902459566b491e2debb136f43f815965910eeee was an attempt to not print
$status twice in the prompt. As a result we print $pipestatus but
not $status, which *usually* is the same as $pipestatus[-1] --- unless
the builtin "not" is used, which inverts the $status of a job (it does
not alter $pipestatus).

As a result, the default prompt prints unexpected status codes:

	~ > not false
	~ [1]> not true
	~ > not true | true
	~ > not false | false
	~ [1|1]>

This commit reintroduces printing of $status after $pipestatus, but only
if it is different from $pipestatus[-1].
Additionally, we only print anything at all if the $status is nonzero,
to avoid confusing output on `not false | false`

	~ > not false
	~ > not true
	~ [0] 1> not true | true
	~ [0|0] 1> not false | false
	~ >

I think this is closer to users' expectations for those cases; they should
not have to think about this implementation detail of the not-statement.

### DIFF
--- a/share/functions/__fish_print_pipestatus.fish
+++ b/share/functions/__fish_print_pipestatus.fish
@@ -1,4 +1,11 @@
 function __fish_print_pipestatus --description "Print pipestatus for prompt"
+    # take $status as optional argument to maintain compatibility
+    set -l last_status
+    if set last_status (string match -r -- '^\d+$' $argv[1])
+        set -e argv[1]
+    else
+        set last_status $argv[-1] # default to $pipestatus[-1]
+    end
     set -l left_brace $argv[1]
     set -l right_brace $argv[2]
     set -l separator $argv[3]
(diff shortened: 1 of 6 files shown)


share/functions/__fish_print_pipestatus.fish
```diff
@@ -1,18 +1,28 @@
...
     set -l separator $argv[3]
     set -l brace_sep_color $argv[4]
     set -l status_color $argv[5]
     set -e argv[1 2 3 4 5]
 
-    # only output status codes if some process in the pipe failed
-    # SIGPIPE (141 = 128 + 13) is usually not a failure, see #6375.
-    if string match -qvr '^(0|141)$' $argv
+    # Only print status codes if the job failed.
+    if test $last_status -ne 0
```
ridiculousfish: Probably best to quote this due to test's parsing quirks. `"$last_status"`

mqudsi:
I agree with the idea, but I think it needs to be a bit more intuitive.
e.g. it is not clear why the number of entries in pipestatus differs for these two:

```
~ [0] 1> not true | true
~ [0|0] 1> not false | false
```

krobelus:
> I agree with the idea, but I think it needs to be a bit more intuitive.
> e.g. it is not clear why the number of entries in pipestatus differs for these two:

`[0] 1>` is the prompt for the previous command, let me format it in a
more obvious to way:

    ~ > not false
    ~ >

    ~ > not true
    ~ [0] 1>

    ~ > not true | true
    ~ [0|0] 1>

    ~ > not false | false
    ~ >

For the record, an alternative I was considering is to replace `$pipestatus[-1]` with `$status`, like so:

    ~ > not true
    ~ [1]>

    ~ > not true | true
    ~ [0|1]>

Which would avoid the trailing status when using `not`, however it is probably not as intuitive when using pipes.

PR #5897, opened 2019-05-27
WIP Add undo

faho:
We add two "undo_list_t", one for undo and one for redo.

If we do anything that changes the commandline, we add an undo
element, if we undo, we add a redo element.

Also we coalesce small changes like normal string insertions and
single-character deletions.

This is only lightly tested and might not work for everything. In particular I'm not sure if I caught everything or how it works for weirder bindings.

I'd just like to know that the general idea is solid.

Fixes #1367.

## TODOs:

- [ ] Changes to fish usage are reflected in user documentation/manpages.
- [ ] Tests have been added for regressions fixed
- [ ] User-visible changes noted in CHANGELOG.md

PR #6570, opened 2020-02-06
Add undo

### LINKED ISSUE #1367: Add support for `undo`
It would be nice to have an `undo` feature.

I don't know about bash, but zsh and emacs use <ctrl + _> for this and it is very very useful.

krobelus:
Add the readline function `undo` which is bound to `\c_` (control + / on
some terminals). Redoing the most recent chain of undos is supported,
`redo` is bound to `\e/` for now.

Closes #1367.
This approach should not have the issues discussed in #5897.

Every single modification to the commandline can be undone individually,
except for adjacent single-character inserts, which are coalesced,
so they can be reverted with a single undo. Coalescing is not done for
space characters, so each word can be undone separately.

When moving between history search entries, only the current history
search entry is reachable via the undo history. This allows to go back
to the original search string with a single undo, or by pressing the
escape key.
Similarly, when moving between pager entries, only the most recent
selection in the pager can be undone.

TODO: (I'm not sure about the first three)

- [ ] decide on the redo hotkey; `\e/` does not seem to work on xterm
      (it is tempting to add [emacs-style redo] but that would probably be less popular)
- [ ] vi mode: find some suitable bindings (`u` and `\cr` are taken)
- [ ] vi mode: find out how to bind in insert mode (which should then also work in the pager's search field)
- [x] update documentation

These areas should be mostly alright based on what I tried but there may be room for improvement:

- make sure the undo grouping behavior works nicely in practise
- make sure the interaction with history search feels right
- make sure the interaction with the pager feels right
- add some tests (interactive tests are missing - expect-based tests are not very nice to write last time I tried)

I think this also allows us to get rid of `cycle_command_line` and `cycle_cursor_pos` because we can obtain those from the history.

[emacs-style redo]: <https://www.gnu.org/software/emacs/manual/html_node/emacs/Undo.html>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Interactive improvements
 
 #### New or improved bindings
+- New readline commands `undo` (Ctrl+_) and `redo` (Alt-/) can be used to revert changes to the command line or the pager search field (#6570).
 
 #### Improved prompts
 - The default and example prompts print the correct exit status for commands prefixed with `not` (#6566).
--- a/share/functions/fish_default_key_bindings.fish
+++ b/share/functions/fish_default_key_bindings.fish
@@ -69,6 +69,8 @@ function fish_default_key_bindings -d "Default (Emacs-like) key bindings for fis
     bind --preset $argv \cf forward-char
     bind --preset $argv \cb backward-char
     bind --preset $argv \ct transpose-chars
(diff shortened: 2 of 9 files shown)


krobelus:
Thanks!
Merging to master; the current behavior feels reasonable for a start.
vi-mode keybindings and a more portable redo hotkey should be added eventually

Issue #4456, opened 2017-10-08
Flatpak completion includes extra text


TingPing:
Fish version: 2.6.0
Flatpak version: 0.9.12

Flatpak completion of applications is currently broken probably as the format of Flatpaks cli output has changed:

https://asciinema.org/a/8dCLh0bJNwqRiWfVvds2C1hIx

mqudsi:
Can flatpack URLs contain spaces?

TingPing:
An app-id cannot contain spaces.

mqudsi:
Can you try the attached diff? You can just apply it to `/usr/local/share/fish/completions/flatpak.fish`:

```diff
diff --git a/share/completions/flatpak.fish b/share/completions/flatpak.fish
index 5860c6b..565ab0b 100644
--- a/share/completions/flatpak.fish
+++ b/share/completions/flatpak.fish
@@ -30,9 +30,9 @@ complete -f -c flatpak -n "not __fish_seen_subcommand_from $commands" -a build-i
 complete -f -c flatpak -n "not __fish_seen_subcommand_from $commands" -a build-sign          -d 'Sign an application or runtime'
 complete -f -c flatpak -n "not __fish_seen_subcommand_from $commands" -a build-update-repo   -d 'Update the summary file in a repository'

-complete -f -c flatpak -n "__fish_seen_subcommand_from run" -a "(flatpak list --app | string trim)"
-complete -f -c flatpak -n "__fish_seen_subcommand_from info uninstall" -a "(flatpak list | string trim)"
-complete -f -c flatpak -n "__fish_seen_subcommand_from remote-ls remote-modify remote-delete" -a "(flatpak remote-list | string trim)"
+complete -f -c flatpak -n "__fish_seen_subcommand_from run" -a "(flatpak list --app | string trim | string match -r '^\S+')"
+complete -f -c flatpak -n "__fish_seen_subcommand_from info uninstall" -a "(flatpak list | string trim | string match -r '^\S+')"
+complete -f -c flatpak -n "__fish_seen_subcommand_from remote-ls remote-modify remote-delete" -a "(flatpak remote-list | string trim | string match -r '^\S+')"

 # Note that "remote-ls" opens an internet connection, so we don't want to complete "install"
 # Plenty of the other stuff is too free-form to complete (e.g. remote-add).
```

TingPing:
That works as expected.

mqudsi:
Thanks, @TingPing 

This is fixed for fish3

[...]

PR #6581, opened 2020-02-10
Fix flatpak completion

brunelli:
The `flatpak` completion is broken since the implementation of columns in v1.2. See [Flatpak commandline design](https://blogs.gnome.org/mclasen/2018/12/19/flatpak-commandline-design/).

This is a minimal change to make it work exactly like it did before.

### DIFF
--- a/share/completions/flatpak.fish
+++ b/share/completions/flatpak.fish
@@ -1,27 +1,50 @@
 # Completions for flatpak, an "Application deployment framework for desktop apps"
 # (http://flatpak.org)
-set -l flatversion (flatpak --version | string match -r '[\d.]+' | string split .)
-set -l commands install update uninstall list info run override make-current enter document-{export,unexport,info,list} \
-    remote-{add,modify,delete,list,ls} build build-{init,finish,export,bundle,import-bundle,sign,update-repo}
+set -l flatversion (flatpak --version | string match -r '[\d.]+' | string split . | head -n 2 | string join .)
+set -l commands install update uninstall list info config repair create-usb search run override make-current enter ps \
+    documents document-{export,unexport,info} permissions permission-{show,reset} remotes remote-{add,modify,delete,ls,info} \
+    build build-{init,finish,export,bundle,import-bundle,sign,update-repo,commit-from} repo
+
+if test $flatversion -ge 1.1
+    set commands $commands history kill
+    complete -f -c flatpak -n "not __fish_seen_subcommand_from $commands" -a history -d 'Show history'
+    complete -f -c flatpak -n "not __fish_seen_subcommand_from $commands" -a kill -d 'Stop a running application'
+end
+
+if test $flatversion -ge 1.5
+    set commands $commands mask permission-{remove,set}
+    complete -f -c flatpak -n "not __fish_seen_subcommand_from $commands" -a mask -d 'Mask out updates and automatic installation'
+    complete -f -c flatpak -n "not __fish_seen_subcommand_from $commands" -a permission-remove -d 'Remove item from permission store'
+    complete -f -c flatpak -n "not __fish_seen_subcommand_from $commands" -a permission-set -d 'Set permissions'
+end
 
 complete -f -c flatpak -n "not __fish_seen_subcommand_from $commands" -a install -d 'Install an application or runtime from a remote'
 complete -f -c flatpak -n "not __fish_seen_subcommand_from $commands" -a update -d 'Update an installed application or runtime'
 complete -f -c flatpak -n "not __fish_seen_subcommand_from $commands" -a uninstall -d 'Uninstall an installed application or runtime'
 complete -f -c flatpak -n "not __fish_seen_subcommand_from $commands" -a list -d 'List installed apps and/or runtimes'
 complete -f -c flatpak -n "not __fish_seen_subcommand_from $commands" -a info -d 'Show info for installed app or runtime'
+complete -f -c flatpak -n "not __fish_seen_subcommand_from $commands" -a config -d 'Configure flatpak'
+complete -f -c flatpak -n "not __fish_seen_subcommand_from $commands" -a repair -d 'Repair flatpak installation'
+complete -f -c flatpak -n "not __fish_seen_subcommand_from $commands" -a create-usb -d 'Put applications or runtimes onto removable media'
+complete -f -c flatpak -n "not __fish_seen_subcommand_from $commands" -a search -d 'Search for remote apps/runtimes'
 complete -f -c flatpak -n "not __fish_seen_subcommand_from $commands" -a run -d 'Run an application'
 complete -f -c flatpak -n "not __fish_seen_subcommand_from $commands" -a override -d 'Override permissions for an application'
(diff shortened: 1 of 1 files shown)


faho:
And if you have an older flatpak, like e.g. CentOS 8's 1.0.6, how will this react?

brunelli:
It'll probably exit with an error, as `--columns` wasn't implemented before 1.2 (release more than a year ago).

Should I add a version check to make sure it won't break with flatpak<1.2?

Also, should we fix this behavior, could this target 3.1/3.1.1? This really brings down the flatpak experience with fish.

faho:
>Should I add a version check to make sure it won't break with flatpak<1.2?

Ideally yes.

You'll want something like

```fish
set -l flatversion (flatpak --version | string match -r '[\d.]+' | string split .)
if test "$flatversion[1]" -gt 1 -o "$flatversion[2]" -gt 2
# do the new thing
else
# do the old thing
end
```

(if it's okay to ignore versions < 1)

>Also, should we fix this behavior, could this target 3.1/3.1.1? This really brings down the flatpak experience with fish.

If I'm confident it works, I'll cherry-pick it to 3.1 if possible.

brunelli:
Okay, it's now working as we've discussed!

I tried to do something so we could have descriptions showing the app name next to its ID. This uses a for loop. Can this be improved somehow?

The other thing I want to do is to add the new commands to completions. They are (as per v1.6.1):

```
mask                   Mask out updates and automatic installation
history                Show history
config                 Configure flatpak
repair                 Repair flatpak installation
create-usb             Put applications or runtimes onto removable media
search                 Search for remote apps/runtimes

ps                     Enumerate running applications
kill                   Stop a running application

documents              List exported files

permissions            List permissions
permission-remove      Remove item from permission store
permission-set         Set permissions
permission-show        Show app permissions
permission-reset       Reset app permissions

remotes                List all configured remotes
remote-info            Show information about a remote app or runtime

build-commit-from      Create new commit based on existing ref
repo                   Show information about a repo
```

The thing is: some were added in 1.2, some were added after that. What would be a clean way to set commands per version?

brunelli:
Okay, some commands inside the `$flatversion -ge 1.0` check were added before `1.0`, but I think it's cleaner to make the cut there.

If you believe it's safe to assume there're no one shipping `flatpak<1.0`, maybe we should skip this check altogether.

share/completions/flatpak.fish
```diff
@@ -30,9 +31,21 @@ complete -f -c flatpak -n "not __fish_seen_subcommand_from $commands" -a build-i
 complete -f -c flatpak -n "not __fish_seen_subcommand_from $commands" -a build-sign -d 'Sign an application or runtime'
 complete -f -c flatpak -n "not __fish_seen_subcommand_from $commands" -a build-update-repo -d 'Update the summary file in a repository'
 
-complete -f -c flatpak -n "__fish_seen_subcommand_from run" -a "(flatpak list --app | string match -r '\S+')"
-complete -f -c flatpak -n "__fish_seen_subcommand_from info uninstall" -a "(flatpak list | string match -r '\S+')"
-complete -f -c flatpak -n "__fish_seen_subcommand_from remote-ls remote-modify remote-delete" -a "(flatpak remote-list | string match -r '\S+')"
+if test (string join . "$flatversion[1]" "$flatversion[2]") -ge 1.2
+    for line in (flatpak list --app --columns=application,name)
+        set -l info (string split \t "$line")
+        complete -f -c flatpak -n "__fish_seen_subcommand_from run" -a "$info[1]" -d "$info[2]"
```
faho: This `string split` is unnecessary, because bits after a tab are already used as a description for the line.

So, in total, this entire for-loop can just be:

```fish
complete -f -c flatpak -n "__fish_seen_subcommand_from run" -a '(flatpak list --app --columns=application,name)'
```

share/completions/flatpak.fish
```diff
@@ -30,9 +31,21 @@ complete -f -c flatpak -n "not __fish_seen_subcommand_from $commands" -a build-i
...
+    end
+    for line in (flatpak list --columns=application,name)
+        set -l info (string split \t "$line")
+        complete -f -c flatpak -n "__fish_seen_subcommand_from info uninstall" -a "$info[1]" -d "$info[2]"
```
faho: Same as above.

share/completions/flatpak.fish
```diff
@@ -30,9 +31,21 @@ complete -f -c flatpak -n "not __fish_seen_subcommand_from $commands" -a build-i
...
+    end
+    complete -f -c flatpak -n "__fish_seen_subcommand_from remote-ls remote-modify remote-delete" -a "(flatpak remote-list --columns=name)"
+else
+    complete -f -c flatpak -n "__fish_seen_subcommand_from run" -a "(flatpak list --app | string match -r '\S+')"
```
faho: If you want to trim leading and/or trailing whitespace `string trim` is a bit more explicit.

share/completions/flatpak.fish
```diff
@@ -30,9 +31,21 @@ complete -f -c flatpak -n "not __fish_seen_subcommand_from $commands" -a build-i
 complete -f -c flatpak -n "not __fish_seen_subcommand_from $commands" -a build-sign -d 'Sign an application or runtime'
 complete -f -c flatpak -n "not __fish_seen_subcommand_from $commands" -a build-update-repo -d 'Update the summary file in a repository'
 
-complete -f -c flatpak -n "__fish_seen_subcommand_from run" -a "(flatpak list --app | string match -r '\S+')"
-complete -f -c flatpak -n "__fish_seen_subcommand_from info uninstall" -a "(flatpak list | string match -r '\S+')"
-complete -f -c flatpak -n "__fish_seen_subcommand_from remote-ls remote-modify remote-delete" -a "(flatpak remote-list | string match -r '\S+')"
+if test (string join . "$flatversion[1]" "$flatversion[2]") -ge 1.2
```
faho: This works because we force the c locale for float comparisons, otherwise you'd have issues with locales that use "," as a decimal separator like e.g. german.

But it does work, and we're not going to change this, so it's okay. It's just an idea to keep in mind, because other programs might not be so accomodating.

I would however ignore stderr here - `2>/dev/null`, in case we didn't get a version.

share/completions/flatpak.fish
brunelli: Is this related to the `string match -r '\S+'` bit?

It was like this in the original code and I left it untouched because I'm not sure of how was the `flatpak list` output before columns (and I assumed there was a good reason for using `string match`).

faho: Ah, okay. I missed that it was there before. It's fine to leave it like that, then!

share/completions/flatpak.fish
brunelli: > I would however ignore stderr here - `2>/dev/null`, in case we didn't get a version.

This will make completions fallback to v1.0. Sounds good.

brunelli:
Just pushed a commit with all the requested changes

share/completions/flatpak.fish
brunelli: After further investigation, it used `string trim` before (98b561929fa01d6f264d4fa0accd14790134e2c0) but it was changed due to #4456.

PR #6599, opened 2020-02-13
Restore `time` as a generic builtin

### LINKED ISSUE #6598: New `time` builtin cannot be detected as a builtin
Hi 👋 

I've just upgraded to 3.1.0 and have just learned about the new `time` builtin. I'm running into a small issue, though, where the `builtin` command doesn't seem to be aware that `time` is now a builtin.

```
$ builtin -n | grep time; echo $status
$ 1
$ builtin -q time; echo $status
$ 1
```

~As far as I can tell, this is because it was removed from `src/builtin.cpp` in 664d6fb13298976f35ae9278ebb7a08e9beb6feb. I'm not very familiar with this project, but would you accept a PR adding it back as a `builtin_generic`?~

Edit: It's a little bit more work than I thought, see #6599.

charlesbjohnson:
## Description

Resolves #6598.

### DIFF
--- a/src/builtin.cpp
+++ b/src/builtin.cpp
@@ -394,6 +394,7 @@ static const builtin_data_t builtin_datas[] = {
     {L"string", &builtin_string, N_(L"Manipulate strings")},
     {L"switch", &builtin_generic, N_(L"Conditionally execute a block of commands")},
     {L"test", &builtin_test, N_(L"Test a condition")},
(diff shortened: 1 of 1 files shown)


zanchey:
This appears to break the use of the command, as the tests are failing on all platforms; does `make test` pass locally for you?

charlesbjohnson:
D'oh, it doesn't pass locally for me either. I guess the fix isn't this simple so I'll just close this for now.

PR #6602, opened 2020-02-14
Work around compilation failure with old Apple SDKs

lilyball:
When building fish-shell with the macOS 10.12 SDK, `<sys/proc.h>` does not include `<sys/time.h>` but references `struct itimerval`. This causes a compilation failure if we don't import `<sys/time.h>` ourselves.

This was previously masked by an import of `<sys/sysctl.h>`, which was removed in fc0c39b6fdb6.

This fixes a build failure when building Fish using Nix, as Nix is currently on the macOS 10.12 SDK.

### DIFF
--- a/src/expand.cpp
+++ b/src/expand.cpp
@@ -17,6 +17,7 @@
 #include <procfs.h>
 #endif
 #if __APPLE__
+#include <sys/time.h> // Required to build with old SDK versions
 #include <sys/proc.h>
 #else
 #include <dirent.h>


src/expand.cpp
```diff
@@ -17,6 +17,7 @@
 #include <procfs.h>
 #endif
 #if __APPLE__
+#include <sys/time.h> // Required to build with old SDK versions
```
zanchey: possibly make this an IWYU pragma?

floam:
Could hide it behind a check specifically for the 10.12 SDK or older before including it. Not really going to make a difference though.

src/expand.cpp
lilyball: I'm not set up to run IWYU so I don't know if it would even flag this. We do use something from this file, it's just that in newer SDKs it's redundant because `<sys/proc.h>` includes it.

lilyball:
@floam I don't know which SDK added the `#include <sys/time.h>` to `<sys/proc.h>` so I don't feel comfortable making it conditional.

zanchey:
SGTM! I think we're definitely going to have to do a 3.1.1, so I'll add this to the milestone and cherry-pick it when I get a chance.

faho:
I've just merged this now, as-is, because it fixes what it purports properly. I think we'd have to re-check a bunch of includes anyway if we reran IWYU (I think the last time was in 2017?).

PR #6618, opened 2020-02-17
Docs: grammar nit in function page

ammgws:
Per title


faho:
Yup, that sure is a grammer.

zanchey:
Ha. I wrote that, reads ok either way to me.

Issue #6617, opened 2020-02-17
rbenv completion spews errors on load


faho:
From the "how-did-nobody-report-this" department:

Try to complete `rbenv`, and you get an error about there being too many arguments to `complete` in this line:

```fish
complete -f -c rbenv -n '__fish_rbenv_needs_command' -a version-file 'Detect the file that sets the current rbenv version
```

Of course the "-d" before the description is missing.

Is anybody using these completions?

Opening an issue mainly to tag it as 3.1.1 to remind us to cherry-pick it.

faho:
Ah, wait, added in 5c8ef19ff8b0eecfa86dd79c6143984ec858227f, which was merged after 3.1.0.

faho:
The optipng completions spew too. Sorry!

PR #6619, opened 2020-02-17
Check all fish files

faho:
## Description

This adds a check to run all fish files (except tests/) through the recently fixed `fish --no-execute` in order to find syntax errors.

It would have caught 63b4a891ff3559a14b13124ea3bbca19fae98b9d (but not #6617).

This is a nice idea, but it comes at a cost of ~4s per test run (on my machine).

So the question is: Are we willing to spend that?

Also there are a bunch of things that we could skip when we do --no-execute, like initializing the environment. Note that currently feature flags don't matter for no-execute because they don't change any syntax *errors* - nothing starts to be syntactically valid or stops being valid.

## TODOs:

- [N/A] User-visible changes noted in CHANGELOG.md

### DIFF
--- a/tests/checks/check-all-fish-files.fish
+++ b/tests/checks/check-all-fish-files.fish
@@ -1,0 +1,8 @@
+#RUN: %fish -C 'set -l fish %fish' %s
+# Test ALL THE FISH FILES
+# in share/, that is - the tests are exempt because they contain syntax errors, on purpose
+
+for file in $__fish_data_dir/**.fish
+    $fish -n $file
+end
+# No output is good output


zanchey:
One possibility is to hide this behind an environment variable, and only set it in one of the Travis builds.

mqudsi:
This is extremely parallelizable; that time could easily come down a good chunk if we run them in parallel. But a better solution might be to "remember" the last-tested mtime in the previous run and resume from there. A clean build would always test everything:

```fish
#RUN: %fish -C 'set -l fish %fish' %s
# Test ALL THE FISH FILES
# in share/, that is - the tests are exempt because they contain syntax errors, on purpose

set -l mnewer (test -f .last_noexec_check && printf -- "-mnewer\n.last_noexec_check\n")
for file in (find $__fish_data_dir -name "*.fish" $mnewer)
    $fish -n $file
end
touch .mnewer
```

If you need to speed up the travis tests too, you can use this instead to test only files that have been changed in the last _n_ commits (unless travis does a `git clone --depth 1`, in which case I think it'll be only "changed in the last commit?"):

```fish
#RUN: %fish -C 'set -l fish %fish' %s
# Test ALL THE FISH FILES
# in share/, that is - the tests are exempt because they contain syntax errors, on purpose

for file in (git diff --name-only HEAD~100)
    $fish -n $file
end
```

The only time the caching would be a problem is if we changed the syntax parsing and previously valid syntax is now invalid.

ridiculousfish:
Seems odd that it would be so slow, I suspect this can be improved.

ridiculousfish:
It's spending all its time in parsing, bleh. Optimizing debug builds is hard. Anyways LGTM, it's a good test. We can work to reduce the test cycle time in the future.

faho:
Merged as 22c0054c1 (I rebased manually to run the tests once more, so github didn't detect it)

PR #6627, opened 2020-02-19
Correction link in doc.

Djabx:
## Description

Correction of a link in documentation.


faho:
Yup, that sure is a broken link.

Thanks, nice catch!

PR #6633, opened 2020-02-21
clang-tidy fixups

neheb:
## Description

clang-tidy fixups.

### DIFF
--- a/src/complete.cpp
+++ b/src/complete.cpp
@@ -216,7 +216,7 @@ completion_t::completion_t(wcstring comp, wcstring desc, string_fuzzy_match_t ma
 completion_t::completion_t(const completion_t &) = default;
 completion_t::completion_t(completion_t &&) = default;
 completion_t &completion_t::operator=(const completion_t &) = default;
-completion_t &completion_t::operator=(completion_t &&) = default;
+completion_t &completion_t::operator=(completion_t &&) noexcept = default;
 completion_t::~completion_t() = default;
 
 __attribute__((always_inline)) static inline bool natural_compare_completions(
@@ -404,7 +404,7 @@ static owning_lock<autoload_t> completion_autoloader{autoload_t(L"fish_complete_
 
 /// Create a new completion entry.
 void append_completion(completion_list_t *completions, wcstring comp, wcstring desc,
-                       complete_flags_t flags, string_fuzzy_match_t &&match) {
+                       complete_flags_t flags, string_fuzzy_match_t match) {
     completions->emplace_back(std::move(comp), std::move(desc), match, flags);
 }
 
@@ -1209,8 +1209,7 @@ bool completer_t::complete_variable(const wcstring &str, size_t start_offset) {
         }
 
         // Append matching environment variables
-        append_completion(&this->completions, std::move(comp), std::move(desc), flags,
-                          std::move(match));
+        append_completion(&this->completions, std::move(comp), std::move(desc), flags, match);
 
         res = true;
     }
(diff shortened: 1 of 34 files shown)


src/iothread.cpp
```diff
@@ -199,7 +199,9 @@ static void enqueue_thread_result(work_request_t req) {
     s_result_queue.acquire()->push(std::move(req));
 }
 
-static void *this_thread() { return (void *)(intptr_t)pthread_self(); }
+static void *this_thread() {
+    return reinterpret_cast<void *>(static_cast<intptr_t>(pthread_self()));
```
faho: Breaks the build on macOS:

>/Users/travis/build/fish-shell/fish-shell/src/iothread.cpp:203:37: error: static_cast from 'pthread_t _Nonnull' (aka '_opaque_pthread_t *') to 'intptr_t' (aka 'long') is not allowed

>return reinterpret_cast<void *>(static_cast<intptr_t>(pthread_self()));

src/complete.cpp
```diff
@@ -214,7 +214,7 @@ completion_t::completion_t(wcstring comp, wcstring desc, string_fuzzy_match_t ma
       flags(resolve_auto_space(completion, flags_val)) {}
 
 completion_t::completion_t(const completion_t &) = default;
-completion_t::completion_t(completion_t &&) = default;
+completion_t::completion_t(completion_t &&) noexcept = default;
```
faho: Breaks the build on Travis on Linux:

>The noexcept thing breaks the build on Travis, and it's useless as we build without exceptions anyway:

>/home/travis/build/fish-shell/fish-shell/src/complete.cpp:219:15: error: function ‘completion_t& completion_t::operator=(completion_t&&)’ defaulted on its redeclaration with an exception-specification that differs from the implicit exception-specification ‘’

neheb: Fixed. Probably older GCC.

src/iothread.cpp
neheb: Switched back to C cast for that since it's static on some platforms and reinterpret_cast on others.

neheb:
Fixed.

src/screen.cpp
```diff
@@ -1153,7 +1153,7 @@ void s_reset(screen_t *s, screen_reset_mode_t mode) {
         if (screen_width > non_space_width) {
             bool justgrey = true;
             if (cur_term && enter_dim_mode) {
-                std::string dim = tparm((char *)enter_dim_mode);
+                std::string dim = tparm(enter_dim_mode);
```
faho: This cast is necessary on some platforms. Specifically I'm pretty sure this breaks the build on NetBSD.

src/screen.cpp
```diff
@@ -1164,22 +1164,22 @@ void s_reset(screen_t *s, screen_reset_mode_t mode) {
             if (cur_term && justgrey && set_a_foreground) {
                 if (max_colors >= 238) {
                     // draw the string in a particular grey
-                    abandon_line_string.append(str2wcstring(tparm((char *)set_a_foreground, 237)));
+                    abandon_line_string.append(str2wcstring(tparm(set_a_foreground, 237)));
```
faho: These casts.

floam:
I don't like the "auto all the things" commit.

faho:
Yeah, I'm gonna merge this partially. I'm dropping:

- "use auto"
- "apply performance changes" because it adds the "noexcept" to the completion_t constructor that breaks the build
- "replace C casting with C++", because it breaks tparm and looks awful
- "mark move constructors noexcept" because we do not use exceptions

To be honest, I think we've reached the limits of what these static tools can help us with. Most changes they now propose either break something or don't improve things (just c++-ification for the sake of it).

So: Thank you, that'll be enough for now!

Merged as b26637042..a410dcb5c

neheb:
Hardly. This is only version 9 of LLVM. GCC's also working on its own static analyzer.

PR #6636, opened 2020-02-22
Fix variable leaking in completion `kill`

wodesuck:
## Description

If we type `kill <tab>`, then global variables `number` and `name` will be set accidentally. This PR fix it.


faho:
Yup, that's a thing.

(In general these completions are a bit messy and would benefit from cleanup - these things have quite some variation across implementations tho, so it needs some care)

PR #6638, opened 2020-02-22
lscpu.fish: disable file completions, localise var

ammgws:
[skip ci]


ammgws:
@faho What's the recommended way to make this complete all options after typing `lscpu TAB`? Right now it only completes them if you manually type in the starting hyphen. Saw stuff like [this](https://github.com/fish-shell/fish-shell/blob/ff29d815328f737a1034ab1cf37b3b7d1d90f037/share/completions/pacman.fish#L23) in the codebase but the comment puts me off from using it.

faho:
>What's the recommended way to make this complete all options after typing lscpu TAB?

The pacman way is about the best you can do right now. I'm not sure what a good design to allow this would look like? An additional flag to `complete` to specify "hey, these arguments are not normal options, offer them even without the -"? Because the default of hiding "-"-prefixed things is absolutely the right one.

Anyway, I wouldn't sweat it for this. People get the bell, figure they can't add files or normal arguments, and type "-". It's not too bad.

PR #6639, opened 2020-02-22
cleanup completions: localise vars rather than erasing

ammgws:
[skip ci]

### DIFF
--- a/share/completions/cargo.fish
+++ b/share/completions/cargo.fish
@@ -7,10 +7,10 @@ complete -c cargo -l list -d 'List installed commands'
 complete -c cargo -s v -l verbose -d 'Use verbose output'
 complete -c cargo -s q -l quiet -d 'No output printed to stdout'
 
-set __fish_cargo_subcommands (cargo --list 2>&1 | string replace -rf '^\s+([^\s]+)\s+(.*)' '$1\t$2')
+set -l __fish_cargo_subcommands (cargo --list 2>&1 | string replace -rf '^\s+([^\s]+)\s+(.*)' '$1\t$2')
 
-complete -c cargo -f -c cargo -n '__fish_use_subcommand' -a '$__fish_cargo_subcommands'
-complete -c cargo -x -c cargo -n '__fish_seen_subcommand_from help' -a '$__fish_cargo_subcommands'
+complete -c cargo -f -c cargo -n '__fish_use_subcommand' -a "$__fish_cargo_subcommands"
+complete -c cargo -x -c cargo -n '__fish_seen_subcommand_from help' -a "$__fish_cargo_subcommands"
 
 for x in bench build clean doc fetch generate-lockfile \
     locate-project package pkgid publish \
@@ -88,4 +88,3 @@ complete -c cargo -x -n '__fish_seen_subcommand_from update' -l precise -d 'Upda
 
 complete -c cargo -x -n '__fish_seen_subcommand_from yank' -l vers -d 'The version to yank or un-yank'
 complete -c cargo -n '__fish_seen_subcommand_from yank' -l undo -d 'Undo a yank, putting a version back into the index'
-
--- a/share/completions/rustup.fish
+++ b/share/completions/rustup.fish
@@ -194,6 +194,7 @@ function __rustup_strip_common_suffix_strict
 end
 
 function __rustup_all_toolchains
+    set -l __rustup_channels "beta" "stable" "nightly"
     printf "%s\n" $__rustup_channels
     printf "%s\n" $__rustup_channels-(__rustup_triples)
 end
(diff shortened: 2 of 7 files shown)


share/completions/cargo.fish
```diff
@@ -89,3 +89,4 @@ complete -c cargo -x -n '__fish_seen_subcommand_from update' -l precise -d 'Upda
 complete -c cargo -x -n '__fish_seen_subcommand_from yank' -l vers -d 'The version to yank or un-yank'
 complete -c cargo -n '__fish_seen_subcommand_from yank' -l undo -d 'Undo a yank, putting a version back into the index'
 
+set -e __fish_cargo_subcommands
```
faho: Please just set this as local instead of erasing it explicitly. That way it will shadow any (theoretical) global of that name.

share/completions/rustup.fish
```diff
@@ -296,3 +296,6 @@ end
 function __rustup_doc_path
     rustup doc --path | string replace --regex '/[^/]*$' ''
 end
+
+set -e __rustup_channels
```
faho: These should just be local-exported (`-lx`). That way functions that need them get a copy.

share/completions/rustup.fish
```diff
@@ -224,9 +224,8 @@ set -l __rustup_toolchains (rustup toolchain list | string replace -r "\s+.*" ''
 set -l __rustup_toolchains_short (__rustup_strip_common_suffix_strict $__rustup_toolchains)
 
 # Needs to be global to retain access from functions
```
faho: This comment would of course also need to be changed - "needs to be accessible to functions, hence the x" or something?

ammgws: Removed it since it should be self-explanatory

share/completions/cargo.fish
```diff
@@ -7,7 +7,7 @@ complete -c cargo -l list -d 'List installed commands'
 complete -c cargo -s v -l verbose -d 'Use verbose output'
 complete -c cargo -s q -l quiet -d 'No output printed to stdout'
 
-set __fish_cargo_subcommands (cargo --list 2>&1 | string replace -rf '^\s+([^\s]+)\s+(.*)' '$1\t$2')
+set -l __fish_cargo_subcommands (cargo --list 2>&1 | string replace -rf '^\s+([^\s]+)\s+(.*)' '$1\t$2')
 
 complete -c cargo -f -c cargo -n '__fish_use_subcommand' -a '$__fish_cargo_subcommands'
```
faho: Okay, these have to be double-quotes now - 

```fish
complete -c cargo -f -c cargo -n '__fish_use_subcommand' -a "$__fish_cargo_subcommands"
```

before it would have worked because the variable was still defined when this was expanded. But now it's a local, so it disappears before.

share/completions/cargo.fish
```diff
@@ -7,7 +7,7 @@ complete -c cargo -l list -d 'List installed commands'
...
 complete -c cargo -x -c cargo -n '__fish_seen_subcommand_from help' -a '$__fish_cargo_subcommands'
```
faho: Same.

share/completions/rustup.fish
```diff
@@ -223,10 +223,8 @@ set -l __rustup_toolchains (rustup toolchain list | string replace -r "\s+.*" ''
 # but a shorter version can be used if it is unambiguous.
 set -l __rustup_toolchains_short (__rustup_strip_common_suffix_strict $__rustup_toolchains)
 
-# Needs to be global to retain access from functions
-set __rustup_channels "beta" "stable" "nightly"
-
-set rustup_profiles "minimal" "default" "complete"
+set -lx __rustup_channels "beta" "stable" "nightly"
```
faho: Okay, I think this actually doesn't work after all. The variable is defined, the function that would use it is defined, the variable goes out of scope, then the function is called. I think `function --inherit-variable` could be used here, but there's no real need because this is only used in a single function (`__rustup_all_toolchains`), so just define it there.

share/completions/setfacl.fish
```diff
@@ -1,4 +1,4 @@
-set __fish_facl_spec_keywords default user group mask other
+set -lx __fish_facl_spec_keywords default user group mask other
```
faho: This thing is only used in the function right after it.

Just put it there, and not even as a variable:

```fish
for keyword in default user group mask other
```

share/completions/zfs.fish
```diff
@@ -10,7 +10,7 @@
 # - for the program command, complete the script to be executed
 # - for commands accepting several arguments of different types, propose arguments in the right order: for get, once the ZFS parameters have been given, only propose datasets
 
-set OS ""
+set -lx OS ""
```
faho: Yeah, this will have to be done via "--inherit-variable", so

```fish
function __fish_zfs_list_permissions -V OS
```

$OS is used in that one function and outside of functions, so that seems to be the simplest way to keep computing it only once.

share/completions/rustup.fish
ammgws: IIRC I couldn't get them to show up prior to this so that's a win-win.

faho:
Yup, that'll do. Thanks!

PR #6644, opened 2020-02-23
completions: mixed bag of updates/additions

ammgws:
Mixed bag of completion updates/additions

### DIFF
--- a/share/completions/telnet.fish
+++ b/share/completions/telnet.fish
@@ -1,25 +1,51 @@
-#
-# Load common telnet options
-#
-# Written by:  Sean C. Higgins
-#
+# First written by:  Sean C. Higgins
 
 complete -x -c telnet -d Hostname -a "(__fish_print_hostnames)"
 
+# Common across all telnet implementations
+complete -c telnet -s 4 -d "Use IPv4 addresses only"
+complete -c telnet -s 6 -d "Use IPv6 addresses only"
 complete -c telnet -s 8 -d (_ "Specifies an 8-bit data path")
-complete -c telnet -s 7 -d (_ "Do not try to negotiate TELNET BINARY option")
 complete -c telnet -s E -d (_ "Stops any character from being recognized as an escape character")
-complete -c telnet -s F -d (_ "Use local Kerberos authentication, if possible")
+complete -c telnet -s e -x -d "Set escape character"
 complete -c telnet -s K -d (_ "Specifies no automatic login to remote system")
-complete -c telnet -s L -d (_ "Specifies an 8-bit data path")
 complete -c telnet -s a -d (_ "Attempt automatic login")
 complete -c telnet -s c -d (_ "Disables reading user's .telnetrc")
-complete -c telnet -s d -d (_ "Sets debug mode")
-complete -c telnet -s S -x -d (_ "Sets IP TOS")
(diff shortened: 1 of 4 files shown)


share/completions/telnet.fish
```diff
@@ -1,25 +1,51 @@
...
 complete -c telnet -s l -x -a "(__fish_complete_users)" -d (_ "User login")
+complete -c telnet -s L -d (_ "Specifies an 8-bit data path")
 complete -c telnet -s n -x -d (_ "Log to tracefile")
-complete -c telnet -s x -d (_ "Turn on encryption")
 complete -c telnet -s r -d (_ "User interface similar to rlogin")
-complete -c telnet -s k -x -d (_ "Use Kerberos realm for authentication")
 
+switch (uname -s)
+  case FreeBSD NetBSD Linux
+    complete -c telnet -s d -d (_ "Sets debug mode")
```
faho: These completions set a terrible precedent - they not only use `_` explicitly (which you don't have to, we already translate descriptions internally, whereas `_` will call the gettext *command*), but they also don't quote it.

This means if you had a translation that either had no lines or more than one line you'd get a syntax error.

Please just use 

```fish
complete -c telnet -s d -d "Sets debug mode"
```

share/completions/whatis.fish
```diff
@@ -1,2 +1,14 @@
-
 complete -xc whatis -a "(__fish_complete_man)"
+
+complete -c whatis -s d -l debug -d "Debug"
+complete -c whatis -s v -l verbose -d "Verbose"
+complete -c whatis -s r -l regex -d "Interpret each keyword as a regex"
+complete -c whatis -s w -l wildcard -d "The keyword(s) contain wildcards"
+complete -c whatis -s l -l long -d "Do not trim output to terminal width"
+complete -rc whatis -s C -l config-file -d "Configuration file"
+complete -xc whatis -s L -l locale -a "(command -sq locale; and locale -a)" -d "Set locale"
```
faho: Good idea with the check!

share/completions/whatis.fish
```diff
@@ -1,2 +1,14 @@
-
 complete -xc whatis -a "(__fish_complete_man)"
+
+complete -c whatis -s d -l debug -d "Debug"
+complete -c whatis -s v -l verbose -d "Verbose"
+complete -c whatis -s r -l regex -d "Interpret each keyword as a regex"
+complete -c whatis -s w -l wildcard -d "The keyword(s) contain wildcards"
```
faho: "Allow wildcards"?

share/completions/telnet.fish
ammgws: Yeah I will remove them. Is there any place for them anymore in completions? There's probably some other older files which still use them.

share/completions/whatis.fish
ammgws: Thanks. It did seem awkward at the time.

share/completions/telnet.fish
faho: In descriptions definitely no. Like I said we translate them in the C++. Which is faster and means not having to type `(_` all the time.

I don't think there are many other uses.

faho:
Merged, thanks!

PR #6645, opened 2020-02-23
Remove gettext cruft from completions

ammgws:
As per https://github.com/fish-shell/fish-shell/pull/6644#discussion_r382986279
> No need for `_` in descriptions definitely no. Like I said we translate them in the C++. Which is faster and means not having to type (_ all the time.


faho:
Merged as f507f4ad1..e2ed4ec01 (because Github's conflict resolution UI apparently doesn't work).

PR #6647, opened 2020-02-23
doc: add interlinks between `true` ←→ `false` and `and` ←→ `or` cmds

Delapouite:
Hi

These new links offer a convenient way to jump back and forth between tightly related commands.

`See Also` sections already exist for a few other commands like `cdh`.

(a few blank lines were removed to keep format consistent between tandem commands)


ridiculousfish:
thank you!

Issue #4019, opened 2017-05-09
vim mode: deletion with movement to space doesn't delete


jacobsa:
I'm using fish 2.4.0 on Linux.

From zsh's vim mode I'm in the habit of deleting arguments by positioning myself at the start of the argument and then pressing `d t <space>`. In fish's vim mode this just moves me to the space, instead of deleting.

Would it be difficult to properly support deletion with a `t` movement, as in vim?

jacobsa:
(Also in the meantime does anybody have a good workaround for deleting long arguments?)

krader1961:
See the `fish_vi_key_bindings` script and the `bind` command. The vi bindings would need to introduce another binding sub mode that is entered when `dt` or `df` is seen. That sub mode would then delete up to (or including) the next character typed. This is certainly possible to implement and patches are welcomed to implement this behavior.

However, my recommendation is to get used to typing [meta-v] or [meta-e] to put the current command line into a real Vi editor for any non-trivial changes. That's because the fish Vi key bindings will probably never implement more than a tiny subset of Vi/Vim behaviors.

jacobsa:
Ah, I didn't know about [meta-v]. Is there documentation for `fish_vi_key_bindings` somewhere? The best I found was a single Stack Overflow question about how to turn it on.

That said, I can't figure out what my meta key is. I'm using the macOS Terminal.app to connect over ssh to Linux, with tmux running on Linux.

krader1961:
No, the fish key bindings are not well documented other than that they exist and how to invoke one of them. That should be improved. Patches are welcomed :smile:

The meta key in macOS Terminal.app is normally the [alt] or [option] key (depending on your keyboard). At least if the profile keyboard preferences has the "Use Option as Meta Key" is checked. That allows you to press the [alt] or [option] key and another key (e.g., `v`) to send the second key prefixed by escape. The [meta] key mechanism as a way to prefix the next key with the escape character dates back 20+ years to the Emacs editor.

krader1961:
Also, the [meta-v] and [meta-e] bindings where introduced in fish 2.5.0 or later. So they won't be available on your system running fish 2.4.0 unless you have installed custom bindings. You can type `bind \ev` to find out if that binding exists.

jacobsa:
Ah thanks, that explains why I haven't seen anybody mention it, and why it doesn't work. I will attempt to get fish 2.5.0 and try it out. Thanks!

jacobsa:
[...]

PR #6648, opened 2020-02-23
Add additional vi bindings

Vvalter:
Adds `dh`, `dl`, `c0`, `cf`, `ct`, `cF`, `cT`, `ch`, `cl`, and `y0` bindings.

## Description

I added a couple of missing vim bindings. This is kind of related to #4019 even though the issues does not mention the missing bindings explicitly.
The `dh` and `ch` bindings delete the first character of the line when used at the beginning of the line. This is different from vim, but I did not find a clean way to fix this and I consider this a minor difference.
I also noticed that the `y0` binding pastes if used at the beginning of the line. This is the same for the already existing `y^` binding so I guess it is not a problem?

## TODOs:


ridiculousfish:
thank you!

ridiculousfish:
Merged as e1a08e3f585a0b3a949f2980579b70b7d9ed7662

PR #6660, opened 2020-02-27
Change safe_strerror to only work for glibc

neheb:
Other platforms have thread-safe strerror implementations.

Amusingly, The %m format in glibc is threadsafe whereas strerror is not.

I know this to be true of at least Windows, macOS, FreeBSD, musl, and uClibc-ng.

### DIFF
--- a/src/wutil.cpp
+++ b/src/wutil.cpp
@@ -297,11 +297,7 @@ static inline void safe_append(char *buffer, const char *s, size_t buffsize) {
 // have to grub through sys_nerr and sys_errlist directly On GNU toolchain, this will produce a
 // deprecation warning from the linker (!!), which appears impossible to suppress!
 const char *safe_strerror(int err) {
-#if defined(__UCLIBC__)
-    // uClibc does not have sys_errlist, however, its strerror is believed to be async-safe.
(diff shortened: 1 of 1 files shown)


neheb:
This comment, dated 2019, says that strerrror under glibc is not yet thread safe: https://sourceware.org/bugzilla/show_bug.cgi?id=447#c3

zanchey:
The concern is not about thread safety, but about async-signal safety - see http://man7.org/linux/man-pages/man7/signal-safety.7.html etc - so I'm not sure this is appropriate,

neheb:
Hmm maybe strerror_l is a better function to use.

faho:
Yeah, no. This is one of those things that require extensive testing, especially on less common platforms (this includes Solaris and NetBSD).

And since there doesn't seem to be any actual benefit (apart from it being a bit cleaner), that does not seem to be worth it.

So I'm rejecting this.

(unless of course I've misunderstood and there is an actual issue this solves)

neheb:
Amusingly, POSIX requires %m to be available to syslog(3). Having it available elsewhere is a convenient extension.

PR #6661, opened 2020-02-27
completions: speed up python startup time

ammgws:
We only use the standard library so it should be OK to use python's [-S](https://docs.python.org/3/using/cmdline.html#id3) flag.

Before and after:
```
>time for i in (seq 50); python -c pass; end
Executed in    1.00 secs   fish           external 
   usr time  816.77 millis    0.00 millis  816.77 millis 
   sys time  183.10 millis   39.58 millis  143.52 millis 

>time for i in (seq 50); python -S -c pass; end
Executed in  505.14 millis    fish           external 
   usr time  336.00 millis    1.10 millis  334.90 millis 
   sys time  169.42 millis   33.09 millis  136.33 millis 
```

Also: found a bug in the bower completions which was causing it to hang. Seems like it was broken from the start.

### DIFF
--- a/share/completions/bower.fish
+++ b/share/completions/bower.fish
@@ -53,7 +53,7 @@ function __bower_list_installed
     if set -l python (__fish_anypython)
         # Warning: That weird indentation is necessary, because python.
         $python -S -c 'import json, sys; data = json.load(sys.stdin);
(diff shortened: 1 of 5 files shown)


faho:
Sweet, merged!

PR #6671, opened 2020-02-28
completions: add mpv

ammgws:
as per title


ammgws:
Actually, the ones generated by fish_update_completions are better since they include the description as well. However it does have one or more incorrect/invalid ones mixed in there.

@faho What's the go with adding completions in this case? Do we clean up the generated completions and commit that to the repo?

krobelus:
> Do we clean up the generated completions and commit that to the repo?

I think so, yes.

faho:
Merged, thanks!

PR #6672, opened 2020-02-28
completion helper functions: do not use gettext

ammgws:
Missed these in #6645


share/functions/__fish_config_interactive.fish
```diff
@@ -20,16 +20,16 @@ function __fish_config_interactive -d "Initializations that should be performed
     set -g __fish_active_key_bindings
 
     if not set -q fish_greeting
-        set -l line1 (_ 'Welcome to fish, the friendly interactive shell')
+        set -l line1 'Welcome to fish, the friendly interactive shell'
```
faho: No, these need gettext.

It's only the *completions* that don't (and hence also functions that are only used in completions).

ammgws:
Gotcha. I'll rebase this leaving just the completion helper functions

ammgws:
Hmm, is the one in __fish_print_packages OK to remove?

faho:
>Hmm, is the one in __fish_print_packages OK to remove?

Yeah. It's also only used in descriptions, and we expect those to be translated implicitly.

For non-completion use, that word would be useless anyway, so you'd expect anything after the tab to be removed.

(note: print_packages is of dubious value, as you only ever use the things it prints with package managers, in which case you only ever want the matching packages for that one)

ammgws:
Build failure due to cdbf0a9d2051a7292a3132df92cb2eea214d524c

faho:
>Build failure due to cdbf0a9?

Yeah, don't worry about it, it's unrelated to this PR.

faho:
Merged, thanks!

PR #6673, opened 2020-02-28
completions: add ps2pdf wrapper functions

ammgws:
as per title

### DIFF
--- a/share/completions/ps2pdf14.fish
+++ b/share/completions/ps2pdf14.fish
@@ -1,0 +1,1 @@
+complete -c ps2pdf14 --wraps=ps2pdf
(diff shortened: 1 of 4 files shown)


share/completions/ps2pdf14.fish
```diff
@@ -0,0 +1 @@
+complete -c ps2pdf14 --wraps=ps2pdf
```
faho: What are these? Old versions of this command?

ammgws: These ps2pdf scripts are installed by the `ghostscript` package. Currently `ps2pdf` targets PDF version 1.4, however it may change in a new version. The other scripts target specific versions of PDF such as 1.2, 1.3, etc.

faho:
Merged, thanks!

PR #6675, opened 2020-02-28
completions: add imv

ammgws:
as per title


share/completions/imv.fish
```diff
@@ -0,0 +1,13 @@
+complete -c imv -s h --description 'Show help message and quit.'
```
faho: Please remove these trailing periods. They're annoying and take up a column.

ammgws: Done

faho:
Thanks, merged!

PR #6685, opened 2020-02-29
added uncompressed pkg.tar files for completion

akspecs:
## Description
something like:
pacman -U name-of-package.pkg.tar
should auto-complete, as pkg.tar files can be files users quickly package without compression.
Fixes issue #
with changes brought forth in 3.1.0 to pacman.fish, uncompressed *.pkg.tar files did not tab-complete as expected, while compressed packages completed as expected.


faho:
Thank you, merged!

PR #6686, opened 2020-02-29
Make the history search term editable & add incremental search

krobelus:
Summary
------

After starting a history search by pressing the up arrow, the highlighted
search term can be edited. After doing so, pressing up again will resume
the search from where you left off but with the updated search term.

This PR also adds `incremental-search` as new readline function.
It is bound to `Control`+`R`, that attempts to fit into the existing history search interface.

Changes
-------

When doing a history search with an empty search string, the behavior
remains unchanged.  In particular, the cursor is positioned at the end
of the line after moving to another history entry.

When doing history search with a non-empty search string, the cursor will
be positioned at the end of the matching search string.  Subsequently,
the search term can be edited as long as the user only inserts or deletes
text inside the highlighted search string, or inserts text right after
the search string. Other edits end history search search as usual.
I added one exception here: if the matched search string extends to the end
of the command line, then appending to it also ends history search,
I think this is more intuitive.

The ability to edit the search term allows to clear it and then navigate
in the vicinity of the currently matched history entry.
For example, imagine you have a history like

        echo Some command I cannot recall
        echo Some command I do remember
        ...

And you want to recall the first command, but remember neither its name
nor any of its arguments.  Now you can simply type "remember" and press
the up arrow to go to the second command.  Then delete the search string
by pressing Control-W and press the up arrow to go to the first command.

Additionally, this changes readline command beginning-of-history to
move to the oldest search match. This seems like an alright fix because
I think it just used to do the same thing as end-of-history (move back to
the original search string).

Note that edits to the search string are part of the search's transient
edit, which means they will all be undone with a single undo
(every search should create exactly one atomic undo group).

TODO
----

- [ ] Do we want `incremental-search-forward`? (traditionally on `Control`+`S`).

- [ ] Currently fish moves the cursor to the end of the commandline
      after going to the next history entry.
      With this change we move it to the end of (non-empty) search strings -
      this could be too intrusive and break people's workflows, so we might
      want to only move it to the match when incremental search is active.

- [ ] I'm not sure what should happen when you search for `foo`, find
      something like `foo 123` and then want to prepend `echo ` to the
      command line. With this PR, fish also prepends `echo ` to the search
      string, so the highlighted search string will be `echo foo`. That
      extra highlighting may be annoying when you don't actually want to
      refine the search, which probably is most of the time.

Follow-ups:

- token search (M-.) should also allow editing the search term
- Add some kind of discoverable UI when incremental search starts, or when it is active. Something like `Type to search.`.


ridiculousfish:
Awesome! I'll give it a try!

krobelus:
@ridiculousfish sounds good; I'm by no means in a rush to merge this. The behaviour should be alright by now but I should split up the first commit and do some more cleanup.

I added incremental search now which should be quite convenient. It is bound to Control-R (I expect that it does not make sense to make it the default search behavior so we do need the extra key binding).
Overall it seems to fit in quite nicely, although I'm sure there are some details that are off.  Also, I've never felt comfortable with the incremental search in bash/zsh so I probably missed some of their features.

krobelus:
The behavior changes should now be only in the last three commits:
https://github.com/krobelus/fish-shell/compare/history-search-edit-needle~3...history-search-edit-needle

Plus the Page-Up fix in [history-search-edit-needle~6](https://github.com/krobelus/fish-shell/compare/history-search-edit-needle~6)
And an earlier fix to not move the cursor after failing search in [history-search-edit-needle~15](https://github.com/krobelus/fish-shell/compare/history-search-edit-needle~15)

shawnps:
> * [ ]  Do we want `incremental-search-forward`? (traditionally on `Control`+`S`).

I would love to have Control+S; coming from zsh it is as ingrained in me as Control+R. I am just one person though so I understand if others don't care about it as much.

krobelus:
After having used this for a while: being able to edit the history search term is cool, but rarely actually helpful in my experience.

Regarding the incremental search: I personally don't find it terribly useful and prefer the old up-arrow search. Naturally this could be reversed for users with different habits.

I think incremental search would be considerably more useful if it was fuzzy and case insensitive, just like the fish pager, or the Ctrl+R binding provided by elvish, or FZF's fish plugin. Implementing that by adapting the fish pager seems like the obvious solution to me now, which should also turn out less complex than trying to cram everything into the commandline buffer.

krobelus:
This needs another approach for the reasons for the reasons stated above.
I've adapted useful changes to master as 25fe353, 2b3b460 and 971c2eb.
Maybe it makes sense to look into how other programs like web browsers do history search. 
'm sure with the right design it's possible to make a nice UI for incremental search, but we should appreciate that the existing up-arrow search covers most scenarios perfectly fine!

ridiculousfish:
I have a branch which adds a control-r binding to [open history in the pager](https://github.com/ridiculousfish/fish-shell/tree/control_r). However the experience isn't yet great - the pager is just ugly and the command line jumps around if you select a line. I think we will need to improve the pager before this can be a good experience.

Abdillah:
I just migrate to Fish Shell, and a great shell btw! Hope this fit into this PR; my two cents, I have an uncovered use case.

Usually, I looked up through history using keyword without really using that as a completion (like "what's that command that can do A/B/C I executed yesterday"). The closest natural implementation in fish would be to add an ability to traverse the history
back and forth only as a suggestion (the completion part still in dim color and the cursor stay where it is).

Going for a bit of imagination, it will be powerfull if fish can do like this.

```
$ up -d

# With a "suggest history up" key would suggest
$ docker-compose up -d
# ^---- dim ---^ ^ hl^ (highligted)

# Doing "suggest history up" again, and another suggestion would shows up
$ roundup -d
# ^dim^^ hl^ (highligted)

# Doing "suggest history down" would suggest to later history

# And finally cancel key or change the keyword would remove the suggestion leave up with only the keyword.
$ up -d
```

This way we can reserve pager for more advanced use case. And IMHO, it is very close to what zsh / bash backward search covers.

ridiculousfish:
@Abdillah how does that idea differ from the existing history search (using up arrow)? Is it just about the colors?

Abdillah:
I've discovered that in the doc, I can cancel after the up arrow search. But I can't reproduce that (in my "cancel" sense). Instead, it stays on the picked up history item and when I press other key it remove the highlight (so, autocomplete on it). Is this the intended behavior or just my setting?

The other different are the cursor position to stay at the end of the keyword not on current history selection; and when pressing edit action (e.g. delete, left arrow), it should edit the keyword not the current picked up history item. It act as a suggestion after all, not autocomplete.

ridiculousfish:
Ok, thank you.

The intent is that if you press the cancel binding (escape by deafult), the history search ends and the commandline is restored to whatever it had before you hit up arrow. That's what I see when I try it.

Abdillah:
Yes, that alone would improve a workflow like this..

krobelus:
@Abdillah I just realized what you mean. I can reproduce like:

1. Enter `echo` and press up-arrow
2. Use any movement key like left-arrow. The `echo` part remains highlighted.
3. Press Escape. Nothing happens, because fish stopped the search in step 2.

This is because normal movement keys stop history search. We could change that, but then what happens when you are editing a multiline command. The cursor should go up but we are still in history search, so a naive implementation would go to the next match.

We could remove the highlighting the highlighting when moving the cursor. Does that make the overall experience better?

Also you can delete the entire line with Control+U instead of holding backspace.
In the upcoming fish 3.2.0 you'll be able to undo the history search with Control+Z

krobelus:
> The closest natural implementation in fish would be to add an ability to traverse the history back and forth only as a suggestion (the completion part still in dim color and the cursor stay where it was).

Yeah that's essentially in this patch series, except the dim part is normal (and the search term is highlighted).
However, I think this approach to editing the search term is too complex.

Abdillah:
> 2. Use any movement key like left-arrow. The echo part remains highlighted.
> 3. Press Escape. Nothing happens, because fish stopped the search in step 2.

Actually I press escape before left-arrow, and still nothing happened. But the search stopped and complete my hightlighted keyword (although in my intuition pressing ESC should "cancel and revert" not "leave on the current completion").

> However, I think this approach to editing the search term is too complex.

I see. I'm blind with the code currently. But implementing what @ridiculousfish [underlined](https://github.com/fish-shell/fish-shell/pull/6686#issuecomment-674650156) enables the user "edit the search term". Although the search will be reset from the start.

> We could remove the highlighting the highlighting when moving the cursor.

Yes for the current behavior, although that's not my idea's focus :smile:.

> Also you can delete the entire line with Control+U instead of holding backspace.

I see. ESC then Ctrl+U then retype the keyword would be good for now. Thank you.

Abdillah:
I am playing a bit with the code and this is a rough implementation on the cancel behavior.

```diff
diff --git a/src/reader.cpp b/src/reader.cpp
index 9928af6d5..ef1de193b 100644
--- a/src/reader.cpp
+++ b/src/reader.cpp
@@ -2689,7 +2689,10 @@ void reader_data_t::handle_readline_command(readline_cmd_t c, readline_loop_stat
             break;
         }
         case rl::cancel: {
-            // The only thing we can cancel right now is paging, which we handled up above.
+            if (history_search.active()) {
+                set_command_line_and_position(&command_line, L"" + history_search.search_string(), history_search.search_string().length());
+                history_search.reset();
+            }
+            // We also cancel paging, which we handled up above.
             break;
         }
         case rl::repaint_mode: {
```

krobelus:
Okay, does that fix it for you? I'm a bit puzzled because it works without this patch for me.
It should already be handled by this snippet in line 3615:

```cpp
            if (history_search.active() && command_ends_history_search(readline_cmd)) {
                // "cancel" means to abort the whole thing, other ending commands mean to finish the
                // search.
                if (readline_cmd == rl::cancel) {
                    // Go back to the search string by simply undoing the history-search edit.
                    clear_transient_edit();
                }
                history_search.reset();
            }
```

Can you show how to reproduce this? Which version of fish are you using? The cancellation works fine for me following the steps below.

1. Start fish without configuration: `env HOME=(mktemp -d) fish`
2. Run two commands: `echo 1` and `echo 2`.
3. Enter `echo`, press up-arrow (selects `echo 1`)
4. Press Escape to go back to `echo`

jcelliott:
@ridiculousfish since you mentioned the pager approach (and because I've been following this conversation for many years 😄), I would like to mention that I've been using [this fuzzy search approach](https://github.com/oh-my-fish/plugin-peco/blob/master/functions/peco_select_history.fish) bound to ctrl-r for several years now and it's really a fantastic experience. It leaves the simpler up-arrow history search untouched, and gives you a more powerful search when you need it. 

So, what I'm really saying is I don't think fish needs any [additional search features](https://fishshell.com/docs/current/design.html#the-law-of-orthogonality). What exists already works well, and users can add more if they want it.

JoeKays:
Just as a note for everyone still not satisfied, here is a program that I used a lot in the past and it puts a different spin on it: https://github.com/dvorka/hstr

alper:
@jcelliott Good that you mention that. I've found that there is a good plugin for fzf history search in fish, so that does solve my problem as well.

sekoenig:
@krobelus I realize from the discussion here that the incremental search as in `krobelus/history-search-edit-needle`  is probably is probably abandoned, but I have actually gotten used to it over the last couple of months and like it.

Now I am trying to make it work with the current HEAD.  I managed to get a manual rebase of the branch to compile, but unfortunately it doesn't work: already the behavior introduced [here](https://github.com/krobelus/fish-shell/commit/8ffc74820a1153a8698d1f0472ffeee273363465) is not as it should be: the highlighting is missing for newly added text and pressing up the second time doesn't update the search string.

Is there any chance you might take a look and help debug this?

krobelus:
@sekoenig your branch looks like good work; as far as I can tell you got most things right.
I just pushed my rebased branch. It seems to work as before (although the added  tests are failing, I didn't investigate why).
It should be usable for the time being.

> already the behavior introduced here is not as it should be

I think I ran into the same problem, it is fixed by using the correct `push_edit` in `insert_string`, see the range-diff between our branches. I didn't test too much, so I don't really know if there are glitches, especially with the recent changes to undo.

>  highlighting is missing for newly added text

Ok I just realized that that's still the case (probably because `history_search_text_if_active` has replace `highlight_search`)

sekoenig:
> @sekoenig your branch looks like good work; as far as I can tell you got most things right.
> I just pushed my rebased branch. It seems to work as before (although the added tests are failing, I didn't investigate why).
> It should be usable for the time being.

Thank you!

> > already the behavior introduced here is not as it should be
> 
> I think I ran into the same problem, it is fixed by using the correct `push_edit` in `insert_string`, see the range-diff between our branches. I didn't test too much, so I don't really know if there are glitches, especially with the recent changes to undo.

I see, that would have taken me a while to figure out.  I was slowly working through the code to understand the logic, very happy to have those efforts cut short now ;)

> > highlighting is missing for newly added text
> 
> Ok I just realized that that's still the case (probably because `history_search_text_if_active` has replace `highlight_search`)

Okay, this I think I could fix by transplanting a few lines from the old `highlight_search` into `paint_layout`.  I will check a bit more more this is fine and then push to my branch.

sekoenig:
Okay, the top commit of `sekoenig/history-search-edit-needle` now uses my brute-force fix for the highlighting.  I believe the proper way to do this is probably calling `make_layout_data` in an appropriate place while the search string is being edited, but I didn't quite find that place...

kobiebotha:
As a first time user to fish, reverse-i-search was basically a non-starter for me. Now I have it thanks to this work. @sekoenig I hope your history-search-edit-needle branch makes it into fish master. I hope this feature makes it into `fish-shell:master`. Wasn't sure where to provide that feedback, so just dropping it here....

jodaka:
> As a first time user to fish, reverse-i-search was basically a non-starter for me. Now I have it thanks to this work. @sekoenig I hope your history-search-edit-needle branch makes it into `fish-shell:master`. Wasn't sure where to provide that feedback, so just dropping it here....

Is there a binary or one need to checkout sources and build from there? Would also like to try that mechanism, but I'm a bit lazy 🙈

sekoenig:
> As a first time user to fish, reverse-i-search was basically a non-starter for me. Now I have it thanks to this work. @sekoenig I hope your history-search-edit-needle branch makes it into `fish-shell:master`. Wasn't sure where to provide that feedback, so just dropping it here....

I just want to point out that all credit for this goes to @krobelus.  He's the original author of the branch, I merely tried (unsuccessfully) to rebase it on top of the recent fish master branch, which @krobelus eventually did successfully.

I also hope this will be merged to master eventually!

sekoenig:
> > As a first time user to fish, reverse-i-search was basically a non-starter for me. Now I have it thanks to this work. @sekoenig I hope your history-search-edit-needle branch makes it into `fish-shell:master`. Wasn't sure where to provide that feedback, so just dropping it here....
> 
> Is there a binary or one need to checkout sources and build from there? Would also like to try that mechanism, but I'm a bit lazy see_no_evil

I believe you'll need to go through the source, but it's really quite easy.  At the time of writing this, `history-search-edit-needle` still rebases smoothly onto the current fish master/main (HEAD affc4e32).

ewtoombs:
> This needs another approach for the reasons stated above.
> I've adapted useful changes to master as [25fe353](https://github.com/fish-shell/fish-shell/commit/25fe353187c8cf2bb9b6eecda6433ec619e25019), [2b3b460](https://github.com/fish-shell/fish-shell/commit/2b3b460264d614cd828cdff9d9d37818e447bb1d) and [971c2eb](https://github.com/fish-shell/fish-shell/commit/971c2eb668bb3d58facb42c8f556858ac92b25c0).
> Maybe it makes sense to look into how other programs like web browsers do history search.
> I'm sure with the right design it's possible to make a nice UI for incremental search, but we should appreciate that the existing up-arrow search covers most scenarios perfectly fine!

The existing search does not cover most scenarios. It only feels that way because the scenarios that it doesn't cover can' be done and the user knows that ahead of time, so the user never attempts them. Only people who used ^R in zsh and bash will tell you this, though. You should listen to them. They're not crazy. Well, they are crazy, but that's not the point. Even a stopped clock is right twice a day. Unless you live somewhere that has adopted 24-hour time.

ewtoombs:
Mind expanding on that, @faho?

faho:
>The existing search does not cover most scenarios.

That's.. just wrong. I've had this conversation a bunch of times, and typically the cause is that people don't understand the existing search.

It can search everything, it does infix and fuzzy search. Almost literally the only difference is that the search term isn't easily editable *while doing the search*. You need to stop, edit the search term and start the search again. Well, that and that you don't need to press "ctrl-r" to start it.

That's why this is so frustrating - we get a lot of people asking, but nobody can really explain what they want, and when they do try to explain it usually turns out that it was already possible, they just didn't even attempt it.

faho:
>Mind expanding on that, @faho?

Expanding on what? Did github show a comment before I was done writing it?

ewtoombs:
No, you did expand on it. I was talking about the downvote.

ewtoombs:
So, how do you stop the fish history search, edit the search term, and start it again?

faho:
Press escape

ewtoombs:
aw, wth! That's so SO much better than what I was doing. But it's undocumented. This desperately needs to be added to the [Searchable command history](https://fishshell.com/docs/current/interactive.html#searchable-command-history) section!!

ewtoombs:
One thing though: perhaps you could em the words "escape key" in the docs. That would be easier to read.

borring:
Is Ctrl+R still being persued, or has it been abandoned?

Abdillah:
> Is Ctrl+R still being persued, or has it been abandoned?

@borring, It worked as in bash, fyi: Ctrl+R for searching history then use Esc to cancel. Mine is fish v3.3.1.

jodaka:
> > Is Ctrl+R still being persued, or has it been abandoned?
> 
> @borring, It worked as in bash, fyi: Ctrl+R for searching history then use Esc to cancel. Mine is fish v3.3.1.

I'm also on 3.3.1 and Ctrl+R does nothing. Do I need to configure something? Using fish with iterm2 on macos

faho:
As you might see, this PR was *not* merged, and there is, as far as I'm aware, no ongoing work in this area.

The reason is that this is a complicated addition for very little gain. Usually users are simply confused and want bash's ctrl-r, while fish actually already offers the functionality they are looking for. Instead of pressing ctrl-r and then searching, simply enter your search string and press up-arrow! That will already search! It will do a nice infix-search, and mark the search string and everything. It is enabled by default! 

The only real difference between that and ctrl-r that anyone has been able to figure out after years of discussions is that ctrl-r allows you to change the search string mid search. Which is a thing, but not super important.

Abdillah:
Ah yes, I'm forgot to add type your keyword first and then press Ctrl+R, Esc to cancel.

ewtoombs:
> As you might see, this PR was _not_ merged, and there is, as far as I'm aware, no ongoing work in this area.
> 
> The reason is that this is a complicated addition for very little gain. Usually users are simply confused and want bash's ctrl-r, while fish actually already offers the functionality they are looking for. Instead of pressing ctrl-r and then searching, simply enter your search string and press up-arrow! That will already search! It will do a nice infix-search, and mark the search string and everything. It is enabled by default!
> 
> The only real difference between that and ctrl-r that anyone has been able to figure out after years of discussions is that ctrl-r allows you to change the search string mid search. Which is a thing, but not super important.

I might add that fish actually *does* let you change the search string without having to type it out all over again, by hitting Esc in the middle of a search. You can't resume the old search at the same point in history, but that is a stupid thing to want to do anyway. You changed the search terms. Why wouldn't you want to start the search from the beginning again?

ewtoombs:
One more thing, lol. After giving it some more thought, there actually is a use case for ctrl-r-style incremental searching that fish doesn't cover. You do one search first to get to the right point in history. Perhaps you want to skip to about six months ago, when you were working on the postgres database. Then, once you're at that point in history, you start a new search with _completely different search terms_, basically using the first search as a filter, to only do the second search on commands that happened around six months ago when you were working on postgres. Getting more specific:

1. Find the right point in history: ctrl-r "psql" ctrl-r ...
2. Stop the first search: Esc
3. Start a new search: ctrl-r "rm" ctrl-r ctrl-s ...

That looks for remove commands that happened in around the last time psql was run, which was six months ago. If you just searched for "rm", you'd be scanning for minutes, probably, since it's such a common command. If you only know _that_ a file was removed, and you can't remember the name, you can't get more specific. And since only commands with the word "rm" are being shown, you will only know whether you're in the right spot if you actually recognise some part of the rm command.

So, bash-style incremental search actually can be useful. But I have personally never tried this. Actually, I only just realised just now, while contriving this scenario that I might actually want to do this. But now that I see it is possible, I might find myself wanting to do this lol

alexeymuranov:
Then it would probably make more sens to save timestamps in history and allow to navigate it alternatively by string search or by time...

ewtoombs:
> Then it would probably make more sens to save timestamps in history and allow to navigate it alternatively by string search or by time...

Well, fish does keep timestamps of all commands. As for time-based search, well, I have no idea how I'd go about doing that. Any ideas?

PR #6688, opened 2020-03-01
Fix expand.cpp compile failed with old SDK version

mczoden:
## Description

Workaround method to fix compile issue on old OSX SDK (10.11.6)
<sys/time.h> should be included ahead of <sys/proc.h>

## TODOs:

### DIFF
--- a/src/expand.cpp
+++ b/src/expand.cpp
@@ -18,6 +18,7 @@
 #endif
 #if __APPLE__
 #include <sys/time.h>  // Required to build with old SDK versions
+// proc.h needs to be included *after* time.h, this comment stops clang-format from reordering.
 #include <sys/proc.h>
 #else
 #include <dirent.h>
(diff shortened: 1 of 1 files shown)


faho:
So this fails on old macOS?

If so, please add a comment to explain the dependency, in a line *between the two*.

These were reordered by running `clang-format`, which sorts includes by default, and as far as I know it only orders "groups", which are separated with empty or comment lines.

I.e. 

```c++
#include <sys/time.h>  // Required to build with old SDK versions
// proc.h needs to be included *after* time.h, this comment stops clang-format from reordering.
#include <sys/proc.h>
```

mczoden:
Yes, failed on OS X 10.11.6 indeed

I will try to add a comment between these 2 lines

faho:
>Do I need to close this request firstly?

Of course not. Just push a new commit (in addition to the old one or via a force push) to the PR branch.

faho:
Thank you, merged!

PR #6697, opened 2020-03-01
Privatise completion-only functions

ammgws:
Leads to less pollution when tab completing `fish_` to find actual end-user functions,
 and brings the naming in line with the other completion helper functions.

Also considered these two but they are used in prompts so left them alone:
- fish_print_git_action.fish
- fish_print_hg_root.fish


share/completions/gradle.fish
```diff
@@ -45,7 +45,7 @@ function __cache_or_get_gradle_completion
     end
     mkdir -m 700 -p $XDG_CACHE_HOME/gradle-completions
 
-    set -l hashed_pwd (fish_md5 -s $PWD)
+    set -l hashed_pwd (__fish_md5 -s $PWD)
```
ridiculousfish: The file was renamed but the function was not, so there isn't actually a __fish_md5 function.

ammgws:
Sorry for missing that. Done.

ridiculousfish:
Merged as 926c1f10161495c032d6b3141e947e923a5aefe5, thank you!

PR #6700, opened 2020-03-02
completions: update GNU wc options

ammgws:
[skip-ci]


krobelus:
As far as I know this is left out intentionally, since other variants of `wc` do not support it.
We can check that the GNU variant is installed (for example by running a command that only succeeds when it is) and then conditionally add the completion.

ammgws:
On phone now but isn't the GNU one where I added it? (The one with long opts)

krobelus:
Oh yeah, of course, my bad.

PR #6701, opened 2020-03-02
Fix 256 color detection on xterm-like

Shaac:
bbc3fecbe introduced a regression where support for 256 color was not
detected on xterm-like terminals that did not define the TERM_PROGRAM
env variable. Almost no terminal on linux define this variable.

## Description

The upgrade from fish 3.0 to 3.1 broke the color on my terminal (kitty). Its `TERM` environment variable is `xterm-kitty`, and it does not define a `TERM_PROGRAM` env variable (neither do xterm, or urxvt). I changed the logic of checking for Terminal.app back to what it was before bbc3fecbe and fixed the problem.

## TODOs:


faho:
Thanks, nice catch. Merged to master, cherry-picked to 3.1.1 as e0645c947.

Issue #3861, opened 2017-02-21
abbreviations are not mentioned in the tutorial or documentation


nhooyr:
They are only mentioned in the commands reference. I've been using fish for several months, have read all of the tutorial and http://fishshell.com/docs/current/index.html but I had absolutely no idea that this immensely useful feature existed.

It should really be mentioned in the tutorial or docs.

PR #6705, opened 2020-03-03
doc: add section about abbreviations in the index

Delapouite:
Hello

In this issue from 2017 https://github.com/fish-shell/fish-shell/issues/3861 :

> I had absolutely no idea that this immensely useful feature existed

I agree that abbreviations deserve a more prominent  place in the doc.
For consistency, the `gco` example is the same than on the [abbr command page](http://fishshell.com/docs/current/cmds/abbr.html).

Regarding the order, this addition comes right after the sections describing related concepts like aliases.


ridiculousfish:
Merged as fba6da983b7f5265bf8364192786e987e258da7c, thank you!

PR #6706, opened 2020-03-03
Move logic to remove trailing spaces when pasting to C++

### LINKED ISSUE #6704: Pasting a lot hangs the shell again
In 3.1.0, if you paste a long string, it results in thread exhaustion which effectively hangs the shell. We should take a fix for 3.1.1.

krobelus:
Fixes #6704

Removing trailing spaces can be quite useful, especially when `fish_clipboard_paste` is not available (ssh/container/vm). I'm trying to do the equivalent thing in C++, I hope this doesn't break anything.

### DIFF
--- a/share/functions/__fish_shared_key_bindings.fish
+++ b/share/functions/__fish_shared_key_bindings.fish
@@ -159,9 +159,6 @@ function __fish_shared_key_bindings -d "Bindings shared between emacs and vi mod
     # into one literal token, to facilitate pasting non-code (e.g. markdown or git commitishes)
     bind --preset -M paste "'" "__fish_commandline_insert_escaped \' \$__fish_paste_quoted"
     bind --preset -M paste \\ "__fish_commandline_insert_escaped \\\ \$__fish_paste_quoted"
-    # Only insert spaces if we're either quoted or not at the beginning of the commandline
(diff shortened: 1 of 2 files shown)


krobelus:
When building with `-fsanitize=thread` the tests fkr.expect, signals.expect and wait.expect
also fail on master for me (fkr also spawns a fish_key_reader that maxes out one CPU).

ridiculousfish:
Thank you for going at this! I think this particular approach is too inflexible, because it "bakes in" the logic in a way that users cannot easily discover it, or disable it. (There was a recent issue where a user asked how to disable paste-magic though I can't find it right now).

My weak preference is to revert the initial strip-spaces change a5a643f854db2e52fe72c56fbd6a8cc6636e7623, but if someone feels strongly about keeping it, my second suggestion is to embody the space-prefix-stripping logic in a new readline function, perhaps `self-insert-lstrip` (or whatever). This would be the same as `self-insert` but with this extra logic.

Then fish would ship:

    bind --preset -M paste ' ' self-insert-lstrip

and we could document it.

krobelus:
Yep a new readline command would be much better if we want to keep this.

Maybe it is better to revert this feature. It's hard to tell how often this would actually be useful, so I don't know.

krobelus:
superseded by #6713

PR #6709, opened 2020-03-04
docs: Move completion tutorial out into its own document

faho:
This is useful, but it's a rather large thing that's uninteresting to
many users. So we put it into its own (top-level, for now) document.

The reason for this PR is that we have a note somewhere saying that we shouldn't have more separate documents (might have been in the old doxygen docs somewhere), but to be quite honest I disagree with this.

And not just for this one case, tho this is egregious, but also potentially for more things. Not that I'd want one document per "chapter", but the *7* docs we have after this change (and that's counting the license and design docs) are too few.

## TODOs:

### DIFF
--- a/doc_src/completions.rst
+++ b/doc_src/completions.rst
@@ -1,0 +1,120 @@
+.. _completion-own:
+
+Writing your own completions
+============================
+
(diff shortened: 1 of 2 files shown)


ridiculousfish:
I very much agree with this direction, the index file is a wall of text and we should break it up. LGTM!

PR #6712, opened 2020-03-04
doc(end): mention the 'function' command and add links to all block starters

Delapouite:
Hi

`function` was missing from the synopsis. Also convenient links have been added to all block constructs commands.


faho:
Thank you, this is a very good catch!

PR #6713, opened 2020-03-05
Add self-insert-nonempty as fast implementation of space-trimming

ridiculousfish:
(Note if we merge this, it will be into 3.1.1 as well as master.)

This implements the idea I sketched in [this comment](https://github.com/fish-shell/fish-shell/pull/6706#issuecomment-594348387). It adds a new readline command `self-insert-nonempty` which inserts characters unless the command line is empty.

The space-trim-on-paste binding then becomes `bind --preset -M paste " " self-insert-nonempty` and it is quite noticeably faster.

For 3.1.1, I am OK merging this or reverting a5a643f854d. Post if you have strong feelings either way.


faho:
This should be documented in cmds/bind.rst.

src/input_common.h
```diff
@@ -111,6 +112,17 @@ class char_event_t {
     /// The type of event.
     char_event_type_t type;
 
+    /// Hackish: the input style, which describes how char events (only) are applied to the command
+    /// line. Note this is set only after applying bindings; it is not set from readb().
+    enum input_style_t : uint8_t {
```
krobelus: I was surprised for a bit because the enum is called `input_style_t` but it's used elsewhere using `char_event_t`.
It's obvious after a second look, but maybe it's better to make it a top-level `enum class` like the others above, since it needs to be public anyways.

src/reader.cpp
```diff
@@ -2542,47 +2543,36 @@ struct readline_loop_state_t {
...
-        }
-        // Only allow commands on the first key; otherwise, we might have data we
-        // need to insert on the commandline that the command might need to be able
-        // to see.
-        auto next_event = inputter.readch(false);
-        if (event_is_normal_char(next_event)) {
-            arr[i] = next_event.get_char();
+        } else if (evt.input_style == char_event_t::style_nonempty && accumulated_chars.empty() &&
+                   active_edit_line()->empty()) {
+            // Skip this ltrim character.
```
faho: Let's see if I understand this correctly: this bit is just an optimisation?

We can skip all normal chars on an empty commandline if the binding used is self-insert-nonempty. If anything else modifies the commandline, we won't be going through this path, and then once we do go back here the commandline might not be empty anymore.

src/reader.cpp
```diff
@@ -2542,47 +2543,36 @@ struct readline_loop_state_t {
...
             break;
-        }
-        // Only allow commands on the first key; otherwise, we might have data we
-        // need to insert on the commandline that the command might need to be able
-        // to see.
-        auto next_event = inputter.readch(false);
-        if (event_is_normal_char(next_event)) {
-            arr[i] = next_event.get_char();
+        } else if (evt.input_style == char_event_t::style_nonempty && accumulated_chars.empty() &&
+                   active_edit_line()->empty()) {
```
krobelus: Presently we do the space trimming whenever the cursor is at the beginning of a commandline.
Doing it instead only when the command line is empty feels like a benign change.

src/reader.cpp
ridiculousfish: Right, this is just an optimization, but an important one . This loop accumulates normal input, while there's input available. If we break out of the loop, we end up redrawing.

src/reader.cpp
ridiculousfish: Good point. After reflection I think the beginning-of-commandline behavior is better. I renamed the binding to `self-insert-notfirst`

ridiculousfish:
Merged into master as e3e82f58739a972c6773f0dca40cea102560cfa7 and Integration_3.1.1 as 4c30e5ad44bf465c7ee3e46bc5bbbaf78d0e924f

PR #6720, opened 2020-03-07
completions: remove unnecessary use of `--erase`

ammgws:
Random use of `--erase` apparently stems from an old misunderstanding of the completion system.


ammgws:
`rustup` is the only one in the list that seems to provide completions, but they aren't eagerly loaded, just provides a feature that generates them for you.

```
FISH:

    Fish completion files are commonly stored in
    `$HOME/.config/fish/completions`. Run the command:

        $ mkdir -p ~/.config/fish/completions
        $ rustup completions fish > ~/.config/fish/completions/rustup.fish

    This installs the completion script. You may have to log out and
    log back in to your shell session for the changes to take affect.
```

mqudsi:
Yes, I intentionally added the `--erase` to the `rustup` completions because they are naive and generate incorrect results that my version of the completions can filter out.

ammgws:
@mqudsi OK I re-added it with a comment

faho:
Wait, hold on.

>Yes, I intentionally added the --erase to the rustup completions because the upstream completions are naive and generate incorrect (well, spurious) results that my version of the completions can filter out.

How are these completions loaded?

>The assumption is that some distro maintainers are generating those completions and including them in the packages.

If the distro maintainers do `rustup completion > /usr/share/fish/vendor_completions.d/rustup.fish` erasing the completions won't have any effect, because these won't be loaded.

Fish loads the first file with the right name in $fish_complete_path. The other files are never touched.

So the only reason you need to erase completions is if you did `rustup complete | source` in config.fish (or something to that effect).

mqudsi:
I didn’t realize multiple completions aren’t loaded.

faho:
Cherry-picked to master as 7fb3880b9. Thanks!

PR #6721, opened 2020-03-07
doc: add interlinks between `break` and `continue` commands

Delapouite:
Hi

This commit adds interlinks between the `break` and `continue` commands


faho:
As always, thank you very much!

PR #6722, opened 2020-03-08
fish_indent: Remove useless quotes

faho:
## Description

This changes fish_indent so that it strips quotes where they don't have any use. 

In particular what it will do is turn this

```fish
set --local composer_cmds 'about' 'archive' 'browse' 'check-platform-reqs' 'clear-cache' 'config'
```

into this:

```fish
set --local composer_cmds about archive browse 'check-platform-reqs' 'clear-cache' config
```

As you can see it currently doesn't handle everything but instead tries unquoting and if that only leaves alphanumeric characters it leaves the quotes off.

(I initially tried a complicated scheme in which I'd try unescaping twice to see if anything changes, but that had issues, in particular with chars like `^` or `%` that might only be special some of the time).

## TODOs:


src/fish_indent.cpp
```diff
@@ -254,9 +254,8 @@ void prettifier_t::prettify_node(const parse_node_tree_t &tree, node_offset_t no
                 }
                 wcstring unescaped{source, node.source_start, node.source_length};
                 // Unescape the string - this leaves special markers around if there are any expansions or anything.
-                // TODO: This also already computes backslash-escapes like \u or \x.
-                // We probably don't want that - if someone picked `\x20` to spell space, they have a reason.
-                unescape_string_in_place(&unescaped, UNESCAPE_SPECIAL);
+                // We specifically tell it to not compute backslass-escapes like \U or \x, because we want to leave them intact.
```
krobelus: `s/ss/sh/`

src/common.h
```diff
@@ -121,7 +121,8 @@ enum escape_string_style_t {
 enum {
     UNESCAPE_DEFAULT = 0,         // default behavior
     UNESCAPE_SPECIAL = 1 << 0,    // escape special fish syntax characters like the semicolon
-    UNESCAPE_INCOMPLETE = 1 << 1  // allow incomplete escape sequences
+    UNESCAPE_INCOMPLETE = 1 << 1,  // allow incomplete escape sequences
+    UNESCAPE_NO_BACKSLASHES = 1 << 2,  // compute backslash escapes
```
krobelus: something like "keep" instead of "compute"?

src/fish_indent.cpp
```diff
@@ -251,9 +252,28 @@ void prettifier_t::prettify_node(const parse_node_tree_t &tree, node_offset_t no
...
+                // Remove INTERNAL_SEPARATOR because that's a quote.
+                auto quote = [](const wchar_t ch) {
+                                 return ch == INTERNAL_SEPARATOR;
+                             };
+                unescaped.erase(std::remove_if(unescaped.begin(), unescaped.end(), quote), unescaped.end());
+
+                // If no non-alphanumeric char is left, use the unescaped version.
+                // This can be extended to other characters, but giving the precise list is tough,
+                // can change over time (see "^", "%" and "?", in some cases "{}") and it just makes people feel more at ease.
+                if (std::find_if_not(unescaped.begin(), unescaped.end(), fish_iswalnum) == unescaped.end() && !unescaped.empty()) {
```
krobelus: Can we add just `_` and `-` to the allowed characters? Since they are commonly used for compound names.

src/fish_indent.cpp
```diff
@@ -251,9 +252,28 @@ void prettifier_t::prettify_node(const parse_node_tree_t &tree, node_offset_t no
...
                 }
-                output.append(source, node.source_start, node.source_length);
+                wcstring unescaped{source, node.source_start, node.source_length};
+                // Unescape the string - this leaves special markers around if there are any expansions or anything.
+                // TODO: This also already computes backslash-escapes like \u or \x.
+                // We probably don't want that - if someone picked `\x20` to spell space, they have a reason.
+                unescape_string_in_place(&unescaped, UNESCAPE_SPECIAL);
+
+                // Remove INTERNAL_SEPARATOR because that's a quote.
+                auto quote = [](const wchar_t ch) {
```
krobelus: I'd not use `const` when passing by value (if that still compiles).

src/common.h
faho: This was supposed to be "don't compute"

src/fish_indent.cpp
faho: I'd like it if we could anchor this to some other character set. I don't think we have a set of "safe" characters anywhere. We could also pick valid variable names, which is this plus "_", but not "-".

Note that this would have real impact on how our completions format - there are some that have "kebab-case" commands.

Or we just declare a random set of valid characters, that wouldn't be too awful.

src/fish_indent.cpp
faho: Yeah, a backlass is something entirely different.

src/fish_indent.cpp
faho: Okay, I pushed an additional list of "-", "_" and "/". To show the difference I added another commit just with those changes, but I'll remove that when merging.

src/fish_indent.cpp
faho: Yeah, compiles.

faho:
Merged as b25f72f39..9f984ee89 after re-running the indent (which also gets around merge conflicts) and making one teensy change to a check.

Issue #6724, opened 2020-03-09
scp remote path completion broken


rnhmjoj:
### Description
`scp` used to have completion of remote paths via ssh.
`rsync` completion, on the other hand, still works as expected.

### How to reproduce
1. Setup a remote host with working ssh connection
2. Type `scp user@host:<tab><tab>`
3. nothing happens

Doing the same with rsync should list the paths in ~/ are `host`.

### Info
OS: NixOS 20.0.3
fish version: 3.0.2
default fish configuration

PR #6725, opened 2020-03-09
completions: fix scp remote path when /bin/ls does not exist

rnhmjoj:
## Description

This should fix issue #6724 
I gave a quick look and figured out what the problem was.

## TODOs:


faho:
Could we not rely on $PATH and just use `ls`?

rnhmjoj:
How would you do that? I honestly don't know.

faho:
Just use `ls` instead of `/bin/ls`.

`ssh` starts a shell, so that should have a working $PATH.

rnhmjoj:
> ssh starts a shell, so that should have a working $PATH.

Ok, that was pretty obvious. I was somehow conviced a shell or executable were needed.

faho:
Merged, thanks!

rnhmjoj:
Thank you!

PR #6726, opened 2020-03-09
time builtin: align output columns on rare cases

afq984:
1. When the wall time and cpu time rows has different units
   e.x. running multiple cores
2. When duration is around 1E3 or 1E6 micros
   printf("%6.2F", 999.995) gives 1000.00 which is 7 digits
---
The included test looks like this without the modification:
```
Executed in  500.00 micros    fish           external 
   usr time    1.00 secs  1000.00 millis  1000.00 micros 
   sys time    1.00 secs  999.99 millis  500.00 micros
```

### DIFF
--- a/src/fish_tests.cpp
+++ b/src/fish_tests.cpp
@@ -28,6 +28,7 @@
 #include <algorithm>
 #include <array>
 #include <atomic>
+#include <chrono>
 #include <cmath>
 #include <cstring>
 #include <cwchar>
@@ -72,6 +73,7 @@
 #include "redirection.h"
 #include "screen.h"
 #include "signal.h"
+#include "timer.h"
 #include "tnode.h"
 #include "tokenizer.h"
 #include "topic_monitor.h"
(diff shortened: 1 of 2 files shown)


krobelus:
> When duration is around 1E3 or 1E6 micros
printf("%6.2F", 999.995) gives 1000.00 which is 7 digits

So `printf` rounds float arguments, it looks like we first need to do that rounding before we know how many digits there are, so we can choose the correct unit.

I imagine printing a float value with `sprintf` and reading it back with `sscanf` should do the trick, though there may be more modern solutions.

afq984:
> I imagine printing a float value with `sprintf` and reading it back with `sscanf` should do the trick, though there may be more modern solutions.

[snprintf with size 0 string would return the number of characters needed for the print](http://man7.org/linux/man-pages/man3/printf.3.html#RETURN_VALUE).
That is `snprintf(0, 0, "%6.2F", 999.995) => 7`; `snprintf(0, 0, "%6.2F", 999.994) => 6`.

---

I implement it as a check for (>= 1000 and >= 999995) because `get_unit` takes `int64_t`, and in master, the only values that trigger the `%6.2F` overflow are
* `1000`, `1000000` (because of the use of `>` instead of `>=`)
* `999995` to `999999` (because of rounding)

There isn't a `1000.00 secs` problem because we switch to `15.00 mins` after `900.00 secs`.

krobelus:
Nice I didn't know that about snprintf.

I can see now why this is sufficient, merged with thanks!

PR #6727, opened 2020-03-09
doc: add links between the string-split and read commands

Delapouite:
Hello

This PR adds various missing links between commands.


faho:
Thank you!

PR #6729, opened 2020-03-10
fix(cd): remove duplicated comment introducing the cd builtin

Delapouite:
Hi

Nothing problematic. I've stumbled upon this comment which is repeated for no reason for `cd`.
Probably the result of an extra copy-paste.


krobelus:
Thanks, it's nice to remove some lines :)

PR #6733, opened 2020-03-10
doc: Add note about opening help in web browser

daviebadger:
## Description

Hello, I was reading the Fish tutorial, tried myself commands by commands, and had a problem with the `help` command (my browser could not open the given help). Hopefully, I found a solution on StackOverflow after a while. However, I think it would be super to note it directly in the tutorial to save readers time with troubleshooting.

Fixes issue #

## TODOs:

### DIFF
--- a/doc_src/tutorial.rst
+++ b/doc_src/tutorial.rst
@@ -77,6 +77,24 @@ Getting Help
     set - handle shell variables
       Synopsis...
 
+.. note::
+
+    If the ``help`` command failed like::
+
+        > help set
+        gio: file:///usr/share/doc/fish/cmds/set.html: No application is registered as handling this file
+
+    then you need to register a web browser to handle these files::
+
+        > # Detect a MIME type
+        > xdg-mime query filetype /usr/share/doc/fish/cmds/set.html
+        application/x-extension-html
+        > # Detect your default browser (*.desktop files are mostly in /usr/share/applications/)
+        > xdg-settings get default-web-browser
+        firefox.desktop
+        > # Register the browser for the MIME type
+        > xdg-mime default firefox.desktop application/x-extension-html
+
 
 
 Syntax Highlighting


krobelus:
Thanks but I don't think telling the operating system how to open HTML files is in scope for fish. Most systems should already have a sane default for this, and if not, your system's documentation should state how to do it. Your example is also specific to Linux.

In [help.fish](https://github.com/fish-shell/fish-shell/blob/master/share/functions/help.fish#L72) we use `xdg-open` if available, many other programs (like file managers) do the same on Linux, and will fail if you have not configured it correctly. If you have installed `xdg-utils`, `xdg-open` should search all your desktop file for anything that can open `text/html` before failing. If you install a different provider for `xdg-open`, then you might need to do the manual steps you describe.

daviebadger:
Just for your information, `xdg-open` works fine for `text/html` files, but the help for the `set` command is **not** `text/html`, but `application/x-extension-html` and that is why it failed at first. Maybe it is only specific to my Linux environment.

krobelus:
Oh sorry, I somewhat missed the actual issue.
Let's find out why it's not detected as `text/html`, maybe we can fix that.

Can you share the output of the commands below? I get this:

```fish
$ file --mime /usr/share/doc/fish/cmds/set.html
/usr/share/doc/fish/cmds/set.html: text/html; charset=utf-8
$ xdg-mime query filetype /usr/share/doc/fish/cmds/set.html
application/xhtml+xml
$ echo $version
3.1.0
```

daviebadger:
The output:

```
$ file --mime /usr/share/doc/fish/cmds/set.html
/usr/share/doc/fish/cmds/set.html: text/html; charset=utf-8
$ xdg-mime query filetype /usr/share/doc/fish/cmds/set.html
application/x-extension-html
$ echo $version
3.1.0
$ file --version
file-5.37
magic file from /etc/magic:/usr/share/misc/magic
$ xdg-mime --version
xdg-mime 1.1.3
$ lsb_release --all
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 19.10
Release:	19.10
Codename:	eoan
```

krobelus:
> $ xdg-mime query filetype /usr/share/doc/fish/cmds/set.html
application/x-extension-html

This is strange, I don't know if it should be like this. On Ubuntu 18 here it gives `text/html`.

Running it with debug output shows that it is just executing `mimetype`:
```sh
$ XDG_UTILS_DEBUG_LEVEL=1 xdg-mime query
Running mimetype --brief --dereference "/usr/share/doc/fish/cmds/set.html"
text/html
```
`mimetype` is a perl script that has a debug flag:

```sh
$ mimetype --version
mimetype 0.28
$ mimetype -D --brief --dereference /usr/share/doc/fish/set.html
> Data dirs are: /home/johannes/.local/share, /usr/local/share, /usr/share, /var/lib/snapd/desktop
> Checking inode type
> Checking globs for basename 'set.html'
> Checking for extension '.html'
text/html
```

I assume the last command shows something different for you? It could be that something in your `~/.local/share/mime*` overrides the file extension check.

---

If this is a common occurrence, we probably want to add a workaround in
https://github.com/fish-shell/fish-shell/blob/master/share/functions/help.fish#L72, because typing `help` should absolutely work in all realistic cases.

daviebadger:
The `mimetype` output:

```
$ mimetype -D --brief --dereference /usr/share/doc/fish/set.html
> Data dirs are: /home/davie/.local/share, /usr/share/budgie-desktop, /usr/local/share, /usr/share, /var/lib/snapd/desktop, /var/lib/snapd/desktop
> Checking inode type
> Checking globs for basename 'set.html'
> Checking for extension '.html'
text/html
```

I also checked why it works as expected on Ubuntu 18.04 and not in 19.10 (who knows how on upcoming 20.04). There is a difference in versions:

* 18.04 has `xdg-utils=1.1.2` -> `text/html`
* 19.10 has `xdg-utils=1.1.3` -> `application/x-extension-html`

See https://launchpad.net/ubuntu/+source/xdg-utils.

krobelus:
Wait, so your `mimetype` on Ubuntu 19.10 prints `text/html`, but `xdg-mime query filetype` doesn't??
What is the output of `XDG_UTILS_DEBUG_LEVEL=1 xdg-mime query filetype /usr/share/doc/fish/cmds/set.html`? I get

```
Running mimetype --brief --dereference "/usr/share/doc/fish/cmds/set.html"
text/html
```

On Ubuntu 19.10 with xdg-utils 1.1.3, I also get `text/html` with both commands just like on 18, even if I create the file `~/.local/share/mime/types` with the same contents as yours.

krobelus:
Do you maybe have overrides for `*.html` in `~/.local/share/mime/packages`? (see https://wiki.archlinux.org/index.php/XDG_MIME_Applications#Shared_MIME_database for an example that overrides `*.foo`)

daviebadger:
I checked the link. Something overrode the mime types in the past:

```
$ cat ~/.local/share/mime/packages/user-extension-html.xml
<?xml version="1.0" encoding="UTF-8"?>
<mime-info xmlns="http://www.freedesktop.org/standards/shared-mime-info">
 <mime-type type="application/x-extension-html">
  <comment>html document</comment>
  <glob pattern="*.html"/>
 </mime-type>
</mime-info>
```
I will try to figure out myself what and how. Nevertheless, I very appreciate your help. Thank you.

Issue #4255, opened 2017-07-27
`commandline --selection` option is undocumented and only prints


faho:
As I found out when I checked the vi-bindings, `commandline` has a `-s` or `--selection` option to print the current selection. This has a number of problems:

- [ ] It's "--selection" when it should be "--current-selection" for consistency

- [ ] It only prints, combinations like `commandline -rs` don't work (at least that's what the code looks like to me)

- [ ] It's undocumented

Because the first is backwards-incompatible (even though it's unlikely to break since it's undocumented), we should probably only add this to 3.0. Unless we want to document it as-is for 2.7.0.

faho:
I've now changed it to "--current-selection" (keeping "-s" as the short version) and added a small note to the documentation.

That means the breaking part has been added for 3.0, the rest can wait until later (though I have most of that already).

faho:
This really isn't important enough to hold up a release.

PR #6738, opened 2020-03-12
feat: add missing 'current-selection' completion for commandline

Delapouite:
Hi

This existing option for `commandline` introduced in #4255 is documented but not autocompleted


krobelus:
Great catch, thanks!

PR #6747, opened 2020-03-13
cppcheck fixes

neheb:
Cppcheck fixes.

### DIFF
--- a/src/color.cpp
+++ b/src/color.cpp
@@ -170,7 +170,7 @@ static const named_color_t named_colors[] = {
 };
 
 wcstring_list_t rgb_color_t::named_color_names() {
-    size_t count = sizeof named_colors / sizeof *named_colors;
+    const size_t count = sizeof named_colors / sizeof *named_colors;
     wcstring_list_t result;
     result.reserve(1 + count);
     for (size_t i = 0; i < count; i++) {
--- a/src/history.cpp
+++ b/src/history.cpp
@@ -398,7 +398,8 @@ void history_impl_t::save_unless_disabled() {
     if (countdown_to_vacuum < 0) {
         unsigned int seed = static_cast<unsigned int>(time(nullptr));
         // Generate a number in the range [0, kVacuumFrequency).
-        countdown_to_vacuum = rand_r(&seed) / (RAND_MAX / kVacuumFrequency + 1);
+        srand(seed);
+        countdown_to_vacuum = rand() / (RAND_MAX / kVacuumFrequency + 1);
     }
 
     // Determine if we're going to vacuum.
--- a/src/builtin.cpp
+++ b/src/builtin.cpp
@@ -152,7 +152,7 @@ int parse_help_only_cmd_opts(struct help_only_cmd_opts_t &opts, int *optind, int
 ///    builtin or function name to get up help for
 ///
 /// Process and print help for the specified builtin or function.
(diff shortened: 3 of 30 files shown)


src/color.cpp
```diff
@@ -170,7 +170,7 @@ static const named_color_t named_colors[] = {
 };
 
 wcstring_list_t rgb_color_t::named_color_names() {
-    size_t count = sizeof named_colors / sizeof *named_colors;
+    const size_t count = sizeof named_colors / sizeof *named_colors;
```
krobelus: So variables are marked const when they are initialized with `sizeof`? Or when they are effectively const and initialized with a constant expression?
There are some cases that suggest the former, but maybe your tool didn't catch those:

[`size_t max = 1024 * 1024 * 64;`](https://github.com/fish-shell/fish-shell/blob/master/src/lru.h#L300) (in a header)
[`size_t expected_evictee = 3;`](https://github.com/fish-shell/fish-shell/blob/master/src/fish_tests.cpp#L5529) and others in this test file, which is compiled separately from fish
[`size_t read_amt = 64 * 1024;`](https://github.com/fish-shell/fish-shell/blob/master/src/env_universal_common.cpp#L1295) which is compiled only on some systems

neheb: The tool was treating this as a non constant expression for some reason.

src/history.cpp
```diff
@@ -398,7 +398,8 @@ void history_impl_t::save_unless_disabled() {
     if (countdown_to_vacuum < 0) {
         unsigned int seed = static_cast<unsigned int>(time(nullptr));
         // Generate a number in the range [0, kVacuumFrequency).
-        countdown_to_vacuum = rand_r(&seed) / (RAND_MAX / kVacuumFrequency + 1);
```
zanchey: If this doesn't need to be reentrant, it might be worth adding a comment explaining why not.

I struggle with these "marked as obsolete" features; they will never actually be removed, and I think it's reasonable to avoid writing new code with them, but there doesn't seem to be a great deal achieved by excising them from existing codebases.

neheb: According to the man page, all three functions are reentrant.

From musl:

```
static uint64_t seed;

void srand(unsigned s)
{
        seed = s-1;
}

int rand(void)
{
        seed = 6364136223846793005ULL*seed + 1;
        return seed>>33;
}

static unsigned temper(unsigned x)
{
        x ^= x>>11;
        x ^= x<<7 & 0x9D2C5680;
        x ^= x<<15 & 0xEFC60000;
        x ^= x>>18;
        return x;
}

int rand_r(unsigned *seed)
{
        return temper(*seed = *seed * 1103515245 + 12345)/2;
}
```

I don't see anything that would indicate that rand is not reentrant.

ridiculousfish: rand() is reentrant but this should be replaced with C++ random number generation

ridiculousfish:
Merged as fee08a87e9d656d967d1d8612b2ddc3d63749c32 except for the `static` changes. I think it's not useful to mark member functions as `static` without also adjusting the call sites. Thank you!

src/history.cpp
zanchey: Yeah, in glib the manpage says "The function rand() is not reentrant, since it uses hidden state that is modified on each call", but STL is much more sensible.

PR #6749, opened 2020-03-14
completions: add basic completions for 'ip netns'

sanga:
## Description

Add some rudimentary completions for `ip netns`. This is by no means complete but it's a small improvement. I read through the contributing guidelines and I think I've done everything I should. Does one add a changelog entry for improved completions?

### DIFF
--- a/share/completions/ip.fish
+++ b/share/completions/ip.fish
@@ -194,6 +194,12 @@ function __fish_ip_scope
         host "Address is only valid on this host"
 end
 
+function __fish_ip_netns_list
+    ip netns list | while read a b c
+        echo -- $a
+    end
+end
+
 function __fish_ip_types
     printf '%s\t%s\n' \
         macvtap "Virtual interface based on link layer address (MAC) and TAP." \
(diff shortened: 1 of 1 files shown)


share/completions/ip.fish
```diff
@@ -194,6 +194,12 @@ function __fish_ip_scope
         host "Address is only valid on this host"
 end
 
+function __fish_list_netnses
```
krobelus: I'd rather use a `__fish_ip` prefix to be consistent with other functions in this file.
Maybe `__fish_ip_netns_list`? I think it's a good idea to use subject-first naming for things in the global namespace.

share/completions/ip.fish
```diff
@@ -194,6 +194,12 @@ function __fish_ip_scope
...
+    ip netns list | while read a b c
+        echo $a
```
krobelus: It looks like it is allowed to create a network namespace with a name that starts with `-`. So let's use `echo --` or `printf`, just in case ;)

share/completions/ip.fish
sanga: Yeah that seems better for sure. I forgot that these all land in the global namespace

share/completions/ip.fish
sanga: Fair point. Fixing.

sanga:
Comments should be addressed. I noticed I also forgot to add `list`

krobelus:
Merged as 5cf24b773 with a typo fix, thank you!

PR #6750, opened 2020-03-14
chore: bump copyright year 2019 → 2020

Delapouite:
Hi

After greping for `2019` over the codebase, it seems to be the only dates that need update.


ridiculousfish:
Thank you!

PR #6751, opened 2020-03-14
Update Arch Linux package tool completions

ammgws:
New: completions for `downgrade`

Changed: added `--installed` option for printing Arch packages


faho:
Yeah that makes sense. Thanks!

PR #6752, opened 2020-03-14
doc: make <kbd> style more prominent

Delapouite:
Hello

I know there are ongoing discussions about building a new nice graphical design for the fish documentation.
But I think there are also minor improvements to the current style that could improve readability regardless of the final skin.

The first low hanging fruit I would like to tackle are the keyboard shortcuts (`<kbd>`). An image is worth 1000 words, so here's a before/after comparison:

before
image

after
image

This style is the one used on Wikipedia so it would feel familiar to a lot of people.

### DIFF
--- a/doc_src/_static/custom.css
+++ b/doc_src/_static/custom.css
@@ -1,1 +1,10 @@
 .sphinxsidebar ul.current > li.current { font-weight: bold }
+
+kbd {
+  background-color: #f9f9f9;
+  border: 1px solid #aaa;
+  border-radius: .2em;
+  box-shadow: 0.1em 0.1em 0.2em rgba(0,0,0,0.1);
+  color: #000;
+  padding: 0.1em 0.3em;
+}


exploide:
One difference to the Wikipedia style is that your keyboard shortcuts are contained within one single kbd element. "Alt + c" vs "Alt" + "c"

Since the element is styled to mimic the look of an actual keyboard key and there is no "Alt + c" key, I would personally prefer the style seen on Wikipedia.

charego:
Also the capitalization is inconsistent. FWIW it seems to be inconsistent [on Wikipedia](https://en.wikipedia.org/wiki/Table_of_keyboard_shortcuts) too:

shortcuts

Personally, I would prefer the characters to be capitalized, i.e. `Control`+`T` over `Control`+`t`.

Delapouite:
You're right but note that this patch only applies CSS without touching the underlying text. I guess I could add another commit to this PR fixing the separation and capitalization.

Delapouite:
Looks like the rst parser does not like when 2 kbd instructions are next to each others without space.

The following text from the FAQ page:

```rst
Additional default key bindings include :kbd:`Ctrl`+:kbd:`P` (up) and :kbd:`Ctrl` + :kbd:`N` (down)
```

render like:

image

Which means that we have to introduce an extra space on each side of the `+`.  @charego is it OK for you?

charego:
Interesting. While [sphinx](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-kbd) defines `:kbd:` as a sequence of keystrokes, it generates a `<kbd>` element. A brief search of [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd) and other sites shows that `<kbd>` is used for a single keystroke.

So I would think that extra spaces, if not ideal, is the right approach.
```
:kbd:`Ctrl` + :kbd:`N`
```

That being said, the CSS change alone looks very good.

krobelus:
Looks great, this definitely improves readability, maybe we can find a sphinx theme that includes good defaults for things like this in future.

It looks like we can escape the spaces to make sphinx render two `kbd` elements separated only by a `+`:

```rst
- :kbd:`Control+K`

- :kbd:`Control` + :kbd:`K`

- :kbd:`Control`\ +\ :kbd:`K`
```

renders as

image

faho:
This indeed does look much better.

Lemme merge this *now*, because as it is it's already a nice improvement. We can handle the separation later.

PR #6753, opened 2020-03-14
Fix screen optarg completions

ammgws:
Fixes this:

```
>screen -list
There are screens on:
    27624.pts-5.ammgws    (Detached)
    26048.fishysmell    (Detached)
# tabbed completion:
>screen -r26048.fishysmell
...
Error: Unknown option r26048.fishysmell
```


krobelus:
I think the same goes for h, s, S, t and T (even if they don't have argument completions)?

ammgws:
Yeah everything that takes an argument

krobelus:
Except for `-c` and `-e` for whatever reason. Merged as 0f681190b, thanks!

PR #6762, opened 2020-03-17
Re-adds `git restore` and `git switch` completions

bonflintstone:
These commands were already added in version 2.23.
See: ecacf346932c4bbee9f40871b2763595243f3c70

Original Author: Shun Sakai | sorairolake

## Description

This has already been part of fish a while ago, but for some reason seems to be missing in current versions. I readded it here from https://github.com/fish-shell/fish-shell/commit/ecacf346932c4bbee9f40871b2763595243f3c70

Absolutely no credit goes to myself :)

If I am missing some reason they were removed or something else please let me know!

Fixes issue #

## TODOs:


faho:
And they [are still in](https://github.com/fish-shell/fish-shell/blob/master/share/completions/git.fish#L1520-L1559), they've just been moved by 42c6b9c17697a3e5062121d307bd5aa952b02053

bonflintstone:
Turns out the version on aapt universe/shells is about a year old which explains why I dont have it locally. Thanks for the quick reply and sorry to bother you, should have checked that first!

Issue #5974, opened 2019-07-03
Add `string sub --end`


zx8:
Would be neat if `string sub --length` accepted negatives indexes, just like `string sub --start`, e.g. 

```fish
$ string sub --length -2 -- foobar
foob
```

AFAIK, the only way to achieve this at the moment is:
```fish
$ set s foobar
$ string sub --length (math (string length -- $s) - 2) -- $s
foob
```

(Also, feel like this option should be (re?)named `--end` to match `--start`)

faho:
That's an entirely different thing?

Like... with a length of 3, you'd go from a start of e.g. 2 to an end of 5.

Which is an okay thing to add, I just don't see why we should remove --length for it.

faho:
E.g. `string sub --start 2 --length 3 foobar` prints "oob" (a 3-long string starting with the second character).

`string sub --start 2 --end 3 foobar` should print either `oo` or just `o`. (Arguably it'd be better to include the end, like we include the start, so `oo`)

zx8:
Aha, misunderstood `--length`, got it!

PR #6765, opened 2020-03-18
Add `string sub --end`

gechr:
Closes https://github.com/fish-shell/fish-shell/issues/5974

Add a new `e/--end` option to allow specifying the end of the substring. Supports positive/negative indices, using Python's string slicing as a reference behaviour. Made `--length` mutually exclusive with `--end` as specifying both doesn't really make sense.

See docs/tests for examples.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ### Scripting improvements
 - Range limits in index range expansions like `$x[$start..$end]` may be omitted: `$start` and `$end` default to 1 and -1 (the last item) respectively.
+- `string sub` has a new `--end` option to specify the end index of a substring (#6765).
 
 ### Interactive improvements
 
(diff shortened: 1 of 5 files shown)


krobelus:
Looks good as far as I'm concerned. I wonder if it makes sense to support the combination of `--length` and `--end` (not all three), even though it's weird. It could save the user a tiny bit of arithmetic in very rare cases.

gechr:
@krobelus Happy to make `--length` and `--end` work together, but would you expect the output of:

```fish
$ string sub --end -1 --length 2 0123456789
```

to be `01` or `78`? That is, should it take characters from the left or right of the string if `--end` is specified?

krobelus:
I'd expect `78` here, but now I see that it is super ambiguous, so let's leave it off for now.
Merged, thanks!

PR #6769, opened 2020-03-19
Add vi 'y' bindings and some i / a support

joncinque:
## Description

Many people have mentioned wanting support for changing / yanking /
deleting between "" and '', meaning the commands `ci' ci" yi' yi" di' di"`,
so this adds that support in a generic, and thus potentially confusing way.
The concept is that we check for the character backwards and forwards
before making the selection.  Unfortunately, this will also work for *any*
character that isn't `w` or `W`, so `cia` could change everything between
two `a` characters.
Looking through the [bind documentation](https://fishshell.com/docs/current/cmds/bind.html)
and input handler at `src/input.cpp`,
this is the best possible solution I could come up with until
`forward-jump` and `backward-jump` can accept input in the call to `bind`,
and not just from stdin, meaning we can write a binding as:
```
bind di\" backward-jump-till \" and repeat-jump-reverse ...`
```
If that were done, then other commands such as `di)` to go between `()` would
be possible.
There are also some added `y` bindings not part of #6648.
Let me know if you need anything else.

## TODOs:

### DIFF
--- a/share/functions/fish_vi_key_bindings.fish
+++ b/share/functions/fish_vi_key_bindings.fish
@@ -153,6 +153,8 @@ function fish_vi_key_bindings --description 'vi-like key bindings for fish'
     bind -s --preset dT begin-selection backward-jump forward-char kill-selection end-selection
     bind -s --preset dh backward-char delete-char
     bind -s --preset dl delete-char
+    bind -s --preset di backward-jump-till and repeat-jump-reverse and begin-selection repeat-jump kill-selection end-selection
+    bind -s --preset da backward-jump and repeat-jump-reverse and begin-selection repeat-jump kill-selection end-selection
 
     bind -s --preset -m insert s delete-char repaint-mode
     bind -s --preset -m insert S kill-whole-line repaint-mode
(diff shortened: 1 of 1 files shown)


faho:
I'm not sure how useful this is - it seems like it really only works for `"` and `'`, which isn't *entirely* uncommon.

But on the other hand, I'm not entirely sure how to bring this into the binding system. How do we break this down?

Adding arguments to bind functions like you propose is one idea,  but that complicates the already weird bind DSL.

Can we find operations that this requires, and add *them* so that we can then implement these things in their terms?

I think we *might* be able to? "di)" to me appears to require the following:

- Find the next ")" - this is just forward-jump
- Enable selection - (well, really, start a new selection. If one already exists we'd have to "stash" it or something?)
- Find the corresponding "(" - this would be "%", and would be good to have anyway. Note that this requires cooperation from the parser.
- Shrink the selection by 1 on both ends (to exclude "()")
- Delete the selection (kill-selection)

And for "di(" it would require a *backward* jump first or a forward-jump with the argument switched around.

----

Which just leaves the question of what we do with this now. And personally, I'd assume that pressing `di"` and getting something is good even if `di)` doesn't work.

So I'm merging this. Thanks!

joncinque:
I think this should all be possible if `forward-jump` and its variants can accept input from the binding command, and not just from stdin.  Given the current setup of the system, I'm not sure how that'll be feasible though.  Anyway, thanks for merging this in!

PR #6776, opened 2020-03-19
feat: add missing 'save' completion for funced

Delapouite:
Hello

The `funced` `-s/save` option is mentioned in the doc and in the code, but the completion was missing.


krobelus:
Nice, I didn't know that option :+1:

PR #6777, opened 2020-03-20
docs: consistent key combinations

charego:
## Description

A few changes to the documentation so that key combinations look consistent throughout.

- keys are capitalized (prefer `S` over `s`)
- no spacing between keys in a combination, i.e. `Alt`+`S` not `Alt` + `S`
  - implementation tip from https://github.com/fish-shell/fish-shell/pull/6752#issuecomment-599238970, if there's a cleaner way let me know
- prefer arrow characters (`↑`, `↓`, `→`, `←`) over names (Up, Down, Right, Left)
  - not sure about this one, names might look better

EDIT: some handy ripgrep commands to search for keys and key combinations.
```
fish-shell/doc_src > rg -o ':kbd:`[^`]+`'
fish-shell/doc_src > rg -o ':kbd:`[^`]+`[\s+\\\]*:kbd:`[^`]+`'
```


doc_src/index.rst
```diff
@@ -1408,27 +1408,27 @@ Command mode
 
 Command mode is also known as normal mode.
 
-- :kbd:`h` moves the cursor left.
+- :kbd:`H` moves the cursor left.
```
faho: The problem with capitalizing letters is that it looks like you're supposed to press that which enters a "H" - Shift+h.

So I would capitalize *key names* like "Tab" instead of "tab", but not letters.

faho: Note that we already have some places where we speak of "Control + C". We should probably make that consistent.

charego: Your interpretation of <kbd>H</kbd> definitely makes sense, but shouldn't the `<kbd>` element represent the key as it appears on your keyboard (capitalized)?

The same argument can be made for leaving the arrow characters as is.

https://en.wikipedia.org/wiki/Keyboard_layout#Mechanical,_visual,_and_functional_layouts

In the end consistency is most important, so I'm fine switching to lowercase if there is consensus.

krobelus: I think capital letters look better and for the places where Shift is used (vi bindings) we add it explicitly, so I don't think there will be anyone confused by this. But lowercase is fine, too, it is more accurate.

faho: Yeah my fear is probably overblown. If we do get confusion we can change it later.

Let's merge this now.

zanchey:
Just discovered this made a lot of the Vi mode documentation wrong - whoops!

PR #6780, opened 2020-03-21
perl completions: remove unnecessary use of `begin`

ammgws:
Was added in 3c8c8a7559999a30d4f3de63b2c7122409b1fdb6 but we shouldn't need it anymore.


faho:
Thanks!

PR #6783, opened 2020-03-21
Add `string join (--left|--right)` options

gechr:
Add new `-l/--left` and `-r/--right` options to `string join`
to allow the separator to be added as a prefix or suffix
to each element in the list.

My personal use case for this is: I have a function with a list of globs
that I need to prefix with a `-g/--glob` flag, and I'm currently doing this:

```fish
set -l flags
set -l globs 'foo*' '*bar' 'ba*z'
for glob in $globs
    set -a flags -g $glob
end
```

...And I figured adding this functionality natively would be neater.
(and included `--right` for completeness).

See docs/tests for examples.


krobelus:
Sounds like [cartesian product expansion](https://fishshell.com/docs/current/index.html#cartesian-product), though this might be more discoverable in some cases.

krobelus:
If this can already be done by cartesian product expansion, would it make sense to instead add a reference to that in the documentation for `string join`?

gechr:
Sure, will close this.

PR #6784, opened 2020-03-21
Update `apropos` completions

ammgws:
Add missing completions, fix up some errors and/or borrow from `whatis` completions

### DIFF
--- a/share/completions/apropos.fish
+++ b/share/completions/apropos.fish
@@ -1,21 +1,29 @@
-
 function __fish_complete_apropos
     if test (commandline -ct)
         set str (commandline -ct)
-        apropos $str 2>/dev/null | sed -e "s/^\(.*$str\([^ ]*\).*\)\$/$str\2"\t"\1/"
+        switch $str
+            case '-**'
+
+            case '*'
+                apropos $str 2>/dev/null | string replace -rf -- "^(.*$str([^ ]*).*)" "$str\$2\t\$1"
+        end
     end
 end
 
 complete -xc apropos -a '(__fish_complete_apropos)' -d "whatis entry"
 
-complete -c apropos -s h -l help -d "Display help and exit"
+complete -f -c apropos -s '?' -l help -d "Display help and exit"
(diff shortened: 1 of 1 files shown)


share/completions/apropos.fish
```diff
@@ -1,21 +1,29 @@
...
 
 complete -xc apropos -a '(__fish_complete_apropos)' -d "whatis entry"
 
-complete -c apropos -s h -l help -d "Display help and exit"
+complete -f -c apropos -s '?' -l help -d "Display help and exit"
+complete -f -c apropos -l usage -d "Display short usage message"
 complete -f -c apropos -s d -l debug -d "Print debugging info"
 complete -f -c apropos -s v -l verbose -d "Verbose mode"
-complete -f -c apropos -s r -l regex -d "Keyword as regex"
+complete -f -c apropos -s r -l regex -d "Keyword as regex (enabled by default)"
```
faho: Just `(default)` would do.

share/completions/apropos.fish
```diff
@@ -1,21 +1,29 @@
-
 function __fish_complete_apropos
     if test (commandline -ct)
         set str (commandline -ct)
-        apropos $str 2>/dev/null | sed -e "s/^\(.*$str\([^ ]*\).*\)\$/$str\2"\t"\1/"
+        switch $str
```
faho: So wait, this is meant to stop it from passing arguments starting with `-` to apropos?

Wouldn't `apropos -- $str` work for that?

ammgws: To supress completions for text starting with "-" as it means the actual apropos option completions won't show:

https://user-images.githubusercontent.com/20397027/77248543-1eda2980-6c32-11ea-9a67-db33c0886376.png

faho: Technically I think it's possible for a manpage name to start with "-". But since I don't think I've ever seen one that does, this is probably alright.

faho:
Thanks!

PR #6786, opened 2020-03-21
help function: special case string commands

ammgws:
> @ammgws
> 23:11
> The help function currently only accepts max of one argument, however regarding the builtin string functions, I often find myself typing `help string replace` instead of `help string-replace`, due to the former being the way the command is actually used. Would a special case for `help string xxx` be welcome?
> @faho
> 23:14
> @ammgws Sure, sounds good!


faho:
Thanks!

PR #3055, opened 2016-05-23
fish_indent: add use_tabs option

onodera-punpun:
`-t` or `--use-tabs` use tabs instead of spaces for indentation

fairly self-explanatory

PR #6790, opened 2020-03-21
fish_indent: Make indentation configurable

gechr:
Introduces a new `-i/--indent` option, allowing the user to override the
default indentation level of 4. I personally prefer 2 spaces-per-indent
and want to enable format-on-save for my personal fish scripts, but am
unable to do so currently because `fish_indent` insists on 4 spaces.

I also renamed the short option for `--no-indent` from `-i` to `-I` as I
could only find one reference to it in the fish codebase, and I really
don't expect end users will be using this. Additionally, with
indentation being configurable now, `--no-indent` becomes equivalent to
`--indent=0`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ### Scripting improvements
 - Range limits in index range expansions like `$x[$start..$end]` may be omitted: `$start` and `$end` default to 1 and -1 (the last item) respectively.
+- `fish_indent` now accepts an `-i`/`--indent` parameter to specify custom levels of indentation. As a result, the short flag for `--no-indent` has been renamed from `-i` to the more fitting `-I` (#6790).
 
 ### Interactive improvements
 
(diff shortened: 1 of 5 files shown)


faho:
Again: No. Making this configurable *defeats the point*.

faho:
That response probably comes across as overly curt and perhaps aggressive, and I'm sorry, that wasn't my intention.

You deserve a longer explanation.

The ideal of fish_indent is to be like gofmt or perhaps black - an autoformatter you set up, that has buy-in across the community, that then defines *the* style for the language.

In doing so, it can remove squibbling over formatting from the equation entirely. Just let fish_indent do the formatting, and you're done.

Adding options for variants then works against all that, and only for a minor improvement in subjective aesthetics. That is very much not worth it.

And that is why I reject PRs that add style-variants to fish_indent.

(we have also previously rejected one to use tabs, see #3055)

gechr:
Understandable, thanks for elaborating.

I thought of `fish_indent` (due to its name more than anything) as more similar to [`shfmt`](https://github.com/mvdan/sh) than formatters as uncompromising as `gofmt` / `black`.

I'll focus future contributions on formatting improvements rather than additional options to tweak the style. 👍

ammgws:
With the latest changes to fish_indent I thought it would be nicer if it was called fish_fmt or something instead, but I guess it's too late now?

PR #6791, opened 2020-03-22
Allow finishing build on OS X <10.13.6

malthejorgensen:
## Description

Building on OS X versions prior to 10.13.6 fails at the very end when running `codesign`.
The `-options runtime`-argument isn't available on these earlier versions of the OS.

Simply running codesign without that argument on OS X <10.13.6 seems to produce a runnable binary with no security warnings. I'm on OS X 10.12.6, and but have not tested on setups than other my own, so I could have a lax security setting allowing this to work.
In any case the situation for people OS X <10.13.6 should be better as the build will at least finish with the.

Fixes issue https://github.com/fish-shell/fish-shell/issues/6792

This code change is most easily understood when ignoring whitespace (`git diff -w`: https://github.com/fish-shell/fish-shell/pull/6791/files?w=1)

Apple docs: https://developer.apple.com/documentation/xcode/notarizing_macos_software_before_distribution/resolving_common_notarization_issues#3087724

## TODOs:

   _I don't think we need to change the docs. This is more a incompatibility fix/bug than a new feature or change._
  _Unless, a specific setup for building on an older OS X setup becomes part of CI (I don't think it should), we can't write a test for this._

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
 
 ### For distributors and developers
 - fish source tarballs are now distributed using the XZ compression method (#5460).
+- Allow finishing builds on OS X <10.13.6 (previously builds would fail at the `codesign` step)
 
 ---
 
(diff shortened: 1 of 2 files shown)


ridiculousfish:
Thank you!

ridiculousfish:
3b38f35168f6ed6d35dfa4f9e87d64cd9db15876

ridiculousfish:
8a068ed9844c6aef66b7e23c638edf355b8e2db5

PR #6793, opened 2020-03-22
Missing flag descriptions in python script

JanczarKurek:
## Description

create_manpage_completions.py has some flags that are in help.
Added description for flags --cleanup-in and -z.


faho:
Thank you very much!

PR #6795, opened 2020-03-23
string/trim: Add VT to list of default chars

gechr:
In line with https://en.cppreference.com/w/cpp/string/byte/isspace


faho:
When was the last time you've seen a vertical tab?

I don't think I've *ever* seen one.

But since we already have \f in there, okay.

Issue #6775, opened 2020-03-19
Allow fish_update_completions to skip already generated files


JanczarKurek:
Hi

I was making some hooks to pacman to regenerate fish completions upon installing/updating packages. Regenerating all completions each time can be pointless, and in most cases we're interested in generating only few completions.

This can be done by getting this info from package manager and manually choose files to fish_update_completions. It is unfortunately distro-dependent and maybe needs some logic already existing in code (mapping manpage names into completion names for example).

My proposal is to implement this as an optional flag into script - for example by checking mod times of manpages and completions. I am willing to implement this change if accepted.

zanchey:
SGTM! One thing to think about is if the path is different (eg first completion generated from `/usr/share/man/man1/foo.1`, when regenerated it is picked up from `/usr/local/share/man/man1/foo.1` instead).

faho:
I would assume you'd get most of the bang for quite a bit less effort if you just passed the paths of all the changed manpages?

If I e.g. update `coreutils`, just passing all the files that touched in /usr/share/man/man1/ (figuring that path out left as an exercise to the reader), like /usr/share/man/man1/ls.1.gz, should allow us to skip most other manpages and allow most of the speed up.

JanczarKurek:
I agree this is possible to achieve outside python script, but e.g. that path figuring part is already done in script (and takes some lines to be as broad as possible).

JanczarKurek:
Ok, first of all - script always removes all existing completions from destination directory, and thats not a nice thing when it comes to regenerating. Imho cleanup in target directory should be at least non mandatory, especially that we already have --cleanup-in flag that can do the same thing. What do you think?

PR #6796, opened 2020-03-23
Allow not removing files from destination dir

JanczarKurek:
Signed-off-by: JanczarKnurek <jacek@beit.tech>

Have something to do with:  #6775

Generating completions always removes all files from destination dir.
Now it can be omitted by passing flag "--save".


faho:
I feel like "keep" is a better term here.

faho:
Merged as  f212aba17, renamed the flag to "--keep" with 6e95e1d79.

Thanks!

PR #6811, opened 2020-03-26
Avoid failing tests if a username starting with `haha` exists

soumya92:
## Description

Currently if the local machine has one or more users starting with `haha`, then the cd completion tests fail, because `~haha` will complete to those username(s).

To ensure that only the directory completions expected by the test are surfaced, switch the tests to use a less likely to collide prefix, `~absolutelynosuchus`

## TODOs:

- [n/a] Tests have been added for regressions fixed
- [n/a] User-visible changes noted in CHANGELOG.md

This is an infrastructure improvement, there is no user-visible impact.


krobelus:
That works, thanks!

Issue #6809, opened 2020-03-26
`jobs` ignores -q for invalid PIDs


soumya92:
**fish, version 3.0.2**

In a script, I want to see whether a previous job is still running. I saw a "--quiet" flag, so I thought the following would work:

```
start-the-job &
set -l pid (jobs -lp)
# ... stuff ...
if jobs -q $pid
  echo "Job is still running"
end
```

But if the job has finished, this still prints `jobs: No suitable job: 47362`.

In fact, if the PID does not point to a fish job, there's no difference in the behaviour of `jobs $badpid` and `jobs -q $badpid`.

Is it reasonable to make the `-q` flag suppress that output as well? I've looked at the code and it doesn't seem like a huge change, but I'm not sure if there are reasons to not make this change.

PR #6812, opened 2020-03-26
jobs: suppress "No suitable job" if -q is given

soumya92:
## Description

This allows code of the form `if jobs -q $some_pid` in scripts to check whether a previously started job is still running. Previously this would return the correct value, but also print an error message.

The invalid argument errors will still be printed.
Added test cases for both.

Fixes issue #6809

## TODOs:

    Sort of. The documentation already describes the expected behaviour:
    > `-q` or `--quiet` prints no output for evaluation of jobs by exit status only


faho:
Yeah, that'll do the trick. Thanks!

PR #6813, opened 2020-03-26
Fix typo: '\E' is not an escape sequence.

soumya92:


faho:
Huh, that's a nice catch!

PR #6820, opened 2020-03-27
Add a $status_generation variable to track when $status was changed

### LINKED ISSUE #6815: Detect $status being carried over from last command
**fish, version 3.1.0**

I did not expect this:

```
soumya@machine /h/soumya> false
soumya@machine /h/soumya [1]> set foo bar
soumya@machine /h/soumya [1]> sleep 2 &
Job 1, 'sleep 2 &' has ended
soumya@machine /h/soumya [1]>

soumya@machine /h/soumya> true
soumya@machine /h/soumya> set foo bar
soumya@machine /h/soumya> sleep 2 &
Job 1, 'sleep 2 &' has ended
soumya@machine /h/soumya> 
```

I understand why this is happening, since both `set` and `&` do not modify the status. And that behaviour is great for scripts.

But it was still surprising to me that the [1] carried over in the prompt. Would it be possible for $status to be cleared somehow in preexec, so that it's only set to a non-zero value if it was explicitly caused by the last command? Or introduce a variable that could indicate that the previous command had no status?

(I tried `set status 0` in `--on-event fish_preexec`, but as expected, that did not work)

soumya92:
## Description

Attempt to differentiate between explicitly setting $status and implicitly re-using a previous $status value through a `$status_generation` electric variable. Currently this only affects backgrounded jobs and variable assignments.

This variable will be automatically incremented after every interactive command, except `command &` or `set foo bar`.

This will allow tools like "undistract me" to correctly report the exit status of the last command, by ignoring `$status` in `fish_postexec` iff the value of `$status_generation` is the same in pre and post exec.

It could potentially be extended to support something like `return --previous` in fish script, if the feature is useful enough to warrant that complexity.

Fixes #6815 #6998

## TODOs:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -70,6 +70,7 @@ Interactive improvements
 -  Control-Z is now available for binding (#7152).
 -  ``fish_key_reader`` sets the exit status to 0 when used with ``--help`` or ``--version`` (#6964).
 -  ``fish_key_reader`` and ``fish_indent`` send output from ``--version`` to standard output, matching other fish binaries (#6964).
+-  A new variable ``$status_generation`` is incremented only when the previous command produces a status. This can be used, for example, to check whether a failure status is a holdover due to a background job, or actually produced by the last run command.
 
 
 New or improved bindings
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -21,8 +21,9 @@ Notable improvements and fixes
 -  A new ``fish_add_path`` helper function to add paths to $PATH without producing duplicates, to be used interactively or in ``config.fish`` (#6960).
 - ``fish_preexec`` and ``fish_postexec`` events are no longer triggered for empty commands.
 - The ``test`` builtin now better shows where an error occured (#6030).
-- builtins may now output before all data is read. For example, `string replace` no longer has to read all of stdin before it can begin to output.
+- builtins may now output before all data is read. For example, ``string replace`` no longer has to read all of stdin before it can begin to output.
 - A number of new debugging categories have been added, including ``config``, ``path``, ``reader`` and ``screen`` (#6511). See the output of ``fish --print-debug-categories`` for the full list.
+- ``set`` and backgrounded jobs no longer overwrite ``$pipestatus``.
 
 Syntax changes and new commands
 -------------------------------
--- a/share/functions/fish_prompt.fish
+++ b/share/functions/fish_prompt.fish
@@ -1,6 +1,8 @@
 # name: Classic + Vcs
 # author: Lily Ballard
 
+set __fish_prompt_status_generation $status_generation
+
 function fish_prompt --description 'Write out the prompt'
     set -l last_pipestatus $pipestatus
     set -l last_status $status
@@ -23,7 +25,12 @@ function fish_prompt --description 'Write out the prompt'
     end
 
     # Write pipestatus
-    set -l prompt_status (__fish_print_pipestatus $last_status " [" "]" "|" (set_color $fish_color_status) (set_color --bold $fish_color_status) $last_pipestatus)
+    set -l bold_flag '--bold'
+    if test $__fish_prompt_status_generation = $status_generation
+        set bold_flag
+    end
+    set __fish_prompt_status_generation $status_generation
+    set -l prompt_status (__fish_print_pipestatus $last_status " [" "]" "|" (set_color $fish_color_status) (set_color $bold_flag $fish_color_status) $last_pipestatus)
 
     echo -n -s (set_color $fish_color_user) "$USER" $normal @ (set_color $color_host) (prompt_hostname) $normal ' ' (set_color $color_cwd) (prompt_pwd) $normal (fish_vcs_prompt) $normal $prompt_status $suffix " "
 end
--- a/src/builtin_set.cpp
+++ b/src/builtin_set.cpp
@@ -805,7 +805,6 @@ static int builtin_set_set(const wchar_t *cmd, set_cmd_opts_t &opts, int argc, w
 
 /// The set builtin creates, updates, and erases (removes, deletes) variables.
 maybe_t<int> builtin_set(parser_t &parser, io_streams_t &streams, wchar_t **argv) {
-    const int incoming_exit_status = parser.get_last_status();
     wchar_t *cmd = argv[0];
     int argc = builtin_count_args(argv);
     set_cmd_opts_t opts;
@@ -838,6 +837,6 @@ maybe_t<int> builtin_set(parser_t &parser, io_streams_t &streams, wchar_t **argv
         retval = builtin_set_set(cmd, opts, argc, argv, parser, streams);
     }
 
-    if (retval == STATUS_CMD_OK && opts.preserve_failure_exit_status) retval = incoming_exit_status;
+    if (retval == STATUS_CMD_OK && opts.preserve_failure_exit_status) return none();
     return retval;
 }
--- a/src/builtin_set.cpp
+++ b/src/builtin_set.cpp
@@ -804,7 +804,7 @@ static int builtin_set_set(const wchar_t *cmd, set_cmd_opts_t &opts, int argc, w
 }
 
 /// The set builtin creates, updates, and erases (removes, deletes) variables.
-int builtin_set(parser_t &parser, io_streams_t &streams, wchar_t **argv) {
+maybe_t<int> builtin_set(parser_t &parser, io_streams_t &streams, wchar_t **argv) {
     const int incoming_exit_status = parser.get_last_status();
     wchar_t *cmd = argv[0];
     int argc = builtin_count_args(argv);
--- a/src/maybe.h
+++ b/src/maybe.h
@@ -3,6 +3,7 @@
 
 #include <cassert>
 #include <type_traits>
+#include <utility>
 
 namespace maybe_detail {
 // Template magic to make maybe_t<T> copyable iff T is copyable.
--- a/src/proc.cpp
+++ b/src/proc.cpp
@@ -999,6 +999,7 @@ void job_t::continue_job(parser_t &parser, bool in_foreground) {
         const auto &p = processes.back();
         if (p->status.normal_exited() || p->status.signal_exited()) {
             parser.set_last_statuses(get_statuses());
+            parser.libdata().status_count++;
         }
     }
 }
--- a/src/proc.cpp
+++ b/src/proc.cpp
@@ -167,17 +167,30 @@ bool job_t::signal(int signal) {
     return true;
 }
 
-statuses_t job_t::get_statuses() const {
+maybe_t<statuses_t> job_t::get_statuses() const {
     statuses_t st{};
+    bool has_status = false;
+    int laststatus = 0;
     st.pipestatus.reserve(processes.size());
     for (const auto &p : processes) {
         auto status = p->status;
+        if (status.is_empty()) {
+            // Corner case for if a variable assignment is part of a pipeline.
+            // e.g. `false | set foo bar | true` will push 1 in the second spot,
+            // for a complete pipestatus of `1 1 0`.
+            st.pipestatus.push_back(laststatus);
+            continue;
+        }
         if (status.signal_exited()) {
             st.kill_signal = status.signal_code();
         }
+        laststatus = status.status_value();
+        has_status = true;
         st.pipestatus.push_back(status.status_value());
     }
-    int laststatus = st.pipestatus.back();
+    if (!has_status) {
+        return none();
+    }
     st.status = flags().negate ? !laststatus : laststatus;
     return st;
 }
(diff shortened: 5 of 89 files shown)


ridiculousfish:
Apologies for delay in review.  I recognize the need here and I want to find a solution, but I think this particular approach is too complex. Let's continue discussion in #6815.

CHANGELOG.rst
```diff
@@ -32,6 +32,7 @@ Notable improvements and fixes
   for empty commands.
 - The ``test`` builtin now better shows where an error occured (#6030).
 - Add a helper function to know if the user is root (#7031).
+- `fish_postexec` has a second argument that's set to `1` when the `$status` is a holdover from the previous command (e.g. because the current command was backgrounded, or was a variable assignment), `0` otherwise.
```
krobelus, line 35: Unfortunately, since we switched this to `rst` from `md` we need two backticks instead of one.

doc_src/cmds/function.rst
```diff
@@ -56,7 +56,7 @@ By using one of the event handler switches, a function can be made to run automa
 
 - ``fish_posterror``, which is emitted right after executing a command with syntax errors. The commandline is passed as the first parameter.
 
-- ``fish_postexec``, which is emitted right after executing an interactive command. The commandline is passed as the first parameter. Not emitted if command is empty.
+- ``fish_postexec``, which is emitted right after executing an interactive command. The commandline is passed as the first parameter. The second parameter indicates whether the value of `$status` is a holdover (for example, backgrounded jobs and variable assignments do not affect `$status`). Not emitted if command is empty.
```
krobelus, line 59: Also here two double backticks around `$status`. We should have a linter for this - Sphinx happily renders `<cite>$status</cite>`.

src/proc.cpp
```diff
@@ -189,17 +189,27 @@ bool job_t::signal(int signal) {
...
-statuses_t job_t::get_statuses() const {
+maybe_t<statuses_t> job_t::get_statuses() const {
     statuses_t st{};
+    bool has_status = false;
+    int laststatus = 0;
     st.pipestatus.reserve(processes.size());
     for (const auto &p : processes) {
         auto status = p->status;
+        if (status.is_empty()) {
+            st.pipestatus.push_back(laststatus);
```
krobelus, line 200: So a process that doesn't produce a status simply uses the status of the previous process in the pipeline?
For example `false | set x` will produce `1 1` (I understand this example is horribly contrived :laughing:)

I wonder if we can get away with dropping it, making pipestatus a singleton `1` in this example.
The user-visible `$pipestatus` will never be completely empty because it is only updated if at least one process in the job sets a status. I tried this and it seems to work well.

CHANGELOG.rst
```diff
@@ -32,6 +32,7 @@ Notable improvements and fixes
   for empty commands.
 - The ``test`` builtin now better shows where an error occured (#6030).
 - Add a helper function to know if the user is root (#7031).
+- `fish_postexec` has a second argument that's set to `1` when the `$status` is a holdover from the previous command (e.g. because the current command was backgrounded), `0` otherwise.
```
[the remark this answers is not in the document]
soumya92: Done.

src/reader.cpp
```diff
@@ -2525,8 +2527,9 @@ static int read_i(parser_t &parser) {
             data->command_line_changed(&data->command_line);
             wcstring_list_t argv(1, command);
             event_fire_generic(parser, L"fish_preexec", &argv);
-            reader_run_command(parser, command);
+            auto eval_res = reader_run_command(parser, command);
             signal_clear_cancel();
+            argv.push_back(std::to_wstring(eval_res.no_status));
```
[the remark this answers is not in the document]
soumya92: Yeah, I like that idea. I'm not a huge fan of 1 and 0 either, and I forgot that we can now use variables as commands.

share/functions/fish_prompt.fish
```diff
@@ -23,7 +23,16 @@ function fish_prompt --description 'Write out the prompt'
...
+    if test $__fish_prompt_status_is_holdover -eq 1
+        set bold_flag
+    end
+    set -l prompt_status (__fish_print_pipestatus $last_status " [" "]" "|" (set_color $fish_color_status) (set_color $bold_flag $fish_color_status) $last_pipestatus)
 
     echo -n -s (set_color $fish_color_user) "$USER" $normal @ (set_color $color_host) (prompt_hostname) $normal ' ' (set_color $color_cwd) (prompt_pwd) $normal (fish_vcs_prompt) $normal $prompt_status $suffix " "
 end
+
+set __fish_prompt_status_is_holdover 0
+function __fish_prompt_dim_holdover_status --on-event fish_postexec -a cmd is_holdover
```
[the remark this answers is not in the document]
soumya92: I tried a few things, but I'm not happy with the results. Maybe you meant something else?

- Add another argument for status_color_when_holdover. This breaks compat
- Move the `--bold` portion to __fish_print_pipestatus: This prevents users from being able to print a non-bold pipestatus, and also does nothing for holdover statuses if status_color is already bold.

CHANGELOG.rst
soumya92: Done

doc_src/cmds/function.rst
soumya92: Done

src/proc.cpp
soumya92: The only reason I didn't do that was because I was worried that having pipestatus not match the number of `|` might be confusing. If you think that's not a big deal, I can make the change.

src/reader.cpp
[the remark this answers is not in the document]
soumya92: Whoops. Pushed to early. Working on it.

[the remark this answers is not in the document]
soumya92: Done

soumya92:
Any idea why switching int -> maybe_t<int> in the builtins is causing `undefined reference to '__atomic_store'` and `undefined reference to '__atomic_load'` on clang?

I don't know enough C++ to figure this one out. I can confirm that without that change clang configure and make works fine, (even though -latomic is not present in the build flags), and after that change it fails (-latomic is still not present in the build flags).

src/maybe.h
```diff
@@ -3,6 +3,7 @@
 
 #include <cassert>
 #include <type_traits>
+#include <utility>
```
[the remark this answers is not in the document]
krobelus: No, it doesn't look like we do that elsewhere.

src/proc.cpp
krobelus: Yeah, it's fine to leave it. I don't think anyone should ever rely on either behavior.
Maybe add a comment explaining in which case this branch is taken, because it is really quite an unusual scenario.

soumya92: Done

krobelus:
The build failure is mysterious.. I could reproduce it at first and now it builds fine with clang.

soumya92:
Added a change to cmake that I think will fix the build issue on clang

share/functions/fish_prompt.fish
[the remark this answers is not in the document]
krobelus: I see, let's keep it this way. The caller of `__fish_print_pipestatus` should maintain control over all colors.

> Add another argument for status_color_when_holdover. This breaks compat

In `5d135d555` I used a hack to maintain compatibility in `__fish_print_pipestatus`. In hindsight I realized we can do better. We can pass named parameters as a local exported (or global unexported) variable.
`(last_status=$last_status __fish_print_pipestatus ...)`

---

On some terminal the difference in colors is a bit subtle. I'd make the entire status grey if its a holdover.

On a side note, the status-color-when-holdover could additionally be a universal variable, initialized in `share/functions/__fish_config_interactive.fish` like we do for many other colors.

src/builtin_set.cpp
```diff
@@ -805,7 +805,6 @@ static int builtin_set_set(const wchar_t *cmd, set_cmd_opts_t &opts, int argc, w
 
 /// The set builtin creates, updates, and erases (removes, deletes) variables.
 maybe_t<int> builtin_set(parser_t &parser, io_streams_t &streams, wchar_t **argv) {
-    const int incoming_exit_status = parser.get_last_status();
     wchar_t *cmd = argv[0];
```
krobelus, line 808: I think this commit breaks the default git prompt, not sure why; it no longer recognises that I'm on `master`.

```sh
fish-shell (master)> HOME=$PWD ./build/fish
Type help for instructions on how to use fish
~ ((cfa92bba…))>
```

soumya92: Yeah... `if not set branch (command git symbolic-ref HEAD 2>/dev/null)`. Looks like the `not` is being ignored there somehow. I will fix that and add a test for it.

soumya92: Thanks for catching this! I've added logic in exec.cpp to handle negation when no status is returned, and added some test cases to cover this.

krobelus:
Alright, I think this is basically ready!

One more thing: the new parameter to `fish_postexec` could break people who are using `$argv` instead of `$argv[1]`.
A quick search on `grep.app` revealed just a couple:

https://github.com/gazorby/fish-abbreviation-tips/blob/b02b13b32db0aedd000e293167cba59e1235ac41/conf.d/abbr_tips.fish
https://github.com/qznc/dot/blob/662f4122a3d01cb088f097c88523edef51bfd65b/config/fish/config.fish
https://github.com/razzius/fish-functions/blob/af3b6b408c2e67254143b732a86f4bb162325298/config.fish
https://github.com/martinklepsch/dotfiles/blob/master/config/fish/functions/notify-me.fish

I guess we can just not care about this, since it's not that many.

However, thinking about this, couldn the interface be better if we expose the holdover status not as hook parameter, but instead - true to its name - in [the "status" builtin](https://fishshell.com/docs/current/cmds/status.html). I think something like `status is-holdover` would be sweet, it simplifies `fish_prompt` since we won't need the postexec hook anymore, and (unlike a [special variable](https://fishshell.com/docs/current/index.html#special-variables)) it is super easy to look up what it does with `help status` or `status -h`.

soumya92:
Yeah, good point regarding compat of this solution. I don't know if status is-holdover is simple to implement and explain, since it needs to use some reference point for checking the holdover state.

I've implemented @ridiculousfish's suggestion on the linked issue (to use a "generation" variable), which does have the downside of documentation as you mention, but probably works better and was much easier to implement. PTAL, thanks.

soumya92:
I tried clang-format, but it made no changes. Is there some configuration I need to import?

```
Rewrite 8f14bf700e9a486b3b97b889e65945521a34b034 (1/5) (0 seconds passed, remaining 0 predicted)    clang-format did not modify any files
Rewrite 3014b94dcddd681219593ffcac59fd8863783c55 (2/5) (1 seconds passed, remaining 1 predicted)    clang-format did not modify any files
Rewrite e61d66669aa5d04903463590678c23795a686c59 (2/5) (1 seconds passed, remaining 1 predicted)    clang-format did not modify any files
Rewrite 4047391a680c7e13d4bc66ada2567e79b88d1078 (2/5) (1 seconds passed, remaining 1 predicted)    no modified files to format
Rewrite 42ceecba55fccb189d5ab9aab9aaed57f59cefee (5/5) (3 seconds passed, remaining 0 predicted)    no modified files to format

WARNING: Ref 'refs/heads/explicit-status-tracking' is unchanged
```

krobelus:
Oh maybe my clang-format 10 is newer than yours, don't worry about it.

krobelus:
Yeah, it looks like `status is-holdover` would require more explanation, so directly exposing the generation count seems like the simplest solution after all.

We could expose a `status status-generation` instead of `$status_generation`, like we have `status current-command` as a replacement for the old-style `$_`. The idea is to not clutter the variables with information most users don't need. Similar for `$pipestatus`, but that's already out there.
Then again, `$status_generation` fits right next to `$status`. This solution is also fine by me. Maybe we can have another opinion, @ridiculousfish?

Empty commands don't cause a "holdover" status anymore as they don't fire `fish_postexec` since #7085. This feels a bit strange but that's very minor.

share/functions/fish_prompt.fish
```diff
@@ -23,7 +23,26 @@ function fish_prompt --description 'Write out the prompt'
...
+
+set __fish_prompt_status_generation $status_generation
+set __fish_prompt_status_is_holdover false
+
+function __fish_prompt_save_generation --on-event fish_preexec
+    set __fish_prompt_status_generation $status_generation
+end
+
+function __fish_prompt_dim_holdover_status --on-event fish_postexec
+    if test $__fish_prompt_status_generation = $status_generation
```
krobelus, line 43: nit: (this hook is no longer necessary, we can do the comparison in `fish_prompt`)

soumya92: Good call. Done. I wasn't too happy with all the underscores anyway

soumya92:
> Empty commands don't cause a "holdover" status anymore as they don't fire fish_postexec since #7085. This feels a bit strange but that's very minor.

With your suggestion of checking generation in fish_prompt, now they do :)

ridiculousfish:
I think this is awesome!! I love the subtle tweak to the default prompt. Also really well factored, easily reviewable change set.

`$status_generation` is going to be a "low level" feature, not something users encounter every day, so it's OK if it's a little un-ergonomic. I think we can improve the name, but I don't want name bikeshedding to block merge, so I'll merge now.

Big thanks to @soumya92 for this contribution!

soumya92:
Thanks! I'm glad I was able to add this feature, since I will use it a lot :)

PR #6824, opened 2020-03-28
Add `fish_kill_signal` to track the signal that terminated the last command

### LINKED ISSUE #6822: Distinction between Ctrl-C and exit 130 is gone
**fish, version 3.1.0-442-gdee2adf03**

Prior to the fix for `fish_postexec` not being called on interrupt, I could set a flag in `fish_preexec` and clear it in `fish_postexec`. If I made it to fish_prompt with the flag set, I would know that the previous command was cancelled (rather than exit 130).

Now, with `fish_postexec` (correctly) being called even after a Ctrl-C, there's no way to distinguish something that exited with 130 on its own from something that was terminated by the user pressing Ctrl-C.

It would be nice if there was some way to do that. Would it be possible to add an event that triggers on Ctrl-C? (`--on-signal INT` doesn't work because the signal is sent to the foreground process, not fish).

I would have called such an event `fish_cancel`, but that name is already in use. `fish_exec_cancel` maybe?

soumya92:
## Description

Adds a new variable `$fish_kill_signal`, which stores the signal that terminated the last interactive command (e.g. 2 for SIGINT). Set to 0 if the command exited normally.

Fixes #6822

## TODOs:

### DIFF
--- a/src/env.h
+++ b/src/env.h
@@ -63,6 +63,10 @@ struct statuses_t {
     /// Status of the last job to exit.
     int status{0};
 
+    /// Signal from the most recent process in the last job that was terminated by a signal.
+    /// 0 if all processes exited normally.
+    int kill_signal{0};
+
     /// Pipestatus value.
     std::vector<int> pipestatus{};
 
--- a/tests/signals.expect
+++ b/tests/signals.expect
@@ -12,6 +12,22 @@ send_line read
 expect -re "\\r\\n?read> $"
 exec -- kill -INT $pid
 expect "fish_postexec spotted"
+expect_prompt
+
+# Verify that the fish_kill_signal is set.
+send_line "functions -e postexec; function postexec --on-event fish_postexec; echo fish_kill_signal \$fish_kill_signal; end"
+expect_prompt
+send_line "sleep 5"
+sleep 0.100
+exec -- pkill -INT sleep -P $pid
+expect "fish_kill_signal 2"
+expect_prompt
+
+send_line "sleep 5"
+sleep 0.100
+exec -- pkill -TERM sleep -P $pid
+expect "fish_kill_signal 15"
+expect_prompt
 
 # Verify that sending SIGHUP to the shell, such as will happen when the tty is
 # closed by the terminal, terminates the shell and the foreground command and
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Notable improvements and fixes
 - `fish --no-execute` will no longer complain about unknown commands or non-matching wildcards, as these could be defined differently at runtime (especially for functions). #977
 - `jobs --quiet PID` will no longer print 'no suitable job' if the job for PID does not exist (e.g. because it has finished). #6809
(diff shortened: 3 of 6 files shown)


krobelus:
Looks good.

I'm curious where something to the effect of `exit 130` is used. I think it happens mostly when a child process (for example another shell) exits after interrupting a job.

faho:
Since we already have cancellation_signal, it seems much easier to just expose that, e.g. via a variable, `fish_kill_signal` or so?

soumya92:
Yeah, that sounds better. I'll update this tomorrow to use an electric var.

For my purpose I could then check that var in the postexec handler.

soumya92:
Done. I ended up using the proc_status since that includes other signals as well which could be useful down the line.

Since cancellation sets the status to sigint already, everything works as expected.

tests/signals.expect
```diff
@@ -12,6 +12,22 @@ send_line read
...
 exec -- kill -INT $pid
 expect "fish_postexec spotted"
+expect_prompt
+
+# Verify that the fish_kill_signal is set.
+send_line "functions -e postexec; function postexec --on-event fish_postexec; echo fish_kill_signal \$fish_kill_signal; end"
+expect_prompt
+send_line "sleep 5"
+sleep 0.100
+exec -- killall -INT sleep
```
krobelus: Let's do a `pkill -INT sleep -P $pid` to not kill unrelated processes.

tests/signals.expect
```diff
@@ -12,6 +12,22 @@ send_line read
...
+expect "fish_kill_signal 2"
+expect_prompt
+
+send_line "sleep 5"
+sleep 0.100
+exec -- killall -TERM sleep
```
krobelus: Same here.

src/env.h
```diff
@@ -63,6 +63,10 @@ struct statuses_t {
     /// Status of the last job to exit.
     int status{0};
 
+    /// Signal from the most recent job that was terminated by a signal.
+    /// 0 if all jobs exited normally.
```
krobelus: I think this should be

```
Signal from the most recent process in the last job that was terminated by a signal.
0 if all processes exited normally.
```

tests/signals.expect
soumya92: Done.

src/env.h
soumya92: Yep, makes sense. Done.

krobelus:
Merged as 61a9cdaa7491a85342651d35f80afc8db2f86eee, thank you!

soumya92:
Happy to contribute :)

zanchey:
Is this for the last job or the last command?

krobelus:
From the last foreground job it's the rightmost command in the pipeline that was terminated by a signal.
If I do `sleep 100 | true` and `pkill sleep`, then it is the kill signal of `sleep`, not `true`.
So "job" might be a better wording than "command" for the docs/changelog?

soumya92:
Yep, job makes more sense and is consistent with $status. I'll send a PR to update that shortly

Issue #6537, opened 2020-01-25
Drop python2 support after 3.1


faho:
As we all probably know, python 2 is EOL, so it might be time to drop support for it.

We currently use python in a few places:

- webconfig
- manpage completion generation
- the docs (sphinx + configuration + our fish_indent_lexer)
- littlecheck
- some completion scripts (npm, bower, yarn, composer - anything that needs json)

All of which are optional and should fail cleanly if python isn't installed.

What I am proposing is that we stop explicitly checking and supporting python 2, and we start recommending python 3 (we already use it over 2 if available), after 3.1 is released. Nothing too rash, if an issue pops up we can still fix it, and if it's just cosmetics we should still prefer allowing python 2 to work.

Or we could take a more extreme stance and improve our code actively with things that don't work in py2, or only do that for some things.

zanchey:
To inform the decision, on our current platforms that I would consider "tier 1":

* RHEL/CentOS 6 does not easily have Python 3 installable - this platform goes EOL in November 30 2020
* macOS does not ship with Python 3 anywhere, as far as I can tell - and it sounds like all Python versions will go in the future (https://developer.apple.com/documentation/macos_release_notes/macos_catalina_10_15_release_notes) 
* RHEL/CentOS 7 as of 7.7 has Python 3.6 installable

Other platforms are fine on cursory inspection, with a general minimum of Python 3.5 available.

geoff-nixon:
> macOS does not ship with Python 3 anywhere, as far as I can tell - and it sounds like all Python versions will go in the future

The OS proper may not, but Xcode 11 ships with a full Python 3.7 at `/Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework`.

floam:
Yes - with Xcode you get python3 now on macOS.

geoff-nixon:
> Yes - with Xcode you get python3 now on macOS.

Yes, and double-checking now, it's included in the "Command Line Tools" package as well: there's even `python3` (etc.) shims at /usr/bin/python3 which call `xcrun`. So if a user doesn't have Xcode or the CLI tools installed, the first time the binary is called, it walks you through a painless download/install. So I'd definitely say macOS is good to go with Python 3.

mqudsi:
We call python in the background at first startup, the `xcrun` hack is insufficient.

zanchey:
[...]

PR #6825, opened 2020-03-28
New interactive test framework based on pexpect

ridiculousfish:
This adds a new interactive test framework, intended to eventually supplant the expect-based interactive tests.

Recall that "interactive tests" simulate a tty. They're for testing key bindings, signal handling, and other UI stuff. Existing fish interactive tests are based on [expect](https://core.tcl-lang.org/expect/index). Despite heroics from @zanchey and others, expect is still hard to work with. TCL is an idiosyncratic language. It uses dollar-sign variables, which collides with fish. The diff-based framework makes authoring and debugging tests tedious.

I propose incrementally migrating expect based tests to a framework based on [pexpect](https://pypi.org/project/pexpect/), which is a tty emulator written in Python. This allows us to write tests in Python instead of TCL, and also provides better authoring and debugging niceties.

## Design

pexpect tests will use the same vocabulary as expect-based tests:

- `send()` or `sendline()` to send text to fish.
- `expect_str()` and `expect_re()` to match against fish's output. If no matching output is produced before a timeout, this fails the test.
- `expect_prompt()` optionally matches a RE, then a prompt.

See [bind.py](https://github.com/ridiculousfish/fish-shell/blob/pexpect/tests/pexpects/bind.py) for an example (compare to [bind.expect](https://github.com/fish-shell/fish-shell/blob/1c880e58881ab5c5e9217857496a688addfb5f67/tests/bind.expect)).

## Failures

A test failure looks like this:

Screen Shot 2020-03-28 at 2 37 25 PM

this shows the line number `bind.py:188` where the failure occurred.

Note the table of "messages" which is text that was successfully sent or matched:

    RECV 3193.86 (bind.py:177): \r\ndefault\r\ninsert\r\npaste\r\nreplace\r\nreplace_one\r\nvisual\r\n
    RECV 3198.16 (bind.py:177): \r[I] prompt 19>
    SENT 3198.57 (bind.py:181): set -g fish_key_bindings fish_default_key_bindings\n
    RECV 3225.20 (bind.py:182): \rprompt 20>
    SENT 3225.66 (bind.py:185): echo fish_escape_delay_cms=$fish_escape_delay_ms\n

This shows us the direction, timestamp (ms since start), and the location in the test file which initiated the send or receive.

## Dependency Changes

Tests will run under Python 3 only. Python 2 support is not worth the hassle. Note this is a test-time only dependency; we'll continue to support python2 for user-facing scripts.

Beyond that the `pexpect` module must be installed: `pip3 install pexpect`. Note that pexpect uses the pty module, which is [untested except on Linux](https://docs.python.org/3/library/pty.html) - this is I think the biggest risk. In practice it appears to work fine on Macs, and course our `expect` tests have historically been flaky on non Linux/Mac platforms.


tests/interactive.fish
```diff
@@ -22,11 +22,13 @@ cd (dirname (status -f))
 set -gx TERM xterm
 set -e ITERM_PROFILE
 
-# Test files specified on commandline, or all *.expect files
+# Test files specified on commandline, or all *.expect files.
 if set -q argv[1]
-    set files_to_test $argv.expect
+    set expect_files_to_test $argv.expect*
```
krobelus: This will break because it matches legacy `*.expect.err` and `*.expect.out` paths.

I think we should drop this argument handling anyway because it should made be easier to run individual tests in the same way the build system would run them.
Merely running `fish interactive.fish bind` is usually wrong, or I just don't understand how to use it. It does not know about the build system, so it does not copy changed tests, `*.err` or `*.out` files, and it does not rebuild fish. Not sure in which scenarios this would be desired.

What I usually do is edit `interactive.fish` to hardcode `files_to_test`, and then run `make test_interactive`.

Maybe we can pass it as env var instead:
`INTERACTIVE_TESTS=./tests/pexpects/*.py ninja test_interactive`

faho:
I love how this uses one file per test. I feel that's the single biggest improvement in littlecheck over our previous system. Switching files got old quick, especially between ".err" and ".out" files.

I also like how this uses python, which we also use elsewhere and which I know and like, over tcl which I don't know and only ever used for these tests.

The dependency seems acceptable, since [it seems to be widely available](https://repology.org/project/python:pexpect/versions), especially via pip. I would love confirmation that it works everywhere (e.g. the BSDs, OpenIndiana/Illumos, cygwin?), but since it's just a *testing* dependency and we don't currently have CI on any of these I would be okay without.

-----

What I don't really love is the output. There is a lot of information here, and it's quite dense, especially considering there's escaped characters in here.

There's a few suggestions I'd make:

- Add some color. This output is all yellow, which appears to be your normal text color. It should distinguish between out/input and "meta" text like "RECV"

- Ditch the "RECV". That's super jargony. Make it "Output"

- Separate the "Buffer" from the rest. It's not super useful in this case, but I think I can see how it can sometimes be useful. Maybe just add an empty line between these?

Also it's currently a bit weird how you have to read top to bottom and then jump to the top again. I.e. you follow the messages, and then the thing that *happens* is the TIMEOUT, mentioned *before* the messages. Maybe add a TIMEOUT message to the "table"? Or maybe not, since that's the important thing that happened? But on the other hand the "unmatched" message is now printed in two locations already.

------

>Tests will run under Python 3 only. Python 2 support is not worth the hassle. Note this is a test-time only dependency; we'll continue to support python2 for user-facing scripts.

Supporting py3 only is acceptable. See also #6537.

build_tools/pexpect_helper.py
```diff
@@ -0,0 +1,261 @@
...
+    def sent(text, when):
+        """ Return a SEND message with the given text. """
+        return Message(Message.DIR_SEND, text, when)
+
+    @staticmethod
+    def received(text, when):
+        """ Return a RECV message with the given text. """
+        return Message(Message.DIR_RECV, text, when)
+
+    def formatted(self):
```
faho: I'm not entirely up on idiomatic python. Should this not be `__str__`? Then you could remove the explicit call later.

faho:
A rough sketch of what I think the output should look like:

Screenshot_20200329_170728

I don't want to overdo it, but I feel coloring in the important bits ("this was the failure, this kind of thing happened, this is the interaction") really aids readibility.

faho:
Also it is entirely too ready to call everything a "TIMEOUT" or is that on purpose?

When I e.g. change a "send" but not the corresponding "except", should it not complain that it's gotten the *wrong* thing?

faho:
Okay, if I'm reading this correctly pexpect ignores any non-matching output.

I'm not a fan of that. It means you can just add more output and the test won't fail. In certain situations our prompt counter could save us, tho.

krobelus:
I love this! it really takes away most of the pain from writing these tests for me, because it's much easier to learn and debug with a bit of python experience.

> Okay, if I'm reading this correctly pexpect ignores any non-matching output.

Yep, this makes it easy to ignore extra redraws; for example:

```python
control_a = '\x01'
sp.send("XYZ" + control_a + "echo ")
```

Here fish has to redraw the line once for each character of `echo `. I'm not sure how `expect` behaves here.

faho:
>Yep, this makes it easy to ignore extra redraws; for example:

That's neat, but what about unintentional output?

If you cause another prompt to appear we'll still catch it with the prompt counting, and if you change a *line* that's bracketed with `\r` or similar you'll break it, but e.g. change

```python
sp.send("echo mno pqr")
```

to

```python
sp.send("echo AAAAAAA; echo mno pqr")
```

or e.g. unintentionally change the output of `bind`, this won't catch it! The expect tests would.

build_tools/pexpect_helper.py
```diff
@@ -0,0 +1,261 @@
...
+
+    def sendline(self, s):
+        """ Cover over pexpect.spawn.sendline().
+            Send the given string + linesep to the tty, returning the number of bytes written.
+        """
+        return self.send(s + os.linesep)
+
+    def expect_re(self, pat, pat_desc=None, unmatched=None, **kwargs):
+        """ Cover over pexpect.spawn.expect().
+            Look through the "new" output of self.spawn until the given pattern is matched.
```
krobelus: Can we add to the docstring that this consumes the output until after the match? Kind of obvious, but useful to someone new to expect.

build_tools/pexpect_helper.py
```diff
@@ -0,0 +1,261 @@
...
+        if isinstance(err, pexpect.EOF):
+            msg = "EOF"
+        elif isinstance(err, pexpect.TIMEOUT):
+            msg = "TIMEOUT"
+        else:
+            msg = "UNKNOWN"
+        filename, lineno, code_context = get_callsite()
+        print("{RED}Failed to match:{NORMAL} {pat}".format(pat=escape(pat), **colors))
+        print(
+            "{msg} from {filename}:{lineno}: {code}".format(
```
krobelus: This message could follow an established error format, so editors can jump to the failing line.
`"{filename}:{lineno}: {msg} in {code}"` should work at least for vim.

That'd also need the full relative path, for example:
`tests/pexpects/history_search.py:59: TIMEOUT in sp.expect_str("5 AAC")`

krobelus:
I see, so with `expect`, the very next line has to match? I assume with lines separated by either of `\n` or `\r`.
It seems reasonable to keep that behavior, similar to `littlecheck`.

build_tools/pexpect_helper.py
faho: So I've done some work on this, and I've found it nicer to remove the formatting responsibility from this class, so you can do relative timestamps.

zanchey:
I really can't take any of the blame, er, kudos for the expect-based framework - that was @lilyball's hard work and @faho has done lots to keep it running. I agree that it is time to move to more modern tooling.

pexpect is packaged in just about every platform we care about, although sometimes the versions are quite old.

faho:
>Okay, if I'm reading this correctly pexpect ignores any non-matching output.

So I've looked some more into this, and I have no idea how expect does it.

You can technically use `spawn.readline()` to read a line or `spawn.read(int)` to read some characters (with timeout exceptions if you don't get enough and such). But always reading back all input becomes boring quickly, so you'd add reading it back to `send`. Only... sometimes you don't get all the input back. And that isn't just a rare case, it's what happens for any non-inserted character, including execute, so you'd basically have to adjust every `send` again, or figure out some way to figure out what fish inserts (and now send any escape sequence).

So: Yeah, it's not all that important. Let's just use it without.

zanchey:
Expect also just drops unmatched output until the timeout arrives - that's one of its strengths, arguably.

.travis.yml
```diff
@@ -14,6 +14,10 @@ matrix:
             - libncurses5-dev
             - libpcre2-dev
             - python
+            - python3
+            - python3-pip
```
zanchey, line 18: Just make this `python3-pexpect`, and drop the `pip` lines on everything except macOS - the system packages install faster and are fine (or I can do it after merge).

.travis.yml
```diff
@@ -14,6 +14,10 @@ matrix:
             - libncurses5-dev
             - libpcre2-dev
             - python
```
zanchey, line 16: Drop `python` alone here if depending on Python 3, though I think they'll both be installed anyway.

tests/interactive.fish
```diff
@@ -39,7 +41,7 @@ if not type -q expect
     exit 0
```
zanchey, line 41: There's a block here which exits with success if `expect` isn't installed; possibly the same should be done with `python3 -c 'import pexpect'`.

build_tools/pexpect_helper.py
ridiculousfish: Sure, good suggestion!

build_tools/pexpect_helper.py
ridiculousfish: Nice, I hadn't even thought of that.

.travis.yml
ridiculousfish: Awesome, I'll do it

ridiculousfish:
Thank you all for awesome review and feedback. @faho's output suggestions are a huge improvement, I will certainly implement those before merging.

faho:
>@faho's output suggestions are a huge improvement, I will certainly implement those before merging.

https://github.com/faho/fish-shell/tree/pexpect-output

tests/interactive.fish
ridiculousfish: done

tests/interactive.fish
ridiculousfish: Fixed the globbing; I will leave making it easier to run specific tests for another day.

ridiculousfish:
Merged as 8a27873598e6c71ce615d0a86f84c49072b9ae3e

@zanchey on Travis, I tried your suggestion of switching from `pip3 install pexpect` to `python3-pexpect`; [see this build](https://travis-ci.org/github/ridiculousfish/fish-shell/builds/695778306). But the .deb was very old (`python3-pexpect_4.0.1-1_all.deb`) and I wish to use a more recent release. So I merged this with the pip3, but feel free to modify this of course.

@faho I think I incorporated your suggestions; of course feel free to modify it further as you please. You have a great eye for this stuff.

Screen Shot 2020-06-07 at 3 27 37 PM

PR #6835, opened 2020-03-30
feat(completion): add missing -P option to read command

Delapouite:
Hello

This PR adds `-P/--prompt-str` completion to `read`. This option is already documented so nothing to change in the doc.


krobelus:
Thanks :)

PR #6837, opened 2020-03-30
doc: add kbd markup to Tab Completion section

Delapouite:
Hi

This PR adds more `kbd` markup.

**Before**
image

**After**
image


krobelus:
Good catch, thanks!

PR #6845, opened 2020-03-31
Added openssl completions

exploide:
## Description

More than one time, I wished fish had autocompletions for `openssl`. But `openssl` with all its dozens  of subcommands and hundreds of options is so complex, I thought it would require an insane amount of work.

Now, actually that's the shortest completion file I've ever written. :laughing: It turns out openssl is pretty good in completing itself with subcommands and options.

I think it's far from being perfect since there is no completion for further arguments and there also is no description text, but it is still an improvement to the current non-completion.

Maybe it is a bit hacky. If you know whether this could be done in a better way, please let me know.

## TODOs:


exploide:
Just force pushed an even shorter version avoiding one of the helper functions.

krobelus:
Neat, merged as 67eaefeb821ccfa6ea2cd8916830e78ac1efe1c8

PR #6846, opened 2020-03-31
GitHub actions

faho:
## Description

This enables running the tests via github actions on Ubuntu and macOS (their "latest" versions, respectively).

It's a potential replacement for Travis, which has caused us some grief.

Note that this disables some of the invocation tests (the `fish -i -c status` and such ones) on Github Actions because that doesn't start a terminal. The expect tests are unimpacted as they start their own pty.

Also note that the display can render color escape sequences, but since it doesn't provide a terminal most tests won't print colors (i.e. you won't see the nice green "OK" in the browser).


faho:
For some reason this doesn't run the actions right now. If you want to see what it looks like, check out [the PR I made for testing](https://github.com/faho/fish-shell/pull/1/checks).

zanchey:
I suspect it doesn't run because the configuration hasn't been enabled on master until after this is merged.

I think if the invocation tests won't run, this might be a non-starter for now - it could be worth checking with GitHub whether they are going to add that, though.

faho:
>I think if the invocation tests won't run

*Some* of the invocation tests - https://github.com/fish-shell/fish-shell/pull/6846/files#diff-e72dd6c1a42d9d2f897a52e7d77f0eddR15-R18

I.e. the ones that check that `status` prints the correct thing inside `fish -i` and `fish -li`. Those two.

I think github can take the hit of us enabling this, even if we don't disable Travis *for now*.

faho:
Okay, so let's see what happens when I merge this.

PR #6851, opened 2020-04-03
Update docs for `fish_kill_signal`, use "job" rather than "command"

soumya92:
From https://github.com/fish-shell/fish-shell/pull/6824#issuecomment-608531710


krobelus:
Great, thanks!

PR #6853, opened 2020-04-04
doc: add section about directory history / stack

Delapouite:
Hi

This PR adds a small section to the documentation dedicated to the notion of "Navigating directories" which was strangely missing.

What was particularly confusing for me as a new comer to fish was the not so obvious distinction between the directory **history** and the directory **stack** concepts. They are pretty similar concepts but nevertheless manipulate a different data structure under the hood.

Therefore, I tried to make the separation more clear between the 2 families of commands:
- history: `dirh`, `cdh`, `prevd` and `nextd`
- stack: `dirs`, `pushd` and `popd`

After adding this section, I did a little cleanup in the "See Also" section of the aforementioned commands.

Here's what it looks like when rendered:

image

### DIFF
--- a/doc_src/index.rst
+++ b/doc_src/index.rst
@@ -1472,10 +1472,25 @@ In addition, when pasting inside single quotes, pasted single quotes and backsla
 
 .. [#] These rely on external tools. Currently xsel, xclip, wl-copy/wl-paste and pbcopy/pbpaste are supported.
 
+.. _multiline:
+
+Multiline editing
+-----------------
+
+The fish commandline editor can be used to work on commands that are several lines long. There are three ways to make a command span more than a single line:
+
+- Pressing the :kbd:`Enter` key while a block of commands is unclosed, such as when one or more block commands such as ``for``, ``begin`` or ``if`` do not have a corresponding :ref:`end <cmd-end>` command.
+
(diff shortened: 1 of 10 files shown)


doc_src/index.rst
```diff
@@ -1498,21 +1513,35 @@ To search for previous entries containing the word 'make', type ``make`` in the
...
 
-The fish commandline editor works exactly the same in single line mode and in multiline mode. To move between lines use the left and right arrow keys and other such keyboard shortcuts.
+Directory stack
+---------------
+
+Another set of commands, usually also available in other shells like bash, deal with the directory stack. Stack handling is not automatic and needs explicit calls of the following commands:
+
+- :ref:`dirs <cmd-dirs>` prints the stack
+- :ref:`pushd <cmd-pushd>` adds a directory on top of the stack and makes it the current working directory
+- :ref:`popd <cmd-popd>` removes the directory on top of the stack and changes the current working directory
```
faho, line 1544: `cd -` seems worth a mention.

zanchey:
If you have time, I think it's worth trying to give some explanation of the stack - that's a concept that many shell users may never have come across. If not, I think we should merge this and look at updating it later.

Delapouite:
@zanchey I'm not sure of the phrasing to define the stack correctly and concisely yet. So I think we can merge it and improve upon in a future PR as you suggest.

PR #6854, opened 2020-04-04
Make ":" and "." proper builtins

faho:
## Description

We've tried to remove "." before, to disastrous results, so these aren't going anywhere. Since it's *simpler* to implement them in C++ than script, and since that makes fish just a little more functional without share/config.fish, I'd like to add them to C++ and make them real builtins.

## TODOs:

### DIFF
--- a/share/functions/__fish_config_interactive.fish
+++ b/share/functions/__fish_config_interactive.fish
@@ -168,7 +168,7 @@ function __fish_config_interactive -d "Initializations that should be performed 
     #
     # Only a few builtins take filenames; initialize the rest with no file completions
     #
-    complete -c(builtin -n | string match -rv '(source|cd|exec|realpath|set|\\[|test|for)') --no-files
+    complete -c(builtin -n | string match -rv '(.|:|source|cd|exec|realpath|set|\\[|test|for)') --no-files
 
     # Reload key bindings when binding variable change
     function __fish_reload_key_bindings -d "Reload key bindings when binding variable change" --on-variable fish_key_bindings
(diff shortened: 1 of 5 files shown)


krobelus:
Sounds good.
`.` should still have file completions, and I prefer `:` to have them as well, but I'm happy to be overruled here.

```diff
diff --git a/share/functions/__fish_config_interactive.fish b/share/functions/__fish_config_interactive.fish
index faef0ef05..1e62b96b2 100644
--- a/share/functions/__fish_config_interactive.fish
+++ b/share/functions/__fish_config_interactive.fish
@@ -168,7 +168,7 @@ function __fish_config_interactive -d "Initializations that should be performed
     #
     # Only a few builtins take filenames; initialize the rest with no file completions
     #
-    complete -c(builtin -n | string match -rv '(source|cd|exec|realpath|set|\\[|test|for)') --no-files
+    complete -c(builtin -n | string match -rv '^(.|:|source|cd|exec|realpath|set|\\[|test|for)$') --no-files
 
     # Reload key bindings when binding variable change
     function __fish_reload_key_bindings -d "Reload key bindings when binding variable change" --on-variable fish_key_bindings
```

faho:
Okay, there is one more issue here: Our `true` actually errors if it receives arguments, so this would break use like the `: job description` bit we used to mark what the `python` in the background was doing.

That seems quite wrong to begin with - `true` should just return true.

share/functions/__fish_config_interactive.fish
```diff
@@ -168,7 +168,7 @@ function __fish_config_interactive -d "Initializations that should be performed
     #
     # Only a few builtins take filenames; initialize the rest with no file completions
     #
-    complete -c(builtin -n | string match -rv '(source|cd|exec|realpath|set|\\[|test|for)') --no-files
+    complete -c(builtin -n | string match -rv '(.|:|source|cd|exec|realpath|set|\\[|test|for)') --no-files
```
ridiculousfish, line 171: I think the `.` has to be escaped here.

ridiculousfish:
As-is this outputs text like `complete -c '['` on launch; escaping the `.` fixes it.

faho:
Merged as be0de5e2d..4eccc0f6d, with the `\.` fix.

Note that also changes `true` and `false` to accept any number of arguments - as they should.

Issue #6838, opened 2020-03-30
Preview file shortcut


exploide:
Sometimes, when I type a command like `mycmd --my-file ./here/comes/some/file.txt` I start wondering whether I actually selected the correct file or if I mixed something up. Then I copy the file path, open a new terminal tab and open the file in `less`.

fish has some cool shortcuts, for example Alt+H to open the man page of the current command.

What about a shortcut for opening the current file in a pager? I think it could be useful.

I'm thinking about what shortcut would be appropriate. Some of them are already occupied for other actions. So I'm leaning towards Alt+O as in Open.

What do you think?

krobelus:
Good idea, seems convenient.

I'm playing around with this. It also works nicely when fish's pager is active.
`bind \eo 'eval less -I (commandline -ct); commandline -f repaint'`

We could also overload the Alt+L binding, to show the current token in a pager if it is a file.

exploide:
@krobelus are you working on it? Otherwise I would give it a try.

But it should be decided whether enhancing Alt+L or introducing a new function for Alt+O is preferred. I think enhancing  `__fish_list_current_token` with file support is great but would incidentally remove 

https://github.com/fish-shell/fish-shell/blob/a29bc127cec6a911fcd5919e0c19dcc14b0ef360/share/functions/__fish_list_current_token.fish#L10-L12

which strips the filename to list the parent directory. That would obviously preview the file then. But I'm not sure if this is a no-go since I cannot imagine this is heavily used.

krobelus:
> Otherwise I would give it a try.

Sure, go for it.

I think a new binding on Alt+O is fine.

Issue #5811, opened 2019-04-11
RFC: `expand` and safer alternatives to `eval`


mqudsi:
I've been looking at usage of fish's `eval` (both within our codebase and by others across github) and I've found that most usages don't actually require a full-blown eval with all the brittleness it entails (aside from the security risks it poses).

We've already made some good progress here with new fish 3.0 features:
* `$cmd_name` may be used as the head of a command - it is escaped before execution, which is exactly what we want in an `eval` alternative
* fish 3.0 also supports `command $command` and `buliltin $builtin` to do the same,
* #5575 would allow the same, but using command substitution for the payload rather than a variable

Remaining non-evil usages of `eval`:
1) A non-array variable needs to be evaluated as a command + space-separated arguments for interoperability (e.g. `$PAGER` being set to `less -R`)
2) Used to expand the contents of a variable literally where expansion is desired but quote bypassing is not

I propose introducing a new builtin `expand` for the latter, which when given an argument would return its expansion with no danger of executing anything, as compared to the extremely dangerous use of `eval` for this purpose, e.g. `eval set expanded $to_expand`, which works fine if `to_expand` is set to `*` or `hello_{foo, bar}` but fails miserably if `to_expand` is set to `'hello; echo "you\'ve been pwnd!"'`:

```fish
~> set to_expand 'hello; echo "you\'ve been pwnd!"'
~> eval set expanded $to_expand
you've been pwnd!
~> echo $expanded 
hello
~> set expanded (expand $to_expand)
~> echo $expanded
hello;
echo
you've been pwnd!
```

While this solves case 2, it also provides a way of solving case 1,

```fish
eval $EDITOR
```

can instead be interpreted as
```fish
set -l editor (expand $EDITOR)
$editor[1] $editor[2..-1]
```

and of course a wrapper can be provided for this, e.g.

```fish
function eval_safe
    set -l cmd (expand $argv[1])
    $cmd[1] $cmd[2..-1] $argv[2..-1]
end
```

faho:
I actually have one written that I need to polish up, but there's a few design decisions to be made here. In particular, what if you only want certain kinds of expansions (e.g. just tilde-expansion), and what about command substitutions?

Some prior discussion in #751.

mqudsi:
I imagine command substitution should not be performed, as it is a safer default and the safe equivalent can be achieved by first saving the result of an intentional command substitution to a variable then running `expand` as needed, but it's definitely a good question.

[...]

PR #6855, opened 2020-04-04
New function __fish_preview_current_file to open current file in pager

exploide:
## Description

New function `__fish_preview_current_file` to open current file in pager. Bound to Alt+O by shared key bindings.

Fixes issue #6838

## TODOs:

### DIFF
--- a/share/functions/__fish_preview_current_file.fish
+++ b/share/functions/__fish_preview_current_file.fish
@@ -1,0 +1,29 @@
+function __fish_preview_current_file --description "Open the file at the cursor in a pager"
+    set -l pager less --
+    set -q PAGER && echo $PAGER | read -at pager
+
+    # commandline -t will never return an empty list. However, the token
+    # could comprise multiple lines, so join them into a single string.
+    set -l file (commandline -t | string collect)
+    # count $file is 1
+
+    # $backslash will parsed as regex which may need additional escaping.
+    set -l backslash '\\\\'
+    not status test-feature regex-easyesc && set backslash $backslash$backslash
+    test -z $file && set file (string replace -ra -- '([ ;#^<>&|()"\'])' "$backslash\$1" (commandline -oc)[-1])
+    # count $file is 0 or 1
+
+    set -q file[1] || return
+    # Now file contains exactly one element, the unexpanded token.
+
+    # strip -option= from token if present
+    set file (string replace -r -- '^-[^\s=]*=(.*)' '$1' $file)
+    true # Clear $status from string replace, we don't want to return because of it.
+
+    eval set -l files $file || return # Return when eval fails.
+
+    if set -q files[1] && test -f $files[1]
+        $pager $files
+        commandline -f repaint
+    end
+end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 
 #### New or improved bindings
 - New readline commands `undo` (Ctrl+_) and `redo` (Alt-/) can be used to revert changes to the command line or the pager search field (#6570).
+- New function `__fish_preview_current_file` (Alt+O) opens the current file at the cursor in a pager (#6838 #6855).
 
 #### Improved prompts
 - The default and example prompts print the correct exit status for commands prefixed with `not` (#6566).
--- a/doc_src/index.rst
+++ b/doc_src/index.rst
@@ -1324,6 +1324,8 @@ Some bindings are shared between emacs- and vi-mode because they aren't text edi
 
 - :kbd:`Alt`\ +\ :kbd:`L` lists the contents of the current directory, unless the cursor is over a directory argument, in which case the contents of that directory will be listed.
 
+- :kbd:`Alt`\ +\ :kbd:`O` opens the current file at the cursor in a pager.
+
 - :kbd:`Alt`\ +\ :kbd:`P` adds the string '``| less;``' to the end of the job under the cursor. The result is that the output of the command will be paged.
 
 - :kbd:`Alt`\ +\ :kbd:`W` prints a short description of the command under the cursor.
--- a/share/functions/__fish_shared_key_bindings.fish
+++ b/share/functions/__fish_shared_key_bindings.fish
@@ -84,6 +84,7 @@ function __fish_shared_key_bindings -d "Bindings shared between emacs and vi mod
     bind --preset $argv \e. history-token-search-backward
 
     bind --preset $argv \el __fish_list_current_token
+    bind --preset $argv \eo __fish_preview_current_file
     bind --preset $argv \ew __fish_whatis_current_token
     # ncurses > 6.0 sends a "delete scrollback" sequence along with clear.
     # This string replace removes it.


share/functions/__fish_preview_current_file.fish
```diff
@@ -0,0 +1,16 @@
+function __fish_preview_current_file --description "Open the current file in a pager"
+    # set pager to user-preferred pager or less
+    set -l pager "less --"
+    set -q PAGER && echo $PAGER | read -at pager
+
+    # get current or last token, use eval to resolve $HOME and ~
+    set -l file (eval echo (commandline -t))
```
krobelus, line 7: I think `eval set -l file (commandline -t)` does the same thing.
Since a token like `a{1,2}` can expand to multiple tokens, maybe it's better to explicitly pick the first one for now? That'd be `set -l file (eval echo (commandline -t))[1]`. Or open all of them in less?

share/functions/__fish_preview_current_file.fish
```diff
@@ -0,0 +1,16 @@
+function __fish_preview_current_file --description "Open the current file in a pager"
+    # set pager to user-preferred pager or less
+    set -l pager "less --"
```
krobelus, line 3: Make this a more idiomatic array (`set -l pager less --`)
That'd be consistent with the next line, since that uses `read -a`.

share/functions/__fish_preview_current_file.fish
```diff
@@ -0,0 +1,16 @@
...
+    set -q PAGER && echo $PAGER | read -at pager
+
+    # get current or last token, use eval to resolve $HOME and ~
+    set -l file (eval echo (commandline -t))
+    test -z "$file" && set -l file (eval echo (commandline -oc)[-1])
+
+    # strip -option= from token if necessary
+    not test -e "$file" && set -l file (eval echo (string replace -r -- '^-[^\s=]*=(.*)' '$1' "$file"))
+
+    test -f "$file" && eval $pager $file
```
krobelus, line 13: With the above change to `pager` there is no longer a need to use `eval` here. So drop it, since prevents previewing files with special characters.

share/functions/__fish_preview_current_file.fish
```diff
@@ -0,0 +1,16 @@
...
+    # set pager to user-preferred pager or less
+    set -l pager "less --"
+    set -q PAGER && echo $PAGER | read -at pager
+
+    # get current or last token, use eval to resolve $HOME and ~
+    set -l file (eval echo (commandline -t))
+    test -z "$file" && set -l file (eval echo (commandline -oc)[-1])
+
+    # strip -option= from token if necessary
+    not test -e "$file" && set -l file (eval echo (string replace -r -- '^-[^\s=]*=(.*)' '$1' "$file"))
```
krobelus, line 11: I think this `eval` can be dropped: `$file` is already expanded, and you are only removing a prefix.

share/functions/__fish_preview_current_file.fish
```diff
@@ -0,0 +1,16 @@
+function __fish_preview_current_file --description "Open the current file in a pager"
```
krobelus, line 1: something like "file at the cursor"

share/functions/__fish_preview_current_file.fish
exploide: Unfortunately not. The `~` only gets expanded when at the beginning of the token. That's not the case when eval runs first. Test for example `eval echo "--long-opt=~/Desktop/foo"`. Hence the second run after removing the options.
I will incorporate your other suggestions tomorrow.

krobelus: Oh yeah you're right. Maybe we should do the replacement before eval.

share/functions/__fish_preview_current_file.fish
exploide: I ignored the `a{1,2}` after trying for a while. I did not get it to work for both, this kind of expansion as well as single filenames containing whitespace. If this is possible, please give me an example.

share/functions/__fish_preview_current_file.fish
exploide: done

share/functions/__fish_preview_current_file.fish
exploide: done

share/functions/__fish_preview_current_file.fish
exploide: done

exploide:
I incorporated most of your suggestions. Please have a second look.

However, I encountered another problem, so I would not merge this now:

This use of `eval`, at least the second and probably also the third use, poses a funny security vulnerability. Imagine you want to examine files originating from someone else. If the person prepared a file with the name `hello; touch pwned` and you do somethink like `>_ your-command 'hello; touch pwned' <Alt+O>`, then the `eval` executes that.
It's a hilarious attack vector, but still it is one :smile: 

I haven't looked yet whether existing functions or completions have the same problem.

Can this easily be solved by using `commandline` and `eval` in a different way? Or do we need another way of expanding arguments? @faho said on Gitter it would need a new shell builtin when not using `eval` for expansion.

krobelus:
I think your solution works fine for most use cases, but it is surprisingly tricky to fix the issues.
I tried to do that with below example that also supports multiple files.

> The problem with (eval echo (commandline -oc)[-1]) is that after
> accessing index [-1], the result is no longer escaped.

Yeah I just realized that. `commandline -o` does tokenize, but not
expand the likes of `$`, `~`, `{}` and `[]`. So passing it to eval is
wrong because eval tokenizes *again*.
My fix is to escape all characters that are relevant for tokenization, but
not the ones that will be expanded later. We should have better support
for use cases like this. Here we could either

1. Have a variant of `eval` that does only expansion, not tokenization.
   This is #5811.

2. Add a new flag to `commandline -o` to make it return an escaped string,
   that can be passed to eval safely. This is similar to my manual escaping.

> This use of eval, at least the second and probably also the third use,
> poses a funny security vulnerability.

Yeah, the third eval evaluates the result of a former evaluation. We
should not do that, since that's not what fish would do when executing
the command. To fix that I do the replacement of a "-option=" prefix
before evaluating. This will fail to work for files that literally start
like that but I don't think we care.

The second eval suffers from the double-tokenization issue; that's
hopefully fixed by escaping.

When the cursor is on something that does not tokenize (for example
unclosed quotes `"asdf`, pressing `Control`+`O` prints the tokenizer
error message. I think that's better than silencing the error.
This message comes from eval, which fails, so I don't run `less`, or repaint if that happens.

I got rid of some variable quoting, instead I explicitly check `set -q
foo[1]`. Of course that's just a matter of style.

----

```fish
function __fish_preview_current_file --description "Open the file at the cursor in a pager"
    set -l pager less --
    set -q PAGER && echo $PAGER | read -at pager

    # commandline -t will never return an empty list. However, the token
    # could comprise multiple lines, so join them into a single string.
    set -l file (commandline -t | string collect)
    # count $file is 1
    # $backslash will parsed as regex which may need additional escaping.
    set -l backslash '\\\\'
    not status test-feature regex-easyesc && set -l backslash $backslash$backslash
    test -z $file && set file (string replace -ra -- '([ ;#^<>&|()"\'])' "$backslash\$1" (commandline -oc)[-1])
    # count $file is 0 or 1

    set -q file[1] || return
    # Now file contains exactly one element, the unexpanded token.

    # strip -option= from token if necessary
    set file (string replace -r -- '^-[^\s=]*=(.*)' '$1' $file)

    true # Clear $status from string replace, we don't want to return because of it.
    eval set -l files $file || return # Return when eval fails.
    
    if set -q files[1] && test -f $files[1]
        $pager $files
        commandline -f repaint
    end
end
```

exploide:
Thanks for continuing this. I will test that tomorrow and come back.

And thanks for linking the issue related to an expand builtin. I think that would be the way to go. Good that @faho already has something like that. The problem with evaluating harmful file names is already present for example in the completions of `unzip`.

exploide:
Seems to work so far, thanks. I pushed that version now. It's ready from my side. At least until we have a safe alternative to eval.

When happens the following? When a filename contains line breaks?

> However, the token could comprise multiple lines, so join them into a single string

krobelus:
> When happens the following? When a filename contains line breaks?

Exactly, you can create a file like `touch a\nb`.
Opening such files with line breaks works if the cursor is directly on the token (though only after I added another `string collect`). If the cursor is past the token, then `(commandline -o)[-1]` is used, so this tries to open only the last line in the last token.

Merged to master as 8025e80fdb90a880aeb4fe6aadfb4919421be138, thank you!

PR #6857, opened 2020-04-04
clang-tidy again

neheb:


neheb:
The first commit causes the test failures. I can't see where the problem is.

neheb:
Fixed with an explicit cast.

src/screen.cpp
```diff
@@ -1155,7 +1155,7 @@ void s_reset(screen_t *s, screen_reset_mode_t mode) {
         if (screen_width > non_space_width) {
             bool justgrey = true;
             if (cur_term && enter_dim_mode) {
-                std::string dim = tparm((char *)enter_dim_mode);
+                std::string dim = tparm(enter_dim_mode);
```
faho, line 1158: NAK. We've been over this. This *needs* to stay because of differences in curses implementations.

faho: (If anyone has a better way of expressing this I'm all ears. The issue is macOS having ncurses configured as taking non-const - an option that was since removed - and NetBSD curses insisting on const, or something along those lines)

faho:
Merged without the breaking (and super ugly IMHO) C++ casting change.

Note that I have complained about *that exact change before*. I don't want to have to complain again.

PR #6862, opened 2020-04-05
In Fish MD5 on BSD now use given String and not -s

Mazorius:
## Description

On a BSD system the `-s` option was used instead of the string passed to `fish_md5`.

Fixes issue -

## TODOs:

- [-] Changes to fish usage are reflected in user documentation/manpages.
- [-] Tests have been added for regressions fixed

Changed to use given String and not option `-s`.


Mazorius:
The error seems to be here: https://github.com/fish-shell/fish-shell/pull/6862/checks?check_run_id=562696937#step:6:277

But I do not find the issue … maybe someone can have a look?

ridiculousfish:
Merged as 77fb54fa9930745b8676ae3a6bc5eab11a2b389b, thank you!

PR #6864, opened 2020-04-06
Improve gradle completion

Mazorius:
## Description

* Add support of gradlew and gdub
* make support for options a lot better
* fix issue in tasks completion
  * missing completion on none project folder without caching

## TODOs:

### DIFF
--- a/share/completions/gradle.fish
+++ b/share/completions/gradle.fish
@@ -12,7 +12,7 @@ function __create_completion_cache_file
     end
     mkdir -m 700 -p $XDG_CACHE_HOME/gradle-completions
 
-    string trim -- $XDG_CACHE_HOME/gradle-completions/(fish_md5 -s $argv[1] | string split ' = ')[2]
+    string trim -- $XDG_CACHE_HOME/gradle-completions/(__fish_md5 -s $argv[1] | string split ' = ')[2]
 end
 
 ##############################
--- a/share/completions/gradle.fish
+++ b/share/completions/gradle.fish
@@ -2,7 +2,7 @@
 # See: https://gradle.org
 
 function __contains_gradle_build
-    test \( -e ./build.gradle -a -f ./build.gradle \) -o \( -e ./build.gradle.kts -a -f ./build.gradle.kts \)
+    test -f build.gradle -o -f build.gradle.kts
 end
 
 function __create_completion_cache_file
@@ -25,20 +25,19 @@ function __get_gradle_default_task_completion
         return
     end
 
-    string replace --all ' - ' \t -- "\
-        buildEnvironment - Displays all buildscript dependencies declared in root project.
-        components - Displays the components produced by root project.
-        dependencies - Displays all dependencies declared in root project.
-        dependencyInsight - Displays the insight into a specific dependency in root project.
-        dependentComponents - Displays the dependent components of components in root project.
-        help - Displays a help message.
-        init - Initializes a new Gradle build.
-        model - Displays the configuration model of root project.
-        projects - Displays the sub-projects of root project.
-        properties - Displays the properties of root project.
-        tasks - Displays the tasks runnable from root project.
-        wrapper - Generates Gradle wrapper files." \
-        | string trim
+    printf '%s\t%s\n' \
+        "buildEnvironment" "Displays all buildscript dependencies declared in root project." \
+        "components" "Displays the components produced by root project." \
+        "dependencies" "Displays all dependencies declared in root project." \
+        "dependencyInsight" "Displays the insight into a specific dependency in root project." \
+        "dependentComponents" "Displays the dependent components of components in root project." \
+        "help" "Displays a help message." \
+        "init" "Initializes a new Gradle build." \
+        "model" "Displays the configuration model of root project." \
+        "projects" "Displays the sub-projects of root project." \
+        "properties" "Displays the properties of root project." \
+        "tasks" "Displays the tasks runnable from root project." \
+        "wrapper" "Generates Gradle wrapper files."
 end
 
 complete --command 'gw' --command 'gradle' --command 'gradlew' \
@@ -52,7 +51,7 @@ function __get_gradle_task_completion
     end
 
     set -l gradle_cache_file (__create_completion_cache_file "{$PWD}-tasks")
-    if not command test -f $gradle_cache_file
+    if not command test -f $gradle_cache_file -a -s $gradle_cache_file
         command gradle -q tasks --all 2>/dev/null | string match --regex '^[a-z][A-z:]+.*' | string replace ' - ' \t >$gradle_cache_file
     end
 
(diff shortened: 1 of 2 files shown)


share/completions/gradle.fish
```diff
@@ -1,60 +1,281 @@
...
-complete -c gradle -l stop -d 'Stop daemon if running'
-complete -c gradle -l continuous -s t -d 'Enable continuous build'
-complete -c gradle -l no-search-upward -s u -d 'Don\'t search in parent folders for settings file'
-complete -c gradle -l version -s v -d 'Print version'
-complete -c gradle -l exclude-task -s x -x -d 'Specify task to be excluded from execution'
-
-# https://github.com/hanny24/gradle-fish/blob/master/gradle.load
-function __cache_or_get_gradle_completion
+function __contains_gradle_build
+    test \( -e ./build.gradle -a -f ./build.gradle \) -o \( -e ./build.gradle.kts -a -f ./build.gradle.kts \)
```
faho, line 5: The existence tests are unneeded if you're checking for fileness

`test -f build.gradle -o -f build.grade.kts`

does the exact same thing - something can't be a file without existing.

share/completions/gradle.fish
```diff
@@ -1,60 +1,281 @@
...
     if test -z $XDG_CACHE_HOME
         set XDG_CACHE_HOME $HOME/.cache/
     end
     mkdir -m 700 -p $XDG_CACHE_HOME/gradle-completions
 
-    set -l hashed_pwd (__fish_md5 -s $PWD)
-    set -l gradle_cache_file $XDG_CACHE_HOME/gradle-completions/$hashed_pwd
-    if not test -f $gradle_cache_file; or command test build.gradle -nt $gradle_cache_file
-        command gradle -q tasks 2>/dev/null | string match -r '^[[:alnum:]]+ - .*' | string replace ' - ' \t >$gradle_cache_file
+    string trim -- $XDG_CACHE_HOME/gradle-completions/(fish_md5 -s $argv[1] | string split ' = ')[2]
```
faho, line 15: 926c1f101 renamed the function to `__fish_md5`.

share/completions/gradle.fish
```diff
@@ -1,60 +1,281 @@
...
+    --exclusive \
+    --arguments "(__get_gradle_task_completion)"
+
+
+###############################
+# Configure Option Completion #
+###############################
+
+function __get_console_completion
+    string replace --all ' - ' \t -- "\
```
faho, line 74: Typically this will be expressed with something like

```fish
printf '%s\t%s\n' \
auto "Enable color and other rich output in the console output when the build process is attached to a console, or to generate plain text only when not attached to a console. This is the default when Gradle is attached to a terminal." \
plain "This option disables all color and other rich output in the console output. This is the default when Gradle is not attached to a terminal."
```

Also these string are *entirely too long* as a description. Please rewrite!

share/completions/gradle.fish
```diff
@@ -1,60 +1,281 @@
...
+        auto - Enable color and other rich output in the console output when the build process is attached to a console, or to generate plain text only when not attached to a console. This is the default when Gradle is attached to a terminal.
```
faho, line 75: E.g. "enable color in terminal"

share/completions/gradle.fish
```diff
@@ -1,60 +1,281 @@
...
+        plain - This option disables all color and other rich output in the console output. This is the default when Gradle is not attached to a terminal.
```
faho, line 76: "disable color"

share/completions/gradle.fish
```diff
@@ -1,60 +1,281 @@
...
+        rich - Enable color and other rich output in the console output, regardless of whether the build process is not attached to a console. When not attached to a console, the build output will use ANSI control characters to generate the rich output.
```
faho, line 77: "always enable color"

share/completions/gradle.fish
```diff
@@ -1,60 +1,281 @@
...
+    string replace --all ' - ' \t -- "\
+        normal - Set the default process priority.
+        low - Set a low process priority." \
+        | string trim
+end
+
+function __get_warning_mode_completion
+    string replace --all ' - ' \t -- "\
+        all - Log all warnings.
+        summary - Suppress all warnings and log a summary at the end of the build.
```
faho, line 110: "only output summary at the end"

share/completions/gradle.fish
```diff
@@ -1,60 +1,281 @@
...
+        fail - Log all warnings and fail the build if there are any warnings.
```
faho, line 111: "Show warnings and fail if there are any"

share/completions/gradle.fish
```diff
@@ -1,60 +1,281 @@
...
+        none - Suppress all warnings, including the summary at the end of the build." \
```
faho, line 112: "Show no warnings"

share/completions/gradle.fish
```diff
@@ -1,60 +1,281 @@
...
+        plain - This option disables all color and other rich output in the console output. This is the default when Gradle is not attached to a terminal.
+        rich - Enable color and other rich output in the console output, regardless of whether the build process is not attached to a console. When not attached to a console, the build output will use ANSI control characters to generate the rich output.
+        verbose - Enable color and other rich output like the rich, but output task names and outcomes at the lifecycle log level, as is done by default in Gradle 3.5 and earlier." \
+        | string trim
+end
+
+function __get_property_completion
+    string replace --all ' - ' \t -- "\
+        org.gradle.cache.reserved.mb - Reserve Gradle Daemon memory for operations.
+        org.gradle.caching - Set true to enable Gradle build cache.
```
faho, line 85: "enable build cache"

share/completions/gradle.fish
```diff
@@ -1,60 +1,281 @@
...
+        org.gradle.console - Set type of console output to generate (plain auto rich verbose).
```
faho, line 86: the options are already given - "type of output"

share/completions/gradle.fish
```diff
@@ -1,60 +1,281 @@
...
+        org.gradle.daemon.debug - Set true to debug Gradle Daemon.
```
faho, line 87: "enable gradle debugging"

share/completions/gradle.fish
Mazorius: ok but the following happens if I change it to `__fish_md5`

```
fish: Unknown command: __fish_md5
```

ammgws: You need to be on fish master

Mazorius: Ah … `__fish_md5` is not in the current release?

ammgws: Not yet because the commit is from a couple weeks ago.

share/completions/gradle.fish
```diff
@@ -1,60 +1,276 @@
...
+
+# Outside of a Project
+function __get_gradle_default_task_completion
+    if __contains_gradle_build
+        return
     end
-    cat $gradle_cache_file
+
+    printf '%s\t%s\n' \
+        "buildEnvironment" "Displays all buildscript dependencies declared in root project." \
```
faho, line 29: To be honest these are still a bit verbose.

"Display buildscript dependencies"

Note the lack of the "s" - for one this is an *order* you give the command, so the imperative makes sense, for another just ditching the "s" saves a character, just like removing the "." does.

And it displaying _all_ dependencies just seems implicitly true. If it says dependencies, I would assume it's all of them, or it should tell me why it isn't all.

share/completions/gradle.fish
```diff
@@ -1,60 +1,276 @@
...
+        "components" "Displays the components produced by root project." \
+        "dependencies" "Displays all dependencies declared in root project." \
+        "dependencyInsight" "Displays the insight into a specific dependency in root project." \
```
faho, line 32: What is "the insight"?

I don't know gradle, so I have no idea what this means. Descriptions are often a good place to explain.

Unless this is something the typical gradle user would know immediately.

My guess is this is something like "Explain a specific dependency"? Or "Show info on a dependency"?

share/completions/gradle.fish
```diff
@@ -1,60 +1,276 @@
...
+        "auto" "Use 'rich' in console otherwise 'plain'" \
+        "plain" "Disable color and rich output." \
+        "rich" "Enable color and rich output." \
+        "verbose" "Enable color, rich output, output task names and outcomes at the lifecycle log level"
+end
+
+function __get_property_completion
+    printf '%s\t%s\n' \
+        "org.gradle.cache.reserved.mb" "Reserve Gradle Daemon memory for operations." \
+        "org.gradle.caching" "Set true to enable Gradle build cache." \
```
faho, line 83: So this is a boolean. Boolean options are typically named so that the thing is done if they are true.

So the "Set true" is just superfluous - "Enable Gradle build cache".

share/completions/gradle.fish
szpak: `dependencyInsight` allows - in general - to filter what you want to see (e.g. one dependency or subset, not all of them). Maybe: "Displays specific dependencies in root project"?

Btw, @Mazorius maybe it would be good to replace "Displays" with "Shows" to save 3 chars in every case?

szpak: On the other hand, currently they are consistent with the naming using in Gradle's task desctiptions :-/

share/completions/gradle.fish
szpak: Assuming you decide that it is ok to diverge from the Gradle naming convention, `all` probably could be removed also for `gradle dependencies` - that task can be reduced to display dependencies only from the specific configuration and then it is all for that configuration, but not all at all :).

faho: It *definitely* is okay to diverge from the exact wording of Gradle's documentation.

These things have a very different purpose.

Now "naming convention" implies that we'd e.g. use our own *terms*. That's probably not a good thing to do. What Gradle calls a "function" we should also call a "function" and not a "method" (to pick a common example).

share/completions/gradle.fish
Mazorius: I improved the description a bit more, removed `.` and change Displays to Display.
Maybe thats good?

For Insight you need to add a dependency to your command … than gradle will show you from which dependencies it is.

share/completions/gradle.fish
```diff
@@ -1,60 +1,276 @@
...
-complete -c gradle -l status -d 'Shows status of running andrecently stopped daemon'
-complete -c gradle -l stop -d 'Stop daemon if running'
-complete -c gradle -l continuous -s t -d 'Enable continuous build'
-complete -c gradle -l no-search-upward -s u -d 'Don\'t search in parent folders for settings file'
-complete -c gradle -l version -s v -d 'Print version'
-complete -c gradle -l exclude-task -s x -x -d 'Specify task to be excluded from execution'
-
-# https://github.com/hanny24/gradle-fish/blob/master/gradle.load
-function __cache_or_get_gradle_completion
+function __contains_gradle_build
```
krobelus, line 4: All these functions are put in the global scope when the completion is loaded.
I think we usually use a common prefix like `__fish_gradle`, so this'd be  `__fish_gradle_contains_build` and so on.

Mazorius: ok … so all function in this completion should start with `__fish_gradle_…`?

ammgws: Yeah it's a convention used throughout the included completions/functions supplied by fish

krobelus: Yes, that'd be nice (unless the function is shared with some other completion).
 I just checked, we do the same in completions for `git`, `ip`, `hg`, `zypper`, `zfs`, and `composer`, for example.

Mazorius: I changed the names. Please resolve if everything is fine now

faho:
Merged, thanks!

PR #6865, opened 2020-04-06
Remove unnecessary string duplication in handle_flag_f.

Liorst4:
Prevents a memory leak.


faho:
Yeah, that seems correct. Merged, thanks!

PR #6867, opened 2020-04-06
Allow `man` completions on catalina if `apropos` is overridden

thomcc:
## Description
It's pretty easy to fix catalina's apropos with [a small tweak](https://apple.stackexchange.com/a/375629), so it would be nice if fish's man completions worked for users who do this and override the default `apropos`. Making that work is just an extra check at the start of the completion handler so it seems worth it to me.

## TODOs:


faho:
You know what would be great? If this billion dollar company would... fix their stuff.

Blergh.

Yeah, the workaround is solid, merging!

zanchey:
Let's take this for 3.1.1 as well.

PR #6868, opened 2020-04-06
Add completions for ripgrep (`rg`).

thomcc:
## Description

These are roughly equivalent to the autogenerated ones but

- Don't need manual installation.
- Offers correct completions `type` and `encoding` with the set of valid args.
- Flags which require arguments / don't accept file args are properly annotated.

It also doesn't have `-n "__fish_use_subcommand"` on each call since I'm pretty sure that's just because of how `clap` (a rust cli arg parsing library) autogenerates fish completions.

See https://github.com/BurntSushi/ripgrep/issues/1543

## TODOs:

### DIFF
--- a/share/completions/rg.fish
+++ b/share/completions/rg.fish
@@ -1,0 +1,190 @@
+complete -xc rg -s A -l after-context -d 'Show NUM lines after each match.'
+complete -xc rg -s B -l before-context -d 'Show NUM lines before each match.'
+complete -xc rg -l color -d 'Controls when to use color.' -a "never auto always ansi"
+complete -xc rg -l colors -d 'Configure color settings and styles.'
+complete -xc rg -s C -l context -d 'Show NUM lines before and after each match.'
+complete -c rg -l context-separator -d 'Set the context separator string.'
+complete -xc rg -l dfa-size-limit -d 'The upper size limit of the regex DFA.'
+complete -xc rg -l engine -d 'Specify which regexp engine to use.' -a "default pcre2 auto"
+complete -rc rg -s f -l file -d 'Search for patterns from the given file.'
+complete -rc rg -s g -l glob -d 'Include or exclude files.'
+complete -rc rg -l iglob -d 'Include or exclude files case insensitively.'
+complete -rc rg -l ignore-file -d 'Specify additional ignore files.'
+complete -xc rg -s M -l max-columns -d 'Don\'t print lines longer than this limit.'
+complete -xc rg -s m -l max-count -d 'Limit the number of matches.'
+complete -xc rg -l max-depth -d 'Descend at most NUM directories.'
+complete -xc rg -l max-filesize -d 'Ignore files larger than NUM in size.'
+complete -xc rg -l path-separator -d 'Set the path separator.'
+complete -rc rg -l pre -d 'search outputs of COMMAND FILE for each FILE'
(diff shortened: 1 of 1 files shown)


thomcc:
I think this CI failure is unrelated.

exploide:
I think if the maintainers decide to merge this, the completion bundled with ripgrep should be removed there. Otherwise yours would never be loaded since `fish/vendor.completions.d` takes precedence over `fish/completions`.

faho:
>I think this CI failure is unrelated.

It is. We've been testing Github Actions CI, and the mac one might possibly be even more flakey than Travis (tho we have some tests that have too strict timing requirements)

faho:
>I think if the maintainers decide to merge this, the completion bundled with ripgrep should be removed there.

That is up to packagers - rg doesn't actually install the completions.

faho:
We used to ship rg completions, quite recently in fact.

f036d019618ff1f0508c539dc88d64a194cd2034 dropped them (and the bat ones) because, as it turns out, packagers installed the completions from the rg package into /usr/share/fish/completions instead of /usr/share/fish/vendor_completions.d where they'd belong, and so the packages conflicted. (to be clear, this wasn't just a distro, but multiple)

And this all took a while to get sorted out (longer than anyone liked, frankly), and in the meantime some people didn't know how to install fish (as they understandably didn't want to run their package manager with the "--ignore-file-conflicts-yes-I-know-this-could-break-stuff" option).

So: I'm quite wary of starting to ship these. I think we'd have to let the pkg-config improvements start to permeate distributions first, and perhaps improve our distribution guidance.

Ideally we'd just be able to coordinate with rg upstream to drop them, but I'm not entirely sure that'd work out.

thomcc:
> Ideally we'd just be able to coordinate with rg upstream to drop them, but I'm not entirely sure that'd work out.

Might be worth asking. In https://github.com/BurntSushi/ripgrep/issues/1543#issuecomment-609767339 @BurntSushi indicated it was mostly supported because it was relatively easy to add. Could be wrong though.

BurntSushi:
I'd be happy to drop the fish completions from the ripgrep build. They are indeed only there because they are auto-generated by the arg parser. I have no hand in it. If they weren't auto-generated, then I definitely wouldn't be maintaining them on my own anyway. So if the fish project wants to own completions for ripgrep, then I'm more than happy with that arrangement.

faho:
>I'd be happy to drop the fish completions from the ripgrep build.

That would be great, if we can get packagers to notice, or to upgrade ripgrep and the next fish at the same time.

Really this is a problem of communication, and I haven't really figured it out.

We could possibly add an explicit note to the packagers to the CHANGELOG? Do they read those?

BurntSushi:
Yeah, it would certainly at least be in the CHANGELOG. Another possibility is to just remove fish completions from the build output so that they can't be used, but that seems a bit harsh.

We could do it in stages. In the next release, I'll add a prominent note in the CHANGELOG that fish completions will be removed in the next major release. Would that work?

faho:
>We could do it in stages. In the next release, I'll add a prominent note in the CHANGELOG that fish completions will be removed in the next major release. Would that work?

Yeah, that would work.

The release after that is out, we'll then merge this one and hope packagers get the idea.

faho:
Okay, it looks like these are going to stay in rg, so I'm closing.

PR #6866, opened 2020-04-06
Support `.jar` and `.aar` files in `unzip` completions

thomcc:
## Description
I've been dealing with these a lot recently (android dev...), and it's
pretty annoying that unzip completions don't recognize them (They're
just zip files with a weird file extension).

## TODOs:

- [ ] Changes to fish usage are reflected in user documentation/manpages.
- [ ] Tests have been added for regressions fixed
- [ ] User-visible changes noted in CHANGELOG.md

PR #6869, opened 2020-04-06
Allow unzip versions not patched by Debian to enjoy .{jar,aar} completions

Asgavar:
A minor follow-up to #6866 (e658a88ab0066987fcec90c40b3e454836e75eb2).
These file types should be properly handled by other unzip flavors too,
regardless of Debian's/non-Linux Unixes' idiosyncrasies.


faho:
Yeah, I didn't notice that bit. Thanks!

PR #6876, opened 2020-04-09
git prompt: resilient check for git stash

charego:
The refs/stash file is deleted by "git gc". The prompt won't display
stash information until another "git stash" command is run.

Using "git rev-parse stash" seems more resilient.

Based the choice on "git stash" test cases:
https://github.com/git/git/blob/master/t/t3903-stash.sh

There is still one edge case that I know of:
- If there is a branch named "stash", and no actual stashes,
this will report that there are stashes

### DIFF
--- a/share/functions/fish_git_prompt.fish
+++ b/share/functions/fish_git_prompt.fish
@@ -414,7 +414,7 @@ function fish_git_prompt --description "Prompt function for Git"
             end
 
             if set -q __fish_git_prompt_showstashstate
-                and test -r $git_dir/refs/stash
+                and test -r $git_dir/logs/refs/stash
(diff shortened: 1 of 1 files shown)


charego:
The informative git prompt checks for `.git/logs/refs/stash` instead of `.git/refs/stash`, and it appears "git gc" doesn't wipe that. If you want I can make that simple change instead.

krobelus:
> The informative git prompt checks for `.git/logs/refs/stash` instead of `.git/refs/stash`, and it appears "git gc" doesn't wipe that. If you want I can make that simple change instead.

Yeah let's do that.

charego:
@krobelus Done, and updated commit message.

krobelus:
Neat, thanks!

Issue #6816, opened 2020-03-26
shell integration with tokenization error


PerBothner:
Summary: Tokenization errors don't work nicely with shell integration, at least as implemented on [DomTerm](https://domterm.org).  A possibly fix to for Fish to emit a `fish_postexec` event in this situation.

To reproduce, type a lone right-parenthesis and then Enter. Fish prints the error `fish: Unexpected ')' for unopened parenthesis` and then re-displays the prompt and incorrect input, ready for editing.  Fish calls `fish_prompt` function, but does not emit any events that I can see.

This causes a problem for shell integration, because it looks like a regular re-display during input editing. In such a situation, DomTerm moves the cursor to the start of the "current" editing area - i.e. what Fish thinks is the previous input area.  The reason for this is that it is impossible for Fish to correctly move to the start of the input area if the window resizes, so when "start prompt" is received, DomTerm ignores wherever Fish has moved the cursor to, and overrides the cursor to move it to the start of the input area.  (The reason Fish cannot correctly move the cursor is that it cannot know the line width of the terminal when its re-display escape sequences are received by the terminal - the width may have changed again.)

There are various ways to fix or work-around this problem.  The simplest (though untested) may be for fish to emit `fish_postexec` after the error message.  This seems consistent with the documentation: "This event will be emitted even if the command is invalid."

Fish version: 3.1.0-439-g69afb1b56
OS: Fedora Linux 31
Terminal: DomTerm (git trunk)

PR #6880, opened 2020-04-10
Fire fish_postexec event on tokenization error

PerBothner:
Fixes issue #6816 "shell integration with tokenization error"

## TODOs:

CHANGELOG.md note could be:
A tokenization error (such as command-line containing just `)`) now raises the `fish_postexec` event, as documented.

### DIFF
--- a/src/reader.cpp
+++ b/src/reader.cpp
@@ -2173,7 +2173,8 @@ parser_test_error_bits_t reader_shell_test(parser_t &parser, const wcstring &b) 
             error_desc.push_back(L'\n');
         }
         std::fwprintf(stderr, L"\n%ls", error_desc.c_str());
-        event_fire_generic(parser, L"fish_postexec");
+        wcstring_list_t argv(1, b);
+        event_fire_generic(parser, L"fish_postexec", &argv);
     }
     return res;
 }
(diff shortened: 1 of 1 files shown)


faho:
So I've been thinking about this, and I think it would be nicer to introduce a *new* event if there's an error.

The reason being that it's not entirely clear that postexec handlers would be useful if nothing was actually executed.

Your function would then just register for both events.

PerBothner:
A distinct signal for error would also work. Wold it be only for tokenization errors (corresponding to my patch), or also other kinds of errors (which would require other changes)?

faho:
>A distinct signal for error would also work. Wold it be only for tokenization errors (corresponding to my patch), or also other kinds of errors (which would require other changes)?

I would imagine all kinds of errors that cause us not to execute a command - I think that might be

- tokenization (i.e. the syntax is wrong)
- empty variable-as-command - e.g. `$PAGER file` if $PAGER is unset
- empty globs - `ls *.nonexistent`

But I think for now just the tokenization will suffice.

ridiculousfish:
In addition to faho's suggestions, please move the code to after where `reader_shell_test` is invoked. [Here](https://github.com/fish-shell/fish-shell/blob/d278ff4e45879d8570d7632914595622598fe0d3/src/reader.cpp#L2932) is where fish acts on the result of `reader_shell_test`.

PerBothner:
Ok - I moved the `event_fire_generic` call as suggested (which I also initially considered) and renamed it to `fish_posterror`.

However, this is still one (less serious) problem: The error message is printed while we're still in input mode, so it is printed with the wrong style.  Fixing this using heuristics is no doubt possible, but it would be less complex/fragile to have an explicit `fish_preerror` event.  However, I'm not sure where it should go, given that the comment before `report_tokenizer_error` says "This is a hack".

Issue #5003, opened 2018-05-21
Implement global abbreviations


admirabilis:
fish, version 2.7.1

It would be best if `abbr` would expand after sudo. So typing something like `sudo mv<SPACE>` would expand to `sudo mv -i `. Maybe a way to implement that would be hardcoding `abbr` to work specifically after the sudo command too...

Currently, I have a `function` that wraps other functions so they may work with sudo. But `abbr` is much safer in that you see what additional arguments are being given.

<s><sup>Another blocking issue for my own reference https://github.com/fish-shell/fish-shell/issues/3233</sup></s>

ridiculousfish:
fish should not establish an abbreviation-savvy whitelist, which users and new commands can't participate in.

IMO we ought to simply implement global abbreviations. These would be abbreviations which work everywhere, not just in command position.  I think this would solve most of the pain around sudo, and also enable e.g. `!!` which we could even ship by default.

mig-hub:
I agree with @ridiculousfish because it is quite a common thing to have commands taking commands as argument like `sudo` but also `bundle exec`, `time`, `dotenv` to name a few and it is also quite common to write functions which take a command as an argument. Keeping a white list would be limiting and also they can be nested.

mqudsi:
@ridiculousfish that would also address the questions we get about aliases that can trigger redirects or piping, which are possible under bash (and zsh?) thanks to the substitution of aliases prior to evaluation.

alxpettit:
I think the option to whitelist/blacklist for global abbreviations would be super useful though -- for example, let's say I whitelist `youtube-dl` to `ytdl`. That's fine. And then I'm cding into the git repo for `youtube-dl` to try the latest commit, and oh no. It just tried to cd to `ytdl` instead. That's not helpful at all.

faho:
>so, in my opinion, each global abbreviation should have a whitelist option, and a blacklist option -- and if neither are used, then it can apply everywhere default. 

Well... that just means you need to spend a whole bunch of time tweaking (and re-tweaking) this behavior.

I'd rather just we use a different binding for expanding global abbrs. E.g. ctrl-space.

teresaejunior:
I'm more inclined to the whitelisting approach too, because if I forget to press Ctrl+Space after an abbr, it would defeat the purpose of setting aliases like `rm -i`, and I would need to keep using functions instead (which have the disadvantage of not showing explicitly what has been modified).

manfredlotz:
[...]

PR #6891, opened 2020-04-13
Add option to aliases to expand next command word like Bash

ariasuni:
## Description

In bash, you can make the alias expansion work after a command by putting a trailing space at the end of the alias. This is very useful for commands like sudo or [doas](https://man.openbsd.org/doas).

According to the man page `bash(1)`:
> If the last character of the alias value is a blank, then the next command word following the alias is also checked for alias expansion.

https://linuxhandbook.com/run-alias-as-sudo/

I also added an “obvious” flag (`--expand-next`) in addition to the compatibility behavior.

## TODOs:

### DIFF
--- a/share/functions/alias.fish
+++ b/share/functions/alias.fish
@@ -1,5 +1,5 @@
 function alias --description 'Creates a function wrapping a command'
-    set -l options h/help s/save
+    set -l options h/help s/save e/expand-next
     argparse -n alias --max-args=2 $options -- $argv
     or return
 
@@ -44,6 +44,9 @@ function alias --description 'Creates a function wrapping a command'
         return 1
     end
 
+    string trim --quiet --right $body
+    set -l has_trailing_space $status
+
     # Extract the first command from the body.
     printf '%s\n' $body | read -lt first_word body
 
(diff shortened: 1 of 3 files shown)


share/functions/alias.fish
```diff
@@ -58,7 +61,32 @@ function alias --description 'Creates a function wrapping a command'
...
+            if set -q \$argv
+                $prefix $first_word $body
+                return
+            end
+
+            set -l tmp (string split -m 1 ' ' -- \$argv)
+            set -l inner_name \$tmp[1]
+            set -l inner_body \$tmp[2]
+
+            for func in (alias)
```
krobelus, line 75: Since aliases are just simple functions, I wonder if makes sense to run whatever `functions $inner_name` prints to include functions, too. Probably not worth it.

zanchey:
> If the last character of the alias value is a blank, then the next command word following the alias is also checked for alias expansion.

In my view this is poor UI, and I'd prefer to avoid introducing it in fish even for compatibility purposes.

faho:
So, ignoring familiarity from bash et al, this is a super non-obvious interface. That an alias would even be expanded after another alias is a tad weird to begin with, but that it would only happen in some cases, and that that would only be if you had a trailing space? Yeah, nah.

Then it's also super fragile. If an alias does anything other than call "command thing --someoptions" it *will* break.

For instance your tests feature " alias expand1='echo ' ". That means after "expand1" other aliases will be expanded. That's okay. But the expand1 alias itself would be expanded incorrectly, because typically `echo` is our echo *builtin*! (call it `echo` instead of `expand1` and it'll feature an explicit "builtin")

Given that aliases *are functions*, there is no possible way we can properly guard against this.

Overall alias itself is a fairly bad interface (even the bash documentation tells you to use functions instead), and this is unfortunately way too non-obvious and leaky. What we should do is add global abbreviations instead, that will just actually expand in-place *while editing* - #5003.

Sorry, but I'm rejecting this!

ariasuni:
I thought a lot about it, and I thought I could try to expand after the first `--` if there is one, then try the second word of the command, then stops. But I would be very unreliable, except by taking all flags into account and knowing how the command would parse them. so I guess you’re right, this is going nowhere.

Still, I’d like to say that I only use aliases, because I wouldn’t like to see e.g. `exa -lg --group-directories-first` each time instead of `l`; it makes it much easier to parse visually the commands, especially when going through a lot of history.

Anyway, I really appreciate fish!

krobelus:
> Still, I’d like to say that I only use aliases, because I wouldn’t like to see e.g. `exa -lg --group-directories-first` each time instead of `l`

Use a script, which is just as good as an alias and will work in every shell and for other users (sudo), too:

```sh
$ cat /usr/local/bin/l
#!/bin/sh
exec exa -lg --group-directories-first "$@"
```

and to get the correct completions in fish:

```fish
complete -c l --wraps exa
```

Issue #6692, opened 2020-03-01
Unexpected behavior when using fzf in vim


yujinyuz:
fish: 3.1.0
Darwin jprdgy.local 19.3.0 Darwin Kernel Version 19.3.0: Thu Jan  9 20:58:23 PST 2020; root:xnu-6153.81.5~1/RELEASE_X86_64 x86_64
tmux-256color

There's a weird behavior when using `fzf` in vim. 

When I'm trying to open fzf inside vim, I can type the letters but the search does not start. As you can see in the asciinema, I'm trying to press <kbd>Esc</kbd> but a repeated `^[` is just appearing, so I have to press <kbd>Ctrl-c</kbd> an re-run the search again. I do several attempts until I can type and search.

https://asciinema.org/a/2PtASVMpGvamM7j9j4V4nncnn

This only happens when using fish `3.1.0`. I've tried downgrading to `3.0.2` and this doesn't happen.

https://asciinema.org/a/5E2L45ZPico5Yc9tJCDdkWnqV

mitchellhenke:
Same issue is happening for me on fish 3.1 and fzf version 0.20.0

Darwin MBP.localdomain 19.3.0 Darwin Kernel Version 19.3.0: Thu Jan  9 20:58:23 PST 2020; root:xnu-6153.81.5~1/RELEASE_X86_64 x86_64

ridiculousfish:
Sorry for the delay. Probably the same as #6624. If so we should probably take something for 3.1.1. Can you please confirm if [this workaround](https://github.com/fish-shell/fish-shell/issues/6624#issuecomment-593012204) addresses the problem for you?

Strayer:
Same problem here, the workaround worked for me in nvim 0.4.3, fzf 0.20.0 and fish 3.1.0

yujinyuz:
@Strayer I tried creating a file `~/.config/fish/functions/fzf.fish` with the following contents

```
function fzf; command fzf $argv; end
```

But I still encounter the issue with 3.1.0. Do you mind showing what you did? I might be doing something wrong here

Strayer:
I simply put that into my config.fish after loading fzf. I do have to admit that I was wrong though, it seems to work once or a few times and then stops working, no idea why :( I'm back on zsh for now, but will keep an eye on this issue.

mitchellhenke:
[...]

Issue #1362, opened 2014-03-24
Command substitution leaks processes if they expect input (command substitutions run in their own process group)


Undeterminant:
Example: `set -l var (cat)`

Expected: `$var` contains value of user input.

Actual: `$var` contains an empty string, and if you run `jobs`, a stopped but not reaped `cat` process lingers unless you run `kill -9` on it.

I understand that in this case, I could use `read`, but this error would still be a thing.

zanchey:
`fzf` recommends a workaround which may be helpful in some instances - see the [fzf README](https://github.com/junegunn/fzf#fish-shell) for an example.

paulcarey:
Rather than calling `$ vim (git ls-files | peco)` which fails for the reason described above, `xargs` can be used to the same effect.

     $ git ls-files | peco | xargs -I _ fish -c 'vim _ < /dev/tty'

I'd guess that the original issue with `set` could be worked around with a wrapper function that set a universal variable, but I haven't tried this.

jcelliott:
Another vote for finding a way to fix this. Is it something that can reasonably be changed?

ridiculousfish:
Oh sure, we just have to fix it.

jcelliott:
Any pointers on where to start looking?

ridiculousfish:
The key change is to run command substitutions in the foreground process group, and to properly handle stopped jobs when doing command substitutions .It would be interesting to investigate how bash handles this - what happens if you stop (e.g. ctrl-Z) a job in a command substitution?

JoshCheek:
Looks like it just kills it.

![bash-interrupt-bg](https://cloud.githubusercontent.com/assets/77495/6248384/52556960-b73c-11e4-9575-53c244921d02.gif)

zanchey:
It swallows Ctrl-Z, but running `kill -SIGSTOP` from another session just queues input until sent `SIGCONT`.

merlinthered:
I'm just bumping this, since it has become a problem for me.

sebastianmarkow:
Just wanted to share a workaround via `read(1)` which is IMHO better than dumping stdout to file (seen elsewhere).

A snippet from my interactive history search keymap:
~~~fish
    history | fzf --tiebreak=index --header="history" -q (commandline -b) | read -l fzf_last_select
    [ $fzf_last_select ]; and commandline -rb $fzf_last_select
~~~

[...]

PR #6893, opened 2020-04-13
Block TTY control signals in processes started from subshells

### LINKED ISSUE #6300: Command substitution in pipe to less may stop job
Sometimes (about every other attempt) the following command is immediately stopped (and sent to the background).

```fish
begin echo (cat /dev/null); end | less
```

Seems to be race condition because two processes are writing to the terminal. The job stops because of SIGTTOU. It never happens without the `begin ... end`or when adding a sleep:

```fish
begin echo (sleep .1; cat /dev/null); end | less
```

Related to at least #1362.

### LINKED ISSUE #6624: fish 3.1.0: __z --list | fzy results in tcsetattr error
fish --version 3.1.0 (now reverted to 3.0.2)

macos 10.14.6
 `echo $TERM` xterm-256color, using normal mac terminal, but happens the same in others

first reported against z https://github.com/jethrokuan/z/issues/86
The reason for why this was an issue for me is I'm using `alias z="cd (__z -l 2>&1 | fzy | sed 's/^[0-9,.]* *//')"` (maybe there is a better way)

behaviour changed from 3.0.2 to 3.1.0:
in short: `__z --list | fzf` results in `Failed to read /dev/tty` while `ls | fzf` works fine, `__z --list | rg llvm` works fine as well 

fzf and fzy show nearly the same behaviour/ only difference is that the output of fzy is `tcsetattr: Input/output error` and fzf is `Failed to read /dev/tty`

the behaviour seems to be timing sensitive, it fails most of the time though

I suspect this commit might have changed the behaviour 
https://github.com/fish-shell/fish-shell/commit/2a3677b38659835858c9d557e7d920630bfbcbb8

krobelus:
Same for processes started from event handlers.

This is consistent with the bash equivalent: from
https://www.gnu.org/software/bash/manual/html_node/Signals.html

	Commands run as a result of command substitution ignore the
	keyboard-generated job control signals SIGTTIN, SIGTTOU, and SIGTSTP.

Fixes #6300 (caused by SIGTTOU from subshell)
Fixes #6624 (caused by SIGTTIN from subshell)
Probably also #6692 which shows very similar symptoms to #6624, but I
have't been able to reproduce that.

## TODOs:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - A variable `fish_kill_signal` will be set to the signal that terminated the last foreground job, or `0` if the job exited normally.
 - On BSD systems, with the `-s` option, `fish_md5` does not use the given string, but `-s`. From now on the string is used.
 - Control-C no longer kills background jobs for which job control is disabled, matching POSIX semantics (#6828).
+- Commands run from subshells and event handlers block signals SIGTTIN, SIGTTOU and SIGTSTP, matching Bash's behavior (#6300, #6624).
 
 ### Syntax changes and new commands
 
--- a/src/exec.cpp
+++ b/src/exec.cpp
@@ -328,13 +328,20 @@ static void run_internal_process_or_short_circuit(parser_t &parser, const std::s
 }
 
 bool blocked_signals_for_job(const job_t &job, sigset_t *sigmask) {
+    bool changed = false;
+    if (job.wants_job_control() && (job.in_subshell() || job.from_event_handler())) {
+        sigaddset(sigmask, SIGTTIN);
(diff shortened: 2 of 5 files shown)


krobelus:
Ugh, this doesn't actually fix #6300, I'll take another look.
Also bash seems to be doing something differently, fish_test_helper does not report blocked signals in bash.

zanchey:
See also #1362, where fish handled the issue in a different way to bash

krobelus:
OK this does definitely not work since the issue is not specific to subshells, see the example in https://github.com/fish-shell/fish-shell/issues/6624#issuecomment-616827000

PR #6896, opened 2020-04-14
More dogfooding of `string split -f`

ammgws:
Continuing on from #6874

### DIFF
--- a/share/functions/__fish_crux_packages.fish
+++ b/share/functions/__fish_crux_packages.fish
@@ -1,4 +1,6 @@
 # a function to obtain a list of installed packages with CRUX pkgutils
 function __fish_crux_packages -d 'Obtain a list of installed packages'
-    pkginfo -i | cut -d' ' -f1
+    pkginfo -i | while read -l line
+        string split -f1 ' ' -- $line
+    end
 end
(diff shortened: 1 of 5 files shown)


share/functions/__fish_crux_packages.fish
```diff
@@ -1,4 +1,6 @@
 # a function to obtain a list of installed packages with CRUX pkgutils
 function __fish_crux_packages -d 'Obtain a list of installed packages'
-    pkginfo -i | cut -d' ' -f1
+    pkginfo -i | while read -l line
```
krobelus, line 3: The `cut` version was quite short and sweet. Isn't it possible to do `string split -f1 ' ' -- (pkginfo -i)`?

ammgws: Due to how `string split` functions the fish version gets a bit longer, but hopefully it isn't an issue if the aim of fish is not to be as short as possible.

```sh
>printf "package1 description1\npackage2 description2\n" | cut -d' ' -f1
package1
package2

>printf "package1 description1\npackage2 description2\n" | string split -f1 ' '
package1

>printf "package1 description1\npackage2 description2\n" | while read -l line
    string split -f1 ' ' -- $line
 end
package1
package2
```

krobelus: I think it's surprising that `string split` only accepts one argument:
I would expect `string split -f1 ' ' 'a b' 'c d'` to print `a` and `c`.

ammgws: Yeah I came across that which is why I ended up going with the `read`.

ammgws:
Addressed with #6911

PR #6897, opened 2020-04-14
Run `fish_indent` on all files

ammgws:
Caught a couple of strays that weren't processed last time this was done.

Is the formatting for that pipe into awk  in `make.fish` working as intended?


faho:
Merged, thanks!

PR #6906, opened 2020-04-16
Activate untracked status in hg prompt

jeanprey:
Correction of a typo that prevented the display of untracked status on the detailed prompt.

## Description

The variable declared on line 65 and the one used on line 75 should be the same.


faho:
Yeah, that's correct. Good catch!

PR #6907, opened 2020-04-16
Change vi-mode tilde to toggle character case

joelakuhn:
This updates the behavior of tilde to match the behavior found in vim.
In vim, tilde toggles the case of the character under the cursor and
advances one character. In visual mode, the case of each selected
character is toggled and the cursor position is unmodified. In fish,
tilde capitalizes the current letter and advances one word. There is
no current tilde command for visual mode.

## Description

This patch adds the readline commands `togglecase-letter` and `togglecase-selection` to accomplish this behavior. It also adds a visual-mode '~' binding to call `togglecase-selection`


zanchey:
It should rebase to master ok

PR #6908, opened 2020-04-17
Change vi-mode tilde to toggle character case

joelakuhn:
This updates the behavior of tilde to match the behavior found in vim.
In vim, tilde toggles the case of the character under the cursor and
advances one character. In visual mode, the case of each selected
character is toggled, the cursor position moves to the beginning of
the selection, and the mode is changed to normal. In fish, tilde
capitalizes the current letter and advances one word. There is no
current tilde command for visual mode in fish.

## Description

This patch adds the readline commands `togglecase-letter` and
`togglecase-selection` to match the behavior of vim more closely. The
only difference is that in visual mode, the cursor is not modified.
Modifying the cursor in visual mode would require either moving it in
`togglecase-selection`, which seems outside its scope or adding
something like a `move-to-selection-start` readline command.

### DIFF
--- a/src/reader.cpp
+++ b/src/reader.cpp
@@ -3219,6 +3219,72 @@ void reader_data_t::handle_readline_command(readline_cmd_t c, readline_loop_stat
             }
             break;
         }
+        case rl::togglecase_char: {
+            editable_line_t *el = active_edit_line();
+            size_t buff_pos = el->position();
+
+            // Check that the cursor is on a character
+            if (buff_pos < el->size()) {
+                wchar_t chr = el->text().at(buff_pos);
+                wcstring replacement;
+
+                // Toggle the case of the current character
+                bool make_uppercase = iswlower(chr);
+                if (make_uppercase) {
+                    chr = towupper(chr);
+                } else {
+                    chr = tolower(chr);
+                }
+
+                replacement.push_back(chr);
+                el->replace_substring(buff_pos, (size_t)1, std::move(replacement));
+
+                // Restore the buffer position since replace_substring moves
+                // the buffer position ahead of the replaced text.
+                update_buff_pos(el, buff_pos);
+
+                command_line_changed(el);
+                super_highlight_me_plenty();
+                reader_repaint_needed();
+            }
+            break;
+        }
+        case rl::togglecase_selection: {
+            editable_line_t *el = active_edit_line();
+
+            // Check that we have an active selection and get the bounds.
+            size_t start, len;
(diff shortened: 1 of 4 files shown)


src/reader.cpp
```diff
@@ -3219,6 +3219,72 @@ void reader_data_t::handle_readline_command(readline_cmd_t c, readline_loop_stat
...
+                } else {
+                    chr = tolower(chr);
+                }
+
+                replacement.push_back(chr);
+                el->replace_substring(buff_pos, (size_t)1, std::move(replacement));
+
+                // Restore the buffer position since replace_substring moves
+                // the buffer position ahead of the replaced text.
+                update_buff_pos(el, buff_pos);
```
krobelus, line 3244: Maybe it makes sense to remove this and the `forward-char` in the binding

joelakuhn: I thought about that, but I felt it violates the principle of least surprise. In this case, the cursor should advance, but if a user want's to bind to togglecase-letter, they probably wouldn't assume it also advances the cursor.

krobelus: Ok, sounds reasonable.

krobelus:
Nice contribution, thank you!

joelakuhn:
Awesome! That's been slightly bugging me for like 8 years and I finally took the time to fix it. There are other vi-mode enhancements like adding matching brace movement and inner and outer brace text objects I'd like to work on if that would be welcome.

krobelus:
> Awesome! That's been slightly bugging me for like 8 years and I finally took the time to fix it. There are other vi-mode enhancements like adding matching brace movement and inner and outer brace text objects I'd like to work on if that would be welcome.

Sure, bindings for `%` and `cib` would make a lot of sense. It's never too late to scratch those itches ;)
There could also be a "job" text object, delimited by any of `; \n & && || #`, or "process" which is delimited by `|`, too.

Of course, heavy duty editing can always be done with `\ev` or `\ee`, but using fish as line editor interacts nicely with autosuggestions and completions.

We might also want to change the default binding of `u` to the new `undo` readline command.
I haven't done this because there is already a  `bind --preset u history-search-backward` and I wasn't using Vi mode myself, so I don't know how useful the existing binding is (`k` does essentially the same thing, right?). Anyway, `fish_hybrid_key_bindings` seems really nice.

I find the [kakoune](http://kakoune.org/) editing model more intuitive, it's like vi but movements like `w` and `W` start or extend the visual selection in addition to moving the cursor. It should be possible to write such bindings already but I doubt anyone has tried yet.

PR #6913, opened 2020-04-18
systemctl completions: Use --plain where appropriate

### LINKED ISSUE #6740: 'systemctl start' completions miss certain unit files
I noticed that certain systemd units do not get autocompleted for `systemctl start` while they do for `systemctl status`. I tried to find out why but I'm a bit confused regarding the completions for systemd.

I think it boils down to what is returned here: https://github.com/fish-shell/fish-shell/blob/master/share/functions/__fish_systemctl.fish#L83 (or in line 77)

This returns nothing for me, while line 77 returns most of the units but not the ones I'm missing. I tried playing with the `--state` flag but as long as I don't remove it completely, the missed unit does not show up.

This is what the `status` gives me about that unit:

```
> systemctl status nessusd.service 
● nessusd.service - The Nessus Vulnerability Scanner
   Loaded: loaded (/usr/lib/systemd/system/nessusd.service; disabled; vendor preset: disabled)
   Active: inactive (dead)
```

Does someone know what state flag is missing here or why this does not show up?

joukewitteveen:
The --plain flag drops the '●'-glyph and generally makes the output more
suitable for automated processing.

## Description

This undoes db7dfddc3be3ba2fd48bfcebecca4c1179c10901 and fixes #6740 in another way.

The changes anticipate systemd/systemd#15458, but are backward compatible.

### DIFF
--- a/share/functions/__fish_systemctl.fish
+++ b/share/functions/__fish_systemctl.fish
@@ -74,12 +74,11 @@ function __fish_systemctl --description 'Call systemctl with some options from t
     # Output looks like
     # systemd-tmpfiles-clean.timer      [more whitespace] loaded    active     waiting   Daily Cleanup[...]
     # Use the last part as the description.
-    # Note that in some cases this prints a "●" or "*" (with C locale) marker at the beginning of the line. We have to remove it.
-    systemctl --no-legend --no-pager --all list-units $passflags | string trim -c ' *●' | string replace -r "(?: +(\S+)){4}" \t'$1'
+    systemctl --full --no-legend --no-pager --plain --all list-units $passflags | string replace -r "(?: +(\S+)){4}" \t'$1'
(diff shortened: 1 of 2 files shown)


faho:
Okay, so what happened here was that originally --no-legend suppressed the bullet, then that was unintentionally broken, readded and now changed so that "--plain" suppresses the bullet.

Sure, that'll work.

joukewitteveen:
For completeness: `--plain` always suppressed the bullet, so these changes are backward-compatible. What changed is that with `--plain`, the empty first column (that would hold bullets) is now hidden. This hiding was previously done with `--no-legend`.

zanchey:
Is this suitable for 3.1.1 instead of the current fix? It seems to be to me.

faho:
It is, yes.

zanchey:
Cherry-picked to 3.1.1 as 1643cf40f

PR #6914, opened 2020-04-18
Fix underflow in `commandline` jump functions

joelakuhn:
This patch fixes an underflow in the jump family of readline commands when called via `commandline -f` outside of a bind context such as `commandline -f backward-jump`. To reproduce, run that command at a prompt and the shell will crash with a buffer underflow.

## Description

This happens because the jump commands have non-zero arity, requiring a character event to be pushed on the function args stack. Pushing the character event is handled in `function_push_args`, called by `inputter_t::mapping_execute`, which checks the arity of the function and enqueues the required number of charcter events. However, `builtin_commandline` calls `reader_queue_ch`, which in turn calls `inputter_t::queue_ch`, which immediately enqueues the readline event without calling `function_push_args`, so the character event is never pushed on the arg stack.

This patch adds a check in inputter_t::queue_ch which checks if the character event is a readline event, and if so, calls `function_push_args`.

### DIFF
--- a/src/input.cpp
+++ b/src/input.cpp
@@ -433,7 +433,12 @@ bool inputter_t::mapping_is_match(const input_mapping_t &m) {
     return true;
 }
 
-void inputter_t::queue_ch(const char_event_t &ch) { event_queue_.push_back(ch); }
+void inputter_t::queue_ch(const char_event_t &ch) {
+    if (ch.is_readline()) {
+        function_push_args(ch.get_readline());
+    }
+    event_queue_.push_back(ch);
+}
 
 void inputter_t::push_front(const char_event_t &ch) { event_queue_.push_front(ch); }
 


joelakuhn:
This also fixes an issue that doesn't exist yet, being that commandline jump functions cannot receive input events from other commandline functions. There is a FIXME in `inputter_t::mapping_execute` that suggests that there should be a way for commandline functions to add character events to the input queue for other functions. This along with non-zero arity readline functions correctly receiving their arguments would allow for non-interactive jump bindings like `commandline -f backward-jump; commandline -f i '('`.

krobelus:
Basically all `commandline` commands are useless outside an interactive session,
I think `commandline` should just return an error when not interactive to communicate this.

> allow for non-interactive jump bindings like commandline -f backward-jump; commandline -i '('

Just to be clear, bindings are always interactive. I think this doesn't work yet, because `backward-jump` reads from the keyboard before the `(` is inserted directly into the commandline.
To support bindings like this there could probably be a `--fake-keys` or `--insert-with-bindings` option to add character events to the input queue.

joelakuhn:
By non-interactive, I just meant jump bindings that don't require reading from the keyboard. A new `--fake-keys` binding would be great. In my local version I called it --insert-key. That, along with this patch, enables such bindings. Without this patch or something like it though, the inserted key event will never be read by `backward-jump` because it won't call `function_push_args` to ask for it.

ridiculousfish:
Thank you!!

PR #6917, opened 2020-04-18
fish_hg_prompt should return 1 when not in a mercurial directory

asomers:
## Description

fish_hg_prompt should return 1 if the PWD is not part of a Mercurial working copy.  Returning 0 prevents the calling script from correctly handling the error.  For example, fish_vcs_prompt will never call fish_svn_prompt on a system that has mercurial installed.

## TODOs:


krobelus:
Good catch!

PR #6929, opened 2020-04-21
Fix string split docs

ammgws:
Sorry didn't notice this before.


krobelus:
Me neither, thanks again!

PR #6940, opened 2020-04-24
fish completions for arguments in options part of man page

slavaGanzin:
## Description

There are commands like GNU parallel that use options section of man page as reference for arguments supplied to command. Arguments do not start with dashes, so currently create_manpage_completions.py ignores them altogether. 

This PR created to start discussion.

I use this script to generate automatic completions from man generated by help2man: [gencomplete](https://github.com/slavaGanzin/fish-gencomplete/blob/master/functions/gencomplete.fish). I want  create_manpage_completions.py to be more relaxed, so I have no need to write completions for my own commands by hand.

## TODOs:

### DIFF
--- a/share/tools/create_manpage_completions.py
+++ b/share/tools/create_manpage_completions.py
@@ -160,6 +160,8 @@ def built_command(options, description):
         elif option.startswith("-") and len(option) > 2:
             # Old style long option (-recursive)
             fish_options.append("-o " + fish_escape_single_quote(option[1:]))
+        else:
+            fish_options.append("-a " + fish_escape_single_quote(option))
 
     # Determine which options are new (not already in existing_options)
     # Then add those to the existing options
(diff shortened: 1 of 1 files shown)


faho:
Sorry, but no. This generates completions like

```fish
complete -c xterm -o class -a string --description 'This option allows you to override \\*n\'s resource class.'

# or

complete -c udevadm -s q -l query -a TYPE --description 'Query the database for the specified type of device data.'
complete -c udevadm -s p -l path -a DEVPATH --description 'The /sys path of the device to query, e. g.  [/sys]/class/block/sda.'
complete -c udevadm -s n -l name -a FILE --description 'The name of the device node or a symlink to query, e. g.  [/dev]/sda.'

```

Which just provides a useless filler argument, or

```fish
complete -c git -s c -a '<name>' -a '<value>' --description 'Pass a configuration parameter to the command.'
```

where "<name>" is even a syntax error.

Even if you were to fix the errors, what are you gonna provide that's not a useless filler string? Manpages simply aren't a stringent enough format to give useful information here.

We have an open issue about argument "hints" that might be filled with this, but that feature would have to be added first.

slavaGanzin:
I thought it could be pitfalls, but I don't expect so obvious ones. 
Thanks for your answer.

PR #6947, opened 2020-04-26
fix pushd's behavior with invalid arguments

LawAbidingCactus:
## Description

`pushd` currently adds `$PWD` to `$dirstack` without checking whether the directory change was successful. For example, `pushd ~/.config/fish/config.fish` from `~` will add prepend `~` to `$dirstack`, even though pushd was given an invalid argument. This also applies to directories the user may not have permission to `cd` into. The change in question fixes this.

## TODOs:

### DIFF
--- a/share/functions/pushd.fish
+++ b/share/functions/pushd.fish
@@ -76,6 +76,6 @@ function pushd --description 'Push directory to stack'
     end
 
     # argv[1] is a directory
-    set -g -p dirstack $PWD
-    cd $argv[1]
+    set -l old_pwd $PWD
+    cd $argv[1]; and set -g -p dirstack $old_pwd
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - On BSD systems, with the `-s` option, `fish_md5` does not use the given string, but `-s`. From now on the string is used.
 - Control-C no longer kills background jobs for which job control is disabled, matching POSIX semantics (#6828).
 - Improve Gradle completion
+- Fixed `pushd`'s behavior with respect to the directory stack when given an invalid argument
 
 ### Syntax changes and new commands
 


share/functions/pushd.fish
```diff
@@ -76,6 +76,10 @@ function pushd --description 'Push directory to stack'
     end
 
     # argv[1] is a directory
-    set -g -p dirstack $PWD
-    cd $argv[1]
+    if fish -c "cd $argv[1]" >/dev/null 2>&1
```
faho, line 79: That's a NAK.

We should not call fish from fish except in special circumstances.

Just check the status of `cd` before adding it to the stack

LawAbidingCactus: Yeah, sorry, I realized this could be done by adding an `and` just now. Resolved.

share/functions/pushd.fish
```diff
@@ -76,6 +76,6 @@ function pushd --description 'Push directory to stack'
     end
 
     # argv[1] is a directory
-    set -g -p dirstack $PWD
-    cd $argv[1]
+    cd $argv[1];
```
faho, line 79: Semicolon is unnecessary if you use a linebreak. `fish_indent` would remove it.

LawAbidingCactus: Ah, understood.

share/functions/pushd.fish
```diff
@@ -76,6 +76,6 @@ function pushd --description 'Push directory to stack'
     end
 
     # argv[1] is a directory
-    set -g -p dirstack $PWD
     cd $argv[1]
+    and set -g -p dirstack $PWD
```
faho, line 80: Ummmm... sorry. This wants the directory *before* changing.

So you'll have to store $PWD, and only add it to dirstack if the cd succeeds.

LawAbidingCactus: whoops, fixed

faho:
Yeah, that'll do. Thanks!

Issue #6834, opened 2020-03-30
The paste is wrong when setting bind.


hotoolong:
The following settings were added to the config file
```
bind \e\] nextd-or-forward-word
```
I reread the config file and pasted the string and got the following
```
200~configs/fish/config.201~fish
```
The same thing did not happen when I removed the bind setting.

I have not been able to confirm if this is being replicated on other machines. 
Is there anyone who is experiencing the same thing?

Fish version: 3.1.0_1
Darwin hotoolong.local 18.7.0 Darwin Kernel Version 18.7.0: Tue Aug 20 16:57:14 PDT 2019; root:xnu-4903.271.2~2/RELEASE_X86_64 x86_64
screen-256color

faho:
The "200~" and "201~" bits are parts of the bracketed paste start/end sequences. That's a terminal function that fish turns on to distinguish between pasted text and manually entered text (so it can e.g. not execute multiline input immediately).

What is your terminal? What did you paste exactly - it appears to be the string "configs/fish/config." and then possibly a "fish".

What does `bind | grep paste` say?

hotoolong:
The copied string is `configs/fish/config.fish.`
The results of "bind | grep paste" are listed below.

```
bind --preset \cv fish_clipboard_paste
bind --preset -m paste \e\[200\~ __fish_start_bracketed_paste
bind --preset -M insert -m paste \e\[200\~ __fish_start_bracketed_paste
bind --preset -M paste \e\[201\~ __fish_stop_bracketed_paste
bind --preset -M paste '' self-insert
bind --preset -M paste \r commandline\ -i\ \\n
bind --preset -M paste \' __fish_commandline_insert_escaped\ \\\'\ \$__fish_paste_quoted
bind --preset -M paste \\ __fish_commandline_insert_escaped\ \\\\\ \$__fish_paste_quoted
bind --preset -M paste ' ' 'if set -q __fish_paste_quoted[1]; or string length -q -- (commandline -c); commandline -i " "; end'
```

faho:
Those are the correct bindings, that's the string I assumed.

What terminal are you using? It appears to be broken.

hotoolong:
I use iTerm2.
I had a similar problem when I checked with Hyper.

faho:
>echo $TERM is screen-256color

So you are using *screen* or *tmux* inside of iTerm2/Hyper?

hotoolong:
Earlier, I tried it without using tmux and had a similar problem.

[...]

PR #6949, opened 2020-04-26
Reject bindings that conflict with already bound bindings

faho:
In e.g. #6834, the reporter bound `\e\[` to something. That *can't* work.

At best it would effectively disable all bindings that have it as a prefix (including all arrow keys!).

So whenever a binding is added, this goes through all bindings and checks if anything prefixes the new sequence or if the new sequence is a prefix of it. If it is it prints an error without adding the binding. That means that:

```fish
bind \e\[ something
bind -k left
```

will fail without binding left, and doing it the other way around it will fail without binding `\e\[`.

One exception is if the prefix ends in an escape (like just `\e`). In that case we should disambiguate it with the escape timeout, so there is no problem. (also the default `""` binding and different modes are handled)

This is one solution to conflicting bindings - erroring out. Other solutions are:

- If there is an ambiguity run a timeout (e.g. if `abc` and `abcdef` are bound, wait after `c` for `d`)
- If a conflict happens remove the earlier bindings

I don't think removing bindings is what you'd want especially in the `\e\[` case (we could even blacklist that one because it makes no sense to bind), and a timeout is a larger change, and one I'm not sold on.

## TODOs:

I would appreciate some testing, since I could have entirely overlooked some cases. It's also likely this triggers for people who are otherwise happy (one of those "your config was technically always broken" situations).

### DIFF
--- a/src/builtin_bind.cpp
+++ b/src/builtin_bind.cpp
@@ -126,6 +126,61 @@ void builtin_bind_t::list(const wchar_t *bind_mode, bool user, io_streams_t &str
     }
 }
 
+/// Check if the given sequence conflicts with one that was already bound.
+/// Returns true if a conflict was found.
+bool builtin_bind_t::check_conflicts(const wcstring &seq, const wchar_t *bind_mode, bool user, io_streams_t &streams) {
+    const std::vector<input_mapping_name_t> lst = input_mappings_->get_names(user);
+
+    // The generic binding is allowed to be a prefix, because it's only used if nothing else matched.
+    if (seq.empty()) return false;
+    for (const input_mapping_name_t &binding : lst) {
+        if (bind_mode && bind_mode != binding.mode) {
+            continue;
+        }
+        auto &bseq = binding.seq;
+        if (bseq.empty()) continue;
+        // We're allowed to be equal because that just overrides it.
+        if (bseq == seq) continue;
+
+        // This is ugly - C++14 guarantees mismatch won't overrun the second element, but we're C++11.
+        if (bseq.size() < seq.size()) {
+            auto mismatch = std::mismatch(bseq.begin(), bseq.end(), seq.begin());
+            if (mismatch.first == bseq.end()) {
+                // Mismatches on escape are allowed because we treat that special (read timeout and such)
+                if (mismatch.first - bseq.begin() > 0 && *(mismatch.first - 1) == L'\e') continue;
+                streams.err.append_format(_(L"%ls: Sequence '%ls' has already bound sequence '%ls' as a prefix\n"), L"bind",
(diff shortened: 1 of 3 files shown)


zanchey:
What about a warning rather than a hard error? That way if the offending binding is removed then the shell will respond as expected.

exploide:
Currently, there is the default keybinding `bind --preset $argv \ct transpose-chars`. A lot of people use `fzf` which comes with own bindings, e.g. `bind \ct fzf-file-widget`.

What would happen in this case? Would loading the fzf fish config file fail because the binding is already occupied?

I like the idea of handling bind conflicts to avoid problems. But I also think loading such "extensions" should work out of the box. Don't know how to solve this best, but I could imagine introducing a force flag to bind that extension developers could use when deliberately overriding default bindings. Users who mess with their custom config would still be alarmed due to the warning/error (at least until they start thinking using the force flag themselves is a good idea).

faho:
>What about a warning rather than a hard error? That way if the offending binding is removed then the shell will respond as expected.

The problem there is that binding `\e\[` would still break *everything* - if you don't know how to use other movement keys, that's hard to recover from.

>What would happen in this case? Would loading the fzf fish config file fail because the binding is already occupied?

@exploide Straight *overwriting* still works. The old binding is removed and the new one inserted instead.

I think that should solve that usecase?

faho:
>What would happen in this case? Would loading the fzf fish config file fail because the binding is already occupied?

Okay, I have now executed `fzf_key_bindings` in both emacs and vi mode, and it returned no errors.

So overwriting works correctly. Now, like I said above, it's possible people will see errors who haven't seen errors before. I consider that a case of the config being silently broken *already*. It never worked *properly*. The issue is that they might have been happy with how it worked - e.g. they prefixed a binding they don't use.

I *think* that won't happen too often - most bindings are one "key", and the way escape sequences work they aren't typically ambiguous.

I think binding something like `jk` to `cancel` in vi-mode should also work (about as well as it ever did), because it *inserts* j via the default, and that only kicks in once it can't be the specific binding (e.g. typing "jo" triggers insertion for "j" and "o"), and some quick testing seems to support that.

ridiculousfish:
The input machinery is supposed to try longer sequences before shorter sequences, so binding `\e\[` should actually be OK. [Here's where that lives](https://github.com/fish-shell/fish-shell/blob/4116829292ddcfe21da86654b0a6a2bd765816b9/src/input.cpp#L222).

The problem is that this logic doesn't cross between preset and user-defined mappings - but it could. What would you think about extending that logic so that it includes presets? So if one mapping is a prefix of another, always try the longer mapping first. This would allow the arrow keys to keep working.

I implemented this [here](https://github.com/ridiculousfish/fish-shell/commit/46216d06c8591069516db1386975682441c94415), it allows the arrow bindings to keep working.

faho:
Yeah, making it work makes much more sense than disallowing, I think.

So there's no need for this PR after all!

PR #6958, opened 2020-04-28
Fix issue if md5sum is used instead of md5

Mazorius:
## Description

Both have a different output which results in different array sized

`md5sum` return `hash`
`md5` return `description = hash`

The completion now checks for array size after split to support both cases.

### DIFF
--- a/share/completions/gradle.fish
+++ b/share/completions/gradle.fish
@@ -7,16 +7,13 @@ end
 
 function __fish_gradle_create_completion_cache_file
     # Set up cache directory
-    if command test -z $XDG_CACHE_HOME
+    if test -z "$XDG_CACHE_HOME"
         set XDG_CACHE_HOME $HOME/.cache
     end
-    mkdir -m 700 -p $XDG_CACHE_HOME/gradle-completions
+    mkdir -m 700 -p "$XDG_CACHE_HOME/gradle-completions"
 
-    set -l md5Hash (__fish_md5 -s $argv[1] | string split ' = ')
-    if command test (count $md5Hash) -gt 1
-        set -l md5Hash $md5Hash[2]
-    end
-    string trim -- $XDG_CACHE_HOME/gradle-completions/$md5Hash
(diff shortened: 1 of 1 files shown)


share/completions/gradle.fish
```diff
@@ -7,12 +7,16 @@ end
 
 function __fish_gradle_create_completion_cache_file
     # Set up cache directory
-    if test -z $XDG_CACHE_HOME
+    if command test -z $XDG_CACHE_HOME
```
faho, line 10: No thank you. We ship a test *builtin*, and calling that is both faster and not reliant on the operating system `test`. It's also, you know, shorter.

faho: This *should* be quoted tho - `if test -z "$XDG_CACHE_HOME"`

share/completions/gradle.fish
```diff
@@ -7,12 +7,16 @@ end
...
         set XDG_CACHE_HOME $HOME/.cache
     end
     mkdir -m 700 -p $XDG_CACHE_HOME/gradle-completions
 
-    string trim -- $XDG_CACHE_HOME/gradle-completions/(__fish_md5 -s $argv[1] | string split ' = ')[2]
+    set -l md5Hash (__fish_md5 -s $argv[1] | string split ' = ')
```
faho, line 15: There are multiple nicer alternatives here:

- Use an index of `[-1]` on the command substitution
- Use `string replace` to remove everything up to and including ` = ` - `string replace -r '.* = ' ''`

Mazorius: Changed it to string replace. Thanks!

share/completions/gradle.fish
Mazorius: Removed command infront of `test` and quoted `$XDG_CACHE_HOME`

zanchey:
Side note: the checks are failing because of an unrelated error which is fixed in master, rebasing should be fine.

Mazorius:
> Side note: the checks are failing because of an unrelated error which is fixed in master, rebasing should be fine.

Rebased to master

faho:
Merged, thanks!

PR #6964, opened 2020-05-02
Make `fish_key_reader` exit normally

charego:
The `--help` and `--version` options were returning status 1 instead of 0.

This aligns the behavior of `fish`, `fish_indent`, and `fish_key_reader`.

Also the file descriptor passed to `print_help` was wrong.

### DIFF
--- a/src/fish_key_reader.cpp
+++ b/src/fish_key_reader.cpp
@@ -284,13 +284,12 @@ static bool parse_flags(int argc, char **argv, bool *continuous_mode) {
                 break;
             }
             case 'h': {
-                print_help("fish_key_reader", 0);
-                error = true;
-                break;
+                print_help("fish_key_reader", 1);
+                exit(0);
             }
             case 'v': {
-                std::fwprintf(stdout, L"%s\n", get_fish_version());
-                return false;
+                std::fwprintf(stdout, _(L"%ls, version %s\n"), program_name, get_fish_version());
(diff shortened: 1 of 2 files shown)


src/fish_key_reader.cpp
```diff
@@ -284,13 +284,12 @@ static bool parse_flags(int argc, char **argv, bool *continuous_mode) {
...
-                print_help("fish_key_reader", 0);
-                error = true;
-                break;
+                print_help("fish_key_reader", 1);
+                exit(0);
             }
             case 'v': {
-                std::fwprintf(stdout, L"%s\n", get_fish_version());
-                return false;
+                std::fwprintf(stderr, _(L"%ls, version %s\n"), program_name, get_fish_version());
```
zanchey, line 291: Probably this should go to `stdout` as before.

charego: Thanks for the review. I copied that snippet from [fish_indent.cpp](https://github.com/fish-shell/fish-shell/blob/3.1.2/src/fish_indent.cpp#L537). Should it be changed to stdout as well?

krobelus: yes the version flags should all print to stdout

charego: They print to stdout now.

zanchey:
Thanks!

PR #6965, opened 2020-05-03
Add `fish_trace` to special variable completions

charego:
Adding `fish_trace` to https://github.com/fish-shell/fish-shell/commit/261198aa3ee21ed03972528301268a5de4660abd. All the special variables [listed here](https://fishshell.com/docs/3.1/index.html#special-variables) are included except `umask` but that should be set using the function of the same name.


zanchey:
Thanks!

PR #6968, opened 2020-05-03
Add optional argument to force fish_vi_cursor to allow a given $TERM

LawAbidingCactus:
## Description

This adds a check to `fish_vi_cursor` that compares `$argv[1]` against `$TERM`, enabling cursor shape setting if it matches. This is useful for people with nonstandard terminal emulation setups that support the relevant termcaps but aren't acknowledged by `fish_vi_cursor` (term-in-terms with tmux come to mind). This is vastly simpler than attempting to account for the enormous ecosystem of terminal emulation implementations and configurations, which quickly becomes untenable. 

## TODOs:


faho:
I don't think whitelisting a particular $TERM is all that helpful, given that it's not necessary or sufficient - usually it's "xterm-whatever".

So just adding a way to force it to skip checks would work better, and since it's often called from fish_vi_key_bindings an option won't work that well, so how about an environment variable called `fish_vi_force_cursor` or so.

If people want to check $TERM, they can then set that variable depending on $TERM.

LawAbidingCactus:
Yeah, that makes sense-- done. I've also removed the flag for forcing cursor shapes on iterm, since that's redundant now.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 - Ctrl-C no longer kills background jobs for which job control is disabled, matching POSIX semantics (#6828).
 - Improve Gradle completion
 - Fixed `pushd`'s behavior with respect to the directory stack when given an invalid argument
+- A new environment variable, `$fish_force_vi_cursor`, has been added. This can be set to force `fish_vi_cursor` to change the cursor shape in vi mode.
```
krobelus, line 11: Maybe drop `environment`, this variable doesn't need to be exported in most scenarios. Also the motivation could be extended by adding something like "without regard to your terminal" at the end.

share/functions/fish_vi_cursor.fish
```diff
@@ -10,51 +10,51 @@ function fish_vi_cursor -d 'Set cursor shape for different vi modes'
         return
     end
 
-    # Emacs Makes All Cursors Suck
-    if set -q INSIDE_EMACS
-        return
-    end
+    # If this variable is set, skip all checks
+    if test -z "$fish_force_vi_cursor"
```
krobelus, line 14: We usually use something like `not set -q fish_force_vi_cursor`, like for the git prompt variables.

LawAbidingCactus:
Makes sense; I don't have a particular preference here.

CHANGELOG.md
LawAbidingCactus: Done.

share/functions/fish_vi_cursor.fish
LawAbidingCactus: Done.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 - Ctrl-C no longer kills background jobs for which job control is disabled, matching POSIX semantics (#6828).
 - Improve Gradle completion
 - Fixed `pushd`'s behavior with respect to the directory stack when given an invalid argument
+- A new variable, `$fish_force_vi_cursor`, has been added. This can be set to force `fish_vi_cursor` to attempt changing the cursor shape in vi mode, regardless of terminal.
```
krobelus, line 11: Now the name here and in docs is out of sync.

LawAbidingCactus: Fixed; also added a note about the option removal.

krobelus:
> I've also removed the flag for forcing cursor shapes on iterm, since that's redundant now (or perhaps it should just be deprecated? Though I didn't find anyone using it after a quick github search).

Could be fine since this is rarely used and only cosmetic a feature? @faho @zanchey

zanchey:
`--force-iterm` was never documented, although it seems like a cheap feature to keep working.

LawAbidingCactus:
I've just deprecated `--force-iterm` for the time being, then; I'll remove it in a future pull request (just so `fish_vi_cursor.fish` stays a bit cleaner).

krobelus:
Merged as f71737e58ab489d9e4030658420fd58c50b0cf57, thanks and sorry for the delay.

PR #6982, opened 2020-05-07
Document addition of `string split --fields`

ammgws:
Noticed that this was missing from the changelog


zanchey:
Awesome, thanks.

PR #6983, opened 2020-05-07
Update /etc/config.fish to use current syntax

neersighted:
The long option usage of `status` is deprecated and to be removed by fish 3.0.


zanchey:
Yes, that makes sense, thanks.

PR #6989, opened 2020-05-10
Add some missing journalctl completions

atanunq:
## Description

Added some missing completions for `journalctl`. Descriptions were derived from its [man page](https://www.freedesktop.org/software/systemd/man/journalctl.html).

## TODOs:


akiirui:
I found still have some options missing, e.g.:

```
--vacuum-size=BYTES     Reduce disk usage below specified size
--vacuum-files=INT      Leave only the specified number of journal files
--vacuum-time=TIME      Remove journal files older than specified time
```

and more.

atanunq:
Yes, I did skip some that were in the from `$option=$value`. I am not sure how to do those. 
Should only the first half be autocompleted? Include the equals sign? Also, is it necessary to add suggestions for the value? I saw a mixture of those options across different completion files.

Thank you for your feedback, any help is appreciated!

akiirui:
> Should only the first half be autocompleted? Also, is it necessary to add suggestions for the value?

I think complete only the `$option` is fine. because `$value` is provided by the user, hard to complete his want. add `BYTES, INT, TIME` to the `--description` is fine.

> Include the equals sign?

Include or not, both work.

Thnaks for your work ~

atanunq:
I think this time I covered all of them. Also, I added argument lists for some of them, based on the original bash completions.
I believe the functions at the start of the file are outdated and there are better ways to achieve their goal, but I don't want to break anything by touching them.

share/completions/journalctl.fish
```diff
@@ -34,7 +34,7 @@ complete -c journalctl -f -s a -l all -d 'Show all fields in full'
 complete -c journalctl -f -s f -l follow -d 'Show live tail of entries'
 complete -c journalctl -f -s n -l lines -d 'Controls the number of journal lines'
 complete -c journalctl -f -l no-tail -d 'Show all lines, even in follow mode'
-complete -c journalctl -f -s o -l output -d 'Controls the formatting' -xa 'short short-monotonic verbose export json json-pretty json-sse cat'
+complete -c journalctl -f -s o -l output -d 'Controls the formatting' -xa '(journalctl --output=help | while read style; echo -e "$style\t"; end)'
```
krobelus, line 37: You can do `(printf %s\t\n (journalctl --output=help))`.

Using `read` like this sets a global variable by default, or would even update a universal `style` variable if such exists.
This can be remedied by using `read --local`.

atanunq: Fantastic, thank you!

krobelus:
Merged, thanks!

PR #7001, opened 2020-05-13
completions/set:  add --path and --unpath

akiirui:
## Description

Add missing options:
```
--path causes the specified variable to be treated as a path variable, meaning it will automatically be split on colons, and joined using colons when quoted (echo "$PATH") or exported.
--unpath causes the specified variable to not be treated as a path variable. Variables with a name ending in "PATH" are automatically path variables, so this can be used to treat such a variable normally.
```

Fixes issue #

## TODOs:


ridiculousfish:
Thanks!

Issue #4183, opened 2017-07-04
glibc deprecated some stuff, leading to warnings


faho:
In glibc 2.25, a few things we use are marked as deprecated. On my system, this leads to the only remaining compiler warnings.

The full list:

- `readdir_r` (recommending plain `readdir`, which is apparently thread-safe as long as no two threads access the same directory stream simultaneously on glibc and solaris - see e.g. https://github.com/rust-lang/rust/issues/34668)

- `sys_errlist` and `sys_nerr` (recommending `strerror` or `strerror_r`)

faho:
Urgh, the `sys_errlist` stuff is annoying - see #1830.

krader1961:
Note that sys_errlist and sys_nerr have been implicitly deprecated for nearly twenty years. It sucks that strerror is not fork safe. If we can't come up with a better solution I recommend calling strerror for all possible errno values and caching the results as soon as we've initialized the environment but before doing anything that would fork.

A solution I've pondered for this type of problem is an internal RPC mechanism. That would allow the child side of a fork to ask a RPC thread in the main fish process (via a fifo or anonymous unix domain socket) to do something that might acquire a lock and get the results back. This would be a non-trivial amount of work to implement but would allow us to simplify the code in other ways such as eliminating `debug_safe()`. However, doing so in a manner that could be used without fear of deadlock on the child side of a fork is going to be a challenge.

A far simpler solution is to use a barrier to ensure all threads are quiescent before calling fork/spawn. The child process is then free to call malloc, strerror, etc. Yes, I know we generally want to avoid blocking on a thread completing an action. But we have very few threads and none are likely to cause a delay in completing the fork/spawn. If we do have such a thread it is generally trivial to add barrier checks in its inner loop(s) to keep the latency low. The theoretical problem with this approach is that threads created "behind our back" by some library function throws a monkey wrench into the scheme. But we're unlikely to use an API that does so and I think the benefits of this approach outweigh the risks.

krader1961:
Note that @ridiculousfish has already proposed something similar to my preferred solution in our [pthread manifesto](https://github.com/fish-shell/fish-shell/wiki/fish's-Fork-and-Pthread-Manifesto):

> However, this is certainly a fragile area so bugs here are likely. fish has no long-running threads, and so fish can also be put into a mode where it waits for its threads to exit before calling fork. This should improve safety, but may re-introduce fish 1.x's slowdowns on slow filesystems.

[...]

Issue #1830, opened 2014-11-26
Use strerror() or strerror_r() instead of  sys_errlist and sys_nerr


gnu-srs:
Hi,

Looks like your code is still using sys_errlist and sys_nerr for error messages on non-**UCLIBC** systems instead of strerror(). It has been specified by POSIX.1-2001, POSIX.1-2008,  C89,  and  C99.
Why so, e.g. even GNU/Linux marks them as obsolete even if found:
/usr/include/x86_64-linux-gnu/bits/sys_errlist.h
/\* sys_errlist and sys_nerr are deprecated.  Use strerror instead.  */

On GNU/Hurd they are not defined, therefore fish fails to build from source there. In case you want a thread-safe version of strerror(), why not use strerror_r(), as specified by  POSIX.1-2001 and POSIX.1-2008.

Seems like the Debian maintainer is not comfortable with that change non-upstream.

Thanks!

Issue #472, opened 2012-12-23
`fish` process is eating 100% CPU again


minchenkov:
Hello,

I have similar problem like described in issue #422.
fish was built from source code, commit  87baa4d3d224939698926dc119b46e6d87a8b9e8 (so fix for #422 included).

How to reproduce: login to server via ssh, run something like "tail -f /var/log/messages", close terminal window. If login again, there will be `fish` process that eating 100% CPU.

OS Debian GNU/Linux wheezy/sid

faho:
There hasn't been any activity here for 2 years, so I'm closing this.

PR #7006, opened 2020-05-14
safe_strerror(): use thread-safe strerror_r() when available

aaptel:
## Description

* add cmake check for strerror_r()
* make safe_strerror() return simple, safe, always-initialized struct
  holding the error message in a stack allocated array
* remove unsafe fallback code returning static array

The static array solution was not thread-safe as the same memory is
shared and written in by all threads.

This removes the linker warnings mentionned in the comments on
platform supporting strerror_r():

    $ make
    ld: libfishlib.a(wutil.cpp.o): in function `safe_strerror(int)':
    wutil.cpp:307: warning: `sys_errlist' is deprecated; use `strerror' or `strerror_r' instead
    ld: wutil.cpp:307: warning: `sys_nerr' is deprecated; use `strerror' or `strerror_r' instead
    ld: libfishlib.a(wutil.cpp.o): in function `safe_strerror(int)':
    wutil.cpp:307: warning: `sys_errlist' is deprecated; use `strerror' or `strerror_r' instead

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -121,7 +121,7 @@ set(FISH_SRCS
     src/wcstringutil.cpp src/wgetopt.cpp src/wildcard.cpp src/wutil.cpp
     src/future_feature_flags.cpp src/redirection.cpp src/topic_monitor.cpp
     src/flog.cpp src/trace.cpp src/timer.cpp src/null_terminated_array.cpp
-    src/operation_context.cpp src/fd_monitor.cpp
+    src/operation_context.cpp src/fd_monitor.cpp src/safe_strerror.cpp
 )
 
 # Header files are just globbed.
--- a/cmake/ConfigureChecks.cmake
+++ b/cmake/ConfigureChecks.cmake
@@ -85,7 +85,6 @@ check_struct_has_member("struct stat" st_mtimespec.tv_nsec "sys/stat.h"
     HAVE_STRUCT_STAT_ST_MTIMESPEC_TV_NSEC LANGUAGE CXX)
 check_struct_has_member("struct stat" st_mtim.tv_nsec "sys/stat.h" HAVE_STRUCT_STAT_ST_MTIM_TV_NSEC
     LANGUAGE CXX)
-check_cxx_symbol_exists(sys_errlist stdio.h HAVE_SYS_ERRLIST)
 check_include_file_cxx(sys/ioctl.h HAVE_SYS_IOCTL_H)
 check_include_file_cxx(sys/select.h HAVE_SYS_SELECT_H)
 check_include_files("sys/types.h;sys/sysctl.h" HAVE_SYS_SYSCTL_H)
@@ -117,8 +116,6 @@ if(HAVE_XLOCALE_H)
 endif()
 list(APPEND WCSTOD_L_INCLUDES "wchar.h")
 check_cxx_symbol_exists(wcstod_l "${WCSTOD_L_INCLUDES}" HAVE_WCSTOD_L)
-
-check_cxx_symbol_exists(_sys_errs stdlib.h HAVE__SYS__ERRS)
 
 cmake_push_check_state()
 set(CMAKE_EXTRA_INCLUDE_FILES termios.h sys/ioctl.h)
--- a/config_cmake.h.in
+++ b/config_cmake.h.in
@@ -85,9 +85,6 @@
 /* Define to 1 if `st_mtim.tv_nsec' is a member of `struct stat'. */
 #cmakedefine HAVE_STRUCT_STAT_ST_MTIM_TV_NSEC 1
 
-/* Define to 1 if the sys_errlist array is available. */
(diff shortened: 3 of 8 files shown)


ridiculousfish:
Thank you for this contribution; unfortunately the problem is worse than thread safety, it is async-safety. See the backtrace [here](https://github.com/fish-shell/fish-shell/issues/472#issuecomment-11921586): the problem is that `strerror_r` may take a lock which is held by a thread which no longer exists, leading to deadlock. So we can't call `strerror_r` after fork either.

zanchey:
See also #1830 and #4183 for possible alternative approaches. Running `strerror` on startup and caching the results is probably the most sensible option.

aaptel:
I can reproduce the hang relatively easily (ryzen 9 3900x cpu with 12 cores) with the following program. Commenting out the `setlocale()` call makes the hang disappear:

```c
// gcc -Wall -lpthread -o bad bad.c
#include <errno.h>
#include <locale.h>
#include <signal.h>
#include <signal.h>
#include <limits.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <sys/types.h>
#include <sys/wait.h>
#include <unistd.h>
#include <pthread.h>
#include <sys/syscall.h>

#define NB_THREAD 10
#define NB_ITERATION     10000
#define NB_SIG_ITERATION 100000

// strlcat from Apple
// https://opensource.apple.com/source/Libc/Libc-825.25/string/strlcat.c.auto.html
size_t
strlcat(char * restrict dst, const char * restrict src, size_t maxlen) {
    const size_t srclen = strlen(src);
    const size_t dstlen = strnlen(dst, maxlen);
    if (dstlen == maxlen) return maxlen+srclen;
    if (srclen < maxlen-dstlen) {
        memcpy(dst+dstlen, src, srclen+1);
    } else {
        memcpy(dst+dstlen, src, maxlen-1);
        dst[dstlen+maxlen-1] = '\0';
    }
    return dstlen + srclen;
}

// async-safe
void warn(const char *prefix, const char *var, int val)
{
	char buf[512] = {0};
	char valbuf[32] = {0};
	int pos = 0;

	strlcat(buf, prefix, sizeof(buf));
	strlcat(buf, " ", sizeof(buf));
	strlcat(buf, var, sizeof(buf));
	strlcat(buf, "=", sizeof(buf));

	if (val < 0) {
		strlcat(buf, "-", sizeof(buf));
		val *= -1;
	}

	pos = sizeof(valbuf)-2;
	do {
		valbuf[pos--] = (val%10)+'0' ;
		val /= 10;
	} while (val && pos >= 0);

	strlcat(buf, valbuf+pos+1, sizeof(buf));
	strlcat(buf, "\n", sizeof(buf));
	write(2, buf, strlen(buf));
}


void usr1_handler(int num)
{
	char buf[512];

	for (int i = 0; i < NB_SIG_ITERATION; i++) {
		int r = strerror_r(i, buf, sizeof(buf));
		if (r)
			warn("strerror_r", "errno", errno);
	}
}

void *thread(void *x)
{
	char buf[512];
	pid_t tid = syscall(__NR_gettid);

	for (size_t times = 0; times < NB_ITERATION; times++) {
		int r = strerror_r(EACCES, buf, sizeof(buf));
		if (r)
			warn("strerrno_r", "errno", errno);
		if (times % (NB_ITERATION/100) == 0) {
			warn("thread", "tid", tid);
			warn("thread", "progress", times*100/NB_ITERATION);
		}
	}
	return NULL;
}

int main()
{

	pthread_t pid[NB_THREAD] = {0};
	int size = 0;
	pid_t self = getpid();

	printf("MAIN PID %d\n", self);

	// commenting this prevents the hang
	setlocale(LC_ALL, "fr_FR.utf8");

	signal(SIGUSR1, usr1_handler);

	for (size = 0; size < NB_THREAD; size++) {
		int r = pthread_create(&pid[size], NULL, thread, NULL);
		if (r) {
			warn("pthread_create", "r", r);
			exit(1);
		}
	}

	for (int i = 0; i < size; i++) {
		//kill(self, SIGUSR1);
		pthread_kill(pid[i], SIGUSR1);
	}

	for (int i = 0; i < size; i++) {
		//int status; wait(&status);
		pthread_join(pid[i], NULL);
	}
	printf("ok\n");
}
```

aaptel:
New version of the branch does the strerror caching proposed elsewhere and removes the other old _sys_errlist and strerror code.

ridiculousfish:
Apologies for delay in review. Is the idea here only to fix the linker warnings? 200+ new string allocations at startup seems like a high price to pay to do that.

aaptel:
* Future proof: those symbols will get removed at some point so this will be needed eventually
* Regarding memory usage: with an average for 30 bytes per string, that's 6KB
* It simplifies the strerror code a lot
* No linker warnings

ridiculousfish:
Sorry, I think this is too heavyweight especially because it impacts startup. The existing solution is not ideal but is not causing any problems, other than linker warnings. Thank you for the contribution but I am going to decline it.

PR #7008, opened 2020-05-15
Remove text duplicate in tutorial

anothernode:
## Description

These exact same sentences appear again just after the example snippet in the same section.


zanchey:
Thanks!

PR #7012, opened 2020-05-16
__fish_prepend_sudo: Toggle "sudo" on multiple presses

enricozb:
At the moment calling __fish_prepend_sudo multiple times does not toggle
sudo, and also unnecessarily uses the `-c` flag to `commandline` to see if
the first token on the commandline is "sudo".

This change removes the `-c` switch and also toggles "sudo" on multiple
calls to __fish_prepend_sudo, while maintaining the cursor position and
while maintaining any spaces between "sudo" and the next token on the
commandline.

The additional logic around the new_cursor position is because the
cursor could be in the "sudo" during the toggle.

## Description

Talk about your changes here.

Fixes issue #

## TODOs:

### DIFF
--- a/share/functions/__fish_prepend_sudo.fish
+++ b/share/functions/__fish_prepend_sudo.fish
@@ -7,11 +7,6 @@ function __fish_prepend_sudo -d "Prepend 'sudo ' to the beginning of the current
         commandline -C (math $cursor + 5)
     else
         commandline -r (string sub --start=6 (commandline -p))
-        set -l new_cursor (math $cursor - 5)
-        if test "$new_cursor" -lt 0
-          commandline -C 0
-        else
-          commandline -C $new_cursor
-        end
+        commandline -C -- (math $cursor - 5)
     end
 end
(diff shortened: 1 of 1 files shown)


share/functions/__fish_prepend_sudo.fish
```diff
@@ -1,9 +1,17 @@
...
+    set -l cmd (commandline -po)
+    set -l cursor (commandline -C)
     if test "$cmd[1]" != sudo
-        set -l cursor (commandline -C)
         commandline -C 0
         commandline -i "sudo "
         commandline -C (math $cursor + 5)
+    else
+        commandline -r (string sub --start=6 (commandline -p))
+        set -l new_cursor (math $cursor - 5)
```
krobelus, line 10: You can do the same with `commandline -C -- (math $cursor - 5)`

enricozb: Ah I can't believe I didn't think of this... any negative result of the math is being counted as an option to commandline that doesn't exist. Nice catch.

krobelus:
Thanks!

PR #7020, opened 2020-05-19
completions: speed up kdeconnect-cli device discovery

ammgws:
Over 150 times faster now. Also added device name and status to the completion description.
This is done by using the `--shell-device-autocompletion` flag provided by `kdeconnect-cli`.

**Before**
```
>time complete -C"kdeconnect-cli -d"
-d689ccf9fd5f3272d	Device ID
-d689ccf9fd5f3272d	Device ID
Executed in    4.04 secs      fish           external
   usr time   24.87 millis    2.08 millis   22.79 millis
   sys time   14.60 millis    0.12 millis   14.48 millis
```

**After**
```
>time complete -C"kdeconnect-cli -d"
-d689ccf9fd5f3272d	Pixel 4 (paired)
Executed in   22.38 millis    fish           external
   usr time    9.54 millis    2.04 millis    7.51 millis
   sys time   11.23 millis    0.00 millis   11.23 millis
```

### DIFF
--- a/share/completions/kdeconnect-cli.fish
+++ b/share/completions/kdeconnect-cli.fish
@@ -1,3 +1,7 @@
+function __complete_devices
+    kdeconnect-cli --shell-device-autocompletion=zsh | string replace -rf '(\w+)\[(.*)]' '$1\t$2'
+end
+
 complete -c kdeconnect-cli -s l -l list-devices -d "List all devices"
 complete -c kdeconnect-cli -s a -l list-available -d "List available (paired and reachable) devices"
 complete -c kdeconnect-cli -l id-only -d "Make --list-devices or --list-available print the devices id, to ease scripting"
@@ -12,7 +16,7 @@ complete -c kdeconnect-cli -l list-notifications -d "Display the notifications o
 complete -c kdeconnect-cli -l lock -d "Lock the specified device"
 complete -c kdeconnect-cli -l send-sms -f -r -d "Sends an SMS. Requires destination" # message
 complete -c kdeconnect-cli -l destination -f -r -d "Phone number to send the message" # phone number
-complete -c kdeconnect-cli -s d -l device -f -r -a '(kdeconnect-cli -l --id-only)' -d "Device ID" # dev
+complete -c kdeconnect-cli -s d -l device -f -r -a '(__complete_devices)'
 complete -c kdeconnect-cli -s n -l name -f -r -d "Device Name" # name
 complete -c kdeconnect-cli -l encryption-info -d "Get encryption info about said device"
 complete -c kdeconnect-cli -l list-commands -d "Lists remote commands and their ids"
--- a/share/completions/kdeconnect-cli.fish
+++ b/share/completions/kdeconnect-cli.fish
@@ -1,5 +1,6 @@
 function __complete_devices
-    kdeconnect-cli --shell-device-autocompletion=zsh | string replace -rf '(\w+)\[(.*)]' '$1\t$2'
+    kdeconnect-cli --shell-device-autocompletion=zsh | string replace -rf -- '(\w+)\[(.*)]' '$1\t$2'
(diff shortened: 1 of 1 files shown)


faho:
`kdeconnect-cli --shell-device-autocompletion=zsh` outputs nothing for me, while `kdeconnect-cli -l --id-only` shows my device's id.

Sure is fast tho.

ammgws:
That's a shame. I'm on Arch too.

ammgws:
This is the code for that flag:
```
    } else if (parser.isSet(QStringLiteral("shell-device-autocompletion"))) {
        //Outputs a list of reachable devices in zsh autocomplete format, with the name as description
        const QStringList devices = blockOnReply<QStringList>(iface.devices(true, false));
        for (const QString &id : devices) {
            DeviceDbusInterface deviceIface(id);
            QString statusInfo;
            const bool isTrusted = deviceIface.isTrusted();
            if (isTrusted) {
                statusInfo = i18n("(paired)");
            } else {
                statusInfo = i18n("(unpaired)");
            }

            //Description: "device name (paired/unpaired)"
            QString description = deviceIface.name() + QLatin1Char(' ') + statusInfo;
            //Replace characters
            description.replace(QLatin1Char('\\'), QStringLiteral("\\\\"));
            description.replace(QLatin1Char('['),  QStringLiteral("\\["));
            description.replace(QLatin1Char(']'),  QStringLiteral("\\]"));
            description.replace(QLatin1Char('\''), QStringLiteral("\\'"));
            description.replace(QLatin1Char('\"'), QStringLiteral("\\\""));
            description.replace(QLatin1Char('\n'), QLatin1Char(' '));
            description.remove(QLatin1Char('\0'));

            //Output id and description
            QTextStream(stdout) << id << '[' << description << ']' << endl;
        }
```

What does this return for you?

`qdbus org.kde.kdeconnect /modules/kdeconnect/devices/yourdevice org.kde.kdeconnect.device.isTrusted`

and

`qdbus org.kde.kdeconnect /modules/kdeconnect org.kde.kdeconnect.daemon.devices)`

faho:
>qdbus org.kde.kdeconnect /modules/kdeconnect org.kde.kdeconnect.daemon.devices)

What looks like a 16-digit hex number. 5219...

>qdbus org.kde.kdeconnect /modules/kdeconnect/devices/yourdevice org.kde.kdeconnect.device.isTrusted

Replacing "yourdevice" with that hex, it returns `true`.

(also we have usable completions for `busctl`, so tbh I prefer using that instead - dbus is just so frickin verbose)

ammgws:
Weird, no idea why `kdeconnect-cli --shell-device-autocompletion=zsh` doesn't work for you then.

TIL about `busctl`. Looks so much better!

faho:
Ah, okay. When `kdeconnect-cli -l` would print the device is "paired *and reachable*" `kdeconnect-cli --shell-device-autocompletion=zsh`  prints the device is "paired". When `-l` prints just "paired", `--shell...` fails (prints nothing, returns 1)

ammgws:
Most of the commands are for using with a connected device, so perhaps it makes sense for it not to complete? Or we could just fallback to `kdeconnect-cli -l` if `kdeconnect-cli --shell-device-autocompletion=zsh` returns 1.

faho:
The device *is* connected tho? It's lying right next to me, there's no rhyme or reason to when it shows up or doesn't, and kdeconnect seems to work without a hitch?

Is this not sending out a beacon or so that full `-l` would?

ammgws:
Yeah this morning mine didn't show up with --shell... either until I opened the app on my phone and sent a ping (the latter might be unnecessary)
> not sending out a beacon

I guess this is the difference between the two:

```...
 blockOnReply(iface.acquireDiscoveryMode(id));
            QThread::sleep(2);
        }
        const QStringList devices = blockOnReply<QStringList>(iface.devices(reachable, false));
```

ammgws:
I ended up adding a fallback to the slower option when the first returns 1.

faho:
Yeah, that should work. Thanks!

PR #7027, opened 2020-05-21
Fix is_wsl() on wsl2

cherichy:
On wsl2, /proc/version contains no "Microsoft" as wsl1 do.
Changing "Microsoft" to "microsoft" will make it work on both wsl1 and wsl2.

### DIFF
--- a/share/tools/web_config/webconfig.py
+++ b/share/tools/web_config/webconfig.py
@@ -47,7 +47,8 @@ def is_wsl():
     """ Return whether we are running under the Windows Subsystem for Linux """
     if "linux" in platform.system().lower() and os.access("/proc/version", os.R_OK):
         with open("/proc/version", "r") as f:
-            if "microsoft" in f.read():
+            # Find 'Microsoft' for wsl1 and 'microsoft' for wsl2
+            if "microsoft" in f.read().lower():
                 return True
     return False
 
--- a/share/tools/web_config/webconfig.py
+++ b/share/tools/web_config/webconfig.py
@@ -47,7 +47,7 @@ def is_wsl():
     """ Return whether we are running under the Windows Subsystem for Linux """
     if "linux" in platform.system().lower() and os.access("/proc/version", os.R_OK):
         with open("/proc/version", "r") as f:
-            if "Microsoft" in f.read():
+            if "microsoft" in f.read():
                 return True
     return False
 


faho:
1. Are our workarounds still required on WSL2? Personally I'd love to just handle that as linux, basically anything we have to do specifically for it should be a bug there as I understand WSL's purpose. (also is WSL1 now dead or still a thing?)

2. That's not the only place we do that check, see also https://github.com/fish-shell/fish-shell/blob/e51f4e5fc9ca5032690c696c87f5656e648d4b64/share/functions/help.fish#L83 and https://github.com/fish-shell/fish-shell/blob/e51f4e5fc9ca5032690c696c87f5656e648d4b64/src/common.cpp#L172.

cherichy:
If we treat wsl2 as linux, "webbrowser.open(fileurl)" is called. But a url like "file:///home/xxx/" doesn't work under wsl2.

ammgws:
> also is WSL1 now dead or still a thing?

WSL2 is not officially out yet as far as I know. And even if it was, people will still be on WSL1 for a while. At least I will, since updates are controlled by company IT and they're slow at rolling out new versions of windows

zanchey:
The "broken" versions referred to in the `common.cpp` check are still in support (until November 2020), and the WSL check is still needed to block `.dll` files from completion (a9b582d2a806a).

faho:
Okay, so if a check for lowercase  "microsoft" works here as well, should we just change that everywhere?

Mind that the checks are typically case-sensitive.

cherichy:
The utsname.release on wsl1looks like "4.4.0-18362-Microsoft" (windows 1903 (18362)).
But on wsl2, it looks like ”4.19.84-microsoft-standard" (with windows 2004 (19041)).
Discussions can be found here. [How to tell in Linux under WSL if using WSL1 vs WSL2 ?](https://github.com/microsoft/WSL/issues/4555)
I am sorry that my patch using lowercase only works for wsl2. I will change it to check both cases.

cherichy:
@faho I have checked the files common.cpp and help.fish. They don't need to change. 
The problem is fish_config falis to start on wsl2. And the workarounds can be applied only in this file.

zanchey:
Yes, sounds good; the `common.cpp` check is another issue which doesn't need to block the merging of this patch.

PR #7028, opened 2020-05-21
Add fish_add_path, a simple way to add to $PATH

### LINKED ISSUE #6960: Simple way to add to the path
(This has been discussed in https://github.com/fish-shell/fish-shell/issues/527 previously)

I can never remember how to set the path persistently in fish, When I google it I end up at issue #527 but the discussion there is a bit confused. I suggest two things:

1. Add a command `add_to_path` that is _idempotent_ and validates the path.
2. Edit the top comment of #527 or the title to point out this new way so people who end up there from google get the right information. Google results changes too slowly!

I have written a draft of an implementation of `add_to_path`:

```fish
#!/usr/bin/env fish
if argparse -n 'fish-add-user-path' -N 1 'h/help' -- $argv
    ;
else
    exit
end

function add_to_path_if_not_there
    set exists 0

    # normalize path to end with /
    switch $argv
    case "*/"
        set new_path $argv
    case "*"
        set new_path "$argv/"
    end

    for x in (string split " " $fish_user_paths)
        if test "$x" = "$new_path"
            set exists 1
            break
        end
    end

    if test "$exists" = "0"
        set -Ux fish_user_paths $fish_user_paths $new_path
    end   
end

# validate that paths exists
for x in (string split " " $argv)
    if test -d $x
        ;
    else
        echo "Path '$x' does not exist"
        exit
    end
end


for x in (string split " " $argv)
    add_to_path_if_not_there $x 
end
```

This is my first fish program so please help me in making it suck less :P

ping @jonlorusso @knightsamar @kegesch @raunyoliveira-hotmart @i-am-the-slime @SamHH @brianmhunt @radoslawc @cumulotimbus @anothercooldude and many more that have seemed to think this is a good idea.

faho:
## Description

```fish
fish_add_path [paths...]
fish_add_path (-h | --help)
fish_add_path [(-g | --global) | (-U | --universal) | (-P | --path)] 
    [(-m | --move)] [(-a | --append) | (-p | --prepend)] 
    [(-v | --verbose) | (-n | --dry-run)] [paths...]
```

This is a function you can either execute once, interactively, or
stick in config.fish, and it will do the right thing.

Some options are included to choose some slightly different behavior,
like setting $PATH directly instead of $fish_user_paths, or moving
already existing components to the front/back instead of ignoring
them, or appending new components instead of prepending them.

The defaults were chosen because they are the most safe, and
especially because they allow it to be idempotent - running it again
and again and again won't change anything, it won't even run the
actual `set` because it skips that if all components are already in.

Fixes #6960.

## TODOs:


faho:
Caveat: We do not `realpath` the existing $PATH. I don't think that's much of an issue - it's more important that a path is searched or not, not with which name it is. It's also mostly for deduplication (to avoid adding a component again and again), so worst case we add a path once unnecessarily - not that it hurts much.

TODO: We need to absolutize non-existent paths still. `realpath` needs to be changed to do this.

faho:
Okay, so things up for discussion:

- The name (maybe just `add_path`? or `addpath`?)
- The semantics, in particular
- Warning when a path doesn't exist

zanchey:
I agree with https://github.com/fish-shell/fish-shell/issues/6960#issuecomment-632497031 that `fish_add_path` is the way to go.

I don't think this is needed for other path variables.

faho:
Okay, I would *really* like to remove the warning, as we've heard from another person who has trouble ssh-ing to a machine because of the $fish_user_paths warning in fish 2.7.

Adding a warning that isn't often triggered to something where you only see it after logging back in is awful, especially in conjunction with ssh where that can block login.

krobelus:
> Okay, I would _really_ like to remove the warning

Seems good, the warning can be a false positive. I guess if we wanna be fancy we can warn only when `status current-filename` is Standard input.

faho:
Okay, merged as 9354dd697..1e17a6813

PR #7033, opened 2020-05-23
fish_vcs_prompt: allow argv passthrough

ammgws:
`fish_git_prompt` encloses its output in brackets, however this can be changed by supplying a format string to it, i.e. `fish_git_prompt %s`.

However when using `fish_vcs_prompt` there's no way to pass on the arg to `fish_git_prompt`, so you need to manually remove it.

`fish_hg_prompt` doesn't have the same format string support as `fish_git_prompt`, but I suppose it could be added later if needed.


faho:
Yeah, that's right.

Tho we should improve the format string idea and document it.

PR #7041, opened 2020-05-26
Funcsave with --directory option

dkrm0:
## Description

Add the option `--directory` to the `funcsave` function, to specify where to save the functions.

## TODOs:

### DIFF
--- a/share/completions/funcsave.fish
+++ b/share/completions/funcsave.fish
@@ -1,2 +1,2 @@
 complete -c funcsave -xa "(functions -na)" -d "Save function"
-complete -c funcsave -s d -l directory -d "Directory to save the functions" -a $fish_function_path -r
+complete -c funcsave -s d -l directory -d "Directory to save the functions" -a '$fish_function_path' -r
--- a/share/completions/funcsave.fish
+++ b/share/completions/funcsave.fish
@@ -1,1 +1,2 @@
 complete -c funcsave -xa "(functions -na)" -d "Save function"
+complete -c funcsave -s d -l directory -d "Directory to save the functions" -r -F
--- a/share/completions/funcsave.fish
+++ b/share/completions/funcsave.fish
@@ -1,2 +1,2 @@
 complete -c funcsave -xa "(functions -na)" -d "Save function"
-complete -c funcsave -s d -l directory -d "Directory to save the functions" -r -F
+complete -c funcsave -s d -l directory -d "Directory to save the functions" -a $fish_function_path -r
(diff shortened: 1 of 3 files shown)


share/completions/funcsave.fish
```diff
@@ -1 +1,2 @@
 complete -c funcsave -xa "(functions -na)" -d "Save function"
+complete -c funcsave -s d -l directory -d "Directory to save the functions" -r -F
```
faho, line 2: I would just complete $fish_function_path here.

Saving a function to a directory that's not in there isn't likely to be helpful, and completing *files* is even less so.

share/functions/funcsave.fish
```diff
@@ -8,20 +8,26 @@ function funcsave --description "Save the current definition of all specified fu
...
+
     if not set -q argv[1]
         printf (_ "%ls: Expected at least %d args, got only %d\n") funcsave 1 0
         return 1
     end
 
-    if not mkdir -p $__fish_config_dir/functions
-        printf (_ "%s: Could not create configuration directory\n") funcsave
+    if not mkdir -p $funcdir
+        printf (_ "%s: %s: Could not create configuration directory\n") funcsave $funcdir
```
faho, line 23: This reads as

    "funcsave: /home/alfa/.config/fish/test/functions: Could not create configuration directory"

I think it's better as 

    "funcsave: Could not create directory '/home/alfa/.config/fish/test/functions'"

share/completions/funcsave.fish
dkrm0: I thought the `-F` options also worked for path.

share/functions/funcsave.fish
dkrm0: Fixed

share/completions/funcsave.fish
faho: Well, yes, it does. But it also works for every other directory, and worse every other file.

`funcsave --directory ~` is unlikely to be what you want (as you then can't actually use the function) and `funcsave --directory /usr/bin/grep` is complete nonsense. So it's unhelpful to complete either of these.

share/completions/funcsave.fish
```diff
@@ -1 +1,2 @@
 complete -c funcsave -xa "(functions -na)" -d "Save function"
+complete -c funcsave -s d -l directory -d "Directory to save the functions" -a $fish_function_path -r
```
faho, line 2: As it is, this will be expanded when the file is loaded, and expanded into multiple arguments *to complete*.

You need to quote it. If you double-quoted it, that would expand it when the file is loaded, as one argument, but it would always keep the value of $fish_function_path when the file was loaded.

So you need to *single*-quote it:

```fish
complete -c funcsave -s d -l directory -d "Directory to save the functions" -a '$fish_function_path' -r
```

This will cause *complete* to expand it when it needs to offer the completions.

dkrm0: Done. Thank you for the explanation :)

faho:
Merged, thanks!

PR #7053, opened 2020-05-27
Fix typo

ammgws:
As per title


faho:
This would, in fact, appear to work better. Thanks!

PR #7054, opened 2020-05-27
gpg completions: complete key id for `--edit-key`

ammgws:
More useful than just user id since can have multiple keys per user.


faho:
Yeah, thanks!

PR #7057, opened 2020-05-28
Convert (most of) the ".md" files to rst

faho:
## Description

This converts (mostly automatically, via `pandoc`) CHANGELOG, CONTRIBUTING and README to .rst files.

The advantage is that we use rst in our documentation so there's no need to use two languages, and that we ~~could~~ can include these in the html documentation, particularly the CHANGELOG.

Currently fishshell.com includes a cheesy rendered version at https://fishshell.com/release_notes.html, but it's not really integrated with the rest of the docs.

Note that this PR is very likely to go stale quickly - as soon as a change to the CHANGELOG is made, but I can run the conversion again.

For a view of how these render on github, check [my repo](https://github.com/faho/fish-shell/blob/all-rst/CHANGELOG.rst)

This does *not* convert the issue/PR templates because I'm not sure they *work* as rst - I think they're taken from master, so I'd have to merge it to my master branch to check, and I've not gotten around to it. They're not critical.

## TODOs:

- [Not really?] Changes to fish usage are reflected in user documentation/manpages.
- [Well, no] User-visible changes noted in CHANGELOG.md -


faho:
Merged as e6f5c78d3..4f6851b19

zanchey:
Fixed the Debian build in 2c6e95ccf.

The only problem I have is that GitHub doesn't support RST when writing the release information in https://github.com/fish-shell/fish-shell/releases, but it sounds like it should be easy to convert using pandoc.

PR #7065, opened 2020-05-30
completions: update xrandr

ammgws:
Main changes: Adds new options from new version of xrandr

Other changes: Shorten descriptions etc

### DIFF
--- a/share/completions/xrandr.fish
+++ b/share/completions/xrandr.fish
@@ -13,6 +13,7 @@ function __fish_print_xrandr_modes --description 'Print xrandr modes'
         end
     end
 end
+
 complete -c xrandr -l verbose -d 'Be more verbose'
 complete -c xrandr -l dryrun -d 'Make no changes'
 complete -c xrandr -l nograb -d 'Apply modifications without grabbing the screen'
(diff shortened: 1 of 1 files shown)


share/completions/xrandr.fish
```diff
@@ -39,6 +39,32 @@ complete -c xrandr -s o -l orientation -d 'Specify the orientation of the screen
...
+end
+
+# Version > 1.4
+if test $ver[1] -gt 1; or test "$ver[2]" -ge 4
+    complete -c xrandr -l listproviders -d 'Print all available providers'
+    complete -c xrandr -l setprovideroutputsource -d 'Set source for a given provider'
+    complete -c xrandr -l setprovideroffloadsink  -d 'Set provider for a given sink'
+end
+
+# Version > 1.3
```
faho, line 57: Version 1.5 was released in 2016, 1.4. in 201**3**. There is no need to check that far back, especially when it just comes to offering options that might not work.

faho: If you wanted to call `xrandr` in the completion script, and that call required a version, I might be convinced to keep the check for 1.5.

But if it's just about offering options? Nah, ditch them.

ammgws: In another project someone on Ubuntu 18.04 LTS (IIRC) still had an older version by default. We can just get rid of all version checks since this is probably one of a couple completions that actually check the version.

ammgws: > Nah, ditch them.

Sayonara

faho: I think you might be mistaken, since Ubuntu **16**.04 (i.e. the oldest still supported version) has 1.5: https://packages.ubuntu.com/search?suite=xenial&searchon=names&keywords=xrandr

Ubuntu *14*.04 might not have, but we don't even build fish 3.1.2 for it in our PPA.

And even *iff* someone ran this on a distribution that old, the failure mode is quite benign - try an option, see that it doesn't work, know you have to find a workaround (like, you know, upgrading to something not old enough to go to school).

ammgws: It was a different X utility, but your point still stands.

faho:
Merged, thanks!

PR #7067, opened 2020-05-31
remove redundant if conditions

JohnCoconut:
All pcre2 resource free functions handle null pointer gracefully. Nothing would be done if a NULL pointer is passed to `pcre2_code_free` or `pcre2_match_data_free`.

1.  https://www.pcre.org/current/doc/html/pcre2_code_free.html
 2. https://www.pcre.org/current/doc/html/pcre2_match_data_free.html


ridiculousfish:
Thank you!

PR #7073, opened 2020-06-01
Add "alacritty" to the list of TERM values of terminals supporting dynamic titles

rye:
# Description

This change fixes support for "dynamic titles" for users of the Alacritty terminal.  Users with the terminfo for `alacritty` installed in their terminfo database would have their `TERM` values set to `alacritty` by default (as opposed to `xterm-256color`), which would break fish-shell's understanding.

Prior to this change, the [`does_term_support_setting_title`](https://github.com/fish-shell/fish-shell/blob/4d487f711d75f19ec78a3743ea60b04185a45fc0/src/env_dispatch.cpp#L421-L444) function would return `false` because the term value `alacritty` matches the pattern `tty` (causing [this line](https://github.com/fish-shell/fish-shell/blob/4d487f711d75f19ec78a3743ea60b04185a45fc0/src/env_dispatch.cpp#L440) to return `false`).  Other values of `TERM` not including the substrings `tty` or `/vc/` and not matching any of the prior patterns would instead be interpreted as having support for dynamic titles, but in this case because the string `alacritty` contains "tty", it would get interpreted as _not_ having support for dynamic titles.

I'd be happy to make any additional changes requested!

## TODOs:

### DIFF
--- a/src/env_dispatch.cpp
+++ b/src/env_dispatch.cpp
@@ -417,7 +417,8 @@ static bool initialize_curses_using_fallback(const char *term) {
 /// One situation in which this breaks down is with screen, since screen supports setting the
 /// terminal title if the underlying terminal does so, but will print garbage on terminals that
 /// don't. Since we can't see the underlying terminal below screen there is no way to fix this.
-static const wchar_t *const title_terms[] = {L"xterm", L"screen", L"tmux", L"nxterm", L"rxvt"};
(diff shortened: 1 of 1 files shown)


rye:
I'm not sure if this PR warrants an addition to `CHANGELOG.rst`, but I'll be happy to make a note of that if it does.

ridiculousfish:
Thank you!

Issue #7071, opened 2020-06-01
Completion file in $fish_complete_path does not work as intented


soraxas:
As I am not familiar with fish-shell's internal, so please correct me if I am wrong.
I think this problem exists because `fish` loads completion files on-demand, i.e., whenever user tries to tab-complete some commands?

So there is this feature in `git` where every binary with the prefix `git-XXX` can be executed as a custom command like `git XXX`, which is how [git-extras](https://github.com/tj/git-extras) and [git-town](https://github.com/git-town/git-town) add extra functionalities to `git`.

When I was writing a completion file for [git-town](https://github.com/git-town/git-town), I experienced these strange behaviours.

### Behaviour 1

When I placed `git.fish` in `~/.config/fish/completions/` and spawns a new shell, the completions work when I tried to tab-complete
```sh
fish town [tab]
```
However, the original git autocomplete is no longer here. I believe it is because `fish-shell` was able to find a completion file in `~/.config/fish/completions/` before it searches in `/usr/share/fish/completions/`, and therefore the official git autocompletion does not get loaded? (similar to $PATH behaviour).

### Behaviour 2a

When I named it as `git-town.fish` in `~/.config/fish/completions/` and spawns a new shell, the completions **does not get loaded** when I tried to tab-complete
```sh
fish town [tab]
```

### Behaviour 2b

Similar as *2a*, but instead I first tried
```sh
fish-town [tab]
```
which obviously doesn't work (because `complete` command in `git-town.fish` uses `git` as the name), but then afterwards when I tried to tab-complete
```sh
fish town [tab]
```
it works just like **Behaviour 1**.

I believe it is because `fish-town [tab]` forces fish-shell to load the `~/.config/fish/completions/git-town.fish` file. And even though it wasn't able to provide completion for `git-town` commnad, it loads my completion for `git town`; hence my subsequence
```sh
fish town [tab]
```
works as intended.

----------------

## Thoughts
1. I guess I understand the reason for fish's loading autocompletion on-demand (for performance reasons). But maybe after the main shell is launched, it should async load the rest of completion files in the background? (might make the shell memory bulky?)
2. Currently, I guess the complete file must be named to be same as the command that we want to provide completion for? (e.g. If I name it as `git-town.fish`, then `fish` only expects to find completion for `git-town` command in the file?)
[...]

PR #7075, opened 2020-06-02
Add support for git-* command auto-completion

soraxas:
Fixes issue #7071

## Description
This PR addresses #7071 by loading `git-*.fish` completion in the main `git.fish`.

This approach addresses the following
1. This does not add overheads to the existing `git` if the user had not installed any `git-*.fish` completion file, other than the searches for completion file during initialisation.
2. Even if the main-repo or vendor provides some `git-*.fish` completion file, this approach won't loads them if the user had not installed the executable. Therefore, it won't clutter the shell by loading unnecessary completion.
3. This follows the traditional loading approach where it only loads the first completion file `git-*.fish` it founds.
4. It directly addresses the issue of currently there is no way (other than eager loading by user) to add completions for `git <foo>`. This approach allows you to add `git-foo.fish` to your completion file and it will load it whens git is invoked.
5. By looking at the current `git.fish` it already tries to provide completion for custom `git-*` executable, so it makes sense to also load their completion, if exists, afterwards.

## TODOs:

### DIFF
--- a/share/completions/git.fish
+++ b/share/completions/git.fish
@@ -1870,9 +1870,20 @@ complete -c git -n '__fish_git_using_command bisect; and __fish_seen_argument --
 
 ## Custom commands (git-* commands installed in the PATH)
 complete -c git -n __fish_git_needs_command -a '(__fish_git_custom_commands)' -d 'Custom command'
+
+# Get the path to the generated completions
+# If $XDG_DATA_HOME is set, that's it, if not, it will be removed and ~/.local/share will remain.
+set -l generated_path $XDG_DATA_HOME/fish/generated_completions ~/.local/share/fish/generated_completions
+
+# We don't want to modify $fish_complete_path here, so we make a copy.
+set -l complete_dirs $fish_complete_path
+
+# Remove the path to the generated completions if it is in the list
+set -l ind (contains -i -- $generated_path[1] $complete_dirs); and set -e complete_dirs[$ind]
+
 # source git-* commands' autocompletion file if exists
 set -l __fish_git_custom_commands_completion
-for git_ext in $fish_complete_path/git-*.fish
+for git_ext in $complete_dirs/git-*.fish
     # ignore this completion as executable does not exists
     set -l cmd (string replace -r '.*/([^/]*)\.fish' '$1' $git_ext)
     not command -q $cmd
--- a/share/completions/git.fish
+++ b/share/completions/git.fish
@@ -1872,16 +1872,14 @@ complete -c git -n '__fish_git_using_command bisect; and __fish_seen_argument --
 complete -c git -n __fish_git_needs_command -a '(__fish_git_custom_commands)' -d 'Custom command'
 # source git-* commands' autocompletion file if exists
 set -l __fish_git_custom_commands_completion
-for complete_path in $fish_complete_path
-    for git_ext in $complete_path/git-*.fish
-        # ignore this completion as executable does not exists
-        set -l cmd (string replace '.fish' '' (basename $git_ext))
-        not command -q $cmd
-        and continue
-        # already sourced this git-* completion file from some other dir
-        contains -- $git_ext $__fish_git_custom_commands_completion
-        and continue
-        source $git_ext
-        set -a __fish_git_custom_commands_completion $git_ext
-    end
+for git_ext in $fish_complete_path/git-*.fish
+    # ignore this completion as executable does not exists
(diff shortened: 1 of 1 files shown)


share/completions/git.fish
```diff
@@ -1870,3 +1870,18 @@ complete -c git -n '__fish_git_using_command bisect; and __fish_seen_argument --
 
 ## Custom commands (git-* commands installed in the PATH)
 complete -c git -n __fish_git_needs_command -a '(__fish_git_custom_commands)' -d 'Custom command'
+# source git-* commands' autocompletion file if exists
+set -l __fish_git_custom_commands_completion
+for complete_path in $fish_complete_path
+    for git_ext in $complete_path/git-*.fish
```
faho, line 1876: This can actually be done in a single loop:

```fish
for git_ext in $fish_complete_path/git-*.fish
```

share/completions/git.fish
```diff
@@ -1870,3 +1870,18 @@ complete -c git -n '__fish_git_using_command bisect; and __fish_seen_argument --
...
+        # ignore this completion as executable does not exists
+        set -l cmd (string replace '.fish' '' (basename $git_ext))
```
faho, line 1878: I'd do this as a single regex:

```fish
set -l cmd (string replace -r '.*/([^/]*).fish' '$1' $git_ext)
```

i.e. replace everything with the "/"-less bit between a "/" and ".fish".

`basename` is one of those things that barely does anything while being surprisingly expensive.

share/completions/git.fish
```diff
@@ -1870,3 +1870,18 @@ complete -c git -n '__fish_git_using_command bisect; and __fish_seen_argument --
...
+        not command -q $cmd
+        and continue
+        # already sourced this git-* completion file from some other dir
+        contains -- $git_ext $__fish_git_custom_commands_completion
```
faho, line 1882: This checks for the full path ($git_ext), should it not just check for the command ($cmd)?

share/completions/git.fish
```diff
@@ -1870,3 +1870,18 @@ complete -c git -n '__fish_git_using_command bisect; and __fish_seen_argument --
...
+        and continue
+        source $git_ext
+        set -a __fish_git_custom_commands_completion $git_ext
```
faho, line 1885: Similarly this would *add* the full path instead of just the command.

krobelus:
This is a really nice approach! We can extend it to similar commands in future.

share/completions/git.fish
soraxas: Thanks for the suggestion! I didn't know you can do that in shell script

share/completions/git.fish
soraxas: Sorry about that this is my oversight

faho:
One problem with this approach:

It reads the `git-$COMMAND` autogenerated files in ~/.local/share/fish/generated_completions.

I'm not sure how to solve it. In a pinch skipping that directory would probably do.

soraxas:
Hmm indeed, those ones that are generated from man page.

So I just checked and in fact, since those `git-*` man page has no associated executable, the code would reject them with the
```fish
not command -q $cmd
and continue
```
guard. (e.g. even though there are bunch of things like `git-pull.fish` and `git-send-email.fish` completion files in `$XDG_DATA_HOME/generated_completions/`, since there are no associated executable (e.g. no such `git-pull` standalone command exists), the `command -q git-pull` would fails.

Having said that, I'm not sure how useful those auto-generated completions are for custom git-* command (um probably?). If we want to save reductant check of `command -q` within `generated_completions` we can probably skip the auto-gen directory without much harm

faho:
>So I just checked and in fact, since those git-* man page has no associated executable, the code would reject them with the

Indeed. 

I'd still rather go through the directory list to remove it. E.g.

```fish
# Get the path to the generated completions
# If $XDG_DATA_HOME is set, that's it, if not, it will be removed and ~/.local/share will remain.
set -l generated_path $XDG_DATA_HOME/fish/generated_completions ~/.local/share/fish/generated_completions

# We don't want to modify $fish_complete_path here, so we make a copy.
set -l complete_dirs $fish_complete_path

# Remove the path to the generated completions if it is in the list
set -l ind (contains -i -- $generated_path[1] $complete_dirs); and set -e complete_dirs[$ind]

for git_ext in $list/git-*.fish
# ...
```

soraxas:
Thanks @faho for your kind suggestions

faho:
Merged. Thank you and sorry for the delay!

PR #7076, opened 2020-06-02
Add `status dirname` and `status basename` convenience commands

faho:
## Description

There's a terrible number of fishscripts that start with

    set path (dirname (status filename))

And that's really just a bit boring.

So let's let it be

    set path (status dirname)

And while we're at it, add "status basename" as well.

There are some alternatives here, including:

- Making `dirname` a builtin - this doesn't solve the problem of having that boilerplate and it has the problems associated with overriding system commands

- Adding a variable - tbh I'm not super fond of adding more electric variables unless necessary

- Adding an option to `status filename` - `status filename --dir-only` - this is a bit uglier, it does cut down on the number of subcommands, but I don't think that's much of an issue

## TODOs:

### DIFF
--- a/doc_src/cmds/status.rst
+++ b/doc_src/cmds/status.rst
@@ -19,6 +19,8 @@ Synopsis
     status is-interactive-job-control
     status current-command
     status filename
+    status basename
+    status dirname
     status fish-path
     status function
     status line-number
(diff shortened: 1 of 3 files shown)


krobelus:
Surprisingly I found [just one occurrence](https://grep.app/search?q=dirname.%2A%5C%28status%20%28--%29%3Ffilename%5C%29&regexp=true&filter[lang][0]=fish) (I must be doing something wrong here). Of course this pattern is super common in (ba)sh scripts so we want have a good story here, even though if it doesn't matter much for interactive use.

faho:
>Surprisingly I found just one occurrence (I must be doing something wrong here).

Yeah, that can't be right.

One problem is that `status filename` is available under... too many names, tbh. One of them is `status -f`, and that finds me a bunch of results, including 

- [in our own test suite](https://github.com/fish-shell/fish-shell/blob/master/tests/interactive.fish)
- [in our test suite again](https://github.com/fish-shell/fish-shell/blob/master/tests/checks/broken-config.fish)
- [another in our test suite](https://github.com/fish-shell/fish-shell/blob/master/tests/test.fish)
- in [ansible](https://github.com/ansible/ansible/blob/7323d5dd0dadad1c706eae331526cd481785456a/hacking/env-setup.fish#L4)
- in [bass](https://github.com/edc/bass/blob/master/functions/bass.fish)

ridiculousfish:
Changes LGTM. I don't have any objection to `status dirname` or `status basename` if this pattern is used often.

PR #7079, opened 2020-06-03
Let `__fish_prepend_sudo` use the last commandline if there is no current one

faho:
## Description

As we all know, `!!` is a frequently asked for feature, but 99.7% of the time it's specifically for `sudo !!`, which is just a super awkward interface.

This makes it so that `__fish_prepend_sudo` uses the previous commandline if the current commandline is empty, so we finally win on keystrokes against `sudo !!`. Truly that is the most important metric.

## TODOs:


exploide:
> so we finally win on keystrokes

I think fish already does, because <kbd>↑</kbd> <kbd>Alt</kbd>+<kbd>S</kbd> is three keystrokes while just typing `sudo` is already four.

But probably a good idea to make it even shorter :smiley:

Issue #4829, opened 2018-03-16
`fish_postexec` should probably not be emitted on blank enter


mqudsi:
The `fish_postexec` event is documented as being emitted "right after executing an interactive command" and is noted to also be executed when an invalid command is entered (e.g. command not found).

However, in implementation the event is emitted too generously, e.g.

```fish
~> <ENTER>
fish_postexec called
~>
```

There's already a `fish_prompt` that is called each time the prompt is displayed.

zanchey:
I think `fish_postexec` is useful also for terminal integration or other software that may ship a configuration snippet rather than modifying `fish_prompt`.

krobelus:
resolved by #7085

Issue #4926, opened 2018-04-18
Do not update CMD_DURATION when pressing enter with no command


MauricioA:
The variable CMD_DURATION stores how long the last command took to run.
When pressing enter without writing anything in the terminal, CMD_DURATION is updated, as if simply pressing enter was a command (that is, it's replaced with 0 most times).

The way I see it, this is inconsistent with history and with the status variable:
- history doesn't consider an empty line as a command, and nothing gets stored in the history if you simply press enter in a terminal
- the status variable doesn't get modified when you press enter on an empty line

For consistency, it might make more sense to only update CMD_DURATION if an actual command was run.

This would be helpful for me, since I'm trying to print the duration of the last command with fish_right_prompt.

If there's a reason for not changing this, then can you tell me if there's a way to tell if the last command was an actual command and not an empty line?

fish, version 2.6.0
Linux 4.9.0-6-amd64 #1 SMP Debian 4.9.82-1+deb9u3 (2018-03-02) x86_64 GNU/Linux
screen-256color

[![asciicast](https://asciinema.org/a/Q6OK1aQz09P8aj6Cc9bJDbkAT.png)](https://asciinema.org/a/Q6OK1aQz09P8aj6Cc9bJDbkAT)

faho:
Sounds alright.

mqudsi:
Possibly related to #4829

swayamraina:
I will be happy to contribute and fix this.
I have cloned the repo and made changes. I need to test the changes. Can anyone guide how to build and test?

ridiculousfish:
Sure, see the [Building](https://github.com/fish-shell/fish-shell#building) instructions in the readme

PR #7085, opened 2020-06-06
Skip pre/post exec events for empty commands (#4829)

awalgarg:
## Description

In the same vein as #4926, firing `fish_preexec` and `fish_postexec` for empty commands is inconsistent with the way `$status`, `history` and `$CMD_DURATION` treat empty commands. `fish_postexec`, for example, receives an incorrect value for `$CMD_DURATION` when an empty command is processed.

As suggested in #4829, this patch skips firing of these events for empty commands altogether.

Fixes issue #4829

## TODOs:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -28,6 +28,8 @@ Notable improvements and fixes
 -  A new ``fish_add_path`` helper function to add paths to $PATH without producing duplicates, to be used interactively or in config.fish (#6960)
 -  The ``_`` helper function to call into fish's gettext catalog has been made a builtin for simplicity and performance (#7036)
 -  ``:`` (for doing nothing) and ``.`` (as a less readable name for ``source``) have also been made proper builtins rather than wrapper functions (#6854).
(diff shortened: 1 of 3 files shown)


faho:
So what *are* we doing for empty commands? Should we possibly just not be doing anything instead?

awalgarg:
I do think simply ignoring empty commands altogether should be fine, but felt it might be too big a change. Empty commands are often used to "separate" a bunch of commands from the rest for readability, or for some reason to refresh a prompt with a time indicator etc.

This patch simply follows the precedent set from CMD_DURATION to fix the issue with postexec hooks receiving an incorrect duration value. If we are okay with ignoring empty commands altogether, I can just do that then.

ridiculousfish:
I agree with faho, let's merge this, and then try skipping even more work for empty commands.

Merged as cb5eb72c6b15a574687465887296a532da3dcd8c
Extended as 61e948454f463e0adcb94e88a712d61958b270f0

Thank you!

awalgarg:
Thanks, @ridiculousfish!

PR #4685, opened 2018-01-26
Fix the build on Clang-6.0

asomers:
## Description

Clang-6.0 does not define __cpp_lib_make_unique even though it does
define std::make_unique.  Use the standard library's std::make_unique
for any C++14 toolchain.

## TODOs:

- [N/A] Changes to fish usage are reflected in user documentation/manpages.
- [N/A] Tests have been added for regressions fixed
- [N/A] User-visible changes noted in CHANGELOG.md

PR #7088, opened 2020-06-06
fix compilation under C++14

neheb:
Currently, setting CMAKE_CXX_STANDARD to 14 breaks compilation. The
CMake check used isn't working.

__cpp_lib_make_unique is a standard C++ macro.


ridiculousfish:
Unfortunately `__cpp_lib_make_unique` is not always trustworthy; see #4685.

 fish compiles successfully for me after changing `CMAKE_CXX_STANDARD` from 11 to 14. What problem are you having?

neheb:
Hrm. It seems I tested this under a wrong branch. It was erroring saying that make_unique was ambiguous. Typically happens with using namespace std.

PR #7092, opened 2020-06-06
run through clang-tidy

neheb:
most of these are minor.

bumping the version to C++17 allowed a few more warnings including modernize-use-nodiscard.


ridiculousfish:
Thank you. I will pick a few of these, others I am declining, some revealed real issues that I want to fix in other ways. Merged as 9636d9f

PR #7093, opened 2020-06-07
Fix string collect examples

mikelward:
collect -N leaves the trailing newline, not the other way around.


faho:
Oh, that's a good catch, thanks!

PR #7094, opened 2020-06-07
builtin_string: Remove redundant conditionals in handle_flag_f

rye:
The removed comparisons (`{begin,end,field} == INT_MIN`) always evaluated to `false`, because at this point in evaluation, `begin <= 0` has already been evaluated to be `false`.  Since `INT_MIN <= 0`, the second conditional in all three of the affected cases was always `false`.  I'm pretty sure this change should be invisible.

The C++ standard seems to guarantee left-to-right evaluation of logical operators, but not necessarily bitwise operators, which were not used here.


ridiculousfish:
Thank you!

PR #7100, opened 2020-06-09
Add completions for yadm

gokulsoumya:
## Description

Add completions for [yadm](https://github.com/TheLocehiliosan/yadm). It's wraps around git and all of git's features are usable.

## TODOs:

- [x] Suppress git completions for `clone` subcommand as yadm has `clone` with different options
- [x] `yadm gitconfig` should inherit completions from `git config`
- [x] Respect `--yadm-repo` option

I'm not sure if suppressing inherited completions for specific options is possible or not, so could use a pointer.


share/completions/yadm.fish
```diff
@@ -0,0 +1,87 @@
...
+    contains -- $cmd $argv
+    and return 0
+    return 1
+end
+
+set -l yadm_work_tree (yadm gitconfig --get core.worktree)
+set -l yadm_repo (yadm introspect repo)
+
+# TODO: --yadm-repo option must override $yadm_repo calue
+complete -c yadm -w "git --work-tree $yadm_work_tree --git-dir $yadm_repo"
```
faho, line 46: Yeah, this won't change the work-tree at runtime. It will only be set once, when the script is loaded.

The easiest thing to do is probably to do the wrapping manually. Something like (untested):

```fish
function __fish_complete_yadm_like_git
    set -l cmdline (commandline -opc; commandline -ct)[2..-1]
    set -l yadm_work_tree (yadm gitconfig --get core.worktree)
    set -l yadm_repo (yadm introspect repo)
    complete -C"git --work-tree $yadm_work_tree --git-dir $yadm_repo $cmdline"
end

complete -c yadm -a '(__fish_complete_yadm_like_git)'
```

gokulsoumya: Thanks for the help; it works like a charm with a few modifications. There are still a few rough edges to smooth out in other regards, and I'll finish it as soon as possible.

gokulsoumya:
The completions have been finished and now react to `--yadm-repo` option. It also correctly inherits (only) needed completions from git.

share/completions/yadm.fish
```diff
@@ -0,0 +1,131 @@
...
+    string join \n v-version h/help C= c=+ 'e-exec-path=?' H-html-path M-man-path I-info-path p/paginate \
+        P/no-pager o-no-replace-objects b-bare G-git-dir= W-work-tree= N-namespace= S-super-prefix= \
+        l-literal-pathspecs g-glob-pathspecs O-noglob-pathspecs i-icase-pathspecs \
+        Y/yadm-dir= R-yadm-repo= n-yadm-config= E-yadm-encrypt= a-yadm-archive= B-yadm-bootstrap=
+end
+
+# Config values accepted by `yadm config`
+# See `yadm gitconfig` for setting git config values
+function __fish_yadm_config_keys
+    yadm introspect configs | string replace \n " "
```
faho, line 15: I don't believe this `string replace` does what you think it does.

I would imagine, and do correct me if I'm wrong, that `yadm introspect configs` prints one config per line, so they look like

```
configone
configtwo
```

What you assume happens when you have a completion like

```fish
complete -c yadm -a '(__fish_yadm_config_keys)'
```

is that the inner command's output is placed directly into the `-a` argument, so this would have to look like

```
complete -c yadm -a 'configone configtwo'
```

 That's not what happens.

Like every other fish command substitution (that doesn't use `string split`), it is split, on newlines. The command's output is taken, each line is used as a separate argument.

So, if the configs are printed one per line, there is no need to change anything, and you could directly run

```fish
complete -c yadm -a '(yadm introspect configs)'
```

Or put it into this function *without the replace*.

If that for some reason doesn't work, I would love to see some example output of `yadm introspect configs`, and some example usage for `yadm config`

gokulsoumya: You're absolutely right, `yadm introspect configs` prints one config per line and the `string replace` is redundant. Thanks for the review !

share/completions/yadm.fish
```diff
@@ -0,0 +1,131 @@
...
+    if set -q _flag_yadm_repo
+        set yadm_repo $_flag_yadm_repo
+        # argparse *always* sets $argv to remaining arguments after consuming specified options
+        set cmdline $argv
+    end
+
+    set -l git_wrapper_cmd "git --work-tree $yadm_work_tree --git-dir $yadm_repo $cmdline"
+
+    # `complete -a` expects each completion to be separated by space, not newline
+    complete -C "$git_wrapper_cmd" | string replace '\n' ' '
```
krobelus, line 74: I think there is no need for this `string replace` and the comment. fish always splits by newlines, also in `complete -a`.

share/completions/yadm.fish
```diff
@@ -0,0 +1,131 @@
...
+        P/no-pager o-no-replace-objects b-bare G-git-dir= W-work-tree= N-namespace= S-super-prefix= \
+        l-literal-pathspecs g-glob-pathspecs O-noglob-pathspecs i-icase-pathspecs \
+        Y/yadm-dir= R-yadm-repo= n-yadm-config= E-yadm-encrypt= a-yadm-archive= B-yadm-bootstrap=
+end
+
+# Config values accepted by `yadm config`
+# See `yadm gitconfig` for setting git config values
+function __fish_yadm_config_keys
+    yadm introspect configs
+end
```
krobelus, line 16: I guess this function could now be inlined, whatever you prefer.

gokulsoumya: I think I'll leave it as is, because `__fish_yadm_config_keys` seems more readable and intuitive to me than `yadm introspect configs`

share/completions/yadm.fish
gokulsoumya: Ha you're right, I didn't notice it when removing a same `string replace` call that Fabian pointed out. Thanks for pointing it out :)

krobelus:
Thanks and sorry for the delay!

PR #7112, opened 2020-06-12
Add completions for github's gh tool

gokulsoumya:
## Description

Add completions for [gh](https://github.com/cli/cli).

## TODOs:

- [ ] ~~Changes to fish usage are reflected in user documentation/manpages.~~
- [ ] ~~Tests have been added for regressions fixed~~


zanchey:
Some packaging systems add these already, which is probably the preferred approach, but I don't think there's any harm in having both.

Issue #5284, opened 2018-10-27
Interactively-defined function loses indentation when saved or edited


mqudsi:
I first defined a function interactively at the prompt, not entering any indentation manually (i.e. fish indentend my code each time I pressed `Enter`):

```fish
mqudsi@Blitzkrieg /m/d/r/fish> function replace_in
                                   # argv[1] is the regex pattern of files to match, fed to `fd`
                                   # argv[2] is the pattern of text to search for
                                   # argv[3] is the pattern of text it should be replaced with
                                   for f in (fd $argv[1] .)
                                       string replace -a -- $argv[2] $argv[3] < $f | rewrite $f
                                   end
                               end
```

The function definition shows the indentation:

```fish
mqudsi@Blitzkrieg /m/d/r/fish> type replace_in
replace_in is a function with definition
function replace_in

    # argv[2] is the pattern of text to search for
    # argv[3] is the pattern of text it should be replaced with
    for f in (fd $argv[1] .)
        string replace -a -- $argv[2] $argv[3] <$f | rewrite $f
    end
end
```

`funced` without `$EDITOR` set shows the following:

```fish
replace_in> function replace_in
                
                # argv[2] is the pattern of text to search for
                # argv[3] is the pattern of text it should be replaced with
                for f in (fd $argv[1] .)
                    string replace -a -- $argv[2] $argv[3] <$f | rewrite $f
                end
            end
```

But this is what happens after I `set -Ux EDITOR vim`, then execute `funced replace_in` again:

![image](https://user-images.githubusercontent.com/606923/47609426-d77f4780-da03-11e8-8b67-752b41b2effb.png)

ridiculousfish:
This is pretty weird. I wasn't able to reproduce it.

Might be due to vim config. What happens if you `set -x EDITOR cat`? Does `cat` show the function as indented or not?

faho:
I can reproduce as well, but it doesn't have anything to do with `funced` - `functions replace_in` already exhibits the issue.

The reason seems pretty simple, actually. When you interactively define a function, fish lies about the indentation.

When you type

```fish
$ function something
       echo a
```

that `echo a` isn't _actually_ indented - when you go to before the `e` and do a single backward-char you end up in the line before.

We probably don't want to lose that latter property - you don't want to mess with indentation in the commandline buffer - so we could either add the apparent indentation later.

[...]

PR #7114, opened 2020-06-13
Explicitly pipe through fish_indent when using funcsave

gokulsoumya:
## Description

Using `funcsave` sometimes produces a file with improper indentation. I've not
been able to reproduce it reliably, and `functions funcname > file.fish` always
seems to produce properly indeneted code. Explicitly piping through `fish_indent`
fixes the problem.

### DIFF
--- a/share/functions/funcsave.fish
+++ b/share/functions/funcsave.fish
@@ -27,7 +27,7 @@ function funcsave --description "Save the current definition of all specified fu
     set -l retval 0
     for funcname in $argv
         if functions -q -- $funcname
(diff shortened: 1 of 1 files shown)


zanchey:
Maybe caused by #5284? An alternative approach is suggested there.

gokulsoumya:
Ah I see; the problem stems from the implementation of the `functions` builtin. Since fixing it only for `funcsave` or even `funced` will not address the core problem and leave it in an inconsistent state, I'm closing this.

PR #7117, opened 2020-06-15
fish_config: 'Webify' color definitions

kppullin:
### Summary

Fix missing colors in `fish_config`'s customization palette.

### Details

The colors defined in `colorutils.js` lack the leading
`#` character and do not fully follow the html/css spec
(https://www.w3.org/TR/css-color-4/#typedef-hex-color).

While this (somehow) works for the most part, a few
colors get lost along the way and do not display in
the customization selector nor in the preview when
selected.

To fix this, let's prepend the missing `#` character
to all colors defined in `colorutils` and adjust a
few other places that were already doing that (as to
avoid doubling up the '#'s).

#### Before

Solarized - 3rd color missing:
image

Fish Default - One missing in each of the two bottom rows:
image

#### After

Solarized:
image

Fish Default:
image

## TODOs:

### DIFF
--- a/share/tools/web_config/js/colorutils.js
+++ b/share/tools/web_config/js/colorutils.js
@@ -3,33 +3,33 @@
 term_256_colors =
     [
       // 247
-      'ffd7d7', 'd7afaf', 'af8787', '875f5f', '5f0000', '870000', 'af0000', 'd70000', 'ff0000',
-      'ff5f5f', 'd75f5f', 'd78787', 'ff8787', 'ffafaf', 'ffaf87', 'ffaf5f', 'ffaf00', 'ff875f',
-      'ff8700', 'ff5f00', 'd75f00', 'af5f5f', 'af5f00', 'd78700', 'd7875f', 'af875f', 'af8700',
-      '875f00', 'd7af87', 'ffd7af', 'ffd787', 'ffd75f', 'd7af00', 'd7af5f', 'ffd700', 'ffff5f',
-      'ffff00', 'ffff87', 'ffffaf', 'ffffd7', 'd7ff00', 'afd75f', 'd7d700', 'd7d787', 'd7d7af',
-      'afaf87', '87875f', '5f5f00', '878700', 'afaf00', 'afaf5f', 'd7d75f', 'd7ff5f', 'd7ff87',
-      '87ff00', 'afff00', 'afff5f', 'afd700', '87d700', '87af00', '5f8700', '87af5f', '5faf00',
-      'afd787', 'd7ffd7', 'd7ffaf', 'afffaf', 'afff87', '5fff00', '5fd700', '87d75f', '5fd75f',
-      '87ff5f', '5fff5f', '87ff87', 'afd7af', '87d787', '87d7af', '87af87', '5f875f', '5faf5f',
-      '005f00', '008700', '00af00', '00d700', '00ff00', '00ff5f', '5fff87', '00ff87', '87ffaf',
-      'afffd7', '5fd787', '00d75f', '5faf87', '00af5f', '5fffaf', '00ffaf', '5fd7af', '00d787',
-      '00875f', '00af87', '00d7af', '5fffd7', '87ffd7', '00ffd7', 'd7ffff', 'afd7d7', '87afaf',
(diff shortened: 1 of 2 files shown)


share/tools/web_config/js/colorutils.js
```diff
@@ -81,8 +81,8 @@ function adjust_lightness(color_str, func) {
     }
 
     /* More hacks */
-    if (color_str == 'black') color_str = '000000';
-    if (color_str == 'white') color_str = 'c0c0c0';
+    if (color_str == 'black') color_str = '#000000';
+    if (color_str == 'white') color_str = '#c0c0c0';
```
charego, line 85: Nice find. But won't this break the following line (i.e. return NaN)?

```
rgb = parseInt(color_str, 16)
```

Also, the beginning of this function explicitly strips `#` from inputs so adding it back in seems confusing at best. I wonder if there's a simpler way to fix the color issue you've found.

kppullin: Thanks for calling that out... errant change from a find+replace over the file that I missed 😞 .

I haven't found a _better_ way yet to fix the issue, at least in my opinion.  Another thought that I first attempted was to apply a function in the angular attributes to apply a function to insert the `#` as needed.  This felt more hacky and non-obvious maintenance wise to me.

Interesting to note that this bug does _not_ manifest in chrome, but does in firefox.  Peeked a bit at the firefox source and it's hard to dig up where these conversions may take place with only a quick glance.  Firefox uses `rust-cssparser` for some color parsing (maybe just pure CSS and not HTML attributes?) and that project's tests fail if given a hex color without the leading `#`.  There may also be things going on in the angularjs code, though nothing I've found thus far.

share/tools/web_config/js/colorutils.js
```diff
@@ -191,32 +191,32 @@ function master_color_for_color(color_str) {
     })
 }
 
-/* Given a color name, like 'normal' or 'red' or 'FF00F0', return an RGB color string (or empty
+/* Given a color name, like 'normal' or 'red' or '#FF00F0', return an RGB color string (or empty
    string) */
 function interpret_color(str) {
```
charego, line 196: Currently `interpret_color` is not being used anywhere. I took a stab at this problem in a different way (by using this function, for one). Please check out https://github.com/charego/fish-shell/commit/b0bb1d0917a9f89b90d8df57b39a25500fea0a03 and let me know what you think. In addition to fixing the missing colors that you found, I also make a naive attempt to handle fish colors specified like '555 brblack' (which is what the "fish default" theme uses for autosuggestions).

See how autosuggestion text is white vs. gray in these examples.

Current version | Fixed version
------------ | -------------
default_theme_1 | default_theme_2

share/tools/web_config/js/colorutils.js
charego: > Another thought that I first attempted was to apply a function in the angular attributes to apply a function to insert the # as needed. This felt more hacky and non-obvious maintenance wise to me.

I guess that's what I did in https://github.com/charego/fish-shell/commit/b0bb1d0917a9f89b90d8df57b39a25500fea0a03. It does feel a bit hacky. However, I like that the theme code defines the colors in "fish color" format (without leading '#'), and there is a translation layer for display in the browser.

> Interesting to note that this bug does not manifest in chrome, but does in firefox.

Confirmed this. I wonder if the ancient AngularJs code can be blamed. The [color picker](https://github.com/fish-shell/fish-shell/blob/master/share/tools/web_config/partials/colors.html#L110-L119) uses a nested loop and in Firefox their dark entries are missing the "style" attribute. That may be due to the Firefox CSS parser as you mention, but then it's mystifying why it works in most other cases.

```html
<table>
	<tbody>
		<tr data-ng-repeat="color_array in colorArraysArray">
			<td ng-repeat="color in color_array"
			    data-ng-style="{'background-color': color, 'color': color}">...</td>
		</tr>
	</tbody>
</table>
```

share/tools/web_config/js/colorutils.js
kppullin: That's a nice approach and great that you were able to make additional improvements!  Shall I close out this PR so you can replace with yours?

charego: Let's leave your PR open for a bit longer in case others want to chime in. Depending on what happens, we might do that later. Thanks!

kppullin:
Fixed in #7179

PR #7121, opened 2020-06-15
Add fc(1posix) - process the command history list

### LINKED ISSUE #5030: feature-request: Add `fc` "fix command"  command, like the shell builtin
I just learned about the bash builtin `fc` command. It takes the last command and opens it in your editor, so you can tweak it, then when you save it, it runs that new command and adds it to history. 

> fc: usage: fc [-e ename] [-lnr] [first] [last] or fc -s [pat=rep] [command]

Would love to see fish ship this command in core.

jspricke:
## Description

This is useful to combine and rerun several commands.
For example having the last 4 commands in $history:

$ string join \n $history[1..4]
./test
git clean -dfx
ls
make

$ fc -4 -1

opens an editor where you can drop the ls, maybe move the git command to
the start, join all lines with '; and' and save. Now the prompt contains
one command, ready to be executed.

Note that the numbering is inverted to $history to comply with POSIX.

The implementation is based on share/functions/edit_command_buffer.fish.

fc is specified by POSIX in

https://pubs.opengroup.org/onlinepubs/9699919799/utilities/fc.html

For now only the first and last options are implemented.

Fixes #5030

## TODOs:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -35,6 +35,8 @@ Notable improvements and fixes
 Syntax changes and new commands
 -------------------------------
 
+- New command ``fc`` to process the command history list.
+
 Scripting improvements
 ----------------------
 
--- a/share/functions/fc.fish
+++ b/share/functions/fc.fish
@@ -1,0 +1,52 @@
+function fc --description 'Edit the command history list in an external editor'
+    set -l f (mktemp)
+    if set -q f[1]
+        mv $f $f.fish
+        set f $f.fish
+    else
+        # We should never execute this block but better to be paranoid.
+        if set -q TMPDIR
+            set f $TMPDIR/fish.$fish_pid.fish
+        else
+            set f /tmp/fish.$fish_pid.fish
+        end
+        touch $f
+        or return 1
+    end
+
+    # Edit the command line with the users preferred editor or vim or emacs.
+    set -l editor
+    if set -q VISUAL
+        echo $VISUAL | read -at editor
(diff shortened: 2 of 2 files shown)


faho:
Thanks, this is solid work.

Unfortunately you have decided that posix-compatibility is the way to go here. I strongly believe that is misguided.

Posix-compatibility has always been a non-goal for fish, and in this case it is almost completely useless (there is no script that this enables you to run because it's not used in scripts) and requires us to implement a command that is suboptimal in several ways.

1. The name is awful - I'm not sure what "fc" is supposed to stand for. Could be "fix command", could be "fancy chocolate". Either way this isn't important enough to deserve a two-letter mnemonic name, if it were mnemonic to begin with.

2. Like you say the numbering is inverted compared to what we actually use. If we kept the name there might be an argument, but we're not going to.

3. It's a separate command to begin with - I would probably expect this to either be `edit_command_buffer` with an option (`--history 5`), or something like `history edit`.

4. It has a grab bag of options and arguments that fit in nowhere else - it allows recalling a line with replacements.

Also `fc -4 -1` is just `commandline -r -- $history[1..4]`. If you have to have an editor, use `edit_command_buffer` after that.

As it is, I'm going to reject this PR, because we're not going to add an `fc` command. Sorry.

jspricke:
Fine with me and thanks for the commandline hint.
For reference, I added this to my config:
```
function fc
  commandline -r -- $history[1..$argv[1]]
end
```

Issue #7031, opened 2020-05-23
support Administrator in fish_prompt.fish


gMan1990:
- fish, version 3.1.0

- http://cygwin.com/
```plain
> uname -a
CYGWIN_NT-10.0 3.1.4(0.340/5/3) 2020-02-19 08:49 x86_64 Cygwin
```

- fish_prompt.fish
```plain
    if contains -- $USER root toor Administrator
```

faho:
Yeah this should probably be a helper function - `fish_is_root_user`.

That way we wouldn't have to change it in *every single prompt*. 9 of our 16 prompts do this, one only checks for "root".

damani42:
I'm really interesting to work on this issue. I will make this soon as possible.

PR #7123, opened 2020-06-17
Add a function to check if the user is root.

damani42:
Add a helper function to check if the user is root. This function can be
useful for the prompts for example. Modify the prompts made root checked
to use the function instead. Add also the support of Administrator like
a root user.

Fixes: #7031

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -31,6 +31,7 @@ Notable improvements and fixes
 - ``fish_preexec`` and ``fish_postexec`` events are no longer triggered
   for empty commands.
 - The ``test`` builtin now better shows where an error occured (#6030).
+- Add a helper function to know if the user is root (#7031).
 
 Syntax changes and new commands
 -------------------------------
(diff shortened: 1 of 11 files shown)


faho:
Thank you for your effort!

What I'd really like to see, and what makes the entire thing with the function worthwhile, is to have it used in *all* of our sample prompts that do this check. E.g. the "debian chroot" prompt has this bit: https://github.com/fish-shell/fish-shell/blob/f2d5f95396ff95d024208a74de3e68657e42cd2b/share/tools/web_config/sample_prompts/debian_chroot.fish#L22-L33

In essence, whenever a prompt checks if the user is "root" anywhere, it should be using that function instead. That includes now using the function whether it used to check for "root" or "toor" or just for "root". The actual coloring and using symbols can stay in the prompt, however.

So: If you're feeling up for it, please check the sample_prompts directory for such uses and replace them with this function.

damani42:
I think now I have modified all prompts. Do you see something else, I can modify?

faho:
Thank you very much, merged!

PR #7124, opened 2020-06-17
Typo.

damani42:
Replace the tabulation by space.


faho:
Yep, nice catch!

It would be nice to have `fish_indent` working on these blocks, but I have no idea how to convince sphinx to pass it there.

zanchey:
The output is run through `fish_indent` (I think), so it shouldn't matter too much.

Issue #6981, opened 2020-05-07
completions: fix long descriptions


ammgws:
# Issue

There are quite a few completions that have very long descriptions, most likely auto-generated ones from the manpage haven't been fixed up by anyone. I think we should list these up so people can work on it in chunks since it's a mundane task...

<details><summary>I ran a simple script to check for all descriptions >80 characters (arbitrary number):</summary>
<p>

```fish
for f in share/completions/*.fish
     set desc_str (string replace -rf "^complete.*-d\s?[\"|']([^-]*)[\"|']\s?.*" '$1' <$f)
     set n 0
     for d in $desc_str
         if test (echo $d | wc -c) -gt 80
             if test $n -lt 1
                 echo $f
                 set n 1
             end
             echo (echo $d | wc -c) chars: $d
         end
     end
     if test $n -gt 0
         echo --------------------------
     end
 end
```
</p>
</details>

Script output is here: https://gist.github.com/ammgws/a0f49d719a5145675c76ff161627c111

# Progress
- [x] adduser.fish (#7550)
- [x] animate.fish
- [x] ansible-playbook.fish
- [x] ansible.fish (9ffaade0db3392801a41c649486e21574151bb53)
- [x] ant.fish
- [x] atom.fish (848f3ac9)
- [x] awk.fish (#7125)
- [x] badblocks.fish (#7788)
- [x] bison.fish (https://github.com/fish-shell/fish-shell/pull/8658/commits/8b423206aafb6815caddda9f0ae946b94288edc1)
- [x] btrfs.fish (f88d7dd31249e7fc338a037425654d72923eea4d)
- [x] bundle.fish (#8034)
- [ ] bzr.fish
- [x] cdrecord.fish
- [ ] cf.fish
- [x] compare.fish
- [ ] composer.fish
- [x] composite.fish
- [x] convert.fish
- [ ] csc.fish
- [x] curl.fish (#7788)
- [ ] cwebp.fish
- [ ] darcs.fish
- [x] dhclient.fish (#8034)
- [x] display.fish
- [x] entr.fish
- [ ] exercism.fish
- [x] feh.fish (#7109)
- [x] find.fish (c065d24632e12c9988bd6629be9b6a02c6d1fdc6)
- [x] fsharpc.fish (#7569)
- [x] fsharpi.fish (#7569)
- [x] fzf.fish (#7081)
- [ ] gcc.fish
- [ ] gdb.fish
- [x] git.fish (#7752)
[...]

PR #7125, opened 2020-06-17
Fix long description in completion.

damani42:
Simplify the description, I think it's sufficient like this.

Fixes: #6981


faho:
Yeah, this looks like someone copied it directly from the standard. Thanks!

Issue #7102, opened 2020-06-09
Feature request: string pad


pmlandwehr:
Would it be possible to get a `string pad` function to complement `string trim`?  I realizing that padding can effectively be accomplished using things like`printf "%<n or negative n>s\n" $strs_to_pad | tr " " "<padding char>"`, but it seems like a natural extension.

I'm at fish, version 3.1.2 on Mac OS X

m0nhawk:
I made an attempt to implement that. It works like that:

```
❯ ./fish -c "string pad -l -r -n 10 --chars c test"
cccccccccctestcccccccccc
❯ ./fish -c "string pad -l -n 4 --chars ' ' foo"
    foo
```

ridiculousfish:
@m0nhawk apologies for long delay, I'd be happy to review your changes if you would like to open a PR! Let us know if you have questions about how to do so.

PR #7127, opened 2020-06-18
builtin_string: add pad command

m0nhawk:
## Description

Added string pad command.

Fixes issue #7102

## TODOs:

### DIFF
--- a/doc_src/cmds/string-pad.rst
+++ b/doc_src/cmds/string-pad.rst
@@ -8,7 +8,7 @@ Synopsis
 
 ::
 
-    string pad [(-l | --left)] [(-r | --right)] [(-c | --char) CHAR] [(-n | --count) INTEGER] [(-q | --quiet)] [STRING...]
+    string pad [(-l | --left)] [(-r | --right)] [(-c | --char) CHAR] [(-n | --count) INTEGER] [(-m | --max) INTEGER] [(-q | --quiet)] [STRING...]
 
 .. END SYNOPSIS
 
@@ -17,7 +17,7 @@ Description
 
 .. BEGIN DESCRIPTION
 
-``string pad`` pads before and after the string specified character for each STRING. If ``-l`` or ``--left`` is given, only padded before string. Left only is the default padding. If ``-r`` or ``--right`` is given, only padded after string. The ``-c`` or ``--char`` switch causes the characters in CHAR to be padded instead of whitespace. The ``-n`` or ``--count`` integer specifies the amount of characters to be padded. The default padding count is 0. Exit status: 0 if string was padded, or 1 otherwise.
+``string pad`` pads before and after the string specified character for each STRING. If ``-l`` or ``--left`` is given, only padded before string. Left only is the default padding. If ``-r`` or ``--right`` is given, only padded after string. The ``-c`` or ``--char`` switch causes the characters in CHAR to be padded instead of whitespace. The ``-n`` or ``--count`` integer specifies the amount of characters to be padded. The default padding count is 0. The ``-m` or ``--max`` integer specifies the resulting length of the after after adding pad characters. Note, that the ``-n`` and ``-m`` flags are mutually exclusive, you can only use one of them. Exit status: 0 if string was padded, or 1 otherwise.
 
 .. END DESCRIPTION
 
--- a/doc_src/cmds/string-pad.rst
+++ b/doc_src/cmds/string-pad.rst
@@ -17,7 +17,23 @@ Description
 
 .. BEGIN DESCRIPTION
 
-``string pad`` pads before and after the string specified character for each STRING. If ``-l`` or ``--left`` is given, only padded before string. Left only is the default padding. If ``-r`` or ``--right`` is given, only padded after string. The ``-c`` or ``--char`` switch causes the characters in CHAR to be padded instead of whitespace. The ``-n`` or ``--count`` integer specifies the amount of characters to be padded. The default padding count is 0. The ``-m` or ``--max`` integer specifies the resulting length of the after after adding pad characters. Note, that the ``-n`` and ``-m`` flags are mutually exclusive, you can only use one of them. Exit status: 0 if string was padded, or 1 otherwise.
+``string pad`` pads before and after the string specified character for each STRING.
+
+The default behaviour is left padding with spaces and default count or max is zero.
+
+If ``-l`` or ``--left`` is given, only padded before string.
+
+If ``-r`` or ``--right`` is given, only padded after string.
+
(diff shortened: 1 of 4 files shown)


doc_src/cmds/string-pad.rst
```diff
@@ -0,0 +1,42 @@
+string-pad - pad characters before and after string
+========================================
```
damani42, line 2: Continue the equal sign to the end of the sentence.

pmlandwehr:
Could you add a switch for padding **to** a length instead of padding **with** a length?

m0nhawk:
@pmlandwehr Sounds useful. I can, but it may require additional discussion for combination with other flags:

- What if the *to* length smaller than a string? Should it yield an error or work as *trim*?
- What about both `-l` and `-r`? How should the string be wrapped, taking in account the odd/even length of string and *to* specification?

I don't want to implement something, which will have a bad design and usage. :)

pmlandwehr:
> What if the to length smaller than a string? Should it yield an error or work as trim?

To go with the `printf` version, it would just not add any characters - b/c you are already *N* chars, no padding is needed.  The best way to do this might be to steal the `-m/max` flag from `string repeat` to limit the number of chars. 

> What about both -l and -r? How should the string be wrapped, taking in account the odd/even length of string and to specification?

I think that for padding to a max length you should be required to only provide either `-l` or `-r`. To provide both flags would raise an error. To get fancier behavior you could pipe s sequence of pads together:

``` 
~> string pad -r -n 3 -c "-" blah | string pad -l  -m 10 -c "-"
---blah---
~> string pad -r -n 3 -c "-" hi | string pad -l  -m 10 -c "-"
---hi-----
~> string pad -r -n 3 -c "-" babadook | string pad -l  -m 10 -c "-"
---babadook
```

If we wanted simultaneous `-r` and `-l` flags  with `-m` I would try to support centering, putting (*N*-*string length*)/2 chars on either side of the string. If *N* is odd, the left side always gets the extra character:
```
~> string pad -l -r -m 10 -c "-" hi
----hi----
~> string pad -l -r -m 10 -c "-" two
----one---
~> string pad -l -r -m 10 -c "-" five
---five---
~> string pad -l -r -m 10 -c "-" verylongword
verylongword
```

For string wrapping, are we talking about newlines included within passed strings? I would think that the newlines should just be treated as a single character - this would create a kind of gross wrapping that tacks on the new string chars after the newline, but seems like most literal treatment.

m0nhawk:
I made a necessary changes to support max value and preferring left padding, over right.

Now it supports all the examples you provided. And, of course, `--count` and `--max` are exclusive:

```
❯ ./fish -c "string pad -l -r --max 10 --count 4 -c "-" verylongword"
pad: Invalid combination of options,
--count and --max are mutually exclusive
```

doc_src/cmds/string-pad.rst
```diff
@@ -0,0 +1,42 @@
...
+--------
+
+.. BEGIN EXAMPLES
+
+::
+
+    >_ string pad -l -n 10 -c ' ' 'abc'
+              abc
+
+    >_ string pad --right --count 5 --char=z foo bar
```
krobelus, line 34: Is there a reason for `--max` not being the default? I imagine this is the more common use case.
JavaScript ships with a function that does exactly that: `"foo".padStart(4)` gives `"  foo"`.

src/builtin_string.cpp
```diff
@@ -1244,6 +1255,83 @@ static int string_collect(parser_t &parser, io_streams_t &streams, int argc, wch
...
+            if (opts.left && !opts.right) {
+                pad_left = opts.max - padded.size();
+            }
+        }
+
+        if (pad_left) {
+            padded.insert(0, pad_left, opts.char_to_pad);
+        }
+
+        if (pad_right) {
```
krobelus, line 1321: The `if` seems redundant - I believe nothing happens if `pad_right` is zero. Same above.

src/builtin_string.cpp
```diff
@@ -1244,6 +1255,83 @@ static int string_collect(parser_t &parser, io_streams_t &streams, int argc, wch
...
+        streams.err.append_format(BUILTIN_ERR_COMBO2, cmd,
+                                  _(L"--count and --max are mutually exclusive"));
+        return STATUS_INVALID_ARGS;
+    }
+
+    size_t npad = 0;
+
+    arg_iterator_t aiter(argv, optind, streams);
+    while (const wcstring *arg = aiter.nextstr()) {
+        wcstring padded = wcstring(*arg);
```
krobelus, line 1289: `wcstring padded_arg = *arg;` should be equivalent.

m0nhawk:
I have made a couple of changes to address the PR comment.

Regarding the default arguments and additional syntax changes:

The `max` by default does makes sense, but I'm not sure how to implement that. The `handle_flag_m` currently requires the integer value, so I'm not sure if I need to change that or have a new flags for `max`/`count` specifically for `string pad`.

krobelus:
I find it more logical to say `string pad 7 blah` instead of `string pad -n 7 blah`. However, `string repeat` also works with `-n` and it's probably better to be consistent, so I think it is fine to have two different flags that take numbers.

I don't know if `max`/`count` are the best names for `string pad`. Instead of `max`, one of `length` or `min` seems more appropriate because the result of padding creates a string of at least that length. I think we can get rid of `-l/--left` to free up the space for a possible `-l/--length`. If a user wants to pad on both sides, they can just invoke `string pad` twice.

Adding new flags is a possibility; of course reusing flags from other `string` commands could make them easier to remember.

ridiculousfish:
Thank you to @m0nhawk for this contribution. I like this command, and I think it can be simplified. I agree with krobelus's suggestions for removing `--left`, and that it should have a min width.

String padding with a count can be done with just catenation:

    echo "   "$var

so we only need to accept a width.

Instead of `length` I suggest `width`, because of Unicode. For example it should know that a emoji is width 2, and pad it less. The reason is that I think this will be used in prompts which often include emoji.

Lastly it should return success unless there was an error in the argument list. (I think it actually does that but the docs say it returns failure if the string was not padded).

m0nhawk do you have the appetite to incorporate those changes? If not I completely understand and I can take it over.

faho:
There've been unaddressed review comments on this for a while, so I'm closing this.

If @m0nhawk (or anyone else) wishes to pick it up again, we can of course reopen!

m0nhawk:
This PR has flew off my radar, finally managed to find time to address the comments, and I agree, having `width` option look much more slick and easy to grasp.

PS: Somehow this PR didn't get an updates of latest commits and I've created a new one: #7340

krobelus:
> PS: Somehow this PR didn't get an updates of latest commits

I think that only happens after you reopen it, but let's move to #7340

PR #7128, opened 2020-06-18
Migrate ll and la function.

damani42:
Migrate ll and la function to use abbr. In this case it's better we
don't need to do anything special with argv. Use the global scope to be
faster.

### DIFF
--- a/share/config.fish
+++ b/share/config.fish
@@ -42,6 +42,9 @@ if status --is-interactive
         set -q fish_term24bit
         or set -g fish_term24bit 1
     end
+
+    abbr --add --global ll ls -lh
+    abbr --add --global la ls -lah
 else
     # Hook up the default as the principal command_not_found handler
     # in case we are not interactive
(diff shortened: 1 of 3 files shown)


faho:
Thank you for your effort, but we don't want to be shipping pre-defined abbrs at the moment. 

They're really best when the user defines them the themselves, rather than just "magic" that sometimes changes some tyoed letters to others (and is annoying to disable) 

There's an argument to be made that these wrappers should be removed entirely, but that would require some discussion.

damani42:
> Thank you for your effort, but we don't want to be shipping pre-defined abbrs at the moment.

You're welcome, I understand.
 
> They're really best when the user defines them the themselves, rather than just "magic" that sometimes changes some tyoed letters to others (and is annoying to disable)

I agree with you. It's easier to disable with abbr I think.
 
> There's an argument to be made that these wrappers should be removed entirely, but that would require some discussion.

Maybe I can do the patch to open the discussion?

PR #7130, opened 2020-06-18
Fix #7113, correct a few regexes.

### LINKED ISSUE #7113: Cannot call help using msys2
_fish version 3.1.2_

Msys2 has the `cygpath` command but no `cygstart`.

This result in `cygpath` not being called:
https://github.com/fish-shell/fish-shell/blob/27f9920c7f3f0a7438f50f7a310003bfb67808ec/share/functions/help.fish#L139-L143

When no browser are set and since it doesn't find `cygstart` it defaults to `cmd.exe` however because `cygpath` is not called the path is wrong which make this : `/c/Windows/System32/cmd.exe /c 'start file:///usr/share/doc/fish/index.html'` fail.

Instead of exiting with an error it then returns to cmd, typing `exit` will return to fish:

image

From my understanding this could be resolved by changing the code quoted above to this:
```fish
        if type -q cygpath
            and string match -qr 'cygstart|.*cmd\.exe' $fish_browser[1]
            set page_url file://(cygpath -m $__fish_help_dir)/$fish_help_page
```

I can submit a push request if needed.

Edit: I updated the regex from `cygstart|cmd\.exe` to `cygstart|.*cmd\.exe`

malobre:
## Description

This allows the `help` command to work with `msys2` which has the `cygpath` command like cygwin, but no `cygstart`. The changes allow path resolution with `cygpath` without using `cygstart`.

I've also fixed a few regexes.

So far the changes were only tested with `msys2`.

Fixes issue #7113

## TODOs:

### DIFF
--- a/share/functions/help.fish
+++ b/share/functions/help.fish
@@ -80,7 +80,7 @@ function help --description 'Show help for the fish shell'
             # We use this instead of xdg-open because that's useless without a backend
             # like wsl-open which we'll check in a minute.
             if test -f /proc/version
-                and string match -riq 'Microsoft|WSL' </proc/version
+                and string match -riq 'Microsoft|WSL|MSYS|MINGW' </proc/version
                 and set -l cmd (command -s cmd.exe /mnt/c/Windows/System32/cmd.exe)
                 # Use the first of these.
                 set fish_browser $cmd[1]
@@ -136,13 +136,13 @@ function help --description 'Show help for the fish shell'
         # Help is installed, use it
         set page_url file://$__fish_help_dir/$fish_help_page
 
-        # For Windows (Cygwin and WSL), we need to convert the base help dir to a Windows path before converting it to a file URL
+        # For Windows (Cygwin, msys2 and WSL), we need to convert the base help dir to a Windows path before converting it to a file URL
         # but only if a Windows browser is being used
         if type -q cygpath
-            and string match -qr cygstart $fish_browser[1]
+            and string match -qr 'cygstart|cmd\.exe$' $fish_browser[1]
             set page_url file://(cygpath -m $__fish_help_dir)/$fish_help_page
         else if type -q wslpath
-            and string match -qr '.exe' $fish_browser[1]
+            and string match -qr 'cmd\.exe$' $fish_browser[1]
             set page_url file://(wslpath -w $__fish_help_dir)/$fish_help_page
         end
     else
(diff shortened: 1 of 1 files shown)


share/functions/help.fish
```diff
@@ -193,7 +194,7 @@ function help --description 'Show help for the fish shell'
     else
         # Work around lynx bug where <div class="contents"> always has the same formatting as links (unreadable)
         # by using a custom style sheet. See https://github.com/fish-shell/fish-shell/issues/4170
-        if string match -q 'lynx*' -- $fish_browser
+        if string match -q '^lynx' -- $fish_browser
```
krobelus, line 197: If you use `^`, you need to use the `--regex`/`-r` flag, by default `string match` will do glob-style matching.

malobre: My bad, didn't see it was `-q` not `-qr` so I read it as a regex, should I revert this change or switch to `-qr` ?

krobelus: Both work fine, you choose :)

malobre: Well, since all but one `string match` uses regex matching I'll switch to `-qr` to homogenize.

share/functions/help.fish
```diff
@@ -136,13 +136,13 @@ function help --description 'Show help for the fish shell'
...
-        # For Windows (Cygwin and WSL), we need to convert the base help dir to a Windows path before converting it to a file URL
+        # For Windows (Cygwin, msys2 and WSL), we need to convert the base help dir to a Windows path before converting it to a file URL
         # but only if a Windows browser is being used
         if type -q cygpath
-            and string match -qr cygstart $fish_browser[1]
+            and string match -qr 'cygstart|cmd\.exe$' $fish_browser[1]
             set page_url file://(cygpath -m $__fish_help_dir)/$fish_help_page
         else if type -q wslpath
-            and string match -qr '.exe' $fish_browser[1]
+            and string match -qr 'cmd\.exe$' $fish_browser[1]
```
faho, line 145: I'm not sure if this needs to be done *only* for cmd.exe?

malobre: Indeed, I was thinking about `\.exe$` but in a WSL environment there might be programs without the `.exe` extension. So this check might be superfluous ?

krobelus: The check makes sense as far as I can tell: if the browser is a non-Windows executable, like `w3m` instead of `firefox.exe`, then we don't want to convert the path.

malobre: I see, I'll switch this to `\.exe$` tommorow. (I know there is a similar match a little further, I'll change that too).

malobre: I think `\.exe *` might be better since it allows `$fish_browser` to contains both `browser.exe` and `browser.exe --with-arguments`. What do you think ?

krobelus: I believe so. `\.exe\b` comes to mind.

malobre: That would match `some.exe--wrong-arg` without the whitespace after exe.
I don't know if that's a problem or not, maybe I'm going too far. Otherwise `\.exe(\s+|$)` seems good.

faho:
Merged, thanks!

PR #7135, opened 2020-06-19
Fix redirection in `adb shell`

Erutuon:
`adb shell` by default sends stderr from the executed command to stdout, so that `adb pull nonexistent<TAB>` will complete the path using the error message from the `find` command: `adb pull find:\ nonexistent\*:\ No\ such\ file\ or\ directory`. If `>` is escaped, so that redirection is done inside the command executed by `adb shell`, no completion will be shown, as expected.

### DIFF
--- a/share/completions/adb.fish
+++ b/share/completions/adb.fish
@@ -49,7 +49,7 @@ function __fish_adb_run_command -d 'Runs adb with any -s parameters already give
 end
 
 function __fish_adb_list_packages
-    __fish_adb_run_command pm list packages 2>/dev/null | string replace 'package:' ''
+    __fish_adb_run_command pm list packages 2\>/dev/null | string replace 'package:' ''
 end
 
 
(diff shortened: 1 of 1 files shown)


share/completions/adb.fish
```diff
@@ -49,7 +49,7 @@ function __fish_adb_run_command -d 'Runs adb with any -s parameters already give
 end
 
 function __fish_adb_list_packages
-    __fish_adb_run_command pm list packages 2>/dev/null | string replace 'package:' ''
+    __fish_adb_run_command pm list packages 2\>/dev/null | string replace 'package:' ''
```
faho, line 52: Please add a comment explaining this cursedness.

Something along the lines of "we need to redirect *inside* the adb shell, because it sends the stderr to its stdout".

Erutuon: I wrote `# Escape > to redirect stderr inside adb shell, because it sends stderr of the command to its stdout.` but forgot about actually pushing it and now the commit has been merged. Oh well.

faho: Oh, no worries. I added explanation later in 9ae880f6a.

PR #7137, opened 2020-06-20
Uncomment most recent comment if commandline is empty when using toggle comment binding

gokulsoumya:
## Description

The current behaviour of `__fish_toggle_comment_commandline` function (bound to `Alt+#`) is to do nothing if the command line is empty. This PR recalls the last comment and uncomments it instead if the command line is empty. This can be achieved by hitting '#' and 'Up', but it feels too clunky IMO and the first suggestion might not be what you want if an innocent `#` appeared in a previous command.

## TODOs:

### DIFF
--- a/share/functions/__fish_toggle_comment_commandline.fish
+++ b/share/functions/__fish_toggle_comment_commandline.fish
@@ -4,12 +4,13 @@
 # retrieving the command from the shell history and removing the comment chars.
 #
 # This deliberately does not execute the command when removing the comment characters to give you an
-# opportunity to modify the command.
+# opportunity to modify the command. Also if the commandline is empty, the most recently commented
+# out history item is uncommented and presented.
 
 function __fish_toggle_comment_commandline --description 'Comment/uncomment the current command'
     set -l cmdlines (commandline -b)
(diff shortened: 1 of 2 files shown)


share/functions/__fish_toggle_comment_commandline.fish
```diff
@@ -4,12 +4,13 @@
...
 # This deliberately does not execute the command when removing the comment characters to give you an
-# opportunity to modify the command.
+# opportunity to modify the command. Also if the commandline is empty, the most recently commented
+# out history item is uncommented and presented.
 
 function __fish_toggle_comment_commandline --description 'Comment/uncomment the current command'
     set -l cmdlines (commandline -b)
     if test "$cmdlines" = ""
-        return
+        set cmdlines (history search -p "#" -z | read -z)
```
faho, line 13: When you're doing `read`, you can just give that a variable name:

```fish
history search -p "#" -z | read -z cmdlines
```

gokulsoumya: Ah right, thanks for pointing that out :)

faho:
Merged, thanks!

PR #7140, opened 2020-06-21
algorithm conversions

neheb:
## Description

Removes several for loops in favor of STL algorithms.

### DIFF
--- a/src/builtin_disown.cpp
+++ b/src/builtin_disown.cpp
@@ -61,14 +61,12 @@ int builtin_disown(parser_t &parser, io_streams_t &streams, wchar_t **argv) {
         // Stopped jobs can be disowned (they will be continued).
         // Foreground jobs can be disowned.
         // Even jobs that aren't under job control can be disowned!
-        job_t *job = nullptr;
-        for (const auto &j : parser.jobs()) {
-            if (j->is_constructed() && (!j->is_completed())) {
-                job = j.get();
-                break;
-            }
-        }
+        auto it = std::find_if(parser.jobs().begin(), parser.jobs().end(),
+                               [](const std::shared_ptr<job_t> &j) {
+                                   return j->is_constructed() && (!j->is_completed());
+                               });
 
+        job_t *job = (it != parser.jobs().end()) ? it->get() : nullptr;
         if (job) {
             retval = disown_job(cmd, parser, streams, job);
         } else {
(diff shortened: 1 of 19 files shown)


src/builtin_disown.cpp
```diff
@@ -101,9 +101,9 @@ int builtin_disown(parser_t &parser, io_streams_t &streams, wchar_t **argv) {
         }
 
         // Disown all target jobs
-        for (const auto &j : jobs) {
-            retval |= disown_job(cmd, parser, streams, j);
-        }
+        return std::accumulate(jobs.begin(), jobs.end(), retval, [&](int r, job_t *j) {
+            return r || disown_job(cmd, parser, streams, j);
```
krobelus, line 105: Looks like you have uncovered a bug in this linter - this should use bitwise "or" to always disown all jobs.

```cpp
#include <stdio.h>
#include <vector>
#include <numeric>

int main() {
	std::vector<int> vs = {0, 1, 2};

	int r1 = 0;
	for (const auto &v : vs)
		r1 |= v;

	int r2 = std::accumulate(vs.begin(), vs.end(), 0, [&](int r, int v) {
		return r || v; // should be r | v
	});

	printf("%d\n", r1); // prints 3
	printf("%d\n", r2); // prints 1
}
```

neheb: Good catch.

All the changes were done manually. This was a mistake on my part.

mqudsi:
I'm not sure if I'm the only one or not, but while some of these are nice the majority increase the verbosity while simultaneously making it less readable. But maybe it's just me?

faho:
>while some of these are nice the majority increase the verbosity while simultaneously making it less readable. But maybe it's just me?

Not just you. But I think I might kinda hate modern C++ style.

Is there any appreciable difference in behavior here? Any performance benefits? Anything that's easier to extend or understand?

Because "avoid for-loops at all costs" is definitely not my style.

neheb:
accumulate generates slightly more efficient code: https://godbolt.org/z/nefgs7 as an example.

No real idea about the other ones. Maybe count_if as well.

They also happen to be constexpr in C++20. Probably not relevant here.

krobelus:
I like the `any_of` one, the others are a bit awkward.

neheb:
Eh. I only agree about the find_if ones.

neheb:
Rebased.

ridiculousfish:
Thank you for the PR and your patience in review. However I am going to decline this. I agree with krobelus that we should prioritize readability. Beyond that for loops have other nice properties, like that you can `return` from inside them.

PR #7144, opened 2020-06-21
Add completion for `systemctl --failed`

moverest:


faho:
That's a nice catch. Thanks!

PR #7146, opened 2020-06-22
completions/function: Update missing completions

gokulsoumya:
## Description

Added missing completions for the `function` builtin:

- Complete signals with --on-signal
- Complete variables with --on-variable and --inherit-variable
- Complete event handlers with --on-event
- Complete commands with --wraps
- Add `complete` spec for --inherit-variable


faho:
Nice work, thanks!

PR #3723, opened 2017-01-11
Implemented Dracula colors

beefsack:
## Description

I've implemented a color scheme using the [Dracula colors](https://github.com/dracula/dracula-theme).  I'm currently customising `/usr/share/tools/web_config/js/colorutils.js` directly to use it, but creating a PR in case it's useful to others.

Fixes issue N/A

## TODOs:

- [ ] Changes to fish usage are reflected in user documenation/manpages.
- [ ] Tests have been added for regressions fixed

PR #7147, opened 2020-06-22
Angular update

faho:
## Description

As recently discussed on gitter, webconfig uses angular 1.0.8, from 2013.

This ports it to angular 1.8, which entered a 3 year LTS period in 2018. Hey, that's almost new.

As far as I can tell that means angular is dead, so we'd have to figure out something to move *to*.

So this is more of a stop gap.

(also webconfig is in dire need of some love anyway)

## TODOs:

Note that I have only applied light testing - I clicked through all the tabs and they seemed to work for me.


charego:
Clicking the Dracula theme gives this error and breaks the "Customize" button.
```
Error: "[ngRepeat:dupes] Duplicates in a repeater are not allowed.
Use 'track by' expression to specify unique keys.
Repeater: color in color_array, Duplicate key: string:44475A, Duplicate value: 44475A
https://errors.angularjs.org/1.8.0/ngRepeat/dupes?p0=color%20in%20color_array&p1=string%3A44475A&p2=44475A"
```

That's the only regression I see from a cursory look through all the tabs.

faho:
That could actually be a bug in the colorscheme? I'm not sure what that "colors" array is good for.

Also is it possible that some of the colorschemes are either broken themselves or have a wrong background color? For "Snow Day" and "Tomorrow" the suggestions aren't visible to me (but my color vision is impaired, so I'm never quite sure).

zanchey:
> Also is it possible that some of the colorschemes are either broken themselves or have a wrong background color?

Look ok here, though also note #7117

charego:
Dracula theme has always had a duplicate color (https://github.com/fish-shell/fish-shell/pull/3723).

> I'm not sure what that "colors" array is good for.

These are the selectable colors that show up when you press "Customize". It's fine to have 11 colors in the array, as evidenced by other themes in colorutils.js. I believe the fix is to remove the duplicate.

Also came across an official Dracula fish theme:
- https://draculatheme.com/fish
- [dracula.fish](https://github.com/dracula/fish/blob/191a3573f155599bdd80cc644805c1914fd0c06b/dracula.fish)

It differs from the settings we have. But that's for a different PR.

ridiculousfish:
This seems to work OK for me too with light testing.

faho:
Yeah, merged with the Dracula fix, let's see if it shakes anything else out.

Issue #3812, opened 2017-02-04
git config options not tab completing


adymitruk:
git config is one of the more frustrating commands to get right. At the very least, it should show --system, --local and --global. But then, it should also auto complete all possibilities like git config log.decorate and core.autocrlf

cprieto:
Git config is pretty long, for example, if we want to do this:

```
git config --<tab> and show [add|replace|unset] etc... 
```

That will be easy but it can be complicated because extended completion as

```
git config <tab>
```

and then show *all* the git config options available (which are a lot of them).

I propose using the simplest right now, just completion of double dash values, what do you think about it?

faho:
From my count, it's about 400 options. That would probably be okay by itself (though of course it'd make the completion script even further if we can't come up with a way to generate it programmatically).

Another solution is to complete _categories_ (i.e. the part before the first ".") first, so it would show 

> advice. (Help messages for new users)

> core. (Core settings)

> color. (Color settings)

and so on.

When you then complete "color.", it'd show

> color.diff (Add color to patches in diff, log, show)

And if you completed that, it would allow a value of `always`, `auto`, `true` and `false`.

Note that this is a large task if we don't find a way to automate it, so the most important options (not "core.packedGitLimit", which is probably only useful if you're paying people to know about it, but stuff like color or gpg signing or maybe even autocrlf) are okay as a first pass.

PR #7150, opened 2020-06-23
Improve git config completions

gokulsoumya:
## Description

All accepted config variables are now shown in the completions for `git config` . Also config values are shown as description if set.

Fixes issue #3812

### DIFF
--- a/share/completions/git.fish
+++ b/share/completions/git.fish
@@ -550,7 +550,7 @@ end
 
 function __fish_git_config_keys
     # Print already defined config values first
-    git config -l | string match -r '[^=]+'
+    git config -l | string replace '=' \t
     # Print all recognized config keys; duplicates are not shown twice by fish
     printf '%s\n' (__fish_git help --config)[1..-2] # Last line is a footer; ignore it
 end
--- a/share/completions/git.fish
+++ b/share/completions/git.fish
@@ -550,7 +550,12 @@ end
 
 function __fish_git_config_keys
     # Print already defined config values first
-    git config -l | string replace '=' \t
+    # Config keys may span multiple lines, so parse using null char
+    # With -z, key and value are separated by space, not "="
+    git config -lz | while read -lz key value
+        # Print only first line of value(with an ellipsis) if multiline
+        printf '%s\t%s\n' $key (string replace \n …\n $value)[1]
+    end
     # Print all recognized config keys; duplicates are not shown twice by fish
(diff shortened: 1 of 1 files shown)


share/completions/git.fish
```diff
@@ -549,7 +549,10 @@ function __fish_git_needs_command
 end
 
 function __fish_git_config_keys
-    git config -l | string match -r '[^=]+'
+    # Print already defined config values first
+    git config -l | string replace '=' \t
```
faho, line 553: Unfortunately config values can be multiline, so this would have to use `git config -lz`:

```fish
git config -lz | while read -lz key value
    printf '%s\t%s\n' $key (printf %s $value)[1]
end
```

- The "-z" format uses a space, not an "=" as a separator, so reading into two variables just works
- The inner "printf" is to just use the first line of each value

gokulsoumya: Ah right, I made the changes and also appended an ellipsis to the first line of multiline values.

share/completions/git.fish
```diff
@@ -549,7 +549,15 @@ function __fish_git_needs_command
...
-    git config -l | string match -r '[^=]+'
+    # Print already defined config values first
+    # Config keys may span multiple lines, so parse using null char
+    # With -z, key and value are separated by space, not "="
+    git config -lz | while read -lz key value
+        # Print only first line of value(with an ellipsis) if multiline
+        printf '%s\t%s\n' $key (string replace \n …\n $value)[1]
+    end
+    # Print all recognized config keys; duplicates are not shown twice by fish
+    printf '%s\n' (__fish_git help --config)[1..-2] # Last line is a footer; ignore it
```
faho, line 560: For me the last line is a footer and the line before it is *empty*, so I'd assume this needs a `-3`?

gokulsoumya:
I did notice that the second last line in blank, but the extra blank line seemed to be trimmed and I didn't give it anymore thought. I'll change the range to `[1..-3]` then. Is there any reason other than clarity though ?

share/completions/git.fish
gokulsoumya: I did notice that the second last line in blank, but the extra blank line seemed to be trimmed and I didn't give it anymore thought. I'll change the range to `[1..-3]` then. Is there any reason other than clarity though ?

faho: Oh, right, we trim blank entries already.

No, that's okay then and the change isn't needed.

faho:
Merged, thanks!

PR #7151, opened 2020-06-23
Add code of conduct

faho:
This adopts the Contributor Covenant Code of Conduct.

There's no email address given because we do not, currently, have a project email address that I'm aware of.

## TODOs:

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,76 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as
+contributors and maintainers pledge to making participation in our project and
+our community a harassment-free experience for everyone, regardless of age, body
+size, disability, ethnicity, sex characteristics, gender identity and expression,
+level of experience, education, socio-economic status, nationality, personal
+appearance, race, religion, or sexual identity and orientation.
+
+## Our Standards
+
+Examples of behavior that contributes to creating a positive environment
+include:
+
+* Using welcoming and inclusive language
+* Being respectful of differing viewpoints and experiences
+* Gracefully accepting constructive criticism
+* Focusing on what is best for the community
+* Showing empathy towards other community members
+
+Examples of unacceptable behavior by participants include:
+
+* The use of sexualized language or imagery and unwelcome sexual attention or
+ advances
+* Trolling, insulting/derogatory comments, and personal or political attacks
+* Public or private harassment
+* Publishing others' private information, such as a physical or electronic
+ address, without explicit permission
+* Other conduct which could reasonably be considered inappropriate in a
+ professional setting
+
+## Our Responsibilities
+
+Project maintainers are responsible for clarifying the standards of acceptable
+behavior and are expected to take appropriate and fair corrective action in
+response to any instances of unacceptable behavior.
+
+Project maintainers have the right and responsibility to remove, edit, or
+reject comments, commits, code, wiki edits, issues, and other contributions
+that are not aligned to this Code of Conduct, or to ban temporarily or
+permanently any contributor for other behaviors that they deem inappropriate,
+threatening, offensive, or harmful.
+
+## Scope
+
+This Code of Conduct applies both within project spaces and in public spaces
+when an individual is representing the project or its community. Examples of
+representing a project or community include using an official project e-mail
+address, posting via an official social media account, or acting as an appointed
+representative at an online or offline event. Representation of a project may be
+further defined and clarified by project maintainers.
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be
+reported by contacting the project team. All
+complaints will be reviewed and investigated and will result in a response that
+is deemed necessary and appropriate to the circumstances. The project team is
+obligated to maintain confidentiality with regard to the reporter of an incident.
+Further details of specific enforcement policies may be posted separately.
+
+Project maintainers who do not follow or enforce the Code of Conduct in good
+faith may face temporary or permanent repercussions as determined by other
+members of the project's leadership.
+
+## Attribution
+
+This Code of Conduct is adapted from the [Contributor Covenant][homepage], version 1.4,
+available at https://www.contributor-covenant.org/version/1/4/code-of-conduct.html
+
+[homepage]: https://www.contributor-covenant.org
+
+For answers to common questions about this code of conduct, see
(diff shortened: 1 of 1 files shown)


damani42:
Nice, it's a good idea to add a code of conduct.

neheb:
I'll just leave this here: https://www.phoronix.com/scan.php?page=news_item&px=FreeBSD-Code-of-Conduct-2020

faho:
>I'll just leave this here: https://www.phoronix.com/scan.php?page=news_item&px=FreeBSD-Code-of-Conduct-2020

@neheb I'll admit I have no idea why you're leaving it there.

neheb:
This CoC was disliked by many FreeBSD users. That's why.

faho:
>This CoC was disliked by many FreeBSD users

It's not the same?

https://web.archive.org/web/20180213113526/https://www.freebsd.org/internal/code-of-conduct.html is the old FreeBSD one. This one is quite substantially different.

ammgws:
Is there actually a need for this though?

ammgws:
@R2robot This is where I'm coming from: https://drewdevault.com/2020/01/17/Effective-project-governance.html

zanchey:
Having suggested this (in private, which is certainly less helpful) some years ago, I am strongly in support.

> Is there actually a need for this though?

As a professional in a tightly regulated occupation, and an active volunteer with a number of formal organisations, it really surprises when participants in a moderate-sized volunteer project don't agree on a (very basic) standard of behaviour. It seems like releasing an open source project without a license.

There have not been any public incidents requiring the enforcement of a code of conduct but after an incident is the worst time to realise you need to lay out some expectations.

In the wider community there have been a number of members of marginalised groups who have indicated the importance of a code of conduct.

In my view (and I'm happy to be told otherwise, publicly or privately) regular contributors and committers generally adhere to the standards set out in this patch, and I don't see that is creates an onerous obligation to participate. On the other hand, I hope that it will provide what mjg59 calls "a list of behaviours that are unlikely to occur within that community".

ammgws:
Fair enough. I guess I am just naive enough to trust participants in our established community to stay civil. Although the kind of people who would make hurtful comments are likely to knowingly ignore code of conduct anyway.

On that note, perhaps it should mention specifically _how_ one can contact the project team when they want to keep the reporting itself private. If there's no project email address, then could gitter be used instead? (Assuming it has private messaging?)

faho:
>Fair enough. I guess I am just naive enough to trust participants in our established community to stay civil.

You know, I used to think that as well. I was wrong, and I apologize for it.

Because there is very much disagreement on what "civil" (if that's even the right word) *means*. There are a lot of places on the internet where, for instance, transphobic behavior is considered acceptable.

And so, by adopting a Code Of Conduct that says so, we make it clear that this is not one of those places.

--------------

>On that note, perhaps it should mention specifically how one can contact the project team

The template includes "contact the project team at <email address>". I left it out because we don't currently *have* such an address.

We might want to create one.

>keep the reporting itself private

Reporting should obviously be confidential.

>could gitter be used instead? 

I'm not sure about that - gitter is another service that you need to agree to the ToS of, it requires another account,...

Email is ideal because everyone already has it.

ammgws:
> Reporting should obviously be confidential.
> Email is ideal because everyone already has it. We might want to create one.

Well I guess that's decided then :P

mqudsi:
I don't see anything that could be considered objectionable in the proposed CoC. As an old-school denizen of teh interwebs, I too like to continue to believe in the innate goodness of people but I think it's important to distinguish that technical and ethical competencies do not necessarily go hand-in-hand. While the internet of old was in many ways a meritocracy and the majority of BDFL-led organizations absolutely _have_ avoided doing anything that goes beyond the line that divides being politically incorrect from being a bigot/bully, it is important to acknowledge that there are absolutely are communities that have fallen victim to being unable to either see or act when such inappropriate behaviour did arise because it came from individuals who either were or had become too valuable to chastise. 

If such a code of conduct makes it clear to both members of a project and others interacting with the organization or using its product that they can expect at least some level of decency, I don't see why not. Basic decency aside, it is in the best interest of the community/organization itself and the spread of its product/ideology to be positively perceived in the community and to be a force of good, not evil.

As to this *specific* CoC being proposed in this PR, I think it's fine. I think having one is better than having a perfect one and obviously (as mentioned) organizations will ultimately do what they want but even if they either have an incomplete CoC that doesn't go into each and every do/don't or else chooses to be lenient and give second chances for repeat offenders, the fact that the CoC is there, is agreed to by all contributors/members, and hasn't been repealed in a revert commit says volumes. 

If this CoC were to go into the nitty gritty and delve into controversial specifics like refraining from using certain terminology because of _perceived_ etymological origins or requiring the usage of certain grammars or whatnot, I could see why people might be up in arms about it. But this is about as harmless and uncontroversial a CoC as one could find and I'm more than happy to see it added. 

(Now with my developer hat on, however, I do have nitpicks about the line wrapping inconsistencies and would certainly love to see a `gq` applied to the document before its merged!)

faho:
>(Now with my developer hat on however, I do have nitpicks about the line wrapping inconsistencies and would certainly love to see a gq applied to the document before its merged! Also: Oxford commas, please!)

This is from a template github provides. I would assume keeping it the same to ease diffing to be more important than minor formatting nitpicks.

zanchey:
On the mailing list, [Pim Dennedal asks](https://sourceforge.net/p/fish/mailman/message/37048288/):

"Why not adopt v2"?

faho:
>"Why not adopt v2"?

Because Github still provides 1.4 as the template.

I've now upgraded to 2.0. Note that, as the commit message states, this includes an "[INSERT CONTACT METHOD]" stub that we would have to fill in.

sorairolake:
Why is the file format Markdown instead of reST?

Related PR: #7057

faho:
Again: Because that is the format this is provided in, and we want to allow diffing with the original.

sorairolake:
@faho I understand. Thanks for answering.

ridiculousfish:
Thank you for taking the lead on this @faho and apologies for my absence on this matter as I took a bit of a GitHub break. I also am very strongly in support, I wish we had done it a while ago and regret not taking action when zanchey raised it privately.

Regarding the contact method stub, I am not sure how other OSS projects manage it. I think it would be OK to delete that sentence for now. I will set up MX records for `fishshell.com` so we can have shared addresses (and private ones if you want!)

faho:
Okay, merged as 0b3b4f3d9. That's it, then.

PR #7153, opened 2020-06-24
Speed up executable command completions

mqudsi:
This brings down the number of syscalls per potential completion result from three or four to just one.

It passes the tests and I'm comfortable with how it works on Linux and BSD, but putting this up here before merging in case anyone spots something non-portable or can think of any edge cases that aren't addressed.

### DIFF
--- a/src/wildcard.cpp
+++ b/src/wildcard.cpp
@@ -403,7 +403,7 @@ static const wchar_t *file_get_desc(const wcstring &filename, int lstat_res,
         return COMPLETE_SOCKET_DESC;
     } else if (S_ISDIR(buf.st_mode)) {
         return COMPLETE_DIRECTORY_DESC;
-    } else if (buf.st_mode & (S_IXUSR | S_IXGRP | S_IXGRP) && waccess(filename, X_OK) == 0) {
+    } else if (buf.st_mode & (S_IXUSR | S_IXGRP | S_IXOTH) && waccess(filename, X_OK) == 0) {
         // Weird group permissions and other such issues make it non-trivial to find out if we can
         // actually execute a file using the result from stat. It is much safer to use the access
         // function, since it tells us exactly what we want to know.
--- a/src/wildcard.cpp
+++ b/src/wildcard.cpp
@@ -357,6 +357,62 @@ bool wildcard_match(const wcstring &str, const wcstring &wc, bool leading_dots_f
     return match != fuzzy_match_none;
 }
 
+static int fast_waccess(const struct stat &stat_buf, uint8_t mode) {
+    // Cache the effective user id and group id of our own shell process. These can't change on us
+    // because we don't change them.
+    static const uid_t euid = geteuid();
+    static const gid_t egid = getegid();
+
+    // Cache a list of our group memberships.
+    static const std::vector<gid_t> groups = ([&]() {
+        std::vector<gid_t> groups;
+        while (true) {
+            int ngroups = getgroups(0, nullptr);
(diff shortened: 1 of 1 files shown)


src/wildcard.cpp
```diff
@@ -328,6 +328,67 @@ bool wildcard_match(const wcstring &str, const wcstring &wc, bool leading_dots_f
...
+    // because we don't change them.
+    static const uid_t euid = geteuid();
+    static const gid_t egid = getegid();
+
+    // Cache a list of our group memberships.
+    static const std::vector<gid_t> groups = ([&]() {
+        std::vector<gid_t> groups;
+        while (true) {
+            int ngroups = getgroups(0, nullptr);
+            assert(ngroups > 0 && "getgroups(0, nullptr) returned 0 group memberships!");
```
zanchey, line 342: I think technically you can have no supplementary groups? The spec is not very clear.

mqudsi: I forgot that the effective group/primary group is not guaranteed to be included and was presuming that would be the at least one response. You're right, we should probably remove that hard assert.

zanchey:
How much faster do you think this makes things?

mqudsi:
Completions are around 25% faster (on Linux 5.3 with pti enabled), taking 34% less system cpu time. I've been leery of syscall-heavy code because it just seems to get slower and slower these days.

```
hyperfine 'fish exec_completion_benchmark.fish' 'build/fish exec_completion_benchmark.fish'
Benchmark #1: fish exec_completion_benchmark.fish
  Time (mean ± σ):      2.142 s ±  0.076 s    [User: 1.421 s, System: 0.722 s]
  Range (min … max):    2.058 s …  2.262 s    10 runs
 
Benchmark #2: build/fish exec_completion_benchmark.fish
  Time (mean ± σ):      1.699 s ±  0.178 s    [User: 1.225 s, System: 0.475 s]
  Range (min … max):    1.554 s …  2.137 s    10 runs
```

src/wildcard.cpp
```diff
@@ -374,7 +434,7 @@ static const wchar_t *file_get_desc(const wcstring &filename, int lstat_res,
         return COMPLETE_SOCKET_DESC;
     } else if (S_ISDIR(buf.st_mode)) {
         return COMPLETE_DIRECTORY_DESC;
-    } else if (buf.st_mode & (S_IXUSR | S_IXGRP | S_IXGRP) && waccess(filename, X_OK) == 0) {
```
mqudsi, line 377: I'm pretty sure this is a bug and was supposed to be `S_IXUSR | S_IXGRP | S_IXOTH`

mqudsi:
On Linux and BSD it is not possible to effectively\* apply separate permissions to a symlink than the destination it points to. Does anyone know if there are any platforms that do not follow this rule (@krobelus @zanchey)?

e.g. 

```bash
> touch ./executable
> chmod 777 ./executable
> ln -s ./executable ./symlink
> chmod 077 ./symlink
> ls -alhtr ./executable ./symlink
----rwxrwx 1 mqudsi mqudsi  0 Jun 28 16:04 ./executable
lrwxrwxrwx 1 mqudsi mqudsi 12 Jun 28 16:04 ./symlink -> ./executable
```

<small>\* the symlink always has full permissions so the target is always the determinate</small>

zanchey:
At least on FreeBSD, the symbolic link can have a different mode than the target file - the `chmod -h` option will do so.

zanchey:
Oh, sorry - effectively - you are right

PR #7155, opened 2020-06-26
Fix binfmts.h typo in a comment

hugomg:
This PR fixes a small typo in a postfork.cpp comment.

The name of the header file being discussed should be binfmts.h, with an "s" at the end.


faho:
Yeah, that's right. I was confused by "binfmt_misc" and "binfmt.d", which both do not have the s.

PR #7159, opened 2020-06-28
__fish_print_hostnames: match IPv6 addresses + cleanup

exploide:
## Description

`__fish_print_hostnames` was only printing the first two bytes of an IPv6 address until the first colon, e.g. `2a03`. This PR changes the regex in order to match the whole IPv6 address.

- accept `:` in hostname, don't know why it was excluded. Maybe to exclude port numbers, but these come after the `]` in `known_hosts` so it should be fine
- filter out empty lines (`replace -f`)
- typo in Cygwin
- removed comment concerning `awk` since it is not used here

## TODOs:


mqudsi:
Looks good, thanks.

PR #7161, opened 2020-06-28
Change natural number to whole number

aadibajpai:
## Description

It's a bit silly really, but the tutorial talks about how status codes are returned as natural numbers. Well, 0 is not a natural number, since natural numbers start from 1 but whole numbers start from 0.

Thanks 😄

### DIFF
--- a/doc_src/tutorial.rst
+++ b/doc_src/tutorial.rst
@@ -397,7 +397,7 @@ To write them on the same line, use the semicolon (";"). That means the followin
 Exit Status
 -----------
 
-When a command exits, it returns a status code as a whole number.
+When a command exits, it returns a status code as a positive whole number.
 
 Unlike other shells, ``fish`` stores the exit status of the last command in ``$status`` instead of ``$?``.
 
--- a/doc_src/tutorial.rst
+++ b/doc_src/tutorial.rst
@@ -397,7 +397,7 @@ To write them on the same line, use the semicolon (";"). That means the followin
 Exit Status
 -----------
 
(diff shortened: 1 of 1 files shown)


faho:
>since natural numbers start from 1 but whole numbers start from 0.

That depends *entirely* on which definition you look up. I've seen both, and in particular I've seen "whole numbers" include negative numbers.

Change it to "positive whole number" and you have a deal.

aadibajpai:
> > since natural numbers start from 1 but whole numbers start from 0.
> 
> That depends _entirely_ on which definition you look up. I've seen natural numbers both with and without zero, and in particular I've never seen "whole numbers" not include negative numbers.
Interesting, in school we're taught that natural numbers are 1...∞ while whole numbers are 0...∞, no negative numbers anywhere (those are integers).
> Change it to "positive whole number" and you have a deal.
I'd be happy to do that (although 0 is not positive 😉)

krobelus:
One could say "non-negative integer" to include zero.

However, I think this sounds too academic. I'd go with `numeric exit code` and perhaps add `between 0 to 255`.

aadibajpai:
> One could say "non-negative integer" to include zero.
> 
> However, I think this sounds too academic. I'd go with `numeric exit code` and perhaps add `between 0 and 255`.

Personally I'd be okay with non-negative integer, describes it perfectly w/o the natural/whole ambiguity (I'm amazed that's a thing tbh).

faho:
Yeah, that'll do.

Merged, thanks!

PR #7163, opened 2020-06-29
Fix long description in completion.

damani42:
Simplify the description sudo for the option l. I think it's sufficient.

Fixes: #6981


mqudsi:
Yeah, that should be fine.

PR #7169, opened 2020-07-02
Add completions for mpc

gokulsoumya:
## Description

Add completions for `mpc`, used to control Music Player Daemon (MPD) from the command line.

### DIFF
--- a/share/completions/mpc.fish
+++ b/share/completions/mpc.fish
@@ -1,0 +1,93 @@
+# Completions for mpc, used to control MPD from the command line
+
+set -l subcommands consume crossfade queued mixrampdb mixrampdelay next \
+    pause play prev random repeat replaygain single seek seekthrough stop \
+    toggle add insert clear crop del mv searchplay shuffle load lsplaylists \
+    playlist rm save listall ls search search find findadd list stats mount \
+    mount unmount outputs disable enable toggleoutput channels sendmessage \
+    waitmessage subscribe idle idleloop version volume update rescan current
+
+# disable file completions
+complete -fc mpc
+
+complete -c mpc -s f -l format -d "Configure the format used to display songs"
+complete -c mpc -l wait -d "Wait for operation to finish (e. g.  database update)"
+complete -c mpc -l range -d "Operate on a range"
+complete -c mpc -s q -l quiet -l no-status -d "Prevents the current song status from being printed"
+complete -c mpc -l verbose -d "Verbose output"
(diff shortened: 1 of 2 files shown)


share/completions/mpc.fish
```diff
@@ -0,0 +1,91 @@
...
+complete -c mpc -n "not __fish_seen_subcommand_from $subcommands" -a idle -d "Wait until an event occurs"
+complete -c mpc -n "not __fish_seen_subcommand_from $subcommands" -a idleloop -d "Keep waiting and printing events as they occur"
+complete -c mpc -n "not __fish_seen_subcommand_from $subcommands" -a version -d "Report version of MPD"
+complete -c mpc -n "not __fish_seen_subcommand_from $subcommands" -a volume -d "Set volume"
+
+complete -c mpc -n "not __fish_seen_subcommand_from $subcommands" -a update -d "Scan for updated files in music directory"
+complete -c mpc -n "not __fish_seen_subcommand_from $subcommands" -a rescan -d "Like update, but also rescans unmodified files"
+complete -c mpc -n "not __fish_seen_subcommand_from $subcommands" -a current -d "Show currently playing song"
+
+complete -c mpc -n "__fish_seen_subcommand_from add insert" -a "(mpc listall)"
```
krobelus, line 86: Since there can be a lot of songs I wonder if it makes sense to use something based on `mpc ls $x` here where `$x` is the dirname of `commandline -ct`. What do you prefer?

gokulsoumya: Reading the man page again I noticed that `search` can be used to search for a *substring* in the filename, and now that looks more useful that either of `'ls` or 'listall`. Also a problem with `ls` I see is that the argument should be the complete dirname, capitalization and all. What do you think ?

krobelus: It depends on the use case: adding all songs from some directory/artist is easier with `ls`, whereas adding a specific song is easier with `listall`.

I think both are fine and we can merge this as is.

> I noticed that search can be used to search for a substring in the filename

Yeah, but fish already filters completions based on the current token so the difference between `listall` and `search` might not matter. `search` might be more efficient.

gokulsoumya: Yeah you're right; `search` vs `listall` doesn't matter much because of the token filtering. So yeah merging as is would be best I think.

> `search` might be more efficient

It could be on very large libraries, but I have only about 500 files, so can't say for sure. Anyways I'll add a comment about `search` being a possible faster alternative to `listall` for larger libraries.

PR #7170, opened 2020-07-02
docs/jobs: Header is stripped in command substitution

gokulsoumya:
## Description

If the `jobs` builtin is used inside a command substitution, the column headers are stripped and this wasn't documented in the man page:

```fish
> sleep 30 &
> jobs
1       20538   0%      running sleep 30 &
> echo (jobs)
1       20538   0%      running sleep 30 &
```


krobelus:
Merged as 6212a584a7089d103bdd68d938bee96375ce8caf, thank you.

PR #7180, opened 2020-07-07
Add files via upload

MichaelC2-718:
This recursive function will help find text in a file in a directory X or any subdirectory of X.

## Description

I added a function to help find text in a file. I made this because I wanted an easier way to search the Linux kernel source code.

## TODOs:

- [ ] Make a man page?
- [ ] Keep end directory the same as original directory (would help with friendliness, but seems a little tricky)

### DIFF
--- a/share/functions/search.fish
+++ b/share/functions/search.fish
@@ -1,0 +1,27 @@
+function search --description "Search for text in any file within a directory"
+    if test (count $argv) -ne 2
+        printf (_ "%s: requires exactly two arguments (search match dir)\n") search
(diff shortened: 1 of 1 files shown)


zanchey:
Could you explain how this is different to `grep --recursive` or `rg` (ripgrep)?

faho:
Yeah, no, this is just `grep -R`, and I have *numerous* issues with the code - not the least of which that you should not parse the output of `ls`.

Thanks for the idea, but I don't think this is necessary.

PR #7185, opened 2020-07-08
create_manpage_completions.py: cleanup

MaxVerevkin:
## Description

create_manpage_completions.py has weird 'skip' boolean assigned to `False` which is never changed to `True` and does not make much sense.


gokulsoumya:
This is the original code and it seems to me that now the remaining if construct and `output_path` can also be safely removed (change introduced in e7c1cb71859ce31ff83e114fa1e82d413b4229a5; parent is 911c37e63e81a60ece7a0689202b36e794780dae)

https://github.com/fish-shell/fish-shell/blob/911c37e63e81a60ece7a0689202b36e794780dae/share/tools/create_manpage_completions.py#L875-L891

PR #7189, opened 2020-07-10
Disown PIDs as well as PGIDs

### LINKED ISSUE #7183: run fish in docker image will spawn a zombie python process
fish version: 3.1.0
docker image: ubuntu 20.04
docker run command:  `docker run -it --rm --init <image id>`
image

However when I run fish on host linux, there is no such issue. Is this docker issue or fish issue? Or can you explain the python zombie process?

zanchey:
This series addresses the findings in #7183, where jobs started in the same PGID                                                
as fish (ie in non-job-control mode) are disowned but not waited on, creating                                                   
zombie processes.                                                                                                              
                                                                                                                                
All PIDs of a job's processes are added to the disown list and waited on (or                         
removed when they become invalid). However, this does also lead to both the PGID                                               
and PIDs for some jobs being added to the list, which might be unnecessary   
duplication. Comments are appreciated.                                  
                                                                                                                                
I think this probably needs tests, but I'm not sure what the best approach would
be.                                                                                  
                                                                                                            
Fixes #7183.

## TODOs:
<!-- Just check off what what we know been done so far. We can help you with this stuff. -->

### DIFF
--- a/src/proc.cpp
+++ b/src/proc.cpp
@@ -326,13 +326,23 @@ bool job_t::has_external_proc() const {
 /// we exit. Poll these from time-to-time to prevent zombie processes from happening (#5342).
 static owning_lock<std::vector<pid_t>> s_disowned_pids;
 
-void add_disowned_pgid(pid_t pgid) {
+void add_disowned_job(job_t *j) {
+    if (j == nullptr) return;
+
     // NEVER add our own (or an invalid) pgid as they are not unique to only
     // one job, and may result in a deadlock if we attempt the wait.
-    if (pgid != getpgrp() && pgid > 0) {
+    auto pgid = j->get_pgid();
+    if (pgid && *pgid != getpgrp() && *pgid > 0) {
         // waitpid(2) is signalled to wait on a process group rather than a
         // process id by using the negative of its value.
-        s_disowned_pids.acquire()->push_back(pgid * -1);
+        s_disowned_pids.acquire()->push_back(*pgid * -1);
+    } else {
+        // Instead, add the PIDs of any external processes
+        for (auto &process : j->processes) {
+            if (process->pid) {
+                s_disowned_pids.acquire()->push_back(process->pid);
+            }
+        }
     }
 }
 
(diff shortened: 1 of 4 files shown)


ridiculousfish:
Changes LGTM.  Regarding how to test it, I think it could be done in `checks/jobs.fish`:

1. start `sleep .25` in the background, capture its pid, and disown it
2. wait until it should exit
3. run something else, to trigger reaping
4. `kill` the pid. That should fail because the pid should no longer exist. If it succeeds it's a bug

mqudsi:
Nice job tracking this one down, @zanchey. 

It looks like the current behavior is "by design," introduced in 4a3ac6e91ec415bfe536f722a71903d9a67bb10d

    What we do is to not allow special pgids in the disowned_pids list.
    That means we might leave a zombie around (though we probably wait on
    0 somewhere), but that's preferable to infinitely looping.

Since this bug only exists iff the process is run under fish's own pgid and since each call to `wait` is a syscall that's going to block the main loop, I think the individual process pids should only be added if the job pgid equals the shell pgid.

src/proc.cpp
```diff
@@ -574,6 +580,11 @@ void remove_disowned_jobs(job_list_t &jobs) {
     while (iter != jobs.end()) {
         const auto &j = *iter;
         if (j->flags().disown_requested && j->job_chain_is_fully_constructed()) {
+            for (auto &process : j->processes) {
```
mqudsi, line 583: See my comment in the conversation, but briefly, can we wrap this in a `if (j->pgid() == shell_pgid)` precondition?

zanchey:
Haha, looks like @faho saw this coming!

> We could also use the PIDs here - that might actually work better, because we currently allow everything that fails this check to become a zombie.

I'll address the review comments and rebase.

src/proc.cpp
zanchey: I've decided to refactor this all into one spot.

zanchey:
My C++ is not the best, but I think I have the semantics of the pointer passing right here.

faho:
So while we're all here: Why *don't* we just use PIDs? How are the PGIDs any better?

mqudsi:
@faho Using pids means you need to issue a wait call for each process each time until the point that the process exits. Using a pgid reduces that to each job until the process exits (when kill with pgid and WNOHANG comes back with no results, you can just move on). If the waits are being done in a separate thread (no reason not to except the lack of thread safety in our code) it would not matter, but the waits block the progress of the main thread so it's reasonable to keep them down to a minimum.

zanchey:
I cannot, for the life of me, get the tests to pass on Travis (but they pass on my Linux machine and on GitHub Actions?)

zanchey:
OK, bumping the Travis environment to Ubuntu Bionic 18.04 has fixed the issues.

zanchey:
Tests are passing now - `kill` behaves differently on older versions of most Linux distributions, and `pgrep` doesn't show zombies on BSD/macOS, so I went with `ps`.

mqudsi:
Thanks for spearheading this, @zanchey. I've merged the changes (with an additional optimization).

PR #7191, opened 2020-07-11
Show one-line descriptions for functions and builtins using Alt-W

### LINKED ISSUE #2083: alt-w doesn't print the description of custom functions
Recently I discovered that [alt-w](http://fishshell.com/docs/current/index.html#editor) prints a short description of the command under the cursor. This function could be super handy, if it would print the [description of custom functions](http://fishshell.com/docs/current/commands.html#function) too, specified by the `-d DESCRIPTION` or `–description=DESCRIPTION` parameter.

gokulsoumya:
## Description

`__fish_whatis_current_token` (bound to `Alt-W` by default) currently prints information only for commands, using `whatis`. This PR aims to print information for builtins, functions (description specified with `-d`) and aliases in addition to the current implementation.

Even though man pages exist for builtins (and certain functions like `funced`), they aren't show by `whatis` because it isn't in $MANPATH (`man` works despite this by being overridden by a function; see `functions man`)

Closes #2083

## TODO

- [x] Print desciptions for:
    - [x] Builtins
    - [x] Functions (and aliases)

### DIFF
--- a/share/functions/__fish_whatis_current_token.fish
+++ b/share/functions/__fish_whatis_current_token.fish
@@ -3,39 +3,40 @@
 function __fish_whatis_current_token -d "Show man page entries or function description related to the token under the cursor"
     set -l token (commandline -pt)
 
-    if test -n "$token"
-        printf "\n"
-        set -l desc "$token: nothing appropriate."
+    test -n "$token"
+    or return
 
-        set -l tokentype (type --type $token 2>/dev/null)
+    printf "\n"
+    set -l desc "$token: nothing appropriate."
 
-        switch "$tokentype"
-            case "function"
-                set -l funcinfo (functions $token --details --verbose)
+    set -l tokentype (type --type $token 2>/dev/null)
 
-                test $funcinfo[5] != "n/a"
-                and set desc "$token - $funcinfo[5]"
+    switch "$tokentype"
+        case "function"
+            set -l funcinfo (functions $token --details --verbose)
 
-            case "builtin"
-                set desc (__fish_print_help $token | awk "/^$token /{print; exit}")
+            test $funcinfo[5] != "n/a"
+            and set desc "$token - $funcinfo[5]"
 
-            case "file"
-                set -l tmpdesc (whatis $token 2>/dev/null)
-                and set desc $tmpdesc
-        end
+        case "builtin"
+            set desc (__fish_print_help $token | awk "/^$token /{print; exit}")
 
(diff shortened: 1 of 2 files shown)


gokulsoumya:
Okay, `whatis` cannot be used to show descriptions for builtins even with setting `$MANPATH` because it relies on `mandb` databases, and the fish man pages are not in the db. Functions like `funced` with man pages are not really affected since they have a description along with the function definition.

share/functions/__fish_whatis_current_token.fish
```diff
@@ -1,11 +1,25 @@
...
-function __fish_whatis_current_token -d "Show man page entries related to the token under the cursor"
-    set -l tok (commandline -pt)
+function __fish_whatis_current_token -d "Show man page entries or function description related to the token under the cursor"
+    set -l token (commandline -pt)
 
-    if test -n "$tok[1]"
+    if test -n "$token"
         printf "\n"
-        whatis $tok[1]
+        whatis $token 2>/dev/null
```
krobelus, line 8: I think we can give precedence to functions and builtins over commands by switching on `type -p $tok[1]`, and only run `whatis` for files.

krobelus:
I guess for builtins it could be like `__fish_print_help builtin | awk '/^builtin /{print; exit}'`

share/functions/__fish_whatis_current_token.fish
gokulsoumya: Ah yeah; functions can override commands

gokulsoumya:
Functions take precedence over commands now, and descriptions for builtins are also printed.

share/functions/__fish_whatis_current_token.fish
```diff
@@ -1,11 +1,25 @@
...
+
+        if test $status -eq 16 # no match found
```
ridiculousfish: The value 16 here is a Linux-ism. Looks like macOS and BSD return success even if nothing was found :( I'm not sure what to do - maybe just leave it.

krobelus:
Merged as 0c72e650711f1cdb06fb80706d44187cd01ae732 with a small fix for Alt-W on `[`, thanks!

share/functions/__fish_whatis_current_token.fish
gokulsoumya: I changed the code so that commands have least precedence when searching for a description and it doesn't depend on`whatis`'s exit status.

PR #7196, opened 2020-07-14
Documentation: fix link target for "other features"

blinry:
Seems the link targets need a blank line above them, otherwise they will be rendered into the document.


faho:
Yeah, that's correct. Nice catch, thanks!

PR #7198, opened 2020-07-15
Minor documentation updates

charego:
Just getting my feet wet. There are 3 places where we answer the question "how to switch to fish", the best of which is in the FAQ. I don't want to link around too much here, the duplication seems reasonable in some cases.

### DIFF
--- a/doc_src/faq.rst
+++ b/doc_src/faq.rst
@@ -174,7 +174,7 @@ add a statement to your :ref:`user initialization file <initialization>` (usuall
 
 How do I customize my syntax highlighting colors?
 -------------------------------------------------
-Use the web configuration tool, :ref:`fish_config <cmd-fish_config>`, or alter the `fish_color family of environment variables <index#variables-color>`__.
+Use the web configuration tool, :ref:`fish_config <cmd-fish_config>`, or alter the :ref:`fish_color family of environment variables <variables-color>`.
 
 I accidentally entered a directory path and fish changed directory. What happened?
 ----------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


doc_src/faq.rst
```diff
@@ -174,15 +174,15 @@ add a statement to your :ref:`user initialization file <initialization>` (usuall
 
 How do I customize my syntax highlighting colors?
 -------------------------------------------------
-Use the web configuration tool, :ref:`fish_config <cmd-fish_config>`, or alter the `fish_color family of environment variables <index#variables-color>`__.
+Use the web configuration tool, :ref:`fish_config <cmd-fish_config>`, or alter the `fish_color family of environment variables <index.html#variables-color>`__.
```
faho, line 177: These links should lead to a *section* (via `:ref:\`something <variables-color>\``), not a .html file, so it can also work in other exports.

doc_src/index.rst
```diff
@@ -93,18 +93,16 @@ For uninstalling fish: see :ref:`FAQ: Uninstalling fish <faq-uninstalling>`.
...
-A scripts written in ``bash`` it would need a first line like this::
+A script written in ``bash`` would need a first line like this::
 
     #!/bin/bash
 
 
 This line tells the shell to execute the file with the bash interpreter, that is located at the path ``/bin/bash``.
 
-For a script, written in another language, just replace the interpreter ``/bin/bash`` with the language interpreter of that other language (for example ``/bin/python`` for a ``python`` script)
+For a script written in another language, just replace the interpreter ``/bin/bash`` with the language interpreter of that other language (for example: ``/bin/python`` for a python script, or ``/usr/local/bin/fish`` for a fish script).
```
faho, line 105: I've never seen python installed in /bin, but that's probably a different change.

doc_src/faq.rst
charego: Ah, much better. I actually forgot about other contexts for a second.

doc_src/index.rst
charego: Didn't want to change that. For what it's worth my system has `/bin/python3` and `/usr/bin/python3`.

charego:
> 
> 
> Looking good, except for the .html links.

Fixed the links and rebased.

faho:
Merged, thanks!

PR #7199, opened 2020-07-15
Human readable file sizes when ls -l is used instead of ll

quackduck:
## Description

Adding the `-h` option in the `ls` function means that `ls -l` outputs human-readable file sizes like `ll`

## TODOs:

I'm unsure as to exactly what I need to do:

### DIFF
--- a/share/functions/ls.fish
+++ b/share/functions/ls.fish
@@ -39,6 +39,6 @@ function ls --description "List contents of directory"
     __fish_set_lscolors
 
     isatty stdout
-    and set -a opt -F
+    and set -a opt -Fh
     command ls $__fish_ls_color_opt $argv
 end


quackduck:
What does `Error: Item dropped from history: 9038 223` mean?

faho:
Thank you, but we don't want to stray from ls' behavior in this way. The color is something that just disappears once you pipe (and in fact it should just be the default), this doesn't.

faho:
>What does Error: Item dropped from history: 9038 223 mean?

It's a test error that can sometimes happen when the testing system is under load, which is unfortunately too common on common  free CI systems.

It's not an issue with your code.

quackduck:
Okay! That makes sense. Doesn't `-F` change default `ls` behavior too?

Issue #7040, opened 2020-05-26
How to disable custom (i.e. per-command) autocompletion?


L-as:
A lot of commands have some custom autocompletion (like stack, systemctl, etc.) where they show the possible commands and such, but to be quite honest, it is very common for them to be somewhat broken, and really just in the way of what I'm trying to do.
I can e.g. not unzip a file where the name does not end in .zip, and often, e.g. for systemctl's autocompletion, an error will happen during it!

So I was wondering if there is any simple way to disable custom autocompletion while still keeping the autocompletion for selecting files (which is very useful).

krobelus:
There is no way to turn it off for all commands, but for specific commands you can do `complete -c unzip -e` or create an empty `~/.config/fish/completions/unzip.fish` to make it persistent.

I usually insert `: ` (alias for `true`) at the beginning of the command to get generic file completions, and then delete it before executing. Which is a bit awkward..

We should fix the incomplete completions, though!
Probably the unzip ones should complete any filename.
The systemctl completion error could be  #5689 which is fixed in 3.1.0.

mqudsi:
I've been thinking of an enhancement to fish where a second `<TAB>` after a first one fails to generate any completions will disable the `complete -x` behavior and revert to the default to address issues like this without sacrificing the default preference for convention-based filename completion.

exploide:
How would that go with cycling through completion items by repeatedly pressing `<TAB>`?

I fear that would be a confusing behaviour. Is it such a common problem? I would still vote for fixing unhelpful completions.

mqudsi:
No, it would only kick in when there are no completions due to the -x

faho:
How is this different from what we used to have (fall back to file completion), other than requiring a second tab press?

Quite often there just *are no sensible completions*, so adding files is absolutely unhelpful.

floam:
That would confuse me on second tab. +1 fix the individual annoying completions.

floam:
The zip completions would be better to complete any file name but sort `*.zip` first.

faho:
Nah, we should just stop the suffix filtering in most cases.

It's wrong so often it's not even funny.

floam:
I think suffix... prioritizing is a better way to go. They at least are more likely to be the desired files.

faho:
Anyway, that's a different point.

The question has been answered, the systemctl issue was fixed a while ago, the unzip issue is now fixed (with potential refinement).

I'm closing the issue.

PR #7208, opened 2020-07-17
completions/unzip: Restrict filename suggestions to .zip files only

gokulsoumya:
## Description

This PR changes `unzip` completions to suggest only filenames ending with `.zip`, like how `gunzip` and `bunzip2` completions currently do for their respective files.

### DIFF
--- a/share/completions/unzip.fish
+++ b/share/completions/unzip.fish
@@ -1,3 +1,5 @@
+complete -c unzip -x -a "(__fish_complete_suffix .zip)"
+
 complete -c unzip -s p -d "extract files to pipe, no messages"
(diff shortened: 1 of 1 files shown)


faho:
NAK.

We've been through this many times, restricting to file suffixes is more trouble than it is worth.

faho:
Notably, `unzip` *can* unzip files without an ".unzip" ending. So if you have a file with a different name, you still want to see it so you can unzip it.

krobelus:
See also #7040

ammgws:
> The zip completions would be better to complete any file name but sort *.zip first.

@floam's suggestion still seems worthwhile. IIRC the apt completions were done that way too (it completes all files but shows .deb first)

PR #7219, opened 2020-07-22
Fix build when ncurses is in nonstandard prefix

### LINKED ISSUE #6600: fish 3.1.0 - build - fatal error: curses.h: No such file or directory
fish: 3.1.0
cmake: 3.14.1
ncurses: 5.9

I have no local /usr installation of ncurses with development files.  It is in a non-standard location.

If I set PKG_CONFIG_PATH to have the ncurses pkgconfig dir, ncurses is not found.
I also tried setting PATH so ncurses5-config could be found.

So I set
```
  -DCMAKE_SYSTEM_PREFIX_PATH="${UTILS_NCURSES};${UTILS_GETTEXT};${UTILS_READLINE}"
```
And then ncurses is found in the path that I expect.

However, it does not seem that the include path gets added to the build, I encounter this error when building:
```
/build/tmp/fish-3.1.0/src/builtin_set_color.cpp:8:20: fatal error: curses.h: No such file or directory
```

So additionally I am setting CFLAGS/CXXFLAGS
```
 -DCMAKE_C_FLAGS="-O2 -I${UTILS_NCURSES}/include" \
 -DCMAKE_CXX_FLAGS="-O2 -I${UTILS_NCURSES}/include" \
```
(the files here in "include" are links to "include/ncursesw" files)

In the end the build looks something like:
```
export PATH=${UTILS_CMAKE}/bin:$PATH

export CC=${UTILS_GCC}/bin/gcc
export CXX=${UTILS_GCC}/bin/g++

export LDFLAGS+=" -static-libgcc -static-libstdc++"
export LDFLAGS+=" -Wl,-rpath,${UTILS_NCURSES}/lib"

mkdir build && cd build
cmake \
 -DCMAKE_BUILD_TYPE=Release \
 -DCMAKE_INSTALL_PREFIX=${PREFIX} \
 -DCMAKE_SYSTEM_PREFIX_PATH="${UTILS_NCURSES};${UTILS_GETTEXT};${UTILS_READLINE}" \
 -DCMAKE_C_FLAGS="-O2 -I${UTILS_NCURSES}/include" \
 -DCMAKE_CXX_FLAGS="-O2 -I${UTILS_NCURSES}/include" \
..

make -j${nproc}
make install
```

But ideally I would not have to set the CFLAGS/CXXFLAGS.

r-burns:
Fixes #6600

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -177,6 +177,8 @@ target_sources(fishlib PRIVATE ${FISH_HEADERS})
 target_link_libraries(fishlib
   ${CURSES_LIBRARY} ${CURSES_EXTRA_LIBRARY} Threads::Threads ${CMAKE_DL_LIBS}
   ${PCRE2_LIB} ${Intl_LIBRARIES} ${ATOMIC_LIBRARY})
+target_include_directories(fishlib PRIVATE
+  ${CURSES_INCLUDE_DIRS})
 
 # Define fish.
 add_executable(fish src/fish.cpp)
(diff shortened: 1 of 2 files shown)


ridiculousfish:
This LGTM, I will merge. Thank you!

PR #7220, opened 2020-07-22
Add completions for p7zip

sorairolake:
## Description

Add completions for [p7zip](http://p7zip.sourceforge.net/) command-line utils (`7z`, `7za` and `7zr`).

Fixes issue #

## TODOs:


MaxVerevkin:
I think you should add `7z` here.

https://github.com/fish-shell/fish-shell/blob/a30630a1dfb490ef6a6458595a977b0c586cac70/share/tools/create_manpage_completions.py#L763-L774

faho:
>I think you should add '7z', '7za' and '7zr' here.

@MaxVerevkin No need. Hand-written completions take priority over generated ones, so the files aren't used.

Adding them would add cruft to that script (and force you to jump through hoops if you wanted to use the generated version).

The reason to have exceptions in there is because the generated files don't make much sense and there's a huge number of them

MaxVerevkin:
@faho 
Would it make sense not to generate an automated completion if manual completion already exists?

faho:
No, because that wouldn't help much (save a couple of kilobytes of disk space I guess) and would in turn make looking at (e.g. because there are new options) or using the generated script (e.g. as a workaround) harder, plus it's more code.

MaxVerevkin:
@faho 
On my machine `fish_update_completions` time decreased from 9.26 to 8.16 secs.

MaxVerevkin:
> plus it's more code

Not really. https://github.com/fish-shell/fish-shell/pull/7221

faho:
Merged, thanks!

PR #7221, opened 2020-07-23
Do not generate an automated completion if manual completion already exists

MaxVerevkin:
## Description
This change may decrease `fish_update_completions` execution time. On my machine change was from 9.26 to 8.16 secs.


faho:
Thank you but no.

This command runs once a month or so, it's simply not worth the complication.

PR #7223, opened 2020-07-24
completions: Remove -f from pacman -U

rumpelsepp:
pacman -U is intended to be used with (among others) files like this:

        pacman -U ./linux-headers-5.6.2.arch1-2-x86_64.pkg.tar.zst

Thus, let's enable file path completion for this kind of operation.


zanchey:
Looks good, thanks!

Issue #6970, opened 2020-05-05
Starting in fish 3.1.0, abbreviations no longer trigger with following semicolon


lwolfsonkin:
I was running fish 3.0.x previously and just upgraded to fish 3.1.0. I use a slew of `abbr` abbreviations for all the `hg` subcommands, and I used to make a chain by typing e.g. the literal `amend; and upload` which would expand to `hg amend; and hg upload` and each of `amend` and `upload` respectively would expand once I typed a following space, enter, or semicolon (at least among probably others).

Unfortunately, in 3.1.0, it no longer triggers. Is this intended? If not, is there a way for me to get the old behavior as long as I'm on this version?

Thanks!

zanchey:
#5762 reduced the list of characters which expand abbreviations from "` ;|&^><`" to, by default, just space. Command execution always expands abbreviations under the cursor.

I think we probably want to put expansion back for all those other characters, though perhaps @faho can comment.

In the meantime, `bind ';' self-insert expand-abbr` should restore the behaviour you want - add it to your `config.fish`.

faho:
Yeah, it's probably fair to add the command terminators - `;`, `|` and `&`.

I'm not sold on the redirection operators.

zanchey:
My thinking is that if it would normally terminate a command then it should expand an abbreviation, so I'll add those back in.

PR #7229, opened 2020-07-28
vi-keybingings: Add missing abbr triggers for command terminators

gokulsoumya:
## Description

Abbr expansion on command terminators were added back in ec74c739 in response to #6970, but vi mode was not updated.


krobelus:
Good catch, thanks!

PR #7235, opened 2020-08-02
docs: use parsed-literal to highlight interactive examples

charego:
## Description

There are a few code blocks where the default highlighting does not
work and the documentation looks bad as a result. Usually this happens
when we are demonstrating an important interactive feature, such as
autosuggestions, syntax highlighting, or tab completion.

The pygments highlighter was not designed for code samples like these.
But it is important to show the behavior clearly in the docs. I am
attempting to make these weird examples look as much like the "normal"
code blocks as possible.

https://docutils.sourceforge.io/docs/ref/rst/directives.html#parsed-literal

## Screenshots

Before | After
------------ | -------------
syntax-highlighting-1 | syntax-highlighting-2
autosuggestions-1 | autosuggestions-2
tab-completions-1 | tab-completions-2
prompt-1 | prompt-2

### DIFF
--- a/doc_src/_static/custom.css
+++ b/doc_src/_static/custom.css
@@ -1,5 +1,22 @@
 .sphinxsidebar ul.current > li.current { font-weight: bold }
 
+.gray { color: #555555 }
+.purple { color: #551a8b }
+.red { color: #FF0000 }
+
+/* Color based on the Name.Function (.nf) class from pygments.css. */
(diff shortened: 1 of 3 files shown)


charego:
Simplified the original markup at the cost of a few duplicated CSS colors.

ridiculousfish:
Awesome, thank you!!

Issue #1285, opened 2014-02-05
Parser: fish fails to traverse pipes across newlines.


geoff-codes:
I believe in any other shell that uses the pipeline syntax fish inherits, it is allowed to break the statement at the pipe itself. That is, if `echo foo | cat - | cat` is valid syntax, so is:

``` sh
echo foo |
cat - |  # Tab space comments.
cat
```

> `foo`

But in fish, the newline must be escaped:

> ```
> Expected a command, but instead found end of the statement
> echo foo |
>           ^
> ```

Seems... buggish.

ridiculousfish:
This got fixed in #4712 🎉

JoshCheek:
Ty!! ❤️

PR #7238, opened 2020-08-04
Allow newlines after && and ||

krobelus:
We do the same for pipes (#1285). This matches POSIX sh behavior.

Are there any downsides to this? Making the grammar more flexible can make it more difficult to produce good error messages. If someone types `foo &&` by accident instead of `foo &` they will no longer get an error message. Seems like an acceptable price to pay.

We could also support newlines *before* these operators. That would be a trivial change, it's enough to [flip some bits](https://github.com/fish-shell/fish-shell/blob/master/src/ast.cpp#L529-L537) in the new AST.

## TODOs:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -27,7 +27,8 @@ Notable improvements and fixes
 Syntax changes and new commands
 -------------------------------
 - A new ``fish_is_root_user`` simplifies checking for superuser privilege, largely for use in prompts (#7031).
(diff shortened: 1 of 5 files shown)


ridiculousfish:
Changes LGTM, nice tests too! Maybe add a `fish_indent` test in `tests/checks/indent.fish`, it seems to work and preserve newlines, but that's the most fragile parts at the moment.

krobelus:
Nice `fish_indent` does the right thing here. Merged as `b947e36`

PR #7239, opened 2020-08-04
Implement XBPS completions

chilledfrogs:
## Description

XBPS has no intelligent `fish` completion, I attempted to fix that.

## (future) TODOs:

- [ ] Make case-correcting package matches
- [ ] Make mode flags which are mutually exclusive actually all work properly
- [ ] Make package versions show next to installed packages in completions, a bit like with `pacman`
- [ ] Command completion for the few utilities that need it


krobelus:
Looks fine, thanks!

PR #7241, opened 2020-08-05
Add completion for `prime-run`

oriko1010:
## Description

Simple completion for the `prime-run` script.

## TODOs:

### DIFF
--- a/share/completions/prime-run.fish
+++ b/share/completions/prime-run.fish
@@ -1,0 +1,6 @@
+function __fish_complete_prime_run_subcommand
+    set -lx -a PATH /usr/local/sbin /sbin /usr/sbin
+    __fish_complete_subcommand --commandline $args
+end
+
+complete -c prime-run -x -a "(__fish_complete_prime_run_subcommand)"
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -126,6 +126,7 @@ Completions
    -  ``mpc``
    -  ``nc``, ``netcat``, ``nc.openbsd``, ``nc.traditional``
    -  ``nmap``, ``ncat``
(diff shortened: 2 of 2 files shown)


share/completions/prime-run.fish
```diff
@@ -0,0 +1,6 @@
+function __fish_complete_prime_run_subcommand
+    set -lx -a PATH /usr/local/sbin /sbin /usr/sbin
+    __fish_complete_subcommand --commandline $args
```
krobelus, line 3: `$args` should be something like `(commandline -opc)[2..]`, or else it will only complete the command name - try
`prime-run grep --` and press Tab. (Though this doesn't seem terribly important here.)

oriko1010:
```
$ prime-run grep --[TAB]
--after-context                          (Print NUM lines of trailing context)
--basic-regexp                                        (Pattern is basic regex)
...
```
Unless you're referring to something else, this does seem to work for me. I used the `sudo` completion as a guide. Though now I realized the `$args` probably isn't even necessary, as `prime-run` has no arguments of its own.

krobelus: Oh I see, this happens to work because

- `$args` is undefined here, so it is an empty list (the parameters to a function are in `$argv`)
- In `__fish_complete_subcommand.fish`, `$subcommand` is set to the empty list.
- As a result the subcommand to complete is based off `(commandline -cop) (commandline -ct)` which is the current command line left of the cursor.

The `--commandline` option is useful for passing a custom commandline when completing commands like `sudo` that accept options on their own. So we don't need it here I guess.
Merged as 8b02a7 with the simplification in 46a69d6, thanks!

PR #7263, opened 2020-08-13
Implement the -P option for cd.

damani42:

### DIFF
--- a/doc_src/cmds/cd.rst
+++ b/doc_src/cmds/cd.rst
@@ -24,6 +24,18 @@ Fish also ships a wrapper function around the builtin ``cd`` that understands ``
 
 As a special case, ``cd .`` is equivalent to ``cd $PWD``, which is useful in cases where a mountpoint has been recycled or a directory has been removed and recreated.
 
+Options
+-------
+
+The following options are available:
+
+- ``-L`` or ``--logical`` Force symoblic links to be followed.
+- ``-P`` or ``--physical`` Use the physical directory structure without following symbolic.
+
+
+The default is to follow symbolic links, as if ``-L`` were specified.
+
+
 Examples
 --------
 
@@ -31,7 +43,7 @@ Examples
 
     cd
     # changes the working directory to your home directory.
-    
+
     cd /usr/src/fish-shell
     # changes the working directory to /usr/src/fish-shell
 
--- a/share/functions/cd.fish
+++ b/share/functions/cd.fish
@@ -3,10 +3,17 @@
 #
 function cd --description "Change directory"
     set -l MAX_DIR_HIST 25
+    set -l options h/help L/logical P/physical
+    argparse -n cd --exclusive L,P --max-args=1 $options -- $argv
+    or return
 
-    if test (count $argv) -gt (test "$argv[1]" = "--" && echo 2 || echo 1)
-        printf "%s\n" (_ "Too many args for cd command")
-        return 1
+    # The argparse removes the options from argv. When the option -L is uses
+    # it's not necessary to do something as fish always resolves the directory
+    # Like POSIX it's the default option for cd.
+
+    # Drop the argument and use the realpath
+    if set -q _flag_P
+        set argv (realpath $argv)
     end
 
     # Skip history in subshells.
--- a/share/completions/cd.fish
+++ b/share/completions/cd.fish
@@ -1,2 +1,4 @@
 complete -c cd -a "(__fish_complete_cd)"
 complete -c cd -s h -l help -d 'Display help and exit'
+complete -c cd -s L -l logical -d 'force symoblic links to be followed'
+complete -c cd -s P -l physical -d 'Use the physical directory structure'


share/functions/cd.fish
```diff
@@ -3,10 +3,13 @@
...
 function cd --description "Change directory"
     set -l MAX_DIR_HIST 25
+    set -l options h/help P
+    argparse -n cd --max-args=2 $options -- $argv
+    or return
 
-    if test (count $argv) -gt (test "$argv[1]" = "--" && echo 2 || echo 1)
-        printf "%s\n" (_ "Too many args for cd command")
-        return 1
+    # Drop the -P argument as fish always resolves the directory
```
faho, line 10: This hasn't been true in a while - since 3.0.

Fish's default is `cd -L`, like POSIX.

share/functions/cd.fish
```diff
@@ -3,10 +3,13 @@
 #
 function cd --description "Change directory"
     set -l MAX_DIR_HIST 25
+    set -l options h/help P
```
faho, line 6: We also want a long-version like our `pwd`s `--physical`.

So this should be `P/physical`.

zanchey:
I think that's partly my mistake, I advised fish always acts like `-P` but got it backwards!

zanchey:
Probably `cd -L` can do nothing, `cd -P` runs `cd (realpath $argv)`.

damani42:
Ok sorry. I will first implement cd -L in this case. And then I will implement -P option.

damani42:
I have updated my patch to implement the -L option. I will implement the -P option to another contribution. It's a bit more harder to implement it.

share/functions/cd.fish
damani42: Sorry for that, you're right. I have fixed that.

share/functions/cd.fish
damani42: Ok and what about -L for the long version?

share/functions/cd.fish
```diff
@@ -3,10 +3,20 @@
...
+    argparse -n cd --max-args=3 $options -- $argv
+    or return
 
-    if test (count $argv) -gt (test "$argv[1]" = "--" && echo 2 || echo 1)
-        printf "%s\n" (_ "Too many args for cd command")
-        return 1
+    # Drop the -L argument as fish always resolves the directory. Like POSIX
+    # it's the default option for cd.
+    if set -q _flag_L
+        set argv (string replace -- '-L' '' $argv)
```
zanchey, line 13: This block isn't needed - `argparse` removes the options from `$argv`. It's a good idea to keep the comment as an explanation of the default though.

share/functions/cd.fish
```diff
@@ -3,10 +3,20 @@
 #
 function cd --description "Change directory"
     set -l MAX_DIR_HIST 25
+    set -l options h/help L P
+    argparse -n cd --max-args=3 $options -- $argv
```
zanchey, line 7: Probably this needs to set the exclusive flag as well:

```suggestion
    argparse -n cd --exclusive L,P --max-args=3 $options -- $argv
```

share/functions/cd.fish
```diff
@@ -3,10 +3,20 @@
...
-        return 1
+    # Drop the -L argument as fish always resolves the directory. Like POSIX
+    # it's the default option for cd.
+    if set -q _flag_L
+        set argv (string replace -- '-L' '' $argv)
+    end
+
+    # Drop the argument and use the realpath
+    if set -q _flag_P
+        set argv (string replace -- '-P' '' $argv)
```
zanchey, line 18: This line can be dropped, see above.

damani42: If I have not this line I have en error:

    cd: Unknown option “-P”

    /usr/share/fish/functions/cd.fish (line 35):
        builtin cd $argv
        ^
    in function 'cd' with arguments '-P ansiconf/'

So I think we need this line.

damani42: No sorry I said something wrong.

share/functions/cd.fish
```diff
@@ -3,10 +3,17 @@
 #
 function cd --description "Change directory"
     set -l MAX_DIR_HIST 25
+    set -l options h/help L P
+    argparse -n cd --exclusive L,P --max-args=3 $options -- $argv
```
krobelus, line 7: The `--max-args` argument should be two because it only restricts the number of non-option arguments, including the command name itself.

```fish
$ argparse -n cd --max-args=2 --exclusive L,P h/help L/logical P/physical -- cd -L $PWD
$ argparse -n cd --max-args=2 --exclusive L,P h/help L/logical P/physical -- cd -L $PWD $PWD
cd: Expected at most 2 args, got 3
```

share/functions/cd.fish
```diff
@@ -3,10 +3,17 @@
 #
 function cd --description "Change directory"
     set -l MAX_DIR_HIST 25
+    set -l options h/help L P
```
krobelus, line 6: this should include the long options: `P/physical`, and `L/logical`

However, do we need the "logical" option at all? I feel like we can remove it because it's the default, or do people expect that option?

share/functions/cd.fish
faho: It should be *1*, because the command name isn't in $argv.

share/functions/cd.fish
damani42: I think it's good to have it. The other shell have this option.

share/functions/cd.fish
damani42: Sorry yes, I will fixed it.

doc_src/cmds/cd.rst
```diff
@@ -24,14 +24,26 @@ Fish also ships a wrapper function around the builtin ``cd`` that understands ``
 
 As a special case, ``cd .`` is equivalent to ``cd $PWD``, which is useful in cases where a mountpoint has been recycled or a directory has been removed and recreated.
 
+Options
+-------
+
+The following options are available:
+
+- ``-L`` or ``--logical`` Force symoblic links to be followed.
```
faho, line 32: Typo: `symoblic` should be `symbolic`

share/functions/cd.fish
```diff
@@ -3,10 +3,17 @@
...
-    if test (count $argv) -gt (test "$argv[1]" = "--" && echo 2 || echo 1)
-        printf "%s\n" (_ "Too many args for cd command")
-        return 1
+    # The argparse removes the options from argv. When the option -L is uses
+    # it's not necessary to do something as fish always resolves the directory
+    # Like POSIX it's the default option for cd.
+
+    # Drop the argument and use the realpath
+    if set -q _flag_P
+        set argv (realpath $argv)
```
faho, line 16: There are some subtle problems here:

1. We have a realpath *builtin*, but we special case it because some realpaths do more, so it's only used with `builtin realpath`. Using the builtin here is preferable for performance and consistency reasons.

2. `realpath` prints its own error if more than one component of the path doesn't exist (so "foo" always works, but "foo/bar" doesn't if foo doesn't exist). This would print both the realpath error and the cd error. You should probably silence realpath and only set argv if it succeeds. 

3. If you used `cd -P -- -foo` to go to a directory called `-foo`, this would try to interpret it as an option - it should be `realpath -- $argv` (or rather `builtin realpath -- $argv`)

faho (changes requested):

doc_src/cmds/cd.rst
krobelus: Shouldn't we simply omit this option, since it's the default? I always find redundant options confusing.

faho: It's there for backwards-compatibility if the default is ever switched.

I don't think we will, in this case (given that we did so in 3.0), but I don't think we ever regretted a default option.

faho:
Okay, I'm closing this for now since it's not ready to be merged. If you ever get back to it, we'll reopen!

PR #7265, opened 2020-08-18
Move fish_greeting to a function

faho:
This adds a "fish_greeting" function that prints the variable of the
same name.

In doing so, it makes $fish_greeting default to a global
variable (this is of little cost because of the `_` builtin)

This means that:

- We have fewer universal variables by default
- If we change the default greeting people will actually get
  it (unless they have a leftover universal, of course)
- If the user changes their language the variable changes with it

## TODOs:

### DIFF
--- a/share/functions/__fish_config_interactive.fish
+++ b/share/functions/__fish_config_interactive.fish
@@ -19,17 +19,6 @@ function __fish_config_interactive -d "Initializations that should be performed 
     set -g __fish_config_interactive_done
     set -g __fish_active_key_bindings
 
-    if not set -q fish_greeting
-        set -l line1 (_ 'Welcome to fish, the friendly interactive shell')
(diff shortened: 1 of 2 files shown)


zanchey:
The "it's a function! and a variable!" nature always gives me the heebie-jeebies, but this sounds sensible to me.

exploide:
When launching a fish for the first time, I usually only need to do two things: First, add the `sbin` directories to the PATH and second, disable the `fish_greeting` message. What is the preferred way for doing the latter, now that it is a global variable instead of a universal? Prior to that I just ran `set fish_greeting`.

faho:
>What is the preferred way for doing the latter, now that it is a global variable instead of a universal? Prior to that I just ran set fish_greeting.

It *defaults* to a global variable. You can still set it as a universal:

```fish
set -U fish_greeting
```

Or set it as a global in config.fish (this is executed after), or customize the function.

ridiculousfish:
This LGTM! Don't forget to update CHANGELOG.

PR #7266, opened 2020-08-19
Fix typo (verb clarification) in math.rst

jonbakke:
## Description
Was: "parameter expansion takes before expressions are evaluated."
Now: "parameter expansion happens before expressions are evaluated."

I suspect the original intent was to use "takes place," but I see "happens" as less idiomatic and therefore may benefit non-English-native users.

Fixes issue # N/A
Let me know if you'd like me to open an issue or use another tracker for record keeping.

## TODOs:

No code was changed, so no changes to add to docs or tests to conduct. I do not believe this merits a changelog entry, but I can propose a line for that if desired.

Thanks for all the fish.


ridiculousfish:
Thank you!

PR #7268, opened 2020-08-19
Fix the long description for ansible.

damani42:


zanchey:
Could drop the `.` here too

zanchey:
Thanks!

PR #7274, opened 2020-08-22
[Documentation] Correct Emacs bindings in vi insert mode

zeorin:
## Description

The original instruction does not actually enable Emacs key bindings in Vi mode. This fixes it.


ridiculousfish:
Thank you!

PR #7278, opened 2020-08-24
fix Subversion prompt

### LINKED ISSUE #6715: Fish SVN prompt regression in 3.1
In version 3.1, fish_svn_prompt does not work.

Expected:
```
⮞ fish -v
fish, version 3.0.2
⮞ __fish_svn_prompt
(26786|MX?||||X||)⏎
```

Observed:
```
⮞ fish -v
fish, version 3.1.0
⮞ fish_svn_prompt
(26786|||||||)⏎
```

I tried with different locations/repos and the result is always the same after the revision number.
It is also very slow for large subversion repositories (not the svn status retrieval but the processing in the loop for each column). I know you wouldn't expect large svn repositories in the wild in 2020, but boy do they exist.

chref:
- after switching to "string match", some SVN status symbols need
  proper escaping
- the __fish_svn_prompt_flag_names list was missing
  "versioned_obstructed" and was therefore not in line with
  the symbols from __fish_svn_prompt_chars
- when checking for individual SVN status symbols, use
  "string match -e" to handle the case where multiple different
  symbols appear in one status column
- use "sort -u" before merging all symbols from a column into
  one line

Fixes #6715

### DIFF
--- a/share/functions/fish_svn_prompt.fish
+++ b/share/functions/fish_svn_prompt.fish
@@ -70,7 +70,7 @@ set -g __fish_svn_prompt_char_token_broken_color --bold magenta
 function __fish_svn_prompt_parse_status --argument flag_status_string --description "helper function that does pretty formatting on svn status"
     # SVN status symbols
     # Do not change these! These are the expected characters that are output from `svn status`, they are taken from `svn help status`
-    set -l __fish_svn_prompt_chars A C D I M R X \\\? ! \~ L + S K O T B
+    set -l __fish_svn_prompt_chars A C D I M R X '\?' ! '~' L '\+' S K O T B
     # this sets up an array of all the types of status codes that could be returned.
     set -l __fish_svn_prompt_flag_names added conflicted deleted ignored modified replaced unversioned_external unversioned missing versioned_obstructed locked scheduled switched token_present token_other token_stolen token_broken
     # iterate over the different status types
@@ -78,7 +78,7 @@ function __fish_svn_prompt_parse_status --argument flag_status_string --descript
         # resolve the name of the variable for the character representing the current status type
         set -l flag_index (contains -i $flag_type $__fish_svn_prompt_flag_names)
         # check to see if the status string for this column contains the character representing the current status type
-        if test (count (string match -e $__fish_svn_prompt_chars[$flag_index] -- $flag_status_string)) -eq 1
+        if string match -q -r $__fish_svn_prompt_chars[$flag_index] -- $flag_status_string
             # if it does, then get the names of the variables for the display character and colour to format it with
             set -l flag_var_display __fish_svn_prompt_char_{$flag_type}_display
             set -l flag_var_color __fish_svn_prompt_char_{$flag_type}_color
--- a/share/functions/fish_svn_prompt.fish
+++ b/share/functions/fish_svn_prompt.fish
@@ -67,18 +67,18 @@ set -g __fish_svn_prompt_char_token_broken_color --bold magenta
 
 # ==============================
 
-function __fish_svn_prompt_parse_status --argument flag_status_string --description "helper function that does pretty formatting on svn status"
+function __fish_svn_prompt_parse_status --description "helper function that does pretty formatting on svn status"
     # SVN status symbols
     # Do not change these! These are the expected characters that are output from `svn status`, they are taken from `svn help status`
-    set -l __fish_svn_prompt_chars A C D I M R X '\?' ! '~' L '\+' S K O T B
+    set -l __fish_svn_prompt_chars A C D I M R X '?' ! '~' L + S K O T B
     # this sets up an array of all the types of status codes that could be returned.
     set -l __fish_svn_prompt_flag_names added conflicted deleted ignored modified replaced unversioned_external unversioned missing versioned_obstructed locked scheduled switched token_present token_other token_stolen token_broken
     # iterate over the different status types
     for flag_type in $__fish_svn_prompt_flag_names
         # resolve the name of the variable for the character representing the current status type
         set -l flag_index (contains -i $flag_type $__fish_svn_prompt_flag_names)
-        # check to see if the status string for this column contains the character representing the current status type
-        if string match -q -r $__fish_svn_prompt_chars[$flag_index] -- $flag_status_string
+        # check to see if the status list for this column contains the character representing the current status type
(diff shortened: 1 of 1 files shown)


share/functions/fish_svn_prompt.fish
```diff
@@ -70,15 +70,15 @@ set -g __fish_svn_prompt_char_token_broken_color --bold magenta
 function __fish_svn_prompt_parse_status --argument flag_status_string --description "helper function that does pretty formatting on svn status"
     # SVN status symbols
     # Do not change these! These are the expected characters that are output from `svn status`, they are taken from `svn help status`
-    set -l __fish_svn_prompt_chars A C D I M R X \? ! ~ L + S K O T B
+    set -l __fish_svn_prompt_chars A C D I M R X \\\? ! \~ L + S K O T B
```
faho, line 73: The `?` should be *quoted*, not escaped - currently the `?` can be disabled via the "qmark-noglob" feature flag (which will be turned on and eventually removed in future), so how many backslashes you need depends on the setting.

Quoting it should always work.

share/functions/fish_svn_prompt.fish
```diff
@@ -70,15 +70,15 @@ set -g __fish_svn_prompt_char_token_broken_color --bold magenta
...
     # this sets up an array of all the types of status codes that could be returned.
-    set -l __fish_svn_prompt_flag_names added conflicted deleted ignored modified replaced unversioned_external unversioned missing locked scheduled switched token_present token_other token_stolen token_broken
+    set -l __fish_svn_prompt_flag_names added conflicted deleted ignored modified replaced unversioned_external unversioned missing versioned_obstructed locked scheduled switched token_present token_other token_stolen token_broken
     # iterate over the different status types
     for flag_type in $__fish_svn_prompt_flag_names
         # resolve the name of the variable for the character representing the current status type
         set -l flag_index (contains -i $flag_type $__fish_svn_prompt_flag_names)
         # check to see if the status string for this column contains the character representing the current status type
-        if test (count (string match $__fish_svn_prompt_chars[$flag_index] -- $flag_status_string)) -eq 1
+        if test (count (string match -e $__fish_svn_prompt_chars[$flag_index] -- $flag_status_string)) -eq 1
```
faho, line 81: I don't see how this works?

$flag_status_string is received from

```fish
__fish_svn_prompt_parse_status $column_status
```

And $column_status is one element (because of `tr -d ' \n'`), so this would only ever succeed if this is the only status character.

That doesn't seem correct?

share/functions/fish_svn_prompt.fish
krobelus: I'm not sure if quoting even works - `string match` will change behavior depending on `status test-feature qmark-noglob`.

```fish
~> fish_features= fish -c "string match '?' x"
x
~> fish_features=qmark-noglob fish -c "string match '?' x"
~ [1]> 
```

`string match --regex` or `contains` should work

chref: Okay, I wasn't aware of the qmark-noglob feature. And no, just quoting it isn't enough.
So maybe this isn't the best way to go ...
Can `contains` be used on a string? I assumed it was meant for lists.

share/functions/fish_svn_prompt.fish
chref: Um, no. $column_status can be, for example, 'AM', if you have one file added and one file modified.
This will effectively result in `string match -e A -- AM` and `string match -e M -- AM`, which will both succeed. Without `-e`, they will not. In both cases the number of matches is 1 (although checking the count seems a bit unnecessary, but I wanted to keep my changes minimal).

share/functions/fish_svn_prompt.fish
krobelus: > Can contains be used on a string? I assumed it was meant for lists.

Yep, I believe that

`        if test (count (string match $__fish_svn_prompt_chars[$flag_index] -- $flag_status_string)) -eq 1
`

can be just

`
        if contains $__fish_svn_prompt_chars[$flag_index] -- $flag_status_string
`

assuming that `$flag_status_string` is a list, otherwise even `test x = y` could work.

chref: > can be just
> 
> `if contains $__fish_svn_prompt_chars[$flag_index] -- $flag_status_string`
> 
> assuming that `$flag_status_string` is a list,

It is not :-)

> otherwise even `test x = y` could work.

Won't work either, because, like I described below, we need to test if x is a substring of y (or more specifically, if the character x is contained in the string y).

share/functions/fish_svn_prompt.fish
faho: Ah, yeah, right, "--entire" *adds* the wildcards at both ends.

I always forget the semantics of string's glob mode (and tbh we should have just gone full-regex).

share/functions/fish_svn_prompt.fish
faho: `$flag_status_string` should probably just be a list. These are single characters, so I don't see why we join them.

share/functions/fish_svn_prompt.fish
```diff
@@ -116,11 +116,12 @@ function fish_svn_prompt --description "Prompt function for svn"
...
-        # 3. remove spaces and newline characters
-        set -l column_status (printf '%s\n' $svn_status_lines | cut -c $col | sort -u | tr -d ' \n')
+        # 3. remove duplicates
+        # 4. remove spaces
+        set -l column_status (printf '%s\n' $svn_status_lines | cut -c $col | sort -u | tr -d ' ')
 
-        # check that the character count is not zero (this would indicate that there are status flags in this column)
-        if [ (count $column_status) -ne 0 ]
+        # check that the column status list does not only contain an empty element (if it does, this column is empty)
+        if [ (count $column_status) -gt 1 -o -n $column_status[1] ]
```
faho, line 124: `test -n` *always* needs to have the variable quoted.

Our `test` (or `[`) bultin follows POSIX, and that allows usage like `test string`, which does the same thing as `test -n string`. Unfortunately it does that by saying that `test` with any one argument needs to *always* be true, so `test -n`, `test -d` etc are all true.

(yes, this is awkward, sorry)

chref: Odd -- because it works ;-)
But just to be sure, I'll add quotes.

faho:
Merged, thanks!

PR #7282, opened 2020-08-25
docs/string: Fix duplicated {} in match example

aureliojargas:
Follow-up fix from c5f06cd.

[ci skip]


faho:
Yeah that's right, thanks!

Issue #7283, opened 2020-08-25
Fix package manager


ryuichi1208:
The base image of the Dockerfile in the repository is centOS: latest.
If you specify the latest, CentOS 8 is selected.
On CentOS8 the package manager is yum, is this deprecated and shouldn't it be changed to dnf?

exploide:
Are you sure `yum` is deprecated on RHEL/CentOS? On Fedora, `yum` was replaced by `dnf` some time ago. On RHEL/CentOS, it is in fact the same `dnf` tool that is installed at least on more recent versions. But as far as I know, it is still named `yum` for some reasons instead of `dnf`. But I might be wrong. Maybe you have a reference (explicitly mentioning RHEL/CentOS, not Fedora where this is clear)?

zanchey:
Yeah, it's an alias, so should be fine to just leave.

```
$ stat `which yum`
  File: /usr/bin/yum -> dnf-3
```

PR #7284, opened 2020-08-25
Fix package manager in Dockerfile

ryuichi1208:
## Description

Fixed package manager from yum to dnf

Fixes issue #7283

## TODOs:


zanchey:
As discussed in #7283, they're equivalent on CentOS.

PR #7286, opened 2020-08-25
docs: Use \ instead of \\ in examples

aureliojargas:
Escaping characters (i.e., `\n` and `\t`) are [expanded when unquoted][1]
and besides a few exceptions, are left untouched when quoted. So unless
one really wants a literal "backslash followed by n" string, using `\\n`
is unnecessary and decreases the command's readability.

This documentation-only pull request fixes all the `\\` instances where
it should be `\` instead. 

Note that in most examples, the double `\\` is innocuous. In some, it
produces a different result. In those cases, the example output was also
adjusted accordingly (see the fixed `string escape` examples).

[1]: https://fishshell.com/docs/current/index.html#escaping-characters

[ci skip]

### DIFF
--- a/doc_src/cmds/echo.rst
+++ b/doc_src/cmds/echo.rst
@@ -64,7 +64,7 @@ Example
    > echo 'Hello World'
    Hello World
 
-   > echo -e 'Top\\nBottom'
+   > echo -e 'Top\nBottom'
    Top
    Bottom
 
--- a/doc_src/cmds/printf.rst
+++ b/doc_src/cmds/printf.rst
@@ -77,7 +77,7 @@ Example
 
 ::
 
-    printf '%s\\t%s\n' flounder fish
+    printf '%s\t%s\n' flounder fish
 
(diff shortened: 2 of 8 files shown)


aureliojargas:
Dear fish maintainers, I've broken this into separate commits to be easier to review each change in isolation, and also revert/discard if necessary. Please tell me if you prefer it differently.

faho:
Oh, that's a good catch, merging!

>Dear fish maintainers, I've broken this into separate commits to be easier to review each change in isolation

Thank you for your effort, but, honestly, this level of granularity is overdoing it. Since these commits are all doing basically the same thing, they can just be one commit.

>also revert/discard if necessary.

Git can squash commits together. If you make a change in a commit you don't like, you can

- Make a commit that changes it back
- Run `git rebase --interactive` and choose "squash" on that commit

This will create a new commit that combines the two commits, and since a change forward and then back is no change it will have that change removed.

So there really is no need to have each change as a separate commit, and changes of this style make sense together.

aureliojargas:
@faho Ok, thanks for the feedback and for the merge :+1:

PR #7295, opened 2020-08-29
Display symbols for control characters in the commandline buffer

### LINKED ISSUE #5274: fish reader loses it when there's a blank line in an indented block
Paste the following at the prompt:
```fish
	clang --autocomplete='--std=' |

		string replace -r -- '^([^ ]+)\s*(.*)' '$1\t\$2'
```

Observe that the text on the first row does not match what was pasted. Try navigating back and forth/up and down with the arrow keys on the different lines. Press ^C and observe that the prompt isn't cleared.

fish loses it.

https://asciinema.org/a/TsAwarGwkZTpVQL3aQ72Dkvkr

### LINKED ISSUE #6923: fish doesn't cope with literal tabs in command history
- fish 3.1.0
- Ubuntu 18.04, Linux 4.15.0-96-generic
- first encountered in Alacritty, reproduced in xterm

If there is a literal tab in command history (aiui this is impossible with fish itself, but I guess it's reading my ~/.bash_history and seeing this there), fish's autocomplete breaks horribly when it prints it.

Demonstration: https://asciinema.org/a/pyYipLwhQAZEvLJkZkDysXAQv

(the command it's completing there is `echo '\t' | xxd`, though obviously with the `\t` replaced by a literal tab character)

krobelus:
Represent tabs on the commandline as "␉", and so on for other ASCII control
characters, see https://unicode-table.com/en/blocks/control-pictures/.
This makes the fish reader use a reasonable width for these characters.

You can see how this works by inserting some control characters on the command line:

	set chars
	for x in (seq 1 0x1F)
		set -a chars (printf "%02x\\\\x%02x" $x $x)
	end
	eval set chars $chars
	commandline -i "echo '" $chars

Fixes #6923
Fixes #5274

We could extend this approach to display a fallback symbol for every unknown
control/nonprintable character.

An alternative way to handle tabs would be to just replace them with spaces
when inserted.  However, this seems slightly better because it still allows
to use tabs and other control characters inside quotes.

## TODOs:

### DIFF
--- a/src/fallback.cpp
+++ b/src/fallback.cpp
@@ -239,6 +239,8 @@ int fish_get_emoji_width(wchar_t c) {
 #include "widecharwidth/widechar_width.h"
 
 int fish_wcwidth(wchar_t wc) {
+    wc = character_to_render(wc);
+
     // The system version of wcwidth should accurately reflect the ability to represent characters
     // in the console session, but knows nothing about the capabilities of other terminal emulators
     // or ttys. Use it from the start only if we are logged in to the physical console.
@@ -260,6 +262,7 @@ int fish_wcwidth(wchar_t wc) {
     // or standalone with a 1 width. Since that's literally not expressible with wcwidth(),
     // we take the position that the typical way for them to show up is composed.
     if (wc >= L'\u1160' && wc <= L'\u11FF') return 0;
+
     int width = widechar_wcwidth(wc);
 
     switch (width) {
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -137,6 +137,7 @@ Improved terminal output
 -  Autosuggestions now show up also when the cursor passes the right
    prompt (#6948).
 -  The cursor shape in Vi mode changes properly in  Windows Terminal (#6999).
+-  Non-printable ASCII control characters in the commandline are display using graphic Unicode symbols.
 -  The spurious warning about terminal size in small terminals has been removed (#6980).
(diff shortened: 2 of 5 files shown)


src/fallback.cpp
```diff
@@ -239,6 +239,8 @@ int fish_get_emoji_width(wchar_t c) {
 #include "widecharwidth/widechar_width.h"
 
 int fish_wcwidth(wchar_t wc) {
+    wc = character_to_render(wc);
```
faho, line 242: Why is this here?

Wouldn't it check a different character from what would then be inserted (if it doesn't hit the "return c" case)?

krobelus: You're right, this breaks things :cold_sweat: 

Without this line, fish would assume width `0` for the tab character, which leads to drawing a command line like

image

We have some places where we compute the width of a tab, for example if the prompt contains a tab.
In that case, `fish_wcwidth` should not be called for the tab character.

However, for other control characters we don't have such logic so this computes the wrong width, as can be seen by typing a multi-line commandline with this prompt:

```fish
function fish_prompt
    printf (string repeat -n 50 \b)'> '
end
```

We might want to let the caller decide whether the character is something to be rendered on the command line, or rendered directly. Not sure if that will cause any trouble.

The most important thing to fix here is \t, so only pretty-printing that is an option.

krobelus:
Okay since I don't have a working solution now, I'll reopen when I get to it.

PR #7301, opened 2020-09-01
Fix examples in tokenizer comment for redirection

lanza:
Four of these examples were incorrect and didn't perform the stated
behavior in neither bash nor fish. Fix them here.

### DIFF
--- a/src/tokenizer.cpp
+++ b/src/tokenizer.cpp
@@ -316,10 +316,10 @@ maybe_t<pipe_or_redir_t> pipe_or_redir_t::from_string(const wchar_t *buff) {
         cmd >? file      noclobber redirection
         cmd >>? file     appending noclobber redirection
         cmd 2> file      file redirection with explicit fd
-        cmd >&2 file     fd redirection with no explicit src fd (stdout is used)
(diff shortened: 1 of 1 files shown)


ridiculousfish:
I will always love comment improvements 🥇  thank you for spotting this and taking the time to write it up!

Looking at these examples, I think the problem is that some reference a file when they should not. That is:

    echo hello >&2

should send stdout to stderr and no file should be involved. So I think the comment fix should just delete the reference to 'file' for fd redirections:

    cmd >&2       fd redirection with no explicit src fd (stdout is used)

does that seem right to you?

lanza:
> Looking at these examples, I think the problem is that some reference a file when they should not. That is:

Well, they don't technically require a file to be used but adding the file demonstration does exemplify how to use them. I think it makes more sense to leave them there. There's the counter argument that this is documentation for the parser and not for the user, but extra exemplary information doesn't really negate from the parser writer's understanding. Hence I left the file usages intact.

ridiculousfish:
I think the 'file's are just a copy and paste error! I'll merge this and delete them. Thank you!

PR #7303, opened 2020-09-02
docs: Consistent background color for code samples

charego:
## Description

Previously I added some markup to improve the appearance of interactive examples. The background color needs a touch up. Below is the _before_ image. The _after_ image is the same, except both blocks have the gray background.

Also included: small change to `__fish_config_interactive` to stop initializing a color variable that is unused in 3.2.

highlight


faho:
Merged, thanks!

PR #7308, opened 2020-09-03
Colourise diff(1) output, if supported

deviant:
## Description

This adds a simple wrapper to enable colour support for diff(1), like those for ls(1) and grep(1).

## TODOs:

### DIFF
--- a/share/functions/diff.fish
+++ b/share/functions/diff.fish
@@ -1,5 +1,5 @@
 # Use colours in diff output, if supported
-if command diff --color=auto /dev/null /dev/null 2>&1
+if command -vq diff; and command diff --color=auto /dev/null{,} >/dev/null 2>&1
     function diff
         command diff --color=auto $argv
     end
--- a/share/functions/diff.fish
+++ b/share/functions/diff.fish
@@ -1,0 +1,6 @@
+# Use colours in diff output, if supported
+if command diff --color=auto /dev/null /dev/null 2>&1
+    function diff
+        command diff --color=auto $argv
+    end
+end
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -103,6 +103,7 @@ Interactive improvements
    revealed.
 -  The output of ``time`` is now properly aligned in all cases (#6726).
 -  The ``pwd`` command supports the long options ``--logical`` and ``--physical``, matching other implementations (#6787).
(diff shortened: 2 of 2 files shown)


share/functions/diff.fish
```diff
@@ -0,0 +1,6 @@
+# Use colours in diff output, if supported
+if command diff --color /dev/null /dev/null 2>&1
+    function diff
+        command diff --color $argv
```
krobelus, line 4: Great idea! Does it make sense to use `--color=auto` to be more explicit?

deviant: I considered it, but it really feels like auto is just there so there's something to accompany --color=never and --color=always. I can make that change if you'd rather, though.

krobelus: The question is, is there a `diff(1)` implementation that accepts `--color` and interprets it as `--color=always` in the GNU variant?
The rationale being that `diff a b > some-file` should probably not add colors.
Which could happen if someone made `diff` a wrapper script to `git diff --no-index` though I don't really know what they ought to expect then.

For `grep` we use `--color=auto` and for `ls` we prefer it if available. So I'd feel safer with that, although it probably doesn't matter.

deviant: I just checked through the BSDs, and none of them have a --color option (save for DragonFly, which uses the GNU diffutils). I'll add it, nonetheless, lest this derail into bikeshedding.

deviant: (although that could be a good argument for adding it, since it futureproofs against other implementations adding differing behaviour)

share/functions/diff.fish
```diff
@@ -0,0 +1,6 @@
+# Use colours in diff output, if supported
+if command diff --color=auto /dev/null /dev/null 2>&1
```
krobelus, line 2:
```suggestion
if command -vq diff; and command diff --color=auto /dev/null{,} >/dev/null 2>&1
```

Some more paranoia, sorry :smile_cat: - we explicitly mention `diff` as requirement in CONTRIBUTING so it's probably not installed on every system.

(I can merge and squash that, no need to push again).

krobelus (approved):

share/functions/diff.fish
deviant: Sure, fine by me. :+1:

PR #7309, opened 2020-09-04
Call "fish_command_not_found" if a command wasn't found

### LINKED ISSUE #7293: Replace command-not-found event with a known function
Fish has long had the `fish_command_not_found` event that is triggered when a command isn't known.

It's always been a bit weird to me that this used the event machinery because

- it makes it awkward to *replace* the handler (because it has already been registered by the time you try)
- there is no real use for multiple handlers for this (unlike eg. pre-exec)

So really, this should just be a function that is invoked by name, `fish_command_not_found`, and if that isn't defined the default "fish: command 'choochoo' not found" one should take over.

This would also allow us to remove some rather weird logic from __fish_config_interactive and just put it in a normal function file.

faho:
Previously, when a command wasn't found, fish would emit the
"fish_command_not_found" *event*.

This was annoying as it was hard to override (the code ended up
checking for a function called `__fish_command_not_found_handler`
anyway!), the setup was ugly,
and it's useless - there is no use case for multiple command-not-found handlers.

Instead, let's just call a function `fish_command_not_found` if it
exists, or print the default message otherwise.

The event is completely removed, but because a missing event is not an error
(MEISNAE in C++-speak) this isn't an issue.

Note that, for backwards-compatibility, we still keep the default
handler function around even tho the new one is hard-coded in C++.

Also, if we detect a previous handler, the new handler just calls it.

This way, the backwards-compatible way to install a custom handler is:

```fish
function __fish_command_not_found_handler --on-event fish_command_not_found
    # do a little dance, make a little love, get down tonight
end
```

and the new hotness is

```fish
function fish_command_not_found
    # do the thing
end
```

Fixes #7293.

## TODOs:

### DIFF
--- a/doc_src/cmds/fish_command_not_found.rst
+++ b/doc_src/cmds/fish_command_not_found.rst
@@ -16,12 +16,11 @@ Synopsis
 Description
 -----------
 
-When fish can't find a command, it invokes this function.
+When fish tries to execute a command and can't find it, it invokes this function.
 
-It can print a message, usually tailored to the OS, to e.g. check which package would contain the missing command,
-by calling out to a handler provided by the operating system.
+It can print a message to tell you about it, and it often also checks for a missing package that would include the command.
 
-Fish ships multiple handlers for various Linux distributions and chooses from them when this function is loaded,
+Fish ships multiple handlers for various operating systems and chooses from them when this function is loaded,
 or you can define your own.
 
 It receives the full commandline as one argument per token, so $argv[1] contains the missing command.
--- a/doc_src/cmds/fish_command_not_found.rst
+++ b/doc_src/cmds/fish_command_not_found.rst
@@ -1,0 +1,77 @@
+.. _cmd-fish_cmd_not_found:
+
+fish_command_not_found - what to do when a command wasn't found
+===============================================================
+
+Synopsis
+--------
+
+::
+
+  function fish_command_not_found
+      ...
+  end
+
+
+Description
+-----------
+
+When fish can't find a command, it invokes this function.
+
+It can print a message, usually tailored to the OS, to e.g. check which package would contain the missing command,
+by calling out to a handler provided by the operating system.
+
+Fish ships multiple handlers for various Linux distributions and chooses from them when this function is loaded,
+or you can define your own.
+
+It receives the full commandline as one argument per token, so $argv[1] contains the missing command.
+
+When you leave ``fish_command_not_found`` undefined (e.g. by adding an empty function file) or explicitly call ``__fish_default_command_not_found_handler``, fish will just print a simple error.
+
+Example
+-------
+
+A simple handler:
+
+::
+
+    function fish_command_not_found
+        echo Did not find command $argv[1]
+    end
+
+    > flounder
+    echo Did not find command flounder
+
+Or the handler for OpenSUSE's command-not-found::
+
+    function fish_command_not_found
+        /usr/bin/command-not-found $argv[1]
+    end
+
+Or the simple default handler::
+
+    function fish_command_not_found
+        __fish_default_command_not_found_handler $argv
+    end
+
+Backwards compatibility
+-----------------------
+
+This command was introduced in fish 3.2.0. Previous versions of fish used the "fish_command_not_found" :ref:`event <event>` instead.
+
+To define a handler that works in older versions of fish as well, define it the old way::
+
+  function __fish_command_not_found_handler --on-event fish_command_not_found
+       echo COMMAND WAS NOT FOUND MY FRIEND $argv[1]
+  end
+
+in which case fish will define a ``fish_command_not_found`` that calls it,
+or define a wrapper::
+
+  function fish_command_not_found
+       echo "G'day mate, could not find your command: $argv"
+  end
+
+  function __fish_command_not_found_handler --on-event fish_command_not_found
+       fish_command_not_found $argv
+  end
--- a/doc_src/cmds/fish_command_not_found.rst
+++ b/doc_src/cmds/fish_command_not_found.rst
@@ -40,7 +40,7 @@ A simple handler:
     end
 
     > flounder
-    echo Did not find command flounder
+    Did not find command flounder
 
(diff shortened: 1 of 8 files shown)


doc_src/cmds/fish_command_not_found.rst
```diff
@@ -0,0 +1,77 @@
...
+
+  function fish_command_not_found
+      ...
+  end
+
+
+Description
+-----------
+
+When fish can't find a command, it invokes this function.
```
krobelus, line 19: Maybe `When fish can't find the command you're trying to execute, it invokes this function.`

doc_src/cmds/fish_command_not_found.rst
```diff
@@ -0,0 +1,77 @@
...
+
+It can print a message, usually tailored to the OS, to e.g. check which package would contain the missing command,
+by calling out to a handler provided by the operating system.
+
+Fish ships multiple handlers for various Linux distributions and chooses from them when this function is loaded,
+or you can define your own.
+
+It receives the full commandline as one argument per token, so $argv[1] contains the missing command.
```
krobelus, line 27: A slightly different, less honest approach for these three paragraphs. Not sure if this is any better:

By default, it prints a message to inform you which package would contain the missing command.
This is done by delegating to an operating-system specific handler. Fish knows about the right handler for various Linux distributions and chooses the appropriate one when this function is loaded.

If you don't want that behavior, you can define `fish_command_not_found` yourself.
It receives the full commandline as one argument per token, so $argv[1] is the missing command.

doc_src/cmds/fish_command_not_found.rst
```diff
@@ -0,0 +1,77 @@
...
+A simple handler:
+
+::
+
+    function fish_command_not_found
+        echo Did not find command $argv[1]
+    end
+
+    > flounder
+    echo Did not find command flounder
```
krobelus, line 43: Extra `echo`

src/parse_execution.cpp
```diff
@@ -749,10 +749,40 @@ end_execution_reason_t parse_execution_context_t::handle_command_not_found(
             event_args.insert(event_args.begin(), cmd_str);
         }
 
-        event_fire_generic(*parser, L"fish_command_not_found", &event_args);
+        wcstring buffer;
+        wcstring error;
+
+        // Redirect to stderr
+        auto io = io_chain_t{};
+        io.append_from_specs({redirection_spec_t{STDOUT_FILENO, redirection_mode_t::fd, L"2"}}, L"/");
```
krobelus, line 757: `L"/"` could even be the empty string since it's never used, right? A shame we can't have NULL here :older_man:

src/parse_execution.cpp
```diff
@@ -749,10 +749,40 @@ end_execution_reason_t parse_execution_context_t::handle_command_not_found(
...
+        if (function_exists(L"fish_command_not_found", *parser)) {
+            buffer = L"fish_command_not_found";
+            for (const wcstring &arg : event_args) {
+                buffer.push_back(L' ');
+                buffer.append(escape_string(arg, ESCAPE_ALL));
+            }
+        } else {
+            // If we have no handler, just print it as a normal error.
+            error = _(L"Unknown command:");
+            if (event_args.size()) {
```
krobelus, line 768: I think there's some clang-tidy check that would suggest `!event_args.empty()` here.

tests/checks/command-not-found.fish
```diff
@@ -0,0 +1,18 @@
+# RUN: %fish -C 'set -g fish %fish' %s
+$fish -c "nonexistent-command-1234 banana rama"
```
krobelus, line 2: I think `set -g PATH` would work as well to ensure that this doesn't exist.

share/functions/fish_command_not_found.fish
```diff
@@ -0,0 +1,61 @@
...
+        # The fish_command_not_found event was removed in fish 3.2.0,
+        # and future versions of fish will just call a function called "fish_command_not_found".
+        # You have defined a custom handler, we suggest renaming it to "fish_command_not_found".
+        __fish_command_not_found_handler $argv
+    end
+    # First check if we are on OpenSUSE since SUSE's handler has no options
+    # but the same name and path as Ubuntu's.
+else if contains -- suse $os || contains -- sles $os && type -q command-not-found
+    function fish_command_not_found
+        /usr/bin/command-not-found $argv[1]
```
krobelus, line 26: This path and the one from the `type -q`  above don't align. Using the basename should be fine, right? It's not in `/sbin` after all...

share/functions/__fish_config_interactive.fish
```diff
@@ -261,63 +261,9 @@ function __fish_config_interactive -d "Initializations that should be performed
...
-                end
-            end
-            # Use standard fish command not found handler otherwise
-        else
-            function __fish_command_not_found_handler --on-event fish_command_not_found
-                __fish_default_command_not_found_handler $argv[1]
-            end
-        end
+    # For backwards-compatibility - the event doesn't exist anymore so it's harmless.
+    function __fish_command_not_found_handler --on-event fish_command_not_found
```
krobelus, line 265: So peopla are calling  `__fish_command_not_found_handler` in their config?

faho: Not really, no.

This bit is really just about downgrading - prior to this PR fish had *no* handler without the function, so if one of those ran this __fish_config_interactive it would behave quite weirdly (it would just print the backtrace, without an actual error message).

It's not too important, tbh, but it's simple enough to add.

share/functions/fish_command_not_found.fish
faho: Yeah, it's in $PATH, so we should use it via $PATH.

src/parse_execution.cpp
faho: Yeah, there is. I kinda hate clang-tidy tbh.

It *can* give sensible hints, but much of it is just busy work like that.

faho:
Merged as 340de7317..58d549e05 (with CHANGELOG).

PR #7311, opened 2020-09-05
docs/isatty: Mention default value for FILE DESCRIPTOR

aureliojargas:
As seen in `share/functions/isatty.fish` (note the empty string):

    switch "$argv"
        case stdin ''
            set fd 0


faho:
That's correct, thanks, merged!

PR #7316, opened 2020-09-07
ffprobe completions: show_stream -> show_streams

jboekesteijn:
This should be `show_streams` instead of `show_stream` according to:

https://git.ffmpeg.org/gitweb/ffmpeg.git/blob/31b6b6685e705f487c16546a3e2e687af99e470b:/doc/ffprobe.texi#l218


faho:
Thanks, merged!

PR #7355, opened 2020-09-27
Fix string match example

aureliojargas:
The `?` requires a char, so `foo` cannot match.


krobelus:
Correct, well spotted!

With [`qmark-noglob`](https://fishshell.com/docs/current/index.html#future-feature-flags) enabled none of them will match but that's a problem for another day.

aureliojargas:
@krobelus thanks for the heads-up, I was not aware that `?` was deprecated. 

BTW, it was https://github.com/aureliojargas/doctest.fish who spotted this small documentation bug. My human eyes are not that accurate :)

PR #7356, opened 2020-09-28
run codebase through clang-tidy

neheb:
Added NOLINT for some problematic sections.

### DIFF
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -10,8 +10,6 @@
 #include "parse_tree.h"
 #include "wutil.h"
 
-namespace {
-
 /// \return tokenizer flags corresponding to parse tree flags.
 static tok_flags_t tokenizer_flags_from_parse_flags(parse_tree_flags_t flags) {
     tok_flags_t tok_flags = 0;
@@ -100,6 +98,8 @@ static parse_token_type_t parse_token_type_from_tokenizer_token(
     DIE("bad token type");
     return parse_token_type_t::invalid;
 }
+
+namespace {
 
 /// A token stream generates a sequence of parser tokens, permitting arbitrary lookahead.
 class token_stream_t {
--- a/src/builtin_string.cpp
+++ b/src/builtin_string.cpp
@@ -726,6 +726,13 @@ class string_matcher_t {
         : opts(std::move(opts_)), streams(streams_), total_matched(0) {}
 
     virtual ~string_matcher_t() = default;
+
+    /// No copying or moving.
+    string_matcher_t(const string_matcher_t &) = delete;
+    string_matcher_t(string_matcher_t &&) = delete;
+    void operator=(const string_matcher_t &) = delete;
+    void operator=(string_matcher_t &&) = delete;
+
     virtual bool report_matches(const wcstring &arg) = 0;
     int match_count() const { return total_matched; }
 };
@@ -753,6 +760,12 @@ class wildcard_matcher_t : public string_matcher_t {
     }
 
     ~wildcard_matcher_t() override = default;
+
+    /// No copying or moving.
+    wildcard_matcher_t(const wildcard_matcher_t &) = delete;
+    wildcard_matcher_t(wildcard_matcher_t &&) = delete;
+    void operator=(const wildcard_matcher_t &) = delete;
+    void operator=(wildcard_matcher_t &&) = delete;
 
     bool report_matches(const wcstring &arg) override {
         // Note: --all is a no-op for glob matching since the pattern is always matched
@@ -822,6 +835,12 @@ struct compiled_regex_t {
         pcre2_match_data_free(match);
         pcre2_code_free(code);
     }
+
+    /// No copying or moving.
+    compiled_regex_t(const compiled_regex_t &) = delete;
+    compiled_regex_t(compiled_regex_t &&) = delete;
+    void operator=(const compiled_regex_t &) = delete;
+    void operator=(compiled_regex_t &&) = delete;
 };
 
 class pcre2_matcher_t : public string_matcher_t {
@@ -885,6 +904,12 @@ class pcre2_matcher_t : public string_matcher_t {
           regex(argv0_, pattern, opts.ignore_case, streams) {}
 
     ~pcre2_matcher_t() override = default;
+
+    /// No copying or moving.
+    pcre2_matcher_t(const pcre2_matcher_t &) = delete;
+    pcre2_matcher_t(pcre2_matcher_t &&) = delete;
+    void operator=(const pcre2_matcher_t &) = delete;
+    void operator=(pcre2_matcher_t &&) = delete;
 
     bool report_matches(const wcstring &arg) override {
         // A return value of true means all is well (even if no matches were found), false indicates
@@ -1041,6 +1066,13 @@ class string_replacer_t {
         : argv0(argv0_), opts(std::move(opts_)), total_replaced(0), streams(streams_) {}
 
     virtual ~string_replacer_t() = default;
+
+    /// No copying or moving.
+    string_replacer_t(const string_replacer_t &) = delete;
+    string_replacer_t(string_replacer_t &&) = delete;
+    void operator=(const string_replacer_t &) = delete;
+    void operator=(string_replacer_t &&) = delete;
+
     int replace_count() const { return total_replaced; }
     virtual bool replace_matches(const wcstring &arg) = 0;
 };
@@ -1059,6 +1091,13 @@ class literal_replacer_t : public string_replacer_t {
           patlen(pattern.length()) {}
 
     ~literal_replacer_t() override = default;
+
+    /// No copying or moving.
+    literal_replacer_t(const literal_replacer_t &) = delete;
+    literal_replacer_t(literal_replacer_t &&) = delete;
+    void operator=(const literal_replacer_t &) = delete;
+    void operator=(literal_replacer_t &&) = delete;
+
     bool replace_matches(const wcstring &arg) override;
 };
 
--- a/src/builtin_string.cpp
+++ b/src/builtin_string.cpp
@@ -1606,7 +1606,7 @@ maybe_t<int> builtin_string(parser_t &parser, io_streams_t &streams, wchar_t **a
 
     static auto begin = std::begin(string_subcommands);
     static auto end = std::end(string_subcommands);
-    string_subcommand search{subcmd_name, 0};
+    string_subcommand search{subcmd_name, nullptr};
     auto binsearch = std::lower_bound(
         begin, end, search, [&](const string_subcommand &cmd1, const string_subcommand &cmd2) {
             return wcscmp(cmd1.name, cmd2.name) < 0;
--- a/src/builtin_string.cpp
+++ b/src/builtin_string.cpp
@@ -1240,7 +1240,7 @@ static int string_split_maybe0(parser_t &parser, io_streams_t &streams, int argc
     int retval = parse_opts(&opts, &optind, is_split0 ? 0 : 1, argc, argv, parser, streams);
     if (retval != STATUS_CMD_OK) return retval;
 
-    if (opts.fields.size() < 1 && opts.allow_empty) {
+    if (opts.fields.empty() && opts.allow_empty) {
         streams.err.append_format(BUILTIN_ERR_COMBO2, cmd,
                                   _(L"--allow-empty is only valid with --fields"));
         return STATUS_INVALID_ARGS;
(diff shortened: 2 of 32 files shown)


src/fish_tests.cpp
```diff
@@ -1821,12 +1821,12 @@ class test_lru_t : public lru_cache_t<test_lru_t, int> {
...
     std::vector<value_type> evicted;
 
-    void entry_was_evicted(const wcstring &key, int val) { evicted.push_back({key, val}); }
+    void entry_was_evicted(const wcstring &key, int val) { evicted.emplace_back(key, val); }
 
     std::vector<value_type> values() const {
         std::vector<value_type> result;
         for (auto p : *this) {
-            result.push_back(p);
+            result.emplace_back(p);
```
krobelus, line 1837: I think if we use `const auto &p` in the loop, then we can elide the move as well (and keep using `push_back`).

src/ast.cpp
```diff
@@ -13,7 +13,7 @@
 namespace {
 
 /// \return tokenizer flags corresponding to parse tree flags.
-static tok_flags_t tokenizer_flags_from_parse_flags(parse_tree_flags_t flags) {
+tok_flags_t tokenizer_flags_from_parse_flags(parse_tree_flags_t flags) {
```
krobelus, line 16: My bad, I didn't realize that these were in an anonymous namespace.

I would prefer if we put the `namespace {}` only around the class (`token_stream_t`) and keep the functions static to make the fact harder to miss, see also [LLVM's rationale on this](http://llvm.org/docs/CodingStandards.html#anonymous-namespaces).

We also have other classes defined in cpp files, like `completer_t`, they should have internal linkage although I don't know if it makes a difference.

src/complete.cpp
```diff
@@ -211,9 +211,9 @@ completion_t::completion_t(wcstring comp, wcstring desc, string_fuzzy_match_t ma
       flags(resolve_auto_space(completion, flags_val)) {}
 
 completion_t::completion_t(const completion_t &) = default;
-completion_t::completion_t(completion_t &&) = default;
+completion_t::completion_t(completion_t &&) = default;  // NOLINT
```
krobelus, line 214: I assume this lint is about adding `noexcept`, which I believe is pointless since we disable exceptions. I get something similar for `lru_node_t` from `clangd`:

`src/lru.h:36:9: warning:Move constructors should be marked noexcept (fix available)`

Can we disable this lint somewhere in the project configuration? I guess clang-tidy/clangd could detect `-fno-exceptions`.
If not, we could perhaps add an explicit `NOLINT(performance-noexcept-move-constructor)`, but that's quite verbose, I'm not sure. `// NOLINT: noexcept` would work too I guess.

neheb: Currently, all performance checks are enabled: https://github.com/fish-shell/fish-shell/blob/8d2e4bbef58645e51271ae36a4f99ea4d7c5b709/.clang-tidy#L2

src/fish_tests.cpp
neheb: Didn't make a difference. clang-tidy still converted to emplace_back.

This loop could also be converted to use std::copy which internally uses push_back.

src/ast.cpp
neheb: I took out all those functions from the namespace. errors.

```
src/ast.cpp:19:41: error: use of undeclared identifier 'tokenizer_flags_from_parse_flags'
        : src_(src), tok_(src_.c_str(), tokenizer_flags_from_parse_flags(flags)) {}
                                        ^
src/ast.cpp:82:30: error: use of undeclared identifier 'parse_token_type_from_tokenizer_token'
        parse_token_t result{parse_token_type_from_tokenizer_token(token.type)};
                             ^
src/ast.cpp:84:26: error: use of undeclared identifier 'keyword_for_token'
        result.keyword = keyword_for_token(token.type, text);

```

src/complete.cpp
krobelus: Right, I disabled it, so the `// NOLINT` should not be necessary.

src/fish_tests.cpp
krobelus: Interesting!

src/complete.cpp
neheb: Removed.

krobelus: Oh sorry I have been to hasty. 
Turns out we still need to mark them `noexcept` for the STL to use them, as stated in http://clang.llvm.org/extra/clang-tidy/checks/performance-noexcept-move-constructor.html, example at https://godbolt.org/z/vzEWcc.

4c08cbd050de7488adfcad000ef2b1b61f26be7c actually does this. Now the question is, why not just add `noexcept` to this line as well? I remember there was some problem, maybe a build failure on an old system?

src/ast.cpp
krobelus: Hmm if I simply move the namespace down on master, it builds fine:

```diff
diff --git a/src/ast.cpp b/src/ast.cpp
index c8749df0f..3e20c0e99 100644
--- a/src/ast.cpp
+++ b/src/ast.cpp
@@ -11,6 +11,4 @@
 #include "wutil.h"
 
-namespace {
-
 /// \return tokenizer flags corresponding to parse tree flags.
 static tok_flags_t tokenizer_flags_from_parse_flags(parse_tree_flags_t flags) {
@@ -102,4 +100,5 @@ static parse_token_type_t parse_token_type_from_tokenizer_token(
 }
 
+namespace {
 /// A token stream generates a sequence of parser tokens, permitting arbitrary lookahead.
 class token_stream_t {
@@ -215,5 +214,4 @@ class token_stream_t {
     wcstring storage_;
 };
-
 }  // namespace
```

src/complete.cpp
neheb: I think so. I remember the buildbots failing on it.

src/ast.cpp
neheb: Fixed.

src/complete.cpp
neheb: Added noexcept. Let's see if anything fails.

neheb:
Rebased and ran through clang-format-11, which found more stuff.

Last commit was done manually. It actually found two more entries but I couldn't figure out how to convert them.

mqudsi:
What's the heuristic for determining whether to add the deleted copy/move constructors?

neheb:
The check description is here: https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines-special-member-functions.html

Note that it has options. I didn't set any of them.

src/parse_execution.h
```diff
@@ -87,7 +87,7 @@ class parse_execution_context_t {
     enum process_type_t process_type_for_command(const ast::decorated_statement_t &statement,
                                                  const wcstring &cmd) const;
     end_execution_reason_t apply_variable_assignments(
-        process_t *proc, const ast::variable_assignment_list_t &variable_assignments,
+        process_t *proc, const ast::variable_assignment_list_t &variable_assignment_list,
```
mqudsi, line 90: Was this automated? Referring to a `vector`, `[]`, `list`, etc. as plural is quite common instead of needlessly repeating the type name.

src/parser.cpp
```diff
@@ -303,8 +303,7 @@ completion_list_t parser_t::expand_argument_list(const wcstring &arg_list_src,
 
     // Get the root argument list and extract arguments from it.
     completion_list_t result;
-    const ast::freestanding_argument_list_t *list =
-        ast.top()->as<ast::freestanding_argument_list_t>();
+    const auto list = ast.top()->as<ast::freestanding_argument_list_t>();
```
mqudsi, line 306: Great reason to use `auto`, but again, `auto *` would be much more useful.

src/topic_monitor.cpp
```diff
@@ -37,12 +37,14 @@ wcstring generation_list_t::describe() const {
...
-binary_semaphore_t::binary_semaphore_t() : sem_ok_(false) {
+#ifdef __linux__
+binary_semaphore_t::binary_semaphore_t() {
     // sem_init always fails with ENOSYS on Mac and has an annoying deprecation warning.
     // On BSD sem_init uses a file descriptor under the hood which doesn't get CLOEXEC (see #7304).
     // So use fast semaphores on Linux only.
-#ifdef __linux__
     sem_ok_ = (0 == sem_init(&sem_, 0, 0));
+#else
+binary_semaphore_t::binary_semaphore_t() : sem_ok_(false) {
```
mqudsi, line 47: Was this automated? I find the original much easier to follow: here's the default value, here's the override for this case. I get that it's a redundant assignment, but realistically how much does that cost?

An alternative I would also be fine with:

```cpp
binary_semaphore_t::binary_semaphore_t() {
    // comment here
#ifdef __linux__
    sem_ok_ = (0 == sem_init(&sem_, 0, 0));
#else 
    sem_ok = false;
#endif 
}
```

Repeating the declaration twice seems like it's asking for trouble, e.g. forgetting to make changes to one of the branches when you change the other.

mqudsi (changes requested):
There are some nice improvements and I definitely find value in automating the formatting and style to enforce a sufficiently uniform codebase, but I find many of the changes to be questionable or at least sufficiently suboptimal to warrant manual review, possibly combined with turning off some automatic transformations (if that's what they are).

src/highlight.cpp
```diff
@@ -1270,7 +1270,7 @@ highlighter_t::color_array_t highlighter_t::highlight() {
     // Underline every valid path.
     if (io_ok) {
         for (const ast::node_t &node : ast) {
-            const ast::argument_t *arg = node.try_as<ast::argument_t>();
+            const auto arg = node.try_as<ast::argument_t>();
```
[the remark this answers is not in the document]
neheb: Can''t say I agree. Let's give an example

```
auto = const ast::argument_t *
const auto = const ast::argument_t * const
const auto* = const ast::argument_t * 
```

Note that this specific change was totally automated based on the currently existing .clang-tidy file.

src/complete.cpp
```diff
@@ -404,10 +404,10 @@ class completer_t {
         std::set<wcstring> visited_wrapped_commands{};
     };
 
-    void complete_custom(const wcstring &cmd, const wcstring &cmdline, custom_arg_data_t *arg_data);
+    void complete_custom(const wcstring &cmd, const wcstring &cmdline, custom_arg_data_t *ad);
 
-    void walk_wrap_chain(const wcstring &cmd, const wcstring &cmdline, source_range_t cmd_range,
-                         custom_arg_data_t *arg_data);
+    void walk_wrap_chain(const wcstring &cmd, const wcstring &cmdline, source_range_t cmdrange,
+                         custom_arg_data_t *ad);
```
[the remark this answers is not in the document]
neheb: This change just matches the function with the declaration.

src/builtin_wait.cpp
```diff
@@ -33,14 +33,12 @@ static job_id_t get_job_id_from_pid(pid_t pid, const parser_t &parser) {
...
-            return false;
-        }
-    }
-    return true;
+    // If any job is not completed, return false.
+    // If there are stopped jobs, they are ignored.
+    return std::none_of(parser.jobs().begin(), parser.jobs().end(),
+                        [](const std::shared_ptr<job_t> &j) {
+                            return j->is_constructed() && !j->is_completed() && !j->is_stopped();
+                        });
```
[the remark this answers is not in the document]
neheb: This is opt in currently. I can remove the commit if needed.

In the case of std::accumulate, better assembly is indeed generated. No idea about the _of family of functions.

src/topic_monitor.cpp
neheb: This was my solution to avoiding the warning.

src/parse_execution.h
neheb: It was automated, yes.

src/builtin_string.cpp
```diff
@@ -726,6 +726,13 @@ class string_matcher_t {
         : opts(std::move(opts_)), streams(streams_), total_matched(0) {}
 
     virtual ~string_matcher_t() = default;
+
+    /// No copying or moving.
+    string_matcher_t(const string_matcher_t &) = delete;
+    string_matcher_t(string_matcher_t &&) = delete;
+    void operator=(const string_matcher_t &) = delete;
+    void operator=(string_matcher_t &&) = delete;
+
```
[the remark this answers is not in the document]
neheb: So the deleted stuff is already provided implicitly. This just makes it explicit.

src/builtin_wait.cpp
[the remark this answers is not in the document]
neheb: Also note that several algorithms are in the codebase already. inner_product if I remember correctly was one example.

ridiculousfish:
I will manually merge some of these; I think we're clearly at the point of diminishing `clang-tidy` returns though.

ridiculousfish:
Merged the first three as cef84cf2c2b6f29771d530893c8281ad36846456; I'm declining the others because I don't think they improve clarity. Thank you.

Issue #602, opened 2013-02-28
Add zsh-like reverse-isearch


ghost:
While I love most things about fish, I had to switch back to zsh because C-r in zsh (or even bash) is more efficient than fish's reverse-search functionality.

In zsh you can type-to-complete substring-matches in history. With fish, you can cycle to see what in your history matches your substring, but you can't easily modify your substring while searching to narrow down the match.

For example if I type "git re" and go up a few times, I may find that I'm only finding a bunch of "git reset"-ish commands, but I know I want a command like "git rebase", so in zsh or bash I would just type a "b" to narrow that down. But in fish I have to restart my search with C-c and try again with "git reb". Otherwise I'm just adding a "b" to the end of whatever the last match was, which isn't what I meant to do.

This may appear trivial, but I rely heavily on C-r for all my commands (which is one reason I like fish, because it predicted that correctly about me), but I rely on it for substring-matches, not just prefix-string matches like fish does with auto-suggestions.

I'm proposing that in addition to the existing reverse-search feature, a zsh-like C-r be implemented as well.

kfirshay:
+++1

munum:
+1

numkem:
+1

Maybe implement what @jbonjean did, works pretty well.

simar7:
Add me to the "me-too" list as well. Thank you for the hardwork!

zhen9ao:
+1

pathikrit:
+1

jinchizhong:
@jbonjean really cool solution!!! thank you very much.

dsbonev:
:1

chrismacdermaid:
+1

davidmh:
For anyone interesed, [FZF](https://github.com/junegunn/fzf) as an amazing implementation of this feature for fish:

https://github.com/junegunn/fzf#key-bindings-for-command-line

![fzf's ctr-r](http://puu.sh/jqGMQ/c034f63c2c.png)

Which, IMO is a great alternative for the people worried about the "weird modal state" and as the huge advantage of the fuzzy search.

coldis:
+1 for FZF

derekstavis:
What about using a completion state for history search? We already have a great UI for it, a very nice search, and up/down arrows.

jfindlay:
[...]

PR #7361, opened 2020-09-30
Bind ctrl-r to search history if the commandline isn't empty

faho:
This is super cheesy.

One of the most common feature requests we get is "control-r must
search", even tho just using history-search-backward via e.g. up-arrow
is perfectly capable. The only real difference is that ctrl-r search
in other shells allows editing the search term by default, while we
stop the history search and edit the new commandline in those cases.

So, since the major problem is muscle-memory on ctrl-r,
let's just use that!

This makes ctrl-r do nothing on empty commandlines, and do
history-search-backward otherwise, so the basic flow of "press ctrl-r
to start history search, enter your search term, press ctrl-r to cycle
through matches" just works (except the first ctrl-r is useless and it
doesn't show anything).

See #602.

## TODOs:

### DIFF
--- a/doc_src/index.rst
+++ b/doc_src/index.rst
@@ -1575,6 +1575,8 @@ Some bindings are shared between emacs- and vi-mode because they aren't text edi
 
 - :kbd:`Control`\ +\ :kbd:`L` clears and repaints the screen.
 
+- :kbd:`Control`\ +\ :kbd:`R` searches the history if there is something in the commandline. This is mainly to ease the transition from other shells, where ctrl+r initiates the history search.
+
 - :kbd:`Control`\ +\ :kbd:`W` moves the previous path component (everything up to the previous "/", ":" or "@") to the `killring <#killring>`__.
 
 - :kbd:`Control`\ +\ :kbd:`X` copies the current buffer to the system's clipboard, :kbd:`Control`\ +\ :kbd:`V` inserts the clipboard contents.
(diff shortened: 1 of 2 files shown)


thernstig:
Will this break fish functions that some of us use, like this?

https://github.com/jethrokuan/fzf

faho:
Of course not, this will just be the *default*. If you bind anything else that takes precedence.

krobelus:
>  The only real difference is that ctrl-r search
in other shells allows editing the search term by default

The incremental search-as-you-type is an important part as well.
The lack thereof might be disappointing. Maybe we should be explicit about that:

```fish
bind --preset \cr '
    if commandline | string length -q
        commandline -f history-search-backward
    else
        echo \n"fish: incremental search is not supported. Please type your search term and press up arrow, Ctrl+P or Ctrl+R"
        commandline -f repaint
    end
'
```

faho:
> The lack thereof might be disappointing. Maybe we should be explicit about that:

But that shows an error message if people just start pressing the keys they would in bash.

The idea here is to just support the simple muscle-memory case. I'm betting 90% of ctrl-r use is just that.

Of course if we did the thing in the pager that would fully obsolete this, but as long as we don't have that, I think this is a good stop-gap.

zanchey:
I only ever used <kbd>Ctrl</kbd>+<kbd>R</kbd> as a starting point, so I don't think this is all that helpful.

faho:
Merged. We had nothing else bound to ctrl+r and this is a bit of a bandaid.

Doesn't have to be in the changelog or anything, but maybe it'll reduce screams for ctrl+r functionality by ~10% or so.

Issue #6615, opened 2020-02-16
Completing command name takes unreasonably long


bdcaf:
I encounter unreasonably slow completion for commands.
With that I mean completing `mkdi` to `mkdir` or `pop` to `popd`. 
These completions take about 10s and more and block the shell.
The examples I am giving are unique on my machine.

I use this commands frequently and also so see valid suggestions for the `Ctrl-L` shortcut.

I am on `fish, version 3.0.2` installed via homebrew on mac.
I did profile with `fish --profile profile.txt  -c 'complete -C"mkdi"'`

[profile.txt](https://github.com/fish-shell/fish-shell/files/4209511/profile.txt)
I think the relevant lines are

```
Time	Sum	Command
1742	10729	> builtin source /usr/local/Cellar/fish/3.0.2/share/fish/config.fish
...
26821	26821	> complete -C"mkdi"
```

faho:
This is a macOS bug that we've worked around in 3.1.0. It should also be fixed in the next macos release.

bdcaf:
good to know. Thanks couldn't  find any mention about it.

Issue #6611, opened 2020-02-15
Completing a command takes several seconds


Midar:
Fish 3.0.2.

When I tab-complete any command, it takes several seconds. I have seen this behavior on several different macOS and Linux machines, with the default configuration.

For example, if I type:
fis
It already suggests fish. If I type ^F to take the suggestion, that's instant. If, however, I press tab, fish hangs for several seconds.

This even happens when I complete files in the local directory. For example, if a directory has a configure script, and I type ./conf, it hangs several seconds as soon as I type tab - again despite already having the auto suggestion.

faho:
Please run the following, and then upload the output:

```fish
fish --profile /tmp/fish.prof -c 'complete -C"fis"'
sort -nk2 /tmp/fish.prof | tail -n 20
```

Feel free to censor anything you consider sensitive information.

My theory is on mac this is down to #6270, which is fixed in the just-released fish 3.1.0. On linux it might be a network filesystem or also down to bad apropos performance.

Midar:
```
105	219	-----> while read -l entry
221	221	-> for file in $__fish_config_dir/conf.d/*.fish $__fish_sysconf_dir/conf.d/*.fish $__extra_confdir/*.fish
10	249	----> if test -f $path_file
30	260	-> if test -f $__fish_data_dir/__fish_build_paths.fish
277	283	--> source /opt/pkg/share/fish/functions/__fish_set_locale.fish
67	330	-----> while read -l entry
8	363	----> if test -f $path_file
26	383	---> for x in $fish_user_paths[-1..1]
56	392	---> for entry in $$argv[1]
5	394	--> if set -q fish_user_paths
108	455	----> for path_file in $argv[2] $argv[3]/*
105	484	-> __fish_set_locale
21	499	-> __fish_reconstruct_path
29	627	---> __fish_macos_set_env 'MANPATH' '/etc/manpaths' '/etc/manpaths.d'
3	646	--> if [ -n "$MANPATH" ]
153	914	---> for path_file in $argv[2] $argv[3]/*
27	1376	--> __fish_macos_set_env 'PATH' '/etc/paths' '/etc/paths.d'
10	2125	-> if command -sq /usr/libexec/path_helper
915	4929	> builtin source /opt/pkg/share/fish/config.fish
6877	6877	> complete -C"fis"
```

faho:
That shows it taking about 13ms. Did that take about 13ms, or longer?

Is it possible there's a contributing factor we're missing?

[...]

PR #7365, opened 2020-10-01
Fast and functional replacement for __fish_describe_command for Macos

folke:
## Description

Macos no longer comes with a whatis database, which makes `__fish_describe_command` and `__fish_complete_man` command very slow, since `apropos` rebuilds the database every time it is invoked.

Trying to update the whatis database as root also doesn't work, because it's on a read only file system.

The solution I implemented, simply creates a whatis database in the user directory at `~/.cache/fish/whatis`

The whatis cache is being updated at most once a day.

The result of these changes on macos are:
* instant command completion
* instant man completion
* correctly working command descriptions

Fixes issues #6270, #6611, #6615 and possibly others.

## TODOs:

### DIFF
--- a/share/functions/__fish_apropos.fish
+++ b/share/functions/__fish_apropos.fish
@@ -1,13 +1,3 @@
-# macOS 10.15 "Catalina" has some major issues.
-# The whatis database is non-existent, so apropos tries (and fails) to create it every time,
-# which takes about half a second.
-#
-# Instead, we build a whatis database in the user cache directory
-# and override the MANPATH using that directory before we run `apropos`
-#
-# the cache is rebuilt one a day
-#
-
 if not type -q apropos
     function __fish_apropos
     end
@@ -15,6 +5,14 @@ if not type -q apropos
 end
 
 function __fish_apropos
+    # macOS 10.15 "Catalina" has some major issues.
+    # The whatis database is non-existent, so apropos tries (and fails) to create it every time,
+    # which takes about half a second.
+    #
+    # Instead, we build a whatis database in the user cache directory
+    # and override the MANPATH using that directory before we run `apropos`
+    #
+    # the cache is rebuilt once a day
     if test (uname) = Darwin
         set -l cache $HOME/.cache/fish/
         if test -n "$XDG_CACHE_HOME"
--- a/share/functions/__fish_apropos.fish
+++ b/share/functions/__fish_apropos.fish
@@ -1,0 +1,41 @@
+# macOS 10.15 "Catalina" has some major issues.
+# The whatis database is non-existent, so apropos tries (and fails) to create it every time,
+# which takes about half a second.
+#
+# Instead, we build a whatis database in the user cache directory
+# and override the MANPATH using that directory before we run `apropos`
+#
+# the cache is rebuilt one a day
+#
+
+if not type -q apropos
+    function __fish_apropos
+    end
+    exit
+end
+
+function __fish_apropos
+    if test (uname) = Darwin
+        set -l cache $HOME/.cache/fish/
+        if test -n "$XDG_CACHE_HOME"
+            set cache $XDG_CACHE_HOME/fish
+        end
+
+        set -l db $cache/whatis
+        set -l max_age 86400 # one day
+        set -l age $max_age
+
+        if test -f $db
+            set age (math (date +%s) - (stat -f %m $db))
+        end
+
+        if test $age -ge $max_age
+            echo "making cache $age $max_age"
+            mkdir -m 700 -p $cache
+            /usr/libexec/makewhatis -o $db (man --path | string split :) >/dev/null 2>&1
+        end
+        MANPATH="$cache" apropos $argv
+    else
+        apropos $argv
+    end
+end
--- a/share/functions/__fish_describe_command.fish
+++ b/share/functions/__fish_describe_command.fish
@@ -6,23 +6,32 @@
 # The whatis database is non-existent, so apropos tries (and fails) to create it every time,
 # which takes about half a second.
 #
-# Instead, we build a whatis database in the user home directory
-# and use grep to query it
+# Instead, we build a whatis database in the user cache directory
+# and override the MANPATH using that directory before we run `apropos`
 #
-# A function `__fish_apropos_update` is provided for updating
+# the cache is rebuilt one a day
 #
-function __fish_apropos -d "Internal command to query the apopos database"
+function __fish_apropos
     if test (uname) = Darwin
-        set db ~/.whatis.db
-
-        function __fish_apropos_update -d "Updates the apropos database $db on Macos"
-            echo "updating apropos / whatis database at $db"
-            man --path | tr ":" " " | xargs /usr/libexec/makewhatis -o $db
+        set -l cache $HOME/.cache/fish/
+        if test -n "$XDG_CACHE_HOME"
+            set cache $XDG_CACHE_HOME/fish
         end
 
-        [ -f $db ] || __fish_apropos_update
+        set -l db $cache/whatis
+        set -l max_age 86400 # one day
+        set -l age $max_age
 
-        /usr/bin/grep -i "$argv" $db
+        if test -f $db
+            set age (math (date +%s) - (stat -f %m $db))
+        end
+
+        if test $age -ge $max_age
+            echo "making cache $age $max_age"
+            mkdir -m 700 -p $cache
+            man --path | string split : | xargs /usr/libexec/makewhatis -o $db >/dev/null 2>&1
+        end
+        MANPATH="$cache" apropos $argv
     else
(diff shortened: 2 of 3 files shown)


folke:
If needed in other parts of the codebase, the `whatis` command can be replaced with a `grep` like so:
```fish
/usr/bin/grep -i "\b$argv\b" ~/.whatis.db
```

faho:
>If needed in other parts of the codebase, the whatis command can be replaced with a grep like so:

Are you *sure*? Is the format actually line-based like that, or is that just a coincidence? What if $argv occurs in other parts of the line?

I'd rather pass the db to whatis, if at all possible

share/functions/__fish_describe_command.fish
```diff
@@ -6,15 +6,25 @@
...
+        set db ~/.whatis.db
+
+        function __fish_apropos_update -d "Updates the apropos database $db on Macos"
+            echo "updating apropos / whatis database at $db"
+            man --path | tr ":" " " | xargs /usr/libexec/makewhatis -o $db
         end
-        # (remember: exit when `source`ing only exits the file, not the shell)
-        exit
+
+        [ -f $db ] || __fish_apropos_update
```
faho, line 23: There are two hard things in computer science:

- cache invalidation
- naming things
- off-by-one errors

This falls afoul of number 1 (or is that 0?). This will never ever update the cache, as long as fish is installed and the user doesn't delete the file. And we don't want to force the user to remember to refresh that file, that's not how we roll.

Ideally (well, ideally ideally apple would fix their stuff, but...) this could check if the man path has changed (new directories or new or changed files) - since macOS ships GNU test IIRC, it might be possible with `command test -ot` - that compares mtimes between files.

Unfortunately I don't think directory mtime works here (is that even a thing? I'm not sure), so you might have to check all files, and that might take too long.

The alternative is to do what some of our package completions do and only regenerate the file once a day or so. See e.g. https://github.com/fish-shell/fish-shell/blob/a1fd9e1b851bd876fb70da2743623f015d955935/share/functions/__fish_print_packages.fish#L63-L81

share/functions/__fish_describe_command.fish
```diff
@@ -6,15 +6,25 @@
...
+# and use grep to query it
+#
+# A function `__fish_apropos_update` is provided for updating
+#
+function __fish_apropos -d "Internal command to query the apopos database"
+    if test (uname) = Darwin
+        set db ~/.whatis.db
+
+        function __fish_apropos_update -d "Updates the apropos database $db on Macos"
+            echo "updating apropos / whatis database at $db"
```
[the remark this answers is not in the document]
faho: Which also makes the function unnecessary.

share/functions/__fish_describe_command.fish
```diff
@@ -6,15 +6,25 @@
...
+            man --path | tr ":" " " | xargs /usr/libexec/makewhatis -o $db
```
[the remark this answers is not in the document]
folke: I actually found a way to use the system `apropos` script directly (and possibly `whatis`), by setting `MANPATH` to the directory where we write the whatis database file.

I'll make these changes in a bit.

share/functions/__fish_describe_command.fish
```diff
@@ -6,15 +6,25 @@
...
         end
-        # (remember: exit when `source`ing only exits the file, not the shell)
-        exit
+
+        [ -f $db ] || __fish_apropos_update
+
+        /usr/bin/grep -i "$argv" $db
```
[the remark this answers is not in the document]
folke: I actually found a way to use the system `apropos` script directly (and possibly `whatis`), by setting `MANPATH` to the directory where we write the whatis database file.

I'll make these changes in a bit.

share/functions/__fish_describe_command.fish
folke: The once a day solution seems the way to go. Will add this.

share/functions/__fish_describe_command.fish
```diff
@@ -6,15 +6,25 @@
...
-    if test "$darwin_version[1]" = 19 -a "$darwin_version[2]" -le 3
-        function __fish_describe_command
+# Instead, we build a whatis database in the user home directory
+# and use grep to query it
+#
+# A function `__fish_apropos_update` is provided for updating
+#
+function __fish_apropos -d "Internal command to query the apopos database"
+    if test (uname) = Darwin
+        set db ~/.whatis.db
```
krobelus, line 16: `set -l` is better, in case someone has a universal variable with that name.. we should really change that.

share/functions/__fish_describe_command.fish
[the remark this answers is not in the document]
folke: done

share/functions/__fish_describe_command.fish
folke: updated!

share/functions/__fish_describe_command.fish
```diff
@@ -6,15 +6,34 @@
...
+        set -l age $max_age
+
+        if test -f $db
+            set age (math (date +%s) - (stat -f %m $db))
+        end
+
+        if test $age -ge $max_age
+            echo "making cache $age $max_age"
+            mkdir -m 700 -p $cache
+            man --path | string split : | xargs /usr/libexec/makewhatis -o $db >/dev/null 2>&1
```
faho, line 32: Nooooooooooooooooo!

Remove the `xargs`, `xargs` splits on whitespace, so this will break if you have a directory with spaces!

Just pass it via a command substitution:

```fish
/usr/libexec/makewhatis -o $db (man --path | string split :) >/dev/null 2>&1
```

folke: Fixed it. Sorry about that

share/functions/__fish_describe_command.fish
```diff
@@ -6,15 +6,34 @@
...
-    # macOS 15 is Darwin 19, this is an issue up to and including 10.15.3.
-    if test "$darwin_version[1]" = 19 -a "$darwin_version[2]" -le 3
-        function __fish_describe_command
+# Instead, we build a whatis database in the user cache directory
+# and override the MANPATH using that directory before we run `apropos`
+#
+# the cache is rebuilt one a day
+#
+function __fish_apropos
+    if test (uname) = Darwin
```
krobelus, line 15: You can probably also remove the Darwin workaround in `share/functions/__fish_complete_man.fish`.

---

I wonder if this should just be a full wrapper around apropos
`share/functions/apropos.fish` could start like

```fish
if not test (uname) = Darwin
    exit
end

function apropos
    ...
end
```

so we only override it on MacOS.

I don't really know, either way is fine by me. If we expect that a user would always prefer our version, than the wrapper is a good idea.

folke: Thought about that as well.

It definitely makes sense, since the wrapper will be able to find more as well.

I can make a similar override for `whatis` using `MANPAGE=$cache`

What do you think @faho?

folke: I've just pushed the following changes:
* created `functions/__fish_apropos.fish`
* changed `__fish_complete_man` to use `__fish_apropos` instead

share/functions/__fish_apropos.fish
```diff
@@ -0,0 +1,41 @@
+# macOS 10.15 "Catalina" has some major issues.
```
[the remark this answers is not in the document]
faho: Right before the (unfortunate) `uname` check would do nicely.

faho (approved):
Looks alright to me, if someone on macOS could confirm that it works that would be lovely.

ridiculousfish:
This is great! The only issue I see is that the `makewhatis` call is visibly slow. I think I will make it run in the background instead, if the database already exists.

ridiculousfish:
Merged as a6a0d43c9c211ad35d379a62b24da1c9b8c1d57e with my changes to run `makewhatis` in the bg as a6686ca3100c7df2e472f5a4eecf69d27175b873. Thank you, awesome contribution!!

folke:
@ridiculousfish that makes a lot of sense. Thanks!

cap10morgan:
> Looks alright to me, if someone on macOS could confirm that it works that would be lovely.

I just tried `brew install --HEAD fish` on Big Sur 11.2.1 and my command autocompletions are very fast again. Thanks!

If the 3.2.0 release isn't imminent, it might make sense to add this as a patch to Homebrew's installation of 3.1.2. I'd be happy to work up a pull request for that if others think it makes sense. This is a _huge_ UX improvement for macOS users.

faho:
>If the 3.2.0 release isn't imminent

It is.

>add this as a patch to Homebrew's installation of 3.1.2

As I understand Homebrew, that's not how they operate.

dorianmariefr:
thanks, `brew install --HEAD fish` fixed it for me ❤️

PR #7397, opened 2020-10-13
Make mysql completions work with log-in

JJK96:
## Description

Allows completing database names based on the names seen when logged in with the credentials in the current mysql command.


share/functions/__fish_complete_mysql.fish
```diff
@@ -0,0 +1,24 @@
...
+            end
+        end
+        if [ -n "$_flag_p" ]
+            set -a mysql_cmd -p$_flag_p
+        end
+        echo $mysql_cmd
+    end
+
+    function __fish_complete_mysql_databases
+        echo 'show databases' | eval (__fish_get_mysql_cmd) 2>/dev/null
```
faho, line 19: Why not just... run the command?

```fish
__fish_mysql_cmd 'show databases'
```

and then make `__fish_mysql_cmd` just run the mysql command instead of `echo`ing it?

share/functions/__fish_complete_mysql.fish
```diff
@@ -0,0 +1,24 @@
+function __fish_complete_mysql
+    set -l command $argv[1]
```
faho, line 2: This isn't used anywhere.

share/functions/__fish_complete_mysql.fish
faho: Or better yet, just do the `argparse` directly in this function and run it directly, no need for the indirection.

share/completions/mysql.fish
```diff
@@ -1,3 +1,5 @@
+__fish_complete_mysql mysql
```
faho, line 1: Just put the function in here. It's not used anywhere else, so this indirection isn't helpful.

share/functions/__fish_complete_mysql.fish
JJK96: Giving the query to `__fish_mysql_cmd` is a good idea. 

Running argparse inside the function is not a good idea in my opinion. Since in the future we possibly might add table completion or something, which would need exactly the same argument parsing. So I have split that functionality in advance to accommodate for that.

share/functions/__fish_complete_mysql.fish
JJK96: It is used by the `mysql` and `mysqldump` completions. I cannot put them both in the `mysql` file, otherwise the `mysqldump` completion will not work.
At least, that is what I understand of it, and it matches my experience (I initially had everything in one file).

share/completions/mysql.fish
JJK96: See https://github.com/fish-shell/fish-shell/pull/7397#discussion_r504672074

share/functions/__fish_complete_mysql.fish
faho: Aaah, okay, it's used down with `complete`. Sorry, I missed that. That's what happens when you try to make a quick pass during break, I guess.

Still I'd prefer if the `complete` calls happened in the complete files, so please just put `complete -c mysql ... -a '(__fish_complete_mysql_databases)` in mysql.fish, and with mysqldump in mysqldump.fish.

We've had a few completions that try to do this thing, and it's always a bit of a mess because it makes it hard to see where the options and conditions and such are defined.

JJK96: If I put `complete -c mysql ... -a '(__fish_complete_mysql_databases)` in the `completions/mysql.fish` file, I need to split `__fish_complete_mysql_databases` and `__fish_mysql_query` into separate files within `functions/`.
That might also make things less clear, let me know what you prefer.

share/functions/__fish_complete_mysql.fish
```diff
@@ -0,0 +1,25 @@
...
+
+    function __fish_mysql_query -a query
```
krobelus, line 4: I think these functions should be defined outside `__fish_complete_mysql`. They are put in the global scope either way.
Looks good otherwise.

PR #7404, opened 2020-10-17
Clear to end of each line in left prompt

### LINKED ISSUE #6532: prompt is broken when the terminal is resized if the prompt string is randomly changed.
Env:
```
$ fish --version
fish, version 3.0.2
$ uname -a
Linux acm-manjaro 4.19.97-1-MANJARO #1 SMP Sat Jan 18 00:17:24 UTC 2020 x86_64 GNU/Linux
$ echo $TERM
xterm-256color
```
I reproduced in clean env(`sh -c 'env -i TERM=$TERM HOME=$(mktemp -d) fish'`)

If I set as
```
$ function fish_prompt
    switch (random choice 1 2)
        case 1
            echo -e 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaa\n>'
        case 2
            echo -e 'bbbbbbbbbbb\n>'
    end
end
```
and I resized window, I god an merged string of them as prompt like
```
bbbbbbbbbbbaaaaaaaaaaaaaaaaaa
>
```

GIF:
deepin-screen-recorder_konsole_20200124215608

soumya92:
## Description

Changes the left prompt rendering to include clr_eol for each rendered line. This prevents issues where if a multiline prompt changes between renderings, the previous prompt shows through.

Fixes #6532

## TODOs:

### DIFF
--- a/src/screen.cpp
+++ b/src/screen.cpp
@@ -368,7 +368,7 @@ prompt_layout_t layout_cache_t::calc_prompt_layout(const wcstring &prompt_str,
     size_t prompt_len = prompt_str.size();
     const wchar_t *prompt = prompt_str.c_str();
 
-    prompt_layout_t layout = {1, 0, 0};
+    prompt_layout_t layout = {{}, 0, 0};
     wcstring trunc_prompt;
 
     size_t run_start = 0;
@@ -389,7 +389,9 @@ prompt_layout_t layout_cache_t::calc_prompt_layout(const wcstring &prompt_str,
 
         wchar_t endc = prompt[run_end];
         if (endc) {
-            layout.line_count += (endc == L'\n' || endc == L'\f');
+            if (endc == L'\n' || endc == L'\f') {
+                layout.line_breaks.push_back(trunc_prompt.size());
+            }
             trunc_prompt.push_back(endc);
             run_start = run_end + 1;
         } else {
@@ -409,7 +411,7 @@ static size_t calc_prompt_lines(const wcstring &prompt) {
     // calc_prompt_width_and_lines.
     size_t result = 1;
     if (prompt.find_first_of(L"\n\f") != wcstring::npos) {
-        result = layout_cache_t::shared.calc_prompt_layout(prompt).line_count;
+        result = layout_cache_t::shared.calc_prompt_layout(prompt).line_breaks.size() + 1;
     }
     return result;
 }
@@ -696,7 +698,8 @@ static void s_update(screen_t *scr, const wcstring &left_prompt, const wcstring 
     const scoped_buffer_t buffering(*scr);
 
     // Determine size of left and right prompt. Note these have already been truncated.
-    const size_t left_prompt_width = cached_layouts.calc_prompt_layout(left_prompt).last_line_width;
+    const prompt_layout_t left_prompt_layout = cached_layouts.calc_prompt_layout(left_prompt);
+    const size_t left_prompt_width = left_prompt_layout.last_line_width;
     const size_t right_prompt_width =
         cached_layouts.calc_prompt_layout(right_prompt).last_line_width;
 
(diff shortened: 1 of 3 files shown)


mqudsi:
Writing a character at a time is extremely suboptimal. I'd rather split on new lines and then use `s_write_str`, even with the extra allocations for each line rather than do it this way. Better yet would be to only do so if the prompt contains a new line in the first place. As I understand it, this fixes two different problems: #6532 which deals with resized terminal windows, and a multiline issue? Or is the multiline issue only in the event of a resized terminal window? If it's the latter, then we can predicate this whole business on having received a SIGWINCH and avoid the slowdown pretty much all the time.

soumya92:
https://github.com/fish-shell/fish-shell/blob/45d60f6398037e2c236f5c231fca5d90bbe52fda/src/screen.cpp#L735 should cover all cases of rendering. AIUI, if we're here, then the prompt is different, which means we _could_ run into the issue that the previous prompt had some lines that were longer and now we need to clear them.

Keeping track of which lines we need to clear is unlikely to be worth the increased complexity, so clearing all lines makes sense. I did not realise that writech would be so expensive. I assumed an internal buffer. I think we can easily take "views" into the string by jumping from one `\n` to the next, and write those in one call. I will update this PR shortly.

soumya92:
Okay, updated to write each line at a time. I'm not sure if c_str() causes an allocation, but I think we can live with that even if it does, since this shouldn't be running too frequently (given the if check already in place).

mqudsi:
Thanks for updating the patch. The `.c_str()` call wouldn't allocate, but the `.substr()` call would. With C++14, `.substr()` could return a `string_view` (directly?) which doesn't, but calling `.c_str()` afterwards necessarily would since it must be both contiguous and null-terminated.

src/screen.cpp
```diff
@@ -734,7 +734,19 @@ static void s_update(screen_t *scr, const wcstring &left_prompt, const wcstring
     // Output the left prompt if it has changed.
     if (left_prompt != scr->actual_left_prompt) {
         s_move(scr, 0, 0);
-        s_write_str(scr, left_prompt.c_str());
+        if (clr_eol) {
+            s_write_mbs(scr, clr_eol);
+        }
```
mqudsi, line 739: I would move this to after the last write to minimize flashes of contents/repaints. But first - in what case is it required, given that it would presumably apply to even single-line prompts but we don't have any problems with those now, even without it?

soumya92: That's a good idea. I will make that change. It might make the logic slightly more complicated, since currently it goes "clr, line+terminator", which will need to be changed into "terminator+line, clr". Just some indexing stuff to take care of :)

I think the fact that it's being printed on a single line prompt is also just an artefact of that, so I can definitely eliminate it. We don't have a problem for single line prompts because rendering the command also clears to the end of the line.

soumya92: Updated, let me know if you have any suggestions to make the logic clearer.

src/screen.cpp
```diff
@@ -734,7 +734,18 @@ static void s_update(screen_t *scr, const wcstring &left_prompt, const wcstring
...
+        size_t next_nl = wcstring::npos;
+        while ((next_nl = left_prompt.find('\n', start)) != wcstring::npos) {
+            s_write_str(scr, left_prompt.substr(prev_nl, next_nl - prev_nl).c_str());
+            if (clr_eol) {
+                s_write_mbs(scr, clr_eol);
+            }
+            prev_nl = next_nl;
+            start = prev_nl + 1;
+        }
+        s_write_str(scr, left_prompt.substr(prev_nl).c_str());
```
krobelus, line 748: We can avoid allocation in the common case by using `left_prompt.c_str() + prev_nl`.

We *could* also avoid allocations from the `substr` above, if we temporarily swap out newlines for null bytes. 

```cpp
        while ((next_nl = left_prompt.find('\n', start)) != wcstring::npos) {
	        left_prompt.at(next_nl) = L'\0';
            s_write_str(scr, left_prompt.c_str() + prev_nl);
	        left_prompt.at(next_nl) = L'\n';
	        ...
        }
```

However, I don't think swapping null bytes is common in the fish code base, so when in doubt I'd leave that part as is. What do others think?

soumya92:
I had to change the logic a little bit because I realised that my first attempt would not handle '\n' inside escape sequences properly.

src/screen.cpp
soumya92: Done for the common case. Agreed that swapping '\n' with '\0' seems dubious. If we're going that far, we might as well make left_prompt a wcstring_list_t instead of a single wcstring. WDYT?

ridiculousfish:
Merged as 80aaae5b74eaadbb0cc563648b9af53d715f94e2, thank you!

soumya92:
Happy to help! I think there's still room to add the wcstring_list_t change as well, so I'll probably come back to it later.

Issue #3566, opened 2016-11-19
Keep part of the prompt_pwd the full length.


sorahn:
Hey guys, I am trying to have my `prompt_pwd` command spit out something like this `~/s/[project]/f/o/o/b/a/r` and I was directed here to try and figure out a generic way to do it.

Ideally the directory that it would keep is the top level folder once you're inside any git repository.

Ideas I had:
* some `fish_expand_directory_in_pwd_by_index` setting that is set to some int. (lazy copout to a real solution)
* testing for existence of a .git directory? (is this too expensive?)
* string matching with something like `git rev-parse --show-toplevel`

Let me know what you guys think.

krader1961:
For git using either `git rev-parse --show-prefix` or `git rev-parse --show-toplevel` is probably the best solution. Both because it is almost certainly faster than having fish walk the directory hierarchy looking for a *.git* directory (although that should be benchmarked) and because it's considerably simpler than doing it "by hand".

Ideally we would also do this for SVN and Mercurial assuming it is cheap enough. Trying each in order and using the first one that is successful. If doing this is expensive (say more than 10 ms) cache the result in a global var that is updated using a `--on-variable PWD` function to calculate a new short form for use in the prompt.

faho:
We already have these at least for hg and git (dunno if the svn one could actually print the toplevel repo directory), and they are already run once every prompt, at least if you are using the `__fish_vcs_prompt` function (which the samples are. And yes, the dunderscore should be removed).

I'd say they are reasonably cheap - whatever you do with git is basically lightning fast unless your repo is huge, mercurial is dog-slow (IIRC half a second just for "is this a repo?"), but we sidestep it.

However, it's going to be tricky doing this without needing to execute these functions twice, especially if it's supposed to be prompt-agnostic (i.e. just a function you stick in there to shorten the pwd with project, and another to print the branch).

Also, I'm unsure how this should look and if it should be turned on by default, or if it would be too unfamiliar and therefore confusing. Maybe another function, `prompt_project_pwd` or so? Or an option, `prompt_pwd --project` or "--vcs"?

Anyway, I have a quick'n'dirty prototype:

```fish
function prompt_pwd --description 'Print the current working directory, shortened to fit the prompt'
    set -q argv[1]
    and switch $argv[1]
        case -h --help
            __fish_print_help prompt_pwd
            return 0
    end

    # This allows overriding fish_prompt_pwd_dir_length from the outside (global or universal) without leaking it
[...]

PR #7414, opened 2020-10-23
Expand the full view directory in prompt

hotoolong:
## Description

Allows you to set the range in which the full directory is displayed in the prompt.

For example
If I was working in
```
/Users/hotoolong/.ghq/github.com/hotoolong/translate.nvim
```

By default, the display is as follows. (For my settings)

```
~/.g/g/h/translate.nvim (master|✔) $ 
```

If you set the full directory to display to 3,

```
set -g fish_prompt_dir_full_name_range 3
```

It appears as follows

```
~/.g/github.com/hotoolong/translate.nvim (master|✔) $
```

The following is similar issue.
#3566

## TODOs:

### DIFF
--- a/share/functions/prompt_pwd.fish
+++ b/share/functions/prompt_pwd.fish
@@ -19,7 +19,13 @@ function prompt_pwd --description "Print the current working directory, shortene
     if [ $fish_prompt_pwd_dir_length -eq 0 ]
         echo $tmp
     else
+        set -q fish_prompt_dir_full_name_range
+        or set -l fish_prompt_dir_full_name_range 1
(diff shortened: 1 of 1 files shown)


faho:
This already exists as $fish_prompt_pwd_dir_length, see https://github.com/fish-shell/fish-shell/blob/master/share/functions/prompt_pwd.fish.

So there is no need for this PR.

hotoolong:
Thanks for the quick reply. I was hoping that some way of displaying the parent directory from the current directory would be added.
It's a shame this time, but I'll try to deal with it by overwriting the function.

Issue #3066, opened 2016-05-25
shmem notifier strategy has dubious security


amluto:
The shmem notifier code calls shm_open on a predictable filename.  On glibc, at least, shm_open has countermeasures to the obvious symlink attacks, but it's still a bit dubious.  If nothing else, another user can DoS you by creating something there.  Worse attacks may be possible.

Using $XDG_RUNTIME_DIR would be better.

This is mitigated by the fact that this isn't used on Linux by default.

ridiculousfish:
Thanks!!

Can you elaborate on what the DoS attack is? Is the idea that, by writing to the file, an attacker could force fish to re-read its variables? Also, how would we go about making this filename unpredictable, while still allowing fish to find it?

amluto:
Another user can just create /dev/shm/fish-shmem-UID before the victim logs in.

I would suggest moving it into $XDG_RUNTIME_DIR (ie. /run/user/UID on Linux) just like the notify pipe.

ridiculousfish:
`shm_open()` can be implemented as an ordinary filesystem path, or a flat namespace separate from the filesystem. Linux and OS X both use the flat namespace approach, which means that the "path" is limited (30 characters on OS X, 255 on Linux, in my tests).  Linux goes so far as to disallow slashes in `shm_open()` entirely. So unfortunately using `$XDG_RUNTIME_DIR` will just fail with EINVAL.

amluto:
On May 25, 2016 11:22 AM, "ridiculousfish" <notifications@github.com> wrote:
>
> shm_open() can be implemented as an ordinary filesystem path, or a flat
namespace separate from the filesystem. Linux and OS X both use the flat
namespace approach, which means that the "path" is limited (30 characters
on OS X, 255 on Linux, in my tests). Linux goes so far as to disallow
slashes in shm_open() entirely. So unfortunately using $XDG_RUNTIME_DIR
will just fail with EINVAL.
>

Oh... I meant using a real file in $XDG_RUNTIME_DIR and mmapping it.  After
all, on Linux, a shm segment is just a regular file.

ridiculousfish:
Gotcha. That makes sense.

zanchey:
Is this a small enough change for 2.3.1?

ridiculousfish:
The shmem notifier is by default used on Cygwin, but not Linux. I'm not sure how much of the XDG stuff is implemented on Cygwin. Anyways I think we should punt it.

PR #7430, opened 2020-10-27
Add universal variable notifier based off of shared memory

### LINKED ISSUE #7429: SIGIO notifier does not work under WSL
The SIGIO uvar notifier does not seem to work under WSL; I have not dug into it to see if SIGIO is being sent or if it's something more subtle.

```
Testing universal variable legacy parsing
Testing universal callbacks
Testing universal format detection
Testing universal Ok to save
Testing universal notifiers with strategy 2
Error: Universal variable notifier (1) 0x1959a60 polled failed to notice changes, with strategy 2
Error: Universal variable notifier (2) 0x19252a0 polled failed to notice changes, with strategy 2
Error: Universal variable notifier (3) 0x193d730 polled failed to notice changes, with strategy 2
Error: Universal variable notifier (4) 0x190e560 polled failed to notice changes, with strategy 2
Error: Universal variable notifier (5) 0x19581f0 polled failed to notice changes, with strategy 2
Error: Universal variable notifier (6) 0x1914370 polled failed to notice changes, with strategy 2
Error: Universal variable notifier (7) 0x18fb6e0 polled failed to notice changes, with strategy 2
Error: Universal variable notifier (8) 0x18fc4a0 polled failed to notice changes, with strategy 2
Error: Universal variable notifier (9) 0x191f060 polled failed to notice changes, with strategy 2
Error: Universal variable notifier (10) 0x18fb700 polled failed to notice changes, with strategy 2
Error: Universal variable notifier (11) 0x195a670 polled failed to notice changes, with strategy 2
Error: Universal variable notifier (12) 0x1918090 polled failed to notice changes, with strategy 2
Error: Universal variable notifier (13) 0x1958310 polled failed to notice changes, with strategy 2
Error: Universal variable notifier (14) 0x1943af0 polled failed to notice changes, with strategy 2
Error: Universal variable notifier (15) 0x18fb570 polled failed to notice changes, with strategy 2
Error: Universal variable notifier (0) 0x1957df0 polled failed to notice changes, with strategy 2
Error: Universal variable notifier (2) 0x19252a0 polled failed to notice changes, with strategy 2
Error: Universal variable notifier (3) 0x193d730 polled failed to notice changes, with strategy 2
Error: Universal variable notifier (4) 0x190e560 polled failed to notice changes, with strategy 2
Error: Universal variable notifier (5) 0x19581f0 polled failed to notice changes, with strategy 2
Error: Universal variable notifier (5) 0x19581f0 polled failed to notice changes, with strategy 2
Error: Universal variable notifier (6) 0x1914370 polled failed to notice changes, with strategy 2
Error: Universal variable notifier (7) 0x18fb6e0 polled failed to notice changes, with strategy 2
Error: Universal variable notifier (8) 0x18fc4a0 polled failed to notice changes, with strategy 2
Error: Universal variable notifier (9) 0x191f060 polled failed to notice changes, with strategy 2
Error: Universal variable notifier (10) 0x18fb700 polled failed to notice changes, with strategy 2
Error: Universal variable notifier (11) 0x195a670 polled failed to notice changes, with strategy 2
Error: Universal variable notifier (12) 0x1918090 polled failed to notice changes, with strategy 2
// 210 lines snipped
Error: Universal variable notifier (13) 0x1958310 polled failed to notice changes, with strategy 2
Error: Universal variable notifier (14) 0x1943af0 polled failed to notice changes, with strategy 2
Testing completion insertions
Testing scenarios that should produce no autosuggestions
Testing autosuggestion combining
Testing history
Testing history merge
^CTesting history_sample_fish_1_x
Testing history_sample_fish_2_0
Testing bash import
Testing history_sample_corrupt1
Testing builtin_string
Testing illegal command exit code
/mnt/d/rand/fish/build/test/temp
Testing maybe_t
Testing path normalization
Testing topic monitor
Torture-testing topic monitor
Testing timer format
Testing termsize
Encountered 240 errors in low-level tests
ninja: build stopped: interrupted by user.
GNUmakefile:59: recipe for target 'test' failed
make: *** [test] Interrupt
```

mqudsi:
This is a conceptually more straight-forward solution for synchronizing data between multiple fish instances, using the textbook shared memory approach. My only concerns with the exceedingly clever SIGIO approach mainly stem from the subtleties involved with relying on side effects of async reads/writes to drive the notifier (e.g. the variety of differences in just when SIGIO is generated), and a fear that if we adopt asynchronous IO in the future that might cause issues with our current, exclusive interpretation of SIGIO.

I'm not sure how the SIGIO notifier is actually "notifying" since it just returns true/false when polled (I imagine the read loop has been modified so as to guarantee the notifier is polled prior to any operation that might read a universal variable?), but if it is desirable to use the notifier in a non-polling manner the `block()` method will sleep the calling thread until a uvar change is detected. If it's desired to use it in an asynchronous context with an event loop (e.g. with select/poll/epoll), it is trivial to add an anonymous pipe and use the self-pipe trick to provide an fd that can be given to the kernel to wait on by calling `notifier.block()` in a new thread and writing (then immediately reading) a byte to/from the pipe when awoken in a loop.

This notifier has been tested on Linux, FreeBSD, WSL, and macOS and passes all the tests. There is currently no `#ifdef` for selecting the strategy, suggestions are welcome.

The code currently uses the GCC atomic extensions (CMake patched accordingly) to atomically read/write the one- to four-byte fields in the shared memory. This may be overkill: presumably the pointer nature of the shared memory object would suffice to prevent the compiler from making assumptions that would lead to stale reads; the addition of a `volatile` modifier to the storage class of either the SHM pointer or the individual fields would definitely suffice, but that's apparently no longer "in vogue" (suggestions again welcome).

Closes #7429

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -183,7 +183,7 @@ add_library(fishlib STATIC ${FISH_SRCS})
 target_sources(fishlib PRIVATE ${FISH_HEADERS})
 target_link_libraries(fishlib
   ${CURSES_LIBRARY} ${CURSES_EXTRA_LIBRARY} Threads::Threads ${CMAKE_DL_LIBS}
-  ${PCRE2_LIB} ${Intl_LIBRARIES} ${ATOMIC_LIBRARY})
+  ${PCRE2_LIB} ${Intl_LIBRARIES} ${ATOMIC_LIBRARY} ${RT_LIBRARY})
 target_include_directories(fishlib PRIVATE
   ${CURSES_INCLUDE_DIRS})
 
--- a/cmake/ConfigureChecks.cmake
+++ b/cmake/ConfigureChecks.cmake
@@ -225,6 +225,18 @@ IF (NOT LIBATOMIC_NOT_NEEDED)
     set(ATOMIC_LIBRARY "atomic")
 endif()
 
+check_cxx_source_compiles("
+#include <sys/mman.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+int main() {
+shm_open(\"cmake-shm-test-foo\", O_CREAT|O_RDWR, 0666);
+}"
+LIBRT_NOT_NEEDED)
+IF (NOT LIBRT_NOT_NEEDED)
+    set(RT_LIBRARY "rt")
+endif()
+
 # Check if mbrtowc implementation attempts to encode invalid UTF-8 sequences
 # Known culprits: at least some versions of macOS (confirmed Snow Leopard and Yosemite)
 try_run(mbrtowc_invalid_utf8_exit mbrtowc_invalid_utf8_compiles ${CMAKE_CURRENT_BINARY_DIR}
--- a/src/env_universal_common.cpp
+++ b/src/env_universal_common.cpp
@@ -5,12 +5,16 @@
 #include <errno.h>
 #include <fcntl.h>
 // We need the sys/file.h for the flock() declaration on Linux but not OS X.
+#include <pthread.h>
 #include <sys/file.h>  // IWYU pragma: keep
 // We need the ioctl.h header so we can check if SIOCGIFHWADDR is defined by it so we know if we're
 // on a Linux system.
 #include <limits.h>
 #include <netinet/in.h>  // IWYU pragma: keep
 #include <sys/ioctl.h>   // IWYU pragma: keep
+#include <sys/mman.h>
+#include <memory>
+#include <sstream>
 #if !defined(__APPLE__) && !defined(__CYGWIN__)
 #include <pwd.h>
 #endif
@@ -36,6 +40,7 @@
 #include <map>
 #include <string>
 #include <type_traits>
+#include <thread>
 #include <utility>
 
 #include "common.h"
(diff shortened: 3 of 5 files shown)


faho:
Note that we also have a strategy called "shmem_polling". How is this different and how does it avoid #3066?

Also I have to admit that I'm not sure what these recent added strategies are supposed to *do*, I'm not aware of many bug reports about uvar notifying, so naively I would just assume that we could leave the situation as it is?

What's the upside here?

mqudsi:
> Note that we also have a strategy called "shmem_polling". 

I primarily am not sure why Cygwin has its own notifier and did not want to make any modifications there.

> How is this different 

The conditional variable in the shared memory section eliminates the need for repeated polling. The poll frequency is set to `0`.

> how does it avoid #3066?

Thanks for linking that; I hadn't seen it. That is trivially solvable by generating a random per-user seed that can be mixed into the hash and stored as a universal variable (starting the notifier isn't required before reading extant universal variables, so there is no recursion here).

> Also I have to admit that I'm not sure what these recent added strategies are supposed to do, I'm not aware of many bug reports about uvar notifying, so naively I would just assume that we could leave the situation as it is?
> 
> What's the upside here?

I'm guessing @ridiculousfish can answer this more definitively; I imagine there was something about the named pipe approach that he disliked. The motivation for this particular PR came from the recent notifier breakage under FreeBSD and WSL, although @ridiculousfish stepped in and fixed the FreeBSD issue very nicely, of course.

ridiculousfish:
So unfortunately I don't think this works. The main responsibility of the notifier is to interrupt `select()`. When the user is waiting at the prompt, `select()` is waiting on stdin (and possibly other file descriptors). When a uvar change occurs, the notifier has to interrupt `select()` so that fish can detect the change and perhaps redraw the command line.

There are three ways to get `select` to return:

1. A file descriptor is readable. This is used by `named_pipe` (and `notifyd` which is Mac only).
2. A signal is delivered. This is used by `sigio`.
3. The call times out. This is used by `shmem_polling`.

Just writing into shared memory does not interrupt select. You can see that by opening two terminals and changing `fish_color_command` in one: the other should update instantly, and that doesn't happen with this new notifier.

Unfortunately the automated tests don't have coverage of this :(

The notifiers add a lot of complexity for this feature, and I think it would be OK to only support the instant-update feature on certain platforms.  On Mac we can use `notifyd`, on Linux we can use...maybe `epoll` or `dbus` or `inotify` to watch the file itself (though [I already tried that](https://github.com/ridiculousfish/travis-inotify-test)), on other platforms we perhaps just disable the feature outright.

Anyways it would be really nice to find a simple solution for this on Linux; I've been racking my brain trying to think of a simple way to announce that an event occurred and everything I've come up with feels way more complex than it needs to be.

ridiculousfish:
So the main responsibility of the notifier is to interrupt `select()`. When the user is waiting at the prompt, `select()` is waiting on stdin (and possibly other file descriptors). When a uvar change occurs, the notifier has to interrupt `select()` so that fish can detect the change and perhaps redraw the command line.

There are three ways to get `select` to return:

1. A file descriptor is readable. This is used by `named_pipe` (and `notifyd` which is Mac only).
2. A signal is delivered. This is used by `sigio`.
3. The call times out. This is used by `shmem_polling`.

Just writing into shared memory does not interrupt select. You can see that by opening two terminals and changing `fish_color_command` in one: the other should update instantly, and that doesn't happen with this new notifier.

As @mqudsi observes, this could be done with a separate thread that waits on a condition variable in shared memory. However I think dedicating a thread for this is too heavyweight, plus `pthread_condattr_setpshared` is not supported everywhere, plus you have to contend with the risk of another fish instance dying while holding the lock, so that it never gets released. So I think this approach is worse overall.

Unfortunately the automated tests don't have coverage of this :(

This instant-update feature has a "wow" factor but the notifiers sure are complex. I think it would be OK to only support the instant-update feature on certain platforms.  On Mac we can use `notifyd`, on Linux I think the best path is `inotify` to watch the file itself (though [I ran into problems when trying that](https://github.com/ridiculousfish/travis-inotify-test)), on other platforms we perhaps just disable the feature outright?

Anyways it would be really nice to find a simple solution for this on Linux; I've been racking my brain trying to think of a simple way to IPC-broadcast that some event occurred, and everything I've come up with feels way more complex than it needs to be.

mqudsi:
Thanks for the additional info; I was perplexed by how the notifier was waking the read loop - I didn't know SIGIO would interrupt a select/poll call. I agree with the concerns about the dangers of having a thread die with the lock held. There doesn't seem to be unified support for recovering a "poisoned" mutex, which is... restrictive.

In this particular case, `inotify` on Linux and `kqueue` on macOS and BSD do sound like pretty good fits since we are literally watching for updates to a file, even if they don't answer the question for the general IPC case. Windows has a very nice solution here: every application is expected to [have a message pump on the main thread](https://en.wikipedia.org/wiki/Message_loop_in_Microsoft_Windows) that listens for messages posted from the OS, other apps, or even itself. It's been around since the very beginning and is really simple to set up (just ignoring all events) and avoids all threading issues by using message passing rather than shared memory.

SysV or POSIX message queues have some similarities here, perhaps they're options?

Issue #7315, opened 2020-09-06
command documentation: add a link to the command sources


aureliojargas:
Sometimes, when the official documentation is not enough, one has to see the actual source code to find the information she/he is looking for.

I would be nice to have a handy link pointing to the source code, right there in the documentation text.

It's currently a little tricky to find the sources since it can be a cpp file:

- Doc: https://fishshell.com/docs/current/cmds/set_color.html
- Source: https://github.com/fish-shell/fish-shell/blob/master/src/builtin_set_color.cpp

or a fish function:

- Doc: https://fishshell.com/docs/current/cmds/isatty.html
- Source: https://github.com/fish-shell/fish-shell/blob/master/share/functions/isatty.fish

One example is the Python documentation, where the link to the source is right there at the top:

> https://docs.python.org/3/library/argparse.html
> ![image](https://user-images.githubusercontent.com/282592/92319289-b7186f00-f016-11ea-9dfb-2f0c4089585f.png)

faho:
Interesting idea.

A few questions:

- How do we figure out the correct *version*? Just linking to the master branch is probably not enough.
- Do we link to functions *on the running system*? That would be much nicer to have. Or do we tell people to use `type $function`?

PR #7437, opened 2020-10-29
docs: Add a `:issue:` role and use it in the CHANGELOG

faho:
This allows us to write the changelog reasonably simply.

The biggest downside is that pandoc won't be able to handle it anymore
when converting to markdown, but
sphinx-markdown-builder (https://github.com/codejamninja/sphinx-markdown-builder)
will turn it into explicit links (`[#number](https://github.com/...)`)

@zanchey Is that okay for you?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,40 +4,40 @@ fish 3.2.0 (released ???)
 Notable improvements and fixes
 ------------------------------
 
--  Undo and redo support for the command-line editor and pager search (#1367). By default, undo is bound to Control+Z, and redo to Alt+/.
+-  Undo and redo support for the command-line editor and pager search (:issue:`1367`). By default, undo is bound to Control+Z, and redo to Alt+/.
 -  builtins may now output before all data is read. For example, ``string replace`` no longer has to read all of stdin before it can begin to output.
    This makes it usable also for pipes where the previous command hasn't finished yet, like::
 
     # Show all dmesg lines related to "usb"
     dmesg -w | string match '*usb*'
 
--  A new variable, ``fish_kill_signal``, is set to the signal that terminated the last foreground job, or ``0`` if the job exited normally (#6824).
+-  A new variable, ``fish_kill_signal``, is set to the signal that terminated the last foreground job, or ``0`` if the job exited normally (:issue:`6824`).
 -  Control-C no longer kills background jobs for which job control is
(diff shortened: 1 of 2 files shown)


krobelus:
OK so this allows to render issue links outside GitHub.
I wonder it's better to use a simple regex replacement that allows us to keep writing `#XYZ`. Does Sphinx support custom preprocessors?

faho:
Github doesn't link the `#XYZ` in .rst files, so I don't think adding an extra preprocessor step is worth it.

The sphinx way here really appears to be custom roles (and I'm planning to add more like `:cmd:` and `:source:` for #7315), I don't think it can do custom arbitrary transformations easily. I also like the more explicit style, even if it is slightly more typing (and slightly awkward on a qwertz layout)

zanchey:
It doesn't link in .md files either (viz https://github.com/fish-shell/fish-shell/blob/2d52335fed81fe1734aef3291aa830b6cfc43385/CHANGELOG.md) but does when the Markdown is pasted into the release description. So I don't think this is needed. The fish-site scripts do the same work.

Annoyingly the releases metadata can only be filled in with Markdown, so either we move the whole CHANGELOG back to Markdown (my preferred option) or I run it through Pandoc before making a release.

faho:
>Annoyingly the releases metadata can only be filled in with Markdown, so either we move the whole CHANGELOG back to Markdown (my preferred option)

That's of course possible.

However I would like it if we could ship the changelog also in the docs, and that requires that it be readable by sphinx, which means either we ship it as .rst (possibly with conversion, but that's annoying in the git-builds), or require a sphinx extension for markdown support.

zanchey:
Sure, I'm not fussed, I can install pandoc or something I guess?

faho:
Pandoc can do the rst format, but as soon as you add a custom role it won't know what to do with it (because it doesn't read Sphinx' config script).

So you'd have to use a sphinx extension like [sphinx-markdown-builder](https://github.com/codejamninja/sphinx-markdown-builder). Just `pip install sphinx-markdown-builder` and then `sphinx-build -b markdown /path/to/doc_src /path/to/output`.

I've uploaded a conversion at https://gist.github.com/faho/a52e328a4e432a03d1114ba615763615.

(one obvious possible enhancement is to add highlighting to the codeblocks via \`\`\`fish, I'm not sure if that's possible)

PR #7438, opened 2020-10-29
Add bluetoothctl completions

moverest:


faho:
Merged, thanks!

I have to say I like the sample output in the comments!

PR #7471, opened 2020-11-09
Don't assume $__fish_git_prompt_char_cleanstate is non-empty

shlemas:
This fixes the case where an empty "clean state" character
can cause a spurious space character at the end of the git prompt.

## Description

I've been trying to integrate fish_git_prompt into a custom prompt and running into minor issues. The first was fixed by discovering the undocumented argument where you can pass in a format string. The second is fixed by this commit. When running the prompt in informative mode with `$__fish_git_prompt_char_cleanstate` set to an empty string while inside a clean repository, the prompt looks like this:

`(master )`

Note the spurious space after `master`. Minor but a bit ugly/annoying. Even when passing in "%s" as the argument to fish_git_prompt and running the result through `string trim`, the output will still be `master ` with a trailing space, presumably because some color codes are getting in the way of the trim operation. So this commit adds a couple simple checks in and around the informative status function to make sure we aren't echoing spurious output.

Thanks!

## TODOs:

### DIFF
--- a/share/functions/fish_git_prompt.fish
+++ b/share/functions/fish_git_prompt.fish
@@ -403,7 +403,10 @@ function fish_git_prompt --description "Prompt function for Git"
                 and test "$dirty" != false
                 and test "$untracked" != false
             end
-            set informative_status "$space"(__fish_git_prompt_informative_status $git_dir)
+            set informative_status (__fish_git_prompt_informative_status $git_dir)
+            if test -n "$informative_status"
+                set informative_status "$space$informative_status"
+            end
         else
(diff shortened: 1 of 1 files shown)


shlemas:
I looked at the test failure but it doesn't appear to be related to my change. Let me know if I'm wrong about that.

faho:
It's not, Github's macos test machines are just awful, and we need to work out some of the kinks.

(necessary because Travis made their Open-source plan unusable)

faho:
Merged, thanks!

PR #7476, opened 2020-11-12
escape a backslash (in 2 places) in argparse docs

marcusatiliusregulus:


faho:
Good catch, thanks!

One of these days I'll understand rst!

PR #7477, opened 2020-11-13
build_tools/fish_xgettext.fish: avoid printing error message when

ericonr:
/tmp/fish doesn't exist.

 The command invocation used was `rm -r /tmp/fish`, when it should have
 used -rf.

### DIFF
--- a/build_tools/fish_xgettext.fish
+++ b/build_tools/fish_xgettext.fish
@@ -22,34 +22,39 @@ set -l implicit_regex '(?:^| +)(?:complete|function).*? (?:-d|--description) (([
 # than messages which should be implicitly translated.
 set -l explicit_regex '.*\( *_ (([\'"]).+?(?<!\\\\)\\2) *\).*'
 
-rm -r /tmp/fish
+# Create temporary directory for these operations. OS X `mktemp` is somewhat restricted, so this block
+# works around that - based on share/functions/funced.fish.
+set -q TMPDIR
+or set -l TMPDIR /tmp
+set -l tmpdir (mktemp -d $TMPDIR/fish.XXXXXX)
+or exit 1
 
-mkdir -p /tmp/fish/implicit/share/completions /tmp/fish/implicit/share/functions
-mkdir -p /tmp/fish/explicit/share/completions /tmp/fish/explicit/share/functions
+mkdir -p $tmpdir/implicit/share/completions $tmpdir/implicit/share/functions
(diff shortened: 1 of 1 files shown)


faho:
I don't think we should be force-deleting this directory at all. The name is super general, and force deleting is quite the dangerous tool.

Instead we should be using `mktemp -d`.

Any chance you'd want to work on that? Otherwise I can do it.

See [funced](https://github.com/fish-shell/fish-shell/blob/7b9c1a60765c0b326aa1ec0ba4324c8fd63b03f4/share/functions/funced.fish#L77-L83) for how to use it properly.

ericonr:
That is indeed a much better solution. I can do it.

ericonr:
I also added a sanity check for empty `tmpdir` when `mktemp -d` fails. Should I add it to `funced`?

build_tools/fish_xgettext.fish
```diff
@@ -22,34 +22,40 @@ set -l implicit_regex '(?:^| +)(?:complete|function).*? (?:-d|--description) (([
 # than messages which should be implicitly translated.
 set -l explicit_regex '.*\( *_ (([\'"]).+?(?<!\\\\)\\2) *\).*'
 
-rm -r /tmp/fish
+# Create temporary directory for these operations. OS X `mktemp` is somewhat restricted, so this block
+# works around that - based on share/functions/funced.fish.
+set -q TMPDIR
+or set -l TMPDIR /tmp
+set -l tmpdir (mktemp -d $TMPDIR/fish.XXXXXX)
+test "$tmpdir"
```
faho, line 30: I'd just directly do `or exit 1` here.

The $status is the one of `mktemp`, and if that fails it should return false. Otherwise your system is beyond repair.

Or at the very least use `test -n` or `test -z` - the one argument form is one of the worst mistakes in POSIX, and cause of most of our issues with `test`.

faho:
>I also added a sanity check for empty tmpdir when mktemp -d fails. Should I add it to funced?

Sure. As a separate commit that'd be cool. Also I think we use `mktemp` in more places.

build_tools/fish_xgettext.fish
ericonr: > The $status is the one of mktemp, and if that fails it should return false. Otherwise your system is beyond repair.

Heh, fair enough :)

ericonr:
I will make a separate PR to fix those. Consider this PR ready for review.

faho:
Yup, that's looking good. Merged, thanks!

PR #7478, opened 2020-11-13
src/builtin_type.cpp: add missing gettext call.

ericonr:
The string "%ls is %ls", which is printed when `type <command>` is ran
for a command in PATH, couldn't be localized, since it was missing _()
around it.


faho:
Thanks, merged!

PR #7480, opened 2020-11-13
po/pt_BR: update translations.

ericonr:
## Description

Do I need to add it to the changelog?


faho:
>Do I need to add it to the changelog?

Nah, we can do it. If we get any more localisations we'd just add a list like

"Updated localisations for pt_BR (#7480), xy_YX (#banana)"

Merged, thanks!

PR #7481, opened 2020-11-13
Localization fixes

ericonr:
## Description

Related to issues I noticed in #7480 

- normalized some strings so they were simpler and there would be less to translate
- worked around a xgettext bug when dealing with situations like `_(L"1 " L"2")`, which generates a msgid of `"1 "`.


faho:
> worked around a xgettext bug when dealing with situations like _(L"1 " L"2"), which generates a msgid of "1 ".

Ooh, that's annoying.

We should check if clang-format breaks this in its overzealous quest for keeping the line limit.

Anyway, merged, thanks!

ericonr:
For the record, I'm tracking the bug here https://lists.gnu.org/archive/html/bug-gettext/2020-11/msg00013.html

PR #7482, opened 2020-11-13
Account for mktemp failures

ericonr:
## Description

Discussed in #7477, catches when `mktemp` errors out and exits the script / function.

I can try to add some error case that exports `TMPDIR` to some place that can't be written, if you find it necessary.

## TODOs:

### DIFF
--- a/build_tools/list_committers_since.fish
+++ b/build_tools/list_committers_since.fish
@@ -13,7 +13,9 @@ if not contains -- $TAG (git tag)
 end
 
 set -l committers_to_tag (mktemp)
+or exit 1
 set -l committers_from_tag (mktemp)
+or exit 1
 
 # You might think it would be better to case-insensitively sort/compare the names
 # to produce a more natural-looking list.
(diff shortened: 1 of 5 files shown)


mqudsi:
Why don't we just add a tests function called `mktemp`? We need to do that anyway to get the tests to pass on older macOS versions that don't support the same syntax that Linux and newer macOS do.

faho:
>Why don't we just add a tests function called mktemp?

We would still have to add `or return 1` everywhere, because the function won't return out of the calling function.

>We need to do that anyway to get the tests to pass on older macOS versions that don't support the same syntax that Linux and newer macOS do.

As far as I'm aware, all these uses work on macOS and linux - the issue was stuff like too short templates or using templates with *files* instead of directories and such.

ericonr:
I don't really have a horse in this race, I'm only extending the patch from my previous PR :)

So I'm happy to proceed with whatever's decided.

faho:
Okay, I merged this because it's a good thing by itself. Making a compatibility function can be done later, and isn't immediately necessary as far as I can tell.

mqudsi:
Agreed. Thanks, all!

PR #7495, opened 2020-11-22
string: Allow `string match --max`

faho:
This is useful in some cases, it also allows faster processing.

As a bonus, also stop reading strings if ``--quiet`` is given - no use
in reading more than we can.

## TODOs:

### DIFF
--- a/doc_src/cmds/string-match.rst
+++ b/doc_src/cmds/string-match.rst
@@ -8,7 +8,7 @@ Synopsis
 
 ::
 
-    string match [(-a | --all)] [(-e | --entire)] [(-i | --ignore-case)] [(-r | --regex)] [(-n | --index)] [(-q | --quiet)] [(-v | --invert)] PATTERN [STRING...]
+    string match [(-a | --all)] [(-e | --entire)] [(-i | --ignore-case)] [(-m | --max) MAX] [(-r | --regex)] [(-n | --index)] [(-q | --quiet)] [(-v | --invert)] PATTERN [STRING...]
 
 .. END SYNOPSIS
 
@@ -28,6 +28,10 @@ If ``--index`` or ``-n`` is given, each match is reported as a 1-based start pos
 If ``--regex`` or ``-r`` is given, PATTERN is interpreted as a Perl-compatible regular expression, which does not have to match the entire STRING. For a regular expression containing capturing groups, multiple items will be reported for each match, one for the entire match and one for each capturing group. With this, only the matching part of the STRING will be reported, unless ``--entire`` is given.
 
 If ``--invert`` or ``-v`` is used the selected lines will be only those which do not match the given glob pattern or regular expression.
+
+If ``--max`` or ``-m`` is used ``string match`` will stop after MAX matching STRINGs - together with ``--all`` it will still print all the matches for those strings.
+
+If ``-q`` or ``--quiet`` is used ``string match`` will only return the exit status without output. It will also stop processing after the first match.
 
 Exit status: 0 if at least one match was found, or 1 otherwise.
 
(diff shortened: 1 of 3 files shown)


faho:
One notable decision: This stops after X *matching arguments*. If `--all` is used it can therefore write more than X matches.

We could change that, but it would require handing the maximum off to the matcher, so it doesn't go over the limit.

faho:
Also:

This changes the semantics when `string` isn't the only thing with access to that input. Now, theoretically you could do

```fish
thing | while string match; read; end
```

and have the `read` read something. However, because the `string` might read too far (remember how `read` has to read byte-by-byte?) it's not guaranteed to be anything sensible.

The only real possibility is to still read everything to drain the pipe, so it doesn't _look_ like you could read anything.

This is mostly theoretical because the `string` won't read from stdin in that case (we're requiring that stdin is *directly* redirected), so it's hard to trigger, but if we added a `--stdin` flag this would make it slightly more awkward.

krobelus:
I wonder how hard it would be to add lazy evaluation, so `echo (seq 1000000 | string match 1)[1]` would implicitly stop after the first match.

mqudsi:
Can you share some sample workflows that would benefit from this? The complexities to the overall state and the discrepancies between matching against arguments vs matching against number of matches are unfortunate.

In particular, does this enable something (realistically speaking) that using index ranges doesn't currently allow? I have a few fish scripts that do something like `set -l matches (string match ...)[1..4]` to effectively limit to just 4 matches. Like you say (except for krobeleus' creative suggestion of allowing lazy evaluation) there isn't going to be a huge performance win here for most cases.

jorgebucaran:
I ran into this the other day and asked @faho if there was a way to stop after the first match, but I'm using `string match --regex ... <file | read line` now. I could have used index ranges as well. I agree that there isn't going to be a huge performance win like I thought, at least not for my use-case. @krobelus suggestion is really interesting too.

krobelus:
Yeah I think this should be solved in a more orthogonal way. A new option doesn't seem worth it. The `--quiet` optimization can still be done though.

> `string match --regex ... <file | read line`

I believe that this also reads the entire file, because builtin `string`'s output is buffered if the output is a pipe.
This could possibly be enhanced in future, see https://github.com/fish-shell/fish-shell/blob//aa0bfa0eb84689aceb4924ea9cd8abd5c983ea1b/src/exec.cpp#L454

So I'd say use something like `sed` if you want to only read parts of a file.
In future maybe `read` could be made faster.

faho:
> In future maybe read could be made faster.

`read`s fundamental problem is that it can't seek in pipes, so it *has to* read character-by-character to ensure it doesn't read too much (because it can't put the data back), and that's quite slow.

I don't see a way around that.

-----

Anyway, since it seems to be surprisingly unpopular I'm just going to commit early returns for `--quiet` for all applicable string subcommands.

faho:
For posterity: 75d1def718d1efb93f8641a12650a23d9854b7c7 made all string commands with `--quiet` quit as soon as possible.

The tests exercise this like `yes | string match -q y`, so it shows that you can use it with infinite streams.

mqudsi:
Doesn't that break some regex imports?

faho:
>Doesn't that break some regex imports?

Good question. I haven't been paying too much attention to that, I think it sets the variables to the last match? If so, this changes it so you get the *first* match instead.

faho:
Yup, it does indeed:

```fish
string match -rq '^(?<foo>.*)$' banana rama
```

sets $foo to banana, without  -q it's rama.

I'd probably just change the import semantics to pick the first match always - given that it's not released yet, and that I don't think there's a real difference between picking the first and the last. Also if we figured out how to make `while string` work it would be easier to work with.

mqudsi:
I didn’t even realize that breaks, I meant in conjunction with —all.

faho:
>I didn’t even realize that breaks, I meant in conjunction with —all.

It does *not* affect how many matches per-argument are done, so no.

Okay, I'm going to commit a change that makes it use the first matching argument's capture groups, but I'm going to hold off on the documentation changes until #7520 is done.

PR #7496, opened 2020-11-22
math: Add --base option

faho:
Currently a bit limited, unfortunately printf's `%a` specifier is
absolutely unreadable.

So we add `hex` and `octal` with `0x` and `0` prefixes respectively,
and also take a number but currently only allow 16 and 8.

The output is truncated to integer, so scale values other than 0 are
invalid and 0 is implied.

The docs mention this may change.

## TODOs:

### DIFF
--- a/doc_src/cmds/math.rst
+++ b/doc_src/cmds/math.rst
@@ -8,7 +8,7 @@ Synopsis
 
 ::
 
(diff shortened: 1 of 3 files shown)


ridiculousfish:
LGTM!

PR #7503, opened 2020-11-24
add completion for icdiff

hellow554:
icdiff is a diff command that uses colors by default and is an advances
version of the original diff command

https://github.com/jeffkaufman/icdiff


faho:
Merged, thanks!

PR #7516, opened 2020-11-29
Implement `d;` and `d,`

Aster89:
With this change I fix address one of the several issues raised in #4019 in a pretty obvious way: some bindings is missing and I'm adding two of them.

As discussed there, this is not the best way to face the inadequacy of the vi-mode that fish offers, but at the moment this is the most I can do (in terms of my time/knowledge/understanding of the existing code/programmer experience). I hope I'll be able to to something better in the future, beside possibly fixing similar issues. (For instance <kbd>y</kbd><kbd>;</kbd> and <kbd>y</kbd><kbd>,</kbd> should be easy to write too.)

## TODOs:

### DIFF
--- a/share/functions/fish_vi_key_bindings.fish
+++ b/share/functions/fish_vi_key_bindings.fish
@@ -166,6 +166,8 @@ function fish_vi_key_bindings --description 'vi-like key bindings for fish'
     bind -s --preset dl delete-char
     bind -s --preset di backward-jump-till and repeat-jump-reverse and begin-selection repeat-jump kill-selection end-selection
     bind -s --preset da backward-jump and repeat-jump-reverse and begin-selection repeat-jump kill-selection end-selection
+    bind -s --preset 'd;' begin-selection repeat-jump kill-selection end-selection
+    bind -s --preset 'd,' begin-selection repeat-jump-reverse kill-selection end-selection
 
     bind -s --preset -m insert s delete-char repaint-mode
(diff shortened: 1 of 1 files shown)


faho:
>With this change I fix issue #4019 in a pretty obvious way: the binding is missing and I'm adding it.

This does *not* fix issue #4019. It solves *a specific subset*, which is `d;` and `d,`.

The commit message mentions `dw`, which it doesn't touch in any way.

It's probably a good change regardless, but the commit message is flat out wrong and misleading. 

Unless, of course, you're planning on adding more to it.

Aster89:
Sorry, man, I'm trying to contribute in a useful way. I'll pay more attention next time when setting a title.

As regards other changes, I'll try to make some more in the future, but doing a bit now is a first step.

faho:
>Sorry, man, I'm trying to contribute in a useful way. I'll pay more attention next time when setting a title.

Oh, terribly sorry, that probably came across as needlessly mean.

I was just trying to stop people from thinking this would fix all of their worries regarding vi-mode, and in trying to get that across quickly was overzealous in my phrasing. Sorry!

>By the way, I've received an email about an unsuccessful job on macOS. I'll give a look at it when I can, ok?

No need, I already took a look. We recently switched to Github Actions and the tests are a bit flakey there (which is always an issue since we need to test interactivity, so we do have *some* timeouts, and common CI systems are *super* resource constrained - you often don't even get a keypress within 100ms).

I reran them and they passed.

----

As for the commit itself? Merging right now.

Issue #7519, opened 2020-11-30
Building sphinx-docs from source gives warning, "Inline emphasis start-string without end-string."


mojotx:
When building from source (3.1.2-1706-geb592092a) on macOS Big Sur, I see the following warnings with the `sphinx-docs` target:

```
FISH_BUILD_VERSION=3.1.2-1706-geb592092a
[  0%] Built target CHECK-FISH-BUILD-VERSION-FILE
Scanning dependencies of target fishlib
[  1%] Building CXX object CMakeFiles/fishlib.dir/src/fish_version.cpp.o
[  1%] Linking CXX static library libfishlib.a
[ 96%] Built target fishlib
[ 97%] Linking CXX executable fish_indent
[ 98%] Built target fish_indent
[100%] Building HTML documentation with Sphinx
../doc_src/cmds/string-match.rst:13: WARNING: Inline emphasis start-string without end-string.
../doc_src/cmds/string-match.rst:13: WARNING: Inline emphasis start-string without end-string.
/Users/mjarvis/src/github.com/fish-shell-mojotx/doc_src/cmds/string-match.rst:30: WARNING: Inline emphasis start-string without end-string.
/Users/mjarvis/src/github.com/fish-shell-mojotx/doc_src/cmds/string-match.rst:30: WARNING: Inline emphasis start-string without end-string.
[100%] Built target sphinx-docs
```

mqudsi:
That's on me; I tested with `pandoc` to verify that I could have italics ending within the middle of a word and was satisfied with the results.

If anyone can shed light how to write the following:

> some words then *n*th object

with only the first letter of "nth" being italicized.

charego:
@mqudsi some suggestions https://github.com/fish-shell/fish-shell/pull/7520#discussion_r533095541

The superscript one could be tidied up with some css I assume.

mojotx:
I like @zanchey's [idea](https://github.com/fish-shell/fish-shell/pull/7520#issuecomment-736252564) of simply rewording the text to eliminate the need of having a word partially italicized.

ridiculousfish:
This got fixed in eb22a9c4db1a2b7ffda030650ca00a393829b03b

PR #7520, opened 2020-11-30
Fix sphinx-docs warning

mojotx:
When building from source, there is a warning:

     ../doc_src/cmds/string-match.rst:13: WARNING: Inline emphasis start-string without end-string.

One fix appears to be putting a space after the emphasized 'n' character, e.g., `*n* th` instead of `*n*th`.

## Description

I'm not an expert with ReStructuredText, but this seems to be a reasonable way to silence the warning. 

Fixes issue #7519 

## TODOs:

- [ ] Changes to fish usage are reflected in user documentation/manpages.  _(N/A)_
- [ ] Tests have been added for regressions fixed. _(N/A)_
- [ ] User-visible changes noted in CHANGELOG.rst _(N/A)_


doc_src/cmds/string-match.rst
```diff
@@ -27,7 +27,7 @@ If ``--index`` or ``-n`` is given, each match is reported as a 1-based start pos
...
 
-When matching via regular expressions, it is possible to directly import matches as fish variables by means of named capture groups using the PCRE2 syntax. This behavior is automatic and occurs in addition to any other match reporting. The default behavior with ``--regex`` results in the initialization of fish variables in the default scope containing the matched text corresponding to each named capture group. A named capture group matching a zero-length string will be initialized as a fish variable containing a likewise empty string (i.e. the equivalent of ``""``), but a named capture group that did not match will result in an empty (null) fish variable. When ``--regex`` is used in conjunction with ``--all``, this behavior changes slightly: for each of the *n* matching sequences, the *n*th index of each named variable will contain the contents of the corresponding named capture group for the *n*th match, but will contain an empty string if the group was empty or if the group did not match.
```
charego, line 30: The literal space looks a little funky: `*n* th`

nth-spaces

Escaping looks slightly better to me: `*n*\ th`

nth-escapes

A superscript looks cool but messes up the line height:
```
*n*:superscript:`th`
```

nth-superscripts

zanchey:
What about rewording? "When ``--regex`` is used with ``--all``, this behavior changes. Each named variable will contain a list of matches, with the first match contained in the first element, the second match in the second, and so on. If the group was empty or did not match, the corresponding element will be an empty string."

mqudsi:
I'm good with @zanchey's suggestion, but would still like to know how the original could be done. As I said in the issue, `pandoc` renders it correctly (or at least, as intended).

charego:
@mqudsi https://github.com/fish-shell/fish-shell/pull/7520#discussion_r533095541 tl;dr: `*n*\ th`

mojotx:
I just pushed a change (3c11ea2a5a62720527e44e44c5addbe06a9e36fd), escaping the space.

charego:
I thought we agreed to reword it? You presented a nail and I had a hammer, but @zanchey's suggestion is better.

mojotx:
@charego: I replaced the text with @zanchey's suggestion, with commit ce32c6dae19f0e05782cad11d3809cadd06e5163

ridiculousfish:
Merged as eb22a9c4db1a2b7ffda030650ca00a393829b03b, thank you!

PR #7532, opened 2020-12-06
pactl completions: guard call to pacmd

ammgws:
Recently moved to pipewire via the `pipewire-pulse` package on Arch Linux. You still end up pulling in `libpulse` (which contains `pactl`), but it allows uninstalling the `pulseaudio` package which provides `pacmd`, and if you do so these completions break.

### DIFF
--- a/share/completions/pactl.fish
+++ b/share/completions/pactl.fish
@@ -40,15 +40,15 @@ end
 
 # This is needed to filter out loaded modules
 function __fish_pa_complete_unloaded_modules
-    # We need to get just the module names
-    set -l loaded (__fish_pa_print_type modules | string replace -r '^\w*\t([-\w]+).*' '$1')
     if command -sq pulseaudio
-        pulseaudio --dump-modules | while read -l name description
-            # This is a potential source of slowness, but on my system it's instantaneous
-            # with 73 modules
-            if not contains -- $name $loaded
-                printf "%s\t%s\n" $name $description
-            end
+        # We need to get just the module names
+        set -l loaded (__fish_pa_print_type modules | string replace -r '^\w*\t([-\w]+).*' '$1')
+            pulseaudio --dump-modules | while read -l name description
+                # This is a potential source of slowness, but on my system it's instantaneous
+                # with 73 modules
+                if not contains -- $name $loaded
+                    printf "%s\t%s\n" $name $description
+                end
         end
     end
 end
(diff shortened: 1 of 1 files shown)


share/completions/pactl.fish
```diff
@@ -7,7 +7,9 @@
 
 
 # HACK: This is the list of commands from pacmd - used so we can use complete -w there
-set -l commands (pacmd help | string match -r '^ +[-\w]+' | string trim)
+if command -sq pacmd
+    set commands (pacmd help | string match -r '^ +[-\w]+' | string trim)
+end
```
faho, line 12: Soooo.... if this isn't used, this should at least be setting $commands to $ctlcommands, or the subcommand detection should fail

ammgws: OK I just did that but realised there are other completions such as `load-module` that are broken too.

faho: Am I understanding correclty that really it's *just* `load-module`, because that checks `pulseaudio --dump-modules`, which won't be available in this system.

Not that I think loading modules for pulseaudio makes sense anyway because you're not running pulseaudio?

ammgws: It just happened to be the first random command I tested. I haven't tested anything else or looked inside all the other pulseaudio helper functions yet so there may be others.

Load-modules probably doesn't do anything when not running native pulseaudio, but fish should at least avoid spamming the terminal with errors if it can.

faho: Oh, sure, it shouldn't spam. But I'm not seeing any other commands failing, as they just call `pactl`.

ammgws: Just checked and that is indeed true.

share/completions/pactl.fish
```diff
@@ -67,7 +67,7 @@ complete -f -c pactl -n "not __fish_seen_subcommand_from $commands" -a remove-sa
 complete -f -c pactl -n "__fish_seen_subcommand_from play-sample remove-sample" -a '(__fish_pa_complete_type samples)'
 
 complete -f -c pactl -n "__fish_seen_subcommand_from unload-module" -a '(__fish_pa_complete_type modules)'
-complete -f -c pactl -n "__fish_seen_subcommand_from load-module" -a '(__fish_pa_complete_unloaded_modules)'
+complete -f -c pactl -n "__fish_seen_subcommand_from load-module; and command -sq pulseaudio" -a '(__fish_pa_complete_unloaded_modules)'
```
faho, line 70: This is the wrong place for that condition, it should be checked *inside* `__fish_pa_complete_unloaded_modules`.

At least conceptually, that could be called from multiple places. Also, the complet condition here should still apply even if no pulseaudio exists to let the `--no-files` take effect.

ammgws: Yeah sorry I guess I took the easy way out

faho:
Merged, thanks!

PR #7534, opened 2020-12-07
Creating cache for xbps-query in __fish_print_packages.fish

d9d6ka:
## Description

I use Void Linux on daily basis and I've been glad to see that completion rules for xbps utils were added to the repo. But I've also noticed that whether the cache-file is not found it's actually created but I have to retype the whole command to make the completion work.

This occurs only after the cache-file is deleted (or not created yet) but I dare offer you my minor patch.

Hope it would be useful. Fish is amazing!


faho:
Merged, thanks!

PR #7542, opened 2020-12-11
Break __fish_print_packages apart

faho:
## Description

This function makes no sense - there is very little reason to show apt packages for pacman completions. It only works most of the time because most people tend to only have one package manager installed, but since you can e.g. install pacman to create archlinux containers on a debian system, it can actually break.

This is incomplete, but it's a good first step.

## TODOs:


faho:
Okay, done, __fish_print_packages is now a dumb shell that just calls the real function, and isn't used by any of our completions anymore. I'm keeping it for the time being since third party completions might be calling it. We might want to call this out in the release notes.

Since this is a *bunch* of rote work, I'd appreciate it if people took a good look and ideally tested it on a variety of distributions.

You *can* just download __fish_print_packages.fish, the __fish_print_x_packages function and the completion script for your package manager and put them into ~/.config/fish/{functions,completions}, but be careful - it's fairly easy to screw up.

faho:
Alright, let's just merge this, since nobody complained, and we've had positive feedback from FreeBSD and I've checked on Arch and Debian.

(some days I wish we had a test squad that we could just ask to test everything)

PR #7544, opened 2020-12-12
completions: git switch only takes branch names

ammgws:
Trying to switch to a remote branch like "upstream/ver2" will error with "fatal: a branch is expected, got remote branch 'upstream/ver2'", so these completions should only print the branch name. There doesn't seem to be a function for printing just the branch names for remotes (branch names can have forward-slashes in them), so I have just left them out for now.


krobelus:
Thanks, great catch!

I think `__fish_git_unique_remote_branches` already covers remote branches, just fine.

PR #7547, opened 2020-12-13
__fish_complete_suffix: complete all files, but sort files with matching suffix first

krobelus:
## Description

This is a compromise between the common case where the user only wants files with a matching suffix,
and the exceptional case when there is no file extension, or one file we don't know about.

See #7040 and others.

## TODOs:


faho:
This looks like a great idea!

krobelus:
Many times when completing one of these commands, a single <kbd>Tab</kbd> already selects the only sensible option.
It's a bit sad to lose that.
Still, completing non-matching files seems like the least contentious solution.

PR #7550, opened 2020-12-14
6981 fix overlong descriptions

danieh65:
Simplify the overlong completion descriptions for some of the listed commands. Remove unnecessary context (which wouldn't be displayed anyway), and simplify overly verbose wording.

Work on issue #6981 
Fixes:
- `adduser.fish`
- `mocha.fish`
- `valgrind.fish`
- `wget.fish`


share/completions/adduser.fish
```diff
@@ -11,18 +11,18 @@ complete -c adduser -l disabled-login -d 'Do not run passwd to set the password'
 complete -c adduser -l disabled-password -d 'Like --disabled-login, but logins are still possible (for example using SSH RSA keys) but not using password authentication'
 complete -c adduser -l force-badname -d 'By default, user and group names are checked against the configurable regular expression NAME_REGEX (or NAME_REGEX if --system is specified) specified in the configuration file'
```
faho, line 12: These two are also very long - this one's a whopping 176 characters!

Maybe "Disable login with password authentication" and "Allow usernames that don't pass restrictions"?

share/completions/adduser.fish
```diff
@@ -11,18 +11,18 @@ complete -c adduser -l disabled-login -d 'Do not run passwd to set the password'
...
 complete -c adduser -l gecos -d 'Set the gecos field for the new entry generated' -r
-complete -c adduser -l gid -d 'When creating a group, this option forces the new groupid to be the given number' -r
+complete -c adduser -l gid -d 'When creating a group, force the groupid to be the given number' -r
 complete -c adduser -l group -d 'When combined with --system, a group with the same name and ID as the system user is created'
```
faho, line 15: While we're here, this one could be "Create a group for the user"?

share/completions/adduser.fish
```diff
@@ -11,18 +11,18 @@ complete -c adduser -l disabled-login -d 'Do not run passwd to set the password'
...
 complete -c adduser -l help -d 'Display brief instructions'
 complete -c adduser -l home -d 'Use specified directory as the user\'s home directory' -x -a '(__fish_complete_directories)'
 complete -c adduser -l shell -d 'Use shell as the user\'s login shell, rather than the default specified by the configuration file' -x -a '(cat /etc/shells)'
```
faho, line 18: "Use this shell as the user's login shell"

share/completions/adduser.fish
```diff
@@ -11,18 +11,18 @@ complete -c adduser -l disabled-login -d 'Do not run passwd to set the password'
...
-complete -c adduser -l ingroup -d 'Add the new user to GROUP instead of a usergroup or the default group defined by USERS_GID in the configuration file' -x -a '(cut -d : -f 1 /etc/group)'
+complete -c adduser -l ingroup -d 'Add the new user to GROUP instead of a usergroup or the default group' -x -a '(cut -d : -f 1 /etc/group)'
 complete -c adduser -l no-create-home -d 'Do not create the home directory, even if it doesni\'t exist'
```
faho, line 20: This includes a typo and could just be "Don't create the home directory"

faho (changes requested):
Some more suggestions - if you're not feeling up to adding them, I can add them later.

share/completions/adduser.fish
danieh65: Definitely too long! I've brought them down to 65 and 59 characters, respectively.

share/completions/adduser.fish
danieh65: Done. I omitted "for the user" because this option is the same as running `addgroup`.

share/completions/adduser.fish
danieh65: I've left "rather than the default" because it provides a little bit of extra context without making the line overlong (61 chars).

share/completions/adduser.fish
danieh65: Done.

danieh65 (commented):
I touched up the descriptions per @faho's requests.

faho:
Merged, thanks!

danieh65:
Thank you! Apologies if I didn't follow the proper change request workflow (this is my first open-source contribution).

faho:
Oh, no, it's quite alright. We're fairly low on bureaucracy, precisely to make contributions like this as painless as possible.

I don't think you really missed anything?

Basically, you do the code, a proper description and some semblance of usable commits, we do the rest - in this case I chose to squash everything into one commit because these aren't really usable on their own - I don't think we'd revert the valgrind one, and if we wanted to undo just it that would still be perfectly doable.

(I only had to solve a conflict with the wget one because I missed that you had already done it and committed it in the meantime)

danieh65:
That's great to hear, I'm a big fan of the project and have been using the shell for about three years now.

PR #7580, opened 2020-12-26
CHANGELOG: Add missing --query

IlanCosman:
Fabian forgot this in https://github.com/fish-shell/fish-shell/commit/a6674483c17f1676fffb003c7919c42ed465ce88 😜

Would be nice to have a way of skipping CI like we used to.


faho:
Oops, yeah, thanks!

PR #7586, opened 2020-12-28
cd first, ask questions later

### LINKED ISSUE #7577: cd relies on stat(3), which on buggy filesystems may fail even if a directory exists
The linux system call `stat()` (stat(3) in linux man page) can return < 0 when invoked over a file or directory that reside in a not properly  configured filesystem (observed in misconfigured permission on fuse).
In that case `stat()` fail even if a directory exists, so `cd`ing to those directories fails and file/folder autocompletion (and probably every time fish use `stat()` in the filesystem) fails as well.
As reported and tested on gitter, `access()` apparently works, as does actually changing the directory.
The following [branch](https://github.com/faho/fish-shell/tree/cd-error2) solves the `cd` problem, but there is still work to do to get autocompletion to properly work.

faho:
Some filesystems are broken and error out on `stat(3)` of existing and
cd-able directories.

So we just try to `fchdir` and report errors later.

Fixes #7577.

## TODOs:

I have *absolutely no idea* how to test this - we need some way to have a `stat` fail but an `fchdir` succeed.

### DIFF
--- a/src/builtin_cd.cpp
+++ b/src/builtin_cd.cpp
@@ -60,7 +60,7 @@ maybe_t<int> builtin_cd(parser_t &parser, io_streams_t &streams, wchar_t **argv)
     errno = 0;
     auto best_errno = errno;
 
-    for (auto &dir: dirs) {
+    for (const auto &dir: dirs) {
         wcstring norm_dir = normalize_path(dir);
 
         // We need to keep around the fd for this directory, in the parser.
@@ -69,8 +69,10 @@ maybe_t<int> builtin_cd(parser_t &parser, io_streams_t &streams, wchar_t **argv)
         bool success = dir_fd.valid() && fchdir(dir_fd.fd()) == 0;
 
         if (!success) {
-            // If the directory doesn't exist, we ignore it and only error if none does.
-            // ENOENT is a low-priority error - if we found something with the name that isn't a directory, we error out.
+            // Some errors we skip and only report if nothing worked.
+            // ENOENT in particular is very low priority
+            // - if in another directory there was a *file* by the correct name
+            // we prefer *that* error because it's more specific
             if (errno == ENOENT) {
                 if (!best_errno) best_errno = errno;
                 continue;
@@ -102,6 +104,7 @@ maybe_t<int> builtin_cd(parser_t &parser, io_streams_t &streams, wchar_t **argv)
     } else if (best_errno == EACCES) {
         streams.err.append_format(_(L"%ls: Permission denied: '%ls'\n"), cmd, dir_in.c_str());
     } else {
+        errno = best_errno;
         wperror(L"cd");
         streams.err.append_format(_(L"%ls: Unknown error trying to locate directory '%ls'\n"),
                                   cmd, dir_in.c_str());
--- a/src/builtin_cd.cpp
+++ b/src/builtin_cd.cpp
@@ -47,55 +47,69 @@ maybe_t<int> builtin_cd(parser_t &parser, io_streams_t &streams, wchar_t **argv)
     }
 
     wcstring pwd = parser.vars().get_pwd_slash();
-    maybe_t<wcstring> mdir = path_get_cdpath(dir_in, pwd, parser.vars());
-    if (!mdir) {
-        if (errno == ENOTDIR) {
-            streams.err.append_format(_(L"%ls: '%ls' is not a directory\n"), cmd, dir_in.c_str());
-        } else if (errno == ENOENT) {
-            streams.err.append_format(_(L"%ls: The directory '%ls' does not exist\n"), cmd,
-                                      dir_in.c_str());
-        } else if (errno == EROTTEN) {
-            streams.err.append_format(_(L"%ls: '%ls' is a rotten symlink\n"), cmd, dir_in.c_str());
-        } else {
-            streams.err.append_format(_(L"%ls: Unknown error trying to locate directory '%ls'\n"),
-                                      cmd, dir_in.c_str());
-        }
+    auto dirs = path_apply_cdpath(dir_in, pwd, parser.vars());
+    if (dirs.empty()) {
+        streams.err.append_format(_(L"%ls: The directory '%ls' does not exist\n"), cmd,
+                                  dir_in.c_str());
 
         if (!parser.is_interactive()) streams.err.append(parser.current_line());
 
         return STATUS_CMD_ERROR;
     }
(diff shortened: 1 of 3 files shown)


zanchey:
> I have _absolutely no idea_ how to test this - we need some way to have a `stat` fail but an `fchdir` succeed.

https://github.com/trailofbits/ebpfault is probably the way to go, but requires root and a newish kernel. Sounds like a fun thing to play with though.

zanchey:
See also https://www.scylladb.com/2016/02/16/fault-injection-filesystem-software-testing/

src/builtin_cd.cpp
```diff
@@ -47,55 +47,69 @@ maybe_t<int> builtin_cd(parser_t &parser, io_streams_t &streams, wchar_t **argv)
...
-
-        status = wstat(dir, &buffer);
-        if (!status && S_ISDIR(buffer.st_mode)) {
-            streams.err.append_format(_(L"%ls: Permission denied: '%ls'\n"), cmd, dir_in.c_str());
-        } else {
-            streams.err.append_format(_(L"%ls: '%ls' is not a directory\n"), cmd, dir_in.c_str());
+    errno = 0;
+    auto best_errno = errno;
+
+    for (auto &dir: dirs) {
```
krobelus, line 63: nit: this can probably be a `const auto &`

src/path.h
```diff
@@ -64,6 +64,10 @@ wcstring_list_t path_get_paths(const wcstring &cmd, const environment_t &vars);
 maybe_t<wcstring> path_get_cdpath(const wcstring &dir, const wcstring &wd,
                                   const environment_t &vars);
```
krobelus, line 65: This declaration should then be removed.

krobelus: oops, nevermind

src/builtin_cd.cpp
```diff
@@ -47,55 +47,69 @@ maybe_t<int> builtin_cd(parser_t &parser, io_streams_t &streams, wchar_t **argv)
...
+        wcstring norm_dir = normalize_path(dir);
+
+        // We need to keep around the fd for this directory, in the parser.
+        errno = 0;
+        autoclose_fd_t dir_fd(wopen_cloexec(norm_dir, O_RDONLY));
```
krobelus, line 68: I think we prefer `{` for initialization, although I'm probably guilty of using `(` as well.

src/builtin_cd.cpp
```diff
@@ -47,55 +47,69 @@ maybe_t<int> builtin_cd(parser_t &parser, io_streams_t &streams, wchar_t **argv)
...
+        bool success = dir_fd.valid() && fchdir(dir_fd.fd()) == 0;
+
+        if (!success) {
+            // If the directory doesn't exist, we ignore it and only error if none does.
+            // ENOENT is a low-priority error - if we found something with the name that isn't a directory, we error out.
```
krobelus, line 73: This refers to the ENOTDIR case, right? The ENOENT case is described in the line above.
Also the second sentence is a bit confusing because we don't error out immediately but still try other CDPATH entries but I don't know how to word it differently.

fabh2o:
I'm currently testing the branch, `cd` seems to works fine, `cd -` works too. Providing directly the name of the directory fails, tab completion fails too, but i know it's not the goal of this commit.
I have tried to get a reproducible non working example but with no chance.

krobelus:
> For the saga of broken filesystem, if the current directory . or parent directory .. files are missing, running .. in fish gets The file '..' is not executable by this user but cd .. works fine.

`path_get_cdpath` is also used by `path_as_implicit_cd` which decides that a command like `..` is a directory (that we should cd to). This check is also used during highlighting. It would probably be quite ugly to make this highlighting or directory completion work - I imagine one could cd there and back instead of using stat.

Does file/directory completion work in bash? It seems they also stat each completion candidate by default in `append_to_match` in `lib/readline/complete.c:1857`. It seems difficult to do much besides merely listing all files/directories if stat doesn't work. Many other programs will fail as well. The only real solution is to fix the filesystem.

krobelus:
> Providing directly the name of the directory fails

Do you mean that running something like `cd foo` fails when it shouldn't?

fabh2o:
> Do you mean that running something like `cd foo` fails when it shouldn't?

No, i mean just run `foo`, as stated in [fish cd doc ](https://fishshell.com/docs/current/cmds/cd.html), where `foo` is the name of a directory in the current folder.


> Does file/directory completion work in bash?

Yes, both of them

krobelus:
Ok, that's "implicit cd", the execution can be fixed by changing the call to `path_as_implicit_cd` in `src/parse_execution.cpp:864`.
Maybe we can reuse some of the new logic in builtin_cd.

I wonder how come completion works in bash. Do you have reproduction steps to configure the file system so we can test on that?

fabh2o:
> Do you have reproduction steps to configure the file system so we can test on that?

None yet, i opened a [thread](https://forum.rclone.org/t/stat-3-fails-on-mounted-folder/21254/2) on the rclone forum (it mounts a remote filesystem via fuse) but had no help so far

faho:
Note: I would be okay if we punted this to after 3.2 - I don't think it's important enough to block 3.2, given that it's a workaround for a broken filesystem that isn't used often, and that it's in a critical-ish path.

src/builtin_cd.cpp
```diff
@@ -47,55 +47,69 @@ maybe_t<int> builtin_cd(parser_t &parser, io_streams_t &streams, wchar_t **argv)
...
+        streams.err.append_format(_(L"%ls: '%ls' is not a directory\n"), cmd, dir_in.c_str());
+    } else if (best_errno == ENOENT) {
+        streams.err.append_format(_(L"%ls: The directory '%ls' does not exist\n"), cmd,
+                                  dir_in.c_str());
+    } else if (best_errno == EROTTEN) {
+        streams.err.append_format(_(L"%ls: '%ls' is a rotten symlink\n"), cmd, dir_in.c_str());
+    } else if (best_errno == EACCES) {
+        streams.err.append_format(_(L"%ls: Permission denied: '%ls'\n"), cmd, dir_in.c_str());
+    } else {
+        wperror(L"cd");
```
ridiculousfish, line 105: Probably should restore `errno` before calling wperror

ridiculousfish:
Left one review comment; otherwise changes LGTM

faho:
Moving this to 3.3. It should also be done for implicit cd, and it needs to be tested.

Issue #7582, opened 2020-12-27
Fish history file path detection is broken for path prefixed with `~`


Blizarre:
Version: 3.1.0
OS: Ubuntu focal
Terminal: st

Thank you for this project, I have been using fish for years and I love it. I looked at the issues but could not find one like this.

## How to reproduce

When I looked at my `~/.local/share/fish/fish_history` file, I found that files starting with `~`  weren't detected properly as paths:

```
$ md5sum fish.spec.in ~/test.sh /tmp/fish.desktop
5edd6151e8eee9c252fe08e77775d993  fish.spec.in
48a61d611ce2b7bce16a49119909e625  /home/simon/test.sh
4a104f9e1d24ad65fb4aca68aa4d96b8  /tmp/fish.desktop
$ tail -n 7 ~/.local/share/fish/fish_history
- cmd: md5sum fish.spec.in ~/test.sh /tmp/fish.desktop
  when: 1609108437
  paths:
    - fish.spec.in
    - /tmp/fish.desktop
- cmd: tail -n 7 ~/.local/share/fish/fish_history
  when: 1609108439

``` 

## Expectations

```
- cmd: md5sum fish.spec.in ~/test.sh /tmp/fish.desktop
  when: 1609108437
  paths:
    - fish.spec.in
    - ~/test.sh
    - /tmp/fish.desktop
- cmd: tail -n 7 ~/.local/share/fish/fish_history
  when: 1609108439
  paths:
    - ~/.local/share/fish/fish_history
```

## Comments

I do not know if it is really an issue or not, but it explain a few questions that I had in the past where I noticed that the autocomplete was inconsistent, and couldn't really understand why.

Blizarre:
After looking at the code, it seems that the function that decide if a path should be added is [path_is_valid](https://github.com/fish-shell/fish-shell/blob/e43913a547487e8fe57310b5fc90e93e922c7a6c/src/path.cpp#L441), and that it is not performing tilde expansion.

I had a quick look at [expand_tilde](https://github.com/fish-shell/fish-shell/blob/e43913a547487e8fe57310b5fc90e93e922c7a6c/src/expand.cpp#L786) and it requires an `environment_t` variable (only really used to extract `HOME`).

[...]

PR #7591, opened 2020-12-31
Expand ~ paths in the history files

Blizarre:
## Description

All paths prefixed with `~` were not detected as valid in the file history.
This change add a call to [expand_tilde](https://github.com/fish-shell/fish-shell/blob/e43913a547487e8fe57310b5fc90e93e922c7a6c/src/expand.cpp#L786) in `path_is_valid`.

Fixes issue #7582

## TODOs:

### DIFF
--- a/src/fish_tests.cpp
+++ b/src/fish_tests.cpp
@@ -4332,12 +4332,15 @@ void history_tests_t::test_history_path_detection() {
     vars.vars[L"HOME"] = tmpdir;
 
     history_t &history = history_t::history_with_name(L"path_detection");
-    history.add_pending_with_file_detection(L"cmd0 not/a/valid/path", tmpdir);
-    history.add_pending_with_file_detection(L"cmd1 " + filename, tmpdir);
-    history.add_pending_with_file_detection(L"cmd2 " + tmpdir + L"/" + filename, tmpdir);
+    history.add_pending_with_file_detection(L"cmd0 not/a/valid/path", vars);
+    history.add_pending_with_file_detection(L"cmd1 " + filename, vars);
+    history.add_pending_with_file_detection(L"cmd2 " + tmpdir + L"/" + filename, vars);
+    history.add_pending_with_file_detection(L"cmd2 ~/" + filename, vars);
+    history.add_pending_with_file_detection(L"cmd2 ~", vars);
+    history.add_pending_with_file_detection(L"cmd2 ~/invalid_path", vars);
     history.resolve_pending();
 
-    constexpr size_t hist_size = 3;
+    constexpr size_t hist_size = 6;
     if (history.size() != hist_size) {
         err(L"history has wrong size: %lu but expected %lu", (unsigned long)history.size(), (unsigned long)hist_size);
         history.clear();
@@ -4349,6 +4352,9 @@ void history_tests_t::test_history_path_detection() {
         {},
         {filename},
         {tmpdir + L"/" + filename},
+        {L"~/" + filename},
+        {L"~"},
+        {}
     };
 
     size_t lap;
--- a/src/highlight.cpp
+++ b/src/highlight.cpp
@@ -468,7 +468,7 @@ bool autosuggest_validate_from_history(const history_item_t &item,
     }
 
     // Did the historical command have arguments that look like paths, which aren't paths now?
-    if (!all_paths_are_valid(item.get_required_paths(), working_directory)) {
+    if (!all_paths_are_valid(item.get_required_paths(), ctx.vars)) {
         return false;
     }
 
--- a/src/history.cpp
+++ b/src/history.cpp
@@ -325,7 +325,7 @@ struct history_impl_t {
 
     // Add a new pending history item to the end, and then begin file detection on the items to
     // determine which arguments are paths
-    void add_pending_with_file_detection(const wcstring &str, const wcstring &working_dir_slash);
+    void add_pending_with_file_detection(const wcstring &str, const environment_t &vars);
 
     // Resolves any pending history items, so that they may be returned in history searches.
     void resolve_pending();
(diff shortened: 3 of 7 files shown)


Blizarre:
The main issue was that [the vars had to be passed by reference](https://github.com/fish-shell/fish-shell/blob/e43913a547487e8fe57310b5fc90e93e922c7a6c/src/history.cpp#L1325). I checked and the concrete implementation of the [`get`](https://github.com/fish-shell/fish-shell/blob/e43913a547487e8fe57310b5fc90e93e922c7a6c/src/env.cpp#L1269) seems to be [protected by a mutex](https://github.com/fish-shell/fish-shell/blob/e43913a547487e8fe57310b5fc90e93e922c7a6c/src/env.cpp#L1262), so it should be safe to pass a reference to envs to a backgroudn thread. 
@krobelus

Blizarre:
Lots of new tests on master, I will merge and restart the tests

Blizarre:
Failure on the [threadsan](https://github.com/fish-shell/fish-shell/pull/7591/checks?check_run_id=1630199069#step:6:159) tests, I will have a look later.

Blizarre:
From the issue @krobelus :
> Sounds good - probably you can pass an entire operation_context_t as returned by get_bg_context to add_pending_with_file_detection(). It is already used for highlighting and and autosuggestions; the file detection while writing history is a similar background task.

I could build a new `operation_context_t` for each call to [reader_data_t::handle_readline_command](https://github.com/fish-shell/fish-shell/blob/e43913a547487e8fe57310b5fc90e93e922c7a6c/src/reader.cpp#L2797), but I fear for the performance implications. `operation_context_t` seems to be used for long-lived data structures, not built for each call.

I don't know if I could build a `operation_context_t` and keep it somewhere?

krobelus:
Most code is not relevant in terms of performance.
`get_bg_context()` is already called twice on each readline command/keystroke - for highlighting and autosuggestions.
I added `FLOG(error, L"get_bg_context called");` to test that. I believe it's lightweight to construct a new context because only pointers and small objects are copied.

Also we only need to build this context when a command was executed.
A different cancel checker is probably also a good idea; `get_bg_context` returns a context that is cancelled on the next keystroke.
I don't know what the cancellation criteria for adding to history should be, maybe if `parser.libdata().exec_count` changes?
This means that there is time to add to history until the user runs the next command.

This change to support `~` looks safe to me, we can add support for other variables in a follow-up.

src/fish_tests.cpp
```diff
@@ -4332,12 +4332,15 @@ void history_tests_t::test_history_path_detection() {
     vars.vars[L"HOME"] = tmpdir;
 
     history_t &history = history_t::history_with_name(L"path_detection");
-    history.add_pending_with_file_detection(L"cmd0 not/a/valid/path", tmpdir);
-    history.add_pending_with_file_detection(L"cmd1 " + filename, tmpdir);
-    history.add_pending_with_file_detection(L"cmd2 " + tmpdir + L"/" + filename, tmpdir);
+    history.add_pending_with_file_detection(L"cmd0 not/a/valid/path", vars);
+    history.add_pending_with_file_detection(L"cmd1 " + filename, vars);
+    history.add_pending_with_file_detection(L"cmd2 " + tmpdir + L"/" + filename, vars);
+    history.add_pending_with_file_detection(L"cmd2 ~/" + filename, vars);
```
krobelus, line 4338: The tests look great, but let's continue the numbering as `cmd3` and so on

src/highlight.cpp
```diff
@@ -468,7 +468,7 @@ bool autosuggest_validate_from_history(const history_item_t &item,
     }
```
krobelus, line 469: > Fix: #7582

The commit message should rather say `Fixes #7582` or `Closes #7582`, otherwise the issue won't be closed automatically

src/history.cpp
```diff
@@ -1207,21 +1207,21 @@ wcstring history_session_id(const environment_t &vars) {
     return result;
 }
 
-path_list_t valid_paths(const path_list_t &paths, const wcstring &working_directory) {
+path_list_t valid_paths(const path_list_t &paths, const environment_t &envs) {
```
krobelus, line 1210: Why not call it `vars` everywhere?

krobelus (commented):
Looks fine - I only have some nitpicks. You can rebase on latest master, then the merge commit will vanish.

ridiculousfish:
The change looks quite nice, thank you! I didn't realize you were planning to make a PR to fix the issue, and so I implemented it in parallel in [this commit](https://github.com/ridiculousfish/fish-shell/commit/1dd85ed06a1ddf3ed7bbb93b7438d3867c0f3ef5). I can merge that, but if you prefer and address review comments, I'll merge yours instead so you get credit in the commit.

If you want to continue with this:

1. In the `add_pending_with_file_detection` call site, use `vars.snapshot()` function to get an immutable copy of the variables. This way there's not a race with `$PWD` and you can capture the shared_ptr by copy (I need to write up some internals docs).
2. Rather than detecting tildes manually, use `expand_string` (or `expand_one`) with the flags to control expansion. This centralizes our home directory handling (plus gives you variable expansion "for free.")
3. Pass along the operation context into `all_paths_are_valid`; that way it can check for cancellation after each path.

Blizarre:
> The change looks quite nice, thank you! I didn't realize you were planning to make a PR to fix the issue, and so I implemented it in parallel in [this commit](https://github.com/ridiculousfish/fish-shell/commit/1dd85ed06a1ddf3ed7bbb93b7438d3867c0f3ef5). I can merge that, but if you prefer and address review comments, I'll merge yours instead so you get credit in the commit.

That's alright! It was fun hacking a bit at a new code base. You can close this PR, your commit is a lot more complete and takes care of all the edge cases I couldn't possibly imagine.

PR #7597, opened 2020-12-31
Normalize and document __fish_status_to_signal

### LINKED ISSUE #7595: normalize and document __fish_status_to_signal
I think that this is a case exactly like `prompt_pwd`, where Fish can offer a convenient service for those creating a prompt. I [currently use](https://github.com/IlanCosman/tide/blob/main/functions/_tide_item_status.fish#L20) this undocumented function in my prompt, Tide (well technically i'm using `__fish_pipestatus_with_signal` because this is before 293a3a628d6d5d6220518e3991e989b38cf50c8e is released).

IlanCosman:
## Description

Converts `__fish_status_to_signal` to `fish_status_to_signal` for external use. Documents `fish_status_to_signal` and adds it to the changelog.

Fixes #7595

## TODOs:

### DIFF
--- a/doc_src/cmds/fish_status_to_signal.rst
+++ b/doc_src/cmds/fish_status_to_signal.rst
@@ -9,7 +9,7 @@ Synopsis
 ::
 
     function fish_prompt
-        echo -n (fish_status_to_signal $pipestatus) (prompt_pwd) '$ '
+        echo -n (fish_status_to_signal $pipestatus | string join '|') (prompt_pwd) '$ '
     end
 
 Description
--- a/doc_src/cmds/fish_status_to_signal.rst
+++ b/doc_src/cmds/fish_status_to_signal.rst
@@ -1,0 +1,29 @@
+.. _cmd-fish_status_to_signal:
+
+fish_status_to_signal - Convert exit codes to human-friendly signals
+====================================================================
+
+Synopsis
+--------
+
+::
+
+    function fish_prompt
+        echo -n (fish_status_to_signal $pipestatus) (prompt_pwd) '$ '
+    end
+
+Description
+-----------
+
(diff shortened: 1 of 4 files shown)


ridiculousfish:
Thank you! LGTM, if there are no other comments I will merge later today or tomorrow.

doc_src/cmds/fish_status_to_signal.rst
```diff
@@ -0,0 +1,29 @@
...
+
+    function fish_prompt
+        echo -n (fish_status_to_signal $pipestatus) (prompt_pwd) '$ '
+    end
+
+Description
+-----------
+
+``fish_status_to_signal`` converts exit codes to their corresponding human-friendly signals if one exists.
+This is likely to be useful for prompts in conjunction with the ``$status`` and ``$pipestatus`` variables.
```
faho, line 19: It's probably useful to note that this is a *guess* - there's nothing that enforces that e.g. status 129 has to be caused by a SIGHUP, a program could just decide to return it itself.

(AFAIK the mapping of signal to status is very common - it might even be guaranteed by the standard)

doc_src/cmds/fish_status_to_signal.rst
```diff
@@ -0,0 +1,29 @@
...
+fish_status_to_signal - Convert exit codes to human-friendly signals
+====================================================================
+
+Synopsis
+--------
+
+::
+
+    function fish_prompt
+        echo -n (fish_status_to_signal $pipestatus) (prompt_pwd) '$ '
```
faho, line 12: This prints:

```fish
SIGINT SIGINT ~ $ 
```

While that *might* be what you want, it might also be surprising, especially since $pipestatus doesn't tend to show as different from $status constantly, so people use this and then it shows up weird sometimes.

Maybe add a `string join` in here?

faho:
Merged, thanks!

PR #7605, opened 2021-01-04
use original theme repo URL

edouard-lopez:
## Description

> The new ayu theme points to https://github.com/edouard-lopez/ayu-theme.fish
> but it should link to the original Ayu theme https://github.com/dempfi/ayu


faho:
Thanks!

PR #7606, opened 2021-01-04
completions/pkg: Add support for "pkg check" sub-command

woodsb02:
Improve completion for the FreeBSD package management tool pkgng, by adding support for the "pkg check" sub-command.

Refer to the FreeBSD [pkg-check(8)](https://man.freebsd.org/pkg-check) manpage for details of the command syntax.


faho:
Merged, thanks!

PR #7609, opened 2021-01-05
(fish_hg_prompt) Added $fish_hg_executable setting

Mekk:
This can be useful in two scenarios:

a) Fish can be asked to use virtualenv- or user-installed mercurial
   when such is preferred to system installation:

     set -U fish_hg_executable ~/.virtualenvs/mercurial-3.8/bin/hg

b) … or simply system installation can be frozen by:

     set -U fish_hg_executable /usr/bin/hg

   (the latter avoids switching mercurial's in case user activates
   virtualenv containing another mercurial version)

By default `$fish_hg_executable` is simply `hg`, so evberything behaves as it used to.

### DIFF
--- a/doc_src/cmds/fish_hg_prompt.rst
+++ b/doc_src/cmds/fish_hg_prompt.rst
@@ -25,6 +25,8 @@ By default, only the current branch is shown because ``hg status`` can be slow o
 
 If you enabled the informative status, there are numerous customization options, which can be controlled with fish variables.
 
+- ``$fish_hg_executable`` specifies Mercurial executable used (``hg`` by default, can be set to absolute path like ``set -U fish_hg_executable /usr/bin/hg``)
+
 - ``$fish_color_hg_clean``, ``$fish_color_hg_modified`` and ``$fish_color_hg_dirty`` are colors used when the repository has the respective status.
 
 Some colors for status symbols:
--- a/share/functions/fish_hg_prompt.fish
+++ b/share/functions/fish_hg_prompt.fish
@@ -20,10 +20,12 @@ set -g fish_prompt_hg_status_unmerged !
 
 set -g fish_prompt_hg_status_order added modified copied deleted untracked unmerged
 
+set -g fish_hg_executable 'hg'
+
 function fish_hg_prompt --description 'Write out the hg prompt'
     # If hg isn't installed, there's nothing we can do
     # Return 1 so the calling prompt can deal with it
-    if not command -sq hg
+    if not command -sq $fish_hg_executable
         return 1
     end
 
@@ -45,7 +47,7 @@ function fish_hg_prompt --description 'Write out the hg prompt'
     echo -n '|'
 
     # Disabling color and pager is always a good idea.
-    set -l repo_status (hg status | string sub -l 2 | sort -u)
+    set -l repo_status ($fish_hg_executable status | string sub -l 2 | sort -u)
 
     # Show nice color for a clean repo
     if test -z "$repo_status"


faho:
Then why not make a function that points there?

```fish
function hg
    /usr/bin/hg $argv
end
```

That has the advantage of working for everything instead of requiring a specific variable just for hg.

Mekk:
Because there can be various hg around.

My simplest use case is `vf activate mercurial-5.6`  (let's say the latter is virtualenv where I just installed fresh mercurial version to be tested). Afterwards, whenever I issue `hg something`, I wanna test this version. But for the purpose of prompt I wanna stable system one (for example because I don't want to see errors about unavailable or broken extensions which are not yet installed for new version).

krobelus:
If you use `PATH=/bin:$PATH fish_vcs_prompt` to always prefer system binaries while computing the hg prompt, would that work?

> tested venv version reports numerous errors, at least initially – due to missing extensions and such

Is it possible to silence these errors (maybe by tweaking the hg prompt).I imagine the prompt wouldn't need many extensions.

Mekk:
Not sure how should I „use PATH…”. I faced the issue while experimenting with some custom prompt which internally used `fish_hg_prompt`, so I did not even see the function directly. Nor is this PATH always the same (on my machines sometimes I have „canonical” mercurial in `/usr/bin/hg`, sometimes in `/usr/local/bin/hg` and sometimes in `~/bin/hg`). Not too mention that in the very end, calling `/usr/bin/hg` is simply more natural and simpler than `PATH=/usr/bin hg`.

For comparison: I faced similar issue inside emacs. Simple
```
(setq monky-hg-executable "/usr/bin/hg")
```
and in 15 seconds I forgot about the problem.

It's standard approach: various tools allow one to specify paths to commands they use internally. As long as everything is OK, nobody cares. When conflict/problem arises, one can find the setting, and set it appropriately.

faho:
>Not sure how should I „use PATH…”.

The idea is, if you wanted to use e.g. /usr/bin/hg, wherever you call `fish_hg_prompt` or `fish_vcs_prompt` you'd change it to say

```fish
PATH=/usr/bin:$PATH fish_hg_prompt
```

instead. This puts /usr/bin first in $PATH *for that function*, so it will have to use /usr/bin/hg if that exists, no matter what the global $PATH says.


>For comparison: I faced similar issue inside emacs. Simple
>(setq monky-hg-executable "/usr/bin/hg")
>dropped to my .emacs, and in 15 seconds I forgot about the problem.
>It's standard approach: various tools allow one to specify paths to commands they use internally. 

An important note: Fish is not emacs. In particular fish heavily disagrees on how many options to expose, and this option is fundamentally a bit cheesy. For one it's just for hg, no other tool in fish has this. Why not for `git` and `svn` as well? And once you're there, why not for `cat` or `grep`, or more realistically `python`? Then it's a bit of a secondary layer - executables are normally found via $PATH, and this adds another bit on top of that, in this specific case.

We try very hard not to *need* options like this. So we either try to find a way to make it just work, to detect what needs to be done. If that doesn't work, it's better to make it scriptable - instead of a fixed option make it an open-ended thing that people can do more interesting things with (this is also why `fish_prompt` is a function and not a special string like $PS1).

krobelus:
> Not sure how should I „use PATH…”

My assumption was that you already use a custom `fish_prompt`, then modifying the call to `fish_vcs_prompt` is not a huge stretch. Otherwise you'd first need a quick `funcsave fish_prompt`.

I agree that using PATH is not the first thing that comes to mind. It has the advantage that it works without special support from the application, but the disadvantage that it could break other executables (which is extremely unlikely in this case).
In general, fish assumes that binaries like `git`, `hg`, `awk` and so on are sane. The fact that this is not the case here seems like an edge case (where a hacky solution is okay). Adding an option just for `hg` seems arbitrary as faho said. I'm not really familiar with `hg`, maybe this scenario is common for some reason? Then the option might be justified.

> Nor is this PATH always the same

That problem is orthogonal to the question of whether to customize the executable path or PATH.

Mekk:
`git` IMHO should have similar option, just like `svn` (I really faced cases where I had to configure where proper git lies - albeit not for fish - old machines with very old system-wide git). Still, it's less likely to strike - as I said, `hg` is particularly vulnerable because it easily interacts with virtualenvs.

Not sure whether/how fish uses python (I noticed `fish_config` but it is unlikely to be used once first day with fish passed).

Whatever. I faced the issue, fixed it for myself, and offered patch considering sb else can benefit. You dislike, your right, scratch it. Nice fish feature is that I can keep my version for myself.

Mekk:
And regarding `PATH` trickery. Standard fish functions are not always used directly, they are also „library”. Imagine tide switches from `fish_git_prompt` to `fish_vcs_prompt`. In such case user would have externally provided `fish_prompt` which calls two levels of standard functions where deepest one calls `hg`. How user is to impact?

Funny thing is that I can control variables to define various colors (say `fish_color_hg_clean`) but I can't control most important setting: the command to be called…

faho:
>Funny thing is that I can control variables to define various colors (say fish_color_hg_clean) but I can't control most important setting: the command to be called…

This is because the colors are purely a matter of taste and affect nothing. The command isn't, there's a correct one to call and 99.997% of the time it's the one in $PATH.

I believe the cost of carrying this option outweighs the benefits as I don't see this being a normal case (unless you're an hg developer, in which case you can surely modify your hg things).

If you really do need it modify fish_hg_prompt like you've already done - that's why it's a function, so you can change it. We do not have to foresee every customization point - this is what I meant earlier as  "make it scriptable".

Sorry, I am rejecting this as I don't think it belongs in fish itself.

krobelus:
> My simplest use case is vf activate mercurial-5.6

I tried `fish_hg_prompt` in such a virtualenv with `set fish_prompt_hg_show_informative_status 1`, didn't see a problem.

PR #7621, opened 2021-01-12
Add `losetup` completions

moverest:


faho:
Oh, it's "lo(op) setup". I was wondering why this tool would be called "lose tup" - wasn't "tup" that build system?

Anyway, merged, thanks!

moverest:
> I was wondering why this tool would be called "lose tup" - wasn't "tup" that build system?

I didn't know it existed. I've learned something today.

Issue #7491, opened 2020-11-20
Right prompt issues with reflowing terminal


TheEvilSkeleton:
> Please tell us which fish version you are using by executing the following:
> 
>   fish --version
>   echo $version

fish, version 3.1.2

> Please tell us if you tried fish without third-party customizations by executing this command and whether it affected the behavior you are reporting:
> 
>   sh -c 'env HOME=$(mktemp -d) fish'

```
Welcome to fish, the friendly interactive shell
Type `help` for instructions on how to use fish
```

I am only using [Tide](https://github.com/IlanCosman/tide) for customization. The maintainer told me it was an upstream issue with right prompts.

> Please tell us which operating system and terminal you are using. The output of `uname -a` and `echo $TERM` may be helpful in this regard although other commands might be relevant in your specific situation.

`uname -a`: `Linux TheMainDesktop 5.9.8-200.fc33.x86_64 #1 SMP Tue Nov 10 21:58:19 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux`

`echo $TERM`: `xterm-256color`

## Issue

Resizing the window of a terminal when using a right prompt will make the terminal look super ugly. The prompt will show up everywhere every time it gets resized.

![fish](https://user-images.githubusercontent.com/50847364/99743022-c5691980-2aa2-11eb-8c92-7bd7aa81e47b.gif)

Related issue: https://github.com/IlanCosman/tide/issues/48

IlanCosman:
Oh gosh, this is awkward 😅 This might not be a fish issue upon closer inspection. 

There _is_ the very similar upstream issue where if you have a multi-line left prompt and a right prompt, then window resizing does this.

Can demonstrate with:

```fish
function fish_prompt
    echo foo
    echo bar
end

function fish_right_prompt
    echo blah
end
```

But this bug doesn't appear if the right prompt isn't displaying anything, so it doesn't apply to your issue.

I'd love to get a fish maintainer's comments on this though, as I couldnt find the bug I described above in the open issues. I'll reopen the Tide issue and poke around on this. Sorry if I was mistaken in kicking this upstream 😅

IlanCosman:
Ok, yeah, this was a Tide issue (though caused by the upstream bug in my defense 😛).

I was setting the right prompt display to a space when I wanted it to not display anything. Space counts as a character obviously, thus producing the bug.

```fish
set -g _tide_fish_right_prompt_display ' '
```

Instead I'll set it to truly empty.

[...]

PR #7623, opened 2021-01-12
Make reflow handling configurable

faho:
## Description

As part of #7491, we turned off our own reflow for some terminals - VTE and alacritty, but declined to turn it off for iTerm (as that apparently behaves better with it on?).

Which shows that 1. it's an awkward list to keep, 2. some of these have it configurable (this PR adds Konsole to the list, which *definitely* has)

So what we do is to add a variable called `$fish_handle_reflow`. If that is set to 1, we will reflow. If not, we won't.

Then we move the list to *before* the winch handler and just set the variable default - if we detect it's a terminal that can do reflow, we turn ours off, otherwise we keep it turned on.

## TODOs:

----

As a personal note: I would really really really pretty please with a cherry on top love it if terminals managed to get something together to just *tell* us if this works, instead of us keeping weird lists and leaving it to the user to fix problems.

### DIFF
--- a/share/functions/__fish_config_interactive.fish
+++ b/share/functions/__fish_config_interactive.fish
@@ -241,24 +241,31 @@ function __fish_config_interactive -d "Initializations that should be performed 
         # __fish_enable_focus
     end
 
-    function __fish_winch_handler --on-signal WINCH -d "Repaint screen when window changes size"
+    # Detect whether the terminal reflows on its own
+    # If it does we shouldn't do it.
+    # Allow $fish_handle_reflow to override it.
+    if not set -q fish_handle_reflow
         # VTE reflows the text itself, so us doing it inevitably races against it.
         # Guidance from the VTE developers is to let them repaint.
         if set -q VTE_VERSION
-            return
-        end
-        # Same for alacritty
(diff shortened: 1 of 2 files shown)


zanchey:
This approach seems sensible to me.

share/functions/__fish_config_interactive.fish
```diff
@@ -241,17 +241,26 @@ function __fish_config_interactive -d "Initializations that should be performed
...
+        or string match -q -- 'alacritty*' $TERM
+        set -g fish_handle_reflow 0
+    else if set -q KONSOLE_VERSION
+        and test "$KONSOLE_VERSION" -ge 210400 2>/dev/null
+        # Konsole since version 21.04(.00)
+        # Note that this is optional, but since we have no way of detecting it
+        # we go with the default, which is true.
+        set -g fish_handle_reflow 0
+    else if not set -q fish_handle_reflow
+        set -g fish_handle_reflow 1
```
faho, line 257: Reminder for myself: This needs to not touch $fish_handle_reflow *at all* if it is set - or we'd override customization from config.fish.

PR #7626, opened 2021-01-13
Fix Git status in Acidhub prompt

henrikhermansen:
## Description

Fix 1: The --quiet flag must be at the end of the command. The way it was I would never get any status symbol in my prompt as the command failed.
Fix 2: After adding files to git, but before committing them, git status is unsorted. This gave me the output "M A M A" after `uniq`, which resulted in 4 status symbols instead of 2. Sorting them before filtering them fixed the problem.

Not sure if these changes needs to be reflected in docs or the changelog.

## TODOs:

### DIFF
--- a/share/tools/web_config/sample_prompts/acidhub.fish
+++ b/share/tools/web_config/sample_prompts/acidhub.fish
@@ -8,7 +8,7 @@ function fish_prompt -d "Write out the prompt"
     if set -l git_branch (command git symbolic-ref HEAD 2>/dev/null | string replace refs/heads/ '')
         set git_branch (set_color -o blue)"$git_branch"
         set -l git_status
-        if command git diff-index HEAD -- --quiet
+        if not command git diff-index --quiet HEAD --
             if set -l count (command git rev-list --count --left-right $upstream...HEAD 2>/dev/null)
                 echo $count | read -l ahead behind
                 if test "$ahead" -gt 0
--- a/share/tools/web_config/sample_prompts/acidhub.fish
+++ b/share/tools/web_config/sample_prompts/acidhub.fish
@@ -8,7 +8,7 @@ function fish_prompt -d "Write out the prompt"
     if set -l git_branch (command git symbolic-ref HEAD 2>/dev/null | string replace refs/heads/ '')
         set git_branch (set_color -o blue)"$git_branch"
         set -l git_status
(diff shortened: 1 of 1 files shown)


faho:
>Fix 1: The --quiet flag must be at the end of the command. The way it was I would never get any status symbol in my prompt as the command failed.

It... must not? Putting `--quiet` after `--` means it's not interpreted as an option by `git diff-index`, so it's interpreted as a "tree-ish" (a commit/branch/tag) instead. Which is 100% broken.

What you actually want is to prefix the command with `not` - `git diff-index --quiet` exits with a true status if there are no differences, but we want this branch to be taken if there *are* differences, so we need to turn it around.

>Fix 2: After adding files to git, but before committing them, git status is unsorted. This gave me the output "M A M A" after uniq, which resulted in 4 status symbols instead of 2. Sorting them before filtering them fixed the problem.

This seems correct.

henrikhermansen:
> It... must not? Putting `--quiet` after `--` means it's not interpreted as an option by `git diff-index`, so it's interpreted as a "tree-ish" (basically a path) instead. Which is 100% broken.

I read in an old SO post that it had to, and as it seemed to fix the issue for me, I didn't consider that information might be outdated. Sorry for not doing a more thorough check up on this.

> What you actually want is to prefix the command with `not` - `git diff-index --quiet` exits with a true status if there are no differences, but we want this branch to be taken if there _are_ differences, so we need to turn it around.

I changed it as you suggested, and this seems to work as expected 😃

faho:
>I read in an old SO post that it had to, and as it seemed to fix the issue for me, I didn't consider that information might be outdated. Sorry for not doing a more thorough check up on this.

I'm pretty sure that question was along the lines of "How do I tell diff-index to check a file called '--quiet'", while we want diff-index to *be* quiet, so we want to pass it as an option. (either that or the answer was just outright wrong - it happens)

It *seemed* to work because you don't have a file called "--quiet", so there is nothing to print, and without "--quiet" diff-index will just always return 0 if run in a git repo (because "--quiet" tells it both to not print output and to exit with 1 if there is a diff). It would however do the wrong thing if the git repo had no differences (because it would still return 0 and take the "there are differences" branch), and especially if it had a file called "--quiet" (because it might actually print the differences in that file!).

faho:
Merged, thanks!

PR #7638, opened 2021-01-19
completions/mtr: init

akiirui:
Add mtr completion.

[ci skip]

## Description

Talk about your changes here.

Fixes issue #

## TODOs:

### DIFF
--- a/share/completions/mtr.fish
+++ b/share/completions/mtr.fish
@@ -1,0 +1,41 @@
+complete -c mtr -x -a "(__fish_print_hostnames)"
+
+complete -c mtr -s h -l help -d 'display this help and exit'
+complete -c mtr -s v -l version -d 'output version information and exit'
+complete -c mtr -s 4 -d 'use IPv4 only'
+complete -c mtr -s 6 -d 'use IPv6 only'
+complete -c mtr -s F -l filename -d 'read hostname(s) from a file' -rF
+complete -c mtr -s r -l report -d 'output using report mode'
+complete -c mtr -s w -l report-wide -d 'output wide report'
+complete -c mtr -s x -l xml -d 'output xml'
+complete -c mtr -s t -l curses -d 'use curses terminal interface'
+complete -c mtr -l displaymode -d 'select initial display mode' -x -a '0 1 2'
+complete -c mtr -s g -l gtk -d 'use the GTK+ interface (if available)'
+complete -c mtr -s l -l raw -d 'output raw format'
+complete -c mtr -s C -l csv -d 'output comma separated values'
+complete -c mtr -s j -l json -d 'output json'
+complete -c mtr -s p -l split -d 'split output'
+complete -c mtr -s n -l no-dns -d 'do not resolve host names'
+complete -c mtr -s b -l show-ips -d 'show IP numbers and host names'
+complete -c mtr -s o -l order -d 'select output fields' -x
+complete -c mtr -s y -l ipinfo -d 'select IP information in output' -x -a '0 1 2 3 4'
+complete -c mtr -s z -l aslookup -d 'display AS number'
+complete -c mtr -s i -l interval -d 'ICMP echo request interval' -x
+complete -c mtr -s c -l report-cycles -d 'set the number of pings sent' -x
+complete -c mtr -s s -l psize -d 'set the packet size used for probing' -x
+complete -c mtr -s B -l bitpattern -d 'set bit pattern to use in payload' -x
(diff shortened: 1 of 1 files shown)


share/completions/mtr.fish
```diff
@@ -0,0 +1,41 @@
+complete -c mtr -x -a "(__fish_print_hostnames)"
```
faho, line 1: Probably good to mention that this is a network diagnostic tool.

https://www.bitwizard.nl/mtr/ seems to be the website?

share/completions/mtr.fish
```diff
@@ -0,0 +1,41 @@
...
+
+complete -c mtr -s h -l help -d 'display this help and exit'
```
faho, line 3: We generally capitalize the descriptions, and this isn't "this help", so it would just be

```
Display help and exit
```

share/completions/mtr.fish
```diff
@@ -0,0 +1,41 @@
...
+complete -c mtr -s v -l version -d 'output version information and exit'
+complete -c mtr -s 4 -d 'use IPv4 only'
+complete -c mtr -s 6 -d 'use IPv6 only'
+complete -c mtr -s F -l filename -d 'read hostname(s) from a file' -rF
+complete -c mtr -s r -l report -d 'output using report mode'
+complete -c mtr -s w -l report-wide -d 'output wide report'
```
faho, line 9: These should be explained. From [the man page](https://github.com/traviscross/mtr/blob/master/man/mtr.8.in), it seems like "wide report" mode means that the hostnames won't be abbreviated?

So something like "Output a statistic report" for `--report` and "Output report with full hostnames" for `--report-wide`?

share/completions/mtr.fish
```diff
@@ -0,0 +1,41 @@
...
+complete -c mtr -s x -l xml -d 'output xml'
+complete -c mtr -s t -l curses -d 'use curses terminal interface'
+complete -c mtr -l displaymode -d 'select initial display mode' -x -a '0 1 2'
```
faho, line 12: These would really benefit from an explanation.

From the man page I see

> 
Use this option to select the initial display mode: 0 (default)
selects statistics, 1 selects the stripchart without latency
information, and 2 selects the stripchart with latency
information.

So:

```fish
complete -c mtr -l displaymode -d 'select initial display mode' -x -a \
'0\t"Statistics" 1\t"Stripchart without latency" 2\t"Stripchart with latency"'
```

?

share/completions/mtr.fish
```diff
@@ -0,0 +1,41 @@
...
+complete -c mtr -s g -l gtk -d 'use the GTK+ interface (if available)'
+complete -c mtr -s l -l raw -d 'output raw format'
+complete -c mtr -s C -l csv -d 'output comma separated values'
+complete -c mtr -s j -l json -d 'output json'
+complete -c mtr -s p -l split -d 'split output'
+complete -c mtr -s n -l no-dns -d 'do not resolve host names'
+complete -c mtr -s b -l show-ips -d 'show IP numbers and host names'
+complete -c mtr -s o -l order -d 'select output fields' -x
+complete -c mtr -s y -l ipinfo -d 'select IP information in output' -x -a '0 1 2 3 4'
```
faho, line 21:
```
0 Display AS number
1 Display IP prefix
2 Display country code of the origin AS
3 Display RIR (ripencc, arin, ...)
4 Display the allocation date of the IP prefix
```

(tho possibly replace some of these terms with simpler ones - maybe write out "Autonomous System number", RIR -> "registry"?)

share/completions/mtr.fish
```diff
@@ -0,0 +1,41 @@
...
+complete -c mtr -s z -l aslookup -d 'display AS number'
+complete -c mtr -s i -l interval -d 'ICMP echo request interval' -x
```
faho, line 23: This interval is given in *seconds*, I'd add that to the description.

Also maybe add some numbers? `-a '(seq 1 20)'`? Makes it clear that this expects numbers.

(same goes for the other options that take numbers here)

faho (changes requested):
Some wording nitpicks

share/completions/mtr.fish
akiirui: @faho
But add `-a "(seq NUM NUM)"` may misleading users, let user think the available values are limited to this range.

akiirui: I will add description to no range limited arguments, and add `-a seq` to range limited arguments.

faho:
Merged, thank you very much!

PR #7642, opened 2021-01-21
Include completion for all pkg alias subcommands

xanderio:
## Description

FreeBSD pkg alias subcommands where missing from the completion list, this adds them in a way where new and user aliases are automatically include in the completion 

## TODOs:

### DIFF
--- a/share/completions/pkg.fish
+++ b/share/completions/pkg.fish
@@ -178,8 +178,6 @@ for alias in (pkg alias -lq)
     end
 end
 
-
-end
 function __fish_pkg_aliases
     for alias in (pkg alias -q)
         echo $alias | read -l name description
--- a/share/completions/pkg.fish
+++ b/share/completions/pkg.fish
@@ -165,3 +165,105 @@ complete -c pkg -n '__fish_pkg_is list' -xa '(pkg query "%n")'
 
 # update
 complete -c pkg -n '__fish_pkg_is add update' -s f -l force -d "Force a full download of a repository"
+
+# alias 
+for alias in (pkg alias -q)
+  set -l name (echo $alias | cut -f 1 -d' ')
+  set -l description (echo $alias | cut -f 2- -d' ')
+
+  switch $name
+  case 'all-depends'
+    set description 'Display all dependencies for a given package'
+    complete -c pkg -n "__fish_pkg_is $name" -xa '(pkg query "%n")'
+
+  case 'annotations'
+    set description 'Display any annotations added to the package'
+    complete -c pkg -n "__fish_pkg_is $name" -xa '(pkg query "%n")'
+
+  case 'build-depends'
+    set description 'Display build dependencies for a given package'
+    complete -c pkg -n "__fish_pkg_is $name" -xa '(pkg query "%n")'
+
+  case 'cinfo'
+    set description 'Display install package matching case-sensitve regex'
+    complete -c pkg -n "__fish_pkg_is $name" -xa '(pkg query "%n")'
+
+  case 'comment'
+    set description 'Display comment off a package'
+    complete -c pkg -n "__fish_pkg_is $name" -xa '(pkg query "%n")'
+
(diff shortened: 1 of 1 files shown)


share/completions/pkg.fish
```diff
@@ -165,3 +165,105 @@ complete -c pkg -n '__fish_pkg_is list' -xa '(pkg query "%n")'
 
 # update
 complete -c pkg -n '__fish_pkg_is add update' -s f -l force -d "Force a full download of a repository"
+
+# alias 
+for alias in (pkg alias -q)
+  set -l name (echo $alias | cut -f 1 -d' ')
+  set -l description (echo $alias | cut -f 2- -d' ')
```
faho, line 172: This is probably nicer as

```fish
echo $alias | read -l name description
```

share/completions/pkg.fish
```diff
@@ -165,3 +165,105 @@ complete -c pkg -n '__fish_pkg_is list' -xa '(pkg query "%n")'
...
+
+  switch $name
+  case 'all-depends'
+    set description 'Display all dependencies for a given package'
```
faho, line 176: We use four space indents - you could just run this through `fish_indent -w pkg.fish`

share/completions/pkg.fish
```diff
@@ -165,3 +165,105 @@ complete -c pkg -n '__fish_pkg_is list' -xa '(pkg query "%n")'
 
 # update
 complete -c pkg -n '__fish_pkg_is add update' -s f -l force -d "Force a full download of a repository"
+
+# alias 
+for alias in (pkg alias -q)
```
faho, line 170: This is executed when the completion file is loaded. That means it won't pick up new aliases (or throw away old ones).

If you wanted to make it dynamic, you'd have to put this into a function, print the things as "name\tdescription" (that's a tab character) and call it in `complete -a`, like

```fish
complete -c pkg -n __fish_pkg_subcommand -xa '(__fish_pkg_aliases)'
```

Of course this depends on if `pkg alias -q` is usably fast. If not, having these cached isn't too bad.

faho (changes requested):

share/completions/pkg.fish
```diff
@@ -165,3 +167,107 @@ complete -c pkg -n '__fish_pkg_is list' -xa '(pkg query "%n")'
...
+
+# alias 
+function __fish_pkg_aliases
+    for alias in (pkg alias -q)
+        echo $alias | read -l name description
+
+        switch $name
+            case all-depends
+                set description 'Display all dependencies for a given package'
+                complete -c pkg -n "__fish_pkg_is $name" -xa '(pkg query "%n")'
```
faho, line 179: These wouldn't be called if you never try completing the subcommand before completing its argument. And if you did it multiple times you'd set up the completion multiple times, which is a bit of a waste (from a quick check it seems fish will actually run it multiple times then - we don't seem to be checking for duplicates).

I'd probably just move these out of the function with the name hardcoded - there's no real harm to setting up this completion even if the alias doesn't exist, because in that case you couldn't call it anyway.

If you want to be absolutely correct, you'd have to check `pkg alias -q` again outside of this.

share/completions/pkg.fish
```diff
@@ -165,3 +167,107 @@ complete -c pkg -n '__fish_pkg_is list' -xa '(pkg query "%n")'
...
+            case size
+                set description 'Display the total size of files installed by a package'
+                complete -c pkg -n "__fish_pkg_is $name" -xa '(pkg query "%n")'
+
+            case '*'
+                set description "alias: $description"
+
+        end
+
+        complete -c pkg -n __fish_pkg_subcommand -xa $name -d "$description"
```
faho, line 270: This should now be:

```fish
printf '%s\t%s\n' $name $description
```

faho (changes requested):

faho:
Merged, thanks!

PR #7648, opened 2021-01-23
Allow profiling without startup time

faho:
## Description

Currently, profiling a command is a bit annoying, as you need to ignore a whole bunch of extraneous information related to fish's startup.

This changes it so `fish --profile` only writes profiling information from *after* startup - anything after reading the config and before executing the postconfig commands (I'm open to suggestions for a different cut-off point), and introduces a different `--profile-startup` option for profiling anything before.

## TODOs:

### DIFF
--- a/doc_src/cmds/fish.rst
+++ b/doc_src/cmds/fish.rst
@@ -31,7 +31,9 @@ The following options are available:
 
 - ``-n`` or ``--no-execute`` do not execute any commands, only perform syntax checking
 
-- ``-p`` or ``--profile=PROFILE_FILE`` when fish exits, output timing information on all executed commands to the specified file
+- ``-p`` or ``--profile=PROFILE_FILE`` when fish exits, output timing information on all executed commands to the specified file. This excludes time spent starting up and reading the configuration.
+
+- ``--profile-startup=PROFILE_FILE`` will write timing information for fish's startup to the specified file. This is useful to profile your configuration.
 
 - ``-P`` or ``--private`` enables :ref:`private mode <private-mode>`, so fish will not access old or store new history.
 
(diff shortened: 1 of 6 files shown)


zanchey:
Sounds like a good idea though I can't comment on the implementation.

faho:
Oh the implementation is simple enough.

It's the UI I'm not completely sure about - this changes `fish --profile` so that it'll only profile the commands you gave it.

So before `fish --profile /tmp/profile -ic 'fish_prompt'` would show the profile for fish_prompt *and* all the loading done in config.fish and such, so you'd have ~150 lines before you even got to anything.

Now, it only shows the time for `fish_prompt` (including all the calls that makes and all autoloading that does), and if you want to have the startup time you'd have to pass `--profile-startup /path/to/file`. And if you want *both*, you'd have to pass both and if you want the old style of having one file you'd pass the same file for both, so it's `fish --profile /tmp/profile --profile-startup /tmp/profile`. My thinking was that that was rare - you typically want to know how fast your fish starts up, so you'd use `--profile-startup` or how fast a function is, so you'd use `--profile`.

The alternative is to add a `--profile-mode` option that you then give an argument to explain what you want profiled - `--profile-mode=startup` or `--profile-mode=run` or `--profile-mode=both`. Make it default to `run`. But that makes the common case of just profiling startup more annoying.

Or keep `--profile` as it is now, add `--profile-startup` and `--profile-run` to only profile one of them. This has the advantage of being strictly backwards-compatible (the current solution does repurpose an option), but it continues to leave the naive option for the wrong behavior - so if someone doesn't read the manpage closely they'd get a more awkward profile than necessary if they just want to profile a function.

zanchey:
I'd change it, it's not a common use-case and it makes more sense this way.

PR #7649, opened 2021-01-23
Enable OSC 7 when running in WezTerm

wez:
Over in https://github.com/wez/wezterm/issues/115#issuecomment-765869705 one of my users noted that fish had automatic OSC 7, but that it wasn't enabled under WezTerm.

You can detect WezTerm through the `$TERM_PROGRAM` environment.  In practical terms, all versions of wezterm in the wild support OSC 7 so a version check is not needed.

I'm not a fish user myself, but I did give the equivalent change to this a try on my Fedora 33 machine (it has an older version of fish).

I can see in this file that there's some stuff with `__fish_enable_focus` that you may also want to enable under wezterm; the escape sequence is supported as are panes, tabs and windows.


faho:
Merged, thanks!

PR #7321, opened 2020-09-09
Some `complete` usability improvements

faho:
## Description

This fixes a few of `complete`s usage warts.

- It's weird to always have `-c command` as the first two arguments. This makes the `-c` optional, if there is one non-option argument it's used as the command
- `complete -c command` without any other options now lists just the (loaded) completions for that command
- Some improvements to the docs

## TODOs:

- [X] Changes to fish usage are reflected in user documentation/manpages.
- [X] Tests have been added for regressions fixed
- [ ] User-visible changes noted in CHANGELOG.rst

PR #7652, opened 2021-01-24
Complete custom "git-foo" commands from "git foo" commandlines

krobelus:
Since #7075, git-foo.fish files are sourced when Git completions are loaded.
However, at least Cobra (CLI framework for Go) provides completions like

	complete -c git-foo ...

This means that completions are only offered when typing "git-foo <TAB>"
and not on "git foo <TAB>". Fix this by forwarding the completion requests.

This patch breaks git-foo.fish files that directly add competions to "git" using

	complete -c git -n '__fish_git_using_command foo' ...

Since we forward to "git-foo" which has no custom completions, "git foo"
would get default file completions. I don't know if this pattern is used
already. Since #7075 is not released yet, it is probably not common.

cc @soraxas 

## TODOs:

### DIFF
--- a/share/completions/git.fish
+++ b/share/completions/git.fish
@@ -1899,6 +1899,16 @@ complete -c git -n '__fish_git_using_command bisect; and __fish_seen_argument --
 ## Custom commands (git-* commands installed in the PATH)
 complete -c git -n __fish_git_needs_command -a '(__fish_git_custom_commands)' -d 'Custom command'
 
+function __fish_git_complete_custom_command -a subcommand
+    set -l cmd (commandline -opc)
+    set -e cmd[1] # Drop "git".
+    set -l subcommand_args
+    if argparse -s (__fish_git_global_optspecs) -- $cmd
+        set subcommand_args $argv[2..] # Drop the subcommand.
+    end
+    complete -C "git-$subcommand $subcommand_args "(commandline -ct)
+end
+
 # Get the path to the generated completions
 # If $XDG_DATA_HOME is set, that's it, if not, it will be removed and ~/.local/share will remain.
 set -l generated_path $XDG_DATA_HOME/fish/generated_completions ~/.local/share/fish/generated_completions
@@ -1920,5 +1930,9 @@ for git_ext in $complete_dirs/git-*.fish
     contains -- $cmd $__fish_git_custom_commands_completion
     and continue
     source $git_ext
+    set -l subcommand (string replace -r '^git-' '' -- $cmd)
+    if [ (complete git-$subcommand | count) -gt 0 ]
+        complete git -f -n "__fish_git_using_command $subcommand" -a "(__fish_git_complete_custom_command $subcommand)"
+    end
     set -a __fish_git_custom_commands_completion $cmd
 end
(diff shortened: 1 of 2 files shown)


faho:
This seems like a good idea, particularly because it matches git's semantics - any command named "git-thing" is available as "git thing", but that means it can also be called as "git-thing". So a "git-thing.fish" *should* have `complete -c git-thing`!

soraxas:
I too reckon this is a good addition to make custom `git-foo` completions available when TAB completing `git foo`. 

About the 
```fish
complete -c git -n '__fish_git_using_command foo' ...
```
 pattern, they are used in projects such as 

https://github.com/tj/git-extras/blob/96b95392ac35d3ff033403e399cb7cfb7e94750b/etc/git-extras.fish#L80

For those sort of patterns, I think the new patch would add a completion that forwards `git foo` to `git-foo` that doesn't work, but since we are still sourcing the original file (`source $git_ext`)m the original completion commands should still work. (i.e. there will be one redundant completion command, but it shouldn't beak the existing functionality?)

krobelus:
> (i.e. there will be one redundant completion command, but it shouldn't beak the existing functionality?)

The redundant completion command would cause fish to use file completions on `git-extras `, since no completions are defined for `git-extras`. I'll guard against that by only doing this if `complete git-extras` is nonempty. Then both ways work.

soraxas:
I see that make sense, it didn't occured to me that `complete -C 'foo '` would provide the default filelist completions even if `foo` is not a valid command. 

Are there any simple way to check for the existence of completion command other than a guard that looks like this?
```fish
complete | string match -qr -- '\s-c git-foo'
```

share/completions/git.fish
```diff
@@ -1899,6 +1899,16 @@ complete -c git -n '__fish_git_using_command bisect; and __fish_seen_argument --
 ## Custom commands (git-* commands installed in the PATH)
 complete -c git -n __fish_git_needs_command -a '(__fish_git_custom_commands)' -d 'Custom command'
 
+function __fish_git_complete_custom_command -a subcommand
+    set -l cmd (commandline -opc)
+    set -e cmd[1] # Drop "git".
+    set -l subcommand_args
+    if argparse -s (__fish_git_global_optspecs) -- $cmd
+        set subcommand_args $argv[2..] # Drop the subcommand.
```
soraxas, line 1907: I think this should be `$argv[2..-1]`

share/completions/git.fish
```diff
@@ -1919,6 +1929,9 @@ for git_ext in $complete_dirs/git-*.fish
     # already sourced this git-* completion file from some other dir
     contains -- $cmd $__fish_git_custom_commands_completion
     and continue
-    source $git_ext
     set -a __fish_git_custom_commands_completion $cmd
+
+    source $git_ext
+    set -l subcommand (string replace -r '^git-' '' -- $cmd)
+    complete git -f -n "__fish_git_using_command $subcommand" -a "(__fish_git_complete_custom_command $subcommand)"
```
soraxas, line 1936: This line is missing a `-c`
```fish
complete -c git -f -n ...
```

share/completions/git.fish
faho: #6574

share/completions/git.fish
faho: #7321

faho:
>Are there any simple way to check for the existence of completion command other than a guard that looks like this?

After loading, with #7321 it's

```fish
complete foo
```

(#7321 makes it so it filters by the command and the `-c` is optional)

PR #7662, opened 2021-01-28
fish_clipboard_paste: trim indentation when pasting multiple lines

krobelus:
## Description

When pasting a multiline command with indented blocks, extra indentation from
spaces, or tabs, is generally undesirable. Discard it unless the cursor is
inside quotes.

---

This is probably not *always* exactly what you wants, since sometimes you might paste multiline tokens that are not commands.
But it's surely better in the majority of all cases.

I also created this PR because I wasn't sure about `__fish_commandline_is_singlequoted`.
Is it okay to rename it? It seems that some people are using it already, but not many.
Maybe it should be `fish_commandline_is_quoted single` for the old behavior, and `fish_commandline_is_quoted single double` for "single or double", but the "print" variant is best for `fish_clipboard_paste`, so we don't need to call it twice.

## TODOs:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -278,6 +278,7 @@ New or improved bindings
 -  As mentioned above, new special input functions ``undo`` (Control+\_ or Control+Z) and ``redo`` (Alt-/) can be used to revert changes to the command line or the pager search field (:issue:`6570`).
 -  Control-Z is now available for binding (:issue:`7152`).
 -  Additionally, using the ``cancel`` special input function (bound to escape by default) right after fish picked an unambiguous completion will undo that (:issue:`7433`).
+- ``fish_clipboard_paste`` (Control+V) trims indentation from multiline commands because fish already indents (:issue:`7662`).
 -  Vi mode bindings now support ``dh``, ``dl``, ``c0``, ``cf``, ``ct``, ``cF``, ``cT``, ``ch``, ``cl``, ``y0``, ``ci``, ``ca``, ``yi``, ``ya``, ``di``, ``da``, ``d;``, ``d,``, ``o``, ``O`` and Control+left/right keys to navigate by word (:issue:`6648`, :issue:`6755`, :issue:`6769`, :issue:`7442`, :issue:`7516`).
 -  Vi mode bindings support ``~`` (tilde) to toggle the case of the selected character (:issue:`6908`).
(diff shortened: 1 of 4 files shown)


faho:
>I also created this PR because I wasn't sure about __fish_commandline_is_singlequoted.
Is it okay to rename it? It seems that some people are using it already, but not many.

It's a dunderscored function, which we technically don't give guarantees for (but tbh we have sooo many of these, some we even recommend in the docs, that that's not a fantastic policy in practice - maybe we should undunder some?), but it's also a very specialized thing.

I say go for it, if you think anyone is using it (my gut says probably not, but I have been wrong before) add a compatibility shim - see `__fish_git_prompt`.

-----

I have two questions here:

1. How well does this handle arbitrary pastes? Escaped things, quoted strings spanning multiple lines and such?
2. Wouldn't this be better solved by stripping it when doing the commandline insert? `fish_clipboard_paste` only does one `commandline -i`, so it should be possible to handle that as one bit and ignore any superfluous indentation. Or maybe actually run the indentation code on it? (I can see, of course, how this is quite a bit simpler)

krobelus:
There are some [occurrences](https://github.com/search?q=__fish_commandline_is_singlequoted&type=code) on GitHub. It's probably still acceptable to rename it.

> How well does this handle arbitrary pastes?

Hmm, sadly it doesn't handle quoted multiline strings. Pasting something like

```
echo '
    indented
```

drops the indent. This is fixable.

> 2. Wouldn't this be better solved by stripping it when doing the commandline insert?

Could be. Adding this to `commandline -i` can work better, because we could always detect if we are inside quotes.
I can't think of cases we would break but `commandline -i` ought to be reliable.

We could also use the logic from `__fish_commandline_is_singlequoted` on `$data` to avoid trimming the quoted string example.
I'll think about it.

A similar issue is when I edit a multiline command with Alt-E and run `fish_indent` on it in the editor. When going back to the commandline, the indentation is doubled. Anyway, that's easily avoidable.

krobelus:
It might be a good idea to un-dunder functions that are likely to be copied by users (for example because they sample  prompts). However, that will just add to the functions we'd want to un-dunder/document/keep supporting. So I'm not sure.

I'm still not entirely happy with the name, maybe it should stay dundered.

I'm printing a deprecation message plus stack trace when the old function is loaded, but keep the behavior. The same should be usable for other deprecated wrappers. I still need to test how this will look like when called from bindings.

PR #7668, opened 2021-01-30
completions/networkctl: add missing commands

akiirui:
## Description

Add missing commands for `networkctl`

Fixes issue #

## TODOs:

### DIFF
--- a/share/completions/networkctl.fish
+++ b/share/completions/networkctl.fish
@@ -1,5 +1,30 @@
-set -l cmds status list lldp
+# Completions for networkctl (a part of systemd)
 
-complete -c networkctl -f -n '__fish_seen_subcommand_from status' -a '(networkctl list --no-pager --no-legend -a | string trim \
-| string replace -r \'([0-9]+) (\w+) .*$\' \'$2\t$1\n$1\t$2\')'
-complete -c networkctl -x -n "not __fish_seen_subcommand_from $cmds" -a "$cmds"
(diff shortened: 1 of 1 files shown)


share/completions/networkctl.fish
```diff
@@ -1,5 +1,30 @@
-set -l cmds status list lldp
+# Completions for networkctl (a part of systemd)
 
-complete -c networkctl -f -n '__fish_seen_subcommand_from status' -a '(networkctl list --no-pager --no-legend -a | string trim \
-| string replace -r \'([0-9]+) (\w+) .*$\' \'$2\t$1\n$1\t$2\')'
-complete -c networkctl -x -n "not __fish_seen_subcommand_from $cmds" -a "$cmds"
+set -l command_with_devices list status lldp delete up down renew forcerenew reconfigure
+set -l commands label reload $command_with_devices
+set -l devices (networkctl list --no-legend 2>/dev/null | string trim | string split -f 2 -- ' ')
```
faho, line 5: This executes once when the completions are *loaded*. Given that e.g. USB network cards are a thing, I'd prefer if this was done at completion time.

Because of how completions work, it would be enough to just quote it:

```fish
set -l devices "(networkctl list --no-legend 2>/dev/null | string trim | string split -f 2 -- ' ')"
```

akiirui: Right! fixed

PR #7670, opened 2021-01-30
Ensure escaped trailing spaces are not trimmed

### LINKED ISSUE #7661: Selecting previous command forgets trailing escaped space
## Problem and Reproduction
Take the following command (where there is a space after `\`, making it escaped)
```sh
echo \ 
```
The command will execute properly. However, when pressing up to re-use the previous command, the trailing space is removed. While removing trailing spaces is usually a good idea, I don't think this should be the case when the last space is escaped.

As this __seems__ like a pretty simple fix, I'd be happy to try my hand at fixing -- assuming this is indeed a bug and someone doesn't beat me to it!

## System Info
Fish version: `3.1.2`
OS: Arch Linux
Present when customizations are disabled

Shizcow:
## Description

When a command has trailing spaces (eg `ls       `) ensure that _escaped_ spaces are not trimmed.

Fixes #7661

## TODOs:
Unsure if this is notable enough for any of the following, left uncompleted:

### DIFF
--- a/src/reader.cpp
+++ b/src/reader.cpp
@@ -3226,8 +3226,29 @@ void reader_data_t::handle_readline_command(readline_cmd_t c, readline_loop_stat
                 }
 
                 // Historical behavior is to trim trailing spaces.
+                // However, escaped spaces ('\ ') should not be trimmed (#7661)
                 while (!text.empty() && text.back() == L' ') {
-                    text.pop_back();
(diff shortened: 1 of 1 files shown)


ridiculousfish:
Merged as cff5aa9130f650a2a497296942972b9922f60cc1. Thank you, that was quick!

Issue #7683, opened 2021-02-05
Syntax warning with argparse.rst


mojotx:
There's a syntax warning about a change introduced with 3eef2959908f6df2a1e7f17fe2707c2b36398ac2. 

```
~/src/fish-shell/doc_src/cmds/argparse.rst:103: WARNING: Literal block ends without a blank line; unexpected unindent.
~/src/fish-shell/doc_src/cmds/argparse.rst:103: WARNING: Literal block ends without a blank line; unexpected unindent.
```

PR #7684, opened 2021-02-05
Fix sphinx doc warning

mojotx:
~/src/fish-shell/doc_src/cmds/argparse.rst:103: WARNING: Literal block ends without a blank line; unexpected unindent.
~/src/fish-shell/doc_src/cmds/argparse.rst:103: WARNING: Literal block ends without a blank line; unexpected unindent.

## Description

Talk about your changes here.

Fixes issue #7683

## TODOs:

- [NA] Changes to fish usage are reflected in user documentation/manpages.
- [NA] Tests have been added for regressions fixed
- [NA] User-visible changes noted in CHANGELOG.rst

### DIFF
--- a/doc_src/cmds/argparse.rst
+++ b/doc_src/cmds/argparse.rst
@@ -100,6 +100,7 @@ Sometimes commands take numbers directly as options, like ``foo -55``. To allow 
 The ``#`` must follow the short flag letter (if any), and other modifiers like ``=`` are not allowed, except for ``-`` (for backwards compatibility)::
 
   m#maximum
+
 This does not read numbers given as ``+NNN``, only those that look like flags - ``-NNN``.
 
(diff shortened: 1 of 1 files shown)


faho:
Merged, thanks!

I think I mentioned this before but you don't need to open an issue *and* a PR, especially not for trivial things like this.

If there's actually going to be discussion on the issue, then open one, but this is just a clean bug that the PR is enough for, the issue is kinda useless.

mojotx:
> Merged, thanks!
> 
> I think I mentioned this before but you don't need to open an issue _and_ a PR, especially not for trivial things like this.
> 
> If there's actually going to be discussion on the issue, then open one, but this is just a clean bug that the PR is enough for, the issue is kinda useless.

Thanks, will do!

PR #7689, opened 2021-02-07
Highlight keywords differently

### LINKED ISSUE #7678: Give syntax commands their own colorization
In "real" languages, syntax elements like `if`, `for`, `while`, etc., are given their own color. While I know (at least I think I do) that Fish's `if`, `end` etc. are really just commands, it would still be nice if they were given their own color to distinguish them from "regular" commands.

faho:
This introduces a new variable $fish_color_keyword that will be used
to highlight keywords. If it's not defined, we fall back on
$fish_color_command as before.

An issue here is that most of our keywords have this weird duality of
also being builtins *if* executed without an argument or with
`--help`.

This means that e.g.

    if

is highlighted as a command until you start typing

    if t

and then it turns keyword.

Fixes #7678.

## TODOs:

Also note that this does *not* add any colors by default (except for the documentation).

I'm not sure if we should merge this *now* or figure out some way to deal with the command/keyword duality and merge it *then*.

### DIFF
--- a/doc_src/fish_indent_lexer.py
+++ b/doc_src/fish_indent_lexer.py
@@ -34,6 +34,7 @@ ROLE_TO_TOKEN = {
     "normal": Name.Variable,
     "error": Generic.Error,
     "command": Name.Function,
+    "keyword": Keyword,
     "statement_terminator": Punctuation,
     "param": Name.Constant,
     "comment": Comment,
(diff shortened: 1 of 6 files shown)


ridiculousfish:
LGTM!

faho:
You know what? This breaks nothing, and it makes the highlighting in the help a little nicer.

So even tho it's not perfect, let's just merge it.

PR #7694, opened 2021-02-08
Indent line continuations

krobelus:
Similar to what fish_indent does. After typing "echo \" and hitting return,
the cursor will be indented.

A possible annoyance is that when you have multiple indented lines

	echo 1 \
	    2 \
	    3 \
	    4 \

If you remove lines in the middle with Control-k, the lines below
the deleted one will start jumping around, as they are disconnected
from and reconnected to "echo".

## TODOs:


ridiculousfish:
Looks great!

krobelus:
Thanks! I took another look, this should be good to go.

PR #7695, opened 2021-02-08
Highlight redirection target as valid if it contains to-be-defined var

### LINKED ISSUE #6652: ros command-line tools autocomplete fail
Please tell us which fish version you are using:

`fish --version # fish version 3.1.0`

Please tell us which operating system and terminal you are using. The output of `uname -a` and `echo $TERM` may be helpful in this regard although other commands might be relevant in your specific situation.

`uname -a # Linux yasha-laptop 5.0.0-37-generic #40~18.04.1-Ubuntu SMP Thu Nov 14 12:06:39 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux`
`echo $TERM # xterm-256color`

Please tell us if you tried fish without third-party customizations by executing this command and whether it affected the behavior you are reporting:
not affected

i try to autocomplete ros command-line tools. In bash i hit tab button twice and it show list of commands. for example:
`yasha@yasha-laptop:~$ rostopic #tab #tab`
`bw    echo  find  hz    info  list  pub   type `
But in fist i just got  list of directories
`yasha@yasha-laptop ~> rostopic #tab #tab`
`Arduino/    Desktop/    Downloads/  Pictures/  sketchbook/  Videos/ ...`

### LINKED ISSUE #6654: fish cannot detect would-be-valid variable
```fish
set tables one two three
for table in $tables
    echo foo > $table.sql
end
```

as `$table` is not a live variable at the time the loop is being composed, the redirect destination is highlighted in the error color.

This behavior is just a particular illustration of the underlying problem. I'm pretty sure the editor and the parser simply were not designed with any of this in mind, so any fixes would have to either be architectural or hacks. Or I could be wrong and this could be a cinch.

krobelus:
If a variable is undefined, but it looks like it will be defined by the
current command line, assume the user knows what they are doing.
This should cover most real-world occurrences.

Closes #6654

## TODOs:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -268,6 +268,7 @@ Interactive improvements
 -  ``bind`` now shows ``\x7f`` for the del key instead of a literal DEL character (:issue:`7631`)
 -  Paths containing variables or tilde expansion are only suggested when they are still valid (:issue:`7582`).
 -  Syntax highlighting can now color a command as invalid even if executed quickly (:issue:`5912`)
+-  Redirection targets are no longer highlighted as error if they contain variables which will likely be defined by the current commandline (:issue:`6654`).
 -  fish is now more resilient against broken terminal modes (:issue:`7133`, :issue:`4873`).
 -  fish handles being in control of the TTY without owning its own process group better, avoiding some hangs in special configurations (:issue:`7388`).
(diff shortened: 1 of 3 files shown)


krobelus:
Whoops, I had referenced the wrong issue :c

PR #7698, opened 2021-02-08
Saturate return value in builtin_set_query

ethulhu:
## Description

`builtin_set_query` returns the number of missing variables. Because the return value passed to the shell is an 8-bit unsigned integer, if the number of missing variables is a multiple of 256, prior to this PR it would overflow to 0.

This PR saturates the return value at 255 if there are more than 255 missing variables.

This issue also seems to apply to `functions --query` and probably others; this PR only fixes `set --query` for now, but could be expanded to fix others.

Currently, it is fixed with an ad-hoc `if`-statement, but it could be fixed with some shared utility function like `int saturate_error_count_to_255(int)`:

```c
/* In builtin_set.cpp */
int builtin_set_query(…) {
    …
    return saturate_error_count_to_255(retval);
}

/* In builtin_functions.cpp */
int builtin_functions_query(…) {
    …
    return saturate_error_count_to_255(retval);
}
```

## TODOs:

### DIFF
--- a/tests/checks/set.fish
+++ b/tests/checks/set.fish
@@ -682,4 +682,11 @@ echo $foo
 echo $bar
 #CHECK: 1 3
 
+
+# Test that `set -q` does not return 0 if there are 256 missing variables
+
+set -lq a(seq 1 256)
+echo $status
+#CHECK: 255
+
 true
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,7 @@ fish next-minor
 Notable improvements and fixes
 ------------------------------
 
+- **Saturating error count** for ``set --query``. Previously, 256 errors were reported at 0 errors.
 
 Syntax changes and new commands
 -------------------------------
--- a/doc_src/cmds/set.rst
+++ b/doc_src/cmds/set.rst
@@ -49,7 +49,7 @@ The following options are available:
 
 - ``-e`` or ``--erase`` causes the specified shell variables to be erased
 
-- ``-q`` or ``--query`` test if the specified variable names are defined. Does not output anything, but the builtins exit status is the number of variables specified that were not defined.
+- ``-q`` or ``--query`` test if the specified variable names are defined. Does not output anything, but the builtins exit status is the number of variables specified that were not defined, saturating at 255 if more than 255 variables are not defined.
 
 - ``-n`` or ``--names`` List only the names of all defined variables, not their value. The names are guaranteed to be sorted.
 
--- a/src/builtin_set.cpp
+++ b/src/builtin_set.cpp
@@ -565,6 +565,13 @@ static int builtin_set_query(const wchar_t *cmd, set_cmd_opts_t &opts, int argc,
         free(dest);
     }
 
+    // The return value is cast to an 8-bit unsigned integer,
+    // so saturate the error count to 255.
+    // Otherwise 256 errors is reported as 0 errors.
+    if (retval > 255) {
+        retval = 255;
+    }
+
     return retval;
 }
 


tests/checks/set.fish
```diff
@@ -682,4 +682,14 @@ echo $foo
...
+
+if set -q a1 a2 a3 a4 a5 a6 a7 a8 a9 a10 a11 a12 a13 a14 a15 a16 a17 a18 a19 a20 a21 a22 a23 a24 a25 a26 a27 a28 a29 a30 a31 a32 a33 a34 a35 a36 a37 a38 a39 a40 a41 a42 a43 a44 a45 a46 a47 a48 a49 a50 a51 a52 a53 a54 a55 a56 a57 a58 a59 a60 a61 a62 a63 a64 a65 a66 a67 a68 a69 a70 a71 a72 a73 a74 a75 a76 a77 a78 a79 a80 a81 a82 a83 a84 a85 a86 a87 a88 a89 a90 a91 a92 a93 a94 a95 a96 a97 a98 a99 a100 a101 a102 a103 a104 a105 a106 a107 a108 a109 a110 a111 a112 a113 a114 a115 a116 a117 a118 a119 a120 a121 a122 a123 a124 a125 a126 a127 a128 a129 a130 a131 a132 a133 a134 a135 a136 a137 a138 a139 a140 a141 a142 a143 a144 a145 a146 a147 a148 a149 a150 a151 a152 a153 a154 a155 a156 a157 a158 a159 a160 a161 a162 a163 a164 a165 a166 a167 a168 a169 a170 a171 a172 a173 a174 a175 a176 a177 a178 a179 a180 a181 a182 a183 a184 a185 a186 a187 a188 a189 a190 a191 a192 a193 a194 a195 a196 a197 a198 a199 a200 a201 a202 a203 a204 a205 a206 a207 a208 a209 a210 a211 a212 a213 a214 a215 a216 a217 a218 a219 a220 a221 a222 a223 a224 a225 a226 a227 a228 a229 a230 a231 a232 a233 a234 a235 a236 a237 a238 a239 a240 a241 a242 a243 a244 a245 a246 a247 a248 a249 a250 a251 a252 a253 a254 a255 a256
```
faho, line 688: This should just directly test that the $status is 255.

Also it's a bit shorter as `set -q a(seq 1 256)`

*and* it should check if these are local variables or you get spurious failures if e.g $a1 is defined for some reason, so

```fish
set -lq a(seq 1 256)
echo $status
# CHECK: 255
```

faho (changes requested):

faho:
> Currently, it is fixed with an ad-hoc if-statement, but it could be fixed with some shared utility function like int saturate_error_count_to_255(int):

Considering that the shared utility function would be about the same amount of code I don't think it helps.

Now, if the builtin return type could be some saturating one or something that *might* be of use, but tbh it still feels like overkill.

tests/checks/set.fish
ethulhu: Done.

faho:
Merged as 6dd6a57c606afa6673f7950364f8a76610e4106d (with the CHANGELOG changed to 3.2.0).

Thanks!

ethulhu:
> > Currently, it is fixed with an ad-hoc if-statement, but it could be fixed with some shared utility function like int saturate_error_count_to_255(int):
> 
> Considering that the shared utility function would be about the same amount of code I don't think it helps.
> 
> Now, if the builtin return type could be some saturating one or something that _might_ be of use, but tbh it still feels like overkill.

It could also be implemented higher up the callstack, e.g. in [`builtin_run()`](https://github.com/fish-shell/fish-shell/blob/42c75111c86882d4b227fe34512c93b0f1e234eb/src/builtin.cpp#L470) or [`proc_status_t`](https://github.com/fish-shell/fish-shell/blob/42c75111c86882d4b227fe34512c93b0f1e234eb/src/proc.h#L54), to saturate any return status higher than 255 from any function.

For example, putting the saturation check into `proc_status_t::from_exit_code(int)` passes all tests:

```c
    /// Construct directly from an exit code.
    static proc_status_t from_exit_code(int ret) {
        // Some paranoia.
        constexpr int zerocode = w_exitcode(0, 0);
        static_assert(WIFEXITED(zerocode), "Synthetic exit status not reported as exited");

        if (ret > 255) {
            ret = 255;
        }
        return proc_status_t(w_exitcode(ret, 0 /* sig */));
    }

```

krobelus:
I'd either move the saturation logic to `builtin_run` (because it only happens with builtins, not with external commands), or apply a similar fix to other callers like `functions -q`.

Maybe adding an `assert(ret < 256)` as well is worth it? Hard to tell.

krobelus:
Looking through the documented exit codes, I think we have four affected builtins:

- `set -q`
- `functions -q`
- `exit`
- `return`

`exit` and `return` should probably print an error when the argument is > 255.
I think we should change the behavior of `set -q` and `functions -q` to always exit with 0 or 1. I doubt that anyone is doing arithmetic on those exit codes?


Anyway, your change looks good and is maybe safer, so feel free to go for it.

ethulhu:
> `exit` and `return` should probably print an error when the argument is > 255.
> I think we should change the behavior of `set -q` and `functions -q` to always exit with 0 or 1. I doubt that anyone is doing arithmetic on those exit codes?
> 
> Anyway, your change looks good and is maybe safer, so feel free to go for it.

I agree that `exit` and `return` should probably raise an error on an invalid value, but also agree that saturating return values to 255 is more compatible with the previous behavior and less likely to cause mysterious errors in user code.

I've filled in the branch I linked earlier, and made it into PR #7702.

PR #7703, opened 2021-02-09
Updated metasploit completions

exploide:
## Description

Just a small update for metasploit completions.

Don't know whether I should target `master` or `Integration_3.2.0`. Initial support for these completions will be in 3.2.0 anyway, so this can probably be backported.

## TODOs:


krobelus:
Thanks; I think everything should go to master, and we cherry pick to the integration branch (or, if they are in sync, push the same to both if they, like I just did).

Issue #7704, opened 2021-02-10
Allow enabling flow control also inside the shell


Depau:
- fish, version 3.1.2
- Linux frigo 5.10.9-arch1-1 #1 SMP PREEMPT Tue, 19 Jan 2021 22:06:06 +0000 x86_64 GNU/Linux
- `sh -c 'env HOME=$(mktemp -d) fish'` - yep, it still does it

When running on a serial console which I explicitly configured with `ixon` and `ixoff`, fish keeps disabling said options every time a new prompt is printed.

I understand that users might think their system is locked up if they press Ctrl+S by mistake and the terminal stops responding, however there are cases in which one may need software flow control.

Is it possible to disable this behavior?

### Reproducing

- Run fish in one terminal
- Run `tty` to find the tty/pty it's attached to
- In another terminal, run `sudo stty -F <tty found above> ixon ixoff` to enable software flow control
- Run `sudo stty -a -F <tty found above>` to verify that `ixon` and `ixoff` have indeed been enabled
- Go back to the initial fish terminal and press enter
- Run `sudo stty -a -F <tty found above>` to verify that `ixon` and `ixoff` have been disabled by fish
- Do the same thing with another shell, such as bash, to verify that it respects the user's decision

(using another terminal makes it easier to demonstrate the issue since otherwise fish resets the terminal right after stty is run)

(to whoever thought "who uses software flow control intentionally in 2021 anyway?", I do :stuck_out_tongue_winking_eye:)

faho:
Duplicate of #2315

faho:
Short answer: The next release will sanitize the terminal modes instead of just resetting them all, this allows you to enable flow control. It remains off by default.

Depau:
(replying here because the other issue is locked)

Seeing that 3.2.0 looks close to being released according to the number of open issues in the GH milestone, I tested this on the latest git (2bab31a as I'm writing this) and it still disables ixon/ixoff, even if I set them before starting fish.

I see that it does restore the terminal to how it found it on exit, however my application needs flow control enabled \*while\* fish is running.

Does the new behavior you mentioned need to be enabled explicitly?

faho:
>I see that it does restore the terminal to how it found it on exit, however my application needs flow control enabled *while* fish is running.

Ah, okay. Yeah, that's not supported.

Like I said, fish sanitizes the modes still, and flow control doesn't seem very useful *inside the shell* - it's already waiting for input, why do you want it stopped?

[...]

PR #7708, opened 2021-02-11
Allow flow control also in the shell

faho:
## Description

This specifically copies IXON/IXOFF from the terminal modes after a command to the shell-modes, so you can use

    stty ixon ixoff

to enable flow control (and `stty -ixon -ixoff` to disable it again). This brings with it the obvious issues that you lose two keychords (ctrl-s and ctrl-q by default, typically) and that it might be confusing if you're not familiar with it (unless your terminal tells you what happened, like e.g. konsole does).

It's technically *possible* someone ends up with flow control enabled accidentally - this doesn't check if the command was `stty` after all - but I'm assuming that's highly unlikely, and even if it did the shell would still basically work.

It also still disables flow control on startup so you specifically have to ask for it - we don't want to give up ctrl-s.

Fixes issue #7704.

## TODOs:

Not sure how to test this - can we just send ctrl-s from pexpect? How do we then know it worked? Confirm that no output happened? Does this require waiting for a timeout?

### DIFF
--- a/build_tools/pexpect_helper.py
+++ b/build_tools/pexpect_helper.py
@@ -26,6 +26,7 @@ import pexpect
 # Default timeout for failing to match.
 TIMEOUT_SECS = 5
 
+UNEXPECTED_SUCCESS = object()
 
 def get_prompt_re(counter):
     """ Return a regular expression for matching a with a given prompt counter. """
@@ -74,6 +75,8 @@ def pexpect_error_type(err):
         return "EOF"
     elif isinstance(err, pexpect.TIMEOUT):
         return "timeout"
+    elif err is UNEXPECTED_SUCCESS:
+        return "unexpected success"
     else:
         return "unknown error"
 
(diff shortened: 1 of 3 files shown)


Depau:
> Not sure how to test this - can we just send ctrl-s from pexpect? How do we then know it worked? Confirm that no output happened? Does this require waiting for a timeout?

(I'm not familiar with how fish tests work) FWIW if you can run `tty` from inside fish and record the pts path, you can then run `stty -a -F /dev/pts/#` externally and verify whether the terminal mode is correct. Even easier if pexpect allows you to retrieve the pty slave name.

This will work on Linux for sure, I don't know about macOS.

Depau:
I confirm that the latest changes address my comment https://github.com/fish-shell/fish-shell/issues/7704#issuecomment-777755645

Thanks :)

src/builtin_bind.cpp
```diff
@@ -35,20 +35,6 @@ struct bind_cmd_opts_t {
...
-// A few builtins, including 'while', 'command' and 'builtin' are not defined here as they are
-// handled directly by the parser. (They are not parsed as commands, instead they only alter the
-// parser state)
-//
-// The builtins 'break' and 'continue' are so closely related that they share the same
-// implementation, namely 'builtin_break_continue.
-//
-// Several other builtins, including jobs, ulimit and set are so big that they have been given their
-// own module. These files are all named 'builtin_NAME.cpp', where NAME is the name of the builtin.
-
```
zanchey, line 51: Is this supposed to be removed?

faho: Yeah, this comment is wrong - it went in there when we split up the builtins.

The commit isn't *really* part of this PR, it's also on master as 056153eddd008233af27904c6ec54320da796882. I forgot to push it before I branched, it'll go away on a rebase.

Issue #7466, opened 2020-11-08
builtins set doesn't follow symlinks when saving variables


etu:
So when using `set --universal name value` it saves to `.config/fish/fish_variables`. If that file is a symlink it removes the symlink and just stores as a new file instead.

This for me is a problem because both my root partition and my home partition is on a tmpfs filesystem, so when this happens the change isn't persistent to any filesystem and gets lost on next reboot.

I've lost things more than once due to this, such as abbreviations that I've saved.

Recording: https://asciinema.org/a/Y3QcT1kfZcVC8tIv5I5quHy4p

System info:
```
$ fish --version
fish, version 3.1.2

$ echo $version
3.1.2

$ uname -a
Linux agrajag 5.9.2 #1-NixOS SMP Thu Oct 29 09:12:22 UTC 2020 x86_64 GNU/Linux

$ echo $TERM
xterm-256color     # (but it's really kitty)
```

I tried to run with a separate home directory:
```
$ sh -c 'env HOME=$(mktemp -d) fish'
```

And could perform the same actions to get the same problem.

More about my setup: https://elis.nu/blog/2020/05/nixos-tmpfs-as-root/ https://elis.nu/blog/2020/06/nixos-tmpfs-as-home/

faho:
You just have to symlink the *directory*, not the file.

We unlink it so we get atomicity, which simplifies things quite a bit.

etu:
Hmm, it's actually not that simple for me to change that... Because the symlinks go to different places. Even though it looks like they go to the sample place, that place just contains symlinks. The symlink for `config.fish` and `functions` goes to a read only place while the symlink for `fish_variables` goes to another symlink that points at a filesystem which is readwrite.

Wouldn't it be possible to resolve the symlinks first, then unlink and replace the file at it's storage location?

ridiculousfish:
Probably doable in principle, not trivial though.

mqudsi:
The easiest approach is probably to not unlink the file at all, and just truncate + write instead?

krobelus:
It needs to be atomic, so other fish always read a consistent state. Though that should already be guaranteed because we are locking the file.

zanchey:
Locking is too unreliable over NFS.

stouset:
Would the project be willing to accept a patch that checks if `${XDG_CONFIG_HOME}/fish/fish_variables` is a symlink, and—if so—performs the same atomic write but to the symlinked location instead?

[...]

Issue #7553, opened 2020-12-15
Feature: Allow to customize fish history file path


jansepke:
It would if the history file path would be customizable. My plan is to built a throw away dev VM with a fish shell. But I would like to not start with a empty history all the time because a filled shell history increases productivity ;-) So my idea is to mount the fish history file into my VM so everything is in sync. Would it be possible to add a config option for this? I already tried symlinks but that gets overridden by fish automatically.

❯ fish --version
fish, version 3.1.2

faho:
Already available - set $fish_history to an identifier for the "session", or empty to not keep history at all - see [the documentation on special variables](https://fishshell.com/docs/current/index.html#special-variables).

I can't say I agree with your idea, since a shell history also means you can search for commands you've already used, but you do you.

faho:
Fish's history path is $XDG_DATA_HOME/fish/"$fish_history"_history.

Set those two variables however you want.

jansepke:
As far as I know $XDG_DATA_HOME is also used by other applications that I maybe don't want to influence (e.g. fonts, nano, ...) and setting only $fish_history is not enough because I want to configure the path and not only the file name. Would be great if I could configure it only for fish shell.

faho:
In that case, symlink the fish/ *directory* in $XDG_DATA_HOME.

I don't think we have interest in making the full location customizable, just like we don't want config.fish's location to be customizable beyond $XDG_CONFIG_HOME.

The issue with symlinking the file is the same as with fish_variables - we want these things to be usable across shells, so we need to have atomicity guarantees, anything beyond moving a file into place is quite complicated (locking doesn't work etc), so I'll refer you to #7466 for that.

jansepke:
Ok, that does works as long as the target directory is not a shared folder between the windows host and the linux VM. But that is now a very special case so I will do a workaround with a cron job or similar.

PR #7728, opened 2021-02-18
Fix `set -U` when `fish_variables` is a symlink

ilyagr:
Previously, `set -U` would overwrite the symlink with a regular file.

Fixes https://github.com/fish-shell/fish-shell/issues/7466

This is a very naive implementation of the solution that the issue thread seemed to agree on. 

Can we rely on `wrealpath` always returning a valid path? Or should we fall back to `vars_path` if `wstat` returns something dodgy?

I wasn't sure which section of the CHANGELOG.rst this belongs it. I put it into "interactive improvements". But perhaps it's a "scripting improvement"? Or a "notable bugfix"?

## TODOs:


ilyagr:
If this looks OK, https://github.com/fish-shell/fish-shell/issues/7553 could be more fully resolved with essentially the same change at

https://github.com/fish-shell/fish-shell/blob/e8c9da100c438b1876fb04fade314e67f92f690d/src/history.cpp#L777

ridiculousfish:
Can't think of any reason not to do this, so merging. Thanks for writing a test!!

ilyagr:
Thank you!

Issue #7741, opened 2021-02-25
fish_config may not work in certain cases in WSL (2)


nosed1ve:
I am aware that the bug in which `fish_config` does not work in WSL 2 because of the fact that the string in `/proc/version`  is `microsoft` and not `Microsoft` is being fixed in the new release, however, in the function `runThing()` in `tools/webconfig/webconfig.py`, we are calling `cmd.exe` as it is, without mentioning it's full path.

Some people(like me) may have `appendWindowsPath = false` in their `/etc/wsl.conf` which turns of the adding of windows path elements to the WSL linux $PATH. If this is the case, `cmd.exe` would not be found in $PATH and hence `fish_config` would fail.

To fix this all that needs to be done is that we need to call `cmd.exe` by it's full path, which is `/mnt/c/Windows/System32/cmd.exe`.

There is one small edge case that might need to be considered however. That is that if the user changed their windows filesys mount location, it wouldn't be /mnt/c anymore, and might be something else, like /windir/c.

I am willing to make a pull request when I have time if required.

faho:
>There is one small edge case that might need to be considered however. That is that if the user changed their windows filesys mount location, it wouldn't be /mnt/c anymore, and might be something else, like /windir/c.

It seems like the best solution here is to add the default location to $PATH temporarily, so we get it either via the user's configured $PATH or the default location. It'll still fail if they changed the mount location (why would you do that, tho) *and* didn't add it to $PATH, and I don't see a way of supporting that.

>I am willing to make a pull request when I have time if required.

That would be great, yeah.

nosed1ve:
>  It'll still fail if they changed the mount location (why would you do that, tho) 

Some people do it for convenience, `/windir/c` makes more sense than `/mnt/c`, or some ppl change mount location to /, so the `c` directory becomes `/c`, which I believe is default on cygwin and is more convenient too.

Anyway, what I was thinking was, we can run a shell script, which finds the location of the `c` directory using a `find` command.
Then we can call `cmd.exe` using the `c` dir path we found from the `find`.

If that is too much, the $PATH solution seems best. Can we assume that the `c` directory PATH is first in `$PATH`? That way, we can get the `c` dir path in the `webconfig.py` script by doing something like `os.environ.get('PATH').split(':')[0]`.

> That would be great, yeah.
Sure thing, I'll do it as soon as I have some time on my hands.

faho:
>If that is too much

It is too much, yes.

[...]

PR #7742, opened 2021-02-25
Ensure web_config works on WSL

nosed1ve:
web_config could sometimes fail on WSL if the user chose not to append
windows directories to their linux $PATH. This change ensures that the
cmd.exe executable is found in most cases even if windows directories
are not appended to $PATH on linux.

An error message letting the user know that cmd.exe was not found, and
that they should add the cmd.exe dir to their $PATH before running
fish_config is displayed if cmd.exe is still not found.

## Description

To check for cmd.exe, I added a (constant) variable named `COMMON_WSL_CMD_PATHS` which is a tuple containing (at the moment) 3 common paths which may be the directories `cmd.exe` is in. Also, I modified the `find_executable()` function to take a `paths` argument which defaults to an empty tuple.

If the system is detected to be WSL, the `COMMON_WSL_CMD_PATHS` is passed as the `paths` argument to the `find_executable()` function, and `cmd.exe` is attempted to be found. If `cmd.exe` is still not found, an error message is printed and the program exits with a non-zero status code.

I tested this on my machine and `web_config` successfully launched the default Windows web browser on WSL 2 with these changes.

Fixes issue #7741

## TODOs:

Edit: added a description

### DIFF
--- a/share/tools/web_config/webconfig.py
+++ b/share/tools/web_config/webconfig.py
@@ -1546,6 +1546,7 @@ def runThing():
             subprocess.call([cmd_path, "/c", "start %s" % url])
         else:
             print("Please add the directory containing cmd.exe to your $PATH")
+            sys.exit(-1)
     elif is_termux():
         subprocess.call(["termux-open-url", url])
     else:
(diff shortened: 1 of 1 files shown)


faho:
Yeah, that works, thanks!

nosed1ve:
Welcome! Happy to help :)

Issue #7738, opened 2021-02-22
Regression in completions with one possible result


mqudsi:
It used to be that completing an argument that evaluates to a single possible result of a directory would first complete `dir` to `dir/` without displaying a list of options on first `<TAB>`/`^I` and then accept that completion and list the contents of the dir as possible completions on second tab. This currently hangs with one completion as the result until the option is accepted, as the completion pager is displayed.

e.g. in the fish source directory:

```fish
~ cat cmake^I
```

No matter how many times tab/^I is pressed, this continues to display a pager with one possible option (`cmake/`) even though the option is accepted (the commandline itself changes from `cmake` to `cmake/`).

I'm unable to bisect it because the behavior in interactive mode (as described) differs from `build/fish -c 'complete -C"nvim cmake"'` which generates multiple entries, but have confirmed it was no present in the last released version of fish.

faho:
It's somewhere around 2b8d2deb0ccc5016a03c3023e0a2709e88c162d1

faho:
Haaaa, it's a smartcase thing:

```fish
complete -c foo -a 'one ONelists' -f
```

reproduces it

faho:
Yeah, let's do the simple fix for now - see that there's only one completion left and bypass the pager.

If we figure out how to display these in the pager later, that can of course go in.

krobelus:
The wrong case in the pager probably needs this fix:

```diff
diff --git a/src/reader.cpp b/src/reader.cpp
index 3a0fd0505..ed768e4a3 100644
--- a/src/reader.cpp
+++ b/src/reader.cpp
@@ -1985,11 +1984,3 @@ bool reader_data_t::handle_completions(const completion_list_t &comp, size_t tok
                 for (idx = 0; idx < max; idx++) {
-                    wchar_t ac = common_prefix.at(idx), bc = el.completion.at(idx);
-                    bool matches = (ac == bc);
-                    // If we are replacing the token, allow case to vary.
-                    if (will_replace_token && !matches) {
-                        // Hackish way to compare two strings in a case insensitive way,
-                        // hopefully better than towlower().
-                        matches = (wcsncasecmp(&ac, &bc, 1) == 0);
-                    }
-                    if (!matches) break;
+                    if (common_prefix.at(idx) != el.completion.at(idx)) break;
                 }
```
On

	$ fish -C "complete -c foo -a 'abc aBd' -f"
	$ foo A^I

[...]

PR #7743, opened 2021-02-25
Only show smartcase-matches if there is no exact match

krobelus:
This makes smartcase completion a bit less smart. I'm not sure if that's okay.
This allows to remove the workaround for #7738


faho:
I'm assuming this is for 3.3?

krobelus:
Yeah that's probably safer. I'm not sure which behavior we want.

Actually, I forgot to disclose: the reason why I made this was that with #7744, d0a8493 no longer works as fix for #7738.
It might be possible to tweak it though, I need to look into that.

krobelus:
This is no longer necessary with the updated #7744.

Issue #7526, opened 2020-12-02
Expansions in file completion are escaped


faho:
As reported by @mojotx on gitter:

```fish
> ls ~/<TAB>
```

escapes the `~`.

Bisects to b38a23a46dfac9fdaf453cced61683765790528c.

faho:
Also goes for variables - `ls $HOME/<TAB>` suffers the same issue, as does `ls /etc/$foo/<TAB>`.

I think this needs to remember what expansions are already in the commandline and let them through, but that seems like a bigger thing and I'm tempted to revert b38a23a46dfac9fdaf453cced61683765790528c in the meantime because it's just cleanup.

faho:
Okay, I've reverted b38a23a for now.

PR #7744, opened 2021-02-25
Use the correct case in completion pager

krobelus:
This resolves the regression mentioned in https://github.com/fish-shell/fish-shell/issues/7738#issuecomment-783681897 and the comments below.

The previous version revived b38a23a, but I didn't manage to fix all problems.
This version uses a simpler fix.

Add a tmux-based tests, including tests for recent interactive regressions #7526 and #7738.
<details><summary>(I coudn't get this pexpect test to work)</summary>

```python
sendline("complete -e echo; complete echo -a 'abc aBd' -f")
send("echo a\t\t\t")
expect_re(r"echo aBc\b")
```

</details>

## TODOs:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,7 @@ Scripting improvements
 
 Interactive improvements
 -------------------------
+- When there are multiple completion candidates, fish inserts their shared prefix. This prefix was computed in a case-insensitive way, resulting in wrong case in the completion pager. This was fixed by only inserting prefixes with matching case (:issue:`7744`).
 
 New or improved bindings
 ^^^^^^^^^^^^^^^^^^^^^^^^
--- a/src/reader.cpp
+++ b/src/reader.cpp
@@ -1996,15 +1996,7 @@ bool reader_data_t::handle_completions(const completion_list_t &comp, size_t tok
                 size_t idx, max = std::min(common_prefix.size(), el.completion.size());
 
                 for (idx = 0; idx < max; idx++) {
-                    wchar_t ac = common_prefix.at(idx), bc = el.completion.at(idx);
-                    bool matches = (ac == bc);
-                    // If we are replacing the token, allow case to vary.
-                    if (will_replace_token && !matches) {
-                        // Hackish way to compare two strings in a case insensitive way,
-                        // hopefully better than towlower().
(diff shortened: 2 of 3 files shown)


faho (commented):
This seems to work, but I've never been comfortable with this bit of the completions, so I wouldn't want to merge this for 3.2.1.

Unless you're really really sure that this time it's correct.

krobelus:
I'm sure it works, I just wasn't sure if it's okay to remove this feature. (Given `touch someCompletion1 somecompletion2`  we would insert the entire prefix when completing  `rm some`.)

ridiculousfish:
Probably fine to merge now. Neat test!

krobelus:
Added changelog.

The `tmux` tests are only run if `tmux` is installed - which is the case with GitHub Actions.
In future we can install it explicitly, and maybe document it.

zanchey:
OK, I'm trying to get my head around this for the release notes. I can't reproduce the "wrong" behaviour in 3.2.1, which I think is because we backed out the case-insensitive completion stuff before the 3.2.0 release? And it doesn't look like we've added it back in?

krobelus:
We didn't back out smartcase completion. This bug was that the pager entries were incorrect after the pager's common-prefix-insertion.

Reproduce with
`fish-3.2.1 -C 'complete -c foo -a "aab aaB" -f'` and typing `foo A` followed by Tab.
The pager was `aaB aaB` but should be `aab aaB`.

zanchey:
Thanks. I think I'm confused about what smartcase completions actually are; with a directory containing a single file `aBC`, the behaviour of `vim ab`<kbd>Tab</kbd> is the same in 3.0.2 and 3.1.1. With a directory containing `aBC` and `abC`, likewise.

krobelus:
Fish uses "tiered" completions, 
1. If any completion prefix-matches, then only prefix matches are shown
2. Else, if any completion substring-matches, then show only these matches (which includes 1.)
3. Else, if any completion substring-matches case-insensitively, then show only these matches (includeing 1 & 2)
4. Finally, show all completions that "subsequence-match", for example `fbr` matches `foobar`

So if some completion on 2 matches the token, then the case-insensitive-only matches (3) are not displayed, if that makes sense.

AIUI, smartcase is about showing both 2 and 3, if the string on the command line is all-lowercase.

PR #7750, opened 2021-02-27
Fix pandoc completion error

### LINKED ISSUE #7747: Pandoc completion errors
The autocomplete commands defined [here](https://github.com/fish-shell/fish-shell/blob/master/share/completions/pandoc.fish#L101-L103):
```fish
complete -c pandoc -r -s F -l filter -a "(find $datadir/filters/** | string replace -- '$datadir/filters/' '')"
complete -c pandoc -r -l template -a "(find $datadir/templates/** | string replace -- '$datadir/templates/' '')"
complete -c pandoc -r -f -l lua-filter -a "(find $datadir/** | string match -r '.lua\$' | string replace -- '$datadir/' '')"
```
Gives the following error when `~/.pandoc` or `~/.pandoc/templates` does not exist
```
fish: No matches for wildcard “/home/birk/.pandoc/templates/**”. See `help expand`.
```

One can reproduce with typing `pandoc --template <TAB>` (note the space before the tab) this results in the following behavior
(see also [recording](https://asciinema.org/a/8A5LD2rP2EFK0YtmcgUeqnDm8)):
```
$ pandoc --template fish: No matches for wildcard “/home/birk/.pandoc/templates/**”. See `help expand`.
find /home/birk/.pandoc/templates/** | string replace -- '/home/birk/.pandoc/templates/' ''
     ^
in command substitution
```

This problem should be pretty easy to fix by modifying the affected completions, however there is probably a point in discussing whether errors in completion commands should be shown at the prompt or not.

birktj:
## Description

Pandoc completions would give an error if the folder `~/.pandoc` does exist as described in #7747 . This fixes #7747

I have checked that changes fix the error I noted by using:
```
$ complete -c pandoc -e
$ complete -c pandoc -r -l template -a "(find $datadir/templates -type f 2>/dev/null | string replace -- '$datadir/templates/' '')"
$ pandoc --template <TAB>
```
However do note that I have not tested if the expected completions work although the find commands seem to work and I see no reason for the completions not to.

## TODOs:

I have not done any of these, however I assume they are not relevant for such a small fix.

### DIFF
--- a/share/completions/pandoc.fish
+++ b/share/completions/pandoc.fish
@@ -98,9 +98,9 @@ complete -c pandoc -r -f -l bibliography -k -a "(__fish_complete_suffix 'ria')"
 complete -c pandoc -r -f -l lua-filter -k -a "(__fish_complete_suffix 'lua')"
 
 # options that take files in DATADIR
-complete -c pandoc -r -s F -l filter -a "(find $datadir/filters/** | string replace -- '$datadir/filters/' '')"
-complete -c pandoc -r -l template -a "(find $datadir/templates/** | string replace -- '$datadir/templates/' '')"
-complete -c pandoc -r -f -l lua-filter -a "(find $datadir/** | string match -r '.lua\$' | string replace -- '$datadir/' '')"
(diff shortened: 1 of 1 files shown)


zanchey:
Yes, that looks sensible. The completions should ideally not be using find anyway, but this will fix the problem.

PR #7754, opened 2021-03-01
Rewrite the real file if history file is a symlink

ilyagr:
This is a followup to https://github.com/fish-shell/fish-shell/pull/7728 that makes `fish_history` behave consistently with the new behavior of `fish_variables`. It should also help with issues like https://github.com/fish-shell/fish-shell/issues/7553.

I combined the tests for the two features, renamed the test file, and fixed the issue where the old test could pass if fish didn't write anything to the expected location of `fish_variables`. I couldn't find a good way to have a nice diff for the rename. Let me know if you prefer separate tests.

The checks/git.fish test is currently failing, but I believe that is because of <...>. **Edit:** that wasn't the reason.

## TODOs:


faho:
>The checks/git.fish test is currently failing, but I believe that is because of 384975c.

Mind posting your error? It's not failing on CI or for me, and it's not skipped either (that test only runs if git is installed). Not that I think it's *related*, but it might show differences in our setup.

ilyagr:
Yes, here you go:

```
Testing file checks/function.fish ... ok (24 ms)
Testing file checks/functions.fish ... ok (20 ms)
Testing file checks/git.fish ... Failure:

  The CHECK on line 29 wants:
    add\tAdd file contents to the index

    adda\tAlias for add :/

  Context:
    adda\tAlias for add :/ <= no check matches
    add\tAdd file contents to the index
    foo\tUntracked file
     (newbranch)

  when running command:
    ../test/root/bin/fish -i checks/git.fish
Testing file checks/glob.fish ... ok (52 ms)
```

It fails on the master branch as well for me. I haven't looked at it in detail, I assumed it's the same problem as https://github.com/fish-shell/fish-shell/runs/1997846755?check_suite_focus=true, but that's probably not the case.

In fact, it's probably because the test is affected by my global `.gitconfig`. I have an alias called `adda` (add all), and I think that confuses the test.

faho:
>In fact, it's probably because the test is affected by my global .gitconfig. I have an alias called adda (add all), and I think that confuses the test.

Yeah, that's it. We either need to disable .gitconfig here or at least anchor the match some more (`grep '^add\t'`). Given that git doesn't allow aliases named like commands that's safe.

Either way, not related.

>I assumed it's the same problem as https://github.com/fish-shell/fish-shell/runs/1997846755?check_suite_focus=true, but that's probably not the case.

Nah, that's lsan being lsan - we do some funky business in the tests that regularly confuses it. But given these are tests, the funky business is the point.

ilyagr:
Yes, exactly. If you like, here's the same fix you mentioned in pull request form: https://github.com/fish-shell/fish-shell/pull/7755. It works for me.

faho:
Alright, at a glance this looks okay but I don't want to add it to 3.2.0 this late (hopefully) in the cycle and risk breaking history.

If we do a 3.2.1 it can go in there.

ilyagr:
Thanks for looking at it!

I didn't realize you were going to release 3.2 today. Congratulations! I'm happy for this to get in later.

There's one minor issue I fixed in a new commit. It was a bit confusing. Let me know if you think it's potentially problematic.

I wanted to see what happens when the symlink points to an invalid location or another file-system. In the case, the history writing usually fails, which I think is a good outcome. However, no error is shown to the user, which is less good. It is also a bit surprising, since in the case of `fish_variables`, very similar code does result in errors shown to the user.

The error for `fish_variables` is shown here:

https://github.com/fish-shell/fish-shell/blob/589eb34571e00da530395776166dd873489d4027/src/env_universal_common.cpp#L486-L491

In `history.cpp`, we have the following.

https://github.com/fish-shell/fish-shell/blob/589eb34571e00da530395776166dd873489d4027/src/history.cpp#L852-L854

I'm not exactly sure why the former code produces output and the latter does not, but it seems to be intended. At least, the later produces output when running tests (modified to point symlinks to bad places). I hope it will be shown in release builds of `fish` as well.

So, I added a commit that adds the first kind of `FLOG`  to `history.cpp`.

faho:
>I'm not exactly sure why the former code produces output and the latter does not, but it seems to be intended

The former produces output because it's FLOG category "error", which is enabled by default. (this is the first argument to "FLOGF")

The latter has the category "history_file", which isn't. You'd have to start fish with `fish -d history-file` to see it.

Making it consistent with the former seems correct. There's a point to be made that they should be "warning" because fish is still mostly functional (but less so without universal variables), but that's a question for another time.

I would just change "history_file" to "error" for now - keep the message because it's more informative, but change the category. Definitely no two FLOGs with different categories.

ilyagr:
Thanks, that sounds good, I'll do it tomorrow.

Should I use the more informative message even though it is not localized?

Also, there is an option to move the entire `env_universal_t::move_new_vars_file_into_place` function to somewhere like `wutil.h` and use it, especially if we want to use the less informative but localized message. I'm not sure how eager I am to do it.

faho:
>Should I use the more informative message even though it is not localized?

Localize it - i.e. put it through `_`. (not that fish's translations are currentl brilliant, but we should at least *give* the common error messages a chance)

tests/checks/symlinks-not-overwritten.fish
```diff
@@ -0,0 +1,45 @@
...
+            when: 1614577746' >$target_file
+    cp $target_file $backup_file
+    ln -sf $target_file $history_file
+    # The one way to ensure non-interactive fish writes to the history file
+    $fish -c 'echo all | history delete deleted | grep echo'
+    # CHECK: [1] echo I will be deleted from history
+
+    if not test -L $history_file
+        echo fish_history has been overwritten
+    else if test "(cat $target_file)" = "(cat $backup_file)"
```
krobelus, line 39: This is always false since the `cat` is not run.

```suggestion
    else if cmp $target_file $backup_file
```

tests/checks/symlinks-not-overwritten.fish
```diff
@@ -0,0 +1,45 @@
...
+    set -l target_file $XDG_CONFIG_HOME/fish/target_fish_variables
+    set -l fish_variables $XDG_CONFIG_HOME/fish/fish_variables
+
+    echo >$target_file
+    ln -sf $target_file $fish_variables
+    $fish -c 'set -U variable value'
+
+    if not test -L $fish_variables
+        echo fish_variables has been overwritten
+    else if test (wc -c < $target_file) -le 3
```
krobelus, line 15: this is a bit tricky - I wonder if there is an easier way to write this.

Maybe `grep -v . $target_file`, or the backup file trick below?

tests/checks/symlinks-not-overwritten.fish
ilyagr: That's very good to know. Apart from working, it's also less ugly.

I'll just fix it in my version to rebase one less time.

ilyagr: I'll also add `>/dev/null`, unless you have a better idea.

tests/checks/symlinks-not-overwritten.fish
ilyagr: I'll just do the backup trick. I guess `grep` would work, but I'm vaguely afraid of `echo` writing `\r\n` on Windows and `grep` not understanding it. Or something.

tests/checks/symlinks-not-overwritten.fish
ilyagr: Actually, `cmp` gives a message to `stderr`: `cmp: EOF on /tmp/tmp.z101oB6q12/fish/target_fish_history which is empty `. So, it'll be &>/dev/null.

tests/checks/symlinks-not-overwritten.fish
```diff
@@ -0,0 +1,45 @@
+#RUN: %fish -C 'set -g fish %fish' %s
+begin
```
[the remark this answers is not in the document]
ilyagr: I copied it from `broken-config.fish`. I can't judge if it's necessary, but I'll trust you on it. The tests still pass.

ilyagr:
Thank you very much for your helpful suggestions, @krobelus and @faho! I put them all in, I believe.

@faho: I updated the error message, and even managed to cause it to happen on my machine. It now looks like this: `error: Error when renaming history file: Invalid cross-device link`.

ilyagr:
In case the rebase is annoying, here are the changes since the previous version: https://github.com/fish-shell/fish-shell/compare/b1a203d...ilyagr:tmp. Maybe I'll try a merge next time.

faho:
Merged, thanks!

ilyagr:
Thank you!

PR #7755, opened 2021-03-01
Fix `git.fish` test failing because of an alias

ilyagr:
I have an alias `adda=add :/` in my global `.gitconfig`, which
made the test fail. This is the fix.
## TODOs:


faho:
Merged, thanks!

PR #7763, opened 2021-03-01
Update rsync completions

Jakeler:
## Description
This adds completions for the newer (and very useful) `info`+`debug` flags, parsed from the help output:
Screenshot_20210228_220622
Also many other options that were missing. Based on the current rsync 3.2.3 man page.

## TODOs:


share/completions/rsync.fish
```diff
@@ -2,8 +2,24 @@ function __rsync_remote_target
     commandline -ct | string match -r '.*::?(?:.*/)?' | string unescape
 end
 
+function __rsync_parse_flags -d "Print info|help FLAGS from help output"
+    set -l helptext $argv[2..-1]
+    for line in $helptext
+        set -l tokens (string match -r "([A-Z]+)(?: {2,})(.+)" $line)
+        if [ (count $tokens) -ge 3 ]
+            echo -e "$tokens[2]\t$tokens[3]"
```
faho, line 10: A few nits here:

- We prefer `test` over `[ ]`
- The `echo -e` isn't necessary here, just use `echo $tokens[2]\t$tokens[3]`

share/completions/rsync.fish
```diff
@@ -2,8 +2,24 @@ function __rsync_remote_target
     commandline -ct | string match -r '.*::?(?:.*/)?' | string unescape
 end
 
+function __rsync_parse_flags -d "Print info|help FLAGS from help output"
+    set -l helptext $argv[2..-1]
```
faho, line 6: If I'm reading this correctly, you're skipping the first line because it's a heading - the output looks like

```
Use OPT or OPT1 for level 1 output, OPT2 for level 2, etc.; OPT0 silences.

BACKUP     Mention files backed up
COPY       Mention files copied locally on the receiving side
DEL        Mention deletions on the receiving side
FLIST      Mention file-list receiving/sending (levels 1-2)
MISC       Mention miscellaneous information (levels 1-2)
MOUNT      Mention mounts that were found or skipped
NAME       Mention 1) updated file/dir names, 2) unchanged names
PROGRESS   Mention 1) per-file progress or 2) total transfer progress
REMOVE     Mention files removed on the sending side
SKIP       Mention files that are skipped due to options used
STATS      Mention statistics at end of run (levels 1-3)
SYMSAFE    Mention symlinks that are unsafe

ALL        Set all --info options (e.g. all4)
NONE       Silence all --info options (same as all0)
HELP       Output this help message
```

If so, there should be a comment that says so, for future readers.

share/completions/rsync.fish
```diff
@@ -2,8 +2,24 @@ function __rsync_remote_target
...
+        end
+    end
+end
+
 complete -c rsync -s v -l verbose -d "Increase verbosity"
+complete -c rsync -l info -d "Fine-grained informational verbosity" -xa "
+(__rsync_parse_flags (rsync --info=help))"
+complete -c rsync -l debug -d "Fine-grained debug verbosity" -xa "
+(__rsync_parse_flags (rsync --debug=help))"
+complete -c rsync -l stderr -xa '{errors,all,client}' -d "change stderr output mode, default: errors"
```
faho, line 20: No need for the brace expansion, just `'errors all client'` will do.

share/completions/rsync.fish
```diff
@@ -16,30 +32,41 @@ complete -c rsync -l suffix -d "Backup suffix (default ~ w/o --backup-dir)"
 complete -c rsync -s u -l update -d "Skip files that are newer on the receiver"
 complete -c rsync -l inplace -d "Update destination files in-place"
 complete -c rsync -l append -d "Append data onto shorter files"
+complete -c rsync -l append-verify -d "--append w/old data in file checksum"
```
faho, line 35: What does this mean?

I'm assuming "w/old" means "with old", but what's old data in the file checksum?

The full explanation in the man page goes into more detail:

>This special copy mode works like --append except that all the data in the file is included in the checksum verification (making it much less efficient but  also potentially  safer)

So it's "append with checksum computed over all the old data".

Unfortunately it then goes on to explain

>This option can be dangerous if you aren't 100% sure that all the files in the transfer are shared, growing files.

And now I'm confused and unsure again - is this safer or dangerous? So we check `--append`, and find out that *appending* is dangerous if not all files grow, and append-verify is a bit safer than that in confirming the checksum, but still bad if things aren't growing.

So something like "append with full file checksum" seems to communicate the basic idea.

share/completions/rsync.fish
```diff
@@ -16,30 +32,41 @@ complete -c rsync -l suffix -d "Backup suffix (default ~ w/o --backup-dir)"
...
 complete -c rsync -s o -l owner -d "Preserve owner (super-user only)"
 complete -c rsync -s g -l group -d "Preserve group"
 complete -c rsync -l devices -d "Preserve device files (super-user only)"
 complete -c rsync -l specials -d "Preserve special files"
 complete -c rsync -s D -d "Same as --devices --specials"
-complete -c rsync -s t -l times -d "Preserve times"
+complete -c rsync -s t -l times -d "Preserve modification times"
+complete -c rsync -s U -l atimes -d "Preserve access (use) times"
+complete -c rsync -l open-noatime -d "Avoid changing the atime on opened files"
+complete -c rsync -l crtimes -d "Preserve create times (newness)"
```
faho, line 59: This is the first time I'm reading the term "newness" for creation time, and I don't think it helps to clarify anything. "create time" or "creation time" is already perfectly clear.

share/completions/rsync.fish
```diff
@@ -16,30 +32,41 @@ complete -c rsync -l suffix -d "Backup suffix (default ~ w/o --backup-dir)"
...
 complete -c rsync -s O -l omit-dir-times -d "Omit directories when preserving times"
+complete -c rsync -s J -l omit-link-times -d "Omit symlinks when preserving times"
 complete -c rsync -l super -d "Receiver attempts super-user activities"
+complete -c rsync -l fake-super -d "Store/recover privileged attrs using xattrs"
 complete -c rsync -s S -l sparse -d "Handle sparse files efficiently"
+complete -c rsync -l preallocate -d "Allocate dest files before writing them"
+complete -c rsync -l write-devices -d "Write to devices as files (implies --inplace)"
 complete -c rsync -s n -l dry-run -d "Show what would have been transferred"
 complete -c rsync -s W -l whole-file -d "Copy files whole (without rsync algorithm)"
+complete -c rsync -l checksum-choice -xa '{xxh128,xxh3,xxh64,md5,md4,none}' -d "Choose the checksum algorithm (aka --cc)"
```
faho, line 69: If it's also known as `--cc` just add that - `-l cc` (can be in the same invocation).

Also unnecessary brace expansion.

share/completions/rsync.fish
```diff
@@ -52,19 +79,27 @@ complete -c rsync -l del -d "An alias for --delete-during"
...
 complete -c rsync -l min-size -xa '(seq 0 10){K,M,G}' -d "Don’t transfer any file smaller than SIZE"
+complete -c rsync -l max-alloc -xa '(seq 0 10){K,M,G}' -d "Change process memory allocation limit"
 complete -c rsync -l partial -d "Keep partially transferred files"
 complete -c rsync -l partial-dir -xa '(__fish_complete_directories)' -d "Put a partially transferred file into DIR"
 complete -c rsync -l delay-updates -d "Put all updated files into place at end"
 complete -c rsync -s m -l prune-empty-dirs -d "Prune empty directory chains from file-list"
 complete -c rsync -l numeric-ids -d "Don’t map uid/gid values by user/group name"
+complete -c rsync -l usermap -xa '(__fish_complete_users)' -d "Custom username mapping"
+complete -c rsync -l groupmap -xa '(__fish_complete_groups)' -d "Custom username mapping"
+complete -c rsync -l chown -xa '(__fish_complete_users;__fish_complete_groups)' -d "Combined username/groupname mapping"
```
faho, line 100: A space after the `;` would be nice.

(also this does user:group, so maybe `(__fish_complete_users):(__fish_complete_groups)` might be helpful, or more complicated logic to only complete the group *after* the user - see `__fish_complete_user_at_hosts` for something similar)

share/completions/rsync.fish
```diff
@@ -74,7 +109,9 @@ complete -c rsync -l compare-dest -xa '(__fish_complete_directories)' -d "Also c
 complete -c rsync -l copy-dest -xa '(__fish_complete_directories)' -d "Also compare received files relative to DIR and include copies of unchanged files"
 complete -c rsync -l link-dest -xa '(__fish_complete_directories)' -d "Hardlink to files in DIR when unchanged"
 complete -c rsync -s z -l compress -d "Compress file data during the transfer"
-complete -c rsync -l compress-level -d "Explicitly set compression level"
+complete -c rsync -l compress-choice -xa '{zstd,lz4,zlibx,zlib,none}' -d "Choose the compression algorithm (aka --zc)"
```
faho, line 112: Same as `--cc`.

share/completions/rsync.fish
```diff
@@ -84,29 +121,43 @@ complete -c rsync -l include -d "Don’t exclude files matching PATTERN"
...
 complete -c rsync -l checksum-seed -x -d "Set block/file checksum seed (advanced)"
 complete -c rsync -s 4 -l ipv4 -d "Prefer IPv4"
 complete -c rsync -s 6 -l ipv6 -d "Prefer IPv6"
-complete -c rsync -l version -d "Display version and exit"
-complete -c rsync -l help -d "Display help and exit"
+complete -c rsync -l daemon -d "Run as an rsync daemon"
+complete -c rsync -l config -r -d "Specify alternate rsyncd.conf file"
+complete -c rsync -l dparam -x -d "Override global daemon config parameter"
+complete -c rsync -l no-detach -x -d "Do not detach from the parent"
+complete -c rsync -s V -l version -d "Display version + other info and exit"
```
faho, line 159: Maybe "version and feature info"? It lists the sort of stuff you'd en/disable via compile flags, like the compression and checksum algorithms.

faho (changes requested):

share/completions/rsync.fish
Jakeler: Yes, you read that correctly, I have added a comment.

share/completions/rsync.fish
Jakeler: I agree, did a quick web search now and could not find other mentions... seems to be only in this man page. 
In `stat` it is called birth or `btime` in the struct, so I have changed it.

share/completions/rsync.fish
Jakeler: `(__fish_complete_users):(__fish_complete_groups)` unfortunately does not work in my testing, if I type `:` and hit TAB it removes the whole option? Is there a any simple solution for the common user:group pattern?

share/completions/rsync.fish
Jakeler: Good idea, changed now.

share/completions/rsync.fish
Jakeler: As far as I understand it the old data is the (smaller) file that already exists on the destination. With normal append it calculates the checksum only over the appended data that gets transferred (possibly over unstable networks).

> appending is dangerous if not all files grow, and append-verify is a bit safer than that in confirming the checksum, but still bad if things aren't growing.

I think that is correct, append is dangerous in general, but the full checksum makes it at least a bit safer.

share/completions/rsync.fish
faho: Aaaah, okay, both of these provide a description (after a tab character), so only the users part is actually used (because the rest is now after a tab).

Here's what the actual chown completions do:

```fish
complete -c chown -d Username -a "(__fish_print_users):"
complete -c chown -d Username -a "(string match -r -- '.*:' (commandline -ct))(__fish_complete_groups)"
```

You could probably make a helper function that

1. if you don't have a ":" in the current token (`commandline -ct`) runs `__fish_complete_users`
2. If you do, prints the current token *plus* everything in `__fish_complete_groups`

Jakeler: Okay, I think it is best to merge as it is and make such a helper function in a separate PR.

faho:
Merged, thanks!

share/completions/rsync.fish
krobelus: Since `:somegroup` is valid, that might need a separate entry (untested):
 `(echo; __fish_print_users):(__fish_complete_groups)`

PR #7789, opened 2021-03-07
Make `help` and `fish_config` work on Chrome OS

ilyagr:
When `fish` is running in the Chrome OS Linux VM (Crostini), both `help` and `fish_config` currently open a "file not found"
page. That is because on Crostini, `BROWSER` is usually set to `garcon-url-handler`, which opens URLs in the host OS Chrome
browser. That browser lacks access to the Linux file system.

This PR fixes these commands. `help` now opens the URL on www.fishshell.com.  `fish_config` now opens the URL for the
server it starts. Previously, it opened a local file that redirects to the same URL.

In the case of `help`, the situation could be improved further by starting a web server to serve help. I don't know of another way to access `/share/fish` from outside the VM without user intervention, and I think that might be a part of the security
model for the Crostini VM.

It's hard to write a test for this. I checked that `help math`, `python2 webconfig.py`, and `python3 webconfig.py` work on my
machine running in Crostini.

----------

I also rearranged the imports a little to group them together. It's in a separate commit, let me know if you prefer that I not do that.

Another alternative is to look much harder for a browser installed in the Linux VM. By default, none are installed, I think. I do think a Linux browser should be used if the user has installed it *and* put it into `BROWSER`, though.

## TODOs:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,7 @@ Scripting improvements
 
 Interactive improvements
 ------------------------
+- ``help`` and ``fish_config`` no longer open to a "Your file couldn't be accessed" page when fish is running in a Chrome OS Crostini Linux VM and URLs are opened in Chrome running outside the VM.
 
 New or improved bindings
 ^^^^^^^^^^^^^^^^^^^^^^^^
--- a/share/functions/help.fish
+++ b/share/functions/help.fish
@@ -142,12 +142,20 @@ function help --description 'Show help for the fish shell'
             set fish_help_page "index.html#$fish_help_item"
     end
 
+    # In Crostini Chrome OS Linux, the default browser opens URLs in Chrome running outside the
+    # linux VM. This browser does not have access to the Linux filesystem. This uses Garcon, see e.g.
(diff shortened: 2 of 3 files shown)


ilyagr:
I realized that the field `webbrowser.get().name` I used is, strictly speaking, [not documented](https://docs.python.org/3/library/webbrowser.html#browser-controller-objects). I checked, it has been around [since Python 2.5](https://github.com/python/cpython/commit/e8f244305ef4f257f6999b69601f4316b31faa5e). I suppose it's not completely inconceivable that it might one day disappear, or that it's not present in PyPy.

I don't know a better way to do this, though. If we are worried enough, we could  sacrifice some aesthetics and make sure it never fails:

```python
try:
    return "garcon-url-handler" in webbrowser.get().name
except NameError:
    # Explanation
    return False
```

faho:
>or that it's not present in PyPy.

Well, that would make PyPy incompatible, which would probably be a bug in PyPy.

----

Considering that the "name" attribute is on the BaseBrowser class I think it's safe enough.

Merged, thanks!

ilyagr:
Thank you, @faho!

PR #7793, opened 2021-03-07
Added descriptions to __fish_print_addresses

### LINKED ISSUE #7787: Enhance __fish_print_addresses
I would like to enhance the function `__fish_print_addresses` a bit. This function is currently used in completions when something like an address to bind to is required.

It could be enhanced to include the interface name as a description, which would help selecting the correct address.

Besides I would introduce an optional argument like `--all` that also outputs `0.0.0.0` and `::`.

Before conducting a PR: can I just change the existing function or should I create a new one? While a change would not affect the fish codebase, theoretically it's possible that 3rd party scripts use this functions and rely on the fact that no descriptions are included. But if I introduce a new function, the existing one would be unused across the fish codebase.

exploide:
## Description

- Added descriptions to __fish_print_addresses
- added --all flag to include 0.0.0.0 and ::
- adapted use of this function in existing completions
- fixes #7787

## TODOs:

### DIFF
--- a/share/completions/msfdb.fish
+++ b/share/completions/msfdb.fish
@@ -24,7 +24,7 @@ complete -c msfdb -l db-port -x -d 'Database port'
 complete -c msfdb -l db-pool -x -d 'Database connection pool size'
 
 # Web Service Options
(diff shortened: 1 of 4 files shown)


faho:
Merged, thanks!

Issue #4961, opened 2018-05-06
String interpolation with undefined string causes loss of data


mqudsi:
```fish
set foo 'hello'
set bar 'world'
set foo $foo$bar
echo $foo #prints 'hello world'
set foo 'hello'
set -e bar
set foo $foo$bar
echo $foo #prints nothing
```

krader1961:
That's the expected behavior. And the behavior is defined. Although the most relevant [section of the documentation](http://fishshell.com/docs/current/index.html#cartesian-product) would benefit from an example and some explanatory text of the dangers.

Works for me if the vars are quoted so that there is actual string interpolation. That is:

```
$ set foo 'hello'
$ set foo "$foo$bar"
$ echo $f
hello
```

I know this has been discussed in the past with the decision to not change current behavior. But in light of the pending 3.0 release perhaps it should be discussed again.

mqudsi:
Thanks, @krader1961. And you're right, the result is an empty string and not an undefined variable.

I'm not sure I understand semantically how quoting it differs, or what rationale would explain why it works in quotes but not without.

krader1961:
When this issue comes up it seems to always be in the context of
concatenating an empty or undefined var to a var with one or more
values. So the obvious question is whether that should be special-cased
to treat the empty or undefined var as if it consisted of a single empty
string. Note that we would only want to do this if at least one var
participating in the cartesian product had at least one value.

Some examples may help. In the following table a value of `(nil)` means
the var expansion produces no value(s). Literal strings are enclosed in
curly brackets. And `{}` means the empty string.

These are the vars used in the second table showing how var expansion
works now and might work in the future.

| Var Definition | Meaning |
| -------- | ------ |
| `set -e undef` | An undefined var |
| `set empty` | A var with no values |
| `set s1 a` | A var with one value: `a` |
| `set s2 x y` | A var with two values: `x` and `y` |
| `set z1 ''` | A var with one empty string |
| `set zs '' 't'` | A var with one empty string and one non-empty string |

| Example Use | Current Behavior | Proposed Behavior |
| ----------- | ---------------- | ----------------- |
| `$undef` | (nil) | (nil) |
| `$empty$undef` | (nil) | (nil) |
| `$s1$undef` | (nil) | {a} |
| `$s2$undef` | (nil) | {x} {y} |
| `$s1$empty` | (nil) | {a} |
| `$s2$empty` | (nil) | {x} {y} |
[...]

Issue #2583, opened 2015-12-07
Should I add a shebang to my script?


edouard-lopez:
I'm new to `fish`, but have some background in `zsh` and `bash`. 

As I'm working to make [pure-fish theme](https://github.com/rafaelrinaldi/pure) supported in `oh-my-fish`, I was a bit surprised that [`omf new theme my_theme`](https://github.com/oh-my-fish/oh-my-fish/wiki/Advanced#creating) command create boilerplate files without a shebang. 

I was expecting to see:

    #!/usr/bin/env fish

Should I add a shebang to my scripts? Is it a `fish` best-practice?

terlar:
Depends on what you plan to do, the shebang is there if you plan to execute/run the script by itself, and doesn't really provide any value for fish scripts that are supposed to be sourced or not run by themselves.

izumidamasa:
Hello. Does not make sense for your theme, which works OK to add ohmyfish support. Let me explain, instead omf should support your theme, or any theme I think.

And btw, I have being useing https://github.com/vkovtash/pure for a few months now. What is the difference with your theme?

edouard-lopez:
@izumidamasa I'm not the author of the fish version, just working on the omf port.

faho:
I'd say you SHOULD have a shebang if it's a standalone program that just happens to be written in fishscript, but you SHOULD NOT have one if it's meant to be sourced (like a function definition).

The reason being that running a function definition by itself is not useful. These files also should not have the executable bit on them.

Does that make sense?

edouard-lopez:
Yep, nice explanation but if it not harmful, I like to have a prologue that tell what interpreter is needed.

edouard-lopez:
Yep, nice explanation but if it not harmful, I like to have a prologue that tell what interpreter is needed.

Issue #1281, opened 2014-02-03
Failed to execute process:startx


ghost:
FreeBSD 10
fish, version 2.1.0

```
Failed to execute process '/usr/local/bin/startx'. Reason:
exec: Exec format error
The file '/usr/local/bin/startx' is marked as an executable but could not be run by the operating system.
```

krader1961:
Someone was asking about this on IRC/Gitter recently because they were seeing the same error. Albeit with different scripts created at their company that did not contain a [shebang line](https://en.wikipedia.org/wiki/Shebang_(Unix)). They noticed that those scripts ran just fine when launched from sh/bash and therefore didn't understand why fish reported an error.

The answer is that the Bourne shells and those derived from it like Bash and Zsh contain a compatibility shim that is no longer necessary. There is a common misconception that the shebang line is read by the shell. That isn't true: it's read by the kernel. Specifically, whatever part of the kernel implements the `execve()` call. Long, long, ago (like 35+ years) shebang lines were new and there was no guarantee your kernel would read them to figure out which shell to run to interpret the script. So if the `execve("/path/to/script")` call failed your shell (/bin/sh and /bin/csh back then was pretty much the only options) would just assume the script was a Bourne shell script and directly call `execve("/bin/sh", "/path/to/script")`.

There is a kernel of truth to the notion that the shebang line is read by the shell. Long ago the Bourne shell would look to see if the file began with `:` or `#` characters which was interpreted as meaning the file contained a Bourne shell script. That behavior was made obsolete and ultimately removed after the introduction of shebang lines.

Bottom line is that today you should not rely on that compatibility shim. Every script must have a shebang line. Full stop. Anyone telling you it's okay to omit the shebang line is propagating a dangerous practice.

PR #7802, opened 2021-03-09
Allow scripts without #!

jart:
This change modifies the fish safety check surrounding execve / spawn so
it can run shell scripts having concatenated binary content. We're using
the same safety check as FreeBSD /bin/sh [1] and the Z-shell [5].  POSIX
was recently revised to require this behavior:

    "The input file may be of any type, but the initial portion of the
     file intended to be parsed according to the shell grammar (XREF to
     XSH 2.10.2 Shell Grammar Rules) shall consist of characters and
     shall not contain the NUL character. The shell shall not enforce
     any line length limits."

    "Earlier versions of this standard required that input files to the
     shell be text files except that line lengths were unlimited.
     However, that was overly restrictive in relation to the fact that
     shells can parse a script without a trailing newline, and in
     relation to a common practice of concatenating a shell script
     ending with an 'exit' or 'exec $command' with a binary data payload
     to form a single-file self-extracting archive." [2] [3]

One example use case of such scripts, is the Cosmopolitan C Library [4]
which configuse the GNU Linker to output a polyglot shell+binary format
that runs on Linux / Mac / Windows / FreeBSD / OpenBSD / NetBSD / BIOS.

Fixes jart/cosmopolitan#88

[1] https://github.com/freebsd/freebsd-src/commit/9a1cd363318b7e9e70ef6af27d1675b371c16b1a
[2] http://austingroupbugs.net/view.php?id=1250
[3] http://austingroupbugs.net/view.php?id=1226#c4394
[4] https://justine.lol/cosmopolitan/index.html
[5] https://github.com/zsh-users/zsh/commit/326d9c203b3980c0f841bc62b06e37134c6e51ea

## Description

Talk about your changes here.

Fixes issue #

## TODOs:

### DIFF
--- a/src/exec.cpp
+++ b/src/exec.cpp
@@ -13,6 +13,7 @@
 #ifdef HAVE_SPAWN_H
 #include <spawn.h>
 #endif
+#include <paths.h>
 #include <stdio.h>
 #include <sys/wait.h>
 #include <unistd.h>
@@ -62,6 +63,46 @@ enum class launch_result_t {
     failed,
 } __warn_unused_type;
 
+static bool is_thompson_shell_payload(const char *p, size_t n) {
+    if (!memchr(p, '\0', n)) return true;
+    bool haslower = false;
+    for (; *p; p++) {
+        if (islower(*p) || *p == '$' || *p == '`') {
+            haslower = true;
+        }
+        if (haslower && *p == '\n') {
+            return true;
+        }
+    }
+    return false;
+}
+
+/// This function checks the beginning of a file to see if it's safe to
+/// pass to the system interpreter when execve() returns ENOEXEC.
+///
+/// The motivation is to be able to run classic shell scripts which
+/// didn't have shebang, while protecting the user from accidentally
+/// running a binary file which may corrupt terminal driver state. We
+/// check for lowercase letters because the ASCII magic of binary files
+/// is usually uppercase, e.g. PNG, JFIF, MZ, etc. These rules are also
+/// flexible enough to permit scripts with concatenated binary content,
+/// such as Actually Portable Executable.
+bool is_thompson_shell_script(const char *path) {
+    int e = errno;
+    bool res = false;
+    int fd = open(path, O_RDONLY | O_NOCTTY);
+    if (fd != -1) {
+        char buf[256];
+        ssize_t got = read(fd, buf, sizeof(buf));
+        close(fd);
+        if (got != -1 && is_thompson_shell_payload(buf, got)) {
+            res = true;
+        }
+    }
+    errno = e;
+    return res;
+}
+
 /// This function is executed by the child process created by a call to fork(). It should be called
 /// after \c child_setup_process. It calls execve to replace the fish process image with the command
 /// specified in \c p. It never returns. Called in a forked child! Do not allocate memory, etc.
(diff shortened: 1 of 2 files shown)


faho:
The title here isn't *technically* incorrect, but it's misleading.

This allows "more" scripts without shebang by allowing basically any scripts at all without a shebang (the current behavior seems to only be a windows thing and I've never actually encountered it and didn't know we had it).

This is a POSIX legacy misfeature we've long rejected (and this new version of it makes it even more awkward by adding a weird heuristic on top)

The reason we've rejected it is that, until now, when a script doesn't have a shebang it's because the author simply forgot to add one, and it's not at all clear that it's actually a /bin/sh script. So instead of printing a sensible error, you'd now print a non-sensible one because you've tried to run a python script with /bin/sh.

Annoyingly, cosmopolitan *might* turn this around by making it more likely that it's actually meant to be without a shebang, and for the record I hate it. I absolutely, viscerally, despise this hack.

>We're using
the same safety check as FreeBSD /bin/sh [1] and the Z-shell [5]. POSIX
was recently revised to require this behavior:

Note that we do not, in general, *care* about compatibility or especially POSIX conformance.

Yet at the same time I feel that this particular bit is no longer a hill worth dying on. Is having a script without a shebang a sensible thing to do? No, because it makes the failure mode of forgetting a shebang worse. Simply adding a shebang always would be better. Is it worth it to keep failing in this case when things keep abusing scripts without shebangs? Probably not. We should probably merge this, for 3.3.

faho:
>This change aims to offer you the best possible recommendation that's the result of an effort to balance the concerns of many groups.

Just out of curiosity, do I understand the heuristic thinks this is a shellscript:

```python
import os
os.sleep(100)
```

and this is a shellscript:

```c++
int main() {
    return 129;
}
```

but this isn't:

```sh
ECHO FOO
```

faho:
To be clear, let me repeat my opinion:

>We should probably merge this, for 3.3.

I just hate it, that's all.

jart:
It's great to hear that! 

The heuristic only applies if there's a NUL character in the first 256 bytes. So `echo 'ECHO FOO' >script` is a valid script. On the hand, `printf 'ECHO HI\0' >script` would not be a valid script. We check for lowercase because ASCII magic for binary files is usually uppercase, e.g. `PNG`, `JFIF`, etc.

faho:
>We check for lowercase because ASCII magic for binary files is usually uppercase, e.g. PNG, JFIF, etc.

That sounds like a good idea for a comment, to be honest.

zanchey:
Cosmopolitan is bonkers and I love it! I think this is good, and fixes #4961, https://github.com/fish-shell/fish-shell/issues/1281#issuecomment-265788224, #2583 also related.

faho:
Important: #2583 is about a *fish script* without a shebang. This will make that case slightly *worse*, because it will try to run it with /bin/sh, which will probably succeed for a part of it - dash runs up to the first syntax error.

This means:

```fish
echo before
set path /banana
rm -rf $path/*
if true
    echo in
end
echo after
```

will print "before" and run `rm -rf /*` (because the set does something else and so $path is empty) and *then* print a syntax error.

zanchey:
The advice there should be all-shebang, all the time, then.

src/exec.cpp
```diff
@@ -71,36 +112,19 @@ enum class launch_result_t {
...
-            }
-
-            execve(sh_command, argv2, envv);
-        }
+    // The shebang wasn't introduced until UNIX Seventh Edition, so if
+    // the kernel won't run the binary we hand it off to the intpreter
+    // after performing a binary safety check, recommended by POSIX: a
+    // line needs to exist before the first \0 with a lowercase letter
+    if (err == ENOEXEC && is_thompson_shell_script(actual_cmd)) {
+        *--argv = const_cast<char *>(_PATH_BSHELL);
```
mqudsi, line 126: Playing fast and loose with constness here is fine but it’s not clear to me that this write is to owned memory?

mqudsi:
> Cosmopolitan is wild and I love it! I think this is good, and fixes #4961, [#1281 (comment)](https://github.com/fish-shell/fish-shell/issues/1281#issuecomment-265788224), #2583 also related.

@zanchey is 4961 a typo? It doesn’t seem related?

zanchey:
> > Cosmopolitan is wild and I love it! I think this is good, and fixes #4961, [#1281 (comment)](https://github.com/fish-shell/fish-shell/issues/1281#issuecomment-265788224), #2583 also related.
> 
> @zanchey is 4961 a typo? It doesn’t seem related?

Fixed, ta

src/exec.cpp
faho: Why don't we just replace the check in the old code with the thompson check and possibly change /bin/sh to _PATH_BSHELL (if that matters)?

ridiculousfish:
Merged as 0048730a67a5e70cafce1fb725a4b28001d924ac, added a test and replaced the `_PATH_BSHELL` trickery in eb71e4555f29948573be4d5f4da764e70fc5e85e. Thank you!

mqudsi:
I would actually like to change this to add an exception to refuse execution of .fish files without a shebang as a) we are not posix compatible anyway, b) it is almost certainly an error to execute them with sh.

ridiculousfish:
Great idea, I will add the exception.

mqudsi:
Thanks. The case @faho mentioned has been gnawing at me.

jart:
Thank you!

Issue #7710, opened 2021-02-12
Tests fail on latest commit


MithicSpirit:
Trying to build the latest commit (`056153eddd008233af27904c6ec54320da796882`) on Archlinux (using the `fish-git` AUR package) and I'm getting errors on the tests. Here is the output of the tests:

<details>
<summary>Logs</summary>

```
Testing indents
Testing utf8
Testing future feature flags
Testing escape_sequences
Testing escaping strings as pcre2 literals
Testing LRU cache
Testing parameter expansion
Testing overflowing expansions
Testing fuzzy string matching
Testing ifind
Testing ifind with fuzzy logic
Testing abbreviations
Testing test builtin
Testing fish_wcstod
Testing path functions
Testing pager navigation
Testing pager layout
Testing word motion
Testing is_potential_path
Testing colors
Testing complete
Testing autoload
Testing input
Testing line iterator
Testing undo/redo setting and restoring text and cursor position.
Testing undoing coalesced edits.
Testing universal variables
Testing universal variable output
Testing universal variable parsing
Testing universal variable legacy parsing
Testing universal callbacks
Testing universal format detection
Testing universal Ok to save
Testing universal notifiers with strategy 2
Testing completion insertions
Testing scenarios that should produce no autosuggestions
Testing autosuggestion combining
Testing history
Testing history merge
Error: Test failed on line 4283: history_contains(everything, text)
Error: Test failed on line 4283: history_contains(everything, text)
Error: Test failed on line 4283: history_contains(everything, text)
Error: Test failed on line 4308: history_contains(hists[i], texts[j])
Error: Test failed on line 4308: history_contains(hists[i], texts[j])
Error: Test failed on line 4308: history_contains(hists[i], texts[j])
Error: Test failed on line 4308: history_contains(hists[i], texts[j])
Error: Test failed on line 4308: history_contains(hists[i], texts[j])
Error: Test failed on line 4308: history_contains(hists[i], texts[j])
Error: Test failed on line 4308: history_contains(hists[i], texts[j])
Error: Test failed on line 4308: history_contains(hists[i], texts[j])
Error: Test failed on line 4308: history_contains(hists[i], texts[j])
Error: Test failed on line 4329: history_contains(reader, more_texts[j])
Error: Test failed on line 4329: history_contains(reader, more_texts[j])
Error: Test failed on line 4329: history_contains(reader, more_texts[j])
Error: Test failed on line 4329: history_contains(reader, more_texts[j])
[...]

Issue #7484, opened 2020-11-15
History race condition tests are failing


faho:
On Github Actions:

https://github.com/fish-shell/fish-shell/runs/1395993326
https://github.com/fish-shell/fish-shell/runs/1396163083
https://github.com/fish-shell/fish-shell/runs/1400100356

Example output:

```
Testing history race conditions
Error: Item dropped from history: 8511 173
Error: Item dropped from history: 8509 107
Error: Item dropped from history: 8510 109
Error: Item dropped from history: 8511 73
Error: Item dropped from history: 8511 72
Error: Item dropped from history: 8511 71
Error: Item dropped from history: 8511 70
[...]

Error: Item dropped from history: 8511 48
Error: Expected 1025 items, but instead got 993 items
Error: Line '8509 0' still left in the array
Error: Line '8510 0' still left in the array
Error: Line '8511 0' still left in the array
Testing history_sample_fish_1_x
```

Weirdly enough this seems to happen more on the 32bit machine?

I'm unsure if we should just give the test more slack, because it's meant to test race conditions, so if it fails that might actually show a problem.

Issue #6477, opened 2020-01-06
history merge test fails on OpenBSD


faho:
Again, on sr.ht (https://builds.sr.ht/~faho/job/133042), this time with OpenBSD:

```

Testing history merge
Error: Test failed on line 3889: history_contains(reader, more_texts[j])
Error: Test failed on line 3889: history_contains(reader, more_texts[j])
Error: Test failed on line 3889: history_contains(reader, more_texts[j])
Error: Test failed on line 3889: history_contains(reader, more_texts[j])
Error: Test failed on line 3889: history_contains(reader, more_texts[j])
Error: Test failed on line 3889: history_contains(reader, more_texts[j])
Error: Test failed on line 3889: history_contains(reader, more_texts[j])
Error: Test failed on line 3889: history_contains(reader, more_texts[j])
Error: Test failed on line 3889: history_contains(reader, more_texts[j])
Error: Test failed on line 3889: history_contains(reader, more_texts[j])
Error: Test failed on line 3889: history_contains(reader, more_texts[j])
Error: Test failed on line 3889: history_contains(reader, more_texts[j])
Error: Test failed on line 3889: history_contains(reader, more_texts[j])
Error: Test failed on line 3889: history_contains(reader, more_texts[j])
Error: Test failed on line 3889: history_contains(reader, more_texts[j])
Testing history race conditions
Error: Item dropped from history: 90318 113
Error: Item dropped from history: 65161 38
Error: Item dropped from history: 65161 37
Error: Item dropped from history: 65161 36
Error: Item dropped from history: 65161 35
Error: Item dropped from history: 65161 34
Error: Item dropped from history: 65161 33
Error: Item dropped from history: 65161 32
Error: Item dropped from history: 65161 31
Error: Item dropped from history: 65161 30
Error: Item dropped from history: 65161 29
Error: Item dropped from history: 65161 28
```

and so on.

(again with 33a234e19, which is 484e590b7 without a rebase)

faho:
Supposed to be fixed by 3ed8a57bc59c931e890c60df38222f4532d09b66.

faho:
Yeah, seems fixed.

Issue #3825, opened 2017-02-08
history race failure


krader1961:
When I merged a simple code style cleanup the Travis macOS (OS X) tests failed with 

```
Testing history race conditions
Error: Item dropped from history: 7815 222
Error: Expected 2561 items, but instead got 2560 items
Testing history_sample_fish_1_x
```

The OS in this case is probably not relevant as a race condition failure is likely to occur on multiple platforms. The challenge is how to both increase the likelihood of triggering this failure and determine the root cause. Since I could not find any previous reports of this failure mode it is presumably rare.

ridiculousfish:
Yes, the history doesn't handle heavy contention exactly right. I have a fix for it that I've been holding onto. I'll merge it later today.

krader1961:
FWIW, the history race test fails with much greater frequency on my FreeBSD 12 VM than any other environment I use to test fish behavior. Out of eight test runs three failed this specific test.

faho:
We don't really see this often. If there is a specific race we can work on it, but I don't see the use in keeping this open.

PR #7807, opened 2021-03-09
Fix parallel build race condition for test targets

jrosdahl:
## Description

When executing `make test -jX` (with X > 1) to build and run tests in a build directory, there is a race condition between the `serial_test_low_level` target and the `test_prep` target (a dependency of `serial_test_fishscript` and `serial_test_interactive`).

As far as I can tell, these events happen in a serial build scenario (`make test` with the "Unix Makefiles" CMake generator):

1. The `fish_tests` binary is built and executed.
2. The `test_prep` target (a dependency of `serial_test_fishscript`) cleans up test directories.
3. Tests in `test.fish` are executed.

In a parallel build scenario, this often happens:

1. Build of the `fish_tests` binary is started.
2. The `test_prep` target cleans up test directories.
3. Build of the `fish_tests` binary is finished.
4. Execution of the `fish_tests` binary starts.
5. Execution of the `fish_tests` binary finishes.
6. Tests in `test.fish` are executed.

However, if building the `fish_tests` binary is fast enough but not instant (e.g when using ccache), this can happen:

1. Build of the `fish_tests` binary is started.
2. Build of the `fish_tests` binary is finished.
3. Execution of the `fish_tests` binary starts.
4. The `test_prep` target cleans up test directories.
5. `fish_tests` tests that depend on said test directories may, depending on timing, fail because they are wiped by `test_prep`.

Fix this by making `test_prep` a dependency of `serial_test_low_level` so that `test_prep` can't interfere with `fish_tests` execution.

I think that it's likely that this fixes/addresses the following bugs:

- #3825: history race failure
- #6477: history merge test fails on OpenBSD
- #7484: History race condition tests are failing
- #7710: Tests fail on latest commit

## TODOs:

    - No changes to fish usage.
    - No regressions fixed.
    - I could do that if I understood where to add a note. :slightly_smiling_face:

### DIFF
--- a/cmake/Tests.cmake
+++ b/cmake/Tests.cmake
@@ -113,6 +113,7 @@ endforeach(TESTTYPE)
 
 # Now add a dependency chain between the serial versions.
 # This ensures they run in order.
+add_dependencies(serial_test_low_level test_prep)
 add_dependencies(serial_test_fishscript serial_test_low_level)
 add_dependencies(serial_test_interactive serial_test_fishscript)
 


faho:
I doubt it fixes

>#6477: history merge test fails on OpenBSD

Because that one fails *every time*.

But the rest? Yeah, probably.

In fact let's just try it. Merged, thanks!

faho:
https://github.com/fish-shell/fish-shell/runs/2075025613 failed like #3825, so that's apparently not fixed.

Issue #7819, opened 2021-03-14
Translation lacks ": " in cdh


lapingenieur:
## Versions

fish : version 3.1.2
gettext : version 0.21

## My issue

When using `cdh` to print last visited directories, it automatically asks for a number/letter. With `$LANG` equals to C it prints : (enters 'a' then press enter)

```bash
~/.config $ cdh
 a  1)  ~
Select directory by letter or number: a
~/ $
```

But if `$LANG` is set to `fr_FR.UTF-8` (standard french value), it's a bit stranger : (enters 'a' then press enter)

```
~/.config $ cdh
 a  1)  ~
Sélectionner un dossier par lettre ou chiffrea
~/ $
```

=> There's no `:` and the answer is just after the output (no space). I tried to fix the problem myself : `cdh` calls a fish function `_` which is a kind of alias for `gettext fish` (gettext is a GNU program). Now it's gettext that causes the bug : it doesn't add `: ` to the end of the line.

However, I don't know how to correct it. I tried to search for this string in the source code (available from GNU savanah repos / gettext) but I didn't find neither the french or the english string.

This issue might not be at the right place here but I don't know where else to put it, so if anyone here knows how to correct this I'd thank him.

faho:
You'll want to look in https://github.com/fish-shell/fish-shell/blob/master/po/fr.po - that's *fish's* french translation.

> the source code (available from GNU savanah repos / gettext)

The ones in the GNU gettext repo are irrelevant.

lapingenieur:
Thanks for this quick answer ! I'll correct then pr it !

PR #7821, opened 2021-03-14
Added a ':' at the end of a french translation

lapingenieur:
## Description

Added a `:` at the end of a french translation ([here](https://github.com/lapingenieur/fish-shell/commit/f663324916476f4ae4d171a0a44f878d5e714be7))

Fixes issue #7819

### DIFF
--- a/po/fr.po
+++ b/po/fr.po
@@ -3556,7 +3556,7 @@ msgstr ""
 
 #: /tmp/fish/explicit/share/functions/cdh.fish:5
 msgid "Select directory by letter or number: "
-msgstr "Sélectionner un dossier par lettre ou chiffre"
+msgstr "Sélectionner un dossier par lettre ou chiffre : "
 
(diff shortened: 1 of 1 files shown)


po/fr.po
```diff
@@ -3556,7 +3556,7 @@ msgstr ""
 
 #: /tmp/fish/explicit/share/functions/cdh.fish:5
 msgid "Select directory by letter or number: "
-msgstr "Sélectionner un dossier par lettre ou chiffre"
+msgstr "Sélectionner un dossier par lettre ou chiffre : "
```
faho, line 3559: This now also has a space *before*. Are you sure that's correct?

lapingenieur: In french, we always write spaces after *and* before `:`. For us it's kinda special to see `!?:;` without beeing spaced, unlike in our strange language.

So yes it's intentionnal.

faho:
Merged, thanks!

PR #7826, opened 2021-03-15
funced : changed 'rm' to 'command rm'

lapingenieur:
## Description

If `rm` is aliased to `rm -i` (ask before deleting) then `rm` will always ask to delete the cache file after funced edited the file, which is anoying. With this change, it won't be asking anymore. (This file is meant to be deleted at the end of `funced` so it's not a problem at all)

## Tests

I made the exact same modifications in `funced` in my session and it worked perfectly.

### DIFF
--- a/share/functions/funced.fish
+++ b/share/functions/funced.fish
@@ -130,7 +130,7 @@ function funced --description 'Edit function definition'
     end
 
     set -l stat $status
-    rm $tmpname >/dev/null
+    command rm $tmpname >/dev/null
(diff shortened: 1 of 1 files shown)


faho:
Yeah, sure.

Note that you don't typically want to alias things with the same name. It's bad form and quite likely to cause problems like this.

(honestly, just use an abbreviation - I have `abbr rm 'rm -I'`, and it works as it should)

But in this case it's simple enough that we can just take the enhancement.

PR #7839, opened 2021-03-19
Write new completion file for cargo

PoignardAzur:
## Description

Updates completions for `cargo`.

Some of existing completions which did "smart" work have been kept unchanged, eg:

```fish
# $x is "build", "run", etc
complete -c cargo -x -n "__fish_seen_subcommand_from $x" -l bin -a "(cargo run --bin 2>&1 | string replace -rf '^\s+' '')"
```

(though from my experience so far I'm not sure most of them work)

All other "dumb" completions which just filled in a subcommand name have been re-generated using [`clap::gen_completions`](https://docs.rs/clap/2.33.3/clap/struct.App.html#method.gen_completions). The code used to generate them is in [this cargo PR](https://github.com/rust-lang/cargo/pull/9288).

I was worried that using a hybrid hand-written/tool-generated config file might be a bad idea, but there's precedent in the codebase (see [`gcc.fish`](https://github.com/fish-shell/fish-shell/blob/d79f864072aa7264a5ab275fa5aefc64bbbe34da/share/completions/gcc.fish#L1-L3)).

I've done some quick testing in my shell to check that the new completions work.

## Why this PR?

The existing file does a lot of work to cover lots of commands, but it ultimately lets a few options slip.

For instance, `cargo check --tes` won't autocomplete now, even though `--test` is a valid option for `cargo check`. Using an auto-generated file makes sure that every option is documented.

Ideally, the generator should also complete cases like `cargo run --example=` by fetching example names, but `clap` doesn't implement that feature yet (and I'm told there's some non-trivial work blocking it).

## TODOs:


krobelus:
Great improvements, thank you!

PoignardAzur:
Huh. That was surprisingly straightforward.

Do I still push a commit with a changelog entry?

zanchey:
I generally don't add updates to completions to the changelog/release notes, because it becomes a bit overwhelming!

PoignardAzur:
Well I guess we're done then. Great!

krobelus:
Yeah! We'll bug you when we find problems with the generated completions ;)
BTW I forgot to mention: I ran `fish_indent` on your changes - it does things like removing some unneeded quotes.

Issue #4693, opened 2018-01-30
Function wrapping "cd" doesn't get correct completions


hamon-e:
If i write this function:
```
function cdtest -w cd
    cd $argv
end
```
and write `cdtest <TAB>` it doesn't autocomplete like cd does.
Is there something wrong with my function ?

```
fish --version
fish, version 2.7.1-708-gea897fcc
```

faho:
> Is there something wrong with my function ?

No. The issue here is that anything called `cd` itself gets special treatment by the completion machinery - see https://github.com/fish-shell/fish-shell/blob/master/src/complete.cpp#L1468.

This is supposed to go through the "wrap chain", but it seems like it's not doing that correctly.

jchilders:
@hamon-e FYI, I was able to create a `cd` wrapper using the following syntax:

```
function cd --wraps cd
  echo "Using our cd function"
  builtin cd $argv
end
```

This was on 2.7.1.

krobelus:
Fixd by #7843

PR #7843, opened 2021-03-20
Fix function wrapping for cd command

Caroleq:
## Fix function wrapping for cd command

As @faho mentioned in the issue discussion `cd` command is treated specially, but when it goes through wrapping chain it gets broken.  

Indicator if command should be treated specially for `cd` is stored in `handle_as_special_cd` flag which was true only if first command in chain was equal to `cd` and ignored commands visited in wrap chain.  This PR fixes the bug by checking if `cd` was visited during walking through wrapping chain - all visited commands (without the initial command) are stored in `arg_data.visited_wrapped_commands`.  

I also added test section for checking correct wrapping of `cd` in functions 

Fixes issue #4693


krobelus:
Great fix thanks!

I added a changelog entry and changed the map lookup to use `count` instead of `find`.

If `complete_t` would implement `operator==` we could simply use `do_test(cd_compl == cdwrap1_compl);` but I don't know if that's worth it since it's used nowhere else?

PR #7847, opened 2021-03-21
[vi-mode] corrected normal/default mode pasting

lapingenieur:
## Description

When using vi-mode, in normal/default mode `p` would put the wanted text just at the position of the cursor and `P` would paste one character before the curosr. But in `vi`, `vim` and other vi-based motions, `p` puts the text after the *next* character and `P` *at* the current character (or before the cursor).

I just changed

```fish
bind -s --preset p yank
bind -s --preset P backward-char yank
```

to : 

```fish
bind -s --preset p forward-char yank
bind -s --preset P yank
```

And did the same modifications for `"*p` and `"*P`

Tested, works even if at begining/end of the line with `p` as well as with `P`.


ridiculousfish:
Thank you!

PR #7851, opened 2021-03-22
Overhaul test integration in build system

mqudsi:
This is my second attempt at overhauling our varied test suites and unifying them under the ctest umbrella. Due to the limitations of the way ctest was hacked on to CMake, `test` remains a separate custom target, but nevertheless utilizes `ctest` under the hood.

The biggest change is that each individual test is run in a clean environment, created just for that test. And "test" here no longer means one of three (low level, pexpect, little check) but rather each test run by those test drivers. (Never again will you have to account for an alias that was added but not removed in a previous test).

This change also made it possible to run the tests in parallel (well, mostly), which lets me run the entire fish test suite, including the overhead of creating a new environment for each test, in 13 seconds. Tests continue to be run sequentially under CI.

An additional improvement is the addition of top-level targets for each test, meaning if test `foo` failed and you would like to retry just that test, `ninja -C build test_foo` will retry that test (regardless of which test suite it belongs to).

That said, the ability to run tests in parallel has shown some tests that are designed with global system state in mind - they need to be changed in order to pass when executed in parallel (or else can be forced to execute serially, but I think it's a good idea to clean up the test suite to make everything less brittle) - this applies mainly to `jobs.fish`. ~~The history merge test suite has issues that I believe might indicate a bug in either the creation of a new environment or what constitutes a unique fish session, the history race condition test continues to be extremely flaky and reports items dropped from history when the system is under heavy load when the test is executed - at this point, I suspect an actual issue.~~ Edit: later commits have changed the serialization level for history tests to prevent multiple history tests being run (serially) in one process (inadvertently in the case of this fork, by design in `master`); this seems to have resolved their flakiness.

The actual changes are fairly free of any hacks with one notable exception: cmake/ctest doesn't have a "discover tests" stage so as part of the "configure" step, ~~a bastardized version of `fish_tests` is built (without any dependencies on fish itself) and invoked to emit a list of tests - the alternative to this is for each low level test to be explicitly named in the CMake files (ugly) or each low level test to be moved to a separate source file (cringe).~~

### DIFF
--- a/build_tools/littlecheck.py
+++ b/build_tools/littlecheck.py
@@ -6,7 +6,6 @@ from __future__ import unicode_literals
 from __future__ import print_function
 
 import argparse
-from collections import deque
 import datetime
 import io
 import re
@@ -468,7 +467,6 @@ class TestRun(object):
             """
             return [s + "\n" for s in s.decode("utf-8").split("\n")]
 
-        PIPE = subprocess.PIPE
         if self.config.verbose:
             print(self.subbed_command)
         proc = runproc(self.subbed_command)
@@ -593,13 +591,13 @@ class Checker(object):
 
         # Find run commands.
         self.runcmds = [RunCmd.parse(sl) for sl in group1s(RUN_RE)]
+        self.shebang_cmd = None
         if not self.runcmds:
             # If no RUN command has been given, fall back to the shebang.
             if lines[0].text.startswith("#!"):
                 # Remove the "#!" at the beginning, and the newline at the end.
                 cmd = lines[0].text[2:-1]
-                if not find_command(cmd):
-                    raise CheckerError("Command could not be found: " + cmd)
+                self.shebang_cmd = cmd
                 self.runcmds = [RunCmd(cmd + " %s", lines[0])]
             else:
                 raise CheckerError("No runlines ('# RUN') found")
@@ -627,10 +625,12 @@ def check_file(input_file, name, subs, config, failure_handler):
         proc = runproc(
             perform_substitution(reqcmd.args, subs)
         )
-        stdout, stderr = proc.communicate()
-        status = proc.returncode
+        proc.communicate()
         if proc.returncode > 0:
             return SKIP
+
+    if checker.shebang_cmd is not None and not find_command(checker.shebang_cmd):
+        raise CheckerError("Command could not be found: " + checker.shebang_cmd)
 
     # Only then run the RUN lines.
     for runcmd in checker.runcmds:
@@ -698,13 +698,16 @@ def main():
     def_subs = {"%": "%"}
     def_subs.update(parse_subs(args.substitute))
 
-    failure_count = 0
+    tests_count = 0
+    failed = False
+    skip_count = 0
     config = Config()
     config.colorize = sys.stdout.isatty()
     config.progress = args.progress
     fields = config.colors()
 
     for path in args.file:
+        tests_count += 1
         fields["path"] = path
         if config.progress:
             print("Testing file {path} ... ".format(**fields), end="")
@@ -714,13 +717,14 @@ def main():
         starttime = datetime.datetime.now()
         ret = check_path(path, subs, config, TestFailure.print_message)
         if not ret:
-            failure_count += 1
+            failed = True
         elif config.progress:
             endtime = datetime.datetime.now()
             duration_ms = round((endtime - starttime).total_seconds() * 1000)
             reason = "ok"
             color = "{GREEN}"
             if ret is SKIP:
+                skip_count += 1
                 reason = "SKIPPED"
                 color = "{BLUE}"
             print(
@@ -728,7 +732,15 @@ def main():
                     duration=duration_ms, reason=reason, **fields
                 )
             )
-    sys.exit(failure_count)
+
+    # To facilitate integration with testing frameworks, use exit code 125 to indicate that all
+    # tests have been skipped (primarily for use when tests are run one at a time). Exit code 125 is
+    # used to indicate to automated `git bisect` runs that a revision has been skipped; we use it
+    # for the same reasons git does.
+    if skip_count > 0 and skip_count == tests_count:
+        sys.exit(125)
+
+    sys.exit(1 if failed else 0)
 
 
 if __name__ == "__main__":
--- a/cmake/Tests.cmake
+++ b/cmake/Tests.cmake
@@ -26,7 +26,7 @@ endif()
 
 # Set a policy so CMake stops complaining when we use the target name "test"
 cmake_policy(PUSH)
-if(${CMAKE_VERSION} VERSION_LESS 3.11.0 AND POLICY CMP0037)
+if(POLICY CMP0037)
   cmake_policy(SET CMP0037 OLD)
 endif()
 add_custom_target(test
--- a/cmake/Tests.cmake
+++ b/cmake/Tests.cmake
@@ -135,3 +135,14 @@ foreach(CHECK ${FISH_CHECKS})
   )
   add_test_target("${CHECK_NAME}")
 endforeach(CHECK)
+
+FILE(GLOB PEXPECTS CONFIGURE_DEPENDS ${CMAKE_SOURCE_DIR}/tests/pexpects/*.py)
+foreach(PEXPECT ${PEXPECTS})
+  get_filename_component(PEXPECT ${PEXPECT} NAME)
+  add_test(NAME ${PEXPECT}
+    COMMAND sh ${CMAKE_CURRENT_BINARY_DIR}/tests/test_driver.sh
+      ${CMAKE_CURRENT_BINARY_DIR}/tests/interactive.fish ${PEXPECT}
+    WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/tests
+  )
+  add_test_target("${PEXPECT}")
+endforeach(PEXPECT)
--- a/cmake/Tests.cmake
+++ b/cmake/Tests.cmake
@@ -1,3 +1,14 @@
+# This adds ctest support to the project
+enable_testing()
+
+# By default, ctest runs tests serially
+if(NOT CTEST_PARALLEL_LEVEL)
+  include(ProcessorCount)
+  ProcessorCount(CORES)
+  message("CTEST_PARALLEL_LEVEL ${CORES}")
+  set(CTEST_PARALLEL_LEVEL ${CORES})
+endif()
+
 # Define fish_tests.
 add_executable(fish_tests EXCLUDE_FROM_ALL
                src/fish_tests.cpp)
@@ -5,6 +16,25 @@ fish_link_deps_and_sign(fish_tests)
 
 # The "test" directory.
 set(TEST_DIR ${CMAKE_CURRENT_BINARY_DIR}/test)
+
+# HACK: CMake is a configuration tool, not a build system. However, until CMake adds a way to
+# dynamically discover tests, our options are either this or resorting to sed/awk to parse the low
+# level tests source file to get the list of individual tests. Or to split each test into its own
+# source file.
+execute_process(
+  COMMAND ${CMAKE_CXX_COMPILER} ${CMAKE_SOURCE_DIR}/src/fish_tests.cpp
+                            -o ${CMAKE_BINARY_DIR}/fish_tests_list
+                            -I ${CMAKE_CURRENT_BINARY_DIR} # for config.h
+                            -lpthread
+                            # Strip actual dependency on fish code
+                            -Wl,-undefined,dynamic_lookup,--unresolved-symbols=ignore-all
+  WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+)
+execute_process(
+  COMMAND ./fish_tests_list --list
+  OUTPUT_FILE low_level_tests.txt
+  WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+)
 
 # The directory into which fish is installed.
 set(TEST_INSTALL_DIR ${TEST_DIR}/buildroot)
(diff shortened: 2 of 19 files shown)


ridiculousfish:
Thanks for tackling this. I'm very very happy to have real isolated tests and also the ability to run individual tests.

@mqudsi can you say a bit about why you chose CTest? I don't know much about it - is it actually good, or just convenient? I see the configure-time test discovery step, I worry about that since it means you have to reconfigure to pick up new tests.

mqudsi:
The primary reason for going with ctest is to avoid any additional dependencies or further complications to our build system (I feel like they're all ugly, so might as well stick with what we have).

CTest has integration for Google Test if we ever want to formally adopt a framework. Looking through the code for that, I have an idea for how we can avoid the configure-time discovery for the low-level tests suite that I might try my hand at (all other tests discovered via globbing do not require a manual reconfigure to be picked up, as the addition of any files matching the glob pattern for the little check and pexpect tests will automatically trigger a CMake run).

On a more positive note, as of last night all the tests are passing. The build driver was fine, but some of the tests had shared state via the `../test/temp` folder that I patched out, and `jobs.fish` was further isolated to prevent asserting anything about the global state of pre-existing or simultaneously created zombies in the session. macOS runs the tests in a different order, which threw me for a loop because I thought it was an OS-specific issue that was causing them to fail when it turned out to be state (or lack thereof) peristing from a previous test.

krobelus (commented):
Running individual `littlecheck` and `pexpect` tests is a great improvement.
With `fish_tests` that already works fairly well, so if test discovery causes problems we could get by without that. Though a better integration is surely nice here as well.

A detail: the individual tests are named just by the basename, like `abbr.fish`. I'd prefer something like `test/abbr.fish` so they are all grouped. I noticed this in the tab-completion results of `ninja` that are  sorted alphabetically by fish, which makes sense since `ninja -t targets` does not give a consistent order.

tests/checks/jobs.fish
```diff
@@ -1,15 +1,5 @@
 #RUN: %fish %s
 
-# # Ensure there's no zombies before we start, otherwise the tests will mysteriously fail.
-# set zombies_among_us (ps -o stat | string match 'Z*' | count)
-# [ "$zombies_among_us" -eq "0" ]
```
krobelus: `zombies_among_us` had a nice vibe :)

krobelus:
I noticed is that "unsupported" littlecheck tests are listed as "passed" with ctest while littlecheck would print

image

 Of course that's a really minor detail since we only have like 2 tests with `REQUIRES`.

krobelus:
I noticed is that "unsupported" littlecheck tests are listed as "passed" with ctest while littlecheck would print

image

 Of course that's a really minor detail since we only have like 2 tests with `REQUIRES`.

cmake/Tests.cmake
```diff
@@ -6,6 +17,25 @@ fish_link_deps_and_sign(fish_tests)
 # The "test" directory.
 set(TEST_DIR ${CMAKE_CURRENT_BINARY_DIR}/test)
 
+# HACK: CMake is a configuration tool, not a build system. However, until CMake adds a way to
+# dynamically discover tests, our options are either this or resorting to sed/awk to parse the low
+# level tests source file to get the list of individual tests. Or to split each test into its own
+# source file.
+execute_process(
```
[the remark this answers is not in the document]
mqudsi: I developed this branch on Linux, compiling with gcc but linking with lldb.  What’s your toolchain?

tests/test_driver.sh
```diff
@@ -0,0 +1,113 @@
...
+        "$fish_script" "$script_args"
+    test_status="$?"
+
+    echo test completed in $homedir with status $test_status
+
+    rm -rf "$homedir"
+    exit "$test_status"
+else
+    echo "Recursive test invocation detected!" 1>&2
+    exit 10
```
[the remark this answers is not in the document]
mqudsi: > I noticed is that "unsupported" littlecheck tests are listed as "passed" with ctest while littlecheck would print
> 
> image
> 
> Of course that's a really minor detail since we only have like 2 tests with `REQUIRES`.

If little check returns a specific error code, we can program that to be a “skipped” marker and have ctest act accordingly. I’m guessing it’s just returning 0 here.

mqudsi:
> I noticed is that "unsupported" littlecheck tests are listed as "passed" with ctest while littlecheck would print
> 
> image
> 
> Of course that's a really minor detail since we only have like 2 tests with `REQUIRES`.

If little check returns a specific error code, we can program that to be a “skipped” marker and have ctest act accordingly. I’m guessing it’s just returning 0 here.

cmake/Tests.cmake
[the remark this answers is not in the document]
krobelus: Linux + gcc (same with clang) + GNU ld

krobelus:
> If little check returns a specific error code, we can program that to be a “skipped” marker

Ok, let's use exit code 125 since git bisect uses the same? Anything other than 0 or 1 should work. We can change littlecheck's behavior here since the `REQUIRE` syntax is fairly new.

faho:
The "wip (do not merge)" label seems superfluous, that's what PR drafts are for - they block merging *in the UI*, so I've taken the liberty of converting this to one.

mqudsi:
> A detail: the individual tests are named just by the basename, like `abbr.fish`. I'd prefer something like `test/abbr.fish` so they are all grouped. I noticed this in the tab-completion results of `ninja` that are sorted alphabetically by fish, which makes sense since `ninja -t targets` does not give a consistent order.

Unfortunately, CMake target names may not contain a slash:

```
CMake Error at cmake/Tests.cmake:101 (add_custom_target):
  add_custom_target called with invalid target name "tests/wait.py".  Target
  names may not contain a slash.  Use ADD_CUSTOM_COMMAND to generate files.
Call Stack (most recent call first):
  cmake/Tests.cmake:155 (add_test_target)
  CMakeLists.txt:207 (include)
```

krobelus:
> Unfortunately, CMake target names may not contain a slash:

Okay, then maybe `test-wait.py` as the next best thing?

ridiculousfish:
@mqudsi do you plan to pick this up again near term? If not I'd be eager to take it over, I'm very excited for this improvement.

mqudsi:
@ridiculousfish, thanks for the ping. I needed some time away from things and when I started dipping my toes back in the water I was too worried about how long I'd abandoned this and how out-of-date it must be by now, which wasn't exactly the kind of motivation I needed. Your enthusiasm, however, is quite inspire. Let me blow the dust off this and see what the deal is; it was already very close to ready iirc, so hopefully it won't be that bad.

mqudsi:
In addition to all the rebasing, I think I've addressed all the issues we've previously talked about including dropping the TMPDIR references from the tests and forgoing the dark magic involved in getting dynamic test discovery for low-level tests working. The tests also have a `test_` prefix when invoked via `ninja` or `cmake` directly, as requested by @krobelus, to make grouping/filtering/ignoring in completions (mentally or actually) easier.

tests/test_driver.sh
```diff
@@ -48,6 +48,12 @@ if test -z "$__fish_is_running_tests"; then
     export XDG_CONFIG_HOME
     mkdir -p $XDG_CONFIG_HOME/fish || die
 
+    # Create a temp/scratch directory for tests to use, if they want (tests shouldn't write to a
+    # shared temp folder).
+    TMPDIR="$homedir/temp"
```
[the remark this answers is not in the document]
krobelus: > # RUN: env TMPDIR=%tmpdir %fish %s

Yeah, or just run `set -l tmpdir (mktemp -d)` in the those tests, which also uses TMPDIR.

krobelus:
Maybe let's drop the changes to fish_tests.cpp for now.
They are not so important because to fish_tests binary already works well,
and it's pretty fast.
Those changes should be fairly independent of the rest, so dropping them
makes it easier to review this (my preprocessor-macro suggestion feels wrong)

The main part for me is that we can run individual isolated littlecheck
tests.  For that I've been using a script that mimics the setup done by
`test_fishscript`.  Happy to finally get rid of that.

I find this a bit hard to review because some things are added in earlier
commits, and removed in later ones.

I think the changes to tests/checks/tmux-complete.fish are no longer
necessary and can be dropped. (Or maybe they are only necessary in intermediate states?)
We still want to be able to run tests with littlecheck, so the `mktemp` calls are still good, and removal of the temp dirs is also nice.

Anyway, most of that is not  too important.  We can decide on whether to keep or
drop the fish_tests.cpp changes and then clean up post-merge.

tests/test_driver.sh
```diff
@@ -0,0 +1,130 @@
...
+}
+
+# To keep things sane and to make error messages comprehensible, do not use relative paths anywhere
+# in this script. Instead, make all paths relative to one of these or $homedir."
+TESTS_ROOT="$(dirname "$0")"
+BUILD_ROOT="${TESTS_ROOT}/.."
+FISH_ROOT="${BUILD_ROOT}/.."
+
+# macOS (still) doesn't have `readlink -f` or `realpath`. It's OK, this is just for aesthetics.
+if which realpath 1>/dev/null 2>/dev/null; then
```
faho, line 29: `command -v`, please. `which` is unnecessary cruft and may not be installed itself.

tests/test_driver.sh
```diff
@@ -0,0 +1,130 @@
...
+    # Set locale information for consistent tests. Fish should work with a lot of locales but the
+    # tests assume an english UTF-8 locale unless they explicitly override this default. We do not
+    # want the users locale to affect the tests since they might, for example, change the wording of
+    # logged messages.
+    #
+    # TODO: set LANG to en_US.UTF-8 so we test the locale message conversions (i.e., gettext).
+    unset LANGUAGE
+    # Remove "LC_" env vars from the test environment
+    for kv in $(env); do
+        if echo "$kv" | grep -Eq "^LC_"; then
```
faho, line 82: This executes one grep per line - why not `$(env | grep)` above?

(yes, technically variables can contain newlines, but then you can't use `env` to begin with)

faho: Alternatively we could just hardcode the LC_ variables we know:

LANG
LC_CTYPE
LC_NUMERIC
LC_TIME
LC_COLLATE
LC_MONETARY
LC_MESSAGES
LC_PAPER
LC_NAME
LC_ADDRESS
LC_TELEPHONE
LC_MEASUREMENT
LC_IDENTIFICATION
LC_ALL

I don't think anyone adds anything more to that silly broken system?

build_tools/littlecheck.py
```diff
@@ -6,7 +6,6 @@
 from __future__ import print_function
 
 import argparse
-from collections import deque
```
faho, line 9: I think all or most changes to littlecheck are upstream and should go away on a rebase?

cmake/Tests.cmake
```diff
@@ -1,11 +1,67 @@
...
+# We will use 125 as a reserved exit code to indicate that a test has been skipped, i.e. it did not
+# pass but it should not be considered a failed test run, either.
+set(SKIP_RETURN_CODE 125)
+
+# Even though we are using CMake's ctest for testing, we still define our own `make test` target
+# rather than use its default for many reasons:
+#  * CMake doesn't run tests in-proc or even add each tests as an individual node in the ninja
+#    dependency tree, instead it just bundles all tests into a target called `test` that always just
+#    shells out to `ctest`, so there are no build-related benefits to not doing that ourselves.
+#  * CMake devs insist that it is appropriate for `make test` to never depend on `make all`, i.e.
```
faho, line 20: Wow.

src/fish_tests.cpp
```diff
@@ -4328,6 +4329,11 @@ void history_tests_t::test_history_races_pound_on_history(size_t item_count, siz
 }
 
 void history_tests_t::test_history_races() {
+    // This always fails under WSL
```
faho, line 4332: Pretty sure this one typically works for me, WSL 1, provided I do it on the linux filesystem. Fails a ton on OpenBSD, tho.

On the windows filesystem the uvar signalling fails?

tests/test_driver.sh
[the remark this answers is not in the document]
mqudsi: Yes, I left tests that create their own temporary (via mktemp) directory as-is. Only tests that assumed it was OK to write to the shared tests directory or that relied on TEMPDIR or similar were modified to use the temporary `$PWD` instead. I was expecting breakage when executing tests in a random dir as `$PWD` but was pleasantly surprised to see no relative references were used and thus broken by that switch.

cmake/Tests.cmake
```diff
@@ -6,6 +17,25 @@ fish_link_deps_and_sign(fish_tests)
...
+execute_process(
+  COMMAND ${CMAKE_CXX_COMPILER} ${CMAKE_SOURCE_DIR}/src/fish_tests.cpp
+                            -o ${CMAKE_BINARY_DIR}/fish_tests_list
+                            -I ${CMAKE_CURRENT_BINARY_DIR} # for config.h
+                            -lpthread
+                            # Strip actual dependency on fish code
+                            -Wl,-undefined,dynamic_lookup,--unresolved-symbols=ignore-all
+  WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+)
+execute_process(
```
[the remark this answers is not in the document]
mqudsi: The hack has been removed and CMake now uses text preprocessing combined with a no-op macro in `fish_tests.cpp` to detect low-level tests. Complexity is considerably reduced and imho now at an acceptable level. I am strongly in favor of making this new system holistic, and believe that all test atomics should be runnable via a unified harness/interface.

tests/test_driver.sh
mqudsi: Sure.

build_tools/littlecheck.py
mqudsi: Yes, I believe so. I upstreamed all my littlecheck changes before using them in this branch. I might have botched the rebase, I'm not sure.

src/fish_tests.cpp
mqudsi: On my WSL machine, the fish source code is in a shared NTFS directory but are you quite sure you're actually running the `history_races` test under WSL? This commit was just re-adding the patch that was dropped in the cleanup of `fish_tests.cpp` in an earlier commit in this branch; I actually (silently) patched out `history_races` under WSL (v1, at least) years ago so there's a good chance you're not actually running it under WSL at all.

tests/test_driver.sh
mqudsi: Thanks. I remember having a particular stupid moment with that bit of code, for some reason I was doing something incredibly convoluted based off of trying to port the previous logic line-by-line from fish and not accounting for the differences between `set` and `env`. The entire thing (including both `grep` invocations) can be executed just once.

build_tools/littlecheck.py
mqudsi: Oh, it's `master` is on a particularly old version of `littlecheck.py`. The `ctest` branch is on a version in between that and littlecheck's `master`. I'll just merge in the latest.

mqudsi:
@krobelus for reviewing, you can actually just completely ignore the changes in the first commit; I'll try to rebase around them but if it's too messy I'm not going to spend too long on that endeavor. I mentioned in another comment but I feel pretty strongly about making all atomic tests runnable via a unified test harness and interface, especially since the complexity it requires out of the build system has been significantly reduced. If it makes you feel any better about the preprocessor suggestion, that was actually my fallback from the start but in hoping to avoid that minor bit of cruft, I ended up with that crazy workaround that I'm not sorry to see go.

mqudsi:
I've rebased again and tried to make the commit history more linear by dropping commits that introduced changes that were later reverted. 

I sporadically see this CI error:

> Testing file checks/psub.fish ... Failure:
> 
>   The CHECK on line 24 wants:
>     foo
> 
>   but there was no remaining output to match.
> 
>   additional output on stderr:1:3:
>     mktemp: X's may only occur at the end of the template
>     usage: mktemp [-d] [-t] [template]
>     note: mktemp is a test function, see tests/test_functions/mktemp.fish
> 
>   Context:
>      <= nothing to match CHECK 'foo' on line 24
>      <= nothing to match CHECK 'bar' on line 25
>      <= nothing to match CHECK 'baz' on line 26
>      <= nothing to match CHECK 'psub file was deleted' on line 34
>      <= nothing to match CHECK 'psub filename ends with .cc' on line 42
>      <= nothing to match CHECK 'psub filename ends with .cc' on line 50
>      <= nothing to match CHECK 'psub directory was deleted' on line 58
> 
>   when running command:
>     ../test/root/bin/fish checks/psub.fish
> 

but `psub.fish` does not call `mktemp` :shrug:

tests/test_driver.sh
```diff
@@ -0,0 +1,128 @@
...
+        printf "%s\n" "$@" 1>&2
+    fi
+    exit 1
+}
+
+# To keep things sane and to make error messages comprehensible, do not use relative paths anywhere
+# in this script. Instead, make all paths relative to one of these or $homedir."
+TESTS_ROOT="$(dirname "$0")"
+BUILD_ROOT="${TESTS_ROOT}/.."
+FISH_ROOT="${BUILD_ROOT}/.."
```
ridiculousfish, line 26: I think this is assuming that the build directory is directly inside the fish source directory, which is often not the case. You can reproduce by making a separate directory and running `cmake /path/to/fish`; the tests will fail.

mqudsi: Do you know if I have to explicitly pass these as command line arguments or environment variables from CMake on invocation or if there's something already set I can use?

mqudsi: nvm. We are (I am?) already passing it as a command line argument, just not using it in the sh script.

mqudsi:
Aside from the issue of littlecheck catching the stderr too late (or something), there appears to have been a bug in the mktemp wrapper function for tests in that it was too sensitive to the presence of an X in the template string. In particular, it was randomly firing when we were calling `mktemp` with a template that contained a directory (e.g. PWD) that was also created via `mktemp` and happened to contain a sequence of `X` as a result. This is patched in the most recent commit.

tests/test_driver.sh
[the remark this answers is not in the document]
mqudsi: Done.

faho:
>but psub.fish does not call mktemp 🤷

`psub` itself does! https://github.com/fish-shell/fish-shell/blob/60a9dcbf0548e0f3445639b2753a9cbd11e8bba7/share/functions/psub.fish#L28-L41

mqudsi:
Thank you - mystery solved! TMPDIR is randomly created by the test driver and then used as part of the template, and it would randomly trigger the overzealous template protection in the test-specific mktemp (now fixed) when it contained something that looked like XX.

src/fish_tests.cpp
faho: Ah, okay. That's it - it doesn't run the history_races test. I missed that because it doesn't say it's been skipped. It just doesn't mention it at all.

mqudsi:
@zanchey @faho @krobelus @ridiculousfish any objections to merging this? It's been weighing on me and I am loathe to keep rebasing.

faho:
>any objections to merging this?

As far as I can tell all the objections have been dealt with and the tests succeed (and aren't much flakier than what we already have).

So I see no reason to hold this back any longer. Merge if you want!

zanchey:
:shipit:

krobelus (approved):
Sorry for the delay - this looks awesome, let's merge.

I was confused for a bit because `ninja test_basic.fish` would take 4 seconds but that's just because I had `BUILD_DOCS=ON` which runs sphinx

tests/checks/features-qmark1.fish
```diff
@@ -1,2 +1,2 @@
-#RUN: env HOME="$(mktemp -d)" XDG_CONFIG_HOME="$(mktemp -d)" %fish --features '' -c 'string match --quiet "??" ab ; echo "qmarkon: $status"'
+#RUN: env %fish --features '' -c 'string match --quiet "??" ab ; echo "qmarkon: $status"'
```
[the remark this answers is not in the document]
mqudsi: I've added that back in but I think making it conceptually easier to create new tests without having to worry about these things is one of the greatest things about using the test driver.

tests/checks/basic.fish
```diff
@@ -42,12 +42,12 @@ end
 # Simple function tests
 
 function foo
-    echo >$TMPDIR/fish_foo.txt $argv
+    echo >./fish_foo.txt $argv
```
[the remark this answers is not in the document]
mqudsi: Sure. Done.

mqudsi:
🎉

thank you, all.

tests/checks/features-qmark1.fish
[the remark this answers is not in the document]
krobelus: Thanks - yeah, I'm probably being overly paranoid here.
If manual invocations ever cause confusion, we could also explicitly prevent them in future.

zanchey:
There are two bits of fallout in the package builders from this merge, neither of which are showstoppers.

It looks like the overrides from 4c2a0de5de have been moved into the test driver, which means the low-level tests go back to running in the user's home directory. This is a bit of an issue on Debian pbuilders which use a non-existent HOME - see d518b01281dfab as well.

Second, old CMake versions call skipped tests "failed". I can't see that this was changed explicitly - it's broken in 3.7 but working in 3.10. This only affects the package builds on LTS version of Debian, so I'll try and chase it down at some stage but it's not a blocker by any means.

zanchey:
> it's broken in 3.7

[  327s] 
[  327s] The following tests FAILED:
[  327s] 	130 - git.fish (Not Run)
[  327s] 	184 - tmux-bind.fish (Not Run)
[  327s] 	185 - tmux-complete.fish (Not Run)
[  327s] 	186 - tmux-prompt.fish (Not Run)
[  327s] Errors while running CTest

> but working in 3.10

The following tests did not run:
	130 - git.fish (Skipped)
	184 - tmux-bind.fish (Skipped)
	185 - tmux-complete.fish (Skipped)
	186 - tmux-prompt.fish (Skipped)

faho:
>Second, old CMake versions call skipped tests "failed". I can't see that this was changed explicitly - it's broken in 3.7 but working in 3.10

Lovely. The alternative here would be to "pass" these tests on our end, maybe even just if cmake < 3.10. It's not like skipped tests are super visible anyway.

zanchey:
> The overrides have indeed been moved into the test driver _but_ at the same time the low-level tests are now executed via the test driver

No, the low-tests are run directly - the test driver always invokes fish first.

```cmake
foreach(LTEST ${LOW_LEVEL_TESTS})
  add_test(
    NAME ${LTEST}
    COMMAND ${CMAKE_BINARY_DIR}/fish_tests ${LTEST}
    WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
  )
  set_tests_properties(${LTEST} PROPERTIES SKIP_RETURN_CODE ${SKIP_RETURN_CODE})
  add_test_target("${LTEST}")
endforeach(LTEST)
```

> I thought maybe the feature to support skipping a test based on the return code was only introduced post CMake 3.7 [but it seems to have been supported since at least version 3.0](https://cmake.org/cmake/help/v3.0/prop_test/SKIP_RETURN_CODE.html)!

They were counted as failures prior to 3.9 (see https://gitlab.kitware.com/cmake/cmake/-/issues/16822).

> Side note: I can't believing viewing the documentation for a CMake API/feature in the latest documentation doesn't include a "introduced in CMake version xxx" or something

Heh. fish's documentation isn't great in that sense either.

mqudsi:
@zanchey  can you please try the ctest_redux branch from  mqudsi/fish-shell? It should fix the debian problem.

zanchey:
@mqudsi, it works great!

zanchey:
Thanks for merging that branch in 93aaa1851e1c9190e6d06e2c5f0d26e434bbd502. Unfortunately that uncovered a new issue in Ubuntu Xenial, which uses CMake 3.5 (which I didn't test until after the merge). The `test` target generate in 3.5 doesn't depend on `fish_tests`! Anyway I think I'll just hack around that in the package build.

mqudsi:
You're most welcome. I saw your patch - that's certainly one way to work around it!

PR #7858, opened 2021-03-23
Fixed completion errors for tshark when running as root

exploide:
## Description

When running as root, many `tshark` completions show the following error:

> Running as user "root" and group "root". This could be dangerous.

This can be fixed by redirecting stderr.

## TODOs:


krobelus:
Makes sense, thanks!

(Looks like the tmux tests are flaky, I'll try to increase the timeout..)

PR #7860, opened 2021-03-23
Add max and min math functions

### LINKED ISSUE #7856: Add max and min math functions
It's not super common, but I've found myself wanting these functions to set upper and lower bounds on things.

Obviously this is just a nice tool to have, but as a specific use case, some `string` commands like `pad` and `repeat` will error if you give them negative numbers. Obviously we can't use `abs` here, so I've been doing `test $blah -lt 0 && set $blah 0`. This works of course, but it is less elegant and requires setting the value to a variable first.

IlanCosman:
## Description

Adds `max` and `min` math functions.

fixes #7856 

## TODOs:

### DIFF
--- a/src/tinyexpr.cpp
+++ b/src/tinyexpr.cpp
@@ -182,6 +182,20 @@ static constexpr double bit_xor(double a, double b) {
     return static_cast<double>(static_cast<long long>(a) ^ static_cast<long long>(b));
 }
 
+static double max(double a, double b) {
+    if (std::isnan(a)) return a;
+    if (std::isnan(b)) return b;
+    if (a == b) return std::signbit(a) ? b : a; // treat +0 as larger than -0
+    return a > b ? a : b;
+}
+
+static double min(double a, double b) {
+    if (std::isnan(a)) return a;
+    if (std::isnan(b)) return b;
+    if (a == b) return std::signbit(a) ? a : b; // treat -0 as smaller than +0
+    return a < b ? a : b;
+}
+
 static const te_builtin functions[] = {
     /* must be in alphabetical order */
     {L"abs", reinterpret_cast<void *>(static_cast<te_fun1>(fabs)), TE_FUNCTION1},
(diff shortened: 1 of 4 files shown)


IlanCosman:
@ridiculousfish Could you review? Fabian is on break and I'd like to spare him the trouble of fixing my terrible C++ 😄

IlanCosman:
Flake test I think?

src/tinyexpr.cpp
```diff
@@ -182,6 +182,10 @@ static constexpr double bit_xor(double a, double b) {
     return static_cast<double>(static_cast<long long>(a) ^ static_cast<long long>(b));
 }
 
+static double max(double a, double b) { return std::max(a,b); }
+
+static double min(double a, double b) { return std::min(a,b); }
```
krobelus, line 187: `clang-format` should add a space between arguments :detective:

krobelus:
> Flake test I think?

Yeah, sorry - I've disabled the test in 4f00ccfd1

src/tinyexpr.cpp
IlanCosman: My bad, I don't actually have `clang` on my machine 😄

ridiculousfish:
Are we OK with `min` and `max` which only support two arguments?  I suppose we can always extend it later.

Regarding the implementations, there's some floating point edge-cases, [we should just match Java](https://docs.oracle.com/javase/8/docs/api/java/lang/Math.html#max-double-double-).

1. Negative zero should be "less than" positive zero. So `math 'max(-0, 0)'` should print 0, and min should print -0.
2. `max` and `min` of anything with NaN should be NaN. `math 'max(1, 0/0)'` should error instead of printing 1.

Something like this should do it:

   double max(double a, double b) {
       if (std::isnan(a)) return a;
       if (std::isnan(b)) return b;
       if (a == b) return std::signbit(a) ? b : a; // treat +0 as larger than -0
       return a > b ? a : b;
   }

   double min(double a, double b) {
       if (std::isnan(a)) return a;
       if (std::isnan(b)) return b;
       if (a == b) return std::signbit(a) ? a : b; // treat -0 as smaller than +0
       return a < b ? a : b;
   }

IlanCosman:
Thanks @ridiculousfish! Yah I decided that two arguments is better than none at all, and in many (most?) cases two is all that is necessary.

Thanks for the changes, I'll add them 😄 Is there any reason that C++'s `max` and `min` don't do those things already?

ridiculousfish:
`std::max` and `std::min` are generic functions, they operate on anything comparable so they are ignorant of floating point's nonsense rules.

ridiculousfish:
Merged as c762c624647f6c188177c79f62608fb844f2ba8f with some fixes to Mac build. Thank you!!

PR #7861, opened 2021-03-24
Fix Sphinx warning about CHANGELOG.rst

mojotx:
When building `sphinx-manpages` and/or `sphinx-docs`, a warning is issued:

```
../CHANGELOG.rst:33: WARNING: Document or section may not begin with a transition.
```

## Description

This is almost identical to the fix in commit 84a89f5195a31b53c56cde644279b3e88232b601. 

Based on a previous discussion with @faho , I didn't bother with creating an issue, and I'm just making the pull request. 🙂

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -29,6 +29,9 @@ Improved prompts
 
 Improved terminal support
 ^^^^^^^^^^^^^^^^^^^^^^^^^
+- *This placeholder silences a sphinx doc warning, so the last section is not empty.*
+
+--------------
 
 fish 3.2.1 (released March 18, 2021)
 ====================================
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -30,8 +30,6 @@ Improved prompts
 Improved terminal support
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 
---------------
-
 fish 3.2.1 (released March 18, 2021)
 ====================================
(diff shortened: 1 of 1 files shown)


CHANGELOG.rst
```diff
@@ -30,8 +30,6 @@ Improved prompts
 Improved terminal support
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 
---------------
```
krobelus, line 33: I think we should keep this because it belongs to the changelog template - we have it between all releases.
It renders as horizontal line (`<hr>`) in HTML (in `build/user_doc/html/relnotes.html`, it took me a while to realize we renamed it).

As soon as we add some text to the "Improved terminal support" section above, the warning will go away.
Maybe there is a way to avoid or work around the warning?

mojotx: That's a good point. I did a test, and you are correct, adding ANYTHING to that last section silences that warning. 

Since the 3.3.0 section is a work-in-progress, perhaps add a "To-do" bullet-point, with the understanding that it's a placeholder for "real" info to be provided later? 

I just pushed another commit, a96bbb598, that does this. I think it looks good. What do you think? 

Screen Shot 2021-03-24 at 5 29 14 PM

zanchey:
Yes, that works better - thanks! Just have to remember to [REMOVE BEFORE FLIGHT].

PR #2755, opened 2016-02-22
Send BEL on completion error

floam:
This patch makes fish send a BEL character to trigger a visual flash or a beep (however they have their terminal setup) at the same time it tries to flash the prompt momentarily to indicate that hitting tab again won’t help when there are zero completion results. Sometimes I just don’t notice a short completion flashing like that, especially because it only does it on the first tap of the tab key. This is nicer.

PR #7875, opened 2021-03-28
fish_bell option to enable or disable bell on reader_t::flash().

ewtoombs:
`fish_bell`'s behaviour is documented in doc_src/language.rst.

## Description

The option was added because visual (and especially audio) feedback after
failed user interaction is usually unnecessary and can be extremely annoying.
It is only useful over bad networks, when it isn't clear whether the
interaction failed or there is network lag.

The terminal bell is still useful for other things, like notifying on the completion of a long command, so disabling the terminal bell entirely is not preferable to many users.

## TODOs:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -27,6 +27,7 @@ Interactive improvements
 - ``__fish_prepend_sudo`` now toggles sudo even when it took the commandline from history instead of only adding it.
 - Fish now defaults job-control to "full" meaning it more sensibly handles assigning the terminal and process groups (:issue:`5036`, :issue:`5832`, :issue:`7721`)
 - ``math`` learned two new functions, ``max`` and ``min`.
+- Terminal bells are now off by default. Enable them by setting ``fish_bell``.
 
 New or improved bindings
 ^^^^^^^^^^^^^^^^^^^^^^^^
--- a/doc_src/language.rst
+++ b/doc_src/language.rst
@@ -1196,6 +1196,8 @@ You can change the settings of fish by changing the values of certain variables.
 
 - ``fish_ambiguous_width`` controls the computed width of ambiguous-width characters. This should be set to 1 if your terminal renders these characters as single-width (typical), or 2 if double-width.
 
+- ``fish_bell``, if set, causes fish to emit BEL characters. This happens when a history search finds nothing, when no suitable completions are found, or when an undo or redo fails. Useful over bad network connections to distinguish failed user interaction from network lag.
+
 - ``fish_emoji_width`` controls whether fish assumes emoji render as 2 cells or 1 cell wide. This is necessary because the correct value changed from 1 to 2 in Unicode 9, and some terminals may not be aware. Set this if you see graphical glitching related to emoji (or other "special" characters). It should usually be auto-detected.
 
 - ``FISH_DEBUG`` and ``FISH_DEBUG_OUTPUT`` control what debug output fish generates and where it puts it, analogous to the ``--debug`` and ``--debug-output`` options. These have to be set on startup, via e.g. ``FISH_DEBUG='reader*' FISH_DEBUG_OUTPUT=/tmp/fishlog fish``.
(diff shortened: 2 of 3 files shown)


ridiculousfish:
Thank you for the PR. Instead of adding configuration, I think we should just remove the bell entirely - I doubt anyone will miss it. Does anyone like the bell and want to keep it?

faho:
Let's kill it.

faho:
Okay I just removed the bell entirely, it's not actually a good idea in this context.

ewtoombs:
Hm. Yep, works for me. Thanks, @faho and @ridiculousfish !

floam:
I l liked the bell, but ya snooze, ya lose.

ewtoombs:
> l liked the bell, (I added it) but ya snooze, ya lose.

I'd still like to know your rationale.

floam:
https://github.com/fish-shell/fish-shell/pull/2755

floam:
I just find a little audible ding or the terminal flashing with a visual bell a decent stimuli to let me know I'm about to attempt an invalid command.

floam:
Also, I really do not understand "It is only useful over bad networks".

ewtoombs:
Invalid keyboard shortcuts elsewhere in the operating system do not produce audible bells. The fact that nothing happens is all the stimulus a user needs to be able to tell something doesn't work. Anything extra is just spurious and annoying. Even the terminal flash is unnecessary and annoying, but not nearly as annoying as the bell.

A general rule of UI development is never to produce audible noises in direct response to keypresses. Modern design overwhelmingly follows this rule. This is not the way UIs have always been, and they are so much better for having made this change. Adding audible bells is a massive step backward from this advance.

ewtoombs:
This idea of preventing the running of invalid commands is new, though. The flash does correlate at least somewhat with a failed history search and when that happens, the command will be invalid. If somebody does a search assuming it will succeed, then immediately presses enter without checking the command, they can accidentally enter (potentially dangerous) nonsense into the terminal. The flash does steer people away from that. That is good. So, the flash is doing something useful. I can agree with that.

ewtoombs:
You may rightly ask why I don't just switch to a visual terminal bell. That's because I use the audible terminal bell to notify me when a long command is complete. It is very useful in this regard. I mentioned this before. I even have an alias for this:
```
alias bell='echo -n \a'
```
Example:
```
> rm -rf huge_freaking_folder; bell
```
It is so useful that it would actually be really cool if fish automatically rang the terminal bell at completion if a command has taken more than, say, 5 seconds and is still running. But I digest.

ewtoombs:
> Also, I really do not understand "It is only useful over bad networks".

Right, let me unpack this one lol. I can see that I didn't really explain this completely.

Suppose that we're in an alternate universe where fish notifies the user of failed history searches by doing nothing at all. So, no bell and no flash either.

Suppose you're in an SSH session and your network is total garbage.  You want to do `ps aux | grep nginx`. On a new line, you type `ps ^P`. Suppose the search will succeed, but nothing happens initially, due to network lag. You might now accidentally assume the search failed, and start typing out the full command. When the network kicks back in, the history search succeeds, and everything you have typed so far is inserted at the end of the line, after the correct command. Now, you are forced to erase what you have added to the end of the command or start over. From this experience, you have learned that you have to wait a few seconds after every history search to make sure it has really failed.

The flash solves this problem nicely, but I don't consider it a priority, because there are a whole bunch of other instances where you'd also have to add flashes in order for fish to be robust against bad networks—something fish hasn't done. So it's pretty clear to me that network lag is not a design consideration in fish. I actually think this is totally fine. Bad networks are so rare these days that designing for it doesn't really make sense any more.

faho:
@ewtoombs Could you take this elsewhere? I get notified for every message in here.

ewtoombs:
> @ewtoombs Could you take this elsewhere? I get notified for every message in here.

With an audible bell, or with a flash? :D

Right, @floam, you can /msg me on libera if you want to keep discussing this. My nick is triteraflops.

Sorry, @faho.

PR #7877, opened 2021-03-29
math: Make function parentheses optional

faho:
It's a bit annoying to use parentheses here because that requires
quoting or escaping.

This allows the parens to be omitted, so

```fish
math sin pi
```

is the same as

```fish
math 'sin(pi)'
```

Function calls have the lowest precedence, so

```fish
math sin 2 + 6
```

is the same as

```fish
math 'sin(2 + 6)'
```

## TODOs:


ridiculousfish:
Nice! This would benefit with tests for nested calls and mixed arities, e.g. `math 'sin cos 3'` and `math 'pow sin 3 5'`

PR #7880, opened 2021-03-30
Make default vi mode prompt a bit nicer to look at

faho:
## Description

The default vi mode prompt is kind of ugly, mostly because we include
this `[I]` with a super bright green background and white text,
which is particularly grating because most prompts don't actually have
a background.

So we get a ton of people asking "How do I remove this [I]" when they
could really benefit from having the mode shown.

There's a few ways to make this look nicer, the simplest is to just
keep the same colors but use them as foreground instead of background
colors, which looks much more understated.

The mode prompt is important, but not more than the actual contents of
the commandline, so it shouldn't have ALARMING colors.

Before:
Screenshot_20210330_171830

After:
Screenshot_20210330_171804
## TODOs:


ridiculousfish:
Nice, looks much better!

PR #7881, opened 2021-03-30
Fix Sphinx warning

mojotx:
## Description

For some reason, the space in seems to cause a problem. I verified that the generated HTML looks good. 

```
../CHANGELOG.rst:30: WARNING: Inline literal start-string without end-string.
```
Screen Shot 2021-03-30 at 10 59 47 AM


faho:
Oh, sphinx.

Yeah, thanks!

PR #7891, opened 2021-04-01
Shortens RPM completions

Phuurl:
## Description

Edits the longer `rpm` completions to be more succinct, and generally around ~60 characters long.

Work as part of #6981


ridiculousfish:
Merged as 13439b399e44861ac0082f636b418cecebe7af0b, thank you!

Issue #7884, opened 2021-03-31
Curate sample prompts


faho:
Our sample prompts are mostly a random grab bag of prompts someone wrote once.

We should see that the selection

- is distinct - there's no real reason for two prompts that just differ in a color or one bit of information
- is well-written
- uses our common facilities (e.g. fish_vcs_prompt) where desirable

Personally, I would:

- kick out the "justadollar" prompt because I don't think it serves most people well and is trivial to write yourself
- remove the "informative vcs" prompt because the informative git status is sloooooow - or make it, ironically, not "informative"
- remove two out of three "classic" prompts
- remove the "screen savvy" prompt because I have no idea what it's for - where is $WINDOW set?
- add debian chroot support to other prompts and remove that one
- rename the prompts named after a single person because some may have... "spicy" opinions that don't align with our CoC and we don't want to appear like we're condoning those

We might also want to add new prompts - what are y'all using? Anyone have a cool prompt they want to share?

IlanCosman:
Highly agreed, I was just thinking about this myself 👍 I think a very simple pure-style prompt would be great addition, though that does kinda step on third-party authors' toes 😄

faho:
Alright, here's a screenshot of what the prompts currently look like, in an actual terminal:

![Screenshot_20210402_220629](https://user-images.githubusercontent.com/5185367/113450216-bd7d9680-93ff-11eb-89b0-6808a0b7a143.png)
![Screenshot_20210402_220642](https://user-images.githubusercontent.com/5185367/113450221-be162d00-93ff-11eb-8694-ff10f7c8416f.png)


The first thing you'll notice here is that these are almost all variations on a very simple theme, basically:

```
user@host pwd status/vcs glyph
```

Some leave off one or more of these or add a little special thing (debian chroot has the debian chroot, "informative" has the time), but the bulk of these stick very closely to that idea.

In particular:

- "debian_chroot" is literally the classic prompt plus a debian chroot display that I'm not sure is actually used - you would need to specifically select it *in the chroot* (this is the sort of thing that's really only worth it as a default, and then it shouldn't just be for debian specifically - I've asked for this before and been met with resistance)

- classic status/vcs are just straight upgrades of each other - these could just be one prompt with easily removable bits
- I have no idea what lonetwin and screen savvy are
[...]

PR #6398, opened 2019-12-08
Add the machine "type" to the default prompt

faho:
This is a description of whether we're connected over ssh or in a
container or something.

It's used to signal to the user that this
particular terminal is "special", and not on the local machine, e.g.
to avoid mistakes like executing `poweroff` on a server instead of
your development laptop.

Unfortunately this (at least currently) isn't dependable enough
to *disable* display of the user@host part on "normal" machines, but
we can add something on ssh.

This is most useful to have if it is the default, because that means you don't have to set it up on every single remote machine you ever come into contact with.

See #6375.

## TODOs:

- [ ] Changes to fish usage are reflected in user documentation/manpages.
- [N/A] Tests have been added for regressions fixed
- [ ] User-visible changes noted in CHANGELOG.md
- [ ] Do we add this to classic+vcs, or make the default prompt yet another version of that?
- [ ] Are there more ways of detection to be added?
- [ ] Do we have to handle multiple stacked things? Ssh-on-container, is that important? Or is there some order of precedence here?
- [ ] Is there a better term than "machine type"?

PR #1351, opened 2014-03-21
Debian chroot environment in prompt

mdesantis:
When you chroot in Debian, bash shows the chroot environment in the prompt:

``` bash
# inside /etc/bash.bashrc

...

# set variable identifying the chroot you work in (used in the prompt below)
if [ -z "${debian_chroot:-}" ] && [ -r /etc/debian_chroot ]; then
    debian_chroot=$(cat /etc/debian_chroot)
fi

# set a fancy prompt (non-color, overwrite the one in /etc/profile)
PS1='${debian_chroot:+($debian_chroot)}\u@\h:\w\$ '

...
```

This is the effect:

```
(chroot_env)user@host:~#
```

It is useful when chrooting, since usually the hostname remains the same and thus you can't distinguish where you are.

My proposal has the same effect. I tested it and it works.

Feel free to discuss it, I made it on the fly.

PR #7897, opened 2021-04-03
Sample prompts

faho:
## Description

This is basically a first pass on the sample prompts.

It removes:

- classic and classic+status - they are just subsets of classic+vcs
- debian chroot - it's just classic plus a function that I don't think is used much (and if it is it should be added to the others as well)
- lonetwin and screen savvy because they are just slight variations on classic
- justadollar because it's unhelpful and easy to build yourself

and performs some light cleanup on the others. It renames the prompts named after people who did not contribute them directly to fish.

Work on #7884.

## TODOs:


faho:
Screenshot of all the prompts after this:

Screenshot_20210403_100641

We are now down to 10, from 16.

ridiculousfish:
Thanks for tackling this. Lots of good cleanup here.

I think we should have at least one simple prompt (not necessarily justadollar, but similarly simple). Some users may prefer a simple prompt, and don't want to learn how to build it; or they can use the simple prompt as a starting point to build from.

faho:
>I think we should have at least one simple prompt (not necessarily justadollar, but similarly simple)

Sure, but I would argue that minimalist is much better in that role. Let me quote the full source:

```fish
# name: Minimalist
# author: ridiculous_fish

function fish_prompt
    set_color $fish_color_cwd
    echo -n (basename $PWD)
    set_color normal
    echo -n ' ) '
end
```

My fear is that justadollar is confusing to people, because it doesn't even contain $PWD. It's also not good as a starting point because it doesn't even show that $PWD exists, or that `set_color` is a thing.

By contrast going from minimalist to something justadollar-like is just removing the first three (or two) lines, which should be intuitively understandable.

Not that I'm married to minimalist in particular - I don't love showing only the last component of $PWD. My preferred "very simple" prompt would be something like `echo -n (prompt_pwd) '$'`.

But I would quite like to remove justadollar because it's a disservice to those who pick it accidentally.

faho:
Alright, I renamed the Classic+VCS prompt to "Default" and added a "Simple" prompt that's the baseline I would expect - it's basically what you get when you look for a nice default $PS1 for bash:

```fish
function fish_prompt
    # This is a simple prompt. It looks like
    # alfa@nobby /path/to/dir $
    # with the path shortened and colored
    # and a "#" instead of a "$" when run as root.
    set -l symbol ' $ '
    set -l color $fish_color_cwd
    if fish_is_root_user
        set symbol ' # '
        set -q fish_color_cwd_root
        and set color $fish_color_cwd_root
    end

    echo -n $USER@$hostname

    set_color $color
    echo -n (prompt_pwd)
    set_color normal

    echo -n $symbol
end
```

Once you've seen this, you should be able to figure out how to make all sorts of prompts.

krobelus:
I appreciate quiet prompts; the `$ ` one is nice when sending terminal contents to someone else because people recognize the dollar as prompt, and other prompt information is usually not relevant. But I agree it need not be in our samples.

I'd shave off the `fish_is_root_user` part of the simple prompt, to make it even simpler.
Or is it normal for people to copy over their prompt to root as well?

faho:
>I'd shave off the fish_is_root_user part of the simple prompt, to make it even simpler.
Or is it normal for people to copy over their prompt to root as well?

This is what `\$` in $PS1 in bash does. The default $PS1 on arch is `[\u@\h \W]\$`, so it does this (and also the weird thing of only printing the basename for $PWD).

This is another one of those things that are most useful when they are the default - people don't tend to customize root much, but when they pick a prompt from `fish_config` it should still make clear that it is root.

We could probably get by with adding a helper function called `prompt_symbol`. Takes two args for the user symbol and the root symbol, and then calls `fish_is_root_user` and prints the corresponding symbol. That doesn't do the PWD color, however, but we could also leave it out for this prompt. Either leave it at $fish_color_cwd or don't color it, or hard-code a different color.

zanchey:
https://github.com/fish-shell/fish-shell/pull/1351 re Debian chroot prompt

faho:
>#1351 re Debian chroot prompt

So the argumentation there was that we don't want to make the default more complicated, so it should be in a separate prompt.

I disagree pretty strongly with that, because, like I've said numerous times: This is most useful when it's the default. When you need to specifically configure it you most likely won't, and then you'll never see it.

It's also a super cheesy solution because it is specific to debian. Any good solution here should probably also work for other kinds of virtualization.

I would argue for:

- Make a "machine id" function
- Add it to the default prompt

I've argued for this in the past with #6398. Note also that Debian apparently defaults to showing chroots for bash.

I don't think a separate prompt is of much use.

faho:
Alright. The perfect being the enemy of the good, I have now merged this.

It doesn't solve *all* issues with the sample prompts, but it is an improvement.

Issue #7105, opened 2020-06-10
Default stderr-nocaret and qmark-noglob to ON


faho:
When we introduced feature flags in #4940, the understanding was that these aren't configuration knobs, but a deprecation mechanism - we don't support setting these online, only via universal variables or a commandline argument.

With that in mind, I think it's time to continue the deprecation process by defaulting some of these to ON instead of OFF, and allowing people to turn them off.

I propose enabling the following:

- stderr-nocaret - the removal of the `^` redirection
- qmark-noglob - the removal of the `?` single-character glob

for the next release, and removing the option after a year (the flag will still be available to be *queried*, but you won't be able to set it).

Ideally we'd also have `fish -n` warn about any uses of these so it's easier to find.

zanchey:
This is going to break heaps of stuff, even if it's easy-ish to fix quickly by turning the flags back on. I think we need to telegraph loudly at least one release beforehand, so I think this would be best noted in the release notes for 3.2 as "this is the last release that they will be left at defaults".

stderr-nocaret is easy, qmark-noglob is harder to work around (I actually use it now and then though I recognised it's a real footgun).

faho:
>qmark-noglob is harder to work around (I actually use it now and then though I recognised it's a real footgun).

What issues have you had?

AFAIR it's basically quoting all `?`. The only other problem is with `case`s including a literal question mark, which now *needs* to be tested. I don't think that happens too often?

zanchey:
That is, I use `?` as a glob, and it's moderately tricky to replace (`*` is not as safe). My history says I use it much more as an escaped question mark than a glob though!

faho:
Okay, how about enabling nocaret by default and trying to figure something out for globs before we remove `?`?

zanchey:
See also #4394

faho:
Alright, we have enabled stderr-nocaret by default, let's close this issue for that and revisit it for other flags.

robey:
If you found this bug because fish started throwing errors in some of your scripts after the last update, the solution is to replace every `^` with `2>`.

That is, `^/dev/null` is now spelled `2>/dev/null`

(It would have been nice to deprecate with a warning before removing, but I guess it's too late now.)

PR #7907, opened 2021-04-07
Enable stderr-nocaret by default

zanchey:
* Teaches the feature flags about a default value - I'm not sure this is the best approach and would appreciate feedback.
* Enable stderr-nocaret by default.

Work on #7105.

## TODOs:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,7 @@ Syntax changes and new commands
 Deprecations and removed features
 ---------------------------------
 - The ``fish_history`` value "default" is no longer an alias for "fish" (:issue:`7650`).
+- Redirection to standard error with the ``^`` character has been disabled by default. It can be turned back on using the ``stderr-nocaret`` feature flag, but will eventually be disabled completely (:issue:`7105`).
 
 Scripting improvements
 ----------------------
--- a/src/future_feature_flags.cpp
+++ b/src/future_feature_flags.cpp
@@ -16,7 +16,7 @@ features_t::features_t() {
 features_t features_t::global_features;
 
 const features_t::metadata_t features_t::metadata[features_t::flag_count] = {
-    {stderr_nocaret, L"stderr-nocaret", L"3.0", L"^ no longer redirects stderr", false},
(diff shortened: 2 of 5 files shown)


faho:
Looks alright to me.

Maybe add a test that checks it can be disabled again?

zanchey:
Looks like Past Fabian was way ahead of us - 9d5620c170

CHANGELOG.rst
```diff
@@ -10,6 +10,7 @@ Syntax changes and new commands
 Deprecations and removed features
 ---------------------------------
 - The ``fish_history`` value "default" is no longer an alias for "fish" (:issue:`7650`).
+- Redirection to standard error with the ``^`` character has been disabled by default. It can be turned back on using the ``stderr-nocaret`` feature flag for interactive sessions, but will eventually be disabled completely (:issue:`7105`).
```
krobelus, line 13: > for interactive sessions

Is it really only for interactive sessions? If not we could drop this bit.

src/future_feature_flags.cpp
```diff
@@ -6,15 +6,20 @@
 
 #include "wcstringutil.h"
 
-features_t::features_t() = default;
+features_t::features_t() {
+    for (const metadata_t &md : metadata) {
```
krobelus, line 10: This seems reasonable. I guess we can also make the constructor private, to make clearer that it's never constructed beyond the `global_features` singleton.

```diff
diff --git a/src/future_feature_flags.h b/src/future_feature_flags.h
index e31b306e7..727cded47 100644
--- a/src/future_feature_flags.h
+++ b/src/future_feature_flags.h
@@ -70,4 +70,2 @@ class features_t {
 
-    features_t();
-
     features_t(const features_t &rhs) { *this = rhs; }
@@ -82,2 +80,4 @@ class features_t {
    private:
+    features_t();
+
     // Values for the flags.
```

krobelus: nevermind sorry, that won't work because we construct it in tests

CHANGELOG.rst
zanchey: It is not. Thanks.

PR #7908, opened 2021-04-07
Update fish_vi_key_bindings.fish

xiruizhao:
## Description

add bindings for undo and redo

## TODOs:

### DIFF
--- a/share/functions/fish_vi_key_bindings.fish
+++ b/share/functions/fish_vi_key_bindings.fish
@@ -206,6 +206,9 @@ function fish_vi_key_bindings --description 'vi-like key bindings for fish'
     bind -s --preset J end-of-line delete-char
     bind -s --preset K 'man (commandline -t) 2>/dev/null; or echo -n \a'
 
+    bind -s --preset u undo
+    bind -s --preset \cr redo
+
     bind -s --preset yy kill-whole-line yank
     bind -s --preset Y kill-whole-line yank
     bind -s --preset y\$ kill-line yank


faho:
These are already bound to history-search-backward/forward: https://github.com/fish-shell/fish-shell/blob/783736c77f904e5177ec98be5678b7785b7b687d/share/functions/fish_vi_key_bindings.fish#L104-L105

Given that undo is also not bound in readline's vi-mode I prefer not to change the default here. Sorry!

xiruizhao:
In [readline 8.1](https://man7.org/linux/man-pages/man3/readline.3.html), `u` is bound to `vi-undo`, although `C-R` is bound to `reverse-search-history`. Conforming to vim bindings is a better idea.

krobelus:
> In readline 8.1, u is bound to vi-undo

I think it makes sense to do the same; the history search one is not a big loss given that `k` already does up-or-search.

It doesn't feel nice: after typing `echo `, pressing `u` once doesn't move the cursor.

I also tried binding undo to `\c_` in insert mode (like readline) but hit #7910.

faho:
In that case, we should however remove the previous bindings.

xiruizhao:
What about binding previous ones to `/` and `?` ? Removing them is also fine.

faho:
A vi-bindings user would have to decide that. I don't know what makes sense here.

(Also note that my bash with readline 8.1 does *not* have "u" bound to vi-undo - `INPUTRC=/dev/null bash -c 'bind -Pm vi'` says nothing about vi-undo or "u")

xiruizhao:
I just tried, `u` actually binds to `vi-undo` (this is also reflected in source code https://github.com/bminor/bash/blob/f3a35a2d601a55f337f8ca02a541f8c033682247/lib/readline/vi_keymap.c#L165), probably a bug of `bind`?

faho:
Doesn't happen for me. `u` doesn't appear to be doing anything. Probably some compile-time settings again (I've checked my local customization, I have very little of it and nothing that should apply to vi-mode let alone `u`).

faho:
Ah, no `undo` is just weirder than I thought - it doesn't undo entering text, but if I `dw` then it'll undo something.

xiruizhao:
The `bind` bug is probably due to https://github.com/bminor/bash/blob/f3a35a2d601a55f337f8ca02a541f8c033682247/lib/readline/funmap.c#L59 not having `vi-undo`. I will notify readline's maintainer.

xiruizhao:
Since `j` and `k` include forward/backward history search functionality. I deleted previous bindings. You can reopen this PR and review it. @faho

krobelus:
@xiruizhao we can't reopen after a force push, but I'll cherry-pick your commit and add a changelog entry.

xiruizhao:
> Ah, no `undo` is just weirder than I thought - it doesn't undo entering text, but if I `dw` then it'll undo something. Still not displayed in `bind -P`.

It would be actually nice to undo insertions/paste.

zanchey:
> @xiruizhao GitHub doesn't allow to reopen after a force push, but I'll cherry-pick your commit and add a changelog entry.

That's f66f9993d765660dff5f0fee702177b4ed842052 from https://github.com/xiruizhao/fish-shell/tree/patch-1 I take it?

krobelus:
Sorry for the delay, cherry-picked as 8bbb06bf5c9efb6d3e27c35af8d88e5f83987260

Issue #7445, opened 2020-11-02
Add option to list kill ring entries


mamiu:
I couldn't find any command to obtain (or variable that holds) the entries of the [killring](https://fishshell.com/docs/current/#killring) feature (except the two bind input functions `yank` and `yank-pop` which don't list all entries).

Or is there another option to get all entries from the kill ring?

faho:
There currently is not.

See also #4028, which could use it.

The obvious question is what this should look like:

- A command that prints the entries?
- A variable that contains them, like $history?

septatrix:
There seems to be `$fish_killring`

faho:
Yup, forgot to close this one (never really sure when exactly Github does that for you).

septatrix:
While at it: Is it somehow possible to remove an entry from the killring? `yank-pop` does not seem to do anything for me (also no rotating which is mentioned in the docs) and `set --erase` does not work for that read-only variable

krobelus:
> While at it: Is it somehow possible to remove an entry from the killring?  `set --erase` does not work for that read-only variable

not yet, probably we should support `set --erase`

> `yank-pop` does not seem to do anything for me (also no rotating which is mentioned in the docs)

it only rotates what is inserted to the commandline after C-y (the GNU Emacs tutorial explains how this kind of killring works)

septatrix:
> not yet, probably we should support `set --erase`

I would also be fine with an extra function. My use case is a `push-line` like functionality with a stack (such that multiple commands can be stored and popped one after another) for which I wanted to use the built in killring. For now I will try to implement it using a normal list.

> > `yank-pop` does not seem to do anything for me (also no rotating which is mentioned in the docs)
> 
> it only rotates what is inserted to the commandline after C-y (the GNU Emacs tutorial explains how this kind of killring works)

Oh that makes a lot more sense now. I did not knew the concept was based on Emacs - that also explains the terminology which was rather confusing to me as I normally associate yanking with copy/cut.

septatrix:
Should I create an issue for that (i.e. some way to remove entries from the killring)?

ridiculousfish:
The killring should probably just be a fish variable.

PR #7916, opened 2021-04-09
List killring entries

Caroleq:
## Description
This is proposal for a new variable containing entries from killring.   

I added a test for that, but tbh I'm not sure if this is needed - there is little that may get broken.  

Fixes issue #7445

## TODOs:


krobelus:
Seems reasonable. You should rebase on latest master, then the merge commit will go away.

Caroleq:
> Seems reasonable. You should rebase on latest master, then the merge commit will go away.

Done.  


Now I think we may want to add this variable to the documentation. I'll update PR today or tomorrow.

zanchey:
I suggest adding this to the list in "Special variables" in `doc_src/interactive.rst` too.

krobelus:
One interesting question is if/how the killring can be changed by the user. I think currently none of the electric variables are modifiable. Either way, I'm sure that's possible to add later.

Caroleq:
> I suggest adding this to the list in "Special variables" in `doc_src/interactive.rst` too.

This section was in language.rst - I added it there.  



> One interesting question is if/how the killring can be changed by the user. I think currently none of the electric variables are modifiable. Either way, I'm sure that's possible to add later.

Hmm, my understanding of kill ring is that it can be modified only by adding new entries.

zanchey:
I apologise for the bikeshedding, but I'm just wondering if it would be better to call this `fish_killring`. Electric variables become reserved variable names, and have the potential to break existing uses (though I can't see any in https://github.com/search?l=fish&p=1&q=killring&type=Code).

krobelus:
> Hmm, my understanding of kill ring is that it can be modified only by adding new entries.

Yeah but maybe there's a good reason to modifying it on-the-fly, even if we can't imagine one yet.

(New subcommands to the builtin `status` are often a viable alternative to new special variables, but a variable seems OK here)

ridiculousfish:
Looks good, merged as 9d66ddc8408b98362d5377cc498af02ff60eb991. Thank you!

PR #5417, opened 2018-12-15
Add fish startup flag `--no-config`, take 2

mqudsi:
This prevents fish from loading the user configuration file, both at
startup and at any point thereafter, in both interactive and
non-interactive modes.

As fish relies extensively on `$__fish_config_dir` for core
functionality (universal variables), this is achieved by creating a new
runtime path (using the existing code to do so) to serve as the user
config dir.

--- 

Either this or #5416 should be merged, depending on how far we want to take this.

PR #5416, opened 2018-12-15
Add fish startup flag `--no-config`, take 1

mqudsi:
This prevents fish from loading the user configuration file at startup.
It is not intended to affect interactive uses of fish, which will end up
sourcing the file automatically, but it does prevent sourcing the user
`config.fish` file when fish is called non-interactively.

i.e. users with bad `config.fish` entries that should not be executed in non-interactive mode (but aren't guarded with `if not status --interactive`) will not be run at startup. But running fish interactively with `--no-config` will end up loading (or creating) the needed files to automatically source functions, etc.

--- 

Either this or #5417 should be merged, depending on how far we want to take this.

Issue #5394, opened 2018-12-07
Don't read configuration if not interactive


faho:
One of the nice things of fish is that we don't have anything as complicated as the bash man page's "INVOCATION" section in our documentation. 

"If it's a non-interactive shell, it does not read bashrc, except when it has been invoked by ssh or rshd, unless it is invoked as sh".

We always read config.fish, and that's nice.

However, as we see in e.g. #5246 and #4432, doing that even when non-interactive can cause some issues. Issues that we currently expect the user to expect, and guard against with `if status is-interactive`.

And I'm not really sure if configuration of non-interactive shells is ever really wanted? If you execute a script via `fish /path/to/script` (or directly via the shebang), you expect a consistent environment.

So, I would argue that we should possibly change this:

- Don't source /etc/fish/config.fish and ~/.config/fish/config.fish (the one in /usr can stay, but that's not really "config")

- Don't read snippets (which are technically in /usr/share/fish/config.fish)

- Don't add the user's functions (in $HOME and /etc) to $fish_function_path

Now, obviously this complicates the logic a bit. It's "if fish is interactive, read config.fish, if not, don't" instead of just "read config.fish". I'd argue that it's still quite simple, and helps unbreak things that are currently quite easy to break - understanding that `stty` touches the terminal is advanced stuff.

Additonally, if configuration of non-interactive shells is wanted (and I don't see how, given that bash doesn't have it), we could add a "config-noninteractive.fish", or simply expect users to run `fish -C 'source ~/.config/fish/config.fish`.

zanchey:
I'm against this. I do have stuff in my non-interactive startup config (path editing dependent on machine, for example), and I use my custom functions from non-interactive mode all the time.

faho:
>I do have stuff in my non-interactive startup config (path editing dependent on machine, for example)

Could you provide some more detail?

Like I said, it's _possible_ that non-interactive configuration is needed, I just believe that the default is backwards. Most configuration should not apply to non-interactive shells, and the current solution is a footgun.

>path editing dependent on machine

You mean $PATH? Shouldn't that be done once in the login shell and then passed down? Or in e.g. pam_env? (There's a bit of an annoyance here in that much of linux expects a shell to run in the startup path, and specifically one that can read /etc/profile* - but fish can't do that to begin with) 

>I use my custom functions from non-interactive mode all the time.

[...]

PR #7921, opened 2021-04-11
Add `--no-config` option

### LINKED ISSUE #1256: Add support for disabling configuration
For testing purposes it would be handy to disable loading of some configuration files. I would suggest adding `-R`/`--no-rc` switch for disabling loading files from `/etc` and `$XDG_CONFIG_HOME` and `-D`/`--no-defaults` for disabling loading from `/usr/share/fish`.

faho:
## Description

This adds a `--no-config` option to fish, to let it start up without reading the config.

In particular it will:

- Keep $fish_function_path set to *just* the datadir (so /usr/share/fish/functions - just the stuff we ship). No sysconfdir, no vendor directory, nothing else. ($fish_complete_path isn't set at all at the moment)
- Disable universal variables. If you try to *set* a universal variable, it will be silently set as a global instead (this is a general change - if fish doesn't manage to set up universal variables, `set -U var val` is "redirected" to global). They aren't read at all.
- Disables history by enabling private mode
- Runs *no* config.fish or conf.d snippets

My hope is that this is basically "independent" of the system entirely. I'm not sure I caught all the places we write something or read something, so anyone else checking is appreciated.

One big issue with this is that it's very uncomfortable interactively because we don't set the colors or set up the handler for $fish_key_bindings. Both of those things happen in `__fish_config_interactive`, which isn't started.

The bindings can be improved later -  {up,down}-or-search shouldn't be functions and then we can just add them to the hardcoded defaults. The colors are tricky because they're usually defined as uvars and we want to allow unsetting a color.

There are a few things we could be doing differently - for instance it would also be possible to do without $fish_function_path entirely, by leaving it unset. This is awkward because fish isn't fully operational without the functions. It's gotten *better* since e.g. `type` and `eval` are no longer functions, but `psub` still is.

Fixes #1256 
Rerun of #5417, #5416
Helps with #5394.

## TODOs:

### DIFF
--- a/share/functions/fish_command_not_found.fish
+++ b/share/functions/fish_command_not_found.fish
@@ -11,6 +11,10 @@ if test -r /etc/os-release
     string replace -r '^ID(?:_LIKE)?\s*=(.*)' '$1' | string trim -c '\'"' | string split " ")
 end
 
+function __fish_default_command_not_found_handler
+    printf "fish: Unknown command: %s\n" (string escape -- $argv[1]) >&2
+end
+
 # If an old handler already exists, defer to that.
 if functions -q __fish_command_not_found_handler
     function fish_command_not_found
--- a/share/functions/fish_prompt.fish
+++ b/share/functions/fish_prompt.fish
@@ -5,6 +5,8 @@ function fish_prompt --description 'Write out the prompt'
     set -l last_pipestatus $pipestatus
     set -lx __fish_last_status $status # Export for __fish_print_pipestatus.
     set -l normal (set_color normal)
+    set -q fish_color_status
+    or set -g fish_color_status --background=red white
 
     # Color the prompt differently when we're root
     set -l color_cwd $fish_color_cwd
--- a/share/tools/web_config/sample_prompts/default.fish
+++ b/share/tools/web_config/sample_prompts/default.fish
@@ -5,6 +5,8 @@ function fish_prompt --description 'Write out the prompt'
     set -l last_pipestatus $pipestatus
     set -lx __fish_last_status $status # Export for __fish_print_pipestatus.
     set -l normal (set_color normal)
+    set -q fish_color_status
+    or set -g fish_color_status --background=red white
 
     # Color the prompt differently when we're root
     set -l color_cwd $fish_color_cwd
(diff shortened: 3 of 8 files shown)


faho:
Alright, I'm merging this now to get more testing, and because the biggest thing (key bindings aren't set) is probably best solved by adding the $fish_key_bindings variable handling to C++, and that's a good PR on its own.

Imagine this as a preview - I think we're on the right track, but the specifics may need hashing out.

robfordww:
will creating scripts by "#!/bin/fish --no-config", be the preferred way now if you want to create predictable scripts for say setup/install or similar.

faho:
@robfordww Basically not. The issue is that you can't guarantee fish's path (and it most likely will not be /bin/fish, but either /usr/bin/fish or /usr/local/bin/fish), so you would have to use

```
#!/usr/bin/env fish
```

and on linux at least the shebang can only have one argument - `#!/usr/bin/env fish --no-config` would pass `fish --no-config` to `env` as one argument, so it would then try to execute a file called `fish --no-config`. Which probably doesn't exist. (technically not even /usr/bin/env has to exist but I have zero sympathy for platforms that don't have it, and they're not ones you're likely to use)

If you *can* guarantee fish's path, then sure, go ahead. If not you'd either need a pre-processing step or a wrapper sh-script to execute `fish --no-config`. But really we need #5394.

robfordww:
@faho Thanks.  I was not aware of the "one-argument" thing in #!
    `#!/bin/env` -S fish --no-config  
has some issues then I am not aware of?

faho:
`env -S` has two issues:

1. You need to know about it and remember to use it
2. It's a GNU extension

So if you can guarantee your script will only ever be run on GNU systems, have at it. It's just not a solution for us and needlessly narrows the platforms your script can run on. Also it's an ugly hack.

kopischke:
Just for the record, FreeBSD added the `-S` flag to `env` in version 6.0 (released in 2007), and macOS inherited that. That still leaves a number of platforms in the lurch, of course, starting with OpenBSD and Solaris …

robfordww:
While waiting for a prettier recipe, the `env -S` option seems like a reasonably portable approach. Linux and Mac OS I would guess is 95% of fish's relevant market.

PR #7925, opened 2021-04-13
alias: print error msgs to stderr

ammgws:
As per title.


faho:
Yup, seems good. Merged!

Issue #736, opened 2013-05-12
fish_config only copies fish_prompt function


aeosynth:
osx 10.7.5, fish HEAD

after selecting the 'Robbyrussell' prompt from the `fish_config` web interface, the shell displays errors:

```
fish: Unknown command '_git_branch_name'
/Users/james/.config/fish/functions/fish_prompt.fish (line 2): _git_branch_name
                                                               ^
in command substitution
    called on line 11 of file '/Users/james/.config/fish/functions/fish_prompt.fish',

in function 'fish_prompt',
    called on standard input,

in command substitution
    called on standard input,
```

manually copy / pasting the displayed code into `fish_prompt.fish` works fine, as does using the 'Classic + Git' prompt

jhdxr:
8 years passed, I'm wondering if this issue get addressed? Sorin uses ` fish_right_prompt` and it seems still not working

ridiculousfish:
This is not fixed properly; however as of b722e9ae329cf407f5a7a74d062bc1a2f677fecf none of the sample prompts define function outside of `fish_prompt`, so it's worked around. That commit will be in 3.2.

jhdxr:
@ridiculousfish Thanks. Just curious, is there any special reason that you always prefer these workarounds rather than modifying webconfig.py?

ridiculousfish:
No, it's just easier. Modifying webconfig.py is the right approach but would have to duplicate some work done by `funcsave`, like fetching `$__fish_config_dir` and creating the functions directory. If you or anyone has the appetite for it then we'd love to merge a PR.

PR #7930, opened 2021-04-14
Handle right prompts in webconfig

faho:
## Description

This allows webconfig to show and save fish_right_prompt if it is defined in a sample (or the current prompt).

It will *not* remove an existing fish_right_prompt if there is one.

It looks like this:

Screenshot_20210414_205112

Fixes issue #736 (at least for fish_right_prompt, I don't think we need more).

In addition (and this is optional, it was just nice to get a better screenshot), this adds a slightly simplified version of my own prompt as the "Disco" sample. Here's a preview:

Screenshot_20210414_202949

(the $PWD color depends on the sha of $PWD - this requires sha256sum to be available. if not it'll gracefully fallback to $fish_color_cwd. It will also only show user@host if necessary)

## TODOs:

### DIFF
--- a/share/tools/web_config/js/controllers.js
+++ b/share/tools/web_config/js/controllers.js
@@ -198,6 +198,7 @@ controllers.controller("promptController", function($scope, $http) {
 
             // Update attributes of current prompt and select it
             $scope.samplePrompts[0].demo = selectedPrompt.demo;
+            $scope.samplePrompts[0].right = selectedPrompt.right;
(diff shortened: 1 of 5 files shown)


faho:
Changed the layout by removing the spacer div, instead letting the left prompt grow. It now looks like

Screenshot_20210417_173256

ridiculousfish:
This looks great to me! Awesome!

PR #7931, opened 2021-04-15
Remove completion files for `ls` aliases; just use `function --wraps`

cstyles:
## Description

This PR fixes a minor bug I encountered. If a user adds their own alias for `ll` (e.g., with `alias ll 'exa -l'`) and then persists the alias with `funcsave ll`, the alias/function will wrap both `exa` and `ls` and both sets of completions will be suggested. I don't encounter this bug when the completion files are absent.

### DIFF
--- a/share/completions/la.fish
+++ b/share/completions/la.fish
@@ -1,1 +1,0 @@
-complete -c la -w ls
--- a/share/completions/ll.fish
+++ b/share/completions/ll.fish
@@ -1,1 +1,0 @@
-complete -c ll -w ls
--- a/share/functions/la.fish
+++ b/share/functions/la.fish
@@ -1,6 +1,6 @@
 #
 # These are very common and useful
 #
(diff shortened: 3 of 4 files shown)


faho:
This means that the completions now only work after you've loaded the function.

Tbh we should probably just remove these.

krobelus:
> This means that the completions now only work after you've loaded the function.

But fish loads the function as soon as the user types `ll`, right? So it's loaded whenever needed with this change.

faho:
>But fish loads the function as soon as the user types ll, right?

No, it loads the *completion* as soon as the user types it, because that's where `--no-files` would be.

krobelus:
I like having `ll` without setup

faho:
Merged. This is correct by itself (as we load the function when required), and removing these functions is a different discussion.

Thanks!

PR #7932, opened 2021-04-15
Add a "prompt_login" helper function

faho:
## Description

This prints a description of the "host". Currently that's

`(chroot:debianchroot) $USER@$hostname`

with the chroot part when needed. That means instead of having one "debian chroot" prompt, we support that out of the box.

This also switches the default and terlar prompts to use it, the other
prompts have slightly different coloring or logic here. It would be possible to allow that by adding options to prompt_host.

Part of #7884.

## TODOs:

## Questions:

- Is this wanted at all? Like I said I think having debian chroots off in a separate prompt just results in nobody using it because they would have to change the prompt *in the chroot* (which means allowing for `fish_config` to work there!), so the alternative seems to be not supporting that at all
- What about the name? `prompt_host` sounds like it might be `prompt_hostname`, but `prompt_chroot_user_at_host` is a bit wordy. `prompt_machine` perhaps?
- Support for things other than the debian chroot thing - do other systems have something similar? How about virtual machines? We already color the host differently if it's ssh.

### DIFF
--- a/doc_src/cmds/prompt_login.rst
+++ b/doc_src/cmds/prompt_login.rst
@@ -1,0 +1,26 @@
+.. _cmd-prompt_login:
+
+prompt_login - describe the login suitable for prompt
+=====================================================
+
+Synopsis
+--------
+
+::
+
+    function fish_prompt
+        echo -n (prompt_login) (prompt_pwd) '$ '
+    end
+
+Description
+-----------
+
+``prompt_login`` is a function to describe the current login. It will show the user, the host and also whether the shell is running in a chroot (currently debian's debian_chroot is supported).
+
+Examples
+--------
+
+::
+
+    >_ prompt_login
+    root@bananablaster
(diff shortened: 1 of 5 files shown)


krobelus:
(#6398 had a more comprehensive solution, using `systemd-detect-virt` and Debian's `ischroot`, at the cost of adding forks at startup)

Not a strong opinion but I'd be fine with dropping the debian chroot prompt. Or replace it with the generic `systemd-detect-virt --chroot`. (We don't care too much about containers/virtualization, because they hopefully have a different hostname from the host, but the chroot one can be nice.) Though I don't know how often this is useful in practise, is chroot still commonly used (over containers)?.

Distribution integration is a bit lacking in general; I wouldn't be surprised if $PATH was not set up correctly when using fish as login shell on a Debian (chroot) system. That would be nice to have fixed but it's not really within our scope.

`prompt_login` would be an alternative name, if we need it.

zanchey (approved):

faho:
Alright, renamed to `prompt_login`. It's not perfect, but at least it doesn't have the host/hostname confusion.

PR #7935, opened 2021-04-16
Fix skipping locale tests on Github Actions

faho:


faho:
Okay, this is now running all the actions *twice* - once for the push, once for the pr. That seems wasteful. Is there a way to change it?

PR #7945, opened 2021-04-21
Disable single-line if statements in clang-format

ridiculousfish:
This switches our clang-format to disallow single-line if statements.

Instead of

     if (retval != STATUS_CMD_OK) return retval;

we now have:

     if (retval != STATUS_CMD_OK)
          return retval;

I think @faho made a comment that he disliked single-line if statements and I agree. Does anyone want to keep them?

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -6,6 +6,7 @@
 BasedOnStyle: Google
 ColumnLimit: 100
 IndentWidth: 4
+AllowShortIfStatementsOnASingleLine: false
 
 # Place config.h first always.
 IncludeCategories:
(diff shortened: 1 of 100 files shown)


faho:
I want it to be *exactly the other way around*. I do not want *unbraced multiline ifs*

I want this:

```c++
 if (retval != STATUS_CMD_OK) return retval;
```

because here there is no misleading indentation. It's clear what it does and short.

NEVER this:

```c++
 if (retval != STATUS_CMD_OK)
      return retval;
```

If you must, use this:

```c++
 if (retval != STATUS_CMD_OK) {
      return retval;
}

So that's a -1 from me.

krobelus:
When writing, I like the minimalism of single-line-if, but that's the wrong place to optimize.
For readers, the Pythonic way is probably better.

The misleading indentation example is valid, but there is a compiler warning (maybe turn that into an error if we'd overlook it?).

ridiculousfish:
Gotcha, looks like there's not much appetite for this so I will close.

PR #7952, opened 2021-04-25
Avoid excessive polling of universal variable file

krobelus:
Reproducible at least on Linux, where the "named pipe" universal
variable notifier is used:

	rm -rf build/test/xdg_config
	XDG_CONFIG_HOME=build/test/xdg_config ./build/fish -c "xterm -e ./build/fish"

The child fish reacts to keyboard input with a noticeable initial
delay.  This is because the universal variable file is polled over
a million times, even when I immediately press Control-D. This polling
prevents readb() from handling keyboard input.
Before commit 939aba02d ("Refactor input_common.cpp:readb"), readb()
reacted to keyboard input even when there were universal variable
notifications.  Restore this behavior, but make sure to call the
universal variable notifier after the new "prepare_to_select" logic.
Maybe the problem is in the notifier but the old behavior was sane.

Fixes the problems described in
https://github.com/fish-shell/fish-shell/commit/7a556ec6f21f16b528be3fb94f5160e2209c3cbf#commitcomment-49773677

Adding `-d uvars-file` shows that we are checking the uvar file repeatedly:

```
uvar-file: universal log sync
uvar-file: universal log sync elided based on fast stat()
uvar-file: universal log no modifications
```

## TODOs:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,7 +11,7 @@ jobs:
     - uses: actions/checkout@v2
     - name: Install deps
       run: |
-        sudo apt install gettext libncurses5-dev libpcre2-dev python3-pip
+        sudo apt install gettext libncurses5-dev libpcre2-dev python3-pip tmux
         sudo pip3 install pexpect
     - name: cmake
       env:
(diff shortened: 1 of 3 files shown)


faho:
Alright, this makes sense to me and seems to fix the problem.

LGTM, merged.

ridiculousfish:
Sorry I didn't get to this; this looked fine to me too.

Issue #7784, opened 2021-03-06
string pad doesn't work with escapes


IlanCosman:
```fish
Welcome to fish, the friendly interactive shell
Type help for instructions on how to use fish
ilan@arch ~> string pad --width 20 hello
               hello
ilan@arch ~> string pad --width 20 \ehello
ello
ilan@arch ~> string pad --width 20 (set_color red)hello
hello
ilan@arch ~> 
```

It would be really nice if `set_color` and `string pad` would work together.

faho:
In general, this is to be expected, interpreting arbitrary escape sequences here would require us to become a terminal.

We could probably handle colors and such.

IlanCosman:
Honestly, now that I think about it, it wouldn't be so insane to design a shell and a terminal together. Yah people couldn't use whatever their favorite terminal was or whatever, but you could probably do some pretty great stuff with that integration. Off topic though 😂

IlanCosman:
Huh, this seems to work now with Fish 3.2.1 on MacOS and Fish master on Arch. I guess something in 3.2.1 fixed it. 👍

krobelus:
I don't think it's fixed, see this:

```
$ string pad --width 20 (set_color red)hello hello
                     hello
               hello
```

The escape sequences take 6 characters.
We'll have to think about whether it makes sense to support colors here. Maybe there should be a different command to remove escapes?

IlanCosman:
Good point, it's not fixed.

> Maybe there should be a different command to remove escapes?

👀 yes please! I've been rolling [my own](https://github.com/IlanCosman/tide/blob/main/functions/_tide_decolor.fish).

faho:
Fun fact: This needs to handle multline strings, and there are some ambiguities.

If you have a string like

```fish
echo aaa\nbb
```

The width of that is 3 and then 2. Calling it "3" (the maximum) doesn't make sense, calling it 6 (the "length) makes even less. Really 3, 2 is the correct answer.

But if you have one like

```fish
echo aaa\rbb
```

The width of that is? "3" sorta makes sense because that is how wide it would show up in isolation (it would look like "bba"), but "2" would also make sense because, if you add anything *to* it it works like it's 2 cells wide (printing a "1" would make it look like "bb1", not "bba1").

And if the "aaa" wraps into the next line? All bets are off. We can't detect it, we can't express it.

[...]

Issue #3625, opened 2016-12-06
fish_config should have a text UI


vbextreme:
fish implements fish_config command that lets you view and select, but not editing, configurations.

It looks very cute, but to create a web server and then open a web browser to display the configurations seem a choice a little absurd.
I was expecting something more consistent as a classic ncurses menus, this would also permit to change configurations via tty, which is now unusable.

krader1961:
Patches are welcomed. It's a small project with just a handful of regular committers. So if someone like yourself, @vbextreme, are passionate about improving fish in some way we'd love to work with you to make that happen.

Since we already have a dependency on python, and have no plans to remove that dependency, the natural place to implement this capability is in the *share/tools/web_config/webconfig.py* program (or a fork of it). Then modify the `fish_config` function to include a flag to select text versus web UIs.

vbextreme:
thanks for the information.
I hope to recover a bit 'of time to add the new functionality.

silverhook:
I have the same issue as @vbextreme – running Fish on a server.

What I think would be already useful enough is if `fish_config` would produce an HTML page that can actually be used in a CLI web browser – whether that be lynx, elinks or w3m.

silverhook:
A different solution to the same problem is also being discussed in bug #1916.

krader1961:
I predict issue #1916 won't be implemented anytime soon. Security is hard. Also, having a curses based solution has the advantage that it's running inside your terminal so accurately displaying colors, especially how they look given the terminals background color, becomes trivial. Personally I wish the person who implemented the web UI had instead spent the time creating a curses based UI.

silverhook:
> Personally I wish the person who implemented the web UI had instead spent the time creating a curses based UI.

Personally, I agree.

But I see how having docs and settings in browser does make it more appealing to the modern-day crowd, so I don’t see it going away any time soon.

Having a good ncurses option for `fish_config` does seem like a sterling idea though.

A(n interim) lighter idea still might be to just have a CLI command to select (maybe even preview) prompts.

faho:
>A(n interim) lighter idea still might be to just have a CLI command to select (maybe even preview) prompts.

I've actually tried to do something like that. It'd be used like `prompt view sorin`. I kinda feel like this CLI feels a bit better than some `dialog`-like ncurses thing would.

[...]

PR #7958, opened 2021-04-27
A cli prompt selector

faho:
It's a bit weird to *have* to fire up a browser to get fish_config to
choose a prompt.

So this adds `prompt`:

- `prompt list` shows all the available prompt names
- `prompt show` demos the available sample prompts
- `prompt choose` sources a prompt
- `prompt save` makes the choice permanent

Part of #3625.

This means that you can e.g. pick the "informative" prompt by adding `prompt choose informative` to config.fish, and it will just always `source` the one from the currently installed fish (so it will update when we update the sample). This is on purpose and has basically no overhead, so it's entirely acceptable - we might even want to recommend it in the docs.

## TODOs:

- [x] This does not currently correctly show the right prompt, which is awkward
to do because we'd have to move it to the right (and this involves width-without-color, see #7784).
- [x] How about the name? `prompt`, or maybe a more general "config tool"? A special mode for `fish_config`, so if you do `fish_config prompt` it'll run in the terminal, `fish_config` will keep opening the browser?

### DIFF
--- a/doc_src/cmds/fish_config.rst
+++ b/doc_src/cmds/fish_config.rst
@@ -8,23 +8,37 @@ Synopsis
 
 ::
 
-    fish_config [TAB]
+    fish_config
+    fish_config browse
+    fish_config prompt (choose | list | save | show)
 
 Description
 -----------
 
-``fish_config`` starts the web-based configuration interface.
+``fish_config`` is used to configure fish.
 
-The web interface allows you to view your functions, variables and history, and to make changes to your prompt and color configuration.
-
-``fish_config`` starts a local web server and then opens a web browser window; when you have finished, close the browser window and then press the Enter key to terminate the configuration session.
-
-``fish_config`` optionally accepts name of the initial configuration tab. For e.g. ``fish_config history`` will start configuration interface with history tab.
+Without arguments or with the ``browse`` command it starts the web-based configuration interface. The web interface allows you to view your functions, variables and history, and to make changes to your prompt and color configuration. It starts a local web server and opens a browser window. When you are finished, close the browser window and press the Enter key to terminate the configuration session.
 
 If the ``BROWSER`` environment variable is set, it will be used as the name of the web browser to open instead of the system default.
 
+With the ``prompt`` command ``fish_config`` can be used to view and choose a prompt from fish's sample prompts inside the terminal directly.
+
+Available subcommands for the ``prompt`` command:
+
+- ``choose`` loads a sample prompt in the current session.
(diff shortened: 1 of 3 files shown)


zanchey:
This is rad.

image

Name: There's only one package in Debian that has a `/usr/bin/prompt` command, and it's `libmodglue1` which is installed on a [tiny fraction of reporting Debian installs](https://qa.debian.org/popcon.php?package=modglue), so I don't think `prompt` is *bad* as such... but I wonder if `fish_choose_prompt` or something would be better. Alternatively, I do like the idea of `fish_config` being able to do both terminal and Web configuration, with autodetection (eg for SSH) and an appropriate override flag.

UI: I think a bare `prompt` should be the same as `prompt list`. I think `prompt choose` should probably imply `prompt save`.

faho:
>I think prompt choose should probably imply prompt save.

The idea here is that you can use `prompt choose` to test-drive a prompt - you can do `prompt choose scales`, then `cd` around a bit, run a few commands, see what it does.

Or, like I added in above, you can just add `prompt choose scales` in config.fish and it will *track the updates* to the sample (this would make #7918 nicer because it would allow us to remove the now-useless function!).

We could *add* a `prompt choose --save` shorthand (like `funced --save`), but I do believe not saving by default is correct, because test-driving is a very normal thing to do.

share/functions/prompt.fish
```diff
@@ -0,0 +1,81 @@
...
+        __fish_print_help prompt
+        return 0
+    end
+
+    if contains -- $cmd list; and set -q argv[1]
+        echo "Too many arguments" >&2
+        return 1
+    end
+
+    set -l prompt_dir $__fish_data_dir/sample_prompts $__fish_data_dir/tools/web_config/sample_prompts
```
faho, line 18: (note: This was added in preparation of moving the sample_prompts directory out of web_config - because it no longer *is* web-exclusive)

faho:
>I think a bare prompt should be the same as prompt list.

Done

faho:
Current output:
Screenshot_20210427_185524

faho:
Alright, moved it over to `fish_config`. It's now `fish_config prompt <subcommand>`.

The web-ui can be accessed via `fish_config` or `fish_config browse`. Selecting the initial tab (which is a feature I doubt many people use or know about) can now only be done with `fish_config browse`.

faho:
Alright, selecting the initial tab is broken anyway (even without this). I'm just gonna remove it.

faho:
Alright, merged.

Issue #7960, opened 2021-04-28
Prompt turns into a mess when resizing with kitty terminal


tomKPZ:
Using the latest fish-shell built from ToT on Arch Linux.  Using oh-my-fish with the shellder prompt.  Please see this screencast:

https://user-images.githubusercontent.com/9124077/116347253-bd737b00-a7a0-11eb-9b8c-2b639910b50b.mp4

This change fixes the issue:
https://github.com/tomKPZ/fish-shell/commit/9ed1972f8c169b214591a92baf802200fc142eb1

PR #7961, opened 2021-04-28
Fix prompt reflow in kitty terminal

tomKPZ:
Fixes https://github.com/fish-shell/fish-shell/issues/7960


faho:
Yeah, merged, thanks!

In future: If you have a trivial PR there's no need to open a separate issue. We're not paid by the issue number and it really doesn't add much.

If you expect discussion of the *kind* of solution (so not slightly adjusting the check but doing an entirely *different* thing to solve the problem), then sure, open an issue. But this is fairly black and white so one thing open suffices.

PR #7981, opened 2021-05-09
Allow configuration of fish_clipboard{copy,paste}

neersighted:
## Description

This is a simple POC for making fish_clipboard_copy/fish_clipboard_paste
customizable. The default logic is sound, but can be enhanced by allowing a
custom choice of programs. This can be used to (for example) use
`reattach-to-user-namespace` on older macOS versions, or implement custom logic
like OSC 52-based clipboard access.

While the user can easily override the default functions with autoloads, the
normalization provided by the default functions is quite solid, and it's a shame
to force the user to duplicate it (and this not benefit from improvements) to
override the program selection logic.

The variable names are subject to change, and documentation and a change log entry
are still needed.

Additionally, we may want to add tests -- I will need guidance there.

## TODOs:

### DIFF
--- a/share/functions/fish_clipboard_copy.fish
+++ b/share/functions/fish_clipboard_copy.fish
@@ -2,7 +2,9 @@ function fish_clipboard_copy
     # Copy the current selection, or the entire commandline if that is empty.
     set -l cmdline (commandline --current-selection | string collect)
     test -n "$cmdline"; or set cmdline (commandline | string collect)
-    if type -q pbcopy
+    if set -q fish_user_clipboard_copy_command
+        printf '%s' $cmdline | $fish_user_clipboard_copy_command
+    else if type -q pbcopy
         printf '%s' $cmdline | pbcopy
     else if set -q WAYLAND_DISPLAY; and type -q wl-copy
         printf '%s' $cmdline | wl-copy
--- a/share/functions/fish_clipboard_paste.fish
+++ b/share/functions/fish_clipboard_paste.fish
@@ -1,6 +1,8 @@
 function fish_clipboard_paste
     set -l data
-    if type -q pbpaste
+    if set -q fish_user_clipboard_paste_command
+        set data ($fish_user_clipboard_paste_command 2>/dev/null)
+    else if type -q pbpaste
         set data (pbpaste 2>/dev/null)
     else if set -q WAYLAND_DISPLAY; and type -q wl-paste
         set data (wl-paste 2>/dev/null)


faho:
The way to configure this is to just edit the function, customization via variables doesn't really help.

Sorry!

neersighted:
> 
> 
> The way to configure this is to just edit the function, customization via variables doesn't really help.
> 
> Sorry!

The justification was this:

> While the user can easily override the default functions with autoloads, the
> normalization provided by the default functions is quite solid, and it's a shame
> to force the user to duplicate it (and this not benefit from improvements) to
> override the program selection logic.

Is the rationale here that it is preferred to force the user to duplicate the logic/overwrite the function, despite the above?

faho:
It is, yes.

Adding variables to everything doesn't scale. It just adds a lot of cases for us to document and for the user to read (and expect).

This is a leaf function that other things don't depend on. You can just have your own version of it.

neersighted:
Fair enough. I know generally we try to steer away from too many knobs -- I just find duplicating a complex function less fish-y than adding a knob. I consider this situation to be somewhat of an exception.

My current solution (to avoid duplicating the function) is to define `pbpaste`/`pbcopy` as a function.

Here's a compromise solution: would breaking out the program selection logic to be separate from the normalization logic be acceptable? That way only the parts most users will be interested in replacing can be easily overridden by an autoload.

faho:
>My current solution (to avoid duplicating the function) is to define pbpaste/pbcopy as a function.

What exactly are you doing with these?

Because if it's required to make them useful *anyway*, like AFAIK that "reattach-to-user-namespace" business is, then this is, in fact, the correct solution.

neersighted:
So there's a couple things that can be done here. `reattach-to-user-namespace` is not required on all macOS version -- recent version of macOS have made it unnecessary. I personally have not used it in quite some time, but it remains a common use case on pre-Cataline (IIRC) macOS.

Personally, I make use of [a script in my dotfiles](https://github.com/neersighted/dotfiles/blob/master/private_dot_local/bin/executable_yankee) that centralizes my copy/paste logic across programs and languages. This allows me to do neat tricks like use OSC 52 to set the clipboard over SSH, or manage tmux paste buffers. To make fish use this script, I either need to autoload `fish_clipboard_{copy,paste}` or resort to trickery like defining `pbcopy`/`pbpaste` since they come first in the dispatch logic.

faho:
>Here's a compromise solution: would breaking out the program selection logic to be separate from the normalization logic be acceptable? That way only the parts most users will be interested in replacing can be easily overridden by an autoload.

Okay, here's a compromise: Make it a function.

A variable isn't the correct choice here, because it can't represent any *logic*. E.g. see how `xclip` and `powershell.exe` need to be handled specially on paste:

```fish
    else if type -q xclip
        set data (xclip -selection clipboard -o 2>/dev/null)
    else if type -q powershell.exe
        set data (powershell.exe Get-Clipboard | string trim -r -c \r)
    end
```

That can't really be modelled with a variable.

Instead, if a thing called `clipboard_paste` exists, use that:

```fish
    set -l data
    if  type -q clipboard_paste
        set data (clipboard_paste)
    else if type -q pbpaste
        set data (pbpaste 2>/dev/null)
```

If anything needs to be done specially to format the data, let that be done there. Same for copy.

neersighted:
Yeah, I realized after I submitted this that it really made more sense as a set of functions. I'll send a follow-up with the function version.

PR #7983, opened 2021-05-09
Change ambiguous sentence in introduction

ripytide:
## Description

I think that the word "here" is slightly ambiguous and so I changed it to specifically refer to brace variable expansion in the variable expansion section of the introduction. That is unless I misunderstood the behavior of quote vs curly braces variable expansion as I am quite new to fish.

Fixes issue #

### DIFF
--- a/doc_src/language.rst
+++ b/doc_src/language.rst
@@ -516,7 +516,7 @@ Without the quotes or braces, fish will try to expand a variable called ``$WORDs
 
 The latter syntax ``{$WORD}`` is a special case of :ref:`brace expansion <expand-brace>`.
 
-If $WORD here is undefined or an empty list, the "s" is not printed. However, it is printed if $WORD is the empty string (like after ``set WORD ""``).
+If $WORD is undefined or an empty list, the "s" is printed after the quotes expansion but not after the braces expansion. However, it is printed if $WORD is the empty string (like after ``set WORD ""``).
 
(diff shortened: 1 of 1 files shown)


doc_src/language.rst
```diff
@@ -516,7 +516,7 @@ Without the quotes or braces, fish will try to expand a variable called ``$WORDs
 
 The latter syntax ``{$WORD}`` is a special case of :ref:`brace expansion <expand-brace>`.
 
-If $WORD here is undefined or an empty list, the "s" is not printed. However, it is printed if $WORD is the empty string (like after ``set WORD ""``).
+If $WORD is undefined or an empty list, the "s" is printed after the quotes expansion but not after the braces expansion. However, it is printed if $WORD is the empty string (like after ``set WORD ""``).
```
krobelus, line 519: > If $WORD is undefined or an empty list, the "s" is printed after the quotes expansion but not after the braces expansion

This is misleading, because nothing will actually be printed due to [Cartesian product expansion](https://fishshell.com/docs/current/index.html#combining-lists-cartesian-product).

What about just dropping this line instead? It describes what happens for some cases but the reasons are only given in the referenced sections (_expand-brace and _cartesian-product).

ripytide: I agree, it would be simpler to leave that particular effect to the relevant section, Cartesian product expansion, as it may over complicate the current topic, variable expansion.

Can I also ask about:

> The latter syntax ``{$WORD}`` is a special case of :ref:brace expansion <expand-brace>.

According to my, albeit rookie, understanding I was under the impression that {$WORD} is not a special case of brace expansion but just normal brace expansion with only was item.

ripytide: Fair enough.

faho:
Sorry, I don't think we want to drop this sentence.

PR #7984, opened 2021-05-09
Explanation of list range example wrong way round.

ripytide:
## Description

I'm assuming the first number before the *..* is the FROM and the number after it is the TO.

Fixes issue #


krobelus:
Good catch, thanks!

PR #7987, opened 2021-05-10
Not quite accurate code example heading

ripytide:
:mage:


faho:
That makes sense, merged!

faho:
As a general point: You don't need to open one PR per sentence, it probably makes more sense to batch a few, to cut down on the notifications and churn. Just general "documentation rewording" works.

ripytide:
Ok, sorry I will finish reading the introduction and add any more changes I may have in one PR.

PR #7989, opened 2021-05-10
Move list section upwards

ripytide:
For sure though I think it should be above the Index range expansion section.
I put the list section somewhere I thought was sensible but there may be better places.
Also I am not familiar with this (markdown-like?) notation so it might be worth checking I haven't messed something up when moving a whole section.


ripytide:
Whoops I'm bad at git hub I meant to do this on my fork first.

Issue #6001, opened 2019-07-22
completion for path should trigger for relative paths


chbiel:
fish version: 3.0.2
Terminal: xterm-256color
Ubuntu 19

Hi,
I currently try to get more familiar with fish completions. My goal to achieve:
I have a bash script `/home/auser/myscript` that can take arguments and options like `./myscript dosomething -s` or `./myscript dosomethingelse`

I would like to create fish autocompletions for that script.

I assumed that `complete -p /home/auser/myscript -f -a 'dosomething dosomethingelse' -d "Some comment"` should do this but I do not get autocomplete options when typing `./myscript <TAB>` or `/home/auser/myscript <TAB>`.

Do I missunderstand something here?

Thanks in advance for your help.

faho:
>I assumed that complete -p /home/auser/myscript -f -a 'dosomething dosomethingelse' -d "Some comment" should do this but I do not get autocomplete options when typing ./myscript <TAB> or /home/auser/myscript <TAB>.

This works for me with the full path (or via $PATH, if it's in), but not the relative `./`.

So it seems we don't resolve the path here. We should.

PR #7992, opened 2021-05-11
Resolve relative paths in complete

Caroleq:
## Description

PR resolves issue that completions for paths don't work for relative paths.  

Reason for a bug is that `path_get_path` does not return full path as it is told in the function description, but returns unmodified `cmd` or `path` + `cmd` where `path` is taken from `PATH` e.g. for `../../dir1/exec1` `path_get_path` will return `../../dir1/exec1`.  

From what I checked this function is used by (I skip functions which don't store path i.e. `out_path` is null and `parse_cmd_string`): 
- `populate_plain_process` passes value from  `path_get_path` to exec function. Relative path should not be changed to absolute here because e.g. whole command path for a process is later seen in process info (this can be checked e.g. with `ps -ef` on Linux)
- `builtin_command` uses value from  `path_get_path` to print command that would be executed - I would also rather leave it as it is now  

My proposal to solve it is to pass additional flag to `path_get_path` indicating if path should be returned as absolute. `parse_cmd_string` will always return absolute path which can be later matched with path completion if a one was defined. I didn't add function converting path to absolute in section checking if `cmd` is located in `$PATH`, because paths in `$PATH` should be absolute already and `cmd` doesn't have any slashes (otherwise it would be handled at the beginning of the function).  

Fixes issue #6001 

## TODOs:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -45,7 +45,7 @@ Interactive improvements
 - ``fish --private`` prints a note on private mode on startup even if ``$fish_greeting`` is an empty list (:issue:`7974`).
 - fish no longer attempts to lock history or universal variable files on remote filesystems, including NFS and SMB. In rare cases, updates to these files may be dropped if separate fish instances modify them simultaneously. (:issue:`7968`).
 - ``wait`` works correctly with jobs that have already exited (:issue:`7210`).
-- completion for paths will resolve relative paths (:issue: `6001`).
+- completion for paths will resolve relative paths (:issue:`6001`).
 
 New or improved bindings
 ^^^^^^^^^^^^^^^^^^^^^^^^
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -45,6 +45,7 @@ Interactive improvements
 - ``fish --private`` prints a note on private mode on startup even if ``$fish_greeting`` is an empty list (:issue:`7974`).
 - fish no longer attempts to lock history or universal variable files on remote filesystems, including NFS and SMB. In rare cases, updates to these files may be dropped if separate fish instances modify them simultaneously. (:issue:`7968`).
 - ``wait`` works correctly with jobs that have already exited (:issue:`7210`).
+- completion for paths will resolve relative paths (:issue: `6001`).
 
 New or improved bindings
 ^^^^^^^^^^^^^^^^^^^^^^^^
--- a/src/builtin_command.cpp
+++ b/src/builtin_command.cpp
@@ -103,7 +103,7 @@ maybe_t<int> builtin_command(parser_t &parser, io_streams_t &streams, const wcha
             }
         } else {  // Either find_path explicitly or just quiet.
             wcstring path;
-            if (path_get_path(command_name, &path, parser.vars(), true)) {
+            if (path_get_path(command_name, &path, parser.vars())) {
                 if (!opts.quiet) streams.out.append_format(L"%ls\n", path.c_str());
                 ++found;
             }
--- a/src/builtin_command.cpp
+++ b/src/builtin_command.cpp
@@ -103,7 +103,7 @@ maybe_t<int> builtin_command(parser_t &parser, io_streams_t &streams, const wcha
             }
         } else {  // Either find_path explicitly or just quiet.
             wcstring path;
-            if (path_get_path(command_name, &path, parser.vars())) {
+            if (path_get_path(command_name, &path, parser.vars(), true)) {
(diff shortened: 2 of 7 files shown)


CHANGELOG.rst
```diff
@@ -44,6 +44,7 @@ Interactive improvements
 - Variable ``fish_killring`` containing entries from killring is now available (:issue:`7445`).
 - ``fish --private`` prints a note on private mode on startup even if ``$fish_greeting`` is an empty list (:issue:`7974`).
 - fish no longer attempts to lock history or universal variable files on remote filesystems, including NFS and SMB. In rare cases, updates to these files may be dropped if separate fish instances modify them simultaneously. (:issue:`7968`).
+- completion for paths will resolve relative paths (:issue: `6001`).
```
faho, line 47: The space is wrong - it's 

```
:issue:`6001`
```

src/parse_execution.cpp
```diff
@@ -848,7 +848,7 @@ end_execution_reason_t parse_execution_context_t::populate_plain_process(
     wcstring path_to_external_command;
     if (process_type == process_type_t::external || process_type == process_type_t::exec) {
         // Determine the actual command. This may be an implicit cd.
-        bool has_command = path_get_path(cmd, &path_to_external_command, parser->vars());
+        bool has_command = path_get_path(cmd, &path_to_external_command, parser->vars(), true);
```
faho, line 851: Why does this use the absolute path?

src/builtin_command.cpp
```diff
@@ -103,7 +103,7 @@ maybe_t<int> builtin_command(parser_t &parser, io_streams_t &streams, const wcha
             }
         } else {  // Either find_path explicitly or just quiet.
             wcstring path;
-            if (path_get_path(command_name, &path, parser.vars())) {
+            if (path_get_path(command_name, &path, parser.vars(), true)) {
```
faho, line 106: Why does this use the absolute path? Currently

```fish
cd /usr/bin
command -s ./env
```

prints "./env". I don't know if that's of much *use*, but it doesn't appear to be obviously *wrong* either?

src/path.h
```diff
@@ -41,17 +41,19 @@ int path_get_config_is_remote();
...
+/// output_or_NULL - If non-NULL, store the path.
 /// vars - The environment variables to use
+/// absolute - Should returned path be absolute (if it is set to false, returned path may be either
+///            absolute or relative depending on the content of cmd argument)
 ///
 /// Returns:
 /// false if the command can not be found else true. The result
 /// should be freed with free().
-bool path_get_path(const wcstring &cmd, wcstring *out_path, const environment_t &vars);
+bool path_get_path(const wcstring &cmd, wcstring *out_path, const environment_t &vars, bool absolute);
```
faho, line 56: This is a case where I'd use a default argument - `bool absolute = false` and just leave it out in the spots you don't need it.

Then, where you do, you can comment it as

```fish
path_get_path(str, path, vars, /* absolute */ true)
```

Alternatively we could require those parts to do the `wrealpath` themselves. If it's only used in one or two places (and I count *one* where I'd agree that we need it) that might be preferable.

faho (changes requested):

src/builtin_command.cpp
krobelus: Yeah, I'd keep the current behavior unless there is a reason for changing this.

src/path.h
krobelus: Yeah I think we only want this in one place, so the caller should call `wrealpath`.

CHANGELOG.rst
Caroleq: Fixed. I'll remember for the future

src/parse_execution.cpp
Caroleq: :facepalm: It was meant be be false everywhere apart from `parse_cmd_string`. I put it there by mistake and didn't check it after. Sorry.

src/builtin_command.cpp
Caroleq: :facepalm: It was meant be be false everywhere apart from `parse_cmd_string`. I put it there by mistake and didn't check it after. Sorry.

src/path.h
Caroleq: Agreed, this makes more sense. I just removed information from `path_get_path` that it returns full path and call `wrealpath` in `parse_cmd_string`.

krobelus:
Merged as 31f3c1685 thanks! I tweaked the patch to avoid the `realpath` syscall in the common case (when we already have an absolute path).
Technically we should always resolve the path when `complete -p` is used, in case the commandline has an absolute but non-canonical path but I don't know if that matters.

PR #7999, opened 2021-05-14
completions: fix pacmd completions when using pipewire-pulse.

ericonr:
In such cases, `pacmd help` prints

  No PulseAudio daemon running, or not running as session daemon.

to stderr, which ends up printed to the user terminal.

Should I add some fallback for when `pacmd` is available but `pacmd help` failed? The completions still seem to work...

## TODOs:


faho:
Merged, thanks.

Note that I have no way to test this because arch packages "pacmd" along with pipewire and the packages conflict. I'm also not sure if pacmd isn't about to disappear.

PR #8004, opened 2021-05-15
Updated hashcat completions to version 6.2.1

exploide:
## Description

Just updated the hashcat completions to version 6.2.1

## TODOs:

### DIFF
--- a/share/completions/hashcat.fish
+++ b/share/completions/hashcat.fish
@@ -1,6 +1,6 @@
 function __fish_hashcat_types --description "Get hashcat hash types"
-    set -l modes (hashcat --example-hashes | string replace -f -r '^MODE: (\d+)' '$1')
-    set -l types (hashcat --example-hashes | string replace -f -r '^TYPE: (.+)' '$1')
+    set -l modes (hashcat --example-hashes | string replace -f -r '^(?:MODE: |Hash mode #)(\d+)' '$1')
+    set -l types (hashcat --example-hashes | string replace -f -r '^(?:TYPE:|\s+Name\.+:)\s+(.+)' '$1')
     for i in (seq (count $modes))
         echo -e "$modes[$i]\t$types[$i]"
     end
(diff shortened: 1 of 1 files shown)


share/completions/hashcat.fish
```diff
@@ -1,6 +1,6 @@
 function __fish_hashcat_types --description "Get hashcat hash types"
-    set -l modes (hashcat --example-hashes | string replace -f -r '^MODE: (\d+)' '$1')
-    set -l types (hashcat --example-hashes | string replace -f -r '^TYPE: (.+)' '$1')
+    set -l modes (hashcat --hash-info | string replace -f -r '^Hash mode #(\d+)' '$1')
+    set -l types (hashcat --hash-info | string replace -f -r '^\s+Name\.+:\s+(.+)' '$1')
```
faho, line 3: Oh noooooo.

When did this change? Is the old way still supported? Is anyone packaging hashcat versions < 6.2.1?

exploide: Ah, are you a hashcat user? I was wondering if someone stumbles upon this.

It's a recent change that is included since 6.2.0 was released yesterday. https://github.com/hashcat/hashcat/pull/2672

`--example-hashes` is still an alias for `--hash-info` but the format changed completely. It was requested to also include a machine readable format, but that hasn't been done yet.

I think only few distributions package hashcat in their official repositories (just a guess) and those who do will probably update quickly. I assume this will happen sooner than the next fish update. However, yes it is a breaking change and we could theoretically special-case the version numbers. But that doesn't make the code prettier and I still think people want to use the most recent hashcat anyway.

faho: >Ah, are you a hashcat user? I was wondering if someone stumbles upon this.

I'm not, it's just that this is awful to support for any third party (including, say, a shell that wishes to ship a completion script), because it's an either-or change.

>--example-hashes is still an alias for --hash-info but the format changed completely. It was requested to also include a machine readable format, but that hasn't been done yet.

So ideally this would be written so it can handle either format, which should be doable - roughly along these lines:

```fish
hashcat --example-hashes | string replace -f -r '^(?:MODE: |Hash mode #)(\d+)' '$1'
```

Use the old option (because that works in both, while the new one throws an error in the old version), and allow both strings. They don't overlap, so this shouldn't be an issue.

Similarly for types.

>I think only few distributions package hashcat in their official repositories (just a guess) and those who do will probably update quickly. I assume this will happen sooner than the next fish update. 

https://repology.org/project/hashcat/versions shows that a bunch of distributions package hashcat, among them the usual candidates for keeping old software alive - Debian Stable, Ubuntu LTS, FreeBSD Ports. The places where software goes to die a long and boring death, but somehow users insist on running *some* new software there.

I do not share your enthusiasm about it being updated quickly. I've been around the block a few times and I know that old distributions exist and are used and are a pain in the behind.

exploide: Ok, I completely underestimated how many distributions package old hashcat versions :D

> So ideally this would be written so it can handle either format

Have done that and force-pushed.

faho:
Merged, thanks!

PR #8011, opened 2021-05-17
Move the cursor to the beginning before printing right prompt

faho:
This makes the right prompt position independent of the width of the
commandline, which prevents staircase effects. That means, with "X"
standing in as a character that the terminal and fish disagree on:

```
> echo X           rightprompt
```

will stay like that instead of creating a staircase like

```
> echo X            rightpromp
t> echo X             rightpromp
pt> echo X
```

and so on.

The cursor still won't be *correct*, but it will be wrong in a less
annoying way.

A screenshot:

Screenshot_20210517_173936

Here fish and gnome-terminal disagree on the width of `☆` (U+2606 WHITE STAR). The first shell has the staircase effect, the second has this patch applied.

The cursor is still *wrong* - the final line has `echo ☆123`, and the cursor is logically *after the 3*. However, that's much less annoying than the staircase.

The cost of this is one cursor move to the start (a `\r`, typically) and longer move to the right prompt per printing of the right prompt, which should be acceptable.

## TODOs:

### DIFF
--- a/src/screen.cpp
+++ b/src/screen.cpp
@@ -877,6 +877,9 @@ static void s_update(screen_t *scr, const wcstring &left_prompt, const wcstring 
 
         // Output any rprompt if this is the first line.
         if (i == 0 && right_prompt_width > 0) {  //!OCLINT(Use early exit/continue)
+            // Move the cursor to the beginning of the line first to be independent of the width.
+            // This helps prevent staircase effects if fish and the terminal disagree.
+            s_move(scr, 0, 0);
             s_move(scr, static_cast<int>(screen_width - right_prompt_width), static_cast<int>(i));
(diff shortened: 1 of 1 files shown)


faho:
Weirdly enough this now means it works better *with* a right prompt.

I'm not *completely* sure why, it's possible we remove a dependency on "current_width" for something more. Just unconditionally doing the move back to 0 before the right prompt improves things also for situations without a right prompt. I imagine that's probably  too chatty, so maybe once before the right prompt and once after the last line?

ridiculousfish:
LGTM!

faho:
Merged with an additional move *after* printing everything, which appears to make fish almost usable with broken width.

Issue #7721, opened 2021-02-16
Handler for --on-job-exit $fish_pid runs constantly


faho:
As reported on gitter by @scorphus, another for the "fixed by status job-control full" (#5036) pile:

```fish
function on-exit --on-job-exit $fish_pid
    echo on-exit
end
```

Run this interactively and it will trigger constantly - presumably whenever you run the prompt or similar.

This is presumably because the prompt doesn't have "job-control" enabled and therefore runs with fish's pgroup.

While on-job-exit doesn't make sense here, it shouldn't trigger constantly.

mqudsi:
Fixed in https://github.com/fish-shell/fish-shell/commit/e7398c0248ccf1240c65ffe2f6f37ab634ec2337 but I fat-fingered the commit message and it auto-closed the wrong issue.

scorphus:
Thanks, @mqudsi

Issue #5832, opened 2019-04-20
$last_pid oddities in test context


mqudsi:
I'm trying to add an interactive test for the following behavior:

```fish
function disown_jobs
    sleep 1&
    set -l sleep_pid $last_pid
    jobs -c
    disown $sleep_pid
    jobs -c
end

disown_jobs
```

but it behaves differently when executed by the test driver, as best as I can tell due to an incorrect `last_pid`.

With this change to the tests:

```diff
diff --git a/tests/jobs.in b/tests/jobs.in
index 7be84be4b..adc947222 100644
--- a/tests/jobs.in
+++ b/tests/jobs.in
@@ -21,3 +21,18 @@ function foo
     jobs -c
 end
 foo
+
+killall -9 sleep
+
+# Verify `jobs` both works in a function and does *not* contain disowned jobs
+function disown_jobs
+    sleep 1&
+    set -l sleep_pid $last_pid
+    echo \$last_pid: $sleep_pid
+    echo pidof sleep: (pidof sleep)
+    jobs -c
+    disown $sleep_pid
+    jobs -c
+end
+
+disown_jobs
```

This is the output:

```diff
--- jobs.out    2019-04-19 19:26:51.102060200 -0500
+++ jobs.tmp.ou

t        2019-04-19 19:27:06.819177300 -0500
@@ -8,3 +8,9 @@
 sleep
 Command
 sleep
+$last_pid: 2261
+pidof sleep: 7554
+Command
+sleep
+Command
+sleep
Error output differs for file jobs.in. Diff follows:
--- jobs.err    2019-04-19 19:26:51.102060200 -0500
+++ jobs.tmp.err        2019-04-19 19:27:06.819177300 -0500
@@ -2,3 +2,4 @@
 fg: No suitable job: 3
 bg: Could not find job '3'
 disown: 'foo' is not a valid job specifier
+disown: Could not find job '2261'
```

But when executed in `build/fish`, the output is as follows:

```
mqudsi@ZBOOK ~/r/fish-shell> # Verify `jobs` both works in a function and does *not* contain disowned jobs
mqudsi@ZBOOK ~/r/fish-shell> function disown_jobs
                                     sleep 1&
                                     set -l sleep_pid $last_pid
                                     echo \$last_pid: $sleep_pid
                                     echo pidof sleep: (pidof sleep)
                                     jobs -c
                                     disown $sleep_pid
                                     jobs -c
                             end
[...]

Issue #5036, opened 2018-06-08
Set job-control to full (was: job control (%last/$last_pid) does not work within scripts)


deanveloper:
fish 2.7.1

Ubuntu 18.04, GNOME terminal (xterm-256color)

third party extensions do not affect the issue

I would expect `%last` to be different for each `sleep` call

# test.fish
```fish
 ><>  ~/Desktop > cat test.fish                                        2:29:12pm

for i in (seq 1 5)
	sleep 5 &
	echo %last
end
```

# Expected
```fish
 ><>  ~/Desktop > fish test.fish                                       2:29:14pm
30528
30529
30530
30531
30532
```
# Actual
```fish
 ><>  ~/Desktop > fish test.fish                                       2:29:14pm
30528
30528
30528
30528
30528
```

Using `(jobs -lp)` as a substitute for `%last` seems to work, although I would expect `%last` to work for this.

My use case for this is that I want to display an image which has some information on it, and I want to record the information within the image, then write that information to a file, close the image, and open the next one.

```fish
for image in (ls *.png)

    eog $image &
    set PID %last    # substituting this with (jobs -lp) makes it work, although %last should work fine...
    echo "What does the image say?"
    read info
    kill $PID
    echo "$image - $info" >> my.file
end
```

faho:
This is caused by "job-control" being enabled only for interactive jobs by default. `status` should print something to that effect.

If you do `status job-control full` at the beginning of the script, it should start working.

I have no idea what the idea behind that was.

@mqudsi: You've stared into this particular abyss for longer than I have. Any ideas why this is a thing?

mqudsi:
Before anything else, note that this usage of `%last` has been deprecated in fish 3.0 and won't work. Using `jobs` is far more explicit and addresses a lot of the issues with job and process expansion.

That said, I think the job control determinations date back to the original code? They're very.. messy. My guess is this is in place (as a hack?) to prevent termination of background jobs when fish is exiting when invoked as a script. It's also used (most definitely as a hack) in lieu of a proper model that determines what sort of interaction should be allowed between child programs and the tty.

faho:
>Before anything else, note that this usage of %last has been deprecated in fish 3.0 and won't work. Using jobs is far more explicit and addresses a lot of the issues with job and process expansion.

Oh, sure, that's correct. This would be `$last_pid` instead of `%last`.

[...]

PR #8021, opened 2021-05-21
Fix $last_pid and re-disable non-interactive job control

ridiculousfish:
This is my proposal to fix `$last_pid`, and then re-disable job control in scripts.

`$last_pid` (and `%last` before it) has always been the pgroup of the last job run in the background. If job control is active, then means the first pid in the pipeline. But if job control is not active, then the job runs in fish's pgroup, so `$last_pid` is the pid of fish itself! This is completely useless: you can't use wait, disown, kill, fg, bg, etc.

This was fixed in [this commit](https://github.com/fish-shell/fish-shell/commit/325599979479ae04ebac1d0556519f973058993e) which unconditionally enabled job control. This made `$last_pid` sane, but regressed signal handling in scripts (see my comment in the commit).

I propose making `$last_pid` always be a real pid in the job, or empty if the job has no pids. This will make `$last_pid` useful, and more closely match other shells, fixing #5036, #5832, #7721 in the same way.

If there is more than one pid in the job, we have to decide which one to use. I think we should just match bash and zsh, which uses the last. For example in `sleep 1 | sleep 2` then bash will set `$!` to the second sleep; fish will do the same. It does not matter very much which pid we choose, as all commands that operate on jobs allow you to pass the pid of any process in the pipeline.

The one place where it did matter was in `--on-job-exit` events, where you really did have to pass the pgroup (and if you passed fish's pgroup it would fire for every job without job control). To fix that, this change allows `--on-job-exit` events to use any pid in the job, matching fg, disown, etc.

This is technically a breaking change but the existing `$last_pid` behavior is pretty useless; I doubt anyone depends on it. Note if you really want the pgroup, you can use `jobs -g PID` to get it.

### pgroup and job control background

Some background on pgroups and job control: each process lives in a process group. Process groups are identified by a pid, which refers to the "leader" of the group. Process groups do two major things:

1. You can signal an entire pgroup. For example, when you run a shell script, control-C will signal the script itself, and also any child processes currently running as part of the script. This is because the script and its children live in the same pgroup, and control-C directs SIGINT to that group.

2. pgroups multiplex the tty. The tty is owned by one process group at any given time. If a process not in the group attempts to read or write from the tty, it gets signalled instead (SIGTTOU or SIGTITN).

Now when job control is enabled, the shell will place each job into its own process group. This is why control-Z will background vim and not the shell: vim is in its own pgroup. But if you are running a script, usually you want the script and all of its children to be cancelled or backgrounded together; this is accomplished by having them share a pgroup. So this explains why job control is enabled by default only in interactive mode.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -30,6 +30,7 @@ Scripting improvements
 - ``$last_pid`` now reports the pid of the last process in the pipeline, allowing it to be used in scripts (:issue:`5036`, :issue:`5832`, :issue:`7721`).
 - ``process-exit`` event handlers now receive the same value as ``$status`` in all cases, instead of receiving -1 when the exit was due to a signal.
 - ``process-exit`` event handlers for pid 0 also received ``JOB_EXIT`` events; this has been fixed.
+- ``job-exit`` event handlers may now be created with any of the pids from the job. The handler is passed the last pid in the job as its second argument, instead of the process group.
 
 Interactive improvements
 -------------------------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -38,7 +38,6 @@ Interactive improvements
 - Commands entered before the previous command finishes will now be properly syntax highlighted.
 - fish now automatically creates ``config.fish`` and the configuration directories in ``$XDG_CONFIG_HOME/fish`` (by default ``~/.config/fish``) if they do not already exist (:issue:`7402`).
 - ``__fish_prepend_sudo`` now toggles sudo even when it took the commandline from history instead of only adding it.
-- fish now defaults job-control to "full" meaning it more sensibly handles assigning the terminal and process groups (:issue:`5036`, :issue:`5832`, :issue:`7721`)
 - ``backward-kill-path-component`` :kbd:`Control-W`) no longer erases parts of two tokens when the cursor is positioned immediately after ``/``. (:issue:`6258`).
 - ``$SHLVL`` is no longer incremented in non-interactive shells. This means it won't be set to values larger than 1 just because your environment happens to run some scripts in $SHELL in its startup path (:issue:`7864`).
 - fish no longer rings the bell when flashing the command line. The flashing should already be enough notification and the bell can be annoying (:issue:`7875`).
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -27,6 +27,7 @@ Scripting improvements
 - ``fish_indent`` allows to write inline variable assignments on multiple lines (ending in a backslash), instead of joining them into one line (:issue:`7955`).
 - fish gained a ``--no-config`` option to disable reading the configuration. This applies both to the user's and the admin's config.fish (typically in /etc/fish/config.fish) and snippets and also sets $fish_function_path to just the functions shipped with fish and disables universal variables and history. (:issue:`7921`, :issue:`1256`).
 - When universal variables are unavailable for some reason, setting a universal variable now sets a global variable instead (:issue:`7921`).
+- ``$last_pid`` now reports the pid of the last process in the pipeline, allowing it to be used in scripts (:issue:`5036`, :issue:`5832`, :issue:`7721`).
 - ``process-exit`` event handlers now receive the same value as ``$status`` in all cases, instead of receiving -1 when the exit was due to a signal.
 - ``process-exit`` event handlers for pid 0 also received ``JOB_EXIT`` events; this has been fixed.
 
--- a/src/proc.cpp
+++ b/src/proc.cpp
@@ -953,6 +953,14 @@ bool job_t::is_foreground() const { return group->is_foreground(); }
 
 maybe_t<pid_t> job_t::get_pgid() const { return group->get_pgid(); }
 
+maybe_t<pid_t> job_t::get_last_pid() const {
+    for (auto iter = processes.rbegin(); iter != processes.rend(); ++iter) {
+        const process_t *proc = iter->get();
+        if (proc->pid > 0) return proc->pid;
+    }
+    return none();
+}
+
 job_id_t job_t::job_id() const { return group->get_id(); }
 
 void job_t::continue_job(parser_t &parser, bool in_foreground) {
--- a/src/proc.cpp
+++ b/src/proc.cpp
@@ -610,6 +610,7 @@ static void save_wait_handle_for_completed_job(const shared_ptr<job_t> &job,
     for (auto &proc : job->processes) {
         if (wait_handle_ref_t wh = proc->get_wait_handle(false /* create */)) {
             wh->status = proc->status.status_value();
+            wh->internal_job_id = job->internal_job_id;
             wh->completed = true;
         }
     }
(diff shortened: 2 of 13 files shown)


CHANGELOG.rst
```diff
@@ -27,8 +27,10 @@ Scripting improvements
 - ``fish_indent`` allows to write inline variable assignments on multiple lines (ending in a backslash), instead of joining them into one line (:issue:`7955`).
 - fish gained a ``--no-config`` option to disable reading the configuration. This applies both to the user's and the admin's config.fish (typically in /etc/fish/config.fish) and snippets and also sets $fish_function_path to just the functions shipped with fish and disables universal variables and history. (:issue:`7921`, :issue:`1256`).
 - When universal variables are unavailable for some reason, setting a universal variable now sets a global variable instead (:issue:`7921`).
+- ``$last_pid`` now reports the pid of the last process in the pipeline, allowing it to be used in scripts.
 - ``process-exit`` event handlers now receive the same value as ``$status`` in all cases, instead of receiving -1 when the exit was due to a signal.
 - ``process-exit`` event handlers for pid 0 also received ``JOB_EXIT`` events; this has been fixed.
+- ``job-exit`` event handlers may be created with any of the pids from the job, instead of just the process group.
```
faho, line 33: So the process group here still works because it's one of the PIDs, not because we also specifically look at the PGID? Do we then still need to mention the PGID at all?

src/proc.cpp
```diff
@@ -674,7 +675,7 @@ static bool process_clean_after_marking(parser_t &parser, bool allow_interactive
             // don't create an event or it's easy to get an infinite loop.
             if (!j->from_event_handler() && j->should_report_process_exits()) {
                 pid_t pgid = *j->get_pgid();
-                exit_events.push_back(event_t::job_exit(pgid));
+                exit_events.push_back(event_t::job_exit(pgid, j->internal_job_id));
```
faho, line 678: I'm not entirely sure why this still uses the pgid? The method itself has it renamed to "pid", so wouldn't that be the last pid then, now? Or is it even necessary at all to store it in this case, if we match all pids anyway?

CHANGELOG.rst
faho: Oh, also the reverted change should have the CHANGELOG entry removed.

zanchey:
It sounds like the obvious uses (`kill $last_pid`) will continue to work, but will now work in scripts as well (instead of killing the fish process)?

CHANGELOG.rst
ridiculousfish: Right. Good point, I deleted the mention of the process group. Also reverted the other changelog entry.

src/proc.cpp
ridiculousfish: It was historical, but you're right, nice catch. I switched this to passing the last pid instead of the process group.

ridiculousfish:
> It sounds like the obvious uses (`kill $last_pid`) will continue to work, but will now work in scripts as well (instead of killing the fish process)?

Yep precisely.

ridiculousfish:
Merged as 08950b10770da8fc4c2d4b1c9316d0cbf95bada0

zanchey:
The tests are failing following the merge of this branch on Ubuntu Xenial and CentOS 7 (GCC 5.4 and 4.8 respectively):

```
Testing file checks/noshebang.fish ... Failure:

  The CHECK on line 28 wants:
    0

    127

  Context:
    127 <= no check matches
    0
    127 <= does not match CHECK '0' on line 29
     <= nothing to match CHECK '0' on line 34
    0
    126
    127 <= does not match CHECK '126' on line 52
    127 <= no check matches this, previous check on line 52
    0
    127 <= does not match CHECK '0' on line 63
    126
    127 <= no check matches this, previous check on line 68
    126
     <= nothing to match CHECK '126' on line 81
     <= nothing to match CHECK '126' on line 86

  when running command:
    ../test/root/bin/fish checks/noshebang.fish
```

faho:
Ah, we've had this before - with redirections, it can sometimes happen that the file isn't directly executable afterwards.

The offending code: https://github.com/fish-shell/fish-shell/blob/34ededa644f0e868db142fdea8b7c5e5d20e93cd/tests/checks/noshebang.fish#L25-L29

This returns 127, i.e. "CMD_UNKNOWN" instead of 0, because it can't find the file yet. The other failures are similar.

Unfortunately I can't find the code or commit, but I'm quite certain we had to add a `sleep`.

faho:
@zanchey 7511de8 should fix this.

It's possible littlecheck could be improved here - we get a "text file busy" error here but swallow it because it has a catch-all CHECKERR already. Maybe we should print the error output as well, even if it matched?

zanchey:
Unfortunately not, it is still erroring out.

See https://launchpadlibrarian.net/540579769/buildlog_ubuntu-xenial-i386.fish_3.2.2-339-g7511de8d8-1~xenial_BUILDING.txt.gz for an example.

faho:
Aaaand another attempt. This time I'm assuming it's about the file not being *closed* by the time we try to run it.

ridiculousfish:
Ugh...ok it's not a timing issue. It's that fish's shebangless-logic doesn't kick in if posix_spawn is used. However posix_spawn has [its own shebangless logic](https://github.com/lattera/glibc/blob/895ef79e04a953cac1493863bcae29ad85657ee1/sysdeps/posix/spawni.c#L59) which does what fish wants; unfortunately it is recent enough that old glibc versions don't have it so we fail on older glibc.

Unfortunately posix_spawn really is an [important optimization](https://twitter.com/ridiculous_fish/status/1232889398485606400). I think we should just retry with fork() if posix_spawn fails. I will work on that tomorrow.

Note this can be reproduced locally if you can stomach Docker:

    ./docker/docker_run_tests.sh --shell-after ./docker/centos7.Dockerfile

you'll be dropped into a shell after running the tests (and they fail).

ridiculousfish:
I ended up fixing this (I hope!) in e74b9d53df886c9c03b89edb57a094354f09f35b by disabling posix_spawn on glibc < 2.24. The next commit c5ec4ef5f9782b6de68abc89f801067b7acc2dab reverts the prior attempts to fix this by adding sleeps.

zanchey:
Looks good!

floam:
> It's not easy to figure out when that happened but it definitely happens in glibc 2.28, and does not happen in glibc 2.17. Presumably the new behavior is present in glibc 2.24 (see BZ#23264) so that's the cutoff: posix_spawn is no longer allowed on glibc < 2.24.

@ridiculousfish here's the glibc commit where it changed:

[posix: New Linux posix_spawn{p} implementation](https://sourceware.org/git/gitweb.cgi?p=glibc.git;h=9ff72da471a509a8c19791efe469f47fa6977410)

floam:
HOWEVER! glibc has two implementations of posix_spawn in the tree, one is Linux-specific and they also have a unix/posix implementation. That one got a similar treatment in after the release of 2.25. 

https://sourceware.org/git/?p=glibc.git;a=commitdiff;h=ccfb2964726512f6669fea99a43afa714e2e6a80

FWIW, Python's subprocess.Popen at least at one point would only use posix_spawn() on macOS, glibc 2.26, or glibc 2.24 if Linux, and it's basically that same commit.

They say that spawn errors aren't properly reported to the parent process unless:

> os.posix_spawn() is available and properly reports errors to the
> parent process: macOS or glibc 2.26 and newer (or glibc 2.24 and
> newer on Linux).

PR #8030, opened 2021-05-24
env: remove trailing null-terminator from default path

d3dave:
## Description
The default value of the `PATH` variable when fish reads its global `config.fish` has a null byte at the end of the last element. You can see this by adding `set -S PATH` to the very beginning of `/usr/local/share/fish/config.fish`:

```
# Main file for fish command completions. This file contains various
# common helper functions for the command completions. All actual
# completions are located in the completions subdirectory.
#
# Set default field separators
#
set -S PATH
...
```

Then run `env -u PATH /usr/local/bin/fish -c ''`:

```
> env -u PATH /usr/local/bin/fish -c ''
$PATH: set in global scope, exported, a path variable with 4 elements
$PATH[1]: |/usr/bin|
$PATH[2]: |/bin|
$PATH[3]: |/usr/sbin|
$PATH[4]: |/sbin\x00|
...
```

Note `$PATH[4]`. This happens because `confstr` writes a null-terminator at the end of the provided buffer, and the length of the `cspath` string includes it. After the fix, `PATH[4]` no longer ends with `\x00`.

### DIFF
--- a/src/env.cpp
+++ b/src/env.cpp
@@ -244,7 +244,11 @@ static void setup_path() {
         // _CS_PATH: colon-separated paths to find POSIX utilities
         std::string cspath;
         cspath.resize(confstr(_CS_PATH, nullptr, 0));
-        confstr(_CS_PATH, &cspath[0], cspath.length());
+        if (cspath.length() > 0) {
+            confstr(_CS_PATH, &cspath[0], cspath.length());
+            // remove the trailing null-terminator
+            cspath.resize(cspath.length() - 1);
+        }
 #else
         std::string cspath = "/usr/bin:/bin";  // I doubt this is even necessary
 #endif


src/env.cpp
```diff
@@ -245,6 +245,8 @@ static void setup_path() {
         std::string cspath;
         cspath.resize(confstr(_CS_PATH, nullptr, 0));
         confstr(_CS_PATH, &cspath[0], cspath.length());
+        // remove the trailing null-terminator
+        cspath.resize(cspath.length() - 1);
```
faho, line 249: This needs to be guarded against length == 0.

faho:
Do you have a system that actually uses this value, or is this academic?

To be honest I'd always assumed that's basically the sign of a broken system, so I'd be interested to know where that's the case.

d3dave:
I stumbled across this while trying to debug another `PATH` issue I was having. I needed `fish` to not inherit my current `PATH` when invoked.

> To be honest I'd always assumed that's basically the sign of a broken system, so I'd be interested to know where that's the case.

Having an empty environment for a shell doesn't seem too insensible to me. It makes sense if the `PATH` is later configured within the global `config.fish`. It just boils down to expectations: if inheritance is expected, it's a fallback; if it's expected to be configured later, it's a default.

src/env.cpp
d3dave: Fixed, thanks 👍

faho:
Merged, thanks!

PR #8035, opened 2021-05-30
Add color to ls output on OpenBSD when colorls is installed

bonds:
## Description

Adds color to ls output on OpenBSD when colorls is installed.

Fixes issue #

## TODOs:

### DIFF
--- a/share/functions/ls.fish
+++ b/share/functions/ls.fish
@@ -17,6 +17,14 @@ function __fish_set_lscolors --description 'Set $LS_COLORS if possible'
     end
 end
 
+function __fish_ls_command --description 'Use colorls if it is installed'
+    if command -sq colorls
+        command colorls $argv
+    else
+        command ls $argv
+    end
+end
+
 function ls --description "List contents of directory"
     # Make ls use colors and show indicators if we are on a system that supports that feature and writing to stdout.
     #
@@ -24,17 +32,12 @@ function ls --description "List contents of directory"
     # BSD, macOS and others support colors with ls -G.
     # GNU ls and FreeBSD ls takes --color=auto. Order of this test is important because ls also takes -G but it has a different meaning.
     # Solaris 11's ls command takes a --color flag.
-    # OpenBSD requires the separate colorls program
+    # OpenBSD requires the separate colorls program for color support.
     # Also test -F because we'll want to define this function even with an ls that can't do colors (like NetBSD).
-    if not set -q __fish_ls_command
-        if which colorls >/dev/null 2>/dev/null
-            set -g __fish_ls_command colorls
-        else
-            set -g __fish_ls_command ls
-        end
+    if not set -q __fish_ls_color_opt
         set -g __fish_ls_color_opt
         for opt in --color=auto -G --color -F
-            if command $__fish_ls_command $opt / >/dev/null 2>/dev/null
+            if __fish_ls_command $opt / >/dev/null 2>/dev/null
                 set -g __fish_ls_color_opt $opt
                 break
             end
@@ -46,5 +49,5 @@ function ls --description "List contents of directory"
 
     isatty stdout
     and set -a opt -F
-    command $__fish_ls_command $__fish_ls_color_opt $argv
+    __fish_ls_command $__fish_ls_color_opt $argv
 end
--- a/share/functions/ls.fish
+++ b/share/functions/ls.fish
@@ -24,11 +24,17 @@ function ls --description "List contents of directory"
     # BSD, macOS and others support colors with ls -G.
     # GNU ls and FreeBSD ls takes --color=auto. Order of this test is important because ls also takes -G but it has a different meaning.
     # Solaris 11's ls command takes a --color flag.
+    # OpenBSD requires the separate colorls program
     # Also test -F because we'll want to define this function even with an ls that can't do colors (like NetBSD).
-    if not set -q __fish_ls_color_opt
+    if not set -q __fish_ls_command
+        if which colorls >/dev/null 2>/dev/null
+            set -g __fish_ls_command colorls
+        else
+            set -g __fish_ls_command ls
+        end
         set -g __fish_ls_color_opt
         for opt in --color=auto -G --color -F
-            if command ls $opt / >/dev/null 2>/dev/null
+            if command $__fish_ls_command $opt / >/dev/null 2>/dev/null
                 set -g __fish_ls_color_opt $opt
                 break
             end
@@ -40,5 +46,5 @@ function ls --description "List contents of directory"
 
     isatty stdout
     and set -a opt -F
-    command ls $__fish_ls_color_opt $argv
+    command $__fish_ls_command $__fish_ls_color_opt $argv
 end
--- a/share/functions/ls.fish
+++ b/share/functions/ls.fish
@@ -17,14 +17,6 @@ function __fish_set_lscolors --description 'Set $LS_COLORS if possible'
     end
 end
 
-function __fish_ls_command --description 'Use colorls if it is installed'
-    if command -sq colorls
-        command colorls $argv
-    else
-        command ls $argv
-    end
-end
-
 function ls --description "List contents of directory"
     # Make ls use colors and show indicators if we are on a system that supports that feature and writing to stdout.
     #
@@ -37,7 +29,7 @@ function ls --description "List contents of directory"
     if not set -q __fish_ls_color_opt
         set -g __fish_ls_color_opt
         for opt in --color=auto -G --color -F
-            if __fish_ls_command $opt / >/dev/null 2>/dev/null
+            if command ls $opt / >/dev/null 2>/dev/null
                 set -g __fish_ls_color_opt $opt
                 break
             end
@@ -49,5 +41,11 @@ function ls --description "List contents of directory"
 
     isatty stdout
     and set -a opt -F
-    __fish_ls_command $__fish_ls_color_opt $argv
+
+    if command -sq colorls
+        command colorls $__fish_ls_color_opt $argv
+    else
+        command ls $__fish_ls_color_opt $argv
+    end
+
 end
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -51,6 +51,7 @@ Interactive improvements
 - ``wait`` and ``on-process-exit`` work correctly with jobs that have already exited (:issue:`7210`).
 - Completion scripts are now loaded when calling a command via a relative path (like ``./git``) (:issue:`6001`, :issue:`7992`).
 - ``__fish_print_help`` (used for ``--help`` output for fish's builtins) now respects $LESS and uses a better default value (:issue:`7997`).
+- ls output is colorized on OpenBSD if colorls utility is installed
 
 New or improved bindings
 ^^^^^^^^^^^^^^^^^^^^^^^^
--- a/README.rst
+++ b/README.rst
@@ -104,6 +104,7 @@ The following optional features also have specific requirements:
    ``wl-copy``/``wl-paste`` or ``pbcopy``/``pbpaste`` utilities
 -  full completions for ``yarn`` and ``npm`` require the
    ``all-the-package-names`` NPM module
+-  ``colorls`` is used, if its installed, to add color when running ``ls``
 
 Switching to fish
 ~~~~~~~~~~~~~~~~~


share/functions/ls.fish
```diff
@@ -24,11 +24,17 @@ function ls --description "List contents of directory"
     # BSD, macOS and others support colors with ls -G.
     # GNU ls and FreeBSD ls takes --color=auto. Order of this test is important because ls also takes -G but it has a different meaning.
     # Solaris 11's ls command takes a --color flag.
+    # OpenBSD requires the separate colorls program
     # Also test -F because we'll want to define this function even with an ls that can't do colors (like NetBSD).
-    if not set -q __fish_ls_color_opt
+    if not set -q __fish_ls_command
+        if which colorls >/dev/null 2>/dev/null
```
faho, line 30: Please don't use `which`. `which` is an external tool that has a variety of issues.

Use `command -sq`, which is builtin.

faho: Also, honestly, there is no reason this needs to be cached in a cheesy global variable.

Just either check it right when you're about to run ls, or define a new function when `colorls` exists.

faho (changes requested):
Never use `which`. `which` is bad.

share/functions/ls.fish
bonds: Great, thank you for the feedback! I'll take another shot at this incorporating your feedback soon.

bonds:
Alright, updated to use ``command -sq`` instead of ``which``, and determine whether to use colorls every time ls is run instead of caching the answer.

IlanCosman:
Is "colorls" here referring to https://github.com/athityakumar/colorls? I believe that is essentially obsolete when things like [exa](https://github.com/ogham/exa) or [lsd](https://github.com/Peltoche/lsd) exist as much faster alternatives. Or does openbsd have some other program called colorls?

bonds:
``colorls`` is, I think, an OpenBSD specific package. Here's the package description:

````
Information for inst:colorls-6.5p0

Comment:
ls(1) that can use color to display file attributes

Description:
This is a simple hack, taken from FreeBSD, to OpenBSD's ls(1) to
use ANSI sequences to display file attributes in color.  There is
a -G flag (somewhat similar to the -F flag).  Take a look at the
man page for details.  The program is called "colorls", so you may
want to use an alias such as ls=/usr/local/bin/colorls.

Note that you need a color-capable terminal to enable colorls.  This
means you should set your TERM to "wsvt25" on the wscons(4) console
and to "sun-color" when using the Sun console, not "vt220" and
"sun", respectively, which are not color-capable in termcap(5).

Maintainer: Christian Weisgerber <naddy@openbsd.org>
````

And this is the man page:

````
COLORLS(1)		    General Commands Manual		    COLORLS(1)

NAME
     colorls  list directory contents

SYNOPSIS
     colorls [-1AaCcdFfGgHhikLlmnopqRrSsTtux] [file ...]

DESCRIPTION
     For each operand that names a file of a type other than directory,
     colorls displays its name as well as any requested, associated
     information.  For each named directory, colorls displays the names of
     files contained within that directory, as well as any requested,
     associated information.

     If no operands are given, the contents of the current directory are
     displayed.	 If more than one operand is given, non-directory operands are
     displayed first; directory and non-directory operands are sorted
     separately and in lexicographical order.  By default, colorls lists one
     entry per line to standard output; the exceptions are to terminals or
     when the -C, -m, or -x options are specified.

     The options are as follows:

     -1	     (The numeric digit one.) Force output to be one entry per line.
	     This is the default when output is not to a terminal.

     -A	     List all entries except for . and ...  Always set for the
	     superuser.

     -a	     Include directory entries whose names begin with a dot (.).

     -C	     Force multi-column output; this is the default when output is to
	     a terminal.

     -c	     Use time file's status was last changed instead of last
	     modification time for sorting (-t) or printing (-g, -l, or -n).

     -d	     Directories are listed as plain files (not searched recursively)
	     and symbolic links in the argument list are not indirected
	     through.

     -F	     Display a slash (/) immediately after each pathname that is a
	     directory, an asterisk (*) after each that is executable, an at
	     sign (@) after each symbolic link, an equal sign (=) after
	     each socket, and a vertical bar (|) after each that is a FIFO.

     -f	     Output is not sorted.  This option implies -a.

     -G	     Enable colorized output.  This option is equivalent to defining
	     CLICOLOR in the environment.  (See below.)

     -g	     List in long format as in -l, except that the owner is not
	     printed.

     -H	     Follow symbolic links specified on the command line.  This is the
	     default behaviour when none of the -d, -F, or -l options are
	     specified.

     -h	     When used with a long format option, use unit suffixes: Byte,
	     Kilobyte, Megabyte, Gigabyte, Terabyte, Petabyte, and Exabyte in
	     order to reduce the number of digits to four or fewer using
	     powers of 2 for sizes (K=1024, M=1048576, etc.).

     -i	     For each file, print its inode number.

     -k	     Modifies the -s option, causing the sizes to be reported in
	     kilobytes.	 Overrides any value specified by the BLOCKSIZE
	     environment variable.

     -L	     If argument is a symbolic link, evaluate the file information and
	     file type to be those of the file referenced by the link, and not
	     the link itself; however, colorls writes the name of the link
	     itself and not the file referenced by the link.

     -l	     (The lowercase letter ell.) List in long format (see below).  A
	     total sum of all file sizes is output on a line before the long
	     listing.  Output is one entry per line.

     -m	     Stream output format; list files across the page, separated by
	     commas.

     -n	     List in long format as in -l, but retain user and group IDs in a
	     numeric format.  The output of -gn and -ng is identical: a long
	     listing with numerical group ID, and no numerical user ID.	 The
	     output of -ln and -nl is identical: a long listing with numerical
	     group and user ID.

     -o	     Include the file flags in a long format (-g, -l, or -n) output.

     -p	     Display a slash (/) immediately after each pathname that is a
	     directory.

     -q	     Force printing of non-graphic characters in file names as the
	     character ?; this is the default when output is to a terminal.

     -R	     Recursively list subdirectories encountered.

     -r	     Reverse the order of the sort to get reverse lexicographical
	     order or the smallest or oldest entries first.

     -S	     Sort by size, largest file first.

     -s	     Display the number of file system blocks actually used by each
	     file, where partial units are rounded up to the next integer
	     value.  Blocks are 512 bytes unless overridden by the -k flag or
	     BLOCKSIZE environment variable.

     -T	     Display complete time information for the file, including month,
	     day, hour, minute, second, and year.  This option has no effect
	     unless one of the long format (-g, -l, or -n) options is also
	     specified.

     -t	     Sort by time modified (most recently modified first) before
	     sorting the operands in lexicographical order.

     -u	     Use file's last access time instead of last modification time for
	     sorting (-t) or printing (-g, -l, or -n).

     -x	     Multi-column output sorted across the page rather than down the
	     page.

     It is not an error to specify more than one of the following mutually
     exclusive options: -1, -C, -g, -l, -m, -n, and -x; and -c, -f, -S, -t,
     and -u.  Where more than one option is specified from the same mutually
     exclusive group, the last option given overrides the others, except that
     -l always overrides -g; and -f always overrides -c, -S, -t, and -u.

   The Long Format
     If the -g, -l, or -n options are given, the following information is
     displayed for each file: mode, number of links, owner (though not for
     -g), group, size in bytes, time of last modification (mmm dd HH:MM),
     and the pathname.	In addition, for each directory whose contents are
     displayed, the first line displayed is the total number of blocks used by
     the files in the directory.  Blocks are 512 bytes unless overridden by
     the -k option or BLOCKSIZE environment variable.

     If the owner or group name is not a known user or group name,
     respectively, or the -n option is given, the numeric ID is displayed.

     If the file is a character special or block special file, the major and
     minor device numbers for the file are displayed in the size field.

     If the -T option is given, the time of last modification is displayed
     using the format mmm dd HH:MM:SS ccyy.

     If the file is a symbolic link, the pathname of the linked-to file is
     preceded by ->.

     The file mode printed under the -g, -l, or -n options consists of the
     entry type, owner permissions, group permissions, and other permissions.
     The entry type character describes the type of file, as follows:

	   -	 regular file
	   b	 block special file
	   c	 character special file
	   d	 directory
	   l	 symbolic link
	   p	 FIFO
	   s	 socket link

     The next three fields are three characters each: owner permissions, group
     permissions, and other permissions.  Each field has three character
     positions:

	   1.	If r, the file is readable; if -, it is not readable.
	   2.	If w, the file is writable; if -, it is not writable.
	   3.	The first of the following that applies:

		      S	    If in the owner permissions, the file is not
			    executable and set-user-ID mode is set.  If in the
			    group permissions, the file is not executable and
			    set-group-ID mode is set.

		      s	    If in the owner permissions, the file is
			    executable and set-user-ID mode is set.  If in the
			    group permissions, the file is executable and set-
			    group-ID mode is set.

		      x	    The file is executable or the directory is
			    searchable.

		      -	    The file is neither readable, writable,
			    executable, nor set-user-ID, nor set-group-ID, nor
			    sticky (see below).

		These next two apply only to the third character in the last
		group (other permissions):

		      T	    The sticky bit is set (mode 1000), but neither
			    executable nor searchable (see chmod(1) or
			    sticky(8)).

		      t	    The sticky bit is set (mode 1000), and is
			    searchable or executable (see chmod(1) or
			    sticky(8)).

     In addition, if the -o option is specified, the file flags (see
     chflags(1)) are displayed as comma-separated strings in front of the file
     size, abbreviated as follows:

	   -	     no flags
	   arch	     archived
	   nodump    do not dump
	   sappnd    system append-only
	   schg	     system immutable
	   uappnd    user append-only
	   uchg	     user immutable

ENVIRONMENT
     BLOCKSIZE	     If the environment variable BLOCKSIZE is set, and the -k
		     option is not specified, the block counts (see -s) will
		     be displayed in units of that size block.

     CLICOLOR	     Use ANSI color sequences to distinguish file types.  See
		     LSCOLORS below.  In addition to the file types mentioned
		     in the -F option some extra attributes (setuid bit set,
		     etc.) are also displayed.	The colorization is dependent
		     on a terminal type with the proper termcap(5)
		     capabilities.  To display the colors on the wscons(4)
		     console, for example, the TERM variable must be set to
		     wsvt25.  Other terminal types may require similar
		     adjustments.  Colorization is silently disabled if the
		     output isn't directed to a terminal unless the
		     CLICOLOR_FORCE variable is defined.

     CLICOLOR_FORCE  Color sequences are normally disabled if the output isn't
		     directed to a terminal.  This can be overridden by
		     setting this flag.	 The TERM variable still needs to
		     reference a color capable terminal however otherwise it
		     is not possible to determine which color sequences to
		     use.

     COLUMNS	     If set to a positive integer, output is formatted to the
		     given width in columns.  Otherwise, colorls defaults to
		     the terminal width, or 80 columns if the output is not a
		     terminal.

     LC_CTYPE	     The character encoding locale(1).	It decides which byte
		     sequences form characters and what their display width
		     is.  If unset or set to "C", "POSIX", or an unsupported
		     value, non-ASCII bytes are replaced by question marks.

     LSCOLORS	     The value of this variable describes what color to use
		     for which attribute when colors are enabled with
		     CLICOLOR.	This string is a concatenation of pairs of the
		     format fb, where f is the foreground color and b is the
		     background color.

		     The color designators are as follows:

			   a	 black
			   b	 red
			   c	 green
			   d	 brown
			   e	 blue
			   f	 magenta
			   g	 cyan
			   h	 light grey
			   A	 bold black, usually shows up as dark grey
			   B	 bold red
			   C	 bold green
			   D	 bold brown, usually shows up as yellow
			   E	 bold blue
			   F	 bold magenta
			   G	 bold cyan
			   H	 bold light grey; looks like bright white
			   x	 default foreground or background

		     Note that the above are standard ANSI colors.  The actual
		     display may differ depending on the color capabilities of
		     the terminal in use.

		     The order of the attributes are as follows:

			   1.	directory
			   2.	symbolic link
			   3.	socket
			   4.	pipe
			   5.	executable
			   6.	block special
			   7.	character special
			   8.	executable with setuid bit set
			   9.	executable with setgid bit set
			   10.	directory writable to others, with sticky bit
			   11.	directory writable to others, without sticky
				bit

		     The default is "exfxcxdxbxegedabagacad", i.e., blue
		     foreground and default background for regular
		     directories, black foreground and red background for
		     setuid executables, etc.

     TERM	     The CLICOLOR functionality depends on a terminal type
		     with color capabilities.

     TZ		     The time zone to use when displaying dates.  See
		     environ(7) for more information.

EXIT STATUS
     The colorls utility exits0 on success, and>0 if an error occurs.

EXAMPLES
     List the contents of the current working directory in long format:

	   $ colorls -l

     In addition to listing the contents of the current working directory in
     long format, show inode numbers, file flags (see chflags(1)), and suffix
     each filename with a symbol representing its file type:

	   $ colorls -lioF

     List the files in /var/log, sorting the output such that the most
     recently modified entries are printed first:

	   $ colorls -lt /var/log

SEE ALSO
     chflags(1), chmod(1), termcap(5), symlink(7), sticky(8)

STANDARDS
     The ls(1) utility is compliant with the IEEE Std 1003.1-2008 (POSIX.1)
     specification, except behaviour for the -o flag differs.

     The flags [-GhT], as well as the BLOCKSIZE environment variable and
     colorls itself, are extensions to that specification.

     The flags [-go] are marked by IEEE Std 1003.1-2008 (POSIX.1) as being
     an X/Open System Interfaces option.

     Historically, the -g flag was used to specify that the group field be
     included in long listings.	 The group field is now automatically included
     in the long listing for files and the meaning of the -g flag has been
     changed in order to be compatible with the IEEE Std 1003.1-2008
     (POSIX.1) specification.

HISTORY
     An ls utility appeared in Version1 AT&T UNIX.

OpenBSD 6.9		       October 24, 2016			   OpenBSD 6.9
````

File: share/functions/ls.fish [hunk 2]
```diff
@@ -17,18 +17,27 @@ function __fish_set_lscolors --description 'Set $LS_COLORS if possible'
     end
 end
 
+function __fish_ls_command --description 'Use colorls if it is installed'
```
faho, line 20: I probably should have been clearer: I meant either do this check in the actual `ls` function, or define an entirely different `ls` function that uses `colorls` if available. But now I see that you'd still want to call `colorls -F` (I assumed it did *just* color), so it's probably simplest to just inline this check in the actual `ls` function.

This indirection is just as cheesy as having a global variable.

bonds: Got it. I was thinking it would be preferable to use the color __fish_ls_color_opt code to test ``colorls`` for supported options before using them, in case some other version of ``colorls`` exists in the world and behaves differently, but perhaps I was over engineering it...if such a ``colorls`` exists, it could very well have different options behaviors that would need to be dealt with. Better to just wait for those to come up than to guess.

Anyhoo, I'll take another shot it. Thanks again for the feedback. :)

faho:
Merged, thanks!

bonds:
You beat me to one more fix I was working on: https://github.com/bonds/fish-shell/commit/fd228351a9254ff9d97232dca42d7bcfb870330a

Should I open a new PR for that?

faho:
>Should I open a new PR for that?

Nah, cherry-picked as ad387307926b3b6fcbadee8370e5a015d4678b53.

kidonng:
@IlanCosman I'm thinking about the same thing. What if people has https://github.com/athityakumar/colorls in path? I tried installing `colorls` and applying this patch, then I got:

```
$ ls
colorls: invalid option: -F
See 'colorls --help'.
```

bonds:
Hmm. Well, in a previous version of this PR I tested the chosen ls command with different parameters to check that they were at least supported...that would avoid this problem. I didn't keep that version because @faho didn't like the global function I wrote to reuse my check for colorls's existence, but perhaps a version that does the same thing without a global function would be good.

faho:
Fixed in a633889244a0db878c5ef0eb097cacc8b9a56a04, we now ignore a "colorls" if it doesn't understand "-F". Which ignores the ruby colorls, but that seems to be closer in spirit to exa than normal ls.

bonds:
Thanks @faho !

PR #8039, opened 2021-06-01
fish_vi_key_bindings: bind s to follow vim visual behaviour

mtoohey31:
## Description

This change adds a binding that sets the s key's behaviour to match the c key's in visual mode. This mirrors vim's behaviour (see `:h v_s` in vim or neovim). It's probably not something anyone other than me uses since I've gotten into the bad habit of using s for changing in visual mode, but it is how vim works and s is current unbound in visual mode, so I figure it might as well be supported.

~~Fixes issue #~~

## TODOs:

Not sure which of these ^ are applicable... let me know if there's anything else I need to do though!

### DIFF
--- a/share/functions/fish_vi_key_bindings.fish
+++ b/share/functions/fish_vi_key_bindings.fish
@@ -302,6 +302,7 @@ function fish_vi_key_bindings --description 'vi-like key bindings for fish'
     end
 
     bind -s --preset -M visual -m insert c kill-selection end-selection repaint-mode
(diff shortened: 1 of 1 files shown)


faho:
Sure, why not.

>Not sure which of these ^ are applicable... let me know if there's anything else I need to do though!

Honestly this doesn't really need testing, I'm adding the docs and changelog.

mtoohey31:
Awesome, thanks!

PR #8041, opened 2021-06-01
Fallback to `cmd --help` when `man cmd` errors

cljoly:
## Description

In __fish_man_page.fish, call `cmd --help` (resp. `cmd sub --help`) when `man
cmd` (resp. `man cmd-sub`) does not exist.

I’m unsure if this is worth adding an entry in CHANGELOG.rst

## TODOs:

### DIFF
--- a/share/functions/__fish_man_page.fish
+++ b/share/functions/__fish_man_page.fish
@@ -27,12 +27,16 @@ function __fish_man_page
             man "$maincmd-$args[2]"
         else if man "$maincmd" &>/dev/null
             man "$maincmd"
+        else if "$maincmd" "$args[2]" --help &>/dev/null
+            "$maincmd" "$args[2]" --help | less
         else
             printf \a
         end
     else
         if man "$maincmd" &>/dev/null
             man "$maincmd"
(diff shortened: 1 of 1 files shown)


faho:
Calling command here in any way is *dangerous*. If the command does not take `--help` as an option you've now done something.

E.g. Solaris' `killall` ignores all arguments and kills all processes (IIRC).

This is not a feature that should be added, sorry!

zanchey:
> E.g. Solaris' `killall` ignores all arguments and kills all processes (IIRC).

Yes. More specifically, it kills all processes belonging to a user, problematic if you are logged in as root. Ask me how I know. `routef` is another landmine for the unwary.

cljoly:
Right, that makes sense.
How would you feel about having a prompt like (the `if cmd --help` part would be removed):
```
Do you want to see `cmd --help`? [Y/n]
```
The main drawback I see is that it makes calls to `__fish_man_page` blocking for the user.

zanchey:
That is not a question that users can give a useful answer to unless they are already sure that the command is safe.

cljoly:
That’s a good point.

PR #8042, opened 2021-06-01
Ignore __fish_ls_color_opt when using colorls

bonds:
## Description

Ignore __fish_ls_color_opt when using colorls, otherwise we're assuming ``ls`` and ``colorls`` respond to the same command line options, which is definitely not the case. Sorry for not catching this 10 minutes earlier @faho :|

Fixes issue #

## TODOs:


bonds:
Not needed, @faho cherry picked the fix.

Issue #6764, opened 2020-03-18
vcs prompt blocks on slow filesystems


cg505:
The default prompt in fish 3.1.0 has VCS info in the prompt. This is cool! But, sometime I am working on filesystems (like sshfs) that are VERY slow. Basic git operations can take seconds. This can also apply to huge git repositories (think chromium) on spinning rust. This means that I have to wait on the prompt to show to do anything.

Of course, I understand that this is a relatively rare scenario, and I can just disable the vcs portion of the prompt. However, it would seem reasonable to **set a very short timeout (100ms?) on these commands as protection against these issues in general**. This would also help with issues like #6681.

I am totally unfamiliar with the fish codebase but would be willing to help with this if it seems like a reasonable idea. Thanks so much for developing this awesome piece of software.

krobelus:
Setting a timeout for the prompt seems reasonable. Maybe have each fish process automatically redefine (process-scoped) `fish_prompt` with a safe fallback prompt after the first prompt times out. Also fish should print a small note to explain why the prompt changed.
Not sure how complex this'd be. It probably has to be implemented in fish, I don't think it'd be realistic to use an external command like `timeout`.

MaxGyver83:
I have the same problem, except that I'm not working on a slow filesystem.

Sometimes, when I do `cd ~/huge-repository`, the shell is blocked for about a minute. Especially after the weekend (this is at work). It's possible that I'm more than 1000 commits behind `master`. I have started `fish` with `-p` before changing into that repository folder. The culprit is `git -c core.fsmonitor= status --porcelain -z $untr | string split0`:

```
Time      Sum       Command
...       ...       ...
15        15        ---------> test "$untracked" = true
162       63251235  ---------> set -l stats (string sub -l 2 (git -c core.fsmonitor= status --porcelain -z $untr | string split0))
264       63251073  ----------> string sub -l 2 (git -c core.fsmonitor= status --porcelain -z $untr | string split0)
63250809  63250809  -----------> git -c core.fsmonitor= status --porcelain -z $untr | string split0
156       312       ---------> set -l invalidstate (string match -r '^UU' $stats | count)
156       156       ----------> string match -r '^UU' $stats | count
106       168       ---------> set -l stagedstate (string match -r '^[ACDMR].' $stats | count)
```

I'm already running `git pull` in a cron job every 15 minutes to keep the delay small. 

[...]

PR #8050, opened 2021-06-08
RFC git prompt: Cache checked out commit description

Werkov:
## Description

Describing tags in repos with many tags may take unconvenient amount of
time. The information is valid until HEAD is switched to something else
therefore we cache the resolved name and use file modification timestamp
for invalidation.

This is complementary to #6764, the caching can be used even with timeout but this patch is specific to a particular git variable.
Because of the existence of this issue and caching possibly controversial, I'm sending this just as RFC to gather feedback/hints if caching is the way to go.

## TODOs:

### DIFF
--- a/share/functions/fish_git_prompt.fish
+++ b/share/functions/fish_git_prompt.fish
@@ -460,28 +460,48 @@ function __fish_git_prompt_operation_branch_bare --description "fish_git_prompt 
     end
 
     if test -z "$branch"
-        if not set branch (command git symbolic-ref HEAD 2>/dev/null)
-            set detached yes
-            set branch (switch "$__fish_git_prompt_describe_style"
-						case contains
-							command git describe --contains HEAD
-						case branch
-							command git describe --contains --all HEAD
-						case describe
-							command git describe HEAD
-						case default '*'
-							command git describe --tags --exact-match HEAD
-						end 2>/dev/null)
-            if test $status -ne 0
-                # Shorten the sha ourselves to 8 characters - this should be good for most repositories,
-                # and even for large ones it should be good for most commits
-                if set -q sha
-                    set branch (string match -r '^.{8}' -- $sha)…
-                else
-                    set branch unknown
+        set -l cache $HOME/.cache/fish/
+        if test -n "$XDG_CACHE_HOME"
+            set cache $XDG_CACHE_HOME/fish
+        end
+        mkdir -p $cache
+        set -l head_file (realpath $git_dir/HEAD)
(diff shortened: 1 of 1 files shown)


faho:
I'm not opposed to the idea of caching, and this seems reasonably doable (unlike e.g. the number of untracked files). 

However, I have never seen this bit take any interesting amount of time. (it's always been things like the number of untracked files)

Do you have any benchmarks where it's an issue? E.g. `fish --profile /tmp/thing.prof -c __fish_git_prompt` in some specific git repo. When I do that in the fish-shell repo, I get the entire `__fish_git_operation_branch_bare` at 2271 *microseconds* of the total 32915. Similar in the llvm-project, freebsd ports and linux repos. And importantly: This time doesn't seem to significantly increase in those larger repos. It stays at the 2-3 millisecond mark.

Unless you have different measurements I'd skip the caching as it doesn't appear to be worth it.

marcusatiliusregulus:
> The information is valid until HEAD is switched to something else

you sure about that?

```fish
$ git tag --contains HEAD
tagtag
$ git rev-parse HEAD
cd0e794370b21ac91efd73ad0e40389a0d04acf7
$ git tag -d tagtag
Deleted tag 'tagtag' (was cd0e794)
$ git rev-parse HEAD
cd0e794370b21ac91efd73ad0e40389a0d04acf7
```

Werkov:
Excerpt from the `sort -rn -k2 /tmp/thing.prof` (in the detached head state, on a tag):

	534     1021306 > __fish_git_prompt
	222     1020722 -> fish_git_prompt $argv
	306     980664  --> set -l rbc (__fish_git_prompt_operation_branch_bare $repo_info)
	188     980358  ---> __fish_git_prompt_operation_branch_bare $repo_info
	65      979582  ----> if test -z "$branch"...
	11      976673  -----> if test $rel_age -le 0...
	24      976399  ------> if not set branch (command git symbolic-ref HEAD 2>/dev/null)...
	442     974236  -------> set branch (switch "$__fish_git_prompt_describe_style"
	76      973794  --------> switch "$__fish_git_prompt_describe_style"
	973718  973718  ---------> command git describe --tags --exact-match HEAD

and I'm quite confident this may not affect everyone

	$ git tag -l | wc -l
	11548

Thanks @marcusatiliusregulus, you're right, I haven't consider such a case.

For now, I reckon this isn't worth having this caching in the fish upstream and I'll make do with a local `__fish_git_prompt_operation_branch_bare` override.

Issue #159, opened 2012-06-20
Cannot save multi-line output in a variable (support $() command substitution syntax)


lilyball:
There seems to be no way to capture multi-line output in a shell variable. Any attempt to do so splits each line into a separate array element. In bash I'd simply put double-quotes around my `$()` invocation, but in fish you cannot quote a command substitution. This is rather unfortunate because it means I cannot capture multi-line output in a variable and send it back to a separate command without really weird contrivances, such as replacing all newlines with NUL when saving and reverting the process when emitting.

derekstavis:
Another great syntax that runs away from `$` is what Swift proposes for string interpolation:

    set CPUINFO "\(cat /proc/cpuinfo)"

jakwings:
I'm opposed to `((...))`, `#{...}`, `{...}`, `"\(...)"`. And `"$(...)"` is the best. If we also change the syntax of unquoted command substitution to `$(...)`, that would be better. Then we can use `()` in `test (...) -a (...)` without escaping.

imgx64:
Maybe I'm missing something, but isn't this already fixed?

From [the documentation](http://fishshell.com/docs/current/index.html#expand-command-substitution):

> If the output is more than one line long, each line will be expanded to a new parameter. Setting IFS to the empty string will disable line splitting.

So the solution is:
```
> count (echo one\ntwo\nthree)
3
> begin; set -l IFS; count (echo one\ntwo\nthree); end
1
```

faho:
@imgx64: That has been mentioned here before.

While mucking with $IFS does solve the issue, it's supremely awkward.

dxlr8r:
```
# echo 1 2 3 | read -z flatvar; count $flatvar
1
```

anordal:
While `read -z` reads up to NUL, an option to read all the way to EOF would be a more proper solution for this use case.

andrewrk:
What's preventing this issue from being resolved? It seems like several reasonable paths forward gave been proposed. Can we pick one and move on with it?

I'll throw in a vote for deprecating `()` in favor of `$()`.

faho:
>What's preventing this issue from being resolved? It seems like several reasonable paths forward have been proposed. Can we pick one and move on with it?

Well, we'd have to actually pick one and then someone has to write the code. Or someone writes code that we can agree on.

----

Personally, I now rather strongly believe that switching to "$()" is the right solution.

[...]

PR #8054, opened 2021-06-11
string: Allow `collect --allow-empty` to avoid empty ellision

faho:
Currently we still have that issue where

    test -n (thing | string collect)

can return true if `thing` doesn't print anything, because the
collected argument will still be removed.

So, what we do is allow ~~`--no-empty`~~ `--allow-empty` to be used, in which case we
print one empty argument.

This means

    test -n (thing | string collect -a)

can now be safely used.

"~~no~~allow-empty" isn't the best name for this flag, but string's design
really incentivizes reusing names, and it's not *terrible*.

## TODOs:

### DIFF
--- a/doc_src/cmds/string-collect.rst
+++ b/doc_src/cmds/string-collect.rst
@@ -8,7 +8,7 @@ Synopsis
 
 ::
 
-    string collect [(-N | --no-trim-newlines)] [STRING...]
+    string collect [(-N | --no-trim-newlines)] [(-n | --no-empty)] [STRING...]
 
 .. END SYNOPSIS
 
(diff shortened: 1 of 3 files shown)


faho:
>"no-empty" isn't the best name for this flag, but string's design
really incentivizes reusing names, and it's not terrible.

I have now switched to using "--allow-empty" - because "--no-empty" does the opposite in split/split0 (it causes them to *remove* empty arguments)

ridiculousfish:
An alternative to this is to support quoted substitutions, old #159:

    test -n "$(thing | string collect)"

the advantage here is it works for all commands, not just `string`. But I don't mind `--no-empty`

faho:
>An alternative to this is to support quoted substitutions, old #159:

Oh, sure. And that's something I'd eventually like to have. But I've looked into it, and it's not trivial to add.

While this exists *now*.

ridiculousfish:
Ok, seems reasonable; go for it.

Aftabalamkhan120:
> >"no-empty" isn't the best name for this flag, but string's design
> 
> really incentivizes reusing names, and it's not terrible.
> 
> 
> 
> I have now switched to using "--allow-empty" - because "--no-empty" does the opposite in split/split0 (it causes them to *remove* empty arguments)
[/Genius-Tech-1998s]

PR #8059, opened 2021-06-15
Support "$(cmd)" command substitution without line splitting

### LINKED ISSUE #159: Cannot save multi-line output in a variable (support $() command substitution syntax)

krobelus:
This adds support for
- `echo "$(cmd)"` - quoted command substitution without line splitting
- `echo $(cmd)` - same as `echo (cmd)`

Closes #159

## TODOs:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,7 @@ Deprecations and removed features
 
 Scripting improvements
 ----------------------
+- fish's command substitution syntax has been extended: ``$(cmd)`` now has the same meaning as ``(cmd)`` but it can be used inside double quotes, to prevent line splitting of the results. (:issue:`159`).
 - ``string collect`` supports a new ``--allow-empty`` option, which will output one empty argument in a command substitution that has no output (:issue:`8054`). This allows commands like ``test -n (echo -n | string collect --allow-empty)`` to work more reliably.
 
 Interactive improvements
--- a/doc_src/cmds/string-collect.rst
+++ b/doc_src/cmds/string-collect.rst
@@ -19,9 +19,11 @@ Description
 
 ``string collect`` collects its input into a single output argument, without splitting the output when used in a command substitution. This is useful when trying to collect multiline output from another command into a variable. Exit status: 0 if any output argument is non-empty, or 1 otherwise.
 
+A command like ``echo (cmd | string collect)`` is mostly equivalent to a quoted command substitution (``echo "$(cmd)"``). The main difference is that the former evaluates to zero or one elements whereas the quoted command substitution always evaluates to one element due to string interpolation.
+
 If invoked with multiple arguments instead of input, ``string collect`` preserves each argument separately, where the number of output arguments is equal to the number of arguments given to ``string collect``.
 
-Any trailing newlines on the input are trimmed, just as with ``"$(cmd)"`` substitution in sh. ``--no-trim-newlines`` can be used to disable this behavior, which may be useful when running a command such as ``set contents (cat filename | string collect -N)``.
+Any trailing newlines on the input are trimmed, just as with ``"$(cmd)"`` substitution. Use ``--no-trim-newlines`` to disable this behavior, which may be useful when running a command such as ``set contents (cat filename | string collect -N)``.
 
 With ``--allow-empty``, ``string collect`` always prints one (empty) argument. This can be used to prevent an argument from disappearing.
 
@@ -33,6 +35,11 @@ Examples
 .. BEGIN EXAMPLES
 
 ::
+
+    >_ echo "zero $(echo one\ntwo\nthree) four"
+    zero one
+    two
+    three four
 
     >_ echo \"(echo one\ntwo\nthree | string collect)\"
     "one
--- a/doc_src/language.rst
+++ b/doc_src/language.rst
@@ -569,6 +569,8 @@ The output of a command (or an entire :ref:`pipeline <pipes>`) can be used as th
 
 When you write a command in parenthesis like ``outercommand (innercommand)``, the ``innercommand`` will be executed first. Its output will be taken and each line given as a separate argument to ``outercommand``, which will then be executed. [#]_
 
+A command substitution can have a dollar sign before the opening parenthesis like ``outercommand $(innercommand)``. This variant is also allowed inside double quotes. When using double quotes, the command output is not split up by lines.
+
 If the output is piped to :ref:`string split or string split0 <cmd-string-split>` as the last step, those splits are used as they appear instead of splitting lines.
 
 The exit status of the last run command substitution is available in the :ref:`status <variables-status>` variable if the substitution happens in the context of a :ref:`set <cmd-set>` command (so ``if set -l (something)`` checks if ``something`` returned true).
@@ -588,7 +590,7 @@ Examples::
 
     # Set the ``data`` variable to the contents of 'data.txt'
     # without splitting it into a list.
-    begin; set -l IFS; set data (cat data.txt); end
+    set data "$(cat data.txt)"
 
     # Set ``$data`` to the contents of data, splitting on NUL-bytes.
     set data (cat data | string split0)
--- a/doc_src/language.rst
+++ b/doc_src/language.rst
@@ -573,7 +573,7 @@ If the output is piped to :ref:`string split or string split0 <cmd-string-split>
 
 The exit status of the last run command substitution is available in the :ref:`status <variables-status>` variable if the substitution happens in the context of a :ref:`set <cmd-set>` command (so ``if set -l (something)`` checks if ``something`` returned true).
 
-Only part of the output can be used, see :ref:`index range expansion <expand-index-range>` for details.
+To use only part of the output, refer to :ref:`index range expansion <expand-index-range>`.
 
 Fish has a default limit of 100 MiB on the data it will read in a command sustitution. If that limit is reached the command (all of it, not just the command substitution - the outer command won't be executed at all) fails and ``$status`` is set to 122. This is so command substitutions can't cause the system to go out of memory, because typically your operating system has a much lower limit, so reading more than that would be useless and harmful. This limit can be adjusted with the ``fish_read_limit`` variable (`0` meaning no limit). This limit also affects the :ref:`read <cmd-read>` command.
 
--- a/doc_src/tutorial.rst
+++ b/doc_src/tutorial.rst
@@ -400,11 +400,11 @@ Unlike other shells, fish does not split command substitutions on any whitespace
     -lgobject-2.0
     -lglib-2.0
 
-If you need a command substitutions output as one argument, without any splits, use ``string collect``::
+If you need a command substitutions output as one argument, without any splits, use quoted command substitution::
 
     > echo "first line
     second line" > myfile
-    > set myfile (cat myfile | string collect)
+    > set myfile "$(cat myfile)"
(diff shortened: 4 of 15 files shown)


zanchey:
If you can hold this for another few days, I think it would be worth waiting until after 3.3.0 to merge.

zx8:
Just want to drop a massive THANK YOU to @krobelus for getting the ball rolling on this – it's the one of the very few pieces of syntactic sugar (along with process substitution `<(...)` – the whole `psub` hack isn't great either) that I still sorely miss from my bash days... and I say that as someone who's been using fish for 6 years now.

I appreciate the implementation isn't perfect in its current form, but having compiled and briefly tested it locally, I'm a very happy end user and can't wait to see it merged! 🎉

The only thing I might suggest is that the following error message is improved slightly:

```
fish: $(...) is not supported. In fish, please use '(foo)'.
```

Perhaps something like this is more suitable now:

```
fish: $(...) without quotes is not supported. In fish, please use '(foo)'.
```

krobelus:
> `fish: $(...) without quotes is not supported. In fish, please use '(foo)'.`

Good idea, maybe even `fish: $(...) without quotes is not supported. In fish, please use '(foo)' or '"$(foo)"'`.

faho:
>Unquoted $() ist left unsupported for now. Maybe it's best to leave it at that, because even in Bash it should rarely be used.

In bash, it should rarely be used because it splits on spaces ($IFS, to be exact).

In fish, we could just split on newlines like we already do in `(foo)`. That makes command substitutions easy - if you want to split (you typically do), use `$(foo)` without quotes. If you do, use it with quotes. Exactly like variables.

krobelus:
Having read through the issues again, I think everyone agrees on adding `"$(cmd)"`.
It adds fairly intuitive syntactic sugar for `string collect`, especially for users of other shells.

If we make `$(cmd)` equivalent to `(cmd)` this means we probably want to phase out `(cmd)` eventually. I'm not sure if that's worth the trouble. If we could change it with no repercussions, then `$(cmd)` would probably be better. Not because this syntax is cleaner (I don't think so), but because of the mainstream appeal.

We could also look at how often the `"$(cmd)"` behavior is needed over `(cmd)`. If `"$(cmd)` is only used in rare cases, then the inconsistency might not matter that much?

> To be honest I find it weird to merge this without support outside of quotes, because it's weirdly in the middle between support and not.

I'm probably guilty of doublethink, but I think it's *sort of* consistent, because it's the obvious way to evaluate a command substitution inside double quotes. So I wouldn't mind adding it earlier.

I'm not against `$(cmd)` - I haven't fully formed my opinion - but adding it means we will have to support it alongside `(cmd)` for many years, so it should be a conscious decision.

---

Another random idea: disable the line splitting in `$(cmd)`. This way `(cmd)` will remain. This is probably a bad idea because the behavior is fairly subtle.

faho:
>I'm probably guilty of doublethink, but I think it's sort of consistent, because it's the obvious way to evaluate a command substitution inside double quotes.

That's just the thing - if it's the obvious way to do a command substitution inside double quotes, then it's also the obvious way to do it *outside*.

Or, to come at it from another angle: This is an FAQ (not formally - there's a question about backticks instead - but given what we see. There is also an entry in fish-for-bash-users). We have users asking for this *all the time*. We have the opportunity to entirely remove that.

I don't see the benefit in removing `()`, but I also don't see any great harm in having both.

kopischke:
I’m with @faho: once `$()` is available in quotes, people will start expecting it to work unquoted too. I agree there is no harm in keeping the legacy `()` form besides unquoted `$()`; there are other cases where fish supports such variants, like `and / or` and `&& / ||`.

krobelus:
The current error message on `$()` should already answer user questions.
The advantages of `$()` that I see are improved consistency and that it's easier to paste (sloppy) bash commands.
Okay I guess it's fine to have both.

>  people will start expecting it to work too

I wonder if that will happen for `var=val cmd` vs. `var=val`

ridiculousfish:
Sorry for delay in getting to this - I've also been working on this but it looks like yours is a lot further along! Agreed with supporting unquoted `$()` I'll certainly review, very exciting!

marcusatiliusregulus:
i like the idea of supporting a way of quoted command substitution because it removes worries about a substitution returning a list or nothing and "corrupting" the string you're building.

however i **heavily disagree** on supporting unquoted `$()` as an alternative to `()`. fish is not bash, it doesn't even try to be fully posix compliant, so doing this just because new users have to relearn some concept is just wrong. on the contrary: i would _love_ to have this error message removed allowing for variable expansion like:

```fish
$ set 1 one; set 2 two; set 3 three
$ echo $(seq 3)
one two three
$ echo $(echo 2)
two
```

if this is not supported and constructs with the new quoted substitution like

```fish
$ set foobar foo bar
$ echo $"foo$(echo -n bar)"
foo bar
```

dont work either, i fear we close the door to generative variable expansion forever and would miss out on a powerful feature for all the wrong reasons.

ridiculousfish:
> i fear we close the door to generative variable expansion forever

If I understand correctly, by "generative variable expansion" you mean for a command substitution to output the name of a variable, and then expand that, as a shorthand:

     echo $(echo foo) ; # equivalent to set tmp (echo foo); echo $tmp

 Can you say more about why that's a powerful feature? Do other shells support this?

krobelus:
That sounds like a nice feature for power users but I think it's rarely used in practice. Moreover, most users will probably expect the Bash behavior, so anything but that or an error message will catch them offguard.

ridiculousfish:
@krobelus I stand in awe. My approach was to teach fish about cmdsubs in quotes everywhere, which meant a lot of complexity and recursion. Your trick of making cmdsubs "temporarily close" quotes is so brilliantly simple, I would not have believed it possible! I learned from it today.

There's a bug with escaping the output in `expand_cmdsubst` if the output itself contains a quote:

    > echo "$(echo '"')" # should output one double-quote

because the double-quote is not escaped. Backslash has the same problem. You may [steal my function here](https://github.com/ridiculousfish/fish-shell/blob/53da49af6d0c227325adf559e1ba2c985a2a7494/src/common.cpp#L852) which escapes a string so it may be inserted unchanged in double quotes.

A second point is brackets:

    > echo (seq 3)[2] # outputs 2
    > echo "$(seq 3)[2]" # outputs 1\n2\n3[2]

do we want to support these? I go back and forth on this: the argument for support is consistency with variable expansion, but it feels like it would be a rarely used feature that users might trip over. Either way we should be deliberate about it and mention it in the docs.

marcusatiliusregulus:
rediculousfish yes, i think you understand correctly.

> Do other shells support this?

im not sure

> Can you say more about why that's a powerful feature?

its about composability. shell scripting is awesome because you just make up whatever you need by connecting and substituting whats needed. but this is severely hindered when its painful to get the value of a variable you know the name of but cant access by simply throwing a dollar at it.

fish doesn't have dictionaries and can't store lists in lists. allowing explicit variable expansion would make getting around that way more easy, performant, intuitive and powerful. having these data structures enables a whole new level of expressivness in your code.

faho:
>allowing explicit variable expansion would make getting around that way more easy, performant, intuitive and powerful.

From what little explanation you've given here, it would *not* be more powerful, as it's simply a shortcut. Would it be any easier, more performant or more intuitive? I'm not sold.

Anyway, if you really wanted this supported, the way to do it would be to write up a proposal and ideally submit a patch, not allude to it in an *entirely different* PR.

And if you want `$()` as the syntax: Sorry, I don't think so. We had discussed `$()` for command substitutions before (#159), and it has a *lot* of alluring properties - one that it's simply `()`, but with a `$` in front (so it already looks sorta like a command substitution in fish), one that it's literally the syntax taken from elsewhere (so that it looks like a command substitution if you come from elsewhere), one that it uses a `$` which is *already special* in double-quotes so it's entirely backwards-compatible.

Pick something else, and explain where the power comes from!

marcusatiliusregulus:
ok, let me try to explain.  why is it more powerful (imo):

variable assignment and -lookup are quicker than manipulating lists
of key-value pairs and other dict tricks. they are intuitive because
well known (appending, prepending, indexing etc), powerful because
they simply support multiline- and multiple elements without any
escaping-unescaping magic and simple as they come for free because
`set` is already there.

is it simply a shortcut? argueably. but considere this: if i cannot
access the variable by expanding it, it is impossible to reliably
get the value from inside the current command, i have to create a
new variable outside of the current command and expand this one.
why? because i cant return from the substitution (or any function
really) other than by writing to stdout and this mangles the
items. so for every time i want to read a value i have to create
a new var before the current command. and all this time while i
already have the processed value in its own var whose name i know.

kopischke:
> A second point is brackets:
>
> ```fish 
> echo (seq 3)[2] # outputs 2
> echo "$(seq 3)[2]" # outputs 1\n2\n3[2]
> ```
> 
> do we want to support these? I go back and forth on this: the argument for support is consistency with variable expansion, but it feels like it would be a rarely used feature that users might trip over. Either way we should be deliberate about it and mention it in the docs.

May I point out that it's not just consistency with variable expansion, but also consistency with unquoted command substitutions? If fish supports unquoted `$()`, making it an equivalent to `()` – and I am in favour of that –, I think bracket expansion should be supported, both in quotes and unquoted. If it were not, this would end up creating a second, subtly different set of command expansion operators instead of a more powerful substitute.

I always find it helpful to imagine how I’d explain the feature to a fish newbie: of “command substitution is `$()`; for legacy reasons, you can also use `()` unquoted“ and “command substitution can be `$()` in or outside quotes, or `()` outside them, with the latter supporting bracket expansion”, I know which I find easier.

krobelus:
> My approach was to teach fish about cmdsubs in quotes everywhere, which meant a lot of complexity and recursion.

Yeah, it's simpler to just reuse the existing command-substitution parsing, since the behavior is almost the same.
It's a bit of a hack, but I haven't found any serious problems.

Fixed the bug with `"` and added index range expansion since that seems more consistent, though I'm not also not sure which one is better.
I also fixed a bug in `parse_util_locate_brackets_of_type`, and added some comments - I wrote most of this a while ago and it got a bit hard to understand.

ridiculousfish (approved):
This LGTM, it's really great!

lilyball:
@marcusatiliusregulus If I understand what you want correctly, surely you can define a function using `--no-scope-shadowing` that takes the name of a variable and returns the value. Maybe call it `get` (as a dual to `set`). This way your proposed `echo $(echo 2)` would be written like `get (echo 2)`.

faho:
If we wanted this the most natural syntax would be `$$(echo foo)` - take the command substitution and expand it, now you have `$foo`, expand that. Two expansions, two dollars, done.

As it stands, I'd recommend spelling it out with a temp variable over having to define your own function with "--no-scope-shadowing":

```fish
set -l tmp (echo foo)
echo $$tmp
```

In my experience it's rarely needed - not never, but also not particularly often. So I'm not sold that we *need* a shortcut for it, but if we did `$$(thing)` seems like the obvious winner.

marcusatiliusregulus:
yeah, looks good to me, ill use that. thanks

Issue #7317, opened 2020-09-07
Remove default universal variables


faho:
Currently, we set a few universal variables on first run:

- The colors - $fish_color_* and $fish_pager_color_*
- $fish_key_bindings
- $__fish_initialized

$__fish_initialized is used to guard

- Setting the colors (as an optimization)
- Upgrading the abbrs from fish 2.3.0

The optimization, on my system, is worth a whopping 0.25 *milli*seconds, to be paid on each startup of an interactive shell, and in turn updating the defaults becomes harder as we can't tell a user's customization from our old value anymore.

I am not sure updating the abbrs still works (it's two upgrade steps! Once to 2.3.0 and once to 3.0), and I don't have a system to test. I would actually assume most people use `abbr` calls in config.fish anyway, and even if they don't calling `abbr` again isn't too odious.

If these two are gone, $__fish_initialized can just go away entirely.

$fish_key_bindings is possibly more acceptable as a universal variable, given that it is a user choice that is quite unlikely to change between shell instances, but given that this is the last default universal variable I believe that behavior to be too surprising, especially since it necessitates actively switching back if the user decides they don't like vi-bindings.

Going further, it would also be possible to always default to global variables even e.g. for `fish_add_path` and `abbr`.

(Personally, I'd be open to removing uvars entirely, but I suspect that's not a popular position)

IlanCosman:
Removing uvars entirely is an interesting idea. That would fundamentally push fish away from the generally CLI-based configuration it has to a more traditional file-based one. There are quite a few advantages to doing it the file-based way. For example, new users often struggle with the fish paradigm when coming from a traditional shell like bash or zsh. It would also be easier to share configs like people do their .zshrc etc. I'm personally open the idea, but it would represent a monumental shift for fish.

zanchey:
I'm not clear what the benefit is, though?

Dropping the colour optimisation seems fine as a first action.

faho:
>I'm not clear what the benefit is, though?

The benefit of removing default uvars?

- Less stuff the user doesn't know or care about in the uvar file
- Easier updating of color defaults - if they're uvars, you only ever get the defaults the first time. If they're global, when we change the defaults they apply on next start (but since we would only set the variable if it isn't set, all you need to do to opt out is to set the variable explicitly)
- Once we stop setting uvars by default it becomes possible to just ignore uvars entirely

[...]

PR #8072, opened 2021-06-19
Stop setting any uvars in fish initialization

ridiculousfish:
This removes all of the universal variables from fish's initialization scripts. With this change, fish will not even create its `fish_variables` file until the user sets a universal variable. See #7317.

Variables which used to be universal (the color variables like `fish_color_cwd`) are now global, instead of universal. As before, they are only set if not already set.

There are two risks here:

1. We no longer migrate the pre-3.0 abbreviation format. It seems like an OK time to stop doing that.
2. Because color variables are now global instead of universal, a command like `set -U fish_color_normal purple` won't appear to work until the user spawns a new shell. This is because global scope trumps universal scope. I think we should fix that by just removing universal variables, but I am saving that for a next pull request.

I think the second risk is serious enough that I will not merge this until we have a solution. My proposal is to [remove universal variables](https://github.com/fish-shell/fish-shell/issues/7317#issuecomment-701165897), and I'm pretty far along on that, but I think this change can be reviewed separately.


krobelus:
Looks good; I agree that this should be merged when we remove the uvar shadowing problem.

zanchey:
If you can hold this for another few days, I think it would be worth waiting until after 3.3.0 to merge.

ridiculousfish:
I'm going to close this for now, my work on replacing uvars isn't going particularly well.

PR #8081, opened 2021-06-22
__fish_list_current_token: Do not use eval

wylfen:
Similarly to b0e3cc4b5 (__fish_complete_suffix: Remove `eval`, 2019-12-28), this use of eval is unsafe and can spew errors if invoked on an incomplete brace expansion.

## Examples
```
$ (exit)
```
With the cursor on this token, `Alt-L` will execute `exit` - the shell exits after the next command.

```
$ this-dir-does-not-exist\n-o\ntrue
```
With the cursor on this token, `Alt-L` will attempt listing the contents of both `this-dir-does-not-exit` and `true` whilst also passing `-o` to the `ls` command.

## TODOs:

I could not find a good way of testing this, since the function relies on `commandline -t`. I'll happily add a test if this is easily possible somehow.

### DIFF
--- a/share/functions/__fish_list_current_token.fish
+++ b/share/functions/__fish_list_current_token.fish
@@ -2,7 +2,7 @@
 # of the directory under the cursor.
 
 function __fish_list_current_token -d "List contents of token under the cursor if it is a directory, otherwise list the contents of the current directory"
-    set -l val (eval echo (commandline -t))
+    set -l val (commandline -t)
     printf "\n"
     if test -d $val
         ls $val


faho:
The issue here is that it now won't list e.g. `$HOME/` or `~/`.

Really we need an expand builtin, see #5811.

When that's done, we can use it here instead.

Anyway, it's probably worth it, so I'm merging.

> I could not find a good way of testing this, since the function relies on commandline -t. I'll happily add a test if this is easily possible somehow.

Theoretically you can use pexpect tests, but I don't think it's really worth it - they're slow, and this is a low-risk change.

PR #8089, opened 2021-06-29
completions/git: suggest modified-staged-deleted files

z3ntu:
## Description

Add those files to the completion of 'add', 'checkout', 'diff', 'difftool', 'restore' and 'stash push'.

Test case:
```
mkdir /tmp/test
cd /tmp/test
git init
touch foo
git add foo
git commit -m "foo"
echo bar > foo
git add foo
rm foo
git add <TAB> *COMPLETION!*
git restore <TAB> *COMPLETION!*
# etc
```

I think I've added it to all relevant options now, I've oriented myself at "deleted" as it's quite similar in handling.

## TODOs:


krobelus:
Good catch, thanks!

PR #8097, opened 2021-07-01
`FISH_USE_POSIX_SPAWN` and `HAVE_SPAWN_H` fixes

evanmiller:
## Description

`FISH_USE_POSIX_SPAWN` is always defined, thanks to the lines

https://github.com/fish-shell/fish-shell/blob/9f54c8d88b4e02f8b6b3f74c18dd57060ef035c4/src/postfork.h#L15-L17

So replace `#ifdef` with `#if` to fix compilation on platforms lacking `spawn.h`. Also make the `spawn.h` inclusion condition consistent across files.

## TODOs:

### DIFF
--- a/src/postfork.cpp
+++ b/src/postfork.cpp
@@ -10,7 +10,7 @@
 
 #include <cstring>
 #include <memory>
-#ifdef FISH_USE_POSIX_SPAWN
+#ifdef HAVE_SPAWN_H
 #include <spawn.h>
 #endif
 #include <cwchar>
--- a/src/postfork.h
+++ b/src/postfork.h
@@ -9,7 +9,7 @@
 #include <unistd.h>
 
 #include "maybe.h"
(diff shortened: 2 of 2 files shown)


faho:
>So replace #ifdef with #if to fix compilation on platforms lacking spawn.h

Which platforms are those?

evanmiller:
@faho I am currently attempting to compile on OSX 10.4 / PowerPC, via MacPorts. My understanding from the commit messages is that some versions of Android also lack spawn.h.

faho:
> My understanding from the commit messages is that some versions of Android also lack spawn.h.

It works on termux [with `-landroid-spawn`](https://github.com/termux/termux-packages/blob/master/packages/fish/CMakeLists.txt.patch), we don't care about Android otherwise.

But anyway, this seems correct and passes CI elsewhere, so I'm merging.

PR #8104, opened 2021-07-02
 docs: Made the abort/edit history feature more discoverable.

ewtoombs:
First, I changed "the escape key" to :kbd:`Esc`. This makes this information
easier to find when scanning the docs because it stands out and because it is
more consistent with the docs's formatting of keyboard keys. Then I added "or
edit" after abort, because the escape key is not just for aborting a search.
Finally, I added a link from the FAQ to history-search to make this mechanism
easier to discover.

This was all to address confusion in former zsh and bash users as to how to
edit a search that is in progress, but this will also help new users. See https://github.com/fish-shell/fish-shell/pull/6686#issuecomment-872960760

### DIFF
--- a/doc_src/faq.rst
+++ b/doc_src/faq.rst
@@ -145,7 +145,7 @@ Note that fish has a default titlebar message, which will be used if the fish_ti
 
 How do I run a command from history?
 ------------------------------------
-Type some part of the command, and then hit the :kbd:`↑` (up) or :kbd:`↓` (down) arrow keys to navigate through history matches. Additional default key bindings include :kbd:`Control`\ +\ :kbd:`P` (up) and :kbd:`Control`\ +\ :kbd:`N` (down).
+Type some part of the command, and then hit the :kbd:`↑` (up) or :kbd:`↓` (down) arrow keys to navigate through history matches. Additional default key bindings include :kbd:`Control`\ +\ :kbd:`P` (up) and :kbd:`Control`\ +\ :kbd:`N` (down). See :ref:`Searchable command history <history-search>` for more information.
 
 Why doesn't history substitution ("!$" etc.) work?
 --------------------------------------------------
--- a/doc_src/interactive.rst
+++ b/doc_src/interactive.rst
@@ -523,7 +523,7 @@ After a command has been executed, it is remembered in the history list. Any dup
 
 By pressing :kbd:`Alt`\ +\ :kbd:`↑` and :kbd:`Alt`\ +\ :kbd:`↓`, a history search is also performed, but instead of searching for a complete commandline, each commandline is broken into separate elements just like it would be before execution, and the history is searched for an element matching that under the cursor.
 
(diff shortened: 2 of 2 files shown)


doc_src/faq.rst
```diff
@@ -145,7 +145,7 @@ Note that fish has a default titlebar message, which will be used if the fish_ti
 
 How do I run a command from history?
 ------------------------------------
-Type some part of the command, and then hit the :kbd:`↑` (up) or :kbd:`↓` (down) arrow keys to navigate through history matches. Additional default key bindings include :kbd:`Control`\ +\ :kbd:`P` (up) and :kbd:`Control`\ +\ :kbd:`N` (down).
+Type some part of the command, and then hit the :kbd:`↑` (up) or :kbd:`↓` (down) arrow keys to navigate through history matches. Additional default key bindings include :kbd:`Control`\ +\ :kbd:`P` (up) and :kbd:`Control`\ +\ :kbd:`N` (down). See :ref:`history <history-search>` for more details about searching history interactively with fish.
```
faho, line 148: This wording is a bit redundant. Maybe just ``See :ref:`Searchable command history <history-search>` for more information``?

ewtoombs: I'm OK with this.

doc_src/interactive.rst
```diff
@@ -523,7 +523,7 @@ After a command has been executed, it is remembered in the history list. Any dup
 
 By pressing :kbd:`Alt`\ +\ :kbd:`↑` and :kbd:`Alt`\ +\ :kbd:`↓`, a history search is also performed, but instead of searching for a complete commandline, each commandline is broken into separate elements just like it would be before execution, and the history is searched for an element matching that under the cursor.
 
-History searches are case-insensitive unless the search string contains an uppercase character, and they can be aborted by pressing the escape key.
+History searches are case-insensitive unless the search string contains an uppercase character, and they can be aborted or edited by pressing :kbd:`Esc`.
```
krobelus, line 526: (Any movement key (like left-arrow) also stops history search. That should be fairly obvious from trial and error.)

I don't think "edited" is accurate here because you can only start a new search.

krobelus: ah, you're right, my bad! Page-Down and Control-Z also do that.

ewtoombs: Really? I don't think pgdn or ^Z are documented. Should they be?

krobelus: I have added my suggestion (because you don't edit the search but the *search string*).

I added page-down, but not the various keys for undo (ctrl-Z, ctrl-_, ctrl-/).
I guess undo is a general feature, and it even has a glitch here:
press `echo` <kbd>Up</kbd> ^Z <kbd>Up</kbd> ^Z
The first undo goes back to the original string but the second clears the command line.

faho:
Yeah, that looks alright, merged! Thanks!

ewtoombs:
No problem, @faho ! This should take care of most of the search-related gripes coming from former zsh and bash users.

PR #8108, opened 2021-07-03
Add missing options for `bind`

kidonng:
## Description

Add `-s/--silent` `--preset` `--user` completions for `bind` which are presented in man pages.

> bind [(-M | --mode) MODE] [(-m | --sets-mode) NEW_MODE] [--preset | --user] [(-s | --silent)] [(-k | --key)] SEQUENCE COMMAND [COMMAND...]

> • --preset  and  --user specify if bind should operate on user or preset bind‐
         ings. User bindings take precedence over preset bindings when fish looks  up
         mappings.  By  default, all bind invocations work on the "user" level except
         for listing, which will show both levels. All invocations except for insert‐
         ing new bindings can operate on both levels at the same time (if both --pre‐
         set and --user are given). --preset should only be used in full binding sets
         (like when working on fish_vi_key_bindings).

## TODOs:


faho:
Yup, thanks!

PR #8113, opened 2021-07-05
Add completion for Rakudo

CIAvash:
[Rakudo](https://rakudo.org/) is an implementation of the [Raku](https://raku.org/) programming language.

I set the `$rakudo` command to the filename, so that users can easily create a `raku.fish` symlink to `rakudo.fish`, if they want Rakudo completions for the `raku`(which is a symlink to an implementation) command.


faho:
>I set the $rakudo command to the filename

If they want that, they can just use

```fish
complete -c raku -w rakudo
```

No need to add special support.

share/completions/rakudo.fish
```diff
@@ -0,0 +1,24 @@
+complete -c rakudo -n __fish_is_first_token -k -f -a '(__fish_complete_suffix)'
```
faho, line 1: This disables file completions via `-f`, only to reenable them via `__fish_complete_suffix`.

Also, `-k` (`--keep-order`) should not be used without a good reason. Please remove it!

CIAvash: I think that's because I looked at [python.fish](https://github.com/fish-shell/fish-shell/blob/727934c6b6deb7a375531f3886e535eb4fc36d2a/share/completions/python.fish#L19).

krobelus: I guess that was more relevant when `__fish_complete_suffix .py` **only** showed `*.py` files.
Now it just sorts them first - this is why `-k` is needed. I guess it's still okay to keep

zanchey:
I think the first two lines can just be `complete -c rakudo -n __fish_is_first_token -F -a - -d 'Read program source from STDIN or start REPL if a TTY'`?

share/completions/rakudo.fish
```diff
@@ -0,0 +1,24 @@
...
+complete -c rakudo -l optimize -x -d 'use the given level of optimization (0..3)' -a '0 1 2 3'
+complete -c rakudo -s o -l output -r -d 'Specify name of output file'
+complete -c rakudo -s v -x -l version -d 'Display version information'
+complete -c rakudo -s V -x -d 'Print configuration summary'
+complete -c rakudo -l rakudo-home -r -d 'Override the path of the Rakudo runtime files'
+complete -c rakudo -l stagestats -d 'Display time spent in the compilation stages'
+complete -c rakudo -l repl-mode -x -d 'Specify the REPL mode' -a 'interactive\tStarts\ Interactive\ REPL\ \(default\) non-interactive\tSTDIN\ is\ read\ entirely\ and\ evaluated\ as\ if\ it\ were\ a\ program'
+complete -c rakudo -l profile -r -d 'Write profile information to a file. Extension controls format'
+complete -c rakudo -l profile-compile -r -d 'Write compile-time profile information to a file. Extension controls format'
+complete -c rakudo -l profile-kind -x -d 'Choose the type of profile to generate' -a 'instrumented\tPerformance\ measurements heap\tRecord\ heap\ snapshots\ after\ every\ garbage\ collector\ run'
```
krobelus, line 21: we can use quotes instead of backslashes, sam ein line 18

```suggestion
complete -c rakudo -l profile-kind -x -d 'Choose the type of profile to generate' -a '
    instrumented\t"Performance measurements"
    heap\t"Record heap snapshots at every garbage collection run"
'
```

share/completions/rakudo.fish
```diff
@@ -0,0 +1,24 @@
+complete -c rakudo -n __fish_is_first_token -a '(__fish_complete_suffix)'
+complete -c rakudo -n __fish_is_first_token -a - -f -d 'Read program source from STDIN or start REPL if a TTY'
+complete -c rakudo -s c -d 'Check syntax only (runs BEGIN and CHECK blocks)'
+complete -c rakudo -l doc -d 'Extract documentation and print it as text'
+complete -c rakudo -s e -x -d 'One line of program, strict is enabled by default'
+complete -c rakudo -s h -l help -x -d 'Display help text'
+complete -c rakudo -s n -d 'Run program once for each line of input'
+complete -c rakudo -s p -d 'Same as -n, but also print $_ at the end of lines'
+complete -c rakudo -s I -r -d 'Add the path to the module search path'
+complete -c rakudo -s M -r -d 'Load the module prior to running the program'
```
krobelus, line 10: the `-c` in `complete -c` is not necessary anymore, I guess we can start by not adding it to new completions?

faho: So far I've not required that, because it means that these completions become incompatible with earlier versions. Which isn't a big problem, but when it's just gratuitous stuff like this it feels weird?

This was changed in fish 3.2, which was released in *march*. I think we can wait a bit longer before we break completions for it *on purpose*.

krobelus: Oh yeah, and we had an issue of a user upgrading fish, and then old shell instances would get this error.
So we should wait at least until we can reasonably expect most users to have upgraded.

share/completions/rakudo.fish
CIAvash: Thought I tried everything before using backslashes, obviously not.

krobelus:
One more quick thing: we usually try to avoid parentheses in completion descriptions, because fish already adds them.
Here are some suggestions, but the current version is fine of course.

```diff
diff --git a/share/completions/rakudo.fish b/share/completions/rakudo.fish
index 06804354c..491fe19e8 100644
--- a/share/completions/rakudo.fish
+++ b/share/completions/rakudo.fish
@@ -2 +2 @@ complete -c rakudo -n __fish_is_first_token -F -a - -d 'Read program source from
-complete -c rakudo -s c -d 'Check syntax only (runs BEGIN and CHECK blocks)'
+complete -c rakudo -s c -d 'Check syntax only. Runs BEGIN and CHECK blocks'
@@ -11 +11 @@ complete -c rakudo -l target -x -k -a 'start parse syntaxcheck ast optimize mast
-complete -c rakudo -l optimize -x -d 'use the given level of optimization (0..3)' -a '0 1 2 3'
+complete -c rakudo -l optimize -x -d 'use the given level of optimization, pass 0-3' -a '0 1 2 3'
@@ -17 +17 @@ complete -c rakudo -l stagestats -d 'Display time spent in the compilation stage
-complete -c rakudo -l repl-mode -x -d 'Specify the REPL mode' -a 'interactive\t"Starts Interactive REPL (default)" non-interactive\t"STDIN is read entirely and evaluated as if it were a program"'
+complete -c rakudo -l repl-mode -x -d 'Specify the REPL mode' -a 'interactive\t"Start the default interactive REPL" non-interactive\t"STDIN is read entirely and evaluated as if it were a program"'
@@ -23 +23 @@ complete -c rakudo -l debug-port -x -d 'Listen for incoming debugger connections
-complete -c rakudo -l tracing -f -d 'Output a line to stderr on every interpreter instr (only if enabled in MoarVM)'
+complete -c rakudo -l tracing -f -d 'Output a line to stderr on every interpreter instr, if enabled in MoarVM'
```

krobelus:
Thanks!

Issue #7379, opened 2020-10-05
Remove universal variables


IlanCosman:
There was some discussion of removing the universal variable functionality in #7317. Some good initial ideas were brought up there.

folke:
Loading universal variables from `fish_variables` is currently much faster than specifying them as globals in your `config.fish`. See also my proposal https://github.com/fish-shell/fish-shell/issues/7432

krobelus:
A lightweight first iteration of this could look like:

1. Whenever a universal variable is set, also set it to the global scope
2. Whenever a global variable is set, if it's also defined in the universal scope, then also update it in the universal scope (persisting it)

This removes support for shadowing of universals with globals, which is probably not used very much<sup>[citation needed]</sup>.

The next obvious step would be to remove the universal scope, since at this point it's just used as flag whether the variable is persisted.

In the end we probably want a [different interface for universal variables](https://github.com/fish-shell/fish-shell/issues/7317#issuecomment-701165897)

planet36:
Is the universal variable feature going to be removed from fish?

faho:
@planet36 That is why the issue is called "Remove universal variables", yes. It means that we would like to remove universal variables, in their current form.

They would be replaced with something that doesn't have their drawbacks, see https://github.com/fish-shell/fish-shell/issues/7317#issuecomment-701165897.

planet36:
Here's a problem I had that I solved using universal variables.  (If there's a better way, please let me know.)

In my config.fish I set environment variable `CXXFLAGS` in a block like this.

```
if status is-login
    set --export CXXFLAGS -O2 -pipe -Wall -Wextra
end
```

The intent is to set my environment variables once upon login, not every invocation of fish.

Using `$CXXFLAGS` works as expected at first.

`count $CXXFLAGS`
> 4

`g++ $CXXFLAGS test.cpp`
_success_

But when I open a nested shell, the variable expansion (or maybe it's the splitting) is different.

_open a nested shell_
`fish`

`count $CXXFLAGS`
> 1

`g++ $CXXFLAGS test.cpp`
> cc1plus: error: argument to '-O' should be non-negative integer, 'g', 's' or 'fast'

I understand the following work-around, but I'd rather not type it out often.

`count (string split -- ' ' $CXXFLAGS)`
> 4

In bash, the environment variable is expanded/split as expected.

_open a nested shell_
`bash`

`g++ $CXXFLAGS test.cpp`
_success_

[...]

PR #8117, opened 2021-07-05
Deduplicate $fish_user_paths automatically

faho:
In the variable handler, we just go through the entire thing and keep
every element once.

If there's a duplicate, we set it again, which calls the handler
again.

This takes a bit of time, to be paid on each startup. On my system,
with 100 already deduplicated elements, that's about 4ms (compared to
~17ms for adding them to $PATH).

It's also semantically more complicated - now this variable
specifically is deduplicated? Do we just want "unique" variables that
can't have duplicates?

However: This entirely removes the pathological case of appending to
$fish_user_paths in config.fish (which should be an FAQ entry!), and the implementation is quite simple.

## TODOs:

### DIFF
--- a/share/config.fish
+++ b/share/config.fish
@@ -98,6 +98,21 @@ end
 
 # Add a handler for when fish_user_path changes, so we can apply the same changes to PATH
 function __fish_reconstruct_path -d "Update PATH when fish_user_paths changes" --on-variable fish_user_paths
+    # Deduplicate $fish_user_paths
+    # This should help with people appending to it in config.fish
+    set -l new_user_path
+    for path in (string split : -- $fish_user_paths)
+        if not contains -- $path $new_user_path
+            set -a new_user_path $path
+        end
+    end
+
+    if test (count $new_user_path) -lt (count $fish_user_paths)
+        # This will end up calling us again, so we return
+        set fish_user_paths $new_user_path
+        return
+    end
+
     set -l local_path $PATH
 
     for x in $__fish_added_user_paths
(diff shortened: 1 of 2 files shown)


krobelus:
Looks good, this will get rid of the scenario where users add to a universal `$fish_user_paths` on each startup.
An alternative would be to warn when duplicates were added during startup, but that seems harder to get right.

faho:
> An alternative would be to warn when duplicates were added during startup, but that seems harder to get right.

Yeah, we *could* warn whenever a duplicate is added, but that's probably too annoying?

Or do naughty things with checking the stacktrace and only warn if it's still in config.fish?

I don't really want this to be something you count on, I'd just like to fix it for the simple case. So I wouldn't even document this, in case we ever decide to remove or replace it.

IlanCosman:
Doesn't this sort of overlap with `fish_add_path`? As you said, this problem is worthy of FAQ status, and the answer would be `fish_add_path`. As long as we market `fish_add_path` in all of the relevant places in the docs, this question should basically go away.

faho:
>Doesn't this sort of overlap with fish_add_path?

Well, yes, obviously. But the point is that, instead of constantly telling people not to do the wrong thing, we can just make the wrong thing not broken.

If you understand universal variables it's obvious that appending to one in config.fish is wrong, but enough people don't understand universal variables (which is also a reason for #7379).

I'm tired of diagnosing "my shell has gotten slow" problems that half the time come down to this (and most of the other half is some plugin that we have no control over).

ridiculousfish:
This looks good to me too. I think this simple targeted fix is right.

faho:
Okay, I added some tests, merging.

PR #8121, opened 2021-07-06
Enable OSC 0 when running in WezTerm

JeanMertz:
WezTerm [has support for OSC 0](https://wezfurlong.org/wezterm/escape-sequences.html#operating-system-command-sequences) (dynamic titles) but somehow this wasn't being picked up by the checks currently in place. To solve this, I added the `TERM` value as a hard-coded accepted value for OSC 0 support.


faho:
Merged, thanks!

PR #8127, opened 2021-07-09
fish_config: Read colorschemes from .theme files

faho:
## Description

As we know, fish_config currently is the only place you can see the colorschemes. There's no way inside the shell to access them, which is profoundly weird.

So, what we do is store the themes in ".theme" files instead. These are a fishscript-influenced key-value format, basically what `set -L` prints. Essentially, every line is either a comment (including a metadata comment) starting with "#", empty, or the arguments for a `set` call from the variable name onwards:

```fish
# name: ayu Light
# url: 'https://github.com/dempfi/ayu'
# preferred_background: FAFAFA

fish_color_normal 575F66
fish_color_command 55B4D4
fish_color_quote 86B300
fish_color_redirection A37ACC
fish_color_end ED9366
fish_color_error F51818
fish_color_param 575F66
fish_color_comment ABB0B6
fish_color_match F07171
fish_color_selection --background=FF9940
fish_color_search_match --background=FF9940
fish_color_history_current --bold
fish_color_operator FF9940
fish_color_escape 4CBF99
fish_color_cwd 399EE6
fish_color_cwd_root red
fish_color_valid_path --underline
fish_color_autosuggestion 8A9199
fish_color_user brgreen
fish_color_host normal
fish_color_cancel -r
fish_pager_color_description B3A06D yellow
fish_pager_color_completion normal
fish_pager_color_prefix normal --bold --underline
fish_pager_color_progress brwhite --background=cyan
```

This makes it easy to read from both webconfig.py, and a fish script (to be written and integrated into `fish_config` as `fish_config theme`).

Little should currently change about the fish_config browser interface, except for that the themes have been reordered. They're now in alphabetical, case-insensitive order except for the current and default themes, which are first.

The motivation for all of this is that it allows us to:

- Add a terminal UI for picking themes
- Add a way to add themes for third-parties (this is safe because there is no *code* here, unlike prompts)
- Make it easier to add themes to fish proper (by not requiring weird javascript - we could even make a theme dumper!)

Part of #3625.

## TODOs:


faho:
Note: I extracted the themes by setting them from webconfig (I hacked it so it only printed the set commands instead of executing them). This means they've lost the relation to the palette, and now it's all just a list of variables.

I don't see how we can keep a non-code format without losing that, and that's imperative to make it usable from python and fish, and to allow others to add things (imagine a ~/.config/fish/themes directory that fish_config just picks up! imagine a theme of the month on fishshell.com!)

cmake/Install.cmake
```diff
@@ -147,6 +149,7 @@ install(DIRECTORY share/tools/web_config
         PATTERN "*.html"
         PATTERN "*.py"
         PATTERN "*.js"
+        PATTERN "*.theme"
```
zanchey, line 152: To answer the question in your commit message, yes, you need this - the `FILES_MATCHING` argument to `install(DIRECTORY)` disables the copying of any non-matching files (see https://cmake.org/cmake/help/latest/command/install.html)

share/tools/web_config/webconfig.py
```diff
@@ -880,7 +880,8 @@ class FishConfigHTTPRequestHandler(SimpleHTTPServer.SimpleHTTPRequestHandler):
     def write_to_wfile(self, txt):
         self.wfile.write(txt.encode("utf-8"))
 
-    def do_get_colors(self):
+    def do_get_colors(self, path=None):
+        """ Read the colors from a .theme file or the current shell config (if no path has been given) """
```
zanchey, line 884: Can I suggest "Read the colors from the .theme file in path, or the current shell if no path provided"?

share/tools/web_config/webconfig.py
```diff
@@ -1299,7 +1310,23 @@ def do_GET(self):
...
+            # The rest in alphabetical order.
+            curcolors, curinfo = self.do_get_colors()
+            defcolors, definfo = self.do_get_colors("themes/fish default.theme")
+            curinfo.update({ "theme": "Current", "colors": curcolors})
+            definfo.update({ "theme": "fish default", "colors": defcolors})
+            output = [curinfo, definfo]
+            paths = sorted(glob.iglob("themes/*.theme"), key=str.casefold)
+            for p in paths:
+                # Strip ".theme" suffix and path
+                theme = os.path.basename(p)[:-6]
```
zanchey, line 1324: Possibly `os.path.splitext` is more obvious here?

zanchey (commented):
This works well in my testing, and it's a great idea.

I'm not able to give much comment on the JS.

Can I suggest running `black` over the Python source, either before and after your changes or just after?

share/tools/web_config/webconfig.py
```diff
@@ -1288,7 +1294,19 @@ def do_GET(self):
...
+            output = [{ "theme": "Current", "colors": self.do_get_colors()},
+                      { "theme": "fish default", "colors": self.do_get_colors("themes/fish default.theme")}]
+            paths = sorted(glob.iglob("themes/*.theme"), key=str.casefold)
+            for p in paths:
+                # Strip ".theme" suffix and path
+                theme = os.path.basename(p)[:-6]
+                if any(theme == d["theme"] for d in output): continue
+                out = self.do_get_colors(p)
+                output.append({ "theme": theme, "colors": out})
+            print(len(output))
```
krobelus, line 1329: looks like leftover debug code

share/tools/web_config/webconfig.py
```diff
@@ -941,7 +941,23 @@ def do_get_colors(self, path=None):
...
+            # Lines starting with "#" can contain metadata.
+            if line.startswith("#"):
+                if not ":" in line: continue
+                key, value = line.split(":", maxsplit=1)
+                key = key.strip("# '")
+                value = value.strip(" '\"")
+                # Only use keys we know
+                if not key in ["name", "preferred_background", "url"]: continue
+                if key == "preferred_background":
+                    if not value in named_colors and not value.startswith("#"):
```
krobelus, line 957: I think we can say `if value not in ...`

share/tools/web_config/webconfig.py
```diff
@@ -941,7 +941,23 @@ def do_get_colors(self, path=None):
...
+            # Ignore empty lines
+            if not line: continue
+            # Lines starting with "#" can contain metadata.
+            if line.startswith("#"):
+                if not ":" in line: continue
+                key, value = line.split(":", maxsplit=1)
+                key = key.strip("# '")
+                value = value.strip(" '\"")
+                # Only use keys we know
+                if not key in ["name", "preferred_background", "url"]: continue
```
krobelus, line 955: I was told that tuples should be used over lists in this case

faho:
So I'm close on the CLI. Should I add that to this PR or merge this (which doesn't really *do* a lot that's visible) and open another?

krobelus:
What if we changed the theme format to (retricted) fish

so `set fish_color_normal 575F66` instead of `fish_color_normal 575F66`?

then it would be fairly obvious how to load the theme - just source the file.
The canonical way would still be through fish_config (also CLI), which will offer proper completions.
Or would that just cause confusion, because people might try to use fish features in their themes, which won't work.
So maybe a new format brings the least trouble.
As long as fish_config can load arbitrary files as themes, so users can experiment, it's probably good enough.

This seems good to merge (I'd run the formatter though).

---

If the addition + removal of the themes were in the same commit, it would be slightly easier to trace the evolution of the themes. But I didn't see a good way to do that while keeping the commit structure.

faho:
>What if we changed the theme format to (retricted) fish

That just means people start running `source` on it, and start putting unrestricted fish in it.

Also that makes it harder to pick the scope of the variable - in the current CLI I run `set -g` or `set -U` depending on the arguments, and once you add a `set` here you need to weirdly insert the scope in the middle.

Overall, I *like* having a clear line between a .theme and a .fish.

An alternative would of course be picking an actual serialization format, but most of them are more annoying to read (and especially more annoying to *generate* from the shell).

Possibly `var=val`, but that makes dealing with lists annoying - how do you store a variable set like `set -U fish_color_status '--background=red'  'white'`? Do you just go `fish_color_status=--background=red white` or is it `fish_color_status='--background=red white'` and you have to tokenize again?

faho:
Alright, black doesn't really change a lot, so I'm gonna merge this now and run a full formatting pass later.

This also allows me to open another PR for the CLI bit of this, I have some open questions.

Issue #3954, opened 2017-04-14
support bash style git completion for git command aliases containing a "null command"


rocketraman:
```
~ fish --version
fish, version 2.3.1
~ echo $version
2.3.1
```

Git completion for bash has a very useful feature in which a null command can be used to assign the completion command for an alias. Here is the description: https://github.com/git/git/blob/master/contrib/completion/git-completion.bash#L26.

It would be great if the git completion built in to fish would support this as well.

Note that, for some reason, I can't get this to work with bash, so I suppose its possible that my aliases are not set up correctly and fish already supports this, but I don't think that's the case.

krader1961:
If I've correctly understood the comment you linked to this is done in fish using the `--wraps` flag. For example,

```
function f --wraps git
    do something
end
```

If that doesn't do what you want we're going to need more information about the behavior you're looking for.

faho:
@krader1961: This is about a _git alias_ that calls a shell function. I.e. you'd define in the git config

    [alias]
        banana = "! somefunc () { : git commit; do something; }; somefunc"

which causes `git banana` to execute the defined "somefunc" function in a shell. I'm not sure how it determines that - it's possible it just uses $SHELL (which is usually not what you want).

Honestly, this doesn't sound supportable for us, because it requires parsing arbitrary (possibly POSIX-) shell script.

What could help here is to allow the `--wraps` option to wrap more than the command - in that case you could add `complete -c "git banana" -w "git commit"` and have completions working. That's #1976.

krader1961:
Thanks, @faho, for pointing me in the right direction. Starting with your question about whether git honors $SHELL a simple experiment shows that git always uses /bin/sh:

```
$ git alias wtf '!testme() { echo hello; }; testme'
$ set -x GIT_TRACE 2
$ git wtf
17:55:11.120624 git.c:595               trace: exec: 'git-wtf'
17:55:11.121216 run-command.c:369       trace: run_command: 'git-wtf'
17:55:11.123133 run-command.c:369       trace: run_command: 'testme() { echo hello; }; testme'
17:55:11.124011 run-command.c:228       trace: exec: '/bin/sh' '-c' 'testme() { echo hello; }; testme' 'testme() { echo hello; }; testme'
hello
```

The bash git completion uses `git config --get alias.wtf` to get the definition of the alias and parses it to extract the completion type. See the `__git_aliased_command` function in the script pointed to in the original comment.

[...]

PR #8129, opened 2021-07-10
Replicate upstream git logic for alias commands

rocketraman:
## Description

This (partially) replicates the git upstream logic for extracting a git command from a git alias. Currently fish ignores all aliases that start with `!` -- we now attempt to parse these instead.

Fixes issue #3954.

## TODOs:

### DIFF
--- a/share/completions/git.fish
+++ b/share/completions/git.fish
@@ -593,16 +593,30 @@ end
 # So instead, we store the aliases in global variables, named after the alias, containing the command.
 # This is because alias:command is an n:1 mapping (an alias can only have one corresponding command,
 #                                                  but a command can be aliased multiple times)
-git config -z --get-regexp 'alias\..*' | while read -lz alias command _
-    # If the command starts with a "!", it's a shell command, run with /bin/sh,
-    # or any other shell defined at git's build time.
-    #
-    # We can't do anything with them, and we run git-config again for listing aliases,
-    # so we skip them here.
-    string match -q '!*' -- $command; and continue
+git config -z --get-regexp 'alias\..*' | while read -lz alias cmdline
+    set -l command (__fish_git_aliased_command $cmdline)
+    string match -q --regex '\w+' -- $command; or continue
     # Git aliases can contain chars that variable names can't - escape them.
     set -l alias (string replace 'alias.' '' -- $alias | string escape --style=var)
     set -g __fish_git_alias_$alias $command
+end
+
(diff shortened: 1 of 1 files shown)


share/completions/git.fish
```diff
@@ -593,18 +593,32 @@ end
...
-    string match -q '!*' -- $command; and continue
+git config -z --get-regexp 'alias\..*' | while read -lz alias cmdline
+    set -l command (__fish_git_aliased_command $cmdline)
+    string match -q --regex '\w+' -- $command; or continue
     # Git aliases can contain chars that variable names can't - escape them.
     set -l alias (string replace 'alias.' '' -- $alias | string escape --style=var)
     set -g __fish_git_alias_$alias $command
 end
 
+# Approximately duplicates the logic from https://github.com/git/git/blob/master/contrib/completion/git-completion.bash#L1130
```
faho, line 604: This is a link directly to the master branch, it is *very* likely to go out of date.

When you click on the "..." menu on the github web interface, select "copy permalink" to get https://github.com/git/git/blob/d486ca60a51c9cb1fe068803c3f540724e95e83a/contrib/completion/git-completion.bash#L1130. This links to the current master *commit*, which would stay the same.

share/completions/git.fish
```diff
@@ -593,18 +593,32 @@ end
...
+# The Git script also finds aliases that reference other aliases via a loop but this is fine for a PoC
+function __fish_git_aliased_command
+    for word in (string split ' ' -- $argv)
+        switch $word
+            case !gitk gitk
+                echo "gitk"
+                return
+            # Adding " to the list
+            case '!*' '-*' '*=*' git '()' '{' : '\'*' '"*'
```
faho, line 613: This includes double-quotes, the bash version only has *single* quote?

I would probably use *both* - we're not tokenizing here, so we can't handle any kind of quotes.

(also this might break on anything with quotes *inside* or any kind of escaping or... I'm guessing that'll just lead to a non-working completion)

rocketraman: Yes, the double quotes are there simply to handle aliases of the form `!sh -c "..."` as well as the form `!sh -c '...'`. I find it a bit surprising the git logic only looks for single quotes -- unless I'm missing something, I don't see any down side to including the double-quote here.

Yes, there are all sorts of scenarios when this type of token-based parsing won't work correctly, but a) that is true for git's own logic as well, and b) this won't be any worse than not attempting parsing at all.

rocketraman:
@faho I've updated the link to the git repo to point to a specific commit. Let me know what else I should do here. I'm using this locally and getting significantly better git alias completion in fish.

faho:
Merged, thanks!

Issue #391, opened 2012-11-14
funced the whole .fish file contents, not only function body


maxfl:
Sometime the <function>.fish file contains not only the function definition, but some variable definitions. Sometimes the function definition depends on a system, so there is if/function/else/function construction.

When funced is invoked it let you edit only the function name, options and body. After applying funcsave all the other information is lost.

It would be very helpful if function could contain the whole contents of the .fish file instead. So when you type funced, you can also modify the preceeding variables, conditions and comments.

krader1961:
This should be easy to do after issue #3295 is implemented.

krader1961:
As discussed in issue #3701 we should do two things:

1) Make it possible to edit the file in which the function was originally defined using the `funced` command.

2) Warn the user about losing their edits if they don't invoke `funced` with the `--save` or `--perm` flag (the flag name is still open for discussion).

To be determined is whether we can, or should, attempt to protect the user from doing `funcsave $func_name` when the file in which the function was defined contains other statements and comments.

PR #8130, opened 2021-07-10
funced: edit the whole file, not just the function definition

zanchey:
Fixes issue #391.

I'd appreciate feedback on the approach, as well as the following specifics
(plus anything else relevant):

- Are the error messages appropriate?
- When used without `--save`, all the extra parts of the file are lost (the
  function definition which has been sourced overrides the autoloaded files).
  They are still in effect but cannot be edited easily any more. Is this ok?
  Should we make `--save` the default? (I don't think so.)

## TODOs:

### DIFF
--- a/share/functions/funced.fish
+++ b/share/functions/funced.fish
@@ -107,7 +107,7 @@ function funced --description 'Edit function definition'
                 end
             end
 
-            if not source $tmpname
+            if not source <$tmpname
                 # Failed to source the function file. Prompt to try again.
                 echo # add a line between the parse error and the prompt
                 set -l repeat
--- a/share/functions/funced.fish
+++ b/share/functions/funced.fish
@@ -83,10 +83,16 @@ function funced --description 'Edit function definition'
     or return 1
     set -l tmpname $tmpdir/$funcname.fish
 
-    if functions -q -- $funcname
-        functions -- $funcname >$tmpname
-    else
+    set -l writepath
+
+    if not functions -q -- $funcname
         echo $init >$tmpname
+    else if functions --details -- $funcname | string match --invert --quiet --regex '^(?:-|stdin)$'
+        set writepath (functions --details -- $funcname)
+        # Use cat here rather than cp to avoid copying permissions
+        cat "$writepath" >$tmpname
+    else
+        functions -- $funcname >$tmpname
     end
 
     # Repeatedly edit until it either parses successfully, or the user cancels
@@ -113,7 +119,7 @@ function funced --description 'Edit function definition'
                 # Failed to source the function file. Prompt to try again.
                 echo # add a line between the parse error and the prompt
                 set -l repeat
-                set -l prompt (_ 'Edit the file again\? [Y/n]')
+                set -l prompt (_ 'Edit the file again? [Y/n]')
                 read -p "echo $prompt\  " response
                 if test -z "$response"
                     or contains $response {Y,y}{E,e,}{S,s,}
@@ -124,6 +130,40 @@ function funced --description 'Edit function definition'
                     continue
                 end
                 echo (_ "Cancelled function editing")
+            else if test -n "$writepath"
+                if not set -q _flag_save
+                    echo (_ "Warning: the file containing this function has not been saved. Changes may be lost when fish is closed.")
+                    set -l prompt (printf (_ 'Save function to %s? [Y/n]') "$writepath")
+                    read --prompt-str "$prompt " response
+                    if test -z "$response"
+                        or contains $response {Y,y}{E,e,}{S,s,}
+                        set _flag_save 1
+                    else if not contains $response {N,n}{O,o,}
+                        echo "I don't understand '$response', assuming 'Yes'"
+                        set _flag_save 1
+                    end
+                end
+                if set -q _flag_save
+                    # try to write the file back
+                    # cp preserves existing permissions, though it might overwrite the owner
+                    if cp $tmpname "$writepath" 2>&1
+                        printf (_ "Function saved to %s") "$writepath"
+                        echo
+                        # read it back again - this ensures that the output of `functions --details` is correct
+                        source "$writepath"
+                    else
+                        echo (_ "Saving to original location failed; saving to user configuration instead.")
+                        set writepath $__fish_config_dir/functions/(basename "$writepath")
+                        if cp $tmpname "$writepath"
+                            printf (_ "Function saved to %s") "$writepath"
+                            echo
+                            # read it back again - this ensures that the output of `functions --details` is correct
+                            source "$writepath"
+                        else
+                            echo (_ "Saving to user configuration failed. Changes may be lost when fish is closed.")
+                        end
+                    end
+                end
             else if set -q _flag_save
                 funcsave $funcname
             end
--- a/share/functions/funced.fish
+++ b/share/functions/funced.fish
@@ -104,6 +104,8 @@ function funced --description 'Edit function definition'
                 set -l new_checksum (__funced_md5 "$tmpname")
                 if test "$new_checksum" = "$checksum"
                     echo (_ "Editor exited but the function was not modified")
+                    # Don't source or save an unmodified file.
+                    break
                 end
             end
 
--- a/share/functions/funced.fish
+++ b/share/functions/funced.fish
@@ -166,6 +166,9 @@ function funced --description 'Edit function definition'
                 end
             else if set -q _flag_save
                 funcsave $funcname
+            else
+                printf (_ "Run funcsave %s to save this function to the configuration directory.") $funcname
+                echo
             end
         end
         break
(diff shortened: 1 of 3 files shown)


share/functions/funced.fish
```diff
@@ -83,10 +83,16 @@ function funced --description 'Edit function definition'
...
     set -l tmpname $tmpdir/$funcname.fish
 
-    if functions -q -- $funcname
-        functions -- $funcname >$tmpname
-    else
+    set -l writepath
+
+    if not functions -q -- $funcname
         echo $init >$tmpname
+    else if functions --details -- $funcname | string match --invert --regex '^-$'
```
faho, line 90: Okay, so this path is for functions that have been defined in a file?

Note that interactively defined functions say "stdin" here, the `-` bit only applies to functions `source`d from stdin.

We might want to clean that up - see src/builtin_functions.cpp.

share/functions/funced.fish
```diff
@@ -104,10 +110,12 @@ function funced --description 'Edit function definition'
                 set -l new_checksum (__funced_md5 "$tmpname")
                 if test "$new_checksum" = "$checksum"
                     echo (_ "Editor exited but the function was not modified")
+                    # Don't source or save an unmodified file.
+                    break
                 end
             end
 
-            if not source $tmpname
+            if not source <$tmpname
```
faho, line 118: This changes it so that `source` doesn't know about the filename anymore.

That means after you've `funced` a function once (without `--save`) it has no file anymore and was now defined on stdin.

Is there any particular reason for that? Would it not remember the tempfile, and then a second `funced` would use that again, so the things outside of the function stay there?

(still not perfect because what if you run `funced fish_git_prompt` and then `funced __fish_git_prompt_operation_branch_bare` or any other of the helper functions)

faho:
>Are the error messages appropriate?

They seem alright to me.

share/functions/funced.fish
zanchey: Thanks, I'll add that to the expression.

share/functions/funced.fish
zanchey: The problem is that the temporary file is cleaned up by `funced`. So yes, without `--save` you may lose the ability to edit the file properly again - hence the warnings. We could leave the tempfile around, perhaps by moving it to `XDG_CACHE_HOME` (people do run programs like `tmpreaper`).

share/functions/funced.fish
```diff
@@ -83,10 +83,16 @@ function funced --description 'Edit function definition'
...
     set -l tmpname $tmpdir/$funcname.fish
 
-    if functions -q -- $funcname
-        functions -- $funcname >$tmpname
-    else
+    set -l writepath
+
+    if not functions -q -- $funcname
         echo $init >$tmpname
+    else if functions --details -- $funcname | string match --invert --regex '^(?:-|stdin)$'
```
ridiculousfish, line 90: This causes it to output the file path to stdout. I think we should add `--quiet` as a flag.

share/functions/funced.fish
```diff
@@ -122,6 +130,26 @@ function funced --description 'Edit function definition'
...
+                        else
+                            # read it back again - this ensures that the output of `functions --details` is correct
+                            source "$writepath"
+                        end
+                    else
+                        # read it back again - this ensures that the output of `functions --details` is correct
+                        source "$writepath"
+                    end
+                else
+                    echo (_ "Warning: file containing function was edited but not saved. Changes will be lost when fish is closed.")
```
ridiculousfish, line 151: I found this message to be a little scary, and it doesn't tell the user what to do. I wish we could suggest `funcsave`, but that reintroduces the problem we are trying to solve. I also wish we could suggest `funced --save` but that doesn't work because the original file is lost.

What would you think about prompting the user to save the function if `--save` is not passed?

    > funced fish_prompt
    ...
    Save function to /home/peter/.config/fish/functions/fish_prompt? [Y/n]  Y
    Function saved to /home/peter/.config/fish/functions/fish_prompt

share/functions/funced.fish
```diff
@@ -122,6 +130,26 @@ function funced --description 'Edit function definition'
...
+                        echo (_ "Saving to original location failed; saving to user configuration instead.")
+                        set writepath $__fish_config_dir/functions/(basename "$writepath")
+                        if not cp $tmpname "$writepath"
+                            echo (_ "Saving to user configuration failed. Changes will be lost when fish is closed.")
+                        else
+                            # read it back again - this ensures that the output of `functions --details` is correct
+                            source "$writepath"
+                        end
+                    else
+                        # read it back again - this ensures that the output of `functions --details` is correct
```
ridiculousfish, line 147: I suggest adding a confirmation message here, perhaps:

    `Function saved to $writepath`

ridiculousfish:
Overall this looks really nice and is a big improvement.

share/functions/funced.fish
zanchey: Yes, that sounds great! I think that's the missing link. I'm tempted to add that for all invocations, not just those that edit a file - what do you think?

share/functions/funced.fish
zanchey: Good idea.

share/functions/funced.fish
ridiculousfish: Yeah I agree!

zanchey: I've done that, but I'm not sure about it - breaks the otherwise simple workflow of `funced`. Anyway perhaps we can merge it and see what people think.

zanchey:
The control flow here is pretty gnarly but I can't see how to untangle it easily. Any testing is appreciated, otherwise I'll merge soon.

share/functions/funced.fish
```diff
@@ -122,8 +130,50 @@ function funced --description 'Edit function definition'
                     continue
                 end
                 echo (_ "Cancelled function editing")
+            else if test -n "$writepath"
+                if not set -q _flag_save
+                    echo (_ "Warning: the file containing this function has not been saved. Changes will be lost when fish is closed.")
+                    set -l prompt (printf (_ 'Save function to %s\? [Y/n]') "$writepath")
+                    read -p "echo $prompt\  " response
+                    if test -z "$response"
+                        or contains $response {Y,y}{E,e,}{S,s,}
```
faho, line 139: The default should probably be to not save.

share/functions/funced.fish
```diff
@@ -122,8 +130,50 @@ function funced --description 'Edit function definition'
                     continue
                 end
                 echo (_ "Cancelled function editing")
+            else if test -n "$writepath"
+                if not set -q _flag_save
+                    echo (_ "Warning: the file containing this function has not been saved. Changes will be lost when fish is closed.")
+                    set -l prompt (printf (_ 'Save function to %s\? [Y/n]') "$writepath")
+                    read -p "echo $prompt\  " response
```
faho, line 137: This can just be `read -P "$prompt "` (`--prompt-str`)

share/functions/funced.fish
```diff
@@ -122,8 +130,50 @@ function funced --description 'Edit function definition'
                     continue
                 end
                 echo (_ "Cancelled function editing")
+            else if test -n "$writepath"
+                if not set -q _flag_save
+                    echo (_ "Warning: the file containing this function has not been saved. Changes will be lost when fish is closed.")
```
faho, line 135: No need for the `echo`, just `_ "Warning..."`

share/functions/funced.fish
faho: (or even `read -p "printf (_ 'Save function to %s\? [Y/n]') \"$writepath\""`, or `read -p "$(printf (_ 'Save function to %s\? [Y/n]') "$writepath")"`

share/functions/funced.fish
```diff
@@ -122,8 +130,50 @@ function funced --description 'Edit function definition'
...
+                        set _flag_save 1
+                    else if not contains $response {N,n}{O,o,}
+                        echo "I don't understand '$response', assuming 'Yes'"
+                        set _flag_save 1
+                    end
+                end
+                if set -q _flag_save
+                    # try to write the file back
+                    # cp preserves existing permissions, though it might overwrite the owner
+                    if not cp $tmpname "$writepath" 2>&1
```
faho, line 149: I suggest using the happy path first - `if cp`, and then the failure case in the `else`.

share/functions/funced.fish
```diff
@@ -122,8 +130,50 @@ function funced --description 'Edit function definition'
...
+                    else
+                        printf (_ "Function saved to %s") "$writepath"
+                        # read it back again - this ensures that the output of `functions --details` is correct
+                        source "$writepath"
+                    end
+                end
             else if set -q _flag_save
                 funcsave $funcname
+            else
+                set -l prompt (_ 'Save function to the fish configuration directory\? [Y/n]')
```
faho, line 168: To be honest, I'm not a massive fan of this idea.

I'd just suggest running `funcsave $funcname` instead of prompting.

faho (changes requested):

share/functions/funced.fish
faho: (I am happy to be outvoted on this, of course)

zanchey: Let's drop it for now.

PR #8132, opened 2021-07-11
fish_config: Add CLI-based theme selector

faho:
## Description

`fish_config theme`:

- `list` to list all available themes (files in the two theme
directories - either the web_config/themes one or
~/.config/fish/themes!)
- `show` to show select (or all) themes right in the terminal - this
starts another fish that reads the theme file and prints the sample
text, manually colored
- `choose` to load a theme *now*, setting the variables globally
- `save` to load a theme and save the variables universally
- `dump` to write the current theme in .theme format (to stdout)

Part of #3625.

A screenshot:

Screenshot_20210711_222310

## TODOs:

Open questions:

- Should we have separate `choose` and `save` commands (like `fish_config prompt`!)? Should we offer universal variables here, or a conf.d snippet? Should we have `save` at all?
- What about pre-defined universal variables? What about saving if there are already globals?
- Should dumping write to a file?
- Do we read from ~/.config/fish/themes? If yes, we should also add it to the web version. (at which point the format of .theme files becomes a lot more important!)

(and yes, docs and changelog and tests are still missing)

### DIFF
--- a/doc_src/cmds/fish_config.rst
+++ b/doc_src/cmds/fish_config.rst
@@ -11,6 +11,7 @@ Synopsis
     fish_config
     fish_config browse
     fish_config prompt (choose | list | save | show)
+    fish_config theme (choose | demo | dump | list | save | show)
 
 Description
 -----------
@@ -29,6 +30,19 @@ Available subcommands for the ``prompt`` command:
 - ``list`` lists the names of the available sample prompts.
 - ``save`` saves the current prompt to a file (via :ref:`funcsave <cmd-funcsave>`).
 - ``show`` shows what the given sample prompts (or all) would look like.
+
+With the ``theme`` command ``fish_config`` can be used to view and choose a theme (meaning a color scheme) inside the terminal.
+
+Available subcommands for the ``theme`` command:
+
+- ``choose`` loads a sample theme in the current session.
+- ``demo`` displays some sample text in the current theme.
+- ``dump`` prints the current theme in a loadable format.
+- ``list`` lists the names of the available sample themes.
+- ``save`` saves the current prompt to :ref:`universal variables <variables-universal>`.
+- ``show`` shows what the given sample theme (or all) would look like.
+
+The themes are loaded from the theme directory shipped with fish or a ``themes`` directory in the fish configuration directory (typically ``~/.config/fish/themes``).
 
 Example
 -------
--- a/share/completions/fish_config.fish
+++ b/share/completions/fish_config.fish
@@ -1,5 +1,6 @@
 complete fish_config -f
 set -l prompt_commands choose save show list
+set -l theme_commands choose demo dump save show list
 complete fish_config -n '__fish_use_subcommand' -a prompt -d 'View and pick from the sample prompts'
 complete fish_config -n "__fish_seen_subcommand_from prompt; and not __fish_seen_subcommand_from $prompt_commands" \
     -a choose -d 'View and pick from the sample prompts'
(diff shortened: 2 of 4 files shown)


ridiculousfish:
This is sweet! It looks awesome and works great for me.

Some nits: for `fish_config theme show` I suggest showing the current theme first, for comparison. Also the list of themes somewhat runs together vertically - perhaps indent the sample text by 2 spaces, and/or add a newline before each title.

> Should we have separate choose and save commands (like fish_config prompt!)?

I think we should be consistent with `fish_config prompt`. If we change one let's change both.

> What about pre-defined universal variables? What about saving if there are already globals?

Could we print a warning if there's a global shadowing the uvar? I really do think universal variables will go away eventually, and there will only be globals.

> Should dumping write to a file?

I love it how it is: prints to stdout and the user is free to redirect it.

> Do we read from ~/.config/fish/themes? 

That's a tough question. I suspect themes will end up being fish script, not JSON or other structured formats, so if the theme files are scripts then I don't think we'll paint ourselves into a corner. Do you have plans for how to populate the `~/.config/fish/themes` directory, e.g. a separate git repo or something?

faho:
>Also the list of themes somewhat runs together vertically - perhaps indent the sample text by 2 spaces, and/or add a newline before each title.

That makes sense!

> Could we print a warning if there's a global shadowing the uvar?

We already do!

```fish
> set -g fish_user_paths $fish_user_paths
> set -U fish_user_paths $fish_user_paths
set: Universal variable 'fish_user_paths' is shadowed by the global variable of the same name.
```

>I suspect themes will end up being fish script, not JSON or other structured formats

In my solution, they are *not* scripts. They really are just *data*. And I think that's quite important.

The syntax is "each line contains whatever comes after `set -U `, with quoting but without expansions".

I'm quite uncomfortable with making them actual scripts, especially if we end up making it available to third parties.

(unfortunately "theme" is used by some third party things to mean "prompt and greeting and such", while this is about a color scheme)

>Do you have plans for how to populate the ~/.config/fish/themes directory, e.g. a separate git repo or something?

If we don't have code here, it doesn't really matter. Someone could make a website, we could make a theme of the month contest on fishshell.com, a separate git repo, downloading random themes on startup, ...

faho:
Okay, I moved the printing of the sample text to a new `demo` subcommand. You can use that to show off your own colorscheme.

This makes it possible to display the current colorscheme first. I'm not *quite* sold on doing that, because it's too long for a single screen, so you'd need to scroll up to see it again. We could do it last, or we could remove a few themes? Are all of these really used?

Or should we make the lines longer? Move the comment on the first line?

faho:
Alright, tacking the comment onto the end of the first line seems to be a winner:

Screenshot_20210712_210338

At least in the terminal, given how little we can use the width otherwise. It's not a problem in the browser, flexbox will take care of it.

ridiculousfish:
The added spacing + comment motion makes it even better, it looks very nice to me.

> This makes it possible to display the current colorscheme first. I'm not quite sold on doing that, because it's too long for a single screen, so you'd need to scroll up to see it again. We could do it last, or we could remove a few themes? Are all of these really used?

My typical $LINES is 34 which fits 6.5 themes. I suppose it could be doubled if we used two columns, or even more columns if we shortened some of the text. "bright vixens jump; dozy fowl quack" is just a silly pangram I found.

But we shouldn't allow perfect to be the enemy of the good. I personally don't mind scrolling, and printing a long theme list to stdout seems like a fine place to start.

faho:
Alright, I added some docs and completions and made the web version also check ~/.config/fish/themes.

Merging!

PR #8141, opened 2021-07-14
Add completion for azure cli (az)

kfkonrad:
## Description

This introduces completions for the Azure CLI `az`. `az` uses the argcomplete library for its completions and the code I've used is little more than a boilerplate snippet to access those completions.

argcomplete has been compatible with fish for over 2 years now ([argcomplete!260](https://github.com/kislyuk/argcomplete/pull/260)), but this is (to my knowledge) the first time this is being added to fish-shell directly. The code is basically copied from a snippet mentioned in argcomplete!260, except I replaced `myscript` with `az`.

## TODOs:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -30,7 +30,7 @@ Completions
   - Angular's ``ng`` (:issue:`8111`)
   - ``zef`` (:issue:`8114`)
   - ``rakudo`` (:issue:`8113`)
-  - ``az``
+  - ``az`` (:issue:`8141`)
 
 - Improvements to many completions.
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -30,6 +30,7 @@ Completions
   - Angular's ``ng`` (:issue:`8111`)
   - ``zef`` (:issue:`8114`)
   - ``rakudo`` (:issue:`8113`)
+  - ``az``
 
 - Improvements to many completions.
 
--- a/share/completions/az.fish
+++ b/share/completions/az.fish
@@ -1,0 +1,16 @@
+function __fish_az_complete
+  set -x _ARGCOMPLETE 1
+  set -x _ARGCOMPLETE_IFS \n
+  set -x _ARGCOMPLETE_SUPPRESS_SPACE 1
+  set -x _ARGCOMPLETE_SHELL fish
+  set -x COMP_LINE (commandline -p)
+  set -x COMP_POINT (string length (commandline -cp))
+  set -x COMP_TYPE
+  if set -q _ARC_DEBUG
+    az 8>&1 9>&2 1>/dev/null 2>&1
+  else
+    az 8>&1 9>&2 1>&9 2>&1
+  end
+end
+
+complete -c az -f -a '(__fish_az_complete)'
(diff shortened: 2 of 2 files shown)


share/completions/az.fish
```diff
@@ -0,0 +1,16 @@
+function __fish_az_complete
+  set -x _ARGCOMPLETE 1
```
faho, line 2: All of these should explicitly be set in local scope - `set -lx`. Otherwise you'd clobber any global (or worse, universal) variable of that name.

share/completions/az.fish
```diff
@@ -0,0 +1,16 @@
...
+  set -x _ARGCOMPLETE_IFS \n
+  set -x _ARGCOMPLETE_SUPPRESS_SPACE 1
+  set -x _ARGCOMPLETE_SHELL fish
+  set -x COMP_LINE (commandline -p)
+  set -x COMP_POINT (string length (commandline -cp))
+  set -x COMP_TYPE
+  if set -q _ARC_DEBUG
```
faho, line 9: We're not them, no need to keep the debugging guff.

share/completions/az.fish
```diff
@@ -0,0 +1,16 @@
...
+    az 8>&1 9>&2 1>/dev/null 2>&1
+  else
+    az 8>&1 9>&2 1>&9 2>&1
```
faho, line 12: What the hedge is that fd awkwardness? If I read this correctly it swaps stdout and stderr around. Do they really send the actual requested output to stderr and errors to stdout?

If so, just redirect stderr to stdout and silence stdout:

```fish
az 2>&1 >/dev/null
```

Or do they actually expect fd 8 and 9 to be set?

What happens if you leave it off?

share/completions/az.fish
```diff
@@ -0,0 +1,16 @@
+function __fish_az_complete
+  set -x _ARGCOMPLETE 1
+  set -x _ARGCOMPLETE_IFS \n
+  set -x _ARGCOMPLETE_SUPPRESS_SPACE 1
+  set -x _ARGCOMPLETE_SHELL fish
+  set -x COMP_LINE (commandline -p)
```
faho, line 6: You typically want `commandline -pc`, to cut off the bit after the cursor.

That allows you to go back and change an argument.

CHANGELOG.rst
```diff
@@ -30,6 +30,7 @@ Completions
   - Angular's ``ng`` (:issue:`8111`)
   - ``zef`` (:issue:`8114`)
   - ``rakudo`` (:issue:`8113`)
+  - ``az``
```
faho, line 33: Since you now know this is PR 8141, you can add the issue number to it as well.

faho (changes requested):

share/completions/az.fish
kfkonrad: Done

share/completions/az.fish
kfkonrad: Done

share/completions/az.fish
kfkonrad: That's actually needed. I found no difference when I left of the third redirection and I've silenced stderr, but the other ones are really needed for az to provide completions correctly

share/completions/az.fish
kfkonrad: I'm not sure this makes sense since az has subcommands with subcommands which don't means swapping out one subcommand in the chain generally won't yield a valid command, but other than that it's fine. I've added the parameter

CHANGELOG.rst
kfkonrad: Done

faho:
Merged, thanks!

PR #8142, opened 2021-07-14
commandline: Add --is-complete option to query whether it's syntactically complete

### LINKED ISSUE #7602: Allow updating fish_prompt after command (transient prompt)
I'm working on trying to replicate powerlevel10k's awesome [transient prompt feature](https://github.com/romkatv/powerlevel10k/#transient-prompt).

Here's my first stab:

```fish
function fish_prompt
    if set -q foo
        echo $foo
        set -e foo
    else
        echo $PWD
    end
end

function fish_preexec --on-event fish_preexec
    set -g foo bar
    commandline -f repaint
end
```

The logic is as follow: 
1. `foo` starts out undefined, so `fish_prompt` echoes `$PWD`
2. Then you hit enter on a command, causing `fish_preexec` to run, which sets foo and reruns the prompt
3. The prompt changes to `$foo` and erases it
4. The command starts running

Thus the currently prompt will always be `$PWD` and all previous ones will be `$foo`

It doesnt work. Is there a way of doing what I want? If not, can it be added?

faho:
This means querying when the commandline is in a state that it could
be executed. Because our `execute` bind function also inserts a
newline if it isn't.

One case that's not handled right now: `execute` also expands
abbreviations, those can technically make the commandline invalid
again.

Unfortunately we have no real way to *check* without doing the
replacement.

Also since abbreviations are only available in command position when
you _execute_ them the commandline will most likely be valid.

This is enough to make transient prompts work:

```fish
function reset-transient --on-event fish_postexec
    set -g TRANSIENT 0
end

function maybe_execute
    if commandline --is-complete
        set -g TRANSIENT 1
        commandline -f repaint
    else
        set -g TRANSIENT 0
    end
    commandline -f execute
end

bind \r maybe_execute
```

and then in `fish_prompt` react to $TRANSIENT being set to 1.

As far as I'm concerned that means it:

Fixes #7602

## TODOs:

### DIFF
--- a/src/builtin_commandline.cpp
+++ b/src/builtin_commandline.cpp
@@ -396,8 +396,13 @@ maybe_t<int> builtin_commandline(parser_t &parser, io_streams_t &streams, const 
     if (is_valid) {
         if (!*current_buffer) return 1;
         parser_test_error_bits_t res =
-            parse_util_detect_errors(current_buffer, NULL, false /* do not accept incomplete */);
-        return res & PARSER_TEST_ERROR ? 1 : 0;
+            parse_util_detect_errors(current_buffer,
+                                     NULL,
+                                     true /* accept incomplete so we can tell the difference */);
+        if (res & PARSER_TEST_INCOMPLETE) {
+            return 2;
+        }
+        return res & PARSER_TEST_ERROR ? STATUS_CMD_ERROR : STATUS_CMD_OK;
     }
 
     switch (buffer_part) {
--- a/src/builtin_commandline.cpp
+++ b/src/builtin_commandline.cpp
@@ -394,10 +394,9 @@ maybe_t<int> builtin_commandline(parser_t &parser, io_streams_t &streams, const 
     }
 
     if (is_valid) {
-        auto buff = reader_get_buffer();
-        if (!*buff) return 1;
+        if (rstate.text.empty()) return 1;
         parser_test_error_bits_t res =
-            parse_util_detect_errors(buff, NULL, false /* do not accept incomplete */);
+            parse_util_detect_errors(rstate.text, NULL, false /* do not accept incomplete */);
         return res & PARSER_TEST_ERROR ? 1 : 0;
     }
 
--- a/src/builtin_commandline.cpp
+++ b/src/builtin_commandline.cpp
@@ -142,6 +142,7 @@ maybe_t<int> builtin_commandline(parser_t &parser, io_streams_t &streams, const 
     bool line_mode = false;
     bool search_mode = false;
     bool paging_mode = false;
+    bool is_complete = false;
     const wchar_t *begin = nullptr, *end = nullptr;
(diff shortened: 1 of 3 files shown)


faho:
Alright, I renamed it to "--is-valid" to avoid the idea that it's about completions. No short option because this is a rarely-used thing.

src/builtin_commandline.cpp
```diff
@@ -388,6 +394,13 @@ maybe_t<int> builtin_commandline(parser_t &parser, io_streams_t &streams, const
         return STATUS_CMD_ERROR;
     }
 
+    if (is_valid) {
+        if (!*current_buffer) return 1;
+        parser_test_error_bits_t res =
+            parse_util_detect_errors(current_buffer, NULL, false /* do not accept incomplete */);
+        return res & PARSER_TEST_ERROR ? 1 : 0;
```
ridiculousfish, line 401: Probably should be `STATUS_CMD_ERROR/OK` instead of `1/0`

ridiculousfish:
There's three possibilities: the command is valid and complete, the command is valid but incomplete, the command is invalid. Maybe we should distinguish those by the status code.

faho:
Okay, this is valid functionality, so I'm merging this now.

It's also *sufficient* to implement transient prompts, but the way to do that would be a bit of a hack. We should probably make either repaint work from a preexec handler or give some other way to set the prompt string from there, that's the obvious way to do it.

IlanCosman:
Is there any documentation on what the terms "valid" and "complete" mean? I don't really understand the difference 😄

faho:
>Is there any documentation on what the terms "valid" and "complete" mean?

"Valid" means that it's a full commandline. It could be executed in that state, and wouldn't give you a syntax error (of course you could still get argument errors and such - `printf --unknown` is syntactically valid, but `printf` will tell you it doesn't know the `--unknown` option).

"Incomplete" means that what is there could be expanded to become a full commandline - it's not done, it might have unclosed quotes or parentheses, but it could become valid.

The third thing is "invalid" or "erroneous". It means that what is there has a syntax error.

I wouldn't put too much into the difference, there's a reason both return falsy and there's no way to select them. This is a fairly direct line into the parser, and it was developed to figure out what to expand and highlight. There are some weird edge cases - `echo $$` is invalid (no valid variable name), `echo "$$` is incomplete (missing closing quote apparently trumps that?). Really, you probably just care about valid vs not.

(and I stopped using the term "complete" here because that sounds like we're talking about completions)

faho:
>There are some weird edge cases - echo $$ is invalid (no valid variable name), echo "$$ is incomplete (missing closing quote apparently trumps that?)

Note: The `"` trumps simply because it is first - `echo $$ "` is invalid again.

Also a commandline being incomplete means, among other things, that pressing enter (triggering the "execute" bind function) inserts a newline instead of executing the commandline - the assumption is that you'd put the needed bit on a new line.

PR #8145, opened 2021-07-15
Add `set --function`

### LINKED ISSUE #565: add function scope option to the `set` command
Doc of `set` builtin says:

> The scoping rules when creating or updating a variable are:
> 1. If a variable is explicitly set to either universal, global or local, that setting will be honored. If a variable of the same name exists in a different scope, that variable will not be changed.
> 2. If a variable is not explicitly set to be either universal, global or local, but has been previously defined, the previous variable scope is used.
> 3. If a variable is not explicitly set to be either universal, global or local and has never before been defined, the variable will be local to the currently executing function. Note that this is different from using the -l or –local flag. If one of those flags is used, the variable will be local to the most inner currently executing block, while without these the variable will be local to the function. If no function is executing, the variable will be global.

The upside is that well-known global variables (`$PATH`, `$EDITOR`, etc.) can be set within functions without an explicit `-g`. However, the behavior of the following function can be confusing:

```
function f
    set var val
end
```

If the user has a global variable named `var`, this function alters the global environment. Otherwise it has no impact on global environment. To make the function safe one has to always write

```
function f
    set -l var val
end
```

which is quite tedious. Therefore I propose that `set` should always imply function scope within functions.

faho:
~~This makes the "unspecified" scope available - the one that is used
when no variable exists.~~

~~It's either the enclosing function's topmost local scope, or global
scope if there is no function.~~ (after deliberation, using global felt awkward because that would make the variable available *everywhere*, so we're just local)

This removes the need to declare variables locally before use. E.g.:

```fish
set -l thing
if condition
    set thing one
else
    set thing two
end
```

could be written as

```fish
if condition
    set -f thing one
else
    set -f thing two
end
```

Note: Many scripts shipped with fish use workarounds like `and`/`or`
instead of `if`, so it isn't easy to find good examples.

Also, if there isn't an else-branch in that above, just with

```fish
if condition
    set -f thing one
end
```

that means something different from setting it before! Now, if
`condition` isn't true, it would use a global (or universal) variable of
te same name!

Some more interesting parts:

Because it *is* a local scope, setting a variable `-f` and
`-l` in the toplevel of a function ends up the same:

```fish
function foo2
    set -l foo bar
    set -f foo baz # modifies the *same* variable!
end
```

but setting it locally inside a block creates a new local variable
that shadows the function-scoped variable:

```fish
function foo3
    set -f foo bar
    begin
        set -l foo banana
        # $foo is banana
    end
    # $foo is bar again
end
```

This is how local variables already work. "Local" is actually "block-scoped".

Also `set --show` will only show the closest local scope, so it won't
show a shadowed function-level variable. Again, this is how local
variables already work, and could be done as a separate change.

Fixes #565

One of the tests explicitly checks one case mentioned by @ridiculousfish in https://github.com/fish-shell/fish-shell/issues/565#issuecomment-259046107:

```fish
function f
    set -l fruit banana
    if true
        set -f fruit orange
    end
    echo $fruit #banana
end
```

This isn't an issue, because function scope *is* the outermost local scope of the function. There is no difference. It's not a new node.

## TODOs:

### DIFF
--- a/doc_src/cmds/set.rst
+++ b/doc_src/cmds/set.rst
@@ -33,11 +33,11 @@ With ``--show``, ``set`` will describe the given variable names, explaining how 
 
 The following options control variable scope:
 
-- ``-l`` or ``--local`` forces the specified shell variable to be given a scope that is local to the current block, even if a variable with the given name exists and is non-local
+- ``-f`` or ``--function`` scopes the variable to the currently executing function. It is erased when the function ends.
 
-- ``-f`` or ``--function`` makes the variable local to the current *function*. This is the same as defining it with ``--local`` outside of a block like ``begin`` or ``for``. When used outside of a function, it uses the top-level local scope instead.
+- ``-l`` or ``--local`` scopes the variable to the currently executing block. It is erased when the block ends. Outside of a block, this is the same as ``--function``.
 
-- ``-g`` or ``--global`` causes the specified shell variable to be given a global scope. Non-global variables disappear when the block they belong to ends
+- ``-g`` or ``--global`` causes the specified shell variable to be given a global scope. Global variables don't disappear and are available to all functions running in the same shell. They can even be modified.
 
 - ``-U`` or ``--universal`` causes the specified shell variable to be given a universal scope. If this option is supplied, the variable will be shared between all the current user's fish instances on the current computer, and will be preserved across restarts of the shell.
 
--- a/doc_src/cmds/set.rst
+++ b/doc_src/cmds/set.rst
@@ -35,6 +35,8 @@ The following options control variable scope:
 
 - ``-l`` or ``--local`` forces the specified shell variable to be given a scope that is local to the current block, even if a variable with the given name exists and is non-local
 
+- ``-f`` or ``--function`` makes the variable local to the current *function*. This is the same as defining it with ``--local`` outside of a block like ``begin`` or ``for``. When used outside of a function, it uses the top-level local scope instead.
+
 - ``-g`` or ``--global`` causes the specified shell variable to be given a global scope. Non-global variables disappear when the block they belong to ends
 
 - ``-U`` or ``--universal`` causes the specified shell variable to be given a universal scope. If this option is supplied, the variable will be shared between all the current user's fish instances on the current computer, and will be preserved across restarts of the shell.
--- a/doc_src/language.rst
+++ b/doc_src/language.rst
@@ -875,14 +875,15 @@ There are three kinds of variables in fish: universal, global and local variable
 - Universal variables are shared between all fish sessions a user is running on one computer.
 - Global variables are specific to the current fish session, and will never be erased unless explicitly requested by using ``set -e``.
 - Local variables are specific to the current fish session, and associated with a specific block of commands, and automatically erased when a specific block goes out of scope. A block of commands is a series of commands that begins with one of the commands ``for``, ``while`` , ``if``, ``function``, ``begin`` or ``switch``, and ends with the command ``end``.
+- Function-local variables are the same as local variables outside of one of these blocks. They go out of scope when the current function ends.
 
-Variables can be explicitly set to be universal with the ``-U`` or ``--universal`` switch, global with the ``-g`` or ``--global`` switch, or local with the ``-l`` or ``--local`` switch.  The scoping rules when creating or updating a variable are:
+Variables can be explicitly set to be universal with the ``-U`` or ``--universal`` switch, global with ``-g`` or ``--global``, local to the current block with ``-l`` or ``--local``, or local to the current function with ``-f`` or ``--function``.  The scoping rules when creating or updating a variable are:
 
 - When a scope is explicitly given, it will be used. If a variable of the same name exists in a different scope, that variable will not be changed.
 
 - When no scope is given, but a variable of that name exists, the variable of the smallest scope will be modified. The scope will not be changed.
 
-- As a special case, when no scope is given and no variable has been defined the variable will belong to the scope of the currently executing *function*. This is different from the ``--local`` flag, which would make the variable local to the current *block*.
+- As a special case, when no scope is given and no variable has been defined the variable will belong to the scope of the currently executing *function*. This is different from the ``--local`` flag, which would make the variable local to the current *block*. Outside of a function, it is global, unlike with explicit function-scope.
 
 There can be many variables with the same name, but different scopes. When you :ref:`use a variable <expand-variable>`, the smallest scoped variable of that name will be used. If a local variable exists, it will be used instead of the global or universal variable of the same name.
 
@@ -897,6 +898,16 @@ Typically inside funcions you should use local scope::
         set -l file /path/to/my/file
         if not test -e "$file"
             set file /path/to/my/otherfile
+        end
+    end
+
+    # or
+
+    function something
+        if test -e /path/to/my/file
+            set -f file /path/to/my/file
+        else
+            set -f file /path/to/my/otherfile
         end
     end
 
@@ -922,16 +933,22 @@ If you want to set some personal customization, universal variables are nice::
 
 Here is an example of local vs function-scoped variables::
 
-    begin
-        # This is a nice local scope where all variables will die
-        set -l pirate 'There be treasure in them thar hills'
-        set captain Space, the final frontier
-    end
+  function test-scopes
+      begin
+          # This is a nice local scope where all variables will die
+          set -l pirate 'There be treasure in them thar hills'
+          set -f captain Space, the final frontier
+          # If no variable of that name was defined, it is function-local.
+          set gnu "In the beginning there was nothing, which exploded"
+      end
 
-    echo $pirate
-    # This will not output anything, since the pirate was local
-    echo $captain
-    # This will output the good Captain's speech since $captain had function-scope.
+      echo $pirate
+      # This will not output anything, since the pirate was local
+      echo $captain
+      # This will output the good Captain's speech since $captain had function-scope.
+      echo $gnu
+      # Will output Sir Terry's wisdom.
+  end
 
 .. _variables-override:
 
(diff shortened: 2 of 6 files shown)


tests/checks/set.fish
```diff
@@ -725,3 +725,68 @@ begin
...
+# Blocks aren't functions, "function" scope is still global:
+begin
+    set -f stillglobal "as global as the moon is wet"
+    echo $stillglobal
+    # CHECK: as global as the moon is wet
+end
+set -S stillglobal
+#CHECK: $stillglobal: set in global scope, unexported, with 1 elements
+#CHECK: $stillglobal[1]: |as global as the moon is wet|
+
```
krobelus, line 747: we could add a test for the case when a global of the same name is already defined

```fish
function test-function-scope
    set -f stillglobal 'setting variable in function scope will not clobber global variable'
end
test-function-scope
echo $stillglobal
#CHECK: as global as the moon is wet
```

krobelus (commented):
Looks good, this is really nice!
We should try to make this the default behavior eventually - during the transition period we could print a deprecation warning when a statement `set foo bar` inside a function sets a global or universal var.

faho:
>We should try to make this the default behavior eventually - during the transition period we could print a deprecation warning when a statement set foo bar inside a function sets a global or universal var.

Wait, you mean requiring an explicit `--global` to set a global variable? That's a large change that *could* work out, but has one wart: You'd have to choose between global and universal.

Anyway, that's for another time.

-----

I want to make *this* kind of scoping prominent. I believe that it's the much better model from `--local` (which is okay for specific situations and should be `set --block`, really).

But switching *that* is a much much much bigger change, so I would appreciate any ideas on how to do that - `--function` is a bit of an awkward name ("does this set a function?"). Is there anything better? `set --real-local`?

krobelus:
> You'd have to choose between global and universal.

That problem can go away once we ["merge" global and universal scopes](https://github.com/fish-shell/fish-shell/issues/7379#issuecomment-864387562)

---

maybe `--function-local`?

faho:
>maybe --function-local?

"Does this make my function a local function?"

Maybe "--function-scoped", but that's a bit asymmetrical to what's already there.

Also neither option really explains what happens when there *isn't* a function - `set --function` is global in that case.

faho:
So, looking over #565, there's a fair amount of talk over what `set -f` should do at the top level. Some are even arguing for making it error out.

I don't agree with that. I want `set -f` to be the *default* you reach for when you don't specifically need something else, and having it error out makes for more thinking about when it can be used.

Now, this PR currently uses the global scope at the top level, and that's probably not great - now suddenly the variable is available *everywhere*. Instead I'm going to make it the top level local scope.

That also makes explaining what this does easy:

> set --function sets a variable in the function's top local scope. If used outside of a function, it sets it in the top local scope.

(this could be made *shorter* by explaining that it's "the top-most accessible local scope" or some such, but that is a very dense explanation)

Of course that means it's not entirely the same as the "undefined scope" anymore, but tbh that's not great anyway.

faho:
Okay, I have now made `set --function` use the top local scope if used outside of a function.

Note that this now means functions with `--no-scope-shadowing` can set local variables:

```fish
> function set-my-thing --no-scope-shadowing
     set -f $argv 
   end
> function foo
      set -S bar
      set-my-thing bar nana
      set -S bar
   end
$bar: set in local scope, unexported, with 1 elements
$bar[1]: |nana|
```

(to be clear that's the `-f` scope from `foo`! Even with `begin; end` around `set-my-thing` it's accessible)

I'm not entirely sure if I like that or not. They could already *read* local variables (that's the point of it!), and they could always set *global* variables, but now they can set local variables as well. I think `--no-scope-shadowing` is a sufficiently expert level thing that it might be okay, and possibly even useful? Maybe? Or is this the most disgusting thing since the discovery of mushrooms?

faho:
Alright, I added the docs, which kinda shows how convoluted all of this is.

We now have:

- Universal
- Global
- Local (which should really be "block-scoped")
- Function-local (which is the same as block-scoped outside of a block!)
- Undefined scope (which is function-local inside of functions, global outside)

This should really be simplified. The classic "local" is probably mostly obsolete, unless you really need a variable only valid in a block, which is needed for:

```fish
begin
    set -lx PAGER cat
    # do things that use $PAGER
end
```

So removing it, or changing the default meaning of `-l` is probably out of the question?

We also can't really make the undefined scope local instead, because millions of config.fish have cruft like

```fish
set -x PAGER cat
```

(alternative: Make exported variables global by default?)

Merging universal and global scope sounds tempting, but what if you e.g. want to try out a new value for a universal variable?

krobelus:
Yeah this is getting fairly complex. I wonder how we can provide a simpler UI with fewer scopes.

My main gripe with (top-level) local scope is that local variables are not inherited by functions by default.

```fish
function f
    echo "not visible inside function:" $foo
end
set --function foo bar # same with set --local
f
```

I'm not sure if not inheriting locals is really useful, especially since as of now, function authors should declare their own variables with `--local`/`--function`. It is occasionally a problem when writing completion scripts that want to use private local variables in functions.

If we could make `set foo bar` inside functions never set globals/universals that would be nice..

> Merging universal and global scope sounds tempting, but what if you e.g. want to try out a new value for a universal variable?

Right, that wouldn't really work, we can't use locals for trying out values.

faho:
>My main gripe with (top-level) local scope is that local variables are not inherited by functions by default.

Oh I *heavily* disagree with this!

Not passing local variables everywhere is good because it encapsulates things. It leaves knowledge of these variables where it is, instead of spreading it all over the place.

Your example here is a toy. In a real situation often the function would be defined elsewhere, even in a different file. So having random variables defined there would be quite confusing.

And if you *do* need to pass a local variable, we have an option for that: Exporting them:

```fish
function f
    echo "not visible inside function:" $foo
end
set --function --export foo bar # same with set --local
f
```

The function will receive a local copy of the variable.

doc_src/language.rst
```diff
@@ -875,14 +875,15 @@ There are three kinds of variables in fish: universal, global and local variable
 - Universal variables are shared between all fish sessions a user is running on one computer.
 - Global variables are specific to the current fish session, and will never be erased unless explicitly requested by using ``set -e``.
 - Local variables are specific to the current fish session, and associated with a specific block of commands, and automatically erased when a specific block goes out of scope. A block of commands is a series of commands that begins with one of the commands ``for``, ``while`` , ``if``, ``function``, ``begin`` or ``switch``, and ends with the command ``end``.
+- Function-local variables are the same as local variables outside of one of these blocks. They go out of scope when the current function ends.
```
ridiculousfish, line 878: I suggest moving this up above `--local`, since now we want to de-emphasize local variables. Also we should not use "function-local" since that risks some confusion with local variables. I think "function variables" is fine.

My attempt:

> Function variables are scoped to the body of the currently executing function, and are erased when the function completes.
>
> Local variables are scoped to the body of the current innermost `for`, `while`, `if`, `begin`, `switch`, or `function` command, and are erased when that command completes. If the command is `function`, then it is a function variable.

doc_src/cmds/set.rst
```diff
@@ -35,6 +35,8 @@ The following options control variable scope:
 
 - ``-l`` or ``--local`` forces the specified shell variable to be given a scope that is local to the current block, even if a variable with the given name exists and is non-local
 
+- ``-f`` or ``--function`` makes the variable local to the current *function*. This is the same as defining it with ``--local`` outside of a block like ``begin`` or ``for``. When used outside of a function, it uses the top-level local scope instead.
```
ridiculousfish, line 38: I suggest not defining function-scoped variables in terms of local variables; instead we should assign the complexity to `--local`, which we want to de-emphasize. That is, move the `--function` description above `--local` and don't use "local" in its description since that could be confusing.

My attempt:

> -f or --function scopes the variable to the currently executing function, or the top-level scope if no function is executing. The variable is erased when the function completes.
>
> -l or --local scopes the variable to the currently executing block. The variable is erased when the block completes. If there is no currently executing block, this is the same as `--function`.

doc_src/language.rst
```diff
@@ -875,14 +875,15 @@ There are three kinds of variables in fish: universal, global and local variable
...
 
-Variables can be explicitly set to be universal with the ``-U`` or ``--universal`` switch, global with the ``-g`` or ``--global`` switch, or local with the ``-l`` or ``--local`` switch.  The scoping rules when creating or updating a variable are:
+Variables can be explicitly set to be universal with the ``-U`` or ``--universal`` switch, global with ``-g`` or ``--global``, local to the current block with ``-l`` or ``--local``, or local to the current function with ``-f`` or ``--function``.  The scoping rules when creating or updating a variable are:
```
ridiculousfish, line 880: Again suggest talking about function variables before local variable.

doc_src/language.rst
```diff
@@ -875,14 +875,15 @@ There are three kinds of variables in fish: universal, global and local variable
...
 
 - When a scope is explicitly given, it will be used. If a variable of the same name exists in a different scope, that variable will not be changed.
 
 - When no scope is given, but a variable of that name exists, the variable of the smallest scope will be modified. The scope will not be changed.
 
-- As a special case, when no scope is given and no variable has been defined the variable will belong to the scope of the currently executing *function*. This is different from the ``--local`` flag, which would make the variable local to the current *block*.
+- As a special case, when no scope is given and no variable has been defined the variable will belong to the scope of the currently executing *function*. This is different from the ``--local`` flag, which would make the variable local to the current *block*. Outside of a function, it is global, unlike with explicit function-scope.
```
ridiculousfish, line 886: I think this can now be simplified:

> If no scope is given and the variable is not yet defined, the variable will default to function scope if inside a function, or global scope if no function is executing.

ridiculousfish (approved):
This looks excellent to me. I agree with having top-level `set -f` set in the top-most local scope, not a global. I think this actually simplifies scoping, since users can now mostly ignore `set -l`.

Conceptually we should not think of "function-local" but just "function variables" to avoid confusion. A local variable is local to the innermost block; if that block is a function it is a function variable. Local variables become the "advanced" case.

One wrinkle is that `set --show` will still call function variables "local" but I think that's not a big deal.

doc_src/cmds/set.rst
faho: That's a brilliant idea, thanks!

faho:
Alright, merged.

PR #8148, opened 2021-07-17
Let "return" exit a script

faho:
Currently, if a "return" is given outside of a function, we'd just
throw an error.

That always struck me as a bit weird, given that scripts can also
return a value.

So simply let "return" outside also exit the script, kinda like "exit"
does.

However, unlike "exit" it doesn't quit an interactive shell - it seems
weird to have "return" do that as well. It sets $status, so it can be
used to quickly set that, in case you want to test something.

## TODOs:


ridiculousfish (approved):
Looks good to me, I like the idea. `INVALID_RETURN_ERR_MSG` can be removed now, it is unused.

PR #8151, opened 2021-07-19
Handle absolute path in CMAKE_INSTALL_DATADIR

### LINKED ISSUE #8150: fish assumes CMAKE_INSTALL_DATADIR must be relative
`CMAKE_INSTALL_DATADIR` can be both relative, in which case CMake will add `CMAKE_INSTALL_PREFIX` or absolute when prefix is not added. `fish` assumes only relative path can be passed and if absolute path is used `pkgconfig` file is broken with `prefix` added:
https://github.com/fish-shell/fish-shell/blob/b0dc72ee7811cc74d4b991f8d800a8189339e458/cmake/Install.cmake#L17
https://github.com/fish-shell/fish-shell/blob/b0dc72ee7811cc74d4b991f8d800a8189339e458/fish.pc.in#L1-L2

jpalus:
Cmake accepts both absolute and relative paths in CMAKE_INSTALL_DATADIR.
For the latter case CMAKE_INSTALL_PREFIX is being prepended
automatically. %{rel_datadir} is derived from CMAKE_INSTALL_DATADIR
which was assumed to be relative and otherwise causes issues in a .pc
file where prefix is being prepended unconditionally.

Make sure %{rel_datadir} is relative by calculating RELATIVE_PATH from
CMAKE_INSTALL_PREFIX to CMAKE_INSTALL_FULL_DATADIR which is known to be
absolute.

Fixes #8150

### DIFF
--- a/cmake/Install.cmake
+++ b/cmake/Install.cmake
@@ -14,8 +14,8 @@ set(bindir ${CMAKE_INSTALL_BINDIR})
 set(sysconfdir ${CMAKE_INSTALL_SYSCONFDIR})
 set(mandir ${CMAKE_INSTALL_MANDIR})
 
-set(rel_datadir ${CMAKE_INSTALL_DATADIR})
 set(datadir ${CMAKE_INSTALL_FULL_DATADIR})
+file(RELATIVE_PATH rel_datadir ${CMAKE_INSTALL_PREFIX} ${datadir})
 
 set(docdir ${CMAKE_INSTALL_DOCDIR})
 


zanchey (approved):
As an explainer for myself, we used to always use a full path but the relative directories were added in #6778 because some packages override `prefix` or `datadir` when running pkg-config. If you use a data directory that's outside the prefix, this will probably work (`rel_datadir` will become `../../some/path`), which _should_ work okay with most installers.

Issue #7474, opened 2020-11-11
"setpgid: No such process" on FreeBSD


raichoo:
Hi,

I'm getting this error from time to time when I issue a command:

```
$ pkg info | grep seat
[…]
Could not send own process 1895, 'grep' in job 1,  'command grep --color auto $argv' from group -1 to group 1894
setpgid: No such process
```

```
fish, version 3.1.2-1589-g36337fc45
FreeBSD lain 12.2-STABLE FreeBSD 12.2-STABLE r365792 GENERIC  amd64
alacritty 0.6.0-dev (78f6cc4)
```

Kind regards,
raichoo

faho:
If I understand you correctly:

- This doesn't happen in 3.1.2

- This only happens sometimes, even with the same commands.

In that case it seems like a race condition in setting up the processes (it's running and has already finished before we give it a pgroup). I'm not entirely sure which path FreeBSD ends up taking in this context (we do sometimes use posix_spawn).

If you could `git bisect` this, that'd be fantastic. Otherwise I'll have to fire up my FreeBSD vm and see if I can trigger it.

Also @ridiculousfish might know more about this.

raichoo:
I'm afraid the bug is too unreliable to trigger (yes I'm assuming a race condition), so doing a `bisect` wouldn't
really help. I don't think I've seen the same happening in 3.1.2.

mqudsi:
So what's happening is that the first command (`pkg info`) is exiting before the second command (`grep`) has finished fork+exec. 

@raichoo how many packages did you have installed at the time, i.e. is the output of `pkg info` extremely short? Normally, the first process would hang when writing to stdout after a short amount of output has been written, until the read end of the pipe is connected to the process being launched (grep). If `pkg info` generates no (or very little) output, it could theoretically exit before we've finished the post-fork handlers. This is all further complicated by the fact that `grep` is a function in the default fish configuration.

You can try to trigger the bug by using a non-builtin that doesn't generate any output as part of a toolchain, e.g. `/usr/bin/true | grep seat` - that might reproduce more readily for you?

mqudsi:
The first command in the job (here, `pkg`) is exiting before the second command (here, the command `grep` called by the *function* `grep`) finishes its post-fork work and calls `exec()`.

[...]

PR #8153, opened 2021-07-20
Work around `setpgid` error on older Apple platforms

evanmiller:
## Description

Expand the #7474 fix to Apple platforms. The issue affects older OS versions, including Mac OS X 10.4.11.

## TODOs:

### DIFF
--- a/src/postfork.cpp
+++ b/src/postfork.cpp
@@ -95,7 +95,7 @@ int execute_setpgid(pid_t pid, pid_t pgroup, bool is_parent) {
             FLOGF_SAFE(proc_pgroup, "setpgid(2) returned EPERM. Retrying");
             continue;
         }
-#ifdef __BSD__
+#if defined(__BSD__) || defined(__APPLE__)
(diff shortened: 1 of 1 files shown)


faho:
I can't help but notice that this isn't conditionalized on the macOS version, which presumably makes fish slower/worse in newer versions as well?

Support for versions *that old* (10.4 was released in 2005!) has to come behind support for newer versions, making newer ones worse isn't acceptable.

evanmiller:
I do not believe this change will make newer OSes any slower. Under normal conditions, with `setpgid` behaving as expected, the function will exit and return 0 earlier in the loop.

I would be happy to add more conditions. However, I've been unable to find any documentation on when the OS behavior changed, so this would require testing several OS versions that I don't currently have access to. A catch-all similar to the BSD condition seemed like the simplest solution without substantive drawbacks.

faho:
Okay, this is unlikely to make any big dents.

It can stay until anyone complains - 10.4 is *16 years old*. It would be allowed to drink beer here. Any later version is, by default, more important than it.

PR #8171, opened 2021-07-26
Refuse $status as a command

faho:
## Description

Another one from the "please no more" department:

Using $status as a command is *virtually guaranteed* to be an error. Nobody has commands called "0", and if they do they certainly don't want them called at semi-random just because some command failed or succeeded.

So, we forbid $status specifically, and print a special error.

It would also be possible to check if $status was used *after* the command wasn't found, but that means it can't be done statically anymore. And, like I said, using $status this way is unlikely to be on purpose, and if it was on purpose it wasn't necessary and you can skip the cutesy trick.

(the first commit is because the test was throwing path errors otherwise, I don't know since when they've been printed, but they are bogus regardless)

## TODOs:

### DIFF
--- a/src/fish.cpp
+++ b/src/fish.cpp
@@ -51,6 +51,7 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA
 #include "intern.h"
 #include "io.h"
 #include "parser.h"
+#include "parse_util.h"
 #include "path.h"
 #include "proc.h"
 #include "reader.h"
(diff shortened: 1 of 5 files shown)


faho:
Huh, surprising how github manages to capitalize the branch name into a title.

ridiculousfish:
This is nice, it will definitely catch some real errors. I suggest moving the check into `detect_errors_in_decorated_statement`, [around here](https://github.com/fish-shell/fish-shell/blob/3359e5d2e9bcbf19d1652636c8e448a6889302ae/src/parse_util.cpp#L1126). The advantage of putting it there is that the error will be caught when the script is parsed, instead of when it is executed.

src/parse_execution.cpp
```diff
@@ -800,6 +800,15 @@ end_execution_reason_t parse_execution_context_t::expand_command(
     wcstring unexp_cmd = get_source(statement.command);
     size_t pos_of_command_token = statement.command.range.start;
 
+    // $status specifically is invalid as a command,
+    // to avoid people trying `if $status`.
```
krobelus, line 804: maybe join these two lines

krobelus (commented):
Good idea.
Moving the check will also avoid false negatives like `set status_cmd echo 123; $status_cmd`

faho:
I'm sorry, is `detect_errors_in_decorated_statement` somehow not run for commands given via `fish -c`?

I've added this there, and if I do `echo 'if $status; true; end' | ./fish` I get the correct error, but if I do `./fish -c 'if $status; true; end'` I get "Unknown command: 0"?

faho:
>I'm sorry, is detect_errors_in_decorated_statement somehow not run for commands given via fish -c?

Okay, added a commit for that. There might be a cleaner way, but I don't know how or where.

faho:
Okay, merged. If someone finds a nicer way to detect_errors for the -c command, feel free to just do that. In the meantime this works.

PR #8172, opened 2021-07-26
Disable job control inside command substitutions

ridiculousfish:
This disables job control inside command substitutions, regardless of `status job-control`. A reason is that control-C doesn't work reliably if it gets directed to the command substitution:

    while true ; echo (sleep 5) ; end

With job control, `sleep` owns the tty so gets the SIGINT, and so the loop just continues on. The simplest way to fix this is to match other shells and not use job control in cmdsubs.

Related is #1362. I think I was just wrong in that issue and it's correct to not use job control.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,7 @@ Notable improvements and fixes
 Deprecations and removed features
 ---------------------------------
 - A new feature flag ``ampersand-nobg-in-token`` makes ``&`` only act as background operator if followed by a separator. In combination with ``qmark-noglob`` this allows to write some URLs without quoting or escaping (:issue:`7991`)
+- Command substitutions no longer respect job control, instead running inside fish's own process group. This more closely matches other shells, and improves control-C reliability inside a command substitution.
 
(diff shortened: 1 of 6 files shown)


faho:
Just to be sure: Does this still properly handle `set foo (fzf)` and `set bar (cat)` and friends?

krobelus:
This also seems more correct for `set bar (cat)` - `$bar` will no longer be set if canceled.

ridiculousfish:
> Just to be sure: Does this still properly handle set foo (fzf) and set bar (cat) and friends?

It's a good question; it works as far as I can tell. Other shells work this way so I think it is low risk.

mqudsi:
I definitely agree this is a step in the right direction for cleaning up fish scripts. It's unfortunate that we only really hear about things breaking after a release. It would be really cool if we had something like rust's crater runs that take code from the wild (cargo or in our case, github) and see if changes to the language/lib break things.

zanchey:
I rebased and merged this as 2ca66cff53.

PR #8180, opened 2021-07-28
Add completion for qmk

kfkonrad:
## Description

This PR adds completion for `qmk`. `qmk` is a CLI for managing, compiling and flashing firmware for custom keyboards. The code is based on the `az` completion since both programs use the same mechanism through argcomplete (see #8141  for reference)

## TODOs:

### DIFF
--- a/share/completions/qmk.fish
+++ b/share/completions/qmk.fish
@@ -1,12 +1,1 @@
-function __fish_qmk_complete
-  set -lx _ARGCOMPLETE 1
-  set -lx _ARGCOMPLETE_IFS \n
-  set -lx _ARGCOMPLETE_SUPPRESS_SPACE 1
-  set -lx _ARGCOMPLETE_SHELL fish
-  set -lx COMP_LINE (commandline -pc)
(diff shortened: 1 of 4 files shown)


share/completions/qmk.fish
```diff
@@ -0,0 +1,12 @@
+function __fish_qmk_complete
```
faho, line 1: Since we now have this twice, I would prefer if it was extracted into a function, named `__fish_argcomplete_complete` for example, that just takes the commandname.

So you'd make a file by that name in share/functions/, add the function, and then just call

```fish
__fish_argcomplete_complete qmk
```

here and

```fish
__fish_argcomplete_complete az
```

in the az completions. The function would call

```fish
$argv 8>&1 9>&2 2>/dev/null
```


That way, adding more of these is just a matter of adding one line into a new file.

faho (changes requested):
Please extract the function.

share/completions/qmk.fish
kfkonrad: done

faho:
Thanks, merged!

Issue #4012, opened 2017-05-04
add a `string width` subcommand


krader1961:
When issues like #3989 and #4011 are opened I frequently find myself hacking a fish binary to tell me what `fish_wcswidth()` returns for the non-ASCII characters. There really should be a standard mechanism for getting that information. It seems like a `string width` subcommand is what is needed. By default it reports the cumulative width of each string one per line. With the `-i` / `--individual` flag it would report the width of each individual character in each string. With the `-v` / `--verbose` flag it would print a symbolic representation of each char and its width.

This would allow you to type `string width \u009C` to learn that fish thinks that char has width -1 (i.e., the width is undefined).

The first implementation would not support ANSI X3.64 escape sequences. So it couldn't be used to determine the "visual" width of something like the output of `fish_prompt`. That's an enhancement we might add in the future.

faho:
So, for the easily confused (i.e. me), could you explain why this shouldn't be changed in or added to `string length` instead?

E.g. `string length é` right now prints 2 (because that "é" is the decomposed version, consisting of "e" and a composing accent), and I can't see how that is useful.

krader1961:
@faho: The reason is that `string length` counts the number of code points in a string. We can't change it without breaking backward compatibility. With respect to combining characters that might be okay but is risky. Too, what do we do about characters that do not have a defined width (which means we default to -1) or a width of 0.5 or 2?

This is basically a steaming pile of ugliness due to legacy issues. Note that Python 3.6 reports a length of 2 for your `é` character. Which is the behavior I think fish `string length` should have since that string consists of two unicode code points.

We could add one or more flags to `string length` to tell it to calculate character or visual widths. But it seems to me safer to introduce a new subcommand to make it crystal clear the new command is not reporting code points and vice-versa.

The primary purpose of this proposal is to make it easy to find out the length of individual code points. So that we don't have to manually hack a version of fish that reports that information when debugging problems #3989 and #4011. The secondary purpose is to potentially make it possible to calculate the visual width of a string that contains combining unicode chars and ANSI X3.64 sequences for more sophisticated prompts.

kballard:
This looks like a good idea. And I agree that it should be a separate command and that `string width` is a good candidate.

terrycloth:
[I've also been looking for a way to count the visible characters in a string](https://sourceforge.net/p/fish/mailman/message/36013313/), when I was directed to this thread.

faho:
[...]

PR #8182, opened 2021-07-29
Let string handle displayed width

faho:
## Description

This makes `string pad` pad to the effective width without escape sequences we recognize, and it adds an option `--visible` to `string length` that makes it show the *visible* length in terminal cells (or width, if you prefer).

This is, of course, necessarily imperfect. We're not the terminal, we don't know which escape codes it recognizes exactly. But we do handle the most common ones like colors and such.

This also changes in response to $TERM, $fish_emoji_width and $fish_ambiguous_width.

Importantly: `string length --visible` handles carriage return and line feed in what I hope is a sensible manner. Because counting a string like

```
abcdef\r123
```

as anything other than a width of 6 seems useless to me. The line feed handling may be surprising because it might print multiple lines of output in response to a single argument, but I also don't see how anything else would help.

It would also be possible to make `string length --visible` a separate command instead, but given how similar these operations are I'm not sure that helps. Or we could change the option to `--width`, but that doesn't include that it also removes escape sequences.

Fixes issue #7784.

## TODOs:


tests/checks/string.fish
```diff
@@ -87,6 +87,31 @@ string pad -c_ --width 5 longer-than-width-param x
...
+    # CHECK: 1
+    # CHECK: 2
+    set -l fish_emoji_width 1
+    string length --visible . 🐟
+    # CHECK: 1
+    # CHECK: 1
+end
+
+# Only the longest run between carriage returns is kept because the rest is overwritten.
+string length --visible (set_color reset)abcdef\rfooba(set_color red)raaa
```
IlanCosman, line 106: From what I can tell, Fish is trying to move away from `set_color reset`. Should this be a `set_color normal`?

IlanCosman:
Also: closes #4012 I believe.

tests/checks/string.fish
faho: Yeah, replaced.

faho:
Do we want `string pad` to require `--visible` as well?

faho:
Okay, this still needs adjustment in `pad`, for CR runs.

E.g.

```fish
string pad -w 4 ab\rfoo
```

has a maximum width of *3*, but we would pad from the beginning, so we influence a run with a width of *2*. So we need to add *2* spaces. If we padded to the right, we would need 1 space.

faho:
Alright, let's do the additional pad stuff later. It's probably overly completionist anyway.

IlanCosman:
> It's probably overly completionist anyway. Merged.

Good call 👍 Thanks for working on this Fabian 😄

PR #8183, opened 2021-07-30
Add zypper subcommands completion

Sam0523:
- rename functions to avoid confusion

## TODOs:

### DIFF
--- a/share/completions/zypper.fish
+++ b/share/completions/zypper.fish
@@ -1,6 +1,7 @@
 # completion for zypper
 
-set -g __fish_zypper_all_commands shell sh repos lr addrepo ar removerepo rr renamerepo nr modifyrepo mr refresh ref clean cc services ls addservice as modifyservice ms removeservice rs refresh-services refs install in remove rm verify ve source-install si install-new-recommends inr update up list-updates lu patch list-patches lp dist-upgrade dup patch-check pchk search se info if patch-info pattern-info product-info patches pch packages pa patterns pt products pd what-provides wp addlock al removelock rl locks ll cleanlocks cl locales lloc addlocale aloc removelocale rloc versioncmp vcmp targetos tos licenses download source-download needs-rebooting ps
+set -g __fish_zypper_all_commands (LC_ALL=POSIX zypper -q -h | sed -n '/[Cc]ommands:$/,$ s/^ \{6\}\([^ ,]\+\(,[ *][^ ,]\+\)*\).*/\1/p' | string split ', ')
+set -g __fish_zypper_subcommands (LC_ALL=POSIX zypper -q subcommand | string split ', ')
 set -g __fish_zypper_pkg_commands install in remove rm info if addlock al removelock rl source-install si update up download
 set -g __fish_zypper_repo_commands repos lr removerepo rr renamerepo nr modifyrepo mr refresh ref clean cc packages pa patches pch patterns pt products pd
 set -g __fish_zypper_file_commands install in addrepo ar
@@ -20,7 +21,7 @@ function __fish_zypper_cmd_in_array
     return 1
 end
 
-function __fish_zypper_no_subcommand
+function __fish_zypper_no_command
     not __fish_zypper_cmd_in_array $__fish_zypper_all_commands
 end
 
@@ -69,7 +70,7 @@ function __fish_zypper_print_packages
         set -e args[1]
     end
 
-    if __fish_zypper_is_subcommand_rm
+    if __fish_zypper_is_command_rm
         set idx /var/cache/zypp/solv/@System/solv.idx
     else if test -z "$repo"
         set idx /var/cache/zypp/solv/*/solv.idx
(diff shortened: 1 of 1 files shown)


share/completions/zypper.fish
```diff
@@ -1,6 +1,7 @@
 # completion for zypper
 
-set -g __fish_zypper_all_commands shell sh repos lr addrepo ar removerepo rr renamerepo nr modifyrepo mr refresh ref clean cc services ls addservice as modifyservice ms removeservice rs refresh-services refs install in remove rm verify ve source-install si install-new-recommends inr update up list-updates lu patch list-patches lp dist-upgrade dup patch-check pchk search se info if patch-info pattern-info product-info patches pch packages pa patterns pt products pd what-provides wp addlock al removelock rl locks ll cleanlocks cl locales lloc addlocale aloc removelocale rloc versioncmp vcmp targetos tos licenses download source-download needs-rebooting ps
+set -g __fish_zypper_all_commands (LC_ALL=POSIX zypper -q -h | sed -n '/[Cc]ommands:$/,$ s/^ \{6\}\([^ ,]\+\(,[ *][^ ,]\+\)*\).*/\1/p' | string split ', ')
```
faho, line 3: What's the output here like?

I would prefer if we could avoid

1. switching to POSIX locale for this
2. using `sed`

What commands are added by doing this?

share/completions/zypper.fish
```diff
@@ -20,7 +21,7 @@ function __fish_zypper_cmd_in_array
     return 1
 end
 
-function __fish_zypper_no_subcommand
+function __fish_zypper_no_command
```
faho, line 24: Please no. It's a subcommand *to zypper*, the function can be named `no_subcommand`.

share/completions/zypper.fish
Sam0523: > What's the output here like?

```

Usage:

    zypper [--GLOBAL-OPTIONS] <COMMAND> [--COMMAND-OPTIONS] [ARGUMENTS]
    zypper <SUBCOMMAND> [--COMMAND-OPTIONS] [ARGUMENTS]

Global Options:

    --help, -h              Help.
    --version, -V           Output the version number.
    --promptids             Output a list of zypper's user prompts.
    --config, -c <FILE>     Use specified config file instead of the default.
    --userdata <STRING>     User defined transaction id used in history and plugins.
    --quiet, -q             Suppress normal output, print only error messages.
    --verbose, -v           Increase verbosity.
    --color
    --no-color              Whether to use colors in output if tty supports it.
    --no-abbrev, -A         Do not abbreviate text in tables. Default: false
    --table-style, -s <INTEGER>
                            Table style (0-11).
    --non-interactive, -n   Do not ask anything, use default answers automatically. Default: false
    --non-interactive-include-reboot-patches
                            Do not treat patches as interactive, which have the rebootSuggested-flag
                            set. Default: false
    --xmlout, -x            Switch to XML output.
    --ignore-unknown, -i    Ignore unknown packages. Default: false
    --terse, -t             Terse output for machine consumption. Implies --no-abbrev and
                            --no-color.


    --reposd-dir, -D <DIR>  Use alternative repository definition file directory.
    --cache-dir, -C <DIR>   Use alternative directory for all caches.
    --raw-cache-dir <DIR>   Use alternative raw meta-data cache directory.
    --solv-cache-dir <DIR>  Use alternative solv file cache directory.
    --pkg-cache-dir <DIR>   Use alternative package cache directory.

  Repository Options

    --no-gpg-checks         Ignore GPG check failures and continue. Default: false
    --gpg-auto-import-keys  Automatically trust and import new repository signing keys.
    --plus-repo, -p <URI>   Use an additional repository.
    --plus-content <TAG>    Additionally use disabled repositories providing a specific keyword. Try
                            '--plus-content debug' to enable repos indicating to provide debug
                            packages.
    --disable-repositories  Do not read meta-data from repositories.
    --no-refresh            Do not refresh the repositories.
    --no-cd                 Ignore CD/DVD repositories.
    --no-remote             Ignore remote repositories.
    --releasever            Set the value of $releasever in all .repo files (default: distribution
                            version)

  Target Options

    --root, -R <DIR>        Operate on a different root directory.
    --installroot <DIR>     Operate on a different root directory, but share repositories with the
                            host.
    --disable-system-resolvables
                            Do not read installed packages.

Commands:

      help, ?               Print zypper help
      shell, sh             Accept multiple commands at once.

  Repository Management:

      repos, lr             List all defined repositories.
      addrepo, ar           Add a new repository.
      removerepo, rr        Remove specified repository.
      renamerepo, nr        Rename specified repository.
      modifyrepo, mr        Modify specified repository.
      refresh, ref          Refresh all repositories.
      clean, cc             Clean local caches.

  Service Management:

      services, ls          List all defined services.
      addservice, as        Add a new service.
      modifyservice, ms     Modify specified service.
      removeservice, rs     Remove specified service.
      refresh-services, refs
                            Refresh all services.

  Software Management:

      install, in           Install packages.
      remove, rm            Remove packages.
      verify, ve            Verify integrity of package dependencies.
      source-install, si    Install source packages and their build dependencies.
      install-new-recommends, inr
                            Install newly added packages recommended by installed packages.

  Update Management:

      update, up            Update installed packages with newer versions.
      list-updates, lu      List available updates.
      patch                 Install needed patches.
      list-patches, lp      List available patches.
      dist-upgrade, dup     Perform a distribution upgrade.
      patch-check, pchk     Check for patches.

  Querying:

      search, se            Search for packages matching a pattern.
      info, if              Show full information for specified packages.
      patch-info            Show full information for specified patches.
      pattern-info          Show full information for specified patterns.
      product-info          Show full information for specified products.
      patches, pch          List all available patches.
      packages, pa          List all available packages.
      patterns, pt          List all available patterns.
      products, pd          List all available products.
      what-provides, wp     List packages providing specified capability.

  Package Locks:

      addlock, al           Add a package lock.
      removelock, rl        Remove a package lock.
      locks, ll             List current package locks.
      cleanlocks, cl        Remove useless locks.

  Locale Management:

      locales, lloc         List requested locales (languages codes).
      addlocale, aloc       Add locale(s) to requested locales.
      removelocale, rloc    Remove locale(s) from requested locales.

  Other Commands:

      versioncmp, vcmp      Compare two version strings.
      targetos, tos         Print the target operating system ID string.
      licenses              Print report about licenses and EULAs of installed packages.
      download              Download rpms specified on the commandline to a local directory.
      source-download       Download source rpms for all installed packages to a local directory.
      needs-rebooting       Check if the reboot-needed flag was set.
      ps                    List running processes which might still use files and libraries deleted
                            by recent upgrades.
      purge-kernels         Remove old kernels.

  Subcommands:

      subcommand            Lists available subcommands.
      appstream-cache       <No manual entry for zypper-appstream-cache>
      autoremove            <No manual entry for zypper-autoremove>
      lifecycle             products and packages lifecycle information
      log                   Zypper logfile reader

Type 'zypper help <COMMAND>' to get command-specific help.

```
The subcommands section may differs to each user.
> 
> I would prefer if we could avoid
> 
>     1. switching to POSIX locale for this
> 
>     2. using `sed`
> 
> 
> What commands are added by doing this?

Actually nothing but subcommands. I copy-pasted this from bash completion of zypper, expecting that maybe we will need less change if a new version of zypper add new subcommands. After consideration I found this idea silly... we still need to add argument completions if that happens.

faho: Okay, so what *sub*commands are added or removed by doing this?

share/completions/zypper.fish
Sam0523: zypper has a feature named "subcommand" that allow standalone executables (user's script, for example) to be run as zypper subcommands, which this commit tries to complete. Isn't that confusing with same names?

Sam0523: FYI:
```
$ zypper subcommand
<subcommand> [--command-options] [arguments]

Zypper subcommands are standalone executables that live in the
zypper_execdir ('/usr/lib/zypper/commands').

For subcommands zypper provides a wrapper that knows where the
subcommands live, and runs them by passing command-line arguments
to them.

If a subcommand is not found in the zypper_execdir, the wrapper
will look in the rest of your $PATH for it. Thus, it's possible
to write local zypper extensions that don't live in system space.

Using zypper global-options together with subcommands, as well as
executing subcommands in 'zypper shell' is currently not supported.

Available zypper subcommands in '/usr/lib/zypper/commands'

      appstream-cache       <No manual entry for zypper-appstream-cache>
      lifecycle             products and packages lifecycle information


Zypper subcommands available from elsewhere on your $PATH

      autoremove            <No manual entry for zypper-autoremove>
      log                   Zypper logfile reader


Type 'zypper help <subcommand>' to get subcommand-specific help if available.
```

faho: Okay, but in *fish*, a "subcommand" is something like the "log" in `git log` or the "install" in `zypper install`.

Which is reflected in helper functions like `__fish_seen_subcommand_from`.

Sam0523: OK, I get it. So any ideas about naming here? Or shall I keep the confusing names?

faho: Just keep the old names, please.

faho:
Merged, thanks!

PR #8192, opened 2021-07-31
Updated ping completions

exploide:
## Description

Wanted to update ping completions because `-4` and `-6` was missing. Found out that there is also the inetutils ping implementation. And busybox.

- support for ping from iputils (version 20210202)
- support for ping from inetutils (version 2.1)
- support for ping from busybox (version 1.33.1)
- support for ping from FreeBSD and macOS (by @juntuu)

## TODOs:

### DIFF
--- a/share/completions/ping.fish
+++ b/share/completions/ping.fish
@@ -1,29 +1,131 @@
-complete -c ping -a "(__fish_print_hostnames)" -x
-complete -c ping -s a -d "Audible ping"
-complete -c ping -s A -d "Adaptive ping"
-complete -c ping -s b -d "Allow pinging a broadcast address"
-complete -c ping -s B -d "Do not allow ping to change source address of probes"
-complete -c ping -s c -d "Stop after specified number of ECHO_REQUEST packets" -x
-complete -c ping -s d -d "Set the SO_DEBUG option on the socket being used"
-complete -c ping -s F -d "Allocate and set 20 bit flow label on ECHO_REQUEST packets" -x
-complete -c ping -s f -d "Flood ping"
-complete -c ping -s i -d "Wait specified interval of seconds between sending each packet" -x
-complete -c ping -s I -d "Set source address to specified interface address" -x -a "(__fish_print_interfaces; __fish_print_addresses)"
-complete -c ping -s l -d "Send the specified number of packets without waiting for reply" -x
-complete -c ping -s L -d "Suppress loopback of multicast packets"
-complete -c ping -s n -d "Numeric output only"
-complete -c ping -s p -d "Pad packet with empty bytes" -x
-complete -c ping -s Q -d "Set Quality of Service -related bits in ICMP datagrams" -x
-complete -c ping -s q -d "Quiet mode"
-complete -c ping -s R -d "Record route"
-complete -c ping -s r -d "Bypass the normal routing tables"
-complete -c ping -s s -d "Specifies the number of data bytes to be sent" -x
-complete -c ping -s S -d "Set socket buffer size" -x
-complete -c ping -s t -d "Set the IP Time to Live" -x
-complete -c ping -s T -d "Set special IP timestamp options" -x
-complete -c ping -s M -d "Select Path MTU Discovery strategy" -x -a "do want dont"
-complete -c ping -s U -d "Print full user-to-user latency"
-complete -c ping -s v -d "Verbose mode"
-complete -c ping -s V -d "Display version and exit"
-complete -c ping -s w -d "Specify a timeout, in seconds, before ping exits" -x
-complete -c ping -s W -d "Time to wait for a response, in seconds" -x
+complete -c ping -xa "(__fish_print_hostnames)"
+
+set -l ping_version (ping -V 2>&1 | string collect)
+
+switch $ping_version
+    case '*iputils*'
+        # completions for ping from iputils
+        complete -c ping -s 4 -d "Use IPv4"
+        complete -c ping -s 6 -d "Use IPv6"
+        complete -c ping -s a -d "Audible ping"
+        complete -c ping -s A -d "Adaptive ping"
+        complete -c ping -s b -d "Allow pinging a broadcast address (IPv4 only)"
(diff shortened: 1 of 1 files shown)


juntuu:
As far as I can see, this would remove flag completions on macos.

Although, the current completions seem to have quite a few incorrect options for macos version of ping, it does have some correct ones as well.

That said, having no completions might be better, than having incorrect ones mixed in. And correct completions for macos could always be added later.

The same might be true for other BSD based systems.

```console
$ uname -mrs
Darwin 20.5.0 arm64
$ ping -V
ping: invalid option -- V
usage: ping [-AaDdfnoQqRrv] [-c count] [-G sweepmaxsize]
            [-g sweepminsize] [-h sweepincrsize] [-i wait]
            [-l preload] [-M mask | time] [-m ttl] [-p pattern]
            [-S src_addr] [-s packetsize] [-t timeout][-W waittime]
            [-z tos] host
       ping [-AaDdfLnoQqRrv] [-c count] [-I iface] [-i wait]
            [-l preload] [-M mask | time] [-m ttl] [-p pattern] [-S src_addr]
            [-s packetsize] [-T ttl] [-t timeout] [-W waittime]
            [-z tos] mcast-group
Apple specific options (to be specified before mcast-group or host like all options)
            -b boundif           # bind the socket to the interface
            -k traffic_class     # set traffic class socket option
            -K net_service_type  # set traffic class socket options
            -apple-connect       # call connect(2) in the socket
            -apple-time          # display current time
```

share/completions/ping.fish
```diff
@@ -1,29 +1,81 @@
...
-complete -c ping -s T -d "Set special IP timestamp options" -x
-complete -c ping -s M -d "Select Path MTU Discovery strategy" -x -a "do want dont"
-complete -c ping -s U -d "Print full user-to-user latency"
-complete -c ping -s v -d "Verbose mode"
-complete -c ping -s V -d "Display version and exit"
-complete -c ping -s w -d "Specify a timeout, in seconds, before ping exits" -x
-complete -c ping -s W -d "Time to wait for a response, in seconds" -x
+complete -c ping -xa "(__fish_print_hostnames)"
+
+if ping -V &| grep -q iputils
```
faho, line 3: This starts `ping` once per branch.

I recommend storing the output in a variable and `switch`ing over it:

```fish
set -l ping_version (ping -V 2>/dev/null)
switch $ping_version
    case '*iputils*'
        # do iputils stuff
    case '*inetutils*'
       # do inetutils stuff
    case '*'
       # unknown version
end
```

exploide: Yes, that's better. I incorporated this in the last force-push. But I needed to redirect stderr to stdout instead of discarding it because opposed to iputils and inetutils, BusyBox prints to stderr...

exploide:
> As far as I can see, this would remove flag completions on macos.

Oh no, I hoped they just use inetutils. Why does everyone implement their own ping...

We could add a case for MacOS. There is the string "Apple" we could match on. Do you have a complete help output or something like this?

If you like, feel free to write the complete lines for this ping. If not, please provide me with sufficient docs.

juntuu:
Ended up writing ones for mac https://gist.github.com/juntuu/ea7b53750d33a96320273d36cff8fb3b

Looking into this a bit more, the apple man page for ping is same as [FreeBSD](https://www.unix.com/man-page/freebsd/8/ping/), with just the following  Apple specific additions:
```
-b boundif
     Bind the socket to interface boundif for sending.  This option is
     an Apple addition.

-C      Prohibit the socket from using the cellular network interface.
     This option is an Apple addition.

-k trafficclass
     Specifies the traffic class to use for sending ICMP packets.  The
     supported traffic classes are BK_SYS, BK, BE, RD, OAM, AV, RV,
     VI, VO and CTL.  By default ping uses the control traffic class
     (CTL).  This option is an Apple addition.

-K netservicetype
     Specifies the network service type to use for sending ICMP pack-
     ets.  The supported network service type are BK_SYS, BK, BE, RV,
     AV, RD, OAM, VI, SIG and VO.  Note this overrides the default
     traffic class (-k can still be specified after -K to use both).
     This option is an Apple addition.

--apple-connect
     Connects the socket to the destination address.  This option is
     an Apple addition.

--apple-time
     Prints the time a packet was received.  This option is an Apple
     addition.
```

I also checked OpenBDS and NetBSD's man pages, and both seem to have their own incompatible options...

Quoting https://man.netbsd.org/ping.8 
> The ping program has evolved differently under different operating sys-
     tems, and in some cases the same flag performs a different function under
     different operating systems.  The -t flag conflicts with FreeBSD.  The
     -a, -c, -I, -i, -l, -P, -p, -s, and -t flags conflict with Solaris.

I wonder, if it would be possible to fall back to using the auto generated completions if the version can not be detected?

exploide:
Great, thanks. Unfortunately, I'm not sure how to detect FreeBSD. `ping -V` outputs nothing useful there.
The workaround I could image: Having a default `case '*'` and within branch on some other conditions, like the system type (from `uname` or wherever?).
That completion grows quickly in complexity. Don't know if this is desired. Any advice @faho ?

juntuu:
> I wonder, if it would be possible to fall back to using the auto generated completions if the version can not be detected?

This could cut down the complexity of trying to match every possible version. I'm not sure how & if it would work, but it could provide at least some completions, given that the system has the man-pages.

A naïve solution would be to check if generated completion file exists, and sourcing that. But I'm not too familiar with the completion system to know if that would work correctly.

kidonng:
> Unfortunately, I'm not sure how to detect FreeBSD.

Maybe `test (uname) = FreeBSD`

exploide:
I incorporated your changes @juntuu, please give it a try on macOS and see if it works.

As far as I know from other completions, falling back to the manpage generated completions is not common.

So that's it so far until someone proposes further changes.

juntuu:
> give it a try on macOS and see if it works

Seems to work just fine

faho:
Yeah, that looks good. Merged, thanks!

PR #8202, opened 2021-08-03
Add `--function` to `set` completion

kidonng:
## Description

Seems like an oversight in #8145.

## TODOs:


share/completions/set.fish
```diff
@@ -71,6 +71,7 @@ end
 complete -c set -n "fish_is_nth_token 1" -s e -l erase -d "Erase variable"
 complete -c set -n "fish_is_nth_token 1" -s x -l export -d "Export variable to subprocess"
 complete -c set -n "fish_is_nth_token 1" -s u -l unexport -d "Do not export variable to subprocess"
+complete -c set -n "fish_is_nth_token 1" -s f -l function -d "Make variable scope function"
```
faho, line 74: "Make variable function-scoped"

kidonng: Agree, I'm just following the existing examples. May as well change other descriptions to use same format.

faho: The other examples are adjectives. A variable "is local", a variable "is global".

A variable is not "function".

So the others can stay.

kidonng: Seems like we should not just say "make variable scope universal" so there will be inconsistency between descriptions. That's fine, I reverted other changes.

faho:
Merged, thanks!

PR #8204, opened 2021-08-04
Always use LC_NUMERIC=C internally

faho:
In most cases, like math, we want C-semantics for floating point
numbers. In particular "." needs to be the decimal separator.

Instead, we pay the price in printf, which is currently the sole place
to output in locale-specific numbers and attempt to read them and
C-style ones.

This is *very likely* to contain GNUisms, not work on BSD, yadda yadda.

This is an RFC, intended to get people to check and to think about.

The results are... interesting.

All tests done on my Archlinux system with LANG=de_DE.UTF-8:

- math is about 10% faster with the new system - this is understandable, since it no longer needs to switch locale
- *external_cmds.fish* is about 1.4x faster than the old system! From 1.5s to 1.05s!

I have absolutely no idea why, but the same speedup is reproducible by just setting LC_NUMERIC=C, so it seems the multibyte locales are just slower at doing numbers?

cc @ridiculousfish


faho:
As a sidenote: `printf`s output being locale-dependent is *super awkward*, because that means

```fish
math (printf '%f' 5.2) + 2
```

might or might not work, depending on the locale. But since math cannot (and definitely SHOULD NOT) try to read numbers in the regional way (because it uses "," as argument separator) the only possible change is to make printf print numbers with ".", always. Which would be a break from e.g. bash (and probably POSIX), but not one I'm opposed to.

ridiculousfish:
I like the change. The switch to `uselocale` instead of `setlocale` is quite nice for thread safety reasons. This works fine on my Mac, though there's no apparent performance difference. We'll see which if any platforms are broken.

faho:
>This works fine on my Mac, though there's no apparent performance difference

Not even faster `math`? This should be as fast as fba86fb82105e65b0b8bef8f1e1d258cfc2f08dd, at least.

src/wutil.cpp
```diff
@@ -574,6 +574,22 @@ locale_t fish_c_locale() {
...
+
+void fish_invalidate_numeric_locale() {
+    fish_numeric_locale_is_valid = false;
+}
+
+locale_t fish_numeric_locale() {
+    // The current locale, except LC_NUMERIC isn't forced to C.
+    static locale_t loc;
+    if (!fish_numeric_locale_is_valid) {
+        auto cur = duplocale(LC_GLOBAL_LOCALE);
```
faho, line 587: I'm not sure if we need to free the old loc here, if it's set?

src/wutil.cpp
```diff
@@ -711,14 +727,14 @@ unsigned long long fish_wcstoull(const wchar_t *str, const wchar_t **endptr, int
...
 double fish_wcstod(const wchar_t *str, wchar_t **endptr) {
+    // We can ignore the locale because we use LC_NUMERIC=C!
     // The "fast path." If we're all ASCII and we fit inline, use strtod().
     char narrow[128];
     size_t len = std::wcslen(str);
     size_t len_plus_0 = 1 + len;
-    auto is_digit = [](wchar_t c) { return '0' <= c && c <= '9'; };
-    if (len_plus_0 <= sizeof narrow && std::all_of(str, str + len, is_digit)) {
+    auto is_ascii = [](wchar_t c) { return 0 <= c && c <= 127; };
+    if (len_plus_0 <= sizeof narrow && std::all_of(str, str + len, is_ascii)) {
```
faho, line 736: Note: I am not sure if the fast path actually helps anymore if we always use LC_NUMERIC=C.

I would appreciate benchmarks on this.

faho:
Ah, yes, forgot that NetBSD does not have uselocale.

So the ideal of thread-safe locale switching goes out of the window on that system. The alternatives are:

- setlocale
- printf just does "." always.

I'm tempted to just do the latter, first on NetBSD, and then *everywhere*, because that makes more sense anyway.

faho:
>external_cmds.fish is about 1.4x faster than the old system! From 1.5s to 1.05s!

Okay, on WSL1 I get 1.8x, but only if I effectively set LC_NUMERIC to de_DE.UTF-8 - en_US.UTF-8 is the same as C. *Something* here is causing massive slowdowns in either that locale or comma-using ones in general.

faho:
Okay, so this works on linux (with glibc), macOS, FreeBSD. It works with `setlocale` on NetBSD (because they believe that "you should use printf_l" and refuse to implement uselocale - yes, really). And it "works" on OpenBSD with the wonderful "CAVEAT":

>On systems other than OpenBSD, the LC_NUMERIC locale(1) category can cause parsing failures; see CAVEATS in setlocale(3) for details.

Yeah, sure, OpenBSD. Removing a feature is totally the same as *unbreaking* things. (LC_NUMERIC is basically unused there, which means it'll never read localized numbers anywhere. Whatever)


For the record: I hate all of this.

faho:
Note: This doesn't need to be merged for 3.4.0.

I would prefer more testing and benchmarks (because this is such a portability mess!), so I'll delay merging it.

faho:
Okay, I am merging this now, because:

1. Nobody seems to have found any issues with this
2. I have found a system where comma-using locales are pathologically slow - by a factor of 20x or so?

ridiculousfish:
Oh wow. That must have been very annoying to track down. Yes  `__GLIBC_MINOR__` is what I should have been using all along, I'll add that.

ridiculousfish:
Looks like `__GLIBC_PREREQ` is the thing to use. f0d4fd85b116a30319c1b42f1f79c33ebaa69441

PR #8208, opened 2021-08-05
prompt_pwd: Allow keeping some directory components at full length

faho:
## Description

This allows specifying to `prompt_pwd` that you'd like to keep the last N components at full length instead of shortening them.

This can be done either by setting $fish_prompt_pwd_full_dirs to the number, or passing the new `--full-length-dirs`/`-D` option.

The rationale for that is that the earlier path components are typically slower changing and known - it's often something like ~/dev/. So by keeping these short, more space can be kept for the components at the end that are more important and faster changing.

Similarly, instead of setting `$fish_prompt_pwd_dir_length`, this adds a `--dir-length`/`-d` option for symmetry, and allows passing the directory so that we can better test this. Options are probably the much nicer interface here anyway.

(we have had someone open a PR about this years ago, which I rejected. I'm afraid I was wrong, sorry!)

## TODOs:

### DIFF
--- a/doc_src/cmds/prompt_pwd.rst
+++ b/doc_src/cmds/prompt_pwd.rst
@@ -17,7 +17,16 @@ Description
 
 ``prompt_pwd`` is a function to print the current working directory in a way suitable for prompts. It will replace the home directory with "~" and shorten every path component but the last to a default of one character.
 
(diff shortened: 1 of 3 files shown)


ridiculousfish:
Seems reasonable to me. Nice test!

PR #8211, opened 2021-08-06
Fix dirs returns false when $dirstack is empty

andrew-schulman:
## Description

In fish 3.3.1, when `$dirstack` is empty, `dirs` returns false. This is undesirable, because no error occurred. 

It happens because the `string join` at the end doesn't join anything, so it returns false. The solution is just to add a `return 0` at the end of the function.

## TODOs:

### DIFF
--- a/share/functions/dirs.fish
+++ b/share/functions/dirs.fish
@@ -16,4 +16,5 @@ function dirs --description 'Print directory stack'
 
     # Replace $HOME with ~.
     string replace -r '^'"$HOME"'($|/)' '~$1' -- $PWD $dirstack | string join " "
+    return 0
 end


kidonng:
I would say current behavior seems right. There is a slight difference between *printing nothing* and *matching nothing*, and adding a `return 0` unconditionally will make them indistinguishable.

krobelus:
> There is a slight difference between printing nothing and matching nothing, and adding a return 0 unconditionally will make them indistinguishable.

I don't understand, what's the difference here? I don't think "matching" applies here because we just print the dirstack entries one by one.

---

I think we should probably change `string join` here. Returning false on `string join ' ' x` is weird. Returning false on `string join ' '` seems okay still, because in that case the output is empty.

andrew-schulman:
`dirs` always prints at least one result, namely the current directory. So the return value is different depending on whether it prints only one (false) or more than one (true), which doesn't seem useful. It should just return true.

faho:
Yeah, makes sense. Merged!

Issue #4033, opened 2017-05-16
`set -q` documentation wrong?


dpc:
https://fishshell.com/docs/current/commands.html#set  :

```
-q or --query test if the specified variable names are defined. Does not output anything, but the builtins exit status is the number of variables specified that were not defined.
```

**not** defined

But when I run:

```
> fish --version
fish, version 2.5.0
> set -q $HOME; echo $status
1
> set -q $HOME $DISPLAY; echo $status
2
> set -q $fooodfsdfa $asdfasdfddd; echo $status
0
```

faho:
No, the documentation is correct. The issue here is that you are expanding the variables, which means `set -q` will check if the _value_ of the variable is itself the name of a defined variable name.

E.g. `set -q $HOME`, will check if "/home/dpc/" (or similar) is a defined variable, which it isn't (it's not even a valid variable name).

`set -q $foo` will expand to `set -q` without other arguments if foo is undefined. It would probably be better to return a false value in that case, but that would necessarily break the idea that the return status is equal to the number of undefined variables - here, there are no undefined variables given.

The correct call would be `set -q foo` or `set -q HOME` - pass just the name, don't expand the variable.

Hope that helps!

dpc:
Ooooooohhh... . Thanks!

Issue #8131, opened 2021-07-10
Warn about deprecated features (qmark-noglob, & in tokens)


krobelus:
Some ideas in https://github.com/fish-shell/fish-shell/pull/7991#issuecomment-876482808

uberjay:
This is almost certainly not the right place to leave this comment, but I'm not sure where is:

Somehow I just encountered the `qmark-noglob` "feature", after upgrading fish. I spent longer than I'd care to admit getting confused trying to move a set of files from one directory to another. After seeing that `~/Downloads/20*-*.pdf` matched too many files, I tried to use `~/Downloads/20??-??.pdf`. Eventually I executed the `mv` command using `bash` and it worked... leading me to eventually run across the desire to completely remove the questionmark glob character in fish.

I guess other people have different workflows! I use `?` as a glob character quite frequently and it's kind of blowing my mind that it's slated to complete removal. Is there a replacement planned? (Other than old version of fish or different shell entirely?)

hholst80:
`?` and `&` are very often used as part of URL requests. 

If you wish to disable the glob for these you can use this command:

```
# set -U fish_features qmark-noglob
```

```
# status features
stderr-nocaret          on  3.0 ^ no longer redirects stderr (historical, can no longer be changed)
qmark-noglob            on  3.0 ? no longer globs
regex-easyesc           on  3.1 string replace -r needs fewer \'s
ampersand-nobg-in-token on  3.4 & only backgrounds if followed by a separator
```

norbertkeri:
I just saw in the release notes of 4.0b1 that this will become a read only flag, and it will be "impossible" to turn off

>The flag will eventually be made read-only, making it impossible to turn off.

Could someone explain the reason for making this "impossible" to turn off? And maybe not do that?

faho:
> Could someone explain the reason for making this "impossible" to turn off? And maybe not do that?

It'll be impossible to turn off because we don't want to have multiple dialects of the language. We'll do that for a transitional phase, but we don't want to have the same effect as bash with "nullglob" and "errexit" and "extglob" and so on, where what a line means depends on global state.

It's turned off because it makes `?` a special character that needs to be quoted and escaped and in particular makes URLs annoying to use. It also doesn't pull its weight because it is often not actually necessary and can be misused and lead to undermatching - I saw archlinux match `python3.?` in a script once, which broke once python3.10 came out.

Fish's globs are very simple, and if you need something more specific you can always use `find`.

[...]

PR #8214, opened 2021-08-09
`set -q`: Return 255 if no variable name was passed

faho:
## Description

Previously this strictly returned the number of unset variables. So if
no variable was given, it would return *true*, which is highly
suspect.

Inspired by re-reading #4033.

## TODOs:

Note: This is *technically* backwards-incompatible. In practice most code that triggers this is probably buggy.

E.g.

```fish
if set -q $foo
```

would enter that if-block (note the stray `$`), which would most likely expect the variable called "foo" to be defined.

### DIFF
--- a/doc_src/cmds/set.rst
+++ b/doc_src/cmds/set.rst
@@ -59,7 +59,7 @@ The following other options are available:
 
 - ``-e`` or ``--erase`` causes the specified shell variables to be erased
 
-- ``-q`` or ``--query`` test if the specified variable names are defined. Does not output anything, but the builtins exit status is the number of variables specified that were not defined, or 255 if more than 255 variables are not defined.
+- ``-q`` or ``--query`` test if the specified variable names are defined. Does not output anything, but the builtins exit status is the number of variables specified that were not defined, up to a maximum of 255. If no variable was given, it also returns 255.
 
 - ``-n`` or ``--names``: List only the names of all defined variables, not their value. The names are guaranteed to be sorted.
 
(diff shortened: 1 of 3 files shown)


krobelus:
Wow yeah, this is another case like `if $status` that is often used wrongly. I have definitely made this error before.

> In practice most code that triggers this is probably buggy.

Right, so what if we report a noisy error in this case? That would give even more user-friendly feedback.
I don't sympathise with code that relies on the old behavior (though I don't know any so who am I to judge).
In any case, the instances of this error easily outnumber the cases where this behavior is used.

(I also can't think of use cases where counting undefined vars with the exit code is useful but I'm sure there is a case)

faho:
>Right, so what if we report a noisy error in this case? That would give even-more user-friendly feedback.

See, experience tells me that error messages in general should be kept for exceptional cases or they'll just be ignored or forgotten. In this case you'd need `if set -q $foo 2>/dev/null`, and you would have to *know* to do that. And that even in the case where you're using it *correctly*, and you expect the `set` to fail if the variable named by foo doesn't exist.

krobelus:
Okay there is nothing wrong with `if set -q $foo` but if it's possible that the variable is undefined I'd write `if set -q foo && set -q $foo`. I'm mainly thinking about the interactive use case, where an explicit error feels better.

faho:
>I'm mainly thinking about the interactive use case, where an explicit error feels better.

Honestly that's more something for a "lint" mode. Imagine turning on a "lint" flog category and getting these, and also warnings about one-argument test or test with unquoted arguments or test with logical operators (we could probably make half of these about test).

Or e.g. warnings about using unset variables. Think of it as an extension of #8131.

Having errors just in the interactive session feels awkward, and it would also be hard to determine when you'd show that. Surely you don't want to do it when you run an autoloaded function (or any plugins you have installed would trigger it). But what about a function defined in the shell? How do you determine if it's defined in the shell? But errroring *just* with a literal `set -q $foo` in the shell would rarely trigger, and wouldn't help at all if you e.g. wrote a plugin in your editor.

PR #8218, opened 2021-08-13
docs: Add copy buttons to all the codeblocks

faho:
This uses a bit of javascript to add copy buttons, so you can directly
copy all the code in a given block to the clipboard!

For codeblocks without prompts, it just copies all the code, for
blocks with prompts, it copies all the lines after prompts, under the
assumption that that's the code to be executed.

It would give you *all* the lines, so the output wouldn't be
interleaved like it is in the html, but good enough.

The buttons appear on hover, so they aren't usable on phones, but
since you won't really have a clipboard on phones and I have no idea
how to make them not always in front of the text otherwise: Eh.

I'm not in love with the javascript here, but it'll do.

## TODOs:

# Screenshots

Firefox, dark theme
Screenshot_20210813_194922

Chromium, light theme
Screenshot_20210813_194956

### DIFF
--- a/doc_src/python_docs_theme/layout.html
+++ b/doc_src/python_docs_theme/layout.html
@@ -58,4 +58,58 @@
     {% trans sphinx_version=sphinx_version|e %}Created using <a href="https://www.sphinx-doc.org/">Sphinx</a> {{ sphinx_version }}.{% endtrans %}
     </div>
 </div>
+
+<script type="text/javascript">
+ function copy_to_clipboard(it) {
+     // Find the pre tag we're interested in.
+     var pre = it.target;
+     while (pre.tagName != "PRE") pre = pre.parentNode;
+     var txt = "";
+     // Cheesy: If we have a prompt,
+     // we only copy prompted lines,
+     // by splitting and matching and stuff
+     if (pre.querySelector('span.gp')) {
+         var texts= [];
+         for (var line of pre.innerText.split('\n')) {
+             if (line.match(/^>_?.*/)) {
+                 texts.push(line.replace(/^>_?/, ""));
+             }
+         }
+         txt = texts.join("\n");
+     } else {
+         // Even cheesier: If we don't have a prompt, we remove the button text from the end.
+         var txt = pre.innerText.substring(0, pre.innerText.length - it.target.innerText.length).trim();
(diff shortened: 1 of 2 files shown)


mqudsi:
I like it. I'm not sure if the smart "prompt elision" could cause any problems (by not being obvious?) or if it's just going to be a nice surprise the first time and then appreciated thereafter.

Do we have `pre` blocks with multiple prompts? How are those handled?

faho:
>Do we have pre blocks with multiple prompts? How are those handled?

Oh, it'll give you all the lines as one, so

```fish
> echo one
one
> echo two
two
```

gives you

```fish
echo one
echo two
```

in the clipboard. That's what I meant by output not being interleaved.

I'd do multiple buttons here, but the linebreaks aren't actually part of the `<span>`s, so I have no idea how to figure out a line once I drop down to the children of the `<pre>`. So doing all lines is a compromise (and the `trim()` removes any trailing newlines, and bracketed paste/fish_clipboard_paste makes it so the lines aren't executed independently).

mqudsi:
I'm just worried that it's too smart. I'm fine if you want to merge it, but just want to say that KISS might apply here. (I feel like either KISS or go all the way with either per-prompt copy buttons like you mentioned or if you *really* want, maybe you can apply a background shading on hover that only applies to the content that would be copied at the same that the copy button becomes active - as an indication of what would be copied?)

As a different matter altogether though, it occurs to me to question if people typically copy-and-paste examples to run directly as-is in the shell? What's the benefit if it's a deterministic process and the result is the same as they see in the browser? Ideally they'd be making changes to the command, which isn't too easy to do post-facto when copy-and-pasting into the shell where it's probably faster to just retype what you want than navigate to the various parts needing changes (until we get mouse support, given that people copying-and-pasting examples are probably not going to be hard-core vim/emacs users).

faho:
>What's the benefit if it's a deterministic process and the result is the same as they see in the browser?

It allows you to see for yourself, easily! And then, once you've seen that, you can alter it.

Of course you could also be typing this stuff, but having a two click option makes it just that little bit quicker and easier that it makes it more tactile.

Now, of course this isn't a massive thing, but it's a *fun* thing that's easily implemented.

PR #8219, opened 2021-08-14
completion: support `--no` prefixes for mpv flag options

mtoohey31:
## Description

mpv options that are labelled as "Flag" under `mpv --list-options` can be prefixed with `--no` to disable that option, but these `--no` prefixed options are currently not completed, so this addition fixes that. It also specifically adds the `--no-audio-display` option (because this is not marked as a "Flag" and so is not caught by the aforementioned changes) and adds a description for it.

~~Fixes issue #~~

## TODOs:

### DIFF
--- a/share/completions/mpv.fish
+++ b/share/completions/mpv.fish
@@ -3,8 +3,14 @@ for opt in $options
     complete -c mpv -l "$opt"
 end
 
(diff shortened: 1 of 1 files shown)


faho:
Merged, thanks!

PR #8221, opened 2021-08-17
edit_command_buffer: use "command" to ignore any functions with the same name

aca:
## Description

When function is defined with the same command name, it causes unwanted side effect.

For example, when we defined function "mv" like this
```
# Defined in /Users/rok/.config/fish/functions/mv.fish @ line 1
function mv
  command mv -n -v $argv
end
```

image


krobelus:
You shouldn't override `mv` like this, see https://github.com/fish-shell/fish-shell/pull/7826#issuecomment-799716818, but sure, why not

PR #8222, opened 2021-08-17
replace push_back with emplate_back

neheb:
The latter forwards the arguments directly.

### DIFF
--- a/src/builtin_argparse.cpp
+++ b/src/builtin_argparse.cpp
@@ -521,7 +521,7 @@ static int validate_and_store_implicit_int(parser_t &parser, const argparse_cmd_
 
     // It's a valid integer so store it and return success.
     opt_spec->vals.clear();
-    opt_spec->vals.push_back(wcstring(val));
+    opt_spec->vals.emplace_back(val);
     opt_spec->num_seen++;
     w.nextchar = nullptr;
     return STATUS_CMD_OK;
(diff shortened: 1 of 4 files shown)


src/builtin_argparse.cpp
```diff
@@ -535,9 +535,9 @@ static int handle_flag(parser_t &parser, const argparse_cmd_opts_t &opts, option
         // short or long flag was given.
         assert(!woptarg);
         if (long_idx == -1) {
-            opt_spec->vals.push_back(wcstring(1, L'-') + opt_spec->short_flag);
+            opt_spec->vals.emplace_back(1, L'-' + opt_spec->short_flag);
```
faho, line 538: I don't think this is correct - the original was just a cheesy way of saying "Make a string containing `-`", via the "fill" constructor. I'm not sure which constructor this is now supposed to use, but it appears to not be working - see the failing tests.

beryllium76 (approved):

faho (changes requested):
It borken.

src/builtin_argparse.cpp
neheb: Right. It's not working because of thr 1. I'll fix.

neheb:
Fixed. Added some extra.

mqudsi:
lgtm, all seem to be valid uses of `emplace_back()`. That said, the benefit probably wouldn't show in even a micro benchmark after all automated optimizations are said and done.

neheb:
Probably not. emplace_back is just shorter.

mqudsi:
Thanks, merged.

PR #8224, opened 2021-08-17
clang-tidy: remove pointless virtual

neheb:
override is already used.


faho:
I'm not entirely up on my C++. I was under the impression that `virtual` here was used so you'd get an error if you instantiated the base class and tried to use these? Which would make it not "pointless", even if the child classes override it.

PR #8225, opened 2021-08-17
tinyexpr: convert typedef to using manually

neheb:
clang-tidy fails hard at transforming this.


krobelus:
this is not really useful - we have a bunch of other places where we use `typedef`.
If we want to change that, then we should change all of them. I'm not sure that's worth the churn.
Let's use `using` for new code and maybe someday switch the old typedefs

krobelus:
ok please report that to the LLVM bug tracker

PR #8226, opened 2021-08-17
clang-tidy: remove redundant initializer

neheb:
It's already defaulted to false.

### DIFF
--- a/src/topic_monitor.cpp
+++ b/src/topic_monitor.cpp
@@ -23,7 +23,7 @@ wcstring generation_list_t::describe() const {
     return result;
 }
 
-binary_semaphore_t::binary_semaphore_t() : sem_ok_(false) {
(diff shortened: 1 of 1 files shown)


src/topic_monitor.cpp
```diff
@@ -23,7 +23,7 @@ wcstring generation_list_t::describe() const {
     return result;
 }
 
-binary_semaphore_t::binary_semaphore_t() : sem_ok_(false) {
+binary_semaphore_t::binary_semaphore_t() {
```
krobelus, line 26: Looking at the surrounding code, I think this initialization helps readability, because there is an 
`if (!sem_ok_)` below. Normally I'd prefer an explicit default value right at the field definition, but here it makes much more sense to place this information at the only constructor.

PR #8227, opened 2021-08-17
change value to rvalue reference

neheb:
clang-tidy recommends using const ref in one of these functions, which
does not really work. This avoids it.

### DIFF
--- a/src/expand.cpp
+++ b/src/expand.cpp
@@ -963,13 +963,13 @@ class expander_t {
     /// An expansion stage is a member function pointer.
     /// It accepts the input string (transferring ownership) and returns the list of output
     /// completions by reference. It may return an error, which halts expansion.
-    using stage_t = expand_result_t (expander_t::*)(wcstring, completion_receiver_t *);
+    using stage_t = expand_result_t (expander_t::*)(wcstring &&, completion_receiver_t *);
 
-    expand_result_t stage_cmdsubst(wcstring input, completion_receiver_t *out);
-    expand_result_t stage_variables(wcstring input, completion_receiver_t *out);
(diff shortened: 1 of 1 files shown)


src/expand.cpp
```diff
@@ -963,13 +963,13 @@ class expander_t {
     /// An expansion stage is a member function pointer.
     /// It accepts the input string (transferring ownership) and returns the list of output
     /// completions by reference. It may return an error, which halts expansion.
-    using stage_t = expand_result_t (expander_t::*)(wcstring, completion_receiver_t *);
+    using stage_t = expand_result_t (expander_t::*)(wcstring &&, completion_receiver_t *);
```
krobelus, line 966: I'm not sure if there is another motivation here but the [style guide](https://google.github.io/styleguide/cppguide.html#Rvalue_references) says to use rvalue references sparingly:

> if you want to "consume" an ordinary function parameter, prefer to pass it by value.

(Is there a way to communicate that to clang-format?)

neheb: It complains on

```
src/expand.cpp:1005:54: warning: the parameter 'input' is copied for each invocation but only used as a const reference; consider making it a const reference [performance-unnecessary-value-param]
```

unescape string takes a const ref parameter and thus does not make sense to use std::move.

krobelus: Yeah, but I don't see a nice way to avoid std::move just for this expansion stage (so it makes no difference whether we take `wcstring &&` or `wcstring`).
It looks like the majority of the expander functions do mutate/move the input parameter, but `stage_variables` doesn't.
Since the all need to have the same signature, we could suppress the warning with
`NOLINT(performance-unnecessary-value-param)` ?

neheb: Might as well.

PR #8228, opened 2021-08-17
convert const ref to rvalue ref

neheb:
clang-tidy wrongly sees an std::move to a const ref parameter and
believes it to be pointless. The copy constructor however is deleted.

### DIFF
--- a/src/fd_monitor.cpp
+++ b/src/fd_monitor.cpp
@@ -199,7 +199,7 @@ void fd_monitor_t::run_in_background() {
     }
 }
 
-void fd_monitor_t::poke_in_background(const poke_list_t &pokelist) {
+void fd_monitor_t::poke_in_background(poke_list_t pokelist) {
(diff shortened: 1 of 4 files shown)


src/fd_monitor.cpp
```diff
@@ -199,7 +199,7 @@ void fd_monitor_t::run_in_background() {
     }
 }
 
-void fd_monitor_t::poke_in_background(const poke_list_t &pokelist) {
+void fd_monitor_t::poke_in_background(poke_list_t &&pokelist) {
```
krobelus, line 202: I think these two are also better as pass-by-value, per https://google.github.io/styleguide/cppguide.html#Rvalue_references. I am not a language lawyer but there should be no behavior difference since we always call `std::move` when calling this function, right?

neheb:
Change to value. There seem to be no ill effects.

PR #8229, opened 2021-08-17
clang-tidy: use for range loops

neheb:

### DIFF
--- a/src/fish_tests.cpp
+++ b/src/fish_tests.cpp
@@ -4089,13 +4089,13 @@ static void test_notifiers_with_strategy(universal_notifier_t::notifier_strategy
     std::unique_ptr<universal_notifier_t> notifiers[notifier_count];
 
     // Populate array of notifiers.
-    for (size_t i = 0; i < notifier_count; i++) {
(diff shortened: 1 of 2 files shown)


src/fish_tests.cpp
```diff
@@ -4089,13 +4089,13 @@ static void test_notifiers_with_strategy(universal_notifier_t::notifier_strategy
...
-    for (size_t i = 0; i < notifier_count; i++) {
-        notifiers[i] = universal_notifier_t::new_notifier_for_strategy(strategy, UVARS_TEST_PATH);
+    for (auto &&notifier : notifiers) {
+        notifier = universal_notifier_t::new_notifier_for_strategy(strategy, UVARS_TEST_PATH);
     }
 
     // Nobody should poll yet.
-    for (size_t i = 0; i < notifier_count; i++) {
-        if (poll_notifier(notifiers[i])) {
+    for (auto &&notifier : notifiers) {
```
krobelus, line 4097: Let's better use normal references in these loops because rvalue references are still a bit obscure.

neheb: Done.

krobelus (commented):
clang-tidy's naming heuristic failed in one place, "sub_re" is not a good loop variable

PR #8230, opened 2021-08-18
fix deleted constructors

neheb:
Some of these have C++98 heritage. C++11 allows public deleted copy/move stuff.

### DIFF
--- a/src/env_universal_common.h
+++ b/src/env_universal_common.h
@@ -179,13 +179,11 @@ class universal_notifier_t {
         strategy_named_pipe,
     };
 
+    universal_notifier_t(const universal_notifier_t &) = delete;
+    universal_notifier_t &operator=(const universal_notifier_t &) = delete;
+
    protected:
     universal_notifier_t();
-
-   private:
-    // No copying.
-    universal_notifier_t &operator=(const universal_notifier_t &);
-    universal_notifier_t(const universal_notifier_t &x);
 
    public:
     static notifier_strategy_t resolve_default_strategy();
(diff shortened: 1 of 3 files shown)


zanchey:
This produces no difference in the compiled output, so I'm not sure it's needed?

src/common.h
```diff
@@ -243,8 +243,6 @@ struct [[gnu::unused]] noncopyable_t {
     noncopyable_t() = default;
     noncopyable_t(noncopyable_t &&) = default;
     noncopyable_t &operator=(noncopyable_t &&) = default;
-    noncopyable_t(const noncopyable_t &) = delete;
-    noncopyable_t &operator=(const noncopyable_t &) = delete;
```
krobelus, line 247: https://google.github.io/styleguide/cppguide.html#Copyable_Movable_Types says

> Explicitly declaring or deleting all four copy/move operations is permitted, but not required

but it also says

> a move-only class should explicitly declare the move operations, and a non-copyable/movable class should explicitly delete the copy operations

I read that as a preference for the explicit (albeit redundant) `delete`. I have no idea if there's a difference between "move-only" and "non-copyable/movable". If they're the same thing then the wording can be simplified.

krobelus: So in a case like this let's maybe disable the motivating clang-tidy check (which one is it?) to match the style guide.

neheb: There's no clang-tidy check that complains about explicit constructors.

neheb: Added them back.

krobelus:
> This produces no difference in the compiled output, so I'm not sure it's needed?

not needed ofc, but explicitly deleting a constructor seems better than the private-constructor trick

PR #8234, opened 2021-08-19
No more sys_errlist

faho:
## Description

This removes safe_strerror and safe_perror (which uses the former internally), and replaces it with hardcoded error messages.

The impetus for this is that glibc actually *removed* sys_errlist and according to #7006 we can't use strerror_r either.

I looked up the used errno values I could find. It's possible these are incomplete, in which case we get more "Unknown error $errno" messages, but I don't think that should be a common thing, and on current glibc that's already *all* of the errors here (except for the exec errors that we already explained).

Fixes issue #4183.

Ideally I would like this to be included in 3.4.0, provided we can get enough testing.

## TODOs:

### DIFF
--- a/src/postfork.cpp
+++ b/src/postfork.cpp
@@ -472,7 +472,10 @@ void safe_report_exec_error(int err, const char *actual_cmd, const char *const *
             FLOGF_SAFE(exec, "Failed to execute process '%s': Unsupported format.", actual_cmd);
             break;
         }
-        case EISDIR:
+        case EISDIR: {
+            FLOGF_SAFE(exec, "Failed to execute process '%s': File is a directory.", actual_cmd);
+            break;
+        }
         case ENOTDIR: {
             FLOGF_SAFE(exec, "Failed to execute process '%s': A path component is not a directory.", actual_cmd);
             break;
--- a/src/postfork.cpp
+++ b/src/postfork.cpp
@@ -68,7 +68,30 @@ void report_setpgid_error(int err, bool is_parent, pid_t desired_pgid, const job
     }
 
     errno = err;
-    safe_perror("setpgid");
+    switch (errno) {
+        case EACCES: {
+            FLOGF_SAFE(error, "setpgid: Process %s has already exec'd", pid_buff);
+            break;
+        }
+        case EINVAL: {
+            FLOGF_SAFE(error, "setpgid: pgid %s unsupported", getpgid_buff);
+            break;
+        }
+        case EPERM: {
+            FLOGF_SAFE(error, "setpgid: Process %s is a session leader or pgid %s does not match", pid_buff, getpgid_buff);
+            break;
+        }
+        case ESRCH: {
+            FLOGF_SAFE(error, "setpgid: Process id %d does not match", pid_buff);
+            break;
+        }
+        default: {
+            char errno_buff[64];
+            format_long_safe(errno_buff, errno);
+            FLOGF_SAFE(error, "setpgid: Unknown error number %s", errno_buff);
+            break;
+        }
+    }
 }
 
 int execute_setpgid(pid_t pid, pid_t pgroup, bool is_parent) {
@@ -200,7 +223,26 @@ pid_t execute_fork() {
         }
     }
 
-    safe_perror("fork");
+    // These are all the errno numbers for fork() I can find.
+    // Also ENOSYS, but I doubt anyone is running
+    // fish on a platform without an MMU.
+    switch (errno) {
+        case EAGAIN: {
+            // We should have retried these already?
+            FLOGF_SAFE(error, "fork: Out of resources. Check RLIMIT_NPROC and pid_max.");
+            break;
+        }
+        case ENOMEM: {
+            FLOGF_SAFE(error, "fork: Out of memory.");
+            break;
+        }
+        default: {
+            char errno_buff[64];
+            format_long_safe(errno_buff, errno);
+            FLOGF_SAFE(error, "setpgid: Unknown error number %s", errno_buff);
+            break;
+        }
+    }
     FATAL_EXIT();
     return 0;
 }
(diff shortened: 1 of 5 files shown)


src/postfork.cpp
```diff
@@ -415,10 +456,31 @@ void safe_report_exec_error(int err, const char *actual_cmd, const char *const *
             FLOGF_SAFE(exec, "Out of memory");
             break;
         }
-
+        case EACCES: {
+            FLOGF_SAFE(exec, "Failed to execute process '%s': The file could not be accessed.", actual_cmd);
+            break;
+        }
+        case ETXTBSY: {
+            FLOGF_SAFE(exec, "Failed to execute process '%s': File is currently open for writing.", actual_cmd);
```
faho, line 464: I specifically made these messages *not* just the ones we get by default.

That's because "text file is busy" is among the worst error messages I have ever read, mostly because it has a 98% chance of appearing when the file is *not* a text file (in the sense any user cares about).

It's kinda like if an ice cream place told you "machine not fillable" when they're out of a certain flavor. Might be true in the technical sense, but so misleading that it might as well not be said, at all.

faho: So I would welcome more wording suggestions.

src/postfork.cpp
```diff
@@ -415,10 +456,31 @@ void safe_report_exec_error(int err, const char *actual_cmd, const char *const *
...
+            FLOGF_SAFE(exec, "Failed to execute process '%s': A path component is not a directory.", actual_cmd);
+            break;
+        }
+        
         default: {
-            const char *err = safe_strerror(errno);
-            FLOGF_SAFE(exec, "%s", err);
+            char errnum_buff[64];
+            format_long_safe(errnum_buff, err);
+            FLOGF_SAFE(exec, "Failed to execute process '%s', unknown error number %d", actual_cmd, errnum_buff);
```
krobelus, line 483:
```suggestion
            FLOGF_SAFE(exec, "Failed to execute process '%s', unknown error number %s", actual_cmd, errnum_buff);
```

src/postfork.cpp
```diff
@@ -415,10 +456,31 @@ void safe_report_exec_error(int err, const char *actual_cmd, const char *const *
             FLOGF_SAFE(exec, "Out of memory");
             break;
         }
-
+        case EACCES: {
+            FLOGF_SAFE(exec, "Failed to execute process '%s': The file could not be accessed.", actual_cmd);
```
krobelus, line 460: probably run clang-format even if it makes this look slightly worse

src/postfork.cpp
```diff
@@ -68,7 +68,30 @@ void report_setpgid_error(int err, bool is_parent, pid_t desired_pgid, const job
...
+        case EINVAL: {
+            FLOGF_SAFE(error, "setpgid: pgid %s unsupported", getpgid_buff);
+            break;
+        }
+        case EPERM: {
+            FLOGF_SAFE(error, "setpgid: Process %s is a session leader or pgid %s does not match", pid_buff, getpgid_buff);
+            break;
+        }
+        case ESRCH: {
+            FLOGF_SAFE(error, "setpgid: Process id %d does not match", pid_buff);
```
krobelus, line 85:
```suggestion
            FLOGF_SAFE(error, "setpgid: Process ID %d does not match", pid_buff);
```

krobelus (commented):
Looks great! The messages read nice.

src/postfork.cpp
faho: Also `%s` - I need to remember these numbers are all pre-stringed.

src/postfork.cpp
```diff
@@ -415,10 +456,31 @@ void safe_report_exec_error(int err, const char *actual_cmd, const char *const *
...
+        }
+        case ELOOP: {
+            FLOGF_SAFE(exec, "Failed to execute process '%s': Too many layers of symbolic links.", actual_cmd);
+            break;
+        }
+        case EINVAL: {
+            FLOGF_SAFE(exec, "Failed to execute process '%s': Unsupported format.", actual_cmd);
+            break;
+        }
+        case ENOTDIR: {
```
faho, line 475: Note: Also EISDIR

src/postfork.cpp
```diff
@@ -415,10 +456,31 @@ void safe_report_exec_error(int err, const char *actual_cmd, const char *const *
...
+            FLOGF_SAFE(exec, "Failed to execute process '%s': A path component is not a directory.", actual_cmd);
+            break;
+        }
+        
         default: {
```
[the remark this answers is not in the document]
faho: I *think* we check for pathname length before fork?

EPERM is interesting here because it's such a weird failure case - "Either setuid/setgid is forbidden or you lack capabilities" is the best I can come up with.

zanchey:
We can use `strerrordesc_np` as a replacement for `sys_errlist` on platforms that support it ([noted by @ridiculousfish](https://gitter.im/fish-shell/fish-shell?at=60f70f2c23956a5aa44834ab)), though that doesn't fix the issues of the poorly-written error descriptions.

faho:
>We can use strerrordesc_np as a replacement for sys_errlist on platforms that support it

I don't know if strerrordesc_np provides any better guarantees than `strerror_r` ,which should be more widespread as it's in POSIX. We rejected that one in #7006.

faho:
Okay, tested on arch, alpine, FreeBSD, NetBSD, and Ubuntu and macOS on Github Actions.

It seems to compile everywhere and I can't find somewhere the errors are actually *wrong*, so I'm merging this now because it improves the status quo and I'd like to have it in 3.4.0.

More improvement is of course welcome!

mqudsi:
I mean, you can’t beat error messages custom tailored to the faulting call - great work @faho.

zanchey:
`strerrordesc_np` is specifically marked as async-signal-safe, but yeah, this is good

PR #8238, opened 2021-08-20
clang-tidy: run through normal checks

neheb:
There's a .clang-tidy file in here.

### DIFF
--- a/src/expand.cpp
+++ b/src/expand.cpp
@@ -739,7 +739,7 @@ static expand_result_t expand_cmdsubst(wcstring input, const operation_context_t
         }
         wcstring sub_res_joined;
         sub_res_joined.reserve(approx_size);
-        for (const wcstring &line : sub_res) {
+        for (wcstring &line : sub_res) {
             sub_res_joined.append(escape_string_for_double_quotes(std::move(line)));
             sub_res_joined.push_back(L'\n');
         }
--- a/src/fds.cpp
+++ b/src/fds.cpp
@@ -100,7 +100,7 @@ fd_event_signaller_t::fd_event_signaller_t() {
 int fd_event_signaller_t::write_fd() const { return write_.fd(); }
 #endif
 
-bool fd_event_signaller_t::try_consume() {
+bool fd_event_signaller_t::try_consume() const {
     // If we are using eventfd, we want to read a single uint64.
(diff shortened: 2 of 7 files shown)


src/expand.cpp
```diff
@@ -740,7 +740,7 @@ static expand_result_t expand_cmdsubst(wcstring input, const operation_context_t
         wcstring sub_res_joined;
         sub_res_joined.reserve(approx_size);
         for (const wcstring &line : sub_res) {
-            sub_res_joined.append(escape_string_for_double_quotes(std::move(line)));
+            sub_res_joined.append(escape_string_for_double_quotes(line));
```
krobelus, line 743: Umm this warning was introduced by faf51e0693b73ceae21a1ed34a6ac452b96cace6
I forgot that we do want to move out of `sub_res` here to avoid a copy, so we can't use `const wcstring &` but rather `wcstring &`. If clang-tidy suggested `const wcstring &` here, that's a bug IMO.

neheb: Fixed. I probably manually did it to be consistent with the surrounding code.

Generally speaking, auto&& should be used in for loops like this. That way you don't worry around bad transformations.

auto&& in for range loops is not the same as a function parameter as mentioned in the google style doc.

krobelus: > bad transformations.

can you give an example of what can go  wrong when using `auto &`?
This might be a good addition to that style guide.

neheb: auto&& transforms into either auto& or const auto&, with the latter being preferred.

When it encounters an std::move, the compiler knows to evaluate to auto& and clang-tidy will not remove the std::move like it did here.

Note that I'm talking about for range loops. function parameters are a different beast.

As a sidenote: In a different project, I transformed all for range loops to auto&& and got a size reduction. I haven't tried here.

krobelus: I'd keep using `const auto &` over `auto &&` because it prevents us from accidentally modifying the value.
Then it only makes sense to also use `auto &` which is simpler and I haven't found any downsides.

>  I transformed all for range loops to auto&& and got a size reduction. I haven't tried here.

Wow, I'd love to find a small example that shows how this changes the generated code.

neheb: If you're curious: https://github.com/gerbera/gerbera/commit/66b3c7736f2ffa7e1faa98ec2d62804ecb555683

PR #8252, opened 2021-08-26
add constexpr

neheb:
This was not const before.

Small size reduction.

### DIFF
--- a/src/env_dispatch.cpp
+++ b/src/env_dispatch.cpp
@@ -556,7 +556,7 @@ static void init_curses(const environment_t &vars) {
     curses_initialized = true;
 }
 
-static const char *utf8_locales[] = {
+static constexpr const char *utf8_locales[] = {
     "C.UTF-8",
     "en_US.UTF-8",
     "en_GB.UTF-8",


src/env_dispatch.cpp
```diff
@@ -556,7 +556,7 @@ static void init_curses(const environment_t &vars) {
     curses_initialized = true;
 }
 
-static const char *utf8_locales[] = {
+static constexpr const char *utf8_locales[] = {
```
krobelus, line 559: The behavior change is hardly relevant (can't the compiler figure out that
these are "effectively constexpr") but the extra constexpr annotation can
be helpful.
It feels like a reasonable change, but this is of course a drop in the ocean
since we have many similar occurrences. I'd rather change all of them in
one commit.
I'd do it since we already use constexpr a bunch and it helps us pick up the
habit of using it consistently across our code but I'm not sure what others
think - is this style improvement worth the churn?  .

faho: Honestly, I don't care about any of these style "improvements".

It doesn't markedly improve readability, it adds a lot of churn, it has previously broken alternative platforms.

In this case there's apparently a slight behavioral improvement so it's probably okay, but I no longer want to merge any things just because it placates some sense of "modern" C++ style.

neheb: I will note again that there is a size reduction with this commit. The compiler does not make this const automatically.

faho: >I will note again that there is a size reduction with this commit

I have noted this in my comment - that's the "behavioral improvement" I was talking about.

However, many of your commits don't have any of these, so I'll ask that you please not submit purely cosmetic style fixes any more. It's simply not worth it.

PR #8256, opened 2021-08-28
Fix completions/coredumpctl.fish and add new complete

akiirui:
## Description

Old completion from `coredumpctl` missing `--arguments` in subcommand complete, Fix it.
And add new options complete from systemd 249.

Fixes issue #

## TODOs:


faho:
Merged, thanks!

Note that we don't use commit names as formal as "fix(" and "feat(", but it's not worth quibbling about.

Issue #8264, opened 2021-09-01
Getting an error "cd: '/Users/x/Downloads/' is a rotten symlink"


elimist3:
Today I logged onto my machine and tried to `cd ~/Downloads`, but to my surprise fish returned an error `cd: '/Users/x/Downloads/' is a rotten symlink`.

It seems to originate from the `builtin_cd` function call [here](https://github.com/fish-shell/fish-shell/blob/master/src/builtin_cd.cpp#L112), however I can't tell what the reason is since my Downloads directory is not a symlink.

I'm able to cd into any other directory just fine and I'm able to access `~/Downloads` through MacOS's file explorer (Finder) so I don't think the directory is corrupt.

When I `chsh -s /bin/bash`, I'm able to `cd ~/Downloads` just fine!

My `config.fish` is just the default config, and I've already tried reinstalling Fish from my package manager (homebrew), to no avail.

**System info**
Fish version: 3.3.1
Operating system: MacOS 11.5.2, arm64
Terminal: Alacritty
Affected by third-party customizations: no, output is the same

I have a feeling this may be difficult to reproduce but I thought I'd put this out there just in case there is an easy solution to be found.

krobelus:
The error message is misleading because `EROTTEN` is a fish invention that was removed in 7ac593273 ("Remove a bunch of dead code identified by cppcheck"). It is defined as 1, which matches [EPERM](https://opensource.apple.com/source/xnu/xnu-201.5/osfmk/libsa/errno.h.auto.html).
I'm not sure how you're getting a permissions error; `fchdir` should never set `EPERM`.
Maybe you've [allowed] bash to access files but not fish? I'm not sure.

```diff
diff --git a/src/builtin_cd.cpp b/src/builtin_cd.cpp
index 9d0fa6e58..1b1910fe2 100644
--- a/src/builtin_cd.cpp
+++ b/src/builtin_cd.cpp
@@ -108,9 +108,7 @@ maybe_t<int> builtin_cd(parser_t &parser, io_streams_t &streams, const wchar_t *
     } else if (best_errno == ENOENT) {
         streams.err.append_format(_(L"%ls: The directory '%ls' does not exist\n"), cmd,
                                   dir_in.c_str());
-    } else if (best_errno == EROTTEN) {
-        streams.err.append_format(_(L"%ls: '%ls' is a rotten symlink\n"), cmd, dir_in.c_str());
-    } else if (best_errno == EACCES) {
[...]

PR #8270, opened 2021-09-05
builtin cd: print error about broken symlink

krobelus:
## Description

Fixes issue #8264

## TODOs:

### DIFF
--- a/src/wutil.cpp
+++ b/src/wutil.cpp
@@ -203,6 +203,29 @@ int make_fd_blocking(int fd) {
     return err == -1 ? errno : 0;
 }
 
+maybe_t<wcstring> wreadlink(const wcstring &file_name) {
+    struct stat buf;
+    if (lwstat(file_name, &buf) == -1) {
+        return none();
+    }
+    ssize_t bufsize = buf.st_size + 1;
+    char target_buf[bufsize];
+    const std::string tmp = wcs2string(file_name);
+    ssize_t nbytes = readlink(tmp.c_str(), target_buf, bufsize);
+    if (nbytes == -1) {
+        wperror(L"readlink");
+        return none();
+    }
+    // The link might have been modified after our call to lstat.  If the link now points to a path
+    // that's longer than the original one, we can't read everything in our buffer.  Simply give
+    // up. We don't need to report an error since our only caller will already fall back to ENOENT.
+    if (nbytes == bufsize) {
+        return none();
+    }
+
+    return str2wcstring(target_buf, nbytes);
+}
+
 /// Wide character realpath. The last path component does not need to be valid. If an error occurs,
 /// wrealpath() returns none() and errno is likely set.
 maybe_t<wcstring> wrealpath(const wcstring &pathname) {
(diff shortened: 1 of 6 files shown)


src/wutil.cpp
```diff
@@ -203,6 +203,27 @@ int make_fd_blocking(int fd) {
...
 }
 
+maybe_t<wcstring> wreadlink(const wcstring &file_name) {
+    struct stat buf;
+    if (lwstat(file_name, &buf) == -1) {
+        return none();
+    }
+    ssize_t bufsize = buf.st_size + 1;
+    char target_buf[bufsize];
+    const std::string tmp = wcs2string(file_name);
```
faho, line 213: (note: We really really do way too much of this wcs2string and back - this will do wcs2string for the `lwstat` above, and then wcs2string again for the readlink. In some benchmarks ~50% of the time is taken in widening and narrowing strings)

(but since this is just for the error, it's okay for now - we should just probably start doing all the path stuff in narrow exclusively, so you can keep one narrow string around if you need to do more than one thing)

src/wutil.cpp
```diff
@@ -705,9 +726,7 @@ file_id_t file_id_for_fd(int fd) {
     return result;
 }
 
-file_id_t file_id_for_fd(const autoclose_fd_t &fd) {
-    return file_id_for_fd(fd.fd());
-}
+file_id_t file_id_for_fd(const autoclose_fd_t &fd) { return file_id_for_fd(fd.fd()); }
```
faho, line 729: This line wasn't touched, at all. That's the issue with running clang-format on save - you touch a lot of unrelated code.

We should probably do another formatting run (but changes like this are the worst - clang-format is super anal about having lines exactly the correct way, which is just useless churn and should be disabled, but I've so far not found a way)

krobelus: > but changes like this are the worst - clang-format is super anal about having lines exactly the correct way, which is just useless churn and should be disabled, but I've so far not found a way

I'm not sure what you mean exactly - it's the formatter's job to be precise, so we don't have to.
This style can be configured with `AllowShortFunctionsOnASingleLine`, see also https://zed0.co.uk/clang-format-configurator/. I somehow have lots of energy for evangelizing clang-format.. it makes refactorings much easier and avoids thinking or communicating about style.

Assuming we want this change, I'd probably still do it because it's small and I'm already touching the file / invalidating builds. Though that's really not a strong reason.

faho: The problem is that it's *annoyingly* precise.

We've had cases where it flip-flopped. E.g. rename this variable to "file_desc" and it might just decide that it *has to* be split. That's reformatting that basically helps nobody. Line length really only matters if lines constantly go over by a lot which requires you to scroll to view the lines you care about.

And because this particular kind of reformatting isn't whitespace-only, `git blame -w` won't ignore it and will show you the reformatting commit first, so you then need to run it again with `somehash^`. There's a way to give `git blame` commits to ignore, but that also only really works if the commits only reformat (or you could be ignoring a functional change!).

So I'm really quite annoyed by clang-format's super precise insistence on line length and would much prefer if it had some sort of slack where it keeps whatever's there, because as it stands it either forces joined style or forces separated style. (that's what I meant by "disabling" it - currently you can only say "always separate" or "separate exactly when clang-format decides to")

And if we can't have that, we could at least restyle the entire codebase every so often in separate commits and add their hashes to some file so I can run `git blame -w --ignore-revs-file :/restylecommits.txt` (or set "blame.ignoreRevsFile") to ignore them, and stop restyling unrelated stuff in regular commits.

faho: Note: I don't particularly care about this one occurence, it's more that I don't want commits to have unrelated formatting changes so `git blame` is nicer to use, and that I consider clang-format to not be a great autoformatter because it is annoyingly precise, which causes it to add a bunch of formatting churn.

PR #8281, opened 2021-09-09
normalize_path: Squash leading slashes even without allow_leading

faho:
This currently changes builtin realpath with the "-s" option:

    builtin realpath -s ///tmp

previously would print "///tmp", now it prints "/tmp".

The only thing "allow_leading_double_slashes" does is allow *two*
slashes.

This is important for `path match`, to be introduced in #8265.

## TODOs:


faho:
Okay, the only use of this is in builtin realpath, all other callsites allow leading double slashes.

So I feel comfortable merging this.

PR #8287, opened 2021-09-12
Prepend command to cat

victordiaz:
*Problem:*
edit_command_buffer uses `cat` to return the modified content.
If a person has an alias for `cat` to a different command such `bat`** the editing will not be useful anymore since bat decorates the text with frames, line counts, etc

*Solution*
Appending command to cat, fish will ignore the alias and execute the real command according to this https://fishshell.com/docs/current/cmds/command.html

** https://github.com/sharkdp/bat

## Description

Talk about your changes here.

Fixes issue #

## TODOs:

### DIFF
--- a/share/functions/edit_command_buffer.fish
+++ b/share/functions/edit_command_buffer.fish
@@ -70,7 +70,7 @@ function edit_command_buffer --description 'Edit the command buffer in an extern
     if test $editor_status -eq 0 -a -s $f
         # Set the command to the output of the edited command and move the cursor to the
         # end of the edited command.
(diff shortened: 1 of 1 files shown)


zanchey:
Several of these pull requests have been merged, because they are essentially harmless, but with the proviso that you are playing with fire here. If you replace a common system command with another that does not do the same thing, when it breaks, you get to keep the pieces. (eg this pull request does not protect against someone installing `bat` as `/usr/bin/cat`).

For interactive use, you are much better off defining an abbreviation rather than using an alias.

PR #8299, opened 2021-09-17
Make less version check compatible with older Fish

kidonng:
## Description

First of all, the background. I've been putting the git repo of fish in my `$fish_function_path` and `$fish_complete_path` for some time, to get faster bugfixes/better completions, due to the relatively long breaks during releases (especially my experiences with *cough, 3.2.0, cough*).

I'm aware this is a rare usage and not supported, but it works fine until today I noticed 05fdee1be781009f663fb995c26142002d612afa breaks `history`, due to usage of newer syntaxes and options. In my opinion, we should try not to break things unless necessary. In this particular case:

- `string match -g` is a valid usage, but explicitly specifying `[2]` would be better and backwards-compatible.
- `"$()"` is unnecessary, and I fear the "abuse" of this syntax will just lead to more confusion when people find out unquoted `$()` behaves differently. We should prefer the Fish way.

## TODOs:

### DIFF
--- a/share/functions/__fish_print_help.fish
+++ b/share/functions/__fish_print_help.fish
@@ -121,7 +121,7 @@ function __fish_print_help --description "Print help message for the specified f
             # similar to man, but add -F to quit paging when the help output is brief (#6227)
             # Also set -X for less < v530, see #8157.
             set -l lessopts isRF
-            if test "$(less --version | string match -rg 'less (\d+)')" -lt 530 2>/dev/null
+            if test (less --version | string match -r 'less (\d+)')[2] -lt 530 2>/dev/null
                 set lessopts "$lessopts"X
             end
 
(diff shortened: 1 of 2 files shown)


faho:
>"$()" is unnecessary, and I fear the "abuse" of this syntax will just lead to more confusion when people find out unquoted $() behaves differently. We should prefer the Fish way.

Oh, this one *will* happen more. In particular with `test`, because `test` sucks. It's the absolute single worst piece of all of fish, and `"$()"` eases the pain quite a bit.

There is no "the Fish way" that's referring the older `()`. `"$()"` is *helpful*, so it will be used. And "unquoted" `$()` is exactly `()`. Quoted variables also behave differently from unquoted ones, so I have no idea what you're on about.

That being said:

As long as the backwards-compatible code isn't significantly worse - either much longer or less functional - using older features is probably wise.

That's probably the case here, so let's merge it.

PR #8321, opened 2021-09-27
Fix typo in `set_colors` command documentation

askoufis:
## Description

Referencing the [previous commit](https://github.com/fish-shell/fish-shell/commit/1643df0d23483f84c79c0c45b014b6d4b8ca8ab4#diff-4c6511de5a67fc5797b0baa3c8fa664dfa2e0f40ea0e93539add05205054eb09R16) that touched this text to try figure out what the intended wording was. An alternative could be to replace `It` with `VALUE` because `It` could be interpreted ambiguously in this context.

### DIFF
--- a/doc_src/cmds/set_color.rst
+++ b/doc_src/cmds/set_color.rst
@@ -13,7 +13,7 @@ Synopsis
 Description
 -----------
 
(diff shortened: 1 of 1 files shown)


floam:
> An alternative could be to replace It with VALUE because It could be interpreted ambiguously in this context.

Yes that is better.
Please update the PR or I can just do it.

askoufis:
> > An alternative could be to replace It with VALUE because It could be interpreted ambiguously in this context.
> 
> Yes that is better. Please update the PR or I can just do it.

I've updated it.

PR #8334, opened 2021-10-05
Update ip.fish

zaleoth:
## Description

Found a bug in "ip" autocompletion and tried to fix it : whenever ip is aliased (like ip=ip -c -br), the autocompletion using the "ip" command fails to find current devices
Add "command" directive in front of ip commands used to create autocompletion to prevent bug from an aliased "ip"

Fixes issue [no issue actually opened for this, shall I've had one ?]

This is my first contribution, hope this will be ok for a small fish script edition

### DIFF
--- a/share/completions/ip.fish
+++ b/share/completions/ip.fish
@@ -176,7 +176,7 @@ function __fish_ip_commandwords
 end
 
 function __fish_ip_device
-    ip -o link show | while read -l a b c
+    command ip -o link show | while read -l a b c
         printf '%s\t%s\n' (string replace ':' '' -- $b) Device
     end
 end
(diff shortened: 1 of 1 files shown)


faho:
>Add "command" directive in front of ip commands used to create autocompletion to prevent bug from an aliased "ip"

Honestly: Don't alias commands to other things. We cannot, in general, be expected to handle that. If something is called `ip`, it should basically act like `ip`.

In this specific case, `ip -c` turns on color indiscriminately, without checking - it's like `grep --color=always`, not `grep --color=auto`. Simply call your alias something slightly different like ´ipc` or use an abbreviation instead, and this'll go away, or make it a function like

```fish
function ip
    if isatty stdout
        command ip -c $argv
    else
        command ip $argv
    end
end
```

That being said, our approach so far has been to merge things like these when they appear to cause issues, so I'm merging this. You should still fix your alias - if any script called `ip` it would be broken as well.

Issue #5791, opened 2019-04-03
`history delete` function seems inadequate for common use cases


alxpettit:
Using fish version `3.0.2-661-g6e525cc5d` on Arch Linux with uxterm.

The main thing I find myself using `history delete` for is clearing a password from the shell history. Thus it would be reasonable to expect it to be easy to delete, say, the last 10 lines from history, or the history from the _current_ session (but spare all other history).

But the interface expects me to delete by search term instead, which wastes time when I could have just typed `history clear-session` or something like `history delete -10` (to delete the last 10 lines, ideally without it bringing up a confirmation thing).

What does everyone think?

floam:
If we made $history not read-only, this could be done with `set -e history[1..10]`.

mqudsi:
But this isn't a complaint about the inability to modify the `history` variable, rather the unfriendliness of the `history` builtin.

alxpettit:
> But this isn't a complaint about the inability to modify the `history` variable, rather the unfriendliness of the `history` builtin.

Yeah, exactly. I was a little bit incredulous reading the docs because I strongly expected it to have a function for clearing only the current session or for clearing the last X number of lines... I half expected someone to tell me that there was such a function that wasn't documented, which is why I didn't phrase this as a feature request.

floam:
It seems like a reasonable feature request to me.

InspectorMustache:
How about this elegant and not at all hacky workaround in the meantime:

```
for i in $history[1..10]
    history delete --exact --case-sensitive -- "$i"
end
```

alxpettit:
@InspectorMustache That works great for now! I'll wrap it in a temporary function until this is fixed.

mqudsi:
I think a nice approach (that would also address some of the concerns in #5482) would be to make a bare `history delete` trigger a `read` prompt for keywords to sanitize from the history, which would a) be easier to use, and b) take care of the privacy hole whereby `history delete foo` deletes `foo` from history but re-introduces it with a new `history delete foo` entry in history

I also like the idea of a separate `history clear-session` in addition to the existing `history clear`, this one wouldn't even need to ask for confirmation as it's No Big Deal™.

faho:
>bare history delete trigger a read prompt for keywords to sanitize from the history, which would a) be easier to use, and b) take care of the privacy hole whereby history delete foo deletes foo from history but re-introduces it with a new history delete foo entry in history

One problem: We keep a `read` history.

[...]

PR #8337, opened 2021-10-07
Adds sub-command clear-session to history command. Issue #5791

s7rul:
## Description
Adds sub-command ``clear-session`` to ``history`` command. Which clear current session history. If ``history merge`` or ``builtin history merge`` is run in a session only the history after this will be erased.

## Implementation
A function ``clear_session`` witch does this is added to ``src/history.cpp`` where it first puts all of the commands in ``new_items`` into ``deleted_items``after this ``new_items`` is cleared and ``first_unwritten_new_item_index`` is reset.

Documentation is updated in ``doc_src`` and completion. Tests were written for the added code (no unit test in ``fish_tests.cpp``).

Fixes issue #5791

## Additional notes
We have done this as part in a school project and it is our first contribution to open-source.

## TODO

- [x] Make test pass
- [x] Make it not affect old history via de-duplication.

### DIFF
--- a/share/completions/history.fish
+++ b/share/completions/history.fish
@@ -1,5 +1,5 @@
 # Note that when a completion file is sourced a new block scope is created so `set -l` works.
-set -l __fish_history_all_commands search delete save merge clear
+set -l __fish_history_all_commands search delete save merge clear clear-session
 
 complete -c history -s h -l help -d "Display help and exit"
 
@@ -33,3 +33,5 @@ complete -f -c history -n "not __fish_seen_subcommand_from $__fish_history_all_c
     -a merge -d "Incorporate history changes from other sessions"
 complete -f -c history -n "not __fish_seen_subcommand_from $__fish_history_all_commands" \
     -a clear -d "Clears history file"
+complete -f -c history -n "not __fish_seen_subcommand_from $__fish_history_all_commands" \
+    -a clear-session -d "Clears all history from the current session"
--- a/share/functions/history.fish
+++ b/share/functions/history.fish
@@ -62,13 +62,15 @@ function history --description "display or manipulate interactive command histor
         set hist_cmd search
     else if set -q _flag_merge
         set hist_cmd merge
+    else if set -q _flag_clear-session
+        set hist_cmd clear-session
     end
 
     # If a history command has not already been specified check the first non-flag argument for a
     # command. This allows the flags to appear before or after the subcommand.
     if not set -q hist_cmd[1]
         and set -q argv[1]
-        if contains $argv[1] search delete merge save clear
+        if contains $argv[1] search delete merge save clear clear-session
             set hist_cmd $argv[1]
             set -e argv[1]
         end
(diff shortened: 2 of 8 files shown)


ridiculousfish:
This is great! Impressively simple implementation, nice tests too. I am able to reproduce the test failures locally, I will investigate to ensure we aren't masking some serious bug.

Regarding touching old history, faho is correct that this will erase items from previous sessions, if the item is also in the current session. I think this can be fixed easily by tracking in `deleted_items` whether the deletion is for all sessions or just this session, and then inspect that field when de-duplicating to decide whether to erase. If you'd like to tackle that please do; otherwise I will have a go and add on top of this commit.

s7rul:
> Regarding touching old history, faho is correct that this will erase items from previous sessions, if the item is also in the current session. I think this can be fixed easily by tracking in `deleted_items` whether the deletion is for all sessions or just this session, and then inspect that field when de-duplicating to decide whether to erase. If you'd like to tackle that please do; otherwise I will have a go and add on top of this commit.

I can give it a try, sounds simple enough.

ridiculousfish:
So the test failures were due to [this hack](https://github.com/fish-shell/fish-shell/blob/8784253282289455ae2168be6eb27232cfdbb1d9/src/history.cpp#L575) and not a problem with this PR.

The gory details: fish shares a single history file with all other instances.  Any instance should only "see" past items, not items from other running instances. It does this via timestamps: a fish instance has a start timestamp. Items with earlier timestamps are "old" and should be considered part of history. Items with later timestamps are "new" and should be ignored by this instance.

A problem arises with items whose timestamp exactly matches fish's start date: those items may be old or new. For example `exec fish` is likely to create an "old" history item that exactly matches fish's start date. Meanwhile scripted tests also create history items in the same second as fish launches. To distinguish these, any new history item created in the same second as fish starts will have its timestamp incremented by one.

Really we should be using higher resolution timestamps. I'll see if I can fix this another way but it shouldn't block merge.

ridiculousfish:
> Also perhaps `history clear --session` instead?

I prefer a separate command `clear-session`. Usually you want commands to do the least dangerous thing and for modifiers to make them more dangerous, e.g. `rm` vs `rm -R`. But with `history clear --session` it would be reversed: the default would be more destructive. So I think a separate command is better in this case.

src/history.cpp
```diff
@@ -1073,6 +1076,15 @@ void history_impl_t::clear() {
     this->clear_file_state();
 }
 
+void history_impl_t::clear_session() {
+    for (size_t i = 0; i < new_items.size(); i++) {
+        deleted_items.insert(new_items.at(i).str());
```
ridiculousfish, line 1081: nit: you could use a range-based for loop here to simplify it:

    for (const auto &item : new_items) {
        deleted_items.insert(item.str());
    }

share/completions/history.fish
```diff
@@ -1,5 +1,5 @@
 # Note that when a completion file is sourced a new block scope is created so `set -l` works.
-set -l __fish_history_all_commands search delete save merge clear
+set -l __fish_history_all_commands search delete save merge clear clear-session
```
zx8, line 2:
```suggestion
set -l __fish_history_all_commands search delete save merge clear clear_session
```

share/functions/history.fish
```diff
@@ -62,13 +62,15 @@ function history --description "display or manipulate interactive command histor
         set hist_cmd search
     else if set -q _flag_merge
         set hist_cmd merge
+    else if set -q _flag_clear-session
```
zx8, line 65:
```suggestion
    else if set -q _flag_clear_session
```

src/history.cpp
s7rul: Changed this in later commit.

ridiculousfish:
Looks good, squash-merged as 049104e8df02a989e3c0ad70d2521251192013cd. Thank you, a wonderful contribution!

PR #8338, opened 2021-10-09
Display local branches first in git switch completion

zeertzjq:
## Description

Usually local branches have remote branches with the same name, and in completion they are currently overshadowed by unique remote branches, making local branches hard to find. Define local branch completion after unique remote completion to show local branches first.

## TODOs:


share/completions/git.fish
```diff
@@ -1764,8 +1764,8 @@ complete -f -c git -n '__fish_git_using_command restore; and contains -- --stage
 complete -F -c git -n '__fish_git_using_command restore; and __fish_contains_opt -s s source'
 # switch options
 complete -f -c git -n __fish_git_needs_command -a switch -d 'Switch to a branch'
-complete -k -f -c git -n '__fish_git_using_command switch' -a '(__fish_git_local_branches)'
 complete -k -f -c git -n '__fish_git_using_command switch' -a '(__fish_git_unique_remote_branches)' -d 'Unique Remote Branch'
+complete -k -f -c git -n '__fish_git_using_command switch' -a '(__fish_git_local_branches)'
```
krobelus, line 1768: Great idea.
`man complete` confirms that `complete -k` are rendered in reverse order

    -k or --keep-order keeps the order of ARGUMENTS instead of sorting alphabetically. Multiple complete calls with -k result in arguments of the later ones displayed first.

Can you add the same change to other callers of __fish_git_unique_remote_branches:
- checkout
- worktree add

(BTW: this issue was already mention in a side note in bca17db96 (Preserve
git's own ordering of branch/tag completions for checkout, 2017-09-09))

The checkout ones are a bit complicated; currently we have

1. recent commits
2. unique remote branches
3. local branches
4. heads
5. tags

I think it should be

1. local branches
2. unique remote branches
3. recent commits
4. heads
5. tags

because showing branches before recent commits probably looks better for
completions on plain "git checkout" (plus I personally never use the recent
commits here).  But I'm okay with keeping "recent commits" first.

zeertzjq: Non-unique remote branches probably should be ordered after unique remote branches.

zeertzjq: Also, maybe `HEAD` should not be considered a remote branch?

PR #8343, opened 2021-10-12
add istioctl completions

kfkonrad:
## Description

This PR adds completion for `istioctl` to fish. `istioctl` provides a command (`istioctl completion fish`) to generate the completion. My commit is that script verbatim. The script is fairly generic and not likely to change. It calls a "secret"/hidden subcommand of `istioctl` called `istioctl __completion` to generate the concrete completion results.

If you don't want to have this autogenerated script in your code base, my suggested alternative for `istioctl.fish` would be this:

```
if type -q "istioctl"
  istioctl completion fish | source
end
```

Let me know which one you prefer.

Fixes issue (no connected issue)

## TODOs:


share/completions/istioctl.fish
```diff
@@ -0,0 +1,176 @@
...
+
+function __istioctl_debug
+    set -l file "$BASH_COMP_DEBUG_FILE"
+    if test -n "$file"
+        echo "$argv" >> $file
+    end
+end
+
+function __istioctl_perform_completion
+    __istioctl_debug "Starting __istioctl_perform_completion"
```
faho, line 11: Please remove the debugging stuff. It's not of interest to us.

share/completions/istioctl.fish
```diff
@@ -0,0 +1,176 @@
...
+    # For Fish, when completing a flag with an = (e.g., <program> -n=<TAB>)
+    # completions must be prefixed with the flag
+    set -l flagPrefix (string match -r -- '-.*=' "$lastArg")
+
+    __istioctl_debug "Comps: $comps"
+    __istioctl_debug "DirectiveLine: $directiveLine"
+    __istioctl_debug "flagPrefix: $flagPrefix"
+
+    for comp in $comps
+        printf "%s%s\n" "$flagPrefix" "$comp"
```
faho, line 51: This can just be

```fish
printf '%s\n' $flagPrefix$comp
```

No need for the for-loop.

share/completions/istioctl.fish
```diff
@@ -0,0 +1,176 @@
...
+    # we need to count the number of valid completions.
+    # To do so, we will filter on prefix as the completions we have received
+    # may not already be filtered so as to allow fish to match on different
+    # criteria than the prefix.
+    if test $nospace -ne 0; or test $nofiles -eq 0
+        set -l prefix (commandline -t | string escape --style=regex)
+        __istioctl_debug "prefix: $prefix"
+
+        set -l completions (string match -r -- "^$prefix.*" $__istioctl_comp_results)
+        set --global __istioctl_comp_results $completions
```
faho, line 122: This global variable is quite suspect. This function should simply print the completion candidates instead of setting a global variable.

share/completions/istioctl.fish
```diff
@@ -0,0 +1,176 @@
...
+    # Also, we use '> /dev/null 2>&1' since '&>' is not supported in older versions of fish.
+    complete --do-complete "istioctl " > /dev/null 2>&1
+end
+
+# Remove any pre-existing completions for the program since we will be handling all of them.
+complete -c istioctl -e
+
+# The call to __istioctl_prepare_completions will setup __istioctl_comp_results
+# which provides the program's completion choices.
+complete -c istioctl -n '__istioctl_prepare_completions' -f -a '$__istioctl_comp_results'
```
faho, line 175: This should just print the completion results directly - `-a '(__fish_istioctl_completions)'`.

share/completions/istioctl.fish
```diff
@@ -0,0 +1,176 @@
...
+        printf "%s%s\n" "$flagPrefix" "$comp"
+    end
+
+    printf "%s\n" "$directiveLine"
+end
+
+# This function does two things:
+# - Obtain the completions and store them in the global __istioctl_comp_results
+# - Return false if file completion should be performed
+function __istioctl_prepare_completions
```
faho, line 60: We name our internal functions `__fish_`

share/completions/istioctl.fish
```diff
@@ -0,0 +1,176 @@
...
+
+# Since Fish completions are only loaded once the user triggers them, we trigger them ourselves
+# so we can properly delete any completions provided by another script.
+# Only do this if the program can be found, or else fish may print some errors; besides,
+# the existing completions will only be loaded if the program can be found.
+if type -q "istioctl"
+    # The space after the program name is essential to trigger completion for the program
+    # and not completion of the program name itself.
+    # Also, we use '> /dev/null 2>&1' since '&>' is not supported in older versions of fish.
+    complete --do-complete "istioctl " > /dev/null 2>&1
```
faho, line 167: This entire block is completely useless in this context and should be removed.

share/completions/istioctl.fish
```diff
@@ -0,0 +1,176 @@
...
+end
+
+# Remove any pre-existing completions for the program since we will be handling all of them.
+complete -c istioctl -e
```
faho, line 171: Remove please. We're setting up the completions, no need to fiddle with anything.

share/completions/istioctl.fish
faho: That also means the condition is unnecessary.

share/completions/istioctl.fish
```diff
@@ -0,0 +1,176 @@
...
+        __istioctl_debug "Received error directive: aborting."
+        # Might as well do file completion, in case it helps
+        return 1
+    end
+
+    set -l filefilter (math (math --scale 0 $directive / $shellCompDirectiveFilterFileExt) % 2)
+    set -l dirfilter (math (math --scale 0 $directive / $shellCompDirectiveFilterDirs) % 2)
+    if test $filefilter -eq 1; or test $dirfilter -eq 1
+        __istioctl_debug "File extension filtering or directory filtering not supported"
+        # Do full file completion instead
```
faho, line 103: We no longer do this. Either the completions say files are okay or they don't.

There are very few options that allow files here - basically the configuration file options (according to https://istio.io/latest/docs/reference/commands/istioctl/). They can simply be added directly, with

```fish
complete -c istioctl -l kubeconfig -rF
```

share/completions/istioctl.fish
```diff
@@ -0,0 +1,176 @@
...
+                # In other cases, to support the "nospace" directive we trick the shell
+                # by outputting an extra, longer completion.
+                __istioctl_debug "Adding second completion to perform nospace directive"
+                set --global __istioctl_comp_results $split[1] $split[1].
+                __istioctl_debug "Completions are now: $__istioctl_comp_results"
+            end
+        end
+
+        if test $numComps -eq 0; and test $nofiles -eq 0
+            # To be consistent with bash and zsh, we only trigger file
```
faho, line 149: We're not consistent with bash and zsh.

share/completions/istioctl.fish
```diff
@@ -0,0 +1,176 @@
...
+            # We don't need descriptions anyway since there is only a single
+            # real completion which the shell will expand immediately.
+            set -l split (string split --max 1 \t $__istioctl_comp_results[1])
+
+            # Fish won't add a space if the completion ends with any
+            # of the following characters: @=/:.,
+            set -l lastChar (string sub -s -1 -- $split)
+            if not string match -r -q "[@=/:.,]" -- "$lastChar"
+                # In other cases, to support the "nospace" directive we trick the shell
+                # by outputting an extra, longer completion.
```
faho, line 141: This is super suspect. Is it useful for anything? Is there any case this matters because istioctl __complete only prints a partial argument, and then you complete again and it completes further?

share/completions/istioctl.fish
```diff
@@ -0,0 +1,176 @@
...
+    # If we want to prevent a space, or if file completion is NOT disabled,
+    # we need to count the number of valid completions.
+    # To do so, we will filter on prefix as the completions we have received
+    # may not already be filtered so as to allow fish to match on different
+    # criteria than the prefix.
+    if test $nospace -ne 0; or test $nofiles -eq 0
+        set -l prefix (commandline -t | string escape --style=regex)
+        __istioctl_debug "prefix: $prefix"
+
+        set -l completions (string match -r -- "^$prefix.*" $__istioctl_comp_results)
```
faho, line 121: Fish already does the matching on the completion candidates. This makes it worse by e.g. breaking fuzzy matching.

share/completions/istioctl.fish
```diff
@@ -0,0 +1,176 @@
...
+
+    set -l requestComp "$args[1] __complete $args[2..-1] $lastArg"
+
+    __istioctl_debug "Calling $requestComp"
+    set -l results (eval $requestComp 2> /dev/null)
+
+    # Some programs may output extra empty lines after the directive.
+    # Let's ignore them or else it will break completion.
+    # Ref: https://github.com/spf13/cobra/issues/1279
+    for line in $results[-1..1]
```
faho, line 29: This entire block can just be

```fish
set results (string trim -- $results | string match -r '.+')
```

if the issue is empty (or whitespace-only) lines.

share/completions/istioctl.fish
kfkonrad: done

share/completions/istioctl.fish
kfkonrad: actually the loop _is_ needed here. If I replace it with a single `printf` it won't complete all options and only ever complete to the first in the list

share/completions/istioctl.fish
kfkonrad: done, but be "warned" this necessitates another loop for the completion to work

share/completions/istioctl.fish
kfkonrad: done, removed the code block

share/completions/istioctl.fish
kfkonrad: done

share/completions/istioctl.fish
kfkonrad: done

share/completions/istioctl.fish
kfkonrad: I don't know, I haven't been able to trigger such a case, but I might have missed a possible argument combination somewhere

share/completions/istioctl.fish
kfkonrad: done

share/completions/istioctl.fish
krobelus: > actually the loop is needed here

I don't think so. Probably you just need to quote flagPrefix since it might be empty (?)

`printf '%s\n' "$flagPrefix"$comp`

PR #8352, opened 2021-10-14
Fix delete-key in st

LetMeByte:
In 'simple terminal' the delete key prints \e\[P by default, which ich related to the different approach the authors of st are taking on the matter of shell configuration. The main problem is the malfunction of the delete key, so we have to use a workaround like this.


LetMeByte (commented):
I know there are more bindings to be fixed in the future.

Issue #2014, opened 2015-04-06
Add forward-kill-arg and backward-kill-arg


jonhoo:
The `Alt+<backspace>` (kill last word) and `Ctrl+W` (kill last path component) family of keyboard shortcuts are very useful, particularly when editing previously issued commands, as they allow deleting chunks of command-line parameters faster than character-by-character. Beneath the hood, they are implementing using the forward/backward-kill-\* functions, and in particular, `*-kill-path-component`. However, in many cases, it would be very convenient to have a keyboard shortcut for killing an entire argument. For example, given the following input with the cursor at the end:

```
$ mplayer --no-video music/a.mp3
```

pressing `Ctrl+W` here will delete `a.mp3`, but there is no shortcut for deleting the entire last argument (`Alt+<backspace>` would delete only `mp3`). This can in many cases be useful when you have a complicated command invocation where there's a a single (but potentially long) argument that you want to change.

I propose another variation of the _-kill-_ commands should be added, namely `forward-kill-arg` and `backward-kill-arg`, that delete the next and previous _argument_ respectively. In particular, in the above example, `backward-kill-arg` would kill the full `music/a.mp3`. I'm not sure what a good shortcut would be for this (nor whether one should necessarily be provided by default), but it certainly seems like a useful feature!

aboisvert:
Love this idea.

snnw:
I like this too.  Can we have a separate bug for the `ls I\ am\ a\ file` case?  That can be fixed separately from this enhancement idea.

jonhoo:
Reported in #2016.

snnw:
Thank you for filing the separate bug.  Let's focus on the "kill entire
argument/path" use-case here.

jonhoo:
I updated the bug description to no reference the bug at all.

zx8:
Massive +1 for this!

faho:
Slightly reinterpreted as \*-kill-**token**, the backward-version of this is _really_ simple:

```fish
bind SOMETHING "commandline -rt ''"
```

while the forward version is rather hard since it's surprisingly difficult to get the _next_ token.

faho:
Wait a minute, what's the difference between this and \*-kill-bigword?

With the following commandline contents

```fish
echo $flounder >& | less; and echo $catfish
```

 backward-kill-bigword will delete like this (cursor at the very end):

```fish
echo $flounder >& | less; and echo
echo $flounder >& | less; and
echo $flounder >& | less; 
[...]

PR #8360, opened 2021-10-17
Add: `forward-path-component`, `backward-path-component`, and `forward-kill-path-component` (no preset bindings)

lgarron:
See https://github.com/fish-shell/fish-shell/issues/2014#issuecomment-945064053

I'm trying to get to know the codebase so I can implement `*-arg` versions, but I figured that these these code changes were minimal enough (because they don't need any significant new code paths)  that they were worth offering as a PR first.

## Description

This adds `forward-path-component`, `backward-path-component`, and `forward-kill-path-component` as a parallel to existing bindings:

| ➡️ | ⬅️ | ✂️ ➡️ | ⬅️ ✂️ |
|-|-|-|-|
| 🆕 `forward-path-component` | 🆕 `backward-path-component` | 🆕 `forward-kill-path-component` | `backward-kill-path-component` |
| `forward-word` | `backward-word` | `forward-kill-word` | `backward-kill-word` |
| `forward-bigword` | `backward-bigword` | `forward-kill-bigword` | `backward-kill-bigword` |


doc_src/cmds/bind.rst
```diff
@@ -87,6 +87,8 @@ The following special input functions are available:
 
 - ``backward-kill-word``, move the word to the left of the cursor to the killring. The "word" here is everything up to punctuation or whitespace.
 
+- ``backward-path-component``, move one path component to the left. A path component is everything likely to belong to a path component, i.e. not any of the following: `/={,}'\":@ |;<>&`, plus newlines and tabs.
```
lgarron, line 90: I don't think "to the left" is great here, because it conflates "left" with "backward" (which is not true of all text, although I certainly don't know how RTL text works across all commandlines).

```suggestion
- ``backward-path-component``, move one path component backward. A path component is everything likely to belong to a path component, i.e. not any of the following: `/={,}'\":@ |;<>&`, plus newlines and tabs.
```

Unfortunately, it wouldn't make sense to change this just here, since "left"/"right" are used throughout.

If I wasn't trying to match existing descriptions, I'd spell it out and use what I wrote in https://github.com/fish-shell/fish-shell/issues/2014#issuecomment-945064053 :

> Move to immediately after the last character in the current path component (i.e. right after the previous folder slash if there is one, else to the start of the word).

(Same applies the other new bindings in this PR.)

lgarron:
> If I wasn't trying to match existing descriptions, I'd spell it out and use what I wrote in #2014 (comment) :
>
> > Move to immediately after the last character in the current path component (i.e. right after the previous folder slash if there is one, else to the start of the word).

Actually, I just realized that this is *not* how `backward-kill-path-component` works, because it always stops at spaces (even when escaped or inside quotes). which makes these new bindings much, much less useful for me. :-/

I'd much rather see something that understands everyday file paths better, but that would require the code change to understand file paths/arguments more like `history-token-search-backward` (which does seem to know that spaces are part of file paths).

PR #8376, opened 2021-10-22
Allow disabling autosuggestions and syntax highlighting

### LINKED ISSUE #1363: improve the experience of using fish over mosh
Is there a way to turn off autosuggest? I find it really disturbing when using fish.

## Please note

The original request, disabling autosuggestions, is being handled in issue #102. This issue has morphed into a request for improving the experience of using fish over a [mosh](http://mosh.org) connection.

faho:
## Description

This allows essentially disabling autosuggestions and syntax highlighting.

For autosuggestions, it adds a variable $fish_autosuggestion_enabled. If set to 0, this will disable suggestions. If set to anything else (or unset), it will keep them enabled.

For highlighting, it adds a "none" theme that disables most colors. It keeps colors for the search match, selection, suggestions and pager descriptions, but sets them to either brblack or --reverse. Everything else is set to "normal". Especially the selection and pager *need* to have color or they aren't discernable at all.

The rationale here is that these features don't impact any scripting. They don't create a language dialect. They also aren't hard to implement, especially since the suggestions change is *already implemented* (to allow for interactive `read`).

So we should allow for simple matters of taste.

One issue here is that the git prompt colors currently can't be set via a theme file, ~~so most likely the git branch will still show up as green~~ (EDIT: This is incorrect, the branch isn't colored by default, please ignore the green in the below screenshots - it's left over from my config). This would either require us to rename the color variables to `fish_color_git` or similar, or allow `__fish` variables to be set in themes (this was quite locked down on purpose, so people don't try to put other things in).

Screenshots:

Screenshot_20211022_171321
Screenshot_20211022_171506
Screenshot_20211022_193324

(as a personal note, I absolutely hate using fish without suggestions - I feel like I have my arm tied behind my back!)

Fixes #1363.

## TODOs:

### DIFF
--- a/share/tools/web_config/themes/none.theme
+++ b/share/tools/web_config/themes/none.theme
@@ -1,0 +1,31 @@
+# name: '(Almost) No Colors'
+# preferred_background: 'black'
+
+fish_color_autosuggestion brblack
+fish_color_cancel
+fish_color_command
+fish_color_comment
+fish_color_cwd normal
+fish_color_cwd_root normal
+fish_color_end
+fish_color_error
+fish_color_escape
+fish_color_history_current
+fish_color_host normal
+fish_color_keyword
+fish_color_match
+fish_color_normal normal
+fish_color_operator
+fish_color_param
+fish_color_quote
+fish_color_redirection
+fish_color_search_match -r
+fish_color_selection -r
+fish_color_status normal
+fish_color_user normal
+fish_color_valid_path
+fish_pager_color_completion normal
+fish_pager_color_description brblack
+fish_pager_color_prefix '--underline'
+fish_pager_color_progress brblack
+fish_pager_color_selected_background --background=brblack
--- a/share/tools/web_config/themes/none.theme
+++ b/share/tools/web_config/themes/none.theme
@@ -16,6 +16,7 @@ fish_color_keyword
 fish_color_match
 fish_color_normal normal
 fish_color_operator
+fish_color_option
 fish_color_param
 fish_color_quote
 fish_color_redirection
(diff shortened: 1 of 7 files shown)


share/tools/web_config/themes/none.theme
```diff
@@ -0,0 +1,31 @@
+# name: '(Almost) No Colors'
+# preferred_background: 'black'
```
krobelus, line 2: I love this theme - I like to read code on my grayscale e-ink screen, and I sometimes can't read text if the background and foreground are too similar. This fixes those issues, very nice!

Actually the one issue I have in mind is that search matches (`fish_color_search_match`) are combined with other syntax highlighting (instead of overriding them), which can make them hard to read, even on a screen that already supports colors.

So if I type `echo` <kbd>Up</kbd> and the matching `echo` is in command position, the default theme will give it dark blue foreground, and as a search match it has black background background (on a light-themed terminal).
If the search match is an argument, it's better because that's lighter by default.
I wonder if we can/should disable our semantic syntax highlighting inside search matches.
I want to try if this makes things better.

share/tools/web_config/themes/none.theme
```diff
@@ -0,0 +1,31 @@
...
+fish_color_search_match -r
+fish_color_selection -r
+fish_color_status normal
+fish_color_user normal
+fish_color_valid_path
+fish_pager_color_completion normal
+fish_pager_color_description brblack
+fish_pager_color_prefix '--underline'
+fish_pager_color_progress brblack
+fish_pager_color_selected_background --background=brblack
```
krobelus, line 31: I think this should be

```suggestion
fish_pager_color_selected_background -r
```

Otherwise I can't read the selected completion (+ description) in the pager

krobelus: I'm using gnome-terminal which likes to draw bright colors a little bit darker than other terminals :confused: 

image

with `-r`

image

hmm ok to get rid of the inconsistency with the description, we could use  `set fish_pager_color_description normal` on top of that. But that removes the nice gray tone for descriptions..

krobelus (approved):

PR #8406, opened 2021-11-05
vmctl completion function call needs to be quoted

sourgrasses:
When no VMs are running, `__fish_get_vmctl_vms` returns nothing, leaving `complete` without an argument for the `a` flag, which returns an error. I put the subcommand in quotes, so there will at least always be an empty string passed.


krobelus:
Yep, looks like that was changed by accident in https://github.com/fish-shell/fish-shell/pull/6523

Issue #4982, opened 2018-05-14
revisit exit code of `while` builtin


mqudsi:
I'm not sure the current exit code for `while` when the loop evaluation condition returns non-zero makes sense. I understand that it passes through the value returned by the condition executable, but I'm not sure that a non-zero exit code accurately reflects the intent of the code, nor is particularly useful.

```fish
while foo
    if bar
        break
    end
end
```

This loop can terminate with either a zero or non-zero exit code. But it's equal to the following, which can only terminate with a zero exit code:

```fish
while true
    if not bar
        break
    end
end
```

Now if the condition failed to run (instead of ran and exited with a non-zero exit code), it would make sense for `while` to return non-zero:

```fish
while does-not-exist
end
```

Here, it would make sense for `$status` to evaluate to non-zero after the loop.

My problem is with a script or function that executes in a loop, an explicit `return/exit 0` is needed to "override" the misleading nonzero exit code after the loop, which _shouldn't_ (imho) ever be necessary unless a failure was detected and is being explicitly ignored.

faho:
How would you like it to behave?

Return 0 when the loop has been run at least once, 1 otherwise?

As best as I can tell, in bash it just always returns 0.

mqudsi:
I like the general principle of "if it's run at least once" but I feel like "if it never runs this is unexpected behavior/an error" is too big of an assumption to make and we'll just get complaints about it later (if anyone actually cares about these things). I think it should just return 1 in case there was an error evaluating the target (i.e. target does not exist).

faho:
>if it never runs this is unexpected behavior/an error

But it's not?

A return of 1 isn't "unexpected behavior". It's just a false return. It's something you have to deal with all the time, and which happens for quite trivial reasons, and which has no huge effect.

It's another way to convey information, not a big issue.

>it should just return 1 in case there was an error evaluating the target (i.e. target does not exist).

If the "target" does not exist, that would warrant an error message, not just a return value of 1.

mqudsi:
> If the "target" does not exist, that would warrant an error message, not just a return value of 1.

Well, yes and no. An error message is for the user, but the return code is for programmatic evaluation (which I understand you already cover with the "not just" in your statement). But my point is that nothing else explicitly warrants a non-zero return code for `while`.

[...]

PR #8409, opened 2021-11-06
Some enhancements to "for" and "while" loop pages

EmilySeville7cfg:
## Description

- Exit status mentioned for `for` command
- `while` manual page title change to correspond `for` command title

### DIFF
--- a/doc_src/cmds/for.rst
+++ b/doc_src/cmds/for.rst
@@ -15,6 +15,8 @@ Description
 
 ``for`` is a loop construct. It will perform the commands specified by ``COMMANDS`` multiple times. On each iteration, the local variable specified by ``VARNAME`` is assigned a new value from ``VALUES``. If ``VALUES`` is empty, ``COMMANDS`` will not be executed at all. The ``VARNAME`` is visible when the loop terminates and will contain the last value assigned to it. If ``VARNAME`` does not already exist it will be set in the local scope. For our purposes if the ``for`` block is inside a function there must be a local variable with the same name. If the ``for`` block is not nested inside a function then global and universal variables of the same name will be used if they exist.
 
+The exit status of the for loop is the exit status of the last iteration of the ``COMMANDS`` executed, or 0 if none were executed.
+
 Example
 -------
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -71,6 +71,7 @@ Scripting improvements
 - On MacOS terminals that are not granted permissions to access a folder, ``cd`` would print a spurious "rotten symlink" error, which has been corrected to "permission denied" (:issue:`8264`).
 - Performance improvements to globbing, especially on systems using glibc (by avoiding its slow iswalnum() function). In some cases (large directories with files with many numbers in the names) this almost doubles performance.
 - Since fish 3.0, for-loops would trigger a variable handler an additional time before the loop was entered. This has been corrected (:issue:`8384`).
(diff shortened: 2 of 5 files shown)


doc_src/cmds/for.rst
```diff
@@ -15,6 +15,8 @@ Description
 
 ``for`` is a loop construct. It will perform the commands specified by ``COMMANDS`` multiple times. On each iteration, the local variable specified by ``VARNAME`` is assigned a new value from ``VALUES``. If ``VALUES`` is empty, ``COMMANDS`` will not be executed at all. The ``VARNAME`` is visible when the loop terminates and will contain the last value assigned to it. If ``VARNAME`` does not already exist it will be set in the local scope. For our purposes if the ``for`` block is inside a function there must be a local variable with the same name. If the ``for`` block is not nested inside a function then global and universal variables of the same name will be used if they exist.
 
+The exit status of the for loop is the exit status of the last iteration of the ``COMMANDS`` executed, or 0 if none were executed.
```
krobelus, line 18: The "or 0 if none were executed" is not always true, for example

	fish -c 'false; for a in; end' # exits with 1.

So the `for`-statement itself does not set an exit status, unlike `while`:

	fish -c 'sh -c "exit 2"; while false; end' # exits with 0

We could drop this commit, because it should go without saying that the
last-executed command sets the exit status.

EmilySeville7cfg: > So the for-statement itself does not set an exit status, unlike while:

Is it the inconsistence must be resolved or it is intended behavior and why? (Should I create new issue about it?)

floam: It would be part two of this: https://github.com/fish-shell/fish-shell/issues/4982

krobelus: Ah right. I believe we should resolve the inconsistency to match other shells. The docs for `while` say

> The exit status of the while loop is the exit status of the last iteration of the ``COMMANDS`` executed, or 0 if none were executed. (This matches other shells and is POSIX-compatible.)

Turns out `for` loops are treated the same way in other shells, try:

    sh -c 'false; for x in; do :; done; echo $?' # prints 0

So we should probably just teach `run_for_statement()` to set the exit code to zero if no loop iteration was executed.
This matches both [bash](https://www.gnu.org/software/bash/manual/html_node/Looping-Constructs.html#Looping-Constructs) and [POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html).

No need to open an issue - we don't get paid in issues ;) - having the discussion here should be enough.

floam: I like it @krobelus.

krobelus: The fix to set $status to 0 when no loop iteration is executed seems okay but there's another edge case with empty for-loop bodies.
This loop doesn't change $status

	false
	for x in 1 2 3
	end
	# status is 1

which is weird when comparing to `for x in; end` which does (with the fix)

I guess we can just ignore this case because an empty body in a for-loop makes no sense. We could even error in that case..

krobelus: ok nevermind, `for` is just like `set` -- it doesn't touch the exit status by itself.
Documented in 0acbbfe2ca7847f8d01e36b17f7e78336d08e5c7

krobelus:
merged as 3c4969fc38d27d0000b2315d7a75aa0228366408 thanks for the patience

Issue #8391, opened 2021-10-28
Support vi-mode cursors in Foot Terminal 


dit7ya:
Currently vi-mode dynamic block and beam cursors do not work in Foot Terminal. 

As discussed in https://codeberg.org/dnkl/foot/issues/772 I am suspecting it is going to be an easy fix by just adding `foot` in the list here https://github.com/fish-shell/fish-shell/blob/b0b6a585a86cb7b8b082418d2c9cbc39d5a9f99b/share/functions/fish_vi_cursor.fish#L34-L55. I can start a PR.

```
fish --version = fish, version 3.3.1
uname -a = Linux nixos 5.14.15 #1-NixOS SMP Wed Oct 27 07:59:56 UTC 2021 x86_64 GNU/Linux
echo $TERM = foot```

faho:
I simply fixed it now.

In future: For issues this simple, if you're prepared to open a PR you can just do so - answering that we would accept a PR and then reviewing it and merging it is more work than just fixing it.

PR #8422, opened 2021-11-08
fish_vi_cursor: add support for alternative foot terminfos

dnkl:
Foot has several terminfos:

* foot - the default one
* foot-direct - 24-bit color terminfo, similar to xterm-direct (used by e.g. emacs)
* foot-extra - alternative to the ncurses provided terminfo, with a couple of extra, non-standard
capabilities
* foot-extra-direct - 24-bit color version of the above

There may also be other distro-custom terminfo names (serving the same purpose as foot-extra*)

Related to issue #8391

## TODOs:

### DIFF
--- a/share/functions/__fish_config_interactive.fish
+++ b/share/functions/__fish_config_interactive.fish
@@ -258,14 +258,16 @@ function __fish_config_interactive -d "Initializations that should be performed 
     # Notify terminals when $PWD changes (issue #906).
     # VTE based terminals, Terminal.app, iTerm.app (TODO), and foot support this.
     if not set -q FISH_UNIT_TESTS_RUNNING
-        and test 0"$VTE_VERSION" -ge 3405 -o "$TERM_PROGRAM" = Apple_Terminal -a (string match -r '\d+' 0"$TERM_PROGRAM_VERSION") -ge 309 -o "$TERM_PROGRAM" = WezTerm -o "$TERM" = foot
-        function __update_cwd_osc --on-variable PWD --description 'Notify capable terminals when $PWD changes'
-            if status --is-command-substitution || set -q INSIDE_EMACS
-                return
+        if string match -q -- 'foot*' $TERM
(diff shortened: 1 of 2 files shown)


faho:
You probably want to touch https://github.com/fish-shell/fish-shell/blob/c5530421f7de9d1cc672d5e388a660e644d6cb80/share/functions/__fish_config_interactive.fish#L261 as well. That's the only other spot we mention foot atm I can find.

dnkl:
> You probably want to touch 

Done. However, I'm very unused to fish shell scripting... let me know if there's a better way to do this.

share/functions/__fish_config_interactive.fish
```diff
@@ -258,7 +258,8 @@ function __fish_config_interactive -d "Initializations that should be performed
     # Notify terminals when $PWD changes (issue #906).
     # VTE based terminals, Terminal.app, iTerm.app (TODO), and foot support this.
     if not set -q FISH_UNIT_TESTS_RUNNING
-        and test 0"$VTE_VERSION" -ge 3405 -o "$TERM_PROGRAM" = Apple_Terminal -a (string match -r '\d+' 0"$TERM_PROGRAM_VERSION") -ge 309 -o "$TERM_PROGRAM" = WezTerm -o "$TERM" = foot
+        and string match -q -- 'foot*' $TERM
+        or test 0"$VTE_VERSION" -ge 3405 -o "$TERM_PROGRAM" = Apple_Terminal -a (string match -r '\d+' 0"$TERM_PROGRAM_VERSION") -ge 309 -o "$TERM_PROGRAM" = WezTerm
```
faho, line 262: This groups wrong - if the `not set -q` is false, the `and` will be skipped, and the `or` could still be true. (you'd have the same issue in sh with `&&` and `||`)

Tbh I'd add an inner `if`:

```fish
if not set -q FISH_UNIT_TESTS_RUNNING
    if string match...
       or test
```

faho: (note: This is extra credit. If you don't want to do it, I'll merge the original version and add this myself)

dnkl: Gotcha. Should be fixed now (anything for extra credits 🤣 - though I'm not really sure I can take credit for this...)

faho:
Thank you very much, merged!

PR #8430, opened 2021-11-11
__fish_seen_subcommand_from and __fish_seen_argument update

EmilySeville7cfg:
## Description

- `__fish_seen_subcommand.fish` added to have consistent naming with `__fish_seen_argument.fish`. `__fish_seen_subcommand_from.fish` is not removed but saved to backward compatibility.
- `-p` option added to commandline in `__fish_seen_argument.fish` (backward incompatible change)

### DIFF
--- a/share/functions/__fish_seen_argument.fish
+++ b/share/functions/__fish_seen_argument.fish
@@ -1,11 +1,11 @@
-function __fish_seen_argument
+function __fish_seen_argument --description 'Check whether argument used'
     argparse 's/short=+' 'o/old=+' 'l/long=+' -- $argv
 
-    set -l cmd (commandline -co)
+    set -l cmd (commandline -poc)
     set -e cmd[1]
     for t in $cmd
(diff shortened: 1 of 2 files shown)


ridiculousfish:
It looks like `__fish_seen_subcommand` isn't used anywhere, do you have plans for it?

EmilySeville7cfg:
Yes, I plan to use it instead of `__fish_seen_subcommand_from`. Naming consistency it's important to me. ;)

faho:
Please don't just duplicate a function.

Simply use `__fish_seen_subcommand_from`.

faho:
Merged the `seen_argument` update as 7f8679488, I do not believe we should duplicate a function to make it slightly more consistent in naming, especially not one that is as widely used in completions everywhere as this one.

If you want to pursue this, I would suggest trying to rename the lesser-used function seen_argument, and leaving a compatibility shim in place. See e.g. #8096.

PR #8433, opened 2021-11-11
fish_opt completion

EmilySeville7cfg:
## Description

Completion for `fish_opt` function.

### DIFF
--- a/share/completions/fish_opt.fish
+++ b/share/completions/fish_opt.fish
@@ -1,0 +1,10 @@
+complete -c fish_opt -f
+
+complete -c fish_opt -s h -l help -d 'Show help and exit'
+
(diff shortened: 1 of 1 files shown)


share/completions/fish_opt.fish
```diff
@@ -0,0 +1,12 @@
+set -g CONDITION '! __fish_seen_argument -s r --long required-val -s o --long optional-val'
```
faho, line 1: This makes a global variable called "CONDITION", visible outside of the script. That means if you had a variable called CONDITION it would change it.

Make it local instead - `set -l CONDITION`.

faho (changes requested):

faho:
Merged, thanks!

Issue #4570, opened 2017-12-01
tilde characters in scripted tab completions are escaped (sometimes)


msteed:
```sh
msteed@bones:~> fish --version
fish, version 2.7.0
msteed@bones:~> echo $version
2.7.0
msteed@bones:~> uname -a
Linux bones 4.13.12-1-ARCH #1 SMP PREEMPT Wed Nov 8 11:54:06 CET 2017 x86_64 GNU/Linux
msteed@bones:~> echo $TERM
tmux-256color
msteed@bones:~>

```

The problem:
```sh
msteed@bones:~> sh -c 'env HOME=$(mktemp -d) fish'
Welcome to fish, the friendly interactive shell
msteed@bones /h/msteed> cd
msteed@bones ~> mkdir -p a/b
msteed@bones ~> cd a/b
msteed@bones ~/a/b> cd
msteed@bones ~> cdh a<TAB>
```
When I hit <kbd>Tab</kbd>, fish supplies this completion:
```sh
msteed@bones ~> cdh \~/a/b
```
If I accept that, fish (correctly) complains:
```sh
cd: The directory “~/a/b” does not exist
```

If I hit <kbd>Tab</kbd> with an empty directory name, i.e., `cdh <TAB>`, fish lists the completions with the `~` characters escaped, but selecting a completion un-escapes the string so the command line is formatted correctly.

Edit: this may be the problem reported in https://github.com/fish-shell/fish-shell/issues/4566

yig:
I just ran into this. Does anyone have any pointers for how I might fix this (and make a pull request)?

faho:
@yig: This is a bit of a tough one.

The problem is, basically, that fish always assumes scripted completions shouldn't contain anything that expands, so if a completion prints `$VAR` that should be escaped to `\$VAR`, or, as in this case, if it prints `~/something` that it should escape the `~`. There's an exception for if the current token already starts with a `~`, but I'm not sure why and if that is intentional.

That seems to be the right thing in most circumstances, and I don't think changing the default is workable (e.g. is there a case where git expects an argument starting with `~`?).

Or it might be possible to just not escape `~`?

Or maybe expand the `~` to the value of $HOME?

It's possible that this could be done with an option to `complete` - `complete --unescaped` that would then have to be used in the `cdh` completions (and everywhere else where it's applicable).

But really, this requires looking at what completions commonly need and figuring out the best way to satisfy that, while not breaking the existing completion scripts.

-----------------

[...]

Issue #3469, opened 2016-10-17
Escape and complete -C (for fzf completion widget)


Ambrevar:
fish 2.3.1, Arch Linux, urxvt 9.22

I am trying to write a widget to replace fish completion menu with fzf. (Awesome, right? ;) )

The idea is simple:

``` fish
## Boilerplate...
complete -C | fzf... | read result
## Some code to replace the token with $result.
```

Problem is, `complete -C` unescapes the tokens a bit too much, which makes variables indistinguishable from escaped dollars.

Example:

``` fish
set HFOO 42
echo \$HOME-$H<binding>
```

Replace `<binding>` with a binding to a function that prints out `complete -C`. Result:

```
$HOME-$HOME
$HOME-$HFOO
```

The result is unusable since there is no way for the calling function to know what should be escaped.

The normal completion does preserve the escaped token though:

```
\$HOME-$HOME
\$HOME-$HFOO
```

I think `complete -C` should do the same.

This might be related to https://github.com/fish-shell/fish-shell/issues/1127.

faho:
Yes, can reproduce.

We have a bunch of issues with escaping (because it's really confusing), but I can't find this particular one.

ViViDboarder:
I've been working on something like this with fzf for certain completions.

I put together the following [function](https://github.com/ViViDboarder/shoestrap/blob/908d4f084d6796248344aab5880dd207a322cef8/assets/default/fish/functions/__fzf_complete.fish) to use, but I see issues as well.

Not every time, but sometimes it will clear the entire command line (including prompt) from the view. Other times it will call fzf multiple times and try to do multiple completions. So far I haven't figured the issues or why they are only sporadic.

Interested to see if you've had any more luck!

Ambrevar:
Correct me if I'm wrong, but I think you are referring to a different issue, namely process management and/or file descriptor management.  fish has had several issues in that area in the past.  You might want to keep looking in the issue tracker.

The issue from the original post related to the badly escaped output of the `complete -C` command.

krobelus:
For `complete -C''\$HOME-$H`, I can imagine two possible behaviors:

1. `\$HOME-/home/johndoe`

That is, expand the completed token. This may be worthwile assuming that completions are short-lived.

2. `\$HOME-$HOME`

I have an implementation of this one at [escape-complete-C](https://github.com/krobelus/fish-shell/commits/escape-complete-C).
[...]

PR #8441, opened 2021-11-13
Don't escape tildes that come from custom completions

cstyles:
## Description

This is just a simple PR to add the `--unescaped` option suggested by @faho in https://github.com/fish-shell/fish-shell/issues/4570#issuecomment-718082497. I didn't bother adding tests/documentation because I wasn't sure if this is the best solution but if y'all are OK with this change, I can go through and flesh out the PR.

Fixes issue #4570

## TODOs:

- [x] ~~Is it OK to reuse the `-u` option that was previously `--unauthoritative` but was removed in 19112980ee5c3703e50f42a9e05781e5a32f1e88?~~
- [ ] Do we need/want a short option for this?
- [ ] Should we update bundled completions that use `__fish_complete_directories`, `__fish_complete_suffix`, etc.?

### DIFF
--- a/src/builtins/complete.cpp
+++ b/src/builtins/complete.cpp
@@ -139,6 +139,7 @@ maybe_t<int> builtin_complete(parser_t &parser, io_streams_t &streams, const wch
     wcstring_list_t path;
     wcstring_list_t wrap_targets;
     bool preserve_order = false;
+    bool escape_tildes = true;
 
     static const wchar_t *const short_options = L":a:c:p:s:l:o:d:fFrxeuAn:C::w:hk";
     static const struct woption long_options[] = {
@@ -249,6 +250,7 @@ maybe_t<int> builtin_complete(parser_t &parser, io_streams_t &streams, const wch
                 break;
             }
             case 'a': {
+                escape_tildes = false;
                 comp = w.woptarg;
                 assert(comp);
                 break;
@@ -420,6 +422,9 @@ maybe_t<int> builtin_complete(parser_t &parser, io_streams_t &streams, const wch
         int flags = COMPLETE_AUTO_SPACE;
         if (preserve_order) {
             flags |= COMPLETE_DONT_SORT;
+        }
+        if (!escape_tildes) {
+            flags |= COMPLETE_DONT_ESCAPE_TILDES;
         }
 
         if (remove) {


floam:
This doesn't need a short option.

src/builtins/complete.cpp
```diff
@@ -422,6 +423,10 @@ maybe_t<int> builtin_complete(parser_t &parser, io_streams_t &streams, const wch
             flags |= COMPLETE_DONT_SORT;
         }
 
+        if (!escape) {
+            flags |= COMPLETE_DONT_ESCAPE;
```
krobelus, line 427: Problem is, if there is a space in the filename we will no longer escape it, so we'll insert incorrect completions again.
This is a difficult issue.

- We could add a hack that replaces a leading  `\~` with `~`. I don't like that because the backslash could be legit.
- Or just not use `__fish_complete_suffix`, but that would pessimate the common case.
- Or expand the ~ to the value of $HOME, as suggested in https://github.com/fish-shell/fish-shell/issues/4570#issuecomment-718082497

Neither of those are nice.
A real fix needs to add metadata to each completion entry, because the knowledge that `~` is actually `$HOME` is lost in serialization.
Sadly this is difficult to do without breakage.

Right now each completion is serialized as

    <completion text> \t <description>

we could extend that to

    <completion text> \t <description> \t <completion text but it's already escaped>

then, if fish sees a completion with the three-column format, we just the last column, and don't escape it.

A related issue is #3469: `complete -C'\\$HOME-$HOM'` outputs `$HOME-$HOME`.

cstyles: Ah, that's a good point. If I change this to `COMPLETE_DONT_ESCAPE_TILDES` instead, it fixes that bug. Doing so essentially treats the completions as regular files completions and escapes them as such. Which in this case is what we want because the completions are output by `__fish_complete_directories` / `__fish_complete_suffix` and correspond to files.

krobelus: Yep that would work until we come up with a more general solution.

I'd like to avoid updating all call sites of `__fish_complete_suffix` plus possibly others -- fish completions should remain easy to write -- so why don't we use `COMPLETE_DONT_ESCAPE_TILDES` always (for all things passed to `complete -a`)? Since `~` is only special at the beginning of a token, I expect few wrong completions. Is there any program that expects arguments that start with a literal `~`? Probably not because shells punish that :-)

cstyles: I like that idea. I just wasn't sure about the possible consequences. I'll go ahead and push a commit with that approach.

krobelus: merged as 4a3e55f69c63b6b46ae81e0ddc2e52326c072872 BTW the `if` was not necessary because the flags are only ever used for `-a` arguments anyway

PR #8446, opened 2021-11-16
Use `cargo run --example` to get list of examples

### LINKED ISSUE #8429: `cargo run --example` completions break with nested example directories
```fish
$ fish --version 
fish, version 3.3.1

$ echo $version 
3.3.1
```

Cargo completions for `--example` seem to be a little off when there are subdirectories in the `examples` directory of a cargo project.

Example, running in a local checkout of this repository: https://github.com/bevyengine/bevy 
```fish
$ cargo run --example # <tab>
2d       asset        game         reflection  ui    
3d       audio        hello_world  scene       wasm  
android  diagnostics  input        shader      window
app      ecs          ios          tools      
```
Expected result would include `z_sort_debug`, `3d_scene`, and many more.
Directory names do not necessarily correlate to actual runnable example names, and file names do not necessarily correspond to example names, although they usually do by convention.

The problem seems to be in this function in `cargo.fish`:
```fish
function __list_cargo_examples
    if not test -d ./examples
        return
    end

    find ./examples -mindepth 1 -maxdepth 1 -type f -name "*.rs" -or -type d \
        | string replace -r './examples/(.*?)(?:.rs)?$' '$1'
end
```

I'm not sure exactly why the original implementation had `-maxdepth 1` or `-type d`.

There are two options as far as I can tell, one would be to run `cargo` itself to list the examples, as a quick hack something like this seems to work pretty well:
```fish
cargo run --example 2>&1 | tail -n+3 | string trim
```

Alternately, more of a heuristic would be to modify the `find` command to be more robust, something like:
```fish
find ./examples -mindepth 1 -type f -name "*.rs" -exec basename '{}' \; | string replace '.rs' ''
```

I can open a PR for one or the other approach, depending on what seems preferable. The first approach is probably more likely to be correct, but could break depending on implementation changes in `cargo`, the second may be incorrect depending on a given project's structure, but probably matches convention fairly well.

ian-h-chamberlain:
Fixes #8429

This behavior matches the way completions are found for `cargo run`,
`cargo test`, etc., and is more robust and correct compared to looking
at filenames.

## Description

Just use a simple `string replace` instead of trying to guess example binary names from filenames. See discussion on #8429 and https://github.com/rust-lang/cargo/pull/6505 for why this works.

## TODOs:

I don't _think_ any of the above should be necessary since this is basically just a tiny bugfix, but let me know and I can update the PR.

### DIFF
--- a/share/completions/cargo.fish
+++ b/share/completions/cargo.fish
@@ -12,17 +12,9 @@ for x in bench b build rustc t test
     complete -c cargo -x -n "__fish_seen_subcommand_from $x" -l test -a "(cargo test --test 2>&1 | string replace -rf '^\s+' '')"
 end
 
-function __list_cargo_examples
-    if not test -d ./examples
-        return
-    end
-
-    find ./examples -mindepth 1 -maxdepth 1 -type f -name "*.rs" -or -type d \
-        | string replace -r './examples/(.*?)(?:.rs)?$' '$1'
-end
 for x in bench b build r run rustc t test
     complete -c cargo -x -n "__fish_seen_subcommand_from $x" -l bin -a "(cargo run --bin 2>&1 | string replace -rf '^\s+' '')"
-    complete -c cargo -x -n "__fish_seen_subcommand_from $x" -l example -a "(__list_cargo_examples)"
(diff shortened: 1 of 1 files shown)


krobelus:
Great work, thanks!

PR #8447, opened 2021-11-16
Skip leading `command` in `__fish_man_page`

kidonng:
## Description

Both `sudo` and `command` will be followed by a program name, `__fish_man_page` should handle the latter as well.

## TODOs:


kidonng:
Hmm, unrelated to this PR though, these checks are probably not needed for files under `share`.

ridiculousfish:
Thank you!

Issue #8320, opened 2021-09-26
vi mode: support increment/decrement of the next number


d-e-s-o:
It would be great if fish supported [increment/decrement of the next number](https://vim.fandom.com/wiki/Increasing_or_decreasing_numbers) when in vi mode (which Vim by default has on Ctrl-a & Ctrl-x). At least I didn't find a function providing such support.

DevAtDawn:
hey, heres a function I made to implement this feature: @d-e-s-o @krobelus 
https://github.com/DevAtDawn/fish_increment_decrement_number

now how would I go about creating a pull request in the future to contribute, if I adjust this code, seems like it would go in fish_vi_key_bindings.fish and 
fish-shell/share/functions/ ? thanks

ridiculousfish:
You would add the new binding to [`fish_vi_key_bindings`](https://github.com/fish-shell/fish-shell/blob/master/share/functions/fish_vi_key_bindings.fish). Yes it makes sense to add your new function in `share/functions`.

DevAtDawn:
@d-e-s-o  did a full rewrite if you want to check it out / add to it https://github.com/DevAtDawn/fish-shell/tree/increment_decrement

PR #8453, opened 2021-11-17
vi mode: support increment/decrement of the next number

DevAtDawn:
## Description

use ctrl+a and ctrl+x to increment or decrement a number in vi default mode 

Fixes issue #8320 

## TODOs:
- [x]  fix incrementing individual digits in a multi-digit number. (treat the number as one and increment it)
- [x] fix incrementing negative numbers (doesn't change back to a positive/remove - operator 
- [ ] 

### DIFF
--- a/share/functions/change_number.fish
+++ b/share/functions/change_number.fish
@@ -1,28 +1,28 @@
 function change_number
     set -l cursor_position (commandline -C)
     set -l current_buffer (commandline -b)
-    set -l num 0
+    set -l count 0
     set -l split_str (string split '' $current_buffer)
-    set -l the_num (math $cursor_position + 1)
-    set -l line_char $split_str[$the_num]
-    set -l new_string ""
+    set -l char_position (math $cursor_position + 1)
+    set -l buffer_char $split_str[$char_position]
+    set new_string ""
     
-    string match -qr '^[0-9]+$' $line_char
+    string match -qr '^[0-9]+$' $buffer_char
     or return
 
     if test "$argv[1]" = increase
-        set new_num (math $line_char + 1)
+        set new_num (math $buffer_char + 1)
     else if test "$argv[1]" = decrease
-        set new_num (math $line_char - 1)
+        set new_num (math $buffer_char - 1)
     end
 
     for x in (string split '' $current_buffer)
-        if [ $num = $cursor_position ]
-            set -l new_string "$new_string$new_num"
+        if [ $count = $cursor_position ]
+            set new_string $new_string$new_num
         else
-            set -l new_string "$new_string$x"
+            set new_string $new_string$x
         end
-        set num (math $num + 1)
+        set count (math $count + 1)
     end
     
     commandline -r $new_string
--- a/share/functions/change_number.fish
+++ b/share/functions/change_number.fish
@@ -1,31 +1,121 @@
-function change_number
+function change_number 
     set -l cursor_position (commandline -C)
     set -l current_buffer (commandline -b)
     set -l count 0
-    set -l split_str (string split '' $current_buffer)
+    set -l split_str (string split '' \ $current_buffer)
+    set list_size (count $split_str)
+    set -l split_str $split_str[2..$list_size]
     set -l char_position (math $cursor_position + 1)
     set -l buffer_char $split_str[$char_position]
     set new_string ""
+    set full_num ""
+    set new_num ""
     
     string match -qr '^[0-9]+$' $buffer_char
     or return
 
-    if test "$argv[1]" = increase
-        set new_num (math $buffer_char + 1)
-    else if test "$argv[1]" = decrease
-        set new_num (math $buffer_char - 1)
+    set last_num (math $char_position + 1) 
+
+    while string match -qr '^[0-9]+$' $split_str[$last_num]
+        set last_num (math $last_num + 1)
+        set cursor_position (math $cursor_position + 1)
     end
 
-    for x in (string split '' $current_buffer)
-        if [ $count = $cursor_position ]
-            set new_string $new_string$new_num
+    set append false
+    set neg false
+    set z_count 0
+
+    for x in $split_str
+        if [ $x = '-' ]
+            set full_num $full_num$x
+        else if string match -qr '^[0-9]+$' $x
+            if [ $x = 0 ]
+                if [ $count != $cursor_position ]
+                    if not $append
+                        set z_count (math $z_count + 1)
+                    end
+                end
+            end
+            set full_num $full_num$x
+
         else
+            set new_string $new_string$full_num$x
+            set full_num ''
+            set z_count 0
+        end
+
+        if $append
             set new_string $new_string$x
         end
+            
+        if [ $count = $cursor_position ]
+            set append true
+            set math_str (math $full_num + 1)
+            if string match -qr '-' \ $math_str
+                set neg true
+            else
+                set neg false
+            end
+
+            if test "$argv[1]" = increase
+                set math_str (math $full_num + 1)
+                if string match -qr '-' \ $math_str
+                    set neg true
+                else
+                    set neg false
+                end
+
+                if [ $full_num = \-1 ]
+                    set new_num (printf "%0"(math $z_count + 0)"d" (math $full_num + 1))
+                    echo $new_num
+                else
+                    if $neg
+                        set new_num (printf "%0"(math $z_count + 2)"d" (math $full_num + 1))
+                    else
+                        set new_num (printf "%0"(math $z_count + 1)"d" (math $full_num + 1))
+                    end
+                end
+
+            else if test "$argv[1]" = decrease
+                set math_str (math $full_num - 1)
+                if string match -qr '-' \ $math_str
+                    set neg true
+                else
+                    set neg false
+                end
+
+                if $neg
+                    set new_num (printf "%0"(math $z_count + 2)"d" (math $full_num - 1))
+                else
+                    set new_num (printf "%0"(math $z_count + 1)"d" (math $full_num - 1))
+                end
+            end
+            set new_string $new_string$new_num
+            set full_num ''
+        end
         set count (math $count + 1)
+    end 
+    set new_split (string split '' \ $new_string)
+    set new_list_size (count $new_split)
+    set new_split $new_split[2..$new_list_size]
+    if [ $new_split[(math $cursor_position + 1)] = '-' ]
+        set cursor_position (math $cursor_position + 1)
+    else if string match -qr '^[0-9]+$' $new_split[$char_position] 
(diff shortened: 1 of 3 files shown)


share/functions/vi_increment_decrement.fish
```diff
@@ -0,0 +1,34 @@
+alias vi_dec 'vi_increment_decrement subtract'
+alias vi_inc 'vi_increment_decrement add'
+
+function vi_increment_decrement --description 'increment or decrement numbers with ctrl+x ctrl+a'
```
d-e-s-o, line 4: I don't believe that a function should mention key bindings. It can hint at defaults, but that is about it.

share/functions/vi_increment_decrement.fish
```diff
@@ -0,0 +1,34 @@
...
+    set lnum (commandline --cursor)
+    set lstr (commandline -b)
+    set num 0
+    set split_str (string split '' $lstr)
+    set the_num (math $lnum + 1)
+    set lchar $split_str[$the_num]
+    set new_string ""
+
+    if string match -qr '^[0-9]+$' $lchar
+        if [ $argv = 'add' ]
```
d-e-s-o, line 14: Is there a chance we can just use a `bool` or something here? A string seems wrong on many levels, but I am not very familiar with what we can work with.

share/functions/vi_increment_decrement.fish
```diff
@@ -0,0 +1,34 @@
+alias vi_dec 'vi_increment_decrement subtract'
+alias vi_inc 'vi_increment_decrement add'
+
+function vi_increment_decrement --description 'increment or decrement numbers with ctrl+x ctrl+a'
+    set lnum (commandline --cursor)
+    set lstr (commandline -b)
```
d-e-s-o, line 6: Would you mind using long options in scripts, please? Short options should only be used in command line usage, not when other people read the code.

share/functions/vi_increment_decrement.fish
```diff
@@ -0,0 +1,34 @@
...
+            else
+                set new_string (echo $new_string$x | string collect)
+            end
+            set num (math $num + 1)
+        end
+        commandline -r $new_string
+        commandline -C $lnum
+        commandline -f repaint
+    else
+        return
```
d-e-s-o, line 32: May I ask why that return is necessary? Is that some `fish` specific syntax requirement? Otherwise, why not drop it and keep content concise?

share/functions/fish_vi_key_bindings.fish
```diff
@@ -268,6 +268,12 @@ function fish_vi_key_bindings --description 'vi-like key bindings for fish'
     # in vim (and maybe in vi), <BS> deletes the changes
     # but this binding just move cursor backward, not delete the changes
     bind -s --preset -M replace -k backspace backward-char
+    
+    #
+    # increment or decrement numbers with ctrl+x ctrl+a
+    #
+    bind -s --preset -M default \ca vi_inc
```
d-e-s-o, line 275: Have you checked whether there is an overlap with an existing Ctrl-a binding somewhere? I recall seeing something.

share/functions/vi_increment_decrement.fish
```diff
@@ -0,0 +1,34 @@
+alias vi_dec 'vi_increment_decrement subtract'
+alias vi_inc 'vi_increment_decrement add'
```
d-e-s-o, line 2: Are these aliases needed to implement the functionality? I am not sure they really should be there, given that they'd probably be user visible.

d-e-s-o (commented):
Thanks for working on this change, @DevAtDawn! It's great to see that implemented.

I tried it out and I am still seeing a bit of room for getting close to actual Vim behavior before landing the change. For one, I don't think we should be incrementing individual digits in a multi-digit number. Rather, we should treat the number as one and increment it.
Probably less important, but it would also be great to have the number detection be a bit more flexible: instead of looking into directly below the cursor I'd suggest the following algorithm:
- If there is a number below the cursor work with that, but consider that case that it may begin before the cursor
- If there is no number directly below it, then search the next number and work on that (repositioning the cursor accordingly)

I think that's closer to Vim behavior and I see value in mirroring it as close as we can to leave a positive experience. Let me know what you think. I can't say much to the other points you raised in the issue and I am in no position to approve, so you will have to wait for a reviewer from the `fish` side of things. Also, I think we should be thinking about adding tests as the pull request template suggests.

share/functions/vi_increment_decrement.fish
faho: However, this breaks if you pass more or less than one argument. That should be checked above.

Use `if test "$argv[1]" = add` or `if contains -- add $argv`.

share/functions/vi_increment_decrement.fish
```diff
@@ -0,0 +1,34 @@
...
+    if string match -qr '^[0-9]+$' $lchar
+        if [ $argv = 'add' ]
+            set new_num (math $lchar + 1)
+        else if [ $argv = 'subtract' ]
+            set new_num (math $lchar - 1)
+        end
+
+        for x in (string split '' $lstr)
+            if [ $num = $lnum ]
+                set new_string (echo $new_string$new_num | string collect)
```
faho, line 22: This `string collect` seems bogus?

```fish
set new_string "$new_string$new_num"
```

share/functions/vi_increment_decrement.fish
```diff
@@ -0,0 +1,34 @@
...
+function vi_increment_decrement --description 'increment or decrement numbers with ctrl+x ctrl+a'
+    set lnum (commandline --cursor)
+    set lstr (commandline -b)
+    set num 0
+    set split_str (string split '' $lstr)
+    set the_num (math $lnum + 1)
+    set lchar $split_str[$the_num]
+    set new_string ""
+
+    if string match -qr '^[0-9]+$' $lchar
```
faho, line 13: Make this a guard instead:

```fish
string match -qr '^[0-9]+$' $lchar
or return
```

faho (changes requested):

DevAtDawn (commented):
updated

share/functions/change_number.fish
```diff
@@ -0,0 +1,31 @@
+function change_number
+    set -l cursor_position (commandline -C)
+    set -l current_buffer (commandline -b)
+    set -l count 0
+    set -l split_str (string split '' $current_buffer)
+    set -l char_position (math $cursor_position + 1)
+    set -l buffer_char $split_str[$char_position]
```
faho, line 7: This only handles a single digit. If you give it a number larger than that, it will add a "0" when it overflows.

E.g.

```
> echo 987
```

with the cursor on the "9" turns into `1087` because it adds 1 to the "9". And decreasing e.g. `echo 907` on the 0 turns it into `echo 9-17`.

I don't think that's how vim behaves in this case.

If you want, you can handle it, or I can do it.

Note: I did not start a review because the PR is still marked as "draft". If you're done, you should mark it as final.

DevAtDawn:
> Thanks for working on this change, @DevAtDawn! It's great to see that implemented.
> 
> I tried it out and I am still seeing a bit of room for getting close to actual Vim behavior before landing the change. For one, I don't think we should be incrementing individual digits in a multi-digit number. Rather, we should treat the number as one and increment it. Probably less important, but it would also be great to have the number detection be a bit more flexible: instead of looking only directly below the cursor I'd suggest the following algorithm:
> 
>     * If there is a number below the cursor work with that, but consider the case that it may begin before the cursor
> 
>     * If there is no number directly below it, then search the next number and work on that (repositioning the cursor accordingly)
> 
> 
> I think that's closer to Vim behavior and I see value in mirroring it as close as we can to leave a positive experience. Let me know what you think. I can't say much to the other points you raised in the issue and I am in no position to approve, so you will have to wait for a reviewer from the `fish` side of things. Also, I think we should be thinking about adding tests as the pull request template suggests.

updated @d-e-s-o

share/functions/change_number.fish
DevAtDawn: updated, btw could you add me to assignees thx

d-e-s-o:
Great work, @DevAtDawn! It works very well. But I found an issue when working on a string such as:
`1234ui5678`. If you attempt to increment the first number, the string between the two numbers gets multiplied every time.
Also, may I ask what is going to happen on this front? Why is this pull request back in draft stage?

DohNoh:
> Great work, @DevAtDawn! It works very well. But I found an issue when working on a string such as:
> 
> `1234ui5678`. If you attempt to increment the first number, the string between the two numbers gets multiplied every time.
> 
> Also, may I ask what is going to happen on this front? Why is this pull request back in draft stage?

Hey thanks for the feedback! And cause there's still a couple things to adjust/change (noticed a bug which needs error handling) + what you mentioned^. I was also considering greatly simplifying/re-writing the code as it is by no means elegant (still getting used to the fish Lang), but then again if it works.... ¯\_(ツ)_/¯. Am I suppose to change it from a draft after everything's complete or? And I guess we still need to add the cmd to the documentation etc, but I think I'll let someone else do that

d-e-s-o:
> Am I suppose to change it from a draft after everything's complete or?

Sure, sounds good to me. It was just not clear to me why it happened, that's why I asked.

> And I guess we still need to add the new cmd to the documentation etc, but I think I'll let someone else do that stuff

I am not sure how realistic it is that somebody else will do it, though. Typically it's done as part of the introduction and forgotten if not. But it's possible `fish` does it differently.

faho:
>Am I suppose to change it from a draft after everything's complete or?

@DevAtDawn Typically yes. I see "draft" as an indication that you're still working on it, so I put off the review because there are still open questions, especially if I don't have a lot of time.

In this case, specifically, I see you've requested my review. I gotta be honest that's barely visible on Github so it doesn't really work. (I only really see it when I've already opened this PR!)

If you're basically done, I'll try to see if I can get in a review this weekend - I recall this needs some testing.

faho:
I am extremely sorry about the delay here, but after further testing this still isn't ready.

In particular when the number changes number of digits:

> = -
    ^
Standard input (line 101): 
if [ $new_split[(math $cursor_position + 1)] = '-' ]
   ^
in function 'change_number' with arguments 'decrease'

E.g. do a `change_number decrease` on `echo 10`, with the cursor on the `10`.

The code uses an unworkable idiom:

```fish
set -l split_str (string split '' \ $current_buffer)
```

This is, as best as I can tell, an attempt at allowing a negative number to not be detected as an option. What you really want is to use `--` to signal the end of options:

```fish
set -l split_str (string split '' -- $current_buffer)
```

However, because the former adds a space to the beginning, the rest of the code relies on it and I haven't managed to make it work with `--`.

So I'm terribly sorry to say, but I will be closing this pending a rewrite.

PR #8455, opened 2021-11-18
Make universal variables global, eliminating universal variable scope

ridiculousfish:
This PR addresses the "scope confusion" aspect of universal variables, as discussed in #7317 and #7379. It is a major change, so would go in after 3.4.0 release.

Today, there is a universal variable scope above the global scope, which causes confusion:

1. A global variable may shadow a universal variable, with a different value.
2. Calling `set` at top level will modify a universal variable if it exists, otherwise create a global. It's not always obvious which one you'll get.
3. If you export a universal variable, then nested fish shell instances will inherit an environment variable, which becomes a global. This is so serious that [we have a hack](https://github.com/fish-shell/fish-shell/blob/1e84164e4ede5b873c5834dd567c87902503028d/src/env.cpp#L432) to prevent the common case.

With this change, universal variables become *global* variables that are also propagated and persisted. "Universal" is now understood as a property of a variable, like "exports". Only global variables may have the universal property set.

1. `set --universal var val` now implies global, like `set --global --universal var val`
2. `set --query --universal` does what it did before: checks for a universal variable.
3. `set --query --global` now *succeeds* for universal variables, because they are just globals. Some code (`abbr`) expected this to fail and is updated.

The most important user-facing change is that **universal variables now take precedence over environment variables.**  Previously, if you tried to "override" an environment variable with a universal variable, the env var would "win" because it goes into the global scope. Now both environment variables and universal variables are global, so the uvar will "win." I think this is desirable and will fix more than it will break, but I could be proven wrong.

I would really prefer to not stage this with a feature flag, since maintaining both code paths would be quite painful.

Some missing bits, which we may or may not want to add:

1. No way to "un-universal" a variable, except to delete it and recreate it. Logically this should be an `--ununiversal` option like `--unexport` or `--unpath` but...

2. Similarly, no direct way to query if a variable is un-universal. You would write `set -qg somevar && ! set -qU somevar`.

3. We may want to add a `--no-overwrite` option to `set` that would decline to overwrite an existing variable in the same scope. This could be persisted and would give the user fine-grained control over which universal variables "win" over environment variables.


faho:
Am I correct in understanding the following:

1. If $fish_user_paths is universal, then starting fish like

```fish
fish_user_paths= fish
```

will *not* unset it?

2. If you want to override $fish_user_paths just for the current shell, that's impossible? You could make a local variable, but that' not passed everywhere so it's probably not useful?

3. If someone manages to `set -U PATH` they need to erase it and re-set it - something like

```fish
set -l oldpath $PATH
set -e PATH
set -g PATH $oldpath
```

(and this would have to be done in all running fish sessions, because the erase hits all of them)

?

The consequence of all this is that removing an entry in $fish_user_paths from $PATH becomes more awkward - removing it manually works, until the next time something alters $fish_user_paths. Basically universal variables are now for things that you absolutely require everywhere and never want to diverge.

That's a rather large list of caveats and niggles and I'm not sure if we wouldn't be trading one set of problems for another.

Maybe if we added that explicit "now save this value" step? So you could keep a diverging copy in a shell as long as you never save it? Or is that awkward without a corresponding "now load it"?

ridiculousfish:
> If $fish_user_paths is universal, then starting fish like `fish_user_paths= fish ` will not unset it?

It's a great point. This would make universal variables even more invasive.

I thought this could be addressed by having a `--no-overwrite` flag, but you've convinced me the sense is backwards. Environment variables need to continue to have precedence by default.

The diverging copy idea might be OK. We could mark variables as "detached" and then they ignore uvar changes until the user writes `varsave` or something else. But this feels like it's just reintroducing the universal variable scope in a back door way. So maybe there's nothing that can be improved about this. :(

andrey-mishchenko:
Have you considered making it illegal to set universal variables as exportable, and warning when launching a fish instance if there is an environment variable whose name collides with that of a universal variable?

ridiculousfish:
Exported universal variables are very useful, e.g. `$EDITOR`, so we certainly want to preserve that.

andrey-mishchenko:
Interesting -- I do `set -gx EDITOR vim` in my `config.fish`. But I made that decision a long time ago before I really knew my way around fish, and particularly before I even understood what universal variables are. Is there a reason to make things like `EDITOR`, `PYTHONPATH`, etc., universal instead of setting them as `-gx` in `config.fish`?

kopischke:
> Is there a reason to make things like EDITOR, PYTHONPATH, etc., universal instead of setting them as -gx in config.fish?

Changing global state. Changes to global variables set in ˋconfig.fishˋ need that file to be edited and only take effect when that file is reloaded, i.e. effectively when (re-)launching your shell. Changes to universal variables only need a ˋsetˋ command to be applied to all running instances of fish. I think the utility of that us agreed upon by the maintainers, but in practice there are scope shadowing issues, which is why uvars being replaced by another mechanism is discussed.

src/builtin.h
```diff
@@ -44,7 +44,10 @@ struct builtin_data_t {
 #define BUILTIN_ERR_COMBO2_EXCLUSIVE _(L"%ls: %ls %ls: options cannot be used together\n")
 
 /// Error message on multiple scope levels for variables.
-#define BUILTIN_ERR_GLOCAL _(L"%ls: scope can be only one of: universal function global local\n")
+#define BUILTIN_ERR_GLOCAL _(L"%ls: scope can be only one of function, global, local\n")
```
krobelus, line 47: Maybe one way forward is to get rid of automatic persistence but make it easy to write the startup file.

A new `set --broadcast` command could eventually replace `--universal`.
`set --broadcast` uses a similar propagation logic to `--universal` but always interacts with globals, never universals.

```
$ set --broadcast fish_color_command green
Updating $fish_color_command in all running shells. 
Use this value for future shells in your startup file? [Ny]
y
echo 'set --global fish_color_command green' >>~/.config/fish/config.fish
```

This will eliminate the need for fish_user_paths:

```
$ set --broadcast --prepend PATH /opt/mycoolthing
Updating $PATH in all running shells. 
Use this value for future shells in your startup file? [Ny]
y
echo 'set --global --export --prepend PATH /opt/mycoolthing' >>~/.config/fish/config.fish
```

Observe that the exported bit is preserved.

Unfortunately this also makes it easy to add too much information to config.fish:

```
$ set --broadcast PATH /opt/mycoolthing $PATH
Updating $PATH in all running shells. 
Use this value for future shells in your startup file? [Ny]
y
echo 'set --global --export PATH /opt/mycoolthing /home/user/.local/bin /usr/local/bin /usr/bin' >>~/.config/fish/config.fish
```

Also the config file can get stuffed when using this a lot (though that's fairly easy to fix).

One advantage of this interactive command is that it avoids ever-growing variables in startup files like `set -Ua fish_user_paths /opt/mycoolthing` (although we have fixed that particular case by deduplicating `fish_user_paths`)

Sadly I don't have an idea of what to do with scripts that run `set -U` (non-interactively) and having both solutions around is not nice.

ridiculousfish:
I'm closing this for now since it needs a lot more thinking. Thank you all for the feedback.

PR #8456, opened 2021-11-18
Slightly improved ansible-galaxy completions

exploide:
## Description

Slightly improved ansible-galaxy completions. Includes subcommand completions now but no further options for subcommands or more special arguments.

While I think this improves upon the current completion file, I'm not really satisfied.

The Ansible commands support [argcomplete](https://github.com/kislyuk/argcomplete). The autogenerated fish completions seem to work okay. Not everything is great, e.g. it suggests options starting with a dash when completing arguments. But at least it is complete.
I'm not really happy duplicating the whole argument structure in a completion file when it already is modeled in the application itself.

Is there a rule of thumb when fish accepts completion files that just trigger the autogeneration engine?

## TODOs:

### DIFF
--- a/share/completions/ansible-galaxy.fish
+++ b/share/completions/ansible-galaxy.fish
@@ -1,4 +1,28 @@
-complete -c ansible-galaxy -a "delete import info init install list login remove search setup"
-complete -c ansible-galaxy -s h -l help -d "Show this help message and exit"
-complete -c ansible-galaxy -s v -l verbose -d "Verbose mode (-vvv for more, -vvvv to enable connection debugging)"
-complete -c ansible-galaxy -l version -d "Display version and exit"
+# global options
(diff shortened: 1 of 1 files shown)


ridiculousfish:
Merged, thank you!

> Is there a rule of thumb when fish accepts completion files that just trigger the autogeneration engine?

It's Ok to have auto-generated completions if they are of decent quality. We do that with [cargo](https://github.com/fish-shell/fish-shell/blob/3a56c97087e06a1c6ab99da8a44073acfc132099/share/completions/cargo.fish#L26) and [gcc](https://github.com/fish-shell/fish-shell/blob/3a56c97087e06a1c6ab99da8a44073acfc132099/share/completions/gcc.fish#L35). I don't know offhand of any completions which integrate with tools like argcomplete, I think that's mostly unexplored.

Issue #7562, opened 2020-12-20
Drop Deno/Rustup completion


kidonng:
Since Deno/Rustup uses [clap](https://github.com/clap-rs/clap), it provides built-in completions via `deno completions fish` (though users have to source the output themselves).

I'm not sure if this means we can drop the completions come with Fish, but seeing similar packages like [`bat` and `ripgrep`'s completions are removed](https://github.com/fish-shell/fish-shell/commit/5c1cf0619acbd7d5275ce94617b26285aa26a68a), I opened this issue to ask.

exploide:
Are the auto-generated build-in completions good enough? Quality of such auto-generated completions varies widely. I've seen them in some Python tools and it was horrible. In some cases they erroneously disabled completion of file names and such. On the other hand, the new completions for `podman` are really awesome. They use features recently added in the Cobra framework (Go) to generate these. How is the quality here?

If the quality is good:

I think completions should work out of the box. If the user needs to be aware of sourcing the output in some config file, that is just bad UX. Two possibilities here.

- The code that makes the completions available is shipped with the software package itself. Do you know if this is happening? Are packagers of various distributions creating a file in `vendor_completions.d` or can they be convinced to to so?
- If not, for whatever reason, fish could ship the proper file instead. This would have the advantage that completions still work out of the box but the actual completion logic can change upstream without the need to adapt something in fish.

(Just my opinion how I think completions should be handled.)

kidonng:
> I've seen them in some Python tools and it was horrible. In some cases they erroneously disabled completion of file names and such. 

Yeah, I believe this is one of the areas Fish should ship its own completions. Communities like pip and npm doesn't provide good completions usually.

> On the other hand, the new completions for `podman` are really awesome. They use features recently added in the Cobra framework (Go) to generate these. How is the quality here?

`clap`'s completion is fairly good, and it's being used by lots of popular CLI tools.

> I think completions should work out of the box. If the user needs to be aware of sourcing the output in some config file, that is just bad UX.

It's not bad UX, that's just the "official" way to load completions. But most of time users don't need to do this manually, see below.

> - The code that makes the completions available is shipped with the software package itself. Do you know if this is happening? Are packagers of various distributions creating a file in `vendor_completions.d` or can they be convinced to do so?

Most distributions/package managers already do it. They run the completion command at build time, saving the output as a completion script.

[...]

PR #8471, opened 2021-11-22
Obtain Deno completions from itself

kidonng:
## Description

This was brought up in #7562. Now that Deno's completions stay as auto-generated, I suggest just calling `deno` itself for up-to-date completions, similar to [`gh`](https://github.com/fish-shell/fish-shell/blob/master/share/completions/gh.fish) and [`kitty`](https://github.com/fish-shell/fish-shell/blob/master/share/completions/kitty.fish).

## TODOs:


ridiculousfish:
Thank you!

PR #8476, opened 2021-11-23
Run fish_indent on all non-test .fish files

IlanCosman:
Ran `fish_indent -w (string match --invert -r '^tests' **.fish)`. 
I avoided tests because they often have intentionally irregular formatting. 
`fish_indent, version 3.3.1-642-g37f385b19`


kidonng:
I wonder if we can add https://github.com/fish-actions/format-check (and maybe https://github.com/fish-actions/syntax-check) to this repository. That way we can require contributors to format instead of us doing batch updates afterwards.

faho:
>I wonder if we can add https://github.com/fish-actions/format-check (and maybe https://github.com/fish-actions/syntax-check) to this repository

That just adds a lot of noisy errors to PRs. We want to keep them simple. Formatting isn't critical, running it every so often is okay.

faho:
Merged, thanks!

PR #8477, opened 2021-11-23
Use test command instead of bracket command

IlanCosman:
VSCode find and replace params:
find: `\[ (.*) \]`
replace: `test $1`
files to include: `*.fish`
files to exclude: `tests/**`

Manually removed a few matches that were incorrect, such as the comments in `ls.fish`.
Ran `fish_indent` to clean up formatting.
Manually reviewed changes and they seem fine.

`test` is the fishier way of doing things, it's more common in the code base, and it runs a whole 1% faster by my benchmarks 😄


ridiculousfish:
I will merge this but please avoid this sort of thing in the future. A change like this adds noise and risk without any real benefit.

IlanCosman:
Fair enough, my apologies. Thanks for merging.

PR #8479, opened 2021-11-24
Fix --shell option for adduser

EmilySeville7cfg:
## Description

Hide all comments.

### DIFF
--- a/share/completions/adduser.fish
+++ b/share/completions/adduser.fish
@@ -15,7 +15,7 @@ complete -c adduser -l gid -d 'When creating a group, force the groupid to be th
 complete -c adduser -l group -d 'Create a group'
 complete -c adduser -l help -d 'Display brief instructions'
 complete -c adduser -l home -d 'Use specified directory as the user\'s home directory' -x -a '(__fish_complete_directories)'
-complete -c adduser -l shell -d 'Use shell as the user\'s login shell, rather than the default' -x -a '(cat /etc/shells)'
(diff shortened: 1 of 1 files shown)


IlanCosman:
Just to be sure: have you tried using `string` commands instead of `sed`? While there are sprinklings of `sed` here and there, we'd prefer not to add any more of them, and remove the existing ones.

EmilySeville7cfg:
Thanks for a comment! I've removed sed usage in my fix.

share/completions/adduser.fish
```diff
@@ -15,7 +15,7 @@ complete -c adduser -l gid -d 'When creating a group, force the groupid to be th
 complete -c adduser -l group -d 'Create a group'
 complete -c adduser -l help -d 'Display brief instructions'
 complete -c adduser -l home -d 'Use specified directory as the user\'s home directory' -x -a '(__fish_complete_directories)'
-complete -c adduser -l shell -d 'Use shell as the user\'s login shell, rather than the default' -x -a '(cat /etc/shells)'
+complete -c adduser -l shell -d 'Use shell as the user\'s login shell, rather than the default' -x -a '(string match --regex --invert "^\s*#" < /etc/shells)'
```
zanchey, line 18: Nice one. The `chsh` completions use `string match -r '^[^#].*'`, which I think is slightly more correct (based on FreeBSD's documentation and the source for Debian's `shadow`).

EmilySeville7cfg: Thanks. ;)

PR #8481, opened 2021-11-24
Add findstr completion

EmilySeville7cfg:
## Description

Windows `findstr` command completion.

## TODOs:


share/completions/findstr.fish
```diff
@@ -0,0 +1,50 @@
+if test "$OSTYPE" != cygwin
```
faho, line 1: If something called `findstr` exists, we should give our best guess as to how to complete it.

In this case the only thing we can guess is that it's the windows findstr, so we should try that.

Case in point: This would also be used in WSL.

Also fish never sets $OSTYPE anywhere.

In short: Please remove this check entirely.

faho:
Merged, thanks!

PR #8485, opened 2021-11-24
Add --paging-full-mode option to commandline to determine if paging is complete

thibaultmarin:
## Description

Trying to prevent the TAB key from cycling through completions, I found this question
(https://unix.stackexchange.com/questions/565196/fish-shell-disable-pager-navigation-through-tab-shift-tab) offering a way to disable TAB when in pager mode (using `bind \t 'if not commandline -P; commandline -f complete; end'`).

However, I would like to have the following functionality:
- first TAB press shows the reduced list of completion (limited to `PAGER_UNDISCLOSED_MAX_ROWS`)
- second TAB press shows all the completions (this is not the case with `bind \t 'if not commandline -P; commandline -f complete; end'`)
- subsequent TAB presses do not cycle through completions.

I can achieve this by exposing the pager's `full_disclosed` flag and storing it in the `commandline_state_t` structure.  I added an option to `commandline` (`-F`) to return this `fully_disclosed` flag and use the following for my TAB keybinding:

`bind \t 'if not commandline -F; commandline -f complete; end'`

which seems to achieve what I was looking for, without, as far as I can tell, breaking anything else.

Is there any interest in merging this change? Is my approach acceptable?

Thanks in advance for the guidance.

## TODOs:

### DIFF
--- a/src/builtins/commandline.cpp
+++ b/src/builtins/commandline.cpp
@@ -143,13 +143,14 @@ maybe_t<int> builtin_commandline(parser_t &parser, io_streams_t &streams, const 
     bool line_mode = false;
     bool search_mode = false;
     bool paging_mode = false;
+    bool paging_full_mode = false;
     bool is_valid = false;
     const wchar_t *begin = nullptr, *end = nullptr;
     const wchar_t *override_buffer = nullptr;
 
     const auto &ld = parser.libdata();
 
-    static const wchar_t *const short_options = L":abijpctforhI:CLSsP";
+    static const wchar_t *const short_options = L":abijpctforhI:CLSsPF";
     static const struct woption long_options[] = {{L"append", no_argument, nullptr, 'a'},
                                                   {L"insert", no_argument, nullptr, 'i'},
                                                   {L"replace", no_argument, nullptr, 'r'},
@@ -167,6 +168,7 @@ maybe_t<int> builtin_commandline(parser_t &parser, io_streams_t &streams, const 
                                                   {L"line", no_argument, nullptr, 'L'},
                                                   {L"search-mode", no_argument, nullptr, 'S'},
                                                   {L"paging-mode", no_argument, nullptr, 'P'},
+                                                  {L"paging-full-mode", no_argument, nullptr, 'F'},
                                                   {L"is-valid", no_argument, nullptr, 1},
                                                   {nullptr, 0, nullptr, 0}};
 
@@ -237,6 +239,10 @@ maybe_t<int> builtin_commandline(parser_t &parser, io_streams_t &streams, const 
             }
             case 'P': {
                 paging_mode = true;
+                break;
+            }
+            case 'F': {
+                paging_full_mode = true;
                 break;
             }
             case 1: {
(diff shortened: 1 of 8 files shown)


krobelus:
> I added an option to `commandline` (`-F`) to return this `fully_disclosed` flag and use the following for my TAB keybinding:
> Is there any interest in merging this change? Is my approach acceptable?

I don't understand why you want to use a binding like that.
Maybe you got used to mashing <kbd>Tab</kbd> since bash is notoriously reluctant to show completion options? In fish that's not necessary, pressing Tab exactly twice will always go into full paging mode if there are enough completions.

That said, it seems fine to add this.

src/pager.h
```diff
@@ -186,6 +186,7 @@ class pager_t {
     bool is_navigating_contents() const;
 
     // Become fully disclosed.
+    bool get_fully_disclosed() const;
```
krobelus, line 189: this would be more consistent with others

```suggestion
    bool is_fully_disclosed() const;
```

also the comment is off now, can you move the new method in a separate paragraph like the other methods above?

src/builtins/commandline.cpp
```diff
@@ -239,6 +241,10 @@ maybe_t<int> builtin_commandline(parser_t &parser, io_streams_t &streams, const
                 paging_mode = true;
                 break;
             }
+            case 'F': {
+                paging_full_mode = true;
```
krobelus, line 245: you forgot to update the check for incompatible options
at `src/builtins/commandline.cpp:332`

src/builtins/commandline.cpp
```diff
@@ -167,6 +168,7 @@ maybe_t<int> builtin_commandline(parser_t &parser, io_streams_t &streams, const
                                                   {L"line", no_argument, nullptr, 'L'},
                                                   {L"search-mode", no_argument, nullptr, 'S'},
                                                   {L"paging-mode", no_argument, nullptr, 'P'},
+                                                  {L"paging-full-mode", no_argument, nullptr, 'F'},
```
krobelus, line 171: I considered giving the existing `-P/--paging-mode` an optional argument, like `--paging-mode=full`, but that would work slightly worse with interactive completions, so yours seems fine.

Maybe just remove the short option, since it's not so obvious?

ridiculousfish:
LGTM, please fix the issues krobelus identified and I'll merge. Thank you.

thibaultmarin:
Thanks for the feedback, it helped me realize that my initial implementation was not exactly what I wanted.

I would like to avoid the slightly inconsistent behavior of the TAB key depending on the length of the completion list: short (fits within PAGER_UNDISCLOSED_MAX_ROWS) or long.  A double press on TAB will start cycling through completions when the list is short and show a full (filling the screen) list of completions when the list is long (a third TAB press would start cycling through completions).  Since I don't mind losing the ability to cycle through possibly long lists of completions using TAB, I would rather have TAB stop interacting with the completion list once it is fully shown (in other words, I would like to disable TAB if the next TAB press will start cycling through completions).

The previous version worked only for long lists.  This version relies on the `remaining_to_disclose` variable, which seems to work regardless of the size of the completion list.

Combined with the following binding:

`bind \t 'if not commandline --paging-full-mode; commandline -f complete; end'`

I get the desired behavior.

Please let me know if this is an acceptable approach.

Thanks again for the guidance.

krobelus:
That all makes sense, merged with some minor changes in ceade1629dd2af316df38debdf8b35255535dbb0

PR #8486, opened 2021-11-25
Windows command completions

EmilySeville7cfg:
## Description

Advanced command completions for 13 Windows commands.

## Main features

- Where option requires smth but path option values are suggested (if exist)
- Where there are two or more mutually exclusive options the first time one of them is suggested (to choose another user have to remove written option)

## ToDos

- [x] cmdkey
- [x] reg
- [x] setx
- [x] attrib
- [x] attributes
- [x] choice
- [x] clean
- [x] cleanmgr
- [x] cmd
- [x] comp
- [x] forfiles
- [x] schtasks
- [x] powershell
- [x] __fish_seen_argument clean-up and Windows-style option support
- [x] make concise descriptions


zanchey:
This could go in the changelog as 
` -  Windows' ``attrib``, ``attributes``, ``choice``, ``clean``, ``cleanmgr``, ``cls``, ``cmd``, ``cmdkey``, ``comp``, ``copy``, ``forfiles``, ``md``, ``pause``, ``powershell``, ``rd``, ``reg``, ``schtasks``, ``setx``, ``ver``, and ``vol`` `

floam:
Please check which of these may conflict on non-Windows systems and gate on `uname` or something.

floam:
For example zsh-completions ships completions for a totally different `reg` utility: https://github.com/zchee/zsh-completions https://github.com/genuinetools/reg

floam:
Are you certain `cls` even does anything? My understanding is there is no `cls.exe` - it is a cmd.exe builtin.

floam:
(Same for goes for VER for example...)

EmilySeville7cfg:
> Are you certain cls even does anything on fish on Windows? My understanding is there is no cls.exe - it is an internal cmd.exe builtin. Not in PATH, right?

Yes, it's my mistake I've created completion for it. I will go through [this](https://ss64.com/nt/index.html) command list and remove completions for internal cmd commands.

EmilySeville7cfg:
I've done the small research on manned.org for utils with the same names and found out:

- [clean - Removes all intermediate files produced by a previous compilation.](https://manned.org/clean.1)
- [comp - compose a message](https://manned.org/comp.1)

EmilySeville7cfg:
Everything is ready, please review this PR. ;)

share/completions/attrib.fish
```diff
@@ -0,0 +1,30 @@
...
+h\tSet the Hidden file attribute
+i\tSet the Not Content Indexed file attribute' | awk -F '\t' "{ printf \"$current_token%s\t%s\n\", \$1, \$2 }"
+    case '-*'
+      echo -e 'r\tClear the Read-only file attribute
+a\tClear the Archive file attribute
+s\tClear the System file attribute
+h\tClear the Hidden file attribute
+i\tClear the Not Content Indexed file attribute' | awk -F '\t' "{ printf \"$current_token%s\t%s\n\", \$1, \$2 }"
+    case '*'
+      if __fish_seen_argument --windows 's'
```
zanchey, line 18: If I run this, I get `__fish_seen_argument: --windows: unknown option` - probably this needs to be 
```suggestion
      if __fish_seen_argument -- --windows 's'
```

share/completions/choice.fish
```diff
@@ -0,0 +1,7 @@
+complete --command choice --no-files --arguments '/c\t"Specify the list of choices to be created"
+/n\t"Hide the list of choices"
+/cs\t"Specify that the choices are case-sensitive"
+/t\t"Specify the number of seconds"
+/d\t"Specify the default choice"
+/m\t"Specify a message"
+/?\t"Show help"'
```
zanchey, line 7: This can be rewritten (and I think is more readable) as:

```fish
complete --command choice --no-files --arguments '/c' --description "Specify the list of choices to be created"
complete --command choice --no-files --arguments '/n' --description "Hide the list of choices"
complete --command choice --no-files --arguments '/cs' --description "Specify that the choices are case-sensitive"
complete --command choice --no-files --arguments '/t' --description "Specify the number of seconds"
complete --command choice --no-files --arguments '/d' --description "Specify the default choice"
complete --command choice --no-files --arguments '/m' --description "Specify a message"
complete --command choice --no-files --arguments '/?' --description "Show help"
```

share/completions/clean.fish
```diff
@@ -0,0 +1 @@
+complete --command clean --no-files --arguments 'all\t"Specify that each and every sector on the disk is set to zero"'
```
zanchey, line 1: This file is missing a new line, which is not usually an issue but occasionally trips things up. Rather than an embedded tab I suggest using the `--description`option.

share/functions/__fish_list_windows_drives.fish
```diff
@@ -0,0 +1,3 @@
+function __fish_list_windows_drives --description 'Lists Windows drives'
+  wmic logicaldisk get name | tail +2
+end
```
zanchey, line 3: We generally call these functions "print" rather than "list".

share/functions/__fish_generate_password.fish
```diff
@@ -0,0 +1,4 @@
+function __fish_generate_password --description 'Generate password'
```
zanchey, line 1: Please do not include this function. It does not generate cryptographically-random passwords, and passwords should not be passed on the command-line anyway (I understand that some Windows programs accept them).

share/completions/cleanmgr.fish
```diff
@@ -0,0 +1,18 @@
...
+  echo -e '/d\tSpecify the drive
+/sageset\tDisplay the Disk Cleanup Settings dialog box and also create a registry keyt
+/sagerun\tRun the specified tasks that are assigned to the n value
+/tuneup\tRun /sageset and /sagerun for the same n
+/lowdisk\tRun with the default settings
+/verylowdisk\tRun with the default settings, no user prompts
+/?\tShow help'
+end
+
+complete --command cleanmgr --no-files --arguments '(__cleanmgr_generate_args)'
```
zanchey, line 18: I think this can be rewritten to be similar to other completions with `__fish_seen_subcommand` and similar. @faho might have a better idea.

share/completions/cmd.fish
```diff
@@ -0,0 +1,50 @@
+function __cmd_generate_args --description 'Function to generate args'
```
zanchey, line 1: This function does not generate the arguments, but actually completes the command-line; it would be better to call it (and similar functions) `__fish_complete_cmd` or similar.

share/completions/cmd.fish
```diff
@@ -0,0 +1,50 @@
...
+6\tYellow
+7\tWhite
+8\tGray
+9\tLight blue
+A\tLight green
+B\tLight aqua
+C\tLight red
+D\tLight purple
+E\tLight yellow
+F\tBright white' | awk -F '\t' "{ printf \"$current_token%s\t%s\n\", \$1, \$2 }"
```
zanchey, line 21: This awk pipeline is pretty gnarly! I think it could be done with Cartesian expansion instead?

share/completions/cmd.fish
```diff
@@ -0,0 +1,50 @@
...
+      echo -e '/s\tModify the treatment of string after /c or /k
+/q\tTurn the echo off
+/d\tDisable execution of AutoRun commands
+/a\tFormat internal command output as ANSI
+/u\tFormat internal command output as Unicode
+/t\tSet the background and foreground color
+/e\tManage command extensions
+/f\tManage file and directory name completion
+/v\tManage delayed environment variable expansion
+/?\tShow help'
```
zanchey, line 46: Just add these as static `complete` statements, as above.

share/completions/attributes.fish
```diff
@@ -0,0 +1,40 @@
+function __attributes_disk_generate_args
+  if not __fish_seen_subcommand_from set clear
```
zanchey, line 2: You can rewrite this outside the function as:

``fish
complete -c attributes  --no-files -n 'not __fish_seen_subcommand_from set clear' -a 'set' -d 'Set attribute'
complete -c attributes  --no-files -n 'not __fish_seen_subcommand_from set clear' -a 'clear' -d 'Clear attribute'
```

**Comment by zanchey** [comment]
Thanks for this massive contribution!

Before merging I think it needs to be brought into line with the way most of the other completions are written. That means shortening descriptions, using more `complete` statements and delegating less to functions, and making the function names more consistent (generally `__fish_print_xyz` for functions that output lists, and `__fish_complete_xyz` for functions that output item-description pairs). I've left a hodgepodge of comments above along these lines, but it applies to the majority of files.

It might be easier to just work on one or two at a time, if you like.

File: share/completions/attrib.fish (hunk 1 above)
Comment (EmilySeville7cfg) [reply share/completions/attrib.fish:18]: Everything works nice on my machine. :thinking: I've copied completion to my terminal and tested with fake ampty `attrib` executable. 

image

File: share/completions/cmd.fish (hunk 8 above)
Comment (EmilySeville7cfg) [reply share/completions/cmd.fish:21]: If I do smth like this:

```fish
echo prefix_(echo "1
2")
```

it prints `prefix_1 prefix_2`. The problem is dealing with embedded spaces:

```fish
echo prefix_(echo "1
2 after-space")
```

How can I print each item on a separate line to pass them as a suggested completions?

krobelus: `printf "$current_token"` is dangerous because `$current_token might contain a format specifier.

> How can I print each item on a separate line to pass them as a suggested completions?

You can use `printf %s\n` instead of `echo`.

Here's the full Cartesian product expansion:

```fish
printf %s\n $current_token(
            printf %s\t%s\n \
                1 Blue \
                2 Green
        )
```

ridiculousfish:
Rebased and merged ending at cdae6536fd345d57c3306cc587806713a7418e26, relnoted at 30216b2d620ec4e354149ab45b3d4d3bacba309f . Thank you!

PR #8499, opened 2021-11-29
Completions for miss_hit tools

EmilySeville7cfg:
## Description

Advanced completions for [miss_hit](https://github.com/florianschanda/miss_hit) tools.

## TODOs:
- [x] mh_style
- [x] mh_metric
- [x] mh_lint
- [x] mh_diff
- [x] mh_copyright


floam:
I'd honestly rather not take all these - could you try to see if upstream would take them first? It's quite niche and it's like five or six new scripts - meh.

EmilySeville7cfg:
> I'd honestly rather not take all these - could you try to see if upstream would take them first?

Do you mean that miss_hit project author must validate these completions first?

floam:
I would suggest a PR to the author to see if they would be willing to install completions with their software.

EmilySeville7cfg:
As I understand we should prefer include in Fish completions for more general commands than very specific such as this one. Right? If yes, completions for non-general commands must be implemented as plugins for fisher or another plugin manager.

zanchey:
> As I understand we should prefer include in Fish completions for more general commands rather than for very specific such as this one. Right? If yes, completions for non-general commands must be implemented as plugins for fisher or another plugin manager.

My approach is that completions should ideally be merged upstream for projects that are willing to carry the maintenance burden. This means that the version of the completions is in step with the version of the project as it is released. However, it does make it harder to use features present in newer versions of fish in the completions.

EmilySeville7cfg:
It makes sense. ;) What about to divide all provided functionality to system administrator's utils and anything else? I mean move everything else (not very important) to separate plugins for Fisher for instance. We can't provide everything out of the box but we can allow user to choose what he want to install. Fisher package manager could be included by default in Fish.

krobelus:
> It makes sense. ;) What about to divide all provided functionality to system administrator's utils and anything else?

Note that system administration utils are not really special here. Ideally all programs ship fish completions (or something better). We're not there yet, but we should encourage other software to ship fish completion. (Also it would be really hard to decide where to draw the line of what to keep.)

> I mean move everything else (not very important) to separate plugins for Fisher for instance.
> We can't provide everything out of the box but we can allow user to choose what he want to install. Fisher package manager could be included by default in Fish.

That would result in a worse user experience. We should not force users to use a new package manager (many users already have one).
Currently completions are shipped with fish, or they are included in your app, so the user doesn't need to do anything.
Just make it so running `cd miss_hit && make PREFIX=$PREFIX install` will install `miss_hit.fish` in `$PREFIX/share/fish/vendor_completions.d`, where $PREFIX is typically /usr/local, /usr or ~/.local.
(There's also `pkgconf fish --variable completionsdir` but this is usually inside /usr/, so using that prevents an install as unprivileged user (to a prefix like ~/.local), so I'm not sure)

EmilySeville7cfg:
I will try to move completions to `miss_hit` repo.

PR #8519, opened 2021-12-02
Use `--almost-all` in `la` function

kidonng:
## Description

`--almost-all` or `-A` option of `ls` is similar to `--all` / `-a`, except for that it doesn't show `.` or `..`. This is a more sensible default IMO.

I didn't bother to look up when this appeared across different implementations, but according to the date of [a StackOverflow answer](https://stackoverflow.com/a/22407549), this option has been existing for more than 7 years, from which we can safely assume any decent system with recent versions of fish should be shipped with a capable `ls` version.

## TODOs:

### DIFF
--- a/share/functions/la.fish
+++ b/share/functions/la.fish
@@ -2,5 +2,5 @@
 # These are very common and useful
 #
 function la --wraps ls --description "List contents of directory, including hidden files in directory using long format"
-    ls -lah $argv
+    ls -lAh $argv
 end


floam:
It should be safe. I put together this table when I was getting getting the `ls` completions perfect a while back. https://github.com/fish-shell/fish-shell/blob/master/share/completions/ls.fish#L97-L102

```
    #              [         IEEE 1003.1-2017 options         ]  [   extension options  ]
    # freebsd: ls -[ikqrs][glno][Aa][Cmx1][Fp][LH][Rd][Sft][cu]  [hbTBWwPUG ZyI,        ] [-D format] [--color=when] [file ...]
    # netbsd:  ls -[ikqrs][glno][Aa][Cmx1][Fp][L ][Rd][Sft][cu]  [hbTBWwP       XMO     ] [file ...]
    # macOS:   ls -[ikqrs][glno][Aa][Cmx1][Fp][LH][Rd][Sft][cu]  [hbTBWwPUG       Oe@   ] [file ...]
    # openbsd: ls -[ikqrs][glno][Aa][Cmx1][Fp][LH][Rd][Sft][cu]  [  TB                  ] [file ...]
    # solaris: ls -[i qrs][glno][Aa][Cmx1][Fp][LH][Rd][ ft][cu]  [hb               e@EvV] [file …]
```

floam:
Thanks!

faho:
It's [in POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/ls.html).

floam:
Yeah, it is; although… there were actually **are** some gaps in the IEEE 1003.1-2017 portion of my table believe it or not 😵‍💫

floam:
(-k, -H, -S)

exploide:
I'm not sure I like this change. I use `la` whenever I'm wondering who the owner of the current working directory is. This is now not sufficient anymore.

But you decide.

zanchey:
I suspect these were originally put together to copy the aliases in Debian's default `.bashrc`:

```
# some more ls aliases
#alias ll='ls -l'
#alias la='ls -A'
#alias l='ls -CF'
```

(note they are commented out as shipped, and have been as long as I remember)

PR #8526, opened 2021-12-03
/define option for pabcnetcclear completion

EmilySeville7cfg:
## Description

/define option for pabcnetcclear completion according to [this PR](https://github.com/pascalabcnet/pascalabcnet/pull/2565).


faho:
Thanks!

PR #8528, opened 2021-12-03
Reword documentation for `filename` in `man status`

andrey-mishchenko:
## Description

The original text of the `filename` section of `man status` is:

```
       o filename  prints  the filename of the currently running script.
         Also current-filename, -f or --current-filename.  This  depends
         on  how the script was called - if it was called via a symlink,
         the symlink will be returned, and if  the  current  script  was
         received via source it will be -.
```

I was confused by the last line of this, because if you `source` a script (like `source path/to/script.fish`) then `status filename` returns the script name, not `-`. So I reworded to:

```
       o filename  prints  the  filename of the currently-running script. Also
         current-filename, -f or --current-filename. If the current script was
         called  via  a  symlink, this will return the symlink. If the current
         script was received by piping into source, then this will return -.
```

## TODOs:

### DIFF
--- a/doc_src/cmds/status.rst
+++ b/doc_src/cmds/status.rst
@@ -54,7 +54,7 @@ The following operations (sub-commands) are available:
 
 - ``current-command`` prints the name of the currently-running function or command, like the deprecated ``_`` variable.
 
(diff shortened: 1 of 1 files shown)


faho:
Merged, thanks!

PR #8530, opened 2021-12-05
Speed up, fix fish_status_to_signal

IlanCosman:
## Description

Some super small changes to `fish_status_to_signal` that grant a massive speed boost.
Running `(count $signals)` is slow, and silly considering that we know the number already. 
And this list of signals is never, ever going to change, so it won't waste any developer time in the future.

```fish
# In config.fish
set -g list (seq 255)

function new_status_to_signal
    for arg in $argv
        if test $arg -gt 128
            set -l sigix (math $arg - 128)
            if test $sigix -le 31
                set -l signals SIGHUP SIGINT SIGQUIT SIGILL SIGTRAP SIGABRT SIGBUS \
                    SIGFPE SIGKILL SIGUSR1 SIGSEGV SIGUSR2 SIGPIPE SIGALRM \
                    SIGTERM SIGSTKFLT SIGCHLD SIGCONT SIGSTOP SIGTSTP \
                    SIGTTIN SIGTTOU SIGURG SIGXCPU SIGXFSZ SIGVTALRM \
                    SIGPROF SIGWINCH SIGIO SIGPWR SIGSYS
                echo $signals[$sigix]
            else
                echo $arg
            end
        else
            echo $arg
        end
    end
    return 0
end
```

```fish
ilan@arch ~> hyperfine --shell=fish "fish_status_to_signal \$list" "new_status_to_signal \$list"
Benchmark 1: fish_status_to_signal $list
  Time (mean ± σ):      23.4 ms ±   3.0 ms    [User: 15.6 ms, System: 10.0 ms]
  Range (min … max):    12.9 ms …  37.4 ms    103 runs
 
Benchmark 2: new_status_to_signal $list
  Time (mean ± σ):      11.7 ms ±   0.9 ms    [User: 7.3 ms, System: 6.0 ms]
  Range (min … max):     6.7 ms …  14.5 ms    177 runs
 
Summary
  'new_status_to_signal $list' ran
    1.99 ± 0.30 times faster than 'fish_status_to_signal $list'
```

### DIFF
--- a/share/functions/fish_status_to_signal.fish
+++ b/share/functions/fish_status_to_signal.fish
@@ -3,8 +3,12 @@ function fish_status_to_signal --description "Print signal name from argument (\
     for arg in $argv
         if test $arg -gt 128
             # Important: each element of $__kill_signals is a string like "10 USR1"
-            string replace -r --filter '^'(math $arg-128)' ' SIG $__kill_signals |
-                read -l signal && echo $signal || echo $arg
+            if set -l signal_names (string replace -r --filter '^'(math $arg-128)' ' SIG $__kill_signals)
+                # Some signals have multiple mnemonics. Pick the first one.
+                echo $signal_names[1]
+            else
+                echo $arg
+            end
         else
             echo $arg
         end
--- a/share/functions/fish_status_to_signal.fish
+++ b/share/functions/fish_status_to_signal.fish
@@ -1,13 +1,13 @@
 function fish_status_to_signal --description "Print signal name from argument (\$status), or just argument"
     for arg in $argv
         if test $arg -gt 128
-            set -l signals SIGHUP SIGINT SIGQUIT SIGILL SIGTRAP SIGABRT SIGBUS \
-                SIGFPE SIGKILL SIGUSR1 SIGSEGV SIGUSR2 SIGPIPE SIGALRM \
-                SIGTERM SIGSTKFLT SIGCHLD SIGCONT SIGSTOP SIGTSTP \
-                SIGTTIN SIGTTOU SIGURG SIGXCPU SIGXFSZ SIGVTALRM \
-                SIGPROF SIGWINCH SIGIO SIGPWR SIGSYS
             set -l sigix (math $arg - 128)
-            if test $sigix -le (count $signals)
+            if test $sigix -le 31 # 31 = number of named signals below
+                set -l signals SIGHUP SIGINT SIGQUIT SIGILL SIGTRAP SIGABRT SIGBUS \
+                    SIGFPE SIGKILL SIGUSR1 SIGSEGV SIGUSR2 SIGPIPE SIGALRM \
+                    SIGTERM SIGSTKFLT SIGCHLD SIGCONT SIGSTOP SIGTSTP \
+                    SIGTTIN SIGTTOU SIGURG SIGXCPU SIGXFSZ SIGVTALRM \
+                    SIGPROF SIGWINCH SIGIO SIGPWR SIGSYS
                 echo $signals[$sigix]
             else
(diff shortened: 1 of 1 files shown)


floam:
Yikes! This speedup is nice but this function is no bueno! For starters, SIGBUS (10, not 7) comes after SIGKILL and before SIGSEGV on my system.

7 is SIGEMT.

floam:
Would you like to also fix that while you are improving this function? We actually cache these - see the completions for `kill`/`killall`. What you want to do is check for `$__kill_signals` otherwise run `__fish_make_completion_signals`.

IlanCosman:
Can do 👍

IlanCosman:
Should be good now, hope I did things right 😄

share/functions/fish_status_to_signal.fish
```diff
@@ -1,17 +1,10 @@
...
-                SIGTTIN SIGTTOU SIGURG SIGXCPU SIGXFSZ SIGVTALRM \
-                SIGPROF SIGWINCH SIGIO SIGPWR SIGSYS
-            set -l sigix (math $arg - 128)
-            if test $sigix -le (count $signals)
-                echo $signals[$sigix]
-            else
-                echo $arg
-            end
+            # Important: each element of $__kill_signals is a string like "10 USR1"
+            string replace -r --filter "^$(math $arg-128) " SIG $__kill_signals |
```
krobelus, line 6: I think we should not use the "$()" feature just yet, to avoid prompt errors when upgrading

share/functions/fish_status_to_signal.fish
```diff
@@ -1,17 +1,10 @@
...
+                read -l signal && echo $signal || echo $arg
```
krobelus, line 7: I think we can even say
`string replace -rf '^'(math $arg-128)' ' SIG $__kill_signals || echo $arg`

IlanCosman: You dare challenge my fish scripting prowess?!? 😄 At first glance I thought you were right, but then I tested it and remembered the problem (which I should have written a comment about). `$__kill_signals` is not deduplicated i.e `17 CHLD` and `17 CLD` are both elements. So we need some way of getting only one match, which `read` lets us do. 

We _can_ shorten it to
```fish
string replace -r --filter '^'(math $arg-128)' ' SIG $__kill_signals | read || echo $arg
```

krobelus: right, that's subtle :o

we don't need to do it in one line, how about

```fish
if set -l signal_names (string replace -r --filter '^'(math $arg-128)' ' SIG $__kill_signals)
    # Some signals have multiple mnemonics. Pick the first one.
    echo $signal_names[1]
else
    echo $arg
end
```

IlanCosman: > we don't need to do it in one line

Blasphemy, a true shell scripter does everything on one line.

Yah sure, we don't need to do it on one line, and I can change it if you'd like. If you use command substitution though, the whole function ends up being 20% slower, because command substitution is much slower than piping in general. And if you use `read`, well then you're basically where we are now, just more verbose.

I'd say just leave it the way it is, with an appropriate comment on the purpose of `read`. But your choice 👍

krobelus: I don't expect this perfomance difference to matter.
Also I'm surprised that command substitution is slower. I think they use the same machinery as when piping to builtins like `read`. We should just fix that. A benchmark might be helpful here.

IlanCosman: Very well, I'll make it multi-line.

Benchmark: https://github.com/fish-shell/fish-shell/issues/8542

PR #8537, opened 2021-12-08
Update fish_title and __fish_pwd

EmilySeville7cfg:
## Description

- use `--description` and `--argument-names` where appropriate
- use long-style options
- simplify code according to existing comments
- use `and`/`or` instead of `&&`/`||`

Fixes issue #

## TODOs:

### DIFF
--- a/share/functions/__fish_pwd.fish
+++ b/share/functions/__fish_pwd.fish
@@ -1,8 +1,4 @@
 function __fish_pwd --description "Show current path"
-    if status test-feature regex-easyesc
-        string replace -r '^/cygdrive/(.)?' '\U$1:' -- $PWD
-    else
-        # TODO: Remove this once regex-easyesc becomes the default
-        string replace -r '^/cygdrive/(.)?' '\\\U$1:' -- $PWD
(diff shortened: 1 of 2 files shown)


floam:
Please do not keep submitting PRs with gratuitous style changes towards your preference for long options.

EmilySeville7cfg:
The problem I don't know what code style to follow, so I use my preferences.

floam:
Just assume PRs will receive some scrutiny; avoid style changes.

floam:
For completions of `my command`, please use the same output you get from `complete`. In general, leave things as they are.

krobelus:
Following the style of most completion scripts is a good starting point. Style changes can be good but we first have to understand why the old style exists.
Interestingly, `complete` prints `--no-files` while most completion scripts use `-f` but I think this is what we want.

Note that `regex-easyesc` still defaults to off, so we can't remove that code path yet.

If you have commits (like the typo fix) that only fixup earlier commits, you should [squash them](https://git-rebase.io/), so the changes become easier to read.

Issue #6695, opened 2020-03-01
docs: undocumented functions


ammgws:
In share/functions there are a bunch of fish_* functions, however the docs only list these:
```
fish_breakpoint_prompt - define the prompt when stopped at a breakpoint
fish_config - start the web-based configuration interface
fish_git_prompt - output git information for use in a prompt
fish_hg_prompt - output Mercurial information for use in a prompt
fish_indent - indenter and prettifier
fish_key_reader - explore what characters keyboard keys send
fish_mode_prompt - define the appearance of the mode indicator
fish_opt - create an option spec for the argparse command
fish_prompt - define the appearance of the command line prompt
fish_right_prompt - define the appearance of the right-side command line prompt
fish_svn_prompt - output Subversion information for use in a prompt
fish_update_completions - Update completions using manual pages
fish_vcs_prompt - output version control system information for use in a prompt
```
Putting aside those that should really be __fish_* functions, there are these functions that perhaps should be documented as well:
```
fish_clipboard_copy.fish
fish_clipboard_paste.fish
fish_default_key_bindings.fish
fish_default_mode_prompt.fish
fish_hybrid_key_bindings.fish
fish_title.fish
fish_vi_cursor.fish
fish_vi_key_bindings.fish
```

Or have they been left out on purpose?

(Came across this while fixing https://github.com/fish-shell/fish-shell/pull/6694)

faho:
>Putting aside those that should really be _fish* functions, 

They should not. They're for use by normal users, so they should just be called `fish_`.

If anything, we're overusing dunderscores.

>Or have they been left out on purpose?

No, just an oversight.

Delapouite:
Related issue about documenting `*bindings.fish` functions: https://github.com/fish-shell/fish-shell/issues/6695

ammgws:
@faho I was referring to things like the npm completion helper function that was mixed in there

charego:
Another related issue (with fewer details): https://github.com/fish-shell/fish-shell/issues/5630

andrey-mishchenko:
As of today on `master`, this is the list of functions (computed by looking at `share/fish/functions` of a clean install) which don't have associated docs (computed by looking at `share/man/man1` of a clean install):

- [ ] N_.fish
- [ ] _validate_int.fish
- [ ] contains_seq.fish
- [ ] delete-or-exit.fish
- [ ] diff.fish
- [ ] down-or-search.fish
- [ ] edit_command_buffer.fish
[...]

PR #8538, opened 2021-12-08
Reimplement prevd-or-backward-word and nextd-or-forward-word in C++

andrey-mishchenko:
## Description

Reimplement `prevd-or-backward-word` and `nextd-or-forward-word` in C++.

(Working on https://github.com/fish-shell/fish-shell/issues/6695.)

## TODOs:


andrey-mishchenko:
As suggested by @faho in https://github.com/fish-shell/fish-shell/issues/6695.

I considered adding an `int exec_subshell(const wcstring &cmd, parser_t &parser, outputter_t &outp, bool apply_exit_status);` overload to dedup the code I added in `reader.cpp` but I'm not really familiar with the codebase yet -- let me know if that is a reasonable thing to do.

andrey-mishchenko:
The failing test is line 21 of `tmux-complete.fish`, in a macOS environment. The test passes on my machine (also macOS), and running the commands in the test file manually seems to work fine on my machine (in a tmux, etc.), and I am having trouble seeing how my code is even getting executed in that test. Hmm, am I missing something?

krobelus:
Hmm yeah the tmux tests can fail in resource-constrained environments (since they rely on sleeps).
Long-term we should probably write our own synchronous mock terminal, and port most of these tests.
I'll try to monitor how often these tests fail.

src/reader.cpp
```diff
@@ -3511,15 +3513,45 @@ void reader_data_t::handle_readline_command(readline_cmd_t c, readline_loop_stat
...
+        case rl::backward_bigword:
+        case rl::prevd_or_backward_word: {
+            if (c == rl::prevd_or_backward_word && command_line.empty()) {
+              // Execute the prevd function. We capture and reprint output because prevd signals
+              // that you are already at the beginning of the directory history with a bell.
+              // TODO: This is duplicated in nextd_or_forward_word. A "exec and print to outputter"
+              // function might be helpful.
+              outputter_t &outp = outputter_t::stdoutput();
+              wcstring_list_t outputs;
+              (void)exec_subshell(L"prevd", parser(), outputs, false /* ignore exit status */);
```
krobelus, line 3525: I'm pretty sure we can just write

	(void)parser().eval(L"prevd", io_chain_t{});

that seems to inherit the output streams

share/functions/nextd-or-forward-word.fish
```diff
@@ -1,9 +0,0 @@
-function nextd-or-forward-word
-    set -l cmd (commandline)
-    if test -z "$cmd"
-        nextd
-        commandline -f repaint
-    else
-        commandline -f forward-word
```
krobelus, line 7: thanks for the great work.

I had concern about losing introspectability, but the new docs in `man bind` make up for it (much easier to find than before).

CHANGELOG.rst
```diff
@@ -49,6 +49,7 @@ Deprecations and removed features
 - ``set --query`` now returns a falsy status of 255 if given no variable names. This means ``if set -q $foo`` will not enter the if-block if ``$foo`` is empty or unset. To restore the previous behavior you would use something like ``if not set -q foo; or set -q $foo``. We do not expect anyone to have used this on purpose, any places this happens are almost certainly buggy (:issue:`8214`).
 - Mac OS X 10.9 is no longer supported. The minimum Mac version is now 10.10 "Yosemite."
 - ``_`` is now a reserved keyword (:issue:`8342`).
+- The special input functions ``nextd-or-forward-word``, ``prevd-or-backward-word``, and ``delete-or-exit`` (for readline functionality in bindings) are no longer available as normal fish functions.
```
krobelus, line 52: I guess to be precise

	New special input functions ``nextd-or-forward-word``, ``prevd-or-backward-word`` replace fish functions of the same name.
	Special input function ``delete-or-exit`` is no longer available as fish function.

Which makes me realize that this is a breaking change for whoever added custom bindings for `nextd-or-forward-word`.
I don't know if that's common (maybe try a quick github search). It might be better to keep a `function nextd-or-forward-word; commandline -f nextd-or-forward-word; end`

floam: What do you mean by custom bindings?

krobelus: Oops I somehow thought this breaks `bind X nextd-or-forward-word` but it only breaks `bind X 'nextd-or-forward-word; echo something else'`. So the risk of breakage seems very low.

andrey-mishchenko: I did have two other thoughts:

1. How much deliberation was there over the name "special input function"? IMHO it's confusing that "special input functions" are not actually (fish) functions -- what do you think about calling them "binding actions" or something else that doesn't have the word function in it?
2. AFAIK there is no way to invoke these things except in a binding (correct me if I'm wrong) -- what do people think about adding a `fish_special_input_function` (or `fish_binding_action`) command which can be used to invoke `fish_binding_action nextd-or-forward-word`, for the benefit of those few people who do somehow use these "functions" in custom bindings?

zanchey: `commandline -f` allows you to invoke input functions.

andrey-mishchenko: Ah, thanks.

Ok, the copy edits should be resolved.

src/reader.cpp
andrey-mishchenko: Thanks, that is indeed better. (We do need to be careful about restoring the status.)

andrey-mishchenko:
Again, the failing test seems racy/perfy and unrelated -- also passes on my machine:

```
 69/227 Test  #64: history_races ............................***Failed    0.92 sec
Proceeding with target execution
warning: Locking the history file took too long (0.436 seconds).
Testing low-level functionality
Testing history race conditions
Error: Item dropped from history: 2 228
Error: Item dropped from history: 0 152
Error: Item dropped from history: 0 102
Error: Item dropped from history: 3 165
Error: Item dropped from history: 1 117
Error: Item dropped from history: 0 77
Error: Expected 1025 items, but instead got 1019 items
Encountered 7 errors in low-level tests
```

krobelus:
LGTM, I only fixed the formatting, see below.

BTW fish also knows special input functions "and" and "or" so in theory this could be
`bind x nextd or backward-word` but mixing functions (nextd) and special input functions is currently not supported (there's not even an error message, we should probably fix that).

not sure about the term, function seems indeed suboptimal, because they are more like builtins.

```
$ git origin/pr-8538@{2}...pr-8538
1:  0d5a1e8ac ! 1:  971a0acca Implement nextd-or-forward-word and prevd-or-backward-word in C++
    @@ src/reader.cpp: void reader_data_t::handle_readline_command(readline_cmd_t c, re
     +        case rl::backward_bigword:
     +        case rl::prevd_or_backward_word: {
     +            if (c == rl::prevd_or_backward_word && command_line.empty()) {
    -+              auto last_statuses = parser().get_last_statuses();
    -+              (void)parser().eval(L"prevd", io_chain_t{});
    -+              parser().set_last_statuses(std::move(last_statuses));
    -+              force_exec_prompt_and_repaint = true;
    -+              inputter.queue_char(readline_cmd_t::repaint);
    -+              break;
    ++                auto last_statuses = parser().get_last_statuses();
    ++                (void)parser().eval(L"prevd", io_chain_t{});
    ++                parser().set_last_statuses(std::move(last_statuses));
    ++                force_exec_prompt_and_repaint = true;
    ++                inputter.queue_char(readline_cmd_t::repaint);
    ++                break;
     +            }
     +
                  auto move_style =
    @@ src/reader.cpp: void reader_data_t::handle_readline_command(readline_cmd_t c, re
     +        case rl::forward_bigword:
     +        case rl::nextd_or_forward_word: {
     +            if (c == rl::nextd_or_forward_word && command_line.empty()) {
    -+              auto last_statuses = parser().get_last_statuses();
    -+              (void)parser().eval(L"nextd", io_chain_t{});
    -+              parser().set_last_statuses(std::move(last_statuses));
    -+              force_exec_prompt_and_repaint = true;
    -+              inputter.queue_char(readline_cmd_t::repaint);
    -+              break;
    ++                auto last_statuses = parser().get_last_statuses();
    ++                (void)parser().eval(L"nextd", io_chain_t{});
    ++                parser().set_last_statuses(std::move(last_statuses));
    ++                force_exec_prompt_and_repaint = true;
    ++                inputter.queue_char(readline_cmd_t::repaint);
    ++                break;
     +            }
     +
                  auto move_style =
2:  d049ade86 = 2:  b7eb0f75d Remove delete-or-exit function, which has a C++ implementation
3:  daea29378 = 3:  969b59c6d Fix delete-or-exit doc wording for local consistency
4:  635da71c0 = 4:  24f9533aa Add documentation for nextd-or-forward-word and prevd-or-backward-word readline functions
5:  66544c872 = 5:  7fb287735 Fix word usage complimenting -> complementing
6:  f218ae18e = 6:  e0bb43c95 Fix underline width
7:  105bb9afc ! 7:  ee04bc857 Add CHANGELOG entry noting special input functions no longer available as fish functions
    @@ CHANGELOG.rst: Deprecations and removed features
      - ``set --query`` now returns a falsy status of 255 if given no variable names. This means ``if set -q $foo`` will not enter the if-block if ``$foo`` is empty or unset. To restore the previous behavior you would use something like ``if not set -q foo; or set -q $foo``. We do not expect anyone to have used this on purpose, any places this happens are almost certainly buggy (:issue:`8214`).
      - Mac OS X 10.9 is no longer supported. The minimum Mac version is now 10.10 "Yosemite."
      - ``_`` is now a reserved keyword (:issue:`8342`).
    -+- New special input functions ``nextd-or-forward-word`` and ``prevd-or-backward-word`` replace fish functions having the same names. The special input function ``delete-or-exit`` is no longer available as fish function.
    ++- New special input functions ``nextd-or-forward-word`` and ``prevd-or-backward-word`` replace fish functions of the same names. The special input function ``delete-or-exit`` is no longer available as fish function (:issue:`8538`).
      
      Scripting improvements
      ----------------------
```

andrey-mishchenko:
Oops, sorry about the indent and thanks for fixing it.

faho:
We should re-add the delete-or-exit function because it is the default for control-d.

That means upgrading to 3.4.0 would trigger errors if you tried closing 3.3.1 fishes with ctrl+d.

andmis:
There is a C++ special input function `delete-or-exit` also present in 3.3.1 -- why wouldn't the ctrl-d binding just invoke the C++ function?

andmis:
I just tried it and on my 3.3.1, `delete-or-exit` doesn't seem to be called. The name is overloaded (there is a special input function and a regular function). It looks (https://github.com/fish-shell/fish-shell/blob/a7b4fc3845ba0dca830553acab761169aa042cb0/src/input.cpp#L396) like `bind` prefers to take the special input functions in this case (when there is an overload).

faho:
>There is a C++ special input function delete-or-exit also present in 3.3.1 

Aaaah, okay. Looks like this was added in 3.1 already - d415350aaf050104d1a8c1d4f4b1179f202aa105.

That should be okay, then.

Issue #522, opened 2013-01-15
Write a converter or add a mode to accept regular old shell scripts


shouze:
This is the only way to adopt fish definitively for me. It's not appropriate to make the switch without that.

pik:
:+1: for this. I've been using fish for a lately but been thinking more and more about switching back to bash and zsh.

For compatibility I always have to script in bash anyways - and having to hack-around obviously simple things such as `eval "$(docker-machine env my_env)"` just because fish has it's own "approach" which is a very minor aesthetic when it comes down to it -.

sghiassy:
I could really use this as well

krader1961:
There seem to be two different use cases. Some people seem to feel we should add full support for bash, or at least sh, syntax. That would allow fish to use native functions (e.g., the `cd` override mentioned earlier) written for those shells. I am quite confident there will never be a full /bin/sh, let alone bash, emulation layer added to fish. It's highly unlikely that even a sufficiently large subset of /bin/sh syntax and behavior will be added to fish to allow arbitrary scripts to be run with confidence  they will work correctly inside fish. That proposal is a non-starter AFAICT.

The second use case is more reasonable: Make it possible to recognize changes to environment vars made by scripts like *nvm.sh* and incorporate those changes into the current fish environment. That's what the [`bass`](https://github.com/edc/bass) aims to support.

I vote for closing this issue since the first goal is impractical and the second goal appears to have already been achieved.

krader1961:
Okay, I'm closing this since the broader goal (full /bin/sh or bash compatibility) will never be implemented. It's akin to asking that Python allow Perl syntax. And the narrower goal of dealing with sh/bash scripts that set environment variables has already been solved. To the extent the latter solution is imperfect I invite people to open a new issue to discuss how to improve handling that use case.

sghiassy:
I'm not entirely sure of the conclusion to close this issue. Cross-platform compatibility to increase a software's reach has ton's of precedence:

- Parallels runs Windows programs on Mac
- Windows now runs Linux shell commands
- PCI.js runs DOS in the browser
- Ecmascripten transpiles C/C++ to JavaScript

If you can transpile C++ to JavaScript then you can certainly figure out a way to run bash scripts in Fish terminal.

Seeing as I'm not willing to do with work, I have no room to complain. But my 2cents is that Fish shell will not gain mainstream traction without such support.

Best of luck,
-- Shaheen

faho:
[...]

PR #8541, opened 2021-12-11
Allow POSIX-style loops with do/done and if statements with then/fi

krobelus:
I sometimes want to compose a for-loop in fish (so I have completion
and can test it right away) and paste the final command line into
a /bin/sh script, or send it to someone who doesn't have fish yet.
Support this by accepting the POSIX syntax for if/for/while statements.

Our motto "Don't be evil like them POSIX shells" has allowed us to
keep the fish syntax clean and easy to learn. To appeal to a broader
audience -- mostly users who are already corrupted by other shells --
why not allow certain evil constructs? Specifically, they are:

        for var in values; do command; done
        while condition; do command; done
        if condition; then command; fi

But don't document them, since that would make "if --help" more
confusing to new users. Maybe mention them somewhere else?

Also, accidentally allow

        begin; done
        function my-function; done

though those could be disabled if we care.

This introduces some breaking changes to the language: the new keywords
"do", "done", "then", "fi", plus the next commit's "elif" can no
longer be used in command position (whether as script or function).
This will probably not cause many problems because POSIX shells
already disallow these keywords as commands.


floam:
I would like to revisit much of our dogma related to POSIX. This isn't the first thing that comes to mind but I should try to be kind and see that it's progress in the right direction and does not break anything.

floam:
My immediate reaction, however, is "CORRUPTION!!"

krobelus:
Yeah, I was a bit conflicted. Every single one of these is objectively worse than our existing syntax, so this feels really dirty.
But it does help a use case, and maybe it is what the people want..

IlanCosman:
My first reaction is also "please don't". I will find out that one of these evil syntaxes, like `elif`. is .01% faster than the alternative and then be _**obligated**_ to use it 😂

ridiculousfish:
I am in favor of adding bash-isms where users encounter friction, especially to support pasted one-liners. But also we shouldn't add bash syntax wherever we can, because fish cannot realistically get to the point of running more than a fraction of bash scripts (what with `set`, parens, etc). I think loop-syntax is in this second category. It's not used very often outside of scripts, and this feature alone will probably not enable any new scripts. So for that reason I think we should not add this.

Note #522 tracks the idea of a bash-compatibility layer. There's no design there at all, should someone want to pick it up.

krobelus:
Makes sense.
I often use for-loops interactively, especially when a task is too complicated for `xargs`.
Though I can see that the scenario where this helps is a bit obscure and doesn't justify the cost / potential confusion.

PR #8549, opened 2021-12-12
Fix "Error: target STRING not available" error in fish_clipboard_paste for xclip

EmilySeville7cfg:
## Description

Talk about your changes here.

## TODOs:

### DIFF
--- a/share/functions/fish_clipboard_paste.fish
+++ b/share/functions/fish_clipboard_paste.fish
@@ -7,7 +7,7 @@ function fish_clipboard_paste
     else if set -q DISPLAY; and type -q xsel
         set data (xsel --clipboard)
     else if set -q DISPLAY; and type -q xclip
-        set data (xclip -selection clipboard -o)
+        set data (xclip -selection clipboard -o 2>/dev/null)
     else if type -q powershell.exe
         set data (powershell.exe Get-Clipboard | string trim -r -c \r)
     end


krobelus:
when does this error happen?

EmilySeville7cfg:
When I paste empty line

krobelus:
That works for me; these commands run without that error

	echo -n | xclip -selection clipboard
	xclip -selection clipboard -o

maybe there is something wrong with your environment?
Which kind of system is this on?

EmilySeville7cfg:
My environment is:

- Fish version: 3.3.1-682-g0cd3ed4b8-dirty (built from source)
- OS version: Linux Mint 20.2 Cinnamon

zanchey:
https://github.com/astrand/xclip/issues/38

krobelus:
OK as far as I can tell this error shouldn't happen for an empty line.
According to [this comment](https://github.com/neovim/neovim/issues/2642#issuecomment-163522558), it happens when the clipboard contains non-text, like an image.

It also happens on Wayland (if the clipboard hasn't been used before) but that shouldn't be the case here.

Anyway, the change looks good. I'd summarize the reason in the commit message but that's not essential

zanchey:
Merged with rebase/commit message edits here:
   1d1bee3ba..077f1f257  master -> master

Thanks!

PR #8550, opened 2021-12-13
Fix typo in __fish_tokenizer_state

EmilySeville7cfg:
## Description

It seems `str` is a typo because it's never used.

## TODOs:

### DIFF
--- a/share/functions/__fish_tokenizer_state.fish
+++ b/share/functions/__fish_tokenizer_state.fish
@@ -13,7 +13,7 @@ function __fish_tokenizer_state --description "Print the state of the tokenizer 
 
     set -l state normal
     if set -q _flag_initial_state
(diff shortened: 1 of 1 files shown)


floam:
Bug introduced 4c1173f2aeb5ae70923c4c351a3ed7749f05b712

zanchey:
Has anyone worked out how to trigger misbehaviour in `fish_clipboard_paste` as a result of this?

krobelus:
When trying to paste these three lines

```
'
'
    not quoted, so this line should be trimmed
```

we'd fail to trim spaces from line three, because we didn't realize that the single quote was already closed.

even worse, we mangled quoted data -- we trimmed line three here:

```
'
' '
    quoted, so this line should not be trimmed
```

Very nice find, thanks!

Issue #8542, opened 2021-12-11
command substitution vs read benchmarks


IlanCosman:
`fish, version 3.3.1-701-gceade1629`

```fish
set -g string (string repeat -n 1 h)
set -q list || set -U list (seq 500)

function command_substitution
    set -l foo
    for i in $list
        set foo (echo $string$i)
    end
end

function builtin_read_command
    set -l foo
    for i in $list
        echo $string$i | read foo
    end
end
```

```fish
ilan@arch ~> hyperfine --shell=fish --warmup=100 --runs=100 command_substitution builtin_read_command
Benchmark 1: command_substitution
  Time (mean ± σ):      53.1 ms ±   7.4 ms    [User: 44.2 ms, System: 18.0 ms]
  Range (min … max):    26.7 ms …  61.9 ms    100 runs
 
Benchmark 2: builtin_read_command
  Time (mean ± σ):      43.1 ms ±   2.4 ms    [User: 21.8 ms, System: 12.4 ms]
  Range (min … max):    27.5 ms …  46.6 ms    100 runs
 
Summary
  'builtin_read_command' ran
    1.23 ± 0.18 times faster than 'command_substitution'
```

`read` performance is highly dependant on string length however. If we set string like so:
```fish
set -g string (string repeat -n 100 h)
```
The following happens:
```fish
ilan@arch ~> hyperfine --shell=fish --warmup=100 --runs=100 command_substitution builtin_read_command
Benchmark 1: command_substitution
  Time (mean ± σ):      61.0 ms ±   8.7 ms    [User: 54.4 ms, System: 15.1 ms]
  Range (min … max):    30.3 ms …  67.9 ms    100 runs
 
Benchmark 2: builtin_read_command
  Time (mean ± σ):      86.3 ms ±   2.1 ms    [User: 34.7 ms, System: 42.0 ms]
  Range (min … max):    70.9 ms …  91.6 ms    100 runs
 
Summary
  'command_substitution' ran
    1.42 ± 0.20 times faster than 'builtin_read_command'
```

The equivalence point is around 32 on my machine.

faho:
The read performance is very dependent on length because read has to, in the general case, read byte-for-byte. This is so:

```fish
echo foo\nbar | begin
    read -l foo # consumes the "foo" and nothing more
    read -l bar
    echo $bar
end
```

[...]

Issue #7422, opened 2020-10-24
Job redirecting output to a fifo cannot be backgrounded


mqudsi:
As reported by @jos128 in #7411, fish can't background a process writing to a fifo:

```
mkfifo fifo
echo foo > fifo &
```

causes fish (not echo) to hang until a reader connects to `fifo` in some other instance.

The attached patch resolves this but incorrectly: fish opens the fifo and gets out of the way, but if the process being launched (such as in the above example) generates less than the maximum pipe buffer length, it will write and exit without blocking at any point, which causes the output it generated to be lost (since the kernel fifo object doesn't stick around if there are no more readers or writers connected to the fifo in question).

The problem is that fish does not fork (or better, launch a new thread) to create a job. POSIX handles the problem by having the writer's call to open(2) block until a reader connects, without that, there's no guarantee the output will not disappear into the ether if a write that does not fill the pipe buffer completes and the writing end is closed before the read end is opened. The supplied patch stops fish from hanging when opening the fifo, but does not provide an alternate mechanism that causes `echo` (or whatever) to hang in its place, so this safety mechanism to ensure the output does not disappear is completely bypassed.

```diff
commit 082d9a260789790f92c4b59f28761bbc2277a3a1 (HEAD -> master)
Author: Mahmoud Al-Qudsi <mqudsi@neosmart.net>
Date:   Sat Oct 24 12:55:41 2020 -0500

    Don't hang when redirecting to a fifo
    
    This is a bit dicey, but fish hangs any time an attempt is made to
    redirect output of a process/job to a fifo when there are no readers
    connnected to the fifo, e.g.
    
    ```
    rm -f fifo
    mkfifo fifo
    echo foo > fifo
    ```
    
    (Although it may be difficult to discern at a first blush whether it's
    fish that's hanging for the FIFO to become available or echo that's
    hanging on the write, there are several telltale signs: first, "foo" is
    almost certainly guaranteed to be smaller than the FIFO buffer, so echo
    wouldn't actually hang on the write if it had a valid fd to the fifo
    already given to it by the shell, and second, we hang even if we try to
    run the job backgrounded, e.g. `echo foo > fifo &`)
    
    The hang happens because the POSIX spec says that open(2) should hang
    when opening a fifo for write (O_WRONLY) when there are no readers that
    already have the same fifo open. O_NONBLOCK doesn't help: it is
    specified as causing the open(2) call to fail with ENXIO when O_WRONLY
    is used.
    
[...]

Issue #5714, opened 2019-03-01
`string` reading from <&0 in a piped function always reads from the tty


mqudsi:
I'm not sure if we're already tracking this as one of the other fd-related issues, but attempting to read from `&0` in a function is interpreted as `/dev/tty` and not the piped input to the function.

```fish
function foo
    string match -r 'world'  <&0
end

echo "hello world" | foo
```

Under fish 2.7.1, this hangs indefinitely (regardless of ^C, ^D, or ^Z) . Under fish 3.0 with the job management fixes, it connects to the tty and takes the input from there (until ^D).

Because of #5457, it's required to explicitly pipe input into `string`. But attempting to do so via handle manipulation just breaks everything.

The workaround is to "abuse" `cat`, which reads from the correct `stdin`:

```fish
function foo
    cat - | string match -r 'world'  <&0
end

echo "hello world" | foo
```

This isn't an issue with fds in general, because this extremely pointless example works fine:

```fish
function foo
    cat - | cat - <&0
end

echo "hello world" | foo
```

(`read -z` may be substituted for `cat -` in any of the examples above)

faho:
Sooo... `<&0` looks to me like it should be a noop, like `>&1` and `2>&2`. 

"Redirect stdin to wherever fd 0 is pointing to", is, given that stdin is fd 0, entirely redundant.

I'd wager we still set the "directly redirected" flag, though, because it's a redirection, and then we miss some redirection (see e.g #206).

Instead of requiring this weird bit of syntax (which could even be a syntax error as it's entirely redundant), why not add a `--stdin` flag to string? (I wouldn't use `-` as an indicator because `string` doesn't take files as arguments)

mqudsi:
> Sooo... `<&0` looks to me like it should be a noop, like `>&1` and `2>&2`.

Yes, it *should* be. I'm pretty sure that is not strictly the case when dealing with blocks; I don't have any concrete examples on me right now, but I'm pretty sure in the past I've had to use it within a block to get a command to read from the block's stdin rather than the terminal (but that behavior may have changed with the improvements to the job control), which is the only reason I tried it in the first place.

I'm not in favor of expanding the api by add flags to support what should be simple "if no input is provided, use stdin" behavior. This isn't restricted to `string`, the concerns about null concatenation entirely voiding the presence of non-quoted strings applies elsewhere (to fish builtins and to external commands) and I don't think we should be going to such lengths to prevent it from reading from stdin.

[...]

Issue #5635, opened 2019-02-10
Remove IO buffering when piping out of function


mqudsi:
Now that @ridiculousfish updated the io buffer code for fish 3.1.0 and we've embraced background threads, I think we can finally do something about our broken function output buffering:

```fish
function foo
    echo hello
    sleep 10
end

foo | command grep hello # no output appears until timeout has elapsed
```

The easy way out is to use the background thread to read from the buffer and write to `grep`'s `stdin`. The correct way would be to drop io buffering entirely since there's absolutely nothing stopping us from making the output pipe for `foo` be the input pipe to `grep`. The *only* place IO buffering is actually useful (and someone correct me if I'm wrong here) is command substitution.

ridiculousfish:
The main reason for my changes is to improve the architecture around process waiting, not responsiveness - though improved architecture can make other changes tractable.

I wonder if a background thread that services the pipe makes sense, or if `io_streams_t` should be enhanced to know how to write directly to an fd. I'm thinking about questions like what if the child process exits, how do we implement back pressure, etc. (We don't handle these today at all but it would be good to know where we want to go).

Note that buffering is useful when piping between two fish blocks or functions. The reason is that only one function in fish can execute at a time so the output must be fully buffered.

mqudsi:
> I wonder if a background thread that services the pipe makes sense, or if `io_streams_t` should be enhanced to know how to write directly to an fd. I'm thinking about questions like what if the child process exits, how do we implement back pressure, etc. (We don't handle these today at all but it would be good to know where we want to go).

It was my thinking that we'd deal with issues along these lines by asking ourselves "how would it behave if this entire block were an external process invoked via a single fork()", but that is contingent on the main thread being able to just launch functions/builtins/processes and move on, so that the shell itself wouldn't have to take any of that into account (i.e. if the shell were replaced with a manually constructed chain of pipes hand-routed between the individual processes start from the very first input fd to the very last output fd), but that ties back directly to the next point:

> Note that buffering is useful when piping between two fish blocks or functions. The reason is that only one function in fish can execute at a time so the output must be fully buffered.

Thanks for reminding me of that. I remember trying to see how much manipulation of global state takes place in executing a block by attempting to start a new thread to execute functions, but I couldn't figure out how to fork the parser state (this was a long time ago). The hypothetical behavior of simultaneous execution of blocks can much more easily be achieved by forking to execute blocks, but personally I would hope we never, ever go in that direction).

[...]

PR #8552, opened 2021-12-14
Read more fasterer

faho:
We can't always read in chunks because we often can't bear to
overread:

```fish
echo foo\nbar | begin
    read -l foo
    read -l bar
end
```

needs to have the first read read `foo` and the second read `bar`. So
here we can only read one byte at a time.

However, when we are directly redirected:

```fish
echo foo | read foo
```

we can, because the data is only for us anyway. The stream will be
closed after, so anything not read just goes away. Nobody else is
there to read.

This dramatically speeds up `read` of long lines through a pipe. How
much depends on the length of the line.

With lines of 5000 characters it's about 15x, with lines of 50
characters about 2x, lines of 5 characters about 1.07x.

See #8542.

## TODOs:

Tentatively scheduling this for after 3.4 because I am not entirely sure if the assumptions hold. Is there a case where we are directly redirected but the fd is used?

### DIFF
--- a/src/builtins/read.cpp
+++ b/src/builtins/read.cpp
@@ -262,7 +262,7 @@ static int read_interactive(parser_t &parser, wcstring &buff, int nchars, bool s
 /// of chars.
 ///
 /// Returns an exit status.
-static int read_in_chunks(int fd, wcstring &buff, bool split_null) {
+static int read_in_chunks(int fd, wcstring &buff, bool split_null, bool do_seek) {
     int exit_res = STATUS_CMD_OK;
     std::string str;
     bool eof = false;
@@ -284,7 +284,7 @@ static int read_in_chunks(int fd, wcstring &buff, bool split_null) {
         if (bytes_consumed < bytes_read) {
             // We found a splitter. The +1 because we need to treat the splitter as consumed, but
             // not append it to the string.
-            if (lseek(fd, bytes_consumed - bytes_read + 1, SEEK_CUR) == -1) {
+            if (do_seek && lseek(fd, bytes_consumed - bytes_read + 1, SEEK_CUR) == -1) {
                 wperror(L"lseek");
                 return STATUS_CMD_ERROR;
             }
@@ -501,8 +501,15 @@ maybe_t<int> builtin_read(parser_t &parser, io_streams_t &streams, const wchar_t
                 read_interactive(parser, buff, opts.nchars, opts.shell, opts.silent, opts.prompt,
                                  opts.right_prompt, opts.commandline, streams.stdin_fd);
         } else if (!opts.nchars && !stream_stdin_is_a_tty &&
-                   lseek(streams.stdin_fd, 0, SEEK_CUR) != -1) {
-            exit_res = read_in_chunks(streams.stdin_fd, buff, opts.split_null);
+                   (streams.stdin_is_directly_redirected || lseek(streams.stdin_fd, 0, SEEK_CUR) != -1)) {
+            // We read in chunks when we either can seek (so we put the bytes back),
+            // or we have the bytes to ourselves (because it's directly redirected).
+            //
+            // Note we skip seeking back even if we're directly redirected to a seekable stream,
+            // under the assumption that the stream will be closed soon anyway.
+            // You don't rewind VHS tapes before throwing them in the trash.
+            // TODO: Do this when nchars is set by seeking back.
+            exit_res = read_in_chunks(streams.stdin_fd, buff, opts.split_null, !streams.stdin_is_directly_redirected);
         } else {
             exit_res =
                 read_one_char_at_a_time(streams.stdin_fd, buff, opts.nchars, opts.split_null);
--- a/src/builtins/read.cpp
+++ b/src/builtins/read.cpp
@@ -494,8 +494,6 @@ maybe_t<int> builtin_read(parser_t &parser, io_streams_t &streams, const wchar_t
     do {
         buff.clear();
 
-        // TODO: Determine if the original set of conditions for interactive reads should be
-        // reinstated: if (isatty(0) && streams.stdin_fd == STDIN_FILENO && !split_null) {
         int stream_stdin_is_a_tty = isatty(streams.stdin_fd);
         if (stream_stdin_is_a_tty && !opts.split_null) {
             // Read interactively using reader_readline(). This does not support splitting on null.
--- a/tests/checks/read.fish
+++ b/tests/checks/read.fish
@@ -366,3 +366,12 @@ function-scoped-read
 # CHECK: $skamtebord[1]: |bar|
 # CHECK: $craaab: set in local scope, unexported, with 1 elements
 # CHECK: $craaab[1]: |baz|
+
+echo foo\nbar\nbaz | begin
+    read -l foo
+    read -l bar
+    cat
+    # CHECK: baz
+    echo $bar
+    # CHECK: bar
+end


faho:
>With lines of 5000 characters it's about 15x, with lines of 50
characters about 2x, lines of 5 characters about 1.07x.

The theoretical upper limit on this is 128x, because that's how many more bytes we read. I don't think it's likely to be reached because there are other factors in play and it would have to line up perfectly.

In practice the maximum I can seem to reach with an end-to-end benchmark of

```fish
for i in (seq NUM1)
    string repeat -n NUM2 a | read -l foo
end
```

is about 24x, with lines of 1 million bytes.

faho:
Alternative branch name: how-much-read-would-a-builtin-read-if-a-builtin-would-read-chunks.

And this also removes the `lseek` when we're directly redirected to a seekable fd. That's the case in `read foo < file`. Here the performance improves more for *short* lines (because it's just one thing less to do). For a 50 char line it's 10%, for a 5000 char one it's 1%. It shouldn't ever be slower.

IlanCosman:
Bwahahah, now there will never be a reason to use command substitution ever again! 😈 

> how-much-read-would-a-builtin-read-if-a-builtin-would-read-chunks

Alternative-alternative branch name, now with 33% more read + past tense read: 
`what-would-the-read-speed-of-builtin-read-be-if-builtin-read-read-chunks`

> Tentatively scheduling this for after 3.4

Noooooo, give performance _**now!**_ 😋

faho:
It's important to note that this is a microbenchmark. It's only really a thing in really really tight loops, like in that benchmark.

In an actual useful fish script most of the time the time is spent in an external command. E.g. I believe this would not speed up `fish_git_prompt` any perceptible amount because that one is stuck waiting on `git status` when it takes any significant amount of time.

And when it is stuck inside fish, it's stuck in something like functions buffering output, heavy stringery in e.g. `string` or heavy file crunching in globs. Which is why things like e.g. #5635 or 6af38962170a2b8f99ac121d73b231493566c7d3 (plus the other recent glob performance work) are much more important for actual perceptible performance.

That's why this is rather low priority, and why I would rather wait a release to ensure there's not an edge case I've missed.

src/builtins/read.cpp
```diff
@@ -494,17 +494,22 @@ maybe_t<int> builtin_read(parser_t &parser, io_streams_t &streams, const wchar_t
...
-            exit_res = read_in_chunks(streams.stdin_fd, buff, opts.split_null);
+                   (streams.stdin_is_directly_redirected || lseek(streams.stdin_fd, 0, SEEK_CUR) != -1)) {
+            // We read in chunks when we either can seek (so we put the bytes back),
+            // or we have the bytes to ourselves (because it's directly redirected).
+            //
+            // Note we skip seeking back even if we're directly redirected to a seekable stream,
+            // under the assumption that the stream will be closed soon anyway.
+            // You don't rewind VHS tapes before throwing them in the trash.
+            // TODO: Do this when nchars is set by seeking back.
+            exit_res = read_in_chunks(streams.stdin_fd, buff, opts.split_null, !streams.stdin_is_directly_redirected);
```
krobelus, line 512: Looks like a reasonable change.

> // TODO: Do this when nchars is set by seeking back.

I'm not sure how that will look like. The stream may not be seekable, so we can't actually lseek() back.
But the same assumption holds: we're draining the stream, so it's okay to spill.
Passing nchars to read_in_chunks seems ugly, so  I'd probably just disable this optimization for `read --nchars`.

> Is there a case where we are directly redirected but the fd is used?

Probably not today but in bash we can do something like

	exec 3< README.rst
	read line1 <&3
	read line2 <&3

which would break if read too much.
That's a pretty low level API. In this case it's more elegant than a loop but it's also obscure and easy to misuse.
I don't expect that fish wants to expose fds directly. Would be interesting to know where this feature is used, but I wouldn't care about that now.

faho: Okay, I believe this could be triggered via `read <&0` already. I'll have to check later.

faho: #5714. Squint and it looks like a feature.

floam: What would be a better way to do this without "abusing" exec? I would like to see the feature land here someday but agree maybe something more intuitive could be found.

floam: exec --man in ksh:

``` 
  If command is specified, then the current shell process will be replaced by
  command rather than running command and waiting for it to complete. Note that
  there is no need to use exec to enhance performance since the shell
  implicitly uses the exec mechanism internally whenever possible.

If no operands are specified, exec can be used to open or close files, or to
  manipulate file descriptors from 0 to 9 in the current shell environment
  using the standard redirection mechanism available with all commands. The
  close-on-exec flags will be set on file descriptor numbers greater than 2
  that are opened this way so that they will be closed when another program is
  invoked.
```

I think their documentation can be instructive, but my take away is that this other modality under which exec is used, is basically for manipulating fds. I would suggest if we don't ever allow for the second form, we add a builtin `fd` or something.

faho: >exec --man in ksh93:

It's *documented*, sure. But how does redirecting arbitrary file descriptors belong to the "make this other command replace me" builtin?

It's two entirely disparate things to do! That's why I call it "abuse". 

But anyway: This has very little to do with `read`, we have other issues open about it.

ridiculousfish:
This is clever but unfortunately it will run afoul of fifos as written, and perhaps some other exotic file-like objects. An example of how to reproduce a failure:

    # terminal 1
    mkfifo fifo
    echo abc\ndef > fifo

    #terminal 2
    read var1 < fifo
    read var2 < fifo

With this change, the first `read` reads both lines, so the second `read` hangs.

However I suspect it's safe with pipes, as pipes themselves buffer, so there's no way to stop a pipe from over-reading. Maybe we could make this safer by adding a `stdin_is_piped` bool, analogous to `out_is_piped` and `err_is_piped`, and only apply this if the stream is seekable or stdin is piped - ignoring file redirections.

krobelus:
> With this change, the first `read` reads both lines, so the second `read` hangs.

This is expected: the second `read` already hung before, same in bash.
This is because fifos are just named pipes.
Opening a fifo for writing will block until someone else opens it for reading (and vice versa).
After that, a normal pipe is established for exclusive communication between the reader and writer processes.

But we also have this problem with read_one_char_at_a_time()

	# terminal 1
	read a < fifo; read b < fifo; echo $a $b

	# terminal 2
	echo 1 > fifo; echo 2 > fifo

On my system, the first `read` occasionally consumes both lines.
I believe this is because
1. the OS buffers pipe reads/writes, so even if we only read the first line, we might have consumed more.
2. we don't close the fifo after the first `echo`
We could probably try to fix that but the same thing happens with bash, so I don't think it's important.
Would be great if we could somehow warn/error on such races but that seems very hard.

One interesting difference from bash is that "/bin/echo > fifo &" (without a reader) blocks in fish, while bash forks a process first.

It gets even trickier if one process uses a fifo in two places concurrently.
Commands like `read a < fifo & read b < fifo` are racy, we might try to warn/error in that case.

> I'm not sure if "directly redirected" is the absolute correct condition here

Yeah, it feels close but probably not the exact criteria

ridiculousfish:
> This is expected: the second `read` already hung before, same in bash. This is because fifos are just (named) pipes. Opening a fifo for writing will block until someone else opens it for reading (and vice versa). After that, a normal pipe is established for communication between the reader and writer processes, and no one else is supposed to tap the pipe.

Whoa, for years I had thought that FIFOs were always unbuffered at the kernel level, and that writes to a FIFO would block until there was a reader. TIL! Ok, then there's no behavior change here. I wonder why I saw inconsistent results.

src/builtins/read.cpp
mqudsi: bash's `exec n< foo` can be subbed with `mkfifo` (and I actually have fish scripts that create, pipe to, and read from FIFOs for advanced parallel processing). Would that be an issue here?

tests/checks/read.fish
```diff
@@ -366,3 +366,12 @@ function-scoped-read
 # CHECK: $skamtebord[1]: |bar|
 # CHECK: $craaab: set in local scope, unexported, with 1 elements
 # CHECK: $craaab[1]: |baz|
+
+echo foo\nbar\nbaz | begin
```
mqudsi, line 370: I would suggest leaving a comment indicating that this is specifically testing chunked vs byte-by-byte reads, just to have pity on whoever needs to track down a problem at a later date :)

mqudsi:
@krobelus 

> One interesting difference from bash is that "/bin/echo > fifo &" (without a reader) blocks in fish, while bash forks a process first.

See https://github.com/fish-shell/fish-shell/issues/7422 - I thought it was a more fundamental issue but it turns out it's just for saner error handling and we can technically patch this today.

mqudsi:
Previous discussion on `read` performance issues caused by unbuffered reading here: https://github.com/fish-shell/fish-shell/issues/4972

faho:
Merged now, for 3.5.0. If we find any issues with it it's trivial to back out.

IlanCosman:
Hmm yah, this did break stuff unfortunately 😞

```fish
begin
    echo 1
    echo 2
end | read -l --line foo bar
echo $foo $bar
```

on fish 3.4 prints `1 2`. On `fish, version 3.4.0-20-g97127c3e0`, it prints `1`.

faho:
Fixed. Yeah, `--line` is a bit of a mess. It's implemented in terms of reading one line per run through the loop n-times, and that doesn't "really" work if we're just doing direct buffering like that.

So let's just let it read byte-by-byte like it did before.

PR #8560, opened 2021-12-20
Escape double quotes when pasting

EmilySeville7cfg:
## Description

To be more consistent in pasting I made fish escape double quotes too when single/double quotes are open.

## TODOs:


faho:
We only escape what is necessary to put it into single-quotes, on purpose. The idea is that you want to *either* paste a literal string, or you want to paste *code*. A literal string is easiest to put into single-quotes, so that's when we do it.

Other than that, whatever you are pasting might fit where you chose it, and so we don't touch it.

E.g. let's say you want to print a variable, but you don't know the name yet. You write `echo "`, go to the docs, see it's `$PAGER`, copy that and paste it in. Now you end with `echo "$PAGER"`, if we did escaping you would end up with `echo "\$PAGER"`, making the `$` inert.

So: Thanks, but no.

PR #8570, opened 2021-12-21
Add basic completion for mvn -pl switch

morozov:
## Description

This patch introduces basic completion of the `-pl|--projects` switch for `mvn`. The implementation is quite naive but it's better than nothing. A more robust implementation would require either scanning the filesystem or running `mvn` which might slow down completion significantly.

Fixes issue #

## TODOs:


krobelus:
Thanks. Improving integration of java tools is an arduous task
In future we could use an XML parser here

PR #8584, opened 2021-12-25
Update fish.desktop

scardracs:
Update .desktop file (fish/Fish and added comment)

## Description

Updated Name and added a comment for DEs like Plasma

### DIFF
--- a/fish.desktop
+++ b/fish.desktop
@@ -6,6 +6,7 @@ Terminal=true
 # Other uses would have to replace these with the correct path.
 Exec=fish
 Name=fish
+Comment=The user-friendly command line shell
 Icon=fish
 # "Shell" would be the proper category, but it's reserved so we can't use it.
 Categories=ConsoleOnly;System;


floam:
Since we are touching it: is there a specific reason this file needs to be in the root?

scardracs:
> Since we are touching it: is there a specific reason this file needs to be in the root?

No, it's not really needed

krobelus:
> Since we are touching it: is there a specific reason this file needs to be in the root?

Yeah I think it's safe to move it to `share/applications` (that shouldn't break distributors since our build system installs it) but that can be a follow-up.

fish.desktop
```diff
@@ -5,7 +5,8 @@ Terminal=true
 # This file is for use in appimages, which rewrite the path.
 # Other uses would have to replace these with the correct path.
 Exec=fish
-Name=fish
+Name=Fish
```
krobelus, line 8: I don't think we use uppercase fish anywhere

floam: Also looking for a reasoning for the capital F fish.

scardracs: It was just more good-looking (at least on Plasma)

floam (approved):
LGTM

PR #8587, opened 2021-12-26
Provide completions for `complete` without `-c`

kidonng:
## Description

Since #7321 `-c` is optional for `command`, but this is not the case for its completions.

## TODOs:


floam:
`complete` completions, that is pretty meta. Looks good to me.

PR #8589, opened 2021-12-27
fish_git_prompt: be careful about git config

ridiculousfish:
fish_git_prompt may run certain git commands which may invoke certain
external programs as specified `.git/config`. Prevent this by suppressing
certain git config options by default. Introduce a variable
`$__fish_git_prompt_trusted_paths` which allows these options to be
re-enabled for a directory hierarchy (possibly `/`).

### DIFF
--- a/share/functions/fish_git_prompt.fish
+++ b/share/functions/fish_git_prompt.fish
@@ -345,18 +345,18 @@ function __fish_git_prompt_staged --description "fish_git_prompt helper, tells w
     # The "diff" functions all return > 0 if there _is_ a diff,
     # but we want to return 0 if there are staged changes.
     # So we invert the status.
-    not command git diff-index --cached --quiet HEAD -- 2>/dev/null
+    not command git -c core.fsmonitor= diff-index --cached --quiet HEAD -- 2>/dev/null
     and echo 1
 end
 
 function __fish_git_prompt_untracked --description "fish_git_prompt helper, tells whether or not the current repository has untracked files"
-    command git ls-files --others --exclude-standard --directory --no-empty-directory --error-unmatch -- :/ >/dev/null 2>&1
+    command git -c core.fsmonitor= ls-files --others --exclude-standard --directory --no-empty-directory --error-unmatch -- :/ >/dev/null 2>&1
     and echo 1
 end
 
 function __fish_git_prompt_dirty --description "fish_git_prompt helper, tells whether or not the current branch has tracked, modified files"
     # Like staged, invert the status because we want 0 to mean there are dirty files.
-    not command git diff --no-ext-diff --quiet --exit-code 2>/dev/null
+    not command git -c core.fsmonitor= diff --no-ext-diff --quiet --exit-code 2>/dev/null
     and echo 1
 end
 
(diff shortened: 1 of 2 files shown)


faho:
>Introduce a variable
$__fish_git_prompt_trusted_paths which allows these options to be
re-enabled for a directory hierarchy (possibly /).

I do not think a configuration knob is warranted here. I don't believe these git config options (core.fsmonitor, which is an optimization, and diff.external, which renders the diff) have an influence on what fish prints, and this isn't the kind of decision we want to ask of users.

share/functions/fish_git_prompt.fish
```diff
@@ -211,6 +212,15 @@ function fish_git_prompt --description "Prompt function for Git"
 
     set -l space "$___fish_git_prompt_color$___fish_git_prompt_char_stateseparator$___fish_git_prompt_color_done"
 
+    # Suppress certain config options in untrusted repos.
+    set -l extra_config -c core.fsmonitor= -c diff.external=
```
krobelus, line 216: Today, overriding diff.external shouldn't make a difference.
- Our "git diff" already uses --no-ext-diff, which has the last word here
- "git diff-index" is a plumbing command that, by default, doesn't use external diff drivers,
  only if you add --ext-diff and --patch (without --patch it doesn't look at content changes)

zanchey:
As noted by @krobelus on Gitter, what about tab-completion?

krobelus:
> I do not think a configuration knob is warranted here. [...] core.fsmonitor, which is an optimization

OK. I wasn't sure at first -- [this article](https://devblogs.microsoft.com/devops/introducing-scalar/) mentions 12 vs. 1.5 seconds for `git status` -- however unless `bash.showInformativeStatus` is set we are probably *much* faster.
So I think it's okay to remove the configuration until someone complains.

> what about tab-completion?

Right, let's make that consistent:

```diff
diff --git a/share/completions/git.fish b/share/completions/git.fish
index 6c1f80e7e..8a6090bd8 100644
--- a/share/completions/git.fish
+++ b/share/completions/git.fish
@@ -17,7 +17,7 @@ function __fish_git
     end
     # Using 'command git' to avoid interactions for aliases from git to (e.g.) hub
     # Using eval to expand ~ and variables specified on the commandline.
-    eval command git $global_args \$saved_args 2>/dev/null
+    eval command git $global_args -c core.fsmonitor= \$saved_args 2>/dev/null
 end
 
 # Print an optspec for argparse to handle git's options that are independent of any subcommand.
```

ridiculousfish:
Ok, you've convinced me to remove the knob. Our story can be "don't use `bash.showInformativeStatus` on watchman-backed repos" and if someone really needs this they can hack up `fish_git_prompt` themselves. Tab-completions I didn't think were necessary as users have to type `git` to run them, but if it's easy we may as well do them at the same time.

tests/checks/git.fish
```diff
@@ -80,3 +80,18 @@ set -g __fish_git_prompt_status_order untrackedfiles
...
+
+# Turn on everything and verify we correctly ignore sus config files.
+set -g __fish_git_prompt_status_order stagedstate invalidstate dirtystate untrackedfiles stashstate
+set -g __fish_git_prompt_showdirtystate 1
+set -g __fish_git_prompt_show_informative_status 1
+set -g __fish_git_prompt_showuntrackedfiles 1
+rm -Rf .git *
+git init >/dev/null 2>&1
+echo -n > ran.txt
+git config core.fsmonitor 'echo fsmonitor >> ran.txt; false'
```
krobelus, line 92: LGTM, thanks!

(fish_indent would make *some* of the redirections consistent (>/dev/null vs > /dev/null) but it can't see inside quotes)

> Ok, you've convinced me to remove the knob. Our story can be "don't use `bash.showInformativeStatus` on watchman-backed repos" and if someone really needs this they can hack up `fish_git_prompt` themselves.

Yeah. Going back and forth on this one. `bash.showInformativeStatus` and fsmonitor are beyond my horizon so I find it hard to empathize.

> I think it's OK to leave tab completions as they are. If you are tab-completing `git` you probably have a reason, and tab-completing e.g. `git add` will run `git status` which really can benefit from watchman.

OK

PR #8593, opened 2021-12-29
Expand wildcards on tab

### LINKED ISSUE #954: Make tab-completion with wildcards make more sense
Example:

```
$ touch file(seq 4)
$ ls
file1  file2  file3  file4
$ echo f*<tab>
```

This autocompletes to `f*ile`. It should autocomplete to `file*` or `file`, imo.

Tab completion with wildcards is pretty awkward at the moment.

ridiculousfish:
Prior to this change, if you tab-completed a token with a wildcard (glob), we would invoke ordinary completions. Instead, expand the wildcard, replacing the wildcard with the result of expansions. If the wildcard fails to expand,  flash the command line to signal an error and do not modify it.

Example:

    > touch file(seq 4)
    > echo file*<tab>

becomes:

    > echo file1 file2 file3 file4
	
whereas before the tab would have just added a space.

Some things to note:

1. If the expansion would produce more than 256 items, we flash the command line and do nothing, since it would make the commandline overfull.
2. The wildcard token can be brought back through Undo (ctrl-Z).

Fixes #954.

### DIFF
--- a/src/complete.cpp
+++ b/src/complete.cpp
@@ -674,11 +674,11 @@ void completer_t::complete_cmd(const wcstring &str_cmd) {
     completion_list_t possible_comp;
 
     // Append all possible executables
-    expand_result_t result =
-        expand_string(str_cmd, &this->completions,
-                      this->expand_flags() | expand_flag::special_for_command |
-                          expand_flag::for_completions | expand_flag::executables_only,
-                      ctx);
+    expand_result_t result = expand_string(
+        str_cmd, &this->completions,
+        this->expand_flags() | expand_flag::special_for_command | expand_flag::for_completions |
+            expand_flag::preserve_home_tildes | expand_flag::executables_only,
+        ctx);
     if (result == expand_result_t::cancel) {
         return;
     }
@@ -690,10 +690,10 @@ void completer_t::complete_cmd(const wcstring &str_cmd) {
     // updated with choices for the user.
     expand_result_t ignore =
         // Append all matching directories
-        expand_string(
-            str_cmd, &this->completions,
-            this->expand_flags() | expand_flag::for_completions | expand_flag::directories_only,
-            ctx);
+        expand_string(str_cmd, &this->completions,
+                      this->expand_flags() | expand_flag::for_completions |
+                          expand_flag::preserve_home_tildes | expand_flag::directories_only,
+                      ctx);
     UNUSED(ignore);
 
     if (str_cmd.empty() || (str_cmd.find(L'/') == wcstring::npos && str_cmd.at(0) != L'~')) {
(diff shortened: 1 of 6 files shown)


krobelus:
This is really powerful, but it disallows some scenarios,
Press tab on

	ls ~/.config/*/conf

(The glob is useful in case there are multiple directories with interesting files.)

previously, we got a set of completions with entries like `~/.config/*/config.fish`.
Now we get nothing (since there is no file named "conf").

Maybe `~/.config/*/conf` should expand to `~/.config/fish/config.fish ~/.config/whatever/we.conf`.
(Don't expand the last path component, but ask the completion engine instead)

Sadly that's still not perfect, because sometimes preserving the glob is nice.
Think of `ls src/*/java/com/example/...` - I want to tab-complete the java/com/example components but not expand the glob.
Of course that case is rare so I don't know how important it is; I'll try it for a while.

Maybe there is a criteria to guess if the user wants expansion or not. Hopefully we can avoid a separate key binding.

krobelus:
Maybe only expand `*` in the current path component. That'd cut this feature back so some useful completions keep working.

Here are some scenarios (where `|` is the cursor)

1. `echo R*|`
   - Expand to `README.rst`

2. `echo *.r|`
   - Glob fails, so offer completion (`*.rst` because `README.rst` exists).

3. `echo *.rst|`
   - Expand to `README.rst`

4. `echo benchmarks/*|`
   - Expand to `benchmarks/benchmarks/ benchmarks/driver.sh`

5. `echo benchmarks/*/|`
   - No glob in current path component, so offer completion (`benchmarks/*/aliases.fish` etc.).

6. `echo ben*/*|`
   - Expand to `benchmarks/benchmarks/ benchmarks/driver.sh`
   - Alternative: only expand the second `*`. That might be harder to implement.

7. `echo benchmarks/*|/`
   - Expand to `benchmarks/benchmarks/` (not driver.sh, because of the trailing slash)

8. `echo ben*|/nosuchfile`
   - Glob fails, so offer completion, which also fails.

9. `echo ben*|/bench`
   - Glob fails, so offer completion (`bench*/benchmarks/`) like we do now.
   - Alternative/future: expand parts of the glob, to `benchmarks/bench` and perhaps immediately complete to `benchmarks/benchmarks/`

10. `echo {benchmarks/*/,benchm}a|`
   - No glob in current path component, so offer completion (not really important I guess).

11. `echo ben**/aliases.fish|`
   - Expand to `benchmarks/benchmarks/aliases.fish`, even though `**` is not in the current path component!

12. `echo ben**/aliases.|`
   - Glob fails, so offer completion (this case is not really important, it doesn't work in master either).

I think these are acceptable because filename completion rarely does much when there is a `*` in the current path component (plus we can fall back to it, see 2).
So we'd need tweaks for 2 and 5 (I haven't looked at the source code yet).
I'm not entirely sure about 11, but that's not so important.

ridiculousfish:
Shout-out to @krobelus for detailed thinking here. I successfully implemented 1 through 10 (and 12 is unchanged). For 6, I didn't implement "expand only one wildcard;" it's probably doable but can be saved for the future.

PR #8597, opened 2021-12-30
completions/opam: add completions for "opam switch"

dmbaturin:
## Description

The completion script for [opam](https://opam.ocaml.org) is a good start, but not a real time-saver yet because it will not complete either sub-commands or their arguments.

For example, `opam switch set $switchName` is a frequent operation for many OCaml users, but without completion, the user has to type it from memory or look it up in `opam switch list`.

This is the first batch of changes intended to improve that.

## TODOs:


dmbaturin:
Test fails at https://github.com/fish-shell/fish-shell/runs/4669840319?check_suite_focus=true#step:6:280

Seems unrelated to my changes, unless I'm missing something.

faho:
>Seems unrelated to my changes, unless I'm missing something.

Yeah it's a flake.

-----

Merged, thanks!

PR #8601, opened 2022-01-01
Work on CHANGELOG for 3.4

IlanCosman:
## Description

Some work on the changelog. Cleans up all remaining except #8589. I don't fully understand what's going on there, nor do I know where to put such a changelog entry.


faho:
Thanks. In future: There's no need to back-merge master. It's typically easier for us to rebase.

If you want to keep on top of developments there, you should rebase, which removes awkward merge commits in the middle.

zanchey:
8403 was already in the ignore list (minor change, not actionable) and I don't usually add new subcommands to the new completion list so I'll revert this, but I really do appreciate the assistance! I have an uncommitted entry for #8589 that I'll add soon.

PR #8603, opened 2022-01-02
Run `fish_indent` on `funcsave`

kidonng:
## Description

`funced` already run `fish_indent` when the editor is fish:

https://github.com/fish-shell/fish-shell/blob/f254692759aa564420702a54e76feb133b50c25d/share/functions/funced.fish#L61-L75

It makes sense to have `funcsave` run it as well.

Maybe we should also run `fish_indent` after `funced` finishes?

## TODOs:

### DIFF
--- a/share/functions/funcsave.fish
+++ b/share/functions/funcsave.fish
@@ -29,6 +29,7 @@ function funcsave --description "Save the current definition of all specified fu
         set -l funcpath "$funcdir/$funcname.fish"
         if functions -q -- $funcname
             functions --no-details -- $funcname >$funcpath
+            and fish_indent --write $funcpath
             and set -q _flag_quiet || printf (_ "%s: wrote %s\n") funcsave $funcpath
         else if test -w $funcpath
             rm $funcpath
--- a/share/functions/funcsave.fish
+++ b/share/functions/funcsave.fish
@@ -28,8 +28,7 @@ function funcsave --description "Save the current definition of all specified fu
     for funcname in $argv
         set -l funcpath "$funcdir/$funcname.fish"
         if functions -q -- $funcname
-            functions --no-details -- $funcname >$funcpath
-            and fish_indent --write $funcpath
+            functions --no-details -- $funcname | fish_indent >$funcpath
             and set -q _flag_quiet || printf (_ "%s: wrote %s\n") funcsave $funcpath
         else if test -w $funcpath
             rm $funcpath


kidonng:
Interestingly, `funcsave`'s manpage says:

> This function will be automatically loaded by current and future fish sessions.

But the functions is clearly not autoloaded:

```fish
$ function t; end
$ functions -Dv t
stdin
n/a
0
scope-shadowing
n/a
```

Is the manpage wrong, or the function can be made autoloaded later via some magic I don't know? If the latter is true, then the function shown in `functions` will be formatted as well once fish autoloads it.

Ideally fish will automatically run `fish_indent` on interactively defined functions. But for this PR, there are some solutions I can think of:

- Do nothing about the interactively defined function. I don't really think it's a problem, since `fish_indent` doesn't change anything except the format. We can mention in the manpage that if people really want to preserve the format for some reason, just run `functions --no-details -- $funcname > $file` themselves.
- Only run `fish_indent` if given a `-f/--format` flag. Surely everything can be an option right?
- `source` the just saved function file. This however changes the description of the function from `Defined interactively` to `Defined in foo.fish @ line 1`.
- `eval` the formatted function and save it. That way the description of the function stays `Defined interactively` but `eval` is evil...

IlanCosman:
> But the functions is clearly not autoloaded:

I don't see where you ran `funcsave t`. You need to do so to have t be autoloaded. And it won't say "autoloaded` in the current shell, because it wasn't autoloaded, you defined it interactively. In future shells it will be autoloaded, and will say so correctly.

faho:
Okay, it seems we don't want to do this. Sorry!

kidonng:
> Okay, it seems we don't want to do this. Sorry!

@faho would you be happy to accept a PR to add the functionality behind a flag, instead of doing it automatically (which _may_ not be what the user expect), as mentioned above?

floam:
Something I was not aware of until just now: functions created interactively seem not to have tabs saved at all:

image

kidonng:
> Something I was not aware of until just now: functions created interactively seem not to have tabs preserved at all:

Right, that's the initial motivation of this PR. I got tired of having to run `fish_indent` manually after `funcsave`, because the saved function is not indented all.

krobelus:
I haven't found time to think about this.
As a workaround, how about teaching your editor to run fish_indent when you save a *.fish file?

krobelus:
this should be addressed by https://github.com/fish-shell/fish-shell/pull/10437 which makes `functions foo` and `type foo` output indented code

Issue #8496, opened 2021-11-28
Feature request: add support for '_' as thousands separator to math command


EdwardBetts:
It would be nice if I could use '_' as a thousands separator as a parameter to the math command. I'd like to be able to write 2000 as 2_000. I'm not asking for the output to change. 

Python, JavaScript, Perl and Ruby all support '_' as a thousands separator. It makes it easier to read large numbers. 

Example of suggested usage:
```
$ math 2_000 + 2
2002
$
```

mhmdanas:
Does it have to be for thousands only? Surely it can be used in other positions.

floam:
why not `,`?

faho:
Because math uses that to separate arguments to functions. It is unavailable.

EdwardBetts:
> Does it have to be for thousands only? Surely it can be used in other positions.

Yes, you're right, any position. I just checked, Javascript, Perl, Python and Ruby support '_' in any position within a number.

mqudsi:
> > why not `,`?
> Because math uses that to separate arguments to functions. It is unavailable.

Also because that's very locale-specific behavior.

andrey-mishchenko:
I looked into this a bit.

This was added to JavaScript recently, and they have a [good summary of prior art](https://github.com/tc39/proposal-numeric-separator) in their proposal. tl;dr: most languages use an underscore, allow it to be placed between digits, and disallow multiple underscores in a row, leading or trailing underscores, underscores next to `0x` or `0b`, etc.

The implementation of fish's `math` builtin uses (a ported-to-C++ and otherwise-modified) [tinyexpr](https://github.com/codeplea/tinyexpr) for expression parsing and evaluation.

Out-of-the-box, to parse out numeric literals, the tokenizer in tinyexpr invokes [strtod](https://en.cppreference.com/w/c/string/byte/strtof) when it arrives at a decimal digit or a dot (`.`). The function `strtod` also advances the next-character-pointer in the expression being parsed.

[...]

PR #8607, opened 2022-01-02
[RFC] Allow underscore separators in numeric literals in `math` builtin

andmis:
## Description

This adds support for underscore separators in the `math` builtin.

After trying a few different things, this solution felt simplest. We have our own parser that preprocesses out the underscores and passes the pruned-down string to `strtod`/`wcstod`.

There are a few smaller things I'm not sure about:

- I initially opted to disallow underscores next to base or exponent specifiers, but actually `0x_FF` and `1.23_e3` read nicely. I think I'd like to support these. Does anyone have strong feelings about the specific rules of where we allow underscores? Other languages differ. Here's [Python](https://www.python.org/dev/peps/pep-0515/) for reference.
- The tinyexpr code around my changes could be cleaned up but I wanted to keep the PR diff small.
- Our docs say that the `math` builtin supports hex, but we use `strtod` under the hood to parse and AFAICT OpenBSD `strtod` doesn't allow hex (I don't have an OpenBSD machine to test this on so I'm not sure). I made a small drive-by change in the `math` man page, see the PR.
- I know you are all busy in the middle of the 3.4 release, I added a changelog entry but TBH I was not sure where to put it. No hurry on this, either.

Fixes issue #8496

## TODOs:

### DIFF
--- a/src/wutil.cpp
+++ b/src/wutil.cpp
@@ -711,6 +711,78 @@ double fish_wcstod(const wchar_t *str, wchar_t **endptr) {
     return std::wcstod(str, endptr);
 }
 
+/// Like wcstod(), but allows underscore dividers between digits in the input string.
+double fish_wcstod_underscores(const wchar_t *str, wchar_t **endptr) {
+    // OpenBSD's strtod/wcstod do not support hex.
+    static const bool allow_hex = (fish_wcstod(L"0x1", nullptr) == 1);
+    const wchar_t *orig = str;
+    while (iswspace(*str)) str++;
+    std::wstring pruned;
+    auto take_char = [&pruned, &str]() {
+        if (*str != L'_') pruned.push_back(*str);
+        str++;
+    };
+    auto is_sign = [](wchar_t c) { return c == L'+' || c == L'-'; };
+    if (is_sign(*str)) take_char();
+    if (!wcscasecmp(str, L"infinity") || !wcscasecmp(str, L"nan")) {
+        return fish_wcstod(orig, endptr);
+    }
+    bool hex = false;
+    if (allow_hex && !wcsncmp(str, L"0x", 2)) {
+        take_char();
+        take_char();
+        hex = true;
+    }
+    auto is_mantissa_digit = [hex](wchar_t c) -> int { return hex ? iswxdigit(c) : iswdigit(c); };
+    bool got_radix = false;
+    bool got_digit = false;
+    while (is_mantissa_digit(*str) || *str == L'.' || *str == L'_') {
+        if (*str == L'_') {
+            if (str == orig || !is_mantissa_digit(*(str - 1)) || !is_mantissa_digit(*(str + 1))) {
+                break;
+            }
+        } else if (*str == L'.') {
+            if (got_radix) break;
+            got_radix = true;
+        } else {
+            got_digit = true;
+        }
+        take_char();
+    }
+    if (!got_digit) {
+        if (hex) {
+            str = orig;
+            while (*str != L'x') str++;
+            // hex = false; // This cannot have an effect in the rest of the function, so skip it.
+        } else {
+            if (endptr) *endptr = const_cast<wchar_t *>(orig);
+            return 0.0;
+        }
+    }
+    auto is_exp_char = [hex](wchar_t c) {
+        if (hex) {
+            return c == L'P' || c == L'p';
+        } else {
+            return c == L'E' || c == L'e';
+        }
+    };
+    // Check if *str is an exponent character followed by an optional sign and a digit.
+    if (is_exp_char(*str) &&
+            (iswdigit(*(str + 1)) ||
+             (is_sign(*(str + 1)) && iswdigit(*(str + 2))))) {
+        take_char(); // E/e/P/p
+        if (is_sign(*str)) take_char();
+        while (iswdigit(*str) || *str == L'_') {
+            if (*str == L'_' && (!iswdigit(*(str - 1)) || !iswdigit(*(str + 1)))) {
+                break;
+            }
+            take_char();
+        }
+    }
+    if (endptr) *endptr = const_cast<wchar_t *>(str);
+    return fish_wcstod(pruned.c_str(), nullptr);
+}
+
 file_id_t file_id_t::from_stat(const struct stat &buf) {
     file_id_t result = {};
     result.device = buf.st_dev;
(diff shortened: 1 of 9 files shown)


src/wutil.cpp
```diff
@@ -711,6 +711,78 @@ double fish_wcstod(const wchar_t *str, wchar_t **endptr) {
     return std::wcstod(str, endptr);
 }
 
+/// Like wcstod(), but allows underscore dividers between digits in the input string.
+double fish_wcstod_underscores(const wchar_t *str, wchar_t **endptr) {
+    // OpenBSD's strtod/wcstod do not support hex.
+    static const bool allow_hex = (fish_wcstod(L"0x1", nullptr) == 1);
```
faho, line 717: I don't believe this helps. We *want* reading hex numbers to fail on OpenBSD. It not having hex number support is a deficiency in OpenBSD, and should be fixed there. It's a compliance issue and technically means OpenBSD doesn't fullfil our requirements (C++11/C99 require hex support).

But anyway, I don't see how this helps: Best case, it only passes the `0` from `0xdeadbeef` to wcstod, which returns 0. Then the rest of the parsing sees an `x` without following whitespace and errors out. That means this is pointless because we could just let it read up to the `x` and fail.

Worst case, it doesn't error out and now `0x5` means "0 times 5" in OpenBSD and "5" elsewhere.

We should fail on OpenBSD so it can be fixed.

src/wutil.cpp
```diff
@@ -711,6 +711,78 @@ double fish_wcstod(const wchar_t *str, wchar_t **endptr) {
...
+            while (*str != L'x') str++;
+            // hex = false; // This cannot have an effect in the rest of the function, so skip it.
+        } else {
+            if (endptr) *endptr = const_cast<wchar_t *>(orig);
+            return 0.0;
+        }
+    }
+    auto is_exp_char = [hex](wchar_t c) {
+        if (hex) {
+            return c == L'P' || c == L'p';
```
faho, line 764: I'm not sure this isn't over-specified. Wouldn't it be simpler to read until a separator (`,`, `)`, whitespace, end of the string)?

Overreading shouldn't be an issue because wcstod can just return the end pointer, which we can then offset by the number of `_`.

Or is there a reason you're doing it like this?

faho (changes requested):

faho:
>Does anyone have strong feelings about the specific rules of where we allow underscores?

Honestly I don't - python's rationale here is that code style would disallow it anyway, but there's also value in having simpler rules.

>Our docs say that the math builtin supports hex, but we use strtod under the hood to parse and AFAICT OpenBSD strtod doesn't allow hex

I've not documented this because I've held out hope that OpenBSD would fix it - which would make its strtod C*99* conformant (AFAICT). But if we do mention it, I would like to specifically call out OpenBSD. "Most platforms" is a bit weasily and doesn't tell people if it will be available on *their* platform.

The odd OS out really does appear to be OpenBSD - we only skip the math tests there: https://github.com/fish-shell/fish-shell/blob/1552c526ba66ee2e9f34a590cad136c5821c84a7/tests/checks/math.fish#L2-L4

So any other platform that runs the tests seems to manage, or people don't care enough to submit a bug.

src/wutil.cpp
faho: I.e.

1. Create the pruned string, counting the number of `_`
2. Call wcstod, passing a second endptr
3. Set the real endptr to start of real string + offset of the second endptr + number of `_`
4. Return wcstod value

src/wutil.cpp
faho: Oh, I see this is a consequence of trying to find the end point that wcstod *would* return.

src/wutil.cpp
faho: This has the advantage that, if wcstod behaves differently than we think, we don't get weird double-counting or skipping, but have the end at the place it gives us.

floam:
https://github.com/openbsd/src/blob/2207c4325726fdc5c4bcd0011af0fdf7d3dab137/lib/libc/gdtoa/strtod.c looks like it handles hex floats.

faho:
>https://github.com/openbsd/src/blob/2207c4325726fdc5c4bcd0011af0fdf7d3dab137/lib/libc/gdtoa/strtod.c looks like it handles hex floats.

It doesn't. I don't know what it is about the code, but OpenBSD's strtod does *not* handle hex floats. I have seen it myself, we have had reports of it, and the man page mentions being C89 conformant, specifically not C99. The difference in those is hex numbers.

floam:
And this guy asserts it. Weird. 

https://marc.info/?l=openbsd-tech&m=155910240816770&w=2

>

faho:
Huh, I just tested an old fish I built on OpenBSD, and `math 0x5` didn't work ("unknown function").

Then I built 935ec3966841 and it started working.

I can't remember a system update I did (I haven't touched this VM in ages), and it also works with the fish (3.1.0) package, so I have no idea how this got fixed. Unfortunately I compiled over the old build so I can't test it anymore.

Well, good news either way. (and now I have no idea where the C99 compliance issue lies?)


Anyway: We should still ideally not try to second guess wcstod. I believe counting `_` is much less brittle.

floam:
I don’t think there is an issue. OpenBSD appears to be using gdtoa for these functions, and the readme for that documents the C99 features. But it does not provide manpages, strtod.1 is something OpenBSD has in a totally different part of their source tree, and it’s old. I think mentioning that it is also C99 compliant is an oversight.

src/wutil.cpp
andmis: It's not quite as simple as consuming all non-underscore characters until a separator and counting how many underscores you dropped in the process for the eventual offset.

First, if you don't implement real parsing logic, then you will convert `1_1.2_2.3_3` to `11.22.33` and `wcstod` will only parse until the second decimal. Which is fine, but you need to know that you should only advance your own `endptr` with an offset assuming you dropped two underscores, not all three.

Second, the situation depends on what rules we want to use for where underscores can appear. If we want to have nontrivial restrictions, then I think we basically need to implement the full parsing logic. For example, deciding whether an `a` is an end-of-parse or a hex digit requires knowing whether the string began with `0x`.

In my mind, originally I was planning to have some rules about where underscores could appear. But if we are ok with being very permissive (as in #8611) that simplifies things and we can get away without a full parse.

I figured that if I was going to do a full parse anyway, then I would just implement the parse corresponding to the C99 spec -- it turns out it's not really that bad. The `allow_hex` runtime feature flag also is not that bad IMO.

It's true that we are dead-reckoning (or praying) in this case that our parse is going to be equal to the system `wcstod` parse, which I agree feels pretty uncomfortable. But if that is not the case then either there is a bug in our parse, or the system `wcstod` is non-conformant. I didn't implement the `endptr` offsetting using the offset we get from the system `wcstod` because (1) if there are no bugs and the system function is conformant then I should be able to assert that it's equal to the end of the pruned string anyway, (2) the fallback logic for offsetting based on number of underscores dropped is not completely trivial (because of the `1_1.2_2.3_3` case above, it's not enough to count), (3) if we have a bug in the parser or the system stdlib function is non-conformant then I didn't see a graceful recovery being likely.

Anyway I do think the new version is better and probably more robust.

andmis:
Closed in favor of #8611, see my comments at https://github.com/fish-shell/fish-shell/pull/8607#discussion_r777793957.

PR #8611, opened 2022-01-04
Allow underscores as separators in the `math` builtin

andmis:
## Description

This change allows underscores as visual separators in the math builtin, so `math 1_000 + 2_000` will return `3000`. We allow leading, trailing, and multiple underscores. See the documentation of the `fish_wcstod_underscores` function in the PR.

See also #8607.

Fixes issue #8496.

## TODOs:

### DIFF
--- a/doc_src/cmds/math.rst
+++ b/doc_src/cmds/math.rst
@@ -59,6 +59,8 @@ Syntax
 For numbers, ``.`` is always the radix character regardless of locale - ``2.5``, not ``2,5``.
 Scientific notation (``10e5``) and hexadecimal (``0xFF``) are also available.
 
+``math`` allows you to use underscores as visual separators for digit grouping. For example, you can write ``1_000_000``, ``0x_89_AB_CD_EF``, and ``1.234_567_e89``.
+
 Operators
 ---------
 
--- a/src/fish_tests.cpp
+++ b/src/fish_tests.cpp
@@ -3000,6 +3000,49 @@ static void test_wcstod() {
     tod_test(L"nope", "nope");
 }
 
+static void test_fish_wcstod_underscores() {
+    say(L"Testing fish_wcstod_underscores");
+
+    auto test_case = [](const wchar_t *s, size_t expected_num_consumed) {
+        wchar_t *endptr = nullptr;
+        fish_wcstod_underscores(s, &endptr);
+        size_t num_consumed = (size_t)(endptr - (wchar_t *)s);
+        do_test(expected_num_consumed == num_consumed);
+    };
+
+    test_case(L"123", 3);
+    test_case(L"1_2.3_4.5_6", 7);
+    test_case(L"1_2", 3);
+    test_case(L"1_._2", 5);
+    test_case(L"1__2", 4);
+    test_case(L" 1__2 3__4 ", 5);
+    test_case(L"1_2 3_4", 3);
+    test_case(L" 1", 2);
+    test_case(L" 1_", 3);
+    test_case(L" 1__", 4);
+    test_case(L" 1___", 5);
+    test_case(L" 1___ 2___", 5);
+    test_case(L" _1", 3);
+    test_case(L"1 ", 1);
+    test_case(L"infinity_", 8);
+    test_case(L" -INFINITY", 10);
+    test_case(L"_infinity", 0);
+    test_case(L"nan(0)", 6);
+    test_case(L"nan(0)_", 6);
+    test_case(L"_nan(0)", 0);
+    // We don't strip the underscores in this commented-out test case, and the behavior is
+    // implementation-defined, so we don't actually know how many characters will get consumed. On
+    // macOS the strtod man page only says what happens with an alphanumeric string passed to nan(),
+    // but the strtod consumes all of the characters even if there are underscores.
+    // test_case(L"nan(0_1_2)", 3);
+    test_case(L" _ 1", 0);
+    test_case(L"0x_dead_beef", 12);
+    test_case(L"None", 0);
+    test_case(L" None", 0);
+    test_case(L"Also none", 0);
+    test_case(L" Also none", 0);
+}
+
 static void test_dup2s() {
     using std::make_shared;
     io_chain_t chain;
@@ -6792,6 +6835,7 @@ static const test_t s_tests[]{
     {TEST_GROUP("abbreviations"), test_abbreviations},
     {TEST_GROUP("builtins/test"), test_test},
     {TEST_GROUP("wcstod"), test_wcstod},
+    {TEST_GROUP("fish_wcstod_underscores"), test_fish_wcstod_underscores},
     {TEST_GROUP("dup2s"), test_dup2s},
     {TEST_GROUP("dup2s"), test_dup2s_fd_for_target_fd},
     {TEST_GROUP("path"), test_path},
--- a/src/wutil.cpp
+++ b/src/wutil.cpp
@@ -18,9 +18,11 @@
 #include <unistd.h>
 #include <wctype.h>
 
+#include <algorithm>
 #include <atomic>
 #include <cstring>
 #include <cwchar>
+#include <iterator>
 #include <string>
 #include <unordered_map>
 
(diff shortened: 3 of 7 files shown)


src/fish_tests.cpp
```diff
@@ -3000,6 +3000,49 @@ static void test_wcstod() {
...
+    // implementation-defined, so we don't actually know how many characters will get consumed. On
+    // macOS the strtod man page only says what happens with an alphanumeric string passed to nan(),
+    // but the strtod consumes all of the characters even if there are underscores.
+    // test_case(L"nan(0_1_2)", 3);
+    test_case(L" _ 1", 0);
+    test_case(L"0x_dead_beef", 12);
+    test_case(L"None", 0);
+    test_case(L" None", 0);
+    test_case(L"Also none", 0);
+    test_case(L" Also none", 0);
```
krobelus, line 3043: Looking at the implementation, these two cases seem redundant

doc_src/cmds/math.rst
```diff
@@ -59,6 +59,8 @@ Syntax
 For numbers, ``.`` is always the radix character regardless of locale - ``2.5``, not ``2,5``.
 Scientific notation (``10e5``) and hexadecimal (``0xFF``) are also available.
 
+``math`` allows you to use underscores as visual separators for digit grouping. For example, you can write ``1_000_000``, ``0x_AB_CD_EF``, and ``1_234.567_e89``.
```
krobelus, line 62: nit:: maybe use more realistic examples here:

```suggestion
``math`` allows you to use underscores as visual separators for digit grouping. For example, you can write ``1_000_000``, ``0x_89_AB_CD_EF``, and ``1.234_567_e89``.
```

src/wutil.cpp
```diff
@@ -711,6 +713,66 @@ double fish_wcstod(const wchar_t *str, wchar_t **endptr) {
...
+    // sometimes it is the end of the parse, etc.
+    while (is_possible_alnum(*str) || is_sign(*str) || *str == L'.' || *str == L'_') take_char();
+    const wchar_t *pruned_begin = pruned.c_str();
+    const wchar_t *pruned_end = nullptr;
+    double result = fish_wcstod(pruned_begin, (wchar_t **)(&pruned_end));
+    if (pruned_end == pruned_begin) {
+        *endptr = (wchar_t *)orig;
+        return result;
+    }
+    auto last_underscore_consumed = std::upper_bound(underscores.begin(), underscores.end(),
```
krobelus, line 768: nit: I think "last" is a bit misleading here, maybe `consumed_underscores_end` fits better

src/wutil.cpp
```diff
@@ -711,6 +713,66 @@ double fish_wcstod(const wchar_t *str, wchar_t **endptr) {
...
+                                                     (size_t)(pruned_end - pruned_begin));
```
krobelus, line 769: I think we use `size_t(foo)` more often than `(size_t)foo`. Didn't check other types.

src/wutil.cpp
```diff
@@ -711,6 +713,66 @@ double fish_wcstod(const wchar_t *str, wchar_t **endptr) {
...
+    std::vector<size_t> underscores;
+    // We keep track of the positions *in the pruned string* where there used to be underscores. We
+    // will pass the pruned version of the input string to the system wcstod, which in turn will
+    // tell us how many characters it consumed. Then we will set our own endptr based on (1) the
+    // number of characters consumed from the pruned string, and (2) how many underscores came
+    // before the last consumed character. The alternative to doing it this way (for example, "only
+    // deleting the correct underscores") would require actually parsing the input string -- for
+    // example, sometimes a dot is a decimal delimiter and sometimes it represents the end of the
+    // valid float, as in "1_2.3_4.5_6".
+    auto take_char = [&pruned, &str, &underscores]() {
```
krobelus, line 746: I guess the single-use lambdas are alright if they serve the comments.

src/wutil.cpp
```diff
@@ -711,6 +713,66 @@ double fish_wcstod(const wchar_t *str, wchar_t **endptr) {
...
+    size_t leading_whitespace = (size_t)(str - orig);
+    auto is_sign = [](wchar_t c) { return c == L'+' || c == L'-'; };
+    auto is_inf_or_nan_char = [](wchar_t c) {
+        return c == L'i' || c == L'I' || c == L'n' || c == L'N';
+    };
+    // We don't do any underscore-stripping for infinity/NaN.
+    if (is_inf_or_nan_char(*str) || (is_sign(*str) && is_inf_or_nan_char(*(str + 1)))) {
+        return fish_wcstod(orig, endptr);
+    }
+    std::wstring pruned;
```
krobelus, line 736: We use our `wcstring` alias almost everywhere, probably also here

src/wutil.cpp
```diff
@@ -711,6 +713,66 @@ double fish_wcstod(const wchar_t *str, wchar_t **endptr) {
...
+    };
+    // In order to be more clever, we would need to do a full parse, because sometimes 'a' is a hex
+    // digit and sometimes it is the end of the parse, sometimes '+' is an exponent sign and
+    // sometimes it is the end of the parse, etc.
+    while (is_possible_alnum(*str) || is_sign(*str) || *str == L'.' || *str == L'_') take_char();
+    const wchar_t *pruned_begin = pruned.c_str();
+    const wchar_t *pruned_end = nullptr;
+    double result = fish_wcstod(pruned_begin, (wchar_t **)(&pruned_end));
+    if (pruned_end == pruned_begin) {
+        *endptr = (wchar_t *)orig;
```
krobelus, line 765: shouldn't we check if endptr is null here?
(I guess we could also move this outside the if, and then use += below, not sure)

krobelus:
looks really nice, I only have nitpicks

src/fish_tests.cpp
andmis: Which ones? The four immediately above test slightly different code paths:

- no leading whitespace, `is_inf_or_nan_char`
- leading whitespace, `is_inf_or_nan_char`
- no leading whitespace, `!is_inf_or_nan_char`
- leading whitespace, `!is_inf_or_nan_char`

I'm happy to delete one or more, your call.

src/wutil.cpp
andmis: Actually I think I had this lambda used multiple times in an earlier draft -- I can remove it if you guy tend not to use this style. LMK. The comments can get moved around.

src/wutil.cpp
andmis: Good catch, sorry about that.

I don't think we should move it because then we need extra logic about not offsetting by the leading whitespace, and the logic of the implementation is easier to read this way.

src/wutil.cpp
andmis:
```
RP/fish. rg -F '(size_t)' | count
17
RP/fish. rg -F 'size_t(' | count
20
RP/fish. 
```

Neck-and-neck. I'll switch it, though.

andmis: I'm 2 of those 17, and in both cases the expression being cast is parenthesized anyway so yeah you're right. :)

src/wutil.cpp
andmis: I agree.

andmis:
Thanks for the review. A missed null pointer check isn't a nitpick. :)

I've made the changes you requested -- one thing is that I added a changelog entry and I'm not sure about where I put it, so before merging someone should double-check that.

zanchey:
I want to hold this until after 3.4.0 is out as we're doing the final QA for the release now.

andmis:
Yep that makes sense -- we'll just have to move the changelog entry before merging (it's under 3.4 right now).

src/fish_tests.cpp
krobelus: Oh I didn't realize that, let's keep them then.
(would be nice to have a tool that shows the coverage of each test case)

src/wutil.cpp
krobelus: I think the code will become simpler by inlining the lambda, though I haven't tried that.

faho:
Merged now, thanks!

PR #8612, opened 2022-01-04
Skip more leading commands in `__fish_man_page`

kidonng:
## Description

Like #8447 but for more commands.

The list is generated via the following command:
```fish
rg -l __fish_complete_subcommand $__fish_data_dir/completions | xargs basename | sort | string replace .fish '' | string join "|"
```

I excluded `git` and `ssh` from the list since their "subcommand" cannot be arbitrary program.

## TODOs:


faho:
Merged, thanks!

PR #8613, opened 2022-01-04
Fix `xargs` option completions

kidonng:
## Description

692a8aef03d52b4bd63f4f99d18ee3334a71e2af added subcommand completions to `xargs` but it doesn't consider the location of the current token.

## TODOs:


zanchey:
Nice catch!

PR #8625, opened 2022-01-09
docs synopsis: add HTML highlighing and automate manpage markup

krobelus:
The RST syntax from the recent synopsis changes does not translate
well to HTML. Alignment and syntax highlighting are lost.

Fix this by adding a Pygments lexer plugin that can tokenize
a synopsis and assign the various highlighting roles
(command/parameter/operator/BNF metacharacter) for HTML.

Demo how this will look like on two examples.

The disadvantage here is that man pages lose bold/underline markup,
which is acceptable? If we want to keep that we might need to use
Sphinx roles. Maybe there is an easy way to do that?

(The best solution would be to use the custom lexer also for manpage
output, sadly that doesn't seem possible in Sphinx 4.3.2).


zanchey:
> The disadvantage here is that man pages lose bold/underline markup,
> which is acceptable?

If possible I think it is better to keep it - the manpage output is markedly improved with formatting.

krobelus:
Hmm there is an easy way to have both:

```rst
string-match - match substrings
===============================

Synopsis
--------

.. BEGIN SYNOPSIS

.. only:: not html

    ``string`` match [**-a** | **--all**] [**-e** | **--entire**] [**-i** | **--ignore-case**]
             \       [**-r** | **--regex**] [**-n** | **--index**] [**-q** | **--quiet**] [**-v** | **--invert**]
             \       *PATTERN* [*STRING*...]

.. only:: html

    .. code-block:: synopsis
    
     string match [-a | --all] [-e | --entire] [-i | --ignore-case]
                  [-r | --regex] [-n | --index] [-q | --quiet] [-v | --invert]
                  PATTERN [STRING...]

.. END SYNOPSIS
```

observe that it's trivial to generate the second one from the first.
Now we just need to find a way to make Sphinx do that on-the-fly, so we don't actually duplicate every synopsis.

zanchey:
I find it hard to believe that nobody else is trying to do this, but I can't find anything with a brief look around.

floam:
I’m kinda sadface here obviously. I feel like getting nice output on the terminal should win.  Longer term, I think we could colorize e.g commands and options differently by using :command:: and those :option:: dealios Sphinx supports outside the box. I’m without a computer here so can’t really add much until Monday.

krobelus:
> Longer term, I think we could colorize e.g commands and options differently by using :command:: and those :option:: dealios Sphinx supports out of the box instead of just bold/italics.

I think for HTML it's important that the synopsis is consistent with code examples, which is hard with (custom) roles.
So I went with literal code blocks.
Eventually we should also have the same fish-like syntax highlighting in code examples in manpages. That just needs to be supported by sphinx (man(1) already supports color). This will probably be easier with literal code blocks.

I have pushed a version that highlights both HTML and manpages.
The manpage output has changed because I wasn't sure if there is a consistent set of
rules in the old markup.

Here is the old highlighting (B = bold, N = regular, U = underlined):

	string join [-q | --quiet] SEP [STRING…]
	BBBBBB NNNN NBB N BBBBBBBN NNN NUUUUUUNN
	PREVIOUS; and COMMAND
	NNNNNNNNN BBB UUUUUUU

and the new one:

	string join [-q | --quiet] SEP [STRING…]
	BBBBBB NNNN NNN N NNNNNNNN NNN NNNNNNNNN
	PREVIOUS; and COMMAND
	BBBBBBBBB UUU BBBBBBB

I personally think that this subtle markup in manpages is better, but we can also preserve some of the old behavior.
Which rules do we want to keep? Something like
- "Options are bold"
- "UPPERCASE tokens are underlined"
- ...

(another small change is `...` -> `…` though that's easy to fix at the souirce)

floam:
And, if there is a synopsis section that has some really complicated stuff going on, like any variables or strings… probably it is not really a good usage synopsis and is acting more like an EXAMPLES section or something and we should consider splitting it up. I did this a couple times.

krobelus:
> Were you able to fix the synopsis section indentation weirdness in the manpages?

yeah, they are no longer indented
		
> Did you try just **doing** our code block formatting with some CSS for the synopsis?

That's what I [did at first](https://github.com/fish-shell/fish-shell/commit/79d1b3e085ede508194b19c907f249787b88ebec#r62788729) but it's not enough; we need to enrich our HTML because we don't want to parse things client-side.
With the current version no CSS changes are necessary since we treat them like code blocks.

> I really think our functions and builtins should have first class `--help` output.

Ok, but what do we want? Are my proposed changes okay? Once we have decided, it will be very easy to implement the markup in `FishSynopsisDirective`; which probably means less manual work in future.

ridiculousfish:
The HTML changes look great, a big improvement, but yeah the man page formatting ought to be preserved too. Both are important. If we can infer the formatting from the synopsis by lexing it, as the new lexer does, then that's ideal.

> Which rules do we want to keep?

GNU's `man man` gives some conventions for how man pages should be formatted:

    bold text          type exactly as shown.
    italic text        replace with appropriate argument.
    [-abc]             any or all arguments within [ ] are optional.
    -a|-b              options delimited by | cannot be used together.
    argument ...       argument is repeatable.
    [expression] ...   entire expression within [ ] is repeatable.

For fish's man pages I suggest:

1. Emphasize **`command`**s and **`--options`** since both are literal
2. Non-literal arguments like FILE get italics (often renders as underline).
3. Subcommands are normal text. These seem uncommon in system man pages, I think it would be too busy to emphasize these.

    **string** join [**-q** | **--quiet**] <ins>SEP</ins> [<ins>STRING</ins> …]
    **string** join0 [**-q** | **--quiet**] [<ins>STRING</ins> …]

this is quite close to what we have today and it would be great if the synopsis lexer could output this, so the .rst files could be simpler. Is that doable?

floam:
>  Are my proposed changes okay?

Please give me some a day or two. I just don't have regular access to a computer but I will tomorrow. I'm not able to really review this properly yet.

floam:
That said, I'm on a phone now and trying to peek. 

> I wasn't sure if there is a consistent set of rules in the old markup.

Well, there were supposed to be. Commands in ``, options bolded, placeholder labels in italics (underlined by man), keywords bolded. It rendered reasonably in man but it was a serious error to reuse bold there in hindsight. I should have just done `````` it sure seems.

I admit being defensive before and having just really scanned the discussion in my inbox -- I'm all for getting an  effective fish-literal-option-synopsis block directive or whatever this does exactly that will let us just expect it to do the right thing, rendering correctly as both HTML and roff. I'll see if I can get time tonight.

krobelus:
> 1. Emphasize **`command`** s and **`--options`** since both are literal
> 2. Non-literal arguments like FILE get italics (often renders as underline).
> 3. Subcommands are normal text. These seem uncommon in system man pages, I think it would be too busy to emphasize these.

Ok, done, except that I made subcommands bold as well, because it's be weird to be inconsistent here. That can be changed easily.

(I added the rule that (italic) placeholders must always be uppercase. Not perfect but it's easy to implement.)

The HTML output of builtins like `break` is a bit weird because a placeholder like COMMANDS is highlighted as command token, but that's okay-ish.

krobelus:
> It seemed to me that if it looks just like some lines of fish script entered into the shell in an example section that could be confusing.

True and I think it's worthwhile to explore rendering the HTML synopsis differently from literal examples. This PR is just a local optimum. The synopsis lexer should help with that; it will be much easier to change everything at once.

doc_src/conf.py
```diff
@@ -10,6 +10,7 @@
 import os.path
 import pygments
```
zanchey, line 11: This import is now unused.

zanchey:
This is great - strictly an improvement over the existing *and* previous implementation. Thanks!

floam:
This is pretty slick. I'm obviously happy with it.

Issue #8621, opened 2022-01-07
`git push <remote>` doesn't complete tags


spenserblack:
fish version: 3.3.1
OS: Ubuntu 20.04.3 LTS
`$TERM`: xterm-256color
Behavior persists after running `sh -c 'env HOME=$(mktemp -d) fish'`

***

Example:
```fish
git tag v1.0.0
git push origin v # press TAB, not completed to v1.0.0
```

Coming from bash, I expected the tag name to be completed just like branches are.
From comments like https://github.com/fish-shell/fish-shell/issues/8516#issuecomment-984341875 I see that PRs are welcome for completions, but I'm still pretty new to fish and wanted to report this in case anybody else wants to try this out :slightly_smiling_face:  
I'll need to learn more about fish first.

PR #8630, opened 2022-01-10
Include tags in `git push` completions

spenserblack:
## Description

This adds tags to the completion of `git push <remote>`. Tags are also references that can be pushed. Tags will be unsorted (`-ka`), similar to other `git <subcommand>` completions.

Fixes issue #8621

## TODOs:

- [ ] User-visible changes noted in CHANGELOG.rst (I think this is already covered by the existing changelog for 3.4.0)

### DIFF
--- a/share/completions/git.fish
+++ b/share/completions/git.fish
@@ -1667,8 +1667,8 @@ complete -f -c git -n '__fish_git_using_command range-diff' -l no-dual-color -d 
 ### push
 complete -f -c git -n __fish_git_needs_command -a push -d 'Update remote refs along with associated objects'
 complete -f -c git -n '__fish_git_using_command push; and not __fish_git_branch_for_remote' -a '(__fish_git_remotes)' -d 'Remote alias'
-complete -f -c git -n '__fish_git_using_command push; and __fish_git_branch_for_remote' -a '(__fish_git_branches)'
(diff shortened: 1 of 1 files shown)


share/completions/git.fish
```diff
@@ -1668,6 +1668,7 @@ complete -f -c git -n '__fish_git_using_command range-diff' -l no-dual-color -d
 complete -f -c git -n __fish_git_needs_command -a push -d 'Update remote refs along with associated objects'
 complete -f -c git -n '__fish_git_using_command push; and not __fish_git_branch_for_remote' -a '(__fish_git_remotes)' -d 'Remote alias'
 complete -f -c git -n '__fish_git_using_command push; and __fish_git_branch_for_remote' -a '(__fish_git_branches)'
+complete -f -c git -n '__fish_git_using_command push; and __fish_git_branch_for_remote' -ka '(__fish_git_tags)' -d Tag
```
krobelus, line 1671: Looks like -k preserves the relative order of tags, but they are still interleaved with branches, which weird (but acceptable).
I'd flip these two lines and add another -k so we always get (local) branches first.

	complete -f -c git -n '__fish_git_using_command push; and __fish_git_branch_for_remote' -ka '(__fish_git_tags)' -d Tag
	complete -f -c git -n '__fish_git_using_command push; and __fish_git_branch_for_remote' -ka '(__fish_git_branches)'

spenserblack: Done. It looks much better this way IMO :smiley:

PR #8632, opened 2022-01-12
Css: Update font stack

faho:
## Description

Unfortunately, the "monospace" and "sans-serif" font families are basically broken. The browser defaults have ossified, e.g. Firefox and Edge on Windows still default to Arial. (this situation, for the record, is horrible)

So, instead we try to use the "system-ui" font. Unfortunately there's a few caveats here:

1. Windows picks a different font depending on locale, and if the locale isn't latin-based it might not be good for latin text. So we prefer "Segoe UI", the latin windows font.
2. For monospace fonts there is "ui-monospace", but that's only supported in Safari, so we still need to give fonts that systems may use by default.

Online version at https://bean.solutions/fish-doc/.

## Screenshots

(in Firefox on Linux)

Before

Screenshot 2022-01-12 at 17-38-03 Fish for bash users — fish-shell 3 3 1-885-gb1ec8af5e documentation

After

Screenshot 2022-01-12 at 19-57-24 Fish for bash users — fish-shell 3 3 1-885-gb1ec8af5e documentation

Note also how there's a colored background behind the inline-code blocks. This is also in light mode, and allows distinguishing the two even if the fonts are similar (like here).

@zanchey If you are comfortable with this I would like it for 3.4.

I do not believe this should ever make anything *worse*. Technically it's possible you'd get e.g. Segoe on linux, but Segoe is an okay font, and very very likely nicer than what "sans-serif" would pick for you.

### DIFF
--- a/doc_src/python_docs_theme/static/pydoctheme.css
+++ b/doc_src/python_docs_theme/static/pydoctheme.css
@@ -228,7 +228,7 @@ tr, pre {
 
 code {
     /* Make inline-code better visible */
-    background-color: rgba(0,0,0, .08);
+    background-color: rgba(20,20,80, .1);
     padding-left: 5px;
     padding-right: 5px;
 }
@@ -519,6 +519,6 @@ div.body .internal.reference:link {
     }
 
     code {
-        background-color: rgba(255, 255, 255, .08);
+        background-color: rgba(200, 200, 255, .1);
     }
 }
--- a/doc_src/python_docs_theme/static/pydoctheme.css
+++ b/doc_src/python_docs_theme/static/pydoctheme.css
@@ -222,8 +222,15 @@ div.body a:hover {
     color: #00B0E4;
 }
 
-tr, code, pre {
+tr, pre {
     background-color: inherit;
+}
+
+code {
+    /* Make inline-code better visible */
+    background-color: rgba(0,0,0, .08);
+    padding-left: 5px;
+    padding-right: 5px;
 }
 
 tt, code, pre, dl > dt span ~ em, #synopsis p, #synopsis code {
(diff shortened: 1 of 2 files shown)


zanchey (approved):
This looks great.

faho:
Okay, Windows now picks Segoe UI and Consolas, it used to pick Arial/Courier New in Firefox and Arial/Consolas in Edge.

On Arch I'm not sure what it picks for the proportional font, possibly Nimbus Sans (that's what `fc-match` tells me for my default). I have no idea how to figure out what "system-ui" resolves to.

On KDE it'll now pick Hack for monospace (and possibly Noto Sans typically for proportional? it's unclear), on macOS San Francisco, on Ubuntu Ubuntu and on Gnome Cantrell and probably Noto Sans Mono (because we can't list Source Code Pro).

At worst it'll pick an unfamiliar but acceptable font or fall back on sans-serif/monospace. This is better than picking Arial and some bad monospace font in most cases.

The downside is that this is no longer user-configurable. That's an unfortunately necessary trade-off because the defaults are horrible.

faho:
I would appreciate testing on macOS given that I've eschewed listing "BlinkMacSystemFont" and "-apple-system" under the belief that all browsers there now support system-ui.

zanchey:
macOS 12.1:

Chrome
Screen Shot 2022-01-13 at 21 22 09

Firefox
Screen Shot 2022-01-13 at 21 22 33

Safari
Screen Shot 2022-01-13 at 21 22 43

faho:
Okay, I just tried picking Cascadia Code (the Windows Terminal default font), and our comment/output styles are italic, so it's awkward. They've improved the italic version so it's not unreadable to non-americans anymore, but it's still *very* different from normal. So I won't be adding it here.

Other than that, this seems to give okay results everywhere, so I'm merging!

PR #8636, opened 2022-01-14
completions(apk.fish): force file/parameter completion

panekj:
Signed-off-by: Jakub Panek <me@panekj.dev>

## Description

Fixes `apk` completion to allow file completions and requires values in parameters

## TODOs:

### DIFF
--- a/share/completions/apk.fish
+++ b/share/completions/apk.fish
@@ -6,8 +6,8 @@ complete -f -c apk -n "__fish_seen_subcommand_from del fix version manifest" -a 
 
 # Global options
 complete -f -c apk -s h -l help -d "Show help"
-complete -x -c apk -s p -l root -d "Install packages to DIR"
-complete -x -c apk -s X -l repository -d "Use packages from REPO"
+complete -r -c apk -s p -l root -d "Install packages to DIR"
+complete -r -c apk -s X -l repository -d "Use packages from REPO"
 complete -f -c apk -s q -l quiet -d "Print less information"
 complete -f -c apk -s v -l verbose -d "Print more information"
 complete -f -c apk -s i -l interactive -d "Ask confirmation for certain operations"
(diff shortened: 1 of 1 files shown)


share/completions/apk.fish
```diff
@@ -6,8 +6,8 @@ complete -f -c apk -n "__fish_seen_subcommand_from del fix version manifest" -a
 
 # Global options
 complete -f -c apk -s h -l help -d "Show help"
-complete -x -c apk -s p -l root -d "Install packages to DIR"
-complete -x -c apk -s X -l repository -d "Use packages from REPO"
+complete -F -c apk -s p -l root -r -d "Install packages to DIR"
```
krobelus, line 9: We usually only use `-F` if really necessary. I think in this case you can just drop it (untestd), since file completion is the default.

share/completions/apk.fish
```diff
@@ -6,8 +6,8 @@ complete -f -c apk -n "__fish_seen_subcommand_from del fix version manifest" -a
...
+complete -F -c apk -s X -l repository -r -d "Use packages from REPO"
```
krobelus, line 10: are you sure that REPO is a filename?

share/completions/apk.fish
```diff
@@ -20,18 +20,18 @@ complete -f -c apk -l force-overwrite -d "Overwrite files in other packages"
 complete -f -c apk -l force-refresh -d "Don't use cached files"
 complete -f -c apk -s U -l update-cache -d "Alias for --cache-max-age 1"
 complete -f -c apk -l progress -d "Show a progress bar"
-complete -f -c apk -l progress-fd -d "Write progress to FD"
+complete -F -c apk -l progress-fd -r -d "Write progress to FD"
```
krobelus, line 23: this should probably stay without file completions; FD = file descriptor, so 1 for stdout, 2 for stderr, etc.

share/completions/apk.fish
panekj: It can point to local directory, so the path completion should be available imo.
Example of `apk add`:
```sh
apk add -X ~/packages/main <package_name>
```

Example of `/etc/apk/repositories`:
```
#/media/cdrom/apks

@aports-m /home/pj/packages/main
@aports-c /home/pj/packages/community
@aports-t /home/pj/packages/testing

https://dl-cdn.alpinelinux.org/alpine/v3.15/main
https://dl-cdn.alpinelinux.org/alpine/v3.15/community
```

share/completions/apk.fish
panekj: Didn't test all but it seems to work fine

PR #8643, opened 2022-01-14
Assert `and` and `or` create no new scope

scorphus:
## Assert `and` and `or` create no new scope

This PR suggests the addition of 2 set of tests that assert that variables declared as part of `and` and `or` statements get declared in the expected scope.

There's not issue related to this change. This PR comes as a result of a [brief conversation I had with @zanchey over Gitter](https://gitter.im/fish-shell/fish-shell?at=61d99374526fb77b316c0218).

## TODOs:

I believe none of the TODOs above apply.

I hope this is a good first contribution and I'm happy to make any change to it.

### DIFF
--- a/tests/checks/set.fish
+++ b/tests/checks/set.fish
@@ -635,6 +635,34 @@ end
 #CHECK: $var6[1]: |ghi|
 #CHECK: $var6[2]: |jkl|
 
+# `and` creates no new scope on its own
+true; and set -l var7a 89 179
+set -q var7a
+echo $status
+#CHECK: 0
+
+# `begin` of an `and` creates a new scope
+true; and begin
+    set -l var7b 359 719
+end
+set -q var7b
+echo $status
+#CHECK: 1
+
+# `or` creates no new scope on its own
+false; or set -l var8a 1439 2879
+set -q var8a
+echo $status
+#CHECK: 0
+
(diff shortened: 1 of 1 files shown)


tests/checks/set.fish
```diff
@@ -635,6 +635,36 @@ end
...
+#CHECK: $var7a[1]: |89|
+#CHECK: $var7a[2]: |179|
+true; and begin
+    set -l -a var7b 359 719
+    set --show var7b
+end
+set --show var7b
+#CHECK: $var7b: set in local scope, unexported, with 2 elements
+#CHECK: $var7b[1]: |359|
+#CHECK: $var7b[2]: |719|
```
krobelus, line 651: I found this a bit hard to follow.

In the gitter link you summarize this as "and/or do not create a new scope".
This is quite important to understand the point of this test, so it should be in a comment or commit message.

You add have scenarios `true; and set someglobal` and `true; and begin; set -l somelocal; end` but I don't see a test that actually ensures that a and/or don't create a new scope.
That would need something like `true; and set -l local; set -q local`.
I guess a minimal test along those lines could work.

scorphus: Your inputs are helpful, @krobelus. Thanks. I'll soon update this PR.

faho:
Merged, thanks!

PR #8645, opened 2022-01-15
Add --unescaped option to `complete -C`

### LINKED ISSUE #3469: Escape and complete -C (for fzf completion widget)

nzig:
This PR is meant to fix #3469.

It is based on https://github.com/krobelus/fish-shell/commit/38b4fe5db278d5dfb1a0009013a74ce626077451 by @krobelus, except that it gates the feature behind a new `--unescaped` flag to not break existing scripts that use `complete -C`.

There should probably be more tests, but I'm not familiar enough to know what tests should be added. I'd appreciate some guidance on what those should look like. https://github.com/krobelus/fish-shell/commit/38b4fe5db278d5dfb1a0009013a74ce626077451 has some more tests that might be relevant.

### DIFF
--- a/tests/checks/complete.fish
+++ b/tests/checks/complete.fish
@@ -9,6 +9,8 @@ complete -c complete_test_alpha3 --no-files -w 'complete_test_alpha2 extra2'
 
 complete -C'complete_test_alpha1 arg1 '
 # CHECK: complete_test_alpha1 arg1
+complete --escape -C'complete_test_alpha1 arg1 '
+# CHECK: complete_test_alpha1\ arg1\
 complete -C'complete_test_alpha2 arg2 '
 # CHECK: complete_test_alpha1 extra1 arg2
 complete -C'complete_test_alpha3 arg3 '
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -94,6 +94,7 @@ Scripting improvements
 - ``command -v`` returns an exit status of 127 instead of 1 if no command was found (:issue:`8547`).
 - ``argparse`` with ``--ignore-unknown`` no longer breaks with multiple unknown options in a short option group (:issue:`8637`).
 - Comments inside command substitutions or brackets now correctly ignore parentheses, quotes, and brackets (:issue:`7866`, :issue:`8022`).
+- ``complete -C`` supports a new ``--escape`` option, which turns on escaping in returned completion strings (:issue:`3469`).
 
 Interactive improvements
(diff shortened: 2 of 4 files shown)


tests/checks/complete.fish
```diff
@@ -9,6 +9,8 @@ complete -c complete_test_alpha3 --no-files -w 'complete_test_alpha2 extra2'
 
 complete -C'complete_test_alpha1 arg1 '
 # CHECK: complete_test_alpha1 arg1
+complete --unescaped -C'complete_test_alpha1 arg1 '
+# CHECK: complete_test_alpha1\ arg1\
```
krobelus, line 13: The `--unescaped` option makes the output escaped, so it should be called `--escape` (I think that's better than `--escaped`)

krobelus (commented):
Looks good; the test is simple and to the point, no need for more test.
I just have a suggestion on the naming.

nzig:
Squashed and updated the documentation.
There was a discussion about waiting until after the release. I'll let you folks decide.

Thanks for reviewing this!

zanchey:
We can take it for 3.4.0.

krobelus:
Thanks, this is a nice solution.
Now someone can update FZF plugins to use this if available (I guess using something like `complete -C/ --escape' >/dev/null 2>&1`)

This is how I tested it, feels pretty slick!

	bind \cg "commandline -rt (complete -C --escape|fzf|cut -d\t -f1)\ "

nzig:
I plan to use this in my [fork](https://github.com/nzig/fzf.fish) of fzf.fish

Issue #8644, opened 2022-01-15
Variadic functions for `math` and tinyexpr rewrite.


juntuu:
Hi,

I've played around with the tinyexpr included in fish. I rewrote it [here](https://github.com/juntuu/fish-shell/tree/te-refactor) using more of c++ (than the very c-like code), and added variadic function support.

Changes:
- direct evaluation since fish didn't use any features (e.g. variables) that required separate parse or optimise steps
- support for variadic functions (and example functions `minimum` and `maximum`)
- functions are stored in a `std::unordered_map`, so that alphabetical order is no longer required and the lookup is simpler

Would this be of interest to you? If you'd like, I could open a pull request.

The current implementation of variadic functions doesn't support reporting errors from the functions, which might be good to have at some point. Also a question: Should some existing functions be made variadic, or should some variadic functions be added?

faho:
>Would this be of interest to you?

Sure, sounds cool. How's the performance?

>Should some existing functions be made variadic

I think the only ones where that really makes sense are minimum and maximum, I don't think e.g. "log2" makes sense as a variadic function.

There are possibilities here for *optional* arguments, e.g. `log(num, base)` or `round(num, digits)`, but I wouldn't add them initially because I'm not entirely sold on them.

>The current implementation of variadic functions doesn't support reporting errors from the functions

I'm not sure what you mean here. Things like division by zero? E.g. the function itself makes no sense to apply?

juntuu:
> How's the performance?

Running the math benchmark shows a small speedup on my machine: 1.02 ± 0.01 times faster. Otherwise I haven't noticed any significant difference, nor would I expect to see one. I also ran the benchmark with few other expressions (using other operators and including function calls) with very similar results.

With enough arguments (1000 +) the variadic functions e.g. `minimum n1,n2,...,nk` start getting faster than equivalent expression using two argument functions `min n1, min n2, ..., min n(k-1), nk`. I would assume the difference comes from the number of function lookups and calls.

> I think the only ones where that really makes sense are minimum and maximum

Do you think these are better to have separately, or changing the current `min` and `max` to be variadic?

> I'm not sure what you mean here.

I was mostly thinking if the function requires at least some number of arguments, or has other limitations on the number of arguments. I agree the optional arguments might not be a good choice, but that would be an example of variadic function with extra arity requirements.

floam:
[...]

PR #8646, opened 2022-01-15
Variadic functions for `math` and tinyexpr rewrite.

juntuu:
This patch adds support for variadic functions in the math builtin and `min` and `max` functions are made variadic.

This also introduces the following changes to tinyexpr, that are not visible to the user:

- direct evaluation since fish didn't use any features (e.g. variables) that required separate parse or optimise steps
- ~functions are stored in a std::unordered_map, so that alphabetical order is no longer required and the lookup is simpler~
- use of c++ features like overloaded functions, compared to the previous c-like code

See issue: #8644 

## TODOs:


src/tinyexpr.cpp
```diff
@@ -199,53 +195,63 @@ static double min(double a, double b) {
...
+    {L"e", M_E},
+    {L"exp", std::exp},
+    {L"fac", fac},
+    {L"floor", std::floor},
+    {L"ln", std::log},
+    {L"log", std::log10},
+    {L"log10", std::log10},
+    {L"log2", std::log2},
+    {L"max", max},
+    {L"maximum", maximum},
```
faho, line 235: These should be one function `max` that is variadic. Same for `min`.

At least I can't see any point for having both.

tests/checks/math.fish
```diff
@@ -140,9 +140,9 @@ not math '(1 pi)'
 # CHECKERR: '(1 pi)'
 # CHECKERR:    ^
 not math '(1 pow 1,2)'
-# CHECKERR: math: Error: Too many arguments
+# CHECKERR: math: Error: Missing operator
```
faho, line 143: Nice that we get this error improved!

src/tinyexpr.cpp
```diff
@@ -259,88 +265,76 @@ static constexpr double divide(double a, double b) {
...
+                    error_ = TE_ERROR_UNKNOWN_FUNCTION;
                 }
             } else {
                 /* Look for an operator or special character. */
-                switch (s->next++[0]) {
-                    // The "te_fun2" casts are necessary to pick the right overload.
+                switch (next_++[0]) {
                     case '+':
-                        s->type = TOK_INFIX;
-                        s->function = reinterpret_cast<void *>(static_cast<te_fun2>(add));
```
faho, line 312: Anything that gets us away from that casting is a win in my book!

src/tinyexpr.cpp
```diff
@@ -199,53 +195,63 @@ static double min(double a, double b) {
...
-    {L"sinh", reinterpret_cast<void *>(static_cast<te_fun1>(std::sinh)), TE_FUNCTION1},
-    {L"sqrt", reinterpret_cast<void *>(static_cast<te_fun1>(std::sqrt)), TE_FUNCTION1},
-    {L"tan", reinterpret_cast<void *>(static_cast<te_fun1>(std::tan)), TE_FUNCTION1},
-    {L"tanh", reinterpret_cast<void *>(static_cast<te_fun1>(std::tanh)), TE_FUNCTION1},
-    {L"tau", reinterpret_cast<void *>(static_cast<te_fun0>(tau)), TE_FUNCTION0},
+static constexpr double kMaximumContiguousInteger =
+    double(1LLU << std::numeric_limits<double>::digits) - 1;
+
+static double maximum(const std::vector<double> &args) {
+    double ret = -kMaximumContiguousInteger;
```
faho, line 202: This will return -kMaximumContiguousInteger if called with no arguments.

That doesn't seem like a good idea. The best I can come up with is interpreting "variadic" as meaning "at least one" for now, and erroring out when no argument is given. 

That seems doable, because "maximum" really only makes sense if you have at least one (arguably two), and so we could just allow variadic functions that allow zero arguments later, if that need ever shows up.

(same with minimum, of course)

faho (changes requested):
Nice work!

Two big things:

1. No difference between max/maximum and min/minimum
2. An error if they're called with no arguments

src/tinyexpr.cpp
andmis: The maximum of an empty set is usually defined to be minus infinity. (And the minimum of an empty set to be infinity.) This is often useful in formal math. But yeah I think that 99% of the time, a user who somehow passes an empty set to `max` did it by accident, rather than with the expectation that `max({}) = -infty`.

juntuu: This was intentional choice, as it is the identity element *within the allowed range*. Infinity might be more correct, but it's not a valid value and thus I can't see any use for it.

I agree, that this could be surprising for many users and more often a mistake, and so requiring at least one argument would be better. This also removes one special case from parsing.

juntuu: Changed it now to require at least one argument in the parser.

I didn't add this assumption to the function implementations though, to keep it simple. But I switched to infinity as the identity value now, as these would otherwise clamp the values to the representable range from one side (and `min max X` would clamp `X` from both sides).

faho:
It's worth testing how this handles nested function calls with multiple arguments.

E.g. I would imagine `math min 5, sin 2` works, while

```fish
math 'min 5, 1, 3, ncr 5, 2, 17
```

is an error because it binds the `, 2` to the `min`? So it would require parentheses (`ncr(5, 2)`). That's okay, but we should check that in the tests so we don't change it later on.

faho:
>And that logic hasn't changed for non-variadic functions.

>To stop parsing after expected number of arguments the condition should be:

To be clear: I am not asking for a change in behavior here. This behavior is acceptable. We have to pick one, and this is okay.

It's just that it wasn't entirely clear to me how variadic functions would interact here, and we should nail that down so we don't change it accidentally.

juntuu:
>  I am not asking for a change in behavior here.

This is what I understood from your earlier comment. I just wanted to understand the reasons for the current behaviour better, as it wasn't quite clear for me.

The only difference with variadic functions is that those always consume any remaining arguments, and thus those would behave like the `ncr` does when it is at the end.

In other words, function call within argument list requires parentheses unless the it is the last argument. And variadic function will always be last (unless parenthesised), because it consumes all available arguments.

faho:
Merged, thanks!

PR #8651, opened 2022-01-17
completion: add completions

dundargoc:


floam:
If you can't figure out a way to get from the tool itself a list of those steps used by two of the options, I'd like to see a comment in here explaining exactly how to derive it, since it seems like they add and drop these not super infrequently.

Also, are these all valid on every OS? A quick look at linux.rs, macos.rs, etc. in the topgrade code has me wondering if for example if the Mac App Store "mas" step is even valid on Linux or DragonFlyBSD, or if steps in linux.rs can be added or excluded on a Mac.

zanchey:
Thanks!

PR #8662, opened 2022-01-22
tar completion: added useful options

exploide:
## Description

Added the incredibly useful option `--one-top-level` to `tar` completions.

`--help` and `--usage` options as well.

And a bit of cleanup for functionality that is available under different long option names.

## TODOs:


zanchey:
Some of these are GNU-only (but other existing options are too).

krobelus:
Agreed, thanks.
We should detect if the options are actually supported but that was already an issue before.

PR #8665, opened 2022-01-24
Update some git completions

spenserblack:
## Description

- Adds `--[no-]recurse-submodules` for `git checkout`.
- Only shows `--abort`, `--skip`, and `--continue` for `git rebase` when a rebase is in progress, because they aren't applicable otherwise. Also adds `edit-todo` with same rule.
- Adds `--prune` for `git fetch` and `git pull`.
- Completes filenames with `git show -- `.

Fixes issue: no specific issue

## TODOs:

### DIFF
--- a/share/completions/git.fish
+++ b/share/completions/git.fish
@@ -798,6 +798,10 @@ format:\tSpecify which information to show"
     end
 end
 
+function __fish_git_is_rebasing
+    test -e (__fish_git rev-parse --git-path rebase-merge)
+end
+
 # general options
 complete -f -c git -l help -d 'Display the manual of a git command'
 complete -f -c git -n __fish_git_needs_command -l version -d 'Display version'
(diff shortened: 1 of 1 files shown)


share/completions/git.fish
```diff
@@ -798,6 +798,10 @@ format:\tSpecify which information to show"
     end
 end
 
+function __fish_git_is_rebasing
+    test -e (__fish_git rev-parse --git-path rebase-merge)
+end
```
spenserblack, line 803: :thinking: It looks like `git -C path/to/other/repo --git-path rebase-merge` resolves to `.git/rebase-merge` instead of `path/to/other/repo/.git/rebase-merge`. So this is actually only hiding options like `git rebase --continue` if the *current* directory contains `.git/rebase-merge`, instead of the actual repo. So it might be best to just remove this check.

spenserblack: Nevermind, I think I figured it out, but this check might still be overkill.

share/completions/git.fish
```diff
@@ -1700,10 +1709,11 @@ complete -f -c git -n '__fish_git_using_command rebase' -a '(__fish_git_branches
...
 complete -f -c git -n '__fish_git_using_command rebase' -a '(__fish_git_recent_commits)'
 complete -f -c git -n '__fish_git_using_command rebase' -a '(__fish_git_tags)' -d Tag
-complete -f -c git -n '__fish_git_using_command rebase' -l continue -d 'Restart the rebasing process'
-complete -f -c git -n '__fish_git_using_command rebase' -l abort -d 'Abort the rebase operation'
+complete -f -c git -n '__fish_git_using_command rebase; and __fish_git_is_rebasing' -l continue -d 'Restart the rebasing process'
+complete -f -c git -n '__fish_git_using_command rebase; and __fish_git_is_rebasing' -l abort -d 'Abort the rebase operation'
+complete -f -c git -n '__fish_git_using_command rebase; and __fish_git_is_rebasing' -l edit-todo -d 'Edit the todo list'
 complete -f -c git -n '__fish_git_using_command rebase' -l keep-empty -d "Keep the commits that don't change anything"
-complete -f -c git -n '__fish_git_using_command rebase' -l skip -d 'Restart the rebasing process by skipping the current patch'
+complete -f -c git -n '__fish_git_using_command rebase; and __fish_git_is_rebasing' -l skip -d 'Restart the rebasing process by skipping the current patch'
```
spenserblack, line 1716: Hiding these options unless the user is rebasing might be overkill. Please let me know what you think :slightly_smiling_face:

share/completions/git.fish
```diff
@@ -798,6 +798,10 @@ format:\tSpecify which information to show"
     end
 end
 
+function __fish_git_is_rebasing
+    count (string join "/" (__fish_git rev-parse --show-toplevel --git-path rebase-))* > /dev/null
```
krobelus, line 802: Very nice changes overall, thanks!

this commit can probably be dropped.
.git/rebase-merge exists during interactive rebase, while
.git/rebase-apply exists during non-interactive rebase. But --continue/--abort etc are not valid there.

krobelus, line 802: The problem here is that `--git-path` will print absolute paths when inside a secondary worktree (`git worktree add`).
How about just using `--absolute-git-dir`? I don't think there are any relocation variables for the `rebase-merge` directory.

share/completions/git.fish
krobelus: > Hiding these options unless a rebase is in progress might be overkill. Please let me know what you think slightly_smiling_face

I think it's a great idea

share/completions/git.fish
spenserblack: Thanks for the suggestion! Applied in https://github.com/fish-shell/fish-shell/pull/8665/commits/9554cc8d64e8260de2e18b0bd434cf7ce38883e4.

PR #8672, opened 2022-01-28
Fix gradle tasks completion

NextAlone:
## Description

Fix for tasks like `TMessagesProj:assembleMiniRelease` 

Fixes issue #

## TODOs:

### DIFF
--- a/share/completions/gradle.fish
+++ b/share/completions/gradle.fish
@@ -54,7 +54,7 @@ function __fish_gradle_get_task_completion
 
     set -l gradle_cache_file (__fish_gradle_create_completion_cache_file "{$PWD}-tasks")
     if not test -f "$gradle_cache_file" -a -s "$gradle_cache_file"
-        command gradle -q tasks --all 2>/dev/null | string match --regex '^[a-z][A-z:]+.*' | string replace ' - ' \t >"$gradle_cache_file"
+        command gradle -q tasks --all 2>/dev/null | string match --regex '^(?!-)[A-z:-]+(\s-\s)+.*|^(?!-)[A-z0-9:-]+$' | string replace ' - ' \t >"$gradle_cache_file"
     end
 
     # return possible tasks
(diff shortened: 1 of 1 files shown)


share/completions/gradle.fish
```diff
@@ -54,7 +54,7 @@ function __fish_gradle_get_task_completion
 
     set -l gradle_cache_file (__fish_gradle_create_completion_cache_file "{$PWD}-tasks")
     if not test -f "$gradle_cache_file" -a -s "$gradle_cache_file"
-        command gradle -q tasks --all 2>/dev/null | string match --regex '^[a-z][A-z:]+.*' | string replace ' - ' \t >"$gradle_cache_file"
+        command gradle -q tasks --all 2>/dev/null | string match --regex '^(?!-)[A-z:-]+(\s-\s)+.*|^(?!-)[A-z0-9:-]+$' | string replace ' - ' \t >"$gradle_cache_file"
```
krobelus, line 57: set -l gradle_cache_file (__fish_gradle_create_completion_cache_file "{$PWD}-tasks")
     if not test -f "$gradle_cache_file" -a -s "$gradle_cache_file"
-        command gradle -q tasks --all 2>/dev/null | string match --regex '^[a-z][A-z:]+.*' | string replace ' - ' \t >"$gradle_cache_file"
+        command gradle -q tasks --all 2>/dev/null | string match --regex '^(?!-)[A-z:-]+(\s-\s)+.*|^(?!-)[A-z0-9:-]+$' | string replace ' - ' \t >"$gradle_cache_file"


A-z is likely wrong because it includes characters like [ ] and \ (see `man ascii`)
we probably want a-zA-Z instead. Also I think we can simplify the regex:

```suggestion
        command gradle -q tasks --all 2>/dev/null | string match --regex '^(?!-)[A-Za-z0-9:-]+( - .*)?$' | string replace ' - ' \t >"$gradle_cache_file"
```

NextAlone: image
Now it will match the description twice, and i will correct it later

NextAlone: or just `^(?!-)[A-Za-z0-9:-]+ - .*$|^(?!-)[A-Za-z0-9:-]+$`

PR #8673, opened 2022-01-28
refactor: declare variable to prevent warning

dundargoc:
It silences a "Maybe uninitiliazed" warning when building.

### DIFF
--- a/src/reader.cpp
+++ b/src/reader.cpp
@@ -233,7 +233,8 @@ bool editable_line_t::undo() {
     maybe_t<int> last_group_id{-1};
     while (undo_history.edits_applied != 0) {
         const edit_t &edit = undo_history.edits.at(undo_history.edits_applied - 1);
-        if (did_undo && (!edit.group_id.has_value() || edit.group_id != last_group_id)) {
+        bool edit_group_is_last_group = edit.group_id != last_group_id;
+        if (did_undo && (!edit.group_id.has_value() || edit_group_is_last_group)) {
             // We've restored all the edits in this logical undo group
             break;
         }


src/reader.cpp
```diff
@@ -233,7 +233,8 @@ bool editable_line_t::undo() {
     maybe_t<int> last_group_id{-1};
     while (undo_history.edits_applied != 0) {
         const edit_t &edit = undo_history.edits.at(undo_history.edits_applied - 1);
-        if (did_undo && (!edit.group_id.has_value() || edit.group_id != last_group_id)) {
+        bool edit_group_is_last_group = edit.group_id != last_group_id;
+        if (did_undo && (!edit.group_id.has_value() || edit_group_is_last_group)) {
```
krobelus, line 237: I don't understand. gcc agrees that everything is initialized without this change.
How do you get the warning?

dundargoc: Weirt. This is the function I'm using.

```fish
function build-fish
    set fish_path "$HOME/programs/fish-shell"
    set fish_build_path "$fish_path/build"

    cmake -S $fish_path -B $fish_build_path -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DCMAKE_BUILD_TYPE=Release -G Ninja
    cmake --build $fish_build_path

    command cp $fish_build_path/compile_commands.json $fish_path
end
```

dundargoc: The exact error message:

```
/home/dundar/programs/fish-shell/src/reader.cpp: In member function ‘bool editable_line_t::undo()’:
/home/dundar/programs/fish-shell/src/reader.cpp:236:53: warning: ‘*(int*)((char*)&last_group_id + offsetof(maybe_t<int>,maybe_t<int>::impl_.maybe_detail::maybe_impl_t<int>::storage[0]))’ may be used uninitialized in this function [-Wmaybe-uninitialized]
  236 |         if (did_undo && (!edit.group_id.has_value() || edit.group_id != last_group_id)) {
```

krobelus: I remember getting the same warning, also for a `maybe_t` variable. This was on a modern gcc, and
the variable was definitely initialized. I never figured out what the problem was.
Let's try again.
If you can reproduce with a `docker/docker_run_tests.sh` command, that would be greatly helpful.
Maybe write a new Dockerfile that matches your environment.

dundargoc: Yep, I think you're absolutely right. I can't recreate this on gcc 9.3.0 so it seems to be a recently introduced bug in gcc. I guess I should close this PR since this isn't related to fish?

krobelus: I get no warning on gcc 11.1.0 (using `touch src/reader.cpp && ninja -C build`)

> I guess I should close this PR since this isn't related to fish?

Yes, I think it's not serious enough to warrant a workaround.
If the gcc bug is not already fixed we should report it to their bug tracker.

dundargoc: [They know :)](https://gcc.gnu.org/bugzilla/buglist.cgi?quicksearch=may%20be%20uninitialized)

Issue #3126, opened 2016-06-08
Make autosuggestions from history case-insensitive


phrolov:
For instance, I have directory named `Projects` in current workind directory. When typing `cd pro` fish suggests `Projects/` as possible completion based on directory name. But when I type `cd Pro` shell suggests `Projects/myproject` string from history.

I think it would be more convenient to search for possible completions from history case-insensitively.

---
Fish version: 2.3.0
Operating system: OS X, Homebrew
Terminal or terminal emulator: OS X Terminal

krader1961:
Seems reasonable to me. Most search engines default to case-insensitive and may not even provide a case-sensitive search option. Having fish behave the same way is likely to be less surprising, and thus friendlier, for people used to Google, etc.

nitsakh:
Hello, I wish to work on this bug. I haven't contributed to fish-shell before.

krader1961:
> I haven't contributed to fish-shell before.

That's okay. We're always interested in seeing people contribute to the core C++ code. We're happy to get you pointed in the right direction by answering questions specific to this project such as how to run unit tests, acceptable code style, etcetera. We're not, however, going to teach you how to use git or program in C++ :smile:

krader1961:
See also issue #2538 which is closely related.

tweetjay:
Hi, just wanted to let you know that I really like this behaviour! If I want to lookup the history case insensitive - I just type small letters. And since most of my Directory paths are starting uppercase I can easily search for them with the first letter uppercase.
So if this is going to be changed, please make it configurable.

nitsakh:
Sorry! I've been pretty busy lately and haven't got to this yet. If someone else is willing to take this up, feel free to do so.

BK1603:
Yeah sure, I think the filenames could be case sensitive, also like while I am at it, I have one other idea/feature request, like while we are in the middle of an incomplete or a half written command, and the shell suggests a completion, could some keybinding automatically execute it? I won't say that pressing enter should 'cause, well, a command like `cd` could have a lot of suggestions but would work on its own too..

ridiculousfish:
@BK1603 you can make your own key binding if you want to accept and execute. For example here with control-F:

    bind \cf accept-autosuggestion execute

mqudsi:
Any fix for this should respect existing fish convention with regards to case: case insensitive options should only be shown if no case-matching options are extant/viable.

ridiculousfish:
[...]

Issue #3978, opened 2017-04-22
Case-insensitive tab completion


benmullins:
I'd like to suggest an option to make tab completion fully case-insensitive. What I mean, for example, is the following situation: I have in my home directory two subdirectories, one starting with uppercase P and the other with lowercase p. For the last year or so I've been using `zsh` with oh-my-zsh, and I've grown accustomed to getting into the directory starting with uppercase P by typing `cd p`, hitting tab twice, and choosing the correct directory of the two.

Besides that inconvenience, the current behavior feels somewhat inconsistent to me. If I type, for example, `cd N`, it matches directories starting with lowercase n, and `cd v` matches directories starting with uppercase V --- because I don't have directories starting with that letter in the other case. So essentially it matches case-insensitively sometimes but not always; specifically iff there are no case-sensitive matches. This confused me a fair bit when I first discovered the behavior, since I was using fish for a little while before I ran into the `cd p` issue I described, and it took me a while to realize why it was immediately auto-completing to the lowercase p directory without giving me a choice.

I understand why this behavior might be undesirable to some people since Unix filesystems are generally case-sensitive, and I know that configurability is the root of all evil and all, but I found this to be something of a pain point while trying fish.

fish version: 2.5.0, stock standard from Homebrew
OS: macOS 10.12.4
Terminal: Terminal.app

faho:
First some facts and definitions:

I'd call what we currently have "case-correcting". I.e. when what you entered _can't_ match a completion, fish tries matches that only differ by case.

An obvious variation on this theme (inspired by vim's "smartcase" search option) is to do case-insensitive matching if the input is all lower-case, and case-sensitive otherwise. Let's call this "smartcase".

Compare #1196, which asks for case-insensitivity specifically for the "fuzzy" completions.

------

The advantage of case-sensitive searching (CSS) is that you can get down to fewer options faster _iff_ you know what case what you are looking for is.

The advantage of case-insensitive searching (CIS) is that you get more options if you don't know (or remember) the case.

The failure mode of CSS (i.e. you enter something in the wrong case) is that you either get the wrong or no option - case-correction removes that "no" options case. When this happens, you need to remove the already picked option (if any) and then change the case of the offending character. But to do that, you need to _know_ which character is wrong.

The failure mode of CIS is that you get more options and need to pick from them. Which is something that you already need to be prepared to do, and which can be made easier with e.g. a nice searchable pager (:wink:).

[...]

PR #8678, opened 2022-01-31
Changing autosuggestions from search history to be case-insensitive

skhatib07:
## Description
Edited the suggestion when the case of a command being typed is not the same as the case of the command from history to not consider the case of the character in the comparison between the two (prefer case insensitive to smart case).
Fixes issue #3126

## TODOs:

### DIFF
--- a/src/complete.cpp
+++ b/src/complete.cpp
@@ -346,11 +346,13 @@ void completions_sort_and_prioritize(completion_list_t *comps, completion_reques
 
     // Lastly, if this is for an autosuggestion, prefer to avoid completions that duplicate
     // arguments, and penalize files that end in tilde - they're frequently autosave files from e.g.
-    // emacs. Also prefer samecase to smartcase.
+    // emacs. Also prefer case insensitive to smartcase.
     if (flags & completion_request_t::autosuggestion) {
         stable_sort(comps->begin(), comps->end(), [](const completion_t &a, const completion_t &b) {
             if (a.match.case_fold != b.match.case_fold) {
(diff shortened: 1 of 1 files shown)


mqudsi:
I don't think preferring case insensitive *over* smart case is what you want - they're discriminated cases of a case comparison, and icase is the last resort after both samecase and smartcase have failed. Did you mean to eliminate the sort-by-case_fold entirely?

In all cases, your modified sort function isn't preferring anything over anything else on the basis of `case_fold`: (`is_less_than(a { case_fold = smartcase }, b { case_fold = icase })` gives you `false` and then `is_less_than(b { case_fold = icase }, a { case_fold = smartcase })` also gives you false).

This would definitely need a test to accompany it and a better comment to explain what the goal is in the code, because as it stands it doesn't mention the *why* aspect of it at all.

krobelus:
I think history autosuggestion should remain case-sensitive.
https://github.com/fish-shell/fish-shell/issues/3978 has some discussion that led to today's behavior.

> When typing cd pro I want the autosuggestion to be the same as typing cd Pro or typing cd PRO

*file* autosuggestion is already case-insensitive, so that should already work.

floam:
> file autosuggestion is already case-insensitive, so that should already work.

Not sure about that. I get different results.

floam:
Oh, well, history autosuggestions can look like file autosuggestions.

faho:
Okay, this doesn't work as-is and it's been 10 days without a reply, so I'm closing this.

If you get back to us we'll reopen!

PR #8687, opened 2022-02-01
completions: add wg-quick

ammgws:
## Description

Adds completions for the Wireguard `wg-quick` utility.

Fixes issue #

## TODOs:

### DIFF
--- a/share/completions/wg-quick.fish
+++ b/share/completions/wg-quick.fish
@@ -1,18 +1,12 @@
 set -l valid_subcmds up down strip save
 
-function __wg_complete_interfaces
-    wg show 2>| string replace -rf 'Unable to access interface (.*): Operation not permitted' '$1'
-end
-
-function __wg_subcmd --no-scope-shadowing
-    set -l subcmd $argv[1]
-    set -e argv[1]
-    complete -f -c wg-quick -n "not __fish_seen_subcommand_from $valid_subcmds" -a $subcmd $argv
+function __fish_wg_complete_interfaces
+    LC_ALL=C wg show 2>| string replace -rf 'Unable to access interface (.*): Operation not permitted' '$1'
 end
 
 complete -c wg-quick -f
-complete -k -c wg-quick -n "__fish_seen_subcommand_from $valid_subcmds" -a '(__wg_complete_interfaces | sort)'
-__wg_subcmd up -d 'Add and set up an interface'
-__wg_subcmd down -d 'Tear down and remove an interface' 
-__wg_subcmd strip -d 'Output config for use with wg'
-__wg_subcmd save -d 'Saves the configuration of an existing interface'
+complete -c wg-quick -n "__fish_seen_subcommand_from $valid_subcmds" -a '(__fish_wg_complete_interfaces)'
+complete -c wg-quick -n "not __fish_seen_subcommand_from $valid_subcmds" -a up -d 'Add and set up an interface'
+complete -c wg-quick -n "not __fish_seen_subcommand_from $valid_subcmds" -a down -d 'Tear down and remove an interface'
+complete -c wg-quick -n "not __fish_seen_subcommand_from $valid_subcmds" -a strip -d 'Output config for use with wg'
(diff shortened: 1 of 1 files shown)


share/completions/wg-quick.fish
```diff
@@ -0,0 +1,18 @@
...
+end
+
+function __wg_subcmd --no-scope-shadowing
+    set -l subcmd $argv[1]
+    set -e argv[1]
+    complete -f -c wg-quick -n "not __fish_seen_subcommand_from $valid_subcmds" -a $subcmd $argv
+end
+
+complete -c wg-quick -f
+complete -k -c wg-quick -n "__fish_seen_subcommand_from $valid_subcmds" -a '(__wg_complete_interfaces | sort)'
```
faho, line 14: Just remove the `-k` here instead of needing to call `sort`.

share/completions/wg-quick.fish
```diff
@@ -0,0 +1,18 @@
+set -l valid_subcmds up down strip save
+
+function __wg_complete_interfaces
+    wg show 2>| string replace -rf 'Unable to access interface (.*): Operation not permitted' '$1'
```
faho, line 4: Ooof, this seems... awkward. Is there really no better way than to parse an error message?

If you get an error message, is the argument actually useful?

share/completions/wg-quick.fish
```diff
@@ -0,0 +1,18 @@
...
+end
+
+function __wg_subcmd --no-scope-shadowing
+    set -l subcmd $argv[1]
+    set -e argv[1]
+    complete -f -c wg-quick -n "not __fish_seen_subcommand_from $valid_subcmds" -a $subcmd $argv
```
faho, line 10: Please just repeat this further down.

This is a fairly awkward abstraction, removing it makes things more normal and simpler.

faho (changes requested):

share/completions/wg-quick.fish
ammgws: `wg-quick` looks for interfaces in `/etc/wireguard/`, which has `700` permissions so we cannot get completion candidates from there. This was the only way I could find that works - however I guess will need to prefix with `LC_ALL=C` here. `wg show` always shows the name of the currently connected interface in its error message, so it is perfect as the completion in this case.

krobelus:
On Tue, Feb 08, 2022 at 08:56:14PM -0800, Jason Nader wrote:
> +function __wg_complete_interfaces

I think we want to name this __fish_wg_complete_interfaces

share/completions/wg-quick.fish
faho: >wg-quick looks for interfaces in /etc/wireguard/, which has 700 permissions so we cannot get completion candidates from there.

Okay, wait... what happens if you are the owner (I'm assuming root?)?

What does `wg show` say then?

faho: Ping @ammgws: Does this work if you're root? I'm assuming it doesn't error out then, but do the completions still work?

faho:
Merged, thanks!

ammgws:
Sorry I never got around to checking what happens when you're `root`

share/completions/wg-quick.fish
ammgws: @faho Sorry for the late reply. It doesn't error out, just gives garbage candidates for the completion:
```sh
>root@mydesktop ~# wg-quick down [TAB]
  allowed ips: 0.0.0.0/0, ::/0        private key: (hidden)                                   
  endpoint: xxx:51820                 public key: xxxx
  fwmark: 0xca6c                      transfer: 414.65 KiB received, 141.07 KiB sent          
  latest handshake: 50 seconds ago    interface: <completion target>                                     
  listening port: 49937               peer: xxxx
 ```

Comment (ammgws) [reply share/completions/wg-quick.fish]: Actually it looks like that works for non-root too, so we can just get rid of the kluge altogether:

```sh
>wg show
Unable to access interface mlvd-ch2: Operation not permitted
[1]>wg show interfaces
my_wg_interface
```

PR #8692, opened 2022-02-02
Prevent signals from tearing multi-char bindings

### LINKED ISSUE #8628: Improper handling of combined focus+resize events on master and 3.3.1
In tmux with `$ tmux set focus-events on` fish will put "garbage" in the command line when it receives/loses focus after closing/creating a split window.

Bisect:
```
7c153a8307c662fa853280fd64657dff0cd9a625 is the first bad commit
commit 7c153a8307c662fa853280fd64657dff0cd9a625
Author: ridiculousfish <rf@fishshell.com>
Date:   Sat Apr 10 20:41:33 2021 -0700

    Introduce prepare_to_select, fixing job_summary.py

    This concerns printing status messages for background jobs which have
    stopped or finished. Previously fish would do this from two places:

    1. Before running a command (including empty string)
    2. If a signal is received during select()

    So if the job finishes while fish is doing something else (like running an
    event handler) then we would not print status messages until the user hit
    return. This caused the job_summary.py test to be flaky.

    Fix this by splitting the interrupt handler into two parts: a part that
    handles signals (e.g. triggering exit from the reader), and a part that
    always runs just before blocking in select(). This second part always
    reaps jobs and prints their status messages. This narrows the window for a
    job exit to be "missed" before fish blocks in select, and should make the
    job_summary.py test more reliable.

 src/input.cpp        | 13 +++++++++----
 src/input.h          |  3 +++
 src/input_common.cpp |  7 +++++++
 src/input_common.h   |  4 ++++
 4 files changed, 23 insertions(+), 4 deletions(-)
```

Tested on both Arch Linux and NixOS unstable.

Recording of observed behaviour in the linux tty on a clean nixos vm: https://asciinema.org/a/ODR7cnvio1vcaaZIdvFfe3oiJ

See https://github.com/fish-shell/fish-shell/issues/8628#issuecomment-1009617569 for steps to reproduce.

ridiculousfish:
Say the user has a multi-char binding (typically an escape sequence), and a signal arrives partway through the binding. The signal has an event handler which enques some readline event, for example, `repaint`. Prior to this change, the readline event would cause the multi-char binding to fail. This would cause bits of the escape sequence to be printed to the screen.

Fix this by noticing when a sequence was "interrupted" by a non-char event, and then rotating a sequence of such interruptions to the front of the queue, allowing them to be handled and still anticipating the remaining characters in the binding.

Fixes #8628

### DIFF
--- a/src/input.cpp
+++ b/src/input.cpp
@@ -673,9 +673,6 @@ void inputter_t::mapping_execute_matching_or_generic(const command_handler_t &co
 
     FLOGF(reader, L"no generic found, ignoring char...");
     auto evt = peeker.next();
-    if (evt.is_eof()) {
-        this->push_front(evt);
-    }
     peeker.consume();
 }
 
@@ -750,7 +747,11 @@ char_event_t inputter_t::read_char(const command_handler_t &command_handler) {
             // If we have EOF, we need to immediately quit.
             // There's no need to go through the input functions.
             return evt;
+        } else if (evt.is_check_exit()) {
+            // Allow the reader to check for exit conditions.
+            return evt;
         } else {
+            assert(evt.is_char() && "Should be char event");
             this->push_front(evt);
             mapping_execute_matching_or_generic(command_handler);
             // Regarding allow_commands, we're in a loop, but if a fish command is executed,
--- a/src/input.cpp
+++ b/src/input.cpp
@@ -504,6 +504,14 @@ class event_queue_peeker_t {
         idx_ = 0;
     }
 
+    /// Test if any of our peeked events are readline or check_exit.
+    bool char_sequence_interrupted() const {
+        for (const auto &evt : peeked_) {
+            if (evt.is_readline() || evt.is_check_exit()) return true;
+        }
+        return false;
+    }
+
     /// Reset our index back to 0.
     void restart() { idx_ = 0; }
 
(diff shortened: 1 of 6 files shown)


src/input.cpp
```diff
@@ -673,9 +673,6 @@ void inputter_t::mapping_execute_matching_or_generic(const command_handler_t &co
 
     FLOGF(reader, L"no generic found, ignoring char...");
     auto evt = peeker.next();
-    if (evt.is_eof()) {
```
ridiculousfish, line 676: This was never true; this function is only called if the first event is a char.

krobelus (approved):

ridiculousfish:
Merged ending at 7f31acbf9b677ffcecea917970146de5e57c2dbb

page-down:
@ridiculousfish 

Can a similar treatment be applied to the vi mode prompt redraw?

```text
https://github.com/fish-shell/fish-shell/issues/3481#issuecomment-1020475058

> ... some overlap between the event handler and the cursor movement ...
> Try this:
> bind -M insert \e '... commandline -f repaint-mode backward-char; ...
```

I'm curious if it's the same kind of problem. Because the escape sequence is cut off, it results in a missing cursor up command when redrawing.

PR #8695, opened 2022-02-04
Fix comment parsing inside command substitutions and brackets

### LINKED ISSUE #7866: Is this a bug in the parser, such that it sees unbalances quotes within a comment in this context (command substitution?) as an error, or...?
Is this a bug in the parser, such that it sees unbalanced quotes within a comment (within this context (command substitution?) as an error, or..?

This works:
```fish
	echo (
	test ""
	and echo a # comment 1 ''
	or  echo b # comment 2 ''
	)
```

But this gets an error:
```fish
	echo (
	test ""
	and echo a # comment 1 '''
	or  echo b # comment 2 ''
	)
```

```
	fish: Unexpected end of string, quotes are not balanced
	and echo a # '
	             ^
```

(fish, version 2.7.1 ([ppa 3.2.1-1~focal](https://launchpad.net/~fish-shell/+archive/ubuntu/release-3)))

(
Also work:
```fish
	echo (
	test ""
	and echo a # comment 1 '''
	or  echo b # comment 2 '''
	)
```
```fish
	echo (
	test ""
	and echo a # comment 1 '
	or  echo b # comment 2 '
	)
```

and of course:
```fish
	echo (
	test ""
	and echo a # comment 1
	or  echo b # comment 2
	)
```

...

a few other notes:

:heavy_check_mark: works:
```fish
	echo (
	test ""
	and echo a # comment 1 '
	or  echo b # comment 2
	#'
	)
```

:x: fails:
```fish
	echo (
	test ""
	and echo a # comment 1 '
	or  echo b # comment 2
	) #'
```

:x: fails:
```fish
	echo (
	test ""
	and echo a # comment 1 '
	or  echo b # comment 2
	)
	#'
```

:heavy_check_mark: works:
```fish
	echo ( #'
	test ""
	and echo a # comment 1 '
	or  echo b # comment 2
	)
```
)

### LINKED ISSUE #8022: Bracket pairing breaks comments
Running fish version 3.2.2-325-g7123e2f25 (built from HEAD via Homebrew) on macOS 11.3.1 in the stock Terminal.app. I tried fish without third-party customisations by executing `sh -c 'env HOME=$(mktemp -d) fish'` and checked whether it affected the behaviour I am reporting (it did not).

---------

This seems related to #7970 and, unlike that issue, does not seem to be covered by #2485: **fish’s somewhat eccentric bracket pairing triggers across comment boundaries, effectively breaking comments.** I submit for your consideration the following code:

```fish
echo f[oo # not valid, no matching ]
```

To my best understanding, the argument to that `echo` command is  `f[oo`, followed by a “not valid, no matching ]” comment, and executing that snippet should trigger the “unexpected end of string” error, as there is an unmatched opening bracket that is not in first word character position. Note GitHub’s fish syntax highlighting concurs with my interpretation, for what that is worth. **However.**

What fish actually does is to treat this as a valid `f[oo # not valid, no matching ]` argument, which it executes without complaining. I find this utterly baffling, as [there is no paired bracket syntax construct outside variable indexing that justifies this](https://github.com/fish-shell/fish-shell/issues/7970#issuecomment-831411751).

The current behaviour can’t be right, can it?

shayaviv:
This change avoids parsing brackets and quotes within comments.

This fixes the following examples (as well as those described in #7866 and #8022):
| Script | Before | After |
| - | - | - |
| <pre lang="fish">echo a (&#xA;    echo b  # I'm using an apostrophe &#xA;)</pre> | (line 1): Unexpected end of string, quotes are not balanced | `a b` |
| <pre lang="fish">echo a (&#xA;    echo b  # Oops, closing paren :)&#xA;)</pre> | (line 3): Unexpected ')' for unopened parenthesis | `a b` |
| <pre lang="fish">echo a[b #c]</pre> | `a[b #c]` | Unexpected end of string, square brackets do not match |

Fixes #7866, and fixes #8022.

## TODOs:
  - _There are no changes in usage._

### DIFF
--- a/src/parse_util.cpp
+++ b/src/parse_util.cpp
@@ -90,6 +90,8 @@ size_t parse_util_get_offset(const wcstring &str, int line, long line_offset) {
 static int parse_util_locate_cmdsub(const wchar_t *in, const wchar_t **begin, const wchar_t **end,
                                     bool allow_incomplete, bool *inout_is_quoted) {
     bool escaped = false;
+    bool is_first = true;
+    bool is_token_begin = true;
     bool syntax_error = false;
     int paran_count = 0;
     std::vector<int> quoted_cmdsubs;
@@ -122,6 +124,10 @@ static int parse_util_locate_cmdsub(const wchar_t *in, const wchar_t **begin, co
         if (!escaped) {
             if (*pos == L'\'' || *pos == L'"') {
                 if (!process_opening_quote(*pos)) break;
+            } else if (*pos == L'\\') {
+                escaped = true;
+            } else if (*pos == L'#' && is_token_begin) {
+                pos = comment_end(pos) - 1;
             } else {
                 if (*pos == L'(') {
                     if ((paran_count == 0) && (paran_begin == nullptr)) {
@@ -161,12 +167,12 @@ static int parse_util_locate_cmdsub(const wchar_t *in, const wchar_t **begin, co
                     }
                 }
             }
-        }
-        if (*pos == '\\') {
-            escaped = !escaped;
+            is_token_begin = is_token_delimiter(pos[0], is_first, pos[1]);
         } else {
             escaped = false;
+            is_token_begin = false;
         }
+        is_first = false;
     }
 
     syntax_error |= (paran_count < 0);
--- a/src/tokenizer.cpp
+++ b/src/tokenizer.cpp
@@ -154,6 +154,7 @@ tok_t tokenizer_t::read_string() {
     int slice_offset = 0;
     const wchar_t *const buff_start = this->token_cursor;
     bool is_first = true;
+    bool is_token_begin = true;
 
     auto process_opening_quote = [&](wchar_t quote) -> const wchar_t * {
         const wchar_t *end = quote_end(this->token_cursor, quote);
@@ -192,6 +193,8 @@ tok_t tokenizer_t::read_string() {
         // has been explicitly ignored (escaped).
         else if (c == L'\\') {
             mode |= tok_modes::char_escape;
+        } else if (c == L'#' && is_token_begin) {
+            this->token_cursor = comment_end(this->token_cursor) - 1;
         } else if (c == L'(') {
             paran_offsets.push_back(this->token_cursor - this->start);
             expecting.push_back(L')');
@@ -278,8 +281,9 @@ tok_t tokenizer_t::read_string() {
         FLOGF(error, msg.c_str(), c, c, int(mode_begin), int(mode));
 #endif
 
-        this->token_cursor++;
+        is_token_begin = is_token_delimiter(this->token_cursor[0], is_first, this->token_cursor[1]);
         is_first = false;
+        this->token_cursor++;
     }
 
     if (!this->accept_unfinished && (mode != tok_modes::regular_text)) {
@@ -540,8 +544,7 @@ maybe_t<tok_t> tokenizer_t::next() {
     while (*this->token_cursor == L'#') {
         // We have a comment, walk over the comment.
         const wchar_t *comment_start = this->token_cursor;
-        while (this->token_cursor[0] != L'\n' && this->token_cursor[0] != L'\0')
-            this->token_cursor++;
+        this->token_cursor = comment_end(this->token_cursor);
         size_t comment_len = this->token_cursor - comment_start;
 
         // If we are going to continue after the comment, skip any trailing newline.
@@ -677,6 +680,10 @@ maybe_t<tok_t> tokenizer_t::next() {
     }
     assert(result.has_value() && "Should have a token");
     return result;
+}
+
+bool is_token_delimiter(wchar_t c, bool is_first, maybe_t<wchar_t> next) {
+    return c == L'(' || !tok_is_string_character(c, is_first, next);
 }
 
 wcstring tok_first(const wcstring &str) {
--- a/tests/checks/basic.fish
+++ b/tests/checks/basic.fish
@@ -533,6 +533,23 @@ echo banana
 echo (echo hello\\)
 # CHECK: hello\
 
+# This used to be a parse error - #7866.
+echo (echo foo;#)
+     )
+# CHECK: foo
+echo (echo bar #'
+     )
+# CHECK: bar
+echo (#"
+      echo baz)
+# CHECK: baz
+
+# Make sure we don't match up brackets within comments (#8022).
+$fish -c 'echo f[oo # not valid, no matching ]'
+# CHECKERR: fish: Unexpected end of string, square brackets do not match
+# CHECKERR: echo f[oo # not valid, no matching ]
+# CHECKERR: {{      }}^
+
 # Should fail because $PWD is read-only.
 for PWD in foo bar
     true
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -93,6 +93,7 @@ Scripting improvements
 - ``abbr -q`` returns the correct exit status when given multiple abbreviation names as arguments (:issue:`8431`).
 - ``command -v`` returns an exit status of 127 instead of 1 if no command was found (:issue:`8547`).
 - ``argparse`` with ``--ignore-unknown`` no longer breaks with multiple unknown options in a short option group (:issue:`8637`).
+- Comments inside command substitutions or brackets now correctly ignore parentheses, quotes, and brackets (:issue:`7866`, :issue:`8022`).
 
 Interactive improvements
 ------------------------
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -523,6 +523,13 @@ const wchar_t *quote_end(const wchar_t *pos, wchar_t quote) {
     return nullptr;
 }
 
+const wchar_t *comment_end(const wchar_t *pos) {
+    do {
+        pos++;
+    } while (*pos && *pos != L'\n');
+    return pos;
+}
+
 void fish_setlocale() {
     // Use various Unicode symbols if they can be encoded using the current locale, else a simple
     // ASCII char alternative. All of the can_be_encoded() invocations should return the same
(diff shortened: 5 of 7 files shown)


floam:
Well I'm a fan, nice job, cool!

The only thing I'd want to think on is if this is the simplest, best fix. These are parts of the codebase that are super convoluted.

shayaviv:
I refactored my changes - they are now the simplest I could get them without touching too much of the existing parsing code. I also added regression tests to `basic.fish`.

krobelus:
Looks pretty nice! :+1:

> Should we note this change?

I think so.

src/parse_util.cpp
```diff
@@ -167,6 +170,7 @@ static int parse_util_locate_cmdsub(const wchar_t *in, const wchar_t **begin, co
         } else {
             escaped = false;
         }
+        is_token_begin = !escaped && (iswspace(*pos) || *pos == L'(');
```
krobelus, line 173: This criteria is missing some other characters that separate tokens, like `;`.
`tok_is_string_character()` should give the complete answer (the API is a bit awkward, sorry).

(Using that, we also might be able to avoid the `- 1` in `comment_end(pos) - 1`)

tests/checks/basic.fish
```diff
@@ -533,6 +533,17 @@ echo banana
...
+# This used to be a parse error - #7866.
+echo (echo foo #)
+     )
+# CHECK: foo
+echo (echo bar #'
+     )
+# CHECK: bar
+echo (#"
+      echo baz)
+# CHECK: baz
```
krobelus, line 545: I think you're also fixing #8022, so how about adding

```fish
# Make sure we don't match up brackets within comments (#8022).
$fish -c 'echo f[oo # not valid, no matching ]'
# CHECKERR: fish: Unexpected end of string, square brackets do not match
# CHECKERR: echo f[oo # not valid, no matching ]
# CHECKERR: {{      }}^
```

src/parse_util.cpp
shayaviv: Thank, but I believe `;` does not separate comment tokens. Here's an example:

```fish
echo a  # ; echo b
```

According to the existing tokenizer.cpp code that finds the end of a comment token (which I extracted to `comment_end()`), the comment token should span all the way to a `L'\n'` or `L'\0'`: https://github.com/fish-shell/fish-shell/blob/964b7a729a7ea50485b9ce56861f5e922ac26a2a/src/tokenizer.cpp#L540-L545

As for the `- 1`, it's there because at the end of the loop iteration, `pos` is incremented (`pos++`) before being checked (`*pos != L'\0'`). Without the `- 1`, `pos` is incremented beyond a null delimiter, and then `*pos` is an an out-of-bounds read.

krobelus: I meant something else: a # will also start a token in `echo foo;#bar`, hence the check is to narrow

example test case:

```fish
echo (;#"
      echo baz)
# CHECK: baz
```

shayaviv: I see - thanks for catching that! I'll fix that.

krobelus:
Nice work! I have found one minor issue

Also, consider squashing the commits, so the changes are easier to follow for third parties (including
future you and future me).

src/tokenizer.cpp
```diff
@@ -679,6 +682,10 @@ maybe_t<tok_t> tokenizer_t::next() {
     return result;
 }
 
+bool is_token_delimiter(wchar_t c, bool is_first, maybe_t<wchar_t> next) {
+    return c == L'(' || !tok_is_string_character(c, is_first, next);
```
krobelus, line 686: +bool is_token_delimiter(wchar_t c, bool is_first, maybe_t<wchar_t> next) {
+    return c == L'(' || !tok_is_string_character(c, is_first, next);

Interesting. This function deviates from tok_is_string_character() because today `(a)(b)(c)` is a single token at parsing time. Only during execution, command substitutions are reparsed and further split.  We could change this in future.

Here is an alternative solution I played around with, but I think we should just discard it, because
`is_comment()` doesn't deal with backslash escapes itself but requires the caller to do so, which is weird.

<details>

```diff
diff --git a/src/parse_util.cpp b/src/parse_util.cpp
index c626804d0..82870c6d9 100644
--- a/src/parse_util.cpp
+++ b/src/parse_util.cpp
@@ -90,8 +90,6 @@ size_t parse_util_get_offset(const wcstring &str, int line, long line_offset) {
 static int parse_util_locate_cmdsub(const wchar_t *in, const wchar_t **begin, const wchar_t **end,
                                     bool allow_incomplete, bool *inout_is_quoted) {
     bool escaped = false;
-    bool is_first = true;
-    bool is_token_begin = true;
     bool syntax_error = false;
     int paran_count = 0;
     std::vector<int> quoted_cmdsubs;
@@ -126,7 +124,7 @@ static int parse_util_locate_cmdsub(const wchar_t *in, const wchar_t **begin, co
                 if (!process_opening_quote(*pos)) break;
             } else if (*pos == L'\\') {
                 escaped = true;
-            } else if (*pos == L'#' && is_token_begin) {
+            } else if (is_comment(pos, in)) {
                 pos = comment_end(pos) - 1;
             } else {
                 if (*pos == L'(') {
@@ -167,12 +165,9 @@ static int parse_util_locate_cmdsub(const wchar_t *in, const wchar_t **begin, co
                     }
                 }
             }
-            is_token_begin = is_token_delimiter(pos[0], is_first, pos[1]);
         } else {
             escaped = false;
-            is_token_begin = false;
         }
-        is_first = false;
     }

     syntax_error |= (paran_count < 0);
diff --git a/src/tokenizer.cpp b/src/tokenizer.cpp
index 861c57f61..9d48cdb7e 100644
--- a/src/tokenizer.cpp
+++ b/src/tokenizer.cpp
@@ -154,7 +154,6 @@ tok_t tokenizer_t::read_string() {
     int slice_offset = 0;
     const wchar_t *const buff_start = this->token_cursor;
     bool is_first = true;
-    bool is_token_begin = true;

     auto process_opening_quote = [&](wchar_t quote) -> const wchar_t * {
         const wchar_t *end = quote_end(this->token_cursor, quote);
@@ -193,7 +192,7 @@ tok_t tokenizer_t::read_string() {
         // has been explicitly ignored (escaped).
         else if (c == L'\\') {
             mode |= tok_modes::char_escape;
-        } else if (c == L'#' && is_token_begin) {
+        } else if (is_comment(this->token_cursor, buff_start)) {
             this->token_cursor = comment_end(this->token_cursor) - 1;
         } else if (c == L'(') {
             paran_offsets.push_back(this->token_cursor - this->start);
@@ -281,7 +280,6 @@ tok_t tokenizer_t::read_string() {
         FLOGF(error, msg.c_str(), c, c, int(mode_begin), int(mode));
 #endif

-        is_token_begin = is_token_delimiter(this->token_cursor[0], is_first, this->token_cursor[1]);
         is_first = false;
         this->token_cursor++;
     }
@@ -682,8 +680,12 @@ maybe_t<tok_t> tokenizer_t::next() {
     return result;
 }

-bool is_token_delimiter(wchar_t c, bool is_first, maybe_t<wchar_t> next) {
-    return c == L'(' || !tok_is_string_character(c, is_first, next);
+bool is_comment(const wchar_t *pos, const wchar_t *tok_start) {
+    if (*pos != L'#') return false;
+    if (pos == tok_start) return true;
+    wchar_t prev = *(pos - 1);
+    bool prev_is_first = (pos - 1) == tok_start;
+    return prev == L'(' || prev == L')' || !tok_is_string_character(prev, prev_is_first, *pos);
 }

 wcstring tok_first(const wcstring &str) {
diff --git a/src/tokenizer.h b/src/tokenizer.h
index fccff61db..5322a2d32 100644
--- a/src/tokenizer.h
+++ b/src/tokenizer.h
@@ -133,8 +133,8 @@ class tokenizer_t : noncopyable_t {
     }
 };

-/// Tests if this character can delimit tokens.
-bool is_token_delimiter(wchar_t c, bool is_first, maybe_t<wchar_t> next);
+/// Tests if the given position starts a comment (possible inside a command substitution).
+bool is_comment(const wchar_t *pos, const wchar_t *tok_start);

 /// Returns only the first token from the specified string. This is a convenience function, used to
 /// retrieve the first token of a string. This can be useful for error messages, etc. On failure,

Comment (krobelus) [inline src/tokenizer.cpp:686]: +bool is_token_delimiter(wchar_t c, bool is_first, maybe_t<wchar_t> next) {
+    return c == L'(' || !tok_is_string_character(c, is_first, next);

We also want closing parens, here is a test case:

    echo ()#)

Comment (shayaviv) [reply src/tokenizer.cpp:686]: I see two different ways to interpret the test case you provided, and both yield the same result:
* expand empty list into an `echo` command, appending `#)` to each arg → a newline is printed
  - This is how I intended my code change to interpret this.
* expand empty list into an `echo` command, followed by a comment → a newline is printed
  - I assume this is how you meant we should interpret this.

Let's look at a different example:
```fish
echo (echo a)#b
```
I assume `a#b` should be printed in this case, since that is the current behavior. It also makes sense to me, since this interpretation is consistent with `echo a#b`. Do you agree?

krobelus: Sorry, you're right, # is not a comment after )
My bad -- using ()#) as example was foolish, since it is trivially empty because of cartesion product expansion

shayaviv: I like your solution! My only reservation is that `)` is interpreted as a delimiter - as I said in my [previous comment](https://github.com/fish-shell/fish-shell/pull/8695#discussion_r801608359), I think it shouldn't.

Escapes are not handled in `tok_is_string_character()` either - for example, it wouldn't know if `c = ';'` is escaped - it assumes that it isn't. So I think it is reasonable to burden the caller with escaping. We can also document that, perhaps like this:
```c
/// \c pos must not be an escaped character.
```

krobelus: > Escapes are not handled in `tok_is_string_character()` either

Yeah, but that doesn't contradict that function's name.
Meanwhile `is_comment()` is a straight up lie if quoted/escaped, and I couldn't find a better name.

So I'd just merge as-is.

Issue #5048, opened 2018-06-14
Commas in process output are used as separators in brace expansion


mqudsi:
_I originally thought that [this bug](https://github.com/fish-shell/fish-shell/commit/35a7e1ef0eb9892c6ec73a7f3fee4d3f6ee9580b#r29369994) affected only fish 3.0 with [the changes to brace expansion](https://github.com/fish-shell/fish-shell/commit/d385248cc8533d7895a80d35e36fbcd1ef7c8f54) that imbued it with significantly more functionality and flexibility, but it turns out that even fish 2.7.1 is affected by this bug._

```fish
~> set -l tokens "foo,bar"
~> printf "'%s' " {$tokens}
'foo,bar'
~> printf "'%s' " {(echo $tokens)}
'foo' 'bar'
```

The output of a process should never be evaluated in this fashion, as this poses a major security risk.

faho:
>as this poses a major security risk.

I would actually dispute that characterization - the output is not "evaluated" as _code_. It's "parsed" as a certain form of expression.

You can't use this to `rm` anything or `cat` anything else. There's really no way for anyone malicious to abuse this.

So I'd merely call it a bug.

mqudsi:
It's not arbitrary code execution, but it it _is_ a vulnerability. Imagine privileged code that executes some command based off of usernames, perhaps something like a script that changes a user's password. It's pretty much along the lines of classical SQL injection vulnerabilities due to lack of argument escaping.

Let's say there's are sensitive resources located on a server that can only be accessed by the user and shared resources that can be accessed by anyone. Charlie wants to gain access to Alice's file and finds somewhere where the sysadmin unwittingly wrote the following code, assuming it to be safe:

```fish
for file in {(get-username), shared}.txt
    echo $file:
    cat $file
end
```

By creating an account with the name `bob,alice` in the system, Charlie gets access to `alice.txt` because `{(get-username), shared}.txt` now evaluates to `[bob.txt, alice.txt, shared.txt]` instead of `[bob,alice.txt shared.txt]`.

Security. Fun stuff.

faho:
>By creating an account with the name bob,alice in the system

If Charlie can create accounts, then why not create alice directly?

mqudsi:
> If Charlie can create accounts, then why not create alice directly?

Because `alice` is already taken by Alice.

anordal:
How is this not just yet another way to split a string into an array? If that's what the privileged code needs to do – fine!

[...]

PR #6059, opened 2019-08-24
Escape separators (colon and equals) to improve completion

krobelus:
Fish completes parts of words split by the separators, so things like
`dd if=/dev/sd<TAB>` work.
This commit improves interactive completion if completion strings legitimately
contain '=' or ':'.  Consider this example where completion will suggest
`a:a:1` and other files in the cwd in addition to `a:1`

```fish
touch a:1; complete -C'ls a:'
```

This behavior remains unchanged, but this commit allows to quote or escape
separators, so that e.g. `ls "a:<TAB>` and `ls a\:<TAB>` successfully complete
the filename.

This also makes the completion insert those escapes automatically unless
already quoted.
So `ls a<TAB>` will give `ls a\:1`.

Both changes match bash's behavior.

----

## TODOs:
Not sure if these are relevant:

- [ ] Changes to fish usage are reflected in user documentation/manpages.
- [ ] Tests have been added for regressions fixed
- [ ] User-visible changes noted in CHANGELOG.md

PR #8717, opened 2022-02-11
Escape commas when expanding command substitutions

shayaviv:
Fixes issue #5048.

## TODOs:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -95,6 +95,7 @@ Scripting improvements
 - ``argparse`` with ``--ignore-unknown`` no longer breaks with multiple unknown options in a short option group (:issue:`8637`).
 - Comments inside command substitutions or brackets now correctly ignore parentheses, quotes, and brackets (:issue:`7866`, :issue:`8022`).
 - ``complete -C`` supports a new ``--escape`` option, which turns on escaping in returned completion strings (:issue:`3469`).
+- Commas in process output are no longer used as separators in brace expansion (:issue:`5048`).
 
 Interactive improvements
 ------------------------
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -987,6 +987,7 @@ static void escape_string_script(const wchar_t *orig_in, size_t in_len, wcstring
                 case L']':
                 case L'{':
                 case L'}':
+                case L',':
                 case L'?':
                 case L'*':
                 case L'|':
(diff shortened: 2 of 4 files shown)


faho:
This breaks ordinary comma output, which is why you had to adjust the test.

Sorry, it's not a workable approach.

shayaviv:
@faho Would you mind giving a specific example for a command output (not an error message) that this change breaks?

@krobelus If it's not too much trouble, I'd like to get you opinion too before we trash this.

faho:
>@faho Would you mind giving a specific example for a command output (not an error message) that this change breaks?

... the test you had to adjust?

```
function t
    test $argv[1] -eq 5
end
t 5,2
```

This function now prints `in function 't' with arguments '5\,2'`. The comma is now escaped. It shouldn't be.

This is simply the wrong place to handle this.

krobelus:
On Fri, Feb 11, 2022 at 06:32:04AM -0800, Shay Aviv wrote:
> 2. Commas need to be escaped, because they *are* special.

I agree that commas are special sometimes but the conclusion "commands need
to be escaped" is wrong. The output of a command subtitution should not be
escaped. It should be exactly what the command would print.

> You can test this without my changes:
>    ```console
>    ~> echo {1,2,3}
>    1 2 3
>    ~> echo {1\,2,3}
>    1,2 3
>    ```

You are fixing one case but breaking others that don't want escaping, like

	echo (string join , 1 2 3)

The real fix is probably somewhere in src/expand.cpp. Doesn't seem straightforward.

faho:
>The string after with arguments is escaped by design. You can test this without my changes: 

Of course, yes. However now it is escaped *differently*, and for a character that doesn't need to be escaped here.

This makes escaping uglier than it needs to be and changes e.g. what `string escape` and completions do!

Do `string escape --no-quoted 5,2`. It now prints `5\,2`.

Do

```fish
complete -c foo -xa '5,2 5,3'
```

and complete. It now completes as `5\,2` and `5\,3`.

This applies to *all* completions, so even filenames with `,` are used as `5\,2` in the commandline.

This doesn't technically "break", but it makes all of this a little more ugly.

>Commas need to be escaped, because they are special. You can test this without my changes: 

They aren't special in a way that they need to be escaped everywhere. They are only special inside braces.

----

This simply is not the right solution, not even in concept.

This isn't a personal slight against you, sometimes you try the wrong thing. You could fill books with all the wrong approaches I've tried over the years. It happens.

But it shouldn't be merged. Sorry!

shayaviv:
I'm not trying to convince you my solution is good - I just wanted to clear things up. Thanks for your time and consideration 🙂

krobelus:
Oh I see, this is a bit like https://github.com/fish-shell/fish-shell/pull/6059, which we had to revert because we (still) can't tell whether the characters in question need escaping. Adding backslashes to `,` might not make a difference, but it makes things quite ugly.

PR #8734, opened 2022-02-16
Added kubectl completion

exploide:
## Description

This PR adds completions for the Kubernetes control tool `kubectl`.

It can generate its own completions, so all we need to do is sourcing them.

While the completions are not perfect yet, they work at least for basic use cases and complete global options and subcommands.

## TODOs:


krobelus:
good idea, thanks!

PR #8738, opened 2022-02-18
completions/sv.fish: update completion for sv, a part of runit.

lelgenio:
This should fix finding the correct active svdir in most cases.

Search for services in the following order:

$SVDIR                      - User defined override
/run/runit/runsvdir/current - Value patched in by Void[1]
/run/runit/service          - Value patched in by Artix[2]
/etc/service                - Value patched in by Debian[3]
/services                   - Default value for runit[4]

Also don't use /etc for searching since it is not guaranteed that there
is an instance of runsvdir running in that directory.

Finally return quietly if there is no svdir.

[1] https://github.com/void-linux/void-packages/blob/75403cef76a93199456befdcb8884fbac39adbb3/srcpkgs/runit/template#L29
[2] https://gitea.artixlinux.org/packagesR/runit/src/commit/c9d691ce864e5be6735025fc51c6d72fdb5aabb0/x86_64/core/PKGBUILD#L9
[3] https://sources.debian.org/src/runit/2.1.2-41/debian/patches/0001-default-directory-for-services-on-Debian-is-etc-servi.diff/
[4] hard-coded in sv.c

### DIFF
--- a/share/completions/sv.fish
+++ b/share/completions/sv.fish
@@ -10,15 +10,24 @@ set -l commands \
     shutdown force-stop force-reload force-restart force-shutdown \
     try-restart check
 
-
 function __fish_complete_sv_list_services
-    for dir in $SVDIR /run/runit/service/ /etc/runit/current /etc/runit/runsvdir/
-        test -d $dir
-        and break
+    set -l svdir
+    for candidate_svdir in \
+        "$SVDIR" \
+        /run/runit/runsvdir/current \
+        /run/runit/service \
+        /etc/services \
+        /services
+        if test -d "$candidate_svdir"
+            set svdir "$candidate_svdir"
+            break
+        end
     end
-    set -l services (string match -r '[^/]*$' $dir/*)
-    set -l out (sv status $services 2>/dev/null)
-    and string replace -r "^(\w+: )(.*?):" '$2\t$1$2:' $out
+    set -q svdir[1]; or return
+    set -l services (command ls "$svdir")
+    set -l sv_status (sv status $services 2>/dev/null |
+                      string replace -ar ';.*$' '')
+    and string replace -r "^(\w+: )(.*?):" '$2\t$1' $sv_status
     or printf "%s\n" $services
 end
 


share/completions/sv.fish
```diff
@@ -10,12 +10,19 @@ set -l commands \
...
+    /services \
+    #
 
 function __fish_complete_sv_list_services
-    for dir in $SVDIR /run/runit/service/ /etc/runit/current /etc/runit/runsvdir/
+    for dir in $SVDIR $search_svdirs
         test -d $dir
         and break
     end
+    test -n $dir; or return
```
faho, line 25: Since $SVDIR is a user-defined variable, this needs to be quoted - `set -l SVDIR; test -n $SVDIR` *succeeds*.

Same for the above `test -d $dir`.

(another use case for #8265 - this entire thing would be `set -l dir (path filter -d -- $SVDIR $search_svdirs)[1]; or return`)

faho (changes requested):
Variables used with `test` must be quoted

share/completions/sv.fish
```diff
@@ -10,12 +10,19 @@ set -l commands \
     shutdown force-stop force-reload force-restart force-shutdown \
     try-restart check
 
+set -l search_svdirs \
+    /run/runit/runsvdir/current \
+    /run/runit/service \
+    /etc/services \
+    /services \
+    #
```
krobelus, line 18: > Finally return quietly if there is no svdir.

I don't think this actually works.
Note that the # doesn't do anything. You probably wanted to write "" here.

this alternative approach would also work 

```fish
set -l svdir
for dir in \
    "$SVDIR" \
    /run/runit/runsvdir/current \
    /run/runit/service \
    /etc/services \
    /services
    if test -d $dir
        set svdir $dir
        break
    end
end
set -q svdir[1]; or return
```

Note that we actually only need to quote $SVDIR (for edge cases), so doing that explicitly seems better.

lelgenio: > Note that the # doesn't do anything. You probably wanted to write "" here.

It was on purpose, to make sure all lines on the list ended with `\`, but I realize it's stupid.

> ```fish
> test -d $dir
> ```
> Note that we actually only need to quote $SVDIR (for edge cases), so doing that explicitly seems better.

We also need to quote variables in `test`, see faho's comment, I had no idea about this quirk with `test` until now.

lelgenio:
I took the liberty to of reverting from `string match -r '[^/]*$' $dir/*` to `command ls "$svdir"`, which is slower but does not give nasty errors when the directory is empty.

I also removed the logger status from the completion description, I am not sure this type of indentation is correct, `fish_indent` does not seem to care.

```fish
set -l sv_status (sv status $services 2>/dev/null |
                  string replace -ar ';.*$' '')
```

krobelus:
> We also need to quote variables in `test`

Actually we don't, and we don't even need to quote $SVDIR either (my bad).

This pattern works fine:

```fish
for dir in $SVDIR /etc/services
  test -d $dir; ...
end
```

If SVDIR is undefined or an empty list (`set SVDIR`), the for loop will just skip over it.
OTOH if SVDIR is not empty, let's say `set SVDIR foo bar`, then we'd run `test -d foo` and `test -d bar`.
for-loop variables never need to  be quoted because they always have exactly one element.

If we use `test -n $svdir` and `svdir` can be empty, then you need to quote, of course.

krobelus:
> I am not sure this type of indentation is correct, `fish_indent` does not seem to care.
> 	```fish
> 	set -l sv_status (sv status $services 2>/dev/null |
> 	                  string replace -ar ';.*$' '')
	```

sounds like a missing feature in fish_indent. It will probably indent the continuation line by one extra level, same as outside command substitutions.
Anyway, your version looks nice, why don't we keep it. Of course this can also fit on one line (we use up to 100 chars).

share/completions/sv.fish
```diff
@@ -10,15 +10,24 @@ set -l commands \
...
+        if test -d "$candidate_svdir"
+            set svdir "$candidate_svdir"
+            break
+        end
     end
-    set -l services (string match -r '[^/]*$' $dir/*)
-    set -l out (sv status $services 2>/dev/null)
-    and string replace -r "^(\w+: )(.*?):" '$2\t$1$2:' $out
+    set -q svdir[1]; or return
+    set -l services (command ls "$svdir")
```
krobelus, line 27: great catch fixing this for empty directories.
I think the idiomatic solution is to rely on nullglob behavior of set/for/count

```diff
-set -l services (string match -r '[^/]*$' $dir/*)
+set -l services $dir/*
+set services (string match -r '[^/]*$' $services)
```

though parsing ls output is probably fine too

krobelus:
merged as 02c34a30eb2abbd9a2fcacd0763a6d4ea69ade47, thanks!

PR #8748, opened 2022-02-27
Better documentation of forward-char

sourcefrog:
This makes it match the code in reader.cpp, and explains why the default
binding of `right` accepts the complete line.

## Description

Motivation: I wanted to bind `right` to have a less drastic effect than
accepting the whole line, and I was confused by the docs apparently saying that
all the default binding does is move the cursor.

## TODOs:

  - Does not change the code
  - Does not change the code
  - Seems too minor to mention

### DIFF
--- a/doc_src/cmds/bind.rst
+++ b/doc_src/cmds/bind.rst
@@ -89,10 +89,11 @@ The following special input functions are available:
     only execute the next function if the previous succeeded (note: only some functions report success)
 
 ``accept-autosuggestion``
-    accept the current autosuggestion completely
+    accept the current autosuggestion
 
 ``backward-char``
-    moves one character to the left
+    move one character to the left.
+    If the completion pager is active, select the previous completion instead.
 
 ``backward-bigword``
     move one whitespace-delimited word to the left
@@ -179,13 +180,15 @@ The following special input functions are available:
     move one whitespace-delimited word to the right
 
 ``forward-char``
-    move one character to the right
+    move one character to the right; or if at the end of the commandline, accept the current autosuggestion.
+    If the completion pager is active, select the next completion instead.
 
 ``forward-single-char``
-    move one character to the right; if an autosuggestion is available, only take a single char from it
+    move one character to the right; or if at the end of the commandline, accept a single char from the current autosuggestion.
 
 ``forward-word``
-    move one word to the right
+    move one word to the right; or if at the end of the commandline, accept one word
+    from the current autosuggestion.
 
 ``history-search-backward``
     search the history for the previous match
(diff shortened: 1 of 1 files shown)


doc_src/cmds/bind.rst
```diff
@@ -179,7 +179,9 @@ The following special input functions are available:
     move one whitespace-delimited word to the right
 
 ``forward-char``
-    move one character to the right
+    move one character to the right; or in a menu select the item to the right;
+    or if it at the end of the input and an autosuggestion is available accept
+    it completely
```
krobelus, line 184: Thanks, this looks nice.

I'm not a fan of the "completely" wording though.
I think our only menu is the completion pager, but we rarely use the word menu (only twice in doc_src/interactive.rst) which is an argument for using "(completion) pager".
While at it, we can also document the same behavior in backward-char

so how about something like

```
diff --git a/doc_src/cmds/bind.rst b/doc_src/cmds/bind.rst
index 9de9ed5ca..b6f45f429 100644
--- a/doc_src/cmds/bind.rst
+++ b/doc_src/cmds/bind.rst
@@ -91,6 +91,7 @@ The following special input functions are available:
 ``accept-autosuggestion``
-    accept the current autosuggestion completely
+    accept the current autosuggestion
 
 ``backward-char``
-    moves one character to the left
+    move one character to the left.
+    If the completion pager is active, select the previous completion instead.
 
@@ -181,6 +182,7 @@ The following special input functions are available:
 ``forward-char``
-    move one character to the right
+    move one character to the right; or if at the end of the commandline, accept the current autosuggestion.
+    If the completion pager is active, select the next completion instead.
 
 ``forward-single-char``
-    move one character to the right; if an autosuggestion is available, only take a single char from it
+    move one character to the right; or if at the end of the commandline, accept a single char from the current autosuggestion.
 
```

sourcefrog: OK, will do - unless you prefer to just merge that yourself?

> -    move one character to the right; if an autosuggestion is available, only take a single char from it

I'd drop "only" from that clause too.

sourcefrog:
I guess the test failure is just a flake; it seems irrelevant.

doc_src/cmds/bind.rst
krobelus: > OK, will do

that'd be nice, it's always helpful to double check these

> I'd drop "only" from that clause too. 

sounds good

doc_src/cmds/bind.rst
```diff
@@ -230,7 +233,8 @@ The following special input functions are available:
     move the next word to the killring
 
 ``nextd-or-forward-word``
-    if the commandline is empty, then move forward in the directory history, otherwise move one word to the right
+    if the commandline is empty, then move forward in the directory history, otherwise move one word to the right.
+    If at the end of the commandline, accept the current autosuggestion.
```
krobelus, line 237: good catch documenting this as well

"accept the current autosuggestion" is not quite right, it's just one word (same as forward-word)
fixed and merged as c0be74c55a6a0a4d7161e1a82ff804ccdf4443d2, thanks!

Issue #6884, opened 2020-04-10
Add tab completion for the iwctl command used to interact with IWD


ericonr:
[IWD](https://iwd.wiki.kernel.org/start) is a "new" wireless daemon that intends to supersede wpa_supplicant,  by being both faster and having simpler configuration. It has a pretty simple CLI, called iwctl, which has a readline interface, but can also take commands directly. It would be great to be able to use those commands with tab completion inside fish.

I am interested in implementing those completions, just want to be sure it's ok and leave this issue open if anyone wants to comment anything.

faho:
Sure, go for it!

mizzunet:
@ericonr Do you have it?

ModProg:
working on this, will revive the pr in the comming days.

ModProg:
Done: https://github.com/fish-shell/fish-shell/pull/9932

Issue #8217, opened 2021-08-13
Command to remove invisible escapes


IlanCosman:
After https://github.com/fish-shell/fish-shell/pull/8182, `string pad` and `string length` can both deal with invisible escapes.

There should be a new command, or perhaps a flag to `string unescape`, that prints the arguments without invisible escapes.

The use case is redirecting functions or commands that use colors. e.g.

- Testing functions with colors, bold etc.

- Redirecting `yay` when it uses colors. Below, `yay` removes it's colors, but fails to remove it's bold and normal escape codes.

```
:: Synchronizing package databases...
 core downloading...
 extra downloading...
 community downloading...
 multilib downloading...
 endeavouros downloading...
:: Starting full system upgrade...
 there is nothing to do
:: [1mSearching databases for updates...[0m
:: [1mSearching AUR for updates...[0m
 -> Missing AUR Packages:  libopenaptx
 there is nothing to do
```

This is a more general solution, so perhaps this new function could also be used in the implementation of `string pad` and `string length --visible`?

faho:
Yeah, sure.

>Redirecting yay when it uses colors. Below, yay removes it's colors, but fails to remove it's bold and normal escape codes.

Are you redirecting stderr without stdout, perchance? From a cursory reading of the yay code, it [does check](https://github.com/Jguer/yay/blob/3de02dbd73e09a0bca7ba30b28b66334c6522b9e/main.go#L72) if std*out* is a terminal (and if you had set pacman to color=always, those lines would also be colored).

Really, typically this should be handled by the tool itself, we can't always get all escape sequences correctly.

IlanCosman:
> Yeah, sure.

❤️

> Are you redirecting stderr without stdout

Nope, just stdout: `yay > foo`

> Really, typically this should be handled by the tool itself, we can't always get all escape sequences correctly.

True. Still applies for Fish functions though (do I really want to add `status is-interactive` everywhere 😄)

faho:
>do I really want to add status is-interactive everywhere

You want `isatty stdout`. We might want to add a `--auto` mode to `set_color`, that only prints the color if stdout is a terminal or if stdout is a command substitution but the *next* stdout in the chain is one, so you can do

```fish
function foo
    # This prints nothing if foo is redirected
    set_color --auto ff2244
    # Neither does the `set_color` here
    echo "This is my text$(set_color --auto normal)"
end
```

mqudsi:
[...]

PR #8756, opened 2022-03-05
Add completion for `iwctl`

mizzunet:
## Description
This is not 100% percent finished.
 It can't properly return completion for `iwctl station wlan0 connect<TAB> ` which would return `\e\[0mWIFI1`. Here `\e\[0m` is due to some colour or something, I'm not what that is. I couldn't get rid of if as well.

Fixes issue #
https://github.com/fish-shell/fish-shell/issues/6884
## TODOs:

### DIFF
--- a/share/completions/iwctl.fish
+++ b/share/completions/iwctl.fish
@@ -1,0 +1,59 @@
+set adhoc '(iwctl ad-hoc list | tail -n -2 | awk \'{print $1}\')'
+set ap '(iwctl ap list | tail -n -2 | awk \'{print $1}\')'
+set adpaters '(iwctl adpater list | tail -n -2 | awk \'{print $1}\')'
+set devices '(iwctl device list | tail -n -2 | awk \'{print $1}\')'
+set dpp '(iwctl dpp list | tail -n -2 | awk \'{print $1}\')'
+set knownnetworks '(iwctl known-networks list | tail -n -2 | awk \'{print $1}\')'
+set stations '(iwctl station list | tail -n -2 | awk \'{print $1}\')'
+set wsc '(iwctl wsc list | tail -n -2 | awk \'{print $1}\')'
+set net '(iwctl station wlan0 get-networks | tail -n -2 | awk \'{print $2}\')'
+
+complete -f iwctl
+
+# First argument
+complete -c iwctl -n "test (__fish_number_of_cmd_args_wo_opts) = 1" \
+   -a "ad-hoc adapter ap device dpp help known-networks station version wsc"
+
+# Second argument
+complete -c iwctl -n "test (__fish_number_of_cmd_args_wo_opts) = 2; and __fish_seen_subcommand_from adapter" \
+    -xa "list $adapters"
+complete -c iwctl -n "test (__fish_number_of_cmd_args_wo_opts) = 2; and __fish_seen_subcommand_from ap" \
+    -xa "list $ap"
+complete -c iwctl -n "test (__fish_number_of_cmd_args_wo_opts) = 2; and __fish_seen_subcommand_from dpp" \
+    -xa "list $dpp"
+complete -c iwctl -n "test (__fish_number_of_cmd_args_wo_opts) = 2; and __fish_seen_subcommand_from ad-hoc" \
+    -xa "list $adhoc"
+complete -c iwctl -n "test (__fish_number_of_cmd_args_wo_opts) = 2; and __fish_seen_subcommand_from device" \
+    -xa "list $devices"
+complete -c iwctl -n "test (__fish_number_of_cmd_args_wo_opts) = 2; and __fish_seen_subcommand_from known-networks" \
+    -xa "list $knownnetworks"
+complete -c iwctl -n "test (__fish_number_of_cmd_args_wo_opts) = 2; and __fish_seen_subcommand_from station" \
+    -xa "list $stations"
+complete -c iwctl -n "test (__fish_number_of_cmd_args_wo_opts) = 2; and __fish_seen_subcommand_from wsc" \
+    -xa "list $wsc"
+
+# Third arguments
+complete -c iwctl -n "test (__fish_number_of_cmd_args_wo_opts) = 3; and __fish_seen_subcommand_from device" \
+    -xa "set-property show"
+complete -c iwctl -n "test (__fish_number_of_cmd_args_wo_opts) = 3; and __fish_seen_subcommand_from dpp" \
+    -xa "start-configurator start-enrollee stop"
+complete -c iwctl -n "test (__fish_number_of_cmd_args_wo_opts) = 3; and __fish_seen_subcommand_from known-networks" \
+    -xa "forget set-property show"
+complete -c iwctl -n "test (__fish_number_of_cmd_args_wo_opts) = 3; and __fish_seen_subcommand_from wsc" \
+    -xa "cancel push-button start-pin start-user-pin"
+
+# Forth arguement
+complete -c iwctl -n "test (__fish_number_of_cmd_args_wo_opts) = 4; and __fish_seen_subcommand_from device set-property" \
+    -xa  "Powered Mode"
+complete -c iwctl -n "test (__fish_number_of_cmd_args_wo_opts) = 4; and __fish_seen_subcommand_from known-networks set-property" \
+    -xa  "AutoConnect"
+complete -c iwctl -n "test (__fish_number_of_cmd_args_wo_opts) = 4; and __fish_seen_subcommand_from station connect" \
+    -xa  '(iwctl station wlan0 get-networks | tail -n -2 | awk \'{print $2}\')'
+
+# Fifth arguement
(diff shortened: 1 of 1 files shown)


share/completions/iwctl.fish
```diff
@@ -0,0 +1,59 @@
...
+    -xa  "Powered Mode"
+complete -c iwctl -n "test (__fish_number_of_cmd_args_wo_opts) = 4; and __fish_seen_subcommand_from known-networks set-property" \
+    -xa  "AutoConnect"
+complete -c iwctl -n "test (__fish_number_of_cmd_args_wo_opts) = 4; and __fish_seen_subcommand_from station connect" \
+    -xa  '(iwctl station wlan0 get-networks | tail -n -2 | awk \'{print $2}\')'
+
+# Fifth arguement
+complete -c iwctl -n "test (__fish_number_of_cmd_args_wo_opts) = 5; and __fish_seen_subcommand_from device set-property Mode" \
+    -xa  "ad-hoc ap station"
+complete -c iwctl -n "test (__fish_number_of_cmd_args_wo_opts) = 5; and __fish_seen_subcommand_from device set-property Powered" \
```
faho, line 56: `__fish_seen_subcommand_from` checks if *at least one* of the given arguments has been seen. That means this condition will succeed with e.g.

```fish
iwctl one two three four set-property
```

I don't think we have something exactly like you want here ready-made. Try:

```fish
function __fish_seen_all_subcommands
    set -q argv[1]
    or return 1
    set -l cmd (commandline -poc)
    set -e cmd[1]
    for i in $cmd
        if contains -- $i $argv[1]
            # This argument is the next subcommand, pop it
            set -e argv[1]
        end
    end
    # If we still have an argument left over, we don't have all subcommands
    not set -q argv[1]
end
```

and then use it like


```fish
test (__fish_number_of_cmd_args_wo_opts) = 5; and __fish_seen_all_subcommands device set-property Powered
```

(note: This can still have false-positives. That is very very hard to avoid - without knowing all option-taking options it's impossible. For now, and this specific command that'll be good enough)

share/completions/iwctl.fish
```diff
@@ -0,0 +1,59 @@
+set adhoc '(iwctl ad-hoc list | tail -n -2 | awk \'{print $1}\')'
+set ap '(iwctl ap list | tail -n -2 | awk \'{print $1}\')'
+set adpaters '(iwctl adpater list | tail -n -2 | awk \'{print $1}\')'
+set devices '(iwctl device list | tail -n -2 | awk \'{print $1}\')'
+set dpp '(iwctl dpp list | tail -n -2 | awk \'{print $1}\')'
+set knownnetworks '(iwctl known-networks list | tail -n -2 | awk \'{print $1}\')'
+set stations '(iwctl station list | tail -n -2 | awk \'{print $1}\')'
+set wsc '(iwctl wsc list | tail -n -2 | awk \'{print $1}\')'
+set net '(iwctl station wlan0 get-networks | tail -n -2 | awk \'{print $2}\')'
```
faho, line 9: Is this unused?

share/completions/iwctl.fish
```diff
@@ -0,0 +1,59 @@
...
+complete -c iwctl -n "test (__fish_number_of_cmd_args_wo_opts) = 3; and __fish_seen_subcommand_from wsc" \
+    -xa "cancel push-button start-pin start-user-pin"
+
+# Forth arguement
+complete -c iwctl -n "test (__fish_number_of_cmd_args_wo_opts) = 4; and __fish_seen_subcommand_from device set-property" \
+    -xa  "Powered Mode"
+complete -c iwctl -n "test (__fish_number_of_cmd_args_wo_opts) = 4; and __fish_seen_subcommand_from known-networks set-property" \
+    -xa  "AutoConnect"
+complete -c iwctl -n "test (__fish_number_of_cmd_args_wo_opts) = 4; and __fish_seen_subcommand_from station connect" \
+    -xa  '(iwctl station wlan0 get-networks | tail -n -2 | awk \'{print $2}\')'
```
faho, line 51: This prints "psk" for me. Is that what you want?

Is the output here maybe not stable?

Here's what it looks like for me (with network names replaced):

```
                               Available networks                              
--------------------------------------------------------------------------------
    Network name                    Security          Signal
--------------------------------------------------------------------------------
  > currentnetwork                  psk               ****  
    pskdnetwork                     psk               ****
    opennetwork                     open              ****
```

faho (changes requested):

share/completions/iwctl.fish
mizzunet: For me, it's `print 3` that prints `psk`.

> Is the output here maybe not stable?

I guess

faho:
>It can't properly return completion for iwctl station wlan0 connect<TAB> which would return \e\[0mWIFI1. Here \e\[0m is due to some colour or something, I'm not what that is. I couldn't get rid of if as well.

Yeah, iwctl is hardcoding color sequences here, and not even checking if maybe they aren't connected to a terminal, or checking terminfo, or even giving an option to disable color.

That's not good, tbh.

Luckily they're only hardcoding [a few simple color sequences](https://git.kernel.org/pub/scm/network/wireless/iwd.git/tree/client/display.h#n26) (and a clear screen sequence that's irrelevant here), so we can remove them with

```fish
string replace -ra '\e\[[\d;]+m' ''
```

e.g. 

```fish
iwctl station wlan0 get-networks | tail -n -2 | string replace -ra '\e\[[\d;]+m' ''
```

(also see #8217)

share/completions/iwctl.fish
mizzunet: I don't get it, seems little complicated to me. 
Can I skip this part? Maybe someone else could update this part.

share/completions/iwctl.fish
faho: Ooof.

>For me, it's print 3 that prints psk.

Does your output look different than what I showed you above, e.g. are the columns swapped, or is this a case of whitespace being different (which we could probably work around).

Because if the columns are swapped, that's terrible news, because it makes this output unusable, which kind of makes the entire script unmergeable.

mizzunet: I guess, this is because I only have one network

This is my output
```
                               Available networks
--------------------------------------------------------------------------------
    Network name                    Security          Signal
--------------------------------------------------------------------------------
  > FTTH-6DBD                       psk               ****
```

faho: Aaaaah, okay, I see what it is:

The selected network line starts with `  > `, while unselected networks start with `    `. awk will collapse the latter because it's whitespace. In the former the `>` is the first field!

So: `awk '{print $2}'` is too simple here. You probably want a regular expression.

E.g.

```fish
string replace -rf '^[\s>]*(\S+)\s*.*' '$1'
```

And the `>` is even colored, so you have that `\e[SOMENUMBERm` to deal with.

The full line would be something like

```fish
iwctl station wlan0 get-networks | tail -n +5 | string replace -ra '\e\[[\d;]+m' '' | string replace -rf '^[\s>]*(\S+)\s*.*' '$1'
```

(note the `tail -n +5`. For me that's necessary because the first four lines are header guff)

share/completions/iwctl.fish
mizzunet: Fixed

share/completions/iwctl.fish
```diff
@@ -6,7 +6,7 @@ set dpp '(iwctl dpp list | tail -n -2 | awk \'{print $1}\')'
 set knownnetworks '(iwctl known-networks list | tail -n -2 | awk \'{print $1}\')'
 set stations '(iwctl station list | tail -n -2 | awk \'{print $1}\')'
 set wsc '(iwctl wsc list | tail -n -2 | awk \'{print $1}\')'
```
faho, line 8: These others also need to have the `tail` adjusted. `tail -n -2` will give you the last two lines, which only works if you have one entry. We want all of them, we just want to skip the header. That's `tail -n +5`, to start at the fifth line of output.

I took a quick glance, as far as I can see they don't mark a "current" or "active" entry so you can probably skip the final `replace`.

You might have to strip the colors tho.

share/completions/iwctl.fish
```diff
@@ -0,0 +1,61 @@
+set adhoc '(iwctl ad-hoc list | tail -n -2 | awk \'{print $1}\')'
+set ap '(iwctl ap list | tail -n -2 | awk \'{print $1}\')'
+set adpaters '(iwctl adpater list | tail -n -2 | awk \'{print $1}\')'
```
faho, line 3: Typo: Should be `adapter`

share/completions/iwctl.fish
```diff
@@ -0,0 +1,61 @@
+set adhoc '(iwctl ad-hoc list | tail -n -2 | awk \'{print $1}\')'
+set ap '(iwctl ap list | tail -n -2 | awk \'{print $1}\')'
```
faho, line 2: This prints "No" for me. Presumably it should be empty instead.

Unfortunately `iwctl` provides no useful exit code (it still exits 0 even if no devices in ap mode are there), so you might have to simply ignore "No"

faho:
Okay, this has unaddressed feedback so I'll be closing it.

If you or anyone else gets back to it we can reopen!

PR #8757, opened 2022-03-05
Add completion for `rc-status`

mizzunet:
`rc-status` is part of OpenRC

## Description

Talk about your changes here.

Fixes issue #

## TODOs:

### DIFF
--- a/share/completions/rc-status.fish
+++ b/share/completions/rc-status.fish
@@ -1,0 +1,17 @@
+complete -f rc-status
+complete -c rc-status -n "test (__fish_number_of_cmd_args_wo_opts) = 1" \
+    -xa "(/bin/ls -1 /etc/runlevels)"
+
+complete -c rc-status -s h -l help -d 'Display the help output'
+complete -c rc-status -s a -l all -d 'Show services from all run levels'
+complete -c rc-status -s f -l format -d 'format status to be parsable'
+complete -c rc-status -s l -l list -d 'Show list of run levels'
+complete -c rc-status -s r -l runlevel -d 'Show the name of the current runlevel'
+complete -c rc-status -s m -l manual -d 'Show manually started services'
+complete -c rc-status -s s -l servicelist
+complete -c rc-status -s S -l supervised
+complete -c rc-status -s u -l unused
+complete -c rc-status -s v -l version
+complete -c rc-status -s q -l quiet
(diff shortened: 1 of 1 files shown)


share/completions/rc-status.fish
```diff
@@ -0,0 +1,17 @@
...
+    -xa "(/bin/ls -1 /etc/runlevels)"
+
+complete -c rc-status -s h -l help -d 'Display the help output'
+complete -c rc-status -s a -l all -d 'Show services from all run levels'
+complete -c rc-status -s f -l format -d 'format status to be parsable'
+complete -c rc-status -s l -l list -d 'Show list of run levels'
+complete -c rc-status -s r -l runlevel -d 'Show the name of the current runlevel'
+complete -c rc-status -s m -l manual -d 'Show manually started services'
+complete -c rc-status -s s -l servicelist
+complete -c rc-status -s S -l supervised
```
faho, line 12: What is "supervised"?

The descriptions are typically the meat of the completions because they help you use an unfamiliar tool.

share/completions/rc-status.fish
```diff
@@ -0,0 +1,17 @@
+complete -f rc-status
+complete -c rc-status -n "test (__fish_number_of_cmd_args_wo_opts) = 1" \
+    -xa "(/bin/ls -1 /etc/runlevels)"
```
faho, line 3: I don't love that we use `ls` here, or that we hard-code it as being in /bin.

Unfortunately the current builtin-only way is:

```fish
set -l levels /etc/runlevels/*
string replace /etc/runlevels/ $levels
```

with #8265 this would become:

```fish
path basename /etc/runlevels/*
```

Anyway, for now the ls is probably fine as it is.

share/completions/rc-status.fish
```diff
@@ -1,7 +1,10 @@
+set -l levels_path /etc/runlevels/*
```
faho, line 1: Please don't do this *here*.

That means it will be done when the completions are *loaded*.

Do it all inside the argument string:

```fish
set -l levels '(set -l p /etc/runlevels/*; string replace /etc/runlevels "" $p)'
```

(and yes, the `set` is necessary because `string` will error out if the glob doesn't match)

faho:
Merged, thanks!

PR #8758, opened 2022-03-05
Add completion for `brightnessctl`

mizzunet:
## Description

Talk about your changes here.

Fixes issue #

## TODOs:


share/completions/brightnessctl.fish
```diff
@@ -0,0 +1,3 @@
+complete -f brightnessctl
+complete -k -c brightnessctl -n "test (__fish_number_of_cmd_args_wo_opts) = 1" \
```
faho, line 2: Please remove the `-k` (short for `--keep-order`) unless you have a good reason to keep it.

The default ordering should be simple, predictable and therefore alphabetical.

faho (changes requested):

share/completions/brightnessctl.fish
mizzunet: Done

PR #8762, opened 2022-03-06
Formatted Files According to the clang-format

Harry-Hopkinson:
## Description

Formatted Files According to the clang-format to incorporate more consistent linting for C++ Code.


faho:
Thanks, but we're trying to prepare a new release, that's a bit of a bad time to do auto-formatting, because it can easily break rebases and reverts.

We'll apply the changes *after*.

Harry-Hopkinson:
Okay - thanks for your response.

Issue #8351, opened 2021-10-14
Suggestion: -n flag for string join mirroring string split


schuelermine:
There is a flag `-n`/`--no-empty` for `string split` that excludes empty result items.
There should be an analogous flag for `string join` that does not add the delimiter next to an empty item.

```
~$ string join -n + a b "" c
   string join + a b "" c
a+b+c
a+b++c
~$ read x; for o in {-n,}
     string join $o -- - a $x (string sub -s 3 $x) b
   end
read> iy
a-iy--b
a-iy-b
```

faho:
Yeah, makes sense.

PR #8774, opened 2022-03-09
New -n flag for string join.

joao-vitor-sr:
This flag jumps the delimiter for empty items.

### DIFF
--- a/src/builtins/string.cpp
+++ b/src/builtins/string.cpp
@@ -736,7 +736,7 @@ static int string_join_maybe0(parser_t &parser, io_streams_t &streams, int argc,
     arg_iterator_t aiter(argv, optind, streams);
     while (const wcstring *arg = aiter.nextstr()) {
         if (!opts.quiet) {
-            if (opts.no_empty && arg->empty()) continue;
+            if (opts.no_empty && arg->length() == 0) continue;
 
             if (nargs > 0) {
                 streams.out.append(sep);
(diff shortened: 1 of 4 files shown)


faho:
Fixes #8351

src/builtins/string.cpp
```diff
@@ -735,6 +736,8 @@ static int string_join_maybe0(parser_t &parser, io_streams_t &streams, int argc,
     arg_iterator_t aiter(argv, optind, streams);
     while (const wcstring *arg = aiter.nextstr()) {
         if (!opts.quiet) {
+            if (opts.no_empty && arg->length() == 0) continue;
```
faho, line 739: `arg->empty()` or clang-tidy will yell at us (yes that's kind of useless churn, sorry)

faho:
Seems okay to me, I'll wait with merging until after 3.4.0 is out (which we're hoping to get done soon).

src/builtins/string.cpp
faho: Have a look at [what actually failed](https://github.com/fish-shell/fish-shell/runs/5501523442?check_suite_focus=true) - the test was flaky, it had nothing to do with this code. It didn't even execute it.

Common CI systems are typically very very resource-constrained and we have a few tests that involve some form of timing, so they do, unfortunately, occasionally fail.

There's no reason to revert this.

faho:
Merged, thanks!

PR #8792, opened 2022-03-15
documentation, variable name definition

qodo-io:
Nitpickery - technically the sentences were incorrect, noticed while writing a regex to verify fish_vars.
Maybe it's useless for most people ... 

## Description

I originally just mentally translated the English sentence to a regex to check for valid variable names (e.g. `9_red_ballons_9`), and later noticed my tests failed, as it would allow me to `set 99 100` - which would break set theory, and people will get upset. :-) 

PS my first PR, hope I did not miss anything
- [x] read guidelines
- [x] read code of conduct

### DIFF
--- a/doc_src/language.rst
+++ b/doc_src/language.rst
@@ -873,7 +873,7 @@ Shell variables
 
 Variables are a way to save data and pass it around. They can be used just by the shell, or they can be ":ref:`exported <variables-export>`", so that a copy of the variable is available to any external command the shell starts. An exported variable is referred to as an "environment variable".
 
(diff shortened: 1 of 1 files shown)


faho:
Yup, this is wrong.

`set 99 100` *works*, and is unambiguous, because to use the variable value you need `$99`.

E.g.

```fish
set 99 100
echo 99 # prints 99
echo $99 # prints 100
math $99 + 99 # prints 199
```

Closing as it's incorrect.

qodo-io:
ok. sorry. I see the error and feel stupid now :-/ Damn intuition. Thanks for clarifying this and sorry for the wrong bug report!

Issue #8787, opened 2022-03-14
Documented fallback behavior for `fish_pager_color_selected_background` is now misleading


jaminthorns:
# Issue

With the recent release of `fish` version `3.4`, the `fish_pager_color_selected_background` variable is now set as a universal variable along with the other pager colors upon interactive initialization:

https://github.com/fish-shell/fish-shell/blob/cd62771d1245b5a49d02bf51ce193c3e96f898d3/share/functions/__fish_config_interactive.fish#L62-L67

It think this is a perfectly reasonable default, but in the [documentation for interactive use](https://fishshell.com/docs/current/interactive.html#pager-color-variables), this is stated:

> When the secondary or selected variables aren’t set, the normal variables are used, except for `$fish_pager_color_selected_background`, where the background of `$fish_color_search_match` is tried first.

This is still technically true, but because of the new initialization behavior for `fish_pager_color_selected_background`, this doesn't happen in practice, so the documentation is now misleading. To get the documented behavior, I can of course just run:

```fish
set -e fish_pager_color_selected_background
```

But the default behavior is still confusing.

# Context

I'm a maintainer of the [Dracula theme for `fish`](https://github.com/dracula/fish), and in this theme, we rely on the documented behavior for `fish_pager_color_selected_background` (falling back to the background color of `fish_color_search_match`). Upon upgrading to `fish` version `3.4`, I was confused as to why the background color of selected autocompletions had changed. Users of this theme are probably having a similar experience.

# Suggested Solutions

- Change the initialized value of `fish_color_search_match` to `-r` and remove the initialization of `fish_pager_color_selected_background`.

  This would change the background color of search matches to be the same as selected autocompletions and rely on the documented fallback behavior for `fish_pager_color_selected_background`.

- Add a note to the documentation that `fish_pager_color_selected_background` is set as a universal variable by default and needs to be unset in order to have the documented fallback behavior.

  This seems like it might be exposing the "guts" of `fish`'s initialization process too much for what is supposed to be user-friendly documentation.

- Don't change initialization code or documentation at all, but instead inform 3rd-party theme maintainers (like myself) that `fish_pager_color_selected_background` now needs to be explicitly set instead of relying on the fallback behavior.

  This issue might suffice, but a page/blurb on this repository's wiki would probably be better.

faho:
>This is still technically true, but because of the new initialization behavior for fish_pager_color_selected_background, this doesn't happen in practice, so the documentation is now misleading. 

[...]

PR #8793, opened 2022-03-15
highlight: Also use the fallback variable if the main is empty

faho:
Currently, when a variable like $fish_color_command is set but empty:

    set -g fish_color_command

what happens is that highlight parses it and ends up with a "normal"
color.

Change it so instead it sees that the variable is empty and goes
on to check the fallback variable, e.g. fish_color_normal.

That makes it easier to make themes that override variables.

This means that older themes that expect an empty variable to be
"normal" need to be updated to set it to "normal".

Following from this, we could make writing .theme files easier by no
longer requiring them to list all variables with specific values.
Either the theme reader could be updated to implicitly set known color
variables to empty, or the themes could feature empty values.

See #8787.

## TODOs:

- [ ] Tests have been added for regressions fixed - I don't know how to test this easily

### DIFF
--- a/doc_src/interactive.rst
+++ b/doc_src/interactive.rst
@@ -134,7 +134,7 @@ Variable                                                   Meaning
 ``fish_color_search_match``                                history search matches and selected pager items (background only)
 ==========================================                 =====================================================================
 
-If a variable isn't set, fish usually tries ``$fish_color_normal``, except for:
+If a variable isn't set or is empty, fish usually tries ``$fish_color_normal``, except for:
 
 - ``$fish_color_keyword``, where it tries ``$fish_color_command`` first.
 - ``$fish_color_option``, where it tries ``$fish_color_param`` first.
@@ -178,7 +178,7 @@ Variable                                                   Meaning
 ``fish_pager_color_secondary_description``                 description of every second unselected completion
 ==========================================                 ===========================================================
 
-When the secondary or selected variables aren't set, the normal variables are used, except for ``$fish_pager_color_selected_background``, where the background of ``$fish_color_search_match`` is tried first.
+When the secondary or selected variables aren't set or are empty, the normal variables are used, except for ``$fish_pager_color_selected_background``, where the background of ``$fish_color_search_match`` is tried first.
 
 .. _abbreviations:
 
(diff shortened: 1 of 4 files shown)


doc_src/interactive.rst
```diff
@@ -134,7 +134,7 @@ Variable                                                   Meaning
 ``fish_color_search_match``                                history search matches and selected pager items (background only)
 ==========================================                 =====================================================================
 
-If a variable isn't set, fish usually tries ``$fish_color_normal``, except for:
+If a variable isn't set or is empty, fish usually tries ``$fish_color_normal``, except for:
```
krobelus, line 137: sounds like a good idea (assuming that not too many themes rely on the empty=normal behavior)

I'm not sure how it enables themes to override variables but I didn't think very hard..

faho: What I mean by that is that this allows making a theme with only global variables, and it will override any potentially set universal variables. 

So you can start by setting all variables to empty and then set the ones you want, you don't have to *erase* variables (which for uvars would be permanent).

>assuming that not too many themes rely on the empty=normal behavior

Of *our* themes only the None theme uses any empty values, and that sets the fallback colors to normal anyway.

I don't imagine this is used much elsewhere because I don't think there is much demand for the "normal" color. Most themes, in my experience, look like [dracula.fish](https://github.com/dracula/fish/blob/master/conf.d/dracula.fish) - a fixed palette of RGB colors.

It's also very simple to fix: If you notice your theme colors being not-normal where you expected normal, simply add a "normal" where you had an empty value. This is even backwards-compatible, *and* it is more like what you would get with `set_color`.

`set_color $fish_color_option $fish_color_param` won't be normal only because $fish_color_option isn't set.

krobelus: sounds good, thanks. The new behavior is more intuitive

faho:
I just pushed commits that make `fish_config` (in both versions) set all variables the theme doesn't mention to empty.

That should make themes more self-contained and make it harder to accidentally inherit color from the previous theme.

Specifically it'll be nicer when we add a color variable, because then it would be set to empty and the fallback would kick in *even if the theme hasn't been updated*.

PR #8819, opened 2022-03-20
Add completion for rclone

page-down:
rclone added fish completion in version 1.57.

https://github.com/rclone/rclone/blob/v1.57.0/docs/content/commands/rclone_completion_fish.md

Also change CHANGELOG to use `kitty` all lowercase official project name.

### DIFF
--- a/share/completions/rclone.fish
+++ b/share/completions/rclone.fish
@@ -1,0 +1,1 @@
+rclone completion fish 2>/dev/null | source
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -39,12 +39,13 @@ Completions
 ^^^^^^^^^^^
 - Added completions for:
   - brightnessctl (:issue:`8758`)
+  - rclone (:issue:`8819`)
   - tuned-adm (:issue:`8760`)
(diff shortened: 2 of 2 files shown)


share/completions/rclone.fish
```diff
@@ -0,0 +1 @@
+command rclone completion fish 2>/dev/null | source
```
faho, line 1: We don't typically add `command` here.

The completions are sourced when something called `rclone` exists, this could also be a function. If that happens, this would error out.

Please remove it.

page-down:
> We don't typically add command here.

I removed `command`, I see lots of completions using `command`, do we need to clean them up?

faho:
>I removed command, I see lots of some completions using command, do we need to clean them up?

In short: I don't think clean up runs like that make too much sense. If someone is in there already they can have a go at it, but changing completions en masse often ends up introducing bugs in the end.

Besides, whether or not `command` is warranted has to be decided on a case-by-case basis. Typically (like here) it's not, but in some cases (like the git completions) it's better.

---

Anyway, thanks, merged!

PR #8821, opened 2022-03-20
feat(completions): add sops completions

budimanjojo:
## Description

Add [SoPS](https://github.com/mozilla/sops) shell completion.
## Issue I can't solve

- I don't know how to check if previous primary flag is a flag that require value. For example, when I want to type in: `sops --add-age "value" --decrypt` If I type in `sops --add-age ` and put in the value that it needs, then I type in `--`, it should give me list of flags that I can put in after that, but it shows me nothing but global flag. The workaround for user is to type it in like this `sops --add-age="value" --decrypt`

## TODOs:

- [ ] I still need help in the issue above


krobelus:
> I don't know how to check if previous primary flag is a flag that require value. For example, when I want to type in: `sops --add-age "value" --decrypt` If I type in `sops --add-age ` and put in the value that it needs, then I type in `--`, it should give me list of flags that I can put in after that, but it shows me nothing but global flag. The workaround for user is to type it in like this `sops --add-age="value" --decrypt`

share/completions/sops.fish
```diff
@@ -0,0 +1,147 @@
...
+    return 1
+  end
+  if string match -qr -- "sops\s+$argv[1]\s+\S*" $cmd
+    return 0
+  end
+  return 1
+end
+
+# Primary commands
+complete -F -c sops -n __fish_is_first_arg -a exec-env -d "execute a command with decrypted values inserted into the environment"
```
krobelus, line 21: I think we typically  start descriptions (`-d`) with an uppercase letter

share/completions/sops.fish
```diff
@@ -0,0 +1,147 @@
...
+complete -F -c sops -n __fish_is_first_arg -a help -d "shows a list of commands or help for one command"
+complete -F -c sops -n __fish_is_first_arg -a h -d "shows a list of commands or help for one command"
+
+# Primary flags without parameters
+complete -F -c sops -n __fish_is_first_token -s d -l decrypt -d "decrypt a file and output the result to stdout"
+complete -F -c sops -n __fish_is_first_token -s e -l encrypt -d "encrypt a file and output the result to stdout"
+complete -F -c sops -n __fish_is_first_token -s r -l rotate -d "generate a new data encryption key and reencrypt all values with the new key"
+complete -F -c sops -n __fish_is_first_token -s i -l in-place -d "write output back to the same file instead of stdout"
+complete -F -c sops -n __fish_is_first_token -s s -l show-master-keys -d "display master encryption keys in the file during editing"
+complete -F -c sops -n __fish_is_first_token -l extract -d "extract a specific key or branch from the input document. Decrypt mode only. Example: --extract '[\"somekey\"][0]'"
```
krobelus, line 36: The description is probably too long and might be cut off, so how about

```suggestion
complete -F -c sops -n __fish_is_first_token -l extract -d "extract a specific key or branch from the decrypted input document"
```

same for some other descriptions. I think 60 chars is our limit. They are summaries, not documentation.

share/completions/sops.fish
```diff
@@ -0,0 +1,147 @@
...
+complete -x -c sops -n __fish_is_first_token -l gcp-kms -d "comma separated list of GCP KMS resource IDs [\$SOPS_GCP_KMS_IDS]"
+complete -x -c sops -n __fish_is_first_token -l aws-profile -d "the AWS profile to use for requests to AWS"
+complete -x -c sops -n __fish_is_first_token -l azure-kv -d "comma separated list of Azure Key Vault URLs [\$SOPS_AZURE_KEYVAULT_URLS]"
+complete -x -c sops -n __fish_is_first_token -l hc-vault-transit -d "comma separated list of Vault's key URI (e.g. 'https://vault.example.org:8200/v1/transit/keys/dev') [\$SOPS_VAULT_URIS]"
+complete -x -c sops -n __fish_is_first_token -l input-type -d "currently json, yaml, dotenv and binary are supported. If not set, sop will use the file's extension to determine the type"
+complete -x -c sops -n __fish_is_first_token -l output-type -d "currently json, yaml, dotenv and binary are supported. If not set, sop will use the input file's extension to determine the output format"
+complete -x -c sops -n __fish_is_first_token -l add-gcp-kms -d "add the provided comma-separated list of GCP KMS key resource IDs to the list of master keys on the given file"
```
krobelus, line 52: `-x` means you're disabling file completion for this option but this option wants a file argument.
Please try `-r` intead. Same for some other options.

share/completions/sops.fish
```diff
@@ -0,0 +1,147 @@
...
+complete -x -c sops -n "__fish_sops_commands exec-env" -l user -a "(__fish_print_users)" -d "the user to run the command as"
+complete -F -c sops -n "__fish_sops_commands exec-env" -l enable-local-keyservice -d "use local key service"
+complete -x -c sops -n "__fish_sops_commands exec-env" -l keyservice -d "specify the key services to use in addition to the local one. Can be specified more than once. Syntax: protocol://address. Example: tcp://myserver.com:5000"
+
+# exec-file flags
+complete -F -c sops -n "__fish_sops_commands exec-file" -l background -d "background the process and don\'t wait for it to complete"
+complete -F -c sops -n "__fish_sops_commands exec-file" -l no-fifo -d "use a regular file instead of a fifo to temporarily hold the decrypted contents"
+complete -x -c sops -n "__fish_sops_commands exec-file" -l user -a "(__fish_print_users)" -d "the user to run the command as"
+complete -x -c sops -n "__fish_sops_commands exec-file" -l input-type -a "json;yaml;dotenv;binary" -d "currently json, yaml, dotenv and binary are supported. If not set, sops will use the file's extension to determine the output format"
```
krobelus, line 88: I'm surprised `-a "json;yaml;dotenv;binary"` works, we usually use spaces instead of ;

```suggestion
complete -x -c sops -n "__fish_sops_commands exec-file" -l input-type -a "json yaml dotenv binary" -d "currently json, yaml, dotenv and binary are supported. If not set, sops will use the file's extension to determine the output format"
```

share/completions/sops.fish
```diff
@@ -0,0 +1,147 @@
...
+complete -x -c sops -n "__fish_sops_commands exec-file" -l output-type -a "json;yaml;dotenv;binary" -d "currently json, yaml, dotenv and binary are supported. If not set, sops will use the input file's extension to determine the output format"
+complete -F -c sops -n "__fish_sops_commands exec-file" -l filename -d "filename for the temporarily file (default: tmp-file)"
+complete -F -c sops -n "__fish_sops_commands exec-file" -l enable-local-keyservice -d "use local key service"
+complete -x -c sops -n "__fish_sops_commands exec-file" -l keyservice -d "specify the key services to use in addition to the local one. Can be specified more than once. Syntax: protocol://address. Example: tcp://myserver.com:5000"
+
+# publish flags
+complete -F -c sops -n "__fish_sops_commands publish" -s y -l yes -d "pre-approve all changes and run non-interactively"
+complete -F -c sops -n "__fish_sops_commands publish" -l omit-extentions -d "omit file extensions in destination path when publishing sops file to configured destination"
```
krobelus, line 96: typo?

```suggestion
complete -F -c sops -n "__fish_sops_commands publish" -l omit-extensions -d "omit file extensions in destination path when publishing sops file to configured destination"
```

share/completions/sops.fish
```diff
@@ -0,0 +1,147 @@
...
+complete -F -c sops -n __fish_is_first_arg -a exec-file -d "execute a command with decrypted contents as a temporary file"
+complete -F -c sops -n __fish_is_first_arg -a publish -d "publish sops file or directory to a configured destination"
+complete -F -c sops -n __fish_is_first_arg -a keyservice -d "start a SOPS key kervice server"
+complete -F -c sops -n __fish_is_first_arg -a groups -d "modify the groups on a SOPS file"
+complete -F -c sops -n __fish_is_first_arg -a updatekeys -d "update the keys of a SOPS file using the config file"
+complete -F -c sops -n __fish_is_first_arg -a help -d "shows a list of commands or help for one command"
+complete -F -c sops -n __fish_is_first_arg -a h -d "shows a list of commands or help for one command"
+
+# Primary flags without parameters
+complete -F -c sops -n __fish_is_first_token -s d -l decrypt -d "decrypt a file and output the result to stdout"
```
krobelus, line 31: > I don't know how to check if previous primary flag is a flag that require value. For example, when I want to type in: `sops --add-age "value" --decrypt` If I type in `sops --add-age ` and put in the value that it needs, then I type in `--`, it should give me list of flags that I can put in after that, but it shows me nothing but global flag. The workaround for user is to type it in like this `sops --add-age="value" --decrypt`

I think you should just drop the  `-n __fish_is_first_token` for all global options.

Then you'll get completions for `sops --add-age value --`

share/completions/sops.fish
```diff
@@ -0,0 +1,147 @@
+# Completion for SOPS (Secrets OPerationS)
+
+function __fish_sops_commands --description "Test if argument(s) match a sops command"
+  if not set -q argv[1]
+    return 1
+  end
+  set -l cmd (commandline -cp)
+  if set -q argv[2]
+    if string match -qr -- "sops\s+$argv[1]\s+$argv[2]\s+\S*" $cmd
```
krobelus, line 9: Can you try to replace uses of this function with `__fish_seen_subcommand_from`?

I believe if there are multiple levels of subcommands, it's okay to use `__fish_seen_subcommand_from subcmd && __fish_seen_subcommand_from subsubcmd` (not always fully correct but good enough)

share/completions/sops.fish
budimanjojo: Here is the problem about `sops` that I encounter. The "global options" is acting kinda weird because it needs to be declared before the subcommand.
This works:
```
sops --decrypt exec-env out.json 'echo $password'
```
These don't work:
```
sops exec-env --decrypt out.json 'echo $password'
sops exec-env out.json 'echo $password' --decrypt
```

That's the reason behind why I used `-n __fish_is_first_token` to make sure the global options are defined before subcommands.

share/completions/sops.fish
budimanjojo: These options wants a value instead of files though. I have `-l output` and `-l config` with `-r` because only these two options wants a file.

share/completions/sops.fish
budimanjojo: Do you mean deleting this whole `__fish_sops_commands` function and use `__fish_seen_subcommand_from` instead?

budimanjojo:
@krobelus Thanks for the reviews! I have commited some of your suggestions, and commented on those I don't quite understand. Please have another look :)

share/completions/sops.fish
krobelus: Extra completions are not as bad as missing completions, so it would be okay-ish to remove the condition

Though the fix should be doable; try this trick (adapted from sudo.fish):

```fish
function __fish_sops_no_subcommand
    not __fish_sops_has_subcommand
end
function __fish_sops_has_subcommand
    set cmdline_args (commandline -opc)
    set -e cmdline_args[1]
    argparse -s A-add-age= D-decrypt -- $cmdline_args
    set -l remaining_args $argv
    set -q remaining_args[1]
end
complete -c sops -n __fish_sops_no_subcommand -l add-age -x -d "add the provided comma-separated list of age recipients fingerprints to the list of master keys on the given file"
complete -c sops -n __fish_sops_no_subcommand -s d -l decrypt -d "decrypt a file and output the result to stdout"
```

share/completions/sops.fish
krobelus: > Do you mean deleting this whole `__fish_sops_commands` function and use `__fish_seen_subcommand_from` instead? 

yep

share/completions/sops.fish
krobelus: Ah sorry, I think I commented on the wrong completion. Looks like there are surprisingly few options that take file arguments.

krobelus:
> I don't know why the check is not successful though, completion files shouldn't affect builds right?

yeah, no need to worry. We have some flaky tests.

share/completions/sops.fish
krobelus: even better, replace it with a function that uses argparse to decide which subcommand is active (similar to `__fish_sudo_print_remaining_args`)

share/completions/sops.fish
budimanjojo: Sorry, I read through `argparse` manpage and I can't seem to find what is `A-add-age=` and `D-decrypt`. Shouldn't it be `add-age=` and `d/decrypt`?

So I have to list all the primary flags to the line `argparse -s OPTION_SPEC -- $cmdline_args`?

share/completions/sops.fish
budimanjojo: The reason why I didn't use `__fish_seen_subcommand_from subcmd1 && __fish_seen_subcommand_from subcmd2` for  multiple levels subcommands is because it accepts anything as long as there are the word `subcmd1` and `subcmd2` ignoring the placement. For example, if I want `group` as `subcmd1` and `add` as `subcmd2` then even if I write it as `sops blablabla add groups` it will still return true. While the function that I'm using it right now (I took the code from `__btrfs_commands` and `__btrfs_command_groups`) will return false unless the command is `sops groups add` and nothing else.

budimanjojo: I might go with the even better "similar to `__fish_sudo_print_remaining_args`" approach, but it might need some time because I'm still figuring out how does `argparse` work. T.T

share/completions/sops.fish
krobelus: > I can't seem to find what is `A-add-age=`

It's this part of the docs

> [..] for backwards compatibility, a ``-`` if it should not be exposed as a valid short flag (in which case it will also not be exposed as a flag variable).

However, I just remembered that the short flag is no longer necessary since fish 3.2.0. So you should write `argparse add-age=` if there's no corresponding short option.
We should also fix that in our other scripts (I copied the legacy style)

> So I have to list all the primary flags to the line `argparse -s OPTION_SPEC -- $cmdline_args`?

Yep, everything that can appear before a command

share/completions/sops.fish
krobelus: Ok makes sense. I think copying `__fish_sudo_print_remaining_args` is the way to go. We should probably add more examples to the argparse docs.

budimanjojo: I have modified the way it works, please have another look @krobelus. I don't know if the code is good as I'm not experienced enough for this (not even sure what I did lol), please correct my code if needed. :)

krobelus:
Thanks!

Issue #8786, opened 2022-03-14
Expose more limits in `ulimit`


septatrix:
* Version: `3.3.1`
* OS: Fedora 35, Kernel Version: `5.16.12-200.fc35.x86_64`

The bash builtin `ulimit` (which is also the fallback/`command ulimit`) exposes many more limits which fish does not show. Some of them might be unique to bash and others seem to have a reason for excluding `-p` the are quite a few which are missing. The list seems to be `eipqrsx`

```fish
~ $ bash -c "type ulimit && ulimit -a"
ulimit is a shell builtin
real-time non-blocking time  (microseconds, -R) unlimited
core file size              (blocks, -c) unlimited
data seg size               (kbytes, -d) unlimited
scheduling priority                 (-e) 0
file size                   (blocks, -f) unlimited
pending signals                     (-i) 95584
max locked memory           (kbytes, -l) 8192
max memory size             (kbytes, -m) unlimited
open files                          (-n) 1024
pipe size                (512 bytes, -p) 8
POSIX message queues         (bytes, -q) 819200
real-time priority                  (-r) 0
stack size                  (kbytes, -s) 8192
cpu time                   (seconds, -t) unlimited
max user processes                  (-u) 95584
virtual memory              (kbytes, -v) unlimited
file locks                          (-x) unlimited
~ $ type ulimit && ulimit -a
ulimit is a builtin
Maximum size of core files created                           (kB, -c) unlimited
Maximum size of a process’s data segment                     (kB, -d) unlimited
Maximum size of files created by the shell                   (kB, -f) unlimited
Maximum size that may be locked into memory                  (kB, -l) 8192
Maximum resident set size                                    (kB, -m) unlimited
Maximum number of open file descriptors                          (-n) 1024
Maximum stack size                                           (kB, -s) 8192
Maximum amount of cpu time in seconds                   (seconds, -t) unlimited
Maximum number of processes available to a single user           (-u) 95584
Maximum amount of virtual memory available to the shell      (kB, -v) unlimited
```

faho:
This isn't something the shell should care about, to be honest. It's awkward platform-specific guff that's only in the shell because that's where it's been placed historically.

Very very few people even care about the ulimit we *do* have. I would be entirely okay with removing it.

So I'm rejecting this.

septatrix:
> This isn't something the shell should care about, to be honest. It's awkward platform-specific guff that's only in the shell because that's where it's been placed historically.

[...]

PR #8823, opened 2022-03-21
Add extra resource limits to ulimit

zanchey:
I've been through the manual pages for Linux, and added most of the the resource
limits that aren't already available. I skipped `RLIMIT_LOCKS` because it only
applies to long-obsolete versions of the kernel.

Fixes issue #8786.

## TODOs:


src/builtins/ulimit.cpp
```diff
@@ -200,10 +220,22 @@ maybe_t<int> builtin_ulimit(parser_t &parser, io_streams_t &streams, const wchar
...
+                what = RLIMIT_NICE;
+                break;
+            }
+#endif
             case 'f': {
                 what = RLIMIT_FSIZE;
                 break;
             }
+#ifdef RLIMIT_SIGPENDING
+            case 'i': {
```
faho, line 234: I'm assuming these option letters are just taken from bash?

zanchey: Most of them aren't in bash, so I've taken them from [prlimit(1)](https://man7.org/linux/man-pages/man1/prlimit.1.html).

src/builtins/ulimit.cpp
```diff
@@ -200,10 +220,22 @@ maybe_t<int> builtin_ulimit(parser_t &parser, io_streams_t &streams, const wchar
                 what = RLIMIT_DATA;
                 break;
             }
+#ifdef RLIMIT_NICE
```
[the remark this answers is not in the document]
zanchey: Maybe set the `what` value to -1 if the option is not defined, and print an error further down?

These are all on various versions of Linux. There are a bunch more on FreeBSD and NetBSD, which I can add. OpenBSD and macOS have a couple of extras too, but they're a subset of the others. Solaris has some too.

zanchey:
I've included the macros from NetBSD and FreeBSD, added some basic tests, updated the completions and the documentation, and made the utility return a specific error if the option is valid but not supported.

faho (approved):
LGTM!

PR #8827, opened 2022-03-23
Autoload: Call the parser directly instead of going via "subshell"

faho:
This used to call exec_subshell, which has two issues:

1. It creates a command substitution block which shows up in a stack
trace
2. It does much more work than necessary

This removes a useless "in command substitution" from an error message
in an autoloaded file, and it speeds up autoloading a bit (not
measurable in actual benchmarks, but microbenchmarks are 2x).

The stack trace goes from:

```
/tmp/autoloads/error.fish (line 1): Missing end to balance this function definition
function error
^
from sourcing file /tmp/autoloads/error.fish
in command substitution
```

to

```
/tmp/autoloads/error.fish (line 1): Missing end to balance this function definition
function error
^
from sourcing file /tmp/autoloads/error.fish
```

## TODOs:

@ridiculousfish I'd like your eyes on this. I believe the subshell stuff *should* be unnecessary here?

(tbh I dislike the term "subshell" here because that's a specific thing in posixy shells and this is not it)

### DIFF
--- a/src/autoload.cpp
+++ b/src/autoload.cpp
@@ -205,6 +205,14 @@ maybe_t<wcstring> autoload_t::resolve_command(const wcstring &cmd, const wcstrin
 }
 
 void autoload_t::perform_autoload(const wcstring &path, parser_t &parser) {
+    // We do the useful part of what exec_subshell does ourselves
+    // - we source the file.
+    // We don't create a buffer or check ifs or create a read_limit
+
     wcstring script_source = L"source " + escape_string(path, ESCAPE_ALL);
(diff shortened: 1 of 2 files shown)


ridiculousfish:
Nice improvement. The only potential problem is that `skip_notification` will no longer be set for the `source` job:

    props.skip_notification =
        ld.is_subshell || parser->is_block() || ld.is_event || !parser->is_interactive();

because `is_subshell` will now be false. But we don't notify for jobs without pids anyway so it's probably benign.

ridiculousfish (approved):

PR #8609, opened 2022-01-03
Add `isatty` completion

kidonng:
## Description

Even though `isatty` is not very useful interactively, completions will help if the user is interactively writing a function which uses `isatty`.

## TODOs:

- [ ] Changes to fish usage are reflected in user documentation/manpages.
- [ ] Tests have been added for regressions fixed
- [ ] User-visible changes noted in CHANGELOG.rst

PR #8840, opened 2022-03-28
Complete `/dev/fd` in `isatty`

kidonng:
## Description

Complement to #8609.

## TODOs:

### DIFF
--- a/share/completions/isatty.fish
+++ b/share/completions/isatty.fish
@@ -1,1 +1,5 @@
-complete -c isatty -x -k -a "stdin stdout stderr"
+complete -c isatty -x 
+complete -c isatty -k -a "(string replace /dev/fd/ '' /dev/fd/*)"
+complete -c isatty -k -a "stderr" -d "2"
+complete -c isatty -k -a "stdout" -d "1"
+complete -c isatty -k -a "stdin" -d "0"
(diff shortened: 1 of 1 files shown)


share/completions/isatty.fish
```diff
@@ -1 +1,5 @@
-complete -c isatty -x -k -a "stdin stdout stderr"
+complete -c isatty -x 
+complete -c isatty -k -a "(string replace /dev/fd/ '' /dev/fd/*)"
```
faho, line 2: This will error out if `/dev/fd` doesn't exist. Which it won't on non-linux systems, if I recall correctly.

floam: FYI it exists here on macOS, and I think it's been in BSD forever but it should still be fixed for cases where it does not exist.

```shell
ls -l /dev/fd
total 0
crw--w----   1 floam  tty    0x10000004 Mar 28 11:35 0
crw--w----   1 floam  tty    0x10000004 Mar 28 11:35 1
crw--w----   1 floam  tty    0x10000004 Mar 28 11:35 2
drw-r-----  55 floam  staff        1760 Mar 28 11:34 3/
dr--r--r--   1 root   wheel           0 Mar 16 15:55 4/
```

faho:
Thanks!

Issue #8465, opened 2021-11-21
Behaviour of --require-parameter is not the same for long-style and old-style options


EmilySeville7cfg:
Fish version: fish, version 3.3.1-622-ge06671512 (built from source)
OS version: Linux Mint 20.2 Cinnamon

`--require-parameter` doesn't affect old-style options correctly:

- wrong: `x` and `y` are shown both for `./exe -old-first<space>` and `./exe -old-second<space>` when pressing `tab` while for long options `x` and `y` are shown only for `./exe -old-second<space>`

```fish
complete --command exe --old-option old-first --arguments='x y'
complete --command exe --old-option old-second --require-parameter --arguments='x y'

complete --command exe --long-option long-first --arguments='x y'
complete --command exe --long-option long-second --require-parameter --arguments='x y'
```

krobelus:
very interesting. I believe that `exe -old-first <TAB>` should not show `x` or `y` because the argument is optional, so file completion should kick in.

Curiously, we have completions that rely on this behavior

```fish
complete -c find -o maxdepth -d "Maximum recursion depth" -a "1 2 3 4 5 6 7 8 9"
```

observe that we forgot to add `-r`. But in most places we do add it.

So it looks like we could fix that, *however*, there could be getopt implementations around that read `-flag-with-optional-arg the-arg` as a flag with an optional argument.
(For example Python's argparse has this weird behavior, but I'm not sure if they have old-style options.)
So we need a bit of research, because this depends on how popular getopt parsers work.

We have this comment in `src/complete.cpp:1078`

> // Optional arguments to a switch can only be handled using the '=', so we add it as
> // a completion. By default we avoid using '=' and instead rely on '--switch
> // switch-arg', since it is more commonly supported by homebrew getopt-like
> // functions.

PR #8863, opened 2022-04-07
Update terraform completions

tomrijnbeek:
## Description

This PR updates the built-in completions for Terraform, which were outdated.

The completions have been checked against the official help commands of the Terraform CLI version 1.1.7.

Terraform uses old style options (long names, but single hyphen, e.g. `-no-color`), which doesn't play super nicely with some of the completion mechanics it seems. For the time being, the global `-chdir=PATH` option isn't supported, as I wasn't sure how to nicely parse it out and still determine whether a subcommand was still needed. In general, having `-option=arg` style options appears problematic, as the completion automatically inserts a whitespace after the `=` symbol. However, Terraform doesn't seem to particularly mind getting the arguments in `-option arg` format either, so this is not a major issue. That being said, for cases where the Terraform help command shows only a single option (e.g. `-lock=false` or `-input=true`), I have opted to put the whole phrase as completion, as the Terraform documentation does not appear to expect other options to be passed in:

image

The subcommands for `state` and `workspace` do not have further completions at this time.

While by no means complete, all in all, I still think this is a worthwhile PR, since it improves the completion over the status quo.

Note that I opted to replace single quotes with double quotes, since variable interpretation didn't seem to play nicely with single quote strings. There does not appear to be a de facto standard across all other completion scripts, but it does provide a bit of a dirty diff. I can revert and use string concatenation if preferred.

## TODOs:

- [x] Changes to fish usage are reflected in user documentation/manpages [N/A]
- [x] Tests have been added for regressions fixed [N/A]

### DIFF
--- a/share/completions/terraform.fish
+++ b/share/completions/terraform.fish
@@ -10,6 +10,10 @@ function __fish_terraform_needs_command
     return 1
 end
 
+function __fish_terraform_workspaces
+    terraform workspace list | string replace -r "^[\s\*]*" ""
+end
+
 # general options
 complete -f -c terraform -n "not __fish_terraform_needs_command" -o version -d "Print version information"
 complete -f -c terraform -o help -d "Show help"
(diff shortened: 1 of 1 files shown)


krobelus:
> Terraform uses old style options (long names, but single hyphen, e.g. `-no-color`), which doesn't play super nicely with some of the completion mechanics it seems.

eh, they should be usable

> For the time being, the global `-chdir=PATH` option isn't supported, as I wasn't sure how to nicely parse it out and still determine whether a subcommand was still needed.

Example completions that do something similar are make, ninja, git and sudo (I think that one is the simplest one to copy). Can be done later of course.

> In general, having `-option=arg` style options appears problematic, as the completion automatically inserts a whitespace after the `=` symbol. However, Terraform doesn't seem to particularly mind getting the arguments in `-option arg` format either, so this is not a major issue.

Yeah I think that's https://github.com/fish-shell/fish-shell/issues/8465
Not sure if we're gonna change anything. Anyway, it's harmless.

> That being said, for cases where the Terraform help command shows only a single option (e.g. `-lock=false` or `-input=true`), I have opted to put the whole phrase as completion, as the Terraform documentation does not appear to expect other options to be passed in:

That's great.

share/completions/terraform.fish
```diff
@@ -1,178 +1,205 @@
...
-complete -f -c terraform -n '__fish_seen_subcommand_from apply' -o state -d 'Path to a Terraform state file'
-complete -f -c terraform -n '__fish_seen_subcommand_from apply' -o state-out -d 'Path to write state'
-complete -f -c terraform -n '__fish_seen_subcommand_from apply' -o target -d 'Resource to target'
-complete -f -c terraform -n '__fish_seen_subcommand_from apply' -o var -d 'Set a variable in the Terraform configuration'
-complete -f -c terraform -n '__fish_seen_subcommand_from apply' -o var-file -d 'Set variables from a file'
+complete -f -c terraform -n "not __fish_terraform_needs_command" -o version -d "Print version information"
+complete -f -c terraform -o help -d "Show help"
+
+### apply/destroy
+set -l apply apply destroy
```
krobelus, line 18: would be easier to review if you placed this refactoring change (the consolidation of some complete-statements) in a separate commit.

share/completions/terraform.fish
```diff
@@ -1,178 +1,205 @@
...
+complete -f -c terraform -n "__fish_seen_subcommand_from refresh" -o backup -d "Path to backup the existing state file"
+complete -f -c terraform -n "__fish_seen_subcommand_from refresh" -o input=true -d "Ask for input for variables if not directly set"
+complete -f -c terraform -n "__fish_seen_subcommand_from refresh" -o lock=false -d "Don't hold a state lock"
+complete -f -c terraform -n "__fish_seen_subcommand_from refresh" -o lock-timeout -d "Duration to retry a state lock"
+complete -f -c terraform -n "__fish_seen_subcommand_from refresh" -o no-color -d "If specified, output won't contain any color"
+complete -f -c terraform -n "__fish_seen_subcommand_from refresh" -o state -d "Path to a Terraform state file"
+complete -f -c terraform -n "__fish_seen_subcommand_from refresh" -o state-out -d "Path to write state"
+complete -f -c terraform -n "__fish_seen_subcommand_from refresh" -o target -d "Resource to target"
+complete -f -c terraform -n "__fish_seen_subcommand_from refresh" -o var -d "Set a variable in the Terraform configuration"
+complete -f -c terraform -n "__fish_seen_subcommand_from refresh" -o var-file -d "Set variables from a file"
```
krobelus, line 149: There is another bug here: `terraform refresh -var-file ` should offer file completions. Same for some others (like `-state`)

You can fix this by replacing `complete -f` with `complete -r`

share/completions/terraform.fish
```diff
@@ -1,178 +1,205 @@
...
 ### version
-complete -f -c terraform -n __fish_use_subcommand -a version -d 'Print the Terraform version'
+complete -f -c terraform -n __fish_terraform_needs_command -a version -d "Print the Terraform version"
+
+### workspace
+complete -f -c terraform -n __fish_terraform_needs_command -a workspace -d "Workspace management"
+complete -f -c terraform -n "__fish_seen_subcommand_from workspace" -a list -d "List workspaces"
+complete -f -c terraform -n "__fish_seen_subcommand_from workspace" -a select -d "Select an workspace"
+complete -f -c terraform -n "__fish_seen_subcommand_from workspace" -a new -d "Create a new workspace"
+complete -f -c terraform -n "__fish_seen_subcommand_from workspace" -a delete -d "Delete an existing workspace"
```
krobelus, line 205: Completely untested but I imagine something like this could be useful

	complete -f -c terraform -n "__fish_seen_subcommand_from workspace && __fish_seen_subcommand_from select delete" -a "(terraform workspace list)"

tomrijnbeek: I really like the idea. It needed a bit more work, but I've now added an autocomplete for the workspace select and delete. We may be able to do something similar with state, but it's a bit less straightforward.

tomrijnbeek (commented):
Thanks for the review comments. I have split the initial work over three separate commits (had to use a force push to override).

krobelus:
Thanks!

Issue #8873, opened 2022-04-13
Starting interactive shell, without setting TERM, causes SIGSEGV


mojotx:
- Fish Version: `fish, version 3.4.1-168-g2bf9b97ab`
- OS: macOS

I recently noticed an [annoying issue](https://github.com/microsoft/vscode/issues/147320) when I tried launching the Microsoft VSCode editor from the macOS Dock. I was presented with an error, _"Unable to resolve your shell environment: Unexpected exit code from spawned shell (code null, signal SIGSEGV)."_

![Screen Shot 2022-04-12 at 13 57 12](https://user-images.githubusercontent.com/5694899/163263030-276ba306-7e6b-437a-a4a7-dbbf4b4d5589.png)

This was puzzling, because I could run a fish in a terminal session within the editor just fine. I did some digging through the Visual Studio Code codebase, and eventually figured out how to bump up the logging level, while still launching the editor as a macOS "app", since launching VSCode from the command-line would not trigger the bug. 

From vscode main.log:
``` 
[2022-04-13 14:08:12.112] [main] [trace] getUnixShellEnvironment#stderr warning: Could not set up terminal.
warning: TERM environment variable not set.
```

### TL;DR:
The bug is caused by the `TERM` environment variable being set. I then checked a Linux box running fish, to see if the problem was specific to macOS, and saw that the problem does NOT occur with the recent fish 3.4.1 release. 

I was then able to bisect the bug to commit 9c96986b3688974f9381e8509d66a73af1dc8a4f, but it's not immediately obvious why it's causing a SIGSEGV, since that was only a change to `src/builtins/set_color.cpp`.  Does `set_color` always get executed?

With fish prior to commit 9c96986b3688974f9381e8509d66a73af1dc8a4f:
```
$ sh -c 'env -i HOME=$(mktemp -d) /Users/mjarvis/src/github.com/fish-shell/build/fish'
warning: Could not set up terminal.
warning: TERM environment variable not set.
Welcome to fish, the friendly interactive shell
Type help for instructions on how to use fish
```

With fish >= fish 3.4.1-9c96986b3:
```
$ sh -c 'env -i HOME=$(mktemp -d) /Users/mjarvis/src/github.com/fish-shell/build/fish'
warning: Could not set up terminal.
warning: TERM environment variable not set.
fish: Job 1, 'sh -c 'env -i HOME=$(mktemp -d)…' terminated by signal SIGSEGV (Address boundary error)
```

[...]

PR #8875, opened 2022-04-14
Remove test for italics_mode and dim_mode on Apple

mojotx:
This resolves an issue on Apple macOS, where fish crashes with SIGSEGV if the TERM environment
variable is not set.

See:
- https://github.com/fish-shell/fish-shell/issues/8873
- https://github.com/microsoft/vscode/issues/147320

## Description

It appears that the set_color built-in is always called on a normal start-up to the shell, and if the `$TERM` environment variable is not set, we dump core with a SIGSEGV. 

To reproduce:

```sh
$ sh -c 'env -i HOME=$(mktemp -d) /Users/mjarvis/src/github.com/mojotx/fish-shell/build/fish'
```

Fixes issue #8873

### DIFF
--- a/src/builtins/set_color.cpp
+++ b/src/builtins/set_color.cpp
@@ -113,17 +113,14 @@ maybe_t<int> builtin_set_color(parser_t &parser, io_streams_t &streams, const wc
 #ifdef __APPLE__
     // Hack in missing italics and dim capabilities omitted from MacOS xterm-256color terminfo
     // Helps Terminal.app/iTerm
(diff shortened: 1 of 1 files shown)


faho:
This reverts 9c96986b3688974f9381e8509d66a73af1dc8a4f - looks alright to me, once the tests pass I'll merge it.

mojotx:
Basically this backs out the test for `enter_italics_mode` and `enter_dim_mode` that is missing from the macOS xterm-256color terminfo. There may be a better way to fix the core dump issue, and still test those values.

PR #8876, opened 2022-04-14
Re-add fix for Apple terminfo

mojotx:
## Description

Talk about your changes here.

Related to issue #8873 and #8875

### DIFF
--- a/src/builtins/set_color.cpp
+++ b/src/builtins/set_color.cpp
@@ -113,14 +113,19 @@ maybe_t<int> builtin_set_color(parser_t &parser, io_streams_t &streams, const wc
 #ifdef __APPLE__
     // Hack in missing italics and dim capabilities omitted from MacOS xterm-256color terminfo
     // Helps Terminal.app/iTerm
-    const auto term_prog = parser.vars().get(L"TERM_PROGRAM");
-    if (!term_prog.missing_or_empty() && (term_prog->as_string() == L"Apple_Terminal" ||
-        term_prog->as_string() == L"iTerm.app")) {
-        const auto term = parser.vars().get(L"TERM");
-        if (!term.missing_or_empty() && (term->as_string() == L"xterm-256color")) {
-            enter_italics_mode = sitm_esc;
-            exit_italics_mode = ritm_esc;
-            enter_dim_mode = dim_esc;
+    const auto term = parser.vars().get(L"TERM");
+    if (!term.missing_or_empty()) {
+        if ((!enter_italics_mode || enter_italics_mode[0] == '\0') ||
+            (!enter_dim_mode || enter_dim_mode[0] == '\0')) {
+            const auto term_prog = parser.vars().get(L"TERM_PROGRAM");
+            if (!term_prog.missing_or_empty() && (term_prog->as_string() == L"Apple_Terminal" ||
+                                                  term_prog->as_string() == L"iTerm.app")) {
+                if (term->as_string() == L"xterm-256color") {
+                    enter_italics_mode = sitm_esc;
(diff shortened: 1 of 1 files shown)


mojotx:
```
sh -c 'env -i HOME=$(mktemp -d) /Users/mjarvis/src/github.com/mojotx/fish-shell/build/fish
warning: Could not set up terminal.
warning: TERM environment variable not set.
Type help for instructions on how to use fish
mjarvis@TMNA-MBP-VST-345 /U/m/s/g/m/f/build (apple_terminfo)>
```

mojotx:
@floam : Can you review this? We had to back out 9c96986b3688974f9381e8509d66a73af1dc8a4f with #8875. 

This resolves #8875. I tried to re-add the logic from your commit, but still fix the SIGSEGV issue. Thanks!

src/builtins/set_color.cpp
```diff
@@ -113,14 +113,21 @@ maybe_t<int> builtin_set_color(parser_t &parser, io_streams_t &streams, const wc
...
-    if (!term_prog.missing_or_empty() && (term_prog->as_string() == L"Apple_Terminal" ||
-        term_prog->as_string() == L"iTerm.app")) {
+    {
         const auto term = parser.vars().get(L"TERM");
-        if (!term.missing_or_empty() && (term->as_string() == L"xterm-256color")) {
-            enter_italics_mode = sitm_esc;
-            exit_italics_mode = ritm_esc;
-            enter_dim_mode = dim_esc;
+        if (!term.missing_or_empty()) {
+            if ((!enter_italics_mode || enter_italics_mode[0] == '\0') ||
```
krobelus, line 119: Changing `enter_dim_mode == ""` to `enter_dim_mode[0] == '\0'` makes sense, although, from a quick look, I still don't understand how the problem happened, and how this fixes it. Care to describe the root cause?

> Basically, move the test for term.missing_or_empty() to the beginning of
> the logic.

Does that really change anything?

mojotx: > Care to describe the root cause?

The root problem is that when you launched fish without setting the `TERM` environment variable, it would die with SIGSEGV. 

I fixed the core dump issue with #8875, which basically reverted @floam's commit 9c96986b3688974f9381e8509d66a73af1dc8a4f. It's good that it fixed the core dump issue, but bad that it stripped out the dim/italic issue that Aaron was addressing. 

To be honest, I am not sure WHY it is dumping core. Obviously, it is accessing memory somewhere that it shouldn't, and checking to see if TERM is set fixes the issue, but I don't know specifically which line is causing the crash. If I get some time, I will generate some core files and see if I can find the problem. However, this is a pretty minor edge case, that only affects Mac users with broken terminfo databases and/or tools that don't set the $TERM environment variable correctly. 

> Does that really change anything?

I don't know for sure if it fixes the dim/italics issue, which was why I was hoping Aaron could chime in with some context.

ridiculousfish:
So the real issue here is that e.g. `enter_italics_mode` is defined as:

    #define CUR cur_term->type.
    ...
    #define enter_italics_mode             CUR Strings[311]

indirecting through the `cur_term` global variable. If this variable is null then reading or writing `enter_italics_mode` will crash.

Beyond that, if we are going to modify `enter_italics_mode`, we should do it when the term is set up, not opportunistically in `set_color`.

Thank you for proposing the fix, I'm going to fix this in a different way.

mojotx:
Great, I am glad you spotted the root cause!

ridiculousfish:
bd9c6a64e39a2d1650c0040e4b03ab04915882e8

PR #8879, opened 2022-04-15
Setup $USER and $HOME if passwd for $USER has different uid

### LINKED ISSUE #8583: Don't trust $USER
fish, version 3.3.1 in chroot

I login with user safety(though it's unsafe at all),why fish shell think im a root user,ash and bash think im not root user

2021-12-25 21-02-23 的螢幕擷圖

2021-12-25 21-04-01 的螢幕擷圖

faho:
This gets the passwd entry for $USER (if it is set). If that gives the
same uid that getuid() gives us, we assume the data is correct.

If not, we reset $HOME and $USER from the passwd value for our UID.

Fixes #8583

## TODOs:

### DIFF
--- a/src/env.cpp
+++ b/src/env.cpp
@@ -325,11 +325,8 @@ void env_init(const struct config_paths_t *paths, bool do_uvars, bool default_pa
         }
     }
 
-    // Some `su`s keep $USER when changing to root.
-    // This leads to issues later on (and e.g. in prompts),
-    // so we work around it by resetting $USER.
-    // TODO: Figure out if that su actually checks if username == "root"(as the man page says) or
-    // UID == 0.
+    // Set $USER, $HOME and $EUID
+    // This involves going to passwd and stuff.
     vars.set_one(L"EUID", ENV_GLOBAL, to_string(static_cast<unsigned long long>(geteuid())));
     setup_user();
 
--- a/src/env.cpp
+++ b/src/env.cpp
@@ -235,15 +235,19 @@ static void setup_user() {
     if (!retval && result) {
         const wcstring uname = str2wcstring(userinfo.pw_name);
         vars.set_one(L"USER", ENV_GLOBAL | ENV_EXPORT, uname);
-        if (userinfo.pw_dir) {
-            const wcstring dir = str2wcstring(userinfo.pw_dir);
-            vars.set_one(L"HOME", ENV_GLOBAL | ENV_EXPORT, dir);
-        } else {
-            // We cannot get $HOME. This triggers warnings for history and config.fish already,
-            // so it isn't necessary to warn here as well.
-            vars.set_empty(L"HOME", ENV_GLOBAL | ENV_EXPORT);
+        // Only change $HOME if it's empty, so we allow e.g. `HOME=(mktemp -d)`.
+        // This is okay with common `su` and `sudo` because they set $HOME.
+        if (vars.get(L"HOME").missing_or_empty()) {
+            if (userinfo.pw_dir) {
+                const wcstring dir = str2wcstring(userinfo.pw_dir);
+                vars.set_one(L"HOME", ENV_GLOBAL | ENV_EXPORT, dir);
+            } else {
+                // We cannot get $HOME. This triggers warnings for history and config.fish already,
+                // so it isn't necessary to warn here as well.
+                vars.set_empty(L"HOME", ENV_GLOBAL | ENV_EXPORT);
+            }
         }
-    } else {
+    } else if (vars.get(L"HOME").missing_or_empty()) {
         // If $USER is empty as well (which we tried to set above), we can't get $HOME.
         vars.set_empty(L"HOME", ENV_GLOBAL | ENV_EXPORT);
     }
--- a/src/env.cpp
+++ b/src/env.cpp
@@ -198,18 +198,54 @@ wcstring_list_t null_environment_t::get_names(int flags) const {
     return {};
 }
 
-/// Set up the USER variable.
-static void setup_user(bool force) {
+/// Set up the USER and HOME variable.
+static void setup_user() {
     auto &vars = env_stack_t::globals();
-    if (force || vars.get(L"USER").missing_or_empty()) {
-        struct passwd userinfo;
-        struct passwd *result;
-        char buf[8192];
-        int retval = getpwuid_r(getuid(), &userinfo, buf, sizeof(buf), &result);
+    auto uid = getuid();
+    auto user_var = vars.get(L"USER");
+    struct passwd userinfo;
+    struct passwd *result;
+    char buf[8192];
+
+    // If we have a $USER, we try to get the passwd entry for the name.
+    // If that has the same UID that we use, we assume the data is correct.
+    if (!user_var.missing_or_empty()) {
+        std::string unam_narrow = wcs2string(user_var->as_string());
+        int retval = getpwnam_r(unam_narrow.c_str(), &userinfo, buf, sizeof(buf), &result);
         if (!retval && result) {
-            const wcstring uname = str2wcstring(userinfo.pw_name);
-            vars.set_one(L"USER", ENV_GLOBAL | ENV_EXPORT, uname);
+            if (result->pw_uid == uid) {
+                // The uid matches but we still might need to set $HOME.
+                if (vars.get(L"HOME").missing_or_empty()) {
+                    if (userinfo.pw_dir) {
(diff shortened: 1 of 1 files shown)


src/env.cpp
```diff
@@ -198,18 +198,54 @@ wcstring_list_t null_environment_t::get_names(int flags) const {
...
+                    }
+                }
+                return;
+            }
+        }
+    }
+
+    // Either we didn't have a $USER or it had a different uid.
+    // We need to get the data *again* via the uid.
+    // TODO: Do we need euid() here instead?
```
septatrix, line 233: Bash and Zsh actually do quite a lot if euid and uid mismatch either by restoring euid to uid or dropping into a so called privileged mode (which to my knowledge fish does not have).

src/env.cpp
```diff
@@ -295,46 +331,7 @@ void env_init(const struct config_paths_t *paths, bool do_uvars, bool default_pa
     // TODO: Figure out if that su actually checks if username == "root"(as the man page says) or
     // UID == 0.
```
septatrix, line 332: This todo should not be needed anymore (or at least moved?)

src/env.cpp
```diff
@@ -198,18 +198,54 @@ wcstring_list_t null_environment_t::get_names(int flags) const {
...
+            const wcstring dir = str2wcstring(userinfo.pw_dir);
+            vars.set_one(L"HOME", ENV_GLOBAL | ENV_EXPORT, dir);
+        } else {
+            // We cannot get $HOME. This triggers warnings for history and config.fish already,
+            // so it isn't necessary to warn here as well.
+            vars.set_empty(L"HOME", ENV_GLOBAL | ENV_EXPORT);
         }
+    } else {
+        // If $USER is empty as well (which we tried to set above), we can't get $HOME.
+        vars.set_empty(L"HOME", ENV_GLOBAL | ENV_EXPORT);
```
septatrix, line 248: Why are you purposefully clearing it here?

src/env.cpp
```diff
@@ -198,18 +198,54 @@ wcstring_list_t null_environment_t::get_names(int flags) const {
...
+    // Either we didn't have a $USER or it had a different uid.
+    // We need to get the data *again* via the uid.
+    // TODO: Do we need euid() here instead?
+    int retval = getpwuid_r(uid, &userinfo, buf, sizeof(buf), &result);
+    if (!retval && result) {
+        const wcstring uname = str2wcstring(userinfo.pw_name);
+        vars.set_one(L"USER", ENV_GLOBAL | ENV_EXPORT, uname);
+        if (userinfo.pw_dir) {
+            const wcstring dir = str2wcstring(userinfo.pw_dir);
+            vars.set_one(L"HOME", ENV_GLOBAL | ENV_EXPORT, dir);
```
septatrix, line 240: Here I think the previous behaviour of only setting HOME when it is empty should be preserved. This is also not required as `su` (and others like `sudo`, `doas` etc.) always update the HOME even when `--login` is not passed.

septatrix (changes requested):
I still think that implementing this as a fish variable (e.g. `fish_prompt_name`) is preferable and more in line with (what user came to expect from) other shells. Setting USER and HOME when empty is fine however.

Modifying the users environment comes at the risk of doing something which the user did not intend. For example now `LOGNAME` and `USER` (the former of which is actually required by posix whereas the latter apparently is not) can become out if sync. Setting a fish variable instead provides users with the same behaviour regarding the environment variables they came to expect from bash/zsh et al. while providing a clear source for the name to display. This way one might also get rid of the [technique proposed by faho](https://github.com/fish-shell/fish-shell/issues/8583#issuecomment-1100154564) which while fantastic still suffers from problems like duplicate usernames.

src/env.cpp
faho: I don't think we're interested in any of that.

faho:
>Modifying the users environment comes at the risk of doing something which the user did not intend. For example now LOGNAME and USER (the former of which is actually required by posix whereas the latter apparently is not) can become out if sync. 

Again: $USER is what people will actually use. And it is what we already use, and what we document people use.

We aren't posix-compatible, so it's *fine* to simply define $USER as the thing to use.

>I still think that implementing this as a fish variable (e.g. fish_prompt_name) is preferable and more in line with (what user came to expect from) other shells. 

Fish is not other shells, and calling this something else just means you have now made the wrong way and the right way. (what is $USER even *for* if it's not corrected?)

Instead let's simply make the one right way to get the user.

septatrix:
Well my underlying problem with this is that modern `su` implementations made the *conscious choice* to not set `USER` and `LOGNAME` for example if a script wants to lower privileges from root back to the UID of the invoking user. This is already broken by the current implementation making this use case impossible. And for chroot it is the same story.
Setting up anything besides `SHELL` and filling in empty variables is not something the shell should about and instead leave it to the login manager. However seeing as we both are very reluctant to change our stance this is something @ridiculousfish should probably finally decide...

faho:
>However seeing as we both are very reluctant to change our stance this is something @ridiculousfish should probably finally decide...

Sorry, I am entirely capable of making decisions here. This is not how any of this works.

septatrix:
> > However seeing as we both are very reluctant to change our stance this is something @ridiculousfish should probably finally decide...
> 
> Sorry, I am entirely capable of making decisions here. This is not how any of this works.

Given that you made a PR for this I figured that you wanted feedback from some of your colleagues and I do not think we will come to an agreement I do not see a point in us repeating our arguments ad infinitum and simply see what your colleagues think.

> > modern su implementations made the conscious choice
> 
> That's quite the assumption. The way I know these things, modern su implementations just do the same thing that su implementations have always done.

Yes they reimplement the behaviour from previous versions instead of changing it to preserve backwards compatibility and here we are talking about a behavior which would nullifylies their effort. I also [asked this in the util-linux repo](https://github.com/util-linux/util-linux/discussions/1646#discussioncomment-2542617) and he confirmed by gist about providing information about the invoking user and performing minimal environment changes unless the login flag is passed.

> > if a script wants to lower privileges from root back to the UID of the invoking user.
> 
> That's both not a common use case and doable in other ways.

It is a more common use case that starting a fish shell inside a chroot and no there is no possible way to know the invoking user if fish overrides that variable. Well actually there is LOGNAME but only because fish does not keep it in sync with USER.

> This might be a case for a "$ORIGINAL_USER" variable.

Then you would implement a workaround variable for a workaround and everyone would have to special case if there was a fish shell between su and the program changing USER and setting ORIGINAL_USER

faho:
>Given that you made a PR for this I figured that you wanted feedback from some of your colleagues and I do not think we will come to an agreement I do not see a point in us repeating our arguments ad infinitum and simply see what your colleagues think.

Yes, and they will give their feedback if they have any, once they have the time.

There is no need to @ specific people, and this specific example comes across a bit like you are asking for my manager to overrule me.

----

>Then you would implement a workaround variable for a workaround

I do not believe it is a "workaround" to have someone asking for specifically the unusual "original user" as opposed to the more typical "current user" to have to specify that.

> performing minimal environment changes unless the login flag is passed.

Yes, and you have been arguing that "There is no reason to use with su without --login/-l/-".

With "--login" it changes $USER.

(tbh I would naively expect it to be the *other* way around - $LOGNAME, i.e. "login name" should only be changed with "--login". Oh well)

>have to special case if there was a fish shell between su and the program changing USER

None of this is standardized or guaranteed. If you are relying on $USER here you are making unportable and easily breakable assumptions already.

septatrix:
> > have to special case if there was a fish shell between su and the program changing USER
> 
> None of this is standardized or guaranteed. If you are relying on $USER here you are making unportable and brittle assumptions already. Relying on $USER to give you the "original" user can very easily be circumvented already.

Yes but if I ship installation instructions with my software and tell the to `su` to a different user and call some scripts I can very well expect the same behaviour for all shells but fish.

> What if I set $USER to someone else before starting your script? Will it then become that other user instead of me? That doesn't sound like a good idea!

Which is why my proposal would be to introduce a fish variable (which could be made read-only to prevent such tampering) which always returns the current `pw_name`. Also the `EUID` variable would be more suited for a fish variable instead of an environment variable.

This way fish could always show the correct user in the prompt and properly check if a red root prompt should be displayed (if caching is not required it would even be possible to update the username direct in the prompt without a religion). User and developers could then be sure that fish does not touch their environment and everything is as one is used from other shells.

faho:
Okay, over two weeks with no comment from anyone else. That means I'm making an executive decision and merging this.

If we get any reports of problems, we can revisit this. I don't foresee any because we've already been doing this in the most common case (when uid is 0).

Issue #741, opened 2013-05-12
`funcrm` and other thoughts on `funced` and `funcsave`


dag:
There doesn't seem to be a way to undo `funcsave` without manually deleting the file.  In fact, `functions -e` doesn't even work in the active session because the function will be loaded again.

In #735 @aeosynth suggested `funcrm`.  I'm also wondering if perhaps we could collapse these three into a single function while also adding support for doing `funced; and funcsave` in one go by passing the right arguments.

krader1961:
The current situation is inconsistent and surprising to a naive user.  I'm not a fan of adding a `funcrm` command. I do think that `functions --erase` should tell the user if the function was auto loaded. We might want to mark the function as no longer being auto loadable. I don't think it should actually remove the file because a) that might not be possible, and b) even if we did remove it the script might have defined other functions other than the one named by the script. The documentation should be clarified that erasing a function is only useful if the function was defined interactively.

mqudsi:
Just copying my comment from #4743 here.

>I've been bothered by the lack of symmetry with funcsave vs functions --erase and my natural instinct is to func<TAB> to remind myself whether I'm looking for funcdel or funcrm, only to remember I need to use functions --erase.

>I think the discoverability here is poor and would suggest adding one (or both) of the funcsave/funcrm aliases for functions --erase. Or maybe just funcerase. (Actually, now it occurs to me that perhaps funcunload is a more apt name for what functions --erase does?)

>(Of course there is still a symmetry mismatch here, since funcsave actually writes out the function to the disk while functions --erase merely unloads it from the function map for the current session, but I feel like it's at least an improvement? Perhaps it could emit a "Use func{del,rm,erase} -p/--persistent to actually delete the function from your user directory" message (maybe just the first time it is used) or something?)

PR #8886, opened 2022-04-19
Create funcdel.fish

chapmanjacobd:
## Description

There does not seem to be a built-in for permanently deleting or renaming user functions. I've been using a variant of this successfully for a few years

Fixes issue #

## TODOs:

### DIFF
--- a/share/completions/funcdel.fish
+++ b/share/completions/funcdel.fish
@@ -1,0 +1,2 @@
+complete -c funcdel -s d -l directory -d "dir to remove function(s) from" -a '$fish_function_path' -r
+complete -c funcdel -s q -d "suppress output" -r
--- a/share/completions/wg-quick.fish
+++ b/share/completions/wg-quick.fish
@@ -1,0 +1,12 @@
+set -l valid_subcmds up down strip save
+
+function __fish_wg_complete_interfaces
+    LC_ALL=C wg show 2>| string replace -rf 'Unable to access interface (.*): Operation not permitted' '$1'
+end
+
+complete -c wg-quick -f
+complete -c wg-quick -n "__fish_seen_subcommand_from $valid_subcmds" -a '(__fish_wg_complete_interfaces)'
+complete -c wg-quick -n "not __fish_seen_subcommand_from $valid_subcmds" -a up -d 'Add and set up an interface'
+complete -c wg-quick -n "not __fish_seen_subcommand_from $valid_subcmds" -a down -d 'Tear down and remove an interface'
+complete -c wg-quick -n "not __fish_seen_subcommand_from $valid_subcmds" -a strip -d 'Output config for use with wg'
+complete -c wg-quick -n "not __fish_seen_subcommand_from $valid_subcmds" -a save -d 'Saves the configuration of an existing interface'
--- a/share/functions/funcdel.fish
+++ b/share/functions/funcdel.fish
@@ -11,7 +11,7 @@ function funcdel --description "Remove the current definition of all specified f
     if set -q _flag_directory
         set funcdir $_flag_directory
     else
-        set funcdir $fish_function_path
+        set funcdir $__fish_config_dir/functions
     end
 
     if not set -q argv[1]
@@ -32,7 +32,6 @@ function funcdel --description "Remove the current definition of all specified f
             printf (_ "%s: not defined as a function '%s'\n") funcdel $funcname >&2
             set retval 1
         else if not test -e $funcpath
-            echo "$funcname does not exist as a user function"
             printf (_ "%s: not a saved user function '%s'\n") funcdel $funcname >&2
             set retval 2
         else
--- a/share/functions/funcdel.fish
+++ b/share/functions/funcdel.fish
@@ -1,18 +1,46 @@
-function funcdel -a funcname
+function funcdel --description "Remove the current definition of all specified functions from session and file"
+    set -l options q/quiet h/help d/directory=
+    argparse -n funcdel $options -- $argv
+    or return
 
-    set funcpath $fish_function_path/$funcname.fish
+    if set -q _flag_help
+        __fish_print_help funcdel
+        return 0
+    end
 
-    if not type -q $funcname
-        echo "$funcname is not defined as a function"
+    if set -q _flag_directory
+        set funcdir $_flag_directory
+    else
+        set funcdir $fish_function_path
+    end
+
+    if not set -q argv[1]
+        printf (_ "%ls: Expected at least %d args, got only %d\n") funcdel 1 0 >&2
         return 1
     end
 
-    if not test -e $funcpath
-        echo "$funcname does not exist as a user function"
-        return 2
+    if not test -d $funcdir
+        printf (_ "%s: Configuration directory does not exist '%s'\n") funcdel $funcdir >&2
+        return 1
     end
 
-    functions -e $funcname
-    rm $funcpath
+    set -l retval 0
+    for funcname in $argv
+        set -l funcpath "$funcdir/$funcname.fish"
 
+        if not functions -q -- $funcname
+            printf (_ "%s: not defined as a function '%s'\n") funcdel $funcname >&2
+            set retval 1
+        else if not test -e $funcpath
+            echo "$funcname does not exist as a user function"
+            printf (_ "%s: not a saved user function '%s'\n") funcdel $funcname >&2
+            set retval 2
+        else
+            functions --erase -- $funcname
+            rm $funcpath
+            and set -q _flag_quiet || printf (_ "%s: removed %s\n") funcdel $funcpath
+        end
+    end
+
+    return $retval
 end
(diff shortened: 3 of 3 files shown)


faho:
Okay, #741 asked for this previously.

Personally, I'm not a massive fan because I don't think removing the file is a super common task, but maybe I'm just using things differently.

However, if we *are* to do this, it would require some more work.

1. It should probably use $fish_function_path.
2. It should see if the file exists and print an error if it doesn't - e.g. if the function wasn't defined in a normal autoload file
3. It should ask for confirmation

And of course all the amenities - docs, tests and so on.

Like I said I'm not in love with this and the reaction to #741 was lukewarm at best. So, unless anyone else champions this in, say, a week, I am planning to reject this (and close #741).

chapmanjacobd:
> It should ask for confirmation

I think it should only do this explicitly if the function path does not match the name of the function (but probably this is impossible to detect?)

Another idea would be to have a folder for deleted functions and fish functions are moved to there instead of deleted.

Another idea is that the file is printed to stdout when it is deleted...

I just think it's a little strange to require confirmation. I'm not aware of any fish built-in or default function that requires confirmation

zanchey:
The reason I think it's worth including functions like this is that it helps to make the interface more seamless - you don't need to know how function saving is implemented to be able to remove a function.

However, it's imperfect and potentially explosive. For example, `funcdel fish_prompt` without a custom user prompt defined produces an error in the current implementation (`rm` tries to remove a nonexistent argument); if extended to use `fish_function_path`, it will either fail with permissions errors or delete a part of the installed fish!

So while I appreciate the improved unity of interface created by this function, I'm not sure it stacks up. On the other hand, the current method of removing a saved function is not so hard to understand that there are lots of people filing issues or questions about it.

I'd come down towards closing #741 as well, but I appreciate it's tricky.

chapmanjacobd:
> helps to make the interface more seamless - you don't need to know how function saving is implemented to be able to remove a function

yes I completely agree

I agree that the value which a funcdel utility adds is very minimal but without it the tools funced and funcsave feel a tiny bit incomplete. For me the scope of this utility would only be to supplement the existing funced/funcsave functions. 

If funcsave can save a function then fundel should be able to delete it. In the `funcdel fish_prompt` case I think returning an error makes sense. funcsave does not overwrite the default fish_prompt so it doesn't make sense to delete the default or the system fish_prompt. funcsave only works with the `$fish_function_path` right? So funcdel would be limited to only undo funcsave operations. Within the context of funcsave and funced I imagine that completely covers the scope that a user could want ?

chapmanjacobd:
> If you have erased a function using :ref:`functions <cmd-functions>`'s ``--erase`` option, ``funcsave`` will remove the saved function definition.
> funcsave.rst

This does not seem to be true?

```
$ type t 
t is a function with definition
# Defined in /home/xk/.config/fish/functions/t.fish @ line 2
function t
	echo hhhh
end
$ functions -e t 
$ type t         
type: Could not find 't'
[1] $ ls /home/xk/.config/fish/functions/t.fish                           
Permissions Size User Date Modified Name
.rw-r--r--    49 xk   23 Apr 22:28  /home/xk/.config/fish/functions/t.fish
$ funcsave t                                                                                       
funcsave: Unknown function “t”
[1] $ ls /home/xk/.config/fish/functions/t.fish                              
Permissions Size User Date Modified Name
.rw-r--r--    49 xk   23 Apr 22:28  /home/xk/.config/fish/functions/t.fish
$ cat /home/xk/.config/fish/functions/t.fish              
# Defined interactively
function t
    echo hhhh
end
```

chapmanjacobd:
```
echo $fish_function_path
/home/xk/.config/fish/functions /home/xk/.local/share/omf/pkg/omf/functions/compat /home/xk/.local/share/omf/pkg/omf/functions/core /home/xk/.local/share/omf/pkg/omf/functions/index /home/xk/.local/share/omf/pkg/omf/functions/packages /home/xk/.local/share/omf/pkg/omf/functions/themes /home/xk/.local/share/omf/pkg/omf/functions/bundle /home/xk/.local/share/omf/pkg/omf/functions/util /home/xk/.local/share/omf/pkg/omf/functions/repo /home/xk/.local/share/omf/pkg/omf/functions/cli /home/xk/.local/share/omf/pkg/bass/functions /home/xk/.local/share/omf/pkg/fish-spec/functions /home/xk/.local/share/omf/pkg/fzf/functions /home/xk/.local/share/omf/pkg/omf/functions /home/xk/.local/share/omf/pkg/z/functions /home/xk/.local/share/omf/lib /home/xk/.local/share/omf/lib/git /home/xk/.local/share/omf/themes/default /home/xk/.local/share/omf/themes/default/functions /etc/fish/functions /home/xk/.local/share/flatpak/exports/share/fish/vendor_functions.d /var/lib/flatpak/exports/share/fish/vendor_functions.d /usr/local/share/fish/vendor_functions.d /usr/share/fish/vendor_functions.d /usr/share/fish/functions
```

Until `funcsave` supports multiple directories it doesn't make sense for `funcdel` to. I'm going to use the same path that `funcsave` does

chapmanjacobd:
```
~/g/o/fish-shell (patch-1|✚1) 🌌 funcdel t
funcdel: removed /home/xk/.config/fish/functions/t.fish

~/g/o/fish-shell (patch-1|✚1) 🌭 ls /home/xk/.config/fish/functions/t.fish
"/home/xk/.config/fish/functions/t.fish": No such file or directory (os error 2)
/usr/bin/ls: cannot access '/home/xk/.config/fish/functions/t.fish': No such file or directory
```

works... and the universe-hotdog combo of my ~~$PS1 randomizer~~ is a good omen

mqudsi:
How does this handle a case where a .fish file has two function declarations and you call `funcdel` against one of them?

chapmanjacobd:
Good question. `funcdel` is complementary to `funced/funcsave` so we should expect them all to have similar uses and edge-case handling. I don't always use `funced/funcsave` to permanently edit fish functions and I don't always use `funcdel` to permanently delete functions. For funcsave keeps one fish function per file--but what if you call `funcsave` on a fish file that has two function declarations? I have not tried that yet. I'll do a quick experiment without reading the `funcsave` code:

```fish
$ tee ~/.config/fish/functions/t1.fish
function t1_inner
  echo "hello"
end
function t1
  t1_inner
end

$ type t1
t1 is a function with definition
# Defined in /home/xk/.config/fish/functions/t1.fish @ line 4
function t1
  t1_inner
end

$ funced t1
Warning: the file containing this function has not been saved. Changes may be lost when fish is closed.
Save function to /home/xk/.config/fish/functions/t1.fish? [Y/n] y
Function saved to /home/xk/.config/fish/functions/t1.fish
$ cat ~/.config/fish/functions/t1.fish
function t1_inner
  echo "hello?"
end
function t1
  t1_inner
end
$ funcsave t1
$ cat ~/.config/fish/functions/t1.fish
# Defined in /home/xk/.config/fish/functions/t1.fish @ line 4
function t1
  t1_inner
end
```

hmm there is some strange behavior here... If the function is edited in the $EDITOR then we get a prompt about saving it and it remains but if we call `funcsave` on it then the extraneous function definition in the t1.fish file is stripped away.

I'm not sure what the desired behavior would be. It could go either way. I think if you define multiple functions per file you wouldn't want to use `funcsave` or `funcdel` to manage that file.

chapmanjacobd:
but as far as `funcdel` goes it is only concerned about the function with the same name as the file--which should always be the case if it was saved with `funcsave`

```fish
$ funcdel t1
funcdel: removed /home/xk/.config/fish/functions/t1.fish
$ funcdel t1_inner
funcdel: not a saved user function 't1_inner'
[2]$ t1_inner 
hello?
```

mqudsi:
Since funcdel doesn't know if a function was created with funcsave (and then never modified), I am against any feature that would result in something that could potentially destroy more than what you specified. 

I've already reported that issue with `funcsave` before and I consider its current behavior to be buggy. I am strongly against anything that further expands the potential for data loss.

chapmanjacobd:
we could add something like this to `funcsave`:

```fish
if test (grep -i function $functionfile | count) -gt 1
  confirm "$functionfile contains more than one function definition. Overwrite? (y / N) "
end
```

and `funcdel`:

```fish
if test (grep -i function $functionfile | count) -gt 1
  confirm "$functionfile contains more than one function definition. Continue deleting file? (y / N) "
end
```

but maybe someone else has a more complete solution for this ?

mqudsi:
I would prefer we leverage the existing parser or ast support to actually determine how many functions are in the file. Naively counting the occurrences of the word "function" is too error prone. Fish supports nested declaration of functions, but they're not private outside their declaration site - not sure if something like this should be considered two functions or one, but I'd be fine with deleting both if `funcdel outer` were called.

```fish
function outer
    function inner
        echo "inner executed"
    end

    echo "executing inner"
    inner
end
```

faho:
Okay, I'm terribly sorry about the delay on this.

I believe this lacks a champion to take it over the finish line, and consensus on what exactly it should do.

So, for those reasons, I think the most honest thing we can do is to close it.

I thank you for your work, and who knows, maybe this'll be picked up again for another attempt?

Issue #8895, opened 2022-04-21
rsync space escaping no longer works since version 3.2.4


farid-fari:
I am using fish version 3.4.1 and rsync version 3.2.4.

If you do `rsync -Pauhr server:begin` then press tab, and that server contains a file that begins with "begin" and contains spaces, for example `"begin file.txt"` then the auto-complete which fish does is `rsync -Pauhr server:begin\\\ file.txt`.

This used to work up to rsync 3.2.3 but with 3.2.4 spaces should only be single escapes. The completed command will fail, and what needs to be executed instead is `rsync -Pauhr server:begin\ file.txt`. I do not know the precise reasons and rsync escaping seems complicated, but the very least I can tell is that the current auto-completion does not work.

Another option for auto-complete would be to not escape spaces but add quotes: `rsync -Pauhr server:"begin file.txt"`.

farid-fari:
See [here](https://download.samba.org/pub/rsync/NEWS#3.2.4) for more info (first point).

PR #8898, opened 2022-04-22
Fixed escaping in rsync completion

farid-fari:
## Description

Removing the escape fixes the escaping problem in rsync (no double escaping).

Fixes issue #8895

## TODOs:

### DIFF
--- a/share/completions/rsync.fish
+++ b/share/completions/rsync.fish
@@ -183,7 +183,6 @@ complete -c rsync -d "Remote path" -n "commandline -ct | string match -q '*:*'" 
 )(
 	# Get the list of remote files from the specified rsync server.
         rsync --list-only (__rsync_remote_target) 2>/dev/null | string replace -r '^d.*' '\$0/' |
-        string replace -r '(\S+\s+){4}' '' | # drop the first four columns
-        string escape -n
+        string replace -r '(\S+\s+){4}' '' # drop the first four columns
 )
 "


share/completions/rsync.fish
```diff
@@ -183,7 +183,6 @@ complete -c rsync -d "Remote path" -n "commandline -ct | string match -q '*:*'"
 )(
 	# Get the list of remote files from the specified rsync server.
         rsync --list-only (__rsync_remote_target) 2>/dev/null | string replace -r '^d.*' '\$0/' |
-        string replace -r '(\S+\s+){4}' '' | # drop the first four columns
-        string escape -n
+        string replace -r '(\S+\s+){4}' '' # drop the first four columns
```
faho, line 186: This breaks it for rsync < 3.2.4.

Unfortunately, long term support distributions exist so this needs to have a version check.

faho (changes requested):

share/completions/rsync.fish
faho: Worse than that, I'm pretty sure we'd still have to escape `$` and `*` and such - what the rsync news refers to as "shell-active" characters.

I can't find an enumeration of those in the rsync docs.

faho: Ah, okay, from https://download.samba.org/pub/rsync/rsync.1#opt--old-args:

>The default in a modern rsync is for "shell-active" characters (including spaces) to be backslash-escaped in the args that are sent to the remote shell. The wildcard characters *, ?, [, & ] are not escaped in filename args (allowing them to expand into multiple filenames) while being protected in option args

So we need to still escape `*`, `?`, `[`, `]`, because we want to pass *one* filename here. I believe we might also need to escape `\` again.

farid-fari: I certainly don't know enough about fish to do all of this... I've just tried a file containing `[` and `]` and it seems to work correctly with my version, though there's a weird anomaly where the completion list contains two versions of the file (one single-escaped and one triple-escaped, the former being correct).

Same thing for `?` and `*` actually.

faho: Okay. Create, on the remote machine, these three files:

- /tmp/a1b
- /tmp/a2b
- /tmp/a?b

Now try to copy `/tmp/a?b` to the local machine, using your updated completions to select it. Does it also create a1b and a2b?

farid-fari: Oh yeah it does. But triple-escaping does not work either: `rsync server:a\\\?b .` says the file doesn't exist, I think because the server sees a slash.

farid-fari: I should also note this is not a regression from my change, the bad behaviour for your example already exists with the current state of the code and rsync 3.2.3. There are two completions suggested, one will copy all of the files and the other will just copy the correct file.

faho: >But triple-escaping does not work either: rsync server:a\\\?b . says the file doesn't exist, I think because the server sees a slash.

Ooof.

What about quoted:

```
rsync server:"'a?b'" .
```

?

farid-fari: With 3.2.4 double quoted does not work

faho: As I understand it, it introduces a rather large regression for older versions.

We need the version check. Lemme see if I can whip something up.

farid-fari: Right, I meant with the version check, sorry. I've seen `systemctl.fish` also has a version check but it's much simpler in that case since their version numbers are valid floats. I guess it would take a bit of string matching to compare these version numbers.

faho: Okay, this is gonna be awkward. Because rsync's version output is up there among the more annoying.

The output looks like

```
rsync  version v3.2.4  protocol version 31
Copyright (C) 1996-2022 by Andrew Tridgell, Wayne Davison, and others.
Web site: https://rsync.samba.org/
Capabilities:
    64-bit files, 64-bit inums, 64-bit timestamps, 64-bit long ints,
    socketpairs, symlinks, symtimes, hardlinks, hardlink-specials,
    hardlink-symlinks, IPv6, atimes, batchfiles, inplace, append, ACLs,
    xattrs, optional protect-args, iconv, prealloc, stop-at, no crtimes
Optimizations:
    SIMD-roll, no asm-roll, openssl-crypto, no asm-MD5
Checksum list:
    xxh128 xxh3 xxh64 (xxhash) md5 md4 none
Compress list:
    zstd lz4 zlibx zlib none

rsync comes with ABSOLUTELY NO WARRANTY.  This is free software, and you
are welcome to redistribute it under certain conditions.  See the GNU
General Public Licence for details.
```

So the actual version number we're interested in is on the first line. There are apparently two spaces between "rsync" and "version". On this system the version starts with a "v", but I can't find that upstream (printed [here](https://github.com/WayneD/rsync/blob/0ac7ebceef70417355f25daf9e2fd94e84c49749/usage.c#L183-L191), version number is stored [here](https://github.com/WayneD/rsync/blob/0ac7ebceef70417355f25daf9e2fd94e84c49749/version.h#L1). Apparently archlinux [builds from a git tag](https://github.com/archlinux/svntogit-packages/blob/packages/rsync/trunk/PKGBUILD), and that causes the "v" prefix.

So, we would extract the number with something like

```fish
rsync --version | string replace -rf '^rsync +version\D+([\d.]+) .*' '$1'
```

match a line starting with "rsync", at least one space, the word "version", something that's not a digit and then a run of digits and dots, followed by whitespace and whatever. Extract the digits and dots.

We would then split that on the dots and store it in a variable, and then we'd test if it's larger than 3.2.3.

Together it would have to look something like

```fish
set -l new_escaping # has an element if the new escaping style introduced in 3.2.4 is required

set -l rsync_ver (rsync --version | 
                  string replace -rf '^rsync +version\D+([\d.]+) .*' '$1' |
                  string split .)

if test "$rsync_ver[1]" -gt 3 2>/dev/null
    or test "$rsync_ver[1]" -eq 3 -a "$rsync_ver[2]" -gt 2 2>/dev/null
    or test "$rsync_ver[1]" -eq 3 -a "$rsync_ver[2]" -eq 2 -a "$rsync_ver[3]" -gt 3 2>/dev/null
    set new_escaping 1
end
```

And then, in the bit here, you'd simply only add the `| string escape -n` if new_escaping doesn't have an element:

```fish
        # drop the first four columns
        string replace -r '(\S+\s+){4}' '' $(set -q new_escaping[1]; or echo ' | string escape -n')
```

mqudsi: FWIW if you execute the invalid `rsync -V`, you'll get a possibly easier to parse output:

```
> rsync -V
rsync: -V: unknown option
rsync error: syntax or usage error (code 1) at main.c(1585) [client=3.1.2]
```

:shrug:

faho: >FWIW if you execute the invalid rsync -V, you'll get a possibly easier to parse output:

Until it starts being valid in some future version.

faho:
Merged, version check added in 3e0bd0b7e13b6a47f7c0df19ca3792ac1d44e288

Issue #8624, opened 2022-01-09
Empty case is valid but unreachable


zanchey:
The documentation says this is valid syntax, and indeed it is:

```fish
switch $foo
    case
        echo Hello
end
```

However, the case without an argument is unreachable. A parameter to `switch` that expands to nothing is equivalent to the empty string (per https://github.com/fish-shell/fish-shell/blob/master/src/parse_execution.cpp#L534-L539).

Should `case` without an argument be a default (equivalent to `case '*'`) or an error? Or should it be left alone?

faho:
I'd probably make a literal empty `case` error out and leave the commandline incomplete. As it stands you can *type* the example interactively, and it looks like it works.

Now, `case` with an empty expansion is another case (hahaha), but there it's probably expected that it won't match. E.g.

```fish
set -l cases # empty

switch $foo
    case foo
        echo foo
    case $cases
        # unreachable!
end
```

This seems to be expected and useful (because the variable also *could* be non-empty), while a literal empty `case` is just nothing.

This follows `switch`, which also needs an argument - it's a syntax error otherwise!

PR #8902, opened 2022-04-22
fix issue#8624 empty case

pltan:
## Empty case is valid but unreachable #8624

We added a condition to check whether the argument is empty. If the argument is empty, then we attach “*” to the case’s argument to convert it into a default argument. By doing so, the case would be reachable.

Fixes issue #8624 

## TODOs:

### DIFF
--- a/src/parse_execution.cpp
+++ b/src/parse_execution.cpp
@@ -561,6 +561,10 @@ end_execution_reason_t parse_execution_context_t::run_switch_statement(
             for (const wcstring &arg : case_args) {
                 // Unescape wildcards so they can be expanded again.
                 wcstring unescaped_arg = parse_util_unescape_wildcards(arg);
+                if (arg.empty()) {
+                    wcstring empty_arg = std::to_wstring('*');
+                    wcstring unescaped_arg = parse_util_unescape_wildcards(empty_arg);
+                }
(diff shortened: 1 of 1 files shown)


faho:
We've mentioned a few possible meanings for an empty case:

1. `case` is the same as `case '*'` and matches everything
2. `case $empty` where the variable is empty matches nothing, literal `case` is an error

You've picked the former. That's a possibility. I'm not sure I'm sold on it, and it requires further discussion/defence vs the other possibility (and possibly anything we hadn't thought of yet).

One issue here is that this runs on the *already expanded* `case`, so:

```fish
switch whatever
    case $var
       echo matched via var
    case foo
       echo matched via foo
    case
       echo matched via empty case # could also be '*'
end
```

would end up always matching the first `case` if the variable is empty. This seems surprising and I'm wary of introducing it. It reminds me of how 0 just blows up all division and you always need to remember it. I don't believe anyone would use an empty variable like that on purpose.

This could *possibly* be solved by only matching everything for a *literal* `case`, and treating expansions that happen to come up empty like today and *not* matching anything, but that might also be hugely surprising.

In terms of the specific code here, there's not actually any need to match the `*` at all - it matches everything.

So it could be simplified to 

```c++
bool match = arg.empty() || wildcard_match(switch_value_expanded, unescaped_arg);
```

So because this needs further design, I'm closing this. Sorry!

Issue #8743, opened 2022-02-23
Killring should be a modifiable fish variable


septatrix:
> The killring should probably just be a fish variable.

_Originally posted by @ridiculousfish in https://github.com/fish-shell/fish-shell/issues/7445#issuecomment-1046139653_

This enables users to e.g. remove entries from the killring if they so desire

chapmanjacobd:
If nothing is selected (I think only used in `vi` mode?) `kill-selection` can effectively clear the killring which can be convenient.

PR #8903, opened 2022-04-22
Issue#8743

pltan:
## Issue#8743

Add a function under kill.h to support removing an existed element in fish_killring.

Fixes issue #8743 

## TODOs:

### DIFF
--- a/src/kill.cpp
+++ b/src/kill.cpp
@@ -59,3 +59,10 @@ wcstring_list_t kill_entries() {
     auto kill_list = s_kill_list.acquire();
     return wcstring_list_t{kill_list->begin(), kill_list->end()};
 }
+
+void kill_remove(wcstring str) {
+    auto kill_list = s_kill_list.acquire();
+    // find the str and remove it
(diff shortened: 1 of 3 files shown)


faho:
1. Please make your PRs independent of each other. This is branched from #8902 
2. Please make your titles more expressive instead of just the issue number
3. This doesn't _work_ because the function isn't *used* anywhere. #8743 calls for it to be a modifiable fish variable. So until you can do something like `set -e fish_killring[424]` and it changes the killring that's not actually fixed.

The tricky part here isn't really writing the code itself. The tricky part is in making sure it works, and works like it should. That means the person who writes it needs to figure out what that "should" is. What's the right behavior here?

In particular how do you *use* this? What do you do with it? I have to confess that's absolutely unclear to me, which is why I haven't worked on #8743.

Because this currently does not contain anything of value, I am closing this. Please do slow down, it's quality over quantity. Instead of filing two PRs in one day make a really really really good one.

pltan:
Thanks a lot for the feedback. Those are very useful insight.

PR #8904, opened 2022-04-22
feat: base completion for fastboot 

NextAlone:
## Description

feat: base completion for fastboot according to its help

## TODOs:

### DIFF
--- a/share/completions/fastboot.fish
+++ b/share/completions/fastboot.fish
@@ -1,0 +1,78 @@
+function __fish_fastboot_no_subcommand -d 'Test if fastboot has yet to be given the subcommand'
+    for i in (commandline -opc)
+        if contains -- $i flashall getvar oem flashing reboot update erase format devices flash get_staged help stage boot fetch
+            return 1
+        end
+    end
+    return 0
+end
+
+function __fish_fastboot_list_partition
+    set -l partitions boot bootloader dtbo modem odm odm_dlkm oem product pvmfw radio recovery system vbmeta vendor vendor_dlkm cache userdata system_ext
+    for i in $partitions
+        echo $i
+    end
+end
+
+function __fish_fastboot_list_partition_and_file
+    for i in (commandline -opc)
+        if contains -- (string replace -r '_[ab]' '' $i) (__fish_fastboot_list_partition)
+            __fish_complete_path
+        else
+            __fish_fastboot_list_partition
+        end
+    end
+end
+
+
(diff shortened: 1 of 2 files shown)


share/completions/fastboot.fish
```diff
@@ -0,0 +1,78 @@
...
+    end
+end
+
+function __fish_fastboot_list_partition_and_file
+    for i in (commandline -opc)
+        if contains -- (string replace -r '_[ab]' '' $i) (__fish_fastboot_list_partition)
+            __fish_complete_path
+        else
+            __fish_fastboot_list_partition
+        end
```
krobelus, line 23: You loop over `commandline -opc` which includes the command itself (`fastboot`), which means the else-branch is taken at least once.
So how about moving the else-branch outside the loop?
Also, I don't know fastboot but I'd imagine a `return` after `__fish_complete_path` could make sense because why would we need two partitions on the commandline?

share/completions/fastboot.fish
```diff
@@ -0,0 +1,78 @@
...
+function __fish_fastboot_list_partition
+    set -l partitions boot bootloader dtbo modem odm odm_dlkm oem product pvmfw radio recovery system vbmeta vendor vendor_dlkm cache userdata system_ext
+    for i in $partitions
+        echo $i
+    end
+end
+
+function __fish_fastboot_list_partition_and_file
+    for i in (commandline -opc)
+        if contains -- (string replace -r '_[ab]' '' $i) (__fish_fastboot_list_partition)
```
krobelus, line 19: I'm curious, why the `_[ab]`?

share/completions/fastboot.fish
```diff
@@ -0,0 +1,78 @@
+function __fish_fastboot_no_subcommand -d 'Test if fastboot has yet to be given the subcommand'
+    for i in (commandline -opc)
+        if contains -- $i flashall getvar oem flashing reboot update erase format devices flash get_staged help stage boot fetch
```
krobelus, line 3: I guess we could use __fish_seen_subcommand_from here

share/completions/fastboot.fish
NextAlone: In newer devices with A/B (Seamless) System Updates, that partitions are divided into foo_a and foo_b, for example, boot_a, boot_b. Then when boot_a and boot_b and boot (represent current slot) is typed, it represent the target patition.

NextAlone: image

PR #8909, opened 2022-04-24
fix: edit_command_buffer with micro without parsecursor

CosmicToast:
## Description

Micro only parses the [FILE]:LINE:COL syntax if the parsecursor option is enabled.
The +LINE:COL syntax is unambigous and always valid, and thus should be preferred.

I didn't bother opening an issue because it's a single-line fix.
The consequence of not doing this is that `edit_command_buffer` appears to do nothing when micro is used as the editor, and does not have the parsecursor option enabled (as in my case, and the default).

## TODOs:

- [x] Changes to fish usage are reflected in user documentation/manpages. *(no changes to fish usage)*
- [x] User-visible changes noted in CHANGELOG.rst *(no user-visible changes)*

### DIFF
--- a/share/functions/edit_command_buffer.fish
+++ b/share/functions/edit_command_buffer.fish
@@ -56,7 +56,7 @@ function edit_command_buffer --description 'Edit the command buffer in an extern
         case subl
             set -a editor $f:$line:$col --wait
         case micro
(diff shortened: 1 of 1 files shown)


krobelus:
Thanks!

Issue #8917, opened 2022-04-27
Extra single quotes are getting stripped in webconfig


YmirKhang:
Fish version 3.1.0
Ubuntu 20.04
xterm-256color

This issue can be reproduced regardless of third party customizations are added or not.

Whenever an abbreviation ending with a single quote is added via fish_config UI, the single quote at the end gets stripped. Even if the command is wrapped with one more layer of single quotes.  

So setting the word `_first` to `awk '{print $1'}` actually executes the following command:

```fish
abbr --add '_first' 'awk \'{print $1}'
```
 
Also setting the work `_first` with another layer of single quotes: `'awk '{print $1'}'`, results in the execution of the same command:

```fish
abbr --add '_first' 'awk \'{print $1}'
```

faho:
Note: Your reproducer has what I assume is a typo:

```fish
awk '{print $1'}
```

works - there's no `'` at the end, so there's nothing to "strip".

However I *can* still reproduce this with

```fish
awk '{print $1}'
```

which also makes much more sense.

-----

Also:

>Fish version 3.1.0

fish 3.1.0 was already outdated by the time Ubuntu 20.04 was released, and that's not even the current LTS anymore (the current LTS 22.04, incidentally also ships an already outdated fish).

It is 3930 released commits behind, out of a total 15379 released commits - that's 25% of all commits ever made to fish (and released)!

Please update if you want to report any further issues.

YmirKhang:
> Also: We do _not_ require separate issues to be filed if the PR is simple.
> 
> In future please just file the PR if you have one, otherwise it's just more clicking for me.
> 
> That being said: Thank you for your report, I believe it is fixed now!

Notes. Thanks

PR #8918, opened 2022-04-27
Update single layer stripping

YmirKhang:
## Description

Fixed words and phrases being stripped of multiple single quotes. As per the comment of the function states, only one layer of single quotas need to be stripped. The `strip` method in python actually removes as many trailing and leading characters as it can.

Fixes issue https://github.com/fish-shell/fish-shell/issues/8917

## TODOs:


faho:
Merged, thanks!

YmirKhang:
> Merged, thanks!

You're welcome

PR #8928, opened 2022-05-02
Document `string split --fields`

henrikhorluck:
## Description

Documentation was lost as part of https://github.com/fish-shell/fish-shell/commit/14d60ccb328fbff47408e914eb8cb27e887669b8

## TODOs:

- [x] Changes to fish usage are reflected in user documentation/manpages.
- [ ] Tests have been added for regressions fixed
- [ ] User-visible changes noted in CHANGELOG.rst

PR #8626, opened 2022-01-09
Command synopsis: fixes and consistency improvements

krobelus:
Some of the consistency changes may not be worth it.
The RST syntax changes might be obsolete with #8625 (depending on which direction it goes).
cc @floam

PR #8929, opened 2022-05-03
Fix more lost string documentation

henrikhorluck:
## Description

More lost documentation. From what I can understand the reason this happened in the first place is because there was [inconsistencies](https://github.com/fish-shell/fish-shell/commit/14d60ccb328fbff47408e914eb8cb27e887669b8#diff-a70e8ee95eade27579cc586715560314feaf9ed86c0783ed65921ab8df4ec842) between the command index and the individual pages — this should perhaps be fixed by de-duplicating it, or automatically generate it from code—but regardless that is outside the scope of this PR.

NOTE: before merge, there is no standard for how you specify arguments, e.g. both `[-o | --option OPTION]`, `[(-o | --option) OPTION]` and `[-o | --option] OPTION` co-exist in the docs currently[^1]. Is there a specific standard or format we should employ to be consistent? If so I can correct it so that it is consistent.

## TODOs:

[^1]: The last one perhaps erroneously introduced in #8928


zanchey:
> [-o | --option OPTION]

This is my preferred approach. It is not strictly compliant with the man(7) format, but I think it's a lot more readable.

krobelus:
Nice fixes thanks!

Previous style discussion in https://github.com/fish-shell/fish-shell/pull/8626#discussion_r780795111 which tried to standardize notation for short/long options with arguments (apparently it forgot some instance, where is that?).
The resuling commit f8e8c23ac cites Git as using the same style (though this pattern is really rare).

> `[-o | --option] OPTION`
> [^1]: The last one perhaps erroneously introduced in #8928

I don't see that anywhere. It's definitely the worst of the three options.

henrikhorluck:
> Previous style discussion in [#8626 (comment)](https://github.com/fish-shell/fish-shell/pull/8626#discussion_r780795111) which tried to standardize notation for short/long options with arguments (apparently it forgot some instance, where is that?).

e.g. https://github.com/fish-shell/fish-shell/blame/master/doc_src/cmds/string.rst#L20
 
> > `[-o | --option] OPTION`
> 
> I don't see that anywhere. It's definitely the worst of the three options.

As mentioned in this PR description footnote, that was _introduced_ by me 🤡 and was not caught in the PR review of the last documentation "fix" (should probably also been caught already in that PR that this might have happened more than one place, and done a check on that)

faho:
Okay, I've now merged this because it seems okay to me. Thanks for opening!

If there are any more issues here those can be handled in a separate PR.

Issue #8046, opened 2021-06-04
kill "not enough arguments" possible regression 


bytefluxio:
`fish --version` fish, version 3.2.2
`echo $version` 3.2.2

`uname -a` MSYS_NT-10.0-21390 DESKTOP-NE8TN2K 3.2.0-340.x86_64 2021-05-24 19:32 UTC x86_64 Msys
`echo $TERM` xterm-256color

Terminal: Windows Terminal, Version: 1.8.1521.0

Same issue with `sh -c 'env HOME=$(mktemp -d) fish'`

---

autocompletion seems to execute kill command before I'm done typing.

---

there doesn't seem to be an asciinema build compatible with msys2, so here's a screenshot:
![WindowsTerminal_43B7PhWtaH](https://user-images.githubusercontent.com/29196542/120852447-81d78800-c57a-11eb-9026-45e8c2c84021.png)

Might be a regression of #6469

faho:
It's not a regression as there haven't been any changes to the completions since that time.

It's just that msys2 uses the same crappy `kill` implementation and we only check for cygwin.

I'm assuming `uname` says `Msys`, with a capital "M"?

bytefluxio:
It does

PR #8944, opened 2022-05-09
Simplified string match.

jtza8:
My Msys fish broke, but when I wanted to write a patch I realised the
version of fish I was using was old and the problem was already fixed.

I did, however, simplify the solution and thought I'd share the patch
as a suggestion anyway.

## Description

A single string match finds all the problematic kill environments.

Re-fixes issue #8046

## TODOs:


faho:
This expression isn't entirely equivalent. It matches cyGwin and CYgWin and cygwin, it matches msys and mingw only at the start of the line.

I don't think we can merge this without checking if that actually matches what these things set, and because it's a minor cleanup I don't believe that research is worth it.

So I'm closing this, sorry!

PR #8953, opened 2022-05-13
Add missing cp completions

dseight:
## Description

All of the missing options were present in the GNU cp for more than 8 years, so it makes sense to add completions for them.

## TODOs:


database64128:
I always use `cp -a --reflink src dest` to copy src to dest. The `--reflink` option's argument is optional and defaults to `auto`. This PR changed the behavior and now Fish refuses to autocomplete src because it thinks I'm missing the argument to `--reflink`.

PR #8955, opened 2022-05-14
fix: completion `git mv` with files

NextAlone:
## Description

Talk about your changes here.

Fixes issue #

## TODOs:

### DIFF
--- a/share/completions/git.fish
+++ b/share/completions/git.fish
@@ -1663,6 +1663,7 @@ complete -c git -n '__fish_git_using_command mergetool' -s O -d 'Process files i
 
 ### mv
 complete -c git -n __fish_git_needs_command -a mv -d 'Move or rename a file, a directory, or a symlink'
+complete -f -c git -n '__fish_git_using_command mv' -a '(git ls-files)'
 complete -f -c git -n '__fish_git_using_command mv' -s f -l force -d 'Force rename/moving even if target exists'
 complete -f -c git -n '__fish_git_using_command mv' -s k -d 'Skip rename/move which can lead to error'
 complete -f -c git -n '__fish_git_using_command mv' -s n -l dry-run -d 'Only show what would happen'


faho:
I'm not sure this is correct. This would enable the normal file completions for `git mv`, but that completes more than what `git mv` operates on.

E.g. go to a git repository and run this:

```fish
touch foo
git mv foo bar
```

You'll get an error:

> fatal: not under version control, source=foo, destination=bar

And the same if "bar" is tracked.

So at the very least, the source needs to be tracked by git already.

So this would need `__fish_git_files`, presumably with the list of all files that aren't untracked - `__fish_git_files all-staged copied deleted modified renamed`?

The destination can't be an existing file at all or you get "target already exists", unless `--force` was given. So technically we could complete *all files* if `--force` was given. I think that's okay to leave.

NextAlone:
> I'm not sure this is correct. This would enable the normal file completions for `git mv`, but that completes more than what `git mv` operates on.
> 
> E.g. go to a git repository and run this:
> 
> ```fish
> touch foo
> git mv foo bar
> ```
> 
> You'll get an error:
> 
> > fatal: not under version control, source=foo, destination=bar
> 
> And the same if "bar" is tracked.
> 
> So at the very least, the source needs to be tracked by git already.
> 
> So this would need `__fish_git_files`, presumably with the list of all files that aren't untracked - `__fish_git_files all-staged copied deleted modified renamed`?
> 
> The destination can't be an existing file at all or you get "target already exists", unless `--force` was given. So technically we could complete _all files_ if `--force` was given. I think that's okay to leave.

I can't get all git files with `__fish_git_files`, with command mentioned above.

faho:
> I can't get all git files with __fish_git_files, with command mentioned above.

What's missing?

What does `git status --porcelain=2` print, in the first two columns, for the missing files?

NextAlone:
> > I can't get all git files with __fish_git_files, with command mentioned above.
> 
> What's missing?
> 
> What does `git status --porcelain=2` print, in the first two columns, for the missing files?

It only gets changed files, while files no changes won't show.

faho:
Aaah, okay, yeah, that's right.

Okay, ls-files is probably the right thing to use, then.

faho:
Okay, merged. Thanks!

NextAlone:
> Okay, merged. Thanks!

Thank you too

Issue #7658, opened 2021-01-25
Feature request: add path subcommand for string utility (`string path`)


mattmc3:
Coming from Zsh, there are a lot of built-in helpers for dealing with path strings. Being a shell and all, this makes sense since path strings are a really common thing to interact with. The typical way to handle these in Zsh is with [variable expansion modifiers](http://zsh.sourceforge.net/Doc/Release/Expansion.html#Modifiers)

I appreciate the simplicity and power of Fish and don't think these kind of magic modifiers are a good fit for fish, but I do think fish suffers from a lack of a great builtin path management utility. I started down the road of creating a `pathmod` plugin, and maybe that is good enough, but as I was developing it I realized the only reason for this to exist is because fish could use it as a builtin, and it would be worth asking the dev team to consider adding this functionality. Perhaps as a new `string path` subcommand of the `string` utility.

Here's how it could work. First off, the canonical Zsh examples:

```zsh
$ # Zsh path manipulation
$ somepath="./README.md" 

$ # get the absolute path
$ echo ${somepath:a}
/Users/mattmc3/code/fish-shell/fish-shell/README.md

$ # get the file extension
$ echo ${somepath:e}
md

$ # get the file name
$ echo ${somepath:t}
README.md

$ # get the file directory
$ echo ${somepath:a:h}
/Users/mattmc3/code/fish-shell/fish-shell

$ # get the file root name
$ echo ${somepath:a:r}
/Users/mattmc3/code/fish-shell/fish-shell/README

$ # chaining works to go up directories
$ echo ${somepath:a:h:h:h}
/Users/mattmc3/code

$ # chaining also works to strip a double extension
$ somepath="./README.md.bak"
$ echo ${somepath:a:r:r}
/Users/mattmc3/code/fish-shell/fish-shell/README
```

In fish currently, it can cumbersome to do some of these simple path string manipulations:

```fish
$ # Fish path manipulation
$ set somepath "./README.md" 

$ # get the absolute path
$ realpath $somepath
/Users/mattmc3/code/fish-shell/fish-shell/README.md

$ # get the file extension (not all paths have one, so we need to account for extension-less paths)
$ echo (string split "." --max 1 --right $somepath)[-1]
md

$ # get the file name
$ basename $somepath
README.md

$ # get the file directory
$ dirname (realpath $somepath)
/Users/mattmc3/code/fish-shell/fish-shell

$ # get the file root name
$ echo (string split "." --max 1 --right $somepath)[1]
[...]

PR #8958, opened 2022-05-14
Builtin path, v2

faho:
## Description

This adds a `path` builtin to deal with paths.

It offers the following subcommands:

- `filter` to go through a list of paths and only print the ones that pass some filter - exist, are a directory, have read permission, ...
- `is` as a shortcut for `filter -q` to only return true if one of the paths passed the filter
- `basename`, `dirname` and `extension` to print certain parts of the path
- `change-extension` to change the extension to a different one (as a string operation)
- `normalize` and `resolve` to canonicalize the paths in various flavors
- `sort` to sort paths, also only using the basename or dirname as a key

The definition of "extension" here was carefully considered and should line up with how extensions are actually used - `~/.bashrc` doesn't have an extension, but `~/.conf.d` does (".d").

These subcommands all compose well - they can read from arguments or stdin (like string), they can use null-delimited input or output (input is autodetected - if a NULL happens in the first PATH_MAX bytes it switches automatically).

It is both a failglob exception (so like `set` if a glob passed to it fails it just doesn't get any arguments for it instead of triggering an error), and passes output to command substitution buffers explicitly split (like `string split0`) so newlines are easy to handle.

This PR is a rerun of #8265 because that was very long and hard to follow.

I intend to merge this for 3.5.0 in two weeks or so.

Fixes issue #7658

## TODOs:

There is one thing I'm not entirely sure about that would be awkward to change later:

- [ ] `filter` currently follows symlinks, so any filter applies to what the link points to.
      This isn't unreasonable, but it might be possible that turning it around and requiring an option to follow is the better choice. Or we leave it as-is and maybe add an option to ignore symlinks?


krobelus:
BTW [here](https://lore.kernel.org/git/alpine.LFD.0.98.0705090825090.4062@woody.linux-foundation.org/) is a creative strategy for cleaning up an ugly sequence of commits but that's okay, I can look at the sum of the changes.

src/builtins/path.cpp
```diff
@@ -0,0 +1,892 @@
...
+    int retval = parse_opts(&opts, &optind, 0, argc, argv, parser, streams);
+    if (retval != STATUS_CMD_OK) return retval;
+
+    int n_transformed = 0;
+    arg_iterator_t aiter(argv, optind, streams, opts.null_in);
+    while (const wcstring *arg = aiter.nextstr()) {
+        auto real = wrealpath(*arg);
+
+        if (!real) {
+            // The path doesn't exist, so we go up until we find
```
ridiculousfish, line 674: There's three important reasons for failure here:

1. `ENOENT/ENOTDIR`: Path does not exist
2. `EPERM`: Path may exist but you can't read it
3. `ELOOP`: Path is a symlink loop

For a symlink loop, it probably makes sense to "resolve as far as we can" as this does, but it would be good to add a test for that case since it's easily overlooked. Also we should change the language in the docs from "nonexistent paths" to "paths which do not exist, or otherwise cannot be resolved" or something, maybe you have ideas here.

doc_src/cmds/path.rst
```diff
@@ -0,0 +1,402 @@
...
+::
+
+    path sort [(-z | --null-in)] [(-Z | --null-out)] \
+        [(-q | --quiet)] [(-v | --invert)] \
+        [--what=basename|dirname|path] [([PATH...]
+
+
+``path sort`` returns the given paths in sorted order. They are sorted in the same order as globs - alphabetically, but with runs of numerical digits compared numerically.
+
+With ``--invert`` or ``-v`` the sort is reversed.
```
ridiculousfish, line 354: I suggest `--reverse` / `-r` instead of `--invert`, to match sort(1).

src/builtins/path.cpp
```diff
@@ -0,0 +1,892 @@
...
+static int path_sort(parser_t &parser, io_streams_t &streams, int argc, const wchar_t **argv) {
+    options_t opts;
+    opts.invert_valid = true;
+    opts.what_valid = true;
+    opts.unique_valid = true;
+    int optind;
+    int retval = parse_opts(&opts, &optind, 0, argc, argv, parser, streams);
+    if (retval != STATUS_CMD_OK) return retval;
+
+    auto func = +[] (const wcstring &x) {
```
ridiculousfish, line 723: LOL the `+ lambda` trick, first time encountering this 👏

src/builtins/path.cpp
```diff
@@ -0,0 +1,892 @@
...
+    while (const wcstring *arg = aiter.nextstr()) {
+        auto real = wrealpath(*arg);
+
+        if (!real) {
+            // The path doesn't exist, so we go up until we find
+            // something that does.
+            wcstring next = *arg;
+            // First add $PWD if we're relative
+            if (!next.empty() && next[0] != L'/') {
+                next = wgetcwd() + L"/" + next;
```
ridiculousfish, line 679: This should use `parser.vars().get_pwd_slash()` for two reasons:

1. It allows sanely handling cases where `getcwd` fails (most importantly if $PWD has been unlinked)
2. It plays better with concurrent execution, where each thread might have its own CWD

Be aware that `get_pwd_slash()` may have symlinks.

Also note the `path_apply_working_directory` function, this may be as simple as

    wcstring next = path_apply_working_directory(*arg, parser.vars.get_pwd_slash());

src/builtins/path.cpp
faho: Yeah, it's a workaround because lambda types are unique:

```c++
auto t = []{};
auto t2 = []{};
```

Here, t and t2 have different types. So reassigning breaks - `t = t2` complains about incompatible types. The `+` coerces it to a function pointer so it works (possibly at some slight cost, I dunno). It could also be possible to use std::function instead but that requires us to spell the type again, AFAICT.

See for example https://stackoverflow.com/questions/18889028/a-positive-lambda-what-sorcery-is-this.

src/builtins/path.cpp
faho: As the kids don't say: Donespacito.

doc_src/cmds/path.rst
```diff
@@ -0,0 +1,402 @@
...
+
+    path basename GENERAL_OPTIONS [PATH...]
+    path dirname GENERAL_OPTIONS  [PATH...]
+    path extension GENERAL_OPTIONS [PATH...]
+    path filter GENERAL_OPTIONS [(-v | --invert)] \
+        [-d] [-f] [-l] [-r] [-w] [-x] \
+        [(-t | --type) TYPE] [(-p | --perm) PERMISSION] [PATH...]
+    path is GENERAL_OPTIONS [(-v | --invert)] [(-t | --type) TYPE] \
+        [-d] [-f] [-l] [-r] [-w] [-x] \
+        [(-p | --perm) PERMISSION] [PATH...]
```
krobelus, line 19: no need for backslashes and "..." should have a leading space.  Also `[(-v | --invert)]` should be `[-v | --invert]` and options should probably be aligned in the source like in string.rst.

doc_src/cmds/path.rst
```diff
@@ -0,0 +1,402 @@
...
+    path normalize GENERAL_OPTIONS [PATH...]
+    path resolve GENERAL_OPTIONS [PATH...]
+    path change-extension GENERAL_OPTIONS EXTENSION [PATH...]
+    path sort GENERAL_OPTIONS [(-v | --invert)] \
+        [-u | --unique] [--key=basename|dirname|path] [([PATH...]
+
+    GENERAL_OPTIONS := [(-z | --null-in)] [(-Z | --null-out)] [(-q | --quiet)]
```
krobelus, line 26: Are we planning to add null switches to other commands as well? I don't see `path` and `string` being all that different in this regard.

doc_src/cmds/path.rst
```diff
@@ -0,0 +1,402 @@
...
+It returns 0 if any normalization or resolution was done, i.e. any given path wasn't in canonical form.
+
+Examples
+^^^^^^^^
+
+::
+
+   >_ path resolve /bin//sh
+   # The "//" is squashed, and /bin is resolved if your system links it to /usr/bin.
+   # sh here is bash (on an Archlinux system)
```
krobelus, line 291:
```suggestion
+   # sh here is bash (on an Arch Linux system)
```

share/completions/path.fish
```diff
@@ -0,0 +1,29 @@
...
+complete -f -c path -n "test (count (commandline -opc)) -lt 2" -a basename -d 'Give basename for given paths'
+complete -f -c path -n "test (count (commandline -opc)) -lt 2" -a dirname -d 'Give dirname for given paths'
+complete -f -c path -n "test (count (commandline -opc)) -lt 2" -a extension -d 'Give extension for given paths'
+complete -f -c path -n "test (count (commandline -opc)) -lt 2" -a change-extension -d 'Change extension for given paths'
+complete -f -c path -n "test (count (commandline -opc)) -lt 2" -a normalize -d 'Normalize given paths (remove ./, resolve ../ against other components..)'
+complete -f -c path -n "test (count (commandline -opc)) -lt 2" -a resolve -d 'Normalize given paths and resolve symlinks'
+complete -f -c path -n "test (count (commandline -opc)) -lt 2" -a filter -d 'Print paths that match a filter'
+complete -f -c path -n "test (count (commandline -opc)) -lt 2" -a is -d 'Return true if any path matched a filter'
+complete -f -c path -n "test (count (commandline -opc)) -lt 2" -a match -d 'Match paths against a glob'
+complete -f -c path -n "test (count (commandline -opc)) -lt 2" -a expand -d 'Expand globs'
```
krobelus, line 13: I think you forgot "sort"

src/builtins/path.cpp
```diff
@@ -0,0 +1,895 @@
...
+    UNUSED(streams);
+    opts->have_key = true;
+    opts->key = w.woptarg;
+    return STATUS_CMD_OK;
+}
+
+/// This constructs the wgetopt() short options string based on which arguments are valid for the
+/// subcommand. We have to do this because many short flags have multiple meanings and may or may
+/// not require an argument depending on the meaning.
+static wcstring construct_short_opts(options_t *opts) {  //!OCLINT(high npath complexity)
```
krobelus, line 378: I think we can drop this OCLINT comment

src/builtins/path.cpp
```diff
@@ -0,0 +1,895 @@
...
+    if (retval != STATUS_CMD_OK) return retval;
+
+    auto func = +[] (const wcstring &x) {
+        return wbasename(x);
+    };
+    if (opts.have_key) {
+        if (std::wcscmp(opts.key, L"basename") == 0) {
+            // Do nothing, this is the default
+        } else if (std::wcscmp(opts.key, L"dirname") == 0) {
+            func = +[] (const wcstring &x) {
```
krobelus, line 733: I guess we don't need the + here. Also maybe call it something like `keyfunc`?

krobelus:
> `filter` currently follows symlinks, so any filter applies to what the link points to.

Right, if /bin is a symlink then both `path filter --type=dir /bin` and `path filter --type=linke /bin` match.
This is a bit weird but it's probably the correct behavior in most situations, so it sounds ok.

krobelus:
I took a tired look at everything, looks fine

doc_src/cmds/path.rst
faho: I believe we should add them to string as well, yes.

The main reason this is in path first is because it's *necessary* for path to be complete - NUL is the only byte you can use to separate paths unambiguously, and so it exists in a lot of commands (`find -print0`). Whereas string already passes through NULs from stdin, and has `string split0` that you can use to *work* with NUL-delimited strings by passing them as arguments.

Unfortunately `string` won't be able to use the PATH_MAX trick, because strings are arbitrary.

src/builtins/path.cpp
faho: Renamed to keyfunc (and "funced" to "key", because that's what it does - it gets the key to a value).

I don't want to drop the "+" here because that leaves us with a weird asymmetry.

src/builtins/path.cpp
faho: I'm not sure if oclint is still a thing. I couldn't get it to run last time I tried.

If it's not, we should remove all of these.

share/completions/path.fish
faho: Yeah, the completions needed an update.

src/builtins/path.cpp
krobelus: > I don't want to drop the "+" here because that leaves us with a weird asymmetry.

Sure. IMO the asymmetry is justified because `+` ist just a short way of spelling a declaration's type.
When I first tried to figure out what the `+` means I removed it and recompiled, but this one actually makes no difference which makes Occam's razor less sharp.

faho:
It is done. Merged.

Issue #8536, opened 2021-12-08
complete --condition given multiple times should complicate predicate


floam:
So, it seems like currently this is basically strange undefined behavior:

```fish
$ complete foo -n false -n true -x -a bar
$ complete -C'foo '
bar
$ complete --erase foo
$ complete foo -n true -n false -x -a bar
$ complete -C'foo '
<file completions>
```

I propose this should have been equivalent to: `complete foo -n "false; and true" …`.

krobelus:
Yeah, that sounds better. Looks like we ignore all but the last -n

floam:
There might be an argument it could be equivalent to `or`, but a quick grep suggests `and` is more useful:

```
~/s/f/s/completions (master|💩2…) $ grep -r ' -n .* and.*-d' | count
398
~/s/f/s/completions (master|💩2…) $ grep -r ' -n .* or.*-d' | count
51
```

PR #8967, opened 2022-05-20
Complete: Allow passing multiple conditions

faho:
## Description

This makes it so `complete -c foo -n test1 -n test2` registers *both*
conditions (instead of using only test2, as is the case at the moment).

When it comes time to check the candidate, both conditions are attempted,
in that order. If any fails it stops, if all succeed the completion is offered.

The reason for this is that it helps with caching - we have a
condition cache, but conditions like https://github.com/fish-shell/fish-shell/blob/86ab81dadf6b8723e41c496d19fe71d9a4535b91/share/completions/string.fish#L51-L58

defeat it pretty easily, because the cache only looks at the entire
script as a string - it can't tell that the first `test` is the same.

So this will run `test (count (commandline -opc)) -lt 2` once, because it's cached, and run

```fish
test (count (commandline -opc)) -ge 2; and contains -- (commandline -opc)[2] repeat
test (count (commandline -opc)) -ge 2; and contains -- (commandline -opc)[2] pad
```

once each - so the `test` is run at least twice.

So this means we separate it into two conditions, like

```fish
complete -x -c string -n "test (count (commandline -opc)) -ge 2" -n "contains -- (commandline -opc)[2] pad" -s r -l right -d "Pad right instead of left"
```

which allows the `test` to be cached.

The entire string completions (which are quite simple) have a total of 3 distinct tests - less than two (and offer a subcommand), greater or equal to two (and offer options) and less or equal to two (and offer --help). So if, say there already are three arguments the le and lt tests only need to be done once each, instead of 17 times. If there are less than two, the gt test only needs to run once instead of 31 times.

In tests, this improves performance for the string completions by 60%
by reducing all the redundant `test` calls.

The `git` completions can also greatly benefit from this.

A slight downside is, of course, that the completions need to be altered to take advantage of this.

## TODOs:

### DIFF
--- a/src/builtins/complete.cpp
+++ b/src/builtins/complete.cpp
@@ -32,7 +32,7 @@
 /// Silly function.
 static void builtin_complete_add2(const wchar_t *cmd, bool cmd_is_path, const wchar_t *short_opt,
                                   const wcstring_list_t &gnu_opts, const wcstring_list_t &old_opts,
-                                  completion_mode_t result_mode, const wchar_t *condition,
+                                  completion_mode_t result_mode, const wcstring_list_t &condition,
                                   const wchar_t *comp, const wchar_t *desc, int flags) {
     for (const wchar_t *s = short_opt; *s; s++) {
         complete_add(cmd, cmd_is_path, wcstring{*s}, option_type_short, result_mode, condition,
@@ -59,7 +59,7 @@ static void builtin_complete_add2(const wchar_t *cmd, bool cmd_is_path, const wc
 static void builtin_complete_add(const wcstring_list_t &cmds, const wcstring_list_t &paths,
                                  const wchar_t *short_opt, const wcstring_list_t &gnu_opt,
                                  const wcstring_list_t &old_opt, completion_mode_t result_mode,
-                                 const wchar_t *condition, const wchar_t *comp, const wchar_t *desc,
+                                 const wcstring_list_t &condition, const wchar_t *comp, const wchar_t *desc,
                                  int flags) {
     for (const wcstring &cmd : cmds) {
         builtin_complete_add2(cmd.c_str(), false /* not path */, short_opt, gnu_opt, old_opt,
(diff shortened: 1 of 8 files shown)


faho:
I have converted the git completions now (locally).

This only helps with the *condition*, so in case where the argument completion actually needs to do work (by calling git) it won't show up a lot.

So e.g. `git checkout <TAB>` is sped up a negligible amount (~2%), while `git stash sa<TAB>` is sped up a lot (~30%). This will matter more on systems where the conditions are slower - #8266 points to msys having issues with `argparse`.

A quick glimpse at a profile shows that it about halves the `__fish_git_using_command` calls, and this could be increased by, ironically, splitting some - `__fish_git_using_command diff show range-diff apply` could be four separate calls so they can be cached.

Note: Rewriting this will, of course, be an issue for the people who try to backport completion scripts to older fishes. I don't believe that's a big problem. We can accomodate them to a degree by not doing backwards-incompatible *cleanup*, but we shouldn't hold back improvements in correctness or performance for that.

faho:
Fixes #8536. Huh, hadn't seen that one

src/builtins/complete.cpp
```diff
@@ -333,12 +334,11 @@ maybe_t<int> builtin_complete(parser_t &parser, io_streams_t &streams, const wch
         }
     }
 
-    if (condition && *condition) {
-        const wcstring condition_string = condition;
+    for (const auto& condition_string : condition) {
```
krobelus, line 337: the `&` is on the wrong side

krobelus:
looks great, nice improvement.

I wonder if we should add a "please restart fish" message when and old fish loads new completions that are known to be incompatible. Or perhaps we can embed them into the binary..

faho:
>I wonder if we should add a "please restart fish" message when and old fish loads new completions that are known to be incompatible

Tbh you should *always* restart fish after updating(*). We *reload* changed function files! (which I still believe we should stop doing, but even if we did autoloading means you get the new files if you do something new anyway)

(*) "updating" here refers to actual version upgrades from 3.3 to 3.4, not a few commits at a time like we do

Issue #8965, opened 2022-05-18
Git completions appear to be missing --signoff for git commit


UbiquitousPhoton:
Please tell us which fish version you are using:

Fish 3.4.1

Please tell us which operating system and terminal you are using:

Manjaro Linux, Kitty terminal, 5.17.6

The Fish completions for `git commit` appear to be missing `--signoff` - this is present for other commands, but missing from this one. It would be relatively simple for me to make a PR for this, but not so simple to test it.

I tried without third party add-ons, but the issue is also pretty clear in the completions code.

Source: Someone who has to signoff every commit.

krobelus:
> but not so simple to test it.

placing the modified git.fish in ~/.config/fish/completions/ should be enough.
Alternatively, if you built fish, you can just run that binary and it will pick up git.fish from your git worktree.

> Source: Someone who has to signoff every commit.

Looks like there is no Git config option to do that. but you can create a Git alias for `commit --signoff` or set the `commit.template` configuration variable.

PR #8969, opened 2022-05-21
Add signoff options to `git commit` completions

mhmdanas:
## Description

Add signoff options to `git commit` completions

Fixes https://github.com/fish-shell/fish-shell/issues/8965.

## TODOs:


share/completions/git.fish
```diff
@@ -1223,6 +1223,8 @@ complete -f -c git -n '__fish_git_using_command commit' -s n -l no-verify -d 'Do
 complete -f -c git -n '__fish_git_using_command commit; and __fish_git_contains_opt fixup squash' -k -a '(__fish_git_recent_commits)'
 complete -f -c git -n '__fish_git_using_command commit' -l allow-empty -d 'Create a commit with no changes'
 complete -f -c git -n '__fish_git_using_command commit' -l allow-empty-message -d 'Create a commit with no commit message'
+complete -f -c git -n '__fish_git_using_command commit' -s s -l signoff -d 'Append Signed-off-by trailer to commit message'
+complete -f -c git -n '__fish_git_using_command commit' -l no-signoff -d 'Do not append Signed-off-by trailer to commit message'
```
krobelus, line 1227: I wonder if we should just leave out the `--no-signoff` option, since it is only useful in special cases

mhmdanas: Depends on whether the completions are intended to be comprehensive or to just cover typical uses cases. I prefer the earlier.

PR #8978, opened 2022-05-29
create_manpage_completions.py: Do not overstrip commands with dots

klemensn:
The best effort parser over-eagerly strips all extensions off a manual
page file's basename, hence commands containing dots will output
completions for a different command.

Prominent examples are the mkfs.*(8) and fsck.*(8) families, e.g.
completions for mkfs.xfs.8.gz are generated for the command `mkfs`
is not only incorrect but can also filename collisions in case .fish
files for multiple commands are put into the same directory.

Thus do not strip everything past the first dot from the left, but
instead merely strip expected extensions from the right.


krobelus:
Great fix, thank you!

Issue #8979, opened 2022-05-29
git prompt extremely slow on git repos with lots of commits.


phryk:
## Meta ##

Fish: 3.4.1
OS: FreeBSD 13.0
Terminal: rxvt-unicode

No third-party customizations.

## Reproduction ##

Use one of the predefined prompts showing git info, like "Informative Vcs" or "Terlar".

```
git clone https://git.freebsd.org/ports.git
cd ports
```

## Expected behavior ##

Change directory and give me access to my shell back, fast.

## Actual behavior ##

Shell session blocked for ~15s.

## Fix ##

Have `__terlar_git_prompt` and `fish_git_prompt` pass `-uno` to `git status`,
which makes it ignore untracked files. This speeds up the prompt generation
by a factor of something like 30-100, even when no untracked files are around.

`fish_git_prompt` has some settings that would make this invalid behavior
(show_informative_status, showuntrackedfiles), but `-uno` should at least
be the default because it makes the prompt on big repos in the ballpark of
100x faster.

I would also argue that a timeout like proposed by @cg505 and @krobelus
in #6764 should be wrapped around calls to prompt functions.

IlanCosman:
Not that this solves the problem, since the default prompts should still be responsive, but you might want to check out https://github.com/IlanCosman/tide, which has an async git prompt. 0 delay ever, always have access to your shell. Sorry for the plug, just want to make sure you have a good fish experience 😅

faho:
If -uno helps, your problem isn't lots of commits - it's lots of untracked files.

And the normal fish_git_prompt *already* doesn't care about untracked files, it requires you to specifically set $__fish_git_prompt_showuntrackedfiles (and to not have git's bash.showuntrackedfiles option set to false on that repo).

What you're using however is the *informative* mode (which the informative sample prompt turns on). In this mode it's meant to *count* all the files. We pass "-unormal" (not "-uno") to not descend into untracked directories. If we passed -uno instead it wouldn't count untracked files *at all*, which doesn't seem right. (for historical reasons, the terlar prompt has its own git prompt implementation - I've been working on unifying the two)

The [fish_git_prompt documentation](https://fishshell.com/docs/current/cmds/fish_git_prompt.html#cmd-fish-git-prompt) already tells you that informative mode might be slow, and already explains you can disable it per-repo with:

```fish
git config --local bash.showInformativeStatus false
```

[...]

PR #8980, opened 2022-05-30
git_prompt: Only show untracked files in informative mode if asked

faho:
This makes it so

1. The informative status can work without showing untracked
files (previously it was disabled if bash.showUntrackedFiles was
false)
2. If untrackedfiles isn't explicitly enabled, we use -uno, so git
doesn't have to scan all the files.

In a large repository (like the FreeBSD ports repo), this can improve
performance by a factor of 5 or up.

----

The impetus for this is that our informative+vcs prompt already tries to disable untracked file display via the (undocumented) __fish_git_prompt_hide_untrackedfiles variable, so in that case we're just wasting time.

We *could* also make it so we recognize that option and skip untracked display (or make it use $__fish_git_prompt_status_order without untrackedfiles), but:

1. We've decided that it's not useful there, so we apparently don't like it in the prompt meant to showcase the informative git prompt!
2. This can be *very* costly and so should probably be enabled explicitly.

Part of #8979 (well, except for the terlar bit)

(we probably also want to make it so you can set $__fish_git_prompt_showuntrackedfiles to 0 to explicitly *disable* it)

## TODOs:

### DIFF
--- a/doc_src/cmds/fish_git_prompt.rst
+++ b/doc_src/cmds/fish_git_prompt.rst
@@ -26,7 +26,10 @@ The ``fish_git_prompt`` function displays information about the current git repo
 
 There are numerous customization options, which can be controlled with git options or fish variables. git options, where available, take precedence over the fish variable with the same function. git options can be set on a per-repository or global basis. git options can be set with the ``git config`` command, while fish variables can be set as usual with the :ref:`set <cmd-set>` command.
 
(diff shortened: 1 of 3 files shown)


faho:
>the (undocumented) __fish_git_prompt_hide_untrackedfiles variable

Here's my plan for the future:

1. Document $__fish_git_prompt_status_order
2. Read the git prompt boolean variables as 0/1 toggles instead of checking if they are set or not, so you can *disable* them
3. Remove __fish_git_prompt_hide_X, as it's now superfluous (and was only ever used with informative status)
4. (possibly rename the variables to without the dunderscore - $fish_git_prompt_show_untrackedfiles etc)

PR #8983, opened 2022-05-31
Improve newline behavior of `kill-whole-line`

SeekingBlues:
## Description

Previously, `kill-whole-line` kills the line and its following newline. This is insufficient when we are on the last line, because it would not actually clear the line. The cursor would stay on the line, which is not the correct behavior for bindings like `dd`.

Also, `cc` in vi-mode used `kill-whole-line`, which is not correct because it should not remove any newlines. We have to introduce another special input function (`kill-inner-line`) to fix this.

## TODOs:


ridiculousfish:
LGTM. Merged as cf620c829b4b826cafb2ed92349f8324d8b5045f. Thank you, great fix!

PR #8990, opened 2022-06-02
fish_git_prompt: drop --ignored flag in git status

zanchey:
The git prompt tests are failing on CentOS 7, because it only has git 1.8. The
git prompt uses `git status --ignored=no`, which only started to work in 2.16.

This was introduced in 542a78a4c3 which has not been released, so no need to add
it to the CHANGELOG.

## TODOs:

### DIFF
--- a/share/functions/fish_git_prompt.fish
+++ b/share/functions/fish_git_prompt.fish
@@ -267,7 +267,9 @@ function fish_git_prompt --description "Prompt function for Git"
                 # With just dirty, it's ~20%.
                 set -l opt -uno
                 test "$untracked" = true; and set opt -unormal
-                set -l stat (command git -c core.fsmonitor= status --porcelain -z --ignored=no $opt | string split0)
+                # Don't use `--ignored=no`; it was introduced in Git 2.16, from January 2018
+                # Ignored files are omitted by default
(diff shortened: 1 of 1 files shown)


share/functions/fish_git_prompt.fish
```diff
@@ -267,7 +267,9 @@ function fish_git_prompt --description "Prompt function for Git"
                 # With just dirty, it's ~20%.
                 set -l opt -uno
                 test "$untracked" = true; and set opt -unormal
-                set -l stat (command git -c core.fsmonitor= status --porcelain -z --ignored=no $opt | string split0)
+                # Don't use `--ignored=no`; it was introduced in Git 2.16, from January 2018
+                # Ignored files are omitted by default
```
faho, line 271: Git's documentation here is a bit misleading:

>--ignored[=<mode>]
> Show ignored files as well.

> The mode parameter is used to specify the handling of ignored files. It is optional: it defaults to
           traditional.

> The possible options are:

> •   traditional - Shows ignored files and directories, unless --untracked-files=all is specified, in which
               case individual files in ignored directories are displayed.

> •   no - Show no ignored files.

What that "it defaults to traditional" means is that `--ignored` is the same as `--ignored=traditional`. Not giving a `--ignored` *at all* is the same as `--ignored=no`.

So leaving this off is okay.

Issue #7762, opened 2021-03-01
Selection includes character under cursor for normal mode (non vi)


daleeidd:
I am using fish in emacs mode (non vi mode). And the selection is off by one.

![0000](https://user-images.githubusercontent.com/5249806/109559849-0d2f1200-7a90-11eb-9178-df3068f13533.jpg)

The above screenshot shows me selecting two characters "he", but also erroneously includes the following "l" character.

Version: 3.1.2
OS: MacOS 11.2.1
Terminal: xterm-256color

I fixed (or caused?) this issue in #3684, but it looks like it regressed with #6329. The real issue appears to be that emacs and vi are handled the same which has led to a back and forth on whether selection should be inclusive or exclusive of the cursor character.

MForster:
How about making this dependent on the mode, or alternatively on the value of `$fish_cursor_default` etc?

MForster:
I'm proposing a change in #8994 that sets the selection according to the currently configured cursor shape, if any. This seems to me to be the most principled way of addressing the different preferences of users with block cursors vs. line cursors.

What do you think?

ridiculousfish:
Let's discuss in the PR #8994

faho:
Closing because it appears to have been solved by #8994.

PR #6329, opened 2019-11-16
Change vi selection mode to be inclusive

rayhogenson:
The current cursor position should be included in the selection to be
consistent with the behavior of vi.

Fixes #5770

## Description

The current cursor position should be included in the selection. Because it makes sense to keep sel_stop_pos as the exclusive upper-bound, it must be adjusted by 1 to make the selection inclusive.

As far as I can tell, emacs mode does not use these code paths, but I'd be happy for someone more familiar with the code base to correct me.

Fixes issue #5770

## TODOs:

- [ ] Changes to fish usage are reflected in user documentation/manpages.
- [ ] Tests have been added for regressions fixed
- [ ] User-visible changes noted in CHANGELOG.md

Issue #3684, opened 2016-12-29
begin-selection: including extra character in selection


daleeidd:
- [x] Have you checked if problem occurs with [fish 2.4.0](/fish-shell/fish-shell/releases/tag/2.4.0)?
- [x] Tried fish without third-party customizations *(check `sh -c 'env HOME=$(mktemp -d) fish'`)*?

**fish version installed** *(`fish --version`)*:
`fish, version 2.4.0`

**OS/terminal used**: 
macOS 10.12.*
iTerm2 3.*

The prompt selection is including an extra character (the one in front of the cursor).

## Reproduction steps
1. Setup:
```
function fish_user_key_bindings
  bind \ce begin-selection
  bind \cr kill-selection
end
```
2. Place the cursor somewhere except the beginning or end
<img width="118" alt="testing" src="https://cloud.githubusercontent.com/assets/5249806/21537643/9504c17e-cde8-11e6-8959-e3915f1b910f.png">
3. `control+e` then `left` or `right`
5. `control+r`

## Results

After left:
<img width="118" alt="testing-left" src="https://cloud.githubusercontent.com/assets/5249806/21537648/a0efa2ba-cde8-11e6-9a8b-75c21cd32024.png">
After left kill:
<img width="87" alt="testing-left-del" src="https://cloud.githubusercontent.com/assets/5249806/21537770/b058f052-cde9-11e6-855c-de1ed22a9f83.png">
After right:
<img width="118" alt="testing-right" src="https://cloud.githubusercontent.com/assets/5249806/21537652/a69fb7d6-cde8-11e6-8ab9-20ef584f9d40.png">
After right kill:

ridiculousfish:
nice find

ridiculousfish:
Thanks, I removed the label

PR #8994, opened 2022-06-04
Consider cursor width when updating selection

MForster:
## Description

This fixes the handling of the right end of the selection. Currently the right end is considered to be at the cursor position + 1. When using a `block` or `underline` cursor this is arguably correct, because the cursor has a width of 1 and spans from the current position to the next:
```
    x x [x x x̲] x
```

This is incorrect though (or at least very unintuitive), when using a `line` cursor:
```
    x x [x x|x] x
```

This pull requests changes the strategy for determining the end of the selection in the following way:

* If the current cursor as determined by `$fish_cursor_<bind_mode>` is set to `line`, then a cursor width of `0` is assumed.
* In all other cases, including `block` and `underscore` as well as when no value is set we retain the previous behavior of assuming a cursor width of `1`.
```
    x x [x x x̲] x
    x x [x x|]x x
```

This change should not affect many users, because the selection is probably used most by vi-mode users, who are also likely to use a block cursor.

Fixes issue #7762. Please consider the history in #3684 and #6329. The back-and-forth in these changes demonstrates how users  of line cursors (including myself) want a different behavior from users of block cursors.

## TODOs:

- [x] Tests have been added ~~for regressions fixed~~


MForster:
Note that I'm happy to improve tests and documentation as well. I just wanted to get feedback about whether this is an acceptable change before putting the effort in.

ridiculousfish:
Thank you for opening this.  I understand the idea: if the cursor shape is a line then it seems to be "between" chars, before the next one; but if it's a block it "overlaps" the next char. It makes sense for text selection to respect that difference. But I think this particular approach mixes different concerns too much. Text selection is logical, while cursor shape is about visual rendering, and it seems wrong to have a logical behavior depend on a visual rendering.

I wonder if this can just be implemented through pure bindings. I didn't know that anyone was using selection with the default emacs-style key bindings! Can you please share the bindings you're using?

MForster:
I am still working on my bindings. I didn't get that far because of this issue. The basis idea would have been this:
```
function __bind_movement_key -a key shiftkey binding
  bind $key $binding
  bind $shiftkey begin-selection -m selecting $binding
  bind -M selecting $key $binding end-selection -m default
  bind -M selecting $shiftkey $binding
end

  __bind_movement_key \e\[C \e\[1\;2C forward-char
  __bind_movement_key \e\[D \e\[1\;2D backward-char
```

The other example of doing the same thing would be https://github.com/daleeidd/natural-selection.

MForster:
I had a few more thoughts that I wanted to share:

> Text selection is logical, while cursor shape is about visual rendering, and it seems wrong to have a logical behavior depend on a visual rendering.

Yes, that makes sense. Arguably it is the current behavior, though, that mixes visual and logical representation. For example, logically, there is little reason why the [right side](https://github.com/fish-shell/fish-shell/blob/49567a060eabf593385178e6b89263d73e629814/src/reader.cpp#L1032) of the selection should be treated differently from the [left side](https://github.com/fish-shell/fish-shell/blob/49567a060eabf593385178e6b89263d73e629814/src/reader.cpp#L1028). A symmetrical approach would be more straight-forward. I'm sure that a vi-mode user may disagree with me here :-). It may also depend on whether you consider empty selections a useful concept.

> I didn't know that anyone was using selection with the default emacs-style key bindings!

I think it's just not very usable now. There is this issue, but then there is the other thing that it's hard to set up the bindings. See how I need four bindings in the example above where it's only one logical operation with four different mode transitions. My intention would have been to find out if it's possible to make this work, and then implement and maybe contribute a shift-based selection mode. 

Also note that I'm not interested in "Emacs" mode either. Emacs is just closer to what I want than VI. I would prefer a mode that is more in line with most graphical user interfaces that pervasively use shift-select. And other people would want that, too:

* https://stackoverflow.com/questions/61858253/how-to-select-text-with-arrow-keys-in-fish-shell
* https://superuser.com/questions/1295038/in-fish-shell-how-to-select-with-the-cursor-while-shift-is-held
* https://github.com/daleeidd/natural-selection

I think that a shift-select mode would be something that fits well into Fish's idea of being a more modern shell that is more accessible to newcomers. Not to complain about VI and Emacs, but they are old, and modern editors usually use shift-select, unless they want to specifically cater to previous VI/Emacs users.

> I wonder if this can just be implemented through pure bindings.

I could live with that, but I'm usure how to approach this. I'm new to Fish scripting. Any advice?

MForster:
> I wonder if this can just be implemented through pure bindings.

So, I've looked through the selection code in `reader.cpp`, and came to the conclusion that it is not currently possible to create a selection that has a line cursor coincide with the end of the selection. All code that changes `selection->stop` adds `+ 1`, and code that moves the cursor updates the selection using `update_buff_pos`.

I can see these alternative ways forward to support selection with both line cursors and block cursors:

1. Have a way to configure how the selection gets extended in a way that is independent of `$fish_cursor_<bind_mode>$`. Maybe a new environment variable? `$fish_extend_selection` or something. I don't have an idea of a good name.
2. Add a new input function, or alternatively builtin to modify the selection prgrammatically. Maybe a new option to the `commandline` builtin?

Would you prefer one of these?

daleeidd:
> I wonder if this can just be implemented through pure bindings. I didn't know that anyone was using selection with the default emacs-style key bindings!

I can concur that it is not possible with the current version. I believe I tried everything to eventually resorting to using a local fork of fish.

MForster:
Thanks, @daleeidd.

Do you have input on how well the suggested alternative solutions would work for you?

daleeidd:
The current PR would work but also having option 2 would be nice generally. I think the simplest solution would be that when not in vi mode, selection would not have the +1 offset.

faho:
>But I think this particular approach mixes different concerns too much. Text selection is logical, while cursor shape is about visual rendering, and it seems wrong to have a logical behavior depend on a visual rendering.

What's more, this doesn't work in case you haven't told fish to change the cursor (currently cursor shape changing is heavily tied to vi-mode), and it doesn't work in case the terminal doesn't obey the cursor shape sequence (this is really really easy to misdetect, and we e.g. don't enable it inside tmux).

I'm afraid this approach is, as the french say, "no bueno" and would prefer a slightly different approach.

I don't believe adding more bind functions is helpful, because you'd have to add a separate variant of *every* bind function that extends the selection, IIUC? So you'd have `forward-bigword` and `forward-bigword-line` and that doesn't really work.

So either a separate toggle "$fish_cursor_is_after_selection", or a rework of the cursor shape mechanism - $fish_cursor_shape, set to the current cursor shape, would be warranted.

MForster:
So, is `$fish_cursor_is_after_selection` generally acceptable to people? I'd be very happy to go that direction.

MForster:
I've pushed a commit to use `$fish_select_char_after_cursor=1`. Would this be acceptable?

MForster:
@ridiculousfish @faho, any advice on how to proceed?

faho:
Terribly sorry, we were busy pushing out a release.

As it stands, I'd have to test but this seems an acceptable solution.

The code would need some work:

1. We typically store these sorts of things via the env_dispatch mechanism. You'd have to add the variable [here](https://github.com/fish-shell/fish-shell/blob/da020c0641e4fcc06f623616de4bad5331c56342/src/env_dispatch.cpp#L316-L330) and see that it's passed down into the reader so it can simply read a C++-level variable instead of always going down into the script-level variable stack. Compare the history settings, for instance.
2. It would have to be hooked up to be set in fish_vi_cursor - if the cursor shape is set to "line" (that is if the variable *contains* "line") it would be set to 0, else to 1.
3. Documentation and tests - at the very least something in https://github.com/fish-shell/fish-shell/blob/master/doc_src/interactive.rst#command-line-editor 

In addition, we'd have to discuss the name - these things matter, so it makes sense to spend a bit of time on seeing if we can find something better. (of course the trick is not being stuck in trying to find the perfect name, but that's fixable by just picking something if no clear winner emerges)

Also I'm not sure if this should have an impact on deletion/cursor movement in general. For instance "x" in vi normal mode is bound to "delete-char forward-single-char backward-char". Is that the same regardless of cursor shape or not? Is that correct or are we sometimes off-by-one if the cursor is a line?

MForster:
> Terribly sorry, we were busy pushing out a release.

No worries whatsoever. I'm well aware how this goes, and this is our spare time after all. I may take some time to proceed with the change as well. Just makings sure that this doesn't get forgotten :-).

> The code would need some work: [...]

Yes, definitely. And I'm very happy to put in the effort. I just wanted to finish the UI design to avoid wasted effort. Thanks for the change suggestions. Very appreciated.

Maybe we can continue the open UI questions in parallel to me working on the code changes?

> In addition, we'd have to discuss the name

Agreed that this is tricky. You suggested `$fish_cursor_is_after_selection`, which only works well for block cursors. And having written this I realized that `$fish_select_char_after_cursor` works best with line cursors and is ambiguous with block cursors. `$fish_selection_includes_current_char` doesn't work if selecting right-to-left with a line cursor. `$fish_selection_end_includes_current_char` is precise but ridiculously long. Maybe that's ok for a setting that users will only touch once and then forget about... 

We could also consider something `$fish_selection_end=left_of_current_char/right_of_current_char`, but that would invite misspellings of the value... Note that I'm suggesting "current char" instead of"cursor", because the latter is more ambiguous, depending on cursor shape.

As far as I'm concerned I'm happy to go with anything you suggest.

> Also I'm not sure if this should have an impact on deletion/cursor movement in general. For instance "x" in vi normal mode is bound to "delete-char forward-single-char backward-char". Is that the same regardless of cursor shape or not? Is that correct or are we sometimes off-by-one if the cursor is a line?

Not being a vi user, I'm out of my depth here. I'm confused already why that binding moves right and immediately left. Isn't that a no-op? Or is this to ensure that the cursor ends up "on" the last character instead of beyond? I agree that this only makes sense with a block cursor. But I don't know what a vi user would expect here. It's probably best to do the same as vi? On my system vim has a line cursor, and it moves it to the left on `x` at the end of the line. So, my suggestion would be to not touch this behavior.

ridiculousfish:
Just a note I haven't forgotten this, I'll take a look as soon as I can

MForster:
No worries. There is no time pressure.

ridiculousfish (changes requested):
see above

MForster:
Thanks for the in-depth reply! `fish_cursor_selection_mode=inclusive/exclusive` sounds good to me.

> The default is `exclusive`.

Just checking: `exclusive` is not the current behavior. Are you suggesting to change the default? That would affect more users than if we made `inclusive` the default.

> Probably this roadmap is more work than you anticipated

Nah, I'm fine. I'm happy that I got the discussion started in earnest. Thank you for the detailed explanations on how to implement the change.

> but I think you've identified a real problem and I'm excited to start building up a set of key bindings more familiar to GUI text editor users.

Let's do it! :-)

tests/pexpects/cursor_selection.py
```diff
@@ -0,0 +1,61 @@
...
+expect_prompt()
+
+sendline("bind -k nul begin-selection")
+expect_prompt()
+sendline("bind ! 'echo -n \"<$(commandline --current-selection)>\"'")
+expect_prompt()
+
+sendline("set fish_cursor_selection_mode inclusive")
+
+sendline(f"echo{home}{select}{dump}")
```
faho, line 19: The test suite is python 3.5+, so no f-strings.

MForster: > The test suite is python 3.5+, so no f-strings.

Fixed in [6569a0e](https://github.com/fish-shell/fish-shell/pull/8994/commits/6569a0ebd397e8a9f2e1db91e408b512ac4d555e).

MForster:
I have addressed the main review feedback:

* The implementation now uses the `env_dispatch` mechanism. I've had a hard time inferring how the source should be ordered. Let me know if there is a convention that I missed.
* I have added basic documentation and a change log.
* I have added a test.

This currently uses `exclusive` as the new default as suggested by @ridiculousfish. I still think that `inclusive` would be a safer choice, but I will defer to your opinion.

As far as I am concerned this is ready for a full review. Please also let me know if I should rebase and/or squash my commits.

ridiculousfish:
This is great!! Thank you! Merged as 91c68ec1af6c9cc682d6c4e912f8aed3db9ab802

PR #8997, opened 2022-06-05
Update `kitty` completion

kidonng:
## Description

This is [an updated version of the completion script](https://github.com/kovidgoyal/kitty/blob/cd7751681d15d665cf817918b88f42938f724b12/shell-integration/fish/vendor_completions.d/kitty.fish) used by kitty's [shell integration](https://sw.kovidgoyal.net/kitty/shell-integration/), as can be seen via `kitty +complete setup fish2`.

## TODOs:

### DIFF
--- a/share/completions/kitty.fish
+++ b/share/completions/kitty.fish
@@ -1,6 +1,7 @@
-function __kitty_completions
-    # Send all words up to the one before the cursor
-    commandline -cop | kitty +complete fish
+function __ksi_completions
+    set --local ct (commandline --current-token)
+    set --local tokens (commandline --tokenize --cut-at-cursor --current-process)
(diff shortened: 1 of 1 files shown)


share/completions/kitty.fish
```diff
@@ -1,6 +1,7 @@
-function __kitty_completions
-    # Send all words up to the one before the cursor
-    commandline -cop | kitty +complete fish
+function __ksi_completions
```
faho, line 1: Can we keep the "kitty" name?

I quite dislike the "ksi" abbreviation here and it has confused users in the past.

faho: Oh, wait, this is a vendored thing.

Ideally packages would simply install their completions instead of us requiring to ship guff like this, but in this case it's probably okay.

PR #9003, opened 2022-06-07
Make doc css not depend on sphinx' css

faho:
This has required workarounds a few times, plus if it changes it might
break our theme. See e.g.

4712da3eb1d2324e5fc624f77d96b06eb8898930
e27456df2494a238f2a8ba18b4f71cde360daa3f
a6d484836e55a0a1aa1f04c8b59ce4461fc323f4
85522036f5edf985efa92d0127da16dcef4c300f

So we import the rules we *use* and throw away the rest. Note that
this might still have rules that are no longer necessary - e.g. some
that are required to work around sphinx bugs would still be left.

It could benefit from some cleanup and simplification, and from
switching to a flex layout instead of the 230px hardcoded
sidebar - sphinx tried that, but it doesn't really work with our
narrow layout, so we disabled it again.

## TODOs:

### DIFF
--- a/doc_src/python_docs_theme/static/pydoctheme.css
+++ b/doc_src/python_docs_theme/static/pydoctheme.css
@@ -188,6 +188,11 @@ p.admonition-title {
   font-weight: bold;
 }
 
+div.admonition p.admonition-title + p {
+  display: inline;
+}
+
+
 div.footer {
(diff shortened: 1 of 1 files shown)


faho:
Okay, merged now so we can get some testing in.

PR #9012, opened 2022-06-09
updated function __fish_print_portage_repository_paths.fish

pl3w5y:
I've updated this function to cover the main ways repos.conf can end up and tested them on my system.

-no file or dir, so just use default
-repos.conf as a file
-repos.conf as a dir containing .conf files
-adds the default if gentoo.conf is not present

There are a couple of ways a you can have a working Portage but not get a full list of completions just like the previous version, but being off-docs or not recommended I didn't want to make the whole function about checking strings just for them.

### DIFF
--- a/share/functions/__fish_print_portage_repository_paths.fish
+++ b/share/functions/__fish_print_portage_repository_paths.fish
@@ -6,9 +6,9 @@ function __fish_print_portage_repository_paths --description 'Print the paths of
     and set b (find $a -type f )
     test -f $a
     and set b $a
-    test -n (echo $b)
-    and string match -q "[gentoo]" (cat $b)
+    test -n "$b"
+    and string match -q "[gentoo]" -- (cat $b)
     and set c $b
     or set -a c $b
-    cat $c | string match -g -r 'location = (.*$)'
+    cat $c | string match -g -r '^location = (.*$)'
 end
--- a/share/functions/__fish_print_portage_repository_paths.fish
+++ b/share/functions/__fish_print_portage_repository_paths.fish
@@ -1,18 +1,14 @@
 function __fish_print_portage_repository_paths --description 'Print the paths of all configured repositories'
-    set a /etc/portage/repos.conf
-    set default /usr/share/portage/config/repos.conf
-    set confs
-    if test -f $a
-        set confs $a
-    else if test -d $a
-        set -p confs (find $a -type f -name '*.conf')
-        if not contains "$a/gentoo.conf" $confs
-            set -p confs $default
-        end
-    else
-        set confs $default
-    end
-    for b in $confs
-        cat $b | string match -g -r 'location = (.*$)'
-    end
+    set -l a /etc/portage/repos.conf
+    set -l b
+    set -l c /usr/share/portage/config/repos.conf
+    test -d $a
+    and set b (find $a -type f )
+    test -f $a
+    and set b $a
+    test -n (echo $b)
+    and string match -q "[gentoo]" (cat $b)
+    and set c $b
+    or set -a c $b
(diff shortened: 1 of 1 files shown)


share/functions/__fish_print_portage_repository_paths.fish
```diff
@@ -1,4 +1,18 @@
...
+    set confs
+    if test -f $a
+        set confs $a
+    else if test -d $a
+        set -p confs (find $a -type f -name '*.conf')
+        if not contains "$a/gentoo.conf" $confs
+            set -p confs $default
+        end
+    else
+        set confs $default
```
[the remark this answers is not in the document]
pl3w5y: repos.conf has extra steps, in general Portage reads everything in order and always overwrites. It works nicely but even the 2nd run of the internal command to dump repo data takes hundreds of ms.

From some further work today I can see that looking for the file gentoo.conf is not a good choice, but thanks to these tips, this especially is something I probably do alot.

pl3w5y:
Hi, while I don't think my git use has been ideal, this has been a very interesting time and thanks for the inspiring tips.

Enchanted by this new if structure, I've ended up forgetting about your last comment until now, however since this works on 3.4.0 I felt it best to commit like this and offer to change to the older syntax if needed.

share/functions/__fish_print_portage_repository_paths.fish
```diff
@@ -1,4 +1,14 @@
 function __fish_print_portage_repository_paths --description 'Print the paths of all configured repositories'
-    # repos.conf may be a file or a directory
-    find /etc/portage/repos.conf -type f -exec cat '{}' + | string replace -r --filter '^\s*location\s*=\s*(\S+)' '$1'
+    set -l a /etc/portage/repos.conf
+    set -l b
+    set -l c /usr/share/portage/config/repos.conf
+    test -d $a
+    and set b (find $a -type f )
+    test -f $a
+    and set b $a
```
krobelus, line 8: I liked the if-else-if-chain you had before :(

share/functions/__fish_print_portage_repository_paths.fish
```diff
@@ -1,4 +1,14 @@
...
+    test -n (echo $b)
```
krobelus, line 9: this can be simplified to `test -n "$b"` or `set -q b[1]`

share/functions/__fish_print_portage_repository_paths.fish
```diff
@@ -1,4 +1,14 @@
...
+    and string match -q "[gentoo]" (cat $b)
```
krobelus, line 10: there's a subtle problem here: what if $b contains a line starting with `--foo`.
`string match` will try to parse that as option and fail.

there are multiple ways to avoid this:

```suggestion
    and string match -q "[gentoo]" -- (cat $b)
```

or

```suggestion
    and cat $b | string match -q "[gentoo]"
```

or

```suggestion
    grep -Fxq "[gentoo]" $b
```

share/functions/__fish_print_portage_repository_paths.fish
```diff
@@ -1,4 +1,14 @@
...
+    and set c $b
+    or set -a c $b
+    cat $c | string match -g -r 'location = (.*$)'
```
krobelus, line 13: If comments are allowed in the location line we might want to exclude them from the match

share/functions/__fish_print_portage_repository_paths.fish
pl3w5y: It's more readable with all the 'end's but as a returning user I felt some desire to use the new syntax, but have no strong opinion about it

PR #9015, opened 2022-06-15
Add function mc, switching $PWD to dir, where MC finished.

mcepl:
Rather crude port of /usr/share/mc/mc-wrapper.sh from the Midnight Commander itself.


faho:
I'm sorry, this isn't the sort of function we tend to add. It's entirely okay if this lives outside of fish.

For mc specifically, I don't believe adding more support for it is all that useful given that [it has been broken with fish since 3.3.0 a year ago](https://midnight-commander.org/ticket/4258).

PR #9020, opened 2022-06-16
Show manpage for command under cursor in `__fish_man_page`

newsch:
## Description

This lets you check the manpage for a leading command by moving
the cursor over it, matching the behavior of tab complete.

It also lets you select the man page for the base of a two-part command
like `string match`.

I added the additional regex case because
`commandline -t` returns an empty string when the cursor is after a
space, e.g. at the end of 'sudo ', which the later checks don't handle.

This diagram shows the manpage picked for different cursor positions:

    > sudo -Es time git commit -m foo
      +-------++---++--++------------+
      |       ||   ||  ||            |
      |       ||   ||  |+------------+
      |       ||   ||  |  git-commit
      |       ||   |+--+
      |       ||   | git
      |       |+---+
      |       | time
      +-------+
         sudo

[asciicast](https://asciinema.org/a/502248)

## TODOs:

### DIFF
--- a/share/functions/__fish_man_page.fish
+++ b/share/functions/__fish_man_page.fish
@@ -1,6 +1,6 @@
 function __fish_man_page
-    # Get all commandline tokens not starting with "-"
-    set -l args (commandline -po | string match -rv '^-')
+    # Get all commandline tokens not starting with "-", up to and including the cursor's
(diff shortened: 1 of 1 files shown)


krobelus:
great change, thanks!

PR #9026, opened 2022-06-20
Try to use xterm{-256color,} if $TERM could not be used

faho:
This is very very very likely to result in an almost fully functional
terminal, as opposed to a "minimally functional" one.

Xterm isn't *entirely* a superset of ansi, but where it isn't it's likely that the running terminal behaves like xterm and not like ansi.

## TODOs:


ridiculousfish:
Saucy! Looks good to me.

PR #9029, opened 2022-06-21
`set --show`: Show the originally inherited value, if any

faho:
This adds a line to `set --show`s output like

```
$PATH: originally inherited as |/home/alfa/.local/bin:/usr/local/sbin:/usr/local/bin:/usr/bin:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl:/var/lib/flatpak/exports/bin|
```

to help with debugging.

Note that this means keeping an additional copy of the original
environment around. At most this would be one ARG_MAX's worth, which
is typically about 2M on linux and typically less elsewhere.

## TODOs:

This is a quick implementation that might not be ideal, I'm not sure if there isn't some better way to implement it, or if storing the entire environment is seen as too wasteful - an alternative would be to just store the variable *names* so we can say "was originally inherited" but not with what value.

### DIFF
--- a/src/builtins/set.cpp
+++ b/src/builtins/set.cpp
@@ -556,14 +556,22 @@ static int builtin_set_show(const wchar_t *cmd, const set_cmd_opts_t &opts, int 
                             const wchar_t **argv, parser_t &parser, io_streams_t &streams) {
     UNUSED(opts);
     const auto &vars = parser.vars();
+    auto inheriteds = env_get_inherited();
     if (argc == 0) {  // show all vars
-        wcstring_list_t names = parser.vars().get_names(ENV_USER);
+        wcstring_list_t names = vars.get_names(ENV_USER);
         sort(names.begin(), names.end());
         for (const auto &name : names) {
             if (name == L"history") continue;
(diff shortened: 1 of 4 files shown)


faho:
Note: On my system, the environment is typically under 2K - 1.8K for my normal user, 0.5K for root. On linux, you can see it with `wc -c /proc/$fish_pid/environ`

ridiculousfish:
Nice fix, LGTM. I agree that the imported environment variables are not a major concern with respect to memory usage.

ridiculousfish (approved):

IlanCosman:
IMO this feels wasteful in complexity and space for the value it provides since `set --show` is only rarely used. I guess this does help with debugging the messing-with-path problem so it does provide some value 🤷‍♂️ Idk it just doesn't "feel" right, but that's not worth much, especially if y'all think it's a good idea 🙂

faho:
It is, indeed, a debug aid.

We have a lot of people coming in with "where does my $PATH come from? what's up with $LC_ALL?", and if we can tell them to just run `set --show PATH`, and that will explain whether it got the values from the parent or not, we can then tell them whether it's in fish's config or elsewhere on the system much quicker.

PR #9039, opened 2022-06-24
Return true consistently from all man parsers

RReverser:
## Description

Maybe I'm missing something and this is intentional (perhaps best to leave a comment if so), but I've noticed that those few parsers don't have `return True` in the end of respective `parse_man_page` methods.

Due to that, the main loop in `parse_manpage_at_path` assumes that those parsers have failed, and keeps walking and trying to re-parse the man page with other parsers, even though the completions for all the options have been already generated.

Was this an accidental oversight or is it some sort of intentional fallthrough to augment output with results from several parsers at once?

Fixes issue #

## TODOs:

### DIFF
--- a/share/tools/create_manpage_completions.py
+++ b/share/tools/create_manpage_completions.py
@@ -318,6 +318,8 @@ class Type1ManParser(ManParser):
             options_section = options_section[options_matched.end() - 3 :]
             options_matched = re.search(options_parts_regex, options_section)
 
+        return True
+            
     def fallback(self, options_section):
         add_diagnostic("Trying fallback")
         options_parts_regex = re.compile("\.TP( \d+)?(.*?)\.TP", re.DOTALL)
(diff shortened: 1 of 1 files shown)


ridiculousfish:
I can't say I remember how this parser works, but when I apply this change and run `fish_update_completions` I get a bunch of bogus completions. For example for `xzcat`:

```
complete -c xzcat -o '0e	256'
complete -c xzcat -o '1e	1'
complete -c xzcat -o '2e	2'
complete -c xzcat -o '3e	4'
complete -c xzcat -o '4e	4'
complete -c xzcat -o '5e	8'
complete -c xzcat -o '6e	8'
complete -c xzcat -o '7e	16'
complete -c xzcat -o '8e	32'
complete -c xzcat -o '9e	64' -d 'For example, there are a total of four presets that use 8 MiB dictionary, who…'
```

I haven't dug into the code to understand why.

RReverser:
That is weird. I wish that code was a bit easier to follow...

PR #9042, opened 2022-06-28
`file` completion nuances on macOS

kidonng:
## Description

The `file` program coming with macOS 12.4 (21F79) has a slight difference with [vanilla version](https://man.archlinux.org/man/core/file/file.1.en#i) in that it uses uppercase `-I` for `--mime`, and `-i` was assigned another functionality:

```
     -i      If the file is a regular file, do not classify its contents.

     -I, --mime
             Causes the file command to output mime type strings rather than the more
             traditional human readable ones.  Thus it may say ‘text/plain;
             charset=us-ascii’ rather than “ASCII text”.
```

## TODOs:


faho:
Thanks!

faho:
Thanks!

PR #9055, opened 2022-07-06
Fix long descriptions in gdb.fish

amordo:
Work on #6981


krobelus:
Thank you!

amordo:
@krobelus I see `Quiet` description is not in single quotes. Does some side effect appear with them there?
I tested this completion putting it in `~/.config/fish/completions/` using quotes everywhere (for consistency), met no problems🤔

krobelus:
Quotes make no difference here, this is just the style that our formatter (`fish_indent`) uses.

PR #9056, opened 2022-07-07
link to faq greeting message

Jeroendevr:
I am not proficient with rst, but I think it would be helpful to link from this cmd to the faq page of the greeting. 
How do find the right :ref: attribute for the faq? couldn't find it in the faq itself. The link to docs
https://fishshell.com/docs/current/faq.html#how-do-i-change-the-greeting-message

## Description

Talk about your changes here.

Fixes issue #

## TODOs:

### DIFF
--- a/doc_src/cmds/fish_greeting.rst
+++ b/doc_src/cmds/fish_greeting.rst
@@ -22,7 +22,7 @@ Description
 
 When an interactive fish starts, it executes fish_greeting and displays its output.
 
-The default fish_greeting is a function that prints a variable of the same name (``$fish_greeting``), so you can also just change that if you just want to change the text.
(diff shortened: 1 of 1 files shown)


faho:
The `ref` for a heading is typically derived from the actual text of it. That's awkward because it changes as soon as you reword it. This is what makes our direct links to changelog entries so weird: https://fishshell.com/docs/current/relnotes.html#fish-3-5-0-released-june-16-2022

So you'd give it a ref manually, like https://github.com/fish-shell/fish-shell/blob/1527edd5428372c4f24de4398f249fd7bc589bd6/doc_src/faq.rst?plain=1#L326-L329


However: In this case, I'd actually go the other way around. Add a full explanation of how to use the variable to cmds/fish_greeting.rst, and then link to that with 

```
:ref:`<cmd-fish_greeting>`
```

from the FAQ.

faho:
I linked it the other way around now, which I feel is the nicer solution. Sorry!

Issue #6114, opened 2019-09-14
test --help emits no output


mqudsi:
With the documentation built and installed, `test --help` emits no output and exits without any errors, setting `$status` to `0`.

floam:
This is by design. `test $x` returns 0 if `x` is not empty. "`--help`" is not empty. It would break compatibility with POSIX test and existing uses in fish.

floam:
Note that all of the switches `test` works on require another argument. `test -f` with no more arguments is not looking for file existence or reporting an error about a missing argument. It's returning 0 because `-f` is not empty.

mqudsi:
Would it really break existing uses, though?

faho:
Yes, it would.

`test` fundamentally sucks, let's not try to paint lipstick on a pig.

faho:
In particular, `test` does not *support* `--`.

So because this necessitates a compatibility break, let's close this. Personally, I'd recommend looking for alternatives to test instead.

Issue #5947, opened 2019-06-19
`test` mis-parse with compound expression involving string comparison where LHS is a flag known to test


lilyball:
`test` takes a number of flags, such as `test -g filename`, but it prioritizes binary operators over interpreting an argument as a flag, e.g. `test -g = -g` is parsed as a string comparison. This is useful when the first arg is actually a variable, e.g. `test "$flag" = -g`.

Unfortunately, this fails once a compound expression is involved. `test -g = -g -o -g = --graph` (which comes from `test "$flag" = -g -o "$flag" = --graph`). In this case it instead parses the `-g` as the unary operator and emits the error

> test: Expected a combining operator like '-a' at index 2

This is true even when using parentheses to group, e.g. `test \( -g = -g \) -o \( -g = --graph \)`.

Reproduced with recent `master` (`fish, version 3.0.2-1297-gb41e5cbbb-dirty`) as well as version `3.0.2`.

lilyball:
Looks like this has been filed before as #5186, but that issue was closed saying this is just an edge case. Closing that was wrong; bash is able to parse this correctly.

faho:
Note that this isn't actually even a violation of POSIX, because:

     >4 arguments:
        The results are unspecified. 

To be perfectly frank, test sucks, and should be replaced. And the replacement should not have combining operators, and most of all should not include a one argument mode.

If anyone cares enough to write a patch for this, we'll merge it.

faho:
Okay, I'm gonna close this as a duplicate of #1337. @lilyball even commented essentially the same example.

Issue #5186, opened 2018-09-14
`test: Expected a combining operator like '-a' at index 2` encountered when using `test` with `-o` to combine two string comparisons


JBlaschke:
Using `test` and `-o` to check string equivalence with two candidates
```fish
set elt "-r"
if test "$elt" = "-r"  -o  "$elt" = "--dependencies"
   echo "hit"
end
```

results in:
```
test: Expected a combining operator like '-a' at index 2
```

### System info
```fish
fish --version
fish, version 2.7.1
```

```fish
uname -a
Darwin MacBert.local 17.7.0 Darwin Kernel Version 17.7.0: Thu Jun 21 22:53:14 PDT 2018; root:xnu-4570.71.2~1/RELEASE_X86_64 x86_64
```

```fish
echo $TERM
xterm-256color
```

JBlaschke:
Great! thank you... it makes sense now!

quackduck:
@zanchey Why doesn't `if test "$argv[1]" = "--update"` throw the same error?

zanchey:
`--update` is not a recognised flag for `test`.

quackduck:
Why would that matter?

PR #9058, opened 2022-07-08
builtin test: Implement -ot, -nt, -ef extensions

### LINKED ISSUE #3589: Extend 'test' with non-POSIX file time comparison extensions

floam:
`-ot` and `-nt` are common beyond-POSIX extensions that test(1) utilities implement, comparing the modification time of two files, as was proposed by myself for fish in #3589 back in 2016. `-ef` is the other one we never got around to: it checks if two files are effectively identical, usually implemented by verifying they share inode and ctime.

test if `file1` is newer than `file2`:
```fish
test file1 -nt file2
```
test if `file1` is older than `file2`:
```fish
test file1 -ot file2
```
test if `file1` and `file2` are the same file:
```fish
test file1 -ef file2
```

## Description

test cases still need to be added and documentation needs to be updated. ~~There might need to be accommodation for `struct stat` platform idiosyncrasies.~~

image

Fixes #3589


floam:
> There might need to be accommodation for struct stat platform idiosyncrasies.

I've changed to using `file_id_t` stuff instead of `stat`ing anything myself, since we already have a pretty exhaustive method by which we try to determine file equality and we we're cooking with gas for modification times on presumably all the platforms already.

floam:
A strange thing: 

```fish
touch -m -t 197001010000.00 epoch
touch -m -t 190112132045.53 old
touch -m -t 190112132045.52 oldest
touch -m -t 203801180314.07 newest
touch norm
```

```fish
$ for x in (ls); echo test $x -nt NONEXIST; test $x -nt NONEXIST && echo true || echo false; end
test epoch -nt NONEXIST true
test newest -nt NONEXIST true
test old -nt NONEXIST false
test oldest -nt NONEXIST false
```

floam:
Ah. Dates earlier than 1970 have a negative time_t, however https://github.com/fish-shell/fish-shell/blob/master/src/wutil.h#L144-L151 was was defaulting to -1: INT_MIN fixes that.

faho:
The implementation here looks okay at first glance.

However, because of our issues with `test`, I'm wary of adding more stuff to it. We constantly need to tell people about quoting problems etc, and so making them more likely to use it doesn't seem like a great idea.

So I'd add this to `path filter` instead:

```fish
# Instead of `test file1 -nt file2`
path filter --newer-than=file2 file1 # and many more files

# Instead of `test file1 -ot file2`
path filter --older-than=file2 file2
```

which can then be combined with other filters - `path filter -rf --newer-than=.timestamp *` to print all readable files in the current directory that are newer than the file ".timestamp".

floam:
I think binary operators are fundamentally less unlikely to cause a surprise if someone accidentally passes an empty argument since we will error out for `test -nt foo` or `test foo -nt` you'd need to actually do it twice. I don't really see the harm.

faho:
Sure, the binary operators are *less* of a problem. But they are still awkward to use - if you do `test $first -nt $second`, what can happen is:

1. $first or $second is empty, so this is an error
2. $first or $second are multiple elements, which would most likely be an error (or could trigger arbitrary conditions)

And even more awkward things like $second is empty and $first is `-n`, but that's admittedly a theoretical problem.

> we can improve test without hurting prior use cases is pretty tangible.

This still sends people to use `test` when we really want to deemphasize that. And anything we add to `test` makes it harder to move away from it.

`path` is also nicer to use with multiple files, nicer to use with multiple conditions and nicer to integrate into a pipeline.

I don't even see an argument for compatibility here, because it's not posix to begin with.

These also aren't very important operators - we've never needed them, and we've not heard many people ask for them in the six years since the feature request was opened. So it's okay to spell them longer than the rather cryptic `-nt`.

(note we *have* needed "is file older than a week", but that's #9057)

floam:
I like `path` - the `path filter` suggestion with dates is interesting. I wasn't suggesting an either or thing, I can see that suggestion particularly useful when trying to kind of whittle down a list of files, the UI is amenable to being able to do stuff like "is it older than x days" or "list the files from 2011 to 2014". That's not how you use `test`.

floam:
Multiple elements are going to be a wrench in a lot of situations - so are empty unquoted variables. These are errors, obviously, but it's tricky to do the right thing and also not be surprising.

floam:
If you have multiple elements in the variable and you're not aware you're doing this, or expecting behavior that isn't how it works, you're going to be surprised in some fashion or another.

```fish
floam@MBP ~/f/f/src (test_time)> set accidentally_two / /dev/null
floam@MBP ~/f/f/src (test_time)> test -d $accidentally_two 
test: unexpected argument at index 3: '/dev/null'
-d / /dev/null
     ^
floam@MBP ~/f/f/src (test_time) [1]> path is --type=dir $accidentally_two 
floam@MBP ~/f/f/src (test_time)> 
```

floam:
image

faho:
>If you have multiple elements in the variable and you're not aware you're doing this, or expecting behavior that isn't how it works, you're going to be surprised in some fashion or another.

`path is` here returned true because at least one of the arguments you gave it was a directory.

`test` blew up instead.

That's not the same.

If you give `path is -d` no further argument, it will be false because you've given it no directory.

If you give `test -d` no further argument, it will be true because `-d` is a non-empty string.

That's not the same.

And sure, `-nt` is a binary operator and doesn't have that latter problem, but it would often be combined with `-d` or `-f` or `-r`, which have. See my example above.

We really want to get away from test. It is bad. Adding more features to it makes it *worse*, not better, because it doesn't solve the problems it has while giving people more reasons/opportunities to use it.

floam:
Look, if we have parts of fish that we are redlining as bad, that we should prefer to ignore, I'd suggest submit a PR or RFC to get rid of it. Do users a favor and let them use a better /bin/[ and unburden the project if that's how it is.

faho:
test having issues isn't new or surprising - see #6124, which starts with the words

>test sucks (https://github.com/fish-shell/fish-shell/issues/1337, https://github.com/fish-shell/fish-shell/issues/5186, https://github.com/fish-shell/fish-shell/issues/5947, https://github.com/fish-shell/fish-shell/issues/6114). Plus I always have to look up the manpage because I can never remember what the options do. Its one saving grace is its compliance with POSIX, so rather than enhancing it beyond that, can we do better?

We now *have* something better for paths, so I'm simply asking that we enhance that.

floam:
The code changes are really quite minimal and I feel pretty good that it was implemented well, #3589 has been open for… 6 years. It's a simple enhancement.

faho (changes requested):
Since my review was specifically requested: My opinion is that this should be added to `path` instead.

The code here seems fine, implementation-wise (tho I'm a bit miffed at the reformatting of the `test` completions to use `complete` without `-c`), but I simply believe this is the wrong place to place the feature.

My opinion is that it should be changed.

faho:
>I feel pretty good that it was implemented well, https://github.com/fish-shell/fish-shell/issues/3589 has been open for… 6 years. 

Yes, and notably that was before `path` was added, so we can now easily do better.

>It's a simple enhancement and the UI doesn't suck.

Let me quote #6124 again:

> test sucks

Is this feature the worst? No, in terms of what `test` can do it's okay.

Except

> I always have to look up the manpage because I can never remember what the options do

Because `test foo -nt bar` is still fairly cryptic. It has a mnemonic ("newer than"), but now there is "-gt", "-lt", "-nt" and "-ot".

But we can now do better than `test`, and that can help us get away from `test`.

share/completions/test.fish
```diff
@@ -1,34 +1,34 @@
...
+complete -c test -k -s L -r -d "File is a symlink"
+complete -c test -k -s O -r -d "File owned by our effective user ID"
+complete -c test -k -s p -r -d "File is a named pipe"
+complete -c test -k -s r -r -d "File is readable"
+complete -c test -k -s s -r -d "File size is non-zero"
+complete -c test -k -s S -r -d "File is a socket"
+complete -c test -k -f -s t -d "FD is a terminal"
+complete -c test -k -s u -r -d "File set-user-ID bit is set"
+complete -c test -k -s w -r -d "File is writable"
+complete -c test -k -s x -r -d "File is executable"
```
floam, line 34: -k is here because these short options are organized better logically in this file than alphabetizing does.

share/completions/test.fish
```diff
@@ -1,34 +1,34 @@
...
```
floam, line 34: -k is here because these short options are organized better logically in this file than alphabetizing does.

floam:
I'd suggest that this is actually simpler to use and understand than adapting this to the `path filter` modality or using the proposed `path mtime` with `math` or whatever, where it'd be more likely to race, and be a fair number of further operations required to wrap your head around. It's not some bike shed I thought up: most `test` utilities implement it.

faho:
> proposed path mtime thing which isn't very well fleshed out yet

`path mtime` is a different thing. It just gives you the mtime as data to do with what you want.

The main usecase is what #9057 already covers - figure out if something is older than a week without resorting to `date` and `stat`, which is rather unstandardized.

It's `stat -c '%Y'` (or `stat -f %m` on macOS), and it's `math (date +%s) - (stat -c %Y)` with the `--relative` option.

It's a building block that you can then do things with.

This is comparing files directly. It's not really easily usable to answer that same question because you would have to have a file that's a week old (meaning you would have to resort to `touch`, and create a file, just to do a comparison with). But if you already have two files? Sure.

I'm not saying both can't exist.

>adapting my approach to the path filter modality

I don't understand what you mean by that.

In terms of user interface, it's quite literally just

```fish
path filter --newer-than=comparefile <files-you-want-to-compare-against-it>
```

And in terms of code, you'd have to add new options `--newer-than` and `--older-than` that take options, then you would have to get the stat buf from those files and compare timestamps in filter_path.

It's probably about the same amount of code as this.

>It's not some bike shed I thought up: most test utilities implement -nt and -ot.

That means that the comparison operation is interesting, not that it needs to be in `test`.

Both bash and zsh also implement an alternative to `test`/`[`.

floam:
> test -f "$file" -a -r "$file" -a "$file" -nt timestampfile

This is basically a worst case scenario, doing a check if something is a file, and also it is readable, and then finally newer than a second file which would imply it exists anyhow - we rarely find it necessary to chain up stuff with a bunch of `-a`'s or multiple `test` commands - but even what you put together here, it's not like some kind of perl enigma. I think we're kind of talking past each other to score points.

floam:
Yes, we do use `-a` once a few times.

floam:
If you look through our codebase, you are going to see most of the date-time stuff we need to laboriously figure out along the lines of `set -l age (math (date +%s) - (stat ...))`. This particular PR doesn't make that kind of stuff extremely easier, and I think maybe it'd be cool if `path` had a way to do **that**.

faho:
>most of the date-time file path stuff we need to laboriously figure out is along the lines of set -l age (math (date +%s) - (stat ...)).

Yes, and that's `path mtime -R`. See #9057, which already replaces those uses:

```diff
- set -l age (math (date +%s) - (stat -c '%Y' $cache_file))
+ set -l age (path mtime -R -- $cache_file)
```

I called that out above.

> You need not worry about this PR cannibalizing much of anything.

I am not. I explicitly said that I'm okay with *comparing* being a separate thing.

I just would prefer to have it in `path` instead of `test`.

krobelus:
Looks like a nice addition.
I think having this functionality in both `test` and `path` is justified because many people already know `test`.
Though it means they might not learn the `path` one.

src/wutil.h
```diff
@@ -145,10 +145,10 @@ struct file_id_t {
...
     ino_t inode{static_cast<ino_t>(-1LL)};
     uint64_t size{static_cast<uint64_t>(-1LL)};
-    time_t change_seconds{-1};
-    long change_nanoseconds{-1};
-    time_t mod_seconds{-1};
-    long mod_nanoseconds{-1};
+    time_t change_seconds{std::numeric_limits<time_t>::min()};
+    unsigned long change_nanoseconds{std::numeric_limits<unsigned long>::min()};
+    time_t mod_seconds{std::numeric_limits<time_t>::min()};
+    unsigned long mod_nanoseconds{std::numeric_limits<unsigned long>::min()};
```
krobelus, line 151: > Fix issue with very very old files
> 
> touch -m -t 197001010000.00 epoch
> touch -m -t 190212112045.40 old
> touch -m -t 190212112045.39 oldest
> touch -m -t 203801080314.07 newest
> 
> Were behaving differently between eachother when used as a left
> hand side file with -nt and then some non-existant file on the
> right hand side.

If we check if both files exist before comparing mtimes, this shouldn't be necessary, right?

Anyway, your solution of giving non-existant files the oldest time possible works fine too and is probably more elegant.

Though I think that making mod_nanoseconds unsigned is wrong because `timespec.tv_nsec` is a signed type.
(Also `std::numeric_limits<unsigned long>::min()` is better written as 0, but I think we should change that anyway.)

src/wutil.cpp
```diff
@@ -862,6 +862,22 @@ static int compare(T a, T b) {
     return 0;
 }
 
+/// this is basically for builtin test
+bool file_id_t::compare_time(const file_id_t &rhs) const {
```
krobelus, line 866: I expect a "compare" method to return one of -1, 0 or 1.
If we return a bool, it should be called "less" or "is_older".

src/builtins/test.cpp
```diff
@@ -742,6 +764,11 @@ static bool binary_primary_evaluate(test_expressions::token_t token, const wcstr
         case test_string_not_equal: {
             return left != right;
         }
```
krobelus, line 766: I guess this means there is not much use for a dedicated function anymore, we can just inline it.

src/wutil.h
floam: Oh you're write about it timespec.tv_nsec - it's `long` here.

zanchey:
> I think having this functionality in both `test` and `path` is justified because many people already know `test`.

I agree

floam:
I'm squashing down the commit history here a bit.

faho:
Okay, `test` and `path` it is then, it appears I have been outvoted.

src/wutil.h
```diff
@@ -145,10 +145,15 @@ struct file_id_t {
     dev_t device{static_cast<dev_t>(-1LL)};
     ino_t inode{static_cast<ino_t>(-1LL)};
     uint64_t size{static_cast<uint64_t>(-1LL)};
```
krobelus, line 147: please don't @ me in commit messages because then I get notifications when any fork pushes that commit

> I'm squashing down the commit history here a bit.

yeah I think that's a good idea, it will make the change easier to review.
Especially since some changes cancel each outer out

src/wutil.h
```diff
@@ -145,10 +145,15 @@ struct file_id_t {
...
-    long change_nanoseconds{-1};
-    time_t mod_seconds{-1};
-    long mod_nanoseconds{-1};
+    /* some platforms handle negative time_t
+       values to represent WW1-era dates, initialize ancient
+       for the sake of comparisons.
+       tv_nsec's meaningful values are REALLY [0, 999999999]
+       this nanosecond component we'll initialize at 0. */
+    time_t change_seconds{std::numeric_limits<time_t>::min()};
+    long change_nanoseconds{0};
```
krobelus, line 154: sorry what I meant is that

- if we use `long`, then `std::numeric_limits<long>::min()` is the obvious "oldest" value
- if we use `unsigned long`, then 0 would be more readable than `std::numeric_limits<unsigned long>::min()`

`long 0` is a weird in-between state (how come it's better than the original -1).
I guess it makes no practical difference, but it would improve the clarity of the change to make it the actual oldest file that we can represent.

floam: I'm using std::numeric_limits<time_t>::min() for the `time_t` - that's a valid "really old" timestamp.
The nsec part is how many additional nanoseconds there are - it's defined between 0 and 1000000. Does INT_MIN make sense there?

krobelus: > how many additional nanoseconds there are - it's defined between 0 and 1000000

I see, then I'd stick to -1 since that's already older than any valid time. 0 is fine too IMO

src/wutil.h
floam: Oh dang I hate it when that happens, sorry about that.

The twists and turns of how this PR was finally implemented are squashed away now.

src/wutil.h
floam: Makes sense.

floam: Those are left at -1 again now.

krobelus (approved):
LGTM!

PR #9060, opened 2022-07-08
Add completion for the "expect" command

Raimondi:
## Description

Add completion for the "expect" command.

Fixes issue #

There is no completion for the "expect" command.


krobelus:
Thanks!

Issue #8746, opened 2022-02-26
readelf has no completion


wuyudi:
fish, version 3.3.1
3.3.1

Linux archlinux 5.16.11-arch1-1 #1 SMP PREEMPT Thu, 24 Feb 2022 02:18:20 +0000 x86_64 GNU/Linux

xterm-256color

`readelf` has no tab completion.

I have run `fish_update_completions` before.

faho:
readelf's man page is awkwardly formatted, which makes it not usable for completion generation.

Someone will have to write a completion script by hand.

amordo:
Hi there! I could give a basic set, relying on `tldr` output (`all, headers, symbols, file-header` options).
Does it make sense?

ridiculousfish:
Sure, go for it.

PR #9070, opened 2022-07-13
Created readelf.fish

kershik:
## Description

Created readelf.fish

Fixes issue #8746


share/completions/readelf.fish
```diff
@@ -0,0 +1,6 @@
+if readelf --version 2>/dev/null >/dev/null
```
floam, line 1: Since we don't actually run `readelf` in the completions, and I'm (assuming we aren't detecting a specific variant of readelf, you can just skip this existence check. We don't try completions for commands that don't exist.

faho:
Okay, no response in a while, closing!

Of course if you get back to us, we'll reopen.

PR #9072, opened 2022-07-13
fix trailing comments in __fish_print_hostnames

drwilly:
/etc/hosts specifies, that everything after a #-character is to be
treated as a comment. The current __fish_print_hostnames however only
considers #-characters at the beginning of a line.
Thus the comment from following valid hosts-entry would end up in the
completion output:

1.2.3.4  myhost # examplecomment

getent hosts properly handles comments.

## Description

Talk about your changes here.

Fixes issue #

## TODOs:


mqudsi:
Because you've dropped comment stripping from the regex: does `getent hosts` ever return a comment?

mqudsi:
LGTM. Merged with thanks.

PR #9074, opened 2022-07-15
Fixed problem where fish would escape '~' when completing an unescaped '['

### LINKED ISSUE #9073: fish prepends back slash when using tilde while autocompleting an escape character
#### Environment
* fish --version 
fish, version 3.5.0

* uname -a 
Darwin Egoist 21.5.0 Darwin Kernel Version 21.5.0: Tue Apr 26 21:08:37 PDT 2022; root:xnu-8020.121.3~4/RELEASE_ARM64_T6000 arm64

* $TERM
xterm-256color
Also happens with Alacritty, Kitty, Terminal.app 

* Issue also affects fish without third party customization

#### Issue

I believe this is similar, yet an edge case of #4570 .  

the backslash bug only occurs when autocompleting something that require escaping without the user escaping it themselves.

to reproduce, run the following commands: 
```
mkdir ~/fish_bug
cd ~/fish_bug
touch '[Hello]World.txt'
vim ~/fish_bug/[ **Tab**           <--- Notice that I deliberately did not escape the bracket.
```
result: vim \\~/fish_bug/\[Hello\]\ World.txt

expected result: vim ~/fish_bug/\[Hello\]\ World.txt

xEgoist:
## Description

 Due to the bracket special case when it comes to completion. The #4570 bug occurs when an unescaped bracket is used in the path.
For example, completing  `~/fish_bug/[` will result in `\~/fish_bug/[` this will cause fish to use an incorrect path. 
As a solution, the `COMPLETE_DONT_ESCAPE_TILDES` was used to ensure we don't escape tildes in the path while completing the unescaped bracket. 

Fixed issue #9073 .

## TODOs:

### DIFF
--- a/src/complete.cpp
+++ b/src/complete.cpp
@@ -1435,6 +1435,8 @@ void completer_t::escape_opening_brackets(const wcstring &argument) {
     for (completion_t &comp : completions.get_list()) {
         if (comp.flags & COMPLETE_REPLACES_TOKEN) continue;
         comp.flags |= COMPLETE_REPLACES_TOKEN;
+        // if path contains tilde and the bracket is unescaped. Don't escape the tilde.
+        comp.flags |= COMPLETE_DONT_ESCAPE_TILDES;
         if (comp.flags & COMPLETE_DONT_ESCAPE) {
             // If the completion won't be escaped, we need to do it here.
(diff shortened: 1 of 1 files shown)


krobelus:
Nice find, thanks! The logic to escape unescaped brackets is quite a hack but this seems like a net improvement.

PR #9083, opened 2022-07-21
Add `zig` completion

sorairolake:
## Description

Add completions for the reference implementation of the the [Zig programming language](https://ziglang.org/).

Fixes issue #

## TODOs:

### DIFF
--- a/share/completions/zig.fish
+++ b/share/completions/zig.fish
@@ -1,0 +1,220 @@
+# Completions for `zig` (https://ziglang.org/)
+# Based on version 0.9.1
+
+# This function is based on the `__fish_complete_clang` function.
+function __fish_complete_zig_cc_and_cpp
+    # If the result is for a value, clang only prints the value, so completions
+    # for `-std=` print `c++11` and not `-std=c++11` like we need. See #4174.
+    set -l prefix (commandline -ct | string replace -fr -- '^(.*=)[^=]*' '$1')
+
+    # first get the completions from clang, with the prefix separated from the value by a comma
+    zig cc --autocomplete=(commandline -ct | string unescape | string replace -- "$prefix" "$prefix,") 2>/dev/null |
+        # and put it in a format that fish understands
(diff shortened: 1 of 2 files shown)


share/completions/zig.fish
```diff
@@ -0,0 +1,282 @@
...
+complete -c zig -n "__fish_seen_subcommand_from $zig_build_generic_commands" -l verbose-link -d "Display linker invocations"
+complete -c zig -n "__fish_seen_subcommand_from $zig_build_generic_commands" -l verbose-cc -d "Display C compiler invocations"
+complete -c zig -n "__fish_seen_subcommand_from $zig_build_generic_commands" -l verbose-air -d "Enable compiler debug output for Zig AIR"
+complete -c zig -n "__fish_seen_subcommand_from $zig_build_generic_commands" -l verbose-mir -d "Enable compiler debug output for Zig MIR"
+complete -c zig -n "__fish_seen_subcommand_from $zig_build_generic_commands" -l verbose-llvm-ir -d "Enable compiler debug output for LLVM IR"
+complete -c zig -n "__fish_seen_subcommand_from $zig_build_generic_commands" -l verbose-cimport -d "Enable compiler debug output for C imports"
+complete -c zig -n "__fish_seen_subcommand_from $zig_build_generic_commands" -l verbose-llvm-cpu-features -d "Enable compiler debug output for LLVM CPU features"
+complete -x -c zig -n "__fish_seen_subcommand_from $zig_build_generic_commands" -l debug-log -d "Enable printing debug/info log messages for scope"
```
krobelus, line 271: unfortunately this disables file completions, can we add something to the `-n` parameter so it doesn't?
I believe that clang completions only trigger if the current token starts with `-`, so maybe that?

Looks really great otherwise!

sorairolake: The argument of `--debug-log` is not files (<https://github.com/ziglang/zig/blob/370793a36b3de34ead8456553a2aa2c56f0d3de8/src/main.zig#L497>). So, this is as expected.

I'll fix the file completion for `zig cc` and `zig c++` to be enabled.

PR #9085, opened 2022-07-22
Fix NetBSD executable path to not use procfs

time-killer-games:
## Description

NetBSD can sometimes give an absolute path but with an odd period and extra slash, for example:

"/path/to/./executable"

to fix this, i would recommend using realpath, or writing your own realpath implementation based on the one NetBSD uses in its source code, but without the PATH_MAX limitation. Let me know what your thoughts are on this.

### DIFF
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -1865,7 +1865,11 @@ std::string get_executable_path(const char *argv0) {
     //
     // (this is broken on NetBSD, while /proc works, so we use that)
     size_t buff_size = sizeof buff;
+    #if defined(__NetBSD__)
+    int name[] = {CTL_KERN, KERN_PROC_ARGS, getpid(), KERN_PROC_PATHNAME};
+    #else
     int name[] = {CTL_KERN, KERN_PROC, KERN_PROC_PATHNAME, -1};
+    #endif
     int result = sysctl(name, sizeof(name) / sizeof(int), buff, &buff_size, nullptr, 0);
     if (result != 0) {
         wperror(L"sysctl KERN_PROC_PATHNAME");
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -1858,12 +1858,10 @@ std::string get_executable_path(const char *argv0) {
     // https://opensource.apple.com/source/adv_cmds/adv_cmds-163/ps/print.c
     uint32_t buffSize = sizeof buff;
     if (_NSGetExecutablePath(buff, &buffSize) == 0) return std::string(buff);
-#elif defined(__BSD__) && defined(KERN_PROC_PATHNAME) && !defined(__NetBSD__)
+#elif defined(__BSD__) && defined(KERN_PROC_PATHNAME)
     // BSDs do not have /proc by default, (although it can be mounted as procfs via the Linux
     // compatibility layer). We can use sysctl instead: per sysctl(3), passing in a process ID of -1
     // returns the value for the current process.
-    //
-    // (this is broken on NetBSD, while /proc works, so we use that)
     size_t buff_size = sizeof buff;
     #if defined(__NetBSD__)
     int name[] = {CTL_KERN, KERN_PROC_ARGS, getpid(), KERN_PROC_PATHNAME};


time-killer-games:
@faho is there some magical reason this was closed without explanation?

faho:
I didn't close it?

This just says "This pull request was closed" to me?

Let me reopen, no idea what happened there.

time-killer-games:
interesting, sorry about that, i don't know what is going on.

faho:
It still shows that message below the last comment at the time?
Screenshot_20220722_123727

WTF is going on here, is github broken right now?

time-killer-games:
As a different pr, I'd like to add a solution for getting the OpenBSD executable path. :) again, my bad, I've come across a lot of wild people on the internet and it kind of keeps me on edge.

time-killer-games:
> It still shows that message below the last comment at the time? Screenshot_20220722_123727
> 
> WTF is going on here, is github broken right now?

I'm getting that too, perhaps a site glitch?

time-killer-games:
its finally gone woohoo

faho:
Okay, so let's go back to the problem at hand.

You don't want to go via procfs, because that's a foreign thing (a plan9-ism, originally, if I'm not mistaken) and might not be available?

And so you use the sysctl path we also use elsewhere instead.

However, this can sometimes print a path that is aesthetically not ideal. It's still a valid, absolute path to the fish executable, but includes superfluous useless components or duplicate (not triplicate) slashes?

If that's the case: Honestly I don't mind. The path *works* for its purpose - which is to figure out where fish is. You might want to look at `normalize_path()` (in wutil.cpp), but even that isn't strictly necessary.

Tbh I'd ask the NetBSD people if this could be fixed, but as far as we're concerned it's a minor cosmetic issue that can just stay. It's certainly not something to write our own realpath over.

time-killer-games:
> Okay, so let's go back to the problem at hand.
> 
> You don't want to go via procfs, because that's a foreign thing (a plan9-ism, originally, if I'm not mistaken) and might not be available?
> 
> And so you use the sysctl path we also use elsewhere instead.
> 
> However, this can sometimes print a path that is aesthetically not ideal. It's still a valid, absolute path to the fish executable, but includes superfluous useless components or duplicate (not triplicate) slashes?
> 
> If that's the case: Honestly I don't mind. The path _works_ for its purpose - which is to figure out where fish is. You might want to look at `normalize_path()` (in wutil.cpp), but even that isn't strictly necessary.
> 
> Tbh I'd ask the NetBSD people if this could be fixed, but as far as we're concerned it's a minor cosmetic issue that can just stay. It's certainly not something to write our own realpath over.

Agreed, rewriting realpath is a bit overkill, if its fine as it stands, then we're on the same page. I just wasn't sure if there was a technical reason this could be problematic. It doesn't always happen, the only time it seems to be reproducible is when you run the exe from the command line using a relative path.

faho:
>It doesn't always happen, the only time it seems to be reproducible is when you run the exe from the command line using a relative path.

Just to be careful: Do you get a *relative* path from this?

Like if you run `../bin/fish`, do you get `../bin/fish` or something like `/usr/local/.//bin/fish`?

Actual relative paths would be problematic, unnormalized absolute paths are okay.

time-killer-games:
It wasn't a relative path, i think NetBSD under the hood fires a call to argv[0] and if the path is relaitve prepend the posix working directory. That would explain why it has an absolute path but not normalized, but it's just speculation. 

I do something similar for my OpenBSD solution. The POSIX working directory is the working directory on application launch, meaning it doesn't change at runtime if the working directory is changed with chdir(). TL;DR yes I am sure it is absolute.

time-killer-games:
You got my curiosity up, im going to run a few tests and see if i can get it to spit out a relative path in any obscure case. Although if i don't find anything within 15 minutes of getting unnormalized absolute paths i think it should be safe to assume we're good here. I'll share screenshots of my results if it helps.

time-killer-games:
Very interesting...

Screenshot_2022-07-22_07-12-32

faho:
Those are still *absolute* paths - they don't depend on the current working directory but start from `/`.

Which makes them usable to us, and I'm okay with leaving the rest up to NetBSD.

Alternatively:

```fish
> path normalize -- /root/../root/../root/xproc/../xproc/./xproc
/root/xproc/xproc
```

That seems to be our `normalize_path` function working as intended. If you want, you can incorporate that. If not I'm okay with merging this as-is.

time-killer-games:
in the first case in the screenshot as you pointed out:

/root/../root/../root/xproc/../xproc/./xproc

... is an absolute path and not the current working directory, because the current working directory is always normalized.

in which case the underlying code is likely using the argv[0] as-is for the executable path, because it is not relative.

However in the second case:

./xproc is argv[0] but it still knew where the exe was despite being relative it still resolved to /root/./xproc

as you might notice /root is the part likely taken from $PWD and that path is then joined with the argv[0] which is ./xproc

thus, "/root" + "/" + "./xproc"

is basically the same as $PWD + "/" + argv[0] underneath. Still speculation, but I'm giving more context to what i meant.

I'll add that function now.

time-killer-games:
i know why this pr was closed now, i accidentally deleted my fork. i'll open a new pr with everything in squash commit

time-killer-games:
Actually i take that back, this can be merged as-is. I don't know how to use wcstring, it appears to be a custom string type you wrote? That's pretty cool, but I'm ready to move on to my OpenBSD pr.

faho:
> I don't know how to use wcstring, it appears to be a custom string type you wrote?

Nah, it's just an alias for std::wstring, which is std::string using wchar_t.

time-killer-games:
I think the more correct solution would be the NetBSD team handles this. We shouldn't have to compensate for an OS-level bug if it doesn't break your code at all. The code logic will just end up getting removed when NetBSD rolls out a fix, anyway.

faho:
Alright, merged, thanks!

PR #9089, opened 2022-07-26
Add Control+R incremental history search in pager

### LINKED ISSUE #602: Add zsh-like reverse-isearch

krobelus:
*This text is wildly outdated, please see the commit messages*

This reimplements [ridiculousfish/control_r](history-search-in-pager) which
is a more future-proof approach than #6686.

I haven't tested this much but it feels good. I wonder though if fuzzy search
(with non-chronological ordering) would be better overall.

Most commits are not very interesting, I'll reintegrate those after a fresh look.
The interesting changes are in the third-to-last and last commits.

A few differences to other shells:

1. Our pager can show multiple results at a time.
2. We use the current token as initial query. This makes it more consistent
with up-arrow search, though I wonder if this is actually useful?
3. We only replace the current token, not the entire commandline. (Does this
make sense?)
4. Other shells allow to use up arrow/down arrow to select adjacent entries
in history. Shouldn't be hard to implement but it might confuse users.
5. In other shells, pressing Return immediately runs the command.
6. We add a space after an accepted completion. This is trivial to change
but the space is harmless since it's trimmed before adding the command
to history.

The pager shows at most 10 history items at a time.  This keeps everything
snappy even with huge histories, though I'm sure there are other solutions
for that.

The UI is a bit arbitrary; we could experiment with using only one column
or allowing multiline commands in pager cells etc.

Closes #602

### DIFF
--- a/src/reader.cpp
+++ b/src/reader.cpp
@@ -3741,7 +3741,7 @@ void reader_data_t::handle_readline_command(readline_cmd_t c, readline_loop_stat
             this->history_pager_history_index_end = 0;
             // Update the pager data.
             pager.set_search_field_shown(true);
-            pager.set_prefix(L"");
+            pager.set_prefix(MB_CUR_MAX > 1 ? L"► " : L"> ", false /* highlight */);
             // Update the search field, which triggers the actual history search.
             insert_string(&pager.search_field_line, command_line.text());
             break;
--- a/src/reader.cpp
+++ b/src/reader.cpp
@@ -2175,8 +2175,6 @@ bool reader_data_t::handle_completions(const completion_list_t &comp, size_t tok
     // Update the pager data.
     pager.set_prefix(prefix);
     pager.set_completions(surviving_completions);
-    // Invalidate our rendering.
-    current_page_rendering = page_rendering_t();
     // Modify the command line to reflect the new pager.
     pager_selection_changed();
     return false;
--- a/src/reader.cpp
+++ b/src/reader.cpp
@@ -335,6 +335,11 @@ void editable_line_t::end_edit_group() {
     }
 }
 
+// Make the search case-insensitive unless we have an uppercase character.
+static history_search_flags_t smartcase_flags(const wcstring &query) {
+    return query == wcstolower(query) ? history_search_ignore_case : 0;
+}
+
 namespace {
 
 /// Encapsulation of the reader's history search functionality.
@@ -506,10 +511,7 @@ class reader_history_search_t {
         match_index_ = 0;
         mode_ = mode;
         token_offset_ = token_offset;
-        history_search_flags_t flags = history_search_no_dedup;
-        // Make the search case-insensitive unless we have an uppercase character.
-        wcstring low = wcstolower(text);
-        if (low == text) flags |= history_search_ignore_case;
+        history_search_flags_t flags = history_search_no_dedup | smartcase_flags(text);
         // We can skip dedup in history_search_t because we do it ourselves in skips_.
         search_ = history_search_t(
             hist, text,
--- a/src/reader.cpp
+++ b/src/reader.cpp
@@ -570,6 +570,7 @@ struct highlight_result_t {
 struct history_pager_result_t {
     completion_list_t matched_commands;
     size_t final_index;
+    bool have_more_results;
 };
 
 /// readline_loop_state_t encapsulates the state used in a readline loop.
@@ -1248,9 +1249,10 @@ static history_pager_result_t history_pager_search(const std::shared_ptr<history
             item.str(), L"", string_fuzzy_match_t::exact_match(),
             COMPLETE_REPLACES_COMMANDLINE | COMPLETE_DONT_ESCAPE | COMPLETE_DONT_SORT});
     }
+    size_t last_index = search.current_index();
     if (direction == history_search_direction_t::forward)
         std::reverse(completions.begin(), completions.end());
-    return {completions, search.current_index()};
+    return {completions, last_index, search.go_to_next_match(direction)};
 }
 
 void reader_data_t::fill_history_pager(bool new_search, history_search_direction_t direction) {
@@ -1283,6 +1285,8 @@ void reader_data_t::fill_history_pager(bool new_search, history_search_direction
                 shared_this->history_pager_history_index_start = index;
                 shared_this->history_pager_history_index_end = result.final_index;
             }
+            shared_this->pager.extra_progress_text =
+                result.have_more_results ? _(L"Search again for more results") : L"";
             shared_this->pager.set_completions(result.matched_commands);
             shared_this->select_completion_in_direction(selection_motion_t::next, true);
             shared_this->super_highlight_me_plenty();
--- a/src/reader.cpp
+++ b/src/reader.cpp
@@ -569,6 +569,7 @@ struct highlight_result_t {
 
 struct history_pager_result_t {
     completion_list_t matched_commands;
+    size_t final_index;
 };
 
 /// readline_loop_state_t encapsulates the state used in a readline loop.
@@ -692,6 +693,9 @@ class reader_data_t : public std::enable_shared_from_this<reader_data_t> {
     reader_history_search_t history_search{};
     /// Whether the in-pager history search is active.
     bool history_pager_active{false};
+    /// The range in history covered by the history pager's current page.
+    size_t history_pager_history_index_start{static_cast<size_t>(-1)};
+    size_t history_pager_history_index_end{static_cast<size_t>(-1)};
 
     /// The selection data. If this is not none, then we have an active selection.
     maybe_t<selection_data_t> selection{};
@@ -749,7 +753,8 @@ class reader_data_t : public std::enable_shared_from_this<reader_data_t> {
     /// Do what we need to do whenever our command line changes.
     void command_line_changed(const editable_line_t *el);
     void maybe_refilter_pager(const editable_line_t *el);
-    void fill_history_pager();
+    void fill_history_pager(bool new_search, history_search_direction_t direction =
+                                                 history_search_direction_t::backward);
 
     /// Do what we need to do whenever our pager selection changes.
     void pager_selection_changed();
@@ -1211,7 +1216,7 @@ void reader_data_t::command_line_changed(const editable_line_t *el) {
         s_generation.store(1 + read_generation_count(), std::memory_order_relaxed);
     } else if (el == &this->pager.search_field_line) {
         if (history_pager_active) {
-            fill_history_pager();
+            fill_history_pager(true, history_search_direction_t::backward);
             return;
         }
         this->pager.refilter_completions();
@@ -1228,31 +1233,56 @@ void reader_data_t::maybe_refilter_pager(const editable_line_t *el) {
 }
 
 static history_pager_result_t history_pager_search(const std::shared_ptr<history_t> &history,
+                                                   history_search_direction_t direction,
+                                                   size_t history_index,
                                                    const wcstring &search_string) {
     // Use a small page size because we don't always offer practical ways to select a item from
     // a large page (since we don't support subsequence filtering here).
     constexpr size_t page_size = 12;
     completion_list_t completions;
     history_search_t search{history, search_string, history_search_type_t::contains,
-                            smartcase_flags(search_string)};
-    while (completions.size() < page_size &&
-           search.go_to_next_match(history_search_direction_t::backward)) {
+                            smartcase_flags(search_string), history_index};
+    while (completions.size() < page_size && search.go_to_next_match(direction)) {
         const history_item_t &item = search.current_item();
         completions.push_back(completion_t{
             item.str(), L"", string_fuzzy_match_t::exact_match(),
             COMPLETE_REPLACES_COMMANDLINE | COMPLETE_DONT_ESCAPE | COMPLETE_DONT_SORT});
     }
-    return {completions};
+    if (direction == history_search_direction_t::forward)
+        std::reverse(completions.begin(), completions.end());
+    return {completions, search.current_index()};
 }
 
-void reader_data_t::fill_history_pager() {
-    auto shared_this = this->shared_from_this();
+void reader_data_t::fill_history_pager(bool new_search, history_search_direction_t direction) {
+    assert(!new_search || direction == history_search_direction_t::backward);
+    size_t index;
+    if (new_search) {
+        index = 0;
+    } else if (direction == history_search_direction_t::forward) {
+        index = history_pager_history_index_start;
+    } else {
+        assert(direction == history_search_direction_t::backward);
+        index = history_pager_history_index_end;
+    }
     const wcstring &search_term = pager.search_field_line.text();
+    auto shared_this = this->shared_from_this();
     debounce_history_pager().perform(
-        [=]() { return history_pager_search(shared_this->history, search_term); },
-        [shared_this, search_term](const history_pager_result_t &result) {
+        [=]() { return history_pager_search(shared_this->history, direction, index, search_term); },
+        [=](const history_pager_result_t &result) {
             if (search_term != shared_this->pager.search_field_line.text())
                 return;  // Stale request.
+            if (result.matched_commands.empty() && !new_search) {
+                // No more matches, keep the existing ones and flash.
+                shared_this->flash();
+                return;
+            }
+            if (direction == history_search_direction_t::forward) {
+                shared_this->history_pager_history_index_start = result.final_index;
+                shared_this->history_pager_history_index_end = index;
+            } else {
+                shared_this->history_pager_history_index_start = index;
+                shared_this->history_pager_history_index_end = result.final_index;
+            }
             shared_this->pager.set_completions(result.matched_commands);
             shared_this->select_completion_in_direction(selection_motion_t::next, true);
             shared_this->super_highlight_me_plenty();
@@ -3376,6 +3406,10 @@ void reader_data_t::handle_readline_command(readline_cmd_t c, readline_loop_stat
             break;
         }
         case rl::pager_toggle_search: {
+            if (history_pager_active) {
+                fill_history_pager(false, history_search_direction_t::forward);
+                break;
+            }
             if (!pager.empty()) {
                 // Toggle search, and begin navigating if we are now searching.
                 bool sfs = pager.is_search_field_shown();
@@ -3694,7 +3728,7 @@ void reader_data_t::handle_readline_command(readline_cmd_t c, readline_loop_stat
         }
         case rl::history_pager: {
             if (history_pager_active) {
-                // TODO Deepen search.
+                fill_history_pager(false, history_search_direction_t::backward);
                 break;
             }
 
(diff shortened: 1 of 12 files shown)


faho:
The reformatting commit here makes review a bit awkward. Could you merge that one first?

src/common.h
```diff
@@ -138,16 +138,19 @@ enum {
...
-// Flags for the escape_string() and escape_string() functions. These are only applicable when the
-// escape style is "script" (i.e., STRING_STYLE_SCRIPT).
+// Flags for the escape_string() function. These are only applicable when the escape style is
+// "script" (i.e., STRING_STYLE_SCRIPT).
 enum {
-    /// Escape all characters, including magic characters like the semicolon.
-    ESCAPE_ALL = 1 << 0,
+    /// Do not escape special fish syntax characters like the semicolon. Only escape nonprintable
+    /// characters and backslashes.
+    ESCAPE_NO_PRINTABLES = 1 << 0,
```
faho, line 146: I would call this "ESCAPE_NONPRINTABLES" or just "ESCAPE_NONPRINT".

"ESCAPE_NO_PRINTABLES" makes it sound like it escapes everything except for printable characters - and a tilde is printable.

faho: With that change, I would be okay with merging everything up to [781a5be](https://github.com/fish-shell/fish-shell/pull/9089/commits/781a5be0baecff88d51f84906c757c5e123cbc2c), so we can concentrate on the pager stuff here.

I'm wary of the change to use the unicode control character symbols - how's the font coverage? - so I would like to check that first.

If possible, that should be extracted to its own PR so we can discuss it separately. A quick check shows that the system I'm on at the moment *has* these symbols, but at the font size in my terminal the `␉` is entirely unreadable and "H T" doesn't scream "tab character" to me (for one because vertical tabs basically aren't a thing and so "horizontal tab" isn't something that really springs to mind).

krobelus: thanks for taking a look

> I would call this "ESCAPE_NONPRINTABLES" or just "ESCAPE_NONPRINT".
>		
>	"ESCAPE_NO_PRINTABLES" makes it sound like it escapes everything except for printable characters - and a tilde is printable.

That sounds right to me - in contrast to the default it does not escape printable characters (except for backslash).
So both ESCAPE_NO_PRINTABLES and ESCAPE_NO_TILDE make us escape fewer characters. We can observe that ESCAPE_NO_PRINTABLES implies ESCAPE_NO_TILDE:

	$ set foo \~\;\1
	$ set -S foo
	$foo: set in global scope, unexported, with 1 elements
	$foo[1]: |~;\ca|

So I'd leave it.

> If possible, that should be extracted to its own PR

definitely

krobelus: I agree that ESCAPE_NONPRINT is more intuitive at first but ESCAPE_NO_PRINTABLES is consistent with the fact that 0=ESCAPE_ALL (and with ESCAPE_NO_TILDE).

We could avoid negations by making the default something like `flags = ESCAPE_NONPRINT | ESCAPE_PRINT`
and then call it like `escape_string(str, ESCAPE_NONPRINT)`.
The problem with that is that it's a bit worse for callers who want other flags like ESCAPE_NO_QUOTED, they'd need `escape_string(str, ESCAPE_ALL | ESCAPE_NO_QUOTED)` where `ESCAPE_ALL = ESCAPE_NONPRINT | ESCAPE_PRINT`.

faho: So ESCAPE_NO_PRINTABLES makes it not escape printable characters?

Ah, okay, I read that the wrong way around then. Fine, ship it - it being everything up to https://github.com/fish-shell/fish-shell/pull/9089/commits/db7b8151ca6b526280c27a16642c79a6c2bdb697.

src/reader.cpp
```diff
@@ -1206,6 +1232,58 @@ void reader_data_t::maybe_refilter_pager(const editable_line_t *el) {
     }
 }
```
krobelus, line 1234: trying to dogfood now with

	function fish_user_key_bindings
	    bind -e --preset \cn
	    bind -e --preset \cp
	end
	set -g fish_autosuggestion_enabled 0

but it's hard getting used to, some details seem off.
I wonder if it should be closer to other shells (execute command immediately on enter, movement keys cancel search) but that seems backwards and would prevent us from using pager features

faho:
> We use the current token as initial query. This makes it more consistent
with up-arrow search, though I wonder if this is actually useful?

> We only replace the current token, not the entire commandline. (Does this
make sense?)

I would expect this to work on the entire *commandline*, by default. So it'll use the commandline as the initial query and replace the commandline.

So if I have `if true` and press ctrl-r, it searches for `if true` (and not just `true` like it does now). That would be consistent with up-arrow.

Token search could be a separate mode.

----

Alright, now for some nitpicks:

1. 10 results seems a bit little - even in a 30x120 terminal you can easily display twice that, without scrolling. I'd probably go for 20-30?
2. If you press ctrl-r again and there are no results, it's a bit awkward to see "(no matches)" with no obvious way to get back. Can we just flash instead?
3. Highlighting in the commandline seems to be off sometimes? Maybe if the commandline becomes longer?
4. If you have a large multiline result and go over it the cursor can get stuck in the commandline instead (reproduced with a looong function and pressing down-arrow i.e. down-or-search)
5. Sometimes it's not clear where one column ends and another begins. For completions you typically have descriptions in between. I'm not sure how to fix this
6. In general it might not be obvious that pressing ctrl-r again shows more results. Can we search for one "page" + 1 and then show "more results available" on the bottom if that's the case?

But overall, this is really cool and slick. Definite improvement over bash's isearch and something I might actually use.

faho:
>Highlighting in the commandline seems to be off sometimes? Maybe if the commandline becomes longer?

Pretty sure we're missing a highlighting pass when filling the pager again - press ctrl-r to search for something, select something and then press ctrl-r again. The highlighting will be off.

src/reader.cpp
```diff
@@ -1206,6 +1233,58 @@ void reader_data_t::maybe_refilter_pager(const editable_line_t *el) {
...
+    }
+    // Paranoid copy to avoid data races.
+    // NOLINTNEXTLINE(performance-unnecessary-copy-initialization)
+    wcstring search_term{pager.search_field_line.text()};
+    auto shared_this = this->shared_from_this();
+    debounce_history_pager().perform(
+        [=]() { return history_pager_search(history, direction, index, search_term); },
+        [shared_this, direction, index](const history_pager_result_t &result) {
+            if (result.search_term != shared_this->pager.search_field_line.text())
+                return;  // Stale request.
```
faho, line 1274:
```suggestion
        [shared_this, direction, index, new_search](const history_pager_result_t &result) {
            if (result.search_term != shared_this->pager.search_field_line.text())
                return;  // Stale request.
                
            if (result.matched_commands.empty() && !new_search) {
                // No more matches, keep the existing ones and flash.
                shared_this->flash();
                return;
            }
```

That should make it flash if the next "page" doesn't exist, but not if it's a new search.

src/reader.cpp
```diff
@@ -3624,6 +3714,32 @@ void reader_data_t::handle_readline_command(readline_cmd_t c, readline_loop_stat
...
+            // Record our cycle_command_line.
+            cycle_command_line = el->text();
+            cycle_cursor_pos = end - buff;
+
+            this->history_pager_active = true;
+            this->history_pager_history_index_start = 0;
+            this->history_pager_history_index_end = 0;
+            // Update the pager data.
+            pager.set_search_field_shown(true);
+            pager.set_prefix(L"");
```
faho, line 3738:
```suggestion
            pager.set_prefix(L"");
            pager.set_fully_disclosed();
```

Discloses immediately.

krobelus:
> I would expect this to work on the entire *commandline*, by default.

agreed, fixed

> 2. If you press ctrl-r again and there are no results, it's a bit awkward to see "(no matches)" with no obvious way to get back. Can we just flash instead?

fixed with your suggestion

> 3. Highlighting in the commandline seems to be off sometimes?

fixed

> 4. If you have a large multiline result and go over it the cursor can get stuck in the commandline instead (reproduced with a looong function and pressing down-arrow i.e. down-or-search)

couldn't reproduce but maybe it's gone now we replace the full commandline

> 5. Sometimes it's not clear where one column ends and another begins. For completions you typically have descriptions in between. I'm not sure how to fix this

Right. The easiest fix would be to place everything in one column but then we lose some screen real estate

> 6. In general it might not be obvious that pressing ctrl-r again shows more results. Can we search for one "page" + 1 and then show "more results available" on the bottom if that's the case?

Good idea, I'll definitely do that if we stick to the current pagination.

krobelus:
> Plus the limit on the number of entries doesn't really prevent expensive searches

True, it's kind of a remnant of a previous implementation.

The current limit is used for two purposes: 1. limit search and 2. limit the page size.

We should decouple those two. A background thread should search everything and then we show small digestible chunks in the pager.

A problem with the current approach is that if there are many similar commands, navigating the pager needs a lot of key presses,  even more so because the history pager does not yet support subsequence matching.
Maybe we should drop the Control-R/Control-S navigation in favor of fuzzy match ranking.

faho:
>couldn't reproduce but maybe it's gone now we replace the full commandline

I can still reproduce. I'm assuming it happens when we don't have room for the pager.

In a terminal of 42 lines, if I do

```fish
commandline -r "$(printf %s\n "echo "(seq 1 40))"
commandline -f execute
```

Then enter `echo `, press ctrl-r, press down until you reach the massive `echo 1` to `echo 40` commandline. If you press ctrl-r or press a character you can page normally, but if you press down you don't select the next entry, but end up flashing the commandline. Pressing up makes you end up in the commandline.

Not that we ever handled the case of having commandlines longer than the terminal particularly well, but with this that's now much easier to reach.

faho:
>Maybe we should drop the Control-R/Control-S navigation in favor of fuzzy match ranking.

I mentioned this before: I abhor fuzzy ranking. The unpredicability of where things end up or are feels terrible.

src/reader.cpp
```diff
@@ -1284,6 +1288,10 @@ void reader_data_t::fill_history_pager(bool new_search, history_search_direction
                 shared_this->history_pager_history_index_start = index;
                 shared_this->history_pager_history_index_end = result.final_index;
             }
+            shared_this->pager.extra_progress_text =
+                (result.can_go_back || result.can_go_further)
+                    ? L"Use Ctrl+R/Ctrl+S to show older/newer results"
```
faho, line 1293: The obvious problem with this is that it won't see rebound keys.

Maybe "trigger search again for more results"? Maybe mention that these keys are the default?

krobelus:
nitpicks 5 and 6 should be mostly done now too

One remaining issue is the glitch when (command line size + pager size) exceeds the screen size.

I'm not yet sure about the page size, I kept it low for now because it's not properly searchable.
(I'm trying page size 12 because that renders nice grids.)
However, there is an argument for showing a whole screenful of commands because that allows visual scanning.
But a screenful has the small disadvantage that it forces the terminal to scroll, so I'm not sure.

src/reader.cpp
krobelus: yeah I think it's fairly unconventional to mention the key names, let's not do that
I just wasn't sure how to mention Ctrl+S as well but it's not that important.

krobelus:
Fixed the down-arrow bug.

A good solution for the page size would probably need a design that's less hacky. Maybe make Control+S spawn a second search field, that only filters pager contents? I also considered making repeated Control+R append to the existing page but that doesn't seem nice.

I'm not super confident in always using set_fully_disclosed() because that might make scenarios with small terminals worse.
I guess the point of that feature is that the pager always fits in 5 (or less) lines. If we disclose everything, it might be 13 lines.
Probably I'm being paranoid. Actually I don't think this is specific to the history pager. If we find a criteria where we can safely use set_fully_disclosed(), we should probably use it in the completion pager as well. Maybe something like `if (pager.rows < $LINES/3) pager.set_fully_disclosed()`.

Overall I think it's usable as-is. 
I got used to having to press Return twice (Control+J comes in handy).
Though I still haven't tested it much.

src/reader.cpp
```diff
@@ -1206,6 +1233,68 @@ void reader_data_t::maybe_refilter_pager(const editable_line_t *el) {
...
+                return;
+            }
+            if (direction == history_search_direction_t::forward) {
+                shared_this->history_pager_history_index_start = result.final_index;
+                shared_this->history_pager_history_index_end = index;
+            } else {
+                shared_this->history_pager_history_index_start = index;
+                shared_this->history_pager_history_index_end = result.final_index;
+            }
+            shared_this->pager.extra_progress_text = (result.can_go_back || result.can_go_further)
```
faho, line 1288: This should use the direction - `((direction == history_search_direction_t::forward && result.can_go_further) || (direction == history_search_direction_t::backward && result.can_go_back))`

Otherwise you'll get it if there is more than one page.

krobelus: Ok, so we don't need to indicate if the opposite direction is available?

krobelus: Yeah that's probably the simplest thing to do for now. In future we can also display the index of the match, and maybe the timestamp (or should that go into the description)

faho:
> Actually I don't think this is specific to the history pager. If we find a criteria where we can safely use set_fully_disclosed(), we should probably use it in the completion pager as well. Maybe something like if (pager.rows < $LINES/3) pager.set_fully_disclosed().

Yeah, the logic here for when to fully disclose immediately seems off - I get a bunch of "and 2 more rows", when there is more than enough space for 2 more rows.

That's a general thing, probably needs a fix in the pager.

>A good solution for the page size would probably need a design that's less hacky. 

Can we get "one screen full" easily? I suppose that requires getting results, finding enough to fill X rows and then discarding the rest?

src/reader.cpp
faho: >In future we can also display the index of the match

Do you mean the index as in "number of history entry"? I don't think that helps much.

Page X/Y would be ideal, but of course that requires figuring out how many pages there are.

>Ok, so we don't need to indicate if the opposite direction is available?

Yeah my idea was that the opposite direction is, after one page, always available.

Of course that doesn't tell people that going back is a thing they can generally do.

>maybe the timestamp (or should that go into the description)

Timestamp descriptions are an interesting idea. They would solve the issue of seeing where entries end.

But it might be overkill, especially if you do full "YYYY-MM-DD HH:MM:SS" everywhere?

faho: Oooh, how about "Page 4 (search again for more)"?

krobelus:
> Can we get "one screenful" easily?

Yeah we could get a large enough number (say 1000) but I assume it's bad UI to render completions that the user can't select very easily.
They can try to extend their query, but I believe that's inconvenient if there are many similar commands. Just an assumption, I'd need to try it in practice.
zsh supports globs in their incremental search, we could copy that to make it easier. We can even prefill the search field with `*current-commandline*` (with the cursor placed second-to-last) to make this discoverable. Though I'd prefer plain subsequence match, with no need for globs.

If we stick with the current approach I'll implement something that gives roughly 0.8 * a screenful or so.
Actually just finding *all* the matches wouldn't be much different, plus it removes the need for extending searches.

I just realized that since editing the search field after extending the search (with Control+R) makes the search starts from scratch, refining the search doesn't always work.

So I'm happy to get rid of search-extending anyway. I think we need some more pager improvements to make it a nice experience.

src/reader.cpp
```diff
@@ -3624,6 +3731,28 @@ void reader_data_t::handle_readline_command(readline_cmd_t c, readline_loop_stat
...
+            // Record our cycle_command_line.
+            cycle_command_line = command_line.text();
+            cycle_cursor_pos = command_line.position();
+
+            this->history_pager_active = true;
+            this->history_pager_history_index_start = 0;
+            this->history_pager_history_index_end = 0;
+            // Update the pager data.
+            pager.set_search_field_shown(true);
+            pager.set_prefix(L"$ ", false /* highlight */);
```
faho, line 3749: One weird nitpick: We typically use ">" to mark the prompt - in our docs, in our default prompt, ...

We could use that here - or see if we have a multibyte locale and use `►`?

faho: `>` looks fine here, but is of course confusable for a redirection. On the other hand the `>` will all line up, so it's easy enough to ignore a column of `>`.

Screenshot_20220728_213251

krobelus: sure, I like `>` better than `$` because it's smaller. Let's try the multibyte one too, we'll see if anyone complains

faho:
As far as I'm concerned this is in a state where it could be merged to possibly get some more feedback. It's also not a major issue to change things later (this is not a script feature so there aren't a lot of compatibility concerns).

Things that could be done:

1. Add color - for the prefix character, possibly configurable, for the search match (use color_search_match). Maybe full highlighting but that sounds expensive.
2. The prefix character *could* be made configurable. I don't know if that's needed.
3. Figure out the count and how paging works
4. A way to filter more, maybe regex, globs,...
5. A feature where you press a key chord to delete the selected history entry - might be neat as a function, just kinda needs refreshing?
6. ...

But none of those seem like solid "must"s to me, this appears to be usable as-is.

So if you agree, feel free to merge at your discretion. LGTM!

krobelus:
Agreed on all points. Will merge tomorrow.

Would be nice to have a plan for a better paging UI but it's not essential.

> Maybe full highlighting but that sounds expensive.

Nah it should be fine. I think the only obstacle is that reverse video and syntax highlighting don't play nice, but we can just exclude the selected entry

src/reader.cpp
floam: Could try to highlight these as commands, I think might be the most recognizable way.

krobelus: Yep, I think that's the next thing on my list

src/reader.cpp
```diff
@@ -1206,6 +1233,67 @@ void reader_data_t::maybe_refilter_pager(const editable_line_t *el) {
     }
 }
 
+static history_pager_result_t history_pager_search(const std::shared_ptr<history_t> &history,
+                                                   history_search_direction_t direction,
+                                                   size_t history_index,
+                                                   const wcstring &search_string) {
+    // Use a small page size because we don't always offer practical ways to select a item from
+    // a large page (since we don't support subsequence filtering here).
+    constexpr size_t page_size = 12;
```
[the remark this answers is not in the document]
faho: We don't really do options like that.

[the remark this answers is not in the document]
faho: #9458

PR #9092, opened 2022-07-28
Executable path fixes and more

time-killer-games:
Uses a "better" method to get executable path on NetBSD (use `-1` instead of `getpid()`), and as we recently discussed in my issue I opened, OpenBSD as well. To get it to build on OpenBSD, use of `to_string()` clang considered ambiguous  and threw errors everywhere your `to_string()` implementation was used. I'm curious, why not just use `std::to_wstring()`? Only way I could build it. If you want, an alternative could be make an alias for `std::to_wstring()` called `to_wcstring()` like what you are already doing with `std::wstring` to use `wcstring`. 

If you want me to not conflate unrelated issues, I can split up this pull request.

### DIFF
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -1867,9 +1867,8 @@ std::string get_executable_path(const char *argv0) {
     // https://opensource.apple.com/source/adv_cmds/adv_cmds-163/ps/print.c
     uint32_t buff_size = sizeof buff;
     if (_NSGetExecutablePath(buff, &buff_size) == 0) return std::string(buff);
-#elif defined(__BSD__) && defined(KERN_PROC_PATHNAME)
 #else
-    size_t buff_size = sizeof buff;
+size_t buff_size = sizeof buff;
 #if defined(__BSD__) && defined(KERN_PROC_PATHNAME)
     // BSDs do not have /proc by default, (although it can be mounted as procfs via the Linux
     // compatibility layer). We can use sysctl instead: per sysctl(3), passing in a process ID of -1
@@ -1886,19 +1885,19 @@ std::string get_executable_path(const char *argv0) {
         return std::string(buff);
     }
 #elif defined(__OpenBSD__)
+    static std::string result;
+    if (!result.empty()) return result;
     int cntp = 0;
     struct stat st;
-    std::string path;
     bool ok = false;
-    char exe[buff_size];
+    std::string path;
     kinfo_file *kif = nullptr;
     static kvm_t *kd = nullptr;
     char errbuf[_POSIX2_LINE_MAX];
-    const char *pwd = nullptr, *cwd = nullptr, *penv = nullptr;
+    const char *pwd = nullptr, *penv = nullptr;
     if (!std::string(argv0).empty()) {
         if (argv0[0] == '/') {
             path = argv0;
-            goto finish1;
         } else if (std::string(argv0).find('/') == std::string::npos) {
             penv = getenv("PATH");
             if (penv && *penv) {
@@ -1907,42 +1906,20 @@ std::string get_executable_path(const char *argv0) {
                 for (std::size_t i = 0; i < env.size(); i++) {
                     path = wcs2string(env[i].c_str(), env[i].length()) + "/" + argv0;
                     if (!stat(path.c_str(), &st) && (st.st_mode & S_IXUSR) && (st.st_mode & S_IFREG)) {
-                        goto finish2;
+                        break;
                     }
-                    path.clear();
                 }
-            }
-        }
-        pwd = getenv("PWD");
-        if (pwd && *pwd) {
-            path = std::string(pwd) + "/" + std::string(argv0);
-            if (!stat(path.c_str(), &st) && (st.st_mode & S_IXUSR) && (st.st_mode & S_IFREG)) {
-                goto finish2;
-            } else {
-                goto fallback;
             }
         } else {
-            fallback:
-            cwd = getcwd(exe, buff_size);
-            if (cwd && *cwd) {
-                path = std::string(cwd) + "/" + std::string(argv0);
-            }
-        }
-        finish1:
-        if (!path.empty()) {
-            if (!stat(path.c_str(), &st) && (st.st_mode & S_IXUSR) && (st.st_mode & S_IFREG)) {
-                finish2:
-                if (realpath(path.c_str(), exe)) {
-                    path = exe;
-                    goto finish3;
-                }
+            pwd = getenv("PWD");
+            if (pwd && *pwd) {
+                path = std::string(pwd) + "/" + std::string(argv0);
             }
         }
     }
-    finish3:
-    if (path.empty()) goto finish4;
+    if (path.empty()) { result = "fish"; return result; }
     kd = kvm_openfiles(nullptr, nullptr, nullptr, KVM_NO_FILES, errbuf); 
-    if (!kd) { path.clear(); goto finish4; }
+    if (!kd) { result = "fish"; return result; }
     if ((kif = kvm_getfiles(kd, KERN_FILE_BYPID, getpid(), sizeof(struct kinfo_file), &cntp))) {
         for (int i = 0; i < cntp; i++) {
             if (kif[i].fd_fd == KERN_FILE_TEXT) {
@@ -1955,14 +1932,15 @@ std::string get_executable_path(const char *argv0) {
             }
         }
     }
-    if (!ok) { path.clear(); }
+    if (!ok) path.clear();
     kvm_close(kd);
-    finish4:
     path.resize(buff_size, '\0');
     path[buff_size] = '\0';
-    strcpy(buff, path.c_str());
-    static std::string result = buff;
-    return (!result.empty()) ? result : "fish";
+    if (realpath(path.c_str(), buff)) {
+        path = buff;
+    }
+    result = (!path.empty()) ? path : "fish";
+    return result;
 #else
     // On other unixes, fall back to the Linux-ish /proc/ directory
     ssize_t len;
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -1880,21 +1880,42 @@ size_t buff_size = sizeof buff;
     } else {
         return std::string(buff);
     }
-#elif defined(__OpenBSD__)
+#elif (defined(__linux__) || defined(sun) || defined(__sun))
+    ssize_t len;
+    len = readlink("/proc/self/exe", buff, sizeof buff - 1);  // Linux
+    if (len == -1) {
+        len = readlink("/proc/self/path/a.out", buff, sizeof buff - 1);  // Solaris
+    }
+    if (len > 0) {
+        buff[len] = '\0';
+        // When /proc/self/exe points to a file that was deleted (or overwritten on update!)
+        // then linux adds a " (deleted)" suffix.
+        // If that's not a valid path, let's remove that awkward suffix.
+        std::string buffstr = {buff};
+        if (access(buff, F_OK)) {
+            auto dellen = const_strlen(" (deleted)");
+            if (buffstr.size() > dellen &&
+                buffstr.compare(buffstr.size() - dellen, dellen, " (deleted)") == 0) {
+                buffstr = buffstr.substr(0, buffstr.size() - dellen);
+            }
+        }
+        return buffstr;
+    }
+#else
     static std::string result;
     if (!result.empty()) return result;
-    std::string strargv0 = ((argv0) ? argv0 : ""), path;
+    std::string strargv0 = ((argv0) ? argv0 : ""), buffstr;
     struct stat st;
     if (!strargv0.empty()) {
         if (strargv0[0] == '/') {
-            path = strargv0;
-        } else if (path.find('/') == std::string::npos) {
+            buffstr = strargv0;
+        } else if (buffstr.find('/') == std::string::npos) {
             const char *penv = getenv("PATH");
             if (penv && *penv) {
                 std::vector<std::string> env = split_string(penv, ':');
                 for (std::size_t i = 0; i < env.size(); i++) {
-                    path = env[i] + "/" + strargv0;
-                    if (!stat(path.c_str(), &st) && (st.st_mode & S_IXUSR) && (st.st_mode & S_IFREG)) {
+                    buffstr = env[i] + "/" + strargv0;
+                    if (!stat(buffstr.c_str(), &st) && (st.st_mode & S_IXUSR) && (st.st_mode & S_IFREG)) {
                         break;
                     }
                 }
@@ -1902,10 +1923,13 @@ size_t buff_size = sizeof buff;
         } else {
             const char *pwd = getenv("PWD");
             if (pwd && *pwd) {
-                path = std::string(pwd) + "/" + strargv0;
+                buffstr = std::string(pwd) + "/" + strargv0;
             }
         }
+        return buffstr;
     }
+#ifdef __OpenBSD__
+    // kd variable is necessary for kvm_getfiles() first argument; RTFM.
     kvm_t *kd = kvm_openfiles(nullptr, nullptr, nullptr, KVM_NO_FILES, nullptr);
     if (!kd) { result = "fish"; return result; }
     int cntp = 0;
@@ -1914,7 +1938,7 @@ size_t buff_size = sizeof buff;
     if ((kif = kvm_getfiles(kd, KERN_FILE_BYPID, getpid(), sizeof(struct kinfo_file), &cntp))) {
         for (int i = 0; i < cntp; i++) {
             if (kif[i].fd_fd == KERN_FILE_TEXT) {
-                if (!stat(path.c_str(), &st)) {
+                if (!stat(buffstr.c_str(), &st)) {
                     if (st.st_dev == (dev_t)kif[i].va_fsid || st.st_ino == (ino_t)kif[i].va_fileid) {
                         ok = true;
                         break;
@@ -1925,38 +1949,14 @@ size_t buff_size = sizeof buff;
     }
     kvm_close(kd);
     if (!ok) { result = "fish"; return result; }
-    path.resize(buff_size, '\0');
-    path[buff_size] = '\0';
-    if (realpath(path.c_str(), buff)) {
-        path = buff;
+#endif
+    buffstr.resize(buff_size, '\0');
+    buffstr[buff_size] = '\0';
+    if (realpath(buffstr.c_str(), buff)) {
+        buffstr = buff;
     }
-    result = path;
+    result = buffstr;
     return result;
-#else
-    // On other unixes, fall back to the Linux-ish /proc/ directory
-    ssize_t len;
-    len = readlink("/proc/self/exe", buff, sizeof buff - 1);  // Linux
-    if (len == -1) {
-        len = readlink("/proc/curproc/file", buff, sizeof buff - 1);  // other BSDs
-        if (len == -1) {
-            len = readlink("/proc/self/path/a.out", buff, sizeof buff - 1);  // Solaris
-        }
-    }
-    if (len > 0) {
-        buff[len] = '\0';
-        // When /proc/self/exe points to a file that was deleted (or overwritten on update!)
-        // then linux adds a " (deleted)" suffix.
-        // If that's not a valid path, let's remove that awkward suffix.
-        std::string buffstr{buff};
-        if (access(buff, F_OK)) {
-            auto dellen = const_strlen(" (deleted)");
-            if (buffstr.size() > dellen &&
-                buffstr.compare(buffstr.size() - dellen, dellen, " (deleted)") == 0) {
-                buffstr = buffstr.substr(0, buffstr.size() - dellen);
-            }
-        }
-        return buffstr;
-    }
 #endif
 #endif
 
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -1,10 +1,6 @@
 // Various functions, mostly string utilities, that are used by most parts of fish.
 #include "config.h"
 
-#ifdef __OpenBSD__
-#include "path.h"
-#endif
-
 #ifdef HAVE_BACKTRACE_SYMBOLS
 #include <cxxabi.h>
 #endif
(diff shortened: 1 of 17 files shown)


src/common.cpp
```diff
@@ -1849,21 +1860,23 @@ bool valid_func_name(const wcstring &str) {
...
-    // This is basically grabbing exec_path after argc, argv, envp, ...: for us
+    // Note: _NSGetExecutablePath() is proprietary and might return a symlink.
+    // Use the open source Darwin / XNU kernel proc_pidpath() function instead.
+    // This is basically grabbing exec_path after argc, argv, envp, ...: for us.
     // https://opensource.apple.com/source/adv_cmds/adv_cmds-163/ps/print.c
-    uint32_t buffSize = sizeof buff;
-    if (_NSGetExecutablePath(buff, &buffSize) == 0) return std::string(buff);
-#elif defined(__BSD__) && defined(KERN_PROC_PATHNAME)
+    // The buffer must have the exact size of macro PROC_PIDPATHINFO_MAXSIZE.
+    char buff[PROC_PIDPATHINFO_MAXSIZE]; uint32_t buff_size = sizeof buff;
```
floam, line 1869: do these on separate lines

ridiculousfish:
Thank you for opening this, I commend your boldness and I recognize how much work was invested in this.

The short answer to your question of "why not std::to_wstring" is that this stuff was written before fish adopted C++11 so the function didn't exist when the code was written. But even today I'm skeptical of `std::to_wstring` because its printf-based implementation  is [rather gross](https://github.com/gcc-mirror/gcc/blob/16e2427f50c208dfe07d07f18009969502c25dc8/libstdc%2B%2B-v3/include/ext/vstring.h#L2840) and there's always a risk of an unwanted locale dependency. fish's simple `(val % 10) + '0'` loop does precisely what we want, and it's several times faster than libstdc++; I think replacing it risks more problems than it would solve. So I want to decline adopting `std::to_wstring`.

The naming issue is a good catch and renaming fish's function to `to_wcstring` seems perfect; I'd be happy to merge that.

Re: finding executable paths, the reason fish does this is for locating resources like docs relative to its *runtime* executable path, with the idea that the user may copy a directory tree around and fish will keep working. This perhaps only ever made sense on the Mac; but the use of `_NSGetExecutablePath` has been there forever and I am loathe to change it without a clear benefit. If you don't care about dynamically-relocatable fish on OpenBSD, it's fine to just return nothing from the function and fish will fall back to its build-time or install-time paths.

Overall I'm excited by the prospect of fish working better with OpenBSD, it's very cool.

faho:
>Re: finding executable paths, the reason fish does this is for locating resources like docs relative to its runtime executable path, 

That's not the only place we use this, it's also used to power `status fish-path`, for users to reexecute the same fish.

floam:
If we are going to get rid of `_NSGetExecutablePath`, I'd rather that be us just doing it ourselves and copying the data after `envp` for bragging rights - I don't see any advantage using libproc here.

src/common.cpp
```diff
@@ -1873,6 +1887,40 @@ std::string get_executable_path(const char *argv0) {
     } else {
         return std::string(buff);
     }
+#elif defined(__OpenBSD__)
+    bool ok = false;
+    std::string narrow;
+    static kvm_t *kd = nullptr;
+    wcstring wargv0 = str2wcstring(argv0);
+    // not sure if I am doing this right??
+    null_environment_t vars = null_environment_t();
```
faho, line 1896: This is an *empty* environment - no variables will be set.

That means `path_get_path` won't look through $PATH.

You need to either get an actual environment, which means passing it from the outside (if you have a parser_t you can use parser_t::vars()), or you need to adjust `path_get_path` so that you can pass a $PATH value in instead of a full environment_t, or you need to redo path_get_path's logic with the normal `getenv("PATH")` $PATH.

I don't think former is possible because this is called really early in startup before the parser is initialized. This would require moving the parser initialization up, and I'm not sure that works before the rest of the init is done.

src/env.cpp
```diff
@@ -336,7 +336,7 @@ void env_init(const struct config_paths_t *paths, bool do_uvars, bool default_pa
 
     // Set $USER, $HOME and $EUID
     // This involves going to passwd and stuff.
-    vars.set_one(L"EUID", ENV_GLOBAL, to_string(static_cast<unsigned long long>(geteuid())));
+    vars.set_one(L"EUID", ENV_GLOBAL, std::to_wstring(static_cast<unsigned long long>(geteuid())));
```
faho, line 339: As a general rule, please don't add unrelated cleanup to a PR.

This has nothing to do with the matter at hand and complicates review.

(of course if it's a line or two in the places you are *already* touching that's fine, but that's not the case here)

floam:
Is `status fish-path` a real thing people are using?

faho:
It is a thing we offer.

time-killer-games:
> If we are going to get rid of `_NSGetExecutablePath`, I'd rather that be us just doing it ourselves and copying the data [after `envp`](https://craftware.xyz/tips/Apple-hidden-parameter.html) for bragging rights - I don't see any advantage using libproc headers here. I'm giving a thumbs down to that part.

I would prefer this as well. How do you @ridiculousfish feel about it?

time-killer-games:
I tried replacing all instances of std::to_wstring() with to_wcstring() via:
```
inline wcstring to_wcstring(long x) {
    wchar_t buff[64];
    format_long_safe(buff, x);
    return wcstring(buff);
}

inline wcstring to_wcstring(unsigned long long x) {
    wchar_t buff[64];
    format_ullong_safe(buff, x);
    return wcstring(buff);
}

inline wcstring to_wcstring(int x) { return to_wcstring(static_cast<long>(x)); }

inline wcstring to_wcstring(size_t x) { return to_wcstring(static_cast<unsigned long long>(x)); }
```

And the "ambiguous" errors came back.

faho:
> I tried replacing all instances of std::to_wstring() with to_wcstring() via:

Please leave that out of this PR. It has nothing to do with the rest.

time-killer-games:
> > I tried replacing all instances of std::to_wstring() with to_wcstring() via:
> 
> Please leave that out of this PR. It has nothing to do with the rest.

That's fine. i will at least need it locally for testing because OpenBSD won't even build without it.

time-killer-games:
`status fish-path` is showing correctly on OpenBSD now. I decided to revert the macOS solution. 

Will split up the pull request if/when the CI passes.

src/common.cpp
```diff
@@ -1888,39 +1886,83 @@ std::string get_executable_path(const char *argv0) {
...
-                                ok = true;
-                                break;
-                            }
-                        }
+    char errbuf[_POSIX2_LINE_MAX];
+    const char *pwd = nullptr, *cwd = nullptr, *penv = nullptr;
+    if (!std::string(argv0).empty()) {
+        if (argv0[0] == '/') {
+            path = argv0;
+            goto finish1;
```
faho, line 1901: Please don't use goto.

Instead, keep the path as an std::string - you're already converting it to and fro twice in many cases.

src/common.cpp
```diff
@@ -1888,39 +1886,83 @@ std::string get_executable_path(const char *argv0) {
...
-    narrow[buff_size] = '\0';
-    strcpy(buff, narrow.substr(0, buff_size).c_str());
-    return ((ok) ? std::string(buff)  : "fish");
+    if (!ok) { path.clear(); }
+    kvm_close(kd);
+    finish4:
+    path.resize(buff_size, '\0');
+    path[buff_size] = '\0';
+    strcpy(buff, path.c_str());
+    static std::string result = buff;
```
faho, line 1964: Defining a static variable this late means you do all the computation anyway on the second go and then throw it all away because the assignment never happens.

faho: It's *fine* to use a static variable.

But what you would do is:

```fish
static std::string ret;

if (!ret.empty()) return ret;

// And now the rest of the function
```

That way you skip *all of the other code*.

faho: Like you add the static variable at the top of the function, before everything else.

If it has something in it you're in the second run and can just return.

time-killer-games: I added a static bool so that it only calculates the path once for each run of fish. Not what you asked, let me know if that's ok, it sould be more efficient than runnting multiple times that code logic.

time-killer-games: Ah. I didn't know static worked like that. I'm going to change it to your suggestion.

time-killer-games:
Ok, I got OpenBSD building with `to_string()` renamed to `to_wcstring()`. Is it ok if I split up the pull request now? Or do you guys want to share any more critiques before I do that?

src/common.cpp
```diff
@@ -1,6 +1,10 @@
 // Various functions, mostly string utilities, that are used by most parts of fish.
 #include "config.h"
 
+#ifdef __OpenBSD__
+#include "path.h"
```
faho, line 5: This should be in between the other `#include ".."`s, in alphabetical order.

src/common.cpp
```diff
@@ -1873,6 +1884,63 @@ std::string get_executable_path(const char *argv0) {
...
+    if (!result.empty()) return result;
+    int cntp = 0;
+    struct stat st;
+    bool ok = false;
+    std::string path;
+    kinfo_file *kif = nullptr;
+    static kvm_t *kd = nullptr;
+    char errbuf[_POSIX2_LINE_MAX];
+    const char *pwd = nullptr, *penv = nullptr;
+    if (!std::string(argv0).empty()) {
```
faho, line 1898: Like I said, argv0 should just be saved as a string, so you don't have to reconstruct it multiple times.

E.g.

```c++
std::string strargv0 = argv0;

if (strargv0.empty()) {

// ...

        if (strargv0[0] == '/') {
            path = argv0;
        } else if (strargv0.find('/') == std::string::npos) {

File: src/common.cpp [hunk 8]
```diff
@@ -1873,6 +1884,63 @@ std::string get_executable_path(const char *argv0) {
...
+    static std::string result;
+    if (!result.empty()) return result;
+    int cntp = 0;
+    struct stat st;
+    bool ok = false;
+    std::string path;
+    kinfo_file *kif = nullptr;
+    static kvm_t *kd = nullptr;
+    char errbuf[_POSIX2_LINE_MAX];
+    const char *pwd = nullptr, *penv = nullptr;
```
faho, line 1897: You only need these in a small block, you can define them there:

```c++
            const char *penv = getenv("PATH");
```

faho: Same goes for the other variables - no need to define `kinfo_file *kif` twenty lines before you need it. This isn't C89, you can define variables later.

src/common.cpp
```diff
@@ -1873,6 +1884,63 @@ std::string get_executable_path(const char *argv0) {
...
     }
+#elif defined(__OpenBSD__)
+    static std::string result;
+    if (!result.empty()) return result;
+    int cntp = 0;
+    struct stat st;
+    bool ok = false;
+    std::string path;
+    kinfo_file *kif = nullptr;
+    static kvm_t *kd = nullptr;
```
faho, line 1895: Why is this static?

src/common.cpp
```diff
@@ -1873,6 +1884,63 @@ std::string get_executable_path(const char *argv0) {
...
+            }
+        } else {
+            pwd = getenv("PWD");
+            if (pwd && *pwd) {
+                path = std::string(pwd) + "/" + std::string(argv0);
+            }
+        }
+    }
+    if (path.empty()) { result = "fish"; return result; }
+    kd = kvm_openfiles(nullptr, nullptr, nullptr, KVM_NO_FILES, errbuf);
```
faho, line 1921: You don't use errbuf, can you just pass a nullptr?

src/common.cpp
```diff
@@ -1873,6 +1884,63 @@ std::string get_executable_path(const char *argv0) {
...
+            pwd = getenv("PWD");
+            if (pwd && *pwd) {
+                path = std::string(pwd) + "/" + std::string(argv0);
+            }
+        }
+    }
+    if (path.empty()) { result = "fish"; return result; }
+    kd = kvm_openfiles(nullptr, nullptr, nullptr, KVM_NO_FILES, errbuf); 
+    if (!kd) { result = "fish"; return result; }
+    if ((kif = kvm_getfiles(kd, KERN_FILE_BYPID, getpid(), sizeof(struct kinfo_file), &cntp))) {
```
faho, line 1923: I believe this is the only place you use kif, so you should be able to do

```c++
if ((kinfo_file *kif = kvm_getfiles(...)) {
```

src/wcstringutil.h
```diff
@@ -133,21 +133,26 @@ wcstring_list_t split_string_tok(const wcstring &val, const wcstring &seps,
...
     wchar_t buff[64];
     format_ullong_safe(buff, x);
     return wcstring(buff);
 }
 
-inline wcstring to_string(int x) { return to_string(static_cast<long>(x)); }
-
-inline wcstring to_string(size_t x) { return to_string(static_cast<unsigned long long>(x)); }
+// prevents 'ambiguous' compiler error in OpenBSD clang
+#ifndef __OpenBSD__
```
faho, line 152: No need to have two `#ifndef __OpenBSD__` blocks, just put these above into the other one.

src/common.cpp
```diff
@@ -1873,6 +1884,63 @@ std::string get_executable_path(const char *argv0) {
...
+                }
+            }
+        } else {
+            pwd = getenv("PWD");
+            if (pwd && *pwd) {
+                path = std::string(pwd) + "/" + std::string(argv0);
+            }
+        }
+    }
+    if (path.empty()) { result = "fish"; return result; }
```
faho, line 1920: As a general question: When does this fail/succeed?

What is the kvm_openfiles block used for? In which cases does that happen?

Your answer should probably be added as a comment.

src/common.cpp
```diff
@@ -1873,6 +1884,63 @@ std::string get_executable_path(const char *argv0) {
     } else {
         return std::string(buff);
     }
+#elif defined(__OpenBSD__)
```
faho, line 1887: Note: The $PATH path is useful for *all* systems that don't have a nicer way of figuring this out.

That bit shouldn't be hidden behind `__OpenBSD__`.

src/common.cpp
```diff
@@ -1873,6 +1884,63 @@ std::string get_executable_path(const char *argv0) {
...
+    char errbuf[_POSIX2_LINE_MAX];
+    const char *pwd = nullptr, *penv = nullptr;
+    if (!std::string(argv0).empty()) {
+        if (argv0[0] == '/') {
+            path = argv0;
+        } else if (std::string(argv0).find('/') == std::string::npos) {
+            penv = getenv("PATH");
+            if (penv && *penv) {
+                wcstring wenv = str2wcstring(penv);
+                wcstring_list_t env = split_string(wenv, L':');
```
faho, line 1905: (note: An overload for `split_string` that handles std::string would be useful here, so you can save on converting from a string to a wcstring and back)

faho (changes requested):

src/common.cpp
time-killer-games: Because the documentation says:

https://man.openbsd.org/kvm_open.3

"The kvm_open() and kvm_openfiles() functions both return a descriptor to be used in all subsequent kvm library calls. The library is fully re-entrant."

Normally this would be a global variable, but a static local should be faster and prefered in this case because it is the only place we are using the kvm_t * return value.

faho: We use this *once*, in this function, because we already have the static guard variable above.

Making this static as well just wastes memory.

faho: I.e. you already *have* the variable:

```c++
kd = kvm_openfiles(nullptr, nullptr, nullptr, KVM_NO_FILES, errbuf); 
```

This sets a local variable. And you then simply pass it to the next function:

```c++
if ((kif = kvm_getfiles(kd, KERN_FILE_BYPID, getpid(), sizeof(struct kinfo_file), &cntp)))
```

And because that code sets our static "result", it is run *once*.

Making kd `static` just means that it's kept around for the entire runtime of `fish`.

time-killer-games: All I know is everywhere in OpenBSD's source code that uses a kvm_t * variable they have it global or static. I'm sure there's a reason for this. I can do it if you really want me to, but this seems incorrect.

time-killer-games: Perhaps you are right about this, because I can't find anywhere they are doing it once like I am here. I'll take your word for it.

time-killer-games: I understand now, my bad.

src/common.cpp
time-killer-games: This works:
```
    kinfo_file *kif = nullptr;
    if ((kif = kvm_getfiles(kd, KERN_FILE_BYPID, getpid(), sizeof(struct kinfo_file), &cntp))) {
```

This gave me errors with the kif variable:
```
    if ((kinfo_file *kif = kvm_getfiles(kd, KERN_FILE_BYPID, getpid(), sizeof(struct kinfo_file), &cntp))) {
```

This gave the same error:
```
    if ((struct kinfo_file *kif = kvm_getfiles(kd, KERN_FILE_BYPID, getpid(), sizeof(struct kinfo_file), &cntp))) {
```
It's really odd because I thought all of these would work.

faho: The first one is fine.

src/common.cpp
```diff
@@ -1873,6 +1880,58 @@ std::string get_executable_path(const char *argv0) {
...
+    }
+    kvm_t *kd = kvm_openfiles(nullptr, nullptr, nullptr, KVM_NO_FILES, nullptr);
+    if (!kd) { result = "fish"; return result; }
+    int cntp = 0;
+    bool ok = false;
+    kinfo_file *kif = nullptr;
+    if ((kif = kvm_getfiles(kd, KERN_FILE_BYPID, getpid(), sizeof(struct kinfo_file), &cntp))) {
+        for (int i = 0; i < cntp; i++) {
+            if (kif[i].fd_fd == KERN_FILE_TEXT) {
+                if (!stat(path.c_str(), &st)) {
```
faho, line 1917: Oh, I just noticed, you're stat'ing `path` over and over again.

Extract this `if` out of the for-loop and just keep `st` around.

Since `path` doesn't change here, if the `stat` fails you can just skip this entire thing and return `"fish"`.

src/common.cpp
```diff
@@ -1873,22 +1880,18 @@ std::string get_executable_path(const char *argv0) {
...
-        }
+        len = readlink("/proc/self/path/a.out", buff, sizeof buff - 1);  // Solaris
     }
     if (len > 0) {
         buff[len] = '\0';
         // When /proc/self/exe points to a file that was deleted (or overwritten on update!)
         // then linux adds a " (deleted)" suffix.
         // If that's not a valid path, let's remove that awkward suffix.
-        std::string buffstr{buff};
+        std::string buffstr = {buff};
```
faho, line 1894: This was fine before. `type name {value}` is another way to initialize a variable in C++. It prevents some baffling constructor choices.

src/common.cpp
```diff
@@ -1873,22 +1880,18 @@ std::string get_executable_path(const char *argv0) {
     } else {
         return std::string(buff);
     }
-#else
-    // On other unixes, fall back to the Linux-ish /proc/ directory
+#elif (defined(__linux__) || defined(sun) || defined(__sun))
```
faho, line 1883: This is unnecessary enumeration, the `#else` can stay.

It's *fine* to try /proc, and if that doesn't work try argv0.

src/common.cpp
```diff
@@ -1898,6 +1901,63 @@ std::string get_executable_path(const char *argv0) {
...
+                    }
+                }
+            }
+        } else {
+            const char *pwd = getenv("PWD");
+            if (pwd && *pwd) {
+                buffstr = std::string(pwd) + "/" + strargv0;
+            }
+        }
+        return buffstr;
```
faho, line 1929: This should set `result`.

src/common.cpp
```diff
@@ -1898,6 +1901,63 @@ std::string get_executable_path(const char *argv0) {
         }
         return buffstr;
     }
+#else
+    static std::string result;
+    if (!result.empty()) return result;
+    std::string strargv0 = ((argv0) ? argv0 : ""), buffstr;
```
faho, line 1907: Please don't define two variables on the same line.

I had trouble figuring out where `buffstr` was defined.

src/common.cpp
```diff
@@ -1898,6 +1901,63 @@ std::string get_executable_path(const char *argv0) {
...
+    }
+#ifdef __OpenBSD__
+    // kd variable is necessary for kvm_getfiles() first argument; RTFM.
+    kvm_t *kd = kvm_openfiles(nullptr, nullptr, nullptr, KVM_NO_FILES, nullptr);
+    if (!kd) { result = "fish"; return result; }
+    int cntp = 0;
+    bool ok = false;
+    kinfo_file *kif = nullptr;
+    if ((kif = kvm_getfiles(kd, KERN_FILE_BYPID, getpid(), sizeof(struct kinfo_file), &cntp))) {
+        if (!stat(buffstr.c_str(), &st)) {
```
faho, line 1939: Can this actually be *reached*?

I don't see any way `buffstr` gets any value without being returned?

faho (changes requested):

faho:
@time-killer-games Can I ask you to work on this a bit in your own time, without immediately pushing everything here?

I feel like there's a lot of work remaining to be done, it's a bit messy.

I would probably start anew, and try to do just the things we know need to be done, one at a time, and make some sensible commits along the way. So you'd have "Add narrow split_string overload" and "Use -1 instead of getpid() for NetBSD" and such.

Leave the "to_string" to "to_wcstring" rename for the very end, as a separate thing. Adding more overloads and the awkward `#ifndef __OpenBSD__` can be done before, but under the old name.

And then, for the main course, think hard about which platform should be getting which solution, and try to cut that into sensible pieces. If the control flow becomes too unwieldy, maybe extract something into a function. Only use `#ifdef __PLATFORM__` if absolutely necessary (i.e. it doesn't compile). Try to keep the flows similar, to make it easier to follow.

src/common.cpp
time-killer-games: that was actually an accident. lol

time-killer-games:
@faho alright I think I addressed the last few things you wanted. Now I'm going to start a new pr's to replace this one like you said. I'm going to wait for CI to pass just out of curiosity.

src/common.cpp
```diff
@@ -1874,13 +1880,14 @@ size_t buff_size = sizeof buff;
...
 #endif
-    int result = sysctl(name, sizeof(name) / sizeof(int), buff, &buff_size, nullptr, 0);
-    if (result != 0) {
+    int ret = sysctl(name, sizeof(name) / sizeof(int), buff, &buff_size, nullptr, 0);
+    if (ret != 0) {
         wperror(L"sysctl KERN_PROC_PATHNAME");
     } else {
-        return std::string(buff);
+        buffstr = buff;
+        return result;
```
faho, line 1888: This sets buffstr and returns result. That means you'll get a wrong answer and it'll be computed every time.

Just set result directly, no need for buffstr here.

faho:
>Now I'm going to start new pr's to replace this one like you said


Okay. Note that you don't have to file one PR per commit or anything, just try to find the separate concerns.

It's *fine* to add the split_string overload in the PR that starts using it. Just make it a separate commit.

time-killer-games:
I just want to make sure everything is perfect before I make my fresh pr's. This should do it.

src/common.cpp
```diff
@@ -1849,61 +1856,122 @@ bool valid_func_name(const wcstring &str) {
...
 #ifdef __APPLE__
     // On OS X use it's proprietary API to get the path to the executable.
     // This is basically grabbing exec_path after argc, argv, envp, ...: for us
     // https://opensource.apple.com/source/adv_cmds/adv_cmds-163/ps/print.c
-    uint32_t buffSize = sizeof buff;
-    if (_NSGetExecutablePath(buff, &buffSize) == 0) return std::string(buff);
-#elif defined(__BSD__) && defined(KERN_PROC_PATHNAME)
+    uint32_t buff_size = sizeof buff;
+    if (_NSGetExecutablePath(buff, &buff_size) == 0) {
+        buffstr.resize(buff_size, '\0');
```
faho, line 1869: You don't need to resize the string before you set it.

Issue #9086, opened 2022-07-22
Get OpenBSD executable path


ghost:
The code in master doesn't allow for this, in OpenBSD to get the executable path, and mounting the procfs is no longer possible.

Feel free to add me under people assigned.

I just want to know in advance if this code will be accepted or if it's too hacky to be taken seriously:
https://github.com/time-killer-games/libfilesystem/blob/2f773e866f99f7ea2228f970dbf43ed3128df821/filesystem.cpp#L541-L587

Of course, I will rewrite this code to be C compliant once the pr will be opened. OpenBSD has no official API for getting the current exe path, and I don't know if it ever did, it very well may have when the procfs was available, but i wasn't an OpenBSD user when that was still available to the OS. This is pretty much the best we can do without an official API.

time-killer-games:
Here's what the solution would look like:

https://github.com/time-killer-games/getexecname/blob/patch-1/getexecname.c

Note this is a custom fork of someone else's code which is a C solution I modified to support getting the $PWD for specific use cases.

faho:
You'll want to look through our codebase, because we do of course have helpers to [find an executable in $PATH](https://github.com/fish-shell/fish-shell/blob/0f13337ae62efc609182087193843dada38e1ae9/src/path.cpp#L87-L94), or to [add $PWD to a path](https://github.com/fish-shell/fish-shell/blob/0f13337ae62efc609182087193843dada38e1ae9/src/path.cpp#L254).

That should make this quite a bit simpler.

faho:
Tbh I hate the complexities of that code, the hoops we'd have to jump through.

Because it relies on $PATH and the working directory, you need to jump through hoops to get the *original* values of those. But even that isn't actually guaranteed to be correct!

Because what you need is the $PATH and cwd of *fish's parent process*, at the time it started fish.

So I'm tempted to just declare that we'll be printing "fish" on OpenBSD until they grow a feature that allows us to actually be correct.

time-killer-games:
if the executable path fails to be retrieved for one reason or another, couldn't we check for nullptr or an empty string and in such a case use hard-coded "fish" as a fallback?

time-killer-games:
> Tbh I hate the complexities of that code, the hoops we'd have to jump through.
> 
> Because it relies on $PATH and the working directory, you need to jump through hoops to get the _original_ values of those. But even that isn't actually guaranteed to be correct!
> 
[...]

PR #9093, opened 2022-07-28
Improve get executable path() Function

time-killer-games:
## Description

- macOS API _NSGetExecutablePath() can sometimes return a symbollic link; added realpath to resolve potential symlinks.
- NetBSD API sysctl() / KERN_PROC_PATHNAME can sometimes return an un-normalized path; added realpath to resolve that.
- Unlike FreeBSD, it is not documented for NetBSD, but you can use a -1 instead of getpid() for the current executable path.
- Added new fallback solution which should work on all POSIX-compliant Unix likes. OpenBSD verifies the correct dev / inode.
- Added std::string split_string() overloads.

Fixes issue #9086

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -167,10 +167,12 @@ endfunction(FISH_LINK_DEPS_AND_SIGN)
 
 # Define libfish.a.
 add_library(fishlib STATIC ${FISH_SRCS} ${FISH_BUILTIN_SRCS})
+execute_process(COMMAND sh "-c" "find /usr/lib/libkvm.a 2> /dev/null" OUTPUT_VARIABLE KVM_LIBRARY)
+string(REPLACE "\n" "" KVM_LIB "${KVM_LIBRARY}")
 target_sources(fishlib PRIVATE ${FISH_HEADERS})
 target_link_libraries(fishlib
   ${CURSES_LIBRARY} ${CURSES_EXTRA_LIBRARY} Threads::Threads ${CMAKE_DL_LIBS}
-  ${PCRE2_LIB} ${Intl_LIBRARIES} ${ATOMIC_LIBRARY})
+  ${PCRE2_LIB} ${Intl_LIBRARIES} ${ATOMIC_LIBRARY} ${KVM_LIB})
 target_include_directories(fishlib PRIVATE
   ${CURSES_INCLUDE_DIRS})
 
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -71,7 +71,6 @@
 
 struct termios shell_modes;
 
-extern 
 const wcstring g_empty_string{};
 
 /// This allows us to notice when we've forked.
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -56,8 +56,14 @@
 #include "wutil.h"  // IWYU pragma: keep
 
 // Keep after "common.h"
+#if defined(__OpenBSD__)
+#include <kvm.h>
+#include <sys/types.h>
+#include <sys/sysctl.h>
+#else
 #ifdef HAVE_SYS_SYSCTL_H
 #include <sys/sysctl.h>
+#endif
 #endif
 #if defined(__APPLE__)
 #include <mach-o/dyld.h>
@@ -65,6 +71,7 @@
 
 struct termios shell_modes;
 
+extern 
 const wcstring g_empty_string{};
 
 /// This allows us to notice when we've forked.
@@ -1849,46 +1856,60 @@ bool valid_func_name(const wcstring &str) {
 
 /// Return the path to the current executable. This needs to be realpath'd.
 std::string get_executable_path(const char *argv0) {
+    static std::string result;
+    if (!result.empty()) return result;
     char buff[PATH_MAX];
-
+    std::string buffstr;
 #ifdef __APPLE__
     // On OS X use it's proprietary API to get the path to the executable.
     // This is basically grabbing exec_path after argc, argv, envp, ...: for us
     // https://opensource.apple.com/source/adv_cmds/adv_cmds-163/ps/print.c
-    uint32_t buffSize = sizeof buff;
-    if (_NSGetExecutablePath(buff, &buffSize) == 0) return std::string(buff);
-#elif defined(__BSD__) && defined(KERN_PROC_PATHNAME)
+    uint32_t buff_size = sizeof buff;
+    if (_NSGetExecutablePath(buff, &buff_size) == 0) {
+        char path[PATH_MAX];
+        if (realpath(buff, path)) {
+            result = path;
+            return result;
+        }
+    }
+    result = "fish";
+    return result;
+#else
+    size_t buff_size = sizeof buff;
+#endif
+#if defined(__BSD__) && defined(KERN_PROC_PATHNAME)
     // BSDs do not have /proc by default, (although it can be mounted as procfs via the Linux
     // compatibility layer). We can use sysctl instead: per sysctl(3), passing in a process ID of -1
     // returns the value for the current process.
-    size_t buff_size = sizeof buff;
 #if defined(__NetBSD__)
-    int name[] = {CTL_KERN, KERN_PROC_ARGS, getpid(), KERN_PROC_PATHNAME};
+    int name[] = {CTL_KERN, KERN_PROC_ARGS, -1, KERN_PROC_PATHNAME};
 #else
     int name[] = {CTL_KERN, KERN_PROC, KERN_PROC_PATHNAME, -1};
 #endif
-    int result = sysctl(name, sizeof(name) / sizeof(int), buff, &buff_size, nullptr, 0);
-    if (result != 0) {
+    int ret = sysctl(name, sizeof(name) / sizeof(int), buff, &buff_size, nullptr, 0);
+    if (ret != 0) {
         wperror(L"sysctl KERN_PROC_PATHNAME");
     } else {
-        return std::string(buff);
+        char path[PATH_MAX];
+        if (realpath(buff, path)) {
+            result = path;
+            return result;
+        }
     }
-#else
-    // On other unixes, fall back to the Linux-ish /proc/ directory
+    result = "fish";
+    return result;
+#endif
     ssize_t len;
     len = readlink("/proc/self/exe", buff, sizeof buff - 1);  // Linux
     if (len == -1) {
-        len = readlink("/proc/curproc/file", buff, sizeof buff - 1);  // other BSDs
-        if (len == -1) {
-            len = readlink("/proc/self/path/a.out", buff, sizeof buff - 1);  // Solaris
-        }
+        len = readlink("/proc/self/path/a.out", buff, sizeof buff - 1);  // Solaris
     }
     if (len > 0) {
         buff[len] = '\0';
         // When /proc/self/exe points to a file that was deleted (or overwritten on update!)
         // then linux adds a " (deleted)" suffix.
         // If that's not a valid path, let's remove that awkward suffix.
-        std::string buffstr{buff};
+        buffstr = buff;
         if (access(buff, F_OK)) {
             auto dellen = const_strlen(" (deleted)");
             if (buffstr.size() > dellen &&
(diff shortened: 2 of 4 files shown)


time-killer-games:
So far I've tested this on OpenBSD, NetBSD, and macOS. I'm going to test Linux, FreeBSD, and Cygwin next.

faho:
Okay, those commits could be better.

A commit should be a self-contained "change". It should do one thing, but no less. The message should explain what that thing is. The tricky thing is, of course, getting the right feel for what "one thing" is.

In your case, you've thought too much about the file boundaries. A commit doesn't have to be contained to one file, if the change logically spans multiple.

First of all: One commit can add a thing to both the source and the header. Adding it just to the header isn't really self-contained, because without being added to the header the function is useless.

So we would have one commit: "Add narrow split_string overload", that adds the "split_string" version that operates on std::string/char, to both the source and the header.

Secondly: "Add libkvm to CMakeLists.txt for OpenBSD." is, by itself, meaningless. It links against a library that isn't used anywhere. It's okay to add a library in the same commit that uses it.

Thirdly: "[skip ci] Improve get_executable_path() Function." says nothing. This message effectively means "make better".

The commit message then explains the four different things it does:

1. macOS API _NSGetExecutablePath() can sometimes return a symbollic link; added realpath to resolve potential symlinks.
2. NetBSD API sysctl() / KERN_PROC_PATHNAME can sometimes return an un-normalized path; added realpath to resolve that.
3. Unlike FreeBSD, it is not documented for NetBSD, but you can use a -1 instead of getpid() for the current executable path.
4. Added new fallback solution which should work on all POSIX-compliant Unix likes. OpenBSD verifies the correct dev / inode.

Now, that's probably a bit *too* granular, you could make a case that it's three changes - "Normalize paths in get_executable_path", "Use -1 instead of getpid() on NetBSD for get_executable_path" and "Add a new fallback way to get_executable_path".

Finally: "Add #include <vector> to header." - what is this used for? Why is that there? If it is newly used in one of the other commits, put it in there. If it was used before and included transitively, that's fine but the message should explain that - "Add missing vector include".

faho:
>It works with Linux and FreeBSD. MSYS2 / MinGW failed. Note sure if that is currently supported?

I don't see that you changed anything that requires locale_t, so it was probably already broken.

CMakeLists.txt
```diff
@@ -167,10 +167,12 @@ endfunction(FISH_LINK_DEPS_AND_SIGN)
 
 # Define libfish.a.
 add_library(fishlib STATIC ${FISH_SRCS} ${FISH_BUILTIN_SRCS})
+execute_process(COMMAND sh "-c" "find /usr/lib/libkvm.a 2> /dev/null" OUTPUT_VARIABLE KVM_LIBRARY)
+string(REPLACE "\n" "" KVM_LIB "${KVM_LIBRARY}")
```
faho, line 171: I am not enthused by this solution, this isn't typically how you use cmake.

I'd have to look up what the correct way is, but this appears very hacky.

src/common.cpp
```diff
@@ -56,15 +56,22 @@
...
 #include <sys/sysctl.h>
 #endif
+#endif
 #if defined(__APPLE__)
 #include <mach-o/dyld.h>
 #endif
 
 struct termios shell_modes;
 
+extern
```
faho, line 74: This is an unrelated change, please remove.

src/common.cpp
```diff
@@ -1849,61 +1856,122 @@ bool valid_func_name(const wcstring &str) {
 
 /// Return the path to the current executable. This needs to be realpath'd.
 std::string get_executable_path(const char *argv0) {
+    static std::string result;
+    if (!result.empty()) return result;
     char buff[PATH_MAX];
-
+    std::string buffstr;
```
faho, line 1862: Please define this where it is used. This probably means defining it *before* that one if-block where it first appears (instead of inside it).

src/common.cpp
```diff
@@ -1849,61 +1856,122 @@ bool valid_func_name(const wcstring &str) {
...
 #if defined(__NetBSD__)
-    int name[] = {CTL_KERN, KERN_PROC_ARGS, getpid(), KERN_PROC_PATHNAME};
+    int name[] = {CTL_KERN, KERN_PROC_ARGS, -1, KERN_PROC_PATHNAME};
 #else
     int name[] = {CTL_KERN, KERN_PROC, KERN_PROC_PATHNAME, -1};
 #endif
-    int result = sysctl(name, sizeof(name) / sizeof(int), buff, &buff_size, nullptr, 0);
-    if (result != 0) {
+    int ret = sysctl(name, sizeof(name) / sizeof(int), buff, &buff_size, nullptr, 0);
+    if (ret != 0) {
```
faho, line 1890: Instead of renaming this unrelated variable, you should name your result string something else. "exe_path"?

src/common.cpp
```diff
@@ -1849,61 +1856,122 @@ bool valid_func_name(const wcstring &str) {
...
     }
-#else
-    // On other unixes, fall back to the Linux-ish /proc/ directory
+    result = "fish";
+    return result;
+#endif
     ssize_t len;
     len = readlink("/proc/self/exe", buff, sizeof buff - 1);  // Linux
     if (len == -1) {
-        len = readlink("/proc/curproc/file", buff, sizeof buff - 1);  // other BSDs
```
faho, line 1881: I am wary of just removing this path. There haven't been any issues with it that I know of and there are more BSDs.

src/common.cpp
```diff
@@ -1849,61 +1856,122 @@ bool valid_func_name(const wcstring &str) {
...
+        char path[PATH_MAX];
+        if (realpath(buff, path)) {
+            result = path;
+            return result;
+        }
     }
-#else
-    // On other unixes, fall back to the Linux-ish /proc/ directory
+    result = "fish";
+    return result;
```
faho, line 1900: Here we return "fish" because the sysctl failed. I think we should try further, and go via $PATH if necessary.

src/common.cpp
```diff
@@ -1849,61 +1856,122 @@ bool valid_func_name(const wcstring &str) {
...
+            }
+        }
+    }
+    buffstr.resize(buff_size, '\0');
+    buffstr[buff_size] = '\0';
+    if (realpath(buffstr.c_str(), buff)) {
+        buffstr = buff;
     }
+#ifdef __OpenBSD__
+    // kd variable is necessary for kvm_getfiles() first argument; RTFM.
```
faho, line 1952: It's obvious that the "kd" is passed in the next function, and keeping some "handle" is a fairly common idiom, no need to insult the reader.

time-killer-games:
Why are you saying the libkvm library isn't used anywhere? OpenBSD is using it, and I thought you knew that. It's treated like an empty string on non-OpenBSD platfroms so it won't effect those.

src/common.cpp
```diff
@@ -1849,61 +1856,122 @@ bool valid_func_name(const wcstring &str) {
...
+    kvm_t *kd = kvm_openfiles(nullptr, nullptr, nullptr, KVM_NO_FILES, nullptr);
+    if (!kd) { result = "fish"; return result; }
+    int cntp = 0;
+    bool ok = false;
+    kinfo_file *kif = nullptr;
+    if ((kif = kvm_getfiles(kd, KERN_FILE_BYPID, getpid(), sizeof(struct kinfo_file), &cntp))) {
```
faho, line 1958: My main issue with this path is that I have no idea in what cases it is useful.

Is it that e.g. a shell calls us and passes a modified $PATH, e.g.

```bash
PATH=/tmp fish
# The "fish" is in /usr/bin, but there is another in /tmp
```

In that case we would find /tmp/fish, and this would then figure out that it's not the correct one and print "fish" instead?

That's not all that helpful, because if you return "fish" people will just go via $PATH *anyway*, so you still get the wrong fish. I don't think that justifies this level of trickery and linking another library.

src/wcstringutil.cpp
```diff
@@ -256,6 +256,20 @@ wcstring_list_t split_string(const wcstring &val, wchar_t sep) {
     return out;
 }
 
+std::vector<std::string> split_string(const std::string &val, char sep) {
+    std::vector<std::string> out;
+    size_t pos = 0, end = val.size();
+    while (pos <= end) {
+        size_t next_pos = val.find(sep, pos);
+        if (next_pos == wcstring::npos) {
```
faho, line 264: This needs to be `std::string::npos`. I think they are typically the same value, but I'm fairly certain they don't have to be.

faho (changes requested):

faho:
>Why are you saying the libkvm library isn't used anywhere? OpenBSD is using it, and I thought you knew that.

I am saying it is used nowhere *in that commit*, so it hangs a bit in the air.

It's fine for the split_string change to be its own thing, because you could build other things from it, but you're linking libkvm specifically for this OpenBSD change, and it's unlikely we would be using it anywhere else.

So it should be part of the commit that adds the kvm_X bit to get_executable_path.

faho:
Worse: In this case you've *first* committed the change that uses it and then you added the library.

That means anything between https://github.com/fish-shell/fish-shell/pull/9093/commits/583461b740f5fa68f2650fd352a893ff1e8cb754 and https://github.com/fish-shell/fish-shell/pull/9093/commits/48d55f2a53143a061ea6796b0cb669e7188b613e won't be buildable on OpenBSD.

That causes issues with bisecting. Each commit should be usable, if at all possible. That means if you do a separate commit that adds a dependency and then another where you use it, the commit that adds it (links the library, adds a method, whatever), needs to be *first*.

src/common.cpp
time-killer-games: I wasn't sure if RTFM meant "Read the f'ing manual" or "Read the fine manual" -- I've heard it used both ways, I was hoping it was the latter.

src/common.cpp
time-killer-games: I put it at the top in front of all the platform checks because it is used by all platforms iirc.

faho: It's not used by the Apple path *or* the KERN_PROC_PATHNAME one.

It appears the first use is in the /proc bit, followed by the $PATH/$PWD bit.

src/common.cpp
```diff
@@ -1849,61 +1855,122 @@ bool valid_func_name(const wcstring &str) {
...
+                std::vector<std::string> env = split_string(penv, ':');
+                for (std::size_t i = 0; i < env.size(); i++) {
+                    buffstr = env[i] + "/" + strargv0;
+                    if (!stat(buffstr.c_str(), &st) && (st.st_mode & S_IXUSR) && (st.st_mode & S_IFREG)) {
+                        break;
+                    }
+                }
+            }
+        } else {
+            const char *pwd = getenv("PWD");
```
faho, line 1939: This should probably use getcwd() instead. $PWD doesn't need to exist as a variable, and it doesn't need to be the correct path.

src/common.cpp
faho: Both have been used and so people reading this might understand it either way, which means it's potentially insulting.

Anyway, the comment isn't very helpful - you can see that "kd" is used again a few lines later - so I'd just remove it.

But see the next comment, I'm not actually sold on the kvm_X stuff at all.

src/common.cpp
time-killer-games: I remembered wrong. I'm working on multiple things right now otherwise I probably would've noticed that.

src/common.cpp
time-killer-games: I'm happy to remove it. That makes things easier.

time-killer-games:
@faho I don't want this to get annoying to you. I'm trying really hard to get everything done correctly, but this is starting to drag out quite a bit. Once I address all of your concerns, would you like me to make a new pull request again, now that I think we're clear on how it must be done finally? I have a lot to learn and I feel bad for how much code review has been needed because I'm not doing it right the first time.

faho:
>Once I address all of your concerns, would you like me to make a new pull request again

Either that, or you can force-push to the pull request branch and github will update it automatically.

But if you're not comfortable with the git trickery required, it's fine to make a new PR.

time-killer-games:
Thank you!

time-killer-games:
> > Once I address all of your concerns, would you like me to make a new pull request again
> 
> Either that, or you can force-push to the pull request branch and github will update it automatically.
> 
> But if you're not comfortable with the git trickery required, it's fine to make a new PR.

I think the amount of work it takes to review my code is a lot more involved and a lot more stress and work than if you guys were to do it yourself. Since you know exactly how you want it, and you are all far more experienced at this stuff than I am, I feel I am being nothing more than a problem rather than a help.

I think I have a lot to learn before I am ready to make contributions, each time I try to contribute to a project I am reminded of this. I'm going to close this. That said, thank you for helping me be more mindful about how I write my code, I've learned a lot from this experience. I am also a bit overwhelmed at the moment.

floam:
Thanks a lot @time-killer-games for the effort, and for letting us know a whole lot about how to do this right on OpenBSD, and your improvements that made it in for NetBSD.

Issue #8766, opened 2022-03-07
--on-process-exit not getting called within a --on-event function


Baspar:
When creating a function with `--on-process-exit` or `--on-job-exit` from within a `--on-event` function, the "callback" function never gets called.

Example:
```fish
function fish_init --on-event fish_prompt
    sleep 3 &
    function fish_callback --on-process-exit (jobs --last --pid)
      commandline -a "Callback DONE"
    end
end
```

Executing manually the inner code seems to work fine

### Info:
**Fish Version**: 3.3.1
**Terminal**: Kitty 
**System**: Tested on MacOS & Linux

ridiculousfish:
Yeah, jobs triggered from within event handlers do not themselves post process-exit events. I made this change in f7e2e7d26be2253f14551a023a339fca136593d2 and while I don't remember the details I think it was to avoid an infinite recursion for on-process-exit events. It should be relaxed for background jobs at least and maybe more; this is definitely something that can be imnproved.

Baspar:
I opened a partial PR to cover the background jobs in #9096, but it might need some additional polish

PR #9096, opened 2022-07-29
Generate job & process exit events for background jobs

Baspar:
## Description

Allow for `--on-process-exit` and `--on-job-exit` to be called inside of an event handler, as long as it is from a background job

(partially) Fixes issue #8766

## Reproducing the issue
```fish
function __test_background_job_exit_event --on-variable dummy_var
  sleep 1&
  function callback --on-process-exit (jobs --last --pid)
    echo -n "Callback called"
  end
end

set dummy_var ""
# after 1s, "Callback called" will be printed on the screen
```


ridiculousfish:
Merged as 2410e27d10ac2c5c8a24393438a070f50b430a4d - thank you!

PR #9105, opened 2022-08-02
Disclose pager to half of screen height immediately

### LINKED ISSUE #2698: Set the number of rows/lines for completions
I have a large screen and would like to use more than 4 rows for autosuggestions. 

screen shot 2016-01-26 at 20 38 53

I know tat I can hit tab again to get the complete list.

faho:
This removes that bit where we only show 4 rows at most at first,
instead we disclose between half and up to the full terminal height.

This results in less pressing of tab to get the other results, and
better visibility of results.

However, it also means that it'll push the shell up to the top of the
terminal a lot.

So this is an experiment to see how that works in practice.

Fixes #2698

----

The direct impetus for this is #9089, which creates up to 12 rows. In that case it feels supremely weird to have to disclose "8 more rows", when there's clearly more room.

An alternative is to increase the maximum either up to a constant, or up to half the terminal height.

----

There's some weirdness here where it doesn't *count* the pager as fully disclosed, and so pressing tab again doesn't immediately move down into it, which doesn't actually save any key presses. I'd look into it, but only after we've decided which way to go.

(also the constant here still exists and such)

---

Edit: Originally this went fullscreen immediately. That's awkward because it removes all scrollback.

### DIFF
--- a/src/pager.cpp
+++ b/src/pager.cpp
@@ -424,9 +424,6 @@ bool pager_t::completion_try_print(size_t cols, const wcstring &prefix, const co
     size_t term_height =
         this->available_term_height - 1 -
         (search_field_shown ? 1 : 0);  // we always subtract 1 to make room for a comment row
-    if (!this->fully_disclosed) {
-        term_height = std::min(term_height, static_cast<size_t>(PAGER_UNDISCLOSED_MAX_ROWS));
-    }
 
     size_t row_count = divide_round_up(lst.size(), cols);
 
--- a/src/pager.cpp
+++ b/src/pager.cpp
@@ -424,6 +424,17 @@ bool pager_t::completion_try_print(size_t cols, const wcstring &prefix, const co
     size_t term_height =
         this->available_term_height - 1 -
         (search_field_shown ? 1 : 0);  // we always subtract 1 to make room for a comment row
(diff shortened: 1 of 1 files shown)


krobelus:
Is it possible to show exactly the number of rows where we don't push the shell up? That would require knowing the current row which I guess we can't.

Anyway, I think the suggested behavior is okay to try and see if anyone complains. I don't think I use the pager much in day-to-day fishing so I can't really tell if it is bothersome.

Substantially improving the pager is hard.
I'm starting to think it should behave more like in an IDE - where the completion pager has a fixed max size, and perhaps show completions automatically, without having to press tab (though that has security implications).

faho:
I've been trying out a different idea: We disclose between 50% and 100%, but at least 4 rows.

In practice that means we disclose half the terminal, which feels okay and leaves a bit of scrollback on-screen and doesn't snap fish off into the stratosphere (i.e. the top of the screen).

I'm tempted to just merge that and see how people get on with it. No shame in reverting it, and I don't really see RFC PRs working too well.

faho:
Okay, I'm merging this speculatively. If people don't like it it's extremely easy to back out or refine, but without adding it to master we're not gonna get enough testing.

PR #9124, opened 2022-08-08
string repeat: Don't allocate repeated string all at once

faho:
This used to allocate one string and fill it with the necessary
repetitions, which could be a very very large string.

Now, it instead uses one buffer and fills it to a chunk size,
and then writes that.

This fixes:

1. We no longer crash with too large max/count values. Before they
caused a bad_alloc because we tried to fill all RAM.
2. We no longer fill all RAM if given a big-but-not-too-big value. You
could've caused fish to eat *most* of your RAM here.
3. It can start writing almost immediately, instead of waiting
potentially minutes to start.

Performance is about the same to slightly faster overall.

## TODOs:

### DIFF
--- a/tests/checks/string.fish
+++ b/tests/checks/string.fish
@@ -527,7 +527,7 @@ string repeat -n 17 ab | string length
 string repeat -n 17 a | string length
 # CHECK: 17
 # And a more tricksy case with a long string that we truncate.
-string repeat -m 5 (string repeat -n 500000000 aaaaaaaaaaaaaaaaaa) | string length
+string repeat -m 5 (string repeat -n 500000 aaaaaaaaaaaaaaaaaa) | string length
 # CHECK: 5
 
 # Test equivalent matches with/without the --entire, --regex, and --invert flags.
--- a/tests/checks/string.fish
+++ b/tests/checks/string.fish
@@ -504,6 +504,32 @@ string repeat -n3 ""
 or echo string repeat empty string failed
 # CHECK: string repeat empty string failed
 
+# See that we hit the expected length
+# First with "max", i.e. maximum number of characters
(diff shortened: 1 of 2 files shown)


faho:
Ah, I think the tests fail because the number isn't *high* enough. This is ram-dependent.

tests/checks/string.fish
```diff
@@ -504,6 +504,32 @@ string repeat -n3 ""
...
+# Then with "count", i.e. number of repetitions.
+# (these are count * length long)
+string repeat -n 17 aab | string length
+# CHECK: 51
+string repeat -n 17 ab | string length
+# CHECK: 34
+string repeat -n 17 a | string length
+# CHECK: 17
+# And a more tricksy case with a long string that we truncate.
+string repeat -m 5 (string repeat -n 500000 aaaaaaaaaaaaaaaaaa) | string length
```
krobelus, line 530: Nice work, LGTM!

The longer version also reaches fish_read_limit for me locally even though I didn't set it.

faho:
Just as a note: I had just assumed that this would change the return code in case the pipe is closed early.

It turns out it mostly doesn't, because we set it to error anyway if a builtin exited successfully but had an io error: https://github.com/fish-shell/fish-shell/blob/df7b97d99dc30020ea8535b12c79f63e1edb4628/src/builtin.cpp#L466-L475

This can be tested with something like

```fish
string repeat -m 100000 aaaa\n | head -n 1
```

It's *possible* this changes in edge cases - if the chunking size differs from the pipe buffer size, and we now end up trying to write more or less. Because SIGPIPE errors are unreliable anyway (which is a major issue with bash's "pipefail" option), I don't think that's a problem.

PR #9129, opened 2022-08-09
Reset the read byte limit to the default when unset

faho:
This used to be kept, so e.g. testing it with

    fish_read_limit=5 echo (string repeat -n 10 a)

would cause the prompt and such to error as well.

Also there was no good way to get back to the default value
afterwards.

## TODOs:

### DIFF
--- a/src/env_dispatch.cpp
+++ b/src/env_dispatch.cpp
@@ -56,6 +56,12 @@
 #include "trace.h"
 #include "wutil.h"  // IWYU pragma: keep
 
+
+// Limit `read` to 100 MiB (bytes not wide chars) by default. This can be overridden by the
+// fish_read_limit variable.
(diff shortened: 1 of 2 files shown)


src/env_dispatch.cpp
```diff
@@ -299,6 +305,8 @@ static void handle_read_limit_change(const environment_t &vars) {
         } else {
             read_byte_limit = limit;
         }
+    } else {
+        read_byte_limit = DEFAULT_READ_BYTE_LIMIT;
```
krobelus, line 309: Nice fix, LGTM.
In future we should probably get rid of most global variables but that doesn't seem relevant now.

faho: Oh, definitely, this thing is super janky - and how exactly would that work with multiple parsers?

PR #9130, opened 2022-08-09
Mark error location across the entire length

faho:
## Description

We currently only mark the start of errors with a `^` caret. This changes it so that we mark the start and end with a caret, with a run of `~` tildes in between.

That makes it easier to see the extent of the erroneous syntax, and is familiar from e.g. gcc.

The output looks like

```
checks/set.fish (line 471): for: a,b: invalid variable name. See `help identifiers`
    for a,b in y 1 z 3
        ^~^

checks/cmdsub-limit.fish (line 67): Too much data emitted by command substitution so it was discarded
    echo this will fail (string repeat --max 513 b) to output anything
                        ^~~~~~~~~~~~~~~~~~~~~~~~~~^
```

Note that this exposes the error location in a way that it wasn't before, so it's likely to shake out mistakes and weirdnesses. This is why I've elected to keep the tests broken for now, so we can easily see what's changed and what needs improvement.

## TODOs:

### DIFF
--- a/src/parse_tree.cpp
+++ b/src/parse_tree.cpp
@@ -126,6 +126,19 @@ wcstring parse_error_t::describe_with_prefix(const wcstring &src, const wcstring
     result.push_back(L'\n');
     result.append(caret_space_line);
     result.push_back(L'^');
+    if (source_length > 1) {
+        // Add a squiggle under the error location.
+        // We do it like this
+        //               ^~~^
+        // With a "^" under the start and end, and squiggles in-between.
+        auto width = fish_wcswidth(src.c_str() + source_start, source_length);
+        if (width > 2) {
+            // Subtract one for each of the carets - this is important in case
+            // the starting char has a width of > 1.
+            result.append(width - 2, L'~');
+            result.push_back(L'^');
+        }
+    }
     return result;
 }
 
--- a/src/parse_tree.cpp
+++ b/src/parse_tree.cpp
@@ -132,7 +132,7 @@ wcstring parse_error_t::describe_with_prefix(const wcstring &src, const wcstring
         //               ^~~^
         // With a "^" under the start and end, and squiggles in-between.
         auto width = fish_wcswidth(src.c_str() + source_start, source_length);
-        if (width > 2) {
+        if (width >= 2) {
             // Subtract one for each of the carets - this is important in case
             // the starting char has a width of > 1.
             result.append(width - 2, L'~');
--- a/src/parse_tree.cpp
+++ b/src/parse_tree.cpp
@@ -44,14 +44,17 @@ parse_error_code_t parse_error_from_tokenizer_error(tokenizer_error_t err) {
 wcstring parse_error_t::describe_with_prefix(const wcstring &src, const wcstring &prefix,
                                              bool is_interactive, bool skip_caret) const {
     wcstring result = prefix;
+    // Some errors don't have their message passed in, so we construct them here.
+    // This affects e.g. `eval "a=(foo)"`
     switch (code) {
         default:
             if (skip_caret && this->text.empty()) return L"";
+            result.append(this->text);
             break;
         case parse_error_andor_in_pipeline:
             append_format(result, INVALID_PIPELINE_CMD_ERR_MSG,
                           src.substr(this->source_start, this->source_length).c_str());
-            return result;
+            break;
         case parse_error_bare_variable_assignment: {
             wcstring assignment_src = src.substr(this->source_start, this->source_length);
             maybe_t<size_t> equals_pos = variable_assignment_equals_pos(assignment_src);
@@ -60,10 +63,9 @@ wcstring parse_error_t::describe_with_prefix(const wcstring &src, const wcstring
             wcstring value = assignment_src.substr(*equals_pos + 1);
             append_format(result, ERROR_BAD_COMMAND_ASSIGN_ERR_MSG, variable.c_str(),
                           value.c_str());
-            return result;
+            break;
         }
     }
-    result.append(this->text);
 
     size_t start = source_start;
     size_t len = source_length;
--- a/src/parse_tree.cpp
+++ b/src/parse_tree.cpp
@@ -64,7 +64,20 @@ wcstring parse_error_t::describe_with_prefix(const wcstring &src, const wcstring
         }
     }
     result.append(this->text);
-    if (skip_caret || source_start >= src.size() || source_start + source_length > src.size()) {
+
+    size_t start = source_start;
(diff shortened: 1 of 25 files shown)


faho:
One big issue: If a command substitution errors, this will show both errors and blame the outermost one in the last error:

```fish
set fish_read_limit 512
echo this will fail (string repeat -n 2 (string repeat --max 300 b)) to output anything
echo this will fail (string repeat -n 2 (string repeat --max 550 b)) to output anything
```

results in

```
checks/cmdsub-limit.fish (line 76): Too much data emitted by command substitution so it was discarded
    echo this will fail (string repeat -n 2 (string repeat --max 300 b)) to output anything
                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
```

for the first and

```
    checks/cmdsub-limit.fish (line 1): Too much data emitted by command substitution so it was discarded
    string repeat -n 2 (string repeat --max 550 b)
                       ^~~~~~~~~~~~~~~~~~~~~~~~~~^
    in command substitution
    \tcalled on line 77 of file checks/cmdsub-limit.fish
    checks/cmdsub-limit.fish (line 77): Too much data emitted by command substitution so it was discarded
    echo this will fail (string repeat -n 2 (string repeat --max 550 b)) to output anything
                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
```

for the second. Doing this right requires threading the error through, and then removing the outer errors because the inner substitution failed and it's already clear where the error is.


----

Even *with* that, I'd argue this is still nicer, because the current error is

```
    echo this will fail (string repeat -n 2 (string repeat --max 550 b)) to output anything
                        ^
```

faho:
And this is very misleading:

```
    fish: Unexpected end of string, square brackets do not match
    echo f[oo # not valid, no matching ]
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
```

This isn't because it detects that `]`, it's a coincidence, it happens to be the last character.

We simply keep iterating our token cursor over the comment and fall off the end.

krobelus:
> Doing this right requires threading the error through, and then removing the outer errors because the inner substitution failed and it's already clear where the error is.

Yeah. It's okay to leave this for later.

> This isn't because it detects that `]`, it's a coincidence, it happens to be the last character.

Right, this is a pathological case. Probably we should rewrite this test case to something like

		    fish: Unexpected end of string, square brackets do not match
		    echo f[oo # not valid, no matching ] to close list index 
		          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^

which should still capture the original intent of this test

src/parse_tree.cpp
```diff
@@ -126,6 +126,19 @@ wcstring parse_error_t::describe_with_prefix(const wcstring &src, const wcstring
...
+        // Add a squiggle under the error location.
+        // We do it like this
+        //               ^~~^
+        // With a "^" under the start and end, and squiggles in-between.
+        auto width = fish_wcswidth(src.c_str() + source_start, source_length);
+        if (width > 2) {
+            // Subtract one for each of the carets - this is important in case
+            // the starting char has a width of > 1.
+            result.append(width - 2, L'~');
+            result.push_back(L'^');
```
krobelus, line 139: I would expect to also print the second carent if `width=2`. I didn't manage to find a test case though

faho:
>Right, this is a pathological case. Probably we should rewrite this test case to something like

Honestly I'm skipping the squigglies for unterminated things.

By definition, the closing character could be anywhere, and is missing. So we just mark what started it.

I don't believe marking to the end of the line (or even the entire file?) is helpful.

So you'll get the old

```
fish: Unexpected end of string, square brackets do not match
echo f[oo # not valid, no matching ]
      ^
```

krobelus:
> Honestly I'm skipping the squigglies for unterminated things.

Of course, that's much better. We even allow newlines before terminators

faho:
Okay, I added the unsquiggling but it's awkward, because we now need to define the error-length for tokenizer errors.

faho:
Okay, I believe I've now fixed all the immediate problems. The test suite is passing again.

Note that a bunch of errors aren't altered by this but could be, and some cases aren't *ideal* (but acceptable).

And littlecheck doesn't check whitespace, so you see a lot of

```fish
echo $foo[d]
#CHECKERR: {{.*}}expansion.fish (line {{\d+}}): Invalid index value
#CHECKERR: echo $foo[d]
#CHECKERR: ^
```

even tho the caret is right smack on the "d", which is correct.

Some more errors we could improve:

```fish
if $status; echo foo; end'

echo $foo[wrongindex]
```

mark just one char instead of the full variable.

```fish
fish -c 'echo "$abc["'
```

doesn't print a caret at all!

(yes, this is detect_errors instead of the proper parser again!)

```fish
not time true&
```

marks the entire statement.

```fish
set -l p ")"; echo $$p
```

marks just the `$`, could mark `$$p`?

But those are improvements for later. I'm looking to merge this soon to get feedback and work on it.

Also I'm going to be changing littlecheck so it doesn't escape quotes anymore. Nothing worse than

```fish
""
^^
```

looking wrong because littlecheck displays it as

```
\"\"
^^
```

After that's done, I dunno, maybe introduce coloring in the error messages? So you get the erroneous part in red?

faho:
> fish -c 'echo "$abc["'

Figured this one out, it's not detect_errors, it's that it skipped the caret if it was meant to be past the end (of the entire script). So

```fish
fish -c 'echo "$abc["; true'
```

would print it (on the `;`, but still). As a workaround/robustness measure I've made it so we clamp the caret position at the end of the script. This has a slight inconsistency in that

```fish
echo "$abc["
```

will blame the `"`, while

```fish
echo "$abc["; true
```

will blame the semicolon, but that's not *too* bad. And at least it shows the issue, instead of quietly pushing it under the rug (by not showing the caret line).

src/parse_tree.cpp
faho: Done, this is triggered by e.g.

```fish
command ""

foo |& bar
```

faho:
This exercise is turning up a few broken error messages.

E.g.

```fish
fish --no-config -c 'echo notprinted; echo foo; a=b'
```

doesn't print a caret at all.

(the commit message says it fails to print the error, that was probably an editing artifact, I can no longer reproduce it - at one point I got just "fish:" without it explaining what was wrong)

faho:
Alright, merged for now, let's see if anything else turns up.

This seems to be working much better than I expected.

PR #9131, opened 2022-08-10
Highlight shell commands in history pager

krobelus:
To make this work correctly for multiline commands,  use Unicode symbols
for rendering control characters. For example, the command

	begin
	    echo 123
	end

used to be rendered as

	begin\necho 123\nend

which means we can't use the highlighting of the original command, because
\n has turned into two characters.
This proposed fix just turns \n into a single character, then we can reuse
the same highlighting info for the rendered string.

### DIFF
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -883,6 +883,8 @@ static void escape_string_script(const wchar_t *orig_in, size_t in_len, wcstring
     const bool no_quoted = static_cast<bool>(flags & ESCAPE_NO_QUOTED);
     const bool no_tilde = static_cast<bool>(flags & ESCAPE_NO_TILDE);
     const bool no_qmark = feature_test(features_t::qmark_noglob);
+    const bool symbolic = static_cast<bool>(flags & ESCAPE_SYMBOLIC) && (MB_CUR_MAX > 1);
+    assert((!symbolic || !escape_printables) && "symbolic implies escape-no-printables");
 
     bool need_escape = false;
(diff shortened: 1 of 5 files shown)


faho:
Here's what this looks like on my system:

Konsole:

Screenshot_20220810_190102

Gnome Terminal:

Screenshot_20220810_190150

Konsole with our default theme:

Screenshot_20220810_190334

With a C locale:

Screenshot_20220810_190432

----

I don't believe the newline symbol here is really "readable", as a newline. In this font (I *think* that's Noto) it's barely visible.

However, I don't think it really *matters*? I don't think it's really super important, in the context of the pager, to see that it's a newline.

But the C version is just broken, I'd either turn off highlighting in that case, or replace `\n` with `;`.

I keep banging on that drum that a C locale is broken anyway, so it's fine if we have reduced functionality.

krobelus:
Right, I was testing with `LC_ALL=C` but my system still had a UTF-8 locale which we fall back to.

> But the C version is just broken, I'd either turn off highlighting in that case, or replace \n with ;.

sounds good

faho:
>Right, I was testing with LC_ALL=C but my system still had a UTF-8 locale which we fall back to.

You need `fish_allow_singlebyte_locale=1 LC_ALL=C fish`.

krobelus:
I hadn't spent much time on this which is partly why this solution is so hacky.
Anyway, it's good enough and I've documented a way to clean it up in future.

PR #9138, opened 2022-08-12
Improved completions for `ip`

exploide:
## Description

- When adding a VLAN-enabled interface, it is named like `enp0s31f6.100@enp0s31f6` with the physical interface being appended behind an `@`. But subsequent `ip` commands operate on the interface name without this suffix, so it needs to be removed when completing interface names in `__fish_ip_device`.
- Added completions for `ip link delete`
- Updated the global options

## TODOs:

### DIFF
--- a/share/completions/ip.fish
+++ b/share/completions/ip.fish
@@ -362,6 +362,21 @@ function __fish_complete_ip
                                     gso_max_segs
                         end
                     case delete
+                        switch $cmd[-2]
+                            case delete
+                                __fish_ip_device
+                                echo dev
+                                echo group
+                            case dev
+                                __fish_ip_device
+                            case group
+                            case type
+                                __fish_ip_types
+                            case '*'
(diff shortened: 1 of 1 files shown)


exploide:
Added another commit that includes rudimentary completions for `ip route`.

But I got a question. Is it somehow possible to complete multiple words at once? There is the following use case:

For `ip r delete` the arguments look like `default via 192.168.1.1 dev wlp0s20f3`, which could be obtained from `ip r show`. But this would require to be able to complete five space-separated tokens at once. I.e. this is not a single argument that should be quoted but overall looks like `ip r delete default via 192.168.1.1 dev wlp0s20f3`.

faho:
Merged, thanks!

>But I got a question. Is it somehow possible to complete multiple words at once?

There is not, and it's not clear how that would be expressed, and where that would be useful (other than ip, where I'm still holding out hope that someone writes a replacement with a better UI).

PR #9150, opened 2022-08-19
Add missing completions for builtins

kidonng:
## Description

Regarding `status` flags, despite the manpage stating they are deprecated, I think it's still useful to provide completions for them until they are truly removed.

## TODOs:


faho:
Merged without the `status` bit as 1727ed7b5fb1398c0214e171a9bc6357eebc2df8..ada9251412156df58e665e642569d26400fb5cd7

The issue is that this doesn't show up in an editor while you're viewing someone else's code, it shows up while you're *writing* something. So it makes it more likely that people accidentally use the deprecated versions, especially since they aren't called out (but even if they were).

PR #9156, opened 2022-08-25
Add "string shorten" command to shorten strings with an ellipsis

faho:
## Description

This is essentially the inverse of `string pad`.
Where that adds characters to get up to the specified width,
this adds an ellipsis to a string if it goes over a specific maximum width.
The char can be given, but defaults to our ellipsis string.
("…" if the locale can handle it and "..." otherwise)

If the ellipsis string is empty, it just truncates.

For arguments given via argv, it goes line-by-line,
because otherwise visible length makes no sense.

If "--no-newline" is given, it adds an ellipsis instead and removes all subsequent lines.

Like pad and `length --visible`, it goes by visible width,
skipping recognized escape sequences, as those have no influence on width.

The default target width is the shortest of the given widths that is non-zero.

If the ellipsis is already wider than the target width,
we truncate instead. This is safer overall, so we don't e.g. move into a new line.
This is especially important given our default ellipsis might be width 3.

We use this in numerous places including fish_job_summary and the git prompt and completions, but we do that in a fundamentally broken way, and it's already a bit complicated:

- We only check $LANG, where we would have to check $LC_ALL and $LC_CTYPE first
- We have no idea if that locale actually *exists* or is used (we set $LC_CTYPE internally!)
- We need to redo it over and over again
- We go by number of codepoints, not width

This already adopts it in those places. It mostly behaves the same, except for one difference in the git prompt when shortening the branch. The original code took $len chars and then added the ellipsis, so a length of 8 could end up printing a length of 10 (and a width of potentially 8*2 + 2 = 18). This takes as many chars as we can so that we can add the ellipsis and keep under the target width. So people will potentially see their git branch names shortened more - typically 1 char so the unicode ellipsis fits.

It also uses "..." as the ellipsis string on singlebyte systems instead of "..". I'd argue that we should probably change our default fallback ellipsis string to "..", but that's a separate change.

## TODOs:

Potential questions:

- [x] Is "ellipsize" the right name? I've also thought of "shorten" and "truncate"
- [ ] Are the options right or do we need to add new options, because e.g. `--char` actually takes a string (unlike `string pad`, which really just allows a single codepoint)
- [ ] Do we need a "--exclusive" mode to add the ellipsis on top of the target width?
- [ ] To support shortening to the lowest width, this needs to read all input in before acting on it. That's potentially costly with big inputs in terms of memory, but it has a nice symmetry with `string pad`.

### DIFF
--- a/doc_src/cmds/string-shorten.rst
+++ b/doc_src/cmds/string-shorten.rst
@@ -1,0 +1,89 @@
+string-shorten - shorten strings to a width, with an ellipsis
+===============================================================
+
+Synopsis
+--------
+
+.. BEGIN SYNOPSIS
+
+.. synopsis::
+
+    string shorten [(-c | --char) CHARS] [(-m | --max) INTEGER] [(-N | --no-newline)] [(-l | --left)]
+               [STRING ...]
+
+.. END SYNOPSIS
+
+Description
+-----------
+
+.. BEGIN DESCRIPTION
+
+``string shorten`` truncates each *STRING* to the given visible width and adds an ellipsis to indicate it. "Visible width" means the width of all visible characters added together, excluding escape sequences and accounting for :envvar:`fish_emoji_width` and :envvar:`fish_ambiguous_width`. It is the amount of columns in a terminal the *STRING* occupies.
+
+The escape sequences reflect what fish knows about, and how it computes its output. Your terminal might support more escapes, or not support escape sequences that fish knows about.
+
+If **-m** or **--max** is given, truncate at the given width. Otherwise, the lowest non-zero width of all input strings is used.
+
+If **-N** or **--no-newline** is given, only the first line (or last line with **--left**) of each STRING is used, and an ellipsis is added if it was multiline. This only works for STRINGs being given as arguments, multiple lines given on stdin will be interpreted as separate STRINGs instead.
+
+If **-c** or **--char** is given, add *CHAR* instead of an ellipsis. This can also be empty or more than one character.
+
+If **-l** or **--left** is given, remove text from the left on instead, so this prints the longest *suffix* of the string that fits. With **--no-newline**, this will take from the last line instead of the first.
+
+The default ellipsis is ``…``. If fish thinks your system is incapable because of your locale, it will use ``...`` instead.
+
+.. END DESCRIPTION
+
+Examples
+--------
+
+.. BEGIN EXAMPLES
+
+::
+
+    >_ string shorten foo foobar
+    # No width was given, we infer, and "foo" is the shortest.
+    foo
+    fo…
+
+    >_ string shorten --char="..." foo foobar
+    # The target width is 3 because of "foo",
+    # and our ellipsis is 3 too, so we can't really show anything.
+    # This is the default ellipsis if your locale doesn't allow "…".
+    foo
+    ...
+
+    >_ string shorten --char="" --max 4 abcdef 123456
+    # Leaving the char empty makes us not add an ellipsis
+    # So this truncates at 4 columns:
+    abcd
+    1234
+
+    >_ touch "a multiline"\n"file"
+    >_ for file in *; string shorten -N -- $file; end
+    # Shorten the multiline file so we only show one line per file:
+    a multiline…
+
+    >_ ss -p | string shorten -m$COLUMNS -c ""
+    # `ss` from Linux' iproute2 shows socket information, but prints extremely long lines.
+    # This shortens input so it fits on the screen without overflowing lines.
+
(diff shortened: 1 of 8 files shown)


faho:
Oh, also it removes the ellipsis from a shortened SHA. I think it's obvious that a sha of 8 characters isn't complete, and so saving one column is actually the better idea.

(plus that code never checked that the codepoint was usable)

IlanCosman:
Wow, Fabian adding the things I didn't know I wanted, but now do 😄 

One thing I'd like to see is an option that truncates the beginning instead of the end. I.e `foobar` --> `…bar` instead of `foo…`. In many cases, e.g. branch names, hostnames, the beginning information is often less useful and specific than the end.

faho:
>One thing I'd like to see is an option that truncates the beginning instead of the end. I.e foobar --> …bar instead of foo…

Done. This is kind of tricky because we need to skip escapes and need to be careful to not hack combiners into pieces, so actually going from the end is hard. So I just went front to back and stopped once the width to the end fits.

This does a bunch more checks than is needed, but tbh that can be done later.

>Also, I'd go with a more generic name like truncate or shorten, because those are common CS/english words, while ellipsize is

Oh, it is actually used!

You have:

1. [People asking for it in java](https://stackoverflow.com/questions/3288716/automatically-ellipsize-a-string-in-java)
2. [Android API](https://stackoverflow.com/questions/13313996/what-does-ellipsize-mean-in-android)
3. [Someone explaining it in C++](https://www.quora.com/What-is-an-example-of-a-string-ellipsize-algorithm?share=1)
4. [GTK](https://docs.gtk.org/gtk3/property.Label.ellipsize.html)
5. [An NPM thing](https://github.com/mvhenten/ellipsize)
6. [wxPython](https://docs.wxpython.org/wx.EllipsizeMode.enumeration.html)
7. [The systemd docs use it a bunch](https://www.man7.org/linux/man-pages/man1/systemctl.1.html) - "By default, this function only shows 10 lines of output and **ellipsizes lines** to fit in the terminal window"

On the other hand you have python with [textwrap.shorten](https://docs.python.org/3.4/library/textwrap.html#textwrap.shorten) in the stdlib, and an example for how to "[Truncate String with Ellipsis](https://css-tricks.com/snippets/css/truncate-string-with-ellipsis/)" in CSS.

My issue with "truncate" is that, to me, it only speaks to the "cutting off" part and is *still* not the easiest term?

"shorten" might be okay, but there we need to explain that we add an ellipsis (which python's textwrap calls a "placeholder" which I don't like?)

>much less likely to be understood by someone with limited english

Just as a sidenote: "ellipsis" comes from a greek term. That means in a bunch of western languages it will also be used - I know "ellipse" is a thing in german. So how "difficult" a term seems for native speakers doesn't always match how it seems for non-native speakers, because it might also be in their language.

In this specific case that's because it's a loan word, but e.g. non-native speakers might have a harder time with everyday kitchen terminology than grammar terminology.

floam:
It's easy to forget @faho is German, considering his perfect use of English.

floam:
I like it. I initially had some thoughts that `truncate` may be better, but when I think about it and how I'd use it, I really just want to ellipsize. Which may not be an actual word but I use it that way.

IlanCosman:
I'm not sure how fancy you want to get, but another option that people sometimes like is truncating the middle, 
i.e `foobarbaz` -> `foo…baz` Probably not worth implementing, but just thought I'd mention it.

mqudsi:
My gut instinct on reading the title was "why not just `string shorten`?" and then I saw that you considered that as an alternative.
I think ellipsize is too arcane of a name.

mqudsi:
In terms of prior art... Back in the day, I used to use the terribly named `PathCompactPath` function from `shlwapi.h` on Windows.

https://docs.microsoft.com/en-us/windows/win32/api/shlwapi/nf-shlwapi-pathcompactpathw

faho:
Alright, I renamed it to "shorten". It's shorter and simpler, and we don't need to be 100% accurate - people will figure out that it adds an ellipsis once they use it.

faho:
One more yak-shaving question: I've named the option to keep text from the end "--right".

`string pad` has "--right" to *add* padding to the right.

Should this be "--left" instead? My intuition was that I think of shortening as "go from the left/right, keep until you can't keep any more and then stop", while I think of padding as "add to the left/right until it's full".

IlanCosman:
I'm torn, but I'd say `--left`. I agree that I and probably most people think in terms of "keeping", but the command is called shorten, and objectively what you are shortening is the left side of the string.

faho:
Okay, renamed to "--left" - the idea is that we name it after where we change the string, like in `pad` and `trim`.

And with that, I'm merging this.

PR #9175, opened 2022-09-02
fix completions for tcpdump (-w and -r option)

sgtm0113:
## Description

Fix completions for tcpdump.

When I use tcpdump with "-w" or "-r" option, fish can't compete files to read.


sgtm0113:
I can't understand why the test on macos failed... :(

faho:
Merged, thanks!

>I can't understand why the check on macos failed... :(

The interactive tests are unfortunately slightly flaky on CI systems because they fundamentally rely on things happening within a certain time, and those systems have very bad latency characteristics. Nothing for you to worry about.

PR #9186, opened 2022-09-07
completions pipenv: updated activation command of built-in completion mechanism

exploide:
## Description

The internal completion mechanism of pipenv has changed. This PR updates the completions to work with the new invocation syntax.

By the way, I removed the `if command -sq pipenv` guard because I can't see where this is good for. The completion file should only be loaded when the command exists anyway? If not I can add that back.

## TODOs:

### DIFF
--- a/share/completions/pipenv.fish
+++ b/share/completions/pipenv.fish
@@ -1,7 +1,12 @@
-if command -sq pipenv
-    # pipenv determines which completions to print via some automagic that might not be perfect.
-    # Force it to be correct.
-    _PIPENV_COMPLETE=source-fish pipenv --completion 2>/dev/null | source
+# try new completion generation of click first
+set -l comps (_PIPENV_COMPLETE=fish_source pipenv 2>/dev/null)
+if test -z "$comps" || string match -q 'Usage:*' -- $comps
+    # fall back to older click-completion used in prior versions
+    set comps (_PIPENV_COMPLETE=source-fish pipenv --completion 2>/dev/null)
 end
 
+set -q comps[1] && printf %s\n $comps | source
+
+
+# manual workaround for pipenv run command completion until this is supported by the built-in mechanism
 complete -c pipenv -n "__fish_seen_subcommand_from run" -a "(__fish_complete_subcommand --fcs-skip=2)" -x


faho:
This is one of those changes (on the pipenv side) that I find annoying - it's pure busywork, and I would prefer they don't do it.

In this case, specifically: When was this changed? Are there old pipenv versions still in use that would be broken by this? Do we need to version-gate this?

> By the way, I removed the if command -sq pipenv guard because I can't see where this is good for.

I *think* the idea was that you might have pipenv2 and pipenv3, and they would wrap the base "pipenv". But since we don't have those completions, and python2 is dead, it's probably alright to remove.

floam:
Seems like the breaking change was [2021.11.9](https://github.com/pypa/pipenv/releases/tag/v2021.11.9).

exploide:
> When was this changed? Are there old pipenv versions still in use that would be broken by this? Do we need to version-gate this?

I did not investigate when this was changed yet. I don't use pipenv frequently. But I checked that an older version does not break in a way that there are errors or nonsense completions. There are no completions at all in this case. This is also not ideal but at least it doesn't explode.

If you are not comfortable with this, I could investigate when this has changed and try to version-gate this.

floam:
See usage for fish in https://click.palletsprojects.com/en/8.1.x/shell-completion/

exploide:
> See usage for fish at https://click.palletsprojects.com/en/8.1.x/shell-completion/

Switching from click-completion package to the new built-in completion logic of the click framework is generally a good idea. The former was less-well maintained and contained some insufficiencies especially concerning fish. When I saw what click provides now, I considered this a real improvement. However, the syntax change is still annoying and as far as I see, pipenv doesn't make use of the new features click provides, so the completions in this special case did not improve technically.

faho:
Blergh.

It would have been great to rename the variable and allow `--completions` so we could do a cross-compatible invocation.

`pipenv --version` takes upwards of 600ms so it's too slow to use.

Here's what we can do:

```fish
set -l comps (_PIPENV_COMPLETE=fish_source pipenv 2>/dev/null)
if string match -q 'Usage:*' -- $comps
    set comps (_PIPENV_COMPLETE=source-fish pipenv --completion 2>/dev/null)
end

set -q comps[1]; and printf %s\n $comps | source
```

This way it will attempt the new version, and if that doesn't work fall back to the old one. Unfortunately running just `pipenv` prints the help on stdout and returns 0, so we need to check which kind of output we got. Checking for "Usage:" (of `Usage: pipenv [OPTIONS] COMMAND [ARGS]...`) seems okay-ish. We don't want to try to `source` arbitrary help output. The biggest issue is if this was suddenly localized. Alternative needles include "--help", under the assumption that that one will always be in the help output and never in the completion script.

Also if someone wants to improve click's output, here's a freebie:

```fish
for value in (env _PIPENV_COMPLETE=fish_complete COMP_WORDS=(commandline -cp) COMP_CWORD=(commandline -t) pipenv);
        set response $response $value;
    end;
```

This is quite directly just

```fish
set response $response (env _PIPENV_COMPLETE=fish_complete COMP_WORDS=(commandline -cp) COMP_CWORD=(commandline -t) pipenv)
```

Even assuming you might not want `set -a` to support old fish versions.

exploide:
> Checking for "Usage:" (of Usage: pipenv [OPTIONS] COMMAND [ARGS]...) seems okay-ish.

Unfortunately I'm not sure you will really get the "Usage:" on older versions.

pipenv-version

faho:
That's easy enough - either "Usage:" or no output at all

exploide:
> That's easy enough - either "Usage:" or no output at all

True. I amended the commit.

faho:
Thanks!

PR #9190, opened 2022-09-08
Improve math errors

faho:
## Description

This introduces a specific error for division/modulo by zero and augments some error reporting with the length of the location.

Examples:

```fish
> math '1 - 1      pi'
math: Error: Missing operator
'1 - 1      pi'
      ^~~~~^
> math '2 + notafunction 5'
math: Error: Unknown function
'2 + notafunction 5'
     ^~~~~~~~~~~^
> math '1 - 1 / (5 - 5)'
math: Error: Division by zero
'1 - 1 / (5 - 5)'
       ^
```

Note: The division-by-zero error *technically* changes behavior, in that it now errors out early instead of carrying on and only complaining at the end that "Result is infinite" or "not a number".

That means something like

```fish
math min 1 / 0, 5
```

will now error out instead of returning 5.

I believe that is the correct choice - division by zero is typically expected to not be defined.

## TODOs:

### DIFF
--- a/src/builtins/math.cpp
+++ b/src/builtins/math.cpp
@@ -258,7 +258,12 @@ static int evaluate_expression(const wchar_t *cmd, const parser_t &parser, io_st
     } else {
         streams.err.append_format(L"%ls: Error: %ls\n", cmd, math_describe_error(error));
         streams.err.append_format(L"'%ls'\n", expression.c_str());
-        streams.err.append_format(L"%*ls%ls\n", error.position - 1, L" ", L"^");
+        if (error.len >= 2) {
+            wcstring tildes(error.len - 2, L'~');
+            streams.err.append_format(L"%*ls%ls%ls%ls\n", error.position - 1, L" ", L"^", tildes.c_str(), L"^");
+        } else {
+            streams.err.append_format(L"%*ls%ls\n", error.position - 1, L" ", L"^");
+        }
         retval = STATUS_CMD_ERROR;
     }
     return retval;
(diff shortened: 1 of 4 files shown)


ChristoferK:
In the context of a `math` function, which is only ever going to be called intentionally, by the user, to perform an arithmetic calculation, one would almost certainly want/need a division-by-zero to be handled gracefully, i.e. by returning a value.  This can be a numerical constant that represents infinity, or it can be a symbolic entity such as `INF`, `∞`, etc.  An exception is virtually never going to be helpful, and doesn't achieve anything that can't also be achieved following a graceful return.  Mathematically, division-by-zero occurs very naturally and often, to the extent that it's more common to treat it similar to a numerical quantity during interim calculations, particularly in physics when normalised infinities of the same magnitude can happily act to cancel each other out.

A benefit of having a means to handle division-by-zero mathematically rather than operationally is that the entity that is used to represent infinity can lend itself to some common programming scenarios (it's pretty common to implement a `min()` algorithm using infinity as a starting value against which the first element of a list is compared).  Consider the majority of programming languages: these _will_, of course, throw exceptions when a division-by-zero arises as a side-effect of some other operation; but for performing calculations, there's usually some sort of value or construct to represent unbounded values, mathematical singularities, infinities, and the like, both as a return value and as an entity available to the user (programmer) to use within calculations and other, broader scopes.  `C` (and `C++` too?) doesn't.  But JavaScript, Python, Ruby, Rust, Fortran, and pretty much all functional languages do.  `bash` doesn't, but thankfully `FiSH` has served to improve upon many of the things `bash` does poorly.

faho:
In the context of a scripting language, used by an ordinary person, it's true.

When they divide by zero, it's very very likely a mistake, and so the error is helpful.

If you want to "implement a min algorithm", use the "min" function. If you want to do more complicated math, use another tool.

Sorry, I don't want to change this.

PR #9199, opened 2022-09-10
Add the possibility to ignore arguments in alliases

Arkaeriit:
In other shells such as Bash, one can end its alias with `#` in order to make the alias ignore additional arguments. Example:

```bash
$ alias test='echo test #'
$ test abcd
test
```

Unfortunately, in Fish, this is not possible yet.

```fish
$ alias test='echo test #'
- (line 1): function: test: cannot use reserved keyword as function name
function test --wraps 'echo test #' --description 'alias test=echo test #';  echo test # $argv
^
from sourcing file -
        called on line 70 of file /usr/local/share/fish/functions/alias.fish
in function 'alias' with arguments 'test=echo\ test\ \#'
```

This PR modifies `alias.fish` to make it possible to create this kind of alias.

## TODOs:

- [ ] Changes to fish usage are reflected in user documentation/manpages. (Non applicable ?)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -97,6 +97,7 @@ Other improvements
 ------------------
 - The css for fish's documentation no longer depends on sphinx' stock "classic" theme. This should improve compatibility with sphinx versions and ease upgrading (:issue:`9003`).
 - The web-based configuration tool now works on systems that have ipv6 disabled (:issue:`3857`).
+- Aliases can ignore arguments by ending them with ``#``.
 
 
 For distributors
--- a/share/functions/alias.fish
+++ b/share/functions/alias.fish
@@ -67,7 +67,8 @@ function alias --description 'Creates a function wrapping a command'
         set wraps --wraps (string escape -- $body)
     end
 
-    echo "function $name $wraps --description $cmd_string; $prefix $body \$argv; end" | source
+    echo "function $name $wraps --description $cmd_string; $prefix $body \$argv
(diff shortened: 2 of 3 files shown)


faho:
You're going to have to adjust the tests as well - see tests/checks/alias.fish.

They print the semicolon so you probably just have to remove it.

The checks/ scripts are normal fish scripts, and their expected output is embedded in the `# CHECK:` and `# CHECKERR:` (for stderr) lines.

---

However, in general: Alias is a cheesy compatibility function. If you want an easier time of working with more control, try defining actual functions! It's much nicer because you don't need to fiddle with that additional layer of quoting, have proper syntax highlighting, can use many more constructs, ...! It's heavily recommended, and it's also been the recommended way to handle it in bash for years.

Arkaeriit:
The tests have been updated.

Thank you very much for your quick presentation of the tests in fish, it is heartwarming to see how welcoming you are with new contributors.

Thank you as well for the details on the best practice of fish use. I do use function for the wast majority of my aliases but there is some cases were a simple alias fell more suitable.

Arkaeriit:
Should the behavior added in this PR be documented? I feel like this is not necessary as this is a very niche behavior, it does not contradict the code example in the alias documentation, bash does not document the behavior, and functions are preferred over aliases. Yet, If you want me to document that in the alias documentation, I can do it.

faho:
>Should the behavior added in this PR be documented? I feel like this is not necessary as this is a very niche behavior, it does not contradict the code example in the alias documentation

Nah, it's fine.

faho:
Merged, thanks!

PR #9212, opened 2022-09-16
Document that break cannot be used in switch

mk12:
Fixed the docs for `break` which incorrectly stated it can be used in `switch`. Also clarified this in the docs for `switch`.

### DIFF
--- a/doc_src/cmds/switch.rst
+++ b/doc_src/cmds/switch.rst
@@ -19,7 +19,7 @@ Each ``case`` command is given one or more parameters. The first ``case`` comman
 
 Note that fish does not fall through on case statements. Only the first matching case is executed.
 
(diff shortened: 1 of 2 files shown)


doc_src/cmds/switch.rst
```diff
@@ -19,6 +19,8 @@ Each ``case`` command is given one or more parameters. The first ``case`` comman
 
 Note that fish does not fall through on case statements. Only the first matching case is executed.
 
+Note that :ref:`break <cmd-break>` cannot be used to exit a case statement early like in other languages. It can only be used in loops.
```
mqudsi, line 22: Can you change "case statement" to "case/switch block" to make it more searchable and easier for beginners to understand?

mk12: Sure, done.

mqudsi:
Merged with thanks.

PR #9220, opened 2022-09-19
Make directory-only wildcards (cd completions) faster on slow filesystems by keeping dir information

faho:
## Description

This uses wreaddir_resolving, which tries to use the dirent d_type
field if it exists. In that way, it can skip the `stat` to determine
if the given file is a directory.

This allows `cd` completions to skip stat in most cases:

```fish
strace -Ce newfstatat fish --no-config -c 'complete -C"cd /tmp/completion_test/"' >/dev/null
```

prints before:
```
% time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ----------------
100,00    0,002627           2      1033         4 newfstatat
```

after:

```
% time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ----------------
100,00    0,000054           1        31         3 newfstatat
```

for a directory with 1000 subdirectories.

(just `fish --no-config -c exit` does 26 newfstatat)

This should improve the situation with slow filesystems like fuse or
network fsen.

In case we have no d_type, we use `stat`, which would yield about the
same results.

The worst case is that we need directories *and* descriptions or the
"executable" flag (which we don't currently check for cd, if I read
this right?). I don't think that case exists or is all that common?

## TODOs:

### DIFF
--- a/src/wildcard.cpp
+++ b/src/wildcard.cpp
@@ -474,7 +474,7 @@ static bool wildcard_test_flags_then_complete(const wcstring &filepath, const wc
         return false;
     }
 
-    if (is_windows_subsystem_for_linux() &&
+    if (executables_only && is_windows_subsystem_for_linux() &&
         string_suffixes_string_case_insensitive(L".dll", filename)) {
         return false;
     }
--- a/src/wildcard.cpp
+++ b/src/wildcard.cpp
@@ -435,7 +435,16 @@ static const wchar_t *file_get_desc(int lstat_res, const struct stat &lbuf, int 
 /// up. Note that the filename came from a readdir() call, so we know it exists.
 static bool wildcard_test_flags_then_complete(const wcstring &filepath, const wcstring &filename,
                                               const wchar_t *wc, expand_flags_t expand_flags,
-                                              completion_receiver_t *out) {
+                                              completion_receiver_t *out, bool known_dir) {
+    const bool executables_only = expand_flags & expand_flag::executables_only;
+    const bool need_directory = expand_flags & expand_flag::directories_only;
+    // Fast path: If we need directories, and we already know it is one,
+    // and we don't need to do anything else, just return it.
+    // This is a common case for cd completions, and removes the `stat` entirely in case the system supports it.
+    if (known_dir && !executables_only && !(expand_flags & expand_flag::gen_descriptions)) {
+        return wildcard_complete(filename + L'/', wc, const_desc(L""), out, expand_flags,
+                                 COMPLETE_NO_SPACE) == wildcard_result_t::match;
+    }
     // Check if it will match before stat().
     if (wildcard_complete(filename, wc, {}, nullptr, expand_flags, 0) != wildcard_result_t::match) {
         return false;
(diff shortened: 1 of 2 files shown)


src/wildcard.cpp
```diff
@@ -435,7 +435,16 @@ static const wchar_t *file_get_desc(int lstat_res, const struct stat &lbuf, int
 /// up. Note that the filename came from a readdir() call, so we know it exists.
 static bool wildcard_test_flags_then_complete(const wcstring &filepath, const wcstring &filename,
                                               const wchar_t *wc, expand_flags_t expand_flags,
-                                              completion_receiver_t *out) {
+                                              completion_receiver_t *out, bool known_dir) {
+    const bool executables_only = expand_flags & expand_flag::executables_only;
+    const bool need_directory = expand_flags & expand_flag::directories_only;
+    // Fast path: If we need directories, and we already know it is one,
```
ridiculousfish, line 441: You can add another fast path: if we are `executables_only` and `known_dir` immediately return false!

ridiculousfish (approved):
This is great! Nice idea and perf improvement.

Curious why this is disabled under WSL?

One way to take this even further: the `is_dir` value is reliable, so if we have `directories_only && !is_dir` then we can return false immediately. This would require a three-way value (is dir, not dir, unknown) but is probably worth it since it would save additional stat calls for files which are not directories.

faho:
> the is_dir value is reliable,

unfortunately it's not in all cases. i tried throwing non-dirs out early and it ran afoul of one of the logical PWD tests.

we would have to absolutize that more eagerly for this to work, and I'm not sure that's worth it.

In fact I'm not convinced there isn't already a logical $PWD bug here ("what if ../foo is a directory physically but isn't logically"). I'll have to see if I can find it.

> Curious why this is disabled under WSL?

There's a check later for it, and the early-out checks simply mimic things we do anything special for later.

faho:
Ah, got it. Two issues:

1. expand_last_segment passes an empty base_dir to wreaddir_resolving, which makes it use "/" as the base dir
2. it doesn't normalize the actual *path*.

fstatat won't work here because we need the normalized path for logical $PWD.

We could use path_normalize_for_cd, but that is too ornery with what it accepts - leading and trailing slashes, and if you get it wrong it assert()s out.

faho:
Okay, benchmarks:

In a directory with 1000 directories and 1000 files, running `complete -C'cd '` yields:

| version | number of syscalls | time | relative |
| ------- | ---- |  ----------- | ------   |
| skipping early | 2215 | 10.3ms | 1.0x |
| not skipping early | 3215 | 13.8ms | 1.35x |
| lots of stat | 4213 | 16.2ms | 1.58x |

The syscall count for the skipping early version does a total of 5 "getdents64", 31 "newfstatat" and 18 "openat" calls. That's 5 newfstatat and 2 openat more than a pure `fish --no-config -c exit` - including function files like cd.fish or source.fish. 

Of the 2215 syscalls, 2000 are write() - we might want to have a look at that because it "write()"s the newline separately, which seems wasteful?

PR #9221, opened 2022-09-20
Preserve order of completion rules when generating completions

mqudsi:
This patch forces completions to be generated in the same order that they were declared (only relevant for completions that bypass sort, e.g. declared with the `-k` flag).

The arbitrary (reverse) sort order expected in `checks/complete.fish` when a bare `complete` is executed has been updated and a new test has been added verifying that `-k` completions for the same command with the same conditions respect their order of declaration.

 **If it's desirable for the output of a bare `complete` call (generating a list of all active completions) to be reverse-chronological, it can be iterated in reverse order without backing out this patch.**

As for the order of the completions generated when `tab` completing: IMHO, this is the intent for the majority of completions; I do not think many/most contributors expect that completions (especially those with `-k`) were actually being emitted in _reverse_ declared order.

@faho I've requested a review from you because a) IMHO this is now the correct order for `git checkout ^I` completions with changed files before tags and branches, etc. but I want to make sure that we're on the same page and defer to you here and b) this breaks the test for `checks/git.fish` and I'm not sure if the new output is  incorrect and indicative of something else gone wrong (this patch did change storage from `std::forward_list` to `std::vector`).

In particular, the final check `$fish -c 'complete -C "git diff -c"'` seems to expect no completions but now generates the following (unexpected) completions:

```
    -ca\tTreat all files as text <= no more checks
    -cB\tBreak complete rewrite changes into pairs of delete and create
    -cb\tIgnore changes in amount of whitespace
    -cC\tDetect copies as well as renames
    -cD\tOmit the preimage for deletes
    -cG\tLook for differences where <regex> matches the patch\'s added/removed lines
    -cM\tDetect and report renames
    -cR\tSwap inputs to create a reverse diff
    -cS\tLook for differences that change the number of occurrences of the string
    -cW\tShow whole surrounding functions of changes
    -cw\tIgnore whitespace when comparing lines
    -cz\tUse NULs as output field/commit terminators
    -cO\tControl the order in which files appear in the output
    -cl\tPrevents rename/copy detection when rename/copy targets exceed the given number
    -c1\tCompare the working tree with the \"base\" version
    -c2\tCompare the working tree with the \"our branch\"
    -c3\tCompare the working tree with the \"their branch\"
    -c0\tOmit diff output for unmerged entries and just show \"Unmerged\"
```

### DIFF
--- a/src/complete.cpp
+++ b/src/complete.cpp
@@ -76,7 +76,7 @@ static const wchar_t *C_(const wcstring &s) {
 static const wcstring &C_(const wcstring &s) { return s; }
 #endif
 
-/// Struct describing a completion option entry.
+/// Struct describing a completion rule for options to a command.
 ///
 /// If option is empty, the comp field must not be empty and contains a list of arguments to the
 /// command.
@@ -92,13 +92,13 @@ namespace {
 struct complete_entry_opt_t {
     // Text of the option (like 'foo').
     wcstring option;
-    // Type of the option: args-oly, short, single_long, or double_long.
+    // Type of the option: args-only, short, single_long, or double_long.
     complete_option_type_t type;
-    // Arguments to the option.
+    // Arguments to the option; may be a subshell expression expanded at evaluation time.
     wcstring comp;
     // Description of the completion.
     wcstring desc;
-    // Conditions under which to use the option.
+    // Conditions under which to use the option, expanded and evaluated at completion time.
     wcstring_list_t condition;
     // Determines how completions should be performed on the argument after the switch.
     completion_mode_t result_mode;
@@ -125,7 +125,7 @@ struct complete_entry_opt_t {
 static relaxed_atomic_t<unsigned int> k_complete_order{0};
 
 /// Struct describing a command completion.
-using option_list_t = std::forward_list<complete_entry_opt_t>;
+using option_list_t = std::vector<complete_entry_opt_t>;
 class completion_entry_t {
    public:
     /// List of all options.
@@ -139,15 +139,17 @@ class completion_entry_t {
     const option_list_t &get_options() const { return options; }
 
     /// Adds an option.
-    void add_option(complete_entry_opt_t &&opt) { options.push_front(std::move(opt)); }
+    void add_option(complete_entry_opt_t &&opt) { options.push_back(std::move(opt)); }
 
     /// Remove all completion options in the specified entry that match the specified short / long
     /// option strings. Returns true if it is now empty and should be deleted, false if it's not
     /// empty.
     bool remove_option(const wcstring &option, complete_option_type_t type) {
-        this->options.remove_if([&](const complete_entry_opt_t &opt) {
-            return opt.option == option && opt.type == type;
-        });
+        options.erase(std::remove_if(options.begin(), options.end(),
+                                     [&](const complete_entry_opt_t &opt) {
+                                         return opt.option == option && opt.type == type;
+                                     }),
+                      options.end());
         return this->options.empty();
     }
 
--- a/src/complete.h
+++ b/src/complete.h
@@ -59,6 +59,7 @@ using description_func_t = std::function<wcstring(const wcstring &)>;
 /// Helper to return a description_func_t for a constant string.
 description_func_t const_desc(const wcstring &s);
 
+/// This is an individual completion entry, i.e. the result of an expansion of a completion rule.
 class completion_t {
    private:
     // No public default constructor.
@@ -210,7 +211,7 @@ enum complete_option_type_t {
 void completions_sort_and_prioritize(completion_list_t *comps,
                                      completion_request_options_t flags = {});
 
-/// Add a completion.
+/// Add an unexpanded completion "rule" to generate completions from for a command.
 ///
 /// Examples:
 ///
--- a/tests/checks/complete.fish
+++ b/tests/checks/complete.fish
@@ -500,6 +500,21 @@ complete -C'oooops '
 echo $oops
 # CHECK: 1
 
+# Validate that completions are generated in the same order as the rules/calls to `complete` were
+# defined.
+
+function foo_to_complete
+end
+
+# Generate completions with `-k` to observe that sort order is preserved
+complete -c foo_to_complete -kf -a "opt_b_2 opt_b_1"
+complete -c foo_to_complete -kf -a "opt_a_1 opt_a_2"
+
+complete -C"foo_to_complete "
+# CHECK: opt_b_2
+# CHECK: opt_b_1
+# CHECK: opt_a_1
+# CHECK: opt_a_2
 
 # See that we load completions only if the command exists in $PATH,
 # as a workaround for #3117.
--- a/tests/checks/complete.fish
+++ b/tests/checks/complete.fish
@@ -55,8 +55,8 @@ complete
 # CHECK: complete --no-files complete_test_alpha1 -a '(commandline)'
 # CHECK: complete --no-files complete_test_alpha2
 # CHECK: complete --no-files complete_test_alpha3
-# CHECK: complete --force-files t -l fileoption
 # CHECK: complete --no-files t -a '(t)'
+# CHECK: complete --force-files t -l fileoption
 # CHECK: complete -p '/complete test/beta1' -s Z -d 'desc, desc'
 # CHECK: complete --require-parameter 'complete test beta2' -d desc\ \'\ desc2\ \[ -a 'foo bar'
 # CHECK: complete --exclusive complete_test_beta2 -o test -n false
@@ -368,8 +368,8 @@ complete -c banana
 # "-c" is optional
 complete banana -a bar
 complete banana
-#CHECK: complete banana -a bar
 #CHECK: complete banana -a '1 2 3'
+#CHECK: complete banana -a bar
 
 # "-a" ain't
 complete banana bar


faho:
So do I correctly understand: The idea behind this is simply to re-order the completions?

So if you have

```fish
complete -c foo -ka '3 2 1'
complete -c foo -ka 'c b a'
```

sorting as "c b a 3 2 1", you want it sorted as "3 2 1 c b a"?

I see an argument for that, but I'm not sold on changing it now - this exists and it is documented as such:

>Keeps the order of ARGUMENTS instead of sorting alphabetically. Multiple complete calls with -k result in arguments of the later ones displayed first.

(from https://github.com/fish-shell/fish-shell/blob/master/doc_src/cmds/complete.rst#L45-L46)

It seems weird to tell people writing fish completions (and they exist even outside of fish-shell/fish-shell) that they now need to swap the order.

It seems a bit callous to me and I would lean towards not doing it unless we have some good reason?

---

>In particular, the final check $fish -c 'complete -C "git diff -c"' seems to expect no completions but now generates the following (unexpected) completions:

On my system, without this PR, that tries completing like an ordinary `git diff` positional argument. So it tries to complete branches and such, and it ends up matching those with `-c` as a substring.

This, on the other hand, appears to interpret `-c` as an option and suggests more short options to group onto it (`ls -lah` style).

Typically, for unknown options, we don't do this, so it seems to recognize `-c`, which implies that it's messing up the conditions.

Here are all the lines that add a `-c` short option candidate:

```fish
complete -x -c git -n __fish_git_needs_command -s c -a '(__fish_git config -l 2>/dev/null | string replace = \t)' -d 'Set a configuration option'
complete -f -c git -n '__fish_git_using_command am' -s c -l scissors -d 'Remove everything in body before scissors'
complete -f -c git -n '__fish_git_using_command branch' -s c -l copy -d 'Copy branch'
complete -f -c git -n '__fish_git_using_command grep' -s c -l count -d 'Instead of showing every matched line, show the number of lines that match'
complete -c git -n '__fish_git_using_command log' -s c
complete -c git -n '__fish_git_using_command ls-files' -s c -l cached -d 'Show cached files in the output'
complete -f -c git -n "__fish_git_using_command notes" -n "__fish_seen_subcommand_from add append" -s c -l reedit-message -a '(__fish_git_commits)' -d 'Copy and edit note from object'
complete -f -c git -n '__fish_git_using_command switch' -r -s c -l create -d 'Create a new branch'
complete -f -c git -n '__fish_git_using_command stripspace' -s c -l comment-lines -d 'Prepend comment character to each line'
complete -f -c git -n '__fish_git_using_command blame' -s c -d 'Use the same output mode as git-annotate'
```

I.e. when git has no subcommand, or when the subcommand is one of "am branch grep log ls-files notes switch stripspace blame".

The most likely culprit is `__fish_git_needs_command`.

mqudsi:
@faho I didn't realize the current behavior was documented. I guess I find the proposed change to the behavior to be the intuitive case, but if this is going to break a lot of completions (note: break as opposed to change) then it's probably best not to merge this. I think most people would naturally write the "most important" completions first then continue on to the less and less likely cases instead of doing it in backwards order. Though I do wonder if there's an argument to be made for the current behavior (if it were a greenfield scenario), perhaps there is something I'm missing.

---

As for the issue with the differing completions behavior that shouldn't result from merely changing the order in which they are iterated:

It seems that the fault lies in the existing logic in `src/complete.cpp`, which juggles between an inner and outer loop. Officially, the only way that iterating over the completions in a different error can cause different results is if a completion causes the completer to reach the maximum number of supported completions, in which case any completion sources that haven't yet been iterated will be skipped. 

But it turns out there is an existing issue that isn't being handled cleanly - if there are multiple completions matching a commandline **regardless of whether or not their conditions are met**, the `last_option_requires_param` flag can bounce back and forth between true and false, settling on whatever the last evaluated `complete` rule was. If `last_option_requires_param` settled on `true`, the subsequent attempt to generate non-switch completions can falsely skip a completion source early because it thinks the last item on the command line is a short option completion requiring a parameter and the current possible completion is a different switch, so it won't offer it.

In your case, you have `complete -x -c git -n __fish_git_needs_command -s c` saying "commandline ends on a short opt, don't offer any switches up as possible completions" but then, e.g. `complete -c git -n '__fish_git_using_command ls-files'` comes along and says `-c` isn't a shortopt requiring a completion, go ahead and offer other switches if you want. Fish doesn't check if the conditions for `ls-files` were met before adopting its position, so the "requires a command" flag is incorrectly overridden. 

The first thing is to only set `last_option_requires_param` if the conditions are a match, but the more complicated question is if a user (incorrectly, I think) specifies two separate completions that both complete the same shortopt and both have their conditions met but one completions says it requires a param and the other doesn't. In that case, I think `last_option_requires_param = o.result_mode.requires_param;` should be `last_option_requires_param = last_option_requires_param && o.result_mode.requires_param` so that if there are multiple conflicting options, we default to suggesting both possible completions for the parameter payload of the shortopt plus possible completions for other switches and the user has to figure out which is correct.

krobelus:
I find input order more natural as well and thought about making the same change Hard  to gage how important compat is here.

faho:
So, my issue is this:

I don't think that adding new completions to the front of the queue is fundamentally unreasonable. You add new completions, they have to go *somewhere*.

For instance, in our git completions we have (slightly simplified):

```fish
complete -f -c git -n '__fish_git_using_command show' -a '(__fish_git_branches)'
complete -f -c git -n '__fish_git_using_command show' -ka '(__fish_git_tags)' -d Tag
complete -f -c git -n '__fish_git_using_command show' -ka '(__fish_git_commits)'
```

You get tags - "e d f", and you get commits, "2 3 1", and it ends up as "2 3 1 e d f".

Now, if I had designed this from the ground up, and had to ask myself which order these should go in, I probably would have decided to make it "e d f 2 3 1".

But if we change this *now*, this means anyone writing an external completion script for other people to consume and who cares about the order here (and if they use `--keep-order` they kinda do) needs to do

```fish
if test $version -gt 3.5.1
    complete -f -c git -n '__fish_git_using_command show' -ka '(__fish_git_tags)' -d Tag
    complete -f -c git -n '__fish_git_using_command show' -ka '(__fish_git_commits)'
else
   complete -f -c git -n '__fish_git_using_command show' -ka '(__fish_git_commits)'
   complete -f -c git -n '__fish_git_using_command show' -ka '(__fish_git_tags)' -d Tag
end
```

and that's not even the actual version check!

Someone who doesn't know about this will just write one of these and then the order will be wrong for any user with a fish version on the other side of the transition. And if they read old documentation they'll see that the order is this, and if they read new documentation they'll see that the order is that - or, better, "the order is now this, but you have to write all this extra guff because it used to be that". And that's annoying documentation to read and to write and to understand.

And if they attempt to write it for both versions they might get the version check wrong and then it's broken in the other version anyway. Or they'll get the version check right for *released versions* like "3.5.1", but they won't have it right for "3.5.1-320-g168d74ab0-dirty" that I could build from git right now, and so it'll be broken for me.

For many projects fish support is somewhere way down on the priority list, something done by drive-by committers who sometimes have trouble finding interested reviewers. I would like to make this reasonably easy on them so we can get good fish support, and so I would like to place a bit of a higher bar on changes like this. It needs to be enough of an improvement to justify the hassle, and I don't think this is.

faho:
>But it turns out there is an existing issue that isn't being handled cleanly - if there are multiple completions matching a commandline regardless of whether or not their conditions are met, the last_option_requires_param flag can bounce back and forth between true and false, settling on whatever the last evaluated complete rule was.

Neat, nice catch!

> The first thing is to only set last_option_requires_param if the conditions are a match, but the more complicated question is if a user (incorrectly, I think) specifies two separate completions that both complete the same shortopt and both have their conditions met but one completions says it requires a param and the other doesn't.

This is, indeed, incorrect. This is Schrödinger's option parameter. Tbh I don't think it's worth thinking too much about because of that. I'd probably lean towards requiring the parameter because that's easier to see, and figure out that your completions make no sense.

mqudsi:
_The option completion bug has been resolved and merged into `master`, and this branch has been rebased on top of it (and therefore now passes all tests)._

This PR isn't a hill I'm going to pick to die on but here are my thoughts in response to the very good points you raised:

* I think the real-world impact of this change is a lot smaller than we're making it out to be. This doesn't affect just "anyone using `-k`" because most `-k` uses are confined to a single `complete` call for an option with all its arguments and conditions specified in one go and will continue to work as always. This change only affects the ordering of completions _across multiple `complete -k` declarations_ where they are intended to additively provide completions for the *same* option (possibly under different conditions) pulling from multiple sources and breaking that up into multiple `complete` invocations rather than putting them all in one `-a` payload.
* `-k` arose from a real need and is self-evident; anyone basing their completions off a sample fish completion that preserves order will see the `-k` and can know to look it up. Similarly, it's a top-level switch documented in the `complete` help synopsis so it's hard to miss. Whereas if you're inspecting a sample fish completion, the order of completions *at runtime* offered across multiple `complete -k` invocations isn't in any way "visible" (unless there happens to be a comment saying "these completions will be offered first because we add them last") and so it's not as likely to be picked up.
* The importance of ordering of completions *across `complete` calls* for the same option is not immediately obvious when testing a completion; if the order of what `-a` evaluated to is jumbled/sorted (i.e. commits are presented in alphabetical rather than chronological order) it's immediately obvious that there is an issue with the completion and people will seek out an answer for that; whereas if there are multiple `complete -k` calls providing completions to the same cmd+option, the results are still "acceptable" just less usable or ergonomic and only the most OCD fish dev will try to fix that. In fact, if there are multiple `complete -k` calls to complete a single cmd+option, there are likely distinct conditions for each one (as with some of our git completions) and it's perfectly plausible that a fish dev tested their completions and didn't even run into the case where the conditions for both, disparate `complete -k` calls were simultaneously true, and if they were, it (again) might not even have occurred to them that there's anything wrong with the order.
* The overall ordering of "groups" of order-preserved completions is extremely unlikely to be important enough for someone to ship multiple implementations based off the fish version. It's not like there's an actual bug, fish crashes, the completions don't work, etc. if the *overall* group order is reversed to what was intended: everything continues to work and most end users will probably not even realize that there's anything different and almost certainly won't complain to the dev for the dev to implement a version-based completion scheme. 
* My final and most important point: fish strives to be, where possible, intuitive. Some things (like conveying to the completion system that you wish to preserve the order of `complete -a` payloads) require explicit annotation and there's nothing we can do about it. But I can't find any reason why anyone would intuitively expect that completions would be generated in the opposite order they were written/loaded, and I think that fish should follow that intuition (especially since, at least so far as I can tell and have outlined above, the cost of breaking the pre-existing behavior now isn't a huge deal).

That said and to be fair, that line was added to the documentation in May 2019 and has been in the fish docs since fish 3.1.0 - I had been hoping it was added more recently than that.

-----

@faho Completely orthogonal to this change, is it your intention that modified files come after all other possible `git checkout` completions? Logically (to me!), file completions are predicated on there being changed files so I would _assume_ **in the presence of uncommitted changes** that a `git checkout ` call would be most likely to revert or partially revert uncommitted changes rather than to checkout a different tag/branch/commit. 

It's actually the reason I went down this rabbit hole, trying to understand why the `git checkout ` completions weren't (to me) very useful when `cd`d into the `fish-shell` repo, as I had checked in the completions script and confirmed that the `complete` call to provide modified files came before the `complete -k` calls to provide tags, branches, and commits. (Even though I was the one that had added `complete -k` support to fish, in my mind the correct order was the opposite of what was specified/exhibited!)

faho:
>Completely orthogonal to this change, is it your intention that modified files come after all other possible git checkout completions? Logically (to me!), file completions are predicated on there being changed files so I would assume in the presence of uncommitted changes that a git checkout call would be most likely to revert or partially revert uncommitted changes rather than to checkout a different tag/branch/commit.

I don't remember. If you want to change that order, go ahead!

faho:
I'm tentatively gonna close this because it already has a conflict with master and I'm against it.

mqudsi:
I left it open hoping you'd reconsider. Like I said, I'm not going to fight anyone on this, but I genuinely believed (and still do) no one writes completions expecting they'll be shown in reverse order and my perusal of the git completions reinforces that.

No worries though! 

In all cases, I was waiting on this to either be merged or closed before I fixed the order of git completions (at least for `git checkout`). I can do that now!

krobelus:
I feel it's okay to do the breaking change because I assume that
- this is a very specific case, probably very few users are affected
- even if a user is affected, it's only the order of completions that changes, likely no big deal

.. but I haven't read through the comments or done the research so I'll leave it to you guys and weigh in later

Issue #7044, opened 2020-05-27
Using aliases with sudo


nihaals:
I was wondering if there's an equivalent to `alias sudo='sudo '` which allows `sudo x` to use an alias for `x`. The reason why something like `sudo -s` wouldn't work is that the alias changes `sudo apt-get` to `sudo apt` for any relevant subcommand which can be useful when copying examples.

I'm using fish 3.1.2 and tried `alias sudo 'sudo '` but had no success.

faho:
There is not, and it would be a massive hack. Use abbreviations instead.

Nihaal Sangha <notifications@github.com> schrieb am Mi., 27. Mai 2020,
14:53:

nihaals:
Since abbreviations only work at the start and can't contain spaces I don't think this would work (otherwise I could add one for `sudo apt-get install` for example), which I think is related to #5003 although it couldn't require a different input e.g. `Ctrl + Space` but instead just on paste. I guess this is a feature request that either expands on #5003 or one to allow for abbreviations with spaces.

faho:
Okay, what do you want to *do* here?

How do you want to override `apt-get`?

nihaals:
As I’m very new to fish I have no idea what the best course would be. My [current solution](https://gist.github.com/nihaals/12c98546531e7df3b6c3614e6e2030dc) completely replaces it then either calls apt or if there isn’t an equivalent, just calls what you gave it, but I think with fish there could be a more efficient way

faho:
Okay, I would just make this a script.

Call it "myapt", give it a shebang (which you already have), put it in $PATH and use it.

(Or, if you have to call it `apt` it needs to call the actual apt with a full path, e.g. `/usr/bin/apt`)

nihaals:
Wouldn’t it need to be called `apt-get`? Also, if they both have the same name how would it decide to use my proxy version instead of the original when I would just be calling `apt-get`?

nihaals:
I think I might have been unclear with my goal, the point is so when I copy something from a README for example, if it uses `apt-get`, it would still use `apt` anyway, effectively replacing `apt-get` with `apt` without changing how you use it

faho:
>Wouldn’t it need to be called apt-get?

It would need to be called whatever you call it. You would of course then have it *call* "apt-get" instead of its name.

>Also, if they both have the same name how would it decide to use my proxy version instead of the original when I would just be calling apt-get?

[...]

PR #9225, opened 2022-09-20
Add explanation on abbreviations vs aliases with `sudo`

lucatrv:


faho:
This isn't actually *correct* as such.

If you have an abbreviation "v" for "vim", you can't really do `sudo v`, because we currently only expand abbreviations in command position.

That makes this misleading, which is why I'm rejecting this. Thanks anyway!

lucatrv:
Sorry I realized it only now! I was referring to issue #7044 and for some reason got convinced that abbreviations were always expanded... now I tested it and found out that it's not the case. Could this be addressed in future, for instance by expanding them not only in command position but also after `sudo`? Or otherwise do you have any other advice that could be reported in the documentation on how to correctly work with `sudo` since aliases cannot be used? Thanks

faho:
>Could this be addressed in future, for instance by expanding them not only in command position but also after sudo?

#5003

faho:
>  Or otherwise do you have any other suggestion that could be reported in the documentation on how to correctly work with sudo since aliases cannot be used?

Enter whatever you want, press alt+s. This adds or removes a `sudo` prefix.

Use abbreviations that already have `sudo` included or start a root shell. Whichever works for you.

PR #9226, opened 2022-09-20
Add reference to abbreviations

lucatrv:


doc_src/tutorial.rst
```diff
@@ -599,6 +599,8 @@ For more, see :ref:`Functions <syntax-function>`.
 
 .. [#] There is a function called :ref:`alias <cmd-alias>`, but it's just a shortcut to make functions.
 
+.. [#] fish also provides :ref:`abbrevisations <interactive-abbreviations>`, through the :ref:`abbr <cmd-abbr>` command.
```
mqudsi, line 602: You have a typo here.

mqudsi (changes requested):

mqudsi:
I’m not sure if we need two footnotes here. Can’t we just merge both into one? Happy to hear your thoughts.

lucatrv:
> I’m not sure if we need two footnotes here. Can’t we just merge both into one? Happy to hear your thoughts.

Yes I agree they could be merged into a single footnote. IMHO the only drawback is that abbreviations would be a little less evident in the text, while I think they are very useful and maybe they would deserve even a full paragraph of the tutorial in future... Bottom line, for me both options are ok, please let me know which one you prefer.
Thanks for asking my opinion BTW.

lucatrv:
> Ok, let's merge the paragraphs for now and we can always write more about `abbr`, move it to its own paragraph, etc. in the future.

Please let me know if I need to do anything else

mqudsi:
Sorry, I can see how my comment could have been taken to signal merging the PR. I meant can you merge the two footnotes into one paragraph.

lucatrv:
> Sorry, I can see how my comment could have been taken to signal merging the PR. I meant can you merge the two footnotes into one paragraph.

Oh, my bad! :)

ridiculousfish:
Merged as 064699cade2bba9f37b1a90317f7d370b8481030, thank you.

ridiculousfish:
I implemented mqudsi's feedback and merged the footnotes together.

PR #9229, opened 2022-09-21
Reduce write()s for builtins

faho:
## Description

Our io_streams_t is, when it comes to pipes (including the terminal), an extremely thin wrapper over `write(2)`.

Whenever you do `streams.out.append()` or `streams.out.push_back()` that corresponds to one `write(2)`. This means doing something like

```c++
for (wchar_t c : somestring) {
    streams.out.push_back(c);
}
```

is extremely slow. The pattern we see a lot is

```c++
streams.out.append(somestring);
streams.out.push_back(L'\n');
```

So, this PR tries to reduce how often we do that, sometimes by constructing strings on-the-fly like

```c++
streams.out.append(somestring + L"\n");
```

Or doing some buffering.

I'd like to call out one egregious case, which is `printf`. This one would `out.push_back` for every character escape, so

```fish
printf (string repeat -n 200 \\x7f)%s\n (string repeat -n 2000 aaa\n)
```

would issue 400000 (four hundred thousand) `write(2)` calls. Now it has been reduced to 2000 (one per argument excluding the format string - I experimented with doing one big write but that turned out to be slower in some cases and would of course have worse latency and memory requirements). This used to take ~400ms, now it takes 16ms.

The rest of the wins are more modest, ranging from ~1.1x to 1.7x.

## TODOs:

### DIFF
--- a/src/builtins/builtin.cpp
+++ b/src/builtins/builtin.cpp
@@ -101,11 +101,8 @@ maybe_t<int> builtin_builtin(parser_t &parser, io_streams_t &streams, const wcha
         wcstring_list_t names = builtin_get_names();
         std::sort(names.begin(), names.end());
 
-        for (const auto &name : names) {
-            auto el = name.c_str();
-
-            streams.out.append(el);
-            streams.out.append(L"\n");
+        for (auto &name : names) {
+            streams.out.append(name + L"\n");
         }
     }
 
--- a/src/builtins/complete.cpp
+++ b/src/builtins/complete.cpp
@@ -416,14 +416,14 @@ maybe_t<int> builtin_complete(parser_t &parser, io_streams_t &streams, const wch
                     // we need to unescape the command line. See #1127.
                     unescape_string_in_place(&faux_cmdline_with_completion, UNESCAPE_DEFAULT);
                 }
-                streams.out.append(faux_cmdline_with_completion);
 
                 // Append any description.
                 if (!next.description.empty()) {
-                    streams.out.push_back(L'\t');
-                    streams.out.append(next.description);
+                    faux_cmdline_with_completion.push_back(L'\t');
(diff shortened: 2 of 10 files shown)


mqudsi:
These changes look good and the wins are appreciable. I had a concern about excessive buffering but we already eschew builtins for cases where fish's internal pipeline buffering causes burdensome delays (such as using `sed` instead of `string ...` in some completions) so I don't think there's any risk of any long-running processes piping to/from builtins effectively becoming less usable if we increase buffering (and I think the additional buffering won't even matter in the cases where all output is buffered in one go, anyway).

ridiculousfish:
This looks fine to me, but it does impose some awkwardness on the builtins.

A more pleasant-to-use but complex alternative would be to support buffering directly in `output_stream_t`, example:

     {
          autobuffer_t buffer = streams.out.autobuffer();
          buffer.append(name);
          buffer.append('\n');
     }

and in buffer's destructor, it flushes its data to its origin stream, in one go.

Either that approach or what you've done here is fine with me, just wanted to toss that out as an alternative.

faho:
>A more pleasant-to-use but complex alternative would be to support buffering directly in output_stream_t, example:

I have thought about that, but it's not really any less effort if you want that line-buffering. Basically you would replace the `streams.out.append(line)` with `buffer.flush()` anyway.

I don't really want to buffer by *length*, because I'm imagining the latency characteristics might be weird? What if we have 490 bytes ready, we flush every 500 bytes and the next 10 bytes take a second because this is a super tricky regex match?

What would be possible would be a "append_line" that appends the `\n` itself, for simple cases. But even that's barely clearer than `append(str + "\n")`

mqudsi:
Regarding a possible `append_line()`: it is/was my intention to add `writev()` support (where available) and expose that via a variadic `append(...)` and `append_line()` functions.

PR #9232, opened 2022-09-24
fish_clipboard_copy: use OSC 52 when running inside SSH

krobelus:
When connecting to a remote system via SSH, Control-X either does nothing,
or if a clipboard-copying program is installed, tries to copy into the remote
system's clipboard, which often fails.

Modern terminal emulators implement the OSC 52 escape sequence for setting
the clipboard of the terminal host system. Use OSC 52 in fish_clipboard_copy
when running inside SSH.  Give it precedence over all other cases since that
behavior is probably better than setting the clipboard of the remote system.

Tested on foot, iTerm2, kitty, [tmux] and xterm (which requires
"XTerm.vt100.allowWindowOps: true").

Should also work in screen and Windows Terminal. On terminals that don't
support OSC 52 (like Gnome Terminal or Konsole), it seems to do nothing (but
with this patch we also don't delegate to pbcopy and friends, so I hope no
one relies on that).

I think there is also an escape sequence to request pasting the system
clipboard but that's less important and less popular, possibly due to
security concerns.

[tmux]: https://github.com/tmux/tmux/wiki/Clipboard

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -83,6 +83,7 @@ Interactive improvements
 - Fish now disables the QUIT terminal sequence when it has the terminal. This frees up a key combination, often ctrl-backslash (``\x1c``) (:issue:`9234`).
 - Fish's vi mode no longer uses iTerm's proprietary escape sequences to signal cursor change, instead using the normal xterm-style sequences. This allows for a blinking cursor and makes it work in complicated scenarios with nested terminals. (:issue:`3741`, :issue:`9172`)
 - Generating descriptions for commands now uses ``manpath`` instead of ``man --path`` on macOS, as that has been removed in macOS Ventura.
+- When running fish on a remote system (e.g. inside SSH or a container), :kbd:`Control-X` now copies to the local client system's clipboard if the terminal supports OSC 52.
 
 
 Fixed Bugs
--- a/share/functions/fish_clipboard_copy.fish
+++ b/share/functions/fish_clipboard_copy.fish
@@ -23,4 +23,15 @@ function fish_clipboard_copy
     else if type -q clip.exe
         printf '%s' $cmdline | clip.exe
     end
+
+    # Copy with OSC 52; useful if we are running in an SSH session or in
+    # a container.
+    if type -q base64
+        if not isatty stdout
+            echo "fish_clipboard_copy: stdout is not a terminal" >&2
+            return 1
+        end
+        set -l encoded (printf %s $cmdline | base64 | string join '')
+        printf '\e]52;c;%s\a' "$encoded"
+    end
 end
--- a/share/functions/fish_clipboard_copy.fish
+++ b/share/functions/fish_clipboard_copy.fish
@@ -33,5 +33,16 @@ function fish_clipboard_copy
         end
         set -l encoded (printf %s $cmdline | base64 | string join '')
         printf '\e]52;c;%s\a' "$encoded"
+        # tmux requires user configuration to interpret OSC 52 on stdout.
+        # Luckily we can still make this work for the common case by bypassing
+        # tmux and writing to its underlying terminal.
+        if set -q TMUX
+            set -l tmux_tty (tmux display-message -p '#{client_tty}')
+            or return 1
+            # The terminal might not be writable if we switched user.
+            if test -w $tmux_tty
+                printf '\e]52;c;%s\a' "$encoded" >$tmux_tty
+            end
+        end
     end
 end


mqudsi:
Johannes and I spoke about this in person and it's an awesome feature but I have concerns about not calling the traditional fallbacks if the OSC escape isn't supported. I know we don't do terminal-identification as a general rule, but I think we need to do something because this could break a lot of configs. 

FWIW: I had to look it up, but xterm's `XTerm.vt100.allowWindowOps` defaults to true.

@krobelus does "Windows Terminal" mean just that app or also cmd.exe?

krobelus:
Ok there are several cases:
1. `ssh localhost` - yes, this one will be broken if the terminal doesn't support OSC 52. We could work around this though I'm not sure how? Probably not a big deal since it's a nonsensical use case.
2. `ssh remote-host`. Here it means we stop writing the clipboard of the remote system but who wants that anyway?
  0. I guess this could matter for some remote desktop setups, but wouldn't the terminal then also run in remote desktop, so without SSH_PTY defined?
  1. I failed to think about X11 forwarding, which means remote xclip uses the local X server. Luckily we can just detect if DISPLAY is defined and skip OSC 52 if yes.
  2. I haven't tested with remote macOS, but I can imagine that `pbcopy`/`pbpaste` work (but on the remote system)? Though that's a skewed expectation; using the local clipboard is more correct.

> FWIW: I had to look it up, but xterm's `XTerm.vt100.allowWindowOps` defaults to true.

On my system it doesn't :(

> does "Windows Terminal" mean just that app or also cmd.exe? 

I'm not sure but probably just the app since it's more modern.

krobelus:
I'm not the Windows expert in the room, is it possible to `ssh` into windows and use the clipboard there?

If there's a chance we break anyone, we can just do both: the old behavior, and additionally write the OSC 52 sequence.

share/functions/fish_clipboard_copy.fish
```diff
@@ -2,7 +2,11 @@ function fish_clipboard_copy
     # Copy the current selection, or the entire commandline if that is empty.
     set -l cmdline (commandline --current-selection | string collect)
     test -n "$cmdline"; or set cmdline (commandline | string collect)
-    if type -q pbcopy
+
+    if set -q SSH_TTY; and not set -q DISPLAY
+        set -l encoded (printf %s $cmdline | base64 | string join '')
+        printf '\e]52;c;%s\a' "$encoded" >/dev/tty
```
faho, line 8: What if `/dev/tty` doesn't exist?

IIRC it's a linuxism?

mqudsi: `/dev/tty` isn't available under BSD.

mqudsi:
> ssh remote-host. Here it means we stop writing the clipboard of the remote system but who wants that anyway?

@krobelus I'm not sure if you remember as we spoke about this in person in passing, but it's not safe to assume `fish_clipboard_copy` and `fish_clipboard_paste` are strictly for the local machine. 

In my minimal fish config, one of the few things I override is fish_clipboard_copy (indirectly, by defining my own `pbcopy` and `pbpaste`) to support remote copy-and-paste over SSH:

https://github.com/mqudsi/fish-config/blob/12b7ebb5bfb049807e583ce76dc8b4eb89553deb/functions/pbcopy.fish

```fish
# Implement remote copy-and-paste via `lemonade` (a go utility) - note that it doesn't support IPv6:
if type -q lemonade; and string length -q $SSH_CLIENT; and not string match -rq ':' $SSH_CLIENT
	set -l host (echo $SSH_CLIENT | cut -f1 -d ' ')
	lemonade --host $host copy 2>/dev/null
	return
end
```

I'm fairly sure I'm not the only one doing something along these lines, even if not everyone has their fish config on GitHub for us to find. 

I'm all for adding support for this escape, but given the patchy client support I think we would have to continue calling `fish_clipboard_copy`. There are issues if they behave differently, e.g. one could trim trailing empty lines as a function of the utility it goes through and the other might not, etc. and I'm not sure how we should resolve that race condition (if it's a race at all; after all, the order we execute them in is deterministic) but that would be better than not calling `fish_clipboard_copy` at all.

I'm also sorry to learn that `XTerm.vt100.allowWindowOps` doesn't default to true on your machine - what's your distribution? It seems that true is the (current?) [default value upstream](https://www.x.org/releases/X11R6.7.0/doc/xterm.1.html#:~:text=allowWindowOps%20(class%20AllowWindowOps,default%20is%20%60%60true.%27%27) but (though saddened) I'm not surprised to learn that it depends on the packager. If we don't even have basic xterm support out-of-the-box that's definitely all the more reason to not rely on the OSC escape on its own.

share/functions/fish_clipboard_copy.fish
krobelus: Right we can probably write to `$SSH_TTY` instead

actually it also seems to work if when write to stdout, that seems ideal

faho: It works writing to stdout unless you do `fish_clipboard_copy | cat`.

Probably just error out if `isatty stdout` fails?

faho:
>In my minimal fish config, one of the few things I override is fish_clipboard_copy (indirectly, by defining my own pbcopy and pbpaste) to support remote copy-and-paste over SSH:

If you override fish_clipboard_copy, you can override fish_clipboard_copy some more?

As I see it: The current copy/paste over ssh is typically useless, and OSC 52 has at least a fighting chance of doing something useful instead.

Of course if you have a better way of doing it on your systems, this doesn't stop you from doing that. It's a function, it's super easy to customize!

mqudsi:
Fair enough!

mqudsi:
I don't think it's a big deal at all, but just fyi and from previous experience, this will cause local fish copy/paste to fail if used locally on a machine with a tmux session that was started remotely (and, vice versa, won't perform remote copy and paste if the tmux session was started locally but that's less of an issue). Again, super niche and not a blocker, but just for posterity.

krobelus:
> what's your distribution?

Arch Linux, [here](https://github.com/archlinux/svntogit-packages/blob/eaa794478cde8da5cd3088f78fab3a9af18405b0/trunk/PKGBUILD) is the xterm build configuration.
Probably some configure flag is missing but I don't really care.

> In my minimal fish config, one of the few things I override is fish_clipboard_copy (indirectly, by defining my own `pbcopy` and `pbpaste`) to support remote copy-and-paste over SSH

Actually I did the same thing - I created a function `pbcopy` that writes the OSC 52 sequence.
It's very hacky but it's the solution with the fewest lines :)

Not sure if we should keep such configurations working. I guess it's basically free.
No strong opinion here.

> this will cause local fish copy/paste to fail if used locally on a machine with a tmux session that was started remotely

Good find. When using detachable terminals, environment variables are not flexible enough to tell the origin of a terminal.
A future terminal protocol extension could do that.
Though the easiest solution is for terminals to support OSC 52, then there's no problem.

share/functions/fish_clipboard_copy.fish
krobelus: > Probably just error out if `isatty stdout` fails?

Ok. Maybe more elegant to write to $SSH_TTY but that might have spooky failures after tmux re-atttach, I'm not sure

CHANGELOG.rst
```diff
@@ -66,6 +66,7 @@ Interactive improvements
 - The :kbd:`Alt-H` binding will now show the manpage of the command under cursor instead of the always skipping ``sudo`` and the likes (:issue:`9020`).
 - If ``$fish_color_valid_path`` contains an actual color instead of just modifiers, those will be used for valid paths even if the underlying color isn't "normal" (:issue:`9159`).
 - Performance improvements to highlighting (:issue:`9180`) and the cd completions (:issue:`9220`) should make using fish more pleasant on slow systems.
+- When running fish inside SSH, :kbd:`Control-X` now copies to the local system's clipboard if the terminal supports OSC 52.
```
mqudsi, line 69: I would change "local system's" to "SSH client's" because it could be confusing what is local and what is remote.

krobelus:
This is more complicated than expected.
I realized I still haven't managed to configure tmux correctly - it only
worked when writing to /dev/tty - I used a similar workaround that also
works if a tmux session was started somewhere else.
I've also left pbcopy in for now and fixed some cases where su/sudo would break fish_clipboard_copy.

share/functions/fish_clipboard_copy.fish
krobelus: > `/dev/tty` isn't available under BSD.

Huh. Apparently it's even part of [POSIX](https://pubs.opengroup.org/onlinepubs/000095399/basedefs/xbd_chap10.html) and available on macOS.

share/functions/fish_clipboard_copy.fish
```diff
@@ -13,4 +14,24 @@ function fish_clipboard_copy
...
         printf '%s' $cmdline | clip.exe
     end
+
+    # Copy with OSC 52. Primarily for SSH, but don't rely on SSH_TTY because
+    # that's dropped by su/sudo.
+    if not isatty stdout
+        echo "fish_clipboard_copy: stdout is not a terminal" >&2
+        return 1
+    end
+    set -l encoded (printf %s $cmdline | base64 | string join '')
```
faho, line 24: Technical not: `base64` is not in posix - it appears to be kind of common, tho.

I would not be opposed to a base64 `string escape` style.

mqudsi: fwiw, `base64` isn't installed by default on FreeBSD.

For my own scripts, I have found `openssl enc base64` to be more universally available than `base64`. I can't remember if the `openssl` command has been pulled from the base installation in new macOS versions, but it was part of the base installation at least as of 10.10.

PR #9233, opened 2022-09-24
completion/adb: add more general options

NextAlone:
only complete device serial when space after '-s' option

## Description

Talk about your changes here.

Fixes issue #

## TODOs:

### DIFF
--- a/share/completions/adb.fish
+++ b/share/completions/adb.fish
@@ -14,7 +14,10 @@ function __fish_adb_get_devices -d 'Run adb devices and parse output'
     set -l procs (ps -Ao comm= | string match 'adb')
     # Don't run adb devices unless the server is already started - it takes a while to init
     if set -q procs[1]
-        adb devices -l | string replace -rf '(\S+).*model:(\S+).*' '$1'\t'$2'
+        set -l token (commandline -opc)
+        if test $token[-1] = '-s'
+            adb devices -l | string replace -rf '(\S+).*model:(\S+).*' '$1'\t'$2'
+        end
     end
 end
 
@@ -93,9 +96,14 @@ end
 
 
 # Generic options, must come before command
-complete -n __fish_adb_no_subcommand -c adb -s s -x -a "(__fish_adb_get_devices)" -d 'Device to communicate with'
-complete -n __fish_adb_no_subcommand -c adb -s d -d 'Communicate with first USB device'
-complete -n __fish_adb_no_subcommand -c adb -s e -d 'Communicate with emulator'
+complete -n __fish_adb_no_subcommand -c adb -s a -d 'Listen on all network interfaces'
+complete -n __fish_adb_no_subcommand -c adb -s d -d 'Use first USB device'
+complete -n __fish_adb_no_subcommand -c adb -s e -d 'Use first TCP/IP device'
+complete -n __fish_adb_no_subcommand -c adb -s s -x -a "(__fish_adb_get_devices)" -d 'Use device with given serial'
+complete -n __fish_adb_no_subcommand -c adb -s t -d 'Use device with given transport id'
+complete -n __fish_adb_no_subcommand -c adb -s H -d 'Name of adb server host'
+complete -n __fish_adb_no_subcommand -c adb -s P -d 'Port of adb server'
+complete -n __fish_adb_no_subcommand -c adb -s L -d 'Listen on given socket for adb server'
 
 # Commands
 complete -f -n __fish_adb_no_subcommand -c adb -a connect -d 'Connect to device'
--- a/share/completions/adb.fish
+++ b/share/completions/adb.fish
@@ -171,7 +171,7 @@ complete -n '__fish_seen_subcommand_from backup' -c adb -o nosystem -d 'Exclude 
 complete -n '__fish_seen_subcommand_from backup' -c adb -f -a "(__fish_adb_list_packages)" -d 'Package(s) to backup'
 
 # reboot
-complete -n '__fish_seen_subcommand_from reboot' -c adb -x -a 'bootloader recovery'
+complete -n '__fish_seen_subcommand_from reboot' -c adb -x -a 'bootloader recovery fastboot'
 
 # forward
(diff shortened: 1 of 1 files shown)


share/completions/adb.fish
```diff
@@ -14,7 +14,10 @@ function __fish_adb_get_devices -d 'Run adb devices and parse output'
     set -l procs (ps -Ao comm= | string match 'adb')
     # Don't run adb devices unless the server is already started - it takes a while to init
     if set -q procs[1]
-        adb devices -l | string replace -rf '(\S+).*model:(\S+).*' '$1'\t'$2'
+        set -l token (commandline -opc)
+        if test $token[-1] = '-s'
+            adb devices -l | string replace -rf '(\S+).*model:(\S+).*' '$1'\t'$2'
+        end
```
mqudsi, line 20: `__fish_adb_get_devices` is executed from two different spots, one to complete `-s` and the other to complete `adb disconnect`. 

For the former (`adb ... -s TAB`), there's no need to add this check because it's already guaranteed (since it's the completion for a `-s` shortcode) and your test will actually fail in some cases (for example, if two shortcodes are combined into one, which getopts allows, such as `-as TAB` since `-a` doesn't require an argument but `-s` does, and they're both shortcodes).

For the latter (`adb disconnect TAB`), I'm not sure how this fits in. In particular, I'm not sure how `adb disconnect` looks with an adb device as the argument. Is it `adb disconnect [one of __fish_adb_get_devices]` or is it `adb disconnect -s [one of __fish_adb_get_devices]`? If it's the first case, this test breaks that. If it's the second, the correct fix would be to just add `-s s` to the `adb disconnect` completion below instead of this (again, in order to use fish's better and more correct argument parsing instead of relying on your test).

NextAlone: For the former, `adb -s<TAB>` will complete as `adb -s1234567` which is not right.

mqudsi: That would imply the problem is with the following line, right?

```fish
complete -n __fish_adb_no_subcommand -c adb -s s -x -a "(__fish_adb_get_devices)" -d 'Use device with given serial'
```

Again, I'm not entirely sure what the valid usages of `-s` would be, if you could illustrate that would help (specifically in case of `adb disconnect` as well).

mqudsi (changes requested):

share/completions/adb.fish
NextAlone: Yes, because option's parameter completes when option is given. 

image
image

And adb disconnect doesn't use `-s` option, it uses parameter after it, but i do not find the specific manpage.

NextAlone: So, just guarantee that there is a space before the serial id.

mqudsi: Ok, that means that the `complete ... -s s ...` rule needs to be `complete ... -o s ...` and if `adb` doesn't support it for the `-s` option, it probably doesn't support it for any of the options. Using `-o` instead of `-s` prevents grouping options together and prevents munging an option (like `-s`) with its value, instead requiring a space between them.

[This SO answer](https://stackoverflow.com/a/46051196/17027) confirms what you're saying about `adb disconnect` only being for TCP connections, so the old line `adb.fish:153` (which says to use `__fish_adb_get_devices` when `-s` is used in conjunction with subcommand `disconnect`) should be deleted. (cc @tomsci who was kind enough [to contribute the adb completions back in 2013](https://github.com/fish-shell/fish-shell/commit/0e7acaa3c4676dca7204deae85c08c2b2e426bb9), in case he had a specific use case in mind).

tomsci: > in case he had a specific use case in mind

If I did, I've long since forgotten :D

mqudsi: Hi @tomsci, nice of you to drop by! 😀

NextAlone:
So that is to say, `adb disconnect` also needs to modify the parameters to filter TCP/IP devices.

share/completions/adb.fish
```diff
@@ -182,8 +174,8 @@ complete -n '__fish_seen_subcommand_from reconnect' -c adb -x -a device -d 'Kick
 
 # commands that accept listing device files
 complete -n '__fish_seen_subcommand_from shell' -c adb -f -a "(__fish_adb_list_files)" -d 'File on device'
-complete -n '__fish_seen_subcommand_from pull' -c adb -f -a "(__fish_adb_completion_for_pull)" -d 'File on device'
-complete -n '__fish_seen_subcommand_from push' -c adb -f -a "(__fish_adb_completion_for_push)" -d 'File on device'
+complete -n '__fish_seen_subcommand_from pull' -c adb -F -a "(__fish_adb_list_files)" -d 'File on device'
+complete -n '__fish_seen_subcommand_from push' -c adb -F -a "(__fish_adb_list_files)" -d 'File on device'
 
 # logcat
 complete -n '__fish_seen_subcommand_from logcat' -c adb -f
```
mqudsi, line 181: Do you know what this line is completing?

NextAlone: Yes, completes host files and device files at the same time.  Found it hard to detect when to complete second parameter because first allows multi.

`adb push [local]... [remote]`
`adb pull [remote]... [local]`

mqudsi: Sorry, GitHub highlighted all 5 lines but I only meant the last one (after `# logcat`).

NextAlone: It means not to complete file for `adb logcat`.

mqudsi: Ok, thanks. This all looks good, thanks for explaining the adb logic to me and for bearing through the requested changes and refactors. If you’re happy with the completions for now, I can squash and merge.

mqudsi:
Ok, thanks. This all looks good, thanks for explaining the adb logic to me and for bearing through the requested changes and refactors. If you’re happy with the completions for now, I can squash and merge.

NextAlone:
> Ok, thanks. This all looks good, thanks for explaining the adb logic to me and for bearing through the requested changes and refactors. If you’re happy with the completions for now, I can squash and merge.

Ok, thank you too.

PR #9237, opened 2022-09-25
add usbip completion

NextAlone:
## Description

Talk about your changes here.

Fixes issue #

## TODOs:

### DIFF
--- a/share/completions/usbip.fish
+++ b/share/completions/usbip.fish
@@ -1,0 +1,69 @@
+function __fish_usbip_no_subcommand -d 'Test if usbip has yet to be given the subcommand'
+    for i in (commandline -opc)
+        if contains -- $i attach detach list bind unbind port
+            return 1
+        end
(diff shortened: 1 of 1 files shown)


share/completions/usbip.fish
```diff
@@ -0,0 +1,69 @@
+function __fish_usbip_no_subcommand -d 'Test if usbip has yet to be given the subcommand'
```
mqudsi, line 1: Instead of hard-coding a list of subcommands (that _might_ change over time, although for some utilities they're pretty much set in stone) I think you can just use `__fish_is_nth_token 1` as the condition, which tests that the current token being completed is the first non `-` or `--` argument to the command, e.g. `usbip TAB` would pass, `usbip --verbose TAB` would pass, but `usbip attach TAB` wouldn't.

mqudsi:
Merged with thanks.

PR #9240, opened 2022-09-25
Make maybe_t safer against accidental misuse

mqudsi:
We've had a few bugs over the years (at least one I caused and just patched, others I caught so no pointing fingers as there's blame to go around!) stemming from accidental misuse of maybe_t with bool-convertible types. This patch disables maybe_t's bool operator if the type T is already bool convertible, forcing the (barely worth mentioning) need to use maybe_t::has_value() instead.

This patch both removes maybe_t's bool conversion for bool-convertible types and updates the existing codebase to use the explicit `has_value()` method in place of existing implicit bool conversions.

This makes it impossible to forget that a function returns a `maybe_t<int>` instead of `int` or `bool` and then test `if (foo())` because `maybe_t<bool/int>` can no longer be used as a unary expression (though `maybe_t<wcstring>` can).

A somewhat different class of maybe_t misuse we've had before is an issue where a dev writes `return 0` in a method that returns a `maybe_t<int>` then checks the return value of the function with `if (foo())` - this patch makes it so you need to write `if (foo.has_value())` but there's still going to be an issue if you expect `maybe_t<int>{0}` to evaluate to false, but hopefully the added friction and explicitly needing to write `if (foo().has_value())` will give one an opportunity to catch their mistake, at least making sure they remember at the call site that the result is a `maybe_t` and not an int, so possibly both `foo.has_value()` and `foo.value()` have to be tested. In other words, this isn't a panacea, but maybe it'll help!

@ridiculousfish I'm requesting a review mainly to see if you feel that the benefits outweigh the costs or if this constitutes excessive hand-holding. If you think it's not worth it, I'm happy to just close this PR. It was not a difficult undertaking (just tedious!) so no hard feelings.

----

If you look at the changes in this patch series, you'll notice that sometimes `maybe_t` return types are assigned "bool-like names" like `found`, which I think is an issue that makes `maybe_t` return values even more likely to be misused. Is `maybe_t<bool>{false}` "found"? These should probably be renamed!

### DIFF
--- a/src/builtin.cpp
+++ b/src/builtin.cpp
@@ -460,7 +460,7 @@ proc_status_t builtin_run(parser_t &parser, const wcstring_list_t &argv, io_stre
         // Resolve our status code.
         // If the builtin itself produced an error, use that error.
         // Otherwise use any errors from writing to out and writing to err, in that order.
-        int code = builtin_ret ? *builtin_ret : 0;
+        int code = builtin_ret.has_value() ? *builtin_ret : 0;
         if (code == 0) code = out_ret;
         if (code == 0) code = err_ret;
 
--- a/src/builtins/jobs.cpp
+++ b/src/builtins/jobs.cpp
@@ -42,8 +42,11 @@ static double cpu_use(const job_t *j) {
 /// Print information about the specified job.
 static void builtin_jobs_print(const job_t *j, int mode, int header, io_streams_t &streams) {
     int pgid = INVALID_PID;
-    if (auto job_pgid = j->get_pgid()) {
-        pgid = *job_pgid;
+    {
+        auto job_pgid = j->get_pgid();
+        if (job_pgid.has_value()) {
+            pgid = *job_pgid;
+        }
     }
 
     wcstring out;
--- a/src/maybe.h
+++ b/src/maybe.h
@@ -134,8 +134,14 @@ class maybe_t : private maybe_detail::conditionally_copyable_t<T> {
     // return whether the receiver contains a value.
     bool has_value() const { return impl_.filled; }
 
-    // bool conversion indicates whether the receiver contains a value.
-    explicit operator bool() const { return impl_.filled; }
+    // A bool operator as a shortcut to test if the maybe_t has a value.
+    // Not enabled if the type T is already bool-convertible to prevent accidental misuse,
+    // otherwise the "typename std::enable_if<....>::type" evaluates to bool, giving us a definition
+    // of `explicit operator bool() const { ... }`
+    template <typename U = T> explicit operator
+    typename std::enable_if<!std::is_convertible<U, bool>::value, bool>::type() const {
+        return impl_.filled;
+    }
 
     // The default constructor constructs a maybe with no value.
     maybe_t() = default;
@@ -210,5 +216,19 @@ class maybe_t : private maybe_detail::conditionally_copyable_t<T> {
 
     ~maybe_t() { reset(); }
 };
+
+// A specialization enabling bool conversion for maybe_t types *can't* be converted to boolean,
+// protecting against accidental misuse.
+//
+//
+// template <typename U>
+// class maybe_t<U, typename std::enable_if<std::is_convertible<U, bool>::value>::type> {
+//         // bool conversion indicates whether the receiver contains a value.
+//         // bool conversion is disabled when the type is maybe_t<bool> or maybe_t<int> to prevent
+//         // accidentally comparing `maybe_t<>.has_value()` instead of `*maybe_t<>`.
+//         template <typename U = T>
+//         typename <std::enable_if<std::is_integral<Integer>::value, bool>::type
+//         explicit operator bool() const { return impl_.filled; }
+// }
 
 #endif
--- a/src/builtins/disown.cpp
+++ b/src/builtins/disown.cpp
@@ -28,7 +28,7 @@ static void disown_job(const wchar_t *cmd, io_streams_t &streams, job_t *j) {
     // Stopped disowned jobs must be manually signaled; explain how to do so.
     auto pgid = j->get_pgid();
     if (j->is_stopped()) {
-        if (pgid) killpg(*pgid, SIGCONT);
+        if (pgid.has_value()) killpg(*pgid, SIGCONT);
         const wchar_t *fmt =
             _(L"%ls: job %d ('%ls') was stopped and has been signalled to continue.\n");
         streams.err.append_format(fmt, cmd, j->job_id(), j->command_wcstr());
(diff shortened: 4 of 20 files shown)


src/builtin.cpp
```diff
@@ -460,7 +460,7 @@ proc_status_t builtin_run(parser_t &parser, const wcstring_list_t &argv, io_stre
         // Resolve our status code.
         // If the builtin itself produced an error, use that error.
         // Otherwise use any errors from writing to out and writing to err, in that order.
-        int code = builtin_ret ? *builtin_ret : 0;
+        int code = builtin_ret.has_value() ? *builtin_ret : 0;
```
krobelus, line 463: interesting. I considered the opposite - removing all uses of has_value()
since i like the contextual conversion to bool - it only happens in places where a bool is expected.
Don't mind me too much though.

src/maybe.h
```diff
@@ -134,8 +134,14 @@ class maybe_t : private maybe_detail::conditionally_copyable_t<T> {
...
 
-    // bool conversion indicates whether the receiver contains a value.
-    explicit operator bool() const { return impl_.filled; }
+    // A bool operator as a shortcut to test if the maybe_t has a value.
+    // Not enabled if the type T is already bool-convertible to prevent accidental misuse,
+    // otherwise the "typename std::enable_if<....>::type" evaluates to bool, giving us a definition
+    // of `explicit operator bool() const { ... }`
+    template <typename U = T> explicit operator
+    typename std::enable_if<!std::is_convertible<U, bool>::value, bool>::type() const {
+        return impl_.filled;
```
krobelus, line 143: Technically we only had problems with `maybe_t<bool>`, not with `maybe_t<size_t>` etc.
I guess `maybe_t<SomeType*>` is dangerous too but I think that happens to us only for typedef'd pointers (though I didn't check).

src/maybe.h
```diff
@@ -211,4 +217,18 @@ class maybe_t : private maybe_detail::conditionally_copyable_t<T> {
     ~maybe_t() { reset(); }
 };
 
+// A specialization enabling bool conversion for maybe_t types *can't* be converted to boolean,
+// protecting against accidental misuse.
+//
+//
+// template <typename U>
+// class maybe_t<U, typename std::enable_if<std::is_convertible<U, bool>::value>::type> {
```
krobelus, line 225: leftover?

src/maybe.h
mqudsi: We've actually had problems with `maybe_t<int>` (or maybe it was `size_t`), where an attempt to test for a non-zero offset was actually testing `maybe_t::has_value()`.

src/maybe.h
mqudsi: Yes; I missed that! Thanks.

src/builtin.cpp
mqudsi: I know what you mean! Once upon a time, I too loved implicit bool operators (especially when `shared_ptr<>` was introduced) as they removed the "boilerplate" of having to explicitly do `ptr != nullptr` in order to please the compiler (I think MSVC complained if you did the equivalent of `if (((void*) ptr)) { ... }`), but these days I value my sanity and code maintenance far more. 

Even now, just looking at the PR changes tab, it leaps at me immediately for a number of newly-converted tests: "should we also be testing the value?" whereas I would have just glossed over those if conditions and continued on my merry way.

src/builtins/jobs.cpp
```diff
@@ -42,8 +42,11 @@ static double cpu_use(const job_t *j) {
 /// Print information about the specified job.
 static void builtin_jobs_print(const job_t *j, int mode, int header, io_streams_t &streams) {
     int pgid = INVALID_PID;
-    if (auto job_pgid = j->get_pgid()) {
-        pgid = *job_pgid;
+    {
+        auto job_pgid = j->get_pgid();
+        if (job_pgid.has_value()) {
```
krobelus, line 47: We could probably upgrade to C++17 soon and use `if` init-statements

```cpp
if (auto job_pgid = j->get_pgid(); job_pgid.has_value()) {
```

krobelus:
Feels like a good change overall, LGTM! It does introduce a difference to `std::optional` but it's not so bad since we are merely restricting ourselves to a subset of that type's API.

faho:
Do we want the bool conversion *at all*?

0217f9c3952e5f2c4ac9ce86693df6e9bd6ae12a is maybe a bit of an edge-case, but it already shows weird platform differences where we can now only use bool conversion on some platforms.

I don't think it would be terrible to just always require `.has_value()`?

src/builtins/jobs.cpp
faho: The cutoff point on C++17 support seems to be gcc 7 (https://gcc.gnu.org/projects/cxx-status.html#cxx17 - with one teensy exception in gcc 8).

According to https://repology.org/project/gcc/versions, that's CentOS/RHEL 8, Version 7 ships with gcc 4.8.5 and is supported until 2024.

So we would be asking people there to upgrade their compiler - which includes us, since we build for CentOS on OBS.

mqudsi:
@faho I didn't want to rock the boat _too_ much and it's arguably safe to assume no one will accidentally test the equivalent of `if (maybe_t<job_t>) { .. }` expecting anything other than nullability or whatever. But yes, 0217f9c definitely gave me pause and I'm personally open to just dropping implicit bool conversion, but I'll defer to others on this.

src/builtins/jobs.cpp
mqudsi: I'm leery of upgrading to a newer spec because I was around for the introduction of C++11 and saw how long it took for compilers to actually get their act together and actually support what they claimed to. I'm sure there are quality of life benefits in C++14 and C++17 but I'm honestly continuing to use the sane subset of C++11 we already do. C++11 vs C++98 was huge and a fundamentally seismic shift in making the standard library usable to write cross-platform software, but I don't think we can say the same about the newer variants (for now). Also, it's a stupid reason but I love that I can dig out old VMs running ancient OSes and get fish to compile and I would genuinely miss being able to do that.

(Though I have to say C++20 concepts would have made disabling bool conversion for bool-convertible types here so much easier than the SIFNAE hell I ended up writing. Apparently you can just write your member function and after the prototype add `requires (!std::is_convertible<T>)` which is nice, but not really world changing and not worth breaking compatibility over.)

zanchey: I built new compilers for old images on OBS and I did not enjoy it!

ridiculousfish (approved):
Please squash into a single commit so the tests pass on each commit, and please remove the commented-out code in maybe.h. Otherwise LGTM, I like this direction.

mqudsi:
Thanks, @ridiculousfish. That was the plan, but I wanted your general approval before sinking any additional work into this.

PR #9078, opened 2022-07-19
Add autocomplete for `tox`

rouge8:
## Description

Add autocomplete for tox. Based on `tox --help` output for tox 3.25.1.

## TODOs:

- [ ] Changes to fish usage are reflected in user documentation/manpages.
- [ ] Tests have been added for regressions fixed
- [ ] User-visible changes noted in CHANGELOG.rst

PR #9241, opened 2022-09-26
Fix the completions of tox

yoshida-ryuhei:
## Description

This is related to #9078 .
Unfortunately, that patch has a typo and does not work as intended.

The number of `_` should be 2, but the 3 in the definition. Therefore the patch does not work.

Thanks.

## TODOs:


faho:
Thanks!

PR #9242, opened 2022-09-26
`complete` docs: Document when to use old-style for short options

mqudsi:
There are many applications with "primitive" argument parsing capabilities that cannot handle munging two short options together (`-xf` for `-x -f`) or a short option and its required value (`-dall` for `-d all`). To prevent fish from suggesting munged arguments/payloads, the options (both long and short, not just long!) can be specified as `-o` or `--old-option` but none of this is documented.

*I'm not confident enough to say my docs don't contain any incorrect info, so a review would be appreciated!*

### DIFF
--- a/doc_src/cmds/complete.rst
+++ b/doc_src/cmds/complete.rst
@@ -36,8 +36,8 @@ The following options are available:
 **-l** or **--long-option** *LONG_OPTION*
     Adds a GNU-style long option to the completions list.
 
-**-o** or **--old-option** *LONG_OPTION*
-    Adds an old-style long option to the completions list (see below for details).
+**-o** or **--old-option** *OPTION*
+    Adds an old-style short or long options (see below for details).
 
 **-a** or **--arguments** *ARGUMENTS*
     Adds the specified option arguments to the completions list.
@@ -76,7 +76,7 @@ Command specific tab-completions in ``fish`` are based on the notion of options 
 
 - Short options, like ``-a``. Short options are a single character long, are preceded by a single hyphen and can be grouped together (like ``-la``, which is equivalent to ``-l -a``). Option arguments may be specified by appending the option with the value (``-w32``), or, if ``--require-parameter`` is given, in the following parameter (``-w 32``).
 
-- Old-style long options, like ``-Wall`` or ``-name``. Old-style long options can be more than one character long, are preceded by a single hyphen and may not be grouped together. Option arguments are specified in the following parameter (``-ao null``) or after a ``=`` (``-ao=null``).
+- Old-style options, long like ``-Wall`` or ``-name`` or even short like ``-a``. Old-style options can be more than one character long, are preceded by a single hyphen and may not be grouped together. Option arguments are specified by default following a space (``-ao null``) or after a ``=`` (``-ao=null``). Short options for commands that do not support concatenating multiple short options in one string (not supporting ``-xf`` as short for ``-x -f``) or a short option and its value in one string (not supporting ``-d9`` instead of ``-d 9``) should be specified as single-character old-style options; for example, ``complete -c foo -o s; complete -c foo -o v`` would never suggest ``foo -ov`` but rather ``foo -o -v``.
 
 - GNU-style long options, like ``--colors``. GNU-style long options can be more than one character long, are preceded by two hyphens, and can't be grouped together. Option arguments may be specified after a ``=`` (``--quoting-style=shell``), or, if ``--require-parameter`` is given, in the following parameter (``--quoting-style shell``).
 
--- a/doc_src/cmds/complete.rst
+++ b/doc_src/cmds/complete.rst
@@ -37,7 +37,7 @@ The following options are available:
     Adds a GNU-style long option to the completions list.
 
 **-o** or **--old-option** *OPTION*
-    Adds an old-style short or long options (see below for details).
+    Adds an old-style short or long option (see below for details).
 
 **-a** or **--arguments** *ARGUMENTS*
     Adds the specified option arguments to the completions list.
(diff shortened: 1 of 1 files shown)


faho (changes requested):

doc_src/cmds/complete.rst
```diff
@@ -76,7 +76,7 @@ Command specific tab-completions in ``fish`` are based on the notion of options
...
 
-- Old-style long options, like ``-Wall`` or ``-name``. Old-style long options can be more than one character long, are preceded by a single hyphen and may not be grouped together. Option arguments are specified in the following parameter (``-ao null``) or after a ``=`` (``-ao=null``).
+- Old-style options, long like ``-Wall`` or ``-name`` or even short like ``-a``. Old-style options can be more than one character long, are preceded by a single hyphen and may not be grouped together. Option arguments are specified by default following a space (``-ao null``) or after a ``=`` (``-ao=null``). Short options for commands that do not support concatenating multiple short options in one string (not supporting ``-xf`` as short for ``-x -f``) or a short option and its value in one string (not supporting ``-d9`` instead of ``-d 9``) should be specified as single-character old-style options; for example, ``complete -c foo -o s; complete -c foo -o v`` would never suggest ``foo -ov`` but rather ``foo -o -v``.
```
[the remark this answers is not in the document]
mqudsi: That's fine with me, but I would like to keep the omissions of "long" from the old-style documentation to make it more inclusive/so it isn't dismissed out-of-hand by someone thinking it can't apply to them in their situation.

fwiw, I don't think the existing completions are necessarily a good metric to go by. As shown in #9233, they may be wrong but just haven't annoyed anyone enough to complain. Option grouping is an old unix greybeard mainstay but there's no _requirement_ for it to be the case and anyone manually parsing arguments in a new application today instead of using getopts or a similar library is probably not handling them. As our completions library has expanded, we're no longer documenting just the most "elite" traditional cli tools, but also a lot of less traditional ones too. 

* Two out of the eleven surveyed C++ arg parsing libraries [in this 2018 roundup](https://attractivechaos.wordpress.com/2018/08/31/a-survey-of-argument-parsing-libraries-in-c-c/) didn't support option grouping
* Any cross-platform tools with Windows origins (like any .NET Core application) isn't likely to support option grouping because it's not a thing on Windows. 
* Commands not written in C/C++ and without a getopt legacy (such as any java applications) don't usually support option grouping either. 
* golang's most common arg parser [doesn't support grouping short options](https://nullprogram.com/blog/2020/08/01/#:~:text=Go%E2%80%99s%20flag%20package,with%20%3D.).
* Python's argparse [apparently broke short option grouping](https://utcc.utoronto.ca/~cks/space/blog/python/ArgparseSomeUnixNotes#:~:text=Changed%20in%20version%203.8%3A%20In%20previous%20versions%2C%20allow_abbrev%20%5Bbeing%20False%5D%20also%20disabled%20grouping%20of%20short%20flags%20such%20as%20%2Dvv%20to%20mean%20%2Dv%20%2Dv.) until Python 3.8.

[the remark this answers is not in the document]
faho: How about this wording:

> Old-style options can't be grouped together even if they are only a single character long, unlike short options.

And then, in the examples, show the example you gave here:

```fish
complete -c foo -o o
complete -c foo -o v
# This will allow "foo -o -v", but not "foo -ov" because old-style options can't be grouped
```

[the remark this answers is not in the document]
faho: >Two out of the eleven surveyed C++ arg parsing libraries [in this 2018 roundup](https://attractivechaos.wordpress.com/2018/08/31/a-survey-of-argument-parsing-libraries-in-c-c/) didn't support option grouping

(Both of those have since been fixed: https://tclap.sourceforge.net/v1.2/manual.html#COMBINE_SWITCHES and https://github.com/adishavit/argh/issues/28)

Yes, weird option parsers exist and we should, within reason, allow them to be supported with completions. I'm just saying that we don't need to front-load that complexity in the docs.

mqudsi:
I've updated the branch per the discussion above. As a later to-do, I don't love how the examples all blend together without a separator but I think the short old-style example works fine there (and doesn't need a separate section).

doc_src/cmds/complete.rst
```diff
@@ -36,8 +36,8 @@ The following options are available:
 **-l** or **--long-option** *LONG_OPTION*
     Adds a GNU-style long option to the completions list.
 
-**-o** or **--old-option** *LONG_OPTION*
-    Adds an old-style long option to the completions list (see below for details).
+**-o** or **--old-option** *OPTION*
+    Adds an old-style short or long options (see below for details).
```
faho, line 40: Broken plural - "Adds an options"

faho: How about:

```
Adds an old-style single-dash option (see below for details
```

?

doc_src/cmds/complete.rst
```diff
@@ -76,7 +76,7 @@ Command specific tab-completions in ``fish`` are based on the notion of options
 
 - Short options, like ``-a``. Short options are a single character long, are preceded by a single hyphen and can be grouped together (like ``-la``, which is equivalent to ``-l -a``). Option arguments may be specified by appending the option with the value (``-w32``), or, if ``--require-parameter`` is given, in the following parameter (``-w 32``).
 
-- Old-style long options, like ``-Wall`` or ``-name``. Old-style long options can be more than one character long, are preceded by a single hyphen and may not be grouped together. Option arguments are specified in the following parameter (``-ao null``) or after a ``=`` (``-ao=null``).
+- Old-style options, long like ``-Wall`` or ``-name`` or even short like ``-a``. Old-style options can be more than one character long, are preceded by a single hyphen and may not be grouped together. Option arguments are specified by default following a space (``-foo null``) or after ``=`` (``-foo=null``).
```
faho, line 79: I was a bit thrown off by the "by default", but it turns out old-style options really don't need "--require-parameter" to accept an option in the next parameter?

That might actually be worth a change!

(not that "require-parameter" is the right way around - it should have been "--optional-parameter" for the few times you have an optional parameter)

doc_src/cmds/complete.rst
mqudsi: Thanks for catching that. Given that the other examples (short, GNU-style long) specify short and long, I figured it makes sense to specify that here, too.

PR #9245, opened 2022-09-29
Decode `\X` escapes immediately

faho:
We forgot to decode (i.e. turn into nice wchar_t codepoints) "byte_literal" escape sequences. This meant that e.g.

```fish
string match ö \Xc3\Xb6

math 5 \X2b 5
```

didn't work, and the latter would print the wonderful error:

```
math: Error: Missing operator
'5 + 5'
   ^
```

In contrast, `math 5 \x2b 5` works because `\x` is currently not "byte_literal" - it's seen as a codepoint number.

So, instead, we decode eagerly.

## TODOs:

----

Sidenote: The diff is an absolute mess - what I did was:

1. Pack the switch/case into a `while` loop so we can `continue`
2. Add a std::string buffer for multiple `\X` escaped values
3. Move an early `if (errored) return none()` into the loop
4. Make the path for "byte_literal" (currently that's only `\X`) put the value into the buffer and `continue` if there's another `\X` following.

A reasonably simple change, but because of the additional indentation this touched most of `read_unquoted_escape`. Sorry about that!

### DIFF
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -1149,161 +1149,185 @@ maybe_t<size_t> read_unquoted_escape(const wchar_t *input, wcstring *result, boo
     bool errored = false;
     size_t in_pos = 1;  // in_pos always tracks the next character to read (and therefore the number
                         // of characters read so far)
-    const wchar_t c = input[in_pos++];
-    switch (c) {
-        // A null character after a backslash is an error.
-        case L'\0': {
-            // Adjust in_pos to only include the backslash.
-            assert(in_pos > 0);
-            in_pos--;
 
-            // It's an error, unless we're allowing incomplete escapes.
-            if (!allow_incomplete) errored = true;
-            break;
-        }
-        // Numeric escape sequences. No prefix means octal escape, otherwise hexadecimal.
-        case L'0':
-        case L'1':
-        case L'2':
-        case L'3':
-        case L'4':
-        case L'5':
-        case L'6':
-        case L'7':
-        case L'u':
-        case L'U':
-        case L'x':
-        case L'X': {
-            long long res = 0;
-            size_t chars = 2;
-            int base = 16;
-            bool byte_literal = false;
-            wchar_t max_val = ASCII_MAX;
+    // For multibyte \X sequences.
+    std::string byte_buff;
+    while (!errored) {
+        const wchar_t c = input[in_pos++];
+        switch (c) {
+                // A null character after a backslash is an error.
+            case L'\0': {
+                // Adjust in_pos to only include the backslash.
+                assert(in_pos > 0);
+                in_pos--;
 
-            switch (c) {
-                case L'u': {
-                    chars = 4;
-                    max_val = UCS2_MAX;
-                    break;
-                }
-                case L'U': {
-                    chars = 8;
-                    max_val = WCHAR_MAX;
-
-                    // Don't exceed the largest Unicode code point - see #1107.
-                    if (0x10FFFF < max_val) max_val = static_cast<wchar_t>(0x10FFFF);
-                    break;
-                }
-                case L'x': {
-                    chars = 2;
-                    max_val = ASCII_MAX;
-                    break;
-                }
-                case L'X': {
-                    byte_literal = true;
-                    max_val = BYTE_MAX;
-                    break;
-                }
-                default: {
-                    base = 8;
-                    chars = 3;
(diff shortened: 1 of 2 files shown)


src/common.cpp
```diff
@@ -1149,161 +1149,185 @@ maybe_t<size_t> read_unquoted_escape(const wchar_t *input, wcstring *result, boo
...
-                if (d < 0) {
-                    // If we have no digit, this is a tokenizer error.
-                    if (i == 0) errored = true;
-                    break;
+                if (!errored && res <= max_val) {
+                    if (byte_literal) {
+                        // Multibyte encodings necessitate that we keep adjacent byte escapes.
+                        // - `\Xc3\Xb6` is "ö", but only together.
+                        // (this assumes a valid codepoint can't consist of multiple bytes
+                        // that are valid on their own, which is true for UTF-8)
```
mqudsi, line 1239: It's true for most multi-byte encodings but the opposite isn't (e.g. instead of `\Xc3\Xb6` this could just as well be `\X51\X51` each of which stands by itself as a character). This'll cause you to keep trying to concatenate adjacent single-byte (or multiple multi-byte) escapes to `byte_buff` at no point stopping to return them upstream... which as far as I can tell, is OK in the general case.

It does mean if an error is encountered parsing a subsequent adjacent escape after consuming one or more bytes that fully form one (or more) character(s), you'll error out from the first escape instead of just the actually broken one, with none of the escaped characters making into the valid payload buffer. I don't know if that matters.

Actually, after testing, it seems that's already the case and pedantically we already report an incorrect error. `fish -c 'echo \\X51\\X'` returns `fish: Invalid token '\X51\X' ...` instead of `fish: Invalid token '\X'`.

faho: >even if it's suboptimal because you might exceed the SSO for byte_buff

If we have been given a large string, we need to handle a large string.

Technically we could force a str2wcstring and clear the buffer after 4 bytes, which should be the limit on any sensible encoding (even UTF-32, the simplest unicode encoding), but I didn't really want to give this function that knowledge, and it seems like an optimization for a special case that very very likely is never going to happen.

If you want to test, see how a file with thousands of `\Xab` in a row fares. My expectation is that it is slower than just using the characters immediately, but not terrible.

faho: I guess we could use MB_CUR_MAX.

faho: Okay, in any realistic case this simply does *not matter*.

My test case is a file created with

```fish
echo echo (string repeat -n 2000000 "\\X82") >echo.fish
```

This will look like

```fish
echo \X82\X82\X82\X82...
```

The timing here is ~500ms without the decoding (i.e. current master), ~720ms with this PR and ~740ms if I add a `byte_buff.size() <= MB_CUR_MAX`.

This sounds like a lot, but it's also a *completely ridiculous file*, and current master is faster *but also wrong*, and flushing the decode early is slower. This isn't something where you can find a short and realistic script that hits a weird edge case and is super slow, this is something where you need to explicitly construct something ridiculous and it's still only slower by a constant factor - it's ~40% *if all you do is reading `\X`*.

So I do not believe this is something where the performance is different in any important way, and we should simply prefer correctness and simplicity.

mqudsi: Yeah, I wasn't suggesting we should do anything about it once I observed that there were no functional regressions. I think it's fine the way it is unless there is some edge case I am missing that _doesn't_ involve erroring out where `read_unquoted_escape()` needs to return exactly one codepoint and no more. Not having to do anything that requires introspection of the current codepage is definitely a win.

faho: >read_unquoted_escape() needs to return exactly one codepoint and no more

Not that it matters, but: It can return multiple, and it will, here - or we would have to know codepoint boundaries, which requires knowing the encoding.

mqudsi: That's what I was trying to say, but I guess didn't do a good job of it :)

krobelus:
the diff reads fine when ignoring whitespace

mqudsi:
Just to share something that's helped me: I manually use [difftastic](https://github.com/Wilfred/difftastic) from time-to-time for looking at patches like this, as it is AST-aware (or at least scope-aware) and can ignore superfluous changes like addition/removal of braces (when it doesn't change meaning), an if condition causing the entire resulting scope to be marked as changed, changes to hard-wrapped lines, etc.

PR #9248, opened 2022-09-29
Completions: sort git branches by recency

mqudsi:
I have about fifty git branches for fish and I almost always `git checkout` between the most recent two or three. Maybe I'm not the only one.

### DIFF
--- a/share/completions/git.fish
+++ b/share/completions/git.fish
@@ -51,11 +51,10 @@ function __fish_git_recent_commits
 end
 
 function __fish_git_branches
-    # This is much faster than using `git branch`,
-    # and avoids having to deal with localized "detached HEAD" messages.
-    __fish_git for-each-ref --format='%(refname)' refs/heads/ refs/remotes/ 2>/dev/null \
-        | string replace -r '^refs/heads/(.*)$' '$1\tLocal Branch' \
-        | string replace -r '^refs/remotes/(.*)$' '$1\tRemote Branch'
+    # This is much faster than using `git branch` and avoids dealing with localized "detached HEAD" messages.
+    # We intentionally only sort local branches by recency. See discussion in #9248.
+    __fish_git for-each-ref --format='%(refname:strip=2)%09Local Branch' --sort=-committerdate refs/heads/ 2>/dev/null
+    __fish_git for-each-ref --format='%(refname:strip=2)%09Remote Branch' refs/remotes/ 2>/dev/null
 end
 
 function __fish_git_submodules
@@ -64,8 +63,7 @@ function __fish_git_submodules
 end
 
 function __fish_git_local_branches
-    __fish_git for-each-ref --format='%(refname:strip=2)' refs/heads/ 2>/dev/null \
-        | string replace -rf '.*' '$0\tLocal Branch'
+    __fish_git for-each-ref --format='%(refname:strip=2)%09Local Branch' --sort=-committerdate refs/heads/ 2>/dev/null
 end
 
 function __fish_git_unique_remote_branches


share/completions/git.fish
```diff
@@ -53,7 +53,7 @@ end
 function __fish_git_branches
     # This is much faster than using `git branch`,
     # and avoids having to deal with localized "detached HEAD" messages.
-    __fish_git for-each-ref --format='%(refname)' refs/heads/ refs/remotes/ 2>/dev/null \
+    __fish_git for-each-ref --format='%(refname)' --sort=-committerdate refs/heads/ refs/remotes/ 2>/dev/null \
```
faho, line 56: I wouldn't do this for remote branches - you can easily have a lot of those, and the more branches you have the harder recency sort becomes to read. E.g. my fish-shell clone with 3 remotes has 326 branches, but only 14 local.

So I would remove it here and only keep it in `__fish_git_local_branches` below - that's also what `git checkout` uses.

(ideally we'd only do this for branches touched in the last week or so, but that's probably prohibitively expensive)

share/completions/git.fish
```diff
@@ -64,8 +64,7 @@ function __fish_git_submodules
 end
 
 function __fish_git_local_branches
-    __fish_git for-each-ref --format='%(refname:strip=2)' refs/heads/ 2>/dev/null \
-        | string replace -rf '.*' '$0\tLocal Branch'
+    __fish_git for-each-ref --format='%(refname:strip=2)%09Local Branch' --sort=-committerdate refs/heads/ 2>/dev/null
```
faho, line 67: What's the reason for removing the description here?

faho (changes requested):

share/completions/git.fish
mqudsi: Sounds good.

mqudsi: @faho I tried it but it seems `git merge` and `git cherry-pick` (among others) don't use the sorted `__fish_git_local_branches` impl, though. I would definitely want local branches sorted by recency there (and honestly, probably everywhere else).

We can split `__fish_git_branches` into two invocations of `git for-each-ref`, once for local (sorted by recency) and the other for remotes (with the default sort order). The runtime cost should be the same since it's roughly linear with regards to the number of refs being considered and we're just splitting it into to, though we are paying the `git` startup cost twice in that case:

```diff
diff --git a/share/completions/git.fish b/share/completions/git.fish
index 1dcf9e037..fdbb25092 100644
--- a/share/completions/git.fish
+++ b/share/completions/git.fish
@@ -53,9 +53,8 @@ end
 function __fish_git_branches
     # This is much faster than using `git branch`,
     # and avoids having to deal with localized "detached HEAD" messages.
-    __fish_git for-each-ref --format='%(refname)' refs/heads/ refs/remotes/ 2>/dev/null \
-        | string replace -r '^refs/heads/(.*)$' '$1\tLocal Branch' \
-        | string replace -r '^refs/remotes/(.*)$' '$1\tRemote Branch'
+    __fish_git for-each-ref --format='%(refname:short)%09Local Branch' --sort=-committerdate refs/heads/ 2>/dev/null
+    __fish_git for-each-ref --format='%(refname:short)%09Remote Branch' refs/remotes/ 2>/dev/null
 end
 
 function __fish_git_submodules
```

(We _do_ remove the `string` invocation in this case, if that offsets the cost any.)

The other option is to call `__fish_git_local_branches` in all of those other completions before calling `__fish_git_branches` and rely on stable sort-and-dedup to remove the duplicate listings but preserve the order - but that would be more expensive.

faho: >We can split __fish_git_branches into two invocations of git for-each-ref, once for local (sorted by recency) and the other for remotes (with the default sort order). The runtime cost should be the same

That does not seem to be the case for me.

```fish
time __fish_git for-each-ref --format='%(refname)' refs/heads/ refs/remotes/ 2>/dev/null \
        | string replace -r '^refs/heads/(.*)$' '$1\tLocal Branch' \
        | string replace -r '^refs/remotes/(.*)$' '$1\tRemote Branch'
```

is ~30ms, while

```fish
time begin; __fish_git for-each-ref --format='%(refname:short)%09Local Branch' --sort=-committerdate refs/heads/ 2>/dev/null
__fish_git for-each-ref --format='%(refname:short)%09Remote Branch' refs/remotes/ 2>/dev/null; end
```

takes ~150ms.

Aaaand the culprit is: the `%(refname:short)` format, specifically for the remote branches.

(this is a different machine, here I have a total of 3000 remote branches)

mqudsi: That makes sense, it has to determine what would be the shortest unambiguous name which requires a quadratic runtime. Does using `%(refname:strip=2)` suffer the same problem? (We could always put `string` back in there as well, but I'd be surprised if that was the fastest way.)

mqudsi: Yeah, I think that would be fine. We're already stripping instead of disambiguating in several other places and it should be easy to fix an "ambiguous ref" error if it ever comes up (and not worth paying the price of avoid it, given how much slower that can be).

mqudsi:
I've updated the branch with the changes (and added some suggestions for future changes in the commit message).

Thanks for the help on this one, @faho.

mqudsi:
FYI and for future reference (as if I'll ever find this again, though!) -- for getting the name of the current branch, the method chosen apparently makes a huge difference:

```
  'git symbolic-ref --short HEAD' ran
    1.05 ± 0.07 times faster than 'git branch --show-current'
    1.16 ± 0.06 times faster than 'git rev-parse --abbrev-ref HEAD'
   10.35 ± 0.42 times faster than 'git name-rev --name-only HEAD'
```

krobelus:
fifty branches is a lot, I only have one, where I add all interesting commits

PR #9249, opened 2022-09-29
Exclude current directory ('.') from rsync completions

cstyles:
## Description

This is what currently happens when I try to tab complete on a directory with a single file (`file.txt`) in the middle of an `rsync` command:

```shell
$ rsync localhost:~/stuff/<TAB>
…:~/stuff/./  (Remote path)  …:~/stuff/file.txt  (Remote path)
```

The current directory (`.`) shows up as an option and stops rsync from automatically selecting the only file. I can't think of a reason why one would actually want to tab complete to `.` (and even if one did, just typing `.` is a hell of a lot quicker than waiting for a response that probably traverses the network). I couldn't find any `rsync` option to exclude the current directory so I just used `string match` to remove it. Hopefully nothing actually useful gets matched (I tried, for example, creating `file.` and `directory./` and they still show up) but I would appreciate it if someone could double-check.

## TODOs:

### DIFF
--- a/share/completions/rsync.fish
+++ b/share/completions/rsync.fish
@@ -205,6 +205,7 @@ complete -c rsync -d "Remote path" -n "commandline -ct | string match -q '*:*'" 
 )(
 	# Get the list of remote files from the specified rsync server.
         rsync --list-only (__rsync_remote_target) 2>/dev/null | string replace -r '^d.*' '\$0/' |
-        string replace -r '(\S+\s+){4}' '' $(set -q new_escaping[1]; or echo ' | string escape -n'; echo)
+        string replace -r '(\S+\s+){4}' '' |
+        string match --invert './' $(set -q new_escaping[1]; or echo ' | string escape -n'; echo)
 )
 "


mqudsi (changes requested):
Thanks for filing this PR. I agree that the . completion is suboptimal. 

I would either pipe the old output to `string match —invert ./` or keep what you did but use a `—` before the subshell to prevent arguments being interpreted as switches. 

I prefer the former since it is easier to follow, reading from left to right in the same order as you’ll the actions you are performing (first generating the list then excluding the one result we don’t want).

To clarify string match behavior for you, the match you used will only match a literal ./ instead of any string containing ./ - to get it to exclude any result with ./ anywhere you would use -e (for entire) in addition to `—invert`.  

If the actual string generated is `stuff/./` what you have without -e won’t block it though (I’m on my phone and can’t really check very well). It does block it if ./ is generated and is then filtered and concatenated to the current path.

cstyles:
> If the actual string generated is `stuff/./` what you have without -e won’t block it though (I’m on my phone and can’t really check very well). It does block it if ./ is generated and is then filtered and concatenated to the current path.

Forgive me if I'm misunderstanding but the output of `rsync` doesn't include the prefix:

```shell
$ rsync --list-only localhost:~/stuff/
drwxr-xr-x            160 2022/10/01 13:25:58 .
-rw-r--r--              0 2022/10/01 13:25:52 file.
-rw-r--r--              0 2022/10/01 13:25:55 file.txt
drwxr-xr-x             64 2022/10/01 13:25:58 dir.
```

So if I run the commands from the `rsync` completion file up until the `string match` command that I added, I get this result:

```shell
$ rsync --list-only localhost:~/stuff/ | string replace -r '^d.*' '$0/' | string replace -r '(\S+\s+){4}' ''
./
file.
file.txt
dir./
```

(The first `string replace` just adds a trailing slash to directories and the second trims file permissions, mtime, etc.)

So I think that just matching on `./` is OK (it works on my end at least). I would appreciate it if someone could confirm on their machine just to be sure.

mqudsi:
Sorry, yes, that's what I meant. Not necessarily the direct output of rsync but after the modifications we make to it in the completion pipeline. Note that the output of rsync changes depending on whether it is piped into something or displayed to the console (you can try `rsync ... | cat` to see what I mean). 

Your use of `string match --invert ./` here is correct, just change it to read from the pipe instead of arguments and we should be good to go.

mqudsi:
Merged.

PR #9255, opened 2022-10-03
Add a fish_delta helper function

faho:
## Description

This goes through $fish_function_path/$fish_complete_path and checks which files you have that aren't in the stock fish install. So it helps you to figure out which functions and completions you've overridden.

For instance, on my system:

```
> fish_delta -fd | head -n 5
New file: /home/alfa/.config/fish/test/functions/__fish_supports_cursor.fish
New file: /home/alfa/.config/fish/test/functions/_fish_systemctl.fish
New file: /home/alfa/.config/fish/test/functions/aur.fish
New file: /home/alfa/.config/fish/test/functions/fish_describe_machine.fish
New file: /home/alfa/.config/fish/test/functions/fish_set_cursor.fish
> fish_delta --no-completions --no-new --no-diff
File /home/alfa/.config/fish/functions/fish_clipboard_copy.fish differs from /usr/share/fish/functions/fish_clipboard_copy.fish
File /home/alfa/.config/fish/functions/fish_greeting.fish differs from /usr/share/fish/functions/fish_greeting.fish
File /home/alfa/.config/fish/functions/fish_prompt.fish differs from /usr/share/fish/functions/fish_prompt.fish
```

(and without "-d"/"--no-diff" it would show you the diff on changed files, with automatic paging - but that's awkward to demonstrate here)

One wrinkle is that it'll only show function *files* - if you've e.g. overridden a function in config.fish, we won't be showing it. I've tried working on that, but it's awkward considering multi-function files like fish_git_prompt.fish.

I think it's still useful, it'll allow me to figure out if I have a completion I'm testing (hence the test/completions directory!) and need to merge.

## TODOs:

- [x] Should we ignore vendor stuff, by default?
- [x] Should we ignore new files by default?

### DIFF
--- a/share/functions/fish_delta.fish
+++ b/share/functions/fish_delta.fish
@@ -1,0 +1,156 @@
+function fish_delta
+    argparse h/help f/no-functions c/no-completions C/no-config d/no-diff n/new V/vendor= -- $argv
+
+    if set -q _flag_help
+        __fish_print_help fish_delta
+        return 0
+    end
+
+    type -q diff
+    or set -l _flag_d --no-diff
+
+    set -l vendormode default
+    if set -ql _flag_vendor[1]
+        if not contains -- $_flag_vendor[-1] default ignore user
+            echo "Wrong vendor mode '$_flag_vendor[-1]'." >&2
+            echo "Valid values are: default, ignore, new" >&2
+            return 2
+        end
+        set vendormode $_flag_vendor[-1]
+    end
+
+    # TODO: Do we want to keep the vendor dirs in here?
+    set -l default_function_path $__fish_data_dir/functions
+    test "$vendormode" = default && set -a default_function_path $__fish_vendor_functionsdirs
+
+    set -l default_complete_path $__fish_data_dir/completions
+    test "$vendormode" = default && set -a default_completions_path $__fish_vendor_completionsdirs
+
+    set -l default_conf_path
+    test "$vendormode" = default && set -a default_conf_path $__fish_vendor_confdirs
+
+    set -l user_function_path
+    set -l user_complete_path
+    set -l user_conf_path $__fish_config_dir/conf.d $__fish_sysconf_dir/conf.d
+    test "$vendormode" = user && set -a user_conf_path $__fish_vendor_confdirs
+
+    for dir in $fish_function_path
+        if contains -- $vendormode ignore default
+            contains -- $dir $__fish_vendor_functionsdirs
+            and continue
+        end
+        contains -- $dir $default_function_path
+        or set -a user_function_path $dir
+    end
+    for dir in $fish_complete_path
+        if contains -- $vendormode ignore default
+            contains -- $dir $__fish_vendor_completionsdirs
+            and continue
+        end
+        # We don't care about generated completions.
+        # They shouldn't be compared at all.
+        contains -- $dir $default_complete_path $__fish_user_data_dir/generated_completions
+        or set -a user_complete_path $dir
+    end
+
+    set -l vars
+    not set -ql _flag_f[1]
+    and set -a vars user_function_path default_function_path
+    not set -ql _flag_c[1]
+    and set -a vars user_complete_path default_complete_path
+    not set -ql _flag_C[1]
+    and set -a vars user_conf_path default_conf_path
+
+    if not set -q vars[1]
+        echo No directories to check! >&2
+        return 2
+    end
+
+    set -l have_diff 0
+
+    if isatty stdout
+        set -f colors "$(set_color normal)" "$(set_color brblue)" "$(set_color bryellow)" "$(set_color green)" "$(set_color red)"
+        set -f pager less
+        set -q PAGER
+        and echo $PAGER | read -at pager
+        if type -q less
+            set -l lessopts isRF
+            if test (less --version | string match -r 'less (\d+)')[2] -lt 530 2>/dev/null
+                set lessopts "$lessopts"X
+            end
+
+            not set -qx LESS
+            and set -xf LESS $lessopts
+        end
+    else
+        set -f colors "" "" "" "" ""
+        set -f pager cat # cannot use a builtin here
+    end
+
+    begin
+        while set -q vars[1]
+            set -l user_var $$vars[1]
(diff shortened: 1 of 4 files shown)


faho:
Okay, ignoring new files makes more sense - they don't *change* anything, they aren't relevant for debugging etc, and there may be a lot of them.

I'm still not happy with the vendor dirs - there are three ways to handle them:

1. Treat them as "user" customization. Anything they do counts as "changed" or "new" compared to what we ship
2. Treat them as "default" - anything the user has of the same name as them counts as "changed"
3. Ignore them, like we do the generated completions - anything the user has of the same name counts as "new" (unless it's also in our dirs, in which case it might be changed)

In a certain sense, number 1 is true. These aren't things we ship, they are customizations compared to a stock fish.

But in another sense, that's entirely not the same as a user overriding things. *I* have a lot in these directories simply because my distro packages ship stuff there! So it's "default" according to what my system ships.

So number 3 feels like the least annoying option, possibly with a flag to switch to 1. (but tbh our vendor path logic is overcomplicated, especially with all the $XDG_DATA_DIRS stuff, and there's an actual possibility we end up getting these paths wrong)

krobelus:
this is intriguing, seems very useful.

> I'm still not happy with the vendor dirs

On a first glance, 2 or 3 sound like viable options.
I certainly would like to simplify the vendor loading logic, somehow..

share/functions/fish_delta.fish
```diff
@@ -0,0 +1,100 @@
...
+            end
+
+            if type -q diff
+                # We execute diff twice - once to figure out if it's changed,
+                # so we can get nicer output.
+                #
+                if not diff -q -- $file $def >/dev/null 2>&1
+                    printf (_ "%sChanged%s: %s\n") $colors[3] $colors[1] $file
+                    not set -ql _flag_d[1]
+                    and diff -u $diff_color  -- $file $def
```
krobelus, line 88: I'm surprised that the user's file is the "old" file in this diff.
I'd expect the other argument order, so the default version is the "old" one and the user modifications are "new".

share/functions/fish_delta.fish
```diff
@@ -0,0 +1,100 @@
...
+        set -f pager less
+        set -q PAGER
+        and echo $PAGER | read -at pager
+        set -l lessopts isRF
+        if test (less --version | string match -r 'less (\d+)')[2] -lt 530 2>/dev/null
+            set lessopts "$lessopts"X
+        end
+
+        not set -qx LESS
+        and set -xf LESS $lessopts
```
krobelus, line 55: I wonder if theres a good way to extract the paging logic.
A function will probably not work because we need to pipe into it.. maybe we can create a function that outputs the pager command

share/functions/fish_delta.fish
```diff
@@ -0,0 +1,100 @@
...
+                set -ql _flag_n
+                and printf (_ "%sNew%s: %s\n") $colors[2] $colors[1] $file
+                continue
+            end
+
+            if type -q diff
+                # We execute diff twice - once to figure out if it's changed,
+                # so we can get nicer output.
+                #
+                if not diff -q -- $file $def >/dev/null 2>&1
```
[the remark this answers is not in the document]
faho: Why is it faster *in theory*?

We give `diff` enough information here that it knows we just want to know if there's a difference. That means it could quit early or even use whatever faster path `cmp` uses.

Regardless - this isn't performance-sensitive (it's quite fast in practice *and* it's okay if this takes a while), and I would rather depend on one command being present and about what we expect than two.

In fact I probably need to see if `--color=always` needs to be checked.

share/functions/fish_delta.fish
faho: You're right, done!

faho:
>In fact I probably need to see if --color=always needs to be checked.

Okay I wasn't in the mood to do feature detection so I just added coloring via `string replace`. I await your screams.

faho:
Alright I implemented a "--vendor=" option to allow picking how these are counted.

By default they are counted as "default" (or "stock") files.

Note: This necessitates new global variables set via share/config.fish - along with e.g. $__fish_data_dir we now have $__fish_vendor_functionsdirs, $__fish_vendor_completionsdirs and $__fish_vendor_confdirs. These are set to the directories we used on startup.

The alternative is to redo the logic for finding the vendor files in fish_delta, which appears worse.

It just feels kind of awkward to make that change *here*, in a semi-related PR, but on the other hand I also don't love doing a separate PR and threading that in here?

mqudsi:
I think it helps to decide what the biggest use case for this would be in determining how to handle vendor dirs.

e.g. is this for the user's discoverability or for us to be able to determine if the user is running a stock config or not (we could ask them to run this when they file an issue and it appears that it's a config-related problem).

faho:
>e.g. is this for the user's discoverability or for us to be able to determine if the user is running a stock config or not (we could ask them to run this when they file an issue and it appears that it's a config-related problem).

As far as I'm concerned: Both.

One use case is finding out "what have *I* changed" - as an example I, personally, used previous evolutions of this to figure out what stuff I have left to commit, because I like test-driving things in my personal config for a while.

Another is finding out "what *is* changed over the stock configuration". This is both for debugging - bug reports to us could include something like this[0] - and for figuring out what sort of stuff differs from a stock fish so you can redo it on other systems.

Both of these necessitate treating vendor dirs differently - in the former, you don't really care about the separation between "the fish project's stuff", so it should all be counted together. This should be the most common and so it is the default mode of operation.

In the latter, you want to know what changed over "the fish project's stuff", so this is "--vendor=user" mode.

The idea behind "--vendor=ignore" was that you'd use it if the vendor stuff is annoying. You don't really choose what gets installed there and removing stuff there is frowned upon (because it's installed from packages and will just be overwritten on update), so you might end up with a bunch of vendor stuff that you don't care about.

---
[0]: But note that people love their aliases in config.fish and this won't help with that, so it's not a panacea.

mqudsi:
Do you want to add it to the changelog right now, while it's still fresh in your mind?

faho:
>Do you want to add it to the changelog right now, while it's still fresh in your mind?

I find doing changelog in the PR only leads to annoying merge conflicts.

PR #9261, opened 2022-10-05
Add feature to fish_commandline_prepend and fix minor issue

charlesmaher:
## Description

Prepending will now respect leading spaces instead of doubling it up, and removing a prefix no longer sends the cursor to the end of the line.

```sh
# new feature:
~>  vi /etc/fstab
#  ^ one leading space

# press <alt+s>

# before this change
~> sudo  vi /etc/fstab
#      ^^ two spaces

# after
~>  sudo vi /etc/fstab
#  ^    ^ one space either side

## fixed issue:
~> sudo systemctl status
#                ^ cursor location

# press <alt+s>

# before
~> systemctl status
#                  ^ cursor location
# always moved to the end

# after
~> systemctl status
#           ^ cursor location
# same relative placement
```

## TODOs:

### DIFF
--- a/share/functions/fish_commandline_prepend.fish
+++ b/share/functions/fish_commandline_prepend.fish
@@ -3,14 +3,24 @@ function fish_commandline_prepend --description "Prepend the given string to the
         commandline -r $history[1]
     end
 
-    set -l escaped (string escape --style=regex -- $argv[1])
     set -l process (commandline -p | string collect)
-    if set process (string replace -r -- "^$escaped " "" $process)
-        commandline --replace -p -- $process
-    else
-        set -l cursor (commandline -C)
-        commandline -C 0 -p
-        commandline -i -- "$argv[1] "
-        commandline -C (math $cursor + (string length -- "$argv[1] "))
+
+    set -l to_prepend "$argv[1] "
+    if string match -qr '^ ' "$process"
+        set to_prepend " $argv[1]"
     end
+
+    set -l length_diff (string length -- "$to_prepend")
+    set -l cursor_location (commandline -pC)
+
+    set -l escaped "$(string escape --style=regex -- $to_prepend)"
+    if set process (string replace -r -- "^$escaped" "" $process)
+        commandline --replace -p -- $process
+        set length_diff "-$length_diff"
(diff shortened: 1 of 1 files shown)


share/functions/fish_commandline_prepend.fish
```diff
@@ -3,14 +3,24 @@ function fish_commandline_prepend --description "Prepend the given string to the
...
+
+    set -l to_prepend "$argv[1] "
+    if string match -qr '^ ' "$process"
+        set to_prepend " $argv[1]"
+    end
+
+    set -l length_diff (string length -- "$to_prepend")
+    set -l cursor_location (commandline -pC)
+
+    set -l escaped "$(string escape --style=regex -- $to_prepend)"
```
charlesmaher, line 16: Now that I'm looking at it again I think this might be better:
```fish
set -l escaped (string escape --style=regex -- "$to_prepend")
```

Let me know if I should change it, or if it doesn't matter.

mqudsi: It doesn't really matter in this context - `$(...)` only makes a difference if you're trying to avoid splitting multiple lines into an array and want them as a single string. However the convention in our codebase would be to not use a quoted subshell. You also don't need the quotes in `"$to_prepend"` in your revised alternative either, as all fish variables are implicitly quoted when passed as parameters. Isn't fish just too easy?

mqudsi:
Above discussion not withstanding, this looks great and it's a nice bugfix. Thank you.

I'll let you update the patch to comply with conventions and then we can merge this pronto!

share/functions/fish_commandline_prepend.fish
charlesmaher: > You also don't need the quotes in `"$to_prepend"`

Right you are, that does make things easier. Thanks!

charlesmaher:
I just pushed your suggested changes, and assuming there's no other concerns I'm happy for this to be merged!

I'm also somewhat interested in abstracting this code so it can be used in `fish_commandline_append` as well, but that's a project for another day...

mqudsi:
Merged with thanks.

PR #9262, opened 2022-10-05
Update $fish_cursor_selection_mode in vi/default bindings

sshilovsky:
Introduced with 3.6.0 `fish_cursor_selection_mode` variable breaks existing vi bindings (for example, input sequence `abc<Esc>0vd` doesn't delete the `a` character as would be expected).

This patch fixes it by switching `fish_cursor_selection_mode` to `inclusive` and back.

## TODOs:


faho:
Sorry for the delay, merged!

(I think this would need to be changed with the cursor shape in future)

PR #9266, opened 2022-10-08
Make `output_stream_t::append()` fallible and handle SIGINT on append

mqudsi:
This patch series addresses a long-standing issue (`output_stream_t::append()` is assumed to never fail and returns void) that made it difficult to handle output termination via Ctrl-C (by halting progress and ending early) for various builtins with lots of output, `history` being the best example because it wrote to the pager (giving the user plenty of time and reason to `Ctrl-C` the running process, but to no effect besides an error message).

------

**Make `output_stream_t::append()` fallible**

Allow errors encountered by certain implementations of `output_stream_t` when
writing to the output sink to be bubbled back to the caller.

**Silently handle fd_output_stream_t append errors in case of SIGINT**

If `EINTR` caused by `SIGINT` is encountered while writing to the
`fd_output_stream_t` output fd, mark the output stream as errored and return
false to the caller but do not visibly complain.

Addressing the outstanding TODO notwithstanding, this is needed to avoid
littering the tty with spurious errors when the user hits Ctrl-C to abort a
long-running builtin's output (w/ the primary example being `history`).

**history: Handle Ctrl-C/SIGINT or other errors on output append**

When there are multiple screens worth of output and `history` is writing to the
pager, pressing Ctrl-C at the end of a screen doesn't exit the pager (`q` is
needed for that) but previously caused fish to emit an error ("write:
Interrupted system call) until we started silently handling SIGINT in
`fd_output_stream_t::append()`.

This patch makes `history` detect when the `append()` call returns with an error
and causes it to end early rather than repeatedly trying (and failing) to write
to the output stream.

### DIFF
--- a/src/history.cpp
+++ b/src/history.cpp
@@ -1482,8 +1482,10 @@ bool history_t::search(history_search_type_t search_type, const wcstring_list_t 
     wcstring_list_t collected;
     wcstring formatted_record;
     size_t remaining = max_items;
+    bool output_error = false;
 
-    // The function we use to act on each item.
+    // The function we use to act on each item. The return value indicates whether the search should
+    // continue (true) or stop (on false).
     std::function<bool(const history_item_t &item)> func = [&](const history_item_t &item) -> bool {
         if (remaining == 0) return false;
         remaining -= 1;
@@ -1493,7 +1495,11 @@ bool history_t::search(history_search_type_t search_type, const wcstring_list_t 
             collected.push_back(std::move(formatted_record));
         } else {
             // We can output this immediately.
-            streams.out.append(formatted_record);
+            if (!streams.out.append(formatted_record)) {
+                // This can happen if the user hit Ctrl-C to abort (maybe after the first page?).
+                output_error = true;
+                return false;
+            }
         }
         return true;
     };
@@ -1514,9 +1520,16 @@ bool history_t::search(history_search_type_t search_type, const wcstring_list_t 
     }
 
     // Output any items we collected (which only happens in reverse).
-    for (auto iter = collected.rbegin(); iter != collected.rend(); ++iter) {
-        streams.out.append(*iter);
+    for (auto iter = collected.rbegin(); !output_error && iter != collected.rend(); ++iter) {
+        if (!streams.out.append(*iter)) {
+            // Don't force an error if output was aborted (typically via Ctrl-C/SIGINT); just don't
+            // try writing any more.
+            output_error = true;
+        }
     }
+
+    // We are intentionally not returning false in case of an output error, as the user aborting the
+    // output early (the most common case) isn't a reason to exit w/ a non-zero status code.
     return true;
 }
 
--- a/src/io.cpp
+++ b/src/io.cpp
@@ -338,8 +338,17 @@ bool fd_output_stream_t::append(const wchar_t *s, size_t amt) {
     if (errored_) return false;
     int res = wwrite_to_fd(s, amt, this->fd_);
     if (res < 0) {
-        // TODO: this error is too aggressive, e.g. if we got SIGINT we should not complain.
-        if (errno != EPIPE) {
+        // Some of our builtins emit multiple screens worth of data sent to a pager (the primary
+        // example being the `history` builtin) and receiving SIGINT should be considered normal and
+        // non-exceptional (user request to abort via Ctrl-C), meaning we shouldn't print an error.
+        if (errno == EINTR && sigcheck_.check()) {
+            // We have two options here: we can either return false without setting errored_ to
+            // true (*this* write will be silently aborted but the onus is on the caller to check
+            // the return value and skip future calls to `append()`) or we can flag the entire
+            // output stream as errored, causing us to both return false and skip any future writes.
+            // We're currently going with the latter, especially seeing as no callers currently
+            // check the result of `append()` (since it was always a void function before).
+        } else if (errno != EPIPE) {
             wperror(L"write");
         }
         errored_ = true;
--- a/src/io.cpp
+++ b/src/io.cpp
@@ -206,9 +206,10 @@ void io_chain_t::push_back(io_data_ref_t element) {
     std::vector<io_data_ref_t>::push_back(std::move(element));
 }
 
-void io_chain_t::append(const io_chain_t &chain) {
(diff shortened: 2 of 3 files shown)


src/io.cpp
```diff
@@ -308,59 +309,74 @@ shared_ptr<const io_data_t> io_chain_t::io_for_fd(int fd) const {
...
+            // We have two options here: we can either return false without setting errored_ to
+            // true (*this* write will be silently aborted but the onus is on the caller to check
+            // the return value and skip future calls to `append()`) or we can flag the entire
+            // output stream as errored, causing us to both return false and skip any future writes.
+            // We're currently going with the latter, especially seeing as no callers currently
+            // check the result of `append()` (since it was always a void function before).
+        } else if (errno != EPIPE) {
             wperror(L"write");
         }
         errored_ = true;
```
faho, line 354: Are there any cases where we *shouldn'*t error?

E.g. if we get EINTR for other signals? EAGAIN?

This now is quite eager to stop writing and I'm scared.

mqudsi: > This now is quite eager to stop writing and I'm scared.

The behavior hasn't changed. It always set `errored_ = true;` in case of `EINTR` but now we just don't call `wperror(..)` before doing so.

src/io.cpp
```diff
@@ -308,59 +309,74 @@ shared_ptr<const io_data_t> io_chain_t::io_for_fd(int fd) const {
...
     // Return a generic 1 on any write failure.
     return errored_ ? STATUS_CMD_ERROR : STATUS_CMD_OK;
 }
 
-void null_output_stream_t::append(const wchar_t *, size_t) {}
+bool null_output_stream_t::append(const wchar_t *, size_t) { return true; }
 
-void string_output_stream_t::append(const wchar_t *s, size_t amt) { contents_.append(s, amt); }
+bool string_output_stream_t::append(const wchar_t *s, size_t amt) {
+    contents_.append(s, amt); return true;
```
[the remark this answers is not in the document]
mqudsi: Oh, I meant to run `git-clang-format` but forgot. I'll do that and force push.

src/io.cpp
```diff
@@ -308,59 +309,75 @@ shared_ptr<const io_data_t> io_chain_t::io_for_fd(int fd) const {
...
 
-void fd_output_stream_t::append(const wchar_t *s, size_t amt) {
-    if (errored_) return;
+bool fd_output_stream_t::append(const wchar_t *s, size_t amt) {
+    if (errored_) return false;
     int res = wwrite_to_fd(s, amt, this->fd_);
     if (res < 0) {
-        // TODO: this error is too aggressive, e.g. if we got SIGINT we should not complain.
-        if (errno != EPIPE) {
+        // Some of our builtins emit multiple screens worth of data sent to a pager (*cough* history
```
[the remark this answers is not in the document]
mqudsi: It's not a sarcastic cough but rather an accusatory one!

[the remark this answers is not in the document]
faho: But why would you accuse it of doing its job, correctly?

src/io.cpp
faho: Huh, you're right, now I'm confused as to why I've never seen an error for EINTR being printed here.

src/io.cpp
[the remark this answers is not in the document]
faho: (just to be clear this isn't a big thing, I'm just unclear on what this comment means to convey with its tone)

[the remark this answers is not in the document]
mqudsi: It was just a more playful way of saying "(especially `history`)". It doesn't matter, I've changed it.

src/io.cpp
```diff
@@ -308,59 +309,75 @@ shared_ptr<const io_data_t> io_chain_t::io_for_fd(int fd) const {
...
+        // Some of our builtins emit multiple screens worth of data sent to a pager (the primary
+        // example being the `history` builtin) and receiving SIGINT should be considered normal and
+        // non-exceptional (user request to abort via Ctrl-C), meaning we shouldn't print an error.
+        if (errno == EINTR && sigcheck_.check()) {
+            // We have two options here: we can either return false without setting errored_ to
+            // true (*this* write will be silently aborted but the onus is on the caller to check
+            // the return value and skip future calls to `append()`) or we can flag the entire
+            // output stream as errored, causing us to both return false and skip any future writes.
+            // We're currently going with the latter, especially seeing as no callers currently
+            // check the result of `append()` (since it was always a void function before).
```
krobelus, line 350: > we can flag the entire output stream as errored, causing us to both return false and skip any future writes.

I guess for EINTR that's somewhat implied because SIGINT cancels builtin execution.

src/history.cpp
```diff
@@ -1514,9 +1520,16 @@ bool history_t::search(history_search_type_t search_type, const wcstring_list_t
...
+    for (auto iter = collected.rbegin(); !output_error && iter != collected.rend(); ++iter) {
+        if (!streams.out.append(*iter)) {
+            // Don't force an error if output was aborted (typically via Ctrl-C/SIGINT); just don't
+            // try writing any more.
+            output_error = true;
+        }
     }
+
+    // We are intentionally not returning false in case of an output error, as the user aborting the
+    // output early (the most common case) isn't a reason to exit w/ a non-zero status code.
```
krobelus, line 1532: .. but in this case we likely exit with a non-zero signal code like SIGINT = 130

src/history.cpp
```diff
@@ -1514,9 +1520,16 @@ bool history_t::search(history_search_type_t search_type, const wcstring_list_t
...
 
     // Output any items we collected (which only happens in reverse).
-    for (auto iter = collected.rbegin(); iter != collected.rend(); ++iter) {
-        streams.out.append(*iter);
+    for (auto iter = collected.rbegin(); !output_error && iter != collected.rend(); ++iter) {
+        if (!streams.out.append(*iter)) {
+            // Don't force an error if output was aborted (typically via Ctrl-C/SIGINT); just don't
+            // try writing any more.
+            output_error = true;
+        }
```
krobelus, line 1528: history is the worst offender (can't seem to hit the `wperror("write")` on Linux though).
There are probably more cases where we could stop writing early (`string repeat`) etc.

Anyway, this seems like a good step forward, LGTM!

PR #9271, opened 2022-10-11
fish_clipboard_copy/paste: Handle redirected stdout/stdin

faho:
## Description

This makes these tools usable in a pipe.

You can run

```fish
some-long-command | fish_clipboard_copy
```

to copy some command's output to your clipboard, and

```fish
fish_clipboard_paste | some-other-command
```

To feed your clipboard to some command.

For instance, let's say I want to insert my patented "commits missing from release X" table.

I've written a function called "gitcomp" to generate the info, and I have another function "mdtable" that transforms it to markdown. But I need it in my clipboard so I can paste it in the browser, and previously I'd use `xsel --clipboard` for that, which is very platform-dependent - if I switched to wayland I'd suddenly have to get used to a different thing!

So now I can instead do:

```fish
LANG=en_US.UTF-8 gitcomp | tail -n 5 | mdtable -d\t release commits percent date | fish_clipboard_copy
```

and paste directly into my browser to get:

|release|commits|percent|date|
|:---|:---|:---|:---|
|3.4.0|1106|5.7%|2022-03-12 22:31:57 +0800 (7 months ago)|
|3.4.1|9|5.7%|2022-03-26 00:27:56 +0800 (7 months ago)|
|3.5.0|433|3%|2022-06-16 21:02:33 +1000 (4 months ago)|
|3.5.1|24|2.9%|2022-07-20 18:15:47 +0800 (3 months ago)|
|HEAD|469|0%||

and if I want to work on something from my clipboard, I can do

```fish
fish_clipboard_paste | something
```

and be reasonably sure it'll work.

## TODOs:

### DIFF
--- a/share/functions/fish_clipboard_copy.fish
+++ b/share/functions/fish_clipboard_copy.fish
@@ -1,7 +1,17 @@
 function fish_clipboard_copy
-    # Copy the current selection, or the entire commandline if that is empty.
-    set -l cmdline (commandline --current-selection | string collect)
-    test -n "$cmdline"; or set cmdline (commandline | string collect)
+    set -l cmdline
+    if isatty stdin
+        # Copy the current selection, or the entire commandline if that is empty.
+        # Don't use `string collect -N` here - `commandline` adds a newline.
+        set cmdline (commandline --current-selection | string collect)
+        test -n "$cmdline"; or set cmdline (commandline | string collect)
+    else
+        # Read from stdin
+        while read -lz line
+            set -a cmdline $line
+        end
+    end
+
     if type -q pbcopy
         printf '%s' $cmdline | pbcopy
     else if set -q WAYLAND_DISPLAY; and type -q wl-copy
--- a/share/functions/fish_clipboard_paste.fish
+++ b/share/functions/fish_clipboard_paste.fish
@@ -1,25 +1,30 @@
 function fish_clipboard_paste
     set -l data
     if type -q pbpaste
-        set data (pbpaste 2>/dev/null)
+        set data (pbpaste 2>/dev/null | string collect -N)
     else if set -q WAYLAND_DISPLAY; and type -q wl-paste
-        set data (wl-paste 2>/dev/null)
+        set data (wl-paste -n 2>/dev/null | string collect -N)
     else if set -q DISPLAY; and type -q xsel
-        set data (xsel --clipboard)
+        set data (xsel --clipboard | string collect -N)
     else if set -q DISPLAY; and type -q xclip
-        set data (xclip -selection clipboard -o 2>/dev/null)
+        set data (xclip -selection clipboard -o 2>/dev/null | string collect -N)
     else if type -q powershell.exe
-        set data (powershell.exe Get-Clipboard | string trim -r -c \r)
+        set data (powershell.exe Get-Clipboard | string trim -r -c \r | string collect -N)
     end
 
     # Issue 6254: Handle zero-length clipboard content
-    if not string match -qr . -- "$data"
+    if not string length -q -- "$data"
         return 1
     end
 
-    # Also split on \r to turn it into a newline,
-    # otherwise the output looks really confusing.
-    set data (string split \r -- $data)
+    if not isatty stdout
+        # If we're redirected, just write the data *as-is*.
+        printf %s $data
+        return
+    end
+
+    # Also split on \r, otherwise it looks confusing
+    set data (string split \r -- $data | string split \n)
 
     # If the current token has an unmatched single-quote,
     # escape all single-quotes (and backslashes) in the paste,
--- a/doc_src/cmds/fish_clipboard_copy.rst
+++ b/doc_src/cmds/fish_clipboard_copy.rst
@@ -1,0 +1,36 @@
+.. _cmd-fish_clipboard_copy:
+
+fish_clipboard_copy - copy text to the system's clipboard
+==============================================================
+
+Synopsis
+--------
+
+.. synopsis::
+
+    fish_clipboard_copy
+
+    foo | fish_clipboard_copy
+
+Description
+-----------
+
+The ``fish_clipboard_copy`` function copies text to the system clipboard.
+
+If stdin is not a terminal (see :doc:`isatty <isatty>`), it will read all input from there and copy it. If it is, it will use the current commandline, or the current selection if there is one.
+
+It is bound to :kbd:`Control`\ +\ :kbd:`X` by default.
+
+``fish_clipboard_copy`` works by calling a system-specific backend. If it doesn't appear to work you may need to install yours.
+
+Currently supported are:
+
+- ``pbcopy``
+- ``wl-copy`` using wayland
+- ``xsel`` and ``xclip`` for X11
+- ``clip.exe`` on Windows.
+
+See also
+--------
+
+- :doc:`fish_clipboard_paste` which does the inverse.
--- a/doc_src/cmds/fish_clipboard_paste.rst
+++ b/doc_src/cmds/fish_clipboard_paste.rst
@@ -1,0 +1,38 @@
+.. _cmd-fish_clipboard_paste:
+
+fish_clipboard_paste - get text from the system's clipboard
+==============================================================
+
+Synopsis
+--------
+
+.. synopsis::
+
+    fish_clipboard_paste
+
+    fish_clipboard_paste | foo
+
+Description
+-----------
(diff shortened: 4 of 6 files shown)


mqudsi:
+1 for me; I forgot this wasn't already supported. It's one of the many clipboard-related things I can't live without as part of [my fish config](https://github.com/mqudsi/fish-config). 

I also have support for copying the contents of a file or pasting the clipboard to a file when invoked as `fish_clipboard_copy source.txt` or `fish_clipboard_paste dest.txt`. This is, of course, just a shortcut for `fish_clipboard_copy < source.txt` and `fish_clipboard_paste > dest.txt` but I found it worth adding support for given how often I do it to copy contents between machines (together w/ the copy/paste-over-ssh support) or to a web browser (often for github issues).

share/functions/fish_clipboard_paste.fish
```diff
@@ -21,6 +21,12 @@ function fish_clipboard_paste
     # otherwise the output looks really confusing.
     set data (string split \r -- $data)
 
+    if not isatty stdout
+        # If we're redirected, just write the data
+        printf %s\n $data
```
mqudsi, line 26: I used to have this as part of my config but it broke my integration with tmux and neovim. (Instead of rewriting all my clipboard handling each time for each application that wants to integrate, I just farm out the clipboard operation to fish.)

The problem is that if you forcibly insert line breaks, some applications won't strip them out on paste (at least by default). I wanted to be able to copy a word to the clipboard in neovim/tmux and then paste it into another doc in another terminal instance, but emitting a trailing `\n` causes a hard line break to be inserted everywhere I paste.

I've since modified my clipboard wrappers to omit any additional line breaks and I've found that to be a much more zero-stress solution. If you _do_ want a line break after pasting, you can always just hit enter/ctrl-j, but adding a line break can cause problems. We already handle commands that emit data without a trailing `\n` very nicely so there's no harm (to us) in excluding that.

Oh yeah, I remembered another issue I ran into: you also especially don't want to force either execution of the clipboard contents or forcibly starting a second line if the user is using `fish_clipboard_paste` to paste into another shell.

krobelus: I agree; I use a similar script for copying/pasting text on any OS and I also made it not produce newlines out of thin air.
That's important when using it to paste into a text editor.

share/functions/fish_clipboard_copy.fish
```diff
@@ -1,7 +1,16 @@
...
-    test -n "$cmdline"; or set cmdline (commandline | string collect)
+    set -l cmdline
+    if isatty stdin
+        # Copy the current selection, or the entire commandline if that is empty.
+        set cmdline (commandline --current-selection | string collect)
+        test -n "$cmdline"; or set cmdline (commandline | string collect)
+    else
+        # Read from stdin
+        while read -lz line
+            set -a cmdline $line
```
krobelus, line 10: That's a neat way to losslessly forward data; notablky it works if stdin does not have a trailing new line.
We should use it for the other as well when we are not reading from stdin. I can take care of that.

share/functions/fish_clipboard_paste.fish
faho: Should it not "produce newlines out of thin air", or have trailing newlines at all?

Those are two different things.

share/functions/fish_clipboard_copy.fish
faho: The other case can just be done via `string collect -N`, if you want to care about not removing trailing newlines (we already don't *add* any).

(we could also use that here if we had a `--stdin` flag to force `string` to read from stdin)

share/functions/fish_clipboard_paste.fish
krobelus: We should print the clipboard contents verbatim. I realized that we also need to make sure that we don't drop actual newlines. This should do it but it requires GNU sed AFAIK

```diff
diff --git a/share/functions/fish_clipboard_paste.fish b/share/functions/fish_clipboard_paste.fish
index 49a2dc1be..5509154ec 100644
--- a/share/functions/fish_clipboard_paste.fish
+++ b/share/functions/fish_clipboard_paste.fish
@@ -1,19 +1,19 @@
 function fish_clipboard_paste
     set -l data
     if type -q pbpaste
-        set data (pbpaste 2>/dev/null)
+        set data "$(pbpaste 2>/dev/null)E"
     else if set -q WAYLAND_DISPLAY; and type -q wl-paste
-        set data (wl-paste 2>/dev/null)
+        set data "$(wl-paste 2>/dev/null)E"
     else if set -q DISPLAY; and type -q xsel
-        set data (xsel --clipboard)
+        set data "$(xsel --clipboard)E"
     else if set -q DISPLAY; and type -q xclip
         set data (xclip -selection clipboard -o 2>/dev/null)
     else if type -q powershell.exe
-        set data (powershell.exe Get-Clipboard | string trim -r -c \r)
+        set data "$(powershell.exe Get-Clipboard | string trim -r -c \r)E"
     end
 
     # Issue 6254: Handle zero-length clipboard content
-    if not string match -qr . -- "$data"
+    if not string match -qr .E -- "$data"
         return 1
     end
 
@@ -23,10 +23,12 @@ function fish_clipboard_paste
 
     if not isatty stdout
         # If we're redirected, just write the data
-        printf %s\n $data
+        printf %s $data | sed '$s/E$//'
         return
     end
 
+    set data (printf %s $data | sed '$s/E$//')
+
     # If the current token has an unmatched single-quote,
     # escape all single-quotes (and backslashes) in the paste,
     # in order to turn it into a single literal token.
```

faho: >set data "$(powershell.exe Get-Clipboard | string trim -r -c \r)E"

You want `string collect -N`. No fiddling with `sed` or awkward special-insert chars. Just:

```fish
set data (xsel --clipboard | string collect -N)
```

share/functions/fish_clipboard_copy.fish
krobelus: Nice, `string collect -N` does the job!

share/functions/fish_clipboard_paste.fish
mqudsi: The default behavior on all platforms I have observed with clipboard functionality is that you can copy a single word (without including/selecting the line break) and then inspect the system clipboard contents as binary and observe that no trailing new line has been added. You should be able to select a word in the middle of a line, copy it, then select it and paste over it without any whitespace changes and without the client needing to intercept and eliminate trailing whitespace.

We don't have an issue with that *in fish* because both presence and absence of a trailing newline is coalesced in any subshell (i.e. `printf "%s" (echo foo)` and `printf "%s" (echo -n foo)` are identical). But when we touch the system clipboard, the ultimate target may or may not be fish and may or may not preserve trailing line breaks.

faho: 5bbe0073e258cf1ad22d48f79ac8b5dca874ea00 should do it.

krobelus: so we can just use `string collect -N` here as well to do the same thing

share/functions/fish_clipboard_paste.fish
```diff
@@ -1,25 +1,30 @@
 function fish_clipboard_paste
     set -l data
     if type -q pbpaste
-        set data (pbpaste 2>/dev/null)
+        set data (pbpaste 2>/dev/null | string collect -N)
     else if set -q WAYLAND_DISPLAY; and type -q wl-paste
-        set data (wl-paste 2>/dev/null)
+        set data (wl-paste 2>/dev/null | string collect -N)
```
krobelus, line 6: this one requires `wl-paste --no-newline` (no idea why that's not the default)

share/functions/fish_clipboard_paste.fish
```diff
@@ -1,25 +1,30 @@
...
-    # otherwise the output looks really confusing.
-    set data (string split \r -- $data)
+    if not isatty stdout
+        # If we're redirected, just write the data *as-is*.
+        printf %s $data
+        return
+    end
+
+    # Also split on \r, otherwise it looks confusing
+    set data (string split \r -- $data | string split \n)
```
krobelus, line 27: is the `string split \n` really necessary? Other than that, LGTM.

faho: It is, yes.

ilyagr:
Thank you so much! I just discovered this gem.

PR #9272, opened 2022-10-12
Add clojure completions

kthu:
## Description

Completions for the Clojure cli tool. Both `clj` and `clojure` is recognized.  Will complete aliases and tools if babahka is installed.

## TODOs:


share/completions/clj.fish
```diff
@@ -0,0 +1,79 @@
...
+       (mapcat aliases-from-config)
+       dedupe))
+
+(run! println
+      (case (first *command-line-args*)
+        "tools" (installed-tools)
+        (aliases)))'
+
+function __fish_clj_aliases
+    which bb >/dev/null 2>&1; or return 1
```
faho, line 42: This should be `command -q bb`.

`which` is a cheesy external tool that might not be installed and might not give the right answer.

share/completions/clj.fish
```diff
@@ -0,0 +1,79 @@
...
+    bb (echo $bb_helper | psub)
```
faho, line 43: Does `bb` take input from stdin?

If so, this could just be

```fish
echo $bb_helper | bb
```

or possibly `echo $bb_helper | bb -`.

share/completions/clj.fish
```diff
@@ -0,0 +1,79 @@
...
+end
+
+function __fish_clj_tools
+    which bb >/dev/null 2>&1; or return 1
```
faho, line 47: (same as above)

share/completions/clj.fish
```diff
@@ -0,0 +1,79 @@
...
+complete -c clj -s J -d "Pass opt through in java_opts, ex: -J-Xmx512m"
+complete -c clj -s i -l init -r -d "Load a file or resource"
+complete -c clj -f -s e -l eval -r -d "Eval exprs in string; print non-nil values"
+complete -c clj -f -l report -d "Report uncaught exception to \"file\" (default), \"stderr\", or \"none\""
+complete -c clj -f -o version -l version -d "Print the version to stdout and exit"
+complete -c clj -f -s m -l main -d "Call the -main function from namespace w/args"
+complete -c clj -f -s r -l repl -d "Run a repl"
+complete -c clj -r -d "Run a script from a file or resource"
+
+complete -c clojure --wraps clj
```
faho, line 79: This won't be loaded at the right time.

Fish finds completion files by command name, so if you try to complete "clojure" it will look for "clojure.fish".

That means once you have tried completing "clj" in the current session, the wrapper will be set up. But if you start a new shell and try to complete "clojure" before "clj" it won't work.

You'll have to add a file called "clojure.fish" that includes just this single line.

However: That's only if "clojure" is a common name set *by upstream* - if it's typically *installed* as both "clojure" and "clj".

If users usually set this up themselves, e.g. by using `alias clojure=clj`, then please just drop this. (`alias` would set up the wrapping)

faho (changes requested):

share/completions/clj.fish
kthu: It can, but it will make noise:

```
❯ echo "(+ 1 2)" | bb -i
Babashka v0.10.163 REPL.
Use :repl/quit or :repl/exit to quit the REPL.
Clojure rocks, Bash reaches.

user=> 3
user=> ⏎                                                                                                                                                                                                          
~
❮ bb "(+ 1 2)"
3

~
❯
```

So I guess `psub` is needed?

share/completions/clj.fish
kthu: Both are typically installed and are widely used. `clj` is just `clojure` with a `rlwrap` wrapper for use in the repl. I'll make a separate clojure.fish file.

Thank you for the pointers. I'll fix the issues you've mentioned.

Clojure's command line is a bit odd. For the execution options -A, -M, -X and -T, it will take a colon separated list of aliases without any spaces between the option and the list (like `clj -Mdev:debug:build`). `__fish_complete_list` handles this nicely, but it would be nice if the description changed when making list suggestions, so that it first shows

```
❯ clj -<tab>
-A                      (Use concatenated aliases to modify classpath)  
...
```

but once the user has typed -A and hits tab it would show something like

```
❯ clj -Ab<tab>
-Abuild                 (Alias)  
-Abenchmark             (Alias)  
...
```

I was not able to figure out how to do this. It is possible?

share/completions/clj.fish
```diff
@@ -0,0 +1,78 @@
...
+       (mapcat aliases-from-config)
+       dedupe))
+
+(run! println
+      (case (first *command-line-args*)
+        "tools" (installed-tools)
+        (aliases)))'
+
+function __fish_clj_aliases -V bb_helper
+    command -q bb
```
faho, line 43: You still need to `; or return` here

share/completions/clj.fish
mqudsi: If the completions for `clj` and `clojure` are identical or close to it, you can always put them in a function that takes the command name as a parameter, and invoke that from both `clj.fish` and `clojure.fish`. There are other workarounds such as defining `clj` as `function clj --wraps clojure; command clj; end;` but that's the kind of thing you would do on your pc and not as part of the base fish installation.

faho:
Merged, thanks!

kthu:
Great! thanks

PR #9273, opened 2022-10-13
Fix `fish_config theme save`

### LINKED ISSUE #9088: `fish_config theme save` returns "Too few arguments"
fish 3.5.0 on macOS Big Sur 11.6.7

Basically, I'm following the example from https://fishshell.com/docs/current/faq.html#how-do-i-customize-my-syntax-highlighting-colors

`fish_config theme show` works
`fish_config theme choose [name_of_theme]` works
`fish_config theme save` returns "Too few arguments"

mqudsi:
The documentation states that running `fish_config theme save` after `fish_config theme choose [theme_name]` will result in "saving" the currently chosen theme, but this does not match the actual behavior of `fish_config theme save` which expects a trailing argument specifying the name of the theme to select/persist.

Given that the documented way has been included in a release and that it makes more sense than calling `fish_config theme save xxx` when you are *loading from* xxx and not *saving to* xxx, this patch revises `fish_config.fish` to support the documented behavior.

When `fish_config theme choose xxx` is used, xxx is saved to a hidden global variable `__fish_last_chosen_theme`. If `fish_config theme save` is called without a theme name specified, then this global variable is searched for the name of the last chosen theme and that is saved/persisted as the default theme.

Closes #9088.

---

I'm requesting a review because I have no idea how `fish_config` is meant to be used and I'm going solely off what I read in the function -- I've never used it myself. I'm just conforming the code to the existing docs.

### DIFF
--- a/share/functions/fish_config.fish
+++ b/share/functions/fish_config.fish
@@ -205,35 +205,22 @@ function fish_config --description "Launch fish's web based configuration"
                         echo "Too many arguments" >&2
                         return 1
                     end
-                    if not set -q argv[1]
+                    # The name of the theme to save *from* is optional for `fish_config theme save`
+                    if not set -q argv[1] && contains -- $cmd choose
                         echo "Too few arguments" >&2
                         return 1
                     end
 
-                    set -l files $dir/$argv[1].theme
-                    set -l file
-
                     set -l scope -g
+                    set -l have_colors
+
                     if contains -- $cmd save
-                        read -P"Overwrite theme? [y/N]" -l yesno
+                        read -P"Overwrite your current theme? [y/N] " -l yesno
                         if not string match -riq 'y(es)?' -- $yesno
                             echo Not overwriting >&2
                             return 1
                         end
                         set scope -U
-                    end
-
-                    for f in $files
-                        if test -e "$f"
-                            set file $f
-                            break
-                        end
-                    end
-
-                    if not set -q file[1]
-                        echo "No such theme: $argv[1]" >&2
-                        echo "Dirs: $dir" >&2
-                        return 1
                     end
 
                     set -l known_colors fish_color_{normal,command,keyword,quote,redirection,\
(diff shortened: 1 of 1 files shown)


share/functions/fish_config.fish
```diff
@@ -205,17 +205,33 @@ function fish_config --description "Launch fish's web based configuration"
...
+                    if not set -q argv[1] && contains -- $cmd choose
                         echo "Too few arguments" >&2
                         return 1
                     end
 
+                    # If we are choosing a theme, save its name to a hidden global variable.
+                    # If we are persisting a theme, either get its name from $argv[1] or
+                    # from the last chosen theme.
+                    if contains -- $cmd choose
+                        set -g __fish_last_chosen_theme $argv[1]
```
faho, line 217: This isn't the right way to go about this.

The current theme is just the set of the current color variables, so running `fish_config theme save` should just persist the current variables (to universal variables)

share/functions/fish_config.fish
```diff
@@ -205,17 +205,33 @@ function fish_config --description "Launch fish's web based configuration"
...
+                    end
+
                     set -l files $dir/$argv[1].theme
                     set -l file
 
                     set -l scope -g
+
                     if contains -- $cmd save
-                        read -P"Overwrite theme? [y/N]" -l yesno
+                        read -P"Permanently overwrite default theme? [y/N] " -l yesno
```
faho, line 234: You wouldn't just overwrite the *default* theme. You'd overwrite whatever theme you currently have saved. So I'd just keep the wording here.

share/functions/fish_config.fish
faho: Basically, what you need to do is just not read any file if $argv is empty, and instead only do the saving part.

So an `if set -q argv[1]` on line 226 .

faho (changes requested):

share/functions/fish_config.fish
faho: In essence, `fish_config theme save` or `fish_config prompt save` is the "I'm happy with this" button. It just saves the current config.

So if you do `fish_config theme choose coolbeans`, and you decide that it's nice, and then you do `fish_config theme save` it should save the coolbeans theme, but if you decide to change a color or two and then `fish_config theme save` it should save the theme *with those alterations*.

mqudsi: That makes sense, thanks.

share/functions/fish_config.fish
faho: Okay, then "Overwrite your theme?" or "Overwrite currently saved theme?"?

(also that should be consistent with the above "Overwrite prompt?")

mqudsi: That works well.

mqudsi:
I pushed an update that addresses your concerns, @faho. (I didn't force push, but I'll squash the commits into the latter if we end up going with that direction).

Note that the latest commit includes an important caveat/note:

> This does *not* catch color variables unknown to fish! If a theme and a
> prompt agree on some variable to hold some color but it's not a color variable
> known to fish, it won't be persisted!

I think we have to search for all variables that match the whitelist filter we use when loading a theme from disk and persist all of those, not just the known colors? 

Basically, replace `for color in $known_colors` with `for color in $known_colors (set | string match -r '^fish_(?:pager_)?color.*$')`

faho:
>Basically, replace for color in $known_colors with for color in $known_colors (set | string match -r '^fish_(?:pager_)?color.*$')

`set --names`, but yeah, that's probably better.

mqudsi:
Actually, thanks for reminding me to use `--names`. I've made that mistake before!

I just pushed an update that does that; I think now we can comfortably use `fish_config theme save` to persist whatever theme was already loaded. 

Separate from this pull request, I think it might be more correct for loading ("choosing") a theme to unset any pattern-matched variables that weren't overwritten in the new theme; e.g. theme A sets '$fish_color_custom' then theme B is loaded - I think loading theme B should maybe erase `$fish_color_custom`? Because a subsequent `fish_config theme dump` will include theme A's variable even after theme B has been chosen.

But I'm not sure if there are prompts that set these colors; we wouldn't want `fish_config theme choose` to erase a variable set by a *prompt* rather than a different theme..

Issue #7120, opened 2020-06-15
Define what variable values count as "true"


faho:
We have a bunch of places where variables are used for configuration, and often they're used as booleans.

E.g. `__fish_git_prompt_show_informative_status` (yes, that's a handful) can be set, and turns on informative status by virtue of being defined.

However, using just the fact of it being defined has issues. For one, it's impossible to turn it *off* via a universal variable, and it's hard to turn it off globally (because you actually need to erase it instead of setting it). Also the choice to have something turned off can't really be differentiated from not having it set at all.

So instead I propose defining some standard set of values for a variable to count as "true" for the purposes of using it as a boolean, and adding a helper function or builtin (or possibly option to `set` or `test`) to test that.

If a variables actual *value* is to be used, that can of course still be done, and if it's unset we can still fall back to some default.

The question here is what the set of values should be? Just "1", which is the easiest check, or any non-zero or greater-than-zero number? Also strings like "Yes", "yes", "y", "j", "true"?

Using just "1" is simple and quick to test, using strings is friendly and compatible (e.g. with `git config`, which uses `true` as the true boolean).

zanchey:
So my key concern here is that the logical (aha) values to use here - 0 for false, 1 for true - are exactly opposite to what `if` uses! Using `set foo (true)` is more consistent, but a lot more typing. Words and initials aren't localisable either, though my unilingual impression is that most Unix systems require at least intermediate English?

faho:
Oh, no, using simple words and initials like "yes", "on" and "true" should be absolutely no problem - we already use `if`, `while`, `for`, `begin`, `end`, and of course the most important english word of all: `argparse`.

I've also thought about the difference with $status, but I don't think it's that large a problem. It boils down to what people intuitively use, and I *think* that's going to be 1 for true?

There's another point that makes this choice important, and that's backwards-compatibility. Because this, strictly speaking, isn't backwards-compatible. If people set the variable to a value that's now interpreted as falsy, the behavior would switch. If people already set these to `1` then the variable would remain being interpreted as truthy and they wouldn't have to change. (yes, this would have to be specifically called out in the release notes)

I think I'm going to check what people use.

IlanCosman:
I've simply been using `test "$blah" = 'true'` as a substitute for proper booleans, which allows me to use universal variables. It would definitely be nice to have a more standardized approach across fish though.

krobelus:
> test "$blah" = 'true'

[...]

PR #9274, opened 2022-10-13
git prompt: Interpret values as true for bools instead of relying on defined-or-not

faho:
This allows explicitly turning these settings off by setting the variable to e.g. 0.

See #7120

This is technically a change in behavior iff you set these variables to anything *but* "1". I don't think I've ever seen that, and it's easy enough to fix.

Note: This lacks the "hide" stuff (`__fish_git_prompt_hide_stagedstate` and friends), it's not documented, nigh-useless now that we properly look at the status order. I want to remove it.

## TODOs:

### DIFF
--- a/doc_src/cmds/fish_git_prompt.rst
+++ b/doc_src/cmds/fish_git_prompt.rst
@@ -28,14 +28,14 @@ There are numerous customization options, which can be controlled with git optio
 
 Boolean options (those which enable or disable something) understand "1", "yes" or "true" to mean true and every other value to mean false.
 
-- ``$__fish_git_prompt_show_informative_status`` or the git option ``bash.showInformativeStatus`` can be set to enable the "informative" display, which will show a large amount of information - the number of dirty files, unpushed/unpulled commits, and more.
+- ``$__fish_git_prompt_show_informative_status`` or the git option ``bash.showInformativeStatus`` can be set to 1, true or yes to enable the "informative" display, which will show a large amount of information - the number of dirty files, unpushed/unpulled commits, and more.
   In large repositories, this can take a lot of time, so you may wish to disable it in these repositories with  ``git config --local bash.showInformativeStatus false``. It also changes the characters the prompt uses to less plain ones (``✚`` instead of ``*`` for the dirty state for example) , and if you are only interested in that, set ``$__fish_git_prompt_use_informative_chars`` instead.
 
   Because counting untracked files requires a lot of time, the number of untracked files is only shown if enabled via ``$__fish_git_prompt_showuntrackedfiles`` or the git option ``bash.showUntrackedFiles``.
 
-- ``$__fish_git_prompt_showdirtystate`` or the git option ``bash.showDirtyState`` can be set to show if the repository is "dirty", i.e. has uncommitted changes.
+- ``$__fish_git_prompt_showdirtystate`` or the git option ``bash.showDirtyState`` can be set to 1, true or yes to show if the repository is "dirty", i.e. has uncommitted changes.
 
-- ``$__fish_git_prompt_showuntrackedfiles`` or the git option ``bash.showUntrackedFiles`` can be set to show if the repository has untracked files (that aren't ignored).
+- ``$__fish_git_prompt_showuntrackedfiles`` or the git option ``bash.showUntrackedFiles`` can be set to 1, true or yes to show if the repository has untracked files (that aren't ignored).
 
 - ``$__fish_git_prompt_showupstream`` can be set to a list of values to determine how changes between HEAD and upstream are shown:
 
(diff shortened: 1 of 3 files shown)


mqudsi:
I'm not sure how I feel about this. I agree that it should be possible to set them to some value to indicate that they are disabled, but to assume that in a loosely typed scripting language everyone will by default understand or muscle memory default to a numeric `1` for truthiness is not sitting well with me.

You've gone from accepting any value as "true" but not having a way to express "false" to understanding every variable to be "false" except for the one and only value reserved as "true".

In writing GUI frontends for CLIs, I have found the greatest success in considering any of `on`, `true`, `1`, or `yes` (all case-insensitive, of course) to be equivalent to `true`. I'm not suggesting we document that (by all means, go ahead and keep the documentation the same and insist that users use `1`) but we should be more liberal in parsing even if that means we need a helper function like `is_true <var_name>`.

faho:
>I agree that it should be possible to set them to some value to indicate that they are disabled, but to assume that in a loosely typed scripting language everyone will by default understand or muscle memory default to a numeric 1 for truthiness is not sitting well with me.

It's quite simply how people use this:

https://github.com/search?q=__fish_git_prompt_show_informative_status&type=code

By a country mile, "1" is used to signify true.

It's also how *we* do it, it's how *we* set $fish_private_mode, $fish_term24bit, $fish_use_posix_spawn, $fish_handle_reflow and other [random](https://github.com/fish-shell/fish-shell/blob/acb77ad1a37aee15969ed80b840cf55642cbe650/share/functions/__fish_npm_helper.fish#L18) [helper](https://github.com/fish-shell/fish-shell/blob/acb77ad1a37aee15969ed80b840cf55642cbe650/share/functions/__fish_shared_key_bindings.fish#L216) variables.

Statistically, 1 has won. "true" is a distant second.

And I'm not sold that having multiple true values is helpful - like yes, we have a cheesy "[bool_from_string](https://github.com/fish-shell/fish-shell/blob/14d2a6d8ff18434a9672887d60009eef92ea21cc/src/wcstringutil.h#L165-L177)" helper that is technically used for some of these, but people don't use that capability, and having one true value makes the check easier.

This is the absolute, 100% correct check to test if $fish_handle_reflow is a true value:

```fish
test "$fish_handle_reflow" = 1
```

This is easy to get right (well, ignoring `test`s issues). On the other hand, this is the check to see if $fish_term24bit is set to true:

```fish
string match -ri '^[yt1]' -- $fish_term24bit
```

So these values are true: "yakshaving", "Tarpit", "1WRONG" and these are not: "ja", "waar", "oui".

mqudsi:
GitHub's default code search includes too many duplicates. [Using cs.github.com](https://cs.github.com/?q=__fish_git_prompt_show_informative_status&p=1&pt=bbebc12f2607dc8300fcafb29693e0bc4fb7bd916c7015ec941cfab9b4d8c4c020f84c21d3623947a562d369a4867d7c10deb0af415f7b5b51ca0aaf87460bdcc9fe667f466b4051c175de74b5f64bf70a46957d3a93060d0117894fef4165ed9b022f39f99fe5a6e66472841b5210c9b18a14b0dd151e4948dd15ac7ee06c664ca32c6e245d37ca1bd26e2f7387bb3c8670a62afb094658830caf57e3b0a6fa175718d5128a3f97d81b047a75e4bc5d5e05105882f599e46df462191f09ce6dd9e52927a89dd7baf83f9448d2c56a619786be021b9bb19d16b4e4a46222a867cf16a3ca7d1efcd5bc3ffcb5bb3afcb62a85bfec93025fd388ae0dca63aae721ce7fe5aca28cb68d860803179d74110602e6e53130876fff0873e34526cd3385f82bc61476aa1a476d8f9250a4e9351e0559dd4fcb2793fb992093329bba302fd3820bc6dd6d7971e872be8179f372985557659c468f94b0a8bf7d0fdd5dc5be0bc5b6994fb2f16cf8d01fafb21b023d9bbc66f59ba5faaa663338a970e05225c1448aec50f519b468b8125ec4caae0f) shows much more variety, including "yes" and "true" - and I didn't even use a regex to ignore the `1` cases.

faho:
>Using cs.github.com shows much more variety

Yeah, they want me to sign up for the waitlist, I'll just take your word for it that there are other cases.

In our code non-1 cases are a rarity.

Anyway: I still don't believe that making this more complicated helps. You tell people that "1" is true, and they'll use "1". Tell people that "true is one of true, yes, on, 1", they'll have to think about which one of them they want to use, and case-insensitive means you can get "yEs", which I see absolutely no reason for (we're not case-insensitive in our keywords or variable names or really a lot of other places) and which I would always want to correct, everywhere.

mqudsi:
I'm (also) saying people already have these set from a time before we told them what to set it to. Their configs will break.

The results show a bunch of themes that people clone/copy/download internally set "yes" or "true" for the prompt status variable. Anyone using these is going to have braeakage.

[The oh-my-fish theming guide](https://github.com/oh-my-fish/oh-my-fish/blob/master/docs/Themes.md) use `yes`, `true`, and `1` all over the place in their theming guide and tutorials.

The results show a whole bunch of these.

mqudsi:
It's nice that this clears up the earlier confusing statement about boolean variables vs git variables.

Issue #8384, opened 2021-10-26
For-loops and variable change events


faho:
While working on reducing the overhead of running for-loops, I stumbled upon one weird fact:

```fish
> function foo --on-variable foo; echo $argv $foo; end
> for foo in 1 2 3; end
VARIABLE SET FOO
VARIABLE SET FOO
VARIABLE SET FOO
VARIABLE SET FOO
```

See how that calls the event handler *four* times? That's because this called parser::set_var_and_fire to set the variable initially, which pins it to the outer scope.

That's probably useless.

However, I'm not sure the events here have a use in general. The reason is that for sets the variable as a *local* to the enclosing block, so no variable handler would ever be able to touch it?

Do we want to just remove the superfluous handler or the event firing here in general?

(or do we even want to abolish events for local variables entirely?)

Is there a use here I'm not seeing?

PR #9280, opened 2022-10-15
Allow erasing in multiple scopes in one go

### LINKED ISSUE #7711: Feature Request: 'set --erase' for multiple/all scopes
Hello!

I'm not in the practice of requesting features, but I feel this is one that is missing. It seems that un-setting a variable can be more difficult than it really should be, requiring a number of logic checks to ensure a setting is really gone. Could we have a:

```
set --ERASE foo
```
wherein, it actually un-sets the variable, regardless of scoping? I will even submit a pull request, if you agree. :-)

Thanks.

Additional details:

  fish --version
  fish, version 3.1.2

  Ubuntu, Groovy
  uname -a
  Linux xxxxxxxx 5.8.0-7642-generic #47~1612288990~20.10~b8113e7~dev-Ubuntu SMP Wed Feb 3 01:55:26 U x86_64 x86_64 x86_64 GNU/Linux

  echo $TERM
  xterm-256color

mqudsi:
While I was working on `fish_config` I found myself trying to execute `set -egl ...` a few times, and I couldn't come up with a good reason for why this wasn't allowed.

This patch extends `set -e` to allow unsetting a variable from multiple scopes in one go. (The docs and changelog would also need to be updated.)

Closes #7711.

### DIFF
--- a/src/builtins/set.cpp
+++ b/src/builtins/set.cpp
@@ -204,11 +204,14 @@ static int validate_cmd_opts(const wchar_t *cmd, const set_cmd_opts_t &opts, int
         return STATUS_INVALID_ARGS;
     }
 
-    // Variables can only have one scope.
+    // Variables can only have one scope...
     if (opts.local + opts.function + opts.global + opts.universal > 1) {
-        streams.err.append_format(BUILTIN_ERR_GLOCAL, cmd);
-        builtin_print_error_trailer(parser, streams.err, cmd);
-        return STATUS_INVALID_ARGS;
+        // ..unless we are erasing a variable, in which case we can erase from several in one go.
+        if (!opts.erase) {
+            streams.err.append_format(BUILTIN_ERR_GLOCAL, cmd);
+            builtin_print_error_trailer(parser, streams.err, cmd);
+            return STATUS_INVALID_ARGS;
+        }
     }
 
     // Variables can only have one export status.
@@ -391,7 +394,7 @@ static maybe_t<split_var_t> split_var_and_indexes(const wchar_t *arg, env_mode_f
     return res;
 }
 
-/// Given a list of values and 1-based indexes, return a new list, with those elements removed.
+/// Given a list of values and 1-based indexes, return a new list with those elements removed.
 /// Note this deliberately accepts both args by value, as it modifies them both.
 static wcstring_list_t erased_at_indexes(wcstring_list_t input, std::vector<long> indexes) {
     // Sort our indexes into *descending* order.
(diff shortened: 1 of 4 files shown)


src/builtins/set.cpp
```diff
@@ -620,45 +623,50 @@ static int builtin_set_show(const wchar_t *cmd, const set_cmd_opts_t &opts, int
...
-        auto split = split_var_and_indexes(argv[i], scope, parser.vars(), streams);
-        if (!split) {
-            builtin_print_error_trailer(parser, streams.err, cmd);
-            return STATUS_CMD_ERROR;
-        }
+    env_mode_flags_t scopes = compute_scope(opts);
+    // `set -e` is allowed to be called with multiple scopes.
+    for (int bit = 0; 1<<bit <= ENV_USER; ++bit) {
+        int scope = scopes & 1<<bit;
+        if (scope == 0 || (scope == ENV_USER && scopes != ENV_USER)) continue;
```
krobelus, line 630: >      int ret = STATUS_CMD_OK;
> -    env_mode_flags_t scope = compute_scope(opts);
> +    env_mode_flags_t scopes = compute_scope(opts);
> +    // `set -e` is allowed to be called with multiple scopes.
> +    for (int bit = 0; 1<<bit <= ENV_USER; ++bit) {
> +        int scope = scopes & 1<<bit;
> +        if (scope == 0 || (scope == ENV_USER && scopes != ENV_USER)) continue;

Supporting `set -elgU` sounds great!

faho:
Fixes #7711

src/builtins/set.cpp
```diff
@@ -620,45 +623,50 @@ static int builtin_set_show(const wchar_t *cmd, const set_cmd_opts_t &opts, int
...
+        for (int i = 0; i < argc; i++) {
+            auto split = split_var_and_indexes(argv[i], scope, parser.vars(), streams);
```
faho, line 632: I'm assuming with indexes this works like:

```fish
set -g foo 1 2 3
set -l foo a b c

set -egl foo[2]
```

makes the global foo "1 3" and the local foo "a c"?

faho: That is indeed how it works, and it's probably a good test case.

Variable events are also fired for each scope, which is probably okay for now.

(tbh I'm still of the opinion local variables shouldn't fire events - you can't read them in an event handler anyway - but last time that turned out to be unpopular)

mqudsi: Yes, what I had in mind the entire time was that `set -elg foo` should behave exactly as if `set -el foo; set -eg foo` (except for the return codes, which are already somewhat non-sensical and need to be fixed later), side effects and all. We can always change that behavior later if we have a good reason to, but I wanted to minimize the scope of the changes for this PR to make this an easier yes/no decision.

> (tbh I'm still of the opinion local variables shouldn't fire events - you can't read them in an event handler anyway - but last time that turned out to be unpopular)

I must have missed that discussion. Do you happen to remember what issue this was in?

mqudsi:
> Fixes #7711

When I was drafting the PR, I wrote "A team member recently mentioned in the chat that this either is possible or should be possible" but then I couldn't find it in the chat and was left confused where I had seen it. Thanks for linking it!

mqudsi:
Added tests and updated docs.

ridiculousfish:
Nice fix, LGTM! Please changelog before merging.

src/builtins/set.cpp
faho: > I must have missed that discussion. Do you happen to remember what issue this was in?

#8384, which asked about `for`-loop variables.

In that case, we chose to keep it because someone could try to do something with `--no-scope-shadowing`.

PR #9281, opened 2022-10-15
Document behavior of `set -f` outside a function?

mqudsi:
We should either document this behavior or emit an error (warning?) when function scope is explicitly requested but no function is executing. I'm leaning more towards the latter, but would like to hear others' thoughts.

### DIFF
--- a/doc_src/cmds/set.rst
+++ b/doc_src/cmds/set.rst
@@ -97,6 +97,7 @@ The scoping rules when creating or updating a variable are:
 
 - If a variable's scope is not explicitly set and there is no existing variable by that name, the variable will be local to the currently executing function. Note that this is different from using the ``-l`` or ``--local`` flag, in which case the variable will be local to the most-inner currently executing block, while without them the variable will be local to the function as a whole. If no function is executing, the variable will be set in the global scope.
 
+- Similarly, a variable explicitly set with ``-f`` or ``--function`` outside of any function (a top-level declaration or at the interactive prompt) will be assigned to the global scope instead.
 
 The exporting rules when creating or updating a variable are identical to the scoping rules for variables:
 


faho:
>emit an error (warning?) when function scope is explicitly requested

Please don't.

The idea behind `set -f` was that it was the option you'd reach for if you don't have a specific reason to pick something else.

It's not specifically "this scope has to be inside a function", it's something that fixes local scope, which should either not exist or be a special "block" scope.

That it is named "function" scope is simply in contrast to that.

Adding a warning just puts people off using something that is absolutely not a problem. It's *fine* to use it outside of a function. It has defined semantics and it's okay. There is no real need to warn people.

doc_src/cmds/set.rst
```diff
@@ -97,6 +97,7 @@ The scoping rules when creating or updating a variable are:
 
 - If a variable's scope is not explicitly set and there is no existing variable by that name, the variable will be local to the currently executing function. Note that this is different from using the ``-l`` or ``--local`` flag, in which case the variable will be local to the most-inner currently executing block, while without them the variable will be local to the function as a whole. If no function is executing, the variable will be set in the global scope.
 
+- Similarly, a variable explicitly set with ``-f`` or ``--function`` outside of any function (a top-level declaration or at the interactive prompt) will be assigned to the global scope instead.
```
faho, line 100: This is wrong. Function-scope outside of a function is simply the top-level local scope.

```fish
set -f foo bar

begin
    set -f foo bar2
end

set -S foo
```

will print:

```
$foo: set in local scope, unexported, with 1 elements
$foo[1]: |bar2|
```

faho (changes requested):

doc_src/cmds/set.rst
mqudsi: Hmm, I'm not sure what I did wrong but I tested this yesterday and *thought* I had confirmed the behavior before I even wrote the doc change. Thanks for the correction.

PR #9295, opened 2022-10-22
Add `nvme` completions

moverest:
This MR adds completions for the `nvme` Linux utility. It completes the sub-commands and some of their options.

### DIFF
--- a/share/completions/nvme.fish
+++ b/share/completions/nvme.fish
@@ -1,0 +1,141 @@
+# NVME user space tooling for Linux
+# See: https://github.com/linux-nvme/nvme-cli
+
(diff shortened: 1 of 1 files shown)


mqudsi:
Looks good. I like the jq-based default and the regular fallback.

Does `nvme` ever require sudo to provide the results?

moverest:
> Does nvme ever require sudo to provide the results?

Yes, it does need root permissions on my machine (Arch Linux). I'm not sure why though because `nvme list` does seem to find them (when looking at the file accesses it makes to `/sys` with `strace`). It just does not print them for some reason. :thinking:

mqudsi:
Ok, so does that mean the completions won't work because they execute `nvme` without sudo?

moverest:
There's a fallback with `string match -vr -- 'p[0-9]+$' /dev/nvme*`.

mqudsi:
Gotcha. Thanks for explaining.

Issue #8905, opened 2022-04-23
History is not updated when running multiple commands in the same line


liuml07:
### Environment

```
$ fish --version
fish, version 3.4.1
$ echo $version
3.4.1
$ uname -a
Darwin Mingliangs-MacBook-Pro-M1-Max.local 21.4.0 Darwin Kernel Version 21.4.0: Fri Mar 18 00:46:32 PDT 2022; root:xnu-8020.101.4~15/RELEASE_ARM64_T6000 arm64
$ echo $TERM
xterm-256color
```

### Problem

Running following commands, I got "true". I expect "false".

```
$ sh -c 'env HOME=$(mktemp -d) fish'
$ true
$ false; echo $history[1]
true
```

faho:
`false; echo $history[1]` is one entry in the history.

The history is updated per-commandline.

krobelus:
yeah I think this is the least surprising behavior.

If there is a concrete use  case, we have a mechanism for tracing every individual process (`set fish_trace 1`) that we could perhaps repurpose but it's probably not worth it.

liuml07:
Thanks for the context.

Is there a way I can get the last command (`false`) in the same command line?

faho:
There is no way to change this behavior so that `foo; bar` counts as two different history elements.

You haven't really explained what you want to *do* here, so we can't say any more.

liuml07:
Probably I'm asking a XY question.   This probably should be a user question instead of an issue.

I will get email after foo completes using command `foo; notify`. I now want to know the last command foo from "notify" fish function so the notification is for foo. Running `bar; notify` will be sending notification including bar.

"notify" includes `$status` which is for command "foo" but `$history[1]` is not foo.

I now learnt history is for per-commandline and it makes sense to me. Thanks

krobelus:
I just type `foo` <kbd>Enter</kbd> `notify` <kbd>Enter</kbd>.
A wrapper script (like `notify foo`) would be a cross-shell solution.
Another potential solution is a post-exec hook (`function --on-event fish_postexec`)

Though I think it's weird that there is no way to access the commandline *while* the command is running.

	$ sleep 3 && echo done, command: (status current-command) commandline: (commandline)
	done, command: sleep commandline:

so `status current-command` works but neither `commandline` nor `history` do.

Probably this is best exposed as `status current-commandline`.

liuml07:
Thanks for providing solutions. They work for me!

[...]

Issue #813, opened 2013-05-23
Rename $_


dag:
The name is inconsistent with the rest of fish and as far as I know not even consistent with other shells.  Not even that consistent with Perl's topic variable!

krader1961:
It is unfortunate this var was named `$_` given that in ksh and bash it refers to the last argument of the previous command rather than the command as in fish. Nonetheless, it is far too late to change this as doing so breaks backward compatibility for very little benefit.

It might be useful to introduce an alias. Which would allow us in the future, when we create the fish 3.0 release, to remove `$_` or change its definition. Does anyone have a recommendation for a more explicit var name?

mqudsi:
I've committed this in d367d57ae9527753e3564a77b757c8ccdbb12f02 with `$_` renamed to `$current_cmd`.

If there are any objections to the name or better suggestions, please chime in!

faho:
>If there are any objections to the name or better suggestions, please chime in!

I don't think removing it _right now_ is worth it. I'd rather have both names for a while, and deprecate the old one.

Also, I don't really like how many variables with generic names we're now using - `$current_cmd` is actually a name someone might have used, as is $hostname.

zanchey:
Please back this out. I agree with @faho entirely.

mqudsi:
Sure thing. If you'll read my commit logs, you'll see similar remarks to your own reservations. I feel `$_` should probably become `status --something`, and I also have huge reservations about using such common names (see commitlog for my `$hostname` implementation as well). There was already a very heated and acrimonious debate about `$version` and some commits one way and then the other, if I'm not mistaken, and I didn't want to get involved in that so I said that I'm just sticking to what we're currently using (no `$fish_` prefix) and we can change them or keep them as the team decides.

faho:
>debate about $version

Yes, there was. The main problem I could see is that $version is something you use to check which version of fish you are dealing with, so it's important that it stays consistent. And unfortunately it was named "$version" from the beginning.

We have a few more variables with generic names - $status, $argv - but those are front and center, something you use constantly. Others, like $PATH or $LANG, are standardized across shells. So there's a reasonable expectation you might know them.

This stuff - $pid, $current_command, $hostname - has names that might already be in use in user's scripts, or that users might want to use. And they're used much less often, so users might not know or might have forgotten about them.

--------------

So, what would I use?

For $pid, I'd use `$fish_pid` - because that makes it clear that it's actually the pid of fish.

[...]

PR #9296, opened 2022-10-22
Parser status vars

mqudsi:
* Removes deprecated `$_` and stores status variables directly in the parser.
* Adds `status current-commandline` for #8905.


src/parser.h
```diff
@@ -384,6 +393,17 @@ class parser_t : public std::enable_shared_from_this<parser_t> {
...
     void set_last_statuses(statuses_t s) { vars().set_last_statuses(std::move(s)); }
 
+    /// Get a parser status variable
+    const wcstring& get_status_var(parser_status_var_t var) const {
+        return status_vars_.at((int) var);
+    }
+
+    /// Set a parser status variable
+    void set_status_var(parser_status_var_t var, wcstring val) {
+        ASSERT_IS_MAIN_THREAD();
```
mqudsi, line 403: @ridiculousfish do you want me to remove this assert?

faho:
For clarity: `status current-command` was introduced in 3.0, released in December 2018 (#813).

I am not against removing $_ - it's a terrible name. However, it is still used - for example [a bunch of oh-my-fish title functions do](https://github.com/search?q=org%3Aoh-my-fish+fish_title&type=code) (for the record oh-my-fish still claims compatibility with fish 2.2.0, which I don't believe is helpful, but it is what it is).

And given that we might want to have 3.6.0 go out sooner rather than later, because we have already found an issue with the imminent new macOS version (53cb3a98fc5be524d6bdaaa7bf1e53e800534a09), I would *not* want to have this in the current cycle.

`status current-commandline` on its own I wouldn't have an issue with. I would like some tests.

mqudsi:
`status current-command` doesn’t currently have any tests because it doesn’t work in a non-interactive setting (though expect-based tests {s,w}hould work?)

I also don’t like the current `current-command` behavior; it isn’t the current command but the head of the current command line which (maybe) has nothing to do with current command. I tried writing tests for it but didn’t like codifying behavior I don’t like, lol. 

IMHO we should ship this and fix current-command together, having it return the *actual* command? Eg `true 1 && foo` where foo is `function foo; status current-command; end` would be nice to return `foo` and not `sleep` while status current-commandline would return the whole thing (as it does already in this PR).

Most status current-command usages are in bindings or completions where the behavior wouldn’t change, but typed out at the interactive prompt its behavior would change. There’s probably no one doing that outside of contrived demonstrations.

faho:
>status current-command doesn’t currently have any tests because it doesn’t work (the same) in a non-interactive setting (though expect-based tests {sh,w}ould work?)

I'm asking for tests for your new command, to be clear. I realize that sometimes we do things that are awkward to test, and I don't mean to require 100% tests for everything, but in this case it seems possible to do a pexpect test, and I'd like that.

>IMHO we should ship this and fix current-command together

I don't really have an opinion on current-command (i.e. the old one) at the moment.

I find it a bit awkward to change things together like this - it introduces a new command *and* changes how internals work *and* removes a deprecated variable. But if that's what's necessary, then sure.

But if the removal of `$_` can't be removed from this and then committed separately later I would not want it for 3.6.0.

mqudsi:
$_ can be added to this easily, I’ll do it later today if I get the chance.

mqudsi:
Re-added `$_` and added a pexpect test.

src/parser.h
ridiculousfish: yeah

src/parser.h
```diff
@@ -239,6 +239,12 @@ struct eval_res_t {
         : status(status), break_expand(break_expand), was_empty(was_empty), no_status(no_status) {}
 };
 
+enum class parser_status_var_t : uint8_t {
```
ridiculousfish, line 242: Unless we expect to have a lot more of these variables, I think we can do without this enum. In every call site we know which status variable we're accessing so we don't benefit from having these be indexable.

Also (as your changes illustrate) any time we want to set `current_command` we probably also want to set `current_commandline`. We can structure these so we're less likely to forget.

I suggest just making a struct:

    struct status_vars_t {
        wcstring command;
        wcstring commandline;
    };
    
and then you can set and get that struct. We don't even really need a setter or getter, just put it in `library_data_t`.

other than that this looks good to me

ridiculousfish (changes requested):
see comments

thomcc:
Totally random thing I coincidentally happened to noticed when playing around with the 3.6.0 changes: `status`'s completions doesn't include `current-commandline`, so perhaps it got missed in this PR? Sorry if this is already fixed elsewhere.

mqudsi:
Hey nice to bump into you here, @thomcc. You're right, the completions for this were overlooked. I'll add them and if we do a 3.6.1 we can backport the fix.

Issue #5924, opened 2019-06-06
Missing feature: excluding certain commands from being added to history


manfredlotz:
Unfortunately, there is no good possiblity to exclude certain commands from being added to history.  `bash` has `HISTIGNORE`. 

Could such a feature be added to fish? Of course, in a fish-like way instead of using a variable like in bash.

Thanks for considering.

zanchey:
This is a duplicate of #2788, which we have decided against.

One workaround is to abbreviate the command you always want to hide to begin with a space, which prevents it from being saved:

```
abbr --add nonsaved ' nonsaved'
```

manfredlotz:
That workaround is not bullet proof.
Let's say I type:  `nonsa<tab>` and after completion the abbr won't used.

I guess you will tell me not to type `nonsa<tab>`. But sometimes this happens automatically, and I think it has to work.

zanchey:
What version are you running? Abbreviations could not be tab-completed before 3.0.0.

manfredlotz:
3.0.2

manfredlotz:
Yesterday, I discussed things with Fabian, and he gave me a suggestion which I finalized like this:
```
function delete-mplayer --on-event fish_postexec
    string match -qr '^mplayer' -- $argv
    and history delete --exact --case-sensitive -- (string trim -r $argv)
end
```
This seems to work no matter if I tab-complete `mplayer` or type the whole command without completion

manfredlotz:
I have to trim when tab-completion, and I have to have the trim inside the history command because if there is nothing to trim there is a status 1

manfredlotz:
For the records. It seem that this works fine.

```
set -l pgms  mplayer my_secret_cmd

for p in $pgms 
  function $p-del-hist --on-event fish_postexec -d "delete $p from history" --inherit-variable p
    string match -qr "^$p" -- $argv
    and history delete --exact --case-sensitive -- (string trim -r $argv)
  end
end
```

faho:
>This is a duplicate of #2788, which we have decided against.

This is a re-run. I'd like to reconsider.

The workaround has a lot of problems (not the least of which that it now requires trimming whitespace, but we want to do some normalization on history), so I'd like to provide some thing to make this easier.

I don't think $HISTIGNORE is a great interface, but a function that we execute to ask whether an element should be added would be alright.

E.g.

```fish
function fish_should_add_history
    for cmd in mplayer cd ls
[...]

Issue #1383, opened 2014-03-29
"Up" key should recall previously entered command instead of previous command from history


develop7:
Repro:
1. execute `echo foo`
2. execute `<Space>echo bar` (prepend command with a space (` `) to prevent storing it in history)
3. press <kbd>&uarr;</kbd> key

Expected: `<Space>echo bar` is recalled

What happens instead: `echo foo` is recalled

This behavior is very convenient if a user wants to try various approaches to run command but doesn't want to flood history with non-working entries and/or entries containing private data.

develop7:
I've added another possible use case to issue description.

Also since core team is not sure if they want to see this implemented, I've got a question: would be accepted a pull request implementing this feature?

ridiculousfish:
zsh says:

> Note that the command lingers in the internal history until the next command is entered before it vanishes, allowing you to briefly reuse or edit the line. If you want to make it vanish right away without entering another command, type a space and press return.

So there's at most one volatile entry in zsh. That's an interesting compromise. What do you think?

develop7:
@ridiculousfish I'm perfectly ok with just one entry. Sorry for not being clear enough, I was sure I was.

pickfire:
What is the progress on this?

MhdAli:
+1

alphapapa:
I'm not sure this is a good idea.  When I prefix a command with a space, it's because I do not want the command recorded.  If this change were made, I would have to kill the fish process, or the command would remain in that shell's history.

At the least, this feature should be optional, and off by default.

develop7:
> If this change were made, I would have to kill the fish process, or the command would remain in that shell's history.

@alphapapa the whole point of this feature the space-prefixed command _won't_ hit history file and will be overwritten by next command

alphapapa:
@develop7 The point I'm making is that, when I prefix a command with a space, I don't want it to be recorded, period--neither on-disk nor in-memory.  If this change were made, the command would remain in the in-memory history list until another command is run or the shell is killed.

This is not only a potential privacy and security issue, but a potential data-loss issue as well.  A common reason to prefix commands with a space and prevent them from being recorded is to prevent potentially dangerous commands from being accidentally run a second time from the command history.

I'm sorry, but this is not a good idea.

develop7:
> I don't want it to be recorded, period--neither on-disk nor in-memory

[...]

PR #9298, opened 2022-10-23
fish_should_add_history function

BCNelson:
## Description

Adding a function `fish_should_add_history` to check if an item should be added to the history. I think that there need to be alot more discussion on how this feature should work. I would also welcome feedback on hope I implemented the feature.

Fixes issue #5924

## TODOs:
- [ x ] Changes to fish usage are reflected in user documentation/manpages.

### DIFF
--- a/doc_src/cmds/fish_should_add_history.rst
+++ b/doc_src/cmds/fish_should_add_history.rst
@@ -1,0 +1,45 @@
+.. _cmd-fish_should_add_history:
+
+fish_should_add_history - decide whether a command should be added to the history
+==============================================================
+
+Synopsis
+--------
+
+.. synopsis::
+
+    fish_should_add_history
+
+::
+
+  function fish_should_add_history
+      ...
+  end
+
+
+Description
+-----------
+
+The ``fish_should_add_history`` function is executed after every command that would normaly be saved to the disk. It is used to decide whether the command should be saved to the history or not.
+
+The function is executed with the command as its first argument. If the function returns a status of 0, the command is saved to the history. If the function returns a status of 255, the command is not saved to the history.
+
+The function can also return a status of 1 or 2, to save the command in ephemeral or memory mode respectively. This is useful for commands that should not be saved to the history file, but should be available for the up and down arrow keys.
+
+Example
+-------
+
+A simple example:
+
+
+
+::
+
+    function fish_should_add_history
+        for cmd in clear
+        string match -qr "^$cmd" -- $argv; and return 255
+        end
+        return 0
+    end
+
+
--- a/doc_src/commands.rst
+++ b/doc_src/commands.rst
@@ -60,6 +60,7 @@ Known functions are a customization point. You can change them to change how you
 - :doc:`fish_command_not_found <cmds/fish_command_not_found>` to tell fish what to do when a command is not found.
 - :doc:`fish_title <cmds/fish_title>` to change the terminal's title.
 - :doc:`fish_greeting <cmds/fish_greeting>` to show a greeting when fish starts.
+- :doc:`fish_should_add_history <cmd/fish_should_add_history>` to decide whether a command should be added to the history.
 
 Helper functions
 ^^^^^^^^^^^^^^^^
--- a/src/reader.cpp
+++ b/src/reader.cpp
@@ -105,6 +105,8 @@
 /// The name of the function for getting the input mode indicator.
 #define MODE_PROMPT_FUNCTION_NAME L"fish_mode_prompt"
 
+#define SHOULD_ADD_HISTORY_FUNCTION_NAME L"fish_should_add_history"
+
 /// The default title for the reader. This is used by reader_readline.
 #define DEFAULT_TITLE L"echo (status current-command) \" \" $PWD"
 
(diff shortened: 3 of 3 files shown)


faho:
I think the biggest question here is how many possibilities this should expose.

Here's what happens currently:

1. If you are running a normal shell, things are typically saved to disk (that's a mode of "disk")
2. If you have turned on private mode, things are "saved" to memory - so they are available in the current session but disappear after
3. If an entry starts with " ", it is an "ephemeral" entry

That last part was changed in #1383. Before, any ignored history entry was actually ignored - you would run ` echo foo` with the leading space, and it was gone *immediately*. There was absolutely no way to recall it, ever.

This was an issue if you made an error in such a command and wanted to run it again with fixes, so you would have to e.g. copy-paste it from the terminal into the terminal again.

So what we did was store it, in memory, until you run another command. So if you execute ` echo foo` and realize it should have been ` echo bar`, you can correct your mistake, but it's still gone.

And this PR here now allows you to both skip entirely, which has these downsides, and select the other kind of saving - so you could store things on disk even in private mode.

With that in mind, I would simplify this:

I would expose two possibilites: Stored as normal or ephemeral.

If `fish_should_add_history` returns 0, the entry is stored as normal, whether that's on disk or in memory (in private mode).

If it returns anything else, the entry simply becomes the ephemeral entry.

BCNelson:
> 1. If you are running a normal shell, things are typically saved to disk (that's a mode of "disk")
> 2. If you have turned on private mode, things are "saved" to memory - so they are available in the current session but disappear after
> 3. If an entry starts with " ", it is an "ephemeral" entry

That tracks with what I understood from the code

> And this PR here now allows you to both skip entirely, which has these downsides, and select the other kind of saving - so you could store things on disk even in private mode.

That is not entirely accurate. `fish_should_add_history` is only evoked when the history item would normally be saved to disk. If the command starts with `L" "` or we are in private mode the items history mode is `ephemeral` or `memory` respectively.

> With that in mind, I would simplify this:
> 
> I would expose two possibilites: Stored as normal or ephemeral.
> 
> If fish_should_add_history returns 0, the entry is stored as normal, whether that's on disk or in memory (in private mode).
> 
> If it returns anything else, the entry simply becomes the ephemeral entry.

I don't love these limitations because it removes one of my uses cases. That being after I run the `clear` to remove the large output of a previous command I don't need the clear command stored in history because the screen is cleared and a never need or want to run `clear` immediately after.

faho:
>That being after I run the clear to remove the large output of a previous command I don't need the clear command stored in history because the screen is cleared and a never need or want to run clear immediately after.

To be clear: The only limitation here would be that, right after you run `clear`, you could press up-arrow and it would show `clear`. As soon as you run another command that `clear` is gone.

The reason I don't want to expose commands being ignored entirely is that then we end up in the same situation as with space - people run a thing, notice that they want to run it again but now they can't.

And it's easy to put the fault on them, because they "could have made the right decision" - "why didn't you set it up correctly", but this is the sort of decision that's quite unclear at the time where it's made and is an undue burden to put on users.

Compare this to history size - that's also a popular knob in other shells, and it's also a decision users need to make long ahead of where it actually matters, because at some point you end up hitting that limit, and once you notice it and notice you chose wrong you would have to use your time machine to go back  to make the correct decision. So, we simply don't play, and we simply don't offer that option.

---

In your case could just press ctrl-l, which will also run `clear` but not via an actual commandline so it never enters history to begin with.

BCNelson:
> In your case could just press ctrl-l, which will also run clear but not via an actual commandline so it never enters history to begin with.
Fair enough

Another potential problem with my implementation is that the logic around when `fish_should_add_history` is run is complex. It can be summed up simply as whenever a history item would be persisted to disk it is run. However the simplicity of that statement is deceptive as the logic around saving to disk is not straight forward nor, as far as I know, document in one place.

An option that was brought up in [#5924](https://github.com/fish-shell/fish-shell/issues/5924#issuecomment-502176484) by @ridiculousfish is to have a default implementation of `fish_should_add_history` that does all of the logic around the saving of history. This does require branching on the return of `fish_should_add_history`.

BCNelson:
@ridiculousfish @faho What are your thoughts on the future of this PR?

Also I have found over the past month, ctrl-l has inconsistent behavior. I have tried in Konsole, vscodes Terminal and Gnome terminal sometimes it does clear the screen. However sometimes it moves the prompt to top of the screen but keeps the history so you can scroll to in still in the terminal. Do you have an other suggestions to get the desired behavior. If not I would still benefit from allowing multiple returns from the functions

faho:
>What are your thoughts on the future of this PR?

I still believe that the modes this exposes should be simplified.

It should not be possible to override private-mode with this (i.e. no forcing "save to disk").

I also remain unconvinced that it's wise to allow not saving something *at all* - experience has shown that that exacerbates mistakes and people end up regretting it. I don't want to include a footgun when the alternative is to minorly inconvenience some extremely specific use case (you'll see the temporary entry until you run something else).

So I'm still of the opinion that it should return 0 for "save however you would save" and "don't save except temporarily as the last entry".

Other than that, I'm okay with the idea. I think this the right way to allow this to be configured, instead of a cheesy variable.

---

>Also I have found over the past month, ctrl-l has inconsistent behavior.

This is probably down to the behavior of `clear`, which sends terminal-specific escape sequences.

Typically it sends one that means "clear the screen", and in some cases it will then *also* send "clear the scrollback buffer" - which is a fricking terrible decision, so we try to work around it. It appears that may be incomplete.

You can try `clear -x`, which, depending on your `clear` might skip the clearing of the scrollback, or you can try `bind \cl 'clear; commandline -f repaint'` without our hacks.

faho:
I'm going to close this for the time being as it has conflicts.

If anyone wants to work on it, be my guest.

PR #9307, opened 2022-10-26
Sort and deduplicate output of `complete -C`

### LINKED ISSUE #9288: complete(1) shows duplicates from symlinked paths
>  fish --version

`fish, version 3.5.1`

> uname -a

`Linux simplify 5.19.14_1 #1 SMP PREEMPT_DYNAMIC Tue Oct 11 15:42:18 UTC 2022 x86_64 GNU/Linux`

> echo $TERM

`xterm-256color`

> Please tell us if you tried fish without third-party customizations by executing this command and whether it affected the behavior you are reporting:
>  sh -c 'env HOME=$(mktemp -d) fish'

Yes, same results.

> Tell us how to reproduce the problem.

Here's a demo:
```fish
> complete -C test | head -n4
test	Check file types and compare values
test	Check file types and compare values
test	Check file types and compare values
test	Check file types and compare values
```

I tracked this down to my system being "usrmerged" so there are actually four symlinked paths that point to the same `test` binary:
````fish
> ls -i (which -a test)
656311 /bin/test  656311 /sbin/test  656311 /usr/bin/test  656311 /usr/sbin/test
````

Since `complete -C` strips paths and just shows the command and its description, it seems like it could dedupe output coming from binaries with the same inode.  (Of course, different paths can have the same filename and need different completion, so this only applies if they're the *same* command, and inode is an easy check.)

### PR DESCRIPTION
**Pull request by mqudsi**
This addresses a long-standing TODO where `complete -C` output isn't deduplicated.

With this patch, the same deduplication and sort procedure that is run on actual pager completions is also executed for `complete -C` completions (with a `-C` payload specified).

This makes it possible to use `complete -C` to test what completions will actually be generated by the completions pager instead of it displaying something completely divorced from reality, improving the productivity of fish completions developers.

Note that completions that wouldn't be shown in the pager are also omitted from the results, e.g. `test/buildroot/` and `test/fish_expand_test/` are omitted from the check matches in `checks/complete_directories.fish` because even if they were generated, the pager wouldn't have shown them. This again makes reasoning about and debugging completions much easier and more sane.

Closes #9288

---

Only remaining question is whether we should also sort and deduplicate when no `-C` is specified. I'm leaning towards no, mainly because it has a different purpose (it doesn't match the semantics of any pager completions in the first place, and I guess someone insane could actually be using it to see all the loaded completions globally, regardless of duplication or not.)

### DIFF
--- a/src/builtins/complete.cpp
+++ b/src/builtins/complete.cpp
@@ -388,11 +388,15 @@ maybe_t<int> builtin_complete(parser_t &parser, io_streams_t &streams, const wch
 
         // Prevent accidental recursion (see #6171).
         if (!parser.libdata().builtin_complete_current_commandline) {
-            if (!have_do_complete_param)
+            if (!have_do_complete_param) {
                 parser.libdata().builtin_complete_current_commandline = true;
+            }
 
             completion_list_t comp = complete(
                 do_complete_param, completion_request_options_t::normal(), parser.context());
+
+            // Apply the same sort and deduplication treatment as pager completions
+            completions_sort_and_prioritize(&comp);
 
             for (const auto &next : comp) {
                 // Make a fake commandline, and then apply the completion to it.
@@ -419,6 +423,8 @@ maybe_t<int> builtin_complete(parser_t &parser, io_streams_t &streams, const wch
 
                 // Append any description.
                 if (!next.description.empty()) {
+                    faux_cmdline_with_completion.reserve(
+                            faux_cmdline_with_completion.size() + 2 + next.description.size());
                     faux_cmdline_with_completion.push_back(L'\t');
                     faux_cmdline_with_completion.append(next.description);
                 }
--- a/tests/checks/complete-group-order.fish
+++ b/tests/checks/complete-group-order.fish
@@ -1,0 +1,32 @@
+#RUN: %fish -C 'set -l fish %fish' %s
+
+function fooc; true; end;
+
+# A non-`complete -k` completion
+complete -c fooc -fa "alpha delta bravo"
+
+# A `complete -k` completion chronologically and alphabetically before the next completion. You'd
+# expect it to come first, but we documented that it will come second.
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: src/builtins/complete.cpp [hunk 1]
```diff
@@ -394,6 +394,10 @@ maybe_t<int> builtin_complete(parser_t &parser, io_streams_t &streams, const wch
             completion_list_t comp = complete(
                 do_complete_param, completion_request_options_t::normal(), parser.context());
 
+            if (have_do_complete_param) {
```
faho, line 397: "have_do_complete_param" means that we're not using the current commandline.

It's this:

```fish
complete -C"git checkout "
# vs
complete -C # <- this uses the current commandline, the argument is optional!
```

Notably it doesn't apply to `complete` (which lists all defined completions - `complete -c foo -l foo-option1` and so on).

If we are in this branch, we already know we are in "do_complete" mode, i.e. we are running completions to return completion candidates.

I'm not entirely sure why we wouldn't run the same logic on the current commandline?

mqudsi: Oh, I see. In that case, I should revisit it and remove that precondition.

faho:
Looks good to me!

krobelus:
I'm not sure if we should do this because it breaks the order (in situations where `--keep-order` is used):

```fish
complete foo -xa '3 2 1'
complete bar -xa '(complete -C "foo ")' --keep-order
```

faho:
>I'm not sure if we should do this because it breaks the order (in situations where --keep-order is used):

It breaks the order because it makes the order the same as the wrapped completion wants.

Or am I missing something?

Do you have an example where this changes things?

krobelus:
Oh I see, this is fine then.

mqudsi:
Yes, I have another test like the pexpect one I just added but for `complete -C` I'll check in along with this commit that asserts that `-k` group sort order is correct.

PR #9311, opened 2022-10-29
fix lsblk column completion

llenck:
## Description

The current completion of `lsblk -o` has been broken since commit [c3a4cfc](https://github.com/util-linux/util-linux/commit/c3a4cfc5795da0a5f7eb2753aef3f543d12c8341) of util-linux. This PR changes the completion so it works with the current help format.

## TODOs:
- n/a Changes to fish usage are reflected in user documentation/manpages.
- n/a Tests have been added for regressions fixed

### DIFF
--- a/share/completions/lsblk.fish
+++ b/share/completions/lsblk.fish
@@ -1,6 +1,5 @@
 function __fish_print_lsblk_columns --description 'Print available lsblk columns'
-    lsblk --help | sed '1,/Available columns:/d; /^$/,$d; s/^\s\+//; s/\s/\t/'
-
+    LC_ALL=C lsblk --help | sed '1,/Available .*columns:/d; /^$/,$d; s/^\s\+//; s/\s/\t/'
 end
 
 complete -c lsblk -s a -l all -d "print all devices"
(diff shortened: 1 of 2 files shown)


share/completions/lsblk.fish
```diff
@@ -1,5 +1,5 @@
 function __fish_print_lsblk_columns --description 'Print available lsblk columns'
-    lsblk --help | sed '1,/Available columns:/d; /^$/,$d; s/^\s\+//; s/\s/\t/'
+    lsblk --help | sed '1,/Available output columns:/d; /^$/,$d; s/^\s\+//; s/\s/\t/'
```
faho, line 2: This should be optional, for people using older versions.

Also, it's localized, so this won't actually work for people not using an english locale - this needs to be `LC_ALL=C lsblk --help`.

I'd probably make it

```fish
LC_ALL=C lsblk --help | sed '1,/Available .*columns:/d; /^$/,$d; s/^\s\+//; s/\s/\t/'
```

llenck: Thanks; I will apply this.

faho:
Thanks!

PR #9312, opened 2022-10-29
Add completions for loadkeys

llenck:
## Description

Adds completion for loadkeys.

## TODOs:

- n/a Changes to fish usage are reflected in user documentation/manpages.
- n/a Tests have been added for regressions fixed

### DIFF
--- a/share/completions/loadkeys.fish
+++ b/share/completions/loadkeys.fish
@@ -1,0 +1,26 @@
+function __fish_print_console_keymaps
+    # The path(s) used may differ by linux distribution, and are compiled into
+    # loadkeys, which doesn't provide a way to check which paths are searched
+    # or which keymaps it can find. localectl can list keymaps, but is part of
+    # systemd, which we shouldn't depend on.
+
+    set -l dirs '/usr/share/kbd/keymaps' '/usr/share/keymaps' '/usr/lib/kbd/keymaps' '/lib/kbd/keymaps' '/usr/src/linux/drivers'
+
+	path filter -f $dirs/** | string replace -rf '.*/(.*)\.k?map(|\..*)$' '$1'
+end
+
+complete -fc loadkeys -a "(__fish_print_console_keymaps)"
+
+complete -c loadkeys -s C -l console -d "Console device to use" -x -a "(path filter --type=char /dev/tty*)"
+complete -c loadkeys -s a -l ascii -d "Force conversion to ASCII"
+complete -c loadkeys -s b -l bkeymap -d "Output binary keymap"
+complete -c loadkeys -s c -l clearcompose -d "Clear kernel compose table"
+complete -c loadkeys -s d -l default -d "Load default keymap"
+complete -c loadkeys -s m -l mktable -d "Output a 'defkeymap.c'"
+complete -c loadkeys -s p -l parse -d "Only search and parse keymap"
+complete -c loadkeys -s s -l clearstrings -d "Clear kernel string table"
+complete -c loadkeys -s u -l unicode -d "Force conversion to Unicode"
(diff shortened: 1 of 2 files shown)


share/completions/loadkeys.fish
```diff
@@ -0,0 +1,15 @@
+complete -fc loadkeys -a "(find /usr/share/kbd/keymaps/ -type f | string replace -rf '.*/(.*).map.gz' '\$1')"
+
+complete -c loadkeys -s C -l console -d "Console device to use" -x -a "(command ls /dev/tty*)"
```
faho, line 3: Use `path filter --type=char /dev/tty*` instead of `command ls` (which might have portability issues, will print directories, requires starting another command, and will print an error if no files match).

share/completions/loadkeys.fish
```diff
@@ -0,0 +1,15 @@
+complete -fc loadkeys -a "(find /usr/share/kbd/keymaps/ -type f | string replace -rf '.*/(.*).map.gz' '\$1')"
```
faho, line 1: I'm not sure these *need* to be .gz, so I'd probably make that second extension optional, maybe even up to

```fish
string replace -rf '.*/(.*).map(\..*)' '\$1'
```

llenck: You're right, and I also just found out that `/usr/share/kbd/keymaps` isn't as standardized as the manual I read made it seem. I will update this later.

llenck: I've updated the PR to work with all paths I could find for different distributions, and all extensions that `loadkeys` looks for, according to its source code.

share/completions/loadkeys.fish
```diff
@@ -0,0 +1,26 @@
+function __fish_print_console_keymaps
+    # The path(s) used may differ by linux distribution, and are compiled into
+    # loadkeys, which doesn't provide a way to check which paths are searched
+    # or which keymaps it can find. localectl can list keymaps, but is part of
+    # systemd, which we shouldn't depend on.
+
+    for d in '/usr/share/kbd/keymaps' '/usr/share/keymaps' '/usr/lib/kbd/keymaps' '/lib/kbd/keymaps' '/usr/src/linux/drivers'
+        [ -d $d ] && find $d -type f | string replace -rf '.*/(.*)\.k?map(|\.gz|\.bz2|\.xz|\.zst)$' '$1'
```
faho, line 8: This can be simplified somewhat:

```fish
set -l dirs /usr/share/kbd/keymaps /usr/share/keymaps /usr/lib/kbd/keymaps /lib/kbd/keymaps /usr/src/linux/drivers
path filter -f $dirs/** | string replace -rf '.*/(.*)\.k?map(|\.gz|\.bz2|\.xz|\.zst)$' '$1'
```

No need to run `find` five times or test for the existence of the directories separately. You can just glob in them and keep the files.


---

Another thing: I wouldn't want to hardcode the compression extension here - I'd just look for a `\.k?map` and throw away any extension after that, so:

```fish
string replace -rf '.*/(.*)\.k?map(|\..*)$' '$1'
```

Now, technically this could be mis-matching something if there was ever a file with a different extension there - `foo.map.this_is_not_a_map`. However, I would call that *less* likely than someone using a compression algorithm you don't know here.

llenck: I doubt supported compression schemes are going to change soon, and there are a bunch of weird extensions in `/usr/src/linux`, though I guess even if a false positive triggering file was added, it wouldn't be that harmful; I'll apply that.

Also sorry, I didn't know `path` had special behaviour for failed wildcard matches, so I didn't want to use that, as e.g. Ubuntu's `loadkeys` comes without any keymaps. I'll apply that as well.

faho:
Merged, thanks!

PR #9325, opened 2022-11-02
completions: added systemd-cryptenroll

exploide:
## Description

Added completions for `systemd-cryptenroll` which allows to configure LUKS2 key slots with FIDO2/PKCS11/TPM2 devices for key unlock.

## TODOs:

### DIFF
--- a/share/completions/systemd-cryptenroll.fish
+++ b/share/completions/systemd-cryptenroll.fish
@@ -1,0 +1,45 @@
+function __fish_cryptenroll_pkcs11_devices
+    systemd-cryptenroll --pkcs11-token-uri=list 2>/dev/null | tail -n +2 | string replace -r '(\S+)\s(\S+).*' '$1\t$2'
+    echo -e "auto\tautomatically determine URI of security token"
+    echo -e "list\tlist available PKCS#11 tokens"
+end
+
+function __fish_cryptenroll_fido2_devices
+    systemd-cryptenroll --fido2-device=list 2>/dev/null | tail -n +2 | string replace -ar ' +' ' ' | string replace -r '(\S+)\s(.*)' '$1\t$2'
+    echo -e "auto\tautomatically determine device node of security token"
+    echo -e "list\tlist available FIDO2 tokens"
+end
+
+function __fish_cryptenroll_tpm2_devices
+    systemd-cryptenroll --tpm2-device=list 2>/dev/null | tail -n +2 | string replace -r '(\S+)\s(\S+).*' '$1\t$2'
+    echo -e "auto\tautomatically determine device node of TPM2 device"
+    echo -e "list\tlist available TPM2 devices"
+end
+
+function __fish_cryptenroll_complete_wipe
+    echo -e "all\twipe all key slots"
+    echo -e "empty\twipe key slots unlocked by an empty passphrase"
+    echo -e "password\twipe key slots unlocked by a traditional passphrase"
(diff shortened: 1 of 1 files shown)


mqudsi (commented):
Thanks for submitting this completion. 

Are there any subcommands that require `sudo`? Does enumerating devices ever required `sudo`?

exploide:
The final command should need to be run with `sudo` I guess. But the enumeration was done without elevated privileges. This should work.

share/completions/systemd-cryptenroll.fish
```diff
@@ -0,0 +1,45 @@
...
+complete -c systemd-cryptenroll -xa "(__fish_complete_blockdevice)"
+
+complete -c systemd-cryptenroll -l password -d "Enroll a regular password"
+complete -c systemd-cryptenroll -l recovery-key -d "Enroll an auto-generated recovery key"
+complete -c systemd-cryptenroll -l pkcs11-token-uri -xa "(__fish_cryptenroll_pkcs11_devices)" -d "Enroll a PKCS#11 security token or smartcard"
+complete -c systemd-cryptenroll -l fido2-device -xa "(__fish_cryptenroll_fido2_devices)" -d "Enroll a FIDO2 security token"
+complete -c systemd-cryptenroll -l fido2-with-client-pin -xa "yes no" -d "Require to enter a PIN when unlocking the volume"
+complete -c systemd-cryptenroll -l fido2-with-user-presence -xa "yes no" -d "Require to verify presence when unlocking the volume"
+complete -c systemd-cryptenroll -l fido2-with-user-verification -xa "yes no" -d "Require user verification when unlocking the volume"
+complete -c systemd-cryptenroll -l tpm2-device -xa "(__fish_cryptenroll_tpm2_devices)" -d "Enroll a TPM2 security chip"
```
mqudsi, line 41: e.g. using `-kxa` instead of `-xa` here.

mqudsi (changes requested):
That's good. I know some other hardware enumeration completions require sudo for the completions themselves which is harder to handle.

Question for you:

In all the `__fish_cryptenroll_xxxx_devices` functions, you have an actual list of enumerated devices then two hard-coded options (`auto` and `list`). Currently they'll be munged in with the actual device names and all of them displayed in alphabetical order. I think you would prefer to use `-k` (see the annotated example) so that these two lines are shown after all the actual device name completions?

mqudsi:
I see that you added `-kxa` everywhere. The only question is if the output of `systemd-cryptenroll --xxxx-device=list` is already sorted (by some strict order even if not alphabetic) or if you need to sort it yourself (`| sort`) now that fish's pager won't do the sort for you.

exploide:
> I think you would prefer to use -k (see the annotated example) so that these two lines are shown after all the actual device name completions?

I don't really care whether the two default keywords are sorted before or after the devices and in most cases you have only one device available anyway. However it would be odd if it gets sorted in between the default keywords, so I added the `-k` option as requested. Though this probably wouldn't have happened because the devices usually start with `/dev` and `/` comes alphabetically before `auto`.

> I see that you added -kxa everywhere. The only question is if the output of systemd-cryptenroll --xxxx-device=list is already sorted (by some strict order even if not alphabetic) or if you need to sort it yourself (| sort) now that fish's pager won't do the sort for you.

I don't know as I don't have more than one of each respective device class. And you typically don't have many of them plugged in simultaneously anyway.

mqudsi:
That makes sense to me. I'll merge this, thanks for the quick revisions and sorry for the review delay.

PR #9326, opened 2022-11-03
Make Fish Shell REUSE-compliant

silverhook:
# Intro

No-one asked me to do this, but since I’m using Fish quite heavily at work, which also involves using and applying [REUSE](https://reuse.software) best practices for licensing and copyright information, I wanted to contribute back to Fish by helping make its codebase REUSE-compatible as well.

If someone finds this useful, I’m happy to help make this happen. But I am not intimately aware of Fish’s history, so it would be very prudent this PR is thoroughly checked for any errors.

# How I added the info with `reuse`

At the start I used [ScanCode](https://nexb.com/scancode/) on a freshly cloned repository, to find any copyright and license notices that already existed in the code. We’ll look at that at the end ;)

The first thing I did was to run the following shell script using the `reuse` helper tool from <https://reuse.software>

I got the licensing and copyright notices from the current [`doc_src_license.rst`](https://github.com/fish-shell/fish-shell/blob/5ee389592cea90e123927aa657b8ad8b790660a3/doc_src/license.rst).

The year I stored only the year of the first commit of that file [and there are good reasons for that](https://matija.suklje.name/how-and-why-to-properly-write-copyright-statements-in-your-code#why-not-bump-the-year-on-change).

```shell
for file in (git ls-files)
    reuse addheader --skip-unrecognised --license GPL-2.0-only $file
    set --local year (git log --follow --format=%as $file | tail -n1 | cut -c-4)
    if test $year -le 2009
        set --local copyright "Axel Liljencrantz"
        reuse addheader --skip-unrecognised --copyright-style spdx-symbol --copyright $copyright --year $year $file
    end
    if test $year -ge 2009
        set --local copyright "fish-shell contributors"
        reuse addheader --skip-unrecognised --copyright-style spdx-symbol --copyright $copyright --year $year $file
    end
end
```

As that still left me with a handful of files that `reuse` could not guess which comment style to use safely, for files I was pretty sure it’s safe to force it, I did so with the following script (replace `……………` with a file or glob):

```shell
for file in ……………
    reuse addheader --license GPL-2.0-only --style python $file
    set --local year (git log --follow --format=%as $file | tail -n1 | cut -c-4)
    if test $year -le 2009
        set --local copyright "Axel Liljencrantz"
        reuse addheader --copyright-style spdx-symbol --copyright $copyright --year $year --style python $file
    end
    if test $year -ge 2009
        set --local copyright "fish-shell contributors"
        reuse addheader --copyright-style spdx-symbol --copyright $copyright --year $year --style python $file
    end
end
```
I ran the above on the following files/globs:

- `.clang*`
- `**.Dockerfile`
- `.oclint`
- `*makefile`
- `**.imp`
- `build_tools/osx_package_scripts/*`
- `fish.desktop`
- `fish.spec.in`
- `share/lynx.lss`
- `build_tools/cppcheck.suppressions` – using `--style c` instead
- `config_cmake.h.in` – using `--style c` instead
- `**.m` – using `--style c` instead
- `src/ast_node_types.inc` – using `--style c` instead
- `src/fallback.cpp` – using `--style c` instead

That still left me with a handful of files that I either knew were not safe to modify or was not sure, so I used the `*.license` option. What this does is instead of modifying `{$filename}`, store the SPDX tags in `{$basename}.license`.

```shell
for file in $argv
    reuse addheader --force-dot-license --license GPL-2.0-only $file
    set --local year (git log --follow --format=%as $file | tail -n1 | cut -c-4)
    if test $year -le 2009
        set --local copyright "Axel Liljencrantz"
        reuse addheader --force-dot-license --copyright-style spdx-symbol --copyright $copyright --year $year $file
    end
    if test $year -ge 2009
        set --local copyright "fish-shell contributors"
        reuse addheader --force-dot-license --copyright-style spdx-symbol --copyright $copyright --year $year $file
    end
end
```

I ran the above on the following files/globs:

- `**.theme` – did not want to mess with the comments, just in case; if it’s fine to add them, it would be better to have the data in the file’s headers instead of the `**.theme.license` files.
- `build_tools/cppcheck.rules`
- `osx/*.entitlements`
- `debian/**`
- `fish.pc.in`
- `share/__fish_build_paths.fish.in`
- `share/groff/fish.tmac`
- `tests/history_sample_bash`
- `tests/history_sample_corrupt1`
- `tests/history_sample_fish_1_x`
- `tests/history_sample_fish_2_0`
- `tests/interactive.config`
- `osx/CMakeMacAppInfo.plist.in`

Ultimately, I had to do `share/completions/rsync.fish`, `src/fallback.cpp`, `share/completions/grunt.fish`, `src/tinyexpr.*`, `share/completions/pandoc.fish` by hand, due to a [bug](https://github.com/fsfe/reuse-tool/issues/621) I just found in `reuse`.

Then I ran `reuse download -all` to download all missing licenses.

# Double-checking with `scancode`

Finally I went through the files that `scancode` identified as already having licensing and copyright information and cleaned those up:

- `share/tools/web_config/js/angular-route.js`, `share/tools/web_config/js/angular-sanitize.js`, `share/tools/web_config/js/angular.js`
- `doc_src/python_docs_theme/globaltoc.html`,  `doc_src/python_docs_theme/localtoc.html`

I don’t know where the Samba team’s snippet begins and ends in `share/completions/rsync.fish` and NetBSD’s in `src/fallback.cpp` but if someone knows, you can use the [SPDX Snippet Tags](https://spdx.github.io/spdx-spec/v2.3/file-tags/#h3-snippet-tags-format) to mark those. Similar issue is with `src/env.cpp`, `share/completions/grunt.fish`, `src/tinyexpr.*`, `share/completions/pandoc.fish`, `share/functions/fish_git_prompt.fish` and a few files in `po/*.po`.

`share/tools/create_manpage_completions.py` seems to be orginally under _BSD-2-Clause_ and copyright by _Siteshwar Vashisht_, someone should double-check that.

`src/tinyexpr.*` seems to be originally under `Zlib_ license and copyright by _Lewis Van Winkle_, someone should double-check that.

`src/wgetopt.*` seems to be originally under _LGPL-2.0-or-later_, someone should double-check that.

`src/builtins/printf.cpp` seems to be originally under _GPL-2.0-or-later` and copyright by _FSF_, someone should double-check that.

`cmake/CheckIncludeFiles.cmake` seems to be originally under _BSD-3-Clause_, someone should double-check that.

I suspect `src/widecharwidth/widechar_width.h` is supposed to be licensed under _CC0-1.0_, someone should double-check that.

`src/utf8.*` seems to be originally under _ISC_ license and copyright by _Alexey Vatchenko_, someone should double-check that.


silverhook:
For the vast majority of the files this should be fine as is. The only question here would be if we should also add `fish-shell contributors` as copyright holders to files that were created before 2009, but later modified – this is a use case the current [doc_src_license.rst](https://github.com/fish-shell/fish-shell/blob/5ee389592cea90e123927aa657b8ad8b790660a3/doc_src/license.rst) does not cover, so I skipped it.

Where some care should be taken is source code that was copied from elsewhere and is as such has other copyright holders and licenses, as well as code that includes authorship or copyright notices either in snippets or files themselves.

Again, happy to help with this and polish the PR to an acceptable form, if it is in the interest of the community/maintainers for Fish to carry [proper licensing info based on community best practices](https://reuse.software). If this is not desired though, I withdraw my well-intended PR and apologise for any inconvenience it might have caused.

BTW, [don’t bump the year in copyright statements](https://matija.suklje.name/how-and-why-to-properly-write-copyright-statements-in-your-code#why-not-bump-the-year-on-change), there is no good reason for that.

build_tools/littlecheck.py
```diff
@@ -1,5 +1,9 @@
 #!/usr/bin/env python
 
+# SPDX-FileCopyrightText: © 2019 fish-shell contributors
+#
+# SPDX-License-Identifier: GPL-2.0-only
```
faho, line 5: This is incorrect - littlecheck is imported [from here](https://github.com/ridiculousfish/littlecheck) and licensed as "public domain" (the intention there obviously being "do whatever you want with it"), by us.

faho: And since it's vendored from another repo, the typical way of updating it is the copy the file over, so this would be broken again.

CODE_OF_CONDUCT.md
```diff
@@ -1,3 +1,8 @@
+<!--
+SPDX-FileCopyrightText: © 2020 fish-shell contributors
+
+SPDX-License-Identifier: GPL-2.0-only
```
faho, line 4: This is the contributor covenant, copied from https://www.contributor-covenant.org/version/2/1/code_of_conduct/ (I'm not sure which version).

It carries a CC-BY 4.0 license

faho:
Can I just say that I would *much* prefer being able to leave the license in a centralized place?

I hate how this spec asks you to add boilerplate to every file.

Specifically, we get a *lot* of drive-by commits that just add a completion script, and now we would have to ask people to add it there as well. (similar problems occur with build systems that don't like globbing files - we would have to ask someone who only wants to write fish script and knows a particular tool to modify the build system)

faho:
>src/tinyexpr.* seems to be originally under `Zlib_ license and copyright by Lewis Van Winkle, someone should double-check that.

Correct.

> src/wgetopt.* seems to be originally under LGPL-2.0-or-later, someone should double-check that.

Pretty sure it's nicked from old glibc.

> src/builtins/printf.cpp seems to be originally under _GPL-2.0-or-later` and copyright by FSF, someone should double-check that.

Definitely grabbed from old GNU stuff. I'm not sure if this also applies to src/builtins/test.cpp.

>I suspect src/widecharwidth/widechar_width.h is supposed to be licensed under CC0-1.0, someone should double-check that.

Widechar_width.h is generated from the unicode data files via https://github.com/ridiculousfish/widecharwidth, which is also one of ours. License there is "public domain or CC0 if that doesn't work".

>share/tools/create_manpage_completions.py seems to be orginally under BSD-2-Clause and copyright by Siteshwar Vashisht, someone should double-check that.

That's @siteshwar, an inactive fish developer who I'm pretty sure added that himself. No idea why that script has a different license but if it says BSD-2-clause I have no reason to doubt that.

build_tools/littlecheck.py
silverhook: Got it.

There’s three practical solutions I see:
1. make that repository REUSE compliant as well, so the file would already carry the right license and copyright info – optimal, but more work
2. move the (right) SPDX/REUSE tags into a new file `build_tools/littlecheck.py.license`, so that info would apply to the file even if it’s later updated – has its downsides, but might be a good compromise here
3. add the right info on every update of the file – annoying

CODE_OF_CONDUCT.md
silverhook: Then perhaps something like this instead:

```suggestion
SPDX-License-Identifier: CC-BY-4.0
```

CODE_OF_CONDUCT.md
```diff
@@ -1,3 +1,8 @@
+<!--
+SPDX-FileCopyrightText: © 2020 fish-shell contributors
```
silverhook, line 2: … and  then also change this

```suggestion
SPDX-FileCopyrightText: © 2014 Coraline Ada Ehmke
```

silverhook:
> I think it's a good idea to clarify our licenses, and having a machine-readable format is great, but I would _much_ prefer being able to leave the license in a centralized place.

I understand your reservation to it, since it is a massive chore (at least initially), but there are really good reasons for why REUSE by default requires the tags in the source code files themselves, if at all possible.

As we saw in this discussion here before, it is quite common that a file is moved out of its original repository, and then it becomes an exercise in community memory and digital archeology to either figure out what the origin and ©/licensing reality of that file really is. The only reliable way to make sure such metadata moves together with the files, is to keep it in the files themselves.
 
> Specifically, we get a _lot_ of drive-by commits that just add a completion script, and now we would have to ask people to add it there as well. (similar problems occur with build systems that don't like globbing files - we would have to ask someone who only wants to write fish script and knows a particular tool to modify the build system)

I see. That could, of course, provide a tiny hurdle, but nothing we can’t make easier or automate completely.

How about setting up a git hook that would check if a new file includes SPDX/REUSE tags, and if not just add them by using Fish’s default license, the year the commit happened and as the copyright holder the standard Fish notice, as is pretty much everywhere?

silverhook:
> > src/wgetopt.* seems to be originally under LGPL-2.0-or-later, someone should double-check that.
> 
> Pretty sure it's nicked from old glibc.

Then probably the most correct version would be (if the copyright notice is correct in the file):

```
// SPDX-FileCopyrightText: Copyright (C) 1987, 88, 89, 90, 91, 92, 93, 94 Free Software Foundation, Inc.
// SPDX-FileCopyrightText: © 2005 Axel Liljencrantz
// SPDX-FileCopyrighTtext: © 2009 fish-shell contributors
//
// SPDX-License-Identifier: LGPL-2.0-or-later
```
 
> > src/builtins/printf.cpp seems to be originally under _GPL-2.0-or-later` and copyright by FSF, someone should double-check that.
> 
> Definitely grabbed from old GNU stuff. I'm not sure if this also applies to src/builtins/test.cpp.

Any way to check?

> > I suspect src/widecharwidth/widechar_width.h is supposed to be licensed under CC0-1.0, someone should double-check that.
> 
> Widechar_width.h is generated from the unicode data files via https://github.com/ridiculousfish/widecharwidth, which is also one of ours. License there is "public domain or CC0 if that doesn't work".

Then it’s CC0-1.0, since you don’t want to rely on “public domain” as in general it doesn’t work (can dive deeper into why and how, if you’re really interested).

BTW, if you can, you might also want to change CC0 to something else like 0BSD or MIT0, to avoid [issues with Fedora](https://hackaday.com/2022/07/29/why-fedora-decided-to-give-cc0-licensed-code-the-boot/) and others who decided similarly due to the potential patent trap.
 
> > share/tools/create_manpage_completions.py seems to be orginally under BSD-2-Clause and copyright by Siteshwar Vashisht, someone should double-check that.
> 
> That's @siteshwar, an inactive fish developer who I'm pretty sure added that himself. No idea why that script has a different license but if it says BSD-2-clause I have no reason to doubt that.

In that case that would probably best be:

```
# SPDX-FileCopyrightText: © 2012 Siteshwar Vashisht
# SPDX-FileCopyrightText: © 2012 fish-shell contributors
#
# SPDX-License-Identifier: BSD-2-Clause
```

silverhook:
Thanks for looking into this. I half-expected this PR to be just shown the door right away, as it introduces changes that have no purely functional gain, just on a meta- and legal/copyright/FOSS level :)

siteshwar:
> > src/builtins/printf.cpp seems to be originally under _GPL-2.0-or-later` and copyright by FSF, someone should double-check that.

Right. It was me who had done the initial implementation of `printf` based on `coreutils`.

> > share/tools/create_manpage_completions.py seems to be orginally under BSD-2-Clause and copyright by Siteshwar Vashisht, someone should double-check that.
> 
> That's @siteshwar, an inactive fish developer who I'm pretty sure added that himself. No idea why that script has a different license but if it says BSD-2-clause I have no reason to doubt that.

I originally wrote that script and I added `BSD` license after discussing licensing with @ridiculousfish.

build_tools/littlecheck.py
silverhook: I created https://github.com/ridiculousfish/littlecheck/pull/11 in order to tackle this issue with the option 1) solution.

If that is not desirable, I’ll change it accordingly.

silverhook:
I made the changes to the 3rd party software licensing & copyright info as discussed here.

In addition, I went ahead and added the `fish-shell contributors` copyright statement to all files that were originally contributed by `Axel Liljencrantz`, but also modified on or after 2009 by the Fish community:

```shell
for file in (grep --recursive --regexp "SPDX-FileCopyrightText: .* Axel Liljencrantz" | cut --delimiter ":" --fields 1)
    set --local year (git log --follow --format=%as $file | head -n1 | cut -c-4)
    if test $year -ge 2009
        set --local copyright "fish-shell contributors"
        reuse addheader --copyright-style spdx-symbol --copyright $copyright --year 2009 $file
    end
end
```

From what I can tell, this should cover it all now, and reflect all the copyright and licensing info for Fish that I am aware of.

A great thing about REUSE-compliant repos is that you can very easily see which all licenses apply to it, without even needing any tools (although some tools are cool too). For example, to see all licenses that are mentioned in the repo, simply `ls LICENSES/` and you see [SPDX short names](https://spdx.org/licenses/) of all the licenses that apply and have all the license texts in one folder too :)

krobelus:
I don't like adding headers to every file without a strong reason because it adds noise and unnecesssary overhead to creating new files.
A top-level COPYING file is obviously the default license for any file without the header, does REUSE not recognize this?
Files with other licenses should have the header I guess.
For simplicity I'd relicense as much as possible under the same license.

mqudsi:
I'm uncomfortable with all the churn here and the explicit GPL2 assertions where they don't really make sense. I guess the latter is on us and we should be more explicit for non-code files and probably need to sit down amongst ourselves and have a discussion, but I feel very petty labeling a doc listing colors for a theme, some docs, translations, even something as silly as a cppcheck rule file, etc as explicitly licensed when most reuse of their components would quite automatically fall under fair use. Even the completions files, given some/many were machine generated or even generated by the programs that we complete themselves.

Additionally, a million separate `.license` files is a complete non-starter from me.

I feel also like the scope here is just ballooning: this was nominally an issue about adding SPDX headers to files but it's turned into a discussion about trying to relicense some components as GPL 2.0 or later instead of the de facto fish GPL 2.0 license.

mqudsi (commented):
I'm uncomfortable with all the churn here and the explicit GPL2 assertions where they don't really make sense. I guess the latter is on us and we should be more explicit for non-code files and probably need to sit down amongst ourselves and have a discussion, but I feel very petty labeling a doc listing colors for a theme, some docs, translations, even something as silly as a cppcheck rule file, etc as explicitly licensed when most reuse of their components would quite automatically fall under fair use. Even the completions files, given some/many were machine generated or even generated by the programs that we complete themselves.

Additionally, a million separate `.license` files is a complete non-starter from me.

I feel also like the scope here is just ballooning: this was nominally an issue about adding SPDX headers to files but it's turned into a discussion about trying to relicense some components as GPL 2.0 or later instead of the de facto fish GPL 2.0 license.

ridiculousfish:
Echoing what others have said - adding this boilerplate to every file is way too much noise given the benefit. We can do this if it can be structured as a top-level default license, with special handling for exceptions with different licenses.

silverhook:
> A top-level COPYING file is obviously the default license for any file without the header, does REUSE not recognize this?

No, because in practice that often turns out not to actually be the case.

And even if it was, let’s say I copy a few files from Fish and put it into my own shell (let’s call it Hsif), which is MIT-licensed. So there is a `COPYING` file that has MIT license in it, and there’s hundreds of files of my own and a dozen of Fish files, none of which have license info in them. How is my user or downstream to know that there the Fish files are under GPL?

Now let’s say someone forks Hsif, modifies it and re-licenses it to something else (can do under MIT). Neither them nor their downstream would know that there are files from Fish in there that are under GPL (and it was violated).

> I don't like adding headers to every file without a strong reason because it adds noise and unnecesssary overhead to creating new files.

I’m surprised that two extra lines that can be very simply put into a very simple [git hook](https://www.git-scm.com/docs/githooks) like this, should be a big overhead:

```shell
reuse addheader --copyright (git config --get user.name)\ \<(git config --get user.email)\> --license GPL-2.0-only
```

or without relying on external tools prepending essentially this:

```shell
echo "// SPDX-FileCopyrightText: © "(date +%Y)" "(git config --get user.name)" <"(git config --get user.email)">"\
\n"//"\
\n"// SPDX-License-Identifier: GPL-2.0-only"
```

The big chore was analysing the state of it as it is, _because_ the files were not appropriately tagged. Which I feel shows the worth of carrying (human- and machine-readable) licensing and copyright info in the files themselves. It may not be your direct benefit, but a benefit to your downstream. Just as Fish would have benefited if the files it copied from elsewhere had REUSE/SPDX tags in them already back then.

> […] but I feel very petty labeling […] file, etc as explicitly licensed when most reuse of their components would quite automatically fall under fair use. Even the completions files, given some/many were machine generated or even generated by the programs that we complete themselves.

You can – and REUSE encourages that – always tag those files to be under a more permissive license like 0BSD, MIT-0, Unlicense, CC0-1.0 and make it explicit that you think those files are not worth copyright protection (or are even original enough to be copyrightable). I did not want to add that additional option into the discussion at this point, as it was complicated enough as is.

“Public domain” and “fair use”/“fair dealings”/limitations&exceptions are very jurisdiction-specific things so relying on them is quite fickle. There are also many false myths about them. And if unless you explicitly say it, how should your downstream know what _you_ intend it to be (and whether the applicable law agrees)?

> Additionally, a million separate `.license` files is a complete non-starter from me.

Regarding the dozen(s) of `.license` files, there is a third option, which is to list or [glob those files in `reuse.dep5`](https://reuse.software/faq/#bulk-license) (or better yet in `reuse.yaml` [once that gets added](https://github.com/fsfe/reuse-docs/issues/81) and dep5 gets deprecated).

Since the main benefit is that _license/© info travels as close as possible together with the source code it applies to_, the preference is 1) tags in $file > 2) tags in `$file.license` > 3) tags as entry in `reuse.dep5`/`reuse.yaml`

> I feel also like the scope here is just ballooning: this was nominally an issue about adding SPDX headers to files but it's turned into a discussion about trying to relicense some components as GPL 2.0 or later instead of the de facto fish GPL 2.0 license.

I do agree that discussing is becoming a bit too complicated a bit soon (and apologise from my side!) – again, I argue, because these files did not carry license/© tags before – and Fish project should first decide whether they want to have the REUSE/SPDX tags at all. If needed/desired, we can track down the weird parts later on.

faho:
>I’m surprised that two extra lines that can be very simply put into a very simple [git hook](https://www.git-scm.com/docs/githooks) like this, should be a big overhead:

And we would have to explain that to every single person who makes a small pull request to add a simple completion script.

>And even if it was, let’s say I copy a few files from Fish and put it into my own shell (let’s call it Hsif), which is MIT-licensed.

That is a much much much less common case than adding files to fish.

This is optimizing for the wrong thing.

>Regarding the dozen(s) of .license files, there is a third option, which is to list or [glob those files in reuse.dep5](https://reuse.software/faq/#bulk-license) (or better yet in reuse.yaml https://github.com/fsfe/reuse-docs/issues/81 and dep5 gets deprecated).

This would be acceptable.

As far as I'm concerned, we can also keep the headers in src/ (which much fewer people touch), but for share/functions and share/completions we don't want a per-file header, because that's where we want to make adding new files as easy as possible.

silverhook:
> > I’m surprised that two extra lines that can be very simply put into a very simple [git hook](https://www.git-scm.com/docs/githooks) like this, should be a big overhead:
> 
> And we would have to explain that to every single person who makes a small pull request to add a simple completion script.

If I understand git hooks correctly, if you `git add` e.g. `.git/hooks/reuse_header.sh` to the repo, it would work automatically, no?

It could also be part of a CI/CD instead.
 
> > Regarding the dozen(s) of .license files, there is a third option, which is to list or [glob those files in reuse.dep5](https://reuse.software/faq/#bulk-license) (or better yet in reuse.yaml [fsfe/reuse-docs#81](https://github.com/fsfe/reuse-docs/issues/81) and dep5 gets deprecated).
> 
> This would be acceptable.
> 
> As far as I'm concerned, we can also keep the headers in src/ (which much fewer people touch), but for share/functions and share/completions we don't want a per-file header, because that's where we want to make adding new files as easy as possible.

This could be a working solution, yes. And if someone contributed a file to `share/functions/*` or `share/completions/*` they could still add headers to their files if they wanted to, and it would be OK. And if they didn’t add them it would default to the glob in dep5/yaml.

> I would also take this over the separate file for each theme - because it's ridiculous to have _two files_ for something that amounts to […]

If we’re already globbing for `share/functions/*` and `share/completions/*` it makes even more sense to add the non-editable files to dep5/yaml.

faho:
> If I understand git hooks correctly, if you git add e.g. .git/hooks/reuse_header.sh to the repo, it would work automatically, no?

I am *reasonably* sure you can't cause git hooks to run on github. Git hooks are arbitrary scripts, so I suspect github doesn't run those server-side.

>It could also be part of a CI/CD instead.

We do not currently do anything via CI other than run the tests. Particularly nothing which *writes back* to the repo, and I would prefer to keep it that way.

It's nice when things are just what you commit, especially if the magic you could add on top is just to add some license boilerplate.

If we *can* add the license information in a (for us) nicer and simpler way, then that's what we'll do.

>If we’re already globbing for share/functions/* and share/completions/* it makes even more sense to add the non-editable files to dep5/yaml.

Sure. That was only meant as a quick example of the big important directories, there may be more places where handling licenses in bulk makes sense.

silverhook:
> > If I understand git hooks correctly, if you git add e.g. .git/hooks/reuse_header.sh to the repo, it would work automatically, no?
> 
> I am _reasonably_ sure you can't cause git hooks to run on github. Git hooks are arbitrary scripts, so I suspect github doesn't run those server-side.

I’m pretty sure of that too. I meant that the (pre-commit) hooks run on the _local_ copy of the repo pre-commit.

Do you get many commits from people editing through GitHub’s UI instead of using `git` locally?

> > It could also be part of a CI/CD instead.
> 
> We do not currently do anything via CI other than run the tests. Particularly nothing which _writes back_ to the repo, and I would prefer to keep it that way.
> 
> It's nice when things are just what you commit, especially if the magic you could add on top is just to add some license boilerplate.

Totally understandable.

> If we _can_ add the license information in a (for us) nicer and simpler way, then that's what we'll do.
> 
> > If we’re already globbing for share/functions/* and share/completions/* it makes even more sense to add the non-editable files to dep5/yaml.
> 
> Sure. That was only meant as a quick example of the big important directories, there may be more places where handling licenses in bulk makes sense.

Once/if there’s an agreement in the Fish maintainership, I’d be happy to tackle the agreed changes.

So, if I understand correctly, the current proposal is:

- put REUSE/SPDX tags in files where possible: `src`, (`doc_src`, `test`, …)
- glob folders that get many drive-by commits into reuse.dep5/.yaml: `share/completions/*`, `share/functions/*`
- put/glob all files that cannot be edited into reuse.dep5/.yaml – all that currently in this PR has a `.license` file

krobelus:
> I’m pretty sure of that too. I meant that the (pre-commit) hooks run on the local copy of the repo pre-commit.

Git hooks must be enabled explicitly due to security reasons. So it's a manual step that we don't want to impose upon new contributors.

silverhook:
@krobelus, I did not know, but did wonder about security of hooks, so that makes perfect sense. I’ll keep it in mind, thanks!

faho:
Alright, I'm gonna close this because it already conflicts and is going to start conflicting more.

I think it's clear what needs to be done to get this into a shape where we'd merge it. That can be done in a new PR.

silverhook:
@faho, if at any point you need help with this, I’m more than happy to help.

PR #9328, opened 2022-11-04
Update `tree` completions

dseight:
## Description

Update completions for the tree command. There are a lot of new options were added since the 1.6.0 release (which apparently was used to create current completions).

Options are also reordered to follow the `tree` [help](https://gitlab.com/OldManProgrammer/unix-tree/-/blob/c70f1b9adbe98016b658e25b05978905f3f85d92/tree.c#L599).

## TODOs:

### DIFF
--- a/share/completions/tree.fish
+++ b/share/completions/tree.fish
@@ -1,5 +1,3 @@
-
-
 ## Listing options
 complete -c tree -s a -d 'All files are listed'
 complete -c tree -s d -d 'List directories only'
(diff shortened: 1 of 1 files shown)


share/completions/tree.fish
```diff
@@ -26,33 +29,40 @@ complete -c tree -s g -d 'Displays file group owner or GID number'
...
+complete -c tree -s r -d 'Reverse the order of the sort'
 complete -c tree -l dirsfirst -d 'List directories before files (-U disables)'
+complete -c tree -l filesfirst -d 'List files before directories (-U disables)'
+complete -c tree -l sort -d 'Select sort' -xa 'name version size mtime ctime'
 
 ## Graphics options
 complete -c tree -s i -d 'Don\'t print indentation lines'
 complete -c tree -s A -d 'Print ANSI lines graphic indentation lines'
-complete -c tree -s S -d 'Print with ASCII graphics indentation lines'
+complete -c tree -s S -d 'Print with CP437 (console) graphics indentation lines'
```
mqudsi, line 52: I feel like "ASCII graphics" is going to be more comprehensible than "CP437 (console) graphics"?

mqudsi (commented):
Thanks for the update. Indeed many of our completions get stale over time, so thanks for updating these. I do have a comment about one change, if you don't mind taking a look.

share/completions/tree.fish
dseight: ASCII contains only 128 values and no pseudo-graphics at all, while CP437  _does_ contain pseudo-graphics — which is enabled by this option. So, CP437 feels to be the correct name here. 

Also, this is what we got in [upstream](https://gitlab.com/OldManProgrammer/unix-tree/-/blob/c70f1b9adbe98016b658e25b05978905f3f85d92/tree.c#L647) right now (the corresponding [change](https://gitlab.com/OldManProgrammer/unix-tree/-/commit/be2d80bd8b64064b4050fd3d8432c762bae09f25#94d3c813985dd997752870f2f64fb02f61c2eaa3_563_654) in tree was introduced more than eight years ago).

mqudsi:
Thanks. Merged.

Issue #8015, opened 2021-05-18
Autocomplete tag names after `vim -t`


cassepipe:
(fish version 3.1.0 on latest Linux Mint)

Would it be possible to have fish look for matching tags in the tags file after entering `vim -t` ?

Thanks for all the great work you have been doing so far.

faho:
Works for me - both complete files.

Please upgrade to fish 3.2.2, you are missing 2357 commits - 17% of all commits ever made to fish. Then try again. Use [our PPA](https://launchpad.net/~fish-shell/+archive/ubuntu/release-3).

faho:
Alright, after checking the zsh completion I can't see any special case of finding a "tags file" (as much as I can read zsh-script), it also doesn't seem like vim requires a special name or path or anything, so completing just them isn't tenable.

We already complete files in this case so there seems to be nothing to do.

cassepipe:
Thanks for your answer.

I upgraded thanks to the ppa (thank you) and I get the same behavior,

Not sure I understand the secondesentence of your last post. 
I guess the solution would be to look up for tag names in the `tags` or `.tags` files whenever `vi -t` or `vim -t` is entered. I don't see what is untenable about it.

zanchey:
zsh can complete tags, by walking up the tree until it finds a tags file then running it through some [fairly terrifying regular expressions](https://sourceforge.net/p/zsh/code/ci/master/tree/Completion/Base/Widget/_complete_tag).

A better approach might be to ask vim to spit out a tags list for the file in question, but I can't manage that with ten minutes of thrashing around.

cassepipe:
@zanchey When you type `vim -t`, vim uses a `tags` or `TAGS` file. In it are the tags generated for all the files of a project (minus what's been excluded by ctags configuration). 
You can't specify a file and then a tag, so there is no "file in quesiton alas.

The idea is not to jump to a specific place in a file but to a specific function you are not really sure where it is (in a big project). 
But one tends to make typos or not remember exactly what was the function name, that's where autocompletion from a tagfile comes in really handy.

Don't you think that feature request deserves being re-opened ?

zanchey:
Yeah, I think it's worth looking into.

zanchey:
I was initially dubious about the 'look for the `tags` file' approach because there was a workflow from vim-fugitive which used `.git/tags` instead, but it turns out that stopped being recommended some years ago so I think the zsh approach is fine.

cassepipe:
[...]

PR #9329, opened 2022-11-06
Autocomplete tag names after `vim -t`

dseight:
## Description

Implement completion for vim tags from any place within the source tree.

To prevent freezes on a huge tags file (e.g., on one from the Linux kernel source tree), amount of completion lines is limited to 10000.

Note that the TAGS file (EMACS-compatible tags file) is not searched here as it would not be used by vim anyway.

Fixes issue #8015

## TODOs:

### DIFF
--- a/share/completions/vim.fish
+++ b/share/completions/vim.fish
@@ -1,6 +1,36 @@
 # these don't work
 #complete vim -a - -d 'The file to edit is read from stdin. Commands are read from stderr, which should be a tty'
 
+function __fish_vim_find_tags_path
+    set -l max_depth 10
+    set -l tags_path tags
+
+    for depth in (seq $max_depth)
+        if test -f $tags_path
+            echo $tags_path
+            return 0
+        end
+
+        set tags_path ../$tags_path
+    end
+
+    return 1
+end
+
+function __fish_vim_tags
+    set -l token (commandline -ct)
+    set -l tags_path (__fish_vim_find_tags_path)
+    or return
+
+    # To prevent freezes on a huge tags file (e.g., on one from the Linux
+    # kernel source tree), limit matching tag lines to some reasonable amount
+    set -l limit 10000
+
+    # tags file is alphabetically sorted, so it's reasonable to use "look" to
+    # speedup the lookup of strings with known prefix
+    command look $token $tags_path | head -n $limit | cut -f1,2 -d \t
+end
+
 # todo
 # +[num]        : Position the cursor on line number
 # +/{pat}       : Position the cursor on the first occurrence of {pat}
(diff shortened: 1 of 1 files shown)


dseight:
Updated PR to show the name of the file containing the tag.

mqudsi (commented):
I don't know if it's related to your performance issues or not, but on Linux `look(1)` does not perform a binary search by default (but it does on BSD and macOS). 

If you're on Linux, try using `look -b` instead and see if that performs better for you? As you said, the `tags` file is already sorted.

We would probably want to use that in all cases, regardless of whether it's a huge tag file or not and regardless of whether you'll keep the two character search minimum for huge files or not. This does mean you'd need to feature test first to figure out whether to execute `look` or `look -b` since the latter does not run under BSD/mac.

krobelus:
TIL about `look`. My `llok` from util-linux 2.38 does not have the `-b` option.

krobelus:
A typical approach to fix the performance issues is to limit the number of results, e.g. via `head -n 1000` but it's not that simple when doing a binary search.

dseight:
> I don't know if it's related to your performance issues or not, but on Linux `look(1)` does not perform a binary search by default (but it does on BSD and macOS).
> 
> If you're on Linux, try using `look -b` instead and see if that performs better for you? As you said, the `tags` file is already sorted.

I do not see `-b` option in look in [util-linux](https://github.com/util-linux/util-linux/blob/21e52455116be0ae48592014aa0290f1120b3c1f/misc-utils/look.c#L113), even on the first imported version (which dates back to 2006). So, it probably would be okay to assume that look on Linux does binary search by default.

dseight:
> A typical approach to fix the performance issues is to limit the number of results, e.g. via `head -n 1000` but it's not that simple when doing a binary search.

Oh, this would be a great thing to just _limit_ completions instead of refusing to show them unless there are several characters typed. Thank you for the suggestion. I'll do that instead of messing with the file size.

mqudsi:
> I do not see -b option in look in [util-linux](https://github.com/util-linux/util-linux/blob/21e52455116be0ae48592014aa0290f1120b3c1f/misc-utils/look.c#L113), even on the first imported version (which dates back to 2006). So, it probably would be okay to assume that look on Linux does binary search by default.

This is weird, Ubuntu 18.04 ships an old BSD version of `look`.

dseight:
So, I investigated a little bit, and it seems that only Debian up to 10 and Ubuntu up to 20.04 are using this old look, while Debian 11 and Ubuntu 22.04 are using look from the util-linux (which is utilizing binary search by default, as any modern BSD look does).

Also, even with the linear search, searching for something like "f" (in a tags file for Linux kernel) and limiting it to 10000 lines will take only 500 ms, which is a good result. Limiting the lines count to 10000 helps a lot to prevent freezes in completions with an initial blank value.

I think it would be okay to go with the current solution, even though it could be a bit slower on older Linux systems — to make things a little bit cleaner.

mqudsi:
Thanks for looking into it. I guess it's fine the way it is. If it ever slows me down, I'll patch it 😂.

The only concern I have for posterity is that you could potentially traverse mountpoints while searching for the tags file, but yours isn't the only completion to do that and it's not as great of a concern when traversing up the tree rather than down it.

Issue #9343, opened 2022-11-12
macOS without Xcode: fish_prompt spams install prompt dialogs


floam:
On macOS,  `/usr/bin/git` exists even if `git` might not be installed - it is a tiny libxcselect.dylib wrapper that basically runs `xcrun git` which finds `git` in the installed Xcode or Command Line Tools (...Developer/) path. It prompts the user to install Xcode if they don't have it, which they very well might not if they installed fish via the .pkg installer or are trying the .app bundle. I am not sure what it does if the user isn't an administrator that can't install Xcode anyhow, like if someone tries fish in a computer lab or something.

So, `git` exists but running `git` prints an error in this situation:

```console
> git
xcode-select: note: no developer tools were found at '/Applications/Xcode.app',
requesting install. Choose an option in the dialog to download the command line
developer tools.
```

… and pops up a modal dialog:
<img width="504" alt="Screenshot 2022-11-11 at 3 30 22 PM" src="https://user-images.githubusercontent.com/291142/201447633-69950b25-74eb-4b55-895a-209d201188b3.png">

Since we try `git` in `fish_git_prompt` which `fish_prompt` tries by default, as long as `command -sq git` works, this makes fish pretty unusable unless the person relents. The only positive thing I can say, is it is throttled to display no more than once per 10 seconds.  It's not our fault, but it's annoying enough I'm calling it a bug.

https://user-images.githubusercontent.com/291142/201447351-6fd7da9d-ac51-4a96-a511-c20654ed76c0.mp4

The solution here is going to be gating `git` behind some check if the developer tools are installed if `command git` is /usr/bin/git on Darwin.

edit: or find a trick to suppress the dialog.

floam:
There is a related issue, that I can't find right now, where after either Xcode updates, macOS updates, or maybe just reboots, the cache `xcrun` uses is regenerated, which makes fish really slow to launch, because `git` can take 3-5 seconds (or more like 10-15 on an old slow HDD iMac I tried once!) to do anything while it regenerates/does stuff. So you are sitting there looking at a blank screen waiting for the prompt to print when you open a terminal.

floam:
The issue was #6625, and it was apparently closed by mistake.

Issue #6625, opened 2020-02-18
git prompt causes FISH to start very slowly on OS X


duganchen:
```
dugan ) fish --version
fish, version 3.1.0
dugan ) echo $version
3.1.0
dugan ) uname -a
Darwin MacBook-Pro.local 19.3.0 Darwin Kernel Version 19.3.0: Thu Jan  9 20:58:23 PST 2020; root:xnu-6153.81.5~1/RELEASE_X86_64 x86_64
dugan ) echo $TERM
xterm-256color-italic
dugan ) sh -c 'env HOME=$(mktemp -d) fish'
```

I had the common OS X issue where FISH would take seconds to load. It would only happen once per reboot. The first time I tried to start FISH after a reboot, it would take seconds to start. If you tried to start it again, it would start instantly. At least, until the next reboot.

I profiled it: and the cause seems to be having a prompt with VCS information in it? This absolutely jumped out at me:
```
356	7369742	--> fish_vcs_prompt
360	379	---> source /usr/local/Cellar/fish/3.1.0/share/fish/functions/fish_vcs_prompt.fish
19	19	----> function fish_vcs_prompt --description "Print the prompts for all available vcsen"
    # If a prompt succeeded, we assume that it's printed the correct info.
    # This is so we don't try svn if git already worked.
    fish_git_prompt
    or fish_hg_prompt
    or fish_svn_prompt
...
394	5204357	---> fish_git_prompt

```

I tried switching to a prompt without VCS information, and now FISH starts instantly. Even right after a reboot.

Both profile results ("FISH Profile with git prompt" and "FISH profile with minimalist prompt" are here:

https://gist.github.com/duganchen/7467164add3bcfa6cae0e2f2741ca8bc

This, more precisely, is what's taking so much time on startup:

```
277	5194027	----> set -l repo_info (command git rev-parse --git-dir --is-inside-git-dir --is-bare-repository --is-inside-work-tree HEAD 2>/dev/null)
5193750	5193750	-----> command git rev-parse --git-dir --is-inside-git-dir --is-bare-repository --is-inside-work-tree HEAD 2>/dev/null
```

And no, I'm (obviously) not in a git repository when FISH starts.

If I change fish_prompt_git to call git with its absolute path: well, that makes no difference:

```
327	4283632	----> set -l repo_info (/usr/bin/git rev-parse --git-dir --is-inside-git-dir --is-bare-repository --is-inside-work-tree HEAD 2>/dev/null)
[...]

PR #9345, opened 2022-11-12
fish_git_prompt: fix macOS lag after reboot and problem when Xcode not installed

floam:
## Description

This fixes the two xcrun-related problems encountered due to the nature of the `/usr/bin/git` shim on macOS:

### Slow `git` after reboot

Due to $TMPDIR being cleaned out on reboots, and a cache file that is slow to generate, it can take multiple seconds for the `git` shim to locate the actual Git binary inside of the developer directory that came with Xcode/CLT, and it does some security checks on the the Xcode.app bundle that can be slow before creating a cache file. We detect the file not existing, and run `git --version &` in the background, skip printing the git prompt, and when an on-job-exit handler for that `git --version` finally runs we reprint the prompt. 

The behavior is nice. You can compare what happens after `rm (xcrun --show-cache-path)`: instead of fish appearing to hang for a while before the prompt finally prints, you get the asynchronous prompt effect with a regular prompt that displays the git portion later.

Fixes issue #6625

### Git might not actually be installed

If Xcode or the smaller command line tools package aren't actually installed or working, running `git` would cause a dialog to appear asking the user to install it. This isn't nice if you are running that binary every time you draw the prompt.

Fixes issue #9343


floam:
I notice the problem isn't as bad as it used to be, the slow cache regeneration. It takes like 900ms on my current M1 Pro  MacBook Pro. It definitely used to be 5, 10+ seconds on older hardware; It's possible Xcode has gotten a little faster over the years too (xcrun has been a thing since, oh 2012/2013 Mavericks I believe.)

ridiculousfish (changes requested):
Man it suck so hard we have to do this. But we do, you're right. Thank you for tackling it.

I think your overall approach is right. I have some suggestions for simplifying it.

Instead of polluting `fish_git_prompt`, let's introduce a new function (maybe `__fish_git_prompt_ready`) whose job it is to figure out if it's safe to run git. That's where we'll add this logic on Mac.

Now instead of adding this top-level `uname` variable we can just define the function conditionally:

    if "$(uname)" = Darwin -a "$(command -v git)" = /usr/bin/git
             function __fish_git_prompt_ready
                        # all the stuff
             end
    else
            function __fish_git_prompt_ready
                  return 0
            end
    end
    
Second, I like the approach of running `git` in the background and ensuring we only do it once (if something goes wrong, we don't want to spawn a bunch of background procs). However I think we can achieve the same by simply running it at global scope. The extra waiting function can also be omitted; we'll just test for the path.

So basically the structure I propose is:

    if "$(uname)" = Darwin -a "$(command -v git)" = /usr/bin/git
             # Kick off git in the background to populate the cache.
             command git --version &>/dev/null 2>
             function __fish_git_prompt_ready
                        test -e "$(xcrun --show-cache-path)" || return 1
                        # Path exists, replace our function.
                        function __fish_git_prompt_ready; return 0; end
                        return 0
             end
    else
            function __fish_git_prompt_ready
                  return 0
            end
    end

ridiculousfish:
[An article about this nonsense](https://lapcatsoftware.com/articles/xcrun.html)

floam:
The reason the security checks are done, is somethuing we might vaguely remember, [XcodeGhost](https://en.wikipedia.org/wiki/XcodeGhost).

floam:
After sleeping on it,I realize there is going to be an additional cold-start "install on demand" issue: python3, when we do that one-off thing with the manpage completion generator script. (macOS no longer comes with Python 2, Python 3 is shipped with the command line tools.)

floam:
To reproduce slowness issue:
```fish
rm (xcrun --show-cache-path)
```

To reproduce the virgin macOS issue:
```fish
sudo rm -rf /Library/Developer/CommandLineTools
sudo mv /Applications/Xcode.app/ /Applications/Xcode_.app
sudo xcode-select --reset
```

ridiculousfish:
@floam do you anticipate getting to this shortly, or shall I take it over?

floam:
I was a bit busy but things should clear up tomorrow. I’ll get into it.

faho:
>For the manpage completions, I'd prefer honestly just skipping what we do on macOS and, if python3 exists, installing the equivalent of a monthly cron job into launchd on macOS. I don't dig how it never re-runs.

We don't do that anywhere else, and adding a scheduled job that runs outside of fish (whether cron or something else) is a bit of a land grab that I'm uncomfortable with. For one we would also have to ensure it's removed when fish is.

It's also not typically much of an issue - the list of commands doesn't change by a large percentage (the long tail of simple commands is already gonna be part of the initial install), and man completions aren't great anyway.

So I would be against this. A compromise I would be okay with is documenting these things and letting the user enable them (e.g. shipping a disabled systemd timer and the launchd equivalent would be okay by me).

ridiculousfish:
I'm proposing https://github.com/fish-shell/fish-shell/pull/9434 which builds on these ideas. It does not include any sort of cron job or launchd agent, it merely skips running python3 by default if it's only a stub.

PR #9349, opened 2022-11-13
completions: added ykman

exploide:
## Description

Just added completions for `ykman`, the YubiKey management tool. Since they are using click, it is sufficient to source their autogenerated completions.

## TODOs:

### DIFF
--- a/share/completions/ykman.fish
+++ b/share/completions/ykman.fish
@@ -1,0 +1,1 @@
+_YKMAN_COMPLETE=fish_source ykman | source


floam:
And their packaging does not install completions in any way, correct?

floam:
(All they'd need to do is drop the output of that on install into `/usr/share/fish/vendor_completions.d/`)

exploide:
Their README states that autocompletions are not automatically enabled.

And in the end, that would ultimately be up to the package maintainer of the distribution. And on Fedora the files are not included for ykman. Which is expected because the repository doesn't contain them either. They would need to be generated.

PR #9355, opened 2022-11-17
completions: Add toot

bartlibert:
## Description

Added completion for the mastodon client [toot](https://github.com/ihabunek/toot)

## TODOs:

### DIFF
--- a/share/completions/toot.fish
+++ b/share/completions/toot.fish
@@ -1,0 +1,87 @@
+set -l _sub_commands 'login login_cli activate logout auth tui whoami whois notifications instance search thread timeline post upload delete favourite unfavourite reblog unreblog reblogged_by pin unpin follow unfollow mute unmute block unblock'
+
+function __fish_complete_toot_accounts
+    toot auth | awk '/^\*/{print $2}'
+end
+
+# No file completion for most subcomands
+complete -c toot -f
+
+# # Subcommands
+complete -c toot -a activate -n "__fish_is_first_token" -d 'Switch between accounts.'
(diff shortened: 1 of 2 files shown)


mqudsi:
These are nicely dynamic, good work. (My, mastadon is suddenly getting popular thanks to Mr Musk!)

My only concern is that some of your completions are overly long. The completion description is not intended in lieu of a man page and shouldn't (with rare exceptions) include additional info or documentation like when to or not to use something, what a valid payload would be, etc. It should ideally be three or four words max.

Do you think you can prune down some of the too-long descriptions?

If you like, you can also use the shorter `__fish_is_first_token` instead of repeating `not __fish_seen_subcommand_from $_sub_commands` everywhere.

bartlibert:
I've updated the descriptions and used `__fish_is_first_token`, I did not know about that one, thanks for the tip!

share/completions/toot.fish
```diff
@@ -0,0 +1,87 @@
+set -l _sub_commands 'login login_cli activate logout auth tui whoami whois notifications instance search thread timeline post upload delete favourite unfavourite reblog unreblog reblogged_by pin unpin follow unfollow mute unmute block unblock'
+
+function __fish_complete_toot_accounts
+    toot auth | awk '/^\*/{print $2}'
```
mqudsi, line 4: This looks like a portable use of `awk` so I'm not really concerned, but fyi `string replace -rf [pattern] [replacement]` *should* be able to also handle this (quicker since no process needs to be spawned).

PR #9382, opened 2022-11-26
reg: completions for key entries

### LINKED ISSUE #9380: Complete registry subkeys for `reg` completion
For instance autocomplete `<keyname>` for [`reg add`](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/reg-add) and other `reg` subcommands.

This issue can be closed later by me or somebody else. It's not prioritized for me as I am Linux user. But I'll try to close it.

EmilySeville7cfg:
## Description

Complete subkeys for correctly written keys via `reg query` (closes #9380) for some subcommands:

- add
- delete
- export
- query
- save

image

Only tested with `wine` 6.0.

## TODOs:


share/completions/reg.fish
```diff
@@ -1,4 +1,24 @@
...
+            set output (echo $output | string replace " " "\n" | string collect)
+            echo $output
+        end
+    #end
+end
+
 function __reg_add_complete_args -a previous_token
+    if test "$previous_token" = add
+        set -l current_token (commandline -toc)
+        __reg_run_reg_safely query $current_token
```
EmilySeville7cfg, line 18: For `reg add HKEY_CURRENT_CONFIG` it should suggest:

```
HKEY_CURRENT_CONFIG
HKEY_CURRENT_CONFIG\Software
HKEY_CURRENT_CONFIG\System
```

But now it doesn't do it. How to fix it?

faho: It's just `commandline -ct`.

share/completions/reg.fish
```diff
@@ -1,4 +1,24 @@
+function reg # placeholder for testing with wine, will be removed at the end of PR
+    wine reg $argv 2> /dev/null
+end
+
+function __reg_run_reg_safely
+    #if which reg > /dev/null # these lines will be uncommented at the end of PR
+        set -l output (reg $argv | tr -d '\r' | tail --lines +2 | string collect)
+        if not string match -q -r "reg: Invalid syntax*" -- $output
+            set output (echo $output | string replace " " "\n" | string collect)
```
faho, line 9: This is just `set output (string split \n -- $output)`.

Or, rather... It's just *not* doing the `string collect` above to begin with.

Just keep the output split per line, and check the first line with that `string match` (`$output[1]`).

share/completions/reg.fish
EmilySeville7cfg: > I'll have to check if there's a reason it shouldn't work, but it's unnecessary so the easy workaround is to just leave it off

I suggest return non zero status or warn user in another way that it's doesn't have to work. Or maybe just note about it in man page for `commandline`.

EmilySeville7cfg:
I've done everything I wanted, this PR can be merged. 😄

share/completions/reg.fish
```diff
@@ -1,4 +1,20 @@
+function __reg_run_reg_safely
+    if which reg > /dev/null
```
faho, line 2: 1. `which` is a cheesy external command that might not be installed and shouldn't be used
2. When `which` doesn't find a command it prints an *error* to stderr, so this spews error messages
3. We already only load the description when the command to be completed is an executable thing

Concretely, this prevents me from testing this via `function reg; wine reg $argv; end`, even tho that would work.

Just remove the entire check. If you *have* to have a check, use `type -q reg`. Just `if type -q reg`. That checks if `reg` is any kind of runnable thing (function, command, builtin).

faho:
Merged, thanks!

PR #9383, opened 2022-11-26
Updated hostnamectl completions

exploide:
## Description

`hostnamectl` has been changed some time ago. The subcommands have different names and there is a bit of new functionality. I updated the completions to conform with the new behavior.

## TODOs:

### DIFF
--- a/share/completions/hostnamectl.fish
+++ b/share/completions/hostnamectl.fish
@@ -1,18 +1,31 @@
-set -l commands status set-hostname set-icon-name set-chassis set-deployment set-location
-
 complete -c hostnamectl -f
-complete -c hostnamectl -n "not __fish_seen_subcommand_from $commands" -a status -d 'Show current hostname settings'
-complete -c hostnamectl -n "not __fish_seen_subcommand_from $commands" -a set-hostname -d 'Set system hostname'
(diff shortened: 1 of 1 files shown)


mqudsi:
So these look like they're breaking changes. Do you know what systems have the old version and what systems have the new version? This ships with the distro and isn't updated out-of-stream, so we can't just assume users have the latest version.

mqudsi (changes requested):
So these look like they're breaking changes. Do you know what systems have the old version and what systems have the new version? This ships with the distro and isn't updated out-of-stream, so we can't just assume users have the latest version.

share/completions/hostnamectl.fish
```diff
@@ -1,18 +1,21 @@
-set -l commands status set-hostname set-icon-name set-chassis set-deployment set-location
```
EmilySeville7cfg, line 1: Optional fix: if `--version` or option with the same semantics is available please use it to decide what completions to provide.

exploide:
Made the subcommand names conditional based on hostnamectl version.

mqudsi:
Thanks, this looks good.

mqudsi (approved):

PR #9386, opened 2022-11-26
Add readelf completions

xfgusta:
## Description

This targets readelf from binutils. There are some variants like elftoolchain and elfutils. Because I don't know how to handle variants and I don't know them very well, I picked the most popular one. They seem to be quite similar, though.

Fixes issue #8746

## TODOs:


share/completions/readelf.fish
```diff
@@ -0,0 +1,75 @@
...
+complete -c readelf -s A -l arch-specific -d "Display architecture specific information"
+complete -c readelf -s c -l archive-index -d "Display the symbol/file index in an archive"
+complete -c readelf -s D -l use-dynamic -d "Use the dynamic section info when displaying symbols"
+complete -c readelf -s L -l lint -l enable-checks -d "Display warning messages for possible problems"
+complete -c readelf -s x -l hex-dump -d "Dump the contents of specified section as bytes" -a $sections -x
+complete -c readelf -s p -l string-dump -d "Dump the contents of specified section as strings" -a $sections -x
+complete -c readelf -s R -l relocated-dump -d "Dump the relocated contents of of specified section" -a $sections -x
+complete -c readelf -s z -l decompress -d "Decompress section before dumping it"
+for c in {a,A,r,c,L,f,F,g,i,o,m,p,t,R,l,s,O,u,T,U}
+	complete -c readelf -o w$c -d "Display the contents of DWARF debug sections"
```
krobelus, line 56: we can simplify the loop:

```fish
for c in a A r c L f F g i o m p t R l s O u T U
```

Actually it could be helpful to add to the description the specific section, for example

	complete -o wa -e 'Display the contents of the .debug_abbrev section'

share/completions/readelf.fish
```diff
@@ -0,0 +1,75 @@
...
+end
+complete -c readelf -l debug-dump -d "Display the contents of DWARF debug sections" -a 'abbrev addr aranges cu_index decodedline frames frames-interp gdb_index info loc macro pubnames pubtypes Ranges rawline str str-offsets trace_abbrev trace_aranges trace_info' -x
```
krobelus, line 58: apparently the argument is not required. For example,

	$ readelf /bin/fish --debug-dump

will work fine but this does not:

	$ readelf /bin/fish --debug-dump abbrev
	[..]
	readelf: Error: 'abbrev': No such file

It should be `--debug-dump=abbrev`.
Hence we should remove `-x` to provide a completion that includes the `=`.
This comment is obsolete if you're motivated to implement the next comment.

krobelus:
nice work! I'll wait a bit before merging, in case you'd like to incorporate the feedback.

share/completions/readelf.fish
xfgusta: > Actually it could be helpful to add to the description the specific section, for example
Yeah, I agree with it.

PR #9387, opened 2022-11-27
`calligra` office suite: add completions

EmilySeville7cfg:
## Description

Add completions for almost all `calligra` suite programs:

- `calligraconverter`
- `calligraflow`
- `calligraplan`
- `calligraplanwork`
- `calligrasheets`
- `calligrastage`
- `calligrawords`

## TODOs:


mqudsi (changes requested):
Thanks for submitting these, Emily. However, with the exception of `calligraconverter`, these are all GUI applications with no *actual* command line options being completed.

I'm not sure how many people are using a GUI application from the terminal, but these completions aren't going to be of any use. Do you have these apps installed? Can you check if they have actual options (besides help, version, license, etc) people would use that are revealed if you use `help-qt` or `help-kde` or `help-all` that we could actually complete instead of these just placeholder completions?

Pretty much every GUI application will have the ability to use `--version` or `--help` at the command line, and I don't feel it serves much purpose to document/complete that **if** that's all we're going to document/complete.

EmilySeville7cfg:
> Do you have these apps installed?

Unfortunately I don't. I've relied on manned.org manuals to write these completions.

mqudsi:
Ok, thanks for trying.

PR #9388, opened 2022-11-27
es: add completion

EmilySeville7cfg:
## Description

Completion for [es - extensible shell](https://manned.org/es.1).

## TODOs:

### DIFF
--- a/share/completions/es.fish
+++ b/share/completions/es.fish
@@ -1,11 +1,11 @@
-complete -c es -s c
-complete -c es -s s
-complete -c es -s i
-complete -c es -s l
-complete -c es -s e
-complete -c es -s v
-complete -c es -s x
-complete -c es -s n
-complete -c es -s p
-complete -c es -s o
-complete -c es -s d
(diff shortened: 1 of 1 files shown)


share/completions/es.fish
```diff
@@ -0,0 +1,10 @@
+complete -c es -s c
+complete -c es -s s
+complete -c es -s i
+complete -c es -s l
+complete -c es -s e
+complete -c es -s v
+complete -c es -s x
+complete -c es -s n
+complete -c es -s p
+complete -c es -s o
```
faho, line 10: You want to look at http://wryun.github.io/es-shell/manpage.html, which is the fork I could install on archlinux.

It has an additional "-d" option.

This also needs descriptions or we would likely get better results from parsing the man page.

faho (changes requested):

share/completions/es.fish
EmilySeville7cfg: Done. 😄

PR #9395, opened 2022-11-28
aliases: add completion

EmilySeville7cfg:
## Description

Simple [`aliases`](https://github.com/sebglazebrook/aliases) completion.

## TODOs:

### DIFF
--- a/share/completions/aliases.fish
+++ b/share/completions/aliases.fish
@@ -1,0 +1,1 @@
+complete -c aliases -a init -d 'Create aliases for the current directory'


mqudsi (commented):
What is `aliases` in this context?

EmilySeville7cfg:
@mqudsi, they are [aliases](https://github.com/sebglazebrook/aliases) literally for Bash.

mqudsi:
If it's for bash, then no one will be using it on fish, right?

EmilySeville7cfg:
From my understanding [this command](https://github.com/sebglazebrook/aliases#usage) can be run from any shell: it just creates file with aliases. But I understand that primary usage will be in Bash anyway.

faho:
Yeah this doesn't appear to be useful with fish, as it stands. The one command this currently completes does basically nothing on its own. It also has an *extremely* generic name and could easily clash with other things of the same name.

So I'm going to do a very rare thing for completions and reject this. Sorry!

Issue #7885, opened 2021-03-31
Perhaps use undo groups in key bindings that do multiple things


krobelus:
This came up [here](https://github.com/fish-shell/fish-shell/commit/7669e8e4977103da9225403824b7f7f03f767779#r45729914).

Every edit to the command line is its own undo group.
Multiple edits can be merged in an undo group using readline functions `begin-undo-group` and `end-undo-group`.

This ticket tracks if and how we should use undo grouping for some bind functions, like ones that do `self-insert expand-abbr`.

One option is to implicitly create undo groups for each binding; and maybe remove the explicit undo-grouping commands entirely, since they are [unused](https://github.com/fish-shell/fish-shell/commit/85ffa77b4e6d3f5f23e6a9ed3e879035eff239f5#r48937529).

ridiculousfish:
Creating a group for each binding seems like a natural and simple approach to me.

mqudsi:
I remember you (@krobelus) suggesting all bindings should use begin/end-undo-group instead of just the paste thing I was hacking at and thought it was a great idea at the time. I think a binding is pretty much "atomic" for the user in all cases.

PR #9398, opened 2022-11-30
builtin commandline: use readline

### LINKED ISSUE #3031: `commandline -f` ops reordered with respect to other commandline operations
I observe strange behaviour regarding the execution order of `commandline` calls _with_ `-f` option and _without_.

Example: I define the following function and bind it to a <kbd>ctrl</kbd><kbd>s</kbd>:

``` fish
function __surround
    commandline -f backward-kill-path-component
    commandline -i '"'
    commandline -f yank
    commandline -a '"'
end

bind \cs '__surround'
```

(in my `fish_user_key_bindings.fish`)

What happens instead:
- cut + paste (those with `-f`)
- insert + append

Here is a demo: [asciicast](https://asciinema.org/a/45721)

This is just a simplified example, I have encountered this behaviour in more situations and if this is not enough, I can provide more examples. 

---
- Fish version: `fish, version 2.3b2` (`homebrew install --devel fish`)
- Operating system: OS X El Capitan
- Terminal emulator: iTerm2

phillipwood:
## Description

Currently in a key-binding `commandline --insert` and `commandline --cursor` are always be executed before `commandline -f`. To work around this `commandline -f` special cases `begin-undo-group` and `end-undo-group` so they work when combined with text insertion but that means they do not work with `delete-char` or `kill-word` etc. This PR uses the suggestion in src/builtin/commandline.cpp to try and fix that. It converts the builtin `commandline` function to use readline functions for moving the cursor and inserting text. This has the advantage that `commandline --insert` and `commandline --cursor` can be freely mixed with `commandline -f` and they are executed in the order that they are written.

### Some notes on the commits:

 - The second commit is larger than I'd like, it would be smaller if we could avoid renaming `enum readline_cmd_t` by choosing a different name for the class but I failed to come up with a good name.

 - The fourth commit should perhaps be squashed into the fifth to avoid having to comment out the temporary test failures

 - The fifth commit would benefit from using `std::variant` but that is not an option in C++11. I did look at some C++11 implementations with a view to importing one but they were all huge.

Fixes #3031

## TODOs:

### DIFF
--- a/src/builtins/commandline.cpp
+++ b/src/builtins/commandline.cpp
@@ -24,13 +24,6 @@
 #include "../wgetopt.h"
 #include "../wutil.h"  // IWYU pragma: keep
 
-/// For text insertion, how should it be done.
-enum {
-    REPLACE_MODE = 1,  // replace current text
-    INSERT_MODE,       // insert at cursor position
-    APPEND_MODE        // insert at end of current token/command/buffer
-};
-
 /// Handle a single readline_cmd_t command out-of-band.
 void reader_handle_command(readline_cmd_t cmd);
 
@@ -38,50 +31,6 @@ void reader_handle_command(readline_cmd_t cmd);
 void commandline_get_part(const wchar_t *current_buffer, size_t current_cursor_pos,
                           commandline_part_t buffer_part, const wchar_t **begin,
                           const wchar_t **end);
-
-/// Replace/append/insert the selection with/at/after the specified string.
-///
-/// \param begin beginning of selection
-/// \param end end of selection
-/// \param insert the string to insert
-/// \param append_mode can be one of REPLACE_MODE, INSERT_MODE or APPEND_MODE, affects the way the
-/// test update is performed
-/// \param buff the original command line buffer
-/// \param cursor_pos the position of the cursor in the command line
-static void replace_part(const wchar_t *begin, const wchar_t *end, const wchar_t *insert,
-                         int append_mode, const wchar_t *buff, size_t cursor_pos) {
-    size_t out_pos = cursor_pos;
-
-    wcstring out;
-
-    out.append(buff, begin - buff);
-
-    switch (append_mode) {
-        case REPLACE_MODE: {
-            out.append(insert);
-            out_pos = out.size();
-            break;
-        }
-        case APPEND_MODE: {
-            out.append(begin, end - begin);
-            out.append(insert);
-            break;
-        }
-        case INSERT_MODE: {
-            long cursor = cursor_pos - (begin - buff);
-            out.append(begin, cursor);
-            out.append(insert);
-            out.append(begin + cursor, end - begin - cursor);
-            out_pos += std::wcslen(insert);
-            break;
-        }
-        default: {
-            DIE("unexpected append_mode");
-        }
-    }
-    out.append(end);
-    commandline_set_buffer(out, out_pos);
-}
 
 /// Output the specified selection.
 ///
@@ -131,7 +80,7 @@ maybe_t<int> builtin_commandline(parser_t &parser, io_streams_t &streams, const 
     bool cut_at_cursor = false;
 
     int argc = builtin_count_args(argv);
-    int append_mode = 0;
+    maybe_t<commandline_insertion_mode_t> append_mode{};
 
     bool function_mode = false;
     bool selection_mode = false;
@@ -180,7 +129,7 @@ maybe_t<int> builtin_commandline(parser_t &parser, io_streams_t &streams, const 
     while ((opt = w.wgetopt_long(argc, argv, short_options, long_options, nullptr)) != -1) {
         switch (opt) {
             case L'a': {
-                append_mode = APPEND_MODE;
+                append_mode = commandline_insertion_mode_t::append;
                 break;
             }
             case L'b': {
@@ -188,11 +137,11 @@ maybe_t<int> builtin_commandline(parser_t &parser, io_streams_t &streams, const 
                 break;
             }
             case L'i': {
-                append_mode = INSERT_MODE;
+                append_mode = commandline_insertion_mode_t::insert;
                 break;
             }
             case L'r': {
-                append_mode = REPLACE_MODE;
+                append_mode = commandline_insertion_mode_t::replace;
                 break;
             }
             case 'c': {
@@ -369,7 +318,7 @@ maybe_t<int> builtin_commandline(parser_t &parser, io_streams_t &streams, const 
 
     // Set default modes.
     if (!append_mode) {
-        append_mode = REPLACE_MODE;
+        append_mode = commandline_insertion_mode_t::replace;
     }
 
     if (!buffer_part) {
@@ -467,20 +416,21 @@ maybe_t<int> builtin_commandline(parser_t &parser, io_streams_t &streams, const 
         return STATUS_CMD_OK;
     }
 
-    commandline_get_part(current_buffer, current_cursor_pos, *buffer_part, &begin, &end);
     int arg_count = argc - w.woptind;
     if (arg_count == 0) {
+        commandline_get_part(current_buffer, current_cursor_pos, *buffer_part, &begin, &end);
         write_part(begin, end, cut_at_cursor, tokenize, current_buffer, current_cursor_pos,
                    streams);
     } else if (arg_count == 1) {
-        replace_part(begin, end, argv[w.woptind], append_mode, current_buffer, current_cursor_pos);
+        reader_queue_ch(
+            readline_cmd_t(commandline_insertion_t(*buffer_part, *append_mode, argv[w.woptind])));
     } else {
         wcstring sb = argv[w.woptind];
         for (int i = w.woptind + 1; i < argc; i++) {
             sb.push_back(L'\n');
             sb.append(argv[i]);
         }
-        replace_part(begin, end, sb.c_str(), append_mode, current_buffer, current_cursor_pos);
+        reader_queue_ch(readline_cmd_t(commandline_insertion_t(*buffer_part, *append_mode, sb)));
     }
 
     return STATUS_CMD_OK;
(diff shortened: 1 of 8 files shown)


krobelus:
This is a great step in the right direction, thanks! I left some minor nitpicks.
Nice to see Git folk stop by :)

tests/pexpects/commandline.py
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 from pexpect_helper import SpawnedProc, control
+from re import escape
```
krobelus, line 3: I think we should generally call this `re.escape` instead of `escape`

tests/pexpects/commandline.py
```diff
@@ -12,6 +13,92 @@
...
+# commandline --cursor
+t("abcdef", "commandline --cursor 3", "3 abcdef")
+t("abc defghi jkl",
+  "commandline --cursor 8; commandline --current-token --cursor 1",
+  "5 abc defghi jkl")
+t("abc def", "commandline --current-token --cursor -- -3", "1 abc def")
+t("abc def", "commandline --current-token --cursor -- -100", "0 abc def")
+t("abc def",
+  "commandline --current-token --cursor 0; commandline --insert x",
+  "5 abc xdef")
```
krobelus, line 44: the tests are great. One alternative syntax is `"abc x%def"` instead of `"5 abc xdef"`

tests/pexpects/commandline.py
```diff
@@ -12,6 +13,92 @@
...
+  "commandline --current-job --cursor 5",
+  "14 echo ===; echo hello | echo there" )
+t("echo hello | echo there",
+  "commandline --current-process --cursor 4",
+  "16 echo hello | echo there")
+
+# commandline --insert
+t("abc", "commandline --cursor 2; commandline --insert x", "3 abxc")
+t("abc defghi jkl",
+  "commandline --cursor 8; commandline --current-token --insert x",
```
krobelus, line 55: Okay, this test shows that `--current-token` is ignored when `--insert` is given

tests/pexpects/commandline.py
```diff
@@ -12,6 +13,92 @@
...
+  "9 abc defgxhi jkl")
+
+# commandline --append
+t("abc def; ghi", "commandline --append x", "12 abc def; ghix")
+t("abc def ghi",
+  "commandline --cursor 5; commandline --current-token --append x",
+  "5 abc defx ghi")
```
krobelus, line 62: OTOH `--current-token` does make a difference with `--append`. Makes sense.

src/reader.cpp
```diff
@@ -138,6 +138,34 @@ static acquired_lock<commandline_state_t> commandline_state_snapshot() {
 
 commandline_state_t commandline_get_state() { return *commandline_state_snapshot(); }
 
+void commandline_get_part(const wchar_t *current_buffer, size_t current_cursor_pos,
```
krobelus, line 141: we should change `current_cursor_pos` to `cursor_pos` here.
Similarly, `current_buffer` should probably be `buff` like in the functions from `parse_util.cpp`.

src/input_common.cpp
```diff
@@ -270,3 +270,125 @@ void input_event_queue_t::prepare_to_select() {}
 void input_event_queue_t::select_interrupted() {}
 void input_event_queue_t::uvar_change_notified() {}
 input_event_queue_t::~input_event_queue_t() = default;
+
+readline_cmd_t::readline_cmd_t(const readline_cmd_t& rl) : id_(rl.id_) {
```
krobelus, line 274: we still use old-school pointer alignment (`readline_cmd_t &code`), `clang-format` should do the trick

src/input.cpp
```diff
@@ -212,12 +216,12 @@ static void input_set_bind_mode(parser_t &parser, const wcstring &bm) {
 }
 
 /// Returns the arity of a given input function.
-static int input_function_arity(readline_cmd_t function) {
-    switch (function) {
```
krobelus, line 216: I think this one should take the enum

src/input_common.cpp
```diff
@@ -270,3 +270,125 @@ void input_event_queue_t::prepare_to_select() {}
...
+        if (rl.id_ == id_t::set_cursor) arg_.cursor_pos = rl.arg_.cursor_pos;
+        id_ = rl.id_;
+    }
+    return *this;
+}
+
+readline_cmd_t::~readline_cmd_t() {
+    // Only destroy arg_.insertion if it is active.
+    if (id_ == id_t::insert_chars) arg_.insertion.~commandline_insertion_t();
+}
```
krobelus, line 329: this is tricky code but it looks correct.

> The fourth commit should perhaps be squashed into the fifth to avoid having to comment out the temporary test failures

I think so too, although it was nice to read this way

src/input_common.h
```diff
@@ -12,91 +12,185 @@
...
+    readline_cmd_t& operator=(readline_cmd_t&& rl) noexcept;
+
+    ~readline_cmd_t();
+
+   private:
+    id_t id_;
+
+    union arg_t {
+        /// Set if id_ is set_cursor
+        commandline_cursor_pos_t cursor_pos;
```
krobelus, line 180: Yeah, this is probably the best solution for now.
Just out of curiosity, here are some possible alternatives to the union:

1.

```cpp
class readline_cmd_arg_t {
	virtual ~readline_cmd_arg_t();
};
class readline_cmd_cursor_pos_arg_t : readline_cmd_arg_t {
	commandline_part_t part;
	long pos;
}
struct readline_cmd_t {
	id_t id_;
	std::unique_ptr<readline_cmd_arg_t> arg_;
}
```

2.

```cpp
class readline_cmd_t {
	virtual ~readline_cmd_arg_t();
};
class readline_cmd_beginning_of_line_t : readline_cmd_t {};
class readline_cmd_cmd_cursor_pos_t : readline_cmd_t {
	commandline_part_t part;
	long pos;
};
```

The second is closest to Rust enums but it probably does not make for a good for developer experience in C++.

src/builtins/commandline.cpp
```diff
@@ -306,17 +249,8 @@ maybe_t<int> builtin_commandline(parser_t &parser, io_streams_t &streams, const
...
-                // all queued `commandline` operations - including buffer modifications - are
-                // executed in order
-                if (mc == rl::begin_undo_group || mc == rl::end_undo_group) {
-                    reader_handle_command(*mc);
-                } else {
-                    // Inserts the readline function at the back of the queue.
-                    reader_queue_ch(*mc);
-                }
+                // Inserts the readline function at the back of the queue.
+                reader_queue_ch(readline_cmd_t(*mc));
```
krobelus, line 253: BTW we currently don't use undo groups.

Our plan is to make it so that every key press either
0. odes not modify undo history
1. adds to an existing undo entry (when adding to the current word)
2. or creates exactly one new undo entry (which needs the undo group if there are multiple changes)

This way undo groups need not be user visible. That's #7885

faho:
I have a question: Is this the right way around?

Here's how I understand this: Currently, when you do `commandline -f something`, we enqueue that `something` in the input queue, but when you do `commandline -i foo`, we insert that "foo" immediately.

That means, effectively, when you have a script that does `commandline -i foo`, you can then rely on `foo` being in the commandline *immediately*, but when you do `commandline -f backward-kill-word`, you can't rely on the word being killed until fish gets around to processing the queue that far.

Now this makes it so we put the input action into the queue as well, so now you can't rely on `foo` being in there immediately.

This should be reproducible with something like

```fish
function foo
    commandline -r foo
    commandline >&2
    commandline -r bar
    commandline >&2
end

bind \cg foo
```

That seems like

1. It has some potential to break scripts that rely on commandline inputs applying immediately
2. Commandline changes not applying immediately is kind of annoying to begin with

So I'm not really sure this is the right way around - maybe, instead of adding input actions to the queue as well we should be doing functions immediately? Or possibly drain the queue for every action?

krobelus:
you're right, that should probably be fixed before merging.
There are two things that can enqueue readline events: keyboard input and the `commandline` builtin.
Both run on the main thread so there is no race.
To fix the coherence I see two options:
1. always drain the input queue after running `commandline` (or equivalently, skip the queue)
2. always drain the input queue before running `commandline` (if it were to read commandline state)

faho:
Tagging this for after 3.6.0 as we'd prefer to get that out sooner and this would need some testing.

tests/pexpects/commandline.py
phillipwood: Ok, will do. Thanks for all your comments, sorry it has taken me so long to reply. I'll push a new version in the New Year.

tests/pexpects/commandline.py
phillipwood: I'm afraid I'm not quite sure what you mean

src/input_common.cpp
phillipwood: Will do. I think `clang-format` chooses whatever is most common in the file unless one sets
```
DerivePointerAlignment: false
PointerAlignment: Right
ReferenceAlignment: Pointer
```
in `.clang-format`

src/input_common.cpp
phillipwood: I'll squash them together.

phillipwood:
> Tagging this for after 3.6.0 as we'd prefer to get that out sooner and this would need some testing.

When is 3.6.0 due for release? @faho thanks for you comment, I need to think about it before replying properly

tests/pexpects/commandline.py
krobelus: I mean that instead of passing a separate parameter for the cursor position, we could designate some character to represent the cursor. Of course it's fine as-is

faho:
>When is 3.6.0 due for release

It's out now.

faho:
Alright, this is conflicting already and the approach needs to be reworked a bit.

We can, of course, revisit it later.

krobelus:
This should be fix for most practical cases by c3cd68dda (Process shell commands from bindings like regular char events, 2024-03-02)

The only limitation is that commands that want to see each other's updates to the commandline need to be passed to bind as separate arguments. So something like

    bind x foo bar

should work fine now; `foo` and `bar` are executed in this order and commandline updates are read in between

PR #9402, opened 2022-12-07
added completion for git branch --remotes (-r)

bagohart:
## Description

The option `--remotes` or `-r` for `git branch` was missing, I added it.

One thing I'm unsure about: If used as `git branch -rd` this will complete to `(__fish_git_local_branches)`, although it might be more helpful to complete it also to the remote branches, e.g. `origin/my-branch-name`. I'm not sure if fish's completion mechanisms support this distinction, and a `(__fish_git_remote_branches)` does not exist anyway, the closest I could find was `__fish_git_unique_remote_branches`, which seems to strip away the remote, which is not what would be helpful in this case.
Still, I guess it's better to add the flag, even if the branch-completion after that does not provide all reasonable results.

## TODOs:

### DIFF
--- a/share/completions/git.fish
+++ b/share/completions/git.fish
@@ -1166,6 +1166,7 @@ complete -f -c git -n '__fish_git_using_command branch' -s M -d 'Force rename br
 complete -f -c git -n '__fish_git_using_command branch' -s c -l copy -d 'Copy branch'
 complete -f -c git -n '__fish_git_using_command branch' -s C -d 'Force copy branch'
 complete -f -c git -n '__fish_git_using_command branch' -s a -l all -d 'Lists both local and remote branches'
+complete -f -c git -n '__fish_git_using_command branch' -s r -l remotes -d 'List or delete (if used with -d) the remote-tracking branches.'
 complete -f -c git -n '__fish_git_using_command branch' -s t -l track -l track -d 'Track remote branch'
 complete -f -c git -n '__fish_git_using_command branch' -l no-track -d 'Do not track remote branch'
 complete -f -c git -n '__fish_git_using_command branch' -l set-upstream-to -d 'Set remote branch to track'


faho:
>this will complete to (__fish_git_local_branches), although it might be more helpful to complete it also to the remote branches, e.g. origin/my-branch-name. 

Here are all the branch completion functions:

- `__fish_git_branches`, which does both local and remote branches, fully qualified with origin/ in the latter case
- `__fish_git_local_branches`, which does local branches
- `__fish_git_unique_remote_branches`, which does those remote branches that are unambiguous and can be used without the origin/ qualification

There is no separate `__fish_git_remote_branches` because that wasn't useful for now.

So you want `__fish_git_branches` if you don't care where the branch is.

faho:
>User-visible changes noted in CHANGELOG.rst

As a side note: We don't tend to list every small enhancement to completions separately in the changelog, because that would be pages and pages and our changelogs tend to be too long anyway.

bagohart:
> So you want `__fish_git_branches` if you don't care where the branch is.

Okay, so to make this work, we would have to either change 
```fish
complete -f -c git -n '__fish_git_using_command branch' -s d -l delete -d 'Delete branch' -xa '(__fish_git_local_branches)'
```
to
```fish
complete -f -c git -n '__fish_git_using_command branch' -s d -l delete -d 'Delete branch' -xa '(__fish_git_branches)'
```
which makes `git branch -d` less useful in most cases, or we would have to add a really ugly and brittle hack such as

```fish
complete -f -c git -n '__fish_git_using_command branch' -s d -l delete -d 'Delete branch' -xa '(__fish_git_branches)' --condition "__uses_rd_and_not_d" # pseudocode for the condition
```

therefore I propose to merge it as is.
I removed the note in the changelog, and sorry for all the force-pushing after your comment, my original branch was branched off a very old commit and I didn't expect to get a reply within 2 minutes :smile:

faho:
Merged, thanks!

PR #9412, opened 2022-12-13
abbr: Allow filtering by command

### LINKED ISSUE #9411: abbr: Allow restricting to a specific command's arguments
This is very much a nice-to-have, it does not need to be in the initial release with "Mega abbr".

One of the examples in #9313:

```fish
function git_c
    echo checkout
end
```

This is supposed to allow adding abbreviations for git subcommands. It would be cool if this could be

```fish
abbr --add --commmand git c checkout
```

Which allows us to save the regex, makes for easier and more correct matching and allows arbitrary transformations with arbitrary regexen without marking everything as not-an-error.

(Currently `abbr hahaha --regex '.*' --function false` will cause commands to always be marked as valid because the regex matches and it can't know that the function would fail. This is probably correct - I don't think we can run the function before and marking things as invalid even tho they could still match an abbr feels weird?)

Note that this would always be restricted to a command *argument* - because otherwise the command won't match. I don't think we need to add `--position argument` or require `--position anywhere` just for this. I also don't think we'd need to walk a command's wrap chain (i.e. `function g; git $argv; end` would need to have the abbrs specified again).

faho:
This allows specifying an abbreviation e.g. just for git:

```fish
abbr --command git c checkout
```

will expand `git c ` to `git checkout `, but not `echo c ` to `echo checkout`, without using regexes.

Fixes #9411.

## TODOs:

This is *not* necessary for 3.6.0, unless we can quickly come to an agreement that it's working and good.

### DIFF
--- a/doc_src/cmds/abbr.rst
+++ b/doc_src/cmds/abbr.rst
@@ -9,7 +9,7 @@ Synopsis
 .. synopsis::
 
     abbr --add NAME [--position command | anywhere] [-r | --regex PATTERN]
-                    [--set-cursor[=MARKER]]
+                    [--set-cursor[=MARKER]] [-c | --command COMMAND]
                     [-f | --function] EXPANSION
     abbr --erase NAME ...
     abbr --rename OLD_WORD NEW_WORD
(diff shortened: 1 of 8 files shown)


krobelus:
I don't think this adds expressiveness, so I'm not sure.
If it's a really common use case then it might be worth it. I don't think it is that common but that's a chicken-and-egg problem of course

faho:
Closing this for now, I think tok_command isn't the right thing to use here as it'll fail for decorators (`command git` triggers abbreviations for `command`, not `git`). Frankly I'm not sure if it's not broken everywhere, seems like a footgun.

Issue #3857, opened 2017-02-19
"NameError: name 'httpd' is not defined" (w/ IPv6 partially disabled)


nullproem:
**OS/terminal used**:
`$ uname -oivrs`
Returns: `Linux 3.16.0-4-amd64 #1 SMP Debian 3.16.39-1 (2016-12-30) unknown GNU/Linux`

GNU Emacs 25.1 "ansi-term"
`$ echo "$TERM"`
Returns: `eterm-color`

Note, I have tried both xterm and urxvt as well with the same result.

**Error message returned when** `fish_config` **is invoked**:
`[Errno 97] Address family not supported by protocol`
`Web config started at 'file:///home/redacted/.cache/fish/web_config-VJB6UM.html'. Hit enter to stop.`
`Traceback (most recent call last):`
`  File "/usr/share/fish/tools/web_config/webconfig.py", line 1135, in <module>`
`    ready_read = select.select([sys.stdin.fileno(), httpd.fileno()], [], [])`
`NameError: name 'httpd' is not defined`

This bug seems to be very similar to issue [#3530](https://github.com/fish-shell/fish-shell/issues/3530) which was ostensibly fixed in 2.5.0. However, for me the ability to use `fish_config` has regressed in the transition from fish 2.4.0 to 2.5.0; 2.4.0 being the version I was last able to use the web-based configuration.

**Relevant system information**:
in `/etc/default/grub` I have ipv6 disabled (but not completely):
`GRUB_CMDLINE_LINUX_DEFAULT="ipv6.disable=1 quiet"`

`$ python -c "from __future__ import print_function; import platform, socket; print(platform.python_version(), '-', socket.has_ipv6);"`
Returns:  `2.7.9 - True`

`$ fish --version`
Returns: `fish, version 2.5.0`

`$ getent hosts localhost`
Returns: `::1             localhost ip6-localhost ip6-loopback`

`$ sh -c 'env HOME=$(mktemp -d) fish'`
Returns the same errorful output as above with only the web_config cache file name changing.

faho:
I can reproduce this, and I don't get why python says ipv6 is available when it is not.

However, I can only recommend that you _don't_ disable ipv6 this way (if you have to at all) since it will cause more issues than just this one (though this is the only one you are likely to have with fish). See https://wiki.archlinux.org/index.php/IPv6#Disable_IPv6. Use `ipv6.disable_ipv6=1` instead.

[...]

PR #9416, opened 2022-12-15
Update webconfig.py to correctly detect IPv6 support.

deltaclock:
## Description

A quick search on the issues shows many users reporting this problem.

Currently the code checks if IPv6 is enabled on the system by using `socket.has_ipv6`. This variable is a compile time constant meaning that Python is compiled with IPv6, **not** that the system currently has support for it.

The common workaround for this was to actually try to bind to an IPv6 access catching the errors. This error checking functionality seems to have been added to python as can be seen [here](https://github.com/python/cpython/blob/b430399d41fa88e9040cd055e55cf9211bf63c61/Lib/socket.py#L870).
My patch simply replaces `socket.has_ipv6` with `socket.has_dualstack_ipv6()`.

A more detailed overview can also be found in https://github.com/urllib3/urllib3/pull/611#issuecomment-100954017.

Fixes issue #3857

## System Info

Latest arch kernel with IPv6 totally disabled (with `ipv6.disable=1`).
Python version 3.10.8

## TODOs:

### DIFF
--- a/share/tools/web_config/webconfig.py
+++ b/share/tools/web_config/webconfig.py
@@ -942,7 +942,7 @@ class FishConfigTCPServer(SocketServer.ThreadingMixIn, SocketServer.TCPServer):
 
     WHITELIST = set(["::1", "::ffff:127.0.0.1", "127.0.0.1"])
 
-    address_family = socket.AF_INET6 if socket.has_ipv6 else socket.AF_INET
+    address_family = socket.AF_INET6 if socket.has_dualstack_ipv6() else socket.AF_INET
 
     def verify_request(self, request, client_address):
         return client_address[0] in FishConfigTCPServer.WHITELIST
@@ -1664,7 +1664,7 @@ authkey = binascii.b2a_hex(os.urandom(16)).decode("ascii")
 
 # Try to find a suitable port
 PORT = 8000
-HOST = "::" if socket.has_ipv6 else "localhost"
+HOST = "::" if socket.has_dualstack_ipv6() else "localhost"
 while PORT <= 9000:
     try:
         Handler = FishConfigHTTPRequestHandler


share/tools/web_config/webconfig.py
```diff
@@ -1664,7 +1664,7 @@ def log_error(self, format, *args):
 
 # Try to find a suitable port
 PORT = 8000
-HOST = "::" if socket.has_ipv6 else "localhost"
+HOST = "::" if socket.has_dualstack_ipv6() else "localhost"
```
faho, line 1667: I'm not sure this is correct (but then python's handling here is extremely confusing so it is possible I'm wrong).

It tells us if "the platform supports creating a SOCK_STREAM socket which can handle both AF_INET and AF_INET6 (IPv4 / IPv6) connections". That's not really something we're interested in.

What we *want* is to use one of the two, not both.

We're already trying to do this by checking if we get "EAFNOSUPPORT" - see below.

It appears what we need to do in addition to that is to check for a ValueError, as python might error out before:

https://github.com/python/cpython/blob/b430399d41fa88e9040cd055e55cf9211bf63c61/Lib/socket.py#L899-L900

(this is an extremely crappy error because instead of giving a specific value like "EAFNOSUPPORT" meaning "I do not support this address family" we now get a flat "This value ain't it, chief")

faho: Okay, first of all: Can you confirm that webconfig.py, as it currently stands in master (*not* 3.5.1, this isn't released yet) since 47ae6bc746c59020f1b3e341fd5a2445d7c497d3, does not work for you?

Because #3857 was closed because of that, and I did test at that time, and it appeared to work.

To test this you can also run `python $PWD/webconfig.py` without installing fish from master (it needs the absolute path tho, hence the $PWD).

>Even if the case might be very rare (especially for localhost) I suggest we replace my dualstack calls with a similar function like:

Please see https://github.com/fish-shell/fish-shell/blob/43d618e51f110dd370f112463ce8034eef889d26/share/tools/web_config/webconfig.py#L1674-L1682. That is almost exactly what we are doing now!

However: From reading the python code you linked, it seems python versions that have the "dualstack" methods can also throw a ValueError instead of the OSError with EAFNOSUPPORT. So all we would have to do is to *also* catch a ValueError *iff* that actually happens.

The issue with has_dualstack_ipv6() is that it is an additional method that I'm not sure does the actual right thing, and that it appears to have been introduced in python 3.7, while the rest of this works with older versions as well (you would be surprised at how many bug reports we get by people using new fish on otherwise ancient systems - RHEL exists, and people do *not* upgrade quickly).

In fact whether it is "dualstack" isn't something we really care about - we only want to know if we can use an ipv6 or an ipv4 address. We don't need a socket that can do both, which is what the "dualstack" thing is apparently about.

>Also I would like to remove the "localhost" string

Sure.

deltaclock: Okay yeah, the master is working for me. There could be some improvements but it doesn't really matter.

Also about the ValueError, I don't think you should worry about it since it needs to be enabled [in](https://github.com/python/cpython/blob/b430399d41fa88e9040cd055e55cf9211bf63c61/Lib/socket.py#L898) which unless I'm missing something, its not.

I agree with the rest, you can close this PR. Sorry for occupying you.

deltaclock:
Bug was already fixed.

PR #9419, opened 2022-12-19
abbr: Let --function use a mandatory argument

faho:
This now means `abbr --add` has two modes:

```fish
abbr --add name --function foo --regex regex
```

```fish
abbr --add name --regex regex replacement
```

This is because `--function` was seen to be confusing as a boolean flag.

Note: I am not sure this is a good idea. I think the example of

```fish
abbr -a --position anywhere --function replace_history -- !!
```

is sufficiently awkward that I think this might just be fixable with documentation changes instead.

Basically, instead of documenting `abbr -a !! --position anywhere --function last_history_item`, document `abbr -a !! --function --position anywhere  last_history_item` or `abbr -fa !! --position anywhere last_history_item`

## TODOs:

### DIFF
--- a/doc_src/cmds/abbr.rst
+++ b/doc_src/cmds/abbr.rst
@@ -9,8 +9,7 @@ Synopsis
 .. synopsis::
 
     abbr --add NAME [--position command | anywhere] [-r | --regex PATTERN]
-                    [--set-cursor[=MARKER]]
-                    [-f | --function] EXPANSION
+                    [--set-cursor[=MARKER]] ([-f | --function FUNCTION] | EXPANSION)
     abbr --erase NAME ...
     abbr --rename OLD_WORD NEW_WORD
     abbr --show
@@ -43,7 +42,7 @@ Abbreviations may be added to :ref:`config.fish <configuration>`.
 .. synopsis::
 
     abbr [-a | --add] NAME [--position command | anywhere] [-r | --regex PATTERN]
-         [--set-cursor[=MARKER]] [-f | --function] EXPANSION
+         [--set-cursor[=MARKER]] ([-f | --function FUNCTION] | EXPANSION)
 
 ``abbr --add`` creates a new abbreviation. With no other options, the string **NAME** is replaced by **EXPANSION**.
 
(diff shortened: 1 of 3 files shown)


faho:
Note: Because this uses a required_argument, using `--function` without one is an unknown option. This means

```fish
abbr --query banana --function
```

says "unknown option: --function" instead of "--function requires --add". I think this needs to be fixed in getopt.

krobelus:
I have a slight preference for this change but no compelling argument. I'm fine with either outcome.

ridiculousfish:
This LGTM, I also prefer this. You can detect the missing argument by adding a ':' to the shortopts string after the '-', and then adding a case for `:`:

```patch
diff --git a/src/builtins/abbr.cpp b/src/builtins/abbr.cpp
index a280512dd..323716263 100644
--- a/src/builtins/abbr.cpp
+++ b/src/builtins/abbr.cpp
@@ -292,17 +292,17 @@ maybe_t<int> builtin_abbr(parser_t &parser, io_streams_t &streams, const wchar_t
     const wchar_t *cmd = argv[0];
     abbr_options_t opts;
     // Note 1 is returned by wgetopt to indicate a non-option argument.
     enum { NON_OPTION_ARGUMENT = 1, SET_CURSOR_SHORT, RENAME_SHORT };
 
     // Note the leading '-' causes wgetopter to return arguments in order, instead of permuting
     // them. We need this behavior for compatibility with pre-builtin abbreviations where options
     // could be given literally, for example `abbr e emacs -nw`.
-    static const wchar_t *const short_options = L"-af:r:seqgUh";
+    static const wchar_t *const short_options = L"-:af:r:seqgUh";
     static const struct woption long_options[] = {
         {L"add", no_argument, 'a'},         {L"position", required_argument, 'p'},
         {L"regex", required_argument, 'r'}, {L"set-cursor", optional_argument, SET_CURSOR_SHORT},
         {L"function", required_argument, 'f'},    {L"rename", no_argument, RENAME_SHORT},
         {L"erase", no_argument, 'e'},       {L"query", no_argument, 'q'},
         {L"show", no_argument, 's'},        {L"list", no_argument, 'l'},
         {L"global", no_argument, 'g'},      {L"universal", no_argument, 'U'},
         {L"help", no_argument, 'h'},        {}};
@@ -386,16 +386,20 @@ maybe_t<int> builtin_abbr(parser_t &parser, io_streams_t &streams, const wchar_t
             case 'g':
             case 'U':
                 // Kept for backwards compatibility but ignored.
                 break;
             case 'h': {
                 builtin_print_help(parser, streams, cmd);
                 return STATUS_CMD_OK;
             }
+            case ':': {
+                builtin_missing_argument(parser, streams, cmd, argv[w.woptind - 1]);
+                return STATUS_INVALID_ARGS;
+            }
             case '?': {
                 builtin_unknown_option(parser, streams, cmd, argv[w.woptind - 1]);
                 return STATUS_INVALID_ARGS;
             }
         }
     }
     opts.args.insert(opts.args.end(), argv + w.woptind, argv + argc);
     if (!opts.validate(streams)) {
```

faho:
Okay, merged with the better error.

Note that ideally `abbr --query banana --function` would error that `--function` isn't a usable option for `--query`, but I don't know how to get getopt to give me the detected option even in case of an error (because otherwise it could also be `abbr -rfq` and the problem is with the `f`).

PR #9424, opened 2022-12-21
completions/git: (git log) show refs before than files

NextAlone:
Signed-off-by: Next Alone <12210746+NextAlone@users.noreply.github.com>

## Description

Show refs before files on git-log completions

Fixes issue #

## TODOs:

### DIFF
--- a/share/completions/git.fish
+++ b/share/completions/git.fish
@@ -1400,8 +1400,8 @@ complete -f -c git -n '__fish_git_using_command init' -l bare -d 'Create a bare 
 ### log
 complete -c git -n __fish_git_needs_command -a shortlog -d 'Show commit shortlog'
 complete -c git -n __fish_git_needs_command -a log -d 'Show commit logs'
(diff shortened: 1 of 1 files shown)


faho:
This is entirely subjective, and in many cases unfiltered completions here are going to be unhelpful whatever we do because it's too much.

The best data point I have here is that most of my own `git log` uses are for branches or tags. Neither order is particularly *wrong* and I don't see a good argument for either, and if we made this change someone else could just come and ask to have it swapped again, and they wouldn't be really "wrong" as such either.

So I'm rejecting this. Sorry.

NextAlone:
but now branches appear among commits，not entirely after them，this maybe a bug

faho:
I don't think I'm seeing what you are seeing, can you open an issue with some more explanation? It doesn't seem like the sort of thing that should be worked around by swapping completions around in any case. Also, if that was the motivation, that should have been in the commit or PR message.

PR #9451, opened 2023-01-06
history pager: allow filtering by multiple words

krobelus:
When using Control-R to search command history, I tend to not navigate the
pager. Instead I like to keep typing the query until my desired command is
on top.

This strategy is awkward when the query is part of a long string that is
shared by many commands.  Let's improve this scenario by allowing arbitrary
gaps in between words in the search string.

This interprets "foo bar" as a glob-search for "foo*bar".  This feels a
bit weird because this syntax has no precedent in fish or other shells. Zsh
supports various globbing syntaxes instead.  We could use globbing syntax too,
but this approach has some upsides:
1. Space is easier to type
2. Space splitting is used in most other search prompts
3. glob syntax would make it awkward to search for actual globs in the
   command history ("to  find `**` type `\*\*`"). I don't think I ever want
   to search for spaces.

### DIFF
--- a/src/history.cpp
+++ b/src/history.cpp
@@ -184,6 +184,14 @@ bool history_item_t::matches_search(const wcstring &term, enum history_search_ty
         case history_search_type_t::prefix: {
             return string_prefixes_string(term, content_to_match);
         }
+        case history_search_type_t::glob: {
+            wcstring wcpattern = parse_util_unescape_wildcards(term);
+            if (wcpattern.empty())
+                wcpattern = ANY_STRING;
+            else if (wcpattern.front() != ANY_STRING && wcpattern.back() != ANY_STRING)
+                wcpattern.push_back(ANY_STRING);
+            return wildcard_match(content_to_match, wcpattern);
+        }
         case history_search_type_t::contains_glob: {
             wcstring wcpattern1 = parse_util_unescape_wildcards(term);
             if (wcpattern1.front() != ANY_STRING) wcpattern1.insert(0, 1, ANY_STRING);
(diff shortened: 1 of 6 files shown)


faho:
So:

First of all, let's mark this for *after* 3.6.0.

Secondly: What happens if you search `git checkout`? That would, of course, also match `git checkout Integration_3.6.0`, which would presumably be what you want, but it could also overmatch?

How does that trade-off work in practice?

My gut feeling is that globs are reasonably intuitive, don't have overmatching because you chose to match more, and given that our glob syntax has one symbol (`*`), I don't think escaping is a big issue?

>I don't think I ever want to search for spaces.

Do you never look for `git checkout`? On the other hand I can't recall searching for `*`?

faho:
Okay, how about this:

1. When we pre-load the filter from the commandline, escape all globs in there
2. Add explicit `*` at the beginning and the end

So, say you type `for i in *` into the commandline and then press ctrl-r. This turns into

`*for i in \**`

in the pager search bar, with the cursor positioned *before* the last `*`. Removing the `*` means you search for the string anchored at the beginning/end - these `*` actually mean something.

That makes globs discoverable.

krobelus:
the discoverability is indeed nice. I'll try this for a while.

> Do you never look for git checkout? On the other hand I can't recall searching for *?

right, searching for space is more common than `*`

faho:
One idea is to escape `*` by doubling them up, this avoids the layers-of-escaping problem, at the cost of turning `**` recursive globs into `****`. Also it's not as discoverable (maybe a note behind the search field?).

(of course we might need escaping here for `\n` if you want to search for a multiline thing?)

krobelus:
I think https://github.com/fish-shell/fish-shell/pull/9476 better addresses my use cases, closing in favor of that one

PR #9458, opened 2023-01-09
Make history pager use more entries

faho:
Like I mentioned in https://github.com/fish-shell/fish-shell/pull/9089, 12 entries is a bit few.

So, instead, we do like we do for completions before disclosing and
pick half the screen (but at least X, in this case 12).

This avoids filling the entire screen, and will avoid an unsightly "X
more entries" (which requires scrolling down to fully disclose)
because it matches what the pager does.

Note: For multiline commands we can be pushed further upwards, and in
case of a multi-column layout we could fit more lines. That would
require asking the pager to fit as many as possible and give us back
the index of the last matching entry and rewinding the history search.

That's gonna be left as an exercise for later if it turns out to be necessary.

### DIFF
--- a/src/reader.cpp
+++ b/src/reader.cpp
@@ -1277,9 +1277,13 @@ static history_pager_result_t history_pager_search(const std::shared_ptr<history
                                                    history_search_direction_t direction,
                                                    size_t history_index,
                                                    const wcstring &search_string) {
-    // Use a small page size because we don't always offer practical ways to select a item from
-    // a large page (since we don't support subsequence filtering here).
-    constexpr size_t page_size = 12;
+    // Limit the number of elements to half the screen like we do for completions
+    // Note that this is imperfect because we could have a multi-column layout.
+    //
+    // We can still push fish further upward in case the first entry is multiline,
+    // but that can't really be helped.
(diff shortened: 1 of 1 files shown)


src/reader.cpp
```diff
@@ -1277,9 +1277,14 @@ static history_pager_result_t history_pager_search(const std::shared_ptr<history
...
                                                    const wcstring &search_string) {
-    // Use a small page size because we don't always offer practical ways to select a item from
-    // a large page (since we don't support subsequence filtering here).
-    constexpr size_t page_size = 12;
+    // Limit the number of elements to half the screen like we do for completions
+    // Note that this is imperfect because we could have a multi-column layout.
+    //
+    // We can still push fish further upward in case the first entry is multiline,
+    // but that can't really be helped.
+    // (subtract 2 for the search line and the prompt)
```
krobelus, line 1285: yeah this behavior is definitely much nicer, thanks.
My original concern was that it can be difficult to select things in the pager.
I think that's not really relevant in practice because most searches can easily be refined to match fewer commands.
Also we'll add globbing soon.

krobelus:
random future ideas: use the alternate screen like fzf
perhaps with a single-column layout if there's enough space, because single-column a bit easier to scan

PR #9460, opened 2023-01-10
webconfig: Remove the abbreviations tab

faho:
Since the new expanded abbreviations in 3.6.0, abbr no longer accepts new universal variables. That means this tab is now non-functional (except that it could technically remove abbrs that were set in universal variables).

Because making it work with the expanded abbreviations requires some awkwardness like a dedicated conf.d snippet (or writing into config.fish!), we simply remove it.

## TODOs:

---

If someone wants to make this functional with the new abbreviations, be my guest - but until that happens it's better to not show this because it's almost entirely non-functional. If you want to bring it back the source is always in git.


ridiculousfish:
Yep this was an oversight. Thanks for fixing this. I agree with 3.6.1 milestone.

ridiculousfish (approved):

PR #9469, opened 2023-01-13
docs: clearify global vs universal variable

shenlebantongying:
I got asked about them. It turns out that without much computer knowledge, those sentences are mega confusing. 

The weirdness of the original wording:

> Global variables don't disappear

What does this even mean? If it doesn't disappear, will it exist after the computer shuts down???

> Global variables ....... and will never be erased .....

So there is no difference between Global & Universal since Global never be erased and it also lives forever???

This is wrong. A global variable can be erased by other means (rather than NEVER): quit the current session.

### DIFF
--- a/doc_src/cmds/set.rst
+++ b/doc_src/cmds/set.rst
@@ -51,8 +51,8 @@ The following scope control variable scope:
 
 **-g** or **--global**
     Sets a globally-scoped variable.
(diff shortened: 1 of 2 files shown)


faho:
Thanks, that's much better (no idea how I came up with "They can *even* be modified" tbh)

PR #9473, opened 2023-01-14
Remove str2wcs special case for MB_CUR_MAX

### LINKED ISSUE #2613: Variables saved with C locale are broken in other locales (was "iTerm2: Wrong $PWD after locale is changed")
Already discussed in another issue: https://gitlab.com/gnachman/iterm2/issues/4083
But it seems more like a problem from fish.

Steps to reproduce:
1. iTerm2 Preferences > Profiles > General > Working Directory: Reuse previous session's directory
2. In config.fish, export LANG="en_US.UTF-8"
3. `cd /path/to/Äpfel; and echo $PWD` outputs `/path/to/Äpfel`
4. Switch to a new tab by pressing <kbd>command</kbd>-<kbd>T</kbd>
5. In the new tab, `echo $PWD`, gets `/path/to/AÌpfel`
6. `pwd` outputs: `/path/to/Äpfel` (normal)

I suspect this is related to Unicode Normalization Form Decomposition (NFD). And here are some JavaScript for comparison:

``` js
$ encodeURI('Äpfel'.normalize('NFD'))
"A%CC%88pfel"
$ decodeURI(encodeURI('Äpfel'.normalize('NFD')))
"Äpfel"
$ unescape(encodeURI('Äpfel'.normalize('NFD')))
"AÌpfel"
```

and two fish commands to illustrate the mechanism of `unescape` above:

``` fish
$ printf '\xcc\x88'
 ̈
$ printf '\u00cc\u0088'
Ì
```

faho:
This meant we didn't actually do our weird en/decoding scheme for e.g. a C locale, which meant that, when you then switch to a proper locale the previous variables were broken.

I don't know how to test this automatically - none of my attempts seem to ever *fail* with the old code, here's what you'd do manually:

- Run fish with an actual C locale (LC_ALL=C fish_allow_singlebyte_locale=1 fish)
- `set -gx foo 💩`
- `set -e LC_ALL`
- `echo $foo` outputs "💩" if it works and "ð⏎" if it's broken.

Fixes #2613

## TODOs:

### DIFF
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -303,16 +303,6 @@ static wcstring str2wcs_internal(const char *in, const size_t in_len) {
     wcstring result;
     result.reserve(in_len);
 
-    // In the unlikely event that MB_CUR_MAX is 1, then we are just going to append.
-    if (MB_CUR_MAX == 1) {
-        size_t in_pos = 0;
-        while (in_pos < in_len) {
-            result.push_back(static_cast<unsigned char>(in[in_pos]));
-            in_pos++;
-        }
-        return result;
-    }
-
     size_t in_pos = 0;
     mbstate_t state = {};
     while (in_pos < in_len) {


ridiculousfish (approved):
This seems reasonable to me. We should be moving towards using Unicode internally anyways.

PR #9475, opened 2023-01-14
abbr: Warn when `-U` is given

faho:
This prints a warning to stderr and then still does the thing.

Because of the error trailer, it points to the abbr help page.

## TODOs:

- [ ] Wording? I tried to keep it reasonably general so it stands a chance of being reused, maybe something more specific is useful?
- [ ] This can now warn on every startup, if you have `abbr -U` in your config.fish. I don't believe there's anything better we can do (we don't want to store a universal variable or write a file to mark that we've errored)

### DIFF
--- a/src/builtins/abbr.cpp
+++ b/src/builtins/abbr.cpp
@@ -393,9 +393,15 @@ maybe_t<int> builtin_abbr(parser_t &parser, io_streams_t &streams, const wchar_t
                 opts.list = true;
                 break;
             case 'g':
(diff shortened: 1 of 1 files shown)


faho:
The big problem here is that it has no chance of warning in what might be the most common case - `abbr` *without* an explicit scope (which was an implicit universal), so I'm not sure this is worth it.

ridiculousfish (approved):
This can't hurt at least.

Issue #1874, opened 2015-01-05
Fuzzy history command match


iven:
It may be useful if we can use fuzzy match when searching history commands(using the UP arrow).

Say we have command history like this:

```
scp 1.txt myhost1:/tmp
ssh myhost1
scp 2.txt myhost2:/root
ssh myhost123
...
```

Currently it's hard to match the first command. If we have fuzzy match, we can just use:

```
$ scphost1<UP ARROW>
```

to match it.

Also it would be great if we exactly match the whole string first, and use fuzzy match as backup.

And maybe we need an UI menu like [ctrlp.vim](https://github.com/kien/ctrlp.vim).

terlar:
For now there is a third party util that will achieve this for you, that comes with fish integration:
https://github.com/junegunn/fzf

iven:
@terlar Seems it doesn't support history completion...

terlar:
It does:
https://github.com/junegunn/fzf/blob/master/install#L305-L310

It binds it to a ctrl-r binding by default which can be modified to your liking.

iven:
Thanks, I didn't see that in README. I'll try it.

2015-01-07 15:58 GMT+08:00 Terje Larsen <notifications@github.com>:

karolyi:
+1

c02y:
It is May 24, 2017 now, it seems this feature is not implemented yet.

krader1961:
> It is May 24, 2017 now, it seems this feature is still not implemented yet.

Patches are welcomed and will be reviewed in a timely manner. Alternatively, I'm starting to think we should setup a [open source code bounty](https://en.wikipedia.org/wiki/Open-source_bounty) for fish. That way people who really want some feature implemented but lack the skills to do so themselves can entice others to work on the feature.

c02y:
Good idea.

ElijahLynn:
Would love this feature, one of the biggest things I find myself wanting on an hourly basis! Love the idea of open source code bounty for fish. www.bountysource.com is an option.

NotWearingPants:
There's exactly this, but only for bash & zsh:
https://github.com/dvorka/hstr

ElijahLynn:
This has been working for me, but I do wonder if I installed 

> It does:
> https://github.com/junegunn/fzf/blob/master/install#L305-L310
> 
> It binds it to a ctrl-r binding by default which can be modified to your liking.

[...]

PR #9476, opened 2023-01-15
Add subsequence search type to `history` command

elebow:
## Description

This adds a subsequence search type to the `history` command. It was easier than I expected because wcstringutil already had a `subsequence_in_string()`!

#1874 discussed adding this behavior to `history-pager`, but `history` seemed easier so I figured it was better to start with this. We can do `history-pager` in a followup PR, or add it to this one.

## TODOs:

### DIFF
--- a/src/history.h
+++ b/src/history.h
@@ -57,6 +57,8 @@ enum class history_search_type_t {
     contains_glob,
     /// Search for commands starting with the given glob pattern.
     prefix_glob,
+    /// Search for commands containing the given string as a subsequence
+    contains_subsequence,
     /// Matches everything.
     match_everything,
 };
--- a/src/reader.cpp
+++ b/src/reader.cpp
@@ -1288,11 +1288,21 @@ static history_pager_result_t history_pager_search(const std::shared_ptr<history
     completion_list_t completions;
     history_search_t search{history, search_string, history_search_type_t::contains,
                             smartcase_flags(search_string), history_index};
-    while (completions.size() < page_size && search.go_to_next_match(direction)) {
+    bool next_match_found = search.go_to_next_match(direction);
+    if (!next_match_found) {
+        // If there were no matches, try again with subsequence search
+        search =
+            history_search_t{history, search_string, history_search_type_t::contains_subsequence,
+                             smartcase_flags(search_string), history_index};
+        next_match_found = search.go_to_next_match(direction);
+    }
+    while (completions.size() < page_size && next_match_found) {
         const history_item_t &item = search.current_item();
         completions.push_back(completion_t{
             item.str(), L"", string_fuzzy_match_t::exact_match(),
             COMPLETE_REPLACES_COMMANDLINE | COMPLETE_DONT_ESCAPE | COMPLETE_DONT_SORT});
+
+        next_match_found = search.go_to_next_match(direction);
     }
     size_t last_index = search.current_index();
     if (direction == history_search_direction_t::forward)
--- a/src/history.cpp
+++ b/src/history.cpp
@@ -195,6 +195,9 @@ bool history_item_t::matches_search(const wcstring &term, enum history_search_ty
             if (wcpattern2.back() != ANY_STRING) wcpattern2.push_back(ANY_STRING);
             return wildcard_match(content_to_match, wcpattern2);
         }
+        case history_search_type_t::contains_subsequence: {
+            return subsequence_in_string(term, content_to_match);
+        }
         case history_search_type_t::match_everything: {
             return true;
         }
--- a/src/wcstringutil.cpp
+++ b/src/wcstringutil.cpp
@@ -119,7 +119,7 @@ bool string_suffixes_string_case_insensitive(const wcstring &proposed_suffix,
 
 /// Returns true if needle, represented as a subsequence, is contained within haystack.
 /// Note subsequence is not substring: "foo" is a subsequence of "follow" for example.
-static bool subsequence_in_string(const wcstring &needle, const wcstring &haystack) {
+bool subsequence_in_string(const wcstring &needle, const wcstring &haystack) {
     // Impossible if needle is larger than haystack.
     if (needle.size() > haystack.size()) {
         return false;
--- a/src/wcstringutil.h
+++ b/src/wcstringutil.h
@@ -34,6 +34,9 @@ bool string_suffixes_string_case_insensitive(const wcstring &proposed_suffix,
 bool string_prefixes_string_case_insensitive(const wcstring &proposed_prefix,
                                              const wcstring &value);
 
+/// Test if a string matches a subsequence of another.
+bool subsequence_in_string(const wcstring &needle, const wcstring &haystack);
+
 /// Case-insensitive string search, modeled after std::string::find().
 /// \param fuzzy indicates this is being used for fuzzy matching and case insensitivity is
 /// expanded to include symbolic characters (#3584).


elebow:
I also wonder if it is reasonable to replace `contains` with `subsequence`, since the latter is a superset.

krobelus:
> I also wonder if it is reasonable to replace contains with subsequence, since the latter is a superset.

when a user expects the contains-check, subsequence semantics would give false positives

krobelus:
> I also wonder if it is reasonable to replace contains with subsequence, since the latter is a superset.

when a user expects the contains-check, subsequence semantics would give false positives

elebow:
I removed the `--prefix` argument and used the fallback logic described at https://github.com/fish-shell/fish-shell/issues/1874#issuecomment-1383054739.

Unfortunately, I was unable to get either `history` or `history-pager` to work as desired.

src/builtins/history.cpp
```diff
@@ -260,6 +260,15 @@ maybe_t<int> builtin_history(parser_t &parser, io_streams_t &streams, const wcha
                                  parser.cancel_checker(), streams)) {
                 status = STATUS_CMD_ERROR;
             }
+            bool no_results = true; // TODO how can we find this?
```
elebow, line 263: I don't know how to check whether there were no results. `history_t::search` seems to write directly to an IO stream, and returns `false` only when the user supplies invalid arguments.

src/reader.cpp
```diff
@@ -1288,11 +1288,21 @@ static history_pager_result_t history_pager_search(const std::shared_ptr<history
...
     history_search_t search{history, search_string, history_search_type_t::contains,
                             smartcase_flags(search_string), history_index};
-    while (completions.size() < page_size && search.go_to_next_match(direction)) {
+    bool next_match_found = search.go_to_next_match(direction);
+    if (!next_match_found) {
+        // If there were no matches, try again with subsequence search
+        history_search_t search{history, search_string, history_search_type_t::contains_subsequence,
+                                smartcase_flags(search_string), history_index};
+        next_match_found = search.go_to_next_match(direction);
+        // TODO next_match_found is 1 even if there are no matches
```
elebow, line 1297: `go_to_next_match` seems to return `true` even when there are no matches. I don't know else to determine whether there were no matches.

src/history.h
```diff
@@ -276,6 +278,9 @@ class history_search_t {
     /// Returns the current search result item. asserts if there is no current item.
     const history_item_t &current_item() const;
 
+    /// Returns whether the current item exists, which indicates whether a search had any results.
+    bool current_item_exists() const;
```
elebow, line 282: This was another attempt to check for no matches, but it doesn't seem to be useful because `current_item` is null before `go_to_next_match()` is called, even if there are matches.

src/builtins/history.cpp
krobelus: it's easier if we move this logic further down the call stack:

```diff
diff --git a/src/history.cpp b/src/history.cpp
index c07a7ebb9..e8a7d7913 100644
--- a/src/history.cpp
+++ b/src/history.cpp
@@ -1475,9 +1475,16 @@ static void do_1_history_search(history_t *hist, history_search_type_t search_ty
     history_search_t searcher = history_search_t(hist, search_string, search_type,
                                                  case_sensitive ? 0 : history_search_ignore_case);
+    bool success = false;
     while (!cancel_check() && searcher.go_to_next_match(history_search_direction_t::backward)) {
+        success = true;
         if (!func(searcher.current_item())) {
             break;
         }
     }
+    if (!success && (search_type == history_search_type_t::contains ||
+                     search_type == history_search_type_t::contains_glob)) {
+        do_1_history_search(hist, history_search_type_t::contains_subsequence, search_string,
+                            case_sensitive, func, cancel_check);
+    }
 }
 
```

src/reader.cpp
krobelus: With this small tweak, it seems to work fine for me.
The problem was that the fallback search goes out of scope, so it would always use the substring search.

```diff
diff --git a/src/reader.cpp b/src/reader.cpp
index a3b54126d..174b48cd5 100644
--- a/src/reader.cpp
+++ b/src/reader.cpp
@@ -1327,10 +1327,10 @@ static history_pager_result_t history_pager_search(const std::shared_ptr<history
     bool next_match_found = search.go_to_next_match(direction);
     if (!next_match_found) {
         // If there were no matches, try again with subsequence search
-        history_search_t search{history, search_string, history_search_type_t::contains_subsequence,
-                                smartcase_flags(search_string), history_index};
+        search =
+            history_search_t{history, search_string, history_search_type_t::contains_subsequence,
+                             smartcase_flags(search_string), history_index};
         next_match_found = search.go_to_next_match(direction);
-        // TODO next_match_found is 1 even if there are no matches
     }
     while (completions.size() < page_size && next_match_found) {
         const history_item_t &item = search.current_item();
```

src/history.h
krobelus: current_item is only set after the first go_to_next_match()

elebow:
Thanks for the tips. I think this PR is now working as intended.

However, I have been using this locally and I think I prefer it: https://github.com/elebow/fish-shell/compare/master...elebow:fish-shell:history-subsequence-search-alternative

I believe it offers a few advantages:
1. Keeping `history --contains` and `history --subsequence` separate allows a user to choose which search they want. With the behavior in this PR, there is no way to do _only_ a substring search.

   I haven't personally used `history` yet (as a new fish user), so I don't know if this is important. But since `history` already offers other search methods (`--prefix`, `--exact`), it makes sense to keep `--contains` and `--subsequence` separate. Though maybe not with those names (`--contains` → `--substring` perhaps?).

1. I prefer the behavior of `history-pager` to do only subsequence searching. It includes all of the same results as substring, but the results are strictly ordered by recency. It just seems easier to use, to me—my desired results are at the top of the list more frequently than with the other method.

If the approach in this PR is accepted as the default behavior, I would at least want a configuration variable or alternative input command (`history-pager-subsequence` maybe) for the alternative behavior.

elebow:
I've been mulling the idea of making `history-pager` into a more generic `selection-pager` that just takes an iterator and shows the selection/paging UI.

Then a history pager could be accomplished with something like:

`history | selection-pager`

Or if a user wanted to start with the oldest entry:

`history | sort -r | selection-pager`

Or any other arbitrary list could be passed in (file glob, git branches, ...)

And with `selection-pager` as a its own input function, it makes more sense to allow multiple search strategies:

```
history | selection-pager-prefix
history | selection-pager-subsequence
```

---

I understand that there is some complication in combining input functions and normal commands, but it might be surmountable (with `commandline -f`?). I'm still new to fish and haven't explored this idea much yet.

krobelus:
I like the change to the history pager a lot, I think it's a strict improvement.
(Sadly when building fish in debug mode, it can be quite laggy on a large history but that's no big deal)

We should probably also add the fallback behavior to up-arrow search... though I'm not 100% sure, we can leave that for later.

I don't think we should change the `history` builtin.
As mentioned above, the equivalent behavior can already achieved with globs (the extra keystrokes are no big deal unlike in the interactive search).
Additionally, the behavior change could break existing scripts.

> If the approach in this PR is accepted as the default behavior, I would at least want a configuration variable or alternative input command (`history-pager-subsequence` maybe) for the alternative behavior.

preferrably `history-pager` is good enough for everyone but if it's just one extra special input function `history-pager-subsequence`, we can probably add that. I don't expect anyone wants a "prefix" variant. We might want a `history-pager-token` one though (I think the fzf integration offers something similar).

> 1. I prefer the behavior of `history-pager` to include all subsequence matches, even if a substring match exists. It just seems easier to use, to me.

I see. I like the tiered behavior because it gives better results when typing an exact substring, which users are already used to doing.
The completion pager has the same behavior (see `completion_t::rank()`).
In future we should teach the interactive history searches to factor in the smartcase match-type like that function does.

> `history | selection-pager`

This would lack history-specific stuff like syntax highlighting and future features like the ability to delete the selected history items or a timestamp in the description.

> Or any other arbitrary list could be passed in (file glob, git branches, ...)

for generic menus, fzf should work well

elebow:
Okay, I've removed the changes to the `history` builtin.

The tiered behavior is in this PR, and the "always include everything" behavior is in https://github.com/fish-shell/fish-shell/compare/master...elebow:fish-shell:history-contains-and-subsequence-search. Maybe in a future PR we can add a second `history-pager-subsequence`, or a `fish_` configuration variable.

krobelus:
I like the new behavior, maybe we should add it to a patch release

PR #9477, opened 2023-01-15
completions: Offer ../ and ./ again

faho:
Inadvertently broken in a2d816710fc336e0a964e11b798cd5fb19043760, this made `cd .` no longer offer `cd ../` (same for file completions like `ls .`, which only offers dot-files but no more `ls ../` for `ls ../foo`)

This adds these paths in one specific case rather than changing dir_iter to always go through them. That should be okay.

## Description

Talk about your changes here.

Fixes issue #

## TODOs:

### DIFF
--- a/src/wildcard.cpp
+++ b/src/wildcard.cpp
@@ -858,6 +858,16 @@ void wildcard_expander_t::expand_last_segment(const wcstring &base_dir, dir_iter
     bool is_dir = false;
     bool need_dir = flags & expand_flag::directories_only;
 
+    if (prefix.empty() && flags & expand_flag::for_completions) {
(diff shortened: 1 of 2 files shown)


ridiculousfish:
Nice find. I think rather than manually inserting these paths, it would be better to have an option in `dir_iter_t`'s constructor which determines whether to skip `.` and `..`. That way we are honestly reporting what the filesystem gives us, and call sites will have to consider whether they want `.` and `..`  or not.

ridiculousfish (changes requested):
see comment

ridiculousfish (approved):
LGTM

Issue #9479, opened 2023-01-15
Are any additional argparse validators planned?


mattmc3:
Are there plans to add additional `argparse` validators? As far as I can tell, fish only ships with `_validate_int`. I know users can implement them for themselves, but I think at least a couple more might be commonly used, and helpful to ship with Fish. For example:

- `_validate_file_exists` and its counterpart, `_validate_file_not_exists`
- `_validate_directory_exists` and its counterpart, `_validate_directory_not_exists`
- etc.

While those read nicely, if we dig in a lot of the useful validators would simply be wrappers for `test` functionality, so it might be clever to just provide that instead like so:

```fish
function _validate_with_test --no-scope-shadowing \
    -d "argparse validator for 'test [TEST] FILE'"

    # save test_args
    set -l test_args $argv

    # strip off not operator
    test $argv[1] != "!" || set argv $argv[2..]

    # verify one and only one test operator
    test (count $argv) -eq 1 || return 2

    # argparse file tests
    set -l options 'b' 'c' 'd' 'e' 'f' 'g' 'G' 'k' 'L' 'O' 'p' 'r' 's' 'S' 't' 'u' 'w' 'x'
    argparse -n _validate_with_test $options -- $argv
    or return 2

    # perform test
    if not test $test_args $_flag_value
        set -l msg (_ "%s: Value '%s' for flag '%s' does not pass 'test %s'\n")
        printf $msg $_argparse_cmd $_flag_value $_flag_name "$test_args" >&2
        return 1
    end
end
```

```fish
$ set _flag_value $__fish_conf_dir
$ argparse --name=foo 'f/file=!_validate_with_test -e' -- --file $_flag_value
# success - directory exists
$ argparse --name=foo 'f/file=!_validate_with_test -f' -- --file $_flag_value
# fail! - is a directory, not a regular file
$ argparse --name=foo 'f/file=!_validate_with_test -d' -- --file $_flag_value
# success - is a directory
$ argparse --name=foo 'f/file=!_validate_with_test ! -f' -- --file $_flag_value
# success - is not a regular file
```

One other handy one might be a `string match` regex wrapper - `_validate_string_matches`.

I don't think there's merit in going too crazy adding validators, but I do think at least a few more of these kinds of validators are common enough that they bear discussing inclusion in Fish itself, rather than left as an exercise for the user.

faho:
[...]

PR #9483, opened 2023-01-16
Add argparse validation examples

mattmc3:
## Description

Added examples to the argparse rst file for flag validator usage.

Fixes issue #9479

## TODOs:

- [X] (N/A?) Changes to fish usage are reflected in user documentation/manpages.
- [X] (N/A?) Tests have been added for regressions fixed
- [ ] User-visible changes noted in CHANGELOG.rst (does this change require that?)

### DIFF
--- a/doc_src/cmds/argparse.rst
+++ b/doc_src/cmds/argparse.rst
@@ -163,6 +163,19 @@ The script should write any error messages to stdout, not stderr. It should retu
 
 Fish ships with a ``_validate_int`` function that accepts a ``--min`` and ``--max`` flag. Let's say your command accepts a ``-m`` or ``--max`` flag and the minimum allowable value is zero and the maximum is 5. You would define the option like this: ``m/max=!_validate_int --min 0 --max 5``. The default if you just call ``_validate_int`` without those flags is to simply check that the value is a valid integer with no limits on the min or max value allowed.
 
+Here are some examples of flag validations::
+
+  # validate that a path is a directory
+  argparse 'p/path=!test -d "$_flag_value"' -- --path $__fish_config_dir
+  # validate that a path does not exist or print an error
+  argparse 'p/path=!test ! -e "$_flag_value" || echo "Path exists: $_flag_value"' -- $argv
+  # validate that a function does not exist
+  argparse 'f/func=!not functions -q "$_flag_value"' -- -f alias
+  # validate that a string matches a regex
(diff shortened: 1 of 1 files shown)


doc_src/cmds/argparse.rst
```diff
@@ -163,6 +163,19 @@ The script should write any error messages to stdout, not stderr. It should retu
 
 Fish ships with a ``_validate_int`` function that accepts a ``--min`` and ``--max`` flag. Let's say your command accepts a ``-m`` or ``--max`` flag and the minimum allowable value is zero and the maximum is 5. You would define the option like this: ``m/max=!_validate_int --min 0 --max 5``. The default if you just call ``_validate_int`` without those flags is to simply check that the value is a valid integer with no limits on the min or max value allowed.
 
+Here are some examples of flag validations::
+
+  # validate that a path is a directory
+  argparse 'p/path=!test -d "$_flag_value"' -- --path $__fish_config_dir
+  # validate that a path does not exist or print an error
+  argparse 'p/path=!test ! -e "$_flag_value" || echo "Path exists: $_flag_value"' -- $argv
```
faho, line 171: This doesn't work. It'll print the error (to stdout), but it will still set $_flag_path.

That's because the `echo` exits true, so the entire script returns true.

You *could* abuse precedence:

```fish
argparse 'p/path=!test ! -e "$_flag_value"; or echo "Path exists: $_flag_value" && false' -- -p /etc
```

The `&&` binds to the `echo`, so this is two separate jobs:

```fish
test ! -e "$_flag_value"
```

or

```fish
echo "Path exists: $_flag_value" && false
```

However that's not something we want to encourage, so I think this should simply be left out here.

(there's a point to be made that a way to get argparse to easily error out if a flag fails validation, and to print a proper error message for it would be good. I don't know how that should work tho)

mattmc3: > that's not something we want to encourage, so I think this should simply be left out here.

Good catch. I've removed that example and pushed a new commit with that example removed.

> there's a point to be made that a way to get argparse to easily error out if a flag fails validation, and to print a proper error message for it would be good. I don't know how that should work tho

One idea might have been better to `echo >&2 "Path exists: $_flag_value"` and then argparse could error if the validation writes anything to stderr. Or, something like `echo "Path exists: $_flag_value" && return 1`. That's a problem for another day though. No need to over complicate it - if the script is that long it's in _write a real validator function_ territory anyway.

faho:
Merged, thanks!

Issue #9497, opened 2023-01-19
Add Fossil support to fish_vcs_prompt


nat-418:
Currently, fossil-scm is supported for fish completions, but not in, e.g., the informative-vcs prompt.

fish, version 3.5.1
Linux laptop 5.15.86 #1-NixOS SMP Sat Dec 31 12:14:48 UTC 2022 x86_64 GNU/Linux
xterm-256color
I tried running fish without third-party customizations and it did not affect behavior.

To reproduce, type `fish_vcs_prompt` in a fossil repository checkout directory.

faho:
This is a textbook example of "patches welcome". We're gonna need someone who knows fossil to write something.

nat-418:
I am working on it. For now I am still trying to figure out how fish_vcs_prompt works. Adding support for Fossil should be pretty easy given that we have `fossil info` and `fossil status` that give us all the information we need in a simple key-value format.

faho:
`fish_vcs_prompt` is just a cover function that goes through all the known vcsen.

It simply calls `fish_git_prompt` for git and `fish_hg_prompt` for mercurial.

What you want is this:

A file called `fish_fossil_prompt.fish`, with a function called `fish_fossil_prompt` (i.e. the same name as the file, no .fish ending) that functions as the entry point. You can add more functions inside the file, but that is the one that ends up being called.

Then, you want the function to at the very least have this:

```fish
function fish_fossil_prompt
    if not type -q fossil
        # no fossil installed, we gracefully fail
        return 127
    end
    
    # do whatever to get the fossil info
    set -l info (fossil info)
    # *jazz hands* MAGIC */jazz hands*

    echo -- $fossil_branch $fossil_repostate # I don't know fossil, put whatever useful info here
end
```

You basically want to look at the [hg prompt](https://github.com/fish-shell/fish-shell/blob/master/share/functions/fish_hg_prompt.fish), not the git one because that's overly complicated for legacy reasons (especially the fifty million configuration variables it exposes).

If this prompt is fast enough, we can add it to fish_vcs_prompt to enable it by default. It especially needs to be *very quick* (on the order of single-digit milliseconds) outside of a fossil repo, even if fossil is installed.

[...]

PR #9500, opened 2023-01-20
feat: add support for fossil-scm in prompt

nat-418:
## Description

Add fish prompt function and include it in the vcs prompt function.

Fixes issue #9497

## TODOs:

### DIFF
--- a/share/functions/fish_fossil_prompt.fish
+++ b/share/functions/fish_fossil_prompt.fish
@@ -5,8 +5,8 @@ function fish_fossil_prompt --description 'Write out the fossil prompt'
     end
 
     # Bail if not a fossil checkout
-    if not test -e .fslckout
-        return 127
+    if not fossil ls &> /dev/null
+	return 127
     end
 
     # Parse fossil info
@@ -16,7 +16,7 @@ function fish_fossil_prompt --description 'Write out the fossil prompt'
     string match --regex --quiet \
         '^tags:\s*(?<fossil_tags>.*)$' $fossil_info
 
-    echo -n '['
+    echo -n ' ['
     set_color --bold magenta
     echo -n $fossil_project_name
     set_color normal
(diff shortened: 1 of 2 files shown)


share/functions/fish_fossil_prompt.fish
```diff
@@ -0,0 +1,53 @@
+function fish_fossil_prompt --description 'Write out the fossil prompt'
+    # Bail if fossil is not available
+    if not command -sq fossil
+        return 1
+    end
+
+    # Bail if not a fossil checkout
+    if not test -e .fslckout
```
faho, line 8: This is too strict - like git and hg looking for a .git or .hg directory, fossil looks in parent directories as well - see https://fossil-scm.org/home/doc/tip/www/tech_overview.wiki:

> For Fossil commands that run from within a working checkout, the first thing that happens is that Fossil locates the checkout database. Fossil first looks in the current directory. If not found there, it looks in the parent directory. If not found there, the parent of the parent. And so forth until either the checkout database is found or the search reaches the root of the file system. (In the latter case, Fossil returns an error, of course.) Once the checkout database is located, it is used to locate the repository database. 

So you need to basically do what `fish_print_hg_root` does, except replacing `.hg/dirstate` with `.fslckout` (and *possibly* `_FOSSIL_` - no idea when that name is used).

Alternatively, if `fossil info` is quick enough outside of a repo, say ~5ms (like git), you can remove this check and do

```fish
set -l fossil_info (fossil info)
or return
```

below (possibly with `fossil info 2>/dev/null` if it writes an error).

share/functions/fish_fossil_prompt.fish
```diff
@@ -0,0 +1,53 @@
...
+    echo -n $fossil_project_name
+    set_color normal
+    echo -n ':'
+    set_color --bold yellow
+    echo -n $fossil_tags
+    set_color normal
+
+    # Parse fossil status
+    set -l fossil_status (fossil status)
+    if string match --index --regex --quiet '^ADDED' $fossil_status
```
faho, line 30: I don't know what the `--index` is for, it seems to not do anything.

faho: You can also use glob-mode here:

```fish
if string match --quiet 'ADDED*' $fossil_status
```

which is a bit shorter.

faho (changes requested):
1. The in-repository will fail in subdirectories
2. Some style nits

faho:
As a general question: How long does this take inside of a repo with fossil installed?

share/functions/fish_fossil_prompt.fish
faho: From what I can find `_FOSSIL_` was the old name, and it's been renamed before 2016. So it's fine to ignore for our use, if we need to check for the checkout database ourselves.

nat-418:
> As a general question: How long does this take inside of a repo with fossil installed?
> 
> You can check with `time fish_fossil_prompt`.
```
$ time fish_fossil_prompt
Executed in   84.28 millis    fish           external
   usr time   13.72 millis    3.16 millis   10.56 millis
   sys time    9.85 millis    0.01 millis    9.84 millis

```

nat-418:
> 1. The in-repository will fail in subdirectories
> 
>     2. Some style nits

Thanks for the quick and helpful feedback. I apologize for forgetting to mark this PR as a draft when I opened it.

share/functions/fish_fossil_prompt.fish
nat-418: changed

share/functions/fish_fossil_prompt.fish
nat-418: changed

faho:
Merged, thanks!

For the record `fossil ls` outside of a fossil repo takes ~3-4ms, so it's about as fast as `git`. We *could* use builtins here like hg, but it's a much different trade-off (`hg status` is 150ms outside of a repo, *hot cache*).

nat-418:
Thank you! I love fish!

Issue #2110, opened 2015-06-01
Feature Request: Interactively delete history entries during auto-completion (i.e. via Alt-Del)


bentolor:
Fish auto-completion is great, but the history can be easily bloated with wrong & unwanted entries

I'm aware about `history --contains keyword --delete` and the interactive history editor via `fish_config` but find them rather cumbersome.

It would be great to have i.e. a key combination like `[Alt-Del]` or `[Ctrl-Del]` to remove the current completion proposal from the history

ridiculousfish:
That's a good idea!

krader1961:
The "current completion proposal" is not added to the interactive command history. You have to actually execute a command for it to be added to the history. By that time it's far too late for a key binding to have any effect.

@bentolor Can you provide more details about your scenario? Perhaps by creating a asciinema recording?

floam:
@krader1961 Why would the item need to be added to history again?

floam:
I think maybe there is a misunderstanding with what he's asking for - the proposal he's referring to is the completion that **fish is proposing to the user** that the user would be rejecting, as I read it.

floam:
So I suspect he's wanting to be able to hit alt-del to nuke what he sees suggested from history, or while searching history with the up arrow remove the current entry from history. In either cased doing the equivalent of `history --contains <COMMAND LINE CONTENTS> --delete` would get the user close and show the a list to pick from that would have their command in it and they can delete it.

The interactive deletion is essential because the desired (undesired) history item may not be the first result due to the context-sensitive nature of autosuggestions and the builtin's inability to do queries on the variables.

bentolor:
> So I suspect he's wanting to be able to hit alt-del to nuke what he sees suggested from history, or while searching history with the up arrow remove the current entry from history. 
You nailed it!
````
$ wget http://user:mysecrepass@foo.baz/ ...
                                 # Argh! Missed to type a leading space. Now my pw is in the history
                                 # --- Option #1:
$ [arrow-up]                     # shows last command
$ [Alt-Del]                      # Clears commandline & removes entry from history
                                 # --- Option #2:
$ mysecret[arrow-up]             # shows last command matching string
$ [Alt-Del]                      # Removes entry from history. Suggest next history match
$ mysecret[arrow-up]             # ..nothing happens (last element).
````

ridiculousfish:
[...]

PR #9515, opened 2023-01-29
history pager: delete selected history entry with Shift-Delete

### LINKED ISSUE #9454: Delete an entry from Ctrl-R history
https://github.com/fish-shell/fish-shell/issues/5791 and https://github.com/fish-shell/fish-shell/commit/049104e8df02a989e3c0ad70d2521251192013cd have made it a lot nicer to work from history.

I'd love to see this go one step further, by allowing a history entry to be deleted from the `Ctrl`-`R` history pager. For example, something similar to Chrome's autofill/omnibox dropdowns, where you can type `Shift`-`Del` to delete the currently selected dropdown item.

This would make it super practical to delete a recent item from history if you accidentally pasted a password, without having to clear any more than necessary:

- Ctrl-R
- Down key a few times
- Shift-Del

It would also bring the history page up to feature parity with the `fish_config` UI, which has exactly two clear features: search all items, and delete any single item.

Screenshot 2023-01-07 at 15 20 59

fish 3.6.0
macOS Ventura 13.1

(previously suggested in https://github.com/fish-shell/fish-shell/pull/9089#issuecomment-1199333297)

krobelus:
After accidentally running a command that includes a pasted password, I want
to delete command from history. Today we need to recall or type (part of)
that command and type "history delete".  Let's maybe add a shortcut to do
this from the history pager.

The current shortcut is Shift+Delete. I don't think that's very discoverable,
maybe we should use Delete instead (but only if the cursor is at the end of
the commandline, otherwise delete a char).

Closes #9454


faho:
This seems like a good idea. My question is if we could make the bind function more general and have it delete the current commandline if the pager isn't open?

>I don't think that's very discoverable,
maybe we should use Delete instead

I'm kind of wary of adding something permanent like this to a very simple key press that you might just use to edit. So I think I'd prefer shift-delete for now.

mqudsi:
shift-delete matches the behavior of native Windows autocomplete inputs w/ history enabled, which was later adopted by Chrom(e|ium) and, I think, Firefox.

krobelus:
sounds good, let's stick to Shift+Delete

> My question is if we could make the bind function more general and have it delete the current commandline if the pager isn't open?

Yeah we should make it work at least during an up-arrow search  (where Shift+Delete shall switch to the next older match).
If no search is active, it should maybe show an interactive prompt for selecting commands to delete, based on the current commandline. That's me spitballing, leaving this case for later.

There's a bug left, deletion does not work for space-prefixed commands. No big deal but would be nice to fix.

zanchey:
Are you happy to merge this for now without that fix?

krobelus:
yeah; actually I can't reproduce the problem with space-prefixed ("ephemeral") commands anymore. Looks like we only ever keep one of them in memory, for simplicity of implementation.

bentolor:
After #2110 has been resolved I was looking into this, but I'm not sure if i get this right:

`Shift`+`Del` should delete a whole history entry if the cursor is at EOL, right? If i du an `Arrow`+`Up` navigation and press `Shift`+`Del` it only deletes the last character, but not the line.

Anything I'm missing @krobelus ?

faho:
> Shift+Del should delete a whole history entry if the cursor is at EOL, right?

No.

It should delete the history entry you are selecting in the history pager, the thing you start with ctrl-r.

We've decided not to do it in the commandline for now.

krobelus:
I still think we should try to make it work in the commandline but it's not clear to me what the UI should look like.
Echoing the deleted command is a possibility but isn't it a bit un-fishy? Not sure

bentolor:
Thanks for clarifying, @faho and @krobelus !

Issue #7902, opened 2021-04-06
Command-not-found case has hard-coded behavior


i077:
Running NixOS with fish version 3.2.1.

I have a little [fish script](https://github.com/i077/system/blob/master/bin/%2C) called `,` (comma) that lets me search nixpkgs (nix's repository of packages) for commands that provide a certain binary, then directly run the command inside that package. [Here](https://asciinema.org/a/gbBMqwaNLVXH0JO5UdsOngpyf) is a screencast just to make it clear.

I want to replace my command-not-found handler with a wrapper around this script, so that if fish can't find the command, it will check if nixpkgs has it, and if it does, offer to run it. This way, I can forego prepending `,` to my command altogether and just have all nixpkgs binaries "available" on my shell. So I implemented a new `fish_command_not_found` function that did exactly that. [Here](https://asciinema.org/a/2RLwbTS3jlWm6ZXnQH2W04OnC) is another screencast that tries to do the same thing as the first, but doesn't quite work as expected.

I noticed in the [function](https://github.com/fish-shell/fish-shell/blob/ed51e2baacc47c831a14ed60a9363f5dbb0b7955/src/parse_execution.cpp#L726-L786) for handling command-not-found (cnf), fish is hard-coded to, among other things:
- Give a status code of 127, and
- Redirect all output from the handler to stderr

Fish seems to make the assumption that a cnf handler will just print an error message, possibly suggesting packages with the given command. This breaks what I want to do, because I don't want my handler to just suggest that a package has a certain command available, I want the handler to _run_ the binary inside that package. I recognize this isn't a very common use case for command-not-found behavior, which is why fish is hard-coded to do the above things, so I was wondering if either
1. This behavior could be redefined so it is up to the handler to return 127 and print errors to stderr, or
2. There is another way I could achieve the same effect (i.e. if a command is not in my PATH, run that `comma` script instead)

For reference, a zsh user implemented a cnf handler with equivalent functionality, and is able to pipe commands that weren't in his path with it, probably because zsh does not do such hardcoding and leaves it up to the handler to implement everything. [Here](https://asciinema.org/a/GUtpGgJ88DENxN0517np6kMRw) is a demo of that.

zanchey:
This seems reasonable to me, but some thought would have to be given to backwards-compatibility.

krobelus:
Hmm, maybe the `preexec` hook should be able to modify the commandline.
What you can do now is to rebind Enter to do the rewriting.

[...]

PR #9517, opened 2023-01-31
move fish_command_not_found exit code and redirection handling up to the error handler function

jhheider:
## Description

This moves the choice of redirection to stderr and the 127 exit code up to the error handler function if defined. Fallback code unchanged. Default error handler config is also updated to maintain backwards compatibility. This matches, as best I can tell, the behavior of zsh/bash. fixes https://github.com/teaxyz/cli/issues/344

Fixes issue #7902

## TODOs:

### DIFF
--- a/share/functions/fish_command_not_found.fish
+++ b/share/functions/fish_command_not_found.fish
@@ -13,6 +13,7 @@ end
 
 function __fish_default_command_not_found_handler
     printf (_ "fish: Unknown command: %s\n") (string escape -- $argv[1]) >&2
+    return 127
 end
 
 # If an old handler already exists, defer to that.
--- a/src/parse_execution.cpp
+++ b/src/parse_execution.cpp
@@ -723,7 +723,8 @@ parse_execution_context_t::ast_args_list_t parse_execution_context_t::get_argume
 end_execution_reason_t parse_execution_context_t::handle_command_not_found(
     const wcstring &cmd_str, const ast::decorated_statement_t &statement, int err_code) {
     // We couldn't find the specified command. This is a non-fatal error. We want to set the exit
-    // status to 127, which is the standard number used by other shells like bash and zsh.
+    // status to 127, which is the standard number used by other shells like bash and zsh,
+    // unless there is an explicit command handler, then we can use its exit code.
 
     const wchar_t *const cmd = cmd_str.c_str();
     if (err_code != ENOENT) {
@@ -770,7 +771,6 @@ end_execution_reason_t parse_execution_context_t::handle_command_not_found(
 
     // Redirect to stderr
     auto io = io_chain_t{};
-    io.append_from_specs({redirection_spec_t{STDOUT_FILENO, redirection_mode_t::fd, L"2"}}, L"");
 
     if (function_exists(L"fish_command_not_found", *parser)) {
         buffer = L"fish_command_not_found";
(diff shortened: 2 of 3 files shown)


jhheider:
Looks like I have some tests to fix. But input on whether this is desired/acceptable/usable style is welcome.

tests/checks/command-not-found.fish
```diff
@@ -2,17 +2,16 @@
...
+$fish -C 'function fish_command_not_found; echo cmd-not-found; end' -c "nonexistent-command-1234 1 2 3 4"
+##CHECKERR: cmd-not-found
+#CHECKERR: fish: Unknown command: nonexistent-command-1234
 #CHECKERR: nonexistent-command-1234 1 2 3 4
 #CHECKERR: ^~~~~~~~~~~~~~~~~~~~~~~^
 $fish -C 'function fish_command_not_found; echo command-not-found $argv; end' -c "nonexistent-command-abcd foo bar baz"
-#CHECKERR: command-not-found nonexistent-command-abcd foo bar baz
-#CHECKERR: fish: 
+##CHECKERR: command-not-found nonexistent-command-abcd foo bar baz
+#CHECKERR: fish: Unknown command: nonexistent-command-abcd
```
jhheider, line 14: this is a regression, unfortunately. need to figure out why this particular needed case fails.

tests/checks/command-not-found.fish
```diff
@@ -27,8 +26,7 @@ echo $status
 
 
 { echo; echo }
-# CHECKERR: {{.*}}: Unknown command: '{ echo; echo }'
-# CHECKERR: {{.*}}: '{ ... }' is not supported for grouping commands. Please use 'begin; ...; end'
+# CHECKERR: {{.*}}: Unknown command: '{ echo; echo }''{ ... }' is not supported for grouping commands. Please use 'begin; ...; end'
```
jhheider, line 29: this is also a regression. these lines shouldn't be joined.

tests/checks/command-not-found.fish
faho: This fails because this runs a fish - `$fish` above - that isn't interactive.

So because you only made it execute `fish_command_not_found` if fish is interactive, and this tries to override fish_command_not_found, it no longer works.

I'm not sure if that's okay.

tests/checks/command-not-found.fish
```diff
@@ -2,17 +2,16 @@
 set -g PATH
 $fish -c "nonexistent-command-1234 banana rama"
 #CHECKERR: fish: Unknown command: nonexistent-command-1234
-#CHECKERR: fish: 
 #CHECKERR: nonexistent-command-1234 banana rama
 #CHECKERR: ^~~~~~~~~~~~~~~~~~~~~~~^
-$fish -C 'function fish_command_not_found; echo cmd-not-found; end' -ic "nonexistent-command-1234 1 2 3 4"
-#CHECKERR: cmd-not-found
-#CHECKERR: fish: 
+$fish -C 'function fish_command_not_found; echo cmd-not-found; end' -c "nonexistent-command-1234 1 2 3 4"
```
faho, line 7: Why this change?

src/parse_execution.cpp
```diff
@@ -770,23 +771,25 @@ end_execution_reason_t parse_execution_context_t::handle_command_not_found(
...
-        parser->set_last_statuses(std::move(prev_statuses));
+
+        return end_execution_reason_t::error;
     } else {
+        // Redirect to stderr _if there's no handler_
+        io.append_from_specs({redirection_spec_t{STDOUT_FILENO, redirection_mode_t::fd, L"2"}},
+                             L"");
         // If we have no handler, just print it as a normal error.
         error = _(L"Unknown command:");
         if (!event_args.empty()) {
```
faho, line 795: This needs a newline (after the `if` which github won't allow me to mark):

```c++
error.push_back(L'\n');
```

I believe that's a pre-existing bug that this just happens to run into.

share/functions/fish_command_not_found.fish
```diff
@@ -13,6 +13,7 @@ end
 
 function __fish_default_command_not_found_handler
     printf (_ "fish: Unknown command: %s\n") (string escape -- $argv[1]) >&2
+    return 127
```
faho, line 16: All the *other* handlers should also default to return 127.

tests/checks/command-not-found.fish
jhheider: Yes, I'm not 100% that this doesn't need more tweaks, but I felt it was a start. A bunch of the error responses changed because the error return was moved to the fish function itself, rather than propagating out of the C++ (as far I understood it). Help is absolutely welcome.

PR #9520, opened 2023-02-01
Upgrade GitHub action `dessant/lock-threads` to v4

LingMan:
In v3 several input parameters where renamed and since v4 it requires Node.js 16.

This resolves warnings about Node.js 12 and `set-output` being deprecated and slated for removal in the `Lock threads` workflow.


zanchey:
Thanks!

PR #9528, opened 2023-02-02
Improve fossil prompt execution time

rymrg:
## Improve fossil prompt execution time

This PR improves the execution time of the fish_fossil_prompt by up to 70ms.
It has the following changes from the original prompt introduced in #9500.

1. Instead of calling `fossil ls` to check if we are inside a fossil repository, we call `fossil branch` instead and cache the value as the current branch. This saves the need of finding the branch later.
2. It also notifies the user if some files in the folder are not ignored yet not added to the repository. This is done by adding `--differ` flag.
3. It makes use of the `changes` command instead of `status` which is cheaper to compute (~20ms).

# Discussion:
1. This PR removes the tags from the current branch and instead keeps only the branch. This is the same behavior if the commit is not tagged. If this behavior is wanted, we can restore it possibly using `fossil tag list` for checking if we are in a repository. Another option is to use `fossil status --differ` to check if the repository exist. The output from status, while slower, has the tags.
2. It removes the repository name from the prompt. Calling `fossil info` takes 50ms on a clean repository. 
3. Are there fish built-ins that I am not aware of that could replace the calls to `grep` `awk` and `sort`?
4. The call to `grep` is probably not required with the use of `changes` instead of `status` but I haven't verified it yet.

## TODOs:

### DIFF
--- a/share/functions/fish_fossil_prompt.fish
+++ b/share/functions/fish_fossil_prompt.fish
@@ -32,7 +32,7 @@ function fish_fossil_prompt --description 'Write out the fossil prompt'
 
     echo -n ' ['
     # Disabling color and pager is always a good idea.
-    set -l repo_status (fossil status --differ 2>/dev/null | grep -v '\w:\|^\s' | awk '{print $1}' | sort -u)
+    set -l repo_status (fossil changes --differ 2>/dev/null | grep -v '\w:\|^\s' | awk '{print $1}' | sort -u)
 
     # Show nice color for a clean repo
     if test -z "$repo_status"
--- a/share/functions/fish_fossil_prompt.fish
+++ b/share/functions/fish_fossil_prompt.fish
@@ -34,9 +34,9 @@ function fish_fossil_prompt --description 'Write out the fossil prompt'
     or set -g fish_prompt_fossil_status_modified '*'
     set -q fish_prompt_fossil_status_renamed
     or set -g fish_prompt_fossil_status_renamed '⇒'
-    set -q fish_prompt_fossil_status_deleted '
+    set -q fish_prompt_fossil_status_deleted
     or set -g fish_prompt_fossil_status_deleted '-'
-    set -q fish_prompt_fossil_status_missing '
+    set -q fish_prompt_fossil_status_missing
     or set -g fish_prompt_fossil_status_missing '✖'
     set -q fish_prompt_fossil_status_untracked
     or set -g fish_prompt_fossil_status_untracked '?'
@@ -48,14 +48,18 @@ function fish_fossil_prompt --description 'Write out the fossil prompt'
 
 
 
-    echo -n ' ['
+    echo -n ' ('
+	set_color magenta
+	echo -n "$branch"
+	set_color normal
+	echo -n '|'
+	#set -l repo_status (fossil changes --differ 2>/dev/null | string match -rv '\w:|^\s' | string split " " -f1 | sort -u)
 	set -l repo_status (fossil changes --differ 2>/dev/null | string match -rv '\w:|^\s' | string split " " -f1 | path sort -u)
 
     # Show nice color for a clean repo
     if test -z "$repo_status"
         set_color $fish_color_fossil_clean
-        echo -n "($branch)"'✓'
-        set_color normal
+        echo -n '✔'
 
     # Handle modified or dirty (unknown state)
     else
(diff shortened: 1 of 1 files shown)


share/functions/fish_fossil_prompt.fish
```diff
@@ -4,48 +4,88 @@ function fish_fossil_prompt --description 'Write out the fossil prompt'
...
     echo -n ' ['
-    set_color --bold magenta
-    echo -n $fossil_project_name
-    set_color normal
-    echo -n ':'
-    set_color --bold yellow
-    echo -n $fossil_tags
-    set_color normal
+    # Disabling color and pager is always a good idea.
+    set -l repo_status (fossil changes --differ 2>/dev/null | grep -v '\w:\|^\s' | awk '{print $1}' | sort -u)
```
faho, line 35: This `grep` can be replaced with a `string match`. Possibly something like

```fish
string match -rv '\w:|^\s'
```

(but note I have no idea what the `\w:` is supposed to match and the dialects are slightly different - string is PCRE)

The awk would then probably be `string split " " -f1` and the `sort -u` can just be `path sort -u` (which is in `path` because it knows about sorting by directory or filename, it's perfectly cromulent for sorting ordinary strings excluding NULL-bytes).

faho: It would be helpful to have a sample of the output and what you want to get here.

share/functions/fish_fossil_prompt.fish
```diff
@@ -4,48 +4,88 @@ function fish_fossil_prompt --description 'Write out the fossil prompt'
...
 
     echo -n ' ['
-    set_color --bold magenta
-    echo -n $fossil_project_name
-    set_color normal
-    echo -n ':'
-    set_color --bold yellow
-    echo -n $fossil_tags
-    set_color normal
+    # Disabling color and pager is always a good idea.
```
faho, line 34: I don't know what this comment refers to, is it outdated?

share/functions/fish_fossil_prompt.fish
```diff
@@ -4,48 +4,88 @@ function fish_fossil_prompt --description 'Write out the fossil prompt'
...
+
+    set -g fish_prompt_fossil_status_added '✚'
+    set -g fish_prompt_fossil_status_modified '*'
+    set -g fish_prompt_fossil_status_renamed '⇒'
+    set -g fish_prompt_fossil_status_deleted '-'
+    set -g fish_prompt_fossil_status_missing '✖'
+    set -g fish_prompt_fossil_status_untracked '?'
+    set -g fish_prompt_fossil_status_conflict '×'
+
+    set -g fish_prompt_fossil_status_order added modified renamed deleted missing untracked conflict
```
faho, line 30: Setting these variables this way means there is absolutely no way they can be overridden, which kinda defeats the point of variables.

You probably want

```fish
set -q fish_prompt_fossil_status_order
or set -g fish_prompt_fossil_status_order added modified renamed deleted missing untracked conflict
```

share/functions/fish_fossil_prompt.fish
```diff
@@ -4,48 +4,88 @@ function fish_fossil_prompt --description 'Write out the fossil prompt'
...
+                case 'MISSING'
+                    set -a fossil_statuses missing
+                case 'RENAMED'
+                    set -a fossil_statuses renamed
+                case 'CONFLICT'
+                    set -a fossil_statuses conflict
+            end
+        end
+
+        if string match -qr '^(ADDED|EDITED|DELETD)' $repo_status
```
faho, line 69: Typo: "DELETED"

faho (changes requested):

share/functions/fish_fossil_prompt.fish
rymrg: Seems like it. 
https://github.com/fish-shell/fish-shell/blob/master/share/functions/fish_hg_prompt.fish#L47

share/functions/fish_fossil_prompt.fish
rymrg: Seems good. Is path split slower than sort or is it just my luck with timing?
Output of the `status` command.
```
repository:   /tmp/aaa/foo.fossil
local-root:   /tmp/aaa/foo/
config-db:    /home/user/.fossil
checkout:     c30b371a740b617cc24f660791189c2b647cbcde 2023-02-02 07:26:59 UTC
parent:       57f300be2dbe273717dc10fd5941999ba5388907 2023-02-02 07:00:50 UTC
tags:         trunk
comment:      foo (user: user)
MISSING    a
EXTRA      b
EXTRA      c
ADDED      d
DELETED    e
RENAMED    b  ->  g
EXTRA      t
```

rymrg: I tried to hyperfine the script. Seems like the change is minor and I just had too much noise.

rymrg:
@nat-418: Could you please test these changes as well?

PR #9539, opened 2023-02-04
completions/meson: rewrite meson completions

dseight:
## Description

Rewrite completions for meson to expose meson commands with their options and subcommands. New completions are based on the meson 1.0.

Subcommands [were introduced](https://mesonbuild.com/Release-notes-for-0-42-0.html#make-all-meson-functionality-invocable-via-the-main-executable) in meson 0.42.0 (August 2017), so new completions will only work for versions after 0.42.0. At this moment, even oldstable Debian (buster) [has meson 0.49.2](https://packages.debian.org/buster/meson) — which means it is unlikely someone will be affected.

## TODOs:

### DIFF
--- a/share/completions/meson.fish
+++ b/share/completions/meson.fish
@@ -1,52 +1,304 @@
 # Completions for the meson build system (http://mesonbuild.com/)
 
-set -l basic_arguments \
-    "h,help,show help message and exit" \
-    ",stdsplit,Split stdout and stderr in test logs" \
-    ",errorlogs,Print logs from failing test(s)" \
-    ",werror,Treat warnings as errors" \
-    ",strip,Strip targets on install" \
-    "v,version,Show version number and exit"
+function __fish_meson_needs_command
+    set -l cmd (commandline -opc)
+    set -e cmd[1]
+    argparse -s 'v/version' -- $cmd 2>/dev/null
+    or return 0
+    not set -q argv[1]
+end
 
-set -l dir_arguments \
-    ",localedir,Locale data directory [share/locale]" \
-    ",sbindir,System executable directory [sbin]" \
-    ",infodir,Info page directory [share/info]" \
-    ",prefix,Installation prefix [/usr/local]" \
-    ",mandir,Manual page directory [share/man]" \
-    ",datadir,Data file directory [share]" \
-    ",bindir,Executable directory [bin]" \
-    ",sharedstatedir,Arch-agnostic data directory [com]" \
-    ",libdir,Library directory [system default]" \
-    ",localstatedir,Localstate data directory [var]" \
-    ",libexecdir,Library executable directory [libexec]" \
-    ",includedir,Header file directory [include]" \
-    ",sysconfdir,Sysconf data directory [etc]"
+function __fish_meson_using_command
+    set -l cmd (commandline -opc)
+    set -e cmd[1]
+    test (count $cmd) -eq 0
+    and return 1
+    contains -- $cmd[1] $argv
+    and return 0
+end
 
-for arg in $basic_arguments
-    set -l parts (string split , -- $arg)
-    if not string match -q "" -- $parts[1]
-        complete -c meson -s "$parts[1]" -l "$parts[2]" -d "$parts[3]"
+function __fish_meson_builddir
+    # Consider the value of -C option to detect the build directory
+    set -l cmd (commandline -opc)
+    argparse -i 'C=' -- $cmd
+    if set -q _flag_C
+        echo $_flag_C
     else
-        complete -c meson -l "$parts[2]" -d "$parts[3]"
+        echo .
     end
 end
 
-for arg in $dir_arguments
-    set -l parts (string split , -- $arg)
-    complete -c meson -l "$parts[2]" -d "$parts[3]" -xa '(__fish_complete_directories)'
+function __fish_meson_targets
+    set -l python (__fish_anypython); or return
+    meson introspect --targets (__fish_meson_builddir) | $python -S -c 'import json, sys
+data = json.load(sys.stdin)
+targets = set()
+for target in data:
+    targets.add(target["name"])
+for name in targets:
+    print(name)' 2>/dev/null
 end
 
-complete -c meson -s D -d "Set value of an option (-D foo=bar)"
+function __fish_meson_subprojects
+    set -l python (__fish_anypython); or return
+    meson introspect --projectinfo (__fish_meson_builddir) | $python -S -c 'import json, sys
+data = json.load(sys.stdin)
+for subproject in data["subprojects"]:
+    print(subproject["name"])' 2>/dev/null
+end
 
-complete -c meson -l buildtype -xa 'plain debug debugoptimized release minsize' -d "Set build type [debug]"
-complete -c meson -l layout -xa 'mirror flat' -d "Build directory layout [mirror]"
-complete -c meson -l backend -xa 'ninja vs vs2010 vs2015 vs2017 xcode' -d "Compilation backend [ninja]"
-complete -c meson -l default-library -xa 'shared static both' -d "Default library type [shared]"
-complete -c meson -l warning-level -xa '1 2 3' -d "Warning level [1]"
-complete -c meson -l unity -xa 'on off subprojects' -d "Unity build [off]"
-complete -c meson -l cross-file -r -d "File describing cross-compilation environment"
-complete -c meson -l wrap-mode -xa 'WrapMode.{default,nofallback,nodownload,forcefallback}' -d "Special wrap mode to use"
+function __fish_meson_tests
+    # --list option shows suites in a short form, e.g. if a test "gvariant"
+    # is present both in "glib:glib" and "glib:slow" suites, it will be shown
+    # in a list as "glib:glib+slow / gvariant". So, just filter out the first
+    # part and list all of the test names.
+    meson test -C (__fish_meson_builddir) --no-rebuild --list | string split -r -f1 ' / '
+end
 
-# final parameter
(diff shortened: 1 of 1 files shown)


share/completions/meson.fish
```diff
@@ -1,52 +1,308 @@
...
-    complete -c meson -l "$parts[2]" -d "$parts[3]" -xa '(__fish_complete_directories)'
+function __fish_meson_targets
+    set -l python (__fish_anypython); or return
+    meson introspect --targets (__fish_meson_builddir) | $python -S -c 'import json, sys
+data = json.load(sys.stdin)
+targets = set()
+for target in data:
+    targets.add(target["name"])
+for name in targets:
+    print(name)' 2>/dev/null
```
mqudsi, line 39: This doesn't return as many targets as `complete -C"ninja -C $(__fish_meson_builddir) "` does. Are all of the targets discovered by `ninja` usable in this context? If so, maybe we should prefer that. (`ninja` is also faster than meson (python) + python).

(Ninja includes the automatic targets like `benchmark`, `clean`, `install`, `reconfigure`, etc.)

share/completions/meson.fish
```diff
@@ -1,52 +1,308 @@
...
+data = json.load(sys.stdin)
+suites = set()
+for test in data:
+    suites.update(test["suite"])
+for name in suites:
+    print(name)' 2>/dev/null
+end
+
+function __fish_meson_help_commands
+    meson help --help | string match -g -r '^  {(.*)}' | string split ,
```
mqudsi, line 70: Let's be a bit more flexible and support a variable number of spaces, can you change that to 

```
meson help --help | string match -g -r '^ *{(.*)}' | string split ,
```

share/completions/meson.fish
```diff
@@ -1,52 +1,308 @@
...
 end
 
-complete -c meson -s D -d "Set value of an option (-D foo=bar)"
+# Each meson command and subcommand has -h/--help option
+complete -c meson -s h -l help -d 'Show help'
+
+# global options
+complete -c meson -n __fish_meson_needs_command -s v -l version -d 'Show version number and exit'
+# implicit "meson setup" usage via just "meson" was deprecated a long time ago,
+# so no "meson setup" option completions in global scope
```
mqudsi, line 79: I'm not an expert on what qualifies as "a long time ago" and what doesn't, but FWIW this was news to me. It's going to take me a long time to stop doing `meson build` or `meson --buildtype debug debug`

Some git sleuthing shows the deprecation was only added to the codebase August 2022 and shipped in meson 0.64.0, which didn't ship until November of 2022. Debian stable is shipping with meson 0.56.2, Ubuntu's bleeding edge is shipping meson 0.63 while 22.04 LTS is shipping 0.61, and 20.04 is shipping 0.53.

If these were new completions altogether, I'd say it's fine. But the old completions supported top-level setup (which, deprecated or not, was how meson themselves were using it in the docs for the longest time), so this is actually a regression in our completions.

I would just change `-n '__fish_meson_using_command setup'` to `-n '__fish_meson_needs_command || __fish_meson_using_command setup'` so that you can automatically support both?

mqudsi (commented):
Wow, really nice work. You went all out with providing dynamic completions everywhere, great job!

I left a few comments, if you don't mind taking a look. It all looks good though I do have a personal disagreement with the deprecation of implicit setup for backwards compatibility reasons.

share/completions/meson.fish
```diff
@@ -1,52 +1,308 @@
...
+
+function __fish_meson_builddir
+    # Consider the value of -C option to detect the build directory
+    set -l cmd (commandline -opc)
+    argparse -i 'C=' -- $cmd
+    if set -q _flag_C
+        echo $_flag_C
     else
-        complete -c meson -l "$parts[2]" -d "$parts[3]"
+        echo .
```
mqudsi, line 27: If I remember correctly, meson doesn't support in-tree builds at all so this wouldn't work for commands that need the build directory and don't work without `-C` at all? I guess what I'm asking is are there any commands that work with `-C .` but don't work without `-C` at all? Because the inverse is true (e.g. `meson configure` can be run in a meson project directory without `-C` but fails with `-C .`).

share/completions/meson.fish
eli-schwartz: `meson compile` doesn't support ninja special targets, just compilation build edges.

Note benchmark is really just a type of test, and it's a bit silly for:
- `meson compile reconfigure` to run ninja and have ninja run `meson setup --reconfigure`.
- `meson compile test` to not compile anything, just run ninja and have ninja run `meson test`
- `meson compile install` to not compile anything, just run ninja and have ninja run `meson install`


clean meanwhile is handled via `meson compile --clean`, it's not a valid target (except directly in ninja).

share/completions/meson.fish
eli-schwartz: There's a lot more context in the relevant Meson commit: https://github.com/mesonbuild/meson/commit/3c7ab542c0c4770241eae149b0d4cd8de329aee0

The commit message calls out the fact that since Meson 0.42.0 (which is indeed "a long time ago") `meson builddir/` is silently rewritten to `meson setup builddir/` -- and also observes that the fact that it wasn't deprecated way back in 0.42.0 was an *accident*, because we deprecated the other entry points via wrapper scripts but forgot to do anything to the main executable.

share/completions/meson.fish
dseight: Also, using targets from ninja will not work for backends other than ninja, whereas the current approach does (e.g., it works perfectly for the `xcode` backend).

share/completions/meson.fish
dseight: Sure. I have updated the commit.

share/completions/meson.fish
mqudsi: Sorry, I meant to link that commit in my response.

> [...] and also observes that the fact that it wasn't deprecated way back in 0.42.0 was an accident

"If a tree falls in the forest and no one hears it...."

> So, in fact, bringing setup option completions into a global scope seems to be harmful, as more people might discover via completions that it works and start to use it in this way.

This is a problem for meson, but fish serves to facilitate a *user's* workflow. The commands and options are supported today and people use them today. The interests of the developers of the tools we provide completions for and the interests of our users are almost always in lock-step but sometimes they do not align perfectly.

If you do not wish to add the completions, that's fine. Fish might still support them anyway.

share/completions/meson.fish
mqudsi: @dseight I wasn't necessarily suggesting using ninja as much as I was wondering about the disparity between the target list. But I did forget about the xcode backend.

@eli-schwartz thanks for clarifying.

share/completions/meson.fish
eli-schwartz: I do not care one way or the other (neither for Meson nor on a personal level), I'm just pointing out some context in case you're worried about personally relying on `meson setup`.

> It's going to take me a long time to stop doing `meson build` or `meson --buildtype debug debug`

etc.

mqudsi: I understand and appreciate your point. Thanks for stopping by and providing a definitive answer there and thanks for the work you do on meson!

share/completions/meson.fish
eli-schwartz: Yes, that is why `meson compile` exists, essentially. As a portability wrapper for xcode/msbuild compatibility.

If you're happy with ninja, there's no reason to ever use `meson compile`. But it might be nicer in a beginner's README, or you might have xcode-related needs, and that's fine too.

share/completions/meson.fish
dseight: > This is a problem for meson, but fish serves to facilitate a user's workflow. The commands and options are supported today and people use them today. The interests of the developers of the tools we provide completions for and the interests of our users are almost always in lock-step but sometimes they do not align perfectly.

Yeah, I think that I got your point. I just thought about these completions more from the meson side, not the fish side. If they were in the meson repo, then it would be perfectly fine not to include these global options. But completions are in the fish repo, and fish cannot assume much about which version of meson the user has. So, you are right, and it's better to support them here.

eli-schwartz: @mqudsi

No worries, that's the way I interpreted it too. :)

share/completions/meson.fish
dseight: > `ninja` is also faster than meson (python) + python

That's true. Calling python twice here isn't a good thing. It actually would be nice to have some convenient builtin to parse json in fish someday, maybe with JSONPath syntax. So this could be reduced to something like:
```fish
meson introspect --targets (__fish_meson_builddir) | json path '$.[name]'
```
This way, we could also replace other python invocations in fish, which are used to parse json.

share/completions/meson.fish
mqudsi: Thanks, @dseight. I'm all for being able to deprecate or change interfaces, but we just want to be nice to everyone :)

This looks good. I'll take a final look and merge if I don't notice anything else!

share/completions/meson.fish
mqudsi: Yeah, we've needed to parse JSON in a number of other completions time and time again. I think if we can find a spiffy and cool way of integrating it into the shell without reimplementing all of `jq` it might be a good addition.

mqudsi:
I pushed a commit to use ordered completions (`complete -k[x]a`) to prevent intermixing directories and subcommands to make the output nicer.

before
Screenshot from 2023-02-11 17-35-57@2x



after 

Screenshot from 2023-02-11 17-34-52@2x

dseight:
Thank you, @mqudsi! Completion with command-first ordering looks much better.

PR #9541, opened 2023-02-05
rust: fix clippy warnings, doc comments

Xiretza:
## Description

First of all, I really like the direction that fish is taking here! I'd love to contribute to the porting effort; as a way to get myself familiar with the rust codebase so far, I decided to fix the low-hanging fruit reported by clippy. This should also allow developers to use use clippy for their newly written code without being overwhelmed by unrelated warnings.

If you're still too early in development to worry about such small stuff, that's fine too of course.

Is there any interest in running clippy in CI? I could work on that (as well as things like checking rustfmt, running tests on the minimum supported rust version and with the minimum dependency versions).

## TODOs:

### DIFF
--- a/fish-rust/src/util.rs
+++ b/fish-rust/src/util.rs
@@ -2,6 +2,7 @@
 
 use crate::ffi::wcharz_t;
 use crate::wchar::wstr;
+use std::cmp::Ordering;
 use std::time;
 
 #[cxx::bridge]
@@ -80,15 +81,19 @@ pub fn wcsfilecmp(a: wcharz_t, b: wcharz_t) -> i32 {
         acl = acl.to_uppercase().next().unwrap();
         bcl = bcl.to_uppercase().next().unwrap();
 
-        if acl < bcl {
-            retval = -1;
-            break;
-        } else if acl > bcl {
-            retval = 1;
-            break;
-        } else {
-            ai += 1;
-            bi += 1;
+        match acl.cmp(&bcl) {
+            Ordering::Less => {
+                retval = -1;
+                break;
+            }
+            Ordering::Equal => {
+                ai += 1;
+                bi += 1;
+            }
+            Ordering::Greater => {
+                retval = 1;
+                break;
+            }
         }
     }
 
@@ -152,15 +157,19 @@ pub fn wcsfilecmp_glob(a: wcharz_t, b: wcharz_t) -> i32 {
         // TODO Compare the tail (enabled by Rust's Unicode support).
         let acl = ac.to_lowercase().next().unwrap();
         let bcl = bc.to_lowercase().next().unwrap();
-        if acl < bcl {
-            retval = -1;
-            break;
-        } else if acl > bcl {
-            retval = 1;
-            break;
-        } else {
-            ai += 1;
-            bi += 1;
+        match acl.cmp(&bcl) {
+            Ordering::Less => {
+                retval = -1;
+                break;
+            }
+            Ordering::Equal => {
+                ai += 1;
+                bi += 1;
+            }
+            Ordering::Greater => {
+                retval = 1;
+                break;
+            }
         }
     }
 
--- a/fish-rust/src/util.rs
+++ b/fish-rust/src/util.rs
@@ -13,10 +13,28 @@ mod ffi {
     }
 
     extern "Rust" {
-        fn wcsfilecmp(a: wcharz_t, b: wcharz_t) -> i32;
-        fn wcsfilecmp_glob(a: wcharz_t, b: wcharz_t) -> i32;
+        #[cxx_name = "wcsfilecmp"]
+        fn wcsfilecmp_ffi(a: wcharz_t, b: wcharz_t) -> i32;
+        #[cxx_name = "wcsfilecmp_glob"]
+        fn wcsfilecmp_glob_ffi(a: wcharz_t, b: wcharz_t) -> i32;
         fn get_time() -> i64;
     }
+}
+
+fn ordering_to_int(ord: Ordering) -> i32 {
+    match ord {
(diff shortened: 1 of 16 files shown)


faho:
>If you're still too early in development to worry about such small stuff, that's fine too of course.

It is, yeah.

And in particular: This codebase will be unpalatable to clippy for a bit. For example, we're sticking to the C++ naming conventions for things that are shared with the C++ like https://github.com/fish-shell/fish-shell/pull/9541/files#diff-a9a4039f39e3644ed0f496b6e22b25e9298ea26939f14425bd748f6f26594289L51-R53.

The intention is to port first and then make it nice, so we can more easily compare.

>Is there any interest in running clippy in CI?

Personally, I do not believe that style nits are enough of a deal to fail the tests, so I never saw the appeal of running formatters etc in CI.

Sorry, I'll be closing this because it's a bit early to worry about these things!

Xiretza:
> For example, we're sticking to the C++ naming conventions for things that are shared with the C++

To be clear about that one in particular, "shared with C++" can also mean "already ported to Rust, but also still implemented in C++", right? Or is that struct actually exposed to C++ in a way that I'm not seeing right now?

> The intention is to port first and then make it nice, so we can more easily compare.

I've heard this before, in my experience the "making it nice" part tends to be much harder to apply to an entire not-yet-nice project rather than incrementally while the project is being built. But if line-by-line comparison is a concern, priorities might just be different.

> Personally, I do not believe that style nits are enough of a deal to fail the tests, so I never saw the appeal of running formatters etc in CI.

The primary concern is the same for linters and formatters: contributers want to and should use them, but if the existing codebase produces heaps of warnings/gets changed by the autoformatter, they become essentially useless because it's so hard to filter the signal from the noise.

> Sorry, I'll be closing this because it's a bit early to worry about these things!

No worries, I'll try to get some actual porting done soon.

krobelus:
Most of these are good changes. Making clippy more usable is a great direction. If there's some rule we don't want to adhere to just yet, we can tell clippy to ignore that (which this PR already does for some rules).

fish-rust/src/util.rs
```diff
@@ -80,15 +81,19 @@ pub fn wcsfilecmp(a: wcharz_t, b: wcharz_t) -> i32 {
...
+        match acl.cmp(&bcl) {
+            Ordering::Less => {
+                retval = -1;
+                break;
+            }
+            Ordering::Equal => {
+                ai += 1;
+                bi += 1;
+            }
+            Ordering::Greater => {
```
krobelus, line 93: I discarded this variant because it seemed worse (more lines).
But I didn't realize it's a clippy lint, so this fix is fine.

We could already rewrite the entire function to return `Ordering`, only the ffi shim needs to return `c_int`.

fish-rust/src/wgetopt.rs
```diff
@@ -21,41 +23,39 @@ License along with the GNU C Library; see the file COPYING.LIB.  If
...
+///
+/// The special argument `--` forces an end of option-scanning regardless of the value of
+/// `ordering`.  In the case of RETURN_IN_ORDER, only `--` can cause `getopt` to return EOF with
+/// `woptind` != ARGC.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 enum Ordering {
-    REQUIRE_ORDER,
-    PERMUTE,
-    RETURN_IN_ORDER,
+    RequireOrder,
```
krobelus, line 51: The docstring and the C++ dual still use screaming snake case.

Keeping the existing name is the easiest way to port code but I think we want to adopt Rust conventions.

I have no strong opinion yet on whether to do this now or later (when the port is done).

I'd like to experiment a bit more before deciding this, so maybe instruct clippy to silence this warning?

fish-rust/src/util.rs
krobelus: I think the proposed version is not really better than the old one and the old one matches the old code (w.r.t. order).
How about we disable this warning at crate-level?
We can re-enable it once we move to more idiomatic Rust.

fish-rust/src/wgetopt.rs
Xiretza: Silenced.

fish-rust/src/util.rs
Xiretza: I spent a couple minutes rewriting it to use `Ordering` instead, let me know what you think. I've added it as a separate commit to make it easier to review or revert.

fish-rust/src/util.rs
```diff
@@ -12,12 +13,30 @@ mod ffi {
     }
 
     extern "Rust" {
-        fn wcsfilecmp(a: wcharz_t, b: wcharz_t) -> i32;
-        fn wcsfilecmp_glob(a: wcharz_t, b: wcharz_t) -> i32;
+        #[cxx_name = "wcsfilecmp"]
+        fn wcsfilecmp_ffi(a: wcharz_t, b: wcharz_t) -> i32;
```
krobelus, line 17: nice, I didn't know about `cxx_name`

fish-rust/src/util.rs
```diff
@@ -12,12 +13,30 @@
...
 
+fn ordering_to_int(ord: Ordering) -> i32 {
+    match ord {
+        Ordering::Less => -1,
+        Ordering::Equal => 0,
+        Ordering::Greater => 1,
+    }
+}
+
+fn wcsfilecmp_glob_ffi(a: wcharz_t, b: wcharz_t) -> i32 {
```
krobelus, line 32: It's great that we can make ffi-only functions private

fish-rust/src/util.rs
Xiretza: Took me quite a while to find this in the cxx docs too.

Issue #6575, opened 2020-02-07
Copying a function doesn't set the line number


faho:
Define func.fish with the following contents:

```fish
function t1
    echo t1
end

functions -c t1 t2

functions t1
functions t2
```

and run it as `fish func.fish`.

This prints:

```output
# Defined in func.fish @ line 1
function t1
    echo t1
end
function t2
    echo t1
end
```

Notice how `t2` does not print a line number? I feel it should, specifically line 5.

I noticed this while porting the "function.in" test to littlecheck, because that runs the tests as a file instead of via stdin.

esdmr:
I was writing a script to setup my OS after reinstall when I noticed this too. I was doing something like:

```fish
for i in (functions -n | string match -re '^setup_')
    functions --copy $i _original_$i

    eval "function $i
    _task_begin || return
    _original_$i || return
    _task_end || return
end"
end
```

Inside of `_original_$1`, I call `status stack-trace` if there were any errors. But I noticed that the trace did not contain any line numbers.

After some `git blame`s, I found the source to be [a single line](https://github.com/fish-shell/fish-shell/commit/208be0f4d4f42abe8e01eae0e41e22e8ac6bae0f#diff-0fc0d4326ef8eca43852e58c14fbc67da67204ce7940f87035410855bc898aeeR248) from 13 years ago, since the first time `--copy` (or rather `--rename`) was implemented back in `v2.0.0`.

[Currently](https://github.com/fish-shell/fish-shell/blob/538518cdf3d52046bb7423efe63690f44a13c205/src/function.cpp#L241), the line has a comment specifying that: “This new instance of the function shouldn't be tied to the definition file of the \[…] original, so clear the filename, etc.” Other than the comment, I did not find any rationales for it.

I think that it was because the original definition did not have the same name and that was deemed confusing.

I tried removing that line:

```diff
-new_props->definition_file = nullptr;
```

And it seems to work (?).

```fish
function hello
    status stack-trace
end

function world
    hello
end

echo ── Normal ──────────────────────────────────────────────────

world
functions hello
functions world

echo ── Copied ──────────────────────────────────────────────────

functions -c world world2

world2
functions world2
```

<details>
<summary>Normal fish output</summary>

```log
── Normal ──────────────────────────────────────────────────
in function 'hello'
	called on line 6 of file test.fish
in function 'world'
	called on line 11 of file test.fish
# Defined in test.fish @ line 1
function hello
    status stack-trace
end
[...]

Issue #3295, opened 2016-08-09
make it easy to learn where a function came from


krader1961:
We regularly get questions about unexpected behavior whose investigation involves verifying that the expected fish functions are being loaded. For example, issue #3210. It should be much easier to answer that question. A first approximation solution is to do something like

```
for fn_dir in $fish_function_path
    test -r $fn_dir/fish_vi_key_bindings.fish
    and echo $fn_dir/fish_vi_key_bindings.fish
end
```

The problem with that solution is there is no guarantee the function came from the corresponding autoload script.

I propose two enhancements. First, augment the `builtin functions` command to take a function name and report where it came from. Second, augment the `type` function to report all the corresponding autoload scripts using something like the snippet above. It seems like `type -P $func_name` should be enhanced to report the first autoload script found or all of them if `-a` is also specified. It's not clear why the current behavior only deals with external commands and not autoloaded functions. If changing `type -P` to also handle autoloaded functions is too confusing, or likely to break existing uses, then we could add an analogous `-F` option to find autoload scripts rather than external commands.

ridiculousfish:
We already track which functions are autoloaded. It would be straightforward to report that fact in the `functions` output.

floam:
This would enable a funced that can optionally work with originals, and easy checking for un-funcsaved functions (I'd like to have mine list them out at exit).

krader1961:
I've created PR#3718 that implements the enhancement described above to the `functions` command. It adds `-f` / `--file` flags. However I am not entirely happy with those flag names since the verbose output includes information not directly related to the file name where the function was defined. Specifically, whether or not the function was defined with the `--no-scope-shadowing` flag. Opinions for alternative flag names or otherwise requiring the non-source related output to be done under other flags is welcome.

faho:
>However I am not entirely happy with those flag names since the verbose output includes information not directly related to the file name where the function was defined.

I can see that. You've implemented this in a function called "show_function_details", so how about "--details" or "--show-details"? Of course it couldn't use "-d" since that's already taken for changing the description, so it would have to be "-D". Or simply call it "--verbose" and either support passing it twice (`functions -vv somefunc`) or print all the details always.

Or since this _is_ metadata, something with that term?

krader1961:
[...]

PR #9542, opened 2023-02-05
`functions --copy`: store file name and line number

esdmr:
## Description

Keeps the location of original function definition, and also stores where it was copied. `functions` and `type` show both locations, instead of none. It also retains the line numbers in the stack trace.

Fixes issue #6575

## TODOs:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,6 +18,8 @@ Scripting improvements
 ----------------------
 - ``abbr --list`` no longer escapes the abbr name, which is necessary to be able to pass it to ``abbr --erase`` (:issue:`9470`).
 - ``read`` will now print an error if told to set a read-only variable instead of silently doing nothing (:issue:`9346`).
+- ``functions`` and ``type`` now show where a function was copied and where it originally was instead of saying ``Defined interactively``.
+- Stack trace now shows line numbers for copied functions.
 
 Interactive improvements
 ------------------------
--- a/doc_src/cmds/functions.rst
+++ b/doc_src/cmds/functions.rst
@@ -34,10 +34,10 @@ The following options are available:
     Causes the specified functions to be erased. This also means that it is prevented from autoloading in the current session. Use :doc:`funcsave <funcsave>` to remove the saved copy.
 
 **-D** or **--details**
-    Reports the path name where the specified function is defined or could be autoloaded, ``stdin`` if the function was defined interactively or on the command line or by reading standard input, **-** if the function was created via :doc:`source <source>`, and ``n/a`` if the function isn't available. (Functions created via :doc:`alias <alias>` will return **-**, because ``alias`` uses ``source`` internally.) If the **--verbose** option is also specified then five lines are written:
+    Reports the path name where the specified function is defined or could be autoloaded, ``stdin`` if the function was defined interactively or on the command line or by reading standard input, **-** if the function was created via :doc:`source <source>`, and ``n/a`` if the function isn't available. (Functions created via :doc:`alias <alias>` will return **-**, because ``alias`` uses ``source`` internally. Copied functions will return where the function was copied.) If the **--verbose** option is also specified then five lines are written:
 
-    - the pathname as already described,
-    - ``autoloaded``, ``not-autoloaded`` or ``n/a``,
+    - the path name as already described,
+    - if the function was copied, the path name to where the function was originally defined, otherwise ``autoloaded``, ``not-autoloaded`` or ``n/a``,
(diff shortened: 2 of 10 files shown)


esdmr:
A couple of points left:

- Should `functions -D` and `type -p` print the path to original definition or its copy?
- Should `functions -Dv` and `type -s` be modified to include the both paths?
- Chaining copies only keeps the location of the last copy. Should we keep them all?
- Do we need to preallocate space for the newline (in case the function was not a copy), or is the performance penalty negligible?

mqudsi (commented):
Thanks for tackling this - a brave undertaking for a first foray into the fish code!

> Should `functions -D` and `type -p` print the path to original definition or its copy?

When two answers are valid, I look at edge cases and see what still makes sense there (maybe nothing will!). 

Imagine `src` is in `src.fish` and `copy` is made in `config.fish`. Printing `~/.config/fish/functions/src.fish` in response to `type -p copy` makes no sense since there is no mention of `copy` in that file. Someone unaware that it's a copy that only executed `type -p` should be given a result that can set them on the right track, so I would say the file that executes `functions -c src copy` should be the output of `type -p`.

I think `functions -D` should behave the same, and the argument for that is even stronger there. Currently `functions -D` would print `stdin` if you source `copy_funcs.fish` and that contains the `functions -c src dst` call. I would say returning the path to `copy_funcs.fish`  would be a step up from that while returning the equivalent of `functions -D src` would be a breaking change from that.

> Should functions -Dv and type -s be modified to include both paths?

Unlike `type -p` (and maybe `functions -D` though not as much), `functions -Dv` and `type -s` aren't intended to be used programmatically and their output isn't considered sacrosanct. The output of `-Dv` is currently machine parseable but it's not well-defined at all (I'm not sure how much use it sees).

For `type -s` I think changing it from the current output for a copied function:

> xxx is a function

to something more useful and more similar to the output for a file-defined function would be a plus:

> xxx is a function (defined in <path to file that executed `functions -c`>:<lineno> as a copy of <original func name>) 

is good.

I don't know what to do about `functions -Dv` because its output is fairly meaningless right now (we dump a fixe format, parseable output but we don't define the output anywhere). You could easily change it from the current

```
stdin
n/a 
0
scope-shadowing
```

to 

```
<path to file that executed `functions -c`>
<path to original file>
0
scope-shadowing
```

(anything would be an improvement over "n/a" displayed instead of "autoloaded" or "not-autoloaded")

but whether or not this is a meaningful change, a breaking change, etc. I honestly can't answer.
 
> Chaining copies only keeps the location of the last copy. Should we keep them all?
I don't think there's a need to do that. 

> This is a bit pedantic, but… do we need to preallocate space for the newline (in case the function was not a copy), or is the performance penalty negligible?

It really doesn't matter. This shouldn't ever be in a hot path, if I understand what you're asking correctly.

esdmr:
@mqudsi Thank you for your comment. I have made these changes:

- `functions -D` and `type -p` now print where the function was copied. (Prints definition if not copied.)
- `functions -Dv` second line prints where the original function was instead of “n/a” or “not-autoloaded”. (Prints autoload state if not copied.)
- `type -s` now shows the same description as `functions` and `type` (“Defined in …, copied in …”).
- Previously, `type -s` was localized. Now, both `type` and `type -s` are localized (because I reused the string).
- However, `functions` is not localized (yet). Should it be?
- More tests.

ridiculousfish:
This is really cool - thank you for implementing this!

> functions -Dv, functions -D, and type -p are machine readable but their output has changed. This is most likely a breaking change

We'll do it anyways, I think it's unlikely to break anyone though we may revert this part if we discover it does. [Here's where the machine readable stuff](https://github.com/fish-shell/fish-shell/issues/3295) came from, it was more of a "someone might want this" than a user request.

> However, functions is not localized (yet). Should it be?

Error messages should certainly be localized. Regarding the emitted comments like "# Defined interactively," maybe they should, but that doesn't have to be part of this commit.

Two last things to do:

1. Please squash (or I can do it)
2. Update the docs `functions.rst` for the `-D/--details` flag to describe the new output.

@mqudsi any further concerns? If not I'll merge once the above is done. Thank you again!

esdmr:
@ridiculousfish Thank you! I updated the docs and squashed the commits.

mqudsi (approved):
No, I'm happy with this. Good idea, comprehensive implementation, docs, test coverage :+1: 

We might want to at some point describe the output of `functions -Dv` but that's separate from this.

zanchey:
I'm not taking this for 3.6.1 because of the behaviour change.

PR #9544, opened 2023-02-05
Port echo builtin to Rust

Xiretza:
## Description

An easy first module, along with some infrastructure changes to make sure `$status` is updated correctly for Rust builtins.

## TODOs:


mqudsi:
I approved the CI workflow.

src/builtin.cpp
```diff
@@ -82,8 +81,8 @@
 #include "wutil.h"  // IWYU pragma: keep
 
 static maybe_t<RustBuiltin> try_get_rust_builtin(const wcstring &cmd);
-static proc_status_t builtin_run_rust(parser_t &parser, io_streams_t &streams,
```
mqudsi, line 85: Is there a reason you aren't using `proc_status_t`? Rust likes strongly types more than C++ does, we want to encourage new types everywhere possible instead of reverting to returning integers.

fish-rust/src/builtins/shared.rs
```diff
@@ -89,17 +123,29 @@ fn rust_run_builtin(
...
+    match run_builtin(parser.unpin(), streams, args.as_mut_slice(), builtin) {
+        None => false,
+        Some(Ok(())) => {
+            *status_code = 0;
+            true
+        }
+        Some(Err(e)) => {
+            *status_code = e.into();
+            false
+        }
```
krobelus, line 136: nit: bad patch split here

fish-rust/src/builtins/shared.rs
```diff
@@ -23,17 +23,50 @@ mod builtins_ffi {
...
+            status_code: &mut i32,
+        ) -> bool;
     }
 
     impl Vec<wcharz_t> {}
 }
 
-/// A handy return value for successful builtins.
-pub const STATUS_CMD_OK: Option<c_int> = Some(0);
+pub enum CommandError {
```
krobelus, line 33: I agree that that an enum is preferable to `c_int` but I'm not yet convinced of using `Result<(), CommandError>` as opposed to including 0 in the enum.
It feels weird to not carry over this comment and the defininition of 0:

```c++
// Return values (`$status` values for fish scripts) for various situations.
enum {
    /// The status code used for normal exit in a command.
    STATUS_CMD_OK = 0,
```

From the implementation's point of view, 0 is treated the same as other status codes.
If we translated the  enum as-is, it would look like

```cpp
enum Status {
	CmdOk = 0,
	CmdError = 1,
	InvalidArgs = 2,
	[...]
}
```

I think the benefit you're going for is the convenient early-exit in

	let (opts, optind) = parse_options(args, parser, streams)?;

where `parse_options()` returns `Result<(Options, usize), CommandError>` 
However the same thing would be possible if you use `Status` instead of `CommandError`.

fish-rust/src/builtins/echo.rs
```diff
@@ -0,0 +1,228 @@
+// Implementation of the echo builtin.
```
krobelus, line 1: should use `//!` instead of `//`

fish-rust/src/builtins/echo.rs
```diff
@@ -0,0 +1,228 @@
...
+        while let Some(c) = chars.next() {
+            if !opts.interpret_special_chars || c != '\\' {
+                // Not an escape.
+                out.push(c);
+                continue;
+            }
+
+            // Most escapes consume one character in addition to the backslash; the numeric
+            // sequences may consume more, while an unrecognized escape sequence consumes none.
+            let Some(next_char) = chars.peek() else {
```
krobelus, line 170: I love `let-else`. We might need to desugar it when we lower our Rust version requirements but let's not worry about that now.

fish-rust/src/builtins/echo.rs
```diff
@@ -0,0 +1,228 @@
...
+                out.push('\\');
+                break;
+            };
+
+            // Most escapes consume one character in addition to the backslash; the numeric
+            // sequences may consume more, while an unrecognized escape sequence consumes none.
```
krobelus, line 176: This comment is duplicated. The second occurrence makes more sense.
Instead of the first I'd write something like `// Incomplete escape sequence is echoed verbatim`.

fish-rust/src/builtins/mod.rs
```diff
@@ -1,2 +1,4 @@
 pub mod shared;
+
 pub mod wait;
+pub mod echo;
```
krobelus, line 4: should be sorted, I think rustfmt does that automatically.

Same goes for `try_get_rust_builtin` and the `RustBuiltin` enum definition.
They should be sorted alphabetically, like the existing C++ builtins are

fish-rust/src/builtins/shared.rs
```diff
@@ -23,17 +23,50 @@ mod builtins_ffi {
             streams: Pin<&mut io_streams_t>,
             cpp_args: &Vec<wcharz_t>,
             builtin: RustBuiltin,
-        );
+            status_code: &mut i32,
+        ) -> bool;
```
krobelus, line 27: CXX doesn't support `std::optional` but we can return a makeshift `SharedPtr<i32>`, which is probably nicer than an output parameter

fish-rust/src/builtins/shared.rs
```diff
@@ -79,7 +112,8 @@
     streams: Pin<&mut builtins_ffi::io_streams_t>,
     cpp_args: &Vec<wcharz_t>,
     builtin: RustBuiltin,
-) {
+    status_code: &mut i32,
```
krobelus, line 115: nit: CXX doesn't support type aliases in the extern declarations, but in the definition we can use `libc::c_int` (but I'm not sure if this is the perfect type, so feel free to ignore)

fish-rust/src/builtins/shared.rs
```diff
@@ -89,17 +123,29 @@
...
+
+    match run_builtin(parser.unpin(), streams, args.as_mut_slice(), builtin) {
+        None => false,
+        Some(Ok(())) => {
+            *status_code = 0;
+            true
+        }
+        Some(Err(e)) => {
+            *status_code = e.into();
+            false
```
krobelus, line 135: a `false` return will be interpreted as "don't update status" which is wrong. That would mean `builtin false` would never change `$status`.

Instead of mixing `Result<(), CommandError>` and `bool` we should probably use `Option<Status>` everwhere (plus the equivalent `SharedPtr<Status>` or `SharedPtr<i32>` at the language boundary).

src/builtin.cpp
krobelus: it's because this commit makes `builtin_run_rust` do the same thing as the actual builtin functions which return a `maybe_t<int>`. We might want to change that when we port `builtin_run` to rust but for now there's no reason

fish-rust/src/builtins/shared.rs
mqudsi: I feel like I'm misunderstanding or missing something you're saying, Johannes. If there is only one OK then there is no need to use a `Result` type. However, if you forgo the result type you can no longer use `?` to early return and bubble errors back to the caller. 

I'm not sure how you could early return if you used status? You'd have to implement `Into<SomeOtherError>` so that `?` could desugar into a return, and you'd have to do that for the return error type of every caller handling a status result.

krobelus: There is no code path that treats `exit 0` different from `exit 1`, so it's wrong to separate the two.
We can still use `Result` the same way this patch proposes to.

faho: In general, I would avoid doing these sorts of plumbing changes in PRs for specific builtins - that means everyone else now needs to rebase on this.

Or is there a real *need* to make this switch? AFAICT returning an Option<c_int> should work for now? It's not idiomatic rust, it's very closely modelled on the C++ version, but then that's what we said we'd be doing.

krobelus: yeah, actually an enum might be a bit overengineered. Enums are mostly useful when used in a switch, so the compiler can tell us that we forgot a case. I don't think we ever switch on an exit code (in a way that wants to handle all cases).

So unless I have missed a benefit of enums here, I think constants model the notion of exit codes better.
We can still make it type-safe by defining a tuple struct, something like

```rust
#[derive(Clone, Copy, PartialEq, Eq)]
pub Status(pub u8);
pub const STATUS_CMD_OK: Status = Status(0);
pub const STATUS_CMD_ERR: Status = Status(1);
```

Xiretza: `Result`'s design *really* does not like success values (`CMD_OK`) to be included in the error type. The `E` type in `Result<T, E>` must represent only error values due to how the `?` short-circuiting logic works - otherwise, a function might return `Err(CMD_OK)`, and if the caller uses `?` on the result, execution is incorrectly aborted.

faho: The return type here, as it is used in the C++, *is* either:

1. A status code. This is a unix api thing, it's a whole number between 0 and 255 inclusive, where 0 is "true" and everything else is "false". We have a convention on when we use which falsey number, but it's not necessary e.g. that "1" is always only used as this - `set -q` returns how many variables weren't set.
2. A marker that the status shouldn't be changed - this is important for `set`, where `set foo bar` doesn't touch $status but `set -q foo` does.

So the C++ returned a maybe_t<int>, with a "None()" for the latter case. Note that none of these are "errors" as such on the fish-internal side. It's just a return value of the builtin.

But anyway: Changing this isn't necessary to implement `echo`, and I would prefer we leave this untouched for now. Especially because multiple people are gonna be implementing multiple builtins at the same time and so switching this plumbing around forces them to adjust their work again.

Xiretza: Right, the equivalent of the C++ return type would be `Option<Status>`.

- If `Status` is simply an integer (or newtype of one), `?` can not be used - `Try` can not be implemented yet, so if `?` should be usable, `Status` needs to be a `Result`.
- Even if we stay with plain integers for now, I would advocate for putting the `Option<Status>` in a newtype - otherwise it's too easy to accidentally use `?` on such a return type, which is a nonsensical operation (short-circuit execution if the called function would not change the status).

faho: >Even if we stay with plain integers for now

These *are* integers. We can have some form of aliasing e.g. 2 to "invalid arguments", but fundamentally the return value here will be stored in $status as an integer, and in some cases it just is a number with no deeper encoded meaning.

> I would advocate for putting the Option<Status> in a newtype

That might, in the end, be the right solution, or it might not. But it doesn't belong in this PR, and it's not a good idea to do it *now*, because you have multiple people working on multiple builtins and so you don't want to change things from under them. Remember that currently you might have one person working on "random" and another on "math" and another on "count".

> otherwise it's too easy to accidentally use ? on such a return type

For context here: Builtins will *always* be called from builtin_run. That's the one place that dispatches builtins. So it's not a big issue if you can't use `?` because it'll only ever be done in that one place, and that one place won't change a lot.

krobelus: > If Status is simply an integer (or newtype of one), ? can not be used - Try can not be implemented yet, so if ? should be usable, Status needs to be a Result.

This is not true. We'd use `Result<(), Status>` to the same effect as now. The only difference is as you said

> a function might return Err(CMD_OK), and if the caller uses ? on the result,

which should never happen but I can see how it might creep in during the transition given that today builtins' helper functions use CMD_OK as what would be `Ok(())`. I guess we can add a runtime check for that in builtin_run()..

krobelus: > But it doesn't belong in this PR, and it's not a good idea to do it now, because you have multiple people working on multiple builtins and so you don't want to change things from under them. 

I think it's no big deal to make such a change now, it's easy to update pending builtins to use the new API.
I agree that it should go into a separate PR.

mqudsi: > This is not true. We'd use Result<(), Status> to the same effect as now. The only difference is as you said

I understood you to mean `-> Status` and not `-> Result<(), Status>`. This would work fine, though `Status` should idiomatically have an error-y name and not include success.

fish-rust/src/builtins/shared.rs
Xiretza: What would that look like on the Rust side? How do I return a `SharedPtr`?

fish-rust/src/builtins/shared.rs
Xiretza: Yep, forgot to push that fixup yesterday night.

> Instead of mixing `Result<(), CommandError>` and bool we should probably use `Option<Status>` everwhere

The builtins' internals often use short-circuiting on `STATUS_CMD_OK`, as discussed above, if you naively do that on `Option<Status>`, you get nonsensical semantics.

Xiretza:
I've addressed the mistakes, changed the error handling to match the old way, and rebased on riir.

fish-rust/src/builtins/shared.rs
ridiculousfish: [SharedPtr docs here](https://cxx.rs/binding/sharedptr.html), but I think this is fine as written.

fish-rust/src/builtins/shared.rs
ridiculousfish: Agree with faho that this should stay as an `Optional<c_int>` for now - fidelity to the existing code is more important than Rust idiomaticity. Also, builtins need to be able to express arbitrary return codes, e.g. `eval`, etc, so an enum won't work.

Xiretza: > Also, builtins need to be able to express arbitrary return codes, e.g. `eval`, etc, so an enum won't work.

The enum already had a `Custom(n)` variant, which was already used by `wait`, so that's not a concern.

krobelus:
looks great, thanks! I'll merge later tonight.
I have two remaining nitpicks: ordering is still off in `shared.rs` and I would return `c_int` instead of `Option<c_int>` in parse_options, to simplify it a bit

Xiretza:
> ordering is still off in `shared.rs`

Sorry about that, must've gotten lost during rebasing.

> and I would return `c_int` instead of `Option<c_int>` in parse_options, to simplify it a bit

`STATUS_INVALID_ARGS` is defined as `Some(2)`, should I change the definition to `2`, `.unwrap()` it in `parse_options()`, or just put a `2` there?

krobelus:
> STATUS_INVALID_ARGS is defined as Some(2), should I change the definition to 2, .unwrap() it in parse_options(), or just put a 2 there?

hmm I missed that. Not sure what's the best solution here. I thought we only add the Option to top-level builtin functions (like in the C++ code). If we use it everywhere a newtype alias might make sense.. but maybe we shouldn't

Issue #9461, opened 2023-01-10
The future of webconfig


faho:
I've thought for quite a while that we need to do something about webconfig - our web-based configuration tool.

It isn't keeping up with new features, like our extended abbreviations - #9460.
It also makes implementing some features more annoying, like #9456 - which would be trivial in fish, but is awkward if you need to add it in two places and marshall the data through.

Another is that angularjs has gone EOL with no easy upgrade path in January 2022 - that's a year ago! (you may now feel old)

In short: It needs some real work put into it, and that isn't happening. Since 3.4.0 (released March 2022), it has received 16 commits (excluding those to themes and sample_prompts), most of them extremely small (5 just to the css). No substantial work on it has happened.

So, I propose one of two possibilities:

- We come up with a plan to port it off angular, and ideally get someone to maintain it and bring it up to speed
- We remove it - to be clear: the function `fish_config` and its CLI with the `theme` and `prompt` subcommands would stay, `fish_config browse`, the web-UI would go.

I realize that the latter is hard - webconfig is a flashy feature, and it has served fish well especially in the time that the CLI-version wasn't available. But now that we have `fish_config prompt` and `fish_config theme`, I believe that it doesn't really add a lot of value and it is time to let go.

ridiculousfish:
I agree with your diagnosis and yet I hope to save web-UI `fish_config`. In my experience it's a powerful way to make users instantly comfortable with fish. Yes we do need to rewrite it from AngularJS, I have the stomach to do so, it's not that much code.

The observation about #9456 is good - `fish_config` does its magic through universal variables and we are trying to move away from those. It's a real problem if web-UI `fish_config` is an obstacle to new features.

Sorry I don't have any concrete conclusions at the moment, I need to think about it a bit.

septatrix:
> Yes we do need to rewrite it from AngularJS, I have the stomach to do so, it's not that much code.

Do you already have a framework in mind? As it is a rather small app I would suggest Svelte or even lit-html as opposed to some of the heavier options like React, Vue, or Angular. I could also help with this if you want

faho:
@septatrix I don't think we really care about which, specifically, just that it's light on dependencies - i.e. no need to run "npm" or ideally any tool whatsoever.

Really what we need here is basically a template engine - "add all themes in this spot".

septatrix:
[...]

PR #9554, opened 2023-02-07
WIP: Proof of concept for replacing AngularJS with Alpine.js in webconfig component

septatrix:
## Description

Alpine.js is a very lightweight JS framework bringing many of the comforts from heavier frameworks like Vue or Angular. A huge benefit of Alpine.js is that it does not require any build steps unlike some other more lighter frameworks such as Svelte.

I have wanted to try out Alpine.js for a while now and worked on this during a train ride. It was very pleasant to work with and easy to understand, so I think will also help to reduce the maintenance burden coming forward.

This is for now just a POC, most importantly the color tab is not yet implemented. I also plan to split the HTML and JS into separate files again.

Fixes issue #9461 (at least the part about replacing EOL AngularJS)

## TODOs:


ridiculousfish:
Thank you! This is very cool. I think alpine.js is an excellent choice because of its minimalism. I'll review.

share/tools/web_config/js/alpine.js
```diff
@@ -0,0 +1,5 @@
...rEach(([o,{value:s,enumerable:a}])=>{if(a===!1||s===void 0)return;let c=i===""?o:`${i}.${o}`;typeof s=="object"&&s!==null&&s._x_interceptor?n[o]=s.initialize(e,c,o):t(s)&&s!==n&&!(s instanceof Element)&&r(s,c)})};return r(e)}function Ee(e,t=()=>{}){let r={initialValue:void 0,_x_interceptor:!0,initialize(n,i,o){return e(this.initialValue,()=>yn(n,i),s=>lt(n,i,s),i,o)}};return t(r),n=>{if(typeof n=="object"&&n!==null&&n._x_interceptor){let i=r.initialize.bind(r);r.initialize=(o,s,a)=>{let c=n.initialize(o,s,a);return r.initialValue=c,i(o,s,a)}}else r.initialValue=n;return r}}function yn(e,t){return t.split(".").reduce((r,n)=>r[n],e)}function lt(e,t,r){if(typeof t=="string"&&(t=t.split(".")),t.length===1)e[t[0]]=r;else{if(t.length===0)throw error;return e[t[0]]||(e[t[0]]={}),lt(e[t[0]],t.slice(1),r)}}var Zt={};function y(e,t){Zt[e]=t}function se(e,t){return Object.entries(Zt).forEach(([r,n])=>{Object.defineProperty(e,`$${r}`,{get(){let[i,o]=ut(t);return i={interceptor:Ee,...i},ye(t,o),n(t,i)},enumerable:!1})}),e}function Qt(e,t,r,...n){try{return r(...n)}catch(i){Z(i,e,t)}}function Z(e,t,r=void 0){Object.assign(e,{el:t,expression:r}),console.warn(`Alpine Expression Error: ${e.message}
```
faho, line 1: I'm reasonably sure distributions would prefer we ship the unminified source, and size isn't really a big deal here.

septatrix: I was using a minified version because the angular files were also minified. The full version is about double the size¹ but I agree that unminified is probably preferred by Distros.

¹ <https://cdn.jsdelivr.net/npm/alpinejs@3.11.1/dist/>

septatrix:
I also hooked up the color tab now. What remains to be done now is cleaning up the code by splitting it up in files again and test some edge cases.

One thing to note: Alpine.js does not support IE11 due to usage of `@vue/reactivity` under the hood. As IE is officially unsupported (finally) and only those using WSL and IE11 would be affected I do not think this should be considered a problem. I suspect most fish users are more tech-adapt and use evergreen browsers anyhow.

Due to the above, we can also reasonably replace a lot of old code with more modern equivalents. Another thing I would like to work on is make the webconfig a bit more accessible: E.g. use proper elements like inputs and buttons instead of divs with onclick handlers everywhere, improve keyboard navigation etc.

mqudsi (commented):
This is a nice change. Dropping IE11 support is probably wholly inconsequential for our purposes.

faho:
Alright, I've checked this out, made the prompt tab work.

I can't find a bunch missing, so I'm going to merge this *now* so we can then collaborate on it together, instead of letting this PR languish in draft mode.

Thanks for all that work!

PR #9560, opened 2023-02-09
completions/trash-cli: add completions for trash-cli

Jay-716:
## Description
Add completions for trash-cli commands:
trash, trash-empty, trash-list, trash-put and trash-restore.

## TODOs:


share/completions/trash-empty.fish
```diff
@@ -0,0 +1,9 @@
+complete -f -c trash-empty -s h -l help -d 'show help message'
+complete -f -c trash-empty -l print-completion -a 'bash zsh tcsh' -d 'print shell completion script'
```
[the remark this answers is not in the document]
Jay-716: Thanks for your swift review.
I'm sorry for making such a mistake and I have to add one more commit to fix that.
I realize that this pr may be abnormal and I can reopen one if you want to.
Thanks. :)

mqudsi (changes requested):
I would request identification of the `trash` executable via some means as it isn't a unique name by any means and there are several implementations in different languages and for different platforms floating around.

CHANGELOG.rst
```diff
@@ -41,7 +41,12 @@ Completions
   - ``otool``
   - ``mix phx``
   - ``neovim``
-  
+  - ``trash``
+  - ``trash-empty``
+  - ``trash-list``
+  - ``trash-put``
+  - ``trash-restore``
```
mqudsi, line 48: Can you collapse these to one line ("`trash` and helper utilities `trash-empty`, `trash-put`, ...")?

Jay-716:
It's my fault to ignore other implementations of trash tools.
There are several things that need to be confirmed:
 - There are many trash tools, and which file should their completions locate? Can I merge all of the completions into a single file ```trash.fish``` or separate them into different .fish files and add implementation-specific completions when there are binaries with the same name?
 - May I collapse those lines in CHANGELOG.rst into ``` ``trash`` and utilities of various implementations``` ?
 - May I squash all the commits into a single one before merging(if possible)?

And there are relatively popular trash cli tools I found:
https://github.com/andreafrancia/trash-cli
https://github.com/sindresorhus/trash-cli
https://github.com/ali-rantakari/trash
https://github.com/PhrozenByte/rmtrash

And as for the means to identify ```trash``` excutables, I think I can use some kind of ```trash --version | string match "xxx"```
Would that be accetable? or any suggestions?
Sorry for my poor English. :)

mqudsi:
No worries and your English is excellent!

To be clear, I'm not asking you to add completions for other versions/implementations of `trash`, but just to make sure that your completions only kick in for the one you intend them to. Matching against `trash --version` (usually not an approach we like because it can break in the future) is fine here.

Jay-716:
Thanks for your patient guidance.
I decided to match the help message instead of the version because it may not be broken so easily.
And I just add the completions for only one implementation because I am not familiar with the others and I don't have an appropriate environment to test them, therefore I cannot ensure their quality. Sorry about that.

share/completions/trash.fish
```diff
@@ -0,0 +1,19 @@
...
+    complete -f -c trash -s i -l interactive -d 'prompt before every removal'
+    complete -f -c trash -s r -s R -l recursive -d 'ignored (for GNU rm compatibility)'
+    complete -F -c trash -l trash-dir -d 'specify trash folder'
+    complete -f -c trash -s v -l verbose -d 'be verbose'
+    complete -f -c trash -l version -d 'show version number'
+end
+
+if string match -qr "https://github.com/andreafrancia/trash-cli" (trash --help)
+    __trash_by_andreafrancia
+end
```
mqudsi, line 19: This is good - you're right, this is less likely to break than `--version`.

There's just one small problem: let's say tomorrow someone installs a different `trash` that doesn't have `--help` (so it'll print an error to stderr) or it writes `--help` output to stderr for some reason. In this case, `trash.fish` will print junk to the terminal because you are only capturing stdout in the subshell `(trash --help)`.

Can you change this to either `(trash --help 2>/dev/null)` or `(trash --help >&1)` to prevent that from happening?

I can change it too but I wanted to share this with you as a learning opportunity.

mqudsi (changes requested):
Looks good except for the small issue with `--help` output that I commented on above.

Jay-716:
I should have taken the missing help option and stderr into consideration.
Thanks for your telling me about that and giving me this valuable learning opportunity.
It's so considerate and friendly of you haha. :)

PR #9561, opened 2023-02-10
Port emit builtin to rust

Xiretza:
I am super unsure about the C++/FFI changes, so please review those carefully.


Xiretza:
47cc98fd5731917f4874420bacf92cd868ca0e78 broke the conversion from `const wcharz_t *` to `std::vector<std::wstring>` somehow, I don't know enough about C++ to figure out how to fix it:

```
In file included from /usr/include/c++/12.2.1/memory:65,
                 from /home/xiretza/dev/fish-shell/src/event.h:12,
                 from /home/xiretza/dev/fish-shell/src/event.cpp:4:
/usr/include/c++/12.2.1/bits/stl_uninitialized.h: In instantiation of ‘constexpr bool std::__check_constructible() [with _ValueType = __cxx11::basic_string<wchar_t>; _Tp = const wchar_t&]’:
/usr/include/c++/12.2.1/bits/stl_uninitialized.h:182:4:   required from ‘_ForwardIterator std::uninitialized_copy(_InputIterator, _InputIterator, _ForwardIterator) [with _InputIterator = const wchar_t*; _ForwardIterator = __cxx11::basic_string<wchar_t>*]’
/usr/include/c++/12.2.1/bits/stl_uninitialized.h:372:37:   required from ‘_ForwardIterator std::__uninitialized_copy_a(_InputIterator, _InputIterator, _ForwardIterator, allocator<_Tp>&) [with _InputIterator = const wchar_t*; _ForwardIterator = __cxx11::basic_string<wchar_t>*; _Tp = __cxx11::basic_string<wchar_t>]’
/usr/include/c++/12.2.1/bits/stl_vector.h:1690:33:   required from ‘void std::vector<_Tp, _Alloc>::_M_range_initialize(_ForwardIterator, _ForwardIterator, std::forward_iterator_tag) [with _ForwardIterator = const wchar_t*; _Tp = std::__cxx11::basic_string<wchar_t>; _Alloc = std::allocator<std::__cxx11::basic_string<wchar_t> >]’
/usr/include/c++/12.2.1/bits/stl_vector.h:706:23:   required from ‘std::vector<_Tp, _Alloc>::vector(_InputIterator, _InputIterator, const allocator_type&) [with _InputIterator = const wchar_t*; <template-parameter-2-2> = void; _Tp = std::__cxx11::basic_string<wchar_t>; _Alloc = std::allocator<std::__cxx11::basic_string<wchar_t> >; allocator_type = std::allocator<std::__cxx11::basic_string<wchar_t> >]’
/home/xiretza/dev/fish-shell/src/event.cpp:493:81:   required from here
/usr/include/c++/12.2.1/bits/stl_uninitialized.h:90:56: error: static assertion failed: result type must be constructible from input type
   90 |       static_assert(is_constructible<_ValueType, _Tp>::value,
      |                                                        ^~~~~
/usr/include/c++/12.2.1/bits/stl_uninitialized.h:90:56: note: ‘std::integral_constant<bool, false>::value’ evaluates to false
```

src/event.cpp
```diff
@@ -488,6 +489,11 @@ void event_print(io_streams_t &streams, const wcstring &type_filter) {
     }
 }
 
+void event_fire_generic(parser_t &parser, wcstring name, const wcharz_t argv[], int argc) {
```
faho, line 492: Note: *Never* use array function arguments in C++. They don't really *exist*.

C++ arrays fall apart and decay into pointers at the first convenience, and they don't actually carry around anything useful (like, you know, length?).

This should be `const wcharz_t *argv`.

Anyway, the construction guff can be sidestepped by just going the long way around:

```c++
void event_fire_generic(parser_t &parser, wcstring name, const wcharz_t *argv, int argc) {
    wcstring_list_t args_vec{};
    for (int i = 0; i < argc; i++) {
        args_vec.push_back(argv[i]);
    }
    event_fire_generic(parser, std::move(name), std::move(args_vec));
}
```

I *think* the kerfuffle could be triggered by const-issues, but tbh I'm so tired of C++ I don't want to learn user conversion operators *now*.

Xiretza: > Note: Never use array function arguments in C++. They don't really exist.
>
> C++ arrays fall apart and decay into pointers at the first convenience, and they don't actually carry around anything useful (like, you know, length?).

oh yeah, I know that from C, I just saw it as an annotation for humans that `argv` isn't a single `wcharz_t`, but rather a pointer to several of them. Not really necessary given the argument name though.

> Anyway, the construction guff can be sidestepped by just going the long way around

Applied, thanks!

krobelus:
> 47cc98f (cc @krobelus) broke the conversion from const wcharz_t * to std::vector<std::wstring> somehow, I don't know enough about C++ to figure out how to fix it:

We can revert 47cc98f and use explicit conversions.
I'm not yet sure how to provoke the problem you saw

fish-rust/src/builtins/emit.rs
```diff
@@ -0,0 +1,51 @@
...
+        Err(err) => panic!("Illogical exit code from parse_options(): {err:?}"),
+    };
+
+    if opts.print_help {
+        builtin_print_help(parser, streams, cmd);
+        return STATUS_CMD_OK;
+    }
+
+    let Some(event_name) = argv.get(opts.optind) else {
+        streams.err.append(WString::from_str(&format!("{cmd}: expected event name\n")));
```
krobelus, line 28: `format!` requires explicit conversion to widestring so we should avoid it.
Use `sprintf!` instead.

```diff
diff --git a/fish-rust/src/builtins/emit.rs b/fish-rust/src/builtins/emit.rs
index 57d99ee93..bb4a5c99f 100644
--- a/fish-rust/src/builtins/emit.rs
+++ b/fish-rust/src/builtins/emit.rs
@@ -6,4 +6,7 @@ use super::shared::{
 use crate::ffi::{self, parser_t, Repin};
-use crate::wchar_ffi::{wstr, W0String, WCharToFFI, WString};
+use crate::wchar_ffi::{wstr, W0String, WCharToFFI};
+use crate::wutil::sprintf;
+use widestring_suffix::widestrs;
 
+#[widestrs]
 pub fn emit(
@@ -27,3 +30,3 @@ pub fn emit(
     let Some(event_name) = argv.get(opts.optind) else {
-        streams.err.append(WString::from_str(&format!("{cmd}: expected event name\n")));
+        streams.err.append(&sprintf!("%s: expected event name\n"L, cmd));
         return STATUS_INVALID_ARGS;
```

There's one open question: `sprintf!` works fine with both `%s` and `%ls` format strings, which one should we use?

fish-rust/src/builtins/shared.rs
```diff
@@ -155,3 +157,54 @@ pub fn builtin_print_help(parser: &mut parser_t, streams: &io_streams_t, cmd: &w
         empty_wstring(),
     );
 }
+
+pub struct HelpOnlyCmdOpts {
+    pub print_help: bool,
```
krobelus, line 162: Given that we add `optind` here which feels a bit out of place, maybe it's better to replace the struct by a pair.

This ties in with another consideration:
You replace `parse_help_only_cmd_opts()` with `HelpOnlyCmdOpts::parse()`.
Whoever ports the next builtin that uses `parse_help_only_cmd_opts()` needs to figure out that it's called `HelpOnlyCmdOpts::parse()` now.

We already change type names (from `foo_bar_t` to `FooBar`) but that's less of an issue because those follow an obvious pattern.
So far we mostly preserve function names.

Actually this is likely not an issue here because it will be obvious how to port the earlier `struct help_only_cmd_opts_t`.
But it's something to keep in mind in general.

I'm guilty of the same thing, I replaced
`keyword_description()` with an implementation of `From<ParseKeyword> for &'static wstr`
`keyword_description()`  is still there but it's private and only used for FFI.
I've been thinking maybe I should make `keyword_description()` public again, and maybe add a TODO that it can be removed once we're done.

Summing up, I think your version is perfectly fine.
I think it's worth considering this alternative:

	let Ok((print_help, optind)) = parse_help_only_cmd_opts();

fish-rust/src/builtins/emit.rs
```diff
@@ -0,0 +1,51 @@
...
+        .collect();
+
+    ffi::event_fire_generic(
+        parser.pin(),
+        event_name.to_ffi(),
+        event_arg_ptrs.as_ptr(),
+        i32::try_from(event_arg_ptrs.len()).unwrap().into(),
+    );
+
+    None
```
krobelus, line 50: No, it should return `STATUS_CMD_OK`.
Only a select few builtins may return no status.

fish-rust/src/builtins/emit.rs
```diff
@@ -0,0 +1,51 @@
...
+    };
+
+    let event_args: Vec<W0String> = argv[opts.optind + 1..]
+        .iter()
+        .map(|s| W0String::from_ustr(s).unwrap())
+        .collect();
+    let event_arg_ptrs: Vec<ffi::wcharz_t> = event_args
+        .iter()
+        .map(|s| ffi::wcharz_t {
+            str_: s.as_ucstr().as_ptr().cast(),
```
krobelus, line 39: Correct. The W0String objects live until the function ends, so it's okay to pass their pointers.

I think we can simplify this, it seems `as_ucstr()` and `cast()` are unnecessary.
Also we can implement `From<&W0String> for wcharz_t`, in case we want to use this more often.
(could also implement `From<&Vec<W0String>> for Vec<wcharz_t>`).
Of course the conversion functions can be misused - it doesn't warn you if the `W0String` goes out of scope, which can  leave the `wcharz_t` as dangling reference.  I think that's acceptable.

```
diff --git a/fish-rust/src/builtins/emit.rs b/fish-rust/src/builtins/emit.rs
index bb4a5c99f..112eff5e1 100644
--- a/fish-rust/src/builtins/emit.rs
+++ b/fish-rust/src/builtins/emit.rs
@@ -38,8 +38,3 @@ pub fn emit(
         .collect();
-    let event_arg_ptrs: Vec<ffi::wcharz_t> = event_args
-        .iter()
-        .map(|s| ffi::wcharz_t {
-            str_: s.as_ucstr().as_ptr().cast(),
-        })
-        .collect();
+    let event_arg_ptrs: Vec<ffi::wcharz_t> = event_args.iter().map(ffi::wcharz_t::from).collect();
 
diff --git a/fish-rust/src/wchar_ffi.rs b/fish-rust/src/wchar_ffi.rs
index cc00c1ea7..14844a567 100644
--- a/fish-rust/src/wchar_ffi.rs
+++ b/fish-rust/src/wchar_ffi.rs
@@ -45,2 +45,8 @@ impl From<&wcharz_t> for WString {
 
+impl From<&W0String> for wcharz_t {
+    fn from(s: &W0String) -> Self {
+        wcharz_t { str_: s.as_ptr() }
+    }
+}
+
 /// Convert a wstr or WString to a W0String, which contains a nul-terminator.
```

fish-rust/src/builtins/emit.rs
```diff
@@ -0,0 +1,51 @@
...
+        return STATUS_INVALID_ARGS;
+    };
+
+    let event_args: Vec<W0String> = argv[opts.optind + 1..]
+        .iter()
+        .map(|s| W0String::from_ustr(s).unwrap())
+        .collect();
+    let event_arg_ptrs: Vec<ffi::wcharz_t> = event_args
+        .iter()
+        .map(|s| ffi::wcharz_t {
```
krobelus, line 38: nit: doesn't really matter but I'm not a huge fan of qualifying `wcharz_t` because it's mostly used in FFI bridge definitions, where we don't qualify it

krobelus:
A conversion from `const wcharz_t *` to  `std::vector<std::wstring>` would only work if the `const wcharz_t *` is a null-terminated array of pointers.
I don't think we want to create that.

Anyway, the code looks fine as is.

fish-rust/src/builtins/emit.rs
```diff
@@ -0,0 +1,51 @@
...
+            str_: s.as_ucstr().as_ptr().cast(),
+        })
+        .collect();
+
+    ffi::event_fire_generic(
+        parser.pin(),
+        event_name.to_ffi(),
+        event_arg_ptrs.as_ptr(),
+        i32::try_from(event_arg_ptrs.len()).unwrap().into(),
```
krobelus, line 47: I think `i32` should be `c_int`. No idea if it makes a difference

fish-rust/src/builtins/emit.rs
Xiretza: > Also we can implement From<&W0String> for wcharz_t, in case we want to use this more often.
 (could also implement From<&Vec<W0String>> for Vec<wcharz_t>).
> Of course the conversion functions can be misused - it doesn't warn you if the W0String goes out of scope, which can leave the wcharz_t as dangling reference.

I considered implementing `From`, but decided against it because it's too easy to cause gnarly UB with it.

fish-rust/src/builtins/emit.rs
Xiretza: `c_int` is just a type alias for `i32`, so it doesn't make a difference, but I've changed it for clarity.

Xiretza:
> A conversion from `const wcharz_t *` to `std::vector<std::wstring>` would only work if the `const wcharz_t *` is a null-terminated array of pointers. I don't think we want to create that.

Here's what the conversion looked like originally: https://github.com/fish-shell/fish-shell/commit/2567bb1db986699f1470fdaf16882a12799db0f6#diff-36f9e9ac4ce412ab321946deb3c9ea1f28ebf8507e813c296609735fd5fdb396R493

i.e. converting a start and end `const wcharz_t *` to a `std::vector<std::wstring>`.

fish-rust/src/builtins/emit.rs
Xiretza: I've used `%ls` for now since `cmd` is utf32.

fish-rust/src/builtins/emit.rs
Xiretza: Yep, silly oversight, I had put `None` there while writing out the function to avoid type complaints, guess I should really use `todo!()` instead.

fish-rust/src/builtins/emit.rs
krobelus: Right. We don't want to make it too easy.
If we use this pattern more often we can extract an `unsafe` function.
One safe alternative would be to not pass `wcharz_t` to C++ but `std::unique_ptr<wcstring>` which can be created with `.to_ffi()`.

Issue #4656, opened 2018-01-11
Using block cursor only for normal state?


dieggsy:
|             |          | 
|-----------------|---------|
|fish version | `2.7.0`|
|uname -a | `Linux Ragnarok 4.14.12-1-ARCH #1 SMP PREEMPT Fri Jan 5 18:19:34 UTC 2018 x86_64 GNU/Linux`|
| TERM |   `xterm-termite` or `xterm-256color` (happens with both) | 
| `sh -c 'env HOME=$(mktemp -d) fish'` | behavior persists |

Basically, I use `fish_vi_key_bindings` along with:
```fish
set -U fish_cursor_insert line
set -U fish_cursor_replace_one underscore
```
However, when running a repl program like `python` (or really anything else where my cursor is visible, like `pacman -S` and other password prompting programs), the cursor I get is a block (normal mode) cursor. Is there a way for the default cursor in all instances/programs to be a line cursor, and only switch to a bar cursor when explicitly switching to normal mode?

dieggsy:
@zanchey That makes the cursor a line in those programs, but also a line in fish's normal mode. I'd like to still have a block cursor in normal mode.

I understand fish doesn't have control over what those programs do, but them having a block cursor is a direct result of fish changing the cursor to a block - since this is a fish-specific issue.

dieggsy:
To clarify, it would be nice to be able to do `set -U fish_cursor_default line` as well as something ilke `set -U fish_cursor_normal block`. The way it is now, normal mode (and all other programs run under fish) inherit from default cursor while insert cursor must be specified/changed. It'd be nice if the normal mode cursor could be specified or changed independently from default cursor.

faho:
So, this would require an actual code change. What we'd need to do is to introduce a new variable that describes the cursor outside of fish. So that, when we start a command, we use that to set the cursor shape instead of `fish_cursor_default`. Any idea for the name of that variable?

- `fish_cursor_outside`

- `fish_cursor_command`

- `fish_cursor_nonfish`

- `fish_cursor_exec`

?

dieggsy:
@faho couldn't you use fish_cursor_default for that and introduce a cursor for nomal mode? (I have no idea how this works, and you obviously know the code better, just wondering). If this isn't possible, I like `fish_cursor_exec`.

faho:
>couldn't you use fish_cursor_default for that

[...]

PR #9565, opened 2023-02-11
add new variable for external cursor mode

quintenroets:
## Description

Fixes issue #4656

## TODOs:

- [x] add variable for external cursor mode
- [x] update documentation
- [x] documented update in changelog.rst


share/functions/fish_vi_cursor.fish
```diff
@@ -92,7 +92,7 @@ function fish_vi_cursor -d 'Set cursor shape for different vi modes'
 
     echo "
           function fish_vi_cursor_handle_preexec --on-event fish_preexec
-              set -l varname fish_cursor_default
+              set -l varname fish_cursor_external
```
faho, line 95: This breaks every existing setup that expects the default key to carry over.

What it *should* do is to use the first of fish_cursor_external, fish_cursor_default and fish_cursor_unknown that is defined.

It would also need documentation.

faho (changes requested):

share/functions/fish_vi_cursor.fish
quintenroets: Thank you for your feedback!

quintenroets:
Please let me know if this pull request requires additional changes.

zanchey:
@faho any thoughts? probably won't take this for 3.6.1

faho:
Merged, thanks!

PR #9567, opened 2023-02-12
Rewrite `wrealpath` from `wutil` in Rust

Tyrubias:
## Description

Rewrite `wrealpath` from `wutil` in Rust. A lot of other builtins use this function so I thought it'd be useful to have before other builtins are worked on/merged. I'm not familiar enough with FFI to understand whether C++ can call a Rust function that returns `Option<WString>`. I'm new to the project so I'd love any feedback people have. Thanks!

Fixes issue #

## TODOs:


fish-rust/src/wchar_ext.rs
```diff
@@ -88,6 +88,11 @@ pub trait WExt {
             self.as_char_slice().iter().copied().rev(),
         )
     }
+
+    /// Equivalent of `basic_string.rfind()` from C++
+    fn find_last_char(&self, c: char) -> Option<usize> {
```
Xiretza, line 93: How about calling it `rfind`, like the [method on `str`](https://doc.rust-lang.org/std/primitive.str.html#method.rfind)?

fish-rust/src/wutil/wrealpath.rs
```diff
@@ -0,0 +1,59 @@
...
+
+/// Wide character realpath. The last path component does not need to be valid. If an error occurs,
+/// `wrealpath()` returns `None`
+pub fn wrealpath(pathname: &wstr) -> Option<WString> {
+    if pathname.is_empty() {
+        return None;
+    }
+
+    let mut real_path = WString::new();
+    let mut narrow_path = pathname.to_owned();
```
Xiretza, line 13: This doesn't correspond to what the C++ code does - the UTF-32 string should be decoded to UTF-8, including fish's Special Use Area trickery. See `wcs2string_callback`.

fish-rust/src/wutil/wrealpath.rs
```diff
@@ -0,0 +1,59 @@
...
+            real_path = narrow_path;
+        } else {
+            // Only call realpath() on the portion up to the last component.
+            if pathsep_idx.is_none() {
+                // If there is no "/", this is a file in $PWD, so give the realpath to that.
+                narrow_result = canonicalize(".");
+            } else {
+                // Only call realpath() on the portion up to the last component.
+                narrow_result = canonicalize(narrow_path[0..pathsep_idx.unwrap()].to_string());
+            }
```
Xiretza, line 40: Flipping the branches around, we can avoid the `.unwrap()` and the clippy warning it causes:

```suggestion
            if let Some(idx) = pathsep_idx {
                // Only call realpath() on the portion up to the last component.
                narrow_result = canonicalize(narrow_path[0..idx].to_string());
            } else {
                // If there is no "/", this is a file in $PWD, so give the realpath to that.
                narrow_result = canonicalize(".");
            }
```

fish-rust/src/wutil/wrealpath.rs
```diff
@@ -0,0 +1,59 @@
...
+
+            if narrow_result.is_err() {
+                return None;
+            }
```
Xiretza, line 44: No need to `.unwrap()` later if we conver it here:

```suggestion
            let Ok(narrow_result) = narrow_result else {
                return None;
            };
```

fish-rust/src/wutil/wrealpath.rs
```diff
@@ -0,0 +1,59 @@
...
+
+            pathsep_idx = pathsep_idx.map(|idx| idx + 1);
```
Xiretza, line 46: `None` corresponds to `npos` in the original code, which is defined as `-1` - thus `pathsep_idx++` is either `x + 1`, or `0` in case of `None`:

```suggestion
            let pathsep_idx = pathsep_idx.map_or(0, |idx| idx + 1);
```

fish-rust/src/wchar_ext.rs
Tyrubias: Sounds good, I wasn't sure whether to name it the same or name it like the `find_char` method in the same trait extension.

fish-rust/src/wutil/wrealpath.rs
Tyrubias: To confirm, this means we cannot use `widestring` to simply convert `pathname` from UTF-32 to UTF-8, right? We have to reimplement the `wcs2string*` family of functions?

Xiretza: Yes, I think so. Really the function argument shouldn't be a `&wstr` but rather a separate type that denotes fish's special string magic, but we don't have one of those.

Tyrubias: I have some questions about the special string magic.

Here is my understanding: In the C++ implementation of UTF-32 to UTF-8 string we convert anything in between `ENCODE_DIRECT_BASE` and `ENCODE_DIRECT_BASE + 256` (end-exclusive) into an ASCII character. This is done by subtracting `ENCODE_DIRECT_BASE` from `wc`, where `wc` is supposed to be a 4 byte `wchar_t`. We also do some locale-related hackery.

Here are my questions:

- How does the UTF-8 to UTF-32 conversion work with the directly encoded characters? It looks like we call `mbrtowc` to actually do the character conversion, then use direct encoding for a bunch of error cases. However, we also use direct encoding if the `wchar_t wc` produced by `mbrtowc` is in the `ENCODE_DIRECT_BASE` and `ENCODE_DIRECT_BASE + 256` range. We do this by adding `ENCODE_DIRECT_BASE` to the `unsigned char` representation of the input `char`. Why do we do this part? How does it work?
- It seems a lot of this is designed to handle idiosyncrasies in string handling between various platforms. How much of this is still needed when we rewrite it in Rust? For example, when we convert between UTF-32 and UTF-8 in Rust, we know that the `char` type in Rust is a valid Unicode scalar value. How does this relate to the direct encoding we do?

I will push some quick code I wrote in my attempt to directly translate the C++ string conversion. Thanks in advance! Please let me know if there's someone else I should reach out to regarding these questions.

Tyrubias: In addition, when we convert back, it doesn't seem like we expect the input to be valid UTF-8. Should we convert into something like `OsString` instead for a path-related function like `wrealpath`?

Tyrubias: I saw that we introduced `unixstring` as a dependency. I considered using it, but apparently it doesn't allow null bytes. In the original PR for the RIIR effort, someone suggested using the `bstr` crate. I was wondering if we should revisit that or if we should in some way or other convert the UTF-32 string to a `Vec<u8>` rather than try to treat it as a UTF-8 string.

zanchey:
I can't reopen this because the target branch has been deleted - @Tyrubias can you retarget to master?

PR #9563, opened 2023-02-11
RIIR event.cpp

Xiretza:
Not even close to done yet (there's no FFI), but I want to avoid toe-stepping.

PR #9568, opened 2023-02-12
builtins: port return/exit to rust

nrjais:
## Description

Port the `return/exit` builtin to Rust. There was an issues with parsing of long options cause `-1` to be parsed as first found long option after comparing the code to CPP I found that it was taking long options as first non exact match even when the option did not have matching prefix.

Fixes issue #

## TODOs:


src/parser.h
```diff
@@ -231,6 +231,9 @@ struct library_data_t {
         /// Used to get the full text of the current job for `status current-commandline`.
         wcstring commandline;
     } status_vars;
+
+    void set_exit_current_script(bool val);
+    void set_returning(bool val);
```
nrjais, line 236: Could not figure out how to update these directly from rust, so exposed setters for these

fish-rust/src/builtins/return_exit.rs
```diff
@@ -0,0 +1,148 @@
...
+#[derive(Debug, Clone, Copy)]
+struct Options {
+    print_help: bool,
+}
+
+impl Default for Options {
+    fn default() -> Self {
+        Self { print_help: false }
+    }
+}
```
Xiretza, line 26: clippy made me aware of this:

```suggestion
#[derive(Debug, Clone, Copy, Default)]
struct Options {
    print_help: bool,
}
```

fish-rust/src/builtins/return_exit.rs
```diff
@@ -0,0 +1,148 @@
...
+fn parse_return_value(args: &mut [&wstr], parser: &mut parser_t, streams: &mut io_streams_t) -> Result<i32, Option<c_int>> {
+    let cmd = args[0];
+    let (opts, optind) = match parse_options(args, parser, streams) {
+        Ok((opts, optind)) => (opts, optind),
+        Err(err @ Some(_)) if err != STATUS_CMD_OK => return Err(err),
+        Err(err) => panic!("Illogical exit code from parse_options(): {err:?}"),
+    };
+    if opts.print_help {
+        builtin_print_help(parser, streams, cmd);
+        return Err(STATUS_CMD_OK);
```
Xiretza, line 125: This isn't great, but I think I've been annoying enough about return value types that I'll let someone else weigh in on it instead.

fish-rust/src/builtins/return_exit.rs
```diff
@@ -0,0 +1,148 @@
...
+        streams
+            .err
+            .append(wgettext_fmt!(BUILTIN_ERR_TOO_MANY_ARGUMENTS, cmd));
+        builtin_print_error_trailer(parser, streams, cmd);
+        return Err(STATUS_INVALID_ARGS);
+    }
+    if optind == args.len() {
+        Ok(parser.get_last_status().into())
+    } else {
+        match fish_wcstoi(args[optind].as_char_slice().iter().copied()) {
```
Xiretza, line 137: This can be simplified a bit:

```suggestion
        match fish_wcstoi(args[optind].chars()) {
```

fish-rust/src/builtins/return_exit.rs
```diff
@@ -0,0 +1,148 @@
...
+pub fn exit(
+    parser: &mut parser_t,
+    streams: &mut io_streams_t,
+    args: &mut [&wstr],
+) -> Option<c_int> {
+    let retval =  match parse_return_value(args, parser, streams) {
+        Ok(v) => v,
+        Err(e) => return e,
+    };
+
```
Xiretza, line 110: Missing comment from original code:

```suggestion

    // Mark that we are exiting in the parser.
    // TODO: in concurrent mode this won't successfully exit a pipeline, as there are other parsers
    // involved. That is, `exit | sleep 1000` may not exit as hoped. Need to rationalize what
    // behavior we want here.
```

src/parser.h
Xiretza: https://github.com/fish-shell/fish-shell/pull/9563/commits/a7aa497ff95db61aa8b884371b737b9a02066a2c might be a solution here, but I don't know if it's a good one.

Xiretza (commented):
Looks good to me overall, please run rustfmt.

src/parser.h
nrjais: Yeah, that also will work, I will leave it like this for now, After your changes are merged then I can update here

fish-rust/src/builtins/return_exit.rs
```diff
@@ -0,0 +1,150 @@
+// Implementation of the return builtin.
+
```
krobelus, line 2: I don't think we should put both in the same file because that's inconsistent.
Rust makes it easy to share code across files.

I think we should name the files `return.rs` and `exit.rs`.
(we'll need to declare the module like `pub mod r#return;`)

fish-rust/src/builtins/return_exit.rs
```diff
@@ -0,0 +1,150 @@
...
+use libc::c_int;
+use num_traits::abs;
+
```
krobelus, line 5: nit: the third commit is a fixup of the first two so it should be squashed

fish-rust/src/builtins/return_exit.rs
```diff
@@ -0,0 +1,150 @@
...
+/// Function for handling the exit builtin.
+pub fn exit(
+    parser: &mut parser_t,
+    streams: &mut io_streams_t,
+    args: &mut [&wstr],
+) -> Option<c_int> {
+    let retval = match parse_return_value(args, parser, streams) {
+        Ok(v) => v,
+        Err(e) => return e,
+    };
```
krobelus, line 103: nit: Could use the question-mark operator here, like

	let retval =  match parse_return_value(args, parser, streams)?;

fish-rust/src/builtins/shared.rs
```diff
@@ -31,6 +31,9 @@ mod builtins_ffi {
     impl Vec<wcharz_t> {}
 }
 
+pub const BUILTIN_ERR_TOO_MANY_ARGUMENTS: &str = "%ls: too many arguments\n";
+pub const BUILTIN_ERR_NOT_NUMBER: &str = "%ls: %ls: invalid integer\n";
```
krobelus, line 35: should carry over comments

fish-rust/src/builtins/shared.rs
```diff
@@ -111,6 +114,8 @@
     match builtin {
         RustBuiltin::Echo => super::echo::echo(parser, streams, args),
         RustBuiltin::Emit => super::emit::emit(parser, streams, args),
+        RustBuiltin::Return => super::return_exit::r#return(parser, streams, args),
+        RustBuiltin::Exit => super::return_exit::exit(parser, streams, args),
```
krobelus, line 118: please sort

fish-rust/src/builtins/return_exit.rs
nrjais: I tried it, but because the builtin returns Option<> and the parse_return_value return Result<i32, Option<>> ? mark operator can't be used.

src/builtin.cpp
```diff
@@ -536,6 +534,12 @@ static maybe_t<RustBuiltin> try_get_rust_builtin(const wcstring &cmd) {
     if (cmd == L"wait") {
         return RustBuiltin::Wait;
     }
+    if (cmd == L"return") {
+        return RustBuiltin::Return;
+    }
+    if (cmd == L"exit") {
+        return RustBuiltin::Exit;
+    }
```
Xiretza, line 542: Please sort these.

src/builtin.h
```diff
@@ -112,5 +112,7 @@ enum RustBuiltin : int32_t {
     Echo,
     Emit,
     Wait,
+    Return,
+    Exit,
```
Xiretza, line 116: Ditto.

fish-rust/src/wgetopt.rs
```diff
@@ -451,18 +451,20 @@ impl<'opts, 'args, 'argarray> wgetopter_t<'opts, 'args, 'argarray> {
...
+                    *exact = true;
+                    break;
+                } else if pfound.is_none() {
+                    // First nonexact match found.
+                    pfound = Some(*p);
+                    *indfound = option_index;
+                } else {
+                    // Second or later nonexact match found.
+                    *ambig = true;
+                }
```
Xiretza, line 467: Would be nice to have this in a separate commit with some explanation, it's not immediately obvious what's happening here.

nrjais: I am also not sure what is happening here, I found the issue by comparing this with actual master CPP code and found the logic issue. I can put some explanation on what I encountered and explain

nrjais: Have updated the comment let if know what you think

krobelus:
Thanks.

The commit order should be reversed so that every commit passes all tests. No big deal though.

I tried porting libdata but haven't succeeded so far.
I made it an opaque Rust type but it still lives in the C++ parser_t.
I exposed it to Rust as `void*`.
Somehow C++ doesn't see the updates made by Rust, and autocxx doesn't seem to support volatile.
Probably it's better to port the entire parser_t.

PR #9569, opened 2023-02-12
fix 3 instances of old command substitution `$()`

mwartell:
## Description

On installation of fish 3.6.0 on Linux, every prompt fails with a syntax error

```
/usr/share/fish/functions/fish_git_prompt.fish (line 173): $(...) is not supported. In fish, please use '(uname)'.
if string match -q Darwin -- "$(uname)" && string match -q /usr/bin/git -- "$(command -s git)" && type -q xcode-select && type -q xcrun
                              ^
from sourcing file /usr/share/fish/functions/fish_git_prompt.fish
        called on line 1 of file ~/.config/fish/functions/fish_prompt.fish
in command substitution
        called on line 1 of file ~/.config/fish/functions/fish_prompt.fish
in command substitution
        called on line 28 of file ~/.config/fish/functions/fish_prompt.fish
in function 'fish_prompt'
in command substitution
```

The commit a77bc70defd47db9c401127465d7e71d05184200 added those old-style substitutions.

Fixes issue #

## TODOs:

### DIFF
--- a/share/functions/fish_git_prompt.fish
+++ b/share/functions/fish_git_prompt.fish
@@ -170,7 +170,7 @@ end
 
 # Decide if git is safe to run.
 # On Darwin, git is pre-installed as a stub, which will pop a dialog if you run it.
-if string match -q Darwin -- "$(uname)" && string match -q /usr/bin/git -- "$(command -s git)" && type -q xcode-select && type -q xcrun
+if string match -q Darwin -- (uname) && string match -q /usr/bin/git -- (command -s git) && type -q xcode-select && type -q xcrun
     if not xcode-select --print-path &>/dev/null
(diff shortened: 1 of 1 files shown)


faho:
>On installation of fish 3.6.0 on Linux, every prompt fails with a syntax error

The reason this happens is that you have an *old* fish running - pre-[3.4](https://github.com/fish-shell/fish-shell/blob/master/CHANGELOG.rst#fish-340-released-march-12-2022).

What you *need* to do, basically after every fish update, is to stop these old fishes and restart them with the new version. This is a consequence of autoloading - there is no way to tell the old fish to never load a new function.

That being said: In this specific case it's not a big deal to just use the old style a bit longer, so I'll be merging this.

Issue #9572, opened 2023-02-13
Display only stash count


Dreamsorcerer:
I'd like to be able to display just the stash count, but the current version of fish seems to have broken this (it mostly worked on 3.1, which is the version installed on Ubuntu 20.04).

If I do:
```
set -U __fish_git_prompt_show_informative_status 1
set -U __fish_git_prompt_showstashstate 1
set -U __fish_git_prompt_status_order stashstate
```

Then the count will not be displayed. I think that's because this disables the informative status entirely:
https://github.com/fish-shell/fish-shell/blob/master/share/functions/fish_git_prompt.fish#L259-L265

Additionally, I'd also point out that it was a little confusing when setting these variables, as to why there were global variables shadowing them. It seems the script sets a global variable if it's not already set, not sure if maybe that should be changed to a local variable to be less confusing?
https://github.com/fish-shell/fish-shell/blob/master/share/functions/fish_git_prompt.fish#L229

Dreamsorcerer:
Wrote a test in #9573 to reproduce it.

Just realised that the variable has also been renamed since then too (2 underscores instead of 3 now).

faho:
> Just realised that the variable has also been renamed since then too (2 underscores instead of 3 now).

I'm reasonably sure it hasn't been. The version with 3 underscores is the cached "effective" version.

The git prompt has a bunch of knobs that influence each other, and so we save the result in a separate variable.

>Then the count will not be displayed. I think that's because this disables the informative status entirely:

This disables the informative status because that is extremely slow in larger repositories because it needs to call `git status` to count all files, as opposed to git calls that can quit early once they found something. So we disable it if the status call isn't necessary, to avoid people getting prompts that are slow by doing useless work.

It would be possible to still count the cache. This would require a change to the non-informative path.

Dreamsorcerer:
> > Just realised that the variable has also been renamed since then too (2 underscores instead of 3 now).
> 
> I'm reasonably sure it hasn't been. The version with 3 underscores is the cached "effective" version.

Hmm, on my Ubuntu 20.04 machine setting the variable with 2 underscores does nothing (I ensured both versions were erased first), while on the Mac setting the one with 3 underscores does nothing...

Though I'm still having issues with the default value appearing as a global variable from time to time, and then shadowing my setting...

> It would be possible to still count the cache. This would require a change to the non-informative path.

OK, that sounds reasonable. I'll see if I can figure out a fix then.

faho:
[...]

PR #9573, opened 2023-02-13
Test displaying only stash count

Dreamsorcerer:
Test for #9572.

Still needs the fix to be implemented.


faho:
Through the magic of rebasing, we should now see this miraculously pass on master (because I just fixed it).

PR #9579, opened 2023-02-15
port abbr.cpp and abbr builtin to rust 

nrjais:
## Description

Talk about your changes here.

Fixes issue #

## TODOs:


nrjais:
@krobelus @Xiretza What is the plan on porting regex to rust? I am currently porting abbreviations to rust and am stuck on exposing `regex_t` I think because of non_copy marker. 
Is there any better way to expose it rust as reference instead of by value?

fish-rust/src/abbrs.rs
```diff
@@ -0,0 +1,442 @@
...
+    // This is used as the token to match unless we have a regex.
+    pub name: WString,
+
+    /// The key (recognized token) - either a literal or a regex pattern.
+    pub key: WString,
+
+    /// If set, use this regex to recognize tokens.
+    /// If unset, the key is to be interpreted literally.
+    /// Note that the fish interface enforces that regexes match the entire token;
+    /// we accomplish this by surrounding the regex in ^ and $.
```
krobelus, line 131: yeah`regex: Option<regex_t>,` will not work but you can make it a `UniquePtr<regex_t>.
I think that works. If we run into problems we can port `re.h` first, it has no dependencies.

PR #9585, opened 2023-02-17
Port random builtin to rust

faho:
## Description

This ports `random` to rust.

The remaining issue here is some weirdness with how the C++ does the numbers - the tests explicitly exercise that you can declare random numbers from the smallest signed long long to the largest with a step size of the difference between the two, and I haven't been able to make that work.

Also the error messages are a bit of a mess in the C++ - a step of "-1" is "invalid integer", but a step of "0" is "STEP must be a non-negative integer" for some reason?

This uses a new crate, rand, which is what seemed to be the recommended course if you needed a seedable RNG, which we do (`random 12` seeds the RNG with "12").

Also I needed *some* way to get wcstoi to tell me it hasn't read the entire token, so I picked a brute-force approach. I would assume there's a more elegant approach?

This represents about 98% of all the rust code I've ever written, so let's see how much I missed!

## TODOs:

### DIFF
--- a/fish-rust/src/builtins/random.rs
+++ b/fish-rust/src/builtins/random.rs
@@ -118,11 +118,13 @@ pub fn random(
                 Ok(x) => start = x
             }
 
-            match parse_ll(argv[i + 1]) {
-                Err(_) => return STATUS_INVALID_ARGS,
-                Ok(x) if x < 0 => {
-                    // XXX: Historical, this read an "unsigned long long"
-                    // This should actually be "must be a positive integer"
+            // Unfortunately we can't use parse_ll here because step is a u64 so it can go
+            // from i64::MIN to i64::MAX
+            //
+            // If we figured out how to tell it to return a u64 we could use it
+            let res: Result<u64, wutil::Error> = fish_wcstoi_radix_all(argv[i + 1].chars(), None, true);
+            match res {
+                Err(_) => {
                     streams.err.append(wgettext_fmt!(
                         "%ls: %ls: invalid integer\n",
                         cmd,
@@ -140,8 +142,17 @@ pub fn random(
                 Ok(x) => step = x
             }
 
-            match parse_ll(argv[i + 2]) {
-                Err(_) => return STATUS_INVALID_ARGS,
+            // XXX Because we've used streams.err above, we can't use parse_ll here because
+            // it needs to be borrowed mutably apparently?
+            let res: Result<i64, wutil::Error> = fish_wcstoi_radix_all(argv[i + 2].chars(), None, true);
+            match res {
+                Err(_) => {
+                    streams.err.append(wgettext_fmt!(
+                        "%ls: %ls: invalid integer\n",
+                        cmd,
+                        argv[i + 1],
+                    ));
+                },
                 Ok(x) => end = x
             }
         },
@@ -162,24 +173,11 @@ pub fn random(
         return STATUS_INVALID_ARGS;
     }
 
-    let real_end : i64 = if start >= 0 || end < 0 {
-        // 0 <= start <= end
-        let diff : i64 = end - start;
-        // 0 <= diff <= LL_MAX
-        start + diff / step
-    } else {
-        // This is based on a "safe_abs" function
-        // in the C++ that just casted a "long long" to "-unsigned long long"
-        // start < 0 <= end
-        let abs_start : u64 = start.abs() as u64;
-        let diff : u64 = (end as u64 + abs_start) as u64;
-        (diff / (step as u64) - abs_start) as i64
-    };
+    // Possibilities can be abs(i64::MIN) + i64::MAX,
+    // so we do this as i128
+    let possibilities = (end as i128 - start as i128) / (step as i128);
 
-    let a = if start < real_end { start } else { real_end };
-    let b = if start > real_end { start } else { real_end };
-
-    if a == b {
+    if possibilities == 0 {
         streams.err.append(wgettext_fmt!(
             "%ls: range contains only one possible value\n",
             cmd,
@@ -187,16 +185,31 @@ pub fn random(
         return STATUS_INVALID_ARGS;
     }
 
-    let rand = engine.gen_range(a..b);
+    let rand = engine.gen_range(0..=possibilities);
 
-    let result = if start >= 0 {
-        start + (rand - start) * step
-    } else if rand < 0 {
-        (rand - start) * step - start.abs()
-    } else {
-        (rand + start.abs()) * step - start.abs()
-    };
-    streams.out.append(format::printf::sprintf!("%d\n"L, result));
+    let result = start as i128 + rand as i128 * step as i128;
 
-    return STATUS_CMD_OK;
+    // We do our math as i128, 
+    // and then we check if it fits in 64 bit - signed or unsigned!
+    match i64::try_from(result) {
+        Ok(x) => {
+            streams.out.append(format::printf::sprintf!("%d\n"L, x));
+            return STATUS_CMD_OK;
+        },
+        Err(_) => {
+            match u64::try_from(result) {
+                Ok(x) => {
+                    streams.out.append(format::printf::sprintf!("%d\n"L, x));
+                    return STATUS_CMD_OK;
+                },
+                Err(_) => {
+                    streams.err.append(wgettext_fmt!(
+                        "%ls: range contains only one possible value\n",
+                        cmd,
+                    ));
+                    return STATUS_INVALID_ARGS;
+                },
+            }
+        },
+    }
 }
--- a/fish-rust/src/builtins/random.rs
+++ b/fish-rust/src/builtins/random.rs
@@ -1,0 +1,209 @@
+use libc::{c_int};
+
+use crate::builtins::shared::{
+    builtin_missing_argument, builtin_print_help, builtin_unknown_option, io_streams_t,
+    STATUS_CMD_OK, STATUS_INVALID_ARGS,
+};
+use crate::ffi::{parser_t};
+use crate::wchar::{widestrs, wstr};
+use crate::wgetopt::{wgetopter_t, wopt, woption, woption_argument_t};
+use crate::wutil::{self, format, fish_wcstoi_radix_all, wgettext_fmt};
+use rand::{Rng, SeedableRng};
+use rand::rngs::SmallRng;
+use once_cell::sync::Lazy;
+use std::sync::Mutex;
+
+static seeded_engine: Lazy<Mutex<SmallRng>> = Lazy::new(|| { Mutex::new(SmallRng::from_entropy()) });
+
+
+#[widestrs]
+pub fn random(
+    parser: &mut parser_t,
+    streams: &mut io_streams_t,
+    argv: &mut [&wstr],
+) -> Option<c_int> {
+    let cmd = argv[0];
+    let argc = argv.len();
+    let print_hints = false;
+
+    const shortopts: &wstr = "+:h"L;
+    const longopts: &[woption] = &[
+        wopt("help"L, woption_argument_t::no_argument, 'h'),
+    ];
+
+    let mut w = wgetopter_t::new(shortopts, longopts, argv);
+    while let Some(c) = w.wgetopt_long() {
+        match c {
+            'h' => {
+                builtin_print_help(parser, streams, cmd);
+                return STATUS_CMD_OK;
+            }
+            ':' => {
+                builtin_missing_argument(parser, streams, cmd, argv[w.woptind - 1], print_hints);
+                return STATUS_INVALID_ARGS;
+            }
+            '?' => {
+                builtin_unknown_option(parser, streams, cmd, argv[w.woptind - 1], print_hints);
+                return STATUS_INVALID_ARGS;
+            }
+            _ => {
+                panic!("unexpected retval from wgeopter.next()");
+            }
+        }
+    }
+
+    let mut engine = seeded_engine.lock().unwrap();
+    let mut start = 0;
+    let mut end = 32767;
+    let mut step = 1;
+    let arg_count = argc - w.woptind;
+    let i = w.woptind;
+    if arg_count >= 1 && argv[i] == "choice" {
+        if arg_count == 1 {
+            streams.err.append(wgettext_fmt!(
+                "%ls: nothing to choose from\n",
+                cmd,
+            ));
+            return STATUS_INVALID_ARGS;
+        }
+
+        let rand = engine.gen_range(0..arg_count - 1);
+        streams.out.append(format::printf::sprintf!("%ls\n"L, argv[i + 1 + rand]));
+        return STATUS_CMD_OK;
+    }
+    let mut parse_ll = |num : &wstr| {
+        let res: Result<i64, wutil::Error> = fish_wcstoi_radix_all(num.chars(), None, true);
+        if res.is_err() || res.unwrap() < 0 {
+            match res {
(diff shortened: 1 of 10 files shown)


fish-rust/src/builtins/random.rs
```diff
@@ -0,0 +1,209 @@
...
+                    streams.err.append(wgettext_fmt!(
+                        "%ls: %ls: invalid integer\n",
+                        cmd,
+                        num,
+                    ));
+                },
+                Ok(_) => {}
+            }
+        }
+        return res;
```
mqudsi, line 88: You're checking if `res.is_err() || res.unwrap() < 0` but you're only handling the case where it's an error?

fish-rust/src/builtins/random.rs
```diff
@@ -0,0 +1,209 @@
...
+                return STATUS_INVALID_ARGS;
+            }
+            if num.unwrap() == 0 {
+                streams.err.append(wgettext_fmt!(
+                    "%ls: STEP must be a positive integer\n",
+                    cmd,
+                ));
+                return STATUS_INVALID_ARGS;
+            }
+            step = num.unwrap();
```
mqudsi, line 147: You have a lot of calls to `.unwrap()` but they're all cases where you can avoid that altogether by using match blocks. That's better because you never want to get into the habit of glossing over `.unwrap()` because you use it liberally.

You can either do 

```let num = match num {
    Ok(num) => num,
    Err(e) => { streams.err.append(...); return STATUS_XXX; }
};

// then continue to just use `num` instead of `num.unwrap()` thereafter
```

or you can convert the entire highlighted block into one match block, which is probably more idiomatic because it lets the compiler do things like check for untested branches, check for duplicate branches, check for branches that can't be taken because of a previous check, etc.

```rust
let step = match num {
    Err(e) => // as before,
    Ok(x) if x < 0 => ..., // or, more compactly, Ok(..0) => ...
    Ok(0) => ....,
    Ok(x) => x,
};
```

mqudsi (commented):
Hey, great job! You probably want to remove `random.cpp` and `random.h` from the codebase altogether so it's clearer where everything is implemented. 

This PR reminds me that anywhere we accept arbitrary input from the user and perform math, we're going to have to use the checked addition/subtraction/multiplication/division/abs/modulo functions so we can handle overflows without panicking.

fish-rust/src/builtins/random.rs
faho: Yeah, this is a helper function to not duplicate that one error six times (specifically since so far we haven't gone back to the BUILTIN_ERR_NOT_NUMBER constant stuff).

If I could cause this to return from the outer function I would, but so far I've not found anything (using `?` doesn't work because the builtin itself doesn't return a Result).

faho: Ah, now I see what you mean.

I'm reasonably sure the `< 0` part was an editing artifact.

fish-rust/src/builtins/random.rs
faho: Okay. I left the one for the rng because that's something we need to look at regarding concurrent execution.

faho:
>we're going to have to use the checked addition/subtraction/multiplication/division/abs/modulo functions so we can handle overflows without panicking.

It just feels like so much progress that those *exist* at all.

Anyway, what I *haven't* been able to find is discussion of how to do math with *both* signed and unsigned types - the issue is that the STEP value is unsigned, and so you can have

```fish
set -l max 9223372036854775807
set -l min -9223372036854775807
set -l diff_max 18446744073709551614

random -- $min $diff_max $max
```

which, in C++, will work out and randomply pick either the minimum long long or the maximum long long. (same for just having a range that's larger than the maximum long long in general, like going from the minimum to 0)

So what I have now resorted to is to do all the math in i128. I realize that by doing that I've trampled all over the "beautiful" math that was in here originally, and it might be a tad slower, depending on hardware support, compiler version, etc?

I would be all ears if anyone has a better solution, but I've been banging my head against this for a while now and I have no idea how else to do it.

krobelus:
Re matching the existing behavior, the relevant differences are
1. Rust has no implicit arithmetic conversions, so we need to write out some of the casts
2. Rust panics on overflow (while in C++ overflow is well-defined for unsigned integers) so we need to use functions like [u64::wrapping_add](https://doc.rust-lang.org/std/primitive.u64.html#method.wrapping_add)

here's a very quick and dirty proof-of-concept (based on previous push) that demonstrates the idea.
I'm sure it's wrong in some places but that one failing test passes.

(I removed widestrs because it obscures error locations, so the backtraces are not really useful)

One more thing, parse() should probably return an Option<> if we never read the actual error

<details>

```diff
diff --git a/fish-rust/src/builtins/random.rs b/fish-rust/src/builtins/random.rs
index d82f8edc1..280c57874 100644
--- a/fish-rust/src/builtins/random.rs
+++ b/fish-rust/src/builtins/random.rs
@@ -1,22 +1,22 @@
-use libc::{c_int};
+use libc::c_int;
 
 use crate::builtins::shared::{
     builtin_missing_argument, builtin_print_help, builtin_unknown_option, io_streams_t,
     STATUS_CMD_OK, STATUS_INVALID_ARGS,
 };
-use crate::ffi::{parser_t};
-use crate::wchar::{widestrs, wstr};
+use crate::ffi::parser_t;
+use crate::wchar::{widestrs, wstr, L};
 use crate::wgetopt::{wgetopter_t, wopt, woption, woption_argument_t};
-use crate::wutil::{self, format, fish_wcstoi_radix_all, wgettext_fmt};
-use rand::{Rng, SeedableRng};
-use rand::rngs::SmallRng;
+use crate::wutil::{self, fish_wcstoi_radix_all, format, wgettext_fmt};
+use num_traits::PrimInt;
 use once_cell::sync::Lazy;
+use rand::rngs::SmallRng;
+use rand::{Rng, SeedableRng};
 use std::sync::Mutex;
 
-static seeded_engine: Lazy<Mutex<SmallRng>> = Lazy::new(|| { Mutex::new(SmallRng::from_entropy()) });
+static seeded_engine: Lazy<Mutex<SmallRng>> = Lazy::new(|| Mutex::new(SmallRng::from_entropy()));
 
-
-#[widestrs]
+// #[widestrs]
 pub fn random(
     parser: &mut parser_t,
     streams: &mut io_streams_t,
@@ -26,10 +26,8 @@ pub fn random(
     let argc = argv.len();
     let print_hints = false;
 
-    const shortopts: &wstr = "+:h"L;
-    const longopts: &[woption] = &[
-        wopt("help"L, woption_argument_t::no_argument, 'h'),
-    ];
+    const shortopts: &wstr = L!("+:h");
+    const longopts: &[woption] = &[wopt(L!("help"), woption_argument_t::no_argument, 'h')];
 
     let mut w = wgetopter_t::new(shortopts, longopts, argv);
     while let Some(c) = w.wgetopt_long() {
@@ -55,125 +53,111 @@ pub fn random(
     let mut engine = seeded_engine.lock().unwrap();
     let mut start = 0;
     let mut end = 32767;
     let mut step = 1;
     let arg_count = argc - w.woptind;
     let i = w.woptind;
     if arg_count >= 1 && argv[i] == "choice" {
         if arg_count == 1 {
-            streams.err.append(wgettext_fmt!(
-                "%ls: nothing to choose from\n",
-                cmd,
-            ));
+            streams
+                .err
+                .append(wgettext_fmt!("%ls: nothing to choose from\n", cmd,));
             return STATUS_INVALID_ARGS;
         }
 
         let rand = engine.gen_range(0..arg_count - 1);
-        streams.out.append(format::printf::sprintf!("%ls\n"L, argv[i + 1 + rand]));
+        streams
+            .out
+            .append(format::printf::sprintf!(L!("%ls\n"), argv[i + 1 + rand]));
         return STATUS_CMD_OK;
     }
-    let mut parse_ll = |num : &wstr| {
-        let res: Result<i64, wutil::Error> = fish_wcstoi_radix_all(num.chars(), None, true);
+    fn parse<T: PrimInt>(streams: &mut io_streams_t, num: &wstr) -> Result<T, wutil::Error> {
+        let res = fish_wcstoi_radix_all(num.chars(), None, true);
         match res {
             Err(_) => {
-                streams.err.append(wgettext_fmt!(
-                    "%ls: %ls: invalid integer\n",
-                    cmd,
-                    num,
-                ));
-            },
+                streams
+                    .err
+                    .append(wgettext_fmt!("%ls: %ls: invalid integer\n", "random", num));
+            }
             Ok(_) => {}
         }
         return res;
-    };
+    }
 
     match arg_count {
         0 => {
             // Keep the defaults
-        },
+        }
         1 => {
             // Seed the engine persistently
-            let num = parse_ll(argv[i]);
+            let num = parse::<i64>(streams, argv[i]);
             match num {
                 Err(_) => return STATUS_INVALID_ARGS,
-                Ok(x) => *engine = SmallRng::seed_from_u64(x as u64)
+                Ok(x) => *engine = SmallRng::seed_from_u64(x as u64),
             }
             return STATUS_CMD_OK;
-        },
+        }
         2 => {
             // start is first, end is second
-            match parse_ll(argv[i]) {
+            match parse::<i64>(streams, argv[i]) {
                 Err(_) => return STATUS_INVALID_ARGS,
-                Ok(x) => start = x
+                Ok(x) => start = x,
             }
 
-            match parse_ll(argv[i + 1]) {
+            match parse::<i64>(streams, argv[i + 1]) {
                 Err(_) => return STATUS_INVALID_ARGS,
-                Ok(x) => end = x
+                Ok(x) => end = x,
             }
-        },
+        }
         3 => {
             // start, step, end
-            match parse_ll(argv[i]) {
+            match parse::<i64>(streams, argv[i]) {
                 Err(_) => return STATUS_INVALID_ARGS,
-                Ok(x) => start = x
+                Ok(x) => start = x,
             }
 
-            match parse_ll(argv[i + 1]) {
+            match parse::<u64>(streams, argv[i + 1]) {
                 Err(_) => return STATUS_INVALID_ARGS,
-                Ok(x) if x < 0 => {
-                    // XXX: Historical, this read an "unsigned long long"
-                    // This should actually be "must be a positive integer"
-                    streams.err.append(wgettext_fmt!(
-                        "%ls: %ls: invalid integer\n",
-                        cmd,
-                        argv[i + 1],
-                    ));
-                    return STATUS_INVALID_ARGS;
-                },
                 Ok(0) => {
-                    streams.err.append(wgettext_fmt!(
-                        "%ls: STEP must be a positive integer\n",
-                        cmd,
-                    ));
+                    streams
+                        .err
+                        .append(wgettext_fmt!("%ls: STEP must be a positive integer\n", cmd,));
                     return STATUS_INVALID_ARGS;
                 }
-                Ok(x) => step = x
+                Ok(x) => step = x,
             }
 
-            match parse_ll(argv[i + 2]) {
+            match parse::<i64>(streams, argv[i + 2]) {
                 Err(_) => return STATUS_INVALID_ARGS,
-                Ok(x) => end = x
+                Ok(x) => end = x,
             }
-        },
+        }
         _ => {
-            streams.err.append(wgettext_fmt!(
-                "%ls: too many arguments\n",
-                cmd,
-            ));
+            streams
+                .err
+                .append(wgettext_fmt!("%ls: too many arguments\n", cmd,));
             return Some(1);
         }
     }
 
     if end <= start {
-        streams.err.append(wgettext_fmt!(
-            "%ls: END must be greater than START\n",
-            cmd,
-        ));
+        streams
+            .err
+            .append(wgettext_fmt!("%ls: END must be greater than START\n", cmd,));
         return STATUS_INVALID_ARGS;
     }
 
-    let real_end : i64 = if start >= 0 || end < 0 {
+    let real_end: i64 = if start >= 0 || end < 0 {
         // 0 <= start <= end
-        let diff : i64 = end - start;
+        let diff: i64 = end - start;
         // 0 <= diff <= LL_MAX
-        start + diff / step
+        start + (diff as u64 / step) as i64
     } else {
         // This is based on a "safe_abs" function
         // in the C++ that just casted a "long long" to "-unsigned long long"
         // start < 0 <= end
-        let abs_start : u64 = start.abs() as u64;
-        let diff : u64 = (end as u64 + abs_start) as u64;
-        (diff / (step as u64) - abs_start) as i64
+        let abs_start: u64 = start.abs() as u64;
+        let diff: u64 = (end as u64 + abs_start) as u64;
+        ((diff / step).wrapping_sub(abs_start)) as i64
     };
 
     let a = if start < real_end { start } else { real_end };
@@ -190,13 +174,15 @@ pub fn random(
     let rand = engine.gen_range(a..b);
 
     let result = if start >= 0 {
-        start + (rand - start) * step
+        start + (rand - start) * (step as i64)
     } else if rand < 0 {
-        (rand - start) * step - start.abs()
+        (rand - start) * step as i64 - start.abs()
     } else {
-        (rand + start.abs()) * step - start.abs()
+        ((rand + start.abs()) as u64 * step) as i64 - start.abs()
     };
-    streams.out.append(format::printf::sprintf!("%d\n"L, result));
+    streams
+        .out
+        .append(format::printf::sprintf!(L!("%d\n"), result));
 
     return STATUS_CMD_OK;
 }
```

</details>

fish-rust/src/wutil/wcstoi.rs
```diff
@@ -102,11 +105,12 @@ where
...
     }
-    Ok(ParseResult { result, negative })
+    let consumed_all = chars.peek() == None;
+    Ok(ParseResult { result, negative, consumed_all })
 }
 
 /// Parse some iterator over Chars into some Integer type, optionally with a radix.
-fn fish_wcstoi_impl<Int, Chars>(src: Chars, mradix: Option<u32>) -> Result<Int, Error>
+fn fish_wcstoi_impl<Int, Chars>(src: Chars, mradix: Option<u32>, consume_all: bool) -> Result<Int, Error>
 where
```
krobelus, line 114: Another solution is to take a mutable iterator (like `src: &mut Chars`),
then the caller can check `src.next().is_some()`.
But that probably causes other inconveniences, I think it's fine.

fish-rust/src/builtins/random.rs
```diff
@@ -0,0 +1,215 @@
...
+        if arg_count == 1 {
+            streams.err.append(wgettext_fmt!(
+                "%ls: nothing to choose from\n",
+                cmd,
+            ));
+            return STATUS_INVALID_ARGS;
+        }
+
+        let rand = engine.gen_range(0..arg_count - 1);
+        streams.out.append(format::printf::sprintf!("%ls\n"L, argv[i + 1 + rand]));
```
krobelus, line 71: `sprintf` I'd use as unqualified name. rust-analyzer has helpful code actions to add import statements

mqudsi:
Note that rust doesn’t panic on overflow in release builds by default, though that can be enabled - and I believe it’s also not a guarantee that it will never panic in release. . Not sure why it panics since the default CMake is RelWithDebInfo but I don’t know how Corrosion translates that. It’s not built as regular debug though because  I had to do a CMake debug build to get mine numbers in the stack trace.

fish-rust/src/builtins/random.rs
```diff
@@ -0,0 +1,215 @@
...
+        return STATUS_CMD_OK;
+    }
+    let mut parse_ll = |num : &wstr| {
+        let res: Result<i64, wutil::Error> = fish_wcstoi_radix_all(num.chars(), None, true);
+        match res {
```
mqudsi, line 76: If you’re only handling the error case and *not* unwrapping the Ok variant (just passing it along) it is better to use if let instead of match because that eliminates the needless all branch. 

```rust
if let Err(e) = res {
   // handle e by writing to stream
}
return res;
```

PR #9586, opened 2023-02-18
Port fd_monitor (and its needed components)

mqudsi:
I needed to rename some types already ported to rust so they don't clash with their still-extant cpp counterparts. Helper ffi functions added to avoid needing to dynamically allocate an FdMonitorItem for every fd (we use dozens per basic prompt).

I ported some functions from cpp to rust that are used only in the backend but without removing their existing cpp counterparts so cpp code can continue to use their version of them (`wperror` and `make_detached_pthread`).

I ran into issues porting line-by-line logic because rust inverts the behavior of `std::remove_if(..)` by making it (basically) `Vec::retain_if(..)` so I replaced bools with an explict enum to make everything clearer.

I'll port the cpp tests for this separately, for now they're using ffi.

Porting closures was ugly. It's nothing hard, but it's very ugly as now each capturing lambda has been changed into an explicit struct that contains its parameters (that needs to be dynamically allocated), a standalone callback (member) function to replace the lambda contents, and a separate trampoline function to call it from rust over the shared C abi (not really relevant to x86_64 w/ its single calling convention but probably needed on other platforms).

I don't like that `fd_monitor.rs` has its own `c_void`. I couldn't find a way to move that to `ffi.rs` but still get cxx bridge to consider it a shared POD. Every time I moved it to a different module, it would consider it to be an opaque rust type instead. I worry this means we're going to have multiple `c_void1`, `c_void2`, etc. types as we continue to port code to use function pointers.

Also, rust treats raw pointers as foreign so you can't do `impl Send for * const Foo` even if `Foo` is from the same module. That necessitated a wrapper type (`void_ptr`) that implements `Send` and `Sync` so we can move stuff between threads.

The code in fd_monitor_t has been split into two objects, one that is used by the caller and a separate one associated with the background thread (this is made nice and clean by rust's ownership model). Objects not needed under the lock (i.e. accessed by the background thread exclusively) were moved to the separate `BackgroundFdMonitor` type.


mqudsi:
Ugh, tsan is failing :(

fish-rust/src/fd_monitor.rs
```diff
@@ -0,0 +1,573 @@
...
+use crate::ffi::void_ptr;
+use crate::flog::FLOG;
+use crate::wutil::perror;
+use cxx::UniquePtr;
+
+#[cxx::bridge]
+mod fd_monitor {
+    /// Reason for waking an item
+    #[repr(u8)]
+    #[cxx_name = "item_wake_reason_t"]
```
krobelus, line 17: interesting, this seems to do mostly the same things as a type alias on the C++ side.
Sadly it doesn't work for enum variants.

fish-rust/src/fds.rs
```diff
@@ -1,13 +1,30 @@
...
+pub struct AutoCloseFd2 {
     fd_: RawFd,
 }
 
-impl autoclose_fd_t {
+#[cxx::bridge]
+mod autoclose_fd_t {
+    extern "Rust" {
+        #[cxx_name = "autoclose_fd_t2"]
+        type AutoCloseFd2;
```
krobelus, line 18: I'd much rather call this `AutoCloseFd` (or `AutocloseFd`?) and use this name also on the C++ side.
Then it's obvious where it's from.

fish-rust/src/fd_monitor.rs
```diff
@@ -0,0 +1,573 @@
...
+    enum ItemWakeReason {
+        /// The fd became readable (or was HUP'd)
+        Readable,
```
krobelus, line 20: I've been keeping enum variant names but I guess we can change them too

fish-rust/src/fd_monitor.rs
```diff
@@ -0,0 +1,573 @@
...
+        #[rust_name = "new_fd_event_signaller"]
+        fn ffi_new_fd_event_signaller_t() -> UniquePtr<FdEventSignaller>;
+    }
+    // unsafe extern "C++" {
+    //     include!("fds.h");
+    //
+    //     /// A helper class for managing and automatically closing a file descriptor.
+    //     #[rust_name = "AutoCloseFd"]
+    //     type autoclose_fd_t; // = crate::ffi::autoclose_fd_t;
+    //     /// Closes the fd if not already closed
```
krobelus, line 64: should probably not add commented code

fish-rust/src/fd_monitor.rs
```diff
@@ -0,0 +1,573 @@
...
+    extern "Rust" {
+        #[cxx_name = "fd_monitor_item_id_t"]
+        type FdMonitorItemId;
+    }
+
+    extern "Rust" {
+        #[cxx_name = "fd_monitor_item_t"]
+        type FdMonitorItem;
+
+        #[cxx_name = "make_fd_monitor_item_t"]
```
krobelus, line 85: The `_t` appendix is only for type names IMO
Also we've used "new" instead of "make".

```suggestion
        #[cxx_name = "new_fd_monitor_item"]
```

fish-rust/src/fd_monitor.rs
```diff
@@ -0,0 +1,573 @@
...
+        fn new_fd_monitor_item_ffi(fd: i32, timeout_usecs: u64, callback: *const c_void, param: *const c_void) -> Box<FdMonitorItem>;
+    }
+
+    extern "Rust" {
+        #[cxx_name = "fd_monitor_t"]
+        type FdMonitor;
+
+        #[cxx_name = "make_fd_monitor_t"]
+        fn new_fd_monitor_ffi() -> Box<FdMonitor>;
```
krobelus, line 94: The reason I didn't add the `_ffi` suffix to such functions is that they
are already private, so the chance of accidental use is lower,
and once they're unused we get a compiler warning.
But it's probably good to have the `_ffi` suffix for clarity.

fish-rust/src/fd_monitor.rs
```diff
@@ -0,0 +1,573 @@
...
+// unsafe impl Sync for AutoCloseFd {}
+
+// TODO: Remove once we're no longer using the FFI variant of AutoCloseFd.
+impl Unpin for FdMonitorItem {}
+
+/// Each item added to fd_monitor_t is assigned a unique ID, which is not recycled. Items may have
+/// their callback triggered immediately by passing the ID. Zero is a sentinel.
+#[repr(transparent)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub struct FdMonitorItemId(u64);
```
krobelus, line 120: probably no need for `repr(transparent)`. C++ won't write this directly, right?

fish-rust/src/fd_monitor.rs
```diff
@@ -0,0 +1,573 @@
...
+/// `FdMonitorItem` objects in the same vector (in [`FdMonitor`]), which isn't possible since
+/// each closure/callback is given its own type.
+///
+/// Additionally, as capturing C++ closures can't be safely used via ffi interop, we have a separate
+/// variant of the type that uses the C abi to invoke a callback. This will be removed when the
+/// dependent C++ code (currently only `src/io.cpp`) is ported to rust
+enum FdMonitorCallback {
+    None,
+    Native(Box<dyn Fn(&mut AutoCloseFd2, ItemWakeReason) + Send + Sync>),
+    Ffi(FfiCallback /* fn ptr */, void_ptr /* param */),
```
krobelus, line 139: The enum seems like a nice solution for isolating the FFI ugliness.

The callback context argument is also a good choice - simple and effective.

fish-rust/src/fd_monitor.rs
```diff
@@ -0,0 +1,573 @@
...
+
+    extern "Rust" {
+        #[cxx_name = "fd_monitor_t"]
+        type FdMonitor;
+
+        #[cxx_name = "make_fd_monitor_t"]
+        fn new_fd_monitor_ffi() -> Box<FdMonitor>;
+
+        #[cxx_name = "add_item"]
+        fn add_item_ffi(&mut self, fd: i32, timeout_usecs: u64, callback: *const c_void, param: *const c_void) -> u64;
```
krobelus, line 97: Might have been a bit less confusing for me if we kept the `add()` function and add a separate `new_fd_monitor_item()` function.

fish-rust/src/fd_monitor.rs
```diff
@@ -0,0 +1,573 @@
...
+        Self {
+            callback: FdMonitorCallback::None,
+            fd: AutoCloseFd2::empty(),
+            timeout: None,
+            last_time: None,
+            item_id: FdMonitorItemId(0),
+        }
+    }
+
+    // fn set_callback_ffi(&mut self, callback: FfiCallback, param: void_ptr) {
```
krobelus, line 249: leftover comment

fish-rust/src/fd_monitor.rs
```diff
@@ -0,0 +1,573 @@
...
+            self.change_signaller.post();
+        }
+    }
+
+    fn poke_item_ffi(&self, item_id: u64) {
+        self.poke_item(FdMonitorItemId(item_id))
+    }
+
+    fn new_ffi() -> Box<Self> {
+        Box::new(Self {
```
krobelus, line 413: Most of this function wants to be reused by Rust once we port `io.cpp`.
Why not call this `new()` and move the `Box::new` wrapping to `new_fd_monitor_ffi()` (since the Box only exists for the CXX bridge)

fish-rust/src/fd_monitor.rs
```diff
@@ -0,0 +1,573 @@
...
+            }
+            if !self.fd.is_valid() {
+                result = ItemAction::Remove;
+            }
+        }
+        return result;
+    }
+
+    /// Invoke this item's callback with a poke, if its id is present in the sorted poke list.
+    fn maybe_poke_item(&mut self, pokelist: &[FdMonitorItemId]) -> ItemAction {
```
krobelus, line 216: ideally the rename from `poke_item()` to `maybe_poke_item()` should be in a separate commit unless it's related to the port?

fish-rust/src/fd_monitor.rs
```diff
@@ -0,0 +1,573 @@
...
+            }
+        }
+        // Return `ItemAction::Remove` if the callback closed the fd
+        match self.fd.is_valid() {
+            true => ItemAction::Retain,
+            false => ItemAction::Remove,
+        }
+    }
+
+    fn new() -> Self {
```
krobelus, line 239: If we want to match the C++ code this should be public.

fish-rust/src/fd_monitor.rs
```diff
@@ -0,0 +1,573 @@
...
+                    return ItemAction::Remove;
+                }
+                return ItemAction::Retain;
+            };
+
+            // Service all items that are either readable or have timed out, and remove any which
+            // say to do so.
+
+            // This line is from the C++ codebase (fd_monitor.cpp:170) but this write is never read.
+            // now = Instant::now();
```
krobelus, line 500: yeah it can be removed, maybe in a separate commit

fish-rust/src/threads.rs
```diff
@@ -0,0 +1,72 @@
...
+
+/// The rusty version of `iothreads::make_detached_pthread()`. We will probably need a
+/// `spawn_scoped` version of the same to handle some more advanced borrow cases safely, and maybe
+/// an unsafe version that doesn't do any lifetime checking akin to
+/// `spawn_unchecked()`[std::thread::Builder::spawn_unchecked], which is a nightly-only feature.
+///
+/// Returns a boolean indicating whether or not the thread was successfully launched. Failure here
+/// is not dependent on the passed callback and implies a system error (likely insufficient
+/// resources).
+pub fn spawn<F: FnOnce() + Send + 'static>(callback: F) -> bool {
```
krobelus, line 14: Hmm why not call this file `iothread.rs`, is it because we use qualified names? Fine I guess.

src/io.cpp
```diff
@@ -102,38 +116,51 @@ void io_buffer_t::begin_filling(autoclose_fd_t fd) {
...
+    item_id_ =
+        fd_monitor().add_item(fd.acquire(), kNoTimeout, (::c_void *)item_callback_trampoline, (::c_void *)args);
 }
 
+/// This is a hack to work around the difficulties in passing a capturing lambda across FFI
+/// boundaries. A static function that takes a generic/untyped callback parameter is easy to
+/// marshall with the basic C ABI.
+void io_buffer_t::item_callback(autoclose_fd_t2 &fd, uint8_t r, callback_args_t *args) {
+    item_wake_reason_t reason = (item_wake_reason_t)r;
+    auto &promise = *args->promise;
```
krobelus, line 132: we might want to port ASSERT_IS_BACKGROUND_THREAD and friends but that's probably not important

fish-rust/src/fd_monitor.rs
mqudsi: I keep _ffi to know what we want out of the codebase with a simple grep.

src/io.cpp
mqudsi: There’s no need since now it’s in a separate type that’s instantiated on the background thread. It used to be a member function of the same object shared on both foreground and background.

fish-rust/src/fd_monitor.rs
mqudsi: I was going to transmute or pointer cast originally.

fish-rust/src/fd_monitor.rs
mqudsi: Hmm Peter had a few make_ that I saw, because I was originally using new_. I think he modeled them after make_unique, mke_shared, etc

fish-rust/src/fd_monitor.rs
mqudsi: Yeah I kept flipping between using the rust and cpp versions of the auto close type. I’ll remove it.

fish-rust/src/fd_monitor.rs
mqudsi: I was going to typedef/using/define it but it was just a single call site so it was just easier to change it cpp side.

fish-rust/src/fds.rs
mqudsi: Not a bad idea.  Then we don’t need a second rename.

fish-rust/src/fd_monitor.rs
mqudsi: Yeah. I tried defining them as consts i ln the shared cpp bridge section but constant aren’t supported.

fish-rust/src/fd_monitor.rs
mqudsi: Fair enough.

fish-rust/src/fd_monitor.rs
mqudsi: I think this is the kind of thing that doesn’t have to match. that Said it’ll have to be public when we no longer use it via ffi.

fish-rust/src/threads.rs
mqudsi: Well it’s not specific to io threads. We never want to launch a new thread that is using our current signal mask. I don’t much like the generic thread name but I didn’t want to waste too much time bike-shedding. This damn PR took way longer and way more work than I was expecting!

mqudsi:
I've been studying the tsan error and as far as I can tell, it seems to be caused by tsan not understanding the atomicity of `Arc<cxx::UniquePtr>` so it thinks that there is possible data race between the background fd monitor thread reading an fd and the teardown of both the foreground and background fd monitor objects sharing an `Arc<UniquePtr<fd_event_signaller_t>>`.

The ordering of events as reported in tsan is correct (read comes before overwrite of fd with `-1`). tsan just doesn't get that Arc is going to make sure the fd_event_signaller_t destructor isn't called until all reads have completed (because they're both going through clones of the same `Arc`). 

I'm going to try to change the previously ported (by @ridiculousfish, I think) `fd_event_signaller_t` cpp ffi to return a `shared_ptr` instead of `unique_ptr` and see if we can share a single `SharedPtr<FdEventSignaller>` between the two objects instead of sharing an `Arc<UniquePtr<FdEventSignaller>>` between them.

fish-rust/src/fd_monitor.rs
krobelus: true there's no obvious winner here. Maybe `new_` is for things that return `Box`

mqudsi:
It's still the same problem since whether you use `SharedPtr` or `UniquePtr` you're still using a rust type and our tsan setup doesn't recognize the order enforced by a drop call. 

Rust has its own way of running thread sanitizer (`RUSTFLAGS="-Z sanitizer=thread" cargo test`) but I don't know what to do in the meanwhile. Do we want to disable our existing TSAN altogether for the duration of the port? That seems excessive, but if it doesn't understand something as simple as `Drop`, it's going to have a harder and harder time to deal with the code we port.

krobelus:
I think it's fine to disable tsan for now

fish-rust/src/fd_monitor.rs
mqudsi: I think I mentioned somewhere the reason I didn't do that is because each fd_monitor_item would need to be dynamically allocated. We generate dozens of them per pipeline so it seemed stupid to force the ffi caller to allocate a Box only for the rust side of the ffi code to unwrap the box on the next line (when the call to add happens).

ridiculousfish (changes requested):
I spotted one serious bug (`now` needs to be updated after the `select`) and also the class should be shareable across threads (`&FdMonitor` should be `Sync`). Overall really impressive work!

fish-rust/src/fd_monitor.rs
```diff
@@ -0,0 +1,573 @@
...
+pub struct FdMonitorItemId(u64);
+
+type FfiCallback = extern "C" fn(*mut AutoCloseFd2, u8, void_ptr);
+
+/// The callback type used by [`FdMonitorItem`]. It is passed a mutable reference to the
+/// `FdMonitorItem`'s [`FdMonitorItem::fd`] and [the reason](ItemWakeupReason) for the wakeup. The
+/// callback may close the fd, in which case the `FdMonitorItem` is removed from [`FdMonitor`]'s
+/// set.
+///
+/// We can't use generics here and need to resort to dynamic dispatch in order to store multiple
```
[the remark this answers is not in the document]
mqudsi: Yes, it's mainly to document the differences between the cpp and rust code. Happy to omit.

fish-rust/src/fd_monitor.rs
```diff
@@ -0,0 +1,573 @@
...
+    pub fn add(&mut self, mut item: FdMonitorItem) -> FdMonitorItemId {
+        // This assert is the equivalent of autoclose_fd_t::is_valid()
+        assert!(item.fd.is_valid());
+        assert!(item.timeout != Some(Duration::ZERO), "Invalid timeout!");
+        assert!(
+            item.item_id == FdMonitorItemId(0),
+            "Item should not already have an id!"
+        );
+
+        self.last_id += 1;
```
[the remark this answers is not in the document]
mqudsi: I assumed fd monitor wasn't shared, so it was ok without one. But if it's true that it needs to be `Sync`, then that's definitely necessary.

fish-rust/src/fd_monitor.rs
```diff
@@ -0,0 +1,573 @@
...
+    item.callback = FdMonitorCallback::Ffi(callback, param);
+    if timeout_usecs != FdReadableSet::kNoTimeout {
+        item.timeout = Some(Duration::from_micros(timeout_usecs));
+    }
+    return Box::new(item);
+}
+
+/// A class which can monitor a set of fds, invoking a callback when any becomes readable (or has
+/// been HUP'd) or when per-item-configurable timeouts are reached.
+pub struct FdMonitor {
```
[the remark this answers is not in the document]
mqudsi: Ah I missed that `fd_monitor_t` is being used across threads in the cpp code. That definitely changes how the code should be ported and we'd need to use interior mutability to be able to use `&self` instead of `&mut self` but still make changes. Sorry, I should have caught that!

As for asserting that it is `Sync`, you can actually just hard code such a check as some const code without a test, e.g. https://github.com/neosmart/securestore-rs/blob/54488de686ac5e17833379ad3cb7f90540f4fa74/securestore/src/lib.rs#L200-L210

fish-rust/src/fd_monitor.rs
```diff
@@ -0,0 +1,573 @@
...
+        #[cxx_name = "fd_monitor_item_id_t"]
+        type FdMonitorItemId;
+    }
+
+    extern "Rust" {
+        #[cxx_name = "fd_monitor_item_t"]
+        type FdMonitorItem;
+
+        #[cxx_name = "make_fd_monitor_item_t"]
+        fn new_fd_monitor_item_ffi(fd: i32, timeout_usecs: u64, callback: *const c_void, param: *const c_void) -> Box<FdMonitorItem>;
```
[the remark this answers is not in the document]
mqudsi: I ran it before changing the signature, I actually just pushed an update w/ everything prettified and cleaned up before I saw your review. Will address your issues and update.

fish-rust/src/fd_monitor.rs
```diff
@@ -0,0 +1,573 @@
...
+            }
+        }
+    }
+
+    /// Poke items in the poke list, removing any items that close their fd in their callback. The
+    /// poke list is consumed after this. This is only called from the background thread.
+    fn poke(&mut self, pokelist: &[FdMonitorItemId]) {
+        self.items.retain_mut(|item| {
+            let fd = item.fd.as_raw_fd();
+            let action = item.maybe_poke_item(&*pokelist);
```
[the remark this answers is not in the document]
mqudsi: It's not. I got into the habit because that syntax is sometimes needed (it's called an "explicit reborrow") such as if you need to pass a mut reference to `self` to another function in a function that's already using using `&mut self`. 

Personal preference as well to always be clear about borrows vs moves which isn't always obvious just by looking at the call site. Clippy doesn't like the unneeded ref + deref either by default, I guess we'll have to decide on what we want to unify around so it's all the same.

fish-rust/src/fd_monitor.rs
```diff
@@ -0,0 +1,573 @@
...
+            change_signaller: Arc::new(new_fd_event_signaller()),
+            last_id: 0,
+        })
+    }
+}
+
+impl BackgroundFdMonitor {
+    /// Starts the background thread.
+    fn run(&mut self) {
+        let mut pokelist: Vec<FdMonitorItemId> = Vec::new();
```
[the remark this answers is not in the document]
mqudsi: Sure, but I want to clarify that it's actually not possible for this to be called in the current code except on the background thread since the `BackgroundFdMonitor` is moved to the new thread and the method requires `&mut self` (actually, in the commit I pushed before seeing your review, it's been changed to `run(self)` to be even clearer). These functions used to be part of the same `fd_monitor_t` instance that was shared between background and foreground threads, but that's no longer the case in the rust code to please the borrow checker - now they're two different objects, each with its own function. The regular foreground code can't call the background function even if it wanted to.

(Of course we can always just add the assert anyway, I have no problem with defense-in-depth.)

fish-rust/src/fd_monitor.rs
```diff
@@ -0,0 +1,573 @@
...
+                    FLOG!(fd_monitor, "Removing fd", fd);
+                    return ItemAction::Remove;
+                }
+                return ItemAction::Retain;
+            };
+
+            // Service all items that are either readable or have timed out, and remove any which
+            // say to do so.
+
+            // This line is from the C++ codebase (fd_monitor.cpp:170) but this write is never read.
```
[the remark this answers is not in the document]
mqudsi: I added a comment, but I also moved it above the closure definition (which was the preceding expression) which makes it a lot clearer.

mqudsi:
@ridiculousfish I've pushed some updated commits that should address the issues you raised.

PR #9587, opened 2023-02-18
Fix issue#8877 Check if redirections symbols occur textually before the statement's command.

Himmel-OvO:
Fix command highlighting issue with special characters as '>echo'.

## Description
This PR fixes an issue with command highlighting, where if redirections command like '<', '>'appeared before the command, the command highlighting would not work correctly. Specifically, some commands that should have been highlighted as errors were not showing the expected color.

The root cause of this issue was that the previous logic did not account for these special characters appearing before the command. This PR addresses the issue by adding additional checks for these special characters, ensuring that the highlighting logic is properly applied in all cases.

Overall, this fix improves the accuracy and reliability of the command highlighting feature, and will provide a better user experience for users of the system.


src/highlight.cpp
```diff
@@ -775,6 +775,20 @@ static void color_string_internal(const wcstring &buffstr, highlight_spec_t base
...
 }
 
+// For Issue#8877
+// Valid commands following redirection symbols at command line start are colored
+// as $fish_color_command instead of $fish_color_error 
+static bool symbol_text_before_command(const wcstring &buffer, size_t before_cmd_len) {
+    wcstring before_str = buffer.substr(0, before_cmd_len);
+
+    if (before_str.find(L'>') != wcstring::npos
+     || before_str.find(L'<') != wcstring::npos) {
```
mqudsi, line 785: You're potentially scanning the string twice here. Better to use `std::any_of` or a simple for loop.

Himmel-OvO: > You're potentially scanning the string twice here. Better to use `std::any_of` or a simple for loop.

Thank you, mqudsi. I will commit next.

ridiculousfish:
So this is interesting: when encountering

    >echo

for syntax highlighting purposes, fish (correctly) creates an error for the `>`, then restarts parsing for the remainder of the string. Now it sees `echo` and thinks that's a fine command.

The fix as proposed is too aggressive. For example in `true > foo; echo` the `echo` will be colored as an error even though the command is fine.

I'm honestly not sure how best to fix this. One possibility is to fix it in the ast parsing to make the error include the next token, but that is nontrivial. A second possibility is to fix it in highlighting: when we encounter an error, we extend it to following text in some circumstances. Unfortunately neither is easy or clean; I'll have to think about it.

ridiculousfish (changes requested):
Unfortunately we can't merge this because it breaks valid cases like `true > foo; echo`; I don't have any suggestions for how to fix that yet though :(

Himmel-OvO:
> So this is interesting: when encountering
> 
> ```
> >echo
> ```
> 
> for syntax highlighting purposes, fish (correctly) creates an error for the `>`, then restarts parsing for the remainder of the string. Now it sees `echo` and thinks that's a fine command.
> 
> The fix as proposed is too aggressive. For example in `true > foo; echo` the `echo` will be colored as an error even though the command is fine.
> 
> I'm honestly not sure how best to fix this. One possibility is to fix it in the ast parsing to make the error include the next token, but that is nontrivial. A second possibility is to fix it in highlighting: when we encounter an error, we extend it to following text in some circumstances. Unfortunately neither is easy or clean; I'll have to think about it.

You're right. Can I ask on what basis Fish considered interpreting the command '>echo' as an error? In Bash, this command is used to clear the contents of the 'echo' file. 
Additionally, I was thinking about highlighting each command when parsing the AST. Would this approach be practical?

faho:
>Can I ask on what basis Fish considered interpreting the command '>echo' as an error? In Bash, this command is used to clear the contents of the 'echo' file.

Fish does not support redirections before or without a command. In bash, `>echo` is the same as `: >echo`, i.e. it automatically inserts the "no-op" command.

In fish, any redirection before a command is an error. Of course you need to only check the actual current "process", as `commandline` calls it.

This is not an error:

```fish
foo; bar 2>/dev/null | baz
```

This is:

```fish
foo; <file bar 2>/dev/null | baz
```

faho:
Alright, I'm closing this for now since there are issues with the approach.

PR #9588, opened 2023-02-18
Rust: add misc bindings, ports

Xiretza:
These have been extracted from #9563, which is slated to be parked for at least a little while longer.

### DIFF
--- a/fish-rust/src/common.rs
+++ b/fish-rust/src/common.rs
@@ -1,0 +1,33 @@
+use std::mem;
+
+/// A scoped manager to save the current value of some variable, and optionally set it to a new
+/// value. When dropped, it restores the variable to its old value.
+///
+/// This can be handy when there are multiple code paths to exit a block.
+pub struct ScopedPush<'a, T> {
+    var: &'a mut T,
+    saved_value: Option<T>,
+}
+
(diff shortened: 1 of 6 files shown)


fish-rust/src/common.rs
```diff
@@ -0,0 +1,94 @@
...
+    Url,
+    Var,
+    Regex,
+}
+
+/// Flags for the [`escape_string()`] function. These are only applicable when the escape style is
+/// [`EscapeStringStyle::Script`].
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub struct EscapeFlags {
+    /// Do not escape special fish syntax characters like the semicolon. Only escape non-printable
```
krobelus, line 50: This means that

	escape_string(value, ESCAPE_NO_PRINTABLES | ESCAPE_NO_QUOTED)

becomes

	escape_string(value, EscapeFlags{
		no_printables: true,
		no_quoted: true,
		no_tilde: false,
		symbolic: false,
	})

which is alright I guess. Maybe we want a shortcut but it's no big deal

Xiretza: It might not make much of a difference, but you should be able to use struct update syntax: `EscapeFlags { no_printables: true, ..EscapeFlags::default() }`

PR #9591, opened 2023-02-19
Move POD components of library_data_t to separate struct

Xiretza:
This allows them to be accessed as regular fields from Rust, rather than having to create setter/getter methods for each of them.

### DIFF
--- a/fish-rust/src/ffi.rs
+++ b/fish-rust/src/ffi.rs
@@ -49,6 +49,7 @@ include_cpp! {
     generate!("job_t")
     generate!("process_t")
     generate!("library_data_t")
+    generate_pod!("library_data_pod")
 
     generate!("proc_wait_any")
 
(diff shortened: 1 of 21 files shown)


fish-rust/src/ffi.rs
```diff
@@ -80,6 +81,12 @@ impl parser_t {
         let ffi_jobs = self.ffi_jobs();
         unsafe { slice::from_raw_parts(ffi_jobs.jobs, ffi_jobs.count) }
     }
+
+    pub fn get_libdata_pod(&mut self) -> &mut library_data_pod {
```
krobelus, line 85: nit: I could do without the `get_` prefix (there are no field/method conflicts in rust) but that's fine.
Looks good.

src/parser.h
```diff
@@ -207,6 +207,12 @@ struct library_data_t {
 
     /// The read limit to apply to captured subshell output, or 0 for none.
     size_t read_limit{0};
+};
+
+/// Miscellaneous data used to avoid recursion and others.
+struct library_data_t {
+    /// Plain-Old-Data components that can be shared over FFI
+    struct library_data_pod pod;
```
krobelus, line 215: If we used inheritance (`library_data_t : public library_data_pod_t`), we could probably avoid the C++ changes without causing problems for Rust, right?

either way `library_data_pod` should have the `_t` suffix.

PR #9592, opened 2023-02-19
port abbrs.cpp and abbr builtin to rust

nrjais:
## Description

Talk about your changes here.

Fixes issue #

## TODOs:


nrjais:
The previous PR was forced closed due to master merge

nrjais:
This is ready for review, I will go through the changes once again later to see if I missed something, but it can be reviewed to see if anything was missed

fish-rust/src/abbrs.rs
```diff
@@ -0,0 +1,479 @@
...
+        is_function: bool,
+        set_cursor_marker: UniquePtr<CxxWString>,
+        has_cursor_marker: bool,
+    }
+
+    struct abbrs_replacement_t {
+        range: SourceRange,
+        text: UniquePtr<CxxWString>,
+        cursor: usize,
+        has_cursor: bool,
```
nrjais, line 46: can't put `usize` in UniquePtr as it is not supported, This can be removed once FFI code is removed.
Also address sanitizer was failing due to having null pointers, so I did not use them at other places also

fish-rust/src/abbrs.rs
```diff
@@ -0,0 +1,479 @@
...
+
+pub fn with_abbrs<R>(cb: impl FnOnce(&AbbreviationSet) -> R) -> R {
+    let abbrs_g = abbrs.lock().unwrap();
+    cb(&abbrs_g)
+}
+
+pub fn with_abbrs_mut<R>(cb: impl FnOnce(&mut AbbreviationSet) -> R) -> R {
+    let mut abbrs_g = abbrs.lock().unwrap();
+    cb(&mut abbrs_g)
+}
```
nrjais, line 101: Let me know if I should keep this or Use `abbrs_get_set` everywhere in rust also

fish-rust/src/abbrs.rs
```diff
@@ -0,0 +1,479 @@
...
+    set_cursor_marker: Option<WString>,
+}
+
+impl From<Replacer> for abbrs_replacer_t {
+    fn from(value: Replacer) -> Self {
+        let has_cursor_marker = value.set_cursor_marker.is_some();
+        Self {
+            replacement: value.replacement.to_ffi(),
+            is_function: value.is_function,
+            set_cursor_marker: value.set_cursor_marker.unwrap_or_default().to_ffi(),
```
nrjais, line 213: The default value is not used if the cursor is not present, using null failed in `asan` checks

fish-rust/src/abbrs.rs
```diff
@@ -0,0 +1,479 @@
...
+        assert!(abbrs_g.has_name(L!("gc")));
+
+        abbrs_g.rename(L!("gc"), L!("gcc"));
+        assert!(abbrs_g.has_name(L!("gcc")));
+        assert!(!abbrs_g.has_name(L!("gc")));
+
+        assert!(!abbrs_g.erase(L!("gc")));
+        assert!(abbrs_g.erase(L!("gcc")));
+        assert!(!abbrs_g.erase(L!("gcc")));
+    })
```
nrjais, line 478: Moved these tests from CPP to here, because I needed to create ffi funtions to test these from CPP

fish-rust/src/common.rs
```diff
@@ -89,5 +92,67 @@ pub fn escape_string(s: &wstr, style: EscapeStringStyle) -> WString {
...
+    let mut result = ffi::try_compile(&regex_make_anchored(L!("(a+)|(b+)")), &flags);
+    assert!(!result.has_error());
+
+    let re = result.as_mut().get_regex();
+    assert!(!re.is_null());
+    assert!(!re.matches_ffi(&L!("").to_ffi()));
+    assert!(!re.matches_ffi(&L!("aabb").to_ffi()));
+    assert!(re.matches_ffi(&L!("aaaa").to_ffi()));
+    assert!(re.matches_ffi(&L!("bbbb").to_ffi()));
+    assert!(!re.matches_ffi(&L!("aaaax").to_ffi()));
```
nrjais, line 157: Moved from CPP regex as this function was only used in abbr

fish-rust/src/ffi.rs
```diff
@@ -115,6 +138,12 @@ impl From<wcharz_t> for wchar::WString {
     }
 }
 
+impl Debug for re::regex_t {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        f.write_str("regex_t")
+    }
+}
```
nrjais, line 145: Should I keep this? Need it for impl Debug on Abbr

fish-rust/src/abbrs.rs
```diff
@@ -0,0 +1,479 @@
...
+    sync::{Arc, Mutex, MutexGuard},
+};
+
+use crate::{
+    wchar::L,
+    wchar_ffi::{WCharFromFFI, WCharToFFI},
+};
+use cxx::{CxxWString, UniquePtr};
+use once_cell::sync::Lazy;
+pub use widestring::{Utf32Str as wstr, Utf32String as WString};
```
krobelus, line 13: No need to do the import dance here again.
Make it `use crate::wchar::{wstr, WString}`.

fish-rust/src/abbrs.rs
```diff
@@ -0,0 +1,479 @@
...
+    /// Construct a replacement from a replacer.
+    /// The \p range is the range of the text matched by the replacer in the command line.
+    /// The text is passed in separately as it may be the output of the replacer's function.
+    fn from(range: SourceRange, mut text: WString, set_cursor_marker: Option<WString>) -> Self {
+        let mut cursor = None;
+        if let Some(set_cursor_marker) = set_cursor_marker {
+            let matched = text
+                .as_char_slice()
+                .windows(set_cursor_marker.len())
+                .enumerate()
```
krobelus, line 241: you can use `.position()` instead of `enumerate()`

fish-rust/src/abbrs.rs
```diff
@@ -0,0 +1,479 @@
...
+                .find(|(_, w)| w.starts_with(set_cursor_marker.as_char_slice()))
```
krobelus, line 242: no need for `starts_with`, we can use `w == set_cursor_marker.as_char_slice()`

fish-rust/src/abbrs.rs
```diff
@@ -0,0 +1,479 @@
...
+    /// Add an abbreviation. Any abbreviation with the same name is replaced.
+    pub fn add(&mut self, abbr: Abbreviation) {
+        assert!(!abbr.name.is_empty(), "Invalid name");
+        let inserted = self.used_names.insert(abbr.name.clone());
+        if !inserted {
+            // Name was already used, do a linear scan to find it.
+            let (index, _) = self
+                .abbrs
+                .iter()
+                .enumerate()
```
krobelus, line 301: same, I think we can use `position()` here too

fish-rust/src/abbrs.rs
```diff
@@ -0,0 +1,479 @@
...
+
+fn abbrs_replacement_from_ffi(
+    range: SourceRange,
+    text: &CxxWString,
+    set_cursor_marker: &CxxWString,
+    has_cursor_marker: bool,
+) -> abbrs_replacement_t {
+    let mut cursor_marker = None;
+    if has_cursor_marker {
+        let _ = cursor_marker.insert(set_cursor_marker.from_ffi());
```
krobelus, line 396: nit: Id write this as `let cursor_marker = if has_cursor_marker { Some(...) } else { None }`

fish-rust/src/abbrs.rs
```diff
@@ -0,0 +1,479 @@
...
+    }
+
+    let replacement = Replacement::from(
+        SourceRange {
+            start: range.start,
+            length: range.length,
+        },
```
krobelus, line 403: here you can directly pass on `range`  I think. We might need to add `#[derive(Clone, Copy)]` to `SourceRange`

fish-rust/src/abbrs.rs
```diff
@@ -0,0 +1,479 @@
...
+        self.g.erase(&name.from_ffi());
+    }
+}
+use crate::ffi_tests::add_test;
+add_test!("rename_abbrs", || {
+    use crate::{
+        abbrs::{Abbreviation, Position},
+        wchar::L,
+    };
+    pub use widestring::{Utf32Str as wstr, Utf32String as WString};
```
krobelus, line 448: same here, prefer importing from `crate::wchar`

fish-rust/src/builtins/abbr.rs
```diff
@@ -0,0 +1,595 @@
...
+
+        return true;
+    }
+}
+
+fn join(list: &[&wstr], sep: &wstr) -> WString {
+    let mut result = WString::new();
+    for s in list {
+        result.push_utfstr(s);
+        result.push_utfstr(sep);
```
krobelus, line 122: This looks wrong, we don't want to add the separator after the last element.
Can we not use [slice::join](https://doc.rust-lang.org/std/primitive.slice.html#method.join)?

fish-rust/src/builtins/abbr.rs
```diff
@@ -0,0 +1,595 @@
...
+        replacement
+    };
+
+    let position = opts.position.unwrap_or(Position::Command);
+
+    // Note historically we have allowed overwriting existing abbreviations.
+    abbrs::with_abbrs_mut(move |abbrs| {
+        abbrs.add(Abbreviation {
+            name: name.clone(),
+            key: key.clone(),
```
krobelus, line 380: nit: we should take Options by value/move, then we could get rid of these clones.
But there's no harm in following the C++ code

fish-rust/src/common.rs
```diff
@@ -89,5 +92,67 @@ pub fn escape_string(s: &wstr, style: EscapeStringStyle) -> WString {
...
+    if name.find_char('\0').is_some() {
+        return false;
+    };
+    true
+}
+
+/// Adjust a pattern so that it is anchored at both beginning and end.
+/// This is a workaround for the fact that PCRE2_ENDANCHORED is unavailable on pre-2017 PCRE2
+/// (e.g. 10.21, on Xenial).
+pub fn regex_make_anchored(pattern: &wstr) -> WString {
```
krobelus, line 119: Probably better to put this in `re.rs`?

fish-rust/src/common.rs
```diff
@@ -89,5 +92,67 @@
...
+    let mut anchored = pattern.to_owned();
+    // PATTERN -> ^(:?PATTERN)$.
+    let prefix = L!("^(?:");
+    let suffix = L!(")$");
+    anchored.reserve(pattern.len() + prefix.len() + prefix.len());
```
krobelus, line 124:
```suggestion
+    anchored.reserve(pattern.len() + prefix.len() + suffix.len());
```

fish-rust/src/env.rs
```diff
@@ -0,0 +1,17 @@
+// Flags that may be passed as the 'mode' in env_stack_t::set() / environment_t::get().
+
+pub mod flags {
+    use autocxx::c_int;
+
+    /// Flag for universal variable.
+    pub const ENV_UNIVERSAL: c_int = c_int(1 << 3);
```
krobelus, line 7: could also port all of these right away

src/parse_constants.h
```diff
@@ -31,12 +31,13 @@ using parse_error_list_t = ParseErrorList;
...
 
-struct source_range_t {
+struct SourceRange {
     source_offset_t start;
     source_offset_t length;
 };
+using source_range_t = SourceRange;
 
-enum class parse_token_type_t : uint8_t {
+enum class parse_token_type_t:uint8_t {
```
krobelus, line 40: nit: this line should be restored.

krobelus:
Really amazing work! I have only minor comments.

fish-rust/src/abbrs.rs
krobelus: > can't put `usize` in UniquePtr

I think`SharedPtr<usize>` works. Your solution is fine too

fish-rust/src/abbrs.rs
krobelus: with_abbrs seems fine, I don't see any particular reason for or against it

fish-rust/src/ffi.rs
krobelus: seems fine to keep

fish-rust/build.rs
```diff
@@ -20,6 +20,7 @@ fn main() -> miette::Result<()> {
     // This must come before autocxx so that cxx can emit its cxx.h header.
     let source_files = vec![
         "src/fd_monitor.rs",
+        "src/abbrs.rs",
```
krobelus, line 23: sort

fish-rust/src/common.rs
```diff
@@ -89,5 +92,67 @@ pub fn escape_string(s: &wstr, style: EscapeStringStyle) -> WString {
         EscapeStringStyle::Regex => ffi::escape_string_style_t::STRING_STYLE_REGEX,
     };
 
-    ffi::escape_string(s.as_ptr(), flags_int.into(), style).from_ffi()
+    ffi::escape_string(c_str!(s), flags_int.into(), style).from_ffi()
```
krobelus, line 95: I think this is the fix referred to by the commit message.
The rest of the commit is probably largely unrelated, please split to make it easier to review

fish-rust/src/wutil/format/format.rs
```diff
@@ -514,3 +514,19 @@ impl Printf for &WString {
...
+    let width: usize = match spec.width {
+        NumericParam::Literal(w) => w,
+        _ => {
+            return Err(PrintfError::Unknown); // should not happen at this point!!
+        }
+    }
+    .try_into()
+    .unwrap_or_default();
+    while s.len() < width {
+        s.insert(0, ' ');
```
krobelus, line 528: should use `str::repeat` to avoid quadratic runtime.

src/reader.cpp
```diff
@@ -1501,8 +1502,12 @@ bool reader_data_t::expand_abbreviation_at_cursor(size_t cursor_backtrack) {
         size_t cursor_pos = el->position() - std::min(el->position(), cursor_backtrack);
         if (auto replacement =
                 reader_expand_abbreviation_at_cursor(el->text(), cursor_pos, this->parser())) {
-            push_edit(el, edit_t{replacement->range, std::move(replacement->text)});
-            update_buff_pos(el, replacement->cursor);
+            push_edit(el, edit_t{replacement->range, *replacement->text});
+            if (replacement->has_cursor) {
```
krobelus, line 1506: This commit fixes a bug introduced by the first commit.
It'd be better to squash it, so the bug is never introduced.
As a reviewer I don't want to know how the sausage is made.
Also, every commit should pass tests.  (This means he formatting fixes should be placed first etc.)
See also https://www.bitsnbites.eu/git-history-work-log-vs-recipe/

fish-rust/src/abbrs.rs
nrjais: Nice, did not know about position

fish-rust/src/builtins/abbr.rs
nrjais: `[&wstr]` Does not implement `Join` trait

fish-rust/src/builtins/abbr.rs
nrjais: Both are already references, as key is same as name when no regex given, So clone will just move above instead of here

fish-rust/src/common.rs
nrjais: Moved

fish-rust/src/env.rs
nrjais: Done

fish-rust/src/wutil/format/format.rs
nrjais: Done

nrjais:
I have rearranged commits, and squased all commits porting abbrs

fish-rust/src/env.rs
```diff
@@ -0,0 +1,37 @@
+/// Flags that may be passed as the 'mode' in env_stack_t::set() / environment_t::get().
+pub mod flags {
+    use autocxx::c_int;
```
krobelus, line 3: I wonder what's the value of this `flags` module.
Constants like ENV_DEFAULT are intended to be used by their unqualified name.
I guess it does make our import statements a bit more transparent, and it's an obvious place for the doc comment.
We might want to apply this pattern to other flags. 

I'll change the // comment it to a doc comment.

src/fish_tests.cpp
```diff
@@ -6828,23 +6812,6 @@ static void test_re_basic() {
     }
     do_test(join_strings(matches, L',') == L"AA,CC,11");
     do_test(join_strings(captures, L',') == L"A,C,1");
```
krobelus, line 6830: this should be in the commit that ports make_anchored. Will fix.
Actually, the make_anchored commit port doesn't compile because C++ is still using it.
Totally not worth adding a temporary FFI wrapper, so let's do a merge commit instead

fish-rust/src/ffi.rs
```diff
@@ -1,7 +1,13 @@
 use crate::wchar;
+use crate::wchar_ffi::WCharToFFI;
+#[rustfmt::skip]
```
krobelus, line 3: need to add rustfmt::skip here

fish-rust/src/abbrs.rs
```diff
@@ -0,0 +1,470 @@
...
+        ));
+    }
+
+    fn erase(&mut self, name: &CxxWString) {
+        self.g.erase(&name.from_ffi());
+    }
+}
+use crate::ffi_tests::add_test;
+add_test!("rename_abbrs", || {
+    use crate::wchar::wstr;
```
krobelus, line 435: unused import here

fish-rust/src/common.rs
```diff
@@ -89,5 +91,23 @@ pub fn escape_string(s: &wstr, style: EscapeStringStyle) -> WString {
...
```
krobelus, line 94: FWIW with `c_str!()` we have to be extra careful because it creates a reference to a temporary.
I believe in this case the temporary lives long enough.

fish-rust/src/env.rs
nrjais: Yeah, there were another set of constants in this file, It seemed better to have separate mod for them. I think if we move from constant to enum or something else later than that will help with namespacing and docs

krobelus: yeah if both have the same ENV prefix that's another justification for the namespace

PR #9593, opened 2023-02-19
Port "random" builtin to rust

faho:
## Description

Repeat of #9585 

This ports the "random" builtin, that is the builtin called `random` and not just any random builtin, to rust.

~~The remaining issue here is some weirdness with how the C++ does the numbers - the tests explicitly exercise that you can declare random numbers from the smallest signed long long to the largest with a step size of the difference between the two, and I haven't been able to make that work.~~ This allows the weird signed 64-bit number + an unsigned 64 bit step logic, by doing all the math in 128-bit space.

Also the error messages are a bit of a mess in the C++ - a step of "-1" is "invalid integer", but a step of "0" is "STEP must be a non-negative integer" for some reason?

This uses a new crate, rand, which is what seemed to be the recommended course if you needed a seedable RNG, which we do (random 12 seeds the RNG with "12").

~~Also I needed some way to get wcstoi to tell me it hasn't read the entire token, so I picked a brute-force approach. I would assume there's a more elegant approach?~~

This represents about ~~98%~~ 99% of all the rust code I've ever written, so let's see how much I missed!

## TODOs:


fish-rust/src/builtins/random.rs
```diff
@@ -0,0 +1,193 @@
...
+        streams.err.append(wgettext_fmt!(
+            "%ls: END must be greater than START\n",
+            cmd,
+        ));
+        return STATUS_INVALID_ARGS;
+    }
+
+    // Possibilities can be abs(i64::MIN) + i64::MAX,
+    // so we do this as i128
+    let possibilities = (end as i128 - start as i128) / (step as i128);
```
Xiretza, line 156:
```suggestion
    let possibilities = end.abs_diff(start) / step;
```

fish-rust/src/builtins/random.rs
```diff
@@ -0,0 +1,193 @@
...
+        streams.err.append(wgettext_fmt!(
+            "%ls: range contains only one possible value\n",
+            cmd,
+        ));
+        return STATUS_INVALID_ARGS;
+    }
+
+    let rand = engine.gen_range(0..=possibilities);
+
+    let result = start as i128 + rand as i128 * step as i128;
```
Xiretza, line 168:
```suggestion
    let result = start.checked_add_unsigned(rand * step).unwrap();
```

fish-rust/src/builtins/random.rs
```diff
@@ -0,0 +1,193 @@
...
+                Err(_) => {
+                    streams.err.append(wgettext_fmt!(
+                        "%ls: range contains only one possible value\n",
+                        cmd,
+                    ));
+                    return STATUS_INVALID_ARGS;
+                },
+            }
+        },
+    }
```
Xiretza, line 192:
```suggestion
    streams.out.append(sprintf!("%d\n"L, result));
    return STATUS_CMD_OK;
```

fish-rust/src/builtins/shared.rs
```diff
@@ -119,6 +119,7 @@ pub fn run_builtin(
         RustBuiltin::Emit => super::emit::emit(parser, streams, args),
         RustBuiltin::Exit => super::exit::exit(parser, streams, args),
         RustBuiltin::Return => super::r#return::r#return(parser, streams, args),
+        RustBuiltin::Random => super::random::random(parser, streams, args),
```
Xiretza, line 122: These should be sorted alphabetically (random before return)

fish-rust/src/builtins/mod.rs
```diff
@@ -3,5 +3,6 @@ pub mod shared;
 pub mod echo;
 pub mod emit;
 pub mod r#return;
+pub mod random;
```
Xiretza, line 6: ditto

Xiretza (commented):
I think I've found an acceptable solution to the integer width problem.

Please also run rustfmt (`cargo fmt`).

fish-rust/src/builtins/random.rs
```diff
@@ -0,0 +1,193 @@
...
+    match arg_count {
+        0 => {
+            // Keep the defaults
+        },
+        1 => {
+            // Seed the engine persistently
+            let num = parse::<i64>(streams, cmd, argv[i]);
+            match num {
+                Err(_) => return STATUS_INVALID_ARGS,
+                Ok(x) => *engine = SmallRng::seed_from_u64(x as u64)
```
Xiretza, line 96: In general `as` casts are iffy and should be avoided for anything other than float<->int conversion, but I guess here it's what we want (converting negative values to large positive values).

Xiretza:
I hope nobody relies on seeded random values to be stable across fish versions.

faho:
>I hope nobody relies on seeded random values to be stable across fish versions.

I do not believe it is possible to translate this *while* keeping the values stable, especially because I don't think the values are stable across C++ implementations?

Truth be told I would be surprised if anyone actually used the seeding functionality.

fish-rust/src/builtins/random.rs
faho: Yeah, this just wants "the bits".

fish-rust/src/builtins/random.rs
faho: Uuhhhh... why are we specifically doing a checked_add only to throw away the error?

Xiretza: There's no unchecked `add_unsigned`, and this can't overflow.

faho: Ah, so the possibilities are positive, so rand and step are positive, so adding it to start mathematically has to fit.

So you're doing checked_add with unwrap here to crash if it doesn't?

Xiretza: The problem is that `rand * step` is a `u64` that's quite possibly larger than `i64::MAX`, so you can't simply convert it to `i64` and add it. There's no implementation for `i64 + u64` at all, so we have to roll our own; alternatively we could also use `overflowing_add_unsigned` (overflow can never happen with these values anyway, so there's no difference in functionality).

krobelus: We could convert to u64, do the computation (with wrapping behavior), and convert the result back to i64.
Probably that's what the C++ version did. The result should be correct because it's two's complement.

PR #9601, opened 2023-02-21
cmake: import corrosion as an in-tree dependency

zanchey:
Importing corrosion into the tree makes running the build on package builders a
lot more straightforward, and since we're not planning to do a release with it,
it never needs to be shipped in an official tarball.


zanchey:
This does slow down the CMake portion of the build a bit.

krobelus:
> This does slow down the CMake portion of the build a bit.

not for incremental builds, right?

zanchey:
> > This does slow down the CMake portion of the build a bit.
> 
> not for incremental builds, right?

Correct, only when CMake needs to regenerate (and there's some commentary about this not applying to CMake 3.19).

The alternative is to package corrosion for our use, which might not be too hard.

mqudsi:
I hate (working with) git submodules with a passion but I wonder if this might not be a good candidate for that, especially if it's short-term? It'll prevent polluting the main repo history and avoid bloating its size (but I checked and this barely registers, has no huge binaries, etc.)

zanchey:
I think I can come up with an alternative solution.

Issue #9600, opened 2023-02-20
If `build.target-dir` is set the build test will fail


akiirui:
If you set `build.target-dir` in `$CARGO_HOME/config.toml`, then the build test will fail.

config.toml:
```toml
[build]
target-dir = "/tmp/cargo"
```

<details>
<summary> Logs </summary>

```
make: Entering directory '/dev/shm/fish-shell/build'
[  2%] Built target fish_test_helper
[  2%] Built target CHECK-FISH-BUILD-VERSION-FILE
[  2%] Built target cargo-prebuild_fish-rust
    Finished release [optimized] target(s) in 0.05s
Copying byproducts `libfish_rust.a` to /home/rui/Memory/fish-shell/build
[  2%] Built target _cargo-build_fish-rust
[  2%] Built target cargo-build_fish-rust
[ 96%] Built target fishlib
[ 98%] Built target fish
[  0%] Built target cargo-prebuild_fish-rust
    Finished release [optimized] target(s) in 0.06s
Copying byproducts `libfish_rust.a` to /home/rui/Memory/fish-shell/build
[  0%] Built target _cargo-build_fish-rust
[  0%] Built target cargo-build_fish-rust
[  1%] Built target pofiles_1
[  2%] Built target pofiles_2
[  3%] Built target pofiles_3
[  4%] Built target pofiles_4
[  5%] Built target pofiles_5
[  6%] Built target pofiles_6
[  7%] Built target pofiles_7
[  7%] Built target CHECK-FISH-BUILD-VERSION-FILE
[ 87%] Built target fishlib
[ 89%] Built target fish
[ 92%] Built target fish_indent
[ 95%] Built target fish_key_reader
[ 96%] Building man pages with Sphinx
[ 96%] Built target sphinx-manpages
[ 97%] Building HTML documentation with Sphinx
[ 97%] Built target sphinx-docs
[ 97%] Built target doc
[100%] Built target fish_test_helper
[100%] Built target build_fish_pc
Install the project...
-- Install configuration: "RelWithDebInfo"
Built target tests_buildroot_target
Copying test files to binary dir
Built target tests_dir
Symlinking fish functions to binary dir
Built target funcs_dir
Built target fish_tests
Test project /dev/shm/fish-shell/build
245/249 Test #248: cargo-test................................***Failed    2.69 sec
   Compiling fish-rust v0.1.0 (/dev/shm/fish-shell/fish-rust)
The following warnings were emitted during compilation:

[...]

PR #9602, opened 2023-02-21
fish-rust/build.rs: Fix hardcode target_dir

akiirui:
## Description

When `FISH_RUST_TARGET_DIR` is not set, `target_dir` will set to `$CARGO_MANIFEST_DIR/target`. 

But if `build.target-dir` or `CARGO_TARGET_DIR` is set,  the real `target_dir` doesn't at the `$CARGO_MANIFEST_DIR/target`.

It causes failure in cargo test.

Then, I do a magic to get real `target_dir` from `OUT_DIR`.

Fixes issue #9600

## TODOs:


akiirui:
I found a better way, just edit `cmake/Tsets.cmake` cargo commands

PR #9607, opened 2023-02-21
bind: port input key binding mappings to rust

nrjais:
## Description

Porting bind builtin with global input mappings to rust

Fixes issue #

## TODOs:

### DIFF
--- a/fish-rust/src/input.rs
+++ b/fish-rust/src/input.rs
@@ -1,23 +1,51 @@
-use crate::wchar::{wstr, WString, L};
+#![allow(clippy::extra_unused_lifetimes, clippy::needless_lifetimes)]
+use crate::ffi::{self, parser_t, readline_cmd_t};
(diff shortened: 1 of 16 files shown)


fish-rust/src/input.rs
```diff
@@ -0,0 +1,395 @@
...
+        };
+        for m in ml {
+            if sequence == m.seq && mode == m.mode {
+                // *out_cmds = m.commands;
+                // *out_sets_mode = m.sets_mode;
+                result = true;
+                break;
+            }
+        }
+        return result;
```
wuyudi, line 154: try

```rust
return ml
    .iter()
    .any(|m| sequence == m.seq && mode == m.mode)
```

zanchey:
@nrjais, thanks for drafting this. I see it needs a little bit of work to bring it up to where the master branch is, and it isn't building at present; do you think you'll be able to keep working on it or should I close this for now?

PR #9608, opened 2023-02-23
Port job_group to rust

mqudsi:
More ugliness with types that cxx bridge can't recognize as being POD. Using pointers to get/set `termios` values with an assert to make sure we're using identical definitions on both sides (in cpp from the system headers and in rust from the libc crate as exported).

I don't know why cxx bridge doesn't allow `SharedPtr<OpaqueRustType>` but we can work around it in C++ by converting a `Box<T>` to a `shared_ptr<T>` then convert it back when it needs to be destructed. I can't find a clean way of doing it from the cxx bridge wrapper so for now it needs to be done manually in the C++ code.

Types/values that are drop-in ready over ffi are renamed to match the old cpp names but for types that now differ due to ffi difficulties I've left the `_ffi` in the function names to indicate that this isn't the "correct" way of using the types/methods.


fish-rust/src/job_group.rs
```diff
@@ -0,0 +1,362 @@
...
+/// Basic thread-safe sorted vector of job ids currently in use.
+/// This is currently deliberately leaked to avoid destructor ordering issues - see #6539.
+///
+/// In the C++ codebase this is `static owning_lock<std::vector<job_id_t>>` but `sync::Lazy<T>` does
+/// its own locking in order to be `Send`/`Sync`-safe so we don't need a second mutex.
+///
+/// TODO: Move `CONSUMED_JOB_IDS`, `JobId::acquire_job_id()`, and `JobId::release_job_id()` to a new
+/// type (e.g. `JobPool`) that's owned by the parser so that in addition to being just atomic the
+/// operations are actually properly thread-safe and we can have multiple instances of the parser
+/// operating in different threads, each with its own set of jobs.
```
mqudsi, line 271: Ignore this!

mqudsi:
Weird. Tests passed locally for me (wsl) and they pass on FreeBSD but not Linux or macOS!

It’s late and in hungry so I apologize for leaving y’all with a broken PR but such is life!

ridiculousfish:
job_group_t is an important foundational type and will be central to our multithreaded story. Please ensure it is `Sync` - it represents THE bits of a single job which may be shared between threads, when we have multiple concurrent fish functions in a pipeline.

Of course don't merge with failing pipelines, I can help debug if you need; but the failures suggest something around signal handling.

mqudsi:
Yup, I have `assert_send()` and `assert_sync()` both. The `set_pgid()` function which was decorated with "don't worry about thread-safety here" comments is `&mut self` presuming the comments are correct and its only set for the first process in a job (before we need to use it from other jobs) but we can change that to be `&self` if it turns out that it won't compile with `&mut self` (that would involve marshaling the `pid_t` to an atomic i32).

I did the same for `set_modes(_ffi)` which wasn't similarly documented but also didn't use any atomic/protected access. IIRC the mode is only set when the job is being created (and when we `fg` it?) so again if `&mut self` won't compile when that code is ported to rust then it'll have to be changed to store the mode behind a mutex or something.

mqudsi:
I spent way too much time chasing down signal-related stuff but then I actually tried using the built binary and quickly realized it was actually a terminal modes issue. Turns out it was a stupid mistake in the job group modes ffi.

Everything should be good now. I want to restructure the Signals into an enum now that I see which ones overlap (Linux repurposing some codes for its own specific use) but that'll be later.

Let me know if you have any suggestions otherwise I'll merge this later.

fish-rust/src/job_group.rs
```diff
@@ -0,0 +1,352 @@
...
+        #[cxx_name = "cancel_with_signal"]
+        fn cancel_with_signal_ffi(&self, signal: i32);
+        fn set_pgid(&mut self, pgid: i32);
+        #[cxx_name = "get_pgid"]
+        fn get_pgid_ffi(&self) -> UniquePtr<pgid_t>;
+        fn has_job_id(&self) -> bool;
+
+        // cxx bridge doesn't recognize `libc::*` as being POD types, so it won't let us use them in
+        // a SharedPtr/UniquePtr/Box and won't let us pass/return them by value/reference, either.
+        unsafe fn get_modes_ffi(&self, size: usize) -> *const u8; /* actually `* const libc::termios` */
```
krobelus, line 41: Interesting. Another option is to define an opaque Rust type like we do for

	type Features;
	fn mutable_fish_features() -> *mut Features;

fish-rust/src/job_group.rs
```diff
@@ -0,0 +1,352 @@
...
+        unsafe fn set_modes_ffi(&mut self, modes: *const u8, size: usize); /* actually `* const libc::termios` */
+
+        // The C++ code uses `shared_ptr<JobGroup>` but cxx bridge doesn't support returning a
+        // `SharedPtr<OpaqueRustType>` nor does it implement `Arc<T>` so we return a box and then
+        // convert `rust::box<T>` to `std::shared_ptr<T>` with `box_to_shared_ptr()` (from ffi.h).
```
krobelus, line 46: Yes, any non-POD Rust type we return from Rust to C++ we need to Box.

src/ffi.h
```diff
@@ -0,0 +1,15 @@
...
+#include "cxx.h"
+#if INCLUDE_RUST_HEADERS
+// For some unknown reason, the definition of rust::Box is in this particular header:
+#include "parse_constants.rs.h"
+#endif
+
+template <typename T>
+std::shared_ptr<T> box_to_shared_ptr(rust::Box<T> &&value) {
+    T *ptr = value.into_raw();
+    std::shared_ptr<T> shared(ptr, [](T *ptr) { rust::Box<T>::from_raw(ptr); });
```
krobelus, line 13: sweet! I had blindly accepted that box is viral which might have led to overcomplicated code

fish-rust/src/job_group.rs
```diff
@@ -0,0 +1,352 @@
...
+        fn get_command_ffi(&self) -> UniquePtr<CxxWString>;
+        #[cxx_name = "get_job_id"]
+        fn get_job_id_ffi(&self) -> i32;
+        #[cxx_name = "get_cancel_signal"]
+        fn get_cancel_signal_ffi(&self) -> i32;
+        #[cxx_name = "cancel_with_signal"]
+        fn cancel_with_signal_ffi(&self, signal: i32);
+        fn set_pgid(&mut self, pgid: i32);
+        #[cxx_name = "get_pgid"]
+        fn get_pgid_ffi(&self) -> UniquePtr<pgid_t>;
```
krobelus, line 36: I would have retunred an `i32` but this is fine too.
If i32 really doesn't work that's surprising to say the least.

fish-rust/src/job_group.rs
```diff
@@ -0,0 +1,352 @@
...
+            Some(value) => UniquePtr::new(pgid_t { value }),
+            None => UniquePtr::null(),
+        }
+    }
+
+    /// Returns the current terminal modes associated with the `JobGroup` for ffi purposes.
+    unsafe fn get_modes_ffi(&self, size: usize) -> *const u8 {
+        assert_eq!(
+            size,
+            core::mem::size_of::<libc::termios>(),
```
krobelus, line 215: nit: we should be consistent about whether we use`std::mem` or `core::mem`


---

I haven't reviewed in detail but this seems nice :+1:

src/ffi.h
```diff
@@ -0,0 +1,15 @@
...
+
+#include "cxx.h"
+#if INCLUDE_RUST_HEADERS
+// For some unknown reason, the definition of rust::Box is in this particular header:
+#include "parse_constants.rs.h"
+#endif
+
+template <typename T>
+std::shared_ptr<T> box_to_shared_ptr(rust::Box<T> &&value) {
+    T *ptr = value.into_raw();
```
krobelus, line 12: can  we consolidate this with `src/rustffi.h`?

fish-rust/src/job_group.rs
mqudsi: i32 works but then we need to use a sentinel value like `-1` to indicate `None`.

src/ffi.h
mqudsi: There is no `rustffi.h` :)

krobelus: I might add it, and there is `rustffi.cpp` which has definitions for  UniquePtr<CxxWString>. Question is whether these two things go into the same bucket. I think yes.

Issue #8924, opened 2022-04-30
Filename suffix "prioritization" is broken for multiple suffixes


mhmdanas:
## System info

- Fish version: 3.4.1
- OS: macOS 12.3.1
- Terminal: Alacritty 0.10.1 with tmux 3.2a

## Bug

A common pattern in Fish completions is something like the following:

https://github.com/fish-shell/fish-shell/blob/908da627b8a37290a23996e6fe0f5f6c8d3273b3/share/completions/unzip.fish#L44-L48

The idea is that files with these suffixes would be prioritized over others. However, what actually happens is that only the first suffix is prioritized. See https://asciinema.org/a/M3W6vObHVvm1Ldygwp3Qc352B for a quick demonstration with `unzip`, but this applies to all completions that use such a pattern.

This bug is reproducible even without third-party customizations.

https://github.com/fish-shell/fish-shell/issues/8512 may be relevant.

mqudsi:
@zanchey I pushed the fix for this into the 3.6.1 branch.

PR #9611, opened 2023-02-24
fix 8924 in `__fish_complete_suffix`

lengyijun:
Before:
hand write arg parse
only accepts one suffix

After:
use `arg_parse` to parse args
accepts multi suffixes

## Description

Fixes issue #8924

## TODOs:

- [ ] Fix comment at the head of __fish_complete_suffix.fish
I'm not native English speaker, please help me with the comment


share/completions/bzip2recover.fish
```diff
@@ -1,11 +1,9 @@
 complete -c bzip2recover -k -x -a "(
-	__fish_complete_suffix .tbz
-	__fish_complete_suffix .tbz2
+	__fish_complete_suffix .tbz .tbz2
 )
```
mqudsi, line 3: If it's now only one command, just place it on a single line instead of multiple lines (same for the others, please).

share/functions/__fish_complete_suffix.fish
```diff
@@ -8,71 +8,36 @@
...
-        # If $comp is "./ma" and the file is "main.py", we'll catch that case here,
+        # If $_flag_comp is "./ma" and the file is "main.py", we'll catch that case here,
         # but complete.cpp will not consider it a match, so we have to output the
         # correct form.
 
         # Also do directory completion, since there might be files with the correct
         # suffix in a subdirectory.
         set all $base*
-        set files_with_suffix (string match -r -- ".*"(string escape --style=regex -- $suff) $all)
+        set files_with_suffix (string match -r -- (string join "|" ".*"$suff) $all)
```
mqudsi, line 40: I don't understand why it's ok to remove the regex escape here?

share/functions/__fish_complete_suffix.fish
```diff
@@ -8,71 +8,36 @@
 # of $PWD.
 
 function __fish_complete_suffix -d "Complete using files"
+    set -l _flag_prefix ""
+    set -l _flag_comp (commandline -ct)
 
-    # Variable declarations
+    argparse 'prefix=' 'desc=' 'comp=' -- $argv
```
mqudsi, line 14: Given how few times they're invoked, I think I would prefer long names instead of `desc` and `comp` as we don't really use those anywhere.

mqudsi (changes requested):
I'm not against cleaning up our internal functions; some of them have grown organically over time and are harder to grok. This is probably a good candidate and it seems that the common usage has remained unchanged so +1 for that.

I have some specific comments if you don't mind addressing or answering.

share/functions/__fish_complete_suffix.fish
lengyijun:
```
# share/functions/__fish_complete_suffix.fish 
# line 6
 argparse 'prefix=' 'description=' 'complete=' -- $argv
```

lengyijun:
cheatsheet of new `__fish_complete_suffix.fish`

```
# without argument, return all files
__fish_complete_suffix.fish

# return files with  suffix .zip or .rar, alphabetically
__fish_complete_suffix.fish .zip .rar
```

lengyijun:
Further work:

`__fish_complete_suffix.fish .zip .rar`
Some user may want `.zip` file before `.rar` file ?

mqudsi:
Grouping all `.zip` before all `.rar` shoudn't be the default behavior, you can add a `-k` flag to match our sorted group flag for completions in theshell.

mqudsi:
@lengyijun I merged this (and fixed the comment at the top of the file as you mentioned). Thank you very much for taking this on, great work!

PR #9613, opened 2023-02-25
Rewrite `wrealpath` from `wutil` in Rust

Tyrubias:
Reopen of #9567, which was closed because the target branch was merged. This is needed for a number of other builtins. `autocxx` didn't generate the overloads for `str2wcstring` so I had to change the order of declaration to make it generate the "right" one.

### DIFF
--- a/fish-rust/src/wchar_ffi.rs
+++ b/fish-rust/src/wchar_ffi.rs
@@ -143,3 +143,15 @@ impl WCharFromFFI<WString> for cxx::SharedPtr<cxx::CxxWString> {
         WString::from_chars(self.as_chars())
     }
 }
+
+impl WCharFromFFI<Vec<u8>> for cxx::UniquePtr<cxx::CxxString> {
+    fn from_ffi(&self) -> Vec<u8> {
+        self.as_bytes().to_vec()
+    }
+}
+
+impl WCharFromFFI<Vec<u8>> for cxx::SharedPtr<cxx::CxxString> {
+    fn from_ffi(&self) -> Vec<u8> {
+        self.as_bytes().to_vec()
+    }
+}
--- a/fish-rust/src/wutil/wrealpath.rs
+++ b/fish-rust/src/wutil/wrealpath.rs
@@ -1,25 +1,25 @@
 use std::{ffi::OsStr, fs::canonicalize, os::unix::prelude::OsStrExt};
 
-use cxx::{CxxString, UniquePtr};
+use cxx::let_cxx_string;
 
 use crate::{
-    ffi::{make_string, str2wcstring, wcs2string},
+    ffi::{str2wcstring, wcs2string},
     wchar::{wstr, WString},
     wchar_ffi::{WCharFromFFI, WCharToFFI},
 };
 
+/// Wide character realpath. The last path component does not need to be valid. If an error occurs,
+/// `wrealpath()` returns `None`
 pub fn wrealpath(pathname: &wstr) -> Option<WString> {
     if pathname.is_empty() {
         return None;
     }
 
-    let mut real_path: UniquePtr<CxxString> = make_string("");
-    let narrow_path_ptr: UniquePtr<CxxString> = wcs2string(&pathname.to_ffi());
-    // We can't `pop` from a `UniquePtr<CxxString>` so we copy it into a vector
-    let mut narrow_path = narrow_path_ptr.as_bytes().to_vec();
+    let mut real_path = Vec::new();
+    let mut narrow_path: Vec<u8> = wcs2string(&pathname.to_ffi()).from_ffi();
 
     // Strip trailing slashes. This is treats "/a//" as equivalent to "/a" if /a is a non-directory.
-    while narrow_path.len() > 1 && narrow_path[narrow_path.len() - 1] as char == '/' {
+    while narrow_path.len() > 1 && narrow_path[narrow_path.len() - 1] == b'/' {
         narrow_path.pop();
     }
 
@@ -29,18 +29,16 @@ pub fn wrealpath(pathname: &wstr) -> Option<WString> {
     let narrow_res = canonicalize(OsStr::from_bytes(&narrow_path));
 
     if let Ok(result) = narrow_res {
-        real_path
-            .pin_mut()
-            .push_bytes(result.as_os_str().as_bytes());
+        real_path.extend_from_slice(result.as_os_str().as_bytes());
     } else {
         // Check if everything up to the last path component is valid.
-        let pathsep_idx = narrow_path.iter().rposition(|&c| c as char == '/');
+        let pathsep_idx = narrow_path.iter().rposition(|&c| c == b'/');
 
         if let Some(0) = pathsep_idx {
             // If the only pathsep is the first character then it's an absolute path with a
             // single path component and thus doesn't need conversion.
-            real_path.pin_mut().clear();
-            real_path.pin_mut().push_bytes(&narrow_path);
+            real_path.clear();
+            real_path.extend(narrow_path);
         } else {
             // Only call realpath() on the portion up to the last component.
             let narrow_res = if let Some(pathsep_idx) = pathsep_idx {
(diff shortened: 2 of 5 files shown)


src/common.h
```diff
@@ -289,10 +289,10 @@ void show_stackframe(int frame_count = 100, int skip_levels = 0);
 ///
 /// This function encodes illegal character sequences in a reversible way using the private use
 /// area.
-wcstring str2wcstring(const char *in);
-wcstring str2wcstring(const char *in, size_t len);
 wcstring str2wcstring(const std::string &in);
 wcstring str2wcstring(const std::string &in, size_t len);
+wcstring str2wcstring(const char *in);
+wcstring str2wcstring(const char *in, size_t len);
```
krobelus, line 295: is there a reason for moving these? Otherwise better move them back

fish-rust/src/wutil/wrealpath.rs
```diff
@@ -0,0 +1,72 @@
...
+
+pub fn wrealpath(pathname: &wstr) -> Option<WString> {
+    if pathname.is_empty() {
+        return None;
+    }
+
+    let mut real_path: UniquePtr<CxxString> = make_string("");
+    let narrow_path_ptr: UniquePtr<CxxString> = wcs2string(&pathname.to_ffi());
+    // We can't `pop` from a `UniquePtr<CxxString>` so we copy it into a vector
+    let mut narrow_path = narrow_path_ptr.as_bytes().to_vec();
```
krobelus, line 19: please use native Rust strings wherever possible. Put the conversions close to the FFI boundary.
For example

	let narrow_path = wcs2string(&pathname.to_ffi()).from_ffi();

or is there a reason we can't use WString like embedded nul characters?

src/common.h
krobelus: ok I just saw your comment. Better to include the motivation in the commit message

fish-rust/src/wutil/wrealpath.rs
faho: >or is there a reason we can't use WString like embedded nul characters?

Paths can't have nul characters.

fish-rust/src/wutil/wrealpath.rs
```diff
@@ -0,0 +1,72 @@
...
+
+    // Strip trailing slashes. This is treats "/a//" as equivalent to "/a" if /a is a non-directory.
+    while narrow_path.len() > 1 && narrow_path[narrow_path.len() - 1] as char == '/' {
+        narrow_path.pop();
+    }
+
+    // `from_bytes` is Unix specific but there isn't really any other way to do this
+    // since `libc::realpath` is also Unix specific. I also don't think we support Windows
+    // outside of WSL + Cygwin (which should be fairly Unix-like anyways)
```
faho, line 28: This is correct.

fish-rust/src/wutil/wrealpath.rs
Tyrubias: I originally used the FFI strings because it was easier to treat them as a "bag of bytes" which is how Rust handles inputs to `std::fs::canonicalize`. I can see why it might be better to minimize the use of FFI types

Tyrubias: We also don't have an implementation of `WCharFromFFI<WString>` for `cxx::UniquePtr<cxx::CxxString>`

faho: Converting "string" to "wstring" is what str2wcstring does.

I can't see any place we use it on the rust side yet, so this might be the first.

I don't believe we want a WcharFromFFI for this, because it's a more involved operation?

Tyrubias: The original implementation of `wrealpath` does the opposite here, converting from a wide string to a "regular" string (i.e. basically an array of bytes) via `wcs2string` so it can be passed to the `realpath` function. If we want to keep the original behavior while avoiding FFI types shouldn't we use a `Vec<u8>` or maybe an `OsStr` here?

Tyrubias: Upon further experimentation it seems like `Vec<u8>` is the most suitable to replicate the structure of the original code. I'll push up a change for y'all to check out.

fish-rust/src/wchar_ffi.rs
```diff
@@ -143,3 +143,15 @@ impl WCharFromFFI<WString> for cxx::SharedPtr<cxx::CxxWString> {
         WString::from_chars(self.as_chars())
     }
 }
+
+impl WCharFromFFI<Vec<u8>> for cxx::UniquePtr<cxx::CxxString> {
```
mqudsi, line 147: Nitpick: please use `WCharFromFfi` to use conventional PascalCase (not all-caps FFI).

fish-rust/src/wutil/wrealpath.rs
```diff
@@ -0,0 +1,70 @@
...
+    let narrow_res = canonicalize(OsStr::from_bytes(&narrow_path));
+
+    if let Ok(result) = narrow_res {
+        real_path.extend_from_slice(result.as_os_str().as_bytes());
```
mqudsi, line 32: Please use `.into_os_string().into_vec()` instead of `as_os_str().as_bytes()` to avoid re-allocating.
Also just assign directly to `real_path` instead of calling `extend_from_slice()` because that might end up allocating again (I'm not sure what std currently does if calling that on a zero-size buffer, but just assigning is clearer because it makes it clear we're not appending).

fish-rust/src/wutil/wrealpath.rs
```diff
@@ -0,0 +1,70 @@
...
+};
+
+/// Wide character realpath. The last path component does not need to be valid. If an error occurs,
+/// `wrealpath()` returns `None`
+pub fn wrealpath(pathname: &wstr) -> Option<WString> {
+    if pathname.is_empty() {
+        return None;
+    }
+
+    let mut real_path = Vec::new();
```
mqudsi, line 18: Instead of declaring this as mut and then assigning to it (and clearing when it hasn't been written to) can you just return this from the `match` block below?

mqudsi (changes requested):

fish-rust/src/wchar_ffi.rs
Tyrubias: The trait is currently named `WCharFromFFI` and not `WCharFromFfi`, are you suggesting that I rename the entire trait?

mqudsi: Yes

mqudsi: Sorry, I thought you added the trait. No, please leave the existing trais as-is and ignore this for now! :)

fish-rust/src/wutil/wrealpath.rs
Tyrubias: I've made this change; originally I thought it'd be cheaper to return a slice. And yes, `extend_from_slice` does allocate again. I think this change combined with the change you suggested to return `real_path` from the `if/else` does really help in reducing the number of allocations. Thank you so much!

fish-rust/src/wutil/wrealpath.rs
```diff
@@ -28,17 +31,16 @@ pub fn wrealpath(pathname: &wstr) -> Option<WString> {
...
 
-    if let Ok(result) = narrow_res {
-        real_path.extend_from_slice(result.as_os_str().as_bytes());
+    let real_path = if let Ok(result) = narrow_res {
+        result.into_os_string().into_vec()
     } else {
         // Check if everything up to the last path component is valid.
         let pathsep_idx = narrow_path.iter().rposition(|&c| c == b'/');
 
         if let Some(0) = pathsep_idx {
```
mqudsi, line 40: Can you use a regular comparison here? (`if pathsep_idx == Some(0) {`)
There's no need to use let binding assignment.

Tyrubias: Ah yes, good point. I got so used to doing it for any assignment that I forgot we didn't always need it.

mqudsi (commented):
This looks much better; thank you for addressing the issues pointed out.

Is it your intention that this should replace the existing usage of wrealpath or only to be used as-needed from rust? Because if it's the former, you now need to work on the other half of the ffi where the cpp code uses the new rust implementation.

Tyrubias:
I intend for this to be used as-needed from Rust (just like the other utility functions in this module). I decided to make this PR after trying to port some other builtins and noticing that a number of them depend on `wrealpath` (most notably `pwd` and `realpath`). I don't think there's an easy way to replace `maybe_t` with `Option` across the FFI boundary when operating in C++.

mqudsi:
Yeah; there’s no good interop between maybe_t and option. The best alternative is to replace Option with UniquePtr which behaves fairly similar to maybe_t except it is heap allocated.

PR #9614, opened 2023-02-25
Port fish_wcstod*() to Rust

Xiretza:
## Description

Unfortunately the Rust standard library does not offer partial parsing (leaving the unparseable suffix alone), so this reimplements both the C standard library's `wcstod` and fish's modifications in Rust.

There are some differences which I haven't been able to paper over so far, I'm not sure how imporant they are:

- e.g. `0x` is a parse error rather than `(0.0, "x")`; for that one in particular there's even a test for the `math` builtin (`math "0x 3"` should be the same as `0 * 3`, but is an error with the new float parser).
- `nan(foo)` creates platform-specific NAN values. The new parser does not recognize brackets after NAN at all.

## TODOs:

### DIFF
--- a/fish-rust/src/wutil/wcstoi.rs
+++ b/fish-rust/src/wutil/wcstoi.rs
@@ -13,6 +13,8 @@ struct ParseResult {
     result: u64,
     negative: bool,
     consumed_all: bool,
+    radix: u32,
+    num_digits: u32,
 }
 
 /// Helper to get the current char, or \0.
@@ -88,7 +90,7 @@ where
         _ => negative = false,
     }
 
-    let mut consumed1 = false;
+    let mut num_digits = 0;
 
     // Determine the radix.
     let radix = match mradix {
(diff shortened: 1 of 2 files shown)


faho:
>math "0x 3" should be the same as 0 * 3, but is an error with the new float parser

This should be reasonably important. It's tested in math (because the `x` special-case came later, because `*` is annoying), but it would also be a problem e.g. if wcsfilecmp used this (which it no longer does - we compare the numbers directly as strings).

So I can see this being an awkward change in weird places.

>nan(foo) creates platform-specific NAN values. The new parser does not recognize brackets after NAN at all.

I believe basically anywhere we currently accept NaN is a likely bug, and a cursory glance shows `test` prints a special error (but still an error), while `math` doesn't even attempt to parse (because it checks if the first char is a possible hex digit first).

So I do not see this being a problem.

ridiculousfish:
Sorry, I'm going to have to decline this as implemented. Precise float parsing [is quite difficult](https://www.reddit.com/r/rust/comments/a6j5j1/making_rust_float_parsing_fast_and_correct/), requiring bignum arithmetic in some cases; it's not a function we can do ad-hoc like this.

I've got this pretty far along in [this commit](https://github.com/fish-shell/fish-shell/commit/d81f435548bcc54916767e54c3fe4430148a68d4), which is based on the venerable dtoa. The [fast-float](https://github.com/aldanor/fast-float-rust) crate is also possible but I'm further along with dtoa. We'll need to teach both about non-period decimal separators.

Xiretza:
Yeah, makes sense, this was more of a hack than an actual correct implementation. I'm glad the `strtod` code mostly works as-is, it doesn't exactly look easy to debug :D

PR #9615, opened 2023-02-26
Port math module and tinyexpr to Rust

Xiretza:
Still requires an implementation of `fish_wcstod_underscores`.


src/builtins/math.cpp
```diff
@@ -1,302 +0,0 @@
-// Implementation of the math builtin.
-#include "config.h"  // IWYU pragma: keep
```
krobelus, line 2: please put the deletion in the commit that obsoletes the file.
That makes recursive blame easier.

Might need to combine math + tinyexpr changes to one commit but that's fine.

fish-rust/src/tinyexpr.rs
```diff
@@ -0,0 +1,710 @@
...
+
+fn npr(n: f64, r: f64) -> f64 {
+    ncr(n, r) * fac(r)
+}
+
+const BUILTINS: &[(&wstr, Function)] = &[
+    // must be in alphabetical order
+    (L!("abs"), Function::Fn1(f64::abs)),
+    (L!("acos"), Function::Fn1(f64::acos)),
+    (L!("asin"), Function::Fn1(f64::asin)),
```
krobelus, line 286: this is where I'd use the widestrs macro

Xiretza:
I've rebased on top of #9666, which adds `wcstod`. `wcstod_underscores` doesn't completely work yet for two reasons:

1. fast-float does not recognize custom NaN values (`nan(123)`) and parses them as `nan` with `(123)` left over
2. fast-float does not recognize hexadecimal float literals (https://github.com/fish-shell/fish-shell/pull/9666#discussion_r1141036423).

The first is probably not much of an issue, the second still needs to be solved in #9666 anyway.

Xiretza:
Alright, should be good to go now! `nan(foo)` is still parsed as `nan` with `(foo)` left over, should I add a changelog entry for that?

faho:
>Alright, should be good to go now! nan(foo) is still parsed as nan with (foo) left over, should I add a changelog entry for that?

How exactly is this accessible to users?

`math` is meant to error out for things like nan and infinite.

Xiretza:
You're right, I actually missed a condition in my `tinyexpr` port, it shouldn't even recognize `nan` as a number. And since `wcstod_underscores` isn't used anywhere else other than tests, I should just remove it from C++.

fish-rust/src/wutil/wcstod.rs
```diff
@@ -2,28 +2,11 @@ use super::errors::Error;
...
-/// The `consumed` parameter is used to return the number of characters
-/// consumed, similar to the "end" parameter to strtod.
-/// This is only meaningful if parsing succeeds.
-///
-/// Error::Overflow is returned if the value is too large in magnitude.
-pub fn wcstod<Chars>(input: Chars, decimal_sep: char, consumed: &mut usize) -> Result<f64, Error>
+fn wcstod_inner<I>(mut chars: I, decimal_sep: char, consumed: &mut usize) -> Result<f64, Error>
 where
-    Chars: IntoCharIter,
+    I: Iterator<Item = char> + Clone,
```
krobelus, line 7: I'm curious where this will be used, especially since it's not public

Maybe we should implement `IntoCharIter` for all types that are `Iterator<Item = char> + Clone`, then a single caller taking `IntoCharIter` would work for those too

fish-rust/src/builtins/math.rs
```diff
@@ -0,0 +1,319 @@
...
+    streams.stdin_is_directly_redirected()
+}
+
+/// Get the arguments from stdin.
+fn get_arg_from_stdin(streams: &io_streams_t) -> Option<WString> {
+    let mut s = WString::new();
+    loop {
+        let mut buf = [0];
+        let c = match read_blocked(streams.stdin_fd().unwrap().as_raw_fd(), &mut buf) {
+            1 => char::from(buf[0]),
```
krobelus, line 135: You are converting each input byte to a wide character and appending it to the string.
This may break on multibyte characters.
To be safe we should read into a `Vec<u8>` and convert it with `str2wcstring`, as the C++ version did.

fish-rust/src/builtins/shared.rs
```diff
@@ -115,6 +123,22 @@ impl io_streams_t {
     pub fn ffi_ref(&self) -> &builtins_ffi::io_streams_t {
         unsafe { &*self.streams }
     }
+
+    pub fn stdin_is_directly_redirected(&self) -> bool {
+        self.ffi_ref().ffi_stdin_is_directly_redirected()
+    }
+
+    pub fn stdin_fd(&self) -> Option<BorrowedFd> {
```
krobelus, line 131: other ports use `RawFd`. I hope this is not causing needless friction.
If we change types like this we need to think ahead and have a clear idea of how to port all related code.
Looks like this implicitly converts to `i32` (I guess we should change the `read_blocked` parameter to `RawFd` at least)

fish-rust/src/builtins/math.rs
```diff
@@ -0,0 +1,319 @@
...
+        ret
+    }
+}
+
+/// Return a formatted version of the value `v` respecting the given `opts`.
+fn format_double(mut v: f64, opts: &Options) -> WString {
+    if opts.base == 16 {
+        v = v.trunc();
+        let mneg = if v.is_sign_negative() { "-" } else { "" };
+        return sprintf!(L!("%s0x%lx"), mneg, v.abs() as u64);
```
krobelus, line 189: We don't need the `L!` macro when using `sprintf!`

fish-rust/src/builtins/math.rs
```diff
@@ -0,0 +1,319 @@
...
+        return sprintf!(L!("%.*f"), opts.scale, v);
+    }
+
+    let mut ret = sprintf!(L!("%.*f"), opts.scale, v);
+    // If we contain a decimal separator, trim trailing zeros after it, and then the separator
+    // itself if there's nothing after it. Detect a decimal separator as a non-digit.
+    if ret.chars().any(|c| !c.is_ascii_digit()) {
+        let trailing_zeroes = ret.chars().rev().take_while(|&c| c == '0').count();
+        let mut to_keep = ret.len() - trailing_zeroes;
+        if ret.as_char_slice()[to_keep - 1] == '.' {
```
krobelus, line 213: This code used to handle alternative decimal separators. I don't think they are still supported by `math` so this should be fine

fish-rust/src/tinyexpr.rs
```diff
@@ -0,0 +1,714 @@
...
+            let mut consumed = 0;
+            let num = wcstod_underscores(*next, get_numeric_locale().decimal_point, &mut consumed)
+                .unwrap();
+            Some((consumed, Some(Token::Number(num))))
+        } else {
+            // Look for a function call.
+            // But not when it's an "x" followed by whitespace
+            // - that's the alternative multiplication operator.
+            if next.first()?.is_ascii_lowercase()
+                && !(*next.first()? == 'x' && next.get(1)?.is_whitespace())
```
krobelus, line 410: I'm not sure if the question marks do what we want here.
Given `next.first().is_none()`, the C++ version would go into the else branch ("Look for an operator or special character") where it would report a MissingOperator error.

fish-rust/src/tinyexpr.rs
```diff
@@ -0,0 +1,714 @@
...
+                if have_open {
+                    // If we *have* an opening parenthesis,
+                    // we need to consume it and
+                    // expect a closing one.
+                    self.next_token();
+                }
+
+                if let Function::Fn0(f) = f {
+                    if have_open {
+                        if let Token::Close = self.current {
```
krobelus, line 513: nit: I don't like the yoda condition. Can we maybe implement Eq or PartialEq?

fish-rust/src/tinyexpr.rs
```diff
@@ -0,0 +1,714 @@
...
+                            self.next_token();
+                        } else if self.no_specific_error() {
+                            self.set_error(ErrorKind::MissingClosingParen, None);
+                        }
+                    }
+                    return f();
+                }
+                if let Function::Constant(n) = f {
```
krobelus, line 521: Why not use `arity == 0` for both Fn0 and Constants as before?

fish-rust/src/wutil/wcstod.rs
```diff
@@ -106,6 +111,94 @@
...
 }
 
+/// Like [`wcstod()`], but allows underscore separators. Leading, trailing, and multiple underscores
+/// are allowed, as are underscores next to decimal (`.`), exponent (`E`/`e`/`P`/`p`), and
+/// hexadecimal (`X`/`x`) delimiters. This consumes trailing underscores -- `consumed` will include
+/// the last underscore which is legal to include in a parse (according to the above rules).
+/// Free-floating leading underscores (`"_ 3"`) are not allowed and will result in a no-parse.
+/// Underscores are not allowed before or inside of `"infinity"` or `"nan"` input. Trailing
+/// underscores after `"infinity"` or `"nan"` are not consumed.
+pub fn wcstod_underscores<Chars>(
```
krobelus, line 121: okay it's free to change this name because there are no C++ callers remaining

fish-rust/src/wutil/wcstod.rs
```diff
@@ -106,6 +111,94 @@
...
+    let is_inf_or_nan_char = |c: char| "iInN".contains(c);
+
+    // We don't do any underscore-stripping for infinity/NaN.
+    let mut is_inf_nan = false;
+    if let Some(&c1) = chars.peek() {
+        if is_inf_or_nan_char(c1) {
+            is_inf_nan = true;
+        } else if is_sign(c1) {
+            // FIXME make this more efficient
+            let mut copy = chars.clone();
```
krobelus, line 152: sounds like it was maybe a bit premature to take an iterator instead of a char slice like in C++.
No big deal either way, this is fine.

fish-rust/src/wutil/wcstod.rs
Xiretza: It's used in both `wcstod` and `wcstod_underscores`.

> Maybe we should implement `IntoCharIter` for all types that are `Iterator<Item = char> + Clone`, then a single caller taking `IntoCharIter` would work for those too

I don't think that's possible, you can't have a blanket impl and impls for concrete types (say, `&str`) at the same time because they might collide.

fish-rust/src/builtins/math.rs
faho: Specifically, we currently set our internal LC_NUMERIC to C so *all* our float reading only ever uses `.`, and for math specifically alternative separators were never supported and the popular `,` *cannot* be supported because that's the argument separator (german Excel switches to `;` so it can support `,` decimal separators).

Xiretza: Note that this code looks at the *output* of `sprintf!()`, which currently always uses `C_LOCALE`, but could use something else as well - I guess that does need special handling then?

fish-rust/src/tinyexpr.rs
Xiretza: Not easily, function pointers can't be compared.

fish-rust/src/builtins/math.rs
faho: Depends - does https://github.com/fish-shell/fish-shell/blob/971d257e67bc6482f5a3dbeba5fd70dcb0265029/fish-rust/src/locale.rs#L119-L121 affect the decimal separator or not?

If it doesn't, we need some other way to change the decimal separator here to always always always be `.` specifically.

fish-rust/src/builtins/shared.rs
Xiretza: I've changed it to `RawFd`, and `read_blocked` now also takes a `RawFd`.

fish-rust/src/tinyexpr.rs
Xiretza: Given `!*next_`, the C++ version would already bail with `type_ = TOK_END;` earlier. You are right about the `next.get(1)?` though; in C++ that would be a null byte that fails `isspace` rather than a bail. My solution isn't the prettiest, but [`Option::is_some_and`](rust-lang/rust#93050) will come soon.

fish-rust/src/tinyexpr.rs
Xiretza: It involves an (unreachable) panic because you still have to extract the enum variant's content inside the `if`. I guess that's still prettier than duplicating the `if` body? Please take a look.

fish-rust/src/tinyexpr.rs
```diff
@@ -0,0 +1,710 @@
...
+                self.error,
+                Some(Error {
+                    kind: ErrorKind::Unknown,
+                    ..
+                })
+            )
+    }
+
+    /// Tries to get the next token from the input. If the input does not contain enough data for
+    /// another token, `None` is returned. Otherwise, the numbre of consumed characters is returned
```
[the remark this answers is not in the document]
Xiretza: Heh, fixed. (was already in the C++ version)

fish-rust/src/builtins/math.rs
Xiretza: That line doesn't seem to do anything, it changes the locale after `lconv_to_locale()` has already populated `result` with the locale's decimal separator. Is that a bug or am I missing something?

faho: I've not kept a close eye on the locale port.

What it is *supposed* to do is to set the process' locale so that everything but the "LC_NUMERIC" category is taken from the environment, while LC_NUMERIC is set to C.

This is, of course, assuming c-like semantics where the global locale affects all sorts of functions - so something like strtod/wcstod would automatically be affected after the `setlocale`.

If that isn't the case, here, this needs to be changed. (and if `libc::setlocale` is ineffective in that way, it would be cool if that were documented)

We want to have a decimal separator of `.` almost everywhere.

Xiretza: Oh, I see, it does affect the C/C++ standard library functions that look directly at the locale. Anyway, `math` has always used C numeric locale for its output, and it still does now due to `sprintf!()` always using C locale. I think this is fine as is then.

faho: See also the [locale-numeric test](https://github.com/fish-shell/fish-shell/blob/master/tests/checks/locale-numeric.fish).

PR #9616, opened 2023-02-26
Run rustfmt and clippy in CI

Xiretza:
## Description

This ensures that master always contains rustfmt and clippy-clean code, allowing contributors to use these tools effectively (without a bunch of false positives in existing code).

## TODOs:

### DIFF
--- a/.github/workflows/rust_checks.yml
+++ b/.github/workflows/rust_checks.yml
@@ -1,0 +1,42 @@
+name: Rust checks
+
+on: [push, pull_request]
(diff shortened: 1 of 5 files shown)


mqudsi:
I'll merge this, but we'll likely want to fine-tune clippy's behavior further in the future.

Xiretza:
That can still be done through lint attributes in the source code, the point is that whatever warnings *are* enabled should not be produced by code on `master`.

PR #9619, opened 2023-02-26
Port color.cpp to Rust

Xiretza:
## Description

Ports the bit-packed `rgb_color_t` to Rust.

Does it makes sense to implement FFI here? I haven't found a way to convince cxx not to require a pointer indirection for every color value, which would probably defeat the whole size optimization.

### DIFF
--- a/fish-rust/src/color.rs
+++ b/fish-rust/src/color.rs
@@ -1,0 +1,422 @@
+use std::{array, cmp::Ordering};
+
+use crate::{
+    wchar::{widestrs, wstr, WExt, WString, L},
+    wutil::sprintf,
+};
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub struct Color24 {
+    pub r: u8,
+    pub g: u8,
+    pub b: u8,
+}
+
+impl Color24 {
+    fn from_bits(bits: u32) -> Self {
+        assert_eq!(bits >> 24, 0, "from_bits() called with non-zero high byte");
+
+        Self {
+            r: (bits >> 16) as u8,
+            g: (bits >> 8) as u8,
+            b: bits as u8,
+        }
+    }
+}
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum Type {
+    // TODO: remove this? Users should probably use `Option<RgbColor>` instead
+    None,
+    Named { idx: u8 },
+    Rgb(Color24),
+    Normal,
+    Reset,
+}
+
+#[derive(Debug, Default, Copy, Clone, PartialEq, Eq)]
+pub struct Flags {
+    pub bold: bool,
+    pub underline: bool,
+    pub italics: bool,
+    pub dim: bool,
+    pub reverse: bool,
+}
+
+impl Flags {
+    // const eval workaround
+    const DEFAULT: Self = Flags {
+        bold: false,
+        underline: false,
+        italics: false,
+        dim: false,
+        reverse: false,
+    };
+}
+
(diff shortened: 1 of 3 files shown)


fish-rust/src/common.rs
```diff
@@ -115,3 +115,75 @@ pub fn valid_func_name(name: &wstr) -> bool {
...
+/// const COLORS: &[(&wstr, u32)] = &[
+///     // must be in alphabetical order
+///     (L!("blue"), 0x0000ff),
+///     (L!("green"), 0x00ff00),
+///     (L!("red"), 0xff0000),
+/// ];
+///
+/// assert_sorted_by_name!(COLORS, 0);
+/// ```
+macro_rules! assert_sorted_by_name {
```
ridiculousfish, line 139: Whoa, I did not expect this to be possible in Rust. We may need to disable this check on older versions, but very cool!

fish-rust/src/color.rs
```diff
@@ -0,0 +1,453 @@
...
+pub struct RgbColor {
+    #[bits = 3]
+    typ: Type,
+    #[bits = 5]
+    flags: Flags,
+    /// Either the R, G and B bytes for [`Type::Rgb`], or the name index for [`Type::Named`].
+    data: B24,
+}
+
+const _: () = assert!(size_of::<RgbColor>() <= 4, "RgbColor is too big");
```
ridiculousfish, line 66: Thank you for faithfully propagating this assert, but let's remove it, this isn't important.

fish-rust/src/color.rs
```diff
@@ -0,0 +1,453 @@
...
+
+#[bitfield]
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub struct RgbColor {
+    #[bits = 3]
+    typ: Type,
+    #[bits = 5]
+    flags: Flags,
+    /// Either the R, G and B bytes for [`Type::Rgb`], or the name index for [`Type::Named`].
+    data: B24,
```
ridiculousfish, line 63: Let's either make this a Rust union, matching the C++; or if you are feeling ambitious go ahead and combine it with `typ` into an `enum`.

ridiculousfish (changes requested):
This looks very solid - thank you, well done!

Let's not use bitfields here - my original use of bitfields was not well motivated and I regret it, and I prefer to not add the complexity of `modular_bitfield`. Please make these ordinary bools in an ordinary struct. That and a few other nits and it looks good. Thank you again!

Xiretza:
Not having to do bitfields certainly makes it a lot easier! It's all regular Rust types now.

fish-rust/src/common.rs
```diff
@@ -115,3 +115,75 @@ pub fn valid_func_name(name: &wstr) -> bool {
...
+
+            let mut i = 0;
+            let mut prev: Option<&wstr> = None;
+            while i < $slice.len() {
+                let cur = $slice[i].$field;
+                if let Some(prev) = prev {
+                    assert!(
+                        matches!(
+                            cmp_slice(prev.as_char_slice(), cur.as_char_slice()),
+                            Ordering::Less
```
mqudsi, line 174: Do you want this to be strictly `Ordering::Less` or do we envision a case where we might actually want `!Ordering::Greater`?

Xiretza: You're right, the C++ macro is `<=`. Fixed.

ridiculousfish:
Merged as 7585ddf9262829ae00426150b2c52bfae2478d34, thank you!

PR #9622, opened 2023-02-27
Port input common to rust

nrjais:
## Description

This is needed to port input mappings and bind builtin

Fixes issue #

## TODOs:

### DIFF
--- a/fish-rust/src/input_common.rs
+++ b/fish-rust/src/input_common.rs
@@ -1,0 +1,547 @@
+#![allow(unused_imports)]
+use crate::wchar_ffi::{WCharFromFFI, WCharToFFI};
+use crate::{
+    env::flags,
+    fd_readable_set::{self, new_fd_readable_set},
+    fds::kNoTimeout,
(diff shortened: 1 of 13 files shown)


nrjais:
I haven't removed CPP code yet, I will remove those once I port input and bind to rust. As it is extending the input_queue class and I thought it is better to put the time to directly port it rather than making it work with FFI

fish-rust/src/input_common.rs
```diff
@@ -0,0 +1,547 @@
...
+    pub const readb_eof: i32 = -1;
+
+    // select() was interrupted by a signal.
+    pub const readb_interrupted: i32 = -2;
+
+    // Our uvar notifier reported a change (either through poll() or its fd).
+    pub const readb_uvar_notified: i32 = -3;
+
+    // Our ioport reported a change, so service main thread requests.
+    pub const readb_ioport_notified: i32 = -4;
```
mqudsi, line 59: Please use SCREAMING_SNAKE_CASE for const values.

PR #9625, opened 2023-03-01
builtins: Port `pwd` to Rust

Tyrubias:
Reopen of #9566, which was closed because the target branch was merged, as well as #9595, which was closed because I accidentally deleted the source branch.

I had to add a wrapper for `env_stack_t::get` because it returns `maybe_t<env_var_t>`, which `autocxx` can't handle.


fish-rust/src/builtins/pwd.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    if w.woptind != argc {
+        streams
+            .err
+            .append(wgettext_fmt!(BUILTIN_ERR_ARG_COUNT1, cmd, 0, argc - 1));
+        return STATUS_INVALID_ARGS;
+    }
+
+    let mut pwd = WString::new();
+    let tmp = parser.vars1().get_pwd_slash();
```
ridiculousfish, line 53: Previously this was `get(L"PWD")`. This will change the output to add an extra slash. Is there a reason that `parser.vars().get(L"PWD")` couldn't be ported? Maybe we need to add an interface for accessing variables?

ridiculousfish (changes requested):

fish-rust/src/builtins/pwd.rs
Tyrubias: I think there has been some discussion on this before, but the essence of the issue is that `env_stack_t::get` returns a `maybe_t<env_var_t>` which `autocxx` can't handle. I know `get_pwd_slash` will produce an extraneous trailing slash so I added code to `pop()` the slash. I guess an alternative might be to write some kind of conversion in the C++ codebase from `maybe_t` to something like `unique_ptr`. However, I think these kinds of issues might only be fully resolved once the parser is ported (there are a number of other builtins that rely on parser functionality).

mqudsi: You can't rely on popping the slash because PWD might have been set with the trailing slash (even if that's not as correct). Unless I'm misunderstanding.

Tyrubias: I think you're right. I'm going to push a change to make sure `$PWD` doesn't have a trailing slash.

Tyrubias: Actually I think it might be more complicated that that. Both `env_stack_t::get` and `env_stack_t::get_pwd_slash` appear to read `$PWD` from `perproc_data.pwd`. `env_stack_t::get_pwd_slash` checks if there is a trailing slash and if there isn't one already, it appends a trailing slash. So we have no way of knowing whether the trailing slash is there "intentionally" or not.

Tyrubias: Per our discussion, I've introduced a wrapper for `env_stack_t::get` like @ridiculousfish suggested.

fish-rust/src/env.rs
```diff
@@ -35,3 +35,17 @@ pub mod status {
...
+    ENV_LOCAL = 1 << 0,
+    ENV_FUNCTION = 1 << 1,
+    ENV_GLOBAL = 1 << 2,
+    ENV_UNIVERSAL = 1 << 3,
+    ENV_EXPORT = 1 << 4,
+    ENV_UNEXPORT = 1 << 5,
+    ENV_PATHVAR = 1 << 6,
+    ENV_UNPATHVAR = 1 << 7,
+    ENV_USER = 1 << 8,
+}
```
mqudsi, line 51: Can you please use idiomatic rust naming here? `EnvMode` as the name and `Default`, `Local`, etc. as the values.

krobelus: These constants already exist a few lines above, as `pub const`.
I'm not sure if it should be an enum, does that make it easy to write something like

    has_scope = mode & (ENV_LOCAL | ENV_FUNCTION | ENV_GLOBAL | ENV_UNIVERSAL);

we could also use the bitflags crate.

mqudsi: @krobelus  it doesn’t - bitflags probably makes more sense.

Tyrubias: The reason I defined this enum again is because the above definitions are `c_int` whereas they're more appropriate as `u16` which is the most similar to their C++ definition. I do think the bitflags crate makes more sense for these constants.

Tyrubias: Out of curiosity, which is more appropriate: [`bitflags`](https://lib.rs/crates/bitflags) or [`enumflags2`](https://lib.rs/crates/enumflags2)? @krobelus @mqudsi

mqudsi (approved):
Looks good. The wrapper for get_or_null() is definitely a cleaner solution.

fish-rust/src/builtins/pwd.rs
```diff
@@ -0,0 +1,80 @@
...
+    wgetopt::{wgetopter_t, wopt, woption, woption_argument_t::no_argument},
+    wutil::{wgettext_fmt, wrealpath},
+};
+
+use super::shared::{
+    builtin_print_help, builtin_unknown_option, STATUS_CMD_ERROR, STATUS_CMD_OK,
+    STATUS_INVALID_ARGS,
+};
+
+/// The pwd builtin. Respect -P to resolve symbolic links. Respect -L to not do that (the default).
```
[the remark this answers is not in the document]
Tyrubias: Ahh makes sense I'll fix that right now

mqudsi:
Merged. Thanks a bunch, @Tyrubias.

PR #9626, opened 2023-03-01
Port trace to Rust

clemenswasser:
## Description

The only notable difference to the C++ version is that I had to replace the `const std::vector<std::wstring>&` with pointer and size as `CxxWString` does not implement the [`VectorElement`](https://docs.rs/cxx/1.0.91/cxx/vector/trait.VectorElement.html) trait which is required for `CxxVector` support. It sadly has to use very unsafe pointer arithmetic, let's hope that this goes away as soon as more is ported to Rust or when someone has a better idea how to do this safely.

Fixes issue #

## TODOs:


clemenswasser:
The `escape_string` wrapper seems to have issues. I'll try to fix this...

clemenswasser:
The bug actually seems to be in `escape_string_script`?
When escaping the string `"1"` the result seems to be `{(wchar_t)49, (wchar_t)22060}`?

fish-rust/src/common.rs
```diff
@@ -89,5 +89,5 @@ pub fn escape_string(s: &wstr, style: EscapeStringStyle) -> WString {
         EscapeStringStyle::Regex => ffi::escape_string_style_t::STRING_STYLE_REGEX,
     };
 
-    ffi::escape_string(s.as_ptr(), flags_int.into(), style).from_ffi()
+    ffi::escape_string_len(s.as_ptr(), s.len(), flags_int.into(), style).from_ffi()
```
Xiretza, line 92: Ouch, that one's on me, but it should already be fixed on master: 7bab4c4ddabd1ce2c4dace55f7215e47e9665736

fish-rust/src/trace.rs
```diff
@@ -0,0 +1,96 @@
...
+fn trace_enabled(parser: &parser_t) -> bool {
+    let ld = parser.ffi_libdata_pod_const();
+    if ld.suppress_fish_trace {
+        return false;
+    }
+    return do_trace.load(atomic::Ordering::Relaxed);
+}
+
+/// Trace an "argv": a list of arguments where the first is the command.
+fn trace_argv(parser: &parser_t, command: wcharz_t, argv: *const CxxWString, argc: usize) {
```
ridiculousfish, line 49: The types `wcharz_t` and `CxxWString` should be confined to the FFI; we don't want these to leak into non-FFI functions because it will make them difficult to call from Rust.

Please factor this into two functions:

1. `trace_argv` as you've written, except it takes `command: &wstr` and `argv: &[&wstr]`
2. A bridge function which takes FFI types, and calls `trace_argv`.

See the functions `run_builtin ` (non-FFI) and `rust_run_builtin` (FFI) for an example of how to do this.

The bridge function should probably take `command: &wcharz_t` and `argv: &Vec<wcharz_t>`; see `rust_run_builtin` for an example of how to pass these.

ridiculousfish (changes requested):

fish-rust/src/trace.rs
clemenswasser: Makes sense 👍

fish-rust/src/trace.rs
```diff
@@ -0,0 +1,81 @@
...
+        include!("parser.h");
+        type wcharz_t = super::wcharz_t;
+        type parser_t = super::parser_t;
+    }
+
+    extern "Rust" {
+        fn trace_set_enabled(do_enable: bool);
+        fn trace_enabled(parser: &parser_t) -> bool;
+        fn rust_trace_argv(parser: &parser_t, command: wcharz_t, args: &Vec<wcharz_t>);
+        fn trace_if_enabled(parser: &parser_t, command: wcharz_t);
```
ridiculousfish, line 22: Let's name this `rust_trace_if_enabled`. The convention is that Rust functions called from C++ should start with `rust_`

src/exec.cpp
```diff
@@ -828,10 +829,14 @@ static launch_result_t exec_process_in_job(parser_t &parser, process_t *p,
 
     // Maybe trace this process.
     // TODO: 'and' and 'or' will not show.
-    if (trace_enabled(parser)) {
-        trace_argv(parser, nullptr, p->argv());
+
+    ::rust::Vec<wcharz_t> rust_argv;
```
ridiculousfish, line 833: This pattern is duplicated three times; can we add a C++ wrapper here? Also the wrapper could check if tracing is enabled before allocating this new vactor.

fish-rust/src/ffi.rs
```diff
@@ -136,6 +140,9 @@ impl From<wcharz_t> for &wchar::wstr {
 /// Allow wcharz_t to be "into" WString.
 impl From<wcharz_t> for wchar::WString {
     fn from(w: wcharz_t) -> Self {
+        if w.str_.is_null() {
```
ridiculousfish, line 143: We shouldn't silently convert null strings to empty; I've added an assertion on the C++ side here.

Probably the simplest thing to do here is just pass an empty string on the C++ side, instead of null.

ridiculousfish (changes requested):

fish-rust/src/trace.rs
clemenswasser: All FFI functions are now prefixed with `rust_`

src/exec.cpp
clemenswasser: I've now introduced a `to_rust_string_vec` function and a `rust_trace_if_enabled` wrapper.

fish-rust/src/ffi.rs
clemenswasser: I'm now using an empty string.

clemenswasser:
Rebased on `master`

clemenswasser:
I don't really understand why clippy complains about the `Vec` parameter of `rust_trace_argv` when it doesn't complain about the same parameter of `rust_run_builtin`?

Xiretza:
> I don't really understand why clippy complains about the `Vec` parameter of `rust_trace_argv` when it doesn't complain about the same parameter of `rust_run_builtin`?

Apparently the explicit `for` iteration over `cpp_args` in `rust_run_builtin` is enough to confuse clippy and stop it from emitting the lint.

fish-rust/src/trace.rs
```diff
@@ -0,0 +1,75 @@
+use crate::{
+    common::{escape_string, EscapeFlags},
+    ffi::{self, parser_t, wcharz_t},
+    wchar::{self, wstr, L},
+    wchar_ffi::WCharToFFI,
+};
+use std::sync::atomic;
+
+#[cxx::bridge]
+mod ffi2 {
```
mqudsi, line 10: Is there a reason this is `ffi2` and not just `ffi` (or `trace_ffi`)?

fish-rust/src/trace.rs
```diff
@@ -0,0 +1,75 @@
...
+
+    extern "Rust" {
+        fn rust_trace_set_enabled(do_enable: bool);
+        fn rust_trace_enabled(parser: &parser_t) -> bool;
+        fn rust_trace_argv(parser: &parser_t, command: wcharz_t, args: &Vec<wcharz_t>);
+        fn rust_trace_if_enabled(parser: &parser_t, command: wcharz_t);
+    }
+}
+
+static do_trace: atomic::AtomicBool = atomic::AtomicBool::new(false);
```
mqudsi, line 26: Please use SCREAMING_SNAKE_CASE for static values per rust naming conventions.

fish-rust/src/trace.rs
```diff
@@ -0,0 +1,75 @@
...
+        include!("parser.h");
+        type wcharz_t = super::wcharz_t;
+        type parser_t = super::parser_t;
+    }
+
+    extern "Rust" {
+        fn rust_trace_set_enabled(do_enable: bool);
+        fn rust_trace_enabled(parser: &parser_t) -> bool;
+        fn rust_trace_argv(parser: &parser_t, command: wcharz_t, args: &Vec<wcharz_t>);
+        fn rust_trace_if_enabled(parser: &parser_t, command: wcharz_t);
```
mqudsi, line 22: If you've removed the existing/old `trace_xxx()` functions (and I see that you have by deleting `trace.cpp` and `trace.h`) there's no reason to name these `rust_trace_xxx()`. 

In this file, use our existing convention of suffixing the ffi functions with `_ffi` instead of prefixing them with `rust_` then in the `extern "Rust" { ... }` section prefix each `trace_xxx_ffi()` declaration with `[cxx_name = "trace_xxx"]`.

We'll know that any fns with `_ffi` in the name should be deleted when we're done with the transition, and by renaming the type with `cxx_name` the old name will still be used in the cpp codebase, minimizing the diff.

fish-rust/src/trace.rs
```diff
@@ -0,0 +1,75 @@
...
+fn rust_trace_enabled(parser: &parser_t) -> bool {
+    let ld = parser.ffi_libdata_pod_const();
+    if ld.suppress_fish_trace {
+        return false;
+    }
+    return do_trace.load(atomic::Ordering::Relaxed);
+}
+
+/// Trace an "argv": a list of arguments where the first is the command.
+/// Allow the `&Vec` parameter as this function only exists temporarily for the FFI
```
mqudsi, line 41: Nitpick: The second comment should be a regular `//` and not a doc comment (`///`).

src/parse_execution.cpp
```diff
@@ -297,7 +297,7 @@ end_execution_reason_t parse_execution_context_t::run_if_statement(
     size_t next_elseif_idx = 0;
 
     // We start with the 'if'.
-    trace_if_enabled(*parser, L"if");
+    rust_trace_if_enabled(*parser, L"if");
```
mqudsi, line 300: With the above suggestion, all the function call changes in this file will be avoided.

mqudsi (changes requested):

fish-rust/src/trace.rs
clemenswasser: I'm importing ffi above. Changing this to the more meaningful `trace_ffi`, thanks! 😄

fish-rust/src/trace.rs
clemenswasser: Sure, wondering why `clippy` or `rustc` don't complain...

fish-rust/src/trace.rs
clemenswasser: Done.

fish-rust/src/trace.rs
clemenswasser: Makes sense 👍

clemenswasser:
I've now rebased on current `master`

fish-rust/build.rs
```diff
@@ -32,6 +32,7 @@ fn main() -> miette::Result<()> {
         "src/redirection.rs",
         "src/smoke.rs",
         "src/timer.rs",
+        "src/trace.rs",
         "src/tokenizer.rs",
```
krobelus, line 36: should be sorted alphabetically

fish-rust/src/trace.rs
```diff
@@ -0,0 +1,79 @@
...
+        #[cxx_name = "trace_enabled"]
+        fn trace_enabled_ffi(parser: &parser_t) -> bool;
+        #[cxx_name = "trace_argv"]
+        fn trace_argv_ffi(parser: &parser_t, command: wcharz_t, args: &Vec<wcharz_t>);
+        #[cxx_name = "trace_if_enabled"]
+        fn trace_if_enabled_ffi(parser: &parser_t, command: wcharz_t);
+    }
+}
+
+static DO_TRACE: atomic::AtomicBool = atomic::AtomicBool::new(false);
```
krobelus, line 30: I'd prefer importing `AtomicBool` and `Ordering` directly
Actually I just pushed a `RelaxedAtomicBool`, let's use that instead (you'll need to add `global_safety` to mod.rs, sorry)

fish-rust/src/trace.rs
```diff
@@ -0,0 +1,79 @@
...
+
+fn trace_set_enabled_ffi(do_enable: bool) {
+    DO_TRACE.store(do_enable, atomic::Ordering::Relaxed);
+}
+
+fn trace_enabled_ffi(parser: &parser_t) -> bool {
```
krobelus, line 36: should preserve the doc comment from the header file

krobelus, line 36: I think there is no reason to add the `_ffi` suffix here because
I don't think we'll ever need two versions of this function.
Today all callers pass `parser_t` no matter if C++ or Rust.
Once we port `parser_t` to `Parser` we'll switch over all callers, including the C++ ones.

Same for `trace_set_enabled`

fish-rust/src/trace.rs
```diff
@@ -0,0 +1,79 @@
...
+    let ld = parser.ffi_libdata_pod_const();
+    if ld.suppress_fish_trace {
+        return false;
+    }
+    return DO_TRACE.load(atomic::Ordering::Relaxed);
```
krobelus, line 41: nit: I'd remove the `return`

fish-rust/src/trace.rs
```diff
@@ -0,0 +1,79 @@
...
+
+    if !command.is_empty() {
+        trace_text.push(' ');
+        trace_text.push_utfstr(command);
+    }
+    for arg in args {
+        trace_text.push(' ');
+        trace_text.push_utfstr(&escape_string(
+            arg,
+            crate::common::EscapeStringStyle::Script(EscapeFlags::default()),
```
krobelus, line 68: with latest master we can call `EscapeStringStyle::default()`

fish-rust/src/trace.rs
```diff
@@ -0,0 +1,79 @@
...
+        #[cxx_name = "trace_argv"]
+        fn trace_argv_ffi(parser: &parser_t, command: wcharz_t, args: &Vec<wcharz_t>);
+        #[cxx_name = "trace_if_enabled"]
+        fn trace_if_enabled_ffi(parser: &parser_t, command: wcharz_t);
+    }
+}
+
+static DO_TRACE: atomic::AtomicBool = atomic::AtomicBool::new(false);
+
+fn trace_set_enabled_ffi(do_enable: bool) {
```
krobelus, line 32: please make the non-FFI functions `pub`. As a general rule, everything that was defined in the header file should be exported because Rust code will eventually need it.

fish-rust/src/trace.rs
```diff
@@ -0,0 +1,79 @@
...
+            crate::common::EscapeStringStyle::Script(EscapeFlags::default()),
+        ));
+    }
+    trace_text.push('\n');
+    ffi::log_extra_to_flog_file(&trace_text.to_ffi());
+}
+
+fn trace_if_enabled_ffi(parser: &parser_t, command: wcharz_t) {
+    if trace_enabled_ffi(parser) {
+        trace_argv_ffi(parser, command, &Vec::new());
```
krobelus, line 77: You also define this function in C++ which means this one is unused. We should naturalize it to accomodate upcoming Rust callers
1. make it `pub`
2. drop `_ffi` (and drop it from the bridge)
3. call `trace_argv` instead of `trace_argv_ffi`
4. add the `argv` argument instead of always passing an empty slice. Rust does not have default arguments so we have to include it always

src/ffi.h
```diff
@@ -13,3 +14,18 @@ inline std::shared_ptr<T> box_to_shared_ptr(rust::Box<T> &&value) {
...
+inline static rust::Vec<wcharz_t> to_rust_string_vec(const wcstring_list_t &strings) {
+    rust::Vec<wcharz_t> rust_strings;
+    rust_strings.reserve(strings.size());
+    for (const wcstring &string : strings) {
+        rust_strings.emplace_back(string.c_str());
+    }
+    return rust_strings;
+}
+
+inline static void trace_if_enabled(const parser_t & parser, wcharz_t command, const wcstring_list_t& args) {
```
krobelus, line 27: nit: should be formatted with `clang-format`

fish-rust/build.rs
clemenswasser: Sorry about that. Fixed

fish-rust/src/trace.rs
clemenswasser: I've now used `RelaxedAtomicBool`

fish-rust/src/trace.rs
clemenswasser: I've re-added all doc comments

clemenswasser: Applied for both

fish-rust/src/trace.rs
clemenswasser: Thanks. Wondering why clippy didn't complain, as there is a lint for this: https://rust-lang.github.io/rust-clippy/master/index.html#needless_return

fish-rust/src/trace.rs
clemenswasser: Done and added a use statement as this was a pretty long path

fish-rust/src/trace.rs
clemenswasser: Marked them as pub

fish-rust/src/trace.rs
clemenswasser: Done

src/ffi.h
clemenswasser: Thanks

clemenswasser:
Sorry, forgot to amend my commit. Should all be fine now.

ridiculousfish:
This is awesome. Thank you!

fish-rust/src/trace.rs
krobelus: currently some clippy lints are disabled at crate-level (in lib.rs), we can turn them on later

PR #9630, opened 2023-03-02
Move escape_string tests to Rust

Xiretza:
This way, both the Rust FFI wrapper and the actual C++ implementation are tested.

Should've done that in #9588 already, and I would've noticed 7bab4c4ddabd1ce2c4dace55f7215e47e9665736 before it landed in master.


ridiculousfish:
Merged as 8427e05bf79d61044604ab9122eb8dbd20f8325e, thank you!

PR #9637, opened 2023-03-05
use default trait for enum

wuyudi:
the default for enum and some other types.

## Description

Use builtin default to shorten the code.

> Note that shared enums automatically always come with impls of Copy, Clone, Eq, and PartialEq, so you're free to omit those derives on an enum.

https://cxx.rs/shared.html

Fixes issue #

## TODOs:

### DIFF
--- a/fish-rust/src/parse_constants.rs
+++ b/fish-rust/src/parse_constants.rs
@@ -90,8 +90,9 @@ mod parse_constants_ffi {
 
     /// IMPORTANT: If the following enum table is modified you must also update token_type_description below.
     /// TODO above comment can be removed when we drop the FFI and get real enums.
-    #[derive(Clone, Copy)]
+    #[derive(Clone, Copy, Default)]
     enum ParseTokenType {
+        #[default]
         invalid = 1,
 
         // Terminal types.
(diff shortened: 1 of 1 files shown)


fish-rust/src/parse_constants.rs
```diff
@@ -90,7 +90,6 @@ mod parse_constants_ffi {
 
     /// IMPORTANT: If the following enum table is modified you must also update token_type_description below.
     /// TODO above comment can be removed when we drop the FFI and get real enums.
-    #[derive(Clone, Copy)]
```
Xiretza, line 93: I think leaving these here for clarity, even if unnecessary, doesn't hurt.

mqudsi: `Clone` and `Copy` are useful to define because you never know when you might need them later on and they make it clear that the developer considers a clone to be a low-cost operation (without having to review the source). 

`Debug` and `Default` are iffier since (in large quantities) they can slow down the compiler if you don't need them, but it's a real chore to go back and add them everywhere as well, so I'm still undecided on whether or not it makes sense to sprinkle them everywhere liberally.

wuyudi: > I think leaving these here for clarity, even if unnecessary, doesn't hurt.

then maybe add all the `Copy, Clone, Eq, and PartialEq` since they are there?

mqudsi:
Implemented in 14d6b1c3dea59b8a8c3836f4d270f1d5c33c57db

PR #9641, opened 2023-03-05
RIIR event.cpp

Xiretza:
Reopen of #9563 (closed because target branch was deleted). Still doesn't build because autocxx doesn't grok that `Event` is a Rust type that it should know about (even with `unsafe impl ExternType for Event`). I'm honestly quite sick of fighting autocxx, so I'll mostly leave this here as a starting point for someone who feels like doing it.


krobelus:
To pass objects like `Event` from Rust to C++ we should use `const void*` which we can cast to a `const Event &event`.
If we need a `Box<Event>` then we should call `event.clone()`.
It's not trivial to transfer ownership of an object from C++ to Rust but we don't have to. It's fine to copy.

This diff compiles. Next we need to pass `JobGroup` in a similar way. The existing C++ logic expects a shared pointer which we can't construct from Rust but we can use `box_to_shared_ptr(job_group.clone())` on the C++ side.
Great progres so far. I'm happy to take over but I don't know when I can get to it.

<details>

```diff
diff --git a/fish-rust/src/event.rs b/fish-rust/src/event.rs
index ad96851c3..2fb40a3e4 100644
--- a/fish-rust/src/event.rs
+++ b/fish-rust/src/event.rs
@@ -9,7 +9,7 @@
 #![allow(unused_variables)]
 
 use autocxx::WithinUniquePtr;
-use cxx::{ExternType, type_id};
+use cxx::{type_id, ExternType};
 use cxx::{CxxVector, CxxWString, SharedPtr, UniquePtr};
 use libc::{pid_t, SIGWINCH};
 use std::{
@@ -28,6 +28,7 @@ use crate::ffi::{
     termsize_container_t, Repin,
 };
 use crate::flog::FLOG;
+use crate::job_group::JobGroup;
 use crate::signal::{sig2wcs, signal_get_desc};
 use crate::wchar::{wstr, WString, L};
 use crate::wchar_ffi::{wcharz_t, WCharFromFFI, WCharToFFI};
@@ -73,6 +74,8 @@ mod event_ffi {
         fn new_event_process_exit(pid: i32, status: i32) -> Box<Event>;
         fn new_event_job_exit(pgid: i32, jid: u64) -> Box<Event>;
         fn new_event_caller_exit(internal_job_id: u64, job_id: i32) -> Box<Event>;
+        #[cxx_name = "clone"]
+        fn clone_ffi(self: &Event) -> Box<Event>;
 
         #[cxx_name = "event_add_handler"]
         fn event_add_handler_ffi(desc: &event_description_t, name: &CxxWString);
@@ -489,6 +492,12 @@ fn new_event_caller_exit(internal_job_id: u64, job_id: i32) -> Box<Event> {
     Box::new(Event::caller_exit(internal_job_id, job_id))
 }
 
+impl Event {
+    fn clone_ffi(&self) -> Box<Event> {
+        Box::new(self.clone())
+    }
+}
+
 fn event_add_handler_ffi(desc: &event_description_t, name: &CxxWString) {
     add_handler(Arc::new(EventHandler::new(
         desc.into(),
@@ -694,15 +703,16 @@ fn event_get_function_handler_descs_ffi(name: &CxxWString) -> Vec<event_descript
 /// event handler, we make sure to optimize the 'no matches' path. This means that nothing is
 /// allocated/initialized unless needed.
 fn fire_internal(parser: &mut parser_t, event: &Event) {
-    let ld = parser.libdata_pod();
-    assert!(ld.is_event >= 0, "is_event should not be negative");
+    assert!(
+        parser.libdata_pod().is_event >= 0,
+        "is_event should not be negative"
+    );
 
-    let _inc_event = {
-        let incremented = ld.is_event + 1;
-        ScopedPush::new(&mut ld.is_event, incremented)
-    };
+    let saved_is_event = parser.libdata_pod().is_event;
+    parser.libdata_pod().is_event += 1;
     // Suppress fish_trace during events.
-    let _suppress_trace = ScopedPush::new(&mut ld.suppress_fish_trace, true);
+    let saved_suppress_fish_trace = parser.libdata_pod().suppress_fish_trace;
+    parser.libdata_pod().suppress_fish_trace = true;
 
     // Capture the event handlers that match this event.
     let fire: Vec<_> = s_event_handlers
@@ -734,7 +744,8 @@ fn fire_internal(parser: &mut parser_t, event: &Event) {
 
         // Event handlers are not part of the main flow of code, so they are marked as
         // non-interactive.
-        let _interactive = ScopedPush::new(&mut ld.is_interactive, false);
+        let saved_is_interactive = parser.libdata_pod().is_interactive;
+        parser.libdata_pod().is_interactive = false;
         let prev_statuses = parser.get_last_statuses().within_unique_ptr();
 
         FLOG!(
@@ -746,24 +757,29 @@ fn fire_internal(parser: &mut parser_t, event: &Event) {
             "'"
         );
 
-        let parser = parser.pin();
-        let b = parser.push_block(block_t::event_block(Box::new(event)).within_unique_ptr());
-        parser.eval(
+        let b = parser
+            .pin()
+            .push_block(block_t::event_block((event as *const Event).cast()).within_unique_ptr());
+        parser.pin().eval_string(
             &buffer.to_ffi(),
             &io_chain_t::new().within_unique_ptr(),
-            &SharedPtr::null(),
+            // TODO job group
             block_type_t::top,
         );
-        parser.pop_block(b);
-        parser.set_last_statuses(prev_statuses);
+        parser.pin().pop_block(b);
+        parser.pin().set_last_statuses(prev_statuses);
 
         handler.fired.store(true, Ordering::Relaxed);
         fired_one_shot |= handler.is_one_shot();
+        parser.libdata_pod().is_interactive = saved_is_interactive;
     }
 
     if fired_one_shot {
         remove_handlers_if(|h| h.fired.load(Ordering::Relaxed) && h.is_one_shot());
     }
+
+    parser.libdata_pod().suppress_fish_trace = saved_suppress_fish_trace;
+    parser.libdata_pod().is_event = saved_is_event;
 }
 
 /// Fire all delayed events attached to the given parser.
diff --git a/fish-rust/src/ffi.rs b/fish-rust/src/ffi.rs
index 29df622f6..9c340954b 100644
--- a/fish-rust/src/ffi.rs
+++ b/fish-rust/src/ffi.rs
@@ -11,8 +11,6 @@ use autocxx::prelude::*;
 use cxx::SharedPtr;
 use libc::pid_t;
 
-use crate::event::Event;
-
 // autocxx has been hacked up to know about this.
 pub type wchar_t = u32;
 
diff --git a/src/env.cpp b/src/env.cpp
index 739cb8289..e3aad73c9 100644
--- a/src/env.cpp
+++ b/src/env.cpp
@@ -1479,6 +1479,8 @@ const std::shared_ptr<env_stack_t> &env_stack_t::principal_ref() {
 
 env_stack_t::~env_stack_t() = default;
 
+env_stack_t::env_stack_t(env_stack_t && ) = default;
+
 #if defined(__APPLE__) || defined(__CYGWIN__)
 static int check_runtime_path(const char *path) {
     UNUSED(path);
diff --git a/src/env.h b/src/env.h
index eea1558a9..7d9b1efd1 100644
--- a/src/env.h
+++ b/src/env.h
@@ -214,7 +214,7 @@ class env_stack_t final : public environment_t {
     friend class parser_t;
 
     /// The implementation. Do not access this directly.
-    const std::unique_ptr<env_stack_impl_t> impl_;
+    std::unique_ptr<env_stack_impl_t> impl_;
 
     /// All environment stacks are guarded by a global lock.
     acquired_lock<env_stack_impl_t> acquire_impl();
diff --git a/src/event.h b/src/event.h
index 21e29ed08..b3b34d571 100644
--- a/src/event.h
+++ b/src/event.h
@@ -14,6 +14,7 @@
 
 #if INCLUDE_RUST_HEADERS
 
+struct parser_t;
 #include "event.rs.h"
 
 #else
diff --git a/src/parse_execution.cpp b/src/parse_execution.cpp
index a20f08b29..e67d2031e 100644
--- a/src/parse_execution.cpp
+++ b/src/parse_execution.cpp
@@ -788,7 +788,7 @@ end_execution_reason_t parse_execution_context_t::handle_command_not_found(
         auto prev_statuses = parser->get_last_statuses();
 
         auto event = new_event_generic(L"fish_command_not_found");
-        block_t *b = parser->push_block(block_t::event_block(std::move(event)));
+        block_t *b = parser->push_block(block_t::event_block(&*event));
         parser->eval(buffer, io);
         parser->pop_block(b);
         parser->set_last_statuses(std::move(prev_statuses));
diff --git a/src/parser.cpp b/src/parser.cpp
index d59b71dd1..20676627e 100644
--- a/src/parser.cpp
+++ b/src/parser.cpp
@@ -778,9 +778,11 @@ bool block_t::is_function_call() const {
 
 block_t block_t::if_block() { return block_t(block_type_t::if_block); }
 
-block_t block_t::event_block(rust::Box<Event> evt) {
+block_t block_t::event_block(const void *evt_) {
+    const auto &evt = *static_cast<const Event *>(evt_);
     block_t b{block_type_t::event};
-    b.event = std::make_shared<rust::Box<Event>>(std::move(evt));
+    b.event =
+        std::make_shared<rust::Box<Event>>(evt.clone());  // TODO Post-FFI: move instead of clone.
     return b;
 }
 
diff --git a/src/parser.h b/src/parser.h
index b22597949..f42633ab9 100644
--- a/src/parser.h
+++ b/src/parser.h
@@ -13,9 +13,9 @@
 #include <vector>
 
 #include "common.h"
-#include "event.h"
 #include "cxx.h"
 #include "env.h"
+#include "event.h"
 #include "expand.h"
 #include "maybe.h"
 #include "operation_context.h"
@@ -29,6 +29,7 @@ class autoclose_fd_t;
 class io_chain_t;
 struct Event;
 struct job_group_t;
+class parser_t;
 
 /// Types of blocks.
 enum class block_type_t : uint8_t {
@@ -101,7 +102,7 @@ class block_t {
 
     /// Entry points for creating blocks.
     static block_t if_block();
-    static block_t event_block(rust::Box<Event> evt);
+    static block_t event_block(const void *evt_);
     static block_t function_block(wcstring name, wcstring_list_t args, bool shadows);
     static block_t source_block(filename_ref_t src);
     static block_t for_block();
@@ -332,11 +333,24 @@ class parser_t : public std::enable_shared_from_this<parser_t> {
                     const job_group_ref_t &job_group = {},
                     block_type_t block_type = block_type_t::top);
 
+    eval_res_t eval_string(const wcstring &cmd, const io_chain_t &io,
+                    // const void* job_group_,
+                    block_type_t block_type) {
+            // const job_group_t& job_group = *static_cast<const job_group_t*>(job_group_);
+            return eval(cmd, io, {} /*TODO job_group*/, block_type);
+    }
+
     /// Evaluate the parsed source ps.
     /// Because the source has been parsed, a syntax error is impossible.
     eval_res_t eval(const parsed_source_ref_t &ps, const io_chain_t &io,
                     const job_group_ref_t &job_group = {},
                     block_type_t block_type = block_type_t::top);
+    eval_res_t eval_parsed_source(const parsed_source_ref_t &ps, const io_chain_t &io,
+                    // const void* job_group_,
+                    block_type_t block_type) {
+            // const job_group_t& job_group = *static_cast<const job_group_t*>(job_group_);
+            return eval(ps, io, {} /*TODO job_group*/, block_type);
+    }
 
     /// Evaluates a node.
     /// The node type must be ast_t::statement_t or ast::job_list_t.
diff --git a/src/proc.cpp b/src/proc.cpp
index 03a0ad6e2..18976e200 100644
--- a/src/proc.cpp
+++ b/src/proc.cpp
@@ -542,7 +542,7 @@ bool job_or_proc_wants_summary(const shared_ptr<job_t> &j) {
 /// Invoke the fish_job_summary function by executing the given command.
 static void call_job_summary(parser_t &parser, const wcstring &cmd) {
     auto event = new_event_generic(L"fish_job_summary");
-    block_t *b = parser.push_block(block_t::event_block(std::move(event)));
+    block_t *b = parser.push_block(block_t::event_block(&*event));
     auto saved_status = parser.get_last_statuses();
     parser.eval(cmd, io_chain_t());
     parser.set_last_statuses(saved_status);
```

</details>

mqudsi:
I should be able to tackle this as I was looking into what to port next. Thanks for the work you've already done (and @krobelus for the patch above).

mqudsi:
Getting the job group reference isn't as simple since we can't clone a job group (it has a destructor that reclaims global state). 

We will need to retrieve the original `shared_ptr<job_group_t>` that we sent to the C++ code and pass that along (because C++ clones/increments that shared_ptr a few times downstream of this call).

mqudsi:
Ok, I've got it working. There were some issues in the implementation that resulted in subtle breakage but that's all been fixed.

I'm not going to merge this yet, though. The code currently is heavy in the hot path when checking for fired events (in the common case where no events have been fired).

mqudsi:
In case GH didn't generate a notification when I clicked the button, FYI the PR should be ready to review if anyone has any feedback.

krobelus:
no notification for me. I'll read through it tomorrow

fish-rust/src/wchar_ffi.rs
```diff
@@ -153,3 +153,20 @@ impl WCharFromFFI<Vec<u8>> for cxx::SharedPtr<cxx::CxxString> {
...
+
+/// Convert from FFI types to a reference to a wide string (i.e. a [`wstr`]) without allocating.
+pub trait AsWstr<'a> {
+    fn as_wstr(&'a self) -> &'a wstr;
+}
+
+impl<'a> AsWstr<'a> for cxx::UniquePtr<cxx::CxxWString> {
+    fn as_wstr(&'a self) -> &'a wstr {
+        wstr::from_char_slice(self.as_chars())
+    }
```
krobelus, line 165: nice, I could have used that a bunch of times.
FWIW the lifetime annotation is redundant I think.

fish-rust/src/event.rs
```diff
@@ -0,0 +1,926 @@
...
+        include!("wutil.h");
+        include!("parser.h");
+        include!("io.h");
+        type wcharz_t = crate::ffi::wcharz_t;
+        type parser_t = crate::ffi::parser_t;
+        type io_streams_t = crate::ffi::io_streams_t;
+    }
+
+    enum event_type_t {
+        any,
```
krobelus, line 39: duplicating the enum seems great because it makes it easy to get the post-FFI cleanup step right

fish-rust/src/event.rs
```diff
@@ -0,0 +1,926 @@
...
+            } else {
+                format!("exit handler for job with pid {pid}")
+            }
+        }
+        EventType::CallerExit { .. } => "exit handler for command substitution caller".to_string(),
+        EventType::Generic { param } => format!("handler for generic event '{param}'"),
+        EventType::Any => unreachable!(),
+    };
+
+    WString::from_str(&s)
```
krobelus, line 605: it would be more orthodox to use `sprintf` instead of `format!`+`WString::from_str`

fish-rust/src/event.rs
```diff
@@ -0,0 +1,926 @@
...
+        }
+
+        // Event handlers are not part of the main flow of code, so they are marked as
+        // non-interactive.
+        let saved_is_interactive = parser.libdata_pod().is_interactive;
+        parser.libdata_pod().is_interactive = false;
+        let prev_statuses = parser.get_last_statuses().within_unique_ptr();
+
+        FLOG!(
+            event,
```
krobelus, line 728: maybe we should add an incomplete definition of `FLOGF` and start using that already, fixing the definition later

src/parser.h
```diff
@@ -334,6 +333,9 @@ class parser_t : public std::enable_shared_from_this<parser_t> {
                     const job_group_ref_t &job_group = {},
                     block_type_t block_type = block_type_t::top);
 
+    /// An ffi overload of `eval(const wcstring &cmd, ...)` but without the extra parameters.
+    eval_res_t eval_string_ffi1(const wcstring &cmd);
```
krobelus, line 337: I think this one doesn't need the `1` suffix.
The other overload can be called `eval_parsed_source` or similar.

fish-rust/src/bitset.rs
```diff
@@ -0,0 +1,275 @@
...
+            + BitOr<T, Output = T>
+            + BitAndAssign<T>
+            + BitOrAssign<T>
+            + Shl<usize, Output = T>
+            + Shr<usize, Output = T>
+            + Not<Output = T>
+            + AsPrimitive<u8>
+            + FromPrimitive
+            + UpperBounded
+            + Num
```
krobelus, line 18: Is there no Integral trait that implies the bit operations?

fish-rust/src/bitset.rs
```diff
@@ -0,0 +1,275 @@
...
+    pub fn set(&mut self, i: usize) {
+        self.0 |= T::one() << i;
+    }
+
+    pub fn unset(&mut self, i: usize) {
+        self.0 &= !(T::one() << i);
+    }
+
+    pub fn toggle(&mut self, i: usize, v: bool) {
+        // if v { self.set(i) } else { self.unset(i) }
```
krobelus, line 34: leftover comment

fish-rust/src/bitset.rs
```diff
@@ -0,0 +1,275 @@
...
+        let v = T::from_u8(v as u8).unwrap();
+        let mask = T::one() << i;
+        let bit: T = ((-v.as_i64()).wrapping_shr(63)).as_();
+        self.0 = (self.0 & !mask) | (bit & mask);
```
krobelus, line 38: I think if we use xor, we don't need `bit`

	self.0 ^= T::one() << i

(untested)

krobelus:
LGTM, great work

src/parser.h
mqudsi: I was thinking more for the future when we add another overload for this one that has the job_group parameter.

fish-rust/src/bitset.rs
mqudsi: Unfortunately not. I think [PrimInt](https://docs.rs/num-traits/latest/num_traits/int/trait.PrimInt.html) is the closest you get.

fish-rust/src/bitset.rs
mqudsi: It's intentional to clarify the logical equivalent. Maybe I should another line saying that.

Xiretza: Could put it in a doc comment instead, maybe even with an example (which becomes a doctest)

fish-rust/src/event.rs
mqudsi: Probably. But also more work if we ever want to leave WString behind. I don't think it matters either way though.

fish-rust/src/wchar_ffi.rs
mqudsi: Yeah, I think the lifetime depends on what version of rustc you're targeting. There's an argument to be made that the excessive from_ffi() to_ffi() calls don't matter since it's just during the transition period, but I still felt ugly enough using them that I thought it was worth my time to add the trait.

fish-rust/src/bitset.rs
mqudsi: That would unconditionally flip the ith bit instead of conditionally either setting it or clearing it regardless of what the old value was, no?

fish-rust/src/bitset.rs
mqudsi: Good idea. doctests aren't working atm (I think because they're all private modules).

fish-rust/src/bitset.rs
```diff
@@ -0,0 +1,275 @@
...
+
+    pub fn unset(&mut self, i: usize) {
+        self.0 &= !(T::one() << i);
+    }
+
+    pub fn toggle(&mut self, i: usize, v: bool) {
+        // if v { self.set(i) } else { self.unset(i) }
+        let v = T::from_u8(v as u8).unwrap();
+        let mask = T::one() << i;
+        let bit: T = ((-v.as_i64()).wrapping_shr(63)).as_();
```
Xiretza, line 37: Why is this shift necessary? If I got this right, `v` is either `0_T` or `1_T` here, which is then converted to either `0_i64` or `-1_i64` - shifting either of those right by 63 is a no-op. Wouldn't `(-i64::from(v)).as_()` (without L35 above) be the same thing?

mqudsi: You're right - great catch. I was originally trying another approach and confused myself into making this more complicated than it needed to be.

fish-rust/src/event.rs
krobelus: OK. It would be nice to get rid of printf-style format specifiers eventually, maybe before we reitre WString. We could write a `wformat!`.

ridiculousfish:
bleh I had some review feedback I didn't get in in time.  I'll incorporate it.

mqudsi:
@ridiculousfish I think you can just request a review from yourself if you have some comments but not the time. I saw the changes you committed - can't say I didn't wince at the BitSet drop but fyi I'm planning on - at some point - new-typing `Signal`, hopefully as an enum, to clean up all the various usages of it and centralize all the associated functionality such as getting the string representation, the code, etc.

PR #9642, opened 2023-03-06
builtins: Port `realpath` to Rust

Tyrubias:
Rewrite `realpath` in Rust. I had to port a number of helper functions too. The code is a little clunkier than I'd like since `WString` is missing some of the creature comforts of regular Rust and C++ strings. We use the `errno` crate (already in dependencies) to duplicate `strerror(errno)` behavior from C++.


ridiculousfish:
Thank you!

I merged this as 88e0c2137a14220109eaa3bb66c87b583baaa86d before your latest force-push (to 6ac0dc5c8752791cea8bc5d96b274586059cf4c2), let me know of any changes that were missed.

fish-rust/src/expand.rs
```diff
@@ -0,0 +1,39 @@
+use crate::wchar::{EXPAND_RESERVED_BASE, EXPAND_RESERVED_END};
+
+/// Private use area characters used in expansions
+#[repr(u32)]
+pub enum ExpandChars {
+    /// Character representing a home directory.
+    HomeDirectory = EXPAND_RESERVED_BASE as u32,
+    /// Character representing process expansion for %self.
+    ProcessExpandSelf,
```
krobelus, line 9: Gonna use `pub const char` instead of the enum, I think that matches the use better

krobelus: just saw there is a From implementation for `char` which helps.. still, these are better of as char constants, the enum doesn't give us anything (since it's not exhaustive for the switches we use)

PR #9643, opened 2023-03-06
Enable rust address sanitizer for asan ci job

mqudsi:
Having automated asan for the rust part too would help catch some porting issues. It would have saved me some time because of a dumb mistake in fd_monitor when I was hacking on it and I suspect there’s a memory misuse in the event port. 

Rust has multiple sanitizers available (with llvm integration). -Zsanitizer=address catches the most likely culprits but we may want to set up a separate job w/ -Zsanitizer=memory to catch uninitialized reads.

It might be necessary to execute `cargo build` as `cargo build -Zbuild-std` to get full coverage.

When we're linking against the hybrid C++ codebase, the sanitizer library is injected into the binary by also include `-fsanitize=address` in CXXFLAGS - we do *not* want to manually opt-into `-lasan`. We also need to manually specify the desired target triple as a CMake variable and then explicitly pass it to all `cargo` invocations if building with ASAN.

Corrosion has been patched to make sure it follows these rules.

The `cargo-test` target is failing to link under ASAN. For some reason it has autocxx/ffi dependencies even though only rust-native, ffi-free code should be tested (and one would think the situation wouldn't change depending on the presence of the sanitizer flag). It's been disabled under ASAN for now.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -73,13 +73,22 @@ jobs:
   ubuntu-asan:
 
     runs-on: ubuntu-latest
+    env:
+        # Rust has two different memory sanitizers of interest; they can't be used at the same time:
+        # * AddressSanitizer detects out-of-bound access, use-after-free, use-after-return,
+        #   use-after-scope, double-free, invalid-free, and memory leaks.
+        # * MemorySanitizer detects uninitialized reads.
+        #
+        RUSTFLAGS: "-Zsanitizer=address"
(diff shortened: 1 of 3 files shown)


mqudsi:
I’m merging this because I need it but I’m not sure which of the `#[test]` we have uses ffi in its dependencies (i.e. isn’t pure rust) but I’m also not sure why the missing autocxx symbols would only manifest with asan enabled as it shouldn’t affect it either way. (cargo-test is currently disabled under asan)

PR #9647, opened 2023-03-08
Add Zabbix completions

szuro:
## Description

Adds completions for Zabbix commands

## TODOs:


share/completions/zabbix_agentd.fish
```diff
@@ -0,0 +1,43 @@
...
+    log_level_decrease=
+
+function __fish_has_target
+    if string match -rq '=' (commandline)
+        return 0
+    end
+    return 1
+end
+
+function __fish_string_in_command -a ch
```
ridiculousfish, line 14: This should have a prefix to "namespace" it to these completions - see e.g. the `apt` completions with functions like `__fish_apt_no_subcommand`

share/completions/zabbix_proxy.fish
```diff
@@ -0,0 +1,70 @@
...
+
+function __fish_has_target
+    if string match -rq '=' (commandline)
+        return 0
+    end
+    return 1
+end
+
+function __fish_string_in_command -a ch
+    if string match -rq $ch (commandline )
```
ridiculousfish, line 20: You don't need an `if` statement here, just do it:

    function __fish_string_in_command -a ch
          string match -rq $ch (commandline)
    end

Also I suggest using globs instead of a regex, since it's not obvious to the caller that regex characters will be respected:

    function __fish_string_in_command -a ch
          string match -q "*$ch*" (commandline)
    end

share/completions/zabbix_server.fish
```diff
@@ -0,0 +1,105 @@
...
+
+# General
+complete -c zabbix_server -s c -l config -d "Path to the configuration file"
+complete -c zabbix_server -f -s f -l foreground -d "Run Zabbix server in foreground"
+complete -c zabbix_server -f -s h -l help -d "Display this help message"
+complete -c zabbix_server -f -s V -l version -d "Display version number"
+
+
+# Log levels
+complete -c zabbix_server -r -f -s R -l runtime-control -n "__fish_has_target; __fish_string_in_command log_level_increase" -a "(__fish_prepend log_level_increase)"
```
ridiculousfish, line 93: The status code of `__fish_has_target` will just be discarded here; you probably want to use a conjuction like &&.

ridiculousfish (changes requested):

share/completions/zabbix_server.fish
szuro: Figured that if it works without catching the exit cone, might as well ditch it.

share/completions/zabbix_agentd.fish
szuro: Done

szuro:
Shortened some of the descriptions

mqudsi:
@zanchey I wanted to merge this for 3.6.2 but I'm not clear on whether or not that ship has sailed?

PR #9653, opened 2023-03-12
Add a proper rust ScopeGuard

mqudsi:
Due to limitations imposed by the borrow checker, there are very few places
where we will be able to use the `ScopedPush` class ported over from the C++
codebase (once you capture the value w/ a `ScopedPush` you can't access the
value - or the mutable reference you used to reach it! - until the `ScopedPush`
object goes out of scope).

This alternative requires binding the previous values to a variable and manually
restoring them in the callback passed to the `ScopeGuard` constructor, but will
work with rust's borrow and `&mut` paradigm.

This is akin to our C++ `cleanup_t` RAII closure wrapper.

### DIFF
--- a/fish-rust/src/common.rs
+++ b/fish-rust/src/common.rs
@@ -8,6 +8,14 @@ use std::ops::{Deref, DerefMut};
 use std::os::fd::AsRawFd;
 use std::{ffi::c_uint, mem};
 
+/// Like [`std::mem::replace()`] but provides a reference to the old value in a callback to obtain
+/// the replacement value. Useful to avoid errors about multiple references (`&mut T` for `old` then
+/// `&T` again in the `new` expression).
+pub fn replace_with<T, F: FnOnce(&T) -> T>(old: &mut T, with: F) -> T {
+    let new = with(&*old);
+    std::mem::replace(old, new)
+}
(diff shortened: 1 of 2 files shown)


fish-rust/src/common.rs
```diff
@@ -3,13 +3,115 @@ use crate::wchar::{wstr, WString};
...
+
+    /// Cancel the unwind operation, e.g. do not call the previously passed-in `on_drop` callback
+    /// when the current scope expires.
+    pub fn cancel(guard: &mut Self) {
+        guard.on_drop.take();
+    }
+
+    /// Cancels the unwind operation like [`ScopeGuard::cancel()`] but also returns the captured
+    /// value (consuming the `ScopeGuard` in the process).
+    pub fn rollback(mut guard: Self) -> T {
```
Xiretza, line 67: I think this leaks `on_drop`, `let _ = self.on_drop;` should be enough to drop it explicitly.

Xiretza:
Clever! Looks good to me.

fish-rust/src/common.rs
mqudsi: Yikes, nice catch!

PR #9654, opened 2023-03-12
Detect `rust-analyzer` in build script to enable `autocxx` completions

Tyrubias:
Currently the `autocxx` generated code does not produce any code intelligence because `rust-analyzer` can't find the generated code since it's not in the workspace. Here, we detect `rust-analyzer` by checking for `RUSTC_WRAPPER=rust-analyzer` and changing the output directory accordingly.

The style of the `if`-statement I added is not super canonical Rust but I believe this is easier to read than the alternatives I considered.

### DIFF
--- a/fish-rust/build.rs
+++ b/fish-rust/build.rs
@@ -50,7 +50,9 @@ fn main() -> miette::Result<()> {
     let include_paths = [&fish_src_dir, &fish_build_dir, &cxx_include_dir];
     let mut builder = autocxx_build::Builder::new("src/ffi.rs", include_paths);
     // Detect `rust-analyzer` and output the autocxx stuff into `target` so we get code intelligence on it
-    if std::env::var("RUSTC_WRAPPER").map_or(true, |wrapper| !wrapper.contains("rust-analyzer")) {
+    if std::env::var("RUSTC_WRAPPER").map_or(true, |wrapper| {
+        !(wrapper.contains("rust-analyzer") || wrapper.contains("intellij-rust-native-helper"))
+    }) {
         // We need this reassignment because of how the builder pattern works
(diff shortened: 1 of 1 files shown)


mqudsi:
Nice find! Unfortunately I still get `&{unknown}` when I hover over `parser_t` references. What difference do you see w/ and w/out this patch?

Tyrubias:
@mqudsi I see full intelligence for things like `parser_t`. I needed to force reload `rust-analyzer` in VSCode to get it to rerun the build script. Unfortunately I don't see a way around that part.

Tyrubias:
You might want to check how your code intelligence tool (I don't know if you're using VSCode, `coc-nvim`, or some other setup) is setting `RUSTC_WRAPPER=rust-analyzer` in the `rust-analyzer` settings. I can try to help debug if that is not the case.

mqudsi:
I'm using neovim with the built-in lsp. As far as I can tell, [rust-analyzer sets RUSTC_WRAPPER="rust-analyzer" by itself](https://github.com/rust-lang/rust-analyzer/blob/f1e51afa450ecf352c1f01fc5a5b6d10792e3779/crates/rust-analyzer/src/bin/rustc_wrapper.rs#L1-L3).

Tyrubias:
@mqudsi I figured out why: `rust-analyzer` sets `RUSTC_WRAPPER` to *the path* to `rust-analyzer`, not the phrase `"rust-analyzer"`. I'm pushing a change to fix this. Unfortunately this might still be a little brittle since invocations of `cargo build` *may* interfere. There's not a editor/platform agnostic way to do this, at least not until [this `rust-analyzer` issue is resolved](https://github.com/rust-lang/rust-analyzer/issues/6007).

mqudsi:
That works much better, thanks.

PR #9656, opened 2023-03-13
Add fish_config note about snap and flatpak broswers

yarcod:
When the user uses a flatpak or snap packaged browser, the `fish_config browse` command returns a broken page.
This adds a note explaining how the user can work around this issue.

### DIFF
--- a/doc_src/cmds/fish_config.rst
+++ b/doc_src/cmds/fish_config.rst
@@ -20,6 +20,7 @@ Description
 Without arguments or with the ``browse`` command it starts the web-based configuration interface. The web interface allows you to view your functions, variables and history, and to make changes to your prompt and color configuration. It starts a local web server and opens a browser window. When you are finished, close the browser window and press the Enter key to terminate the configuration session.
 
 If the ``BROWSER`` environment variable is set, it will be used as the name of the web browser to open instead of the system default.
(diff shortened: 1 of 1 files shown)


yarcod (commented):
Another option would be for ``fish_config`` to output the http link directly, instead of going via the tmp file

faho:
Note: Current fish versions will *already* print the http link:

> Web config started at file:///tmp/web_configBLABLABLA.html

> If that doesn't work, try opening http://localhost:8000/SOMESTRING

That's both easier and less hacky than what you are proposing, so I am closing this.

yarcod:
> The reason there's no direct HTTP link is that it's potentially visible to other users in the output of `ps`, which is a security problem.

> Note: Current fish versions will _already_ print the http link:
> 
> That's both easier and less hacky than what you are proposing, so I am closing this.

Thanks for the updates! While I certainly agree that printing out the http link is the most convenient way to do it, doesn't the current printout give rise the security issue as @zanchey mentioned? Or is that only present if the application runs the browser from the http directly without ever involving /tmp?

Lastly, apologies for reporting on something which already had a solution in a newer version than I ran!

faho:
>doesn't the current printout give rise the security issue as @zanchey mentioned?

It doesn't. The disclosure happens when it's used as a commandline argument - e.g. when you run a concrete browser like `firefox https://localhost:8000/aiojsdoisa`.

Those are shown in the output of, e.g., `ps`, even when run as other users.

PR #9658, opened 2023-03-13
ScopeGuard: remove potential memory leak

Xiretza:
Calling `ScopeGuard::rollback()` would leak the `on_drop` callable; this is a problem for `Box<dyn FnOnce>` or closures containing `Drop` data. I [originally suggested](https://github.com/fish-shell/fish-shell/pull/9653/commits/992480b1a0d0f108c25038c9d3e7bb919cc09e8c#r1133321862) `let _ = self.on_drop;` as a fix for this, but that's actually a no-op; using a named binding instead shows that this can't possibly work because you can't move out of a type implementing `Drop`. `Option::take()` replaces the field with `None` instead.

### DIFF
--- a/fish-rust/src/common.rs
+++ b/fish-rust/src/common.rs
@@ -65,7 +65,7 @@ impl<T, F: FnOnce(&mut T)> ScopeGuard<T, F> {
     /// Cancels the unwind operation like [`ScopeGuard::cancel()`] but also returns the captured
     /// value (consuming the `ScopeGuard` in the process).
     pub fn rollback(mut guard: Self) -> T {
-        let _ = guard.on_drop;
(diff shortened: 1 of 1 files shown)


mqudsi:
That's funny: that's what I originally wrote because I reasoned you can't move out of a structure implementing `OnDrop`. Then I tried your suggestion just for laughs and when it worked because `Option<T>` implements `Copy` I just kept your approach!

Xiretza:
> it worked because Option<T> implements `Copy`

I don't think that's it (`Option<T>` is only `Copy` if `T` is); `_` is simply special and doesn't actually cause anything to be moved. If you use `foo` instead of `_`, it errors.

PR #9661, opened 2023-03-14
Port non-logic from signals.cpp via rusty Signal newtype

mqudsi:
A signal code has been raised to its own newtype, and the existing rust code has been updated to use `Signal` instead of various-sized integers. `sig2wcs()` has been replaced with `Signal::name()` and `signal_get_desc()` has been replaced with `Signal::desc()`.

For now, the logic that the libc crate uses to decide which signals are available (basically, the os) has been used instead of testing for per-signal availability at build-time. I think this is sufficient, but we can always switch it to build-time detection (with much awkwardness and complexity) at a later date if there are platforms we actually target which don't support the basic set of signals we enumerate.

I do want to introduce a build-time auto-detected "BSD" feature instead of using the very narrowly drilled-down `target_os = "freebsd"/whatever` syntax that rust supports oob.

The `Signal` newtype was originally an `enum` but I found it made more sense to treat it as a `struct`. The original enum code is in the commits but I'll squash it out if/when this is approved.

### DIFF
--- a/fish-rust/src/event.rs
+++ b/fish-rust/src/event.rs
@@ -16,7 +16,7 @@ use crate::builtins::shared::io_streams_t;
 use crate::common::{escape_string, replace_with, EscapeFlags, EscapeStringStyle, ScopeGuard};
 use crate::ffi::{self, block_t, parser_t, signal_check_cancel, signal_handle, Repin};
 use crate::flog::FLOG;
-use crate::signal::{sig2wcs, signal_get_desc};
+use crate::signal::Signal;
 use crate::termsize;
 use crate::wchar::{wstr, WString, L};
 use crate::wchar_ffi::{wcharz_t, AsWstr, WCharFromFFI, WCharToFFI};
@@ -107,7 +107,7 @@ pub enum EventType {
     /// well).
     Any,
     /// An event triggered by a signal.
-    Signal { signal: i32 },
+    Signal { signal: Signal },
     /// An event triggered by a variable update.
     Variable { name: WString },
     /// An event triggered by a process exit.
@@ -206,7 +206,7 @@ impl From<&event_description_t> for EventDescription {
             typ: match desc.typ {
                 event_type_t::any => EventType::Any,
                 event_type_t::signal => EventType::Signal {
-                    signal: desc.signal,
+                    signal: Signal::new(desc.signal),
                 },
                 event_type_t::variable => EventType::Variable {
                     name: desc.str_param1.from_ffi(),
@@ -243,7 +243,7 @@ impl From<&EventDescription> for event_description_t {
         };
         match desc.typ {
             EventType::Any => (),
-            EventType::Signal { signal } => result.signal = signal,
+            EventType::Signal { signal } => result.signal = signal.code(),
             EventType::Variable { .. } => (),
             EventType::ProcessExit { pid } => result.pid = pid,
             EventType::JobExit {
@@ -490,8 +490,8 @@ struct PendingSignals {
 impl PendingSignals {
     /// Mark a signal as pending. This may be called from a signal handler. We expect only one
     /// signal handler to execute at once. Also note that these may be coalesced.
-    pub fn mark(&self, which: usize) {
-        if let Some(received) = self.received.get(which) {
+    pub fn mark(&self, sig: usize) {
+        if let Some(received) = self.received.get(sig) {
             received.store(true, Ordering::Relaxed);
             self.counter.fetch_add(1, Ordering::Relaxed);
         }
@@ -551,19 +551,17 @@ static OBSERVED_SIGNALS: [AtomicU32; SIGNAL_COUNT] = [ATOMIC_U32_0; SIGNAL_COUNT
 /// temporarily moved here. There was no mutex around this in the cpp code. TODO: Move it back.
 static BLOCKED_EVENTS: Mutex<Vec<Event>> = Mutex::new(Vec::new());
 
-fn inc_signal_observed(sig: i32) {
-    if let Ok(index) = usize::try_from(sig) {
-        if let Some(sig) = OBSERVED_SIGNALS.get(index) {
-            sig.fetch_add(1, Ordering::Relaxed);
-        }
+fn inc_signal_observed(sig: Signal) {
+    let index: usize = sig.into();
+    if let Some(sig) = OBSERVED_SIGNALS.get(index) {
+        sig.fetch_add(1, Ordering::Relaxed);
     }
 }
 
-fn dec_signal_observed(sig: i32) {
-    if let Ok(index) = usize::try_from(sig) {
-        if let Some(sig) = OBSERVED_SIGNALS.get(index) {
-            sig.fetch_sub(1, Ordering::Relaxed);
-        }
+fn dec_signal_observed(sig: Signal) {
+    let index: usize = sig.into();
+    if let Some(sig) = OBSERVED_SIGNALS.get(index) {
+        sig.fetch_sub(1, Ordering::Relaxed);
     }
 }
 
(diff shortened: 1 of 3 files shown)


fish-rust/src/event.rs
```diff
@@ -817,7 +821,9 @@ fn event_fire_delayed_ffi(parser: Pin<&mut parser_t>) {
 /// Enqueue a signal event. Invoked from a signal handler.
 pub fn enqueue_signal(signal: usize) {
     // Beware, we are in a signal handler
-    PENDING_SIGNALS.mark(signal);
+    if let Some(s) = Signal::try_new(signal as i32) {
```
ridiculousfish, line 824: This is too much scary code in a signal handler. We should have `mark()` work in primitive types only.

fish-rust/src/event.rs
```diff
@@ -491,7 +498,8 @@ struct PendingSignals {
 impl PendingSignals {
     /// Mark a signal as pending. This may be called from a signal handler. We expect only one
     /// signal handler to execute at once. Also note that these may be coalesced.
-    pub fn mark(&self, which: usize) {
+    pub fn mark(&self, sig: Signal) {
```
ridiculousfish, line 501: As this is run within a signal handler, it should continue to take a `usize` to ensure it is easy to reason about.

fish-rust/src/job_group.rs
```diff
@@ -100,8 +101,8 @@ pub struct JobGroup {
     /// "Simple block" groups like function calls do not have a job id.
     pub job_id: Option<JobId>,
     /// The signal causing the group to cancel or `0` if none.
-    /// Not using an `Option<NonZeroI32>` to be able to atomically load/store to this field.
-    signal: AtomicI32,
+    /// Not using an `Option<Signal>` to be able to atomically load/store to this field.
+    signal: AtomicU8,
```
ridiculousfish, line 105: Why change from I32 to U8? I prefer I32 because signals have type int in the system interface.

fish-rust/src/job_group.rs
```diff
@@ -146,11 +147,9 @@ impl JobGroup {
...
 
     /// Gets the cancellation signal, if any.
-    pub fn get_cancel_signal(&self) -> Option<NonZeroI32> {
-        match self.signal.load(Ordering::Relaxed) {
-            0 => None,
-            s => Some(NonZeroI32::new(s).unwrap()),
-        }
+    pub fn get_cancel_signal(&self) -> Option<Signal> {
+        let s = self.signal.load(Ordering::Relaxed);
+        Signal::try_new(s as i32)
```
ridiculousfish, line 152: This makes it look like the Signal conversion may "fail." We should always return a signal if `self.signal` is nonzero.

fish-rust/src/signal.rs
```diff
@@ -65,3 +67,272 @@ pub fn signal_get_desc(sig: i32) -> &'static wstr {
 
     wstr::from_ucstr(s).expect("signal description should be valid utf-32")
 }
+
+#[derive(Clone, Copy, Eq, PartialEq, PartialOrd, Ord)]
+/// A wrapper around the system signal code.
+pub struct Signal(NonZeroU8);
```
ridiculousfish, line 73: Signals have type `int`, so I expect this to wrap an int.

fish-rust/src/signal.rs
```diff
@@ -65,3 +67,272 @@ pub fn signal_get_desc(sig: i32) -> &'static wstr {
...
+            Signal::SIGSTKFLT => "Stack fault",
+            #[cfg(target_os = "linux")]
+            Signal::SIGPWR => "Power failure",
+            Signal(_) => "Unknown",
+        }
+    }
+
+    pub const fn code(&self) -> i32 {
+        // Safety: Types share size and alignment but one just has additional semantic constraints
+        (unsafe { *(&self.0 as *const _ as *const u8) }) as i32
```
ridiculousfish, line 253: I don't understand why we need unsafe code here? I would expect to just have an int, and return it.

fish-rust/src/signal.rs
```diff
@@ -65,3 +67,272 @@ pub fn signal_get_desc(sig: i32) -> &'static wstr {
...
+    }
+
+    pub const fn new(sig: NonZeroU8) -> Self {
```
ridiculousfish, line 256: I think we should just have a single function which either accepts a NonZeroI32, or an i32 and asserts that it's non-zero. I don't see a reason to make a fallible version here.

fish-rust/src/signal.rs
```diff
@@ -65,3 +67,272 @@ pub fn signal_get_desc(sig: i32) -> &'static wstr {
...
+        Signal(sig)
+    }
+
+    pub fn try_new(sig: i32) -> Option<Self> {
```
ridiculousfish, line 260: I think we should just have a single function which either accepts a NonZeroI32, or an i32 and asserts that it's non-zero. I don't see a reason to make a fallible version here.

fish-rust/src/signal.rs
```diff
@@ -65,3 +67,272 @@ pub fn signal_get_desc(sig: i32) -> &'static wstr {
...
+        let name = self.name();
+        if name != Self::UNKNOWN_SIG_NAME {
+            f.write_str(name)
+        } else {
+            f.write_fmt(format_args!("Unrecognized Signal {}", self.code()))
+        }
+    }
+}
+
+impl From<NonZeroU8> for Signal {
```
ridiculousfish, line 285: I think we can do without these int -> Signal `From` implementations. It's fine to make Signal explicit.

fish-rust/src/event.rs
```diff
@@ -201,13 +201,15 @@ pub struct EventDescription {
     typ: EventType,
 }
 
-impl From<&event_description_t> for EventDescription {
-    fn from(desc: &event_description_t) -> Self {
-        EventDescription {
+impl TryFrom<&event_description_t> for EventDescription {
```
ridiculousfish, line 204: There's no reason to make this fallible; nobody is prepared to handle this error. If you want to add additional logic such that `--on-signal 0` reports an error to the user then that's OK, but we shouldn't complicate this interface for that silly case.

ridiculousfish (changes requested):

fish-rust/src/signal.rs
mqudsi: It's just because of the `const` fn. I can drop the `const` since I guess we will probably be able to live without it.

fish-rust/src/signal.rs
mqudsi: It was to avoid asserting/panicking but keep the option of passing in an i32. If you think it's OK to panic if we get a zero, that makes things easier.

fish-rust/src/job_group.rs
mqudsi: Just to match the Signal inner type. I originally wrapped Signal around NonZeroI32 but went with code purity instead since signal numbers just don't go that high even though they're returned as int by the system (largely because that's the smallest type the system really uses, especially since char isn't defined as being signed or unsigned). But I can change everything back to i32/NonZeroI32 to get rid of any downsizing safety checks.

fish-rust/src/signal.rs
mqudsi: I originally wrapped Signal around NonZeroI32 but went with code purity instead since signal numbers just don't go that high even though they're returned as int by the system (largely because that's the smallest type the system really uses, especially since char isn't defined as being signed or unsigned). But I can change everything back to i32/NonZeroI32 to get rid of any downsizing safety checks.

fish-rust/src/event.rs
mqudsi: That was the scenario I had in mind, yes. I figured it would be safer to handle it centrally by returning an error rather than require all callers to assert the non-zero value but since there aren't many I suppose that is easier.

mqudsi:
@ridiculousfish I've changed `Signal` to wrap a `NonZeroI32` and updated the code to remove the now-unnecessary checks wherever it was possible. I also ported `wcs2sig()` though it's not yet used in the rust code.

PR #9662, opened 2023-03-14
Edited the Readme to add Fedora and others

trytomakeyouprivate:
## Description

Fedora and probably other Distros use a different, sudo-requiring command to switch the shell. I added that command, as well as the command for switching back.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -44,7 +44,7 @@ Packages for Linux
 
 Packages for Debian, Fedora, openSUSE, and Red Hat Enterprise
 Linux/CentOS are available from the `openSUSE Build
-Service <https://software.opensuse.org/download.html?project=shells%3Afish&package=fish>`__.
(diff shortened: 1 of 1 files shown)


exploide:
> On Fedora and other RHEL-based systems, the `chsh` command is not available for security reasons.

I don't know where this is coming from. I'm using fish on Fedora for a decade and have always set my login shell via `chsh`.

```console
> cat /etc/redhat-release 
Fedora release 37 (Thirty Seven)

> which chsh
/usr/bin/chsh
```

Or is this maybe in some special system configuration? Or on rpm-ostree based systems?

trytomakeyouprivate:
Weird yes, I read that this is the case on Fedora. Maybe they reverted it back?
 
Closing this here in that case.

Changing non-root shell should also be no problem for security.

zanchey:
`util-linux` isn't in the cloud image, but I don't think that's the typical experience.

PR #9666, opened 2023-03-15
Rewrite builtin printf in Rust, and add some locale and number parsing stuff

ridiculousfish:
The big lift here is rewriting builtin_printf in Rust. This required some other pieces which came along for the ride.

1. Add some locale pieces sufficient for printf - that's basically decimal separator, thousands separators, and digit grouping. I think the usage here will be confined to builtin_printf - it doesn't leak everywhere like in C.

2. We need a way to parse floats from strings. Add the [`fast-float`](https://github.com/fastfloat/fast_float) crate which has an acceptable license. Change it to allow parsing from any char iterator, instead of just a C string; also allow the caller to specify the decimal separator.

3. We need a better underlying `printf` implementation - the one we have is too buggy. Switch to [`printf-compat`](https://github.com/lights0123/printf-compat) which is somewhat less buggy. Then try to fix all of its bugs, and teach it to do locale-savvy output.

 It's frustrating how much stuff we have to re-implement here but at least it's done.


fish-rust/src/builtins/shared.rs
```diff
@@ -45,7 +45,9 @@ pub const STATUS_CMD_OK: Option<c_int> = Some(0);
 /// The status code used for failure exit in a command (but not if the args were invalid).
 pub const STATUS_CMD_ERROR: Option<c_int> = Some(1);
 
-/// A handy return value for invalid args.
+/// The status code used for invalid arguments given to a command. This is distinct from valid
+/// arguments that might result in a command failure. An invalid args condition is something
+/// like an unrecognized flag, missing or too many arguments, an invalid integer, etc. But
```
zanchey, line 50: Is there something missing from the sentence here?

fish-rust/src/wutil/wcstod.rs
```diff
@@ -0,0 +1,444 @@
...
+///
+/// The `decimal_sep` parameter is used to specify the decimal separator.
+/// '.' is a normal default.
+///
+/// The `consumed` parameter is used to return the number of characters
+/// consumed, similar to the "end" parameter to strtod.
+/// This is only meaningful if parsing succeeds.
+///
+/// Error::Overflow is returned if the value is too large in magnitude.
+pub fn wcstod<Chars>(input: Chars, decimal_sep: char, consumed: &mut usize) -> Result<f64, Error>
```
Xiretza, line 20: `&mut` arguments are much less ergonomic than just returning multiple values - for `wcstod` there's also no risk of mixing up the two return values:

```suggestion
pub fn wcstod<Chars>(input: Chars, decimal_sep: char) -> Result<(f64, usize), Error>
```

fish-rust/src/wutil/wcstoi.rs
```diff
@@ -164,32 +230,38 @@ where
...
+///  - 0 means octal, 0x means hex
+///  - Leading + is supported.
+/// The number of consumed characters is returned in out_consumed.
+pub fn fish_wcstoi_partial<Int, Chars>(
     src: Chars,
-    radix: Option<u32>,
-    consume_all: bool,
+    options: Options,
+    out_consumed: &mut usize,
 ) -> Result<Int, Error>
```
Xiretza, line 259: Same here re: return value vs. `&mut` parameter, although there's a risk of confusing the `Int` and `usize`.

fish-rust/src/common.rs
```diff
@@ -106,6 +106,16 @@ impl<T, F: FnOnce(&mut T)> Drop for ScopeGuard<T, F> {
...
+// These are in the Unicode private-use range. We really shouldn't use this
+// range but have little choice in the matter given how our lexer/parser works.
+// We can't use non-characters for these two ranges because there are only 66 of
+// them and we need at least 256 + 64.
+//
+// Note: We don't use the highest 8 bit range (0xF800 - 0xF8FF) because we know
+// of at least one use of a codepoint in that range: the Apple symbol (0xF8FF)
+// on Mac OS X. See http://www.unicode.org/faq/private_use.html.
+pub const ENCODE_DIRECT_BASE: u32 = 0xF600;
+pub const ENCODE_DIRECT_END: u32 = ENCODE_DIRECT_BASE + 256;
```
Xiretza, line 118: Already exist in `src/wchar.rs`.

fish-rust/src/builtins/printf.rs
```diff
@@ -0,0 +1,817 @@
...
+                esc_value = esc_value * 16 + p.char_at(0).to_digit(16).unwrap();
+                esc_length += 1;
+                p = &p[1..];
+            }
+            if esc_length == 0 {
+                self.fatal_error(wgettext!("missing hexadecimal number in escape"));
+            }
+            self.append_output(
+                char::from_u32(ENCODE_DIRECT_BASE + esc_value % 256)
+                    .expect("Escape should be encodeable"),
```
Xiretza, line 636:
```suggestion
                wchar::wchar_literal_byte(esc_value % 256)
```

fish-rust/src/builtins/printf.rs
```diff
@@ -0,0 +1,817 @@
...
+                p = &p[1..];
+            }
+            while esc_length < 3 && is_octal_digit(p.char_at(0)) {
+                esc_value = esc_value * 8 + p.char_at(0).to_digit(8).unwrap();
+                esc_length += 1;
+                p = &p[1..];
+            }
+            self.append_output(
+                char::from_u32(ENCODE_DIRECT_BASE + esc_value % 256)
+                    .expect("Escape should be encodeable"),
```
Xiretza, line 653:
```suggestion
                wchar::wchar_literal_byte(esc_value % 256),
```

fish-rust/src/builtins/printf.rs
```diff
@@ -0,0 +1,817 @@
...
+        self.exit_code = STATUS_CMD_ERROR.unwrap();
+        self.early_exit = true;
+    }
+
+    /// Print a \ escape sequence starting at ESCSTART.
+    /// Return the number of characters in the string, *besides the backslash*.
+    /// That is this is ONE LESS than the number of characters consumed.
+    /// If octal_0 is nonzero, octal escapes are of the form \0ooo, where o
+    /// is an octal digit; otherwise they are of the form \ooo.
+    fn print_esc(&mut self, escstart: &wstr, octal_0: bool) -> usize {
```
Xiretza, line 617: Future work: reuse the mostly-same code in `builtins/echo.rs`

fish-rust/src/wutil/wcstod.rs
```diff
@@ -0,0 +1,444 @@
...
+where
+    Chars: IntoCharIter,
+{
+    let chars = input.chars();
+    if chars.clone().next().is_none() {
+        *consumed = 0;
+        return Err(Error::Empty);
+    }
+
+    let ret = parse_partial_iter(chars.clone().fuse(), decimal_sep);
```
Xiretza, line 30: fastfloat does not support hexadecimal floats, try e.g. `printf %f 0x2`:

old: `2.000000`
new:
```
0x2: value not completely converted (can't convert 'x2')
0.000000
```

`hexf-parse` is used in `fast-float` tests for this, we should probably pull that in as well?

.github/workflows/main.yml
```diff
@@ -79,7 +79,7 @@ jobs:
         #   use-after-scope, double-free, invalid-free, and memory leaks.
         # * MemorySanitizer detects uninitialized reads.
         #
-        RUSTFLAGS: "-Zsanitizer=address"
+        RUSTFLAGS: "-Zsanitizer=address -C link-args=-lasan"
```
mqudsi, line 82: Did you run into a problem that led you to add this?

I spent a lot of time debugging this one to hopefully get it right - I don't think this change is correct. I tried it this way at one point, but it can cause issues depending on the toolchain.

If everything is configured correctly (meaning all components regardless of language are built with `-fsanitize-address` or `-Zsanitizer=address`), `-lsan` is not required (assuming you're compiling with an llvm toolchain and not gcc). 

Forcibly linking against libsan can make it look like memory sanitizer is working but will cause problems - when the compiler is aware of asan (via correct setup) it ensures the symbols end up correctly using the asan versions. But merely linking against asan ensures that the asan symbols are present but doesn't guarantee the asan version of external calls will be used instead of the real version because you're not guaranteeing the final link order.

fish-rust/src/builtins/shared.rs
```diff
@@ -61,6 +63,11 @@ impl output_stream_t {
     pub fn append<Str: AsRef<wstr>>(&mut self, s: Str) -> bool {
         self.ffi().append1(c_str!(s))
     }
+
+    /// Append a char.
+    pub fn append1(&mut self, c: char) -> bool {
+        self.append(wstr::from_char_slice(&[c]))
+    }
 }
```
mqudsi, line 71: Just fyi, the rust convention is `push_str()` and `push()` so you could use `append_str()` and `append()` if you like. Or just `append()` and `push()` for now if the goal is to minimize differences with the cpp code.

mqudsi (commented):
Is there a reason you're pulling the new crates by copying their code (and they use submodules 😬)  instead of forking and adding a git dependency to Cargo.toml like you did with the others? It's significantly harder to see what changes you made from upstream this way. (If the answer is not wanting to clutter the github repo with more forks, I get that and fully sympathize!)

fish-rust/src/locale.rs
```diff
@@ -0,0 +1,133 @@
...
+
+/// Read the numeric locale, or None on any failure.
+// TODO: figure out precisely which platforms have localeconv_l, or use a build script.
+#[cfg(any(
+    target_os = "macos",
+    target_os = "dragonfly",
+    target_os = "freebsd",
+    target_os = "openbsd",
+    target_os = "netbsd",
+))]
```
mqudsi, line 69: `localeconv_l(3)` is available on BSD, but it doesn't seem to be exposed via the libc crate. Blergh.

krobelus: I think we can work around this by declaring it ourselves in an `extern "C"` block.

krobelus (approved):

fish-rust/src/locale.rs
krobelus: that seems to have fixed compilation, now there's only a test failure in locale-specific decimal separators on BSD

    3,45: value not completely converted (can't convert ',45')

fish-rust/src/builtins/shared.rs
ridiculousfish: fixed

src/builtin.h
```diff
@@ -121,5 +121,6 @@ enum RustBuiltin : int32_t {
     Realpath,
     Return,
     Wait,
+    Printf,
```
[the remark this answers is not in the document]
ridiculousfish: fixed

fish-rust/src/wutil/wcstoi.rs
ridiculousfish: I prefer to match the existing C++ at this time. Refactoring can come later.

mqudsi:
@ridiculousfish do you mind addressing some of the unanswered questions here for peace of mind?

fish-rust/src/common.rs
ridiculousfish: Whoops, this slipped through. Fixed in b0a3e14832dfb6eb1c86618f575641a2a4e74c89. Apologies for missing this.

fish-rust/src/wutil/wcstod.rs
ridiculousfish: Well shit. That's incredibly well spotted. I have to fix this.

ridiculousfish:
Yeah I f'ed up here, the observation of hex floats is really good. I should not have merged this. I will revert.

ridiculousfish:
Reverted in 9f7e6a6cd17255ed87a6e11d37537775433eb380. There were outstanding review comments; thank you @mqudsi for calling me out.

fish-rust/src/builtins/printf.rs
ridiculousfish: fixed

.github/workflows/main.yml
ridiculousfish: Yes, the problem was that printf-compat was being compiled with ASAN but not linking against the ASAN libs, causing link errors. I'll check if this is still needed; if not I'll revert it.

fish-rust/src/locale.rs
ridiculousfish: I succcessfully fixed this, it was a real bug.

fish-rust/src/wutil/wcstod.rs
ridiculousfish: I've opted to fix this using (a forked, what else) version of [hexponent](https://docs.rs/hexponent/latest/hexponent/). I'm not happy about it since it's known incorrect for denormals but I don't know what else to do.

fish-rust/src/builtins/printf.rs
ridiculousfish: done

ridiculousfish:
This got re-merged as a487b1ecf26cd13a8d06c3636d2398d11fb09172. The ASAN changes were reverted, and printf-compat is now a separate project in the fish-shell group instead of being checked in directly.

mqudsi:
> The ASAN changes were reverted, and printf-compat is now a separate project in the fish-shell group instead of being checked in directly.

Thanks!

---

I didn't express my appreciation earlier for the obvious amount of work that went into this, thank you!

PR #9669, opened 2023-03-17
Add rpm-ostree completion

yuntaz2:
## Description

Add basic command completion for rpm-ostree. Fedora 37 Silverblue and Kinoite use rpm-ostree as their hybrid image/package system. 
This should improve the user experience for fish users using rpm-ostree.

Fixes issue #

## TODOs:


mqudsi (changes requested):
Thanks for these - they look good but the descriptions are too long. Descriptions don't get too much room in the pager, can you try to keep them below 50 chars or so max? They don't have to be full sentences, perfect grammar, etc - just enough to convey the message.

PR #9671, opened 2023-03-18
builtins: Rewrite `ulimit` in Rust

Tyrubias:
In this PR, we rewrite the `ulimit` builtin in Rust. Something of concern: the `nix` crate doesn't seem to expose the same set of resources that are actually available on macOS.

### DIFF
--- a/fish-rust/src/builtins/ulimit.rs
+++ b/fish-rust/src/builtins/ulimit.rs
@@ -12,11 +12,10 @@ use crate::{
         builtin_unknown_option, BUILTIN_ERR_TOO_MANY_ARGUMENTS, STATUS_CMD_ERROR,
         STATUS_INVALID_ARGS,
     },
-    ffi::{fish_wcswidth2, parser_t},
+    ffi::parser_t,
     wchar::{wstr, L},
-    wchar_ffi::WCharToFFI,
     wgetopt::{wgetopter_t, wopt, woption, woption_argument_t::no_argument},
-    wutil::{fish_wcstoi, wgettext_fmt},
+    wutil::{fish_wcstoi, fish_wcswidth, wgettext_fmt},
 };
 
 use super::shared::{builtin_wperror, io_streams_t, STATUS_CMD_OK};
@@ -227,7 +226,7 @@ fn print_all(hard: bool, streams: &mut io_streams_t) {
     let mut w = 0;
 
     for resource in RESOURCE_ARR {
-        w = max(w, fish_wcswidth2(&resource.desc.to_ffi()).0);
+        w = max(w, fish_wcswidth(resource.desc));
     }
 
     for resource in RESOURCE_ARR {
@@ -281,8 +280,9 @@ fn set_limit(
     value: u64,
     streams: &mut io_streams_t,
 ) -> Option<c_int> {
-    let (mut soft_limit, mut hard_limit) =
-        getrlimit(resource).expect("getrlimit should return valid limits");
+    let Ok((mut soft_limit, mut hard_limit)) = getrlimit(resource) else {
+        return STATUS_CMD_ERROR;
+    };
 
     if hard {
         hard_limit = value
@@ -412,7 +412,7 @@ pub fn ulimit(
         hard = true;
     }
 
-    let new_limit: u64;
+    let mut new_limit: u64;
     if args[woptind].is_empty() {
         streams.err.append(wgettext_fmt!(
             "%ls: New limit cannot be an empty string\n",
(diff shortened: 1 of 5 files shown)


fish-rust/src/builtins/ulimit.rs
```diff
@@ -0,0 +1,354 @@
...
+    /// description of resource
+    desc: &'a wstr,
+    /// switch used on commandline to specify resource
+    switch_char: char,
+    /// the implicit multiplier used when setting getting values
+    multiplier: i32,
+}
+
+/// Array of `Resource` structs, describing all known resource types.
+const resource_arr: &[Resource] = &[
```
mqudsi, line 34: I'm not sure if you intend for this to be `const` and not `static`. `ulimit` isn't a hotspot in the code so I would prefer to prevent allocating space/memory beforehand (i.e. `const`) but at the same time making it `static` might end up  with better overall codegen.

Also, please use idiomatic naming and use screaming snake case here.

fish-rust/src/builtins/ulimit.rs
```diff
@@ -0,0 +1,354 @@
...
+        switch_char: 'l',
+        multiplier: 1024,
+    },
+    #[cfg(any(
+        target_os = "android",
+        target_os = "freebsd",
+        target_os = "netbsd",
+        target_os = "openbsd",
+        target_os = "linux",
+    ))]
```
mqudsi, line 93: You're missing dragonflybsd here, though you have it elsewhere. I'm introducing a build-time feature `bsd` that groups all BSDs together, and can be used instead of micro-targeting the exact os name for BSD derivatives. 

I'll merge that into master soon so you can rebase on top of it and take advantage of it for anything that isn't actually FreeBSD-exclusive.

This is, of course, still not doing feature detection but it'll make it easier not to miss things.

Is the lack of macOS here intentional due to a missing symbol in the libc or nix crates? macOS supports RLIMIT_RSS (I think any \*nix would).

fish-rust/src/builtins/ulimit.rs
```diff
@@ -0,0 +1,354 @@
...
+    Resource {
+        resource: resource::Resource::RLIMIT_NPROC,
+        desc: L!("Maximum number of processes available to current user"),
+        switch_char: 'u',
+        multiplier: 1,
+    },
+    #[cfg(not(any(target_os = "freebsd", target_os = "netbsd", target_os = "openbsd")))]
```
mqudsi, line 145: This should probably not be `not(bsd)` but rather the supported platforms. 

RLIMIT_AS is available on macOS but it's broken and not to be used. Same with RLIMIT_DATA.
Much has been written about macOS w/ RLIMIT_DATA and RLIMIT_AS:

https://bugs.chromium.org/p/chromium/issues/detail?id=853873#c2
https://stackoverflow.com/questions/15633474/python-on-macos-totally-ignoring-rlimit
https://rdrr.io/cran/unix/man/rlimit.html
https://github.com/avast-tl/retdec/issues/379

This would be a change from fish 3.6.0 so maybe it should be held back for a later release (cc @ridiculousfish)

fish-rust/src/builtins/ulimit.rs
Xiretza: > ulimit isn't a hotspot in the code so I would prefer to prevent allocating space/memory beforehand

The data needs to be part of the binary anyway, adding a `static` (`LazyCell` or similar I guess?) would be strictly more expensive because the data needs to be copied to a heap allocation, no?

ridiculousfish: Seconded screaming snake case (`RESOURCE_ARR`). The use of the static const here is fine, it matches the C++ and there's no unnecessary allocations.

mqudsi:
I've added a `bsd` feature to the code in `master`. Can you please rebase and then change any BSD features that aren't specifically FreeBSD features (e.g. kqueue would be specifically FreeBSD) from `target_os = "xxxbsd"` to `feature = "bsd"`?

fish-rust/src/builtins/ulimit.rs
```diff
@@ -0,0 +1,354 @@
+use std::cmp::max;
+
+use libc::{c_int, RLIM_INFINITY};
```
mqudsi, line 3: Remove the second using here then add the following const definition:

```rust
const RLIM_INFINITY: u64 = libc::RLIM_INFINITY as u64;
```

That'll take care of most of the errors you have under 32-bit Ubuntu.

faho:
I'm closing this for now since it's now conflicting and didn't seem to work (the CI logs are unfortunately expired).

If you do get back to it we can of course reopen it.

PR #9675, opened 2023-03-21
completion/ssh-copy-id: add completion

NextAlone:
- completion/ssh-copy-id: add completion

## Description

Talk about your changes here.

Fixes issue #

## TODOs:

### DIFF
--- a/share/completions/ssh-copy-id.fish
+++ b/share/completions/ssh-copy-id.fish
@@ -2,16 +2,10 @@
 complete -c ssh-copy-id -d Remote -xa "(__fish_complete_user_at_hosts)"
 complete -c ssh-copy-id -d Remote -k -fa '(__ssh_history_completions)'
 
-
+# Options
 complete -c ssh-copy-id -s i -d IdentityFile -r -F
 complete -c ssh-copy-id -s p -d Port -r -x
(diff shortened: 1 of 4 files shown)


share/completions/ssh-copy-id.fish
```diff
@@ -0,0 +1,113 @@
+# Retrieve `user@host` entries from history
+function __ssh_history_completions
+    history --prefix ssh --max=100 | string replace -rf '.* ([A-Za-z0-9._:-]+@[A-Za-z0-9._:-]+).*' '$1'
+end
```
mqudsi, line 4: I'm not sure how I feel about this. You're copying the function definition from ssh.fish but you're not giving it a new name. If someone runs into an issue with ssh.fish's function at some later date, they might end up running into issues trying to see why their function definition isn't being reflected.

Since there's a good chance ssh.fish was already loaded, you can try `type -q __ssh_history_completions || source ssh.fish` at the top of this script, or move `__ssh_history_completions` to its own file in `share/functions/__ssh_history_completions.fish`.

NextAlone: Got that, just extract to a split function in function folder.

PR #9676, opened 2023-03-21
Update deno task completions to handle deno.jsonc and package.json

sigmaSd:
I remembered that deno handles deno.jsonc and recently pacakge.json

I want to specify that deno task only takes one arguemnt so for example `deno task a b` is invalid but `deno task a` is, because currently it keep give the completions after each space, is there a flag for that ?

The second completion should not happen
image

### DIFF
--- a/share/completions/deno.fish
+++ b/share/completions/deno.fish
@@ -2,23 +2,18 @@ deno completions fish | source
 
 # complete deno task
 set searchForDenoFilesCode '
+// order matters
 const denoFile = ["deno.json", "deno.jsonc", "package.json"];
 for (const file of denoFile) {
   try {
     Deno.statSync(file);
     // file exists
-    if (file === "package.json") {
-      console.log(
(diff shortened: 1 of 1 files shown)


share/completions/deno.fish
```diff
@@ -1,2 +1,24 @@
...
+      console.log(
+        Object.keys(JSON.parse(Deno.readTextFileSync(file)).tasks).join("\n"),
+      );
+    }
+    break;
+  } catch {
+  }
+}
+'
+complete -f -c deno -n "__fish_seen_subcommand_from task" -a "(deno eval '$searchForDenoFilesCode')"
```
mqudsi, line 24: You can add `-n "__fish_is_nth_token 2"` (I think 2, but you can try 3 if that doesn't work) or you can add `-n "__fish_prev_arg_in task"` to accomplish what you want.

sigmaSd: thanks that works

mqudsi:
Merged.

(GitHub tricked me into rebasing instead of squashing)

Issue #9686, opened 2023-03-26
[Feature Request] Can the "alias" command be made to report individual aliases?


jensenr30:
## System Info
- fish, version 3.1.2
- Linux truenas 5.15.79+truenas #1 SMP Tue Dec 13 12:40:04 UTC 2022 x86_64 GNU/Linux
- My issue is reproducible even when I try it after executing it without 3rd party customizations.

## Feature request: make alias print individual alias definitions

When I run the `alias` command in fish, all of my aliases are printed.  It would be really nice if the `alias` command would accept the name of an existing alias as an argument and display what that alias is currently set to.

Currently, when I try to run the command `alias ll`, I see the following error message: `alias: Body cannot be empty`.  But I have already defined an alias named `ll`, so I think it would be more helpful to fish users if fish instead printed the definition of the existing alias in this case.  For example, if I run the command: `alias ll`, it would be nice to see what my `ll` alias is currently set to: `alias ll 'ls -Al1sh'`

I could probably implement this feature myself and make a pull request, but I'm having trouble finding where the alias command is defined in the source code.  Could someone point me in the right direction?

Thanks,
Ryan

faho:
Fish does not *have* aliases. `alias` is a [convenience function](https://github.com/fish-shell/fish-shell/blob/master/share/functions/alias.fish) to create a specific kind of functions, and even the listing is just a convenience hack (that checks the function description).

What you want is just `functions ll`.

I do not want to adjust alias here because that makes it easier to misuse, and I maintain that most uses of alias are a smell and you should just define a function.

faho:
Alright, question was answered and I don't want to add the change, so I'm closing this.

Use `functions` to see the definition.

PR #9687, opened 2023-03-26
alias can now print the definition of a single alias

jensenr30:
## Description

It is now possible to use alias to print the definition of a single alias.

usage:
`alias my_alias`
fish.alias will now print out the definition of my_alias

if the given alias is not in the list of functions, the error message "alias: Body cannot be empty" is printed as it was before.

Fixes issue #9686 

## TODOs:

Please let me know if this feature will be accepted.  If this feature gets approved, I will make updates to the docs, test, and changelog.

Thanks,
Ryan


faho:
As explained in #9686: I do not want to add something like this to alias.

Alias is a cheesy compatibility function and this makes it easier to misuse.

Sorry, but no.

PR #9688, opened 2023-03-26
Port common.{h,cpp} to Rust

krobelus:
Functions like mbrtowc are not exposed by the libc crate, so declare them
ourselves.
Since we don't know the definition of C macros, add two big hacks to make
this work:
1. Replace MB_LEN_MAX and mbstate_t with values (resp. types) that should
   be large enough for any implementation.
2. Detect the definition of MB_CUR_MAX in the build script. This requires
   more changes for each new libc. We could also use this approach for 1.

Can we live with these hacks?

### DIFF
--- a/fish-rust/build.rs
+++ b/fish-rust/build.rs
@@ -1,6 +1,8 @@
 use miette::miette;
 
 fn main() -> miette::Result<()> {
+    cc::Build::new().file("src/compat.c").compile("libcompat.a");
+
     let rust_dir = std::env::var("CARGO_MANIFEST_DIR").expect("Env var CARGO_MANIFEST_DIR missing");
     let target_dir =
         std::env::var("FISH_RUST_TARGET_DIR").unwrap_or(format!("{}/{}", rust_dir, "target/"));
@@ -25,6 +27,7 @@ fn main() -> miette::Result<()> {
     let source_files = vec![
         "src/abbrs.rs",
         "src/event.rs",
+        "src/common.rs",
         "src/fd_monitor.rs",
         "src/fd_readable_set.rs",
         "src/fds.rs",
--- a/fish-rust/src/common.rs
+++ b/fish-rust/src/common.rs
@@ -9,6 +9,120 @@ use std::mem::ManuallyDrop;
 use std::ops::{Deref, DerefMut};
 use std::os::fd::AsRawFd;
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum EscapeStringStyle {
+    Script(EscapeFlags),
+    Url,
+    Var,
+    Regex,
+}
+
+impl Default for EscapeStringStyle {
+    fn default() -> Self {
+        Self::Script(EscapeFlags::default())
+    }
+}
+
+bitflags! {
+    /// Flags for the [`escape_string()`] function. These are only applicable when the escape style is
+    /// [`EscapeStringStyle::Script`].
+    #[derive(Default)]
+    pub struct EscapeFlags: u32 {
+        /// Do not escape special fish syntax characters like the semicolon. Only escape non-printable
+        /// characters and backslashes.
+        const NO_PRINTABLES = 1 << 0;
+        /// Do not try to use 'simplified' quoted escapes, and do not use empty quotes as the empty
+        /// string.
+        const NO_QUOTED = 1 << 1;
+        /// Do not escape tildes.
+        const NO_TILDE = 1 << 2;
+        /// Replace non-printable control characters with Unicode symbols.
+        const SYMBOLIC = 1 << 3;
+    }
+}
+
+/// Replace special characters with backslash escape sequences. Newline is replaced with `\n`, etc.
+pub fn escape_string(s: &wstr, style: EscapeStringStyle) -> WString {
+    let (style, flags) = match style {
+        EscapeStringStyle::Script(flags) => {
+            (ffi::escape_string_style_t::STRING_STYLE_SCRIPT, flags)
+        }
+        EscapeStringStyle::Url => (
+            ffi::escape_string_style_t::STRING_STYLE_URL,
+            Default::default(),
+        ),
+        EscapeStringStyle::Var => (
+            ffi::escape_string_style_t::STRING_STYLE_VAR,
+            Default::default(),
+        ),
+        EscapeStringStyle::Regex => (
+            ffi::escape_string_style_t::STRING_STYLE_REGEX,
+            Default::default(),
+        ),
+    };
+
+    ffi::escape_string(c_str!(s), flags.bits().into(), style).from_ffi()
+}
+
+/// Test if the string is a valid function name.
+pub fn valid_func_name(name: &wstr) -> bool {
+    if name.is_empty() {
+        return false;
+    };
+    if name.char_at(0) == '-' {
+        return false;
+    };
+    // A function name needs to be a valid path, so no / and no NULL.
+    if name.find_char('/').is_some() {
+        return false;
+    };
+    if name.find_char('\0').is_some() {
+        return false;
+    };
+    true
+}
+
+/// A rusty port of the C++ `write_loop()` function from `common.cpp`. This should be deprecated in
+/// favor of native rust read/write methods at some point.
+///
+/// Returns the number of bytes written or an IO error.
+pub fn write_loop<Fd: AsRawFd>(fd: &Fd, buf: &[u8]) -> std::io::Result<usize> {
+    let fd = fd.as_raw_fd();
+    let mut total = 0;
+    while total < buf.len() {
+        let written =
+            unsafe { libc::write(fd, buf[total..].as_ptr() as *const _, buf.len() - total) };
+        if written < 0 {
+            let errno = errno::errno().0;
+            if matches!(errno, libc::EAGAIN | libc::EINTR) {
+                continue;
+            }
+            return Err(std::io::Error::from_raw_os_error(errno));
+        }
+        total += written as usize;
+    }
+    Ok(total)
+}
+
+/// A rusty port of the C++ `read_loop()` function from `common.cpp`. This should be deprecated in
+/// favor of native rust read/write methods at some point.
+///
+/// Returns the number of bytes read or an IO error.
+pub fn read_loop<Fd: AsRawFd>(fd: &Fd, buf: &mut [u8]) -> std::io::Result<usize> {
+    let fd = fd.as_raw_fd();
+    loop {
+        let read = unsafe { libc::read(fd, buf.as_mut_ptr() as *mut _, buf.len()) };
+        if read < 0 {
+            let errno = errno::errno().0;
+            if matches!(errno, libc::EAGAIN | libc::EINTR) {
+                continue;
+            }
+            return Err(std::io::Error::from_raw_os_error(errno));
+        }
+        return Ok(read as usize);
+    }
+}
+
 /// Like [`std::mem::replace()`] but provides a reference to the old value in a callback to obtain
 /// the replacement value. Useful to avoid errors about multiple references (`&mut T` for `old` then
 /// `&T` again in the `new` expression).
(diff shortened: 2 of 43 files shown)


fish-rust/src/wchar.rs
```diff
@@ -78,3 +42,9 @@ pub fn wchar_literal_byte(byte: u8) -> char {
     char::from_u32(u32::from(ENCODE_DIRECT_BASE) + u32::from(byte))
         .expect("private-use codepoint should be valid char")
 }
+
+pub fn wchar_decode_byte(c: char) -> u8 {
+    (u32::from(c) - u32::from(ENCODE_DIRECT_BASE))
+        .try_into()
+        .unwrap()
```
Xiretza, line 49: Can we make this function return an `Option`? That way it can be called with any character, and the call site doesn't have to do checks manually (`if c >= ENCODE_DIRECT_BASE && c < ENCODE_DIRECT_END { wchar_decode_byte(c) }` becomes `if let Some(c) = wchar_decode_byte(c)`).

fish-rust/src/wutil/mod.rs
```diff
@@ -29,29 +30,20 @@ pub fn perror(s: &str) {
...
-    assert_eq!(join_strings(&[], '/'), "");
-    assert_eq!(join_strings(&[L!("foo")], '/'), "foo");
-    assert_eq!(
-        join_strings(&[L!("foo"), L!("bar"), L!("baz")], '/'),
-        "foo/bar/baz"
-    );
+/// We need this because there are too many implementations that don't return the proper answer for
+/// some code points. See issue #3050.
+pub fn fish_iswalnum(c: char) -> bool {
+    !fish_reserved_codepoint(c) && !fish_is_pua(c) && c.is_alphanumeric()
```
Xiretza, line 48: Shouldn't be necessary, Rust unicode functions don't touch platform stdlibs.

fish-rust/src/common.rs
```diff
@@ -1,13 +1,1707 @@
...
+            out.push(char::from_digit(nibble1.into(), 16).unwrap());
+            out.push(char::from_digit(nibble2.into(), 16).unwrap());
+            need_escape = true;
+            need_complex_escape = true;
+            continue;
+        }
+        match c {
+            '\t' => {
+                if symbolic {
+                    out.push('␉');
```
Xiretza, line 191:
```suggestion
                    out.push('\t');
```

same below, having these raw characters in source code isn't very nice

fish-rust/src/common.rs
```diff
@@ -1,13 +1,1707 @@
...
+                out.push(c);
+            }
+            _ => {
+                let cval = u32::from(c);
+                if cval < 32 {
+                    need_escape = true;
+                    need_complex_escape = true;
+
+                    if symbolic {
+                        out.push(char::from_u32(0x2400 + cval).unwrap());
```
Xiretza, line 280: What's this magic number? Can we add a constant for it?

fish-rust/src/common.rs
```diff
@@ -1,13 +1,1707 @@
...
+    for c in input.chars() {
+        if c >= ENCODE_DIRECT_BASE && c < ENCODE_DIRECT_END {
+            let val = wchar_decode_byte(c);
+            out += "\\X";
+
+            let nibble1 = val / 16;
+            let nibble2 = val % 16;
+
+            out.push(char::from_digit(nibble1.into(), 16).unwrap());
+            out.push(char::from_digit(nibble2.into(), 16).unwrap());
```
Xiretza, line 183: `sprintf!("%%%02X"L, val)[..]` as below in `escape_string_url()`?

Xiretza (commented):
That's a nice chunk of work! I'll have another look later, this is what I've noticed for now.

fish-rust/src/common.rs
krobelus: they are not the same. A literal tab would render as whitespace (of variable width).
This is from the Unicode Control Pictures, it's a symbol for tab that conveniently fits into one display column https://www.fileformat.info/info/unicode/char/2409/index.htm
This is used for displaying control characters to the user in the completion pager

fish-rust/src/common.rs
krobelus: it's not clear to me which version is better `sprintf("\\X%02X", val)` is nice to read but it might add an allocation.
In future I'd like to move away from printf-style format strings eventually to `format!`

fish-rust/build.rs
```diff
@@ -103,12 +113,49 @@ fn detect_features() {
...
+        .stdin(Stdio::piped())
+        .stdout(Stdio::piped())
+        .spawn()
+        .unwrap();
+    {
+        let mut stdin = child.stdin.take().unwrap();
+        stdin
+            .write_all(indoc!(
+                b"#include <stdlib.h>
+                  MB_CUR_MAX"
```
zanchey, line 146: This breaks cross-compilation, right? Not that I guess it works at the moment anyway.

krobelus: Yeah, it's a shame because Rust makes cross compilation much easier.

We could try to replace `mbrtowc` with Rust's decoding routines (the stdlib supports at least UTF-8 and UTF-16). I'm not sure how complex the requisite locale handling is, and whether we want to support other encodings.

fish-rust/src/common.rs
Xiretza: Ooh, tricky, thanks for clearing that up.

ridiculousfish (commented):
This is huge, very impressive!

There's a family of `test_convert*` functions in `fish_tests.cpp` that could be ported to Rust and checked in (duplicated tests), to give us some more confidence that these are right.

fish-rust/src/wutil/encoding.rs
```diff
@@ -0,0 +1,37 @@
+extern "C" {
+    pub fn wcrtomb(s: *mut libc::c_char, wc: libc::wchar_t, ps: *mut mbstate_t) -> usize;
+    pub fn mbrtowc(
+        pwc: *mut libc::wchar_t,
```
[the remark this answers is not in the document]
krobelus: I assumed `wchar_t` -> `char` conversion is zero-extension but I guess it depends on the locale.
I agree we should postpone it

fish-rust/src/common.rs
```diff
@@ -1,13 +1,1707 @@
...
+    }
+    result
+}
+
+/// Returns a newly allocated multibyte character string equivalent of the specified wide character
+/// string.
+///
+/// This function decodes illegal character sequences in a reversible way using the private use
+/// area.
+pub fn wcs2string(input: &wstr) -> Vec<u8> {
```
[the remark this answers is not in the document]
krobelus: Make it two functions?

```rust
pub fn wcs2string(input: &wstr) -> Vec<u8>; // Or OsString
pub fn wcs2zstring(input: &wstr) -> CString;
```

Most callers will use `wcs2zstring`. Maybe we should rename `wcs2string`.
Could also do the rename on the C++ side if that helps.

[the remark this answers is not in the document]
ridiculousfish: Two functions makes some sense to me, since today interior nul bytes can just data to be dropped. Example:

    echo text > abc\0def.txt

will populate a file `abc`. We might want to emit an error for that case, which requires us to know if the string is going to be used as a path or not.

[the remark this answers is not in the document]
mqudsi: @krobelus IMHO `OsString` would be preferable over `Vec<u8>` since it makes the intention and usage clearer.

-----

> I think we need a real nul-terminated string type that allows interior nuls. I'm not sure what it should be. What do you think?

@ridiculousfish what would a nul-terminated string that supports interior nuls look like? You'd necessarily have to use the count/len on one end to support that interior nul meaning the terminating nul would be hanging out there at the end "just because" (you couldn't rely on it for anything) since either both sides of the interface use the count (rendering the nul-termination moot) or one side uses the len but the other side uses the nul termination, meaning the string gets truncated at first nul.

IMHO we shouldn't use nul-termination for anything except where the (legacy/system) callee requires it. In that case, what would make the most sense is to simply have either a type that cannot contain interior nulls (the setter/new takes a regular `&str`/`&wstr` but errors out or returns `None` if it contained an interior nul) or use a conversion function at the call site that copies to a nul-terminated buffer but similarly throws/panics/fails if the source string contained an interior `nul`.

fish-rust/build.rs
mqudsi: I think this only breaks cross-compilation because `cc` is being invoked directly here. We just need to copy-and-paste the code from above to use the `cc-rs` crate to get the compiler and then it should be OK as cross-compilation implies that the headers for the target are automatically called up.

fish-rust/build.rs
```diff
@@ -104,12 +114,49 @@ fn detect_features() {
...
+    match String::from_utf8_lossy(&output.stdout)
+        .lines()
+        .last()
+        .unwrap()
+    {
+        "__mb_cur_max" => Ok(Some("apple_libc")),
+        "(___mb_cur_max())" | "((size_t)___mb_cur_max())" => Ok(Some("bsd_libc")),
+        "(__ctype_get_mb_cur_max ())" => Ok(Some("glibc")),
+        def => panic!("Unsupported definition of MB_CUR_MAX: {}", def),
+    }
```
mqudsi, line 161: Does this work without also adding these features to `Cargo.toml`? If so, that's really cool. It would make it harder for users to manually specify features that are meant to be auto-detected.

fish-rust/src/wutil/encoding.rs
```diff
@@ -0,0 +1,37 @@
...
+
+// HACK This should be mbstate_t from libc but that's not exposed.  Since it's only written by
+// libc, we define it as opaque type that should be large enough for all implementations.
+pub type mbstate_t = [u64; 16];
+pub fn zero_mbstate() -> mbstate_t {
+    [0; 16]
+}
+
+// HACK This should be the MB_LEN_MAX macro from libc but that's not easy to get.
+pub const AT_LEAST_MB_LEN_MAX: usize = 32;
```
mqudsi, line 37: I think the value here is safe to use, but I've been planning on adding build-time constant retrieval to `build.rs`. This is a good candidate.

fish-rust/src/wutil/encoding.rs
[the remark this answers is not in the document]
mqudsi: I suggest that we can just use rust's std conversion facilities (that only support UTF-8/16/32) for Cygwin.

fish-rust/src/common.rs
```diff
@@ -1,13 +1,1714 @@
...
+
+pub fn unescape_string(input: &wstr, style: UnescapeStringStyle) -> Option<WString> {
+    match style {
+        UnescapeStringStyle::Script(flags) => unescape_string_internal(input, flags),
+        UnescapeStringStyle::Url => unescape_string_url(input),
+        UnescapeStringStyle::Var => unescape_string_var(input),
+    }
+}
+
+pub fn unescape_string_in_place(s: &mut WString, style: UnescapeStringStyle) -> bool {
```
mqudsi, line 420: Can you add a comment:

```rust
/// TODO: Make this actually in-place or rename it to make it clear that it's not an in-place unescape.
```

fish-rust/src/common.rs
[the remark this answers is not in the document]
ridiculousfish: A nul-terminated string that supports interior nuls looks like `std::string`. If you pass it to a syscall like `open` which expects nul-termination then it will truncate your filename; but if you pass it to `write` which has pointer/length pair, it will happily write the zero byte, as it should. So it depends.

What you propose is the two-function approach: interior nuls for some cases (a binary file stream), prohibit them for others (a file redirection), and the call sites have to decide which to use. If that's achievable then I'm for it. It may also be too difficult - we use `wcs2string` a lot, and maybe it's too difficult to tease apart these two use cases, so we should just have some Rust analog to `std::string`.

I want to defer to krobelus here; we should aim to match the existing behavior and we can defer any refactoring to later.

fish-rust/src/wutil/encoding.rs
krobelus: how should we retrieve the constants, I think the most robust way requires us to run a small C program.

fish-rust/src/common.rs
krobelus: actually the comment should say we plan to remove it

fish-rust/build.rs
krobelus: I think we'd need to use [cargo:rustc-cfg=KEY[="VALUE"]](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-cfg)
so we set it as cfg value directly, not as feature

fish-rust/build.rs
krobelus: thanks for the pointer to `cc-rs`. It looks like we can use it to define a small C library that the rest of the code can link against, so I'll put a portable wrapper around `MB_CUR_MAX` there

fish-rust/src/common.rs
```diff
@@ -1,13 +1,1714 @@
...
+/// A function type to check for cancellation.
+/// \return true if execution should cancel.
+pub type CancelChecker = dyn Fn() -> bool;
+
+/// Converts the narrow character string \c in into its wide equivalent, and return it.
+///
+/// The string may contain embedded nulls.
+///
+/// This function encodes illegal character sequences in a reversible way using the private use
+/// area.
```
krobelus, line 1037: The documentation of str2wcstring says

	/// Converts the narrow character string \c in into its wide equivalent, and return it.
	///
	/// This function encodes illegal character sequences in a reversible way using the private use
	/// area.
	pub fn str2wcstring(inp: &[u8]) -> WString;

Surprisingly, it may produce surrogates (0xD800..0xDFFF) and characters that are outside the Unicode codespace (> 0x10FFFF) because mbrtowc does not prevent this.

For instance

	#include <assert.h>
	#include <locale.h>
	#include <stdint.h>
	#include <stdio.h>
	#include <string.h>
	#include <wchar.h>

	int main() {
	    setlocale(LC_ALL, "");
	    wchar_t wc = L'\0';
	    uint8_t input[4] = {245, 155, 155, 143};
	    mbstate_t ps;
	    memset(&ps, '\0', sizeof(mbstate_t));
	    ssize_t ret = mbrtowc(&wc, (const char*)&input[0], 4, &ps);
	    assert(ret == 4);
	    assert(wc == 0x15b6cf);
	}

Rusts `char::from_u32` fails for such code points, breaking many assumptions.
I think due to the broken invariant of `WString`, its `char_at()` does not fail but silently return an invalid character.

To make this easier to handle, I'll encode invalid code points with the private-use-are trick.

fish-rust/src/common.rs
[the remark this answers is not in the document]
krobelus: > IMHO OsString would be preferable over Vec<u8> since it makes the intention and usage clearer.

Yes but I'm not sure where to draw the line. In the strictest interpretation, we would change almost every use of `Vec<u8>` to `OsString` (and `&[u8]` to `OsStr`) because we almost never deal with binary data.
OsString is a more limited API, for example it doesn't have `pop()`.
Also, `OsString::push()` takes a `OsSt`, we cannot push a char directly.
This suggests that we should still use `Vec<u8>` when manipulating a string and convert to `OsStr` later.
There is another problem: `OsStr` does not support slice indexing.

I'm positive that it's worth trying to adopt `OsStr` but it needs more work and thought.
While we use `Vec<u8>`, we do need to convert to `OsStr` whenever we calling something that takes `AsRef<Path>` but that's a small price to pay.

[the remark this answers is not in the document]
Xiretza: The point of `OsStr` is basically that you can't do anything with it directly, since its representation is platform-dependent. That's fine for "take this data from the kernel, put it back in the kernel somewhere else", but less useful for a shell that actually has to look at the contents and manipulate them. I think for the majority of use cases in fish, `OsStr` is the wrong abstraction.

fish-rust/src/wutil/encoding.rs
ridiculousfish: 32 seems fine to me, it's very conservative. The largest value I can find in the wild is 16.

fish-rust/src/common.rs
ridiculousfish: Yuck. Yeah your proposed private-use-area trick seems right to me.

fish-rust/src/common.rs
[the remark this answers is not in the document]
ridiculousfish: Agree with Xiretza to avoid OsString.

Looking at the call sites, I think the two function approach should be very doable. The great majority of `wcs2string` calls are in preparation for a system call (e.g. `wunlink`); these can use `std::ffi::CString` which is what it's there for.

The only important exception is `io_buffer_t` which holds a buffer, e.g. the output of a command when executing a command substitution. This holds a `std::string`, needs to support interior nuls, but does NOT need to be nul-terminated. A `Vec<u8>` would be totally fine here.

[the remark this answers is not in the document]
mqudsi: > It may also be too difficult - we use wcs2string a lot, and maybe it's too difficult to tease apart these two use cases, so we should just have some Rust analog to std::string.

From the perspective of simply finishing the port, I fully agree. 

However I would say silently truncating *some* strings at the first null but not others is a bug - the operation should certainly fail rather than half-succeed.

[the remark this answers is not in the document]
krobelus: CString does not allow interior nulls so this check is easy to implement.
We'd need to make `wcs2zstring` return an error instead of truncating the string.
Probably not worth the effort though.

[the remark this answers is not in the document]
mqudsi: If we get idiomatic rust error handling going, it would be as easy as adding a question mark at any call site and defining an `Into` for the error that automatically sets the error message to something appropriate. But probably best left for later.

PR #9706, opened 2023-04-03
completions/mix: Add options for phx.new in 1.7.2

mfilej:
## Description

This adds two missing `phx.new` options as of 1.7.2.

It also adds the possible values for `--database`. I changed `-f` to `-x` because the option doesn't make sense without a value, but I wanted to double check if that's ok. (cc @woutdp)

## TODOs:


woutdp:
> It also adds the possible values for --database. I changed -f to -x because the option doesn't make sense without a value, but I wanted to double check if that's ok. (cc @woutdp)

Nice makes sense

PR #9707, opened 2023-04-03
Fix adb path completion

shilangyu:
## Description

Fixed path completion for adb-connected devices. There was a bug where suffixing paths was done using the first segment of a line (ie up to a separator, here a space) in awk: `$1`. Fixed by taking the whole line instead: `$0`. Is this fix worth mentioning in the changelog? If so, I will add an entry.

### Before

image

### After

image

## TODOs:


mqudsi:
Thanks, nice catch. There's no need to changelog this.

PR #9718, opened 2023-04-07
completions/git: Allow switch to complete remote branches

ericvw:
## Description

completions/git: Allow switch to complete remote branches
    
While it is true that `git switch <remote-branch>` errors to disallow a detached
head without the `-d` option, it is valid to use any starting point (commit or
reference) in conjunction with the `-c` option. Additionally, the starting point
can occur before any option.
    
This enables the following completions:
    
* `git switch -c <local-name> <any-branch>`
* `git switch <any-branch> -c <local-name>`
* `git switch -d <any-starting-point>`
* `git switch <any-branch> -d`
    
The trade-off is this does allow for `git switch <remote-branch>` to be
completed with an error.
    
Note that this logically reverts 7e3d3cc30f61d466f450a61ebee7c7fcd264967f.

## TODOs:

### DIFF
--- a/share/completions/git.fish
+++ b/share/completions/git.fish
@@ -1956,7 +1956,7 @@ complete -F -c git -n '__fish_git_using_command restore' -n '__fish_git_contains
 # switch options
 complete -f -c git -n __fish_git_needs_command -a switch -d 'Switch to a branch'
 complete -f -c git -n '__fish_git_using_command switch' -ka '(__fish_git_unique_remote_branches)' -d 'Unique Remote Branch'
-complete -f -c git -n '__fish_git_using_command switch' -ka '(__fish_git_local_branches)'
+complete -f -c git -n '__fish_git_using_command switch' -ka '(__fish_git_branches)'
 complete -f -c git -n '__fish_git_using_command switch' -s c -l create -d 'Create a new branch'
 complete -f -c git -n '__fish_git_using_command switch' -s C -l force-create -d 'Force create a new branch'
 complete -f -c git -n '__fish_git_using_command switch' -s d -l detach -d 'Switch to a commit for inspection and discardable experiment' -rka '(__fish_git_refs)'


share/completions/git.fish
```diff
@@ -1956,10 +1956,10 @@ complete -F -c git -n '__fish_git_using_command restore' -n '__fish_git_contains
 # switch options
 complete -f -c git -n __fish_git_needs_command -a switch -d 'Switch to a branch'
 complete -f -c git -n '__fish_git_using_command switch' -ka '(__fish_git_unique_remote_branches)' -d 'Unique Remote Branch'
-complete -f -c git -n '__fish_git_using_command switch' -ka '(__fish_git_local_branches)'
+complete -f -c git -n '__fish_git_using_command switch' -ka '(__fish_git_branches)'
 complete -f -c git -n '__fish_git_using_command switch' -s c -l create -d 'Create a new branch'
 complete -f -c git -n '__fish_git_using_command switch' -s C -l force-create -d 'Force create a new branch'
-complete -f -c git -n '__fish_git_using_command switch' -s d -l detach -d 'Switch to a commit for inspection and discardable experiment' -rka '(__fish_git_refs)'
+complete -f -c git -n '__fish_git_using_command switch' -s d -l detach -d 'Switch to a commit for inspection and discardable experiment'
```
faho, line 1962: This should be kept. Giving commits to `switch --detach` is okay.

ericvw: Since the `--detach` can come after the commit (e.g., `switch <start-point> --detach`), I'll modify argument completion to use `__fish_git_refs` instead of `__fish_git_branches` because `_fish_git_refs` includes `__fish_git_branches`.

faho: No. We complete what we know, so we complete refs if we got `--detach`.

Sometimes offering less is more. Keeping branches otherwise is fine!

faho: The issue is completing random commits isn't very useful here, and if you do `git switch 374637deadbeef` you'll then *have to* add `--detach`.

It's *fine* if that's not modelled 100% - just keep it simple and offer the much more useful branches unless `--detach` is already given, then switch to offering all "refs".

ericvw: Ahh — so for the form `switch --detach <starting-point>` to use `__fish_git_refs` will result in that form to get all references.

I'll make that adjustment now.

ericvw: I've updated the PR to bring back `__fish_git_refs` for `--detach` and updated the commit message appropriately.

faho:
Merged, thanks!

PR #9725, opened 2023-04-10
Port widecharwidth, fallback, wutil, io

krobelus:


Xiretza (commented):
Great to see the io stuff ported! Hopefully the builtins can use it soon, that's currently a bit of a nasty FFI interface in my opinion.

fish-rust/src/io.rs
```diff
@@ -0,0 +1,951 @@
...
+pub type IoDataRef = Rc<dyn IoData>;
+
+#[derive(Default)]
+pub struct IoChain(pub Vec<IoDataRef>);
+
+impl IoChain {
+    pub fn new() -> Self {
+        Default::default()
+    }
+    pub fn remove(&mut self, element: &IoDataRef) {
```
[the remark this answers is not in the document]
krobelus: We cannot compare fat trait pointers, for example given

```
        let a = Rc::as_ptr(element);
        let b = Rc::as_ptr(element);
        assert!(std::ptr::eq(a, b));
```

clippy says

```
error: comparing trait object pointers compares a non-unique vtable address
   --> src/io.rs:594:17
    |
594 |         assert!(std::ptr::eq(a, b));
    |                 ^^^^^^^^^^^^^^^^^^
    |
    = help: consider extracting and comparing data pointers only
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#vtable_address_comparisons
    = note: `#[deny(clippy::vtable_address_comparisons)]` on by default

error: could not compile `fish-rust` due to previous error
```

That's why we need the pointer to the underlying object to match the behavior of the C++ code.
The  AST uses the same trick.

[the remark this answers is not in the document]
Xiretza: You can do as it suggests and compare the data pointer only, ignoring the pointer metadata:

```rust
let a = Rc::as_ptr(element) as *const ();
let b = Rc::as_ptr(element) as *const ();
assert!(std::ptr::eq(a, b));
```

fish-rust/src/common.rs
```diff
@@ -1649,11 +1662,11 @@ static IS_WINDOWS_SUBSYSTEM_FOR_LINUX: Lazy<bool> = Lazy::new(|| {
     let dash = info.release.iter().position('-');
 
     if dash
-        .map(|d| unsafe { libc::strtod(std::ptr::addr_of!(info.release[d + 1]), std::ptr::null()) } >= 17763)
+        .map(|d| unsafe { libc::strtod(&info.release[d + 1], std::ptr::null()) } >= 17763)
```
faho, line 1665: Could this use [wcstoi](https://github.com/fish-shell/fish-shell/blob/master/fish-rust/src/wutil/wcstoi.rs) instead? Version numbers aren't floats.

fish-rust/src/fallback.rs
```diff
@@ -0,0 +1,123 @@
...
+    }
+    result
+}
+
+// Replacement for mkostemp(str, O_CLOEXEC)
+// This uses mkostemp if available,
+// otherwise it uses mkstemp followed by fcntl
+pub fn fish_mkstemp_cloexec(name_template: CString) -> (RawFd, CString) {
+    let name = name_template.into_raw();
+    #[cfg(not(target_os = "macos"))]
```
faho, line 96: (would be great if we could check if mkostemp is available, but I gather that's a general problem?)

fish-rust/src/job_group.rs
```diff
@@ -255,11 +294,15 @@ impl JobGroup {
...
+    const NONE: MaybeJobId = MaybeJobId(None);
+
+    pub fn new(value: NonZeroU32) -> Self {
+        JobId(value)
+    }
 
     /// Return a `JobId` that is greater than all extant job ids stored in [`CONSUMED_JOB_IDS`].
     /// The `JobId` should be freed with [`JobId::release()`] when it is no longer in use.
-    fn acquire() -> Option<Self> {
+    fn acquire() -> MaybeJobId {
```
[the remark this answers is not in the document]
faho: Would match [the old C++](https://github.com/fish-shell/fish-shell/blob/f39bc9317d252ce2229ebc11a8a6fe2d98abf430/src/job_group.cpp#L17-L24), which just returned a "job_id_t" (i.e. [an int](https://github.com/fish-shell/fish-shell/blob/f39bc9317d252ce2229ebc11a8a6fe2d98abf430/src/job_group.h#L13-L15)).

Presumably this *could* return an error if acquiring the mutex failed or all job are full?

fish-rust/src/wutil/mod.rs
```diff
@@ -32,6 +82,402 @@ pub fn perror(s: &str) {
...
+    wd_comps.extend(path_comps.into_iter().skip(erase_count));
+
+    let mut result = join_strings(&wd_comps, sep);
+    result.insert(0, '/');
+    result
+}
+
+/// Wide character version of dirname().
+#[widestrs]
+pub fn wdirname(path: WString) -> WString {
```
faho, line 326: In general, when we're dealing with paths and wstrings, I have a question:

How do these work with NULs? Because UNIX paths can't have NULs, so this should obey c-string semantics and only operate until the first NUL.

fish-rust/src/wutil/mod.rs
```diff
@@ -50,6 +496,324 @@ pub fn fish_iswalnum(c: char) -> bool {
...
+            ]
+            .contains(&&dent.d_name[..3])
+        {
+            return self.next();
+        }
+
+        self.entry.reset();
+        let d_name: Vec<u8> = dent.d_name.iter().map(|b| *b as u8).collect();
+        self.entry.name = cstr2wcstring(&d_name);
+        self.entry.inode = dent.d_ino;
```
faho, line 806: I'm confused as to why this fails on FreeBSD - is libc incomplete there?

fish-rust/src/wutil/mod.rs
krobelus: yeah that's an interesting case. While `wdirname()` accepts interior NULs, an eventual system call will not.
Rust makes things better because CString will report interior NULs as error instead of std::string's silent truncation.

For now I maintained the existing behavior in the place where most CStrings are created:

```rust
pub fn wcs2zstring(input: &wstr) -> CString {
    if input.is_empty() {
        return CString::default();
    }

    let mut result = vec![];
    wcs2string_callback(input, |buff| {
        result.extend_from_slice(buff);
        true
    });
    let until_nul = match result.iter().position(|c| *c == b'\0') {
        Some(pos) => &result[..pos],
        None => &result[..],
    };
    CString::new(until_nul).unwrap()
}
```

which is of course super surprising (in combination with nul-agnostic `wdirname` and such) but it has always been that way and not caused problems

fish-rust/src/wutil/mod.rs
krobelus: looks like it's called `d_ino` on some platforms (https://docs.rs/libc/latest/libc/struct.dirent.html)
but `d_fileno` [on BSD](https://rust-lang.github.io/libc/x86_64-unknown-freebsd/doc/libc/struct.dirent.html)
which is totally weird because `fileno()` is a libc function for something totally different

faho: wtfudge

faho: This goes back to https://github.com/rust-lang/libc/pull/24, which called it d_fileno on freebsdlike and openbsdlike. I don't see an explanation.

(NetBSD was added later and also received "d_fileno")

krobelus: `d_fileno` is also the name documented in the FreeBSD [man page for dirent](https://manpage.me/?q=dirent)
Not sure how we could use `d_ino` on all platforms in C++, maybe the BSD C library defines d_ino for compatibillity

faho: 1. I hate posix.
2. Seems we need one of those `#cfg` thingamabobs here

PR #9729, opened 2023-04-10
Rewrite builtin `type` and `command` in Rust

faho:
## Description

This reimplements the `type` builtin in rust.

It's quite unclean, mainly because I could not figure out how to finagle a few things through the FFI:

- wcstring_list_t (i.e. `std::vector<std::wstring>`) (and https://github.com/fish-shell/cxx/pull/2 unfortunately doesn't seem to help)
- function_properties_t (and the "ref" shared_ptr)

It runs and passes the tests, but I'm assuming it's quite slow because it looks up the function multiple times, and runs str2wcstring/wcs2string more than necessary.

## TODOs:


mqudsi:
Just fyi I already implemented https://github.com/fish-shell/cxx/pull/2 in our cxx fork. I can't remember if I updated the cargo dependency - if not, I'll do that now.

mqudsi:
@faho https://github.com/mqudsi/fish-shell/commit/4e95ea9fb2e1dd94e3044b822a2e6cdf1af81a8c reverts your "Add method to return narrow strings" and gets ride of the wide-to-narrow and then narrow-back-to-wide for every path by using the VectorElement impl but also working around some weird autocxx behavior I haven't figured out that doesn't read the vector element as CxxWString.

faho:
Okay, the wide path trickery works on my system, but fails on CI - and apparently Github Actions is just down at the moment, so only Cirrus is running?

I did some pretty disgusting trickery to use the function_properties_t - this involves making the _ref_t *not const* (because [cxx apparently forgets the const?](https://github.com/dtolnay/cxx/issues/850)). Which is terrible, but considering it always was const it's probably something we can live with until it's ported.

src/path.h
```diff
@@ -62,7 +63,9 @@ struct get_path_result_t {
 get_path_result_t path_try_get_path(const wcstring &cmd, const environment_t &vars);
 
 /// Return all the paths that match the given command.
-wcstring_list_t path_get_paths(const wcstring &cmd, const environment_t &vars);
+std::vector<std::wstring> path_get_paths(const wcstring &cmd, const environment_t &vars);
+
+std::vector<std::string> path_get_paths_from_parser(const wcstring &cmd, const parser_t &parser);
```
krobelus, line 68: we should mark this wrapper somehow so we know not to port it to Rust. Usually we add an `_ffi` suffix

mqudsi: I wish I knew why autocxx was OK with the wide version of this everywhere except on macOS.

faho: It wasn't happy on FreeBSD either: https://cirrus-ci.com/task/5127876329078784.

(my guess is compiler versions, but tbh given that this is *not* a huge performance penalty I'm just gonna take it in this case - it's much more pressing in others)

krobelus:
you could define a `path_get_path_ffi` wrapper that returns a `std::unique_ptr<wcstring>`.
I don't care much about a transient performance loss but we'd want something to remind us to fix it ASAP.
(We can also wait for the path.cpp port. it requires env which isn't far off)
Either way it would be good to make every commit pass all tests, by squashing most of them

faho:
>you could define a path_get_path_ffi wrapper that returns a std::unique_ptr<wcstring>.

That lets us run into our old friend, `std_basic_string_wchar_t_AutocxxConcrete` again.

(I'm *really* not impressed with how easy it is to make autocxx confused. Of course my own inexperience doesn't help either)

>I don't care much about a transient performance loss but we'd want something to remind us to fix it ASAP.

That's simple enough - it uses `path_get_paths_ffi`. Since that won't survive the path port, you get an automatic compile error, and the TODO comment right above it tells you that you should also do that.

Or, if you want, we can add another kind of marker - this won't be the only case.

> Either way it would be good to make every commit pass all tests, by squashing most of them

Should be okay now.

src/path.cpp
```diff
@@ -168,6 +168,15 @@ wcstring_list_t path_get_paths(const wcstring &cmd, const environment_t &vars) {
     return paths;
 }
 
+std::vector<std::string> path_get_paths_ffi(const wcstring &cmd, const parser_t &parser) {
```
ridiculousfish, line 171: With the new `wcstring_list_ffi_t` this can be:

    wcstring_list_ffi_t path_get_paths_ffi(const wcstring &cmd, const parser_t &parser) {
         return path_get_paths(cmd, parser);
    }
    
Then on the Rust side:

    paths: Vec<WString> = ffi:: path_get_paths_ffi(cmd, parser).from_ffi();
    
pretty nice

ridiculousfish (commented):
very nice work

src/path.cpp
faho: Oooh, nice.

Done in three commits so I can make them disappear.

fish-rust/src/builtins/command.rs
```diff
@@ -0,0 +1,108 @@
...
+    };
+
+    // Quiet implies find_path.
+    if !opts.find_path && !opts.all && !opts.quiet {
+        builtin_print_help(parser, streams, cmd);
+        return STATUS_INVALID_ARGS;
+    }
+
+    let mut res = false;
+    for arg in argv.iter().take(argc).skip(optind) {
```
Xiretza, line 83:
```suggestion
    let optind = w.woptind;
    for arg in argv.iter().take(argc).skip(optind) {
```

This avoids the mess above.

faho: Huh, it surprises me that *that* works.

But oh well, let's take that version and just stick with it

faho:
Alright, I'm merging this now, and I'm gonna push builtin builtin after, because it's closely related and basically trivial (once builtin_get_names works).

fish-rust/src/builtins/command.rs
Xiretza: The problem was that it desugared to something like this:

```rust
let temp = argv.iter(); // `argv` immutably borrowed here
let temp = temp.take(argc);
let temp = temp.skip(w.woptind); // `w` used here, which means it must be alive - including its mutable reference to `argv`, causing overlap
```

If you pull out `woptind` first, then never use `w` again, its lifetime can be shortened such that it doesn't overlap with the `argv` borrow (thanks to [NLL](https://blog.rust-lang.org/2022/08/05/nll-by-default.html), I believe).

PR #9732, opened 2023-04-12
Ansible completion: fix typo in `--limit-hosts`

cosi1:
## Description

I believe it's a typo in the Ansible completion file: there's no such option as `--limit-hosts`. The correct option would be `--list-hosts`, which is missing. (The option's description seems to confirm it.)


faho:
Thanks!

PR #9734, opened 2023-04-12
Fix `composer require` completion

jtojnar:
When no development dependencies are installed, the completion would crash with:

    KeyError: 'require-dev'


faho:
Merged, thanks!

PR #9742, opened 2023-04-21
add recent commits to completion for `git switch --detach`

### LINKED ISSUE #9741: `git switch --detach` does not query and complete for commits, even though `git checkout` does
### issue

when typing

```
$ git checkout [commit message or hash]
```

and pressing `tab`, it queries and lists possible commits, but doing

```
$ git switch --detach [commit message or hash]
```

only queries for branches, and not commits, even though both commands should do the same

#### extra info

also occurs when using `sh -c 'env HOME=$(mktemp -d) fish'`

```
$ fish --version && echo $version
fish, version 3.6.1
3.6.1
```

m4rch3n1ng:
## Description

fixes #9741  
adds `'(__fish_git_recent_commits --all)'` to the `git switch --detach` command

## TODOs:


faho:
Merged, thanks!

Note for the future: It's fine, for simple things like this, to just open a PR. The issue isn't of much use if the fix is already clear and doesn't need discussion.

m4rch3n1ng:
thanks for the merge! yeah i know, i didn't initially plan on creating a pr for the issue but after creating the issue i was like "i can probably make a pr for this" and i could

Issue #8604, opened 2022-01-02
fish_remove_path or equivalent, or documentation update


losvedir:
```
~> fish --version                                                                                                                                                                 06:22
fish, version 3.3.1

~> echo $version                                                                                                                                                                  06:22
3.3.1

~> uname -a                                                                                                                                                                       06:22
Darwin Gabes-MacBook-Pro.local 21.2.0 Darwin Kernel Version 21.2.0: Sun Nov 28 20:28:41 PST 2021; root:xnu-8019.61.5~1/RELEASE_ARM64_T6000 arm64

~> echo $TERM                                                                                                                                                                     06:22
xterm-256color

~>  sh -c 'env HOME=$(mktemp -d) fish'
# yes, behavior is present
```

I know this is probably a stupid question, but I've spent a bit of time trying to figure out how to do this simple thing and haven't had any luck.

How do you remove an item from the path?

I added an item to `$PATH` using the handy `fish_add_path`:

```
~> fish_add_path "/Users/gabedurazo/Downloads"
```

I then see it in `$PATH`:

```
~/zig> echo $PATH                                                                                                                                                                 06:21
/Users/gabedurazo/Downloads [other paths...]
```

I guessed `fish_remove_path`, but that isn't a known command. So I went to the [`fish_add_path` documentation](https://fishshell.com/docs/current/cmds/fish_add_path.html), and was surprised to see no mention of it there. However, it mentioned `$fish_user_paths` and pointed me to the concept of [universal variables](https://fishshell.com/docs/current/language.html#variables-universal), so I thought maybe I was supposed to remove it manually from `$fish_user_paths`. From the documentation there, it seemed that `$fish_user_paths` is a list, and so I tried using `set -e` to remove the item following the example in the documentation there, but it didn't work:

```
~/zig> echo $fish_user_paths[1]                                                                                                                                                   06:24
/Users/gabedurazo/Downloads

~/zig> set -e $fish_user_paths[1]                                                                                                                                                 06:26
set: Variable name '/Users/gabedurazo/Downloads' is not valid. See `help identifiers`.
```

At this point, I'm not sure what to try next. I think maybe ultimately this might be a suggest a documentation fix, which I'd be happy to do.

faho:
You can remove it the same way you would remove any other element from a list: Find he index, then erase it:

```fish
if set -l index (contains -i /path/to/whatever $PATH)
    set -e PATH[$index]
end
```

[...]

Issue #9147, opened 2022-08-18
Enhanchement Remove path from $PATH


ShalokShalom:
I suggest `fish_remove_path` or something similar.

Currently, there is no way (in the documentation) to remove strings from the PATH. 

https://fishshell.com/docs/current/tutorial.html#path
https://fishshell.com/docs/current/cmds/fish_add_path.html#cmd-fish-add-path

faho:
Duplicate of #8604

faho:
Short answer is it's impossible to permanently remove a path from $PATH outside of the place where it's added, and so fish_remove_path would have a gaping hole that can't be plugged, so it's better to not pretend to have it.

PR #9744, opened 2023-04-22
Add --remove to fish_add_path

c31io:
## Description

A new option for cleaning $fish_user_paths with some warnings.
This change makes maintaining $fish_user_path easier for beginners without introducing a stand-alone footgun fish_remove_path function.

Fixes issue #8604 #9147

## TODOs:

### DIFF
--- a/doc_src/cmds/fish_add_path.rst
+++ b/doc_src/cmds/fish_add_path.rst
@@ -10,7 +10,7 @@ Synopsis
 .. synopsis::
 
     fish_add_path path ...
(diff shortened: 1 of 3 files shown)


faho:
Those issues were closed on purpose: This has trade-offs that I find unacceptable.

Sorry, but no.

alifarazz:
>This has trade-offs that I find unacceptable.

Okay. Then what other method do you propose?
Because, currently, there is not command in fish that does the opposite of `fish_add_path`.

---

One can use `$ set -e fish_user_paths[$idx]`, where `idx` is the entry index in `fish_user_paths` var, but that's not obvious to  beginners.

KyunLFA:
As a non-advanced user I would personally benefit from being able to remove a path from PATH just as easily as I can add it. Just my own personal opinion.

develmac:
"the friendly interactive shell" guess not user friendly then?

PR #9746, opened 2023-04-22
Fix rust version of is_wsl() check

mqudsi:
Somewhat counter-intuitively, this code is active when compiling under *Linux* and is always false when compiling under Windows. The logic was incorrectly reversed before (it's easier to reason about when you realize that fish doesn't even compile under Windows because it uses tons of libc functions).

As the code was actually never compiled, it wasn't actually tested for validity either and there were some issues that prevented it from compiling that have since been fixed. The logic has also been adjusted a bit to make it possible to use the rust-native int parsing instead of `libc::strtod()`.

The code has been changed to use `once_cell::race::OnceBool` instead of `once_cell::sync::Lazy<T>` which imposes a greater runtime burden with locking and other overhead. We don't care if the code runs more than once on init (if calls were to race, though they probably don't) - just that the code isn't subsequently executed on each call. The `once_cell::race` module is a better fit here, though it doesn't expose the ergonomic `Lazy<T>` façade around its types.

### DIFF
--- a/fish-rust/src/common.rs
+++ b/fish-rust/src/common.rs
@@ -1634,6 +1634,13 @@ pub fn restore_term_foreground_process_group_for_exit() {
     }
 }
 
+fn slice_contains_slice<T: Eq>(a: &[T], b: &[T]) -> bool {
+    a.windows(b.len()).any(|aw| aw == b)
+}
+
+#[cfg(target_os = "linux")]
+static IS_WINDOWS_SUBSYSTEM_FOR_LINUX: once_cell::race::OnceBool = once_cell::race::OnceBool::new();
+
 /// Determines if we are running under Microsoft's Windows Subsystem for Linux to work around
 /// some known limitations and/or bugs.
 /// See https://github.com/Microsoft/WSL/issues/423 and Microsoft/WSL#2997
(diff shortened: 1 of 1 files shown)


fish-rust/src/common.rs
```diff
@@ -1634,6 +1634,13 @@ pub fn restore_term_foreground_process_group_for_exit() {
     }
 }
 
+fn slice_contains_slice<T: Eq>(a: &[T], b: &[T]) -> bool {
+    a.windows(b.len()).any(|aw| aw == b)
+}
+
+#[cfg(target_os = "linux")]
+static IS_WINDOWS_SUBSYSTEM_FOR_LINUX: once_cell::race::OnceBool = once_cell::race::OnceBool::new();
```
krobelus, line 1642: Right, I always use `once_cell:sync` even if it's not necessary, I should recheck all of them

fish-rust/src/common.rs
```diff
@@ -1642,55 +1649,61 @@
...
-    *IS_WINDOWS_SUBSYSTEM_FOR_LINUX
-}
-
-fn slice_contains_slice<T: Eq>(a: &[T], b: &[T]) -> bool {
-    a.windows(b.len()).any(|aw| aw == b)
-}
-
-#[cfg(not(windows))]
-static IS_WINDOWS_SUBSYSTEM_FOR_LINUX: Lazy<bool> = Lazy::new(|| false);
-#[cfg(windows)]
```
krobelus, line 1654: my bad, I somehow mixed up cfg(windows) with the WSL define. 
LGTM!

FWIW could keep using macros like that with `fish-rust/src/compat.{c,rs}` but it's probably not worth it here

fish-rust/src/common.rs
```diff
@@ -1642,55 +1649,61 @@
...
     }
 
-    // Sample utsname.release under WSL, testing for something like `4.4.0-17763-Microsoft`
-    if !slice_contains_slice(&info.release, b"Microsoft") {
-        return false;
-    }
-    let dash = info.release.iter().position('-');
+    #[cfg(target_os = "linux")]
+    IS_WINDOWS_SUBSYSTEM_FOR_LINUX.get_or_init(|| {
+        let mut info: libc::utsname = unsafe { mem::zeroed() };
```
krobelus, line 1659: BTW I always use `mem::zeroed()` to initialize C structs; there ought to be a better way, that hides the `unsafe`..

mqudsi: We could make a safe version of zeroed that asserts that the type does not have a destructor nor does it contain any types that have a destructor (recursively), via https://doc.rust-lang.org/std/mem/fn.needs_drop.html

I'm not absolutely certain that means a type is safe to initialize via `mem::zeroed()` but I think it'll be close enough to it for our purposes?

krobelus: great idea, I'll try that. It's not safe for NonZeroInt but we only want to use it for C structs

mqudsi: Ugh, it would need to be a macro because you can't use the outer generic type to make a const call, and the type would need to be specified explicitly whereas all our current `zeroed()` uses infer the type. Maybe not worth it.

fish-rust/src/common.rs
mqudsi: I think any use of `Lazy<Mutex<...>>` is suspect since Lazy does locking for you. I see an instance of that in common.rs for shell modes.

fish-rust/src/common.rs
Xiretza: There is no generic way to check if a type can be zero-initialized; if there was, `bytemuck` wouldn't have to manually implement its [`Zeroable`](https://docs.rs/bytemuck/latest/bytemuck/trait.Zeroable.html) trait for all zero-initializable std types.

mqudsi: Of course, this was merely an attempt to see if there was some sort of good-enough compromise.

tbh, I don't see a problem with `unsafe { std::mem::zeroed() }` but that was going to be my other suggestion here as well: we can add our own trait and implement it for specific libc types as needed if we really needed to cut down on that unsafe for whatever reason.

PR #9750, opened 2023-04-25
Remove kitty completion in favor of official integration

kidonng:
## Description

This completion is currently broken with latest kitty (`Error: Invalid shell state specification: fish2`), the latest version lives at https://github.com/kovidgoyal/kitty/blob/master/shell-integration/fish/vendor_completions.d/kitty.fish.

But now kitty's shell integration has been out for a long time, and enabled by default. This means vendoring the completion in fish itself has almost no benefit - it either gets ignored or conflicts with the current version. So I suggest just removing it outright.

## TODOs:


zanchey:
Yeah, makes sense.

PR #9753, opened 2023-04-25
complete: fix condition to suppress variable autocompletion

Xiretza:
This makes the code behave like the comment describes; if that's not actually the intended behaviour, I'll change the comment instead.


ridiculousfish:
Yep it's a good fix! The intention of this code is to not autosuggest with a bare $. Consider:

    echo $

fish has no idea what the user will type next, so it offers nothing at all. This happens to work with a single $, but the test fails for more than one:

    echo $PATH $

here fish autosuggests something (the first variable, alphabetically) and the reason is the bug you found. Very well spotted.

ridiculousfish:
I don't think this is worth a changelog but it's a good fix. Thank you!!

PR #9754, opened 2023-04-26
Riir iothreads

mqudsi:
Largely routine but for the trampolines in iothread.h and iothread.cpp which
    were a real PITA to get correct w/ all their variants.

    Integration is complete with all old code ripped out and the tests using the
    rust version of the code.


mqudsi:
All the memory leaks have finally been squashed. The logic in the new `iothread.h` and `callback.h` is more complicated than the entirety of the port, smh.

This memory leak seems unrelated to the changes in this particular port but perhaps comes from replacing previous calls to old cpp code with its rust equivalent:

```
Testing history path detection
Encountered 0 errors in low-level tests

=================================================================
==8865==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 32 byte(s) in 1 object(s) allocated from:
    #0 0x5596be12eae8 in __interceptor_calloc (/home/runner/work/fish-shell/fish-shell/build/fish_tests+0xba1ae8) (BuildId: 24621f9e17668547c64c974a09687a08a3a8a5e5)
    #1 0x7f1bb8b6acb2 in __cxa_thread_atexit_impl (/lib/x86_64-linux-gnu/libc.so.6+0x45cb2) (BuildId: 69389d485a9793dbe873f0ea2c93e02efaa9aa3d)
```

It seems like a false positive to me. I suppose we'll have to maintain a whitelist of memory errors for asan to ignore committed alongside the code?

mqudsi:
The code is ready for review but I'm documenting what I ran into below for posterity.

----

All threading is now done via rust, there is no pthread anything (other than the mutexes and cond variables, that is) left (directly visible, at any rate) in the codebase. All thread-management functions have also been moved over to rust, from setting up the thread ids, to spawning threads, to checking and asserting background/main thread status or forked/parent proc status.

---

For all existing C++ to rust interop that involved passing in callbacks for rust to call back later, we've been able to move the lambda to a standalone function, wrap its arguments or any closures into a `new`d struct, then call it from rust as a void function taking a `void *` (or `uint8_t *` to placate cxxbridge) parameter that we new'd before. For cleanup, the function invoked (the callback) would itself `delete (original_type_t *) param` before ending execution, guaranteeing that all resources were cleaned up.

This approach doesn't work for the `debounce_t` port because a) there are a lot of random callback lambdas, so it would be tedious to convert each to a standalone function and pull its closures and parameters into a struct, and more importantly, the callback might or might not end up actually getting executed. If we turned the lambdas into standalone functions, the `std::function` equivalent wouldn't be leaked (only because it would be code present in the text section of the binary forever) but the `new`d struct containing its closed over variables (as references, shared pointers, or `new`d variables) would be leaked because it would always be allocated from C++ at the time of *assigning* the callback via the rust ffi call, but the rust code may or may not call the function (since subsequent calls to  the same debouncer overwrite the callback and/or its args).

This necessitated moving away from the void C function callback model and moving to a proper RAII wrapper that contained the callback (lambda) to execute and its parameters/closed-over variables. But ffi can't do generic (since each lambda would be mangled to a different type) so there additionally needs to be type erasure involved to make sure that each such RAII wrapper can be executed with no external knowledge of its types.

The new header file `callback.h` includes this magic glue in the form of the C++ struct `callback_t` exposed to rust via the ffi module in `threads.rs` as `CppCallback`. It wraps a type-erased std::function that takes the place of the lambda, its closure, and optionally, its arguments. Importantly, this is passed to rust as a `SharedPtr<OpaqueCppType>` meaning that its destructor is called when rust is done with it. (I would have used `UniquePtr<CppCallback>` but cxxbridge can only pass opaque type wrappers as arguments by reference (via the pointer model), so in order to transfer ownership of the `shared_ptr<callback_t>` from C++ to Rust, C++ hangs on to its copy, passes it to rust by reference, then rust clones the `SharedPtr<CppCallback>` thereby incrementing the `shared_ptr`'s internal count just before C++ destructs its instance, decreasing the count. That kind of manipulation isn't possible with a `UniquePtr`.

The final bit is that `debounce_t` has the convenience function `perform_with_callback()` that queues for execution one callback on a background thread then passes its result to another callback which is queued for execution on a different thread. This means that the result of the first callback needs to be stored somewhere - however we can't just delete it when the second (internal) callback has completed evaluating because `debounce_t` also has a concept of a timeout and if the timeout has elapsed by the time the first callback was executed, the second callback is never called. Our type-erased callback entities for the first and second callback would be erased, but the temporarily allocated result to store the result of the first callback is allocated and never freed in this case.

As such, each `callback_t` also has a vector of "type-erased destructors" (an `std::function<void()>` instance that wraps the typed destructor call) that gets appended to and is executed when the `callback_t`/`CppCallback` instance is destroyed, regardless of whether or not the calback(s) were executed. So the first callback allocates temporary storage to hold its result then configures the second callback to free that allocated variable when its destructor is called.

Without this, ASAN would complain (deterministically but not identically reproducible) of memory leaks in the callback instance or callback state (without the `callback_t` black magic) for any callbacks that didn't get invoked and complain about the temporary parameter stored to chain the result of a callback performer to its main thread completion for cases where the first callback *was* executed but it timed out and didn't end up evaluating its completion. At least we know all the convoluted `debounce_t` logic involving overwriting subsequent invocations, timing out callbacks, and more is working correctly!

----

There appears to be a runtime-caused memory leak (as detailed in the previous comment) that ASAN began catching when all the thread and proc management code was switched over to rust. It appears to be separate from the debounce and iothread threadpool stuff and doesn't seem to originate in our code (for the ASAN CI run, we build the rust standard library itself instead of linking against it prebuilt, so any leaks in the runtime or standard library get reported against our code but the stack trace lets us see where it comes from). Since it seems to be a deterministic leak (just one individual allocation, always 32 bytes, always the same stack trace, always stemming from the call to `__cxa_thread_atexit_impl()`) it doesn't hurt us - as such, I've set up the CI to use a blacklist file that contains memory leaks ASAN should ignore (it's included in `./build_tools/asan_blacklist.txt`) and added a reference to this allocation/function.

fish-rust/src/threads.rs
```diff
@@ -22,6 +26,125 @@ static THREAD_ASSERTS_CFG_FOR_TESTING: AtomicBool = AtomicBool::new(false);
...
+        callback.invoke();
+    });
+}
+
+/// A [`ThreadPool`] work request.
+type WorkItem = Box<dyn FnOnce() + 'static + Send>;
+
+/// The queue of [`WorkItem`]s to be executed on the main thread.
+///
+/// If work is only ever enqueued from the main thread (I don't know?) this can be changed to use a
```
ridiculousfish, line 143: Work is enqueued from background threads. `Send` is correct here.

ridiculousfish (approved):
This is awesome! Very well done.

Issue #8845, opened 2022-03-31
 Prevent fish command from executing through fish_preexec


yonas:
fish, version 3.3.1
FreeBSD 13.0-RELEASE-p10  root@amd64-builder.daemonology.net:/usr/obj/usr/src/amd64.amd64/sys/GENERIC  amd64

Similar to #5068, this is a feature request. Can we add '--check-exit-code' flag - or any other name - so that:

`
function example --on-event fish_preexec --check-exit-code
`

would check that `example` completes with exit code zero.

Fish would run these functions asynchronously, but wait their completion and verify their exit code is zero before running the command.

faho:
Sorry, no. This isn't how events work currently, and I do not believe we want to add it.

Make your own "check-execute" function and bind it to enter.

Issue #5068, opened 2018-06-23
Prevent fish command from executing through fish_preexec


enricozb:
fish, version 2.6.0

I was wondering if there is a way to prevent a fish command from executing. I know that `fish_preexec` exists, but I can't seem to find a way to prevent the execution of a command through this function, without quitting the shell of course. After asking on stack overflow, the best answer that appeared was to bind the enter key, like this

```
function execute_ifnotls
    if test (commandline) = 'ls'
        echo
        echo "No"
        commandline ""
        commandline -f repaint
    else
        commandline -f execute
    end
end
bind \r execute_ifnotls
bind \n execute_ifnotls
```

However, this does not work with multiple lines.

Thanks for any help.

zanchey:
It works with multiple lines, but causes some visual glitches, at least for me on 2.7.1.

enricozb:
@zanchey If you type in `auto \` and enter, it goes onto a new line?

faho:
>If you type in auto \ and enter, it goes onto a new line?


@enricozb: For me it does, though I've not tested it in 2.6.0, only 2.7.1 and git master.

There is a bug related to multi-line commandlines, though. This bit:

```fish
if test (commandline) = 'ls'
```

if the commandline has multiple lines, e.g. it looks like

```fish
line1
line2
```

then commandline will print multiple lines, which the command substitution will turn into multiple arguments, so after expansion it looks like

```fish
if test line1 line2 = 'ls'
```

which will break `test` and print errors. If you wanted that `ls` test, you probably wanted to see either if the entire commandline was ls, in which case you would save it as a variable and double-quote it like

```fish
set -l cmd (commandline)
if test "$cmd" = ls
```

or you want to know if any of the lines starts with `ls`, in which case you'd use `string match` instead of `test` like

```fish
if string match -rq '^ls' -- (commandline)
```

-----

Anyway, that was @zanchey's simple example for a condition. The basic principle is sound. It's possible that fish 2.6.0 had a bug that we've fixed in 2.7.0, which means you want to upgrade.

mqudsi:
To answer the original question: no, there isn't because `fish_preexec` is an asynchronous event and not a callback. It is not awaited and the result of the function call is not evaluated.

enricozb:
@faho Thank you for the MWE. I'll update to 2.7.1. @mqudsi Thank you for the response to the question. As that answers my question, and since there has been an alternative solution shown, I'll close the issue.

PR #9761, opened 2023-04-28
Command blocking by preexec handler

mahdifrmz:
This PR gives a guard-like functionality to `fish_preexec` event handlers (like pre-push hooks for example). If any of the handlers exit with status code other that zero, then the command will not execute. This example shows how the handler can block the shutdown command

``` fish
function intercept --on-event fish_preexec
    if test $argv[1] = 'shutdown'
        return 1
    end
end
```

## Why adding this?

This is by no means a security feature to prevent certain commands being executed on the machine, but a handy command interceptor which can help to contain harmful effects of unintentional command execution. This is pretty common (at least for me) to just iterate over the shell history to find a command that had been run before and press enter on another command by mistake, which shouldn't have been executed (I have lost my work on a local repo several times just by rerunning `git restore .`)
This can also help to run tests and linting before running publish sub-command of package managers which might not support pre-publish hooks. I'm in need of such functionality badly :)

### DIFF
--- a/fish-rust/src/event.rs
+++ b/fish-rust/src/event.rs
@@ -83,13 +83,13 @@ mod event_ffi {
             parser: Pin<&mut parser_t>,
             name: &CxxWString,
             arguments: &CxxVector<wcharz_t>,
-        );
+        ) -> bool;
         #[cxx_name = "event_get_desc"]
         fn event_get_desc_ffi(parser: &parser_t, evt: &Event) -> UniquePtr<CxxWString>;
         #[cxx_name = "event_fire_delayed"]
         fn event_fire_delayed_ffi(parser: Pin<&mut parser_t>);
         #[cxx_name = "event_fire"]
-        fn event_fire_ffi(parser: Pin<&mut parser_t>, event: &Event);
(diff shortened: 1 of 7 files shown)


faho:
Sorry, no.

This isn't a feature we want. Not to mention that adding it in now will break every preexisting handler that happens to return something.

Duplicate of #5068

faho:
Here's what you can do:

```fish
function check
    if test "$(commandline)" = shutdown
        echo
        read --prompt-str "Are you sure (y/N)?" --nchars 1 -l response
        or begin
            commandline -f repaint
            return
        end
        test "$response" = y
        or begin
            commandline -f repaint
            return
        end
    end
    commandline -f repaint
    commandline -f execute
end

bind \r check
```

mahdifrmz:
@faho  Don't be sorry, that's pretty fine :)
Maybe I should've searched a little in the issue's list. Just wanted to give contributing to fish-shell a try.

And yes, character binding seems to be the solution for now.

I had the problem of breaking current handlers which return non-zero status codes in mind, and came up with printing a special string to stdout instead of the status code, or just using a `--check-exit-code` as suggested in #8845.

Anyways, thank you for your time.

PR #9763, opened 2023-04-29
Make (un)escape_string_xxx() portable and use rust version

mqudsi:
These are intended to have fixed, reproducible behavior but they were actually
    locale-sensitive and did not necessarily encode non-ASCII input as hex-encoded
    UTF-8 sequences (and instead used whatever encoding wctombb() returned).
    
    Additionally, the unescape functions had code that could not be reached (such as
    checking if c == '_' when c was not ASCII) and duplicate that wasn't needed
    (such as checking for '\0' before checking if a char is a hex digit). The new
    unescape functions might also be stricter in rejecting invalid input.
    
    The escape functions no longer convert the entirety of the wide input into a
    narrow string, thereby avoiding unnecessary allocations. It is similarly
    possible to decode directly into a WString result in the unescape functions, but
    it would increase branchiness and complexity of the code in order to handle
    sequences of 2, 4, or 8 hex characters per wide char instead of just appending
    all UTF-8 bytes and then widening the string at the end of the operation.

(Note that URL %-encoding with user's current mbstr encoding might produce URLs that won't even be valid to actually use if the the user's locale is neither CP1252 nor UTF-8.)

### DIFF
--- a/fish-rust/src/common.rs
+++ b/fish-rust/src/common.rs
@@ -324,17 +324,17 @@ fn escape_string_script(input: &wstr, flags: EscapeFlags) -> WString {
 /// (converted to UTF-8 then each byte encoded as a hex nibble preceded by a %).
 fn escape_string_url(input: &wstr) -> WString {
     let mut out = WString::with_capacity(input.len());
-    for c in input.as_char_slice() {
-        if c.is_ascii_alphanumeric() || ['/', '.', '~', '-', '_'].contains(c) {
+    for c in input.chars() {
+        if c.is_ascii_alphanumeric() || ['/', '.', '~', '-', '_'].contains(&c) {
             // The above characters don't need to be encoded.
-            out.push(*c);
+            out.push(c);
             continue;
         }
         // All other chars need to have their UTF-8 representation encoded in hex.
         // The maximum size any Unicode character can take in UTF-8 representation is four bytes.
         let mut scratch = [0u8; 4];
         for byte in c.encode_utf8(&mut scratch).as_bytes() {
-            out += &sprintf!(L!("%%%02X"), byte)[..];
+            sprintf!(=> &mut out, L!("%%%02X"), byte);
         }
     }
     out
@@ -398,7 +398,7 @@ fn escape_string_var(input: &wstr) -> WString {
             let mut scratch = [0u8; 4]; // Max length of any single UTF-8 sequence is 4 bytes.
             let utf8_str = c.encode_utf8(&mut scratch);
             for byte in utf8_str.as_bytes() {
-                out += &sprintf!(L!("%02X_"), byte)[..];
+                sprintf!(=> &mut out, L!("%02X_"), byte);
             }
             prev_was_hex = true;
         }
@@ -1177,7 +1177,7 @@ pub fn wcs2zstring(input: &wstr) -> CString {
 
 /// Like wcs2string, but appends to \p receiver instead of returning a new string.
 pub fn wcs2string_appending(output: &mut Vec<u8>, input: &wstr) {
-    output.reserve(input.len());
+    output.reserve(output.len() + input.len());
     wcs2string_callback(input, |buff| {
         output.extend_from_slice(buff);
         true
@@ -2173,16 +2173,23 @@ mod common_ffi {
         type escape_string_style_t = crate::ffi::escape_string_style_t;
     }
     extern "Rust" {
-        fn rust_unescape_string(
+        #[cxx_name = "rust_unescape_string"]
+        fn unescape_string_ffi(
             input: *const wchar_t,
             len: usize,
             escape_special: u32,
             style: escape_string_style_t,
         ) -> UniquePtr<CxxWString>;
+
+        #[cxx_name = "rust_escape_string_url"]
+        fn escape_string_url_ffi(input: *const wchar_t, len: usize) -> UniquePtr<CxxWString>;
+
+        #[cxx_name = "rust_escape_string_var"]
+        fn escape_string_var_ffi(input: *const wchar_t, len: usize) -> UniquePtr<CxxWString>;
     }
 }
 
-fn rust_unescape_string(
+fn unescape_string_ffi(
     input: *const ffi::wchar_t,
     len: usize,
     escape_special: u32,
@@ -2202,4 +2209,14 @@ fn rust_unescape_string(
         Some(result) => result.to_ffi(),
         None => UniquePtr::null(),
     }
+}
+
+fn escape_string_var_ffi(input: *const ffi::wchar_t, len: usize) -> UniquePtr<CxxWString> {
+    let input = unsafe { slice::from_raw_parts(input, len) };
+    escape_string_var(wstr::from_slice(input).unwrap()).to_ffi()
+}
+
+fn escape_string_url_ffi(input: *const ffi::wchar_t, len: usize) -> UniquePtr<CxxWString> {
+    let input = unsafe { slice::from_raw_parts(input, len) };
+    escape_string_url(wstr::from_slice(input).unwrap()).to_ffi()
 }
--- a/fish-rust/src/common.rs
+++ b/fish-rust/src/common.rs
@@ -318,55 +318,142 @@ fn escape_string_script(input: &wstr, flags: EscapeFlags) -> WString {
     out
 }
 
-/// Escape a string in a fashion suitable for using as a URL. Store the result in out_str.
-#[widestrs]
+/// Escape a string in a fashion suitable for using as a URL.
+///
+/// Url-safe ASCII input chars are passed through while everything else is UTF-8 percent-encoded
+/// (converted to UTF-8 then each byte encoded as a hex nibble preceded by a %).
 fn escape_string_url(input: &wstr) -> WString {
-    let narrow = wcs2string(input);
-    let mut out = WString::new();
-    for byte in narrow.into_iter() {
-        if (byte & 0x80) == 0 {
-            let c = char::from_u32(u32::from(byte)).unwrap();
-            if c.is_alphanumeric() || [b'/', b'.', b'~', b'-', b'_'].contains(&byte) {
-                // The above characters don't need to be encoded.
-                out.push(c);
-                continue;
-            }
+    let mut out = WString::with_capacity(input.len());
+    for c in input.as_char_slice() {
+        if c.is_ascii_alphanumeric() || ['/', '.', '~', '-', '_'].contains(c) {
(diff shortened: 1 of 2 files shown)


fish-rust/src/common.rs
```diff
@@ -318,57 +318,146 @@ fn escape_string_script(input: &wstr, flags: EscapeFlags) -> WString {
...
-                    out.push('_');
-                    prev_was_hex_encoded = false;
-                }
-                out.push(c);
-                continue;
-            }
-        } else if byte == b'_' {
-            // Underscores are encoded by doubling them.
-            out += "__"L;
-            prev_was_hex_encoded = false;
```
mqudsi, line 362: Unless I'm missing something, this was unreachable.

fish-rust/src/common.rs
```diff
@@ -687,80 +776,6 @@ fn unescape_string_internal(input: &wstr, flags: UnescapeFlags) -> Option<WStrin
...
-        if c > '\u{7F}' {
-            return None; // invalid character means we can't decode the string
-        }
-        if c == '%' {
-            let c1 = input.char_at(i + 1);
-            if c1 == '\0' {
-                return None;
-            } else if c1 == '%' {
-                result.push(b'%');
-                i += 1;
```
mqudsi, line 706: This doesn't correspond to the escape code - a `%` is encoded as `%25` and not `%%`.

fish-rust/src/common.rs
```diff
@@ -318,57 +318,144 @@ fn escape_string_script(input: &wstr, flags: EscapeFlags) -> WString {
...
 }
 
-/// Escape a string in a fashion suitable for using as a URL. Store the result in out_str.
-#[widestrs]
+/// Escape a string in a fashion suitable for using as a URL.
+///
+/// Url-safe ASCII input chars are passed through while everything else is UTF-8 percent-encoded
+/// (converted to UTF-8 then each byte encoded as a hex nibble preceded by a %).
 fn escape_string_url(input: &wstr) -> WString {
-    let narrow = wcs2string(input);
```
mqudsi, line 324: It's not necessary to narrow the whole string upfront.

fish-rust/src/common.rs
```diff
@@ -318,57 +318,144 @@ fn escape_string_script(input: &wstr, flags: EscapeFlags) -> WString {
...
+
+/// Escape a string in a fashion suitable for using as a fish variable name.
+///
+/// Non-ASCII characters are encoded as UTF-8 and written one hex nibble at a time, preceded and
+/// followed by an underscore. ASCII characters are passed through as-is except if they could be
+/// confused with an encoded UTF-8 hex sequence (0-9 or capital A-F after an underscore), in which
+/// case they too are encoded.
 fn escape_string_var(input: &wstr) -> WString {
-    let mut prev_was_hex_encoded = false;
-    let narrow = wcs2string(input);
```
mqudsi, line 345: It's not necessary to narrow the whole string upfront.

fish-rust/src/common.rs
```diff
@@ -318,57 +318,144 @@ fn escape_string_script(input: &wstr, flags: EscapeFlags) -> WString {
...
+///
+/// Url-safe ASCII input chars are passed through while everything else is UTF-8 percent-encoded
+/// (converted to UTF-8 then each byte encoded as a hex nibble preceded by a %).
 fn escape_string_url(input: &wstr) -> WString {
-    let narrow = wcs2string(input);
-    let mut out = WString::new();
-    for byte in narrow.into_iter() {
-        if (byte & 0x80) == 0 {
-            let c = char::from_u32(u32::from(byte)).unwrap();
-            if c.is_alphanumeric() || [b'/', b'.', b'~', b'-', b'_'].contains(&byte) {
```
mqudsi, line 329: The first equivalent of "is ascii" check is not needed since we can check directly for the chars we want.

fish-rust/src/common.rs
```diff
@@ -318,57 +318,144 @@ fn escape_string_script(input: &wstr, flags: EscapeFlags) -> WString {
...
+                prev_was_hex = true;
+                continue;
+            } else if prev_was_hex && c1.is_ascii() {
+                // We insert an underscore after a hex sequence to improve readability, so it's
+                // normal to get a non-hex character here.
+                result.push(c1 as u8);
+            } else {
+                // All characters must be ASCII and an underscore is only allowed before an escape.
+                return None;
+            }
```
mqudsi, line 447: I think this else block might not have been caught/enforced previously?

fish-rust/src/common.rs
```diff
@@ -318,57 +318,144 @@ fn escape_string_script(input: &wstr, flags: EscapeFlags) -> WString {
...
+            // The above characters don't need to be encoded.
+            out.push(*c);
+            continue;
         }
         // All other chars need to have their UTF-8 representation encoded in hex.
-        out += &sprintf!("%%%02X"L, byte)[..];
+        // The maximum size any Unicode character can take in UTF-8 representation is four bytes.
+        let mut scratch = [0u8; 4];
+        for byte in c.encode_utf8(&mut scratch).as_bytes() {
+            out += &sprintf!(L!("%%%02X"), byte)[..];
```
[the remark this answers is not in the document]
mqudsi: I was very much tempted to use native rust to format then manually widen the result to avoid an allocation - thanks for making that change!

fish-rust/src/common.rs
```diff
@@ -318,57 +318,144 @@ fn escape_string_script(input: &wstr, flags: EscapeFlags) -> WString {
...
-                out.push(c);
-                continue;
-            }
-        } else if byte == b'_' {
-            // Underscores are encoded by doubling them.
-            out += "__"L;
-            prev_was_hex_encoded = false;
+    let mut prev_was_hex = false;
+    let mut out = WString::with_capacity(input.len() + 8);
+    for c in input.as_char_slice() {
```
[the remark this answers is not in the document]
mqudsi: Thanks; I'll change it.

I could have sworn that's what I tried but either the compiler or rust-analyzer didn't like it for wsr/WString.

ridiculousfish (changes requested):
This changes the behavior from its current locale-sensitive encoding to UTF-8. Is there a reason to make such a change? It is potentially breaking.

mqudsi:
There are a few issues I could see:

For variable and url escaping: 
* the previous approach resulted in non-stable output, meaning if someone used `fish escape --style=var/url` on their end then copy-and-pasted the result into a script, unescape might fail when that script is executed on another PC, or if the user used it on their own PC but tried to unescape the result with a different locale, it would fail.
* the code can't handle special characters with an ascii or other restricted locale, which seems like it would actually be one of the very reasons to escape things, e.g. trying to unescape with locale set to C
* a modern cli app that *isn't* locale aware (and few are, these days) is probably going to generate non-ascii output as UTF-8, so if you try to escape the output of an external command but your locale is missing the characters/codepoints in question, escaping then unescaping will be a lossy operation. (Same goes for extracting text content from a file and then escaping the result)

For url escaping:
* The biggest issue is that it's extremely unlikely to be able to actually *use* the escaped URL if your locale isn't UTF-8. It's fine (except for the issues above) when you're using fish to both escape and unescape the URL, but if you try to `curl` the resulting escaped URL or copy-and-paste it into your web browser, the server is going to (with 99% probability) use UTF-8 to decode the hex-escaped URL sequences and it'll either be an invalid result or not the intended result. That's probably the main usage of `string escape --style=url` since I don't think anyone is going to be implementing both the http client *and* the http server in fish, so it's important to use the standardized encoding to generate the escaped url.

mqudsi:
Actually, you can ignore what I said above because it's besides the point: this is only a breaking change *for the cases that were already broken*. 

i.e. transiently escaped then unescaped contetn in the same session that worked fine before will continue to work fine (because before the same locale was used to encode and decode, assuming it didn't change mid-session, and now a different but fixed locale is used to encode and decode). 

The only cases that *will* break are cases where the output of `string escape` is persisted beyond a single fish session, meaning these were cases where the user could have changed locales between escaping and unescaping,  *or* the escaped content was being sent to another machine for unescaping and previously there was no guarantee that locales would line up on both ends. 

There is nothing I can think of that will break if we change to fixed UTF-8 that wasn't already possibly broken.

ridiculousfish:
Yes, these functions are broken and their behavior makes no sense. But that doesn't matter now: the goal of the rewrite is to preserve existing behaviors.

> The only cases that will break are cases where the output of string escape is persisted beyond a single fish session 

That's more than enough reason to avoid this. For example, abbreviations stored in universal variables use the "variable" escape style (perhaps the only usage). This change therefore could potentially break abbreviations.

IMO these functions are not worth "fixing." We should instead deprecate them. But if we do decide to fix them, it will be a breaking change, and that should be a separate effort from the Rust port.

mqudsi:
I'm fine with tabling this until after the port but my point is that starting fish *today* with a different locale will cause the abbreviations to fail to load since the unescape no longer lines up. As it's already broken, this can only help.

faho:
>starting fish today with a different locale will cause the abbreviations to fail to load since the unescape no longer lines up.

That's just for the *variables*, which are deprecated.

>I'm also unclear on how we could drop them since they serve a real purpose (even if we end up not using them internally).

We do expose all of these escaping styles, via `string escape`.

The URL-escaping specifically is used in [our OSC7 cwd-reporting](https://github.com/fish-shell/fish-shell/blob/21e31c9b595256ea08715a5a7f37d83b78efc764/share/functions/__fish_config_interactive.fish#L277), and a few other escapes are also defined as taking URLs, including OSC8 for hyperlinks and [iTerm's RemoteHost escape](https://iterm2.com/documentation-escape-codes.html). I don't believe there is a good generally available portable command alternative for this.

`regex` escaping is used for `string match` and `string replace`, and is *necessary* if you want to match a dynamic string properly - otherwise if you got `foo.*` it would suddenly be treated as a regex! (an alternative design would allow placeholders in the regex)

`script` escaping is needed for `eval` and friends, but also e.g. for `fish_command_not_found`'s output - so we can tell you "that's the command that failed" without it bleeding over into the containing string.

We currently use `var` escaping only in the git-completions, to cache git aliases. Before we used to use it in the abbr function (and we still do in the importing code for old-style variables). All of this is a workaround for dictionaries.

I don't believe we can remove or deprecate any of these. Var-style is closest, but it's still an awkward compatibility break and I do not believe it really hurts.

mqudsi:
> We do expose all of these escaping styles, via `string escape`.

That's the point I was trying to make. We expose them for a reason and people (us and others) use them for good reason.

faho:
I'm closing this because it has conflicts and I'm not quite sure what the current state is.

PR #9764, opened 2023-04-29
Avoid ASAN false positives caused by TLS variables

mqudsi:
The previous commit to blacklist the glibc function `__cxa_thread_atexit_impl()` didn't actually work due to ASAN limitations on what functions can be blacklisted/ignored (they have to be at the top of the stack and since we don't compile gnulibc from source, we can't blacklist that function) - it's just that the false positive is non-deterministic (seeming to depend on background threads being alive when main() exits).

    ASAN can report spurious memory leaks that are apparently caused by main()
    exiting while there are still background threads that haven't exited. This is OK
    - they'll be killed and their memory will be reclaimed - but causes ASAN to
    report false positives for memory leaks originating in the libc runtime.
    
    Some research shows that rust internally uses the weakly-linked glibc routine
    __cxa_thread_atexit_impl() to register destructors for TLS variables to run when
    a thread has exited. It seems that in addition to memory consumed by TLS
    variables not being freed if background threads are summarily executed, the init
    work done in __cxa_thread_atexit_impl() causes additional temporary memory to be
    allocated and won't be freed until the thread exits normally.
    
    If compiled with the asan feature enabled, we now track all extant thread join
    handles and will do our best to join them before our main process exits by
    calling asan_before_exit() in exit_without_destructors(). This function is a
    no-op in the regular case (when the code isn't compiled with the asan feature
    enabled).

and

    This pulls in the call to asan_before_exit() everywhere the C++ code calls
    exit_without_destructors(). It is a no-op if not compiled with the asan feature
    and called from the parent fish process.
    
    fish_tests.cpp didn't use exit_without_destructors (probably also to avoid
    memory leak reports) so we insert a manual call into asan_before_exit() just
    before the end of main() there.

### DIFF
--- a/cmake/Tests.cmake
+++ b/cmake/Tests.cmake
@@ -166,6 +166,12 @@ endforeach(CHECK)
 FILE(GLOB PEXPECTS CONFIGURE_DEPENDS ${CMAKE_SOURCE_DIR}/tests/pexpects/*.py)
 foreach(PEXPECT ${PEXPECTS})
   get_filename_component(PEXPECT ${PEXPECT} NAME)
+  if(DEFINED ASAN)
+    if("${PEXPECT}" STREQUAL "exit_handlers.py")
+      # This times out due to changes to fish's exit behavior under ASAN
+      continue()
+    endif()
+  endif()
   add_test(NAME ${PEXPECT}
     COMMAND ${CMAKE_SKIPPED_HACK} sh ${CMAKE_CURRENT_BINARY_DIR}/tests/test_driver.sh
       ${CMAKE_CURRENT_BINARY_DIR}/tests/interactive.fish ${PEXPECT}
--- a/fish-rust/src/threads.rs
+++ b/fish-rust/src/threads.rs
@@ -49,6 +49,21 @@ static NOTIFY_SIGNALLER: once_cell::sync::Lazy<&'static crate::fd_monitor::FdEve
         result
     });
 
+/// A tuple of `(ThreadId, JoinHandle)` of all threads we've (directly) spawned that haven't yet
+/// exited.
+///
+/// Under glibc 2.18+, rust uses the weakly linked symbol __cxa_thread_atexit_impl to register
+/// destructors for thread local variables (TLS is used by some rust std::thread features).
+/// If remaining background threads aren't joined before the main process exits, ASAN can report a
+/// memory leak with the stack trace originating from __cxa_thread_atexit_impl(). Unfortunately, the
+/// -fsanitize-blacklist/-fsanitize-ignorelist compiler option can only be used to suppress the
+/// top-most stack frame and since we do not compile glibc from scratch under ASAN, we can't
+/// suppress the reported memory leak.
+///
+/// We can instead ensure that all live threads are joined before exiting when running under ASAN.
+#[cfg(feature = "asan")]
+static THREAD_HANDLES: Mutex<Vec<(ThreadId, std::thread::JoinHandle<()>)>> = Mutex::new(Vec::new());
+
 #[cxx::bridge]
 mod ffi {
     unsafe extern "C++" {
@@ -75,6 +90,7 @@ mod ffi {
     extern "Rust" {
         #[cxx_name = "make_detached_pthread"]
         fn spawn_ffi(callback: &SharedPtr<CppCallback>) -> bool;
+        fn asan_before_exit();
     }
 
     extern "Rust" {
@@ -265,12 +281,34 @@ pub fn spawn<F: FnOnce() + Send + 'static>(callback: F) -> bool {
     // We don't have to port the PTHREAD_CREATE_DETACHED logic. Rust threads are detached
     // automatically if the returned join handle is dropped.
 
-    let result = match std::thread::Builder::new().spawn(callback) {
+    let result = match std::thread::Builder::new().spawn(move || {
+        (callback)();
+        #[cfg(feature = "asan")]
+        {
+            let thread_id = std::thread::current().id();
+            let mut handles = THREAD_HANDLES.lock().expect("Mutex poisoned!");
+            // We can't guarantee that the parent thread was scheduled after we ran, so the handle might
+            // not be in the vector.
+            if let Some(idx) = handles.iter().rev().position(|h| h.0 == thread_id) {
+                handles.swap_remove(idx);
+            }
+        }
+    }) {
         Ok(handle) => {
-            let id = handle.thread().id();
-            FLOG!(iothread, "rust thread", id, "spawned");
-            // Drop the handle to detach the thread
-            drop(handle);
+            let thread_id = handle.thread().id();
+            FLOG!(iothread, "rust thread", thread_id, "spawned");
+            #[cfg(feature = "asan")]
+            {
+                let mut handles = THREAD_HANDLES.lock().expect("Mutex poisoned!");
+                if !handle.is_finished() {
+                    handles.push((thread_id, handle));
+                }
+            }
+            #[cfg(not(feature = "asan"))]
+            {
+                // Drop the handle to detach the thread
+                drop(handle);
+            }
             true
         }
         Err(e) => {
@@ -297,6 +335,28 @@ fn spawn_ffi(callback: &cxx::SharedPtr<ffi::CppCallback>) -> bool {
     spawn(move || {
         callback.invoke();
     })
+}
+
+/// When running under ASAN, join child threads before exiting the parent process to avoid spurious
+/// memory leak warnings regarding TLS variables.
+///
+/// This function is always defined but is a no-op if not running under ASAN. This is to make it
+/// more ergonomic to call it where needed in general and to make it possible via ffi at all.
+pub fn asan_before_exit() {
+    #[cfg(feature = "asan")]
+    if !is_forked_child() && is_main_thread() {
+        // We can't keep THREAD_HANDLES locked as threads acquire it before exiting themselves.
+        // A child thread can spawn additional threads, so we loop until no threads remain.
+        loop {
+            let mut handles = std::mem::take(&mut *THREAD_HANDLES.lock().expect("Mutex poisoned!"));
+            if handles.is_empty() {
+                break;
+            }
+            for h in handles.drain(..) {
+                let _ = h.1.join();
+            }
+        }
+    }
 }
 
 /// Data shared between the thread pool [`ThreadPool`] and worker threads [`WorkerThread`].
--- a/cmake/Rust.cmake
+++ b/cmake/Rust.cmake
@@ -25,6 +25,7 @@ set(fish_rust_target "fish-rust")
 
 set(fish_autocxx_gen_dir "${CMAKE_BINARY_DIR}/fish-autocxx-gen/")
 
+set(FISH_CRATE_FEATURES "fish-ffi-tests")
 if(NOT DEFINED CARGO_FLAGS)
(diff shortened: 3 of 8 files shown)


ridiculousfish:
Interesting finding. Is there a specific TLS variable that's triggering this? We should work hard to avoid destructors in TLS variables - we should limit TLS to simple scalars like the thread ID. But it's possible that Rust creates this unconditionally.

mqudsi:
[It's unconditional](https://github.com/rust-lang/rust/blob/eb62877597000ccf8bb99ab131b5977344afdfa3/library/std/src/sys_common/thread_info.rs#L8-L13) and it appears to be used to [track the thread id, thread name, and an optional stack guard](https://github.com/rust-lang/rust/blob/eb62877597000ccf8bb99ab131b5977344afdfa3/library/std/src/thread/mod.rs#L1148-L1152).

I think it has a destructor because these are placed in an Arc to allow access to them from the main thread.

ASAN CI just happened to still run into the same leak after these changes so I'm pushing some test commits to debug what's going on.

cmake/Tests.cmake
```diff
@@ -166,6 +166,12 @@ endforeach(CHECK)
 FILE(GLOB PEXPECTS CONFIGURE_DEPENDS ${CMAKE_SOURCE_DIR}/tests/pexpects/*.py)
 foreach(PEXPECT ${PEXPECTS})
   get_filename_component(PEXPECT ${PEXPECT} NAME)
+  if(DEFINED ASAN)
+    if("${PEXPECT}" STREQUAL "exit_handlers.py")
```
ridiculousfish, line 170: Hmm, this was working before. If this is now broken under ASAN, we should fix it. Any idea what changed? Maybe we should bisect.

fish-rust/src/threads.rs
```diff
@@ -49,6 +49,21 @@ static NOTIFY_SIGNALLER: once_cell::sync::Lazy<&'static crate::fd_monitor::FdEve
         result
     });
 
+/// A tuple of `(ThreadId, JoinHandle)` of all threads we've (directly) spawned that haven't yet
```
ridiculousfish, line 52: This contains a LOT of code to work around an ASAN limitation. I'd be OK with a line or two, but don't think we should do something so invasive. I'd rather just disable ASAN's leak detection than complicate our own code in this way.

ridiculousfish (changes requested):

mqudsi:
It seems waiting for threads to exit took care of that leak because it would be reported in a `fish_tests` run of the  `history_paths` test and that stopped happening for about the dozen or so last commits I force-pushed.

I also pushed another attempt at blacklisting (hopefully just these particular) TLS leaks but it's just a Hail Mary since while these very much *are* the functions that call into `__cxa_at_threadexit_impl()`, ASAN for some reason doesn't report them in its memory leak report's stack trace - and you'd think ASAN can only ignore what it can see. But these are the two call sites (one for tls init upon spawn and the other for tls init upon accessing the current thread) in the standard library itself and I have configured the ASAN CI job to use `cargo build -Z build-std` so there's a tiny chance it'll work.

I did have to skip one test under the new ASAN conditions - the pexpect test `exit_handlers.py` would throw an exception on the call to `wait()` with the `force` parameter set to `false` complaining that the child process (fish) could not be killed.

ridiculousfish:
Is this happening in fish itself or just the tests? If it's just the tests perhaps we should switch from returning from main normally, to using `_exit` (i.e. `exit_without_destructors`), if that suppresses the bogus leak report.

mqudsi:
I've only ever seen ASAN complain about it during individual executions of `fish_tests` entrypoint/executable, but I cannot rationalize why that would actually be the case so I won't commit to saying it couldn't happen in a regular fish execution.

The leaked memory [is allocated in the call to `__cxa_thread_atexit_impl()`](https://github.com/bminor/glibc/blob/a443bd3fb233186038b8b483959ecb7978d1abea/stdlib/cxa_thread_atexit_impl.c#L106) which is made in the middle of regular program/test execution, and that allocated memory is then presumably freed in the fn pointer that gets actually called by the runtime/pthreads upon thread exit. (EDIT: [yes, that's what happens](https://github.com/bminor/glibc/blob/a443bd3fb233186038b8b483959ecb7978d1abea/stdlib/cxa_thread_atexit_impl.c#L163).)

*Logically*, skipping C++'s process teardown should only aggravate the situation and not make it better (I'm presuming of course that in the course of normal teardown `_exit()` is called in addition to other things). But maybe _exit() doesn't always get called in the normal course of business? 

gdb shows a regular fish_tests run ends up calling into the following:

```
#0  __GI__exit (status=status@entry=0) at ../sysdeps/unix/sysv/linux/_exit.c:27
#1  0x00007ffff7dac9c2 in __run_exit_handlers (status=0, listp=<optimized out>, run_list_atexit=run_list_atexit@entry=true, run_dtors=run_dtors@entry=true) at exit.c:132
#2  0x00007ffff7daca60 in __GI_exit (status=<optimized out>) at exit.c:139
#3  0x00007ffff7d8a08a in __libc_start_main (main=0x244fc0 <main(int, char**)>, argc=2, argv=0x7fffffffe4c8, init=<optimized out>, fini=<optimized out>, rtld_fini=<optimized out>, stack_end=0x7fffffffe4b8)
    at ../csu/libc-start.c:342
#4  0x000000000020fc5e in _start ()
```

`__run_exit_handlers()` is where [the call to `__call_tls_dtors()` is made](https://github.com/bminor/glibc/blob/a443bd3fb233186038b8b483959ecb7978d1abea/stdlib/exit.c#L43-L44) (and `__run_exit_handlers()` **is** called with `run_dtors=true` - at least on my local machine.

As for  `_exit()` - it skips all that and basically [just calls `abort()`](https://github.com/bminor/glibc/blob/a443bd3fb233186038b8b483959ecb7978d1abea/posix/_exit.c#LL25C6-L25C6) - and abort pretty much just, well, [aborts](https://github.com/bminor/glibc/blob/a443bd3fb233186038b8b483959ecb7978d1abea/stdlib/abort.c#L48) so there's no way that the memory assigned in `__cxa_thread_atexit_impl()` gets freed with that approach. From a brief glance, the main thread's tls_dtor is statically allocated and wouldn't normally leak, but any call to `__cxa_thread_atexit_impl()` is supposed to allocate and prepend to that linked list, so realistically we *should* be seeing ASAN leak warnings for regular fish execution (since it never calls `exit()` but calls `_exit()` instead).

This is all complicated considerably by the fact that I've never observed an ASAN leak locally - only under CI. And my glibc version doesn't match CI's, among any of a thousand other confounding variables.

cmake/Tests.cmake
mqudsi: I think it breaks because the main thread delays until all child threads have exited. The python `spawn.wait()` call only gives a process 0.1 secs after SIGHUP to exit. Maybe we can avoid calling `asan_before_exit()` if not exiting normally.

ridiculousfish:
`_exit` does not call `abort()`, which is good because `abort()` does a bunch of stuff that we don't want to do. `_exit` simply wraps the syscall - [this is the real glibc implementation](https://github.com/bminor/glibc/blob/master/sysdeps/unix/sysv/linux/_exit.c).

All the memory allocated by the process will be freed after a call to `_exit`, including anything in `__cxa_thread_atexit_impl()`. There's only "leaks" in the sense of memory that was not manually freed; that memory is instead simply freed by the OS. Manually freeing memory just slows down exit, so why bother.

We've had destructor ordering problems in the past in C++, which are only a fake test problem, since as you observe the real fish exits with `_exit`. Probably the tests should exit in the same way. Destructor ordering and fake "leaks" of TLS data are not real problems.

mqudsi:
Huh, that's odd. I wonder what the `_exit()` I linked to is, then.

I understand that all memory is freed when the process is reclaimed by the kernel - but that reclamation isn't taken into account by valgrind or asan. I'm trying to understand why `_exit()` - which definitively does *not* run the tls destructors nor free the memory that ASAN is complaining about - somehow gets a pass from ASAN while calling `exit()` normally doesn't.

And using `_exit()` does indeed seem to work - [I reran the asan job three times](https://github.com/mqudsi/fish-shell/actions/runs/4841452589) in this alternative branch [where the only thing I did](https://github.com/fish-shell/fish-shell/compare/master...mqudsi:fish-shell:asan_fix3) was use `_exit()` at the end of `fish_tests.cpp`'s `main()` and it passed each time. It could still be a fluke, but I've not been that lucky with the CI ASAN runs.

cmake/Tests.cmake
ridiculousfish: Let's get rid of `asan_before_exit()`. ASAN is not telling us anything interesting here; fish does not try to deallocate all of its memory before exit.

ridiculousfish:
`exit()` runs atexit handlers, `_exit()` does not. That probably explains the ASAN differences.

mqudsi:
But the memory isn’t allocated in the atexit handler - that’s where it’s freed!

You can see why I’m confused. The same memory is allocated in both cases, but asan (sometimes) complains when the one that *does* free is called but doesn’t *appear* to complain when the one that *doesn’t* free is called.

mqudsi:
It’s obvious - I suspect calling _exit() cuts asan itself short. It doesn’t get a chance to dump its findings of whether or not anything leaked in the main thread’s own lifetime/scope. 

It registers its own thread exit handlers - unlike valgrid, it’s not running in a separate process. So if we _exit we aren’t even getting the actual asan exit status. ASAN *isn’t* giving us a pass if we use _exit - we are deciding we don’t want to hear from asan in that case!

ridiculousfish:
Yep that's right.

We should use ASAN leak detection to the extent it is helpful, but we should not contort ourselves to work around its bugs or Rust incompatibilities. A line or two is OK, and if we can support LeakSanitizer with some minor changes, then let's do it. But keeping a separate vector of thread handles, or making fish wait around extra long when it wants to exit, are examples of contortions which IMO go too far.

mqudsi:
> or making fish wait around extra long when it wants to exit, are examples of contortions which IMO go too far.

I just want to seek clarification here - this only happens when running ASAN (it and the vector of thread handles are conditionally compiled). Are you saying main waiting for threads to exit *only when running under ASAN CI* is too intrusive?

ridiculousfish:
Yes, it is my opinion that we should not wait for threads to exit in fish's main(), because they may not exit.

1. A thread may be hung in I/O, such as talking to an unresponsive NFS hard mount 
2. We have various threads that deliberately "wait around" for a bit (e.g. iothreads, the fd monitor thread), these may not exit in a timely fashion.
3. In the future we are likely to have background threads that are executing fish script, and these may also hang for various reasons (though we haven't sorted out the signal handling story for these yet).

Basically we can't guarantee that all threads will exit, so we shouldn't wait for them, even in ASAN.

mqudsi:
Ok, sounds good.  I’ve been exploring different options in all cases. 

Something that’s nagging at me is that there’s only one leak report per run despite many threads being launched making me think it might be the main thread onexit that isn’t being called - but then I get no leaks with this PR which doesn’t do anything special for the main thread itself so I don’t know. 

I want to see if I can manually call the thread dtor then _exit to pin down the problem further. 

Also, I think I may have found a legit memory leak in read.fish when I changed fish.cpp to use exit() under ASAN but I’ll dig into that later.

zanchey:
@mqudsi, I think we can close this?

mqudsi:
Yessir. Thanks for the ping.

PR #9765, opened 2023-04-29
builtins/printf: avoid string copies by formatting directly to buffer

Xiretza:


fish-rust/src/builtins/printf.rs
```diff
@@ -262,21 +262,22 @@ impl<'a, 'b> builtin_printf_state_t<'a, 'b> {
...
-                    sprintf_locale(
-                        &mut target,
-                        $fmt,
-                        &self.locale,
-                        &[$($arg.to_arg()),*]
-                    );
-                    target
+                // Don't output if we're done.
+                if self.early_exit {
+                    return;
```
krobelus, line 272: should probably not return from the surrounding function, only skip the rest of the macro

krobelus:
good idea. In future we want the same for `OutputStream`, maybe we can define a trait

fish-rust/src/builtins/printf.rs
Xiretza: Whoops, absolutely right - the danger of macros...

Xiretza:
I think `printf_compat::output::WideWrite` is that trait, just needs to be implemented for `OutputStream`.

PR #9767, opened 2023-04-30
Port most of env to Rust

ridiculousfish:
This reimplements variables in Rust (`env_var_t` -> `EnvVar`) and also the environment stack, and other environments.

We retain the C++ classes, but both now wrap Rust implementations.

There's still a lot of initialization code in env.cpp which will be migrated to Rust later, but this is a good milestone.

### DIFF
--- a/fish-rust/src/env/env_ffi.rs
+++ b/fish-rust/src/env/env_ffi.rs
@@ -1,0 +1,119 @@
+use super::var::{EnvVar, EnvVarFlags};
+use crate::ffi::{wchar_t, wcharz_t, wcstring_list_ffi_t};
+use crate::wchar_ffi::WCharToFFI;
+use crate::wchar_ffi::{AsWstr, WCharFromFFI};
+use cxx::{CxxWString, UniquePtr};
+use std::pin::Pin;
+
+#[allow(clippy::module_inception)]
+#[cxx::bridge]
+mod env_ffi {
+    /// Return values for `EnvStack::set()`.
+    #[repr(u8)]
+    #[cxx_name = "env_stack_set_result_t"]
+    enum EnvStackSetResult {
+        ENV_OK,
+        ENV_PERM,
+        ENV_SCOPE,
+        ENV_INVALID,
+        ENV_NOT_FOUND,
+    }
+
+    extern "C++" {
+        include!("wutil.h");
+        type wcstring_list_ffi_t = super::wcstring_list_ffi_t;
+        type wcharz_t = super::wcharz_t;
+    }
+
+    extern "Rust" {
+        type EnvVar;
+
+        fn is_empty(&self) -> bool;
+
+        fn exports(&self) -> bool;
+        fn is_read_only(&self) -> bool;
+        fn is_pathvar(&self) -> bool;
+
+        #[cxx_name = "equals"]
+        fn equals_ffi(&self, rhs: &EnvVar) -> bool;
+
+        #[cxx_name = "as_string"]
+        fn as_string_ffi(&self) -> UniquePtr<CxxWString>;
+
+        #[cxx_name = "as_list"]
+        fn as_list_ffi(&self) -> UniquePtr<wcstring_list_ffi_t>;
+
+        #[cxx_name = "to_list"]
+        fn to_list_ffi(&self, out: Pin<&mut wcstring_list_ffi_t>);
+
+        #[cxx_name = "get_delimiter"]
+        fn get_delimiter_ffi(&self) -> wchar_t;
+
+        #[cxx_name = "get_flags"]
+        fn get_flags_ffi(&self) -> u8;
+
+        #[cxx_name = "clone_box"]
+        fn clone_box_ffi(&self) -> Box<EnvVar>;
+
+        #[cxx_name = "env_var_create"]
+        fn env_var_create_ffi(vals: &wcstring_list_ffi_t, flags: u8) -> Box<EnvVar>;
+
+        #[cxx_name = "env_var_create_from_name"]
+        fn env_var_create_from_name_ffi(
+            name: wcharz_t,
+            values: &wcstring_list_ffi_t,
+        ) -> Box<EnvVar>;
+    }
+}
+pub use env_ffi::EnvStackSetResult;
+
+impl Default for EnvStackSetResult {
+    fn default() -> Self {
+        EnvStackSetResult::ENV_OK
+    }
+}
+
+/// FFI bits.
+impl EnvVar {
+    pub fn equals_ffi(&self, rhs: &EnvVar) -> bool {
+        self == rhs
+    }
+
+    pub fn as_string_ffi(&self) -> UniquePtr<CxxWString> {
+        self.as_string().to_ffi()
+    }
+
+    pub fn as_list_ffi(&self) -> UniquePtr<wcstring_list_ffi_t> {
+        self.as_list().to_ffi()
+    }
+
+    pub fn to_list_ffi(&self, mut out: Pin<&mut wcstring_list_ffi_t>) {
+        out.as_mut().clear();
+        for val in self.as_list() {
+            out.as_mut().push(val);
+        }
+    }
+
+    pub fn clone_box_ffi(&self) -> Box<Self> {
+        Box::new(self.clone())
+    }
+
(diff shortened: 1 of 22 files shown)


fish-rust/src/env/env_ffi.rs
```diff
@@ -0,0 +1,368 @@
...
+        ) -> EnvStackSetResult;
+        fn remove(&self, name: &CxxWString, flags: u16) -> EnvStackSetResult;
+        fn get_pwd_slash(&self) -> UniquePtr<CxxWString>;
+        fn set_pwd_from_getcwd(&self);
+
+        fn push(&mut self, new_scope: bool);
+        fn pop(&mut self);
+
+        // Returns a Box<OwningNullTerminatedArrayRefFFI>.into_raw() cast to a void*.
+        // This is because we can't use the same C++ bindings to a Rust type from two different bridges.
```
krobelus, line 126: we can, for example `ParseErrorListFfi` is used by bridges in `parse_constants.rs`, `parse_tree.rs` and `ast.rs`
It needs this incantation:

```rust
unsafe impl cxx::ExternType for OwningNullTerminatedArrayRefFFI<'_> {
    type Id = cxx::type_id!("OwningNullTerminatedArrayRefFFI");
    type Kind = cxx::kind::Opaque;
}
```

fish-rust/src/env/env_ffi.rs
```diff
@@ -0,0 +1,368 @@
...
+    /// FFI helper.
+    /// This returns either null, or the result of Box.into_raw().
+    /// This is a workaround for the difficulty of passing an Option through FFI.
+    fn getf_ffi(&self, name: &CxxWString, mode: u16) -> *mut EnvVar {
+        match self.getf(
+            name.as_wstr(),
+            EnvMode::from_bits(mode).expect("Invalid mode bits"),
+        ) {
+            None => std::ptr::null_mut(),
+            Some(var) => Box::into_raw(Box::new(var)),
```
krobelus, line 345: returning `*mut Box<EnvVar>` would probably make things simpler

krobelus:
I took a quick look, seems fine

> There's still a lot of initialization code in env.cpp which will be migrated to Rust later, but this is a good milestone.

ok that's only env_init and some static functions

fish-rust/src/env/environment.rs
```diff
@@ -1,44 +1,462 @@
...
+        None
+    }
+
+    fn get_names(&self, _flags: EnvMode) -> Vec<WString> {
+        Vec::new()
+    }
+}
+
+/// A helper type for wrapping a type-erased Environment.
+pub struct EnvDyn {
```
krobelus, line 106: Haven't tried it but I'd remove this type and have functions take `&dyn Environment`, which is simpler.
(Of course we'd need to keep the FFI version)

src/env_dispatch.cpp
```diff
@@ -195,6 +195,10 @@ void env_dispatch_var_change(const wcstring &key, env_stack_t &vars) {
     s_var_dispatch_table->dispatch(key, vars);
 }
 
+void env_dispatch_var_change_ffi(const wcstring &key) {
+    return env_dispatch_var_change(key, env_stack_t::principal());
```
krobelus, line 199: interesting that we don't have a test for this. Fair enough, we'll fix it later

fish-rust/src/env/environment_impl.rs
```diff
@@ -0,0 +1,1225 @@
...
+    }
+
+    // Exported variable array used by execv.
+    pub fn export_array(&mut self) -> Arc<OwningNullTerminatedArray> {
+        assert!(!is_forked_child());
+        if self.export_array_needs_regeneration() {
+            self.export_array = Some(self.create_export_array());
+
+            // Have to pull this into a local to satisfy the borrow checker.
+            let mut generations = std::mem::take(&mut self.export_array_generations);
```
krobelus, line 695: The C++ used clear() here. are we guaranteed that generations is empty here?

fish-rust/src/env/environment_impl.rs
```diff
@@ -0,0 +1,1225 @@
...
+
+        let mut result = ModResult::new(EnvStackSetResult::ENV_OK);
+        if query.has_scope {
+            // The user requested a particular scope.
+            // If we don't have uvars, fall back to using globals.
+            if query.universal && !UVAR_SCOPE_IS_GLOBAL.load() {
+                self.set_universal(key, val, query);
+                result.uvar_modified = true;
+            } else if query.global || (query.universal && UVAR_SCOPE_IS_GLOBAL.load()) {
+                Self::set_in_node(&mut self.base.globals, key, val, flags);
```
krobelus, line 791: missing `result.global_modified = true` I think

fish-rust/src/env/environment_impl.rs
```diff
@@ -0,0 +1,1225 @@
...
+                panic!("Unknown scope");
+            }
+        } else if let Some(mut node) = Self::find_in_chain(&self.base.locals, key) {
+            // Existing local variable.
+            Self::set_in_node(&mut node, key, val, flags);
+        } else if let Some(mut node) = Self::find_in_chain(&self.base.globals, key) {
+            // Existing global variable.
+            Self::set_in_node(&mut node, key, val, flags);
+            result.global_modified = true;
+        } else if !UVAR_SCOPE_IS_GLOBAL.load()
```
krobelus, line 826: the !UVAR_SCOPE_IS_GLOBAL is a new invention. It seems correct but it's weird that it isn't used in the remove() function below

fish-rust/src/env/environment_impl.rs
```diff
@@ -0,0 +1,1225 @@
...
+            popped = mem::replace(&mut self.base.locals, next);
+        } else {
+            // Exhausted the inner scopes, put back a shadowing scope.
+            if let Some(shadowed) = self.shadowed_locals.pop() {
+                popped = mem::replace(&mut self.base.locals, shadowed);
+            } else {
+                panic!("Attempt to pop last local scope")
+            }
+        }
+        let var_names = popped.borrow().env.keys().cloned().collect();
```
krobelus, line 949: nit: should avoid cloning the keys if possible

fish-rust/src/env/environment_impl.rs
```diff
@@ -0,0 +1,1225 @@
...
+            };
+            if !matches {
+                return Some(EnvStackSetResult::ENV_SCOPE);
+            }
+        }
+
+        // Handle computed mutable electric variables.
+        if key == "umask" {
+            return Some(set_umask(val));
+        } else if key == "PWD" {
```
krobelus, line 1016: TIL we can compare wstr with plain string literals.
This is nice, we should probably use native string literals wherever possible.

fish-rust/src/env/env_ffi.rs
ridiculousfish: I am not sure this is possible without constructing a `Box<Box<EnvVar>>`

fish-rust/src/env/environment_impl.rs
ridiculousfish: Yikes, I missed this. Great catch. Fixed

fish-rust/src/env/environment_impl.rs
ridiculousfish: Thanks, I'm not sure why I added that; removed.

fish-rust/src/env/environment_impl.rs
ridiculousfish: fixed

fish-rust/src/env/env_ffi.rs
ridiculousfish: nice, TIL. Fixed.

fish-rust/src/env/environment.rs
ridiculousfish: fixed

src/env_dispatch.cpp
ridiculousfish: there's only one parser at the moment, eventually we'll have more

fish-rust/src/env/environment_impl.rs
ridiculousfish: This was actually a bug in the C++ implementation - we were potentially reading these keys without holding the lock.

PR #9771, opened 2023-05-02
Rework ASAN integration 

mqudsi:
After fighting with ASAN for three days, I think I have something better. 

* The workaround for the virtualization bug w/ ASAN segfaulting randomly was to use use_tls=0 which didn’t use the ASAN tls interposer. Flipping that makes threads use ASAN for tls management and everything works better (I think ASAN runs after they're killed and reaped?). If ASAN runs again into the `Tracer caught signal 11` error randomly because of this change, well, it's better than it failing every CI run.
* I also tweaked the other ASAN options a bit and added a separate LSAN suppressions list. The individual commits for those explain it. 
* Now the only thing we do differently under ASAN is free the ncurses `TERMINAL` pointer and use `exit()` instead of `_exit()` so LSAN gets a chance to report (this is for `fish.cpp`).
* There was an actual memory leak in `init_curses()` - I’m not sure how many times it could theoretically be called but it is not just once. That’s patched.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -119,9 +119,9 @@ jobs:
           # environments. Follow https://github.com/google/sanitizers/issues/1342 and
           # https://github.com/google/sanitizers/issues/1409 to track this issue.
           # UPDATE: this can cause spurious leak reports for __cxa_thread_atexit_impl() under glibc.
-          LSAN_OPTIONS: verbosity=0:log_threads=0:use_tls=1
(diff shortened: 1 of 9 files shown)


ridiculousfish (approved):
This is awesome! I'm happy you found an easy way to keep ASAN leak detection working. Thank you for humoring me.

mqudsi:
Thanks for pushing me to do better!

PR #9776, opened 2023-05-06
completions: fix qdbus Q_NOREPLY method completion

rocka:
## Description

`qdbus` may print method ["tag"](https://doc.qt.io/qt-5/qmetamethod.html#tag) as `Q_NOREPLY`[^1] when it has annotation `org.freedesktop.DBus.Method.NoReply`[^2], and this is the only "tag" value it can be[^3]. This PR just adds optional `Q_NOREPLY` to the regex.

## TODOs:

[^1]: https://github.com/qt/qttools/blob/v5.15.9-lts-lgpl/src/qdbus/qdbus/qdbus.cpp#L185
[^2]: https://github.com/qt/qtbase/blob/v5.15.9-lts-lgpl/src/dbus/qdbusabstractinterface_p.h#L63
[^3]: https://github.com/qt/qtbase/blob/v5.15.9-lts-lgpl/src/dbus/qdbusmetaobject.cpp#L294-L296

### DIFF
--- a/share/completions/qdbus.fish
+++ b/share/completions/qdbus.fish
@@ -11,7 +11,7 @@ function __fish_qdbus_complete
     set argc (count $argv)
     if test $argc -le 3
         # avoid completion of property value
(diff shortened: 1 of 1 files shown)


ridiculousfish:
Thank you!

PR #9778, opened 2023-05-06
Change Job and Process to use interior mutability and fix fallout

mqudsi:
I removed the `RwLock`s but then all the `&T` and `&mut T` lifetimes were conflicting and nothing made sense so I changed to using a bare `&mut Job` and `&mut Process` when they are first created then after mutation has finished switching the `Job` to `Arc<Job>` (well, now it's an `Rc<Job>`). All access to `Job` after it's been handed to the `Parser` and all access to `Process` after it's been added to a `Job` happens through interior mutability.

----

    Use interior mutability for most Job and Process operations

    We only use `&mut Job` from the creation of a new `Job` until we put it in an
    `Arc` and hand it over to the parser. From that point on, all `Job` access must
    be read-only.

    Aside from not needing an `RwLock` because a Job is never sent to another
    thread and is only accessed from the same thread it was created, I believe the
    old code was prone to deadlocks since there were too many read borrows flying
    around at once for us to reliably say that no one will try to get a simultaneous
    write lock.

    The change necessitated the same treatment for `Process`: aside from when it is
    first created, we no longer use `&mut Process` for anything. All methods that
    can be called during the lifetime of a Process now use interior mutability.

    Note that with these changes we need to make sure that Job, JobRef, Process,
    etc. are all !Send and !Sync. Our interior mutability uses atomics but it
    doesn't use locks when updating multiple fields, meaning outside observers from
    another thread would see torn writes to the structs as a whole.

    With this change, the usages of `Parser::jobs_mut()` have been whittled down to
    just two (those that add a new job and those that remove an existing job) - all
    other access can now happen via `&Job` read-only references.

----
    Make Process and Job !Send and !Sync

    They are not intended to be shared across threads and our interior mutability
    functions are only sufficient for same-core access.

    With this change, we can also use Rc<Job> instead of Arc<Job> since there's no
    point to paying the atomic price here.

----

    Use interior mutability in generation_list_t

    The previous approach transmuting the values to `Cell<u64>` was extremely
    unsound since the compiler doesn't treat structs containg `Cell<T>` the same way
    that it treats `T`.

    It wasn't possible with the ffi because `at_mut()` returns a mutable reference
    and we really don't want to use `RefCell` here when the usecase is practically
    crying for `Cell` instead.

### DIFF
--- a/fish-rust/src/topic_monitor.rs
+++ b/fish-rust/src/topic_monitor.rs
@@ -31,10 +31,8 @@ use nix::unistd;
 use std::cell::UnsafeCell;
 use std::mem;
 use std::pin::Pin;
-use std::sync::{
-    atomic::{AtomicU8, Ordering},
-    Condvar, Mutex, MutexGuard,
-};
+use std::sync::atomic::{AtomicU64, AtomicU8, Ordering};
+use std::sync::{Condvar, Mutex, MutexGuard};
 
 #[cxx::bridge]
 mod topic_monitor_ffi {
@@ -42,8 +40,11 @@ mod topic_monitor_ffi {
     /// This should be kept in sync with topic_t.
     #[derive(Default, Copy, Clone, Debug, PartialEq, Eq)]
     pub struct generation_list_t {
+        // TODO: Replace with AtomicU64 for interior mutability
         pub sighupint: u64,
+        // TODO: Replace with AtomicU64 for interior mutability
         pub sigchld: u64,
+        // TODO: Replace with AtomicU64 for interior mutability
         pub internal_exit: u64,
     }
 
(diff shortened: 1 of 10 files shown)


fish-rust/src/topic_monitor.rs
```diff
@@ -28,68 +28,56 @@ use crate::wchar::{widestrs, wstr, WString};
...
+    sighupint = 0,     // Corresponds to both SIGHUP and SIGINT signals.
+    sigchld = 1,       // Corresponds to SIGCHLD signal.
+    internal_exit = 2, // Corresponds to an internal process exit.
+}
 
-    extern "Rust" {
-        type topic_monitor_t;
-        fn new_topic_monitor() -> Box<topic_monitor_t>;
+// XXX: Is it correct to use the default or should the default be invalid_generation?
+#[derive(Clone, Default, PartialEq, PartialOrd, Eq, Ord)]
```
krobelus, line 50: Default is correct, we need to initialize to zero

PR #9784, opened 2023-05-09
use macro to clean the code

wuyudi:
## Description

Use macro to reduce repeated function signatures. Now it's auto convert from `aa_bb` to `AaBb`.

Fixes issue #

## TODOs:

### DIFF
--- a/fish-rust/src/ast.rs
+++ b/fish-rust/src/ast.rs
@@ -28,6 +28,7 @@ use crate::wutil::printf::sprintf;
 use crate::wutil::wgettext_fmt;
 use cxx::{type_id, ExternType};
 use cxx::{CxxWString, UniquePtr};
+use paste::paste;
 use std::ops::{ControlFlow, Index, IndexMut};
 use widestring_suffix::widestrs;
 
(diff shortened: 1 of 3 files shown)


faho:
Sorry, this diff is massive and this change isn't helpful.

It also introduces an additional dependency just for, at best, minor cleanup.

I am rejecting this and asking you to not send similar PRs in future.

I also have my suspicions that this may be AI generated in which case I'm asking you to stop.

fish-rust/src/ast.rs
```diff
@@ -160,6 +161,18 @@ trait NodeMut: Node + AcceptorMut + ConcreteNodeMut {
     fn as_node(&self) -> &dyn Node;
 }
 
+macro_rules! fn_return_option_none {
+    ($($name:ident),*) => {
+        $(
+            paste!{
+                fn [<as_ $name>](&self) -> Option<&[<$name:camel>]> {
```
krobelus, line 168: interesting, I didn't know this was possible without defining a proc macro

Xiretza: To be clear `paste!{}` *is* a proc macro, it's just being wrapped up in another declarative macro.

PR #9786, opened 2023-05-10
Riir env_dispatch

mqudsi:
General port of env_dispatch was routine, but I ran into issues as I began to call rustified versions of the cpp fish init functions which left the C++ side of things in a broken state. I had to synchronize some variables (by exporting the rust ones and declaring them as `extern int` on the cpp side) and for other things I had the rust init function call into to the C++ init function afterwards to keep both init.

The only other interesting part was porting the first bit of code that uses the (n)curses functionality. ncurses presents a unified interface over both termcap and terminfo capabilities/databases; I originally switched to `terminfo` but that doesn't support installations where there isn't a separate, explicit terminfo install w/ the databases available (ran into this on Linux and BSD) then I [wrote my own termcap](https://github.com/mqudsi/termcap) crate as a fallback for that but decided it made more sense to just use the curses library to query for the features via some ffi calls, as that's what fish regularly does.

The other issue I ran into is that we use some curses functionality that doesn't seem to be exposed in the actual library but was rather made possible via some macro magic. In particular, we were overriding some missing capability strings on macOS to shim support for enter/exit italics mode and enter dim mode (for some reason *not* exit dim mode?) - this doesn't seem to be a feature of the actual, abi-stable part of the curses library since `tgetstr(3)` - the function we would use to query for a string capability  - returns a null pointer in case a capability isn't found.

Anyway, long story short, there's a minimal safe wrapper around the system curses functionality and we'll no longer to be using the curses features that diverge between platforms and distributions that give our cmake system such a hard time. Instead our `curses::Term` just checks if the string capability is present and uses a local override list to allow polyfilling missing capabilities. When we port the code that actually calls tputs(3) or tparm(3), we'll expose that through `curses::Term` as well, but the applied workaround also works if we call those directly.

~~I think I'll make `curses::Term` use internal mutability in the future, but for now there's a global `TERM: curses::Term` variable in `env/mod.rs` that's set/reset by `env_dispatch.rs` and would be used by `screen` or `reader` or whatnot.~~ (It now uses interior mutability.)

EDIT:

Oh yeah, `Environment`/`environment_t` ffi is gnarly. It was a pita to add an ffi version of the C++ functions that takes `UniquePtr<env_var_t>` instead of `environment_t` to be able to call them from rust.


fish-rust/src/env_dispatch.rs
```diff
@@ -0,0 +1,800 @@
...
+    crate::ffi::reader_schedule_prompt_repaint();
+}
+
+fn handle_change_ambiguous_width(vars: &dyn Environment) {
+    let new_width = vars
+        .get(L!("fish_ambiguous_width"))
+        .as_ref()
+        .map(|v| v.as_string())
+        // C++ code disregarded fish_wcstol() return of 0, so we'll unwrap to the same.
+        .map(|fish_ambiguous_width| fish_wcstoi(fish_ambiguous_width.chars()).unwrap_or(0))
```
faho, line 218: Ditto.

fish-rust/src/env_dispatch.rs
```diff
@@ -0,0 +1,800 @@
...
+    crate::ffi::complete_invalidate_path();
+}
+
+fn handle_tz_change(var_name: &wstr, vars: &dyn Environment) {
+    handle_timezone(var_name, vars);
+}
+
+fn handle_locale_change(vars: &dyn Environment) {
+    init_locale(vars);
+    // We need to re-guess emoji width because the locale might have change to a multibyte one.
```
faho, line 275: (typo: `changed`)

fish-rust/src/env_dispatch.rs
```diff
@@ -0,0 +1,800 @@
...
+/// primarily intended for testing, but could also be used directly by users in special situations.
+fn handle_read_limit_change(vars: &dyn Environment) {
+    let read_byte_limit = vars
+        .get_unless_empty(L!("fish_read_limit"))
+        .map(|v| v.as_string())
+        .and_then(|v| match v.to_string().parse() {
+            Ok(v) => Some(v),
+            Err(_) => {
+                // XXX: In the C++ code, this warning wasn't behind an "is_interactive_session()"
+                // check.
```
faho, line 307: Not really sold on this - $fish_read_limit is valid even non-interactively.

fish-rust/src/env_dispatch.rs
```diff
@@ -0,0 +1,800 @@
...
+                None
+            }
+        })
+        .unwrap_or(DEFAULT_READ_BYTE_LIMIT);
+    READ_BYTE_LIMIT.store(read_byte_limit, Ordering::Relaxed);
+}
+
+fn handle_fish_trace(vars: &dyn Environment) {
+    let enabled = vars.get_unless_empty(L!("fish_trace")).is_some();
+    crate::trace::trace_set_enabled(enabled);
```
faho, line 320: (this is a faithful translation of the c++, but we might want to rethink only enabling this if the value is some "truthy" value - numeric and > 0 or "yes" or "true" or something)

(I'm planning on adding additional $fish_trace values to allow tracing event handlers and such so this'll probably happen eventually)

fish-rust/src/env_dispatch.rs
```diff
@@ -0,0 +1,800 @@
...
+    }
+}
+
+/// Try to initialize the terminfo/curses subsystem using our fallback terminal name. Do not set
+/// `$TERM` to our fallback. We're only doing this in the hope of getting a functional shell.
+/// If we launch an external command that uses `$TERM`, it should get the same value we were given,
+/// if any.
+fn initialize_curses_using_fallbacks(vars: &dyn Environment) {
+    // xterm-256color is the most used terminal type by a massive margin, especially counting
+    // terminals that are mostly compatbile.
```
faho, line 486: (typo "compatible")

fish-rust/src/env_dispatch.rs
```diff
@@ -0,0 +1,800 @@
...
+fn apply_term_hacks(vars: &dyn Environment) {
+    // Midnight Commander tries to extract the last line of the prompt, and does so in a way that is
+    // broken if you do '\r' after it like we normally do.
+    // See https://midnight-commander.org/ticket/4258.
+    if vars.get(L!("MC_SID")).is_some() {
+        crate::ffi::screen_set_midnight_commander_hack();
+    }
+
+    // Be careful, variables like `enter_italics_mode` are #defined to dereference through
+    // `cur_term`.
```
faho, line 530: Is this comment still accurate?

fish-rust/src/env_dispatch.rs
```diff
@@ -0,0 +1,800 @@
...
+            FLOGF!(warning, wgettext!("Could not set up terminal."));
+            if let Some(term) = term {
+                FLOGF!(
+                    warning,
+                    wgettext!("TERM environment variable set to: "),
+                    term
+                );
+                FLOGF!(
+                    warning,
+                    wgettext!("Check that this terminaly type is supported on this system.")
```
faho, line 643: (typo: "terminal type")

src/env_dispatch.cpp
```diff
@@ -1,3 +1,4 @@
+/*
```
faho, line 1: (these should be removed before merge)

fish-rust/src/curses.rs
```diff
@@ -0,0 +1,304 @@
...
+struct Code {
+    /// The two-char termcap code for the capability, followed by a nul.
+    code: [u8; 3],
+}
+
+impl Code {
+    const fn new(code: &str) -> Code {
+        let code = code.as_bytes();
+        Code {
+            code: [code[0], code[1], b'\0'],
```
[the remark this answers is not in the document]
mqudsi: It's a `const` fn so a panic here could be a compile-time error. The two-byte limit is hardcoded into termcap, it's not from me.

fish-rust/src/curses.rs
```diff
@@ -0,0 +1,304 @@
...
+
+impl<'a> Capability<'a> for StringCap {
+    type Result = Option<&'a str>;
+
+    fn lookup(&self, term: &'a mut Term) -> Self::Result {
+        let id = self.0;
+        match term.overrides.binary_search_by(|entry| entry.0.cmp(self)) {
+            Ok(idx) => Some(&term.overrides[idx].1),
+            Err(idx) => {
+                let mut result = vec![b'\0'; 100];
```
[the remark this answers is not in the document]
mqudsi: It's about a 10x conservative estimate based off my perusal of the `termcap` database. ncurses doesn't document the length and `tgetstr()` doesn't accept a length parameter to prevent an overflow!!

fish-rust/src/curses.rs
```diff
@@ -0,0 +1,304 @@
...
+{
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        (self.view_mut)(&mut self.value)
+    }
+}
+
+const OK: i32 = 0;
+const ERR: i32 = -1;
+
+mod ffi {
```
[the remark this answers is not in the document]
mqudsi: Yeah, that's not a bad point. It would indeed by semi-permanent.

fish-rust/src/env_dispatch.rs
mqudsi: I had the same thought.

fish-rust/src/env_dispatch.rs
mqudsi: Thanks. I found it faster to retype than copy-and-paste so some typos snuck in.

fish-rust/src/env_dispatch.rs
mqudsi: No, but trying to get the value of `ENTER_ITALICS_MODE` et. al. without curses having been initialized will panic. I should update that comment.

fish-rust/src/curses.rs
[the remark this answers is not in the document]
faho: My concern really is just that someone tries to construct a Code with the wrong string - like the probably better known terminfo name instead of the termcap one.

A compile-time error would do just fine. I'd assume that only works for literals? (which would be fine)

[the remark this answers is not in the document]
mqudsi: I removed it from the latest update by not caching ncurses-returned strings at all.

fish-rust/src/curses.rs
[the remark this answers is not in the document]
mqudsi: I removed it from the latest update by not caching ncurses-returned strings at all.

fish-rust/src/curses.rs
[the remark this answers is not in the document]
mqudsi: Yeah, if it's not a literal string then it wouldn't be a compile-time error but the type is also private to the module and not directly exposed so I think the panic I added per your request should cover it.

fish-rust/src/env/mod.rs
```diff
@@ -5,4 +5,49 @@ pub mod var;
...
+pub static TERM_HAS_XN: AtomicBool = AtomicBool::new(false);
+
+mod ffi {
+    extern "C" {
+        pub fn setenv_lock(
+            name: *const libc::c_char,
+            value: *const libc::c_char,
+            overwrite: libc::c_int,
+        );
+        pub fn unsetenv_lock(name: *const libc::c_char);
```
krobelus, line 36: these should be private, same for other extern C decls below

𑁍 5ec2fc1688916039f00716f7d3b353654abd4e12 fish-rust/src/env_dispatch.rs:48 + -18
+    extern "C" {
+        /// The ncurses `cur_term` TERMINAL pointer.
+        pub static mut cur_term: *const core::ffi::c_void;
+
+        /// setupterm(3) is a low-level call to begin doing any sort of `term.h`/`curses.h` work.

we can use RelaxedAtomicBool instead of AtomicBool

𑁍 5ec2fc1688916039f00716f7d3b353654abd4e12 fish-rust/src/env_dispatch.rs:51 + -18
+
+        /// setupterm(3) is a low-level call to begin doing any sort of `term.h`/`curses.h` work.
+        /// It's called internally by ncurses's `initscr()` and `newterm()`, but the C++ code called
+        /// it directly from [`initialize_curses_using_fallbacks()`].
+        pub fn setupterm(

this no longer leaks. Presumably this is fine

𑁍 5ec2fc1688916039f00716f7d3b353654abd4e12 fish-rust/src/env_dispatch.rs:143 + -18
+    pub fn dispatch(&self, key: &wstr, vars: &EnvStack) {
+        if let Some(named) = self.named_table.get(key) {
+            (named)(key, vars);
+        }
+        if let Some(anon) = self.anon_table.get(key) {

I don't think we can translate wcs2zstring(env_var_name) as env_var_name.to_string() because the latter does not know about our encoding scheme for codepoints that lie outside the valid Unicode ranges. 
We should use `wcs2zstring` which returns the CString we need.
Same for var.to_string().

𑁍 5ec2fc1688916039f00716f7d3b353654abd4e12 fish-rust/src/env_dispatch.rs:168 + -18
+    } else {
+        unsetenv_lock(&name);
+    }
+
+    extern "C" {

missing the "overwriting default" FLOGF

𑁍 5ec2fc1688916039f00716f7d3b353654abd4e12 fish-rust/src/env_dispatch.rs:309 + -18
+        USE_POSIX_SPAWN.store(false, Ordering::Relaxed);
+    } else if let Some(var) = vars.get(L!("fish_use_posix_spawn")) {
+        let use_posix_spawn =
+            var.is_empty() || crate::wcstringutil::bool_from_string(&var.as_string());
+        USE_POSIX_SPAWN.store(use_posix_spawn, Ordering::Relaxed);

The C++ code didn't update READ_BYTE_LIMIT in the error case

𑁍 5ec2fc1688916039f00716f7d3b353654abd4e12 fish-rust/src/env_dispatch.rs:539 + -41
+                    warning,
+                    wgettext!("Could not set up terminal using the fallback terminal type: "),
+                    term,
+                );
+            }

we could also add some tiny helpers in `compat.c`

krobelus, line 36: these should be private, same for other extern C decls below

fish-rust/src/env_dispatch.rs
```diff
@@ -0,0 +1,796 @@
...
+    L!("fish_allow_singlebyte_locale"),
+];
+
+#[rustfmt::skip]
+const CURSES_VARIABLES: &[&wstr] = &[
+    L!("TERM"), L!("TERMINFO"), L!("TERMINFO_DIRS")
+];
+
+/// Whether to use `posix_spawn()` when possible.
+static USE_POSIX_SPAWN: AtomicBool = AtomicBool::new(false);
```
krobelus, line 41: we can use RelaxedAtomicBool instead of AtomicBool

fish-rust/src/env_dispatch.rs
```diff
@@ -0,0 +1,796 @@
...
+
+/// Whether we think we can set the terminal title or not.
+static CAN_SET_TERM_TITLE: AtomicBool = AtomicBool::new(false);
+
+/// The variable dispatch table. This is set at startup and cannot be modified after.
+static VAR_DISPATCH_TABLE: once_cell::sync::Lazy<VarDispatchTable> =
```
krobelus, line 47: this no longer leaks. Presumably this is fine

fish-rust/src/env_dispatch.rs
```diff
@@ -0,0 +1,796 @@
...
+        env_var_name,
+        "| => |",
+        var.as_ref()
+            .map(|v| v.as_utfstr())
+            .unwrap_or(L!("MISSING/EMPTY")),
+        "|"
+    );
+    let name = env_var_name.to_string();
+    if let Some(var) = var {
+        setenv_lock(&name, &var.to_string(), true);
```
krobelus, line 145: I don't think we can translate wcs2zstring(env_var_name) as env_var_name.to_string() because the latter does not know about our encoding scheme for codepoints that lie outside the valid Unicode ranges. 
We should use `wcs2zstring` which returns the CString we need.
Same for var.to_string().

fish-rust/src/env_dispatch.rs
```diff
@@ -0,0 +1,796 @@
...
+
+/// Update the value of [`FISH_EMOJI_WIDTH`].
+fn guess_emoji_width(vars: &dyn Environment) {
+    use crate::fallback::FISH_EMOJI_WIDTH;
+
+    if let Some(width_str) = vars.get(L!("fish_emoji_width")) {
+        // The only valid values are 1 or 2; we default to 1 if it was an invalid int.
+        let new_width = fish_wcstoi(width_str.as_string().chars()).unwrap_or(1);
+        let new_width = new_width.min(2).max(1);
+        FISH_EMOJI_WIDTH.store(new_width, Ordering::Relaxed);
```
krobelus, line 167: missing the "overwriting default" FLOGF

fish-rust/src/env_dispatch.rs
```diff
@@ -0,0 +1,796 @@
...
+                // XXX: In the C++ code, this warning wasn't behind an "is_interactive_session()"
+                // check.
+                if is_interactive_session() {
+                    FLOGF!(warning, "Ignoring invalid $fish_read_limit");
+                }
+                None
+            }
+        })
+        .unwrap_or(DEFAULT_READ_BYTE_LIMIT);
+    READ_BYTE_LIMIT.store(read_byte_limit, Ordering::Relaxed);
```
krobelus, line 315: The C++ code didn't update READ_BYTE_LIMIT in the error case

fish-rust/src/env_dispatch.rs
mqudsi: Actually, all rust static variables are leaked exactly because of the problems caused by C++ *not* leaking static variables.

fish-rust/src/env_dispatch.rs
mqudsi: It's an exported environment variable, presumably that is intended for use outside of fish itself and therefore shouldn't have private use codepoints? I think it would be a mistake to make those externally visible anywhere, no?

fish-rust/src/env_dispatch.rs
mqudsi: Hmmm, good catch - if the read byte limit was previously overridden with a valid value then this would reset it to the default. But surely resetting it to the default in case of an invalid value makes more sense than preserving the prior value? After all, the user *did* intend to change it? We do warn about the invalid value, so the user would have a chance to try again if they needed/wanted to, no?

fish-rust/src/env_dispatch.rs
krobelus: if we are given `set var \xff`, that's a one-byte string that's not a valid UTF-8 sequence.
We use `encode_byte_to_char()` to convert it to a valid wide string, using the private use area. In this case `"\u{F6FF}"`
Now when we pass the value of `$var` to the system we must not encode those private-use codepoints to UTF-8.
Instead we should reverse the private-use encoding using `decode_byte_from_char()`, which is what the `wcs2*string` family does.

fish-rust/src/env/mod.rs
mqudsi: No, actually. They're in a submodule so making them `pub` makes them visible to `env/mod.rs` itself. We'd have to `pub use self::ffi::*` to actually make them public to the crate.

fish-rust/src/env_dispatch.rs
mqudsi: Thanks for clarifying!

mqudsi: Wait, at least thus far in the rust code the only uses of `setenv_lock()`/`unsetenv_lock()` are only with fixed strings (the locale variables, the curses variables, and the timezone variable `TZ`.

mqudsi: The `env_var_name.to_string()` bit is actually fine because the only uses of it are with fixed/hard-coded strings (the curses variable names, the locale variable names, and the timezone variable name `TZ`).

But the `value = var.as_string().to_string()` bit is the one that could contain user-specified content.

fish-rust/src/env_dispatch.rs
mqudsi: Yeah, that makes sense. All the other warnings are being suppressed in non-interactive mode so I thought this could join them.

ridiculousfish:
This is failing a lot of tests on my Mac - [here's the output](https://gist.github.com/ridiculousfish/ba844fc59aa94caabb6934b84a29915a). Not sure why it's passing the CI! Let me know if you want my Mac-assistance debugging.

mqudsi:
So I don't know if it actually changes anything or not but I noticed a subtle difference in the behavior of the two codebases that *might* have something to do with this.

In the C++ code we were explicitly initializing VAR_DISPATCH_TABLE after running some the startup init routines and ``env_var_dispatch_var_change()` was obviously checking if the global pointer was initialized before dereferencing it (out of necessity). But in rust, we used `Lazy<T>` to create the `VAR_DISPATCH_TABLE` and so it was automatically initialized on first access, so I didn't need to port that null pointer check. However that means that the codebase was reacting to variable changes triggered by the init routines before they had completed, so I changed it so that `env_var_dispatch_var_change()` never materializes an as-of-yet-uninitialized `VAR_DISPATCH_TABLE` (and forcibly materialize the table at the same point in time that the C++ codebase did).

https://github.com/fish-shell/fish-shell/pull/9786/commits/fdfde0c604f45808c039721e91debf07bdfcfe2a

mqudsi:
Oh, I think I figured it out. These error messages are all localized with wgettext, and I added a hack to the `env_dispatch.cpp` to get it to pass CI until I open a new PR that correctly detects when `_nl_msg_cat_cntr` is available. In your case, fish is compiled *with* gettext but I hard-coded that `_nl_msg_cat_ctr` is unavailable on macOS because under macOS CI we build without gettext. 

Since `_nl_msg_cat_cntr` doesn't get incremented on your machine, gettext remains blissfully unaware of the localization changes we made and butchers those localized error messages.

I don't want to fix `_nl_msg_cat_cntr` in this PR as I have a proper fix in another branch. If you *really* want, I can land that fix to master then rebase on top of it so that when this PR gets merged everything is functioning correctly.

krobelus:
> I don't want to fix _nl_msg_cat_cntr in this PR as I have a proper fix that actually sets up build.rs with correct header/symbol/library detection in another branch. If you really want, I can land that fix to master then rebase on top of it so that when this PR gets merged everything is functioning correctly.

I would normally not deviate from principles (every commit should pass tests, fixup-style commits should be squashed) without a good reason.

fish-rust/src/env_dispatch.rs
```diff
@@ -0,0 +1,809 @@
...
+
+impl VarDispatchTable {
+    fn observes_var(&self, name: &wstr) -> bool {
+        self.named_table.contains_key(name) || self.anon_table.contains_key(name)
+    }
+
+    /// Add a callback for the variable `name`. We must not already be observing this variable.
+    pub fn add(&mut self, name: &'static wstr, callback: NamedEnvCallback) {
+        let prev = self.named_table.insert(name, callback);
+        assert!(prev.is_none(), "Already observing {}", name);
```
ridiculousfish, line 112: The previous code checked `observes_var` which looks in both tables. It's not super-important but let's preserve the existing behavior.

fish-rust/src/env_dispatch.rs
```diff
@@ -0,0 +1,809 @@
...
+    }
+
+    let term = vars
+        .get(L!("TERM_PROGRAM"))
+        .map(|v| v.as_string())
+        .unwrap_or_else(WString::new);
+    let version = vars
+        .get(L!("TERM_PROGRAM_VERSION"))
+        .map(|v| v.as_string().to_string())
+        .and_then(|v| v.parse().ok())
```
ridiculousfish, line 183: This was `strtod` in the previous code and should be `wcstod` in Rust.

For example my `$TERM_PROGRAM_VERSION` is `3.4.19`; this will fail with `parse()` but succeeds with `wcstod`.

fish-rust/src/env_dispatch.rs
```diff
@@ -0,0 +1,809 @@
...
+
+fn handle_change_ambiguous_width(vars: &dyn Environment) {
+    let new_width = vars
+        .get(L!("fish_ambiguous_width"))
+        .as_ref()
+        .map(|v| v.as_string())
+        // The only valid values are 1 or 2; we default to 1 if it was an invalid int.
+        .map(|fish_ambiguous_width| fish_wcstoi(fish_ambiguous_width.chars()).unwrap_or(0))
+        .unwrap_or(1)
+        .clamp(1, 2);
```
ridiculousfish, line 230: The C++ code doesn't clamp this; rather it just sets it to at least zero.

fish-rust/src/env_dispatch.rs
```diff
@@ -0,0 +1,809 @@
...
+
+fn handle_fish_term_change(vars: &dyn Environment) {
+    update_fish_color_support(vars);
+    crate::ffi::reader_schedule_prompt_repaint();
+}
+
+fn handle_change_ambiguous_width(vars: &dyn Environment) {
+    let new_width = vars
+        .get(L!("fish_ambiguous_width"))
+        .as_ref()
```
ridiculousfish, line 225: nit: this `as_ref` isn't needed

fish-rust/src/env_dispatch.rs
```diff
@@ -0,0 +1,809 @@
...
+}
+
+fn handle_fish_cursor_selection_mode_change(vars: &dyn Environment) {
+    use crate::reader::CursorSelectionMode;
+
+    let inclusive = vars
+        .get(L!("fish_cursor_selection_mode"))
+        .as_ref()
+        .map(|v| v.as_string())
+        .map(|v| v == L!("inclusive"))
```
ridiculousfish, line 251: Just a note you can compare against strs: `v == "inclusive"` is fine.

fish-rust/src/env_dispatch.rs
```diff
@@ -0,0 +1,809 @@
...
+    }
+}
+
+/// Allow the user to override the limits on how much data the `read` command will process. This is
+/// primarily intended for testing, but could also be used directly by users in special situations.
+fn handle_read_limit_change(vars: &dyn Environment) {
+    let read_byte_limit = vars
+        .get_unless_empty(L!("fish_read_limit"))
+        .map(|v| v.as_string())
+        .and_then(|v| match v.to_string().parse() {
```
ridiculousfish, line 313: The previous code used `fish_wcstoull` which does something different than `parse`. This should probably use `fish_wcstoi`.

fish-rust/src/env_dispatch.rs
ridiculousfish: We should just do what we did before; no reason to change behavior as part of the port.

fish-rust/src/env_dispatch.rs
```diff
@@ -0,0 +1,809 @@
...
+    guess_emoji_width(vars);
+    update_wait_on_escape_ms(vars);
+    handle_read_limit_change(vars);
+    handle_fish_use_posix_spawn_change(vars);
+    handle_fish_trace(vars);
+}
+
+/// Updates our idea of whether we support term256 and term24bit (see issue #10222).
+fn update_fish_color_support(vars: &dyn Environment) {
+    let max_colors = curses::term().get(curses::MAX_COLORS);
```
ridiculousfish, line 360: The previous code guarded this with `cur_term != nullptr`; it looks like this will crash if cur_term is not set?

fish-rust/src/env_dispatch.rs
```diff
@@ -0,0 +1,809 @@
...
+        // We assume that if this is iTerm it can't also be st, so having this check inside is okay.
+        if !it.contains(':') {
+            supports_24bit = true;
+            FLOGF!(term_support, "True-color support: enabled for iTerm");
+        }
+    } else if term.starts_with("st-") {
+        supports_24bit = true;
+        FLOGF!(term_support, "True-color support: enabling for st");
+    } else if let Some(vte) = vars.get(L!("VTE_VERSION")).map(|v| v.as_string()) {
+        if fish_wcstoi(vte.chars()).unwrap_or(0) > 3600 {
```
ridiculousfish, line 467: This was `fish_wcstod` not `fish_wcstoi`

fish-rust/src/env_dispatch.rs
```diff
@@ -0,0 +1,809 @@
...
+/// If we launch an external command that uses `$TERM`, it should get the same value we were given,
+/// if any.
+fn initialize_curses_using_fallbacks(vars: &dyn Environment) {
+    // xterm-256color is the most used terminal type by a massive margin, especially counting
+    // terminals that are mostly compatible.
+    const FALLBACKS: [&str; 4] = ["xterm-256color", "xterm", "ansi", "dumb"];
+
+    let termstr = vars
+        .get_unless_empty(L!("TERM"))
+        .map(|v| v.as_string().to_string())
```
ridiculousfish, line 502: Please do not use `to_string` here. It can just be removed.

fish-rust/src/env_dispatch.rs
```diff
@@ -0,0 +1,809 @@
...
+
+    #[cfg(target_os = "macos")]
+    {
+        // Hack in missing italics and dim capabilities omitted from macOS xterm-256color terminfo.
+        // Helps Terminal.app and iTerm.
+        let term_prog = vars
+            .get(L!("TERM_PROGRAM"))
+            .map(|v| v.as_string())
+            .unwrap_or(WString::new());
+        if term_prog == L!("Apple_Terminal") || term_prog == L!("iTerm.app") {
```
ridiculousfish, line 554: note you can compare against string literals: `term_prog == "Apple_Terminal"`

fish-rust/src/env_dispatch.rs
```diff
@@ -0,0 +1,809 @@
...
+        } else {
+            FLOGF!(env_locale, "locale var", name, "is missing or empty");
+            unsetenv_lock(&name);
+        }
+    }
+
+    let locale = unsafe { CStr::from_ptr(libc::setlocale(libc::LC_ALL, b"\0".as_ptr().cast())) };
+
+    // Try to get a multibyte-capable encoding.
+    // A "C" locale is broken for our purpsose: any wchar function will break on it. So we try
```
ridiculousfish, line 715: not - purposes misspeeled

fish-rust/src/env_dispatch.rs
ridiculousfish, line 1: This file was a ton of work! Thank you for tackling it.

ridiculousfish (changes requested):
The env_dispatch changes look good overall.

Regarding curses, my suggestion is to have a list of all of the capabilities fish cares about, and then just eagerly load them (and copy the strings) in setupterm. This allows the type to actually be thread safe and should simplify other aspects.

fish-rust/src/curses.rs
```diff
@@ -0,0 +1,261 @@
...
+use self::sys::*;
+use crate::supercow::ArcCow;
+use once_cell::sync::Lazy;
+use std::ffi::{CStr, CString};
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::{Arc, Mutex};
+
+// This is `static mut` because a `Lazy` can only be init once but we may need to re-init the Term
+// value after a second+ call to curses::setup() in response to environment changes.
+pub static mut TERM: Lazy<Term> = Lazy::new(|| panic!("TERM not yet initialized!"));
```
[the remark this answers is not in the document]
mqudsi: Yes, my preference is also to require unwrapping at each call. When I wrote the code I was unaware that we could actually be running without curses having been successfully initialized and assumed we would tire of checking if TERM was set. It seems `cur_term` can actually remain null throughout fish's duration which means we always have to check.

fish-rust/src/curses.rs
```diff
@@ -0,0 +1,261 @@
...
+
+/// Returns a mutable reference to the global [`Term`] singleton.
+///
+/// Panics on deref if [`curses::setup()`](self::setup()) hasn't been called successfully.
+pub fn term() -> &'static Term {
+    unsafe { &TERM }
```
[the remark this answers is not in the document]
mqudsi: Thanks. It used to borrow a mutable reference but I changed it many times and that's now a stale comment.

fish-rust/src/curses.rs
```diff
@@ -0,0 +1,261 @@
...
+        let mut overrides = self.overrides.lock().expect("Mutex poisoned!");
+        match overrides.binary_search_by(|entry| entry.0.cmp(&id)) {
+            Ok(idx) => overrides[idx] = (id, value),
+            Err(idx) => overrides.insert(idx, (id, value)),
+        }
+        self.has_overrides.store(true, Ordering::Relaxed);
+    }
+}
+
+pub trait Capability<'a> {
```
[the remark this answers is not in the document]
mqudsi: Adding a function for each feature instead of passing the name of the capability to look up is just less rusty. And I don't want to add a getter *and* a setter for each capability we might want to override. Making the capability just a parameter to get/set is a lot less tedious and the ergonomics don't suffer for it.

fish-rust/src/curses.rs
```diff
@@ -0,0 +1,261 @@
...
+pub const EXIT_ITALICS_MODE: StringCap = StringCap::new("ZR");
+pub const ENTER_DIM_MODE: StringCap = StringCap::new("mh");
+
+// Number capabilities
+pub const MAX_COLORS: Number = Number::new("Co");
+
+// Flag capabilities
+pub const EAT_NEWLINE_GLITCH: Flag = Flag::new("xn");
+
+pub struct Term {
```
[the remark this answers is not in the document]
mqudsi: See [my previous comment](https://github.com/fish-shell/fish-shell/pull/9786#discussion_r1193217691) - there are no mutexes flying around here, the mutex is restricted only to the interior of the `get()`/`set()` call and never outlives those. 

>  An "override" would just be overwriting any values loaded from setupterm.

The devil hides behind that "just" - we can't (safely) overwrite a value without an &mut reference to the cached list of strings. We can refactor the code later to have `env_dispatch` both initialize term (getting a mutable `Term` instance and not a reference) and set its overrides with that `mut` reference *then* store the `TERM` singleton so that any subsequent borrows of TERM can be `&Term` and we'd never need `&mut Term`. But the way the calling code is currently structured, we arbitrarily need `&mut Term` references so we need the mutex (and the `has_overrides` machinery is just to bypass the mutex when there are no overrides - it's solely an optimization).

fish-rust/src/curses.rs
```diff
@@ -0,0 +1,261 @@
...
+}
+
+/// Calls the curses `setupterm()` function with the provided `$TERM` value `term` (or a null
+/// pointer in case `term` is null) for the file descriptor `fd`.
+///
+/// Note that the `errret` parameter is provided to the function, meaning curses will not write
+/// error output to stderr in case of failure.
+///
+/// Any existing references from `curses::term()` will be invalidated by this call!
+pub fn setup(term: Option<&str>, fd: i32) -> bool {
```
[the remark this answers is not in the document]
mqudsi: I contend that all valid `TERM` values are strictly ascii (and can guarantee from my own perusal of it that at least in the human-readable termcap database and not the binary terminfo database compiled into libcurses, all TERM identefiers are ascii).

Using `&str` just makes it more ergonomic everywhere (you would reject an invalid `&str` before reaching here, but that invalid `&str` was never going to match the termcap/terminfo database anyway), but I can see why you might want to reserve the option to pass in a CStr retrieved from an environment variable to make things easier and minimize the need to parse beforehand and will make the change.

fish-rust/src/curses.rs
```diff
@@ -0,0 +1,261 @@
...
+    }
+}
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+struct Code {
+    /// The two-char termcap code for the capability, followed by a nul.
+    code: [u8; 3],
+}
+
+impl Code {
```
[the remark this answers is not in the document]
mqudsi: That's not untrue. I personally thought the ergonomic wins and more importantly, the maintainability with no one ever forgetting the trailing `\0` or accidentally using the long-form terminfo name instead of the short-form termcap name was worth the extra dozen SLOC.

fish-rust/src/env/mod.rs
```diff
@@ -5,4 +5,55 @@ pub mod var;
...
+    }
+}
+
+/// Sets an environment variable after obtaining a lock, to try and improve the safety of
+/// environment variables.
+///
+/// As values could contain non-unicode characters, they must first be converted from &wstr to a
+/// `CString` with [`crate::common::wcs2zstring()`].
+pub fn setenv_lock(name: &str, value: &CStr, overwrite: bool) {
```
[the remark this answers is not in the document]
mqudsi: You are correct but as far as I can see we only call it on well-defined, hard-coded variable names (the CURSES_VARS and NCURSES_VARS lists plus `TZ`) so I made it `&str` for ergonomic reasons. I can change it, but also the entire thing is going to be dropped.

fish-rust/src/fallback.rs
```diff
@@ -10,23 +10,39 @@ use std::cmp;
...
-extern "C" {
-    pub fn wcwidth(c: libc::wchar_t) -> libc::c_int;
+mod fallback_ffi {
+    extern "C" {
+        pub fn wcwidth(c: libc::wchar_t) -> libc::c_int;
+    }
 }
-fn system_wcwidth(c: char) -> i32 {
+
+pub fn wcwidth(c: char) -> i32 {
```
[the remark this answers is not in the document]
mqudsi: We never had a wrapper function around wcwidth in the C++ code and always just called `wcwidth()`; the name `system_wcwidth()` only came about during the rust port and was only given due to the namespace clash between the extern `wcwidth()` and the safe rust wrapper around it. I resolved the namespace conflict so figured the real name could stay.

fish-rust/src/env_dispatch.rs
mqudsi: That was an oversight on my part. Thanks for catching it.

fish-rust/src/env_dispatch.rs
mqudsi: I was confused by the existing code and diverged from the C++ codebase. I should have mentioned the change. 

I ended up looking at what we were using `TERM_PROGRAM_VERSION` for and what the valid contents of that string are for the cases we are interested in and found that for `TERM_PROGRAM == "Apple_Terminal"` (which is the only case the code cares about) the version would be an integral value (we check for `version >= 400`).

tbh I have been trying to move away in the little places that I can from our custom code where the standard library functions will suffice.

fish-rust/src/env_dispatch.rs
mqudsi: Sorry, I confused myself tracking down the different comments in all the headers and cpp files around the g_fish_ambiguous_width and g_fish_emoji_width variables (they were even declared in multiple headers as extern). I'll change that and fix the comment.

fish-rust/src/env_dispatch.rs
mqudsi: I actually intentionally used `usize's` `FromStr` impl to avoid the previous behavior which silently truncated on 32-bit platforms when assigning the `unsigned long long` result to `size_t read_byte_limit`. 

Note that using `fish_wcstoi` would technically be a breaking change as well as read limits over 4GB would become invalid (I don't have a problem with that, but I assumed there was a reason `read_byte_limit` was a `size_t`).

fish-rust/src/env_dispatch.rs
mqudsi: Yup, that was lost in the rewrite. Thanks for catching. Changing `curses::term()` to return an `Option` is certainly going to be the safest approach.

fish-rust/src/env_dispatch.rs
mqudsi: I'd actually like to change it to `u32`'s `FromStr` impl (`string::parse()`) with [the same rational as above](https://github.com/fish-shell/fish-shell/pull/9786#discussion_r1193220614) as it seems the type of VTE_VERSION we are interested in (as there seem to be many) are the ones with higher integral values (in the hundreds or thousands) and not a lower program version like "3.2.0" so that would suffice.

fish-rust/src/env_dispatch.rs
mqudsi: Yup. I'll change it the function to take a `CStr`.

fish-rust/src/curses.rs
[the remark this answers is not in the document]
mqudsi: Actually, since I changed from `&str` to `Arc<CString>` for the reasons explained above, I can drop the mutex and change the vec to `Vec<Cell<Arc<CString>>>` - everything stays the same but we don't need a mutex since the clone out of the cell is "free". I can also go back to caching the result of ncurses lookups since there would never be a mutex so we don't need to avoid caching in order to avoid locking.

fish-rust/src/curses.rs
```diff
@@ -0,0 +1,261 @@
...
+                // so we can safely unwrap here.
+                result => CStr::from_ptr(result),
+            }
+        };
+        Some(ArcCow::Borrowed(result))
+    }
+}
+
+impl<'a> Capability<'a> for StringCap {
+    type Result = Option<ArcCow<'a, CStr, CString>>;
```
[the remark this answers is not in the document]
mqudsi: Actually, since I changed from `&str` to `Arc<CString>` for the reasons explained above, I can drop the mutex and change the vec to `Vec<Cell<Arc<CString>>>` - everything stays the same but we don't need a mutex since the clone out of the cell is "free". I can also go back to caching the result of ncurses lookups since there would never be a mutex so we don't need to avoid caching in order to avoid locking, meaning we always return `Arc<CString>` so there's no need for the `Cow`.

fish-rust/src/fallback.rs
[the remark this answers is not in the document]
krobelus: yeah, `wcwidth()` is the least surprising name for the system function

fish-rust/src/env_dispatch.rs
krobelus: I have been porting this check as `int has_cur_term() { return cur_term != NULL; }`

fish-rust/src/supercow.rs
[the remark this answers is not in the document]
mqudsi: It was needed to support returning either an `Arc<CString>` or `&CStr` because of the lifetime requirements imposed by interior mutability but read my comments below, I have found a way around it.

fish-rust/src/env_dispatch.rs
mqudsi: @krobelus We'll want to unify the "uses curses" code between our forks. I have a `curses::is_initialized()` function that does that check but with the most recent commits that change `curses::term()` to return an `Option<&'static Term>` we wouldn't actually use it anywhere. All curses functionality should be restricted to `curses.rs` and all functionality requiring a valid `cur_term` ptr should be restricted to methods on the `Term` type which is only available if `cur_term != nullptr`.

fish-rust/src/env/mod.rs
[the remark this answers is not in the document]
mqudsi: Never mind; there are a lot of places where we want have a `&str` or `&wstr` but need to call a function that doesn't require that its arguments by valid unicode so I have added the trait `ToCString` that is defined for both CString and the various string references and am using that here and elsewhere.

mqudsi:
@ridiculousfish I addressed most of the points that you raised. To clarify what the deal is with `curses` now (as that's where the majority of your concerns lie) and assuming you've seen my comments above for why it's more complicated than it looks: I've removed all the mutexes and have gone back to prepopulating the list of string capabilities. `Term::set(&self, ..)` continues to use interior mutability but I'm using a `RefCell<Option<Arc<CString>>>` (unfortunately you cannot have a `Cell<Arc<T>>` as it does not implement `Clone` to close a soundness hole [in certain rare and contrived cases](https://users.rust-lang.org/t/why-does-cell-require-copy-instead-of-clone/5769/3?u=mqudsi) so it's slightly larger than `Cell<Arc<CString>>` (which would be invisible and is a compile-time-only construct) but such is life).

There are a few remaining discussions above that aren't resolved, namely about the use of std parsing instead of wcstoi/wcstod but I've posted my rationale for going with the former and maintain that there are no observable changes to the behavior when we parse version numbers as an int vs as a double *for the specific values we care about* (and for the ones that we don't the result is never used).

I agree that I should try and maximize the number of individual, independent commits for this branch to try and isolate changes so much as possible - that's going to take some time and I think `git rebase -i --exec `ninja -C ../build` is going to be my friend.

fish-rust/src/curses.rs
```diff
@@ -0,0 +1,300 @@
...
+pub const EXIT_ITALICS_MODE: StringCap = STRING_CAPS[StringCapIdx::ExitItalicsMode.idx()];
+pub const ENTER_DIM_MODE: StringCap = STRING_CAPS[StringCapIdx::EnterDimMode.idx()];
+
+// Number capabilities
+pub const MAX_COLORS: Number = Number::new("Co");
+
+// Flag capabilities
+pub const EAT_NEWLINE_GLITCH: Flag = Flag::new("xn");
+
+pub trait Capability {
```
ridiculousfish, line 134: It's very confusing how some of the Capabilities look at their passed in `Term`, and others look at whatever `cur_term` has. The thread-safety story is completely broken here.

Again I think the simplest and best thing to do is to eagerly load all the capabilities we care about in `setupterm()`, while holding a lock.

fish-rust/src/curses.rs
```diff
@@ -0,0 +1,300 @@
...
+        } else {
+            TERM = None;
+            false
+        }
+    }
+}
+
+/// Whether or not the curses library has been initialized.
+pub fn is_initialized() -> bool {
+    unsafe { !sys::cur_term.is_null() }
```
ridiculousfish, line 228: This should look at our own TERM variable; we should only look at C's `cur_term` in `setupterm()`. Consider again what happens if this is called concurrently with `setupterm`.

fish-rust/src/curses.rs
```diff
@@ -0,0 +1,300 @@
...
+}
+
+/// Resets the curses `cur_term` TERMINAL pointer. Any previous term objects are invalidated!
+pub unsafe fn reset() {
```
ridiculousfish, line 232: This should not be `pub`; it needs to be synchronized with `setupterm`.

fish-rust/build.rs
```diff
@@ -131,3 +145,59 @@ fn detect_bsd() -> miette::Result<bool> {
...
+                libraries.push(library);
+                println!("cargo:warning=Found symbol {symbol} in lib{library}");
+                found += 1;
+                continue;
+            }
+        }
+    }
+    match found {
+        0 => {
+            println!("cargo:warning=gettext not found");
```
ridiculousfish, line 193: Please do not introduce this warning by default. I don't need to see this every time I build. We should only emit this warning if the user explicitly requests gettext and it's not found.

fish-rust/src/common.rs
```diff
@@ -1985,6 +1995,100 @@ pub fn get_by_sorted_name<T: Named>(name: &wstr, vals: &'static [T]) -> Option<&
     }
 }
 
+/// Takes ownership of a variable and `Deref`s/`DerefMut`s into a projection of that variable.
+///
+/// Can be used as a workaround for the lack of `MutexGuard::map()` to return a `MutexGuard`
+/// exposing only a variable of the Mutex-owned object.
+pub struct Projection<T, V, F1, F2>
```
ridiculousfish, line 2002: `Projection` isn't used anywhere and I don't understand what it does; please remove it.

fish-rust/src/common.rs
```diff
@@ -1985,6 +1995,100 @@ pub fn get_by_sorted_name<T: Named>(name: &wstr, vals: &'static [T]) -> Option<&
...
+    }
+}
+
+impl ToCString for &Utf32String {
+    fn to_cstring(self) -> CString {
+        self.as_utfstr().to_cstring()
+    }
+}
+
+impl ToCString for &String {
```
ridiculousfish, line 2086: I'm OK with `ToCString` but we should NOT have an implementation for `&String` or `&str`. There is no guarantee that the user's encoding us UTF8.

fish-rust/src/env_dispatch.rs
```diff
@@ -0,0 +1,814 @@
...
+        }
+    }
+
+    true
+}
+
+// Initialize the curses subsystem
+fn init_curses(vars: &dyn Environment) {
+    for var_name in CURSES_VARIABLES {
+        let name = var_name.to_string();
```
ridiculousfish, line 643: This is where `String::ToCString` gets used, but we should not introduce that because then we are adding APIs which fundamentally do the wrong thing, and future contributors will be tempted to use them.

`CURSES_VARIABLES` should contain `&wstr` or a byte string literal. We should not introduce types that assume UTF8 encoding like `str` or `String`.

fish-rust/src/env_dispatch.rs
ridiculousfish: It's a `size_t` because it effectively controls the size of a memory buffer. I'm OK expanding it to `u64` or whatever.

But (IMO) the more important difference is that `fish_wcstoull` recognizes a different set of strings than `parse()`. For example, the `fish_` variants skip leading whitespace while `parse` does not. So switching to `parse()` would technically be a breaking change; that is why we should continue to use the `fish_*` set of string parsing functions everywhere they appear in the current C++ code.

fish-rust/src/env_dispatch.rs
ridiculousfish: Moving towards standard library functions may be a worthy goal, but it should be done as a separately factored commit, not in this one.

fish-rust/src/supercow.rs
ridiculousfish, line 1: `supercow` is now unused; please remove this file.

ridiculousfish (changes requested):
I still have concerns about `curses.rs` and in particular the thread safety story.

The main scenario to think about is "what if one thread sets $TERM while another thread runs `set_color`." Then you have a race between reading and modifying `TERM`. We should make `curses.rs` not vulnerable to that race, especially because it's easy to do.

Basically the way to think about this is that all manipulations of the C `cur_term` need to be guarded by a lock. That includes `setupterm()`, `del_curterm()`, and any of the interrogation functions (`tigetflag`, etc).

The Rust Term should expose an immutable, thread-safe public interface. Having `term()` return an `Option<Arc<Term>>` makes sense to me. Then `setupterm()` would create a *new* Term.

(I realize this deviates a bit from "preserve existing C++ code" but we fundamentally can't directly port the C++ code, because it depends on lots of preprocessor defines, so we may as well get the Rust code right.)

fish-rust/Cargo.toml
```diff
@@ -33,6 +33,7 @@ cc = { git = "https://github.com/mqudsi/cc-rs", branch = "fish" }
 cxx-build = { git = "https://github.com/fish-shell/cxx", branch = "fish" }
 cxx-gen = { git = "https://github.com/fish-shell/cxx", branch = "fish" }
 miette = { version = "5", features = ["fancy"] }
+rsconf = { git = "https://github.com/mqudsi/rsconf", branch = "master" }
```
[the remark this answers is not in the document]
mqudsi: Yeah, I'll add the readme for sure. I spent most of the time adding doc comments, the readme comes next.

The `cc` crate is pretty foundational to the rust language and it's under the provenance of the rust-lang team (its github repo is `rust-lang/cc-rs`).

fish-rust/src/curses.rs
```diff
@@ -0,0 +1,300 @@
...
+    }
+}
+
+pub struct Term {
+    /// A list of cached string values corresponding to known curses string capabilities. Individual
+    /// values may be overridden with [`Term::set()`].
+    ///
+    /// The list is prepopulated on initialization in [`Term::new()`], as such the `Vec` itself is
+    /// never manipulated - only potentially the values within its individual cells.
+    strings: Vec<RefCell<Option<Arc<CString>>>>,
```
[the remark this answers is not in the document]
mqudsi: I think maybe where you and I are not on the same page is that *right now* the `Term` instance *is* prepopulated but **not immutably so**. As I mentioned previously, the way the C++ code sets up the terminal it is initialized in one place and at one point of time then the overrides are applied at another place at a later point in time. I *do* want to change that make it so that we have all our overrides available from the start to make it possible to just init an immutable `Term` (at least until the next one comes around), but that's not possible right now.

Your point about `RefCell` not being thread-safe is a good one but the `Arc` is there to allow the lifetime to extend past the `RefCell` borrow (in this case I believe the `RefCell` borrow is stripped out by the compiler that recognizes we only borrow to clone). I think the `RefCell` is safe to use because *in practice* we init and modify from the same thread at startup only, but that's not provable via the rust type system so I guess that means that we need to go back to a `Mutex` here after all.

The suggestion to use a boxed slice instead of a Vec is good; it makes it clearer that this structure is not intended to be manipulated and cannot be resized.

fish-rust/src/curses.rs
```diff
@@ -0,0 +1,300 @@
...
+        Term { strings }
+    }
+}
+
+#[repr(u8)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+// Clippy deduces the color of all the cows in Scotland from the first three entries we happen to
+// have here.
+#[allow(clippy::enum_variant_names)]
+enum StringCapIdx {
```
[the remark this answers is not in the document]
mqudsi: Again it's because of how the overrides are applied at a later time, so I would need to add both a getter and a setter for each one, which I am absolutely loathe to do. And for that same reason I can't have `Option<CString>` but need the `Arc` there. 

Basically everything is complicated by the fact that the C++ code init *then* modified the term overrides rather than doing them upfront and together. As mentioned, I want to change that so this can all be cleaner, but I wanted to be faithful to the C++ code for the initial commit then change it to be prettier and faster in a separate PR.

fish-rust/src/curses.rs
mqudsi: I didn't realize you wanted to preload numeric and flag capabilities as well. 

I considered the thread story but it's actually "safe" for non-string capabilities, though the results are not defined. If you assume that `cur_term` is swapped out (or zeroed) atomically (and that's a safe assumption on all platforms since the pointer size is guaranteed to be the size of an atomic read) then if that happens the worst that you end up with is that `Term` was initialized for `xterm-256color` but you get back the results for `tmux` (I don't know how curses does it internally and whether there are memory barriers that would prevent stale reads, etc. so maybe there's a chance that you'll end up with one or the other but not both?).

fish-rust/build.rs
mqudsi: This was for debugging for myself as I tried to figure this out.

fish-rust/src/curses.rs
mqudsi: I'm not sure what you mean by "synchronized with `setupterm`", as `del_curterm` is a valid call after `setupterm()` and we set both `TERM` to `None` as `cur_term` gets nulled out. Obviously if we end up adding a mutex then it'll need to use the mutex, though.

(It needs to be public because it's the equivalent of `env_cleanup()` and gets called by asan at exit.)

fish-rust/src/common.rs
mqudsi: IMHO that's all the more reason to *keep* that trait impl (but change it) because there should be an easy way to use `&str` or else people will just be `CString::new(s).unwrap()` with the `&str`. 

I'll change the impl to convert to a wide string then use that impl (which goes through `wstr2zstring()` or whatever its called and is safe).

fish-rust/src/common.rs
mqudsi: You reviewed this before I finished rebasing and squashing everything. `Projection<T>` is used now (to refactor the changes that were added to `ScopeGuard` to make it support `scoped_push()` that drastically increased its complexity and made it unreadable) and I have more uses for it in the short-term.

fish-rust/src/env_dispatch.rs
mqudsi: See my not above. I'll change the impl to convert to a CString safely, respecting the user's locale.

fish-rust/src/supercow.rs
mqudsi: That's fine. I have another use for it in a separate PR but I'll remove it for now.

mqudsi:
@ridiculousfish When I pushed my rebased/squashed commits I added some documentation to the `TERM` declaration explaining the safety. If you don't mind reading my individual replies above first, the idea is that another call to `setupterm()` might result in getting either a result for the old term or for the new term but shouldn't result in anything specifically unsafe (such that would lead to a segfault) since no function borrows from `Term`. The strings clone the `Arc<CString>` and the ints/flags just query against whatever `cur_term` happens to be (and return `None` if `cur_term` was `None`).

Note that I'm *not* saying that this is ideal, but just that this is the safest way that remains faithful to the existing C++ code (which did no synchronization and had no thread safety here whatsoever and in fact used `* const char` ptrs that pointed to data curses allocated against the term fd for the prev `$TERM` value that would be nulled out if `setupterm()` were called with a different `$TERM`).

I'm fine with adding a `Mutex` that protects manipulation of `$TERM` but I don't *think* it's necessary to safeguard calls to `tigetflag()` or `tigetnum()` behind that mutex since they don't return any pointers or reference any memory; either they'll get the old `cur_term` and serve a stale value, get a `nullptr` `cur_term` and return what we interpret as `None`, or see the new `cur_term` and return the new value. 

I *would* rather add a `Mutex` than return an `Arc` from `curses::term()` as I think it would be a cleaner api and should perform better when there's basically never any contention.

fish-rust/src/common.rs
krobelus: I don't like to introduce inconsistency / multiple obvious ways to do the same thing.
If this trait has advantages over a free function, then we should update the entire tree to use the `to_cstring` name and delete `wcs2zstring` including on the C++ side.

fish-rust/src/curses.rs
mqudsi: Sorry, I meant to drop this function altogether.

mqudsi:
@ridiculousfish I went ahead and restructured the calling code from `env_dispatch` so that we can get an actually immutable `Term` and forget all the interior mutability stuff. 

I started off with only an Arc but I realized I had to use both a Mutex and an Arc to prevent a potential race condition that would read freed memory in the case where we are resetting the Arc to a new value. If thread A first obtains the old value, another thread B swaps out the old value and frees the old, decrementing the refcount to zero in the process, then thread A gets around to cloning the arc, the arc will be zeroed out. So we need to use a mutex around the arc to prevent that from happening.

I don't think it'll be a bottle neck as we usually have just the one (same) thread accessing at any given time.

fish-rust/src/common.rs
```diff
@@ -1143,8 +1144,19 @@ pub fn wcs2osstring(input: &wstr) -> OsString {
...
-    if input.is_empty() {
+/// Converts a Unicode string into a [`CString`], taking into account the user's current locale. As
+/// our wide string input may represent non-unicode characters or else Unicode characters falling
+/// into the private-use range, we shift these into fish's own private use range to make
+/// round-tripping through a wide string a lossless operation. `wcs2zstring()` undoes that.
+///
+/// While this is ostensibly for wide string inputs, we look at the input string char-by-char and to
+/// perform this operation we necessarily allocate, so this can actually be used for any type that
+/// can represent a sequence of characters.
+pub fn wcs2zstring<IntoChars>(input: IntoChars) -> CString
```
ridiculousfish, line 1155: This function should *not* be generic over character sequences.

The only thing this function does is reverse the effects of `str2wcstring`, and add a nul terminator. If the argument did not go through `str2wcstring` (or is a widestring literal), then it should not be passed to `wcs2zstring`. And `str2wcstring` gives us a `WString` so this should accept a `&wstr`; any other argument is likely a latent bug.

This should continue to take a `&wstr` (that also allows you to trivially implement the empty input optimization if you want).

fish-rust/src/common.rs
```diff
@@ -1742,44 +1763,24 @@ pub type Cleanup<T, F, C> = ScopeGuard<T, F, C>;
 ///
 /// // hello will be written first, then goodbye.
 /// ```
-pub struct ScopeGuard<T, F: FnOnce(&mut T), C> {
+pub struct ScopeGuard<T, F: FnOnce(&mut T)> {
```
ridiculousfish, line 1766: FYI I don't object to you merging this since I recognize implementing `scoped_cleanup` is hard in Rust. But it ends up looking pretty big and confusing in Rust Analyzer, (`&mut &mut &mut...`) so I will likely take a run at simplifying this.

Screenshot 2023-05-21 at 2 05 48 PM

fish-rust/src/wcstringutil.rs
```diff
@@ -248,7 +248,10 @@ pub fn string_fuzzy_match_string(
 /// This invokes \p func with (const char*, size_t) pairs.
 /// If \p func returns false, it stops; otherwise it continues.
 /// \return false if the callback returned false, otherwise true.
-pub fn wcs2string_callback(input: &wstr, mut func: impl FnMut(&[u8]) -> bool) -> bool {
+pub fn wcs2string_callback<IntoChars>(input: IntoChars, mut func: impl FnMut(&[u8]) -> bool) -> bool
```
ridiculousfish, line 251: This should only be called with stuff that was either a wide string literal, or came from `str2wcstring`. Any other argument is likely a bug; therefore this should continue to take `&wstr`.

fish-rust/src/curses.rs
```diff
@@ -0,0 +1,276 @@
...
+/// no guarantee that a simultaneous call to [`setup()`] won't result in this refcount being
+/// decremented to zero and the memory being reclaimed before we can clone it, since we can only
+/// atomically *read* the value of the pointer, not clone the `Arc` it points to.
+pub static TERM: Mutex<Option<Arc<Term>>> = Mutex::new(None);
+
+/// Returns a reference to the global [`Term`] singleton or `None` if not preceded by a successful
+/// call to [`curses::setup()`].
+pub fn term() -> Option<Arc<Term>> {
+    const NULL: *mut Arc<Term> = core::ptr::null_mut();
+    TERM.lock()
```
ridiculousfish, line 34: nit: consider the simpler `TERM.lock().expect("Mutex poisoned!").clone()`

ridiculousfish (changes requested):
The curses changes are looking good! Just some nits there.

This makes a lot of changes to make it easier to use `&str` interchangeably with `&wstr`. These changes are wrong: fish avoids using `&str` for reasons spelled out in the "Strings" section of our Rust development guide (`rust-devel.md`). I'm going to insist we do not introduce this `&str` support.

ridiculousfish (changes requested):
The curses changes are looking good! Just some nits there.

This makes a lot of changes to make it easier to use `&str` interchangeably with `&wstr`. These changes are wrong: fish avoids using `&str` for reasons spelled out in the "Strings" section of our Rust development guide (`rust-devel.md`). I'm going to insist we do not introduce this `&str` support.

To add some context: I'm being a stickler here because, if we do allow `&str` to creep in, then it is likely to introduce bugs that will be hard to find. Example: an environment variable value that is not valid UTF-8 - rare but not impossible! And we only have test coverage for a few of these cases.

If we confine ourselves to UTF32 wide strings, and narrow "raw byte" strings, then we can be sure these bugs will not occur. If we throw `&str` into the mix, then we are likely to introduce latent bugs. And `&str` doesn't really add that much value anyways.

fish-rust/src/curses.rs
```diff
@@ -0,0 +1,276 @@
...
+            // Number capabilities
+            max_colors: NumberCap::new("Co").lookup(),
+
+            // Flag/boolean capabilities
+            eat_newline_glitch: FlagCap::new("xn").lookup(),
+        }
+    }
+}
+
+pub trait Capability {
```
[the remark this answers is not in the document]
mqudsi: This (and the others) is left over from when it was required. Thanks for catching it.

fish-rust/src/curses.rs
```diff
@@ -0,0 +1,276 @@
...
+        *global_term = Some(term.clone());
+        Some(term)
+    } else {
+        *global_term = None;
+        None
+    }
+}
+
+/// Whether or not the curses library has been initialized.
+pub fn is_initialized() -> bool {
```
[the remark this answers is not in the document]
mqudsi: Yeah, this function functionally became obviated when `term()` began returning an `Option<Term>` instead of `&Term` or `panic!()`.

fish-rust/src/common.rs
```diff
@@ -1985,6 +1969,122 @@ pub fn get_by_sorted_name<T: Named>(name: &wstr, vals: &'static [T]) -> Option<&
     }
 }
 
+/// Takes ownership of a variable and `Deref`s/`DerefMut`s into a projection of that variable.
+///
+/// Can be used as a workaround for the lack of `MutexGuard::map()` to return a `MutexGuard`
```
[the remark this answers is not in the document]
mqudsi: That's what I was using it for before using it for scoped_push then no longer using it for the mutex. But it's a good example, imho.

fish-rust/src/common.rs
```diff
@@ -1985,6 +1969,122 @@ pub fn get_by_sorted_name<T: Named>(name: &wstr, vals: &'static [T]) -> Option<&
...
+
+impl ToCString for &CStr {
+    fn to_cstring(self) -> CString {
+        self.to_owned()
+    }
+}
+
+/// Safely converts from `&wstr` to a `CString`, taking into account non-Unicode values that have
+/// been shifted into the private-use range by using [`wcs2zstring()`].
+impl ToCString for &wstr {
```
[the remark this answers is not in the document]
mqudsi: You cannot do that without giving up all other impls because "upstream may someday add AsRef<wstr> to CString, making it ambiguous which trait impl should be used."

fish-rust/src/common.rs
```diff
@@ -1985,6 +1969,122 @@ pub fn get_by_sorted_name<T: Named>(name: &wstr, vals: &'static [T]) -> Option<&
...
+where
+    F1: Fn(&T) -> &V,
+    F2: Fn(&mut T) -> &mut V,
+{
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        (self.view_mut)(&mut self.value)
+    }
+}
+
+/// A trait to make it more convenient to pass ascii/Unicode strings to functions that can take
```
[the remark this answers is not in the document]
mqudsi: To be pedantic, that's why *CString* exists, not this trait. But yeah, sure.

fish-rust/src/common.rs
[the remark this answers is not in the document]
mqudsi: That's what I was using it for before using it for scoped_push then no longer using it for the mutex. But it's a good example, imho.

fish-rust/src/env_dispatch.rs
```diff
@@ -0,0 +1,826 @@
...
+    let version = vars
+        .get(L!("TERM_PROGRAM_VERSION"))
+        .map(|v| v.as_string())
+        .and_then(|v| {
+            let mut consumed = 0;
+            crate::wutil::wcstod::wcstod(&v, '.', &mut consumed).ok()
+        })
+        .unwrap_or(0.0);
+
+    match term.to_string().as_str() {
```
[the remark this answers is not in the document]
mqudsi: We are comparing with an ASCII string so I used whatever was most convenient. I don't think it's fair to look at the code without that context in mind. Generally the receiver decides what type it wants (i.e. a function takes an &wstr that is the variable it wants) and the caller knows to provide it in that format. Here the destination is an ascii &str so it is clearly fine. But sure.

mqudsi:
@ridiculousfish I'll be honest: I fundamentally disagree with you on the `&str` stuff. To be sure, I understand your concerns. But `&str` doesn't come from out of the blue - either it's an ascii value/string literal that's created on the spot and you can do anything with it because it's well-defined (the vast majority of the cases and what we're dealing with the most here), or it comes from `wstr::to_string()` in which case it can be treated identically to a &wstr including reversing the private range stuff (which is what I was doing before). And, yes, it could also  come from the environment (e.g. `std::env::var()`), but in that case it would involve *explicitly unwrapping* to assert that the unknown contents coming from the outside world don't contain any non-unicode values, which should be sign enough that you're doing something wrong.

The problem is that you're never going to make it *impossible* to use &str, you're just going to make it inconvenient. i.e. you asked me to remove the `ToCString` impls for `&str` but before we had those, I was using `CString::new(str)` everywhere to convert the ascii string to a CString. *That* option remains available even now *and it's less universally correct* than the `ToCString` impl (which I've now removed) that at least transformed the private-range-encoded non-unicode values back (assuming the string came from a wstr in the first place) and put non-ascii strings into the correct locale.

I understand your concerns but as long as the easiest string to use in rust is `&str` and getting an `&str` from a `&wstr` is as simple as calling `to_string()` it's kind of a losing battle. And as I've said, the odds are actually that the implementation will be correct because that string has to come from somewhere and we can make sure that all those "somewheres" are valid (and from ascii, a raw string literal, or from wstr are all "correct" in that sense).

In all cases, I'm deferring to you and have removed the `&str` usage even when I'm comparing an ascii value to an ascii value (which makes no sense to me since it's clear that the type doesn't matter there).

FYI the commits were already rebased and squashed and each already individually passed all checks. I've pushed the requested changes and rebased/squashed accordingly so this branch should be OK to merge commit.

ridiculousfish (approved):
LGTM. Nice work.

Re: strings, if you think there are acceptable uses of String and &str, start by making n PR against `rust-devel.md` to explain what those uses are, and when to prefer &str over &wstr.

My view is that the only acceptable use is string literals for comparing against &wstr, and perhaps when displaying error messages. Any other use is probably wrong. For example, passing a `&str` to `File::open()` is almost always wrong.

Look at nushell: a Rust shell that uses strings freely. It simply cannot handle non-UTF8 filenames. There are files you cannot `rm` or `mv`. There are files you cannot `ls` - it just pretends that the file doesn't exist. And as far as I can tell this is basically unfixable: it depends on crates like `glob` which explicitly only handle UTF-8. I will not allow fish to suffer the same fate.

I don't agree it's a losing battle: we'll keep supporting non-UTF8 data even if it means adding crazy lint rules or other heroics. "fish can handle every file bash can handle" is a strict requirement.

Now if you have a way to use String that preserves that requirement, I would legitimately love to hear it.

Xiretza:
I don't understand why there would be a difference with regards to "There are files you cannot `rm` or `mv`"-style problems between UTF-8 (`&str`) and UTF-32 (`&wstr`). Both can represent all unicode codepoints, and only unicode codepoints; they require workarounds such as the Private Use Area characters to represent things other than "a stream of unicode code points" (such as the raw byte sequences that are paths etc).

ridiculousfish:
@Xiretza yes you are correct. We could use Rust Strings, and attempt to enforce that String is never passed to a function like `File::open` or `fs::Metadata`. I think that would be more difficult to enforce than just not creating Strings in the first place. But I'm open to alternative proposals.

Xiretza:
Personally I'd model that as wrapper types around `&str`/`String`. They can be 
- constructed from normal strings (no-op) or byte sequences (with escaping)
- compared to normal strings (trivial) or byte sequences (taking into account escaping)
- converted back to byte sequences (with unescaping)
- but not converted back to normal strings without explicitly acknowledging that they may contain escape characters. These calls can then be audited easily.

mqudsi:
I personally wasn't commenting on the general use of `&str` vs `&wstr` (which I was saving for after the port, though I agree with @Xiretza that they're fundamentally no different aside from one being slightly more difficult to lint against than the other) but rather just a specific subset of string operations.

In particular, my main argument was that the use of `&str` with *string literals* (regardless of ascii or unicode) should be OK on their own in all cases (as they are self-contained and internal to the codebase, not subject to the external environment, though with the caveat that you might have to convert to the user's locale, which the `ToCString` impl attempted to do). And to a lesser extent, as an extension of that, (correctly) converting external strings to `&str` for comparison with `&str` string literals should also be allowed, since it's transitively safe (if the external input contained non-unicode then it can't ever match against a string literal which by definition cannot).

@ridiculousfish I'm not sure if you saw my question [regarding `scoped_push()`](https://github.com/fish-shell/fish-shell/pull/9786#discussion_r1202691752)?

ridiculousfish:
I don't really understand your question yet, I haven't looked too closely at Projection and scoped_push; I'll do that before long.

PR #9800, opened 2023-05-17
update npm completions

### LINKED ISSUE #9799: outdated npm completions
the completions in [/share/completions/npm.fish](https://github.com/fish-shell/fish-shell/blob/master/share/completions/npm.fish) are incorrect, outdated and incomplete. i have tested a bit and written some code and could almost have a pr ready to fix those issues (if you want). that pr would:

1. replace the built-in `npm uninstall` completions (from `$ npm completion`) with the one that is already used by `yarn` for local packages and a new one for global packages (`npm uninstall -g`)
2. update and remove outdated options and commands
3. add missing command aliases, commands and options

since this is a more drastic change than just a one-line fix (and npm has quite a large number of commands and aliases) i thought it would probably be better to open an issue first so i know if i should make the pr.

if i make the pr and it gets merged i could also update and expand on the yarn completions, but since yarn is much less used and seems more up-to-date i wanted to fix npm first

let me know if i should open that pr. ty.

```
$ fish --version
fish, version 3.6.1-519-g67d1d80f9
```

m4rch3n1ng:
## Description

fixes #9799

1. replaces the built-in `npm uninstall` completions (from `$ npm completion`) with the one that is already used by `yarn` for local packages and a new one for global packages (`npm uninstall -g`)
2. updates and remove outdated options and commands
3. adds missing command aliases, commands and options

### notes

the global completions are done [here](https://github.com/fish-shell/fish-shell/blob/84545861d030adc46a73e094d7ffcbbc584d1099/share/functions/__fish_npm_helper.fish#L53)

this is not fully done, i still need to update and complete for more npm commands

## TODOs:

### DIFF
--- a/share/completions/npm.fish
+++ b/share/completions/npm.fish
@@ -127,11 +127,18 @@ complete -f -c npm -n __fish_npm_needs_command -a get -d 'Echo the config value 
 complete -f -c npm -n __fish_npm_needs_command -a set -d 'Sets the config key to the value'
 
 # install
-for c in install isntall i
-    complete -c npm -n __fish_npm_needs_command -a "$c" -d 'install a package'
-    complete -c npm -n "__fish_npm_using_command $c" -l save-dev -d 'Save to devDependencies in package.json'
-    complete -c npm -n "__fish_npm_using_command $c" -l save -d 'Save to dependencies in package.json'
-    complete -c npm -n "__fish_npm_using_command $c" -s g -l global -d 'Install package globally'
+for c in install add i 'in' ins inst insta instal isnt isnta isntal isntall
+    complete -c npm -n __fish_npm_needs_command -a "$c" -d 'Install a package'
+    complete -f -c npm -n "__fish_npm_using_command $c" -s B -l save-bundle -d 'Also save to bundleDependencies in package.json'
+    complete -f -c npm -n "__fish_npm_using_command $c" -s E -l save-exact -d 'Save dependency with exact version rather than a semver range'
+    complete -f -c npm -n "__fish_npm_using_command $c" -l no-save -d 'Prevents saving to dependencies in package.json'
+    complete -f -c npm -n "__fish_npm_using_command $c" -s O -l save-optional -d 'Save to optionalDependencies in package.json'
+    complete -f -c npm -n "__fish_npm_using_command $c" -s P -l save-prod -d 'Save to dependencies in package.json'
+    complete -f -c npm -n "__fish_npm_using_command $c" -s D -l save-dev -d 'Save to devDependencies in package.json'
+    complete -f -c npm -n "__fish_npm_using_command $c" -s g -l global -d 'Install package globally'
+    complete -x -c npm -n "__fish_npm_using_command $c" -l omit -a 'dev optional peer' -d 'Omit dependency type'
+    complete -f -c npm -n "__fish_npm_using_command $c" -l ignore-scripts -d 'Do not run scripts specified in package.json'
+    complete -f -c npm -n "__fish_npm_using_command $c" -l install-links -d 'Pack and install file: protocol dependences as regular dependencies instead of symlinks'
 end
 
 # list
--- a/share/completions/npm.fish
+++ b/share/completions/npm.fish
@@ -129,6 +129,12 @@ end
 complete -f -c npm -n __fish_npm_needs_command -a get -d 'Echo the config value to stdout'
 complete -f -c npm -n __fish_npm_needs_command -a set -d 'Sets the config key to the value'
 
+# init
+for c in init create innit
+    complete -c npm -n __fish_npm_needs_command -a "$c" -d 'Create a package.json file'
+    complete -f -c npm -n "__fish_npm_using_command $c" -s y -l yes -d 'Automatically answer "yes" to all prompts'
+end
+
 # install
 for c in install add i 'in' ins inst insta instal isnt isnta isntal isntall
     complete -c npm -n __fish_npm_needs_command -a "$c" -d 'Install a package'
--- a/share/completions/npm.fish
+++ b/share/completions/npm.fish
@@ -162,8 +162,7 @@ for c in r remove rm un uninstall unlink
     complete -f -c npm -n __fish_npm_needs_command -a "$c" -d 'remove package' -xa '(__yarn_installed_packages)'
     complete -x -c npm -n "__fish_npm_using_command $c" -s g -l global -d 'remove global package'
     complete -x -c npm -n "__fish_npm_using_command $c" -l save -d 'Package will be removed from your dependencies'
-    complete -x -c npm -n "__fish_npm_using_command $c" -l save-dev -d 'Package will be removed from your devDependencies'
-    complete -x -c npm -n "__fish_npm_using_command $c" -l save-optional -d 'Package will be removed from your optionalDependencies'
+    complete -x -c npm -n "__fish_npm_using_command $c" -l no-save -d 'Do not remove package from your dependencies'
 end
 
 # search
--- a/share/completions/npm.fish
+++ b/share/completions/npm.fish
@@ -196,7 +196,7 @@ complete -x -c npm -n '__fish_npm_using_command publish' -l otp -d '2FA one-time
 # remove
 for c in uninstall unlink remove rm r un
     complete -f -c npm -n __fish_npm_needs_command -a "$c" -d 'Remove a package'
-    complete -x -c npm -n "__fish_npm_using_command $c" -d 'remove package' -a '(__yarn_installed_local_packages)'
+    complete -x -c npm -n "__fish_npm_using_command $c" -d 'remove package' -a '(__npm_installed_local_packages)'
     complete -x -c npm -n "__fish_npm_using_command $c" -s g -l global -d 'remove global package' -a '(__npm_installed_global_packages)'
     complete -x -c npm -n "__fish_npm_using_command $c" -l save -d 'Package will be removed from your dependencies'
     complete -x -c npm -n "__fish_npm_using_command $c" -l no-save -d 'Do not remove package from your dependencies'
@@ -262,4 +262,4 @@ complete -f -c npm -n __fish_npm_needs_command -a unstar -d 'Remove star from a 
 complete -f -c npm -n __fish_npm_needs_command -a 'version verison' -d 'Bump a package version'
 complete -f -c npm -n __fish_npm_needs_command -a 'view info v show' -d 'View registry info'
 complete -f -c npm -n __fish_npm_needs_command -a whoami -d 'Display npm username'
-complete -f -c npm -n '__fish_seen_subcommand_from add i install; and not __fish_is_switch' -a "(__yarn_filtered_list_packages \"$npm_install\")"
+complete -f -c npm -n '__fish_seen_subcommand_from add i install; and not __fish_is_switch' -a "(__npm_filtered_list_packages \"$npm_install\")"
--- a/share/completions/npm.fish
+++ b/share/completions/npm.fish
@@ -166,6 +166,14 @@ complete -f -c npm -n '__fish_npm_using_command pack' -l dry-run -d 'Only report
 complete -f -c npm -n '__fish_npm_using_command pack' -l json -d 'Output JSON data'
 complete -c npm -n '__fish_npm_using_command pack' -l pack-destination -d 'Tarball destination directory'
 
+# publish
+complete -f -c npm -n __fish_npm_needs_command -a publish -d 'Publish a package'
+complete -x -c npm -n '__fish_npm_using_command publish' -l tag -d 'Upload to tag'
+complete -x -c npm -n '__fish_npm_using_command publish' -l access -d 'Restrict access' -a "public\t'Publicly viewable' restricted\t'Restricted access (scoped packages only)'"
+complete -f -c npm -n '__fish_npm_using_command publish' -l dry-run -d 'Only report what it would have done'
+complete -x -c npm -n '__fish_npm_using_command publish' -l otp -d '2FA one-time password'
+## todo provenance ?
+
 # remove
 for c in r remove rm un uninstall unlink
     complete -f -c npm -n __fish_npm_needs_command -a "$c" -d 'Remove a package'
@@ -204,7 +212,6 @@ complete -f -c npm -n __fish_npm_needs_command -a 'link ln' -d 'Symlink a packag
 complete -f -c npm -n __fish_npm_needs_command -a outdated -d 'Check for outdated packages'
 complete -f -c npm -n __fish_npm_needs_command -a prefix -d 'Display NPM prefix'
 complete -f -c npm -n __fish_npm_needs_command -a prune -d 'Remove extraneous packages'
-complete -c npm -n __fish_npm_needs_command -a publish -d 'Publish a package'
 complete -f -c npm -n __fish_npm_needs_command -a 'rb rebuild' -d 'Rebuild a package'
 complete -f -c npm -n __fish_npm_needs_command -a 'root ' -d 'Display npm root'
 complete -f -c npm -n __fish_npm_needs_command -a 'run-script run' -d 'Run arbitrary package scripts'
--- a/share/completions/npm.fish
+++ b/share/completions/npm.fish
@@ -115,11 +115,11 @@ end
 # access
 set -l access_commands 'list get set grant revoke'
 complete -f -c npm -n __fish_npm_needs_command -a access -d 'Set access level on published packages'
-complete -x -c npm -n "__fish_npm_using_command access; and not __fish_seen_subcommand_from $access_commands" -a list -d 'List access info'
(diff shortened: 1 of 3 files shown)


share/completions/npm.fish
```diff
@@ -69,9 +69,13 @@ end
...
 # except options completion (because it sucks at it)
 # and run-script completion (reading package.json is faster).
+# and uninstall (reading dependencies is faster and more robust and yarn needs the functions anyways)
 # see: https://github.com/npm/npm/issues/9524
 # and: https://github.com/fish-shell/fish-shell/pull/2366
-complete -f -c npm -n 'not __fish_npm_needs_option; and not __fish_npm_using_command run; and not __fish_npm_using_command run-script' -a "(__fish_complete_npm)"
+complete -f -c npm -n 'not __fish_npm_needs_option; and not __fish_npm_using_command run; and not __fish_npm_using_command run-script;
+and not __fish_npm_using_command audit;
+and not __fish_npm_using_command r; and not __fish_npm_using_command remove; and not __fish_npm_using_command rm; and not __fish_npm_using_command un;
+and not __fish_npm_using_command uninstall; and not __fish_npm_using_command unlink' -a "(__fish_complete_npm)"
```
faho, line 78: How about we just fix `__fish_npm_using_command` to check multiple commands?

Just change this

https://github.com/fish-shell/fish-shell/blob/67d1d80f94107e26585d192d194295180cb9f73b/share/completions/npm.fish#L24

to

```fish
if contains -- $cmd[2] $argv
```

Then this can be simplified to

```fish
not __fish_npm_needs_option; and not __fish_npm_using_command run run-script audit r remove rm un uninstall unlink
```

share/completions/npm.fish
```diff
@@ -102,18 +106,33 @@ for k,v in data["scripts"].items(): print(k + "\t" + v[:18])' <package.json 2>/d
...
     complete -f -c npm -n "__fish_npm_using_command $c" -a "(__fish_npm_run)"
 end
 
+# audit
+complete -f -c npm -n __fish_npm_needs_command -a audit -d 'Run a security audit'
+complete -f -c npm -n '__fish_npm_using_command audit' -a signatures -d 'Verify registry signatures'
+complete -f -c npm -n '__fish_npm_using_command audit' -a fix -d 'Automatically install any compatible updates to vulnerable dependencies'
+complete -x -c npm -n '__fish_npm_using_command audit' -l audit-level -a 'info low moderate high critical none' -d 'Audit level'
+complete -f -c npm -n '__fish_npm_using_command audit' -l dry-run -d 'Only report what it would have done'
+complete -f -c npm -n '__fish_npm_using_command audit' -l force -d 'Removes various protections against unfortunate side effects, common mistakes, and malicious input'
```
faho, line 120: This description should be shorter - just "Remove various protections"

share/completions/npm.fish
```diff
@@ -102,18 +106,33 @@ for k,v in data["scripts"].items(): print(k + "\t" + v[:18])' <package.json 2>/d
...
-for c in run run-script
+for c in run-script run rum urn
+    complete -f -c npm -n __fish_npm_needs_command -a "$c" -d 'Run arbitrary package scripts'
     complete -f -c npm -n "__fish_npm_using_command $c" -a "(__fish_npm_run)"
 end
 
+# audit
+complete -f -c npm -n __fish_npm_needs_command -a audit -d 'Run a security audit'
+complete -f -c npm -n '__fish_npm_using_command audit' -a signatures -d 'Verify registry signatures'
+complete -f -c npm -n '__fish_npm_using_command audit' -a fix -d 'Automatically install any compatible updates to vulnerable dependencies'
```
faho, line 117: "Install compatible updates to vulnerable deps"

share/completions/npm.fish
```diff
@@ -102,18 +106,33 @@ for k,v in data["scripts"].items(): print(k + "\t" + v[:18])' <package.json 2>/d
...
+complete -x -c npm -n '__fish_npm_using_command audit' -l audit-level -a 'info low moderate high critical none' -d 'Audit level'
+complete -f -c npm -n '__fish_npm_using_command audit' -l dry-run -d 'Only report what it would have done'
+complete -f -c npm -n '__fish_npm_using_command audit' -l force -d 'Removes various protections against unfortunate side effects, common mistakes, and malicious input'
+complete -f -c npm -n '__fish_npm_using_command audit' -l json -d 'Show information in JSON format'
```
faho, line 121: "Output JSON"

m4rch3n1ng:
one thing i would also like to know if i should fully replace this with manually written completions:
https://github.com/fish-shell/fish-shell/blob/84545861d030adc46a73e094d7ffcbbc584d1099/share/completions/npm.fish#L40-L77

it can get excruciatingly slow, doesn't have labels / explanations, and we already need to completely write the options manually anyways (which has already been noted [here](https://github.com/fish-shell/fish-shell/blob/84545861d030adc46a73e094d7ffcbbc584d1099/share/completions/npm.fish#L70-L71)).

i already replaced it for `npm audit` in https://github.com/fish-shell/fish-shell/blob/4d089e14c4ff941fa74a6f168a4b0241d210de5c/share/completions/npm.fish#L116-L117 with significant performance gains and added descriptions for what they do.

keeping it would also mean having to keep this ugly code, so it doesn't interfere with the hand-written stuff for `npm run` and `npm uninstall`
https://github.com/fish-shell/fish-shell/blob/4d089e14c4ff941fa74a6f168a4b0241d210de5c/share/completions/npm.fish#L75-L78

share/functions/__fish_npm_helper.fish
```diff
@@ -50,7 +50,25 @@ function __yarn_find_package_json
     return 1
 end
 
-function __yarn_installed_packages
+function __npm_installed_global_packages
+	set -l prefix (npm prefix -g)
+	set -l node_modules "$prefix/lib/node_modules"
+
+	for path in $node_modules/*
+		set -l mod (string split '/' $path)[-1]
```
faho, line 58:
```suggestion
		set -l mod (path basename -- $path)
```

share/functions/__fish_npm_helper.fish
```diff
@@ -50,7 +50,25 @@ function __yarn_find_package_json
...
+
+		if test (string match -r "^@" $mod)
```
faho, line 60:
```suggestion
		if string match -rq "^@" $mod
```


*NEVER* use single-argument test. It is the devil.

share/completions/npm.fish
```diff
@@ -150,58 +182,87 @@ complete -f -c npm -n '__fish_npm_using_command owner' -a ls -d 'List package ow
...
+complete -x -c npm -n '__fish_npm_using_command publish' -l otp -d '2FA one-time password'
+## todo provenance ?
+
 # remove
-for c in r remove rm un uninstall unlink
-    complete -f -c npm -n __fish_npm_needs_command -a "$c" -d 'remove package' -xa '(__yarn_installed_packages)'
-    complete -x -c npm -n "__fish_npm_using_command $c" -s g -l global -d 'remove global package'
+for c in uninstall unlink remove rm r un
+    complete -f -c npm -n __fish_npm_needs_command -a "$c" -d 'Remove a package'
+    complete -x -c npm -n "__fish_npm_using_command $c" -d 'remove package' -a '(__yarn_installed_local_packages)'
```
faho, line 202: Hold on, is it right that this npm completion tries the yarn installed packages?

share/completions/npm.fish
```diff
@@ -126,16 +145,29 @@ end
...
+    complete -f -c npm -n "__fish_npm_using_command $c" -s D -l save-dev -d 'Save to devDependencies in package.json'
+    complete -f -c npm -n "__fish_npm_using_command $c" -s g -l global -d 'Install package globally'
+    complete -x -c npm -n "__fish_npm_using_command $c" -l omit -a 'dev optional peer' -d 'Omit dependency type'
+    complete -f -c npm -n "__fish_npm_using_command $c" -l ignore-scripts -d 'Do not run scripts specified in package.json'
+    complete -f -c npm -n "__fish_npm_using_command $c" -l install-links -d 'Pack and install file: protocol dependences as regular dependencies instead of symlinks'
 end
 
 # list
-for c in la list ll ls
+for c in ls list ll la
```
faho, line 170: Why this change?

faho (changes requested):

faho:
>one thing i would also like to know if i should fully replace this with manually written completions:

If you're up for it. I gather people tired of writing npm completions, and tbh I can't blame them. I'd hate having to list all the awkward misspellings.

m4rch3n1ng:
> If you're up for it. I gather people tired of writing npm completions, and tbh I can't blame them. I'd hate having to list all the awkward misspellings.

if you're fine with it i'll probably do that while going through all the options for the commands anyways. i need to list the awkward misspellings and aliases anyways if i want to also have completions for them. on the previous version `npm in` for example did not have any option completions and now they do.

share/completions/npm.fish
faho: I noticed you didn't add it, I just thought it was suspect and might be a mistake.

Fine to leave it as-is.

m4rch3n1ng: should i rename it (and all its derivatives) to `__yarn_`, it probably makes more sense for yarn to be using npm functions than the other way around?

faho: Yeah, seems right.

share/completions/npm.fish
m4rch3n1ng: will do, the problem does not get that much better tho since i actually forgot to include all the aliases / misspellings for the command (there are only 2/4 for `npm run` and 3/6 for `npm uninstall`).

looking through the code and testing i actually think that this code runs on any tab without doing anything almost ever. i looked and found that i had missed 1 alias that it was auto completing and i don't see literally anything else. the only thing other than commands that are auto completed through this function are superseded and overwritten [here](https://github.com/fish-shell/fish-shell/blob/84545861d030adc46a73e094d7ffcbbc584d1099/share/completions/npm.fish#L121-L126), [here](https://github.com/fish-shell/fish-shell/blob/84545861d030adc46a73e094d7ffcbbc584d1099/share/completions/npm.fish#L114-L116) and [here](https://github.com/fish-shell/fish-shell/blob/4d089e14c4ff941fa74a6f168a4b0241d210de5c/share/completions/npm.fish#L116-L117). removing it will at this point probably only improve overall performance.

m4rch3n1ng:
one more consideration would be actually removing the completions for `npm install` since those are entirely unusable for almost anything that is not scoped (i.e. `@m4rch/`) since it is sorted alphabetically and takes like almost two seconds to pull up on my machine. i am not tied to this was just throwing it out as a suggestion since it is also annoying if you accidentally `<tab>` and is also another place where i have to list all the aliases (and forgot to list them).

*attempting to install `"vite"` with the autocomplete*

image
image

m4rch3n1ng:
converting to draft since i still want to add most (if not all) of the other completions, just wanted to pr already to get some feedback.

m4rch3n1ng:
one more quick feedback: should i consistently use `"dep(s)"` over `"dependenc(y/ies)"` in descriptions for options or do i stick to the more "official" but more verbose naming?

faho:
>should i consistently use "dep(s)" over "dependenc(y/ies)" in descriptions for options or do i stick to the more "official" but more verbose naming?

Keep descriptions short. The shorter the descriptions are, the more likely it is we can fit more columns.

Use common abbreviations like "deps" over "dependencies", but don't invent your own or do awkward initialisms.

It's fine if the description isn't 100% exhaustively accurate, it'll either remind people who already know or let them look it up if it seems to them that they might want it.

m4rch3n1ng:
is there a way to suggest to the completion that it should only complete for directories and not all files?

faho:
> is there a way to suggest to the completion that it should only complete for directories and not all files?

`__fish_complete_directories` - i.e. use `complete -f` and then create the arguments with that function.

share/completions/npm.fish
m4rch3n1ng: another issue with this implentation is that it actually doesn't properly work.

here's the generated completion for `npm help`
image

but if i type anything and try to complete that it doesn't want to
image

m4rch3n1ng:
one more note on https://github.com/fish-shell/fish-shell/pull/9800#issuecomment-1551882361: it currently only works with plain `npm install`, if we want to keep it, should i also add it to support when running `npm install -D`, `npm install -B`, etc.?

faho:
>one more note on https://github.com/fish-shell/fish-shell/pull/9800#issuecomment-1551882361:

I told you if you're up for replacing it to replace it. Adding more to it violates my firm belief in the lipstick-on-a-pig principle.

m4rch3n1ng:
ok ty for pinging, i will wait if they approve.

on an unrelated note: how do i add "positional" arguments for completions? i need some way to define `<predefined-word>` for `npm view <user-input> <predefined-word>`.

faho:
>on an unrelated note: how do i add "positional" arguments for completions?

You'll have to see if that argument is already there or not - as a shortcut you can sometimes count the arguments (via `__fish_is_nth_token`).

Unfortunately often you can't, and solving it in general is gnarly - it's only ever possible if you know the options and which of them take arguments, so you'd need something like `argparse`. (the general problem is this: in `git -C status stash oush -m save log`, which is the subcommand?)

m4rch3n1ng:
ty, since it is fairly simple a `-n "__fish_npm_using_command $c; and __fish_is_nth_token 3"` worked. very happy someone else has already done the git completions, those sound like a pain tbh

faho:
>npm has npm search but the all-the-package-names approach is significantly faster

What is being asked here is that all-the-package-names is apparently slow enough, and there are sooooo many packages and no descriptions that it's not helpful and maybe it's a good idea to just remove that completion, and to let people look up the package name elsewhere.

> there's no reason you should end up using the wrong option if you're on fish and the more options we add to the fish completions the more junk you have to sift through

What I'd do here is to *accept* the typo options, but to not offer them.

E.g. you would never offer `npm rum` or `npm urn`, but if someone does `npm rum <TAB>` you'd complete it like `npm run`.

That's easy to do, just add it in the `using_command foo` part.

m4rch3n1ng:
> `npm` has a bunch of typo or shortcut flags/switches. I don't bother supporting them because the way I see it, those help you when you are using a dumb shell like `sh` that can't give you any completions and to protect against mistakes from shells with an inferior UX, but `fish` can tell you what options are available and so there's no reason you should end up using the wrong option if you're on `fish` **and** the more options we add to the fish completions the more junk you have to sift through to actually find a completion (because every shortcut and misspelling supported by `npm` is going to appear in the list of possible completions) so I purposely omitted those.

that's actually a fair consideration and i agree with @faho on this one, but the initial reason i opened this pr in the first place is being annoyed that i frequently didn't get option completions for commands because i misspelled them, (the main example being typing `npm in`). also because you already used the completions from `npm completions` those misspelling still showed up, they just didn't have descriptions.  
i might actually remove those duplicates from the suggestions tho (except for a few frequently used ones like `run` and `i`)

mqudsi:
I think I would prefer that. If you find yourself misspelling things a lot, try using an abbreviation - I think the new abbreviations engine in 3.6.0 makes it possible to use them in cases like these with spaces and even more advanced options.

e.g. I have `maek` as an abbreviation for `make` in my fish configuration, because I mistype that a lot.

mqudsi (changes requested):
I think I would prefer that. I don't *think* I wrote the original code that uses `npm completions`, and I might suggest dumping `npm completions > npm_completions.fish` and pruning out the useless ones from that file before pasting them into our `npm.fish` instead of sourcing them from `npm` at runtime, especially if it provides inferior completions without descriptions.

If you find yourself misspelling things a lot, try using an abbreviation - I think the new abbreviations engine in 3.6.0 makes it possible to use them in cases like these with spaces and even more advanced options.

e.g. I have `maek` as an abbreviation for `make` in my fish configuration, because I mistype that a lot.

m4rch3n1ng:
i am currently in the process of getting rid of the `npm completions` builtin, partly because it doesn't even really work and partly because it's just excruciatingly slow (see https://github.com/fish-shell/fish-shell/pull/9800#discussion_r1196996393) and so i can implement only suggestions for common misspellings, that seems more intelligent. can you elaborate a bit on the "abbreviations" that you mentioned?

faho:
>can you elaborate a bit on the "abbreviations" that you mentioned?

https://fishshell.com/docs/current/interactive.html#abbreviations

m4rch3n1ng:
this should be mostly done now, i fully removed and replaced the `npm completion` command, and now it's significantly faster and more feature complete.

now i just need an explicit "ok" or "not ok" on whether to remove the package completions with `npm install <tab>`.  
i would also consider waiting for npm/cli#6483 npm/cli#6481 and npm/cli#6480 to merge, since i added those flags to the completion, but the npm team is fast and it should be fine regardless i think

mqudsi:
I am strongly against removing `npm install <tab>` suggestions. We can improve the existing code to handle namespaced packages but this is a *very* useful feature that makes no sense to throw out. If you trigger it accidentally you can always press `^C` to cancel.

m4rch3n1ng:
yep ok, then we can keep it and this pr should be ready to merge

faho:
Merged, thanks!

PR #9807, opened 2023-05-23
feat: sync the dracula official theme

whck6:
Signed-off-by: Wenhao Ho <wh.ho@outlook.com>

## Description

Sync dracula official theme: https://github.com/dracula/fish/blob/master/themes/Dracula%20Official.theme


zanchey:
Looks good. I'll add something to the licensing document to indicate that this is derived.

PR #9816, opened 2023-05-28
Simplify scoped_push and ScopedGuard

ridiculousfish:
This makes some simplifications to scoped_push and ScopeGuard:

1. ScopeGuard no longer uses ManuallyDrop; the memory management is now trivial and no longer requires `unsafe`.

2. The functions `cancel`, `rollback`, and `commit` have been removed, as these were unused. They can be added back later if needed.

3. `scoped_push` has been simplified in both signature and implementation.

4. `Projection` is no longer required and has been removed.

Also add a test.

### DIFF
--- a/fish-rust/src/common.rs
+++ b/fish-rust/src/common.rs
@@ -24,7 +24,7 @@ use num_traits::ToPrimitive;
 use once_cell::sync::Lazy;
 use std::env;
 use std::ffi::{CStr, CString, OsString};
-use std::mem::{self, ManuallyDrop};
+use std::mem;
 use std::ops::{Deref, DerefMut};
 use std::os::fd::{AsRawFd, RawFd};
 use std::os::unix::prelude::OsStringExt;
@@ -1717,16 +1717,6 @@ fn get_executable_path(argv0: &str) -> PathBuf {
     std::env::current_exe().unwrap_or_else(|_| PathBuf::from_str(argv0).unwrap())
 }
 
-/// Like [`std::mem::replace()`] but provides a reference to the old value in a callback to obtain
-/// the replacement value. Useful to avoid errors about multiple references (`&mut T` for `old` then
-/// `&T` again in the `new` expression).
-pub fn replace_with<T, F: FnOnce(&T) -> T>(old: &mut T, with: F) -> T {
-    let new = with(old);
-    std::mem::replace(old, new)
-}
-
-pub type Cleanup<T, F> = ScopeGuard<T, F>;
-
 /// A RAII cleanup object. Unlike in C++ where there is no borrow checker, we can't just provide a
 /// callback that modifies live objects willy-nilly because then there would be two &mut references
 /// to the same object - the original variables we keep around to use and their captured references
@@ -1752,47 +1742,19 @@ pub type Cleanup<T, F> = ScopeGuard<T, F>;
 ///
 /// // hello will be written first, then goodbye.
 /// ```
-pub struct ScopeGuard<T, F: FnOnce(&mut T)> {
-    captured: ManuallyDrop<T>,
-    on_drop: Option<F>,
-}
+pub struct ScopeGuard<T, F: FnOnce(&mut T)>(Option<(T, F)>);
 
-impl<T, F> ScopeGuard<T, F>
-where
-    F: FnOnce(&mut T),
-{
+impl<T, F: FnOnce(&mut T)> ScopeGuard<T, F> {
     /// Creates a new `ScopeGuard` wrapping `value`. The `on_drop` callback is executed when the
     /// ScopeGuard's lifetime expires or when it is manually dropped.
     pub fn new(value: T, on_drop: F) -> Self {
-        Self {
-            captured: ManuallyDrop::new(value),
-            on_drop: Some(on_drop),
-        }
+        Self(Some((value, on_drop)))
     }
 
-    /// Cancel the unwind operation, e.g. do not call the previously passed-in `on_drop` callback
-    /// when the current scope expires.
-    pub fn cancel(guard: &mut Self) {
-        guard.on_drop.take();
-    }
-
-    /// Cancels the unwind operation like [`ScopeGuard::cancel()`] but also returns the captured
-    /// value (consuming the `ScopeGuard` in the process).
-    pub fn rollback(mut guard: Self) -> T {
-        guard.on_drop.take();
-        // Safety: we're about to forget the guard altogether
-        let value = unsafe { ManuallyDrop::take(&mut guard.captured) };
-        std::mem::forget(guard);
-        value
-    }
-
-    /// Commits the unwind operation (i.e. applies the provided callback) and returns the captured
-    /// value (consuming the `ScopeGuard` in the process).
+    /// Invokes the callback and returns the wrapped value, consuming the ScopeGuard.
     pub fn commit(mut guard: Self) -> T {
-        (guard.on_drop.take().expect("ScopeGuard already canceled!"))(&mut guard.captured);
-        // Safety: we're about to forget the guard altogether
-        let value = unsafe { ManuallyDrop::take(&mut guard.captured) };
-        std::mem::forget(guard);
+        let (mut value, on_drop) = guard.0.take().expect("Should always have Some value");
+        on_drop(&mut value);
         value
     }
 }
@@ -1801,23 +1763,35 @@ impl<T, F: FnOnce(&mut T)> Deref for ScopeGuard<T, F> {
     type Target = T;
 
     fn deref(&self) -> &Self::Target {
-        &self.captured
+        &self.0.as_ref().unwrap().0
     }
 }
 
 impl<T, F: FnOnce(&mut T)> DerefMut for ScopeGuard<T, F> {
     fn deref_mut(&mut self) -> &mut Self::Target {
-        &mut self.captured
+        &mut self.0.as_mut().unwrap().0
     }
 }
 
 impl<T, F: FnOnce(&mut T)> Drop for ScopeGuard<T, F> {
     fn drop(&mut self) {
-        if let Some(on_drop) = self.on_drop.take() {
-            on_drop(&mut self.captured);
+        if let Some((mut value, on_drop)) = self.0.take() {
+            on_drop(&mut value);
         }
-        // Safety: we're in the Drop so `self` will never be accessed again.
-        unsafe { ManuallyDrop::drop(&mut self.captured) };
+    }
+}
+
+/// A trait expressing what ScopeGuard can do. This is necessary because scoped_push returns an
+/// `impl Trait` object and therefore methods on ScopeGuard which take a self parameter cannot be
+/// used.
+pub trait ScopeGuarding: DerefMut {
+    /// Invokes the callback and returns the wrapped value, consuming the ScopeGuard.
+    fn commit(guard: Self) -> Self::Target;
+}
+
+impl<T, F: FnOnce(&mut T)> ScopeGuarding for ScopeGuard<T, F> {
+    fn commit(guard: Self) -> T {
+        ScopeGuard::commit(guard)
     }
 }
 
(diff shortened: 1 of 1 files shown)


ridiculousfish:
@mqudsi as promised here's the best I can do to simplify scoped_push and ScopeGuard.

Despite my best efforts I wasn't able to make the rust-analyzer type more compact, but this is still a nice simplification.

fish-rust/src/common.rs
```diff
@@ -1827,22 +1779,15 @@ pub fn scoped_push<Context, Accessor, T>(
...
-    // we are changing, and a saved snapshot of the previous values of those variables in a tuple,
-    // then use ScopeGuard's `on_drop` parameter to restore the saved values when the scope ends.
-    let scope_guard = ScopeGuard::new((ctx, accessor, saved_value), |data| {
-        let (ref mut ctx, accessor, saved_value) = data;
-        *accessor(ctx) = *saved_value;
-    });
-    // `scope_guard` would deref to the tuple we gave it, so use Projection<T> to map from the tuple
-    // `(ctx, accessor, saved_value)` to the result of `accessor(ctx)`.
-    Projection::new(scope_guard, |sg| &sg.0, |sg| &mut sg.0)
+    let saved = mem::replace(accessor(&mut ctx), new_value);
```
krobelus, line 1786: just a heads up, this accessor won't work for cases where we need to acquire a lock guard or similar.
For example given a `RefCell` field `library_data`

```
    pub fn libdata_mut(&self) -> RefMut<'_, LibraryData> {
        self.library_data.borrow_mut()
    }
```

there is no kosher function that returns a mutable reference without the `RefMut`.
I worked around this by writing

```
    let parser = scoped_push_replacer(
        parser,
        |parser, new_value| std::mem::replace(&mut parser.libdata_mut().pods.is_event, new_value),
        is_event + 1,
    );
```

but I realize we should instead make it so `accessor` returns something like `impl Deref<&mut T>`

fish-rust/src/common.rs
```diff
@@ -1752,72 +1742,34 @@ pub type Cleanup<T, F> = ScopeGuard<T, F>;
...
-        guard.on_drop.take();
-        // Safety: we're about to forget the guard altogether
-        let value = unsafe { ManuallyDrop::take(&mut guard.captured) };
-        std::mem::forget(guard);
-        value
-    }
-
-    /// Commits the unwind operation (i.e. applies the provided callback) and returns the captured
-    /// value (consuming the `ScopeGuard` in the process).
-    pub fn commit(mut guard: Self) -> T {
```
krobelus, line 1791: I'm already using `commit()` because it's the equivalent of the C++ `scoped_push::restore`. We should unify the names

krobelus (commented):
I'm not sure what to think of the inlay type hints. If they are helpful why aren't they added to the source code? I'll try them for a while.

fish-rust/src/common.rs
```diff
@@ -1827,22 +1779,15 @@ pub fn scoped_push<Context, Accessor, T>(
     mut ctx: Context,
     accessor: Accessor,
     new_value: T,
-) -> impl Deref<Target = Context> + DerefMut<Target = Context>
+) -> impl DerefMut<Target = Context>
```
mqudsi, line 1782: I'm nitpicking but I just want to point out that the change to the inline hints that this PR brings about (which I understand was the motivation for this) could have been brought with just this single line change.

fish-rust/src/common.rs
```diff
@@ -1717,16 +1717,6 @@ fn get_executable_path(argv0: &str) -> PathBuf {
     std::env::current_exe().unwrap_or_else(|_| PathBuf::from_str(argv0).unwrap())
 }
 
-/// Like [`std::mem::replace()`] but provides a reference to the old value in a callback to obtain
-/// the replacement value. Useful to avoid errors about multiple references (`&mut T` for `old` then
-/// `&T` again in the `new` expression).
-pub fn replace_with<T, F: FnOnce(&T) -> T>(old: &mut T, with: F) -> T {
-    let new = with(old);
-    std::mem::replace(old, new)
-}
```
mqudsi, line 1726: Yes, this was no longer used once @krobelus introduced `scoped_push()`; it was an ergonomic workaround used when there was only `ScopeGuard` and should have been dropped before. I mentioned it elsewhere but the `allow(unused)` feature flag prevented the compiler from reporting this.

mqudsi (commented):
I started off with a `ScopeGuard` fairly identical to what you have but changed to the `ManuallyDrop` model to avoid unwrapping on each deref since it's natural that the result will be deref'd repeatedly and all over the place. I felt that the invariant where `captured` is always `Some(...)` was more semantically correct than using an `Option<T>` there. The `ManuallyDrop` necessity was a little unfortunate but I personally felt that aside from unfortunately needing an `unsafe` it was ultimately simpler to prove/grok since we just prevent `Drop` from being called rather than calling `.unwrap()` everywhere - the unsafe aside, it's obvious what the code is doing whereas `.unwrap()` just - to me! - reeks of a workaround.

As @krobelus mentioned, `rollback()` is being used to restore the state in the other branch. I am also reluctant to "unwrite" code at this point - I had need of `Projection<T>` for other things (the mutex workaround) during development and I can see myself needing it in short order again. 

Rust has an exceptional story for tree shaking and statically reporting unused code bits at compile time which we have intentionally switched off with `#![allow(unused)]` in `lib.rs` but we'll be removing that when the port nears completion and it'll show us everything that's (still) not used. I'm not arguing for including a ton of crap for no reason, but I think if one has a hunch something will be used based off their familiarity with the codebase and what they've run into during the porting process it should be OK to leave unused code in *for now* and see how that unfolds. Again thanks to the compiler's support it is really easy to remove that later, so aside from add a negligible amount to the compilation time there's really no harm in keeping things until we need them.

More importantly, I think writing the facade for certain features is easier and more correct at the time when you first develop a component - how everything comes together is perfectly clear and fresh in your mind and you can easily write an additional (unused) helper fn like rollback or restore correctly the first time and with almost zero additional effort, but going in to add it later can require architectural changes, time to either familiarize w/ the original code if you're not the one that wrote it or refamiliarize yourself with it if you wrote it more than a week ago, and a good chance of getting something wrong because you might write it with a less complete mental model of how things work than what the naturally more complete and cohesive mental model you had at the time you were first writing it.

e.g. I wrote the function that @krobelus is using without looking at the cpp codebase because I had a feeling it would either be required or just that its presence would lead one to discover a better way of going about accomplishing a simple task, and it turned out to be the case.

For some context as to why `Projction<T>` is useful and to clarify the mutex workaround comment, let's imagine you have a medium-grained lock protecting a structure (in my case, it was  `Mutex<Option<Term>>` and you want to return a borrow against the mutex. You would be required to return `MutexGuard<Option<Term>>` from your abstraction, but with Projection<T> you can effectively return `MutexGuard<Term>` or `MutexGuard<MutableTermField>` so that your api is a) self-documenting, and b) more ergonomic. e.g. it would allow you to write `curses::get_term_option()` returning `MutexGuard<SomeFoo>` and allow callers to use `curses::get_term_option()` directly instead of having every caller need to call `curses::get_term()` then check if the `Option` was safe to unwrap (when the impl *knows* that it was safe to unwrap when it returned the value) then deref to the `Term` field you needed to expose (requiring it to be `public` in the process).

fish-rust/src/common.rs
```diff
@@ -24,7 +24,7 @@ use num_traits::ToPrimitive;
 use once_cell::sync::Lazy;
```
krobelus, line 24: the `&mut &mut &mut` we may be able to get rid of by replacing (untested)

```
let zelf = scoped_push(zelf, ...);
```

with

```
let zelf = scoped_push(&*zelf, ...);
```

mqudsi: I think there has to be a better way, along the lines of `std::remove_reference()` so that you don't have to do this manually at the call-site. Perhaps if we took `Borrow<Context>` instead of `Context` directly?

fish-rust/src/common.rs
```diff
@@ -1752,72 +1742,34 @@ pub type Cleanup<T, F> = ScopeGuard<T, F>;
...
-
-    /// Cancel the unwind operation, e.g. do not call the previously passed-in `on_drop` callback
-    /// when the current scope expires.
-    pub fn cancel(guard: &mut Self) {
-        guard.on_drop.take();
-    }
-
-    /// Cancels the unwind operation like [`ScopeGuard::cancel()`] but also returns the captured
-    /// value (consuming the `ScopeGuard` in the process).
-    pub fn rollback(mut guard: Self) -> T {
```
krobelus, line 1781: Elsewhere I've used the C++ stdlib's equivalent of `cancel()` like this:

```rust
let mut populate_error = ScopeGuard::new(populate_error_fn);
if error-scenario-1 {
   return Err(());
}
if error-scenario-2 {
   return Err(());
}
populate_error.cancel();
Ok(())
```

I don't expect this to be a common pattern but it's probably useful in some situations.

I wonder if cancel() and rollback() actually need `ManuallyDrop`?

mqudsi: That's actually a very interesting approach, though I'll be honest and say it would take me a second to understand what was being done. It's funnily enough a very timely example of what I said above ("because I had a feeling it would either be required or just that its presence would lead one to discover a better/easier way of going about accomplishing a simple task")

fish-rust/src/common.rs
krobelus: Yeah this could have been a separate commit

krobelus:
This looks good to me except we should keep `commit()`.

> For some context as to why Projection<T> is useful and to clarify the mutex workaround comment, let's imagine you have a medium-grained lock protecting a structure (in my case, it was Mutex<Option<Term>> and you want to return a borrow against the mutex. You would be required to return MutexGuard<Option<Term>> from your abstraction, but with Projection you can effectively return MutexGuard<Term> or MutexGuard<MutableTermField> so that your api is a) self-documenting, and b) more ergonomic. e.g. it would allow you to write curses::get_term_option() returning MutexGuard<SomeFoo> and allow callers to use curses::get_term_option() directly instead of having every caller need to call curses::get_term() then check if the Option was safe to unwrap (when the impl knows that it was safe to unwrap when it returned the value) then deref to the Term field you needed to expose (requiring it to be public in the process).

yes that sounds like a good use of Projection but we don't need it for the scope guard. Except maybe if the value we modify with scoped_push is behind a mutex guard or similar as described in https://github.com/fish-shell/fish-shell/pull/9816#discussion_r1208471386, I'll need to check.
I wouldn't mind removing it and adding it back when we have a definite use.

fish-rust/src/common.rs
krobelus: Yeah I wouldn't have discovered this if it wasn't in the stdlib but I'm not sure if it's actually useful in Rust.

ridiculousfish:
> I started off with a ScopeGuard fairly identical to what you have but changed to the ManuallyDrop model to avoid unwrapping on each deref since it's natural that the result will be deref'd repeatedly and all over the place. I felt that the invariant where captured is always Some(...) was more semantically correct than using an Option<T> there. The ManuallyDrop necessity was a little unfortunate but I personally felt that aside from unfortunately needing an unsafe it was ultimately simpler to prove/grok since we just prevent Drop from being called rather than calling .unwrap() everywhere - the unsafe aside, it's obvious what the code is doing whereas .unwrap() just - to me! - reeks of a workaround.

These unwraps are trivially optimized out; plus there was already a usage of an Option.

More importantly this `ManuallyDrop` and `mem::forget` pattern tied us to a manual memory management scheme. If we were to add more fields to `ScopeGuard` (which the riir-parser branch does), we would have to remember to manually free them in e.g. `rollback` or else have a leak - [indeed that already happened](https://github.com/fish-shell/fish-shell/commit/b39715434b4ff876fea088fa0e96213e1d68142a).

That's a strong signal that this design was more complex than necessary. We should just use regular automatic memory management, not this manual stuff. KISS. 

> As @krobelus mentioned, rollback() is being used to restore the state in the other branch. I am also reluctant to "unwrite" code at this point - I had need of Projection<T> for other things (the mutex workaround) during development and I can see myself needing it in short order again.

I've restored `commit()` and added some tests. I don't see `rollback()` being used; if I'm wrong it's fine to add it.

> More importantly, I think writing the facade for certain features is easier and more correct at the time when you first develop a component

But (not trying to be mean) it wasn't correct the first time. To put it bluntly, it was code that was unused, untested, and buggy. We should be deleting code like that, not writing more of it.

If a component is so conceptually complex that we feel pressure to add extra features now, before we forget how it works: that's a good moment to take a step back and try hard to simplify it.

If we allow ourselves to write stuff just because it might be useful later, we will make components more complex than necessary. YAGNI.

> For some context as to why Projection<T> is useful and to clarify the mutex workaround comment, let's imagine you have a medium-grained lock protecting a structure (in my case, it was  Mutex<Option<Term>> and you want to return a borrow against the mutex.

Maybe. I see the argument though I am not convinced.

But if the problem is mapping MutexGuard, then we should design something specifically for MutexGuard. We should not immediately reach for the most general, maximally abstract solution. Abstractions should emerge from factoring concrete components together.

PR #9818, opened 2023-05-29
Rewrite and adopt print_help in Rust

zanchey:
I'm still pretty new to Rust, but here's an attempt to port the tiny print_help
function. It currently uses C `char *` so I've switched to OsString; let me know
if a wide string is better.


fish-rust/src/print_help.rs
```diff
@@ -0,0 +1,33 @@
...
+mod ffi2 {
+    extern "Rust" {
+        unsafe fn unsafe_print_help(command: *const c_char);
+    }
+}
+
+fn print_help(command: &OsStr) {
+    let mut cmdline = OsString::new();
+    cmdline.push("__fish_print_help ");
+    cmdline.push(command);
```
mqudsi, line 21: This is all 100% correct but for ergonomics post-port (and since the existing C code already used ascii c strings) you can change `print_help()` to either take an `&str` directly or a generic `AsRef<OsStr>` (which is what the rust `Command` functions all take). If you take `&str` then you can use `CStr`'s [`to_str()`](https://doc.rust-lang.org/std/ffi/struct.CStr.html#method.to_str) (and unwrap since we control the calling code so we should guarantee it is only called with ascii) from the unsafe ffi helper func to get an `&str` out of the passed-in `command_buf`, or if you use `AsRef<OsStr>` then you can use `CStr::to_bytes()` like you already do and [`OsStrExt::from_bytes()`](https://doc.rust-lang.org/std/os/unix/ffi/trait.OsStrExt.html#tymethod.from_bytes) to get the `OsStr` from that.

zanchey: Thanks, I think `&str` is a better option, but I'm not sure when I'll get a chance to do that so I'll merge this for now and take a look it at again in a week or two.

PR #9821, opened 2023-05-30
Port killring

clemenswasser:
## Description

This PR ports the `kill.cpp` ("killring") to Rust by reimplementing all functions.
The implementation of `kill_entries` is not optimal in a multithreaded context, since we can't easily access a `std::vector<wcstring>` from Rust, so I had to work around that limitation. Additionally `kill_replace` now removes **all** "old" entries, since using retain is more idiomatic and IMHO more correct anyway.

### DIFF
--- a/fish-rust/src/kill.rs
+++ b/fish-rust/src/kill.rs
@@ -1,0 +1,99 @@
+//! The killring.
+//!
+//! Works like the killring in emacs and readline. The killring is cut and paste with a memory of
+//! previous cuts.
+
+use cxx::CxxWString;
+use std::collections::VecDeque;
+use std::pin::Pin;
+use std::sync::Mutex;
+
+use crate::ffi::wcstring_list_ffi_t;
+use crate::wchar::WString;
+use crate::wchar_ffi::WCharFromFFI;
+
+#[cxx::bridge]
+mod kill_ffi {
+    extern "C++" {
+        include!("wutil.h");
+        type wcstring_list_ffi_t = super::wcstring_list_ffi_t;
+    }
+
+    extern "Rust" {
+        #[cxx_name = "kill_add"]
+        fn kill_add_ffi(new_entry: &CxxWString);
+        #[cxx_name = "kill_replace"]
+        fn kill_replace_ffi(old_entry: &CxxWString, new_entry: &CxxWString);
+        #[cxx_name = "kill_yank_rotate"]
+        fn kill_yank_rotate_ffi(mut out_front: Pin<&mut CxxWString>);
+        #[cxx_name = "kill_yank"]
+        fn kill_yank_ffi(mut out_front: Pin<&mut CxxWString>);
+        #[cxx_name = "kill_entries"]
+        fn kill_entries_ffi(mut out: Pin<&mut wcstring_list_ffi_t>);
+    }
+}
+
+static KILL_LIST: once_cell::sync::Lazy<Mutex<VecDeque<WString>>> =
+    once_cell::sync::Lazy::new(|| Mutex::new(VecDeque::new()));
+
+fn kill_add_ffi(new_entry: &CxxWString) {
+    kill_add(new_entry.from_ffi());
+}
+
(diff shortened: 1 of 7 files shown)


fish-rust/src/kill.rs
```diff
@@ -0,0 +1,101 @@
...
+}
+
+fn kill_replace_ffi(old_entry: &CxxWString, new_entry: &CxxWString) {
+    kill_replace(old_entry.from_ffi(), new_entry.from_ffi())
+}
+
+/// Replace the specified string in the killring.
+pub fn kill_replace(old_entry: WString, new_entry: WString) {
+    let mut kill_list = KILL_LIST.lock().unwrap();
+    kill_list.retain(|str| str != &old_entry);
```
ridiculousfish, line 52: For better or for worse, the C++ code here only removed the first matching string, not all matching strings.

fish-rust/src/kill.rs
```diff
@@ -0,0 +1,101 @@
...
+    kill_list.push_front(new_entry);
```
ridiculousfish, line 53: There was also a check that `new_entry` is not empty.

fish-rust/src/kill.rs
```diff
@@ -0,0 +1,101 @@
...
+}
+
+/// Rotate the killring.
+pub fn kill_yank_rotate() -> WString {
+    let mut kill_list = KILL_LIST.lock().unwrap();
+    kill_list.rotate_left(1);
+    kill_list.front().cloned().unwrap_or_default()
+}
+
+fn kill_yank_ffi(mut out_front: Pin<&mut CxxWString>) {
```
ridiculousfish, line 70: This is fine but FYI it is also possible to return a `UniquePtr<CxxWString>`, example:

```
fn kill_yank_rotate_ffi() -> cxx::UniquePtr<CxxWString> {
    kill_yank_rotate().to_ffi()
}
```

fish-rust/src/kill.rs
```diff
@@ -0,0 +1,101 @@
...
+pub fn kill_yank() -> WString {
+    let kill_list = KILL_LIST.lock().unwrap();
+    kill_list.front().cloned().unwrap_or_default()
+}
+
+pub fn kill_entries_len() -> usize {
+    KILL_LIST.lock().unwrap().len()
+}
+
+fn kill_entries_ffi(mut out_entries: Pin<&mut CxxVector<CxxWString>>) -> bool {
```
ridiculousfish, line 85: This is also fine as written but see `env_ffi.rs as_list_ffi()` for a more direct way to do this.

ridiculousfish (changes requested):
Thank you, it looks really good! A few nits in `kill_replace` and some other suggestions which you may take or not at your option.

fish-rust/src/kill.rs
clemenswasser: I'm aware of the change. It seems I overlooked mentioning it in my PR description.
Personally, I find this behavior change to be more logical and can be expressed very idiomatic in Rust via `retain`.

fish-rust/src/kill.rs
clemenswasser: Thanks! Fixed in the latest commit.

fish-rust/src/kill.rs
clemenswasser: Yes, I know. But I wanted to avoid any unnecessary heap allocations.

fish-rust/src/kill.rs
clemenswasser: Thanks for mentioning. I've changed it to use `wcstring_list_ffi_t`.

fish-rust/src/kill.rs
faho: Be that as it may, the initial port is supposed to keep fidelity with the C++ code as it is.

Please change this.

clemenswasser: Okay, done.

ridiculousfish:
Thank you!!

PR #9828, opened 2023-06-02
improved print CSS for fish_config

korve:
## Description

This commit introduces a fishconfig_print.css that contains special CSS styles that only apply when printing the fishconfig page. This is especially useful when the user wants to print out the key bindings. 

Also the filter input is now hidden when printing any of the tabs on the config page.

Before:

before

After:

after

## TODOs:

### DIFF
--- a/share/tools/web_config/fishconfig.css
+++ b/share/tools/web_config/fishconfig.css
@@ -664,3 +664,7 @@ img.delete_icon {
         color: #AAA;
     }
 }
+
+.print_only {
+    display: none;
+}
--- a/share/tools/web_config/fishconfig_print.css
+++ b/share/tools/web_config/fishconfig_print.css
@@ -1,0 +1,30 @@
+body {
+    background: white;
+    font-size: 8pt;
+}
+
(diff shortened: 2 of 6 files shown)


Xiretza:
The result looks great! Something that would probably be good for the regular stylesheet as well: could the right column's width be reduced to fit as much of the left column as possible, rather than always splitting the page in the middle? The bindings are almost always longer than the key sequences.

korve:
> The result looks great! Something that would probably be good for the regular stylesheet as well: could the right column's width be reduced to fit as much of the left column as possible, rather than always splitting the page in the middle? The bindings are almost always longer than the key sequences.

Its kinda difficult to find a width for every screen/orientation so the right column doesn't get too small. Also, you can expand rows when clicking on them, so more content is revealed on the right column.

Some keys are really weird, for example `commandline\ -f\ expand-abbr\;\ commandline\ -i\ \\n`. What does it even do? Maybe more descriptive keys would be a better solution.

faho:
Merged, thanks!

Issue #9812, opened 2023-05-26
Mass files completion, list not refresh when selecting last one


NextAlone:
**YOUR TEXT HERE**
Version: fish, version 3.6.1-461-g85d8f2b27

Tried with no-config one.

When completing files in a directory that contains many folders, use the up and down arrows to select the desired folder. However, you may notice that the last folder in the list may not be displayed. In such cases, the name of the last folder can be completed directly at the command line.

![image](https://github.com/fish-shell/fish-shell/assets/12210746/2f9f9ac2-2e09-4509-9d48-c77265d44eba)

Where 2023-05-26-17-33-23/ is the 66th one.

ridiculousfish:
In your screenshot, I would not expect to see the last directory, because it is showing rows 1-26 of 66.

I was not able to reproduce this. Here's what I tried:

1. Make a bunch of directories: `for x in (seq 504); mkdir (printf "long_dir_%03d" $x); end`
2. Hit tab twice to open the pager
3. Use the arrow keys to scroll to the bottom

All the directories are shown.

<img width="1146" alt="Screenshot 2023-05-27 at 12 43 54 PM" src="https://github.com/fish-shell/fish-shell/assets/920838/bd829e60-3c81-4d0f-9246-5dfe04774386">

simonborje:
It seems to work if one scrolls down from the top. but if the first arrow press is "Up" so that the selection wraps around to the bottom item/row, the list is not redrawn to show the selected item. Pressing "Up" a second time does make the list redraw however.

NextAlone:
> It seems to work if one scrolls down from the top. but if the first arrow press is "Up" so that the selection wraps around to the bottom item/row, the list is not redrawn to show the selected item. Pressing "Up" a second time does make the list redraw however. ![explorer_ByGy8PPtFG](https://user-images.githubusercontent.com/3867784/241434427-8fa8e3cb-deb4-482f-a26b-b4b59bc9b84b.gif)

Yes, that is the key, if you press the up arrow to move to the last item, then the list will not be redrawn, I think this is a bug,

NextAlone:
> 1. for x in (seq 504); mkdir (printf "long_dir_%03d" $x); end


https://github.com/fish-shell/fish-shell/assets/12210746/1f120ea1-fe34-447d-92e1-82c0c9e5f7d6

When seleting 001，press up，to de-select，press up again to select 504，this time, not redrawn list.

simonborje:
[...]

PR #9833, opened 2023-06-04
Redraw pager on new selection when nothing was selected previously

simonborje:
## Description

An attempt to solve the issue described in #9812.

With this change the `select_next_completion_in_direction` function will no longer return early in the case when nothing was previously selected. The early return would skip updating the `suggested_row_start` variable, leading to the pager not scrolling down to show the new selection when using the "Up" arrow to wrap to the last item.

This change seems to solve the issue based on manual testing.

Please let me know if anything under "TODO" should be updated.

Fixes issue #9812

## TODOs:

### DIFF
--- a/src/pager.cpp
+++ b/src/pager.cpp
@@ -664,8 +664,8 @@ bool pager_t::select_next_completion_in_direction(selection_motion_t direction,
         return false;
     }
 
-    // Handle the case of nothing selected yet.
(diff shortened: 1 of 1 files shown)


ridiculousfish:
Looks great - thank you! This is quite the fix especially for a first-time contributor. Amazing!

NextAlone:
Congratulations!

PR #9841, opened 2023-06-10
docs: Add "Writing your own prompt" doc

faho:
## Description

We have reasonable "reference" style documentation, but we kind of lack tutorials, so it's hard for people to even find out how to approach something.

So this attempts to rectify it by giving a simple example on how to write your own prompt.

### DIFF
--- a/doc_src/prompt.rst
+++ b/doc_src/prompt.rst
@@ -1,0 +1,187 @@
+Writing your own prompt
+=======================
+
+Fish ships a number of prompts that you can view with the :doc:`fish_config <cmds/fish_config>` command, and many users have shared their prompts online.
+
+However, if you are anything like the author of this document, you sooner or later want to write your own or at least adjust the prompt you have.
+
+Unlike other shells, fish's prompt is built by running a function - :doc:`fish_prompt <cmds/fish_prompt>`. Or, more specifically, three functions:
+
+- :doc:`fish_prompt <cmds/fish_prompt>`
+- :doc:`fish_right_prompt <cmds/fish_right_prompt>` - which is shown on the right side of the terminal.
+- :doc:`fish_mode_prompt <cmds/fish_mode_prompt>` - which it runs if :ref:`vi-mode <vi-mode>` is used.
+
+These functions are run, and whatever they print is displayed as the prompt (minus one trailing newline).
+
+.. note::
+   This document uses formatting to show what a prompt would look like. If you are viewing this in the man page,
+   you probably want to switch to looking at the web version instead. Run ``help custom-prompt`` to view it.
+
+Our first prompt
+----------------
+
+Let's look at a very simple example::
+
+  function fish_prompt
+      echo $PWD '>'
+  end
+
+This just prints the current working directory (:envvar:`PWD`) and a ``>`` symbol to show where the prompt ends.
+
+Because we've used :doc:`echo <cmds/echo>`, it added spaces between the two so it ends up looking like (assuming ``_`` is your cursor):
+
+.. role:: white
+.. parsed-literal::
+    :class: highlight
+
+    :white:`/home/tutorial >`\ _
+
+Formatting
+----------
+
+If you don't want the space, there are other ways to print this:
+
+- ``echo -s``, which tells echo to not print a space.
+- We could construct one string: ``echo "$PWD>"``
+- :doc:`printf <cmds/printf>`, which allows us to give a format string to format our... string
+- :doc:`string join <cmds/string-join>`, if we want to have one thing appear between all our components.
+
+For now, let's use ``string join ''`` to write all our components without anything inbetween::
+
+  function fish_prompt
+      string join '' -- $PWD '>'
+  end
+
+The ``--`` tells ``string`` that no options can come after it, in case we extend this with something that can start with a ``-``.
+
+Adding color
+------------
+
+This prompt is functional, but a bit boring. We could add some color.
+
+In general, telling a terminal to print text in color involves sending it an escape sequence. For example you could send it
+
+.. role:: red
+.. parsed-literal::
+    :class: highlight
+
+    ESCAPE [ 3 1 m
+
+where "escape" is the ascii escape character stored as 0x1b.
+
+We could do this in fish, by writing escape as ``\e`` or ``\x1b``::
+
+  echo \e\[31mfoo
+
+This will tell the terminal to print "foo" in red and will look like this:
+
+.. parsed-literal::
+    :class: highlight
+
+    :red:`foo`
+
+To make it blue, we would use ``\e\[34mfoo``. However, obviously it's a bit annoying to have to remember the numbers for these colors, and modern terminals can do quite a bit more - most can not only do 256 colors but even 24-bit RGB.
+
+So, fish offers the :doc:`set_color <cmds/set_color>` command instead, so you can do::
+
+  echo (set_color red)foo
+
+``set_color`` can also handle RGB colors like ``set_color 23b455``, and other formatting options including bold and italics.
+
+So, taking our previous prompt and adding some color::
+
+  function fish_prompt
+      string join '' -- (set_color green) $PWD (set_color normal) '>'
+  end
+
+A "normal" color tells the terminal to go back to its normal formatting options.
+
+Shortening $PWD
+---------------
+
+This is alright, but our $PWD can be a bit long, and we are typically only interested in the last few directories. We can shorten this with the :doc:`prompt_pwd <cmds/prompt_pwd>` helper that will give us a shortened working directory::
+
+  function fish_prompt
+      string join '' -- (set_color green) (prompt_pwd) (set_color normal) '>'
+  end
+
+``prompt_pwd`` takes options to control how much to shorten. For instance if we want to display the last two directories, we'd use ``prompt_pwd --full-length-dirs=2`` or ``prompt_pwd -D 2``::
+
+  function fish_prompt
+      string join '' -- (set_color green) (prompt_pwd -D 2) (set_color normal) '>'
+  end
+
+With a current directory of "/home/tutorial/Music/Lena Raine/Oneknowing", this would print
+
+.. role:: green
+.. parsed-literal::
+    :class: highlight
+
+    :green:`~/M/Lena Raine/Oneknowing`>_
+
+Status
+------
+
+One important bit [#]_ of information that every command returns is the :ref:`status <variables-status>`. This is either 0 if the command returned successfully, or a whole number from 1 to 255 if not, so it's best understood as an error code.
+
+It's useful to display this in your prompt, but showing it when it's 0 seems kind of wasteful.
+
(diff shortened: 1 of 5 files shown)


doc_src/prompt.rst
```diff
@@ -0,0 +1,189 @@
...
+Our first prompt
+----------------
+
+Let's look at a very simple example::
+
+  function fish_prompt
+      echo $PWD '>'
+  end
+
+This just prints the current working directory (:envvar:`PWD`) and a ``>`` symbol to show where the prompt ends. The ``>`` is :ref:`quoted <quotes>` because otherwise it would signify a :ref:`redirection <redirects>`.
```
zanchey, line 31: I'd drop the "just" here.

doc_src/prompt.rst
```diff
@@ -0,0 +1,189 @@
...
+
+Because we've use :doc:`echo <cmds/echo>`, it adds spaces between the two so it ends up looking like (assuming ``_`` is your cursor):
```
zanchey, line 33: "use" should be "used"?

doc_src/prompt.rst
```diff
@@ -0,0 +1,189 @@
...
+
+Formatting
+----------
+
+If you don't want the space, there are other ways to print this:
+
+- ``echo -s``, which tells echo to not print a space.
+- We could construct one string: ``echo "$PWD>"``
+- :doc:`printf <cmds/printf>`, which allows us to give a format string to format our... string
+- :doc:`string join <cmds/string-join>`, if we want to have one thing appear between all our components.
```
zanchey, line 49: I'd rearrange this to be "If you don't want the space, you can use string join, like this:

Other methods include..." (or drop them entirely)

doc_src/prompt.rst
```diff
@@ -0,0 +1,189 @@
...
+
+For now, let's use ``string join ''`` to write all our components without anything inbetween::
```
zanchey, line 51: in-between

doc_src/prompt.rst
```diff
@@ -0,0 +1,189 @@
...
+
+  function fish_prompt
+      string join '' -- $PWD '>'
+  end
+
+The ``--`` tells ``string`` that no options can come after it, in case we extend this with something that can start with a ``-``.
```
zanchey, line 57: This is a personal stylistic preference, but I avoid "tell program X to Y" because it's not a person. What about "The -- indicates that no further options are being used, which allows arguments that start with - to work correctly."

doc_src/prompt.rst
```diff
@@ -0,0 +1,189 @@
...
+
+Adding color
```
zanchey, line 59: Please add a u here. I know you will have to take it right out again, but it will help me sleep at night.

doc_src/prompt.rst
```diff
@@ -0,0 +1,189 @@
...
+------------
+
+This prompt is functional, but a bit boring. We could add some color.
+
+In general, telling a terminal to print text in color involves sending it an escape sequence. For example you could send it
```
zanchey, line 64: I'd stay away from escape sequences entirely. I love learning about terminal emulator protocols as much as the next person, but I think the reader we're aiming for just wants to put an emoji in their prompt. You could rephrase as "The easiest way is to use fish's set_color command ... set_color sends escape sequences, which instruct the terminal to change the color of the text. You can learn more about escape sequences at Wikipedia [or some similar reference]".

doc_src/prompt.rst
```diff
@@ -0,0 +1,189 @@
...
+
+So, taking our previous prompt and adding some color::
+
+  function fish_prompt
+      string join '' -- (set_color green) $PWD (set_color normal) '>'
+  end
+
+A "normal" color tells the terminal to go back to its normal formatting options.
+
+Shortening $PWD
```
zanchey, line 101: I'd call this "shortening the working directory"

doc_src/prompt.rst
```diff
@@ -0,0 +1,189 @@
...
+---------------
+
+This is alright, but our $PWD can be a bit long, and we are typically only interested in the last few directories. We can shorten this with the :doc:`prompt_pwd <cmds/prompt_pwd>` helper that will give us a shortened working directory::
```
zanchey, line 104: Alright's status as a word is controversial - what about "fine"? I'd use the envvar link here for PWD.

doc_src/prompt.rst
```diff
@@ -0,0 +1,189 @@
...
+
+  function fish_prompt
+      string join '' -- (set_color green) (prompt_pwd) (set_color normal) '>'
+  end
+
+``prompt_pwd`` takes options to control how much to shorten. For instance if we want to display the last two directories, we'd use ``prompt_pwd --full-length-dirs=2`` or ``prompt_pwd -D 2``::
```
zanchey, line 110: Probably a comma after "for instance"; I'd just use the longer option version here rather than both.

doc_src/prompt.rst
```diff
@@ -0,0 +1,189 @@
...
+.. role:: green
+.. parsed-literal::
+    :class: highlight
+
+    :green:`~/M/Lena Raine/Oneknowing`>_
+
+Status
+------
+
+One important bit [#]_ of information that every command returns is the :ref:`status <variables-status>`. This is either 0 if the command returned successfully, or a whole number from 1 to 255 if not, so it's best understood as an error code.
```
zanchey, line 127: Not sure about the footnote - it will definitely make some people laugh but confuse others! The formal name for the status is "exit status" and we've tried to stick to that pretty consistently. Also non-0 statuses are not always errors! I'd replace the second sentence with something like "This is a number between 0 and 255, and 0 generally indicates that the command completed successfully, with other numbers used to indicate various error conditions".

doc_src/prompt.rst
```diff
@@ -0,0 +1,189 @@
...
+
+It's useful to display this in your prompt, but showing it when it's 0 seems kind of wasteful.
+
+First of all, since every command (except for :doc:`set <cmds/set>`) changes the status, you need to store it for later use as the first thing in your prompt. Use a local variable so it will be confined to your prompt function::
```
zanchey, line 131: Do you want a link here to the variable scopes reference?

doc_src/prompt.rst
```diff
@@ -0,0 +1,189 @@
...
+  set -l stat
+  if test $last_status -ne 0
+      set stat (set_color red)"[$last_status]"(set_color normal)
+  end
+
+And to print it, we add it to our ``string join``::
+
+  string join '' -- (set_color green) (prompt_pwd) (set_color normal) $stat '>'
+  
+In case that $last_status was 0, $stat is empty, and so it will simply disappear.
```
zanchey, line 147: For formatting I'd put these variable names in code blocks.

doc_src/prompt.rst
```diff
@@ -0,0 +1,189 @@
...
+
+We have now built a simple but working and usable prompt, but of course more can be done.
+
+- Fish offers more helper functions:
+  - ``prompt_login`` to describe the user/hostname/container or ``prompt_hostname`` to describe just the host
+  - ``fish_is_root_user`` to help with changing the symbol for root.
+  - ``fish_vcs_prompt`` to show version control information (or ``fish_git_prompt`` / ``fish_hg_prompt`` / ``fish_svn_prompt`` to limit it to specific systems)
+- You can add a right prompt by changing :doc:`fish_right_prompt <cmds/fish_right_prompt>` or a vi-mode prompt by changing :doc:`fish_mode_prompt <cmds/fish_mode_prompt>`.
+- Some prompts have interesting or advanced features
+  - Add a date?
```
zanchey, line 185: Nit - list items should either all finish with punctuation, or all without

doc_src/prompt.rst
```diff
@@ -0,0 +1,189 @@
...
+  - Show various integrations like python's venv
+  - Color the parts differently.
+
+You can look at fish's sample prompts for inspiration.
```
zanchey, line 189: Is there any way of linking to these? What about mentioning `fish_config` here?

zanchey (commented):
I think this is a great idea. I've got a mixture of feedback in this review, much of which I admit is my own style rather than the one true way of writing documentation.

doc_src/prompt.rst
faho: Done, good idea!

doc_src/prompt.rst
faho: I don't want to remove the hint *entirely*, because I both want to have people hear the termin "escape sequences", and I want to try to get them to see how nice set_color is.

Juggling colo(u)r numbers is annoying!

doc_src/prompt.rst
faho: A compromise "(u)" was added.

doc_src/prompt.rst
faho: I really *like* the style because I find it easy to read. Languages are great at anthropomorphizing stuff because we talk about people a lot.

Your example kind of removes the responsible party - indicates to *what*? Which part of the code does this? Oh, it's `string` that sees it and stops?

Which is kind of important information - this works if the command does it, it's not a magic piece of syntax.

So I switched to "indicates". Is that acceptable?

doc_src/prompt.rst
faho: You're right, this is a bit long. I mentioned `echo -s` and `printf`, but just by reference.

doc_src/prompt.rst
```diff
@@ -0,0 +1,189 @@
...
+- :doc:`fish_prompt <cmds/fish_prompt>`
+- :doc:`fish_right_prompt <cmds/fish_right_prompt>` - which is shown on the right side of the terminal.
+- :doc:`fish_mode_prompt <cmds/fish_mode_prompt>` - which it runs if :ref:`vi-mode <vi-mode>` is used.
+
+These functions are run, and whatever they print is displayed as the prompt (minus one trailing newline).
+
+Here, we will just be writing a simple fish_prompt.
+
+.. note::
+   This document uses formatting to show what a prompt would look like. If you are viewing this in the man page,
```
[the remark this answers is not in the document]
faho: Yeah I put it up top and made it only visible in the man page, but I also made it a warning.

doc_src/prompt.rst
```diff
@@ -0,0 +1,189 @@
+Writing your own prompt
+=======================
+
+Fish ships a number of prompts that you can view with the :doc:`fish_config <cmds/fish_config>` command, and many users have shared their prompts online.
```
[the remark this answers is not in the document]
faho: I find it hard enough to not capitalize nouns to begin with, and starting a sentence with lower-case feels extremely off to me.

faho:
Alright I've merged this now so it's easier to work on.

doc_src/prompt.rst
zanchey: Yes, I think that's great.

doc_src/prompt.rst
zanchey: Hahah. I'll take it out, through my tears.

PR #9844, opened 2023-06-13
Add blueutil completions

nakulj:
## Description

Add completions for [blueutil](https://github.com/toy/blueutil).

## TODOs:

### DIFF
--- a/share/completions/blueutil.fish
+++ b/share/completions/blueutil.fish
@@ -1,0 +1,9 @@
+complete -c blueutil -f
+
+set -l commands \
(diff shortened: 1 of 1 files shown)


zanchey:
Thanks for making your first PR for fish!

I think nobody's merged it because this is a bit of an unusual approach - completions would generally be broken out into separate `complete` commands for each of the commands given here, and use the long/short/description options to give more context. Do you think that's something you could tackle?

nakulj:
Makes sense, I’ll try again when I’m back from vacation. Thanks 😊

faho:
Closing for now. If you get back to it we can reopen it!

PR #9845, opened 2023-06-14
Port builtins/argparse to Rust

henrikhorluck:
## Description

- I'm not certain about the FFI-naming

## TODOs:


fish-rust/src/builtins/argparse.rs
```diff
@@ -0,0 +1,1039 @@
...
+    wopt(L!("stop-nonopt"), woption_argument_t::no_argument, 's'),
+    wopt(L!("ignore-unknown"), woption_argument_t::no_argument, 'i'),
+    wopt(L!("name"), woption_argument_t::required_argument, 'n'),
+    wopt(L!("exclusive"), woption_argument_t::required_argument, 'x'),
+    wopt(L!("help"), woption_argument_t::no_argument, 'h'),
+    wopt(L!("min-args"), woption_argument_t::required_argument, 'N'),
+    wopt(L!("max-args"), woption_argument_t::required_argument, 'X'),
+];
+
+fn exec_subshell(
```
henrikhorluck, line 95: I don't think this is supposed to be in this file, but I don't know where else to put it

src/parser.h
```diff
@@ -434,6 +434,8 @@ class parser_t : public std::enable_shared_from_this<parser_t> {
     /// Remove the outermost block, asserting it's the given one.
     void pop_block(const block_t *expected);
 
+    /// Avoid maybe_t usage for ffi, sends a empty string in case of none.
+    wcstring get_function_name_ffi(int level);
```
henrikhorluck, line 438: Is there a better way to handle `maybe_t` => Rust?

fish-rust/src/builtins/argparse.rs
```diff
@@ -0,0 +1,1033 @@
...
+
+    if retval != STATUS_CMD_OK {
+        return retval;
+    }
+
+    // It's a valid integer so store it and return success.
+    opt_spec.vals.clear();
+    opt_spec.vals.push(val.into());
+    opt_spec.num_seen += 1;
+    w.nextchar = L!("");
```
henrikhorluck, line 739: This used to set `nextchar` to `nullptr`, and `nextchar` was private before this change, is there a method on `wgetopter_t` that should be used instead?

ridiculousfish:
This is outstanding - thank you! I plan to merge tomorrow.

ridiculousfish:
This is fantastic! Thank you for tackling this - I recognize this isn't the most pleasant code in the codebase to port. Merged as 292f7b2be11cd0ab17cc9830c0e19233e46bc95c

For future reference, prefer the wstr functions like `slice_from` and `char_at` over using char slices. `char_at` even implements a "nul-termination" like behavior which can be useful when porting from C++. I switched to those in 6936c944c157edbcff1baf0902af88a742ad43ae

fish-rust/src/builtins/argparse.rs
ridiculousfish: No what you did (making it public) is fine.

src/parser.h
ridiculousfish: The usual way is via unique_ptr. So this could return `unique_ptr<wcstring>` which would be NULL if the value were None.

fish-rust/src/builtins/argparse.rs
ridiculousfish: It's fine here for now.

PR #9856, opened 2023-06-21
path: drop path_get_paths_ffi

zanchey:
f77dc24 provides the pieces to call `path_get_paths` directly from Rust
code. Drop the C++ implementation and its FFI.

I'm not totally convinced the dereferencing of the `EnvStackRef` returned by
`get_vars()` here is right.

### DIFF
--- a/fish-rust/src/builtins/command.rs
+++ b/fish-rust/src/builtins/command.rs
@@ -5,8 +5,8 @@ use crate::builtins::shared::{
     STATUS_CMD_OK, STATUS_CMD_UNKNOWN, STATUS_INVALID_ARGS,
 };
 use crate::ffi::parser_t;
-use crate::path::path_get_paths;
-use crate::wchar::{wstr, WString, L};
+use crate::path::{path_get_path, path_get_paths};
+use crate::wchar::{wstr, L};
 use crate::wgetopt::{wgetopter_t, wopt, woption, woption_argument_t};
 use crate::wutil::sprintf;
 
@@ -71,10 +71,14 @@ pub fn r#command(
     let mut res = false;
     let optind = w.woptind;
     for arg in argv.iter().take(argc).skip(optind) {
-        // TODO: This always gets all paths, and then skips a bunch.
-        // For the common case, we want to get just the one path.
-        // Port this over once path.cpp is.
-        let paths: Vec<WString> = path_get_paths(arg, &*parser.get_vars());
+        let paths = if opts.all {
+            path_get_paths(arg, &*parser.get_vars())
+        } else {
+            match path_get_path(arg, &*parser.get_vars()) {
+                Some(p) => vec![p],
+                None => vec![],
+            }
+        };
 
         for path in paths.iter() {
             res = true;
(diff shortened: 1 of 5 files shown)


fish-rust/src/builtins/command.rs
```diff
@@ -75,7 +74,7 @@ pub fn r#command(
         // TODO: This always gets all paths, and then skips a bunch.
         // For the common case, we want to get just the one path.
         // Port this over once path.cpp is.
-        let paths: Vec<WString> = path_get_paths_ffi(&arg.to_ffi(), parser).from_ffi();
```
krobelus, line 78: this one is special in that we want to remove the TODO workaround above, to match the C++ version again.
Something like

```diff
diff --git a/fish-rust/src/builtins/command.rs b/fish-rust/src/builtins/command.rs
index 1ac1dc407..7d8fc0e71 100644
--- a/fish-rust/src/builtins/command.rs
+++ b/fish-rust/src/builtins/command.rs
@@ -65,27 +77,31 @@ pub fn r#command(
 
     // Quiet implies find_path.
     if !opts.find_path && !opts.all && !opts.quiet {
         builtin_print_help(parser, streams, cmd);
         builtin_print_help(parser, streams, w.cmd());
         return STATUS_INVALID_ARGS;
     }
 
     let mut res = false;
     let optind = w.woptind;
     for arg in argv.iter().take(argc).skip(optind) {
-        // TODO: This always gets all paths, and then skips a bunch.
-        // For the common case, we want to get just the one path.
-        // Port this over once path.cpp is.
-        let paths: Vec<WString> = path_get_paths_ffi(&arg.to_ffi(), parser).from_ffi();
+        if opts.all {
+            let paths = path_get_paths(arg, parser.vars());
+            for path in paths {
+                res = true;
+                if opts.quiet {
+                    return STATUS_CMD_OK;
+                }
 
-        for path in paths.iter() {
-            res = true;
-            if opts.quiet {
-                return STATUS_CMD_OK;
+                streams.out.append(&sprintf!("%ls\n", path));
             }
-
-            streams.out.append(sprintf!("%ls\n", path));
-            if !opts.all {
-                break;
+        } else {
+            // Either find_path explicitly or just quiet.
+            if let Some(path) = path_get_path(arg, parser.vars()) {
+                if opts.quiet {
+                    return STATUS_CMD_OK;
+                }
+                res = true;
+                streams.out.append(&sprintf!("%ls\n", path));
             }
         }
     }
```

faho: Yeah, that's the simple solution.

Note that the same principle applies to `type`, but that never had that fast path. It appears type's author (whoever that may be, I hear he's very handsome) forgot.

So it's not for parity with C++, but still important performance-wise e.g. on WSL (where WSL2 by default gets a bunch of windows paths in $PATH that are slow to look up).

(the rust way to do this would probably be an iterator that just lazily returns the next matching path, but that's a bigger change for later)

faho: Might be simpler to go for this for now:

```rust
let paths = if opts.all {
    path_get_paths(&arg, &*parser.get_vars())
} else {
    match path_get_path(&arg, &*parser.get_vars()) {
        Some(p) => vec!(p),
        None => vec!(),
    }
};
```

This does have a vec even for one path, but tbh that's not a big deal and it saves on the annoying duplication, and makes it easier to switch to an iterator if we decide to do that.

Issue #9739, opened 2023-04-19
Rust builtins behave differently with NUL in arguments (mostly crash)


faho:
Check this out:

```fish
echo a\x00b
```

>thread '<unnamed>' panicked at 'String contained intermediate NUL character: ContainsNul { index: 1, inner: Some([97, 0, 98, 10]) }', src/wchar_ffi.rs:49:39

Mostly, rust builtins will crash when trying to print any argument that contains an embedded NUL like that (because the `\x00` will actually generate a NUL - but the same thing happens with e.g. `echo (find . -print0)`). Some will instead continue, and in some case "work", but differently than before:

- `type -q` will probably always return false (while `type a\x00b` will crash trying to print "Could not find 'aNULb'")
- `contains a\x00b a` will now fail, `contains a\x00b a\x00b` will succeed
- `echo` and `printf` crash

and so on.

A quick refresher so we're all on the same page: 

The C++ code passed the builtin arguments as `wchar_t*`, meaning they're NUL-terminated C-style strings. That means passing any NUL to a builtin would silently truncate the argument - `contains a\x00b a` would succeed. This matches what happens when you pass arguments to external commands, where you have to go through the c-string-using unix api.

We *could*, in future, change it so passing NULs to builtins is allowed, but it would have to have a few changes - `test`'s path tests should error out, for instance, `path` should either error out or automatically split on NUL like it does for stdin (`path dirname` specifically should *not* allow NULs because that's contrary to UNIX path semantics).

(also I've mentioned automatic splitting on NUL before and I still like that behavior)

For now, it's probably easiest to keep the current behavior by truncating on NUL when running a rust builtin - which would also fix the crashes.

ridiculousfish:
Agreed with truncating on NUL when passing strings as arguments at least for now. This keeps our builtins "honest" since they're on the same level with external commands, which of course cannot receive NUL.

mqudsi:
Taking a step back, two obvious solutions:

* conditionally fork only when we have to for cases like this, or 
* if we *really* want to avoid that, we could additionally compile each builtin to its own binary (or all of them to a single binary, hard-linked to the different builtin names the same way busybox does) and run that binary directly - but this means making sure builtins no longer used shared state, which is no small undertaking

cons are that builtin performance when used in a pipeline like this will suffer, which is probably acceptable.

faho:
[...]

PR #9854, opened 2023-06-20
Port/rewrite builtins/string to Rust

henrikhorluck:
- [x] - For regex replace/subsitution, I'm planning on going through rust-pcre2
- [x] - This requires a patched rust-pcre2 that makes `Captures` public to build (so far)
- [x] - I'm not a particular fan of how each subcommand was set up before or with this new setup, still a lot of boilerplate/duplication

Individual sub-commands to review:

- [x] Collect
- [x] Escape
- [x] Join
- [x] Length
- [x] Lower
- [x] Match
- [x] Pad
- [x] Repeat
- [x] Replace
- [x] Shorten
- [x] Split
- [x] Sub
- [x] Trim
- [x] Unescape
- [x] Upper

Please feel free to only look at the overall setup/some individual subcommands and not review everything at once, this is quite a lot

Merging is blocked on https://github.com/fish-shell/rust-pcre2/pull/1, unless you want to always point to my fork

PR #9859, opened 2023-06-22
Make NULs work for builtins

faho:
This switches from passing a c-string to output_stream_t::append to passing a proper string.

That means a builtin that prints a NUL no longer crashes with "thread '' panicked at 'String contained intermediate NUL character: ".

Instead, it will actually handle the NUL, even as an argument.

That means something like

`echo foo\x00bar` will now actually print a NUL instead of truncating after the `foo` because we passed c-strings around everywhere.

The former (not crashing) is *necessary* for e.g. `string`, the latter is a change that on the whole makes dealing with NULs easier, but it is a behavioral change.

To restore the c-string behavior we would have to truncate arguments at NUL.

See #9739.

The reason I am not outright declaring it *fixed* is because I would like consensus about this change in behavior.

For the record my opinion is that we *should* make it, that it is a good idea and that I don't believe there's a real concern about backwards-compatibility - accidental NULs basically don't happen.

## TODOs:

### DIFF
--- a/fish-rust/src/builtins/shared.rs
+++ b/fish-rust/src/builtins/shared.rs
@@ -1,7 +1,7 @@
 use crate::builtins::{printf, wait};
 use crate::ffi::{self, parser_t, wcstring_list_ffi_t, Repin, RustBuiltin};
 use crate::wchar::{wstr, WString, L};
-use crate::wchar_ffi::{c_str, empty_wstring, WCharFromFFI};
+use crate::wchar_ffi::{c_str, empty_wstring, ToCppWString, WCharFromFFI};
 use crate::wgetopt::{wgetopter_t, wopt, woption, woption_argument_t};
 use libc::c_int;
 use std::os::fd::RawFd;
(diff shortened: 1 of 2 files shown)


fish-rust/src/builtins/shared.rs
```diff
@@ -86,9 +86,9 @@ impl output_stream_t {
         unsafe { (*self.0).pin() }
     }
 
-    /// Append a &wtr or WString.
-    pub fn append<Str: AsRef<wstr>>(&mut self, s: Str) -> bool {
-        self.ffi().append1(c_str!(s))
+    /// Append a &wstr or WString.
+    pub fn append<Str: AsRef<wstr> + ToCppWString>(&mut self, s: Str) -> bool {
+        self.ffi().append(&s.into_cpp())
```
henrikhorluck, line 91:
```suggestion
        self.ffi().append(&s.as_ref().into_cpp())
```
Then you don't need the bound on `ToCppWString`, which saves some implementations when using e.g. `Cow<`a, wstr>`

faho: Thanks, implemented!

faho:
Okay I've merged this now. It's reasonably straightforward and fixes a real bug (output streams can contain NULs!).

@henrikhorluck You'll probably want to rebase #9854.

PR #9861, opened 2023-06-23
Restore NUL truncation for builtin arguments

faho:
## Description

This makes it so builtins can no longer see NUL (0x00) again. `echo foo\x00bar` will only see `foo`.

We do this by explicitly truncating the string ourselves, because of course rust no longer passes nul-terminated wchar_t* around.

This is likely to not be the final word on the issue, but until we have a good design for how working with NULs should work, this is the safest choice.

Fixes issue #9739 

## TODOs:


ridiculousfish (approved):
nice test!

PR #9866, opened 2023-06-28
Allow π and τ additionally for pi and tau as constants for math

porky11:
## Description

Now it's possible to use π and τ in the inbuilt math command.

Fixes no issue, I just created a pull request instead of an issue.


faho:
Thanks, but no. I don't believe this is useful and I don't want to further complicate this. "pi" and "tau" are short enough and have the advantage of being ASCII, so there's no issue with locale.

Sorry!

PR #9871, opened 2023-07-04
add stash completions to git show and git diff

m4rch3n1ng:
## Description

you can use `git show` and `git diff` with stashes

```
$ git show stash@{1}
$ git diff stash@{0} stash@{1}
$ git diff stash@{0}
```

this adds completions to those  
especially the former (`git show`) is fairly useful because `git stash show` only shows like with `--stat`

#### notes:
* you can also `git checkout` and `git switch --detach` (to) stashes, but i didn't want to encourage that, as you should probably use `git stash apply` or `git stash pop`, so you aren't missing new commits you added since creating the stash. if you want me to i can add that as well
* technically this should also apply to `git cherry-pick` (and probably `git cherry`) but that should probably be done via `git stash apply`, so i didn't add it
```
$ git cherry-pick stash@{1} -m 1
```

## TODOs:


zanchey (approved):

zanchey:
Thanks!

PR #9874, opened 2023-07-04
Make test_convert seedable, but generate the seed

henrikhorluck:


henrikhorluck:
@ridiculousfish Can you reproduce the failure locally with this change?

ridiculousfish:
Merged as a99fa201b618a22c4bff5549ec4b5d61358638b5, thank you

The test_convert failures are a race with setlocale, I've just called setlocale in the test_convert test in 98d88e06ff72993c1cf014c17a889c4738f36ee0

PR #9878, opened 2023-07-07
Fix rclone autocompletion script sourcing issue in fish shell

raxpd:
## Description

This pull request addresses the issue of the `rclone completion fish 2>/dev/null | source` command not properly sourcing the generated autocompletion script in the fish shell.

## Changes Made

Modified the `rclone completion fish` command to write the autocompletion script to stdout using the `-` option: `rclone completion fish - | source`.

## Documentation

The rclone completion fish command documentation can be found [here](https://rclone.org/commands/rclone_completion_fish/). This pull request ensures that the provided documentation reflects the correct usage of the command.

## TODOs:

### DIFF
--- a/share/completions/rclone.fish
+++ b/share/completions/rclone.fish
@@ -1,1 +1,1 @@
-rclone completion fish 2>/dev/null | source
+rclone completion fish - | source


faho:
Wow I'm surprised it needs that.

I would call this one a problem in rclone - if you don't give it `-` it'll error out with 

> open /etc/fish/completions/rclone.fish: permission denied

Which is... what the fudge? How? Why?

Regardless: Merged, thanks!

raxpd:
> 
> I would call this one a problem in rclone - if you don't give it `-` it'll error out with

It was actually a change that got introduced recently. I came across this yesterday, where rclone completion worked without this patch. 

```bash
❯ rclone version ; rclone completion fish | head
rclone v1.62.2
- os/version: arch (64 bit)
- os/kernel: 6.2.10-1-aarch64-ARCH (aarch64)
- os/type: linux
- os/arch: arm64 (ARMv8 compatible)
- go/version: go1.20.4
- go/linking: dynamic
- go/tags: none
# fish completion for rclone                               -*- shell-script -*-

function __rclone_debug
    set -l file "$BASH_COMP_DEBUG_FILE"
    if test -n "$file"
        echo "$argv" >> $file
    end
end

function __rclone_perform_completion
```

I then checked on two systems running `rclone v1.63`. Both needed `-`

However, if rclone version `< v1.63`, adding `-` will cause it to error.

```bash
❯ rclone version ; rclone completion fish - | head
rclone v1.62.2
- os/version: arch (64 bit)
- os/kernel: 6.2.10-1-aarch64-ARCH (aarch64)
- os/type: linux
- os/arch: arm64 (ARMv8 compatible)
- go/version: go1.20.4
- go/linking: dynamic
- go/tags: none
Error: unknown command "-" for "rclone completion fish"
Usage:
  rclone completion fish [flags]
```

faho:
Fudging heck that change is a pretty terrible idea and I wish they hadn't done it.

PR #9879, opened 2023-07-09
Rewrite cd builtin in Rust

zanchey:
riir cd

I've left the cwd_fd setting commented, as it's annoying to pass across the FFI,
and it's never actually used in the current code (but I guess might be in the
future?)


fish-rust/src/builtins/cd.rs
```diff
@@ -0,0 +1,181 @@
...
+                continue;
+            } else if errno::errno().0 == ENOTDIR {
+                best_errno = errno::errno().0;
+                continue;
+            }
+            best_errno = errno::errno().0;
+            break;
+        }
+
+        // Port note: sending the AutocloseFd across the FFI interface requires additional work
```
ridiculousfish, line 122: I would prefer to fix this now. I made a PR-against-your-PR if you want to take it. [Here's the commit.](https://github.com/zanchey/fish-shell/pull/3/commits/38c72f0e59e2feab26e7481ec0db073b71a4e025)

ridiculousfish (approved):
This is awesome!

PR #9881, opened 2023-07-09
Clean up Feature Flags API, add support for scoped feature tests

henrikhorluck:
Previously you needed `unsafe` to set feature flags from Rust. This also allows
scoped feature tests that makes testing feature flags thread-safe.

Note for posterity: you only needed `unsafe` since the `set`/`set_from_string` took `&mut self` when they could take `&self` since the feature flags are stored in atomics

### DIFF
--- a/fish-rust/src/builtins/status.rs
+++ b/fish-rust/src/builtins/status.rs
@@ -180,10 +180,10 @@ const LONG_OPTIONS: &[woption] = &[
 fn print_features(streams: &mut io_streams_t) {
     // TODO: move this to features.rs
     let mut max_len = i32::MIN;
-    for md in &features::METADATA {
+    for md in features::METADATA {
         max_len = max_len.max(md.name.len() as i32);
     }
-    for md in &features::METADATA {
+    for md in features::METADATA {
         let set = if feature_test(md.flag) {
             L!("on")
         } else {
@@ -433,7 +433,7 @@ pub fn status(
             }
             use TestFeatureRetVal::*;
             let mut retval = Some(TEST_FEATURE_NOT_RECOGNIZED as c_int);
-            for md in &features::METADATA {
+            for md in features::METADATA {
                 if md.name == args[0] {
                     retval = match feature_test(md.flag) {
                         true => Some(TEST_FEATURE_ON as c_int),
--- a/fish-rust/src/builtins/status.rs
+++ b/fish-rust/src/builtins/status.rs
@@ -10,7 +10,7 @@ use crate::ffi::{
     get_job_control_mode, get_login, is_interactive_session, job_control_t, parser_t,
     set_job_control_mode, Repin,
 };
-use crate::future_feature_flags::{feature_metadata, feature_test};
+use crate::future_feature_flags::{self as features, feature_test};
 use crate::wchar::{wstr, L};
 use crate::wchar_ffi::{AsWstr, WCharFromFFI};
 use crate::wgetopt::{
@@ -180,10 +180,10 @@ const LONG_OPTIONS: &[woption] = &[
 fn print_features(streams: &mut io_streams_t) {
     // TODO: move this to features.rs
     let mut max_len = i32::MIN;
-    for md in feature_metadata() {
+    for md in &features::METADATA {
         max_len = max_len.max(md.name.len() as i32);
     }
-    for md in feature_metadata() {
+    for md in &features::METADATA {
         let set = if feature_test(md.flag) {
             L!("on")
         } else {
@@ -192,10 +192,10 @@ fn print_features(streams: &mut io_streams_t) {
         streams.out.append(sprintf!(
             "%-*ls%-3s %ls %ls\n",
             max_len + 1,
-            md.name.as_wstr(),
+            md.name,
             set,
-            md.groups.as_wstr(),
-            md.description.as_wstr(),
+            md.groups,
+            md.description,
         ));
     }
 }
@@ -433,8 +433,8 @@ pub fn status(
             }
             use TestFeatureRetVal::*;
             let mut retval = Some(TEST_FEATURE_NOT_RECOGNIZED as c_int);
-            for md in &feature_metadata() {
-                if md.name.as_wstr() == args[0] {
+            for md in &features::METADATA {
+                if md.name == args[0] {
                     retval = match feature_test(md.flag) {
                         true => Some(TEST_FEATURE_ON as c_int),
                         false => Some(TEST_FEATURE_OFF as c_int),
--- a/fish-rust/src/future_feature_flags.rs
+++ b/fish-rust/src/future_feature_flags.rs
@@ -2,8 +2,6 @@
 
 use crate::ffi::wcharz_t;
 use crate::wchar::wstr;
-use crate::wchar_ffi::WCharToFFI;
-use std::array;
 use std::sync::atomic::AtomicBool;
 use std::sync::atomic::Ordering;
 use widestring_suffix::widestrs;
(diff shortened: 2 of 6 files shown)


fish-rust/Cargo.toml
```diff
@@ -29,6 +29,7 @@ rand = { version = "0.8.5", features = ["small_rng"] }
 unixstring = "0.2.7"
 widestring = "1.0.2"
 rand_pcg = "0.3.1"
+parking_lot = "0.12.1"
```
krobelus, line 32: parking_lot is a drop-in replacement for the standard sync primitives?
We should not have multiple ways of doing the same thing without a good reason.

fish-rust/src/future_feature_flags.rs
```diff
@@ -31,74 +28,49 @@ mod future_feature_flags_ffi {
...
-    values: [AtomicBool; metadata.len()],
+    values: [bool; METADATA.len()],
 }
 
 /// Metadata about feature flags.
-struct FeatureMetadata {
+pub struct FeatureMetadata {
     /// The flag itself.
-    flag: FeatureFlag,
+    pub flag: FeatureFlag,
```
krobelus, line 53: this commit should be split up, it does at least two things (the rwlock change and the refactoring)

fish-rust/src/builtins/status.rs
```diff
@@ -180,10 +180,10 @@ const LONG_OPTIONS: &[woption] = &[
 fn print_features(streams: &mut io_streams_t) {
     // TODO: move this to features.rs
     let mut max_len = i32::MIN;
-    for md in feature_metadata() {
+    for md in &features::METADATA {
         max_len = max_len.max(md.name.len() as i32);
     }
-    for md in feature_metadata() {
+    for md in &features::METADATA {
```
krobelus, line 186: good. The Rust code should never have used the feature_metadata FFI shim.

fish-rust/src/future_feature_flags.rs
```diff
@@ -134,36 +106,57 @@ const metadata: [FeatureMetadata; 4] = [
 ];
 
 /// The singleton shared feature set.
-static mut global_features: Features = Features::new();
+static FEATURES: RwLock<Features> = RwLock::new(Features::new());
```
krobelus, line 109: I don't see a reason for using locking instead of atomic bools.
(Also our current stance is to use the simpler Mutex unless we really need concurrent multiple readers)

IIRC feature flags are set early on during startup and never changed.
From that PoV, atomic bools make most sense because they are usually zero-overhead

fish-rust/Cargo.toml
henrikhorluck: The `std::sync::RwLock` does not support upgradeable readers/downgradeable writer, which is necessary for `scoped_test` to atomically change from a writer to a reader for the test to actually read the feature value. Sadly the diff for adding `scoped_test` got mangled. This is getting added to avoid race-conditions as tests in Rust are multi-threaded by default, unlike the C++ ones,  see https://github.com/fish-shell/fish-shell/pull/9854#discussion_r1257339267 for where this will be used.

We should be able to use `parking_lot` as a dev-dependency and use `std::sync:RwLock` otherwise, but I don't see the point in doing so as it just causes unnecessary `cfg[]`'s ,and `parking_lot::RwLock` is generally faster afaik https://crates.io/crates/parking_lot

krobelus: can we tell Rust to not run this test in parallel to others?
In Go you have to explicitly enable parallelism for tests..

fish-rust/src/future_feature_flags.rs
henrikhorluck: TBH this change can be `#[cfg(test)]`'d. I don't see the point arguing about the performance for functions that are called very rarely. This is a `RwLock` since we need to guarantee that a test that's dependant on one particular feature flag actually has that value during the test. `AtomicBool` does not work for that, and neither does a `Mutex`, unless I am misunderstanding how you are suggesting to use the mutex.

fish-rust/Cargo.toml
henrikhorluck: That requires either `cargo test -- --test-threads=1` or using a separate test-runner like https://nexte.st and using test groups https://nexte.st/book/test-groups.html

fish-rust/src/future_feature_flags.rs
henrikhorluck: Done!

ridiculousfish:
The cleanup commit looks good, thank you!

I agree with krobelus here: I am generally suspicious of read-write locks, and I would prefer for the common case to be a simple read of a global boolean, and not have the design be overly influenced by the needs of the tests.

How about introducing a thread-local `Option<Features>` which, if set, is used in preference to the global one?

The advantage is that `Features` can continue to use atomics, and the code to check the thread-local value can be made conditional on whether we are running the tests. This seems like a nice way to maintain test parallelism without overly distorting the design of `Features`.

henrikhorluck:
> I agree with krobelus here: I am generally suspicious of read-write locks, and I would prefer for the common case to be a simple read of a global boolean, and not have the design be overly influenced by the needs of the tests.
> 
> How about introducing a thread-local `Option<Features>` which, if set, is used in preference to the global one?
> 
> The advantage is that `Features` can continue to use atomics, and the code to check the thread-local value can be made conditional on whether we are running the tests. This seems like a nice way to maintain test parallelism without overly distorting the design of `Features`.

I don't think it matters enough to argue about. Changed it to a thread-local

fish-rust/src/future_feature_flags.rs
```diff
@@ -133,38 +106,77 @@ const metadata: [FeatureMetadata; 4] = [
...
+            fc.borrow()
+                .as_ref()
+                .unwrap_or(&FEATURES)
+                .set_from_string(wstr)
+        });
+    }
+    #[cfg(not(any(test, feature = "fish-ffi-tests")))]
+    {
+        FEATURES.set_from_string(wstr)
+    }
```
henrikhorluck, line 155: BTW I don't think this cfg is built in CI, as if so it should've failed in one of my pushes

mqudsi:
Instead of using parking_lot + RwLock, we can use an atomic counter and a mutex, no?

I feel like features are such a fringe feature that they shouldn't impose such a burden to the rest of the code. The use of `unsafe` for features support isn't really the end of the world.

mqudsi (commented):
Instead of using parking_lot + RwLock, we can use an atomic counter and a mutex, no?

I feel like features are such a fringe feature that they shouldn't impose such a burden to the rest of the code. The use of unsafe for features support isn't really the end of the world.

henrikhorluck:
> Instead of using parking_lot + RwLock, we can use an atomic counter and a mutex, no?

I really do not see any point in rolling our own atomic/sync container implementation just for features, this is not important for performance and I want correctness without having to think about it.

> I feel like features are such a fringe feature that they shouldn't impose such a burden to the rest of the code.

By using `scoped_test` everywhere instead of `set` with storing the previous value as the tests currently do that is what we achieve, no? The `scoped_test` addition is meant to make features more painless to use and test.

> The use of unsafe for features support isn't really the end of the world.

As noted in the PR description unsafe was not really needed, the API was just garbage, as you needed an `&mut Features` to modify features, but atomics can be modified with `&self`. The cleanup commit makes the whole `Features`-struct an implementation detail.

ridiculousfish:
Merged, thank you!

krobelus:
> the API was just garbage

yeah this was ported literally from C++

faho:
>yeah this was ported literally from C++

And that's a sort of mismatch that is to be expected in a port like this. There is going to be a certain amount of friction between C++ and Rust, and a fairly literal port is the best way to make it tractable - divide and conquer, so you don't get lost in architectural adventures. The trick is in knowing when not to do that.

And as a side note: While I know that calling someone else's code "garbage" is common in other projects, I would prefer not to have that here.

henrikhorluck:
> And as a side note: While I know that calling someone else's code "garbage" is common in other projects, I would prefer not to have that here.

Poor choice of words from my side.

PR #9894, opened 2023-07-15
Add preludes for wchar and builtins

henrikhorluck:
- Add a wchar prelude
- Add a builtin prelude

This makes is a lot easier to use in particular WExt, and ToWString-methods, making it more natural to use the widestr types. Also reduces a lot of repetition when porting builtins.

See https://doc.rust-lang.org/std/prelude/index.html#prelude-contents for what is included in `std`, which we should in at least some way mirror

I'm open for suggestions for more additions/removals

I would prefer if `builtins/string` was handled before this, as I'd prefer to rebase this


henrikhorluck:
Would it be preferable to instead have this PR as only introducing the prelude, and rather incrementally adopt it later to avoid merge conflicts?

fish-rust/src/wchar_ext.rs
```diff
@@ -274,8 +274,8 @@ impl WExt for wstr {
 
 #[cfg(test)]
 mod tests {
-    use super::WExt;
-    use crate::wchar::{wstr, WString, L};
+    use super::*;
+    use crate::wchar::L;
```
krobelus, line 278: I guess this one can go as well

fish-rust/src/builtins/mod.rs
```diff
@@ -28,3 +28,20 @@ pub mod wait;
...
+    pub(crate) use crate::{
+        ffi::{self, parser_t, separation_type_t, Repin},
+        wchar::prelude::*,
+        wchar_ffi::{c_str, AsWstr, WCharFromFFI, WCharToFFI},
+        wgetopt::{
+            wgetopter_t, wopt, woption,
+            woption_argument_t::{self, *},
+            NONOPTION_CHAR_CODE,
+        },
+        wutil::{fish_iswalnum, fish_wcstoi, fish_wcstol, fish_wcstoul},
```
krobelus, line 45: Nit: I'm not sure where to draw the line on what to include.
Some of these are not specific to builtins but I guess that's fine for the FFi types.
That leaves `fish_iswalnum` etc. which could be in another prelude.. but we can fix that later.

krobelus:
I think this is a good change, it makes porting easier and barely affects readability.

krobelus (approved):

fish-rust/src/wchar_ext.rs
henrikhorluck: `L` is not imported in the `wchar_ext` mod, and I somewhat wanted to avoid circular imports

fish-rust/src/builtins/mod.rs
henrikhorluck: If we compare this to plain Rust, this is mostly handled by `str::parse`, and therefore this is "by default available", which is why I included them. They are all namespaces behind `fish_` so including them should not be problematic. 

I guess `fish_is_walnum` is an odd-one out in that regard, and does not follow the above rationale. I will remove it.

I will also split out the prelude-adding and the adoption, sorry for making it hard to find the actual changes!

PR #9896, opened 2023-07-16
Rewrite the function store and builtin_function in Rust

ridiculousfish:
The first commit rewrites the global function store in Rust. The second commit adopts it, and simultaneously reimplements the function builtin, for definining functions (I found these to be too tightly coupled to handle separately).

Don't miss the new notion [NodeRef](https://github.com/ridiculousfish/fish-shell/blob/1e7321450f479835c641a1839f89c72f84c43332/fish-rust/src/parse_tree.rs#L119). This wraps up a parsed Ast along with a reference to a Node within that Ast, which points into the Ast. I expect this to be a foundational type.


fish-rust/src/global_safety.rs
```diff
@@ -16,3 +16,9 @@ impl RelaxedAtomicBool {
         self.0.swap(value, Ordering::Relaxed)
     }
 }
+
+impl Clone for RelaxedAtomicBool {
```
henrikhorluck, line 20: https://doc.rust-lang.org/std/sync/atomic/struct.AtomicBool.html does not implement Clone, so would probably be more appropriate to `impl From<bool>`, so you end up with `x.load().into()` vs `x.clone()`

fish-rust/src/function.rs
```diff
@@ -0,0 +1,711 @@
...
+            let func: &WString = &entry.name;
+            if !get_hidden && func.char_at(0) == '_' {
+                continue;
+            }
+            let suffix: Option<usize> = func.chars().rposition(|x| x == '.');
+            // We need a ".fish" *suffix*, it can't be the entire name.
+            if let Some(suffix) = suffix {
+                if suffix > 0 && entry.name.slice_from(suffix) == ".fish" {
+                    // Also ignore directories.
+                    if !entry.is_dir() {
```
mqudsi, line 177: TODO: move the "entry is dir" check before/outside the slicing logic, since that's just a bitwise check.

fish-rust/src/function.rs
```diff
@@ -0,0 +1,711 @@
...
+
+    /// List of all named arguments for this function.
+    pub named_arguments: Vec<WString>,
+
+    /// Description of the function.
+    pub description: WString,
+
+    /// Mapping of all variables that were inherited from the function definition scope to their
+    /// values.
+    pub inherit_vars: HashMap<WString, Vec<WString>>,
```
mqudsi, line 37: TODO: It would be much cheaper memory-wise to use a sorted vec rather than a `HashMap` here (this is per function!), and probably faster too. (How many variables does the typical function which shadows a variable actually end up shadowing?)

An empty vec is alloc-free but I think the same isn't true for an empty `HashMap`.

fish-rust/src/function.rs
```diff
@@ -0,0 +1,711 @@
...
+            event::remove_function_handlers(name);
+            true
+        } else {
+            false
+        }
+    }
+
+    /// Get the properties for a function, or None if none.
+    fn get_props(&self, name: &wstr) -> Option<FunctionPropertiesRef> {
+        self.funcs.get(name).cloned()
```
mqudsi, line 88: Am I the only one that would find all our rust code to imminently more readable if we got rid of the type aliases? Seeing `clone()` used everywhere gives me instinctual panic, then I realize the type we are returning is an `Arc`.

fish-rust/src/common.rs
```diff
@@ -1376,7 +1375,7 @@ pub fn narrow_string_safe(buff: &mut [u8; 64], s: &wstr) {
 }
 
 /// Stored in blocks to reference the file which created the block.
-pub type FilenameRef = Rc<WString>;
+pub type FilenameRef = Arc<WString>;
```
mqudsi, line 1378: Is the atomicity needed everywhere `FilenameRef` was already being used? (i.e. are there places where `Rc` is good enough but new places where `Arc` is required or does everything actually need `Arc` but the code exposing that need wasn't written until now?)

If so, getting rid of the alias and just using `Arc` or `Rc` as needed would be better than paying the `Arc` price everywhere.

mqudsi (approved):
LGTM. Left a few TODO comments for a future commit (maybe you can add a `// TODO: ` comment so they don't get lost?) and one question.

fish-rust/src/global_safety.rs
ridiculousfish: It's an interesting idea but the advantage of Clone is that it allows `#[derive(Clone)]` for other types containing `RelaxedAtomicBool`, which is what we're using here.

Presumably `AtomicBool` does not implement Clone because it is ambiguous which memory ordering to use; but RelaxedAtomicBool is specifically Relaxed so I think Clone is sensible.

fish-rust/src/function.rs
ridiculousfish: `is_dir()` may call `stat()` so we wish to defer it as long as possible.

fish-rust/src/function.rs
ridiculousfish: Good suggestion. I am pretty sure an empty HashMap does not allocate but it's still [gross](https://rust.godbolt.org/z/76YanTPPo). I didn't know there was TLS data involved.

 I switched to a `&[(WString, Vec<WString>)]`, it's nice now because the variables will be in a deterministic order.

fish-rust/src/common.rs
ridiculousfish: The reason to use Arc here is that, under concurrent execution, we may have two threads with the same filename reference, for example piping one block into another.

fish-rust/src/function.rs
ridiculousfish: I've removed this type.

fish-rust/src/common.rs
mqudsi: I understand; I was asking if there are any *other* usage sites  that are exclusively intra-thread rather than cross-thread, though. I’ll grep the code base later when I have a chance.

fish-rust/src/function.rs
mqudsi: Sorry, I thought we had already stat’d the entry by this point.

PR #9902, opened 2023-07-20
Rustify O_NONBLOCK operations

zanchey:
Add & adopt Rust implementations of `make_fd_blocking` and `make_fd_nonblocking`.

I've tried to make these more Rust idiomatic than the C++ versions. I don't like the naming of `perror_io` and would appreciate any other suggestions.

### DIFF
--- a/fish-rust/src/fds.rs
+++ b/fish-rust/src/fds.rs
@@ -3,10 +3,10 @@ use crate::ffi;
 use crate::wchar::{wstr, L};
 use crate::wutil::perror;
 use libc::EINTR;
-use libc::O_CLOEXEC;
+use libc::{fcntl, F_GETFL, F_SETFL, O_CLOEXEC, O_NONBLOCK};
 use nix::unistd;
 use std::ffi::CStr;
-use std::io::{Read, Write};
+use std::io::{self, Read, Write};
 use std::os::unix::io::{AsRawFd, FromRawFd, RawFd};
 
 pub const PIPE_ERROR: &wstr = L!("An error occurred while setting up pipe");
@@ -186,4 +186,30 @@ pub fn exec_close(fd: RawFd) {
             break;
         }
     }
+}
+
+/// Mark an fd as nonblocking
+pub fn make_fd_nonblocking(fd: RawFd) -> Result<(), io::Error> {
+    let flags = unsafe { fcntl(fd, F_GETFL, 0) };
+    let nonblocking = (flags & O_NONBLOCK) == O_NONBLOCK;
+    if !nonblocking {
+        match unsafe { fcntl(fd, F_SETFL, flags | O_NONBLOCK) } {
+            0 => return Ok(()),
+            _ => return Err(io::Error::last_os_error()),
+        };
+    }
+    Ok(())
+}
+
+/// Mark an fd as blocking
+pub fn make_fd_blocking(fd: RawFd) -> Result<(), io::Error> {
+    let flags = unsafe { fcntl(fd, F_GETFL, 0) };
+    let nonblocking = (flags & O_NONBLOCK) == O_NONBLOCK;
+    if nonblocking {
+        match unsafe { fcntl(fd, F_SETFL, flags & !O_NONBLOCK) } {
+            0 => return Ok(()),
+            _ => return Err(io::Error::last_os_error()),
+        };
+    }
+    Ok(())
 }
--- a/fish-rust/src/io.rs
+++ b/fish-rust/src/io.rs
@@ -3,8 +3,9 @@ use crate::common::{str2wcstring, wcs2string, EMPTY_STRING};
 use crate::fd_monitor::{
     FdMonitor, FdMonitorItem, FdMonitorItemId, ItemWakeReason, NativeCallback,
 };
-use crate::fds::{make_autoclose_pipes, wopen_cloexec, AutoCloseFd, PIPE_ERROR};
-use crate::ffi;
+use crate::fds::{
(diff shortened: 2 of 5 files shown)


fish-rust/src/fds.rs
```diff
@@ -187,3 +187,29 @@ pub fn exec_close(fd: RawFd) {
         }
     }
 }
+
+/// Mark an fd as nonblocking
+pub fn make_fd_nonblocking(fd: &RawFd) -> Result<(), io::Error> {
+    let flags = unsafe { fcntl(*fd, F_GETFL, 0) };
+    let nonblocking = (flags & O_NONBLOCK) == O_NONBLOCK;
+    if !nonblocking {
+        match unsafe { fcntl(*fd, F_SETFL, flags & O_NONBLOCK) } {
```
krobelus, line 196:
```suggestion
       match unsafe { fcntl(*fd, F_SETFL, flags | O_NONBLOCK) } {
```

fish-rust/src/io.rs
```diff
@@ -343,10 +344,13 @@ impl IoBufferfill {
...
         // from direct buffer transfers.
-        if ffi::make_fd_nonblocking(autocxx::c_int(pipes.read.fd())).0 != 0 {
-            FLOG!(warning, PIPE_ERROR);
-            perror("fcntl");
-            return None;
+        match make_fd_nonblocking(&pipes.read.fd()) {
+            Ok(_) => (),
+            Err(e) => {
+                FLOG!(warning, PIPE_ERROR);
+                perror_io("fcntl", &e);
```
krobelus, line 351: Going forward we should move the information about fcntl into the callee. Perhaps by making our custom error type. Something like

```rust
pub struct LibcError { // or SyscallError, or simply Error
	causer: &'static str,
	error: io::Error,
}
```

and define `std::fmt::Display` to print it as usual.

Here we'd get a `LibcError{causer: "fcntl", error: io::Error::last_os_error()}`

fish-rust/src/wutil/mod.rs
```diff
@@ -81,6 +81,10 @@ pub fn perror(s: &str) {
     let _ = stderr.write_all(b"\n");
 }
 
+pub fn perror_io(s: &str, e: &io::Error) {
```
krobelus, line 84: We can't overload `perror` to take either `str` or `LibcError` but we can define a `Perror` trait and implement it for either type (`"fcntl".perror()` looks funny alright).

fish-rust/src/fds.rs
```diff
@@ -187,3 +187,29 @@ pub fn exec_close(fd: RawFd) {
         }
     }
 }
+
+/// Mark an fd as nonblocking
+pub fn make_fd_nonblocking(fd: &RawFd) -> Result<(), io::Error> {
```
krobelus, line 192: moving this from `wutil` to `fds` without moving the C++ version as well can catch future porters offguard. I can live with this one

fish-rust/src/io.rs
henrikhorluck: Why not just use nix’s (some are even safe!) wrappers? https://docs.rs/nix/latest/nix/fcntl/index.html

krobelus: I don't think the error returned by Nix's fcntl includes an indication that it was fcntl that failed but yes we should cut back on raw libc calls. I mostly use raw libc calls because that's the easiest for porting. It probably wouldn't be a mistake to do so here as well, to make sure that porting callers is trivial.

fish-rust/src/wutil/mod.rs
henrikhorluck: If this is not used over FFI you can just have both be `impl Debug`, no need to reinvent [`std::error::Error`](https://doc.rust-lang.org/std/error/trait.Error.html)

krobelus: Debug wouldn't work because the implementation for `str` would do what `perror()` does today.
Of course eventually we want to get rid of `perror("fcntl")` etc. entirely but I wonder what's the best way to get there. My gut feeling is to do it after the port

fish-rust/src/fds.rs
zanchey: 4b4bf541d1 moved the declaration but not the implementation, which I think was a mistake

ridiculousfish: Yeah sloppy of me 😬

ridiculousfish, line 192: RawFd is just an int, it can be taken by value (`fd: RawFD`)

fish-rust/src/fds.rs
ridiculousfish: right, the idea is to preserve existing flags

fish-rust/src/fds.rs
```diff
@@ -187,3 +187,29 @@ pub fn exec_close(fd: RawFd) {
...
+        match unsafe { fcntl(*fd, F_SETFL, flags & O_NONBLOCK) } {
+            0 => return Ok(()),
+            _ => return Err(io::Error::last_os_error()),
+        };
+    }
+    Ok(())
+}
+
+/// Mark an fd as blocking
+pub fn make_fd_blocking(fd: &RawFd) -> Result<(), io::Error> {
```
ridiculousfish, line 205: can be just `fd: RawFd` here as well

fish-rust/src/io.rs
ridiculousfish: Agree with all of this but I think it doesn't have to be part of this PR

zanchey: OK, I've left it as is for now. Possibly using straight `perror` is fine, but because there's a `FLOG` first it's theoretically not correct.

PR #9903, opened 2023-07-20
Add Krita completions

EmilySeville7cfg:
## Description

Support Krita CLI.

Fixes issue #

## TODOs:


EmilySeville7cfg:
```fish
function __krita_complete_image_format -d 'Function to generate image format completion'
    set -l previous_token (commandline -oc)[-1]
    set -l current_token (commandline -opc)
    echo $current_token >&2

    if test $previous_token = --new-image
        switch $current_token
            case '*,'
                echo -e 'U8
U16
F16
F32' | awk "{ printf \"$current_token%s\n\", \$1 }"
            case '*'
                echo -e 'RGBA
                XYZA
                LABA
                CMYKA
                GRAY
                YCbCrA'
        end
    end
end
```

It doesn't work as intended: I wanna show color formats for `--new-image` but when there is no comma inside `--new-image` argument. Otherwise I want to display image depth. It should work like this:

- `krita --new-image<space><tab>`, result: `RGBA XYZA ...`
- `krita --new-image<space>RGBA,<tab>`, result: `U8 U16 ...`

EmilySeville7cfg:
I am not sure but if I manage it to implement dynamic suggestions for currently available workspaces (--workspace option), I provide such feature (but in the next PR).

share/completions/krita.fish
```diff
@@ -28,6 +28,11 @@ YCbCrA' $current_token
     end
 end
 
+function __krita_list_workspaces
+    ls --format single-column --color=never $HOME/.local/share/krita/workspaces/ |
```
faho, line 32: Please don't parse `ls` output, and don't mess with gnu-specific options.

Also the `string replace` doesn't use `-f` / `--filter`, so any files not called `.kws` will just be passed through as-is.

This should be:

```fish
path basename ~/.local/share/krita/workspaces/*.kws | path change-extension ''
```

share/completions/krita.fish
```diff
@@ -0,0 +1,56 @@
+function __krita_append_image_format_value -a value current_token
+    echo $value |
```
faho, line 2: `string` can take the value on the commandline, so the `echo` isn't useful:

`string replace -r '^' -- $current_token $value`

share/completions/krita.fish
```diff
@@ -0,0 +1,56 @@
...
+        string replace -r '^' $current_token |
+        string replace -r '$' ','
+end
+
+function __krita_complete_image_format
+    set -l previous_token (commandline -oc)[-1]
+    set -l current_token (commandline -t)
+
+    if test $previous_token = --new-image
```
faho, line 11: Any variable used with `test` needs to be quoted:

```fish
if test "$previous_token" = --new-image
```

share/completions/krita.fish
```diff
@@ -0,0 +1,55 @@
+function __krita_append_image_format_value -a value current_token
+    string replace -r '^' -- $current_token $value |
+        string replace -r '$' ','
```
faho, line 3: Okay, I didn't get what this was supposed to do at first, but this is just

```fish
string join , $value $current_token
```

I would remove this function and just do it in-line in __krita_complete_image_format.

share/completions/krita.fish
```diff
@@ -0,0 +1,55 @@
...
+
+    if test "$previous_token" = --new-image
+        switch $current_token
+            case '*,*,*'
+                # nothing is completed as arbitrary width and height are expected
+            case '*,'
+                __krita_append_image_format_value 'U8
+U16
+F16
+F32' $current_token
```
faho, line 18: So this would be `string join -- , $current_token U8 U16 F16 F32` - without the awkward line-breaks.

share/completions/krita.fish
```diff
@@ -0,0 +1,55 @@
...
+            case '*'
+                __krita_append_image_format_value 'RGBA
+XYZA
+LABA
+CMYKA
+GRAY
+YCbCrA' $current_token
```
faho, line 25: (same idea here)

share/completions/krita.fish
```diff
@@ -12,17 +7,17 @@ function __krita_complete_image_format
             case '*,*,*'
                 # nothing is completed as arbitrary width and height are expected
             case '*,'
-                __krita_append_image_format_value 'U8
+                echo 'U8
```
faho, line 10: No need for the `echo` and awkward line-breaks.

Just give it as arguments to `string join ,`, like I explained.

EmilySeville7cfg: This code doesn't work as expected:

```fish
function __krita_complete_image_format
    set -l previous_token (commandline -oc)[-1]
    set -l current_token (commandline -t)

    if test "$previous_token" = --new-image
        switch $current_token
            case '*,*,*'
                # nothing is completed as arbitrary width and height are expected
            case '*,'
                string join , $current_token U8 U16 F16 F32
            case '*'
                echo 'RGBA
XYZA
LABA
CMYKA
GRAY
YCbCrA' | string replace -r '$' ','
        end
    end
end
```

Completions for U8 U16 F16 F32 are broken: when I press tab having say XYZA already typed I get `krita --new-image GRAY,,U8,U16,F16,F32` instead of list of choices.

faho: Ah, hold on, you're expecting output like

```
XYZA,
LABA,
CMYKA,
```

That's easier as

```fish
printf '%s,\n' XYZA LABA CMYKA # ...
```

which also gets us away from the awkward linebreaks and echo-to-thing piping.

EmilySeville7cfg: Fixed. ;)

PR #9905, opened 2023-07-20
Feature: support Blender

EmilySeville7cfg:
## Description

Support GIMP CLI.

Fixes issue #

## TODOs:

### DIFF
--- a/share/completions/blender.fish
+++ b/share/completions/blender.fish
@@ -1,0 +1,11 @@
+function __blender_list_scenes -a file
+    blender -b $file --python-expr 'import bpy
+
+    for name in [scene.name for scene in list(bpy.data.scenes)]:
+        print(name)'
+end
+
+function __blender_list_addons
+    ls --format single-column /usr/share/blender/scripts/addons --color=never |
+        string replace --regex '.py$' ''
+end


share/completions/blender.fish
```diff
@@ -0,0 +1,125 @@
+function __blender_list_scenes -a file
+    stdbuf --output=0 blender --background test.blend --python-expr 'import bpy
+
+for name in [scene.name for scene in list(bpy.data.scenes)]:
+    print(name)' | head --lines=-2 | tail --lines=+6
+end
+
+function __blender_list_addons
+    ls --format single-column /usr/share/blender/scripts/addons --color=never |
```
faho, line 9: See https://github.com/fish-shell/fish-shell/pull/9903#discussion_r1271431744

share/completions/blender.fish
```diff
@@ -0,0 +1,125 @@
+function __blender_list_scenes -a file
+    stdbuf --output=0 blender --background test.blend --python-expr 'import bpy
+
+for name in [scene.name for scene in list(bpy.data.scenes)]:
+    print(name)' | head --lines=-2 | tail --lines=+6
```
faho, line 5: These head and tail numbers seem a bit arbitrary. What are they needed for?

share/completions/blender.fish
```diff
@@ -0,0 +1,125 @@
+function __blender_list_scenes -a file
+    stdbuf --output=0 blender --background test.blend --python-expr 'import bpy
```
faho, line 2: stdbuf is non-standard - it's a GNU thing, we can't use it unless absolutely necessary.

share/completions/blender.fish
faho: (this is why splitting these up into separate PRs is awkward - you're adding code of the same form to all of these, but now I need to cross-reference my comments from the other PRs)

share/completions/blender.fish
EmilySeville7cfg: Without these numbers there are extra unneded output:
```
Color management: using fallback mode for management
Color management: Error could not find role data role.
Blender 3.0.1
Read prefs: /home/emilyseville7cfg/.config/blender/3.0/config/userpref.blend
Color management: scene view "Filmic" not found, setting default "Standard".
Read blend: /home/emilyseville7cfg/Documents/open-source/other/forks/fish-shell/test.blend
Scene
Scene.001

Blender quit
```

Everything is being put in stdout, so I can't just discard anything in an easy way.

share/completions/blender.fish
EmilySeville7cfg: Without it piping to head/tail messes up the stdout output:

image

share/completions/blender.fish
faho: Yeah that seems very much like it won't always be six lines.

What do the lines you *do* want look like? Can you `print("MARK" + name)` and then `string replace --filter` only the marked lines?

share/completions/blender.fish
faho: See above, you want to match the lines you're interested in.

Also: What `stdbuf` is doing here is turning off buffering on stdout. If that changes the order things appear in, that's suspect.

I would double-check that things aren't being printed on stderr (which is unbuffered by default!).

share/completions/blender.fish
EmilySeville7cfg: > What do the lines you do want look like? Can you print("MARK" + name) and then string replace --filter only the marked lines?

AFAIK they can be any non-empty lines. They can even contain just spaces...

share/completions/blender.fish
EmilySeville7cfg: I've added a guard condition to print nothing when stdbuf is not available. I've tried redirect `stderr` to `/dev/null` but it didn't help, unfortunately.

share/completions/blender.fish
faho: Here's what comes up for me:

```
Blender 3.6.0 (hash c7fc78b81ecb built 2023-07-07 13:20:07)
Failed to open dir (No such file or directory): /run/user/1000/gvfs/
Error: Cannot read file "/home/alfa/test.blend": No such file or directory

Blender quit
```

Note that these are very different errors from what you get. It doesn't print anything usable, probably because I don't have any scenes.

Note: The `test.blend` error goes to stderr, the rest goes to stdout.


What you want to test is:

```fish
for name in [scene.name for scene in list(bpy.data.scenes)]:
    print("COMPLETION:", name)' | string replace -rf '^COMPLETION:' ''
```

i.e. instead of the `head` and `tail` guff that, as you see above, isn't usable (because the number of errors differs!), mark the completion line with a string and then only get those.

share/completions/blender.fish
```diff
@@ -1,18 +1,19 @@
...
-    ls --format single-column /usr/share/blender/scripts/addons --color=never |
-        string replace -r '.py$' ''
+    path basename /usr/share/blender/scripts/addons/*.py | path change-extension ''
 end
 
 function __blender_list_engines
-    stdbuf --output=0 blender --background --engine help |
-        tail --lines=+6 | string replace -r '^\s+' ''
+    blender --background --engine help |
+        tail --lines=+7 | string replace -r '^\s+' ''
```
faho, line 16: Try `string replace -rf`, remove the `tail`.

The output I get:

```
Blender 3.6.0 (hash c7fc78b81ecb built 2023-07-07 13:20:07)
Blender Engine Listing:
	BLENDER_EEVEE
	BLENDER_WORKBENCH
	CYCLES
```

So the engines are indented, the rest isn't. `tail` isn't usable because we get different numbers of lines.

(also `tail` long options aren't usable because they are a gnu-ism!)

The gvfs error goes to stderr, so that needs to be silenced.

share/completions/blender.fish
faho: And once you've done that, the order no longer matters and you can remove the `stdbuf`.

share/completions/blender.fish
```diff
@@ -1,10 +1,10 @@
...
-    if command -q stdbuf
-        stdbuf --output=0 blender --background test.blend --python-expr 'import bpy
+    blender --background $file --python-expr 'import bpy
 
 for name in [scene.name for scene in list(bpy.data.scenes)]:
-    print(name)' | head --lines=-2 | tail --lines=+6
-    end
+    print(f"\t{name}")' |
+        string match -e -r '^\s+' |
+        string replace -r '^\s+' ''
```
faho, line 7: `string replace -rf`, then you can remove the `match`.

The `-f` is short for `--filter`, and tells `string replace` to only let things through where it needs to replace, so it's short for *exactly* this kind of `match | replace` pipeline.

share/completions/blender.fish
```diff
@@ -0,0 +1,124 @@
...
+        string split -n ' ' |
+        string match -r -v '^-' |
+        head --lines=1
+end
+
+function __blender_complete_addon_list
+    set -l previous_token (commandline -oc)[-1]
+    set -l current_token (commandline -t)
+
+    if test $previous_token = --addons
```
faho, line 28: This still needs to be quoted.

EmilySeville7cfg: I am confused: I see my quotation fix [there](https://github.com/EmilySeville7cfg/fish-shell/commit/5745c7bc1ad2f93f9f99661dbd8544b187e1568e) but not [here](https://github.com/fish-shell/fish-shell/pull/9905/files#diff-886e83748a6c40680df274dcd2694c8ad5ca4383bad2ec7a1553ea070963730eR28).

EmilySeville7cfg: image

I feel something went wrong: in a web editor I see quotes around $previous_token, while in your comment - not.

faho:
Merged, thanks!

PR #9915, opened 2023-07-26
Docs: correct small grammatical error in read.rst

asfaltboy:
## Description

Just noticed this sentence doesn't make sense (the construct is typically used as  'while something, something else' but we just end the sentence early); I assumed it intended to reference the `while` command, so I attempted to link to it to help the reader to read more if they wished so.

Fixes issue #

## TODOs:


faho:
The intended construct here is "The following code stores the value 'hello' in $foo: (some code), while this can be used to read line-by-line".

Which is a tad awkward so let's use this PR instead. Thanks, merged!

PR #9726, opened 2023-04-10
WIP: Port parser.h

krobelus:

PR #9916, opened 2023-07-26
Port wildcard to Rust

henrikhorluck:
- Port the easy part of wildcard.{h,cpp}
- Port wilcard.{cpp,h} to Rust

Please avoid looking at this until after `builtins/string`, I might end up porting more of `expand/complete` so that this is actually used and tested before this is merge-able


fish-rust/src/wildcard.rs
```diff
@@ -11,3 +33,1259 @@ pub const ANY_STRING_RECURSIVE: char = char_offset(WILDCARD_RESERVED_BASE, 2);
...
+            // Trailing slashes and accepting incomplete, e.g. `echo /xyz/<tab>`. Everything is added.
+            let Ok(mut dir) = self.open_dir(base_dir, false) else {
+                return;
+            };
+
+            // wreaddir_resolving without the out argument is just wreaddir.
+            // So we can use the information in case we need it.
+            let need_dir = self.flags.contains(ExpandFlags::DIRECTORIES_ONLY);
+
+            while let Some(Ok(entry)) = dir.next() {
```
henrikhorluck, line 706: This had an unused `path` 
https://github.com/fish-shell/fish-shell/blob/ed881bcdd8d7351c5b48daf22ca6e3d9344c2d5d/src/wildcard.cpp#L692C1-L692C1

Introduced by this:

https://github.com/fish-shell/fish-shell/commit/a2d816710fc336e0a964e11b798cd5fb19043760#diff-d6f14992d417e6d5eb82c1590cd41b90fa804b60fd38ad6c2cf9efb66d73d1f5L740

Xiretza:
See also https://github.com/fish-shell/fish-shell/commit/b21e11cc24172607de2e486b762a8292c938dce5 in the seemingly dead `riir-parser` branch.

krobelus:
Yeah we have draft implementations of most of the core modules but they are untested and have no C++ bridges.
I haven't worked on it in months.. I'll clean it up a bit and submit a draft PR. From August 6 I'll have some time to complete it.
I'm happy to let someone else take over and follow their preferred approach but otherwise, [here](https://github.com/fish-shell/fish-shell/pull/9726#issuecomment-1510432132) is a checklist of our pending implementations (plus most of history IIRC).

krobelus:
I've pushed my progress to the `riir-parser`. Still missing the history patches.

There is a lot of `todo!()` comments left in there, some questionable changes, but usually with a good reason.
It ports the parser and all its dependencies. Doesn't fully compile rn.
At this point, it's easier to port the builtins (that use the parser API) than to try to bridge them. I ported `disown` and `jobs` today.
I understand it's probably not easy for others to work with this mess..

If we can break out things that work on their own we should merge them first of course. But many modules are a bit intertwined. For example wildcard+complete+expand is a clique. I didn't check if it's worth to define FFI bridges for wildcard.

henrikhorluck:
> If we can break out things that work on their own we should merge them first of course. But many modules are a bit intertwined. For example wildcard+complete+expand is a clique. I didn't check if it's worth to define FFI bridges for wildcard.

FWIW: the first commit here actually runs the tests and works, yours had issues with using slices+indices instead of `&wstr` + `.char_at()`. (yes I know this has some improper variable-renaming, but it is now actually closer to the source-reference). Though it only ports `wildcard_match` itself.

> At this point, it's easier to port the builtins (that use the parser API) than to try to bridge them. I ported disown and jobs today.
I understand it's probably not easy for others to work with this mess..

I am not familiar enough with the codebase to determine what is the best path forward regarding merging your `riir-parser`-branch. I would guess it is advisable to "pick your battles" and split some of it out (like the trio wildcard, expand, complete), but that still requires `operation_context_t` and some of history.

I cherry-picked the clique-changes and put it in another (compiling and CI-passing) branch: https://github.com/henrikhorluck/fish-shell/commit/d4684c893e413f8e21a9352d31b168ca8e224375, which now needs:

- `history_t` over FFI (or combine with your implementation) for `expand_variables` and `complete_variables`
- `parset_t.transient_commandlines` (should be fairly trivial) 
- OperationContext working over FFI (what is in the branch is know completely wrong, but compiles, no need to comment on it)
- I have not gone over this vs your `wildcard.rs` — currently has this version, but will do that later
- `exec_subshell_for_expand` — again, should be relatively easy
- Add FFI-bindings, use it and then fix the plethora of bugs

I would like help on FFI of `OperationContext` and `history_t` — that could hopefully be a separate PR (or two) by itself, only adding FFI-bindings + minimal tests of them.

But that is only one way to do it, you have more information to decide if that is a reasonable way forward or if it is better to keep working on `riir-parser` directly until it works. I am not comfortable with overtaking ownership of the full changes.

Aside: regarding IoStreams, I think it is better to avoid object-safety and use an `OutputStream`-enum that delegates, like [`StringMatcher`](https://github.com/fish-shell/fish-shell/blob/2ec36338f2b1f21f4074ff7184c37619a4a61e69/fish-rust/src/builtins/string/match.rs#L157) from `string match`. Then we can avoid the API breakage of no generics. Using `dyn` would be fine if we used [`io::Write`](https://doc.rust-lang.org/stable/std/io/trait.Write.html) + [`write!(streams.out, output)`](https://doc.rust-lang.org/stable/std/macro.write.html), but that's a _much_ larger change that I don't think we want right now. I think that a delegating enum is the closest we have to subclassing from C++, even if it is a bit of boilerplate.

I can also cherry-pick the changes to `wgetopt` that you sent with `builtins/string`, as those should be independent.

krobelus:
I'm about to go offgrid for a week now so apologies for being curt/incoherent.
Essentially, anything that you can extract (whether it be started in `riir-parser` or otherwise) in a way that is bridged to the rest of the code and passes all tests, we should definitely merge first.
Or if you want to add some stuff to `riir-parser` that would also work just fine I think but might be difficult to parallelize.
To me personally it feels too slow to port&bridge everything file-by-file. It feels more fun to go in bigger strides. But this might very easily be a fallacy. The extra debugging time might not be worth it. Also it's a rabbit hole. Almost anything involving the parser or anything else that's somewhat complex I don't really feel like bridging; it's easier to port the dependants.
BTW wildcard, expand and history were ported by @Xiretza.

> I can also cherry-pick the changes to wgetopt that you sent with builtins/string, as those should be independent.
EDIT: scratch that, it doesn't work since it ends up with index out of bonds

They have some conflicts with argparse which I resolved in `riir-parser`.. it should definitely be extracted into a precursor commit.

fish-rust/src/common.rs
```diff
@@ -1074,7 +1074,8 @@ pub static EMPTY_STRING_LIST: Vec<WString> = vec![];
 
 /// A function type to check for cancellation.
 /// \return true if execution should cancel.
-pub type CancelChecker = dyn Fn() -> bool;
+/// This needs a lifetime to allow closures that capture for less than static
+pub type CancelChecker<'closure> = dyn FnMut() -> bool + 'closure;
```
henrikhorluck, line 1078: This might actually disagree with what `OperationContext` want — in that case these should rather be separate types as `FnMut` is more general than `Fn`

henrikhorluck (commented):
This now has a combination of the two attempts, I think the most sensible way forward is to merge this in a partially adopted state, and later adopt and actually test the majority of it

fish-rust/src/common.rs
krobelus: I'm not sure we want to change all references to `&mut` just for this type; its implementations don't mutate anything (though you're right they could in theory).

krobelus: (references of OperationContext I mean)

henrikhorluck: We just need to take it by value here in `Wildcard`, see this playground https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ce59c7510e0307fa51c18c28bc795e74

henrikhorluck: My suggestion: Take it by value, get rid of the type alias (there are a lot of type-aliases that I honestly think just obscure the code, this is one of them)

krobelus:
I have integrated these changes in my branch.
Since this PR deletes the C++ version of `wildcard_match`, it is probably reasonably well tested and we can and should merge it separately.
I haven't had time for a thorough review yet.

fish-rust/src/common.rs
Xiretza: I don't know about removing the type entirely, it does serve as immediate documentation in function signatures. How about making it a transparent newtype (`pub struct CancelChecker<'closure>(pub dyn FnMut() -> bool + 'closure);`) instead? Maybe with a `.check()` method if you don't want to access `.0` all the time.

henrikhorluck: I don't think that is a valuable abstraction _for fish_. We already have `OperationContext::check_cancel`.

krobelus: Right, in the C++ version, wildcard expander also took cancel checker by value, so our default should be to match that.
In general there are lots of things to clean up, would be nice if someone could tackle that in future.

henrikhorluck: Just finish the port first, then cleaning stuff up will be much easier :)

henrikhorluck: Done https://github.com/fish-shell/fish-shell/pull/9916/commits/54437d70314a429af0afbbfa7d9c535cef3ee011, I had to rebase to make CI pass as well

krobelus: does this mean that OperationContext references need to be `mut`? I haven't looked into why we actually need FnMut

henrikhorluck: No, as the playground showed you can use `&dyn Fn() -> bool` where the type is `impl FnMut() -> bool + 'closure`.

The reason I changed it to `FnMut` was because I was trying to add FFI-bindings, and that ended up something like:

```rust
// this is pseudo-code
fn use_opaque(x: &mut std::function-ish) -> bool {
    call(x)
}

let opaque_cpp_cancel_checker: std::function-ish = X;
let cancel_checker: impl FnMut() -> bool = || {
    // we capture `x` here for the lifetime of the closure, which is 'closure
    use_opaque(&mut opaque_cpp_cancel_checker)
};
```

And I got compiler errors due to `FnMut`, and my reaction was: hmm, it makes sense to allow `FnMut`, let me change that

krobelus: great

fish-rust/src/complete.rs
```diff
@@ -0,0 +1,185 @@
...
+        if vec.len() > self.limit - self.completions.len() {
+            return false;
+        }
+        self.completions.extend(vec);
+        true
+    }
+
+    /// Swap our completions with a new list.
+    pub fn swap(&mut self, lst: &mut Vec<Completion>) {
+        // XXX: This completly breaks our completions.len() <= self.limit invariant
```
krobelus, line 157: good catch on the broken invariant. Let's remove this function entirely, it's unused.

fish-rust/src/complete.rs
```diff
@@ -0,0 +1,185 @@
...
+    /// \return the list of completions, clearing it.
+    pub fn take(&mut self) -> Vec<Completion> {
+        std::mem::take(&mut self.completions)
+    }
+
+    /// \return a new, empty receiver whose limit is our remaining capacity.
+    /// This is useful for e.g. recursive calls when you want to act on the result before adding it.
+    pub fn subreceiver(&self) -> Self {
+        // XXX: this should not need to be saturating, we have a faulty invariant
+        let remaining_capacity = self.limit.saturating_sub(self.completions.len());
```
krobelus, line 182: Yeah, let's crash (in a separate commit) rather than trying to fix things here

fish-rust/src/wildcard.rs
```diff
@@ -11,3 +48,1249 @@ pub const ANY_STRING_RECURSIVE: char = char_offset(WILDCARD_RESERVED_BASE, 2);
...
+    completion: &mut &wstr,
+    expand_flags: ExpandFlags,
+    description_func: Option<&'f dyn Fn(&wstr) -> WString>,
+) -> WString {
+    if let Some(complete_sep_loc) = completion.find_char(PROG_COMPLETE_SEP) {
+        // This completion has an embedded description, do not use the generic description.
+        assert!(
+            completion.len() > complete_sep_loc,
+            "resolve_description lacks a length assumption"
+        );
```
krobelus, line 75: This assertion is trivially true, let's remove it?

fish-rust/src/wildcard.rs
```diff
@@ -11,3 +48,1249 @@ pub const ANY_STRING_RECURSIVE: char = char_offset(WILDCARD_RESERVED_BASE, 2);
...
+        let (comp, description) = completion.split_at(complete_sep_loc + 1);
+        *completion = comp;
```
krobelus, line 77: Testing revealed there's an off-by-one error in this code; "completion" should not include
the separator. I changed it to

        let description = completion[complete_sep_loc + 1..].to_owned();
        *completion = &completion[..complete_sep_loc];

fish-rust/src/wildcard.rs
```diff
@@ -11,3 +48,1249 @@ pub const ANY_STRING_RECURSIVE: char = char_offset(WILDCARD_RESERVED_BASE, 2);
...
+    // Maybe we have no more wildcards at all. This includes the empty string.
+    if next_wc_char_pos.is_none() {
+        // Try matching
+        let Some(m) = string_fuzzy_match_string(wc, s, false) else {
+            return WildcardResult::NoMatch;
+        };
+
+        // If we're not allowing fuzzy match, then we require a prefix match.
+        let needs_prefix_match = !params.expand_flags.contains(ExpandFlags::FUZZY_MATCH);
+        if needs_prefix_match && m.is_exact_or_prefix() {
```
krobelus, line 146:
```suggestion
        if needs_prefix_match && !m.is_exact_or_prefix() {
```

I guess I should have done the review before testing (but then I might have missed these)

fish-rust/src/wildcard.rs
```diff
@@ -11,3 +48,1249 @@ pub const ANY_STRING_RECURSIVE: char = char_offset(WILDCARD_RESERVED_BASE, 2);
...
+) -> &'static wstr {
+    let Some(lstat) = lstat else {
+        return *COMPLETE_FILE_DESC;
+    };
+
+    fn is_executable(buf: &fs::Metadata, filename: &wstr) -> bool {
+        // Weird group permissions and other such issues make it non-trivial to find out if
+        // we can actually execute a file using the result from stat. It is much safer to
+        // use the access function, since it tells us exactly what we want to know.
+        (buf.mode() as mode_t & (S_IXUSR | S_IXGRP | S_IXOTH) != 0) && waccess(filename, X_OK) == 0
```
krobelus, line 331: nit: it would feel cleaner to use

`mode_t::from(buf.mode())`

instead of `buf.mode() as mode_t`.
However if the from fails on some platform it's probably not worth our time.

fish-rust/src/wildcard.rs
```diff
@@ -11,3 +48,1249 @@ pub const ANY_STRING_RECURSIVE: char = char_offset(WILDCARD_RESERVED_BASE, 2);
...
+    }
+
+    // stat was only queried if lstat succeeded
+    let stat = stat.unwrap();
+    if lstat.is_symlink() {
+        return match stat {
+            Ok(stat) if stat.is_dir() => *COMPLETE_DIRECTORY_SYMLINK_DESC,
+            Ok(stat) if is_executable(&stat, filename) => *COMPLETE_EXEC_LINK_DESC,
```
krobelus, line 339: what happened to the `definitely_executable` parameter?

fish-rust/src/wildcard.rs
```diff
@@ -11,3 +48,1249 @@ pub const ANY_STRING_RECURSIVE: char = char_offset(WILDCARD_RESERVED_BASE, 2);
...
+        && string_suffixes_string_case_insensitive(L!(".dll"), filename)
+    {
+        return false;
+    }
+
+    // Compute the description.
+    let desc = if expand_flags.contains(ExpandFlags::GEN_DESCRIPTIONS) {
+        let mut desc = file_get_desc(filename, lstat, stat).to_owned();
+
+        if !is_directory && !is_executable {
```
krobelus, line 453: missing the `file_size >= 0` check

fish-rust/src/wildcard.rs
```diff
@@ -11,3 +48,1249 @@ pub const ANY_STRING_RECURSIVE: char = char_offset(WILDCARD_RESERVED_BASE, 2);
...
+        }
+
+        Some(desc)
+    } else {
+        None
+    };
+
+    // Append a / if this is a directory. Note this requirement may be the only reason we have to
+    // call stat() in some cases.
+    let x = |_: &wstr| match desc.as_ref() {
```
krobelus, line 467: x should probably be called desc_func (to be shadowed immediately)

fish-rust/src/wildcard.rs
```diff
@@ -11,3 +48,1249 @@ pub const ANY_STRING_RECURSIVE: char = char_offset(WILDCARD_RESERVED_BASE, 2);
...
+            flags: ExpandFlags,
+            cancel_checker: &'e mut dyn FnMut() -> bool,
+            resolved_completions: &'e mut CompletionReceiver,
+        ) -> Self {
+            Self {
+                cancel_checker,
+                working_directory,
+                completion_set: resolved_completions
+                    .iter()
+                    .map(|c| c.completion.to_owned())
```
krobelus, line 544: isn't clone() more correct semantically than to_owned()?

fish-rust/src/wildcard.rs
```diff
@@ -11,3 +48,1249 @@ pub const ANY_STRING_RECURSIVE: char = char_offset(WILDCARD_RESERVED_BASE, 2);
...
+                return;
+            }
+            // Trailing slashes and accepting incomplete, e.g. `echo /xyz/<tab>`. Everything is added.
+            let Ok(mut dir) = self.open_dir(base_dir, false) else {
+                return;
+            };
+
+            // wreaddir_resolving without the out argument is just wreaddir.
+            // So we can use the information in case we need it.
+            let need_dir = self.flags.contains(ExpandFlags::DIRECTORIES_ONLY);
```
krobelus, line 723: missing this part:

        if (flags & expand_flag::special_for_cd) {
            path = this->working_directory;
            append_path_component(path, base_dir);
            // cd operates on logical paths.
            // for example, cd ../<tab> should complete "without resolving symlinks".
            path = normalize_path(path);
        }

looks like we don't have a test for it

fish-rust/src/wildcard.rs
```diff
@@ -11,3 +48,1249 @@ pub const ANY_STRING_RECURSIVE: char = char_offset(WILDCARD_RESERVED_BASE, 2);
...
+        ) {
+            let is_dir = false;
+            let need_dir = self.flags.contains(ExpandFlags::DIRECTORIES_ONLY);
+
+            while !self.interrupted_or_overflowed() {
+                let Some(Ok(entry)) = base_dir_iter.next() else {
+                    break;
+                };
+
+                if need_dir && entry.is_dir() {
```
krobelus, line 877:
```suggestion
                if need_dir && !entry.is_dir() {
```

fish-rust/src/wildcard.rs
```diff
@@ -11,3 +48,1249 @@ pub const ANY_STRING_RECURSIVE: char = char_offset(WILDCARD_RESERVED_BASE, 2);
...
+                        }
+
+                        px += 1;
+                        nx += 1;
+                        continue;
+                    }
+                }
+                c => {
+                    // ordinary char
+                    if nx < name.len() && name.char_at(nx) == c {
```
krobelus, line 1191: ok so we know that given "c == ANY_CHAR && nx == name.len()" we will never match this condition,
because name never contains ANY_CHAR. Therefore it's fine to include that case in the above ANY_CHAR

fish-rust/src/wildcard.rs
krobelus: nevermind there's already a thread

krobelus:
Awesome work, this was very easy to review

fish-rust/src/wildcard.rs
henrikhorluck: Uhm, it appears I forgot to re-add that when I rebased at some point. I can fix it but that will probably lead to some minor merge conflicts in your branch

krobelus: That explains it. When you rebased, you got a conflict an resolved it in favor of your version. You could have used something like `git log riir/wildcard..origin/master src/wildcard.cpp` to show why the conflict exists.

> I can fix it but that will probably lead to some minor merge conflicts in your branch

Please do. Merge conflicts are our friends. Learn to love them.

henrikhorluck (commented):
I also modified `CompletionReceiver::extend` to use `ExactSizeIterator`

```rust
#[must_use]
    pub fn extend(
        &mut self,
        iter: impl IntoIterator<Item = Completion, IntoIter = impl ExactSizeIterator<Item = Completion>>,
    ) -> bool
```

The version in C++ only had `add_list` so this is still a relaxation, and allows us to skip `collect`-ing before checking the length, but with an assert afterwards that can detect bogus `ExactSizeIterator` implementations

fish-rust/src/wildcard.rs
krobelus: sorry I somehow thought it's != 0

fish-rust/src/wildcard.rs
krobelus: In the C++ code, the case of `c == ANY_CHAR && nx = name.len()` was handled in the "ordinary char case" not in the `ANY_CHAR` case, that's what was interesting

PR #9919, opened 2023-07-28
add gcc completion lm lz lrt

wuyudi:
## Description

add `-lm -lz -lrt` options.

ref https://stackoverflow.com/questions/5663097/gcc-lm-lz-lrt-options-what-are-they-about

I found that it would be better to auto search lib under `/usr/lib/lib*.so`, but I haven't figured out how to write it.

Fixes issue #

## TODOs:


zanchey:
Thanks!

asukaminato0721:
> I found that it would be better to auto search lib under /usr/lib/lib*.so, but I haven't figured out how to write it.

use

```sh
ls /usr/lib/ | string match -r -g "^lib(.*?).so\$"
```

Issue #7401, opened 2020-10-15
Set timeout for waiting for a key sequence to complete


comfortablynick:
fish, version 3.1.2-1530-g06bd1e934

With a clean environment, I run these commands:
```fish
fish_vi_key_bindings

# kj -> Esc
bind -M insert kj 'if commandline -P; commandline -f cancel; else; set fish_bind_mode default; commandline -f backward-char repaint-mode; end'
```

Any time I press `k`, fish waits indefinitely for the `j`. Is there an equivalent of `timeoutlen` in Vim? By default, Vim will only wait 1 sec for `j` before assuming I'm just pressing `k`. 

```
'timeoutlen' 'tm'       number  (default 1000)
                          global
Time in milliseconds to wait for a mapped sequence to complete.
```

Fish seems to assume I'm *never* just pressing `k`. There is `$fish_escape_delay_ms` but it doesn't seem to affect this. Shouldn't there be something like this for regular key bindings?

ridiculousfish:
There's nothing like that now except for escape. This seems like an interesting general purpose idea - it would make multi-key bindings more tractable.

yanshay:
Is there any way to workaround this by more complex coding using fish functions or some other way to get the 'kj'/'jk' sequence work as expected?

faho:
>Is there any way to workaround this by more complex coding using fish functions or some other way to get the 'kj'/'jk' sequence work as expected?

No.

yanshay:
I managed to get this working for my "kj" scenario with a simple change to input.cpp as below, but I'm not sure about potential implications on more complex scenarios. Can someone familiar with the code comment on implications?

```
--- a/src/input.cpp
+++ b/src/input.cpp
@@ -475,7 +475,11 @@ class event_queue_peeker_t {
         if (idx_ == peeked_.size()) {
             char_event_t newevt{L'\0'};
             if (!timed) {
-                newevt = event_queue_.readch();
+                if (auto mevt = event_queue_.readch_timed()) {
+                    newevt = mevt.acquire();
+                } else {
+                    return false;
+                }
             } else if (auto mevt = event_queue_.readch_timed()) {
                 newevt = mevt.acquire();
             } else {
```

faho:
>Can someone familiar with the code comment on implications?

What you're doing is *basically* removing the "timed"/"not timed" distinction.

[...]

PR #9926, opened 2023-07-29
Added fish_sequence_key_delay_ms to set time to wait between sequence key presses (#7401)

yanshay:
## Description

This resolves issue #7401

When binding key sequences, currently, fish waits indefinitely for sequence to complete. So if one wants to press only a single character which is a beginning of a sequence, that character will not appear. To make it appear need to press another character, not in the sequence and then backspace. Not very reasonable.
For example, if using 'kj' to exit insert mode in fish vi mode, if one presses 'k' it will not show until another key is pressed, so to type 'k' at end of command need to press 'k', then another character and then backspace. Not to speak that to enter kj as characters its even more difficult.

The solution introduced here is to set fish_sequence_key_delay_ms variable (along the lines of fish_escape_delay_ms) which is the time to wait for the next key in a sequence. Once that time passes the key will be output as usual. By default the behavior is as today and setting the key changes the behavior.

I wrote the code and tested it manually. If you expect it to enter future versions I will update documentation and changelog as well, and will try to add tests for the new behavior.

I also think it's worth considering making this the default behavior using some reasonable value. 

## TODOs:


faho:
I finally got a chance to test this.

Thanks, this seems like solid work! I'll be adding some docs and tests, and we'll see if a default timeout makes sense - nvim sets it to 1 second.

faho:
Tagging for 3.7.0 because it's *possible* to backport, and would be cool to have - but it would require some work.

yanshay:
When will 3.7 come out?

faho:
"When it's done".

I'm afraid experience shows we can't really do estimates.

yanshay:
Can I back-port this PR?
If I recall correctly I originally implemented it on top of 3.6 branch and only later had to contribute to master.
I use it all the time (practically whenever I type J or K on command line which is probably quite a few times an hour ...)

faho:
> Can I back-port this PR?

If you want, you can do it, otherwise I'll see once I get the time.

You'll want to target [the Integration_3.7.0 branch](https://github.com/fish-shell/fish-shell/tree/Integration_3.7.0).

yanshay:
> You'll want to target [the Integration_3.7.0 branch](https://github.com/fish-shell/fish-shell/tree/Integration_3.7.0).

Maybe I misunderstood something.

I thought you tagged it to 3.7, therefore it won't come out in 3.6.2.
Is 3.7 is going to be based on Master (with more rust code, and my PR was adjusted for that)?
So if I want it to be released in 3.6.2 shouldn't I merge it to Integration 3.6.2?

faho:
> I thought you tagged it to 3.7, therefore it won't come out in 3.6.2

3.7.0 is going to be the next C++-based release. We decided that it was a bit much for a 3.6.2. Integration_3.7.0 is branched from Integration_3.6.1, not master.

The rust-based version will probably be 4.0, *possibly* 3.8.

yanshay:
> We decided that it was a bit much for a 3.6.2.

'it' meaning this change? So no point in submitting a PR for this on 3.6.2?

faho:
>  Or do you mean there isn't going to be a 3.6.2?

There isn't going to be a 3.6.2. What was intended for 3.6.2 became a bit much for a patch release, and so we're going to call it 3.7.0 instead.

yanshay:
I submitted a PR with back-port to 3.7.0 : #10008

PR #9927, opened 2023-07-31
In .editorconfig replace `max_line_length: none` with `off`.

ModProg:
## Description
According to
https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#max_line_length
that is the correct value to disable this property.


faho:
Thanks!

PR #9929, opened 2023-08-02
Fix build

henrikhorluck:


faho:
Oops, sorry, I had the same idea: 5de19d2

PR #9930, opened 2023-08-02
Avoid segfaulting if user has invalid locale

henrikhorluck:
- Don't segfault if user has an invalid locale

Some of the added `assert!`s are definitively not necessary, but that is a question of code style I would like answered in a review. I also realised I did not follow the comment-convention regarding full sentences...

### DIFF
--- a/fish-rust/src/env_dispatch.rs
+++ b/fish-rust/src/env_dispatch.rs
@@ -11,8 +11,10 @@ use crate::wchar::{wstr, WString};
 use crate::wchar_ext::WExt;
 use crate::wutil::fish_wcstoi;
 use crate::wutil::wgettext;
+use std::borrow::Cow;
 use std::collections::HashMap;
 use std::ffi::{CStr, CString};
+use std::ptr;
 use std::sync::atomic::{AtomicBool, Ordering};
 
 #[cxx::bridge]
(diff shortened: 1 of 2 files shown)


fish-rust/src/env_dispatch.rs
```diff
@@ -693,11 +695,14 @@ fn init_locale(vars: &EnvStack) {
         "C.UTF-8", "en_US.UTF-8", "en_GB.UTF-8", "de_DE.UTF-8", "C.utf8", "UTF-8",
     ];
 
-    let old_msg_locale: CString = unsafe {
-        let old = libc::setlocale(libc::LC_MESSAGES, std::ptr::null());
+    let old_msg_locale: CString = {
+        let old = unsafe { libc::setlocale(libc::LC_MESSAGES, ptr::null()) };
+        assert_ne!(old, ptr::null_mut());
```
Xiretza, line 700:
```suggestion
        assert!(!old.is_null());
```

henrikhorluck: It was intentional to have it more verbose, `assert!(!` is a bit syntax-heavy, but there is no `is_not_null`

faho:
Thanks!

PR #9931, opened 2023-08-03
Remove a waccess call when completing executables

### LINKED ISSUE #9699: Fish Running as Root Fails to Suggest All Executable Commands in PATH
In fish 3.6.1 as well as 3.3.1 at least, when running as root, commands (scripts or binaries) that are in $PATH, not owned by user root and/or group root, and are not world-executable will not be offered as completions, even though they are executable by root. This appears to affect Enterprise Linux (RHEL, CentOS, and Oracle Linux) 7 and 8, but does not affect macOS (any version). This happens whether ssh-ing to the Linux system via Apple Terminal or connecting directly to the machine's console. Running without third-party customizations does not affect the behavior.

How to reproduce:
1. Ensure your user account on the Linux system can sudo to root and that /usr/local/bin is in your $PATH
2. Create /usr/local/bin/testprog, using sudo if needed, containing _something_ that could execute. An example:
```bash
#!/usr/bin/env fish
echo "Hello World"
```
3. Set the owner and group of testprog to something OTHER than root:root. For full effect, make sure your user account is either the owner or belongs to the group.
5. Set the permissions of testprog to 0550, 0750, or 0770
6. Attempt to auto-complete or tab-complete the command `testprog` (without typing it fully) and notice that fish will happily
7. Change to a root session running fish and ensure that /usr/local/bin is in your path as root
8. Attempt to auto-complete or tab-complete the command `testprog` (without typing it fully) and notice that fish will not
9. Run the command `testprog` (notice that fish will color the command as valid once it's typed fully)
10. Attempt to auto-complete or tab-complete the command `testprog` (without typing it fully) and notice that fish will now _suggest_ it, since it's in your command history, but will not _tab-complete_ it.
12. Switch to a root shell running bash
13. Attempt to tab-complete the command `testprog` (without typing it fully) and notice that bash will happily
14. Do any one of the following: change the owner of testprog to root, change the group of testprog to root, or set testprog's permissions to 0555, 0750, 0775, or 0777
15. Switching back to a fish root prompt, try tab-completing `testprog` again. Note that fish will happily tab-complete it now.

What should happen:
* If a command is executable at all, it's executable by root, and should be offered as a completion

What does happen:
* Fish only seems to offer commands for completion when the following are true: it is executable by anyone OR it is executable specifically by you OR it is executable by a group you belong to OR it's in your command history

faho:
When we run these completions, we have already run waccess with X_OK. We already *know* the file is executable.

There is no reason to check again.

Restores some of the speedup from the fast_waccess hack that was removed to fix #9699.

In my tests it's ~1.14x - if fish with c67d77f reverted takes 1x, this takes 1.2 times as long and fish right before this takes 1.37x as long. (vs 3.6.1 it's 1.22x reverted / 1.5x this / 1.7x master)

The absolute time difference will scale with the size of $PATH.

Note: This will conflict with any port of wildcard, like #9916. I would still like to merge this because it should be in 3.6.2 and offset some of the loss of #9699.

I'm aware that this adds another argument to an already argument-heavy function, but it's only called in one spot and I would leave that cleanup to after a port (a lazy "file information" object would be quite cool to have here, but might be awkward considering TOCTTOU issues, but at the very least the "stat_res" and "stat buf" arguments would just be some combined thing like an Option or Result).

## TODOs:

### DIFF
--- a/src/wildcard.cpp
+++ b/src/wildcard.cpp
@@ -340,7 +340,7 @@ wildcard_result_t wildcard_complete(const wcstring &str, const wchar_t *wc,
 /// \param err The errno value after a failed stat call on the file.
 static const wchar_t *file_get_desc(const wcstring &filename, int lstat_res,
                                     const struct stat &lbuf, int stat_res, const struct stat &buf,
-                                    int err) {
+                                    int err, bool definitely_executable) {
     if (lstat_res) {
         return COMPLETE_FILE_DESC;
     }
@@ -350,10 +350,12 @@ static const wchar_t *file_get_desc(const wcstring &filename, int lstat_res,
             if (S_ISDIR(buf.st_mode)) {
                 return COMPLETE_DIRECTORY_SYMLINK_DESC;
             }
-            if (buf.st_mode & (S_IXUSR | S_IXGRP | S_IXOTH) && waccess(filename, X_OK) == 0) {
+            if (definitely_executable || (buf.st_mode & (S_IXUSR | S_IXGRP | S_IXOTH) && waccess(filename, X_OK) == 0)) {
                 // Weird group permissions and other such issues make it non-trivial to find out if
                 // we can actually execute a file using the result from stat. It is much safer to
                 // use the access function, since it tells us exactly what we want to know.
+                //
+                // We skip this check in case the caller tells us the file is definitely executable.
                 return COMPLETE_EXEC_LINK_DESC;
             }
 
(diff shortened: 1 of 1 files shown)


src/wildcard.cpp
```diff
@@ -444,7 +444,7 @@ static bool wildcard_test_flags_then_complete(const wcstring &filepath, const wc
     // Compute the description.
     wcstring desc;
     if (expand_flags & expand_flag::gen_descriptions) {
-        desc = file_get_desc(filepath, lstat_res, lstat_buf, stat_res, stat_buf, stat_errno);
+        desc = file_get_desc(filepath, lstat_res, lstat_buf, stat_res, stat_buf, stat_errno, executables_only);
```
henrikhorluck, line 447: Can we at least make this its own variable? The control flow here is not very obvious.

So this:

```cpp
bool file_is_executable = false;
if (executables_only) {
    if (is_executable && waccess(filepath, X_OK) == 0)) {
        file_is_executable = true;
    } else {
        return false;
    }
}
```

where we have `if (executables_only && (!is_executable || waccess(filepath, X_OK) != 0))`

henrikhorluck (commented):
I'm fine with adding the parameter, that function will have way fewer parameters in Rust

src/wildcard.cpp
henrikhorluck: LGTM then

zanchey:
Note for me - if we're taking this for 3.6.2 we will need to take #9699 as well

PR #9935, opened 2023-08-06
Use aws_completer in aws completions

mdwint:
## Description

This updates the AWS CLI completions to use `aws_completer`, which [it comes bundled with](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-completion.html) nowadays.

## TODOs:

### DIFF
--- a/share/completions/aws.fish
+++ b/share/completions/aws.fish
@@ -1,3 +1,10 @@
+# AWS CLI comes bundled with a completion command.
+# https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-completion.html
+if type -q aws_completer
+    complete -c aws -f -a "(env COMP_LINE=(commandline -pc) aws_completer)"
+    exit 0
+end
+
 # These are very much incomplete completions for the aws-cli suite.
 # In addition to a complete list of services, the `aws s3` completions are mostly complete
 # (and are the primary reason this file exists). The automatically generated completions
--- a/share/completions/aws.fish
+++ b/share/completions/aws.fish
@@ -1,7 +1,7 @@
 # AWS CLI comes bundled with a completion command.
 # https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-completion.html
 if type -q aws_completer
-    complete -c aws -f -a "(env COMP_LINE=(commandline -pc) aws_completer)"
+    complete -c aws -f -a "(COMP_LINE=(commandline -pc) aws_completer)"
     exit 0
 end
 
--- a/share/completions/aws.fish
+++ b/share/completions/aws.fish
@@ -1,7 +1,13 @@
 # AWS CLI comes bundled with a completion command.
 # https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-completion.html
 if type -q aws_completer
-    complete -c aws -f -a "(COMP_LINE=(commandline -pc) aws_completer)"
+    function __aws_completer
+        set -lx COMP_LINE (commandline -pc)
+        set -lx COMP_POINT (string length $COMP_LINE)
+        aws_completer
+    end
+
+    complete -c aws -f -a "(__aws_completer)"
     exit 0
 end
 


share/completions/aws.fish
```diff
@@ -1,3 +1,10 @@
+# AWS CLI comes bundled with a completion command.
+# https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-completion.html
+if type -q aws_completer
+    complete -c aws -f -a "(env COMP_LINE=(commandline -pc) aws_completer)"
```
zanchey, line 4: I don't think the `env` invocation is needed here

faho:
How good are these completions?

They're using the bash way of figuring out the commandline, which sounds to me like they won't have descriptions, which are pretty useful to have.

Also typically you would set $COMP_LINE and $COMP_POINT which tells you where the cursor is. See https://github.com/fish-shell/fish-shell/blob/f4132af114c7f8e6bdaca39ae41c573d4bdf4bb6/share/functions/__fish_argcomplete_complete.fish#L6-L7 - which you could probably reuse.

mdwint:
> How good are these completions?
> 
> They're using the bash way of figuring out the commandline, which sounds to me like they won't have descriptions, which are pretty useful to have.

True, they are bash completions without descriptions, but to me that's an improvement over the ones in `share/completions/aws.fish` before. Many commands were missing, and AWS CLI keeps expanding.

Example:

```
$ aws ecs list-services
…-services  …-services-by-namespace

$ aws ecs list-services --c
--ca-bundle        --cli-binary-format    --cli-input-json  --cli-read-timeout  --color
--cli-auto-prompt  --cli-connect-timeout  --cli-input-yaml  --cluster
```

The AWS CLI contains hundreds of subcommands, each with a handful of options. Descriptions of each command and option could be extracted from the help text I guess, but I'm not sure how robust that would be.

Speaking from personal experience, when I use the AWS CLI I'm usually not trying to discover what a command does (you'd need the docs for that anyway), I'm just trying to get the spelling right without consulting `--help`. For example: "Is it `--service` or `--service-name` this time?".

---

> Also typically you would set $COMP_LINE and $COMP_POINT which tells you where the cursor is.

Thanks. I'll add `COMP_POINT`.

mqudsi:
My previous `aws s3` completions are fully dynamic and query the S3 bucket to suggest bucket and path completions - does aws_completer do that?

mdwint:
> My previous `aws s3` completions are fully dynamic and query the S3 bucket to suggest bucket and path completions - does aws_completer do that?

No, `aws_completer` does not complete S3 bucket names or paths.

I haven't tried integrating my changes with your existing completions yet. Maybe that's worth a try? I wasn't sure because I've seen them fail in some cases, so I (wrongly) assumed they were broken. For example, when passing a `--profile`:

```console
$ aws --profile my-profile s3 ls s3://<tab>
An error occurred (InvalidAccessKeyId) when calling the ListBuckets operation: The AWS Access Key Id you provided does not exist in our records.
```

The `aws s3 ls` command inside the completion doesn't take into account the profile passed via the CLI. It does work when `AWS_PROFILE` is set in the environment.

Do you think it's viable to make the existing S3 bucket/path completions more robust, and combining them with those provided by `aws_completer`? I'm open to giving that a try.

mqudsi:
The completion callback would need to introspect the arguments for variables that might change the success/failure outcome. I purposely didn't bother with access keys specified at the command line as it's just not something I'd ever do (or recommend others do) but I can definitely see the value in supporting `--profile`. We have (lots of?) existing examples of completions where the completion function re-uses arguments (if present) from the original command line you could model this after.

Using *both* the (patched) existing "smart" completions and the "dumb" completions from aws_completer can be done the wrong way (load both) or the hard way (either preload only the non-s3 aws_completer completions or dynamically choose between using aws_completer or the handwritten completions depending on whether it's an s3 completion at time of invocation).

scooter-dangle:
How hard would it be to transform the JSON help documentation (under site-packages/awscli/botocore/data/ in the Python installation) into Fish completions and fold that in as another step as part of `fish_update_completions`?

E.g., for my installation of aws, the file /opt/homebrew/Cellar/awscli/2.9.16/libexec/lib/python3.11/site-packages/awscli/botocore/data/glue/2017-03-31/service-2.json has entries like the following under the top-level `operations` key:

```json
    "BatchCreatePartition":{
      "name":"BatchCreatePartition",
      "http":{
        "method":"POST",
        "requestUri":"/"
      },
      "input":{"shape":"BatchCreatePartitionRequest"},
      "output":{"shape":"BatchCreatePartitionResponse"},
      "errors":[
        {"shape":"InvalidInputException"},
        {"shape":"AlreadyExistsException"},
        {"shape":"ResourceNumberLimitExceededException"},
        {"shape":"InternalServiceException"},
        {"shape":"EntityNotFoundException"},
        {"shape":"OperationTimeoutException"},
        {"shape":"GlueEncryptionException"}
      ],
      "documentation":"<p>Creates one or more partitions in a batch operation.</p>"
    }
```

where the `input.shape` corresponds to another entry in the same file that looks like this

```json
    "BatchCreatePartitionRequest":{
      "type":"structure",
      "required":[
        "DatabaseName",
        "TableName",
        "PartitionInputList"
      ],
      "members":{
        "CatalogId":{
          "shape":"CatalogIdString",
          "documentation":"<p>The ID of the catalog in which the partition is to be created. Currently, this should be the Amazon Web Services account ID.</p>"
        },
        "DatabaseName":{
          "shape":"NameString",
          "documentation":"<p>The name of the metadata database in which the partition is to be created.</p>"
        },
```

For context, I'm curious about ingesting this metadata directly rather than via `aws_completer` in part due to having gone through the exercise of porting fish's manpage completion generator to Rust ([rust-dc/fish-manpage-completions](https://github.com/rust-dc/fish-manpage-completions)) with others in the Rust DC meetup.

zanchey:
My impression from the discussion above is that we don't want to take this as it stands, so I'm pushing out of 3.7.0.

faho:
Alright, I'm closing this because it seems to be stalled for the moment.

If you (or anyone else) returns to work on it we'd of course happily consider it!

PR #9939, opened 2023-08-08
Simplify and fix `__fish_is_zfs_feature_enabled`

glaeqen:
## Description

Previously `__fish_is_zfs_feature_enabled` was doing
`<whitespace>$queried_feature<whitespace>` pattern matching which
was skipping the state part expected in the follow-up checking code.

Passing the dataset/snapshot in a `target` argument is pointless. As
none of the existing code attempts to do this plus it is also a
private function (`__` prefix), rename of the argument and removal
of extra text replacement should not be considered a breaking change.

## TODOs:

### DIFF
--- a/share/functions/__fish_is_zfs_feature_enabled.fish
+++ b/share/functions/__fish_is_zfs_feature_enabled.fish
@@ -1,17 +1,7 @@
-function __fish_is_zfs_feature_enabled -a feature target -d "Returns 0 if the given ZFS feature is available or enabled for the given full-path target (zpool or dataset), or any target if none given"
-    type -q zpool
-    or return
-    set -l pool (string replace -r '/.*' '' -- $target)
-    set -l feature_name ""
-    if test -z "$pool"
-        set feature_name (zpool get -H all 2>/dev/null | string match -r "\s$feature\s")
-    else
-        set feature_name (zpool get -H all $pool 2>/dev/null | string match -r "$pool\s$feature\s")
-    end
-    if test $status -ne 0 # No such feature
-        return 1
-    end
-    set -l state (echo $feature_name | cut -f3)
-    string match -qr '(active|enabled)' -- $state
-    return $status
+function __fish_is_zfs_feature_enabled \
+    -a feature pool \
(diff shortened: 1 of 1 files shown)


share/functions/__fish_is_zfs_feature_enabled.fish
```diff
@@ -1,17 +1,7 @@
...
-        return 1
-    end
-    set -l state (echo $feature_name | cut -f3)
-    string match -qr '(active|enabled)' -- $state
-    return $status
+function __fish_is_zfs_feature_enabled \
+    -a feature pool \
+    -d "Returns 0 if the given ZFS pool feature is active or enabled for the given pool or for any pool if none specified"
+
+    type -q zpool && \
```
faho, line 5: Instead of doing this awkward `&& \`, just use `and` on the next line, or do `|| return` on this.

I'm pretty sure `fish_indent`, our autoformatter, would remove all your line continuations here.

glaeqen: > Instead of doing this awkward `&& \`, just use `and` on the next line, or do `|| return` on this.

Fixed 👍🏻 

> I'm pretty sure `fish_indent`, our autoformatter, would remove all your line continuations here.

That's true, it'd would. Good point about the formatter, wasn't aware it exists. It fixed a single whitespace there.

Thanks for the quick response :) Great project, I use it everyday 💪🏻❤️

faho:
>You seem not to use automatic squash on the PR merge, do you prefer me to squash them manually into a single commit so it looks nicer on master?

We can just squash it by pressing the "squash and merge" button, which I'm doing now. Thanks!

Strum355:
Is it possible to get this backported to 3.6? I've been having to keep a [patched (via nixos) version of fish](https://github.com/Strum355/nix-dotfiles/blob/master/pkgs/fish.nix) containing this PR (+ another diff so that the patch from this PR applies correctly) and would be happy to have this in a release :slightly_smiling_face:

zanchey:
It'll be in 3.7.0 which is on its way soon.

PR #9944, opened 2023-08-10
Rewrite builtin functions in rust

faho:
## Description

This is a rough draft of rewriting builtin functions in rust. Basically the first time it passed the tests.

It fixes two pre-existing bugs:

1. `functions --details foo bar` will mention whatever the last option is in the error - so `functions --details --verbose foo bar` will mention "--verbose". I don't believe the error is all that great anyway, but now it always uses `--details` because that's the "mode" that functions is in
2. The `event::print` code compared the enum values too deeply, and so it printed the "Event thing" header too often

Honestly the second felt extremely roundabout and awkward, and I'm pretty sure there's a nicer way.

In terms of implementation decisions I mostly decided not to make a report_function_metadata helper function because using that many mutable variables felt like going against the grain and didn't seem like it would have helped a lot.

## TODOs:


fish-rust/src/builtins/functions.rs
```diff
@@ -0,0 +1,413 @@
...
+        if args.len() != 1 {
+            streams.err.append(wgettext_fmt!(
+                BUILTIN_ERR_ARG_COUNT2,
+                cmd,
+                // This error is
+                // functions: --details: expected 1 arguments; got 2
+                // The "--details" was "argv[optind - 1]" in the C++
+                // which would just give the last option.
+                // This is broken because you could do `functions --details --verbose foo bar`, and it would error about "--verbose".
+                "--details",
```
faho, line 198: This is a bit dirty because it means we'll no longer use `-D` if that was passed, but given that it could mention other options, and otherwise this would mean reparsing the options or storing the options I felt that was an acceptable trade-off.

fish-rust/src/builtins/functions.rs
```diff
@@ -0,0 +1,413 @@
...
+                            &props.description,
+                            EscapeStringStyle::Script(
+                                EscapeFlags::NO_PRINTABLES | EscapeFlags::NO_QUOTED
+                            )
+                        ),
+                    ));
+                }
+            }
+        } else if opts.verbose {
+            streams.out.append(L!("n/a\nn/a\n0n/a\nn/a\n"));
```
faho, line 251: This would need to be adjusted if we add any more metadata - the C++ would set the different entries to "n/a" on initialization.

I could also add mutation like that or maybe return multiple values or something?

fish-rust/src/builtins/functions.rs
```diff
@@ -0,0 +1,413 @@
...
+    }
+
+    if opts.query && args.is_empty() {
+        return STATUS_CMD_ERROR;
+    }
+
+    if opts.list || args.is_empty() {
+        let mut names = function::get_names(opts.show_hidden);
+        names.sort();
+        if !streams.out_is_redirected && unsafe { isatty(STDOUT_FILENO) == 1 } {
```
faho, line 280: We've had this a few times, a helper method might be nice to avoid multiple "unsafe" blocks.

fish-rust/src/builtins/functions.rs
```diff
@@ -0,0 +1,413 @@
...
+            let mut buff = WString::new();
+            let mut first: bool = true;
+            for name in names {
+                if !first {
+                    buff.push_utfstr(L!(", "));
+                }
+                buff.push_utfstr(&name);
+                first = false;
+            }
+            // TODO: reformat_for_screen
```
faho, line 290: See the TODO - this would have to be implemented before it can be merged.

(not that the reformatting or even the ","-separated format is all that great - I consider it a mistake, tbh)

fish-rust/src/builtins/functions.rs
```diff
@@ -0,0 +1,413 @@
...
+    let mut res: c_int = STATUS_CMD_OK?;
+
+    let mut first = true;
+    for arg in args.iter() {
+        if opts.query {
+            if !function::exists(arg, parser) {
+                res += 1;
+            }
+        } else if let Some(props) = function::get_props_autoload(arg, parser) {
+            let path = props.definition_file().unwrap_or(L!(""));
```
faho, line 357: This is awkward, taken from type. This should probably be a match, and could also be a helper function.

Would that go in builtins/shared.rs?

fish-rust/src/builtins/functions.rs
```diff
@@ -0,0 +1,413 @@
...
+        builtin_print_error_trailer(parser, streams, cmd);
+        return STATUS_CMD_OK;
+    }
+
+    let describe = !opts.description.is_empty();
+    if [describe, opts.erase, opts.list, opts.query, opts.copy]
+        .into_iter()
+        .filter(|b| *b)
+        .count()
+        > 1
```
faho, line 141: This seemed like the cleanest way to do this sort of "only one" check. The C++ just added the bools together, that's possible here but would require an `as i32` cast for every bool.

fish-rust/src/builtins/functions.rs
```diff
@@ -0,0 +1,413 @@
...
+        }
+        return STATUS_CMD_ERROR;
+    }
+
+    let mut res: c_int = STATUS_CMD_OK?;
+
+    let mut first = true;
+    for arg in args.iter() {
+        if opts.query {
+            if !function::exists(arg, parser) {
```
faho, line 353: This is a slight change in that we now check function::exists for query, instead of seeing that the properties don't exist.

fish-rust/src/builtins/functions.rs
```diff
@@ -0,0 +1,413 @@
...
+            if !streams.out_is_redirected && unsafe { isatty(STDOUT_FILENO) == 1 } {
+                let col = colorize_shell(&def.to_ffi(), parser.pin()).from_ffi();
+                streams.out.append(col);
+            } else {
+                streams.out.append(def);
+            }
+        } else {
+            res += 1;
+        }
+        first = false;
```
faho, line 409: (this always feels extremely clunky in any programming language for what is a very common thing - "only print a newline for every following entry")

fish-rust/src/event.rs
```diff
@@ -845,19 +845,27 @@ pub fn print(streams: &mut io_streams_t, type_filter: &wstr) {
...
-        if last_type.as_ref() != Some(&evt.desc) {
-            if last_type.is_some() {
+        // Print a "Event $TYPE" header for each event type.
+        // This compares only the event *type*, not the entire event,
+        // so we don't compare variable events for different variables as different.
+        //
+        // This assumes EventDescription::Any is not a valid value for an event to have
+        // - it's marked "unreachable!()" below!
+        if last_type == EventDescription::Any
+            || std::mem::discriminant(&last_type) != std::mem::discriminant(&evt.desc)
```
faho, line 862: This is what [std::mem::discriminant](https://doc.rust-lang.org/std/mem/fn.discriminant.html) is made for.

I don't love using the Any sentinel value, but keeping the Option felt extremely awkward.

Maybe we can store the discriminant value directly?

(again it's for skipping the newline the first go around)

tests/checks/functions.fish
```diff
@@ -9,6 +9,10 @@ end
 functions    --details f1 f2
 #CHECKERR: functions: --details: expected 1 arguments; got 2
 
+# Verify that it still mentions "--details" even if it isn't the last option.
+functions    --details --verbose f1 f2
+#CHECKERR: functions: --details: expected 1 arguments; got 2
```
faho, line 14: This used to print "--verbose: expected"

tests/checks/functions.fish
```diff
@@ -171,3 +175,17 @@ functions --no-details --details t
...
+function term1 --on-signal TERM
+end
+function term2 --on-signal TERM
+end
+function term3 --on-signal TERM
+end
+
+functions --handlers-type signal
+# CHECK: Event signal
+# CHECK: SIGTRAP fish_sigtrap_handler
```
faho, line 188: This used to print a second "Event signal" after the SIGTRAP

fish-rust/src/builtins/functions.rs
henrikhorluck: Once IO is ported, we can bump our MSRV to 1.70 so we can use [`is_terminal`](https://doc.rust-lang.org/stable/std/io/trait.IsTerminal.html). Unless that has some behavior-difference I am not aware of, have not seen the internals of it

faho: The idea of increasing a dependency requirement to replace a single line with a slightly shorter single line sounds unappealing to me.

Especially when that line does:

https://github.com/rust-lang/rust/blob/4d7a80d48697171ed151c4667b74fc7d784c1836/library/std/src/sys/unix/io.rs#L79-L82

The *exact same thing we already do* - excluding the "out_is_redirected" check.

I would simply add a helper method to streams - `streams.out_is_terminal()` that encapsulates this exact check.

I know that projects in the rust ecosystem like to move very fast wrt MSRV, and like to include a lot of dependencies, but that's not to my taste and I consider it to be one of the annoyances of the ecosystem.

faho: 22d92b8f39893588be5d5fff6d7ee3f9abae6af2 should be the answer here.

fish-rust/src/builtins/functions.rs
faho: Turns out it's easier than I thought because reformat_for_screen was already ported.

henrikhorluck (commented):
Mostly nits, some wrong short-options that should have tests, and the event-signal appears to have a sorting difference that should be mentioned in release-notes or reverted, otherwise LGTM

tests/checks/functions.fish
faho: > fix an earlier regression introduced in the Rust code

Yes, the issue was new in the rust version, caused by enum comparisons being deep.

Probably something to look out for - if we replace C++ enums with Rust enums, the C++ version only compares variants.

>2. there is a behavior difference in the sorting, which is probably fine, but should perhaps be mentioned in the release notes?

I do not believe this is a thing that is expected to be stable. This *isn't* sorted alphabetically, it's sorted *by definition order* - and used to be in reverse definition (oldest last) order.

fish-rust/src/builtins/functions.rs
```diff
@@ -0,0 +1,420 @@
...
+            no_metadata: false,
+            verbose: false,
+            handlers: false,
+            handlers_type: WString::new(),
+            description: WString::new(),
+        }
+    }
+}
+
+const NO_METADATA_SHORT: char = 2 as char;
```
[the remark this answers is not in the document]
faho: Oh I quite misunderstood what NONOPTION_CHAR_CODE meant, then?

So it's used to be the char code for the one option that is long-only?

And if we have multiple options that are long-only?

Tbh I prefer naming the constant after the option.

faho:
Okay, `reformat_for_screen` is currently *broken* - it does not seem to ever finish. Running `functions` interactively let's fish spin and isn't even interruptible with ctrl-c.

I'm gonna try to find out where the issue is, if I can't find it I'm gonna remove it from functions for now.

fish-rust/src/builtins/functions.rs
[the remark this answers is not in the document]
henrikhorluck: It appears I might be one misunderstanding how `NONOPTION_CHAR_CODE` is supposed to be used then, I just assumed it should be used instead of `1 as char / '\x01' / '\u{1}'`, which is how I have used it in `builtins/string/{escape,unescape}` and `builtins/path`.

Looking at the [commit](https://github.com/fish-shell/fish-shell/commit/ecfabf4db860ffd7466fd3c75b22e493638b27f1) that introduced it, it appears `'\x01'` has a special meaning as a non-option argument (duh), but only if the ordering is `REQUIRE_ORDER | RETURN_IN_ORDER`, which is set by `+` and `-` in short_opts respectively. 

When the ordering is the default aka `PERMUTE`, `\x01` does not have that special meaning.

My takeaway here is:

1. I dislike `wgetoptr_t`, since it mixes special meanings with legitimate return values, only configured through a char in a string. A minimal way of improving it would be to instead return an enum like:

```rust
enum WGetOptRetVal {
    LongOpt(usize), // where the usize is the "id" of the long-only option
    Opt(char),
    NonOption,
}
```

We could also make `wopt` accept using a literal number, making it result in a different enum-variant, to make this difference in usage much clearer. (This requires a `IntoOption` trait)

2. We should probably avoid using `\x01` (regardless of whether it is as `NONOPTION_CHAR_CODE` or another named constant) as a short-option for long-only arguments to avoid accidentally mixing this usage, in that way your change to 2 instead of 1 as was present in the original code is good.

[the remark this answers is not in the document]
faho: > I dislike wgetoptr_t, since it mixes special meanings with legitimate return values, only configured through a char in a string

Yup, it has some real crusty C heritage.

>A minimal way of improving it would be to instead return an enum like:

This would return a different value for `-D` and `--details`, which is awkward. I don't *think* you could match like

```rust
LongOpt(id) | Opt(char) => {
}
```

even if you called the variables the same?

> We should probably avoid using \x01 (regardless of whether it is as NONOPTION_CHAR_CODE or another named constant) as a short-option for long-only arguments to avoid accidentally mixing this usage

Agreed. Tho note that "PERMUTE" is the overwhelming majority of uses and changing from that to a different ordering is a pretty big change that is never going to happen for most uses.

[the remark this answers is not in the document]
henrikhorluck: > This would return a different value for -D and --details, which is awkward. I don't think you could match like
> `LongOpt(id) | Opt(char)`
> even if you called the variables the same?

Ah, I named it a bit unspecificly, I intended `LongOpt` to be `LongOnlyOpt`, if that makes more sense? And then keep the same behaviour for `Opt` if the option has both long and short variants

But a move towards something like your interpretation would make it trivial to determine whether we got the argument as a short- or long-opt, just with another enum-variant. 

We could also add more variants to move the `optarg` to this enum, avoiding all the `w.optarg.unwrap()`-s.

Issue #8877, opened 2022-04-14
Valid commands following redirection symbols at command line start are colored as $fish_color_command instead of $fish_color_error


schuelermine:
```
fish, version 3.4.1
3.4.1
```

Typing something like `>echo` doesn’t color the `echo` as an error, but it should

schuelermine:
https://asciinema.org/a/487589

neerajram30:
Hi there, is the issue solved ?

schuelermine:
No

neerajram30:
Can I work on solving this issue?

faho:
@neerajram30 It's open and nobody has said they're working on it, so go for it!

pradyotRanjan:
I would like to work on this.

floam:
Sure, go ahead, it was nearly a month ago the other individual said they were interested in solving it.

mikoreal:
I'm trying to solve this, however I have a bit of trouble finding how and in which files it's defined how colors are autosuggested based on validity of the commands. Could anyone provide any info on that?

krobelus:
@mikoreal probably this can be fixed inside `highlighter_t::visit(const ast::decorated_statement_t &stmt)`.
If a redirection's source_range occurs textually before the statement's command, then we should probably color at least the command token as error. The syntax tree nodes are defined in `src/ast.h`.

himmel-rider:
@krobelus  Thanks your guidance. I have submitted the PR for review. Could you please take a look when you have a chance? Thank you again! Link is  https://github.com/fish-shell/fish-shell/pull/9587

99jte:
May be in over my head with it (to my credit, it's still labeled a "good first issue"), but have there been any further thoughts on the options to solve this discussed in https://github.com/fish-shell/fish-shell/pull/9587#issuecomment-1436036722 ?

ridiculousfish:
No. Probably this should be fixed at the AST level: if we encounter an erroneous redirection, then try consuming the next token and if it's a string then pull that into the error range.

PR #9947, opened 2023-08-12
Include the target of bad redirects in the error

99jte:
## Description

When a redirect generates an error, peek at the next token, and if it is a string, consume it, and include it in the error range; the latter is done by adding a method `combine` to `SourceRange` that merges two of them. Test was from @krobelus.

Fixes issue #8877

## TODOs:

### DIFF
--- a/fish-rust/src/ast.rs
+++ b/fish-rust/src/ast.rs
@@ -3192,6 +3192,15 @@ impl<'s> Populator<'s> {
                     }
                 }
             }
+            ParseTokenType::redirection if self.peek_type(0) == ParseTokenType::string => {
+                let next = self.tokens.pop();
+                parse_error_range!(
+                    self,
+                    next.range().combine(tok.range()),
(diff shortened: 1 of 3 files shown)


fish-rust/src/ast.rs
```diff
@@ -3192,6 +3192,15 @@ impl<'s> Populator<'s> {
                     }
                 }
             }
+            ParseTokenType::redirection if self.peek_type(0) == ParseTokenType::string => {
+                let next = self.tokens.pop();
```
krobelus, line 3196: what if the redirection is something like `2>&1`, does that work? We probably don't want to pop the next token in that case

krobelus: ok it's correct, the `1` is actually a separate token

99jte: I'm glad it turned out not to be an issue, but I hadn't actually considered that.

PR #9950, opened 2023-08-13
Use a Cargo workspace

henrikhorluck:
- Fix clippy lint in widestring-suffix
- Use a cargo workspace
- Make CI use the workspace, so we format/check all

See https://doc.rust-lang.org/cargo/reference/workspaces.html for why we should use a workspace, but essentially it boils down to sharing dependencies across crates in one `Cargo.lock` and running `fmt/clippy/check` across crates from root.

I have explicitly chosen to not move the entire of `fish-rust/Cargo.toml` to root, as that would require manually specifying paths for `build.rs` and `lib.rs`, but that could also be done in this change. This is currently focusing on the minimal changes to adopt a workspace


henrikhorluck:
Ugh, I think this actually needs changes to CMake since this removes the patch from fish-rust's Cargo.toml

zanchey:
I think we want this, but only after CMake is gone. We're probably going to need the `cargo xtask` pattern.

henrikhorluck:
> I think we want this, but only after CMake is gone. We're probably going to need the `cargo xtask` pattern.

I think we want to get this first because it 1. fixes the rather annoying fact that you cannot run fmt/etc from root, and 2. allows piecewise migration to the xtask-pattern or however we choose to port the littecheck tests and similar

cmake/Tests.cmake
```diff
@@ -193,16 +193,16 @@ endif()
 if(NOT DEFINED ASAN)
     add_test(
         NAME "cargo-test"
-        COMMAND cargo test ${CARGO_FLAGS} --target-dir target ${cargo_target_opt}
-        WORKING_DIRECTORY "${CMAKE_SOURCE_DIR}/fish-rust"
+        COMMAND cargo test ${CARGO_FLAGS} --package fish-rust --target-dir target ${cargo_target_opt}
+        WORKING_DIRECTORY "${CMAKE_SOURCE_DIR}"
```
henrikhorluck, line 197: In an earlier commit I wrongly stated that this change makes CMake workspace aware. As far as I can tell that is wrong, as `cargo` will correctly identify that a crate is a member of a workspace even if run from within the directory of a crate that does not inherit package-attributes/dependencies from the workspace (likely from the lack of a `Cargo.lock`?, not entirely certain), therefore this change essentially only fixes the `target-dir` path to be `fish-shell/target` instead of `fish-shell/fish-rust/target`

Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+    "fish-rust",
+    "fish-rust/widestring-suffix",
+]
+default-members = ["fish-rust"]
+
+[workspace.package]
+rust-version = "1.67"
+edition = "2021"
+
+# TODO: Move fish-rust to src, make it the root package of this workspace
```
zanchey, line 13: We'll definitely want to do that eventually!

Cargo.lock
```diff
@@ -15,9 +15,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.0.2"
```
krobelus, line 18: the Cargo.lock changes are unrelated right? Would have been better to as a separate commit with then.

henrikhorluck: This combines the two lock files we had, so some of the changes are required, I don't remember if I did that with `cargo update` or `cargo generate-lockfile`, but if the output is breaking in any way that is an issue non-locked dependencies in `Cargo.toml`.

PR #9952, opened 2023-08-13
Add appendln

henrikhorluck:
- Add an appenln method to output_stream_t
- Adopt appendln

- I did remove a couple instances of what I determined "untranslatable" uses of `wgettext`

```diff
- streams.out.append(wgettext_fmt!("%ls\n", f));
+ streams.out.appendln(f);
```

### DIFF
--- a/fish-rust/src/builtins/shared.rs
+++ b/fish-rust/src/builtins/shared.rs
@@ -114,6 +114,12 @@ impl output_stream_t {
         self.ffi().append(&s.as_ref().into_cpp())
     }
 
+    /// Append a &wstr or WString with a newline
+    pub fn appendln(&mut self, s: impl Into<WString>) -> bool {
+        let s = s.into() + L!("\n");
+        self.ffi().append(&s.into_cpp())
+    }
+
     /// Append a char.
     pub fn append1(&mut self, c: char) -> bool {
         self.append(wstr::from_char_slice(&[c]))
--- a/fish-rust/src/builtins/type.rs
+++ b/fish-rust/src/builtins/type.rs
@@ -119,15 +119,13 @@ pub fn r#type(
                     }
                     if opts.path {
                         if let Some(orig_path) = props.copy_definition_file() {
-                            streams.out.append(orig_path);
+                            streams.out.appendln(orig_path);
                         } else {
-                            streams.out.append(path);
+                            streams.out.appendln(path);
                         }
-                        streams.out.append1('\n');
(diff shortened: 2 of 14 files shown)


fish-rust/src/builtins/shared.rs
```diff
@@ -114,6 +114,12 @@ impl output_stream_t {
         self.ffi().append(&s.as_ref().into_cpp())
     }
 
+    /// Append a &wstr or WString with a newline
+    pub fn appendln(&mut self, s: impl Into<WString>) -> bool {
```
henrikhorluck, line 118: I think this trait-bound should ideally be `impl ToWString`, but afaik we only really implement that for some integer-types, and not for `T where T: ToString`, or some other very generic impl.

Note that any type that `impl Display` automatically `impl ToString` per the [docs](https://doc.rust-lang.org/std/fmt/trait.Display.html) .

TBH I'm probably rather of the opinion we should just remove `ToWString` and instead use `.to_string()` and convert to UTF32, unless there is some behaviour-difference I am not aware of

fish-rust/src/builtins/type.rs
```diff
@@ -119,15 +119,13 @@ pub fn r#type(
...
                         } else {
-                            streams.out.append(path);
+                            streams.out.appendln(path);
                         }
-                        streams.out.append1('\n');
                     } else if !opts.short_output {
                         streams.out.append(wgettext_fmt!("%ls is a function", arg));
-                        streams.out.append(wgettext_fmt!(" with definition"));
-                        streams.out.append1('\n');
+                        streams.out.appendln(wgettext!(" with definition"));
```
henrikhorluck, line 128: I find this splitting of `wgettext` really weird, there should be some guidelines against doing that, same with including the trailing `\n` in the text to be translated and properly separating punctuation-differences (as we observed in the builtins/string rewrite)

fish-rust/src/builtins/type.rs
```diff
@@ -146,7 +144,7 @@ pub fn r#type(
                         streams.out.append(wgettext_fmt!(" (%ls)\n", comment));
                     }
                 } else if opts.get_type {
-                    streams.out.append(L!("function\n"));
+                    streams.out.appendln("function");
```
faho, line 147: (these are functional strings that probably shouldn't be translated - just like the scripting language itself isn't translated. we should keep that in mind for any translation guidelines)

faho:
Thanks, this is a good idea!

Issue #751, opened 2013-05-16
Add a way to expand tokens


dag:
Let's say the binding is CTRL-X.

Now we can be more selective: `rm *.png` CTRL-X `rm a.png bc.png foo.png` CTRL-W `rm a.png bc.png`

We can confirm that we typed the right thing before executing: `mkinitrd (uname -r)` CTRL-X `mkinitrd 3.8.11-200.fc18.x86_64`

Or that we typed enough: `kill %fi` CTRL-X `kill 1580 1684 2021 3431 3822` oops I only wanted firefox

Possible problems to solve:
- Evaluation has side-effects (at least for subshells) such as printing to stderr.  Completions have the same problem though.
- Are the problems with `eval` relevant to this?
- The result needs to be properly quoted or escaped so that it works exactly like running the commandline without expanding anything (other than the obvious stuff like `(date)` depending on when it's run).

This idea came out of #731 but I now think they're two separate ideas.

krader1961:
I was going to close this as "will-not-implement" because we most definitely do not want to expand constructs like subcommands. However, expanding wildcards is safe from side-effects and could be useful. So I'm adding the "RFC" (request for comment) label to solicit more feedback on the idea.

zanchey:
I think @faho was working on something along these lines.

faho:
Yep, I have attempted to make an "expand" builtin. The way it currently works is that it'll perform all safe expansions on its arguments, unless an option is given to skip a particular one. So `expand '{$a}*' will perform variable-expansion, brace-expansion and wildcard-expansion, while `expand --no-wildcard '{$a}*'` will not do the wildcard-expansion.

I still have a few open questions:

- Should this have an option to _do_ command substitutions? (Currently it doesn't, and I'm quite sure it should not do them by default)

- Should this already escape the string, or have an easy option to do so?

- How about newlines? A "-z" option to do NUL-delimited output?

- Are there any changes to make it more useful for other usecases? (My initial motivation was to make fuzzy expansion usable by completion scripts, which would require some more work)

krader1961:
It seems to me that it must escape all strings so that when word splitting is done you get the same tokens that would occur from the unexpanded args. This includes escaping newlines and other non-printable chars (e.g., tabs).

[...]

PR #9954, opened 2023-08-14
fix __fish_list_current_token not recognizing ~ as $HOME

Axlefublr:


Axlefublr:
it's not gonna hurt if I elaborate: I'd expect `cd ~/.config` to display the contents of .config, but it doesn't because it doesn't recognize ~ as home, which is incredibly annoying

I will argue that not seeing dotfiles and dotdirectories is also annoying, but I can see how not using the -A flag for the ls is a design choice. However I'm gonna jump up thrice if you agree that it's a good idea to change the ls in there to `ls -A` instead

henrikhorluck:
I cannot reproduce your error, even in `sh -c 'env HOME=$(mktemp -d) XDG_CONFIG_HOME= fish'` (note you probably want to add another file to your temp home's `.config` there, only `fish` exists by default).

Tab-completing with `cd ~/.config/` already correctly displays the contents of `~/.config/` for me on  `master`, and it also works on `3.6.1`. This works regardless of whether `AppleShowAllFiles` is set or not (I am on macOS)

Axlefublr:
what's completely hilarious is that "$HOME/.config" does work, but $HOME/.config doesn't

the former being a situation you would practically never find yourself in

zanchey:
The full right answer here would be #751, probably as a subcommand of `path`.

Axlefublr:
Thank you for merging! 

Was scared that this was going to be a situation of "we can't do it perfectly so we won't do it at all", so it's nice to see the feature being merged at least in this simplified fashion for the time being

PR #9955, opened 2023-08-14
Enable PWD reporting for iTerm2

gpanders:
## Description

iTerm2 supports reporting PWD changes via OSC 7 (and has for quite some time). Remove the `TODO` comment and enable OSC 7 on iTerm2.

I am not sure why this was left as a TODO for so long. I searched the issue tracker, but didn't find anything.

## TODOs:


faho:
Merged, thanks!

> I am not sure why this was left as a TODO for so long. I searched the issue tracker, but didn't find anything.

The simple truth is that we have a bunch of macos-specific issues that we could use help with.

I can't really work on them because I don't have a mac.

Issue #7086, opened 2020-06-06
Deal with escaped characters in autogenerated man completions


ammgws:
I am not familiar with how escaping works in manpages but it seems like some escape single quotes as '\(cq' etc, which gets picked up as-is by the fish completion generation script. Might be nice to see if this could be fixed trivially.

```
.\" -----------------------------------------------------------------
.\" * MAIN CONTENT STARTS HERE *
.\" -----------------------------------------------------------------
.SH "NAME"
pacstrap \- install packages to the specified new root directory
.SH "SYNOPSIS"
.sp
pacstrap [options] root [packages\&...]
.SH "DESCRIPTION"
.sp
pacstrap is designed to create a new system installation from scratch\&. The specified packages will be installed into a given directory after setting up some basic mountpoints\&. By default, the host system\(cqs pacman signing keys and mirrorlist will be used to seed the chroot\&.
.sp
If no packages are specified to be installed, the \fIbase\fR group will be installed\&.
.SH "OPTIONS"
.PP
\fB\-C\fR <config>
.RS 4
Use an alternate config file for pacman\&.
.RE
.PP
\fB\-c\fR
.RS 4
Use the package cache on the host, rather than the target\&.
.RE
.PP
\fB\-G\fR
.RS 4
Avoid copying the host\(cqs pacman keyring to the target\&.
.RE
.PP
\fB\-i\fR
.RS 4
Prompt for package confirmation when needed (run interactively)\&.
```

![image](https://user-images.githubusercontent.com/20397027/83945620-e2889380-a846-11ea-9269-3e07ba91bc28.png)

faho:
Troff is just always worse than you thought.

We already detect "cq" in some capacity, but I'm guessing it's just just `\cq`?

faho:
Okay, deroff.py handles this, but generating the completion doesn't?

faho:
Huh, we have multiple parsers here? I didn't expect that. It's probably down to handling multiple "dialects", and this doesn't detect `\(cq`, possibly because of that weird awful no-good `(` in there (seriously, groff is *awful*).

Anyway, let's just hack this in there, since we haven't had too many complaints, and man parsing is a bit of a dead end anyway.

PR #9961, opened 2023-08-17
Replace more escapes with quotes in man parser

ysthakur:
## Description

Similar to #7086, which is about `\(cq`. At least one man page ([`df`](https://linux.die.net/man/1/df)) uses `\(aq`, so this PR replaces those with single quotes. I haven't found any examples other than `df`, though.

I don't understand roff, but looking at the [GNU Troff manual](https://www.gnu.org/software/groff/manual/groff.html#Character-Classes), `dq`, `lq`, `oq`, and `rq` may also have to be replaced with quotes, in addition to `cq` and `aq`. `dq` is a neutral double quote. `lq` seems to be used for left double quotes and `rq` for right double quotes, but I assume both can safely be replaced with neutral double quotes. I have no idea what `oq` is.

## TODOs:

- [x] Changes to fish usage are reflected in user documentation/manpages. <- Is this necessary?
- [x] <s>User-visible changes noted in CHANGELOG.rst</s> <- This change isn't particularly user-visible, right?
- [x] Try to find man pages that use `dq`, `lq`, `oq`, or `rq` and, if they exist, add them to this PR?

### DIFF
--- a/share/tools/create_manpage_completions.py
+++ b/share/tools/create_manpage_completions.py
@@ -260,8 +260,12 @@ def remove_groff_formatting(data):
     data = data.replace(r"\-", "-")
     data = data.replace(".I", "")
     data = data.replace("\f", "")
+    data = data.replace(r"\(oq", "'")
     data = data.replace(r"\(cq", "'")
     data = data.replace(r"\(aq", "'")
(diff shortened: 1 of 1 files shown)


faho:
Merged, thanks!

Issue #9925, opened 2023-07-29
#9018 fixed, but broke in 3.6.1-714 again


NextAlone:
#9018 fixed but in 3.6.1-714 version, it broke again after upgrading fish-shell through pacman on ArchLinux
```
 status fish-path
/usr/bin/fish (deleted)
```

faho:
cc @krobelus - the rust implementation of `get_executable_path` uses rust's [std::env::current_exe()](https://doc.rust-lang.org/std/env/fn.current_exe.html), which unfortunately seems to just defer to the OS.

NextAlone:
Hi, any progress?

faho:
No. Typically if there is any, we'll note in the issue.

Krobelus is away for a bit.

Honestly, currently I recommend just using 3.6.1 if all you want to do is *use* the thing - it's faster and less buggy than master, and master has few features and improvements over it.

NextAlone:
> No. Typically if there is any, we'll note in the issue.
> 
> Krobelus, who is currently the person best suited to work on it, is away for a bit.
> 
> Honestly, currently I recommend just using 3.6.1 if all you want to do is _use_ the thing - it's faster and less buggy than master, and master has few features and improvements over it.

Thanks for your advice, i'll use 3.6.1 for these days.

Issue #3914, opened 2017-03-26
Treat empty $PATH component as equivalent to "."


faho:
As recently discovered in #3912:

```fish
$ printf '%s\n' '#!/bin/sh' 'echo banana' > banana
$ chmod +x banana
$ set PATH "" $PATH
$ banana
# Command not found
$ set PATH "." $PATH
$ banana
banana
```

As far as I can see, fish just effectively ignores the empty component. The compatible behavior is to treat it like it was ".", i.e. the current directory. This is important because fish inherits these variables from the outside and exports them, so interoperability with POSIX is useful here.

This possibly also applies to $MANPATH, and was recently fixed for $CDPATH (#3901, #2106).

krader1961:
FWIW, I think fish should treat empty path components as equivalent to `.` for compatibility with POSIX shells. The POSIX shell behavior is FUBAR in my opinion. Anyone wanting to include the CWD can do so by putting `.` in the path (which is what I do). Nonetheless, this discrepancy is something that only makes it harder for people to switch to fish shell and does not have a tangible benefit that I can find.j

faho:
>Nonetheless, this discrepancy is something that only makes it harder for people to switch to fish shell and does not have a tangible benefit that I can find.

One benefit is that empty components make it harder to iterate over these lists.

E.g.

```fish
for command in $PATH/*
    test -x $command -a ! -d $command; and echo $command
end
```

would list all executable commands iff there was no empty $PATH component that is treated specially. And if we allow those, it's hard to miss.

(This was obviously found in #3912 - see how I complicated the implementation for empty $PATH components)

Now, the problem is that we pretty much _can't_ get by just ignoring empty $PATH components because we can inherit them from the outside. It might be possible to convert them (to ".") when importing instead.

However, whatever we do, it should be consistent between $PATH, $CDPATH (which currently allows empty components) and $MANPATH.

krader1961:
My vote is to magically munge all three vars when set/imported to translate empty elements to `.`. That way scripts don't have to do anything special and constructs like `for command in $PATH/a_command` work sensibly without having to do anything special. This is trivial to implement by augmenting `react_to_variable_change()` to special-case those three env vars like we do for curses and locale env vars.

krader1961:
[...]

Issue #2784, opened 2016-03-01
Use first environment variable, not last, when there are duplicates


ridiculousfish:
If duplicate environment variables appear in envp, fish currently uses the value implied by the last one, while getenv returns the first one. "This could result in an ambiguous environment causing environment variables to be propagated to subprocesses, despite the protections supposedly offered by taint checking."

fish should follow Perl and others in using the first, not last, environment variable.

Related CVE-2016-2381,  https://lists.debian.org/debian-security-announce/2016/msg00072.html

zanchey:
:clap: I don't think there's necessarily a vulnerability in fish here but this is a good fix.

PR #9962, opened 2023-08-18
Port fish.cpp to Rust

henrikhorluck:
This is based on https://github.com/fish-shell/fish-shell/pull/9950 to have `CARGO_MANIFEST_DIR` == `CMAKE_SOURCE_DIR`

- Fix clippy lint in widestring-suffix
- Use a cargo workspace
- Make CI use the workspace, so we format/check all
- Use the workspace from CMake
- Specify default-members as `fish-rust`
- Move edition and MSRV to workspace

Changes in this PR are:

- Make CARGO_MANIFEST_DIR for fish-rust equal to src
- Forward CMake configs to Cargo
- Make PROGRAM_NAME safe
- Port history::start_private_mode to Rust
- Make ParsedSource::new pub
- Make methods in path not unnecessairly take &mut
- Port MISSING_HELP and MISSING (arg)
- Add various FFI-interop-functions
- Make ConfigPaths store as `PathBuf`
- Port more of env_init to Rust
- Port and adopt main written in Rust

Main now uses `wgetoptr_t` instead of `getopt`.

Typing-wise I attempted to prefer `PathBuf` over `WString` since we are operating on paths.

I encountered a couple issues: `fwprintf!` appears to not do anything? `FLOGF!` does not format


fish-rust/build.rs
```diff
@@ -103,6 +126,25 @@ fn main() {
     rsconf::rebuild_if_paths_changed(&source_files);
 }
 
+fn find_version() -> Result<String, Box<dyn std::error::Error>> {
```
faho, line 129: This appears to me to be what https://crates.io/crates/git-version says it does.

It just runs `git` itself, which I would find preferable over depending on the entirety of the git reimplementation that is libgit2.

faho: That's also basically [what we do now](https://github.com/fish-shell/fish-shell/blob/a29aa44183582d268ee2952e4eea050d0c96e1d1/build_tools/git_version_gen.sh)

henrikhorluck: We could probably also just use the script directly (with `--stdout`). I don't know how the envisioned full replacement of CMake with Cargo will look like, and wanted this change to be somewhat independent of that

Cargo.toml
```diff
@@ -1,11 +1,42 @@
...
+cc = { git = "https://github.com/mqudsi/cc-rs", branch = "fish" }
+
+
+[profile.release]
+overflow-checks = true
+
 [package]
 name = "fish-rust"
 version = "0.1.0"
-edition = "2021"
```
krobelus, line 4: what's the selling point? I like repo-relative paths but that's not really an important consideration

Cargo.toml
```diff
@@ -35,10 +66,12 @@
 cc = { git = "https://github.com/mqudsi/cc-rs", branch = "fish" }
 cxx-build = { git = "https://github.com/fish-shell/cxx", branch = "fish" }
 cxx-gen = { git = "https://github.com/fish-shell/cxx", branch = "fish" }
+git-version = "0.3"
```
krobelus, line 69: could you squash this commit?

fish-rust/src/common.rs
```diff
@@ -1240,7 +1240,10 @@ pub const TESTS_PROGRAM_NAME: &wstr = "(ignore)"L;
 /// like `debug()`. It is only intended to suppress diagnostic noise from testing things like the
 /// fish parser where we expect a lot of diagnostic messages due to testing error conditions.
 pub fn should_suppress_stderr_for_tests() -> bool {
-    unsafe { !PROGRAM_NAME.is_empty() && *PROGRAM_NAME != TESTS_PROGRAM_NAME }
+    PROGRAM_NAME
+        .get()
+        .map(|p| p == TESTS_PROGRAM_NAME)
+        .unwrap_or_default()
```
krobelus, line 1246: right, the old comparison incorrectly used `!=` instead of `==`

fish-rust/src/history.rs
```diff
@@ -0,0 +1,8 @@
+use crate::env::{EnvMode, EnvStack};
+use crate::wchar::prelude::*;
+
+/// Sets private mode on. Once in private mode, it cannot be turned off.
+pub fn start_private_mode(vars: &EnvStack) {
```
krobelus, line 5: should copy the docstring from the header file too

src/parse_util.cpp
```diff
@@ -1060,6 +1060,11 @@ static bool detect_errors_in_block_redirection_list(
     return false;
 }
 
+parser_test_error_bits_t parse_util_detect_errors_ffi(const ast::ast_t* ast, const wcstring &buff_src,
```
krobelus, line 1063: nit: I'm surprised this is needed, the version that passes a const reference should work fine.

fish-rust/src/env/environment.rs
```diff
@@ -371,8 +379,315 @@ lazy_static! {
...
+    } else {
+        None
+    }
+}
+
+fn setup_user(vars: &EnvStack) {
+    let uid = unsafe { libc::geteuid() };
+    let user_var = vars.get_unless_empty(L!("USER"));
+
+    let mut userinfo: MaybeUninit<libc::passwd> = MaybeUninit::zeroed();
```
krobelus, line 417: interesting. I've been using `unsafe { std::mem::zeroed() }`, we should make things consistent in future.
MaybeUninit seems superior because it has `.as_mut_ptr()` etc.
Any particular reason not to use `MaybeUninit::uninit()`? That might allow tools to find bugs.

fish-rust/src/env/environment.rs
```diff
@@ -371,8 +379,315 @@
...
+            None
+        } else {
+            Some(res)
+        }
+    } else {
+        None
+    }
+}
+
+fn setup_user(vars: &EnvStack) {
```
krobelus, line 413: nit: missing docstring

fish-rust/src/env/environment.rs
```diff
@@ -371,8 +379,315 @@
...
+                buf.as_mut_ptr(),
+                buf.len(),
+                &mut result,
+            )
+        };
+        if retval == 0 && !result.is_null() {
+            let userinfo = unsafe { userinfo.assume_init() };
+            if unsafe { *result }.pw_uid == uid {
+                if vars.get_unless_empty(L!("HOME")).is_none() {
+                    if !userinfo.pw_dir.is_null() {
```
krobelus, line 438: missing this comment I guess:

	// The uid matches but we still might need to set $HOME.

fish-rust/src/env/environment.rs
```diff
@@ -371,8 +379,315 @@
...
+            }
+        }
+    } else if vars.get_unless_empty(L!("HOME")).is_none() {
+        // If $USER is empty as well (which we tried to set above), we can't get $HOME.
+        vars.set_empty(L!("HOME"), EnvMode::GLOBAL | EnvMode::EXPORT);
+    }
+}
+
+/*
+fn setup_path() {
```
krobelus, line 493: probably better to remove commented code

fish-rust/src/env/environment.rs
```diff
@@ -371,8 +379,315 @@
...
+    let vars = &PRINCIPAL_STACK;
+
+    let env_iter: Vec<_> = std::env::vars_os()
+        .map(|(k, v)| (str2wcstring(k.as_bytes()), str2wcstring(v.as_bytes())))
+        .collect();
+    // Import environment variables. Walk backwards so that the first one out of any duplicates wins
+    // (See issue #2784).
+    // PORTING: this should behave like in C++ in regards to the above comment, but that might be
+    // with assumptions on the standard library's behavior.
+    for (key, val) in env_iter.into_iter().rev() {
```
krobelus, line 517: right. Maybe we should call `environ()` ourselves

fish-rust/src/env/environment.rs
```diff
@@ -371,8 +379,315 @@
...
+        if val.is_empty() {
+            // PORTING: This is making assumptions that env::vars_os set val to empty if no equal sign
+            // PORTING: That assumption appears to be wrong https://github.com/rust-lang/rust/blob/2ceed0b6cb9e9866225d7cfcfcbb4a62db047163/library/std/src/sys/unix/os.rs#L584C30-L584C30
+            // it appears they allow names starting with =, but do not turn malformed lines
+            // into the variable name with an empty value
```
krobelus, line 522: I think the behavior change of skipping env lines that don't contain `=`
is fine because those are invalid and I don't think any language that
produces them.
However, we must not treat zero-length strings (e.g. `FOO=`) as *empty list* (which is what `set_empty()` does).

src/env.cpp
```diff
@@ -38,14 +38,6 @@
...
-/// Some configuration path environment variables.
-#define FISH_DATADIR_VAR L"__fish_data_dir"
-#define FISH_SYSCONFDIR_VAR L"__fish_sysconf_dir"
-#define FISH_HELPDIR_VAR L"__fish_help_dir"
-#define FISH_BIN_DIR L"__fish_bin_dir"
-#define FISH_CONFIG_DIR L"__fish_config_dir"
-#define FISH_USER_DATA_DIR L"__fish_user_data_dir"
-
 /// At init, we read all the environment variables from this array.
 extern char **environ;
```
krobelus, line 43: yeah `key` makes no sense, it should be `key_and_val`. Testament to how this code was dead in practice and we are better of skipping such entries

faho: Note: This code is pretty new, and the "inheriteds" mechanism is there for one reason: To have `set --show foo` show `$foo: originally inherited as |banana|`. See #9029.

fish-rust/src/fish.rs
```diff
@@ -0,0 +1,877 @@
...
+use std::env;
+use std::ffi::{CString, OsStr, OsString};
+use std::fs::File;
+use std::io;
+use std::mem::MaybeUninit;
+use std::os::unix::prelude::*;
+use std::path::{Path, PathBuf};
+use std::sync::atomic::Ordering;
+use std::sync::Arc;
+
```
krobelus, line 59: > 3.6.1-99999-dirty

Ideally the commit that fixes this comes before this one (to minimize eyebrow-raising)

fish-rust/src/fish.rs
```diff
@@ -0,0 +1,877 @@
...
+// FIXME: when the crate is actually called fish and not fish-rust, read this from cargo
+// See: https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates
+// for reference
+const PACKAGE_NAME: &str = "fish"; // env!("CARGO_PKG_NAME");
+const DOC_DIR: &str = {
+    match option_env!("DOCDIR") {
+        Some(e) => e,
+        None => "share/doc/fish",
```
krobelus, line 67: important to note that for any "real" build DOCDIR will be defined and will not need a default.
However we will probably need it for things like `cargo test` and `rust-analyzer`, it least until we move this to `build.rs`
The definition of `DEFAULT_PATH` has the same issue, see its TODO comment.
Can we clarify that these exist purely to facilitate building by using a value like `"(unused)"` everywhere?

fish-rust/src/fish.rs
```diff
@@ -0,0 +1,877 @@
...
+
+fn print_rusage_self(mut dest: impl io::Write) {
+    let mut rs = MaybeUninit::zeroed();
+    if unsafe { libc::getrusage(libc::RUSAGE_SELF, rs.as_mut_ptr()) } != 0 {
+        let s = CString::new("getrusage").unwrap();
+        unsafe { libc::perror(s.as_ptr()) }
+        return;
+    }
+    let rs: libc::rusage = unsafe { rs.assume_init() };
+    let rss_kb = if cfg!(target_vendor = "apple") {
```
krobelus, line 140: I'm pretty sure that we can use `target_os = "macos"` which is more consistent with the other Rust

fish-rust/src/fish.rs
```diff
@@ -0,0 +1,877 @@
...
+    let _ = writeln!(dest, "        max rss: {} kb", rss_kb);
+    let _ = writeln!(dest, "        signals: {}", rs.ru_nsignals);
+}
+
+fn determine_config_directory_paths(argv0: &wstr) -> ConfigPaths {
+    // PORTING: why is this not just an associated method on ConfigPaths?
+
+    let mut paths = ConfigPaths::default();
+    let mut done = false;
+    // FIXME: osstr
```
krobelus, line 165: yeah, `get_executable_path()` should probably take `&[u8]` or similar instead of `&str`.
If it helps, let's make it return `Vec<u8>`. We're gonna need to get rid of `std::env::current_exec()` see https://github.com/fish-shell/fish-shell/issues/9925

fish-rust/src/fish.rs
```diff
@@ -0,0 +1,877 @@
...
+    let exec_path = get_executable_path(&argv0.to_string());
+    if let Ok(exec_path) = exec_path.canonicalize() {
+        FLOGF!(
+            config,
+            format!("exec_path: {:?}, argv[0]: {:?}", exec_path, argv0)
```
krobelus, line 170: should remove the separate format call -- FLOGF already does that (soon) --
unless we really need the Debug representations? Is Display not enough?

fish-rust/src/fish.rs
```diff
@@ -0,0 +1,877 @@
...
+                );
+                Some(just_a_fish)
+            } else {
+                None
+            };
+
+            if let Some(suffix) = suffix {
+                let seems_installed = suffix == installed_suffix;
+
+                let base_path: PathBuf = {
```
krobelus, line 210: nit: I'm not sure if there is really much value in using `PathBuf` here. Possibly better to use `Vec<u8>` for easier resizing and convert only after that, like before

fish-rust/src/fish.rs
```diff
@@ -0,0 +1,877 @@
...
+                    }
+                    // remove whatever remains in remaining_suffix
+                    base.pop();
+                    base
+                };
+
+                paths = if seems_installed {
+                    ConfigPaths {
+                        data: base_path.join("share/fish"),
+                        sysconf: base_path.join("etc"),
```
krobelus, line 226: this should be `/etc/fish` (if it's installed)

fish-rust/src/fish.rs
```diff
@@ -0,0 +1,877 @@
...
+    let escaped_pathname = escape_string(dir, EscapeStringStyle::default()) + L!("/config.fish");
+    if waccess(&config_pathname, libc::R_OK) != 0 {
+        FLOGF!(
+            config,
+            "not sourcing %ls (not readable or does not exist)",
+            escaped_pathname
+        );
+        return;
+    }
+    FLOGF!(config, "sourcing", escaped_pathname);
```
krobelus, line 295: Bad format string here.

fish-rust/src/fish.rs
```diff
@@ -0,0 +1,877 @@
...
+                name_width += 2;
+                for cat in cats.iter() {
+                    let desc = wgettext_str(cat.description);
+                    // https://doc.rust-lang.org/std/fmt/#syntax
+                    // this is left-justified
+                    println!("{:<width$} {}", cat.name, desc, width = name_width);
+                }
+                std::process::exit(0);
+            }
+            'p' => opts.profile_output = Some(OsString::from_vec(wcs2string(w.woptarg.unwrap()))),
```
krobelus, line 458: nit: comment is missing for 'p' case

fish-rust/src/fish.rs
```diff
@@ -0,0 +1,877 @@
...
+                // With "--profile-startup" we immediately turn profiling on.
+                opts.profile_startup_output =
+                    Some(OsString::from_vec(wcs2string(w.woptarg.unwrap())));
+                PROFILING_ACTIVE.store(true);
+                ffi::set_profiling_active(true);
+            }
+            'P' => opts.enable_private_mode = true,
+            'v' => {
+                // FIXME: this was _(L"%s, version %s\n"), but rust-fwprintf! takes a literal instead of an expr
+                // and appears to not print anything
```
krobelus, line 469: right so maybe that's why fwprintf doesn't do anything. You have my blessing to change `$format:literal` to `$format:expr` to support translated format strings. Either way, using `println` and `eprintln` is fine

fish-rust/src/fish.rs
```diff
@@ -0,0 +1,877 @@
...
+    let mut args_for_opts: Vec<&wstr> = owning_args.iter().map(WString::as_utfstr).collect();
+
+    let mut opts = FishCmdOpts::default();
+    my_optind = fish_parse_opt(&mut args_for_opts, &mut opts);
+
+    let args = args_for_opts;
+
+    // Direct any debug output right away.
+    // --debug-output takes precedence, otherwise $FISH_DEBUG_OUTPUT is used.
+    // PORTING: this is a slight difference from C++, we now skip reading the env var if the argument is an empty string
```
krobelus, line 571: the behavior change seems acceptable

fish-rust/src/fish.rs
```diff
@@ -0,0 +1,877 @@
...
+                );
+                let rel_filename = &args[my_optind - 1];
+                // PORTING: this used to be `scoped_push`
+                let old_filename = parser.pin().current_filename_ffi().from_ffi();
+                parser.pin().set_filename_ffi(rel_filename.to_ffi());
+                res = ffi::reader_read_ffi(parser.pin(), c_int(f.as_raw_fd())).into();
+                if res != 0 {
+                    FLOGF!(
+                        warning,
+                        wgettext_fmt!("Error while reading file %ls\n", path.to_string_lossy())
```
krobelus, line 780: do not pass user input as format string to `FLOGF`. Let's use `wgettext!` instead of `wgettext_fmt` and let `FLOGF` do the formatting

Cargo.toml
```diff
@@ -35,10 +66,12 @@ autocxx-build = "0.23.1"
...
 rsconf = { git = "https://github.com/mqudsi/rsconf", branch = "master" }
```
krobelus, line 71: not sure why we'd need shallow clones

krobelus (approved):

krobelus:
I ran `ninja test` 10 times and could not reproduce a failure.

fish-rust/src/fish.rs
faho: Should probably just be `FLOG!` - which joins the arguments.

fish-rust/src/fish.rs
henrikhorluck: I would rather make an extension-method  to `PathBuf` that we can test than change this type. If something is a path using `PathBuf` is the right type. `PathBuf` does have [`shrink_to`](https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.shrink_to), so we can actually perform the resizing, but finding the correct length is not readily available, since the Cpp-code used `strlen`, and a mix of `WString` and `std::string` here, I think this is preferable.

fish-rust/src/fish.rs
henrikhorluck: Good catch! Will fix

fish-rust/src/history.rs
henrikhorluck: Will fix

Cargo.toml
henrikhorluck: CI uses shallow clones

src/parse_util.cpp
henrikhorluck: I really don't feel like fighting with autocxx over this, maybe there is a way to make it work without, I don't know autocxx/C++ well enough.

Cargo.toml
krobelus: I look at individual commits so I totally didn't mind this one

fish-rust/src/env/environment.rs
krobelus: `struct passwd userinfo;` was not zero initialized, it was uninitialized.

src/parse_util.cpp
krobelus: yeah no need to bother for temporary code, it's a nit

fish-rust/src/fish.rs
henrikhorluck: Okay I would like some clarification here:

As far as I understand it `fwprintf` in C++ _outputs_ in UTF32, my understanding of fish's UTF32-usage was that it is supposed to be an internal detail? Therefore changing it to `println!/eprintln!` makes it now print in UTF8. Is that a fine behaviour-change? As far as I can tell `fwprintf!` in Rust has only been used in `io.rs`, which I think is unadopted, so it might actually be non-working?

fish-rust/src/env/environment.rs
henrikhorluck: If it is not already clear: I am not very familiar with C++

In that case I will change it to `uninit`

fish-rust/src/env/environment.rs
krobelus: > I will introduce that as a new commit that can be squashed later since

In general, I prefer squashing eagerly, to incrementally move towards the ideal patch series instead of a big-bang approach.
Also this has the advantage that it helps any reader who doesn't have a good recollection of the previous state (so new revewers but also often the author and previous readers).
If needed I can still see what changed with `git range-diff` etc.

> confstr is available in libc on macOS https://github.com/rust-lang/libc/pull/2883, but we would need to make a PR to add it for other platforms

let's just preserve existing behavior for now, we can put it into `compat.c`

fish-rust/src/fish.rs
krobelus: yeah I think DEFAULT_PATH as fallback in `setup_path` would make sense. Probably has no practical impliciations but it's a simplification

fish-rust/src/fish.rs
krobelus: yeah if we ever need that we can revisit all of the target_os directives

fish-rust/src/fish.rs
krobelus: it's fine to keep the type. I'm surprised though why it's hard to compute `strlen`, it should be the byte length of our UTF-8 string

fish-rust/src/fish.rs
krobelus: yeah C++ uses wide output functions, while in Rust we always convert to bytes before printing.
I'm not sure what's the practical implications of this but I think I had cases where a mixture of the two would lead to output being dropped.
Maybe this is the case here as well? Not sure. If it causes problems, maybe replace it with a todo!, we'll probably not need it that much

fish-rust/src/fish.rs
faho: >The Cpp code here used std::string with as far as I can understand potentially invalid UTF8.

`std::string` is basically a resizable array/vector of arbitrary bytes, yes.

C++ basically... punts on encoding almost entirely.

fish-rust/src/fish.rs
faho: > As far as I understand it fwprintf in C++ outputs in UTF32,

It will output in the locale's encoding, which is typically UTF-8.

It will convert ("narrow") the arguments, or [in C++-speak](https://cplusplus.com/reference/cwchar/wprintf/):

"The external representation of wide characters in stdout are multibyte characters: These are obtained as if [wcrtomb](https://cplusplus.com/wcrtomb) was called to convert each wide character (using the stream's internal [mbstate_t](https://cplusplus.com/mbstate_t) object)."

Of course this will have awkward results if one of fish's internal private use codepoints gets out.

> I think I had cases where a mixture of the two would lead to output being dropped.

The case I know was that we mixed fwprintf (converting output) and fprintf ("direct" output), and with buffering turned on this would break, presumably because the two need separate buffers or need to have them flushed or something?

But it's important to know that `fwprintf(stderr, L"ä");` results in the bytes 0xc3 and 0xa4 (provided the locale is utf-8).

fish-rust/src/fish.rs
henrikhorluck: Yeah, that would have been a huge issue in C. Will fix.

It is very strange to have a literal format-specifier just be passed directly through `gettext`. This should be warned against in the some guidelines, like other weird things https://github.com/fish-shell/fish-shell/pull/9952#discussion_r1293354154, and this is a non-issue with `format!()` since it does not support runtime-specified-format templates

fish-rust/src/fish.rs
faho: Just to confirm, I have tried

```c++
std::fwprintf(stderr, L"%ls\n", L"\u00E4");
```

which results in output of 0xC3 0xA4 0x0A - which is utf-8 for `ä` (and a newline).

fish-rust/src/history.rs
henrikhorluck: Wait: which docstring do you mean here? This already has the doc from the header-file for `start_private_mode`:

https://github.com/fish-shell/fish-shell/blob/0f19d7118b10ee4354317888963c7291334e0173/src/history.h#L345

krobelus: yeah looks like it was added in a later commit (603c5b76e (Remove C++ version of `start_private_mode`, 2023-08-18))

henrikhorluck: Ah, my bad, will move that change to the correct commit

fish-rust/src/fish.rs
henrikhorluck: Sorry, I commented too hastily. It is much easier to find the length than I thought (osstr has it). I will replace this with something closer to the original.

Also: [`nix`](https://docs.rs/nix/latest/nix/trait.NixPath.html) does supply it for `PathBuf` and `Path`, I don't think we need to use that

fish-rust/src/env/environment.rs
henrikhorluck: Okay after some investigation, I have found the following:

`fish` threats `PATH=""` as equal to `PATH="."`, see https://github.com/fish-shell/fish-shell/issues/3914 and 

https://github.com/fish-shell/fish-shell/blob/0f19d7118b10ee4354317888963c7291334e0173/fish-rust/src/env/environment.rs#L171

so this is only reachable if you specifically _unset_ `PATH`. That makes the use of `get_unless_empty` misleading, since it is never empty. 

Is seems like this is intended to actually threat the unset and empty string in the same way?

Adding `panic!("the disco");` to within the `if path.is_none()` does not cause the tests to fail, so this is not covered by our tests. 

How to test manually:
```fish
set PATH_TO_FISH_EXE "$PWD/build/fish"
sh -c "env -u PATH $PATH_TO_FISH_EXE"
```

As far as I can tell the `strconf` was introduced in https://github.com/fish-shell/fish-shell/commit/290d07a833d3c55654e38a69934e3e022bd968fa

henrikhorluck: Okay I squashed the fallback into the wrong commit (will fix), but I have now added a fallback using `compat.c`

krobelus: > That makes the use of get_unless_empty misleading, since it is never empty.
>
> Is seems like this is intended to actually treat the unset and empty string in the same way?

yeah, we often want to do that

we can add a test like

```
env -u PATH $fish -c 'echo $PATH'
# CHECK: ???
```

but it's not clear what to check for so maybe leave it

fish-rust/src/fish.rs
henrikhorluck: I think I found my confusion: I was looking at [fwprintf](https://en.cppreference.com/w/c/io/fwprintf)[^1] then [std_streams](https://en.cppreference.com/w/c/io/std_streams)[^2], which lead me to [FILE](https://en.cppreference.com/w/c/io/FILE)

Which states:
> Narrow and wide orientation
A newly opened stream has no orientation. The first call to fwide or to any I/O function establishes the orientation: a wide I/O function makes the stream wide-oriented; a narrow I/O function makes the stream narrow-oriented. 

Which I then understood as `fwprintf` being a wide I/O function on `STDOUT`, and thus setting it to be wide-oriented **and thus outputting in wide**. But, I think my understanding of a `FILE`'s orientation is wrong, since that appears to only deal with what you send to it, not what it outputs, but cppreference.com does not make that distinction very clear.

It later states (in the middle of the same paragraph)

> Wide I/O functions convert between wide and multibyte characters as if by calling [std::mbrtowc](https://en.cppreference.com/w/cpp/string/multibyte/mbrtowc) or [std::wcrtomb](https://en.cppreference.com/w/cpp/string/multibyte/wcrtomb) with the conversion state as described by the stream.

Which I think is equivalent what cplusplus.com says about external representation, and is what I skipped.

For later: Is cppreference.com not a good place to look for C++ documentation? Should I prefer cplusplus.com?

It appears I mistakenly was on the C instead of C++ side of the documentation, but I don't think that should matter

[^1]: Which says nothing about the orientation/encoding/width
[^2]: Which says stdin/stderr is implicitly unoriented, whatever that means, turns out to be narrow/wide

henrikhorluck: With that said: can I ask for a blessing to just use `println!/eprintln!` here? Using this: 

```rust
fwprintf!(libc::STDOUT_FILENO, "%s, version %s", PACKAGE_NAME, crate::BUILD_VERSION);
```

Prints nothing, and I don't want to debug `fwprintf!`. If we want to keep printf-style formatting I would rather do `println!("{}", sprintf!(...))`

krobelus: cppreference.com is the most useful one.
Yeah it's fine to use println.
It's weird because the Rust fwprintf does narrow IO. Maybe it fails to set the orientation or something.

fish-rust/src/env/environment.rs
henrikhorluck: I added this:
```
# FIXME: Rationalize behavior when PATH is explicitly unset, should this not behave like PATH=""?
# "" is threated like ".", see https://github.com/fish-shell/fish-shell/issues/3914
env -u PATH $fish --no-config -c 'contains -- /usr/bin $PATH; and contains -- /bin $PATH'
and echo PATH has something when unset
# CHECK: PATH has something when unset
```

src/env.cpp
henrikhorluck: I am going to not fix this now — `inheriteds` will not be the same on Rust vs Cpp due to the change in how we read env-vars (Rust allows starting with `=`, that turns into an empty name in our case, which is malformed).
I think that difference is is fine until `builtins/set` is ported

fish-rust/src/env/environment.rs
krobelus: the default path might technically be platform dependent (unless POSIX specifies this?) so I'm not sure if we can write a meaningful test

fish-rust/build.rs
henrikhorluck: I ended up just using the script directly

fish-rust/src/fish.rs
henrikhorluck: This does print it with quotes now:
`config: exec_path: "/ABSOLUTE_PATH/fish-shell/build/fish", argv[0]: "./build/fish"`
We probably want to specify when we want to print it in debug vs lossy, but since this is meant for debugging I think Debug is appropriate.

The error/warning-ones uses lossy currently

henrikhorluck:
I can no longer reproduce the failure in `test_convert_ascii` locally, I still have no clue what caused it, this is a CI run that included the error: https://github.com/henrikhorluck/fish-shell/actions/runs/5905056592/job/16018406020

Seeing as it can no longer be reproduced I see no reason to block merging on that

fish-rust/src/env/environment.rs
henrikhorluck: It should be equal to `getconf PATH`, with that info we can actually test this

krobelus:
I'll merge tomorrow. We're getting close, only the reader and its dependencies are missing

fish-rust/build.rs
henrikhorluck: Contra-message: To avoid change-detection issues, I re-added `git_version`, see https://github.com/fish-shell/fish-shell/pull/9962/commits/0ea11f5e7d5ae531cbc04e5f2846ba8145c49827

tests/checks/default-setup-path.fish
```diff
@@ -0,0 +1,12 @@
+#RUN: %fish -C 'set -g fish %fish' %s
+
+# FIXME: Rationalize behavior when PATH is explicitly unset, should this not behave like PATH=""?
+# "" is threated like ".", see https://github.com/fish-shell/fish-shell/issues/3914
+if command -q getconf
+    env -u PATH $fish -c 'test "$PATH" = "$('(command -s getconf)' PATH)"; and echo Success'
+else
+    # this is DEFAULT_PATH
+    # This techinically depends on PREFIX set in CMake
+    env -u PATH $fish -c 'test "$PATH" = "/usr/local/bin:/usr/bin:/bin"; and echo Success'
```
krobelus, line 10: I think we should make this test work with any prefix, for example by testing only `$PATH[2 3]`

henrikhorluck: I used `[2..]` to ensure it checks all elements

fish-rust/src/fish.rs
```diff
@@ -0,0 +1,875 @@
...
+    let mut rs = MaybeUninit::uninit();
+    if unsafe { libc::getrusage(libc::RUSAGE_SELF, rs.as_mut_ptr()) } != 0 {
+        let s = CString::new("getrusage").unwrap();
+        unsafe { libc::perror(s.as_ptr()) }
+        return;
+    }
+    let rs: libc::rusage = unsafe { rs.assume_init() };
+    let rss_kb = if cfg!(target_os = "macos") {
+        // mac use bytes.
+        rs.ru_maxrss / 1024
```
henrikhorluck, line 143: https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/getrusage.2.html

This says `ru_maxrss` is in kB already. I will investigate which version that changed.

henrikhorluck: It is due to it being in kB on iOS and in bytes on macOS, my local man-page says bytes...

fish-rust/src/fish.rs
henrikhorluck: This would actually have introduced a bug for our (potential) iOS users, since it is kB on iOS https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/getrusage.2.html

henrikhorluck:
The `test_convert_ascii` flakiness has returned (in the CI in my fork) https://github.com/henrikhorluck/fish-shell/actions/runs/5980653829/job/16227194870#step:7:705

tests/checks/default-setup-path.fish
```diff
@@ -0,0 +1,12 @@
+#RUN: %fish -C 'set -g fish %fish' %s
+
+# FIXME: Rationalize behavior when PATH is explicitly unset, should this not behave like PATH=""?
```
henrikhorluck, line 3: Answering https://github.com/fish-shell/fish-shell/pull/9990#discussion_r1311176549 by @faho 

> It is very unlikely that that's what people would expect, and a $PATH of "." is basically unusable. If that's explicitly set, sure, whatever - but an unset $PATH can also be caused by misconfiguration or e.g. a thing that just clears the environment.
> 
> $PATH is also something where we should act like other things because it's a system-wide config knob, and other tools also fall back to their default paths - try env -u PATH $shell -c 'echo $PATH', which will print some default for bash, dash and zsh.

IMO: This discussion does not need to be answered now, it is more important to finish the port. I don't think this should block merging this PR, and I don't know why it hasn't been merged yet. If there are any unresolved questions regarding my changes post the initial review I would love to know.

I can open a separate issue and update the comment to link to that. Is that a satisfactionary interim solution?

I think it would make sense to mark the comment I am answering as off-topic, as the change would be introduced by this PR, and not the one that was commented on.

---

Would love to hear if you have a specific example of a situation where you would (un)intentionally unset `PATH`. I am not familiar enough with this situation to determine what is the expected behaviour — hence the comment. `ksh` and `tcsh` appears to not set a default `PATH`, not that those are very relevant.

If this only happens due to misconfigurations, or using fish in non-recommended situations, I think it would be in the spirit of the `fi` in `fish` to still have a default but also print a warning message.

If unsetting `PATH` is done as a way to reset the environment, shouldn't we then nudge the user in the direction of `--no-config` + `PATH=(getconf PATH)`?

faho: >IMO: This discussion does not need to be answered now, it is more important to finish the port. I don't think this should block merging this PR

Just to be clear: It doesn't block the PR. It's a note about a misunderstanding I've seen.

I'm not sure why this isn't merged, I think @krobelus, who intended to merge, was just busy.

Or possibly he's just merged it to riir-parser, which I feel woefully unqualified to do anything about. Truth be told I would not have expected it to get that big or be that long-lived.

-------

>If unsetting PATH is done as a way to reset the environment, shouldn't we then nudge the user in the direction of --no-config + PATH=(getconf PATH)?

Not something that people do for a specific shell. You might have a process unset $PATH that's three processes up the chain and doesn't know anything about fish.

Fundamentally the PATH="" case is a consequence of what an empty PATH *component* means, which is the current directory - if you `set PATH "" /usr/bin`, that is exported as `PATH=:/usr/bin`, and any empty component is understood to be the current directory. It's not a "default" in any sense, it's just an alternative way to spell `PATH="."`.

It would be a bad default for an unset PATH with very little use. And since the convention is to use some "proper" default, and we already do, I don't think that a change is helpful.

>If this only happens due to misconfigurations, or using fish in non-recommended situations, I think it would be in the spirit of the fi in fish to still have a default but also print a warning message.

We have, over time, moved away from printing a lot of warnings. We used to print a warning in case someone added a $PATH component that didn't exist. Only it turns out that people like adding $PATH components unconditionally (this was before fish_add_path), and people like using ssh and scp - and scp (the old-school version, don't know about the new sftp-based thing) will launch an actual shell and if that prints anything will abort the transfer.

It would even be possible that some ssh implementation unsets PATH and expects the shell to come up with a good default. We would be throwing quite annoying warnings in that case that our users couldn't do anything about.

faho:
Alright I've merged it now, thank you very much!

fish-rust/src/env/environment.rs
krobelus: probably we should keep walking environ backwards, unless we know that there is no security issue here

henrikhorluck: This is walking backwards, it calls `.rev()`, the std-implementation walks forwards, as in https://github.com/rust-lang/rust/blob/2ceed0b6cb9e9866225d7cfcfcbb4a62db047163/library/std/src/sys/unix/os.rs#L560, and AFAIK this did not fix any known vulnerabilities when it was added, ref. https://github.com/fish-shell/fish-shell/issues/2784

branchvincent:
fyi, after https://github.com/fish-shell/fish-shell/commit/eacbd6156d9474b544d247894a7bc7668832aac7 I'm seeing a garbage `__fish_data_dir` (causing fish to skip its initialization in [`share/config.fish`](https://github.com/fish-shell/fish-shell/blob/master/share/config.fish)):

```console
$ git checkout eacbd6156
$ sudo make install
$ fish_trace=1 /usr/local/bin/fish -c 'set --show __fish_data_dir __fish_sysconf_dir'
> source /Users/branch/.config/fish/config.fish
--> status is-login
--> exit
> set --show __fish_data_dir __fish_sysconf_dir
$__fish_data_dir: set in global scope, unexported, with 1 elements
$__fish_data_dir[1]: |fish|
$__fish_sysconf_dir: set in global scope, unexported, with 1 elements
$__fish_sysconf_dir[1]: |fish|
```

Note the actual binary seems fine until you `mv` it somewhere else:

```console
$ fish_trace=1 ./build/fish -c 'set --show __fish_data_dir __fish_sysconf_dir'
> source /Users/branch/.cache/brew/fish--git/share/config.fish
--> set -g IFS \n\ \t
--> ... 
--> ... all expected initialization here
--> ...
> set --show __fish_data_dir __fish_sysconf_dir
$__fish_data_dir: set in global scope, unexported, with 1 elements
$__fish_data_dir[1]: |/Users/branch/.cache/brew/fish--git/share|
$__fish_sysconf_dir: set in global scope, unexported, with 1 elements
$__fish_sysconf_dir[1]: |/Users/branch/.cache/brew/fish--git/etc|
```

faho:
@branchvincent Have you cleaned your build directory? That was the issue for me.

Just remove it and start anew.

branchvincent:
Hmm, not sure why but still seeing the same even with a fresh clone

```console
$ git clone --depth=1 https://github.com/fish-shell/fish-shell /tmp/fish
$ sudo make -C /tmp/fish install
$ /usr/local/bin/fish -c 'set --show __fish_data_dir'
$__fish_data_dir: set in global scope, unexported, with 1 elements
$__fish_data_dir[1]: |fish|
```

faho:
Ah, okay, it only happens on install.

Alright, seems we don't set these hard enough. Try d6a6d77.

(also gosh I always forget about our awkward wrapper makefile)

branchvincent:
thanks a lot @faho, looks good now!

PR #9963, opened 2023-08-18
Port builtin count to rust

faho:
## Description

This is one of the simpler builtins.

Not much to say, except that it punts the actual argument reading to "Arguments" - which presumably *could* be extended with modes to only use stdin, but this isn't *too* bad.

## TODOs:

### DIFF
--- a/fish-rust/src/builtins/count.rs
+++ b/fish-rust/src/builtins/count.rs
@@ -21,7 +21,7 @@ pub fn count(
     // This means excluding lines that don't end in a newline!
     numargs += Arguments::new(&[] as _, &mut zero, streams, COUNT_CHUNK_SIZE)
         // second is "want_newline" - whether the line ended in a newline
-        .filter(|x| x.1)
+        .filter(|(_arg, has_trailing_newline)| has_trailing_newline)
         .count();
 
     streams.out.appendln(numargs.to_string());
--- a/fish-rust/src/builtins/count.rs
+++ b/fish-rust/src/builtins/count.rs
@@ -21,7 +21,7 @@ pub fn count(
     // This means excluding lines that don't end in a newline!
     numargs += Arguments::new(&[] as _, &mut zero, streams, COUNT_CHUNK_SIZE)
         // second is "want_newline" - whether the line ended in a newline
-        .filter(|(_arg, has_trailing_newline)| has_trailing_newline)
+        .filter(|x| x.1)
         .count();
 
     streams.out.appendln(numargs.to_string());
--- a/fish-rust/src/builtins/count.rs
+++ b/fish-rust/src/builtins/count.rs
@@ -1,0 +1,33 @@
+use super::prelude::*;
+
+// How many bytes we read() at once.
+// Since this is just for counting, it can be massive.
+const COUNT_CHUNK_SIZE: usize = 512 * 256;
+
+/// Implementation of the builtin count command, used to count the number of arguments sent to it.
+pub fn count(
+    _parser: &mut parser_t,
+    streams: &mut io_streams_t,
+    argv: &mut [&wstr],
+) -> Option<c_int> {
+    // Always add the size of argv (minus 0, which is "count").
+    // That means if you call `something | count a b c`, you'll get the count of something _plus 3_.
+    let mut numargs = argv.len() - 1;
+
+    // (silly variable for Arguments to do nothing with)
+    let mut zero = 0;
+
+    // Count the newlines coming in via stdin like `wc -l`.
+    // This means excluding lines that don't end in a newline!
+    numargs += Arguments::new(&[] as _, &mut zero, streams, COUNT_CHUNK_SIZE)
+        // second is "want_newline" - whether the line ended in a newline
+        .filter(|x| x.1)
+        .count();
+
+    streams.out.appendln(numargs.to_string());
+
+    if numargs == 0 {
+        return STATUS_CMD_ERROR;
+    }
+    STATUS_CMD_OK
+}
--- a/fish-rust/src/builtins/count.rs
+++ b/fish-rust/src/builtins/count.rs
@@ -24,7 +24,7 @@ pub fn count(
         .filter(|x| x.1)
         .count();
 
-    streams.out.appendln(numargs.to_string());
+    streams.out.appendln(numargs.to_wstring());
 
     if numargs == 0 {
         return STATUS_CMD_ERROR;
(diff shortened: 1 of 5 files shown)


fish-rust/src/builtins/count.rs
```diff
@@ -0,0 +1,33 @@
...
+    let mut zero = 0;
+
+    // Count the newlines coming in via stdin like `wc -l`.
+    // This means excluding lines that don't end in a newline!
+    numargs += Arguments::new(&[] as _, &mut zero, streams, COUNT_CHUNK_SIZE)
+        // second is "want_newline" - whether the line ended in a newline
+        .filter(|x| x.1)
+        .count();
+
+    streams.out.appendln(numargs.to_string());
```
henrikhorluck, line 27: Okay, so since `appendln` uses `impl Into<WString>` this does not need to be `.to_wstring()`, but I think that just supports that we should remove `ToWString` assuming the behaviour is identical, if not this should be `.to_wstring()`

src/builtin.cpp
```diff
@@ -210,42 +210,6 @@ static maybe_t<int> implemented_in_rust(parser_t &, io_streams_t &, const wchar_
...
-        assert(streams.stdin_fd >= 0 &&
-               "Should have a valid fd since stdin is directly redirected");
-        char buf[COUNT_CHUNK_SIZE];
-        while (true) {
-            long n = read_blocked(streams.stdin_fd, buf, COUNT_CHUNK_SIZE);
-            if (n == 0) {
-                break;
-            } else if (n < 0) {
-                wperror(L"read");
-                return STATUS_CMD_ERROR;
```
henrikhorluck, line 232: This errored out completely if `read` failed, and printed an error message. Arguments just silently ignores those errors. We can either duplicate code from Arguments directly here or add:

```rust
pub struct ResultArguments(Arguments);
```

And make `get_arg_stdin` return an `io::Result`, forward that directly in `Iteraror for ResultArguments`, while moving the `.ok()` to `next()` in `Iteraror for Arguments`

We could also add `StdinOnlyArguments` or another constructor to `ResultArguments` that makes it unnecessary to send the dummy-`args`

henrikhorluck: Note that you can `let res: io::Result<usize> = iter_of_result.collect()`

faho: > This errored out completely if read failed, and printed an error message. Arguments just silently ignores those errors.

This is the biggest change here, but tbh I've looked over how `read` *can* fail - assuming this fails like read(2), and I am not sure what the failure modes here *are*?

My man page says:

- EBADF, EFAULT, EINVAL, EISDIR - these are checked when we open the fd
- EAGAIN / EWOULDBLOCK - only relevant for O_NONBLOCK, which I think we would have to do on the fd, and we don't?
- EIO - *possibly* for network filesystems when the lock is lost? maybe?
- EINTR - handled by read_blocked
- "other errors"

> Note that you can let res: io::Result<usize> = iter_of_result.collect()


Sorry, I don't understand what you are saying.

faho: > We could also add StdinOnlyArguments or another constructor to ResultArguments that makes it unnecessary to send the dummy-args

I said as much in the opening message. My gut says this isn't a problem often enough to warrant it.

fish-rust/src/builtins/count.rs
```diff
@@ -0,0 +1,33 @@
...
+    let mut numargs = argv.len() - 1;
+
+    // (silly variable for Arguments to do nothing with)
+    let mut zero = 0;
+
+    // Count the newlines coming in via stdin like `wc -l`.
+    // This means excluding lines that don't end in a newline!
+    numargs += Arguments::new(&[] as _, &mut zero, streams, COUNT_CHUNK_SIZE)
+        // second is "want_newline" - whether the line ended in a newline
+        .filter(|x| x.1)
```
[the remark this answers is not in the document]
faho: Renaming it in arguments sounds good.

I don't really "name" it here as such because I get the tuple, and I'm not sure you can access those by name?

src/builtin.cpp
henrikhorluck: >  > Note that you can let res: io::Result = iter_of_result.collect()

> Sorry, I don't understand what you are saying.

I thought it might be relevant to mention that you can short-circuit if you have an iterator of `Result`, which is not a super-obvious feature of `.collect()`. I wrongly used  `io::Result<usize>` instead of `io::Result<Vec<bool>>` which I think caused the confusion, as you cannot collect to a `usize`

```rust
fn main() {
    let results = [Ok(1), Ok(2), Err("Failed"), Ok(4)];
    let filtered: Result<Vec<usize>, &str> = results.into_iter().collect();
    // prints Err("Failed")
    println!("{:?}", filtered);
    let results = [Ok(1), Ok(2), Ok(3), Ok(4)];
    let filtered: Result<Vec<usize>, &str> = results.into_iter().collect();
    // prints Ok([1, 2, 3, 4])
    println!("{:?}", filtered);
}
```

fish-rust/src/builtins/count.rs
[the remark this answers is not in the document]
henrikhorluck: You would deconstruct it:  
```suggestion
        // second is "want_newline" - whether the line ended in a newline
        .filter(|(_args, has_trailing_newline)| has_trailing_newline)
```

src/builtin.cpp
henrikhorluck: Thinking over it, given that we ignore the result in string and path, I agree with you that this can just be ignored, would be weird to handle it here but not there

fish-rust/src/builtins/count.rs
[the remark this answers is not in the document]
faho: Ah, okay - I tried `|arg, newline|` directly. Rust's parentheses confuse me.

src/builtin.cpp
faho: Funnily enough I had a comment to that effect originally but removed it for being weird meta-commentary that I tend to add too much of.

fish-rust/src/builtins/count.rs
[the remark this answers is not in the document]
henrikhorluck: Ugh, it also needs to dereference `has_trailing_newline`, usually you would just use `|&a| a` if `a` is `Copy`, since that is equal to `|a| *a`
But `_arg` here is not `Copy` so we can't combine it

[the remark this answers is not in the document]
faho: Alright, reverted.

PR #9966, opened 2023-08-19
Implement FLOGF formatting

faho:
## Description

As complained about in #9962, this adds the actual formatting capabilities to FLOGF, by just calling sprintf!.

Because of that some calls that didn't use a format string were no longer correct and I switched them to FLOG. I *think* I got all of them.

Mostly, I'm not sure that this actually is so simple?

One change from the C++ behavior is that `FLOGF!(category, string);` will refuse to compile - it *requires* some argument for the format string. I think that's probably correct because in that case you should use `FLOG!` and have probably made a mistake, that might be benign or might not.

### DIFF
--- a/fish-rust/src/env_dispatch.rs
+++ b/fish-rust/src/env_dispatch.rs
@@ -3,7 +3,7 @@ use crate::curses::{self, Term};
 use crate::env::{setenv_lock, unsetenv_lock, EnvMode, EnvStack, Environment};
 use crate::env::{CURSES_INITIALIZED, READ_BYTE_LIMIT, TERM_HAS_XN};
 use crate::ffi::is_interactive_session;
-use crate::flog::FLOGF;
+use crate::flog::FLOG;
 use crate::function;
 use crate::output::ColorSupport;
 use crate::wchar::prelude::*;
@@ -130,7 +130,7 @@ impl VarDispatchTable {
 
 fn handle_timezone(var_name: &wstr, vars: &EnvStack) {
     let var = vars.get_unless_empty(var_name).map(|v| v.as_string());
-    FLOGF!(
+    FLOG!(
         env_dispatch,
         "handle_timezone() current timezone var:",
         var_name,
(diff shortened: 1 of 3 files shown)


henrikhorluck (commented):
LGTM! Regarding #9948 this could use `format!`, but that would introduce mandatory noise, unlike that PR. I think it is better to merge this as is. A potential swap to [log](https://crates.io/crates/log) should come later to not have an intermediate state of FLOG-ported-but-using-rust-fmt-strings.

> As complained about in https://github.com/fish-shell/fish-shell/pull/9962

I only pointed out that I encountered a couple issues. The text in question is

> I encountered a couple issues: fwprintf! appears to not do anything? FLOGF! does not format

I specifically formulated that to not sound like a complaint. The issue here is rather that I had not looked closely enough on the currently ported implementations, and I am not up-to-data on the porting-status of every piece of `fish`, which I don't think anyone should be expected to be. My intention was therefore to ask if someone could clarify if I was doing something wrong or the previously ported code was bugged or only partially implemented.

The answer is that it was partially implemented, but did not error when used in an unimplemented way, and this now fixes that. Thank you!

If there is something about the way I phrased it that makes it come across as complaining I would love to know. I think complaining in that situation would be somewhat rude which was not my intention at all! For example, would it be clearer if I said "FLOGF! appears to not format?" instead of "FLOGF! does not format". Do I need to be more specific about my intentions?
I believe both of us have English as a second language so a misinterpretation would also be possible.

Aside: Using `todo!()` makes it a lot clearer that something is ported in a not-yet-working state, e.g. I attempted to use `parse_util_detect_errors_in_ast`, but that function relies on `OperationContext::empty()`, which currently is just a `todo!()`, which made it easier to find out that I needed to use the C++ version for now

faho:
> I only pointed out that I encountered a couple issues.

Oh, sure, yes, yes. I didn't mean to imply that this was annoying or anything, you were entirely correct to point it out!

If you read my use of the term "complaining" as anything else then I apologize that was not my intention. Please imagine I had said "pointed out" instead.

>I believe both of us have English as a second language so a misinterpretation would also be possible.

Chronologically speaking, English is technically my third language.

> Regarding https://github.com/fish-shell/fish-shell/pull/9948 this could use format!, but that would introduce mandatory noise, unlike that PR

Yeah I looked into this and once I saw how small the actual needed change was I felt like that was simple enough to do now, swapping the formatting language can come later.

> LGTM! 

Alright, merging then!

Issue #9908, opened 2023-07-22
Markupsafe missing in the build process


remyleone:
```
$ make                                    
[  0%] Built target cargo-prebuild_fish-rust
    Finished release [optimized] target(s) in 0.42s
Copying byproducts `libfish_rust.a` to /Users/sieben/workspace/fish-shell
[  0%] Built target _cargo-build_fish-rust
[  0%] Built target cargo-build_fish-rust
[  1%] Built target pofiles_1
[  2%] Built target pofiles_2
[  3%] Built target pofiles_3
[  4%] Built target pofiles_4
[  6%] Built target pofiles_5
[  7%] Built target pofiles_6
[  8%] Built target pofiles_7
[  8%] Built target CHECK-FISH-BUILD-VERSION-FILE
[ 86%] Built target fishlib
[ 88%] Built target fish
[ 91%] Built target fish_indent
[ 93%] Built target fish_key_reader
[ 95%] Building man pages with Sphinx
Traceback (most recent call last):
  File "/usr/local/bin/sphinx-build", line 5, in <module>
    from sphinx.cmd.build import main
  File "/usr/local/lib/python3.9/site-packages/sphinx/cmd/build.py", line 25, in <module>
    from sphinx.application import Sphinx
  File "/usr/local/lib/python3.9/site-packages/sphinx/application.py", line 32, in <module>
    from sphinx.config import Config
  File "/usr/local/lib/python3.9/site-packages/sphinx/config.py", line 27, in <module>
    from sphinx.util.tags import Tags
  File "/usr/local/lib/python3.9/site-packages/sphinx/util/tags.py", line 11, in <module>
    from jinja2 import nodes
  File "/usr/local/lib/python3.9/site-packages/jinja2/__init__.py", line 6, in <module>
    from markupsafe import escape
ModuleNotFoundError: No module named 'markupsafe'
make[3]: *** [CMakeFiles/sphinx-manpages] Error 1
make[2]: *** [CMakeFiles/sphinx-manpages.dir/all] Error 2
make[1]: *** [all] Error 2
make: *** [all] Error 2
```

Does CMake ensure that all python dependencies are installed?

faho:
> Does CMake ensure that all python dependencies are installed?

Cmake does not handle dependencies. At least not the way we're using it. The good news is it's on its way out.

Your sphinx installation is incomplete. I recommend installing markupsafe.

PR #9971, opened 2023-08-21
add a requirements.txt to specify that sphinx is needed to build

remyleone:
## Description

I think specifying that sphinx and maybe other dependencies are needed could make it more explicit

https://github.com/fish-shell/fish-shell/issues/9908


faho:
This isn't a python project, requirements.txt looks like it is. This is misleading because it makes people think *only* sphinx is required, which is not the case.

And specifically: #9908 *had* sphinx installed! Sphinx was there, but sphinx was incomplete! So this wouldn't have helped in that case because you already had sphinx.

The dependencies are explained in [the README](https://github.com/fish-shell/fish-shell/blob/master/README.rst#dependencies-1).

Sorry, I am rejecting this.

PR #9980, opened 2023-08-24
Run `cargo fmt` with Rustfmt 1.6.0

henrikhorluck:
- "1.6.0" now supports formatting let-else statements which we use liberally,
  and appears to have some fixes in regards to long-indented-lines with macros
  like `wgettext_fmt!`
- This commit updates the formatting so that devs with the latest stable don't
  see random format-fixes upon running `cargo fmt` (and so that CI can pass, we use always run clips and fmt with latest stable)

Release-notes for `rustfmt`: https://github.com/rust-lang/rustfmt/blob/master/CHANGELOG.md (the release-date in `rustup` says it was released yesterday along with Rust `1.72.0`)

PR #9981, opened 2023-08-24
Replace rust toolchain setup in CI

henrikhorluck:
- Replace ATiltedTree/setup-rust with rust-toolchain

This hopefully makes it not necessary to delete the cache on every new Rust release, see https://github.com/fish-shell/fish-shell/pull/9980, https://github.com/fish-shell/fish-shell/commit/5c29ff52fb76ad360c2e1078289f4a30022875d5, and https://github.com/fish-shell/fish-shell/commit/54fa1ad6ec7adb0d9c14076843ef5e51c18ecc7a, likely at the cost of not caching, but this is network-limited regardless, so I don't think it should be that much of a slowdown

EDIT: there is also a warning with every [ATiltedTree/setup-rust](https://github.com/ATiltedTree/setup-rust) use:
```
The following actions uses node12 which is deprecated and will be forced to run on node16: ATiltedTree/setup-rust@v1. For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/
```

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,10 +16,7 @@ jobs:
 
     steps:
     - uses: actions/checkout@v3
-    - name: SetupRust
-      uses: ATiltedTree/setup-rust@v1
-      with:
-        rust-version: 1.67
+    - uses: dtolnay/rust-toolchain@1.67
     - name: Install deps
       run: |
         sudo apt install gettext libncurses5-dev libpcre2-dev python3-pip tmux
@@ -46,11 +43,9 @@ jobs:
 
     steps:
     - uses: actions/checkout@v3
-    - name: SetupRust
-      uses: ATiltedTree/setup-rust@v1
+    - uses: dtolnay/rust-toolchain@1.67
       with:
-        rust-version: 1.67
-        targets: "i686-unknown-linux-gnu" # setup-rust wants this space-separated
+        targets: "i686-unknown-linux-gnu" # rust-toolchain wants this comma-separated
     - name: Install deps
       run: |
         sudo apt update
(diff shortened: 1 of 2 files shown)


henrikhorluck:
Timings: 
- https://github.com/ATiltedTree/setup-rust with cache save 45s https://github.com/fish-shell/fish-shell/actions/runs/5964922610/job/16181541533?pr=9980
- https://github.com/ATiltedTree/setup-rust with cache hit 6s https://github.com/fish-shell/fish-shell/actions/runs/5956496844/job/16157383929
- https://github.com/dtolnay/rust-toolchain (no-caching) 12s https://github.com/fish-shell/fish-shell/actions/runs/5965307730/job/16182434699?pr=9981

henrikhorluck:
We could probably also fix this by disabling caching in the existing action, but I can't find a way to do so, and since it is forced to run with a potentially incompatibly node-version, I see no reason to keep it.

The rustup-issue that it appears we are running into is https://github.com/rust-lang/rustup/issues/2417 as far as I can tell

We can add caching on top of this action, but we need a better cache-key to workaround the rustup-issue. We can either resolve `stable` to the current version by calculating it based on the 6-week release schedule (which appears to be how the dtolnay-action supports [relative stables](https://github.com/dtolnay/rust-toolchain/blob/00b49be78f40fba4e87296b2ead62868750bdd83/action.yml#L36C1-L36C1)), which seems relatively error-prone and needs some tests, or querying some endpoint (I am not entirely sure which)

henrikhorluck:
It appears you not only need to purge both the default-branch cache and each individual PR, ref https://github.com/fish-shell/fish-shell/actions/runs/5980654222/job/16227196307?pr=9962

henrikhorluck:
I re-wrote the commit-message to no longer say that the dtolnay-toolchain-action is a fork, as afaik that is not correct. It is rather a new take on the same idea. It is implemented in what is essentially a single file https://github.com/dtolnay/rust-toolchain/blob/master/action.yml, and I think it is a much more sensible dependency than unnecessary javascript of most third-party Github Actions

PR #9984, opened 2023-08-25
 wcstod_underscores: Stop going quadratic with sums without spaces 

faho:
## Description

Something I regularly do:

```fish
somecommand | string join + | math
```

Imagine my surprise when that suddenly took over 3 seconds with 10000
numbers:

```fish
seq 10000 | string join + | math
```

while the very similar

```fish
seq 10000 | string join " + " | math
```

took 80ms (going by `time`).

This is caused by our underscore extraction:

We go through the number string, and take every char that could possibly be
in a number, *including signs* (+/-).

That means, if we get 1+2+3+4+5+6+7..+1000, without spaces in-between,
we will:

- copy the entire string
- pass it off to the number parser
- that will tell us it consumed 1 byte
- we return the result
- math calls us again with "2+3+4+5+6+7..+1000"
- repeat

So in effect we have gone quadratic over the length of the string.

I don't love this solution, because it requires us to awkwardly
pre-parse the number instead of leaving that to the number parser.
Unfortunately, I don't see another way.

Benchmarks:

```fish
set -l seq (seq $number | string join +); math $seq
```

takes 4.15s without this, 18.7ms with (according to hyperfine) - that's a factor of 222x speedup. For 100 numbers it's 3% faster, and that's about the measuring threshold for one run.

It is ever so slightly slower for `for i in (seq 10000); echo 1 + 1 | math; end` - that's 10000 runs of one computation, and about 2% slower. For `echo 1+1 | math` there is no measurable difference even with 10000 runs.

## TODOs:

- [N/A] Tests have been added - we don't really have performance regression tests - do we want a benchmark for this?
- [ ] Backport to 3.6.1? The issue already existed there

I would like our resident number fiddler @ridiculousfish to take a look at this.

### DIFF
--- a/fish-rust/src/wutil/wcstod.rs
+++ b/fish-rust/src/wutil/wcstod.rs
@@ -7,12 +7,27 @@ where
     I: Iterator<Item = char> + Clone,
 {
     let mut whitespace_skipped = 0;
+    let mut first_is_digit_or_sep = false;
+    let mut possible_hex = false;
     // Skip leading whitespace.
+    // And while we're here, remember if the first non-whitespace char
+    // could start a hex number or possible overflow
     loop {
         match chars.clone().next() {
             Some(c) if c.is_whitespace() => {
                 whitespace_skipped += 1;
                 chars.next();
+            }
+            Some('+' | '-') => {
+                possible_hex = true;
+                break;
+            }
+            Some(c) if c.is_ascii_digit() || c == decimal_sep => {
+                first_is_digit_or_sep = true;
+                if c == '0' {
+                    possible_hex = true;
+                }
+                break;
             }
             None => {
                 *consumed = 0;
@@ -23,7 +38,7 @@ where
     }
 
     // If it's a hex float, use hexponent.
-    if is_hex_float(chars.clone()) {
+    if possible_hex && is_hex_float(chars.clone()) {
         let mut n = 0;
         let res = hexponent::parse_hex_float(chars, decimal_sep, &mut n);
         if res.is_ok() {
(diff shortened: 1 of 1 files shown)


faho:
One other benchmark is the number of brk calls:

```fish
strace -ce brk ./fish-new --no-config -c 'seq 10000 | string join + | math'
```

27993 in the old version, 13 now. We would allocate ourselves silly.

ridiculousfish:
This took way too long for me to review. I think your fix is good, but also we can do even better. Instead of constructing a "pruned" string to parse, we can create an iterator that filters underscores as it goes. That way there's no separate allocation.

I attempted to realize this in [this commit](https://github.com/ridiculousfish/fish-shell/commit/c9785704271f114ae912e448e1fb3b758c2708c6). This avoids the separate "pruned" string entirely.

faho:
> Instead of constructing a "pruned" string to parse, we can create an iterator that filters underscores as it goes. That way there's no separate allocation.

Oooh, nice. I had thought about that, but I thought it would require a change to the parsing functions - I didn't realize you could just make a new iterator with .filter() and pass it already!

That's of course much better, and something that would have been quite a bit harder in C++. Gosh, I love a good iterator protocol.

> I attempted to realize this in [this commit](https://github.com/ridiculousfish/fish-shell/commit/c9785704271f114ae912e448e1fb3b758c2708c6). This avoids the separate "pruned" string entirely.

As far as I can tell that looks good to me, so I'm closing this, thanks!

faho:
@ridiculousfish Did you ever end up committing that fix?

faho:
@ridiculousfish I've cherry-picked your fix as 49c848ed42c37003fe6e26f19556923ff357794d for 4.1.

It is *not* in 3.7 or 4.0.

PR #9990, opened 2023-08-27
WIP Port execution

krobelus:
Not yet ready for review.


fish-rust/src/builtins/string.rs
```diff
@@ -287,82 +284,79 @@ fn escape_code_length(code: &wstr) -> Option<usize> {
...
-    let subcmd_name = args[1];
+    let (cmd_names, args) = args.split_at_mut(2);
+    let cmd = &cmd_names[0];
+    let subcmd_name = &cmd_names[1];
 
-    match subcmd_name.to_string().as_str() {
-        "collect" => collect::Collect::default().run(parser, streams, args),
-        "escape" => escape::Escape::default().run(parser, streams, args),
-        "join" => join::Join::default().run(parser, streams, args),
+    let mut subcmd: Box<dyn StringSubCommand<'_>> = match subcmd_name.to_string().as_str() {
```
henrikhorluck, line 318: https://github.com/fish-shell/fish-shell/commit/d6b41b33f4c888bbce3cba1e14e3dbf9dadc0afd#r124996810

> I wrote: @krobelus I disagree with using dynamic dispatch here, that change is noise and I don't see any good reasoning for it. I do have a regret from the string-rewrite in that it mixes data-and-behaviour — each subcommand-module should rather only expose `BuildinCmd`s, not any structs, which removes all the repeated arguments without using dynamic dispatch.

> @krobelus responded: I want to only construct the wgetopter_t in one place, not once per subcommand. Maybe this needs a quick macro

henrikhorluck: I don't understand what you mean by constructing `wgetopter_t` in only one place, IMO this change should just be reverted.

fish-rust/src/builtins/shared.rs
```diff
@@ -1,64 +1,66 @@
...
-        fn rust_run_builtin(
-            parser: Pin<&mut parser_t>,
-            streams: Pin<&mut io_streams_t>,
-            cpp_args: &wcstring_list_ffi_t,
-            builtin: RustBuiltin,
-            status_code: &mut i32,
-        ) -> bool;
-    }
-}
+pub type BuiltinCmd = fn(&Parser, &mut IoStreams<'_>, &mut [WString]) -> Option<c_int>;
```
henrikhorluck, line 26: https://github.com/fish-shell/fish-shell/commit/d6b41b33f4c888bbce3cba1e14e3dbf9dadc0afd#r124997196
> I wrote: I disagree with the changing of `&mut [&wstr]` to `&mut [WString]`. It is _very_ noisy and I don't see a good reasoning for it. In my opinion `&wstr` is actually a more useful type since we essentially never want to modify it directly. Utilising any spare capacity or mutating in-place seems like premature optimisations.

> @krobelus responded: 
> I changed [&wstr] to [WString] (I think I mentioned it on your PR before) because unlike [&wstr], it occurs naturally in our code.
> It seemed more elegant to not require conversion for builtin execution.
> My suspicion was that the existing Rust uses [&wstr] because it didn't account for the the whole code base so I thought I was correcting an oversight here.. which is unlikely I guess.
> 
> &wstr is actually a more useful type since we essentially never want to modify it directly.
> This property we could probably encode independent of whether we use a reference.
> 
> I might very well be wrong. You mention some good points in https://matrix.to/#/!YLTeaulxSDauOOxBoR:matrix.org/$B2oLWotpXrSVble7xAPpoV_mi-4FNA6n2BwSH46l8lE?via=gitter.im&via=matrix.org&via=tchncs.de
> My opinion is not very strong here. I'm happy to defer to whoever takes ownership of this code.
> 
> I wonder if we could get the best of both worlds by taking something like [AsRef<wstr>].. however I don't want to invest too much time in this relatively minor issue (it's pretty much isolated to the wgetopt module), I'd rather get the rest of the system working. There's a whole lot to clean up

henrikhorluck: > I wonder if we could get the best of both worlds by taking something like [AsRef].. however I don't want to invest too much time in this relatively minor issue (it's pretty much isolated to the wgetopt module), I'd rather get the rest of the system working. There's a whole lot to clean up

My two cents: make `wgetoptr_t` take `AsRef<wstr>`, extract what you changed about it to a separate PR.

If we want to change the `BuiltinCmd`-type to use `WString`, that should be a separate PR that should land before this (even if that requires cloning to support it until this lands, since it is so breaking). I still disagree with changing it from `&wstr` and I would rather make the private encoding optional and use `BStr` aka `Vec<u8>`, but that's a very different discussion I'd rather do after the port is finished

fish-rust/src/io.rs
```diff
@@ -894,8 +947,8 @@ impl OutputStream for BufferedOutputStream {
 
 pub struct IoStreams<'a> {
     // Streams for out and err.
-    pub out: &'a dyn OutputStream,
-    pub err: &'a dyn OutputStream,
+    pub out: &'a mut dyn OutputStream,
```
henrikhorluck, line 950: https://github.com/fish-shell/fish-shell/commit/d6b41b33f4c888bbce3cba1e14e3dbf9dadc0afd#r124997448

> I wrote: OutputStream should really not use dynamic dispatch IMO, see my previous comment https://github.com/fish-shell/fish-shell/pull/9916#issuecomment-1655563903
> 
> > Aside: regarding IoStreams, I think it is better to avoid object-safety and use an OutputStream-enum that delegates, like [StringMatcher](https://github.com/fish-shell/fish-shell/blob/2ec36338f2b1f21f4074ff7184c37619a4a61e69/fish-rust/src/builtins/string/match.rs#L157) from string match. Then we can avoid the API breakage of no generics. Using dyn would be fine if we used [io::Write](https://doc.rust-lang.org/stable/std/io/trait.Write.html) + [write!(streams.out, output)](https://doc.rust-lang.org/stable/std/macro.write.html), but that's a much larger change that I don't think we want right now. I think that a delegating enum is the closest we have to subclassing from C++, even if it is a bit of boilerplate.
> 
> I think you can use generics if you implement something directly on `impl dyn OutputStream`, but I'm not very familiar with that, you generally want to avoid dynamic dispatch if possible due to requiring object-safety

> @krobelus responded: 
> > Then we can avoid the API breakage of no generics.
> yeah good point

fish-rust/src/builtins/string.rs
```diff
@@ -26,118 +25,116 @@ macro_rules! string_error {
...
+                    );
                     return Err(STATUS_INVALID_ARGS);
                 }
                 '?' => {
-                    string_unknown_option(parser, streams, cmd, args_read[w.woptind - 1]);
+                    string_unknown_option(parser, streams, w.cmd(), &args_read[w.woptind - 1]);
                     return Err(STATUS_INVALID_ARGS);
                 }
                 c => {
-                    let retval = self.parse_opt(cmd, c, w.woptarg);
```
henrikhorluck, line 79: https://github.com/fish-shell/fish-shell/commit/d6b41b33f4c888bbce3cba1e14e3dbf9dadc0afd#r124997793
> I wrote: I don't get why this isn't just a change to `let retval = self.parse_opt(w.cmd(), c, w.woptarg())`. This is very noisy. (even if changed to `WString` those methods should give references, I think)

> @krobelus responded:
> > I don't get why this isn't just a change to `let retval = self.parse_opt(w.cmd(), c, w.woptarg())`. This is very noisy. (even if changed to `WString` those methods should give references, I think)
> 
> yeah that's probably wrong (maybe I thought one of them was a mut reference).
The noisiness we can handle by reorganizing commits later ;)

henrikhorluck (commented):
I am re-posting some comments made on an earlier specific commit

fish-rust/src/io.rs
henrikhorluck: We could use both `io::Write` to skip private-encode-undo-ing[^2] and `fmt::Write` to take possibly-private-encoded UTF8[^1], and undo the private-encoding before writing it to the stream. (but that is for a potential UTF8-swap)

[^2]: That sounds mildly risky
[^1]: `WString` currently has a dual purpose of UTF32 + private-encoded

tests/checks/default-setup-path.fish
```diff
@@ -0,0 +1,12 @@
+#RUN: %fish -C 'set -g fish %fish' %s
+
+# FIXME: Rationalize behavior when PATH is explicitly unset, should this not behave like PATH=""?
```
faho, line 3: It is very unlikely that that's what people would expect, and a $PATH of "." is basically unusable. If that's *explicitly* set, sure, whatever - but an unset $PATH can also be caused by misconfiguration or e.g. a thing that just clears the environment.

$PATH is also something where we should act like other things because it's a system-wide config knob, and other tools also fall back to their default paths - try `env -u PATH $shell`, which will print some default for bash, dash and zsh.

krobelus:
Thanks so much for the comments, I think I addressed them.
current status:
1. there is one remaining test failure in `test_history.py`
2. macOS build is failing
3. there are (harmless) memory leaks at the history FFI boundary
4. I'm working on porting the unit tests
5. once done I'll try to read through everything. There should be no need for jumping around, so we can read 1k lines / hour. It's still a lot at 25k lines

fish-rust/src/io.rs
```diff
@@ -189,8 +194,29 @@ pub trait IoData {
...
     // The address of the object, for comparison.
     fn as_ptr(&self) -> *const ();
+    fn as_bufferfill(&self) -> Option<&IoBufferfill> {
+        None
+    }
 }
 
+// todo!("later: this should be safe because of how it's used. Rationalize this better.")
+pub trait IoDataSync: IoData + Send + Sync {}
+unsafe impl Send for IoClose {}
```
krobelus, line 204: These are a bit ugly. I'm pretty sure there is no problem in practice but I didn't check yet.

krobelus:
For now I cut some corners with respect to making the patches reviewable but I don't think there is anything interesting here.
Some more testing with a debug build would be good.

ridiculousfish (commented):
Just incredible work overall. Unbelievable.

I spotted some local errors and have a question about the use of `&mut Parser`. I have not finished reviewing all the files but I'm leaving my accumulated comments now, and will finish the review as soon as I can.

fish-rust/src/builtins/cd.rs
```diff
@@ -9,10 +9,11 @@ use crate::{
 };
 use errno::{self, Errno};
 use libc::{fchdir, EACCES, ELOOP, ENOENT, ENOTDIR, EPERM, O_RDONLY};
+use std::rc::Rc;
 
 // The cd builtin. Changes the current directory to the one specified or to $HOME if none is
 // specified. The directory can be relative to any directory in the CDPATH variable.
-pub fn cd(parser: &mut parser_t, streams: &mut io_streams_t, args: &mut [&wstr]) -> Option<c_int> {
+pub fn cd(parser: &Parser, streams: &mut IoStreams, args: &mut [&wstr]) -> Option<c_int> {
```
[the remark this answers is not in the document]
krobelus: I don't remember, would need to try. Maybe because it allowed a more direct port of the operation context, which is passed by const reference. C++ allows mutating the operation context's parser field because it's inside a shared pointer but Rust doesn't. We'd add a Mutex to `ParserRef`.

faho:
I just tried running our benchmarks on this, and the results are encouraging.

In general, running any sort of rust-builtin with this is ~60% faster than without, which shows the cost of FFI nicely. I've seen this before, anything that crosses the C++/Rust boundary pays a fairly heavy cost.

Globs are ~80% faster than without (and 20% faster than 3.6.1, but that could also be due to some of our glob changes).

The aliases.fish one (that makes 1000 aliases) is about 20% faster than 3.6.1 (and about twice as fast as git master). That's the benchmark that'll exercise this the most (builtins only, a bunch of runs through a function).

Ironically no_execute.fish (which you would expect to exercise the "parser" a bunch, but then our "parser", uh, kinda isn't) is still 70% slower than 3.6.1, but a lot of that is due to startup being ~40% slower - it starts one fish per-file.

krobelus:
cool. It will be interesting to compare the finished port with `LastC++11` (with relevant changes backported). I don't expect much of a difference.

fish-rust/src/builtins/read.rs
```diff
@@ -1,7 +1,800 @@
...
+            let b = b[0];
+
+            nbytes += 1;
+            if MB_CUR_MAX() == 1 {
+                res = char::from(b);
+                finished = true;
+            } else {
+                let sz = unsafe {
+                    mbrtowc(
+                        std::ptr::addr_of_mut!(res).cast(),
```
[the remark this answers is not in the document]
Xiretza: There it was plain incorrect (it was taking a pointer to a slice, which is one indirection too many). Here it's not wrong, but that macro is almost always unnecessary.

fish-rust/src/wcstringutil.rs
```diff
@@ -464,38 +470,34 @@ pub fn trim(input: WString, any_of: Option<&wstr>) -> WString {
 /// \p idx may be "one past the end."
 pub fn count_preceding_backslashes(text: &wstr, idx: usize) -> usize {
     assert!(idx <= text.len(), "Out of bounds");
-    let mut backslashes = 0;
-    while backslashes < idx && text.char_at(idx - backslashes - 1) == '\\' {
-        backslashes += 1;
-    }
-    backslashes
+    text.chars().take(idx).take_while(|&c| c == '\\').count()
```
ridiculousfish, line 473: This looks wrong - this looks to be counting backslashes directly after `idx` but it should count the number of backslashes directly before `idx`.

ridiculousfish (approved):
This was a truly Herculean lift. It is astonishing. I'm left speechless.

Midway through the review I figured out your approach: to not worry too much about comments remaining correct and applicable, but rather to propagate the comments unchanged, so we can orient ourselves should we need to bisect. We can fix up the coments later. This makes a lot of sense to me, so I've endeavored to leave feedback only for legitimate bugs and nontrivial user-visible changes (of which there were very few).

krobelus:
Yeah I copied C++ comments verbatim to save time but also because I read them less often than using them as search string in git-blame output. Comments are great search strings because they have high entropy.
If I touched comments I'd probably trim lots of implementation details, especially now that we have a more usable type system.
In general I kinda ran out of time that I wanted to invest in this. I'll finish it soon.

zanchey:
Incredible work

PR #10000, opened 2023-09-05
ast: Require `--help` to parse more keywords as decorated statement

faho:
This makes it so

```fish
if -e foo
    # do something
end
```

complains about `-e` not being a command instead of `end` being used outside of an if-block.

That means both that `-e` could now be used as a command name (it already can outside of `if`!) *and* that we get a better error!

The only way to get `if` to be a decorated statement now is to use `if -h` or `if --help` specifically (with a literal option).

The same goes for switch and while.

It would be possible, alternatively, to disallow `if -e` and point towards using `test` instead, but the "unknown command" message should already point towards using `test` more than pointing at the "end" (that might be quite far away).

## TODOs:

- [ ] The comments here confuse me quite a bit. Have I understood the intent correctly? Do we want to change this even more?

### DIFF
--- a/fish-rust/src/ast.rs
+++ b/fish-rust/src/ast.rs
@@ -3514,6 +3514,7 @@ impl<'s> Populator<'s> {
             // If we are one of these, then look for specifically help arguments. Otherwise, if the next token
             // looks like an option (starts with a dash), then parse it as a decorated statement.
             let help_only_kws = [
+                ParseKeyword::kw_begin,
                 ParseKeyword::kw_function,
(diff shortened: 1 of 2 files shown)


ridiculousfish (approved):
LGTM. I think this can also be extended this to `begin`.

Recall that `begin`, for historical reasons, doesn't need a statement terminator. This is OK:

    begin echo hello; end

Today `begin --foo; end` fails with the same "end outside of a block" but I think it should enjoy your nice fix here too.

PR #6144, opened 2019-09-27
Use dynamic path resolve for chsh command instruction

MainShayne233:
## Description

After installing `fish`, the location of the executable was different from what the README.md stated (in `/usr/bin/fish`). Though this was *easy* to resolve, me from maybe even a year ago might have been very confused as to what is going on, and having this path get dynamically resolved seems a bit more user friendly.

Fixes issue #
N/A

## TODOs:
N/A

PR #6051, opened 2019-08-20
Default install directory for repo install + log out/in

josephrocca:

PR #5715, opened 2019-03-02
Updated readme

henrycunh:
There were cases where the install path wasn't clear, so using `chsh -s `which fish`` is a better option. Also explained what the command represents.

PR #10002, opened 2023-09-05
Make default shell configuration more generic

agirault:
## Description

Update command line instructions to use `$(which fish)` instead of hardcoded `/usr/local/bin/fish`

## TODOs:

- [ ] Tests have been added for regressions fixed -> NA
- [ ] User-visible changes noted in CHANGELOG.rst -> NA?

### DIFF
--- a/doc_src/index.rst
+++ b/doc_src/index.rst
@@ -75,13 +75,13 @@ To change your login shell to fish:
 
 1. Add the shell to ``/etc/shells`` with::
 
-    > echo /usr/local/bin/fish | sudo tee -a /etc/shells
+    > echo $(which fish) | sudo tee -a /etc/shells
 
 2. Change your default shell with::
 
-    > chsh -s /usr/local/bin/fish
+    > chsh -s $(which fish)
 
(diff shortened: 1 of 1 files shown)


faho:
Unfortunately, `which` is an external tool that may or may not be installed, and we want to keep the syntax use here to a minimum so it'll work from whatever prior shell.

Additionally, this doesn't make it easier to *understand*.

Sorry, but no.

Duplicate of #6144, #6051, #5715 - unfortunately this is an easy point to want to *fix*, more likely than it is an actual point of confusion.

agirault:
> Duplicate of https://github.com/fish-shell/fish-shell/pull/6144, https://github.com/fish-shell/fish-shell/pull/6051, https://github.com/fish-shell/fish-shell/pull/5715

my bad

> unfortunately this is an easy point to want to fix, more likely than it is an actual point of confusion.

I messed that up initially, hence the suggestion. Seeing all the previous duplicates above, my guess is that they ran into the same issues and were proactive about it, which others might not. I agree with @josephrocca's point [here](https://github.com/fish-shell/fish-shell/pull/6051#issuecomment-523089591)

> Sorry, but no.

np, though maybe that'd another data point to share that current doc isn't optimal in terms of usability. Cheers.

faho:
Alright, I gave explicit instructions from inside fish - so we can guarantee a sensible `command`.

PR #10004, opened 2023-09-06
completions: add ibmcloud

6A61736F6E206E61646572:
Adds completions for ibmcloud cli tool

### DIFF
--- a/share/completions/ibmcloud.fish
+++ b/share/completions/ibmcloud.fish
@@ -1,0 +1,11 @@
+set -l commands (ibmcloud --generate-bash-completion)
+
+complete -c ibmcloud --no-files \
+    --condition "not __fish_seen_subcommand_from $commands" \
+    --arguments "$commands"
+
+for cmd in $commands
+    complete -c ibmcloud --no-files \
+        --condition "__fish_seen_subcommand_from $cmd" \
+        --arguments "(ibmcloud $cmd --generate-bash-completion)"
+end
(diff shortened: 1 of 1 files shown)


share/completions/ibmcloud.fish
```diff
@@ -0,0 +1,11 @@
+set -l commands (ibmcloud --generate-bash-completion)
```
faho, line 1: What does this output look like? Could you give us a sample?

The "bash" part here is a bit suspect.

faho: Ah, that's what I thought.

It expects to be run like bash would run its completions. Importantly, that exports the commandline and where the cursor is.

If you don't do that, you only ever get the first argument, which isn't very helpful.

What you want is to run it with the same variables that bash would - like [__fish_argcomplete_complete](https://github.com/fish-shell/fish-shell/blob/master/share/functions/__fish_argcomplete_complete.fish) does.

(technically the additional file descriptors there aren't necessary, but they're unlikely to hurt)

Try something like

```fish
complete -c ibmcloud -f -a '(__fish_argcomplete_complete ibmcloud --generate-bash-completion)'
```

faho: Oh god, it's using half the bash completion system.

Okay, you need to arrange for the second to run `ibmcloud iam --generate-bash-completion`.

You might want to *try* to run `__fish_argcomplete_complete (commandline -opc) --generate-bash-completion`, or it needs to be more complicated.

faho:
Thanks, merged!

PR #10005, opened 2023-09-06
reader: Use existing search string when opening the history pager

faho:
I sometimes find myself doing something like this:

- Look for a commandline that includes "echo" (as an example)
- Type echo, press up a few times
- I can't immediately find what I'm looking for
- Press ctrl-r to open up the history pager
- It uses the current commandline as the search string, so now I'm looking for "echo foobar"

This makes it so if the search string already is in use, that's what the history-pager picks as the initial search string.

## TODOs:

I am entirely fine with shelving this if it gets in the way of anyone's porting. On the other hand it's pretty trivial.


krobelus (approved):
I think I hit this as well. Great call.

PR #10007, opened 2023-09-08
add gcc completion for link lib

asukaminato0721:
## Description

add completion for lib.

ref for `-ldl` https://stackoverflow.com/questions/19146525/meaning-of-library-dl-in-gcc

Fixes issue #

## TODOs:

### DIFF
--- a/share/completions/gcc.fish
+++ b/share/completions/gcc.fish
@@ -482,9 +482,7 @@ complete -c gcc -s c -d 'Compile or assemble the source files, but do not link.'
 complete -c gcc -s S -d 'Stop after the stage of compilation proper; do not assemble'
 complete -c gcc -s E -d 'Stop after the preprocessing stage; do not run the compiler proper'
 complete -c gcc -o llibrary -d 'Search the library named library when linking'
-for i in (ls /usr/lib/ | string match -r -g "^lib(.*?)\.so.*")
-	complete -c gcc -o "l$i"
-end
+complete -c gcc -a '-l(path basename /usr/lib/lib*.so* | string match -r -g "^lib(.*?)\.so.*")'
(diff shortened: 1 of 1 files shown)


share/completions/gcc.fish
```diff
@@ -482,7 +482,11 @@ complete -c gcc -s c -d 'Compile or assemble the source files, but do not link.'
 complete -c gcc -s S -d 'Stop after the stage of compilation proper; do not assemble'
 complete -c gcc -s E -d 'Stop after the preprocessing stage; do not run the compiler proper'
 complete -c gcc -o llibrary -d 'Search the library named library when linking'
+for i in (ls /usr/lib/ | string match -r -g "^lib(.*?)\.so.*")
```
faho, line 485: Calling ls here is awkward for a variety of reasons - it's an external command that may or may not escape its output or be aliased to any number of other things.

Use

```fish
for i in (path basename /usr/lib/lib*.so* | string match -r -g "^lib(.*?)\.so.*")
```

Which is also easily extendable to other directories - I would assume that /lib and the "lib64" variants are also of interest

faho, line 485: This has a *bunch* of problems.

For one: 

Calling ls here is awkward for a variety of reasons - it's an external command that may or may not escape its output or be aliased to any number of other things.

Try

```fish
path basename /usr/lib/lib*.so* | string match -r -g "^lib(.*?)\.so.*"
```

Which is also easily extendable to other directories - I would assume that /lib and the "lib64" variants are also of interest

For another:

This generates all libraries at the start, once the gcc completions are *loaded*.

What you want is instead to make this an *argument* so it's generated when needed and e.g. adding libraries later works without having to restart the shell:

```fish
complete -c gcc -a '-l(path ...)'

faho:
Thanks!

PR #10008, opened 2023-09-09
added support for fish_sequence_key_delay_ms to set how long to wait between sequence key presses

yanshay:
#### This is a back-port of PR #9926 (which is on Master) to 3.7.0.

Reminder what it is about:

This resolves issue https://github.com/fish-shell/fish-shell/issues/7401

When binding key sequences, currently, fish waits indefinitely for sequence to complete. So if one wants to press only a single character which is a beginning of a sequence, that character will not appear. To make it appear need to press another character, not in the sequence and then backspace. Not very reasonable.
For example, if using 'kj' to exit insert mode in fish vi mode, if one presses 'k' it will not show until another key is pressed, so to type 'k' at end of command need to press 'k', then another character and then backspace. Not to speak that to enter kj as characters its even more difficult.

The solution introduced here is to set fish_sequence_key_delay_ms variable (along the lines of fish_escape_delay_ms) which is the time to wait for the next key in a sequence. Once that time passes the key will be output as usual. By default the behavior is as today and setting the key changes the behavior.

I wrote the code and tested it manually. If you expect it to enter future versions I will update documentation and changelog as well, and will try to add tests for the new behavior.

I also think it's worth considering making this the default behavior using some reasonable value.


faho:
Thanks again!

I've finally made some headway on 3.7.0 so I finally got to this, sorry for the delay!

PR #10010, opened 2023-09-10
Add completions for pnpm command

u1aryz:
## Description

This PR introduces completion support for pnpm in the fish shell. The completion script is based on the one provided by the pnpm/tabtab repository.
~https://github.com/pnpm/tabtab/blob/fish-script/lib/scripts/fish.sh~
https://github.com/pnpm/tabtab/blob/main/lib/scripts/fish.sh

Please review and let me know if any changes are required.

## TODOs:

### DIFF
--- a/share/completions/pnpm.fish
+++ b/share/completions/pnpm.fish
@@ -1,0 +1,20 @@
+function _pnpm_completion
+  set cmd (commandline -o)
+  set cursor (commandline -C)
+  set words (count $cmd)
+
+  set completions (eval env DEBUG=\"" \"" COMP_CWORD=\""$words\"" COMP_LINE=\""$cmd \"" COMP_POINT=\""$cursor\"" pnpm completion -- $cmd)
+
+  if [ "$completions" = "__tabtab_complete_files__" ]
+    set -l matches (commandline -ct)*
+    if [ -n "$matches" ]
+      __fish_complete_path (commandline -ct)
+    end
+  else
+    for completion in $completions
+      echo -e $completion
+    end
+  end
+end
+
+complete -f -d 'pnpm' -c pnpm -a "(_pnpm_completion)"


share/completions/pnpm.fish
```diff
@@ -0,0 +1,20 @@
+function _pnpm_completion
+  set cmd (commandline -o)
```
faho, line 2: All of these variables should explicitly be scoped as *local* variables, or you would override any global or universal variable by that name:

```fish
set -l cmd (commandline -o)
```

share/completions/pnpm.fish
```diff
@@ -0,0 +1,20 @@
...
+  set cursor (commandline -C)
+  set words (count $cmd)
+
+  set completions (eval env DEBUG=\"" \"" COMP_CWORD=\""$words\"" COMP_LINE=\""$cmd \"" COMP_POINT=\""$cursor\"" pnpm completion -- $cmd)
```
faho, line 6: This `eval` gives me the heeby-jeebies, and this `env` is unnecessary.

I also don't like that this sets a variable called "DEBUG", and I'm not sure you *need* COMP_CWORD - our bash-compatible completion thing is [__fish_argcomplete_complete](https://github.com/fish-shell/fish-shell/blob/master/share/functions/__fish_argcomplete_complete.fish), which doesn't set it (but could!).

share/completions/pnpm.fish
```diff
@@ -0,0 +1,20 @@
...
+
+  if [ "$completions" = "__tabtab_complete_files__" ]
+    set -l matches (commandline -ct)*
+    if [ -n "$matches" ]
+      __fish_complete_path (commandline -ct)
+    end
+  else
+    for completion in $completions
+      echo -e $completion
```
faho, line 15: That `-e` is suspect - this will re-evaluate backslash-escapes you got in the output of `pnpm completion`!

I doubt pnpm expects its output to be treated like that.

faho (changes requested):

share/completions/pnpm.fish
```diff
@@ -0,0 +1,20 @@
+function _pnpm_completion
+  set cmd (commandline -o)
+  set cursor (commandline -C)
+  set words (count $cmd)
+
+  set completions (eval env DEBUG=\"" \"" COMP_CWORD=\""$words\"" COMP_LINE=\""$cmd \"" COMP_POINT=\""$cursor\"" pnpm completion -- $cmd)
+
+  if [ "$completions" = "__tabtab_complete_files__" ]
```
faho, line 8: Where do you get this from? It's not in the original script?

u1aryz: These conditions are based on the following.
https://github.com/pnpm/tabtab/commit/63aef480af1a0aee23f82e86cc167a6d2dc22eac

u1aryz (commented):
Thank you for the review.

u1aryz:
I'll come back out.

PR #10014, opened 2023-09-11
Improve vim 'word' movements.

AlexSWall:
## Description

I've improved the behaviour of vim mode's movement:
- `ge` and `gE` now go to the end of the previous word.
- Makes word movements in visual mode the same as in normal mode.


faho:
> `bind -s --preset gE forward-word backward-word backward-bigword forward-bigword backward-char`

That's a ginormous hack that kinda just shows our word definition is out of step with vi(m). I have no idea how that behaves because it's a weird back and forth that somehow seems to move.

I don't think it's a good idea to add this, instead I would like to figure out how these differ and try to add a way to get them included.

This needs a bigger rework - see #10016

PR #10016, opened 2023-09-12
Adding forward-*next*-[big]word movements.

AlexSWall:
## Description

Fish's movements, particularly in vim mode, are hindered by the lack of a movement that allows one to move to the start of the next word.

Currently, the shortcomings of fish's `forward-word` and `forward-bigword` can be easily seen in the case of the following, where both `forward-word` and `forward-bigword` take one from `1` to `2` and then `2` to `3`.
```
foo     bar      baz
^   ^       ^
1   2       3
```
Instead, we want the behaviour of 
```
foo     bar      baz
^       ^        ^
1       2        3
```

We want to leave `forward-word` and `forward-bigword` as they are as they're currently used plenty, but we should add a new `forward-next-word` and `forward-next-bigword` (alternative names welcome) to give us this functionality, and then I'll add a separate PR for integrating these into vim mode.

The case of having multiple spaces between arguments may seem rather specific, but I believe the functionality can actually be quite important to have reliable vim keybindings, as they otherwise include fragile workarounds to get this same behaviour.

At this stage, I want a thumbs-up from an owner that they're happy with this addition, and/or receive suggestions for changes such that such an owner _would_ be, and then I will work on documentation, testing, and the like.

### DIFF
--- a/src/input.cpp
+++ b/src/input.cpp
@@ -126,6 +126,8 @@ static constexpr const input_function_metadata_t input_function_metadata[] = {
     {L"forward-char", readline_cmd_t::forward_char},
     {L"forward-jump", readline_cmd_t::forward_jump},
     {L"forward-jump-till", readline_cmd_t::forward_jump_till},
+    {L"forward-next-bigword", readline_cmd_t::forward_next_bigword},
+    {L"forward-next-word", readline_cmd_t::forward_next_word},
     {L"forward-single-char", readline_cmd_t::forward_single_char},
     {L"forward-word", readline_cmd_t::forward_word},
     {L"history-pager", readline_cmd_t::history_pager},
--- a/src/input_common.h
+++ b/src/input_common.h
@@ -22,6 +22,8 @@ enum class readline_cmd_t {
     backward_word,
     forward_bigword,
     backward_bigword,
+    forward_next_word,
+    forward_next_bigword,
     nextd_or_forward_word,
     prevd_or_backward_word,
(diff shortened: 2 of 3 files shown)


faho:
As it stands, I am going to reject this.

Not because I don't believe the bindings are valuable to have, or because I think the code is of poor quality.

Rather it's because it's too ad-hoc. We have 79 bind functions, and adding more movement like this makes it easy to get into a combinatorial explosion of sorts - what if I want to skip whitespace after a *word*, not a bigword? Or a *char*? Or a path-component? What if I want to move by that and also remove the text? That's `(forward + backward) * (word + bigword + char + path-component) * (move + kill)` == 16 new bind functions, that we would be stuck maintaining. And that's if we don't add any more movement styles like that.

What we need is more involved than that: We need to figure out

1. How vi's word definitions work, and how they differ from emacs' and fish's (and if vim differs from the original vi, we want to take the vim way instead)
2. How to implement that with the fewest number of bind functions possible

In this specific case, a "\*-whitespace" family would be more useful because you can combine it with the rest:

```fish
bind -M default E forward-bigword forward-whitespace
bind \cg backward-kill-bigword backward-kill-whitespace
```

And that's not necessarily the final answer either - it needs more research than that.

(there's also the question if we want to make these movements token-aware, so e.g. forward-whitespace only moves in-between tokens, for separating spaces like you have up there, but not e.g. in a quoted argument: `"foo   bar"`)

PR #10020, opened 2023-09-15
Fix out of scope opt variable

mathijshenquet:
## Description

I had a weird bug on a fresh fish install and I traced it back to this rather weird piece of code. The $opt variable is not set when the if at L29 fails and so just contained random junk from an earlier function call. Moreover, with every subsequent call of of ls it would grow an extra `-F`! I believe the following is the intended meaning of the author.

You can look at the blame of linet 53 and 66 to see that this state of things was caused by an old refactor of the ls function.

Fixes issue: if $opt is filled with some junk before calling ls it will result in unexpected behavior

~> set fish_trace on
~> set opt test
> set opt test
~> ls
> ls
(...)
--> ls --color=auto test -F
ls: cannot access 'test': No such file or directory

### DIFF
--- a/share/functions/ls.fish
+++ b/share/functions/ls.fish
@@ -50,7 +50,6 @@ function ls --description "List contents of directory"
     __fish_set_lscolors
 
     isatty stdout
-    and set -a opt -F
 
     # Terminal.app doesn't set $COLORTERM or $CLICOLOR,
     # but the new FreeBSD ls requires either to be set,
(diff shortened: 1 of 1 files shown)


share/functions/ls.fish
```diff
@@ -50,7 +50,6 @@ function ls --description "List contents of directory"
     __fish_set_lscolors
 
     isatty stdout
```
faho, line 52: The intention is to apply -F only if the output goes to the terminal.

The simplest fix is to set $opt before:

```fish
set -l opt
isatty stdout
and set opt -F
```

Note that we have some cases above where we might apply "-F" even if the output isn't a tty, those should be removed. It's either when we have `colorls` or when `ls` doesn't support any of the color flags.

faho (changes requested):

faho:
Thanks!

PR #10024, opened 2023-09-23
 remove unnecessary heap allocation

thechampagne:
## Description

Talk about your changes here.

Fixes issue #

## TODOs:

### DIFF
--- a/fish-rust/src/common.rs
+++ b/fish-rust/src/common.rs
@@ -1749,8 +1749,7 @@ pub fn fish_reserved_codepoint(c: char) -> bool {
 pub fn redirect_tty_output() {
     unsafe {
         let mut t: libc::termios = mem::zeroed();
-        let s = CString::new("/dev/null").unwrap();
-        let fd = libc::open(s.as_ptr(), O_WRONLY);
+        let fd = libc::open("/dev/null\0".as_ptr() as _, O_WRONLY);
         assert!(fd != -1, "Could not open /dev/null!");
         for stdfd in [STDIN_FILENO, STDOUT_FILENO, STDERR_FILENO] {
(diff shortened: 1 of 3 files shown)


Xiretza:
Ignoring the fact that this doesn't even build, having to explicitly specify the NUL byte is a footgun.

Most of these allocations either happen *at most once* per fish invocation, or only happen in conjunction with some job control syscalls/signals, which are *way more* overhead than a single small allocation. This kind of reduction in readability and safety would need some seriously convincing benchmark results.

faho:
Yeah, I'm rejecting this as not worth it. It's an irrelevant performance optimization and usage pessimization.

Sorry!

thechampagne:
> Ignoring that this doesn't build, having to explicitly specify the NUL byte is a footgun.

@Xiretza why specifying the NULL byte is a footgun ?

faho:
>why specifying the NULL byte is a footgun ?

@thechampagne it's a footgun (i.e. an awkward trap to stumble over) because it is easy to forget. Few things require you to spell out the NUL for typical strings - rust doesn't, C++ doesn't.

And if you do forget it it would break in interesting ways.

That's why we're going for the more idiomatic, easier to read version that we already have, especially because the difference is not going to be measurable.

Performance improvements should come with benchmarks so we can see what impact they have, and if the impact was large enough we would also take a bit of a hit in readability - but that's not the case here.

thechampagne:
I would force myself to remember it instead of using useless runtime code.

faho:
And that is a trade-off we won't make.

We'll lean towards maintainability in favor of an increase in performance that's so small that it's lost in the statistical noise.

Anyway: I do believe this has run its course. Thank you for your interest in fish, I am sorry that I did not accept your pull request. If you do want to work on performance, which is certainly a worthwhile task and something that could much improve fish, please do add a benchmark.

PR #10029, opened 2023-09-27
completions/pkill: use locals.

charlottia:
## Description

Use `read -l` to not (over)write `number` and `name` in the shell context.

## TODOs:

I don't know that I feel any of the above is necessary for a change of this minority, but I'm happy to be guided otherwise.


faho:
Nice catch, thanks!

charlottia:
Thank you! I was relying on a different local `name` (set by `nix shell` etc.) in my `fish_prompt`, and every now and then would notice it was replaced with `USR2` :D

PR #10038, opened 2023-09-29
completion(loginctl): fix sessions with ssh or other states

NextAlone:
## Description

Just like SESSION 7

image

Fixes issue #

## TODOs:

### DIFF
--- a/share/completions/loginctl.fish
+++ b/share/completions/loginctl.fish
@@ -30,7 +30,7 @@ complete -c loginctl -f -n "not $seen $commands" -l version -d "Show package ver
 
 
 function __fish_loginctl_list_sessions
-    loginctl list-sessions --no-legend --no-pager --output=short | string replace -r '(\d+) \d+ (\S+) \S+ (\S+) .*' '$1\t$2 at $3'
+    loginctl list-sessions --no-legend --no-pager --output=short | string replace -r '(\d+) \d+ (\S+) (\S*|\s+) (\S+) .*' '$1\t$2 at $4'
 end
 
 
--- a/share/completions/loginctl.fish
+++ b/share/completions/loginctl.fish
@@ -30,7 +30,7 @@ complete -c loginctl -f -n "not $seen $commands" -l version -d "Show package ver
 
 
 function __fish_loginctl_list_sessions
-    loginctl list-sessions --no-legend --no-pager --output=short | string replace -r '(\d+) \d+ (\S+) (\S*|\s+) (\S+) .*' '$1\t$2 at $4'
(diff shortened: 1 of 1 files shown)


share/completions/loginctl.fish
```diff
@@ -30,7 +30,7 @@ complete -c loginctl -f -n "not $seen $commands" -l version -d "Show package ver
 
 
 function __fish_loginctl_list_sessions
-    loginctl list-sessions --no-legend --no-pager --output=short | string replace -r '(\d+) \d+ (\S+) \S+ (\S+) .*' '$1\t$2 at $3'
+    loginctl list-sessions --no-legend --no-pager --output=short | string replace -r '(\d+) \d+ (\S+) (\S*|\s+) (\S+) .*' '$1\t$2 at $4'
```
faho, line 33: Neither of these work in one test of mine. The output looks like

```
322 1000 alfa   pts/0
```

Where it goes wrong is the ` .*` at the end - there is no space following that `pts/0`.

The easy fix appears to be to just not require that space - `(\S+).*`. The `\S+` should be greedy, so the `.*` can only ever start with something that isn't matched by the `\S`.

Tho it is early in the morning here, so maybe I'm just not awake enough to deal with regexen.

share/completions/loginctl.fish
```diff
@@ -30,7 +30,7 @@ complete -c loginctl -f -n "not $seen $commands" -l version -d "Show package ver
 
 
 function __fish_loginctl_list_sessions
-    loginctl list-sessions --no-legend --no-pager --output=short | string replace -r '(\d+) \d+ (\S+) (\S*|\s+) (\S+) .*' '$1\t$2 at $4'
+    loginctl list-sessions --no-legend --no-pager --output=short | string replace -r '(\d+)\s+\d+\s+(\w+)\s+(seat\d+\s+)?(tty\d+|pts\/\d+).*' '$1\t$2 at $4'
```
faho, line 33: 1. Those are actual words you're matching, they might very well be localized. If you did want to match them you would have to pass a locale to `loginctl` - `LC_ALL=C loginctl ...`.
2. `\w` ("word"-characters) is not a very useful class to use here. This output will have useful information *every line* (and if it didn't you would have to throw it away). It should be `\S` ("not whitespace").
3. What if there is something that's not "tty" or "pts"?

In essence this runs the risk of under-matching (i.e. not matching input that should be matched), and hence would let untransformed input through.

I like the earlier version more than this.

NextAlone: Got that.

NextAlone: Columns are aligned with spaces, this regex looks better.

faho:
Nice, thanks!

PR #10042, opened 2023-10-01
docs: test drive fish section to Readme

strasdat:
## Description

For somebody like me who like to use fish in their daily workflow, but also have a convenient fallback to a more traditional shell, I converged to the trick described in this PR (adding "fish" to last line of ".bashrc").

I'm not sure this is a common workflow, or whether it has serious shortcomings. If this is not the right place for such a workflow recommendation/proposal, feel free to just close the PR!

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -112,6 +112,17 @@ The following optional features also have specific requirements:
 -  ``colorls`` is used, if installed, to add color when running ``ls`` on platforms
    that do not have color support (such as OpenBSD)
 
+
+Test driving fish
+~~~~~~~~~~~~~~~~~
+
+If you want to try fish but want to have a fallback to your old workflow, you can use the following trick:
+Let assume without loss of generality that your default shell is bash. Just add``fish`` at the end of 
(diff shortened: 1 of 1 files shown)


faho:
This has a few downsides, in that it'll always happen when bash is started. The best that e.g. [the arch wiki](https://wiki.archlinux.org/title/Fish) has come up with is a great deal more complicated:

```sh
if [[ $(ps --no-header --pid=$PPID --format=comm) != "fish" && -z ${BASH_EXECUTION_STRING} ]]
then
	shopt -q login_shell && LOGIN_OPTION='--login' || LOGIN_OPTION=''
	exec fish $LOGIN_OPTION
fi
```

I do not believe that's worthy of being added to the README, and this isn't really what the README is *for*.

It's really just to explain how to build fish and to point people to our [actual documentation](https://fishshell.com/docs/current/index.html).

So I'm rejecting this, sorry.

strasdat:
Thanks for the pointer to the arch wiki!

PR #10044, opened 2023-10-03
Add a clear-screen bind function to clear the screen

faho:
## Description

This adds a bind function called `clear-screen` and binds it to ctrl+l by default, removing the rather awkward invocation of the `clear` command.

What that does is:

1. Print the "clear_screen" escape sequence, if possible
2. Redraw with the old prompt
3. Reexecute the prompt
4. Repaint again

This makes for a smooth experience removing any flicker, and still leaves an up-to-date prompt (eventually, once it comes through).

I've tested this with TERM=dumb (it'll do nothing) and with an awkwardly slow prompt - it will show the old prompt, won't react to input and then update the prompt and the commandline.

## TODOs:

### DIFF
--- a/doc_src/cmds/bind.rst
+++ b/doc_src/cmds/bind.rst
@@ -142,6 +142,9 @@ The following special input functions are available:
 ``capitalize-word``
     make the current word begin with a capital letter
 
+``clear-screen``
+    clears the screen and redraws the prompt. if the terminal doesn't support clearing the screen it does nothing.
+
 ``complete``
     guess the remainder of the current token
 
--- a/doc_src/cmds/bind.rst
+++ b/doc_src/cmds/bind.rst
@@ -143,7 +143,7 @@ The following special input functions are available:
     make the current word begin with a capital letter
 
 ``clear-screen``
-    clears the screen and redraws the prompt. if the terminal doesn't support clearing the screen it does nothing.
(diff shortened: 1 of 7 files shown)


faho:
I'm reasonably sure the test failure is down to $TERM/terminfo being wrong on these machines, and so it doesn't do anything.

I'm not sure why `command clear` would work in that case, I'll have to check.

faho:
Ah, this was a *crash* - `clear_screen` was NULL.

This should really be shown in pexpect, also the test that failed doesn't really check clearing because the pexpect tests are run with TERM=dumb.

ridiculousfish:
Seems fine to me - though I use the terminal's clear function which also clears scrollback.

faho:
>Seems fine to me - though I use the terminal's clear function which also clears scrollback.

That is of course a valid choice.

We already removed scrollback clear from `clear`'s output because it makes clearing *destructive*, which seems like a surprising and therefore bad default. So there's no change related to that here.

It would be *possible* to add a "repaint-old" function that repaints without reexecuting. This would allow you to add a fast repaint after your clear by doing `bind \cl 'clear; commandline -f repaint-old repaint'`. It does of course make it possible to accidentally carry around a stale prompt. Or a function that does "repaint-old repaint", by first redrawing the cached information and then reexecuting - this would of course send a bit more in case both need to be done and is typically unnecessary.

So I chose this version because skipping the awkward `clear | string replace` was on my agenda anyway and this is simple and impossible to misuse.

PR #10049, opened 2023-10-07
Use git diff instead of git status if dirty and not untracked

notEvil:
## Description

I found `git diff --quiet` to be significantly faster than `git status`. The difference is noticable on high latency storage (e.g. network share over internet).

## TODOs:

### DIFF
--- a/share/functions/fish_git_prompt.fish
+++ b/share/functions/fish_git_prompt.fish
@@ -290,6 +290,9 @@ function fish_git_prompt --description "Prompt function for Git"
                 # Only untracked, ls-files is faster.
                 command git -c core.fsmonitor= ls-files --others --exclude-standard --directory --no-empty-directory --error-unmatch -- :/ >/dev/null 2>&1
                 and set untrackedfiles 1
+            else if test "$dirty" = true; and not test "$untracked" = true
+                # Only dirty, diff is faster
+                command git -c core.fsmonitor= diff --quiet HEAD; or set dirtystate 1
             else if test "$dirty" = true
                 # With both dirty and untracked, git status is ~10% faster.
                 # With just dirty, it's ~20%.


faho:
Tests are failing:

> fatal: ambiguous argument 'HEAD': unknown revision or path not in the working tree.

I *think* you can just leave off the `HEAD`:

```fish
command git -c core.fsmonitor= diff --quiet ; or set dirtystate 1
```

notEvil:
Thanks! `HEAD` is required because `diff` would otherwise compare against cache/staging. Hope I didn't miss another edge case.

faho:
> Thanks! HEAD is required because diff would otherwise compare against cache/staging

If I'm not completely missing something right now that's correct?

"dirtystate" is true if there is any file that matches the regex `'^.[ACDMR]'` in the `git status --porcelain` output.

That means: Any file that has an *unstaged change*. And an "unstaged" change is a change relative to the staging area. This would be `git diff --quiet`.

The "stagedstate" meanwhile must match `'^[ACDMR].'` - which is any file that has a *staged* change. This would be `git diff --quiet --staged`.

Of course that means that you run the latter even if the former didn't show changes.

This also explains the [current test failures](https://github.com/fish-shell/fish-shell/actions/runs/6441952908/job/17492270674):

> (newbranch) <= does not match CHECK on line 103: (newbranch +)

This should be "staged" (`+`) and it says it's clean. You're not checking the `--staged` diff because the first `git diff` didn't show changes (i.e. return 1).

> (newbranch +*) <= does not match CHECK on line 147: (newbranch +)

This should show only staged changes, it instead claims it's staged and dirty (`*`).

notEvil:
Sry to cause so much effort! You are right in all points and I'm currently trying to setup local tests to not waste CI time for the next and hopefully final update.

The current state now is
```fish
                command git -c core.fsmonitor= diff --quiet
                test $status -eq 1; and set dirtystate 1

                if test -n "$sha"
                    command git -c core.fsmonitor= diff --quiet --staged
                    test $status -eq 1; and set stagedstate 1
                else
                    set invalidstate 1
                end
```

notEvil:
Unfortunately my tests yield bad news. With a fresh repository, 1000 empty files at top level, all commited, no dirt, mounted using sshfs (Arch linux) over cell network:
- `git status --porcelain` took ~4.5 seconds most of the time and sometimes ~27 seconds
- `git diff --quiet` took ~27 seconds every single time

So `git status` seems to be smarter and therefore faster.

What motivated this PR was the observation that `git diff` was 3x faster than `git status` locally (small repository, sshfs over ethernet; ~30ms vs ~100ms) and the assumption that it must be faster since it stops at first difference.

faho:
>Unfortunately my tests yield bad news.

That is indeed bad news.

Okay, seems like there's nothing here, so I'm closing this.

Thanks for checking tho!

PR #10052, opened 2023-10-10
 wildcard: Rationalize file/command completions

faho:
## Description

Alternative to #10050. **NOTE** The rust version is not up-to-date, it's #10050. I based it on that, I will of course adjust that before merging if we do decide that this is an acceptable approach.

This keeps the entry_t as long as possible, and asks it, so especially
on systems with working d_type we can get by without a single stat in
most cases.

Then it guts file_get_desc, because that is only used for command
completions - we have been disabling file descriptions for *years*,
and so this is never called there.

That means we have no need to print descriptions about e.g. broken symlinks, because those are not executable.

Put together, what this means is that we, in most cases, only do
an *access(2)* call instead of a stat, because that might be checking
more permissions.

So we have the following constellations:

- If we have d_type:
  - We need a stat() for every _symlink_ to get the type (e.g. dir or regular)
    (this is for most symlinks, if we want to know if it's a dir or executable)
  - We need an access() for every file for executables
- If we do not have d_type:
  - We need a stat() for every file
  - We need an lstat() for every file if we do descriptions
    (i.e. just for command completion)
  - We need an access() for every file for executables

As opposed to the current way, where every file gets one lstat whether
with d_type or not, and an additional stat() for links, *and* an
access. And this is on top of the stat() that the entry might already have done - even without d_type, we save that.

So we go from two syscalls to one for executables, while improving accuracy compared to #10050.

## Numbers

All acquired via `strace -c -e %file fish --no-config -c $thething`, counting "file" syscalls like access and the stat family. This is an up-to-date linux system on ext4, *with* d_type.

`complete -C "ls ~/complete_test/"`, where complete_test is a directory with 10k directories and 10k files:

- Current pre-3.7.0 and #10050: 20k stat
- This: 0 stat (well, 32, but in comparison it's basically 0)

`complete -C ""` (command completion), with 2870 files in $PATH, 2865 of which are executable files and 407 are links:

- Current pre-3.7.0: 6239 "file" syscalls (3331 newfstatat, 2869 access)
- #10050: 3382, 3328 of which are newfstatat (3.6.1 would have similar numbers)
- This: 3361, 2868 access + 442 newstatat (remember, 407 links)

I will have to see if I can get numbers for without d_type. The gains there are that we can reuse the stat from is_dir() if we are completing directories specifically, and that we can skip the lstat() for non-executables.

## TODOs:

### DIFF
--- a/fish-rust/src/wildcard.rs
+++ b/fish-rust/src/wildcard.rs
@@ -426,17 +426,31 @@ fn wildcard_test_flags_then_complete(
         stat = None;
     }
 
-    let (file_size, is_directory, is_executable) = if let Some(Ok(md)) = &stat {
-        (md.len(), md.is_dir(), md.is_file())
+    let (file_size, is_directory, is_regular_file, perms) = if let Some(Ok(md)) = &stat {
+        (
+            md.len(),
+            md.is_dir(),
+            md.is_file(),
+            md.permissions().mode() as mode_t,
+        )
     } else {
-        (0, false, false)
+        (0, false, false, 0)
     };
 
     if need_directory && !is_directory {
         return false;
     }
 
-    if executables_only && (!is_executable || waccess(filepath, X_OK) != 0) {
+    // If the file has all executable bits set, we assume it is executable.
+    // This does not account for:
+    // 1. Capabilities (but neither does access!)
+    // 2. noexec filesystems (but then why add those to $PATH?)
+    // 3. MAC (SELinux etc)
+    // 4. ACLs
+    // Since this is for completions, that's fine.
+    let assume_executable = perms & (S_IXUSR | S_IXGRP | S_IXOTH) == (S_IXUSR | S_IXGRP | S_IXOTH);
+    if executables_only && !assume_executable && (!is_regular_file || waccess(filepath, X_OK) != 0)
+    {
         return false;
     }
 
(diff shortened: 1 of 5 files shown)


src/wildcard.cpp
```diff
@@ -238,55 +238,28 @@ wildcard_result_t wildcard_complete(const wcstring &str, const wchar_t *wc,
...
-        return COMPLETE_BLOCK_DESC;
-    } else if (S_ISFIFO(buf.st_mode)) {
-        return COMPLETE_FIFO_DESC;
-    } else if (S_ISSOCK(buf.st_mode)) {
-        return COMPLETE_SOCKET_DESC;
-    } else if (S_ISDIR(buf.st_mode)) {
+/// \param is_dir Whether the file is a directory or not (might be behind a link)
+/// \param is_link Whether it's a link (that might point to a directory)
+/// \param definitely_executable Whether we know that it is executable, or don't know
+static const wchar_t *file_get_desc(const wcstring &filename, bool is_dir,
```
ridiculousfish, line 244: This whole function is so much nicer now!

src/wildcard.cpp
```diff
@@ -238,55 +238,28 @@ wildcard_result_t wildcard_complete(const wcstring &str, const wchar_t *wc,
...
+                                    bool is_link, bool definitely_executable) {
+    if (is_link) {
```
ridiculousfish, line 246: This whole function is so much nicer now!

ridiculousfish (approved):
This is fantastic. I love it.

fish-rust/src/wildcard.rs
```diff
@@ -379,6 +379,13 @@ fn wildcard_test_flags_then_complete(
         return false;
     }
 
+    if executables_only
+        && is_windows_subsystem_for_linux()
```
faho, line 383: (@mqudsi this should really also be a thing on WSL2)

mqudsi (approved):
LGTM! Nice work, much cleaner.

I agree on the WSL2 bit; I can open a PR with the unification of WSL1 and WSL2 stuff. Maybe there's nothing that it won't hurt to do under WSL2.

PR #10054, opened 2023-10-13
Simpler error messages

faho:
## Description

These are three fairly independent commits with a common goal:

To make our error messages better.

The first adds some extra cases where we used to print "Unknown error while evaluating command substitution". The messages it adds are pretty redundant, but not adding an error there would require some extra scaffolding. Or we could pull these errors apart, which requires separate status codes. I would like to add this one to 3.7.0 as-is, and think it's fairly uncontroversial.

The second puts the "called on" on the same line as the block description in stack traces. So instead of

```
in command substitution
        called on line 14 of file foo.fish
```

you get

```
in command substitution called on line 14 of file foo.fish
```

This makes some possibly very large messages a bunch smaller.

One potential issue is that this also changes the output of `status print-stack-trace`, which people could potentially rely on. Last [time we did that](7095de628c8f05b2142827a74511cd1c979c5c90) I can't remember anyone complaining.

It also changes all these messages, which means the translations would need to be fixed (which should be fine to just do because it's removing a `\n`).

The third one removes a potential redundant "fish:" line we get in some cases, because we add an error without a message to add context. It's a bit hacky, but not *too* odious I think. Mostly it would be easier if `describe_with_prefix` didn't have to add error messages for bare variable assignments or andor_in_pipeline, because then the outside could just drive it, and decide to set the prefix to empty if there is no message.

There's some slight concern that there are errors where it's unclear they are from fish - if so, I'd argue that they should just be improved.

Here's a demonstration of how messages change:

`fish -c 'nonexistent'` changes from

```
fish: Unknown command: nonexistent
fish: 
nonexistent
^~~~~~~~~~^
```

to

```
fish: Unknown command: nonexistent
nonexistent
^~~~~~~~~~^
```

And a nonsense script I wrote changes from:

```
fish: Unknown command: nonexistent
foo.fish (line 4): 
    if nonexistent
       ^~~~~~~~~~^
in function 'this-errors' with arguments '2'
	called on line 11 of file foo.fish
in function 'call-this'
	called on line 1 of file foo.fish
in command substitution
	called on line 14 of file foo.fish
foo.fish (line 1): for: status: cannot overwrite read-only variable
for status in 1 2 3; echo $status; end
    ^~~~~^
in command substitution
	called on line 7 of file foo.fish
in function 'this-errors' with arguments '2'
	called on line 11 of file foo.fish
in function 'call-this'
	called on line 1 of file foo.fish
in command substitution
	called on line 14 of file foo.fish
foo.fish (line 7): Unknown error while evaluating command substitution
    echo (for status in 1 2 3; echo $status; end)
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
in function 'this-errors' with arguments '2'
	called on line 11 of file foo.fish
in function 'call-this'
	called on line 1 of file foo.fish
in command substitution
	called on line 14 of file foo.fish
```

to

```
fish: Unknown command: nonexistent
foo.fish (line 4): 
    if nonexistent
       ^~~~~~~~~~^
in function 'this-errors' with arguments '2' called on line 11 of file foo.fish
in function 'call-this' called on line 1 of file foo.fish
in command substitution called on line 14 of file foo.fish
foo.fish (line 1): for: status: cannot overwrite read-only variable
for status in 1 2 3; echo $status; end
    ^~~~~^
in command substitution called on line 7 of file foo.fish
in function 'this-errors' with arguments '2' called on line 11 of file foo.fish
in function 'call-this' called on line 1 of file foo.fish
in command substitution called on line 14 of file foo.fish
foo.fish (line 7): Invalid arguments
    echo (for status in 1 2 3; echo $status; end)
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
in function 'this-errors' with arguments '2' called on line 11 of file foo.fish
in function 'call-this' called on line 1 of file foo.fish
in command substitution called on line 14 of file foo.fish
```

A 50% reduction in number of lines - from 30 lines to 20, for a single wrong variable name and nonexistent command. The script itself is 15 lines long. Or see the stack-overflow test, where this removes 130 lines.

It would also be possible to reword this, while we're at it (I've seen things phrase line positions as "file:line", so "called on foo.fish:14").

## TODOs:

I would be entirely fine if we decided to go with just the first one, or if we decided to put the rest off for 4.0, or if the rest gets in the way of someone's porting and I need to write it again later, or if we decide to go for a cleaner solution.


src/parser.cpp
```diff
@@ -249,27 +248,22 @@ static void append_block_description_to_stack_trace(const parser_t &parser, cons
...
                           user_presentable_path(*source_dest, parser.vars()).c_str());
-            print_call_site = true;
             break;
         }
         case block_type_t::event: {
             assert(b.event && "Should have an event");
             wcstring description = *event_get_desc(parser, **b.event);
-            append_format(trace, _(L"in event handler: %ls\n"), description.c_str());
-            print_call_site = true;
+            append_format(trace, _(L"in event handler: %ls"), description.c_str());
```
krobelus, line 266: The only potential downside is it could look worse if the argument strings are overlong.
I think that's an edge case so probably this is fine and overall a great change.

---

As a programmer, the stack traces are a bit confusing to me, if I see something like

	function foo file bar.fish:3
	function bar ...

I expect that line 3 in bar.fish is inside function `foo`. But that's a separate concern

Also I expect the ordering file,line,function,arguments  instead of function,arguments,file,line (which would alleviate the concern about long arguments)

---

> I would be entirely fine if we decided to go with just the first one, or if we decided to put the rest off for 4.0, or if the rest gets in the way of someone's porting and I need to write it again later, or if we decide to go for a cleaner solution.

I don't mind merging things like this first (it's small and I don't expect too many of them).

faho: > The only potential downside is it could look worse if the argument strings are overlong.

There are a bunch of ways to solve this:

1. Truncate/ellipsize the arguments (not great)
2. Move the arguments onto another line if they are very long

Something like

```
fish: Unknown command: nonexistent
foo.fish (line 4): 
    if nonexistent
       ^~~~~~~~~~^
in function 'this-errors' with arguments:
'foo\ bar aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa baz'
 called on line 11 of file foo.fish
in function 'call-this' called on line 1 of file foo.fish
```

for the latter. I have something whipped up, I'll add it here.

>Also I expect the ordering file,line,function,arguments instead of function,arguments,file,line

I think that would be awkward because we don't tend to use tons of files. So you would have a lot of

```
foo.fish:50:foo bar baz
foo.fish:42:function2 arguments
foo.fish:38:function1 arguments
```

which is a bunch of noise.

I've been experimenting with only showing the file when it *changed*

faho:
Alright, I'm going to be merging the first commit of this, also to 3.7.0 and then rework the rest later.

faho:
8d5d0c24aa0d9dd087d8b7d6c1e4cd8ec786c6c5 / 67faa107b0993564ebbb5c0dd0e94d0c35b47975

Issue #10080, opened 2023-11-02
Builtin `functions/ls.fish` Is Globally Setting `LS_COLORS`


nlordell:
Fish provides a very helpful `ls`, however the current implementation will globally set and export `LS_COLORS` (`set -gx`), which may have unintended consequences with other CLI tools that use this environment variable (and has different semantics for `LS_COLORS` not set and `LS_COLORS=""`). In particular, this affects the output of `fd` which will use `LS_COLORS` if it is set.

<img width="418" alt="image" src="https://github.com/fish-shell/fish-shell/assets/4210206/f8cc4169-472c-4c75-aa53-0ecd2c0e7435">

Personally, although you can argue that this may not be a `fish` issue, but with the respective tools and how they deal with `LS_COLORS`, it _feels_ like `fish` should not be modifying the global environment with its builtin functions like `ls`, as it may cause some weird inconsistencies depending on whether or not `ls` has been called or not.

I don't mind contributing a fix for this (keep the scope of `LS_COLORS` being set limited to the `ls` function call but also potentially caching its generated value to avoid re-calling `dircolors` every time `ls` is invoked), but wanted to do a temperature check before working on it.

### Version

```
fish, version 3.6.1
```

Looking at [the source](https://github.com/fish-shell/fish-shell/blob/496fc03b9880891ea3da4989cee50cf7381e707c/share/functions/ls.fish#L12) it seems to still be the case on `master`.

faho:
My preferred solution is to just get out of $LS_COLORS entirely - GNU ls and busybox ls will just use the defaults if it's unset. We'd have to check if other ls implementations are smart enough to figure it out. The function is annoyingly complicated as-is.

Barring that, we can do what you say and only set it in ls.

faho:
FreeBSD's ls [has a default](https://man.freebsd.org/cgi/man.cgi?ls):

> The default is "exfxcxdxbxegedabagacad"

It also uses $LSCOLORS where we set $LS_COLORS (see the underscore), so we're probably just using the defaults there to begin with.

OpenBSD and NetBSD don't have color at all, but ship a "colorls" that apparently is just FreeBSD's ls.

macOS' ls is FreeBSD's.

That means the majority of implementations (GNU, busybox, BSD) is covered, and if someone has an ls that supports color but doesn't have a default value they need to set $LS_COLORS/$LSCOLORS themselves. 

[...]

PR #10082, opened 2023-11-02
`ls` No Longer Sets `LS_COLORS`

nlordell:
## Description

Fixes issue #10080

This PR changes the builtin `ls` function to no longer load `LS_COLORS` before executing `command ls`. See the aforementioned issue for more details.

Additionally, a small refactor was done to the computed `ls` flags WRT the `-F` flag. In particular, the `ls` function was documented to:

> Make ls [...] show indicators if we are on a system that supports that feature and writing to stdout.

In particular this wasn't really checked as:
- For `colorls` systems (OpenBSD), this would always append the `-F` flag regardless of whether or not it was printing to STDOUT
- For `ls` systems with support for a colour flag (`--color=auto`, `-G` or `--color`), it would append `-F` when printing to STDOUT regardless of whether or not it was supported by the system
- For `ls` that didn't support a colour flag (`--color=auto`, `-G` or `--color`) but that supported `-F`, it would append `-F` regardless of whether or not it was printing to STDOUT

## TODOs:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,8 @@ We have tried to keep these to a minimum, but in some cases it is unavoidable.
 - ``functions --handlers`` will now list handlers in a different order.
   Now it is definition order, first to last, where before it was last to first.
   This was never specifically defined, and we recommend not relying on a specific order (:issue:`9944`).
+- ``LS_COLORS`` is no longer set automatically by ``ls`` (:issue:`10080`). Users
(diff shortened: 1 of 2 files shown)


nlordell:
I'm not sure what the best way to approach:

> Tests have been added for regressions fixed

would be...

faho:
Thanks, merged!

>I'm not sure what the best way to approach:

In general there are a lot of things that are extremely hard to test, especially things like this that are really system-dependent.

So we try not to sweat it too much.

PR #10084, opened 2023-11-05
fix(git): add `force-if-includes` completion

hwadii:
## Description

This PR adds `--force-if-includes` completion to git.


faho:
Thanks!

PR #10086, opened 2023-11-07
Add __fish_anypager helper

### LINKED ISSUE #10074: fish: Unknown command: less
**YOUR TEXT HERE**
fish, version 3.6.1
Linux archlinux 6.5.9-arch2-1 #1 SMP PREEMPT_DYNAMIC Thu, 26 Oct 2023 00:52:20 +0000 x86_64 GNU/Linux

```sh
root@archlinux ~# history
fish: Unknown command: less
/usr/share/fish/functions/history.fish (line 105):
                builtin history search $search_mode $show_time $max_count $_flag_case_sensitive $_flag_reverse $_flag_null -- $argv | $pager
                                                                                                                                      ^~~~~^
in function 'history'
```

faho:
This makes it easier to get *any pager* in the number of places we do.

Unfortunately:

1. It can't just execute the pager because that might block
2. We can't really set the necessary options for less here so they still need to be set outside.

This

Fixes #10074

by falling back to `cat` in that case. We could also decide to abort instead of using a non-pager, but for history that's probably fine.

## TODOs:

### DIFF
--- a/share/functions/__fish_anypager.fish
+++ b/share/functions/__fish_anypager.fish
@@ -1,0 +1,31 @@
+function __fish_anypager --description "Print a pager to use"
+    set -l pager
+    # We prefer $PAGER if we have it
+    set -q PAGER
+    and echo $PAGER | read -at pager
+
+    # or even $MANPAGER if we're allowed to
+    if test "$argv[1]" = "--with-manpager"
+        set -q MANPAGER
+        and echo $MANPAGER | read -at pager
+    end
+
+    # We use them if they *exist*
+    if command -q $pager[1]
+        printf %s\n $pager
+        return 0
+    end
+
+    # Cheesy hardcoded list of pagers.
+    for cmd in bat lv most less more
+        if command -q $cmd
+            echo -- $cmd
+            return 0
+        end
+    end
+
+    # We have no pager.
+    # We could fall back to "cat",
+    # but in some cases that's probably not helpful.
+    return 1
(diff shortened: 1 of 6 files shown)


share/functions/__fish_anypager.fish
```diff
@@ -0,0 +1,31 @@
...
+    end
+
+    # We use them if they *exist*
+    if command -q $pager[1]
+        printf %s\n $pager
+        return 0
+    end
+
+    # Cheesy hardcoded list of pagers.
+    for cmd in bat lv most less more
```
krobelus, line 20: I don't think we should prefer any concrete pager over `less` because `less` is the most widely used default. Both `man` and `git` use it as default. A user may have installed `bat` to use it for syntax highlighting.

averyterrel: I agree, `bat` appearing as my pager in place of `less` was quite unwelcome. I don't think telling users to set `$PAGER` or `$MANPAGER` **and changing the default pager** is a reasonable alternative to defaulting to the most widely used pager that is present on nearly every system.

krobelus: it's fixed in latest master by b83f3b0e9 (__fish_anypager: prefer less over other pagers, 2023-12-17) but not in a release yet unfortunately

averyterrel: Thank you very much!

PR #10090, opened 2023-11-08
Replace some direct uses of libc with wrappers

faho:
## Description

EDIT: Switched to our own wrappers, nix has problems compiling on FreeBSD with the "fs" feature.

This replaces our uses of libc's geteuid, getegid, getpid and isatty with the corresponding functions from the nix crate.

The main advantage is that it removes a bunch of awkward `unsafe` blocks. In turn nix has made some type choices like wrap pid_t in a Pid newtype that we then need to cast into an i32 again so we can use it.

This isn't the only solution. Alternatively we could:

- Make our own wrappers, possibly more direct
- Live with the unsafes
- Use nix *more*, integrating it more fully (e.g. add to_wstring to Pid)

Fundamentally, I would like for `unsafe` to be a rare thing that's easy to audit. In this case libc gets in the way because it defaults to marking things as unsafe, which includes things like getpid where I do not understand what the unsafety is supposed to be.

So I would like to wrap these things away, and these four functions were an easy start.

There are some other rust "style" things that I would like to change, including our uses of .unwrap(), which scares me (it's effectively an "assert"!), and I would like to remove a bunch of them.

### DIFF
--- a/fish-rust/src/builtins/test.rs
+++ b/fish-rust/src/builtins/test.rs
@@ -84,8 +84,7 @@ mod test_expressions {
         // Return true if the number is a tty().
         fn isatty(&self, streams: &mut IoStreams) -> bool {
             fn istty(fd: libc::c_int) -> bool {
-                // Safety: isatty cannot crash.
-                unsafe { libc::isatty(fd) > 0 }
+                crate::nix::isatty(fd)
             }
             if self.delta != 0.0 || self.base > i32::MAX as i64 || self.base < i32::MIN as i64 {
                 return false;
@@ -928,8 +927,7 @@ mod test_expressions {
             }
             Token::filetype_G => {
                 // "-G", for check effective group id
-                // Safety: getegid cannot fail.
-                stat_and(arg, |buf| unsafe { libc::getegid() } == buf.gid())
+                stat_and(arg, |buf| crate::nix::getegid() == buf.gid())
             }
             Token::filetype_g => {
                 // "-g", for set-group-id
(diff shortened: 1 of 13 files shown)


faho:
Lovely. This tries to build fspacectl on FreeBSD, which fails for reasons not entirely clear to me.

I've found https://github.com/nix-rust/nix/pull/2122, which seems related, but isn't in a nix release yet.

So we would have to wait for a release or adopt our own wrappers. To be honest this decreases my trust in nix.

faho:
Alright, added our own wrappers.

I do not believe that fiddling around with getting this nix feature to build on FreeBSD is in any way worth it given the wrappers are a single line each, *and* we get to avoid the newtypes which in our use would just get in the way.

This doesn't mean we couldn't switch to nix for other things and then *also* switch this.

fish-rust/src/builtins/test.rs
```diff
@@ -84,8 +84,7 @@ mod test_expressions {
         // Return true if the number is a tty().
         fn isatty(&self, streams: &mut IoStreams) -> bool {
             fn istty(fd: libc::c_int) -> bool {
-                // Safety: isatty cannot crash.
-                unsafe { libc::isatty(fd) > 0 }
+                crate::nix::isatty(fd)
```
krobelus, line 87: I think we should generally use unqualified function names

I've been using the `libc::` qualifier for C functions, to give a hint to folks who don't know the function.
But no qualifier for constants like `SIGHLD` because those are usually used in a `libc::` function call already.

Not sure what's the future. Here I think it's fine to use unqualified `isatty(fd)` always

krobelus, line 87: this function can be inlined now

krobelus (approved):

PR #10100, opened 2023-11-14
 Add more subcommands for apt

CicadaSeventeen:
## Description

Add more subcommands for command `apt`. For example: `download` `showrc`

### DIFF
--- a/share/completions/apt.fish
+++ b/share/completions/apt.fish
@@ -60,15 +60,24 @@ __fish_apt_option list -l all-versions -d 'Show all versions of any package'
 # Search
 __fish_apt_subcommand search -r -d 'Search for packages'
 
-# Search
+# Show
 __fish_apt_subcommand show -r -d 'Show package information'
+
+# Showsrc
+__fish_apt_subcommand showsrc -r -d 'Show information of source of given package'
+
+# Source
+__fish_apt_subcommand source -r -d 'Download source package'
 
 # Install
 __fish_apt_subcommand install -r -d 'Install packages'
 __fish_apt_option install -l reinstall -d 'Reinstall package'
 
+# Reinstall
+__fish_apt_subcommand reinstall -r -d 'Reinstall packages'
+
 # Remove
-__fish_apt_subcommand remove -r -d 'Remove packages'
+__fish_apt_subcommand remove -x -d 'Remove packages'
 
 # Edit sources
 __fish_apt_subcommand edit-sources -d 'Edit sources list'
@@ -81,6 +90,7 @@ __fish_apt_subcommand upgrade -r -d 'Upgrade packages'
 
 # Full Upgrade
 __fish_apt_subcommand full-upgrade -r -d 'Upgrade packages, removing others when needed'
+__fish_apt_subcommand disk-upgrade -r -d 'Same as full-upgrade'
 
 # Purge
 __fish_apt_subcommand purge -x -d 'Remove packages and delete their config files'
@@ -91,6 +101,16 @@ __fish_apt_subcommand changelog -r -d 'Download and display package changelog'
 # Autoremove
 __fish_apt_subcommand autoremove -d 'Remove packages no longer needed as dependencies'
 
+# Autoremove
+__fish_apt_subcommand autopurge -d 'Remove packages no longer needed as dependencies and delete their config files'
+
+# Clean
+__fish_apt_subcommand clean -d 'Clean all cache without lock'
+
+# Autoclean
+__fish_apt_subcommand autoclean -d 'Clean cache largely useless'
+#__fish_apt_subcommand auto-clean -d 'Clean cache largely useless'
+
 # Policy
 __fish_apt_subcommand policy -x -d 'Display source or package priorities'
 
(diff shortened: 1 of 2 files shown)


faho:
I don't see these subcommands even in the [man page for experimental](https://manpages.debian.org/experimental/apt/apt.8.en.html) or the `--help` output for apt 2.6.1.

They seem to *do* something, but if they aren't documented I don't want to add them because they seem like internal things.

Of course the man page also says:

> Much like apt itself, its manpage is intended as an end user interface and as such only mentions the most used commands and options partly to not duplicate information in multiple places and partly to avoid overwhelming readers with a cornucopia of options and details

which, frankly, seems like a terrible idea.

CicadaSeventeen:
`apt` never documents itself fully at all. These subcommands I add here are some of the 'not-documented' subcommands that really often to use. Of course I do not test any possible subcommands but to include often-used subcommands that I use every day and that `bash` or `zsh` completion have.

`bash` and `zsh` have included such completion. In fact what I do is to 'copy' completion of `bash`. (I just realize that 'copying' `zsh`'s completion is better since there are description about what such commands do) 

I have to say that without completion of like `download` or `source` or `build-dep` makes using `apt` under `fish` a bad experience- since such subcommands are too often used. So you can decide whether to include some of undocumented subcommmands here.

BTW,

1. `fish`'s `complete` also have half-undocumented option `-d` (only appears in examples but not listed as options ). I do not know why it is so.

2. You make mistake that copying comment of `search` to `show`. At least you may correct it.

share/completions/apt.fish
```diff
@@ -91,6 +101,16 @@ __fish_apt_subcommand changelog -r -d 'Download and display package changelog'
...
 
+# Autoremove
+__fish_apt_subcommand autopurge -d 'Remove packages no longer needed as dependencies and delete their config files'
+
+# Clean
+__fish_apt_subcommand clean -d 'Clean all cache without lock'
+
+# Autoclean
+__fish_apt_subcommand autoclean -d 'Clean cache largely useless'
+#__fish_apt_subcommand auto-clean -d 'Clean cache largely useless'
```
faho, line 112: This doesn't add the synonym, but `disk-upgrade` is there.

We should either add both or neither. To be honest my preference is for neither, because they just clog the suggestions.

share/completions/apt.fish
```diff
@@ -60,15 +60,24 @@ __fish_apt_option list -l all-versions -d 'Show all versions of any package'
 # Search
 __fish_apt_subcommand search -r -d 'Search for packages'
 
-# Search
+# Show
 __fish_apt_subcommand show -r -d 'Show package information'
 
+# Showsrc
+__fish_apt_subcommand showsrc -r -d 'Show information of source of given package'
```
faho, line 67: What does this description mean? What is the "information of source"?

Is that about the source package, or about the sources of the package?

faho (changes requested):

faho:
> apt never documents itself fully at all.

I would much prefer if they would just add documentation. It's *fine* to list the most-used commands first, or even to add a separate page that documents everything, but documentation also shows the *contract* of what the tool is supposed to offer, and what you can rely on.

The way this situation is, at the moment, means the apt people could pull any non-documented command and then go "oh we never meant that to be used that's why it's not documented".

(of course oversights do happen, sometimes a command ends up undocumented because it was forgotten about. that is simply a bug that should be fixed)

> fish's complete also have half-undocumented option -d (only appears in examples but not listed as options ). I do not know why it is so.

Corrected, thanks.

share/completions/apt.fish
```diff
@@ -91,6 +101,16 @@ __fish_apt_subcommand changelog -r -d 'Download and display package changelog'
...
 __fish_apt_subcommand autoremove -d 'Remove packages no longer needed as dependencies'
 
+# Autoremove
+__fish_apt_subcommand autopurge -d 'Remove packages no longer needed as dependencies and delete their config files'
+
+# Clean
+__fish_apt_subcommand clean -d 'Clean all cache without lock'
+
+# Autoclean
+__fish_apt_subcommand autoclean -d 'Clean cache largely useless'
```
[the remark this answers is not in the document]
CicadaSeventeen: `apt clean` clean (delete) all the cached packages unless you manually lock them, while `autoclean` only clean the packages that "can no longer be downloaded, and are largely useless"

share/completions/apt.fish
CicadaSeventeen: Maybe better to change suggestion to like `same as full-update`. 

In my opinion, to skip completion of `auto-clean` is OK because this alias is not really common (`bash` do not complete for it ) and has same beginning with `autoclean`.

As for `disk-upgrade` the problem is different. Though `disk-upgrade` is out-of-date, because it is used for ages for `apt-get`, many `apt` users are still more familiar to it rather than `full-upgrade`, so that to skip this completion can cause confusion.

share/completions/apt.fish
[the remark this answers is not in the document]
faho: Okay, I would phrase that as "Clean cache of packages that can no longer be downloaded", or possibly even a bit looser as "Clean cache of outdated packages".

Without that context the "largely useless" sounds like a judgement of the package, and absolutely not in what way.

share/completions/apt.fish
faho: How about "Show source package information"?

The term "source package" is a thing in debian.

share/completions/apt.fish
faho: Hold on, did you mean `dist-upgrade`? With a "t" in "dist" instead of "disk"? Short for "distro"?

I know that one, I've never heard of "disk-upgrade".

CicadaSeventeen: > Hold on, did you mean `dist-upgrade`? With a "t" in "dist" instead of "disk"? Short for "distro"?
> 
> I know that one, I've never heard of "disk-upgrade".

Sorry, I type mistake here. (Oh I am too often to spell wrong) . You are right.

Thanks.

zanchey:
Merged with rebase and some cleanup as b9b850f28. Thanks!

Issue #10088, opened 2023-11-08
`fish_vi_cursor` issues with `read`


bitraid:
fish version: 3.6.1, 723bfb9aa
OS: ArchLinux, Alpine
Terminals: foot, tmux

- **Non-interactive/scripts**:
`fish_cursor_*` settings are not applied to `read`, because of:
https://github.com/fish-shell/fish-shell/blob/f81c9cba50482fe6928bf04ba1c39cbead758e9d/share/functions/fish_vi_cursor.fish#L3-L5

- **Interactive**:
`fish_cursor_default` is not applied to read, unless it switches back from another mode.
This can be fixed by adding `--on-event fish_read` to:
https://github.com/fish-shell/fish-shell/blob/f81c9cba50482fe6928bf04ba1c39cbead758e9d/share/functions/fish_vi_cursor.fish#L70

PR #10101, opened 2023-11-14
Fix vi cursor for interactive and non-interactive read

bitraid:
## Description

Fixes issue #10088

## TODOs:

### DIFF
--- a/share/functions/fish_vi_cursor.fish
+++ b/share/functions/fish_vi_cursor.fish
@@ -67,7 +67,7 @@ function fish_vi_cursor -d 'Set cursor shape for different vi modes'
     or set -g fish_cursor_unknown block
 
     echo "
-          function fish_vi_cursor_handle --on-variable fish_bind_mode --on-event fish_postexec --on-event fish_focus_in
+          function fish_vi_cursor_handle --on-variable fish_bind_mode --on-event fish_postexec --on-event fish_focus_in --on-event fish_read
               set -l varname fish_cursor_\$fish_bind_mode
               if not set -q \$varname
                 set varname fish_cursor_unknown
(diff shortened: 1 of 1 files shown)


share/functions/fish_vi_cursor.fish
```diff
@@ -1,9 +1,4 @@
 function fish_vi_cursor -d 'Set cursor shape for different vi modes'
-    # If we're not interactive, there is effectively no bind mode.
-    if not status is-interactive
-        return
-    end
-
```
ridiculousfish, line 6: Nice find - however this check:

https://github.com/fish-shell/fish-shell/blob/6361362996bc6d915a22f010720aefa2f451500e/share/functions/fish_vi_cursor.fish#L2-L5

was added in 39905d41354358471c41b0f5de2d043f204ca0db to avoid the case of emitting an escape sequence for `scp` and similar non-interactive cases. So removing that check seems likely to cause regressions.

I'm not sure what to do. Maybe we could check if stdin is a tty instead?

ridiculousfish (changes requested):

share/functions/fish_vi_cursor.fish
bitraid: I tested the proposed change and it seems to work, so I pushed it.

krobelus:
can you share commands to reproduce the two issues?

bitraid:
> can you share commands to reproduce the two issues?

Just run `read` (from both interactive and script) after setting `fish_cursor_*` in vi mode.
For example (assuming default terminal cursor style is not *line*):
```
fish_vi_keybindings
set fish_cursor_insert line
read
fish -c read
```
For the first `read`, the cursor shape initially is not line (but changes once it switches out and then back to insert mode).
For the second `read` (non interactive), the cursor shape doesn't change at all.

PR #10103, opened 2023-11-16
completions: add chezmoi completions

nachtjasmin:
## Description

Added completions for [chezmoi](https://chezmoi.io).

I used the completions for kubectl as an inspiration, which means that it's up to chezmoi to always
generate the correct fish completions.

If this is not intended and the actual output should be committed instead,
I'd change this. Furthermore, there were no guidelines about commit messages
inside the CONTRIBUTING.rst, therefore I've oriented after the commit history of the master
branch.

## TODOs:

### DIFF
--- a/share/completions/chezmoi.fish
+++ b/share/completions/chezmoi.fish
@@ -1,0 +1,1 @@
+chezmoi completion fish | source


ridiculousfish:
Thank you!

ridiculousfish:
Actually I reverted this - it looks like chezmoi does indeed package its own completions, in my case in `/opt/homebrew/share/fish/vendor_completions.d/chezmoi.fish`. Any reason to not just allow it to ship its own?

nachtjasmin:
Actually, yes! You can also use [chezmoi for containers](https://www.chezmoi.io/user-guide/machines/containers-and-vms/) and in this case, it just installs itself as a binary into `~/.local/bin` instead of using the package manager, therefore missing the completions.

As a rule for further contributions, should I always check whether the completions are already packaged? But then, I've just checked: kubectl also has it's fish completions packaged, at least in the Arch Linux repository, yet it's inside of the fish repository. :D

nachtjasmin:
Oh, found out about the `$XDG_CONFIG_HOME/fish/completions` folder, that's gonna make things easier for me. Still curious about the general policy for contributions for completions tho ^^

PR #10104, opened 2023-11-16
completions: add just

nachtjasmin:
## Description

Add completions for [just](https://github.com/casey/just).

(same restrictions as in #10103 apply)

### DIFF
--- a/share/completions/just.fish
+++ b/share/completions/just.fish
@@ -1,0 +1,1 @@
+just --completions fish | source


ridiculousfish:
It looks like `just` ships its own completions - in my case it added them to `/opt/homebrew/share/fish/vendor_completions.d`. Any reason to not just use the ones packaged with `just`?

PR #10111, opened 2023-11-20
Add support for the fish_config command in the Windows MSYS environment.

ahaoboy:
## Description
This PR adds support for the fish_config command in the Windows MSYS environment.

## share/functions/fish_config.fish
Windows 11 comes with a default python.exe, but it's merely an empty installation package. It doesn't executing .py files via the command line, so add a detection to display a prompt message when script failed.

## share/tools/web_config/webconfig.py
- Under Windows fish_bin_name will append the .exe extension, and replace '\\' with '/' in the filepath. 
- Because the select function is not supported on Windows for sys.stdin, I use a socket as a replacement. I couldn't find any other solution here. Using multiple processes would add a lot of complexity and result in extensive code changes.
- To fix https://github.com/python/cpython/issues/112277, on Windows, the temporary file is closed after writing and flushing, set delete to True to ensure cleanup when program exit.
- To fix https://github.com/zephyrproject-rtos/windows-curses/issues/49, use empty string when get_tparm return None

### DIFF
--- a/share/functions/fish_config.fish
+++ b/share/functions/fish_config.fish
@@ -19,6 +19,14 @@ function fish_config --description "Launch fish's web based configuration"
         set -lx __fish_bin_dir $__fish_bin_dir
         if set -l python (__fish_anypython)
             $python "$__fish_data_dir/tools/web_config/webconfig.py" $argv
+
+            # If the execution of 'webconfig.py' fails, display python location and return.
+            if test $status -ne 0
+                echo "Please check if Python has been installed successfully."
+                echo "You can find the location of Python by executing the 'command -s $python' command."
+                return 1
+            end
+
         else
             echo (set_color $fish_color_error)Cannot launch the web configuration tool:(set_color normal)
             echo (set_color -o)"fish_config browse"(set_color normal) requires Python.
--- a/share/tools/web_config/webconfig.py
+++ b/share/tools/web_config/webconfig.py
@@ -82,6 +82,11 @@ def is_wsl():
     return False
 
 
+def is_windows():
+    """Return whether we are running under the Windows"""
+    return sys.platform.startswith("win")
+
+
 def is_sailfish_os():
     """Return whether we are running on Sailfish OS"""
     if "linux" in platform.system().lower() and os.access(
@@ -583,7 +588,8 @@ def get_special_ansi_escapes():
                     val = curses.tparm(key)
                 if val:
                     val = val.decode("utf-8")
-                return val
+                # Use an empty string instead of None.
+                return "" if val is None else val
 
             # Just a few for now
             g_special_escapes_dict["exit_attribute_mode"] = get_tparm("sgr0")
@@ -1583,9 +1589,13 @@ redirect_template_html = """
 # find fish
 fish_bin_dir = os.environ.get("__fish_bin_dir")
 fish_bin_path = None
+
+# only need the '.exe' extension on Windows
+fish_bin_name = "fish.exe" if is_windows() else "fish"
+
 if not fish_bin_dir:
     print("The $__fish_bin_dir environment variable is not set. " "Looking in $PATH...")
-    fish_bin_path = find_executable("fish")
+    fish_bin_path = find_executable(fish_bin_name)
     if not fish_bin_path:
         print("fish could not be found. Is fish installed correctly?")
         sys.exit(-1)
@@ -1593,7 +1603,7 @@ if not fish_bin_dir:
         print("fish found at '%s'" % fish_bin_path)
 
 else:
-    fish_bin_path = os.path.join(fish_bin_dir, "fish")
+    fish_bin_path = os.path.join(fish_bin_dir, fish_bin_name)
 
 if not os.access(fish_bin_path, os.X_OK):
     print(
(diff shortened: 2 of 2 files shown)


share/tools/web_config/webconfig.py
```diff
@@ -82,6 +82,11 @@ def is_wsl():
     return False
 
 
+def is_win():
```
faho, line 85: I would prefer if this were called "is_windows". We don't need to save the characters and that makes it easier to read and see.

faho (changes requested):

share/functions/fish_config.fish
```diff
@@ -19,6 +19,14 @@ function fish_config --description "Launch fish's web based configuration"
         set -lx __fish_bin_dir $__fish_bin_dir
         if set -l python (__fish_anypython)
             $python "$__fish_data_dir/tools/web_config/webconfig.py" $argv
+
+            # If the execution of 'webconfig.py' fails, display python location and return.
+            if test $status -ne 0
+                echo "Please check if Python has been installed successfully."
+                echo "You can find the location of Python by executing the 'which $python' command."
```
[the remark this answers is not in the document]
ahaoboy: done

share/tools/web_config/webconfig.py
```diff
@@ -583,7 +588,9 @@ def get_tparm(key):
                     val = curses.tparm(key)
                 if val:
                     val = val.decode("utf-8")
-                return val
+
+                # Use an empty string instead of None.
+                return "" if val is None else val
```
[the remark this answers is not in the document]
ahaoboy: On Windows, these values might be None.

share/tools/web_config/webconfig.py
```diff
@@ -1717,13 +1731,52 @@ def runThing():
...
+    sys.exit(-1)
+
+
+def capture_enter(port):
+    """Read keyboard events and establish a socket connection when pressing the Enter."""
+    import msvcrt
+    while True:
+        if msvcrt.kbhit():  # 检查是否有按键按下
+            key = msvcrt.getch().decode()  # 获取按下的键并解码为字符串
+            if key == "\r" or key == "\n":  # 如果是回车键，则结束输入
```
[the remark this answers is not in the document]
ahaoboy: done

share/tools/web_config/webconfig.py
ahaoboy: done

share/functions/fish_config.fish
```diff
@@ -19,6 +19,14 @@ function fish_config --description "Launch fish's web based configuration"
         set -lx __fish_bin_dir $__fish_bin_dir
         if set -l python (__fish_anypython)
             $python "$__fish_data_dir/tools/web_config/webconfig.py" $argv
+
+            # If the execution of 'webconfig.py' fails, display python location and return.
+            if test $status -ne 0
+                echo "Please check if Python has been installed successfully."
+                echo "You can find the location of Python by executing the 'command -s python $python' command."
```
faho, line 26: `$python` is still there.

ahaoboy: It should be `command -s $python`, bacause $python maybe python/python3

faho: No, $python won't be defined outside of this.

If you want you can do `command -s python python3`.

ahaoboy: Are you sure we can't access $python? This python variable is defined above. I tried it, and it worked. However, I'm not very familiar with Fish. If you're certain, I'll modify it to `command -s python python3`.
image

faho: It's a local variable - `set -l`.

That means once the function returns it'll be gone.

If the user runs `command -s $python` outside, it won't be the same $python.

share/tools/web_config/webconfig.py
```diff
@@ -589,16 +600,18 @@ def get_tparm(key):
             g_special_escapes_dict["exit_attribute_mode"] = get_tparm("sgr0")
             g_special_escapes_dict["bold"] = get_tparm("bold")
             g_special_escapes_dict["underline"] = get_tparm("smul")
+
+            # If get_tparm return None then use default value
+            for key in g_default_special_escapes_dict:
+                if g_special_escapes_dict[key] is None:
+                    g_special_escapes_dict[key] = g_default_special_escapes_dict[key]
```
faho, line 607: This isn't useful.

If the terminfo entry doesn't include these things, then we can't extract them, but e.g. `set_color` also won't add them.

These hardcoded escapes are, if the terminfo entry is that awkward, not guaranteed to mean what this thinks they do.

This isn't specific to Windows, it can also happen e.g. if you set $TERM to "dumb".

Please remove this change.

share/functions/fish_config.fish
ahaoboy: I used a wrong python3.exe, and got this output.
image

The echo output is: 
```
fish_config
Please check if Python has been installed successfully.
You can find the location of Python by executing the 'command -s python3' command.
```
`$python` will be replaced by the python command used by the user to execute the script. So users won't use $python 
 in the command line, did I misunderstand?

share/tools/web_config/webconfig.py
ahaoboy: ok. I will remove this

faho:
Thanks, merged!

zanchey:
sadly I think this is going to be in vain for 4.0, as the MSYS build won't work

PR #10121, opened 2023-11-26
Port builtin ulimit

thomasqueirozb:
## Description

This is not the best port ever but follows the C++ version closely. All of the RLIMIT variables are present in libc but AFAIK there is no way to check if each one of them exists like there is in C/C++ using `#ifdef`. Because of this I added them all as functions inside compat.c/compat.rs, which isn't pretty but is what resembles most the current functionality.

## TODOs:


thomasqueirozb:
CI failed in bind.py only in ubuntu-asan (didn't fail in previous runs). Doesn't seem to be anything related to my changes. Maybe rerun that job?

fish-rust/src/builtins/ulimit.rs
```diff
@@ -1,16 +1,489 @@
...
+    let resource = resource as c_int;
+
+    let result = unsafe { setrlimit(resource, rl as *const rlimit) };
+
+    if result == 0 {
+        Ok(())
+    } else {
+        Err(result)
+    }
+}
```
SteveLauC, line 45: For these 2 functions, we can use the safe interfaces exposed by the [Nix](https://docs.rs/nix/latest/nix/sys/resource/fn.getrlimit.html) crate, so that we don't need to bother with the unsafe libc stuff

SteveLauC: > Just implemented this but still had to use unsafe to convert i32 into the Resource enum unfortunately

Yeah, this is unfortunate, maybe we should provide a `try_from(i32)` function here

thomasqueirozb: Don't know how feasible it is since the values from different RLIMITs can change across platforms. We'd be reimplementing nix's or libc's code with a try_from implementation. The values come directly from C and are either defined or -1 but when the transmutation happens it's already a uint, so it's safe to do so.

SteveLauC: > We'd be reimplementing nix's or libc's code with a try_from implementation

Yeah, implementing it in fish does not make sense, I was thinking about doing it at Nix. Just realized that the macro used in Nix already has this functionality, [but it requests you to specify a `repr`](https://github.com/nix-rust/nix/blob/39ad47bc80af68fb9ad67d37ab699684bea0e2a9/src/macros.rs#L316), which is not suitable for things that are different across OSes,  I am thinking about adding it using third-party macros like [`num_enum`](https://crates.io/crates/num_enum).

FYI, `Resource` is `u32` on Linux(gnu) and Linux(uclibc), on musl and macOS, it has repr `i32`:

```rs
    #[cfg_attr(all(target_os = "linux", any(target_env = "gnu", target_env = "uclibc")), repr(u32))]
    #[cfg_attr(any(
            bsd,
            target_os = "android",
            target_os = "aix",
            all(target_os = "linux", not(any(target_env = "gnu", target_env = "uclibc")))
        ), repr(i32))]
```

thomasqueirozb: Yeah unfortunately TryFrom is not implemented in this case... And I didn't actually notice that it's only u32 on Linux GNU/ulibc. Either way I think the way it is written now should be fine since i32 and u32 should be equivalent in memory. I also don't think `num_enum` would work since it is a proc macro and the enum is defined inside `nix`, inside of their `libc_enum` macro. Is there anything else we could do or just leave it as is?

SteveLauC: > Is there anything else we could do or just leave it as is?

Leaving it as-is looks good

fish-rust/src/builtins/ulimit.rs
```diff
@@ -1,16 +1,468 @@
+use std::cmp::Ordering;
+
+use libc::{c_uint, rlim_t, RLIM_INFINITY};
+use nix::errno::Errno;
+use once_cell::sync::Lazy;
+
+use crate::{compat::*, fallback::*};
```
krobelus, line 7: Should not use wildcard import for `fallback`. Not sure about `compat`.

fish-rust/src/builtins/ulimit.rs
```diff
@@ -1,16 +1,468 @@
...
+        rlim_max,
+    )
+}
+
+/// Print the value of the specified resource limit.
+fn print(resource: c_uint, hard: bool, streams: &mut IoStreams) {
+    let l = get(resource, hard);
+
+    if l == RLIM_INFINITY {
+        streams.out.append(wgettext_fmt!("unlimited\n"));
```
krobelus, line 32: We have momentarily banned wgettext_fmt without format args

```suggestion
        streams.out.append(wgettext!("unlimited\n"));
```

fish-rust/src/builtins/ulimit.rs
```diff
@@ -1,16 +1,468 @@
...
+        streams.err.append(wgettext_fmt!(
+            "%ls: Resource limit not available on this operating system\n",
+            cmd
+        ));
+
+        builtin_print_error_trailer(parser, streams.err, cmd);
+        return STATUS_INVALID_ARGS;
+    }
+
+    let what = opts.what as c_uint; // Only negative value is RLIMIT_UNKNOWN
```
krobelus, line 273: better to use try_from()/try_into() + unwrap() IMO.
(In future we'll hopefully use an enum.)

fish-rust/src/builtins/ulimit.rs
```diff
@@ -1,16 +1,468 @@
...
+            desc,
+            switch_char,
+            multiplier,
+        }
+    }
 }
 
 /// Array of resource_t structs, describing all known resource types.
-const resource_arr: &[Resource] = &[];
+static resource_arr: Lazy<Box<[Resource]>> = Lazy::new(|| {
```
krobelus, line 349: can you name this `RESOURCE_ARR` because it's a global

fish-rust/src/compat.c
```diff
@@ -81,3 +82,130 @@ static const bool uvar_file_set_mtime_hack =
 #endif
 #undef UVAR_FILE_SET_MTIME_HACK
 bool UVAR_FILE_SET_MTIME_HACK() { return uvar_file_set_mtime_hack; }
+
+#ifndef RLIMIT_UNKNOWN
+#define RLIMIT_UNKNOWN -1
+#endif
```
krobelus, line 88: I think we can define this unconditionally. It shouldn't be defined by the stdlib.

fish-rust/src/compat.c
```diff
@@ -81,3 +82,130 @@ static const bool uvar_file_set_mtime_hack =
...
+
+#define RLIMIT_SIMPLE_FN(NAME) \
+    int C_##NAME() { return NAME; }
```
krobelus, line 91: (I'd probably inline this macro)

fish-rust/src/compat.rs
```diff
@@ -33,12 +33,67 @@ pub static _PATH_BSHELL: AtomicPtr<i8> = AtomicPtr::new(std::ptr::null_mut());
...
+macro_rules! CVAR {
+    ($cfn:ident, $cvar:ident, $type:ident) => {
+        #[allow(non_snake_case)]
+        pub fn $cvar() -> $type {
+            unsafe { $cfn() }
+        }
+    };
+}
+
+CVAR!(C_RLIMIT_SBSIZE, RLIMIT_SBSIZE, i32);
```
krobelus, line 45: > All of the RLIMIT variables are present in libc but AFAIK there is no way to check if each one of them exists like there is in C/C++ using #ifdef

Right. The limits are defined as enum variants like

        #[cfg(any(target_os = "freebsd", target_os = "dragonfly"))]
        #[cfg_attr(docsrs, doc(cfg(all())))]
        /// The maximum size (in bytes) of socket buffer usage for this user.
        RLIMIT_SBSIZE,

We don't want to use `#[cfg(any(target_os = "freebsd", target_os = "dragonfly"))]` because that requires extra work whenever OSs change.

We could outsource the checks to a compiler invocation in `build.rs`, similar to `compiles("fish-rust/src/cfg/w_exitcode.cpp")`.
I'm not sure how to correctly invoke `rustc`. We could use a C compiler or preprocesor too..

Probably too complicated for now. It's fine to leave it. Maybe there is an established pattern for this kind of problem.

krobelus:
looks good, thanks. A couple minor comments.
Ideally prefer to rebase instead of merge master, and squash most (probably all) commits.

fish-rust/src/builtins/ulimit.rs
```diff
@@ -1,16 +1,468 @@
...
+        wopt(L!("kernel-queues"), woption_argument_t::no_argument, 'K'),
+        wopt(L!("ptys"), woption_argument_t::no_argument, 'P'),
+        wopt(L!("threads"), woption_argument_t::no_argument, 'T'),
+        wopt(L!("help"), woption_argument_t::no_argument, 'h'),
+    ];
+
+    let mut opts = Options::default();
+
+    let mut shim_args: Vec<&wstr> = args.to_vec();
+    let mut w = wgetopter_t::new(SHORT_OPTS, LONG_OPTS, &mut shim_args);
```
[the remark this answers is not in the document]
thomasqueirozb: I didn't even know `w.argv` was a thing. I just mindlessly copied what other rust files did haha. This is way better than having to copy the args to a different vector

fish-rust/src/builtins/ulimit.rs
```diff
@@ -1,16 +1,468 @@
...
+
+/// Print values of all resource limits.
+fn print_all(hard: bool, streams: &mut IoStreams) {
+    let mut w = 0;
+
+    for resource in resource_arr.iter() {
+        w = w.max(fish_wcswidth(resource.desc));
+    }
+    for resource in resource_arr.iter() {
+        let (rlim_cur, rlim_max) = getrlimit(resource.resource).unwrap();
```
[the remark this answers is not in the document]
thomasqueirozb: I don't really think this approach is the best but made the change regardless. `set_limit` sets limits based on return values from `getrlimit` if the `soft`/`hard` flags are set. This means that if `getrlimit` fails, `setrlimit` can possibly be called with 0 instead of the intended/current value. Returning early before `setrlimit` (not necessarily crashing by unwrapping) would be best IMO.

thomasqueirozb:
> Ideally prefer to rebase instead of merge master, and squash most (probably all) commits.

Rebased the branch and commited each change separately. Feel free to squash everything into one commit once you feel this is ready to merge

PR #10128, opened 2023-11-29
Make wgettext_fmt! require an argument *to* format

faho:
Like FLOGF!, this now needs at least one argument to format.

This avoids some issues with missing variables and broken format strings - it is how I found 13ba5bd405c50ffa7b7c7fc4082e8acebb5b68c - where disown had a format string, with two placeholders, but no arguments to fill them with.

For use in e.g. macros, where it's otherwise hard to tell if we have something to format or not, this adds a wgettext_maybe_fmt! version to "maybe" format, if necessary.

### DIFF
--- a/fish-rust/src/builtins/type.rs
+++ b/fish-rust/src/builtins/type.rs
@@ -88,9 +88,9 @@ pub fn r#type(parser: &Parser, streams: &mut IoStreams, argv: &mut [&wstr]) -> O
                     let mut comment = WString::new();
 
                     if path.is_empty() {
-                        comment.push_utfstr(&wgettext_fmt!("Defined interactively"));
+                        comment.push_utfstr(&wgettext!("Defined interactively"));
                     } else if path == L!("-") {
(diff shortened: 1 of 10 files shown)


fish-rust/src/builtins/type.rs
```diff
@@ -103,9 +103,9 @@ pub fn r#type(parser: &Parser, streams: &mut IoStreams, argv: &mut [&wstr]) -> O
                     if props.is_copy() {
                         let path = props.copy_definition_file().unwrap_or(L!(""));
                         if path.is_empty() {
-                            comment.push_utfstr(&wgettext_fmt!(", copied interactively"));
+                            comment.push_utfstr(&wgettext!(", copied interactively"));
                         } else if path == L!("-") {
-                            comment.push_utfstr(&wgettext_fmt!(", copied via `source`"));
+                            comment.push_utfstr(&wgettext!(", copied via `source`"));
```
krobelus, line 108: good improvement for our current situation.
In future we should move to `std::fmt` syntax for which there are off-the-shelf checks that can run at compile time.

PR #10135, opened 2023-12-05
completions: add smerge (Sublime Merge CLI tool)

zgracem:
## Description

Adds completions for `smerge`, [Sublime Merge's CLI](https://www.sublimemerge.com/docs/command_line).

## TODOs:

### DIFF
--- a/share/completions/smerge.fish
+++ b/share/completions/smerge.fish
@@ -9,5 +9,6 @@ complete -c smerge -s v -l version -d 'Show version and exit'
 complete -c smerge -a search -x -d 'Search for commits in the current repository'
 complete -c smerge -a blame -r -d 'Blame the given file in the current repo'
 complete -c smerge -a log -r -d 'Show the file history in the current repo'
-complete -c smerge -a mergetool -r -d 'Open the merge tool for the given files'
+complete -c smerge -a mergetool -rF -d 'Open the merge tool for the given files'
 complete -c smerge -l no-wait -d "Don't wait for the application to close" -n '__fish_seen_subcommand_from mergetool'
+complete -c smerge -s o -rF -d "Merged output file" -n "__fish_seen_subcommand_from mergetool"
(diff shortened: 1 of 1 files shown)


share/completions/smerge.fish
```diff
@@ -0,0 +1,13 @@
+# Sublime Merge CLI tool
+
+complete -c smerge -s n -l new-window -d 'Open a new window'
+complete -c smerge -l launch-or-new-window -d 'Open new window only if app is running'
+complete -c smerge -s b -l background -d "Don't activate the application"
+complete -c smerge -l safe-mode -d 'Launch in a sandboxed environment'
+complete -c smerge -s h -l help -d 'Show help (this message) and exit'
+complete -c smerge -s v -l version -d 'Show version and exit'
+complete -c smerge -a search -x -d 'Search for commits in the current repository'
```
EmilyGraceSeville7cf, line 9: What about suggesting commits like [here](https://github.com/fish-shell/fish-shell/blob/master/share/completions/git.fish#L30)?

share/completions/smerge.fish
```diff
@@ -0,0 +1,13 @@
...
+complete -c smerge -a blame -r -d 'Blame the given file in the current repo'
+complete -c smerge -a log -r -d 'Show the file history in the current repo'
+complete -c smerge -a mergetool -r -d 'Open the merge tool for the given files'
+complete -c smerge -l no-wait -d "Don't wait for the application to close" -n '__fish_seen_subcommand_from mergetool'
```
EmilyGraceSeville7cf, line 13: It seems like `-o` option is not mentioned here for `mergetool` subcommand.

zgracem: Added `-o <file>` (and `--force-files`) to the `mergetool` subcommand in 781fc38

share/completions/smerge.fish
zgracem: I don't think that would be useful. The expected argument to `search` is a query string; the subcommand launches Sublime Merge with its argument pre-filled into the commit _message_ search field, which doesn't search commit hashes.

faho:
Thanks!

PR #10136, opened 2023-12-05
completions: add rename

zgracem:
## Description

Adds completions for [`rename`](http://plasmasturm.org/code/rename), [`rename`](https://www.mankier.com/1/rename), and [`rename`](https://metacpan.org/dist/File-Rename).

## TODOs:


faho:
`rename` is unfortunately an overloaded name - there's one in util-linux with [completely different options](https://www.mankier.com/1/rename).

There is also one that the arch package calls "perl-rename". I can't find yours packaged for my system.

This would unfortunately require parsing the `rename --version` output, see e.g. https://github.com/fish-shell/fish-shell/blob/8c36c21e2e1e59f2e31299da2ded6fb9020993b2/share/completions/ls.fish#L22 - except you need to see if it contains `util-linux`.

zgracem:
The `rename` on my Mac is the one Homebrew installs, which is _a_ Perl `rename`:

```
> brew info rename
==> rename: stable 1.601 (bottled), HEAD
Perl-powered file rename script with many helpful built-ins
http://plasmasturm.org/code/rename
```

It turns out I have access to a system with the `util-linux` version too, so I can edit this to make it account for both.

zgracem:
I found a _third_ version (from `sudo apt install rename` on Ubuntu) which is probably the real `perl-rename`, since it comes with Perl libraries.

```
me@sapphire ~ > rename --version
rename from util-linux 2.23.2

me@amethyst ~ > rename --version
/usr/bin/rename using File::Rename version 1.30, File::Rename::Options version 1.10

me@imac ~ > rename --version
# ...exits w/ status 2, prints usage info to stderr...
```

This PR now accounts for all three.

faho:
Nice work, merged!

Issue #10131, opened 2023-12-03
Support search pattern in history-pager


shouya:
Fish version: 3.6.1

I just migrated from zsh to fish and was generally very happy with it. However, one feature I am missing a lot was the ability to type search patterns in the "reverse-i-search" (aka "history-pager" in fish, bound to <kbd>C-r</kbd> by default). Although this feature needs to be enabled by binding `history-incremental-pattern-search-backward` as described [here](https://unix.stackexchange.com/a/16629), I usually have it bound when configuring my zsh.

Put simply, this feature enables one to type the `*` as a wildcard during the incremental search session, which matches against arbitrary string. For example, input `A*B` matches string `AwhateverinbetweenBandpossiblemoreafterwards`.

Why is this feature so useful? Let me demonstrate a common use case. I typically have a lot of  `curl` commands in the history. Typically, I have quite a number of varying options between the `curl` command and the `<url>`. However, when I am trying to recall a specific history entry, I usually only remember the two pieces of information: `curl` and the domain name I want. This is when I would search for stuff like `curl*google` and hope the results to be narrowed down significantly.

Now without this feature, what I have to do is to type `C-r curl`, scroll through the myriads of irrelevant history, and hope to find what I want somewhere. If that fails, I may have to fall back to running `history search curl | grep google`, which is also quite a lot of keystrokes.

krobelus:
yeah this sounds reasonable. A suggestion for the UI is https://github.com/fish-shell/fish-shell/pull/9451#issuecomment-1373446233
So a glob would match irrespective of word boundaries or newlines

krobelus:
note that in latest master we already fall back to fuzzy search if there are no exact matches
(commit 00692bcdf (Include subsequence matches in history-pager, 2023-01-15)).
Maybe we still want globs

shouya:
As for the suggested UI in https://github.com/fish-shell/fish-shell/pull/9451#issuecomment-1373446233, I personally prefer not requiring the `*` around the search term to trigger this (at least not the initial `*`). The reason is that it's often the case that one want to search something and find the input term not restrictive enough and need more terms to help narrow down. In this case, it would be a little inconvenient to:

1. go to the beginning of line, insert `*`,
2. go to the end of line, insert `*`
3. continue to type on whatever next term is.

[...]

PR #10143, opened 2023-12-08
Enable globbing in history-pager

shouya:
## Description

I made an attempt to implement the feature described in #10131 to support search pattern in history-pager. From my limited manual testing it works as expected for me.

Now the question becomes whether or not we really want this feature, at all? Or if we do want this feature, should it be promoted to the default (or the only) behavior, or gated through feature flag, or a fish variable? I'm happy to try implement either one of the options.

Fixes issue #10131.

## TODOs:

### DIFF
--- a/fish-rust/src/pager.rs
+++ b/fish-rust/src/pager.rs
@@ -102,6 +102,9 @@ pub struct Pager {
     // Whether we show the search field.
     pub search_field_shown: bool,
 
+    // Whether or not the pager should refilter the supplied completions.
+    pub refilter_disabled: bool,
+
     // The filtered list of completion infos.
     completion_infos: Vec<PagerComp>,
 
@@ -958,6 +961,15 @@ impl Pager {
 
     // Updates the completions list per the filter.
     pub fn refilter_completions(&mut self) {
+        // the result is already considered as filtered, we don't need
+        // to refine it again.
+        if self.refilter_disabled {
+            self.completion_infos = self.unfiltered_completion_infos.clone();
+            return;
+        }
+
+        // otherwise we need to filter through the completions
+        // according to the search field.
         self.completion_infos.clear();
         for i in 0..self.unfiltered_completion_infos.len() {
             if self.completion_info_passes_filter(&self.unfiltered_completion_infos[i]) {
@@ -1263,6 +1275,8 @@ mod pager_ffi {
         #[cxx_name = "set_completions"]
         fn set_completions_ffi(&mut self, completions: &CompletionListFfi);
         fn set_fully_disclosed(&mut self);
+        fn enable_refilter(&mut self);
+        fn disable_refilter(&mut self);
         #[cxx_name = "set_prefix"]
         fn set_prefix_ffi(&mut self, prefix: &CxxWString, highlight: bool);
         fn set_search_field_shown(&mut self, flag: bool);
@@ -1308,6 +1322,12 @@ impl Pager {
             Some(completion) => completion as *const Completion,
             None => std::ptr::null(),
         }
+    }
+    pub fn enable_refilter(&mut self) {
+        self.refilter_disabled = false;
+    }
+    pub fn disable_refilter(&mut self) {
+        self.refilter_disabled = true;
     }
     fn set_prefix_ffi(&mut self, prefix: &CxxWString, highlight: bool) {
         self.set_prefix(prefix.as_wstr(), highlight);
--- a/fish-rust/src/pager.rs
+++ b/fish-rust/src/pager.rs
@@ -15,7 +15,6 @@ use crate::fallback::{fish_wcswidth, fish_wcwidth};
 use crate::future::IsSomeAnd;
 use crate::highlight::{highlight_shell, HighlightRole, HighlightSpec};
 use crate::operation_context::OperationContext;
-use crate::parse_util::parse_util_contains_wildcard;
 use crate::screen::{Line, ScreenData};
 use crate::termsize::Termsize;
 use crate::wchar::prelude::*;
@@ -379,11 +378,6 @@ impl Pager {
         }
 
         let needle = self.search_field_line.text();
-
-        // Bypass glob pattern
-        if parse_util_contains_wildcard(needle) {
-            return true;
-        }
 
         // Match against the description.
         if string_fuzzy_match_string(needle, &info.desc, false).is_some() {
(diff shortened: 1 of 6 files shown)


fish-rust/src/pager.rs
```diff
@@ -379,6 +380,11 @@ impl Pager {
 
         let needle = self.search_field_line.text();
 
+        // Bypass glob pattern
+        if parse_util_contains_wildcard(needle) {
+            return true;
+        }
```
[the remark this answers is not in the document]
krobelus: the filtering is useful only for tab-completions, when the ctrl-s search field is used.
For the history pager we can skip the filtering completely.
There is an existing hack that detects the history pager based on `comp.replaces_commandline()`.
We can probably also pass a `bool` to `set_completions()` that's set to true for the history pager.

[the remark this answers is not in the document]
shouya: Upon closer look into how the pager is implemented, I realize it's indeed far more complex than I previously thought.

I first attempted implementing a toggle on disabling refilter in https://github.com/fish-shell/fish-shell/pull/10143/commits/4db3b67b36879174770baf401d87dce79367d102. Then I found that the tab-completion and C-r share the same pager instance, making it inconvenient to do set/reset refilter option every time. So I reverted that change and followed your original suggestion by adding a bool argument to `set_completions()`.

fish-rust/src/pager.rs
```diff
@@ -379,6 +380,11 @@ impl Pager {
 
         let needle = self.search_field_line.text();
 
+        // Bypass glob pattern
+        if parse_util_contains_wildcard(needle) {
```
krobelus, line 384: this is a leftover (it's wrong for the completion pager).
Also it's no longer necessary, if you squash your commits that should be obvious

src/reader.cpp
```diff
@@ -1136,17 +1136,11 @@ static history_pager_result_t history_pager_search(const HistorySharedPtr &histo
     size_t page_size = std::max(termsize_last().height / 2 - 2, (rust::isize)12);
 
     rust::Box<completion_list_t> completions = new_completion_list();
+
     rust::Box<HistorySearch> search =
-        rust_history_search_new(history, search_string.c_str(), history_search_type_t::Contains,
+        rust_history_search_new(history, search_string.c_str(), history_search_type_t::ContainsGlob,
```
krobelus, line 1141: if the value is from the commandline, we need to escape it:

```
index b10700988..6c45d6123 100644
--- a/src/reader.cpp
+++ b/src/reader.cpp
@@ -3672,7 +3672,8 @@ void reader_data_t::handle_readline_command(readline_cmd_t c, readline_loop_stat
             pager.set_prefix(MB_CUR_MAX > 1 ? L"► " : L"> ", false /* highlight */);
             // Update the search field, which triggers the actual history search.
             if (!history_search.active() || history_search.search_string().empty()) {
-                insert_string(pager.search_field_line(), *command_line.text());
+                insert_string(pager.search_field_line(),
+                              parse_util_escape_wildcards(*command_line.text()));
             } else {
                 // If we have an actual history search already going, reuse that term
                 // - this is if the user looks around a bit and decides to switch to the pager.
```

(need to add `parse_util_escape_wildcards`)

src/reader.cpp
```diff
@@ -1136,17 +1136,11 @@ static history_pager_result_t history_pager_search(const HistorySharedPtr &histo
...
                                 smartcase_flags(search_string), history_index);
     bool next_match_found = search->go_to_next_match(direction);
-    if (!next_match_found) {
-        // If there were no matches, try again with subsequence search
-        search = rust_history_search_new(history, search_string.c_str(),
-                                         history_search_type_t::ContainsSubsequence,
-                                         smartcase_flags(search_string), history_index);
-        next_match_found = search->go_to_next_match(direction);
-    }
```
krobelus, line 1149: removing the subsequence match fallback is somewhat unrelated.
I don't think we should do it in this PR, if at all.

The reason why people may expect subsequence matching is because the completion pager has it.
This overloaded behavior is weird but it usually doesn't matter.

We should make this fallback search for `parse_util_unescape_wildcards(search_string)` but without
the ANY_CHAR and ANY_STRING because they are implied by subsequence.

fish-rust/src/pager.rs
shouya: Fixed in https://github.com/fish-shell/fish-shell/pull/10143/commits/740718c08d26da5f35784af027c2d9c885dd57e9.

src/reader.cpp
shouya: I agree. The existing behaviour is restored in https://github.com/fish-shell/fish-shell/pull/10143/commits/f73c4a36dc130355e107204226f135725969c7c3.

Note that I think it makes little sense to fallback to subsequence search if the user include wildcard in their search term, so I added a check for that.

src/reader.cpp
krobelus: > Question 1: If the user decide to cancel the history pager (by pressing enter on no match), this will result in the user getting the escaped chars on their command line. So we need to unescape the wildcards again, right?

Good point, yes we should probably unescape the search text before copying it to the commandline. Not a big deal though. I'm not sure what's the least surprising behavior overall.

> Or is it possible to get the behavior as if user pressed esc so that their original input is restored?

Escape (aka `rl::cancel`) uses undo to restore the original input, but on Enter we want to copy whatever is in the pager search field at the moment, not necessarily the original commandline.

> Question 2: the semantic of the word "escape" in the parse_util is not very consistent.
> IMHO, what the current parse_util_unescape_wildcards does is not what I would call "unescape", at least not as in that escape_wildcards and unescape_wildcards are the inverse function of the other one.
> Perhaps we should name parse_util_escape_wildcards differently?

Yeah `parse_util_escape_wildcards` works on special internal representation of globs. 
We could put the one with the behavior we want here under `escape_string`.

LGTM overall.

shouya: Thanks. Is there anything else I should do in this PR before it was accepted? I can add a note about the change in `CHANGELOG.rst`, but I'm not sure under which version I put it. I saw this PR is tagged [fish next-3.x](https://github.com/fish-shell/fish-shell/milestone/36), but I wasn't so sure about what that means.

zanchey: Add it under the "fish 3.8.0" section, I've just cleaned it up to allow that.

PR #10149, opened 2023-12-11
fix deno task completion for jsonc

sigmaSd:
The JSON parse doesn't work for jsonc files that actually contains comments (with //) , this fixes it by using jsonc parser from deno std.

### DIFF
--- a/share/completions/deno.fish
+++ b/share/completions/deno.fish
@@ -1,20 +1,4 @@
 deno completions fish | source
 
 # complete deno task
-set searchForDenoFilesCode '
-import * as JSONC from "https://deno.land/std@0.208.0/jsonc/mod.ts";
-// order matters
-const denoFile = ["deno.json", "deno.jsonc", "package.json"];
-for (const file of denoFile) {
-  try {
-    Deno.statSync(file);
-    // file exists
-    const props = file === "package.json" ? "scripts" : "tasks";
-    console.log(
-      Object.keys(JSONC.parse(Deno.readTextFileSync(file))[props]).join("\n"),
-    );
-    break;
-  } catch {}
-}
-'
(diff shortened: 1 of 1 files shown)


branchvincent:
I noticed this too but forgot to make a PR. imo, it's easier to just parse `deno task` output (which would also handle tasks discovered from `package.json`). something like `deno task &| string match -rg '^- (\S*)'`

For reference:
<details>
<summary><b>`deno task` output</b></summary>
<p>

```console

Available tasks:
- prepare (package.json)
    npm run build
- build (package.json)
    rm -rf dist && tsc
- lint (package.json)
    tsc --noEmit && eslint --fix './src/**.ts'
- test (package.json)
    vitest
```

</p>
</details>

branchvincent:
oh sorry I didn't mean to imply you should close this, feel free to just update your pr with that idea :)

sigmaSd:
I tried it like this
```
complete -f -c deno -n "__fish_seen_subcommand_from task" -n "__fish_is_nth_token 2" -a "deno task &| string match -rg '^- (\S*)'"
```
but I get an error
```
complete: deno task &| string match -rg '^- (\S*)': contains a syntax error
complete: Expected a string, but found a pipe
```

branchvincent:
I think `-a` should be a subshell and I also realized we need to remove color codes from deno's output, so try:

```fish
complete -f -c deno -n "__fish_seen_subcommand_from task" -n "__fish_is_nth_token 2" -a "(deno task &| string match -rg '^- (\S*)' | string replace -ar -- '\x1b\x5b.*?m' '')"
```

sigmaSd:
cool that works, for the color deno respects NO_COLOR so I used that

ridiculousfish:
Thank you!

PR #10150, opened 2023-12-12
completions: improve Blender

PiOverFour:
## Description

Completions for Blender were added in #9905. This commit adds many improvements to these completions:

- Replace short options with old-style options: even though they are single-letter, Blender's options cannot be combined.
- Add comments to separate blocks of options, mirroring Blender's help message.
- Add missing options: render-frame, python-use-system-env, register-allusers, unregister, unregister-allusers.
- Remove options: debug-gpu-disable-ssbo, debug-gpu-renderdoc, -R. Those are no longer available.
- Add animation playback options: -a is a subcommand not compatible with the main ones. Related options use a condition to check for the presence of the -a option.
- Fix typos:
  - frame-end (was end-start)
  - no-window-focus (was no-native-pixels)
  - debug-depsgraph-eval (was -evel)
  - debug-depsgraph-pretty (was -time)
  - app-template (was open-last)
- Update output formats:
  - Add descriptions.
  - Add HDR, TIFF, OpenEXR, OpenEXR Multilayer, FFmpeg, Cineon, DPX, JPEG 2000, and WebP, which are optional but generally available.
  - Remove IRIZ, which is no longer available.
- Fix arguments for --use-extension: they should be 0 or 1, not true or false.
- Make env-system-* options require a parameter.
- Improve --add-ons by querying the list of add-ons inside Blender rather than a hardcoded path. This is because Blender's add-on modules may come from many different paths which depend on platform.
- Fix __blender_echo_input_file_name, by relying on extension.
- Fix listing of scene datablocks inside Blend file.
- Add listing of Python text datablocks to execute inside Blend file.

## TODOs:

### DIFF
--- a/share/completions/blender.fish
+++ b/share/completions/blender.fish
@@ -1,73 +1,109 @@
-function __blender_list_scenes -a file
-    blender --background $file --python-expr 'import bpy
-
-for name in [scene.name for scene in list(bpy.data.scenes)]:
-    print(f"\t{name}")' |
-        string replace -r -f '^\s+' ''
+function __blender_echo_input_file_name
+    # Find last argument ending in .blend (or .blend1, etc.)
+    # This is because a Blender invocation can open multiple blend file
+    # sequentially, so we need to find the last one up to this point.
+    set -l path (commandline -poc |
+        string match -r '.*\\.blend[0-9]*$' |
+        tail --lines=1)
+    # Using eval to expand ~ and variables specified on the commandline.
+    eval echo $path
 end
 
-function __blender_list_addons
-    path basename /usr/share/blender/scripts/addons/*.py |
-        path change-extension ''
+function __blender_list_scenes
+    blender --background (__blender_echo_input_file_name) --python-expr 'import bpy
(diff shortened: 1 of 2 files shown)


EmilyGraceSeville7cf:
Thanks for this improvement, LGTM. 😄

PiOverFour:
> Thanks for this improvement, LGTM. 😄

Thanks for adding the completions in the first place ;)

share/completions/blender.fish
```diff
@@ -1,124 +1,183 @@
...
-    set -l current_token (commandline -t)
-
-    if test "$previous_token" = --addons
-        __blender_list_addons |
-            string replace -r '^' $current_token |
-            string replace -r '$' ','
+    set -l current_token (commandline -t | string replace -r '[^,]*$' '')
+    set -l addons (__blender_list_addons)
+    for a in $addons
+        echo "$current_token$a,"
```
krobelus, line 45: prefer to use `__fish_append`

PiOverFour: Thanks! I didn’t know about it.

Issue #7186, opened 2020-07-09
Can't change certain bash prompt colors


Bug-Reaper:

faho:
First of all, an important clarification: This is fish, not bash. There's not a bash prompt in sight.

Secondly: 

>These weren't in the docs about setting colors on the prompt 

These were added in 14b953e414caf2d52df47cdfffa8cbf3b94ae65d?

>namely the @ seperator

That's because it's not colored at all (well, colored "normal").

-----

Yes, these should be added to fish_config. If you want, go for it!

Bug-Reaper:
Hey wait, I still don’t know how to change the color of the `@` or the `>` presumably their color can be customized right?

faho:
>Hey wait, I still don’t know how to change the color of the @ or the > presumably their color can be customized right?

Not with a variable, no.

You can, however, change *everything* about the prompt by actually changing the function - `funced fish_prompt`.

Bug-Reaper:
Alright cool cool! Thanks @faho, I’ll give it a shot

Bug-Reaper:
Is there anywhere I can look at the different `fish_prompt` functions you can choose from in `fish_config`

faho:
>Is there anywhere I can look at the different fish_prompt functions you can choose from in fish_config

@Bug-Reaper [share/tools/web_config/sample_prompts](https://github.com/fish-shell/fish-shell/tree/master/share/tools/web_config/sample_prompts)

Bug-Reaper:
Kinda swamped with some work right now but I’ll return to this to add better docs and more features to the GUI style configurator

Ahanaf-Ether:
@faho help i dont know which varrible to change in order to change the "@"

Ahanaf-Ether:
[https://paste.pics/967cb9b5552946a4e167b8b16bc67b98](url) This is what i want, i have it in bash. But i want the same thing for fish.

vext01:
In `prompt_login.fish` there is:

```
    echo -n -s (set_color $fish_color_user) "$USER" (set_color normal) @ (set_color $color_host) (prompt_hostname) (set_color normal)
```

I wonder why we don't add a `fish_color_at` (or whatever) colour so that people can customise the colour without duplicating the whole function into their config?

zanchey:
I think that would be fine

vext01:
@zanchey I drafted https://github.com/fish-shell/fish-shell/pull/10164 for this.

PR #10164, opened 2023-12-20
Use `fish_color_at` to decide the colour of the @ (if present) in the prompt

vext01:
## Description

As discussed [here](https://github.com/fish-shell/fish-shell/issues/7186), you can't easily change the colour of the `@` in the prompt (if present).

This draft proposes a fix.

## TODOs:

- [ ] Discuss if this is the correct way to do it.
- [ ] Update all themes, adding `fish_color_at=normal`, so that they all render the same if the user doesn't specify.
- [ ] Do something similar for any trailing prompt character? (in the linked issue, the user can't change the colour of the `>`)

### DIFF
--- a/share/functions/prompt_login.fish
+++ b/share/functions/prompt_login.fish
@@ -24,5 +24,5 @@ function prompt_login --description "display user name for the prompt"
         set color_host $fish_color_host_remote
     end
 
-    echo -n -s (set_color $fish_color_user) "$USER" (set_color normal) @ (set_color $color_host) (prompt_hostname) (set_color normal)
+    echo -n -s (set_color $fish_color_user) "$USER" (set_color $fish_color_at) @ (set_color $color_host) (prompt_hostname) (set_color normal)
 end
--- a/share/tools/web_config/themes/fish default.theme
+++ b/share/tools/web_config/themes/fish default.theme
@@ -33,3 +33,4 @@ fish_color_option cyan
 fish_color_keyword blue
 fish_color_host_remote yellow
 fish_color_status red
+fish_color_at normal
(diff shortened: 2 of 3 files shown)


krobelus:
how about

```fish
functions -c fish_prompt default_fish_prompt
function fish_prompt
    default_fish_prompt | string replace @ (set_color red)@
end
```

vext01:
> how about

I guess that's a workaround, but it doesn't allow individual themes to describe their desired colour.

krobelus:
is there any theme that wants to this to be different than `$fish_color_normal`?
If not then we can do

```diff
diff --git a/share/functions/prompt_login.fish b/share/functions/prompt_login.fish
index 632a462bd..3ec1647fd 100644
--- a/share/functions/prompt_login.fish
+++ b/share/functions/prompt_login.fish
@@ -26,3 +26,3 @@ function prompt_login --description "display user name for the prompt"
 
-    echo -n -s (set_color $fish_color_user) "$USER" (set_color normal) @ (set_color $color_host) (prompt_hostname) (set_color normal)
+    echo -n -s (set_color $fish_color_user) "$USER" (set_color $fish_color_normal) @ (set_color $color_host) (prompt_hostname) (set_color normal)
 end
```

AFAICT `fish_color_normal` is used as fallback for syntax highlighting on text that doesn't have a distinguishing role, or for cases where the highlighting for that role has been explicitly disabled (such as `set -e fish_color_command `).

I think that's a relatively obvious approach. In future we should also give prompt variables like `$fish_color_user` the same behavior of falling back to `$fish_color_normal` if unset.

vext01:
I'm closing this for now.

PR #10168, opened 2023-12-24
themes: Synchronize Nord theme in adherence to color palette guidelines

ericvw:
## Description

Adhere as best as possible to the style guidelines at https://www.nordtheme.com/docs/colors-and-palettes. Some adaptations were made so that `functions <function>` is also syntax highlighted per the upstream recommendations.

Additionally, the theme file has been reordered to follow the order of variables defined in interactive syntax-highlighting-variables documentation.


ridiculousfish:
Thank you!

zanchey:
Probably this needs adding to the license file, I will do that before a release

PR #10180, opened 2024-01-01
Rewrite input machinery in Rust

ridiculousfish:
## Description

This ports `input.cpp`, `input_common.cpp`, and `builtin bind`.

Bind came along for the ride because I found having it go over the FFI to be too much work.


ridiculousfish:
Looks like @krobleus is basing his changes in #10179 off of this PR - that seems fine to me, I'll be happy to review those.

fish-rust/src/input_common.rs
```diff
@@ -1,21 +1,586 @@
...
+        WAIT_ON_ESCAPE_MS.store(WAIT_ON_ESCAPE_DEFAULT, Ordering::Relaxed);
+        return;
+    };
+    let fish_escape_delay_ms = fish_escape_delay_ms.as_string();
+    match fish_wcstol(&fish_escape_delay_ms) {
+        Ok(val) if (10..5000).contains(&val) => {
+            WAIT_ON_ESCAPE_MS.store(val.try_into().unwrap(), Ordering::Relaxed);
+        }
+        _ => {
+            eprintln!(
```
krobelus, line 230: I found that we can't mix `print!` with `printf!` because `print!` uses a buffer that our
`printf!`/`wwrite_to_fd` doesn't use. The `println!` variants should be fine though.

fish-rust/src/input_common.rs
```diff
@@ -1,21 +1,586 @@
...
+        // We are not prepared to handle a signal immediately; we only want to know if we get input on
+        // our fd before the timeout. Use pselect to block all signals; we will handle signals
+        // before the next call to readch().
+        let mut sigs: libc::sigset_t = unsafe { std::mem::zeroed() };
+        unsafe { libc::sigfillset(&mut sigs) };
+
+        // pselect expects timeouts in nanoseconds.
+        const NSEC_PER_MSEC: u64 = 1000 * 1000;
+        const NSEC_PER_SEC: u64 = NSEC_PER_MSEC * 1000;
+        let wait_nsec: u64 = (wait_time_ms as u64) * NSEC_PER_MSEC;
```
krobelus, line 380: should be a checked cast?

fish-rust/src/input_common.rs
```diff
@@ -1,21 +1,586 @@
...
+    }
+}
+
+#[test]
+fn test_push_front_back() {
+    let mut queue = InputEventQueue::new(0);
+    queue.push_front(CharEvent::from_char('a'));
+    queue.push_front(CharEvent::from_char('b'));
+    queue.push_back(CharEvent::from_char('c'));
+    queue.push_back(CharEvent::from_char('d'));
```
krobelus, line 536: not so important here but when I'm navigating the call graph, I almost always want
to ignore test code. Separate files makes that easy

krobelus (approved):

PR #10184, opened 2024-01-02
feat: adb completion cleared of awk

JosefLitos:
Ticks off adb in awk list in issue #6520

The latest available package on Archlinux also still doesn't have fixed `$1` to `$0` in the `awk command` that was replaced. It's fixed here already though, so I tried to make a small step forward in some other issue in hope for the next release to ship that adb completion fixed.

### DIFF
--- a/share/completions/adb.fish
+++ b/share/completions/adb.fish
@@ -74,7 +74,7 @@ function __fish_adb_list_files
     end
 
     # Return list of directories suffixed with '/'
-    __fish_adb_run_command find -H "$token*" -maxdepth 0 -type d 2\>/dev/null | awk '{print $0"/"}'
+    __fish_adb_run_command find -H "$token*" -maxdepth 0 -type d 2\>/dev/null | string -r '$' /
     # Return list of files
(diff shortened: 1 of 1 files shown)


faho:
Thanks!

NextAlone:
This commit breaks function `__fish_adb_list_files`
```
string: string: invalid subcommand

/usr/local/Cellar/fish/HEAD-b75e5ee/share/fish/completions/adb.fish (line 76):
    __fish_adb_run_command find -H "$token*" -maxdepth 0 -type d 2\>/dev/null | string -r '$' /
    ^
in function '__fish_adb_list_files'
in command substitution

(Type 'help string' for related documentation)
```

krobelus:
thanks, reverted as 6eddaa37af592253f75ac50990373d4810533712

NextAlone:
@krobelus Maybe this commit is his/her original meaning, to supplement '/' after the directories.

NextAlone:
image
Aimed at supplement the '/' after items, which are directories.

PR #10188, opened 2024-01-03
Remove FFI code and C++ files

krobelus:
Depends on https://github.com/fish-shell/fish-shell/pull/10179

Turn FFI tests into native Rust tests.
To make "cargo test" work we need to link against curses in build.rs.
(`cargo test` assumes the build directory is "build/" and the CMake `tests_dir` goal has been made)

Remove most C++ files.

There's a lot more we can remove, like
- cxx, autocxx
- now-unused CMake code
- C++ entry points
- remaining mentions of "ffi"


faho:
macOS and FreeBSD fail because of gettext.

As far as I can tell that's currently non-functional, so we could just comment it out for the time being. We're gonna have to figure out translation again later.

krobelus:
I wasn't aware they are nonfunctional. To fix the build we probably only need to link against `Intl_LIBRARIES`. Unless `have_gettext` does somehow not match `HAVE_GETTEXT` on these platforms.

faho:
Hold on, it appears this actually *is* working now, but it doesn't reload properly:

```fish
LANG=de_DE.UTF-8 fish -c '_ file; echo; set -gx LANG en_US.UTF-8; _ file'
```

prints "Datei" and "Datei". But it also does that with 3.7.0, so this is a preexisting bug.

fish-rust/src/wutil/gettext.rs
```diff
@@ -1,17 +1,102 @@
...
-    wstr::from_slice(slice).expect("Invalid UTF-32")
+    // Preserve errno across this since this is often used in printing error messages.
+    let err = errno();
+
+    // TODO don't allocate
+    let key = text.to_vec();
+
+    wgettext_init_if_necessary();
+    #[allow(clippy::type_complexity)]
+    static WGETTEXT_MAP: Lazy<Mutex<HashMap<Vec<wchar_t>, Pin<Box<WString>>>>> =
```
faho, line 81: This cache needs to be invalidated if we switch locale (e.g. `set -gx LANG foo` in config.fish).

I *think* it would have to be a global for that?

See e.g.

```fish
LANG=de_DE.UTF-8 /usr/bin/fish -c 'set -gx LC_ALL en_US.UTF-8; _ file'
```

Which will correctly print "file" in english. But if you add another `_ file` before the `set`, it will have cached it and print "Datei" both times.

faho:
Ah, dangit, I caused a merge conflict. Sorry!

PR #10196, opened 2024-01-07
Move from errno crate to nix errno

PolyMeilex:
## Description

`nix` crate already has rusty errno API backed in, without a need for any additional feature flags, so there is no need to depend on extra crate for that.

There was no way to set errno via `nix`, but I [submitted a patch for that already](https://github.com/nix-rust/nix/pull/2283), once it lands `errno` crate can be removed all together just by swapping set functions.
In the meantime I used a small wrapper for setting the errnos, just to make this mergable if you wish to.

Not sure if those sorts of changes are even wanted, but yeah I guess I'm asking after the fact :smile: 

If using more of nix is welcome, then let me know if enabling more of its feature flags like `fs`/`signals`/`user` is also fine, as I'd love to work on porting more stuff from raw libc calls to nix equivalents in the future. (obviously that would be in smaller, more review friendly portions)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,7 +45,7 @@ printf-compat = { git = "https://github.com/fish-shell/printf-compat.git", branc
 autocxx = "0.23.1"
 bitflags = "2.4.0"
 cxx = "1.0"
-errno = "0.2.8"
+_errno = { package="errno", version = "0.2.8" }
 lazy_static = "1.4.0"
 libc = "0.2.137"
 lru = "0.10.0"
--- a/fish-rust/src/fd_monitor.rs
+++ b/fish-rust/src/fd_monitor.rs
@@ -9,8 +9,8 @@ use crate::fds::AutoCloseFd;
 use crate::flog::FLOG;
 use crate::threads::assert_is_background_thread;
 use crate::wutil::perror;
-use errno::errno;
-use libc::{self, c_void, EAGAIN, EINTR, EWOULDBLOCK};
+use libc::{self, c_void};
+use nix::errno::Errno;
 
 #[cfg(not(HAVE_EVENTFD))]
 use crate::fds::{make_autoclose_pipes, make_fd_nonblocking};
@@ -100,11 +100,11 @@ impl FdEventSignaller {
                     std::mem::size_of_val(&buff),
                 )
             };
-            if ret >= 0 || errno().0 != EINTR {
+            if ret >= 0 || Errno::last() != Errno::EINTR {
                 break;
             }
         }
-        if ret < 0 && ![EAGAIN, EWOULDBLOCK].contains(&errno().0) {
+        if ret < 0 && ![Errno::EAGAIN, Errno::EWOULDBLOCK].contains(&Errno::last()) {
             perror("read");
         }
         ret > 0
@@ -127,12 +127,12 @@ impl FdEventSignaller {
                     std::mem::size_of_val(&c),
                 )
             };
-            if ret >= 0 || errno().0 != EINTR {
+            if ret >= 0 || Errno::last() != Errno::EINTR {
                 break;
             }
         }
         // EAGAIN occurs if either the pipe buffer is full or the eventfd overflows (very unlikely).
-        if ret < 0 && ![EAGAIN, EWOULDBLOCK].contains(&errno().0) {
+        if ret < 0 && ![Errno::EAGAIN, Errno::EWOULDBLOCK].contains(&Errno::last()) {
             perror("write");
         }
     }
@@ -511,7 +511,7 @@ impl BackgroundFdMonitor {
                     .map(|duration| duration.as_micros() as u64)
                     .unwrap_or(FdReadableSet::kNoTimeout),
             );
-            if ret < 0 && errno::errno().0 != libc::EINTR {
+            if ret < 0 && Errno::last() != Errno::EINTR {
                 // Surprising error
                 perror("select");
             }
--- a/fish-rust/src/signal.rs
+++ b/fish-rust/src/signal.rs
@@ -4,11 +4,12 @@ use crate::common::{exit_without_destructors, restore_term_foreground_process_gr
 use crate::event::{enqueue_signal, is_signal_observed};
 use crate::nix::getpid;
 use crate::reader::{reader_handle_sigint, reader_sighup};
+use crate::set_errno;
 use crate::termsize::TermsizeContainer;
 use crate::topic_monitor::{generation_t, topic_monitor_principal, topic_t, GenerationsList};
 use crate::wchar::prelude::*;
 use crate::wutil::{fish_wcstoi, perror};
-use errno::{errno, set_errno};
+use nix::errno::Errno;
 use std::sync::atomic::{AtomicI32, Ordering};
 
 /// Store the "main" pid. This allows us to reliably determine if we are in a forked child.
@@ -57,7 +58,7 @@ extern "C" fn fish_signal_handler(
     _context: *mut libc::c_void,
 ) {
     // Ensure we preserve errno.
-    let saved_errno = errno();
+    let saved_errno = Errno::last();
 
     // Check if we are a forked child.
     if reraise_if_forked_child(sig) {
--- a/fish-rust/src/wutil/mod.rs
+++ b/fish-rust/src/wutil/mod.rs
@@ -19,8 +19,8 @@ use crate::flog::FLOGF;
 use crate::wchar::{wstr, WString, L};
 use crate::wchar_ext::WExt;
 use crate::wcstringutil::{join_strings, split_string, wcs2string_callback};
-use errno::errno;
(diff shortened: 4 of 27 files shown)


fish-rust/src/wutil/mod.rs
```diff
@@ -68,13 +68,13 @@ pub fn wperror(s: &wstr) {
 
 /// Port of the wide-string wperror from `src/wutil.cpp` but for rust `&str`.
 pub fn perror(s: &str) {
-    let e = errno().0;
+    let e = Errno::last();
```
krobelus, line 71: The weird thing about `nix::errno::Errno` is that they treat "unknown error" the same as "no error".
This could cause subtle issues if the OS ever returns an error that hasn't been added to `nix` yet.
Unless they have a way of guaranteeing consistency with the host libc?

---

I'm not sure `enum` is the right type for `Errno`.
It's already marked `#[non_exhaustive]` so client code can't do exhaustive matching.
I guess they could make it:

```
enum Errno {
    NoError = 0
    EPERM = libc::EPERM,
    ...
    UnknownErrno(i32),
}
```

but at that point it's probably better to use a struct that wraps `libc::c_int` (or `i32`?), similar to `std::io::Error`.

---

I guess we could use the `nix::errno::Errno::last_raw()` (which is `nix::errno::errno()` today) and `nix::errno::Errno::set_raw()` precisely in the cases where we don't want to ignore unknown errors.
But the problem of being easy to misuse remains.

Cargo.toml
```diff
@@ -45,7 +45,7 @@ printf-compat = { git = "https://github.com/fish-shell/printf-compat.git", branc
 autocxx = "0.23.1"
 bitflags = "2.4.0"
 cxx = "1.0"
-errno = "0.2.8"
+_errno = { package="errno", version = "0.2.8" }
```
krobelus, line 48: nice, I didn't know this was possible

fish-rust/src/fd_monitor.rs
```diff
@@ -11,8 +11,8 @@ use crate::ffi::void_ptr;
 use crate::flog::FLOG;
 use crate::threads::assert_is_background_thread;
 use crate::wutil::perror;
-use errno::errno;
-use libc::{self, c_void, EAGAIN, EINTR, EWOULDBLOCK};
+use libc::{self, c_void};
```
krobelus, line 14: (if we take this, should squash all the "Port" commits into one, and add a bit more info to the commit message)

fish-rust/src/signal.rs
```diff
@@ -4,13 +4,14 @@ use crate::common::{exit_without_destructors, restore_term_foreground_process_gr
...
 use crate::reader::{reader_handle_sigint, reader_sighup};
+use crate::set_errno;
 use crate::termsize::TermsizeContainer;
 use crate::topic_monitor::{generation_t, topic_monitor_principal, topic_t, GenerationsList};
 use crate::wchar::prelude::*;
 use crate::wchar_ffi::{AsWstr, WCharToFFI};
 use crate::wutil::{fish_wcstoi, perror};
 use cxx::{CxxWString, UniquePtr};
-use errno::{errno, set_errno};
+use nix::errno::Errno;
```
krobelus, line 14: > If using more of nix is welcome, then let me know if enabling more of its feature flags like fs/signals/user is also fine,

Sure, we should try to increase the level of abstraction.
Do nix APIs depend on `Errno`? That would be an argument for it.

faho:
I've now triggered the FreeBSD builds, let's see if they work out.

Last time I tried to move something to nix they did not, which is why I'm kind of wary of nix now.


-----

As a general thing: What's the future of errno-the-variable in fish?

I suppose we should replace our own uses of it with proper return values - e.g. [input_terminfo_get_sequence](https://github.com/fish-shell/fish-shell/blob/6a64ba6638482225419af1ee4f340badaf7891d0/fish-rust/src/input.rs#L1129-L1148) should just return a Result.

But do we still need to 1. clear errno, 2. read errno itself from functions we get via libc? Or how would this work with [std::io::Error::last_os_error](https://doc.rust-lang.org/std/io/struct.Error.html#method.last_os_error)?

krobelus:
I think the only place where we have an inherent need for `set_errno` is `fish_signal_handler`.
Everywhere else we should use wrappers that return `Result`. Preferrably we can use nix wrappers but they would need to work on BSD (see https://github.com/fish-shell/fish-shell/pull/10090).

fish-rust/src/wutil/mod.rs
PolyMeilex: Yeha, I noticed that as well, it may cause problems in places where we do if errno != 0. Where unknown error will be treated as success.

The best way would be to remove != 0 checks in favour of Result.is_ok, and in places where we actually need to fiddle with errnos do the raw ones.

So simply, we would not use NoErrno anywhere and assume that we don't care about the difference between unsupported errno types, would that be fine?

Obviously I would be open to working with nix to perhaps add a way to distinguish those, they probably simply never needed that as they do Results instead of NoErrno values everywhere.

fish-rust/src/fd_monitor.rs
PolyMeilex: Sure, this was done to make it easy to drop commits as needed, but I don't hear any pushback related to specific modules so I will start squashing stuff down to a more reasonable amount.

fish-rust/src/wutil/mod.rs
krobelus: > Yeha, I noticed that as well, it may cause problems in places where we do if errno != 0

Right, this is actually a rare case so it does little harm.

> So simply, we would not use NoErrno anywhere and assume that we don't care about the difference between unsupported errno types, would that be fine?

Sounds good. To be clear, cases like `perror` that handle arbitrary errors should use the raw errno, just in case.

It's probably wise to first convert things like `wrealpath` to return `Result`, to make it easy to review.

PolyMeilex: > It's probably wise to first convert things like wrealpath to return Result, to make it easy to review.

Yeah, this is a better approach tbh, I think I will draft this PR for now, and in separate smaller PRs focus on moving stuff to nix & Results rather than focusing on removing errno crate.

After that is done, I will come back to the topic of errno removal.

krobelus: whatever works. We can start merging all the uses of errno where we only compare against concrete errors

SteveLauC: Hi, one of the Nix maintainers here, happy to help to make Nix serve fish better!

> The weird thing about nix::errno::Errno is that they treat "unknown error" the same as "no error".
This could cause subtle issues if the OS ever returns an error that hasn't been added to nix yet.

Yeah, this is indeed a bad design, I think we can do the following approach:
>
> ```rs
> enum Errno {
>     NoError = 0
>     EPERM = libc::EPERM,
>      ...
>     UnknownErrno(i32),
> }
> ```

But maybe `NoError` should not be included as one of them, and for `nix::errno::Errno::last()`, we can make it return an `Option<Errno>`:

```rs
impl Errno {
    pub fn last() -> Option<Errno> {
        match Self::last_raw() {
            0 => None,
            raw => Some(Self::from_raw(raw)),
        }
    }
}
```

krobelus, line 71: The weird thing about `nix::errno::Errno` is that they treat "unknown error" the same as "no error".
This could cause subtle issues if the OS ever returns an error that hasn't been added to `nix` yet.
Unless they have a way of guaranteeing consistency with the host libc?

---

I'm not sure `enum` is the right type for `Errno`.
It's already marked `#[non_exhaustive]` so client code can't do exhaustive matching.
I guess they could make it:

```
enum Errno {
    NoError = 0
    EPERM = libc::EPERM,
    ...
    UnknownErrno(i32),
}
```

but at that point it's probably better to use a struct that wraps `libc::c_int` (or `i32`?), similar to `std::io::Error`.

---

I guess we could use the `nix::errno::Errno::last_raw()` (which is `nix::errno::errno()` today) and `nix::errno::Errno::set_raw()` precisely in the cases where we don't want to ignore unknown errors.
But the problem of being easy to misuse remains.

faho:
Okay, do I understand correctly that the state here is that we're waiting for nix to make an unknown error detectable?

This has quite a few conflicts, so I'm inclined to close it for now - that's not to disparage the good work here, but simply acknowledging that it's not useful to keep it open.

krobelus:
I don't think this is blocked on Nix (or anything else), at least not for the 95% of cases where we only handle a set of concrete errors.
The one case where need to perfectly forward arbitrary (potentially unknown) errnos we can keep as it is today. Or if nix provides an API, use that. Personally I don't really like the enum approach but it's probably the best compromise in practice.

But yeah, should probably close this until we resume working on it.

PolyMeilex:
Yep, fine to close this one, I just went with a different approach, of resultifying APIs one by one, and therefore minimizing the use of errno in the first place.

PR #10198, opened 2024-01-08
Use Rust for executables

zanchey:
Drops the C++ entry points and restructures the Rust package into a library and three binary crates.

Renames the fish-rust package to fish.

Excises all the carefully-added visibility/privacy control to allow access across multiple crates.


fish-rust/src/abbrs.rs
```diff
@@ -28,7 +28,7 @@ pub fn abbrs_get_set() -> MutexGuard<'static, AbbreviationSet> {
     ABBRS.lock().unwrap()
 }
 
-/// Controls where in the command line abbreviations may expand.
+// Controls where in the command line abbreviations may expand.
```
krobelus, line 31: > They confuse the test drivers

How? Doc comments should be standard Rust so I'm surprised they cause problems.

CMakeLists.txt
```diff
@@ -146,33 +146,30 @@ include(cmake/PCRE2.cmake)
...
   ${CURSES_LIBRARY} ${CURSES_EXTRA_LIBRARY} Threads::Threads ${CMAKE_DL_LIBS}
-  ${PCRE2_LIB} ${Intl_LIBRARIES} ${ATOMIC_LIBRARY}
-  "fish-rust")
+  ${PCRE2_LIB} ${Intl_LIBRARIES} ${ATOMIC_LIBRARY})
 target_include_directories(fishlib PRIVATE
   ${CURSES_INCLUDE_DIRS})
 
+set_property(TARGET fishlib PROPERTY fish_extra_cpp_files
+    "${fish_autocxx_gen_dir}/cxx/gen0.cxx"
+)
```
krobelus, line 164: `fish_extra_cpp_files` is stale, it has been deleted already

fish-rust/src/abbrs.rs
zanchey: Most of them were fine, but there are a few that actually look like tests, but were either not syntactic Rust:

https://github.com/fish-shell/fish-shell/blob/29bd6eebd038a2984ac90ad5b4e6b3c8a88affed/fish-rust/src/tokenizer.rs#L886
(I'm not sure why this one was picked up as a code block)

Or didn't compile:

https://github.com/fish-shell/fish-shell/blob/29bd6eebd038a2984ac90ad5b4e6b3c8a88affed/fish-rust/src/common.rs#L1793-L1804

krobelus: So it looks like turning the crate from `staticlib` to `rlib` enabled `cargo test --doc`. Interesting.

Searching for

    /// ```

shows only twelve occurrences.

Additionally there are like one or two `///    ` in a new paragraph (probably it uses Markdown rules for code blocks).

I think we should rather fix those few places by removing the backticks, and for the pipe one delete the blank line or unindent it.

Doc comments are not super important but they can be nice (in LSP hover info, completion labels etc). Doc tests are also nice because they show compile-checked usage examples.

zanchey: Ah, I didn't know that's where they showed up! OK, I will do that instead.

zanchey: I think let's just disable them for now. That will speed things up too.

Xiretza: Code blocks in doc comments that shouldn't become doctests can be tagged accordingly: https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#attributes

Cargo.toml
```diff
@@ -50,8 +50,21 @@ cc = { git = "https://github.com/mqudsi/cc-rs", branch = "fish" }
 rsconf = { git = "https://github.com/mqudsi/rsconf", branch = "master" }
 
 [lib]
-crate-type = ["staticlib"]
+crate-type = ["rlib"]
 path = "fish-rust/src/lib.rs"
+doctest = false
+
+[[bin]]
+name = "fish"
```
krobelus, line 58: I have a slight preference for`src/fish.rs` over `src/bin/fish.rs` because the `fish` prefix already implies it's an entrypoint. Not a strong opinion though.

Cargo.toml
```diff
@@ -50,8 +50,21 @@
...
+path = "fish-rust/src/bin/fish.rs"
+
+[[bin]]
+name = "fish_indent"
+path = "fish-rust/src/bin/fish_indent.rs"
+
+[[bin]]
+name = "fish_key_reader"
```
krobelus, line 66: This introduces a few compile warnings because attributes like `#![allow(non_camel_case_types)]` from `fish-rust/src/lib.rs` apply only to the lib crate.
I don't think there is a way to share those attributes across crates.
It would be good to fix either by copying the attributes to the `fish_*.rs` files or fixing the underlying problems.

Cargo.toml
```diff
@@ -50,8 +50,21 @@
 rsconf = { git = "https://github.com/mqudsi/rsconf", branch = "master" }
 
 [lib]
-crate-type = ["staticlib"]
+crate-type = ["rlib"]
 path = "fish-rust/src/lib.rs"
+doctest = false
```
krobelus, line 55: every commit should pass tests, so this should be reordered or squashed (same for the other commits)

fish-rust/src/common.rs
```diff
@@ -1177,7 +1179,7 @@ pub fn cstr2wcstring(input: &[u8]) -> WString {
     str2wcstring(&input[0..strlen])
 }
 
-pub fn charptr2wcstring(input: *const libc::c_char) -> WString {
+pub unsafe fn charptr2wcstring(input: *const libc::c_char) -> WString {
```
krobelus, line 1182: I think we can avoid the need for `unsafe` by using `pub(crate)` instead of `pub` (preserving existing meaning)

Cargo.toml
Xiretza: > I don't think there is a way to share those attributes across crates.

There is, actually, thanks to the relatively new [`[lints]` table](https://doc.rust-lang.org/cargo/reference/manifest.html#the-lints-section).

krobelus:
I wonder what the CI failure is about, I believe the build processes should all have `CURSES_LIBRARY_LIST` set. If that's true they should link against curses. Else maybe there's a race condition because we're writing that to a file without a lock

Cargo.toml
krobelus: great. I guess we don't want to bump the version yet

faho: Would adding that section break *building* with older versions?

Our lint-CI already runs stable, and any one of us will presumably have >=1.74.

So the only thing left would be warnings for our other CI and people who actually need an older release - and even in the C++ version people with older compilers would get warnings.

cmake/Install.cmake
```diff
@@ -41,15 +41,6 @@ set(extra_confdir
...
 
 
-corrosion_set_env_vars(${fish_rust_target}
-    "PREFIX=${prefix}"
-    # Temporary hack to propogate CMake flags/options to build.rs.
-    "CMAKE_WITH_GETTEXT=${CMAKE_WITH_GETTEXT}"
-    "DOCDIR=${CMAKE_INSTALL_FULL_DOCDIR}"
-    "DATADIR=${CMAKE_INSTALL_FULL_DATADIR}"
-    "SYSCONFDIR=${CMAKE_INSTALL_FULL_SYSCONFDIR}"
-    "BINDIR=${CMAKE_INSTALL_FULL_BINDIR}"
```
krobelus, line 51: @faho do we need this? Maybe for builds that use a custom `SYSCONFDIR`?

faho: Sorry, I don't quite recall the context here.

I *think* it was an issue with fish not finding the config when run from the build directory?

faho: I cannot find an issue, I can both run fish from the build directoy (and `$__fish_sysconf_dir` is effectively `../etc`), and I can `sudo ninja install` (and `$__fish_sysconf_dir` is /usr/local/etc with a prefix of /usr/local).

faho:
> fixed ubuntu build but asan is failing now

See https://github.com/fish-shell/fish-shell/blob/b82dad0160db2dfd86d05dd45c7afdc9701af2ef/build_tools/lsan_suppressions.txt, it appears curses likes to leak.

Not sure why that's not applying, maybe 244d88298491ede71e1fda2646a155e160878f2d?

PR #10187, opened 2024-01-03
Log original exit code used when a builtin returns a negative exit code

tyilo:
## Description

In fish 3.7.0 executing `exit` with a negative argument results in:
```
$ fish -c 'exit -123'
warning: builtin exit returned invalid exit code 133
```

It would make more sense if the error was:
```
$ fish -c 'exit -123'
warning: builtin exit returned invalid exit code -123
```

## TODOs:

- [ ] Changes to fish usage are reflected in user documentation/manpages.
- [x] Tests have been added for regressions fixed
- [ ] User-visible changes noted in CHANGELOG.rst

PR #10199, opened 2024-01-09
Fix crash when running ELF w/ interpreter missing

ksyx:
## Description

The function `stat` as defined in `include/x86_64-linux-gnu/sys/stat.h` marks its arguments as nonnull as in below. This UB causes crash in release builds with variable `interpreter` assumed to be nonnull. Along with failing `stat` returning nonzero value, this ultimately causes `strlen` to be called with NULL as argument. This PR fixes the problem by prefixing `stat` call with a check to avoid calling it with NULL argument.

Definition of `stat`:
```c
extern int stat (const char *__restrict __file,
		 struct stat *__restrict __buf) __THROW __nonnull ((1, 2));
```
Reproduce:
```fish
> # interp.c is any vaild single file C source
> gcc ./interp.c -Wl,--dynamic-linker=/bad -o interp
> echo './interp' > in.txt
> ./fish < in.txt
'./fish < in.txt' terminated by signal SIGSEGV (Address boundary error)
```

Problem identified and fixed together with @moodyhunter

## TODOs:

  - No change to usage
  - No user-visible change

### DIFF
--- a/src/postfork.cpp
+++ b/src/postfork.cpp
@@ -463,7 +463,7 @@ void safe_report_exec_error(int err, const char *actual_cmd, const char *const *
             const char *interpreter =
                 get_interpreter(actual_cmd, interpreter_buff, sizeof interpreter_buff);
             struct stat buf;
-            auto statret = stat(interpreter, &buf);
+            auto statret = !interpreter || stat(interpreter, &buf);
             if (interpreter && (0 != statret || access(interpreter, X_OK))) {
                 // Detect Windows line endings and complain specifically about them.
                 auto len = strlen(interpreter);


zanchey:
I can't reproduce this crash with 3.6.0, 3.7.0 or master on Linux. What environment are you using?

(I get the expected `exec: Failed to execute process './interp': The file exists and is executable. Check the interpreter or linker?`)

(I probably wouldn't roll a 3.7.1 for this fix alone.)

moodyhunter:
Hi @zanchey, it is an Arch Linux build with GCC 13:

```
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-pc-linux-gnu/13.2.1/lto-wrapper
Target: x86_64-pc-linux-gnu
Configured with: /build/gcc/src/gcc/configure --enable-languages=ada,c,c++,d,fortran,go,lto,objc,obj-c++ --enable-bootstrap --prefix=/usr --libdir=/usr/lib --libexecdir=/usr/lib --mandir=/usr/share/man --infodir=/usr/share/info --with-bugurl=https://bugs.archlinux.org/ --with-build-config=bootstrap-lto --with-linker-hash-style=gnu --with-system-zlib --enable-__cxa_atexit --enable-cet=auto --enable-checking=release --enable-clocale=gnu --enable-default-pie --enable-default-ssp --enable-gnu-indirect-function --enable-gnu-unique-object --enable-libstdcxx-backtrace --enable-link-serialization=1 --enable-linker-build-id --enable-lto --enable-multilib --enable-plugin --enable-shared --enable-threads=posix --disable-libssp --disable-libstdcxx-pch --disable-werror
Thread model: posix
Supported LTO compression algorithms: zlib zstd
gcc version 13.2.1 20230801 (GCC) 
```

Passing NULL to `stat` itself is already an undefined behavior, regardless of whether fish crashes or not :)

faho:
I can reproduce this and I don't see any reason not to merge it.

We already have #10187 on Integration_3.7.0, so if we do end up making another C++-release this'll be included. Thanks!

Issue #5063, opened 2018-06-20
pdf version of the documentation


sabinem:
I just feel the documentation of `fish` would urgently need a PDF version. I could not find one.

Have you ever thought about porting the documentation to Read The Docs? May be this is not a good idea, it is just that you all the things such as PDF version and search field and a bunch of formatting options for free there.

faho:
What specifically would a PDF version do better than the existing formats?

>Have you ever thought about porting the documentation to Read The Docs?

From my research, we wouldn't so much port it to "Read The Docs" as we'd port it to markdown or rst. Which isn't something I'm opposed to - in fact I think nobody is in love with the current doxygen-based system - but which would be a whole bunch of work.

It's also important to us that all of the docs be available locally - so you don't have to have an internet connection just to be able to understand your shell. Is that possible with Read The Docs?

sabinem:
@faho:  

> What specifically would a PDF version do better than the existing formats?

I usually like to download documentation on my ipad and read it as a pdf when I am traveling or commuting. I feel it allows for a more focused reading experience.

>  as we'd port it to markdown or rst ... but which would be a whole bunch of work

Yes that is probably true, there might be tools for this. That would have to be researched.

> It's also important to us that all of the docs be available locally - so you don't have to have an internet connection just to be able to understand your shell. Is that possible with Read The Docs?

For now as far as I know Read The Docs is not immediately available offline: it offers an html version, but people would have to download it themselves.

mqudsi:
fwiw since Github is huge on markdown, pandoc can do markdown to pdf, html, and more.

reed9999:
> What specifically would a PDF version do better than the existing formats?

Asked 6 years ago, I know, but here's my use case: I often use Android apps to read documentation to me when I'm able to listen but not read my screen, with mixed results. For example sometimes I use the `@Voice` app. If I have to "share" every HTML page separately to the app then it may not be worth the bother. If I can give it a PDF and let it read for several minutes, then it's suitable for talking a walk or drive. 

To be sure not all doc makes sense when read aloud, but at least a single file lets me listen seamlessly to those portions that make sense when read.

zanchey:
This feels like it should not be that hard to do with the Sphinx toolchain, but I have not yet been successful.

gaoqiangks:
Is there a pdf version now?

faho:
You *can* build your own:

```fish
sphinx-build -M latexpdf doc_src/ /path/to/output
```

[...]

PR #10200, opened 2024-01-09
doc_src: add print media support to the Sphinx theme

perillo:
## Description

Update the pydoctheme.css file to add support for print media.

The code was adapted from the existing support for screens that are less than 700px wide, with the following changes:

  - Remove the documents and sections index
  - Remove the quick search
  - Remove dead CSS code

Additionally, add section numbers and ensure that code blocks are never split across multiple pages.

Since it is now possible to generate good PDF documents using the browser, maybe #5063 can be closed.  This PR **does not** close it.

## NOTES

While reading the existing code I found that:
  - The `div.content` selector does not match any element
  - The `div.sphinxsidebarwrapper > h3:nth-child(5)` also seems to be dead code

I checked the generated files, and I'm quite sure (but not 100%).

### DIFF
--- a/doc_src/python_docs_theme/static/pydoctheme.css
+++ b/doc_src/python_docs_theme/static/pydoctheme.css
@@ -611,6 +611,68 @@ div.documentwrapper {
     }
 }
 
+/* On print media remove anything non-essential. */
+@media print {
+    .inline-search {
+        display: none;
+    }
+
+    div.sphinxsidebar {
+        display: none;
+    }
+
+    div.bodywrapper {
+        margin-left: 0;
+    }
+
+    div.sphinxsidebar ul {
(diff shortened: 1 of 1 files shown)


faho:
> The div.content selector does not match any element
  The div.sphinxsidebarwrapper > h3:nth-child(5) also seems to be dead code


Seems like that to me as well.

> Since it is now possible to generate good PDF documents using the browser, maybe https://github.com/fish-shell/fish-shell/issues/5063 can be closed. This PR does not close it.

I don't think it should. Going through the docs page by page and manually printing to get 135 separate documents isn't exactly a great experience.

We should look at some of the sphinx-to-pdf extensions again.

Anyway, thanks, merged!

Issue #6901, opened 2020-04-15
is support of dot env files in the roadmap?


shouze:
I'm pretty sure this issue has already been discussed but [I didn't find any related issue](https://github.com/fish-shell/fish-shell/issues?page=14&q=is%3Aissue+sort%3Aupdated-desc+env+file+is%3Aclosed).

Currently I'm able to write this kind of `.env` file:
```dotenv
set -Ux SOME_VAR some_value
```
And source it like this:
```fish
source .env
```

But... this is not very interoperable. I would prefer from far using the conventional form of dotenv files:
```
SOME_VAR=some_value
```

It remains of course the question of the scope. To me, flags should be `-Ux` in this precise case but this is of course subject of debate.

Anyway this should imply that the fish source builtin should be able to work with 2 parsing modes. To avoid over engineering on that, I guess it could be a good idea to put a special pragma instruction on the first line of the .env file to explicit the parsing mode. Something like that:
```dotenv
#pragma: fish.source.interoperable_format
SOME_VAR=some_value
```

We can even imagine to configure set flags with this pragma directive. What do you think?

faho:
It is not, and it won't be.

The reason being that it's neither in super widespread use nor super useful, we don't want fish going around sourcing random builtins or add multiple ways to define variables.

Also it's quite simple to implement yourself. A rough sketch:

```fish
function readenv --on-variable PWD
    if test -r .env
        while read -l line
            set -l kv (string split -m 1 = -- $line)
            set -gx $kv # this will set the variable named by $kv[1] to the rest of $kv
        end
   end
end
```

Note also that setting a *universal* variable is basically wrong by definition here - universal variables are shared by all of your running fishes, but $PWD *isn't*. You want a global variable.

faho:
Note also that [direnv](https://direnv.net/) has support for fish.

shouze:
Thank you, closing this one.

shouze:
@faho BTW in my specific case I want a universal variable because I set it in a Makefile and I want that this variable to be propagated to my parent fish process.

But... yes another way to do is to use direnv, it's more like dependency inversion and far more scoped in the right way, thank you for your time and precious advices.

vagoff:
```
function parse-env
    for line in (cat $argv[1])
        set line (string trim $line)
        if test -z $line
        or string match -q "#*" $line
            continue
        end
        set name_value (string split -m 1 = $line)
        set name (string trim $name_value[1])
[...]

PR #10203, opened 2024-01-09
feat: Support for built-in loading of dotenv files

Bit0r:
Support for built-in loading of dotenv files, making Fish more user-friendly for beginners.

#6901 

## Description

Allowing users to use dotenv without installing additional tools. This reduces confusion for beginners around concepts like dotenv files and environment variables.


faho:
Sorry, I do not believe we want to add this.

PR #10210, opened 2024-01-12
Remove widestring-suffix

faho:
## Description

This removes the widestring-suffix crate, which I don't believe is pulling its weight.

In particular it seems to inhibit cargo's view into `#[widestrs]` annotated functions, so now clippy lints start showing up, and I've had compiler errors fail to figure out the correct location due to it in the past.

## TODOs:
A lot of this is sed-patched and I had one accident with a mistaken regex that I hope I undid. I've had multiple looks over it, but it's possible I overlooked something.


faho:
Also the tests take a sizable chunk of the total test time - I've seen them take over 30 seconds on Github Actions, and I'm reasonably sure the cargo tests are run after the rest, so that's 30 seconds serialized.

fish-rust/src/builtins/math.rs
```diff
@@ -189,24 +187,26 @@ fn evaluate_expression(
 
             streams
                 .err
-                .append(sprintf!("%ls: Error: %ls\n"L, cmd, error_message));
-            streams.err.append(sprintf!("'%ls'\n"L, expression));
+                .append(sprintf!(L!("%ls: Error: %ls\n"), cmd, error_message));
+            streams.err.append(sprintf!(L!("'%ls'\n"), expression));
```
krobelus, line 191: while we are at I think we should remove it from printf format strings; those magically work with plain literals. Same elsewhere.

fish-rust/src/env/environment_impl.rs
```diff
@@ -339,57 +339,61 @@ impl EnvScopedImpl {
...
 
-    #[widestrs]
     fn try_get_computed(&self, key: &wstr) -> Option<EnvVar> {
         let ev = ElectricVar::for_name(key);
         if ev.is_none() || !ev.unwrap().computed() {
             return None;
         }
 
-        if key == "PWD"L {
+        if key == L!("PWD") {
```
krobelus, line 348: these comparisons should work fine with plain string literals, which is preferrable. Same elsewhere.

fish-rust/src/builtins/wait.rs
```diff
@@ -177,15 +176,16 @@ pub fn wait(parser: &Parser, streams: &mut IoStreams, argv: &mut [&wstr]) -> Opt
 
     // Get the list of wait handles for our waiting.
     let mut wait_handles: Vec<WaitHandleRef> = Vec::new();
-    for i in w.woptind..argc {
-        if iswnumeric(argv[i]) {
+    let optind = w.woptind;
+    for item in argv.iter().take(argc).skip(optind) {
```
krobelus, line 180: `argc` is not the length, so the equivalent would be `take(argc - optind)`.

better to write `for arg in &argv[optind..argc]`. Clippy not smart enough

krobelus:
I think this change is good.
It makes string tables a bit worse but that's about the only place where they shine and we don't spend much time reading string tables.

In tests I like to define a macro that adds the `L!()` automatically.

fish-rust/src/env/environment_impl.rs
faho: Done, except for one - https://github.com/fish-shell/fish-shell/blob/ca972f6e0fbc3dc236e68b9ebfcbe2deb4a48be1/src/future_feature_flags.rs#L189

This gives "^^ no implementation for `[char] == str`". So that one gets a `L!()`.

fish-rust/src/builtins/wait.rs
faho: Blergh.

I *think* I stumbled over that one before? I probably should trust clippy less.

Issue #10194, opened 2024-01-06
Fish tries to substitute (execute) a subshell command during completion (tab) even when surrounded with double quotes


aureq:
### What happened?

I have a command that I have a shell completion for (`grafana-manager`) and for which the shell completion is generated by `github.com/spf13/cobra` (Golang).

If I run my command as show below, then it work as expected (no error message) and the program receives the the `--dashboard-name` value as written on the shell prompt.

```fish
./grafana-manager public-dashboards delete --organization-id 3 --dashboard-name "k8s (public)"
```

However, if I attempt to add further complete my command by pressing `tab`, `fish` appears to try to execute the command `public`. This results in `fish` showing errors on the terminal and the completion to fail.

```
fish: Unknown command: public
in command substitution
        called on line 1 of file -
in command substitution
        called on line 25 of file -
in function '__grafana_manager_perform_completion'
        called on line 1 of file -
in command substitution
        called on line 68 of file -
in function '__grafana_manager_prepare_completions'
in command substitution
- (line 1): Unknown command
GRAFANA_MANAGER_ACTIVE_HELP=0 ./grafana-manager __complete public-dashboards delete --organization-id 3 --dashboard-name k8s (public) --
                                                                                                                             ^~~~~~~^
in command substitution
        called on line 25 of file -
in function '__grafana_manager_perform_completion'
        called on line 1 of file -
in command substitution
        called on line 68 of file -
in function '__grafana_manager_prepare_completions'
in command substitution
```

### Output of `fish --version`

`3.7.0`

### Additional context

This should quite dangerous situation if the provided value between the parenthesis was to do something _nasty_.

If I escape the parenthesis, then the shell completion works as expected ✅ however, running the command fails ❌ because the (escaped) value supplied to the program is not as expected.

```
./grafana-manager public-dashboards delete --organization-id 3 --dashboard-name "k8s \(public\)" --[tab]
```

zanchey:
Can you post the output of the completions file that cobra is generating?

aureq:
So, I'm using `github.com/spf13/cobra v1.5.0` (upgrade to 1.8.0 is planned)

I usually inject the completion via `./grafana-manager completion fish | source`.

And below is the actual completion.

```
# fish completion for grafana-manager                      -*- shell-script -*-

function __grafana_manager_debug
    set -l file "$BASH_COMP_DEBUG_FILE"
    if test -n "$file"
        echo "$argv" >> $file
    end
end

[...]

PR #10212, opened 2024-01-14
builtin commandline: --tokenize to expand strings, to avoid need for eval

krobelus:
NOTE for 3rd party completion scripts: you should migrate from "commandline -o" to "commandline -x". For example use this to work with both fish before and after this change:

    set -l tokens
    if commandline -x >/dev/null 2>&1
        set tokens (commandline -xpc)
    else
        set tokens (commandline -opc)
    end

---

Issue #10194 reports Cobra completions do

    set -l args (commandline -opc)
    eval $args[1] __complete $args[2..] (commandline -ct | string escape)

The intent behind "eval" is to expand variables and tildes in "$args".
Fair enough. Several of our own completions do the same, see below.

The problem with "commandline -o" + "eval" is that the former already removes
quotes that are significant for "eval". This becomes a problem if $args
contains quoted () or {}, for example this command will wrongly execute a
command substituion:

    git --work-tree='(launch-missiles)' <TAB>

It is possible to escape the string the tokens before running eval, but
then there will be no expansion of variables etc.  The problem is that
"commandline -o" only unescapes tokens so they end up in a weird state
in-between source and expanded version.

Remove the need for "eval" by making "commandline -o" expand things like
variables and braces I don't think there is any use case where not expanding is
better. This enables custom completion scripts to be aware of shell variables
without eval, see the added test for completions to "make -C $var/some/dir ".

Some completion scripts use eval without escaping the tokens.  This is
already broken in some ways but this changes adds more cases of breakage :
when a variable expansion contains spaces or other shell metacharacters,
"eval" will do the wrong thing.
There are some third-party uses of "eval (commandline -opc)".  Besides Cobra
that's at least [pip](https://github.com/pypa/pip/pull/9727).

The new expansion skips command substitutions, in the hope that this matches
user expectation better than running them. They are passed through as-is
(instead of cancelling or expanding to nothing) to make custom completion
scripts work well in the common case.


faho:
I'm in favor of having a way to get the expanded tokens, but I'm not sure

1. this is correct in general for *all* uses of `commandline --tokenize` - consider bindings
2. this is okay for completions that currently use `eval` - like [our git completions](https://github.com/fish-shell/fish-shell/blob/10891d29851847d6485bef05df5bb09368ee016d/share/completions/git.fish#L20) (tho they do try to avoid some obvious issues)

The latter one is what you're aluding to with `if a variable expansion contains spaces or
other shell metacharacters, "eval" will do the wrong thing`.

The change here is that *now* you get double-expansion - `$bar"foo"` with `$bar` set to `(echo hahaha)` will turn into `(echo hahaha)foo` and then pass to `eval` and actually run the `echo hahaha` that wasn't run before.

The obvious alternative is to add a second option, `commandline --expanded-tokens`, or `commandline --expand --tokenize`.

krobelus:
> I'm in favor of having a way to get the expanded tokens, but I'm not sure
> 
> 1. this is correct in general for *all* uses of `commandline --tokenize` - consider bindings

I don't think we can find a case where the current behavior is more useful.

> 2. this is okay for completions that currently use `eval` - like [our git completions](https://github.com/fish-shell/fish-shell/blob/10891d29851847d6485bef05df5bb09368ee016d/share/completions/git.fish#L20) (tho they do try to avoid some obvious issues)
> 
> The latter one is what you're aluding to with `if a variable expansion contains spaces or
> other shell metacharacters, "eval" will do the wrong thing`.
> 
> The change here is that *now* you get double-expansion - `$bar"foo"` with `$bar` set to `(echo hahaha)` will turn into `(echo hahaha)foo` and then pass to `eval` and actually run the `echo hahaha` that wasn't run before.

I think `eval (commandline -opc)` is not a very common idiom because `eval` is bad practice here. It was always a workaround.
Completions that use `eval` were already vulnerable but yeah, this change amplifies it.
But pushing people to get off `eval` for completions is a good thing.

> 
> The obvious alternative is to add a second option, `commandline --expanded-tokens`, or `commandline --expand --tokenize`.

FWIW, if we want two modes, then, besides an expansion mode we could have one that gives raw tokens (without today's bogus unescaping). But I wouldn't add that without a use case and I don't think we can find one.

My hope is that changing existing behavior will be a net positive because it gives the improved behavior to third party completions for free.
Most "forwarding" uses of `commandline -opc` do not call `eval` but forward them as unexpanded argv, like `myprogram complete (commandline -opc)`.
The expansion behavior is a strict improvement here.
For example Rust's `clap` library provides dynamic completions with

	r#"complete -x -c {bin} -a "("'{COMPLETER}'" complete --shell fish -- (commandline --current-process --tokenize --cut-at-cursor) (commandline --current-token))""#

Essentially everything that is broken by this change was already broken in similar ways; we can fix those first.
So far we only know of Cobra. Once they merge https://github.com/spf13/cobra/pull/2095/files, their logic will be safe both before and after this change.

I'm not really sure how to search for third party code. What command line completion frameworks other than clap/Cobra, or other notable third party scripts can we find?

faho:
> What command line completion frameworks other than clap/Cobra, or other notable third party scripts can we find?

The pip completions [also eval](https://github.com/pypa/pip/blob/5efa3e83d90b1d64fbf0104d2ba84bef8d7e9648/src/pip/_internal/commands/completion.py#L41-L46) the output of `commandline -o`.

(for the record, pipenv does not, at least the version I have)

rclone currently does, but it uses cobra. I'm not sure what updating them would entail.

stack doesn't, starship doesn't.

Those are the first five of our `| source` completions I could get my hands on.

krobelus:
I have made PRs to pip and cobra.
I think it's fair to maintain compatibilty. Though third party scripts will probably never upgrade.

krobelus:
I prefer to break users here (short sharp shock) but it sounds like there's an objection so I'll add a separate flag?

faho:
> I prefer to break users here (short sharp shock) but it sounds like there's an objection so I'll add a separate flag?

I would prefer that, yes. The issue is that it wouldn't *be* a "short sharp shock", it would appear to work in most cases, and then in edge cases you get a face full of `eval`.

And that's not just for completion generation projects, but also e.g. users who keep a modified version of our own git completions.

faho:
Note: For *bindings* we probably need a "raw" mode, where you get the full token with quotes and such intact.

Otherwise if you try to make a binding act on tokens it will be wrong as soon as any quotes enter the picture.

Now we have three possible modes, which makes me think we might want it to be one option with arguments.

faho:
> I don't know of a use case for getting the list of raw tokens but it does make much more sense than unescaped ones.

"Insert the last token again".

Currently my best attempt is

```fish
function last_arg
    set -l arg (commandline -opc)[-1]
    commandline -i -- $arg
end

bind \cg last_arg
```

but that fails if the last token is quoted, like `"foo bar"`. Add `string escape` and then it'll turn it into `'foo bar'` with single-quotes.

Generally any time you want to iterate over the commandline to transform it you need it not unescaped.

> I realized read --tokenize is weird in the same way as commandline --tokenize; it does unescaping but leave variables.

`read --tokenize` is mainly useful for reading quotes - turn `"foo bar" baz` into `foo bar` and `baz`. Picture reading from an ini-style file, or we use it for figuring out $EDITOR
 or $PAGER, where you might have escaped spaces (non-standard but they do work in some tools, I think systemd had some pretty complicated logic here).

In those cases you typically do not want variables and globs to be expanded.

krobelus:
both makes sense, thanks.
So `commandline` flags `--expand-tokens` and `--raw-tokens` or `--tokens=expanded` and `--tokens=raw`. But if we do the latter `-x` should still be an abbreviation for `--tokens=expanded`

faho:
> So commandline flags --expand-tokens and --raw-tokens or --tokens=expanded and --tokens=raw. But if we do the latter -x should still be an abbreviation for --tokens=expanded

The latter sounds fine. Closest short opt we can do for "raw" is `-R` or `-w`.

PR #4713, opened 2018-02-07
Control-C to simply clear the line again

ridiculousfish:
In commit 5f849d02646b, issue #2904, the control-C behavior was changed to
print an inverted ^C and then a newline. This behavior has not been well-received, judging by positive reactions for how to undo it ([1](https://github.com/fish-shell/fish-shell/issues/2904#issuecomment-261759961), [2](https://github.com/fish-shell/fish-shell/issues/3535#issuecomment-259538701).

This resets the behavior to the original behavior of clearing the command line. It keeps the `__fish_cancel_commandline` override point because it's now invoked from three places, so it makes sense to keep the behavior centralized.

PR #10213, opened 2024-01-14
Control-C to simply clear commandline buffer again

krobelus:
Commit 5f849d0 changed control-C to print an inverted ^C and then a newline.

The original motivation was

> In bash if you type something and press ctrl-c then the content of the line
> is preserved and the cursor is moved to a new line. In fish the ctrl-c just
> clears the line. For me the behaviour of bash is a bit better, because it
> allows me to type something then press ctrl-c and I have the typed string
> in the log for further reference.

This sounds like a valid use case in some scenarios but I think that most
abandoned commands are noise. After all, the user erased them. Also, undo
can be used to get back a limited set of canceled commands.

I believe the original motivation for existing behavior (in other shells) was
that TERM=dumb does not support erasing characters. Similarly, other shells
like to leave behind other artifacts, for example when using tab-completion
or in their interactive menus but we generally don't.

Control-C is the obvious way to quickly clear a multi-line commandline.
IPython does the same. For the other behavior we have Alt-# although that's
probably not very well-known.

Restore the old Control-C behavior of simply clearing the command line.

Our unused __fish_cancel_commandline still prints the ^C. For folks who
have explicitly bound ^C to that, it's probably better to keep the existing
behavior, so let's leave this one.

Previous attempt at #4713 fizzled.

TODO: changelog, instructions to restore the other behavior.


faho:
I'm fine either way. I will note that this is nicer now that we have undo, so you can actually get the old commandline back by pressing ctrl-z.

PR #10218, opened 2024-01-15
Support command for jumping to word location based on key

freqmod:
## Description

I have added a command that can be used to jump to a word in the command line. Each word is represented with a letter and pressing that letter jumps to that word. It can also be combined with visual mode to select to that word.  I have made a quick video outlining its operation.

[recording.webm](https://github.com/fish-shell/fish-shell/assets/60908/601b34d7-7061-46ab-bb66-e90a73740635)

I have implemented something similar for my fork of the helix editor (it uses partly the same code and is available on my github account)  and used it for a few moths now. I find it very useful instead of using vim like find to command when the character in question is repeated (like the / in paths).

I am adding this pull request as a start of a discussion on how to implement this and I am open to adapting it more to match the architecture of the reader subsystem of fish if necessary.

I use the following fish config with the command (since i use colemak the default variables are set up for that, but I have tried to make a qwerty variant for users of less esoteric keyboard layouts)

```
set fish_jump_anchors_before 'tsradpfwvcxgqbz54321TSRADPFWVCXGQBZ!@#$%'
set fish_jump_anchors_after 'neiohluym,.kj;09876NEIOHLUYM<?KJ:^&*()'
set fish_jump_anchors_before_qwerty 'fdsagrewcxztqvz54321FDSAGREWCXZTQVZ%$#@!'
set fish_jump_anchors_after_qwerty 'jkl;hoiu.,mpyn/09876-=JKL:HOIU><MPYN)(*&^_+'

bind -M default \co 'move-jump-anchor'
bind -M insert \co 'move-jump-anchor'
bind -M visual \co 'move-jump-anchor'
bind \co 'move-jump-anchor'
```

I can help adding documentation and changelog once I know that this is something that is found worthwhile and that the shape of the feature won't change too much.

## TODOs:

- [ ]

### DIFF
--- a/src/input_common.rs
+++ b/src/input_common.rs
@@ -111,6 +111,7 @@ pub enum ReadlineCmd {
     DisableMouseTracking,
     // ncurses uses the obvious name
     ClearScreenAndRepaint,
+    MoveJumpAnchor,
     // NOTE: This one has to be last.
     ReverseRepeatJump,
 }
--- a/src/input.rs
+++ b/src/input.rs
@@ -10,6 +10,7 @@ use crate::parser::Parser;
 use crate::proc::job_reap;
 use crate::reader::{
     reader_reading_interrupted, reader_reset_interrupted, reader_schedule_prompt_repaint,
+    reader_set_show_overlay_state,
 };
 use crate::signal::signal_clear_cancel;
 #[cfg(test)]
@@ -173,6 +174,7 @@ const INPUT_FUNCTION_METADATA: &[InputFunctionMetadata] = &[
     make_md(L!("kill-selection"), ReadlineCmd::KillSelection),
     make_md(L!("kill-whole-line"), ReadlineCmd::KillWholeLine),
     make_md(L!("kill-word"), ReadlineCmd::KillWord),
+    make_md(L!("move-jump-anchor"), ReadlineCmd::MoveJumpAnchor),
     make_md(L!("nextd-or-forward-word"), ReadlineCmd::NextdOrForwardWord),
     make_md(L!("or"), ReadlineCmd::FuncOr),
     make_md(L!("pager-toggle-search"), ReadlineCmd::PagerToggleSearch),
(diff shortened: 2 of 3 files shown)


src/input_common.rs
```diff
@@ -111,6 +111,7 @@ pub enum ReadlineCmd {
     DisableMouseTracking,
     // ncurses uses the obvious name
     ClearScreenAndRepaint,
+    MoveJumpAnchor,
```
krobelus, line 114: I think this is definitely a problem we should consider addressing.

The classic use case is if you want to edit `NEEDLE` after pasting this command:

	echo '/some/really/long/path/name with spaces/aaa/bbb/ccc' NEEDLE 'other token with spaces' a/b/c/

Today I usually fall back on Alt+e to open my editor.

Possible alternative approaches are
- use Shift+Left and Shift+Right which are bound to `forward-bigword` and `backward-bigword` by default -- I didn't realize that until today
- allow searching within the commandline (Ctrl-S would be the obvious Emacs-binding)
- allow moving by full tokens (the Alt+Shift+b binding will be open once we support CSI u)
- add commands that move by `$COLUMNS/2` or similar

Your approach sounds reasonable as well and the UI is fairly well known and obvious.
Do you have an example use case?

I briefly tried it and it has some glitches on multiline commands (when using \co twice) and the contrast is bad (light theme).

freqmod: Thanks for having a look.

I tried using it on multiline commands, and it does not seem to behave in any unexpected way to me, even when pressing ctrl+o twice, however i do agree that the behaviour probably is undesirable, especially replacing the newlines with jump letters, and thus temporarily altering the line breaking.  

There are also some challenges with regards to the fact that it skips adding jump characters if a word is too short leading to inconsistent marking of characters depending on the location of the cursor.

I haven't really used this feature much in the context of fish yet, so I am not sure how well it fits for command line editing except for jumping in paths as i mentioned, but as I wrote i find it useful in the context of a programming text editor; for example when jumping in a line of code that consists of an  an if-test with lots of brackets, quotes etc.

When it comes to the highlighting colour I agree that the contrast is poor, i just chose the first sensible colour enumeration in the theme I could find. We can change it to something else, or make a completely new color category.

Dominik50111 (commented):
Ghb

PR #10223, opened 2024-01-16
Cache sourced completions

faho:
## Description

We have a lot of completions that look like

```fish
pip completion --fish 2>/dev/null | source
```

That's *fine*, upstream gives us some support.

However, the scripts they provide change very rarely, usually not even every release, and so running them again for every shell is extremely wasteful.

In particular the python tools are very slow, `pip completion --fish` takes about 180ms on my system with a hot cache, which is quite noticeable every time you write "pip" in a fresh shell.

So this does two things:

1. It moves our cache to ~/.cache/fish (insert $XDG_CACHE_HOME as appropriate) and allows easily creating subdirectories
2. It caches these completions by running them once, and storing them in a file in our cache
directory

The first one will also move all pre-existing cached completions - this is basically exclusively package managers, so people will have one awkward hidden file in ~/.cache called something like ".pac-cache.alfa" that is ~300K in size that they'd have to remove. We don't try to because running `rm` like that is iffy.

The second one does cache invalidation via the command's mtime.

We store the mtime of the command we ran, and compare against that for
future runs. If the mtime differs - so if the command was up or
downgraded, we run it again.

But if there is no command, no file to compare mtime to, we can't use the cache.

This PR adopts the cache for all egregious offenders I could find. Mostly those are the python tools - pipenv takes a whopping 320ms on my system just to print 18 lines.

I have tested a lot of the other tools (basically everything I could find in the arch repos), and most are more like ~30ms (starship takes 3ms, deno takes 15ms), in which case the cache is probably better avoided.

## TODOs:

- [?] Tests have been added for regressions fixed
- [ ] This removes a few old workarounds for pip{,env}. Do we still need them? My research tells me pip's bug was fixed in 2017 and pipenv's behavior changed in 2019. Given that these are packaging tools that do cryptography and such... do we still need these?
- [ ] Do we need to remove the old cache files? Or do we add a note in the release notes? At worst this is ~300K of wasted space.
- [x] We should rename the package manager cache files to be without a "." - `~/.cache/fish/pac-cache` would be fine (tbh they never should have been hidden, this is what .cache is *for*!)
- [ ] Are there any edgecases I did not think through?
- [ ] Alternatively: Do we just ship the output ourselves? This is of course awkward to keep up-to-date and may have licensing issues.

### DIFF
--- a/share/completions/pip.fish
+++ b/share/completions/pip.fish
@@ -1,1 +1,2 @@
-pip completion --fish 2>/dev/null | source
+__fish_cache_sourced_completions pip completion --fish 2>/dev/null
+or pip completion --fish 2>/dev/null | source
--- a/share/completions/pip2.fish
+++ b/share/completions/pip2.fish
@@ -1,5 +1,2 @@
-# pip[2|3] emits (or emitted) completions with the wrong command name
-# See discussion at https://github.com/fish-shell/fish-shell/pull/4448
-# and pip bug at https://github.com/pypa/pip/pull/4755
-# Keep this even after pip fix is upstreamed for users not on the latest pip
-pip2 completion --fish 2>/dev/null | string replace -r -- " -c\s+pip\b" " -c pip2" | source
+__fish_cache_sourced_completions pip2 completion --fish 2>/dev/null
+or pip2 completion --fish 2>/dev/null | source
--- a/share/functions/__fish_cache_sourced_completions.fish
+++ b/share/functions/__fish_cache_sourced_completions.fish
@@ -1,0 +1,45 @@
+function __fish_cache_sourced_completions
+    # Allow a `--name=foo` option which ends up in the filename.
+    argparse -s name= -- $argv
+    or return
+
+    set -q argv[1]
+    or return 1
+    
+    set -l cmd (command -s $argv[1])
+    or begin
+        # If we have no command, we can't get an mtime
+        # and so we can't cache
+        # The caller can more easily retry
+        return 127
+    end
+
+    set -l cachedir (__fish_make_cache_dir completions)
+    or return
+
+    set -l stampfile $cachedir/$argv[1].stamp
+    set -l compfile $cachedir/$argv[1].fish
+
+    set -l mtime (path mtime -- $cmd)
+
+    set -l cmtime 0
+    path is -rf -- $stampfile
+    and read cmtime < $stampfile
+
+    # If either the timestamp or the completion file don't exist,
+    # or the mtime differs, we rerun.
+    #
+    # That means we'll rerun if the command was up- or downgraded.
+    if path is -vrf -- $stampfile $compfile || test "$cmtime" -ne "$mtime" 2>/dev/null
+        $argv > $compfile
+        # If the command exited unsuccessfully, we assume it didn't work.
+        or return 2
+        echo -- $mtime > $stampfile
+    end
+
+    if path is -rf -- $compfile
+        source $compfile
+        return 0
+    end
+    return 3
+end
--- a/share/completions/pip3.fish
+++ b/share/completions/pip3.fish
@@ -1,5 +1,2 @@
-# pip[2|3] emits (or emitted) completions with the wrong command name
-# See discussion at https://github.com/fish-shell/fish-shell/pull/4448
-# and pip bug at https://github.com/pypa/pip/pull/4755
-# Keep this even after pip fix is upstreamed for users not on the latest pip
-pip3 completion --fish 2>/dev/null | string replace -r -- " -c\s+pip\b" " -c pip3" | source
+__fish_cache_sourced_completions pip3 completion --fish 2>/dev/null
+or pip3 completion --fish 2>/dev/null | source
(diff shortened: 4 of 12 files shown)


share/functions/__fish_cache_sourced_completions.fish
```diff
@@ -0,0 +1,45 @@
+function __fish_cache_sourced_completions
+    # Allow a `--name=foo` option which ends up in the filename.
+    argparse -s name= -- $argv
+    or return
```
krobelus, line 4: `name` is unused; I guess we don't need it at this point.

Maybe apply `path basename` before adding it to the cache path so this function never accidentally overwrites files outside the cache path.

If we want a full cache key we could serialize the entire argv but I guess that's not necessary.

share/functions/__fish_cache_sourced_completions.fish
```diff
@@ -0,0 +1,45 @@
...
+
+    set -q argv[1]
+    or return 1
+    
+    set -l cmd (command -s $argv[1])
```
krobelus, line 9: maybe also apply `realpath` so the timestamp logic also works if the executable path is a symlink (super rare for the commands in question I guess)

share/completions/pip.fish
```diff
@@ -1 +1,2 @@
-pip completion --fish 2>/dev/null | source
+__fish_cache_sourced_completions pip completion --fish 2>/dev/null
+or pip completion --fish 2>/dev/null | source
```
krobelus, line 2: Looks good to me.
It's probably doable to fix `pip` slowness.
They use `pyparsing`; when I used that it needed 2 seconds to parse 1000 lines.

share/completions/pip2.fish
```diff
@@ -1,5 +1,2 @@
-# pip[2|3] emits (or emitted) completions with the wrong command name
-# See discussion at https://github.com/fish-shell/fish-shell/pull/4448
-# and pip bug at https://github.com/pypa/pip/pull/4755
-# Keep this even after pip fix is upstreamed for users not on the latest pip
-pip2 completion --fish 2>/dev/null | string replace -r -- " -c\s+pip\b" " -c pip2" | source
+__fish_cache_sourced_completions pip2 completion --fish 2>/dev/null
+or pip2 completion --fish 2>/dev/null | source
```
krobelus, line 2: nit: we could probably wrap it in another function that does the
`or pip2 completion --fish 2>/dev/null | source` automagically

share/functions/__fish_cache_sourced_completions.fish
faho: We do `stat` for the mtime, not `lstat`, so we should get the real file.

Even if we didn't, in the worst case we'd be running the command one time too often.

share/functions/__fish_cache_sourced_completions.fish
faho: I mainly wanted to add an argparse here so it's easier to add options later.

I guess I can remove the name, but that would be the sort of thing we would add.

share/completions/pip.fish
faho: Honestly, a *lot* of python commands in particular have startup slowness.

That's also why we try not to launch hg until we absolutely have to.

Not that it matters, we're dealing with these now, as they are.

share/completions/pip2.fish
faho: I don't want to add a second function file for that, and adding it to the main function is awkward because you would have to add it everywhere we return, or wrap it in an if.

krobelus:
> Do we need to remove the old cache files? 

I think it makes sense to remove the old cache file in the code path where we create a new one (but only if the new one didn't exist). I guess we could implement that with an argparse option but I'm not sure if it's worth the extra code.
If we want to be more paranoid we could use the `$__fish_initialized` machinery to run it exactly once but that's surely not worth it.
Going forward we can clear the whole cache/fish directory after interesting changes.

share/completions/pip2.fish
krobelus: ok that's all fine. Maybe if we had anonymous functions.

```fish
set -l lambda $(function --anonymous -V foo; ...; end)
$lamdba || error
```

zanchey:
Yes, I think this is a good approach.

I wouldn't bother deleting the old cache.

It might be worth regenerating the completions once a month or so as a fallback?

faho:
> It might be worth regenerating the completions once a month or so as a fallback?

I would prefer, if there is a problem with generating this cache, that people *notice* and tell us so we can fix it (worst case by reverting this). If we regenerate once a month anyway that makes that harder to do.

krobelus:
we could have a background process check that the cache is still valid, and then print some kind of error message.
But I don't think we have a reason to believe that the cache would go stale. If `pip` crashes due to wrong Python deps we'll not write a cache entry.
(test failure is a fluke right?)

zanchey:
> > It might be worth regenerating the completions once a month or so as a fallback?
> 
> I would prefer, if there is a problem with generating this cache, that people _notice_ and tell us so we can fix it (worst case by reverting this). If we regenerate once a month anyway that makes that harder to do.

OK! SGTM

faho:
Okay I'm going to merge this now. If any issues turn up in master we can fix them there.

PR #10224, opened 2024-01-16
commandline: Add "--current-command"

faho:
This can be used to get the current command, even following decorators like `command` or `exec` or `and`.

That makes it useful e.g. to make abbrs only apply to a specific command.

You can also do `commandline --replace --current-command foo` to change

```
echo bar | git commit
```

to

```
echo bar | foo commit
```

For example, to make an abbreviation expand only for `git` as the command:

```fish
function abbr_git_commit
    test "$(commandline -m)" = git
    or return
    echo commit
end
abbr c --position anywhere --function abbr_git_commit
```

It would still be possible to implement a shortcut for abbr - `abbr --command git c commit`, but this is a nice building block for other things regardless.

We could already claim this handles #9411.

## TODOs:

- [ ] The ast-parsing I added to commandline is mostly duplicated with highlight::autosuggest_parse_command(). We should put it somewhere it can be shared, but I'm not sure where - ast directly?
- [ ] I didn't have this expand the command on purpose, pending #10212. We should follow what happens there (and adjust the docs).
- [ ] I'm not sure I caught all the edge-cases, there's weirdness with foo=bar variable assignments.

### DIFF
--- a/doc_src/cmds/commandline.rst
+++ b/doc_src/cmds/commandline.rst
@@ -70,6 +70,9 @@ The following options change what part of the commandline is printed or updated:
 **-t** or **--current-token**
     Selects the current token
 
+**-m** or **--current-command**
+    Selects just the command token in the current process, without any decorators like ``exec`` or ``and``, unless they are the only thing in the process. This can be empty if the commandline doesn't currently have one. Note that it is *not* expanded, so e.g. ``$PAGER foo`` has a command of ``$PAGER``.
+
 The following options change the way ``commandline`` prints the current commandline buffer:
 
 **-c** or **--cut-at-cursor**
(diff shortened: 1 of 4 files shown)


faho:
Awkward cases if you do `bind \ch 'commandline --current-command --replace git'`: 

`for foo in bar; ... ; end` with the cursor on the "end" will pick the empty spot at the beginning of the line, making "gitfor". That's probably due to the attempt at fixing variable assignments. Not sure what the correct *place* to pick is, maybe replacing should just give up?

`foo=bar` without a space will think the command is empty, replacing will create `foo=bargit`, not `foo=bar git`. It should insert a space.

Command substitutions work, `||` conjunctions work.

faho:
Blergh this apparently needs walking - `if true` fails because it doesn't yet have the decorated statement.

(or I need something better than parse_util_process_extent)

faho:
Lemme close this, maybe I'll get to it later

Issue #9911, opened 2023-07-24
Forbid `&&` after backgrounding


faho:
As discovered on matrix/gitter:

```fish
foo & && disown
```

is seen as syntactically valid, but doesn't do anything sensible. The backgrounded command doesn't set $status and so the `&&` reacts to the previous $status - execute `false` before and `disown` won't be executed.

As a test:

```fish
false
sleep 10 & && echo foo
```

won't print foo, but replace "false" with "true" and it will.

Conceptually this also makes no sense - you can't react to the status of a backgrounded command like that because it doesn't *have* a status once the next command is executed.

So the only sensible thing to do here is to make it an error, like it already is for `and`:

```fish
sleep 10 & and echo foo
```

prints

````
fish: The “and” command can not be used immediately after a backgrounded job
sleep 10 & and echo foo
           ^~^
````

PR #10228, opened 2024-01-17
fix: disallow background operator preceding any conjunction

lavafroth:
## Description

For a given conjunction continuation, we traverse up to its parent if it exists. If any of the continuations need to be backgrounded, we error out stating that token (`&&` or `||`) cannot be used after backgrounding. There is probably a more elegant way to solve this.

Fixes issue #9911

Screenshot_20240117_133225

## TODOs:


src/parse_util.rs
```diff
@@ -1120,6 +1120,17 @@ pub fn parse_util_detect_errors_in_ast(
             if jcc.conjunction.has_source() && jcc.job.try_source_range().is_none() {
                 has_unclosed_conjunction = true;
             }
+            let continuations = jcc
+                .parent()
+                .and_then(|job_conj_list| job_conj_list.as_job_conjunction_continuation_list());
```
krobelus, line 1125: should squash the two commits because the first one was wrong it seems

src/parse_util.rs
```diff
@@ -1120,6 +1120,17 @@ pub fn parse_util_detect_errors_in_ast(
...
+            if let Some(continuations) = continuations {
+                for continuation in continuations.contents() {
```
krobelus, line 1127: use `continuations` instead of `continuations.contents()`, similarly below. It should behave like a regular slice,
`contents()` should probably be private but no need to do that here.

src/parse_util.rs
```diff
@@ -1120,6 +1120,17 @@ pub fn parse_util_detect_errors_in_ast(
...
+                    if continuation.job.bg.is_some() {
```
krobelus, line 1128: looking at only continuations that have `&` is not fully correct because it means we fail to
check for `foo & && bar`. The first command is not a continuation.
Consider these test cases:

```diff
diff --git a/src/parse_util.rs b/src/parse_util.rs
index 2ab2d1001..3d3e11909 100644
--- a/src/parse_util.rs
+++ b/src/parse_util.rs
@@ -1822,7 +1822,7 @@ pub fn parse_util_expand_variable_error(
 }

 /// Error message for use of backgrounded commands before and/or.
-const BOOL_AFTER_BACKGROUND_ERROR_MSG: &str =
+pub(crate) const BOOL_AFTER_BACKGROUND_ERROR_MSG: &str =
     "The '%ls' command can not be used immediately after a backgrounded job";

 /// Error message for backgrounded commands as conditionals.
diff --git a/src/tests/parse_util.rs b/src/tests/parse_util.rs
index 3b6748f5a..d4b7cd14f 100644
--- a/src/tests/parse_util.rs
+++ b/src/tests/parse_util.rs
@@ -5,7 +5,7 @@ use crate::parse_constants::{
     ERROR_NOT_ARGV_AT, ERROR_NOT_ARGV_COUNT, ERROR_NOT_ARGV_STAR, ERROR_NOT_PID, ERROR_NOT_STATUS,
     ERROR_NO_VAR_NAME,
 };
-use crate::parse_util::parse_util_detect_errors;
+use crate::parse_util::{parse_util_detect_errors, BOOL_AFTER_BACKGROUND_ERROR_MSG};
 use crate::tests::prelude::*;
 use crate::wchar::prelude::*;
 use crate::wchar_ext::WExt;
@@ -72,4 +72,9 @@ fn test_error_messages() {
     validate!("echo foo\"$\"bar", ERROR_NO_VAR_NAME);
     validate!("echo \"foo\"$\"bar\"", ERROR_NO_VAR_NAME);
     validate!("echo foo $ bar", ERROR_NO_VAR_NAME);
+    validate!("echo 1 & && echo 2", BOOL_AFTER_BACKGROUND_ERROR_MSG);
+    validate!(
+        "echo 1 && echo 2 & && echo 3",
+        BOOL_AFTER_BACKGROUND_ERROR_MSG
+    );
 }
```

src/parse_util.rs
```diff
@@ -1402,6 +1413,39 @@ fn detect_errors_in_backgrounded_job(
...
+
+    // Disallow background immediately before continuation conjunctions. For example:
+    // foo & && bar
+    // foo & || baz
+    if let Some(list) = job
+        .parent()
+        .unwrap()
+        .as_job_conjunction_continuation()
+        .and_then(|j| j.parent())
+        .and_then(|j| j.as_job_conjunction_continuation_list())
```
krobelus, line 1425: I feel like this should probably not be done in `detect_errors_in_backgrounded_job()` but in the caller where we already check the parent type.
Else we might report the same error twice, no?

src/parse_util.rs
lavafroth: Sure

src/parse_util.rs
lavafroth: Yep, just noticed the error. It should be fine now.

src/parse_util.rs
lavafroth: You're right, this was probably a by-product of debugging attempts!

lavafroth:
Thanks!

Issue #328, opened 2012-09-24
Not loading descriptions for functions in $fish_function_path


anbotero:
At first I thought it was just the functions I had in `~/.config/fish/functions`, but I tried auto-completion suggestions by typing:

``` shell
  va<Tab>
```

And noticed that there wasn’t a description for `vared` either, even though there was one when I checked `functions vared`.

The miscellaneous functions I load separately with `.` load their description just fine.

**PS:** I’m on OS X 10.8.2 (Mountain Lion), using fish OpenBeta_r2 version.

frederickjh:
I would also like to add that even if you use the temporary work-around in fish-shell/fish-shell#1915 posted by **VosaXalo**, creating a function with the same name as a command produces some interesting tab completion results.

Example: I have [hub](https://github.com/github/hub) installed (the Github wrappper for git). In **~/.config/fish/functions/git.fish**  I have:
```
function git --description 'Alias for Github wrapper for git called hub, that adds some Github sugar to git.'
	hub $argv
end
```
Entering `g` **Tab** gives this:
![g_tab_result](https://cloud.githubusercontent.com/assets/864159/14062775/bfab732c-f3ab-11e5-9f76-a06294a0caf4.png)
This is the description from the function.

Entering `gi` **Tab** gives this:
![gi_tab_result](https://cloud.githubusercontent.com/assets/864159/14062776/c53c249e-f3ab-11e5-870f-cf0ed299710e.png)
This is the description from the git man page.

Entering `git` **Tab** gives this:
![git_tab_result](https://cloud.githubusercontent.com/assets/864159/14062778/cac89154-f3ab-11e5-8922-d2944d4235f0.png)
Just an executable without a description this time.

I am not sure what is going on here but something does not seem correct. I think that it should get the same description each time. I also think that  the description of a function (an alias) with the same name as a command should be chosen over the description of the command itself.

floam:
Could possibly be worked around by the manpage completions? Or use `whatis` for all commands unless overwritten?

herrbischoff:
> This is a longstanding bug present in the original fish. Since the completions infer the function name from the file, and doesn't actually load the function, so it can't produce the description.

I was wondering about this lately. Everywhere I looked I found mentions of `--description 'whatever'` but absolutely no mention of how to actually put it to use. So, to conclude: for the time being (and considering the age of this issue, probably for a long time), the `--description` flag is basically just an awkwardly place comment and therefore useless, correct?

[...]

PR #10229, opened 2024-01-17
Add warning about setting fish as login in the readme

ElectreAAS:
In 5af1e64, a warning was added to the index of the documentation about setting up fish as your login shell. This PR also puts this warning in the Readme.

If the warning is no longer necessary and the problems mentionned in #328, #5394, and [on the debian forums](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1001584) have a workaround, do tell me, as I think this is the source of [a very weird situation in opam](https://github.com/ocaml/opam/issues/5761#issuecomment-1893943441).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -115,25 +115,25 @@ The following optional features also have specific requirements:
 Switching to fish
 ~~~~~~~~~~~~~~~~~
 
-If you wish to use fish as your default shell, use the following command, from inside fish:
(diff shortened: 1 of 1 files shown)


faho:
To be honest I've been thinking of removing that entire section from the README and send people to the docs for more instead.

faho:
Alright, I've removed the section, which should solve this issue.

PR #10237, opened 2024-01-19
Move more into build.rs

faho:
## Description

This moves a bunch of logic from cmake into build.rs.

It entirely removes config.h because we no longer use any of it.

The end result is that now, on my system, `cargo build` from a clean directory without any cmake results, builds a "working" fish. It can start and do things, but it won't enter our special "run from build directory" mode, so it won't load the config files.

It is possible that it won't build that way on other systems because it assumes that curses is called "ncurses" and found in whatever default search path it uses. Anything more complicated will still have to use cmake for the time being.

Next steps:

- Figure out how to find out we're in the build directory for cargo (not something cargo wants to expose apparently)
- Move the RLIMIT_ vars to rsconf like the other cmakedefines?
- Remove more and more from cmake
- Spicy: Make the default paths ~/.local/{bin,share} and allow easy installing in the user's home directory

I would mostly like a check that

1. what I did makes sense
2. I didn't miss any use of any of the variables I removed. I checked e.g. WCHAR_T_BITS, and the only thing we use it for is finding the right pcre, so I kept it in cmake


zanchey:
WCHAR_T_BITS doesn't get used for PCRE any more, because the Rust crate handles that bit.

PR #10238, opened 2024-01-20
Remove share/completions/highlight.fish

p-ouellette:
Highlight ships its own completion script:
https://gitlab.com/saalen/highlight/-/blob/master/sh-completion/highlight.fish

I don't know if this requires a changelog entry.


faho:
Thanks!

PR #10241, opened 2024-01-20
Use nix read/write

PolyMeilex:
## Description

Moving from libc::read/write to nix read/write with Result types instead of -1 whenever possible.

---
Somewhat related to work in #10196


ridiculousfish:
Squash-merged as f3e8272c5dab769b0e1afdea120a6641f9e74685. Thank you!

PR #10244, opened 2024-01-21
Dynamic Completion for Justfile Recipes

danielb2:
Dynamic completion for justfile recipes

## Description

This pull request adds dynamic completion for `justfile` recipes. This enhancement is designed to streamline workflows by providing quick access to `justfile` recipes, enhancing user experience similar to the existing completions for `make`.

### DIFF
--- a/share/completions/just.fish
+++ b/share/completions/just.fish
@@ -1,4 +1,4 @@
 function _justfile_targets
-    just --summary 2>/dev/null | string split ' ' | string match -v '.*/.*'
+    just --summary 2>/dev/null | string split ' '
 end
 complete -c just -a '(_justfile_targets)' -f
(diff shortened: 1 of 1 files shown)


share/completions/just.fish
```diff
@@ -0,0 +1,4 @@
+function _justfile_targets
+    just --summary 2>/dev/null | string split ' ' | string match -v '.*/.*'
```
faho, line 2: That last `string match` is in glob-mode, so you'll filter out any lines that have a period, then something, a "/." and then something.

Is that what you want or do you need to add `-r`?

danielb2: I removed the match altogether since it wasn't doing anything. just a rememnant

faho:
Merged, thanks!

PR #10246, opened 2024-01-21
Enable LTO

Berrysoft:
## Description

This PR enables LTO in Cargo.toml. It could shrink the binary size significantly and improve performance (in theory).

The original C++ version binary size is ~1.6MB (stripped). Before LTO, the Rust version is ~4.2MB (stripped). After enabling LTO, the Rust version is ~3.7MB (stripped). If you could accept `opt-level="z"`, the size will be ~2.9MB (stripped).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,6 +9,8 @@ edition = "2021"
 
 [profile.release]
 overflow-checks = true
+lto = true
+codegen-units = 1
 
 [package]
 name = "fish"


faho:
>If you could accept opt-level="z", the size will be ~2.9MB (stripped).

opt-level "z" turns off loop-vectorization, I do not believe that's wise.

Some benchmarking shows a ~50% performance loss for doing that in the [aliases benchmark](https://github.com/fish-shell/fish-shell/blob/master/benchmarks/benchmarks/aliases.fish) and 80% on [math](https://github.com/fish-shell/fish-shell/blob/master/benchmarks/benchmarks/math.fish) for doing that.

I do not see binary size as a huge concern, given that we are talking about under a megabyte of difference.

-----

Thank you, we'll accept LTO.

Issue #10248, opened 2024-01-22
`command` doesn't work properly when ran from a variable


SArpnt:
using fish version 3.7.0 on arch linux with the zen kernel in foot terminal

same behavior happens without customizations

most commands you can run by putting the arguments in a variable and running the variable, for example:
```
set -gx EDITOR kak -c global
$EDITOR
```

in my case because of muscle memory i want to alias `kak` to $EDITOR, but that causes a loop, so i used `command kak -c global` instead. what i found is that command seems to somehow handle the arguments improperly when ran from a variable.

```
> command echo test
test
> set -l var command echo test
> $var

command - run a program

command [OPTIONS] [COMMANDNAME [ARG ...]]
```
(the documentation continues)

```
> command kak -c global
```
(runs kakoune as normal)
```
> set -gx EDITOR command kak -c global
> $EDITOR
command: -c: unknown option

(Type 'help command' for related documentation)
```

```
> set -l com command
> $com kak -c global
command: -c: unknown option

(Type 'help command' for related documentation)
```

faho:
You want either `command $EDITOR` or `set -gx EDITOR (command -s kak) -c global`.

The reason `command foo` doesn't work in a variable is because it's syntactically important, just like e.g. `if` won't work there.

SArpnt:
could this give a better error message?

faho:
`command` is both a keyword (for `command foo`) and a builtin (for `command --search foo` and `command --help`).

What you are running is effectively `builtin command kak -c` because it skips the keyword step.

krobelus:
maybe we can make it a regular builtin, I don't off-hand see an advantage to decorators being keywords except maybe easier implementation. Things like `if` are different because it kind of needs to be known before variable expansion.

faho:
Even if we did that, setting $EDITOR like that would still be a bad idea - it's also used outside of fish, and e.g. `git` sure won't know what a `command kak` is.

PR #10249, opened 2024-01-22
Forbid subcommand keywords in variables-as-commands

faho:
This stops you from doing e.g.

```fish
set pager command less
echo foo | $pager
```

Currently, it would run the command *builtin*, which can only do `--search` and similar, and would most likely end up printing its own help.

That means it very very likely won't work, and the code is misguided - it is trying to defeat function resolution in a way that won't do what the author wants it to.

The alternative would be to make the command *builtin* execute the command, *but*

1. That would require rearchitecting and rewriting a bunch of it and the parser
2. It would be a large footgun, in that `set EDITOR command foo` will only ever work inside fish, but $EDITOR is also used outside.

I don't want to add a feature that we would immediately have to discourage.

## TODOs:

- [ ] This reuses the "subcommand" keywords list, which should be *okay*. Alternatively we can restrict it to just "builtin", "command" and "function", but that would lack "if" and "while", which people could also be trying

### DIFF
--- a/src/parse_execution.rs
+++ b/src/parse_execution.rs
@@ -572,7 +572,12 @@ impl<'a> ParseExecutionContext {
         // This is an attempt to defeat function resolution.
         //
         // Make an exception for "time" because that is frequently used as a command and does fundamentally the same thing.
-        if parser_keywords_is_subcommand(out_cmd) && !unexp_cmd.starts_with(out_cmd.chars()) && out_cmd != L!("time") {
+        // (skipping in no-exec because we don't have the actual variable value)
+        if !no_exec()
+            && parser_keywords_is_subcommand(out_cmd)
+            && !unexp_cmd.starts_with(out_cmd.chars())
+            && out_cmd != L!("time")
+        {
             return report_error!(
                 self,
                 ctx,
--- a/src/parse_execution.rs
+++ b/src/parse_execution.rs
@@ -568,15 +568,14 @@ impl<'a> ParseExecutionContext {
                 "The expanded command was empty."
             );
         }
-        // Complain if we've expanded to a subcommand keyword like "command" or "if".
-        // This is an attempt to defeat function resolution.
+        // Complain if we've expanded to a subcommand keyword like "command" or "if",
+        // because these will call the corresponding *builtin*,
+        // which won't be doing what the user asks for
         //
-        // Make an exception for "time" because that is frequently used as a command and does fundamentally the same thing.
         // (skipping in no-exec because we don't have the actual variable value)
         if !no_exec()
             && parser_keywords_is_subcommand(out_cmd)
-            && !unexp_cmd.starts_with(out_cmd.chars())
-            && out_cmd != L!("time")
+            && &unexp_cmd != out_cmd
         {
             return report_error!(
                 self,
--- a/src/parse_execution.rs
+++ b/src/parse_execution.rs
@@ -573,10 +573,7 @@ impl<'a> ParseExecutionContext {
         // which won't be doing what the user asks for
         //
         // (skipping in no-exec because we don't have the actual variable value)
-        if !no_exec()
-            && parser_keywords_is_subcommand(out_cmd)
-            && &unexp_cmd != out_cmd
-        {
+        if !no_exec() && parser_keywords_is_subcommand(out_cmd) && &unexp_cmd != out_cmd {
             return report_error!(
                 self,
                 ctx,
--- a/src/parse_execution.rs
+++ b/src/parse_execution.rs
@@ -34,6 +34,7 @@ use crate::parse_constants::{
 use crate::parse_tree::{NodeRef, ParsedSourceRef};
 use crate::parse_util::parse_util_unescape_wildcards;
 use crate::parser::{Block, BlockId, BlockType, LoopStatus, Parser, ProfileItem};
+use crate::parser_keywords::parser_keywords_is_subcommand;
 use crate::path::{path_as_implicit_cd, path_try_get_path};
 use crate::pointer::ConstPointer;
 use crate::proc::{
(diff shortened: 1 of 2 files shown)


mqudsi (commented):
The change itself LGTM although I'm not sure how much this is warranted — but I assume you have seen a lot of this and this is in response?

This seems like a fundamental misunderstanding of what variables are and the difference between variable expansion and the use of `eval`. Which leads me to ask

1) Do we *want* to point people to eval in cases like this?
2) How do we feel about filling the codebase with workarounds for people that aren't (yet) understanding the fundamentals of shell architecture and behavior?

faho:
> The behavior this would intercept seems like a fundamental misunderstanding of what variables are and the difference between variable expansion and the use of eval

I'm *fine* with keywords not being seen as such in variable expansion. You wouldn't want to have `set EDITOR "if" foo ";" bar "else" baz "end"`. I also believe allowing `set EDITOR command foo` to run the command `foo` does more harm than good (because now e.g. git won't open the correct editor). That's all okay.

My big problem here is that the error behavior is pretty terrible. We have a `command` *builtin* in addition to the keyword for mostly internal technical reasons, and this leaks the fact, and tries to run `builtin command` which isn't prepared for it and doesn't do anything sensible.

To clarify, you do `set EDITOR command vim` and run `$EDITOR` and you get the entirety of the `command` man page.

And the best place I can find to improve that is directly in the variable-expansion code, because if you changed it in the command builtin it would have to answer how you even got there and explain how not to get there. And we would have to do that in the `if` builtin and the `while` builtin and a bunch of others.

> understanding the fundamentals of shell architecture and behavior?

I disagree that this specific case is all that fundamental. It's a weird and subtle edge case interaction between a number of different parts, exacerbated by our builtin/keyword duality.

And yes, we *have* a bunch of tweaks to prevent common errors by new people - try `if $status` or `function test`.

krobelus:
This looks fine but I think we should consider removing the decorators from the ast and and make them proper builtins.
I think the existing state is historical. Unless we care about the highlighting.
I don't see a reason why `set EDITOR command vim; $EDITOR` shouldn't work.
I don't think it's a footgun because as soon as you run that outside the shell you get the "command not found" error.
The problem in https://github.com/fish-shell/fish-shell/issues/10248 was that we didn't give a good error message in fish but that will go away once we support this.

faho:
> I don't think it's a footgun because as soon as you run that outside the shell you get the "command not found" error.

That is *why* it's a footgun. You set it, it works for some things, and then *later* you find out that `git commit` suddenly fails.

Basically: We would enable something that we would then always recommend people don't use.

Think about it in terms of teaching - if we enabled it and you saw someone doing `set EDITOR command vim` you would have to tell them "yeah that won't work with external commands, use `set EDITOR (command -s vim)` and it will always work". Why not just avoid that entirely?

krobelus:
`EDITOR` is a bad example; what I wanted to say is it's perfectly valid for variables that are used inside the shell.
I don't have much time to delve into this now so I may be wrong but I believe the language would be better off if the commands were normal builtins.
They don't have any special syntactic requirements so I think they should work like other builtins (`set x eval echo 123; $x` etc.).
We wouldn't prohibit that either

faho:
> EDITOR is a bad example; what I wanted to say is it's perfectly valid for variables that are used inside the shell.

EDITOR is *the typical example* for variables-as-commands. $EDITOR, $PAGER and a few like this are widely set and used.

> I don't have much time to delve into this now so I may be wrong but I believe the language would be better off if the commands were normal builtins.

I don't really mind if internally these things have the keyword/builtin duality or not.

I can tell you that having them as decorators makes some parts cleaner - like asking "what's the command for this commandline".

Note that this PR also touches on `if` and `while` etc, which would otherwise just exit. Which makes it seem to me like the builtins are a liability more than the keywords are for them. That's what makes keyword highlighting awkward (`if` is highlighted as a command until you arrive at `if x`, when it turns into a keyword).

> They don't have any special syntactic requirements so I think they should work like other builtins (set x eval echo 123; $x etc.).

The difference there is that "set" is a thing on its own. It *does* something, it sets variables.

By contrast `command` doesn't perform a task, it changes how the *next* word is interpreted. The important thing about `command foo` is the "foo". That's why it's kinda natural as a decorator.

And tbh I question how often anyone would stumble over `set EDITOR set foo bar`. There's an attractive nuisance use for `command` in a variable, I don't think there really is for set or e.g. argparse.

src/parse_execution.rs
```diff
@@ -567,6 +568,19 @@ impl<'a> ParseExecutionContext {
                 "The expanded command was empty."
             );
         }
+        // Complain if we've expanded to a subcommand keyword like "command" or "if".
+        // This is an attempt to defeat function resolution.
+        //
+        // Make an exception for "time" because that is frequently used as a command and does fundamentally the same thing.
+        if parser_keywords_is_subcommand(out_cmd) && !unexp_cmd.starts_with(out_cmd.chars()) && out_cmd != L!("time") {
```
krobelus: I'm confused by `!unexp_cmd.starts_with(out_cmd.chars())`. Given

    set cmd command 1 2 3
    $cmd

`out_cmd` will be `command` (the args are put in a separate variable).
So we should be able to use `unexp_cmd != out_cmd`, right?
A test case would be `command""` - it expands to `command` and we probably want to report this error.
Unless this is only about variable expansion?

krobelus: in addition to `time` we could allow `not` which ships with LLVM though I think we're fine either way.
LGTM other than my above comment.

krobelus (approved):

src/parse_execution.rs
faho: This was an attempt at a very cheesy check for "there was some expansion".

I.e. we do allow a literal `command foo`, we don't allow `set -l var command; $var foo`.

I'll have to check if unexp_cmd is actually *just* the command, I think I might have been confused there.

>A test case would be command"" - it expands to command and we probably want to report this error.
Unless this is only about variable expansion?

`command"" echo foo` currently uses `command` as a decorator and works. I don't see a problem with keeping that, the problem is really *just* with things that expand and are understood to be the command *builtin*.

faho: I've been thinking of removing the `time` exception, actually, because it doesn't work:

```fish
set -l time time
$time sleep 5
```

Gives you the `time` man page instead of running the `time` command.

krobelus: > I'll have to check if unexp_cmd is actually just the command

yeah it is, it's `let unexp_cmd = self.node_source(&statement.command);`  and `statement.command` is a singular string token in command position

krobelus: since it never worked that sounds fine, we can make it work later if anyone ever cares

PR #10251, opened 2024-01-23
Retry open_cloexec for signals other than SIGINT

### LINKED ISSUE #10250: Interrupted system calls not handled on redirection
Fish version 3.7.0 (arch package), no customizations
kernel 6.7.0-zen3-1-zen (arch package)
konsole terminal (xterm-256color) or tmux

```
~> mkfifo /tmp/test
~> cat /dev/urandom > /tmp/test

*** resize terminal window here, or split the tmux window ***

warning: An error occurred while redirecting file '/tmp/test'
open: Interrupted system call
~ [1]>
```

bash/zsh handle this without problem.
I think relevant discussion on bash fix for this issue was here: https://unix.stackexchange.com/a/487260

faho:
This retries {w,}open_cloexec if it gets EINTR, unless it got a cancel signal (i.e. SIGINT).

That means if you do `cat >fifo` (where "fifo" is a fifo without a reader), you can only stop it if you ctrl-c, not if you e.g. resize the terminal.

It also no longer prints the warning when you press ctrl-c - SIGINT is expected, we don't print anything in a lot of cases.

This affects *all* uses of open_cloexec, e.g. in `cd`. Most probably won't see a difference, the man page for open(2) on my system claims:

> While blocked waiting to complete an open of a slow device (e.g., a FIFO; see fifo(7))

Fixes #10250

## TODOs:

### DIFF
--- a/src/fds.rs
+++ b/src/fds.rs
@@ -1,9 +1,11 @@
 use crate::common::wcs2zstring;
 use crate::flog::FLOG;
+use crate::signal::signal_check_cancel;
 #[cfg(test)]
 use crate::tests::prelude::*;
 use crate::wchar::prelude::*;
 use crate::wutil::perror;
+use errno::{errno, set_errno};
 use libc::{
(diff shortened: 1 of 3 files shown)


mqudsi (approved):
LGTM. Lots of places in our code that can be improved to be more resilient to EINTR.

PR #10253, opened 2024-01-23
Automatic vi normal mode on paging

unixpariah:
I added a function that makes fish automatically enter vi normal mode when opening pager.
To enable simply add this to your fish.config:

```shell
fish_vi_on_paging
```


faho:
This seems like fine customization for yourself, but I don't believe we need to ship it with fish.

If you want to share it with others, why not make your own repository where people can download it?

PR #10254, opened 2024-01-23
Use Result<OwnedFd> for open_cloexec

PolyMeilex:
Rework of error handling for `w/open_cloexec` function. 

- Return `Result<T>` from `open_cloexec` instead of `RawFd(-1)`
- Return Rust's `OwnedFd` instead of `RawFd`
- Now that `open_cloexec` returns `OwnedFd` also remove some use of `AutoCloseFd`
  - `AutoCloseFd` is basically the same as std `OwnedFd`, except `OwnedFd` does not have a hidden invalid state, it is guarantied to always be valid. Perhaps in other parts of the codebase that's needed, but in cases where `open_cloexec` was being used the FDs were supposed to always be valid anyway, so I assume the type change is appropriate.
- Lastly, use nix's `OFlag` and `Mode` bitflags for `open_cloexec` attributes
  - I was not sure if you prefer the octal numeric mode notation or bitflag one, for now I did fully named flags, I can revert to octal representation as needed. (as that does not really matter that much, I'd rather focus on Result/OwnedFd stuff)

### DIFF
--- a/src/autoload.rs
+++ b/src/autoload.rs
@@ -325,7 +325,7 @@ fn test_autoload() {
     use crate::common::{charptr2wcstring, wcs2zstring, write_loop};
     use crate::fds::wopen_cloexec;
     use crate::wutil::sprintf;
-    use libc::{O_CREAT, O_RDWR};
+    use nix::fcntl::OFlag;
 
     macro_rules! run {
         ( $fmt:expr $(, $arg:expr )* $(,)? ) => {
@@ -340,7 +340,7 @@ fn test_autoload() {
 
         let fd = wopen_cloexec(
             path,
-            O_RDWR | O_CREAT,
+            OFlag::O_RDWR | OFlag::O_CREAT,
             Mode::S_IRUSR
                 | Mode::S_IWUSR
                 | Mode::S_IRGRP
--- a/src/autoload.rs
+++ b/src/autoload.rs
@@ -336,7 +336,19 @@ fn test_autoload() {
     }
 
     fn touch_file(path: &wstr) {
-        let fd = wopen_cloexec(path, O_RDWR | O_CREAT, 0o666).unwrap();
+        use nix::sys::stat::Mode;
+
+        let fd = wopen_cloexec(
+            path,
+            O_RDWR | O_CREAT,
+            Mode::S_IRUSR
+                | Mode::S_IWUSR
+                | Mode::S_IRGRP
+                | Mode::S_IWGRP
+                | Mode::S_IROTH
+                | Mode::S_IWOTH,
+        )
+        .unwrap();
         write_loop(&fd, "Hello".as_bytes()).unwrap();
         unsafe { libc::close(fd) };
     }
--- a/src/autoload.rs
+++ b/src/autoload.rs
@@ -336,8 +336,7 @@ fn test_autoload() {
     }
 
     fn touch_file(path: &wstr) {
-        let fd = wopen_cloexec(path, O_RDWR | O_CREAT, 0o666);
-        assert!(fd >= 0);
+        let fd = wopen_cloexec(path, O_RDWR | O_CREAT, 0o666).unwrap();
         write_loop(&fd, "Hello".as_bytes()).unwrap();
         unsafe { libc::close(fd) };
     }
--- a/src/autoload.rs
+++ b/src/autoload.rs
@@ -350,7 +350,6 @@ fn test_autoload() {
         )
         .unwrap();
         write_loop(&fd, "Hello".as_bytes()).unwrap();
-        unsafe { libc::close(fd) };
     }
 
     let mut t1 = "/tmp/fish_test_autoload.XXXXXX\0".as_bytes().to_vec();
--- a/src/autoload.rs
+++ b/src/autoload.rs
@@ -341,12 +341,7 @@ fn test_autoload() {
         let fd = wopen_cloexec(
             path,
             OFlag::O_RDWR | OFlag::O_CREAT,
-            Mode::S_IRUSR
-                | Mode::S_IWUSR
-                | Mode::S_IRGRP
-                | Mode::S_IWGRP
-                | Mode::S_IROTH
-                | Mode::S_IWOTH,
+            Mode::from_bits(0o666).unwrap(),
         )
         .unwrap();
         write_loop(&fd, "Hello".as_bytes()).unwrap();
--- a/src/builtins/source.rs
+++ b/src/builtins/source.rs
@@ -7,8 +7,8 @@ use crate::{
     parser::Block,
     reader::reader_read,
 };
-use libc::{c_int, O_RDONLY, S_IFMT, S_IFREG};
-use nix::sys::stat::Mode;
+use libc::{c_int, S_IFMT, S_IFREG};
+use nix::{fcntl::OFlag, sys::stat::Mode};
 
 use super::prelude::*;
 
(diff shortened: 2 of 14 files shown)


faho:
See also #10251, which would have to be incorporated into this if we decide it's the way to go.

I would wait for feedback on that one, it is possible we know a case where EINTR is meaningful and should be handled outside. (but I can't think of any)

PolyMeilex:
Sure, I will simply rebase on that as needed. (feel free to ping me in case I forget to)

src/history.rs
```diff
@@ -804,34 +809,38 @@ impl HistoryImpl {
...
+            };
+
             // Exclusive lock on the entire file. This is released when we close the file (below). This
             // may fail on (e.g.) lockless NFS. If so, proceed as if it did not fail; the risk is that
             // we may get interleaved history items, which is considered better than no history, or
             // forcing everything through the slow copy-move mode. We try to minimize this possibility
             // by writing with O_APPEND.
             unsafe {
-                Self::maybe_lock_file(fd.fd(), LOCK_EX);
+                Self::maybe_lock_file(fd.as_raw_fd(), LOCK_EX);
```
mqudsi, line 829: Can we change this function to take a reference to an owned fd?

src/io.rs
```diff
@@ -993,7 +1003,12 @@ const FILE_ERROR: &wstr = L!("An error occurred while redirecting file '%ls'");
 const NOCLOB_ERROR: &wstr = L!("The file '%ls' already exists");
 
 /// Base open mode to pass to calls to open.
-const OPEN_MASK: libc::c_int = 0o666;
+const OPEN_MASK: Mode = Mode::S_IRUSR
+    .union(Mode::S_IWUSR)
+    .union(Mode::S_IRGRP)
+    .union(Mode::S_IWGRP)
+    .union(Mode::S_IROTH)
+    .union(Mode::S_IWOTH);
```
mqudsi, line 1011: This would be more readable as `Mode::from_bits(0o666).unwrap()` (but I don't know if that's stabilized or we would need to use a match and a panic in `const`).

src/autoload.rs
```diff
@@ -336,10 +336,20 @@ fn test_autoload() {
...
+
+        let fd = wopen_cloexec(
+            path,
+            OFlag::O_RDWR | OFlag::O_CREAT,
+            Mode::S_IRUSR
+                | Mode::S_IWUSR
+                | Mode::S_IRGRP
+                | Mode::S_IWGRP
+                | Mode::S_IROTH
+                | Mode::S_IWOTH,
```
mqudsi, line 349: Can you use just use `Mode::from_bits(0o666).unwrap()` here? Easier to grok for us greybeards.

mqudsi (commented):
This is great work and an excellent thing to do at this point. I was going to ask you to keep the `libc` to `nix` migrations to a later commit/pr but appreciated that they're part of the same effort to strongly type the whole thing, though I pointed out a few places where we can compromise by using a mix of octal and nix enums.

I haven't had a chance to test this locally but from the review here on GH, it looks good!

src/io.rs
PolyMeilex: We can also do `from_bits_retain` that accepts unknown variants as well, therefore does not require an unwrap.

faho:
Alright, #10251 is merged, have at it.

The gist is that

1. We loop for EINTR unless it's a "cancel_signal" (i.e. SIGINT)
2. If we did get a cancel_signal, we don't print a warning

src/history.rs
PolyMeilex: Indeed `maybe_lock_file` is no longer called with raw_fd anywhere, replaced with generic `AsFd`. Should work with `OwnedFd`, ` BorrowedFd` and `File`

src/io.rs
PolyMeilex: Turns out we are on older version of nix that uses bitflags 1.0 rather than 2.0, so `from_bits_retain` is not available. I used `unasfe from_bits_unchecked` instead (does the same thing as from_bits_retain, only difference is that in 1.0 it used to be considered unsafe, and in 2.0 it no longer is), will remove it once we update.

src/autoload.rs
PolyMeilex: Done

PolyMeilex:
Rebased, and addressed the comments.

src/builtins/source.rs
```diff
@@ -48,19 +51,23 @@ pub fn source(parser: &Parser, streams: &mut IoStreams, args: &mut [&wstr]) -> O
...
+            Ok(fd) => {
+                opened_fd = fd;
+            }
+            Err(_) => {
+                let esc = escape(args[optind]);
+                streams.err.append(wgettext_fmt!(
+                    "%ls: Error encountered while sourcing file '%ls':\n",
+                    cmd,
+                    &esc
+                ));
```
krobelus, line 64: Ok so the difference between `AutoCloseFd` and `OwnedFd` is that `OwnedFd` is always valid?
Should probably have a commit that gets rid of all uses of `AutoCloseFd`.

src/fds.rs
```diff
@@ -222,12 +222,16 @@ pub fn set_cloexec(fd: RawFd, should_set: bool /* = true */) -> c_int {
...
 /// Wide character version of open() that also sets the close-on-exec flag (atomically when
 /// possible).
-pub fn wopen_cloexec(pathname: &wstr, flags: i32, mode: libc::c_int) -> RawFd {
+pub fn wopen_cloexec(
+    pathname: &wstr,
+    flags: OFlag,
+    mode: nix::sys::stat::Mode,
+) -> nix::Result<OwnedFd> {
     open_cloexec(wcs2zstring(pathname).as_c_str(), flags, mode)
 }
```
krobelus, line 232: in general it's better to use checked versions and `unwrap()` to panic on failure.

src/autoload.rs
```diff
@@ -336,10 +336,15 @@ fn test_autoload() {
...
 
     fn touch_file(path: &wstr) {
-        let fd = wopen_cloexec(path, O_RDWR | O_CREAT, 0o666);
-        assert!(fd >= 0);
+        use nix::sys::stat::Mode;
+
+        let fd = wopen_cloexec(
+            path,
+            OFlag::O_RDWR | OFlag::O_CREAT,
+            Mode::from_bits(0o666).unwrap(),
```
krobelus, line 344: this commit should be squashed

src/builtins/source.rs
PolyMeilex: > Should probably have a commit that gets rid of all uses of AutoCloseFd.

Yep, IMO in an ideal world they should all be gone, but I'm just trying to refrain myself from doing unmaintainable unreviewable monster PRs, will go forward in reasonable increments.

src/fds.rs
PolyMeilex: What are you referring to?

krobelus: meh github doesn't really support fine grained code review outside the web UI. I'd prefer to use email.

I was talking about

```diff

 /// Narrow versions of wopen_cloexec.
-pub fn open_cloexec(path: &CStr, flags: i32, mode: libc::c_int) -> RawFd {
+pub fn open_cloexec(path: &CStr, flags: i32, mode: libc::c_int) -> nix::Result<RawFd> {
+    let flags = unsafe { OFlag::from_bits_unchecked(flags) };
+    let mode = unsafe { nix::sys::stat::Mode::from_bits_unchecked(mode as u32) };
```

which are gone with later changes

PR #10260, opened 2024-01-25
Update just.fish to handle descriptions for completions

danielb2:
## Description

This change updates fish completions to also include descriptions for justfile recipes. It has been tested with descriptions for recipes with arguments as well


share/completions/just.fish
```diff
@@ -1,4 +1,4 @@
 function _justfile_targets
-    just --summary 2>/dev/null | string split ' '
+    just -l | tail -n +2 | sed -e 's/^[[:blank:]]*//' -e 's/[[:blank:]]*#/\t/' -e 's/[[:blank:]]*[*+][^[:blank:]]*//'
```
faho, line 2: Have you considered `string trim` and `string replace`?

We're trying to get away from `sed`, it's got a bunch of terrible implementations in use (I believe some don't support `\t` on the replacement-side).

danielb2: yes, this should do it then ;)

faho:
Thanks!

PR #10262, opened 2024-01-25
Add feature flag to turn off `%self`

faho:
This is the last remnant of the old percent expansion.

It has the downsides of it, in that it is annoying to combine with anything:

```fish
echo %self/foo
```

prints "%self/foo", not fish's pid.

We have introduced $fish_pid in 3.0, which is much easier to use - just like a variable, because it is one.

If you need backwards-compatibility for < 3.0, you can use the following shim:

```fish
set -q fish_pid
or set -g fish_pid %self
```

So we introduce a feature-flag called "remove-percent-self" to turn it off.

"%self" will simply not be special, e.g. `echo %self` will print "%self".

## Current uses of %self:

Here's a few notable uses of %self I can still find in the wild:

- Oh-my-fish's [installer](https://github.com/oh-my-fish/oh-my-fish/blob/d427501b2c3003599bc09502e9d9535b5317c677/bin/install#L140) (last commit to oh-my-fish/oh-my-fish was a year ago)
- The [await plugin](https://github.com/oh-my-fish/plugin-await/blob/master/functions/await.fish#L5) (last commit 5 years ago)
- [theme-budspencer](https://github.com/oh-my-fish/theme-budspencer/blob/eb3bf2919014fc33c136667bf3ae7b0c5b195d3c/functions/fish_prompt.fish#L58) (last commit 2 years ago)

There are in total [386 matches on Github](https://github.com/search?q=%25self+language%3Afish+&type=code) I can find. These are all easily, almost mechanically replaced with `$fish_pid`, with optional backwards-compatibility by setting the variable yourself if it doesn't exist (see above).

It is worth pointing out that $fish_pid was introduced in fish 3.0, which is 45% of commits up to 3.7.0 behind. It was released 5 years ago.

### DIFF
--- a/doc_src/language.rst
+++ b/doc_src/language.rst
@@ -1987,6 +1987,7 @@ You can see the current list of features via ``status features``::
     qmark-noglob            on  3.0 ? no longer globs
     regex-easyesc           on  3.1 string replace -r needs fewer \\'s
     ampersand-nobg-in-token on  3.4 & only backgrounds if followed by a separating character
+    remove-percent-self     off 3.8 %self is no longer expanded (use $fish_pid)
 
 Here is what they mean:
 
@@ -1994,6 +1995,7 @@ Here is what they mean:
 - ``qmark-noglob`` was also introduced in fish 3.0 (and made the default in 3.8). It makes ``?`` an ordinary character instead of a single-character glob. Use a ``*`` instead (which will match multiple characters) or find other ways to match files like ``find``.
 - ``regex-easyesc`` was introduced in 3.1. It makes it so the replacement expression in ``string replace -r`` does one fewer round of escaping. Before, to escape a backslash you would have to use ``string replace -ra '([ab])' '\\\\\\\\$1'``. After, just ``'\\\\$1'`` is enough. Check your ``string replace`` calls if you use this anywhere.
 - ``ampersand-nobg-in-token`` was introduced in fish 3.4. It makes it so a ``&`` i no longer interpreted as the backgrounding operator in the middle of a token, so dealing with URLs becomes easier. Either put spaces or a semicolon after the ``&``. This is recommended formatting anyway, and ``fish_indent`` will have done it for you already.
+- ``remove-percent-self`` turns off the special ``%self`` expansion. It was introduced in 3.8. To get fish's pid, you can use the :envvar:`fish_pid` variable.
 
 
 These changes are introduced off by default. They can be enabled on a per session basis::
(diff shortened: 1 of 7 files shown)


faho:
> If the goal is to eventually remove it, couldn't it first be deprecated

This *is* the deprecation.

We're telling you that we're planning to turn it off, and individual users can already try it out.

And that is also how users catch it. In the meantime, the status quo is untouched, everything still works as before.

Warnings just make it more annoying, especially in this case where finding uses is trival - you can actually just run `grep '%self'`.

epidemian:
Ok, that's fair, thanks for your thorough reply.

I was thinking more from the perspective of:

1. I'm a regular fish user. I add some functions that use `%self` to my fish config.
2. Years later, my package manager has a new version of fish. I let it update, as i always do. I don't check the changelog, and even if i did, by now i have completely forgotten about the detail of how my functions are implemented. I still use them every now and then.

If this version starts warning about uses of `%self`, i have a chance of catching them and i also get to learn how to fix them, while things still work. But if it doesn't warn, then later, when the version that disables this old behavior by default is released, my functions will suddenly stop working, without any indication of why.

I don't really know if this is too far fetched. It probably is. But it's the kind of use-case that came immediately to mind :sweat_smile:

tests/checks/features-percent-self1.fish
```diff
@@ -0,0 +1,4 @@
+#RUN: %fish --features=remove-percent-self -C 'set -g fish_indent %fish_indent' %s
```
krobelus, line 1: copy-paste error (extra fish_indent)

doc_src/language.rst
```diff
@@ -1987,13 +1987,15 @@ You can see the current list of features via ``status features``::
     qmark-noglob            on  3.0 ? no longer globs
     regex-easyesc           on  3.1 string replace -r needs fewer \\'s
     ampersand-nobg-in-token on  3.4 & only backgrounds if followed by a separating character
+    remove-percent-self     off 3.8 %self is no longer expanded (use $fish_pid)
```
krobelus, line 1990: sounds fine to me. Actually turning it on is a separate decision, possibly not for many years

faho: It will be a while, yes, but I would imagine shorter than qmark-noglob.

I expect fewer issues than with stderr-nocaret - that one was bloody everywhere, and it was harder to find than this.

%self was never a huge thing, the main reason why there are still uses is that oh-my-fish is dead more than it being important.

PR #10263, opened 2024-01-26
Import FindRust and otherwise remove Corrosion from the build 

zanchey:
Corrosion is very clever, but basically none of its features are required. Grab FindRust from the upstream source and drive Cargo directly.

Some of this could be macroised but I don't think it's worth it.

This simplifies the build process.


CMakeLists.txt
```diff
@@ -3,6 +3,8 @@
 # CMake 3.21 is needed for file(IMPORTED_RUNTIME_ARTIFACTS)
```
faho, line 3: Unless I'm overlooking something, this removes IMPORTED_RUNTIME_ARTIFACTS, so we might be able to reduce the minimum cmake version again?

zanchey: Yep, I was leaving that for another commit.

cmake/Rust.cmake
```diff
@@ -1,52 +1,25 @@
...
+# Just let rustup do its job
+set(Rust_RESOLVE_RUSTUP_TOOLCHAINS Off)
 
-    FetchContent_Declare(
-        Corrosion
-        GIT_REPOSITORY https://github.com/mqudsi/corrosion
-        GIT_TAG fish
-    )
+include(FindRust)
+find_package(Rust REQUIRED)
```
faho, line 6: We could make this check the version:

```cmake
find_package(Rust 1.67 REQUIRED)
```

As I understand with only a single version that's a lower bound, and equal or later versions are accepted.

krobelus (approved):

PR #10266, opened 2024-01-27
Open command script in external editor on Alt+o

krobelus:
Fish functions are great for configuring fish but they don't integrate
seamlessly with the rest of the system. For tasks that can run outside fish,
writing scripts is the natural approach.

To edit my scripts I frequently run

    $EDITOR (which my-script)

Would be great to reduce the amount typing for this common case (the names
of editor and scripts are usually short, so that's a lot of typing spent on
the boring part).

Our Alt+o binding opens the file at the cursor in a pager.  When the cursor
is in command position, it doesn't do anything (unless the command is actually
a valid file path). Let's make it open the resolved file path in an editor.

Also enable this functionality in Alt+e.  This is experimental, I will probably
drop it. The motivation is that when the cursor is on the command token of a
single-process commandline, the existing Alt+e behavior is probably not very
helpful. So maybe we can improve discoverability of this feature by plugging
it in here. Though I'm not sure how many users would even try Alt+e on the
command token.

In future, we should teach this binding to delegate to "funced" upon seeing
a function instead of a script. I didn't do it yet because funced prints
messages, so it will mess with the commandline rendering if used from
a binding.  (The fact that funced encourages overwriting functions that
ship with fish is worrysome. Also I'm not sure why funced doesn't open the
function's source file directly (if not sourced from stdin). Persisting the
function should probably be the default.)

Alternative approach: I think other shells expand "=my-script" to
"/path/to/my-script".  That is certainly an option -- if we do that we'd want
to teach fish to complete command names after "=".  Since I don't remember
scenarios where I care about the full path of a script beyond opening it in
my editor, I didn't look further into this.

### DIFF
--- a/share/functions/__fish_anyeditor.fish
+++ b/share/functions/__fish_anyeditor.fish
@@ -1,0 +1,16 @@
+function __fish_anyeditor --description "Print a editor to use, or an error message"
+    set -l editor
+    if set -q VISUAL
+        echo $VISUAL | read -at editor
+    else if set -q EDITOR
+        echo $EDITOR | read -at editor
+    else
+        echo
+        echo (_ 'External editor requested but $VISUAL or $EDITOR not set.')
+        echo (_ 'Please set VISUAL or EDITOR to your preferred editor.')
+        commandline -f repaint
+        return 1
+    end
+    string join \n $editor
+    return 0
+end
--- a/share/functions/__fish_edit_command_if_at_cursor.fish
+++ b/share/functions/__fish_edit_command_if_at_cursor.fish
@@ -1,0 +1,24 @@
+function __fish_edit_command_if_at_cursor --description 'If cursor is at the command token, edit the command source file'
+    set -l editor $argv
+    set -l tokens (commandline -xpc)
+    set -l command
+    if not set -q tokens[1]
+        set command (commandline -xp)[1]
+    else if test (count $tokens) = 1 && test -z "$(commandline -t)"
+        set command $tokens[1]
+    end
+
+    set -q command[1]
+    or return 1
+    set -l command_path (command -v -- $command)
+    or return 1
+    test -w $command_path
+    or return 1
+    string match -q 'text/*' (file --brief --mime-type -- $command_path)
+    or return 1
+
+    __fish_disable_bracketed_paste
+    $editor $command_path
+    __fish_enable_bracketed_paste
+    return 0
+end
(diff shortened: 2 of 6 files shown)


share/functions/__fish_edit_command_if_at_cursor.fish
```diff
@@ -0,0 +1,24 @@
...
+        set command $tokens[1]
+    end
+
+    set -q command[1]
+    or return 1
+    set -l command_path (command -v -- $command)
+    or return 1
+    test -w $command_path
+    or return 1
+    string match -q 'text/*' (file --brief --mime-type -- $command_path)
```
faho, line 17: `file` would be a new dependency, we should probably error out if it doesn't exist.

An alternative is to look at the first line and see if it's a shebang.

share/functions/__fish_anyeditor.fish
```diff
@@ -0,0 +1,16 @@
+function __fish_anyeditor --description "Print a editor to use, or an error message"
+    set -l editor
+    if set -q VISUAL
+        echo $VISUAL | read -at editor
+    else if set -q EDITOR
+        echo $EDITOR | read -at editor
+    else
+        echo
+        echo (_ 'External editor requested but $VISUAL or $EDITOR not set.')
+        echo (_ 'Please set VISUAL or EDITOR to your preferred editor.')
```
faho, line 10: Any particular reason this doesn't just go to stderr?

faho:
> Alternative approach: I think other shells expand "=my-script" to
"/path/to/my-script". That is certainly an option -- if we do that we'd want
to teach fish to complete command names after "=". Since I don't remember
scenarios where I care about the full path of a script beyond opening it in
my editor, I didn't look further into this.

I have, for a laugh, written an abbreviation that does this. I would be okay with shipping that by default after we hash out how default abbrs would work.

I would be quite against making it core syntax, and it would be an awkward compatibility break to make `=` special in that regard.

share/functions/__fish_edit_command_if_at_cursor.fish
krobelus: TIL `file` is POSIX but the specified version doesn't have any of the options we need.
But it seems everyone uses the same implementation (except for OpenBSD but it's still compatible).  
I'm leaning towards adding it to the dependencies, and print an error if it fails here. (We can still early-return though it makes no functional difference).

share/functions/__fish_anyeditor.fish
krobelus: yeah, stderr is better here

share/functions/__fish_edit_command_if_at_cursor.fish
faho: In general long options aren't POSIX.

I'm *fine* with relying on file, I would just like a proper error here.

PR #10267, opened 2024-01-27
completions: fix wg-quick interface completions

ammgws:
wg outputs space separated list, so need to convert to newline separated for `complete`


faho:
Thanks!

Issue #6475, opened 2020-01-06
Publish AppImage


jrop:
It would be convenient if Fish published an [AppImage](https://appimage.org/) along with its other release artifacts.

My use-case for such a thing: It would be very useful to be able to curl an AppImage on a system that I do not have administrative access over, and start using fish-shell even there.  The workflow looks like:

```
curl -LO <..>/fish.AppImage
chmod +x fish.AppImage
./fish.AppImage
... fish starts ...
```

If there is any interest in this, I can investigate putting together a PR that would build an AppImage.

faho:
Okay, most of this is already in place and I managed to make an appimage in ~10 minutes.

The gist of it is:

- You need a .desktop file and an icon, and the "linuxdeploy" program is quite strict about both. The categories have to be right and everything

- You build fish like normal, but with CMAKE_INSTALL_PREFIX=/usr *explicitly*, then you install with DESTDIR=tempdir

- Then you run `linuxdeploy`, in my case `./linuxdeploy-x86_64.AppImage --appdir tempdir --executable tempdir/usr/bin/fish --output appimage`

This pulls in all the libraries etc against which fish has been built and puts it into a squashfs that then forms the appimage. The resulting file is 30-50mb in size (I've seen both, dunno why exactly?) and can be executed immediately.

It seems to pick up the config and such? It's not sandboxed, but $__fish_sysconf_dir and similar are set to something like "/tmp/.mount_fish-3ds1GPr/usr/etc/fish".

Doing `ldd /tmp/.mount_fish-3yvgdVf/usr/bin/fish` reveals that it's copied over libncursesw.so.6 and libpcre2-32.so.0, but not libstdc++ or libc. I'm not entirely sure how wise that is or how well that works.

Overall it seems quite simple to do, but I'm not sure how well it really works and if we want to add yet another item onto the release todo list.

jrop:
@faho Thanks for being so responsive.

Regarding exactly _what_ makes it into the app image, perhaps it would be helpful to reference what NeoVim does.  I have not opened up their AppImage, but it may be telling of what a good strategy is.  Periodically I have found their AppImage useful.

> and if we want to add yet another item onto the release todo list.

Perhaps the following is better suited to another ticket, but is this not something that could be automatically delegated to CI/CD?  Are releases cut manually at the moment?  These are all things I would potentially be willing to create PRs for.

zanchey:
[...]

PR #10269, opened 2024-01-27
Remove (n)curses dependency

faho:
## Description

Note: This PR is an exploration. It's to show that this is possible and to see if it would be workable across a variety of systems.

Here's what this does:

It replaces our (very few) uses of curses the C library with the [rust-terminfo](https://github.com/meh/rust-terminfo) crate, which reads the terminfo database on its own.

That means we no longer link against curses, we don't need to *find* curses, which simplifies the build quite a bit.

It also removes quite a bit of C-interoperability gunk - there is no more "cur_term" pointer that can be null and needs to be released or we get a memory leak. It also removes the pain of "what argument does tputs take" because there no longer *is* a tputs.

Some open questions:

- [ ] The default search path for terminfo does not include /usr/local/share/terminfo, where FreeBSD would place it. I have hacked the tests to set $TERMINFO_DIRS and opened https://github.com/meh/rust-terminfo/pull/40.
- [ ] I have replaced "tputs" with `write(tparm(foo))` - tputs did some weird locking around its output and required a function pointer and it was all a bit messy. Was there an actual reason for that?
- [ ] This requires a "non-hashed" terminfo database, meaning there is apparently a way to build (n)curses so it can't read it. I have never seen systems like that. On FreeBSD you install terminfo-db and it's done
- [ ] Are there any systems that do not have terminfo *at all*? I have not found any so far, FreeBSD 14 defaults to using terminfo (from [the release notes](https://www.freebsd.org/releases/14.0R/relnotes/): "The[ ncurses(3X)](https://man.freebsd.org/cgi/man.cgi?query=ncurses&sektion=3X&format=html) library is now able to use[ terminfo(5)](https://man.freebsd.org/cgi/man.cgi?query=terminfo&sektion=5&format=html) as well as[ termcap(5)](https://man.freebsd.org/cgi/man.cgi?query=termcap&sektion=5&format=html), and uses terminfo preferentially") and FreeBSD 13 also has it available

## TODOs:


faho:
One possibility of making fish work on awkward systems without a working terminfo is to hard-code a fallback Term object, which could even just include the values of xterm-256color we use.

faho:
FreeBSD generally works, and the PR to add the standard directory has been accepted.

NetBSD has its own format, stored in /usr/share/misc/terminfo.cdb, which `file` calls a "NetBSD Constant Database, version 1, for 'NetBSD terminfo'".

You *can* install ncurses via `pkgin install ncurses`, and find the database for that in /usr/pkg/share/terminfo. Running fish with `TERMINFO_DIRS=/usr/pkg/share/terminfo fish` works.

n1000:
I tried this branch out on OpenBSD 7.4 as well, it seemed to work fine there also (at least I did not notice any obvious issue). I had to additionally bump the libc crate version up to 0.2.152 to bring in some missing libc posix_spawn* declarations.

meowcoder:
Can't compile without `libncurses-dev` installed, can't find `term.h`. 
Had to remove the include line from `libc.c` to fix this.

mqudsi:
I am against replacing a package literally available everywhere (curses) with dependencies that might need to be manually installed (even if via the package manager) on various platforms. My FreeBSD installation has been upgraded from version to version since 2009 or so, is heavily abused, contains a ton of non-portable stuff, and yet still doesn't have `terminfo-db` installed.

faho:
> dependencies that might need to be manually installed (even if via the package manager) on various platforms

Keep in mind that most people would either already have it installed or just get them with the fish package.

So it's only those that are on a platform where this isn't already the default *and* decide to install fish outside of their distro package. And for those people, it's an 800K download.

Of course I recognize that it will hit you personally because you'll want to compile fish yourself, but even then you'll just have to install a package. (unfortunately FreeBSD 14's ncurses will *prefer* a terminfo database if it's available, but it isn't a dependency)

In return we get an easier build with some pretty gnarly interoperability code removed. That trade-off sounds attractive to me.

faho:
I'm also absolutely not above just including a few terminfo entries ourselves.

This already includes xterm-256color, which should be the most widely used single entry. If we added ~10 more we could probably get it so most people never actually need their terminfo database. E.g. my fresh FreeBSD VM apparently defaults to TERM=xterm.

And it's not like these things change constantly - and if they do, you'll get terminals like alacritty shipping their own, which this also just picks up.

mqudsi:
For what it's worth, I've read more about it and `terminfo-db` was introduced in FreeBSD 13 (con: if we depend on it, we will be permanently icing out FreeBSD 12 and below users, [which just hit EOL](https://endoflife.date/freebsd)) and currently is built from the ncurses sources (it's a build dependency). But the plan is to invert the dependency tree in FreeBSD 14 so that `ncurses` depends on `terminfo-db`, which means it'll basically be installed everywhere by default in that case.

In light of this, I take back a good part of my resistance to this change, though I continue to have concerns about the availability of terminfo on non-mainstream platforms (where curses is almost always available).

faho:
> We could use terminfo and fall back to termcap.

I am not opposed to adding termcap, but only if that is ever used.

I have looked, and so far I have not found a single platform that isn't EOL that doesn't ship a readable terminfo database as a package but has a termcap database.

- NetBSD has a terminfo database by default, but in a separate format. ncurses terminfo is available as a package. 
  It does not have a separate termcap database (see https://man.netbsd.org/termcap.5, "the termcap database file is no longer shipped with NetBSD" - the page is from 2012)
- FreeBSD 14 will include ncurses terminfo, it's available as a package in 13+.
- OpenBSD apparently ships terminfo according to https://github.com/fish-shell/fish-shell/pull/10269#issuecomment-1913463415
- Any linux distro I know has ncurses terminfo.
- macOS has terminfo
- Haiku [has an ncurses package](https://depot.haiku-os.org/#!/pkg/ncurses6/haikuports/haikuports_x86_64/6/3/-/-/2/x86_64?bcguid=bc224-IOCL)
- Illumos [appears to have terminfo](https://www.illumos.org/issues/10360)

Can you name a system that has termcap but doesn't have terminfo? What about one where you can't easily get terminfo as a package?

Because it seems to me like adding support for [NetBSD's terminfo format](https://man.netbsd.org/NetBSD-6.0/cdb.5) would help more than adding termcap.

> in the absence of terminfo, forward 2-char capabilities to termcap and return None for the rest

This part is easy, we already only use termcap names.

mqudsi:
I'm assuming where you say "terminfo" or "ncurses" in your last comment, you specifically mean a terminfo directory and not just terminfo via ncurses? 

I am surprised that the list is that comprehensive; FreeBSD *usually* isn't the last to bring in newer tech. But I just checked and even my dated Solaris 11 and OpenBSD 6.8 virtual machines have terminfo databases (at `/usr/share/lib/terminfo` and `/usr/share/terminfo`, respectively).

After thinking about it some, another option to would be to ship with our own "fallback terminfo db" containing just an entry for `xterm` and `xterm-256` and we would be ok even on platforms without terminfo support at all?

faho:
> I'm assuming where you say "terminfo" or "ncurses" in your last comment, you specifically mean a terminfo directory and not just terminfo via ncurses?

Yes. There is a "hashed database" format that ncurses introduced pretty late, which isn't used much from what I can see. I believe FreeBSD used to, but they have terminfo-db now.

OpenBSD explicitly dropped the hashed format in 2015.

rust-terminfo cannot "currently" read the hashed format.

(tbh I find this a pretty pointless optimization nowadays - reminds me of how fortune ships those ".dat" files to point to the actual fortunes, when it's 2MB of text and you can just read it directly)

> After thinking about it some, another option to would be to ship with our own "fallback terminfo db" containing just an entry for xterm and xterm-256 and we would be ok even on platforms without terminfo support at all?

https://github.com/fish-shell/fish-shell/pull/10269/commits/c845eb3f3c5be3d9ed1c2bf15765dadb45e0f90c adds xterm-256color, and uses it if all else fails.

We would probably want to use that if $TERM is xterm-256color and reading it from disk failed (and probably without an error message at all). If we want to make it semi-comprehensive we could add "xterm", "screen" and "tmux" plus their -256color variants.

faho:
Alright I added a test that compares the hardcoded xterm-256color I added (from my ncurses install) to the system's, and we see two differences:

1. macOS' is outdated. We know this, that's why we [hack around it](https://github.com/fish-shell/fish-shell/blob/263197dcb76fcc54a59a0838f48b9624332f25f9/src/env_dispatch.rs#L531-L559) to add italics and such.
2. FreeBSD's entry has backspace and delete swapped.

That latter one is annoying, but probably not fixable - it's a decades old confusion that terminals make configurable. So if we do want to add a fallback definition, we would have to live with this. (I don't believe `#cfg`ing it out really helps)

mqudsi:
Can you expand on this:

> FreeBSD's entry has backspace and delete swapped.

How does that contrast to the current behavior when retrieving the value from curses under FreeBSD?

faho:
> How does that contrast to the current behavior when retrieving the value from curses under FreeBSD?

Sorry, it's not delete-the-key, it's the ascii "DEL" character.  It's the backspace key and ctrl-backspace that are swapped, because one of them sends DEL and the other the "backspace" character. It only affects if `bind -k backspace` triggers for backspace or ctrl-backspace.

The good news is that we bind both the same in our bindings *because* this is a common confusion between terminals, so it's not a huge deal. It is also configurable in a lot of them (because of course it is), so you can't actually rely on terminfo to begin with.

----
<details><summary>The useless details</summary>

My archlinux system says "kbs=^?", i.e. ctrl-? (aka `\x7f`, aka "DEL")

The FreeBSD terminfo says "kbs=^H", i.e. ctrl-H

When I run unconfigured and unmodified actual xterm on my arch machine and press backspace I get ctrl-H, when I press ctrl-backspace I get ctrl-?.

That means fish_key_reader (reading from curses) locally on my linux machine says pressing actual backspace in xterm is `\b` and pressing ctrl-backspace is `-k backspace` and `\x7f`. Connecting to FreeBSD via ssh says backspace is `-k backspace` and `\b` and ctrl-backspace is `\x7f`.

In addition, gnome terminal, konsole and wezterm, all claiming to be xterm-256color (and I would guess much more popular than crusty old actual xterm), agree with arch's definition by default (as do the alacritty, kitty and tmux terminfos), so I get `-k backspace` for backspace and `\b` for ctrl-backspace locally.
</details>

faho:
Alright, I believe this is pretty successful.

It gets us around linking to (n)curses, which was always one of the more awkward parts of our build. Which also means it fixes #10316, and makes building any form of self-contained package like appimages easier (because including ncurses there was a crapshoot), see https://github.com/fish-shell/fish-shell/issues/6475#issuecomment-588296175. And it does that by making the code simpler (no more cur_term!) and making a nonexistent curses less of a problem.

The downside is that the BSD packages will have to depend on their terminfo database, because some of them don't install it by default - but even that will mostly work because xterm-256color is the single most common $TERM value.

So I would like to see if there is any appetite to merge this.

zanchey:
I think it's a good idea. rust-terminfo looks fine (supports home directory databases and `TERMINFO_DIRS`).

src/curses.rs
```diff
@@ -441,33 +366,42 @@ impl Term {
...
+            term.to_string()
+        } else if let Ok(name) = env::var("TERM") {
+            name
         } else {
-            sys::setupterm(core::ptr::null(), fd, &mut err)
-        }
-    };
+            return Err(x);
+        };
+        let mut path = PathBuf::from("/usr/pkg/share/terminfo");
```
krobelus, line 392: if this is a NetBSD-specific workaround then maybe enable it only there?
Also, maybe this should use `env!("PREFIX")` instead of hardcoding `/usr`?
The only place where we hardcode `/usr` is in the `DEFAULT_PATH` but PREFIX comes first there.

faho: > Also, maybe this should use env!("PREFIX") instead of hardcoding /usr?

This path is not dependent on where fish is installed, so using PREFIX would be wrong. I am unsure if pkgsrc can be set to use a different path, but this is the default one.

> if this is a NetBSD-specific workaround then maybe enable it only there?

This is the path used on NetBSD if you install from pkgsrc (the default package manager there), but pkgsrc itself can also be used on other platforms.

More than that, this is just one more path to search, if we find a valid terminfo entry there we should use it regardless of platform.

krobelus: ok that sounds fine. Would be nice to get rid of this workaround eventually

src/curses.rs
```diff
@@ -479,90 +413,116 @@ where
...
-            // Ignore the result of del_curterm() as the only documented error is that
-            // `cur_term` was already null.
-            let _ = sys::del_curterm(get_curterm());
-            let _ = sys::set_curterm(core::ptr::null());
-        }
-        *term = None;
-    }
+pub fn setup_fallback_term() -> Arc<Term> {
+    let mut global_term = TERM.lock().expect("Mutex poisoned!");
+    // These values extracted from xterm-256color from ncurses 6.4
```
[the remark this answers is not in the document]
faho: Note that we [already](https://github.com/fish-shell/fish-shell/pull/9026) try xterm-256color if $TERM could not be found.

What this changes is:

1. If TERM is xterm-256color and could not be found, we use our fallback xterm-256color, without an error message or trying the other fallbacks
2. If we can't find *any* of of $TERM or our fallback TERM values, we use xterm-256color. Before, we would go on without *any* terminfo, and a lot of things would just *break*.

Both of these would be a good idea even if we kept using curses.

They are, however, more helpful with the rest of this, because it is now a bit more likely that we don't get a terminfo database (because you need to install curses to build against it, and because curses would also read the hashed database or awkward locations for that database).

src/curses.rs
faho: We can, of course, also propose this [upstream](https://github.com/meh/rust-terminfo).

Unlike the FreeBSD path, I didn't because this one is a bit less generic and I wanted to see what looking up custom paths would look like, in case it turned out that we would have to add e.g. build options to add more search paths.

faho: Or propose a method to terminfo that allows passing a list of directories to search.

krobelus: `man terminfo` says that after checking `TERMINFO_DIRS` etc it does:

> •   Finally, ncurses searches these compiled-in locations:
>    •   a list of directories (no default value), and
>    •   the system terminfo directory, /usr/share/terminfo (the compiled-in default).

I guess "(no default value)" means it's platform specific and we have to reimplement that?

faho:
Good news about the hashed database:

- According to [Dickey's writeup of where ncurses is used](https://invisible-island.net/ncurses/ncurses.faq.html#what_platforms), nobody does anymore
- It apparently still depends on berkeley db, which is now AGPL and owned by Oracle, who stopped development in 2020. *Gentoo* has it [permanently disabled](https://gitweb.gentoo.org/repo/gentoo.git/tree/sys-libs/ncurses/ncurses-6.4_p20230401.ebuild#n225)

I do not believe we need to worry about that.

src/curses.rs
faho: I would assume that is referring to the [--with-terminfo-dirs](https://invisible-island.net/ncurses/INSTALL.html#option:with-terminfo-dirs) ncurses build option.

Technically, distributions *could* set that one, I doubt many do.

We already have checked a *bunch* of systems here, and we already have them covered.

faho: Note: Debian [uses --with-terminfo-dirs](https://salsa.debian.org/debian/ncurses/-/blob/b42033ffb63582d99294f1aea351e3dbcfa1e94c/debian/rules#L136), but to add "/etc/terminfo:/lib/terminfo:/usr/share/terminfo", all of which [are already checked by rust-terminfo](https://github.com/meh/rust-terminfo/blob/870327dd79beaecf50db09a15931d5ee1de2a24d/src/database.rs#L180-L182). Likewise [for fedora](https://src.fedoraproject.org/rpms/ncurses/blob/rawhide/f/ncurses.spec#_146). [NixOS](https://github.com/NixOS/nixpkgs/blob/bc79dc30fdf28f42fadea0a62529ee589b8e0824/pkgs/development/libraries/ncurses/default.nix#L50) adds "/run/current-system/sw/share/terminfo" (tho that seems to only compile it in for static binaries?).

----

I am inclined to merge this as-is, with the search path it has, and if anyone needs other paths we either add them or, worst case, add a build option to add more paths.

krobelus:
xterm everywhere sounds worth trying.
One problem with terminal databases is that inside SSH or containers they may not match the client terminal.
Assuming xterm could solve some of these problems. I'm not sure what if anything would get worse.

> I want away from this system. I want it to be replaced with something usable, and I want that to be a hell of a lot simpler.

Maybe `XTGETTCAP` which talks to the terminal directly.

faho:
> Maybe XTGETTCAP which talks to the terminal directly.

I don't like requiring a response. Especially over ssh that's not a great idea.

I want a system where most sequences are just *usable*. For most things we communicate to the terminal, we do not actually need a response, and we don't even need the terminal to *do* anything in particular. It just can't break and e.g. barf half the sequence onto the screen.

For example if we send the cursor shape or OSC-7 PWD we don't actually care if the terminal changes the cursor shape or does anything with the PWD.

And for any kind of query, a sufficiently terrible terminal can just barf the query instead of giving a response. And sufficiently terrible terminals unfortunately exist, so it isn't helpful.

What I want is a mode where the terminal just sets "$GOODTERM=1", and that means we can just use any OSC etc sequences and it will either support or ignore them. Unfortunately this requires terminal authors to talk to each other and come to an agreement and in our current reality I do not see that happening.

--------------

What I would do in practice is to simply downplay terminfo errors and fall back to our included xterm-256color without much fuss.

I would like to check where in practical use that one gets any big errors, and if we don't find anything remove the scary warnings.

PR #10270, opened 2024-01-27
Fix build on NetBSD

faho:
Notably:

1. A typo in `f_flag` vs `f_flags` - this was probably never tested. I don't know if we couldn't use the ST_LOCAL path everywhere?
2. Some pointless name differences  - `st_mtimensec` vs `st_mtime_nsec`
3. The big one: This said that LC_GLOBAL_LOCALE was -1 "everywhere". Well, not on NetBSD.

Does anyone have a better idea for how to solve these? This works, but it's quite ugly, and hard to keep up-to-date given that we have no CI and, to be honest, not a lot of users on NetBSD.

Especially the name difference is one I just find fudging annoying. I had hoped we could get away from pointless stuff like that.

### DIFF
--- a/src/wutil/mod.rs
+++ b/src/wutil/mod.rs
@@ -537,9 +537,15 @@ impl FileId {
         result.change_seconds = buf.st_ctime;
         result.mod_seconds = buf.st_mtime;
         #[allow(clippy::unnecessary_cast)] // platform-dependent
+        #[cfg(not(target_os = "netbsd"))]
         {
             result.change_nanoseconds = buf.st_ctime_nsec as _;
             result.mod_nanoseconds = buf.st_mtime_nsec as _;
+        }
+        #[cfg(target_os = "netbsd")]
+        {
+            result.change_nanoseconds = buf.st_ctimensec as _;
+            result.mod_nanoseconds = buf.st_mtimensec as _;
         }
         result
     }
--- a/src/libc.c
+++ b/src/libc.c
@@ -182,6 +182,8 @@ int C_RLIMIT_NTHR() {
 #endif
 }
 
+#ifdef LC_GLOBAL_LOCALE
 locale_t C_LC_GLOBAL_LOCALE() {
     return LC_GLOBAL_LOCALE;
 }
+#endif
--- a/src/libc.c
+++ b/src/libc.c
@@ -1,3 +1,4 @@
+#include <locale.h>
 #include <paths.h>
 #include <stdbool.h>
 #include <stdint.h>
@@ -179,4 +180,8 @@ int C_RLIMIT_NTHR() {
 #else
     return -1;
 #endif
+}
+
+locale_t C_LC_GLOBAL_LOCALE() {
+    return LC_GLOBAL_LOCALE;
 }
--- a/src/libc.rs
+++ b/src/libc.rs
@@ -102,3 +102,8 @@ extern "C" {
     fn C_RLIMIT_NPTS() -> i32; // ifndef: -1
     fn C_RLIMIT_NTHR() -> i32; // ifndef: -1
 }
+
+#[cfg(target_os = "netbsd")]
+extern "C" {
+    pub fn C_LC_GLOBAL_LOCALE() -> libc::locale_t;
+}
--- a/src/locale.rs
+++ b/src/locale.rs
@@ -60,12 +60,17 @@ unsafe fn lconv_to_locale(lconv: &libc::lconv) -> Locale {
 
 /// Read the numeric locale, or None on any failure.
 #[cfg(localeconv_l)]
+#[allow(non_snake_case)]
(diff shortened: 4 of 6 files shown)


faho:
Remaining test failures on NetBSD:

```
Testing file checks/tmux-prompt.fish ... Failure:

  The CHECK on line 20 wants:
    prompt 0> <changed>

  which failed to match line stdout:11:
    prompt 0> <>

  Context:
    prompt 0> <>
    prompt 0> <> <= does not match CHECK on line 20: prompt 0> <changed>

  when running command:
    ../test/root/bin/fish checks/tmux-prompt.fish
```

and in the cargo tests:

```
error[E0277]: `*mut sem` cannot be sent between threads safely
   --> src/tests/topic_monitor.rs:57:41
    |
57  |           threads.push(std::thread::spawn(move || {
    |  ______________________------------------_^
    | |                      |
    | |                      required by a bound introduced by this call
58  | |             for _ in 0..1 << 11 {
59  | |                 let before = gens.clone();
60  | |                 let _changed = monitor.upgrade().unwrap().check(&gens, true /* wait */);
...   |
65  | |             let _amt = completed.upgrade().unwrap().fetch_add(1, Ordering::Relaxed);
66  | |         }));
    | |_________^ `*mut sem` cannot be sent between threads safely
    |
    = help: within `UnsafeCell<*mut sem>`, the trait `Send` is not implemented for `*mut sem`
note: required because it appears within the type `UnsafeCell<*mut sem>`
   --> /pbulk/work/lang/rust/work/rustc-1.73.0-src/library/core/src/cell.rs:1987:12
    = note: required for `Unique<UnsafeCell<*mut sem>>` to implement `Send`
note: required because it appears within the type `Box<UnsafeCell<*mut sem>>`
   --> /pbulk/work/lang/rust/work/rustc-1.73.0-src/library/alloc/src/boxed.rs:195:12
note: required because it appears within the type `Pin<Box<UnsafeCell<*mut sem>>>`
   --> /pbulk/work/lang/rust/work/rustc-1.73.0-src/library/core/src/pin.rs:410:12
note: required because it appears within the type `binary_semaphore_t`
   --> src/topic_monitor.rs:159:12
    |
159 | pub struct binary_semaphore_t {
    |            ^^^^^^^^^^^^^^^^^^
note: required because it appears within the type `topic_monitor_t`
   --> src/topic_monitor.rs:336:12
    |
336 | pub struct topic_monitor_t {
    |            ^^^^^^^^^^^^^^^
    = note: required for `std::sync::Weak<topic_monitor_t>` to implement `Send`
note: required because it's used within this closure
   --> src/tests/topic_monitor.rs:57:41
    |
57  |         threads.push(std::thread::spawn(move || {
    |                                         ^^^^^^^
note: required by a bound in `std::thread::spawn`
   --> /pbulk/work/lang/rust/work/rustc-1.73.0-src/library/std/src/thread/mod.rs:680:1

For more information about this error, try `rustc --explain E0277`.
error: could not compile `fish` (lib test) due to previous error
```

To be clear this is in the cargo *tests*. The actual binary builds and runs.

Also I *have* seen check-all-fish-files get a stack overflow, but I have no idea how to reproduce it. It happened twice and now I can't seem to trigger it again. Even running the main loop manually without the timestamp checks, which should run *everything* I can't get it to trigger on purpose.

src/wutil/mod.rs
```diff
@@ -537,10 +537,16 @@ impl FileId {
...
         result.mod_seconds = buf.st_mtime;
         #[allow(clippy::unnecessary_cast)] // platform-dependent
+        #[cfg(not(target_os = "netbsd"))]
         {
             result.change_nanoseconds = buf.st_ctime_nsec as _;
             result.mod_nanoseconds = buf.st_mtime_nsec as _;
         }
+        #[cfg(target_os = "netbsd")]
+        {
+            result.change_nanoseconds = buf.st_ctimensec as _;
```
krobelus, line 547: > Especially the name difference is one I just find fudging annoying. I had hoped we could get away from pointless stuff like that.

I think the `libc` crate already does a bunch of platform-dependent renaming of struct fields, so this can be fixed over there eventually.

mqudsi:
> The big one: This said that LC_GLOBAL_LOCALE was -1 "everywhere". Well, not on NetBSD.

I have an unpublished commit to rsconf that resolve macro values and supports cross-compilation. [`rsconf::get_i32_value()`](https://docs.rs/rsconf/latest/rsconf/struct.Target.html#method.get_i32_value) can be used to do that right now, but might break under cross-compilation.

faho:
> If it's literally just a different name but the entirety of the struct is binary-compatible, just casting it to a basic struct stat with whatever names we want would work (dangerously)

I don't believe it would make the code at hand any nicer, and I think I have a dim understanding of the dangers and I don't want them.

---

Alright, it appears we don't have a better answer at the moment, so I'm merging this.

PR #10276, opened 2024-01-30
fix builtin help ignoring redirects

### LINKED ISSUE #10274: builtin commands ignore redirects when printing man pages
fish, version 3.7.0-1382-g54bc19691
OS: Arch Linux, kernel: 6.6.14-1-lts
Terminal: Konsole 23.08.4-2
Same behaviour with/without third-party customisations.

Noticed when debugging a local test failure in `builtinbuiltin.fish`. The test script does this:
https://github.com/fish-shell/fish-shell/blob/54bc19691882e2dafcf094b5c87493b3eed8a852/tests/checks/builtinbuiltin.fish#L13-L15

This works when the man page is missing, but fails when it is present, even though the match string accounts for this case.
The failure is caused by the function that prints the man page:
https://github.com/fish-shell/fish-shell/blob/54bc19691882e2dafcf094b5c87493b3eed8a852/src/builtins/shared.rs#L544-L568

Notice towards the end of the function that if the man page is missing, we simply append the error message to stderr:
```rust
    if res.status.normal_exited() && res.status.exit_code() == 2 {
        streams
            .err
            .append(wgettext_fmt!(BUILTIN_ERR_MISSING_HELP, name_esc, name_esc));
    }
```

All good so far. But taking a look at the code running the man page, we do something entirely different:
```rust
    let mut ios = IoChain::new();
    if !error_message.is_empty() {
        cmd.push_utfstr(&escape(error_message));
        // If it's an error, redirect the output of __fish_print_help to stderr
        ios.push(Arc::new(IoFd::new(STDOUT_FILENO, STDERR_FILENO)));
    }
    let res = parser.eval(&cmd, &ios);
```
Since we're creating a new `IoChain`, any existing redirects held in `streams.io_chain` are ignored, leading to the help text being printed to stdout (or stderr if there is help text) regardless of whether it is being redirected. This causes the test to fail (and as a side note, this behaviour is pretty inconsistent).

I believe this also affects other builtins, for example `read --help &> /dev/null` will still print the man page to stdout.

Humandoodlebug:
## Description

Build on existing `IoChain` rather than starting with a blank slate when printing man pages for builtin commands.
Unsure how to test this in a way that will run in CI, since I think man pages will be unavailable.

## TODOs:

- [ ] Changes to fish usage are reflected in user documentation/manpages. _Probably not required?_

Fixes #10274

### DIFF
--- a/src/builtins/shared.rs
+++ b/src/builtins/shared.rs
@@ -1,7 +1,7 @@
 use super::prelude::*;
 use crate::builtins::*;
 use crate::common::{escape, get_by_sorted_name, str2wcstring, Named};
-use crate::io::{IoChain, IoFd, OutputStream};
+use crate::io::{IoFd, OutputStream};
 use crate::parse_constants::UNKNOWN_BUILTIN_ERR_MSG;
 use crate::parse_util::parse_util_argument_is_help;
 use crate::parser::{Block, BlockType, LoopStatus};
@@ -553,7 +553,7 @@ pub fn builtin_print_help_error(
     }
     let name_esc = escape(cmd);
     let mut cmd = sprintf!("__fish_print_help %ls ", &name_esc);
-    let mut ios = IoChain::new();
+    let mut ios = streams.io_chain.clone();
     if !error_message.is_empty() {
         cmd.push_utfstr(&escape(error_message));
         // If it's an error, redirect the output of __fish_print_help to stderr
--- a/tests/checks/print-help.fish
+++ b/tests/checks/print-help.fish
@@ -1,0 +1,15 @@
+# RUN: %fish %s
+# Test redirecting builtin help with a pipe
+
+set -lx __fish_data_dir (mktemp -d)
+mkdir -p $__fish_data_dir/man/man1
+# Create $__fish_data_dir/man/man1/and.1
+echo '.\" Test manpage for and (not real).
+.TH "AND" "1" "Feb 02, 2024" "3.7" "fish-shell"
+.SH NAME
+and \- conditionally execute a command' >$__fish_data_dir/man/man1/and.1
+
+# help should be redirected to grep instead of appearing on STDOUT
+builtin and --help | grep -q "and - conditionally execute a command"
+echo $status
+#CHECK: 0


mqudsi (commented):
LGTM. 

Let's wait a bit and see if anyone has any objections to this change; perhaps there is a reason for the existing behavior I'm not aware of.

mqudsi (approved):

krobelus:
Yeah this was an unintentional regression in 61486954b (Use a pager to view long outputs of builtin --help, 2019-10-20).
I definitely noticed at the time but couldn't find the fix.

The `This is only used by the source builtin.` comment is not really true anymore, it is used by multiple builtins.

I think we can test this by hand-crafting a small man page, here's tiny start `tests/checks/print-help.fish`

```fish
# RUN: %fish %s

set -lx __fish_data_dir (mktemp -d)
mkdir -p $__fish_data_dir/man/man1
# Create $__fish_data_dir/man/man1/foo.1

__fish_print_help foo | sed s,^--,,
# CHECK: ...
```

though it's not required to test this

Humandoodlebug:
I'll give that a go. On a vaguely-related note, is it expected that some tests would fail when run locally on a system with `mandoc` installed? It seems like many of the tests expect displaying builtin help pages to fail (as they seem to in CI), but on my system they actually display fine, causing other failures similar to this.

faho:
>  On a vaguely-related note, is it expected that some tests would fail when run locally on a system with mandoc installed? It seems like many of the tests expect displaying builtin help pages to fail

The tests run `--help` to see that we get output, and the easiest is to just see it fail.

This runs `__fish_print_help`, which is supposed to just look in $__fish_datadir/man. It's possible it accidentally runs via $MANPATH for mandoc.

Humandoodlebug:
Seems like it _does_ look in `$__fish_datadir/man`, but `$__fish_datadir` ends up being `repos/fish-shell/share`. :thinking:

Humandoodlebug:
Think I figured out why from debug logs - it thinks it's running in the target directory:
```
config: exec_path: "/home/sam/source/repos/fish-shell/build/test/buildroot/usr/local/bin/fish", argv[0]: "../test/root/bin/fish"
config: Running out of target directory, using paths relative to CARGO_MANIFEST_DIR:
 /home/sam/source/repos/fish-shell
config: determine_config_directory_paths() results:
paths.data: /home/sam/source/repos/fish-shell/share
paths.sysconf: /home/sam/source/repos/fish-shell/etc
paths.doc: /home/sam/source/repos/fish-shell/user_doc/html
paths.bin: /home/sam/source/repos/fish-shell/build/test/buildroot/usr/local/bin/fish
config: sourcing /home/sam/source/repos/fish-shell/share/config.fish
config: sourcing /home/sam/source/repos/fish-shell/etc/config.fish
config: not sourcing /tmp/tmp.RhXyefDm0s/xdg_config_home/fish/config.fish (not readable or does not exist)
```

It seems like this is at least partially intentional behaviour: https://github.com/fish-shell/fish-shell/blob/54bc19691882e2dafcf094b5c87493b3eed8a852/src/bin/fish.rs#L160-L175
... so I think I'd better open another issue for it. I would think we should try to use installed paths if fish "seems installed" (as it would in the test virtual root) before falling back to checking CARGO_MANIFEST_DIR.

faho:
> Think I figured out why from debug logs - it thinks it's running in the target directory:

Yep, that's correct.

> I would think we should try to use installed paths if fish "seems installed" (as it would in the test virtual root) before falling back to checking CARGO_MANIFEST_DIR.

That's not the issue here. It's fine and expected that it looks up the repo's files (otherwise it would have to be rebuilt with DATADIR set to the test root), what we can't allow to happen is that `mandoc` opens your installed man pages, because then it changes depending on whether you have an installed fish.

Humandoodlebug:
Oops - indeed I must have built man pages. Sorry for wasting time!

mqudsi:
Merged with thanks!

Issue #10181, opened 2024-01-01
Installing a Rust release


zanchey:
This issue is to discuss how to drive the installation process of the rewritten-in-rust fish.

`cargo install` is not suitable for a complete, working fish install. The following tasks are required:

- Generate documentation
- Install binaries, completions, functions, tools, documentation, translations, and the pkgconfig file
- Support installation using both prefix and DESTDIR options

It may also:
- Generate source tarballs
- Build Mac apps and installers

Some options:
- Do nothing. Teach fish to run from the cargo build directory, publish to crates.rs, and leave everything else up to distributors (I do not like this option)
- Continue to use CMake. All the logic is already there, and Corrosion may not be necessary. The actual CMake parts can be slimmed down significantly.
- Write a small Makefile. Portable and well-understood, but may end up requiring a configure step.
- Autotools strikes back: if we're going to need a configure step beyond what cargo provides, should we just reach back to our old friends automake/autoconf? (I would prefer to stick with CMake)
- Write all the tasks in `cargo xtask` patterns. Major downside is that there's a lot of things that Make/CMake provide for "free" that would need re-implementation in Rust.

https://github.com/axodotdev/cargo-dist is cool, but does not really solve the problem of making a source tarball that can then be built by distrbutions.

Conan-Kudo:
As a package maintainer of fish in Fedora and an ardent user of fish shell, I would prefer that you keep using CMake. It's straightforward and already does what you need, and works well with packaging and shipping fish in distributions.

zanchey:
Regrettably, there will be some work required for distributors; Fedora probably has it easiest as the imperative nature of RPM scripts means that including the macros that handle build dependencies and so on is straightforward. But at a minimum there will need to be additions made to handle the requirements of cargo being online and package builders being offline.

CMake does not support much in the way of Rust by default; Corrosion is helping but requires either vendoring or online access to work properly and the method of getting CMake variables into the Rust build seems fragile to me.

Conan-Kudo:
Fedora has a method of being able to tell Cargo to use a local index of offline crates packaged as RPMs, so that's not an issue.

cdluminate:
I maintain fish for Debian family (including Ubuntu etc). Feel free to at me here on GitHub, or reach out throught email. I'll have to check debcargo and get back to you soon.

faho:
It's probably worth looking at an example of how rust programs are packaged today:

[...]

PR #10279, opened 2024-01-31
Make `cargo install` only mostly instead of entirely broken

faho:
## Description

(title is a bit of an overstatement, I don't want anyone to post this to HN with "fish can now be installed with `cargo install`!!!111eleven!!!")

Currently, if someone downloads the fish source and goes "It's a rust project, I know this!", they'll run `cargo install --path .` and end up with a fish install that is almost entirely broken in non-obvious ways.

As far as I know, it is impossible to disable `cargo install` officially. We could break it on purpose by setting an environment variable in cmake and check it in build.rs, but this would also break `cargo build`.

So this goes a different path:

1. If we can't read share/config.fish (that would be /usr/share/fish/config.fish in a typical package-managed install, or /usr/local/share/fish/config.fish with classical `sudo make install`), it prints an error and refuses to read the rest of the configuration - which would most likely throw more errors. At that point, fish is running in basically emergency mode.
2. It tells you where the data path is and to copy share/ there. If we did decide to support this, this could be a link to a "fish-4.0-share.tar.xz" that you would extract instead.
3. It defaults $PREFIX to ~/.local

The first change is split out because I believe it to be good on its own, regardless of the rest. Without it, with my own configuration, running a not-fully-installed fish would complain about `fish_add_path` not existing.

Together, this means the experience would be sort of like the following:

1. Run `cargo install --path .`
2. Fish ended up in ~/.cargo/bin/fish, run that
3. Fish tells us:
> error: Fish cannot find its asset files in '/home/alfa/.local/share/fish'.
Please copy the share/ directory from the fish repository there and restart fish.
Refusing to read configuration because of this.
4. `cp -r share/* ~/.local/share/fish`
5. Restart fish
6. It works, with some caveats! `help` will just open the online documentation, which is fine. fish_config *almost* works if we can finagle $__fish_bin_dir (or use `status fish-path`, but I know that's broken on OpenBSD). man pages won't be there, so `builtin --help` will tell you to run `help builtin`.

Of course this isn't exactly ideal. It involves annoying manual steps, and especially upgrading would be a pain - you would have to install the assets again, and it's not obvious that you would have to! (we could theoretically compare mtime of config.fish vs the fish binary)

But it's simple enough to add, reasonably obvious to use, and allows people to install fish locally for themselves, which is something we historically struggled with (see our attempts at an appimage).

If cargo ever learned to install assets, we could extend this to build on that.

### DIFF
--- a/src/bin/fish.rs
+++ b/src/bin/fish.rs
@@ -287,7 +287,9 @@ fn read_init(parser: &Parser, paths: &ConfigPaths) {
         let escaped_pathname = escape(&datapath);
         FLOGF!(
             error,
-            "Fish cannot find its asset files in '%ls'. Refusing to read configuration.",
+            "Fish cannot find its asset files in '%ls'.\n\
+             Please copy the share/ directory from the fish repository there and restart fish.\n\
+             Refusing to read configuration because of this.",
             escaped_pathname
         );
         return;
--- a/build.rs
+++ b/build.rs
@@ -208,7 +208,14 @@ fn setup_paths() {
         var
     }
 
-    let prefix = PathBuf::from(env::var("PREFIX").unwrap_or("/usr/local".to_string()));
+    let prefix = if let Ok(pre) = env::var("PREFIX") {
+        PathBuf::from(pre)
+    } else if let Ok(home) = env::var("HOME") {
+        PathBuf::from(home).join(".local/")
+    } else {
+        panic!("Need either $PREFIX or $HOME");
+    };
+
     if prefix.is_relative() {
         panic!("Can't have relative prefix");
     }


src/bin/fish.rs
```diff
@@ -274,11 +275,27 @@ fn source_config_in_directory(parser: &Parser, dir: &wstr) {
...
+    if !source_config_in_directory(parser, &datapath) {
+        // If we cannot read share/config.fish, our internal configuration,
+        // something is wrong.
+        // That also means that our functions won't be found,
+        // and so any config we get would almost certainly be broken.
+        let escaped_pathname = escape(&datapath);
+        FLOGF!(
+            error,
+            "Fish cannot find its asset files in '%ls'.\n\
+             Please copy the share/ directory from the fish repository there and restart fish.\n\
```
Xiretza, line 293: Wording nit: 

```suggestion
             Please copy the contents of the share/ directory from the fish repository there and restart fish.\n\
```

faho: Oh, we're gonna be arguing about the wording *a bunch*. This is very much preliminary, a rough draft.

Personally, I would make it:

```
Please download https://fishshell.com/data/fish-$version.tar.xz and extract it to $datadir
```

But that requires us to

1. decide that this is a way of installation worth supporting
2. commit to uploading tarballs that contain just share/ to fishshell.com

There are a bunch of other ways to go about this, including embedding all our scripts with something like [rust-embed](https://lib.rs/crates/rust-embed), potentially in a second binary (`fish_installer`), and then tell you to extract them.

I would like to come to an agreement on that bigger picture first.

zanchey:
I don't think this is worth doing. It creates issues where the running fish binary, the installation in `~/.local/share` and potentially an installation in `/usr/local` can all be different versions. These will be unpleasant and unpredictable for users and a support nightmare to untangle. It's reasonable to expect users to read the README.

zanchey:
> would like to make it print an error when share/config.fish couldn't be found, and stop reading the config. That's the first commit in this PR.
> 
> What's your opinion on that?

Sensible!

faho:
Alright, I have now added the first commit of this to master: https://github.com/fish-shell/fish-shell/commit/662fde7b71bb994c310a067ecdce993dec4c955d

That means when fish can't find share/config.fish on startup, it prints an error and refuses to read the rest of the configuration. It will still *run*, but tell you there's something wrong with your install.

----

I do still believe that this area is worth exploring. Something akin to [my fish-installer branch](https://github.com/faho/fish-shell/tree/fish-installer) is pretty close to being able to `cargo install fish-shell`, and that seems like a convenient installation method.

Alternatively we can look into appimages again, which should be a lot easier with #10269 - ncurses was a bit of a pain in that, IIRC.

krobelus:
I haven't had time to look into this but writing our own installer code sounds appealing because it would open a door to getting off cmake (other things we'd need to replace are sphinx and ctest drivers, which sounds easier).
If we go this route, we can have both cmake and its replacement in the tree for a while until we are confident in making the switch.

matklad:
As an out there idea, perhaps fish could bundle assets with the binary itself? This can't work for things like manpages, but should work for anything that fish itself needs, like https://github.com/fish-shell/fish-shell/tree/master/share/functions.

This can be feature-gated, such that `cargo install` installs a self-contained version, while distro package fish in a conventional way. 

This could be also useful for things like statically linking a `fish` binary and just `scp`ing it onto whatever server you have misfortune to debug in prod :) 

And even for manpages, if they are bundled, then they can be self-installed by running `fish --install-manpages`.

This is, of course, a terrible idea.

faho:
> As an out there idea, perhaps fish could bundle assets with the binary itself? 

https://github.com/fish-shell/fish-shell/pull/10279#issuecomment-1919518714:

> Adding this generation to the fish binary directly, so you can do fish --generate, bloats it to 11M unstripped, 8.4M stripped from 6.1M/3.8M. That seems too much to bear.

That's why I would make a separate binary. Doubling the size is a lot. The idea with the separate binary exists at https://github.com/faho/fish-shell/tree/fish-installer.

It's generally usable, but still requires a separate step - but if we had these files in the fish binary we would probably still require you to explicitly invoke it.

faho:
Closing this, see https://github.com/fish-shell/fish-shell/issues/10181#issuecomment-1975320006 for what my future ideas are.

PR #10281, opened 2024-02-02
tab completions: avoid completing gpg --use-embedded-filename

dkg:
`gpg`'s `--use-embedded-filename` is a dangerous option that can cause `gpg` to write arbitrary content to arbitrary files.

According to the GnuPG maintainer, this is [not an option recommended for use](https://dev.gnupg.org/T4500).  Fish shouldn't encourage users to supply it.

I've offered https://dev.gnupg.org/T6972 to upstream to make it even more clear that this option is a bad idea.

While removing it, we might as well also remove `--no-use-embedded-filename`, since it is effectively a no-op.

## TODOs:

I don't think any of the templated checklist above are necessary, but i'm happy to supply them if they're desired.


faho:
Thanks!

(tbh I would like it if gnupg removed about half their options)

PR #10290, opened 2024-02-09
Fix scp completion for WSL (with ssh.exe)

RomainGiraud:
## Description

On WSL, it is common to use ssh.exe (from the Windows part) to share the same `.ssh` folder with the host.

The completion script is not able to extract the version number from `ssh.exe`, which has this form:
```
OpenSSH_for_Windows_8.6p1, LibreSSL 3.4.3
```

This PR fixes this issue.

### DIFF
--- a/share/completions/scp.fish
+++ b/share/completions/scp.fish
@@ -45,7 +45,7 @@ complete -c scp -d "Local Path" -n "not string match @ -- (commandline -ct)"
 
 # Remote path
 # Get the list of remote files from the scp target.
-string match -rq 'OpenSSH_(?<major>\d+)\.*' -- (ssh -V 2>&1)
+string match -rq 'OpenSSH_\w*(?<major>\d+)\.*' -- (ssh -V 2>&1)
 if test "$major" -ge 9
     complete -c scp -d "Remote Path" -f -n "commandline -ct | string match -e ':'" -a "
     (__scp_remote_target):( \
--- a/share/completions/scp.fish
+++ b/share/completions/scp.fish
@@ -45,7 +45,7 @@ complete -c scp -d "Local Path" -n "not string match @ -- (commandline -ct)"
 
 # Remote path
 # Get the list of remote files from the scp target.
-string match -rq 'OpenSSH_\w*(?<major>\d+)\.*' -- (ssh -V 2>&1)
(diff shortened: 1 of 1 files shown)


share/completions/scp.fish
```diff
@@ -45,7 +45,7 @@ complete -c scp -d "Local Path" -n "not string match @ -- (commandline -ct)"
 
 # Remote path
 # Get the list of remote files from the scp target.
-string match -rq 'OpenSSH_(?<major>\d+)\.*' -- (ssh -V 2>&1)
+string match -rq 'OpenSSH_\w+(?<major>\d+)\.*' -- (ssh -V 2>&1)
```
Xiretza, line 48: This should probably be `\w*`?

RomainGiraud: Indeed, typo.

RomainGiraud: I also changed for a more explicit regexp as you have recommend.

ammgws:
> On WSL, it is common to use ssh.exe

Off topic but do you alias `ssh` to `/full/path/to/ssh.exe` to do this? Any agent issues?

RomainGiraud:
> Though that raises a point, the rest of the completions use `command ssh` which would use the WSL ssh even if you define a ssh.fish function in ~/.config/fish/functions to override it to the Windows one...?

Indeed, if I define a function alias for ssh, `command ssh` will use the WSL ssh and not the alias. Should the script be adapted to use `ssh` without the `command`?

> Off topic but do you alias `ssh` to `/full/path/to/ssh.exe` to do this? Any agent issues?

I alias `ssh` to `ssh.exe` (without the full path), the agent is the one from the Windows host, more precisely I use [1Password-agent](https://developer.1password.com/docs/ssh/agent/).

mqudsi:
For what it's worth, I'm a very heavy WSL and I use native `ssh` under WSL, without aliasing it to `ssh.exe`. In fact, I'm more likely to use the WSL `ssh` instead of `ssh.exe` even from Windows (via [`RunInBash`](https://github.com/neosmart/RunInBash)).

mqudsi (approved):

RomainGiraud:
Thanks, I will take a look to RunInBash :)

PR #10291, opened 2024-02-10
Use OwnedFd in AutoClosePipes

PolyMeilex:
Static typing for invalid states in `AutoClosePipes`.

While at it, use nix pipe/pipe2 and Results.

Probably best reviewed per commit, as nix/result changes are unrelated.


ridiculousfish:
Merged as b9ba9e57e8bb9516b6ac29f62f6f6a1fb48c64d3 - thank you!

PR #10293, opened 2024-02-10
feat: cobra-cli support

EmilyGraceSeville7cf:
## Description

https://github.com/spf13/cobra-cli/blob/main/README.md

Fixes issue #

## TODOs:

### DIFF
--- a/share/completions/cobra-cli.fish
+++ b/share/completions/cobra-cli.fish
@@ -1,0 +1,1 @@
+source (cobra-cli completion fish | psub)
--- a/share/completions/cobra-cli.fish
+++ b/share/completions/cobra-cli.fish
@@ -1,1 +1,1 @@
-source (cobra-cli completion fish | psub)
+cobra-cli completion fish | source


EmilyGraceSeville7cf:
I expected `eval (cobra-cli completion fish)` to work, but it failed for some reason: no completion appeared after this command was entered in the interactive Fish session. Can somebody explain why it failed to work? `cobra-cli completion fish` writes output to stdout, so I expected `eval` to correctly interpret it's output.

faho:
What you *actually* want is a third thing:

```fish
cobra-cli completion fish | source
```

This is also what our completions for e.g. [starship](https://github.com/fish-shell/fish-shell/blob/master/share/completions/starship.fish) or [deno](https://github.com/fish-shell/fish-shell/blob/master/share/completions/deno.fish) do

Here's why eval didn't work:

```fish
echo "echo foo
echo bar" > script
```

Now run `eval (cat script)`. This ends up running `eval "echo foo" "echo bar"`, only eval doesn't care about separate arguments. It ends up running `echo foo echo bar`, printing "foo echo bar".

The first line of the output of `cobra-cli completion fish` is a comment, so `eval` sees one big comment and does nothing.

faho:
Thanks!

Issue #4754, opened 2018-02-28
running a command should immediately add it to history


nhooyr:
Version: `2.7.1`
OS: macOS
Terminal: Terminal.app

Did not try without any customizations, pretty sure it won't make a difference.

Anyhow, my issue is that if I run `cat /dev/urandom > /dev/null` and then open a second terminal and press up, `cat /dev/urandom > /dev/null` does not show up in my history. I think it should. I get very confused sometimes as I want to run the same command in parallel with different parameters often.

krader1961:
Works for me. After you run the command that you believe isn't in your history do `tail ~/.local/share/fish/fish_history`. Do you not see the command you just ran? I see it in the history whether I run that `tail` command in the same session or a different session. Which means a new session should see it.

nhooyr:
I do not see it in that file after I begin running the command. I only see it when I kill the command with ctrl+c.

krader1961:
Ah, yes! That explains it. The command is only written to the history file after it completes and a new shell prompt is displayed. That is by design. Whether that should be changed is a question for the developers actively maintaining fish. Personally, I think the answer is "no". I've been using fish for several years and have never considered this behavior a problem. It does differ from ksh, for example, which writes the command to the history file before the command completes. But ksh doesn't write anything other than the command line. Whereas the fish command history is much richer and extensible. For example, it could be extended to include how long the command ran.

krader1961:
> ..., it would just need to edit the history entry once the command finishes.

@nhooyr, We're not talking about a database such as Mysql or Postgress. It isn't possible to "edit the history entry". It is possible to append pre and post execution data about a command to the history file. Whether that is worth the cost is debatable.

nhooyr:
Why don't we turn it into a database? Especially since we're considering https://github.com/fish-shell/fish-shell/issues/4154

mqudsi:
This is in many ways #907

faho:
I cannot reproduce the original issue.

I do a command that doesn't end and it is saved and recallable in new sessions, which also matches what I saw in #10302.

PR #10302, opened 2024-02-15
Allow deciding if a command should be saved to history

faho:
## Description

This adds a function called `fish_should_add_history` (EDIT: name changed to `fish_should_add_to_history` after review) that is run when fish decides whether to add a command to history. If it returns 0 the command is saved, if it returns anything else the command is ephemeral (and only available until the next command, like anything space-prefixed now).

(EDIT: This was removed) One notable change needed here is that we now save the history *after* the command was run. This is necessary if we want to allow people to access e.g. $status.

It's visible in some cases, like `history clear` or setting $fish_private_mode - now `set -g fish_private_mode 1` is *not* stored, but `set -eg fish_private_mode` is, because the history setting is already changed by the time fish uses it.

This will also affect commands that end the shell like `exec` or anything that crashes fish. I don't see any way out of that (adding and then removing later doesn't gel with not keeping history), so the alternative would be to not allow access to $status.

Another decision is that this keeps the current setting - space-prefixed commands aren't saved - when no fish_should_add_history exists. It doesn't add a default, and once a function exists it is entirely responsible and would have to check for the space itself. This is in contrast to @ridiculousfish's [comment in 5924](https://github.com/fish-shell/fish-shell/issues/5924#issuecomment-502176484). My rationale is that I want fewer default functions.

A rerun of #9298
Fixes issue #5924

## TODOs:


src/reader.rs
```diff
@@ -4675,14 +4675,31 @@ impl ReaderData {
         }
     }
 
+    fn should_add_history(&mut self, text: &wstr, parser: &Parser) -> bool {
+        if !function::exists(L!("fish_should_add_history"), parser) {
```
krobelus, line 4679: > It gets the right-trimmed (WIP) text as the argument.

that sonds fine because that's what will be added, not the original

(Also how about using a mailing list instead of github, to make code review easier)

src/reader.rs
```diff
@@ -618,6 +618,7 @@
         if !eval_res.no_status {
             STATUS_COUNT.fetch_add(1, Ordering::Relaxed);
         }
+        data.add_to_history(command.clone(), parser);
```
krobelus, line 621: We need to think about whether the behavior change is worth it.
Logging the command before running it seems like a safe default behavior,
which makes a difference in some cases like `exec` you mention.

If I start a long-running command, open a new terminal and press Up-arrow
before the first command is finished, what should the behavior be?
Maybe it doesn't matter maybe it does.

I think we don't have to decide now; we can err on the side of caution and merge this without support for querying
`$status` etc, especially since no one explicitly asked for that.
I think `$status` is not super relevant because most failed commands will be shadowed by their fixed versions.
Users who use a history pager might see things differently.. but only if they configure this behavior, so I wouldn't care too much.
Intuitively I'd tell users who want to use `$status` to remove those command from history in a `postexec` hook?

src/reader.rs
faho: > (Also how about using a mailing list instead of github, to make code review easier)

God no. I don't use email.

src/reader.rs
faho: > We need to think about whether the behavior change is worth it.

Yup.

> $status etc, especially since no one explicitly asked for that.

We have had people ask for it, e.g. #5 is relevant.

[sponge](https://github.com/meaningful-ooo/sponge) features it.

If $status isn't available, there's the question what fish_should_add_history would in practice do. If it's just matching against a list or maybe regex we could also get away with a variable.

Would people check against, say, $PWD?

krobelus:
> Would people check against, say, $PWD?

Yeah that sounds like something that should be possible

faho:
Alright, considering that this was a change (tho I can't find *where* it was changed) and specifically asked for in #4754, I don't believe we want to move history back to after the command.

So checking $status is gone.

If this is needed, we could add a "defer" mode, where you return e.g. 2 and then we run a second function after the command to decide? Tho that sounds pretty complicated. The alternative is to do what people do and delete it on postexec.

krobelus (approved):
LGTM, thanks. How about calling it `fish_should_add_to_history` instead of the ambiguous `fish_should_add_history`?

faho:
Renamed to fish_should_add_to_history and merged.

PR #10303, opened 2024-02-16
Provide man and completion paths when running from build dir

krobelus:
(EDIT: currently broken when using `-GMake` but works with `-GNinja`)

We offer the usual installation method of `make && make install`.  Since are
a leaf package, installation is not inherently necessary.  It would be nice
if instead of installing I could simply symlink `fish` into my `$PATH`.

This would lower the barrier to running (possibly multiple) forks.
I can imagine that users who just want to fix some completion scripts may
not want to clobber their system with a full install.

It already kind of works but there are two issues:

1. Documentation and completions and some scripts are not available
   For example `man path` will fail.

2. vendor completion/function/snippet directories will point to `PREFIX`

This PR is a quick and dirty fix for issue 1. Barely tested.

Issue 2 is tricky, but also sort of unrelated.
It also happens when using `make install`.
Let's say the system package manager installs 3rd party completions to `/usr/share/fish/vendor_completions.d/`.
By default, `make install` will install to `/usr/local/`, so only `/usr/local/share/fish/vendor_completions.d/` is guaranteed to be included.
We do include `$XDG_DATA_DIRS` at runtime as a non-fish-specific way to resolve this.
But it wold be nice if this didn't require manual intervention.

[*] Same for `fish_indent` and `fish_key_reader` but those rarely change,
so it's fine if they come from the system package manager.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -98,6 +98,24 @@ add_definitions(-D_REENTRANT)
 # not run cargo, so we need to tell it *something*
 FILE(GLOB sources src/* src/*/* src/*/*/*)
 
+if(NOT FISH_IN_TREE_BUILD)
+  add_custom_target(data_dir ALL
+    COMMAND mkdir -p ${CMAKE_BINARY_DIR}/share
+    COMMAND # Don't run ln twice or it will create a new link in the link.
+            CreateLink() {
+                if ! test -e \"$$2\" ";" then
+                    ln -sf \"$$1\" \"$$2\" ";"
+                fi
+            }
+    COMMAND CreateLink ${CMAKE_SOURCE_DIR}/share/functions ${CMAKE_BINARY_DIR}/share/functions
+    COMMAND CreateLink ${CMAKE_SOURCE_DIR}/share/completions ${CMAKE_BINARY_DIR}/share/completions
+    COMMAND CreateLink ${CMAKE_SOURCE_DIR}/share/config.fish ${CMAKE_BINARY_DIR}/share/config.fish
+    COMMAND CreateLink ${CMAKE_SOURCE_DIR}/share/groff ${CMAKE_BINARY_DIR}/share/groff
+    COMMAND CreateLink ${CMAKE_SOURCE_DIR}/share/tools ${CMAKE_BINARY_DIR}/share/tools
+    COMMAND CreateLink ${CMAKE_BINARY_DIR}/user_doc/man ${CMAKE_BINARY_DIR}/share/man
+  )
+endif()
+
 # Define a function to link dependencies.
 function(FISH_LINK_DEPS_AND_SIGN target)
   add_custom_command(
@@ -115,11 +133,12 @@ function(FISH_LINK_DEPS_AND_SIGN target)
     USES_TERMINAL
     # Don't use VERBATIM here, to allow the generator expressions above to expand to nothing rather than an empty string
   )
+
   add_custom_target(${target} ALL
     COMMAND "${CMAKE_COMMAND}" -E copy
       "${rust_target_dir}/${rust_profile}/${target}"
       "${CMAKE_CURRENT_BINARY_DIR}"
-    DEPENDS ${rust_target_dir}/${rust_profile}/${target}
+    DEPENDS ${rust_target_dir}/${rust_profile}/${target} data_dir
   )
   codesign_on_mac(${target})
 endfunction(FISH_LINK_DEPS_AND_SIGN)
--- a/cmake/Tests.cmake
+++ b/cmake/Tests.cmake
@@ -37,7 +37,7 @@ add_custom_target(fish_run_tests
           FISH_SOURCE_DIR=${CMAKE_SOURCE_DIR}
           ${CMAKE_CTEST_COMMAND} --force-new-ctest-process # --verbose
           --output-on-failure --progress
-  DEPENDS tests_dir funcs_dir tests_buildroot_target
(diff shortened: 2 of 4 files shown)


faho:
So the idea here is that you would "install" fish by building it and adding a link in $PATH, keeping the scripts in the build directory?

Tbh my preferred idea is to instead install in your home directory, like #10279 does - put the scripts in a place like ~/.local/share/fish. This would also work with `cargo install`.

An alternative is to find the data directories relative to the fish binary, like "../share", so you can make a fish binary in ~/dev/fish-shell/build/fish and then have it resolve the path on start so it looks in ~/dev/fish-shell/share/{functions,completions} and such.

> I can imagine that users who just want to fix some completion scripts may
not want to clobber their system with a full install.

The easy way to update a single completion script is to take it and copy it to ~/.config/fish/completions.

That way you don't even have to build anything.

krobelus:
> So the idea here is that you would "install" fish by building it and adding a link in $PATH, keeping the scripts in the build directory?

Yep. I use it for several programs, it's nice because it allows easy uninstallation.

> 
> Tbh my preferred idea is to instead install in your home
> directory, like #10279 does - put the scripts in a place like
> ~/.local/share/fish.

I think these two approachs are complementary, there's no real conflict, and yours already works?

> This would also work with `cargo install`.

That could be nice in future; though I'm not sure I understand how changing the default prefix helps cargo?

> An alternative is to find the data directories relative to the
> fish binary, like "../share", so you can make a fish binary in
> ~/dev/fish-shell/build/fish and then have it resolve the path on
> start so it looks in ~/dev/fish-shell/share/{functions,completions}
> and such.

The problem is that `build/fish/../share` is `share` and not `build/share` so it is missing the generated files (which can be different per build-directory).

I'd really like using `$(status fish-path)/../share`, making it "relocatable"; `gcc` and a few others do the same.
It might remove the need for `__fish_build_paths` (unless variables like `extra_completionsdir` are set by the user).

So I'd want to add that.
For compatibility, I'd move the binary to `build/fish/bin/fish` and symlink `build/fish` to point there?

faho:
> That could be nice in future; though I'm not sure I understand how changing the default prefix helps cargo?

It's in the home directory, so no root necessary. `cargo install` will put fish at ~/.cargo/bin/fish, and my current approach has a separate binary called "fish_installer" that includes the entirety of share/ and will extract it to ~/.local/share/fish/.

That means you can install it without keeping the build directory around - which would allow you to `cargo install` from crates.io.

Unfortunately you still need that manual step to extract the data files. (technically it would be possible to include them in the main fish binary instead and have it auto-extract them, but that about triples the size of it, and that might have ramifications on startup time etc)

> I'd really like using `$(status fish-path)/../share`, making it "relocatable"; `gcc` and a few others do the same.

Agreed. This would also enable "easy uninstallation" if you can just install fish into a single directory.

> It might remove the need for `__fish_build_paths` (unless variables like `extra_completionsdir` are set by the user).

Probably not. That one's basically 100% added for nix, which has weird needs when it comes to path.

faho:
Fundamentally, I think we need to decide how many "installation" (or not) methods we want to support.

This one is fairly dirty - it adds a bunch of symlinks to any build directory that are exclusively required to enable this, and they don't really *contain* the "installation". If you e.g. checkout a different branch it will track along. That means you can't really have *separate* installations of this because they'll share, well, share/.

So I don't see how it enables "running (possibly multiple) forks", unless you mean "an installed fish from your package manager" and "one fish built from git".

An alternative is to *copy* things in, but that would disallow what I typically do, which is adjust scripts in share/ and run the fish from build/ to test.

So my question is if this is for *developers* to run a fish for development testing, or for *users* to override a completion (in which case copying it over would be easier).

(side note on my system user_doc is in the *build* directory, not the source one, so that link seems useless?)

krobelus:
sorry it's not for overriding completions but for being able to build and build and run fish without installing. Crucially, also being able to "uninstall" without having to use a package manager

faho:
> vendor completion/function/snippet directories will point to PREFIX

This uses `$__fish_data_dir`, which in "running in build directory" mode is changed to point to share/ in CMAKE_SOURCE_DIR (and would be ../share if we went relocatable).

We could add a second variable `$__fish_real_data_dir` that isn't changed, and use that.

zanchey:
I'm not that keen on this change; it seems like a significant increase in complexity for a relatively rare use-case.

krobelus:
I think if we decide to remove support for build directories - in favor of "in-tree" builds, then this will be more attractive.
We already want to be able to run `build/fish` for development; instead of making it only half-supported we can make it fully equal to installation.
I think that will be simpler overall -- we can get rid of `build/test/buildroot` and the `build/fish`/`target/debug/fish` schism; installation logic will be more obvious etc.

So I think we should revisit this when we get rid of build directories.

faho:
> I think if we decide to remove support for build directories - in favor of "in-tree" builds, then this will be more attractive.

I'm not entirely sure how you got that idea?

Out-of-tree builds are *good*, it's one of the major advantages of cmake over the previous autotools system. I don't want to get rid of them.

krobelus:
If we remove out-of-tree builds, `cargo` cache will still work fine so I don't know if they still help.
I guess the `make clean` implementation would be more complicated.

faho:
@krobelus: Are you still interested in this? It appears zanchey and me don't really like it.

krobelus:
I'm not actively working on it, let's see what the other installation mechanisms bring

PR #10308, opened 2024-02-18
Use OwnedFd & Result for tempfile functions

PolyMeilex:


src/env_universal_common.rs
```diff
@@ -494,45 +493,43 @@ impl EnvUniversal {
...
 
-        if !result.is_valid() {
+        if let Err(errno) = result {
             FLOG!(
                 error,
                 wgettext_fmt!(
                     "Unable to open temporary file '%ls': %s",
-                    out_path,
-                    saved_errno.to_string()
+                    tmp_name_template,
```
PolyMeilex, line 516: This is a change in behavior, reading man pages for `mkstemp` /`mkostemp` suggests that after error occurs, the template string can be left in undefined state. So I figure we better not rely on that.

I'm pretty sure it puts the file path in there, but that's not defined as far as I can tell.

src/fallback.rs
```diff
@@ -100,7 +101,7 @@ pub fn fish_wcswidth(s: &wstr) -> isize {
 // Replacement for mkostemp(str, O_CLOEXEC)
 // This uses mkostemp if available,
 // otherwise it uses mkstemp followed by fcntl
-pub fn fish_mkstemp_cloexec(name_template: CString) -> (RawFd, CString) {
+pub fn fish_mkstemp_cloexec(name_template: CString) -> nix::Result<(OwnedFd, CString)> {
```
PolyMeilex, line 104: The reason why I used `Result<(Fd, Path)>` signature, rather than putting the path outside of the result (like the existing usage of this function would suggest) is described in https://github.com/fish-shell/fish-shell/pull/10308/files#r1493507686

src/env_universal_common.rs
```diff
@@ -750,23 +746,19 @@ impl EnvUniversal {
...
-        let private_fd = self.open_temporary_file(directory, &mut private_file_path);
-        let mut success = private_fd.is_valid();
+        let private_fd = self.open_temporary_file(directory);
 
-        if !success {
+        let Ok((private_fd, private_file_path)) = private_fd else {
             FLOG!(uvar_file, "universal log open_temporary_file() failed");
-        }
+            // We can early return because file creation failed, so we don't have to clean it up
+            return false;
```
PolyMeilex, line 755: This is also a change in behavior, previously if `mkstemp` /`mkostemp` failed we called `unlink` on whatever was left in the template buffer, but as mentioned in the comment above, it seems that contents of that buffer after error occurs are not defined.

So here I'm assuming that mkstemp failure does not require cleanup.

mqudsi:
Hey @PolyMeilex, I'm sorry, I missed this PR and actually ended up merging in https://github.com/fish-shell/fish-shell/commit/58a6eb6e45cbbf229d74196ecbe8bb83031c5a4d as part of some work on `history.rs` that overlaps with your proposal.

mqudsi (changes requested):
For more context, my earlier commit was made with the intention of supporting https://github.com/fish-shell/fish-shell/pull/10355 which changes the return value of these function calls from a raw fd to not just an `OwnedFd` but an actual `File` object.

PolyMeilex:
Sure, no problem, I will review if there is anything from this that is still usable shortly

PolyMeilex:
Ok, looks like there is still stuff here that is usable, I will rebase.

Could you take a look at my comment: https://github.com/fish-shell/fish-shell/commit/58a6eb6e45cbbf229d74196ecbe8bb83031c5a4d#r139586934 ? In this PR I made `open_temporary_file` return an error after 10 failed attempts, in your version it retries forever (and logs an error only on attempt number 10), just making sure this is intended.

mqudsi:
> in your version it retries forever (and logs an error only on attempt number 10), just making sure this is intentional.

The code is misleading, the reason I was able to refactor the error out is because the previous code never returned in case of failure. `FLOG(error, ...)` (with `error` *specifically*) is "log-and-abort".

mqudsi:
(I also intentionally didn't close this because I thought you might want to rebase in case there's anything left you still want merged in.)

PolyMeilex:
Ok, this is now rebased, the only thing left is resultifing of write_to_fd function.
I was also written as if `FLOG(error,` was not aborting internally, so I kept that shape of the code for now.

> FLOG(error, ...) (with error specifically) is "log-and-abort".

I can't find any indication of this being the case in the code, but perhaps I missed something.

mqudsi:
> I can't find any indication of this being the case in the code, but perhaps I missed something.

You are correct, I can't find out when this behavior changed. Now I need to fix my earlier commit!

mqudsi (approved):

mqudsi:
Thanks. Sorry this was stuck without review for this long.

PolyMeilex:
Thanks, no problem!
Glad I could help uncover some `FLOG(error)` mysteries.

PR #10317, opened 2024-02-20
Add completion for pstree

artemisart:
## Description

Add completion for pstree <https://www.man7.org/linux/man-pages/man1/pstree.1.html>.

Is it possible to show a custom description for each argument value? Here I want to complete PIDs, it would be nice to show the process name in the tab options/search, but only complete with the PID.

pstree help for reference:

```
Usage: pstree [-acglpsStuZ] [ -h | -H PID ] [ -n | -N type ]
              [ -A | -G | -U ] [ PID | USER ]
       pstree -V
Display a tree of processes.

  -a, --arguments     show command line arguments
  -A, --ascii         use ASCII line drawing characters
  -c, --compact       don't compact identical subtrees
  -h, --highlight-all highlight current process and its ancestors
  -H PID,
  --highlight-pid=PID highlight this process and its ancestors
  -g, --show-pgids    show process group ids; implies -c
  -G, --vt100         use VT100 line drawing characters
  -l, --long          don't truncate long lines
  -n, --numeric-sort  sort output by PID
  -N type,
  --ns-sort=type      sort by namespace type (cgroup, ipc, mnt, net, pid,
                                              user, uts)
  -p, --show-pids     show PIDs; implies -c
  -s, --show-parents  show parents of the selected process
  -S, --ns-changes    show namespace transitions
  -t, --thread-names  show full thread names
  -T, --hide-threads  hide threads, show only processes
  -u, --uid-changes   show uid transitions
  -U, --unicode       use UTF-8 (Unicode) line drawing characters
  -V, --version       display version information
  -Z, --security-context
                      show SELinux security contexts
  PID    start at this PID; default is 1 (init)
  USER   show only trees rooted at processes of this user
```

## TODOs:

### DIFF
--- a/share/completions/pstree.fish
+++ b/share/completions/pstree.fish
@@ -1,0 +1,22 @@
+complete -c pstree -f
+complete -c pstree -s a -l arguments -d 'Show command line arguments'
+complete -c pstree -s A -l ascii -d 'use ASCII line drawing characters'
+complete -c pstree -s c -l compact -d 'don\'t compact identical subtrees'
+complete -c pstree -s h -l highlight-all -d 'highlight current process and its ancestors'
+complete -c pstree -s H -l highlight-pid -d 'highlight this process and its ancestors' -a '(ps -Ao pid | string trim)'
+complete -c pstree -s g -l show-pgids -d 'show process group ids; implies -c'
+complete -c pstree -s G -l vt100 -d 'use VT100 line drawing characters'
+complete -c pstree -s l -l long -d 'don\'t truncate long lines'
+complete -c pstree -s n -l numeric-sort -d 'sort output by PID'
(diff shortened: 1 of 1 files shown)


faho:
> Is it possible to show a custom description for each argument value? Here I want to complete PIDs, it would be nice to show the process name in the tab options/search, but only complete with the PID.

See the [completions for kill](https://github.com/fish-shell/fish-shell/blob/master/share/completions/kill.fish), which call a function that is made for this called [__fish_complete_pids](https://github.com/fish-shell/fish-shell/blob/master/share/functions/__fish_complete_pids.fish).

What you do is you print the completion candidate followed by a tab and its description, one candidate per line. But in this case you should be able to just call that function.

artemisart:
Thanks @faho ! that's perfect

Issue #4959, opened 2018-05-03
String: [STRING…] arguments beginning with dashes are interpreted as options


adamkiss:
Let me begin by saying thank you, to all the maintainters — it is a joy, using fish. So, **thank you**.

```
✨ fish --version; uname -a; echo $TERM
fish, version 2.7.1
Darwin km.local 15.6.0 Darwin Kernel Version 15.6.0: Fri Feb 17 10:21:18 PST 2017; root:xnu-3248.60.11.4.1~1/RELEASE_X86_64 x86_64
xterm-256color
```

**Description**:
When trying to manipulate strings which begin with dashes, `string` mistakenly parses input as options, and fails with either 'Unknown option' (if whatever follows dashes isn't valid option), 'Missing argument' (if string ends) or just uses it as option (?) — I am not completely sure about all the ins and outs of fish yet.

**Command**:
```
✨ string sub -l 2 'aa--'; and string sub -l 2 'a--a'; and string sub -l 2 '--aa'
```

**Expected output**:
```
aa
a-
--
```

**Actual output**:
```
aa
a-
string sub: Unknown option '--aa'

…
```

As far as I can tell, every subcommand od string does it — tested with match, replace, sub and upper.

krader1961:
It's not just the `string` command. Almost every fish builtin, function, and many (most?) external commands behave the same way. If there is any chance a non-option argument might begin with a hyphen you should place a `--` after any flags and before any non-flag arguments. For example,

```
string sub -l 2 -- '--aa'
```

A `--` argument tells the option parser to stop looking for flags and treat all subsequent arguments as non-flag args even if they would otherwise look like a flag.

krader1961:
Not a bug. It's a feature of the `getopt()` libc function that fish uses to parse CLI args.  It's something you just learn about after using the UNIX CLI for a while. It is documented in a few places in the fish documentation. For example, the [argparse](http://fishshell.com/docs/current/commands.html#argparse) command which is how fish functions perform argument parsing. It's just a wrapper around the `getopt()` function that the fish builtin commands use. I don't think it should be mentioned in every single command but it might be worthwhile to add it to the `string` command since it seems to be needed there more often than most other commands.

adamkiss:
I think I see. Thank you.

templateK:
I wasted at least two hours to concatenating string  which is a very basic manipulation.
If I had found this workaround at the official help document then that must been awesome.
 If it is not a bug, it should be documented.

[...]

PR #10319, opened 2024-02-21
functions: handle hostname starting with dash

pgier:
## Description

If a hostname starts with a dash `-` character, the prompt_hostname function fails because the `string` function interprets it as an option instead of an argument.

Related to #4959


zanchey:
Thanks!

PR #10325, opened 2024-02-26
fish_indent: Collapse successive newlines

faho:
## Description

This makes it so fish_indent turns a run of empty lines into one.

So if you have a file like

```fish
echo foo



echo bar
```

it will turn it into

```fish
echo foo

echo bar
```

but

```fish
echo foo
echo bar
```

is untouched.

As a special case, even a single newline at the beginning of a file, before any code, will be removed.

The second commit goes through our share/ to reformat our files according to the new style.

## TODOs:


tests/checks/indent.fish
```diff
@@ -96,10 +95,8 @@ end
 ' | $fish_indent
 
 #CHECK: switch aloha
-#CHECK: 
 #CHECK: {{    }}case alpha
 #CHECK: {{    }}{{    }}echo sup
-#CHECK:
```
faho, line 102: I have to admit I am surprised by this change.

I'm not sure why it happens (do we synthesize an "end" token for this?), but honestly I'm okay with it.

krobelus:
On Mon, Feb 26, 2024 at 07:26:15AM -0800, Fabian Boehm wrote:
> This makes it so fish_indent turns a run of empty lines into one.

we are already pretty opinionated, so this looks fine to me

PR #10329, opened 2024-02-28
Add safe Send/Sync wrapper for main thread data

mqudsi:
Resolve some of the safety issues indicated in https://github.com/fish-shell/fish-shell/commit/e7b94454df61c11c98d29fdbd38e248aa155f690

So long as we insist on having a function that can return a statically allocated copy of data that belongs to and should only be accessed by the main thread, we will need a wrapper that can safely marshal that access.

All (safe) `static` variables need to be `Send` and `Sync`, so this poses a problem for our one-time-alloc, lazy-init variables with `&'static` reference duration that are intended to be accessed only by the main thread.

A `Send` + `Sync` wrapper `MainThread<T>` is introduced which wraps non-`Send`, non-`Sync` type `T` and asserts both `Send` and `Sync`. To ensure safety, only a ~~scoped~~ reference to the underlying `T` may be resolved ~~(in a callback)~~, after asserting the thread id each time. Since the underlying `&T` or `&mut T` is already `!Send` and `!Sync`, we don't have to worry about ~~the callback~~ forwarding that reference to another thread - the compiler will uphold the safety variant.

### DIFF
--- a/src/screen.rs
+++ b/src/screen.rs
@@ -8,6 +8,7 @@
 //! of text around to handle text insertion.
 
 use crate::pager::{PageRendering, Pager};
+use crate::threads::MainThread;
 use std::collections::LinkedList;
 use std::ffi::{CStr, CString};
 use std::io::Write;
@@ -176,7 +177,7 @@ pub struct Screen {
     pub autosuggestion_is_truncated: bool,
 
     /// Receiver for our output.
-    outp: &'static mut Outputter,
+    outp: &'static MainThread<Outputter>,
 
     /// The internal representation of the desired screen contents.
     desired: ScreenData,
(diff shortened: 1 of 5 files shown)


src/screen.rs
```diff
@@ -637,9 +639,9 @@ impl Screen {
             // Either issue a cr to go back to the beginning of this line, or a nl to go to the
             // beginning of the next one, depending on what we think is more efficient.
             if new_y <= zelf.actual.cursor.y {
-                zelf.outp.push(b'\r');
+                zelf.outp.with_mut(|outp| outp.push(b'\r'));
             } else {
-                zelf.outp.push(b'\n');
+                zelf.outp.with_mut(|outp| outp.push(b'\r'));
```
epidemian, line 644: Beware: a `\n` was changed into a `\r` here.

mqudsi: Yikes! Nice catch!

mqudsi:
I pushed an update that reworks the abstraction to remove the reference to `MainThread` from the definition of `Screen`.

mqudsi:
Also pushed an update that addresses the issue that was found with `parser::principal_parser()`, made possible by the change to the `MainThread<T>` abstraction.

PR #10331, opened 2024-02-29
Use `memmap2` crate instead of hand-rolled memory mapping for history files

abhiaagarwal:
## Description

Replace the hand-rolled implementation of memory mapping with the `memmap2` crate. This allows enforcement within the type system that the memory-mapped file is read-only. Removes a bunch of `unsafe`-marked code as well and replaces the various remaining libc `lseek`s with safe `nix` versions.

## TODOs:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -107,6 +107,7 @@ dependencies = [
  "lazy_static",
  "libc",
  "lru",
+ "memmap2",
  "nix",
  "num-traits",
  "once_cell",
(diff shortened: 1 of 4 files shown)


mqudsi (commented):
I'm ambivalent about this. 

The changes look OK at a first glance, but the majority of it is just placing one layer of indirection over the unsafe code we currently have, at the added cost of introducing a dependency.

faho:
Hah, I thought I was gonna be alone in that, I assumed it was just C++-brain talking.

Yeah, I don't love introducing dependencies just to remove a few lines either.

Is there an advantage to memmap2 over our code, or is it just slight *abstraction*?

mqudsi:
Abstractions can be useful (even if they come via an additional dependency) in a lot of cases. If it saves you from having to write a ton of boilerplate for each instance, if it comes with full integration with the rust type system and borrow checker, etc. etc. so I don't at all begrudge you opening this PR. This is only a question of pragmatism for us, and we just have to decide if what benefit *this particular* ffi/interop crate/abstraction provides is worth the cost it brings to the table (more dependencies, worrying about upstream issues, not having full control over the implementation ourselves, possibly bringing in more than we need, etc.)

abhiaagarwal:
If you don't want to include the crate, I can quickly rewrite this to essentially convert the original abstraction to be read_only and the map_anon function to handle the process of converting the shared memory segment into read only!

faho:
Alright, closing this as-is because it appears we have no appetite for it.

If you do get around to those enhancements to what we have that sounds cool!

PR #10336, opened 2024-03-01
Reimplement git version generation ourselves

faho:
This allows us to remove two dependency crates

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -24,6 +24,8 @@ fn main() {
             .unwrap(),
     );
 
+    rsconf::set_env_value("FISH_BUILD_VERSION", &get_version(build_dir));
+
     rsconf::rebuild_if_path_changed("src/libc.c");
     cc::Build::new()
         .file("src/libc.c")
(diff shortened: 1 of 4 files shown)


zanchey:
I haven't had a chance to test, but does this do any better with dirty/non-dirty trees? The CMake/sh version always picked up if the tree was dirty, but the git-version.rs approach only worked if a Rust source file had been modified.

mqudsi:
@zanchey we would have to use [rebuild_if_paths_changed](https://docs.rs/rsconf/latest/rsconf/fn.rebuild_if_paths_changed.html) and pass in all additional paths (share, mainly?) we want build.rs to be re-run if they change. This will mean slightly slower incremental builds after changing only fish files, but *I think* it shouldn't result in actual work for the compiler.

build.rs
```diff
@@ -193,3 +195,27 @@ fn setup_paths() {
...
+
+fn get_version(build_dir: &str) -> String {
+    use std::fs::read_to_string;
+    use std::process::Command;
+
+    if let Ok(var) = std::env::var("FISH_BUILD_VERSION") {
+        return var;
+    }
+
+    let path = PathBuf::from(build_dir).join("version");
```
mqudsi, line 207: I would like to (at some point) rename this file to `FISH_BUILD_VERSION` to make everything more consistent.

mqudsi:
LGTM. I see no need for that dependency.

faho:
> I haven't had a chance to test, but does this do any better with dirty/non-dirty trees?

This is basically the same thing as before. It's really mostly about not needing a dependency that we barely use, and partially about being able to adjust it just to our liking.

Alright, merging, it's simple enough to revert if there are any issues.

build.rs
faho: Tbh my preference would be to pass the version as an environment variable - this would also be much easier for sphinx to read when building manpages (which would allow us to build them before building the binaries, which makes it possible to embed them in a binary).

I haven't figured out how to get cmake to do that, tho

PR #10338, opened 2024-03-02
Fix Vi mode autosuggestions and others

### LINKED ISSUE #3299: fish_vi_mode $ command issues with autosuggestions
- [x] Have you checked if problem occurs with [fish 2.3.1](/fish-shell/fish-shell/releases/tag/2.3.1)?
- [x] Tried fish without third-party customizations _(check `sh -c 'env HOME=$(mktemp -d) fish'`)_?

**fish version installed** _(`fish --version`)_:
git f3e93f06669dc36c68dd86573b4ca7caf59db63a

**OS/terminal used**: 
freebsd via putty, $TERM=xterm
## Reproduction steps
1. Launch fish
2. Type something matching the history so that an autocomplete suggestion comes up
3. Press Esc to leave insert mode
4. Press $ to jump to end of line
## Expected results

Fish should jump to end of line
## Actual results

The first time `$` is pressed, the cursor is moved by one character. Pressing `$` again jumps to the end of line.

### LINKED ISSUE #4683: Unnecessary spaces : kill-word
I have those bindings

```
bind dW kill-bigword
bind dw kill-word
```

When my command line look like this:
<code>echo&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;ls</code>

Doing either one of those keys make this: (when my cursor is at the beginning)
<code>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;ls</code>

The result in vim would be only `ls`

And if my command line look like this (cursor at the beginning also):
<code>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;ls</code>

either one of those keys delete all the input
Whereas in vim the result would be `ls`

What do you think of standardize this behavior and do like vim does ?

(I just realize that same rules was applying to forward-word)

### LINKED ISSUE #10286: Vi-mode poor autosuggestion behavior / Invalid cursor placement and length in normal mode
fish, version 3.7.0
3.7.0
OS: Linux 6.6.14 NixOS (but happens on any OS)
Term: Kitty (but any terminal does this, over ssh, etc.)

----

In vi mode, I noticed an autocomplete issue, say you start here:
image

If you then press `ESC` to enter normal mode, then `l` to attempt to complete the autosuggestion, it doesn't work. You have to press `l` twice to complete.

It seems that the reason for this issue with the autosuggestion or the fact that `ESC` moves you back, but that vi mode lets you move your *normal mode cursor* to *after the last character*, which isn't actually a valid vi place to put the cursor (or a "non-extant character").

Then, when autosuggestion logic checks if you're at the last possible cursor position when you call `forward-char` with `l`, it *should* consider you at the last possible normal mode position, but instead it thinks that there is that one extra place for the cursor to be.

krobelus:
Add a hack to Vi mode to avoid placing the cursor beyond the last character.
This fixes an issue with autosuggestions.  Also fix some other issues.

Vi mode has lots of issues remaining but with hybrid mode it seems usable
to me.

Closes #10286
Closes #3299

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,7 +12,7 @@ fish 3.8.0 (released ???)
    10198 10200 10201 10204 10210 10214 10219 10223 10227 10232 10235 10237 10243 10244 10245
    10246 10251 10260 10267 10281 10347 10366 10368 10370 10371 10263 10270 10272 10276 10277
    10278 10279 10291 10293 10305 10306 10309 10316 10317 10327 10328 10329 10330 10336 10340
-   10345 10346 10353 10354 10356 10372 10373
+   10345 10346 10353 10354 10356 10372 10373 3299 10360
 
 The entirety of fish's C++ code has been ported to Rust (:issue:`9512`).
 This means a large change in dependencies and how to build fish.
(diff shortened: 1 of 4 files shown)


krobelus:
I would merge this soon, the only uncertainty is whether we want to repurpose `fish_cursor_selection_mode` like this.
For Vi mode, the changes are fine but it would completely break users who set `fish_cursor_selection_modee=inclusive` in emacs mode (because they will no longer be able to append to the command line in that mode).
So if there's a chance that anyone legitimately uses this, I'll add a second knob. Of course we don't have telemetry so it's hard to tell

krobelus:
ok it looks like users might set their cursor to an underscore even outside Vi mode? And then `fish_cursor_selection_mode=inclusive` is probably valid even in Emacs style?

PR #10340, opened 2024-03-03
docs: Correct default value of `read` function in `read.rst`

shuuji3:
## Description

Hi, I found the documentation about the default value of `read` function is a bit different from the actual implementation (no `-n` in the documentation). 

ref. 
- https://github.com/fish-shell/fish-shell/blob/33a7172ee8f99a32a18ac627642831bbf9ac8523/src/builtins/read.rs#L453
- https://github.com/fish-shell/fish-shell/blob/master/src/builtins/shared.rs#L22

Fixes issue N/A

## TODOs:


faho:
Thanks!

Just to be clear: The second `-n` doesn't really have an effect - a trailing newline on the prompt is removed (or otherwise most `fish_prompt` functions would have to use `echo -n`).

PR #10346, opened 2024-03-06
cmake: Use cargo --target-dir

faho:
This isolates cargo's output in the cmake build directory, so it no longer creates a target/ in the repository.

Unfortunately cargo still decides to create
"x86_64-unknown-linux-gnu/debug/" (target-triple/profile), there's no way to get it to build *in this directory*.

Note that this *will* create about ~600M in the build directory instead of sharing among them.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,12 +59,13 @@ FILE(GLOB sources src/* src/*/* src/*/*/*)
 # Define a function to link dependencies.
 function(FISH_LINK_DEPS_AND_SIGN target)
   add_custom_command(
-    OUTPUT ${rust_target_dir}/${rust_profile}/${target}
+    OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/${Rust_CARGO_TARGET}/${rust_profile}/${target}
     COMMAND ${CMAKE_COMMAND} -E env
       ${VARS_FOR_CARGO}
     ${Rust_CARGO} ARGS build --bin ${target}
       $<$<CONFIG:Release,RelWithDebInfo>:--release>
       --target ${Rust_CARGO_TARGET}
+      --target-dir ${CMAKE_CURRENT_BINARY_DIR}
       ${CARGO_FLAGS}
       ${FEATURES_ARG}
     DEPENDS ${sources} src/bin/${target}.rs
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -59,12 +59,13 @@ FILE(GLOB sources src/* src/*/* src/*/*/*)
...
 function(FISH_LINK_DEPS_AND_SIGN target)
   add_custom_command(
-    OUTPUT ${rust_target_dir}/${rust_profile}/${target}
+    OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/${Rust_CARGO_TARGET}/${rust_profile}/${target}
     COMMAND ${CMAKE_COMMAND} -E env
       ${VARS_FOR_CARGO}
     ${Rust_CARGO} ARGS build --bin ${target}
       $<$<CONFIG:Release,RelWithDebInfo>:--release>
       --target ${Rust_CARGO_TARGET}
+      --target-dir ${CMAKE_CURRENT_BINARY_DIR}
```
krobelus, line 68: we already override the target dir witih `CARGO_TARGET_DIR` which is passed via `VARS_FOR_CARGO`.
This is why `docker/docker_run_tests.sh docker/opensuse-tumbleweed.docker` works, which mounts the source tree read-only.
But it looks like our `"CARGO_TARGET_DIR=${FISH_RUST_BUILD_DIR}"` is inconsistent with `${rust_target_dir}`.
We should probably fix that to use the same target dir everywhere, at least inside cmake land.

I think today the top-level `target/` directory should only be created by `cargo b` and things like `rust-analyzer` and `cargo clippy`.
I don't know if there's a good way to override this but long-term we should try to get rid of cmake anyway.
Right now, adding a new source file requires a manual cmake rerun.

faho: > we already override the target dir witih CARGO_TARGET_DIR which is passed via VARS_FOR_CARGO.

Huh, cool. This is unnecessary then.

I think my "target" might have been created by rust-analyzer then.

> But it looks like our "CARGO_TARGET_DIR=${FISH_RUST_BUILD_DIR}" is inconsistent with ${rust_target_dir}.

"rust_target_dir" *is* FISH_RUST_BUILD_DIR plus the arch-os-triple.

> we should try to get rid of cmake anyway

Of course.

henrikhorluck: We could make a `.cargo/config.toml` with `target-dir = "build/cargo/build"`, which rust-analyzer and friends pick up

https://doc.rust-lang.org/cargo/reference/config.html#configuration

krobelus:
> We could make a `.cargo/config.toml` with `target-dir = "build/cargo/build"`, which rust-analyzer and friends pick up
> 
> https://doc.rust-lang.org/cargo/reference/config.html#configuration

Sounds reasonable. It would be good to share the cache.

It would be slightly weird if the build directory is not "build/" but that's a harmless cosmetic issue in most cases.
I'm assuming that cargo creates the target dir if it doesn't exist.

PR #10347, opened 2024-03-07
Shortened some which.fish completions.

amiyzku:
## Description

Shortened some which.fish completions.

Related to https://github.com/fish-shell/fish-shell/issues/6981.

## TODOs:


share/completions/which.fish
```diff
@@ -4,9 +4,9 @@ if which -v >/dev/null 2>/dev/null # GNU
     complete -c which -l skip-alias -d "Ignore option '--read-alias'"
     complete -c which -l read-functions -d "Read shell function definitions from stdin, reporting matching ones on stdout"
     complete -c which -l skip-functions -d "Ignore option '--read-functions'"
-    complete -c which -l skip-dot -d "Skip directories in PATH that start with a dot"
-    complete -c which -l skip-tilde -d "Skip directories in PATH that start with a tilde and executables which reside in the HOME directory"
-    complete -c which -l show-dot -d "If a directory in PATH starts with a dot and a matching executable was found for that path, then print './programname'"
+    complete -c which -l skip-dot -d "Skip directories in PATH starting with a dot"
+    complete -c which -l skip-tilde -d "Skip directories in PATH starting with a tilde and executables in the HOME directory"
```
mqudsi, line 8: Can shorten this further to "starting with tilde and executables in \$HOME"

share/completions/which.fish
```diff
@@ -4,9 +4,9 @@ if which -v >/dev/null 2>/dev/null # GNU
...
+    complete -c which -l show-dot -d "If a directory in PATH starts with a dot and a matching executable was found, print './programname'"
```
mqudsi, line 9: "For matches in PATH dirs that start with a dot, print './programname'" is shorter

mqudsi (commented):
This is fine, but I made some suggestions to shorten it further. Also you can change "directories" to "dirs" everywhere *if you like*.

amiyzku:
Thank you for the review! I've incorporated your feedback and made the suggested changes. Please take another look and let me know if there's anything else that needs to be addressed.

mqudsi (changes requested):
FYI I had intentionally written `\$HOME` instead of `$HOME` because in a double quoted (`"`) string, `$HOME` will expand to the actual value of `$HOME`. You need to either escape the `$` sign like `\$` or use `'` single quotes instead of double quotes `"`.

amiyzku:
Thank you for the review! I've updated the code to escape  as suggested.

mqudsi (approved):

PR #10351, opened 2024-03-08
fire fish_prompt on prompt repaint

### LINKED ISSUE #10350: `command -f repaint` does not fire event `fish_prompt`
fish, version 3.6.4

How to reproduce:

```
function on_prompt --on-event fish_prompt
    echo hello
    echo
end
bind \eg 'echo; commandline -f repaint'

# press Alt+G
```

The event should be fired under

```
rl::RepaintMode | rl::ForceRepaint | rl::Repaint => {
   ...
}
```

but I am not sure where exactly.

Related: #10289

iacore:
## Description

Now `commandline -f repaint` will emit the event `fish_prompt`. Probably some other ways will trigger this event too.

Fixes #10350 

## TODOs:

- [x] (N/A: not regression) Tests have been added for regressions fixed

### DIFF
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -2115,6 +2115,9 @@ impl ReaderData {
                     }
                     // Else we repaint as normal.
                 }
+                if !self.conf.event.is_empty() {
+                    event::fire_generic(self.parser(), self.conf.event.to_owned(), vec![]);
+                }
                 self.exec_prompt();
                 self.screen.reset_line(/*repaint_prompt=*/ true);
                 self.layout_and_repaint(L!("readline"));


src/reader.rs
```diff
@@ -2115,6 +2115,9 @@ impl ReaderData {
                     }
                     // Else we repaint as normal.
                 }
+                if !self.conf.event.is_empty() {
+                    event::fire_generic(self.parser(), self.conf.event.to_owned(), vec![]);
```
krobelus, line 2119: what is `fish_prompt` used for today and what do you want to use it for?

iacore: shell integration https://github.com/kovidgoyal/kitty/issues/7200

iacore:
Maybe `commandline -f repaint-mode` need to trigger this event too? I'm not sure. Set this PR as draft.

src/reader.rs
krobelus: I think we should first consider implementing `OSC 133` prompt/output marking ourselves, as upstream suggests.
This way terminals and shells don't need to know about each other.
We can probably bake it into the Rust code unless we want to control it from script for whatever reason.
As is, this change is not really appropriate for our only real use of `--on-event fish_prompt`, which is `__fish_enable_bracketed_paste` - though we might need to move that anyway because the enable/disable logic doesn't seem to work in all cases (That's my guess because I made it toggle CSI u too and that sometimes lingers when switching to bash etc.).

iacore: > OSC 133

That's probably better.

PR #10352, opened 2024-03-08
Add support for OSC 133

iacore:
## Description

Support marking prompts with OSC 133

Supercedes https://github.com/fish-shell/fish-shell/pull/10351.
## TODOs:
- [ ] Find out how feature detection is done

### DIFF
--- a/src/curses.rs
+++ b/src/curses.rs
@@ -79,13 +79,6 @@ pub struct Term {
     pub clr_eol: Option<CString>,
     pub clr_eos: Option<CString>,
 
-    // OSC 133  docs: https://iterm2.com/documentation-escape-codes.html
-    pub prompt_start: Option<CString>,  // "\x1b133;A\x07"
-    pub prompt_end: Option<CString>,    // "\x1b133;B\x07"
-    pub command_start: Option<CString>, // "\x1b133;C\x07"
-    // pub command_end_before_status: Option<CString>, // "\x1b133;D;<exitcode>;\x07"
-    // pub command_end_after_status: Option<CString>,
-
     // Number capabilities
     pub max_colors: Option<usize>,
     pub init_tabs: Option<usize>,
@@ -244,10 +237,6 @@ impl Term {
             parm_right_cursor: get_str_cap(&db, "RI"),
             clr_eol: get_str_cap(&db, "ce"),
             clr_eos: get_str_cap(&db, "cd"),
-
-            prompt_start: if support_osc133 { Some(CString::new("\x1b]133;A\x1b\x5c").unwrap()) } else { None },
-            prompt_end: if support_osc133 { Some(CString::new("\x1b]133;B\x1b\x5c").unwrap()) } else { None },
-            command_start: if support_osc133 { Some(CString::new("\x1b]133;C\x1b\x5c").unwrap()) } else { None },
 
             // Number capabilities
             max_colors: get_num_cap(&db, "Co"),
--- a/src/curses.rs
+++ b/src/curses.rs
@@ -79,6 +79,13 @@ pub struct Term {
     pub clr_eol: Option<CString>,
     pub clr_eos: Option<CString>,
 
+    // OSC 133  docs: https://iterm2.com/documentation-escape-codes.html
+    pub prompt_start: Option<CString>,  // "\x1b133;A\x07"
+    pub prompt_end: Option<CString>,    // "\x1b133;B\x07"
+    pub command_start: Option<CString>, // "\x1b133;C\x07"
+    // pub command_end_before_status: Option<CString>, // "\x1b133;D;<exitcode>;\x07"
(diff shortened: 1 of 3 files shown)


src/curses.rs
```diff
@@ -235,6 +245,10 @@ impl Term {
             clr_eol: get_str_cap(&db, "ce"),
             clr_eos: get_str_cap(&db, "cd"),
 
+            prompt_start: if support_osc133 { Some(CString::new("\x1b]133;A\x1b\x5c").unwrap()) } else { None },
+            prompt_end: if support_osc133 { Some(CString::new("\x1b]133;B\x1b\x5c").unwrap()) } else { None },
+            command_start: if support_osc133 { Some(CString::new("\x1b]133;C\x1b\x5c").unwrap()) } else { None },
```
krobelus, line 250: no need to open a new pull request, you can update the old one, or at least cross-reference https://github.com/fish-shell/fish-shell/pull/10351 for future readers for the benefit of future readers
Looks good to me overall.

I think we don't even need to define `prompt_start` fields, can simply hardcode them at the callsite.
Also make sure to remove unused code.

src/screen.rs
```diff
@@ -832,6 +832,7 @@ impl Screen {
         // Output the left prompt if it has changed.
         if left_prompt != zelf.actual_left_prompt {
             zelf.r#move(0, 0);
+            zelf.write_mbs_if_some(&term.and_then(|term| term.prompt_start.as_ref()));
             let mut start = 0;
             for line_break in left_prompt_layout.line_breaks {
                 zelf.write_str(&left_prompt[start..line_break]);
```
krobelus, line 838: > run bind \eg 'echo; git status; git remote -v; echo; commandline -f repaint'

right, we offer bindings like `\el` and `\w` so this should definitely be supported

src/screen.rs
```diff
@@ -832,6 +832,7 @@
         // Output the left prompt if it has changed.
         if left_prompt != zelf.actual_left_prompt {
             zelf.r#move(0, 0);
+            zelf.write_mbs_if_some(&term.and_then(|term| term.prompt_start.as_ref()));
```
krobelus, line 835: How is this feature used by terminals?
If it's just about not overwriting output, we should do it for right prompts too.
If it's meant to allow jumping between prompts, then the right prompt shouldn't get it.

src/curses.rs
```diff
@@ -201,6 +208,9 @@
     /// Initialize a new `Term` instance, prepopulating the values of all the curses string
     /// capabilities we care about in the process.
     fn new(db: terminfo::Database) -> Self {
+        // todo: how to do feature detection of OSC 133?
+        let support_osc133 = false;
```
krobelus, line 212: We can probably enable it unconditionally.
I'm not aware of a relevant terminal that doesn't have basic xterm compatibility.
We can add a knob to specify xterm compatibility that or turn off OSC 133 specifically but I think that can be done later.

We do have a bug where we fail to turn off bracketed paste and focus reporting for commands launched from bindings,
like `bind \eg sh` for example.
But we can definitely fix that, and it's not relevant here.

src/curses.rs
```diff
@@ -79,6 +79,13 @@
     pub clr_eol: Option<CString>,
     pub clr_eos: Option<CString>,
 
+    // OSC 133  docs: https://iterm2.com/documentation-escape-codes.html
+    pub prompt_start: Option<CString>,  // "\x1b133;A\x07"
+    pub prompt_end: Option<CString>,    // "\x1b133;B\x07"
+    pub command_start: Option<CString>, // "\x1b133;C\x07"
+    // pub command_end_before_status: Option<CString>, // "\x1b133;D;<exitcode>;\x07"
+    // pub command_end_after_status: Option<CString>,
```
krobelus, line 87: Might as well add command start/end too here.
Emit it next to where we emit `fish_preexec`/`fish_postexec`, passing `eval_res.status.status_value()`.
If you choose to do that, make sure it plays well with fish's default prompt which already renders a failing status in bold red.

src/screen.rs
iacore: as for kitty, it only needs marking of

- start of PS1
- start of PS2
- start of application output

https://sw.kovidgoyal.net/kitty/shell-integration/#notes-for-shell-developers

src/curses.rs
iacore: if the output is not a terminal, what will happen if it receives those escseq

iacore:
Why is `CString` used everywhere? On every write it literally recalculates the length of the string

```rust
    pub fn tputs(&mut self, str: &CStr) {
        self.begin_buffering();
        let _ = self.write(str.to_bytes());
        self.end_buffering();
    }
```

iacore:
I am not sure where to print the other escape sequences. Anyone can take over this PR now.

PR #10353, opened 2024-03-09
Simplify Parser's shared-from-this pattern

The0x539:
## Description
Rust 1.60.0 added `Rc::new_cyclic`, which makes it a lot easier to set up an intentional reference cycle. Ideally we'd completely remove this oddity, especially because it looks like only one `Parser` is ever constructed or passed around (and it lives in a static variable!). However, that's a more disruptive change than I'm comfortable attempting as a complete newcomer.

## TODOs:
- [x] Changes to fish usage are reflected in user documentation/manpages. (N/A)
- [x] Tests have been added for regressions fixed (N/A)
- [x] User-visible changes noted in CHANGELOG.rst (N/A)

### DIFF
--- a/src/global_safety.rs
+++ b/src/global_safety.rs
@@ -1,6 +1,5 @@
 use crate::flog::FLOG;
-use std::cell::{Ref, RefCell, RefMut};
-use std::rc::{Rc, Weak};
+use std::cell::{Ref, RefMut};
 use std::sync::atomic::{AtomicBool, AtomicPtr, Ordering};
 use std::sync::MutexGuard;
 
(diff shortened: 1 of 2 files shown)


mqudsi (approved):
Thanks. This looks good to me, and I'm personally fine with the MSRV. I agree with not getting rid of the cyclic reference just yet; there is some groundwork laid out for supporting proper backgrounding and multithreading that will (someday) use multiple `Parser` instances.

I'll leave this open for a bit in case anyone else has any thoughts, if there are no objections I'll merge it later this week DV.

zanchey:
MSRV is already 1.67, so there's no bump here :-)

PR #10354, opened 2024-03-09
Address new clippy lints

The0x539:
Changes should be largely self-explanatory - nightly added some new lints.
Fixes all clippy diagnostics except for `incompatible_msrv`: a few files use `Option::is_some_and` and `Result::is_ok_and`, which were stabilized in 1.70 vs the current MSRV of 1.67, and I wasn't sure whether to bump the MSRV to match reality, write a polyfill, or switch to something uglier.
I'm curious as to how these scattered `let`-`else` usages happened.

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -2288,9 +2288,7 @@ impl<'a> Traversal<'a> {
 impl<'a> Iterator for Traversal<'a> {
     type Item = &'a dyn Node;
     fn next(&mut self) -> Option<&'a dyn Node> {
-        let Some(node) = self.stack.pop() else {
-            return None;
-        };
+        let node = self.stack.pop()?;
         // We want to visit in reverse order so the first child ends up on top of the stack.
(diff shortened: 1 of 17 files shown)


krobelus:
> Fixes all clippy diagnostics except for `incompatible_msrv`: a few
> files use `Option::is_some_and` and `Result::is_ok_and`, which were
> stabilized in 1.70 vs the current MSRV of 1.67, and I wasn&#39;t
> sure whether to bump the MSRV to match reality, write a polyfill,
> or switch to something uglier.

we have a `is_some_and()` etc. polyfills, we should silence any warnings

> I'm curious as to how these scattered `let`-`else` usages happened.

I don't think it's clear which one is better. Your changes look reasonable to me.

The0x539:
> we should silence any warnings
Can do.

An even newer nightly added some additional lints. `assigning_clones` seems like a bit much.
With that warning silenced, we're now all green on 2024-03-06 nightly.

PR #10355, opened 2024-03-09
Begin switching to `File` from `OwnedFd` in a few places

mqudsi:
This is a step towards converting `wopen_cloexec()` to return `File` instead of `OwnedFd`/`AutocloseFd`.¹

In addition to letting us use native standard library functions instead of unsafe libc calls, we gain additional semantic safety because `File` operations that manipulate the state of the fd (e.g. `File::seek()`) require a `&mut` reference to the `File`, whereas using `RawFd` or `OwnedFd` everywhere leaves us in a position where it's not clear whether or not other references to the same fd will manipulate its underlying state.

¹ We actually wouldn't even need `wopen_cloexec()` at all (just a widechar wrapper) as Rust's native `File::open()`/`File::create()` functionality uses `FD_CLOEXEC` internally.

The patch itself is fine, tests pass, etc. but I would like to open the floor for any discussions on the nature of the change itself (moving more code from our libc wrappers to rust std). As you can see, this change also uses native `File` functionality as exposed via the rust standard library (seek, etc) instead of calling libc operations on the raw fd itself, reducing our ffi code and removing a number of `unsafe` calls.

The idea is that after all call sites have been converted to use `File`, we can go back in and change `wopen_cloexec()` to return a `File` directly.

### DIFF
--- a/src/history.rs
+++ b/src/history.rs
@@ -21,17 +21,18 @@ use std::{
     borrow::Cow,
     collections::{BTreeMap, HashMap, HashSet, VecDeque},
     ffi::CString,
-    io::{BufRead, Read, Write},
+    fs::File,
+    io::{BufRead, Read, Seek, SeekFrom, Write},
     mem,
     num::NonZeroUsize,
     ops::ControlFlow,
-    os::fd::{AsFd, AsRawFd, OwnedFd, RawFd},
+    os::fd::{AsFd, AsRawFd, RawFd},
     sync::{Arc, Mutex, MutexGuard},
     time::{Duration, SystemTime, UNIX_EPOCH},
 };
 
 use bitflags::bitflags;
-use libc::{fchmod, fchown, flock, fstat, ftruncate, lseek, LOCK_EX, LOCK_SH, LOCK_UN, SEEK_SET};
+use libc::{fchmod, fchown, flock, fstat, LOCK_EX, LOCK_SH, LOCK_UN};
 use lru::LruCache;
 use nix::{fcntl::OFlag, sys::stat::Mode};
 use rand::Rng;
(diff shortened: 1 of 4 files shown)


src/history.rs
```diff
@@ -475,8 +476,12 @@ impl HistoryImpl {
...
         let _profiler = TimeProfiler::new("load_old");
         if let Some(filename) = history_filename(&self.name, L!("")) {
-            let Ok(fd) = wopen_cloexec(&filename, OFlag::O_RDONLY, Mode::empty()) else {
-                return;
+            let mut file = {
+                let Ok(fd) = wopen_cloexec(&filename, OFlag::O_RDONLY, Mode::empty()) else {
+                    return;
+                };
+                // TODO: Return `File` directly from wopen_cloexec
+                unsafe { File::from_raw_fd(fd.into_raw_fd()) }
```
PolyMeilex, line 484: Why not safe from? That's kinda the main reason why I made it an `OwnedFd`.
```suggestion
                File::from(fd)
```

src/history.rs
```diff
@@ -654,45 +659,41 @@ impl HistoryImpl {
...
             }
 
-            let target_fd_before = wopen_cloexec(
+            let target_file_before = wopen_cloexec(
                 &target_name,
                 OFlag::O_RDONLY | OFlag::O_CREAT,
                 HISTORY_FILE_MODE,
-            );
+            )
+            .map(|fd| unsafe { File::from_raw_fd(fd.into_raw_fd()) });
```
PolyMeilex, line 676: Same case here, the fd is already a safe Owned variant.
```suggestion
            .map(File::from);
```

src/history.rs
mqudsi: Thanks. I forgot about that conversion!

mqudsi:
Thanks, @PolyMeilex. I forgot about that conversion and to me, it isn't obvious that it is safe to convert any owned fd to a file (but in retrospect, since all io operations are fallible, it is fine).

PR #10365, opened 2024-03-11
Deprecate builtin test's one- and zero-argument modes

### LINKED ISSUE #2037: test -n not working as expected
I have the following block:

``` fish
if begin; test -n $SSH_CLIENT; or test -n $SSH_TTY; end
echo "ssh (0)"
else
echo "not ssh (0)"
end
```

If I am correct, it should only print "ssh" if either of the environment variables `$SSH_CLIENT` of `$SSH_TTY` are defined. To verify that neither are defined:

```
   eric@ubuntu ~> printenv | grep SSH
   SSH_AUTH_SOCK=/run/user/1000/keyring-bKvxzV/ssh
```

and yet the output I get is "ssh". If I reverse the code block and do:

``` fish
if begin; test -z $SSH_CLIENT; and test -z $SSH_TTY; end
echo "not ssh (1)"
else
echo "ssh (1)"
end
```

It works as expected. 

So, running these two blocks where neither `$SSH_CLIENT` of `$SSH_TTY`are defined, the output I get is:

```
ssh (0)
not ssh (1)
```

and running these two blocks where both $SSH_CLIENT of $SSH_TTY are defined, the output I get is:

```
ssh (0)
ssh (1)
```

The -n flag does not work. The -z flag does.

```
> fish --version
fish, version 2.1.1

> uname -a
Linux ubuntu 3.16.0-30-generic #40~14.04.1-Ubuntu SMP Thu Jan 15 17:43:14 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux
```

faho:
This introduces the "test-require-arg" feature flag.

When enabled, it removes builtin test's zero and one argument special modes.

That means:

- `test -n` returns false
- `test -z` returns true
- `test -x`, `test foo` and `test ""` error out

Currently, all of these return true, except for the last which returns false, all silently.

This is a recognition that this is a frequent cause of problems. While we could (and still should) introduce a better replacement for `test`, this will fix a frequently recurring question, and remove awkward explanations we need to give.

Nobody *needs* zero- or one-argument `test`, and few people *want* it as a shortcut.

Fixes #2037.

## TODOs:

- [ ] Is this the right fix or do we e.g. error out for *all* one-argument calls?
- [x] Documentation
- [x] Do we add a deprecation FLOG for these if the feature flag is disabled? Or do we give people a mostly-correct regex to `grep` their code instead?
- [ ] CHANGELOG

### DIFF
--- a/src/builtins/test.rs
+++ b/src/builtins/test.rs
@@ -1,5 +1,6 @@
 use super::prelude::*;
 use crate::common;
+use crate::future_feature_flags::{feature_test, FeatureFlag};
 
 mod test_expressions {
     use super::*;
@@ -552,6 +553,10 @@ mod test_expressions {
                 );
             }
 
+            if feature_test(FeatureFlag::test_require_arg) {
+                return self.error(start, sprintf!("Unknown option at index %u", start));
+            }
+
             return JustAString {
                 arg: self.arg(start).to_owned(),
                 range: start..start + 1,
@@ -718,9 +723,6 @@ mod test_expressions {
             err: &mut WString,
             program_name: &wstr,
         ) -> Option<Box<dyn Expression>> {
-            // Empty list and one-arg list should be handled by caller.
-            assert!(args.len() > 1);
-
             let mut parser = TestParser {
                 strings: args,
                 errors: Vec::new(),
@@ -1019,15 +1021,32 @@ pub fn test(parser: &Parser, streams: &mut IoStreams, argv: &mut [&wstr]) -> Opt
         .collect();
     let args: &[WString] = &args;
 
-    if argc == 0 {
-        return STATUS_INVALID_ARGS; // Per 1003.1, exit false.
-    } else if argc == 1 {
-        // Per 1003.1, exit true if the arg is non-empty.
-        return if args[0].is_empty() {
-            STATUS_CMD_ERROR
-        } else {
-            STATUS_CMD_OK
-        };
+    if feature_test(FeatureFlag::test_require_arg) {
+        if argc == 0 {
+            streams.err.appendln(wgettext_fmt!(
+                "%ls: Expected at least one argument",
+                program_name
+            ));
+            builtin_print_error_trailer(parser, streams.err, program_name);
+            return STATUS_INVALID_ARGS;
+        } else if argc == 1 {
+            if args[0] == "-n" {
+                return STATUS_CMD_ERROR;
+            } else if args[0] == "-z" {
+                return STATUS_CMD_OK;
+            }
+        }
+    } else {
+        if argc == 0 {
+            return STATUS_INVALID_ARGS; // Per 1003.1, exit false.
+        } else if argc == 1 {
+            // Per 1003.1, exit true if the arg is non-empty.
+            return if args[0].is_empty() {
+                STATUS_CMD_ERROR
+            } else {
+                STATUS_CMD_OK
+            };
+        }
     }
 
     // Try parsing
--- a/src/builtins/test.rs
+++ b/src/builtins/test.rs
@@ -1,6 +1,7 @@
 use super::prelude::*;
 use crate::common;
 use crate::future_feature_flags::{feature_test, FeatureFlag};
+use crate::should_flog;
 
 mod test_expressions {
     use super::*;
@@ -1036,17 +1037,31 @@ pub fn test(parser: &Parser, streams: &mut IoStreams, argv: &mut [&wstr]) -> Opt
                 return STATUS_CMD_OK;
             }
         }
-    } else {
-        if argc == 0 {
-            return STATUS_INVALID_ARGS; // Per 1003.1, exit false.
-        } else if argc == 1 {
-            // Per 1003.1, exit true if the arg is non-empty.
-            return if args[0].is_empty() {
-                STATUS_CMD_ERROR
-            } else {
-                STATUS_CMD_OK
-            };
+    } else if argc == 0 {
+        if should_flog!(deprecated_test) {
+            streams.err.appendln(wgettext_fmt!(
+                "%ls: called with no arguments. This will be an error in future.",
+                program_name
+            ));
+            streams.err.append(parser.current_line());
         }
+        return STATUS_INVALID_ARGS; // Per 1003.1, exit false.
+    } else if argc == 1 {
+        if should_flog!(deprecated_test) {
+            if args[0] != "-z" {
+                streams.err.appendln(wgettext_fmt!(
+                    "%ls: called with one argument. This will return false in future.",
+                    program_name
+                ));
+                streams.err.append(parser.current_line());
+            }
+        }
+        // Per 1003.1, exit true if the arg is non-empty.
+        return if args[0].is_empty() {
+            STATUS_CMD_ERROR
+        } else {
+            STATUS_CMD_OK
+        };
     }
 
     // Try parsing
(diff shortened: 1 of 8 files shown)


faho:
(I'm pretty sure this will conflict with #10357, that's *fine*, this is easy to redo)

mqudsi:
For posterity: this breaks with posix?

faho:
> For posterity: this breaks with posix?

Yes. Posix explicitly demands the zero- and one-argument behavior as we implement it currently.

This is a regular source of confusion and so I believe the break is warranted.

mqudsi:
After mulling it over, I'm more open to this change. The "error out" behavior is easier to stomach vs the more "silent breakage" changes, but we've seen *so many* of these issues that I think it's overall warranted.

Funnily enough, the part I hate the most about this is "oh no, not another feature flag!"

src/builtins/test.rs
```diff
@@ -1028,15 +1030,32 @@ pub fn test(parser: &Parser, streams: &mut IoStreams, argv: &mut [&wstr]) -> Opt
...
-            STATUS_CMD_OK
-        };
+    if feature_test(FeatureFlag::test_require_arg) {
+        if argc == 0 {
+            streams.err.appendln(wgettext_fmt!(
+                "%ls: Expected at least one argument",
+                program_name
+            ));
+            builtin_print_error_trailer(parser, streams.err, program_name);
+            return STATUS_INVALID_ARGS;
```
krobelus, line 1040: Good change.

One alternative is to keep returning the same exit code as today, but only print the error. This way we wouldn't really need a feature flag.  I'm fine with either.

krobelus (approved):

src/builtins/test.rs
mqudsi: That's not a bad idea. That could be our "depreciation phase".

krobelus: correction, we'd still want a feature flag

faho:
Alright, summarizing, I think we're agreed that a change is good. I think it's not possible to do it without a feature flag.

I also think we're agreed that `test` and `test foo` (where "foo" is anything that's not a valid operator) should error out. This also applies to `test -n "..." -a foo`. The implicit string test is horrible and should be removed.

So, the remaining cases are:

- `test -n` and `test -z` with no further argument - currently these are both true (because it's `test -n "-n"`), my change here makes `test -n` false. The alternative is to make this error out.
- `test -d` with no further argument (where `-d` is any valid operator that isn't `-n` or `-z`). These currently are all true, this makes them error out. The alternative is to make them return false ("the empty filename isn't a directory").

My idea is that this proposal allows you to do `test -n $foo`, with a variable $foo that you *know* is either zero or one elements (or do specifically `test -n $foo[1]`), and it will do the right thing for if it is empty.

The counter-argument is that that will still break if $foo contains *more* than one element and that you should quote it. The counter to that is that `test` is fundamentally not suited for testing lists - `test -d "$foo"` will not do the right thing if $foo has multiple elements no matter what you do.

krobelus:
> The counter-argument is that that will still break if $foo contains more than one element

right so sounds like we don't want to approve of any one-argument invocation -- knowing this, I'd keep the current exit code, to prevent disatrous breakage (but do print an error)

faho:
> right so sounds like we don't want to approve of any one-argument invocation

Well, no. We don't approve of `test "foo"`, but `test -n` and `test -z` I can see being okay - that's the question I was posing.

> knowing this, I'd keep the current exit code, to prevent disatrous breakage (but do print an error)

I still do not believe this is helpful - you would have to silence the error to prevent breakage *anyway*, and if you are touching the line anyway you can just quote the variable instead.

An error message in a prompt or keybinding, or when you redirect stderr (because many things print on stderr! often useful output!) is already breakage.

We could add a FLOG message to any one-argument test invocation that you can turn on? So you would do `fish -d deprecated-test`, and it would show any deprecated invocation (with location).

krobelus:
I understand that extra output is breakage, sorry if I gave the wrong impression.
I was only talking about disastrous breakage like `test $dir && rm -rf "$dir/"`

faho:
>  I was only talking about disastrous breakage like test $dir && rm -rf "$dir/"

`test $dir` would change from true to false even if $dir is non-empty (and would stay false if it is empty), so this wouldn't remove the directory. This would never run `rm -rf /`.

What would break is something like `test -d $dir || rm -rf "$dir/"`, which would print an error and run `rm -rf /`.

Personally, I feel like having that behind a feature flag is already a lot. We could go further by adding FLOG messages. It would also be quite cool to have some form of *static* checking for this (unfortunately writing a good `grep` invocation here is pretty hard).

It would also be possible to have an in-between step where we *warn* about the change, but I would not want to have that as the endpoint. I would like to end up without errors where it returns true.

So how about this:

1. We add the feature flag (default off) and a FLOG category (default off) that logs affected test invocations
2. Some releases later: We turn on the FLOG category by default (but only if the feature flag is off)
3. Another release or two later: We turn on the feature flag by default

Note that FLOG isn't redirectable via stderr - it'll go to fish's `--debug-output`. We could make it so it only fires if test's stderr isn't redirected (or we make it an stderr message that is only printed if the FLOG category is enabled).

faho:
Alright, I've added the FLOG message (in category "deprecated-test") and documented it.

Since this is a feature flag that would introduce a backwards-incompatible change, I'm going to wait for a sort of "final call" for comments before merging.

src/flog.rs
```diff
@@ -79,6 +79,8 @@ pub mod categories {
 
         (warning_path, "warning-path", "Warnings about unusable paths for config/history (on by default)", true);
 
+        (deprecated_test, "deprecated-test", "Warning about using test's zero- or one-argument modes (`test -d $foo`), which will be changed in future.");
```
krobelus, line 82: Fine by me.

Given what I know I'm fine with enabling this warning now already
(and I'd probably not change the exit code but I'm not gonna fight)
In the light of that I'd also rephrase "will change" to "may change"

faho: To be clear: As this PR stands, *by default*, it will not do anything.

If the warning is enabled, it will show the warning, the exit code will remain.

If the feature flag is enabled, the exit code will change.

The end goal is to have the behavior that I consider best, which is that `test -n` and `test -z` are fine with no additional arguments, anything else is an error.

The plan is to stage this carefully:

- People can, once this is merged, enable the warning. Once they've fixed their code, they can set the feature flag
- The next step is to enable the warning, but keep the flag off by default, so the exit codes stay
- The step after that is to enable the warning and the flag, but allow disabling them
- The step after that is to make the flag read-only and remove the warning - now we have the behavior we would want from `test`

That means we have *a step* where there's a warning but no change in the exit code (by default), but it's only a step. I do not consider "warn about bad behavior but keep it" to be a good final outcome, I would like to go further and end at "okay behavior".

But I also believe that giving people some time where they *know* the change is coming without already throwing warnings at them - which is, as we have discussed, itself a breaking change for many uses, even if it is easy to revert. That's why I would like to have that first step.

krobelus: I think everything that can be said has been said already. You go ahead.
I don't fully agree but my expertise in this area is quite limited.

krobelus: > The plan is to stage this carefully:
> 
> People can, once this is merged, enable the warning. Once they've fixed their code, they can set the feature flag

yeah you're right that announcing things in release notes before we change anything is the proper way to do it.
Then plugin authors can react to it. Hard to tell if it matters in this case but it's good practice

PR #10367, opened 2024-03-12
Allow building fish as self-installable

faho:
# Description

Here's the teaser:

```fish
cargo install --git https://github.com/faho/fish-shell --branch fish-installer
~/.cargo/bin/fish --install
~/.cargo/bin/fish
```

This is how you can get a running fish without root permissions.

You can also use it to build a 100% statically linked fish that you can then copy to another system (with the same architecture and operating system). I've been able to do that for x86_64 linux, and cross-compile one for aarch64 linux. (using musl because glibc doesn't do static linking)

I believe it would be possible to allow this as an option, and to add a static build to fishshell.com so people can just download a fish, without compilation.

I believe this is superior to appimages, which I have recently discovered require libfuse2, which I didn't have installed, along with other dependencies like a compatible libc (so you would have to build on an old glibc system, and additionally a musl one, and that for each architecture)

Supersedes #6475.

---------------------------

What this does is embed the entirety of share/ - all the functions, completions and fish_config, and then write it out when you run `fish --install` (or `fish --install=noconfirm` to skip the confirmation).

It writes them out to ~/.local/share/fish/install, so uninstallation is `cargo uninstall fish; rm -r ~/.local/share/fish/install`.

Technically, this is done by adding a Cargo feature, "installable", that's enabled by default (so `cargo install` will do it by default). When it is enabled, rust_embed will be added as a dependency. Default features are disabled in the cmake build so nothing changes there. (a slight wrinkle is that, if we did add another feature, we would have to turn it *on* in cmake)

This about doubles the fish binary size (6M to 12M unstripped, 4M to 8.7M stripped). I have run benchmarks where it had little impact (7% increase in `--no-config` startup time). I do not think this is important. Binaries of similar size on my system: ctest, emacs, agg, gdb, perf. Bigger binaries include: kitten, clang-tidy, qemu, clangd. I have never thought of the size of any of them.

# TODO:

- [x] Documentation
- [ ] Tests
- [x] The version currently only includes the commit hash (e.g. "4ec25d894"). I believe that's because cargo only does a shallow clone. This would have to find another source, e.g. the crate version, but even then it would look different because the number of commits is lost. Would something like "3.7.1-g4ec25d894" be an okay version string?

The following are more "future directions" than things I feel must be included now for this to be useful:

- [ ] Stop including __fish_build_paths.fish.in (this is harmless)
- [x] Include man pages - these aren't currently included. It might be possible to get them in by generating them in build.rs.
  HTML isn't possible (it needs fish_indent to do the syntax highlighting), but `help` is always available and will simply open the online version.
- [ ] Include translations
- [x] Make fish_indent/fish_key_reader available as builtins (and potentially by calling fish *as* them) so you really have a single-file fish - #10876
- [ ] The data path is currently pretty set as ~/.local/share/fish/install. It would be nice to make this "relocatable" in some manner, so people can download a build and place it in e.g. /usr/local/bin/ or /opt/. (unfortunately cargo installs to ~/.cargo/bin/ by default but ~/.cargo/share/ isn't *really* a thing, so a "../share" feels off to me?)

Edit: The man pages are now included, fish_indent/key_reader is #10876.

The data path is computed at runtime in $HOME. So you *can* place this in /opt, but every user running it would have to --install separately. I believe that's fine for now, a future possibility is to install unconditionally.

The version is now CRATE_VERSION-gHASH. This could throw off some regexen because it doesn't match the VERSION-DISTANCE-gHASH format. We could add a fake DISTANCE, e.g. 9999, given that you can't rely on "number of commits" for anything anyway.


faho:
One annoyance of this solution: You have to rerun `fish --install` on upgrade. I experimented with having it check the mtime of share/config.fish and printing a note if it was older than the fish binary. This generally works, but will trigger too often (e.g. if you move the fish binary after installing).

It might be possible to write a file with the version that was installed and check that on startup instead.

matklad:
>I believe it would be possible to allow this as an option, and to add a static build to fishshell.com so people can just download a fish, without compilation.

Would be prudent to double-check performance of the statically linked version. Four-ish years ago, statically linking with musl could have tanked your performance due to musl's `malloc` being much slower than glibc one or jemalloc. This issue prevented rust-analyzer from going the fully statically linked way. But I remeber that musl wanted to rewrite the allocator, and, presumably, they had done that since.

dingobits:
> The alternative is to offer a self-installable build dynamically linked against glibc, but that obviously requires a compatible (i.e. same version or newer) glibc to what it was built against. This would _still_ be better than the appimage, because that needs libfuse2 _and_ glibc on the host. These are also simple enough to build that we could do both.

Binaries built with newer glibc isn't necessarily incompatible with older glibc. glibc uses symbol versioning to indicate compatibility. I pulled your branch and ran `cargo build --release` with glibc 2.37, which was released in 2023. `readelf -V target/release/fish` shows binary requires `GLIBC_2.34`, which was released in 2021. I don't know how many years of backwards compatibility is most appropriate, but I think building from Debian oldstable or the Ubuntu equivalent should suffice for most people.

Justinzobel:
So if I had a fleet of 100 servers on varying versions of Ubuntu LTS I could build it on the oldest server and it should be able to be copied to any of the same version server or higher and it would simply run?

faho:
> I could build it on the oldest server and it should be able to be copied to any of the same version server or higher and it would simply run

If you build dynamically against glibc, yes. If you build statically against musl you can build it *anywhere* (probably even cross-compile it from macos), at the cost of some performance.

This will give you a per-user fish that needs to install its datafiles by running `fish --install`, after that `fish` will be fully capable.

faho:
Alright, after a rebase this now crashes on startup - do `cargo install --path .`, run ~/.cargo/bin/fish, and you'll get a segfault.

Backtrace is:

```
#0  0x0000555555801744 in fish::fork_exec::blocked_signals_for_job ()
#1  0x00005555557ef2ee in fish::fork_exec::spawn::PosixSpawner::new ()
#2  0x00005555557ea0fc in fish::exec::exec_process_in_job ()
#3  0x00005555557e4bdf in fish::exec::exec_job ()
#4  0x00005555556804ea in fish::parse_execution::ParseExecutionContext::run_1_job ()
#5  0x000055555567f378 in fish::parse_execution::ParseExecutionContext::run_job_conjunction ()
#6  0x000055555567f20a in fish::parse_execution::ParseExecutionContext::run_job_list ()
#7  0x00005555556776a4 in fish::parse_execution::ParseExecutionContext::eval_node ()
#8  0x000055555576fd4c in fish::parser::Parser::eval_node ()
#9  0x000055555576d7fe in fish::parser::Parser::eval_with ()
#10 0x00005555557f6ba9 in fish::exec::exec_subshell_internal ()
#11 0x000055555572d2f3 in fish::expand::expand_cmdsubst ()
```

krobelus:
debug build might help. This is still pre-fork so that should rule out a class of problems

faho:
> debug build might help. 

Debug build gives an empty backtrace (only `??`, no symbols).

faho:
Okay I'm reasonably sure `cargo install` pulled in a newer libc. Pinning it to 0.2.151 (the last version compatible with rust < 1.71) seems to work.

zanchey:
The thing that bites people when using `make install` into their homedir for projects is that it does not delete files that are no longer installed. (This is why I worked on the development build packages - the only way to reliably manage your filesystem is with a package manager that uninstalls old files. `make uninstall` is not supported in CMake and in autotools it required a lot of care on the part of developers, which we did not always take.) `cargo install` will manage the binaries ok, but the `fish --install` step should possibly install to a versioned directory or similar.

faho:
> fish --install step should possibly install to a versioned directory or similar.

It currently installs into ~/.local/share/fish/install, so there's the binaries in the cargo binary dir and exactly one directory.

That means uninstallation is `cargo uninstall fish; rm -rf ~/.local/share/fish/install`, which I feel is simple enough. (or if you want to entirely stop using fish you can just `rm -rf ~/.local/share/fish`)

A versioned directory could lead to multiple of these being left behind instead of simply overwritten on upgrade.

(plus the biggest issue left here is that installing via git like `cargo install --git https://github.com/faho/fish-shell --branch fish-installer` will not give you a version because cargo won't pull the tags, so you only get the commit hash)

zanchey:
> A versioned directory could lead to multiple of these being left behind instead of simply overwritten on upgrade.

That's fine, but in that case the destination directory should be removed before the extraction starts - it doesn't look like this happens at present. Otherwise, files that are present in one version but not the next will be left behind.

faho:
> That's fine, but in that case the destination directory should be removed before the extraction starts

Good point, done.

faho:
Alright, those last two commits are a bit weird.

The immediate issue is that `cargo install --git https://github.com/faho/fish-shell/ --branch fish-installer-test` won't clone the tags. So if you do that *right now*, you'll get something in $FISH_VERSION (and `fish --version`) like 

```
fish, version f3fc743fc
```

which will break a lot of uses of $FISH_VERSION - there is no *version* there, only the commit sha.

So this detects that case and turns it into

```
fish, version 3.7.95-g5b00071a7
```

instead, by using the crate version (in Cargo.toml).

I'm unsure if that is also needed for packages on crates.io.

I changed the crate version (which was still at 0.1.0!) for testing. The version I picked is arbitrary and preliminary, and I am fine with removing it again. Tho I suppose it's not harmful to keep it in?

----

I would like to try to get some more people to test this - and therefore the next fish version.

We still have more bugs remaining than I would like, especially in corners we don't visit as often and this is really easy to get going as a *separate* version of fish without breaking your main install.

It would also show if people like this way of installing it, and if anyone has an idea for getting the install/upgrade experience working.

zanchey:
Not sure how much work it would be but getting some form of CI running would be good.

We could publish the crates as [pre-release versions](https://stackoverflow.com/a/46373403/125549). I think it makes sense to do that with a tagged tarball as well?

faho:
Okay, I'm not going to be doing much work on fish for the foreseeable future for personal reasons.

I would be glad if someone were to take this over.

As far as I'm concerned, there are two things that need to be done before this can be let loose on users:

- The data directory should be using $HOME at runtime, currently it's compiled in
- It should detect if the data is out of date and suggest you rerun `fish --install` (possibly by writing out a file with the version)

The former means it can't really be used in a "build on one system, copy elsewhere" manner (including as a static binary on fishshell.com). If I build on my system where $HOME is /home/alfa, I can't use it as root or as another user on another system (not even reinstalling it).

The latter means it's too easy to let the fish binary and the data files get out of sync.

faho:
Alright, got it using the home directory at runtime. The code is pretty ugly but it appears to work for now.

faho:
I have managed to get Github Actions to create static builds - https://github.com/faho/fish-shell/commit/8afa27b7a71bf8d88d5325d7a2d7c29918745f88.

This will lead to something like https://github.com/faho/fish-shell/actions/runs/10321626509, where the "Artifacts" contains a fish.zip (unfortunately) that contains a fish.tar.xz.

This should run on basically any x86_64 linux by just extracting it and running `/path/to/fish --install` and then `/path/to/fish`.

So if you would like to test this, that's an easy way to do it.

(I'm sure there's *some* minimum kernel version, possibly 3.2, which is the minimum for rust since 1.64. It was released in 2012)

faho:
So. This works and it's pretty easy to use. I believe it is a convenient way to install fish. It is also easy to generate builds, including nightly, which seems useful for debugging.

I would now like to have some feedback. 

- Do you want to use this?
- Is this a version we should support?
- Should we add some builds to fishshell.com (I would recommend x86_64 and aarch64 linux, statically linked with musl)?

tangowithfoxtrot:
Thank you so much for adding this!

> Do you want to use this?

Absolutely! This is great for distrobox containers, or similar not-quite-ephemeral environments where interactive shell use is needed. In such environments, I'd default to installing using this method so I don't have to worry about getting older builds from the distro's package manager.

> Is this a version we should support?

I can only speak as a user of Fish, but I'd definitely like to see it supported in the sense that I'd like it to be reliable when installing any tagged release version of Fish.

> Should we add some builds to fishshell.com (I would recommend x86_64 and aarch64 linux, statically linked with musl)?

Making static bins available for Linux would be very nice, especially for systems/containers that don't have `cargo` installed. I assume the installation process would just be something like

```sh
# curl it instead of needing cargo
curl -LO https://github.com/fish-shell/fish-shell/releases/download/{version}/fish-{version}-musl-{arch}.tar.xz
tar xf fish-*.xz
mv ./path-to-fish /usr/local/bin
fish --install
```

then? If so, that's pretty much ideal for me.

Thanks again for making Fish easier to install!

faho:
> I assume the installation process would just be something like
> `mv ./path-to-fish /usr/local/bin`

Not quite.

This is a per-user installation. It is made to be used in situations where you don't *have* root and therefore can't write to /usr/local, and without having to keep it in a specific place. (if you have root, you could also just install the package, which would give you automatic updates and no `--install` step)

So `fish --install` will write to ~/.local/share/fish/install *of the invoking user*, and running `fish` will look in that user's ~/.local/share/fish/install.

So while you *can* place it in /usr/local/bin, that 1. involves being root, 2. means every user who runs it will have to run `fish --install` before they can use it. That includes if you do `sudo /path/to/fish`, which requires `sudo /path/to/fish --install` first.

I would imagine ~/bin/ or similar is usually a better idea to avoid confusing people.

tangowithfoxtrot:
That makes sense. Thanks for clarifying.

In that case, I prefer `~/.local/bin` for user installations, but that's just a minor detail since it can be installed this way now 😄

krobelus:
I haven't taken a look at the code but if we the bandwidth to support this, this sounds like a nice feature.
I used to not know how to install fish without root rights, so with the right marketing that would have helped.
Long-term I'd prefer if we don't advertise multiple different ways of doing the same thing; if this removes a reason to use cmake, that's great

Today, `make install` has the following problems
1. There's no uninstall, which leads to weird issues when a new release removes a completions script.
2. To support `vendor_completions.d` and friends it requires manual configuration (they default to PREFIX but the actual vendor_completions.d is probably located in `/usr/share`).
I use this hack to work around it:

```fish
set -a fish_complete_path /usr/share/fish/vendor_completions.d
set -a fish_function_path /usr/share/fish/vendor_functions.d/
```

the installer solves 1 but I'm not sure if we can solve 2. It's a rabbit hole..

faho:
Alright, now it'll write out its version on install, and check it on start so it can warn, so people don't forget to run `fish --install`.

If the version is wrong, it will still start and read the config.

The assumption is that this would mostly be rare updates from, say, 4.0 to 4.1, and they'll be 99% compatible.

We could also automate this - just write out the new files, or ask about that and then do it, but I'm wary of doing that.

rafi:
This is awesome. Works great on Debian 10+glibc 2.28 and Ubuntu 24.04+glibc 2.39, most useful! ([ref](https://github.com/rafi/awesome-cli-binaries/blob/master/fish.Dockerfile))

faho:
Alright, I merged a few commits of this that were good independent of a self-installable fish.

My intention here is to merge this in a week or so unless anyone objects.

I believe we should offer static builds on fishshell.com for releases and nightlies, for x86_64 and aarch64 linux (macOS has the .app and the .pkg).

The future directions remain:

- Incorporate translations - but that might go easier if we throw out gettext and replace it with something better, which we should do anyway
- Incorporate help text, which needs man pages currently (but we could also write it separately because the full help is a bit much there)
- Make fish_indent/fish_key_reader available as builtins so you get one file instead of three (and if you call fish via a "fish_indent" symlink it becomes fish_indent, so you can use symlinks to call it from the outside)
- Possibly ask to install on startup
- Make functions readable directly from the binary, removing the need to install it for those (you would still need to install for the docs, fish_config, etc)

krobelus:
This is amazing, a great leap forward.
I think everyone agrees that one should be using their package manager for most of their software.
But especially for people who want to track the latest development version[^*], this is a great improvement -- it's more obvious how to install (also mostly solves #10799) and much easier to uninstall. I expect that this will enable some people to contribute.

Looks like I no longer need cmake or make in my install script; I changed

    mkdir -p build && cd build && cmake .. -GNinja -DCMAKE_INSTALL_PREFIX=$HOME/.local && ninja install

to this (most users probably don't want `--debug`)

    cargo install --path . --debug && echo yes | fish --install

---

The main thing that I would like to see fixed is man pages (all of `man status`, `status -h`, `status --help`).
I'm not sure I'd be comfortable merging without knowing a safe path to fix that.
But it should be doable, right? We can run `sphinx` from build.rs...

---

We can (in future) probably simplify the UI by removing the `--install` flag and just always install without asking, probably to `~/.cache/fish/` instead of `~/.local/share/fish/`.
In future the cache path would contain a hash of the assets, so one can even run multiple different versions concurrently without manually choosing a data-dir-prefix.
Add a simple form of garbage collection and the user won't ever notice. We can still have a predictable path by creating a symlink to the freshest fish's assets.

That should make it relocatable enough, so it should still be easy to install to `/usr/local/bin`.

---

> I believe we should offer static builds on fishshell.com for releases and nightlies, for
> x86_64 and aarch64 linux (macOS has the .app and the .pkg).

That would be nice. We can link to them from the README

> Make fish_indent/fish_key_reader available as builtins so you get one file instead of three

Good idea. Being able to simply add a symlink to `some/random/build/dir/fish` to PATH and have it just work would be perfect.
That is my favorite way to install software to that I might hack on myself, because it's very easy to undo. Also for reproducing bugs on latest versions.

[^*]: somehow my Ubuntu system at work literally has `fish 3.1.0`; it would be pretty rough to run that

.github/workflows/staticbuild.yml
```diff
@@ -0,0 +1,46 @@
...
+        fetch-depth: 0
+    - name: Prepare
+      run: |
+           rustup target add x86_64-unknown-linux-musl
+           rustup target add aarch64-unknown-linux-musl
+           sudo apt install musl-tools crossbuild-essential-arm64 -y
+    - name: Build
+      run: |
+           CMAKE_WITH_GETTEXT=0 CC=aarch64-linux-gnu-gcc RUSTFLAGS="-C linker=aarch64-linux-gnu-gcc -C link-arg=-lgcc" cargo build -r --target aarch64-unknown-linux-musl
+           cargo build -r --target x86_64-unknown-linux-musl
```
krobelus, line 35: better to use long options like `--release` instead of `-r` here for readability

build.rs
```diff
@@ -223,22 +223,43 @@ fn setup_paths() {
...
+
     let bindir = get_path("BINDIR", "bin/", prefix.clone());
     rsconf::set_env_value("BINDIR", bindir.to_str().unwrap());
     rsconf::rebuild_if_env_changed("BINDIR");
 
-    let sysconfdir = get_path("SYSCONFDIR", "etc/", datadir.clone());
+    let sysconfdir = get_path(
+        "SYSCONFDIR",
+        // If we get our prefix from $HOME, we should use the system's /etc/
+        // ~/.local/share/etc/ makes no sense
```
krobelus, line 259: TIL I have a bunch of things installed to `~/.local/share/etc/`..
I agree it never makes sense, and redirecting to `/etc` would make those installs strictly less broken.

src/bin/fish.rs
```diff
@@ -71,9 +71,99 @@ use std::{env, ops::ControlFlow};
...
+        std::process::exit(1);
+    };
+    let dir = PathBuf::from(home).join(DATA_DIR).join(DATA_DIR_SUBDIR);
+
+    // TODO: Translation,
+    // FLOG?
+    // - Install: Translations
+    // - Install: Manpages (build via build.rs)
+    // - Don't install: __fish_build_paths.fish.in
+    if !noconfirm {
```
krobelus, line 102: double negation is awkward, I'd maybe change it to `fn install(confirm: bool)`.

src/bin/fish.rs
```diff
@@ -71,9 +71,99 @@
...
+             Please enter 'yes' to continue.",
+            dir.display()
+        );
+        eprint!("> ");
+        let _ = stderr().flush();
+
+        let mut input = String::new();
+        stdin()
+            .read_line(&mut input)
+            .expect("Failed to read from stdin");
```
krobelus, line 114: should not crash here but print the error and `process::exit(1)`.
On some terminals it's easy to enter invalid UTF-8 by accident etc.
(We changed our panic handler to create coredumps so preferably only internal errors should crash)

src/bin/fish.rs
```diff
@@ -71,9 +71,99 @@
...
+
+        if input != "yes\n" {
+            eprintln!("Exiting without writing any files");
+            std::process::exit(1);
```
krobelus, line 118: absolute nit but when I use this noninteractive version, I get an abandoned
line (fish would fix it after printing ⏎ but bash doesn't)

	[johannes@localhost fish-shell]$ echo yes | fish --install
	This will write fish's data files to '/home/johannes/.local/share/fish/install'.
	Please enter 'yes' to continue.
	> [johannes@localhost fish-shell]$ 

I'm not yet sure what's the easiest fix (I guess we could reuse our normal
reader loop..).  I guess that's one upside of `fish --install=noconfirm`
in addition to making it more obvious to everyone how to do noninteractive installs.

src/bin/fish.rs
```diff
@@ -71,9 +71,99 @@
...
+        };
+        let res = File::create(&path);
+        let Ok(mut f) = res else {
+            eprintln!("Creating file '{}' failed", path.display());
+            continue;
+        };
+        // This should be impossible.
+        let d = Asset::get(&file).expect("File was somehow not included???");
+        f.write_all(&d.data).expect("FAILED TO WRITE");
```
krobelus, line 148: nit: should also do a non-panicking error check for `write_all()`,
in case we run out of disk space or something

src/bin/fish.rs
```diff
@@ -261,7 +351,57 @@
...
+        let (_, Some(home)) = fish::env::get_user_home() else {
+            FLOG!(
+                error,
+                "Cannot find home directory and will refuse to read configuration"
+            );
+            return;
+        };
+
+        let p = PathBuf::from(home).join(paths.data.clone());
+        let Some(homestr) = p.to_str() else {
```
krobelus, line 364: we can avoid requiring UTF-8 here (like in the `else` branch)

```diff
diff --git a/src/bin/fish.rs b/src/bin/fish.rs
index 5bede0d43..3b5d68cda 100644
--- a/src/bin/fish.rs
+++ b/src/bin/fish.rs
@@ -361,14 +361,7 @@ fn read_init(parser: &Parser, paths: &ConfigPaths) {
         };
 
         let p = PathBuf::from(home).join(paths.data.clone());
-        let Some(homestr) = p.to_str() else {
-            FLOG!(
-                error,
-                "Cannot decode home directory and will refuse to read configuration"
-            );
-            return;
-        };
-        str2wcstring(homestr.as_bytes())
+        str2wcstring(p.as_os_str().as_bytes())
     } else {
         str2wcstring(paths.data.as_os_str().as_bytes())
     };
```

src/bin/fish.rs
```diff
@@ -261,7 +351,57 @@
...
+        // When fish is installable, we write the version to a file,
+        // now we check it.
+        let verfile =
+            PathBuf::from(fish::common::wcs2osstring(&datapath)).join("fish-install-version");
+        let Ok(version) = std::fs::read_to_string(verfile) else {
+            let escaped_pathname = escape(&datapath);
+            FLOGF!(
+                error,
+                "Fish cannot find its asset files in '%ls'.\n\
+                 Refusing to read configuration because of this.",
```
krobelus, line 387: should probably also log the underlying error, like

```rust
let version = match read_to-string(verfile) {
	Ok(version) = version,
	Err(err) => {
		FLOG!(error, "fish cannot read version file from installed asset files:", err);
	}
};
```

src/bin/fish.rs
```diff
@@ -366,6 +513,21 @@
...
+            'I' => {
+                let noconfirm = match w.woptarg {
+                    None => false,
+                    Some(n) if n == L!("noconfirm") => true,
+                    _ => {
+                        FLOGF!(
+                            error,
+                            "Unknown argument to --install: '%ls'",
+                            w.woptarg.unwrap()
+                        );
```
krobelus, line 525: I believe we should avoid `FLOG!` flog for user errors that can realistically happen (for consistency and also translations).
So I'd probably use `eprintf!` or `eprintln!` here

src/env/environment.rs
```diff
@@ -452,6 +454,41 @@ fn get_hostname_identifier() -> Option<WString> {
     }
 }
 
+/// Get values for $USER and $HOME via getpwuid,
+/// without trusting $USER or $HOME.
+pub fn get_user_home() -> (Option<String>, Option<String>) {
```
krobelus, line 459: The "user" return value is unused, so it's better to remove it.

Again, we can avoid requiring UTF-8 here (unlike much of the Rust ecosystem)
by returning a `Vec<u8>`, which is suitable for consumption by `str2wcstring`

src/env/environment.rs
```diff
@@ -595,12 +632,33 @@
...
     if let Some(paths) = paths {
-        vars.set_one(
-            FISH_DATADIR_VAR,
-            EnvMode::GLOBAL,
-            str2wcstring(paths.data.as_os_str().as_bytes()),
-        );
+        let datadir = if cfg!(feature = "installable") {
+            if let Some(home) = vars.get(L!("HOME")) {
+                if let Ok(strhome) = std::str::from_utf8(&wcs2string(&home.as_string())) {
+                    let dir = PathBuf::from(strhome)
```
krobelus, line 644: I guess since we're on Unix we could also hardcode the slashes.
But going via PathBuf seems better. We have a convenience method for creating an `OsStr` which imlements `AsRef<Path>` which is required by `PathBuf::from`:

```diff
diff --git a/src/env/environment.rs b/src/env/environment.rs
index 041fc926b..00d49250d 100644
--- a/src/env/environment.rs
+++ b/src/env/environment.rs
@@ -642,12 +642,7 @@ pub fn env_init(paths: Option<&ConfigPaths>, do_uvars: bool, default_paths: bool
             if let Some(home) = vars.get(L!("HOME")) {
-                if let Ok(strhome) = std::str::from_utf8(&wcs2string(&home.as_string())) {
-                    let dir = PathBuf::from(strhome)
-                        .join(env!("DATADIR"))
-                        .join(env!("DATADIR_SUBDIR"));
-                    str2wcstring(dir.as_os_str().as_bytes())
-                } else {
-                    // This should be impossible given we already read_init.
-                    FLOG!(error, "Cannot decode $HOME");
-                    "".into()
-                }
+                let strhome = wcs2osstring(&home.as_string());
+                let dir = PathBuf::from(strhome)
+                    .join(env!("DATADIR"))
+                    .join(env!("DATADIR_SUBDIR"));
+                str2wcstring(dir.as_os_str().as_bytes())
             } else {
```

src/env/environment.rs
```diff
@@ -618,21 +676,13 @@
...
-            let mut scstr = paths.data.clone();
-            scstr.push("functions");
-            vars.set_one(
-                L!("fish_function_path"),
-                EnvMode::GLOBAL,
-                str2wcstring(scstr.as_os_str().as_bytes()),
-            );
+            // TODO: Trailing slash?
+            let mut scstr = datadir;
+            scstr.push_str("/functions");
```
krobelus, line 681: Probably we control the datadir so there should already be no trailing slash here.
If not, we should "sanitize" the datadir ASAP wherever it is passed in by the user

src/bin/fish.rs
krobelus: also this disallows running `target/debug/fish`. We should probably skip the version check if we're colocated like that,
because `share/` has the right version by definition

jamesamcl:
> This is amazing, a great leap forward.
> I think everyone agrees that one should be using their package manager for most of their software.
> But especially for people who want to track the latest development version[1](#user-content-fn-*-06abfb9ec00174bcabb02a1225458394), this is a great improvement

Just wanted to comment that this is not the only use case. I have been following this issue for some time because I am a user of a shared computing environment (HPC) where users do not have root access and I would love to be able to use fish.

krobelus:
> do not have root access and I would love to be able to use fish.

you can also do that without the self-installation mechanism,
`make PREFIX=$HOME/.local install` and add `~/.local/bin` to your `PATH`.
Currently this build step needs cmake which you can download as a static binary from cmake.org.

src/bin/fish.rs
krobelus: when running directly as `target/*/fish`, we should use the asset files in the source tree but we currently can't build them without cmake

src/bin/fish.rs
faho: I disagree - FLOG! is how we print errors all over the place.

See e.g. all the history errors, which can and do happen.

The only reason I'm using eprintln et al at all here is because they're used before initialization is done.

src/bin/fish.rs
faho: I made it skip the prompt if stdin isn't a tty.

krobelus:
I'd squash the fixup commits, they only help a hypothetical person
who has context of a previous snapshot.

faho:
Of course, I'll do that before merging.

I've been working on getting the man pages in, but I haven't been able to spend too much time on it.

faho:
Okay, got a pass of the man pages in.

I noticed another blocking issue: The $__fish_X_dir variables are borken, because of where I put in the code to find the data dir. That should go another place (determine_config_directory_paths).

But that throws up another question: What is $__fish_bin_dir supposed to be here? It should probably go to the directory that `status fish-path` shows, but we cannot get that everywhere.

---------

Longer term, it would probably be cool to incorporate the usage text directly in the builtin, generated at build time. `__fish_print_help` is terrifying.

(we could make it just the synopsis, which has been asked for before, and would allow us to skip piping to a pager here)

krobelus:
On Mon, Nov 18, 2024 at 01:06:23PM -0800, Fabian Boehm wrote:
> But that throws up another question: What is $__fish_bin_dir supposed to be here?
> It should probably go to the directory that `status fish-path` shows, but we cannot get that everywhere.

I think it's mainly OpenBSD that doesn't support getting the executable path.
Could add an option for this case (and error out if it's missing). Or use `PATH`. It's not used very much.

faho:
Okay, $__fish_bin_dir is now empty if unavailable and $__fish_help_dir is always empty in installable mode (because the html files can't be built before fish_indent exists - note that "paths.doc" is a misnomer, it's only the html docs, not the man pages).

Our only use of $__fish_bin_dir is in fish_config, and that already tries `status fish-path` and falls back on $PATH otherwise.

Note that this *can* break if you install an installable fish and don't put it in $PATH, on OpenBSD.

The code could use some cleanup and there are some small nits. E.g. build.rs isn't rerun if only the docs are changed.

rafi:
Started getting errors when building:

```
20.80   Installing fish v4.0.0-alpha1 (https://github.com/faho/fish-shell?branch=fish-installer#eb2f936f)
27.48 warning: fish@4.0.0-alpha1: Could not build man pages: Os { code: 2, kind: NotFound, message: "No such file or directory" }
49.79 error: expected identifier, found `#`
49.79   --> src/bin/fish.rs:41:5
49.79    |
49.79 41 |     #[allow(unused_imports)]
49.79    |     ^ expected identifier
```

Reproduce `Dockerfile`:
```
FROM debian:buster AS fish-builder

ARG DEBIAN_FRONTEND=noninteractive
RUN apt-get update \
    && apt-get upgrade --yes --show-upgraded \
    && apt-get install --yes build-essential curl \
    && apt-get purge --yes manpages manpages-dev \
    && apt-get autoremove --yes \
    && rm -rf /var/lib/apt/lists/*

SHELL ["/bin/bash", "-o", "pipefail", "-c"]
RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y

WORKDIR /root
RUN . .cargo/env && cargo install --git https://github.com/faho/fish-shell --branch fish-installer
```

faho:
@rafi https://github.com/fish-shell/fish-shell/pull/10367/commits/ea2fa25d4a197f1c6d731e0797e1297a6125974d

rafi:
@faho works great now. Looks like GitHub actions passed as-well. Thanks!

By the way, noticed the `2.0.0` tag on this fork diverged from the `2.0.0` tag on upstream.

faho:
> Looks like GitHub actions passed as-well

Note that it passes, but did not build the man pages correctly. I suspect some sphinx version difference.

> By the way, noticed the 2.0.0 tag on this fork diverged from the 2.0.0 tag on upstream.

https://github.com/faho/fish-shell/tree/2.0.0 and https://github.com/fish-shell/fish-shell/tree/2.0.0 both show 84978b71329500fd00087d1db64cab95803f3bc3 for me.

faho:
Okay, the [staticbuilds](https://github.com/faho/fish-shell/actions/workflows/staticbuild.yml) now include the man pages.

The issue was that rsconf::set_env_value does not set the variable for child processes, so `sphinx-build` would try to invoke `fish_indent --version`, and `fish_indent` wasn't built yet (and no previous install was available).

(this was pretty annoying to debug because github actions is pretty intransparent and I don't know of a good way to just get all output from a child process via build.rs)

VuiMuich:
For the `version` TODO as a reference, `leftwm` [does this](https://github.com/leftwm/leftwm/blob/main/leftwm/src/bin/leftwm-check.rs#L33):
```rust
    println!(
        "\x1b[0;94m::\x1b[0m LeftWM version: {}",
        env!("CARGO_PKG_VERSION")
    );
    println!(
        "\x1b[0;94m::\x1b[0m LeftWM git hash: {}",
        option_env!("GIT_HASH").unwrap_or(git_version::git_version!(fallback = "unknown"))
```
---
Looking so much forward when this PR gets merged :heart_eyes:

src/env/environment.rs
faho: This comes from pathbuf.join everywhere so it's fine

faho:
Alright, no use in waiting any longer on this, I'm pressing the big blue button (possibly green for y'all, I'm reasonably sure I have deuteranopia mode turned on).

krobelus:
did you manage to fix `HOME=$(mktemp -d) fish` and man pages?

faho:
> did you manage to fix `HOME=$(mktemp -d) fish` and man pages?

I did add the man pages, there's no way to fix `HOME=$(mktemp -d)` with this.

krobelus:
On Fri, Dec 06, 2024 at 01:17:20PM -0800, Fabian Boehm wrote:
> > did you manage to fix `HOME=$(mktemp -d) fish` and man pages?
> 
> there's no way to fix `HOME=$(mktemp -d)` with this.
> 
> That's what `--no-config` is for.

it's not the same.
I think a fixed location is the wrong approach, among other things it makes it unnecessarily hard to install concurrent versions

faho:
> I think a fixed location is the wrong approach

I await your PR. The big problem is that `../share` is a terrible directory to use.

Frankly I think our effort would be better invested loading the functions straight from the binary without unpacking, as well as incorporating the help text (that latter one also in non-installable builds).

krobelus:
> The big problem is that ../share is a terrible directory to use.

Haven't tried it yet but I think everything would go into ~/.cache/fish, not sure where `../share` comes into play.

Anyway, with the current approach, our stance should be that users should not rely on the default location of the support files, because we might change that in future. Probably this is more or less obvious.

> Frankly I think our effort would be better invested loading the functions straight from the binary without unpacking, as well as incorporating the help text (that latter one also in non-installable builds).

yes that would be better (with the same interface). Having the raw files available in `share/functions` is nice but probably not a hard requirement. One can always look at the source

faho:
> Haven't tried it yet but I think everything would go into ~/.cache/fish, not sure where ../share comes into play.

Whether it goes into ~/.cache/fish or ~/.local/share/fish isn't really important and doesn't solve your question of installing concurrent versions - it's still a fixed location.

You *could* of course make versioned directories (in either path), but that means you need to figure out how to clean old directories or you're gonna leave them around.

If instead the functions, completions and documentation are read from the binary, you'd only have fish_config and themes there, and those are just much less important to keep up to date.

(the trickiest part is the man pages, which we would have to write out to `man` for formatting)

lukasbindreiter:
@faho - I just tried downloading the last `staticbuild` artifact for linux `aarch64` - however the artifact has already expired - and since the build is only triggered manually there is no newer run. Could you maybe trigger a new one? And any chance the artifact lifetime could be increased a bit for that workflow?

faho:
@lukasbindreiter The [4.0b1 release](https://github.com/fish-shell/fish-shell/releases/tag/4.0b1) contains copied artifacts at the bottom.

I've now triggered [a new run for Integration_4.0.0](https://github.com/fish-shell/fish-shell/actions/runs/13028212308).

> And any chance the artifact lifetime could be increased a bit for that workflow?

We don't want to keep around arbitrary checkpoints for long. We're probably gonna keep the releases.

You can also always build it yourself with `cargo build --git https://github.com/fish-shell/fish-shell.git --target aarch64-unknown-linux-musl` (this requires a musl C toolchain, I've had some success with [cargo-zigbuild](https://github.com/rust-cross/cargo-zigbuild) for targets that zig supports). If you don't need it static, `cargo install` just works.

lukasbindreiter:
@faho Thank you very much, this resolved my issue! 🙏 

Makes sense that you don't want to keep them around for too long - I wasn't aware they were part of the release and didn't check there unfortunately - thanks for the pointer.

PR #10368, opened 2024-03-12
Add completions for dmidecode

Xiretza:


share/completions/dmidecode.fish
```diff
@@ -0,0 +1,22 @@
+function __fish_dmidecode_complete_keywords
+	dmidecode -s 2>&1 1>/dev/null | tail -n +4 | string trim
```
faho, line 2: I'm always a bit wary of magic line number counts.

I think I would turn this into

```fish
dmidecode -s 2>&1 >/dev/null | string match -rg '^\s+(.*)'
```

instead - match lines starting with whitespace instead of skipping exactly 4 lines.

Xiretza: Good idea, done.

Xiretza:
Any idea what the CI failures are about?

faho:
Nothing to do with this, in any case.

I'm pretty sure fish_test_helper returns false, and so the test exits.

(this is masked by our pexpect setup trying to read a signal value and getting a None instead).

PR #10371, opened 2024-03-15
completions/git: don't offer files when running `git commit --fixup`

sanga:
## Description

So I'm running fish 3.7 when I see this behaviour but it looks like the same problem should exist in master. Basically, when I run `git commit --fixup` I'm expecting to only see the recent commits, however before this change, I'd also see completions for untracked files in the repo. This should fix that (it seems to for 3.7 at least). I guess this isn't worth mentioning in the changelog?


krobelus:
`git commit --fixup HEAD README.rst` is valid though so we should only suppress filenames if the last token is fixup/squash.
I'll commit a hacky fix, in future we should use argparse or perhaps ask Git itself

PR #10373, opened 2024-03-15
ast: Replace can_parse with static dispatch

The0x539:
## Description
My original goal was to get rid of the `None` variants on `StatementVariant` and `BlockStatementHeaderVariant`, but I discovered that that would require a larger refactor.
This change stands on its own, but doubles as a small step towards that refactor by reducing our reliance on "uninitialized" node values.

No intended behavior changes, but I wouldn't be surprised by a minor performance boost.

## TODOs:

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -446,6 +446,12 @@ pub trait List: Node {
     }
 }
 
+/// This is for optional values and for lists.
+trait CheckParse {
+    /// A true return means we should descend into the production, false means stop.
+    fn can_be_parsed(pop: &mut Populator<'_>) -> bool;
+}
+
 /// Implement the node trait.
 macro_rules! implement_node {
     (
@@ -520,7 +526,7 @@ macro_rules! implement_leaf {
 
 /// Define a node that implements the keyword trait.
 macro_rules! define_keyword_node {
-    ( $name:ident, $($allowed:expr),* $(,)? ) => {
+    ( $name:ident, $($allowed:ident),* $(,)? ) => {
         #[derive(Default, Debug)]
         pub struct $name {
             parent: Option<*const dyn Node>,
@@ -553,7 +559,7 @@ macro_rules! define_keyword_node {
                 &mut self.keyword
             }
             fn allowed_keywords(&self) -> &'static [ParseKeyword] {
-                &[$($allowed),*]
+                &[$(ParseKeyword::$allowed),*]
             }
         }
     }
@@ -561,7 +567,7 @@ macro_rules! define_keyword_node {
 
 /// Define a node that implements the token trait.
 macro_rules! define_token_node {
-    ( $name:ident, $($allowed:expr),* $(,)? ) => {
+    ( $name:ident, $($allowed:ident),* $(,)? ) => {
         #[derive(Default, Debug)]
         pub struct $name {
             parent: Option<*const dyn Node>,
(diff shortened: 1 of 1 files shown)


The0x539:
I have no idea why `make test / ubuntu-asan` keeps failing.

Xiretza:
Known issue, nothing to do with your PR.

src/ast.rs
```diff
@@ -446,6 +446,16 @@ pub trait List: Node {
     }
 }
 
+/// This is for optional values and for lists.
+trait CheckParse {
+    /// A true return means we should descend into the production, false means stop.
+    fn can_be_parsed(pop: &mut Populator<'_>) -> bool;
+}
```
krobelus, line 453: Really nice work.

Getting rid of virtual dispatch seems like a good direction, a big one would be  https://github.com/fish-shell/fish-shell/pull/9636#issuecomment-1890470676

src/ast.rs
```diff
@@ -446,6 +446,16 @@
...
+
+trait StaticTokenInfo {
```
krobelus, line 455: is it usual to prefix type-only traits with `Static`?

src/ast.rs
```diff
@@ -446,6 +446,16 @@
...
+    const ALLOWED_TOKENS: &'static [ParseTokenType];
```
krobelus, line 456: FWIW the reason why types like `enum Type` are at the bottom is purely historical.
They should probably be closer to the top of the file.

src/ast.rs
```diff
@@ -1424,6 +1465,12 @@
         Some(self)
     }
 }
+impl CheckParse for ElseifClause {
+    fn can_be_parsed(pop: &mut Populator<'_>) -> bool {
```
krobelus, line 1469: I think you should squash all commits into one, to simplify things overall.
I don't want to read several different approaches in full, only the best one.

src/ast.rs
The0x539: It is not, but I wasn't sure how else to name it. I wanted to put the item in the regular `Token` trait, but object safety got in the way.
We're not actually using it abstractly, so I'll move it to an inherent `impl`.

src/ast.rs
The0x539: Sorry about the fine-grained commits.
The initial position of (some of?) my new items was mostly a result of being overwhelmed by the contents of this module and wanting to just get to prototyping.

src/ast.rs
The0x539: I'll look into that once this step is merged.
I'm quite fond of `enum_dispatch`, but it might not be the best fit here. A custom macro (perhaps with the help of `paste`) would allow us to deduplicate more code (I'm mostly thinking about the downcasting).

krobelus: possibly but we also need to implement generic methods like `Node::typ()` and `Node::source_range()`, ideally without splitting the node definition paragraphs.

> My original goal was to get rid of the None variants on StatementVariant and BlockStatementHeaderVariant, but I discovered that that would require a larger refactor.

One problem is that those union nodes are not real nodes. It's unclear if they should be but the status quo requires a lot of boilerplate that even spreads to macros (`visit_union_field_mut`).
Maybe we should remove union nodes like `Statement` and replace them with something that's not a full node but simply delegates to the concrete statement.

PR #10376, opened 2024-03-16
Un-hide documentation of non-fish shell builtins

krobelus:
When writing scripts for other shells, it can be confusing and annoying
that our `man` function shadows other manual pages, for example `exec(1p)`
from [Linux man-pages]. I almost never want to see the fish variant for such
contended cases (which obviuosly don't include fish-specific commands like
`string`, only widely-known shell builtins).

For the contented cases like `exec`, the POSIX documentation is more
substantial and useful, since it describes a (sub)set of languages widely
used for scripting.

Because of this I think we should stop overriding the system's man pages.
Nowadays we offer `exec -h` as intuitive way to show the documentation for
the fish-specific command (note that `help` is not a good replacement because
it uses a web browser).

Looking through the contended commands, it seems like for most of them,
the fish version is not substantially different from the system version.
A notable exception is `read` but I don't think it's a very important one.

So I'm saying: we can sacrifice a bit of the native fish-scripting experience
in exchange for playing nicer with the outside world.  I think the latter
is more important because scripting is not our focus, the way I see it.
So maybe put our manpath at the end.

In lieu of that, we should at least have `exec.rst` reference the system or
POSIX variants, which is what this draft patch does, to set the stage.

[Linux man-pages]: https://www.kernel.org/doc/man-pages/

### DIFF
--- a/doc_src/cmds/alias.rst
+++ b/doc_src/cmds/alias.rst
@@ -54,9 +54,11 @@ The following code will create ``rmi``, which runs ``rm`` with additional argume
     alias chrome='/Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome banana'
 
 
-See more
+See Also
 --------
 
-1. The :doc:`function <function>` command this builds on.
-2. :ref:`Functions <syntax-function>`.
-3. :ref:`Defining aliases <syntax-aliases>`.
+- The :doc:`function <function>` command this builds on.
+- :ref:`Functions <syntax-function>`.
+- :ref:`Defining aliases <syntax-aliases>`.
+- ``command man alias`` -- show docs for the non-fish variant.
+- The POSIX variant: <https://pubs.opengroup.org/onlinepubs/9699919799/utilities/alias.html>
--- a/doc_src/cmds/bg.rst
+++ b/doc_src/cmds/bg.rst
@@ -45,3 +45,9 @@ If only 123 and 789 exist, it will still background them and print an error abou
 ``bg 123 banana`` or ``bg banana 123`` will complain that "banana" is not a valid job specifier.
 
 ``bg %2`` will background job 2.
+
+See Also
+--------
+
+- ``command man bg`` -- show docs for the non-fish variant.
+- The POSIX variant: <https://pubs.opengroup.org/onlinepubs/9699919799/utilities/bg.html>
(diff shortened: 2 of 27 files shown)


faho:
The `test` documentation already includes something for this:

https://github.com/fish-shell/fish-shell/blob/46cde6cc556487ccb5be1bc7be33710dddd582bd/doc_src/cmds/test.rst?plain=1#L17-L21

And I think `test` is a good example for where we *want* to have our man page first - we're not 100% pure posix (we have extensions like floating point, other posixy shells may have other extensions, we may break what posix-compatibility we *have* - #10365).

Other builtins that are different from posix in ways that can catch people off-guard:

- `read` - this is *substantially* different and reasonably important
- `bg` - the other man page goes into detail on what the output looks like
- `exec` - it explains how it can open/close file descriptors, which ours does not do
- `echo` - does not have `-s`
- `exit` - talks about `trap`

These are important differences that make those manpages confusing when you are inside of fish. 
So I don't think we should put our manpath last - you are inside fish, if you're looking for `man test` that should reflect what `test` would do.

It's fine to add links to the other man pages, tho I would do it like test does, or at least do it consistently. Personally I also don't think the link to POSIX is useful - it's not very easy to read and describes a weird maybe-subset of what is actually on your system.

krobelus:
Fundamentally this is a political issue. I don't think you value my use case (wanting to write sh scripts, not fish) as much and at the same time I don't really empathize with people who write fish-specific scripts (beyond the occasional multi-line commandline)

faho:
Yup, basically.

Fundamentally I don't believe that running `man foo`, which is *the thing to get help* - the way everything teaches you to get help in unix - in a running fish should give you the sh-version.

So this is confusing to 1. new people, who have only learned that you do `man test` to get information on "test" and 2. people already used to fish, who know that `man test` in fish gets you information on fish's test.

And if I write sh scripts, I fire up bash and do a lot of line-by-line testing, so I would be running `man foo` in that.

> I don't really empathize with people who write fish-specific scripts (beyond the occasional multi-line commandline)

I *like* fish scripting and want it to be used more and easier to use and viable for more things. If I didn't like fish scripting I would probably not work on fish. I have to confess I am a bit taken aback by your position here.

-----

My position stays: What this currently does (add mentions of the non-fish docs) is *fine*, I would make it consistent with what we already have in the test man page (either way around, tho I have a slight preference for adding it at the top and only in the man page) and I would probably ditch the posix link because it's not relevant.

PR #10377, opened 2024-03-16
Update clasp completion

EmilyGraceSeville7cf:
## Description

- refactor code
- provide completions for projects, versions, deployments, advanced services and .js functions

Fixes issue #

## TODOs:


share/completions/clasp.fish
```diff
@@ -1,104 +1,124 @@
+function clasp_list_projects
```
krobelus, line 1: `__fish_clasp_list_projects`

EmilyGraceSeville7cf:
I am not sure whether it's the best approach to extract all JS function names for `clasp run {{function}} {{argument...}}` command:

```fish
function __fish_clasp_list_functions
    grep --extended-regexp '^\\s*function\\s+\\w+' --recursive --include '*.js' --no-filename --color=never . |
        string replace --regex '\\s*function\\s+(\\w+).*' '$1'
end
```

I guess the best solution is to utilize some JS parser to provide function names in a more reliable way (regex-based approach is not always accurate). This change also will allow us to provide some hints for function parameters like their default values. I have to do some research how to implement it, but it's my dream to do it.

share/completions/clasp.fish
```diff
@@ -1,104 +1,171 @@
...
+        string replace --regex '\\s*function\\s+(\\w+).*' '$1'
+end
+
+function __fish_clasp_list_subcommands
+    clasp --help |
+        string match --regex '^  [a-z]' --entire |
+        string replace --regex '^\\s{2}([a-z]+).*' '$1'
+end
+
+function __fish_seen_subcommands_from
```
krobelus, line 51: `__fish_seen_subcommands_from` has a name clash.
If the generic one doesn't work for some reason, it's fine to pu it into `clasp.fish` but use the `__fish_clasp` prefix if you do that.
Else things will behave differently depending on which completion script happens to be loaded first.

share/completions/clasp.fish
```diff
@@ -1,104 +1,171 @@
...
+        sheets 'Advanced Sheets' \
+        slides 'Advanced Slides' \
+        classroom Classroom \
+        people 'Advanced People' \
+        contacts-people 'Migrate from Contacts service to People API' \
+        tasks Tasks
+end
+
+function __fish_clasp_list_functions
+    grep --extended-regexp '^\\s*function\\s+\\w+' --recursive --include '*.js' --no-filename --color=never . |
```
krobelus, line 41: Some implementations of `grep` don't support --include or --color.
It's probably not relevant to most users of clasp so I don't think this is blocking.. I think we can find a better solution anyway

I think it would be preferrable to add a command like `clasp run --list-functions` to output the list of available functions because they will know best.

faho: Also long options are a compatibility issue in general, see e.g. [the OpenBSD man page](https://man.openbsd.org/grep) - which supports *some* long options, but only for compatibility with GNU grep where no short option is available.

share/completions/clasp.fish
```diff
@@ -1,104 +1,170 @@
...
+        sheets 'Advanced Sheets' \
+        slides 'Advanced Slides' \
+        classroom Classroom \
+        people 'Advanced People' \
+        contacts-people 'Migrate from Contacts service to People API' \
+        tasks Tasks
+end
+
+function __fish_clasp_list_functions
+    find . -name '*.js' -exec fish --command "cat {} | string replace --regex --filter '^\s*function\s+(\w+).*\$' '\$1'" \;
```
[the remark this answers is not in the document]
EmilyGraceSeville7cf: I've used `sed` as suggested to increase the performance.

PR #10382, opened 2024-03-19
ast: Use enum dispatch for tree traversal

The0x539:
## Description
A lot of the complexity here is a result of needing separate reference-flavored enums, because things that used to take `&dyn Node` still need an "abstractly-typed reference" even though the node's owner knows exactly what type it is.

I tried to avoid introducing custom proc macros, but [a mysterious hygiene bug](https://gist.github.com/The0x539/5ee18715cc3ce4a174aa16d7c2f65f20) forced my hand. Plus, it's arguably more readable than the declarative version, and the `blanket_ref` attribute is a nice bonus (those blanket impls are necessary for enum_dispatch to do its thing on the ref enums)

Future work:
- Change parent pointers to use `NonNull` and possibly `Pin`
- A derive macro for AST nodes
- Eliminate the `fn(&mut Populator, &mut NodeEnum)` pattern in favor of `fn<N: Node>(&mut Populator) -> N`, which can go in a trait that each node implements (taking the place of `AcceptorMut`).
  - This should also eliminate the need for the `Mut` flavor of node enum.
- Rearrange the enums a bit to make some invalid states unrepresentable. Major examples:
  - If the `top` field of the `Ast` struct is only expected to be two kinds of node, there should be an enum for that.
  - If a node can only have a specific type as a parent, then that can be encoded in the type of its parent pointer. We'd probably still have the current unknown-type `parent` method, but wouldn't rely on it as much.
- Do something about the "union nodes" (enums with names ending in `Variant`) as previously discussed. I don't fully understand their nature yet, but I can tell there's some redundancy. ~~Would they work similarly to tokens and keywords?~~

## TODOs:
No intended user-facing changes. Performance characteristics unknown.


macros/src/ast_node.rs
```diff
@@ -0,0 +1,172 @@
...
+                flavor.suffix_and_lifetime(inner)
+            } else {
+                flavor.ampersand(inner)
+            };
+            let middle_ty = flavor.suffix(middle);
+            let outer_ty = flavor.suffix_and_lifetime(outer);
+
+            let lifetime = (flavor != Flavor::Val).then_some(quote!(<'n>));
+
+            quote! {
```
krobelus, line 71: Unfortunately our build system is somewhat broken, so you'll need to extend this workaround to get reliable rebuilds when using cmake:

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 342e6fb55..b332387f3 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,3 +56,3 @@ add_definitions(-DCMAKE_SOURCE_DIR="${REAL_CMAKE_SOURCE_DIR}")
 # not run cargo, so we need to tell it *something*
-FILE(GLOB sources src/* src/*/* src/*/*/*)
+FILE(GLOB sources Cargo.* src/* src/*/* src/*/*/* macros/Cargo.toml macros/src/*)
```

would be nice to fix this in a general way, or at least detect this problem in CI.
But we don't add macro crates often.

src/ast.rs
```diff
@@ -2457,7 +1921,7 @@ pub struct Extras {
 pub struct Ast {
     // The top node.
     // Its type depends on what was requested to parse.
-    top: Box<dyn NodeMut>,
+    top: Box<NodeEnum>,
```
krobelus, line 1924: Does this still need to be boxed?
I guess we still need a stable address for `parent` but we should probably be using `Pin` for that.
In doubt we can leave it until we pin

src/ast.rs
```diff
@@ -2834,91 +2298,44 @@
...
-                    _ => (),
-                }
             }
+            // This field is an embedding of an array of (pointers to) ContentsNode.
+            // Parse as many as we can.
+            NodeRefMut::List(list) => match list {
+                ListRefMut::AndorJobList(l) => self.populate_list(l, false),
+                ListRefMut::ArgumentList(l) => self.populate_list(l, false),
+                ListRefMut::ArgumentOrRedirectionList(l) => self.populate_list(l, false),
+                ListRefMut::CaseItemList(l) => self.populate_list(l, false),
```
krobelus, line 2327: I wonder if we should define a generic type for lists, they're all the same.
We used to have that in C++ but I guess it goes against the exhaustive-enum-matching approach, so probably not.

src/ast.rs
```diff
@@ -3891,11 +3298,11 @@
     if top_type == Type::job_list {
         let mut list = pops.allocate::<JobList>();
         pops.populate_list(&mut *list, true /* exhaust_stream */);
-        ast.top = list;
+        ast.top = Box::new(NodeEnum::from(*list));
```
krobelus, line 3301: I believe the only place where we really need the generic `From` impl is in the `as_node()` and `as_mut_node()`.

If possible we should not use it in cases like this where we know the concrete type,
because that gives  more information. So I'd write it like

```
ast.top = Box::new(NodeEnum::Branch(BranchEnum::FreestandingArgumentList(
	*list,
)));
```

and so on.

Which begs the question whether we actually need a From impl?
Is there a better way to implement `as_node()` and `as_mut_node()`?
In a way that prevents us from accidentally using NodeEnum::from like this..
I don't have a concrete suggestion so it's fine to leave it.

krobelus, line 3301: > I tried to avoid introducing custom proc macros, but a mysterious hygiene
> bug forced my hand.

I get `error[E0412]: cannot find type `BranchEnum` in this scope`
Can't that be fixed?

> Plus, it's arguably more readable than the declarative version, and the
> blanket_ref attribute is a nice bonus (those blanket impls are necessary
> for enum_dispatch to do its thing on the ref enums)

I wonder how much we'd miss if we implemented enum_dispatch ourselves.
I have a preference for avoiding proc macros -- it should be possible
to implement a light version of enum_dipatch with normal macros,
maybe not match statements but if-else chains -- but I agree the
`enum` syntax works great. So this is probably the best way forward.

src/bin/fish_indent.rs
```diff
@@ -571,7 +573,7 @@ impl<'source, 'ast> PrettyPrinterState<'source, 'ast> {
         }
     }
 
-    fn emit_node_text(&mut self, node: &dyn Node) {
+    fn emit_node_text(&mut self, node: impl Node) {
```
krobelus, line 576: I think it's surprising that this is `impl Node` and not `&impl Node`. It works because this is a `NodeRef` in the concrete case.
Maybe it's better to be globally consistent about this (unless there are cases where we don't pass NodeRef?).

src/highlight.rs
```diff
@@ -1317,7 +1315,7 @@ impl<'s> Highlighter<'s> {
             self.pending_variables.push(var_name);
         }
     }
-    fn visit_semi_nl(&mut self, node: &dyn Node) {
+    fn visit_semi_nl(&mut self, node: NodeRef<'_>) {
```
krobelus, line 1318: `visit_semi_nl` can now take a `semi_nl: &SemiNl`.
Since `color_node` takes `impl Node`, this will work fine.

Same for other `visit_semi_nl` functions.

src/parse_execution.rs
```diff
@@ -175,15 +175,15 @@ impl<'a> ParseExecutionContext {
...
         &self,
         ctx: &OperationContext<'_>,
-        node: &dyn Node,
+        node: &impl Node,
         associated_block: Option<BlockId>,
     ) -> EndExecutionReason {
-        match node.typ() {
-            ast::Type::statement => {
-                self.eval_statement(ctx, node.as_statement().unwrap(), associated_block)
+        match node.as_node() {
```
krobelus, line 181: This is really nice overall, especially for writing match statements; LSP type definition is useful for `NodeRef` since it jumps to the enum node hierarchy.

It's a bit ugly that we have to use `node.as_node()` to convert from impl Node to the enum_dispatched enum.
Do you think it would be possible to get rid of the traits altogether and pass around something like `NodeRef`?

Combining traits with enum_dispatch means we need all those blanket trait impls for references.
I guess it's no big deal but since enum_dispatch doesn't need traits, I wonder if we still do?

For the methods we need to call generically, can't we call them `NodeEnum` already?
I guess it would be a bit ugly to copy all the `trait List` default impls
to every single list node but it's not like `define_list_node` is small either way.

Can we look at the cases where we inherently need traits, and maybe use enum dispatch everywhere else?
I know that the parent node pointer is one case where we need to use virtual dispatch,
but I don't think that applies to subtraits like Leaf.
I'm not sure what's the answer here but maybe you have enough context to find it.

src/parse_util.rs
```diff
@@ -1433,26 +1462,28 @@ fn detect_errors_in_backgrounded_job(
...
+    else {
         return false;
     };
 
-    if job_conj.parent().unwrap().as_if_clause().is_some()
-        || job_conj.parent().unwrap().as_while_header().is_some()
+    let mut errored = false;
+
+    if let NodeRef::Branch(BranchRef::IfClause(_) | BranchRef::WhileHeader(_)) =
+        job_conj.parent().unwrap().as_node()
```
krobelus, line 1478: maybe this is better written using `matches!`?

src/parse_util.rs
```diff
@@ -1505,14 +1536,18 @@
...
+    };
 
     // Walk up to the job.
     let mut job = None;
     let mut cursor = dst.parent();
     while job.is_none() {
         let c = cursor.expect("Reached root without finding a job");
-        job = c.as_job_pipeline();
+        if let NodeRef::Branch(BranchRef::JobPipeline(j)) = c.as_node() {
+            job = Some(j);
```
krobelus, line 1549: fwiw this loop should be rewritten to use `break job;`, then job doesn't need to be mutable

macros/src/ast_node.rs
```diff
@@ -0,0 +1,172 @@
...
+                    }
+                }
+            }
+            .to_tokens(stream)
+        }
+    }
+}
+
+#[derive(Copy, Clone, PartialEq)]
+enum Flavor {
```
krobelus, line 104: maybe better to move this up, before it's referenced

src/ast.rs
```diff
@@ -452,6 +208,103 @@
...
 }
 
+define_node! {
+    #[derive(Debug)]
+    #[enum_dispatch(Acceptor, Node)]
+    #[enum_dispatch(AcceptorMut, NodeMut)]
+    pub enum Node {
+        Branch,
+        Leaf,
+        List,
```
krobelus, line 218: Probably for a future commit/PR:
Can we get rid of `Type` and `Category` now?
They should be mostly redundant (already before but maybe it's easier now).
`ast_type_to_string()` is only used for debugging, it's fine to swap that out with a derived `Debug` impl that simply outputs the type name.

macros/src/ast_node.rs
```diff
@@ -0,0 +1,172 @@
...
+            };
+            let middle_ty = flavor.suffix(middle);
+            let outer_ty = flavor.suffix_and_lifetime(outer);
+
+            let lifetime = (flavor != Flavor::Val).then_some(quote!(<'n>));
+
+            quote! {
+                impl #lifetime From<#inner_ty> for #outer_ty {
+                    fn from(value: #inner_ty) -> Self {
+                        Self::from(#middle_ty::from(value))
```
krobelus, line 74: is the transitive thing something that `enum_dispatch` doesn't support?

macros/src/ast_node.rs
```diff
@@ -0,0 +1,172 @@
...
+            .to_tokens(stream)
+        }
+    }
+}
+
+#[derive(Copy, Clone, PartialEq)]
+enum Flavor {
+    Val,
+    Ref,
+    Mut,
```
krobelus, line 107: isn't this usually called RefMut? Maybe use `Value` instead of `Val` then.

macros/src/blanket_ref_impl.rs
```diff
@@ -0,0 +1,108 @@
...
+}
+
+/// Input:
+/// ```
+/// fn bar(&self, baz: Qux) -> Quux
+/// ```
+///
+/// Output:
+/// ```
+/// { T::bar(self, baz) }
```
krobelus, line 70: the input/output comments are helpful, maybe add them elsewhere too?
Maybe a high level one for blanket_ref, etc.
Perhaps use "realistic" names that sound like real AST nodes, so it's easier to categorize

AFAICT, T is the concrete node here so we implement all traits for `&IfStatement` and `&mut IfStatement`.
The mutable variant is questionable because we never need that outside the populator.

macros/src/ast_node.rs
```diff
@@ -0,0 +1,172 @@
+use proc_macro2::TokenStream;
+use quote::{format_ident, quote, ToTokens};
+use std::collections::HashMap;
+use syn::{parse_quote, punctuated::Punctuated, Fields, Ident, ItemEnum, Token};
+
+pub struct State<'a> {
```
krobelus, line 6: can you put the relevant PR description into the commit message, to help `git blame`

krobelus:
I think this is probably fine to merge with little modification but we should look into whether our trait hierarchy is still necessary, given that it overlaps with the methods added by enum_dispatch.

src/ast.rs
The0x539: Good eye. I'm pretty sure top level will still need a `Box`, so that moving an `Ast` doesn't invalidate the first level descendants. None of the other boxes are necessary, but I'm trying to limit the scope of each of these PRs.

I believe a fully idiomatic implementation would use `Pin<Box<T>>` for the root and `Option<Pin<NonNull<T>>>` for the children, as well as a `PhantomPinned` marker field in every node that has children.

However, `Pin` is mostly a guardrail. I think we already establish the necessary guarantees by:
- not exposing `&mut` access once parsing is complete
- not populating parent pointers until the tree is constructed

I plan on looking into `Pin`, but it may or may not prove cumbersome to work with here.

macros/src/ast_node.rs
The0x539: Correct. It doesn't maintain awareness of the other enums.

src/highlight.rs
The0x539: Done.

macros/src/ast_node.rs
The0x539: Done.

src/ast.rs
The0x539: I already have plans that will make this entire function go away. Specifically:
```rs
trait Parse {
    fn parse(pop: &mut Populator<'_>) -> Self;
}
```
Leaves, branches, lists, and variants(?) will all have their own implementations of this trait, using logic relocated from `impl Populator`.
Branch impls will basically do this:
```rs
Self {
    parent: None,
    kw_else: Parse::parse(pop),
    semi_nl: Parse::parse(pop),
    body: Parse::parse(pop),
}
```
There will still be special handling for the "missing `end`" case that `VisitResult` handles, but it will look quite different.
Maybe towards the end of these changes, parsing could even use `Result` and the `?` operator. Then `Parse` impls for branches with a `KeywordEnd` would just need to attach context from their level of the tree before propagating.

src/bin/fish_indent.rs
The0x539: This is a manifestation of what I consider a wart in my design, one that also appears in the form of the inherent `accept` method.
`&'a IfStatement` and `BranchRef<'a>` are both one indirection. `&NodeRef<'_>` is a double indirection. The type system doesn't really "understand" this, leaving us in our current situation.

I suspect things will get a little better as I make further changes to the design, but for now, I have no strong feelings about how to approach this.

Commit `5d9659ce464df78477fb0e7287f0e2106e293546` rids the codebase of any text matching the regex `: impl (ast::)?(Node|Keyword|Token)`, *mostly* by inserting `&` but in some cases I replaced the `impl Trait` with a specific "by-value ref-flavored enum".
Is this satisfactory?

macros/src/ast_node.rs
The0x539: Updated to `FILE(GLOB sources Cargo.* src/* src/*/* src/*/*/* macros/* macros/*/*)`.

src/ast.rs
The0x539: This ties into my third future-work bullet point: `Ast` eventually shouldn't even contain a general `NodeEnum`, it should be an `AstRoot` enum, or possibly even a generic.

src/parse_execution.rs
The0x539: Future-work bullet point 3: This code path will eventually use either a narrower `EvaluatableNodeRef<'a>` enum or a generic bounded by a new `EvaluatableNode` trait (names absolutely not final).

src/parse_util.rs
The0x539: I'll do you one better: that entire block can be a `match`, and we don't need an `errored` variable.

src/parse_execution.rs
The0x539: There is definitely room to _reduce_ our trait surface area, but the way to make that happen is somewhat complex. There are a few cases where it genuinely makes more sense to use a trait, as each callsite knows the exact concrete type it's passing and would have to upcast.

src/ast.rs
The0x539: Not yet, but it's definitely in the cards. `Category` especially.

macros/src/ast_node.rs
The0x539: I'm sorry, I don't understand what you mean by this.

src/parse_execution.rs
krobelus: > There are a few cases where it genuinely makes more sense to use a trait, as each callsite knows the exact concrete type it's passing and would have to upcast.

Do you have an example?

macros/src/ast_node.rs
krobelus: It's often a good idea to explain why some decisions where made,
the traditional place to put such explanation is the commit message.
For example it might be good explain why `blanket_ref` and `NodeRef` exist.

See also https://github.com/git/git/blob/79bdd48716a4c455bdc8ffd91d57a18d5cd55baa/Documentation/SubmittingPatches#L110-L225

src/parse_execution.rs
The0x539: A lot of stuff in `highlight` is the main thing that comes to mind. For example, `color_as_argument` doesn't do any downcasting and has three callsites, each of which has a known concrete type.

macros/src/ast_node.rs
The0x539: This enum's days are numbered anyway: we only use `NodeEnum` for the AST root, and we only use `NodeRefMut` for parsing. In both cases, it's private to the `ast` module and I've already described a refactor that would eliminate the enum.

src/parse_execution.rs
krobelus: I don't think `color_as_argument` qualifies; one reason is that it doesn't read its `node: impl Node` argument except for `source_range()` which we could compute at the call site. The other reason is that we want the `Node` trait regardless (because of parent pointers).
So only the other traits are relevant, we might want to get rid of the `Leaf` trait because we can pass a `LeafEnum` instead.

macros/src/blanket_ref_impl.rs
The0x539: True, the two traits we actually need the `#[blanket_mut]` *attribute* for are quite simple.
However, to implement that, we still need e.g., `impl<T: Node> Node for &mut T`, which isn't so trivial.
As long as we need to keep the mutable flavor around, we might as well have the attribute for it.

src/parse_execution.rs
The0x539: Point taken. My drafts of the `trait Parse` refactor do shrink down some of the traits significantly, which should put us in a better position to experiment with removing traits and/or swapping out `impl Trait` parameters for enums.

macros/src/ast_node.rs
The0x539: Since I've pushed some additional commits since your initial review, should I do this as part of a second squash commit or what?

src/ast.rs
The0x539: > I get "error[E0412]: cannot find type `BranchEnum` in this scope"
Not sure what you mean by that. On 1.77.0, the MRE I linked produces "cannot find value `inner` in this scope".
xiretza found [this old PR](https://gitlab.com/antonok/enum_dispatch/-/merge_requests/22) that allegedly fixeds the problem.
However, if we went back to a declarative macro, we'd need to use `paste`, and when I tried using that here, the readability was terrible.

There is a possible future where the design simplifies to the point that we can just have the enums defined without a macro, but we need to get rid of the `Enum` and `RefMut` flavors first to avoid descending (deeper?) into boilerplate hell.

macros/src/blanket_ref_impl.rs
The0x539: Doc comment added to top level.

krobelus:
I'd like to understand better why we need `NodeRef`, can we not pass a normal reference to a node enum?
Also I'd like to evaluate the impact of proc macros (on compile time), and how bad it would be to replace them with plain macros.
Finally, I'd get rid of all Node sub-traits. Also, if we get rid of parent nodes (which seems doable in theory, we could also get rid of Node.

That being said I'm not sure when I'll get around to it and I'm happy to delegate to you.
If you already have follow-up work, feel free to add it here or elsewhere;
I don't think you're blocked by this not merging yet, there shouldn't be an nontrivial conflicts.

krobelus:
prefer rebase on master instead of merging for this type of change. Merge commits are a complex thing

The0x539:
this has atrophied to the point that it would be better to start over when I can muster the attention

PR #10389, opened 2024-03-22
Added pbpaste command completions

faluciano:
## Description

Added pbpaste command completions

### DIFF
--- a/share/completions/pbpaste.fish
+++ b/share/completions/pbpaste.fish
@@ -1,0 +1,3 @@
+complete -c pbpaste -s help -d "Display command help"
+complete -c pbpaste -s pboard -d "Specifies which paste specifies which pasteboard to copy to or paste from. If no pasteboard is given, the general pasteboard will be used by default. {general | ruler | find | font}"
+complete -c pbpaste -s Prefer -d "Tells pbpaste what type of data to look for in the pasteboard first. {txt | rtf | ps}"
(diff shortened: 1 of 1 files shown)


share/completions/pbpaste.fish
```diff
@@ -0,0 +1,3 @@
+complete -c pbpaste -s help -d "Display command help"
+complete -c pbpaste -s pboard -d "Specifies which paste specifies which pasteboard to copy to or paste from. If no pasteboard is given, the general pasteboard will be used by default. {general | ruler | find | font}"
```
faho, line 2: Is "{general | ruler | find | font}" the list of possible pboards?

If so, you can just add them as arguments:

```fish
complete -c pbpaste -s pboard -d 'Which pasteboard to use' -xa 'general ruler find font'
```

Also these descriptions are pretty long, we try to keep it short.

faho: Oh also `-s pboard` is invalid - `-s` introduces a *short* option, meaning it is `-X`, with one character after the dash, and can be grouped etc (like in `ls -lah`).

If this is `pbpaste --pboard`, use `-l`, if it is `pbpaste -pboard` (one dash), use `-o`.

mqudsi (approved):
Match what I see on macOS, except I don't have `-pboard` on 10.10.

PR #10390, opened 2024-03-23
[RUSTSEC-2020-0070] Bump serial_test dependency to 1.0.0

Pi-Cla:
## Description
The current version of serial_test we use (0.4.0) depends on parking_lot 0.10.2 which in turn depends on lock_api 0.3.4.
This version of lock_api is vulnerable to [RUSTSEC-2020-0070](https://rustsec.org/advisories/RUSTSEC-2020-0070) This was patched in lock_api 0.4.2 but we need to update serial_test to get the update.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,11 +4,11 @@ version = 3
 
 [[package]]
 name = "ahash"
-version = "0.8.9"
+version = "0.8.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d713b3834d76b85304d4d525563c1276e2e30dc97cc67bfb4585a4a29fc2c89f"
+checksum = "e89da841a80418a9b391ebaea17f5c112ffaaa96f621d2c285b5174da76b9011"
 dependencies = [
- "cfg-if 1.0.0",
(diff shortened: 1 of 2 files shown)


mqudsi (changes requested):
I've applied it locally and confirmed that no older versions of the lock_api crate remain with this patch. 

Can you please re-submit this patch without the formatting changes to the toml file as it's a one-line change but it looks like a lot more currently.

mqudsi (approved):

zanchey:
This pulled in a huge chunk of dependencies, including some previously removed (eg jobserver) and a bunch of Windows targets.

krobelus:
I'll remove the dependency, it must be a simple one-liner that takes a lock

zanchey:
https://github.com/palfrey/serial_test/issues/6 - `parking_lot` was pulled in for its reentrant lock support, which didn't exist in std back then and isn't stabilised yet (https://doc.rust-lang.org/nightly/std/sync/struct.ReentrantLock.html)

zanchey:
Slimmed things down a bit with 792eff42b

PR #10391, opened 2024-03-23
Change git to git-core in specfile

Pi-Cla:
## Description
The git package in openSUSE and Fedora installs extra things besides git itself. git-core installs just the git cli without extra dependencies

### DIFF
--- a/fish.spec.in
+++ b/fish.spec.in
@@ -9,8 +9,8 @@ Group:                  System/Shells
 URL:                    https://fishshell.com/
 
 Source0:                %{name}_@VERSION@.orig.tar.xz
-# Corrosion requires git, and it's not always preinstalled
-BuildRequires:          cargo gettext gcc-c++ xz pcre2-devel git
+# Corrosion requires git-core, and it's not always preinstalled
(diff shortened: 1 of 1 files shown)


faho:
Corrosion is no longer a thing, so I think we can do without the git dependency entirely?

faho:
Turns out bdde2b2b35253ad7c39549e293369e47cb1befb6 was based on a misunderstanding to begin with, so I just reverted it.

mqudsi:
If fish.spec.in is indeed no longer used for any of the builds, shouldn’t we drop it?

mqudsi:
Also, the OBS nightly build logs [are still referencing corrosion](https://build.opensuse.org/package/live_build_log/shells:fish:nightly:master/fish/openSUSE_Tumbleweed/x86_64), fwiw.

zanchey:
The OBS uploads are turned off, until I can make them work properly again. So the spec file will be part of that. The errors you're seeing are from a 12-month old upload. I think they keep getting retriggered by distro changes; I can turn them off.

mqudsi:
Thanks, that explains it!

zanchey:
Sorry, this stuff is all completely clear in my head!

PR #10396, opened 2024-03-26
abbr: expand command abbrs after decorators

faho:
Currently, we expand command-abbrs (those with `--position command`) after `if`, but not after `command` or `builtin` or `time`:

```fish
abbr --add gc "git checkout"
```

will expand as `if gc` but not as `command gc`.

This was explicitly tested, but I have no idea why it shouldn't be?

Looking at the git history nothing really springs up as explanation.

## TODOs:

### DIFF
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -4337,7 +4337,7 @@ fn expand_replacer(
     ))
 }
 
-// Extract all the token ranges in \p str, along with whether they are an undecorated command.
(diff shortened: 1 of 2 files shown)


ridiculousfish:
I think I considered `command` to be an escape hatch in case you wanted to run some real command. But it's probably not wise to make abbreviations that are also real commands. Anyways I'm fine with this change.

ridiculousfish (approved):

Issue #10397, opened 2024-03-28
Autosuggestions sometimes get accepted without input


jm355:
```
~ $ fish --version
fish, version 3.7.0
~ $ echo $version
3.7.0
~ $ echo $TERM
xterm-kitty
```
On Arch Linux

I wasn't able to reproduce in a few minutes with the customization-free command, and I can't consistently reproduce on my system.

The issue is that often I will type a letter, and fish will complete the autosuggestion as if I had pressed the right arrow. This is most notable when I try to update. I have `u` as an alias for `sudo pacman -Syu`, and I have other commands I occasionally run like my `usbguard-allow` and `usbguard-deny` aliases. What will happen is, say the most recent "u*" command I've run is `usbguard-allow`, I will boot up my computer, open the terminal, type `u`, hit enter, and realize that `usbguard-allow` was expanded right before I hit enter and that's what gets run.

Sometimes it's not instant like that, and it happens maybe 5-10% of the time. I have no idea what's going on, it's super strange

kovidgoyal:
https://github.com/kovidgoyal/kitty/issues/4530

mqudsi:
> Ideally, fish would have some input function that moves the cursor right
without accepting an auto-suggestion. Then the kitty fish shell integration
could bind that function to some otherwise unused escape sequence and we
would not have this issue. However, there is no such function:
https://fishshell.com/docs/current/cmds/bind.html that I can see.
>
> Even more ideally, fish would add a function that has semantics of
"move cursor to the cell number X on line number Y, where Y is counted
from the line that starts the prompt".

The first seems eminently doable. The second might be a more difficult ask as things currently stand.

@kovidgoyal What do you expect the first to do if the cursor is at the end of input? Bell/alarm would be annoying.

kovidgoyal:
On Wed, Mar 27, 2024 at 09:11:39PM -0700, Mahmoud Al-Qudsi wrote:
> > Ideally, fish would have some input function that moves the cursor right
> without accepting an auto-suggestion. Then the kitty fish shell integration
> could bind that function to some otherwise unused escape sequence and we
> would not have this issue. However, there is no such function:
> https://fishshell.com/docs/current/cmds/bind.html that I can see.
> >
> > Even more ideally, fish would add a function that has semantics of
> "move cursor to the cell number X on line number Y, where Y is counted
> from the line that starts the prompt".
> 
> The first seems eminently doable. The second might be a more difficult ask as things currently stand.
> 
[...]

Issue #10268, opened 2024-01-27
No bind mode for search fields, and no way to make forward-char move the cursor


krobelus:
https://www.reddit.com/r/fishshell/comments/1ac09ed/rebind_the_arrow_keys_in_the_history_pager/ shows that in the search field (as created by Shift+Tab for normal completions or Control+R for the history pager)

1. `forward-char` and `backward-char` always move to the next/previous completion, with no way to move the cursors left/right
2. rebinding navigation keys doesn't really work; doing something like `bind \ck up-line` would break things because it also applies outside search mode

Not sure if one of the active contributors will consider this as worth doing themselves but it sounds reasonable to make bindings more flexible.
Maybe the history pager makes this important enough.
Though I kind of doubt it, it's better to get used to the default key bindings IMO.
Just I wanted to open this so it doesn't get lost.

wadeduvall:
So, I actually tried `bind \ck up-line`/`bind -M INSERT \ck up-line` etc and this doesn't take effect in the pager. So I really think this is a missing feature. I could be wrong (just switched to fish) but

faho:
> I actually tried bind \ck up-line/bind -M INSERT \ck up-line etc and this doesn't take effect in the pager. 

Nah, it's just spelt "insert", not "INSERT".

wadeduvall:
@faho ` bind -M insert \cj up-line` also does not work in the pager...

faho:
Works for me.

`\cj` *is* newline (as in the terminal sends the actual bytes for newline when you press control and j), it is possible you've bound something else.

wadeduvall:
@faho yup, tmux was grabbing it.

faho:
Anyway, the underlying issue here is that the pager is a bit of a mess, binding-wise.

It was added before the introduction of modes (which were added when the vi-bindings were, which IIRC was the only time someone got a bounty for working on fish), and the entire system doesn't work *particularly* well.

For the most part it works because we've hacked the defaults to do things in the pager. E.g. complete-and-search will navigate backwards, which doesn't make a lot of sense except when you consider that it's bound to shift-tab by default.

The way I believe this was originally envisioned was that you would script bindings with `commandline --paging-mode`:

```fish
if commandline --paging-mode
    commandline -f up-line
else
    commandline -f forward-char
end
```

Which has some issues, especially given that we enqueue the commandline functions again, so if you're scripting bindings you run into the limits quite quickly.

[...]

PR #10398, opened 2024-03-28
Add forward-char-passive

mqudsi:
Add a `forward-char-passive` binding. (Better suggestions for the name are being accepted!)

The idea is to add a "non-destructive" version of `forward-char` that strictly moves but does not change the "meta state" of the shell (does not accept auto completion, does not change completion pager selection). This has been requested by a terminal emulator developer and it makes sense to me at a high (enough) level.

I'm not going to pretend to be aware of where each binding is used, but it seems to me that *ideally* instead of adding something new to the existing `forward-char` and `forward-single-char` bindings, it would make more sense for one of them (`forward-single-char`) to not automatically accept suggestions. It seems `forward-single-char` is primarily used in the vi key bindings, and it isn't especially clear to me that accepting the autosuggestion makes sense there. But out of an abundance of precaution and to avoid breaking things, I am adding this as a new keybinding altogether.

Fixes issue #10397 

## TODOs:

- [X] Tests have been added for regressions fixed (test is confirmed to fail if `forward-single-char` is used instead)

### DIFF
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -2511,6 +2511,16 @@ impl ReaderData {
                     self.update_buff_pos(elt, Some(el.position() + 1));
                 }
             }
+            rl::ForwardCharPassive => {
+                let (elt, el) = self.active_edit_line();
+                if self.is_navigating_pager_contents() {
+                    // Do nothing
+                } else if self.is_at_end(el) {
+                    // Do nothing
+                } else {
+                    self.update_buff_pos(elt, Some(el.position() + 1));
+                }
+            }
             rl::BackwardKillWord | rl::BackwardKillPathComponent | rl::BackwardKillBigword => {
                 let style = match c {
                     rl::BackwardKillBigword => MoveWordStyle::Whitespace,
@@ -4495,6 +4505,7 @@ fn command_ends_paging(c: ReadlineCmd, focused_on_search_field: bool) -> bool {
         | rl::HistoryPager
         | rl::BackwardChar
         | rl::ForwardChar
+        | rl::ForwardCharPassive
         | rl::ForwardSingleChar
         | rl::UpLine
         | rl::DownLine
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -2497,6 +2497,14 @@ impl ReaderData {
                     self.update_buff_pos(elt, Some(el.position() - 1));
                 }
             }
+            rl::BackwardCharPassive => {
+                let (elt, el) = self.active_edit_line();
+                if el.position() != 0 {
+                    if elt == EditableLineTag::SearchField || !self.is_navigating_pager_contents() {
+                        self.update_buff_pos(elt, Some(el.position() - 1));
+                    }
+                }
+            }
             rl::ForwardChar | rl::ForwardSingleChar => {
                 let (elt, el) = self.active_edit_line();
                 if self.is_navigating_pager_contents() {
@@ -2513,12 +2521,10 @@ impl ReaderData {
             }
             rl::ForwardCharPassive => {
                 let (elt, el) = self.active_edit_line();
-                if self.is_navigating_pager_contents() {
-                    // Do nothing
-                } else if self.is_at_end(el) {
-                    // Do nothing
-                } else {
-                    self.update_buff_pos(elt, Some(el.position() + 1));
+                if !self.is_at_end(el) {
+                    if elt == EditableLineTag::SearchField || !self.is_navigating_pager_contents() {
+                        self.update_buff_pos(elt, Some(el.position() + 1));
+                    }
                 }
             }
             rl::BackwardKillWord | rl::BackwardKillPathComponent | rl::BackwardKillBigword => {
@@ -4504,6 +4510,7 @@ fn command_ends_paging(c: ReadlineCmd, focused_on_search_field: bool) -> bool {
         | rl::CompleteAndSearch
         | rl::HistoryPager
         | rl::BackwardChar
+        | rl::BackwardCharPassive
         | rl::ForwardChar
         | rl::ForwardCharPassive
         | rl::ForwardSingleChar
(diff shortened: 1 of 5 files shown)


kovidgoyal:
It would be nice if we had a pair of them for forward and back. back is not strictly necessary since as far as I know the regular backward-char doesnt do anything but that may not always remain true.

src/reader.rs
```diff
@@ -2511,6 +2511,16 @@ impl ReaderData {
                     self.update_buff_pos(elt, Some(el.position() + 1));
                 }
             }
+            rl::ForwardCharPassive => {
+                let (elt, el) = self.active_edit_line();
+                if self.is_navigating_pager_contents() {
+                    // Do nothing
```
krobelus, line 2517: > > Even more ideally, fish would add a function that has semantics of
> > "move cursor to the cell number X on line number Y, where Y is counted
> > from the line that starts the prompt".
> 
> The first seems eminently doable. The second might be a more difficult ask as things currently stand.

So the only difficulty is that move(X, Y) needs to potentially skip over Y-1 soft-wrapped lines?
We can probably use the current rendering if we want to do that.
If the current prompt rendering exceeds `$LINES`, does it still start from the prompt, or the first visible character?
Probably the first. Either way this scenario has always been somewhat broken in fish.

Anyway, making it work with something like `forward-char` seems more conventional here [*].

I'd expect that we'll eventually want a forward-char-like command that does not move in the pager and probably also not accept autosuggestion.
Crucially, when the pager is active, it should move the cursor inside the pager search field (which is impossible today).
This was suggested in https://github.com/fish-shell/fish-shell/issues/10268#issuecomment-1913142361;
With the recent input event queueing changes this might be possible now.
Quick sketch:

```
bind \cf '
	if commandline --paging-mode
	    commandline -f move-right
	else if commandline --is-at-end
	    commandline -f accept-autosuggestion-char
	end
	    commandline -f forward-char # different name to avoid breakage?
	end
'
```

I'm not sure if the autosuggestion part makes sense.
In general, it's not clear where the line should be between scripts and builtin behavior.

[*]: Note that currently we use this to move to a X/Y coordinates (logical, ignoring soft-wrapping).  Maybe it's worth making this more ergonomic.

```
    commandline -C 0
    for _line in (seq $line)[2..]
        commandline -f down-line
    end
    commandline -f beginning-of-line
    for _column in (seq $column)[2..]
        commandline -f forward-single-char
    end
```

tests/checks/tmux-complete.fish
```diff
@@ -102,3 +102,15 @@ isolated-tmux send-keys C-e M-f Space nothing
...
+isolated-tmux send-keys C-u 'bind \cs forward-char-passive' Enter C-l
+isolated-tmux send-keys C-u 'bind \cp backward-char' Enter C-l
+isolated-tmux send-keys C-u 'echo do not accept this' Enter C-l
+tmux-sleep
+isolated-tmux send-keys 'echo do not accept thi' C-p C-p Delete C-p C-s 'h'
+tmux-sleep
+isolated-tmux send-keys C-s C-s C-s 'x'
+tmux-sleep
+isolated-tmux capture-pane -p
+tmux-sleep
```
krobelus, line 115: the first and last sleep are unnecessary because there is no relevant async work going on.
Also, I'd `C-b` instead of `C-p` would be less confusing

src/reader.rs
krobelus: If we take `forward-char-passive`, I expect it should also move the cursor in paging mode (in the pager search field). I'm not sure what kitty integration expects.
Given that the `forward-char` naming is nicer for rebinding that the above if/else command,
it seems fine to have an overlapping low-level command. Perhaps we can find some kind of naming scheme for low-level commands.

kovidgoyal:
On Thu, Mar 28, 2024 at 12:45:22AM -0700, Johannes Altmanninger wrote:
> @krobelus commented on this pull request.
>
>
>
> > @@ -2511,6 +2511,16 @@ impl ReaderData {
>                      self.update_buff_pos(elt, Some(el.position() + 1));
>                  }
>              }
> +            rl::ForwardCharPassive => {
> +                let (elt, el) = self.active_edit_line();
> +                if self.is_navigating_pager_contents() {
> +                    // Do nothing
>
> If we take `forward-char-passive`, I expect it should also move the cursor in paging mode (in the pager search field). I'm not sure what kitty integration expects.

kitty shell integration only triggers if the cursor is at the prompt,
not elsewhere. I dont use fish so I dont know what the pager is, but
assuming it means the cursor is not at the shell prompt, kitty shell
integration wont care.

krobelus:
Yeah sounds good; FWIW the (completion) pager shows on Tab and the pager search field shows on Shift+Tab, Control+S or Control+R which move the cursor to that search field. Ideally we get rid of this search field but that's complicated

krobelus (approved):

mqudsi:
I think if we also handle the `self.pager.is_search_field_shown()` case then it's the same thing as just globally updating the position if `position != 0` (for backwards char) or if `!self.is_at_end(el)` (for forwards char)?

src/reader.rs
```diff
@@ -4495,6 +4511,7 @@ fn command_ends_paging(c: ReadlineCmd, focused_on_search_field: bool) -> bool {
         | rl::HistoryPager
         | rl::BackwardChar
         | rl::ForwardChar
+        | rl::ForwardCharPassive
```
krobelus, line 4514: this is missing the backward variant

mqudsi: Good catch.

mqudsi:
The new logic is 

```rust
                if !self.is_at_end(el) {
                    if elt == EditableLineTag::SearchField || !self.is_navigating_pager_contents() {
                        self.update_buff_pos(elt, Some(el.position() + 1));
                    }
                }
```

`elt` is `SearchField` only if both `is_navigating_pager_contents()` and `is_search_field_shown()` so this should match the logic you specified. If `!is_navigating_pager_contents()` then the cursor is moved regardless of whether the pager is empty, which should match the existing logic.

src/reader.rs
mqudsi: I'm going to update that match block to remove the `_` case so that that's a compilation error. `_` shouldn't be left in there for cases where new items need to be manually classified.

mqudsi:
Other name suggestions: 
* `forward-char-cursor`
* `forward-char-cursor-only`
*  `forward-cursor` 
*  `edit-cursor-right`

But it's probably just bike shedding at this point.

PR #10400, opened 2024-03-30
build.rs: pick up version file for tarballs from source, not build, directory

zanchey:
The version file for the tarballs ends up in the source directory, but build.rs
is looking in the build directory. This means that tarball builds report "fish,
version unknown" from `fish --version`.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -3,7 +3,7 @@
 use rsconf::{LinkType, Target};
 use std::env;
 use std::error::Error;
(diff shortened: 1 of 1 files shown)


mqudsi (approved):
The changes look good to me based off your description of the issue. What ends up creating the `version` file? Would it make more sense to change that file to be created in the build directory instead of in the source directory so that we don't pollute the source directory?

PR #10410, opened 2024-04-03
completions/insmod: Add support for insmod command

FirstLoveLife:
## Description

This commit introduces a new completion script for the `insmod` command, enabling Fish shell users to autocomplete file for `.ko` (kernel module) files in the current directory.

### DIFF
--- a/share/completions/insmod.fish
+++ b/share/completions/insmod.fish
@@ -1,0 +1,9 @@
+function _insmod_complete
+    # List all .ko files in the current directory for completion.
+    for file in (ls *.ko)
+        echo $file
+    end
+end
+
+complete -c insmod -a '(_insmod_complete)' -f
+


share/completions/insmod.fish
```diff
@@ -0,0 +1,9 @@
+function _insmod_complete
+    # List all .ko files in the current directory for completion.
+    for file in (ls *.ko)
+        echo $file
```
faho, line 4: This for-loop is broken in a number of ways:

1. `ls` can take options
2. `ls *` is duplicating work - we already *know* that the given files exist once the glob matches
3. It only works in the current directory and won't complete paths (like e.g. "/lib/modules/foo/", which you very likely want to do)
4. The entire loop is unnecessary, you can just do `printf '%s\n' *.ko`

But mostly, this is duplicating `__fish_complete_suffix .ko`. Just use that, remove the entire function.

FirstLoveLife: Hi faho, thanks for your quick reply.

> It only works in the current directory and won't complete paths (like e.g. "/lib/modules/foo/", which you very likely want to do)

IMO "/lib/modules/foo/" can be left for modprobe, insmod is more commonly used to insert a ko in the local directory. Including /lib/modules for insmod seems somewhat overhead(at least to me).

> But mostly, this is duplicating __fish_complete_suffix .ko. Just use that, remove the entire function.

Thanks for your tip! But I encounter an issue:
`complete -c insmod -a "(__fish_complete_suffix .ko)" -f`

This doesn't work somehow and <kbd>TAB</kbd> still lists all files, could you please provide some insights? Thanks.

faho: > IMO "/lib/modules/foo/" can be left for modprobe, insmod is more commonly used to insert a ko in the local directory. Including /lib/modules for insmod seems somewhat overhead(at least to me).

I see no reason to *not* support directories. 

> This doesn't work somehow and TAB still lists all files, could you please provide some insights? Thanks.

`__fish_complete_suffix` will complete *all* files, but leave matching ones at the top.

This is because the suffix is not actually authoritative - you can name files whatever you want.

This is a very common problem when people try to complete e.g. ".txt" files, and then notice that they have a lot of text files with another or no suffix.

What you want is

```fish
complete -c insmod -a "(__fish_complete_suffix .ko)" -kf
```

where the "-k" tells "complete" to not sort the completions but *k*eep the order they're in.

mqudsi:
`kldload` is a good one to copy here (it uses `__fish_complete_suffix` as well, but takes a default directory).

zanchey:
Merged upstream? https://github.com/kmod-project/kmod/pull/138

evelikov:
> Merged upstream? [kmod-project/kmod#138](https://github.com/kmod-project/kmod/pull/138)

Indeed `{ins,ls,rm}mod` completions have landed in upstream kmod for bash, fish and zsh. IMHO it makes sense to open PR there (as needed) and close this one.

PR #10418, opened 2024-04-04
Replace `__funced_md5` with `__fish_md5`

phanen:
`___funced_md5` seems redundant here, since we've already got `__fish_md5`


faho:
Thanks!

Issue #7781, opened 2021-03-05
Event handlers aren't traced


faho:
As I found out in #7780, when `fish_trace` is in effect variable handlers (and presumably other event handlers) don't show up.

E.g. (`>_` is my prompt, `>` is the trace and the line without a symbol is ordinary output):

```fish
>_ function foo --on-variable foo; echo foo handler; end
>_ set -g fish_trace 1
>_ set -g foo bar
> set -g foo bar
foo handler
```

It might be cool to show those as well, ideally somehow marked as event handlers. Maybe something like

```fish
>_ function foo --on-variable foo; echo foo handler; end
>_ set -g fish_trace 1
>_ set -g foo bar
> set -g foo bar
(event $foo) > echo foo handler
foo handler
```

?

mqudsi:
Note that they're intentionally not traced at the moment, in the original commit that introduced `fish_trace` (a7f1d2c) @ridiculousfish writes

>  The user's prompt is not traced unless you run it explicitly. Events are not traced because it is noisy; however autoloading is.

It seems the concern was too much potentially irrelevant output rather than confusing intermixing of standard and event-sourced trace messages.

faho:
Well, $fish_trace is for debugging, and right now it's quite awkward to not have a way to see *everything* that actually runs.

I would either enable events, or give a setting that enables them - set $fish_trace to 2 or "with-events" or something.

CodingAnarchy:
In the case of #7780, the problem was the result of an event handler brought in by `asdf-java`. Without being able to trace that in the event handler, debugging was not obvious.

ridiculousfish:
Yeah, fish currently checks if `$fish_trace` is non-empty but otherwise ignores its value. The plan was to later allow the value to control what gets traced, and the format of the output (e.g. maybe show line numbers). Maybe the simplest way to start is just with levels - 1 and 2 as faho proposes.

faho:
Numeric levels here seem kind of awkward. They remind me of the numbered debug levels.

I'd probably go with something like flog categories - `set fish_trace normal` would set it as it is currently, but `set fish_trace normal,events` would add events to that. I probably wouldn't go overboard and allow `set fish_trace normal,event-variable-PWD` to *just* trace the variable handler for $PWD.

Unless we have absolutely no idea what else would ever be added for this, in which case either "1" and "2" or "normal" and "with-events" are okay.

faho:
Okay, places where we suppress traces:

- Events
- The title
[...]

PR #10424, opened 2024-04-06
fish_trace filters to avoid suppressing traces from binds etc

krobelus:
Today fish_trace=1 traces all fish script code except bind/events/prompt/title.
(Also fish_should_add_to_history which is part of bind).

Let's allow enabling these categories.  This is mainly to help debugging; I'm
not sure how useful this is to people who are not very familiar with fish. I'd
only list this in the changelog once we have a better design (reject typos
like "fish_trace=binds" and maybe allow turning off the default script trace).

### DIFF
--- a/share/completions/set.fish
+++ b/share/completions/set.fish
@@ -54,7 +54,6 @@ function __fish_complete_special_vars
         FISH_DEBUG_OUTPUT "debug output path" \
         umask "current file creation mask" \
         fish_handle_reflow "if fish should repaint prompt when the term resizes" \
-        fish_trace "print cmds as they execute, like set -x" \
         fish_emoji_width "cols wide fish assumes emoji render as" \
         fish_key_bindings "name of function that sets binds" \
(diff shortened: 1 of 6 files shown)


faho:
Fixes #7781

krobelus:
also `set -lx fish_trace 1` should override any existing suppression I think, since this is a debuggin-only feature.

faho:
> also set -lx fish_trace 1 should override any existing suppression I think, since this is a debuggin-only feature.

You mean setting it to "1" should show *all* traces? I would rather have it stay as it is, because otherwise it's overly chatty.

How about an "all" option that turns on all traces instead?

krobelus:
I was just talking about the problem that there is currently no way to "un-suppress" tracing.
I agree "1/on/true/script" should (as it does today) trace only normal commands.
Agreed with the "all" option (or `*` I guess), this patch already includes it.

faho:
This now has a bunch of conflicts that would need to be fixed. I'm closing it for now.

PR #10427, opened 2024-04-07
Rewrite `wgetopt.rs` to Rustier syntax and naming

vertesians:
## Description

Almost all of the actual changes are in `wgetopt.rs`. The others are from replacing the old type and method names with the new ones.

I've left most of the logic from the original code intact. A full Rustification -- something more Iterator-based? -- would require a much more invasive set of changes. I've also tried to update the documentation and make it more consise, since it seemed wordy and left over from the original C code.

I don't know what other documentation needs to be updated, so I'm leaving the checkboxes below unticked.

## TODOs:


src/wgetopt.rs
```diff
@@ -25,341 +26,337 @@ Cambridge, MA 02139, USA.  */
...
-        if self.woptind != argc && self.argv[self.woptind] == "--" {
-            self.woptind += 1;
+        // The `--` element prevents any further scanning of options. We skip it like
+        // a null option, then swap with non-options as if were an option, then skip
+        // everything else like a non-option.
+        if self.wopt_index != argc && self.argv[self.wopt_index] == "--" {
+            self.wopt_index += 1;
 
-            if self.first_nonopt != self.last_nonopt && self.last_nonopt != self.woptind {
+            if self.first_nonopt == self.last_nonopt {
```
ridiculousfish, line 294: The original code read:

```rust
           if self.first_nonopt != self.last_nonopt && self.last_nonopt != self.woptind {
                self.exchange();
            } else if self.first_nonopt == self.last_nonopt {
                self.first_nonopt = self.woptind;
            }
```

so this has dropped a condition. I can be convinced that `self.last_nonopt != self.woptind` will always be true, given that we just incremented `woptind`; but in that case we should assume it's true in both branches.

src/wgetopt.rs
```diff
@@ -25,341 +26,337 @@ Cambridge, MA 02139, USA.  */
...
-        if self.ordering == Ordering::PERMUTE {
-            // If we have just processed some options following some non-options, exchange them so
-            // that the options come first.
-            if self.first_nonopt != self.last_nonopt && self.last_nonopt != self.woptind {
-                self.exchange();
-            } else if self.last_nonopt != self.woptind {
-                self.first_nonopt = self.woptind;
+    /// Advance to the next element in `argv`. Returns nothing on success, or
+    /// an optional value if we need to stop.
+    fn next_argv(&mut self) -> Result<(), Option<char>> {
```
ridiculousfish, line 263: This return type doesn't match the comment.

Before this change, we had the following return types:

1. `None` => options are fully processed, caller should exit their loop, `woptind` now points at remaining arguments.
2. `Some(c) if c != '\0'` => a special return; I think `NON_OPTION_CHAR` is the only possible return here
3. `Some('\0')` => an option was encountered and the caller should handle it. It can be determined by inspecting `self.argv[self.woptind]` and `nextchar`

Probably worth pulling out this return type into a real enum; in particular `Result` suggests an error when it's actually fine and expected.

src/wgetopt.rs
```diff
@@ -369,251 +366,187 @@ impl<'opts, 'args, 'argarray> wgetopter_t<'opts, 'args, 'argarray> {
...
-        if let Some(pfound) = pfound {
-            self._update_long_opt(&pfound, nameend, longind, indfound, retval);
-            return true;
+    /// Check for a matching long-named option.
+    fn handle_long_opt(&mut self, longopt_index: &mut usize) -> Option<char> {
+        let name_end = self
+            .remaining_text
+            .chars()
+            .take_while(|c| !matches!(c, '\0' | '='))
+            .count();
```
ridiculousfish, line 481: For reference the previous code was:

```
        let mut nameend = 0;
        while self.nextchar.char_at(nameend) != '\0' && self.nextchar.char_at(nameend) != '=' {
            nameend += 1;
        }
```

Note that `char_at` returns '\0' for one-past-the-end; we don't expect it to actually be in the option's name. So this should just find the position of any `=` sign; or `nameend.len()` if none.

ridiculousfish (changes requested):
Lots of great cleanup here - thank you!!

I have some comments on, well, comments, and also on `exchange()`. Also please squash into a single commit (or alternatively leave it to me, I can do that on merge).

src/wgetopt.rs
vertesians: In the original code, the condition `self.last_nonopt != self.woptind` is only checked if `self.first_nonopt != self.last_nonopt` is true -- which is by definition equivalent to `self.first_nonopt == self.last_nonopt` being false.

Therefore, checking `self.first_nonopt == self.last_nonopt` and only checking `self.last_nonopt != self.woptind` if it's false should be logically equivalent to the original.

vertesians:
Glad to hear you like the changes, applied most of your suggestions.

I'm not sure how to go about squashing commits -- I didn't even know git had that feature until now. Sorry about the trouble, I'd appreciate it if you handled that this time. I'll make sure to properly squash commits in future PRs.

ridiculousfish:
Looks like tests are failing now - argparse used `usize::MAX` as a sentinel meaning not found, now we're getting 0 back.

vertesians:
Should not have trusted clippy blindly there, whoops. It should work now.

vertesians:
... sorry, tried to squash the last few commits but ended up making a mess of things. Embarrasing. :confused:

ridiculousfish:
Squash-merged as 13230cdda0c2565c7f501e2f29c9871b11ca84f5. Thank you!

Issue #10432, opened 2024-04-09
cd fails for directory without read permission, displays `Permission denied`


TheJonny:
`cd` or `pushd` into a directory without read permission (but with x permission) does not work in fish.

```
jonny@charon /> sh -c 'env HOME=$(mktemp -d) XDG_CONFIG_HOME= fish'
Welcome to fish, the friendly interactive shell
Type help for instructions on how to use fish
jonny@charon /> echo $version
3.7.1
jonny@charon /> ls -ld /srv
drwxr-x--x 1 root root 48 14. Jan 23:51 /srv/
jonny@charon /> cd /srv
cd: Permission denied: '/srv'
jonny@charon / [1]> bash
[jonny@charon /]$ cd /srv
[jonny@charon srv]$ pwd
/srv

```
This transscript was on archlinux. The  same behaviour on debian stable with fish `3.6.0`

it is possible to cd through the directory, if I have read permission in the target, e.g. `cd /srv/www`.

TheJonny:
here an strace.

[strace.log](https://github.com/fish-shell/fish-shell/files/14917869/strace.log)

the only EACCES occurs in line 2947, followed by printing the error message:
```
2947 openat(AT_FDCWD, "/srv", O_RDONLY|O_CLOEXEC) = -1 EACCES (Permission denied)
2948 write(2, "cd: Permission denied: '/srv'\n", 30) = 30
```

TheJonny:
`builtins/cd.rs` calls `fds::wopen_dir`, which calls `fds::open_dir`, which sets `O_RDONLY`. changing it to O_PATH works for me on Linux, but is neither portable nor correct for a function called `open_dir`

faho:
Fixed by merging #10442.

PR #10433, opened 2024-04-09
Draft: cd without a file discriptor, patch out cwd_fd in Parser

TheJonny:
Fixes issue #10432, but removes the (currently unused) functionality of having a working directory in the Parser.

## TODOs:

### DIFF
--- a/src/builtins/cd.rs
+++ b/src/builtins/cd.rs
@@ -2,15 +2,13 @@
 
 use super::prelude::*;
 use crate::{
+    common::wcs2zstring,
     env::{EnvMode, Environment},
-    fds::wopen_dir,
     path::path_apply_cdpath,
     wutil::{normalize_path, wperror, wreadlink},
 };
 use errno::Errno;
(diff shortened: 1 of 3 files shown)


src/builtins/cd.rs
```diff
@@ -124,12 +114,6 @@ pub fn cd(parser: &Parser, streams: &mut IoStreams, args: &mut [&wstr]) -> Optio
             }
         };
 
-        // We need to keep around the fd for this directory, in the parser.
-        let dir_fd = Arc::new(fd);
-
-        // Stash the fd for the cwd in the parser.
-        parser.libdata_mut().cwd_fd = Some(dir_fd);
```
krobelus, line 131: Since this was explicitly re-added to the parser fairly recently (for possible concurrent execution) I'd keep the `cwd_fd`.

I'm not sure why a directory would ever have `--x` permissions (why keep the execute bit, who are you helping?) but I guess we could and should support that case.
chdir() first and open() later sounds fine to me.

TheJonny: I would close this in favor of #10436

Issue #4506, opened 2017-10-27
Implement `history insert`


admirabilis:
I have created two functions that would be interesting to have ported as plugins. One is a precision calculator directly from the prompt (ex. running `3*(1/2)` returns `1.5` instead of expanding globbing and command substitution, and the other runs [thefuck](https://github.com/nvbn/thefuck) automatically on the fish_command_not_found event.

Now, both of them write to the history file, and would benefit greatly from a `history insert` command. I'm currently using the following:

    if set -q XDG_DATA_HOME
        set __fish_history $XDG_DATA_HOME/fish/fish_history
    else
        set __fish_history ~/.local/share/fish/fish_history
    end

    echo -e '- cmd: '(commandline)'\n  when: '(date +%s) >>$__fish_history
    history --merge

The problem with this approach is that `commandline` is written to the history file as is (without considering newlines, without escaping special characters, etc.) and, of course, `history --merge` brings the undesired side-effect of merging history from all other sessions.

ridiculousfish:
`history push` sounds OK to me if you want to do it!

mqudsi:
Note that writing directly to the history file should not be done. We obtain a lock in the C++ code any time we write to it to prevent cross-thread/process corruption of the history file.

The only way this could be done without C++ code _somewhere_ would be if we switched the history file lock to `flock`, which I would personally be averse to.

zanchey:
The history file does lock with `flock`, although it doesn't try very hard (and is particularly unreliable on NFS, and also is implemented in terms of `fcntl` on systems without `flock`).

mqudsi:
Awesome, I should have checked instead of presuming it was a mutex. Thank you!

PR #10439, opened 2024-04-13
 add history append subcommand 

Accurate0:
## Description

This adds the history subcommand for `append` to allow history items to be added without executing the command.

Fixes issue #4506

## TODOs:


src/builtins/history.rs
```diff
@@ -343,6 +346,13 @@ pub fn history(parser: &Parser, streams: &mut IoStreams, args: &mut [&wstr]) ->
             history.save();
         }
         HistCmd::HIST_UNDEF => panic!("Unexpected HIST_UNDEF seen"),
+        HistCmd::HIST_PUSH =>
+        {
+            #[allow(clippy::unnecessary_to_owned)]
+            for arg in args.iter().copied() {
+                history.add_commandline(arg.to_owned());
+            }
```
Xiretza, line 354: No need to silence the lint:

```suggestion
            for &arg in args {
                history.add_commandline(arg.to_owned());
            }
```

Xiretza (approved):
Looks good to me overall except for a small nit; could you squash the commits down to one (or maybe an extra one for the whitespace changes) as well?

src/builtins/history.rs
Accurate0: I get this without silencing the lint, same as the `HIST_DELETE`

Xiretza: Yes, if you follow the clippy suggestion and then fix the type error (oops, clippy bug), you get `for &arg in args` as in my suggestion.

Accurate0: Oh sorry, I didn't notice the change, thanks. Will fix.

krobelus:
LGTM. Should add the new command to `share/completions/history.fish`.
I wonder if we ever want to insert in the middle, I guess that's unlikely

krobelus:
I guess we could call it `append` so it's clear that it's not a stack of history files.. though I'm not sure if that matters

Accurate0:
I've added it to the completions as push, let me know if you'd prefer it changed to append. I went off the original suggestion in the issue

krobelus:
I think I prefer "append" because that's the term used for opening files with `>>`, and history is still a text file.
It also fits `set --append` which we use for variables. Note that we  also have `$history` which is in reverse order so to append we'd want to use `--prepend`.. but the variable is read-only, fortunately.
I don't expect that we want it writable because misuse would be destructive.

Accurate0:
> I think I prefer "append" because that's the term used for opening files with `>>`, and history is still a text file. It also fits `set --append` which we use for variables. Note that we also have `$history` which is in reverse order so to append we'd want to use `--prepend`.. but the variable is read-only, fortunately. I don't expect that we want it writable because misuse would be destructive.

Seems fair to me, I've made the changes and squashed to keep the commits clean

Accurate0:
Check failure doesn't look related to my changes, let me know if you disagree otherwise it might just need a rerun

PR #10440, opened 2024-04-13
Add ruff completions

adamjstewart:
## Description

[ruff](https://github.com/astral-sh/ruff) is rapidly replacing tools like [black](https://github.com/psf/black), [isort](https://github.com/PyCQA/isort), and [flake8](https://github.com/PyCQA/flake8) within the Python community. This PR adds tab completions for ruff.

These completions were generated by running:
```console
> ruff generate-shell-completion fish > ruff.fish
```
and should be updated from time to time.

## TODOs:


faho:
>These completions were generated by running:

I would prefer we did here what we do for pip:

https://github.com/fish-shell/fish-shell/blob/master/share/completions/pip3.fish

That would automatically update the completions, so no updating is necessary.

faho:
Thanks!

PR #10442, opened 2024-04-15
cd: open directory with O_PATH or O_SEARCH

TheJonny:
## Description

fish implements the `cd` builtin using the system calls `open` and `fchdir`. On platforms where it is supported, this PR changes the flags argument for `open` to `O_SEARCH` or `O_PATH`, which does not require read permission on the directory we change into. This is more consistent with the behavior of `chdir` or `cd` in other shells.

also it cleans up the unused "mode" argument to `fds::[w]open_dir`.

Fixes issue #10432
Alternative to #10436 or #10433

## TODOs:

- [x] Changes to fish usage are reflected in user documentation/manpages. (no usage changes should appear)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -159,6 +159,7 @@ Other improvements
 - ``fish_indent`` will now collapse multiple successive empty lines into one (:issue:`10325`).
 - The HTML-based configuration UI (``fish_config``) now uses Alpine.js instead of AngularJS (:issue:`9554`).
 - ``fish_config`` now also works in a Windows MSYS environment (:issue:`10111`).
+- `cd` into a directory that is not readable but accessile (permissions `--x`) is now possible (:issue:`10432`).
 
 .. _rust-packaging:
 
--- a/src/builtins/cd.rs
+++ b/src/builtins/cd.rs
@@ -3,7 +3,7 @@
 use super::prelude::*;
 use crate::{
     env::{EnvMode, Environment},
-    fds::wopen_dir,
+    fds::{wopen_dir, BEST_O_SEARCH},
     path::path_apply_cdpath,
     wutil::{normalize_path, wperror, wreadlink},
 };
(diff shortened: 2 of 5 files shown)


TheJonny:
I opened https://github.com/rust-lang/libc/pull/3659 to add O_SEARCH to the libc crate for macos.
 - do we want to hardcode the constant for macos or wait?
 - do we want to add it to nix first?

zanchey:
From the linked source, it looks like they only exist on macOS 13+. This probably needs a build.rs check.

TheJonny:
> From the linked source, it looks like they only exist on macOS 13+. This probably needs a build.rs check.

that would also be cleaner, I think.

i would add cfg flags `HAVE_O_SEARCH` and `HAVE_O_PATH` in `build.rs`, and not hardcode `O_SEARCH = 0x40100000`, but wait for it to land in libc.

any opinions on this?

TheJonny:
> From the linked source, it looks like they only exist on macOS 13+. This probably needs a build.rs check.

@zanchey you are right, O_SEARCH does not work on macos 12. but the `#define` exists in the libc - at least in the github action. (I don't have access to another macos 12)
It works fine with macos-13.

I tried it here: https://github.com/TheJonny/mactest/

TheJonny:
O_SEARCH = (0x40000000 | 0x00100000)
does the right thing on macos 13, and can open directories with read permission on macos 12, so it should silently fall back to the current behavior without a compile time version check.

ridiculousfish:
For macOS, rather than a build.rs check, I'd rather do a runtime check - that way we can provide a single package for 10.9+. I will be happy to implement that.

TheJonny:
> For macOS, rather than a build.rs check, I'd rather do a runtime check 

After a quick experiment using the macos-12 github runner, just adding the O_SEARCH flag from macos-13 has no effect, opening readable directories still works. in this case, the runtime check is not needed in fish, because it is also implemented (by ignoring it) in the kernel. but we should verify it on more cases / versions / platforms

ridiculousfish:
I do intend to validate the behavior on macOS <=12 but I just discovered the Python tests haven't been running on my Mac for a while now and they've bitrotted...anyways I haven't forgotten this

TheJonny:
Also the next nix release should contain the O_SEARCH flag for all platforms :)

ridiculousfish:
Now that pexpect tests are passing for me again on macOS, I was able to verify that `O_EXEC` (0x40000000) is indeed ignored on macOS < 12, so we don't need a runtime check after all!

Merged as 32ba8d93afd0ae4250535bfc6c857997356b34d4. Thank you, great contribution!!

PR #10449, opened 2024-04-18
Release automation script

krobelus:
We provide releases in several different formats and channels.
There are many manual steps involved in making a release.
This makes it hard to learn or share improvements (especially share improvements that last once-and-for-all).
Also there is potential for error -- for example our Debian PPAs don't include the latest release and our website doesn't link to the latest tarball.
With our current approach, packaging tends to get more complicated, for example we'll want to add things like "Depends: librust-hexponent" to our Debian package, and set the version based on Cargo.toml.

Add a script to automate the boring steps of the release process.
This should obsolete most of https://github.com/fish-shell/fish-shell/wiki/Release-checklist

It seems to work on my Linux system but it's obviously not battle-tested yet.
The most essential TODO comments ones are for uploading Debian packages.

Iteration on the script itself is usually quite fast.  I run something like

    git add build_tools/release.sh
    git commit -m wip
    build_tools/release.sh 3.8.0 --dry-run --no-test # plus eventual --no-rebuild --no-rebuild-debs

As a temporary hack, the Debian packages are currently built in a Docker container.
It'd be easy to skip Docker when debmake is installed on the host system.
We don't yet pass through gpg-agent into Docker, so the debuild process requires interactive password input.

It takes a really long time (> 15 minutes) to build all Debian packages because we rebuild them sequentially for all Ubuntu releases (mantic, jammy, focal, bionic and xenial) even though only "debian/changelog" is different.
For iterating on the script, we can use "--no-rebuild-debs" to reuse the the packages after they've been built once.
Of course this is only valid if the tarball has not changed since.
We could probably ask debuild to validate this, or speed up the process in a different way, at least parallelize.

Future ideas:
- we could run the script in CI instead, to keep us even more honest.
- we should try to simplify the website release process (should not need to
  talk to the github API, the script already has enough information).
- do something similar for the the macOS-specific parts


zanchey:
> Also there is potential for error -- for example our Debian PPAs don't include the latest release and our website doesn't link to the latest tarball.

I don't think this is right. The Debian archive at https://software.opensuse.org/download.html?project=shells%3Afish%3Arelease%3A3&package=fish definitely contains the latest release, and the tarball link at fishshell.com is correct too (although it doesn't appear in the list of "Release history", which could probably be changed).

.github/workflows/release.yml
```diff
@@ -0,0 +1,261 @@
+name: Create a new release
+
+on:
+  push:
+    tags:
+      - '*.*.*'
```
ridiculousfish, line 6: Elegant!

.github/workflows/release.yml
```diff
@@ -0,0 +1,261 @@
...
+  packages-for-linux:
+    needs: [is-release-tag]
+    name: Build single-file fish for Linux (experimental)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          # Workaround for https://github.com/actions/checkout/issues/882
+          ref: ${{ github.ref }}
+      - uses: ./.github/actions/rust-toolchain@oldest-supported
```
ridiculousfish: A self-note that I should check that this works back to what we nominally support for 4.0.x (macOS 10.9). 4.1.x will drop support for these very old versions.

krobelus: macOS needs it yeah, but I don't think `x86_64-unknown-linux-musl` or `aarch64-unknown-linux-musl` have a particular reason to require 1.70 (requirements on linux version did not change), so I'll change those to use stable

PR #10452, opened 2024-04-20
Allow restricting abbreviations to specific commands

### LINKED ISSUE #9411: abbr: Allow restricting to a specific command's arguments

faho:
## Description

This allows defining abbreviations that are only triggered when the current command matches, so you can define

```fish
abbr --add --commmand git c checkout
```

Which turns `c` into `checkout` but only for `git`. This can also be combined with `--function`, expanding a git *subcommand*, replacing git aliases, would be something like

```fish
function git_checkout
    set -l opts version h/help C= c=+ 'exec-path=?' html-path man-path info-path p/paginate \
        P/no-pager no-replace-objects bare git-dir= work-tree= namespace= super-prefix= \
        literal-pathspecs glob-pathspecs noglob-pathspecs icase-pathspecs
    # skip command and abbr token
    set -l toks (commandline -xpc)[2..-2]
    argparse $opts -- $toks 2>/dev/null
    or return

    set -q argv[1]
    and return 1

    echo checkout
    return 0
end

abbr --command=git co --function git_checkout
```

(because, as always, options are complicated!)

One limitation of the current implementation is that the command is *not* part of the abbr "ID", the "name" still needs to be unique across commands.

I believe that's acceptable for now, the available workaround is `--regex`, and this could be lifted in future. For a specific abbr to be used with multiple commands you can specify `--command` multiple times and any of these will be allowed (including an empty command, indicating the command position).

Also parsing the command out could be made easier by additional options to `commandline` - `--current-command` and `--current-arguments` (for everything *after* the command).

And the function could be made more useful if you could pass arguments, so you can do multiple git aliases easily without writing a function for each - `abbr --command git co --function 'git_alias checkout'`. But that's out of scope for this issue.

Fixes #9411 
Previous attempt at #9412 

## TODOs:


ccoVeille:
I'm awaiting this for years 🤩👍

ccoVeille:
Thanks again for this PR. It would make things so easier.

> One limitation of the current implementation is that the command is not part of the abbr \"ID\", the \"name\" still needs to be unique across commands.

What are the next steps? About these limitations? Do you have plan? Is there an opened issue about it?

Thanks

faho:
> What are the next steps? About these limitations? Do you have plan? Is there an opened issue about it?

Let's first ship this and see how it works out.

The "limitations" here are idle musings, there's no need to open an issue about them. If it ends up being a common problem, we can handle it then.

g0t4:
I assume this isn't intended to work with nested subcommands? 

i.e.:
```shell
docker compose d<TAB>
docker image p<TAB>
```

faho:
Correct. The shell cannot determine what is a subcommand and what isn't. E.g. in `git -C log merge`, the "merge" is the subcommand, while in `git -P log merge` it's "log".

You'll have to write a function that knows `docker`.

PR #10462, opened 2024-04-26
Fix the acidhub prompt for a commitless repository

ClaXXX:
## Description

Deletes the error message generated by git when comparing indexes for a commitless repository and print '-' as the branch name.

To test:
```
 mkdir test && pushd test
git init
```
`fatal: bad revision 'HEAD'` is then printed in front of the prompt after each command execution as long as the current directory is the one we just created (`test`).

## TODOs:


faho:
Thanks!

(in future, please don't back-merge master, it only messes up the later merge. if you need to pull in new commits, rebase and force-push to your branch instead)

PR #10466, opened 2024-04-27
Move push_timer to measure command substitution timing too

### LINKED ISSUE #9100: builtin time does not measure command substitutions
I was trying to figure out why some benchmarking of a git.fish completion change I was considering, seemed not to be evidencing the same difference if I used `time`, versus if I ran it a bunch of times and compared with `CMD_DURATION`.

Specifically, the change was to the line `test "$root/$relfile" = (pwd -P)/$relfile;`:

```shell
~/f/fish-shell $ set relfile ./fish.png
~/f/fish-shell $ set root /Users/floam/fish/fish-shell/
~/f/fish-shell $ for i in (seq 100); test "$root/$relfile" = (pwd -P)/$relfile; end
~/f/fish-shell $ echo $CMD_DURATION
25
~/f/fish-shell $ for i in (seq 100); test "$root/$relfile" -ef ./$relfile; end
~/f/fish-shell $ echo $CMD_DURATION
11
```

So it using `test … -ef …` here appears to be twice as fast. 

However, time reports:

```shell

~/f/fish-shell $ time test "$root/$relfile" -ef ./$relfile

Executed in   61.00 micros    fish           external
   usr time    9.00 micros    9.00 micros    0.00 micros
   sys time   55.00 micros   55.00 micros    0.00 micros

~/f/fish-shell $ time test "$root/$relfile" = (pwd -P)/$relfile

Executed in    8.00 micros    fish           external
   usr time    8.00 micros    8.00 micros    0.00 micros
   sys time    4.00 micros    4.00 micros    0.00 micros
```

Accurate0:
## Description

I've moved the push timer to happen before where command substitution seems to occur, works correctly with the simple repro described on the ticket:

image

Fixes #9100

## TODOs:

### DIFF
--- a/src/parse_execution.rs
+++ b/src/parse_execution.rs
@@ -1585,13 +1585,18 @@ impl<'a> ParseExecutionContext {
             0
         };
 
+        let job_wants_timing = job_node_wants_timing(job_node);
+        let job_is_background = job_node.bg.is_some();
+        let job_is_simple_node = self.job_is_simple_block(job_node);
+
+        // Start the timer here to ensure command substitution is measured
+        let _timer = push_timer(job_wants_timing && (job_is_simple_node || !job_is_background));
+
         // When we encounter a block construct (e.g. while loop) in the general case, we create a "block
         // process" containing its node. This allows us to handle block-level redirections.
         // However, if there are no redirections, then we can just jump into the block directly, which
         // is significantly faster.
-        if self.job_is_simple_block(job_node) {
-            let do_time = job_node.time.is_some();
-            let _timer = push_timer(do_time);
+        if job_is_simple_node {
             let mut block = None;
             let mut result =
                 self.apply_variable_assignments(ctx, None, &job_node.variables, &mut block);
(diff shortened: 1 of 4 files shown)


src/parse_execution.rs
```diff
@@ -1672,6 +1672,8 @@ impl<'a> ParseExecutionContext {
             job.internal_job_id,
         );
 
+        // Start the timer here to ensure command substitution is measured
+        let _timer = push_timer(job.wants_timing());
```
mqudsi, line 1676: Is there a reason you dropped the `no_exec` check here?

Accurate0: yes, this function is guarded with a `no_exec` check

https://github.com/fish-shell/fish-shell/blob/14451ff051e099ea6ecd25434a1c5d3f77b65dc0/src/parse_execution.rs#L1564-L1566

krobelus, line 1676: the behavior change looks good (bash does the same) although I think we should consolidate the two calls to `push_timer()` into one. I don't think there is a reason to treat "simple blocks" specially here.
Then we can also get rid of `JobProperties::wants_timing`.
Once we support timing background functions we would want to add something similar back

Accurate0: seems fair, I made some changes, but it feels a little weird.

it maintains the current behavior of `fish -c "time begin; echo -n (sleep 1); end &"` returning timing even though timing a background job is an error, since the block isn't run as a background job

not sure if that should be an error too? technically it doesn't do anything wrong since the block isn't backgrounded, but it feels weird syntactically for that to be accepted

Accurate0: from what i understand, the timing background rule seems to not apply to blocks?

krobelus: `begin; end &` is not supported yet and the `&` is simply ignored. So it's fine if the `time` behavior is weird.

Accurate0: Okay sure, in that case I think the PR is ready for another review, I don't believe the changes I made differ in behavior for the `begin; end &` situation

PR #10467, opened 2024-04-28
Increase MSRV to 1.70

faho:
## Description

This increases the minimum supported rust version to 1.70, from 1.67.

In this case it allows us remove a few backports - "is_some_and" and "is_ok_and" and "is_err_and". I could not find "is_none_or", so I kept it in.

The reason for increasing it is mostly that 1.67 does not *gain* us anything. According to https://repology.org/project/rust/versions, none of the distros we build packages for are on 1.67, 1.68 or 1.69 in supported releases.

Here are the distributions on those versions in supported releases:

- Gobolinux on 1.68 - which currently ships [fish 3.1.2](https://github.com/gobolinux/Recipes/tree/master/Fish), so I don't believe they have a lot of fish users
- ArchLinux 32 (an unofficial port of Arch to 32-bit intel arches) on 1.69, which has [1.77 in staging](https://archlinux32.org/packages/i686/extra-staging/rust/), so hopefully they'll upgrade soon

it also lists Buildroot and YiffOS, but that appears to be outdated information - Buildroot 2024.02 has rust 1.74, YiffOS lists 1.70 in [the online package browser](https://packages.yiffos.gay/rust) (and frankly seems to be pretty dead).

-------------------

We have never really "formalized" an MSRV policy. I would like it to be roughly "we try to enable distro packages with distro compilers".

Basically we don't increase MSRV willy-nilly (no "last X releases" policy), but we also don't keep it lower where it doesn't help. We're not going to support Debian 11's 1.48, so keeping 1.67 doesn't help there.

### DIFF
--- a/src/builtins/test.rs
+++ b/src/builtins/test.rs
@@ -6,7 +6,6 @@ use crate::should_flog;
 mod test_expressions {
     use super::*;
 
-    #[allow(unused_imports)]
     use crate::nix::isatty;
     use crate::wutil::{
         file_id_for_path, fish_wcswidth, lwstat, waccess, wcstod::wcstod, wcstoi_opts, wstat,
--- a/src/builtins/test.rs
+++ b/src/builtins/test.rs
@@ -7,7 +7,6 @@ mod test_expressions {
     use super::*;
 
     #[allow(unused_imports)]
-    use crate::future::IsOkAnd;
     use crate::nix::isatty;
     use crate::wutil::{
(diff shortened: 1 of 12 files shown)


zanchey (approved):

src/builtins/test.rs
```diff
@@ -7,7 +7,6 @@ mod test_expressions {
     use super::*;
 
     #[allow(unused_imports)]
```
krobelus, line 9: should remove the attached `#[allow(unused_imports)]` as well

krobelus (approved):

PR #10471, opened 2024-04-30
Upgrade to Alpine v3.19 and add cargo as dependency

fossdd:
## Description

Alpine v3.13 is out-of-date since 2021-01-14 and is not getting any security updates anymore: https://alpinelinux.org/releases/

Cargo and rustc seems to be required in order to pass tests.

Fixes issue #

## TODOs:


krobelus:
thanks. A lot of these (test-only) dockerfiles are somewhat stale, we should autogenerate them

Issue #9406, opened 2022-12-09
Question: complete inside quotes


jdupak:
Example of situation:
```shell
hyperfine "git status"
```
Hyperfine takes a command as an argument inside quotes. I managed to get full command autocomplete without quotes the way sudo does it, but I found no way to influence competition inside the quotes. Right now I a getting just commands there, but not the subcommand/flag competitions. Is there any way to do that?

krobelus:
This will be a nice feature though not easy to implement

ridiculousfish:
As a hack, one could do the following in fish script:

1. Get the token via `commandline -t`
2. Use `string unescape` to strip the quotes
3. Use `complete -C` to complete it as a command
4. Use `string escape` to re-escape it, and offer that as a completion

krobelus:
we also need a way to tell fish not to escape those completions again (via something like `complete -c hyperfine -x --arguments-no-escape "__fish_complete_quoted_command"`).
This quick attempt technically works but fails to handle many cases.

```diff
diff --git a/share/completions/sh.fish b/share/completions/sh.fish
new file mode 100644
index 000000000..0f724c488
--- /dev/null
+++ b/share/completions/sh.fish
@@ -0,0 +1 @@
+complete sh -s c -d "Execute argument as command" -xa '(__fish_complete_quoted_command)'
diff --git a/share/functions/__fish_complete_quoted_command.fish b/share/functions/__fish_complete_quoted_command.fish
new file mode 100644
index 000000000..8721cb683
--- /dev/null
+++ b/share/functions/__fish_complete_quoted_command.fish
@@ -0,0 +1,20 @@
+function __fish_complete_quoted_command
+    set -l cmd "$(commandline -t | string unescape)"
+    set -l missing_closing_quote true
+    if string match -qr -- '[\'"]$' $cmd
+        set missing_closing_quote false
+    end
+    set -l earlier_tokens
+    printf %s\n $cmd | read --array --tokenize earlier_tokens
+    if string match -qr -- '\S$' $cmd
+        set -e earlier_tokens[-1]
+    end
+    set -l escaped_candidates
+    for candidate in (complete -C $cmd --escape | string replace -r '\t.*$' '' | string unescape)
+        set -a escaped_candidates "$(printf "%s %s" "$earlier_tokens" $candidate | string escape)"
+    end
+    if $missing_closing_quote
+        set escaped_candidates (string replace -r '[\'"]$' '' -- $escaped_candidates)
+    end
[...]

Issue #1439, opened 2014-04-28
Provide subshell like functionality


jorbsd:
While fish avoids forking/execing, having subshell like functionality would still be very nice.

In bash/zsh:

pwd; (cd <otherdir>; <dostuff>; pwd;); pwd

should be in the first dir, change and do stuff, then show that you are back in the original did

This closure of shell state in the subshell (vars, directories, etc.) is very handy.

bash/zsh modern syntax uses $() to be similar to () in fish, both replacing ``

c22:
If I'm not mistaken, this problem kinda goes away if #563 is fixed, right?

ridiculousfish:
Maybe, maybe not, depending on how it's fixed. A subshell is both concurrent execution and independent state (multiple $PWDs, etc).

d:
variables (exported or not) is a case where I'd love to have a subshell:

### Scenario
```fish
begin --subshell
    docker-machine env aws-fast | source
    docker run --rm --volume /ccache ubuntu tar c /ccache
end | begin --subshell
    docker-machine env aws-slow | source
    docker run --rm --volume /ccache tar x
end
```

ridiculousfish:
The `begin --subshell` syntax is pretty clever

alphapapa:
My two cents: I'd love to have a `withd`-style command.  My situation: running a shell in Emacs, I want to run `make` in the parent directory, but end up back in the current directory, without having to explicitly save and restore the directory.  This is easy in Bash with subshells, but not in Fish.  A `withd` could be put together as a script now, one that takes a string, but it would preclude syntax highlighting.

nrnrnr:
Either `cd -l` or `withd` would work for me.

stephane-chazelas:
subshells are also useful to isolate a portion of code with regards to signal delivery (think SIGPIPE for instance).

Note that ksh93 _fakes_ subshells without forking (so holds a copy of all original resources (umask, cwd, traps, aliases...) which are restored when the subshell exits). You may want to have a look at that for your `cd -l`, alias -l, umask -l, function -l (like for cd -l open the current directory on a fd with O_CLOSEEXEC, and fchdir to that latter).

ridiculousfish:
That's kind of amazing that ksh93 does that. Thanks for the tip, I definitely want to take a look.

aaronjwood:
If fork/exec was implemented, at least optionally, parallelism could also be restored. For example:
```
sleep 10 &
sleep 10 &
[...]

PR #10472, opened 2024-05-01
Raw-quoting syntax

krobelus:
TL;DR: this makes it so that typing `(` at the start of a command opens a subshell;
completions, quoting etc. work just like inside command substitutions.
This is enabled by the first commit that adds raw quoting syntax.

---

Other shells allow to create subshells using `(cd somewhere && do-something)`.
Our answer to this is to create a subshell like a regular process (#1439):

	$SHELL -c 'cd somewhere && do-something'

This syntax works the same way for non-shell programs

	watch 'ps aux | grep foo'
	git bisect run $SHELL -c 'make || exit 125; make test'
	hyperfine '...'

Issues with our solution are:
- the argument has no command-highlighting, completions (#9406) or indentation
- the  user may need to escape the argument multiple times ("quoting hell").
This primarily affects arguments containing backslashes and quotes
(for example `$SHELL -c 'printf %s\\\n *'`).
In many other cases backslash escaping is not necessary however the rules
can be confusing.

One workaround is to compose the subcommand in a separate shell prompt and
copy-paste it (inside quotes for automatic escaping).
That's not ideal because it breaks the writer's flow, and it doesn't work
well if the command needs to be refined.

Add a `(( ))` raw quoting syntax for subshells and deferred commands.
Syntax highlighting, completion and indentation work the same as in top-level
commands. No expansion/unescaping takes place:

	$SHELL -c ((printf %s\\n *))

Nested parentheses are okay, similar to Tcl's braced strings.
Imbalanced quotation marks are an error.
This means that the quoted string cannot contain an imbalance of its quote
characters.
(The good side is that, in many cases, imbalance is not intentional).

If we stick to this syntax, it would be good to also support `(< >)` and
`({ })` with the same meaning as `(( ))`, to make it easier to use when the
quoted content contains an unmatched `(`.

---

Also add a `{{ }}` raw quoting syntax that does not get shell highlighting,
completion and indentation.
This is meant for languages like Perl, awk, Python and regex dialects (no
more need to escape backslashes).
The would be a breaking change, we might want to avoid that by picking a
different syntax.

To-do:
- figure out how we want to tweak syntax, possibly find something that can
be added to Bash.
- adjust __fish_tokenizer_state
- docs/changelog/tests

### DIFF
--- a/share/completions/ip.fish
+++ b/share/completions/ip.fish
@@ -401,7 +401,7 @@ function __fish_complete_ip
                             case netns
                             case link-netnsid
                             case vf mac
-                            case {{max,min}_tx_,}rate
+                            case {max,min}_tx_rate rate
                             case node_guid port_guid
                             case state
                                 echo auto
--- a/share/functions/__fish_shared_key_bindings.fish
+++ b/share/functions/__fish_shared_key_bindings.fish
@@ -120,6 +120,16 @@ function __fish_shared_key_bindings -d "Bindings shared between emacs and vi mod
         # Shift-space behaves like space because it's easy to mistype.
         bind --preset $argv shift-space 'commandline -i " "' expand-abbr-backtrack
 
+        bind --preset $argv "(" __fish_insert_subshell
+        function __fish_insert_subshell
+            if string match -qr '^\s*$' -- (commandline --current-process --cut-at-cursor | string collect) &&
+                    not string match -qr '\($' -- (commandline --cut-at-cursor)
+                commandline -i '$SHELL -c (( '
+            else
+                commandline -i "("
+            end
+        end
+
         bind --preset $argv enter execute
         bind --preset $argv ctrl-j execute
         bind --preset $argv ctrl-m execute
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -3773,6 +3773,7 @@ impl<'s> Populator<'s> {
                 && [
                     TokenizerError::unterminated_quote,
                     TokenizerError::unterminated_subshell,
+                    TokenizerError::unterminated_raw_quote,
                 ]
                 .contains(&self.peek_token(0).tok_error)
             {
@@ -3809,6 +3810,7 @@ impl<'s> Populator<'s> {
                 && [
                     TokenizerError::unterminated_quote,
                     TokenizerError::unterminated_subshell,
+                    TokenizerError::unterminated_raw_quote,
                 ]
                 .contains(&self.peek_token(0).tok_error)
             {
--- a/src/common.rs
+++ b/src/common.rs
@@ -524,6 +524,8 @@ fn unescape_string_internal(input: &wstr, flags: UnescapeFlags) -> Option<WStrin
         Unquoted,
         SingleQuotes,
         DoubleQuotes,
+        QuotedSubshell(usize),
+        RawQuoted(usize),
     }
     let mut mode = Mode::Unquoted;
 
@@ -603,7 +605,15 @@ fn unescape_string_internal(input: &wstr, flags: UnescapeFlags) -> Option<WStrin
                     }
                 }
                 '{' => {
-                    if unescape_special {
+                    if input.char_at(input_position + 1) == '{' {
+                        input_position += 1;
+                        mode = Mode::RawQuoted(1);
+                        to_append_or_none = if unescape_special {
+                            Some(INTERNAL_SEPARATOR)
+                        } else {
+                            None
+                        };
+                    } else if unescape_special {
                         brace_count += 1;
                         to_append_or_none = Some(BRACE_BEGIN);
                         // We need to store where the brace *ends up* in the output.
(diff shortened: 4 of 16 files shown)


faho:
Note: I changed the title to avoid confusing people. "Subshell" is a very specific thing, this doesn't really add it, and we've overused the term in the past.

----

My first instinct here is that `((` does not strike me obvious as quoting, and it is not familiar to people coming from bash or python.

We've brought up `"""` python-style quoting before, I feel like that's more intuitive.

There is also the question of how to deal with indentation in quotes - one problem with embedding python scripts specifically is this:

```fish
if foo
    python -c "print('foo')
print('bar')" # <- this *cannot* be indented or it breaks python
end
```


I also am not a fan of deferring to `$SHELL` in the keybinding, which can be any sort of shell, and often is not fish. To be honest, I would remove the binding.

ridiculousfish:
Agreed that we should not promote the "run fish recursively" psuedo-subshell in key bindings. We'll have a real solution for that which preserves variables, functions, etc. which would be incompatible with invoking $SHELL.

+1 to `"""` for raw strings.

krobelus:
Does this mean we agree or disagree on adding a command-flavored quoting syntax to enable
`with-cd ./office (( do work ))`?

faho:
> we agree or disagree on adding a command-flavored quoting syntax

As I understand it, it has the exact same semantics, language-wise, it's only different interactively in that it will tell the reader "this in here is a command, highlight it like one"?

I'm not really sold on it, I don't think fish-script-inside-of-fish-script is a common enough thing to add more syntax for.

Specifically for some cases like `fish -c ... &` to background scripts we'd be better off just allowing blocks/functions to be backgrounded (which, granted, is a bigger task) than add syntax to make the workaround nicer *interactively*.

And again the symbols are alien (`(())` is arithmetic expansion in bash) except if you're used to TCL, which I don't have fond memories of (it's a big part of why we moved off of expect to pexpect), and which I don't think is very popular. `{{}}` is probably a bit worse because there's more backwards-incompatibility there, but I don't like `(())` as a choice either. Funnily enough `(<>)` would at least be a syntax error at the moment (also in bash, so there's less chance for confusion).

What moving to `"""` and `'''` would gain us is more familiar quoting (it's in python) and an obvious way to have "quoting with less escaping" but still interpolate variables, which seems like an important part.

krobelus:
actually for my use case plain quoting is enough. Raw quoting is `r""`.

krobelus:
The primary and concrete problem I'm trying to solve is to quickly run some commands in a different directory.
I don't know how to do this in a nice way while making sure that `$PWD` is restored on `ctrl-c`.
I believe this is an important scenario.

So far I've been using `sh -c 'cd ... && ...'`.
The immediate problem with that is that completion is not yet implemented.
I have an implementation that is almost finished, so that should cover many simple cases.
(As you noted, it has false positives for other shells but I think we can accept that as an intermediate step?)

I think there's a reasonably good chance that we want something better (native, no quoting hell etc.)

Raw quoting / deferred commands is one approach. It's definitely not the only one, and I'd love to hear other ideas such as

```
cd --scoped ./somewhere
    echo do something somewhere
    echo do something else
end
```

There's also some ideas leveraging `begin/end` (like `cd -l`) that would work well if they could be written inside `{}` or `()` but that's probably not an easy sell either.

I think that we should probably figure that out before we talk about a concrete quoting syntax.
For that there are countless possibilities, I'm sure we can find something that does not mislead users.
However, if the syntax could hold a fish script payload I expect that it will be foolish to not support trivial nesting.

Raw quoting for generic strings (regex etc) is also worthwhile exploring but it's less related to my problem and I haven't yet found real-world use case where it helps.

> """ still interpolate variables, which seems like an important part.

For the generic one maybe but it's probably less needed for quoted commands (`--no-scope-shadowing` is a rare exception).

faho:
> So far I've been using sh -c 'cd ... && ...'.
The immediate problem with that is that completion is not yet implemented.

Of course now you're in sh-land, and fish won't really ever be able to complete sh-commands.

But it's also just... a much bigger tool than you would need for the idea at hand.

> The primary and concrete problem I'm trying to solve is to quickly run some commands in a different directory.

I don't know why you would need a special quoting syntax for that, it seems like the right solution for this is something much more focused, like allowing $PWD to be set locally, so you can do `PWD=/some/dir; begin foo; end`.

Even if you *did* want a full "subshell" experience, that could just be something like `begin --subshell` - with no extra quoting *at all*, and no awkward switching into an entire different shell and language.

>  However, if the syntax could hold a fish script payload I expect that it will be foolish to not support trivial nesting.

I don't believe quoted fish-script inside of quoted fish-script is something we really need to be concerned with much. I don't believe it's a common thing, and when it comes up you could always just refactor (e.g. make multiple scripts).

krobelus:
> I don't know why you would need a special quoting syntax for that,

I did not say that.

> it seems like the right solution for this is something much more focused,
> like allowing $PWD to be set locally, so you can do PWD=/some/dir; begin
> foo; end.

`PWD=some-dir foo` sounds good to me. Discoverability will be worse but so be it

> and no awkward switching into an entire different shell and language.

?
sh is just an example, could be proper fish.
I don't think that playing nice with the outside world is a bad thing.

> I don't believe quoted fish-script inside of quoted fish-script is something
> we really need to be concerned with much. I don't believe it's a common
> thing, and when it comes up you could always just refactor (e.g. make
> multiple scripts).

you're repeating yourself / stating the obvious

krobelus:
how about `{}` or `()` for command grouping then?

faho:
Not possible considering backwards-compatibility.

I also don't like having `{}` pull double-duty like that.

krobelus:
`()` does not have a backwards compatibility issue

PR #10477, opened 2024-05-03
Clear gettext cache on locale switch

faho:
This fixes the issue where if you do `_ file` (which prints "file" in english) and then switch locale to e.g. de_DE.UTF-8 and do `_ file` again, it will still print "file" instead of the german "Datei".

CACHE INVALIDATION!!!111eleven!!! *shakes fist*

## TODOs:

This is awkward to test because it needs to require a non-english locale, and I'm not sure how to do that (or do we do that anywhere?).

### DIFF
--- a/src/env_dispatch.rs
+++ b/src/env_dispatch.rs
@@ -762,6 +762,7 @@ fn init_locale(vars: &EnvStack) {
                 _nl_msg_cat_cntr += 1;
             }
         }
+        crate::wutil::gettext::wgettext_clear_cache();
     }
 }
 
--- a/src/wutil/gettext.rs
+++ b/src/wutil/gettext.rs
@@ -64,6 +64,13 @@ enum MaybeStatic<'a> {
     Local(&'a wstr),
 }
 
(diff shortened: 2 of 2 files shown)


krobelus:
This seems like a reasonable change.
Most programs are expected to read environment variables only at startup, but the shell's `LANG` etc. are not only environment variables, we can generally try to react immediately.
Removing from the milestone for now

zanchey:
This is a regression from 3.7 though - all the `_nl_msg_cat_cntr` machinery was to enable runtime switching.

faho:
> This is a regression from 3.7 though - all the _nl_msg_cat_cntr machinery was to enable runtime switching.

@zanchey Nope, the same cache existed: https://github.com/fish-shell/fish-shell/blob/80394ea4e38138e0ce0bf23b016dc931e22ff788/src/wutil.cpp#L43-L44

I believe this either never worked or hasn't worked in quite a while.

For posterity: The issue with this specific commit is LSAN complains about a leak.

The *idea* is correct, but it needs to be done in a way that it doesn't leak these.

Maybe dropping the cache is alright (would need benchmarks tho)?

Issue #9030, opened 2022-06-21
"block" builtin is useless


faho:
Fish has the [block](https://fishshell.com/docs/current/cmds/block.html) builtin to "temporarily block delivery of events".

This sounds like a nice idea. Unfortunately I do not believe I have ever seen it used. It doesn't appear in any of the packages in oh-my-fish's packages-main, and it has a few design issues that cause it to be impossible to use for anything I can think of.

The only case where I ever wanted to use it was in our keybinding functions in order to prevent this:

https://github.com/fish-shell/fish-shell/blob/a004a10a807625379bec301ca1fd345d51ce3cab/share/functions/fish_default_key_bindings.fish#L10-L19

The two issues that prevented me are:

1. `block` blocks *all events*, so it has a larger impact than I am comfortable with here
2. `block` only *delays* these events

That second one makes it almost useless for this - we are trying to *prevent* recursion, but instead we've only delayed it. We still get the second, useless, function call.

It's also very very easy to abuse - run `block` interactively and notice that none of your event handlers run.

------

As I see it, there are two ways to deal with block:

Either we improve block to the point of usefulness, or we remove it and try to implement the good parts elsewhere.

That former bit is hard, because we'd want to add an event filter (which requires being able to specify events for specific variables!) and we would have to add a flag or change the default behavior to entirely *drop* events instead of delaying them.

This would look something like

```fish
block -l --drop --on-variable fish_key_bindings --on-event fish_postexec
```

to drop these two events for the current block of code.

The latter is much easier - the only bit I've ever wanted could be implemented with `set --no-event`. Just do `set --no-event fish_key_bindings fish_default_key_bindings` and drop that specific event. This would be easy to implement and, I find, easier to *read* and reason about.

We could then, if we find uses for other forms of event suppression, add them back later, in other ways.

faho:
Ah, I have found a use in the wild: https://github.com/IlanCosman/tide/commit/8d356d0d49e62399bef89abecac3c09286bb54af

This was later removed, without comment, in https://github.com/IlanCosman/tide/commit/0a33c9cc0b0cb155eaab15f3f146b1799e6b3f08

faho:
Punting from the 3.6.0 milestone because we clearly did not do it early in the release.

zanchey:
How about now?

PR #10480, opened 2024-05-06
Add `set --no-event` to change a variable without triggering an event

faho:
## Description

As mentioned in #9030 this adds a `--no-event` option to `set`, so you can set or erase a variable without triggering the event.

Example:

```fish
function on-my-var --on-variable myvar
    echo myvar was changed!
end

set --no-event myvar foo # <- this will not print "myvar was changed!"
```

There is no short option because this is an advanced option that should be used with care.

This does what a potential use of the `block` builtin would do in a simpler way, and it can be used to remove one of the gnarliest parts of our key binding functions: They set the variable, causing the handler to be run, which then runs the function again.

This upgrades our functions to use `--no-event`, but notably in a way that it'll handle being run by older fishes. This is necessary because otherwise we'd be left with no bindings at all. Tho the binding functions have basically been rewritten to handle the new bind notation, I'm not entirely sure how an old fish accidentally running the new version would work?

## TODOs:

### DIFF
--- a/doc_src/cmds/set.rst
+++ b/doc_src/cmds/set.rst
@@ -9,7 +9,7 @@ Synopsis
 .. synopsis::
 
     set
-    set (-f | --function) (-l | --local) (-g | --global) (-U | --universal)
+    set (-f | --function) (-l | --local) (-g | --global) (-U | --universal) [--no-event]
     set [-Uflg] NAME [VALUE ...]
     set [-Uflg] NAME[[INDEX ...]] [VALUE ...]
     set (-a | --append) [-flgU] NAME VALUE ...
@@ -101,6 +101,11 @@ Further options:
     If no variable names are given then all variables are shown in sorted order.
     It shows the scopes the given variables are set in, along with the values in each and whether or not it is exported.
     No other flags can be used with this option.
+
+**--no-event**
+    Don't generate a variable change event when setting or erasing a variable.
+    We recommend using this carefully because the event handlers are usually set up for a reason.
+    Possible uses include modifying the variable inside a variable handler.
 
 **-L** or **--long**
     Do not abbreviate long values when printing set variables.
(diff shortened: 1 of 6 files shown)


mqudsi (commented):
> Tho the binding functions have basically been rewritten to handle the new bind notation, I'm not entirely sure how an old fish accidentally running the new version would work?

I'd rather not introduce backwards compat workarounds into the codebase if we're not sure they're necessary. The only case I think we have to support is "user upgrades fish, is still running old fish because they haven't started a new session."¹ I don't think there's a way to re-`exec` the *old* fish binary (we always load it from the disk) so the new version of startup-only code (like the key bindings) should *theoretically* never be seen/executed by the older fish install, no?

If there's a situation I'm not seeing, then by all means, please enlighten me. Otherwise, we have never supported multi-versioning of fish and standalone, unpackaged builds of fish should be using the `share/` directory from the source code distribution and not from `$(PREFIX)/usr/share`, right?

¹ For the record, we have shipped upgrades in the past that broke this scenario. It's obvious what's happened if you are intimately familiar with fish, but I pity those that had an OS auto-update in the background and then started seeing random script errors when sourcing various completions or autoloading now-incompatible functions.

faho:
> I'd rather not introduce backwards compat workarounds into the codebase if we're not sure they're necessary

To be clear: Working around a lack of --no-event is necessary.

I was talking about how an old fish would interpret the new bind sequences - it would probably error out a lot, but I *think* there's enough there that the bindings should still be enough to at least reexec fish?

> If there's a situation I'm not seeing, then by all means, please enlighten me. 

The problem is that fish reloads functions that have changed on-disk.

So you have a long-running fish session, you upgrade fish, and in that long-running fish session *something* triggers a binding change. If that happens after it reloaded the function, you've now run `bind --erase --all --preset`, but *not* set the new bindings. You have completely empty bindings. *Nothing* does anything. (with the exception of any user-specified bindings)

You now have a hanging fish that is impossible to kill from inside.

I believe that's egregious and should not be allowed.

(personally I would lean towards removing the auto-reloading of functions to remove this scenario entirely, it's a major PITA on upgrade especially)

mqudsi:
> The problem is that fish reloads functions that have changed on-disk.

When and where does this happen? I always have to constantly `source /path/to/foo.fish` any time I edit the declaration of `foo` without `funced`.

krobelus (approved):

mqudsi:
Yeah, I think the feature itself makes sense and simplifies a lot of the trickier cases involving events. It's a straightforward escape hatch.

mqudsi (approved):

PR #10502, opened 2024-05-18
Use ahash for our own HashMap and HashSet objects

mqudsi:
We already transitively take a dependency on the ahash crate, so might as well use it. While not all of our HashMap/HashSet objects are performance-sensitive, some of them are and there's no cost to migrating them all over.

By default, rust uses a DoS-resistant, cryptographic hash (SipHash) to hash the keys for HashMap/HashSet entries into buckets, which is slower than what we used to use in the C++ world. We don't use any of it in a security-sensitive context or based off of unvalidated/untrusted/remote input, so there are no security reasons to continue to use SipHash.

Some of fish's HashMap/HashSet instances could benefit from being faster, e.g. the functions cache, various completion caches, environment variable tables, and history lookups, so this isn't just a "use it because it's there" kind of thing.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,6 +9,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e89da841a80418a9b391ebaea17f5c112ffaaa96f621d2c285b5174da76b9011"
 dependencies = [
  "cfg-if",
+ "getrandom",
  "once_cell",
  "version_check",
  "zerocopy",
@@ -98,6 +99,7 @@ source = "git+https://github.com/fish-shell/fast-float-rust?branch=fish#9590c33a
 name = "fish"
 version = "0.1.0"
 dependencies = [
+ "ahash",
  "bitflags 2.5.0",
  "cc",
  "errno",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,7 @@ once_cell = "1.17.0"
 rand = { version = "0.8.5", features = ["small_rng"] }
 widestring = "1.0.2"
 terminfo = "0.9.0"
+ahash = "0.8.11"
 
 [dev-dependencies]
 rand_pcg = "0.3.1"
--- a/src/abbrs.rs
+++ b/src/abbrs.rs
@@ -179,7 +179,7 @@ pub struct AbbreviationSet {
 
     /// Set of used abbrevation names.
     /// This is to avoid a linear scan when adding new abbreviations.
-    used_names: HashSet<WString>,
+    used_names: HashSet<WString, ahash::RandomState>,
 }
 
 impl AbbreviationSet {
--- a/src/autoload.rs
+++ b/src/autoload.rs
@@ -26,10 +26,10 @@ pub struct Autoload {
     env_var_name: &'static wstr,
 
     /// A map from command to the files we have autoloaded.
-    autoloaded_files: HashMap<WString, FileId>,
+    autoloaded_files: HashMap<WString, FileId, ahash::RandomState>,
 
     /// The list of commands that we are currently autoloading.
-    current_autoloading: HashSet<WString>,
+    current_autoloading: HashSet<WString, ahash::RandomState>,
 
     /// The autoload cache.
     /// This is a unique_ptr because want to change it if the value of our environment variable
@@ -190,7 +190,7 @@ struct AutoloadFileCache {
 
     /// The set of files that we have returned to the caller, along with the time of the check.
     /// The key is the command (not the path).
-    known_files: HashMap<WString, KnownFile>,
+    known_files: HashMap<WString, KnownFile, ahash::RandomState>,
 }
 
 impl Default for AutoloadFileCache {
@@ -205,7 +205,7 @@ impl AutoloadFileCache {
         Self {
             dirs,
             misses_cache: MissesLruCache::new(NonZeroUsize::new(1024).unwrap()),
-            known_files: HashMap::new(),
+            known_files: HashMap::default(),
         }
     }
 
--- a/src/complete.rs
+++ b/src/complete.rs
@@ -454,8 +454,8 @@ type CompletionEntryMap = BTreeMap<CompletionEntryIndex, CompletionEntry>;
 static COMPLETION_MAP: Mutex<CompletionEntryMap> = Mutex::new(BTreeMap::new());
 
 /// Completion "wrapper" support. The map goes from wrapping-command to wrapped-command-list.
-type WrapperMap = HashMap<WString, Vec<WString>>;
-static wrapper_map: Lazy<Mutex<WrapperMap>> = Lazy::new(|| Mutex::new(HashMap::new()));
+type WrapperMap = HashMap<WString, Vec<WString>, ahash::RandomState>;
+static wrapper_map: Lazy<Mutex<WrapperMap>> = Lazy::new(|| Mutex::new(HashMap::default()));
 
 /// Clear the [`CompleteFlags::AUTO_SPACE`] flag, and set [`CompleteFlags::NO_SPACE`] appropriately
 /// depending on the suffix of the string.
@@ -569,7 +569,7 @@ struct CustomArgData<'a> {
     /// When completing, variable assignments are really set in a local scope.
     var_assignments: &'a mut Vec<WString>,
     /// The set of wrapped commands which we have visited, and so should not be explored again.
-    visited_wrapped_commands: HashSet<WString>,
+    visited_wrapped_commands: HashSet<WString, ahash::RandomState>,
 }
 
 impl<'a> CustomArgData<'a> {
@@ -581,7 +581,7 @@ impl<'a> CustomArgData<'a> {
             do_file: true,
             wrap_depth: 0,
             var_assignments,
-            visited_wrapped_commands: HashSet::new(),
+            visited_wrapped_commands: HashSet::default(),
         }
     }
 }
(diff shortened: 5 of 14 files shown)


mqudsi:
Benchmark run before/after patch (`-old`/`+new`):

```diff
--- ./bench-pre	2024-05-17 20:43:46.485733321 -0500
+++ ./bench-post	2024-05-17 20:43:56.209657621 -0500
@@ -1,41 +1,41 @@
 Benchmark 1: release/fish --no-config benchmarks/benchmarks/aliases.fish
-  Time (mean ± σ):     711.9 ms ± 173.6 ms    [User: 520.5 ms, System: 218.2 ms]
+  Time (mean ± σ):     783.1 ms ±  94.3 ms    [User: 575.3 ms, System: 232.8 ms]
 --
 Benchmark 1: release/fish --no-config 'benchmarks/benchmarks/external_cmds.fish'
-  Time (mean ± σ):     840.8 ms ±   9.0 ms    [User: 406.0 ms, System: 427.4 ms]
+  Time (mean ± σ):     849.6 ms ±  23.7 ms    [User: 402.4 ms, System: 442.5 ms]
 --
 Benchmark 1: release/fish --no-config benchmarks/benchmarks/glob.fish
-  Time (mean ± σ):     153.1 ms ±   7.5 ms    [User: 115.5 ms, System: 37.6 ms]
+  Time (mean ± σ):     156.8 ms ±  16.2 ms    [User: 115.3 ms, System: 41.5 ms]
 --
 Benchmark 1: release/fish --no-config 'benchmarks/benchmarks/load_completions.fish'
-  Time (mean ± σ):      1.946 s ±  0.217 s    [User: 1.434 s, System: 0.558 s]
+  Time (mean ± σ):      1.601 s ±  0.139 s    [User: 1.186 s, System: 0.465 s]
 --
 Benchmark 1: release/fish --no-config benchmarks/benchmarks/math.fish
-  Time (mean ± σ):     861.8 ms ±  42.7 ms    [User: 788.8 ms, System: 73.8 ms]
+  Time (mean ± σ):     824.4 ms ±  97.6 ms    [User: 753.8 ms, System: 71.4 ms]
 --
 Benchmark 1: release/fish --no-config 'benchmarks/benchmarks/no_execute.fish'
-  Time (mean ± σ):     510.0 ms ±  24.9 ms    [User: 322.0 ms, System: 186.6 ms]
+  Time (mean ± σ):     518.5 ms ±  33.3 ms    [User: 318.7 ms, System: 198.6 ms]
 --
 Benchmark 1: release/fish --no-config benchmarks/benchmarks/printf-escapes.fish
-  Time (mean ± σ):      17.2 ms ±   2.5 ms    [User: 13.6 ms, System: 3.5 ms]
+  Time (mean ± σ):      18.6 ms ±   2.4 ms    [User: 14.7 ms, System: 3.8 ms]
 --
 Benchmark 1: release/fish --no-config benchmarks/benchmarks/printf.fish
-  Time (mean ± σ):     929.6 ms ± 109.9 ms    [User: 846.1 ms, System: 84.5 ms]
+  Time (mean ± σ):     816.9 ms ±  73.1 ms    [User: 742.9 ms, System: 74.9 ms]
 --
 Benchmark 1: release/fish --no-config benchmarks/benchmarks/read.fish
-  Time (mean ± σ):      1.076 s ±  0.015 s    [User: 0.430 s, System: 0.667 s]
+  Time (mean ± σ):      1.026 s ±  0.094 s    [User: 0.441 s, System: 0.612 s]
 --
 Benchmark 1: release/fish --no-config 'benchmarks/benchmarks/seq_echo.fish'
-  Time (mean ± σ):      68.1 ms ±   5.3 ms    [User: 56.7 ms, System: 11.5 ms]
+  Time (mean ± σ):      56.8 ms ±   6.4 ms    [User: 47.8 ms, System: 9.1 ms]
 --
 Benchmark 1: release/fish --no-config 'benchmarks/benchmarks/set_long.fish'
-  Time (mean ± σ):     508.9 ms ± 119.5 ms    [User: 490.5 ms, System: 18.9 ms]
+  Time (mean ± σ):     410.6 ms ±  17.9 ms    [User: 393.4 ms, System: 18.2 ms]
 --
 Benchmark 1: release/fish --no-config benchmarks/benchmarks/string-repeat.fish
-  Time (mean ± σ):     343.5 ms ±  10.2 ms    [User: 331.6 ms, System: 11.7 ms]
+  Time (mean ± σ):     341.5 ms ±  24.6 ms    [User: 328.2 ms, System: 13.0 ms]
 --
 Benchmark 1: release/fish --no-config benchmarks/benchmarks/string-wildcard.fish
-  Time (mean ± σ):     586.8 ms ±  12.6 ms    [User: 513.7 ms, System: 73.7 ms]
+  Time (mean ± σ):     589.3 ms ±  33.8 ms    [User: 513.5 ms, System: 76.6 ms]
 --
 Benchmark 1: release/fish --no-config benchmarks/benchmarks/string.fish
-  Time (mean ± σ):     974.6 ms ±  41.5 ms    [User: 885.0 ms, System: 94.1 ms]
+  Time (mean ± σ):     878.8 ms ±  29.2 ms    [User: 802.8 ms, System: 80.0 ms]
```

ridiculousfish:
I don't mind the implications of the change, but it is noisy - it touches every HashMap and HashSet. If this could be set globally I'd be in favor, but I don't like cluttering up the code without a clear, measurable benefit.

We should just use the default hash algorithm generally, and switch to a faster algorithm in any cases where profiling shows hashing to be a bottleneck.

Thomas-airmatrix:
Just some thoughts by a passer-by. 

I have found either always using std or always using ahash is nice because you don't have spend effort on making a choice each time. Although, a potential issue of choosing a non-std map globally is libraries that non-generically consume/produce std ones however I doubt that will be an issue for fish.

Would be cool to see some benchmarks for the global change though.

A potentially less noisy approach might be to use the provided aliases in the ahash crate. 
https://docs.rs/ahash/latest/ahash/type.HashMap.html

The std hashmap already has to be imported so the diff is mainly the import path. (Will still need the `new()` to `default()` as `new()` hardcodes the hasher).

```diff
- use std::collections::{HashMap, HashSet};
+ use ahash::{HashMap, HashSet};
```

Although, you then have to remember to auto import the right HashMap.

Note: Using `use ahash::{AHashMap as HashMap}` would allow using the `new()` method but that would cause 3 levels of newtype wrapping of generic structs:
 `AHashMap -> std::_::HashMap -> hashbrown::HashMap`
 
With the current structure of the code base you could also define an alias somewhere like `common` and then import from there. That would allow a global change in the future. Also since rust-analyzer's auto-import sorting prefers crate level imports  by default it'll import the correct one with an alias defined in common. However, if y'all ever split fish up into smaller crates in a workspace that might be less helpful.

mqudsi:
I mainly feel the same way about this as @ridiculousfish does (despite being the one that opened the issue). Since the runtime/build cost of the change is zero, I initially went for a blanket find-and-replace approach (not literally, since that actually doesn't compile) but need to be convinced by benchmarks that the churn is worth it because of the degraded contribution experience. It's "just" `HashMap` and `HashSet` right now, but we can easily end up in a case like the poor folk at Google or Meta (née Facebook) that basically can't compile a single line of code without using custom types for everything. Given that the most "one place to change things" solution in lieu of a genuine crate-level option is to `use ahash::HashMap as HashMap` in `src/lib.rs` then using `use crate::HashMap` instead of `use std::collections::HashMap` everywhere, it might indeed make more sense to see which maps/sets truly benefit from the change (though this is the extra work on behalf of the maintainers that I was hoping to avoid).

@Thomas-airmatrix:

I don't like the `ahash::HashMap` because it's not a type alias but a type wrapper (as you end up mentioning), and feel like for a type that offers an **almost** but not quite one-for-one API with the standard library it's not a great option when, thanks to rust type inference, the solution was just to add a second/third generic type parameter to each HashMap/HashSet decaration and then the rest of the uses *sort of* just work with type inference.

I think the crate-level use-indirection I mentioned above would be a more elegant solution than `use ahash::HashMap as HashMap` in file, but it's certainly still less than ideal.

One thing about this PR that I did not enjoy: while in most cases changing `struct Foo { map: HashMap<K,V> }` to `struct Foo { map: HashMap<K,V,AH>` and `HashMap::new()` to `HashMap::default()` did the trick (thankfully we no longer have to manually `use std::default::Default` in each file!), the fact that I couldn't change `HashMap::from(...)` to `HashMap::<_, _, AH>::from(..)` was not fun. I originally included `ahash` in `Cargo.toml` with `use-default-features = false` because I wasn't using the `ahash::HashMap` "alias" but ended up having to re-add the default features just to be able to use `ahash::HashMap::from(..).into()` instead (again necessitating the `.into()` call just because it is a newtype and not an alias).

faho:
So do we have any benchmarks here?

I don't think most of these are very hot, except for maybe the function set and the wildcard sets.

(of course the difference with e.g. the highlight maps is that you don't need *throughput* - it's plenty fast already because it's a purely interactive thing, you want reduced *resource use*)

faho:
Okay, our benchmark suite (plus some additions I've been working on) doesn't think this changes a lot.

Even if I try to specifically exercise this, like I run `builtin history search` a bunch, or make a completion with a huge list of conditions, some duplicated and some not, I get a result that's barely above system noise (1%-2%).

That means, because this complicates the code slightly, I would be disinclined to merge it.

Of course if you have any benchmarks to the contrary (in particular I haven't found a good way to benchmark highlighting or the reader history search), that's something to consider.

mqudsi:
I'm closing this out because it doesn't seem to net us any obvious across-the-board wins. I might revisit it later if/when I have more bandwidth (and either a faster machine or more patience to deal with release mode compile times).

PR #10505, opened 2024-05-18
completions: improved ip completions

exploide:
## Description

- complete routing table names/IDs
- rudimentary completions for `ip rule` subcommand

## TODOs:

### DIFF
--- a/share/completions/ip.fish
+++ b/share/completions/ip.fish
@@ -9,7 +9,8 @@ set -l ip_addr a ad add addr addre addres address
 set -l ip_link l li lin link
 set -l ip_neigh n ne nei neig neigh neighb neighbo neighbor neighbour
 set -l ip_route r ro rou rout route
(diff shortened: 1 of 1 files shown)


mqudsi (approved):
LGTM. Regexes are fine, though I think a positive lookbehind match like `string match -r '(?<=\stable\s)\S+'` is clearer in intent, but most of our completions use `string replace -r` so that's also fine.

I don't see anything that could be non-backwards-compatible or anything. The output you parse is very implementation-specific, but I don't think there are multiple implementations of `ip` besides the one  `iproute2`?

exploide:
>  I don't think there are multiple implementations of ip besides the one iproute2?

Good point. There is busybox implementation of `ip` and it seems to lack the `-d` option. Hence, it prints usage on stderr. Don't know if it is likely to run into this, but changing it to `ip -d route show table all 2>/dev/null` is probably cleaner.

Should I open another PR?

mqudsi:
Nice catch. If you want to open a PR, I'll merge it right away.

PR #10507, opened 2024-05-20
Reimplement printf library

ridiculousfish:
## Description

This reimplements printf (the library part, not the builtin), allowing us to drop the gnarly `printf-compat` crate.

The new `printf` crate lives at top level, in the fish workspace. fish itself only required minor changes to adopt it.

This is probably not feasible to review in detail, but printf-compat wasn't really reviewed either, so we're no worse off.

### Design

The new printf is based on musl - porting it from C to Rust, providing longer names and comments, and adding locale support. This is an improvement because the musl printf is widely used, and `printf-compat` is not, so it is more likely to be standards conforming.

There are other improvements:

- Locale support is direct instead of being tacked on later.
- No dependencies on libc printf. No unsafe code at all!
- You can `printf` to anything that uses `std::fmt::Write` such as `String`, `WString`, or a file - no more `WideWrite`.
- Rounding is implemented directly. The floating point rounding mode is respected.
- Supports `%n` (printf-compat did not).
- Better code organization.
- Significantly faster in most cases.

### Testing

This includes all of the printf-compat tests AND all of the musl tests. Naturally it passes all of them. This is substantially more test coverage than printf-compat has.

### Fun Fact

This uncovered a [musl printf buf, which got fixed](https://www.openwall.com/lists/musl/2024/04/12/1)!

### DIFF
--- a/doc_src/license.rst
+++ b/doc_src/license.rst
@@ -181,6 +181,8 @@ MIT License
 
 ``fish`` includes a copy of Alpine.js, which is copyright 2019-2021 Caleb Porzio and contributors, and licensed under the MIT License. It also uses FindRust.cmake, from the Corrosion project, which is copyright 2018 Andrew Gaspar and licensed under the MIT license. It also includes the Dracula theme, which is copyright 2018 Dracula Team, and the Nord theme, which is copyright 2016-present Sven Greb. These themes are also used under the MIT license.
 
+``fish`` contains code derived from  `musl-libc <https://www.musl-libc.org>`_, which is copyright 2005-2020 Rich Felker, et al., to implement printf. This code is used under the terms of the MIT license.
+
 The MIT license follows.
 
 Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
--- a/printf/src/arg.rs
+++ b/printf/src/arg.rs
@@ -1,0 +1,240 @@
+use super::printf_impl::Error;
+use std::result::Result;
+use widestring::{Utf32Str as wstr, Utf32String as WString};
+
+/// Printf argument types.
+/// Note no implementation of ToArg constructs the owned variants (String and WString);
+/// callers can do so explicitly.
+#[derive(Debug, PartialEq)]
+pub enum Arg<'a> {
+    Str(&'a str),
+    WStr(&'a wstr),
+    String(String),
+    WString(WString),
+    UInt(u64),
+    SInt(i64, u8), // signed integers track their width as the number of bits
+    Float(f64),
+    USizeRef(&'a mut usize), // for use with %n
+}
+
+impl<'a> Arg<'a> {
+    pub fn set_count(&mut self, count: usize) -> Result<(), Error> {
+        match self {
+            Arg::USizeRef(p) => **p = count,
+            _ => return Err(Error::BadArgType),
+        }
+        Ok(())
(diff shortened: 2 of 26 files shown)


doc_src/license.rst
```diff
@@ -188,3 +188,18 @@ Permission is hereby granted, free of charge, to any person obtaining a copy of
 The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
 
 THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+License for musl libc
+-----------------
+
+``fish`` contains code derived from  `musl-libc <https://www.musl-libc.org>`_ to implement printf. This code is distributed under the terms of the MIT license.
```
zanchey, line 195: There's already a copy of the MIT license directly above this, the notes about musl-libc can just be added to the paragraph describing the elements using MIT-licensed sources.

printf/src/arg.rs
```diff
@@ -0,0 +1,239 @@
+use super::printf_impl::Error;
+use std::result::Result;
+use widestring::{Utf32Str as wstr, Utf32String as WString};
+
+/// Printf argument types.
+///  Note no implementation of ToArg constructs the owned variants (String and WString);
+/// callers can do so explicitly.
+#[derive(Debug, PartialEq)]
+pub enum Arg<'a> {
```
mqudsi, line 9: Very minor nit, but can we keep the code structured the same way but make this `pub(crate)` and re-export it in `printf/lib.rs` as `pub using arg::Arg`? (Same with the traits)

I don't see the point in downstream users having to do `use printf::sprintf;` and `use printf::arg::Arg` separately from separate modules when you need `Arg` anyway.

doc_src/license.rst
ridiculousfish: fixed

printf/src/arg.rs
ridiculousfish: Turns out that if a type is `pub(crate)` it can't be re-exported as `pub` (Rust modules continue to confound) but your suggestion of adding `pub use` to lib.rs was good, I adopted it.

ridiculousfish:
We need printf support for builtin printf, and also for localized strings as you say - Rust's format strings are required to be compile-time, and the [available crates for runtime formatting don't look so great](https://www.reddit.com/r/rust/comments/13p3nnn/library_for_formatting_userprovided_format_strings/) but maybe one will do.

For other non-localized stuff it's fine to move from printf to native Rust formatting, though there's no urgency for that either.

faho:
> the [available crates for runtime formatting don't look so great](https://www.reddit.com/r/rust/comments/13p3nnn/library_for_formatting_userprovided_format_strings/) but maybe one will do.

We currently mostly (only?) use very simple features - our format strings are mostly `%ls: got %d arguments\n`, so the reddit OP's problems with justification and precision etc probably don't apply to us (right now).

> For other non-localized stuff it's fine to move from printf to native Rust formatting

I would prefer if we had a story for localized stuff first, so we don't end up switching formatting languages constantly.

----

Anyway, the PR at hand sounds like a good idea to me. I had a quick test and didn't find anything immediately wrong with it.

PR #10515, opened 2024-05-22
Try to add 'completions' label to PRs automatically

mqudsi:
The desired behavior is for a PR that modifies share/completions/*.fish and does not modify any other paths except ignoring changes to share/functions/*.fish gets a 'completions' label.

(There's another approach I was looking at first that can automatically set the 'fish next-3.x' milestone as well by doing all this manually via the api instead of using the labeler@v5 action, but it requires setting up GitHub secret tokens and I don't feel like doing that right now.)

I'm not sure how to test this 😅


zanchey:
One approach to testing is to use your fork - does mean you need to turn issues/PRs on temporarily but I think it can even be done in a private fork

mqudsi:
Trying that now. I have to merge this into my `master` branch because that name is hard-coded in the yaml file, though.

PR #10517, opened 2024-05-22
Automatically label and milestone completions

mqudsi:
This automatically assigns the 'completions' label and the 'fish next-3.x' milestone to completions-only PRs. Follow-up to https://github.com/fish-shell/fish-shell/pull/10515

A completions-only PR is defined as being one that touches `share/completions/*.fish` but does not touch any files outside of `share/`

We might want to update this to add an allow for `CHANGELOG.rst` in the future as well, but most PRs are without it because we didn't like how often that caused merged conflicts.

Tested on my own repo, you can see this sample PR that caused the GitHub actions bot to add the 'completions' label and the 'fish next-3.x' milestone label: https://github.com/mqudsi/fish-shell/pull/4

### DIFF
--- a/.github/workflows/autolabel_prs.yml
+++ b/.github/workflows/autolabel_prs.yml
@@ -1,0 +1,66 @@
+name: Auto-Label PRs
+
+on:
+  pull_request_target:
+    types: [opened, synchronize]
+
+jobs:
+  label-and-milestone:
+    runs-on: ubuntu-latest
+    steps:
+    # - name: Checkout repository
+    #   uses: actions/checkout@v2
+
+    - name: Set label and milestone
+      id: set-label-milestone
+      uses: actions/github-script@v7
+      with:
+        script: |
+          const completionsLabel = 'completions';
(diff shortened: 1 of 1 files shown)


krobelus:
sounds fine. We could also delete the `completions` label entirely, I don't think it's used.

This is what I use:

```
gh pr edit "$url" --add-label completions
gh pr edit "$url" --milestone "$milestone"
gh pr merge --rebase "$url"
```

mqudsi:
> We could also delete the completions label entirely, I don't think it's used.

We (ok, mainly @zanchey) use it to find completions for changelog fixes and release notes. Also for backporting changes to minor releases (though that can probably be done easier with a command line `git log -- 'share/completions/*'` filter).

If it's applied automatically, it also goes from being a chore to being an asset. If a completion PR is opened but it doesn't have the label, then someone did something wrong and it's touching more files than it should.

krobelus:
yeah I think `git log` is easier and more correct

PR #10521, opened 2024-05-24
Add `path basename --no-extension`

mqudsi:
Add `path basename --no-extension`

This makes `path basename` a more useful replacement for the stock `basename`
command, which can be used with `-s .ext` to trim `.ext` from the base name.

Previously, this would have required the equivalent of

    path change-extension "" (path basename $path)

but now it can be just

    path basename -E $path

## TODOs:

### DIFF
--- a/doc_src/cmds/path.rst
+++ b/doc_src/cmds/path.rst
@@ -8,7 +8,7 @@ Synopsis
 
 .. synopsis::
 
-    path basename GENERAL_OPTIONS [PATH ...]
+    path basename GENERAL_OPTIONS [(-E | --no-extension)] [PATH ...]
     path dirname GENERAL_OPTIONS  [PATH ...]
     path extension GENERAL_OPTIONS [PATH ...]
     path filter GENERAL_OPTIONS [-v | --invert]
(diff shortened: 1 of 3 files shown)


faho:
LGTM.

A more general solution would be a `path print` that you can give the parts you want (`path print --no-extension --no-dirname`), but this one is probably the most used combination, and it would be easy to allow nonsensical combinations (how do you print the dirname and the extension without the basename?).

mqudsi:
Only question I had was the short flag. I used the capitalized E in the sense that some utils use a capitalized version of a short flag to indicate the opposite (so no Extension). Other options were lowercase e, x, or something else. 

Another thought is that in the future it might make sense to add `-s .ext` to trim specific suffix like the standard `basename` so maybe `-S` for automatic version of that. 

Also if the long flag should be --no-ext or if that's something we should silently support but not advertise.

faho:
> I used the capitalized E in the sense that some utils use a capitalized version of a short flag to indicate the opposite (so no Extension). Other options were lowercase e, x, or something else.

-E seems fine to me. We don't currently have either `-e` or `-E` here. The closest we have is `--end`/`--entire` in *string*, both of which have a `-e` short flag.

`-x` has the issue that it's "executable" in `path filter`. Of course (like we see in string above), short flags can have multiple meanings, but if we wanted to add `--no-extension` to `path filter` we couldn't use `-x` there.

So `-E` is immediately fine, no collisions, and it also seems okay looking at future additions.

> Also if the long flag should be --no-ext or if that's something we should silently support but not advertise.

Because of our getopt, this *would* currently allow `--no-ext`. But I'm not a fan of having the canonical version be abbreviated like that given that we have change-extension.

> Another thought is that in the future it might make sense to add -s .ext to trim specific suffix like the standard basename so maybe -S for automatic version of that.

I don't believe I've ever needed the specific behavior of `basename -s`. I *have* needed "give me the extension", and I *have* needed "make whatever the extension is this instead", but never "remove '.xyz' if it ends in that".

And if I did I'd either match all .xyz files and work on them, or I'd use string to remove the .xyz extension specifically.

mqudsi:
The only reason for `basename -s` is to handle cases like `.tar.gz` where a naive `basename -E` would only strip one layer of extensions. But it's rare enough that I would just use `string replace`.

PR #10524, opened 2024-05-25
Explain `function --argument-names` in more detail.

t-nil:
## Description

`--argument-names` was very confusing for me. This should add enough clarification.

Tested on my system. Both HTML and man pages look good to me.


mqudsi (approved):
LGTM

PR #10528, opened 2024-05-25
feat: improve completion for konsole

yuntaz0:
## Description

Improve the existing completion for konsole, as the current version is outdated. Similar to the current version, the improved version only contains options from --help, but not from --help-all, which includes more Qt-specific options that are less likely to be used by users.
Additionally, it can complete --profile based on existing profiles, provided the profile name does not contain any space. (Please let me know if there is a better way to implement this. Thank you in advance.)

Fixes issue #

## TODOs:

### DIFF
--- a/share/completions/konsole.fish
+++ b/share/completions/konsole.fish
@@ -1,35 +1,36 @@
-# Completion for `konsole
+# Completion for konsole
 
-complete -c konsole -l help -d 'Show help and exit'
-complete -c konsole -s v -d 'Show version and exit'
-complete -c konsole -l version -d 'Show version and exit'
-complete -c konsole -l help-qt -d 'Show Qt specific options and exit'
-complete -c konsole -l help-kde -d 'Show KDE specific options and exit'
+complete -c konsole -f
+
+complete -c konsole -l help -xs h -d 'Show help and exit'
 complete -c konsole -l help-all -d 'Show all options and exit'
-complete -c konsole -l author -d 'Show author information and exit'
-complete -c konsole -l license -d 'Show license information and exit'
-complete -c konsole -l name -d 'Window class'
-complete -c konsole -l ls -d 'Login shell'
-complete -c konsole -s T -d 'Window name'
-complete -c konsole -l tn -d 'Terminal type'
-complete -c konsole -l noclose -d 'Do not close when command exists'
-complete -c konsole -l nohist -d 'Do not save history'
-complete -c konsole -l nomenubar -d 'Hide menu bar'
-complete -c konsole -l notabbar -d 'Hide tab bar'
-complete -c konsole -l notoolbar -d 'Hide tab bar'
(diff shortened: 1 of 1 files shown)


share/completions/konsole.fish
```diff
@@ -1,35 +1,36 @@
...
+complete -c konsole -l version -xs v -d 'Show version and exit'
+complete -c konsole -l author -d 'Show author info and exit'
+complete -c konsole -l license -d 'Show license info and exit'
+complete -c konsole -l desktopfile -d 'Base file name of desktop entry'
+
+set -l profiles (konsole --list-profiles)
+complete -c konsole -l profile -d 'Name of profile to use'
+for p in $profiles
+    complete -c konsole -n "__fish_seen_subcommand_from --profile; and not __fish_seen_subcommand_from $profiles" -a $p
+end
```
mqudsi, line 16: Instead of calling `konsole --list-profiles` when this completion is sourced (bad, hangs shell for some ms), this should be dynamic. No need for `__fish_seen_subcommand_from --profile` and `not __fish_seen_subcommand_from $profiles` because, if I understand it correctly, this is just a single mandatory argument to the long option `--profile`, so just complete it as a long argument `complete -c konsole -l profile -rxa (konsole --list-profiles)`

share/completions/konsole.fish
```diff
@@ -1,35 +1,30 @@
...
+complete -c konsole -l separate -l nofork -d 'Run in a separate process'
+complete -c konsole -l show-menubar -d 'Show the menubar'
+complete -c konsole -l hide-menubar -d 'Hide the menubar'
+complete -c konsole -l show-tabbar -d 'Show the tabbar'
+complete -c konsole -l hide-tabbar -d 'Hide the tabbar'
+complete -c konsole -l fullscreen -d 'Start Konsole in fullscreen mode'
+complete -c konsole -l notransparency -d 'Disable transparent backgrounds'
+complete -c konsole -l list-profiles -d 'List the available profiles'
+complete -c konsole -l list-profile-properties -d 'List all the profile properties'
+complete -c konsole -xs p -d '<P=V> Change profile property'
```
mqudsi, line 28: It would be nice to dynamically complete `konsole -p' by providing the options to it, perhaps `-xa "(konsole --list-profile-properties)="` (would need to be tested)

yuntaz0: The output of --list-profile-properties is in the form of `Property : Type`, such as `Path : QString`, so I'm thinking of using: konsole --list-profile-properties | sed "s/ : /=/g" to format the output to `Property=Type`

mqudsi: Well, it expects `key=<arbitrary value>` so putting it in `key=<type>` doesn't help. I would just use `string replace -fr '(.*?) .*' '$1='` to put it in the form and have the user provide the arbitrary value. Can you try that, does it work?

PR #10531, opened 2024-05-27
Add `set --keep` to avoid overwriting an existing variable

gahr:
## Description

This is my first time contributing to fish, so please point out any flaws in the procedure. Also, I would not consider myself a very power user of fish, so I might be missing something obvious, too.

This adds support for a `-k` or `--keep` option to the `set` command. When used in a `set` invokation that would set a variable to a value, the option turns the command into a no-op if a variable is already there.

I was looking at the changes in 3.7.1 after noticing a nice prompt in a fossil checkout of mine. This lead me to read the fossil prompt function and to stumble upon [this pattern](https://github.com/fish-shell/fish-shell/blob/7850142bef794db4d031fb6e1ec9dd7906a9f459/share/functions/fish_fossil_prompt.fish#L11-L47).

I thought there must be a better way... so here it is: `set -k -g foo bar` is (supposed to be) equivalent to `set -q foo or set -g foo bar`.

## TODOs:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -121,6 +121,7 @@ Scripting improvements
 - ``string repeat`` now allows omission of ``-n`` when the first argument is an integer. (:issue:`10282`)
 - ``functions --handlers-type caller-exit`` once again lists functions defined as ``function --on-job-exit caller``, rather than them being listed by ``functions --handlers-type process-exit``.
 - ``set`` has a new ``--no-event`` flag, to set or erase variables without triggering a variable event. This is useful e.g. to change a variable in an event handler. (:issue:`10480`)
+- ``set`` has a new ``-keep`` flag, to keep an existing variable from being overwritten: ``set -k -g foo bar`` is equivalent to ``set -q foo or set -g foo bar``. (:issue:`10531`)
 - Commas in command substitution output are no longer used as separators in brace expansion, preventing a surprising expansion in rare cases (:issue:`5048`).
 - Universal variables can now store strings containing invalid Unicode codepoints (:issue:`10313`).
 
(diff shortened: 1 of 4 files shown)


faho:
Hah, I had actually worked on something similar myself, but never found myself in the mood to go through with it.

I called it `set --default`. My big question was how it is supposed to interact with the *other* options:

- How should `set --local --keep foo bar` work?
- What about `set --append --keep foo bar` (and prepend)?

This, currently, will not alter the variable if it exists *in that scope*, otherwise it will create it with the new value.

So it's like

```fish
set --keep --local foo bar

# is

set -ql foo
or set -l foo bar
```

The design I came up with was that it would copy any existing values into the specified scope, so more like

```fish
set -q foo
and set -l foo $foo
or set -l foo bar
```

(or an equivalent `if`-block, except that would have issues with local scope)

With `--append` in contrast, it would append the new values to the old ones, no matter what scope. The alternative is to forbid append/prepend.

These together have the nice property that when you do `set --scope --keep`, you *know* that the variable exists in that scope after. Which is pretty useful for local scope, so you *know* you can modify it after without stepping on anyone's toes.

faho:
Alright, what I had was almost finished, I needed to add a single `if` check.

@gahr How do you feel about #10536?

gahr:
> Alright, what I had was almost finished, I needed to add a single `if` check.
> 
> @gahr How do you feel about #10536?

Sure, that works.

To me, copying an existing value from an outer scope and not overwriting an existing value sound like two orthogonal features... spontaneous question would be: how do I overwrite a local `foo` with the value of a global `foo`?

Also, I'm not a fan of `--default`, but I don't have a better alternative and naming things is hard...

Having said that, whatever works for you, works for me.

faho:
Alright, closing this one then.

PR #10536, opened 2024-05-29
Add `set --default`

### LINKED ISSUE #2504: Coalesce is too verbose?
According to this document there are [no options available](http://hyperpolyglot.org/unix-shells) and according to this stack-exchange discussion [it doesn't look to nice either](http://unix.stackexchange.com/questions/81421/how-do-i-reference-a-variable-in-fish-shell-with-a-default-fallback). The best I can think of something like:

```
set -q $MY_PROGRAM_ROOT; and set program_root ~/.local/share/my_prorgram; \
or set program_root $MY_PROGRAM_ROOT
```

Is there a better alternative available?

faho:
This allows quickly setting a variable to a value, or taking the existing value.

The simple use is this:

```fish
set --default foo bar
```

If $foo was already set, it stays. If $foo wasn't set, it is now set to "bar".

The trickier use is when it's combined with a scope or `--append` or `--prepend`. In those cases, it will ensure that:

1. The variable is now set with the given scope
2. If you use append/prepend, those values are included

These are nice properties to have to rely on.

In particular if you do `set --local foo bar`, you now have a *local* $foo, and so you can modify it without stepping on anyone's toes. If there was a global variable by that name, it still exists, but you've nicked the variable.

Alternative to #10531 
Fixes #2504 

## TODOs:

### DIFF
--- a/doc_src/cmds/set.rst
+++ b/doc_src/cmds/set.rst
@@ -12,7 +12,8 @@ Synopsis
     set (-f | --function) (-l | --local) (-g | --global) (-U | --universal) [--no-event]
     set [-Uflg] NAME [VALUE ...]
     set [-Uflg] NAME[[INDEX ...]] [VALUE ...]
-    set (-a | --append) [-flgU] NAME VALUE ...
+    set (-a | --append) (-p | --prepend) [-flgU] NAME VALUE ...
+    set (-d | --default) [-ap] [-flgU] NAME VALUE ...
     set (-q | --query) (-e | --erase) [-flgU] [NAME][[INDEX]] ...]
     set (-S | --show) [NAME ...]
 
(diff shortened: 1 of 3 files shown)


mqudsi:
I like the idea in principle but I'm leery of all the edge cases that might lead to some convoluted logic or violations of mental expectations.

I understand why you might want to initialize a local variable to some default value unless it already has a value in the global scope, in which case you wish to shadow that variable with a local copy of it so you can manipulate the local copy without affecting the global one (did I get this right?). 

But I think that it only makes sense with the scope flowing in one direction. In particular, I'm not sure the opposite makes sense (`set --default -g foo` stealing a value from a *local* `foo`).

faho:
> I'm not sure the opposite makes sense (set --default -g foo stealing a value from a local foo).

I mean if you have a local variable, you're gonna know about it and won't use set --default --global.

The logic isn't all that complicated - if the variable exists as you ask, keep it, if it exists with a different scope, copy it, if it doesn't exist, use the default.

The mental model is "after this the variable will exist in the scope I asked, either with the existing value or the default I gave", which I don't believe is too hard to understand?

mqudsi:
You might be acutely aware of all local variables but the same might not hold true if you set in the universal scope but have a global variable with the same name set. But that's not the only confusing part. Explicitly scoping with `set -gd foo` could also very reasonably be assumed to behave as it says on the tin: "set it in the global scope unless it already exists (in the global scope)". I can see the argument for doing it but I don't see how you can argue that it doesn't at the same time increase the complexity of the underlying scoping rules and logic considerably.

faho:
Okay, it appears there's no consensus here and I do not currently have the bandwidth to work on it, closing.

Issue #9117, opened 2022-08-06
math: Add scale argument to "round"


kseistrup:
The fish `math` functions seems to get it wrong sometimes, when doing integer math:

```sh
> math --scale=0 "22 / 5"         # this is all fine
4
> math --scale=0 "22 / 5 - 5"     # this ought to give -1
-0
> math --scale=0 "-5 + (22 / 5)"  # parentheses or reverting expression doesn't help
-0
```

This has been a problem for all fish3 releases, as far as I can tell, but I always thought it was an error in my scripts.

```sh
> fish --version
fish, version 3.5.1
> uname -a
Linux home 5.18.16-arch1-1 #1 SMP PREEMPT_DYNAMIC Wed, 03 Aug 2022 11:25:04 +0000 x86_64 GNU/Linux
```

faho:
Integer math is truncated, not rounded. The documentation explains this.

faho:
After a further look, the wording there is fairly confusing:

> A scale of zero causes results to be rounded down to the nearest integer. So 3/2 returns 1 rather than 2 which 1.5 would normally round to. This is for compatibility with bc which was the basis for this command prior to fish 3.0.0. Scale values greater than zero causes the result to be rounded using the usual rules to the specified number of decimal places.

So, yes, it does explain that it truncates via the example, but "rounded down" and "to the nearest" is somewhere between misleading and outright wrong. I've now tried to improve that.

Anyway, this was implemented as an awkward compatibility compromise, and so it probably needs to stay an awkward compatibility compromise, because I don't believe it's worth the cost of fixing it (because *finding* places where people even use integer math is hard, let alone figuring out which rounding they expect).

So this will just be how the cookie crumbles.

kseistrup:
I respect your decision, but finding a place where people use integer math is not hard:

* https://codeberg.org/kas/qtime/issues/5

And `bc` does get it right:

```sh
$ echo '22 / 5 - 5' | bc --scale=0
-1
```

As I said, I respect your decision, but the conclusion will be that using integer math in fish to calculate array indices cannot be done reliably and should be avoided.

kseistrup:
And PS:

> Integer math is truncated, not rounded.

That is what I expected in my calculations. `math 22 / 5` should give `4`, and `math 22 / 5 - 5` should give the same as `math 4 - 5`: `-1`.

faho:
>I respect your decision, but finding a place where people use integer math is not hard:

[...]

PR #10538, opened 2024-05-30
Add --scale-mode options for built-in math. Including "round"

Looouiis:
## Description

Add '--scale-mode' option. Incluing "round", "floor", "trunc", "ceil". The default mode is "trunc" and from now on we can freely choose the method to deal with --scale=0.

Fixes issue #9117 

## TODOs:


src/builtins/math.rs
```diff
@@ -132,7 +163,12 @@ fn format_double(mut v: f64, opts: &Options) -> WString {
     // As a special-case, a scale of 0 means to truncate to an integer
     // instead of rounding.
     if opts.scale == 0 {
-        v = v.trunc();
+        v = match opts.zero_scale_mode {
```
faho, line 166: So the way this works right now is that you can do `math --scale-mode=round`, and it will do absolutely nothing because you forgot to set `--scale=0` as well.

I would make it so setting scale-mode implies a scale of 0, and explicitly giving any *other* scale is an error (for now, it could potentially be changed later)

faho: An alternative is to have a special argument to `--scale`, something like `--scale=0-truncate`.

src/builtins/math.rs
```diff
@@ -62,6 +76,23 @@ fn parse_cmd_opts(
                     opts.scale = scale as usize;
                 }
             }
+            'm' => {
+                let optarg = w.woptarg.unwrap();
+                if optarg.eq(utf32str!("trunc")) {
```
faho, line 81: No need to make this short, the options should be "truncate" and "ceiling" ("floor" and "round" are fine, of course).

src/tests/math_tests.rs
```diff
@@ -0,0 +1,77 @@
...
+    let _cleanup = test_init();
+    let parser = &Parser::principal_parser();
+    let mut out = OutputStream::String(StringOutputStream::new());
+    let mut err = OutputStream::String(StringOutputStream::new());
+    let io_chain = IoChain::new();
+    let mut stream = IoStreams::new(&mut out, &mut err, &io_chain);
+    let mut ans_lst = String::new();
+
+    let mut argv = [
+        utf32str!("math"),
```
faho, line 34: I would usually add these tests as scripted tests in https://github.com/fish-shell/fish-shell/blob/46f6aa802489342b73782f61d7a669349b4874a3/tests/checks/math.fish instead.

That's more easily readable, because it is actually used in a script, and can skip all this setup gunk.

Cargo tests are fine for something that's not directly exposed in scripting, but this is, so that's the nicer place to test it.

Looouiis:
My bad, for not thinking carefully. I will fix them immediately when I got spare time.

src/builtins/math.rs
Looouiis: I'm not very familiar with this project and want to know if there are something like 'log level' in fish shell? If so, I want to make use of that.
Of course the default behavior is leaving error to case above. But it will also leave room for intended behavior and only throw warning or even nothing.

faho: > Of course the default behavior is leaving error to case above

I'm saying that `--scale-mode=foo` should automatically set `--scale=0`, and error out if you gave a non-zero scale value because that's nonsensical.

That's specifically an error in how it was called, and it should be an error.

> only throw warning or even nothing

Warnings aren't a good idea here.

We have [a logging system](https://github.com/fish-shell/fish-shell/blob/master/src/flog.rs) for fish-internal things, but it's not something you would use for builtins - which should mostly be treated as if they were standalone programs.

Looouiis: Got it

Looouiis:
I think it's a very little modification, so do we need to add change log for it? I have already drafted the doc and change log.

src/builtins/math.rs
```diff
@@ -1,17 +1,29 @@
...
 /// i.e. this is the first value such that x + 1 == x (or == x + 2, depending on rounding mode).
 const MAX_CONTIGUOUS_INTEGER: f64 = (1_u64 << f64::MANTISSA_DIGITS) as f64;
 
+enum ZeroScaleMode {
+    TRUNC,
+    ROUND,
+    FLOOR,
+    CEIL,
+}
+
```
mqudsi, line 21: I know we don't (yet) have a consistent naming policy across the crate, but we are trying to adopt rust conventions especially for new code. Could you please change these to be idiomatic (e.g. `ZeroScaleMode::Trunc`), and the same for any other all-caps non-`const` values/variables?

Looouiis: Got it

epidemian:
Why is this new `--scale-mode` option only defined when `--scale=0`?

If i run `math --scale=2 --scale-mode=trunc '2/3'`, i think it would be reasonable to print `0.66` instead of failing.

mqudsi:
I actually had the same question as @epidemian and was coming back to this PR to read through it and see if I had missed something :)

Looouiis:
> Why is this new `--scale-mode` option only defined when `--scale=0`?
> 
> 
> 
> If i run `math --scale=2 --scale-mode=trunc '2/3'`, i think it would be reasonable to print `0.66` instead of failing.
> 
> 
> 
> Defining it also for the rounding behavior of non-integer results would require less special-casing on the argument parsing logic  and on the docs :)

In my opinion, these scaling function is originally make for casting float value to integer. If user enable the scale mode. Math app should make the scale mode do their duty.

But for the convenience of using this math app. It make sense to allow user using scale mode when non zero scale is given. But what will math app do will depends on fish philosophy I guess.

By the way. I think automatically setting scale to zero may make user confused in some extreme situations. So I think we can inform user for some automatically behavior.

epidemian:
Playing around with this a bit, i've noticed that the rounding behavior of `math` is not consistent between the fractional part and the integer part of the number (on released version 3.7.1):

```fish
~ $ math --scale=2 200/3
66.67
~ $ math --scale=1 200/3
66.7
~ $ math --scale=0 200/3
66
```

The fractional digits get rounded, while the integer digits get truncated. Seeing the first two command outputs, i would've expected the third to be `67`.

So, assuming this behavior is intentional, maybe what i said about treating `--scale-mode` the same way for cases of `--scale` > 0 doesn't make much sense, as fish seems to already be treating `--scale=0` cases differently.

mqudsi:
What I would expect with this feature is that it unifies the handling for all scale modes.

epidemian:
> What I would expect with this feature is that it unifies the handling for all scale modes.

@mqudsi are you thinking about breaking backwards compatibility on how `math` rounds things? Or adding another "scale mode" that matches what `math` currently does (i.e., "round if it's a decimal place, but truncate if it's the integer part")?

I personally wouldn't mind either approach. If the latter is taken, maybe that mode could be called "default", and of course be the default.

mqudsi:
I'd be leery of changing the default behavior and breaking backwards compatibility but I wouldn't offer a mode=default option. Default if no mode is specified, but with an explicit mode it would do one of the already enumerated behaviors regardless of whether integral or decimal result.

mqudsi:
@Looouiis I noticed you pushed some commits - we don't get a notification when that happens! Is this branch ready for review again?

Looouiis:
@mqudsi I think its ready

mqudsi:
Thanks for working on this and revisiting it after the feedback! I'll review this today or tomorrow, though I do want to merge #10582 first as this will likely break tests under i586 again.

mqudsi:
Merged in 480d48351cfa52e897fed91da77ed7c10d96cc76

Thank you very much!

PR #10546, opened 2024-06-03
[RFC] [WIP] Remove fast-float dependency

mqudsi:
This is a strictly inferior option to shipping with the fast-float library given the packaging concerns some project members have raised. I have not run it through the full float pointing test suite. I have not optimized the code beyond reusing a dynamically allocated buffer per-thread. This was a quick attempt to see how feasible this option is, whether its performance is completely unacceptable, and what the temperature of the room is around this change.

**I do not know if this is a good idea. I am only checking to see if it is even an option to begin with.**

(That said, also bear in mind that our current `wcstod()` impl is not exactly a negligible wrapper around the fast-float simd logic to begin with.)


faho:
First benchmarks (vs 01e958d2f) are looking encouraging.

The math benchmark is slightly *faster* with this (~5%), the printf benchmark possibly very slightly slower (~1%). Both of which seem *fine*.

Varying the exact tests (longer numbers, more runs, more complicated expressions) doesn't seem to change the results much.

If it stays in that ballpark, doesn't have any awkward cliffs otherwise and is correct, that's a definitive win.

mqudsi:
I pushed a few commits with some minor optimizations and more tests to the branch. No changes to the logic were needed; it turns out the rules governing valid floating point literals are as simple as they seem. Tests from the `libc-test` project used by musl were ported to rust, specifically those from `strtod` and `sscanf` (but there weren't as many as I thought there would be).

I also made a minor tweak to `wcstod_underscores` to directly call `parse_partial_iter()` in case of inf/nan instead of calling into `wcstod_inner()` to avoid redoing a lot of the same work again.

faho:
Superseded by 20e9c9493ccaf79c2f0e2a9f70f0de86fec7e720

PR #10548, opened 2024-06-05
Update tutorial.rst

trishankatdatadog:
## Description

Link to some best practices for universal variables.

Fixes issue #

## TODOs:

### DIFF
--- a/doc_src/tutorial.rst
+++ b/doc_src/tutorial.rst
@@ -750,7 +750,7 @@ Now in another shell::
     > echo $EDITOR
     vim
 
-Best practices for using universal variables include `avoiding <https://fishshell.com/docs/current/language.html#universal-variables>`__ directly manipulating ``config.fish`` and ```.config/fish/fish_variables `, and preferring :doc:`funcsave <cmds/fish_add_path>`.
+Best practices for using universal variables include `avoiding <https://fishshell.com/docs/current/language.html#universal-variables>`__ directly manipulating ``config.fish`` and ```.config/fish/fish_variables``, and preferring :doc:`funcsave <cmds/fish_add_path>`.
 
 Ready for more?
 ---------------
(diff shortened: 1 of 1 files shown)


doc_src/tutorial.rst
```diff
@@ -750,6 +750,7 @@ Now in another shell::
     > echo $EDITOR
     vim
 
+Best practices for using universal variables include `avoiding <https://fishshell.com/docs/current/language.html#universal-variables>`__ directly manipulating ``config.fish`` and ```.config/fish/fish_variables``, and preferring :doc:`funcsave <cmds/fish_add_path>`.
```
faho, line 753: As it stands, this goes a bit deep immediately for the tutorial - which is supposed to be a very quick overview (and tbh could do with a bit of a rewrite).

How about this:

1. Link to the language docs on it ("See more on universal variables here")
2. Explain: "Universal variables only need to be set once, you don't need to put them in config.fish."
3. Leave off fish_add_path, funcsave (that's probably a mistake?), fish_variables

faho: Also don't link to the documentation via the URL - if this is installed locally it will go to the web, and this would *always* link to the current version of documentation. Link like

```
:ref:`link text here <universal-variables>`
```

instead.

trishankatdatadog: Agreed. WDYT?

doc_src/tutorial.rst
```diff
@@ -750,6 +750,7 @@ Now in another shell::
     > echo $EDITOR
     vim
 
+See more on universal variables :ref:`here <universal-variables>` (especially on how to set them once).
```
faho, line 753: This sounds weird to me - setting them once is easy, you just `set -U`.

The important part is that you only *need* to set them once.

How about:

```
You only need to set universal variables once, interactively. There's no need to add them to your :ref:`config files <initialization>`.

For more, see :ref:`Universal Variables <universal-variables>`
```

?

trishankatdatadog: Done! (With minor edits.) WDYT?

faho:
Thanks, merged!

PR #10569, opened 2024-06-14
docs: use fish_vi_key_bindings in the example section

chapeupreto:
## Description

In order to start using vi key bindings, `fish_vi_key_bindings` needs to be used.


ridiculousfish:
This should document default (emacs) key bindings, not vi. But good catch! Fixed in 4d300b7f3272b3914b969b0aee211fff9c5ce036

PR #10574, opened 2024-06-18
Extend diskutil eject to suggest volumes

### LINKED ISSUE #10573: disktutil eject autocomplete not showing volumes on MacOS
fish version 3.7.0 on arm64 Darwin latest versions installed using xterm-256color (iTerm2)

## Problem

The `diskutil eject <TAB>`autocomplete currently only shows disk devices, however the `diskutil eject` subcommand supports also ejecting mounted volumes:

```
❯ diskutil eject
Usage:  diskutil eject MountPoint|DiskIdentifier|DeviceNode
```

I would like to extend the auto completion to also offer ejecting volumes, maybe with the exception of the system volume `Macintosh HD`.

## Solution Approach

Extend https://github.com/fish-shell/fish-shell/blob/e95fc104ec60f5499ea69d2b7ae78a809eb8081e/share/completions/diskutil.fish#L54-L56 to also suggest volumes

schlomo:
Extend `diskutil eject` completion to also suggest mounted volumes.

Fixes #10573

## Description

image

## TODOs:

- [X] User-visible changes noted in CHANGELOG.rst (mentioned already improved completions

### DIFF
--- a/share/completions/diskutil.fish
+++ b/share/completions/diskutil.fish
@@ -52,8 +52,8 @@ complete -f -c diskutil -n __fish_use_subcommand -a umountDisk -d 'Unmount an en
 complete -f -c diskutil -n '__fish_diskutil_using_not_subcommand umountDisk' -a '(__fish_diskutil_mounted_volumes)'
 
 # eject
-complete -f -c diskutil -n __fish_use_subcommand -a eject -d 'Eject a disk'
-complete -f -c diskutil -n '__fish_diskutil_using_not_subcommand eject' -a '(__fish_diskutil_devices)'
+complete -f -c diskutil -n __fish_use_subcommand -a eject -d 'Eject a volume or disk'
+complete -f -c diskutil -n '__fish_diskutil_using_not_subcommand eject' -a '(__fish_diskutil_mounted_volumes) (__fish_diskutil_devices)'
 
 # mount
 complete -f -c diskutil -n __fish_use_subcommand -a mount -d 'Mount a single volume'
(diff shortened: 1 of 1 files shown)


ridiculousfish:
Merged as b3a16e0993e04eb31c85d34471306fa9866d66c1 - thank you!

zanchey:
This doesn't work for me; `stat` reports `drwxr-xr-x` but `os.access('/Volumes/some_disk_image', os.W_OK)` in Python reports False.

schlomo:
@zanchey as what user did you try this?

With me for example I plug in a USB thumb drive and the mountpoint actually belongs to my user:
image

zanchey:
Ah, it's because disk images are immutable. The mountpoint itself appears writable, but `touch /Volumes/some_disk_image` fails.

Having just mounted a network filesystem, a USB drive with Time Machine enabled, and a disk image, none of them appear in the output of `path filter -w /Volumes/*`. Only USB drives with plain filesystems are owned by and actually writeable by my users.

zanchey:
Annoyingly, the output of `mount` is pretty unhelpful - the image and the network filesystem appear as "mounted by myusername", but the Time Machine device and the plain USB drive do not.

schlomo:
Ah, good catch and I indeed didn't think about testing that. Can you suggest a better filter? Maybe simply exclude `Macintosh HD` or maybe show only real directories and exclude links?

Thank you @zanchey !

zanchey:
Might as well just match everything? `diskutil` sensibly refuses to eject "Macintosh HD". There's a `com.apple.TimeMachine.localsnapshots` which appears when Time Machine is backing up, but I didn't try to eject that, so I don't know if it's safe or not.

zanchey:
Made that change in 98fcb1c6b

BusyJay:
@zanchey The change doesn't work as `__fish_diskutil_volumes` is undefined.

Issue #10474, opened 2024-05-02
Test failures pre 3.8.0


zanchey:
The tests are failing on the package builders in a few places.

I'm filing an issue for posterity and also to make sure they actually get fixed.

1.  32-bit builds fail in src/tests/common.rs:127:

```
assertion `left == right` failed
  left: "-9223372036854775808"
 right: "0"
```
@KamilaBorowska spotted this one as the argument to sprintf being long, which is 32 bits on a 32-bit platform rather than defined as 64-bit.

2. 32-bit builds fail in src/wutil/wcstod.rs:575:9

```
thread 'wutil::wcstod::test::tests' panicked at 'assertion failed: `(left == right)`
 left: `Ok(8.925500000000001e-18)`,
 right: `Ok(8.9255e-18)`'
```
@mqudsi notes this is the wrong comparison to be doing as floating point values should be compared by epsilon

I don't know why these two aren't picked up by the 32-bit CI build.

3. Unwritable home directories make the history tests fail.

```
---- tests::history::test_history_formats stdout ----
thread 'tests::history::test_history_formats' panicked at src/tests/history.rs:561:32:
Failed to get data directory

---- tests::history::test_history_merge stdout ----
thread 'tests::history::test_history_merge' panicked at src/tests/history.rs:395:9:
assertion failed: history_contains(&everything, text)

---- tests::history::test_history_races stdout ----
thread 'tests::history::test_history_races' panicked at src/tests/history.rs:336:5:
assertion `left == right` failed
  left: 1
 right: 1025
```

HOME is set to something unwritable in the Debian package builds. This wasn't a problem pre-cargo test, because `tests/test_env.sh` was run beforehand. The cargo tests should be equally hermetic. It probably means reimplementing `test_env.sh` in Rust, because sourcing it before running `cargo test` stuffs up any rustup-installed toolchains.

mqudsi:
I pushed some updates that got tests working for me targeting `i586-unknown-linux-gnu` under Debian 12 i686.

mqudsi:
`test_history_formats` fixed in 476b360eb8f6492f891ab5edacfe3c7c08f934c6; everything should be passing now.

mqudsi:
@zanchey tests appear to pass with the following under Debian 12 i686, which I assume does the trick for you?

```sh
set fish_home (mktemp -d)
sudo chown root:root $fish_home
env HOME=$fish_home cargo test --target i586-unknown-linux-gnu
```

[...]

PR #10578, opened 2024-06-20
Run 32-bit CI tests as i586

mqudsi:
This *should* show the errors we keep running into as in #10474.


mqudsi:
Cool, seems to work. I'll fix the i586 issues then merge this.

PR #10582, opened 2024-06-21
Fix i586 floating point test failures

mqudsi:
This PR tries to find a balance between ignoring floating point tests/errors under i586 (i.e. x86 without SSE2) and ensuring that we still have somewhat reasonable results even when run with reduced accuracy. The printf unit tests are ignored under i586 but the littlecheck tests use prefix matching or reduced `math --scale` values to ensure that the results aren't ridiculous and the core preconditions are still being tested. The reduced accuracy tests are run under all platforms to ensure they don't rot, but the regular precision tests are only run if not i586.

This patch series includes the addition of a new and undocumented `status` builtin variants (`status target` and `status target-arch`) that are used to check at runtime if we should expect tests to run with reduced accuracy. There are obviously other ways to expose this information to the test environment, but I see a lot of (internal) use for this. 
(NB: I'm not saying "let's not document this" but rather just punting on the question of whether or not this will be a documented `status` interface for now.)

### DIFF
--- a/printf/src/tests.rs
+++ b/printf/src/tests.rs
@@ -342,6 +342,10 @@ fn test_ptr() {
 }
 
 #[test]
+#[cfg_attr(
+    all(target_arch = "x86", not(target_feature = "sse2")),
+    ignore = "i586 has inherent accuracy issues, see rust-lang/rust#114479"
+)]
 fn test_float() {
     // Basic form, handling of exponent/precision for 0
     assert_fmt1!("%a", 0.0, "0x0p+0");
@@ -448,6 +452,10 @@ fn test_float() {
 }
 
 #[test]
+#[cfg_attr(
+    all(target_arch = "x86", not(target_feature = "sse2")),
+    ignore = "i586 has inherent accuracy issues, see rust-lang/rust#114479"
+)]
 fn test_float_g() {
     // correctness in DBL_DIG places
     assert_fmt1!("%.15g", 1.23456789012345, "1.23456789012345");
@@ -583,8 +591,12 @@ fn test_prefixes() {
     assert_eq!(sprintf_check!("%ls", "cs"), "cs");
 }
 
-#[allow(clippy::approx_constant)]
 #[test]
+#[cfg_attr(
+    all(target_arch = "x86", not(target_feature = "sse2")),
+    ignore = "i586 has inherent accuracy issues, see rust-lang/rust#114479"
+)]
+#[allow(clippy::approx_constant)]
 fn negative_precision_width() {
     assert_fmt!("%*s", -10, "hello" => "hello     ");
     assert_fmt!("%*s", -5, "world" => "world");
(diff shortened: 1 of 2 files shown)


zanchey:
I've just realised that i586 is explicitly not supported on either [Debian 11](https://www.debian.org/releases/bullseye/i386/ch02s01.en.html) or [Debian 12](https://www.debian.org/releases/bookworm/i386/ch02s01.en.html). So I don't think we should particularly bother; I'll see if the OBS maintainers can set their architecture to i686 instead.

mqudsi:
The rest of the workarounds are already there, so I'm fine with merging this and punting on continuing to support i586 for later. It hurts my heart to sunset support for yet-another-target.

zanchey:
I don't think it's sunsetting a target that can be actually used, though. Nothing will boot on actual i586 machines - Debian has been i686-only since 2015, openSUSE requires a Pentium 4, RHEL has been 64-bit only since RHEL 7, Fedora is similar. i686 processors started coming out 28 years ago and i586 was only a small number of processors anyway.

I'm just confused as to why the OBS builds are producing i586 code. All the logs say the build target is `i686-unknown-linux-gnu`.

zanchey:
Oh, wow, OK. Upstream rustc defaults to `pentium4` CPUs for `i686-unknown-linux-gnu`, but Debian [overrides this to `pentiumpro`](https://salsa.debian.org/rust-team/rust/-/blob/debian/sid/debian/patches/behaviour/d-rustc-i686-baseline.patch) so that people can [use CPUs from 1997](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=973414).

Can I suggest moving the `status` subcommands to `__target` or similar?

mqudsi:
Good idea.

mqudsi:
Actually, if they patch `i686-unknown-linux-gnu` to use `pentiumpro` instead of `pentium4` then that means that merely checking for `i586` will fail because fish will have been ostensibly compiled against `i686-unknown-linux-gnu` but in a severely crippled form. Which explains why your test failures didn't have i586.

I'll rework the patch to check for `target-arch` as `x86` combined with a runtime test instead of merely checking for `target` prefixed with `i586`.

mqudsi:
Ok, that seems to have worked. What do you think?

ridiculousfish (changes requested):
Thank you for tackling this. I'm fine with skipping printf tests as you do, but we shouldn't be adding stuff to `builtin status` for this rather niche case.

I think the simplest way to fix the `math.fish` tests is to just use regular expressions. For example (using made-up numbers):

```
math "ln(e^0.1)"
# CHECK: {{(0.1)|(0.10000015}}
```

or whatever numbers. Or just skip the math test outright via `REQUIRES`.

mqudsi:
Sure, that's fine. I had preferred to be as precise as possible in determining whether we would accept a reduced accuracy result but I suppose we have enough coverage with the architecture-specific unit tests that we can afford to loosen our checks here (which are mainly to ensure the syntax parses and the correct order of operations is applied rather than to check precision and accuracy, to be honest).

I've pushed the simplified patch set.

PR #10585, opened 2024-06-23
give the two pages of the functions tab individual scrollbars

JoeIsCool:
## Description
Added individual scrollbars for each page of the functions tab in the web interface. The existing version of the page is awkward to use if the user has hundreds of functions or is looking at a function that is hundreds of lines long because the you must scroll both of them at the same time.
With the function list pane and the function body pane each having an individual scrollbar, a user can scroll down to the bottom of their functions list without scrolling down in the function body, which is more convenient if the function only has a few lines. Conversely, if they are examining a function that is hundreds of lines long, they can scroll through it without scrolling the function list.

## TODOs:

### DIFF
--- a/share/tools/web_config/fishconfig.css
+++ b/share/tools/web_config/fishconfig.css
@@ -54,7 +54,7 @@ body {
 }
 
 #tab_parent{
-		border: red 1px;
+    border: red 1px;
 }
 #tab_parent :first-child {
     border-top-left-radius: 8px;
@@ -99,13 +99,13 @@ body {
     margin-right: 12px;
     display: flex;
     flex-direction: row;
-    height: calc(100vh - 129px);
+    height: 90vh;
     overflow: hidden;
 }
 
 .function-list {
     text-align: right;
-    min-width: 200px;
+    min-width: 250px;
     font-size: 16pt;
     vertical-align: top;
     overflow-y: auto;
--- a/share/tools/web_config/fishconfig.css
+++ b/share/tools/web_config/fishconfig.css
@@ -53,6 +53,9 @@ body {
     background-color: #DDE;
 }
 
+#tab_parent{
+		border: red 1px;
+}
 #tab_parent :first-child {
     border-top-left-radius: 8px;
     border-left: none;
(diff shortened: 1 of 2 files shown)


share/tools/web_config/fishconfig.css
```diff
@@ -91,25 +94,25 @@ body {
...
 
-.master_detail_table {
-    display: table;
-    margin-top: 10px;
+.function_tab {
     margin-left: 12px;
     margin-right: 12px;
+    display: flex;
+    flex-direction: row;
+    height: calc(100vh - 129px);
```
mqudsi, line 102: Can we do without the hard-coded values here? I think we can find a workaround that doesn't involve subtracting the height; usually by adding another layer of div wrappers for everything.

mqudsi (changes requested):
Thanks, this is a nice improvement to the usability. The web layout has been neglected in comparison to the rest of the code, so this is good to see.

Aside from the height calculation comment I left, this seems to (with my list of functions, with their names being as wide as they are, etc, on my PC running under Firefox) introduce a horizontal scroll to the function names container that was not there without your patch.

It turns out this is because it the function list no longer scales in width with the length of your function names:

Before:
image

After:
image

I think it's reasonable to scale the width *up to a point* before introducing scroll bars (because we wouldn't want to increase it indefinitely), but also, I don't think this was intentional fallout from your change from `display: table` to `display: flex` and we'd want to address it.

share/tools/web_config/fishconfig.css
```diff
@@ -53,6 +53,9 @@ body {
     background-color: #DDE;
 }
 
+#tab_parent{
+		border: red 1px;
```
mqudsi, line 57: Spacing error here

share/tools/web_config/fishconfig.css
JoeIsCool: After looking at the css, I see this is the only child of an element with min-height:90hv, so in my new push, Iset height to 90hv

ridiculousfish:
Squash merged as 2ad4454e52e2bcdc9a2e97baee4c9ef92e1950ce. Thank you, great fix!

Issue #10580, opened 2024-06-21
Is possible to detect if an auto-suggestion is being displayed?


Susensio:
fish, version 3.7.1

I'm trying to create a conditional binding based on the current state of autosuggestion:
`\cl` should accept line if autosuggestion present, clear screen otherwise.

I've found this issue https://github.com/fish-shell/fish-shell/issues/5000 that addresses this topic.
It appears that testing commandline autosuggestion was once possible (https://github.com/fish-shell/fish-shell/commit/dc17869a84f671c0395a38a7bc6827a44d7d6e88) but this stopped working (https://github.com/fish-shell/fish-shell/commit/73998b81b4441d15ed039a46354eb9a93b8b8368)

What is the alternative? Is there any other way of knowing if autosuggestion is present on the commandline?

mqudsi:
I can't think of an obvious alternative for you, but I've opened #10586 that would implement this as `commandline --showing-suggestion`.

Issue #5000, opened 2018-05-19
Is possible to detect if an auto-suggestion is being displayed?


da99:
Is it possible to detect if an auto-suggestion is being displayed? For example:

    commandline --auto-suggestion

(The reason is I want to `forward-char` when no suggestion is being displayed, but use `forward-word` when an auto-suggestion is being displayed.)

krader1961:
Try this:

bind \cf 'set -l b (commandline --current-buffer); set -l c (commandline --cut-at-cursor); if test "$b" = "$c"; commandline -f forward-word; else; commandline -f forward-char; end'

Obviously you'll want to add the same binding for the right-arrow key. With that in place [ctrl-F] and [right-arrow] accept the autosuggestion one word at a time if the cursor is after the last typed character otherwise it moves forward one char. And you can still use [ctrl-E] to jump to the end of the line and accept the entire autosuggestion.

da99:
Once again, you save the day @krader1961. It works  _perfectly_. Should that be included in the [documentation for `commandline`](http://fishshell.com/docs/current/commands.html#commandline)? Something like:

   _-b or --current-buffer select the entire buffer,_ including any displayed autosuggestion _(default)_

Here is the code expanded:
```

function my_forward
  set -l buff   (commandline --current-buffer)
  set -l cmd    (commandline --cut-at-cursor)

  if test "$buff" = "$cmd"
    commandline --function forward-word
  else
    commandline --function forward-char
  end 
end  
```

PR #10586, opened 2024-06-24
Add `commandline --showing-suggestion`

mqudsi:
Returns 0 (true) in case an autosuggestion is currently being displayed.

This was first requested in #5000 then again in #10580 after the existing workaround for this missing functionality was broken as part of a change to the overall behavior of `commandline` (for the better).

~~I haven't written the documentation or put in much work besides adding it and checking it more or less works as I would like to see if anyone has any objections to this.~~

### DIFF
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -942,6 +942,18 @@ pub fn reader_execute_readline_cmd(parser: &Parser, ch: CharEvent) {
     }
 }
 
+pub fn reader_showing_suggestion(parser: &Parser) -> bool {
+    if !is_interactive_session() {
+        return false;
+    }
+    if let Some(data) = current_data() {
+        let reader = Reader { parser, data };
+        !reader.autosuggestion.is_empty()
+    } else {
+        false
+    }
+}
+
 /// Return the value of the interrupted flag, which is set by the sigint handler, and clear it if it
 /// was set. If the current reader is interruptible, mark the reader as exit_loop_requested.
 pub fn reader_reading_interrupted(data: &mut ReaderData) -> i32 {
--- a/doc_src/cmds/commandline.rst
+++ b/doc_src/cmds/commandline.rst
@@ -113,6 +113,10 @@ The following options output metadata about the commandline state:
     If it is, it would be executed when the ``execute`` bind function is called.
     If the commandline is incomplete, return 2, if erroneus, return 1.
 
+**--showing-suggestion**
+    Evaluates to true (i.e. returns 0) when the shell is currently showing an automatic history completion/suggestion, available to be consumed via one of the `forward-` bindings.
+    For example, can be used to determine if moving the cursor to the right when already at the end of the line would have no effect or if it would cause a completion to be accepted (note that `forward-char-passive` does this automatically).
+
 Example
 -------
 
--- a/src/builtins/commandline.rs
+++ b/src/builtins/commandline.rs
@@ -13,7 +13,7 @@ use crate::parse_util::{
 use crate::proc::is_interactive_session;
 use crate::reader::{
     commandline_get_state, commandline_set_buffer, commandline_set_search_field,
-    reader_execute_readline_cmd,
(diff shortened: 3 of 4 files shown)


mqudsi:
Added unit tests and documentation.

mqudsi:
Ok, we might have to skip the tests for this one. Under CI, they always fail because of how `commandline` is currently structured. It has no access to the actual `Reader` instance the shell is using, so the approach we've been taken is to add `reader_xxx()` functions to `reader.rs` which instantiate a *new* reader and then perform whatever test is needed against it.

In this case though, we construct a new reader and check if an autosuggestion is shown, but autosuggestions are generated in a background thread and so there's a race condition. No amount of `sleep` calls in the pexpect test itself can solve this - the issue is that rust-side we construct the Reader then test for an autosuggestion immediately afterwards, too quickly for an overloaded CI instance.

The correct approach of course would be to somehow make a `Reader` instance available for `commandline` to inspect so we are actually getting the same view the shell sees and not merely hoping to recreate the same view. The hack to get the tests to pass under CI is the following, possibly conditionally executed when a CI environment variable is found:

```diff
diff --git a/src/reader.rs b/src/reader.rs
index e008f8f83..c20143c23 100644
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -948,6 +948,8 @@ pub fn reader_showing_suggestion(parser: &Parser) -> bool {
     }
     if let Some(data) = current_data() {
         let reader = Reader { parser, data };
+        // HACK: try to work around CI issues <JUST FOR TESTING>
+        std::thread::sleep(Duration::from_millis(100));
         !reader.autosuggestion.is_empty()
     } else {
         false
```

but I think the best course of action here right now is to add the pexpect test but leave it commented out until the `Reader`/`commandline` relatioship is restructured down the line.

src/reader.rs
```diff
@@ -942,6 +942,28 @@ pub fn reader_execute_readline_cmd(parser: &Parser, ch: CharEvent) {
...
+        let reader = Reader { parser, data };
+        // Work around the fact that autosuggestions happen asynchronously and we don't share state
+        // with the actual Reader used by the shell, so while a suggestion might have already been
+        // populated at the tty, we still need time for our background thread to populate one here.
+        // (We are blocking in a call to `commandline --showing-suggestion` here. We could use a
+        // condition variable instead of this loop, but this is incredibly far from the hot path,
+        // never called in a stock fish installation, and expected to generally never end up
+        // actually calling yield_now() except under CI or on really slow machines.)
+        while reader.autosuggestion.is_empty() && !reader.in_flight_autosuggest_request.is_empty() {
+            std::thread::yield_now();
```
ridiculousfish, line 959: This won't have the intended effect - autosuggestions are determined in the background and are applied via a callback (see `autosuggest_completed`). Simply yielding to the scheduler won't cause this to run.

I suggest simply removing this unless something breaks without it, in which case we'll have to understand why.

ridiculousfish (changes requested):
I like the new feature but the `thread::yield` looks wrong to me.

mqudsi:
I don't like it either but I don't think it is worth complicating the code for. 

Ideally it would not be an async operation at all because command line would be able to access &Reader and query the *actual* and *current* state, not try to recreate it from scratch.

src/reader.rs
mqudsi: CI is broken without it.

ridiculousfish: Yes, the completion runs on the same thread - aka the "main" thread. 

The CI [passed for me without this loop](https://github.com/ridiculousfish/fish-shell/actions/runs/9831107931). Indeed if the loop were ever entered, it would be an infinite, hang which I was able to reproduce with some testing.

CI failures here are probably the normal runners-are-slow stuff which can be addressed by increasing the sleep in the tests.

Please remove this bit and up the sleep if necessary. Otherwise this looks good.

mqudsi: Thanks for checking it again. I can confirm it's passing now.

Issue #1842, opened 2014-12-02
vim text objects


tannhuber:
It would be nice to have [vim text objects](http://vimdoc.sourceforge.net/htmldoc/motion.html) in vi mode. See also [here](http://blog.carbonfive.com/2011/10/17/vim-text-objects-the-definitive-guide/).

leoauri:
It would be really nice to be able to `ci"` to change within quotes in vi mode...  One day!

krobelus:
@leoauri For now you can use `Alt-e` or `Alt-v` to edit the commandline in your `$EDITOR`.

PR #10593, opened 2024-06-29
Vi mode: support `%` motion and `ib`/`ab` text objects

nikitabobko:
## Description

I didn't add new tests, since I didn't find that fish has any existing tests for input functions.

It's my first time writing in Rust, so I might did some noob mistakes. I tried to stick with the existing code style as much as possible, but I might made mistakes.

I understand that I introduce a new API, which could be a reason to reject the PR.

The implementation is obviously not 100% the same as in VIM, but it should cover the major cases. ~~The biggest missing part is real matching of "matching brackets", I just scan the text to find next the first occurence of the bracket, but I think it's good enough.~~ (UPD: supported)

My commits don't cover `ci(`, `ci{`, `ci[`, and friends (The introduced `jump-to-matching-bracket` can't help here, because it would match any bracket). I think that if fish supported `jump-to-matching-char` (brackets match their respecitve brackets, regular chars match themselves), mentioned bindings could be added (by changing existing `b,i` binding). If you are ok with the proposal, I can implement it in the next PR

Makes progress in issue #1842

Thank you for checking the PR regardless of your decision! And thank for the great shell!

## TODOs:

- [ ] Tests have been added for regressions fixed. 
  It's not a regression. So I guess check?
- [x] I read `CONTRIBUTING.rst`

### DIFF
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -1739,7 +1739,7 @@ impl ReaderData {
         }
     }
 
-    fn jump(
+    fn jump_and_remember_last_jump(
         &mut self,
         direction: JumpDirection,
         precision: JumpPrecision,
@@ -1749,7 +1749,16 @@ impl ReaderData {
         self.last_jump_target = Some(target);
         self.last_jump_direction = direction;
         self.last_jump_precision = precision;
+        self.jump(direction, precision, elt, vec![target])
+    }
 
+    fn jump(
+        &mut self,
+        direction: JumpDirection,
+        precision: JumpPrecision,
+        elt: EditableLineTag,
+        targets: Vec<char>,
+    ) -> bool {
         let el = self.edit_line(elt);
 
         match direction {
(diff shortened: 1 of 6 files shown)


src/reader.rs
```diff
@@ -3098,17 +3107,69 @@ impl<'a> Reader<'a> {
...
                     self.input_data.function_set_status(success);
                 }
             }
+            rl::JumpToMatchingBracket | rl::JumpTillMatchingBracket => {
+                let (elt, _el) = self.active_edit_line();
+                let el = self.edit_line(elt);
+                let l_brackets = vec!['(', '[', '{'];
+                let r_brackets = vec![')', ']', '}'];
+                let position = el.position();
+                let cur_char = match c {
```
nikitabobko, line 3122: Alternative variable name: `jump_from`. This way `targets` can be renamed to `jump_to` to make the code symmetrical

nikitabobko: Applied the suggestion

nikitabobko:
> The biggest missing part is real matching of "matching brackets"

Actually, I think it's not that hard. I will implement real brackets matching

ridiculousfish:
Thank you!

ridiculousfish:
Merged as e03e5e116dca6afadf2cd72f399e0c5ac9268f65

PR #10594, opened 2024-06-29
don't error if only dependencies or devDependencies are present

KTibow:
## Description

`package.json` may not have both `dependencies` and `devDependencies`. right now, the completion fails due to a keyerror if this is the case. this fixes it by falling back to an empty array.

## TODOs:


mqudsi (approved):
LGTM. Thanks.

Issue #10595, opened 2024-06-30
Default completion for less doesn't include all options


visrosa:
In case it matters, I'm using fish 3.7.1 on Linux  6.5.0-41-generic #41~22.04.2-Ubuntu SMP PREEMPT_DYNAMIC Mon Jun  3 11:32:55 UTC 2 x86_64 x86_64 x86_64 GNU/Linux

I was  looking for the color output option for less with `less -[tab][tab]` since I didn't remember the exact spelling and I noticed that the autocompletion didn't list that one.

If I delete /usr/share/fish/completions/less.fish and regenerate from the man pages the options do get listed.

Is the non-exhaustive completion options by design? If this is not a feature, is a PR welcome? For `less` and whatever else I find missing completions? Any guidelines? I am aware some programs have different variants (gnu, busybox, etc) and will keep that in mind.

ridiculousfish:
#10596 - thank you!

PR #10596, opened 2024-06-30
Update less.fish

visrosa:
Preliminary work.  Might be important to check version if options I added aren't widely available.

Changed some short options to old-style options since they can't be grouped and don't even need spaces before their arguments, such as `less -ooutputfile` which creates `outputfile`.

The -Dxcolor argument is commented out because its arguments follow complex rules I didn't look into in depth

Fixes issue #10595 

## TODOs:


ridiculousfish:
Merged as 7260774fc38131c2a733838f7f9554eccde08c8f - thank you!

PR #10601, opened 2024-07-04
fish_vi_key_bindings: add bindings for semicolon and comma in visual mode

nikitabobko:
fish_vi_key_bindings: add bindings for semicolon and comma in visual mode

They are already presented in normal mode, and I presume were forgotten to be added in visual mode

## TODOs:

  No need to. It's expected to work out of the box
  It's not a regression (at least, I'm not aware of)
  I don't add it to `./CHANGELOG.rst` because it's a minor change that can be considered as a bug fix


nikitabobko:
The original commit that added `repeat-jump*` binds https://github.com/fish-shell/fish-shell/commit/1d68b52cbc3e64591383efc3a5a1150503e8c3b4

It only added them in normal mode, I think it was a simple oversight, and they were not added to visual by mistake

faho:
Thanks!

PR #10603, opened 2024-07-04
Restore terminal state better.

amosbird:
## Description

Restore terminal state better when running commands.

1. Only update TTY_MODES_FOR_EXTERNAL_CMDS when the command finishes successfully.
2. Pop a few entries as a best effort to disable Kitty's progressive enhancement.

## TODOs:

### DIFF
--- a/src/input_common.rs
+++ b/src/input_common.rs
@@ -484,7 +484,7 @@ impl Drop for TerminalProtocols {
         let sequences = concat!(
             "\x1b[?2004l",
             "\x1b[>4;0m",
-            "\x1b[<1u", // Konsole breaks unless we pass an explicit number of entries to pop.
+            "\x1b[<9u", // Pop a few number of entries as a best effort to disable kitty progressive enhancement.
             "\x1b>",
         );
         FLOG!(
--- a/src/input_common.rs
+++ b/src/input_common.rs
@@ -484,7 +484,7 @@ impl Drop for TerminalProtocols {
         let sequences = concat!(
             "\x1b[?2004l",
             "\x1b[>4;0m",
-            "\x1b[<9u", // Pop a few number of entries as a best effort to disable kitty progressive enhancement.
+            "\x1b[=0u", // Clear kitty progressive enhancement.
             "\x1b>",
         );
         FLOG!(
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -3666,8 +3666,10 @@ pub fn term_copy_modes() {
 }
 
 /// Grab control of terminal.
-fn term_steal() {
-    term_copy_modes();
+fn term_steal(copy_modes: bool) {
+    if copy_modes {
+        term_copy_modes();
+    }
     while unsafe { libc::tcsetattr(STDIN_FILENO, TCSANOW, &*shell_modes()) } == -1 {
         if errno().0 == EIO {
             redirect_tty_output();
@@ -4971,7 +4973,7 @@ fn reader_run_command(parser: &Parser, cmd: &wstr) -> EvalRes {
         );
     }
 
-    term_steal();
+    term_steal(eval_res.status.is_success());
 
     // Provide value for `status current-command`
     parser.libdata_mut().status_vars.command = (*PROGRAM_NAME.get().unwrap()).to_owned();


src/reader.rs
```diff
@@ -4971,7 +4973,7 @@ fn reader_run_command(parser: &Parser, cmd: &wstr) -> EvalRes {
         );
     }
 
-    term_steal();
+    term_steal(eval_res.status.is_success());
```
krobelus, line 4976: I'm not yet sure if we should do this but I guess the right criteria is "exited via signal"

src/input_common.rs
```diff
@@ -484,7 +484,7 @@ impl Drop for TerminalProtocols {
         let sequences = concat!(
             "\x1b[?2004l",
             "\x1b[>4;0m",
-            "\x1b[<1u", // Konsole breaks unless we pass an explicit number of entries to pop.
+            "\x1b[<9u", // Pop a few number of entries as a best effort to disable kitty progressive enhancement.
```
krobelus, line 487: we should never need to pop more than one set of flags because we never push more than one set at a time.
It seems like the motivation for changing this is to work around buggy child processes, which I'm not sure we should do. The problems are that we'd mask bugs in child processes, and we'd break nesting (by popping flags set by a parent process).

I'm not sure what's your motivating example but to give a concrete one, if I run `fish` and then run Kakoune from fish, then a crashing Kakoune will leak progressive enhancement flags. It seems appropriate to fix the leak in Kakoune (fish does that when it panics, which covers most cases).

amosbird: >  work around buggy child processes,

The goal is to address the issue of crashing child processes. This can be easily reproduced by running a child fish shell and then using `kill -9` to terminate it. After doing this, any newly created child processes will always have the kitty keyboard protocol enabled, regardless of whether they support it or not.

krobelus: any particular reason why your child process crashes? Usually it should be possible to clean up in a crash handler.

I think the expectation that a crashing child process not leak terminal state makes sense.
However we should not implement it by violating the protocol. Instead we could extend the [protocol](https://sw.kovidgoyal.net/kitty/keyboard-protocol/#progressive-enhancement) to maybe allow both getting the stack size and truncating the stack later

amosbird: > any particular reason why your child process crashes? 

"Crash" is a general term in this context. A more specific example is the loss of connection to a remote shell.

krobelus: right, when doing `ssh -> fish`, then a disconnect will leave the terminal in a bad state.
We could extend the protocol as described above and use that in SSH, to restore flags upon exit (assuming that flags are never leaked intentionally). Then it will work also if the local shell is not fish.
Until then, we could stop pushing protocols if `$SSH_TTY` is set but `$TMUX` (or equivalent) is not? Is there any other relevant scenario?

src/reader.rs
krobelus, line 4976: actually your suggestion of copying modes only on success makes a lot of sense.
Do you maybe have a concrete example of bad modes copied, was it something from ssh too?

src/input_common.rs
krobelus: Yeah actually I think the benefits of the stack are somewhat theoretical, so if we don't find a better solution, we should abandon the stack and set/clear flags directly.

I think the best case is if we can change SSH to clean up. I've openend https://github.com/kovidgoyal/kitty/issues/7603 to summarize my thoughts for a different audience.

As for gauging the importance of this issue. I think if `ctrl-p` stops working in bash that's pretty bad. The connection drop scenario is an everyday one, so we should fix it. Any crashes are not so important I think; those cases should be rare that it's reasonable to run `reset` or restart the terminal

src/reader.rs
amosbird: >  was it something from ssh too?
Yes.

krobelus: ok but was it something that is fixed by `term_fix_external_modes()`? Since that's been re-added in latest master.
It's not very clearly defined what we should do here so I'd prefer a concrete motivation (like `ECHO` not being set or similar)

amosbird: The `term_fix_external_modes()` function is incomplete. I attempted to manually restore additional options like `echoe` and `echok`. However, some options, such as `ignpar` and `ignbrk`, don't even exist in the rust libc package. Additionally, I'm unsure if more options need to be restored. Using `stty sane` results in a different set of options compared to the initialization of fish shell, which is why I opted for a different approach instead.

src/input_common.rs
mqudsi: I respectfully disagree, if we assume crashes don't happen then the bulk of the reasoning for managing terminal state ourselves goes away altogether.

krobelus: > if we assume crashes don't happen

we don't have numbers but the scenario of "SSH disconnect, then run a program that doesn't support CSI u" sounds like it could happen once in a while to the unsuspecting user.
At the same time, I'm still trying to find a tangible benefit for using the flag stack at all; it seems better to set/clear the flags directly, to fix the above scenario.

PR #10608, opened 2024-07-09
Return true if there was an autosuggestion to accept

diniamo:
## Description

Return true from `accept-autosuggestion` if there was an autosuggestion to accept, just like `suppress-autosuggestion`. 

## TODOs:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -155,6 +155,7 @@ Interactive improvements
 - ``read --help`` and friends no longer ignore redirections. This fixes a regression in version 3.1 (:issue:`10274`).
 - Measuring a command with ``time`` now considers the time taken for command substitution (:issue:`9100`).
 - ``fish_add_path`` now automatically enables verbose mode when used interactively (in the commandline), in an effort to be clearer about what it does (:issue:`10532`).
+- The `accept-autosuggestion` input function now returns true when there was something to accept.
 
 New or improved bindings
 ^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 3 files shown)


CHANGELOG.rst
```diff
@@ -155,6 +155,7 @@ Interactive improvements
 - ``read --help`` and friends no longer ignore redirections. This fixes a regression in version 3.1 (:issue:`10274`).
 - Measuring a command with ``time`` now considers the time taken for command substitution (:issue:`9100`).
 - ``fish_add_path`` now automatically enables verbose mode when used interactively (in the commandline), in an effort to be clearer about what it does (:issue:`10532`).
+- The `accept-autosuggestion` input function now returns true when there was something to accept.
```
mqudsi, line 158: Pedantically, I think the real change is that it now returns false if there *wasn't* something to accept.

mqudsi (changes requested):
I left a comment about changing the documentation to be "returns true when.." to "returns false when..", if you don't mind changing that.

It would also be nice for this to be accompanied by a test. You can probably model this directly off [the recent addition of a pexpect test](https://github.com/fish-shell/fish-shell/commit/936f7d9b8d3faeb49de4e617d76eaedabce09aaa) for `commandline --showing-suggestion`.

diniamo:
I'm really not sure how to write the test, so I would appreciate it if someone did it for me.

mqudsi:
In what context are you using this? Can you give an example of how you are invoking this input function? (relevant for how the status code is returned)

diniamo:
Thanks!

krobelus:
going through the changelog I realized we already have `commandline --showing-suggestion` which should be the canonical version (because it's more flexible) so I'll not changelog this. In fact we can completely get rid of `and` and `or` special bind functions, deprecate them and IMO remove them from docs.

PR #10612, opened 2024-07-10
Implement (and use) `string unescape --style=regex`

### LINKED ISSUE #10611: Implement `string unescape --style=regex`
When I added support for `string escape --style=regex`, I recall we intentionally did not implement `string unescape --style=regex` (despite the completions machinery currently offering it up as a valid option) with the rationale being that fish itself doesn't produce any regex-escaped content so there's no natural need to "un-regex" an expression or value.

This remains true, `string escape --style=regex` gets used because we naturally ingest PCRE2 regex expressions as part of `string`'s machinery/features but `string` doesn't generate PCRE2 syntax (with the exception of the output of `string escape --style=regex` itself). However, given that the shell's raison d’être is to make the user's life easier and because PCRE2 is more or less the lingua franca of a lot cli regex support, I think it might be worth adding this.

I _technically_ had need of this for the recently turbo-charged `npm`/`yarn` completions [as they now read literal regex expressions](https://github.com/fish-shell/fish-shell/commit/11be48fc38ef3f009f23445f00f1f836e132929d) out of the `names.json` that ships with `all-the-package-names`. Pedantically, the output of line 41 (below) *should* be piped into `string unescape --style=regex`, although I'm pretty sure npm naming conventions don't allow for any characters with special meaning in a PCRE2 context and so regardless of whether or not `all-the-package-names` PCRE2-escapes names before sticking them in the JSON file (and it presumably does), the output is almost certainly equivalent to the input (once you remove the quotes).

However, such machinery would be required for maximum correctness.

mqudsi:
Add support for `string unescape --style=regex`

    This complements existing support for `string escape --style=regex`.

    PCRE2's spec is rather extensive and supports a massive number of escape
    sequences, but we are only interested in those that are formed by escaping a
    literal string (i.e. not actually regexes).

    In addition to reversing the effects of `string escape --style=regex`, support
    for unescaping some common "in-the-wild" escape variants is also added. For
    example, `printf "hello\nworld\n"` is passed through unmodified by `string
    escape` but it's easy to imagine that a different regex escape library would
    encode that as the (fish-encoded) `hello\\nworld\\n` (escaping the new lines).
    We support both, since the goal here isn't to strictly unmap fish's `string
    escape` output (which would be of limited utility) but to also interact with
    regex expressions found in files or emitted by executables in the course of
    writing shell scripts (such as completions).

Closes #10611.

### DIFF
--- a/doc_src/cmds/string-escape.rst
+++ b/doc_src/cmds/string-escape.rst
@@ -26,7 +26,9 @@ Description
 
 **--style=regex** escapes an input string for literal matching within a regex expression. The string is first converted to UTF-8 before being encoded.
 
-``string unescape`` performs the inverse of the ``string escape`` command. If the string to be unescaped is not properly formatted it is ignored. For example, doing ``string unescape --style=var (string escape --style=var $str)`` will return the original string. There is no support for unescaping **--style=regex**.
+``string unescape`` performs the inverse of the ``string escape`` command. For example, doing ``string unescape --style=var (string escape --style=var $str)`` will return the original string. It assumes its input is valid and typically coming unmodified from the output of the matching ``string escape`` operation. If the string to be unescaped is not properly formatted, it is ignored.
+
+With regards to unescaping regex, in addition to reversing the effects of ``string escape --style=regex``, its counterpart ``string unescape --style=regex`` also unescapes *most* escape sequences that, in the most commonly used regex specs, *may* be escaped but are not in the current fish ``string escape`` implementation. For example, ``string escape --style=regex`` does not escape new lines or control characters and instead passes them through as-is (which is fine since they do not clash with any special regex syntax and *are* supported inputs to regex libraries), but to handle regex expressions formed from literal strings that were escaped outside of fish, ``string unescape --style=regex`` will also correctly detect and un-escape sequences like ``\n`` (literal backslash + n), mapping it to a new line. Note that ``string unescape --style=regex`` expects its input to be a string literal that has been previously escaped. Specifically, this means that it assumes all input is syntactically valid and well-formed regex and does not contain anything other than a literal regex pattern to match against (i.e. does not use special operators like ``|`` or ``(``/``)``, does not contain back references, does not make use of character classes, etc).
 
 .. END DESCRIPTION
 
--- a/share/functions/__fish_npm_helper.fish
+++ b/share/functions/__fish_npm_helper.fish
@@ -38,7 +38,7 @@ function __npm_filtered_list_packages
             test -f $path || continue
             # Using `string replace` directly is slow because it doesn't know to stop looking after a match cannot be
             # found in the alphabetically sorted list. Use `look` for its binary search support.
-            look '  "'(commandline -ct) $path | string replace -rf '^  "(.*?)".*' '$1'
+            look '  "'(commandline -ct) $path | string replace -rf '^  "(.*?)".*' '$1' | string unescape --style=regex
             break
         end
     end
--- a/src/common.rs
+++ b/src/common.rs
@@ -131,6 +131,13 @@ pub enum UnescapeStringStyle {
     Script(UnescapeFlags),
     Url,
     Var,
+    /// Unescaping a regex is confined to input that was formed strictly by escaping literal strings
+    /// to begin with.
+    ///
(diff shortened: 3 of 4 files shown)


doc_src/cmds/string-escape.rst
```diff
@@ -26,7 +26,9 @@ Description
...
...ing the effects of ``string escape --style=regex``, its counterpart ``string unescape --style=regex`` also unescapes *most* escape sequences that, in the most commonly used regex specs, *may* be escaped but are not in the current fish ``string escape`` implementation. For example, ``string escape --style=regex`` does not escape new lines or control characters and instead passes them through as-is (which is fine since they do not clash with any special regex syntax and *are* supported inputs to regex libraries), but to handle regex expressions formed from literal strings that were escaped outside of fish, ``string unescape --style=regex`` will also correctly detect and un-escape sequences like ``\n`` (literal backslash + n), mapping it to a new line. Note that ``string unescape --style=regex`` expects its input to be a string literal that has been previously escaped. Specifically, this means that it assumes all input is syntactically valid and well-formed regex and does not contain anything other than a literal regex pattern to match against (i.e. does not use special operators like ``|`` or ``(``/``)``, does not contain back references, does not make use of character classes, etc).
```
ridiculousfish, line 31: We should not ask our users to read a whole paragraph about `string unescape --style regex`. If we cannot say what it does succintly, then we should not add it.

src/common.rs
```diff
@@ -131,6 +131,13 @@ pub enum UnescapeStringStyle {
     Script(UnescapeFlags),
     Url,
     Var,
+    /// Unescaping a regex is confined to input that was formed strictly by escaping literal strings
+    /// to begin with.
+    ///
+    /// i.e. the assumption is that the input will never contain "functional" regex operators like
+    /// `[..]`, `(..)`, `|`, etc. and we will be just round-tripping escaped literals like `\.` or
+    /// `\\` back to `.` and `\`, respectively.
+    Regex,
```
ridiculousfish, line 140: I don't understand what "the assumption" means here. What if it is violated?

share/functions/__fish_npm_helper.fish
```diff
@@ -38,7 +38,7 @@ function __npm_filtered_list_packages
             test -f $path || continue
             # Using `string replace` directly is slow because it doesn't know to stop looking after a match cannot be
             # found in the alphabetically sorted list. Use `look` for its binary search support.
-            look '  "'(commandline -ct) $path | string replace -rf '^  "(.*?)".*' '$1'
+            look '  "'(commandline -ct) $path | string replace -rf '^  "(.*?)".*' '$1' | string unescape --style=regex
```
ridiculousfish, line 41: I'm not sure what's going on here - this looks like it was something hyper-specific to npm completions? We're already using `look` so maybe we can optimize this completion specifically without messing with builtins.

src/common.rs
```diff
@@ -789,6 +805,54 @@ fn unescape_string_url(input: &wstr) -> Option<WString> {
...
+            // It is invalid for the input to contain a trailing, unpaired `\\`
+            '\\' => result.push(match chars.next()? {
+                'a' => '\x07',                               // alarm
+                'c' => (chars.next()? as u8 & 0x1F) as char, // control codes
+                'e' => '\x1b',                               // escape
+                'f' => '\x0c',                               // form feed
+                'n' => '\n',
+                'r' => '\r',
+                't' => '\t',
+                // TODO: Consider implementing support for the following, noting that the likelihood
```
ridiculousfish, line 839: `string escape` should do the right thing. If we add what is described by the comments it will be a breaking change.

ridiculousfish (changes requested):
This seems motivated by optimizing npm completions, but I can't get behind it: it's not clear to me why we need to change a builtin here, especially with all of the caveats.

Can't we just optimize npm completions more directly?

mqudsi:
This was more about adding missing complementary functionality than anything else. npm completions already fly after I changed to use look. This would have just made that change more pedantically correct.

My todo comment was not suggesting any changes to string escape, which already does the right thing and works fine. 

I have a lot going on right now so I'm just going to close this because I don't have it in me to argue for this change.

ridiculousfish:
On a re-read my review came across as harsh. I did not intend that and I'm sorry.

PR #10622, opened 2024-07-20
Tidy up Cargo.toml to better support packaging

kjnsn:
I had a look at how fedora builds rust applications based on the documentation at https://docs.fedoraproject.org/en-US/packaging-guidelines/Rust/

The guidelines recommend not vendoring dependencies, and instead using the rust2rpm script to determine the correct deps from Cargo.toml.

I then tried to use rust2rpm, which was relatively successful. Fedora has packaged all dependencies required, except for rsconf. The available versions of `terminfo` and `errno` were slightly older than what was specified in Cargo.toml, so those have been downgraded to what is available. I am unsure if this will cause major issues, but likely the fedora packages are due for a version bump.

Also added some metadata into Cargo.toml, as it makes some automated tools really happy (like rust2rpm).

Addresses issue #10181

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,8 +1,6 @@
 [workspace]
 resolver = "2"
-members = [
-    "printf"
-]
+members = ["printf"]
 
 [workspace.package]
 rust-version = "1.70"
@@ -18,12 +16,16 @@ version = "0.1.0"
 edition.workspace = true
 rust-version.workspace = true
 default-run = "fish"
+license = "GPL-2.0-only AND BSD-3-Clause AND ISC AND LGPL-2.0-or-later AND MIT AND PSF-2.0"
+repository = "https://github.com/fish-shell/fish-shell"
+homepage = "https://fishshell.com"
+readme = "README.rst"
 
 [dependencies]
 pcre2 = { git = "https://github.com/fish-shell/rust-pcre2", branch = "master", default-features = false, features = ["utf32"] }
 
 bitflags = "2.5.0"
-errno = "0.3.9"
+errno = "0.3.0"
 lazy_static = "1.4.0"
 libc = "0.2.155"
 # lru pulls in hashbrown by default, which uses a faster (though less DoS resistant) hashing algo.
@@ -36,7 +38,7 @@ once_cell = "1.19.0"
 fish-printf = { path = "./printf" }
 rand = { version = "0.8.5", features = ["small_rng"] }
 widestring = "1.1.0"
-terminfo = "0.9.0"
+terminfo = "0.7.3"
 
 [dev-dependencies]
 rand_pcg = "0.3.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,7 +52,7 @@ once_cell = "1.19.0"
 fish-printf = { path = "./printf" }
 rand = { version = "0.8.5", features = ["small_rng"] }
 widestring = "1.1.0"
-terminfo = "0.7.3"
+terminfo = "0.9.0"
 
 [dev-dependencies]
 rand_pcg = "0.3.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,27 +16,16 @@ version = "0.1.0"
 edition.workspace = true
 rust-version.workspace = true
 default-run = "fish"
-# see also doc_src/license.rst
-# GPLv2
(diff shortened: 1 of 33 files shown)


Cargo.toml
```diff
@@ -18,12 +16,16 @@ version = "0.1.0"
 edition.workspace = true
 rust-version.workspace = true
 default-run = "fish"
+license = "GPL-2.0-only AND BSD-3-Clause AND ISC AND LGPL-2.0-or-later AND MIT AND PSF-2.0"
```
ridiculousfish, line 19: Can you please share how you got this? fish itself is released under GPL 2.0

kjnsn: It was pulled from https://src.fedoraproject.org/rpms/fish/blob/rawhide/f/fish.spec#_33

ridiculousfish: Ah interesting, thank you! Some of those dependencies are no more so we can rework that. Thank you for tackling this!

kjnsn: Had a quick check and updated this with an explanation of where things are coming from. Shoutout to the fedora packaging team for doing this initially :)

I'm unsure about `doc_src/python_docs_theme/`, as the files in there reference the BSD license, but `license.rst` mentions the PSF-2.0

zanchey:
@kjnsn is it possible for you to share the generated spec file? I will have to spin up another chroot to run it and I'm not sure when I'll be able to do that.

kjnsn:
@zanchey https://gist.github.com/kjnsn/834aee9d7e7a258dd8dd5fd91df8375c

Cargo.toml
```diff
@@ -18,12 +16,30 @@ version = "0.1.0"
...
+repository = "https://github.com/fish-shell/fish-shell"
+homepage = "https://fishshell.com"
+readme = "README.rst"
 
 [dependencies]
 pcre2 = { git = "https://github.com/fish-shell/rust-pcre2", branch = "master", default-features = false, features = ["utf32"] }
 
 bitflags = "2.5.0"
-errno = "0.3.9"
+errno = "0.3.0"
```
zanchey, line 42: This is probably ok. I assume that at 0.x there are essentially no guarantees of stability, so bumping a crate might cause issues elsewhere in the distribution?

zanchey (changes requested):
Thanks. I think the metadata changes here are good to include, the changes to the dependencies less so (see comments).

(I note that the generated specfile is wrong, although probably not completely broken. It is fine to merge this with further changes regardless.)

Cargo.toml
```diff
@@ -36,7 +52,7 @@ once_cell = "1.19.0"
 fish-printf = { path = "./printf" }
 rand = { version = "0.8.5", features = ["small_rng"] }
 widestring = "1.1.0"
-terminfo = "0.9.0"
```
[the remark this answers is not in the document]
kjnsn: Gotcha, done

src/curses.rs
```diff
@@ -10,6 +10,7 @@ use std::ffi::{CStr, CString};
 use std::path::PathBuf;
 use std::sync::Arc;
 use std::sync::Mutex;
+use terminfo::expand;
```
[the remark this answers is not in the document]
kjnsn: Yep, reverted

Cargo.toml
kjnsn: Ack. Unsure if this will cause issues with any other packages in fedora, but speaking from experience, packaging is hard and brittle in general.

zanchey: Yeah, I think the openSUSE approach (don't bundle every library as its own package) has some merit

Cargo.toml
zanchey: The docs theme is PSF-2 per https://pypi.org/project/python-docs-theme/2020.1/ but there are some Sphinx bits included from them that are BSD-licensed.

zanchey:
Sorry, I ended up down a rabbit hole of licensing review here. I've pushed 8b028c37e to bring the existing files up to date. I really, really hate having licensing information spread over three (and now four) files.

The licensing string can be changed to ""GPL-2.0-only AND LGPL-2.0-or-later AND MIT AND PSF-2.0", I think.

kjnsn:
Nice! Done

zanchey:
Merged, thanks!

PR #10627, opened 2024-07-24
Completion for `gem install --user-install` and `timeout`

tesuji:


faho:
Merged, thanks and sorry for the delay!

PR #10631, opened 2024-07-26
Upgrade fish_greeting.fish with terseness

marceloslacerda:
## Description

"Go fish" could also become the official slogan of the project

Fixes the lack of puns in the greeting function


zanchey:
Thanks, but no

Issue #10415, opened 2024-04-04
mipsel arch lacks AtomicU64 support


hdhoang:
Hi, I'm one of the users in #8653. I am trying to build master fish-shell again on [openwrt 23.05.3](https://github.com/openwrt/docker-openwrt/pkgs/container/sdk), basing on their `bottom` Makefile.

However, this CPU platform `mipsel` [doesn't support `AtomicU64` type](https://doc.rust-lang.org/core/sync/atomic/index.html#portability):

> PowerPC and MIPS platforms with 32-bit pointers do not have AtomicU64 or AtomicI64 types.

I attempted to `sed /AtomicU64/AtomicU32/` and fixed up call-sites as well, and [could build](https://github.com/openwrt/packages/issues/19386#issuecomment-2028616526) `fish, version a29cc8f-dirty`. Its line-reader was non-functional though. The cargo build command looks like this, for reference. (4.0.0 is a string I set in custom Makefile)

```bash
export PATH=$PATH:/builder/staging_dir/target-mipsel_24kc_musl/host/bin/:/builder/staging_dir/toolchain-mipsel_24kc_gcc-12.3.0_musl/bin/

[...]

PR #10635, opened 2024-07-28
Import portable_atomic::AtomicU64 when std does not provide it

hdhoang:
Congrats for getting close to a release!

I dedicate my contribution to the public domain.

## Description

Restores support for 32-bit powerpc and mips. I am using this build in an openwrt-rust-1.80 toolchain. I intend to add mipsel+rust-nightly check if it makes sense.

I will run a branch on top of faho's installer PR in a few days too.

Fixes issue #10415.

## TODOs:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -93,6 +93,7 @@ dependencies = [
  "num-traits",
  "once_cell",
  "pcre2",
+ "portable-atomic",
  "rand",
  "rand_pcg",
  "rsconf",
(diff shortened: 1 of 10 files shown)


hdhoang:
otoh, testing tier 3 platforms in CI would waste much credits

zanchey:
I'll let others decide on whether this is something we want to take, but I do want to note that we cannot take it as "dedicated to the public domain" - this is a boring legal thing but just complicates matters if contributions are not licensed to us under GPLv2.

mqudsi:
I'd rather not take the empty `.rustfmt.toml` at the moment, you can always just leave it in the repository but not commit it or add it to `.git/info/exclude` as a local `.gitignore` alternative.

But the changes themselves look good. It is unfortunate that crates.io doesn't allow shadowing of the system namespace so you could just use the same import path instead of needing to remember to use both. We could provide a type alias in the crate and then reference that everywhere, but it's still something you need to remember to use every time so it does seem like this is the kind of thing that CI would need to catch. (Unless we blacklist the import with clippy?)

hdhoang:
Thanks for the positive feedbacks, I have updated the main commit.

Regarding a lint, I drafted one at https://github.com/hdhoang/fish-shell/commit/f21ccd6621b4 which passes clippy on CI https://github.com/hdhoang/fish-shell/actions/runs/10332735112/job/28604284218. However, the openwrt rust I have on hand doesn't have a clippy entry.

Do you think this would be burdensome on mainstream platform support? Downstream packagers on the affected platform can escalate build failures if their users care more.

PR #10637, opened 2024-07-28
Fix `cd ..` to the root directory

### LINKED ISSUE #10636: `cd ..` to `/` confuses fish
on 4108306e4

```
~ % RUST_BACKTRACE=1 fish
~> cd /etc
/etc> cd ..
.>
```

Things are a little interesting here (see the PWD) but hitting <kbd>c</kbd> crashes one of the threads and puts fish into a very strange state where some keys work and others don't.

```
thread '<unnamed>' panicked at src/wutil/mod.rs:286:5:
Invalid working directory, it must start and end with /
stack backtrace:
   0: _rust_begin_unwind
   1: core::panicking::panic_fmt
   2: fish::path::path_apply_cdpath
   3: fish::path::path_get_cdpath
   4: core::ops::function::FnOnce::call_once{{vtable.shim}}
   5: core::ops::function::FnOnce::call_once{{vtable.shim}}
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
fish crashed, please report a bug. Debug PID 28696 or press Enter to exit
```

Xiretza:
The leading slash always needs to be present, even if there aren't any other components. This was introduced by the Rust port.

Fixes #10636

## TODOs:

### DIFF
--- a/tests/checks/cd.fish
+++ b/tests/checks/cd.fish
@@ -194,6 +194,12 @@ cd $old_path
 cd file
 cd $old_path
 
+# Test that going up to the root directory using .. works
+cd /(string split --no-empty -f 1 / (pwd))
+cd ..
+pwd
+#CHECK: /
+
 # cd back before removing the test directory again.
 cd $oldpwd
 rm -Rf $base
--- a/src/wutil/mod.rs
+++ b/src/wutil/mod.rs
@@ -319,8 +319,12 @@ pub fn path_normalize_for_cd(wd: &wstr, path: &wstr) -> WString {
     paths.extend(path_comps);
     let mut result =
         WString::with_capacity(paths.iter().fold(0, |sum, s| sum + s.len()) + paths.len() + 1);
-    for p in &paths {
-        result.push(SEP);
+    result.push(SEP);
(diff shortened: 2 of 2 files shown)


tests/checks/cd.fish
```diff
@@ -194,6 +194,12 @@ cd $old_path
 cd file
 cd $old_path
 
+# Test that going up to the root directory using .. works
+cd /(string split --no-empty -f 1 / (pwd))
```
Xiretza, line 198: Not too happy with this, but couldn't come up with a better way to make this work with arbitrary directories in `/`.

ridiculousfish: How about something boring like `for i in (seq 100); cd ..; end` ?

ridiculousfish:
Wow, great find!

ridiculousfish:
Might also consider adding a test for `path_normalize_for_cd`

tests/checks/cd.fish
Xiretza: Hmm, I can imagine failure cases where going to `..` in the root directory would normalize it again.

ridiculousfish:
Thank you!

PR #10647, opened 2024-08-03
Completion for `wg`

metamuffin:
## Description

Adds completions for wireguard's `wg` tool.

Completions for the `set` subcommand are a bit more permissive than the command, fixing that would (as far as I can see) require a function for parsing which is probably not worth it.

Completion improvements are already mentioned in CHANGELOG.rst so there is no changed needed there.


faho:
Merged, thanks!

PR #10649, opened 2024-08-04
Update xbps-install.fish

Zapeth:
`-R` or `--repository` require an argument which is in most cases a url, but can also refer to a directory path on the file system

## Description

Currently when trying to auto-complete the command with this flag, only package names are suggested, which are not valid arguments.

This change should fix that (note: I didn't test this, I'm just assuming it should work based on other completion examples and the documentation of the completion arguments)


faho:
Thanks!

PR #10655, opened 2024-08-07
Update `aura` completions

fosskers:
## Description

This PR updates the completions for Aura, of which I am the author.

## TODOs:
  - Already covered under:
  
> - Various new completion scripts and numerous updates to existing ones.

### DIFF
--- a/share/completions/aura.fish
+++ b/share/completions/aura.fish
@@ -5,7 +5,7 @@ set -l listall "(__fish_print_pacman_packages)"
 set -l listrepos "(__fish_print_pacman_repos)"
 set -l listgroups "(pacman -Sg | sed 's/\(.*\)/\1\tPackage group/g')"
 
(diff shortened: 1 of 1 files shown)


faho:
Merged, thanks!

zanchey:
@fosskers, we're very happy to ship these, but if you're writing them anyway would you prefer to keep them in the upstream repo?

fosskers:
I have the Bash and Zsh completions upstream; if I recall correctly, these Fish completions were submitted here by a user some years ago. If there's a way for Aura's install process to copy them to `/usr/share/fish/whatever/` etc., and have Fish auto-detect them, then I'm happy to do that do. It would reduce friction (albeit minor). I supposed I had assumed they _had_ to be here, for whatever reason.

zanchey:
`cargo install`'s [limited functionality strikes again](https://github.com/fish-shell/fish-shell/issues/10181)!

There's advantages and disadvantages both ways; if you make a new release you're not waiting the (often many) months for fish to be released to update the completions, for example. Alternatively, do both - the copy installed by the program gets precedence over the ones shipped by fish.

They can be installed to `XDG_CONFIG_HOME/fish/completions` for a local user install, or `$PREFIX/share/fish/vendor_completions.d` by a package manager.

Issue #10657, opened 2024-08-09
Completion for run0 command is missing


munzirtaha:
```
fish, version 3.7.1
Operating System: Arch Linux
```

Just like `sudo` we need the autocomplete for `run0`

PR #10662, opened 2024-08-10
Create run0.fish

munzirtaha:
Initial run0 completion which could later be improved.

Fixes issue #10657

### DIFF
--- a/share/completions/run0.fish
+++ b/share/completions/run0.fish
@@ -1,0 +1,21 @@
+#
+# Completion for run0
+#
+
+complete -c run0 -l no-ask-password -d 'Do not query the user for authentication for privileged operations'
+complete -c run0 -l unit -d 'Use this unit name instead of an automatically generated one'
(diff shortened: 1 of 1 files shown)


share/completions/run0.fish
```diff
@@ -0,0 +1,21 @@
+#
+# Completion for run0
```
ccoVeille, line 2: I have no idea what is run0

I would expect this comment to provide a link to the repository and a short explanation about what run0 is about

munzirtaha:
If I had found an explanation on sudo.fish, I would have done it on run0.fish ;)

It's a systemd-based, more secure replacement for sudo with no suid and a fancy red background. I can add
`# See: https://www.freedesktop.org/software/systemd/man/devel/run0.html`
but I think those kind of explanations for inclusion is better suited here than on that file.

ccoVeille:
Thanks for providing the context. It's clearer

Let's maintainers come back to you

faho:
Thanks, merged!

PR #10665, opened 2024-08-12
Extract yaml functions into a dedicated yaml place

kjnsn:
This extracts all yaml related functions into a yaml module, extracting them from the file specific logic from history. This also adds some basic tests for some of the functionality.

Why: some of this logic is relatively complicated and delicate, and is worth testing. The yaml functions also work exclusively with strings & HistoryItem, not files or reading/writing. It is relatively easy to test functions that work on strings & structs.

Potential improvements: where should this file go? Should some functions belong somewhere else?


src/history/file.rs
```diff
@@ -234,320 +236,3 @@ fn read_zero_padded(file: &mut File, mut dest: &mut [u8]) -> std::io::Result<()>
...
-fn unescape_yaml_fish_2_0(s: &[u8]) -> Vec<u8> {
-    // Safety: we never read from this box; we only write to it then read what we've written.
-    // Rationale: This function is in a very hot loop and this benchmarks around 8% faster on
-    // real-world escaped yaml samples from the fish history file.
-    //
-    // This is a very long way around of writing `Box::new_uninit_slice(s.len())`, which
-    // requires the unstablized nightly-only feature new_unit (#63291). It optimizes away.
-    let mut result: Box<[u8]> = std::iter::repeat_with(std::mem::MaybeUninit::uninit)
-        .take(s.len())
-        .map(|b| unsafe { b.assume_init() })
```
Xiretza, line 281: Preexisting, but this is 100% UB. The safety comment doesn't actually address the safety requirements of `assume_init()`, which are totally unrelated to whether or not the value is "read". Needs fixing.

cc @mqudsi

faho:
I have to confess I am not a big fan of this.

These are functions that are called in one place, so moving them out really just makes more files.

Unless anyone else disagrees, I would reject the move. The tests are fine.

src/history/file.rs
mqudsi: @Xiretza https://github.com/fish-shell/fish-shell/commit/6c37103b7c4715e47f9ab5492c1db0dd8f8e19d8

kjnsn:
> I have to confess I am not a big fan of this.
> 
> These are functions that are called in one place, so moving them out really just makes more files.
> 
> Unless anyone else disagrees, I would reject the move. The tests are fine.

Ack, no worries :)

ccoVeille:
PR was closed, but shouldn't the changes about the tests being kept and pushed maybe in another PR

src/history/file.rs
krobelus: It seemed fine to me, I don't think it's UB if it never happens.

Xiretza: What do you mean by "it never happens"? This called `MaybeUninit::uninit().assume_init()` `s.len()` times.

krobelus: The values are never read, so the values don't matter. It's true that we are cheating/lying to `MaybeUninit` but that's fine because the destructor of `u8` is trivial.

Xiretza: No, it's not. The documentation for `MaybeUninit` *specifically* calls out that `MaybeUninit::<i32>::uninit().assume_init()` is UB, the same is true for `u8`.

kjnsn: My view here: I think the concern is regarding the potential for UB, i.e. code paths that the compiler cannot guarantee are safe. The code in it's current state does not result in undefined behavior, as the values are never read. However it does feel fragile to me, as this code could change in the future, and it's not immediately obvious to me that reading at a certain point could be unsafe.

krobelus: Okay you're right we were UB according to the terms of `MaybeUninit`. There are no practical consequences to violating this contract but yeah it should be written differently

PR #10667, opened 2024-08-13
Improve apk completions for apk 3.x

triallax:
## Description
Improve completions for apk 3.x.

- -q silenced warnings in apk 2.x but not in in 3.x, so redirect stderr
  to /dev/null to avoid seeing warnings while completing (-q is still
  passed to `apk search` as it strips package versions and releases)
- Drop `-q` from `apk info`, as on apk 3.x it prevents apk info from
  outputting anything at all

I've tested these changes on both Chimera Linux (which uses apk 3.x)
and Alpine Linux (which is still using 2.x).

## TODOs:

### DIFF
--- a/share/completions/apk.fish
+++ b/share/completions/apk.fish
@@ -1,10 +1,10 @@
 # Completions for apk (Alpine Package Keeper)
 
 # Package name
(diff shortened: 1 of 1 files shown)


faho:
Thanks!

PR #10681, opened 2024-08-22
Add `delta` completions

khipp:
## Description

This PR adds completions for [delta](https://github.com/dandavison/delta).

## TODOs:

### DIFF
--- a/share/completions/delta.fish
+++ b/share/completions/delta.fish
@@ -1,0 +1,1 @@
+delta --generate-completion fish | source


faho:
On my system, `delta --generate-completion fish` takes 7ms. So I would remove the caching and let it be loaded at runtime. That avoids any potential issues with caching and saves some disk space and IO - it's not gonna be faster to check the file and load it.

faho:
Thanks!

PR #10685, opened 2024-08-26
fix: add Ghostty to terminals that support vi cursors

nfejzic:
## Description

There's a new terminal [Ghostty](https://mitchellh.com/ghostty) that is currently in closed beta but should go public until the end of this year. Ghostty supports shell integration features and custom cursors. However, the fish shell did not setup cursors for `vi` key bindings correctly, basically skipped Ghostty because of unknown `$TERM`.

Ghostty sets the `$TERM` to `xterm-ghostty`, so I just added it to the list in fish functions. This is my first contribution to fish shell, so I'm not really familiar with the code base. In any case, please let me know if I need to do some more changes for this.

I didn't create an issue for this, so no issue to be linked to. 

## TODOs:


nfejzic:
Thanks for merging the PR this quickly! 

> That's fine. When there's nothing really to discuss before implementing, an issue doesn't help - all it does is make us click more buttons to label/close/etc.

Makes sense! Thank you, good to know. 🙏👍

PR #10686, opened 2024-08-26
Add OpenTofu completions

nihaals:
## Description

This is based on the existing Terraform completions with minimal changes. [OpenTofu](https://opentofu.org/) is a fork of Terraform and still uses "terraform" in some places like directory names for some amount of compatibility so keeping some mentions of Terraform in the script doesn't seem like a huge stretch and just means there's a smaller diff between them when maintaining them (although from now on they are diverging so there probably won't be many shared changes). I'm not sure if this is the best way to have 2 completions that are so similar at the moment though.

## TODOs:

### DIFF
--- a/share/completions/tofu.fish
+++ b/share/completions/tofu.fish
@@ -1,0 +1,212 @@
+# Returns 0 if the command has not had a subcommand yet
+# Does not currently account for -chdir
+function __fish_terraform_needs_command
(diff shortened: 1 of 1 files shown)


faho:
Yeah, this seems fine, given that I understand the fork was about a license change and therefore they're bound to diverge. Merged, thanks!

PR #10693, opened 2024-08-30
fish_vi_cursor: Remove terminal checks

faho:
We keep having to extend these with new terminals, and I can no longer find a terminal that fails this.

Even emacs' ansi-term can now at least reliably ignore the sequence.

## TODOs:

### DIFF
--- a/share/functions/fish_vi_cursor.fish
+++ b/share/functions/fish_vi_cursor.fish
@@ -10,60 +10,6 @@ function fish_vi_cursor -d 'Set cursor shape for different vi modes'
         return
     end
 
-    # If this variable is set, skip all checks
-    if not set -q fish_vi_force_cursor
-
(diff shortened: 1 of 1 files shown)


share/functions/fish_vi_cursor.fish
```diff
@@ -86,7 +32,7 @@ function fish_vi_cursor -d 'Set cursor shape for different vi modes'
               if not set -q \$varname
                 set varname fish_cursor_unknown
               end
-              $function \$\$varname
+              __fish_cursor_xterm \$\$varname
```
krobelus, line 35: LGTM, the reason why I didn't remove these checks is simply that I didn't look into what the cursor control sequences look like.
Looks like they use CSI (and have a space in the middle, oddly) which we already send.

PR #10695, opened 2024-09-01
Add `word` and more, to `delete-*` and `backward-delete-*`

sgtm0113:
## Description
Add fllowing input functions:
- `backward-delete-bigword`
- `backward-delete-path-component`
- `backward-delete-word`
- `delete-bigword`
- `delete-word`

## TODOs:
- [x] Add new items to `enum ReadlineCmd`
- [x] Add mapping about new functions to `INPUT_FUNCTION_METADATA`
- [x] Add descriptions about new functions to `doc_src/cmds/bind.rst`
- [ ] I have not found tests relevant to input functions

(Something may be wrong because I'm not so good at writing both of Rust and English.)


ridiculousfish:
We have these bindings already - replace `delete` with `kill`. For example `backward-kill-bigword`

The only difference is how they interact with the kill ring - is that the intention of the PR?

sgtm0113:
> The only difference is how they interact with the kill ring - is that the intention of the PR?

Yes, I wanted to not modify kill ring.

faho:
Yeah, I'm inclined to close this.

faho:
Alright, it appears we don't want to take this, closing.

Sorry!

PR #10698, opened 2024-09-03
add completions for `git diff --color-moved` and `--color-moved-ws`

m4rch3n1ng:
## Description

add completions for `git diff --color-moved` and `git diff --color-moved-ws`.

see https://git-scm.com/docs/diff-options#Documentation/diff-options.txt---color-movedltmodegt.

didn't add descriptions yet, but i can add them if requested.

## TODOs:


mqudsi (changes requested):
No need to add descriptions for the individual completion values, but please use `-a 'opt1 opt2 opt3'` instead of each on a new line if there are no `\tdescription` after them as it is easier to refactor.

Otherwise looks good. Thanks for the contribution.

m4rch3n1ng:
done

mqudsi (approved):

PR #10700, opened 2024-09-05
vi mode: support increment/decrement of numbers

DevAtDawn:
## Description

did a full rewrite for #8320, use ctrl+x ctrl+a to increase/decrease a number
Fixes issue #8320

"Support incrementing/decrementing the number below the cursor
Vim supports incrementing & decrementing the number below the cursor (or
after it) via Ctrl-a and Ctrl-x, respectively. Given fish's Vi mode
support, it makes sense to provide similar functionality when working on
the command line, to provide a more natural environment for Vim users.
With this change we add the necessary functionality."

## TODOs:

### DIFF
--- a/share/functions/fish_vi_key_bindings.fish
+++ b/share/functions/fish_vi_key_bindings.fish
@@ -123,6 +123,10 @@ function fish_vi_key_bindings --description 'vi-like key bindings for fish'
 
     bind -s --preset -M insert ctrl-n accept-autosuggestion
 
+    # increment or decrement numbers with ctrl+x ctrl+ain vi mode
+    bind -s --preset -M default \ca 'increment_decrement add1'
+    bind -s --preset -M default \cx 'increment_decrement subtract1'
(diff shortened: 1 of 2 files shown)


faho:
Okay, I'm sorry, I started reviewing this and then I noticed that I would rewrite almost every single line.

Particular issues:

- You're using a ton of external commands when `string` would do
- You're using global variables to pass along state
- You're modifying these global variables in functions named "is_number"
- You're overusing small helper functions - instead of checking "$is_neg", set by "is_number", *just do the check there*

I'm afraid this would have to be completely rearchitected again before it's acceptable to merge. I'm going to close this, sorry.

DevAtDawn:
no problem, it was more of a working prototype that could inc/dec given various tokens/buffers+workarounds, i defs realize it  needs to be refactored/improved, I was more curious about wether it successfully acted like vim in each scenario vs/before refining the code

Issue #8571, opened 2021-12-22
Feature? Pipe to commandline


DevAtDawn:
<!--
fish, version 3.3.1

![image](https://user-images.githubusercontent.com/86164137/147042533-253ee3cd-1dbd-4d36-b601-9dff18ddf236.png)

In order to place the output of a command like "which python" on the commandline buffer (like fzf ctrl+t), I need to re-run it as "commandline -a (which python3)", but I wanted a way to easily pipe the output to "commandline -a" after pressing up arrow, ex: "which python3 | commandline -a", so I made these functions:

function clia
    read -l line; commandline -a $line
end

function fzfa
    fzf | clia
    # cat | fzf | clia
end

![image](https://user-images.githubusercontent.com/86164137/147046385-1cea2f01-a272-408b-8e82-ad7a2bfd4f4f.png)

Now I can run the simplified "tldr tar | fzfa" to select and append a command to the buffer, where normally, id need to do something like: "https://medium.com/xster-tech/get-output-of-last-command-on-fish-shell-7697bc532ca6".

Would it possible to implement a native way that allows pipeing directly to "commandline -a", ex: commandline --pipe", or does this feature exist/is there already a way to do this? - Thanks

ridiculousfish:
`commandline` doesn't know how to read from stdin today. I think it would be OK to add a flag here like `--stdin`, but it's also pretty easy to implement this with a function as you showed.

DevAtDawn:
Ok thanks, I can work on adding a commandline --stdin flag/feature, what would the short name be?

ridiculousfish:
Probably we can omit a short option for this since I expect it to be infrequently used.

vctls:
I'm surprised this hasn't been requested more than once.  
I was hoping of working around this with xargs, but I suppose it can't work in this context.

PR #10701, opened 2024-09-05
 Feature: Pipe to commandline 

DevAtDawn:
## Description

feature for  Pipe to commandline  #8571, added function fish_commandline_stdin

Fixes issue #8571 

## TODOs:


faho:
Sorry, this is *another* thing to use, what we need is to add the actual reading to the commandline builtin itself.

It should also handle multiple lines, which this won't.

PR #10702, opened 2024-09-05
clearer completions on test builtin

abelcha:
## Description

`test -e` (File exists) is true for both regular files and a directories, which might lead to tragic errors on a very common use case

changed the completion wording to use "Path exist" "Path is a Regular file" "Path is a directory" etc

This is how its already described in the official fish tutorial
https://github.com/fish-shell/fish-shell/blob/79a9c0e0f2160a0fb3fe0a974f62d2bfdd6fb7a6/doc_src/tutorial.rst?plain=1#L521-L525

Also do you think it's enviseagable to propose new features to test (the builtin)  ? just having long flags could be helpful  I don't know about all the posix shennanigans

Fixes issue #

## TODOs:


faho:
Merged, thanks!

> Also do you think it's enviseagable to propose new features to test (the builtin)

The test builtin is probably the single biggest mistake in fish's design, so probably not.

It needs to be replaced with something better, that meshes better with fish's design, and that requires *incompatible* changes. Adding more on to it makes it harder to replace.

PR #10711, opened 2024-09-08
Install fish-* man pages in default MANPATH

krobelus:
fish adds ~/.local/share/fish/man to its MANPATH for builtins etc.  But pages
like fish-doc are unambiguous so it seems like they should be accessible
from outside fish by default.


faho:
This looks fine to me.

PR #10713, opened 2024-09-10
Fix `man` completions on macOS with symlinked `manpath`

9999years:
When `manpath` prints a symlink to a directory, `/usr/libexec/makewhatis` ignores the entire directory:

```
$ /usr/libexec/makewhatis -o /tmp/whatis \
    (/usr/bin/manpath | string split :)
makewhatis: /Users/wiggles/.nix-profile/share/man: Not a directory
```

This means that the built-in `man` completions will not index any commands in these directories.

If we pass the directories to `readlink -f` first, `makewhatis` correctly indexes the `man` pages.

```
$ /usr/libexec/makewhatis -o /tmp/whatis \
    (/usr/bin/manpath | string split : | xargs readlink -f)
```

## Description

Talk about your changes here.

Fixes issue #

## TODOs:

### DIFF
--- a/share/functions/__fish_apropos.fish
+++ b/share/functions/__fish_apropos.fish
@@ -40,7 +40,7 @@ if test $status -eq 0 -a (count $sysver) -eq 3
 
         if test $age -ge $max_age
             test -d "$dir" || mkdir -m 700 -p $dir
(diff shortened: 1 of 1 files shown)


ridiculousfish:
Thank you!

PR #10716, opened 2024-09-11
docs: Update synopsis for 'set'

mattmc3:
- Include --export flag
- Include --prepend flag
- Make -Uflg order consistent

## Description

I noticed that the synopsis for `set` wasn't complete nor consistent. The synopsis is useful as a quick TLDR;, so I updated it with a more useful blurb.

## TODOs:

- [ ] Tests have been added for regressions fixed (N/A???)


mattmc3:
Sounds good. The missing `--prepend` was the original reason I started the PR, but I agree having `--export / --unexport` represented differently, and on its own line is better. I took a crack at it, and force pushed, but feel free to go your own way too if I didn't get it right. I don't know how inclusive the synopsis should be, but `--path/--unpath` usage could also be added:

```
set (--path) (--unpath) [NAME ...]
```

faho:
Thanks!

PR #10720, opened 2024-09-14
improved vi keybind. Very minor touch-up to match original vim keybinds.

Houdiee:
A keybind was missing so I added it.


ridiculousfish:
Thank you!

PR #10722, opened 2024-09-14
Add completions for: `termux-*`

EmilyGraceSeville7cf:
## Description

Add completions for: [`termux-*`](https://wiki.termux.com/wiki/Termux:API).

Fixes issue #

## Features

- [x] Positional and option arguments are completed in cases documentation clearly states what valid values are
- [x] Default values for switches are suggested
- [x] File completion is disabled where appropriate

## Progress

- [x] Completions are created for all Termux commands where help page includes at least one positional argument or switch
- [x] Completion descriptions are standardized
- [x] Code is refactored

## TODOs:

### DIFF
--- a/share/functions/__fish_termux_api__complete_group_ids.fish
+++ b/share/functions/__fish_termux_api__complete_group_ids.fish
@@ -1,0 +1,15 @@
+function __fish_termux_api__complete_group_ids
+    set -l command termux-notification-list
+    set ids ($command | jq --raw-output '.[].group')
+    set titles ($command | jq --raw-output '.[].title')
+
+    for index in (seq (count $ids))
+        set -l id $ids[$index]
+        set -l title $titles[$index]
+
+        test -z "$id" && continue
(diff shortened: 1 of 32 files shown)


share/functions/__fish_termux_api__complete_group_ids.fish
```diff
@@ -0,0 +1,15 @@
+function __fish_termux_api__complete_group_ids
+    set -l command termux-notification-list
+    set ids ($command | jq --raw-output '.[].group')
+    set titles ($command | jq --raw-output '.[].title')
```
faho, line 4: `jq` is not guaranteed to be installed, so you would have to check if it is available first.

It's also pretty slow - we've sped up our json reading by using `python` instead.

And this $command is useless, please just call the command directly.

faho (changes requested):

faho:
Alright, I'm going to be closing these because there are unaddressed concerns.

PR #10728, opened 2024-09-17
fish_indent: Cleanup all *.fish files not in tests

mattmc3:
## Description

I am in the process of [writing a linter](https://github.com/mattmc3/fishcheck) for Fish scripts, and of course the first thing to check with a linter would be if the Fish project itself comes out clean. Obviously, the first step of that is to run `fish_indent` (aka: dogfooding). Test should be exempt because they support various formatting choices because they are... you know... tests of stuff.

This PR is the result of running:

```
fish_indent -w (string match --invert --entire /tests/ ./**/*.fish)
```

A couple of questions:
1. Is there any reason not to run `fish_indent` against this repo regularly/automatically?
  a. Do we trust fish_indent to be mature enough to not have/cause edge cases that impact the project?
  b. Do we care about anything from a maintainability standpoint (like alignment) that `fish_indent` fights us on?
2. Does running this belong in a github action as part of CI?

## TODOs:


faho:
Thanks!

PR #10738, opened 2024-09-21
add token-cmds

chapmanjacobd:
## Description

This adds four fish commands for working with command-line arguments at the token level:

    bind \e\[1\;5C forward-token
    bind \e\[1\;5D backward-token
    bind ctrl-backspace backward-kill-token
    bind ctrl-delete kill-token

    bind \e, kill-selection backward-kill-token yank yank

    echo 1\ 2 '3 4' "5 6"
    echo t t  t   t    t     t     t  
	for i in ( seq 4 )
	  echo $i
	end

Fixes issue #2014

## TODOs:


krobelus:
> But for some reason bind \b isn't working for me in master.

Not 100% sure what exactly your problem was.
Master interprets `\b` (or `\x08`) as `ctrl-h` (it gets translated automatically)

```
$ bind \b 'echo foo'
$ bind \b
bind --preset ctrl-h backward-delete-char
bind ctrl-h 'echo foo'
```

and the backspace key can be mapped independently, using `bind backspace ...`.

Maybe `\b` and `\x08` should translate to `backspace` instead (similar to what we do for `\t`, `\e` and `\r`), since it's already obvious how to map `ctrl-h`?
I think I'll do that.

chapmanjacobd:
Ah interesting... yes, if I bind ctrl-h then ctrl-Backspace also works

But when I try `bind -k backspace backward-kill-token` that binds to backspace, not ctrl-Backspace

krobelus:
> But when I try `bind -k backspace backward-kill-token` that binds to backspace, not ctrl-Backspace

well that's intended, it's the difference between `bind backspace` and `bind ctrl-backspace`

> People are likely already using \b in their configs to mean ctrl-h because that is what fish_key_reader has said the key is.

Right but they might also be using it to mean `ctrl-backspace` since that also used to map to `\b` here.
(an actual `backspace` maps to `\0x7F` but it might also be `\b` depending on VERASE).

In short, user's bindings that rely on ambiguous keys will no longer work because the ambiguity is gone.

> In 3.7.1 it seems like \b is mapping to ctrl-h.

To clarify, `\b` is just a short-hand notation for `\x08` or `\ch`. They're all the same thing, so bind can't really treat them differently.
When we read this byte from an old-school terminal, it might come from either `ctrl-h` or `backspace`.

> but it might be good to fix fish_key_reader to print ctrl-h if that is the actual key being read

I think we already do? Maybe I missed something.

src/reader.rs
```diff
@@ -4908,9 +5021,11 @@ fn command_ends_paging(c: ReadlineCmd, focused_on_search_field: bool) -> bool {
         | rl::KillInnerLine
         | rl::KillWord
         | rl::KillBigword
+        | rl::KillToken
         | rl::BackwardKillWord
         | rl::BackwardKillPathComponent
         | rl::BackwardKillBigword
+        | rl::BackwardKillToken
```
krobelus, line 5028: I think the existing commands that operate on tokens are only`transpose-words` and `history-token-search-{forward,backward}`

src/reader.rs
```diff
@@ -3329,6 +3377,69 @@ impl<'a> Reader<'a> {
             }
         }
     }
+
+    fn _backward_token(&mut self) -> Option<(usize, EditableLineTag, usize)> {
```
krobelus, line 3381: why the underscore?

src/reader.rs
```diff
@@ -3329,6 +3377,69 @@ impl<'a> Reader<'a> {
...
+        let start_buff_pos = el.position();
+        let boundary = el.len();
+        if start_buff_pos == boundary {
+            return None;
+        }
+
+        let buff_pos = el.position()
+            + el.text()[el.position()..]
+                .chars()
+                .take_while(|c| !c.is_alphanumeric())
```
krobelus, line 3424: I don't think alphanumeric is right because it treats `_` differently for example.
Can we not parse without this offset and then use `prev_tok` instead of `tok`?

krobelus:
fixed the misleading `\b` in fish_key_reader

src/reader.rs
chapmanjacobd: hmm I copied most of this code from `transpose_words` but I'll try replacing `tok` with `prev_tok` instead of this and see if it still works

src/reader.rs
chapmanjacobd: the function felt a bit specific to `ForwardToken` and `KillToken`. 

I'll remove the underscore

I'm also open to removing the functions and inlining the code in the four functions but AS-IS helps to make the differences between `ForwardToken` and `KillToken` more obvious

src/reader.rs
chapmanjacobd: If I remove `let buff_pos` and `self.update_buff_pos` from both functions...

```
        let buff_pos = el.position()
            + el.text()[el.position()..]
                .chars()
                .take_while(|c| !c.is_alphanumeric())
                .count();

        self.update_buff_pos(elt, Some(buff_pos));
```

Everything still works but when the cursor starts at a space then it only moves forward a single space

But you are right. The behavior in this case:

```
 _ '3 4'
```

is undesired. `ForwardToken` moves from `_` to `'`. 

I think we could replace `!c.is_alphanumeric()` with `c.is_whitespace()`. I will do that

lgarron:
I love this, I would be so glad to see it land! 🤩

I tried out a build of this branch and I thought I'd report some inconsistent behaviour, though: `backward-token` and `forward-token` traverse command substitutions differently backwards vs. forward. Here's a screencast exhibiting three things with a command substitution at the end of a line, in stereotypical `for` loop code:


```shell
for i in ( seq 4 )
  echo $i
end
```


https://github.com/user-attachments/assets/c64ae72c-4c4c-4150-bc95-56e898ebe1f5


- When the cursor is at the beginning of the next line, `backward-token` jumps to the start of the command substitution.
- When the cursor is just after the end of the command substitution, `backward-token` jumps to the start of the final command of the command substitution.
- When the cursor is just outside of the command substitution of `backward-token`  jumps just past the end of it.

I think:

1. This current inconsistency would the feature rather confusing to work with.
2. Skipping command substitutions would make this feature significantly less useful, as I may very well want to use these bindings to edit them without having to click or dance around.

I think that either of the following would make more sense:

### Normal character

`(` are treated like any other character argument when traversing. For example, if `👇` represents the cursor:

- `👇(seq 4)` + `forward-arg` → `(seq👇 4)`
-  `(seq 4)👇` + `backward-arg` → `(seq 👇4)`
- `👇( seq 4 )` + `forward-arg` → `(👇 seq 4 )`
-  `( seq 4 )👇` + `backward-arg` → `( seq 4 👇)`
  
### Special character
  
`(` are treated as their own token when they are syntactically delimiting a command substitution (but not when they are part of an argument, e.g. in a file name):

- `👇(seq 4)` + `forward-arg` → `(👇seq 4)`
-  `(seq 4)👇` + `backward-arg` → `(seq 4👇)`
- `👇( seq 4 )` + `forward-arg` → `(👇 seq 4 )`
-  `( seq 4 )👇` + `backward-arg` → `( seq 4 👇)`

krobelus:
I didn't try it yet but skipping over command substitutions seems appropriate for this command.
(Unless the cursor is already inside the command substitution.) Not sure when that would be a hindrance.
I imagine one would bind this to something like `alt-B` and still keep `alt-b` for raw text movement.
Unfortunately it's hard to find a good set of default bindings for "big movements"; `alt-shift` and `ctlr-alt` are often taken. Maybe `ctrl-shift` though xterm/tmux can't disambiguate that. Maybe we should add the "super" modifier

chapmanjacobd:
Thanks for your testing and detailed report [lgarron](https://github.com/lgarron)!

I think this is intended behavior. In the case that you don't want to move across the whole command substitution token, I would use the `backward-word` command with the `Alt+Left` default preset binding. For me, I don't see this command as replacing those commands. I use both `backward-word` and `backward-token`, etc.

However, I do think the newline behavior could be less confusing. I'm putting together some code that should improve it a bit

lgarron:
> However, I do think the newline behavior could be less confusing. I'm putting together some code that should improve it a bit
> 
> edit: I was able to identify the source of the inconsistency and it should be resolved now

Tried it, and it's working fab (also including the `kill-token` commands)! 🤩

I'm sympathetic to the reasons for the subcommand behaviour, and recognize that I'm not part of the project or PR author. But I would still disagree about the subcommand functionality. The current PR implementation would treat sub-commands sometimes as if they are one big argument and sometimes as if they are their own commands, which I think really contradicts the intuitive "it just works" nature of `fish`. Let me see if I can lay this out visually:


## Characters

Consider the following code:

```shell
ls -al ~/Downloads
```

I can use little arrows to draw the effect of `forward-char` (green) and backward-char (orange). This is very straightfroward:

chars

## Words

Now consider the following code:

```shell
cat "~/Downloads/Design files/final-draft.svg" | npm run format > (cat package.json | jq -r ".src.designSVGPath")
```

Using `forward-word` and `backward-word`, there are some characters that may or may not get skipped depending on which direction you're going, but it's still pretty intuitive and linear. In particular, the sub-command is syntactically transparent[^1], and you can get to any word on the line by pressing `forward-word` and `backward-word` often enough.

words

[^1]: I don't know know if that's a real concept, but I just mean "there's nothing preventing" you from looking inside a sub-command. Or perhaps another way to look at it: the `(` and `)` do nothing special to the enclosing code, and are treated similar to characters like `;` and `|`.
## Lines

Now consider this multi-line code:

```shell
function echo_prefixed_warning                    
  echo -n "[warning] "                            
  echo $argv                                      
end                                               

                                                  
if not test -f (cat package.json | jq -r ".main") 
  echo_prefixed_warning "File does not exist."    
end                                               
```

This is also very straightforward when you use the arrow keys for line navigation. Moving a line up or down does not jump across function definitions or `if` blocks:

lines

Again, the sub-command is syntactically transparent: if the cursor is over the word "does", then pressing the up key does not try to avoid stepping inside the subcommand. 

## Tokens (current PR)

Now consider the following code:

```shell
if not test -f (cat package.json | jq -r ".main"); echo "File does not exist."; end
```

`forward-token` and `backward-token` in the current PR act like this:

tokens-current-pr

Unlike the other examples:

- The sub-command is not syntactically transparent. If you are outside you can't get in. But if you're inside, you *can* get out `forward-token`.
- But `fish` *does* understand that there are tokens inside sub-commands. If your cursor already happens to be inside the sub-command, navigation inside works as if you were in the main command.
- Unlike with the equivalent commands for chars, words, or lines, you can't reach every visible token using `forward-token` and `backward-token`.
- The navigation graph is not linear. In fact, you have to do an arbitrary amount of context-sensitive parsing in your head to predict where the cursor will end up.

By contrast, I think this would be a lot more intuitive for commands using these names:

tokens-simple

For example, if I want to remove an argument from any command invocation on that line, I can just press `backward-token` until I'm at the start of that argument, then press `kill-token`.

----

I think it's reasonable to disagree with this, but I think either way `fish` would really benefit from functionality that is consistent about:

1. whether a sub-command is itself always treated like an entire token or not, and/or
2. whether navigation using the new commands is connected between the inside and outside of the sub-command.

I do think it might be nice to have a way to skip the start/end of the current enclosing block or sub-command, but I hope I've demonstrated why I think it doesn't make sense for these `-token` commands to do so.

chapmanjacobd:
I don't have _strong_ feelings about the subcommand behavior specifically, but without these `*-token` commands there isn't a good way of skipping over large tokens like that.

On the other side of things, there are already many ways to start navigation within subcommands. You can use any of the `*-char` or `*-word` commands.

Tokens (simplified) might seem like it would be a simpler implementation, "a simple linear traversal of all tokens" but after playing around with it for a few hours the code becomes a lot more complicated :/

Maybe more people can voice their opinion about how it should work--or if a fish maintainer wants to take over this PR and change it to be simplified I will also be happy with that

krobelus:
Interesting discussion; I think trying it for a while will help.

If we had a nice default binding for `jump-to-matching-bracket`, that would make the decision easier.
For me the easiest way to do large edits is usually `alt-e`.
It's hard to make Emacs bindings scale well; it needs too many bindings.

With the current approach I'd change the following:
1. call update_buff_pos only once, to get a predictable cursor position on undo+redo
2. stop treating newlines specially; move by tokens and nothing else
   - except if we would hit the beginning/end of a cmdsub; exit the cmdsub in that case I guess
- (you forgot to change `chars()` into `chars().rev()` in `backward_token()` but actually we don't need it the `buff_pos` computation all)

<details>

```diff
diff --git a/src/reader.rs b/src/reader.rs
index 0a600694a..db1b9776c 100644
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -2795,51 +2795,53 @@ impl<'a> Reader<'a> {
                 );
             }
             rl::BackwardKillToken => {
-                let (start_buff_pos, elt, new_position) = match self.backward_token() {
-                    Some(value) => value,
-                    None => return,
+                let Some(new_position) = self.backward_token() else {
+                    return;
                 };
 
+                let (elt, _el) = self.active_edit_line();
                 if elt == EditableLineTag::Commandline {
                     self.suppress_autosuggestion = true;
                 }
 
+                let (elt, el) = self.active_edit_line();
                 self.data.kill(
                     elt,
-                    new_position..start_buff_pos,
+                    new_position..el.position(),
                     Kill::Prepend,
                     self.rls().last_cmd != Some(rl::BackwardKillToken),
                 );
             }
             rl::BackwardToken => {
-                let (_start_buff_pos, elt, new_position) = match self.backward_token() {
-                    Some(value) => value,
-                    None => return,
+                let Some(new_position) = self.backward_token() else {
+                    return;
                 };
+                let (elt, _el) = self.active_edit_line();
                 self.update_buff_pos(elt, Some(new_position));
             }
             rl::KillToken => {
-                let (start_buff_pos, elt, new_position) = match self.forward_token() {
-                    Some(value) => value,
-                    None => return,
+                let Some(new_position) = self.forward_token() else {
+                    return;
                 };
 
+                let (elt, _el) = self.active_edit_line();
                 if elt == EditableLineTag::Commandline {
                     self.suppress_autosuggestion = true;
                 }
 
+                let (elt, el) = self.active_edit_line();
                 self.data.kill(
                     elt,
-                    start_buff_pos..new_position,
+                    el.position()..new_position,
                     Kill::Append,
                     self.rls().last_cmd != Some(rl::KillToken),
                 );
             }
             rl::ForwardToken => {
-                let (_start_buff_pos, elt, new_position) = match self.forward_token() {
-                    Some(value) => value,
-                    None => return,
+                let Some(new_position) = self.forward_token() else {
+                    return;
                 };
+                let (elt, _el) = self.active_edit_line();
                 self.update_buff_pos(elt, Some(new_position));
             }
             rl::BackwardWord | rl::BackwardBigword | rl::PrevdOrBackwardWord => {
@@ -3380,33 +3382,22 @@ impl<'a> Reader<'a> {
         }
     }
 
-    fn backward_token(&mut self) -> Option<(usize, EditableLineTag, usize)> {
-        let (elt, el) = self.active_edit_line();
-        let start_buff_pos = el.position();
-        let boundary = 0;
-        if start_buff_pos == boundary {
+    fn backward_token(&mut self) -> Option<usize> {
+        let (_elt, el) = self.active_edit_line();
+        let pos = el.position();
+        if pos == 0 {
             return None;
         }
 
-        // If we are not in a token, look for one behind
-        let buff_pos = start_buff_pos
-            - el.text()[..start_buff_pos]
-                .chars()
-                .take_while(|c| matches!(*c, '\t' | '\x0C' | ' '))
-                .count();
-        self.update_buff_pos(elt, Some(buff_pos));
-
-        let (elt, el) = self.active_edit_line();
-        let text = el.text();
-
         let mut tok = 0..0;
         let mut prev_tok = 0..0;
-        parse_util_token_extent(text, el.position(), &mut tok, Some(&mut prev_tok));
+        parse_util_token_extent(el.text(), el.position(), &mut tok, Some(&mut prev_tok));
 
         // if we are at the start of a token, go back one
-        let new_position = if tok.start == start_buff_pos {
-            if prev_tok.start == start_buff_pos {
-                start_buff_pos - 1
+        let new_position = if tok.start == pos {
+            if prev_tok.start == pos {
+                let cmdsub = parse_util_cmdsubst_extent(el.text(), prev_tok.start);
+                cmdsub.start.saturating_sub(1)
             } else {
                 prev_tok.start
             }
@@ -3414,39 +3405,34 @@ impl<'a> Reader<'a> {
             tok.start
         };
 
-        Some((start_buff_pos, elt, new_position))
+        Some(new_position)
     }
 
-    fn forward_token(&mut self) -> Option<(usize, EditableLineTag, usize)> {
-        let (elt, el) = self.active_edit_line();
-        let start_buff_pos = el.position();
-        let boundary = el.len();
-        if start_buff_pos == boundary {
+    fn forward_token(&self) -> Option<usize> {
+        let (_elt, el) = self.active_edit_line();
+        let pos = el.position();
+        if pos == el.len() {
             return None;
         }
 
         // If we are not in a token, look for one ahead
-        let buff_pos = start_buff_pos
-            + el.text()[start_buff_pos..]
+        let buff_pos = pos
+            + el.text()[pos..]
                 .chars()
-                .take_while(|c| matches!(*c, '\t' | '\x0C' | ' '))
+                .take_while(|c| c.is_ascii_whitespace())
                 .count();
-        self.update_buff_pos(elt, Some(buff_pos));
-
-        let (elt, el) = self.active_edit_line();
-        let text = el.text();
 
         let mut tok = 0..0;
-        let mut prev_tok = 0..0;
-        parse_util_token_extent(text, el.position(), &mut tok, Some(&mut prev_tok));
+        parse_util_token_extent(el.text(), buff_pos, &mut tok, None);
 
-        let new_position = if tok.end == start_buff_pos {
-            tok.end + 1 // this will never equal/exceed boundary
+        let new_position = if tok.is_empty() {
+            let cmdsub = parse_util_cmdsubst_extent(el.text(), tok.end);
+            cmdsub.end + 1
         } else {
             tok.end
         };
 
-        Some((start_buff_pos, elt, new_position))
+        Some(new_position)
     }
 }
 
```

</details>

chapmanjacobd:
Thanks for taking a close look at the code and the provided patch. In my testing it works well. I'll continue playing around with it to see if I can find any other edge cases but feel free to merge.

Is there an idiomatic way to clear the killring without pasting?

chapmanjacobd:
ah I get a crash:

```
assertion failed: position <= self.len()
```

when I do `forward-token` from here:

```
I( seq 4 ))
^ cursor
```

chapmanjacobd:
hmm if the cursor is at a semicolon then `forward-token` will jump to the end (`3I`)

    I;
    test 1 2 3

tbh... I don't see many differences in behavior with/without the `parse_util_cmdsubst_extent` code for forward-token. But the above is one and it seems strange. `backward-token` does not have this problem but it is also using `parse_util_cmdsubst_extent` in a different way.

If there are specific examples where you can see `parse_util_cmdsubst_extent` is helping for `forward-token` please let me know

krobelus:
Great work, thanks. Merged as a9cee9e75. My attempt to add some default bindings is in #10766

krobelus:
This has been quite useful.
I noticed that it fails to accept autosuggestions (token-wise). We should fix that; looks like today's `MoveWordStateMachine` doesn't  quite cut so we probably want to add a special case

PR #10759, opened 2024-10-02
adds completion for the default Elm cli tool

kemelzaidan:
Add completion for the default CLI tool for the [Elm language](https://elm-lang.org/) which was lacking and neither provided for the tool authors. The completion code was written by myself.

### DIFF
--- a/share/completions/elm.fish
+++ b/share/completions/elm.fish
@@ -1,0 +1,24 @@
+set -l commands repl init reactor make install bump diff publish
+
+complete -c elm -f
+# repl completions
+complete -c elm -n "not __fish_seen_subcommand_from $commands" -a repl -d 'Open up an interactive programming session'
+complete -c elm -n "__fish_seen_subcommand_from repl" -l no-colors -d 'Turn off the colors in REPL'
+complete -c elm -n "__fish_seen_subcommand_from repl" -l interpreter= -d 'Path to an alternative JS interpreter'
+# reactor completions
+complete -c elm -n "not __fish_seen_subcommand_from $commands" -a reactor -d 'Compile code with a click'
+complete -c elm -n "__fish_seen_subcommand_from reactor" -l port= -d 'Compile code with a click'
+# make completions
+complete -c elm -n "not __fish_seen_subcommand_from $commands" -a make -d 'Compiles Elm code in JS or HTML'
+complete -c elm -n "__fish_seen_subcommand_from make" -l output= -F -r -d 'Specify the name of resulting JS file'
+complete -c elm -n "__fish_seen_subcommand_from make" -l debug -d 'Turn on the time-travelling debugger'
+complete -c elm -n "__fish_seen_subcommand_from make" -l optimize -d 'Turn on optimizations to make code smaller and faster'
+complete -c elm -n "__fish_seen_subcommand_from make" -l docs= -d 'Generate a JSON file of documentation for a package'
+#other commands completions
+complete -c elm -n "not __fish_seen_subcommand_from $commands" -a init -d 'Start an Elm project'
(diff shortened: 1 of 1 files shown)


share/completions/elm.fish
```diff
@@ -0,0 +1,24 @@
+set -l commands repl init reactor make install bump diff publish
+
+complete -c elm -f
```
EmilyGraceSeville7cf, line 3: Personally, I would recommend putting `elm` in the variable like `set -l command elm`. However, till the command name is short and not gonna change then this solution may be pretty fine. 😄

share/completions/elm.fish
```diff
@@ -0,0 +1,24 @@
...
+# repl completions
+complete -c elm -n "not __fish_seen_subcommand_from $commands" -a repl -d 'Open up an interactive programming session'
+complete -c elm -n "__fish_seen_subcommand_from repl" -l no-colors -d 'Turn off the colors in REPL'
+complete -c elm -n "__fish_seen_subcommand_from repl" -l interpreter= -d 'Path to an alternative JS interpreter'
```
EmilyGraceSeville7cf, line 7: `-l interpreter` for me looks a little bit suspicious, if you wanna an option to be assignable via `=` sign then don't add `-r` flag to `complete`.

EmilyGraceSeville7cf, line 7: I would rather put `"__fish_seen_subcommand_from repl"` in a some variable like so: `set -l repl_condition "__fish_seen_subcommand_from repl"`.

share/completions/elm.fish
```diff
@@ -0,0 +1,24 @@
...
+complete -c elm -n "__fish_seen_subcommand_from make" -l optimize -d 'Turn on optimizations to make code smaller and faster'
+complete -c elm -n "__fish_seen_subcommand_from make" -l docs= -d 'Generate a JSON file of documentation for a package'
+#other commands completions
+complete -c elm -n "not __fish_seen_subcommand_from $commands" -a init -d 'Start an Elm project'
+complete -c elm -n "not __fish_seen_subcommand_from $commands" -a install -d 'Fetches packages from Elm repository'
+complete -c elm -n "not __fish_seen_subcommand_from $commands" -a bump -d 'Figures out the next version number based on API changes'
+complete -c elm -n "not __fish_seen_subcommand_from $commands" -a diff -d 'See what changed in a package between different versions'
+complete -c elm -n "not __fish_seen_subcommand_from $commands" -a publish -d 'Publishes your package on package.elm-lang.org so anyone in the community can use it'
+
+complete -c elm -l help -d "Show a more detailed description"
```
EmilyGraceSeville7cf, line 24: From my personal experience, I suggest moving `-h|--help` and `-v|--version` options always to the top. The point being that user should always be able to quickly understand how to get some help (although the mentioned flags are common, however, sometimes they vary a little bit).

EmilyGraceSeville7cf (commented):
Please respond to my comments, I try to do my best while reviewing the code. 😄

share/completions/elm.fish
faho: I wouldn't do that, this completion file will only ever work for elm.

My position on that idiom is that it's fake flexibility. You're never gonna change that one line to make it magically work for something else - the most common case is something like python/python2/python3, where you can use `complete --wraps` to reuse one completion script.

Or the tool renames itself entirely, and you just rename the file and do `s/-c elm/-c banana/` on it.

So the variable gives you one additional line and a lot of $command everywhere for no real benefit.

zanchey:
Cherry-picked into 4.0 as 7ea368b6d with fixup suggested by @EmilyGraceSeville7cf as 1b6e10713

PR #10762, opened 2024-10-03
completions/portage: Allow symbolic links in repos.conf

wylfen:
## Description

Repository configuration files in the repos.conf/ directory may also be symbolic links, so make sure to find those as well. There's no need to follow links in the find(1) command since the subsequent invocation of cat(1) will follow them automatically.

## TODOs:


share/functions/__fish_print_portage_repository_paths.fish
```diff
@@ -3,7 +3,7 @@ function __fish_print_portage_repository_paths --description 'Print the paths of
     set -l b
     set -l c /usr/share/portage/config/repos.conf
     test -d $a
-    and set b (find $a -type f 2>/dev/null)
+    and set b (find $a -type f,l 2>/dev/null)
```
faho, line 6: This will of course also find links to directories, which will break the subsequent `cat`.

To be honest I quite dislike how convoluted this configuration system is, and I think it's fine to not support every single possibility anyone could conceivably come up with here.

faho: I *think* you mean `-xtype f`, but I am unsure if this is a GNU extension.

Alternatively you could make this `path filter -f -- $a/**`, which will check the pointed-to type.

wylfen: > To be honest I quite dislike how convoluted this configuration system is, and I think it's fine to not support every single possibility anyone could conceivably come up with here.

That's fair, I have had this change sitting around locally for ages and randomly thought to contribute it, so I don't feel too strongly about having it rejected.

> I _think_ you mean `-xtype f`, but I am unsure if this is a GNU extension.

I think it is, I couldn't find `-xtype` in the POSIX specification.

> Alternatively you could make this `path filter -f -- $a/**`, which will check the pointed-to type.

We could also have `find` follow symbolic links with `-L` and drop the `-type l` check. I didn't want to do that but also didn't consider that `-type l` would find links to directories, too.

Up to you. If you don't want to support a configuration like that I think that is fair.

faho: > I think it is, I couldn't find -xtype in the POSIX specification.

To be clear, we're already in GNU extension land with "f,l". I am unsure if any Gentoo user ever will have a non-GNU find.

> We could also have find follow symbolic links with -L and drop the -type l check.

Depends - does portage follow directory symlinks? If so, that's the correct thing to use.

wylfen: > Depends - does portage follow directory symlinks? If so, that's the correct thing to use.

The documentation is unclear on this, so I looked it up, it does indeed follow symlinks to files and directories.

faho: In that case, `-L` seems like the way to go.

wylfen: Actually, just looking at this again, why don't we use `path filter -f $a/**`?

faho:
Thanks!

PR #10766, opened 2024-10-05
Default bindings for token movement commands

krobelus:
There is no natural default binding for token movements. Add
alt-{left,right,backspace,delete} bindings, breaking some existing behavior.

Many terminals don't support distinguishing ctrl-backspace from ctrl-h yet.
Add a special hack for those terminals, to keep the existing alt-backspace
behavior of killing a word, since that one seems more important to have
access to than kill-token.

I guess we could also add:

    bind alt-B backward-token
    bind alt-F forward-token
    bind ctrl-W backward-kill-token
    bind alt-D kill-token

Those might be intercepted by the terminal on Linux, but I don't know where
that happens.

Tested on foot, kitty, alacritty, xterm, tmux, konsole and gnome-terminal.

### DIFF
--- a/share/functions/fish_default_key_bindings.fish
+++ b/share/functions/fish_default_key_bindings.fish
@@ -74,9 +74,10 @@ function fish_default_key_bindings -d "emacs-like key binds"
     bind --preset $argv alt-u upcase-word
 
     bind --preset $argv alt-c capitalize-word
-    bind --preset $argv alt-backspace backward-kill-word
+    bind --preset $argv alt-backspace backward-kill-token
     bind --preset $argv ctrl-backspace backward-kill-word
     bind --preset $argv ctrl-delete kill-word
+    bind --preset $argv alt-delete kill-token
     bind --preset $argv alt-b backward-word
     bind --preset $argv alt-f forward-word
     if test "$TERM_PROGRAM" = Apple_Terminal
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -31,6 +31,8 @@ Notable backwards-incompatible changes
   See :ref:`below <changelog-new-bindings>` for details.
 - Fish no longer supports terminals that fail to ignore OSC or CSI sequences they don't recognize.
   The typical problem is that terminals echo the raw sequences sent by fish instead of silently ignoring them.
+- :kbd:`alt-left` and :kbd:`alt-right` will now move by one argument (which may contain quoted spaces), not just one word.
+- :kbd:`alt-backspace` will delete an entire argument, not just one word.
 - ``random`` now uses a different random number generator and so the values you get even with the same seed have changed.
   Notably, it will now work much more sensibly with very small seeds.
   The seed was never guaranteed to give the same result across systems,
@@ -162,6 +164,7 @@ New or improved bindings
 - During up-arrow history search, :kbd:`shift-delete` will delete the current search item and move to the next older item. Previously this was only supported in the history pager.
   Same for autosuggestions.
 - :kbd:`ctrl-Z` (also known as :kbd:`ctrl-shift-z`) is now bound to redo.
+- :kbd:`alt-delete` now deletes the argument (which may contain quoted spaces) right of the cursor.
 - Some improvements to the :kbd:`alt-e` binding which edits the commandline in an external editor:
   - The editor's cursor position is copied back to fish. This is currently supported for Vim and Kakoune.
   - Cursor position synchronization is only supported for a set of known editors. This has been extended by also resolving aliases. For example use ``complete --wraps my-vim vim`` to synchronize cursors when ``EDITOR=my-vim``.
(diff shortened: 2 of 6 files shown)


share/functions/fish_default_key_bindings.fish
```diff
@@ -74,9 +74,10 @@ function fish_default_key_bindings -d "emacs-like key binds"
     bind --preset $argv alt-u upcase-word
 
     bind --preset $argv alt-c capitalize-word
-    bind --preset $argv alt-backspace backward-kill-word
+    bind --preset $argv alt-backspace 'bind --has-kitty-progressive-enhancement && commandline -f backward-kill-token || commandline -f backward-kill-word'
```
krobelus, line 77: I learned that `ctrl-backspace` is not a thing on macOS; only `alt-backspace` will delete a word.
On Linux it is the other way round.
I'm leaning towards making this OS-dependent, and making the unused key binding delete an entire token.
Of course the OS-dependent behavior can be confusing inside SSH or containers

serban: Please do not make this platform-dependent. As you point out, it will be very confusing.

I use both macOS and Linux daily and always have terminals open to both macOS and Linux hosts. Over the years I have developed a mental model of both macOS and X idioms. In my mental model, the terminal is a third and independent platform, and it should remain internally consistent to itself.

Otherwise, I'm looking forward to this pull request landing!

chapmanjacobd: `--has-kitty-progressive-enhancement` is likely a useful addition for people to use in own configs but I'm not sure it should be used in a default due to difference in behavior across OS/SSH/containers.

`Ctrl-{backspace,delete,left,right}` seems to be pretty common (firefox, many text editors) for deleting/moving across words (in those same programs `Alt-{backspace,left,right}` have different functionality or are unbound). So I agree that we probably shouldn't use `Ctrl-{backspace,delete,left,right}`

If we are going to change the default of `Alt-{backspace,delete,left,right}` I think it makes sense to go all in... What are the conflicting legacy keys for `alt-backspace` or was `--has-kitty-progressive-enhancement` left in from attempting to bind to `ctrl-backspace`?

krobelus: The problem is that terminals like tmux by default don't allow distinguishing `ctrl-backspace` from `ctrl-h` (which means the same as `backspace`, for historical reasons). I guess `alt-backspace` is the obvious workaround. The git blame shows that it's inherited from Emacs.

This is me typing `backspace`, `ctrl-backspace` and `alt-backspace` in tmux.

```
$ fish_key_reader -cV

To terminate this program type "exit" or "quit" in this window,
or press ctrl-c or ctrl-d twice in a row.

Press a key:
# decoded from: \x7f
bind backspace 'do something'
# decoded from: \x08
bind ctrl-h 'do something'
# decoded from: \e\x7f
bind alt-backspace 'do something'
```

So this special hack binding `alt-baackspace` is because otherwise those terminals would get no CUA binding for `backward-delete-word`.
Not sure..

Going forward we should simply fix the affected terminals. There is [ongoing work to fix it in tmux](https://github.com/tmux/tmux/pull/4038) and for other, non-nested terminals it won't be too hard.

krobelus: Basically, if we'd do `bind alt-backspace backward-delete-token` we'd add more pressure/friction to legacy terminals.
Maybe it doesn't matter enough..

> I use both macOS and Linux daily and always have terminals open to both macOS and Linux hosts. Over the years I have developed a mental model of both macOS and X idioms. In my mental model, the terminal is a third and independent platform, and it should remain internally consistent to itself.

One overarching thing we're trying to do is make the terminal less alien.
It is very painful to watch someone press backspace a million times when pair-programming.
(I guess the next thing to fix here is to add mouse support).
What I'm saying is we should also try to be consistent with GUI apps where there's no inherent difference.

Anyway, I also get your point that it's doable to seamlessly switch languages (as long as you do it regularly) so I guess it's fine

chapmanjacobd: I don't use Mac OS personally but on my friend's macbook in Chrome it looks like:

- control-backspace removes the previous word (`backward-kill-word`)
- option-backspace removes the previous word (`backward-kill-word`)
- command-backspace removes everything from the cursor backwards (a bit like `backward-kill-line`)

But in Terminal.app `command-backspace` makes a beep and `backspace,control-backspace,option-backspace` all turn into `\x7f` in fish_key_reader... :/ It seems like there is no `backward-delete-word` by default in (at least in fish v3.6.1)

I can test with iTerm2 and fish-master later if that is helpful

serban: > One overarching thing we're trying to do is make the terminal less alien. It is very painful to watch someone press backspace a million times when pair-programming. (I guess the next thing to fix here is to add mouse support). What I'm saying is we should also try to be consistent with GUI apps where there's no inherent difference.

I hear ya. That is a good and noble goal, and I appreciate that unifying these worlds is often like fitting a square peg in a round hole.

krobelus:
merged the same thing but without the feature detection hack

PR #10777, opened 2024-10-12
Fix old history being momentarily forgotten after history file rewrite

### LINKED ISSUE #6606: history isn't shown if your system clock is set wrong
fish wants to use a single history file, but also ignore item from other sessions that were ran after this session was started. To achieve this it compares the item's timestamp against the current clock. The problem is that if the clock is invalid or set backwards, then old items may appear new and will be ignored.

The real problem is that fish ever re-reads the file. What ought to happen is that fish uses whatever is in the file at session startup or when `history merge` is run, and then ignores the timestamp.

krobelus:
(Not yet sure which of the 3 approaches to take, maybe there is feedback)

Commit 1b9b89316 (After reading corrupted history entry, keep reading older
entries, 2024-10-06) made history search skip over corrupted history entries,
and try the next one.

This fixes some forms of on-disk corruption
However there is an easily reproducible, form of in-memory corruption.
Let the history be

    older-command
    some-newer-command

If fish runs a history search it will load the offsets of each item into
memory.

Now if another fish runs "older-command" (and up to 25 irrelevant commands), it
may vacuum (deduplicate) the history file, rewriting it in least-recently-used
order, which is:

    some-newer-command
    older-command

If the original fish re-runs the same history search, it will use stale
offsets to read from the memory-mapped file.  It actually wants to read from
a snapshot but mmap(3p) doesn't guarantee that:

> It is unspecified whether modifications to the underlying object done
> after the MAP_PRIVATE mapping is established are visible through the
> MAP_PRIVATE mapping.

It looks like on Linux we get the current version of the file contents,
hence history search fails because offsets don't point to valid items anymore.
This corruption goes away the next time fish vacuums the history file itself.

This commit fixes it by listening to history file changes.

Since this is a race between the notification system and the user performing
a history search, we can still hit it in theory, so we still need to check
lines for correctness and silently fail if they are broken in any way.

Alternative approaches (that don't have this imperfection).
1. take a shared lock before calling item_at_index().
   Haven't explored this very much; it seems like a reasonable approach.
2. create a private snapshot at startup (this would fix #6606):

    flock --shared ~/.local/share/fish_history
    cp ~/.local/share/fish_history ~/.local/share/fish_history.$fish_pid
    flock --shared ~/.local/share/fish_history --unlock

and then read old items only from that snapshot. It's only updated on
"history merge". We'll still write to the shared history file the same way.

Would be even better if we could keep the snapshot in memory.
Ideally we use copy-on-write for snapshots; I don't know if there are
portable implementations.

If we don't have copy-on-write we could compress the history (and still mmap+compress it on the fly).
Need to measure how much memory/disk space we want to spend...

### DIFF
--- a/src/history.rs
+++ b/src/history.rs
@@ -141,6 +141,8 @@ const HISTORY_FILE_MODE: Mode = Mode::from_bits_truncate(0o600);
 /// the file and taking the lock
 const MAX_SAVE_TRIES: usize = 1024;
 
+pub const VACUUM_FREQUENCY: usize = 25;
+
 /// If the size of `buffer` is at least `min_size`, output the contents `buffer` to `fd`,
 /// and clear the string.
 fn flush_to_fd(buffer: &mut Vec<u8>, fd: RawFd, min_size: usize) -> std::io::Result<()> {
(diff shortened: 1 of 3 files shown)


ridiculousfish:
As designed, we 100% ought to be reading from a snapshot. The idea is:

1. fish (perhaps multiple instances) memory maps the file.
2. The file may be appended to. That doesn't interfere with existing fish instances, which only read up to the end of their mapping.
3. When vacuuming, the file may be replaced outright, via the `rename` system call. But this creates a new "underlying object": the original object remains referenced and unmodified.

So this is pretty weird - I'll poke around with the test but please refrain from merging until we have a better understanding of the bug.

krobelus:
> But this creates a new "underlying object": the original object remains referenced and unmodified.

Ok if we need a new inode then we can revert 5db0bd587 (Lock history file before reading it, 2024-10-09)
or change it to use the rename code path.

If we revert and timestamp granularity is really a problem we could probably use inotify make the timestamp checking logic more reliable.

PR #10792, opened 2024-10-16
Add completions for: `guile`, `guild`

EmilyGraceSeville7cf:
## Description

Add completions for:

- [x] [`guile`](https://www.gnu.org/software/guile/manual/html_node/Command_002dline-Options.html)
- [x] [`guild`](https://www.gnu.org/software/guile/manual/html_node/Compilation.html)

## Notes

I have no clue how to extract all function names from the current file purely from Scheme code, that's why for now I use [regular expressions](https://github.com/fish-shell/fish-shell/pull/10792/files#diff-8bc0c9a2375d402b31006ee4100d542f6b5779d0ad50354b5ba924ce284af3f4R49) to find function names.

## TODOs:

### DIFF
--- a/share/completions/guild.fish
+++ b/share/completions/guild.fish
@@ -15,7 +15,6 @@ function __fish_guild__complete_optimizations
 end
 
 set -l command guild
-
 complete -c $command -f
 
 set -l compile_condition '__fish_seen_subcommand_from compile'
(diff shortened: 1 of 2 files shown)


ridiculousfish:
Thank you

PR #10796, opened 2024-10-19
completions/diff: allow file completion for --unified

ammgws:
I often use diff like `diff --unified file1 file2` however the current completions disable file completion when using that flag.

Seems valid to enable file completions since that flag has a default value.


faho:
The man page for GNU diffutils diff is annoyingly unclear on this

> -u, -U NUM, --unified[=NUM]
              output NUM (default 3) lines of unified context

What that means is: `-u` and `--unified` take an optional argument, `-U` takes a required argument.

The long version should be moved to the `-u` one.

ammgws:
If squash on merge is not possible I will fix the commits later.

faho:
Thanks!

PR #10801, opened 2024-10-20
fix documentation for `path normalize`

giorgiga:
## Description

This removes `(except for two leading "//")` from the documentation of `path normalize`, since `path normalize` does actually squash leading "//" (I tested the current HEAD and 3.7.1 and both behave the same)

> `path normalize` returns the normalized versions of all paths. That means it squashes duplicate "/" ~~(except for two leading "//")~~, collapses "../" with earlier components and removes "." components.

Please verify by running

```fish
❱ path normalize //usr/bin
/usr/bin 
```

NB: this should probably be merged both on master and on the branch for the next non-rust release

## TODOs:

*(none applies, so I am checking all off as done)*

### DIFF
--- a/doc_src/cmds/path.rst
+++ b/doc_src/cmds/path.rst
@@ -279,7 +279,7 @@ Examples
 
     path normalize [-z | --null-in] [-Z | --null-out] [-q | --quiet] [PATH ...]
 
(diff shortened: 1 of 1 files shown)


giorgiga:
Apparently my editor trimmed a few whitespace-only lines (which I assume is bad?)

I've undone that and force-pushed, sorry for not checking beforehand.

faho:
Thanks!

PR #10808, opened 2024-10-22
remove $ from a $(command)

doomdagadiggiedahdah:
## Description

Fixes typo in Command Substitutions[¶](https://fishshell.com/docs/current/tutorial.html#command-substitutions) of Tutorial

Current example gives this:
> echo In (pwd), running $(uname)
fish: $(...) is not supported. In fish, please use '(uname)'.
echo In (pwd), running $(uname)
> echo In (pwd), running $uname
In /home, running
> echo In (pwd), running (uname)
In /home, running Linux

Fixes issue #

## TODOs:

### DIFF
--- a/doc_src/tutorial.rst
+++ b/doc_src/tutorial.rst
@@ -386,7 +386,7 @@ Command Substitutions
 
 Command substitutions use the output of one command as an argument to another. Unlike other shells, fish does not use backticks `` for command substitutions. Instead, it uses parentheses with or without a dollar::
 
-    > echo In (pwd), running $(uname)
+    > echo In (pwd), running (uname)
     In /home/tutorial, running FreeBSD
 
 A common idiom is to capture the output of a command in a variable::


doomdagadiggiedahdah:
This example (also from cmd sub) also doesn't work. Checked version and am on: fish, version 3.7.1

```
 touch "testing_$(date +%s).txt"
 ls *.txt
```

faho:
> Checked version and am on: fish, version 3.7.1

No, you're not.

Your running fish is much older, run `echo $version`.

`$()` has been supported since 3.4

doomdagadiggiedahdah:
Apologies, I had installed with `apt install fish`, updated to 3.7.1 without restarting the terminal (so `fish --version` would still output `fish, version 3.7.1` like I included above but was still running prev version). Both commands and `$()` work as expected.

Sorry about that!

PR #10810, opened 2024-10-23
Turn off release-mode overflow checks again

krobelus:
Not sure about this but "no overflow checks" is the status quo, so
if we want to keep the checks beyond the port, this should be stated
explicitly, as implied by e616de544 (Enable rust overflow checks in
release mode, at least for now, 2023-02-20).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,6 @@ rust-version = "1.70"
 edition = "2021"
 
 [profile.release]
-overflow-checks = true
 lto = true
 
 [package]


ridiculousfish:
Up to you, no objection here.

mqudsi:
I would prefer to keep them on if there are no major changes to the benchmarks. They've caught problems in the past.

krobelus:
It's not about benchmarks, it's about a potential behavior change for unsigned numbers; sometimes it's better not to crash. I don't care much either way.

PR #10824, opened 2024-11-01
Add libncurses-dev to apt install list

parker-research:
## Description

Very simple change to add `libncurses-dev` to `apt install` list. Apparently on some very old Ubuntu Docker images (e.g., Ubuntu 20.04), `libncurses-dev` isn't installed by default. Having it listed here ensures it gets installed successfully.

## TODOs:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -181,7 +181,7 @@ On Debian or Ubuntu you want these packages:
 
 ::
 
-   sudo apt install build-essential cmake libpcre2-dev gettext
+   sudo apt install build-essential cmake libpcre2-dev gettext libncurses-dev
 
 On RedHat, CentOS, or Amazon EC2 everything should be preinstalled.
 


README.rst
```diff
@@ -181,7 +181,7 @@ On Debian or Ubuntu you want these packages:
 
 ::
 
-   sudo apt install build-essential cmake libpcre2-dev gettext
+   sudo apt install build-essential cmake libpcre2-dev gettext libncurses-dev
```
krobelus, line 184: If you run `git blame` on this line you'll find commit f62f1aaf9 (Remove remaining mentions of curses, 2024-02-23).
Since an earlier neighboring commit, ncurses is no longer a dependency but terminfo is.
But actually almost no user should even need terminfo because we default to assume xterm compatibilty.
So I'm not sure if we should even recommend installing that. It be great to move away from terminfo too, to avoid issues in SSH etc. and make the behavior more predictable

parker-research:
Damn, fair enough. The issue is that this README suggests that it contains build instructions for the latest release (from 6+ months ago), but that the instructions aren't complete. That's fair enough though that the build instructions associated with the release should be used to build that release.

Any idea when the `main` branch is going to see its next release? I'm excited to see what the Rust migration brings!

krobelus:
I guess "Compiling fish from a tarball" doesn't necessarily imply that that tarball is already released.

zanchey:
I've started to remove all the C++ bits from the README in preparation for the 4.0b1 tarball, sorry.

PR #10833, opened 2024-11-05
Emit only sane pgid value for `jobs` output

mqudsi:
We were previously printing the internal INVALID_PID value (since removed), which was a meaningless `-2` constant, when there was no pgid associated with a job.

This PR changes that to `-` to indicate no pgid available, which I prefer over something like `0` or `-1`, but will cause problems for any code out there that is hardcoded to convert this field to an integral value (do we have any code like that ourselves?).

Other options include 0 or -1 as mentioned above, or printing fish's own pgid.

EDIT:

I don't recommend printing 0 or fish's own pgid because if those get through validation we could end up killing ourself or something equally disasterous.


krobelus:
LGTM. This fixes the output of `fish -c 'sleep 10 & jobs'`

PR #10840, opened 2024-11-10
Remove already deprecated IFS support

mqudsi:
This has been officially deprecated for a long, long time.

Additionally, pretty much all aspects of fish itself (aside from the core language) break when the user messes with `$IFS` (it was intentionally not a read-only variable) as all fish functions that rely on interpreting output of commands have their behavior changed in a way they can't handle. We've already decided to remove `$IFS` a long time ago, but if we wanted to keep it we would have to make it a special variable that only applies to the user script being executed but somehow doesn't apply to fish completions, default functions, runtime helpers, etc.

As an example, executing 

```
> set IFS
> help
```

no longer works, because fish now thinks *all* of these potential paths are one long path:

```
> help
help: Help is being displayed in /mnt/c/WINDOWS/System32/WindowsPowerShell/v1.0/powershell.exe
/mnt/c/WINDOWS/system32/cmd.exe
/mnt/c/Windows/System32/cmd.exe.
/mnt/c/WINDOWS/System32/WindowsPowerShell/v1.0/powershell.exe
/mnt/c/WINDOWS/system32/cmd.exe
/mnt/c/Windows/System32/cmd.exe: command not found
```

### DIFF
--- a/doc_src/cmds/read.rst
+++ b/doc_src/cmds/read.rst
@@ -84,7 +84,7 @@ The following options control how much is read and how it is stored:
     Makes ``read`` return after reading *NCHARS* characters or the end of the line, whichever comes first.
 
 **-t** -or **--tokenize**
-    Causes read to split the input into variables by the shell's tokenization rules. This means it will honor quotes and escaping. This option is of course incompatible with other options to control splitting like **--delimiter** and does not honor :envvar:`IFS` (like fish's tokenizer). It saves the tokens in the manner they'd be passed to commands on the commandline, so e.g. ``a\ b`` is stored as ``a b``. Note that currently it leaves command substitutions intact along with the parentheses.
+    Causes read to split the input into variables by the shell's tokenization rules. This means it will honor quotes and escaping. This option is of course incompatible with other options to control splitting like **--delimiter**. It saves the tokens in the manner they'd be passed to commands on the commandline, so e.g. ``a\ b`` is stored as ``a b``. Note that currently it leaves command substitutions intact along with the parentheses.
 
 **-a** or **--list**
     Stores the result as a list in a single variable. This option is also available as **--array** for backwards compatibility.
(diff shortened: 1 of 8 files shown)


mqudsi:
Ok, I changed the milestone.

I do think fish 4.0 would have been a good opportunity to finally break away from oh-my-fish issues, but it is true that we never emitted a "deprecated feature" message when IFS was set.

In the meantime I might add a hack to disable globally setting IFS?

ridiculousfish:
I don't think we should do anything about IFS as part of the 4.0 release. We can add a feature flag in the next minor release.

faho:
(converted this to a draft so the "merge" button is unavailable to click by accident)

krobelus:
Not knowing much about this, I'd like to keep it for `read`.
This use might feel dated but it's not broken, not enough reason to outlaw it IMO.

For command substitutions, it might be broken and a good idea to remove, I'm not sure.

mqudsi:
I like @krobelus's suggestion; that's a good compromise we could keep long-term and definitely beats making it a special local-only variable.

mqudsi:
I've reworked the commit to keep $IFS as a (still deprecated?) means of setting `read --delimiter` but removed it from everywhere else in the shell. This makes our own `/usr/share/fish/` code much more resistant to `$IFS` manipulations (though we might have issues in the few places we use `read` in completions, etc if we aren't explicitly setting `--delimiter` ourselves), removing much of the need to remove `$IFS` altogether and making it easier to stomach keeping it around for legacy `read` compatibility purposes.

I think with this approach this *should* be included in 4.0 as the only thing it does is make fish itself more resilient to breakage (since globally setting `$IFS` basically broke everything, I believe it's safe to assume that removing that option does not hurt anyone) while preserving the local-only `$IFS` to dictate `read` behavior we intend to officially deprecate and hopefully remove at some point in the future.

faho:
> I've reworked the commit to keep $IFS as a (still to be someday removed) means of setting read --delimiter but removed it from everywhere else in the shell. 

That will still break setting IFS to empty to avoid splitting command substitutions. The replacements for that were added in 3.1 (`string collect`) and 3.4 (`"$()"`).

> I think with this approach this should be included in 4.0 as the only thing it does is make fish itself more resilient to breakage

I would not want to add it to 4.0 at this point. We could merge it first thing after, to possibly gain some information on what breaks, and back it out before 4.1 if it's too much.

mqudsi:
That's a good point and sounds reasonable to me.

PR #10844, opened 2024-11-11
Use strongly typed `Pid` and `Option<Pid>` types

### LINKED ISSUE #10832: Job id expansion fails for mixed process/function pipelines
```
> function test_jobs; end;
> cat | test_jobs &
fish: Job 1, 'cat | test_jobs &' has stopped
> jobs
1       77562   0%      stopped cat | test_jobs &
> fg %1
fg: No suitable job: 0
> jobs -p %1
77562
0
```

The problem is that `__fish_expand_pid_args` returns pids `77625` and the invalid/incorrect `0` for the one job that is running (in that order), and the `fg` function forwards only the last value to the actual `fg` command (`fg $args[-1]`).

`jobs -p 0` naturally fails with `jobs: No suitable job: 0`

We can hack around this by filtering out `0` in `__fish_expand_pid_args`, but we might want to stop emitting `0` altogether in `jobs -p`.

mqudsi:
There's a separate `AtomicPid` type because relying on not storing zero or negative values into a `NonZeroI32` was not working. This found a number of bugs or unhandled cases in the codebase.

One thing I want to make sure of is that it is indeed an error to use `bg %0` or `jobs -q 0` in the sense that they can never match a job?

Closes #10832

### DIFF
--- a/src/bin/fish.rs
+++ b/src/bin/fish.rs
@@ -50,7 +50,7 @@ use fish::{
     printf,
     proc::{
         get_login, is_interactive_session, mark_login, mark_no_exec, proc_init,
-        set_interactive_session,
+        set_interactive_session, Pid,
     },
     reader::{reader_init, reader_read, term_copy_modes},
     signal::{signal_clear_cancel, signal_unblock_all},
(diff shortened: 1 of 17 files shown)


mqudsi:
Is the macOS tmux failure related? I'd say no but it failed twice.

PR #10854, opened 2024-11-18
completions/tmux: add skeleton "Windows and Panes" bindings (plus minor changes)

ilyagr:
These are quite mechanical, but include all the commands (as of tmux
3.5a), and all the target-pane/session/client/window flags, and the -F
flags. After this, `tmux pane<tab>` will hopefully be useful.

A few TODOs mention low-hanging fruit for somebody who better
understands fish's complete syntax (or a future me).

Another piece of low-hanging fruit would be completion for all
the `target-window` flags. This PR merely lists them.

## TODOs:
- (n/a) Changes to fish usage are reflected in user documentation/manpages.
- (n/a) Tests have been added for regressions fixed
- (n/a) User-visible changes noted in CHANGELOG.rst

PR #10855, opened 2024-11-18
completions/tmux: dynamic completions for all `tmux` subcommands and flags using `tmux lscm`

ilyagr:
**Update:** This now completes all flags and all commands. So, `tmux if-shell -<TAB>` now works as well.

**Original description:** For example, `tmux shell<tab>` now completes to `if-shell` and `run-shell`, though
no additional information is provided.

This PR and #10854 complement each other, but either could be used independently from the other if necessary.

## TODOs:
- (n/a) Changes to fish usage are reflected in user documentation/manpages.
- (n/a) Tests have been added for regressions fixed
- (n/a) User-visible changes noted in CHANGELOG.rst

### DIFF
--- a/share/completions/tmux.fish
+++ b/share/completions/tmux.fish
@@ -20,6 +20,10 @@ end
 #don't allow dirs in the completion list...
 complete -c tmux -x
 
+# Complete even commands not explicitly listed below, as long as `tmux list-commands` works
+set -l all_commands (tmux list-commands -F "#{command_list_name} #{command_list_alias}" 2>/dev/null)
+and complete -c tmux -n __fish_use_subcommand -a "$all_commands"
+
 ###############  Begin: Front  Flags ###############
 #these do not require parameters
 complete -c tmux -n __fish_use_subcommand -s 2 -d 'Force tmux to assume the terminal supports 256 colours'
--- a/share/completions/tmux.fish
+++ b/share/completions/tmux.fish
@@ -20,9 +20,51 @@ end
 #don't allow dirs in the completion list...
 complete -c tmux -x
 
-# Complete even commands not explicitly listed below, as long as `tmux list-commands` works
-set -l all_commands (tmux list-commands -F "#{command_list_name} #{command_list_alias}" 2>/dev/null)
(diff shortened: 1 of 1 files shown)


share/completions/tmux.fish
```diff
@@ -20,6 +20,10 @@ end
 #don't allow dirs in the completion list...
 complete -c tmux -x
 
+# Complete even commands not explicitly listed below, as long as `tmux list-commands` works
+set -l all_commands (tmux list-commands -F "#{command_list_name} #{command_list_alias}" 2>/dev/null)
+and complete -c tmux -n __fish_use_subcommand -a "$all_commands"
```
ilyagr, line 25: Is there a way to ensure this completion is always tried last, only if no manual completion matches?

Putting it first in this file makes it not interfere with anything else *in this file*,
but I worry a bit about any users/distributions/etc that define custom tmux completions (I doubt
there are many, but still).

ilyagr: The completions for all the flags are now also provided as a separate commit. If it's harder to review, merging just the first of the commits is an option.

ilyagr: > Is there a way to ensure this completion is always tried last, only if no manual completion matches?

After the matrix discussion about rclone [1](https://matrix.to/#/!YLTeaulxSDauOOxBoR:matrix.org/$L4rb8F0WDM3GSw0ZOtJkSbBqZT-QuFH3vWwL6fewqsw?via=matrix.org&via=gitter.im&via=one.ems.host), [2](https://matrix.to/#/!YLTeaulxSDauOOxBoR:matrix.org/$BHA9IgsHKsRXc5WdflpUY3IdbaL-I5EdKppDjuHss1A?via=matrix.org&via=gitter.im&via=one.ems.host), I think this is fine as is. The built-in completions are not meant to be evaluated if the user provides their own earlier in the `fish_complete_path`.

So, I'll resolve this for now.

zanchey:
Cool, this works well for me.

zanchey:
Cherry-picked as d29d63d930d8..9385a25c22b0

PR #10861, opened 2024-11-19
Flush stdout when control returns to shell

mqudsi:
This prevents kernel-buffered stdout contents from being written to the terminal after the prompt itself, which can happen when the external process is terminated with a signal and hasn't executed the libc/runtime-provided exit handlers that flush stdout.

Screenshots of afl-fuzz being terminated by `^C` before and after the change:

Before:

image

After:

image

I tried to minimize any performance impact by trying to reduce the number of times this is called (by placing it as high up as possible, but bubbling up reap status) and by not performing the call in non-interactive mode (where fish shouldn't be painting to the shell bypassing stdout). stderr isn't flushed because it's typically unbuffered.

Let me know if you think this should be included in the changelog afterwards.

## TODOs:

### DIFF
--- a/src/proc.rs
+++ b/src/proc.rs
@@ -1052,8 +1052,9 @@ impl Job {
         );
 
         // Wait for the status of our own job to change.
+        let mut external_proc_exited = false;
         while !fish_is_unwinding_for_exit() && !self.is_stopped() && !self.is_completed() {
-            process_mark_finished_children(parser, true);
+            external_proc_exited |= process_mark_finished_children(parser, true);
         }
         if self.is_completed() {
             // Set $status only if we are in the foreground and the last process in the job has
@@ -1064,6 +1065,16 @@ impl Job {
                 if let Some(statuses) = self.get_statuses() {
                     parser.set_last_statuses(statuses);
                     parser.libdata_mut().status_count += 1;
+                }
+            }
+
+            // Prevent kernel-buffered output from being written to the terminal after the prompt.
+            // We don't need to this if non-interactive because no prompt is shown. See #10861.
+            if external_proc_exited && parser.is_interactive() {
+                // Normally we'd only need to check if the last proc exited abnormally, but it could
+                // be an earlier proc in the pipeline connected to stdout if redirections are used.
+                if procs.iter().rev().any(|p| p.status.signal_exited()) {
+                    let _ = std::io::stdout().flush();
                 }
             }
         }
@@ -1354,7 +1365,7 @@ fn handle_child_status(job: &Job, proc: &Process, status: &ProcStatus) {
     }
 }
 
-/// Wait for any process finishing, or receipt of a signal.
+/// Wait for any process finishing, or receipt of a signal. Used only by the `wait` builtin.
 pub fn proc_wait_any(parser: &Parser) {
     process_mark_finished_children(parser, true /*block_ok*/);
     let is_interactive = parser.libdata().is_interactive;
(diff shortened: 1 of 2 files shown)


src/proc.rs
```diff
@@ -1066,6 +1067,12 @@ impl Job {
                     parser.libdata_mut().status_count += 1;
                 }
             }
+
+            // Prevent kernel-buffered output from being written to the terminal after the prompt.
+            // We don't need to this if we're running non-interactively, where no prompt is shown.
+            if external_proc_exited && parser.is_interactive() {
+                let _ = std::io::stdout().flush();
```
krobelus, line 1074: Seems reasonable.
Maybe we should do this only if the job is `signal_exited`?
I don't know if this scenario is also possible on normal exit.

Another approach is to do it in `Reader::readline()` before `exec_prompt`,
using something like `exec_external_count` to decide whether we want to flush.

(Both approaches won't handle the case when the output comes from a background job that's still running).

> Let me know if you think this should be included in the changelog afterwards.

Good question; can go both ways I think.
I don't think there is an obligation to add it

zanchey:
I'd mention it:

"Standard output is flushed after external commands exit, in an attempt to reduce interleaving their output with the prompt (:issue:`10861`)."

src/proc.rs
mqudsi: I think keeping the flush call closer to the exec code makes more sense than putting it in the reader, btw.

ridiculousfish (changes requested):
This doesn't make any sense to me. In my understanding, `std::io::stdout().flush()` flushes Rust's *internal* buffer. Because this is in-process, there is no way for an external command, such as afl-fuzz, to append to it, so there is nothing to flush, so I would not expect this to do anything at all.

If there is something to flush, then it ought to have been flushed before the external command ran!

Can you please write a command line program that demonstrates this in action? afl-fuzz doesn't seem available on Apple Silicon.

mqudsi:
You're right; I think I ran into a change in timing that resulted in the "corrected" behavior. I was thinking in terms of `fsync()` rather than `flush()` but that doesn't really mean anything for a fifo buffer. I really don't know why I didn't catch this, thank you for bringing me to my senses.

Are race conditions here inevitable?

ridiculousfish:
I think *some* race conditions are inevitable, especially with curses-style apps such as afl-fuzz. For example, if the app moves the cursor to some point on the screen and then dies, then fish will inherit a weird cursor position and I can see how that might screw up the prompt - though I would expect the `PROMPT_SP` hack to handle that (odd that it doesn't).

If you want to pursue this, it would be useful to write a custom tool that reproduces this, so we can explore more.

mqudsi:
It actually reproduces when afl is still starting up and only producing a colorized log of text to stdout with no real tui to speak of, but it must be using absolute positioning ansi escapes or similar because regular stdout is strictly linear. 

I wonder if I can just record and replay the output of afl instead of trying to independently reproduce the race.

krobelus:
`script` from util-linux has a `--log-timing` option that should enable almost-perfect replay

mqudsi:
Thanks, that's useful. Probably have to use socat to trick it into thinking its running in a terminal, though.

mqudsi:
I wrote this last week but never hit submit:

-------------


I didn't need to use `socat` as `script` uses a pseudotty itself.

Here's a sample (replay with `scriptreplay --timing=timing.log cargo-afl.log`) that reproduces the issue: [cargo-afl.log](https://github.com/user-attachments/files/17894249/cargo-afl.log), [timing.log](https://github.com/user-attachments/files/17894250/timing.log)

Looking at the output with `sequin` makes it easier to understand all the output, but it's not particularly elucidating. There doesn't seem to be any absolute positioning of text, in fact it just looks like fish took over but `cargo afl` kept writing, probably via a child process that inherited the same tty.

One obvious answer is to set `+tostop` at all times except when fish has explicitly launched a background process, but there are obvious drawbacks.

PR #10863, opened 2024-11-20
Add completion for `path basename -E/--no-extension`

kidonng:
## Description

Missed in #10521

## TODOs:


mqudsi (approved):
Nice catch and thanks for the PR!

The usage of `contains` to compare two strings is unorthodox, but it's shorter than `string match` and more similar to the neighboring code so I'll allow it :)

PR #10869, opened 2024-11-25
posix_spawn: Unconditionally default all signals (except HUP)

faho:
As I understand it, we want all signals for processes we spawn (except for HUP, because of nohup) to have the default signal handlers.

So I don't know why we need to check what action a specific signal currently has, we can just default it anyway.

This saves ~30 syscalls *per process* we spawn, so:

```fish
for f in (seq 1000)
    command true
end
```

has ~30000 fewer rt_sigaction calls. These take up about ~30% of the total time spent in syscalls according to strace.

We could also compute this set once at startup and then reuse it.

### DIFF
--- a/src/fork_exec/spawn.rs
+++ b/src/fork_exec/spawn.rs
@@ -3,7 +3,7 @@
 use super::blocked_signals_for_job;
 use crate::proc::Job;
 use crate::redirection::Dup2List;
-use crate::signal::get_signals_with_handlers;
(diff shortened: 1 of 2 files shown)


ridiculousfish (approved):
Great find! I can't think of any reason to not do this.

krobelus:
> We could also compute this set once at startup and then reuse it.

yeah we can cache these indefinitely, other shells do that too

PR #10876, opened 2024-11-27
Builtin indent and key reader

faho:
## Description

This makes it so fish_key_reader and fish_indent are available as builtins. That makes it easier to distribute fish as a single file (especially when paired with #10367).

However, to allow calling them from the outside, the main `fish` binary now checks its argv[0] - if that uses a filename "fish_indent" or "fish_key_reader", it will immediately become the respective program and exit. This is like how busybox can be ls or rm or a shell.

## TODOs:

This is pretty rough and WIP. A quick attempt to see how well it works.

- [x] These two use direct {e,}printf! and so aren't redirectable - [ ] Cleanup

And the big picture:

- [ ] Do we want this?
- [ ] Do we want to ship links by default?
- [ ] How?
- [ ] Do we want to encourage calling `fish -c fish_indent` instead? Deprecate the links?
- [ ] Is `fish_key_reader` a big enough thing that it deserves a link? Or is it only ever called inside of fish?
- [ ] Do we want *other* builtins to be available via direct call?


zanchey:
`cargo install` already installs all three binaries, right? So this would be more to allow us to publish a single binary as a download? I'm fine with it, it makes the fish binary slightly larger, the magic of LTO (I think) means `fish_key_reader` stays the same size at about 25% of the main binary. CMake can be taught to install as a symlink, which saves on disk space. I don't think it will reduce package sizes because they should be compressed anyway.

faho:
> cargo install already installs all three binaries, right?

That's right, and this doesn't change it.

We *could* remove one or both of them in future if we go through with this. `fish_indent` is probably more useful outside of fish given that it may be called by editors or CI scripts and such.

I doubt fish_key_reader sees much use (it's too tied to fish's bind notation for it).

> So this would be more to allow us to publish a single binary as a download?

That's the big thing. We could have a fish.xz, and you would decompress and put it wherever you want.

Installing fish on a machine would be `curl https://fishshell.com/current/fish.xz; xz -d fish.xz; ./fish --install; ./fish`. On a remote machine you'd have `scp fish user@host`.

(I *think* we may still want a tarball because of permissions, I'm unsure)

Uninstallation is `rm -r /path/to/fish ~/.local/share/fish/install`.

Plus we could guarantee that `fish_indent` is available inside of fish - there's no way to break the installation by removing it or forgetting to install it.

ridiculousfish:
This is a good idea; I like this direction.

src/builtins/fish_indent.rs
```diff
@@ -0,0 +1,1200 @@
...
+    let mut only_indent = false;
+    let mut only_unindent = false;
+    // File path for debug output.
+    let mut debug_output = None;
+
+    let short_opts: &wstr = L!("+d:hvwicD:");
+    let long_opts: &[WOption] = &[
+        wopt(L!("debug"), ArgType::RequiredArgument, 'd'),
+        wopt(L!("debug-output"), ArgType::RequiredArgument, 'o'),
+        wopt(L!("debug-stack-frames"), ArgType::RequiredArgument, 'D'),
```
faho, line 805: I am inclined to remove these debug options - they've never been very useful and are awkward from inside fish. The third is also unused and has been for years now.

I would also move "--version" to the throwing_main because it is weird to have a builtin write its version?

src/builtins/fish_indent.rs
```diff
@@ -0,0 +1,1200 @@
...
+            'w' => output_type = OutputType::File,
+            'i' => do_indent = false,
+            '\x04' => only_indent = true,
+            '\x05' => only_unindent = true,
+            '\x01' => output_type = OutputType::Html,
+            '\x02' => output_type = OutputType::Ansi,
+            '\x03' => output_type = OutputType::PygmentsCsv,
+            'c' => output_type = OutputType::Check,
+            'd' => {
+                activate_flog_categories_by_pattern(w.woptarg.unwrap());
```
faho, line 846: This currently allows activating debug options globally by running `fish_indent`.

I'm not sure we currently have anything else to do that?

src/builtins/fish_indent.rs
```diff
@@ -0,0 +1,1200 @@
...
+        wopt(L!("dump-parse-tree"), ArgType::NoArgument, 'P'),
+        wopt(L!("no-indent"), ArgType::NoArgument, 'i'),
+        wopt(L!("only-indent"), ArgType::NoArgument, '\x04'),
+        wopt(L!("only-unindent"), ArgType::NoArgument, '\x05'),
+        wopt(L!("help"), ArgType::NoArgument, 'h'),
+        wopt(L!("version"), ArgType::NoArgument, 'v'),
+        wopt(L!("write"), ArgType::NoArgument, 'w'),
+        wopt(L!("html"), ArgType::NoArgument, '\x01'),
+        wopt(L!("ansi"), ArgType::NoArgument, '\x02'),
+        wopt(L!("pygments"), ArgType::NoArgument, '\x03'),
```
faho, line 815: This is used by our docs - we highlight the code samples by shelling out to `fish_indent`.

That would have to be adjusted to call `fish -c 'fish_indent $argv' args` if we decide to drop the second binary.

src/builtins/fish_indent.rs
faho: Note: `fish_indent --version` is called from doc_src/conf.py to determine the version for the docs.

This should probably be passed in from the build script via an environment variable, like it already is in #10367.

faho:
Also: This gives us the opportunity to rename `fish_indent`.

We could call it `fishfmt`, given that it does a bunch more than indentation, and for consistency with `cargo fmt` and `gofmt`.

Not saying we *have to*, but I wouldn't be opposed.

src/builtins/fish_indent.rs
faho: I've removed these - we would have to do some work to make them *not* affect the running fish session, and they don't really apply to the fish_indent itself.

If you want to debug how some script parses, you can use `fish --no-execute --debug ...`.

krobelus:
Sounds good.

Renaming would be great (also underscores are odd, I'd rather use "-" or no separator).

`fish_key_reader` could go into `bind --which-keys` eventually, not sure.

src/builtins/fish_indent.rs
```diff
@@ -0,0 +1,1149 @@
...
+            future_feature_flags::set_from_string(s.as_utfstr());
+        }
+    }
+
+    do_indent(&mut streams, args)
+}
+
+pub fn fish_indent(_parser: &Parser, streams: &mut IoStreams, args: &mut [&wstr]) -> Option<c_int> {
+    let args = args.iter_mut().map(|x| x.to_owned()).collect();
+    Some(do_indent(streams, args))
```
krobelus, line 770: `future_feature_flags::set_from_string(s.as_utfstr());` is not necessary when running as builtin, fish already does that

src/builtins/fish_key_reader.rs
```diff
@@ -0,0 +1,278 @@
...
+    0
+}
+
+/// Setup our environment (e.g., tty modes), process key strokes, then reset the environment.
+fn setup_and_process_keys(streams: &mut IoStreams, continuous_mode: bool, verbose: bool) -> i32 {
+    signal_set_handlers(true);
+    // We need to set the shell-modes for ICRNL,
+    // in fish-proper this is done once a command is run.
+    unsafe { libc::tcsetattr(0, TCSANOW, &*shell_modes()) };
+    terminal_protocol_hacks();
```
krobelus, line 138: these 3 statements should not be necessary when running as builtin

src/builtins/fish_key_reader.rs
```diff
@@ -0,0 +1,278 @@
...
+
+    if continuous_mode {
+        streams
+            .err
+            .appendln("To terminate this program type \"exit\" or \"quit\" in this window,");
```
krobelus, line 143: not sure if the leading blank line here is dropped intentional

src/builtins/shared.rs
```diff
@@ -495,6 +503,7 @@ pub fn builtin_get_desc(name: &wstr) -> Option<&'static wstr> {
         _ if name == "exit" => wgettext!("Exit the shell"),
         _ if name == "false" => wgettext!("Return an unsuccessful result"),
         _ if name == "fg" => wgettext!("Send job to foreground"),
+        _ if name == "fish_key_reader" => wgettext!("explore what characters keyboard keys send"),
```
krobelus, line 506: why is fish_key_reader added here but fish_indent isn't?

src/builtins/fish_key_reader.rs
krobelus: hmm except for `TCSANOW` which we need to keep it seems. Should try to clean this up eventually

src/builtins/fish_indent.rs
faho: Yes, it's in throwing_main, which is only used when run as a command.

src/builtins/shared.rs
faho: Simple: I forgot.

src/builtins/fish_key_reader.rs
faho: I have no idea what it's for, so I dropped it.

src/builtins/fish_key_reader.rs
faho: The important bit here is the shell_modes, which we unset while running a builtin, apparently?

Tbh I'd want to leave this for now, it seems harmless either way.

faho:
> Renaming would be great (also underscores are odd, I'd rather use "-" or no separator).

If we did rename, I would want it to be more substantial change than `fish-indent`, especially given that we use underscores everywhere - `fish_prompt`, not `fish-prompt`.

src/builtins/fish_key_reader.rs
faho: Okay I put it back

faho:
An alternative is to replace `fish_indent` with an option to the main fish binary - `fish --format`.

This would make it easier to call from the outside if you don't have symlinks (e.g. if you installed via `cargo install`).

However, there are some issues with that:

- We would have to add the other fish_indent options like `--check`, `--write`, `--ansi`, `--html` there too
- How do you call it from inside fish? Do we still add a builtin?

The issue with renaming is that we would have to keep the old name around longer in some way, even from the outside.

So we'd leave a `fish_indent` binary and/or symlink in place, but what gets editor plugins to move to the newer version then?

Overall, thinking more about it I don't think it's worth it.

Those are the questions I would like to have answers to before merging this, because that's the easiest time to do so.

zanchey:
Disk is cheap and compression is good (there's essentially no difference in size between a .xz with one or two copies of the same binary, if they're stripped), I'd not bother with a symlink for now. Maybe make a note in the README for distributors who want to symlink instead.

I also don't think there's much to be gained from a rename.

faho:
I've decided to merge this now so we can get some experience in.

There are improvements to be made, but it's already useful.

krobelus:
`echo asdf | fish_indent` hangs while `/bin/echo asdf | fish_indent` works.
Maybe we're not reading from stdin the way a builtin should.

PR #10881, opened 2024-12-03
(WIP) Remove dependencies and use std ones instead

InfyniteHeap:
## Description

Some of containers, like `Lazy` and `OnceCell` which are provided by third-party crate `once_cell`, have their equivalent implementation in standard library: `LazyLock` and `OnceLock`.

Use std rather than third-party crates seems better because std can provide safer and more stable implementations.

The only thing that should be noticed is that the MSRV will be `1.80.0`, because it was this version that first introduced `LazyLock` in std.

## TODOs:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "allocator-api2"
-version = "0.2.18"
+version = "0.2.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5c6cb57a04249c6480766f7f7cef5467412af1490f8d1e243141daddada3264f"
+checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
 name = "autocfg"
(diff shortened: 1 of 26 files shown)


faho:
> The only thing that should be noticed is that the MSRV will be 1.80.0, because it was this version that first introduced LazyLock in std.

That's a NAK.

InfyniteHeap:
> > The only thing that should be noticed is that the MSRV will be 1.80.0, because it was this version that first introduced LazyLock in std.
> 
> That's a NAK.
> 
> We do not want to increase MSRV at this time.
> 
> I'm sorry that you've invested all this work, but that's something you should have asked about before.
> 
> If you want to replace lazy_static *with* once_cell or vice-versa that would be cool.

Ok, I got it. But can I try fixing the `Arc<RefCell<T>>` in `src/env/environment_impl.rs`?

I read the comments above this struct, but I still think this exists a vulnerability about thread safety. `lazy_static!` seems cannot provide the safety guarantee, yet it seems helped this struct to avoid from thread-safe checking. In my view, it can be replaced with `Arc<RwLock<T>>`. What do you think?☺️

faho:
>Ok, I got it. But can I try fixing the Arc<RefCell<T>> in src/env/environment_impl.rs?

If you can do that without increasing MSRV, sure.

If you want to add another dependency, I would suggest asking about that first.

InfyniteHeap:
> >Ok, I got it. But can I try fixing the Arc<RefCell<T>> in src/env/environment_impl.rs?
> 
> If you can do that without increasing MSRV, sure.
> 
> If you want to add another dependency, I would suggest asking about that first.

Ok, I got it!

Issue #9796, opened 2023-05-17
Crashes in prompt cause a repaint loop


faho:
`fish_job_summary` calls `commandline -f repaint`, which, if it is called via something in fish_prompt, causes it to loop.

I found this when I noticed that fossil crashes if $PWD doesn't exist, so the test case is:

```fish
function fish_prompt; fossil branch current; end
mkdir foo
cd foo
rm -rf $PWD
```

It'll end up printing a constant stream of

> fish: Job 1, 'fossil branch current' terminated by signal SIGABRT (Abort)

until you press ctrl-c. While this should be a bug in fossil, the general principle stands.

I'm not sure how to detect this and turn fish_job_summary off.

lavafroth:
Hi. I have been looking into this but I can't seem to reproduce the bug. Perhaps this has been fixed?

faho:
Yup, the repaint was incidentally removed in 9158395d10cfa584a988ba093cf7a93a70b48578

faho:
Ah, turns out testing with SIGKILL doesn't work.

I had used `bash -c 'kill -9 $$'`, which doesn't trigger fish_job_summary. It does with `kill -6`.

ridiculousfish:
The looping is because of this:

```rust
    fn prepare_to_select(&mut self) {
        // Fire any pending events and reap stray processes, including printing exit status messages.
        event::fire_delayed(self.parser);
        if job_reap(self.parser, true) {
            reader_schedule_prompt_repaint();
        }
    }
```

that is, right before we wait for user input, we reap processes and allow printing messages; if a message is printed we schedule a prompt repaint and then react to that instead of select:

```rust
            // We are going to block; but first allow any override to inject events.
            self.prepare_to_select();
            if let Some(mevt) = self.try_pop() {
                return Some(mevt);
            }
```

still unclear what the right approach is. It's worth noting that this issue seems to have existed forever; still if we can fix it we should since it's a very bad state to get into.

krobelus:
Ah so we never even read stdin.
We could defer reaping jobs from prompt generation X until the current prompt generation is X+1.
This change need only apply to jobs spawned within `exec_prompt`.
That should slow down the loop to only move one cycle whenever there is a legit repaint.

I haven't explored all alternatives, one of them would be an interrupt key that we always prioritize over all other events (possibly such keys could replace some signal-handling logic)

ridiculousfish:
I think I have a pretty solid fix (one liner!) in #10891

PR #10882, opened 2024-12-03
fix: give up after some specified attempts of repainting the prompt

lavafroth:
## Description

- Adds a repaint attempts counter to `ReaderData` struct
- The number of retries can bet set using a newly introduced `fish_redraw_attempts` environment variable
- The default value is set to 3
- If the subshell exit status of the prompt command is not zero for `threshold` number of times,
`redraw_attempt.exceeded()` is true and we paint the default prompt

Fixes issue #9796

## TODOs:

### DIFF
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -543,6 +543,26 @@ pub struct ReaderData {
     in_flight_autosuggest_request: WString,
 
     rls: Option<ReadlineLoopState>,
+    redraw_attempt: RedrawAttempt,
+}
+
+pub struct RedrawAttempt {
+    threshold: usize,
+    attempts: usize,
+}
+
+impl RedrawAttempt {
+    fn exceeded(&self) -> bool {
+        self.threshold == self.attempts
+    }
+
+    fn reset(&mut self) {
+        self.attempts = 0;
+    }
+
+    fn record(&mut self) {
+        self.attempts += 1;
+    }
 }
 
 /// Reader is ReaderData equippeed with a Parser, so it can execute fish script.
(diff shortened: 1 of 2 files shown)


faho:
> The number of retries can bet set using a newly introduced fish_redraw_attempts environment variable

Just as a general thing: I wouldn't introduce a variable like that unless we can justify why someone would want to set it.

We try not to introduce options, if we can.

Just hard-code it at 3.

------------

Also I am reasonably sure this will end up calling `fish_job_summary` 3 times, and print the message 3 times, in the case of #9796.

Which is definitely better than looping endlessly, but of course still not *ideal*. I don't know how to achieve that ideal, tho.

src/reader.rs
```diff
@@ -922,6 +942,24 @@ pub fn reader_set_autosuggestion_enabled(vars: &dyn Environment) {
...
 }
 
+/// Change the number of attempts to redraw a prompt before giving up.
+pub fn reader_set_handle_redraw_attempts_change(vars: &dyn Environment) {
+    // We don't need to _change_ if we're not initialized yet.
+    let Some(data) = current_data() else {
+        return;
+    };
+    let attempts: usize = vars
+        .get(L!("fish_redraw_attempts"))
```
faho, line 952: Remove this variable and just hard-code the number of attempts.

lavafroth:
I'll get back to this by tonight.

src/reader.rs
```diff
@@ -4168,16 +4210,20 @@ impl<'a> Reader<'a> {
...
+                    if zelf.redraw_attempt.exceeded() || left_prompt_deleted {
+                        zelf.redraw_attempt.reset();
                         DEFAULT_PROMPT
                     } else {
                         &zelf.conf.left_prompt_cmd
                     },
                     zelf.parser,
                     Some(&mut prompt_list),
-                    /*apply_exit_status=*/ false,
+                    /*apply_exit_status=*/ true,
```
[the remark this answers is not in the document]
lavafroth: Turns out `exec_subshell` returns `$status` of the command executed anyways. Therefore, we would not have to apply the exit status.

src/reader.rs
```diff
@@ -543,6 +543,26 @@ pub struct ReaderData {
     in_flight_autosuggest_request: WString,
 
     rls: Option<ReadlineLoopState>,
+    redraw_attempt: RedrawAttempt,
+}
+
+pub struct RedrawAttempt {
+    threshold: usize,
```
krobelus, line 550: if we stick with this approach: this field is unused and should be removed (which unlocks further simplification)

lavafroth:
Closing this in favor of #10891

PR #10884, opened 2024-12-04
Remove `lazy_static!` and Use `Lazy` Instead

InfyniteHeap:
## Description

`Arc<RefCell<T>>` isn't thread-safe and should be replaced with `Arc<RwLock<T>>`, but `lazy_static!` bypassed this check so that it could probably cause a vulnerability. This PR aims to fix this protential issue.

## TODOs:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -80,7 +80,6 @@ dependencies = [
  "cc",
  "errno",
  "fish-printf",
- "lazy_static",
  "libc",
  "lru",
  "nix",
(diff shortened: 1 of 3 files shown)


ridiculousfish:
It's a good thing to be thinking about, but the code is correct as written. We explicitly do NOT want fine-grained locking for environment nodes, as it makes certain operations non-atomic and risks deadlock due to lock inversion.

That's why we use [a global lock](https://github.com/fish-shell/fish-shell/blob/fcf8ed062810efd3df0e001c8e3e44ad83d34449/src/env/environment_impl.rs#L1097), and that lock is what makes our `Arc<RefCell>` thread-safe.

I agree the code looks odd as written because Rust doesn't have good facilities for expressing data protected by a global lock. If you have ideas for how to improve that, we'd welcome it, but fine-grained locking isn't what we want.

Closing but thank you anyways!

PR #10885, opened 2024-12-04
doc: expand &| reference to full example

frenchy64:
## Description

I'm new to fish and wasn't aware of `|&` in bash. At this point in the tutorial I had to do a web search to understand what `&|` did.

FWIW at this point I tried `print &| | less` because I thought the comment was saying `&|` abbreviated `2>&1` rather than `2>&1 |`.

Built locally and verified with sphinx.

## TODOs:

### DIFF
--- a/doc_src/language.rst
+++ b/doc_src/language.rst
@@ -266,8 +266,8 @@ Consider this helper function::
 Now let's see a few cases::
 
   # Redirect both stderr and stdout to less
-  # (can also be spelt as `&|`)
   print 2>&1 | less
+  print &| less       # equivalent using &|
 
   # Show the "out" on stderr, silence the "err"
   print >&2 2>/dev/null


doc_src/language.rst
```diff
@@ -266,8 +266,8 @@ Consider this helper function::
 Now let's see a few cases::
 
   # Redirect both stderr and stdout to less
-  # (can also be spelt as `&|`)
   print 2>&1 | less
+  print &| less       # equivalent using &|
```
faho, line 270: this is saying `&|` is equivalent to `&|`, which seems redundant?

I would probably do it like this:

```fish
print 2>&1 | less
# or
print &| less
```

faho, line 270: this is saying `&|` is equivalent to `&|`, which seems redundant?

I would probably do it like this:

```fish
print 2>&1 | less
# or
print &| less
```

faho:
Thanks!

PR #10890, opened 2024-12-08
Swap variable overrides and time in not statement

krobelus:
This is allowed

	time a=b echo 123

but -- due to an oversight in 3de95038b0 (Make "time" a job prefix,
2019-12-21) -- this is not allowed:

	not time a=b echo 123

Instead, this one one works:

	not a=b time echo 123

which is weird because without the "not" this would run "/bin/time".

Swap the order for consistency.


faho (approved):
LGTM

PR #10891, opened 2024-12-08
Fix infinite prompt loop if status message is printed in prompt

### LINKED ISSUE #9796: Crashes in prompt cause a repaint loop

ridiculousfish:
fish will print messages for some jobs when they exit abnormally, such as with SIGABRT. If a job exits abnormally inside the prompt, then (prior to this commit) fish would print the message and re-trigger the prompt, which could result in an infinite loop. This has existed for a very long time.

Fix it by reaping jobs after running the prompt, and NOT triggering a redraw based on that reaping. We still print the message but the prompt is not executed.

Add a test.

Fixes #9796

### DIFF
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -4204,6 +4204,11 @@ impl<'a> Reader<'a> {
         // hack to work.
         reader_write_title(L!(""), zelf.parser, false);
 
+        // Reap jobs but do NOT trigger a repaint.
+        // This is to prevent infinite loops in case a job from the prompt triggers a repaint.
+        // See #9796.
+        job_reap(zelf.parser, true);
+
         // Some prompt may have requested an exit (#8033).
         let exit_current_script = zelf.parser.libdata().exit_current_script;
         zelf.exit_loop_requested |= exit_current_script;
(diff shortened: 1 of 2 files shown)


src/reader.rs
```diff
@@ -4204,6 +4204,11 @@ impl<'a> Reader<'a> {
         // hack to work.
         reader_write_title(L!(""), zelf.parser, false);
 
+        // Reap jobs but do NOT trigger a repaint.
+        // This is to prevent infinite loops in case a job from the prompt triggers a repaint.
+        // See #9796.
+        job_reap(zelf.parser, true);
+
```
krobelus, line 4211: LGTM, this will fix it for foreground jobs in the prompt, which is the most common scenario (I can't think of a legit case for a background job that's not disowned).

PR #10893, opened 2024-12-09
Crash reporting

krobelus:
Add an interactive prompt to the panic handler, as an easy way to
report info like panic messages, stacktraces and version info, via Sentry
(https://sentry.io).  I made the [fish-shell](https://fish-shell.sentry.io/)
organization.  Some features require subscription; we can probably ask for a
[sponsored account](https://sentry.io/for/open-source/).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,30 @@
 version = 3
 
 [[package]]
+name = "addr2line"
+version = "0.24.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
+dependencies = [
+ "gimli",
+]
+
+[[package]]
+name = "adler2"
+version = "2.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
+
+[[package]]
+name = "aho-corasick"
+version = "1.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8e60d3430d3a69478ad0993f19238d2df97c507009a52b3c10addcd7f6bcb916"
+dependencies = [
+ "memchr",
+]
+
+[[package]]
 name = "allocator-api2"
 version = "0.2.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


faho:
To be honest I don't like it.

I don't believe we've had enough crashes to really be worth it, and adding uploading functionality to a proprietary service scares me.

Xiretza:
In addition to what @faho said, I'm not very confident in our ability to make an HTTPS request from the same process after we've crashed. That's a mighty complex operation, and as a shell we do some weird things as part of normal execution that can have even weirder effects when interrupted.

krobelus:
> I don't believe we've had enough crashes to really be worth it

It's true that we haven't had a lot of crashes in released versions so
far. Depends on our definition of worth it, I'd say a few fixable crashes
a year is already great. I'm not sure how we estimate the cost, I haven't
yet looked at how heavy this dependency is.  Rust is more likely to trigger
panics where C++ would have UB.

Some of our peer projects (that were written from scratch in Rust) have
panics that are easy to trigger but most users probably won't put in the
effort to report it.

> adding uploading functionality to a proprietary service scares me.

Why? Does the proprietary bit make a difference?
I think there are free alternatives; ideally crash reports are public.
(Even if the server claims to run non-proprietary code it can't really be trusted but that's not necessary.)

> I'm not very confident in our ability to make an HTTPS request from the
> same process after we've crashed. That's a mighty complex operation, and
> as a shell we do some weird things as part of normal execution that can
> have even weirder effects when interrupted.

I can't think of anything fish-specific that makes it harder (we set signal
handlers and call terminal APIs, both shouldn't matter).

It's true that we might panic while holding a lock, poisoning the mutex.
If crash reporting needs to take the same lock, that would be a problem.

If we think that resource cleanup would help, we could unwind first.

---

One potential issue is that distro packages use "strip" to remove debug
information; (for precise backtraces) my plan is to compensate by integrating
something like `debuginfod`.

krobelus:
The goal is to show the user exactly what would be sent, in plaintext, and allow editing it.
The current patch doesn't really do that.

faho:
Okay, both @Xiretza and I don't like this.

We've now released the 4.0 beta, and it doesn't appear to be particularly crash prone, so I'm not sure if this would have given us good info to begin with.

@krobelus, do you want to continue pushing for this or can it be closed?

krobelus:
> do you want to continue pushing for this or can it be closed?

I feel people sometimes conflate criticism of a suboptimal
implementation with "please don't touch this".

So closing it myself is counterproductive because it might cause
future action to improve things in this area to be shut down as
"previously rejected".

I have some points (looking at what other projects do) but I don't
think it's a good use of my time to be arguing when that's unlikely
to achieve anything.

Feel free to manage your issue tracker however it helps you do the
job you want to do. Again: different perspectives.

faho:
Okay, this has conflicts and the implementation is controversial. It does not appear to be moving forward at this point.

I'm closing this because of that. Crash reporting in general is something we can discuss, but it would have to deal with the criticisms raised (a third-party proprietary platform that we can use because of their goodwill, additional dependencies, https support in the shell, privacy concerns).

That can just be by talking about it and deciding that it's worth it (and auditing the received data for potential private information and getting the user's consent).

PR #10896, opened 2024-12-10
Add `status buildinfo`

faho:
This can be used to get some information on how fish was built - the version, the build system, the operating system and architecture, the features.

The idea is that we can tell people to run `status buildinfo | fish_clipboard_copy` and copy it into bug reports.

Example output:

```output
> ./fish -c 'status buildinfo'
Build system: CMake
Version: 3.7.1-2571-gaadcb292b
Target (and host): x86_64-unknown-linux-gnu
Profile: release
Features: gettext

> ../target/x86_64-unknown-linux-musl/debug/fish -c 'status buildinfo'
Build system: Cargo
Version: 3.7.1-2571-g79aee4d60
Target: x86_64-unknown-linux-musl
Host: x86_64-unknown-linux-gnu
Profile: debug
Features: installable
```

This seems like it might be useful especially since we have multiple build systems and the potential for cross-compilation now.

The exact set of things to print isn't set in stone, of course, this was just what I came up with.

An alternative is to make a `status debuginfo` command that is a bit broader in scope, and also prints e.g. terminal or locale information, possibly even the operating system.

Importantly, this should *not* collect information like the time it was compiled to not break build reproducability, nor should it (for obvious reasons) collect personal information.

## TODOs:

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -29,6 +29,11 @@ fn main() {
             .unwrap(),
     );
 
+    // Some build info
+    rsconf::set_env_value("BUILD_TARGET_TRIPLE", &env::var("TARGET").unwrap());
+    rsconf::set_env_value("BUILD_HOST_TRIPLE", &env::var("HOST").unwrap());
+    rsconf::set_env_value("BUILD_PROFILE", &env::var("PROFILE").unwrap());
+
     let version = &get_version(&env::current_dir().unwrap());
     // Per https://doc.rust-lang.org/cargo/reference/build-scripts.html#inputs-to-the-build-script,
     // the source directory is the current working directory of the build script
--- a/cmake/Rust.cmake
+++ b/cmake/Rust.cmake
@@ -49,6 +49,8 @@ set(VARS_FOR_CARGO
     "PREFIX=${CMAKE_INSTALL_PREFIX}"
     # Temporary hack to propogate CMake flags/options to build.rs.
     "CMAKE_WITH_GETTEXT=${CMAKE_WITH_GETTEXT}"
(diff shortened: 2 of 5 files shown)


mqudsi:
fyi I previously added separate status modes `target` and `target-arch` as part of https://github.com/fish-shell/fish-shell/pull/10582 but as they were incidental to the fix rather than an intentional feature (though I found them valuable in all cases) @ridiculousfish was of the opinion that they should be removed.

+1 from me :)

faho:
>  @ridiculousfish was of the opinion that they should be removed.

I'm aware, I understood the issue to be that you added an undocumented command to provide this information for use in one insignificant test.

I want to have it as a proper thing. Documented and meant to be used.

The impetus here is that I'm expecting some amount of confusion about whether to build with cargo or cmake. And so I would at least want to have an easy way to figure that out.

mqudsi:
Suggestion: I would add at least `crt-static` to the features list, possibly `pcre2-static` as well.

faho:
> Suggestion: I would add at least crt-static to the features list, possibly pcre2-static as well.

crt-static: Sure, why not

pcre2-static: I don't know how to do that. It's not a feature of the fish crate, it's one of the pcre2-sys crate that's decided when that is compiled. I also don't think it's going to help because I don't think it's a source of problems.

faho:
Okay, I'd like to get this one in for 4.0, because it'll be useful in case someone e.g. manages to accidentally build as self-installable.

faho:
Cherry-picked to 4.0 as 5f76fc3e414d39a554402250bfc5af500e0190fa

PR #10898, opened 2024-12-11
Allow { } for command grouping, like begin / end

### LINKED ISSUE #10895: Command grouping with `{ }`
TL;DR: add command grouping with `{}` (NOT `()`) to improve Fish efficiency and facilitate adoption

## Feature request
Command grouping using `{}` is a widely used feature of POSIX shells
- In error handling:
  - `command || { echo "failed; exit 1 }`
- To compose append command in a pipe chain:
  - `{ head -n1 csv; grep "key" csv } | nc localhost 123`
- To handle fallback:
  - `cmd1 && { cmd2 || cmd3 } && cmd4`

## Workaround
Fish offer the equivalent `begin; end`, nevertheless:
- it is long, which it is not convenient for the command line
- it is different than `{}` which shell users have been using for >30 years

## Backward compatibility
Fish is currently not using `{}` for anything other than [braces expansions](https://fishshell.com/docs/current/language.html#expand-brace) (just like POSIX) so there should be no backward compatibility issues.

## This issue does NOT advocate for `()`
Posix shell also uses `()` for command grouping, with the major difference that it spawns a subshell.
The author believes `()` not to be a good fit for Fish as:
- users rarely care about subshell, and most POSIX usage of `()` should be replaced by `{}`
  - subshell should not be encouraged, especially in Fish
- it would conflict with Fish command expansion (`()` <=> `$()`)
- it was already rejected in #8083

`{}` has none of those disadvantages.

krobelus:
Closes #10895

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -3089,6 +3089,18 @@ impl<'s> NodeVisitorMut for Populator<'s> {
         };
 
         if let Some((header_kw_range, missing_end, enclosing_stmt)) = header {
+            let next_token = self.peek_token(0);
+            if next_token.typ == ParseTokenType::string
+                && matches!(
+                    next_token.keyword,
+                    ParseKeyword::kw_case
+                        | ParseKeyword::kw_else
+                        | ParseKeyword::kw_end
+                        | ParseKeyword::kw_rbrace
+                )
+            {
+                self.consume_excess_token_generating_error();
+            }
             parse_error_range!(
                 self,
                 header_kw_range,
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -1255,7 +1255,7 @@ impl CheckParse for JobConjunction {
         token.typ == ParseTokenType::string
             && !matches!(
                 token.keyword,
-                ParseKeyword::kw_end | ParseKeyword::kw_else | ParseKeyword::kw_case
+                ParseKeyword::kw_case | ParseKeyword::kw_end | ParseKeyword::kw_else
             )
     }
 }
@@ -3337,6 +3337,14 @@ impl<'s> Populator<'s> {
             ParseTokenType::string => {
                 // There are three keywords which end a job list.
                 match tok.keyword {
+                    ParseKeyword::kw_case => {
+                        parse_error!(
+                            self,
+                            tok,
+                            ParseErrorCode::unbalancing_case,
+                            "'case' builtin not inside of switch block"
+                        );
+                    }
                     ParseKeyword::kw_end => {
                         parse_error!(
                             self,
@@ -3351,14 +3359,6 @@ impl<'s> Populator<'s> {
                             tok,
                             ParseErrorCode::unbalancing_else,
                             "'else' builtin not inside of if block"
-                        );
-                    }
-                    ParseKeyword::kw_case => {
-                        parse_error!(
-                            self,
-                            tok,
-                            ParseErrorCode::unbalancing_case,
-                            "'case' builtin not inside of switch block"
                         );
                     }
                     _ => {
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -49,12 +49,12 @@ impl<T: Acceptor> Acceptor for Option<T> {
     }
 }
 
-pub struct MissingEndError {
+pub struct MissingEndKeywordError {
     allowed_keywords: &'static [ParseKeyword],
     token: ParseToken,
 }
 
-pub type VisitResult = ControlFlow<MissingEndError>;
+pub type VisitResult = ControlFlow<MissingEndKeywordError>;
 
 trait NodeVisitorMut {
     /// will_visit (did_visit) is called before (after) a node's fields are visited.
@@ -202,6 +202,12 @@ pub trait ConcreteNode {
     fn as_block_statement(&self) -> Option<&BlockStatement> {
         None
     }
+    fn as_brace_header(&self) -> Option<&BraceHeader> {
+        None
+    }
+    fn as_brace_statement(&self) -> Option<&BraceStatement> {
+        None
+    }
     fn as_if_clause(&self) -> Option<&IfClause> {
         None
     }
@@ -318,6 +324,12 @@ trait ConcreteNodeMut {
         None
     }
     fn as_mut_block_statement(&mut self) -> Option<&mut BlockStatement> {
+        None
+    }
+    fn as_mut_brace_header(&mut self) -> Option<&mut BraceHeader> {
+        None
+    }
+    fn as_mut_brace_statement(&mut self) -> Option<&mut BraceStatement> {
         None
     }
     fn as_mut_if_clause(&mut self) -> Option<&mut IfClause> {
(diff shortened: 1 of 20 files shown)


mqudsi (commented):
Look at the code, I'm not sure if there's anything that would stop you from changing Begin/End to be enums instead of structs with variants Regular or Symbolic? That would cut down on the boilerplate for this change considerably.

My bigger concern resides with which cases, if any, conflict with our existing braces-for-expansion support. Is there ever a case where `{` would be anywhere but in the command position yet still imply grouping rather than expansion? And the corollary to that, is there anywhere non-command-only where we *would* want {} for grouping but are prevented from doing so due to the existing expansion case?

krobelus:
> Look at the code, I'm not sure if there's anything that would stop you
> from changing Begin/End to be enums instead of structs with variants
> Regular or Symbolic? That would cut down on the boilerplate for this
> change considerably.

I used to do that (because it's easy to implement); one problem is that a
naive implementation would allow `begin; }` and `{; end`.

It's possible to fix that by checking for this error after parsing the rogue statement.
But that would probably mean we'd need to add another hack to make sure the indents are

	begin
	    }

and not

	begin
	}

I'm not 100% sure and I'm happy to try this (it should not change behavior).
But my feeling is that we'll end up with a few subtle hacks that might be
worse than the code cloning here.

> My bigger concern resides with which cases, if any, conflict with our
> existing braces-for-expansion support.

> Is there ever a case where { would be anywhere but in the command position
> yet still imply grouping rather than expansion?

It should be only special in command position. (I'll clean up the logic to
implement this but the behavior should be okay.)

Since we only group if the command is exactly `{`, we can still use it in the command name.

	$ {echo,echo}
	echo
	$ e{cho,cho,cho,cho}
	echo echo echo

So this change only breaks odd cases where brace expansion is used with a
space/semicolon after the brace:

	$ { echo,foo; }
	# used to print "foo;", now «Unknown command "echo,foo"»
	
	$ function ';'; echo $argv; end
	$ {;,foo; }
	# same as above, used to print foo;, now «Unknown command ";,foo"»

> And the corollary to that, is there anywhere non-command-only where we
> would want {} for grouping but are prevented from doing so due to the
> existing expansion case?

yeah, the two above examples are inevitable conflicts. I think starting a brace
expansion with "{ " is rare, which is why I think it's okay to pick grouping.

mqudsi:
Thinking about this some more and I think it just *might* be a shame to ship this without expanding the usecases a little just in case we are prevented from doing so in the future by backwards compatibility constraints.

Specifically, I would like to propose that if we're going to do this we take it a moment to consider if we don't want to this a step further and support syntax like `if foo { ... }` and `while true { ... }`

Keeping `{` as `begin` and `}` as `end` would mean the parser can't directly distinguish between 

```
if foo
    begin
        foo
    end
```

and 

```
if foo {
    begin
        foo
    end
}
```

leading to unbalanced expressions.

Of course this might already be impossible to support, but I have a feeling it's possible.

src/ast.rs
```diff
@@ -1404,6 +1422,62 @@ impl ConcreteNodeMut for BlockStatement {
     }
 }
 
+#[derive(Default, Debug)]
+pub struct BraceHeader {
+    parent: Option<*const dyn Node>,
+    pub kw_lbrace: KeywordLBrace,
+    /// Note that '{' does NOT require a semi or nl afterwards.
```
ridiculousfish, line 1429: By a historical quirk the same is true for `begin`; it would be good to add a similar comment there.

src/builtins/shared.rs
```diff
@@ -466,6 +474,8 @@ pub fn builtin_get_desc(name: &wstr) -> Option<&'static wstr> {
         _ if name == "." => wgettext!("Evaluate contents of file"),
         _ if name == ":" => wgettext!("Return a successful result"),
         _ if name == "[" => wgettext!("Test a condition"), // ]
+        _ if name == "{" => wgettext!("Create a block of code"), // } {
+        _ if name == "}" => wgettext!("Create a block of code"),
```
ridiculousfish, line 478: Should be `"End a block of commands"` or similar to match `end`

src/tests/parser.rs
ridiculousfish, line 1: It would be good to add tests for the following cases, either here or in a littlecheck .fish test:

```
begin ; echo hi;  }  # should error
echo { 1 2 }           # should output '{ 1 2 }' - this differs from zsh!
{echo,hello}         # should output 'hello' - another difference from zsh
```

src/highlight.rs
```diff
@@ -1399,6 +1402,11 @@ impl<'s> Highlighter<'s> {
         self.visit(&block.end);
         self.pending_variables.truncate(pending_variables_count);
     }
+    fn visit_brace_statement(&mut self, brace_statement: &BraceStatement) {
```
ridiculousfish, line 1405: This is missing:

```
        self.visit(&brace_statement.header);
```

otherwise the initial `{` isn't highlighted.

ridiculousfish (approved):
This is great. I'm in favor of this change and I can't see any problems with adding this. I like the explicit treatment.

Let's hold off until post 4.0 since it's a large change.

krobelus:
> Let's hold off until post 4.0 since it's a large change.

that is not a reliable statement since we don't know when either will be ready.
Usually things can go in if they are ready before the merge window closes, no need to spend more effort on deciding. There are a lot other, riskier changes that do not get this special treatment.

The tokenizer hack means that `and {` doesn't work yet.
We could move subcommand keyword detection to the parser,
but it sounds much simpler to do something closer to other shells:
treat `{` as keyword (and not brace expansion) iff followed by space or semicolon/newline.
That breaks only unreasonable use cases.
We can still allow `{foo, bar}` which bash/zsh do not. 

> support syntax like `if foo { ... }` and `while true { ... }`

those are fine to add without conflict;
other cases like `if foo; { ... }` will conflict but we can throw a syntax error until we implement them.

As far as sigils are concerned, fish has been mostly a subset of POSIX, not a superset (a the notable exception is the frequently-used `()` command substitution).
I think this is generally good because it encourages users to stick to simple, orthogonal commands.
In some ambiguous cases, fish throws an error, which helps learning.

Since `if foo { ... }` is not really ambiguous it's probably fine to add.
Compound conditions may look like `if { ... } { ... }` which may be confusing but still fine.
`if foo { ... }` is already invalid today (it's not unequivocally good that we're deviating from POSIX on `echo {` but I think the difference hardly ever matters in practice).

So yeah, I think it would be worth trying.
Not yet sure how it interacts with other constructs.
Docs will be a bit more confusing I guess.
For the average user, braces may be harder to type on some non-English keyboard layouts (no idea if that's relevant).

POSIX shell also allows pretty if statements:

```shell
if true; then {
	echo some commands
} fi
```

The fact that `} fi` works is quite magical, I'll try to do that here too.

faho:
I agree that it should not go in for 4.0.

dpc:
Oh, this will be implemented? Cool. BTW. I think if you use `Fix #xyz` in the PR it will close the issue as "Solved" instead of "Closed".

PR #10899, opened 2024-12-11
Ask to install installable builds on interactive launch

faho:
## Description

One downside installable builds currently have is the need to run `fish --install` every so often.

This makes it so interactive runs (real interactive runs, `fish -i` doesn't suffice, you need a terminal so we can ask) do that automatically, including asking for it.

That means you can do

```terminal
> ~/.cargo/bin/fish
Warning: Fish's asset files are missing or out of date. Trying to install them.
This will write fish's data files to '/home/alfa/.local/share/fish/install'.
Please enter 'yes' to continue.
> yes
# and the prompt shows, we're in a session
```

You can still decline (by entering anything other than "yes"). If the files are missing, it will run without reading any config - this is quite uncomfortable, it won't even have the regular keybindings. (if they are outdated, things will "most likely" work out okay, but it will ask again next run)

Notably, for non-interactive runs, this won't install automatically, and it won't even complain about *outdated* files anymore.

If the files don't exist, on the other hand, it will print an error and then run.

## TODOs:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -55,10 +55,7 @@ Notable improvements and fixes
     cargo install --path . # in a clone of the fish repository
     # or `cargo build --release` and copy target/release/fish{,_indent,_key_reader} wherever you want
 
-    # and then, wherever you use it, run
-    /path/to/fish --install # or --install=noconfirm for non-interactive use
-
-  This will extract all the data files to (currently) ~/.local/share/fish/install/. To uninstall, remove the fish binaries and that directory.
+  The first time it runs interactively, it will extract all the data files to (currently) ~/.local/share/fish/install/. To uninstall, remove the fish binaries and that directory.
 
   This configuration is experimental.
   It does not affect the main configuration, which is a regular install via ``cmake``.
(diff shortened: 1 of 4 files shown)


faho:
Okay, some spicy changes that I hope make this nicer to use:

- `fish --install` now installs without asking, and does not exit
- `fish --install=noconfirm` is gone

That means you install or update this. If you just run `fish`, it will ask to install. If you answer yes, it will install. If you answer something else, it will still try to run, but if the files are missing it will skip all configuration.

If you want to skip the confirmation, you can run `fish --install`. This will still run fish, so it will give you an interactive session or run a script. If you *just* want to install the files, you can do `fish --install -c exit`.

Note: If fish wasn't built as installable, `fish --install` will *still run fish* after printing an error that it wasn't built as installable.

In total, what I hope this does is remove the need for people to *think* about fish being installable - they're just asked to update the assets.

I'm sure this won't be the last of it, but I'm reasonably happy with this for now.

faho:
Alright, I decided to just merge this now, we can work on it later.

PR #10905, opened 2024-12-14
curses: Just use our hardcoded xterm-256color entry as the fallback

faho:
The values we would try are:

xterm-256color, xterm, ansi, dumb

This is a pretty useless list, because systems without "xterm-256color" but with "ansi" basically don't exist, and it is very likely that the actual terminal is more xterm-compatible than it is ansi.

So instead we just use our xterm-256color definition, which has a high likelihood of being basically correct.

### DIFF
--- a/src/env_dispatch.rs
+++ b/src/env_dispatch.rs
@@ -501,62 +501,6 @@ fn update_fish_color_support(vars: &EnvStack) {
     crate::output::set_color_support(color_support);
 }
 
-/// Try to initialize the terminfo/curses subsystem using our fallback terminal name. Do not set
-/// `$TERM` to our fallback. We're only doing this in the hope of getting a functional shell.
-/// If we launch an external command that uses `$TERM`, it should get the same value we were given,
-/// if any.
-fn initialize_curses_using_fallbacks(vars: &EnvStack) {
-    // xterm-256color is the most used terminal type by a massive margin, especially counting
-    // terminals that are mostly compatible.
-    const FALLBACKS: [&str; 4] = ["xterm-256color", "xterm", "ansi", "dumb"];
-
-    let current_term = vars
-        .get_unless_empty(L!("TERM"))
-        .map(|v| v.as_string())
-        .unwrap_or(Default::default());
-
-    let mut success = false;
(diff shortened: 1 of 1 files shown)


faho:
I am *this close* to just ditching any and all reading of the terminfo database.

I have [a script](https://gist.github.com/faho/db2d7630a6d25492245c2340858c22de) that tries to figure out how different terminfo entries actually are, and the answer is they barely are.

The trick is that you need to filter out a lot of entries - those for 1980s hardware terminals and those invented by ncurses (as well as various variants that just repeat the same information).

E.g. there is an "ms-terminal" entry for Microsoft's terminal, which is useless because that terminal uses xterm-256color and to a first approximation *nobody* sets $TERM (nor should they). It is also just flat out incorrect in some ways, like claiming that "cursor_down" is `\E[B`. It isn't, it behaves like xterm.

Or the tmux entry which claims that exit_attribute_mode is `\E[m^O` when the xterm sequence just works. Or the rxvt entry, which *does* have some legitimate differences in terms of keys, but also has the two parts of exit_attribute_mode swapped - which is not a functional difference.

I am extremely disillusioned with the entire terminfo system and would like to get away from it.

krobelus:
Agreed, thanks.

terminfo seems like a needless indirection; instead of standardizing a terminfo sequence-name it's easier to standardize the actual escape-sequences, because that one needs to exist regardless.

I don't know much about this; if there is a chance a user wants to keep terminfo for now (as workaround), we could start with an `ignore-terminfo` feature flag? I guess we might find that when we turn it on, things will already work better on average across the userbase, since we'll no longer read wrongly configured values. It's probably fine to change it on `master`, as a trial balloon

krobelus (approved):

zanchey:
> I have [a script](https://gist.github.com/faho/db2d7630a6d25492245c2340858c22de) that tries to figure out how different terminfo entries actually are, and the answer is they barely are.

kitty is the only modern terminal I know of that maintains a terminfo.

krobelus:
I think foot too, and both support querying terminfo capabilites via `XTGETTCAP`.
Querying the terminal directly instead of reading a terminfo file seems like the best approach.

PR #10906, opened 2024-12-14
rclone.fish: assume a recent version if the version cannot be parsed

ilyagr:
The version of rclone is set during compilation and could be any crazy string depending on the packager, whether it's a dev build, etc. If it cannot be parsed, let's assume a recent version.

Follows up on cc8fa0f7808130ece03f1c2ac0c95f9ede2aaa1d. Cc @faho 

## TODOs:

### DIFF
--- a/share/completions/rclone.fish
+++ b/share/completions/rclone.fish
@@ -1,10 +1,14 @@
-set -l rclone_version (rclone version | string match -rg 'rclone v?(.*)' | string split .)
-or return
-
-# Yes, rclone's parsing here has changed, now they *require* a `-` argument
-# where previously they required *not* having it.
-if test "$rclone_version[1]" -gt 1; or test "$rclone_version[2]" -gt 62
-    rclone completion fish - 2>/dev/null | source
+if set -l rclone_version (rclone version | string match -rg 'rclone v?(.*)' | string split .) &&
+        test "$rclone_version[1]" -lt 1 ||
+        test "$rclone_version[1]" -eq 1 &&
+        test "$rclone_version[2]" -le 62
+  # version is definitely <= 1.62, adding a `-` would be an error
+  rclone completion fish
 else
-    rclone completion fish 2>/dev/null | source
-end
+  # For newer versions, this requires an `-`. Without a `-`, it would
+  # try to write to /etc/completions/fish.
+  # If we can't determine the version, assume a recent one. An error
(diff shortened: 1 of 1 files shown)


mqudsi:
Can you give an example of an actual in-the-wild version string that currently doesn't parse, mainly just for posterity?

ilyagr:
I don't have an example in the wild, but the version can be [any string](https://github.com/NixOS/nixpkgs/pull/365264/commits/58d62863d8b6e91f416f847f0166fa32325fd7f6#diff-608e307d75a52f5d22e45895a23cf092eca725b245e0194945ff81ff46c14390R50), whatever the packager sets. So, I thought it's better to be more robust.

I initially thought that it'd be empty if that flag is not specified, which would have made this PR more important. It does have a default value, though.

It is of course up to you whether to use this.

share/completions/rclone.fish
```diff
@@ -1,5 +1,7 @@
-set -l rclone_version (rclone version | string match -rg 'rclone v?(.*)' | string split .)
-or return
+set -l rclone version_str (rclone version 2> /dev/null)
+or return 1
+set -l rclone_version (string match -rg 'rclone v?(.*)' "$rclone_version_str" | string split .)
+or set -l rclone_version 99 99  # Could be a dev build, assume a recent version
```
krobelus, line 4: can we do this instead

````c++
begin
    rclone completion fish - || rclone completion fish
end 2>/dev/null | source
```

ilyagr: If `rclone completion fish -` exits with an error for some unrelated reason on a recent version of `rclone`, `rclone completion fish` would (insanely) try to install the completions to `/etc/fish/completions`.

krobelus: how about

```fish
rclone completion fish (
    if set -l rclone_version (rclone version | string match -rg 'rclone v?(.*)' | string split .) &&
        test "$rclone_version[1]" -lt 1 ||
        test "$rclone_version[1]" -eq 1 &&
        test "$rclone_version[2]" -le 62
        # version is definitely <= 1.62, use the old style
    else
        # assume new
        echo -
    end
) 2>/dev/null | source
```

krobelus: no need for `type -q rclone` since fish already checks that before sourcing the completion

> Actually, I'd find the following more readable

sure yeah that's better

PR #10909, opened 2024-12-16
Completions for `mksquashfs`

metamuffin:
## Description

The PR adds competions for the `mksqashfs` command from [`squashfs-tools`](https://github.com/plougher/squashfs-tools) that is available on most distros.

Option descriptions were mostly take from --help, truncated to the most important part then edited or sometimes rewritten to fit.


mqudsi (commented):
Thanks, this looks good! FY, we typically have descriptions start with a capital letter.

The fixme thing is *fine*. We should probably change `__fish_seen_argument` to support n arguments that are checked for in succession, but that might still be too limiting and it's very much a hack so no one has been enthused to take that on. 

However, in your case since you are not redefining `-commonArg` to be, for example, one thing in the case of `gzip` and another in case `lz4`, I don't know if you want/need to go this route at all. The way you have it coded right now, you won't get completions for `-Xhc` (at all, with or without the description) until you've already entered `-comp lz4` -- which is *probably* the most common use case, but just something to be aware of in case it locks out a usecase that people might be using.

metamuffin:
Thanks for the review.

You are right. The argument order does not matter and so that is valid. I changed it such that it always shows all compressor options and only for -Xcompression-level checks the previous arguments to specialize the description to that compressor.

I also uppercased the descriptions. The phrasing of those is kind of inconsistent though: "Set" vs "Sets" and sometimes just a noun. But mksquashfs' --help does that too, so I guess its fine.

mqudsi (approved):

PR #10917, opened 2024-12-18
simplified compute_gaps

sharma-shray:
## Description

Talk about your changes here.

Fixes issue #

## TODOs:

### DIFF
--- a/src/bin/fish_indent.rs
+++ b/src/bin/fish_indent.rs
@@ -162,39 +162,36 @@ impl<'source, 'ast> PrettyPrinter<'source, 'ast> {
 
     // Return the gap ranges from our ast.
     fn compute_gaps(&self) -> Vec<SourceRange> {
-        let range_compare = |r1: SourceRange, r2: SourceRange| {
-            (r1.start(), r1.length()).cmp(&(r2.start(), r2.length()))
-        };
-        // Collect the token ranges into a list.
-        let mut tok_ranges = vec![];
-        for node in Traversal::new(self.ast.top()) {
+        let mut tok_ranges: Vec<SourceRange> = Traversal::new(self.ast.top())
+        .filter_map(|node| {
             if node.category() == Category::leaf {
-                let r = node.source_range();
-                if r.length() > 0 {
-                    tok_ranges.push(r);
+                let range = node.source_range();
+                if range.length() > 0 {
+                    Some(range)
+                } else {
+                    None
                 }
+            } else {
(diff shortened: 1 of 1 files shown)


mqudsi:
I think this PR hasn't been touched because it's not immediately obvious if there are any subtle changes to the algorithm. 

Do you mind adding some unit tests to illustrate that behavior is unchanged?

mqudsi (changes requested):
I think this PR hasn't been touched because it's not immediately obvious if there are any subtle changes to the algorithm.

Do you mind adding some unit tests to illustrate that behavior is unchanged?

src/bin/fish_indent.rs
```diff
@@ -162,39 +162,36 @@ impl<'source, 'ast> PrettyPrinter<'source, 'ast> {
...
+                if range.length() > 0 {
+                    Some(range)
+                } else {
+                    None
                 }
+            } else {
+                None
             }
+        })
+        .collect();
```
epidemian, line 178: A little idea in regards to code simplification: using some shorter `.map()`s and `.filter()`s instead of a general `.filter_map()`, those dangling `else { None }`s could be avoided

```rust
        let mut tok_ranges: Vec<SourceRange> = Traversal::new(self.ast.top())
            .filter(|node| node.category() == Category::leaf)
            .map(|leaf| leaf.source_range())
            .filter(|range| range.length() > 0)
            .collect();
```

I find this a bit easier to follow than the nested `if`s :)

(BTW, i cannot vouch for the correctness of these changes; this is just a drive-by comment.)

src/bin/fish_indent.rs
```diff
@@ -162,39 +162,36 @@ impl<'source, 'ast> PrettyPrinter<'source, 'ast> {
...
+
+    tok_ranges.push(SourceRange::new(self.state.source.len(), 0)); // Add end marker.
+
+    assert!(tok_ranges.is_sorted_by(|x, y| Some((x.start(), x.length()).cmp(&(y.start(), y.length())))));
```
epidemian, line 182: Oh, and if this is comparing tuples of the same things, `(range.start(), range.length())`, it should be the same to use `.is_sorted_by_key`:

```suggestion
    assert!(tok_ranges.is_sorted_by_key(|r| (r.start(), r.length())));
```

faho:
No reaction for a while and merge conflicts with master, closing.

If you get back to it we can reopen.

Issue #10916, opened 2024-12-18
Fish Beta Version: Missing Asset Files Warning on Startup Despite Installation


MrAru:
Hi developers, I'm thrilling with the beta version of fish release and I immediately tried it today.  I download the static build from github release (https://github.com/fish-shell/fish-shell/releases/download/4.0b1/fish-static-linux-x86_64.tar.xz), decompress the tar file and move `fish*` executable files into `~/.local/bin`.

When I switch to the beta version of fish, the fish shell keeps displaying the following message during startup (sorry for the extensive redaction, as my use case is on a company server).

```
PS C:\Users\...> ssh ...@xxx.xx.xx.xx
Warning: Permanently added 'xxx' (ECDSA) to the list of known hosts.
Last login: Wed Dec 18 13:38:24 2024 from xxx

... in ~                                                                                                                                                                                               at 21:51:53
❯ echo $SHELL
/bin/bash

... in ~                                                                                                                                                                                               at 21:51:58
❯ which fish
~/.local/bin/fish

... in ~                                                                                                                                                                                               at 21:52:02
❯ fish
warning: Fish's asset files are missing. Trying to install them.
Can't find $HOME
error: Fish cannot find its asset files in '/mnt/lustre/home/.../.local/share/fish'.
Refusing to read configuration because of this.
error: If you installed via `cargo install`, please run `fish --install` and restart fish.
```

This prompt appears every time I start a new shell. I also tried `fish --install` but nothing changed. Strangely, whether in bash or fish, `echo $HOME` always works, and it returns my home path on lustre:

```
> echo $HOME
/mnt/lustre/home/...
```

Could you suggest any directions for troubleshooting? Thank you very much. My environment is as follows:

```
> fish --version
fish, version 4.0b1-4-g687001f8b
> echo $version
4.0b1-4-g687001f8b

> uname -a
Linux ... 4.18.0-425.3.1.el8.x86_64 #1 SMP Wed Nov 9 20:13:27 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

> echo $TERM
xterm-256color
```

Btw, when I run `sh -c 'env HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_DIRS= fish'`, the same prompt shows again so I cannot provide this info.

faho:
The error is this:

> Can't find $HOME

[...]

Issue #10748, opened 2024-09-27
__fish_sysconf_dir points to /usr/etc/fish


moritz-meier:
fish, version 3.7.0
Fedora Silverblue 40.20240926.0

When installed by RPM on Fedora Silverblue, `__fish_sysconf_dir` points to `/usr/etc/fish` and not to `/etc/fish`.

Fedora Silverblue is an atomic distro where `/usr/etc `and `/etc `exists. `/usr/etc `contains the original files from the installed packages and is read-only, `/etc` contains the writeabe package config files.

When i install fish in a normal (non-atomic) Fedora container I get:

```
config: exec_path: '/usr/bin/fish', argv[0]: 'fish'
config: Using compiled in paths:
config: determine_config_directory_paths() results:
paths.data: /usr/share/fish
paths.sysconf: /etc/fish
paths.doc: /usr/share/doc/fish
paths.bin: /usr/bin

```

On Fedora Silverblue i get:

```
config: exec_path: '/usr/bin/fish', argv[0]: 'fish'
config: determine_config_directory_paths() results:
paths.data: /usr/share/fish
paths.sysconf: /usr/etc/fish
paths.doc: /usr/share/doc/fish
paths.bin: /usr/bin
```

I suspect this is wrong: https://github.com/fish-shell/fish-shell/blob/48a65506883a76aeea9f8268d039fb630ba82d8d/src/bin/fish.rs#L178

The parent of the parent of `/usr/bin/fish` is `/usr` and thus the sysconf path becomes `/usr/etc/fish`. On normal Fedora it is overwritten by the compiled-in path, because `/usr/etc/fish` does not exist, but on Fedora Silverblue `/usr/etc/fish` actually exists (because of the OS) so it is not overwritten.

faho:
I have to confess I have never needed the relocatable stuff, and I'm not sure if it is used in practice.

I believe @ridiculousfish put in a bunch of work there, maybe he can explain.

Otherwise my preference would be to remove it, because I don't see how else this would work.

moritz-meier:
This could be a quick fix: https://github.com/fish-shell/fish-shell/commit/9900d11b81fe08290ba7119d767644ec98de0673

With this it should derive the right sysconf_dir also when running from `/usr/bin/` and then it also shouldn't have to fallback to compiled-in paths.

faho:
Alternatively, instead of specifically coding for yet more specific paths:

[...]

PR #10923, opened 2024-12-19
Allow installable builds to be installed into a specific path

faho:
## Description

This reuses the "relocatable tree" idea to allow installable builds to be put into one.

So you do

```fish
fish --install=$HOME/.local
```

and it

- installs the data files into ~/.local/share/fish/install (that "install" part is still important so we don't clear out the data dir and blow away your history, ask me how I know)
- copies the fish binary to ~/.local/bin/fish

And if you then run ~/.local/bin/fish, it just finds the files.

The main weirdness here is that this picks up the relocatable idea of using PREFIX/etc as sysconfdir. I do not understand this, and it is what causes #10748. Personally, I would remove this, but it would mean fish would detect a relocatable tree as soon as ../share/fish exists (and we *also* allow fish next to share/).

When fish can't get the home directory (because we do not want to trust $HOME because that breaks with `sudo`), it will tell you to install it somewhere specific, so it

Fixes issue #10916

## TODOs:

- [N/A] Tests have been added for regressions fixed - I do not know how to test installable builds yet. Ideas welcome!

### DIFF
--- a/doc_src/cmds/fish.rst
+++ b/doc_src/cmds/fish.rst
@@ -40,9 +40,11 @@ The following options are available:
 **-i** or **--interactive**
     The shell is interactive.
 
-**--install**
+**--install[=PATH]**
     When built as self-installable (via cargo), this will unpack fish's datafiles and place them in ~/.local/share/fish/install/.
     Fish will also ask to do this automatically when run interactively.
+    If PATH is given, fish will install itself into a relocatable directory tree rooted at that path.
+    That means it will install the datafiles to PATH/share/fish and copy itself to PATH/bin/fish.
 
 **-l** or **--login**
     Act as if invoked as a login shell.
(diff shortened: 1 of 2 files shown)


ridiculousfish (approved):
Awesome! Might be worth adding a CI test job for this even.

faho:
> Might be worth adding a CI test job for this even.

We have the "staticbuilds" job, that's manually triggered for now. But that only confirms it *compiles*.

If we want to run our tests on installable builds, we would have to restructure how our test harness is run a bit. Currently it's basically impossible to run outside of cmake, and this would have to be called something like `test_driver.sh target/debug/fish`. Plus fish_test_helper would have to be built manually (which shouldn't be too hard) or tests with it skipped.

I've been looking into doing some of that.

-----

@ridiculousfish Do you have any comment on ../etc? As far as I know you've been entirely spearheading this idea of having fish "relocatable", and I'm not sure what you use it for.

Is ../etc the right directory for sysconfdir? It is not in #10748, and I don't believe it is for installable builds either.

Is there a use case I am not seeing?

I am going to set sysconfdir to the compiled-in path ("/etc", not "../etc") for installable builds, but if I'm doing that here I have to wonder if it isn't the right choice always.

faho:
Okay, I merged it with the "../etc". Let's see how that works out.

Issue #2855, opened 2016-03-24
how to bind the "Ctrl-l" to empty the screen instead of clearing the terminal buffer?


c02y:
I use fish with tmux, fish provides a default binding for "C-l":

```
bind \cl 'clear; commandline -f repaint'
```

but it will clear the terminal buffer (leaving only one whole screen of content if I scroll it)  and I won't be able to scroll the terminal to view the content or the result  of the history commands.

I can make it work by defining a key binding in tmux configuration file, but it seems there is a bug with it because of the conflict of two bindings in fish and tmux. And I have to call a function defined in ~/.config/fish/functions when hitting Ctrl-l (clear and then call the function), so it is better to make a binding in config.fish to replace the default one.

So how can I do that?

krader1961:
The way to create your own key bindings, including overriding a default binding, is to create a function named `fish_user_key_bindings` and put your own bindings in that function. For example, you might put something like the following in your *config.fish*:

```
if status --is-interactive
    source $HOME/.config/fish/interactive.fish
    function fish_user_key_bindings
       bind \n backward-delete-char
       bind -M insert \n backward-delete-char
       bind -m insert \n backward-delete-char
    end
end
```

c02y:
I know how to bind a key in fish,

Please read the description.

krader1961:
> Please read the description.

I did and couldn't make sense of what you're asking. But you ended with

> so it is better to make a binding in config.fish to replace the default one. So how I do that?

So I took a guess that you needed a pointer on how to create a binding to replace the default one for [ctrl-L].

faho:
Of note here is that `clear` is an external command that as far as I can tell just does the same thing as `echo \e\[H\e\2J` - an escape sequence that the terminal recognizes and then proceeds to do its thing.

So the question here is if the mechanism is just deficient for you in tmux (which acts as a terminal), or in your actual terminal (to which tmux talks).

What happens for me (and presumably @krader1961, which is why he does not understand what you're asking), is that the terminal adds a full screen's worth of whitespace, pushing the old content out. I can still scroll back, though. This is with KDE's konsole (version 15.12) and no multiplexer (screen or tmux).

faho:
Works about the same in tmux for me, except that tmux will buffer the output instead of the actual terminal, so I'd need to use the tmux bindings to scroll up or down ("Ctrl-b [", then arrow keys to move around, then "q" to stop scrolling).

c02y:
[...]

PR #10934, opened 2024-12-23
ctrl-l to scroll content instead of erasing screen

### LINKED ISSUE #1448: Cursor becomes glitchy in wide iTerm2 windows
I'm honestly not sure whether to open this here or in iTerm, but...

When I make my Terminal window wide enough (170 columns is the magic number for me) and start typing commands, the cursor glitches out and starts flickering on random columns in the prompt line.

The reason I'm not sure where to open this issue is that this graphical glitch happens in iTerm2 but not Terminal.app. However, it only happens with fish-shell. It does not happen in other shells I have tested (bash, zsh, etc.)

krobelus:
On ctrl-l we send `\e[2J` (Erase in Display).  Some terminals interpret
this to scroll the screen content instead of clearing it. This happens
on VTE-based terminals like gnome-terminal for example.

The traditional behavior of ctrl-l erasing the screen (but not the
rest of the scrollback) is weird because:

1. `ctrl-l` is the easiest and most portable way to push the prompt
   to the top (and repaint after glitches I guess). But it's also a
   destructive action, truncating scrollback. I use it for scrolling
   and am frequently surprised when my scroll back is missing
   information.
2. the amount of lines erased depends on the window size.
   It would be more intuitive to erase by prompts, or erase the text
   in the terminal selection.

Let's use scrolling behavior on all terminals.

The new command could also be named "push-to-scrollback", for
consistency with others. But if we anticipate a want to add other
scrollback-related commands, "scrollback-push" is better.

This currently uses "CSI n A" for cursor movement, I guess to be safe
we could query for support for that as it's done in fa68c2619f (Use
parm_left_cursor and parm_right_cursor for bulk cursor motions. Fixes
#1448, 2014-05-09).

The implementation is currently async: keys sent immedately after
`ctrl-l` may be processed before the repaint completes.  We can
fix that.  For now change tests to reflect the new async behavior.

Ref: https://codeberg.org/dnkl/foot/wiki#how-do-i-make-ctrl-l-scroll-the-content-instead-of-erasing-it
as of wiki commit b57489e298f95d037fdf34da00ea60a5e8eafd6d

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -3,9 +3,7 @@ use crate::curses;
 use crate::env::{Environment, CURSES_INITIALIZED};
 use crate::event;
 use crate::flog::FLOG;
-use crate::input_common::{
-    CharEvent, CharInputStyle, InputData, InputEventQueuer, ReadlineCmd, R_END_INPUT_FUNCTIONS,
-};
+use crate::input_common::{CharEvent, CharInputStyle, InputData, InputEventQueuer, ReadlineCmd};
 use crate::key::ViewportPosition;
 use crate::key::{
     self, canonicalize_raw_escapes, ctrl, Key, Modifiers, MouseButton, MouseEvent, MouseEventType,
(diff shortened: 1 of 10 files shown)


krobelus:
I'm pondering failure modes for when the terminal doesn't respond to CSI 6 n. It's probably
safe to assume this will never happen, so I guess fish should not process other input until
the response has arrived.  That would get rid of the weird out-of-order execution here. Could
still allow ctrl-c as escape hatch (canceling the repaint command in this case).

src/input.rs
```diff
@@ -218,27 +218,17 @@ const INPUT_FUNCTION_METADATA: &[InputFunctionMetadata] = &[
...
-
-// Keep this function for debug purposes
-// See 031b265
-#[allow(dead_code)]
-pub fn describe_char(c: i32) -> WString {
-    if c > 0 && (c as usize) < R_END_INPUT_FUNCTIONS {
-        return sprintf!("%02x (%ls)", c, INPUT_FUNCTION_METADATA[c as usize].name);
-    }
-    return sprintf!("%02x", c);
-}
```
zanchey, line 241: Were these functions supposed to be removed?

zanchey:
#2855 has some background here

src/input.rs
krobelus: I didn't end up removing them

PR #10937, opened 2024-12-24
refactor: use format_args!() when String creation is unnecessary

Integral-Tech:
## Description

- When `String` creation is unnecessary, replace `format!()` with `format_args!()` to avoid overhead of heap allocation.

## TODOs:

### DIFF
--- a/src/timer.rs
+++ b/src/timer.rs
@@ -108,10 +108,12 @@ impl TimerSnapshot {
 
         let mut output = String::new();
         if !verbose {
-            output += "\n_______________________________";
-            output += &format!("\nExecuted in  {:6.2} {}", wall_time, wall_unit.long_name());
-            output += &format!("\n   usr time  {:6.2} {}", usr_time, cpu_unit.long_name());
-            output += &format!("\n   sys time  {:6.2} {}", sys_time, cpu_unit.long_name());
+            output += &format!(
+                "\n_______________________________\n{}\n   {}\n   {}",
+                format_args!("Executed in  {:6.2} {}", wall_time, wall_unit.long_name()),
+                format_args!("usr time  {:6.2} {}", usr_time, cpu_unit.long_name()),
+                format_args!("sys time  {:6.2} {}", sys_time, cpu_unit.long_name()),
+            );
         } else {
             let fish_unit = Unit::for_micros(fish_sys.max(fish_usr).as_micros() as i64);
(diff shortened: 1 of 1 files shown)


Integral-Tech:
Any progress on this pull request?

src/timer.rs
```diff
@@ -129,15 +131,17 @@ impl TimerSnapshot {
...
-                width2 = fish_unit.len() + 7
+                "\n________________________________________________________\n{}\n   {}\n   {}",
+                format_args!(
+                    "Executed in  {wall_time:6.2} {wall_unit:<width1$}    {fish:<width2$}  external",
+                    width1 = column2_unit_len,
+                    fish = "fish",
+                    width2 = fish_unit.len() + 7
+                ),
+                format_args!("usr time  {usr_time:6.2} {cpu_unit:<column2_unit_len$}  {fish_usr_time:6.2} {fish_unit}  {child_usr_time:6.2} {child_unit}"),
+                format_args!("sys time  {sys_time:6.2} {cpu_unit:<column2_unit_len$}  {fish_sys_time:6.2} {fish_unit}  {child_sys_time:6.2} {child_unit}")
```
krobelus, line 143: thanks for pointing out `format_args`.
I think as-is this PR makes things worse overall; readability seems worse.
The (arguably subordinate) performance optimization is neither-here-nor-there,
because there is still a needless `format!` here, which is confusing.
If this is worth doing, we should try something like `write!` which needs no allocation.

Maybe we should first unify relevant places that write to stdout/stderr,
to enable us to use the same writing mechanisms everywhere (which might also help testing).
Then we can use our (existing) buffered writing mechanism possibly with something like `write!`.
Our buffered writer lives in `Outputter` but we don't use it everywhere, for better or worse.

Making a small change to use `Outputter` here might be an improvement, I haven't tried that.

Integral-Tech: @krobelus Okay, I see

juntuu: This could use [std::fmt::Write](https://doc.rust-lang.org/std/fmt/trait.Write.html#impl-Write-for-String), which is implemented for `String`.

For example:
```rust
use std::fmt::Write as _;
_ = writeln!(&mut output, "Executed in  {:6.2} {}", wall_time, wall_unit.long_name());
```

It should be relatively simple to later switch to another output, as long as it has `Write` implementation.

faho:
As it stands, we don't think this improves readability and in this case the performance simply doesn't matter.

Closing, thanks anyway!

PR #10942, opened 2024-12-25
docs: Use grid in the CSS

faho:
Instead of hardcoded 230px margin.

This also makes the ToC only take up a third of the screen when narrow, and lets you scroll the rest.

Without, you'd have to scroll past the *entire* ToC, which is awkward.

Remaining issue is the search box up top. Since this disables the one in the sidebar again once the window gets too narrow, that one is important, and it isn't *great*.

(we had disabled the search box before but sphinx changed it from a `<div>` to a `<search>`)

## Screenshots

These are from Firefox, with the "Galaxy S20 Android 11" view (800 x 360). Chromium looks similar, and moving it around feels okay - it's nice and responsive, switching to the "narrow" view with the sidebar up top at the correct point.

Old:
Bildschirmfoto am 2024-12-25 um 11 06 42
Bildschirmfoto am 2024-12-25 um 11 06 46

Note the "quick search" in the middle and the fact that the *entire screen* is table of contents, no actual content.

New:

Bildschirmfoto am 2024-12-25 um 11 06 58
Bildschirmfoto am 2024-12-25 um 11 07 00

## TODOs:

### DIFF
--- a/doc_src/python_docs_theme/static/pydoctheme.css
+++ b/doc_src/python_docs_theme/static/pydoctheme.css
@@ -71,12 +71,10 @@ div.related h3 {
 }
 
 div.related li.right {
-  float: right;
   margin-right: 5px;
 }
 
 div.sphinxsidebar {
-  width: 230px;
   overflow-wrap: break-word;
 }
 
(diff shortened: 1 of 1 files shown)


faho:
Note: I'll be adding this to 4.0, and I'm also immediately uploading it to the 4.0b1 docs online because the current mobile view isn't *great*.

PR #10946, opened 2024-12-25
Fix cursor position restore

phanen:
## Description
vim/nvim cannot restore cursor position now, the reason:
1. multi-line string cannot be handled correctly to create a valid autocmd.
2. I think the column number should +1 offset.


krobelus:
> multi-line string cannot be handled correctly to create a valid autocmd.

I see, this is a Vim 8 thing.

> I think the column number should +1 offset.

I don't think so, I think `alt-e :q` should not move the cursor

phanen:
> > I think the column number should +1 offset.
> 
> I don't think so, I think `alt-e :q` should not move the cursor

This problem seems happened with kitty terminal. Reproduce by:
```fish
kitty --config NONE -o shell=fish
```

Maybe its shell-integration do some thing wrong...

PR #10948, opened 2024-12-26
Convert builtin functions return type from Option<c_int> to Result<StatusOk, StatusError>.

sandercm:
## Description

Talk about your changes here.

Currently all error codes are wrapped in an Option type, this is not a causes the Option type to leak into a lot of function return types when it's not needed. 

In this PR I removed all these Options around the static variables aswell as reworked all `Option<c_int>` return types to be `Result<StatusOk, StatusError>`. `StatusOk` is an enum that can be a regular `OK` or an `OK_PRESERVE_ERROR_CODE`. There's a lot of small improvements that are now possible for example you can now use `?` to handle errors a lot easier.

There's still a lot more places that could be improved but this PR is limited to replacing all the builtin functions.

## TODOs:

### DIFF
--- a/src/builtins/bg.rs
+++ b/src/builtins/bg.rs
@@ -10,7 +10,7 @@ fn send_to_bg(
     streams: &mut IoStreams,
     cmd: &wstr,
     job_pos: usize,
-) -> Option<c_int> {
+) -> Result<StatusOk, StatusError> {
     {
         let jobs = parser.jobs();
         if !jobs[job_pos].wants_job_control() {
@@ -24,7 +24,7 @@ fn send_to_bg(
             )
             };
             builtin_print_help_error(parser, streams, cmd, &err);
-            return STATUS_CMD_ERROR;
+            return Err(StatusError::STATUS_CMD_ERROR);
         }
 
         let job = &jobs[job_pos];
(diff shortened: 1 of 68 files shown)


src/builtins/bg.rs
```diff
@@ -46,16 +41,20 @@ fn send_to_bg(
...
-        Err(err @ Some(_)) if err != STATUS_CMD_OK => return err,
+        Err(err) if err != STATUS_CMD_OK => return err,
         Err(err) => panic!("Illogical exit code from parse_options(): {err:?}"),
     };
 
-    let cmd = args[0];
+    let cmd = match args.get(0) {
+        Some(cmd) => cmd,
+        None => return STATUS_INVALID_ARGS,
+    };
```
epidemian, line 54: Very fly-by comment: this can be written more succinctly as:

```suggestion
    let Some(cmd) = match args.get(0) else {
        return STATUS_INVALID_ARGS;
    };
```

src/builtins/set.rs
```diff
@@ -1012,7 +1012,7 @@ pub fn set(parser: &Parser, streams: &mut IoStreams, args: &mut [&wstr]) -> Opti
     };
 
     if retval == STATUS_CMD_OK && opts.preserve_failure_exit_status {
-        return None;
+        return STATUS_PRESERVE_FAILURE;
```
ridiculousfish, line 1015: This is the crux - the reason that there's so many Options everywhere is because `builtin set` should not modify the exit status if it succeeds. We should not reserve a value like `STATUS_PRESERVE_FAILURE` because then that becomes unavailable as the exit status of other commands.

ridiculousfish (changes requested):
Thank you for tackling this. I agree the time is right to improve how exit statuses are expressed. But we can't do it by reserving new values like `STATUS_PRESERVE_FAILURE` since we will then misinterpret that exit status from other commands.

I think we should stop making `c_int` the primary representation - that's a holdover from the C++ port. I think it's time to make a real type which is convenient to use, can express any integer exit status, and also encapsulates the idea of "preserve status" for builtin_set.

If you're willing to take that further step then let's leave this open.

sandercm:
> Thank you for tackling this. I agree the time is right to improve how exit statuses are expressed. But we can't do it by reserving new values like `STATUS_PRESERVE_FAILURE` since we will then misinterpret that exit status from other commands.
> 
> I think we should stop making `c_int` the primary representation - that's a holdover from the C++ port. I think it's time to make a real type which is convenient to use, can express any integer exit status, and also encapsulates the idea of "preserve status" for builtin_set.
> 
> If you're willing to take that further step then let's leave this open.

Okay, I was initially planning on doing that but wanted confirmation that the idea would have support 😅ill give it a shot and see how far I get

src/builtins/set.rs
sandercm: I have refactored it now to return a Result<StatusOk, StatusError> where StatusOk represents either succes or succes but keep error value. StatusError is the error obviously.

PR #10953, opened 2024-12-27
Fix buff pos for CursorEndMode::Inclusive

### LINKED ISSUE #10286: Vi-mode poor autosuggestion behavior / Invalid cursor placement and length in normal mode

### LINKED ISSUE #10952: fish 4.0b1: vi-mode: gU doesn't capitalize the whole word
```console
> fish --version
4.0b1-4-g687001f8b
> uname -a
Linux machine 6.8.0-49-generic #49-Ubuntu SMP PREEMPT_DYNAMIC Mon Nov  4 02:06:24 UTC 2024 x86_64 GNU/Linux
> echo $TERM
xterm
```

> executing this command and whether it affected the behavior you are reporting:
    sh -c 'env HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_DIRS= fish'

Yes, it still does.

**How to reproduce the problem**

Install the required fish version.
I download and install the [self-installing static artifact][1].

```
# Change to vi binding mode:
fish_vi_key_bindings
# type something
asdf[^ESC]
# press ESC to switch to normal mode
# go to the first column with `0`
# let's try to capitalize the whole word with gU
# and the result is
ASDf
# f is still a lowercase char
```

fish 3.7.1 behaves as expected, i.e. whole word is capitalized as ASDF.

cc #10633

[1]: https://github.com/fish-shell/fish-shell/releases/download/4.0b1/fish-static-linux-x86_64.tar.xz

tesuji:
## Description

I'm not familiar with the test infrastructure. And PR that fixed #10286 doesn't
include a test. Please guide me to write a test if needed.

Seems to work OK when manually test my case and #10286.

Fixes issue #10952.

## TODOs:

### DIFF
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -1314,12 +1314,8 @@ impl ReaderData {
     fn update_buff_pos(&mut self, elt: EditableLineTag, mut new_pos: Option<usize>) -> bool {
         if self.cursor_end_mode == CursorEndMode::Inclusive {
             let el = self.edit_line(elt);
-            let mut pos = new_pos.unwrap_or(el.position());
-            if !el.is_empty() && pos == el.len() {
-                pos = el.len() - 1;
-                if el.position() == pos {
-                    return false;
-                }
+            if new_pos.is_none() && !el.is_empty() && el.position() == el.len() {
+                let pos = el.len() - 1;
                 new_pos = Some(pos);
             }
         }


src/reader.rs
```diff
@@ -1314,12 +1314,8 @@ impl ReaderData {
...
         if self.cursor_end_mode == CursorEndMode::Inclusive {
             let el = self.edit_line(elt);
-            let mut pos = new_pos.unwrap_or(el.position());
-            if !el.is_empty() && pos == el.len() {
-                pos = el.len() - 1;
-                if el.position() == pos {
-                    return false;
-                }
+            if new_pos.is_none() && !el.is_empty() && el.position() == el.len() {
+                let pos = el.len() - 1;
```
krobelus, line 1318: I'm queueing something like

```patch
Subject: [PATCH 1/2] Add missing test for Vi mode $

PR #10953 reports missing coverage for the change to update_buff_pos()
in d51f669647 (Vi mode: avoid placing cursor beyond last character,
2024-02-14).

Add a case demonstrating how $ should not move the cursor past the
last character. Goes without saying that it's really ugly that we
update_buff_pos() must be so defensive here, ideally we wouldn't pass
it out-of-bounds positions.
---
 tests/pexpects/bind.py | 6 ++++++
 1 file changed, 6 insertions(+)

diff --git a/tests/pexpects/bind.py b/tests/pexpects/bind.py
index c42e0e7e03..d1588464f4 100644
--- a/tests/pexpects/bind.py
+++ b/tests/pexpects/bind.py
@@ -241,6 +241,12 @@ expect_prompt("foo")
 # send("hh~~bbve~\r")
 # expect_prompt(TO_END + "SOME TeXT\r\n", unmatched="Couldn't find expected output 'SOME TeXT")
 
+send("echo 125")
+send("\033")
+sleep(0.200)
+send("0$i34\r")
+expect_prompt("echo 12345")
+
 # Now test that exactly the expected bind modes are defined
 sendline("bind --list-modes")
 expect_prompt(
-- 
2.47.1


Subject: [PATCH 2/2] Fix off-by-one error in Vi-style upcase-word at
 commandline end

cursor_selection_mode=inclusive means the commandline position is
bounded by the last character. Fix a loop that fails to account
for this.

Fixes d51f669647 (Vi mode: avoid placing cursor beyond last character,
2024-02-14).

Fixes #10952
Closes #10953

Reported-by: Lzu Tao <taolzu@gmail.com>
---
 src/reader.rs          | 11 ++++++++++-
 tests/pexpects/bind.py |  6 ++++++
 2 files changed, 16 insertions(+), 1 deletion(-)

index 59f8b83650..63523fd1b8 100644
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -3204,7 +3204,16 @@ impl<'a> Reader<'a> {
                 );
                 let (elt, el) = self.active_edit_line();
                 let mut replacement = WString::new();
-                while pos < el.position() {
+                while pos
+                    < if self.cursor_selection_mode == CursorSelectionMode::Inclusive
+                        && self.is_at_end(el)
+                        && !el.is_empty()
+                    {
+                        el.len()
+                    } else {
+                        el.position()
+                    }
+                {
                     let chr = el.text().as_char_slice()[pos];
 
                     // We always change the case; this decides whether we go uppercase (true) or
diff --git a/tests/pexpects/bind.py b/tests/pexpects/bind.py
index d1588464f4..991325e33e 100644
--- a/tests/pexpects/bind.py
+++ b/tests/pexpects/bind.py
@@ -241,6 +241,12 @@ expect_prompt("foo")
 # send("hh~~bbve~\r")
 # expect_prompt(TO_END + "SOME TeXT\r\n", unmatched="Couldn't find expected output 'SOME TeXT")
 
+send("echo echo")
+send("\033")
+sleep(0.200)
+send("bgU\r")
+expect_prompt("echo ECHO")
+
 send("echo 125")
 send("\033")
 sleep(0.200)
-- 
2.47.1
```

PR #10955, opened 2024-12-27
Make tests runnable without CMake

faho:
## Description

This moves some of the setup our test infrastructure does from cmake, and into the test_driver.sh script, in order to allow

```fish
for test in tests/test.fish tests/interactive.fish
    FISHDIR=$PWD/target/debug/ tests/test_driver.sh $test
end
```

to run the test suite, to successful completion.

To do this, it:

- Moves the python test drivers into tests/, from build_tools. Tbh that's cleaner anyway
- Moves building fish_test_helper into test_driver.sh (it's just `cc fish_test_helper.c`) - this takes about 50ms, *per test_driver invocation* (i.e. once for the scripted tests, once for the interactive ones)
- Lets test_driver.sh call fish via $FISHDIR, not $PATH - so you can pass it explicitly
- Makes the tests run in the actual temporary $HOME again and adapts them for it (the alternative is to change all files the tests touch to be in $HOME instead of $PWD)
- Assorted

$FISHDIR is a bit ugly - it can't be an argument, because we need `fish`, `fish_indent` and `fish_key_reader`, and passing three variables struck me as awkward. This would be fixed by #10876.

We could also remove the checks for fish_test_helper, declare that we need a C compiler (we do for compiling fish anyway!) and remove them again.

## TODOs:

- [Haha] Tests have been added for regressions fixed


faho:
Note: This not only reduces our dependency on cmake, but is also needed to be able to test self-installable fish - because that doesn't use cmake.

ridiculousfish (commented):
Change looks directionally good, moving us away from CMake. Two thoughts:

1. Can we have test_driver.sh `cd`  to the right directory?
2. Would it be useful if fish_test_helper were written in Rust? I think it should be very straightforward to do and then you won't need to conditionalize the code on its presence.

tests/checks/abbr.fish
```diff
@@ -141,7 +141,7 @@ echo $status
 abbr --query banana --function
 echo $status
 # CHECKERR: abbr: --function: option requires an argument
-# CHECKERR: checks/abbr.fish (line 141):
+# CHECKERR: {{.*}}checks/abbr.fish (line 141):
```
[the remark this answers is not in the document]
faho: We have two directories here:

- The "test root", which is in test/ in the build directory when ran via cmake, and just wherever you run the tests from otherwise
- The "homedir", which is a temporary directory (/tmp/randomstring)

What happened before was that test_driver would cd to the homedir, but then the test runner (test.fish or interactive.fish) would `cd (status dirname)` back to the test root.

I removed that last part because that just undid the former.

Now, ending up in the test root means that we get the nice "checks/foo.fish", but it also means that the running test is in a non-temporary directory. Which isn't a huge deal when running with cmake because the test root will also be temporary (it copies the tests there), but outside of cmake it's just your $PWD.

So the test would have to know to `cd` to $HOME if it needs to create files instead. I went with the solution that leaves the $PWD clean by default.

faho:
> Would it be useful if fish_test_helper were written in Rust?

Not really. We require a c compiler to compile anyway - libc.c, the rsconf checks.

I also don't know how we would add a separate binary that should not be installed but should be available to test scripts. Which [target type](https://doc.rust-lang.org/cargo/reference/cargo-targets.html) should it be? It's not a binary, a library or really an example. It's also not a "test" binary because the semantics for that are that `cargo test` runs its `#[test]` functions.

Tbh we could probably remove the `#REQUIRES` lines - that would only stop you from running the tests on e.g. a packaged fish binary that you got from elsewhere if you don't have the means of building fish.

tests/test.fish
```diff
@@ -38,11 +46,11 @@ if set -q files_to_test[1]
...
 
-    $python -S ../littlecheck.py \
+    $python -S $scriptdir/littlecheck.py \
         --progress $force_color \
-        -s fish=../test/root/bin/fish \
-        -s fish_test_helper=../test/root/bin/fish_test_helper \
-        -s filter-control-sequences='../test/root/bin/fish ../tests/filter-control-sequences.fish' \
+        -s fish=$FISHDIR/fish \
+        -s fish_test_helper=$fish_test_helper \
+        -s filter-control-sequences="$FISHDIR/fish $scriptdir/filter-control-sequences.fish" \
```
faho, line 53: This will fail if $FISHDIR or $scriptdir include spaces.

I don't know how to solve it yet.

ridiculousfish (approved):
Ok LGMT - let's get the bind.py test resolved and also please add a note to the "Local testing" section of CONTRIBUTING.md describing how to run the tests, if you want to promote this CMake-free process.

tests/pexpects/bind.py
```diff
@@ -358,19 +358,20 @@
...
-        commandline --insert b
-        commandline -f backward-char
-        commandline -f backward-char
-        commandline -f delete-char
-    "
-""".strip())
-send('\x07') # ctrl-g
-send('\r')
-expect_prompt("foobar")
+# TODO: This test breaks constantly on my system when run directly?
```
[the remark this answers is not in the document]
faho: I can no longer reproduce this.

tests/checks/abbr.fish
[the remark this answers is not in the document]
faho: Okay, this sent me on a bit of a hike.

The multiple layers of posix shell here make this an absolute pain in the gluteals

(seriously, *how* do you source a script *while* keeping your arguments intact without that script getting them?)

but I did manage.

tests/test_driver.sh
```diff
@@ -1,3 +1,4 @@
+#!/bin/sh
```
krobelus, line 1: > Note it will succeed under bash, but fail in alpine's sh

This used to work, no?

CONTRIBUTING.rst
```diff
@@ -209,6 +214,15 @@ The tests can be run on your local computer on all operating systems.
    cmake path/to/fish-shell
    make test
 
+Or you can run them on a fish, without involving cmake::
+
+  cargo build
+  FISHDIR=target/debug tests/test_driver.sh tests/test.fish # script tests, the checks
```
krobelus, line 220: Maybe `FISHDIR=target/debug` should be the default when not invoked by cmake.
Or perhaps `./build`. Currently it's `$PWD`, I'm not sure if that's ever useful.

CONTRIBUTING.rst
```diff
@@ -209,6 +214,15 @@
...
+  FISHDIR=target/debug tests/test_driver.sh tests/interactive.fish # interactive tests, the pexpects
```
krobelus, line 221: in future we want to re-add some ctest niceties, such as printing a list of all failed test at the end

krobelus, line 221: We should maybe have a script that's more user-friendly
Not requiring specifying `tests/test.fish` and `tests/interactive.fish`
is one idea, and allow passing full tests paths.

Going further, we could make each test standalone script, then it'll be
extremely obvious how to run an individual test.  Git's test suite is
structured that way. I think that's a good fit.

faho: First thing I want to do is pass the path to a fish binary, but that requires #10876.

But tbh I'm not actually sure the "test_driver" helps in any real way, and its posixness hurts quite a bit. So, since we're using the fish being tested to run test.fish/interactive.fish anyway, why not just make test.fish/interactive.fish the entry points?

I don't really want to make each test a standalone script because that means more boilerplate for each script.

krobelus: > I don't really want to make each test a standalone script because that means more boilerplate for each script.

It should be only one line for most cases (something like `source ./test/lib.sh`, like what Git uses), perhaps even inside the littlecheck `# RUN` line... No need to rush into that of course.

> But tbh I'm not actually sure the "test_driver" helps in any real way, and its posixness hurts quite a bit. So, since we're using the fish being tested to run test.fish/interactive.fish anyway, why not just make test.fish/interactive.fish the entry points?

There is a prominent comment `# POSIX sh test driver to reduce dependency on fish in tests.`,
which makes sense because if `build/fish` is so broken that it fails to run the test framework properly, then we might get unintelligible error messages on our test failures.
But I guess we've been running `test.fish` with `build/fish` forever and the only time I remember hitting weird problems was when porting execution components.
The test driver script itself is a bit more delicate than `test.fish`; if the driver's `rm -rf` misfires, that would be bad. Not sure if that's a realistic problem, but maybe we should not using a moving fish for that (but pretty much everything else)

The tmux tests currently add another level of nesting, which has caused confusion in the past.
Those tests have some other problems that cause unintelligible failures. Being able to run the test directly would make it much easier to temporarily add a `tmux attach` for debugging.

Reducing the amount of nested shells and different languages sounds good overall.

faho: > But I guess we've been running test.fish with build/fish forever and the only time I remember hitting weird problems was when porting execution components.

Yeah, that's kinda the thing - here's the nesting:

- cmake runs
- test_driver.sh, which runs
- test.fish (via the tested fish!), which runs
- littlecheck (python), which runs
- the to-be-tested fish script in the tested fish, and
- filter-control-sequences.fish, in the tested fish
- (potentially) tmux

Basically, we're already running the to-be-tested fish at least two, possibly three times.

filter-control-sequences used to be an sh script, which was overall an acceptable idea - but posix sh wasn't a good language for this use and it ended up calling out to the to-be-tested fish anyway.

I would see value in reducing this nesting and the number of languages here, and I see three possibilities:

- rewrite test_driver.sh in fish, leaving python and fish
- rewrite test_driver.sh, test.fish, filter-control-sequences.fish in python, leaving python (and the actual test scripts in fish)
- rewrite test_driver.sh, test.fish, filter-control-sequences.fish in rust, leaving rust (and the test scripts)

The first option could either go with the to-be-tested fish or we just require an existing fish install to run the tests. This wouldn't be as bad as it sounds because you can just build fish and then keep it to run future tests, but there *is* a bit of bootstrappiness we'd lose.
(and I maintain that one major downside of using fish for scripts is the need for `--no-config`)

I am unsure how exactly this would work in rust - how would it be hooked up? - and I don't love the idea of rewriting littlecheck, which relies on the python standard library a bunch.

So I think my favorite so far is actually python. I'll see if I can whip up a proof of concept.

faho: Oh, right, if we wanted to eliminate python from the tests we would also have to find a replacement for pexpect, which I wouldn't be looking forward to.

CONTRIBUTING.rst
faho: No, the default is `${BUILD_ROOT}/test/root/bin`, i.e. what cmake would use.

The issue with making it target/debug is that you might be testing a release build instead, and so it would mislead you into thinking the tests pass.

I'm planning on making the path to fish the first argument with #10876.

faho: Okay, I made it error out if $FISHDIR isn't given, to avoid any confusion.

PR #10958, opened 2024-12-28
docs: Distinguish documents in sidebar

kidonng:
## Description

#10942 is a great change, but also makes it hard to tell one document from another in the sidebar. Previously, each document are on their own line, but now a document with long title can span multiple lines.

This pull request restores the previously hidden list item marker to make it easier to distinguish the documents.

| | Before | After |
|--------|--------|--------|
| Wide | CleanShot 2024-12-28 at 15 25 38@2x | CleanShot 2024-12-28 at 15 27 16@2x |
| Narrow | CleanShot 2024-12-28 at 15 27 47@2x |CleanShot 2024-12-28 at 15 27 01@2x | 

## TODOs:

### DIFF
--- a/doc_src/python_docs_theme/static/pydoctheme.css
+++ b/doc_src/python_docs_theme/static/pydoctheme.css
@@ -53,7 +53,6 @@ body {
 div.related ul {
   margin: 0;
   padding: 0 0 0 10px;
-  list-style: none;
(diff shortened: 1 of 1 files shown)


faho:
Yeah that's a good idea. Thanks!

PR #10961, opened 2024-12-28
Update German translation

be-we:


faho:
This is pretty hard to review because there are a lot of formatting changes mixed in with actual changes.

So far, the new or changed translations are good, but it's going to take a while to get through it. You did not fall into the obvious trap of translating functional strings like "generic" (as an argument to `functions --handlers-type`), which is good.

We really need better tooling here - github refuses to show the diff on this because it's too big, and there are a *lot* of unused "fuzzy" translations in these files that gunk up the diff.

faho:
Okay, there are about 300 useful lines in here, and ~7000 useless formatting changes.

There are two kinds of problems:

One is change in wrapping for the file lines:

```diff
-#: share/completions/btrfs.fish:256 share/completions/btrfs.fish:257                      
-#: share/completions/btrfs.fish:258                                                       
+#: share/completions/btrfs.fish:256 share/completions/btrfs.fish:257 share/completions/btrfs.fish:258
```

The other is a change in wrapping for msgid/msgstr:

```diff
-msgid ""                                                                                  
-"Trash all qgroup numbers and scan the metadata again with the current config."           
+msgid "Trash all qgroup numbers and scan the metadata again with the current config."     
```

Frankly, these tools are terrible, and we need something better.

I have merged the useful parts as 83586aebcc18b0b71b4b59ee5f1b7e53f89532ce, thanks!

Now we should look for a better translation system.

krobelus:
> Frankly, these tools are terrible, and we need something better.

could use a semantic diff tool. I didn't find a  treesitter parser,
which would enable tools like difftastic

PR #10966, opened 2024-12-29
README: use `cmake -B` to simplify build commands

louiswpf:
## Description
Note that `-B` was added in CMake 3.13 [1]
and CMake 3.15 or later is already required by fish. [2]

[1] https://cmake.org/cmake/help/latest/release/3.13.html#command-line
[2] commit 044cea1bf (update CMake requirement, 2024-12-26)

## TODOs:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -146,10 +146,9 @@ To install into ``/usr/local``, run:
 
 .. code:: bash
 
-   mkdir build; cd build
-   cmake ..
-   cmake --build .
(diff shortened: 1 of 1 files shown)


faho:
I don't believe this is helpful.

It's very specific to cmake - mkdir and cd are well-known, `cmake -B` isn't.

It's not like this needs to be as short as possible.

Sorry.

PR #10968, opened 2024-12-29
Improve documentation for `string escape`

nertzy:
Before, it unnecessarily stated that there are three `--style` options, when
there are actually four.

I also align the default `--style=script` argument to the beginning of the line
to match the other options visually for easier scanning.


ridiculousfish:
Thank you!

ridiculousfish:
Cherry-picked to Integration_4.0.0 as 1227b6765c9e02f14c69be858c481f7307470b69

PR #10971, opened 2024-12-29
completions/exercism: use generate script

kblcuk:
Exercism ships with it's own completions and a generation script, so let's use
that one instead.

Originally this PR was simply copying shell completions provided by Exercism repo,
but it was pointed out that we simply use completion generation script if that is
provided by the command itself.


faho:
Instead of committing an empty file, you can just delete it.

kblcuk:
@faho sorry about that, something wonky must've happened with my previous commit.
But, as also mentioned in the PR description, I'm fine with deleting the file altogether.

zanchey:
Probably this can be replaced with `exercism completion fish | source`?

kblcuk:
> Probably this can be replaced with `exercism completion fish | source`?

Hm, if that's what other competion scripts here doing as well (haven't checked) -- then yea, sounds reasonable! I'll update the PR once I'm back at the keyboard.

PR #10975, opened 2024-12-30
Code cleanup in `src/bin/fish.rs` to make it more idiomatic

needlesslygrim:
## Description
Updated some of the code in `src/bin/fish.rs` to make it more idiomatic, e.g. using if-expressions instead of:
```rust
let mut foo = None;
if (something) {
  foo = Some(something_else)
}
```

## TODOs:

### DIFF
--- a/src/bin/fish.rs
+++ b/src/bin/fish.rs
@@ -873,16 +873,13 @@ fn throwing_main() -> i32 {
     // Stomp the exit status of any initialization commands (issue #635).
     parser.set_last_statuses(Statuses::just(STATUS_CMD_OK.unwrap()));
 
-    match (&opts.profile_startup_output, &opts.profile_output) {
-        (Some(profile_startup_output), Some(profile_output))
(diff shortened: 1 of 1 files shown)


src/bin/fish.rs
```diff
@@ -871,13 +873,16 @@ fn throwing_main() -> i32 {
...
+        (Some(profile_startup_output), Some(profile_output))
+            if profile_startup_output != profile_output =>
+        {
+            parser.emit_profiling(profile_startup_output);
+            // If we are profiling both, ensure the startup data only
+            // ends up in the startup file.
+            parser.clear_profiling();
+        }
+        _ => (),
     }
```
mqudsi, line 886: I find the old code to be easier to immediately understand. I'd rather wait until if let chains are stabilized (and we can use them) rather than switch to the match block.

needlesslygrim: Sure, reverted that part of the patch

mqudsi (approved):

PR #10977, opened 2024-12-30
Add fish_cmd_entered event for modifying commandline after user hits enter

danzimm:
## Description

As I was hacking on some scripts I tried to use `fish_preexec` to alter the `commandline` prior to execution. I quickly realized that's not possible, and in fact it seems `fish_preexec` might not be the actual event I was looking for.

To enable such an operation (commandline editing prior to execution) I've added a new event that occurs earlier on, once the `Reader` determines that it's finished reading.

With this new event it's possible to define the following function (N.B. this may, and probably is, bad code):
```
function remap_foo --on-event fish_cmd_entered
  set cli (commandline --tokens-raw)
  set n (count $cli)
  test $n -eq 0; and return
  for i in (seq 1 $n)
    test "$cli[$i]" = foo; and set cli[$i] bar
  end
  commandline --replace "$cli"
end
```
and then
```
echo foo<enter>
```
would get replaced just before execution with
```
echo bar
```

---

This event is essentially enabling `abbr --regex '...' --position anywhere --function '...'`, but for the full commandline instead of just the matched token. I'm not sure if this is the correct way of going about this, or if modifying `abbr` is a better route.

Or, there's also the possibility that this sort of feature is out of scope for fish, which is ok too!

Fixes issue #

## TODOs:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,7 @@ Deprecations and removed features
 
 Scripting improvements
 ----------------------
+- `fish_cmd_entered` event emitted after the user hits enter
 
 Interactive improvements
 ------------------------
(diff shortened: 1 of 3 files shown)


danzimm:
Wow that was a fast response ! Sounds good, thank you for the direction!

PR #10985, opened 2024-12-31
environment_impl: drop usage of lazy_static

tranzystorekk:
## Description

This change gets rid of `lazy_static` as direct dependency for good.

In order to make `once_cell` happy, the manual `Sync` marker needs to be moved below
the `Arc<...>` level, so I added a newtype wrapper for the `RefCell<EnvNode>` itself.

## TODOs:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -118,7 +118,6 @@ dependencies = [
  "cc",
  "errno",
  "fish-printf",
- "lazy_static",
  "libc",
  "lru",
  "nix",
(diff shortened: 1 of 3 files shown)


mqudsi:
Instead of papering over the assertion I would prefer if the type statically asserted that the global lock is held.

ridiculousfish:
Thanks!

ridiculousfish:
> Instead of papering over the assertion I would prefer if the type statically asserted that the global lock is held.

Same but I'm not sure how to do it nicely. Maybe you can figure out a way.

PR #10988, opened 2025-01-01
Move `getrusage` wrapper in `timer.rs` to shared `nix` wrapper module

needlesslygrim:
## Description

There was a safe wrapper for `getrusage` in `timer.rs`, but it wasn't in the `nix` module. This change moves it there, and makes `bin/fish.rs` use it instead of `libc::getrusage`.

## TODOs:

### DIFF
--- a/src/nix.rs
+++ b/src/nix.rs
@@ -47,7 +47,8 @@ pub enum RUsage {
     RChildren,
 }
 
-/// A safe wrapper around `libc::getrusage()`
+/// A safe wrapper around `libc::getrusage()`. If `libc::getrusage()` fails,
+/// `wutil::perror()` will be called and a zeroed struct will be returned.
 pub fn getrusage(resource: RUsage) -> libc::rusage {
     let mut rusage = std::mem::MaybeUninit::uninit();
     let result = unsafe {
--- a/src/nix.rs
+++ b/src/nix.rs
@@ -2,6 +2,8 @@
 
 use std::time::Duration;
 
+use crate::wutil;
+
 #[allow(clippy::unnecessary_cast)]
 pub const fn timeval_to_duration(val: &libc::timeval) -> Duration {
     let micros = val.tv_sec as i64 * (1E6 as i64) + val.tv_usec as i64;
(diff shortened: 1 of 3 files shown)


src/nix.rs
```diff
@@ -37,3 +37,33 @@ pub fn isatty(fd: i32) -> bool {
...
+        libc::getrusage(
+            match resource {
+                RUsage::RSelf => libc::RUSAGE_SELF,
+                RUsage::RChildren => libc::RUSAGE_CHILDREN,
+            },
+            rusage.as_mut_ptr(),
+        )
+    };
+
+    // getrusage(2) says the syscall can only fail if the dest address is invalid (EFAULT) or if the
```
ridiculousfish, line 61: The old code invoked `perror` if `getrusage` failed - please restore that.

ridiculousfish (changes requested):
Please restore the error reporting and also `MaybeUninit` is now unused in fish.rs and can be removed. I'll be happy to merge. Thanks!

src/nix.rs
needlesslygrim: Isn't it impossible for `getrusage` to fail in this case? [The FreeBSD man page](https://man.freebsd.org/cgi/man.cgi?query=getrusage#end) says it fails if:
- 'The who [resource] argument is not a valid value'
  - This is impossible since we have the `RUsage` enum
- 'The address specified	by the rusage argument is not in a valid part of the process address space'
  - This shouldn't(?) be possible because we're using `MaybeUninit`

I think the best behaviour would probably be to panic if it doesn't return 0?

needlesslygrim: > also MaybeUninit is now unused in fish.rs and can be removed

Whoops, forgot :^)

krobelus: Both panic and perror would work in practice but if we can recover safely (from a libc bug or new error), that seems better than panicking. We have our own `perror` wrapper, no need to use `libc::perrror`.

needlesslygrim: Sure, would it be a good idea to also have a `try_getrusage` which returns `Result`?

krobelus: > Sure, would it be a good idea to also have a try_getrusage which returns Result?

there's no use for this at the moment because no caller cares about that

src/nix.rs
```diff
@@ -37,3 +39,37 @@ pub fn isatty(fd: i32) -> bool {
...
+    // requested resource type is invalid. Since we're in control of both, we can assume it won't
+    // fail. In case it does anyway (e.g. OS where the syscall isn't implemented), we can just
+    // return an empty value.
+    match result {
+        0 => unsafe { rusage.assume_init() },
+        _ => unsafe {
+            wutil::perror("getrusage");
```
krobelus, line 71: perror should probably not be inside unsafe

Issue #10987, opened 2025-01-01
Fish should flush the input queue before resuming control of the terminal.


OkamiW:
This can help eliminating a class of terminal emulator echoback vulnerabilities.

Demo: 
In terminal 1:
```
$ cat
test
```
Killing `cat` before it gets to read the input would cause the input be read by fish instead, which can lead to potential command execution.

This demonstration is for the sake of simplicity. 

In the real world situation, it could as well be `cat malicious-file.txt` feeding malicious control sequences to the terminal emulator, which would later echo back the malicious commands to the input queue.

`tcsetattr(3)` or `tcflush(3)` can be used to flush input queue to avoid this kind of issues.

Credit for the idea of flushing input queue:
Craig Barnes

See Also:
1. https://codeberg.org/dnkl/foot/issues/1867
2. https://dgl.cx/2023/09/ansi-terminal-security

krobelus:
This would also swallow characters the user typed, right (which is a frequently used feature I think)?
Would be nice if we could differentiate between those and any replies prompted by a child process.

dnkl:
This was split out to a separate issue on the foot tracker: https://codeberg.org/dnkl/foot/issues/1894

My stance: I'd _prefer_ the shell to handle this, as it should be able to handle the whole class of issues, whereas in foot (and terminals in general) all we can do is disable escapes (making them pretty much useless).

If you decide there's nothing you can do on the shell side, I'll just remove the escapes completely from foot. No point in having disabled-by-default escapes that pretty much no one knows when to enable or not.

OkamiW:
> This would also swallow characters the user typed, right (which is a frequently used feature I think)? Would be nice if we could differentiate between those and any replies prompted by a child process.

Yes.
IMO, since they're typed during the execution of last command, shell shouldn't execute them anyway.
Shell should only execute the command typed explicitly to execute, that is after the command prompt.

zanchey:
> In the real world situation, it could as well be `cat malicious-file.txt` feeding malicious control sequences to the terminal emulator, which would later echo back the malicious commands to the input queue.

Sorry, I don't understand this bit. The first example is the user typing "test" into what they think is the cat program, but which leaves the input in the queue after cat (say) crashes and then fish reads it from the input queue. The second example is `malicious-file.txt` containing control characters which cause the terminal emulator to write something into the input queue, and I guess crashing cat somehow as well? Or running a malicious program on the remote side of an SSH link which does something similar?

[...]

PR #10991, opened 2025-01-02
Don't execute off of buffered keys

krobelus:
If I run "sleep 3", type a command and hit enter, then there is no
obvious way to cancel or edit the imminent command other than ctrl-c
but that also cancels sleep, and doesn't allow editing. (ctrl-z sort
of works, but also doesn't allow editing).

Let's limit ourselves to inserting the buffered command (translating
enter to a newline), and only execute once the user actually presses
enter after the previous command is done.

This seems more user-friendly in general, at the risk of breaking
expectations in some cases.

This also fixes a class of security issues where a command like
`cat malicious-file.txt` might output escape sequences, causing the
terminal to echo back a malicious command; such files can still insert
into the command line but at least not execute it directly.
(Since it's only fixed partially I'm not really sure if the security
issue is a good enough motivation for this particular change.)

Note that bracketed paste probably has similar motivation.

Part of #10987
See also https://codeberg.org/dnkl/foot/issues/1894#issuecomment-2562154

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,9 @@ Scripting improvements
 Interactive improvements
 ------------------------
 - Autosuggestions are now also provided in  multi-line command lines. Like `ctrl-r`, autosuggestions operate only on the current line.
+- When typing a command while the previous one is still running, :kbd:`enter` will no longer execute the would-be command, and keys that are bound to shell commands will be ignored.
+  This mitigates a security issue where a command like ``cat malicious-file.txt`` could write terminal escape codes prompting the terminal to write arbitrary text to fish's standard input.
+  Such a malicious file can still potentially insert arbitrary text into the command line but can no longer execute it directly (:issue:`10987`).
 
 New or improved bindings
 ^^^^^^^^^^^^^^^^^^^^^^^^
--- a/src/input.rs
+++ b/src/input.rs
@@ -5,7 +5,7 @@ use crate::event;
 use crate::flog::FLOG;
 use crate::input_common::{
     CharEvent, CharInputStyle, ImplicitEvent, InputData, InputEventQueuer, ReadlineCmd,
-    WaitingForCursorPosition, R_END_INPUT_FUNCTIONS,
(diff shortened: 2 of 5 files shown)


CHANGELOG.rst
```diff
@@ -13,6 +13,9 @@ Scripting improvements
 Interactive improvements
 ------------------------
 - Autosuggestions are now also provided in  multi-line command lines. Like `ctrl-r`, autosuggestions operate only on the current line.
+- When typing a command while the previous one is still running, :kbd:`enter` will no longer execute the would-be command, and keys that are bound to shell commands will be ignored.
+  This mitigates a security issue where a command like ``cat malicious-file.txt`` could write terminal escape codes prompting the terminal to write arbitrary text to fish's standard input.
+  Such a malicious file can still potentially insert arbitrary text into the command line but can no longer execute it directly (:issue:`10987`).
```
OkamiW, line 18: Note that this change may not completely prevent the attacker to directly execute the command.
If the attacker manage to do the timing attack, malicious commands will be injected anyway.
And in case the terminal emulator allows "\n", the commands will be executed.

mqudsi:
I feel like we would have to test drive this behavior as our daily fish for an extended period of time to get a feel for whether or not this hinders natural expectations.

krobelus:
yep, I'll fix tests to make it more convenient for everyone to test drive.
Could also add a temporary feature flag so people don't have to carry the patch.

krobelus:
I added a feature flag, see `status features`.
I'm not yet sure which behavior I like better.

As for addressing the security issue::

I'll fix our OSC/DCS parser to drop unknown sequences in entirety (instead of stopping as soon as the prefix tells us that there is nothing interesting for us too extract). This will prevent the echoback requests that give the most freedom to insert near-arbitrary text.

However as mentioned in the foot issue, that doesn't work because we may only see parts of the sequence.
So we'd still need to flush or find a way to ask the terminal if it's done writing any replies (I don't know how that can be done, probably needs cryptography)

PR #10993, opened 2025-01-03
Added initial iOS support

kekeimiku:
Tested on an iPad Pro M2, which I have been using daily for over six months. 
It would be great if upstream support for building on iOS were available.

image


build.rs
```diff
@@ -199,14 +199,14 @@ fn have_gettext(target: &Target) -> Result<bool, Box<dyn Error>> {
 /// 0.5 MiB is small enough that we'd have to drastically reduce MAX_STACK_DEPTH to less than 10, so
 /// we instead use a workaround to increase the main thread size.
 fn has_small_stack(_: &Target) -> Result<bool, Box<dyn Error>> {
-    #[cfg(not(any(target_os = "macos", target_os = "netbsd")))]
+    #[cfg(not(any(target_os = "macos", target_os = "ios", target_os = "netbsd")))]
```
faho, line 202: These should probably be using something like our [bsd config](https://github.com/fish-shell/fish-shell/blob/996fec87f4114e8424bd0a92e94e74e487187688/build.rs#L91) - I think these platforms are 99% similar, and so it's easy to add something for macos and forget ios.

I *think* this would then be

```rust
#[cfg(not(any(apple, target_os = "netbsd")))]
```

?

ent.xml
```diff
@@ -0,0 +1,392 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
faho, line 1: This file should be somewhere else, like in build_tools with osx_distribution.xml.

Also ideally it would have a more descriptive name - "ios_ent.xml" (or whatever "ent" stands for).

.cargo/config.toml
```diff
@@ -0,0 +1,2 @@
+[env]
+IPHONEOS_DEPLOYMENT_TARGET="13.1"
```
faho, line 2: I would like to avoid this file.

As I understand, you would have to set this variable when building for ios, which sounds fine to me.

faho (changes requested):

faho:
This is not a platform that I have any interest in or ability to test, but as long as it's a few `#cfg` additions where we already special-case macos it's fine, I guess.

Out of curiosity how do you build this? Does it require an apple developer account?

----------

Note also that you want to rebase your branch on top of master before we can merge it.

Generally, I find that much nicer than back-merging anyway.

kekeimiku:
> Out of curiosity how do you build this? Does it require an apple developer account?

It does not require a developer account. 

Most of the time, iOS and macOS are very similar, so maybe #cfg(apple..)

.cargo/config.toml
kekeimiku: It must be specified as version 13+, Rust uses version 10+ by default, which will cause `pcre2` error.

Of course, you don't have to use this file.

faho: What I mean is I don't want to add a .cargo/config.toml if we can help it, and I don't believe this is important enough to justify it.

If you have to run `IPHONEOS_DEPLOYMENT_TARGET="13.1" cargo build` manually, then that's fine by me.

I would rather the few people who build for ios have to do a thing than force this file on everyone.

build.rs
kekeimiku: Currently, it seems so.

faho:
> Merge branch 'fish-shell:master' into rootless-ios-dev](https://github.com/fish-shell/fish-shell/pull/10993/commits/fe908859150711f448238a4ea3d2d4718b0b1e41

Please run `git rebase master` and then `git push --force-with-lease` the result.

Also don't merge master back into your branch.

kekeimiku:
@faho I just cleaned up the branch

.cargo/config.toml
zanchey: Adding this breaks the Linux build pipelines.

build_tools/ios_entitlements.xml
```diff
@@ -0,0 +1,392 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
```
zanchey, line 4: There's a lot of entitlements here! I presume this is just to get full disk access?

kekeimiku: > There's a lot of entitlements here! I presume this is just to get full disk access?

yes

ridiculousfish:
How does one run this? There's no Terminal app built into iOS.

kekeimiku:
> How does one run this? There's no Terminal app built into iOS.

jailbreak

build.rs
faho: What I am saying is I would like you to add something like that bsd config to our build.rs, and then use it.

That makes it easier to not accidentally forget ios when we add another macos special path.

faho:
Okay I think this isn't very complicated or wrong or anything, so the question really is just this:

- Is this a platform we want to support?

I am not going to be of any assistance because I have no apple devices, so unless one of the core devs steps forward and wants to take this up, I'm going to close this in a week or so.

kekeimiku:
If it can't be merged, that's OK; I'll maintain it in my own branch and distribute the binaries.

krobelus:
Besides the entitlements, it should be a one-line change if we change `#[cfg(any(target_os = "macos", target_os = "ios"))]` to `#[cfg(apple)]` (adding our own definition in `build.rs`, like we do for `#[cfg(bsd)]`)

zanchey:
I think it's fine to take it, a bit like Cygwin/Solaris support - if it breaks someone else will have to fix it because the number of developers with jailbroken iOS devices is pretty limited

PR #10998, opened 2025-01-04
refactor: misc cleanup

stefanboca:
## Description

Just some small fixes I noticed

## TODOs:

### DIFF
--- a/src/env/var.rs
+++ b/src/env/var.rs
@@ -199,7 +199,7 @@ impl EnvVar {
     }
 
     /// Returns a copy of the variable with new values.
-    pub fn setting_vals(&mut self, values: Vec<WString>) -> Self {
+    pub fn setting_vals(&self, values: Vec<WString>) -> Self {
         EnvVar {
             values: values.into(),
             flags: self.flags,
@@ -207,7 +207,7 @@ impl EnvVar {
     }
 
     /// Returns a copy of the variable with the export flag changed.
-    pub fn setting_exports(&mut self, export: bool) -> Self {
+    pub fn setting_exports(&self, export: bool) -> Self {
         let mut flags = self.flags;
         flags.set(EnvVarFlags::EXPORT, export);
         EnvVar {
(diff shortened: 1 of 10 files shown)


src/env/var.rs
```diff
@@ -111,8 +111,8 @@ impl Default for EnvVar {
     fn default() -> Self {
         use std::sync::OnceLock;
         /// A shared read-only empty list.
-        static EMPTY_LIST: OnceLock<Arc<Box<[WString]>>> = OnceLock::new();
-        let empty_list = EMPTY_LIST.get_or_init(|| Arc::new(Box::new([])));
+        static EMPTY_LIST: OnceLock<Arc<[WString]>> = OnceLock::new();
```
stefanboca, line 114: Is this `OnceLock` still needed? I would imagine initializing a new Arc isn't very expensive.

krobelus: I believe the reason for making the `Arc` a singleton instead of creating a new `Arc` everytime is that `Arc` owns a heap allocation, so this reduces the number of allocations.
I haven't checked how OnceLock is implemented but I think all it needs to do in the common case is to check if it's initialized (with the right memory ordering); even if multiple threads were doing that there would be no contention between the readers.

krobelus (approved):

src/signal.rs
```diff
@@ -426,7 +426,7 @@ fn equals_ascii_icase(left: &wstr, right: &wstr) -> bool {
         return false;
     }
     for (lc, rc) in left.chars().zip(right.chars()) {
-        if lc.to_ascii_lowercase() != rc.to_ascii_lowercase() {
+        if !lc.eq_ignore_ascii_case(&rc) {
```
mqudsi, line 429: "Same as `to_ascii_lowercase(a) == to_ascii_lowercase(b)`, but without allocating and copying temporaries."

src/env/var.rs
stefanboca: Makes sense. I'll leave it as is.

stefanboca: Makes sense. I'll leave it as is.

mqudsi (approved):

mqudsi (commented):
I personally prefer `assert_eq!(xxx, false)` because the `!` in `assert_eq!(!xxx)` can be missed (esp. after the ! from the macro) but given that we haven't disabled that lint I cannot complain too much about this change.

PR #11003, opened 2025-01-04
Make new ctrl-l logic more portable

mqudsi:
The behavior of the new scrollback-push is not consistent across various terminals and the functionality of ctrl-l is too important to break in the default bindings. This hack should preserve the benefits of scrollback-push while keeping the portability of the old clear-screen: first we ask the terminal to scroll everything offscreen *then* we clear the screen. The addition of clear-screen should be a no-op on any terminal that fully supports the desired behavior of invoking the new scrollback-push.

(There are no real performance-related concerns with ctrl-l.)

This handles multiple possible issues (observed or otherwise) with scrollback-push:
* fish starts a new prompt at the top of the screen, scrollback is lost, screen is not cleared (observed under Prompt/iOS, FreeBSD console session, Linux 6.1 console session -- note that Linux kernel removed scrollback in 5.9)
* termcap SF isn't supported
* termcap UP isn't supported ("Only a few terminal descriptions provide these commands, and most programs do not use them.")

It still breaks in an ugly way if terminal position reporting isn't supported, but I'll address that separately.

Update:

As @krobelus mentioned, a number of terminal emulators map the old ctrl-l to scroll-then-clear. But the new implementation (both current and with this patch) actually *breaks* scrollback for these tty emulators (tested: conhost/Windows 11 and under gnome terminal 3.46), losing the last screen altogether.

Under FreeBSD console sessions, `scrollback-buffer` additionally causes fish to hang indefinitely until some time after ctrl-c is used (my random guess is that waiting for cursor position isn't being properly async?).

With these regressions in mind, I would prefer to close this PR and instead revert the default `ctrl-l` binding to `clear-screen` (keeping `scrollback-push` around for anyone that wants to manually opt into it).

### DIFF
--- a/share/functions/__fish_shared_key_bindings.fish
+++ b/share/functions/__fish_shared_key_bindings.fish
@@ -66,7 +66,7 @@ function __fish_shared_key_bindings -d "Bindings shared between emacs and vi mod
     bind --preset $argv alt-l __fish_list_current_token
     bind --preset $argv alt-o __fish_preview_current_file
     bind --preset $argv alt-w __fish_whatis_current_token
-    bind --preset $argv ctrl-l scrollback-push repaint
+    bind --preset $argv ctrl-l scrollback-push clear-screen
     bind --preset $argv ctrl-c cancel-commandline
     bind --preset $argv ctrl-u backward-kill-line
     bind --preset $argv ctrl-w backward-kill-path-component


ccoVeille:
I'm a simple fish user and fan, but I agree with you

krobelus:
> Under FreeBSD console sessions, `scrollback-buffer` additionally causes fish to hang indefinitely
> until some time after ctrl-c is used

One question I don't yet know the answer to is whether our approach to decoding keyboard input
is correct.

We essentially do:

```rust
fn parse_escape_sequence() {
    while select(timeout=0) {
        read(next_byte);
    }
}
```

Which breaks if `select()` returns "no data available" in the middle of an escape sequence.

I wonder if the PTY can possibly be full so that we can only read parts of a (small) escape sequences?
My intuition is that writes smaller than `PIPE_BUF` should available but probably there is no guarantee.

This tearing definitely happened in Prompt (but that one has other bugs).
You're saying it happened in the FreeBSD console too; I'll try that.

> (my random guess is that waiting for cursor position isn't being properly async?).

viewed from the outside, it should **not** be async.  When we request the cursor position, we
defer handling of all other input until we receive a cursor position response.  This eliminates
a large class of problems (users expect input to be handled in order). Also makes it easier
to find/diagnose/debug problems.

> Prompt/iOS,

Currently, we don't support terminals that break on unknown CSI/OSC sequences, but if we stick
to the current approach, I'm happy to add a terminal-specific workaround or config option.

> * termcap SF isn't supported
> * termcap UP isn't supported ("Only a few terminal descriptions provide these commands, and most programs do not use them.")
> It still breaks in an ugly way if terminal position reporting isn't supported, but I'll address that separately.

will push a fix for these, which fixes the Linux console

I'm not 100% sure if it's the best choice to add a smart fallback when cursor position reporting
is not supported.  This is because so far I don't know how to find a terminal that doesn't support
it (it may be ubiquitous?); and requiring it simplifies things overall (for terminal developers).
But the status quo is definitely a graceful fallback.

PR #11005, opened 2025-01-04
Add completions for Google Cloud commands.

cornmander:
## Description

The `gcloud` and `gsutil` Google Cloud commands use argcomplete, so integrating them is easy with the `__fish_argcomplete_complete` function.


mqudsi:
LGTM. Merged with thanks.

PR #11020, opened 2025-01-07
Add support tier document

faho:
## Description

This is a quick and dirty draft of a "support" tier document of sorts to provide some guidance and serve as a checklist of sorts before a release.

I may have missed some platforms, and none of the rankings are final. A lot of them are incomplete.

E.g. I do not know which terminals our core developers run, which is important here.

This is *not* an RFC because users really don't matter here - what matters is what CI we have and what the regular contributors use.

It is a reflection of the reality of the situation, not what we *want* it to be.

### DIFF
--- a/doc_internal/support-tiers.md
+++ b/doc_internal/support-tiers.md
@@ -1,0 +1,119 @@
+# "Support" Tierlist
+
+This document is guidance on what platforms and terminals fish supports to what degree.
+
+It is supposed to explain to users to what degree they can expect fish to work, and to help us decide when features impact compatibility.
+
+In an ideal world, everything would work everywhere. But of course, we have limited resources and need to not only decide where to put them,
+but also see where they currently *happen to be*. And there are times where we need to decide between supporting most users better,
+and supporting a small group of users. This may even mean dropping platforms or terminals when we can't make them work without impacting
+more important ones.
+
+It is also a list of targets we want to check before release.
+
+For the purposes of this list, terminal multiplexers like tmux and other forms of terminal-in-a-terminal like mc are "terminals".
+
+None of this is a legal document, and fish remains developed by hobbyists in their free time. Nobody is allowed to call me up at night to fix any of this.
+
+## The Tiers
+
+1. Working git at all times - can you expect a random checkout to be usable on this particular platform/terminal?
+   If this is broken, we should revert the offending commits ASAP.
+2. Tested before release, we don't want to ship with it broken
+3. We hope this works and will try to fix it if it doesn't, but no guarantees
+4. We don't care. If it works, that's cool.
+
+These tiers depend on how important a given target is to us, and how much information we get about it.
+
+That means, a tier 1 target requires a core developer to run it in anger, and working CI for operating systems/architectures.
+Since we don't have automated testing of terminals, CI isn't required there.
+
+A tier 2 target needs us to invest a conscious effort to test it before release, but of course if we don't run it ourselves we may not catch things
+that aren't covered in tests. Users can help here. So a tier 2 target is better the more it is used. Note: We may decide to skip some tests for patch releases.
+
+A tier 3 target is best effort. These can move up if they become more important, or someone on the core team uses and maintains it.
+Or it can move down if it conflicts with improving other targets.
+
+A tier 4 target is one that we aren't interested in, bug reports will most likely be closed, but we may invest some time trying to help.
+If this works often enough it may be promoted to tier 3. We may also reject patches if they are too invasive.
+
+The distinction between tier 1 and 2 is really just an expression of our confidence that any random git checkout or release will work.
+The distinction between tier 2 and 3 is much more important - these aren't supported beyond "we'll try if you tell us".
+And tier 4 is just a list of platforms we know fish doesn't work on.
+
+## The current targets: Operating Systems / Architectures
+
+Unless anything else is mentioned, these refer to current mainstream versions.
+
+Tier 1:
+
+- Linux, with current glibc, on x86_64
+- macOS, aarch64
+
+- WSL (as "basically glibc linux" with some use among the core developers)
+
+Tier 2+ (have CI but not enough use among core devs):
+
+- Linux, musl, x86_64
+- Linux, older glibc, x86_64
+- FreeBSD
+
+Tier 2:
+
+- NetBSD, x86_64
+- Linux, glibc, aarch64?
+- macOS x86_64?
+
+Tier 3:
+
+- OpenBSD
+- OpenSolaris/Illumos/OpenIndiana?
+- Architectures other than x86_64 and aarch64 (and possibly i686)
+
+Tier 4:
+- Architectures and operating systems not supported by rust.
+- Cygwin/Msys
+- Haiku?
+- [...]
+- [...]
+- [Long scroll, look at your watch or phone]
+- [...]
+- Windows native
+- Nonstop
+
+## The current targets: Terminals
+
+Tier 1:
+
+- Konsole
+- MS Terminal
+- tmux (used in our CI)
+- other terms the core devs use
+
+Tier 2:
+
+- Ghostty
+- Wezterm
+- Alacritty
+- Kitty?
+- xterm?
+- iTerm?
+- Terminal.app?
+- Gnome Terminal / Console?
+- zellij?
+- foot?
+
+Tier 3:
+
+- putty?
+- linux tty?
+- freebsd tty?
+- rxvt(-unicode)
+- Midnight Commander?
+
+Tier 4:
+
+- emacs ansi-term
+- amazon web console
+- dvtm
+- cool-retro-term
--- a/doc_internal/support-tiers.md
+++ b/doc_internal/support-tiers.md
@@ -25,7 +25,8 @@ None of this is a legal document, and fish remains developed by hobbyists in the
 
 These tiers depend on how important a given target is to us, and how much information we get about it.
 
-That means, a tier 1 target requires a core developer to run it in anger, and working CI for operating systems/architectures.
+That means, a tier 1 target requires a core developer to run it regularly or enough users that we are confident we will catch bugs quickly,
+and working CI for operating systems/architectures.
 Since we don't have automated testing of terminals, CI isn't required there.
 
 A tier 2 target needs us to invest a conscious effort to test it before release, but of course if we don't run it ourselves we may not catch things
@@ -83,11 +84,16 @@ Tier 4:
 
 ## The current targets: Terminals
 
+These are expected to be used in their majority configuration - usually the defaults.
+No switching $TERM required.
+
 Tier 1:
 
 - Konsole
-- MS Terminal
+- MS Terminal / Conhost
+- Gnome Terminal
 - tmux (used in our CI)
+- foot
 - other terms the core devs use
 
 Tier 2:
@@ -95,13 +101,11 @@ Tier 2:
 - Ghostty
 - Wezterm
 - Alacritty
-- Kitty?
-- xterm?
+- Kitty
+- xterm
 - iTerm?
 - Terminal.app?
-- Gnome Terminal / Console?
-- zellij?
-- foot?
+- VSCode / xterm.js
 
 Tier 3:
 
@@ -110,6 +114,10 @@ Tier 3:
 - freebsd tty?
 - rxvt(-unicode)
 - Midnight Commander?
+- termux
+- screen?
+- st?
+- zellij?
 
 Tier 4:
 


faho:
(this could also be a wiki page, because really only the latest version matters. it is entirely unrelated to the code)

septatrix:
> I may have missed some platforms, and none of the rankings are final. A lot of them are incomplete.

Adding Termux somewhere would be nice, maybe lvl 3. Same for the VSCode console which should be xterm.js (though I am not sure how customizable that is regarding CSI etc so listing them separate might make sense)

faho:
We do not have termux CI or core devs, so it would at most be a tier 3.

I don't know if any core dev uses VS Code.

--------

What this doesn't mention is how many workarounds we add to terminals, or *can* add (e.g. a terminal may not be detectable by variable).

And it doesn't mention customization, we would usually assume the defaults.

doc_internal/support-tiers.md
```diff
@@ -0,0 +1,119 @@
+# "Support" Tierlist
```
krobelus, line 1: in future it will be nice to have an externally-facing document of what we are looking for in a terminal, i.e.
- required things like
  - termcaps we need for core functionality
  - the ability to parse arbitrary CSI and OSC commands
- nice-to have features like OSC 133 etc.

then any terminal developer that wants to fully support fish can simply tick
all the boxes in that list and be done, no need to read source code.

doc_internal/support-tiers.md
```diff
@@ -0,0 +1,119 @@
...
+Tier 4:
+- Architectures and operating systems not supported by rust.
+- Cygwin/Msys
+- Haiku?
+- [...]
+- [...]
+- [Long scroll, look at your watch or phone]
+- [...]
+- Windows native
+- Nonstop
```
krobelus, line 82: :)

doc_internal/support-tiers.md
```diff
@@ -0,0 +1,119 @@
...
+
+1. Working git at all times - can you expect a random checkout to be usable on this particular platform/terminal?
+   If this is broken, we should revert the offending commits ASAP.
+2. Tested before release, we don't want to ship with it broken
+3. We hope this works and will try to fix it if it doesn't, but no guarantees
+4. We don't care. If it works, that's cool.
+
+These tiers depend on how important a given target is to us, and how much information we get about it.
+
+That means, a tier 1 target requires a core developer to run it in anger, and working CI for operating systems/architectures.
```
krobelus, line 28: in anger?

doc_internal/support-tiers.md
```diff
@@ -0,0 +1,119 @@
...
+- Ghostty
+- Wezterm
+- Alacritty
+- Kitty?
+- xterm?
+- iTerm?
+- Terminal.app?
+- Gnome Terminal / Console?
+- zellij?
+- foot?
```
krobelus, line 104: I use foot daily, it's great. Also iTerm but only with tmux which is probably why I didn't notice key encoding bugs

doc_internal/support-tiers.md
```diff
@@ -0,0 +1,119 @@
...
+
+Tier 3:
+
+- putty?
+- linux tty?
+- freebsd tty?
+- rxvt(-unicode)
+- Midnight Commander?
```
krobelus, line 112: > We hope this works and will try to fix it if it doesn't, but no guarantees

I expect we won't have too many feature-related problems in keeping (most of) these working; it's mainly about remembering to test them.
I'm happy to do basic testing for linux tty and `mc` but I don't really use either.

doc_internal/support-tiers.md
```diff
@@ -0,0 +1,119 @@
...
+- tmux (used in our CI)
+- other terms the core devs use
+
+Tier 2:
+
+- Ghostty
+- Wezterm
+- Alacritty
+- Kitty?
+- xterm?
```
krobelus, line 99: conhost may be missing (not sure which tier but we can definitely keep it working)

doc_internal/support-tiers.md
```diff
@@ -0,0 +1,119 @@
...
+- iTerm?
+- Terminal.app?
+- Gnome Terminal / Console?
```
krobelus, line 102: I'd definitely add VSCode console, maybe tier 2, I think that's an important one (with a lot of potential for problems unfortunately)

doc_internal/support-tiers.md
```diff
@@ -0,0 +1,119 @@
...
+Unless anything else is mentioned, these refer to current mainstream versions.
+
+Tier 1:
+
+- Linux, with current glibc, on x86_64
+- macOS, aarch64
+
+- WSL (as "basically glibc linux" with some use among the core developers)
+
+Tier 2+ (have CI but not enough use among core devs):
```
krobelus, line 55: I don't think "core dev" is the perfect criteria for 1st tier.
I think it's more like "is there someone using it with recent master who will point out bugs"

zanchey:
I reckon this is better in the wiki

doc_internal/support-tiers.md
zanchey: [possibly this idiom does not translate well](https://english.stackexchange.com/a/30943/99151)

doc_internal/support-tiers.md
mqudsi: I would want to distinguish between how we currently treat the kernel tty emulators and the way we treat the others in tier 3 here. I would personally move them to tier 2 but with the caveat that they function "perfectly but with features possibly missing", however we want to phrase that.

doc_internal/support-tiers.md
```diff
@@ -0,0 +1,119 @@
...
+
+Tier 3:
+
+- putty?
+- linux tty?
+- freebsd tty?
+- rxvt(-unicode)
+- Midnight Commander?
+
+Tier 4:
```
mqudsi, line 114: Isn't this basically everything else?

doc_internal/support-tiers.md
faho: Currently, the only people who are around enough that I'm confident enough we'll get bugs *are* the core devs. If we had some dedicated group of testers, that would also be enough, but I don't think we currently have that.

It's not that we have some magic power, just that we're around.

(plus the distinction between tiers 1 and 2 isn't that important)

doc_internal/support-tiers.md
```diff
@@ -0,0 +1,119 @@
...
+- Nonstop
+
+## The current targets: Terminals
+
+Tier 1:
+
+- Konsole
+- MS Terminal
+- tmux (used in our CI)
+- other terms the core devs use
```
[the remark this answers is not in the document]
faho: > I would add conhost and gnome-terminal to tier 1

Who among us uses gnome-terminal? Are we confident enough that we'll see bugs reported within a day or so if something breaks there?

Would we get bugs if there's an issue in another conhost-using terminal but not in MS Terminal? Is that even a thing? Or do we just mention MS Terminal because that's the default one we're also using?

> I don't know if we want to specify for tmux that TERM has to be screen-256color specifically or not.

I think the terms should be using the majority settings, usually the default.

doc_internal/support-tiers.md
faho: This is really about terminals *breaking* more so than supporting the full set of features.

E.g. Konsole is tier 1 because I use it, but it does not support the kitty keyboard protocol or some of the neat OSC sequences. It's fine if a terminal doesn't support all our features as long as what the terminal supports *works*.

But yeah, I can add some text to that effect.

doc_internal/support-tiers.md
faho: E.g. for FreeBSD we have CI, which means it'll keep building and passing the tests,
but while @mqudsi uses it I believe he doesn't use it constantly.

That means if something in FreeBSD breaks that isn't caught by the test suite we'll have some delay in catching it.

doc_internal/support-tiers.md
faho: We could rephrase this as "regularly". What I mean to say is that people who are around use that target so regularly that we're likely to catch regressions within roughly "a day" (or possibly week, it's all a bit wishy washy).

doc_internal/support-tiers.md
krobelus: When you describe the state of the world based on a superficial criteria,
this will mislead people to believe that there is an inherent causal relationship here,
and that will perpetuate the status quo.

"No one cares to help so I need to do it myself" is a self-fulfilling prophecy.

The FreeBSD example is irrelevant to the question of "who needs to test it".

iTerm is an example where users deliver most bug reports.

---

In general, I don't anticipate us having a reason to drop support for any more terminals unless they misinterpret some command which is very unlikely.

faho: This is fundamentally a statement of *confidence*.

For Tier 1, we're confident that "any" random git checkout will work.

For Tier 2, we're confident that any *release* will work. We may not know about the state of git, but we'll test and fix before release.

For Tier 3, we hope that it'll work (we believe it usually does work) and will fix it once we're told.

Tier 1 implies that we have enough visibility into these targets that we find out at least soon after a breaking commit was added, and will revert or fix it quickly. That means I want someone who is regularly here regularly running it, which currently means one of us.

For some targets we may have enough users who regularly test git checkouts,
but really that's just linux+glibc and macos+iterm again.

> "No one cares to help so I need to do it myself" is a self-fulfilling prophecy.

I didn't say that anywhere. We *will* help for anything up to tier 3, and may for tier 4.

And tier 4 is a list of targets that have some deficiency, so improving the target is going to be the first step. E.g. Cygwin is currently broken on their end.

If you think any of that isn't clear enough, please help me find better wording.

zanchey:
I'm still not really clear what problem this is attempting to solve. The terminal I use on a day-to-day basis might change tomorrow and I don't know that "remembering to update the fish wiki" is going to be on my list of things to do. It seems more like a point-in-time snapshot than a guarantee of anything in particular?

faho:
> I'm still not really clear what problem this is attempting to solve. 

This came up because we were talking about whether it is okay to break emacs ansi-term even more in order to support new features like the kitty keyboard protocol.

So tier 4 is a list of things that we don't have to support at all. If we notice that ansi-term doesn't handle some new feature we're doing, that's fine. If fish no longer works on Nonstop, okay - send in patches and we'll consider them.

Additionally, it is useful as a checklist before release (remember to test NetBSD and Wezterm!), and as guidance stating our confidence on how well specific targets work.

NetBSD will probably break fairly regularly in git, but should work in releases, while Linux and iTerm should work all the time.

We *could* also break it down into two parts: Things we support and will test before release, and things we do not support. But I believe having a bit more granularity is useful even if only to see that yes, we do need your bug reports for OpenBSD because none of us is running that.

> The terminal I use on a day-to-day basis might change tomorrow

It's fine if this isn't always entirely up-to-date. The distinction between tiers 1, 2 and even 3 isn't *that* important - it's only a guess anyway. And it's not like you switching from iTerm to Ghostty is going to break iTerm immediately.

doc_internal/support-tiers.md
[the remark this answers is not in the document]
mqudsi: fwiw Gnome Terminal is my default on my Linux laptop, conhost is the default on my Windows/WSL workstations. I use FreeBSD on the daily but usually over ssh from Windows/Linux, so I tend to catch fish breakage more than terminal breakage there.

I agree terminfo is useless because no one apart from tmux and kitty bother to identify as anything other than xterm. But screen-256color vs tmux-256color is fine, my concern was with people using xterm-256color + tmux, which is a common misconfiguration I've seen in the wild. But you're right, it is a misconfiguration which should be easily remedied so it's not worth worrying over.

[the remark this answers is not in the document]
Xiretza: > no one apart from tmux and kitty bother to identify as anything other than xterm

alacritty too, not that it makes that much of a difference.

mqudsi:
My reasons for specifically highlighting conhost and gnome terminal is because they (unlike iTerm 2, Windows Terminal, foot, tmux, etc) are system-provided packages that don't get updated out-of-band. Anyone going out of their way to install a 3rd party terminal can be reasonably told "update to a newer version of said terminal" but we can't do that for conhost, gnome terminal, apple terminal.

I think we might want to rethink the distinction between tier 1 and tier 2 (as you said, there isn't a great difference between them in the document as currently written) on this basis. tier 1 are ones we should support older versions of (people still on Windows 10 won't see updates that Windows Terminal introduces to its conhost fork, and there's a whole lot of them; the latest Debian stable only has gnome terminal 3.46 available and people are stuck on older versions of Debian/mainstream distros than that; we support OS X 10.9+ and it's fine to require legacy OS X users to use a third party terminal but we should support the system apple terminal for supported macOS (not OS X) versions).

Perhaps tier 1 would be these where we will take into account compatibility with old releases vs tier 2 would be those where we will take into account the restrictions of the latest release? And anything breaking tier 1 or tier 2 would be patched or reverted?

doc_internal/support-tiers.md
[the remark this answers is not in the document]
faho: >I agree terminfo is useless because no one apart from tmux and kitty bother to identify as anything other than xterm. But screen-256color vs tmux-256color is fine, my concern was with people using xterm-256color + tmux

Nope, there are fewer differences in the capabilities that we use between tmux-256color and xterm-256color than to screen-256color.

They are (tmux then xterm):

```
# xterm clearing scrollback, a horrible idea
clear_screen: '\E[H\E[J', '\E[H\E[2J'.
# Useless, both work
cursor_normal: '\E[34h\E[?25h', '\E[?12l\E[?25h'.
# Useless, both work:
cursor_up: '\EM', '\E[A'.
# Useless - both work
exit_attribute_mode: '\E[m^O', '\E(B\E[m'.
# Not actually used anymore, we can handle either
key_home: '\E[1~', '\EOH'.
```

So yes, terminfo is useless.

mqudsi:
Sorry, I wasn't saying that would be "for every commit" but rather a different way of segmenting tier 1 and tier 2 instead of "for every commit" vs "for every release"

sidebar: I have concerns with "for every release" where it might be difficult (psychologically, not technically) to see features backed out months after development leading to a lot of resistance or clouded judgement.

faho:
> I have concerns with "for every release" where it might be difficult (psychologically, not technically) to see features backed out months after development 

Nobody is saying you can't test before.

But either these are important enough to be a release blocker, or they aren't and shouldn't be in that tier.

Also you can fix it. In practice I can't remember something we ever fully backed out after months. If it was really completely broken on an OS, worst case we'd cfg it out, and if it was broken on a terminal, we'd check some variable.

>  a different way of segmenting tier 1 and tier 2

Sure, but that's still a gigantic and slow to test matrix.

If you want to test old terminals, you need to create an old VM, install them there and manually check them. And for macOS specifically, I don't even have that option.

*And* it is still aspirational instead of descriptive, and I don't think that's very useful.

And it loses the grading of how well things work in practice and how quick we would be to find and fix breakage.

mqudsi:
My point is that there isn't a point in specifying how often a terminal should be tested if there is no way to guarantee that it happens on that schedule (i.e. mandate commits go through PRs and have a magic CI that can verify compatibility across all the tier 1 list, because that's just not possible and also not how we want to live life).

Your tier 1 and tier 2 are, technically, going to be caught in the same way: whenever one of the core devs or neigh-mythical nightly users catches the regression in the time between releases, and, ideally, in a regular pre-release effort to make sure that nothing slipped. You can't *make* anyone check them more regularly than that, no matter whether they're tier 1 or tier 2, so my point was that the difference between tier 1 and tier 2 on *that* basis is a "distinction without a difference".

The *only* prescriptive part of this document is going to be "once made aware of it, revert changes breaking these terminals (when and where they are found)".

I am saying that gives us an opportunity to rethink the classification based off of what we can realistically expect:

* Tier 1: terminals where breakage for tty versions m through n will work "out of the box", either with compatibility fixes or by reverting unsupported features if there is absolutely no way to paper over the breakage
* Tier 2: terminals where breakage for the latest tty version n will be reverted, unless the issue has been reported upstream and a newer release with a patch will be released "soon", again either with compatibility hacks (that may include *gracefully* disabling specific fish features) or by reverting features that cannot be supported
* Tier 3: terminals where there may (or may not!) be known breakage that can be worked around with configuration changes (to fish, the terminal emulator, or the os); we will also accept PRs to fix this with no burden on providing the fixes ourselves
* Tier 4: unsupported terminals where fish is known to be broken and we have no inclination to fix it because of the esoteric/proprietary/obscure/crappy nature of the terminal in question

In this breakdown, tier 1 and tier 2 would *ideally* be explicitly tested before release. Realistically, "we" may or may not be able to test everything in older versions of tier 1 or every terminal in tier 2 on our own: if an issue is user/dev-discovered post-release the expectation is that we would release a minor update to fix the breakage *somehow*. tier 3 we have no burden to schedule bugfix releases for, tier 4 is what it is.

faho:
> I am saying that gives us an opportunity to rethink the classification based off of what we can realistically expect:

Okay, that's fairly heavy on prescriptive policy. I don't really want us to have a legal document here, and I don't believe that sort of prescription is warranted.

We haven't had a huge amount of version trouble in practice (consider that people ssh from new desktops to old servers, not the other way around).

And I don't think we *can* make a distinction of "xterm gets 5 versions of support, gnome-terminal gets 7" - because how often these release can change, and it only matters what is used in practice (i.e. what is packaged by LTS distributions).

And that has massive potential for bikeshedding - which is why I would like to have the status quo explained, which is relatively easy and relatively objective. With some light policy on top in the form of "we do not support emacs ansi-term and these three other terminals".

Consider [rust's target tier policy](https://doc.rust-lang.org/nightly/rustc/target-tier-policy.html):

> Rust provides no guarantees about tier 3 targets; they exist in the codebase, but may or may not build.
    Rust's continuous integration checks that tier 2 targets will always build, but they may or may not pass tests.
    Rust's continuous integration checks that tier 1 targets will always build and pass tests.

Since we do not have CI for terminals, the closest we can get is "we know it works"/"we think it works"/"we hope it works" (and the implicit fourth "we don't care if it works").

doc_internal/support-tiers.md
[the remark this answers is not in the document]
Conan-Kudo: Note that GNOME Terminal (gnome-terminal), GNOME Console (kgx), and GNOME Terminal (ptyxis) all use VTE, so they should behave more or less the same.

doc_internal/support-tiers.md
Conan-Kudo: Is this something you would _want_ Konsole to support? I don't recall if anyone's filed bug reports with KDE asking for these things...

faho:
Alright, I don't believe in its current state this is useful.

If someone were to attempt again, I think the highest value we can get from this is: A clear policy on what bugs we're going to *reject*. E.g. if a terminal mishandles CSI u, we *have* been closing the bugs as "upstream". Tell people that the issue is in their terminal, tell them about a workaround and (optionally) send a patch to that terminal.

And we have been rejecting bugs asking for support for macOS 10.5 (which would require abandoning rust) and Windows.

This would either involve listing *features* we require (rust support, a posix environment, certain escape sequences) or platforms that lack them.

The second highest value is listing platforms that we know well and that we know work well, but those are for the most part going to be the mainstream unixy platforms (linux+glibc, macos), and that's where the bikeshedding starts because everyone wants their favorite platform included. It's hard to make a good assessment here.

Issue #1671, opened 2014-09-08
no input function for next/prev history (without search)


timbertson:
I'd like to rebind my up/down keys to be pure `history-backward` and `history-forward` actions, rather than the default `up-or-search` / `down-or-search` functions, as discussed in #1668.

Currently, there are no such input functions - only the `history-search-backward/forward` variant.

(I expect I'll actually want `up-or-history-backward`, but like `up-or-search` that can probably just be a user-level function).

hupfdule:
As I stumbled upon the same problem, may I ask if this is likely to be happen?

The current history-search behaviour is my main obstacle when working with fish.

krobelus:
Sure, someone just needs to implement it. It would be a similar change as https://github.com/fish-shell/fish-shell/pull/6143, which added prefix history search as an alternative to token history search. 
For this I'd just add a history search method that matches anything.

nougcat:
late happy 10 year anniversary to this issue 🥳

musjj:
Is there anyone working on this feature?

PR #11021, opened 2025-01-07
Fix ctrl-r history search inconsistent ordering

kerty0:
## Description

When searching, `history_pager_search` returns elements in the order in which it encountered them. For example, given the sequence `(b, a, b, c)`, a backward search would return `(b, a, c)`, but a forward search would return `(a, b, c)`. This also causes `history_index_start` to be unequal to `0` when searching forward if the first element has a duplicate earlier in the searched interval.
Additionally, it skips every 13th element.

Now, when searching forward, `history_pager_search` finds a new `history_index_start` and only when searches backward. Also, `history_index_end` now points to an element outside the history that the search displays (like `history_index_start`).

I found this bugs in 3.7.1, 4.0b1 and current Git version.

PS: while debugging Fish with `lldb`, I found out that you can write your own summary providers, so I have written some for Fish types (mainly Utf32Str and Utf32String). Should I commit them to the main repo and maybe add a mention of them to `CONTRIBUTING.rst`, or is there a better place for this?

### DIFF
--- a/src/history.rs
+++ b/src/history.rs
@@ -1905,6 +1905,11 @@ impl HistorySearch {
 
             // We're done if it's empty or we cancelled.
             let Some(item) = self.history.item_at_index(index) else {
+                self.current_index = match direction {
+                    SearchDirection::Backward => self.history.size() + 1,
+                    SearchDirection::Forward => 0,
+                };
+                self.current_item = None;
                 return false;
             };
 
@@ -1923,6 +1928,12 @@ impl HistorySearch {
             self.current_index = index;
             return true;
         }
+    }
+
+    /// Move current index so there is `value` matches in between new and old indexes
+    pub fn search_forward(&mut self, value: usize) {
+        while self.go_to_next_match(SearchDirection::Forward) && self.deduper.len() <= value {}
+        self.deduper.clear();
     }
 
     /// Returns the current search result item.
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -5083,6 +5083,7 @@ struct HistoryPagerResult {
     matched_commands: Vec<Completion>,
     range: Range<usize>,
     first_shown: usize,
+    motion: Option<SelectionMotion>,
 }
 
 #[derive(Eq, PartialEq)]
@@ -5095,6 +5096,7 @@ enum HistoryPagerInvocation {
 fn history_pager_search(
     history: &Arc<History>,
     direction: SearchDirection,
+    motion: Option<SelectionMotion>,
     history_index: usize,
     search_string: &wstr,
 ) -> HistoryPagerResult {
@@ -5144,10 +5146,22 @@ fn history_pager_search(
         next_match_found = search.go_to_next_match(SearchDirection::Backward);
     }
     let last_index = search.current_index();
-    HistoryPagerResult {
-        matched_commands: completions,
-        range: first_index..last_index,
-        first_shown,
+    let range = first_index..last_index;
+    if completions.is_empty() && range != (0..history.size() + 1) {
+        history_pager_search(
+            history,
+            SearchDirection::Forward,
+            Some(SelectionMotion::Prev),
+            history.size() + 1,
+            search_string,
+        )
+    } else {
+        HistoryPagerResult {
+            matched_commands: completions,
+            range,
+            first_shown,
+            motion,
+        }
     }
 }
 
@@ -5184,7 +5198,7 @@ impl ReaderData {
         let performer = {
             let history = self.history.clone();
             let search_term = search_term.clone();
-            move || history_pager_search(&history, direction, index, &search_term)
+            move || history_pager_search(&history, direction, motion, index, &search_term)
         };
         let canary = Rc::downgrade(&self.canary);
         let completion = move |zelf: &mut Reader, result: HistoryPagerResult| {
@@ -5217,7 +5231,7 @@ impl ReaderData {
                 zelf.pager.set_selected_completion_index(old_pager_index);
                 zelf.pager_selection_changed();
             }
-            if let Some(motion) = motion {
+            if let Some(motion) = result.motion {
                 zelf.select_completion_in_direction(motion, true);
             }
             zelf.super_highlight_me_plenty();
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -5180,9 +5180,9 @@ impl ReaderData {
                 }
             }
             HistoryPagerInvocation::Refresh => {
-                // Redo the previous search previous direction.
+                // Make backward search from current position
                 let history_pager = self.history_pager.as_ref().unwrap();
-                direction = history_pager.direction;
+                direction = SearchDirection::Backward;
                 index = history_pager.history_index_start;
                 old_pager_index = Some(self.pager.selected_completion_index());
             }
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -1271,6 +1271,7 @@ impl ReaderData {
                 if self.history_pager.is_some() {
                     self.fill_history_pager(
                         HistoryPagerInvocation::Anew,
+                        Some(SelectionMotion::Next),
                         SearchDirection::Backward,
                     );
                     return;
@@ -2723,6 +2724,7 @@ impl<'a> Reader<'a> {
                     }
                     self.fill_history_pager(
                         HistoryPagerInvocation::Advance,
+                        Some(SelectionMotion::Next),
                         SearchDirection::Forward,
                     );
                     return;
@@ -2992,6 +2994,7 @@ impl<'a> Reader<'a> {
                     }
                     self.fill_history_pager(
                         HistoryPagerInvocation::Advance,
+                        Some(SelectionMotion::Next),
                         SearchDirection::Backward,
                     );
                     return;
@@ -3071,6 +3074,7 @@ impl<'a> Reader<'a> {
                     self.history.save();
                     self.fill_history_pager(
                         HistoryPagerInvocation::Refresh,
+                        None,
                         SearchDirection::Backward,
                     );
                 }
@@ -5151,6 +5155,7 @@ impl ReaderData {
     fn fill_history_pager(
         &mut self,
         why: HistoryPagerInvocation,
+        motion: Option<SelectionMotion>,
         mut direction: SearchDirection, /* = Backward */
     ) {
         let index;
@@ -5172,7 +5177,7 @@ impl ReaderData {
                 let history_pager = self.history_pager.as_ref().unwrap();
                 direction = SearchDirection::Backward;
                 index = history_pager.start;
-                old_pager_index = Some(self.pager.selected_completion_index());
+                old_pager_index = self.pager.selected_completion_index();
             }
         }
         let search_term = self.pager.search_field_line.text().to_owned();
@@ -5209,11 +5214,11 @@ impl ReaderData {
                 };
             zelf.pager.set_completions(&result.matched_commands, false);
             if why == HistoryPagerInvocation::Refresh {
-                zelf.pager
-                    .set_selected_completion_index(old_pager_index.unwrap());
+                zelf.pager.set_selected_completion_index(old_pager_index);
                 zelf.pager_selection_changed();
-            } else {
-                zelf.select_completion_in_direction(SelectionMotion::Next, true);
+            }
+            if let Some(motion) = motion {
+                zelf.select_completion_in_direction(motion, true);
             }
             zelf.super_highlight_me_plenty();
             zelf.layout_and_repaint(L!("history-pager"));
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -145,7 +145,7 @@ use crate::wcstringutil::{
 };
 use crate::wildcard::wildcard_has;
 use crate::wutil::{fstat, perror};
-use crate::{abbrs, event, function, history};
+use crate::{abbrs, event, function};
 
 /// A description of where fish is in the process of exiting.
 #[repr(u8)]
@@ -521,7 +521,7 @@ pub struct ReaderData {
     /// The history search.
     history_search: ReaderHistorySearch,
     /// In-pager history search.
-    history_pager: Option<HistoryPager>,
+    history_pager: Option<Range<usize>>,
 
     /// The cursor selection mode.
     cursor_selection_mode: CursorSelectionMode,
@@ -2717,7 +2717,7 @@ impl<'a> Reader<'a> {
             }
             rl::PagerToggleSearch => {
                 if let Some(history_pager) = &self.history_pager {
-                    if history_pager.history_index_start == 0 {
+                    if history_pager.start == 0 {
                         self.flash();
                         return;
                     }
@@ -2986,7 +2986,7 @@ impl<'a> Reader<'a> {
             }
             rl::HistoryPager => {
                 if let Some(history_pager) = &self.history_pager {
-                    if !history_pager.can_go_backwards {
+                    if history_pager.end > self.history.size() {
                         self.flash();
                         return;
                     }
@@ -3001,12 +3001,7 @@ impl<'a> Reader<'a> {
                 self.cycle_command_line = self.command_line.text().to_owned();
                 self.cycle_cursor_pos = self.command_line.position();
 
-                self.history_pager = Some(HistoryPager {
-                    direction: SearchDirection::Backward,
-                    history_index_start: 0,
-                    history_index_end: 0,
-                    can_go_backwards: false,
-                });
+                self.history_pager = Some(0..1);
                 // Update the pager data.
                 self.pager.set_search_field_shown(true);
                 self.pager.set_prefix(
@@ -5082,8 +5077,7 @@ impl<'a> Reader<'a> {
 
 struct HistoryPagerResult {
     matched_commands: Vec<Completion>,
-    final_index: usize,
-    have_more_results: bool,
+    range: Range<usize>,
 }
 
 #[derive(Eq, PartialEq)]
@@ -5091,15 +5085,6 @@ enum HistoryPagerInvocation {
     Anew,
     Advance,
     Refresh,
-}
-
-struct HistoryPager {
-    /// The direction of the last successful history pager search.
-    direction: SearchDirection,
-    /// The range in history covered by the history pager's current page.
-    history_index_start: usize,
-    history_index_end: usize,
-    can_go_backwards: bool,
 }
 
 fn history_pager_search(
@@ -5114,29 +5099,34 @@ fn history_pager_search(
     // We can still push fish further upward in case the first entry is multiline,
     // but that can't really be helped.
     // (subtract 2 for the search line and the prompt)
-    let page_size = usize::try_from(std::cmp::max(termsize_last().height / 2 - 2, 12)).unwrap();
-
-    let mut completions = vec![];
+    let page_size = usize::try_from(cmp::max(termsize_last().height / 2 - 2, 12)).unwrap();
+    let mut completions = Vec::with_capacity(page_size);
     let mut search = HistorySearch::new_with(
         history.clone(),
         search_string.to_owned(),
-        history::SearchType::ContainsGlob,
+        SearchType::ContainsGlob,
         smartcase_flags(search_string),
         history_index,
     );
-    let mut next_match_found = search.go_to_next_match(direction);
-    if !next_match_found && !parse_util_contains_wildcards(search_string) {
+    if !search.go_to_next_match(direction) && !parse_util_contains_wildcards(search_string) {
         // If there were no matches, and the user is not intending for
         // wildcard search, try again with subsequence search.
         search = HistorySearch::new_with(
             history.clone(),
             search_string.to_owned(),
-            history::SearchType::ContainsSubsequence,
+            SearchType::ContainsSubsequence,
             smartcase_flags(search_string),
             history_index,
         );
-        next_match_found = search.go_to_next_match(direction);
+        search.go_to_next_match(direction);
     }
+    // When searching, first we need to find the element before first shown.
+    search.search_forward(match direction {
+        SearchDirection::Forward => page_size,
+        SearchDirection::Backward => 0,
+    });
+    let first_index = search.current_index();
+    let mut next_match_found = search.go_to_next_match(SearchDirection::Backward);
     while completions.len() < page_size && next_match_found {
         let item = search.current_item();
         completions.push(Completion::new(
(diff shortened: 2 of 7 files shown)


krobelus:
Awesome!

> Should I commit them to the main repo and maybe add a mention of them to CONTRIBUTING.rst, or is there a better place for this?

yes please! I'd probably put it into `build_tools/` somewhere.
(FWIW at least gdb pretty printers can be embedded into the binary directly, that would be a goal)

krobelus:
Couple things I observed (I don't have much time right now, I'll try to write tests for your scenarios).

Sorry for the bad/tricky code. Maybe we can extract the off-by-one fix to a separate commit?

1. One minor issue: pressing `ctrl-r` `ctrl-r` `ctrl-s` should not show the
   `Search again for more results` banner. Here's my attempted fix:

<details>

```diff --git a/src/history.rs b/src/history.rs
index 357fed5008..e26dbd4b1d 100644
--- a/src/history.rs
+++ b/src/history.rs
@@ -1932,3 +1932,3 @@ impl HistorySearch {
     /// Move current index so there is `value` matches in between new and old indexes
-    pub fn search_forward(&mut self, value: usize) {
+    pub fn search_forward(&mut self, value: usize) -> bool {
         let mut next_match_found = self.go_to_next_match(SearchDirection::Forward);
@@ -1941,3 +1941,4 @@ impl HistorySearch {
             self.current_item = None;
-        }
+        };
+        next_match_found
     }
diff --git a/src/reader.rs b/src/reader.rs
index 19d31007d5..8fb3f071dd 100644
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -5098,6 +5098,8 @@ fn history_pager_search(
     let mut final_index = None;
+    let mut have_more_results = false;
     if direction == SearchDirection::Forward {
-        search.search_forward(page_size);
+        have_more_results = search.search_forward(page_size);
         final_index = Some(search.current_index());
-        next_match_found = search.go_to_next_match(SearchDirection::Backward);
+        let ok = search.go_to_next_match(SearchDirection::Backward);
+        assert!(ok);
     }
@@ -5113,2 +5115,5 @@ fn history_pager_search(
     }
+    if direction == SearchDirection::Backward {
+        have_more_results = next_match_found;
+    }
     HistoryPagerResult {
@@ -5116,3 +5121,3 @@ fn history_pager_search(
         final_index: final_index.unwrap_or(search.current_index()),
-        have_more_results: next_match_found,
+        have_more_results,
     }
```

</details>

2. `shift-delete` (to delete the current search match)
   should move to the next older item, not the next younger one.
   It seems like your patch breaks this when on the second (or higher) page of search results.

3. (Possibly related to 2) I hit a crash (with my change on top, which should hopefully be unrelated).

The comment next to the assertion says
`// Current users are off by one at most.`
Maybe that has been invalidated.

Unfortunately I forgot to attach GDB and I rebuilt so the debug info in my core dump seems no
longer useful. But I suspect we may be off by two here because of 2.
Could also be entirely unrelated.

<details>

```
search: thread 'main' panicked at src/pager.rs:866:9:
assertion failed: new_index.is_none_or(|new_index| new_index <= self.completion_infos.len())
stack backtrace:fish
  17:     0x55d05490acf3 - core::panicking::panic_fmt::h3eea515d05f7a35e
                               at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/core/src/panicking.rs:74:14
  18:     0x55d05490ad7c - core::panicking::panic::h102d65dbfa674afe
                               at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/core/src/panicking.rs:148:5
  19:     0x55d054d33945 - fish::pager::Pager::set_selected_completion_index::h886804b1bb28f0c3
                               at /home/johannes/git/fish-shell/src/pager.rs:866:9
  20:     0x55d054dddb55 - fish::reader::ReaderData::fill_history_pager::{{closure}}::h1e77e649f8c442e8
                               at /home/johannes/git/fish-shell/src/reader.rs:5190:17
  21:     0x55d054b71379 - core::ops::function::FnOnce::call_once::hc9d9808eceec96b6
                               at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/core/src/ops/function.rs:250:5
  22:     0x55d054cea7b7 - fish::threads::Debounce::perform_with_completion::{{closure}}::{{closure}}::h237fe0c8947049ee
                               at /home/johannes/git/fish-shell/src/threads.rs:632:17
  23:     0x55d054b6daf9 - core::ops::function::FnOnce::call_once{{vtable.shim}}::hd09ba391fa1aeeb9
                               at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/core/src/ops/function.rs:250:5
  24:     0x55d054d1b000 - <alloc::boxed::Box<F,A> as core::ops::function::FnOnce<Args>>::call_once::h7efbf4c335489998
                               at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/alloc/src/boxed.rs:2231:9
  25:     0x55d054ac9fb7 - fish::threads::iothread_service_main::h2e77c7b57d5e69b6
                               at /home/johannes/git/fish-shell/src/threads.rs:517:9
  26:     0x55d054d6128b - fish::input::<impl fish::input_common::InputEventQueuer for fish::reader::Reader>::ioport_notified::habae12fd69d2027c
                               at /home/johannes/git/fish-shell/src/input.rs:442:9
  27:     0x55d054dd087b - fish::input_common::InputEventQueuer::try_readch::hea496b58ed81c58f
                               at /home/johannes/git/fish-shell/src/input_common.rs:721:21
  28:     0x55d054d61dc0 - fish::input_common::InputEventQueuer::readch::h5d2f04f341d26a90
                               at /home/johannes/git/fish-shell/src/input_common.rs:676:15
  29:     0x55d054d61dc0 - fish::input::<impl fish::reader::Reader>::read_char::h88a85534aafd321b
                               at /home/johannes/git/fish-shell/src/input.rs:766:23
  30:     0x55d054d7383d - fish::reader::Reader::read_normal_chars::hb4dffb99ff695497
                               at /home/johannes/git/fish-shell/src/reader.rs:2244:23
  31:     0x55d054dda4d4 - fish::reader::Reader::handle_char_event::{{closure}}::hc442b6ccf25f01fa
                               at /home/johannes/git/fish-shell/src/reader.rs:2319:42
  32:     0x55d0549440a6 - core::option::Option<T>::or_else::h7214d615246cdfb8
                               at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/core/src/option.rs:1545:21
  33:     0x55d054d74087 - fish::reader::Reader::handle_char_event::hf3bea4bcf459ca68
                               at /home/johannes/git/fish-shell/src/reader.rs:2318:38
  34:     0x55d054d72adc - fish::reader::Reader::readline::hfd31a2904b251cb7
                               at /home/johannes/git/fish-shell/src/reader.rs:2136:16
  35:     0x55d054d65bb7 - fish::reader::read_i::h4bbe74d5da0aaf31
                               at /home/johannes/git/fish-shell/src/reader.rs:658:29
  36:     0x55d054d653de - fish::reader::reader_read::hf01179162bffe4f5
                               at /home/johannes/git/fish-shell/src/reader.rs:621:9
  37:     0x55d0549131d1 - fish::throwing_main::hfe1ea241e904d3af
                               at /home/johannes/git/fish-shell/src/bin/fish.rs:915:15
```

</details>

krobelus:
Here's a possible fix; if you use them, feel free to squash.

This makes the search return a proper range rather than reusing `index`.
This might be the most convenient way.

I tested with

```fish
set fish_history tmp
echo yes | history --clear
for x in (seq 50)
    tmux send-keys echo Space $x Enter
end
```

<details>

```
From 4cd9d4cf5b890e039b6f2257e866e568e5e8d18c Mon Sep 17 00:00:00 2001
From: Johannes Altmanninger <aclopte@gmail.com>
Date: Thu, 9 Jan 2025 21:57:06 +0100
Subject: [PATCH] Make shift-delete always move backward

---
 src/reader.rs | 40 +++++++++++++++++++++++++---------------
 1 file changed, 25 insertions(+), 15 deletions(-)

index 8fb3f071dd..fec9254dcd 100644
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -2672,7 +2672,7 @@ impl<'a> Reader<'a> {
             }
             rl::PagerToggleSearch => {
                 if let Some(history_pager) = &self.history_pager {
-                    if history_pager.history_index_start == 0 {
+                    if history_pager.history_index_start <= 1 {
                         self.flash();
                         return;
                     }
@@ -5035,7 +5035,7 @@ impl<'a> Reader<'a> {
 
 struct HistoryPagerResult {
     matched_commands: Vec<Completion>,
-    final_index: usize,
+    range: Range<usize>,
     have_more_results: bool,
 }
 
@@ -5058,7 +5058,7 @@ struct HistoryPager {
 fn history_pager_search(
     history: &Arc<History>,
     direction: SearchDirection,
-    history_index: usize,
+    mut history_index: usize,
     search_string: &wstr,
 ) -> HistoryPagerResult {
     // Limit the number of elements to half the screen like we do for completions
@@ -5068,10 +5068,10 @@ fn history_pager_search(
     // but that can't really be helped.
     // (subtract 2 for the search line and the prompt)
     let page_size = usize::try_from(cmp::max(termsize_last().height / 2 - 2, 12)).unwrap();
-    let history_index = match direction {
-        SearchDirection::Forward => history_index + 1,
-        SearchDirection::Backward => history_index.saturating_sub(1),
-    };
+    if direction == SearchDirection::Backward {
+        // The given index should be the first match.
+        history_index = history_index.saturating_sub(1);
+    }
     let mut completions = Vec::with_capacity(page_size);
     let mut search = HistorySearch::new_with(
         history.clone(),
@@ -5093,15 +5093,20 @@ fn history_pager_search(
         );
         next_match_found = search.go_to_next_match(direction);
     }
+    let first_index = if next_match_found {
+        search.current_index()
+    } else {
+        0
+    };
     // When searching forwards, first we need to find the element before first shown,
     // and only then compose the completions list to maintain order after deduplication.
-    let mut final_index = None;
     let mut have_more_results = false;
+    let mut last_index = 0;
     if direction == SearchDirection::Forward {
         have_more_results = search.search_forward(page_size);
-        final_index = Some(search.current_index());
         let ok = search.go_to_next_match(SearchDirection::Backward);
         assert!(ok);
+        last_index = search.current_index();
     }
     while completions.len() < page_size && next_match_found {
         let item = search.current_item();
@@ -5115,10 +5120,15 @@ fn history_pager_search(
     }
     if direction == SearchDirection::Backward {
         have_more_results = next_match_found;
+        last_index = search.current_index() + if next_match_found { 0 } else { 1 };
     }
     HistoryPagerResult {
         matched_commands: completions,
-        final_index: final_index.unwrap_or(search.current_index()),
+        range: if direction == SearchDirection::Forward {
+            last_index..first_index + 1
+        } else {
+            first_index..last_index
+        },
         have_more_results,
     }
 }
@@ -5146,7 +5156,7 @@ impl ReaderData {
             HistoryPagerInvocation::Refresh => {
                 // Redo the previous search previous direction.
                 let history_pager = self.history_pager.as_ref().unwrap();
-                direction = history_pager.direction;
+                direction = SearchDirection::Backward;
                 index = history_pager.history_index_start;
                 old_pager_index = Some(self.pager.selected_completion_index());
             }
@@ -5169,13 +5179,13 @@ impl ReaderData {
             history_pager.direction = direction;
             match direction {
                 SearchDirection::Forward => {
-                    history_pager.history_index_start = result.final_index;
-                    history_pager.history_index_end = index + 1;
+                    history_pager.history_index_start = result.range.start;
+                    history_pager.history_index_end = result.range.end;
                     history_pager.can_go_backwards = true;
                 }
                 SearchDirection::Backward => {
-                    history_pager.history_index_start = index.saturating_sub(1);
-                    history_pager.history_index_end = result.final_index;
+                    history_pager.history_index_start = result.range.start;
+                    history_pager.history_index_end = result.range.end;
                     history_pager.can_go_backwards = result.have_more_results;
                 }
             };
-- 
2.47.1
```

</details>

kerty0:
Thanks for help!

>1.  One minor issue: pressing ctrl-r ctrl-r ctrl-s should not show the
Search again for more results banner.

I think it would be better if banner was changed to something similar to completions banner:
`Items 1 to 12 of 500`.

>This makes the search return a proper range rather than reusing index.
This might be the most convenient way.

krobelus:
On Fri, Jan 10, 2025 at 01:27:25AM -0800, kerty0 wrote:
> Thanks for help!
> 
> >1.  One minor issue: pressing ctrl-r ctrl-r ctrl-s should not show the
> Search again for more results banner.
> 
> I think it would be better if banner was changed to something similar to completions banner:
> `Items 1 to 12 of 500`.

Sounds good, at least for the `ctrl-s` case.

After `ctrl-r` we often don't know how many results there are left.
We could exhaust the search (breaking with traditional behavior),
or simply add an ellipsis when there are more items available.

	Items 1 to 12 of 12…

We already use the ellipsis for autosuggestions and truncated lines etc.

> 
> >This makes the search return a proper range rather than reusing index.
> This might be the most convenient way.

Yeah it's nice if start always points to a valid item, and end always points to one-past-last.

Well, not exactly "always" because the user can run `history merge` and `history save` from bindings while the history search is active.
So we cannot be sure that these indexes are still valid, unless we just found them and didn't yet return control back to the event loop (which should be the case for the assertion I added)

Given that the history search runs in a background thread, we might get stale or inconsistent results when racing with the main thread reloading the history file.
But I think we're always checking the validity of indexes at some point while holding the history lock, so there shouldn't be too many crashes.

krobelus:
> Maybe change HistoryPager to range too.

sure

kerty0:
> After `ctrl-r` we often don't know how many results there are left.

Are you sure? I think the history file is always loaded completely, so `zelf.history.size()` should work fine.

kerty0:
```rust
zelf.pager.extra_progress_text =
    if (1..zelf.history.size() + 1) != history_pager.history_range {
        wgettext_fmt!(
            "items %lu to %lu of %lu",
            history_pager.range.start,
            history_pager.range.end - 1,
            zelf.history.size()
        )
    } else {
        L!("").to_owned()
    };
```
How about this? It would show that piece of full history pager shows and hide banner if pager covers full history. It easy to implement and shows more info than current banner.

krobelus:
I like the direction. It seems unobtrusive; if it's more or less clear that it counts all items and not only the displayed ones, this should work.
One odd bit might be that `history_pager.range.end - 1,` will basically never reach `zelf.history.size()`,
so it might less obvious of an indication that we've exhausted the search space.
I think that info may be more important than the precise indexes; I need to try it.

FWIW the history pager might enable us to find a UI for moving to
the next/prev history item relative to the one we found via search (https://github.com/fish-shell/fish-shell/issues/1671).
I haven't found intuitive key bindings for that unfortunately.
Clearing the search field will restart the search from the beginning. Maybe it shouldn't, I'm not sure

kerty0:
I hope I got it right this time...

Main changes:
1. Banner now should show index of first element of page (if there exists elements before it, otherwise 1) and index of first element of next page - 1 (or history.size() if there no next pages).
2. When you delete last element on last page it will search forward automatically.

src/reader.rs
```diff
@@ -5046,12 +5030,22 @@ fn history_pager_search(
...
         );
         next_match_found = search.go_to_next_match(direction);
     }
+    // When searching forwards, first we need to find the element before first shown,
+    // and only then compose the completions list to maintain order after deduplication.
+    let mut first_index = 0;
+    if direction == SearchDirection::Forward {
+        search.search_forward(page_size);
+        first_index = search.current_index();
+        next_match_found = search.go_to_next_match(SearchDirection::Backward);
```
krobelus, line 5045: Great work overall.
This is quite tricky; to de-risk we could create tests and split commits.
But it seems fine as-is.

FWIW I figured out a better way to manually test it:

	build/fish -C 'set fish_history my; echo yes | history clear; for i in (seq 50); history append "echo $i"; end'

src/reader.rs
```diff
@@ -5116,29 +5108,29 @@
...
-            };
-            zelf.pager.extra_progress_text = if result.have_more_results {
-                wgettext!("Search again for more results")
-            } else {
-                L!("")
+
+            if result.matched_commands.is_empty() && result.range != (0..history_len + 1) {
+                history_pager.start = history_len;
+                zelf.fill_history_pager(HistoryPagerInvocation::Advance, SearchDirection::Forward);
+                return;
```
krobelus, line 5117: This "recursive" call is sus;
It's for the the scenario where we delete the last element on the last page, right.
It might be better to do this only for that invocation (in HistoryPagerDelete),
and maybe earlier, inside the performer.

I guess it's at odds with the current approach  to `HistoryPagerInvocation::Refresh` when at the last page.
Each deletion makes the page smaller by one. Maybe we should change that so the page size always stays the same (until there are not enough items in history left)

But I think it's also okay to leave it like this for now, and maybe change it later.

src/reader.rs
```diff
@@ -5116,29 +5108,29 @@
...
             }
-            .to_owned();
+
+            assert!(result.range.start < result.range.end);
+            assert!(result.range.end <= history_len + 1);
```
krobelus, line 5121: I don't think this assertion holds, because the history file may have been modified and reloaded after search results were computed but before this callback was invoked.
I think we should clamp the end to the actual history_len.

(the above recursive call to `fill_history_pager()` seems to protect us from some cases but not all)

src/reader.rs
```diff
@@ -5116,29 +5108,29 @@
...
+
+            *history_pager = result.range;
+            zelf.pager.extra_progress_text = if *history_pager != (0..history_len + 1) {
+                wgettext_fmt!(
+                    "Items %lu to %lu of %lu",
+                    history_pager.start + 1,
+                    history_pager.end - 1,
+                    history_len
+                )
```
krobelus, line 5130: > Banner now should show index of first element of page (if there exists elements before it, otherwise 1) and index of first element of next page - 1 (or history.size() if there no next pages).

Looks good.

krobelus, line 5130: > Only thing I want is maybe add new SearchDirection::ForwardPrev that would use zelf.select_completion_in_direction(SelectionMotion::Prev, true); instead of zelf.select_completion_in_direction(SelectionMotion::Next, true); like Forward and Backward does in order to change №2 to be more smooth.

which action/key is this about? (I'm not sure what `№2` is)

src/reader.rs
kerty0: > I don't think this assertion holds, because the history file may have been modified and reloaded after search results were computed but before this callback was invoked.

Will delete it.

> I think we should clamp the end to the actual history_len.

`history.size()` return size of history without zero element, so history[history.size()] is valid history element while history[history.size() + 1] is not.

src/reader.rs
kerty0: Sorry, my explanation was unclear. 
When deleting last element on last page pager will show previous page but select first element. I want to add new `SearchDirection` so it would select last:
```rust 
if result.matched_commands.is_empty() && result.range != (0..history_size + 1) {
    history_pager.start = history_size;
    zelf.fill_history_pager(HistoryPagerInvocation::Advance, SearchDirection::ForwardPrev);
    return;
}
```
and 
```rust
if why == HistoryPagerInvocation::Refresh {
    zelf.pager
        .set_selected_completion_index(old_pager_index.unwrap());
    zelf.pager_selection_changed();
} else if direction == SearchDirection::ForwardPrev {
    zelf.select_completion_in_direction(SelectionMotion::Prev, true);
} else {
    zelf.select_completion_in_direction(SelectionMotion::Next, true);
}
```

src/reader.rs
kerty0: > This is quite tricky; to de-risk we could create tests and split commits.

I'm working on splitting commits, but to be honest I've never written tests before.

krobelus: Don't sweat it, but if you're interested, b7f7dcf7881a24d57dd65e335f261e31bec7b795 is an example test

src/reader.rs
krobelus: yes, that would be a great change. Good eye

kerty0:
> Don't sweat it, but if you're interested, https://github.com/fish-shell/fish-shell/commit/b7f7dcf7881a24d57dd65e335f261e31bec7b795 is an example test

I will try to write some next week.

> Clearing the search field will restart the search from the beginning. Maybe it shouldn't, I'm not sure

 I tried to make the search start from the selected item in https://github.com/kerty0/fish-shell/commit/aeeb8a18666a8f6234bbf5c502cd08d16856baeb, but it requires more polish.
 
 > Given that the history search runs in a background thread, we might get stale or inconsistent results when racing with the main thread reloading the history file.
 
 I also tried to make `history_pager_search` maintain history lock in https://github.com/kerty0/fish-shell/commit/9d75d680b4389c776f90b4faf7423402a1f80de9, but I don’t know how well I did and whether there are any pitfalls with this.

src/history.rs
```diff
@@ -1905,6 +1905,11 @@ impl HistorySearch {
 
             // We're done if it's empty or we cancelled.
             let Some(item) = self.history.item_at_index(index) else {
+                self.current_index = match direction {
+                    SearchDirection::Backward => self.history.size() + 1,
+                    SearchDirection::Forward => 0,
+                };
+                self.current_item = None;
```
krobelus, line 1913: from this commit alone, it's not obvious why this is needed.
It's not necessary for normal up-arrow history search, only for the pager (I'm not yet sure why?). Might be worth mentioning that in the commit message, if this stays  a separate commit.
Or maybe it's better if we track this in the caller, since that one already needs to handle the `bool` return value, no?
Not sure if that's equivalent.

Still haven't re-read the rest yet

kerty0: Sorry, you're right. I will combine this commit with [Fix history pager inconsistent ordering and skipping 13th elements](https://github.com/fish-shell/fish-shell/pull/11021/commits/a170b9be8d5c77493483116c3cb18ea038296317).

> Or maybe it's better if we track this in the caller, since that one already needs to handle the bool return value, no?
Not sure if that's equivalent.

Yes, that equivalent. I added this just to eliminate 2 very similar checks in `history_pager_search` and 1 in `HistorySearch::search_forward`.

kerty0:
I'm not sure if I should disable the new `tmux-history-pager.fish` test in github actions. A few times I ran it, it was successful, but the other disabled tests were successful as well.

faho:
Merged as 059e7424c18d9828c1a5d59f63e8ded8b4e53b93..f26ebac8dd0d97b9f71ef30461bed79b6529d65c

krobelus:
> Given that the history search runs in a background thread, we might get stale or inconsistent results when racing with the main thread reloading the history file.
> 
> I also tried to make history_pager_search maintain history lock in kerty0@9d75d68, but I don’t know how well I did and whether there are any pitfalls with this.

From a quick glance, this seems reasonable. The history pager was written without caring about locking at all.

I think it will be an improvement to hold the lock for the entire duration of the search.
I don't have an example where we currently get inconsistent results but it must be possible.

There seems to be no risk of deadlock.

---

Currently if you create a really long commandline, history search (up-arrow or history pager) will take a long time.

We should maybe allow cancelling that with `ctrl-c` (`OperationContext` is the thing to use).
But I'm not sure if that's an improvement by itself, because with long commandlines, other things can take a long time too.

PR #11028, opened 2025-01-08
Rewrite test driver in python

faho:
## Description

This replaces the test_driver.sh/test.fish/interactive.fish system with a test driver written in python that calls into littlecheck directly and runs pexpect in a subprocess.

This means we reduce the reliance on the fish that we're testing, and we remove a posix sh script that is a weird stumbling block (see my recent quest to make it work on directories with spaces).

To run specific tests, e.g. all the tmux tests and bind.py:

```fish
tests/test_driver.py target/release/ tests/checks/tmux*.fish tests/pexpects/bind.py
```

To run all tests, leave off the file argument or use `tests/checks/*.fish tests/pexpects/*.py`. I.e. you just pass the filenames.

That means, to build fish and run the entire test suite:

```fish
cargo build
cargo test
tests/test_driver.py target/debug/
```

This will run the fish from the given directory (and colocated fish_indent and fish_key_reader) in a temporary directory and environment, once per test. The temporary directory stays between runs.

Note that we do still require /bin/sh for littlecheck - `#RUN` lines and `#REQUIRE` lines are run with it.

## TODOs:

- littlecheck prints output on its own, which means on failure its output comes between the test name and the "FAIL (x ms)" line
- ~~fish_test_helper is compiled once and left in $PWD. That's unclean, but without the time for cmake tests (which run the test_driver once per test) goes up a bit. We could have it reuse an existing fish_test_helper and compile it to the temp directory otherwise~~
- ~~histfile.py is a weird test that passes $XDG_DATA_HOME as a string. This would break if it includes metacharacters (like space). Tho we do get it from a temporary directory.~~
- ~~I removed filter-control-sequences because the littlecheck tests succeed without, and the script slots in weirdly here. If we want to add it, it should be a filter given to littlecheck~~

Note that I have not tested with older python versions, especially not python 2.


faho:
Hmm, if you run these in random order some state carries over.

We should probably clean up the test directory between tests.

mqudsi:
> The temporary directory stays between runs.

Can I ask why? Running each test in its own directory was one of the big improvements we did.

faho:
It was easier to do and I wasn't sure it mattered - we have a bunch of `mktemp` in the littlecheck scripts anyway.

We should probably remove those.

faho:
Okay, every test now gets a new temp dir.

According to https://github.com/fish-shell/fish-shell/pull/11030#issuecomment-2579448843, filter-control-sequences isn't useful anymore, so I'm crossing that off.

PR #11030, opened 2025-01-08
Bump CTEST_PARALLEL_LEVEL to match CTEST_BUILD_PARALLEL_LEVEL

### LINKED ISSUE #11006: CI: run tests in parallel
In https://github.com/fish-shell/fish-shell/blob/7eb254f2bac0be4d2575ae085c44de0473fcb434/.github/workflows/main.yml#L6 the tests in CI are explicitly set to run serially. This was set up in https://github.com/fish-shell/fish-shell/pull/7851.

Is there a reason not to try to run them in parallel? I understand that the tests may still have some global state that would cause them to flake when run in parallel, but maybe that can be addressed as part of this work.

mqudsi:
This is an experiment to see if it causes any of the tests to flake and/or if it even appreciably speeds up CI to begin with.

I note that there are tests added in 8bf8b10 that mutate global terminal state but also note that local tests without CTEST_PARALLEL_LEVEL set at all have been running to completion just fine without any observed flakiness *and* that our Cirrus CI tests have this hard-coded to 6.

Closes #11006

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -3,7 +3,7 @@ name: make test
 on: [push, pull_request]
 
 env:
-  CTEST_PARALLEL_LEVEL: "1"
+  CTEST_PARALLEL_LEVEL: "4"
   CMAKE_BUILD_PARALLEL_LEVEL: "4"
 
 permissions:


mqudsi:
Test time for (macos, ubuntu, ubuntu-32, asan) went from (4, 6, 9, 9) minutes to (2, 4, 4, 5) minutes 👍

faho:
I would also add https://github.com/fish-shell/fish-shell/commit/8bf8b10f685d964101f491b9cc3da04117a308b4#diff-9743e38c3f157c2a7cc13cddf318b6b6e3740b63a07f863f8838d5f92c3a2be4L4-L36 back.

I've just tried running that for an hour and a half and had no failure (with 8304fd0fd088a9d0c738e8c55f272471fe863e8c) - and it doesn't affect CI since that overrides the parallel level.

mqudsi:
Oh, I didn't realize that was yanked out altogether. Retrying with it now.

krobelus:
The terminal state in question is modified by actual interactive shells.

I think the reason why the parallelism never used to be a problem
is that the "interactive shells" in our tests/checks/*.fish --
see `git grep %filter-control-sequences` --
are only pretend-interactive;
they still use `read_ni()` rather that `read_i()`.

`%filter-control-sequences` and serialization seem no longer necessary,
since the criteria for modifying terminal state is "is stdin a TTY" now rather than "-i".

faho:
Let's just merge this and see how it works out.

PR #11034, opened 2025-01-09
Fix missing of builtin token description

phanen:
## Description

With default config, press alt-w on a builtin function, e.g.
```sh
builtin
```

It will show `NAME` only rather than a brief description for the `builtin`.


faho:
Thanks, cherry-picked to 4.0 as fbf9ac8046bd9059c5d706f2847398677752b1c9

Issue #11002, opened 2025-01-04
DNF5 completions break with sqlite3 installed


AldanTanneo:
**OS:** Linux, Fedora 41
**Fish version:** 3.7.1, 4.0b1

DNF5 does not create or update the `/var/cache/dnf/packages.db` sqlite file anymore. When `sqlite3` is installed on the system, the current DNF completion for the `install` tries to access it, then fails and falls back to displaying file names instead of available packages.

Additionally, DNF5 added a `--complete` option for autocompletion, which is slightly faster than the existing `dnf repoquery --cacheonly "$tok*" --qf "%{name}\n" --available`

(example of what I used in my local completions to fix this: `dnf --complete=2 dnf install "$tok*" </dev/null 2>/dev/null`).

Detecting DNF5 could be done with `readlink`, like in the [bash completions](https://github.com/rpm-software-management/dnf5/blob/d4c74147df7764bf3f80facea8ece9907e7618a4/dnf5/bash-completion/dnf5#L22), and then pick the strategy appropriately?

(Follow up to #9862)

mqudsi:
@AldanTanneo do you mind opening a PR with your proposed change?

PR #11035, opened 2025-01-09
DNF5 completion support

AldanTanneo:
## Description

DNF5 does not create or update the `/var/cache/dnf/packages.db` sqlite file anymore. When sqlite3 is installed on the system, the current DNF completion for the install tries to access it, then fails and falls back to displaying file names instead of available packages.

Additionally, DNF5 added a `--complete` option for autocompletion, which is slightly faster than the existing `dnf repoquery`

Fixes issue #11002

### DIFF
--- a/share/completions/dnf.fish
+++ b/share/completions/dnf.fish
@@ -2,6 +2,11 @@
 # Completions for the dnf command
 #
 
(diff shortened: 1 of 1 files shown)


faho:
Thanks!

PR #11045, opened 2025-01-14
Improve mouse support

kerty0:
## Description
Currently, using a mouse leads to crashes when clicking on the pager, autosuggestions, multi-line prompts, or the command line when the pager search field is open. Additionally, clicking past the command line moves the cursor to the start of the command if it is a single line.

This pull request fixes these crashes, makes cursor position changes consistent when clicking and adds the ability to select pager elements by clicking.

Other fixes:
- https://github.com/fish-shell/fish-shell/commit/2ba662be8c3f4b811aacd565f9632583c8468a30 fixes the issue where the screen height could exceed the terminal height when using a multi-line prompt. This would result in KiTTY no longer reporting mouse clicks.

- https://github.com/fish-shell/fish-shell/commit/25f242cc11a252d52aaf3a813970217cdd513d7b fixes unsaved screen modifications that would trigger a prompt redraw when clicking with a mouse or calling scrollback-push. This would led to duplication of prompt lines in multi-line prompts.

### DIFF
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -1392,6 +1392,7 @@ impl ReaderData {
         assert!(self.cursor_position_wait == CursorPositionWait::None);
         self.cursor_position_wait = cursor_position_wait;
         let _ = out.write(b"\x1b[6n");
+        self.save_screen_state();
     }
 
     pub fn mouse_left_click(&mut self, cursor: ViewportPosition, click_position: ViewportPosition) {
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -120,7 +120,7 @@ use crate::proc::{
     print_exit_warning_for_jobs, proc_update_jiffies,
 };
 use crate::reader_history_search::{smartcase_flags, ReaderHistorySearch, SearchMode};
-use crate::screen::{screen_clear, screen_force_clear_to_end, Screen};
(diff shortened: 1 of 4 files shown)


src/reader.rs
```diff
@@ -1392,6 +1392,7 @@ impl ReaderData {
         assert!(self.cursor_position_wait == CursorPositionWait::None);
         self.cursor_position_wait = cursor_position_wait;
         let _ = out.write(b"\x1b[6n");
+        self.save_screen_state();
```
krobelus, line 1395: out of curiousity, how did you trigger this bug? kitty on macos? (IIRC all the stdout timestamp bugs were macos-specific)

krobelus (approved):

src/reader.rs
kerty0: No, I’m using Kitty on Linux. But I was able to trigger this bug on other terminals (Alacritty and Ghostty) with this:
Set up a multi-line prompt, open the `ctrl+r` history pager, wait ~8 seconds, and then use `ctrl+l`.
This sequence causes the prompt lines (except the last one) to duplicate.

>IIRC all the stdout timestamp bugs were macos-specific

I observed that the timestamp incremented by a minimum of 8 seconds, so maybe on macOS it happens more often, and the timestamp bugs are more visible?

PR #11046, opened 2025-01-15
completions/jj: use dynamic completions by default, also fix them

ilyagr:
This uses jj's dynamic completions when possible.

This avoids an annoying problem. Because of 04a4e5c4 (so, in fish 4.0beta, but not in fish 3.7.1), jj's dynamic completions (added to jj recently, see the second paragraph of <https://jj-vcs.github.io/jj/latest/install-and-setup/#command-line-completion>) do not work very well in fish if the user puts `COMPLETE=fish jj | source` in their `~/.config/fish/config.fish`. When the user types `jj <TAB>`, they are instead overridden by fish's built-in non-dynamic completions.

The difference is subtle. One problem I saw is that `jj new <TAB>` works as expected (and shows revisions) while `jj new -A <TAB>` becomes broken (and shows files).

If the user puts `COMPLETE=fish jj | source` in `~/.config/fish/completions/jj.fish` there is no problem. However, users might be confused if they run `COMPLETE=fish jj | source` or put it in their config and it works in a broken fashion. I certainly was, this took a while to debug.

Meanwhile, I checked that if the user has `jj completion fish | source` in their `config.fish`, executing `COMPLETE=fish jj __this_command_does_not_exist | source` afterwards still seems to work correctly. So, this PR should not cause the opposite
problem to the one it tries to solve.

Let me know if there's a better approach to this problem.

## TODOs:

### DIFF
--- a/share/completions/jj.fish
+++ b/share/completions/jj.fish
@@ -1,1 +1,10 @@
-jj util completion fish | source
+# The reason for `__this-command-does-not-exist` is that, if dynamic completion
+# is not implemented, we'd like to get an error reliably. However, the
+# behavior of `jj` without arguments depends on the value of a config, see
+# https://jj-vcs.github.io/jj/latest/config/#default-command
+if set -l completion (COMPLETE=fish jj __this-command-does-not-exist 2>/dev/null)
+    # jj is new enough for dynamic completions to be implemented
+    printf %s\n $completion | source
+else
+    jj util completion fish | source
+end


ilyagr:
There are two other options that come to mind. I'll mark this as a draft until I figure out how viable option 2 is. I'd love to resolve this between fish 4 is released (if possible, of course).

The two other options I see are:

1. Remove `jj` completions from `fish` for now
2. Change `jj` so that `jj util completion fish | source` turns on dynamic completions. This may or may not require changes to the CLI parsing library that generates the completions (`clap`). I *think* the dynamic completions are good enough that this is otherwise possible, though they might still have bugs I don't know about.

ilyagr:
After reading https://github.com/clap-rs/clap/issues/5668, it seems like Clap would not recommend using the same command for static and dynamic completions, since they do not want distributions to package the output of `COMPLETE=fish jj` in the way they package the output of static completions (e.g. `jj util completion fish`). Also, AFAICT, this would require a change to `clap` (a small one, technically, but I now doubt it would be accepted).

So, I'm back to thinking that this PR is the best we can do. If it has a good chance to be accepted, I'll try adding a test to `jj` that `COMPLETE=fish jj __this-command-does-not-exist` does print dynamic completions in the same way as `COMPLETE=fish jj`.

share/completions/jj.fish
```diff
@@ -1 +1,10 @@
-jj util completion fish | source
+# The reason for `__this-command-does-not-exist` is that, if dynamic completion
+# is not implemented, we'd like to get an error reliably. However, the
+# behavior of `jj` without arguments depends on the value of a config, see
+# https://jj-vcs.github.io/jj/latest/config/#default-command
+if set -l completion (COMPLETE=fish jj __this-command-does-not-exist 2>/dev/null)
+    # jj is new enough for dynamic completions to be implemented
+    echo "$completion" | source
+else
+    jj completion fish | source
```
krobelus, line 9: changing from `jj util completion fish` to `jj completion fish` (which is no longer suported in current `jj`) seems odd.
I guess you know that this works for all relevant versions?

share/completions/jj.fish
```diff
@@ -1 +1,10 @@
-jj util completion fish | source
+# The reason for `__this-command-does-not-exist` is that, if dynamic completion
+# is not implemented, we'd like to get an error reliably. However, the
+# behavior of `jj` without arguments depends on the value of a config, see
+# https://jj-vcs.github.io/jj/latest/config/#default-command
+if set -l completion (COMPLETE=fish jj __this-command-does-not-exist 2>/dev/null)
+    # jj is new enough for dynamic completions to be implemented
+    echo "$completion" | source
```
krobelus, line 7: sure this seems like the right solution -- assuming "dynamic" completions are a superset of "static" completions.

We'll take whatever is a stable way to get all completions. Looks fine as is.

Keeping it seems valuable since otherwise adding completions requires manual steps for the user.
(Either way we need to make sure that `COMPLETE=fish jj __this-command-does-not-exist` always outputs safe fish for any relevant fish version.)

I did not yet have a chance to test dynamic completions but I saw from

```
$ COMPLETE=fish jj asdf
complete --keep-order --exclusive --command jj --arguments "(COMPLETE=fish "'jj'" -- (commandline --current-process --tokenize --cut-at-cursor) (commandline --current-token))"
```

That it uses `--tokenize`; this should be updated to `--tokens-expanded` (see https://github.com/fish-shell/fish-shell/pull/10212),
else completions for `jj --repository=$HOME/my-repo ...` will not work -- unless jj is overstepping its boundaries by expanding shell syntax

(not a blocker; it's probably not a regression)

krobelus, line 7: This joins multiple lines by spaces which may break if future `jj` outputs multiple lines.
To reduce surprise and eyebrow-raising, we should use something like `printf %s\n $completion`.
(could also elide the variable with `if ...; end | source`)

share/completions/jj.fish
```diff
@@ -1 +1,10 @@
-jj util completion fish | source
+# The reason for `__this-command-does-not-exist` is that, if dynamic completion
+# is not implemented, we'd like to get an error reliably. However, the
+# behavior of `jj` without arguments depends on the value of a config, see
+# https://jj-vcs.github.io/jj/latest/config/#default-command
+if set -l completion (COMPLETE=fish jj __this-command-does-not-exist 2>/dev/null)
```
krobelus, line 5: `jj` instructions should probably state that it's only necessary for fish below version 4.
Else some users would load duplicate completion definitions (at least Git completions can be slow).
Could make it say something like `if [ -z "$(complete -C'jj ')" ]; COMPLETE=fish jj | source; end`
which does nothing if completions are already installed

share/completions/jj.fish
senekor: This is coming from `clap_complete`. I doubt it can adopt the new option, since it probably wants to stay compatible with older versions of fish. Once clap_complete does decide to adopt this, jj will automatically get it.

senekor:
> Change `jj` so that `jj util completion fish | source` turns on dynamic completions. This may or may not require changes to the CLI parsing library

Why do you think `clap` would need to be changed? We could just have `jj util completion fish` literally output the string `COMPLETE=fish jj | source`, right?

That would also solve the problem that packagers shouldn't include the output of `COMPLETE=fish jj` statically.

senekor:
After a quick look at https://github.com/clap-rs/clap/issues/5668, I think the main issue there is that other shells don't do lazy loading of completions. This is not a problem for fish. We _can_ just drop the literal string `COMPLETE=fish jj | source` in `completions/jj.fish`. IIRC, it will be executed once per shell, at the moment jj is run for the first time in that shell.

share/completions/jj.fish
krobelus: yes it can, with feature testing as suggested in that link:

```fish
set -l tokens
if commandline -x >/dev/null 2>&1
    set tokens (commandline -xpc)
else
    set tokens (commandline -opc)
end
```

I'll send a patch to clap eventually (unless someone else does it first)

share/completions/jj.fish
ilyagr: No, this is a silly mistake. Thanks for catching it!

share/completions/jj.fish
ilyagr: >  assuming "dynamic" completions are a superset of "static" completions.

Yes, they are.

ilyagr: I think I'll go with `printf`. I think that, in practice, `if |source` would work, but I'm not entirely sure whether some of the old versions of fish might print the error from `jj this-command-doesn-not-exist` to stdout.

ilyagr:
I fixed up most of krobelus's comments, thanks for catching my bugs!

> I think the main issue there is that other shells don't do lazy loading of completions.

I agree. For `fish`, it seems like having `jj util completion fish` output `COMPLETE=fish jj | source`, and keeping `jj util completion fish | source` in `completions/jj.fish` would work fine. But for bash, the analogue might be slow.

I'm not sure whether it makes sense for `jj util completion fish` to switch to doing dynamic completions if we need another setup for bash's dynamic completions. Or maybe it does.

Aside: As discussed [on Discord](https://discord.com/channels/968932220549103686/1329253162871164998), nixpkgs already does the analogous (slow) thing for bash and zsh: [nixpkgs](https://github.com/NixOS/nixpkgs/blob/eb62e6aa39ea67e0b8018ba8ea077efe65807dc8/pkgs/by-name/ju/jujutsu/package.nix#L95-L98). I think brew caches the output of `COMPLETE=shell jj` ([1](https://github.com/Homebrew/homebrew-core/blob/92957939a59d0ba058b43da82447fbd6f1097bf1/Formula/j/jj.rb#L36), [2](https://github.com/Homebrew/brew/blob/e30e3d3be8cedd7624dec6cdede6df952b9ed395/Library/Homebrew/formula.rb#L2177) ), which is faster but not recommended by clap.

Another aside: perhaps we could implement feature detection in `jj` so that you could check `jj feature check dynamic-completions-v1` instead of the hack in this PR.

share/completions/jj.fish
```diff
@@ -1 +1,10 @@
...
+    # jj is new enough for dynamic completions to be implemented
+    printf %s\n $completion | source
```
krobelus, line 7: Another issue is that computing completions purely inside `jj` means that
it won't work if there are shell metacharacters. Examples:

- `jj --repository ~/`
- `jj --repository $HOME/`

These don't get any completions. Note that `commandline --tokens-expanded` won't
work because (without a heuristic) fish doesn't know how to apply `/home/user/foo` non-destructively.

The currently recommended approach is to delegate back to fish's file completions if
jj finds that we are completing a file path.
Normally file completions are disabled via `complete --exclusive` but we can do something like

```fish
function __fish_jj_complete
    set -g __fish_jj_completions (COMPLETE=fish jj -- (commandline --current-process --tokenize-expanded --cut-at-cursor) (commandline --current-token))
    test $status -ne 5 # status 5 means "fall back to file completions"
end
complete jj  --keep-order --exclusive --command jj \
    --condition __fish_jj_complete --arguments '(string join \n $__fish_jj_completions)'
```

I guess in this specific case we wouldn't even fallback but instead compute directory completions, using the existing:

```
complete -c jj -n "__fish_jj_needs_command" -s R -l repository -d 'Path to repository to operate on' -r -f -a "(__fish_complete_directories)"
```

This is why hand-written completions are often better - but of course we can fix them (probably needs another change in Clap).

I guess this is not blocking, because using variables in file paths is somewhat rare for jj (given that
most file paths for `jj diff` and others get completed just fine). But we should fix that.

share/completions/jj.fish
krobelus, line 5: Using `jj 0.23.0-0ca6f00421f6f893e974e0f5349e126abd41a812`

Pressing tab on

`$ jj --`


prints a spurious

`Hint: Use `jj -h` for a list of available commands.`

Similar on `jj --repository /`.

share/completions/jj.fish
senekor: > Note that `commandline --tokens-expanded` won't
work because (without a heuristic) fish doesn't know how to apply `/home/user/foo` non-destructively.

I don't understand this, but my fish version doesn't support that option yet, so I can't experiment with it. What does `--tokens-expanded` do exactly? Why doesn't it expand `jj --repository ~/` to `/home/user/` before passing it to clap for dynamic completions?

ilyagr: This is interesting. I think this is far from being specific to `jj`. I don't immediately see a better solution than the commits you linked to this PR.

I think it will only become more of a problem as more CLI libraries support dynamic completion (I think there's already `clap`, https://github.com/spf13/cobra , and some Python libraries whose name I forgot). I hope they will; naively it seems like an ultimately better approach than shipping static completions in shell-and-package-manager dependent ways.

To me, this suggests the need for some more detailed communication protocol between the shell and `clap` or the like. It's hard for me to decide whether `--token-expanded` is a sufficient protocol for this. I'm glad it exists. I haven't followed the details, but I'm guessing that it's an 80% solution, and something better will eventually be needed for the other 20% (which will take 500% as much work).

krobelus: > but I'm guessing that it's an 80% solution

yes absolutely. `--tokens-expanded` was added as an incremental improvement to fix one bad aspect of `--tokens` 

> and something better will eventually be needed for the other 20% (which will take 500% as much work).

`--tokens-expanded` works fine for getting the list of tokens before the token at cursor.
This list can be fed into a completion generator that has no idea of shell syntax.
Some remaining problems are:
1. it breaks on multiline tokens (and tokens that expand to multiple lines) -- should probably add a `--zero` option
2. when expansion would produce too many values, we pass on the unexpanded string. This is hacky but unlikely to cause problems in practice
3. The most difficult one: `--tokens-expanded` only works for tokens to the left, not for the token at cursor.
    To fix this we indeed need some kind of communication protocol for passing arbitrary metadata.
Consider `myapp {foo,` where
the shell would need to tell the completion generator that it should generate something based off `myapp foo`
Also given `myapp $HOME/`, the shell would ask for completions based on the expanded `$HOME`, but would want to un-expand the variable before applying the completion.

I think it's unrealistic that every completion client implements their own fish tokenizer.
All those completions like `ls {foo,` already work in fish, so we could get there by plugging this into external completions.

IME apps providing completion generation are reluctant to use wildly different code for fish than for bash/zsh,
so I'd love to find an interface that can be easily implemented in any shell.
That will also be a soft prerequisite for this thing to be adopted (fish is small enough to have self-interesting in not being evil ;).

share/completions/jj.fish
krobelus: AFAICT the only remaining regression is https://github.com/clap-rs/clap/pull/5949

I'm planning to merge this PR after that clap fix is in jj. Though it may be worth merging before that, not sure

PR #11056, opened 2025-01-16
Add feature flag for turning off keyboard protocols

faho:
To work around terminal bugs.

The flag "keyboard-protocols" defaults to "on" and enables keyboard protocols, but can be turned off by setting "no-keyboard-protocols".

This has downsides as a feature flag - if you use multiple terminals and one of them can't do it you'll have to disable it in all, but anything else would require us to hook this up to env-dispatch, and ensure that we turn the protocols *off* when the flag is disabled.

Since this is a temporary inconvenience, this would be okay to ask zellij and Jetbrains-with-WSL users.

## TODOs:

### DIFF
--- a/doc_src/language.rst
+++ b/doc_src/language.rst
@@ -2024,6 +2024,7 @@ You can see the current list of features via ``status features``::
     ampersand-nobg-in-token on  3.4 & only backgrounds if followed by a separating character
     remove-percent-self     off 4.0 %self is no longer expanded (use $fish_pid)
     test-require-arg        off 4.0 builtin test requires an argument
+    keyboard-protocols      on  4.0 Use keyboard protocols (kitty, xterm's modifyotherkeys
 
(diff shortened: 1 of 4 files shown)


zanchey:
My only concern is that if the broken terminals get fixed, how do users know to turn the flag back off?

faho:
I guess they'll have to try regularly?

That's fundamental to anything that allows you to turn it off - even if this were a shell variable you'd have to remove it again if your terminal is fixed.

Tho at a certain point we'll phase this flag out again (as all feature flags should be), which would re-enable it.

zanchey:
OK, I think it's fine to include this as an escape hatch for 4.0.

krobelus:
For 4.0 sure but as of today it's not needed in master.
It's a shame we can't use `fish | sed ...` to filter out problematic sequences.

Issue #10903, opened 2024-12-13
no autocompletion for `git subtree`


assertnotnull:
In version 3.7.1 the git completion is missing the support for `git subtree`

faho:
PRs welcome.

nav1s:
I saw that there was a PR to address this issue but for some reason I still don't completions for ```git subtree```

### version
fish, version 4.1.2
### Operating System:
Archlinux

Edit: it does seem to work when I run ```source /usr/share/fish/completions/git-subtree.fish```

krobelus:
> I still don't completions for git subtree

they're only autoloaded if `git-subtree` is in `$PATH`,
which is also required for `git subtree` to work.
You might want to install a symlink to `/usr/lib/git-core/git-subtree` and restart fish - there should be no need to source it yourself

PR #11063, opened 2025-01-18
Adds git subtree completion

FeryET:
## Description

Adds subtree completion for git.

Fixes issue #10903

## TODOs:

### DIFF
--- a/share/completions/git-subtree.fish
+++ b/share/completions/git-subtree.fish
@@ -1,0 +1,27 @@
+complete -f -c git -a 'subtree' -d 'Manage git subtrees'
+# Git subtree common completions
+complete -f -c git -n '__fish_git_using_command subtree;' -s 'q' -l 'quiet' -d 'Suppress output'
+complete -f -c git -n '__fish_git_using_command subtree;' -s 'd' -l 'debug' -d 'Debug output'
+complete -f -c git -n '__fish_git_using_command subtree;' -s 'P' -l 'path' -d 'Path to the subtree'
+
+# Git subtree subcommands
+complete -f -c git -n '__fish_git_using_command subtree;' -a "add" -d "Add a new subtree to the repository"
+complete -f -c git -n '__fish_git_using_command subtree;' -a "merge" -d "Merge changes from a subtree into the repository"
+complete -f -c git -n '__fish_git_using_command subtree;' -a "split" -d "Extract a subtree from the repository"
+complete -f -c git -n '__fish_git_using_command subtree;' -a "pull" -d "Fetch and integrate changes from a remote subtree"
+complete -f -c git -n '__fish_git_using_command subtree;' -a "push" -d "Push changes to a remote subtree"
+
+# Completions for push and split subcommands
+complete -f -c git -n '__fish_git_using_command subtree; and __fish_seen_subcommand_from push split' -l 'annotate' -d 'Annotate the commit'
+complete -f -c git -n '__fish_git_using_command subtree; and __fish_seen_subcommand_from push split' -s 'b' -l 'branch' -d 'Branch to split'
+complete -f -c git -n '__fish_git_using_command subtree; and __fish_seen_subcommand_from push split' -l 'ignore-joins' -d 'Ignore joins during history reconstruction'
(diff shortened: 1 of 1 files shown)


share/completions/git.fish
```diff
@@ -622,6 +622,76 @@ function __fish_git_config_keys
...
+        echo "--annotate Annotate the commit"
+        echo "-b Branch to split"
+        echo "--branch Branch to split"
+        echo "--ignore-joins Ignore joins during history reconstruction"
+        echo "--onto Specify the commit ID to start history reconstruction"
+        echo "--rejoin Merge the synthetic history back into the main project"
+    end
+
+    ## git subtree completion when invoking git
+    complete -f -c git -n __fish_git_needs_command -a subtree -d 'Manage git subtrees'
```
krobelus, line 666: `subtree` is  in contrib, so not installed by default. Can we only load this one if it exists?
I guess this can be moved to `git-subtree.fish`, then that should happen automatically.

share/completions/git.fish
```diff
@@ -622,6 +622,76 @@ function __fish_git_config_keys
...
+        complete -f -c git -n '__fish_seen_subcommand_from subtree; and __fish_seen_subcommand_from merge' -a $item -d $desc
+        complete -f -c git -n '__fish_seen_subcommand_from subtree; and __fish_seen_subcommand_from pull' -a $item -d $desc
+    end
+
+    # Completions when invocking 'split' and 'push'
+    for line in (__fish_git_subtree_split_push_options)
+        set -l item (echo $line | cut -d' ' -f1)
+        set -l desc (echo $line | cut -d' ' -f2-)
+        complete -f -c git -n '__fish_seen_subcommand_from subtree; and __fish_seen_subcommand_from split' -a $item -d $desc
+        complete -f -c git -n '__fish_seen_subcommand_from subtree; and __fish_seen_subcommand_from push' -a $item -d $desc
```
krobelus, line 689: It would be good to get rid of the external processes (invoked once at startup).
I don't think you need the loop here, you can simply change the condition to `__fish_seen_subcommand_from push split` and then list everything  in the `-a` argument inline, no need for a function

share/completions/git.fish
```diff
@@ -616,6 +616,76 @@ function __fish_git_config_keys
...
+        set -l desc (echo $line | cut -d' ' -f2-)
+        complete -f -c git -n '__fish_seen_subcommand_from subtree' -a $item -d $desc
+    end
+
+    # Completions when invoking 'add' and 'merge'
+    for line in (__fish_git_subtree_add_merge_options)
+        set -l item (echo $line | cut -d' ' -f1)
+        set -l desc (echo $line | cut -d' ' -f2-)
+        complete -f -c git -n '__fish_seen_subcommand_from subtree; and __fish_seen_subcommand_from add' -a $item -d $desc
+        complete -f -c git -n '__fish_seen_subcommand_from subtree; and __fish_seen_subcommand_from merge' -a $item -d $desc
```
krobelus, line 674: You're adding options via `-a`. This doesn't really fit; `-a` is for option *arguments*.

Please use

	complete -f -c git -n '__fish_seen_subcommand_from subtree; and __fish_seen_subcommand_from add merge pull' -l squash -d Squash
	complete -f -c git -n '__fish_seen_subcommand_from subtree; and __fish_seen_subcommand_from add merge pull' -s m -l message -d Message -r

similarly elsewhere. I don't think we'll need to keep any of the functions unless we really want to avoid deduplication, which seems unlikely

Also please move your changes to `share/completions/git-subtree.fish`

share/completions/git.fish
```diff
@@ -616,6 +616,76 @@ function __fish_git_config_keys
...
+    # Completions when invocking 'split' and 'push'
+    for line in (__fish_git_subtree_split_push_options)
+        set -l item (echo $line | cut -d' ' -f1)
+        set -l desc (echo $line | cut -d' ' -f2-)
+        complete -f -c git -n '__fish_seen_subcommand_from subtree; and __fish_seen_subcommand_from split' -a $item -d $desc
+        complete -f -c git -n '__fish_seen_subcommand_from subtree; and __fish_seen_subcommand_from push' -a $item -d $desc
+    end
+end
+# Run the git subtree completion function
+__fish_render_complete_git_subtree_completion
```
krobelus, line 687: please get rid of the function, I don't think it adds anything

Issue #11041, opened 2025-01-12
Sudo completion breaks with apostrophe present in argument


WaitingAtTheCrossroads:
```bash
$ fish --version
fish, version 4.0b1-4-g687001f8b

$ echo $version
4.0b1-4-g687001f8b

$ uname -a
Linux openmediavault 6.9.10+bpo-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.9.10-1~bpo12+1 (2024-07-26) x86_64 GNU/Linux

$ echo $TERM
tmux-256color # Windows Terminal v1.21.3231.0
```

New fish user here.  😄 

I'm experiencing an issue where tab completion fails when a command prefixed with `sudo` has an apostrophe in its argument.

For instance, the following works fine:

```bash
touch foo

mv "foo" /[TAB] # Works as expected

sudo mv "foo" /[TAB] # Works as expected
```

But the following causes tab completion to fail:

```bash
touch "b'ar"

mv "b'ar" /[TAB] # Works as expected

sudo mv "b'ar" /[TAB] # Tab completion fails
```

The same happens with fish 3.6.0 from my debian bookworm install, so I guess the behavior has been present for a while.

Thanks for looking! 😃

PR #11067, opened 2025-01-19
Fix unnecessary token expansion in sudo completion

kerty0:
## Description

Fixes issue #11041.

`commandline -x` expands strings so `__fish_complete_subcommand` recieve `mv b'ar` instead of `mv "b'ar"`.

### DIFF
--- a/share/completions/doas.fish
+++ b/share/completions/doas.fish
@@ -3,7 +3,7 @@
 #
 
 function __fish_doas_print_remaining_args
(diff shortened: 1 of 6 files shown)


krobelus:
whoa nice find, that's embarassing. I tried to add the remaining cases, with a slightly different approach.
(This approach is broken for multiline tokens but a lot of things in fish are. Not sure which one to prefer)

faho:
Fixed in 6b5ad163d3084f91e756f27dacfb2e8d9032aecd

PR #11068, opened 2025-01-19
Fix completion for celsius flag in acpi(1)

ccoVeille:
## Description

The acpi parameter is named `--celsius` (without a typo), not `--celcius`

## TODOs:

- [X ] Changes to fish usage are reflected in user documentation/manpages.
- [ ] Tests have been added for regressions fixed

PR #11069, opened 2025-01-19
Fix typos in completions scripts

ccoVeille:
## Description

Fixed typo I found in share/completions

I used:
- typos
- codespell
- my own eyes


ccoVeille:
I dare to ping people as there is a high risk of conflicts if the PR stays open for too long

- @ridiculousfish because you approved my other PR #11071
- @zanchey we had fun with the `--celcius` flag with acpi completion https://github.com/fish-shell/fish-shell/pull/11068#issuecomment-2622718958

share/completions/cargo.fish
```diff
@@ -3,7 +3,7 @@
 ## --- WRITTEN MANUALLY ---
 set -l __fish_cargo_subcommands (cargo --list 2>&1 | string replace -rf '^\s+([^\s]+)\s*(.*)' '$1\t$2' | string escape)
 
-# Append user-installed extensions (e.g. cargo-foo, invokable as `cargo foo`) to the list of subcommands (à la git)
+# Append user-installed extensions (e.g. cargo-foo, invocable as `cargo foo`) to the list of subcommands (à la git)
```
zanchey, line 6: I reckon either is acceptable?

share/completions/cdrecord.fish
```diff
@@ -47,7 +47,7 @@ complete -c cdrecord -o setdropts -d "Set driver specific options"
 complete -c cdrecord -o checkdrive -d "Checks if a driver for the current drive is present"
 complete -c cdrecord -o prcap -d "Print the drive capabilities for SCSI-3/mmc compliant drives"
 complete -c cdrecord -o inq -d "Do an inquiry for the drive, print the inquiry info"
-complete -c cdrecord -o scanbus -d "Scan all SCSI devices on all SCSI busses and print the inquiry strings"
+complete -c cdrecord -o scanbus -d "Scan all SCSI devices on all SCSI buses and print the inquiry strings"
```
zanchey, line 50: This one is ok both ways too - https://www.merriam-webster.com/grammar/plural-of-bus

ccoVeille: Both are valid, so keeping the existing wording or using the one I suggest are valid.

I would the then tend to avoid having diff in git history for nothing, and I agree with not updating it

But in the exact article you mentioned, you can read

> The plural of the noun bus is buses. You might see the plural busses, but that form is so rare that it seems like an error to many people.

So I would like to keep the change I'm introducing here

share/completions/cargo.fish
ccoVeille: You are right. Both are valid, so my suggestion too 😅

Also there is a discussion about this exact debate in the tool I used to detect/fix typo

- https://github.com/crate-ci/typos/issues/1065

So I would say, it's up to you.

I can either change it, or we leave it like that, and the PR can be merged as is, considering my [reply](https://github.com/fish-shell/fish-shell/pull/11069#discussion_r1935408675) on "busses"

ridiculousfish: let's run with "invocable"

share/completions/cdrecord.fish
ridiculousfish: going with "buses"

zanchey:
Sorry, I hadn't finished reading the whole patch (it's a big one!) - but all looks good to me now.

share/completions/mksquashfs.fish
```diff
@@ -45,7 +45,7 @@ complete -c mksquashfs -o xattrs -d 'Store extended attributes'
 complete -c mksquashfs -o xattrs-exclude -xd 'Exclude any xattr names matching <regex>'
 complete -c mksquashfs -o xattrs-include -xd 'Include any xattr names matching <regex>'
 complete -c mksquashfs -o xattrs-add -xd 'Add the xattr <name> with <val> to files'
-complete -c mksquashfs -o version -d 'Print version, licence and copyright message'
+complete -c mksquashfs -o version -d 'Print version, license and copyright message'
```
ridiculousfish, line 48: keeping 'licence' 🇦🇺

PR #11074, opened 2025-01-20
completions/tmux: some windows and panes boolean flag completions

ilyagr:
This documents some non-argument options for the window and panes commands. The choice of what to document is somewhat arbitrary, this commit is biased towards options that I find confusing or misleading without documentation (is `-a` "all" or "after"?) and the command that seem more useful to me.

I also didn't cover the options that would be covered by #10855 (though this PR can be used independently). I'm not sure how much difference this made, it might not matter at all.


zanchey:
Cherry-picked as f1bb4e02f

Issue #11050, opened 2025-01-15
Tab autocompletion in quoted filenames


MarcinKonowalczyk:
## Description

Run the following commands:

```
mkdir temp && cd temp
touch 'File Name With Spaces' 
touch file_name_without_spaces
touch file_name_without_spaces && echo 'a bunch of other commands' 2>&1 > /dev/null
```

Now say i want to re-run the latest command on `'File Name With Spaces'`. Press up arrow, navigate to `file_name_without_spaces` and delete it. Start writing `File\ N<tab>` and the autocompletion kicks in no problem. Repeat but start writing `'File N<tab>` and the autocompletion does not work. This seems to happen *only* in the middle of editing a historical command, hence the setup above.

https://asciinema.org/a/3FvbVP8qXwYq3KAcCRujilMUA

## Context

- fish, version 4.0b1 💪
- Darwin marcin.local 24.1.0 Darwin Kernel Version 24.1.0: Thu Oct 10 21:02:26 PDT 2024; root:xnu-11215.41.3~2/RELEASE_ARM64_T8122 arm64
- alacritty 0.14.0 (22a4475)
- Still happens with `sh -c 'env HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_DIRS= fish'`

kerty0:
When you do `touch 'File N<tab> && echo 'a bunch of other commands' 2>&1 > /dev/null` Fish will attempt to complete the string `'File N && echo'` and fail.
To achieve the desired result, end the string first:
`touch 'File N'<tab> && echo 'a bunch of other commands' 2>&1 > /dev/null`

MarcinKonowalczyk:
Yup, can confirm it works, but it does autocomplete like this: touch 'File Name With Spaces'  which is a bit ugly. I guess not a bug but an enhancement proposal then?

1) Handle the `touch 'File N<tab> && echo 'a bunch of other commands' 2>&1 > /dev/null` case by splitting at command delimiters (`&&`, `||`, `>`, `>>`, `|` etc) and autocompleting in blocks?
2) Handle the `touch 'File N<tab> && echo no\ quote\ marks\ here 2>&1 > /dev/null` case to autocomplete to `touch 'File Name With Spaces' ...` as opposed to `touch 'File Name With Spaces'`

There might be strong reasons against either of those though and i don't think this is a major thing.

kerty0:
[...]

PR #11080, opened 2025-01-22
Improve flash

kerty0:
## Description

This pull request changes the behavior of flash in the following ways:
1. Failed history token search: now highlights the matched token if it isn't empty; otherwise, highlights the full command line.
2. Other failed history search: now highlights the full command line.
3. Failed deletion of autosuggestion: now highlights the autosuggestion.
4. Failed undo/redo: now highlights the full command line.
5. Failed completion: now highlights the completed token (#11050).

Other Changes:
- Converted `SourceRange` to use `usize` instead of `u32`.
- Made undo/redo not update autosuggestion if failed.
- Make `parse_util_token_extent` return its output instead of mutating input.
- Some other code refactoring.

### DIFF
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -1825,7 +1825,7 @@ impl ReaderData {
                 .as_ref()
                 .is_some_and(|saved_autosuggestion| {
                     self.conf.autosuggest_ok
-                        && self.history_search.is_at_end()
+                        && self.history_search.is_at_present()
                         && edit.replacement.is_empty()
                         && edit.range.start == saved_autosuggestion.search_string_range.end
                         && !edit.range.is_empty()
@@ -1873,7 +1873,7 @@ impl ReaderData {
 
     /// Apply the history search to the command line.
     fn update_command_line_from_history_search(&mut self) {
-        let new_text = if self.history_search.is_at_end() {
+        let new_text = if self.history_search.is_at_present() {
             self.history_search.search_string()
         } else {
             self.history_search.current_result()
@@ -2939,7 +2939,7 @@ impl<'a> Reader<'a> {
 
                 let was_active_before = self.history_search.active();
 
-                if self.history_search.is_at_end() {
+                if self.history_search.is_at_present() {
                     let el = &self.data.command_line;
                     if mode == SearchMode::Token {
                         // Searching by token.
@@ -2994,7 +2994,7 @@ impl<'a> Reader<'a> {
                 if !found && !was_active_before {
                     self.history_search.reset();
                 } else if found
-                    || (dir == SearchDirection::Forward && self.history_search.is_at_end())
+                    || (dir == SearchDirection::Forward && self.history_search.is_at_present())
                 {
                     self.update_command_line_from_history_search();
                 }
@@ -3053,7 +3053,7 @@ impl<'a> Reader<'a> {
             #[allow(deprecated)]
             rl::HistoryDelete | rl::HistoryPagerDelete => {
                 // Also applies to ordinary history search.
-                let is_history_search = !self.history_search.is_at_end();
+                let is_history_search = !self.history_search.is_at_present();
                 let is_autosuggestion = self.is_at_autosuggestion();
                 if is_history_search || is_autosuggestion {
                     self.input_data.function_set_status(true);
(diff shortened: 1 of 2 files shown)


src/parse_constants.rs
```diff
@@ -38,25 +38,10 @@ bitflags! {
...
-            length: 0,
-        }
-    }
-}
-
-impl SourceRange {
-    pub fn as_usize(&self) -> std::ops::Range<usize> {
-        (*self).into()
-    }
+    pub start: usize,
```
krobelus, line 43: I'm not sure if we should do this; so far the approach is to store things as u32 but perform all computation as `usize`.

if we want to make this change, we should apply it in all places that use `SourceOffset`, not just here.

Until then
1. `SourceRange` should use `SourceOffset` too.
   If you look at the git blame, you can read between the lines that
   the only reason why it no longer uses `SourceOffset` (alias `source_offset_t`)
   is that CXX does not support type aliases in Rust FFI blocks.
2. the u32/SourceOffset members should be made private, and only get/via via methods
   that take `usize`, to remove bad `start as usize` casts at callsites.

This will make a future migration trivial.
I guess the reason for that is to be more idiomatic (and not to support >4GB files).

Maybe move this to a different PR.

src/reader_history_search.rs
```diff
@@ -110,6 +111,17 @@ impl ReaderHistorySearch {
         self.search().original_term()
     }
 
+    /// Return the range of the original search string or current match in the command line.
+    pub fn token_range(&self) -> Range<usize> {
+        let length = if !self.active() || self.is_at_end() {
```
krobelus, line 116: The function name is misleading;
actually we can simplify this because it's only ever called if `self.is_at_end()` is true.
So this will always use the search string range.

Better to remove dead code even if it may be useful in future.

Maybe add `assert!(self.active() && self.is_at_end())`
and call it `search_range_assume_at_end()`
for consistency with `search_range_if_active()`

Could also change `search_range_if_active` to return something
if at end. Then we just need one function, but it may be
awkward to adjust existing callers, so the above seems more obvious.

CHANGELOG.rst
```diff
@@ -23,6 +23,7 @@ Interactive improvements
   Such a malicious file can still potentially insert arbitrary text into the command line but can no longer execute it directly (:issue:`10987`).
 - The history search now preserves ordering between :kbd:`ctrl-s` forward and :kbd:`ctrl-r` backward searches.
 - Left mouse click now can select pager items.
+- Flash now highlights the commandline section relevant to its cause (:issue:`11050`).
```
krobelus, line 26: Trying to add some context:

```suggestion
- When there are no more search matches or completions, fish flashes the search string instead if all text left of the cursor (:issue:`11050`).
```

src/reader.rs
```diff
@@ -4000,14 +4010,15 @@ impl<'a> Reader<'a> {
     }
 
     /// Flash the screen. This function changes the color of the current line momentarily.
-    fn flash(&mut self) {
+    fn flash(&mut self, flash_range: Range<usize>) {
         // Multiple flashes may be enqueued by keypress repeat events and can pile up to cause a
         // significant delay in processing future input while all the flash() calls complete, as we
         // effectively sleep for 100ms each go. See #8610.
         let now = Instant::now();
         if self
```
krobelus, line 4018: thanks, it was very irritating that this didn't flash when the cursor is at the beginning of the command line

src/parse_constants.rs
kerty0: > Maybe move this to a different PR.

Ok

src/reader_history_search.rs
kerty0: > actually we can simplify this because it's only ever called if self.is_at_end() is true.

No, we call it when we reach the other end of the history as well. `self.is_at_end()` is misleading itself; it returns true only if the current index equals 0, which means we're at the start of the history.

krobelus:
merged all but the last commit

src/reader_history_search.rs
krobelus: I'm not sure if `is_at_start()` is a net improvement over `is_at_end()`,
because "end of history" means "present", so it does imply the right thing.
But I think it is a bit ambiguous. Maybe we can say something like "youngest" instead.

kerty0: Maybe `is_at_present()`? After all, `self.matches[0]` isn't part of the history but is the current search term.

PR #11086, opened 2025-01-23
language.rst: make description of features more consistent (minor)

ilyagr:
The version where a feature became the default is now described inline, to make it a single source of truth. I could have fixed the other section where this was described, but this seemed easier.

(Feel free to fix it up in any other way instead of using this PR as well)


ridiculousfish:
Thank you.

zanchey:
Cherry-picked into Integration_4.0.0 as fa5de3ece8cc8e

Issue #11029, opened 2025-01-08
Long multi-line commands are truncated without scrollback


mqudsi:
If you paste a multi-line commandline (or retrieve it from history) that is longer than the number of tty lines visible on the screen, then scroll up in the terminal emulator, the contents of the commandline are truncated at the height of the tty emulator. The prompt is missing.

Scrolling up via fish (by using the up arrow) works as the contents are redrawn line-by-line, but the contents are excluded from scrollback so scrolling via the terminal emulator itself is completely broken. The workaround is to use alt-v to edit the commandline in $VISUAL.

Noted under tmux and conhost. This probably came from the changes to how multiline commandlines are now rendered.

krobelus:
FWIW, the full command will be in scrollback once executed.

The last two comments in https://github.com/fish-shell/fish-shell/issues/10827 are my to-dos to fix the scrollback while composing the command.

mqudsi:
Oh, OK. 

I marked it as a regression because if we ship 4.0 with these outstanding TODOs then it would be a regression in terms of buggy behavior that wasn't previously present in a release.

krobelus:
Right, I missed that scrollback used to have the entire command (despite the glitches).

To fix the regression, it may be enough to remove the truncation logic.
This would mean that whenever we redraw, we'd add another `(count (commandline)) - $LINES` worth of lines to the scrollback
(so similar to 3.7.1 behavior but hopefully without the glitches).

Unless we switch to using the alternate screen (which would also break scrollback), the other fix depends on `CSI Ps S`/`CSI Ps T` (scroll forward/backward).

Given that Putty supports the scroll forward/backward commands, they're probably widely available.
If scrolling commands are ignored by the terminal it would behave as described above, adding extra lines to scrollback on redraw.

kerty0:
From what I've gathered, [CSI Ps S](https://ghostty.org/docs/vt/csi/su) and [CSI Ps T](https://ghostty.org/docs/vt/csi/sd) only operate on the viewport and can't modify the scrollback buffer.

Therefore, it is the choice between only rendering the viewport, or spamming the scrollback with extra lines when redrawing or moving up in a long command line. 

I think only rendering the viewport is the better approach.

krobelus:
That's right!

Currently we always redraw the whole commandline.
Once we only redraw the text that changed, we'd avoid adding garbage to scrollback in some cases, but not in all:
[...]

PR #11094, opened 2025-01-26
Some fixes for prompt and commandline

kerty0:
## Description
[Remove space for prompt if it occupies full line](https://github.com/fish-shell/fish-shell/commit/73f44c73fc1995023c73d4204d1fc79f0c35dca1):
If the prompt occupies the full width of the first line, each new line will soft wrap, leaving empty lines between. This change removes the indentation that matches the start of the line to the end of the prompt if the prompt width equals the terminal width. (Perhaps this could be extended to cases where the difference between the terminal width and prompt width is small enough, e.g., 2-4 characters?)

[Fix not printing osc_133_prompt_start without prompt](https://github.com/fish-shell/fish-shell/commit/854089d3ff9040c6b56d79b80eaa2271d0a961e9):
If the left prompt is defined to be empty, osc_133_prompt_start would not be printed at all. This commit fixes that.

[Fix crash when clicking on empty lines](https://github.com/fish-shell/fish-shell/commit/a198c4d1a286e8d6b34f3484e06cdabea32fb17f):
This fixes a crash that occurred when clicking on lines without characters, such as soft wrapped lines or first line without a prompt.

[Fix variable completion autosuggestion check](https://github.com/fish-shell/fish-shell/commit/d0db37379bdd01594316a02ee4fc808dd097ed14):
Currently, fish adds descriptions to completions only if description and autosuggestion flags are active. However, autosuggestions do not use completion descriptions. So, I inverted the check for the autosuggestion flag, and with this change, the completion of variables now shows their values.

[Fix pager occupying empty soft wrapped lines](https://github.com/fish-shell/fish-shell/commit/2419447415eddbc51e15f48309dd3e53bdc66161):
This commit adds an additional line to the desired screen if the previous line was full, preventing the pager from being drawn on soft-wrapped lines or empty prompt line.

[Fix command line not rendering at the prompt position when scrolled](https://github.com/fish-shell/fish-shell/commit/791592febd8d83fd163ed1b9ef686fc4df484629):
This fixes an issue where, when scrolled, the space that the prompt would occupy if not scrolled is not rendered.

[Fix deletion of prompt in last 2 columns when soft wrapped](https://github.com/fish-shell/fish-shell/commit/35a8577424183c47990b1e3c79863441bcfcc357):
This fixes an issue where the prompt is erased in the last 2 columns of the terminal if the first line is wrapped. When drawing a soft wrapped line, the last 2 characters of the previous line are redrawn. If these 2 characters are occupied by the prompt, they are replaced with `' '`. This commit ensures that the prompt is redrawn if it occupies the full line instead, maintaining soft wrap.

[Fix inconsistent autosuggestion truncation on soft wrapped lines](https://github.com/fish-shell/fish-shell/commit/5f78bd37f42519980d486f628d214d295d0d1019):
This fixes an issue where, if the right prompt exists or is empty, the autosuggestion would be truncated if line isn't soft wrapped and not shown on soft wrapped lines at all. If the autosuggestion is hidden, it would be truncated on the first line, but on other lines, it would not be limited at all. This commit ensures that autosuggestions cannot soft wrap. (Perhaps autosuggestions should always soft wrap unless the right prompt is in the way?)

[Make at least one character of the autosuggestion always shown](https://github.com/fish-shell/fish-shell/commit/84532e607928c14c50a7b6aeb4b2101ccbcf418d):
Currently, there are situations where an autosuggestion exists but is not shown at all. This commit guarantees that at least the ellipsis_char is shown, regardless of the situation.


src/complete.rs
```diff
@@ -1695,7 +1695,7 @@ impl<'ctx> Completer<'ctx> {
             };
 
             let mut desc = WString::new();
-            if self.flags.descriptions && self.flags.autosuggestion {
+            if self.flags.descriptions && !self.flags.autosuggestion {
```
krobelus, line 1698: Wow - `git blame` shows this regressed in 17bd7d0e40 (Switch completion_request_options_t from a list of flags to a struct, 2022-06-07).
Crazy that now one noticed.
Will cherry-pick this one to the release branch.

src/screen.rs
```diff
@@ -375,7 +375,11 @@ impl Screen {
         }
 
         // If overflowing, give the prompt its own line to improve the situation.
-        let first_line_prompt_space = layout.left_prompt_space;
```
krobelus, line 378: Some preliminary comments; I still need to finish reading the last two commits;
I don't expect many changes will be necessary.

src/screen.rs
```diff
@@ -375,7 +375,11 @@
...
+        let first_line_prompt_space = if layout.left_prompt_space == screen_width {
```
krobelus, line 378: I'd prefer to put change descriptions into the commit message, especially
if the commit has no dependency on the rest of the PR.
I know github makes it hard to find commit messages but other tools work great with that.
It will help us next time we touch this code.

--

Looks good -- confirmed the fix with

	fish -C 'function fish_prompt; printf %0"$COLUMNS"d 0; end'
	# type "begin" and Enter
	# observe that there was a spurious empty line

 > (Perhaps this could be extended to cases where the difference between the terminal width and prompt width is small enough, e.g., 2-4 characters?)

Playing with `function fish_prompt; printf %0"$(math $COLUMNS - 10)"d 0; end`
shows that the behavior is weird for multline commands.
every line that overflows $COLUMNS (from indentation or actual text) is broken up,
the start is on the right side of the screen, and the tail is wrapped.

My quick thought: if this scenario happens (e.g. the left prompt occupies too much space),
then we could place the command line on its own line (with 0 indentation).

Not a very common scenario of course.

src/screen.rs
```diff
@@ -443,6 +447,16 @@
...
         };
 
+        // Add an empty line if the last line was soft-wrapped or if there are no lines.
+        if !self
+            .desired
+            .line_datas
+            .last()
+            .is_some_and(|line| line.len() != screen_width)
+        {
+            self.desired.add_line();
```
krobelus, line 457: right; tested with this command

	target/debug/fish -C '
	    function fish_prompt; printf \>; end
	    bind ctrl-g \'commandline  "ls $(printf %0"$(math $COLUMNS - 5)"d 0) "\''

The added line may end up empty; I wonder if we should suppress it in this case?

src/screen.rs
```diff
@@ -601,8 +615,18 @@
         let x = viewport_position.x - viewport_prompt_x;
         let line = self.actual.line(y);
         let x = x.max(line.indentation);
-        let char = line.text.get(x).or(line.text.last()).unwrap();
-        match char.offset_in_cmdline {
+        let offset = line
+            .text
+            .get(x)
+            .or(line.text.last())
+            .or(if y > 0 {
```
krobelus, line 622: Interesting. I thought it was guaranteed that every rendered line has at
least one character (if only the trailing newline) because why else would
we draw it. I guess that was wrong, I'll try to reproduce with kitty.

src/screen.rs
```diff
@@ -1057,19 +1089,21 @@
...
         if zelf.scrolled() && !is_final_rendering {
             zelf.r#move(0, 0);
             zelf.outp
                 .borrow_mut()
                 .tputs_if_some(&term.and_then(|term| term.clr_eol.as_ref()));
             zelf.actual_left_prompt.clear();
             zelf.actual.cursor.x = 0;
-        } else if left_prompt != zelf.actual_left_prompt || (zelf.scrolled() && is_final_rendering)
+        } else if left_prompt != zelf.actual_left_prompt
+            || left_prompt.is_empty()
```
krobelus, line 1102: is this because `actual_left_prompt` is initialized as empty (and probably reset after every command)?
If true, wouldn't it be more obvious to make it an `Option<WString>`? 
Since `function fish_prompt; end` is valid (which seems fine), so we want to properly distinguish between "empty" and "uninitialized".

src/screen.rs
```diff
@@ -1086,6 +1120,20 @@
...
+            zelf.r#move(0, 0);
+            let start = if left_prompt_layout.line_breaks.is_empty() {
+                osc_133_prompt_start(&mut zelf);
+                0
+            } else {
+                left_prompt_layout.line_breaks.last().unwrap() + 1
+            };
+            zelf.write_str(&left_prompt[start..]);
+            zelf.actual.cursor.x = left_prompt_width;
+            zelf.soft_wrap_location = Some(Cursor { x: 0, y: 1 })
```
krobelus, line 1136: I wonder if this logic could be folded into the previous else-if case.
But this may be the best solution. I'd need to take a closer look.
Do you have a test case for this code block?
Something that seems odd is that `osc_133_prompt_start` is not printed in all cases.

The rest of this commit looks great.

src/screen.rs
```diff
@@ -1099,7 +1147,11 @@
         for i in 0..zelf.desired.line_count() {
             zelf.actual.create_line(i);
 
-            let start_pos = if i == 0 { left_prompt_width } else { 0 };
+            let start_pos = if i == 0 && !zelf.scrolled() {
```
krobelus, line 1150: > This fixes an issue where, when scrolled, the space that the prompt would
 > occupy (if not scrolled) is not rendered.

How to reproduce?
If I run

	fish -C 'bind ctrl-g \'commandline $(seq $LINES)\''

and type `ctrl-g shift-enter` things work as expected.
The (one-line) prompt is only rendered if we're not scrolled.
Maybe this is about multi-line prompts?

src/screen.rs
```diff
@@ -1985,81 +2017,70 @@
...
-    let mut truncated_autosuggestion = |indent: &mut Vec<i32>, right_prompt_width: usize| {
-        let width = if let Some(pos) = commandline_before_suggestion
-            .chars()
-            .rposition(|c| c == '\n')
-        {
-            left_prompt_width
-                + usize::try_from(indent[pos]).unwrap() * INDENT_STEP
-                + autosuggestion_line_explicit_width
-        } else {
-            left_prompt_width + right_prompt_width + first_command_line_width
```
krobelus, line 2016: > This fixes an issue where, if the right prompt exists or is empty, the
 > autosuggestion would be truncated if line isn't soft wrapped
 
 I can reproduce that and the fix looks good.
 
 > and not shown on soft wrapped lines at all.
 
 (not yet sure how to provoke that)
 
 > If the autosuggestion is hidden, it
 > would be truncated on the first line, but on other lines, it would not
 > be limited at all.
 
 you mean the autosuggestion would occupy multiple lines? I haven't managed to trigger that.
 
 > This commit ensures that autosuggestions cannot soft
 > wrap. (Perhaps autosuggestions should always soft wrap unless the right
 > prompt is in the way?)
 
 Maybe, but for the case where the prompt is at the bottom-most line of the screen,
 the autosuggestion should not push the commandline up.
 But yeah, in the general case I think rendering the full autosuggestion would be an overall improvement.
 
 Perhaps we can even suggest multi-line command lines. Not yet sure.

src/screen.rs
```diff
@@ -1985,81 +2017,70 @@
...
-        result.left_prompt = left_prompt;
-        result.left_prompt_space = left_prompt_width;
-        result.autosuggestion = truncated_autosuggestion(indent, 0);
-        return result;
+    // Now we should definitely fit.
+    assert!(left_prompt_width + right_prompt_width <= screen_width);
+
+    // Calculate space available for autosuggestion.
+    let pos = commandline_before_suggestion
+        .chars()
```
krobelus, line 2051: this is now unused, as of the last commit

src/screen.rs
```diff
@@ -1985,81 +2017,70 @@
...
+        .rposition(|c| c == '\n');
+    let width = (left_prompt_width
+        + autosuggestion_line_explicit_width
+        + commandline_before_suggestion
```
krobelus, line 2055: yeah this is great. The hidden autosuggestion was just confusing.

src/screen.rs
kerty0: > If the autosuggestion is hidden,

My mistake: I meant right prompt instead. But it also wrong. Right version:

If the first line is not soft wrapped, the autosuggestion will be truncated and not shown on soft wrapped lines. If the first line is soft wrapped, the autosuggestion will not have any limit.

src/screen.rs
kerty0: > Something that seems odd is that `osc_133_prompt_start` is not printed in all cases.

We reprint only last line of prompt, so we print `osc_133_prompt_start` only if prompt is one line.

> Do you have a test case for this code block?

I added it to `tmux-prompt.fish`. Here is another version:
`fish -C 'function fish_prompt; printf %0"$(math $COLUMNS)"d 0; end'` and type anything.

src/screen.rs
kerty0: > The added line may end up empty; I wonder if we should suppress it in this case?

If we don't always add an empty line, then we end up with inconsistent clicking behavior: 
`fish -C 'function fish_prompt; end; bind ctrl-g \'commandline $(printf %0"$(math $COLUMNS)"d 0) $(printf %0"$(math$COLUMNS)"d 0)\''`
There is a difference in behavior when clicking on the first empty line versus the second empty line.

src/screen.rs
kerty0: The problem is that we don't store `'\n'` in `screen.actual` or `screen.desired`, but perhaps we should. The current implementation moves the cursor to position 0 when clicking on an `alt-Enter` newline. Storing `'\n'` would allow for a much simpler solution to this issue.

src/screen.rs
krobelus: > place the command line on its own line

This might also be too much magic to be worth it; we probably want to avoid
scenarios where a single key stroke causes chaotic layout change.

src/screen.rs
krobelus: > Storing `'\n'` would allow for a much simpler solution to this issue.

I see. I don't know of any strong reasons against (or for) including the trailing newline.

The trailing newline can be missing (in the last line of a commandline).
But I guess it still helps because we only need to check if we're at the very end.

Either way seems fine, I guess it depends on what other changes we want to make in future.

src/screen.rs
krobelus: > We reprint only last line of prompt, so we print `osc_133_prompt_start` only if prompt is one line.

yeah that seems fine. OSC 133 probably doesn't clearly define this case;
and for the current uses (`ctrl-shift-{j,k}`) its fine because we will render the entire prompt & commandline before executing.

> I added it to `tmux-prompt.fish`.

that test adds mutation coverage for the other changes in this commit but not for this else-if block; it still passes after I remove the whole
`else if Some(left_prompt_width) == screen_width && zelf.should_wrap(0) { ... }`.
I'm not yet sure in which case this code makes a difference.
This code path is actually taken in the tmux-prompt.fish test,
maybe we're missing an assertion on the output.

src/screen.rs
krobelus: Right that makes sense.
FWIW we might want to rework scrolling a bit, see https://github.com/fish-shell/fish-shell/issues/11029

src/screen.rs
krobelus: > If the first line is soft wrapped, the autosuggestion will not have any limit.

Right, this inconsistency was weird. It's fixed by the "Refactor
compute_layout" commit, probably because that one also truncates the
autosuggestion in case 5 -- this inconsistency was added recently in 1c4e5cadf2
(Autosuggestions in multi-line command lines, 2024-12-15).

krobelus:
LGTM :pray: 
Only thing I wanna figure out is the `else if Some(left_prompt_width) == screen_width && zelf.should_wrap(0) { ... }.`,
if it's a legit change that's probably fairly easy to do.

krobelus (approved):

src/screen.rs
kerty0: `else if Some(left_prompt_width) == screen_width && zelf.should_wrap(0) { ... }` maintains terminal soft wrap when  the first line changes.

krobelus:
nice, the latest push fixes another edge case when both the prompt and the first line of the command line are soft wrapped, and we are scrolled.

```fish
fish -C '
function fish_prompt
    printf "$(string repeat -N $COLUMNS -)!!"
end
bind ctrl-g \'commandline $(string repeat -N $COLUMNS 0)soft-wrapped $(seq $(math $LINES - 1))\''
```

`ctrl-g ctrl-p` used to wrongly render one line as blank.
I'll add a test

PR #11104, opened 2025-01-29
Remove fish.desktop

Justinzobel:
## Description

Removes fish.desktop file and CMake install entry. This is as a result of discussion https://github.com/fish-shell/fish-shell/discussions/11102


faho:
Thanks!

You missed a few bits, I added them in ef4fad763febfcd91f3d08c9c721047f82ea574f

Justinzobel:
> Thanks!
> 
> You missed a few bits, I added them in [ef4fad7](https://github.com/fish-shell/fish-shell/commit/ef4fad763febfcd91f3d08c9c721047f82ea574f)

Ah thanks. I didn't know the pixmap stuff was related and my CMake-foo is definitely very basic.

PR #11105, opened 2025-01-29
Make alt-{b,f} move in directory history if commandline is empty

krobelus:
`alt-{left,right}` move in the directory history (like in browsers). Arrow keys can be inconvenient to reach on some keyboards, so let's alias this to `alt-{b,f}`, which already have similar behavior. (historically the behavior was the same; we're considering changing that back on some platforms).

This happens to fix `alt-{left,right}` in Terminal.app (where we had a workaround for some cases), Ghostty and probably some versions of iTerm2, though that alone should not be the reason for this change.


faho:
This is one of those things where I can only say "fine by me" because I never use it.

I don't believe I've ever pressed alt-b/alt-f on purpose.

ridiculousfish:
What faho said.

PR #11107, opened 2025-01-30
On macOS, swap alt-{left,right,backspace,delete} with ctrl-*

### LINKED ISSUE #10926: behaviour of option+backspace differs between v3 and v4
Steps to reproduce:
* Install fish v4b1. I got mine from [here](https://github.com/fish-shell/homebrew-fish-beta-4)
* Type in the terminal something like this `vim .config/git/.gituser`
* Press `option`+`backspace` shortcut

Expected behaviour: one word starting from the end gets deleted
Actual behavior: whole path gets deleted

I am not sure if you consider this a bug, but behavior has definitely been changed

Env
```
vv@my-precious ~ $ fish --version
fish, version 4.0b1
vv@my-precious ~ $ kitty --version
kitty 0.38.0 created by Kovid Goyal
vv@my-precious ~ $ uname -a
Darwin my-precious.local 23.6.0 Darwin Kernel Version 23.6.0: Fri Nov 15 15:13:15 PST 2024; root:xnu-10063.141.1.702.7~1/RELEASE_ARM64_T6000 arm64
vv@my-precious ~ $ echo $TERM
xterm-kitty
vv@my-precious ~ $ 
```

krobelus:
Commit 6af96a81a8 (Default bindings for token movement commands,
2024-10-05) changed the behavior of keys like `alt-left` to move by
tokens instead of words.

This seems fine on non-macOS systems where GUI text fields use
`ctrl-left` for word movement.

But it's surprising on macOS where word movement is `alt-left`
everywhere.

Sublime Text uses the native key binding for word movement, and uses
the vacant chord (e.g. `alt-left` on Linux) for sub-word movement
(in camel case words).

Let's try to do the same, but for token movements.

There are not too many terminals that run on macOS. Try to detect
some of them in various ways, some more hacky than others.
1. Kitty supports reporting the client OS via `XTGETTCAP`[^XTGETTCAP].
   This is the most correct method, overriding others[^os-name]. 
   By default tmux doesn't let this request pass through.
   Add a workaround to bypass tmux, just like we do in
   `fish_clipboard_copy`.
2. We can detect iTerm2 based on LC_TERMINAL, even inside SSH.
3. Assume macOS if fish is built for macOS. This is not fully correct
   since the terminal may run on another system.

One of my concerns is that so far I personally override this change.
(when on macOS and running SSH into fish). Part of the reason is that
the version of tmux I use doesn't support binding ctrl-backspace
yet. I'm not sure if users prefer a native experience or consistency
across platforms. But users who regularly switch platforms have to
switch muscle memory regardless.

I don't know yet what I'd apply to the release branch.  Maybe a
subset of detection methods, or revert 6af96a81a8, I'm not sure how
many people are affected.

Alternatives:
1. revert and/or find different key bindings
2. leave things as they are

[^XTGETTCAP]: The current patch detects macOS only after sourcing
              config files. We can fix this by reshuffling some code.
[^os-name]: We might want to expose the (arbitrary) OS name,
            however the name reported by Kitty is like
            "bsd"/"linux"/"macos", which does not match `uname(1)`,
            so I'm not sure if that's stable enough for us. Hence
            the boolean macOS flag.

Closes #10926

## TODOs:

### DIFF
--- a/doc_src/cmds/status.rst
+++ b/doc_src/cmds/status.rst
@@ -17,6 +17,7 @@ Synopsis
     status is-no-job-control
     status is-full-job-control
     status is-interactive-job-control
+    status is-in-macos-terminal
     status current-command
     status current-commandline
     status filename
@@ -61,6 +62,10 @@ The following operations (subcommands) are available:
 
 **is-no-job-control** or **--is-no-job-control**
     Returns 0 if no job control is enabled.
+
+**is-in-macos-terminal**
+    Returns 0 if the terminal emulator is running on macOS.
+    Attempts to see through SSH and containers, unlike ``uname(1)``.
 
 **current-command**
     Prints the name of the currently-running function or command, like the deprecated :envvar:`_` variable.
--- a/share/completions/status.fish
+++ b/share/completions/status.fish
@@ -13,6 +13,7 @@ complete -f -c status -n "not __fish_seen_subcommand_from $__fish_status_all_com
 complete -f -c status -n "not __fish_seen_subcommand_from $__fish_status_all_commands" -a is-no-job-control -d "Test if new jobs are never put under job control"
 complete -f -c status -n "not __fish_seen_subcommand_from $__fish_status_all_commands" -a is-interactive-job-control -d "Test if only interactive new jobs are put under job control"
 complete -f -c status -n "not __fish_seen_subcommand_from $__fish_status_all_commands" -a is-full-job-control -d "Test if all new jobs are put under job control"
+complete -f -c status -n "not __fish_seen_subcommand_from $__fish_status_all_commands" -a is-in-macos-terminal -d "Test if the shell is running in a macOS terminal"
 
 # The subcommands that are not "is-something" which don't change the fish state.
 complete -f -c status -n "not __fish_seen_subcommand_from $__fish_status_all_commands" -a buildinfo -d "Print information on how this version fish was built"
--- a/share/functions/__fish_config_interactive.fish
+++ b/share/functions/__fish_config_interactive.fish
@@ -217,6 +217,12 @@ end" >$__fish_config_dir/config.fish
     end
     __fish_update_cwd_osc # Run once because we might have already inherited a PWD from an old tab
 
+    if set -q TMUX
+        # NOTE This will be invalidated when the next tmux client attaches
+        # to fish's window; but we only use it to detect macOS.
+        set -g __fish_tmux_client_tty (tmux display-message -p '#{client_tty}' 2>/dev/null)
+    end
+
     # Bump this whenever some code below needs to run once when upgrading to a new version.
     # The universal variable __fish_initialized is initialized in share/config.fish.
     set __fish_initialized 3800
--- a/share/functions/__fish_shared_key_bindings.fish
+++ b/share/functions/__fish_shared_key_bindings.fish
@@ -21,8 +21,20 @@ function __fish_shared_key_bindings -d "Bindings shared between emacs and vi mod
     $legacy_bind --preset $argv -k left backward-char
 
     # Ctrl-left/right - these also work in vim.
-    bind --preset $argv ctrl-right forward-word
-    bind --preset $argv ctrl-left backward-word
+    bind --preset $argv ctrl-right '
+        if status is-in-macos-terminal
+            commandline -f forward-token
+        else
+            commandline -f forward-word
+        end
+    '
+    bind --preset $argv ctrl-left '
+        if status is-in-macos-terminal
+            commandline -f backward-token
+        else
+            commandline -f backward-word
+        end
+    '
 
     bind --preset $argv pageup beginning-of-history
     bind --preset $argv pagedown end-of-history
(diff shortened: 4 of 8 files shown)


faho:
This is pretty dirty.

It's a special `status` subcommand just for a specific OS. I'm also not a fan of executing `tmux` inside of __fish_config_interactive.

How about this instead:

Add a `status client-operating-system` subcommand that *tries* to get the OS via query, and if it can't falls back on uname (or target_os or whatever).

That's more generally useful and cleaner, and the key bindings could then be `if contains -- (status client-operating-system) Darwin; or test "$LC_TERMINAL" = iTerm2`.

krobelus:
yeah I think we should either  ask kitty for `uname` names, so we can have
a consistent fallback. Or do lossy translation ("uname=FreeBSD" to "bsd"),
which is probably fine but I'm not sure.

We can probably pass tmux' underlying tty to `status client-operating-system`
to remove the tmux-specific code from Rust sources.
We could keep caching the result, or not.
Either way there will be some tmux scenarios that will not work.

faho:
For 4.0, I wouldn't want to add this since it adds querying.

So I'd probably either keep the bindings consistent (by reverting the token bindings or keeping them everywhere), or use a simple `uname` check (possibly with $TERM_PROGRAM or $LC_TERMINAL) there.

Tbh I'm not entirely sure if this query is worth it because it'll currently only work on kitty, and other than for bindings I can't think of anything we need the *terminal* operating system for. E.g. if you want to know what flavor of coreutils you're dealing with `uname` is a decent proxy, the terminal OS isn't.

ridiculousfish:
I don't think we should add this - a new `status` command seems like overkill and all of the checks around tmux, kitty, seem fragile and difficult to test. As faho says, if we must have checks, then let's make them simple based on `uname` and certain environment variables set by the system, not the terminal (sadly, none are ideal). I suggest:

```
set -q XPC_FLAGS || set -q __CFBundleIdentifier || test (uname) = "Darwin"
```

Reverting 6af96a81a8c6ff5e632d4dda7448f5f01d7a5d35 is also fine with me, but I'm primarily a Mac user so I'm biased.

Pulling this into 4.0 - I didn't realize this had regressed since apparently my custom key bindings were masking it. But it really is a regression on macOS

tommyknows:
> But it's surprising on macOS where word movement is alt-left everywhere.

It's not just that - the more painful thing is that the new `prevd-or-backward-word` mapping is on `ctrl-left` and `ctrl-right` for forward-movements.

On MacOS, these keys are (often) mapped to switch to the left and right desktop, respectively.

The first thing I did was map `alt-<arrow>` back to `backward-word` with a custom `bind`, until I realised that the directory-switching doesn't work anymore 😄 

I was also a bit confused as `functions` doesn't seem to list the old `prevd-or-backward-word` implementation:
```
$ functions | rg 'prevd'
prevd
prevd-or-backward-token
```

but it seems like `bind -M insert alt-left prevd-or-backward-word` works anyways 🤔 

I appreciate that this PR might not be the solution you strive for, and I guess as users it's easy enough to get the old bindings back 🙏 

It would be nice to make this a bit clearer in the release notes though, as it took me a second to get to the bottom of this :)

krobelus:
A relevant question is whether an application running in a terminal may have key bindings that depend on the terminal OS; this question can been seen as independent from current implementation details. There are arguments for both ways.

krobelus:
well then we're on the same page.
This PR is not a snap proposal and we do not need to rush it.
I think the chicken-and-egg problem ("only kitty") is very minor. I intentionally did not bother other terminal emulator developers until we know we have a concrete use case. Case in point -- we don't have consensus yet.

On the release branch I'd also prefer a revert to a half-baked solution but I don't super care.

`shift-delete` is another regression on macOS. I don't know if it's relevant.

faho:
378f452eaa0abbe1a6e3d41963adfcee11803a09 reverts the `*-token*` bindings to their old state.

Technically we could have kept alt-delete because that wasn't bound before, but I thought it weird to have *one* token-moving binding in there.

I hadn't really paid close attention to this, to me it seems like it could supersede the "bigword" behavior - which we barely used. I'm not entirely clear why it's a third movement style instead of just a change to bigword, but I know changing keybindings is annoying because people get particular about their muscle memory.

-----------------

Regarding the feature at hand, in 4.1:

The only use we would ever have for a client-OS query is bindings. With anything else I can come up with  you're interested in the OS that runs fish, not the terminal (and that's easy to get, via `uname` - tho I've played with the idea of removing that dependency by introducing a variable or `status` subcommand).

Or does anyone know any *other* feature that we would want to gate on where the terminal runs?

krobelus:
replacing uses of bigword with token would be a good change I think, but regardless we want better bindings than shift-arrows  (shift is not suitable for multiple reasons as explained in the issue)

krobelus:
I think `test (uname) = Darwin` is favored over the status quo, so if we don't want the rest,
I'd add that for now.

Using the client OS makes more sense to me but ultimately I don't care;
macOS users should (eventually) decide which behavior they want inside
`ssh`. Not sure if we want to work around tmux; not doing so seems okay.

`status is-in-macos-terminal` sure is awkward but that's easily remedied by replacing it with
something like `status xtgettcap uname`[^1]. Then any user can debug it based on `bind alt-left`
output. They may see the behavior vary when inside a plain terminal, tmux, mosh etc.

> The only use we would ever have for a client-OS query is bindings. With anything else I can
> come up with you're interested in the OS that runs fish, not the terminal (and that's easy
> to get, via uname - tho I've played with the idea of removing that dependency by introducing
> a variable or status subcommand).
>
> Or does anyone know any other feature that we would want to gate on where the terminal runs?

Can't think of any -- I believe mainly because many features (such as mouse/clipboard support)
already require support by the terminal[^2]. This means we can usually ask the terminal if it has
a feature.

But the "native key bindings" feature is different, it doesn't depend
on the terminal.  Given that a lot of platform-dependent keystrokes are
intercepted by the terminal and not sent to fish, it's a small feature.
But the differences can be jarring and it feels somewhat odd to decide for one behavior
when native bindings are opposites.

[^1]: `status xtgettcap` wants to be a builtin instead of an external command, else we might
      drop keyboard input that immediately precedes the XTGETTCAP response).
[^2]: `termios(3)` seems to list exceptions - it allows fish to talk to the terminal on the client

krobelus:
@ridiculousfish I don't know if your concerns about using the client OS are purely due to this awkward/opaque interface (which can be fixed) or if it's also about the behavior.

hameerabbasi:
As a macOS user, this would be a large QoL improvement for me.

hameerabbasi:
This was superceded by 2bb5cbc95943b0168c8ceb5b639391299e767e72 and can be closed.

PR #11113, opened 2025-02-01
Fix twice tokenize editor_cmd

phanen:
```fish
export VISUAL='nvim --cmd let\ g:flatten_wait=1'
funced -s fish_prompt
```

`editor_cmd[3]` would be `let` rather than `let g:flatten_wait=1`

### DIFF
--- a/share/functions/funced.fish
+++ b/share/functions/funced.fish
@@ -20,7 +20,7 @@ function funced --description 'Edit function definition'
     if set -q _flag_interactive
         set editor fish
     else if set -q _flag_editor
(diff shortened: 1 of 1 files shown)


ridiculousfish:
Nice find. Looks like the double-escape was introduced in 5f459446a8f20a179b339021083d549cf2e3fd99 , so this is a 4.0.0 regression.

ridiculousfish:
Cherry-picked to Integration_4.0.0 as bf91da5979df4e5741503c402338902970e4c0df

Issue #11108, opened 2025-01-31
Terminal queries should have a timeout


faho:
Since commit 64859fc24281d5d97b7f30dcf20a3a0fb80b4e72, we block on startup waiting for a reply to our terminal queries.

One of these is the "Primary Device Attribute" query, which is reasonably widely supported, which is why we haven't heard of reports of hangs yet.

But it *does not have to be* supported, and the failure mode is that fish just hangs indefinitely.

This should have a timeout.

To test, try:

```patch
diff --git a/tests/pexpect_helper.py b/tests/pexpect_helper.py
index 85c0bab6c..40347926b 100644
--- a/tests/pexpect_helper.py
+++ b/tests/pexpect_helper.py
@@ -175,8 +175,8 @@ class SpawnedProc(object):
         )
         self.spawn.delaybeforesend = None
         self.prompt_counter = 0
-        if env.get("TERM") != "dumb":
-            self.spawn.send('\x1b[?123c') # Primary Device Attribute
+        # if env.get("TERM") != "dumb":
+        #     self.spawn.send('\x1b[?123c') # Primary Device Attribute
 
     def time_since_first_message(self):
         """Return a delta in seconds since the first message, or 0 if this is the first."""
diff --git a/tests/pexpects/set_color.py b/tests/pexpects/set_color.py
index 2534ef3c9..2cbd505b3 100644
--- a/tests/pexpects/set_color.py
+++ b/tests/pexpects/set_color.py
@@ -19,7 +19,7 @@ expect_prompt()
 # See that --print-colors prints the colors colored.
 # Note that we don't check *all* of them, just a few.
 sendline("set_color --print-colors")
-expect_str("black")
+expect_str("black", timeout=120)
 expect_str("blue")
 expect_str("brblack")
 expect_str("brblue")
```

This will hang for two minutes before giving up on the pexpect side. That's because pexpect, by default, does not reply to the device attribute query.

So we already know one case that *will* break.

mqudsi:
I wrote a fix for this then realized I hadn't updated master. 64859fc24281d5d97b7f30dcf20a3a0fb80b4e72 that changes all the blocking waits to be hard-coded to `None`/`false` https://github.com/fish-shell/fish-shell/commit/64859fc24281d5d97b7f30dcf20a3a0fb80b4e72#diff-b9baeb92d8b45fbfa98ffd05d1e39d8453a791729c4f297aafe78852005d5a47R1513-R1522

PR #11117, opened 2025-02-02
Add a timeout to terminal queries

mqudsi:
The default timeout is currently set to two seconds; I would like to bring this down to something closer to 200ms if that is feasible based off tty response times and taking into account SSH and similar.

I removed a few panics that could have been triggered simply by tty race conditions, hopefully replacing them with no-op alternatives.

`try_pop()` does not reset the wait upon timeout to give a chance for other code to more gracefully react to the timeout (i.e. by clearing support capabilities or similar) to ensure that the terminal state remains consistent and we do not repeatedly trigger the same timeout.

---

**This does not _yet_ actually solve the problem enough for the modified tests in #11108 to pass.**

Because everything is run by the event loop, characters input *after* the timeout will trigger the timeout handling code added in this PR, but the asynchronous nature of the event loop means that if input were queued _before_ the timeout elapsed, they still need *some event* to occur after in order for the timeout handler to be triggered.

We could theoretically queue a thread to raise an alarm after the global timeout is reached when we first start blocking waiting for an async response (but we'd have to update the wait code to check because currently we ignore spurious waits) or we can register a self-pipe that receives an input (or a semaphore that is set) from another thread after the timeout elapses to nudge the read loop into handling the timeout.

If the general format of this change is a step in the right direction (i.e. the addition of a `since` field and changing `is_blocked()` and co to return an error on timeout) then this can be merged before it diverges too much from master and the timeout can be added separately. But if there's a cleaner way of doing this (i.e. locally calling the read loop but with a timeout as a hybrid synchronous/asynchronous approach) then that should be preferred.

(Separate note: I think I would update this PR to remove the timeout handling from the same sites that trigger the blocking wait because the odds of them seeing the error are very slim, instead just have them handle the success case and handle all the timeout fallouts (such as setting kitty supported to false) at the start of the main loop in `try_pop()` where `CURSOR_POSITION_REPORTING_SUPPORTED` is currently set to false.)

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -7,8 +7,9 @@ use crate::flog::FLOG;
 #[allow(unused_imports)]
 use crate::future::IsSomeAnd;
 use crate::input_common::{
-    BlockingWait, CharEvent, CharInputStyle, CursorPositionWait, ImplicitEvent, InputData,
-    InputEventQueuer, ReadlineCmd, ReadlineCmdEvent, READING_BUFFERED_INPUT, R_END_INPUT_FUNCTIONS,
+    BlockReason, BlockingWaitExpired, CharEvent, CharInputStyle, CursorPositionWait, ImplicitEvent,
+    InputData, InputEventQueuer, ReadlineCmd, ReadlineCmdEvent, READING_BUFFERED_INPUT,
+    R_END_INPUT_FUNCTIONS,
 };
 use crate::key::ViewportPosition;
 use crate::key::{self, canonicalize_raw_escapes, ctrl, Key, Modifiers};
@@ -462,19 +463,23 @@ impl<'a> InputEventQueuer for Reader<'a> {
         )));
     }
 
-    fn is_blocked(&self) -> bool {
-        self.blocking_wait.is_some()
-    }
-    fn unblock_input(&mut self) -> bool {
-        if !self.is_blocked() {
-            return false;
-        }
-        self.blocking_wait = None;
-        true
+    fn is_blocked(&self) -> Result<bool, BlockingWaitExpired> {
+        self.blocking_wait.is_blocked()
     }
 
-    fn blocking_wait(&self) -> Option<&BlockingWait> {
-        self.blocking_wait.as_ref()
+    fn unblock_input(&mut self) -> bool {
+        match self.is_blocked() {
+            Ok(false) => false,
+            Ok(true) | Err(_) => {
+                self.blocking_wait.reset();
+                true
+            }
+        }
+    }
+
+    fn blocking_wait(&self) -> Result<Option<&BlockReason>, BlockingWaitExpired> {
+        self.is_blocked()?;
+        Ok(self.blocking_wait.reason())
     }
 
     fn on_mouse_left_click(&mut self, position: ViewportPosition) {
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -82,7 +82,9 @@ use crate::history::{
 };
 use crate::input::init_input;
 use crate::input_common::kitty_progressive_enhancements_query;
+use crate::input_common::BlockReason;
 use crate::input_common::BlockingWait;
+use crate::input_common::BlockingWaitExpired;
 use crate::input_common::Capability;
 use crate::input_common::CursorPositionWait;
 use crate::input_common::ImplicitEvent;
@@ -521,7 +523,7 @@ pub struct ReaderData {
     /// The representation of the current screen contents.
     screen: Screen,
 
-    pub blocking_wait: Option<BlockingWait>,
+    pub blocking_wait: BlockingWait,
 
     /// Data associated with input events.
     /// This is made public so that InputEventQueuer can be implemented on us.
@@ -1176,7 +1178,7 @@ impl ReaderData {
             last_flash: Default::default(),
             flash_autosuggestion: false,
             screen: Screen::new(),
-            blocking_wait: Some(BlockingWait::Startup(Queried::NotYet)),
+            blocking_wait: BlockingWait::new(BlockReason::Startup(Queried::NotYet)),
             input_data,
             queued_repaint: false,
             history,
@@ -1413,8 +1415,8 @@ impl ReaderData {
         cursor_position_wait: Option<CursorPositionWait>,
     ) {
         if let Some(cursor_position_wait) = cursor_position_wait {
-            assert!(self.blocking_wait.is_none());
-            self.blocking_wait = Some(BlockingWait::CursorPosition(cursor_position_wait));
+            self.blocking_wait
+                .set(BlockReason::CursorPosition(cursor_position_wait));
         }
         let _ = out.write_all(b"\x1b[6n");
         self.save_screen_state();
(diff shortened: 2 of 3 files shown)


src/input.rs
```diff
@@ -7,8 +7,9 @@ use crate::flog::FLOG;
 #[allow(unused_imports)]
 use crate::future::IsSomeAnd;
 use crate::input_common::{
-    BlockingWait, CharEvent, CharInputStyle, CursorPositionWait, ImplicitEvent, InputData,
-    InputEventQueuer, ReadlineCmd, ReadlineCmdEvent, READING_BUFFERED_INPUT, R_END_INPUT_FUNCTIONS,
+    BlockReason, BlockingWaitExpired, CharEvent, CharInputStyle, CursorPositionWait, ImplicitEvent,
```
krobelus, line 10: I forgot that we currently hang when the terminal is either Midnight Commander
or `dvtm`, and the same terminal is running `ssh` / `docker exec` with fish inside.
Is it this use case that we're seeking to fix, or is it about others like `pexpect`?

src/reader.rs
```diff
@@ -2497,16 +2508,26 @@ impl<'a> Reader<'a> {
...
+                        }
+                        Err(_) => {
+                            // Unexpected timeout: ignore. Safe to reset at this point
+                            self.blocking_wait.reset();
+                            return ControlFlow::Continue(());
+                        }
                     };
                     match stage {
-                        Queried::NotYet => panic!(),
+                        Queried::NotYet => return ControlFlow::Continue(()),
```
krobelus, line 2530: this code is very bad but I think this panic was correct.
`Queried::NotYet` is only set initially, and changed before we read anything from stdin that could reach here.
We should get rid of the `Queried::NotYet` enum variant and use a separate static bool instead, to reduce scope.

src/reader.rs
```diff
@@ -3786,20 +3808,28 @@ impl<'a> Reader<'a> {
...
-                    self.request_cursor_position(
-                        &mut Outputter::stdoutput().borrow_mut(),
-                        Some(CursorPositionWait::ScrollbackPush),
-                    );
-                    return;
-                };
-                match wait {
-                    BlockingWait::Startup(_) => panic!(),
-                    BlockingWait::CursorPosition(_) => {
+                match self.blocking_wait() {
```
krobelus, line 3811: (this panic seemed correct too; we do not execute any readline command until
we're either done waiting for startup response or we have canceled it via ctrl-c or timeout)

src/input.rs
faho: Any of them.

It is a failure mode that shouldn't exist in general.

Hanging and crashing just shouldn't happen, period.

krobelus: Thanks for keeping us on our toes.

Speaking to the failure cases (not the failure mode just yet), because regardless of whether
we want to add a timeout, we should try to address these.

In general, interactive fish assumes it's connected to a terminal.  I don't think we want
to support arbitrary environments that are not terminals, but of course, we can add specific
workarounds for concrete environments that have worked historically.

There are some sources online claiming that "all" terminals respond to Primary DA, for example
[`notcurses`](https://github.com/dankamongmen/notcurses/blob/da7151929d9f87a73cabb4c5b54668b3a42cfcaf/TERMINALS.md?plain=1#L35-L37(). I think people would agree that we should require this from any future terminals.)

Let's look at all the concrete examples of potential breakage:

1. The Midnight Commander scenario is not actually affected.
   It might be possible that there is a scenario where `SHELL` is a wrapper for `ssh -t host
   fish`, but so far I haven't found that. I'll try again later.
2. `dvtm` has a proposed fix, in addition to our workaround based on `$TERM`.
   Until the fix is rolled out, something like `TERM=xterm dvtm` or `dvtm ssh -t -o SetEnv=TERM=xterm host fish`
   will hang until ctrl-c on startup. My feeling was that at least the latter is a rather uncommon scenario
   (usually you'd want to run dvtm on the server, not the client); dvtm was proposed for the
   "we don't care if it works" tier (I personally do care but maybe not to the extent of supporting
   edge cases where it's not possible to workaround with either a faithful `$TERM` or `XTVERSION`).

3. Expect/`pexpect` are special; they are not terminal emulators.

Both make a child fish's `isatty(STDIN_FILENO)` return true, even in the cases when the child fish
does not have access to a real terminal. Same goes for `script`.

Consider this Expect script that runs fish and tries to exit with `ctrl-d` --
which does not currently work because we're waiting for the primary DA response.

    #!/usr/bin/expect -f
    spawn fish
    sleep 1
    send "\x04" # ctrl-d
    expect {
        -re .+ {
            puts -nonewline $expect_out(buffer)
        }
    }
    close
    wait

When this above script is run **without a terminal in context** (for example
with Kakoune `kak -e 'exec !./fish.expect<ret>'` not yet sure what's the standard Unix way to do this),
our queries like `\x1b[0c` will be echoed and end up in the output of the expect script.
Naturally, fish will hang waiting for Primary DA response until it's HUP'd.

When the above script is run **inside a terminal**, Expect will relay fish's queries to the
enclosing terminal. **However**, Expect does not seem to relay the response back to fish,
hence fish hangs the same way.

    $ ./fish.expect
    spawn fish
    ^[[?0u^[[3;1R^[[?2026;2$y^[[?62;4;22;28c⏎

The above means that -- whether we have a timeout or not -- Expect scripts driving fish
may have had broken output ever since fish added querying (4.1).
Of course broken output is way less likely to cause actual breakage than hanging,
but in theory there would be new potential breakage everytime we add any query.
Not something that would be easy to support.

Perhaps we should look into an alternative fix (ideally one that doesn't
make expect script slower or include this garbage output). I wonder if we
can detect Expect/`pexpect` somehow but haven't found that yet.  OTOH the
current behavior is the most honest..

To help inform our decision, let's try to find existing Expect scripts running fish.  I have
searched on grep.app and github but no luck so far.  Alternatively an example script could work too.
My idea is to check if those expect scripts can tolerate the queries added in fish 4.1
and thus will realistically benefit from a timeout.

---

I haven't had time to comment on the failure mode yet.
FWIW [`libvaxis`](https://github.com/rockorager/libvaxis) (cc @rockorager) uses a 1 second timeout, and `notcurses` will hang forever.
I'm curious to research how our other peers address this.

faho: >  Of course broken output is way less likely to cause actual breakage than hanging,

Yes, that's the important point.

I don't think hanging indefinitely is an acceptable way of failing, at all.

Broken output on a terminal, i.e. seeing "=0u" isn't *great*, but since that's a terminal bug it would be something I could live with.

But I do not want fish to hang on startup with no user-discernable cause. The user would basically have no way to find out why it hangs unless they file a bug with us.

I do not expect *anyone* to come up with the `self.spawn.send('\x1b[?123c')` workaround on their own.

> but in theory there would be new potential breakage everytime we add any query.

Expect-style scripts will have to deal with programs writing output depending on the terminal. That's the nature of the thing.

Having the output change when the client's version changes is a fact with these. You will have to update these from time to time.

> To help inform our decision, let's try to find existing Expect scripts running fish.

Enumerating cases of breakage is fine when the breakage is a minor graphical glitch, not when it's a hang.

> FWIW [libvaxis](https://github.com/rockorager/libvaxis) (cc @rockorager) uses a 1 second timeout, and notcurses will hang forever.

fish isn't notcurses. It's a shell, and as such will be your main entry point to a machine. It will be the first thing run in an expect-style thing, it will be the first thing that runs over ssh.

If your text editor hangs, kill it and try a different one. If your *shell* hangs over ssh? No such luck. (and it's not obvious that your terminal is at fault, so there's no obvious thing to replace)

fish should be more resilient than the typical terminal program to hanging.

> Perhaps we should look into an alternative fix (ideally one that doesn't
make expect script slower or include this garbage output)

If you have one, I'm all ears. The only thing I can come up with is to stop querying again - which may mean some graphical glitches in buggy terminals.

Other than that, waiting for a second is preferable to hanging indefinitely - and it would be even if all the terminals we know supported the Primary DA query.

We could print an error if the timeout elapses so people know their terminal is at fault.

faho: > dvtm has a proposed fix

Unless you know of a fork I'm not sure that's ever going to be accepted - dvtm [hasn't had a commit in 4 years](https://github.com/martanne/dvtm/), and a release in 9.

krobelus: Regarding the failure mode: for the expect-style use case a timeout sounds like a reasonable compromise.
I'm in favor of also printing an error when the timeout elapses - that should reduce the pressure to make the timeout super low.
(If we use a long timeout *and* no error message, that could lead to a situation like with old tmux, where the high default escape delay was very annoying to unsuspecting users -- but if we tell the user what's going on they can safely `ctrl-c` out).
If we know that there is a human in front of the keyboard, printing an error on timeout but still waiting for `ctrl-c` would be an option, but unfortunately we don't know that.

Issue #15, opened 2012-05-30
Internalized scripts is far too aggressive


ridiculousfish:
The internalize scripts build phase internalizes far too much, including fish_prompt. We should measure to see what benefit, if any, is being derived from it, and restrict it to only those functions that really ought not to be changed (if any).

PR #11143, opened 2025-02-07
Read functions and completions directly from the fish binary

faho:
# Description

This builds on the previous "installable" builds that embedded the assets (functions, completions, man pages etc)
and wrote them out on start or whenever you ran `fish --install`, and turns them into "embed-data" builds.

These will just read the assets directly from the binary when needed. 

When autoloading, these are considered *last* (except for completions, they're considered before the generated completions), so overriding them via $fish_function_path or $fish_complete_path works like before.

That means, for users, this creates a fully capable single-file fish that does not depend on any other files on the filesystem. Link it statically and you can copy it wherever.

## Internals

Some scripts use the new `status list-files` and `status get-file` commands to get the files.
The `fish --install` option is removed - if you need to extract the files you can easily script that with the status commands.

- Webconfig is written to a temporary directory where it can be opened by `python`
- The manpage completion generator is extracted to a temporary directory when you run `fish_update_completions`.
  This is not done automatically at first start to avoid creating directories. Tbh the manpage completions aren't all that useful anymore.
- funced knows to load these files
- our `man` wrapper and `builtin --help` write out the man pages when needed

As before, the way to do this is to build with the on-by-default "embed-data" feature (which is disabled by cmake, so nothing changes for packagers).

# Why?

The idea is that you can copy a single fish binary to a system and run it without having to touch that system's files or care about keeping them up-to-date.

This also makes fish more resilient to misinstallation - there is no more data directory to keep in-sync, uninstallation is just removing the binary. Completions in other packages can no longer conflict with our completions.

It is no longer possible to break a lot of fish by setting $fish_function_path wrong.

# Compatibility concerns

$__fish_data_dir is set to empty. This is to avoid embed-data builds reading outdated assets from old installable builds and because it is no longer a useful directory.

That means scripts that go through that directory, e.g. to load another file, are affected will need to be changed to use the status commands.
I've done that for our things (and hopefully got everything), but third-party plugins would also have to be adjusted.

We could also keep $__fish_data_dir *set* to the old directory, 
but that would not help - if anything does `source $__fish_data_dir/completions/git.fish` that breaks if there is no git.fish there,
and not having a git.fish there is the point. So there is necessary breakage, and at that point it's easier to make both versions work if the directory is empty.

This concern is why I'm not pushing for this to become the default right now - it's still "experimental" so peope have time to adjust their plugins.

An additional point is that $__fish_data_dir also contains the vendor directories. If we did make this the default also for packaged builds, we would have to do something about that. I'm not entirely sure what, one possibility is to always check /usr/share/fish and /usr/local/share/fish, but that can obviously run into problems if the versions diverge (new fish on an LTS system).

Installable builds never read the vendor directories so nothing changes in that respect.

# Future

This could be the default way to build fish, and it could also be used to remove cmake.

The way to build fish for a package could be

```fish
# Configure this one with environment variables like $PREFIX, where necessary.
cargo build --release
# html docs can only be built after fish_indent is built
# we could make this easier by making it find fish automatically and using the indent builtin
FISH_BUILD_VERSION=$(target/release/fish --version) PATH="$PWD/target/release:$PATH" sphinx-build doc_src -M html
cargo test
python3 tests/test_driver.py target/release/fish -f /tmp

# create vendor directories, __fish_build_paths.fish, maybe the man pages separately
# copy the files wherever needed to make a package
```

As before, embed-data builds do not contain html documentation (because we need fish_indent to build them) or translations (because gettext needs them in a file).

`help` will open the online docs and translations aren't very useful at the moment.

# TODO

- [x] Check docs for fish_data_dir
- [x] Another pass over functions/completions
- [ ] Add a feature flag so plugins can do `if status test-feature embed-data` instead of trying `status list-files`?
- [ ] Or make the status commands transparently open files when needed?
- [ ] Rename the status commands to be more explicit about *which* files they touch? "status list-fish-asset-files"? "status list-embedded-files"?
- [ ] CHANGELOG

Proposed CHANGELOG wording:

The "self-installable" builds have been changed to read their datafiles directly from the fish binary instead of having to write them out on start. The `--install` option was removed, instead there are more granular `status list-files` and `status get-file FILE` commands that can be used to get at these files.

That means it is now possible (combined with the changes to make fish_key_reader and fish_indent builtins) to build fish as a single file that you can copy to a compatible system and just run it, including as a different user. As before, this is the default when building fish via cargo (e.g. `cargo install`).

We're considering making this the default in future because it makes it easier to install fish.

Note to developers of fish integration or plugins: These builds will have $__fish_data_dir set to empty because it would not contain any files. If you want to get a file, consider falling back to `status get-file path/to/file` (where "path/to/file" would be the path relative to the datadir, for instance "completions/git.fish") or if you want to load other completions just running them with `complete -C"foo" >/dev/null`.

<details>
<summary> the old description </summary>
## Description

~~This is a rough draft. In particular, for some reason I have not figured out, `cargo test` gets stuck at completing `echo (ls ` .~~

~~If anyone sees why that is (it works in actual usage and the scripted tests!), any help would be appreciated.~~ (edit: [figured it out](https://github.com/fish-shell/fish-shell/pull/11143#issuecomment-2766031066), it's a bug that's uncovered by this and now worked around)

Anyway, what does it do?

The worst part of self-installable fish currently is that it needs to write out its functions and completions before being in any sensible meaning of the word "functional". Without installing you've got a paper (well, bit-) weight.

So this will instead read these files right from the binary, if necessary.

So if you try to autoload e.g. the fish_prompt function, it will go through the regular $fish_function_path lookup, and if it doesn't find anything, load the version of fish_prompt.fish that was embedded in the binary. Same for completions.

Note that this is not entirely the same thing as currently, when using $__fish_data_dir - it is *always* last, and cannot be skipped. Not being skippable is probably fine, we might want to add a special "embedded" entry for this that you can add wherever you want (or just interpret the data dir as that). There's a hack to skip the generated completions.

Installing now only writes out the docs and webconfig (with the themes!). It skips functions and completions entirely.

## TODOs:

- [x] This isn't gated on the "installable" feature for now
- [x] Some things still need to be installed - webconfig, the man pages etc. This no longer prompts on start to install them. Is that okay?
- [x] `builtin --help` could be made to work by reading the docs from the binary and piping them to `groff`
- [x] What about $__fish_data_dir? It's no longer really a thing for $fish_function_path
- [x] We could make embedding the functions and completions the default configuration even via cmake - no real need to have files on disk we need to keep in sync.
- [x] Aforementioned hang
- [x] Cleanup
- [x] Tests

</details>


faho:
- [ ] Completing `functions ` doesn't work completely because it'll just check $fish_function_path

faho:
I think the hang might be pre-existing.

It hangs doing

```rust
completions = do_complete(
    L!("echo (ls test/complete_test/testfil"),
    CompletionRequestOptions::default(),
);
```

And specifically and *only* that one. Now, `ls` has a completion script, which is loaded, and that calls `ls`, which is also a function, which is also loaded.

Only... without embedding the functions, these aren't ever loaded because $fish_complete_path and $fish_function_path aren't actually defined here, so it will use the default file completions. That's also why, when you comment out that specific one, some other test_complete tests will fail.

When the functions are embedded, it hangs in running an external command - the first would be `uname`, but comment that out and it hangs running `command ls` (inside the ls function). It leaves defunct processes, so it seems to not be reaping them correctly from inside the `cargo test` environment.

zanchey:
We did try this in about 2013, IIRC, though the motivation was performance rather than portability.

faho:
Yup, it's #15 - which was disabled in fishfish beta 2 (which later became fish 2.0).

The big downside there was that "internalized" functions could not be overridden.

This solution doesn't have that problem - it does the normal lookup, and only if that fails it'll load the embedded scripts. That means it always runs as if $fish_function_path/$fish_complete_path had the embedded scripts as the last entry (with the one exception of generated_completions, which I'll grant is a hack).

I don't believe performance will really be measurably better - we still look at all the paths (including the empty $__fish_data_dir path for now), and we still need to convert the embedded scripts to wstring.

The benefit I'm excited by really *is* that it makes fish easier to install and more resilient - there's no more "what if someone set $fish_function_path wrong or erased /usr/share/fish", there's no more issue with `make install` leaving outdated scripts.

Which would be cool even if you install fish via a package manager, which is why I think this could be a default.

zanchey:
Yeah, that makes senses, and it's a great idea

faho:
Alright, we have a few places where we kinda wanna get the file from a function. Both `__fish_print_help` and `fish_delta` could benefit from this. I'm not sure how to do the latter without, and the former would be tricky.

So I am tempted to add a `status get-file` command that will print these files - `status get-file functions/fish_prompt.fish` would print the default fish_prompt.fish.

This would be weird if it were possible to not have embedded files at all, because either it would print files from the filesystem (in which case we've built a `cat`), or it wouldn't be available in some configurations.

faho:
Okay, I now made it so it asks to update the extracted assets if they exist. That means we allow an install *without* (e.g. just copying fish onto a machine that you ssh to), but if they are used we prevent them from going stale.

I don't know if that still requires confirmation, we could just write them out when needed.

faho:
I added two special commands:

- `status get-file`, to get a single embedded file's contents
- `status list-files`, to list embedded files (starting with a string - this could also do wildcard matching)

These allow us to make a lot of things work without extracting:

- `fish_config` (the CLI version) (prompts TODO)
- `__fish_print_help` (and thereby `builtin --help`)
- `fish_delta`
- `funced` - if we didn't have `status get-file` it couldn't correctly edit e.g. fish_git_prompt (which has multiple functions)

That means the only remaining features that need files to be extracted are:

- webconfig
- the man page completion generator

That's because these have multiple files that need to be passed to external tools. webconfig has a bunch of js, the completion generator consists of two python files.

We could potentially extract webconfig on-demand, make the man generator a single file and remove `fish --install` entirely. In that case webconfig would need to be taught to call fish to print embedded files.

`status` isn't really the *ideal* place to put these subcommands but I don't know a better one.

Remaining TODOs:

- [ ] fish_config prompt
- [ ] docs for status get-file/list-files

faho:
Alright, made webconfig extract on demand.

faho:
Test builds are available at https://github.com/faho/fish-shell/actions/runs/14088378224, but of course you can also build this with

```fish
cargo build --git https://github.com/faho/fish-shell.git --branch embed-functions
```

or with a git clone.

faho:
I've renamed the feature to "embed-data", because "installable" is now misleading.

Nothing otherwise changes - cmake turns all default features off, embed-data is a default feature, so just `cargo install` will have it on.

faho:
Alright, found an awkward bug: It read all embedded files in interactive mode (i.e. job control on).

This manifested as a SIGTTOU when loading the `diff` function.

Which brings us to something to remember when testing this: The embedded files will only be read once it is *installed*. Otherwise the "running in build directory" logic takes over.

faho:
Alright, I finally figured out why the complete test gets stuck with this:

https://github.com/fish-shell/fish-shell/blob/bf65b9e3a74028533fdeb557294c6aadc3d63db9/src/tests/mod.rs#L105-L106

we reset *all* signals to the default handlers. This includes SIGCHLD, which means fish won't get notification that a child process exited and will wait for it indefinitely.

We could remove at least internally used signals from that list (or possibly only default specific signals like HUP and TERM?), but that would currently fail on the complete tests not running an interactive parser (and potentially other issues in other tests, I didn't run them without signal defaulting). Anyway, it's a different bug that's only unmasked by this, preserving the status quo by not loading any functions/completions from the unit tests is fine for now.

faho:
One small concern here is that this still reads $__fish_data_dir, so if you have a `fish --install` installation left over, it'll use outdated files.

The alternative is to set $__fish_data_dir to empty.

I'm not sure how many scripts that would break - mostly things that do "introspection" of fish's data (like the things I've fixed here, e.g. `funced`). I assume those things can be fixed eventually.

faho:
Okay, the only uses of $__fish_data_dir I can find all would either do nothing (like [tacklebox adding it to $fish_function_path](https://github.com/justinmayer/tacklebox/blob/2b729fbfd0597c7c2c7f711c64192283f3a2dd5b/tacklebox.fish#L144)) or be broken whether we set it to empty or not like [gitnow loading completions/git.fish](https://github.com/joseluisq/gitnow/blob/91bda1d0ffad2d68b21a1349f9b55a8cb5b54f35/functions/__gitnow_load_git_functions.fish#L5-L7)).

So we might as well set it to empty to avoid loading stale files.

faho:
To fix: Completions that load other completions for the functions, like https://github.com/fish-shell/fish-shell/blob/d234ebf484958f32999c22f0cbbd79894ff36f68/share/completions/cargo.fish#L35 or nvim.

This could be a helper function or they could just do `complete -C"foo " &>/dev/null`.

And making __fish_data_dir would cause an issue in our own config: https://github.com/fish-shell/fish-shell/blob/d234ebf484958f32999c22f0cbbd79894ff36f68/share/config.fish#L38-L40 (this could be quoted, but that would then check "/__fish_build_paths.fish", which is unlikely to succeed but still wrong, so it should use `path is` instead).

Cargo.toml
```diff
@@ -91,9 +91,9 @@ name = "fish_key_reader"
 path = "src/bin/fish_key_reader.rs"
 
 [features]
-default = ["installable"]
+default = ["embed-data"]
```
krobelus, line 94: My `cargo install --path . --debug && ~/.cargo/bin/fish` tries to source files from `/usr/share/fish/functions`
This is because I have this hack in my config:

	set -a fish_complete_path /usr/share/fish/vendor_completions.d /usr/share/fish/completions
	set -a fish_function_path /usr/share/fish/vendor_functions.d/ /usr/share/fish/functions

of course the added "/usr/share/fish/functions" is wrong, I'm not sure why I added it

I think the vendor_completions.d used to be necessary for running installations to ~/.local,
but they should not be necessary if I set `XDG_DATA_DIRS`.. probably there's a reason I still have it

Cargo.toml
```diff
@@ -91,9 +91,9 @@ name = "fish_key_reader"
...
 benchmark = []
-installable = ["dep:rust-embed"]
+embed-data = ["dep:rust-embed"]
```
krobelus, line 96: > This concern is why I'm not pushing for this to become the default right now - it's still
   > "experimental" so people have time to adjust their plugins.

Yeah it would be nice if we can directly notify the most important plugins
(or better, give them fixes, even if they're untested).
I haven't checked how many are affected.

   > Installable builds never read the vendor directories so nothing changes in that respect.

It seems `~/.local/share/fish/vendor_completions.d` is still in `fish_complete_path` when I run an embed-data build with:

	env -u XDG_DATA_DIRS HOME=$(mktemp -d) fish

so I assume it's not completely ignored?

build.rs
```diff
@@ -41,14 +41,27 @@ fn main() {
...
+        let sec1dir = targetman.join("man1");
+        let _ = std::fs::create_dir_all(sec1dir.to_str().unwrap());
+    }
+
     rsconf::rebuild_if_paths_changed(&["src", "printf", "Cargo.toml", "Cargo.lock", "build.rs"]);
+
+    // These are necessary if built with embedded functions,
+    // but technically only in release builds (because debug builds read from the filesystem).
+    #[cfg(feature = "embed-data")]
+    rsconf::rebuild_if_paths_changed(&["doc_src", "share"]);
```
krobelus, line 63: Is rust-embed not usable in debug builds?
Maybe avoid rebuilds for debug builds then:

	#[cfg(all(feature = "embed-data"), debug_assertions)]

doc_src/cmds/status.rst
```diff
@@ -31,6 +31,8 @@ Synopsis
     status features
     status test-feature FEATURE
     status buildinfo
+    status get-file FILE
+    status list-files FILE
```
krobelus, line 35: maybe


```suggestion
    status list-files [PATH]
```

doc_src/cmds/status.rst
```diff
@@ -106,6 +108,15 @@ The following operations (subcommands) are available:
     This prints information on how fish was build - which architecture, which build system or profile was used, etc.
     This is mainly useful for debugging.
 
+**get-file** *FILE*
```
krobelus, line 111: `show-file` would be an alternative name

   > Rename the status commands to be more explicit about which files they touch? "status list-fish-asset-files"? "status list-embedded-files"?

maybe we shouldn't even call them file because to the user, they might not be files.
But something like `list-resource` is not better.

Actually, it looks like we mainly use this for man pages and webconfig.
Maybe we can be more concrete like `status list-man-pages` and `status show-man-src`.
But I'm not sure if that's worth it; it might be nice if we could
do it, and avoid exposing paths like `share/completions` but I
guess we should have a way to allow users to see the contents of
`share/completions/*.fish`, load it in a custom way etc.
I guess `status show-completions-src abbr` would abstract the `share` part..

share/functions/__fish_print_help.fish
```diff
@@ -11,7 +11,7 @@ function __fish_print_help --description "Print help message for the specified f
     end
 
     # Do nothing if the file does not exist
-    if not test -e "$__fish_data_dir/man/man1/$item.1" -o -e "$__fish_data_dir/man/man1/$item.1.gz"
+    if not path is -- $__fish_data_dir/man/man1/$item.1 $__fish_data_dir/man/man1/$item.1.gz; and not status get-file man/man1/$item.1 >/dev/null
```
krobelus, line 14: people may be surprised that we're stat-ing paths like "/man/man1/abbr.1",
if we really care we can add `set -q __fish_data_dir[1]`.
(AFAICT __fish_data_dir[1] is only set when turning off the embed feature)

src/autoload.rs
```diff
@@ -54,26 +96,103 @@ impl Autoload {
     /// After returning a path, the command is marked in-progress until the caller calls
     /// mark_autoload_finished() with the same command. Note this does not actually execute any
     /// code; it is the caller's responsibility to load the file.
-    pub fn resolve_command(&mut self, cmd: &wstr, env: &dyn Environment) -> Option<WString> {
+    pub fn resolve_command(&mut self, cmd: &wstr, env: &dyn Environment) -> Option<AutoloadPath> {
+        use crate::autoload::AutoloadPath;
```
krobelus, line 100: this import statement seems redundant

src/autoload.rs
```diff
@@ -54,26 +96,103 @@ impl Autoload {
...
+
+        // HACK: In cargo tests, this used to never load functions
+        // It will hang for reasons unrelated to this.
+        #[cfg(test)]
+        return None;
+
+        #[cfg(feature = "embed-data")]
+        {
+            let narrow = wcs2string(cmd);
+            let cmdstr = std::str::from_utf8(&narrow).ok()?;
```
krobelus, line 142: okay rust embed doesn't support non-UTF-8 paths. We can probably live with that.

src/bin/fish.rs
```diff
@@ -440,61 +302,6 @@ fn fish_parse_opt(args: &mut [WString], opts: &mut FishCmdOpts) -> ControlFlow<i
             'f' => opts.features = w.woptarg.unwrap().to_owned(),
             'h' => opts.batch_cmds.push("__fish_print_help fish".into()),
             'i' => opts.is_interactive_session = true,
-            'I' => {
```
krobelus, line 443: Need to remove this from the optspecs as well ("fish --install" is crashing with unexpected retval from WGetopter)

Cargo.toml
krobelus, line 94: ah so this is the default for `cargo` commands (making it easy to test) but cmake builds pass `--no-default-features`.

krobelus (approved):

Cargo.toml
faho: The installable builds never read the vendor stuff, yes.

I'm not sure if it is a good idea to do that or not, because these builds could be much newer than what the vendor stuff is prepared for.

For now, I'm explicitly punting on that for later - if we end up making this style of build the default, we'll have to handle it, but until then it's the status quo.

Cargo.toml
faho: That's $XDG_DATA_HOME.

faho: Ah, no, it's $__fish_user_data_dir.

We *really* added too much guff here.

doc_src/cmds/status.rst
faho: > Actually, it looks like we mainly use this for man pages and webconfig.
Maybe we can be more concrete like status list-man-pages and status show-man-src.

See e.g. `funced`.

They're used more for man pages and webconfig and stuff, but they can be used for everything and I prefer having something general over making special commands for every kind of resource.

> avoid exposing paths like share/completions

The path is "completions/...", no "share". I think that's fine because it also groups them by type.

share/functions/__fish_print_help.fish
faho: > people may be surprised that we're stat-ing paths like "/man/man1/abbr.1",

We aren't - the variable is an empty list, the entire argument is eliminated.

src/autoload.rs
faho: Yeah, we would have to name one of our functions with non-utf8, and I think we can just decide not to do that.

Cargo.toml
faho: Okay, I don't know what to do with that one. I missed it being introduced.

We can either set it to empty too, or keep it (it's in your home dir, so it's more likely that it's up-to-date and usable).

share/functions/fish_config.fish
```diff
@@ -268,12 +303,23 @@ function fish_config --description "Launch fish's web based configuration"
...
+                            end
                         end
 
-                        while read -lat toks
+                        set -l content
+                        if string match -qr '^tools/' -- $file
+                            set content (status get-file $file)
+                        else
+                            read -az content < $file
+                        end
```
krobelus, line 320: something here must be broken. Visual mode selections are no longer highlighted. I'm  using

	target/debug/fish -C 'set -g fish_key_bindings fish_vi_key_bindings'

Confirmed that "set -S fish_color_selection" still shows "-r"

faho: How are you setting the theme?

When I do that I get my regular selection color and it works.

faho: I'm also not sure how that should be caused by this PR - it does not touch any of the highlighting, so if the variable is set I don't know how it should affect it.

faho: Ah, okay: It needs to be `read -z`, without the `-a`.

Note that that means you would have an *empty* fish_color_selection, not "-r". So presumably you have a universal set to "-r", and are doing "fish_config theme choose None", and that sets a global to empty.

PR #11145, opened 2025-02-09
Complete entries for bootctl

ModProg:
## Description

Adds completion support for entries in `bootctl set-oneshot` and `set-default`. Uses `jq` to parse and doesn't add these extra completions if not available.

## TODOs:


ridiculousfish:
Thank you

PR #11150, opened 2025-02-12
Curses to term

zanchey:
Now that curses isn't actually used, this PR renames the module to term and removes an unneeded lock.

@krobelus I hope this doesn't mess up too much WIP?


krobelus:
Good change.  Note that we also plan to stop querying the file-based terminfo database.  The new
name is appropriate for that.

As a general rule, I avoid abbreviations in globally-visible names, (so I'd probably
change `term` to `terminal`) but it's not a big deal, it's still unambiguous here I think.

krobelus (approved):

zanchey:
OK, I renamed the module to terminal, although I left the name of the Term struct the same to avoid churn

Issue #10633, opened 2024-07-27
Let's release Rust-based fish


ridiculousfish:
# Let's do it

Here's a pass at what remains to be done to enable the first release of Rust-based fish. I hope this will be a checklist to drive the release forwards.

tl;dr: resolve blocking issues, perform internal testing, pair a beta release with a blog post, resolve issues found in the beta, then a proper release.

## Current state

fish "finished" the Rust port months ago; however fish was not then in a release-ready state, due to gaps in packaging and dependencies on forked crates. Since then the port has matured greatly and we are now close.

But I think our first release should be an open beta.

## Case for a beta release

When we set out to do the port, we expected that we would introduce regressions which might not be discovered immediately. That's what happened! We fixed what was found, and there's possible remaining regressions which we have yet to find.

fish has also accumulated many changes unrelated to the port since the last major release; these may also have introduced bugs.

It would bad if we were to release and then major bugs are discovered, prompting us to do multiple patch releases. It risks breaking users' systems, would be a large time sink, and would reflect poorly on the fish project. While this is always a risk, it's a larger risk for this upcoming release because of the port.

A beta will allow us to discover these issues before a release, through volunteers.

## Beta process

We should pair the beta with a blog post discussing the port and inviting users to test it, with the goal of drumming up interest in fish.

- [x] Draft the blog post discussing the port and inviting users to join the beta
- [x] On Linux, provide install instructions (what should they be? Use our OBS builds?)
- [x] On macOS, we can provide a `fish-beta` Homebrew formula, and the standard installer/app

Ideally we provide instructions to back out should users experience breakage. We also can provide instructions for reporting issues (probably just through GitHub).

## Bug fixes
We'll use the [fish next-3.x](https://github.com/fish-shell/fish-shell/milestone/36) milestone to track blocking bugs.

## Dependencies

We have one major dependency whose fate is unclear:

- [ ] pcre2 crate fork
    <details>
    <summary>What to do about our fork of pcre2 crate?</summary>
    fish uses a forked version of the pcre2 crate, with UTF-32 support added. Currently this is fetched via git. Some options:
    
    - Clean up and publish our fork on crates.io as `pcre2-utf32`. This is viable because we've [already gotten it to PR ready state](https://github.com/BurntSushi/rust-pcre2/pull/38). I'm leaning towards this.
    - Vendor the forked crate, stripping out what we don't need.
    - Abandon the fork and round trip through UTF-8 for regex matching. Simplest but may hurt performance.

[...]

PR #11172, opened 2025-02-26
Fix `create-base-directories.fish` with SELinux

lnicola:
## Description

Untested, see https://github.com/fish-shell/fish-shell/issues/10633#issuecomment-2686020746.

## TODOs:


lnicola:
Asking here to spam fewer people, was that the last test failure?

Conan-Kudo:
No. This was the only other one: https://github.com/fish-shell/fish-shell/pull/11173

zanchey:
weird that this doesn't happen in the Fedora builders

lnicola:
> weird that this doesn't happen in the Fedora builders

I couldn't find those, but if they're containers running on a host with SELinux disabled?

PR #11173, opened 2025-02-26
tests/checks: version: Relax regular expression for valid versions

Conan-Kudo:
This relaxes the regular expression determining valid version strings such that packaging version strings used by Fedora and other Linux distributions can be successfully embedded.

This is related to #10633 work.

### DIFF
--- a/tests/checks/version.fish
+++ b/tests/checks/version.fish
@@ -1,2 +1,2 @@
 #RUN: %fish -v
-# CHECK: fish, version {{[-.gabcdeflphat0-9]*(irty)?}}
+# CHECK: fish, version {{[-.abcdefghilpt0-9~\^]*(irty)?}}


zanchey:
> packaging version strings used by Fedora

Can you give me some context? How are these being used/embedded?

Conan-Kudo:
Sure. The version of fish being packaged right now that includes this patch is `4.0.0~beta.1^git1.b82d0fcbcc4`. This gets embedded in via the `version` file so that when you ask fish what version it is, it reports the exact one that is used for the packaging itself.

faho:
At that point, I think we could just make it something like `fish, version {{[0-9].*}}` - just see that the version starts with a digit.

We added that test because packagers managed to build a package that said "fish, version unknown" IIRC.

faho:
Alright, I made it just require a leading digit.

PR #11175, opened 2025-02-27
Fix tag in README

mo8it:


mo8it:
The installation works. It tells you that you installed the beta version, but `fish --version` shows

```
fish, version 4.0.0-beta.1-geb336889b
```

Which means that it compiles from the release commit although it shows beta.

PR #11190, opened 2025-02-28
Orphan background tasks to work around terminals being sensitive to unreaped processes

### LINKED ISSUE #11181: Problematic behavior with macOS Terminal related to `disown` in `fish_git_prompt.fish`
When starting a new Fish shell in Terminal, for example by opening a new tab with Command+T, the title bar will indicate that `bash` is running and trying to close the tab will show the "Closing this tab will terminate the running process bash." warning.

Running `ps` in this state shows the following:

```
$ ps
  PID TTY           TIME CMD
25211 ttys000    0:00.03 -fish
25219 ttys000    0:00.00 <defunct>
```

When the process disappears seems to be a bit random, usually it disappears when pressing `enter`, sometimes it disappears after a short while without it.

I am admittedly not sure what exactly is happening here, but after some troubleshooting, I found that commenting out this line:

https://github.com/fish-shell/fish-shell/blob/a9c61394a89686a247a07b0d21f231a333e6609d/share/functions/fish_git_prompt.fish#L184

seems to fix the issue.

I also tried the following:

* Replacing `fish_prompt` with a stub function fixes/avoids the issue.
  * Running `fish_git_prompt` manually can still trigger it.
* I had an LLM add a bunch of `echo` commands throughout `fish_prompt`, which also "fixed" it.
   * Probably by slowing down the function, maybe the new blazingly-fast Rust code brought this issue to the surface?

**Edit:** I added a very short `sleep` to the end of `fish_prompt` and it indeed "fixed" the issue similar to the `echo`s I mentioned above.

**Edit 2:**

```
$ fish -v
fish, version 4.0.0
$ which fish
/opt/homebrew/bin/fish
```

Issue started after the update to `4.0.0`.

krobelus:
When a command like `long-running-command &` exits, the resulting SIGCHLD
is queued in the topic monitor. We do not process this signal immediately
but only after e.g. the next command has finished. Only then do we reap the
child process.

Some terminals, such as Terminal.app, refuse to close when there are unreaped
processes associated with the terminal -- as in, having the same session ID,
see setsid(3).

In future, we might want to reap proactively.

For now, apply an isolated workaround: instead of taking care of a child
process, double-fork to create an orphaned process. Since the orphan will
be reaped by PID 1, we can eventually close Terminal.app without it asking
for confirmation.

	/bin/sh -c '( "$@" ) >/dev/null 2>&1 &' -- cmd arg1 arg2

This fix confines the problem to the period during which a background process
is running. To complete the fix, we would need to call setsid to detach the
background process from a controlling terminal. That seems to be desirable
however macOS does provide a setsid utility.

	setsid cmd arg1 arg2 >/dev/null 2>&1

Fixes #11181

### DIFF
--- a/share/functions/__fish_apropos.fish
+++ b/share/functions/__fish_apropos.fish
@@ -40,8 +40,7 @@ if test $status -eq 0 -a (count $sysver) -eq 3
 
         if test $age -ge $max_age
             test -d "$dir" || mkdir -m 700 -p $dir
-            /usr/libexec/makewhatis -o "$whatis" (/usr/bin/manpath | string split : | xargs realpath) >/dev/null 2>&1 </dev/null &
-            disown $last_pid
+            /bin/sh -c '( "$@" ) >/dev/null 2>&1 </dev/null &' -- /usr/libexec/makewhatis -o "$whatis" (/usr/bin/manpath | string split : | xargs realpath)
         end
     end
 else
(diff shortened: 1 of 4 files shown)


krobelus:
I botched the initial PR description, it's fixed now

PR #11195, opened 2025-03-01
Change `gettext` paths to be relocatable

hameerabbasi:
## Description

See https://docs.conda.io/projects/conda-build/en/stable/resources/define-metadata.html#detect-binary-files-with-prefix and https://github.com/conda-forge/fish-feedstock/pull/49#issuecomment-2691968880.

## TODOs:


faho:
Try this patch instead:

```diff
diff --git a/src/wutil/gettext.rs b/src/wutil/gettext.rs
index ab968ed34f..78cfad2149 100644
--- a/src/wutil/gettext.rs
+++ b/src/wutil/gettext.rs
@@ -1,4 +1,5 @@
 use std::collections::HashMap;
+use std::ffi::CStr;
 use std::ffi::CString;
 use std::sync::Mutex;
 
@@ -48,7 +49,7 @@ use internal::*;
 // Really init wgettext.
 fn wgettext_really_init() {
     let package_name = CString::new(PACKAGE_NAME).unwrap();
-    let localedir = CString::new(env!("LOCALEDIR")).unwrap();
+    let localedir = CStr::from_bytes_until_nul(env!("LOCALEDIR").as_bytes()).unwrap_or_default();
     fish_bindtextdomain(&package_name, &localedir);
     fish_textdomain(&package_name);
 }
```

Conda's patching here is pretty broken and would ideally be fixed there. I don't know if this sort of thing won't appear again in other places.

hameerabbasi:
To me, right fix would just be to not hardcode paths into the binary. 😅 But I'm not entirely sure how to do that refactor.

krobelus:
There are various other uses of `env!`, like
`const DOC_DIR: &str = env!("DOCDIR");`,
I wonder if they are affected too / why not. If they are, they should crash too unless we're using an unsafe API that doesn't go through CString..

> To me, right fix would just be to not hardcode paths into the binary.

I'm not 100% it's what you would expect but we already have a native solution for that, that seems to work well in practice. For example

`cargo install --path . --debug && echo yes | fish --install`

I'm not sure if `src/gettext.rs` uses the same "relocatable" logic as `determine_config_directory_paths()`; we may want to fix that.

faho:
> To me, right fix would just be to not hardcode paths into the binary

These paths are for system integration. They exist for packagers to say "this is where we put our localizations", like other paths (that may have the same issue) exist to say "this is where we have our system configuration" and "this is where we keep our system data".

The right fix would be for conda to set these paths correctly. These are set *at compile time*, so they should just be compiled correctly.

(tho there are a bunch of reasons we should move away from gettext imho, and LOCALEDIR is one of them)

----

Anyway, try this one:

```diff
diff --git a/src/wutil/gettext.rs b/src/wutil/gettext.rs
index ab968ed34f..a083c48619 100644
--- a/src/wutil/gettext.rs
+++ b/src/wutil/gettext.rs
@@ -48,7 +48,11 @@ use internal::*;
 // Really init wgettext.
 fn wgettext_really_init() {
     let package_name = CString::new(PACKAGE_NAME).unwrap();
-    let localedir = CString::new(env!("LOCALEDIR")).unwrap();
+    let localedir = env!("LOCALEDIR");
+    let localedir = CString::new(match localedir.chars().position(|c| c == '\0') {
+        Some(nul_pos) => &localedir[..nul_pos],
+        None => localedir,
+    }).unwrap();
     fish_bindtextdomain(&package_name, &localedir);
     fish_textdomain(&package_name);
 }
```

that's basically common::truncate_at_nul, except for &str.

If there are any other places where we need this, it should be extracted into a function (or the one in common made generic or whatever).

hameerabbasi:
> These paths are for system integration. They exist for packagers to say "this is where we put our localizations", like other paths (that may have the same issue) exist to say "this is where we have our system configuration" and "this is where we keep our system data".

That works well if the paths can be determined reliably at compile time. However, there are distributions and package managers(such as NixOS and conda) that distribute binaries compiled ahead of time, but the prefix isn't known until install-time.

Would you be willing to propose an alternative solution for such systems? To me, the fix is simply to compute these paths exactly once at runtime rather than at compile time based of a data directory.

faho:
> Would you be willing to propose an alternative solution for such systems? To me, the fix is simply to compute these paths exactly once at runtime rather than at compile time based of a data directory.

The only real alternative is to use a path relative to the fish binary, like we do for the data files (functions and such). That's what krobelus was talking about above - using the "relocatable" logic.

There isn't really a way outside of fish to set an environment variable or similar, because the shell is often the first thing launched, and distributions tend to not configure fish.

Also, to be entirely honest, the translations are barely useful at this point anyway (because they simply are not nearly complete enough) and the best solution is to get off gettext to something that includes the strings in the binary directly.

So for now, "not crashing" should be fine.

hameerabbasi:
Okay; I've done the following changes:

1. Moved the config paths into its own module in `fish::env`.
2. Added `locale` to `ConfigPaths`.
3. Used the `trim_nul` function to get rid of null chars in all paths in the binary.
4. Introduced a lazy static variable `CONFIG_PATHS`.

faho:
Okay that's a bit much for a patch release.

What bit do you actually *need* other than the LOCALEDIR thing?

hameerabbasi:
> Okay that's a bit much for a patch release.

I'll make another PR for a patch to 4.0.0 with just the `LOCALEDIR` fix -- everything else seems to be working in my tests.

krobelus:
is there a particular reason you don't want to patch this downstream? E.g. the minimal fix;
for upstream I'd rather have a complete fix.

But it seems a bit hacky. Where do we draw the line?
Canonical solutions are "recompile it" or "add a runtime option", I'm not sure why that's not possible here.

> everything else seems to be working in my tests.

we're probably lucky because fish builtins truncate arguments at the first
NUL byte (to match external programs).

But things may be subtly broken:

```
$ cargo b --no-default-features && target/debug/fish -c 'set -S __f
ish_help_dir'
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.06s
$__fish_help_dir: set in global scope, unexported, with 1 elements
$__fish_help_dir[1]: |/usr/share/doc/fish\x00\x00\x00\x00\x00\x00|
```

this is with

```diff
diff --git a/src/env/environment.rs b/src/env/environment.rs
index e4722029bd..807d4f226f 100644
--- a/src/env/environment.rs
+++ b/src/env/environment.rs
@@ -25,6 +25,7 @@
 use crate::wchar::prelude::*;
 use crate::wcstringutil::join_strings;
 use crate::wutil::{fish_wcstol, wgetcwd, wgettext};
+use std::str::FromStr;
 use std::sync::atomic::Ordering;
 
 use libc::{c_int, confstr, uid_t, STDOUT_FILENO, _IONBF};
@@ -644,10 +645,11 @@
         );
 
         if !cfg!(feature = "installable") {
+            let fake_path = std::path::PathBuf::from_str("/usr/share/doc/fish\0\0\0\0\0\0").unwrap();
             vars.set_one(
                 FISH_HELPDIR_VAR,
                 EnvMode::GLOBAL,
-                str2wcstring(paths.doc.as_os_str().as_bytes()),
+                str2wcstring(fake_path.as_os_str().as_bytes()),
             );
         } else {
             vars.set_empty(FISH_HELPDIR_VAR, EnvMode::GLOBAL);
```

hameerabbasi:
I'm open to both possilities -- this PR currently is more of a "complete fix". For downstream; I've patched just `LOCALEDIR`: https://github.com/conda-forge/fish-feedstock/pull/49

hameerabbasi:
If translations aren't useful and `gettext` removal is planned; we can even drop the dependency in future builds.

> Until then, conda should probably stop binary-patching fish and ensure it's used as a "relocatable" build, which uses paths relative to the binary.

Is there a way to ensure this during build? That's exactly what this PR tries to achieve, other than the `trim_nul` part (which I can remove if necessary). `gettext` init was the only path still using the compile-time env var; which is now removed in this PR IIUC.

hameerabbasi:
For context, most of this logic was in `fish.rs` which was inaccessible to the `gettext.rs` file -- hence the move to another file.

faho:
> if someone cares enough we can probably add compile-time options to make fish trivially relocatable.

it mostly *is* - it'll find the datafiles via "../share".

The issue is that it's pretty easy to break (move the fish binary, use it on an OS that doesn't give you the binary path, possibly call it via a symlink?), and there's currently no way to *disable* the absolute paths (I think? Maybe if you pass them as empty?).

Also it doesn't apply to the translations, which this is about (and translations are currently not embedded either).

hameerabbasi:
> move the fish binary

Unfixable indeed.

> use it on an OS that doesn't give you the binary path, possibly call it via a symlink

`argv[0]` is indeed empty on some systems. https://github.com/fish-shell/fish-shell/pull/11195/commits/ee183a089a85a90096ed86e953e05c9c2f3f8490 should hopefully fix that and follow symlinks as well.

krobelus:
> The issue is that it's pretty easy to break (move the fish binary,

I think `make install` has the same problem.
We probably already print warnings on startup if we're missing some important files.

> use it on an OS that doesn't give you the binary path,

this can be a compile error (the relocatable feature need not be active by default)

> possibly call it via a symlink?)

this should be fine because the binary path will always be the link's destination

> Also it doesn't apply to the translations, which this is about (and translations are currently not embedded either).

why? Is it because translations are not in `share/fish`?
Maybe something like `(status fish-path)/../share/locale` could work..

hameerabbasi:
> why?

It's read directly from a compile-time env var currently, which is a path encoded into the binary. This PR tries to make it use the same logic as the other paths.

faho:
> this can be a compile error (the relocatable feature need not be active by default)

First I would like to know who it is *for* - because we have had it get in the way in #10748 and I couldn't get an answer from @ridiculousfish as to how it is used on macOS et al.

(the main system where it won't work is OpenBSD, IIRC)

src/env/config_paths.rs
```diff
@@ -0,0 +1,174 @@
...
+
+const DOC_DIR: &str = env!("DOCDIR");
+const DATA_DIR: &str = env!("DATADIR");
+const DATA_DIR_SUBDIR: &str = env!("DATADIR_SUBDIR");
+const SYSCONF_DIR: &str = env!("SYSCONFDIR");
+const BIN_DIR: &str = env!("BINDIR");
+
+pub static CONFIG_PATHS: Lazy<ConfigPaths> = Lazy::new(|| {
+    // Read the current executable and follow all symlinks to it.
+    determine_config_directory_paths(std::env::current_exe().unwrap().canonicalize().unwrap())
```
faho, line 20: This would crash fish on startup, unconditionally, on platforms that don't support std::env::current_exe().

faho: https://github.com/rust-lang/rust/issues/60560 - apparently it currently silently returns a wrong path, but should probably fail instead. OpenBSD downstream applies a patch to rust, so if you use the distro's rust compiler it will produce a crashing fish.

We have our own way to resolve the binary, and it doesn't work on OpenBSD - #9086. It's what `status fish-path` uses.

faho:
Moving this out of the 4.0.1 milestone because it's patched downstream, and we're not sure what the solution here should be.

I think #11143 would solve this class of problem, except for for translations.

For that we would "just" have to get the translations included (and read directly from inside the binary instead of writing them out to a file), which probably means getting off gettext.

src/env/config_paths.rs
hameerabbasi: There seems to be very little we can do for OpenBSD in general, other than checking existence of the path if it is returned and falling back to `argv[0]` on failure.

hameerabbasi: Done in https://github.com/fish-shell/fish-shell/pull/11195/commits/fa02518c4e8b40a498c1d7bcc7403289129ab727

faho:
Okay, I'm inclined to merge this, mostly because it is good cleanup and makes getting the directories *elsewhere* easier.

I'm planning on printing them in `status buildinfo` so we can check how they're set in different circumstances.

Finding the locale the same way as our other directories seems like a good idea even if we *do* want to get off separate localization files in general and gettext specifically.

hameerabbasi:
That'd be much appreciated! The FreeBSD 13 build failure seems to be unrelated, it's not able to find the Rust compiler.

faho:
Thanks!

zanchey:
This broke translations completely as far as I can tell, I haven't dug into why yet

PR #11200, opened 2025-03-02
edit_command_buffer: support passing cursor position to helix

pvsr:
## Description

Detect both `hx`, the default binary name from upstream, and `helix`, the name used by at least [Arch](https://gitlab.archlinux.org/archlinux/packaging/packages/helix/-/blob/main/helix.install?ref_type=heads).

## TODOs:


krobelus:
would be better if they supported one of the standard formats

PR #11203, opened 2025-03-02
Fix kitty integration by passig  cmdline_url in the 133;C message

leiserfg:
## Description

Talk about your changes here.

Fixes issue #

## TODOs:

### DIFF
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -691,7 +691,10 @@ fn read_i(parser: &Parser) -> i32 {
         data.clear(EditableLineTag::Commandline);
         data.update_buff_pos(EditableLineTag::Commandline, None);
         // OSC 133 End of command
-        data.screen.write_bytes(b"\x1b]133;C\x07");
+
+        let cmd_url = escape_string(&command, EscapeStringStyle::Url);
+        let osc = format!("\x1b]133;C;cmdline_url={cmd_url}\x07");
+        data.screen.write_bytes(osc.as_bytes());
         event::fire_generic(parser, L!("fish_preexec").to_owned(), vec![command.clone()]);
         let eval_res = reader_run_command(parser, &command);
         signal_clear_cancel();


adamcstephens:
Link to the kitty report here: https://github.com/kovidgoyal/kitty/issues/8385

septatrix:
@krobelus

> It's debatable whether the shell or the terminal should provide this feature but I think we should fix this regression?

Having this done by the shell has the advantage that it also works when the terminal does not have access to the procfs (e.g. VMs or SSH). A more featureful standard for that is currently being initiated by systemd: [OSC 3008: Hierarchical Context Signalling](https://github.com/systemd/systemd/blob/main/docs/OSC-CONTEXT.md#osc-3008-hierarchical-context-signalling)

kovidgoyal:
On Mon, Mar 03, 2025 at 05:39:02AM -0800, Nils K wrote:
> septatrix left a comment (fish-shell/fish-shell#11203)
>
> @krobelus
>
> > It's debatable whether the shell or the terminal should provide this feature but I think we should fix this regression?
>
> Having this done by the shell has the advantage that it also works when the terminal does not have access to the procfs (e.g. VMs or SSH). A more featureful standard for that is currently being initiated by systemd: [OSC 3008: Hierarchical Context Signalling](https://github.com/systemd/systemd/blob/main/docs/OSC-CONTEXT.md#osc-3008-hierarchical-context-signalling)

It already works without access to procfs in VMs or SSH. I dont know
what usecase 3008 is supposed to solve, but it isn't this.

krobelus:
to correct myself, it definitely has to go through the terminal somehow; a shell inside SSH or VM can't call `notify-send` on the host

kovidgoyal:
On Mon, Mar 03, 2025 at 05:50:08AM -0800, Johannes Altmanninger wrote:
> krobelus left a comment (fish-shell/fish-shell#11203)
>
> to correct myself, it definitely has to go through the terminal somehow; a shell inside SSH or VM can't call `notify-send` on the host

The shell can implement it using OSC 99
https://sw.kovidgoyal.net/kitty/desktop-notifications/
which works regardless of SSH/VM.

It is implemented separately in kitty because the most commonly used
shells, bash/zsh are not likely to implement it anytime soon.

PR #11220, opened 2025-03-05
wgetopt: remove support for abbreviated long options. Closes #7341.

### LINKED ISSUE #7341: Don't allow abbreviated long options in getopt
I just noticed that our getopt implementation, taken from glibc and modified for wchar, allows unambiguous long option abbreviations.

That means `string match --reg` works even tho the option is `--regex`.

I don't think I have ever seen anyone use this on purpose.

Personally, I consider this to be a terrible design decision because it means that adding new long flags can break previous calls (e.g if we added `--regex-dialect` it would break all uses of `string match --reg`), and would like to remove it.

This is likely to be controversial, and the change would have to be backwards-incompatible (although, like I said, I have never seen anyone use this on purpose).

m1kc:
## Description

As per #7341, this PR removes support for abbreviated long options from wgetopt implementation — for example, `--forc` is no longer considered a valid form of `--force`. The PR removes inexact matching of long options altogether and adds a new test to make sure it doesn't come back somehow.

&nbsp;

Regarding documentation, I kinda failed to find the right words. For example,

```
- Support for for abbreviated long options has been removed, e.g. `--reg` is no longer considered
  a valid form of `--regex`.
```

is technically correct, but I'm worried that some users might overestimate the scope if it's phrased like that.

## TODOs:

- [ ] (AFAIK, support for abbreviated long options isn't mentioned anywhere, so maybe no changes required) Changes to fish usage are reflected in user documentation/manpages.

### DIFF
--- a/src/tests/wgetopt.rs
+++ b/src/tests/wgetopt.rs
@@ -41,3 +41,40 @@ fn test_wgetopt() {
     assert_eq!(arguments.len(), 3);
     assert_eq!(join_strings(&arguments, ' '), "emacsnw emacs -nw");
 }
+
+#[test]
+fn test_wgetopt_proper_longopts() {
+    // Test for issue #7341 (disable abbreviated long options, e.g. --forc shouldn't work as an alias to --force).
+    const short_options: &wstr = L!("-f");
+    const long_options: &[WOption] = &[wopt(L!("force"), ArgType::NoArgument, 'f')];
+    let mut argv = [L!("rm"), L!("--forc"), L!("filename1"), L!("filename2")];
+    let mut w = WGetopter::new(short_options, long_options, &mut argv);
+    let mut f_count = 0;
+    let mut unknown_count = 0;
+    let mut arguments = vec![];
+    while let Some(opt) = w.next_opt() {
+        match opt {
+            'f' => {
+                f_count += 1;
+            }
+            '\x01' => {
+                // non-option argument
+                arguments.push(w.woptarg.as_ref().unwrap().to_owned());
+            }
+            '?' => {
+                // unrecognized option
+                unknown_count += 1;
+                if let Some(arg) = w.argv.get(w.wopt_index - 1) {
+                    arguments.push(arg.to_owned());
+                }
+            }
+            _ => {
+                panic!("unexpected option: {:?}", opt);
+            }
+        }
+    }
+    assert_eq!(f_count, 0);
+    assert_eq!(unknown_count, 1);
+    assert_eq!(arguments.len(), 3);
+    assert_eq!(join_strings(&arguments, ' '), "--forc filename1 filename2");
+}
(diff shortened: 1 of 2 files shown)


m1kc:
Turns out I wasn't running all the tests, but that's not the worst part.

1) `string shorten` arguments is called `--chars`, but in the docs it's called `--char`. Both worked before, now it's broken.

```rust
impl<'args> StringSubCommand<'args> for Shorten<'args> {
    const LONG_OPTIONS: &'static [WOption<'static>] = &[
        // FIXME: documentation says it's --char
        wopt(L!("chars"), RequiredArgument, 'c'),
```

Moreover, at least two of fish's own tests use `--char` too. And I guess it's been in the wild for a quite a while — at this point it's probably not nice to suddenly make `--char` invalid because [people use that](https://github.com/search?q=string+shorten+--char+path%3A*.fish&type=code). 

2) Likewise, `function` accepts `--argument-names` but `tests/checks/function-definition.fish` uses `--argument` instead. And yes, [people use that too](https://github.com/search?q=function+--argument+path%3A*.fish&type=code).

I'm not sure what to do with all that.

faho:
Yeah, that's about why I hadn't really made progress on this so far either.

There are, as I see it, four ways we can go about it:

1. Abandon this plan and keep this weird misfeature indefinitely
2. Tough it out and fix our scripts, tell everyone else to fix theirs (the fix is always backwards-compatible)
3. Add a *warning* for when this is used and migrate slowly with a [future feature flag](https://fishshell.com/docs/current/language.html#future-feature-flags) - see #10365 
4. Add these specific options as alternative spellings (`string shorten` gets a "--char" option that works exactly like "--chars" - note that this will break "--ch" because it's now ambiguous)

I think I'm leaning towards 2, unless we can see this is used a lot in the wild and then 3.

m1kc:
Well, `--char` is _the only_ documented form right now, so it's reasonable to expect ~100% of scripts to use that. In fact, [nobody on GitHub](https://github.com/search?q=%22string+shorten+--chars%22+path%3A*.fish&type=code) uses `--chars`. At this point, it might even be easier to just rename the flag to `--char` (or at least support both forms for a very long time).

`--argument`, on the contrary, wasn't documented anywhere, so here you at least _can_ say "who even told you it works that way?". But sadly enough, this form also seems to be more popular than `--argument-names` (4.4k vs 1.7k if my search query is correct). Basically, this change is going to break scripts for two thirds of people who use functions with named arguments.

I'm also thinking about all the other possible cases that weren't caught by CI.

Overall, option 3 seems to be more reasonable, but I'm not a maintainer of a popular shell, so it's a bit hard for me to judge. To play the devil's advocate, fish isn't as popular as bash, so it's not going to break millions of scripts — and the required fixes aren't complicated and probably can be done pretty quickly even in large codebases.

Issue #11221, opened 2025-03-05
Crash (abort) doing completion on filenames >= 256 bytes


g2p:
This is possible on bcachefs (stat --file-system /mnt/bcachefs says 512 bytes in a dirent, https://github.com/koverstreet/bcachefs/issues/733 has a test script; not sure if that limit includes a final NUL).

Create the file like this:

    i=256; touch "$(printf "%05d" ${i})$(head -c $((${i}-5)) /dev/zero | tr '\0' 'a')"

The backtrace looks like (as of 49023adec738da36d017b3d1a8979512daff0237):

```
   5: fish::common::cstr2wcstring
             at ./src/common.rs:1185:18
   6: fish::wutil::dir_iter::DirIter::next
             at ./src/wutil/dir_iter.rs:282:27
   7: fish::wildcard::expander::WildCardExpander::expand_trailing_slash
             at ./src/wildcard.rs:667:41
   8: fish::wildcard::expander::WildCardExpander::expand
             at ./src/wildcard.rs:528:21
```

Apparently there are 256 byte slices in dirent structs returned from the libc crate: https://docs.rs/libc/latest/libc/struct.dirent.html
gdb can printf the rest of the string so the slice length is incorrect.
std's own read_dir will probably work better.

g2p:
Discussed here: https://github.com/rust-lang/libc/issues/2669

The stdlib now does this (before &raw there were complex macros):

                let name = CStr::from_ptr((&raw const (*entry_ptr).d_name).cast());
                let name_bytes = name.to_bytes();

And the alternative is to wrap stdlib and re-insert dotdot when needed.

g2p:
&raw needs Rust 1.82 and fish seems to limit itself to 1.70 though.
https://blog.rust-lang.org/2024/10/17/Rust-1.82.0.html#native-syntax-for-creating-a-raw-pointer

g2p:
Nix, which fish already uses, also has a [dir module](https://docs.rs/nix/latest/nix/dir/) that would simplify the wutil::dir_iter code.  std is more correct wrt unsafe code but doesn't expose the dirfd interface or rewinding.  And [rustix's Dir::read](https://docs.rs/rustix/latest/rustix/fs/struct.Dir.html#method.read) looks even better, it doesn't inherit problems from the libc crate's struct because the default backend does the syscalls directly.

faho:
[...]

PR #11224, opened 2025-03-05
Fix dirent->d_name usage

g2p:
Closes https://github.com/fish-shell/fish-shell/issues/11221

Tested manually with a 300 byte name.

### DIFF
--- a/src/wutil/dir_iter.rs
+++ b/src/wutil/dir_iter.rs
@@ -1,5 +1,5 @@
 use super::wopendir;
-use crate::common::{cstr2wcstring, wcs2zstring};
+use crate::common::{str2wcstring, wcs2zstring};
 use crate::wchar::{wstr, WString};
 use crate::wutil::DevInode;
 use libc::{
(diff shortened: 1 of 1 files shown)


src/wutil/dir_iter.rs
```diff
@@ -267,14 +266,16 @@ impl DirIter {
...
-        let d_name_cchar = &dent.d_name;
-        let d_name = unsafe {
-            slice::from_raw_parts(d_name_cchar.as_ptr() as *const u8, d_name_cchar.len())
-        };
+
+        // TODO: This should use &raw from Rust 1.82 on
+        // https://github.com/rust-lang/libc/issues/2669
+        // https://github.com/fish-shell/fish-shell/issues/11221
+        let d_name_ptr = addr_of!((*dent).d_name);
+        let d_name = unsafe { std::ffi::CStr::from_ptr(d_name_ptr.cast()) }.to_bytes_with_nul();
```
faho, line 274: It's probably worth it to explain what's going on here - we can't trust the length (because dirent lies, a long-standing kernel C convention), but we can trust that it is NUL-terminated,
with a link to the libc issue.

ridiculousfish:
Merged as a0dada561897be9d6c3ba76910b829e167a0c7fc - thank you!!

PR #11230, opened 2025-03-06
completions/flatpak: don't ignore packages with uppercase in name

phanen:
`[^A-Z]+` here seems to trim header section in old version flatpak (avoid one more `tail -n -2`?).
But this can ignore packages have uppercase name like `org.gtk.Gtk3theme.Adwaita-dark`.

```fish
$ flatpak remote-ls --columns=application,name | string replace -rf '^([^A-Z]+)(?: +|\t)(.*?)\s*$' '$1\t$2' | wc -l
1254

$ flatpak remote-ls --columns=application,name | string replace -rf '^([^A-Z].*?)(?: +|\t)(.*?)\s*$' '$1\t$2' | wc -l
5059
```

And I add a `__fish_print_flatpak_packages` since complete `flatpak install xx` is really slow.


ridiculousfish:
Thank you!

Issue #10956, opened 2024-12-28
Please support windows msys2


ahaoboy:
**YOUR TEXT HERE**

As a developer, I kindly request adding support for windows msys2, I understand that this is not part of the plan and might be extremely challenging, but as Windows developers, we desperately need a shell to replace pwsh. Currently, I use fish 3.7.x installed via msys2, and it works amazingly well—intelligent, fast, and efficient. If version 4.0 could also support windows msys2, it would be perfect.

![image](https://github.com/user-attachments/assets/ad1b4dc3-5f9f-4997-881b-a77ec45da917)

msys2-fish-3.7.1-3
https://packages.msys2.org/packages/fish?variant=x86_64

faho:
That's not up to us.

Unfortunately, there is no cygwin/msys target for rust, and so no way to compile fish for it.

We had hoped it would appear in the last 2 years, but it appears the efforts have seized. See https://github.com/rust-lang/rust/issues/79854.

So there is nothing we can do.

Aworldc:
Really? When I install msys2 and then normal windows rustup, it just works and I can use libraries installed in msys2, in rust, and compile that rust, without needing or using msvc or visual studio build tools, so I don't see why you'd need a special msys2 rust toolchain.

faho:
Rustup will give you the *windows* target, which means everything that isn't available for windows, like `std::os::unix::fs::MetadataExt`, won't be available.

We specifically need a unix target, which would be a toolchain for msys/cygwin/whatever, targetting its unix features.

Fish is a unix shell that relies on unix features.

faho:
If you try to compile for the x86_64-pc-windows-gnu target, you'll get some errors in our libc.c - which hooks up some libc constants that don't (yet) exist in the libc crate.

Stub those out and you get 3887 lines of errors, including:

- file descriptors ("unresolved import `std::os::fd`")
- file modes (mode_t/W_OK/X_OK/...)
- termios
- pgroups (tcsetpgrp/tcgetpgrp/setpgid)
- pid_t and uid_t
- `kill()`
- signals - SIGWINCH/SIGHUP/
- posix_spawn
- mmap
- readlink

Basically almost anything that talks to the OS would have to be changed. This touches 54 out of 185 .rs files.

That's not going to happen.

SerJaimeLannister:
Have we tried https://github.com/uutils/coreutils/ , this is written in rust , so maybe this can help ?

faho:
[...]

PR #11238, opened 2025-03-07
Support cygwin

Berrysoft:
## Description

This PR adds support for Cygwin & MSYS2.

* Cygwin `wchar_t` is UTF-16.
  * `wcwidth` is ~~replaced by `unicode-width`.~~ removed.
  * `wcrtomb` and `mbrtowc` are replaced by `c32rtomb` and `mbrtoc32`.
* The target is cross-compiled from Windows.
* There are some bugs on cygwin, so I have to add some workarounds.

Fixes issue #10956, though it has been closed.

Tracking:
- [x] https://github.com/rust-lang/rust/pull/137621
- [x] https://github.com/nix-rust/nix/pull/2610
- [x] https://github.com/lambda-fairy/rust-errno/pull/106

TODO:
- [x] Link libintl correctly
- [x] Link libpcre2 correctly

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -6,8 +6,12 @@ use std::error::Error;
 use std::path::{Path, PathBuf};
 
 fn main() {
-    setup_paths();
-
+    if cfg!(windows) {
+        // Cross compiling from Windows, keep the paths in Unix style.
+        rsconf::set_env_value("DATADIR_SUBDIR", "fish");
+    } else {
+        setup_paths();
+    }
     // Add our default to enable tools that don't go through CMake, like "cargo test" and the
     // language server.
 
--- a/build.rs
+++ b/build.rs
@@ -97,7 +97,12 @@ fn detect_cfgs(target: &mut Target) {
             Ok(target.has_symbol("localeconv_l"))
         }),
         ("FISH_USE_POSIX_SPAWN", &|target| {
-            Ok(target.has_header("spawn.h"))
+            if env::var("CARGO_CFG_TARGET_OS").unwrap() == "cygwin" {
+                // `POSIX_SPAWN_SETSIGDEF` is broken
+                Ok(false)
+            } else {
+                Ok(target.has_header("spawn.h"))
+            }
         }),
         ("HAVE_PIPE2", &|target| {
             Ok(target.has_symbol("pipe2"))
--- a/build.rs
+++ b/build.rs
@@ -97,12 +97,7 @@ fn detect_cfgs(target: &mut Target) {
             Ok(target.has_symbol("localeconv_l"))
         }),
         ("FISH_USE_POSIX_SPAWN", &|target| {
-            if env::var("CARGO_CFG_TARGET_OS").unwrap() == "cygwin" {
-                // `POSIX_SPAWN_SETSIGDEF` is broken
-                Ok(false)
-            } else {
-                Ok(target.has_header("spawn.h"))
-            }
+            Ok(target.has_header("spawn.h"))
         }),
         ("HAVE_PIPE2", &|target| {
             Ok(target.has_symbol("pipe2"))
--- a/build.rs
+++ b/build.rs
@@ -3,15 +3,11 @@
 use rsconf::{LinkType, Target};
 use std::env;
 use std::error::Error;
-use std::path::{Path, PathBuf};
+use std::path::Path;
 
 fn main() {
-    if cfg!(windows) {
-        // Cross compiling from Windows, keep the paths in Unix style.
-        rsconf::set_env_value("DATADIR_SUBDIR", "fish");
-    } else {
-        setup_paths();
-    }
+    setup_paths();
+
     // Add our default to enable tools that don't go through CMake, like "cargo test" and the
     // language server.
 
@@ -236,6 +232,11 @@ fn has_small_stack(_: &Target) -> Result<bool, Box<dyn Error>> {
 }
 
 fn setup_paths() {
+    #[cfg(unix)]
+    use std::path::PathBuf;
+    #[cfg(windows)]
+    use unix_path::{Path, PathBuf};
+
     fn get_path(name: &str, default: &str, onvar: &Path) -> PathBuf {
         let mut var = PathBuf::from(env::var(name).unwrap_or(default.to_string()));
         if var.is_relative() {
--- a/src/exec.rs
+++ b/src/exec.rs
@@ -12,6 +12,7 @@ use crate::common::{
     ScopeGuard,
 };
 use crate::env::{EnvMode, EnvStack, Environment, Statuses, READ_BYTE_LIMIT};
+#[cfg(FISH_USE_POSIX_SPAWN)]
 use crate::env_dispatch::use_posix_spawn;
 use crate::fds::make_fd_blocking;
 use crate::fds::{make_autoclose_pipes, open_cloexec, PIPE_ERROR};
@@ -34,10 +35,12 @@ use crate::null_terminated_array::{
     null_terminated_array_length, AsNullTerminatedArray, OwningNullTerminatedArray,
 };
 use crate::parser::{Block, BlockId, BlockType, EvalRes, Parser};
+#[cfg(FISH_USE_POSIX_SPAWN)]
+use crate::proc::Pid;
 use crate::proc::{
     hup_jobs, is_interactive_session, jobs_requiring_warning_on_exit, no_exec,
-    print_exit_warning_for_jobs, InternalProc, Job, JobGroupRef, Pid, ProcStatus, Process,
-    ProcessType, TtyTransfer,
+    print_exit_warning_for_jobs, InternalProc, Job, JobGroupRef, ProcStatus, Process, ProcessType,
+    TtyTransfer,
 };
 use crate::reader::{reader_run_count, restore_term_mode};
 use crate::redirection::{dup2_list_resolve_chain, Dup2List};
@@ -447,6 +450,7 @@ fn launch_process_nofork(vars: &EnvStack, p: &Process) -> ! {
 // To avoid the race between the caller calling tcsetpgrp() and the client checking the
 // foreground process group, we don't use posix_spawn if we're going to foreground the process. (If
 // we use fork(), we can call tcsetpgrp after the fork, before the exec, and avoid the race).
+#[cfg(FISH_USE_POSIX_SPAWN)]
 fn can_use_posix_spawn_for_job(job: &Job, dup2s: &Dup2List) -> bool {
     // Is it globally disabled?
     if !use_posix_spawn() {
--- a/src/exec.rs
+++ b/src/exec.rs
@@ -12,7 +12,6 @@ use crate::common::{
     ScopeGuard,
 };
 use crate::env::{EnvMode, EnvStack, Environment, Statuses, READ_BYTE_LIMIT};
-#[cfg(FISH_USE_POSIX_SPAWN)]
 use crate::env_dispatch::use_posix_spawn;
 use crate::fds::make_fd_blocking;
 use crate::fds::{make_autoclose_pipes, open_cloexec, PIPE_ERROR};
@@ -35,12 +34,10 @@ use crate::null_terminated_array::{
     null_terminated_array_length, AsNullTerminatedArray, OwningNullTerminatedArray,
 };
 use crate::parser::{Block, BlockId, BlockType, EvalRes, Parser};
-#[cfg(FISH_USE_POSIX_SPAWN)]
-use crate::proc::Pid;
 use crate::proc::{
     hup_jobs, is_interactive_session, jobs_requiring_warning_on_exit, no_exec,
-    print_exit_warning_for_jobs, InternalProc, Job, JobGroupRef, ProcStatus, Process, ProcessType,
-    TtyTransfer,
+    print_exit_warning_for_jobs, InternalProc, Job, JobGroupRef, Pid, ProcStatus, Process,
+    ProcessType, TtyTransfer,
 };
 use crate::reader::{reader_run_count, restore_term_mode};
 use crate::redirection::{dup2_list_resolve_chain, Dup2List};
@@ -450,7 +447,6 @@ fn launch_process_nofork(vars: &EnvStack, p: &Process) -> ! {
 // To avoid the race between the caller calling tcsetpgrp() and the client checking the
 // foreground process group, we don't use posix_spawn if we're going to foreground the process. (If
 // we use fork(), we can call tcsetpgrp after the fork, before the exec, and avoid the race).
-#[cfg(FISH_USE_POSIX_SPAWN)]
 fn can_use_posix_spawn_for_job(job: &Job, dup2s: &Dup2List) -> bool {
     // Is it globally disabled?
     if !use_posix_spawn() {
--- a/src/fallback.rs
+++ b/src/fallback.rs
@@ -8,10 +8,10 @@ use crate::{common::is_console_session, wchar::prelude::*};
 use errno::{errno, Errno};
 use once_cell::sync::Lazy;
 use std::cmp;
+use std::ffi::CString;
 use std::fs::File;
 use std::os::fd::FromRawFd;
 use std::sync::atomic::{AtomicIsize, Ordering};
-use std::{ffi::CString, mem};
 
 /// Width of ambiguous East Asian characters and, as of TR11, all private-use characters.
 /// 1 is the typical default, but we accept any non-negative override via `$fish_ambiguous_width`.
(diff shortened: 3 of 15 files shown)


src/fallback.rs
```diff
@@ -32,16 +32,25 @@ pub static FISH_EMOJI_WIDTH: AtomicIsize = AtomicIsize::new(1);
...
 
-    const _: () = assert!(mem::size_of::<libc::wchar_t>() >= mem::size_of::<char>());
+    const _: () = assert!(std::mem::size_of::<libc::wchar_t>() >= std::mem::size_of::<char>());
     let width = unsafe { wcwidth(c as libc::wchar_t) };
     isize::try_from(width).unwrap()
 }
 
+/// A safe wrapper around the system `wcwidth()` function
+#[cfg(target_os = "cygwin")]
+pub fn wcwidth(c: char) -> isize {
```
faho, line 48: What you're doing here is adding a dependency on a unicode width table. Fish already ships its own, so there's no need for a second one.

We call wcwidth in two places:

https://github.com/fish-shell/fish-shell/blob/90218821f9a373684b439abe5cf200dc467fa7fe/src/env_dispatch.rs#L200-L204 - which figures out if Unicode >= 9 is supported. That one can just be hardcoded as "2".

https://github.com/fish-shell/fish-shell/blob/90218821f9a373684b439abe5cf200dc467fa7fe/src/fallback.rs#L51-L53, as an attempt to support "the system console", i.e. something like linux' or freebsd's kernel VTs. That does not apply to Cygwin.

So this can be removed entirely.

src/fork_exec/spawn.rs
```diff
@@ -114,7 +114,10 @@ impl PosixSpawner {
 
         // Set our flags.
         let mut flags: i32 = 0;
-        flags |= libc::POSIX_SPAWN_SETSIGDEF;
+        // Bug of cygwin: `posix_spawn` reports error if `POSIX_SPAWN_SETSIGDEF` is set.
+        if cfg!(not(target_os = "cygwin")) {
+            flags |= libc::POSIX_SPAWN_SETSIGDEF;
```
faho, line 119: I would expect this means that signals aren't correctly reset in Cygwin, which could cause some problems.

The only solution I can imagine is to not use posix_spawn on cygwin.

faho (changes requested):

src/fork_exec/spawn.rs
Berrysoft: This flag makes cygwin trying to set handler for SIGINT, and cause errors. I think it never worked before in the C++ version. The legacy fork + exec works fine.

faho: As this code stands, cygwin could potentially use posix_spawn but not reset signals, which could then lead to issues in launched commands.

Unless setsigdef is somehow not needed for cygwin, that means posix_spawn should be disabled entirely for cygwin.

src/fallback.rs
```diff
@@ -32,16 +32,24 @@ pub static FISH_EMOJI_WIDTH: AtomicIsize = AtomicIsize::new(1);
...
+    // https://github.com/fish-shell/fish-shell/pull/11238#discussion_r1985414670
+    2
```
faho, line 50: You want a hardcoded width of 2 for `😃`, which is our test codepoint to see if the system knows Unicode 9.

You *don't* want this to return 2 for everything.

I think this function should not be compiled on cygwin at all - instead the 2 should be hardcoded in https://github.com/fish-shell/fish-shell/blob/90218821f9a373684b439abe5cf200dc467fa7fe/src/env_dispatch.rs#L200-L204

faho (changes requested):

src/fallback.rs
faho: And https://github.com/fish-shell/fish-shell/blob/90218821f9a373684b439abe5cf200dc467fa7fe/src/fallback.rs#L51-L53

should be stubbed out - cygwin doesn't have a "console session".

src/fallback.rs
Berrysoft: Seems that there's another usage:
https://github.com/fish-shell/fish-shell/blob/90218821f9a373684b439abe5cf200dc467fa7fe/src/fallback.rs#L75-L78
Should I hardcode 0 here?

faho: Yeah, that sounds right.

build.rs
```diff
@@ -227,6 +232,11 @@ fn has_small_stack(_: &Target) -> Result<bool, Box<dyn Error>> {
 }
 
 fn setup_paths() {
+    #[cfg(unix)]
+    use std::path::PathBuf;
+    #[cfg(windows)]
+    use unix_path::{Path, PathBuf};
```
faho, line 238: What's the need for unix_path here?

src/exec.rs
```diff
@@ -12,6 +12,7 @@ use crate::common::{
     wcs2zstring, write_loop, ScopeGuard,
 };
 use crate::env::{EnvMode, EnvStack, Environment, Statuses, READ_BYTE_LIMIT};
+#[cfg(FISH_USE_POSIX_SPAWN)]
 use crate::env_dispatch::use_posix_spawn;
```
faho, line 16: Do these *need* to be cfg'd out?

build.rs
Berrysoft: When cross compiling from windows, the paths need to keep unix style.

src/exec.rs
Berrysoft: Just to make the code free from warnings.

build.rs
faho: In what way? "/" vs "\"?

If this really is needed, the dependency should be made windows-specific in Cargo.toml.

Berrysoft: The original code tries to get the full path from the env vars, making the env vars containing "C:\" even "D:\" depending on the host machine.

I'll make the dep windows only later.

src/fd_monitor.rs
```diff
@@ -389,7 +389,8 @@ impl BackgroundFdMonitor {
             drop(data);
             let ret =
                 fds.check_readable(timeout.map(Timeout::Duration).unwrap_or(Timeout::Forever));
-            if ret < 0 && !matches!(errno().0, libc::EINTR | libc::EBADF) {
+            // Cygwin reports ret < 0 & errno == 0 as success.
+            if ret < 0 && !matches!(errno().0, libc::EINTR | libc::EBADF | 0) {
```
krobelus, line 393: can we gate the `|| errno == 0` workaround behind a `#[cfg(target_os = "cygwin")]`

build.rs
Berrysoft: I just realized that it's impossible to make `unix-path` windows-specific, because `[target.'cfg(windows)']` means the *target* is Windows, but not *host* is Windows. There's no way to specify a Windows-host only build dependency.

src/fd_monitor.rs
Berrysoft: Fixed.

build.rs
faho: I would be fine with just panic!ing if the passed paths start with "X:".

These aren't paths we get from *the system*, these are paths we're passed by *the build configuration*. It should just give them as unix versions.

Berrysoft: Yes, they are passed by the *build configuration*, but the env var `PREFIX` should be *absolute* according to the code below, and then all relative env vars will be set to absolute path according to `PREFIX`. If you just panic if the passed paths start with drive letter, but also panic if `PREFIX` is not absolute, it would be nearly impossible to compile for cygwin, because the most possible way to compile for cygwin is to cross-compile from Windows.
`unix-path` is used here to minimize and simplify the change of code. Could you explain why you would not like a build dependency that's even not compiled into the final binary?

faho: > the most possible way to compile for cygwin is to cross-compile from Windows.

Wouldn't you do something like `cmake -DPREFIX=/usr` or `PREFIX=/usr cargo build`?

> Could you explain why you would not like a build dependency that's even not compiled into the final binary?

Because it would make everyone on linux/macos/bsd and especially our packagers (debian wants to have one source package per rust dependency) install a dependency that they would never need.

If we absolutely have to okay, but I'd like to find a solution that doesn't impact everyone not using cygwin.

Berrysoft: I'm afraid we have to use it. `/usr` is not an absolute path on Windows, and `\` is not a valid path separator on Unix. Maybe someday there could be a cygwin-hosted rustc, and then we would be able to get rid of this dependency, but it's far from now.

Another possible solution is to create a new crate to handle the paths, which could decide dependency based on platform. It could be a stub crate that re-exports `std::path::*` on Unix and re-exports `unix_path::*` on Windows. We can reference it as a build-dependency and the packagers might be happier about that. (Might be, because anyway `unix_path` will be in the lock file.)

Berrysoft: Oh, there's another dirtier solution: make `unix_path` as cygwin target only build dependency.

faho: What's dirtier about that?

Berrysoft: We wish to find sth like `[host.'cfg(windows)'.build-dependencies]`, and it's roughly equivalent to `[target.'cfg(target_os = "cygwin")'.build-dependencies]`, because only cygwin target is cross-compiled from Windows. The dirty part is that it's actually possible to cross-compile to cygwin from Linux (I have never tried, but someone did). If you prefer this solution, it's OK to me.

faho: If you cross-compiled to cygwin from linux, wouldn't you have unix paths to begin with?

But yeah, that would sound alright.

Berrysoft: OK, I have set `unix_path` as cygwin only.

Berrysoft: Oh, maybe it's my fault. `[target.'cfg(windows)'.build-dependencies]` just works fine now...

faho (approved):
Looks fine by me, I think @ridiculousfish would like to take a look.

Awesome work, btw

src/fallback.rs
```diff
@@ -75,8 +75,16 @@ pub fn fish_wcwidth(c: char) -> isize {
...
-            wcwidth(c)
+            #[cfg(not(target_os = "cygwin"))]
+            {
+                // Fall back to system wcwidth in this case.
+                wcwidth(c)
+            }
+            #[cfg(target_os = "cygwin")]
+            {
+                // No system wcwidth for UTF-32 on cygwin.
+                0
```
riking, line 86: should this be Ambiguous instead of 0?

Berrysoft: They are 0 in cygwin. You can see in the following code that it returns 0 for non-spacing characters.

https://github.com/cygwin/cygwin/blob/main/newlib/libc/string/wcwidth.c

We don't use the cygwin `wcwidth` function because it's UTF-16.

ahaoboy:
Thanks for your excellent work. I tried to test it with the latest version(version 4.1.0-alpha0-g509826571) under msys2, but encountered a strange bug. Every time I type a character, a new line is created. I don't know whether I should open a new issue because cygwin does not officially support it yet.

fish

Berrysoft:
@ahaoboy Do you use `oh-my-posh`? It might be a bug of it, a think.

ahaoboy:
> @ahaoboy Do you use `oh-my-posh`? It might be a bug of it, I think. I haven't dig into the code of `oh-my-posh` so not that sure.


I use starship, which works fine on fish 3.7.1

https://packages.msys2.org/base/fish

```
 starship --version
starship 1.22.1
branch:master
commit_hash:d605196
build_time:2025-01-11 17:09:54 +00:00
build_env:rustc 1.84.0 (9fc6b4312 2025-01-07),stable-x86_64-pc-windows-msvc
```

Berrysoft:
OK, seems that the problem is a little complex. I'll look into it. You may try to disable starship temporarily and see if the problem is solved.

faho:
This is an issue where I would suspect the *terminal*, not the prompt.

Fish 4.0 is more demanding on the terminal, and that can trigger bugs.

This isn't a manifestation I've seen, but I know the most frequent *trigger* is the kitty keyboard protocol, so try `set -Ua fish_features no-keyboard-protocols` to turn that off (and restart fish after).

ahaoboy:
I removed all plugins and configurations, and only added a prompt configuration. The printf function will add a new line every 3 or 4 inputs.

I don't know if there are other possible reasons. It depends on whether other Windows users can reproduce the problem with the same configuration.

```
# ~/.config/fish/config.fish

set -Ua fish_features no-keyboard-protocols

function fish_prompt
    echo

    # This line of code will cause a newline to be added every 3 to 4 inputs.
    printf '%s@%s %s%s%s > ' $USER $hostname (set_color $fish_color_cwd) (prompt_pwd) (set_color normal)
end
```

Berrysoft:
Well, I can reproduce such problem with oh-my-posh, both with Windows Terminal and mintty. It might be some misc problems between fish and cygwin api and terminal...

faho:
> set -Ua fish_features no-keyboard-protocols

Don't put that in config.fish. It will cause the universal variable to grow.

It's enough to run it interactively. If you must have it in config.fish, use

```fish
contains -- no-keyboard-protocols $fish_features
or set -Ua fish_features no-keyboard-protocols
```

faho:
> It's adding a new line each second you input.

Terminal timestamps, see https://github.com/fish-shell/fish-shell/blob/360cfdb7ae7af9a73cc3f357a78bd35b5b12e829/src/screen.rs#L854-L873

Berrysoft:
Thank you, @faho ! I have pushed the change and now the problem should be fixed.

ahaoboy:
> Thank you, @faho ! I have pushed the change and now the problem should be fixed.

Can you update https://github.com/Berrysoft/fish-shell/releases? I also want to test if there are any other issues in starship

Berrysoft:
My custom release has been updated.

ahaoboy:
> My custom release has been updated.

Thank you very much. I found a new problem during use. When I use an empty config.fish and paste text via ctrl+v or right click, fish will display an error: `fish: Unknown command: __fish_paste`

Berrysoft:
I cannot reproduce this problem.

ahaoboy:
> I cannot reproduce this problem.

I copied all the [functions ](https://github.com/fish-shell/fish-shell/tree/master/share/functions) to local `.config/fish/functions`

But commandline seems to have some problems on Windows, I simply modified the __fish_paste function, at least it works now

```
 commandline: --insert-smart: unknown option

~/.config/fish/functions/__fish_paste.fish (line 47): 
        commandline --insert-smart -- $data
        ^
in function '__fish_paste' with arguments 'looks'

(Type 'help commandline' for related documentation)
```

```
function __fish_paste
    # Also split on \r, otherwise it looks confusing
    set -l data (string split \r -- $argv[1] | string split \n)
    set data (string replace -ra [[:cntrl:]] '' -- $data)
    commandline --insert -- $data
end

```

Berrysoft:
The fish package doesn't install scripts to your home directory. The one from the package is in `/usr/share/functions`.

ahaoboy:
> The fish package doesn't install scripts to your home directory. The one from the package is in `/usr/share/functions`.

Thanks for the tip. I used your bin/fish to replace the local executable file. Now I reinstall the zst file and it works perfectly.

Berrysoft:
I re-enable `posix_spawn` because cygwin has fixed this bug in 3.6.0, and MSYS2 will release it soon.

krobelus:
ah ok, `cargo build -Zbuild-std --target=x86_64-pc-cygwin`  seems to work (on a Linux host, a Windows one can't find `#include <paths.h>` when compiling `src/libc.c`).

But it seems we need a new libc release (for `libc::statfs`), and update the nix crate to be compatible with that.
I wonder if anyone has done the nix changes already.

krobelus:
> https://github.com/Berrysoft/fish-msys2/

thanks, that's very useful

krobelus:
merged; I squashed some of your commits (hope that's not a problem) and changed only one line I think (`1isize` -> `1_isize`).
Will add a changelog entry once we pull the requisite dependencies

ahaoboy:
Will update the document to add Windows installation and compilation steps?

https://github.com/fish-shell/fish-shell?tab=readme-ov-file#windows

ahaoboy:
I encountered a strange problem. When inputting fish may get stuck for a few seconds. At this time, fish seems to be stuck and will not respond to the input, but it seems to cache the input and then execute all the cached input. This problem occurs several times a day and is not easy to reproduce, so there is no screen recording. Has anyone else encountered this problem?

A2uria:
> Seems that `poll` doesn't return -1 on success any more. I removed the workarounds in `fd_monitor.rs`.

This issue still occurs on master (https://github.com/fish-shell/fish-shell/commit/db0f9c1d53e64721251663fc513ccfb16fed4f13) with the latest version of `msys2-runtime` (`3.6.4-1`).

Note that patch is required as https://github.com/fish-shell/fish-shell/commit/6644cc9b0e29841e3d0a85fbc672a95c4a2fd000 introduced compatibility issues.

```diff
diff --git a/src/path.rs b/src/path.rs
index 58954a5..7d7339d 100644
--- a/src/path.rs
+++ b/src/path.rs
@@ -667,7 +667,7 @@ fn create_dir_all_with_mode<P: AsRef<std::path::Path>>(path: P, mode: u32) -> st
 /// Return whether the given path is on a remote filesystem.
 pub fn path_remoteness(path: &wstr) -> DirRemoteness {
     let narrow = wcs2zstring(path);
-    #[cfg(target_os = "linux")]
+    #[cfg(any(target_os = "linux", cygwin))]
     {
         let mut buf = MaybeUninit::uninit();
         if unsafe { libc::statfs(narrow.as_ptr(), buf.as_mut_ptr()) } < 0 {
@@ -701,7 +701,7 @@ pub fn path_remoteness(path: &wstr) -> DirRemoteness {
             }
         }
     }
-    #[cfg(not(target_os = "linux"))]
+    #[cfg(not(any(target_os = "linux", cygwin)))]
     {
         fn remoteness_via_statfs<StatFS, Flags>(
             statfn: unsafe extern "C" fn(*const i8, *mut StatFS) -> libc::c_int,
```

I'm using `cargo install` to make fish with embedded data.

```bash
rustup component add rust-src --toolchain nightly-x86_64-pc-windows-gnu
FISH_BUILD_VERSION=4.1.0-alpha0 cargo +nightly-x86_64-pc-windows-gnu install -Z build-std --target x86_64-pc-cygwin --path . --bin fish
```

However, I'm not using `cargo install --git` because `*.fish` will end with CRLF in this case. The same goes for Git for Windows, regardless of `core.autocrlf=false`. Tarballs works well.

And with `msys2-runtime-3.5` (with msys2, as well as with git-for-windows <=2.49) I got another error. (before prompt?)

```
exec: Failed to execute process '/usr/bin/uname': Unsupported format.
```

Berrysoft:
I haven't come into this bug after that comment. If you can reproduce it, you're the best person to fix it, together with other patches, maybe.

And yes, fish needs cygwin 3.6. There's a bug in `posix_spawn` before.

Issue #11243, opened 2025-03-09
fish-printf does not support empty arguments


ahaoboy:
If there are no arguments, should the fmt string be returned directly?

```
fn main() {
    use fish_printf::sprintf;
    let s = sprintf!("%0.5g", 123456.0);
    println!("{s}");
    let s = sprintf!("hello");
    println!("{s}");
}

```

```
   |
5  |     let s = sprintf!("hello");
   |                             ^ missing tokens in macro arguments
   |
```

ridiculousfish:
Thanks for filing!

PR #11244, opened 2025-03-09
fish-printf: support empty arguments

ahaoboy:
## Description

fish-printf: support empty arguments
```
    let s = sprintf!("hello");
    println!("{s}");
    let mut s = String::new();
    sprintf!(=> s, "hello");
    println!("s2 {s}");
```
Fixes issue https://github.com/fish-shell/fish-shell/issues/11243

## TODOs:

### DIFF
--- a/printf/src/lib.rs
+++ b/printf/src/lib.rs
@@ -62,6 +62,26 @@ macro_rules! sprintf {
             ).unwrap()
         }
     };
+
+    // Handling the case where there is no arguments
+    (
+      $fmt:expr // format string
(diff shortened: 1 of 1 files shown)


ridiculousfish:
It shouldn't return the string unchanged (consider `sprintf!("%d")` which should remain an error) but we  can tweak the macros to support zero arguments - 7d3b157f1316550adaf70a2a5eabd37d9be35288

Thanks for filing!

ahaoboy:
Thanks, this is useful in some projects converting C to Rust, although it is a very rare case

Issue #8618, opened 2022-01-05
Remove folder name prefix in `cd` tab autocompletion suggestions


AlexSWall:
**fish version.** 3.3.1
**OS.** MacOS Monterey v12.1
**Terminal.** iTerm2 3.4.14
Behaviour persists on running `sh -c 'env HOME=$(mktemp -d) fish'`.

When typing `cd` and going into a directory, using tab completion prefixes the last directory in front of all results, sometimes truncated and prefixed by `...`. For example, `bar` below is needlessly prefixed.

<img width="488" alt="Screenshot 2022-01-05 at 23 03 56" src="https://user-images.githubusercontent.com/31999802/148302207-6f2c9eea-7e2d-4f40-b836-633f12b872a9.png">

I'm finding this very inconvenient when, for example, trying to tab complete at `~/Documents/`, which prefixes `...ocuments/` in front of every result. More than half of the text returned is useless. It looks something like the below (but worse).

<img width="623" alt="Screenshot 2022-01-05 at 23 08 02" src="https://user-images.githubusercontent.com/31999802/148302620-a34f74c2-c240-49c5-a958-90c0673eabdb.png">

How can I have fish only list the directory names?

krobelus:
This has previously been requested in #7883 (and @thithib's #8617)

I think it's reasonable to special-case file completion here, because it's already special: we complete one path component at a time, not the entire token.
So I believe we could accept this inconsistency in the pager (custom completions like for `git rm` will still have the current pager).
This means that filtering (with Control-S) will only consider the basename, but that's a good thing (today you can even filter for the ellipsis character (`…`), which seems like an accident).

This should probably be kept consistent with other completions of sub-tokens, like for `grep --color=` or `dd if=`

Sadly it's not so easy to do the same for completions that recursively call `complete -C` for file completions, because fish can't tell if these are files.
Those are not used very much, I think __fish_complete_suffix is the main one.

I don't think we need a configuration option (in theory it's possible to write your own pager that parses `complete -C`)

faho:
I don't see how this helps. It's an inconsistency, for very little gain.

You can see for yourself - the screenshot does use "half" of the space for the prefix, but why does it matter? It's still on one line!

I'm not sure how the length of the ellipsized prefix is determined today and there might be room for improvement, but having this consistent across completions is *nice*. Having a marker "hey, this is where this differs, this is how it fits in" is nice.

[...]

PR #11250, opened 2025-03-09
Remove folder name prefix in autocompletion suggestions

TheRealFalcon:
## Description

Remove folder name prefix in autocompletion suggestions

Fixes issue #8618

## Additional info

Example behavior:
image

This didn't break any tests locally for me. I'm willing to add new tests for this behavior, but I could use some direction as to where they would need to go and what sort of setup and/or mocking would be required.

## TODOs:

### DIFF
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -6511,10 +6511,22 @@ impl<'a> Reader<'a> {
             prefix.push_utfstr(&tok);
             prefix.push_utfstr(&common_prefix);
         } else {
-            // Append just the end of the string.
+            // Collapse parent directories and append end of string
             prefix.push(get_ellipsis_char());
+
             let full = tok + &common_prefix[..];
-            prefix.push_utfstr(&full[full.len() - PREFIX_MAX_LEN..]);
+            let truncated = &full[full.len() - PREFIX_MAX_LEN..];
+            let parts: Vec<&wstr> = truncated.split('/').collect();
+            if parts.len() < 2 {
+                // No path separators were found in the common prefix, so we can't collapse
+                // any further
+                prefix.push_utfstr(&truncated);
+            } else {
+                // Discard any parent directories and include whats left
+                let last_part = parts.last().expect("parts is non-empty");
+                prefix.push('/');
+                prefix.push_utfstr(last_part);
+            }
         }
 
         // Update the pager data.
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -6516,17 +6516,16 @@ impl<'a> Reader<'a> {
 
             let full = tok + &common_prefix[..];
             let truncated = &full[full.len() - PREFIX_MAX_LEN..];
-            let parts: Vec<&wstr> = truncated.split('/').collect();
-            if parts.len() < 2 {
+            let (i, last_component) = truncated.split('/').enumerate().last().unwrap();
+            if i == 0 {
                 // No path separators were found in the common prefix, so we can't collapse
                 // any further
                 prefix.push_utfstr(&truncated);
             } else {
                 // Discard any parent directories and include whats left
-                let last_part = parts.last().expect("parts is non-empty");
                 prefix.push('/');
-                prefix.push_utfstr(last_part);
-            }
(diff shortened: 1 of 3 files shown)


src/reader.rs
```diff
@@ -6511,10 +6511,22 @@ impl<'a> Reader<'a> {
...
         } else {
-            // Append just the end of the string.
+            // Collapse parent directories and append end of string
             prefix.push(get_ellipsis_char());
+
             let full = tok + &common_prefix[..];
-            prefix.push_utfstr(&full[full.len() - PREFIX_MAX_LEN..]);
+            let truncated = &full[full.len() - PREFIX_MAX_LEN..];
+            let parts: Vec<&wstr> = truncated.split('/').collect();
+            if parts.len() < 2 {
```
krobelus, line 6520: Seems like a good change. I'll try it for a while.
I haven't re-read the issue yet - maybe we can briefly summarize the reason for this change.

I gues zsh will only show the basename even if there is no truncation going on,
but we don't need to go there - your approach is safer.

src/reader.rs
```diff
@@ -6511,10 +6511,22 @@ impl<'a> Reader<'a> {
...
+                // No path separators were found in the common prefix, so we can't collapse
+                // any further
+                prefix.push_utfstr(&truncated);
+            } else {
+                // Discard any parent directories and include whats left
```
krobelus, line 6525: this works for absolute paths but it doesn't seem to work for `~/some/path`, can we fix that?

src/reader.rs
krobelus, line 6520: `parts.len() < 2` is a bit confusing because there's always at least one part.
Better make that obvious by checking `parts.len() == 1`. 

Actually we should also get rid of the gratuitous vector, how about


	    ```rust
            let (i, last_component) = truncated.split('/').enumerate().last().unwrap();
            if i == 0 {
                // No path separators were found in the common prefix, so we can't collapse
                // any further
                prefix.push_utfstr(&truncated);
            } else {
                // Discard any parent directories and include whats left
                prefix.push('/');
                prefix.push_utfstr(last_component);
            };
	    ```

krobelus, line 6520: I'm wondering if we should also allow `=` and `:` as separators, since we sort of do that when computing completions.
I guess let's wait until we find a use case; the LHS is not usually very large.

src/reader.rs
```diff
@@ -6511,10 +6511,22 @@ impl<'a> Reader<'a> {
             prefix.push_utfstr(&tok);
             prefix.push_utfstr(&common_prefix);
         } else {
-            // Append just the end of the string.
+            // Collapse parent directories and append end of string
```
krobelus, line 6514: should probably add to the changelog, maybe something like

	- When tab completion results are truncated, the common directory name is omitted.

krobelus, line 6514: > This didn't break any tests locally for me. I'm willing to add new tests for this behavior, but I could use some direction as to where they would need to go and what sort of setup and/or mocking would be required.

We don't have mocking for `src/reader.rs` yet, so either add that or write
a system test, maybe in `tmux-pager.fish`. To run, use

	cargo b && tests/test_driver.py target/debug tests/checks/tmux-pager.fish

krobelus (approved):

src/reader.rs
krobelus: after re-reading the issue, I think we should eventually have a generic way
of making sure that the common prefix is not way larger than any of the suffixes.

But that can definitely be done later, and I expect that we'll still want to keep the special handling of "/".

src/reader.rs
TheRealFalcon: Unless I'm misunderstanding you, it should already work for relative directories, but it depends on the length of the prefix. For example, here's my fish config dir:
image

But even with an absolute path, a short prefix won't collapse:
image

krobelus: yeah my bad

TheRealFalcon:
I believe I have addressed all of the comments and added a new check file with some basic tests for this behavior.

omentic:
Thanks! :-)

Issue #10756, opened 2024-09-30
Feature request: add a new history-last-token-search-backward that behaves more like bash's yank-last-arg


jfly:
I'm trying out Fish after many years in Zsh and many years of Bash before that. One difference in behavior that's bothering me is how `Alt+.` works:

- In Fish (`history-token-search-backward`, added in https://github.com/fish-shell/fish-shell/commit/02bd933c580ec0015ed298341d0b14cdc8768588), it loops backwards through all tokens of previous commands
- In Bash/Zsh (`yank-last-arg`), it loops backwards through all the *last* tokens of previous commands

This discrepancy was discussed in https://github.com/fish-shell/fish-shell/issues/89, but there didn't really seem to be anyone advocating for the `yank-last-arg` behavior:

https://github.com/fish-shell/fish-shell/issues/89#issuecomment-6494054

> Just curious.. what use cases do you find yank-last-arg an advantage over history-token-search-backward?

- https://github.com/fish-shell/fish-shell/issues/89#issuecomment-7271843

> it does cycle through all parameters, not only the last, but I'd consider that a feature :)

So here I am, 12 years later ;) I personally find the last argument behavior quite useful. Here's a common use case:

```shell
$ something that may have a lot of tokens > /tmp/before
$ something that may have a lot of tokens > /tmp/after
$ diff (Alt+..) (Alt+.)   # this would expand to `diff /tmp/before /tmp/after`
```

Would Fish be open to a new `history-last-token-search-backward` that behaves like the existing `history-token-search-backward`, but it only searches through the *last* tokens of previous commands? I poked at this in a fork (https://github.com/fish-shell/fish-shell/compare/3.7.1...jfly:fish-shell:add-history-last-token-search?expand=1), and it doesn't seem to be too difficult to implement (caveat: I am a complete noobie to Fish, Fish's codebase, and C++).

PR #11258, opened 2025-03-11
add history-last-token-search-backward/forward command

Frefreak:
## Description
This add two commands `history-last-token-search-backward` and `history-last-token-search-forward` which behaves like bash's yank-last-arg. So similar to `history-token-search-*` but only considers the last argument for each command.

Fixes issue #10756

## TODOs:

### DIFF
--- a/doc_src/cmds/bind.rst
+++ b/doc_src/cmds/bind.rst
@@ -265,6 +265,12 @@ The following special input functions are available:
 ``history-token-search-forward``
     search the history for the next matching argument
 
+``history-last-token-search-backward``
+    search the history for the last argument of the previous command
+
+``history-last-token-search-forward``
+    search the history for the last argument of the next command
+
 ``forward-jump`` and ``backward-jump``
     read another character and jump to its next occurrence after/before the cursor
 
--- a/doc_src/cmds/bind.rst
+++ b/doc_src/cmds/bind.rst
@@ -266,10 +266,10 @@ The following special input functions are available:
     search the history for the next matching argument
 
 ``history-last-token-search-backward``
-    search the history for the last argument of the previous command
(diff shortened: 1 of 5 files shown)


src/reader_history_search.rs
```diff
@@ -235,9 +238,15 @@ impl ReaderHistorySearch {
...
-            for tok in local_tokens.into_iter().rev() {
-                self.add_if_new(tok);
+            if self.mode == SearchMode::LastToken {
+                if let Some(tok) = local_tokens.into_iter().next_back() {
+                    self.add_if_new(tok);
+                }
+            } else {
+                // Make sure tokens are added in reverse order. See #5150
+                for tok in local_tokens.into_iter().rev() {
+                    self.add_if_new(tok);
```
krobelus, line 248: wouldn't it be simpler to add `if self.mode == SearchMode::LastToken { break }` here instead

doc_src/cmds/bind.rst
```diff
@@ -265,6 +265,12 @@ The following special input functions are available:
 ``history-token-search-forward``
     search the history for the next matching argument
 
+``history-last-token-search-backward``
+    search the history for the last argument of the previous command
+
+``history-last-token-search-forward``
+    search the history for the last argument of the next command
```
krobelus, line 272: I guess this could be phrased as

	search the history for the next matching last argument

krobelus (approved):

src/reader_history_search.rs
Frefreak: Indeed. I initially used `break`, but it didn't work due to another part. Then I used this form to verify something and forgot to change it back. Now updated.

jfly:
Thanks @Frefreak and @krobelus!

PR #11264, opened 2025-03-12
doc: `bind ctrl-c clear-commandline`

lengyijun:
If I understand correctly, `commandline -r ""` works the same as   `clear-commandline`

### DIFF
--- a/doc_src/interactive.rst
+++ b/doc_src/interactive.rst
@@ -549,8 +549,8 @@ Custom bindings
 
 In addition to the standard bindings listed here, you can also define your own with :doc:`bind <cmds/bind>`::
 
-  # Just clear the commandline on control-c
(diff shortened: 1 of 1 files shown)


doc_src/interactive.rst
```diff
@@ -550,7 +550,7 @@ Custom bindings
 In addition to the standard bindings listed here, you can also define your own with :doc:`bind <cmds/bind>`::
 
   # Just clear the commandline on control-c
-  bind ctrl-c 'commandline -r ""'
+  bind ctrl-c clear-commandline
```
krobelus, line 553: yeah I think this is equivalent but since it's the default we should probably use `cancel-commandline` as example here

faho:
Thanks!

Issue #11273, opened 2025-03-14
Tab completion for pacman query breaks with alias pacman='pacman --color=always'


ank426:
using fish version 4.0.1

I need `alias pacman='pacman --color=always'` for less.
It is in the `if status is-interactive ... end` block in config.fish.

With the alias:
![Image](https://github.com/user-attachments/assets/6cd50381-99dd-45d1-a194-0d5f59fa81ab)

Without the alias:
![Image](https://github.com/user-attachments/assets/f1ff09cc-1b78-4499-94fd-ce778d9d306f)

faho:
Yup, don't do that. Call your alias somethint else.

ank426:
I need the alias for less. When piping a query to less, unless the --color=always flag is set, it won't show colors.

faho:
Call your alias something else or make it an abbreviation.

ank426:
Nvm. I found a workaround. Copy /usr/share/fish/functions/__fish_print_pacman_packages.fish to /.config/fish/functions/ and in line 25 change `pacman` to `command pacman`.

PR #11274, opened 2025-03-14
change pacman to command pacman in __fish_print_pacman_packages.fish

ank426:
## Description
Fixes issue #11273

## TODOs:


faho:
As a policy issue, we don't pervasively use `command foo` everywhere because your function might also do something useful.

So: Call your alias something else or use an abbreviation.

Issue #5243, opened 2018-10-09
Treat `<(....)` as `(... | psub)`


mqudsi:
bash translates `<(foo)` by redirecting the output to a named file descriptor:

```bash
mqudsi@ZBook:/mnt/c/Users/mqudsi$ echo <(cat .gitconfig)
/dev/fd/63
```

Could fish support this syntax by merely translating `<(foo)` into `<(foo | psub)` and are there corner cases I'm not catching here?

zx8:
@ridiculousfish Google may have updated their search algorithm to support symbols since your comment, but `<()` seems to yield appropriate results for me...

![image](https://user-images.githubusercontent.com/8544900/68860466-7a733400-06e1-11ea-884d-567ccaebbc4b.png)

stephane-chazelas:
Seeing the "posix-me-harder" tag, note that that feature is not POSIX. It comes from ksh (already in ksh86) and is also found in zsh and bash. In yash, `<(...)` is for process *redirection*, not process *substitution*.

In `yash`, you'd do `cmd1 /dev/fd/3 3<(cmd2)` as the equivalent of ksh's `cmd1 <(cmd2)`.

faho:
>Seeing the "posix-me-harder" tag, note that that feature is not POSIX.

Sure. The tag is a joking way to say "make fish behave more like 'normal' shells", where "normal" shells are the ones most users are familiar with, like bash and zsh.  It's a reference to $POSIX_ME_HARDER (later renamed to $POSIXLY_CORRECT), which is an environment variable that some GNU tools use to turn on posix-compatible behavior in cases where GNU disagrees (like the size unit used in `df`, where POSIX demands 512 byte blocks).

As a general rule we don't care about POSIX compatibility.

5225225:
I think even if fish doesn't *support* it, making an attempt to *tell* the user that ran `diff <(ls) <(ls -l)` that they probably meant `diff (ls | psub --fifo) (ls -l | psub --fifo)` is a good idea. There's already an "Invalid redirection target" error, it just needs to also mention `psub` if applicable.

For a while I just assumed fish *doesn't have* this feature, since I couldn't find it in the docs and fish itself didn't tell me about it.

[...]

PR #11287, opened 2025-03-15
mention psub in the error for "Invalid redirection target"

5225225:
Currently fish errors out with

```fish
fish: Invalid redirection target:
rev <(ls)
    ^~~~^
```

This isn't very helpful in telling the user what they *should* be doing instead `rev (ls | psub)`.

## Description

Not a *fix* for https://github.com/fish-shell/fish-shell/issues/5243 , but helps guide the user to what they probably should be doing instead.

Now, the error looks like

```fish
# rev <(ls)
fish: Invalid redirection target: 
To do subprocess subsitution, use the psub command
see: `help psub'
rev <(ls)
    ^~~~^
```

Unsure if this is the best way to do this, but I figure it's better than *nothing*. Ideally we'd show the user psub being used correctly, but I have no idea how to use the error printing in fish, and changing the error message *text* was easy enough.

## TODOs:

### DIFF
--- a/src/parse_execution.rs
+++ b/src/parse_execution.rs
@@ -33,7 +33,10 @@ use crate::parse_constants::{
     WILDCARD_ERR_MSG,
 };
 use crate::parse_tree::{LineCounter, NodeRef, ParsedSourceRef};
-use crate::parse_util::parse_util_unescape_wildcards;
+use crate::parse_util::{
+    parse_util_locate_cmdsubst_range, parse_util_unescape_wildcards,
+    MaybeParentheses::CommandSubstitution,
+};
 use crate::parser::{Block, BlockData, BlockId, BlockType, LoopStatus, Parser, ProfileItem};
 use crate::parser_keywords::parser_keywords_is_subcommand;
 use crate::path::{path_as_implicit_cd, path_try_get_path};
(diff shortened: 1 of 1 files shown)


src/parse_execution.rs
```diff
@@ -1485,7 +1485,9 @@ impl<'a> ExecutionContext {
                     ctx,
                     STATUS_INVALID_ARGS.unwrap(),
                     redir_node,
-                    "Invalid redirection target: %ls",
+                    "Invalid redirection target: %ls\n\
+                     To do subprocess subsitution, use the psub command\n\
+                     see: `help psub'",
```
krobelus, line 1490: typo in `subprocess substitution`. Also the `see: help psub` should comfortably fit on the previous line, no?

src/parse_execution.rs
```diff
@@ -1485,7 +1485,9 @@ impl<'a> ExecutionContext {
                     ctx,
                     STATUS_INVALID_ARGS.unwrap(),
                     redir_node,
-                    "Invalid redirection target: %ls",
+                    "Invalid redirection target: %ls\n\
+                     To do subprocess subsitution, use the psub command\n\
```
krobelus, line 1489: Good idea but we should make some more effort to avoid false positive warnings

For example we should not mention psub on errors such as `wc <$nosuchvariable`.

maybe something like (pseudo code):

```fish
            let target_unexpanded = self.node_source_owned(&redir_node.target);
            if oper.mode == RedirectionMode::input
                && parse_util_locate_cmdsubst_range(target_unexpanded) == CommandSubstitution
            // && if expand_string() succeeds but returns 0 elements or more than one elements
            {
		// mention psub
            }
```

even with that, there's still a chance that this warning is confusing,
if the user actually meant to do something that is valid like

	wc <(echo README.rst)

but somehow made an error causing the command substitution to return 0 or multiple elements.
But that's fine, this is a rare scenario

src/parse_execution.rs
zanchey: also it's just "process substitution", not subprocess

src/parse_execution.rs
5225225: Adjusted, this seems to work? At least it *doesn't* mention psub for `cat <$asdf`, but *does* for `cat <(asdf)` (and `cat <$(asdf)` which. seems fine I guess?)

Unhappy with the branch here to just change the text, but no real other way of doing it, since here it *needs* to be a string literal from the looks of it. Not too bad, in any case.

PR #11289, opened 2025-03-16
fish_git_prompt: use informative status counts in main function

### LINKED ISSUE #11199: `fish_git_prompt` not applying branch colors when `show_informative_status` is set
Here's some settings from my prompt:

```fish
set -g __fish_git_prompt_showcolorhints yes
set -g __fish_git_prompt_showdirtystate yes
set -g __fish_git_prompt_showuntrackedfiles yes
set -g __fish_git_prompt_show_informative_status yes
set -g __fish_git_prompt_color_branch_dirty red
set -g __fish_git_prompt_color_branch_staged yellow
```

When `show_informative_status` is enabled, branch color is always default green, when it's disabled it shows branch colors correctly

zanchey:
I'm not 100% sure this is the best approach here, but `fish_git_prompt` is a pretty gnarly function and I didn't want to add more paths through it.

This commit tries to ensure that the informative path is taken for any situation where counts are needed.

The counts are bubbled up to the fish_git_prompt function, so there are some additional checks to ensure that 0 and empty behave the same.

Fixes #11199. Tested as not breaking #9572.

### DIFF
--- a/share/functions/fish_git_prompt.fish
+++ b/share/functions/fish_git_prompt.fish
@@ -244,6 +244,7 @@ function fish_git_prompt --description "Prompt function for Git"
     set -l informative
     set -l dirty
     set -l untracked
+    set -l stash
     command git config -z --get-regexp 'bash\.(showInformativeStatus|showDirtyState|showUntrackedFiles)' 2>/dev/null | while read -lz key value
         switch $key
             case bash.showinformativestatus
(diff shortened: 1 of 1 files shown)


share/functions/fish_git_prompt.fish
```diff
@@ -270,15 +271,21 @@ function fish_git_prompt --description "Prompt function for Git"
...
-        # enabled globally (via the fish variable) and dirty or untracked are not false.
+        # enabled globally (via the fish variable) and dirty, untracked, and stash are not false.
+        # (supports #9572)
         #
         # This is to allow overrides for the repository.
         if test "$informative" = true
             or begin
                 contains -- "$__fish_git_prompt_show_informative_status" yes true 1
-                and test "$dirty" != false
+                and test "$dirty" != false -o "$untracked" != false -o "$stash" != false
```
faho, line 288: This means that whenever you ask for stash state it will switch to informative mode, which is *much* slower.

It's probably worth preserving stash state that doesn't also incorporate calling `git status` to count everything.

Issue #11277, opened 2025-03-14
Ghostty + Fish 4.1: `Could not set up terminal for $TERM 'xterm-ghostty'`


syhol:
Mac M3 arm64
Ghostty: 1.1.2
Fish: 4.0.1 (homebrew)

Getting for following warning:

```
warning: Could not set up terminal for $TERM 'xterm-ghostty'. Falling back to hardcoded xterm-256color values
```

I don't think this had caused any behavioural issues, the warning on startup is just annoying.

I started getting this warning after upgrading from Fish 3.

It also looks like I'm not the only one: https://news.ycombinator.com/item?id=43197765

faho:
That means that the ghostty terminfo file can't be found. Fish searches a variety of hardcoded directories and what is set in $TERMINFO and $TERMINFO_DIRS: https://github.com/meh/rust-terminfo/blob/7bbf236c1bdcdb39f4e40ecf2f840da1c576a461/src/database.rs#L159-L185

How did you install ghostty and where did it put the terminfo file?

(to be honest we can probably disable that warning, it's not going to affect anything)

faho:
See also https://ghostty.org/docs/help/terminfo

faho:
Okay, those are just the regular semantics of $TERMINFO (if it is set, it is searched, if it is not set, ~/.terminfo is searched - meaning if it is set ~/.terminfo is *not* searched unless that's the value), so I'm calling that a configuration error.

For fun, here's the differences between "xterm-256color" and ncurses terminfo's new "ghostty" entry (I don't think they included "xterm-ghostty" because they don't like anything calling itself "xterm") in the capabilities that fish uses:

```
ghostty
        key_backspace: '\177', '^H'.
```

 i.e. ghostty has backspace as "del", xterm-256color has it as ctrl-h - which is something that's indistinguishable anyway and many terms have configurable. Plus we use the kitty keyboard protocol in ghostty anyway so we don't care about the regular key sequences. That means: We have *zero* notable difference between ghostty and xterm-256color and so we could just silence the warning in this case and probably others.

Other terminfo entries that differ in a similarly useless manner: alacritty, contour, foot, hterm-256color, mintty, mosh-256color, rio, terminology, termite, wezterm and xterm.js.

(also tmux-256color - which *has* more differences in terminfo, except they're all misleading - the xterm sequences work)

Those are the majority of terminals in common use today (anything else will claim to be xterm-256color to begin with!).

syhol:
[...]

PR #11290, opened 2025-03-16
Downgrade $TERM warnings to a term-support flog

faho:
The chances that xterm-256color breaks anything are miniscule.

In the features we use, there are basically no differences, especially when you consider that we decode keys independently.

E.g. tmux-256color has differences, but they are either just taste questions (xterm's clear_screen will also clear scrollback), or they're just... not actually different?

Terminfo will claim that it uses a different cursor_up and exit_attribute_mode, but it also understands the xterm ones, and it sends a different key_home,
but we decode that even with TERM=xterm-256color.

In some cases, terminfo is also just outright *wrong* and will claim something does not support italics when it does.

So, since the differences are very likely to simply not matter, throwing a warning is more confusing than it is helpful.

## TODOs:

I would like to include this in 4.0.2 because we've already had #11277, and that happened for at least two people. Even tho that was arguably a configuration error, the upshot is if we had just treated ghostty as xterm-256color literally nothing would have broken because there is no useful difference between these two for us.

I would actually go a step further and abolish our terminfo use altogether, but for that I would have to *quantify* just how many differences we have in practice, and that's hard because you need to throw away all the useless and sometimes outright *wrong* information terminfo gives you (e.g. the "konsole" terminfo entry is not used and does not document konsole's behavior, so you need to ignore it. some sequences like "cursor_normal" are sometimes equivalent but have their parts swapped, so `infocmp` will tell you there's a difference but it doesn't matter. tmux will understand xterm's sequences, so will screen (for the most part - but which parts exactly?). key_backspace is sometimes delete and sometimes ctrl-h but differs between terminals using the same TERM and can be configured).

So this is the next step.

### DIFF
--- a/src/env_dispatch.rs
+++ b/src/env_dispatch.rs
@@ -555,19 +555,16 @@ fn init_terminal(vars: &EnvStack) {
     if terminal::setup(None, |term| apply_term_hacks(vars, term)).is_none() {
         if is_interactive_session() {
             let term = vars.get_unless_empty(L!("TERM")).map(|v| v.as_string());
-            // We do not warn for xterm-256color at all, we know that one.
-            if term != Some("xterm-256color".into()) {
-                if let Some(term) = term {
-                    FLOG!(
-                        warning,
-                        wgettext_fmt!("Could not set up terminal for $TERM '%ls'. Falling back to hardcoded xterm-256color values", term)
-                    );
-                } else {
-                    FLOG!(
-                        warning,
-                        wgettext!("Could not set up terminal because $TERM is unset. Falling back to hardcoded xterm-256color values")
-                    );
-                }
+            if let Some(term) = term {
+                FLOG!(
+                    term_support,
+                    wgettext_fmt!("Could not set up terminal for $TERM '%ls'. Falling back to hardcoded xterm-256color values", term)
+                );
+            } else {
+                FLOG!(
+                    term_support,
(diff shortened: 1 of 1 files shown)


krobelus (approved):

krobelus:
> key_backspace is sometimes delete and sometimes ctrl-h but differs between terminals using the same TERM and can be configured).

for the default binding we use `VERASE` nowadays (which should obsolete the terminfo-backed `bind -k backspace`).

faho:
Cherry-picked to 4.0.2 as 76d2419228627764876d1b001c3b74e6222499bb

faho:
Here are the terminfo entries that have absolutely no difference compared to xterm-256color in the capabilities that we use:

alacritty foot ghostty hterm-256color rio xterm-256color xterm-ghostty 

That's absolutely *zero*, not even useless swaps (`cursor_normal: '\E[?12h\E[?25h', '\E[?12l\E[?25h'`). Ignore those and add xterm-kitty.

if you ignore the sequences we only use to remove them from  input (e.g. for width calculations), add contour and mintty.

And that's *going by terminfo* to begin with. If you discount keys (which we decode independently) and discount differences that aren't actually a difference in practice (i.e. the xterm sequence works or the feature isn't supported and it is ignored), you can add basically *every single terminal in use*, including tmux-256color, st-256color, rxvt-unicode-256color (but we do not currently decode those keys - I don't think anyone cares?). Even Eterm-256color (that's emacs' ansi-term) should be fine (it breaks due to things that don't have terminfo capabilities).

PR #11299, opened 2025-03-18
Fix all current rustdoc warnings (apart from mentioning private items)

5225225:
## Description

After this, `cargo doc --document-private-items` is clean, and the docs are somewhat more useful to read. (Developer docs are *very* useful for new contributors to find what functions actually do).

There's a fair amount of `\param` usage that gets mangled when rendered as markdown, see 

https://github.com/fish-shell/fish-shell/blob/d203ee4d531b10cb6204d7b327b0ecd88fa30d1b/src/wildcard.rs#L1048-L1053

I didn't tackle that in this PR, but I could go and update those to a different format , such as:

> * `filename`: The file for which to find a description string
> * `is_dir`: Whether the file is a directory or not (might be behind a link)

Some of these docs I had to change the wording of because it was just wrong, such as:

https://github.com/5225225/fish-shell/blob/38c8c59bb91c8dc26f2003cc59d2337a97cef475/src/fd_monitor.rs#L189-L193

## TODOs:

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -1582,7 +1582,7 @@ impl ConcreteNodeMut for IfStatement {
 #[derive(Default, Debug)]
 pub struct CaseItem {
     parent: Option<*const dyn Node>,
-    /// case <arguments> ; body
+    /// case \<arguments\> ; body
     pub kw_case: KeywordCase,
     pub arguments: ArgumentList,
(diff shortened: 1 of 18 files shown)


5225225:
Also, it might be a good idea to add `cargo doc --document-private-items` to CI.

PR #11311, opened 2025-03-21
Update lru to 0.13.0

musicinmybrain:
## Description

This updates the `lru` crate dependency to the latest version.

Looking at the [changelog](https://github.com/jeromefroe/lru-rs/blob/master/CHANGELOG.md#v0130---2025-01-27), the only breaking change was that the MSRV was updated to 1.65.0. This is not a problem because it is still less than fish’s MSRV of 1.70. For extra confidence, I built fish and ran the tests (`cmake . && make fish_run_tests`).

## TODOs:

- [ ] Changes to fish usage are reflected in user documentation/manpages. **N/A**
- [ ] Tests have been added for regressions fixed **N/A**
- [ ] User-visible changes noted in CHANGELOG.rst  **N/A**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -194,9 +194,9 @@ checksum = "a7a70ba024b9dc04c27ea2f0c0548feb474ec5c54bba33a7f72f873a39d07b24"
 
 [[package]]
 name = "lru"
-version = "0.12.5"
+version = "0.13.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "234cf4f4a04dc1f57e24b96cc0cd600cf2af460d4161ac5ecdd0af8e1f3b2a38"
+checksum = "227748d55f2f0ab4735d87fd623798cb6b664512fe979705f829c9f81c934465"
 dependencies = [
  "hashbrown",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,7 @@ libc = "0.2"
 # lru pulls in hashbrown by default, which uses a faster (though less DoS resistant) hashing algo.
 # disabling default features uses the stdlib instead, but it doubles the time to rewrite the history
 # files as of 22 April 2024.
-lru = "0.12.3"
+lru = "0.13.0"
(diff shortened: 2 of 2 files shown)


krobelus:
LGTM. I don't know how often we want to update, maybe it depends on downstream consumers.
Gratuitous updates can needlessly invalidate caches but I don't know if that's a problem.

musicinmybrain:
> LGTM. I don't know how often we want to update, maybe it depends on downstream consumers. Gratuitous updates can needlessly invalidate caches but I don't know if that's a problem.

As far as I can tell, the `lru` crate only handles in-memory caches. If I’m skimming the fish sources correctly, it’s used for in-memory caches associated with some things that are persistent on the disk, like history and autoload files, but it has no role in reading and writing file caches. If that’s correct, cache invalidation shouldn’t be a concern because `lru`’s data structures can’t outlive the process.

In any case, the last couple of SemVer-incompatible `lru` updates have been about a year apart, so the churn rate doesn’t seem to be excessive.

For context, the direct motivation here is to remove a dependency on the `rust-lru0.12` compat package in Fedora, and especially to avoid branching new compat packages while shipping fish 4 in EPEL 10 (assuming the EPEL Steering Committe approves the update). Obviously, this PR is something we can carry as a downstream-only patch, but it’s our policy to offer all patches upstream whenever possible and to try to stay as close to upstream as we reasonably can.

krobelus:
> As far as I can tell, the lru crate only handles in-memory caches

sorry I meant cargo build cache. But that's not relevant if the update helps anyone

> For context, the direct motivation here is to remove a dependency on the rust-lru0.12 compat package

makes sense, thanks

zanchey:
> For context, the direct motivation here is to remove a dependency on the rust-lru0.12 compat package

this is a really useful point! thanks!

krobelus:
we can cherry-pick it to 4.0.2 too. LRU itself had no relevant changes itself and we've already been using hashbrown 0.15.2 before

krobelus:
I didn't cherry-pick it because most distros have probably already packaged it with the previous `lru` version.
Not sure what's better

PR #11315, opened 2025-03-24
completions/cryptsetup: complete device mapping names

memchr:
## Description

commands 'close', 'resize', and 'status' each take 'name' as their solo argument.


share/completions/cryptsetup.fish
```diff
@@ -1,4 +1,9 @@
 #cryptsetup 2.3.5
+function __fish_dm_names
+    set names (path basename /dev/mapper/*)
+    set -e names[(contains -i control $names)]
+    echo $names
```
faho, line 5: This won't work correctly because these need to be on separate *lines*.

You can just make this function

```fish
path basename /dev/mapper/* | string match -v control
```

faho (changes requested):

share/completions/cryptsetup.fish
```diff
@@ -1,4 +1,7 @@
 #cryptsetup 2.3.5
+function __fish_dm_names
```
krobelus, line 2: technically this should be namespaced according to the file name, so `__fish_cryptsetup_dm_names`.
Could also just inline this function, no?

PR #11316, opened 2025-03-24
completions/iwctl: fix parsing bug with multi-column characters

memchr:
## Description

`__iwctl_filter` parses column information by splitting based on a fixed number
of characters. However, some runes, such as `📡`, are multiple columns wide. The
iwctl client uses wcwidth to visually align strings containing such runes. The
current implementation does not take this into account, resulting in errors in
the `__iwctl_connect` function.

## PoC

```fish
set wide "📡📡📡  0123456"
set norm "memchr  0123456"

# the columns are visually aligned
echo $wide
echo $norm

echo (string sub -l 9 $wide)+
echo (string sub -l 9 $norm)+
```
result
Pasted image

### DIFF
--- a/share/completions/iwctl.fish
+++ b/share/completions/iwctl.fish
@@ -20,6 +20,8 @@ function __iwctl_filter -w iwctl
 
     if set -ql _flag_all_columns
         for line in (string match "  *" -- $results[5..] | string sub -s (math $leading_ws + 1))
+            set -l column_widths $column_widths
+            set column_widths[1] (math $column_widths[1] - (string length -V $line) + (string length $line))
             for column_width in $column_widths
                 printf %s\t (string sub -l $column_width -- $line | string trim -r)
                 set line (string sub -s (math $column_width + 1) -- $line)
@@ -29,7 +31,9 @@ function __iwctl_filter -w iwctl
     else if set -q column_widths[1]
         # only take lines starting with `  `, i.e., no `No devices ...`
         # then take the first column as substring
-        string match "  *" $results[5..] | string sub -s (math $leading_ws + 1) -l $column_widths[1] | string trim -r
+        for line in (string match "  *" -- $results[5..] | string sub -s (math $leading_ws + 1))
+            string sub -l (math $column_widths[1] - (string length -V $line) + (string length $line)) $line | string trim -r
+        end
     end
 end
 


memchr:
But the other way to get this information is currently much more complicated and tedious. You need to use dbus. for instance. replicating the `iwctl station wlan0 get-networks rssi-dbms` command requires a series of D-Bus calls.


First, you need to resolve the D-Bus path of the `wlan0` interface. This involves querying the ObjectManager:


```fish

dbus-send --system --dest=net.connman.iwd --type=method_call --print-reply / org.freedesktop.DBus.ObjectManager.GetManagedObjects

```


This returns a large data structure from which you must extract the specific path for `wlan0`, such as `/net/connman/iwd/0/6`.  Then, you use this path to call `net.connman.iwd.Station.GetOrderedNetworks` to retrieve a list of

available networks and their signal strengths:


```fish

dbus-send --system --dest=net.connman.iwd --type=method_call --print-reply /net/connman/iwd/0/6 net.connman.iwd.Station.GetOrderedNetworks

```
Which returns something like this
```
method return time=1742828878.809836 sender=:1.10 -> destination=:1.970 serial=2782 reply_serial=2
   array [
      struct {
         object path "/net/connman/iwd/0/6/f7ff34cf5fba9b3c4a8da65e_psk"
         int16 -3600
      }
      struct {
         object path "/net/connman/iwd/0/6/321d4af54ad7ab2bf35e1f21_pks"
         int16 -6400
      }
      struct {
         object path "/net/connman/iwd/0/6/aac1ea8a15b95ff55b96c62d_psk"
         int16 -2700
      }
      struct {
         object path "/net/connman/iwd/0/6/f4cd3e5f9eeebaf35ab9ea23_psk"
         int16 -7300
      }
      struct {
         object path "/net/connman/iwd/0/6/45de6eb5c67a1ddbbbf8ecb6_psk"
         int16 -7000
      }
      struct {
         object path "/net/connman/iwd/0/6/5aef4d35f7cdce8aaf3ba56d_psk"
         int16 -2000
      }
      struct {
         object path "/net/connman/iwd/0/6/b68bd34f7ab6ac9d95dbfae0_psk"
         int16 -7600
      }
      struct {
         object path "/net/connman/iwd/0/6/e3a4c98b65d8af59ad9dbd5e_psk"
         int16 -7100
      }
      struct {
         object path "/net/connman/iwd/0/6/94bdf36f77bcf2e89ce7e22d_psk"
         int16 -7200
      }
      struct {
         object path "/net/connman/iwd/0/6/b67c7588a1b4fa34b9be9bb5_psk"
         int16 -6900
      },
```

Finally, for each network path returned, you need to call `org.freedesktop.DBus.Properties.Get` with the parameters

`('net.connman.iwd.Network', 'Name')` to obtain the network name.


This process is only for retrieving scanned networks, different and incompatible logic would be required to retrieve other information used by this completion script. 


Furthermore, the output of `dbus-send` (or `busctl`, which itself depends on systemd, potentially limiting portability) needs to be parsed. The resulting shell script would be incredibly complex and difficult to maintain.

krobelus:
maybe use `python -c 'import dbus; ... '`? See `__fish_anypython`.
Then we don't need to parse text.

memchr:
Seems a bit too long for in-line use

known network
```python
import dbus; bus = dbus.SystemBus(); manager = dbus.Interface( bus.get_object("net.connman.iwd", "/"), "org.freedesktop.DBus.ObjectManager"); [ print(str(interfaces["net.connman.iwd.KnownNetwork"]["Name"])) for _, interfaces in manager.GetManagedObjects().items() if "net.connman.iwd.KnownNetwork" in interfaces ]
```

krobelus:
> Seems a bit too long for in-line use

should probably use a multiline string, like

python -c '
do something
do something else
'

faho:
Note: the python dbus module is a separate thing, and may not be installed.

memchr:
iwd has no interface other than dbus, so for a meh version you need python-dbus (parsing busctl/dbus-send output would still be too much), but it is a [common enough](https://pkgstats.archlinux.de/packages?query=python-dbus) [dependency](https://debian.pkgs.org/sid/debian-main-amd64/python3-dbus_1.4.0-1_amd64.deb.html).


Here is a PoC
```fish
function __iwctl_list
    argparse -i d/devices n/networks -- $argv
    set -l python (__fish_anypython); or return
    set -l payload "
import dbus
bus = dbus.SystemBus()
objects = dbus.Interface(bus.get_object('net.connman.iwd', '/'), 'org.freedesktop.DBus.ObjectManager').GetManagedObjects()
DEVICE = 'net.connman.iwd.Device'
def names(interface): [print(i[interface]['Name']) for _, i in objects.items() if interface in i]
def devices(t=None): [print(i[DEVICE]['Name']) for _, i in objects.items() if DEVICE in i and t in i]
def networks(name):
    station = next(
        dbus.Interface(bus.get_object('net.connman.iwd', path), 'net.connman.iwd.Station')
        for path, i in objects.items()
        if 'net.connman.iwd.Station' in i and i[DEVICE]['Name'] == name)
    for network_path, signal in station.GetOrderedNetworks():
        print(f'{objects[network_path]['net.connman.iwd.Network']['Name']}\\0{signal}')
"
    if set -q _flag_devices
        python -c "$payload""devices('net.connman.iwd.$argv[1]')"
    else if set -q _flag_networks
        python -c "$payload""networks('$argv[1]')"
    else
        python -c "$payload""names('net.connman.iwd.$argv[1]')"
    end
end


echo "Adapter: $(__iwctl_list Adapter)"
echo "Station: $(__iwctl_list -d Station)"
echo "Networks:"
__iwctl_list -n wlan0
```

PR #11321, opened 2025-03-25
completions/iwctl: interact with iwd via python-dbus

memchr:
Also fix parsing bug with multi-column characters.

conflicts with #11316


share/completions/iwctl.fish
```diff
@@ -1,35 +1,26 @@
...
+def names(interface): [print(i[interface]['Name']) for _, i in objects.items() if interface in i]
+def devices(t=None): [print(i[DEVICE]['Name']) for _, i in objects.items() if DEVICE in i and t in i]
+def networks(name):
+    station = next(
+        dbus.Interface(bus.get_object('net.connman.iwd', path), 'net.connman.iwd.Station')
+        for path, i in objects.items() if 'net.connman.iwd.Station' in i and i[DEVICE]['Name'] == name)
+    [print(f'{objects[network_path]['net.connman.iwd.Network']['Name']}\\t{signal}') for network_path, signal in station.GetOrderedNetworks()]
+"
+    if set -q _flag_devices
+        python -c "$payload""devices('net.connman.iwd.$argv[1]')"
```
faho, line 19: This will currently fail if the dbus module isn't installed.

It probably shouldn't print an error. It's probably enough to silence stderr (`2>/dev/null`)

Also the arguments would be passed as arguments, use sys.argv. That avoids an awkward python command injection.

faho (changes requested):

share/completions/iwctl.fish
memchr: > Also the arguments would be passed as arguments, use sys.argv. That avoids an awkward python command injection

Then it would look like this:
```python
python -c "$payload devices(sys.argv[1])" -- $argv
```
Or, if we avoid all injections, we'll probably have to move argparsing to Python as well, since the difference in how we get network and device names can't be eliminated.

What do you think?
```python
import argparse

parser = argparse.ArgumentParser(add_help=False)
parser.add_argument('-d', '--device', action='store_true')
parser.add_argument('-n', '--network', action='store_true')
parser.add_argument("name", type=str)
args = parser.parse_args()

if args.device:
    devices(args.name)
elif args.network:
    networks(args.name)
else:
    names(args.name)
```

memchr:
An alternate approach is to use Gio via Python-gi, which boasts a broader base of dependent packages (inc KDE, gnome and other gtk based DE) and is therefore more likely to be pre-installed, but the execution time with this module is three times longer than with python-dbus., yielding an overall lag of over 100ms, potentially perceptible to discerning human users.

```python
import sys, getopt, gi; gi.require_version("Gio", "2.0"); from gi.repository import Gio
SERVICE = "net.connman.iwd"
DEVICE = f"{SERVICE}.Device"
bus = Gio.bus_get_sync(Gio.BusType.SYSTEM, None)
def get(path, interface):
    return Gio.DBusProxy.new_sync(bus, Gio.DBusProxyFlags.NONE, None, SERVICE, path, interface, None)
objects: dict = get("/", "org.freedesktop.DBus.ObjectManager").GetManagedObjects()

def names(ns): [print(i[ns]["Name"]) for i in objects.values() if ns in i]
def devices(ns): [print(i[DEVICE]["Name"]) for i in objects.values() if ns in i and DEVICE in i]
def networks(name):
    station = next(get(path, f"{SERVICE}.Station")
        for path, i in objects.items()
        if f"{SERVICE}.Station" in i and i[DEVICE]["Name"] == name)
    for path, signal in station.GetOrderedNetworks():
        print(f"{objects[path][f'{SERVICE}.Network']['Name']}\\t{signal}")

opts, args = getopt.getopt(sys.argv[1:], "dn")
for opt, _ in opts:
    if opt == "-d": devices(f"{SERVICE}.{args[0]}"); break
    if opt == "-n": networks(args[0]); break
else: names(f"{SERVICE}.{args[0]}")
```

faho:
Okay, I'm not sure any of this is really shippable.

iwctl's formatting is unusable, but python-dbus isn't common enough.

I would have hoped there's some /sys file or something we could read instead.

So I would really love if iwd gains *some* machine-readable CLI output (instead of claiming everything is "unstable" - it's been years now, folx, time to nail it down).

memchr:
However, the current way iwctl completion handles multi-column characters makes it unusable. Would you be open to the previous fix? #11316
```
❯ iwctl station wlan0 connect Invalid number:
/usr/share/fish/completions/iwctl.fish (line 65):
        if test $strength -ge -6000
           ^
in function '__iwctl_connect'
in command substitution
Invalid number:
/usr/share/fish/completions/iwctl.fish (line 67):
        else if test $strength -ge -6700
                ^
in function '__iwctl_connect'
in command substitution
Invalid number:
/usr/share/fish/completions/iwctl.fish (line 69):
        else if test $strength -ge -7500
                ^
in function '__iwctl_connect'
in command substitution
```

faho:
> However, the current way iwctl completion handles multi-column characters makes it unusable. Would you be open to the previous fix? 

To be honest: The previous fix is super rickety. It can easily be defeated by width calculations differing between fish and iwctl.

I don't think it is a suitable approach. The format simply is not usable.

memchr:
> iwctl's formatting is unusable, but python-dbus isn't common enough.

What about the busctl command from systemd, I think it should be common enough among the demographic of iwd users. Here is a reworked implementation using its json output and python without any third party modules.

PR #11322, opened 2025-03-26
completions/git: add --filter option

memchr:
supported subcommands:
- clone
- fetch
- submodule update
- rev-list


faho:
Thanks!

PR #11330, opened 2025-03-27
Attempt at fixing history corruption

danielrainer:
- Replace the write_loop call with Writer::write_all.
- Call sync_all on the file to ensure it gets written to storage.

## Description

This is an attempt at fixing #10300, which I have been experiencing for months.

The idea is to introduce flushing and using Rusts `Writer::write_all` instead of the custom `write_loop` function (which should probably be replaced by `write_all` in the entire codebase, alongside `read_loop`).

I have first tested running cc455ff97677c0085de480f37af05d07d127739e, on which I observed the history corruption after a few days. Then, I applied the changes present in this PR on top of cc455ff97677c0085de480f37af05d07d127739e, and have not observed any history corruption for over two weeks now. This does not mean that the issue is definitely fixed by these changes, but I think it might be useful for other people to have a look at this and maybe test the changes on their systems to see if history corruption still occurs for them with these changes applied.

Fixes issue #10300, maybe.

## TODOs:

- [ ] Have others who experience history corruption test these changes.

### DIFF
--- a/src/history.rs
+++ b/src/history.rs
@@ -36,7 +36,7 @@ use std::{
     num::NonZeroUsize,
     ops::ControlFlow,
     os::{
-        fd::{AsFd, AsRawFd, RawFd},
+        fd::{AsFd, AsRawFd},
         unix::fs::MetadataExt,
     },
     sync::{Arc, Mutex, MutexGuard},
@@ -52,7 +52,7 @@ use rand::Rng;
 use crate::{
     ast::{Ast, Node},
     common::{
-        str2wcstring, unescape_string, valid_var_name, wcs2zstring, write_loop, CancelChecker,
+        str2wcstring, unescape_string, valid_var_name, wcs2zstring, CancelChecker,
         UnescapeStringStyle,
     },
     env::{EnvMode, EnvStack, Environment},
@@ -154,14 +154,18 @@ pub const VACUUM_FREQUENCY: usize = 25;
 
 /// If the size of `buffer` is at least `min_size`, output the contents `buffer` to `fd`,
 /// and clear the string.
-fn flush_to_fd(buffer: &mut Vec<u8>, fd: RawFd, min_size: usize) -> std::io::Result<()> {
+fn flush_to_file(buffer: &mut Vec<u8>, file: &mut File, min_size: usize) -> std::io::Result<()> {
     if buffer.is_empty() || buffer.len() < min_size {
         return Ok(());
     }
 
-    write_loop(&fd, buffer)?;
+    file.write_all(buffer)?;
     buffer.clear();
-    return Ok(());
+    // Ensure that the file content is actually written to persistent storage.
+    // This might be unnecessary.
+    // It is used to help eliminate possible causes for
+    // https://github.com/fish-shell/fish-shell/issues/10300
+    file.sync_all()
 }
 
 struct TimeProfiler {
@@ -633,7 +637,7 @@ impl HistoryImpl {
         let mut buffer = Vec::with_capacity(HISTORY_OUTPUT_BUFFER_SIZE + 128);
         for item in items {
             append_history_item_to_buffer(&item, &mut buffer);
-            if let Err(e) = flush_to_fd(&mut buffer, dst.as_raw_fd(), HISTORY_OUTPUT_BUFFER_SIZE) {
+            if let Err(e) = flush_to_file(&mut buffer, dst, HISTORY_OUTPUT_BUFFER_SIZE) {
                 err = Some(e);
                 break;
             }
@@ -647,7 +651,7 @@ impl HistoryImpl {
 
             false
         } else {
-            flush_to_fd(&mut buffer, dst.as_raw_fd(), 0).is_ok()
+            flush_to_file(&mut buffer, dst, 0).is_ok()
         }
     }
 
@@ -865,7 +869,7 @@ impl HistoryImpl {
             }
         }
 
-        if let Some(history_file) = history_file {
+        if let Some(mut history_file) = history_file {
             // We (hopefully successfully) took the exclusive lock. Append to the file.
             // Note that this is sketchy for a few reasons:
             //   - Another shell may have appended its own items with a later timestamp, so our file may
@@ -894,11 +898,7 @@ impl HistoryImpl {
                 let item = &self.new_items[self.first_unwritten_new_item_index];
                 if item.should_write_to_disk() {
                     append_history_item_to_buffer(item, &mut buffer);
-                    res = flush_to_fd(
-                        &mut buffer,
-                        history_file.as_raw_fd(),
-                        HISTORY_OUTPUT_BUFFER_SIZE,
-                    );
+                    res = flush_to_file(&mut buffer, &mut history_file, HISTORY_OUTPUT_BUFFER_SIZE);
                     if res.is_err() {
                         break;
                     }
@@ -908,7 +908,7 @@ impl HistoryImpl {
             }
 
             if res.is_ok() {
-                res = flush_to_fd(&mut buffer, history_file.as_raw_fd(), 0);
+                res = flush_to_file(&mut buffer, &mut history_file, 0);
             }
 
             // Since we just modified the file, update our history_file_id to match its current state
--- a/src/history.rs
+++ b/src/history.rs
@@ -821,8 +821,7 @@ impl HistoryImpl {
     }
 
     /// Saves history by appending to the file.
-    // TODO: return error information
-    fn save_internal_via_appending(&mut self) -> bool {
+    fn save_internal_via_appending(&mut self) -> std::io::Result<()> {
         FLOGF!(
             history,
             "Saving %lu items via appending",
@@ -831,31 +830,34 @@ impl HistoryImpl {
         // No deleting allowed.
         assert!(self.deleted_items.is_empty());
 
-        let mut ok = false;
-
         // If the file is different (someone vacuumed it) then we need to update our mmap.
         let mut file_changed = false;
 
         // Get the path to the real history file.
         let Some(history_path) = history_filename(&self.name, L!("")) else {
-            // TODO: Why have a successful return here?
-            return true;
+            // No history should be saved if fish_history is set to the empty string,
+            // so nothing needs to be done here.
+            return Ok(());
         };
 
         // We are going to open the file, lock it, append to it, and then close it
         // After locking it, we need to stat the file at the path; if there is a new file there, it
         // means the file was replaced and we have to try again.
         // Limit our max tries so we don't do this forever.
-        let mut history_file = None;
-        for _i in 0..MAX_SAVE_TRIES {
-            let Ok(mut file) = wopen_cloexec(
+        let mut num_attempts = 0;
+        let mut history_file = loop {
+            if num_attempts == MAX_SAVE_TRIES {
+                return Err(std::io::Error::new(
+                    std::io::ErrorKind::Other,
+                    "Number of unsuccessful attempts to open history file exceeds maximum.",
+                ));
+            }
+            // Return immediately if an error occurs here.
+            let mut file = wopen_cloexec(
                 &history_path,
                 OFlag::O_WRONLY | OFlag::O_APPEND,
                 Mode::empty(),
-            ) else {
-                // can't open, we're hosed
-                break;
-            };
+            )?;
 
             // Exclusive lock on the entire file. This is released when we close the file (below). This
             // may fail on (e.g.) lockless NFS. If so, proceed as if it did not fail; the risk is that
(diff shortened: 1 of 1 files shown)


danielrainer:
For context: Corruption for me always occurred when shutting down the system. This is on btrfs with fish as the login shell.

src/history.rs
```diff
@@ -154,14 +154,18 @@ pub const VACUUM_FREQUENCY: usize = 25;
...
 /// If the size of `buffer` is at least `min_size`, output the contents `buffer` to `fd`,
 /// and clear the string.
-fn flush_to_fd(buffer: &mut Vec<u8>, fd: RawFd, min_size: usize) -> std::io::Result<()> {
+fn flush_to_file(buffer: &mut Vec<u8>, file: &mut File, min_size: usize) -> std::io::Result<()> {
     if buffer.is_empty() || buffer.len() < min_size {
         return Ok(());
     }
 
-    write_loop(&fd, buffer)?;
+    file.write_all(buffer)?;
```
dpc, line 162: `buffer.flush()` should be done to check for errors.

src/history.rs
```diff
@@ -154,14 +154,18 @@ pub const VACUUM_FREQUENCY: usize = 25;
...
     buffer.clear();
-    return Ok(());
+    // Ensure that the file content is actually written to persistent storage.
+    // This might be unnecessary.
+    // It is used to help eliminate possible causes for
+    // https://github.com/fish-shell/fish-shell/issues/10300
+    file.sync_all()
```
dpc, line 168: This might lead to slowdown (I mean - maybe a millisecond, could be more in some situations where the storage is odd (like NFS)). Not sure if it's a problem, just pointing it out.

src/history.rs
dpc: Huh. This function is called `flush_to_x` and it doesn't do a flush already?

dpc: I see. Technically it still needs to be called then.

danielrainer: I added it for completeness.

src/history.rs
danielrainer: I agree. The original idea was to just keep the changes as simple as possible, but I modified the code now to be a bit more sensible.

In the case where a new history file is written, there is only on flush+sync call at the end of the loop now.

In the other case, where items are appended and a counter is incremented for each item, I think it makes sense to flush+sync after every item, to ensure that the counter and the file system are consistent. According to a comment there the expected case there is to only append a single item, which would mean that there is no increase in system calls. If `save_internal_via_appending` is called with a large number of items to be saved, this approach might have to be reconsidered, but I think it will be hard to ensure that `first_unwritten_new_item_index` is consistent with the file system without one flush+sync per item.

src/history.rs
```diff
@@ -894,11 +906,12 @@ impl HistoryImpl {
...
-                        &mut buffer,
-                        history_file.as_raw_fd(),
-                        HISTORY_OUTPUT_BUFFER_SIZE,
-                    );
+                    // Ensure that each item is written individually and makes it to storage.
+                    // This is done to keep `self.first_unwritten_new_item_index` consistent with
+                    // the file system.
+                    // Flushing and syncing each iteration adds overhead,
+                    // but hopefully there are not that many items to write when appending.
+                    res = drain_buffer_into_file_and_flush(&mut buffer, &mut history_file);
```
krobelus, line 914: If no one is unplugging the disk, I wonder how this could possibly help.
I think the next fish should still see the updated file, regardless of whether we flush or not.

I guess I might be wrong: the size change might somehow be visible before the content change, which would lead to trailing zeroes. But I'm not sure if it still happens on typical file systems, see https://unix.stackexchange.com/questions/588586/why-does-closing-a-file-wait-for-sync-when-overwriting-a-file-but-not-when-crea
What's your file system and mount options?

Unrelated to this, if we receive `SIGTERM` at the wrong time, we might exit after a half-done write.
If that's the problem, we can work around by blocking the signal, or better, integrate it in the event loop like `SIGHUP`.

src/history.rs
```diff
@@ -827,6 +838,7 @@ impl HistoryImpl {
 
         // Get the path to the real history file.
         let Some(history_path) = history_filename(&self.name, L!("")) else {
+            // TODO: Why have a successful return here?
```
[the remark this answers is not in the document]
danielrainer: Ah yes, that makes sense.

I added some documentation and restructured the function a bit, to make the control flow more explicit. It also returns actual error messages now, the logs will contain more information if any error occurs in this function or its callees.

src/history.rs
danielrainer: Your point about might be valid. I would have to investigate the internals to know more.

One scenario which I imagined could pose a problem is running into size limitations (e.g. a full drive). If flushing+syncing after every entry and then incrementing the counter, we can ensure that if the counter is at a certain value, all previous entries have actually been written. An alternative approach would be to write all entries first without syncing, then syncing once after the loop, and if successful increase the value of `first_unwritten_new_item_index` by the amount of items written. Both of these approaches could result in duplicate entries, but not in skipped ones. At least in the case of `save_internal_via_appending` this seems useful, because if this function fails, `save_internal_via_rewrite` is attempted next, which takes its information about which items to write from `first_unwritten_new_item_index`. I have not checked the logic which actually uses the history file contents, but I would assume that duplicates are better than missing items. If we don not know for sure that this syncing is useless, I would take the cautious approach of leaving it in unless it has significant disadvantages.

Considering signals is definitely useful. These seem like a plausible cause for the corruption. I have not looked into fish's signal handling at all yet, but handling `SIGTERM` seems especially relevant in connection with the corruption occurring in connection with a system shutdown/reboot.

> What's your file system and mount options?

```text
btrfs           rw,noatime,compress=lzo,ssd,space_cache,subvolid=256,subvol=/@root,subvol=@root 0 0
btrfs           rw,noatime,compress=lzo,ssd,space_cache,subvolid=257,subvol=/@home,subvol=@home 0 0
```
Reading the issue it seems that corruption occurs on several file systems. There might of course be more than one cause for it.

faho:
First: As a general thing, please avoid back-merging master into your PR unless there are some changes there that you need. It's easier to review and merge without the merge conflicts.

----

So, the gist here is basically running fsync() after writing the history file?

That sounds like a good idea, but I'm not sure it causes all the issues here.

@danielrainer, have you been seeing a *missing* history file or one that contains NUL somewhere?

danielrainer:
> First: As a general thing, please avoid back-merging master into your PR unless there are some changes there that you need. It's easier to review and merge without the merge conflicts.

Yes, I see your point. In this case, the first merge was because I noticed the CI pipeline failing, so I merged in a commit from master which passed the pipeline, to ensure that the issues are not caused by my changes. The second merge was to get your fix from #11332, which makes development more convenient but is not necessary. Normally I'd rebase, but I did not want to modify the branch history while someone might be testing specific commits on this branch. If you'd like I can rebase my changes to get rid of the merges.

> So, the gist here is basically running fsync() after writing the history file?

Yes.

> That sounds like a good idea, but I'm not sure it causes all the issues here.

I agree, judging from the various reports in #10300 and the issues closed in favor of it I consider it quite plausible that multiple problems might cause history corruption. As suggested in https://github.com/fish-shell/fish-shell/pull/11330#discussion_r2018065443, signal handling might be problematic as well, especially for signals sent related to termination of the shell. For me, the corruption always occurred in conjunction with a system shutdown, so having a look at signal handling in this context seems worthwhile.
 
> @danielrainer, have you been seeing a _missing_ history file or one that contains NUL somewhere?

For me it has always been a sequence of a few tens of NUL bytes in the history file around shutdown commands, which resulted in fish not showing any completions from the history. Otherwise the history file remained intact, and removing the NUL bytes or waiting for fish to rewrite the history file fixed the issue.

src/history.rs
krobelus: full disk sounds unlikely to be the issue here.
If it is really possible to write metadata but not contents, that should be easy to reproduce artificially.

danielrainer: Yes, the corruption issues have been unrelated to full disks. It's just one situation I thought of when considering potential issues about incrementing the counter without ensuring that the corresponding history entries are actually persisted. For me, one plausible scenario in which code could break is sufficient to try to fix the code, so that's what I tried to do there.

krobelus: I don't understand how lack of `fsync` could cause corruption so I wouldn't add it without verifying that

danielrainer: It might not cause corruption in the sense of adding extra NUL bytes, but it could plausibly result in the loss of history items, e.g. if fish tries to append a few items to the file in the loop and increments the counter each iteration. Without `fsync`, the changes are buffered and only written out later on. So if the disk is full at the time the actual write to disk is attempted (or some other issue prevents the write from succeeding), the write will fail, but fish would assume that the items have been written due to the value of `first_unwritten_new_item_index`. Maybe some other part of the code would catch that and fix the history if possible, but if that is the case, I think it would be appropriate to clearly document how this happens if consistency is not ensured locally in the append loop.

krobelus: sorry yeah we should definitely try to fix any inconsistent state that happens when we're out of disk space.
But we should start by testing it. Maybe a sync alone is either not enough or irrelevant to the issue.

faho: A sync also can't really hurt, and we've had trouble reproducing this.

I'd merge the sync (I'm not sure about the attached cleanup), and then see if people are still seeing corruption.

danielrainer: > But we should start by testing it. Maybe a sync alone is either not enough or irrelevant to the issue.

I would love to have tests as well, but so far I have not found a way of reliably reproducing any corruption issues. I have not tested what happens when running out of storage. That could be done, but is time consuming and annoying to automate, especially if we want to test different file systems (and possibly even storage media). And the actual observed corruption is not caused by out-of-storage situations, so even with such tests we would not know if the changes here are relevant for the observed corruption. My reasoning is that if we have no way of reproducing the issues, the next best thing we can do is to write more conservative code which prevents some issues, and add logging to hopefully simplify debugging any remaining issues if the arise. This is what I have tried to do in this PR.

> I'd merge the sync (I'm not sure about the attached cleanup)

Is there anything specific you dislike about the changes or do they just not provide sufficient value to bother with including them?

danielrainer: As I see it, the changes currently in this PR do these things:
1. Restructure some control flow to have more suitable and understandable points for syncing and providing error messages.
2. Actually perform the flush+sync at the relevant spots.
3. Improve error handling in `save_internal_via_appending`.
4. Fix some trivial things in comments.

The first two are somewhat interdependent and I don't think it would be an improvement to split them up. The third and fourth point could be removed, if you'd like. The changes included here might look like more than they actually are due to the way indentation changes are displayed by line-based diff tools.

krobelus: yeah `git show -w` is the way to go.
Definitely get rid of the merge commits if you have the time, if we want to keep old commits beyond the garbage collection horizon there's other ways to do that.

faho:
Alright, let's merge this and see how it shakes out.

I'm not sure if we want to add it to 4.0.2 - it's a valuable fix, but also some new filesystem interaction.

zanchey:
I think we should take this for 4.0.2 - it's a real issue that people are seeing

krobelus:
I don't think we have a reason to believe that the `fsync()` added here helps, I think we should revert it until we have a reason to call `fsync()`.
(FWIW the problem [reoccurred](https://github.com/fish-shell/fish-shell/issues/10300#issuecomment-2876718382)).

danielrainer:
Yes, although I think it is still useful to `fsync` once at the end. The code for appending should probably be rewritten such that it writes all items to a in-memory buffer, appends that buffer to the history file, and fsyncs.

Two things which definitely cause race conditions in the current code:
- The rewrite path reads the history file without locking it (to get the content for the tmp file)
- Locking gets disabled when it takes too long to obtain a lock (see https://github.com/fish-shell/fish-shell/pull/11492#issuecomment-2881522129)

The memory mapping is also questionable and I think is should be replaced by a normal read into a `Vec<u8>`.

krobelus:
On Wed, May 14, 2025 at 01:59:02PM -0700, danielrainer wrote:
> danielrainer left a comment (fish-shell/fish-shell#11330)
> 
> Yes, although I think it is still useful to `fsync` once at the end.

Why? What about the other places where we write a file, are they not affected?

> The code for appending should probably be rewritten such that it
> writes all items to a in-memory buffer, appends that buffer to the
> history file, and fsyncs.
> 
> Two things which definitely cause race conditions in the current code:
> - The rewrite path reads the history file without locking it (to get the content for the tmp file)
> - Locking gets disabled when it takes too long to obtain a lock (see https://github.com/fish-shell/fish-shell/pull/11492#issuecomment-2881522129)
> 
> The memory mapping is also questionable and I think is should be replaced by a normal read into a `Vec<u8>`.

I don't understand how this is related to fsync but let's discuss that in that PR.

Issue #11332, opened 2025-03-28
Fish version identifier only changes upon clean rebuild


danielrainer:
When developing fish, I like being able to check the version of the shell currently running, either via `echo $FISH_VERSION` or `/path/to/fish --version`. These version identifiers contain a prefix of the corresponding git commit hash (and indicate whether there were uncommited changes by marking the version as `dirty`). Unfortunately, changing the commit (e.g. by creating a new one, or checking out a different branch) and rebuilding does not update this version info, making it inaccurate.

Steps to reproduce:
1. Check out some commit (e.g. 11c7310f17ea2e8507ca4fc0f7444ef6a22a071f)
2. Run a clean build (`cargo clean && cargo build --release`)
3. `target/release/fish --version` prints `fish, version 4.0.1-548-g11c7310f1`
4. Check out a different commit (e.g. d88f5ddbaf5d8c60da2da7a5f7d4a316eb6240c3)
5. Build again, this time with the build artifacts from the previous build in place (`cargo build --release`)
6. `target/release/fish --version` still prints `fish, version 4.0.1-548-g11c7310f1`, instead of the desired `fish, version 4.0.1-547-gd88f5ddba`, which would be obtained from a clean build on this commit.

danielrainer:
Thank you for addressing this! I think it might make sense to trigger a rebuild in more cases. Definitely upon changes to `Cargo.toml`, `Cargo.lock`, and `build.rs`, since these have a direct effect on the compilation.

In addition, it might make sense to rebuild on any changes in the repo root, except for build artifacts. The tradeoff there is ensuring that the version info in the executable always matches the current commit (and dirty state) at the cost of having to rebuild even if nothing about the executable would change except for the version info. However, there is no need to build at all if the executable does not need to change, so the only situation I can think of where the mostly useless rebuilds would actually happen is when a language server is running while someone modifies files irrelevant for the executable.

danielrainer:
I can open a PR for either of these options, depending on the preferred approach.

krobelus:
> I can open a PR for either of these options, depending on the preferred approach.

sounds good; I think I'd go with explicit includes (of Cargo.toml
etc) rather than explicit excludes (like share/, in addition to the
implicit gitignored files) but either seems fine.

Don't forget the `printf` directory.

danielrainer:
Explicit includes are definitely much simpler. I just opened #11333 for that.

[...]

PR #11333, opened 2025-03-30
Rebuild on changes relevant to build artifacts

danielrainer:
## Description

If anything changes in one of the listed directories or files, trigger a rebuild.

Rebuilds are needed to update the version info built into the binaries. The version info includes an abbreviated git commit hash, as well as information about whether the repo was in a dirty state (uncommitted changes) when building. Changes to files not explicitly listed will not trigger a rebuild, so to get accurate version info upon such changes (also when checking out a different commit which does not differ in any of the listed files), a rebuild needs to be triggered manually (e.g. via `cargo clean` or `touch build.rs`).

Fixes issue #11332

## TODOs:
- [ ] Check if any other files/directories should be included

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -48,8 +48,7 @@ fn main() {
         let targetman = cman.as_path().join("target").join("man");
         build_man(&targetman);
     }
-    rsconf::rebuild_if_path_changed("src/libc.c");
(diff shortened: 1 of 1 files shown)


faho:
Thanks!

Issue #11278, opened 2025-03-14
Regression: After updated to Rust, `bind -k nul` does not work


Ddystopia:
Hello, I just updated to Rust rewrite. Last 2 years I was not updating. 

I am using Alacritty, and have that in `fish_user_key_bindings.fish`:

```fish
function fish_user_key_bindings
  fish_vi_key_bindings
  bind -k nul -M insert accept-autosuggestion
end
```

Before, when I hit `ctrl-space`, autosuggestion was triggered. It worked fine for the whole time I used fish. Now, it doesn't work. `ctrl-n`, default binding for accepting autosuggesion, works.

Do you need more information from me?

faho:
Fish now uses the kitty keyboard protocol in alacritty, which means ctrl-space is actually distinguishable from a number of other key combinations (like ctrl-2).

Use

    bind ctrl-space accept-autosuggestion

Ddystopia:
I just tried your suggestion and it doesn't work. Issue should probably be reopened.

Ddystopia:
Oh, it strange. If I put this line before `fish_vi_key_bindings`, it works. But not after, and not manually

faho:
When you run `bind ctrl-space -M insert`, that should print

```fish
bind --preset -M insert ctrl-space 'test -n "$(commandline)" && commandline -i " "'
bind -M insert ctrl-space accept-autosuggestion
```

That first line has "--preset", that second line doesn't, because that second line is *your* binding. Our bindings only touch presets, and so if you define it before or after running fish_vi_key_bindings doesn't matter. And I just tried with fish 4.0.1, and it doesn't.

So it is possible that you have something else fiddling with the bindings.

Issue #3176, opened 2016-06-28
fish doesn't understand the color white


mikez302:
I would like the current directory in my fish prompt to be light gray, as it was in bash before I started using fish. I tried running `set -U fish_color_cwd white` (since white is the closest color to light gray that is listed in the [`set_color` documentation](https://fishshell.com/docs/current/commands.html#set_color)) and it didn't do anything. The text was green after, just like it was before.

I then tried `set -U fish_color_cwd red` to see what happened and the text turned red, like I expected. Then I tried `set -U fish_color_cwd white --bold` and it became bold, but green again. It seems like it understands all of the colors I tried, but not white.

I am using fish 2.3.0 on OS X 10.11.5 and the built-in Terminal. Here is a screenshot demonstrating what is happening:
![white not working in fish shell](https://cloud.githubusercontent.com/assets/510781/16399572/9674a980-3c88-11e6-9b11-76d5379e1450.png)

floam:
You have customize the text color in your Terminal.app profile configuration. One of the whites or text/bold text colors in here is going to show that it's been set to bright green (other than the standard ANSI green/bright green):
<img width="676" alt="screenshot 2016-06-27 at 5 47 56 pm" src="https://cloud.githubusercontent.com/assets/291142/16400370/873b142a-3c8f-11e6-83de-7827528ad955.png">

floam:
(Or, you are using Apple's "Homebrew" profile which has green for the unstyled text already)

faho:
@mikez302: To test if floam is right, try `echo (set_color white) banana`.

Also, you can pass RGB values to set_color and the color variables and fish will do the right thing (or try to).

floam:
I'm not right. Of course the "Last login" output there is not colored with ASCII escapes. He's probably using the included Homebrew profile which has both Text and Bold Text as green, like I first guessed... 😒

krader1961:
The important thing to keep in mind is that color names like "white" are simply mapped to the corresponding [ANSI X3.64 SGR sequence](https://en.wikipedia.org/wiki/ANSI_escape_code). Those sequences map each color name to a number. The terminal is free to map the number to whatever color it likes and different terminals use different default colors. However, in your case I strongly suspect that @floam is correct and you've reconfigured your terminal so that "white" appears as bright green.

On my terminal (iTerm2 with the default color palette) I get the expected result:

[...]

Issue #789, opened 2013-05-20
unicode \u23CE ⏎ character preceding every prompt in ConEmu


samvasko:
Everytime I run fish using ssh from cmd on windows. I get this character at the end.
⏎
In cmd it just looks like crossed box. Not sure if that is bug on your side, but can you help with some workaround?

Maximus5:
When I'm trying to run fish, it shows the `?` even **before first prompt**. And before each next prompt of course... It's unclear to me, why fish shows it, what "output from a command" do you mean?

![2016-03-16_18-35-28](https://cloud.githubusercontent.com/assets/1222388/13830343/1850cbe8-ebdd-11e5-88b3-7765eeaaae6f.png)

krader1961:
What OS are you running fish on? How are you connecting to that fish shell?

"Output from command" means just that, the output from the most recent command that was run. For example, consider the difference between `echo hello` and `echo -n hello`. The latter will cause fish to include the down-left-arrow (aka carriage-return) symbol to the output while the former will not.

As for why you're seeing that before the first prompt I can only guess. A likely explanation is something in your config.fish is writing to the terminal without terminating the output with a newline.

Maximus5:
Where the `config.fish` may be located? There is no such file in my cygwin's /home. Also, I did not configure anything special for fish, just "installed" via cygwin setup.

OS: Windows 8, Windows 8.1 x64.
Fish.exe from latest cygwin x86.

I run `t:\cygwin\bin\fish.exe -l -i` directly from `Win+R`.
Also, the screenshot was created when almost the same command was started inside ConEmu.
If you are interested about strange `/fish_shmem_1049690: Permission denied` - that happens every time few hours before, but is not reproduced ATM.

Maximus5:
OK, just found two `config.fish` files and append simple echo to the end of each. Get the following:

![2016-03-17_02-12-29](https://cloud.githubusercontent.com/assets/1222388/13831587/d6dce760-ebe5-11e5-9358-9577b8e8ae8d.png)

So, I suppose the problem is inside fish's prompt setting. How it is configured?

krader1961:
The easiest way to configure things like the prompt is to run the `fish_config` command. Once you have more experience you can create a **fish_prompt** function to have total control over the prompt.

[...]

PR #11342, opened 2025-04-01
Remove "bind -k" terminfo key names

krobelus:
I don't think there's a relevant terminal where the "bind -k" notation is
still needed. The remaining reason to keep it is backwards compatibility.

But "bind -k" is already subtly broken on terminals that implement either
of modifyOtherKeys, application keypad mode or the kitty keyboard protocol,
since those alter the byte sequences (see #11278).

Having it randomly not work might do more harm than good. Remove it.

This is meant go into 4.1, which means that users who switch back and forth
between 4.1 and 4.0 can already use the new notation.

If someone wants to use the bind config for a wider range of versions they
could use "bind -k 2>/dev/null" etc.


krobelus:
On Tue, Apr 01, 2025 at 08:17:50AM -0700, Fabian Boehm wrote:
>              'h' => opts.print_help = true,
> -            'k' => opts.use_terminfo = true,
> -            'K' => opts.mode = BIND_KEY_NAMES,
> 
> This will remove these options so it'll fall back to the "unknown option" error.
> 
> I think we should make these print a more specific error that tells you to rewrite your bindings.

Yeah actually we should probably keep "bind --key-names" around (with all named keys... modifiers should probably get a separate flag)
I'm going with something like

	bind -k nul 'echo foo'
	# CHECKERR: bind: the -k/--key key name style is no longer supported. See `bind --key-names` for key names


> -    return Box::new([
> -        terminfo_add!(key_a1), terminfo_add!(key_a3), terminfo_add!(key_b2),
> -        terminfo_add!(key_backspace), terminfo_add!(key_beg), terminfo_add!(key_btab),
> 
> Does that mean we can remove the keys from terminal.rs too?

yeah. I have started to look at removing all of terminfo.
I need to check again but it looks like we use terminfo for some workarounds
for terminals that are probably still not fixed, like
58347d494a (update PROMPT_SP heuristic, 2016-12-23) https://github.com/fish-shell/fish-shell/issues/789
3669805627 (Improve compatibility with 0-16 color terminals., 2016-07-21) https://github.com/fish-shell/fish-shell/issues/3176

faho:
> I need to check again but it looks like we use terminfo for some workarounds
for terminals that are probably still not fixed

The biggest issue with terminfo is how often it is just plain wrong.

I could not find a current terminal that has the xn capability unset correctly. You can *find* definitions that have it turned off, sure. You'll find an outdated "wezterm" definition (but wezterm sets $TERM to xterm-256color), you'll find "xterm+direct" (but xterm doesn't set that either). And you will find literally hundreds of museum pieces like the "wy350" from 1985.

I can't find the terminfo definition for conemu (might be one called "msys"?), but if that's the only one we're better off hardcoding a hack for it or not supporting it.

The only thing that is an even vaguely sensible use of terminfo is to check colors, and I don't know of an existing current terminal that does not understand the 256 color sequences (cool-retro-term ignores them because it doesn't do color, but I do not know of a terminal that is actually limited to 16 colors and barfs on these). We could just allow "dumb" and "xterm-mono" to turn off color.

krobelus:
> We could just allow "dumb" and "xterm-mono" to turn off color,

that sounds good, because it's unlikely that these are set by accident.
(of course the recommended approach is still `fish_config theme choose None`).

> I don't know of an existing current terminal that does not understand the 256 color sequences

that's good to know.

I was also wondering what's the deal with `set_a_foreground` / `set_foreground`.
But given that our fallback doesn't define `set_foreground`, it might be safe to assume the `set_a_foreground` one is widely supported.

faho:
> of course the recommended approach is still fish_config theme choose None

(different things - that turns off syntax highlighting, but what if you can't have color escapes at all?)

> But given that our fallback doesn't define set_foreground, it might be safe to assume the set_a_foreground one is widely supported

Here's a full list of all terminfo definitions on my system that have setf but not setaf:

```
ctrm
emots
gs6300
hft-old
ibm+color
mgterm
ncr260wy325pp
ncr260wy325wpp
ncr260wy350pp
ncr260wy350wpp
qansi
qansi-g
qansi-m
qansi-t
qansi-w
qnx
qnx4
qnxm
qnxt
qnxt4
qnxtmono
qnxw
tek4205
tw100
wy350
wy350-vb
wy350-w
wy350-wvb
wy370
wy370-101k
wy370-105k
wy370-EPC
wy370-nk
wy370-rv
wy370-vb
wy370-w
wy370-wvb
wyse350
wyse350-vb
wyse350-w
wyse350-wvb
wyse370
```

Yes it's all museum pieces.

Here's the script I used:

```fish
for term in (path filter -rf /usr/share/terminfo/** | path basename)
    tput -T $term setf &>/dev/null && not tput -T $term setaf &>/dev/null && echo $term
end 
```

PR #11345, opened 2025-04-02
Stop reading terminfo database

### LINKED ISSUE #11344: Stop using terminfo
As I've alluded to before, I think we should stop listening to the terminfo database *entirely*. This is because:

- we barely use it
- it is frequently wrong
- it is massively incomplete and shows no signs of improving
- it does not have versioning so old entries remain in use even on new terminals
- it would remove a dependency and make fish more resilient to misconfiguration and over ssh
- it is a source of confusion

We've already stopped using its keys except for when you use `bind -k`, and that's slated to be removed with #11342. The rest of the sequences we use basically do not differ usefully between terminals in actual use.

Quantifying this is pretty annoying, mostly because terminfo is full of ancient guff and misinformation. A lot of the entries are made up to describe terminals that use xterm-256color (these are wrong very often - see "konsole" and "ms-terminal"). I will be discounting these entirely. If an entry isn't used by a terminal *by default* it isn't used (one exception being tmux-256color - tmux at least used to tell people to switch to that if they had it)

The sequences we use are:

https://github.com/fish-shell/fish-shell/blob/3fe6d4922463eacbebe65d38a12b4a1c5a907c80/src/terminal.rs#L42-L80

Now, excluding the keys, the terminfo entries that are *exactly the same* as xterm-256color are: alacritty foot ghostty hterm-256color rio terminology xterm-ghostty

Excluding the sequences that we only *remove* from text for measuring runs adds contour mintty xterm-kitty - a lot of these just don't have "enter_secure_mode" and similar. I don't believe I've ever seen them used anywhere.

And then there are a lot of differences that are just incorrect. They're either actively wrong (and don't work like that), or they're useless because the terminal also understands the xterm-256color sequence. These include e.g.

```
tmux-256color
        cursor_up: '\EM', '\E[A'.
        exit_attribute_mode: '\E[m^O', '\E(B\E[m'.
```

All of these work in tmux, so we can just treat it like xterm-256color. Sometimes they're also just cosmetic differences - rxvt-unicode-256color lists ` exit_attribute_mode: '\E[m\E(B` instead of `'\E(B\E[m'`, but that's just the two parts of the sequence swapped.

Some sequences just do not differ at all in any terminal made after the 80s - e.g. "init_tabs" and "auto_right_margin" don't have any differences in any of the entries that feature at least 256 colors (as a proxy for "modern"). There are terminals that list a difference for "eat_newline_glitch", I don't believe them (because it includes wezterm, which uses xterm-256color). In other cases the sequence is marked as "NULL" but the terminal correctly ignores it, so we can just use it.

If anyone wants, I have written a script that I could share that lists a bunch of differences, I can go through each and every one and explain why exactly it is useless.

The only vaguely sensible remaining use of terminfo is to turn off color. That's a bit weird for us because it makes any use of `set_color` non-functional. We might want to support $NO_COLOR instead. A better solution is the "Almost no color" theme because we can't really do autosuggestions or marking the current entry in the pager without color.

So I propose we:

- remove the terminfo crate dependency
- keep the xterm-256color sequences for what we use them for (but remove init_tabs/auto_right_margin/eat_newline_glitch and other pointless ones)
- if necessary add our own quirks like we already do for things that terminfo doesn't specify

The only sequences where we even add any arguments are set_a_foreground/background, and in those cases we don't need the terminfo DSL, we can just `sprintf("\x1b[%dm", number)`. We already do that for 24-bit color and it's worked fine for years.

This will cut off fish use on the Data General Dasher D220 and Wyse WY-350 for good. To be honest: I am fine with that. It is not a usecase I am interested in. Retrocomputing is neat, but I ain't spendin' any time on it. If it turns out we don't agree on that, we presumably need to keep using terminfo, but I don't believe fish is functional there to begin with.

krobelus:
(changes are messy, will clean up later)

* assume soft wrapping is supported (auto_right_margin)
* assume tab width is 8
* assume there is no weird newline glitch behavior
  * TODO this possibly breaks ConHost or Windows Console, see 58347d494a
    (update PROMPT_SP heuristic, 2016-12-23)
    https://github.com/fish-shell/fish-shell/issues/789
* assume it's fine to output ANSI colors, 256 colors and RGB colors
  * TODO test if this breaks macOS Terminal.app
    ("Enable 256 colors if :envvar:`TERM` contains "xterm", except for known
    exceptions (like MacOS 10.6 Terminal.app)")
    See also 3669805627 (Improve compatibility with 0-16 color terminals.,
    2016-07-21) https://github.com/fish-shell/fish-shell/issues/3176
* Remove `fish_term256` and `fish_term24bit` knobs.
  * TODO find out if they are useful and add them back.
* Now that `TERM` is ignored, add back knobs to turn off color.
  As a first approximation:
  * `TERM=dumb`
  * `TERM=linux-m`
  * `TERM=xterm-mono`
  * `NO_COLOR=1`
  * TODO: `fish_color=0`?
* TODO: decide what to do with "TERM=dumb".
  There is at least one easily reproducible (preexisting) crash.
  What are the requirements and use cases for `TERM=dumb`?

Closes #11344


03040000926 (commented):
Great news about the new update on my phone

faho:
> Remove fish_term256 and fish_term24bit knobs.

> TODO find out if they are useful and add them back.

I know of one use, and that is turning them off to force fish to defer to the terminal palette, for people who want to switch that (e.g. to base16).

krobelus:
ok so I'm assuming `set fish_term256 0` should mean: colors 0-15 are output and colors 16-255 are not printed

faho:
> ok so I'm assuming set fish_term256 0 should mean: colors 0-15 are output and colors 16-255 are not printed

"everything is mapped to the closest color in 0-15"

i.e. `set_color 554433` doesn't *fail*, it falls back to the same thing as `set_color yellow`.

Which ends up picking weird semi-random colors, but only from the palette. So you'd set your palette and then pick the color variables to match, but at no point is anything unstyled or eliminated.

mqudsi:
Should `IS_DUMB` and `is_dumb()` take into account `common::is_console_session()`?

krobelus:
Probably not, I'm not aware of any problems in the console.
I don't really know what are the use cases for `TERM=dumb`.
I guess it works well for terminals that print their output to paper.
fish code suggests the only console quirk is that it doesn't handle multli-byte characters, even though we call `unicode_start` :shrug:

krobelus:
I added `doc_src/terminal-compatibility.rst`; it needs more work to list the exact control sequences we need. But I guess that's not a blocker.

The `doc_src` and changelog changes could probably use a review.

I'll test for ConEmu regressions tomorrow. Putty and Terminal.app and some others seem fine

CHANGELOG.rst
```diff
@@ -42,6 +44,9 @@ Completions
 
 Improved terminal support
 ^^^^^^^^^^^^^^^^^^^^^^^^^
+- ``man terminal-compatibility`` documents required and optional features.
+- fish supports `set NO_COLOR 1 <https://no-color.org/>`_ to completely disable color output (syntax highlighting, prompt, builting output, error messages).
+  However grayscale colors (for autosuggestions) are still enabled.
```
faho, line 49: A different solution is to turn off autosuggestions and force the pager selection to use underline.

That's probably more in the spirit of NO_COLOR.

doc_src/terminal-compatibility.rst
```diff
@@ -0,0 +1,156 @@
...
+Here is a table of the relevant commands and features.
+
+.. list-table::
+   :widths: auto
+   :header-rows: 1
+
+   * - Required?
+     - Description
+     - Origin/Standard
+   * - Required
```
faho, line 21: This should clarify what "Required" vs "Optional" means.

As far as I can tell for *all* of these we require that they're not misparsed, but for the "require" ones we need them for some core functionality.

doc_src/terminal-compatibility.rst
```diff
@@ -0,0 +1,156 @@
...
+     - ``\e[K`` -- clear to end of line (Terminfo: ``clr_eol`` / ``el``)
+     - VT100
+   * - Required
+     - ``\e[J`` -- clear to the end of screen (Terminfo: ``clr_eos`` / ``ed``)
+     - VT100
+   * - Required
+     - ``\e[2J`` -- clear the entire screen (Terminfo: ``clear_screen`` / ``clear``)
+     - VT100
+   * - Required
+     - ``\e[0c`` -- request primary device attribute. The terminal must respond. An example response is ``^[[?62;4;22;28c``.
```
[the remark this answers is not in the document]
faho: Also this should note that we don't care about the exact response.

(I'm still not sold on querying and how we do it. if we have to require synchronized output *and* the alternate screen *and* primary device attribute to do querying it seems simpler to just send the things we query for)

faho (changes requested):

doc_src/cmds/set_color.rst
```diff
@@ -72,24 +74,3 @@ Examples
...
-
-In particular it will:
-
-- Enable 256 colors if :envvar:`TERM` contains "xterm", except for known exceptions (like MacOS 10.6 Terminal.app)
-- Enable 24-bit ("true-color") even if the $TERM entry only reports 256 colors. This includes modern xterm, VTE-based terminals like Gnome Terminal, Konsole and iTerm2.
-- Detect support for italics, dim, reverse and other modes.
-
-If terminfo reports 256 color support for a terminal, 256 color support will always be enabled.
-
-To force true-color support on or off, set :envvar:`fish_term24bit` to "1" for on and 0 for off - ``set -g fish_term24bit 1``.
```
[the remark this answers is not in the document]
krobelus: yeah done -- also linked `fish_term256` from `set_color.rst` I guess -- this wasn't linked before

doc_src/terminal-compatibility.rst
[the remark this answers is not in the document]
krobelus: > Do we want to hang if the terminal does not respond?

My plan is to make it print an error after a short timeout. Haven't gotten around to that one.

> (I'm still not sold on querying and how we do it. if we have to require synchronized output *and* the alternate screen *and* primary device attribute to do querying it seems simpler to just send the things we query for)

Yes, most features can be queried by just trying them and then checking if they worked
(report cursor position is one of them, and incidentially it can also be used to check
if other features are present, i.e. if they moved the cursor).

I agree that behavior testing is better than feature-flag testing.
I haven't found a way to do this for scroll-forward (which is the only thing we use today).

The awkward dance with synchronized output + alternate screen is supposed to be temporary.
It was originally a workaround for the Putty-family (I think ConHost is similar, not sure).
Putty merged a patch last week to parse DCS sequences (which are basically OSC sequences with one character changed).
I wonder how many remaining terminal can't deal with DCS sequences, like

	printf '\x1bP+q616d\x1b\\'; cat

CHANGELOG.rst
krobelus: I have not used the `NO_COLOR` behavior myself much yet,
but I do use the None theme, so it's possible that I want `NO_COLOR`.
I wouldn't use it without autosuggestions.

No strong opinion on the pager selection; I'd think that reverse video is allowed in `NO_COLOR` but haven't checked.

It's probably true that the `NO_COLOR` spec doesn't want anything but black-on-white (or the other way round).
To not openly violate NO_COLOR, maybe we should use a different name? `set -g fish_grayscale 1`?

doc_src/terminal-compatibility.rst
krobelus: > This should clarify what "Required" vs "Optional" means.

Ok I'll make it say something like: implementations for the "Required" ones need to be sound (i.e. correct) and complete.
Implementations for the "Optional" ones must also be sound but need not be complete (can be completely absent).

So this is only talking about what's needed to avoid glitches.

But of course users will notice if color doesn't work.
I'm happy to move some more to required, or define tiers (because a lot of the optional ones are quite minor) if someone has opinions

> As far as I can tell for *all* of these we require that they're not misparsed,

basically yes, but this is technically contained within
"parse and ignore arbitrary `CSI` and `OSC` sequences"
when combined with the implicit assumption
«don't misinterpret things like "CSI = 5 u" as DECRC ("CSI u")»
that should go without saying.

Of course that's more helpful for someone working on a new terminal rather than an existing one.

faho: A few of the ones that are marked as required allow ignoring, which I don't believe is "complete"?

As far as I can tell the difference is "can be ignored" and "needs to be acted upon in the prescribed manner" (and not be confused with something else)? E.g. if you ignored cursor_down, you would get glitches because the cursor is misplaced. If you ignored OSC7, you just won't be able to act on the working directory.

A weird in-between is color - if you're making a terminal that doesn't do color (e.g. as an art piece like cool-retro-term) that's fine, but if you reset styling on rgb sequences instead it's broken.

krobelus: Gotta run now, I'll think about it later.
But as I see it, if any of the ones in Required is missing, there will be glitches. I don't think any of those can be moved to Optional.

faho: I *think* the confusion is that you've listed "parse OSC format" and "parse CSI format" as required, which I would understand to mean "all CSI/OSC commands are required to be implemented".

The OSC commands are, as far as I can tell, all "optional" (can be stubbed out) and the CSI commands range from extremely required (cursor movement) to optional (italics).

I also would not use "sound (i.e. correct)". Just say "correct" instead if you feel the need to define the term with a better understood synonym.

What I would say is roughly: "A terminal that wants to be able to run fish needs to be prepared to receive any of these sequences. 'Required' means it needs to act on them in the prescribed manner, while 'Optional' sequences can also be ignored or partially (but correctly) handled. None of these sequences should be misparsed to do anything else or be shown on-screen.

We expect that the CSI and OSC formats are parsed correctly and may use more of those sequences in future, especially as optional enhancements.".

And then remove the "parse CSI" and "parse OSC" entries from the table.

----

Is it worth it to explain that we can work around your terminal if you give us something to identify it like $TERM_PROGRAM? Or do we not want to encourage that?

doc_src/language.rst
```diff
@@ -1727,9 +1725,18 @@ Fish also provides additional information through the values of certain environm
...
 
-   the type of the current terminal. When fish tries to determine how the terminal works - how many colors it supports, what sequences it sends for keys and other things - it looks at this variable and the corresponding information in the terminfo database (see ``man terminfo``).
+   the type of the current terminal.
 
-   Note: Typically this should not be changed as the terminal sets it to the correct value.
+   Only the following values affect fish, others are ignored.
+
+   Set it to ``xterm-16color`` to restrict fish to a color palette of 16 instead of 256 colors (unless force-enabled by :envvar:`fish_term256`) and to prevent fish from using 24-bite RGB colors (unless force-enabled by :envvar:`fish_term24bit`).
+
+   Set it to one of ``ansi-m``, ``linux-m`` or ``xterm-mono`` to get same effect as setting :envvar:`NO_COLOR``.
```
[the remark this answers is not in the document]
faho: Sounds good.

doc_src/terminal-compatibility.rst
krobelus: Ok, I have rewritten it a bit.
I'm leaning towards keeping it terse since this is reference documentation (for experts I guess), so it feels a bit weird to state the obvious.
There's probably space for a section in the way of a [tutorial/explanation/how-to-guide](https://ubuntu.com/blog/diataxis-a-new-foundation-for-canonical-documentation) on how these things work, but that can be a separate page, or blog post etc.

> Is it worth it to explain that we can work around your terminal if you give us something to identify it like $TERM_PROGRAM? Or do we not want to encourage that?

I have added an entry to the FAQ, asking to report bugs on glitches, I guess that could cover it.

doc_src/terminal-compatibility.rst
```diff
@@ -0,0 +1,179 @@
...
+
+.. list-table::
+   :widths: auto
+   :header-rows: 1
+
+   * - Sequence
+     - Description
+     - Origin
+   * - ``\t``
+     - move the cursor to the next tab stop (à 8 colums).
```
faho, line 68: typo: columns

doc_src/terminal-compatibility.rst
```diff
@@ -0,0 +1,179 @@
+Terminal Compatibility
+======================
+
+To make full use of your terminal, fish emits a number of control sequences and control characters.
```
faho, line 4: Fish doesn't "make full use of" the terminal - it doesn't even touch a lot of features (I keep hearing about box drawing and stuff, and we've never done any of that).

How about "Fish uses a number of control sequences and characters to implement its features"?

doc_src/faq.rst
```diff
@@ -340,6 +341,14 @@ This also means that a few things are unsupportable:
 - Non-monospace fonts - there is *no way* for fish to figure out what width a specific character has as it has no influence on the terminal's font rendering.
 - Different widths for multiple ambiguous width characters - there is no way for fish to know which width you assign to each character.
 
+.. _faq-unicode:
```
faho: Wrong link. Maybe "faq-terminal-glitches", tho this isn't linked anywhere and could also be left out.

doc_src/terminal-compatibility.rst
faho: This also lacks just a basic statement that these are our *requirements*.

Something like

> Fish requires certain terminal features to be present and won't work correctly without. In addition it tries to enable optional features and needs those sequences to either be handled or ignored.

doc_src/terminal-compatibility.rst
```diff
@@ -0,0 +1,179 @@
...
+
+The terminal must be able to parse Control Sequence Introducer (CSI) commands, Operating System Commands (OSC) and optionally Device Control Sequences (DCS).
+These are defined by ECMA-48.
+If a valid CSI, OSC or DCS sequence does not represent a command implemented by the terminal, the terminal will ignore it.
```
faho, line 8: I would just say "should ignore it" here, unless we're going full RFC "MUST".

doc_src/terminal-compatibility.rst
```diff
@@ -0,0 +1,179 @@
...
+Optional Commands
+-----------------
+
+.. list-table::
+   :widths: auto
+   :header-rows: 1
+
+   * - Sequence
+     - Description
+     - Origin
```
faho, line 66: I'm not sure the "Origin" column helps.

doc_src/terminal-compatibility.rst
faho: And also that we're *not* doing terminfo:

> Fish expects specific *sequences* to be handled and does not read terminfo. The required sequences are the same across all modern terminals, and the optional sequences are not typically mentioned in terminfo, so it is of no use to us.

Otherwise a terminal could get the idea that they can just add some capability to terminfo informing us that they're doing things differently, or file a bug saying we should check it.

This is on purpose.

doc_src/terminal-compatibility.rst
```diff
@@ -0,0 +1,179 @@
...
+     - clear to end of line (Terminfo: ``clr_eol`` / ``el``)
+     - VT100
+   * - ``\e[J``
+     - clear to the end of screen (Terminfo: ``clr_eos`` / ``ed``)
+     - VT100
+   * - ``\e[2J``
+     - clear the entire screen (Terminfo: ``clear_screen`` / ``clear``)
+     - VT100
+   * - ``\e[0c``
+     - request primary device attribute. The terminal must respond. An example response is ``\e[?62;4;22;28c``.
```
faho, line 51: Do we care about the contents of the response?

doc_src/terminal-compatibility.rst
krobelus: Yeah I fear that frequently mentioning Terminfo in the table could give the wrong impression.
We should leave out the capability name, and only include the TI code (``cud1``)

I've mainly stolen from https://codeberg.org/dnkl/foot/src/branch/master/doc/foot-ctlseqs.7.scd here,
looking through that I see that they barely mention terminfo.
There is a Name column with values like `CUD` which coincides with terminfo codes.
Actually it should be a separate column.. I was put off because the HTML table is already quite narrow but it looks fine.

TI code seems enough, we don't need multiple names for these.

---

> of no use to us

I left out the justification part since I'm not sure it's necessary for this document but feel free to add it later.
Technically it *could* be useful if we use it for detecting support for scroll forward (indn) and for literally nothing else.
But we should prefer the alternatives.

As a last resort, it doesn't seem terrible.
Not suggesting we do that now but maybe at some point in future, if it's worth the cost.

doc_src/terminal-compatibility.rst
krobelus: This is also stolen from `man foot-ctlseqs`, it helped me orient myself a bit;
My goal is to actually add more information here eventually: every command should link to its spec, to help (terminal) implementors.

doc_src/terminal-compatibility.rst
krobelus: no, I changed it to state only what we require (`CSI ?` prefix and `c` suffix).
I originally gave a real-world example because the parameters indicate capabilities
so just `CSI ? 0 c` might be wrong. But I guess that's not our problem,
people can see refer to the XTerm ctlseqs.html

doc_src/terminal-compatibility.rst
krobelus: This PR does both
1. stop using terminfo for anything
2. remove the actual terminfo dependency

if there is a chance we want to use terminfo selectively in future,
*and* removing the dependency now makes it hard to add it back later,
then we should perhaps keep it.
There's probably little downside to that.
I guess a user might see terminfo in the dependency tree and be surprised that fish doesn't actually rely on it.

doc_src/terminal-compatibility.rst
```diff
@@ -0,0 +1,188 @@
+Terminal Compatibility
```
krobelus, line 1: everything but soft-wrapping is about control sequences here,
maybe we should call it `terminal-control-sequences` or `terminal-ctlseqs`.

I wonder what else we could put in there.
`fish_handle_reflow` comes to mind (that's one of the remaining places where we do user-agent-sniffing)
But if we switch to a different solution here, it will likely be based on a control sequence as well.
So I'll try the rename.

src/builtins/set_color.rs
```diff
@@ -217,22 +207,19 @@ pub fn set_color(parser: &Parser, streams: &mut IoStreams, argv: &mut [&wstr]) -
 
     // Test if we have at least basic support for setting fonts, colors and related bits - otherwise
     // just give up...
-    let Some(term) = terminal::term() else {
-        return Err(STATUS_CMD_ERROR);
-    };
-    let Some(exit_attribute_mode) = &term.exit_attribute_mode else {
+    if is_dumb() {
         return Err(STATUS_CMD_ERROR);
```
krobelus, line 211: this is a behavior change;
previously `TERM=dumb fish -C 'set_color red'` would work
(but the same fish would not use color for prompt etc., so the red even leaks).

Maybe a use case for dumb is things like expect.

I think it's definitely called for to move away from `TERM=dumb`.
Its only upside is that it has some mindshare -- I guess if a user's terminal echoes a control sequence,
then they might remember `TERM=dumb` as an escape hatch.
(of course that's a band-aid; it makes fish much worse, and it's also unintuitive).

For the problem of "certain control sequences break my terminal"
we could instead provide a regex option like

	set -g fish_filter_ctlseqs '^\e[P.*' # Suppress DCS requests

I guess it's fine to keep `is_dumb()`, and have it emit only these:

	$ infocmp dumb
	#	Reconstructed via infocmp from file: /usr/share/terminfo/d/dumb
	dumb|80-column dumb tty,
		am,
		cols#80,
		bel=^G, cr=\r, cud1=\n, ind=\n,

But I'm not sure what's the use;
if it's for debugging terminal emulator behavior, there should be better options.

doc_src/terminal-compatibility.rst
faho: We don't link to any C library anymore, and the terminfo database is installed on 99% of systems (basically *just* some of the BSDs, sometimes).

So going back to terminfo, which I don't foresee us doing, should just be adding the dependency back.

I would want to be careful in merging this - test a bunch, get some buy-in, so we don't have to walk it back later.

src/builtins/set_color.rs
faho: > we could instead provide a regex option like

I don't think most users will understand that.

Maybe something more of a yes/no flag like "fish_my_terminal_is_bad" that turns off the optional things, or a flag like the kitty keyboard one.

I see this being less and less of an issue as time goes on.

> But I'm not sure what's the use;

Tbh most TERM values of "dumb", "vt220" or "ansi" are either misconfiguration or a tool being overly conservative in its defaults - e.g. systemd [sets vt220 for some things](https://github.com/systemd/systemd/issues/3342) (I think that's also the default in nspawn containers?).

doc_src/terminal-compatibility.rst
faho: An alternative, if we're not entirely sure this would work:

Add a feature flag. If it's set, always use the included xterm-256color definition (but apply any quirks).

That way we could add it and easily back it out if it turns out there's something that terminfo is still useful for. If it turns out (as I suspect) that there isn't, we eventually force the flag to on and remove the dependency.

krobelus: > I would want to be careful in merging this - test a bunch, get some buy-in, so we don't have to walk it back later.

I've done some testing.

So far the only problem I see is Terminal.app.

We use a `$TERM_PROGRAM` based workaround. If that one 
is not in effect, RGB colors like `set_color ff0000` will not work.
I think the main concern here is `ssh`. Possibly also `docker`.

I don't personally care (note that we have other workarounds hinging on `TERM_PROGRAM`); not sure what macOS users think.
Either way, I'll try to find a channel to ask Terminal.app to implement `XTVERSION` (if the quirks can't be fixed).

Maybe ignore this issue and see if macOS/Terminal.app users complain.
If that happens, we can easily add back terminfo and use it only for enabling RGB-colors (and grudgingly turn them off by default).

It will probably be fine because our default theme doesn't use any RGB colors.

---

Here's a test script I tried. It should touch almost all things that have changed.

For each terminal, I run this before and after this change and check if the
screen looks the same.

This approach is really dumb, your approach of diffing the terminfo and checking only differences is better, I think it was https://gist.github.com/faho/db2d7630a6d25492245c2340858c22de.

But I guess it still gives a finding (Prompt and zellij bugs), and a feeling of what terminals could be missing (?).
I'll test with the Windows/Mac ones later.

There's some terminals at the bottom that I might not test.

<details>

<summary> Test script </summary>

```fish
# clear_screen (TI clear)
echo CLEAR THIS
printf '\x1b[2J'

# clr_eos (TI ed)
echo "SHOULD HAVE CLEARED ABOVE"
echo "DO CLEAR THIS 1"
printf '\x1b[A' # up
printf '\x1b[J'

# clr_eol (TI el)
printf 'DONT CLEAR THIS DO CLEAR THIS 2'
printf '\x1b[16D' # left
printf '\x1b[K'
printf '\n'

# set_title (TI ts)
function fish_title
echo TEST TITLE
end

# init_tabs (TI it)
printf '        8 SPACES\n'
printf '\tTAB\n'

# auto_right_margin (TI am)
printf "%0$(math $COLUMNS + 5)d\n" 0
printf "^^^^^ wrap extent\n"

# text attributes, colors
echo (set_color --bold)BOLD(set_color normal)
echo (set_color --italics)ITALICS(set_color normal)
echo (set_color --dim)DIM(set_color normal)
echo (set_color --underline)UNDERLINE(set_color normal)
# N.B. we used to fallback on enter_standout_mode if reverse mode was not in terminfo.
echo (set_color --reverse)REVERSE(set_color normal)

# We have been using "set_a_foreground" if possible for many years,
# so we hope this works.
echo (set_color --bold red) ANSI RED (set_color normal)
echo (set_color --background=red) ANSI RED BG (set_color normal)

# On terminals that don't support RGB, this could fail to paint things
# red; we are considering adding exceptions (for Terminal.app and screen).
echo (set_color --bold ff0000) RGB BRIGHT RED (set_color normal)
echo (set_color --background=ff0000) RGB BRIGHT RED BG (set_color normal)

# This prints "CSI 38;5;200m". If it shows a different color, the terminal
# may not support 256 colors.
echo (fish_term24bit=0 set_color --bold ff00dd) PINK-200 (set_color normal)
echo (fish_term24bit=0 set_color --background=ff00dd) PINK-200 BG (set_color normal)

# Included just for comparison; this should be different from the PINK-200 above.
echo (set_color --bold brmagenta) BRMAGENTA (set_color normal)
echo (set_color --background=brmagenta) BRMAGENTA BG (set_color normal)
```

</details>

<details>

<summary> Tested terminals </summary>

- [x] abduco `TERM=dvtm-256color` (failed to render dim text before this change)
- [x] alacritty `TERM=alacritty` (colors look wrong before and after)
- [x] cool-retro-term `TERM=xterm`
- [x] dvtm `TERM=dvtm-256color` (failed to render dim text before this change, also exhibits random breakage before and after)
- [x] emacs M-x vterm `TERM-xterm-256color`
- [x] foot `TERM=foot`
- [x] ghostty `TERM=xterm-ghostty`
- [x] gnome-terminal `TERM=xterm-256color` (could also test other VTE-based ones)
- [x] JetBrains (IntelliJ/PyCharm) `TERM=xterm-256color`
- [ ] iTerm2
- [x] kgx (aka gnome-console) `TERM=linux` (interesting TERM value; now colors are no longer wrong, and italics are actually rendered)
- [x] kitty `TERM=xterm-kitty`
- [x] konsole
- [x] midnight commander (this one is mostly transparent; and it does have a correct CSI parser)
- [x] mosh `TERM=xterm-256color`
- [x] Prompt 3 `TERM=xterm-256color` (their CSI/DCS parser has minor issues but it looks like they have been addressing bug reports)
- [x] putty `TERM=xterm` (even though puttty terminfo exists)
- [x] rio `TERM=rio`
- [x] screen (with a $TERM-based workaround to disable RGB colors). This is used to fail to render italics before
- [x] st `TERM=st-256color`
- [ ] Terminal.app
- [x] terminology (claims to be xterm-256color)
- [x] tmux `TERM=tmux-256color`
- [x] urxvt `TERM=rxvt-unicode-256color`
- [x] Vim `:terminal` `TERM=xterm-256color`
- [x] VSCode `TERM=xterm-256color`
- [x] wezterm `TERM=xterm-256color`
- [ ] Windows ConHost
- [ ] Windows Terminal
- [x] xterm (note that `set_color --background=brmagenta` is now actually bright (105 instead of 45) because we assume xterm-256color, not xterm)
- [ ] zellij (inherits `TERM`) (there is a behavior difference that triggers [a Zellij bug](https://github.com/zellij-org/zellij/issues/4122))
- [x] emacs ansi-term `TERM=eterm-color` (behavior unchanged; but it's broken for other reasons, I wonder why emacs still advertises this and not `M-x vterm`)
- [x] fbterm `TERM=fbterm` (seems no more broken than before; it doesn't seem to have colors; it's sort of unmaintained)

I'm not sure if I want to test the following myself. I'd prefer if others do but I guess I can do mintty, termux, connectbot and xterm.js:
- [ ] amazon linux web console
- [ ] bitvise SSH
- [ ] connectbot
- [ ] domterm
- [ ] FreeBSD console
- [ ] juicessh
- [ ] mintty
- [ ] termux
- [ ] trinity-desktop-environment's fork of konsole
- [ ] xterm.js

<details>

krobelus: If it's just about RGB colors, we already have `fish_color24bit` for overriding.
If there is anything else that could possibly go wrong, a `respect-terminfo` feature flag (defaulting to off) might make sense.
Case in point: I realized that this PR changes the behavior for `TERM=xterm`, because that one doesn't have 256 colors.

We could interpret `TERM=xterm` as "don't use more than 16 colors", for compatibility.
But this interpretation has false positives, in particular for `xterm`, `putty` and `cool-retro-term` which set it by default. So I think I prefer only respecting `xterm-16color` by default.

Based on that conclusion, users might want to reach for setting `respect-terminfo`, to maintain stable behavior without fiddling with `TERM` (which might affect other programs).
Obviously we intend to freeze this feature flag eventually, but it will give users a chance to tell us why they want us to respect the `TERM` variable (and which ones).

faho: > Either way, I'll try to find a channel to ask Terminal.app to implement XTVERSION (if the quirks can't be fixed).

My experience is you'll have better luck teaching a brick wall japanese than getting Apple to do anything, but godspeed to you.

Anyway:

> Maybe ignore this issue and see if macOS/Terminal.app users complain.
If that happens, we can easily add back terminfo and use it only for enabling RGB-colors (and grudgingly turn them off by default).

We do not use terminfo for enabling RGB-colors!

Like.. we have code that checks max_colors, but like I said that code never ever triggers. Nobody uses the "-direct" entries, and everything else claims 256 colors. It especially won't help with Terminal.app because it uses TERM=xterm-256color.

We can add the other *quirks*, sure, but those are all based on env var sniffing, none of them check terminfo.

> This approach is really dumb, your approach of diffing the terminfo and checking only differences is better, I think it was https://gist.github.com/faho/db2d7630a6d25492245c2340858c22de.

Yup, that was it. Of course the list of sequences needs to be pruned, we're no longer using keys.

> Case in point: I realized that this PR changes the behavior for TERM=xterm, because that one doesn't have 256 colors.

I don't think it changes the behavior, because we've worked around TERM=xterm for years and assumed any "xterm" supports 256 colors.

I don't believe a single person ever complained.

Or do you mean it enables RGB in xterm-256color?

krobelus: > I don't think it changes the behavior,

ah my bad - it's not about 256 colors.
But this command changes behavior on xterm (it's visible with the combination of background *and* bright color but looking at the actual control sequences, it also affects non-bright colors):

    echo (set_color --background=brmagenta) BRMAGENTA BG (set_color normal)

Previously, we used the non-bright color here.
This is because `xterm` reports `colors#8` which we interpret as:

```rust
            // this allows the non-bright color to happen instead of no color working at all when a
            // bright is attempted when only colors 0-7 are supported.
            //
            // TODO: enter bold mode in builtin_set_color in the same circumstance- doing that combined
            // with what we do here, will make the brights actually work for virtual consoles/ancient
            // emulators.
            if term.max_colors == Some(8) && idx > 8 {
                idx -= 8;
            }
```

Anyway, this seems like a fine instance of this change since max_colors=8 was clearly a lie.

It's possible that we're making terminals worse that actually only support 8 colors but if those exist, we can probably find a better solution?

krobelus: I reported the missing RGB-support and missing XTVERSION support to Terminal.app (apparently this requires the Feedback Assistant app).

Also reported the Prompt bug about wrong DCS parsing and wrong CSI parsing on `'\x1b[?2026$'` (also reported that one to Terminal.app),

src/screen.rs
```diff
@@ -666,60 +665,16 @@ impl Screen {
...
-            let mut justgrey = true;
-            let add = |abandon_line_string: &mut WString, s: Option<CString>| {
-                let Some(s) = s else {
-                    return false;
-                };
-                abandon_line_string.push_utfstr(&str2wcstring(s.as_bytes()));
-                true
-            };
-            if let Some(enter_dim_mode) = term.and_then(|term| term.enter_dim_mode.as_ref()) {
-                if add(&mut abandon_line_string, Some(enter_dim_mode.clone())) {
```
krobelus, line 680: not to beat the dead horse too much, but it's pretty sad that our logic of

	if dim mode is supported, print dim mode,
	else if max_colors >= 238, print 237
	else ...

didn't work in practice because some terminals ignore dim even though
dim support is advertised (not sure if that's a bug in the terminfo
or a misunderstanding in our code)

doc_src/terminal-compatibility.rst
krobelus: actually we might want to document what kinds of key encodings we support (in addition to the kitty keyboard protocol).
Probably keeping the name I guess

krobelus:
I added a feature flag that can be used like `set -U fish_features no-ignore-terminfo`. I don't expect we'll need it, but just in case someone wants to quickly go back to the old behavior without patching and recompiling.

I also made `TERM=dumb` suppress all escape sequences for now,
since that helps our current pexpect tests.
I don't expect that we'll need this as a product feature;
in future I think we should print escape sequences even if `TERM=dumb`.

I guess there is an argument to be made for keeping
the behavior of using 24-bit RGB colors only if advertised via `COLORTERM`.
No strong opinion here but I guess it's fine to use them without checking.

faho:
> I added a feature flag that can be used like set -U fish_features no-ignore-terminfo. I don't expect we'll need it, but just in case someone wants to quickly go back to the old behavior without patching and recompiling.

Alright, so we're staging the switch. That's fine by me.

> in future I think we should print escape sequences even if TERM=dumb.

I have no idea what TERM=dumb means or is for. I've only really seen it as a misconfiguration or being overly conservative.

> I guess there is an argument to be made for keeping
the behavior of using 24-bit RGB colors only if advertised via COLORTERM.

My opinion is that RGB colors are *good* - we can't really make a good theme while going through an arbitrary palette.

But I recognize that some users prefer to have everything go through their palette, and $COLORTERM would be a good knob for that. It means these people can choose the palette lifestyle once and don't have to configure fish separately (if you want palette colors, you want palette colors everywhere).

I see $COLORTERM is set to "truecolor" on my system in alacritty, ghostty, konsole, gnome console, kitty, wezterm. It's not set in xterm and dropped by ssh.

That seems acceptable.

krobelus:
> Alright, so we're staging the switch. 

The merge will stop using terminfo by default (the `ignore-terminfo` feature flag will be on by default).
Unless we need it for colors..

> I have no idea what TERM=dumb means or is for.

Yeah I think I'll remove it. We use it only for pexpect tests (because if we always repaint the full command line, it's easier to write assertions), where we can call it `FISH_TEST_NO_COLORS_OR_CURSOR_MOVEMENT`.
FWIW we can avoid this if we implement a simple terminal on top of pexpect internals, for system tests that don't depend on tmux. 

> But I recognize that some users prefer to have everything go through their palette

I assume this only makes sense if their palette is 8-16 colors, because a 256 color palette would hardly prevent bad contrast.
If that's true, and assuming the only obvious way to communicate this is terminfo's max_colors, then we need to decide between two evils.
Either break users who explicitly set `TERM=xterm` or similar for a small palette, or don't unbreak users who have wrong/outdated terminfo.

Long-term I'd rather get away from terminfo (and tell people to use something like `fish_term256`).
Given that, I don't think there's a reason for a knob for color precision, see below.

---

> $COLORTERM would be a good knob for that

If we use it for both capability and configuration, we need to be careful because as mentioned, `COLORTERM` is sometimes not set when it should be (ssh), which is [fixable](https://github.com/termstandard/colors?tab=readme-ov-file#querying-the-terminal).
So I guess we can do
1. if `COLORTERM` is not defined, query the terminal directly (and turn RGB colors on if the terminal advertises support)
2. if `COLORTERM` is defined, use it to decide if we use RGB colors (except if inside GNU screen)

---

Personally I'm not sure why "translate RGB to palette" is still a desirable feature. I can see two potential reasons for having it
1. the terminal doesn't support RGB colors, so the downsampling is the best we can do
2. an RGB colorscheme has bad contrast

Reason 1 seems mostly obsolete, unless we think that this feature is important enough that we also want it to work on the handful of terminals that don't implement RGB colors (I don't think it is since default colors should use only palette colors). 

To reason 2, I feel like the concerns about bad contrast could be addressed easily in a less roundabout way -- if you want palette colors, then specify palette colors. Likewise for RGB colors.

Actually if we still care about reason 2, then we technically shouldn't even move from 16 color to the 256 color palette. Even if all  terminals support it, it may not be what the user wanted and is used to.
But I guess enough terminals use `TERM=xterm-256color` instead of `TERM=xterm` so a lot of users will be used to the 256 palette already.

The root of the problem is that we/terminfo is mixing capabilities with user configuration, which adds resistance to improvements. Again, this is why I was wary of using `COLORTERM` for configuration too, but I guess it might be fine.

---

Maybe we should even get rid of the RGB-to-palette translation.
Then we also no longer care about detecting support for RGB. We simply do what the user tells us to.
I don't really use colors myself, so I'm not sure.

Short of that, it seems very unlikely to me that anyone who has been using the RGB-to-256-translation color palette (which should be most users) would notice a significantly worse contrast when we instead write exact RGB sequences. There's plenty of colors in 256 that have bad contrast, so it would likely have been a problem before.

faho:
>  Unless we need it for colors..

We *do not* need it for colors. It isn't useful for colors.

We have assumed that TERM=xterm does 256 colors for years and nobody has ever complained.

The only other active terminfo entry I know of that advertises less than 256 colors is "linux", and that one can even do RGB now.

> Long-term I'd rather get away from terminfo (and tell people to use something like fish_term256).

The point of using $COLORTERM for this is that it is a global knob for *all* applications. If you don't want RGB color you don't want RGB color anywhere. You want all colors to follow your palette.

But that isn't terminfo, $TERM is in fact useless.

> I assume this only makes sense if their palette is 8-16 colors, because a 256 color palette would hardly prevent bad contrast.

I guess that's their problem to deal with. Like I said, we assume 256 colors for "xterm", so most people will be using a 256 color palette.

> Maybe we should even get rid of the RGB-to-palette translation.

We should not, because that means you cannot set a global knob to remove all RGB color.

You would have to set all color variables to not-RGB, and you would have to get every other use of set_color, including in scripts and functions you get from other people, to use not-RGB.

krobelus:
ok. If `COLORTERM` is missing from the environment, can we either assume RGB is supported, or query for support?
I want to avoid needing to fiddle with `COLORTERM` to make things work in ssh sessions.
To use it to turn off truecolor one can still set `COLORTERM=` or `COLORTERM=no`.
I dont' think this use case is standardized yet.

PR #11352, opened 2025-04-03
Fix profiling time log

danielrainer:
Fix profiling time log

`Duration::as_nanos()`
"Returns the total number of nanoseconds contained by this Duration."
https://doc.rust-lang.org/core/time/struct.Duration.html#method.as_nanos
    
Because we are indicating the time in milliseconds, we cannot use
`Duration::subsec_nanos()` but we need to manually get the whole duration in
nanoseconds and subtract the duration in milliseconds to get the fractional
part.


danielrainer:
I also managed to get confused myself by this. Now I think it should be correct, but make sure to double check.

PR #11372, opened 2025-04-11
Try to enable 24-bit RGB colors by default

krobelus:
`COLORTERM=truecolor` doesn't work in some cases; I think we may want to
either default to assuming truecolor is supported, or use this [recommended
fix](https://github.com/termstandard/colors?tab=readme-ov-file#querying-the-terminal).

See the discussion around
https://github.com/fish-shell/fish-shell/pull/11345#issuecomment-2794920900

TODO: docs
- [ ] `fish_term24bit`
- [ ] `doc_src/cmds/set_color.rst`
- [ ] `doc_src/terminal-compatibility.rst`

### DIFF
--- a/src/env_dispatch.rs
+++ b/src/env_dispatch.rs
@@ -53,6 +53,7 @@ static VAR_DISPATCH_TABLE: once_cell::sync::Lazy<VarDispatchTable> =
         }
 
         table.add(L!("TZ"), handle_tz_change);
+        table.add_anon(L!("COLORTERM"), handle_fish_term_change);
         table.add_anon(L!("fish_term256"), handle_fish_term_change);
         table.add_anon(L!("fish_term24bit"), handle_fish_term_change);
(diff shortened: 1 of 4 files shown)


faho:
So this does querying, which I think is overkill for this and I'm not a fan of.

I think we can just assume no COLORTERM is a misconfiguration (or ssh's mistake) and assume RGB-color is supported.

src/env_dispatch.rs
```diff
@@ -408,19 +409,9 @@ fn update_fish_color_support(vars: &EnvStack) {
                 "disabled"
             }
         );
-    } else if vars.get(L!("STY")).is_some() || term.starts_with(L!("eterm")) {
```
faho, line 410: Screen won't unset $COLORTERM, so STY should stay.

faho (approved):

PR #11381, opened 2025-04-13
completions/git: show url as description for remote completion

kpbaks:
## Description

Before:

image

After:

image

## TODOs:

- ~[ ] Changes to fish usage are reflected in user documentation/manpages.~
- ~[ ] Tests have been added for regressions fixed~

### DIFF
--- a/share/completions/git.fish
+++ b/share/completions/git.fish
@@ -96,7 +96,9 @@ function __fish_git_refs
 end
 
 function __fish_git_remotes
-    __fish_git remote 2>/dev/null
+    # Example of output parsed:
+    # "remote.upstream.url git@github.com:fish-shell/fish-shell.git" -> "upstream\tgit@github.com:fish-shell/fish-shell.git"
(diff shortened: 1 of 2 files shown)


share/completions/git.fish
```diff
@@ -96,7 +96,10 @@ function __fish_git_refs
 end
 
 function __fish_git_remotes
-    __fish_git remote 2>/dev/null
+    __fish_git config --get-regexp 'remote\.[a-z]+\.url' | while read key url
```
faho, line 99: This can be

```fish
__fish_git config --get-regexp 'remote\.[a-z]+\.url' | string replace -rf 'remote\.(.*)\.url (.*)' '$1\t$2'
```

Which anchors the remote name on the "remote." and the "url. " - knowing git, remote names can probably be anything (so technically even this isn't unambiguous, but it's quite unlikely to be a problem).

kpbaks: I have applied your feedback now

PR #11388, opened 2025-04-15
Curly and colored underlines

### LINKED ISSUE #7619: Support for coloured underlines
### System information:
fish, version 3.1.2
Running on Arch

### Question:
With increasing support in terminals (Kitty, VTE, mintty, and iTerm2, and Alacritty/Alacritty#4660) for ANSI SGR parameter 58, underline colour. It would be awesome if fish would support it as well.

Personally was thinking that maybe it could be an optional parameter on the `set_color` `-u` flag so that it would be backwards compatible.

### Example
`set_color -u red`

#### Edits:
2021-03-09: SGR parameter 53 should be 58 of course. Thanks @plgruener

krobelus:
Consider reviewing the last three commits.

Add curly and colored underlines, and use curly underlines for errors.

Add a new `--underline-color`` option to set_color (instead of adding an optional
color argument to --underline); this allows to set the underline color
independently of underline style (line, curly, etc.). I don't think this
flexibility is very important but this approach is probably the least hacky.

Closes #7619

### DIFF
--- a/doc_src/cmds/set_color.rst
+++ b/doc_src/cmds/set_color.rst
@@ -37,6 +37,9 @@ The following options are available:
 **-b** or **--background** *COLOR*
     Sets the background color.
 
+**--underline-color** *COLOR*
+    Set the underline color.
+
 **-c** or **--print-colors**
     Prints the given colors or a colored list of the 16 named colors.
 
--- a/doc_src/cmds/set_color.rst
+++ b/doc_src/cmds/set_color.rst
@@ -52,8 +52,8 @@ The following options are available:
 **-r** or **--reverse**
     Sets reverse mode.
 
-**-u** or **--underline**
-    Sets underlined mode.
+**-u** or **--underline** or **--curly-underline**
+    Set the underline mode.
 
 **-h** or **--help**
     Displays help about using this command.
--- a/share/functions/__fish_config_interactive.fish
+++ b/share/functions/__fish_config_interactive.fish
@@ -40,7 +40,7 @@ end" >$__fish_config_dir/config.fish
         __init_uvar fish_color_param cyan
         __init_uvar fish_color_redirection cyan --bold
         __init_uvar fish_color_comment red
-        __init_uvar fish_color_error brred
+        __init_uvar fish_color_error brred --curly-underline
         __init_uvar fish_color_escape brcyan
         __init_uvar fish_color_operator brcyan
         __init_uvar fish_color_end green
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -65,6 +65,7 @@ Completions
 
 Improved terminal support
 ^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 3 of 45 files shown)


doc_src/cmds/set_color.rst
```diff
@@ -52,8 +55,8 @@ The following options are available:
 **-r** or **--reverse**
     Sets reverse mode.
 
-**-u** or **--underline**
-    Sets underlined mode.
+**-u** or **--underline** or **--curly-underline**
+    Set the underline mode.
```
krobelus, line 59: actually it seems better to give it an optional arg.
Then it's more obvious that we can't combine two underline shapes.

	-u --underline, --underline=curly, --underline=double
	-u[STYLE] --underline[=STYLE] where STYLE is `single` (default), `curly` or `double`

could also call it undercurl but that doesn't work with this approach

share/functions/__fish_config_interactive.fish
```diff
@@ -40,7 +40,7 @@ end" >$__fish_config_dir/config.fish
         __init_uvar fish_color_param cyan
         __init_uvar fish_color_redirection cyan --bold
         __init_uvar fish_color_comment red
-        __init_uvar fish_color_error brred
+        __init_uvar fish_color_error brred --curly-underline
```
faho, line 43: What happens if a terminal doesn't support curly underlines?

Do they show underlines or not?

If it's the former I wouldn't want this in the default theme.

faho:
This has become one of those PRs where I would like to see bits merged separately.

A bunch of the cleanup here seems fine at a glance, but doesn't have anything to do with adding underline shape and color.

share/functions/__fish_config_interactive.fish
faho: Okay, ms-terminal colors the text and underline brred and blinks.

So that's no good, especially not for a default.

krobelus: There are some terminals that misinterpret these sequences (will post details later).

The most blocking one is probably that Terminal.app interprets curly underlines as yellow background,
so I won't make it an unconditional default anytime soon (because a `$TERM_PROGRAM` workaround is too unreliable).

As for deciding whether to make it a default that's conditional on the terminal explicitly advertising support via `Smulx` or `Su`,

> Do they show underlines or not?

is the relevant question indeed.
Normal terminals do not[^1] but nestable terminals generally do.
This includes tmux, GNU screen, Vim's `:terminal`, emacs ansi-term and `vterm`.
We could probably change tmux, see https://github.com/orgs/tmux/discussions/4477,
which is certainly the most relevant one. Hopefully at least vterm too.
Then we could think about making it the default after a few years.

Alternatively, we could add a special case to ignore `Smulx`/`Su` if they come from a multiplexing terminal
(AFAICT that's tmux, GNU screen, the emacs ones and dvtm/abduco. and perhaps Vim in clientserver mode)

Anyway, leaving this for later.

[^1]: `terminology` is an exception but that can be fixed easily

PR #11390, opened 2025-04-16
status get/list-files: fix minor bugs when compiled without embedded files

ilyagr:
## TODOs:

### DIFF
--- a/src/builtins/status.rs
+++ b/src/builtins/status.rs
@@ -474,7 +474,7 @@ pub fn status(parser: &Parser, streams: &mut IoStreams, args: &mut [&wstr]) -> B
             }
             #[cfg(not(feature = "embed-data"))]
             {
-                streams.err.append(wgettext_fmt!(
+                streams.err.appendln(wgettext_fmt!(
                     "%ls: fish was not built with embedded files",
                     cmd,
                 ));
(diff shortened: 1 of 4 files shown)


src/builtins/status.rs
```diff
@@ -475,7 +475,7 @@ pub fn status(parser: &Parser, streams: &mut IoStreams, args: &mut [&wstr]) -> B
             #[cfg(not(feature = "embed-data"))]
             {
                 streams.err.append(wgettext_fmt!(
```
faho, line 477: This should be appendln instead of adding the newline - which has the advantage of not leaving the newline in the translated string.

ilyagr: Sounds good, will fix in a moment. FYI, this is not done consistently currently, see e.g.

https://github.com/fish-shell/fish-shell/blob/471fa78bdc383cae6c5c838c2c763a337ef850e1/src/builtins/status.rs#L242

https://github.com/fish-shell/fish-shell/blob/471fa78bdc383cae6c5c838c2c763a337ef850e1/src/builtins/status.rs#L233

I don't want to fix it myself, since I don't know whether those strings were already translated or not.

ilyagr: No problem, and makes sense. Fixed.

faho:
Thanks!

ilyagr:
Thank you!

PR #11394, opened 2025-04-16
Update completion for tmutil

samdoran:
## Description

- Reorganize completions and options so they are easier to read.
- Add destination UUIDs to completetions as well as descriptions
- Add a few missing sub commands

The biggest change is adding the destination UUIDs to the completion. This is very helpful with `tmutil startbackup -d [destination ID]` or `tmutil setquota [destination ID]`.

I used the neat feature of `fish` to dynamically populate the completion description as well as the value.

One thing I could use help with is how to make the completion values update dynamically when destinations are added/removed. Currently I have to source the completions again to update the destination IDs. I know there is a way to do this but I couldn't quite figure it out. Thanks!

Fixes issue #

## TODOs:

### DIFF
--- a/share/completions/tmutil.fish
+++ b/share/completions/tmutil.fish
@@ -5,7 +5,6 @@ function __destination_ids
         if string match -q '*===*' -- $line
             # New section so clear out variables
             set -f name ''
-            set -f ID ''
             set -f kind ''
             continue
         end
--- a/share/completions/tmutil.fish
+++ b/share/completions/tmutil.fish
@@ -30,13 +30,12 @@ function __destination_ids
     end
 end
 
-complete -c tmutil -f
-complete -c tmutil -n '__fish_seen_subcommand_from status' -f
 complete -c tmutil -n __fish_use_subcommand -a status -d 'Display Time Machine status'
+complete -c tmutil -f -n '__fish_seen_subcommand_from status'
 
-complete -c tmutil -F -n __fish_use_subcommand -a addexclusion -d 'Add an exclusion not to back up a file'
-complete -c tmutil -F -n '__fish_seen_subcommand_from addexclusion' -s v -d 'Volume exclusion'
-complete -c tmutil -F -n '__fish_seen_subcommand_from addexclusion' -s p -d 'Path exclusion'
+complete -c tmutil -f -n __fish_use_subcommand -a addexclusion -d 'Add an exclusion not to back up a file'
+complete -c tmutil -r -n '__fish_seen_subcommand_from addexclusion' -s v -d 'Volume exclusion'
+complete -c tmutil -r -n '__fish_seen_subcommand_from addexclusion' -s p -d 'Path exclusion'
 
 complete -c tmutil -r -n __fish_use_subcommand -a associatedisk -d 'Bind a snapshot volume directory to the specified local disk'
 
@@ -62,7 +61,7 @@ complete -c tmutil -f -n '__fish_seen_subcommand_from compare' -s U -d 'Ignore l
 
 complete -c tmutil -r -n __fish_use_subcommand -a delete -d 'Delete one or more snapshots'
 complete -c tmutil -r -n '__fish_seen_subcommand_from delete' -s d -d 'Backup mount point'
-complete -c tmutil -r -n '__fish_seen_subcommand_from delete' -s t -d 'Timestamp'
+complete -c tmutil -r -f -n '__fish_seen_subcommand_from delete' -s t -d 'Timestamp'
 complete -c tmutil -r -n '__fish_seen_subcommand_from delete' -s p -d 'Path'
 
 complete -c tmutil -r -n __fish_use_subcommand -a deletelocalsnapshots -d 'Delete all local Time Machine snapshots for the specified date (formatted YYYY-MM-DD-HHMMSS)'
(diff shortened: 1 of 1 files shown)


share/completions/tmutil.fish
```diff
@@ -1,61 +1,136 @@
...
-complete -f -c tmutil -n __fish_use_subcommand -a stopbackup -d 'Cancel a backup currently in progress'
-complete -r -c tmutil -n __fish_use_subcommand -a thinlocalsnapshots -d 'Thin local Time Machine snapshots for the specified volume'
-complete -r -c tmutil -n __fish_use_subcommand -a uniquesize -d 'Analyze the specified path and determine its unique size'
-complete -r -c tmutil -n __fish_use_subcommand -a verifychecksums -d 'Verify snapshot'
-complete -f -c tmutil -n __fish_use_subcommand -a version -d 'Print version'
-
-complete -f -c tmutil -n '__fish_seen_subcommand_from destinationinfo isexcluded compare' -s X -d 'Print as XML'
+function __destination_ids
+    for line in $(tmutil destinationinfo)
+        if test -n "$(string match '*===*' $line)"
```
faho, line 5: `string match -q`, no `test`.

Also use `--` before the variable, or it'll break on lines starting with `-`.

So:

```fish
if string match -q '*===*' -- $line
```

Same goes for the other `string` uses.

share/completions/tmutil.fish
```diff
@@ -1,61 +1,136 @@
...
+            # New section so clear out variables
+            set -f name ''
+            set -f ID ''
```
faho, line 8: ID doesn't need to be a separate variable here because it's never used without being printed.

share/completions/tmutil.fish
```diff
@@ -1,61 +1,136 @@
...
+complete -c tmutil -r -n __fish_use_subcommand -a listlocalsnapshots -d 'List local Time Machine snapshots of the specified volume'
+
+complete -c tmutil -f -n __fish_use_subcommand -a listbackups -d 'Print paths for all snapshots'
+
+complete -c tmutil -f -n __fish_use_subcommand -a localsnapshot -d 'Create new local Time Machine snapshot of APFS volume in TM backup'
+
+complete -c tmutil -f -n __fish_use_subcommand -a machinedirectory -d 'Print the path to the current machine directory'
+
+complete -c tmutil -f -n __fish_use_subcommand -a removedestination -d 'Removes a backup destination'
+complete -c tmutil -f -n '__fish_seen_subcommand_from removedestination' -d "$(__destination_ids)"
```
faho, line 100: You want this to be an *argument*, not a description, presumably? I.e. `-a` instead of `-d`?

share/completions/tmutil.fish
```diff
@@ -1,61 +1,136 @@
...
+complete -c tmutil -f -n '__fish_seen_subcommand_from setdestination' -s a -d 'Add to the list of destinations'
+complete -c tmutil -f -n '__fish_seen_subcommand_from setdestination' -s p -d 'Enter the password at a non-echoing interactive prompt'
+
+complete -c tmutil -f -n __fish_use_subcommand -a snapshot -f -d 'Create new local Time Machine snapshot'
+
+complete -c tmutil -f -n __fish_use_subcommand -a startbackup -d 'Begin a backup if one is not already running'
+complete -c tmutil -f -n '__fish_seen_subcommand_from startbackup' -s a -l auto -d 'Automatic mode'
+complete -c tmutil -f -n '__fish_seen_subcommand_from startbackup' -s b -l block -d 'Block until finished'
+complete -c tmutil -f -n '__fish_seen_subcommand_from startbackup' -s r -l rotation -d 'Automatic rotation'
+complete -c tmutil -f -n '__fish_seen_subcommand_from startbackup' -s d -l destination -r -a "$(__destination_ids)" -d 'Backup destination'
```
faho, line 119: This:

```fish
"$(...)"
```

is a command substitution that will be expanded when this is sourced.

What you want is to use `-a`'s ability to do that when the argument is needed:

```fish
-a "(__destination_ids)"
```

(or `'$(__destination_ids)'` - the important part is that the command substitution is quoted so it doesn't run when the `complete` command is run, but when the `-a` is evaluated)

faho (changes requested):

share/completions/tmutil.fish
samdoran: Oh that's much nicer. Thanks!

share/completions/tmutil.fish
samdoran: Yes. I corrected one of those but I missed one. Thank you for catching it.

share/completions/tmutil.fish
samdoran: This was extra diligence to make sure a value for ID from a previous section didn't get reused, but I guess that's unnecessary.

share/completions/tmutil.fish
samdoran: That's brilliant! Essentially avoid expanding the command substitution within the quotes and it gets run later.

This works as expected in my testing.

samdoran:
I've got a few more subtle behaviors to iron out mainly around when to offer file completion for certain subcommands. I think the `-F` use is incorrect.

samdoran:
This should be good to merge now. Thanks for the very helpful review.

samdoran:
Thank you for merging and for the great review.

PR #11396, opened 2025-04-16
Use stdlib functions to show profiling info

danielrainer:
## Description

Refactor `print_profile` to use Rust's standard formatting and writing functionality.

## TODOs:
- [x] Ensure that UTF32 in `item.cmd` is converted correctly.
- [x] Ensure that scripts do not rely on exact formatting of profiling output.
- [x] Error handling: Do we want to do anything about errors if they occur during writing?
- [x] Style: `write!` could be used instead of `write_all` + `format!` + `as_bytes`, which would be a bit more concise, but does not retry when interrupted.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -3,8 +3,8 @@
 use crate::ast::{self, Ast, List, Node};
 use crate::builtins::shared::STATUS_ILLEGAL_CMD;
 use crate::common::{
-    escape_string, CancelChecker, EscapeFlags, EscapeStringStyle, FilenameRef, ScopeGuarding,
-    ScopedCell, ScopedRefCell, PROFILING_ACTIVE,
+    escape_string, wcs2string, CancelChecker, EscapeFlags, EscapeStringStyle, FilenameRef,
+    ScopeGuarding, ScopedCell, ScopedRefCell, PROFILING_ACTIVE,
 };
 use crate::complete::CompletionList;
 use crate::env::{EnvMode, EnvStack, EnvStackSetResult, Environment, Statuses};
@@ -38,8 +38,9 @@ use portable_atomic::AtomicU64;
 use std::cell::{Ref, RefCell, RefMut};
 use std::ffi::{CStr, OsStr};
 use std::fs::File;
+use std::io::Write;
 use std::num::NonZeroU32;
-use std::os::fd::{AsRawFd, OwnedFd};
+use std::os::fd::OwnedFd;
 use std::rc::Rc;
 #[cfg(target_has_atomic = "64")]
 use std::sync::atomic::AtomicU64;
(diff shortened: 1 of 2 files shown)


src/parser.rs
```diff
@@ -1252,13 +1260,14 @@ fn print_profile(items: &[ProfileItem], out: &mut File) {
...
-
-        fprintf!(out, "> %ls\n", item.cmd);
+        let level = item.level.unsigned_abs().saturating_add(1);
+        let _ = out.write_all(
+            format!(
+                "{:>col_width$} {:>col_width$} {:->level$} {}\n",
+                self_time, total_time, '>', item.cmd
+            )
+            .as_bytes(),
+        );
```
krobelus, line 1270: this (moving from `\t` to a fixed-with, right-aligned columns) looks like a good and long overdue change.

Also mentioning that this is microseconds is great, that would have helped me many times.
(I'd also be interested in converting it to milliseconds e.g. 12.345 instead of 12345).
To make sure that we don't break any existing scripts, how about adding the unit to the first row?
So `TIme (μs)   Sum (μs)` (or ms for milliseconds if we do that).
I'm fine with the extra row too, it's very unlikely that we're causing significant breakage if we add a header row.


> Error handling: Do we want to do anything about errors if they occur during writing?

it's probably fine to keep going on errors since we'll exit quickly either way

> Ensure that UTF32 in item.cmd is converted correctly.

[Yeah](https://github.com/fish-shell/fish-shell/pull/11397#discussion_r2050759693), we should use `out.write_all(&wcs2string(item.cmd))` here.

(As an aside, by the time a PR is ready to merge, it would be good to have the most important things like maybe a reason for the change etc. in the commit message, not just the PR description)

krobelus:
> but does not retry when interrupted.

yeah that's important for us (probably not in this specific case, but in general, and we better be consistent). We get things like SIGWINCH.
If we started from scratch, we'd try to use `SA_RESTART` as much as possible, to make this less fiddly

src/parser.rs
danielrainer: > To make sure that we don't break any existing scripts, how about adding the unit to the first row?
So TIme (μs)   Sum (μs) (or ms for milliseconds if we do that).

Sounds good, done.

> I'd also be interested in converting it to milliseconds e.g. 12.345 instead of 12345

I am not a huge fan of this idea. Not having integers makes the output a bit more annoying to parse, but the main issue I see is that it would negatively affect readability in many cases. For example, I just profiled running a command with the debug build, and the longest cumulative duration was 593 microseconds. Using milliseconds in the output would result in the entire column looking like `0.xxx` (in most cases `0.0xx`), which makes it harder to visually identify the items which take most time. Microseconds seems like a good unit for the purposes of this output.

> Yeah, we should use out.write_all(&wcs2string(item.cmd)) here.

Done.

Issue #11304, opened 2025-03-19
Getting off UTF-32


nc7s:
https://github.com/fish-shell/fish-shell/issues/10633 stated (though in a ⯈ `<detail>` under the pcre2 crate fork item):

> we expect to eventually get off of UTF-32

but it's postponed to give way to the 4.0.0 release. Now that _that_ is out, it's time to start pruning!

faho suggested on matrix to test the waters with the string builtins and wildcard, and make a plan.

Thus my first was the string builtins. The `wgetopt` module is converted to a UTF-8 `getopt` module, removing `wopt!` and `L!` usage in the string builtins. However, src/builtins/shared.rs, `IoStreams` and formatting macros mean this quickly escalates, as many parts are intertwined. (Not sure if this `getopt` module itself could/should be replaced with other, more "Rust idiomatic" libraries, but that's another topic.)

Took a quick look over the wildcard module. Seems to be string operations for the most part.

### Originators

Here are the wide char modules, the functionalities they provide, and their "eliminatability":

- `wchar`, could be eliminated
  - Re-export `widestring::{Utf32Str as wstr, Utf32String as WString}`, `widestring::utf32str!` as `L!`
- `wchar_ext`, could be eliminated
  - `trait ToWString`
- `wcstringutil`
  - Some common string operations for wide strings, could be replaced with std?
  - Fuzzy matching, could be replace with a crate?
- `wgetopt`, converted
  - Wide `getopt`
- `widecharwidth`, could be replaced with [unicode-width](https://docs.rs/unicode-width/)?
  - Tables and utility to calculate Unicode character width
- `wutil`
  - `mod.rs`
    - Wide string versions of some C functions, could be replaced by std or libc?
    - Some wide string only functions could be eliminated?
  - `dir_iter`, could be replaced with [walkdir](https://docs.rs/walkdir/)?
    - Wide string directory walking
  - `encoding`, could be eliminated?
    - Imports `wcrtomb()`, `mbrtowc()`, `struct mbstate_t` from C
  - `error`, could be eliminated?
    - `enum Error { Overflow, Empty, InvalidChar }`
  - `fileid`, seems should be kept
    - `struct FileId` and supporting utilities to uniquely identify a file
  - `gettext`, could be replaced with i18n/l10n crates?
    - Macros `wgettext!`, `wgettext_fmt!`, `wgettext_maybe_fmt!`
  - `hex_float`, lift?
    - Independent
[...]

PR #11397, opened 2025-04-17
Use standard library formatting and printing functions

danielrainer:
## Description

This replaces the `printf!` and `eprintf!` macros by `print!` and `eprint!`, respectively. The `fprintf!` macro can also be removed once #11396 is merged. This would then only leave the `sprintf!` macro, which is also used by some `wgettext` macros, making it more annoying to replace due to having to adjust the translations accordingly.

### DIFF
--- a/src/bin/fish.rs
+++ b/src/bin/fish.rs
@@ -38,10 +38,9 @@ use fish::{
         environment::{env_init, EnvStack, Environment},
         ConfigPaths, EnvMode, Statuses, CONFIG_PATHS,
     },
-    eprintf,
     event::{self, Event},
     flog::{self, activate_flog_categories_by_pattern, set_flog_file_fd, FLOG, FLOGF},
-    fprintf, function, future_feature_flags as features,
+    function, future_feature_flags as features,
     history::{self, start_private_mode},
     io::IoChain,
     nix::{getpid, getrusage, isatty, RUsage},
@@ -181,7 +180,7 @@ fn read_init(parser: &Parser, paths: &ConfigPaths) {
         let ret = parser.eval_file_wstr(src, fname, &IoChain::new(), None);
         parser.libdata_mut().within_fish_init = false;
         if let Err(msg) = ret {
-            eprintf!("%ls", msg);
+            eprint!("{msg}");
         }
     }
     #[cfg(not(feature = "embed-data"))]
@@ -238,7 +237,7 @@ fn run_command_list(parser: &Parser, cmds: &[OsString]) -> Result<(), libc::c_in
             retval = Ok(());
         } else {
             let backtrace = parser.get_backtrace(&cmd_wcs, &errors);
-            eprintf!("%s", backtrace);
+            eprint!("{backtrace}");
             // XXX: Why is this the return for "unknown command"?
             retval = Err(STATUS_CMD_UNKNOWN);
         }
--- a/src/bin/fish.rs
+++ b/src/bin/fish.rs
@@ -50,7 +50,6 @@ use fish::{
     parse_util::parse_util_detect_errors_in_ast,
     parser::{BlockType, CancelBehavior, Parser},
     path::path_get_config,
-    printf,
     proc::{
         get_login, is_interactive_session, mark_login, mark_no_exec, proc_init,
         set_interactive_session, Pid,
@@ -336,8 +335,8 @@ fn fish_parse_opt(args: &mut [WString], opts: &mut FishCmdOpts) -> ControlFlow<i
             }
             'P' => opts.enable_private_mode = true,
             'v' => {
-                printf!(
-                    "%s",
+                print!(
+                    "{}",
                     wgettext_fmt!("%s, version %s\n", PACKAGE_NAME, fish::BUILD_VERSION)
                 );
                 return ControlFlow::Break(0);
--- a/src/autoload.rs
+++ b/src/autoload.rs
@@ -189,7 +189,7 @@ impl Autoload {
                 widename.push_str(name);
                 let ret = parser.eval_file_wstr(src, Arc::new(widename), &IoChain::new(), None);
                 if let Err(msg) = ret {
-                    eprintf!("%ls", msg);
+                    eprint!("{msg}");
                 }
             }
         }
--- a/src/builtins/fg.rs
+++ b/src/builtins/fg.rs
@@ -126,7 +126,7 @@ pub fn fg(parser: &Parser, streams: &mut IoStreams, argv: &mut [&wstr]) -> Built
     } else {
         // If we aren't redirecting, send output to real stderr, since stuff in sb_err won't get
         // printed until the command finishes.
-        eprintf!("%s", wgettext_fmt!(FG_MSG, job.job_id(), job.command()));
+        eprint!("{}", wgettext_fmt!(FG_MSG, job.job_id(), job.command()));
     }
 
     let ft = tok_command(job.command());
--- a/src/io.rs
+++ b/src/io.rs
@@ -205,7 +205,7 @@ impl IoData for IoClose {
         -1
     }
     fn print(&self) {
-        eprintf!("close %d\n", self.fd)
+        eprintln!("close {}", self.fd)
     }
 }
 
@@ -231,7 +231,7 @@ impl IoData for IoFd {
         self.source_fd
     }
     fn print(&self) {
-        eprintf!("FD map %d -> %d\n", self.source_fd, self.fd)
+        eprintln!("FD map {} -> {}", self.source_fd, self.fd)
     }
 }
 
@@ -260,7 +260,7 @@ impl IoData for IoFile {
         self.file.as_raw_fd()
     }
     fn print(&self) {
-        eprintf!("file %d -> %d\n", self.file.as_raw_fd(), self.fd)
+        eprintln!("file {} -> {}", self.file.as_raw_fd(), self.fd)
     }
 }
 
@@ -292,8 +292,8 @@ impl IoData for IoPipe {
         self.pipe_fd.as_raw_fd()
     }
     fn print(&self) {
-        eprintf!(
-            "pipe {%d} (input: %s) -> %d\n",
+        eprintln!(
+            "pipe {{{}}} (input: {}) -> {}",
             self.source_fd(),
             if self.is_input { "yes" } else { "no" },
             self.fd
@@ -376,11 +376,7 @@ impl IoData for IoBufferfill {
         self.write_fd.as_raw_fd()
     }
     fn print(&self) {
-        eprintf!(
-            "bufferfill %d -> %d\n",
-            self.write_fd.as_raw_fd(),
-            self.fd()
-        )
+        eprintln!("bufferfill {} -> {}", self.write_fd.as_raw_fd(), self.fd())
     }
     fn as_bufferfill(&self) -> Option<&IoBufferfill> {
         Some(self)
@@ -623,20 +619,17 @@ impl IoChain {
     /// Output debugging information to stderr.
     pub fn print(&self) {
         if self.0.is_empty() {
-            eprintf!(
-                "Empty chain %s\n",
-                format!("{:p}", std::ptr::addr_of!(self))
-            );
+            eprintln!("Empty chain {:p}", std::ptr::addr_of!(self));
             return;
         }
 
-        eprintf!(
-            "Chain %s (%ld items):\n",
-            format!("{:p}", std::ptr::addr_of!(self)),
+        eprintln!(
+            "Chain {:p} ({} items):",
+            std::ptr::addr_of!(self),
             self.0.len()
         );
         for (i, io) in self.0.iter().enumerate() {
-            eprintf!("\t%lu: fd:%d, ", i, io.fd());
+            eprint!("\t{}: fd:{}, ", i, io.fd());
             io.print();
         }
     }
--- a/src/panic.rs
+++ b/src/panic.rs
@@ -34,21 +34,21 @@ pub fn panic_handler(main: impl FnOnce() -> i32 + UnwindSafe) -> ! {
             if let Some(at_exit) = AT_EXIT.get() {
                 (at_exit)(false);
             }
-            eprintf!(
-                "%s crashed, please report a bug.",
-                PROGRAM_NAME.get().unwrap(),
+            eprint!(
+                "{} crashed, please report a bug.",
+                PROGRAM_NAME.get().unwrap()
             );
             if !is_main_thread() {
-                eprintf!("\n");
+                eprintln!();
                 std::thread::sleep(Duration::from_secs(1));
             } else {
-                eprintf!(" Debug PID %d or press Enter to exit\n", unsafe {
+                eprintln!(" Debug PID {} or press Enter to exit", unsafe {
                     libc::getpid()
                 });
                 let mut buf = [0_u8; 1];
                 while let Ok(n) = read_blocked(STDIN_FILENO, &mut buf) {
                     if n == 0 || matches!(buf[0], b'q' | b'\n' | b'\r') {
-                        eprintf!("\n");
+                        eprintln!();
                         break;
                     }
                 }
--- a/src/parse_execution.rs
+++ b/src/parse_execution.rs
@@ -257,7 +257,7 @@ impl<'a> ExecutionContext<'a> {
 
             // Print it.
             if !should_suppress_stderr_for_tests() {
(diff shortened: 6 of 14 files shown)


src/bin/fish.rs
```diff
@@ -181,7 +179,7 @@ fn read_init(parser: &Parser, paths: &ConfigPaths) {
         let ret = parser.eval_file_wstr(src, fname, &IoChain::new(), None);
         parser.libdata_mut().within_fish_init = false;
         if let Err(msg) = ret {
-            eprintf!("%ls", msg);
+            eprint!("{msg}");
```
krobelus, line 182: Looks reasonable -- a step in the right direction.

Related: https://github.com/fish-shell/fish-shell/pull/9948

In future we should figure out what's needed so we can write
Rust format strings and still use them with gettext.

src/bin/fish.rs
```diff
@@ -238,7 +236,7 @@
             retval = Ok(());
         } else {
             let backtrace = parser.get_backtrace(&cmd_wcs, &errors);
-            eprintf!("%s", backtrace);
+            eprint!("{backtrace}");
```
krobelus, line 239: note that `print!` locks stdout but `printf!` macros don't.
I don't think we need locking anywhere since we don't use threads very much but I guess it's harmless.

src/bin/fish.rs
danielrainer: I had a look and it seems that the recommendation for using `gettext` with Rust is using the `gettext-rs` crate and `formatx` for formatting (which is similar to to stdlib `format!`, but has some restrictions and differences)
https://www.gnu.org/software/gettext/manual/html_node/Rust.html

To get a sense of the scope of refactoring this, I ran
```fish
rg '^msgid.*%.*$' messages.pot | wc -l
```
which reports 328 lines containing at least one `%`. A few of them are not format strings, the vast majority just use the default formatting for the data type, and a few have some extra formatting, which could be reproduced using `formatx!`. One format string (`"Unicode character out of range: \\%c%0*x"`) might need special handling.

The `gettext-rs` crate only handles UTF-8, so some extra conversions might be needed while UTF-32 is still in use.

Before working on this, it might make sense to clean up the current translation code, see #11400.

src/bin/fish.rs
```diff
@@ -238,7 +236,7 @@ fn run_command_list(parser: &Parser, cmds: &[OsString]) -> Result<(), libc::c_in
             retval = Ok(());
         } else {
             let backtrace = parser.get_backtrace(&cmd_wcs, &errors);
-            eprintf!("%s", backtrace);
```
krobelus, line 241: Hmm actually I think there are two issues.

TL;DR: I think we should try to switch to Rust's format string syntax
but otherwise keep the current behavior for now.

There are quite some uses of `e?print(ln)?!` already; we should at least get
rid of the ones in non-test code.

src/bin/fish.rs
krobelus, line 239: Issue 1: The default formatters bypass fish's internal encodings schema.

E.g. given

```
let backtrace = WString::from(...);
eprint!("{backtrace}");
```

Even though WString is from a third party crate, fish uses the Unicode
private use area (PUA) to encode sequences of bytes that either

1. cannot be decoded at all in the user's locale
   For a test case see 02842923927 (builtin read to pass through invalid UTF-8; reader to ignore invalid codepoints, 2025-04-15)
2. decode to a number that's not a valid Unicode codepoint
   For a test case see d9ba27f58fe (builtin read: always handle out-of-range codepoints (Rust port regression), 2025-04-15)

This is somewhat un-idiomatic (would be less surprising if it was our own type);
but this means that for now all text input and output must go through str2wcstring (or the equivalent code-clones that read one byte at a time) and wcs2string respectively.
Otherwise we'll potentially output a PUA character when we should output the same bytes we read.

Can we do the format string syntax change (ideally also for translations) while keeping the current IO behavior.


---

Optionally, maybe even redefine/shadow the standard macros so that

	eprint!("{backtrace}");

internally translates to

	std::eprint!("{wrapper(backtrace)}");

where (very roughly):

```rust
fn wrapper(fish_str: &wstr) -> Wrapper {
    todo!()
}
struct Wrapper(&wstr);
impl std::fmt::Write for Wrapper {
    fn write(&self, fmt) {
	fmt.write(wcstr2string(self.0));
    }
}
```

I'm not sure if this is very easy to do. Maybe it's not worth the effort.

We can keep using different names.
But it would be good to add a clippy lint to tell us not to use the standard ones.

src/bin/fish.rs
```diff
@@ -337,8 +335,8 @@
             }
             'P' => opts.enable_private_mode = true,
             'v' => {
-                printf!(
-                    "%s",
+                print!(
+                    "{}",
```
krobelus, line 339: Issue 2: `println!()` will panic when stdout is closed (since write fails with `EPIPE`).
(we ignore SIGPIPE globally but that doesn't help us here).

Reproduce this by applying the diff below, then run

	cargo b && target/debug/fish -c '' | cat

and hit `ctrl-c`. This will crash fish next time it tries to
`println!()`.  I do not know if this is a real problem, but it's
enough reason to separate the format string syntax change from a
possible switch to the standard print routines.

```diff
diff --git a/src/bin/fish.rs b/src/bin/fish.rs
index a078a45dc9..2effe399f8 100644
--- a/src/bin/fish.rs
+++ b/src/bin/fish.rs
@@ -566,6 +566,10 @@
     signal_clear_cancel();
 
     if !opts.batch_cmds.is_empty() {
+        for _i in 0..10 {
+            println!("HELLO");
+            std::thread::sleep(std::time::Duration::from_secs(1));
+        }
         // Run the commands specified as arguments, if any.
         if get_login() {
             // Do something nasty to support OpenSUSE assuming we're bash. This may modify cmds.
```

src/bin/fish.rs
krobelus: formatx sounds perfect!

> The gettext-rs crate only handles UTF-8, so some extra conversions might be needed while UTF-32 is still in use.

I suspect those conversions are already taking place today

danielrainer: Ok, apparently the build does work, it is just really slow.

Possibly relevant: https://github.com/gettext-rs/gettext-rs/issues/138

krobelus: does moving to `formatx` imply we absolutely need gettext-rs?
Today we link against the system's gettext library directly. It seems reasonable to get rid of it eventually but maybe there's no hurry.
If we want to use it, it would be good to fix this build slowness first.
And perhaps we want to continue to offer the option to use the system's gettext library instead of compiling it ourselves, I think that some distros will want that option.

> The gettext-rs crate only handles UTF-8

Note that we want `gettext("some path: %s", path)` to keep working even if path contains non-UTF-8 bytes.
I think the gettext C function does support this

danielrainer: The crate has a feature which allows using the system installation of gettext instead of compiling from source. With this feature enabled, build times are not an issue on systems with gettext present. See https://github.com/fish-shell/fish-shell/issues/11400#issuecomment-2816336533

src/bin/fish.rs
danielrainer: I found lints which might be helpful: #11402

src/bin/fish.rs
danielrainer: Some further tests have shown that `gettext-rs` and `formatx` do work together.

When wrapping the `gettext` function in a macro the strings passed to the macro unfortunately do not get extracted into the po file via `xtr`/`xgettext`. Using macros is desirable for combining translation and formatting, even without the encoding conversions. Until this is fixed, it seems that `gettext-rs` is not ready to be used in fish. Possibly this could be done using our own implementation on top, but I have not looked into how the extraction is actually done, so I do not know how difficult it would be to add macro support.

danielrainer: I think we could replace the `printf`-style formatting with `formatx!`. AFAICT this would require converting all the `%` format specifiers in the source and po files to the curly brace alternatives. In addition, we should ensure that existing curly braces are appropriately escaped. This requires checking whether a string is used as a format string or not, for each string containing curly braces. (I guess the same applies to `%` at the moment.)

`formatx!` seems to ignore extra arguments (arguments which do not have a corresponding placeholder in the format string) and returns an error when there is no argument for a placeholder. This means that we have no way of noticing missing placeholders in translations.

For missing arguments, we would have to decide how to handle them:
- Try the string from the source code as a fallback?
- Print some error message?
- panic? (either immediately or if the fallback also fails)

src/bin/fish.rs
danielrainer: I've spent some time investigating this, and I don't see a way of making this work using formatting via the normal `Display` implementation while still supporting arbitrary bytes in the output.

The `wcs2string` function returns `Vec<u8>`, but the formatter works on Rust strings/chars (https://doc.rust-lang.org/std/fmt/trait.Write.html). So to make this work we either have to reimplement formatting for everything (similar to what is currently done in `printf`-style), or we have to decode `WString` into `String`, using replacement characters or dropping invalid bytes. Neither seems like a good option.

One question I now have is why is the `WString`/`&wstr` type used for data which can consist of arbitrary bytes? This seems to invite unintentional misuse. And more generally, why is it used at all? Is there any situation where `WString` is preferable to all of `String`, `OsString`, `CString`, and `Vec<u8>`? There is discussion about this in  #11304, but I don't know if there is any consensus on how to move away from `WString`.

danielrainer:
I'll close this, since `std` formatting does not work when we want non-UTF-8 output as discussed above.

For now, the only way we have found to resolve this issue is writing our own formatting code, like the current printf-style code. While a more rusty formatting implementation could replace internal use of the printf macros, we still need the printf code for the `printf` built-in command.

src/bin/fish.rs
danielrainer: The historical usage influencing the choices for the Rust port makes sense.

> There is no way to enforce all that to conform to utf-8. It's just not how unix works.

Yes, of course. The question is how do we want to represent such data. We can of course switch to UTF-8 `String`, but that only addresses some space and conversion overhead, not the issue of having to use the PUA encoding hack to represent arbitrary data, which is extra overhead and can easily result in conversions being forgotten.

The proposal to use bstr in #11415 seems to be an alternative, but I have not yet looked into the crate's capabilities much. Do you have opinions on what would be best?

src/bin/fish.rs
bhaible: > Some further tests have shown that `gettext-rs` and `formatx` do work together.

Sure. There is even a full sample showing how these play together: https://gitweb.git.savannah.gnu.org/gitweb/?p=gettext.git;a=tree;f=gettext-tools/examples/hello-rust

> When wrapping the `gettext` function in a macro the strings passed to the macro unfortunately do not get extracted into the po file via `xtr`/`xgettext`.

With `xgettext` you can get the desired result, simply by passing a command-line option. If your macro name is, say, `foo!`, and the internationalized string is the first argument, you need to pass `-k foo!:1` to `xgettext`.

danielrainer: Thank you for your interest! We found that approaches which use Rust's normal formatting are not particularly suitable for fish, since the interface is built assuming valid UTF-8 strings, whereas we want to be able to work with arbitrary byte sequences, so for now we keep using fish's custom `printf` formatting, which is very similar to `printf` in C.

Regarding message extraction, we decided to use a custom approach based on Rust macros, which allows us to have a type-safe way of constructing localizable strings and more concise ways of declaring message constants. It also simplifies working with widestrings. (Fish uses a 32-bit widestring type internally, which adds some annoyances such as needing an extra macro for declaring widestring constants. We want to move away from that approach eventually, but doing so requires significant effort.) An additional benefit of our approach is that it works on systems which do not have a recent, Rust-aware version of `xgettext`, which as far as I am aware has only added Rust support earlier this year. Because our macros are evaluated by the compiler, we also trivially avoid false positives an negatives. If you are interested, you can check `src/wutil/gettext.rs`, `crates/gettext-extraction/src/lib.rs`, and `build_tools/fish_xgettext.fish` for the implementation.

We are also working on a custom runtime implementation, primarily discussed in https://github.com/fish-shell/fish-shell/pull/11726. It simplifies building since we will no longer have to link to gettext, which allows us to get rid of a bunch of build-tooling we don't need for anything else, as well as support localization on systems without the gettext library installed. A custom implementation also allows us to build the maps into the executable, either by bundling MO file data and parsing it at runtime, or by directly building lookup maps into the executable, so no parsing needs to happen at runtime. The main advantage of such embedding is that localization works regardless of the state of the filesystem, which is particularly useful for executables installed via `cargo install`, which does not support installing data files alongside the executable.

One issue which has been discussed in [this thread](https://github.com/fish-shell/fish-shell/pull/11726#discussion_r2297374957) is the exact way of handling language fallback. It would be interesting to get your perspective on that.

krobelus: > If your macro name is, say, foo!, and the internationalized string is the
> first argument, you need to pass -k foo!:1 to xgettext.

wow that's good to know

> We found that approaches which use Rust's normal formatting are not
> particularly suitable for fish, since the interface is built assuming
> valid UTF-8 strings, whereas we want to be able to work with arbitrary
> byte sequences, so for now we keep using fish's custom printf formatting,
> which is very similar to printf in C.

FWIW we definitely want to change the syntax from `%s` to `{}` but yeah, we
still want `sprintf!("file not found: {}", path)` to work even when `path`
is not valid UTF-8.  (We encode invalid bytes using the private-use-area,
and convert them back on output.)

> Regarding message extraction, we decided to use a custom approach based
> on Rust macros, which allows us to have a type-safe way of constructing
> localizable strings and more concise ways of declaring message constants. It
> also simplifies working with widestrings.

Yeah our approach seems solid and I don't think we have a reason to switch,
but if we do, we could probably switch to xgettext as extraction tool,
I don't expect any significant problems with that?
Widestrings should be handled transparently because `wgettext!` takes a
narrow string argument, which is palatable to xgettext.

> An additional benefit of our approach is that it works on systems which do
> not have a recent, Rust-aware version of xgettext, which as far as I am
> aware has only added Rust support earlier this year.

I see; but since it's a dev-dependency we would be fine with requiring the
latest version.

> The main advantage of such embedding is that localization works regardless
> of the state of the filesystem, which is particularly useful for executables
> installed via cargo install, which does not support installing data files
> alongside the executable.

maybe `gettext-rs` should allow the same, i.e. be able to source virtual
*.mo file contents from a "virtual file system" provided by the app.

danielrainer: > wow that's good to know

Yes, I knew about the `-kmacro_name!` format, but I didn't know about the possibility of adding numbers either. It's not in the man page of xgettext I have installed, which is probably why I missed it. The info docs talk about the feature.

> I don't expect any significant problems with that?

We did use xgettext before switching to the custom implementation and it mostly worked, but with some annoying edge cases. It also ran on macro-expanded sources, instead of the real source code, which might have been due to the integration predating Rust support in xgettext, but possibly also other issues. Some features we use now, like the `localizable_consts!` macro which allows specifying an arbitrary number of localizable constants with associated English string literals, can probably not work with xgettext.

The main benefits of using xgettext over our implementation (assuming xgettext can handle everything correctly) would be somewhat reduced implementation complexity (we could get rid of our proc-macro and instead maintain a list of macro names as arguments to xgettext in our script for running extraction) and not needing a Rust toolchain for extracting message strings. The former might be worth considering, although our implementation is actually fairly small and not that complex, just weird and potentially unexpected for people used to normal xgettext. The latter seems mainly relevant for contributors who don't work with Rust code. There it depends how we want to handle PO file updates in PRs going forward.

Overall, I agree we don't have a convincing enough case to change our implementation at the moment.

> maybe `gettext-rs` should allow the same, i.e. be able to source virtual *.mo file contents from a "virtual file system" provided by the app.

`gettext-rs` wraps the GNU gettext library, so for that to work there would probably have to be support for it in GNU's implementation. We basically have our own version of `gettext-rs` already in the code base, minus the part which builds and statically links gettext if it's not available as a shared lib at build time. The latter part is implemented quite poorly, as it `make`s the entire gettext tarball if the shared lib is not available, which takes so long that I initially thought the `gettext-sys` crate (dependency of `gettext-rs`) is broken (See earlier in this thread). I then found https://github.com/gettext-rs/gettext-rs/issues/138, which was reported by bhaible (with a proposed fix) in February and has not been addressed since, aside from a comment (by someone who AFIACT is not a contributer to the project), which highlights how the dynamic lib can be used instead. `gettext-rs` does not seem to be maintained well enough that I'd be comfortable with having it as a dependency. The latest commit to master happened 9 months ago.

PR #11399, opened 2025-04-18
Fix translation file generation

danielrainer:
The previous version generates files which do not preserve the line number from the original fish script file, resulting in translation not working.

The new approach is quite ugly, and might have some issues, but at least it seems to work in some cases

For extracting gettext strings from rust code, I also added quotes around the string such that grep only looks for the string when it appears in quotes. This prevents scenarios where the string content also appears outside of a string unquoted from producing garbage.

### DIFF
--- a/build_tools/fish_xgettext.fish
+++ b/build_tools/fish_xgettext.fish
@@ -6,7 +6,7 @@
 # works around that - based on share/functions/funced.fish.
 set -q TMPDIR
 or set -l TMPDIR /tmp
-set -l tmpdir (mktemp -d $TMPDIR/fish.XXXXXX)
+set tmpdir (mktemp -d $TMPDIR/fish.XXXXXX)
 or exit 1
 
 # This is a gigantic crime.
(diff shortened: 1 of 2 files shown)


danielrainer:
It seems that the output generated with this code might be sensible now, but it results in diffs of over 100k lines per language, so there might still be some issues hiding in there.

build_tools/fish_xgettext.fish
```diff
@@ -50,14 +50,29 @@ mkdir -p $tmpdir/explicit/share/completions $tmpdir/explicit/share/functions
...
+    # Then, replace all lines without this prefix by empty lines.
+    # These lines are kept to ensure that the correct line number makes it into the pot file.
+    # Replacing irrelevant lines by empty lines must happen before unescaping, because otherwise
+    # multi-line commands (lines ending on \) would get merged, changing the line count.
+    # Double quotes are escaped, and finally unescaped double quotes are added around the string.
+    # The result will be a file consisting of empty lines and potentially some lines prefixed with
+    # 'N_ ', followed by a double-quoted string.
+    string replace --regex $explicit_regex 'N_ $1' <$f \
+        | sed '/^N_ / !{s/^.*$//}' \
+        | string unescape \
```
krobelus, line 64: FWIW you can write

```fish
foo |
    bar |
    baz
```

without backslashes.

build_tools/fish_xgettext.fish
```diff
@@ -50,14 +50,29 @@ mkdir -p $tmpdir/explicit/share/completions $tmpdir/explicit/share/functions
...
-    # shell.
-    string replace --filter --regex $implicit_regex '$1' <$f | string unescape \
-        | string replace --all '"' '\\"' | string replace -r '(.*)' 'N_ "$1"' >$tmpdir/implicit/$f
+    # Works analogously to the explicit version.
+    string replace --regex $implicit_regex 'N_ $1' <$f \
+        | sed '/^N_ / !{s/^.*$//}' \
+        | string unescape \
+        | string replace --all '"' '\\"' \
+        | string replace --regex '^N_ (.*)$' 'N_ "$1"' \
+        >$tmpdir/implicit/$f
```
krobelus, line 75: looks good but wouldn't it be better to extract a function or a loop like `for target in explicti implicit; string replace $$target ... >$tmpdir/$/$ftarget` (ideally in a precursor commit),
then you'd only have to fix one place

build_tools/fish_xgettext.fish
danielrainer: Good to know, I changed that.

build_tools/fish_xgettext.fish
danielrainer: done

PR #11402, opened 2025-04-19
Add clippy lints about using stdlib print macros

danielrainer:
Context why we do not want to use these macros:
- https://github.com/fish-shell/fish-shell/pull/11397#discussion_r2050759690
- https://github.com/fish-shell/fish-shell/pull/11397#discussion_r2050759693
- https://github.com/fish-shell/fish-shell/pull/11397#discussion_r2050759696

These lints are not strictly about the macros, but at the moment they are implemented to flag `print!`, `println!`, `eprint!`, and `eptintln!`.

- https://rust-lang.github.io/rust-clippy/master/#print_stdout
- https://rust-lang.github.io/rust-clippy/master/#print_stderr
- https://doc.rust-lang.org/stable/clippy/lint_configuration.html#allow-print-in-tests

# TODO

- [ ] At the moment, there are a few uses of these macros in the code.
- [ ] For some reason, detection seems a bit flaky. Using rust-analyzer in my editor seems to work fine, but running `cargo clippy` manually does not find all problematic macro uses. (e.g. when adding a test containing `print!` and setting `allow-print-in-tests = false`, the `print!` usage is flagged by rust-analyzer in my editor but not by `cargo clippy`. I have no idea why.
- [ ] Which level do we want for the lint? (`warn`/`deny`)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -107,3 +107,9 @@ rust.unstable_name_collisions = "allow"
 clippy.manual_range_contains = "allow"
 clippy.needless_return = "allow"
 clippy.needless_lifetimes = "allow"
+
+# We do not want to use the e?print(ln)?! macros.
+# These lints flag their use.
+# In the future, they might change to flag other methods of printing.
+clippy.print_stdout = "deny"
+clippy.print_stderr = "deny"


krobelus:
>  Which level do we want for the lint? (warn/deny)

I think deny is fine because this is not something that will randomly change between different versions of rustc

> running cargo clippy manually does not find all problematic macro uses

interesting, let's create a minimal repro and check if it's a caching issue (`cargo clean`)

danielrainer:
> I think deny is fine because this is not something that will randomly change between different versions of rustc

Done

> interesting, let's create a minimal repro and check if it's a caching issue (`cargo clean`)

https://github.com/danielrainer/clippy_print_repro

Also compare the output of `cargo clippy` on this branch to the output of `rg 'e?print(ln)?!' src`. If you set `allow-print-in-tests = false` in `clippy.toml` you should get warnings for all usages of the print macros, but there are significantly more usages in the code than clippy reports. Can you reproduce this?

danielrainer:
The errors in https://github.com/fish-shell/fish-shell/actions/runs/14560323715/job/40842764058?pr=11402 match what I get when running `cargo clippy`.

danielrainer:
Ok, with the right flags to analyze everything, clippy reports the expected lints. The CI config already uses the flags. The issue there is probably that the build does not finish because we deny usage of the print macros. Locally, I get all lints when setting the severity to `warn`. So I think we can merge this as soon as all usages of `e?print(ln)?!` are removed.

It might also make sense to add an explanation to the docs where people running into this lint can find it easily. Maybe in `CONTRIBUTING.rst`?

krobelus:
> as soon as all usages of `e?print(ln)?!` are removed.

sounds good; I think there's no reason not to do that

> It might also make sense to add an explanation to the docs
> where people running into this lint can find it easily. Maybe in
> `CONTRIBUTING.rst`?

I'd say the commit message is the canonical place for this information.
It can be easily found via git-blame on the lines that add this lint.
That should be enough.

danielrainer:
> sounds good; I think there's no reason not to do that

At the moment, I think the only way of doing this would be switching to the `e?printf` macros. Adding some macros which do the formatting via stdlib `format!` and then write into `std::io::std{err,out}` is fairly simple, but they do not address the problem of supporting arbitrary output bytes. It would of course be possible to use such a macro just for situations where this problem does not occur, but there is no way of preventing users of such a macro from passing data which does not get decoded properly via the `Display` trait. (The private use area code points encoding arbitrary bytes would not get decoded.) Therefore, I am not a fan of introducing such a macro.

> I'd say the commit message is the canonical place for this information. It can be easily found via git-blame on the lines that add this lint. That should be enough.

The commit message is a good place to put information about this, but it requires finding out where the lints are enabled and relies on `git blame` still identifying the original commit adding these lines as the commit which last touched the lines.

In addition, the commit message will not change, so it might easily be outdated when someone runs into the lint. At the moment, we don't even know how we want to handle formatting and output in the future, so if we just use `e?printf` for now and add the lint, the commit message can of course explain why `e?print(ln)?` is undesirable, but its recommendations about alternatives might soon be outdated.

The issue is complex and important enough that I think proper documentation which can be found without having to search the git history is valuable to have.

Cargo.toml
```diff
@@ -107,3 +107,9 @@ rust.unstable_name_collisions = "allow"
 clippy.manual_range_contains = "allow"
 clippy.needless_return = "allow"
 clippy.needless_lifetimes = "allow"
+
+# We do not want to use the e?print(ln)?! macros.
+# These lints flag their use.
+# In the future, they might change to flag other methods of printing.
```
krobelus, line 113: I guess this line is redundant; we already say which four macros we are addressing here,
and I don't think we have any particular reason to believe that the lints will change.

clippy.toml
```diff
@@ -0,0 +1 @@
+allow-print-in-tests = false
```
krobelus, line 1: this doesn't seem to be working for me. I'm running `cargo clippy --workspace --all-targets` and I added `println!` in a unit test

krobelus: ah it's false and not true.
I don't have a strong opinion on this, either way seems fine for now.
If we go with false, we don't need to do so explicitly, since that's the default.

danielrainer: Yes, I had it set to true initially, but then thought it would be more consistent to disallow the macros everywhere, e.g. to avoid issues with `wstr` default formatting, so I set it to false. But you're right, there is not much point in explicitly setting the option to its default value.

danielrainer: I removed `clippy.toml`, so the print macros are still disallowed everywhere.

PR #11414, opened 2025-04-22
Remove e?println macros

danielrainer:
These macros are problematic because:
- They panic when the output stream is closed.
- They are not aware of fish's encoding of arbitrary bytes into a section of a Unicode private use area. The custom printf macros handle this.

More detailed explanations in #11397

In preparation for #11402

To avoid unnecessary complications, everything is passed to printf as a `(w)str` type, so the same formatting as before is applied to non-strings via `.to_string()`.

This should not result in behavioral changes aside from the differences between the `e?println` and `e?printf` macros. One exception is the behavior for `--print-debug-categories`. There, the `println`-based padding was broken, although I do not understand why. The `printf`-based version has padding.

### DIFF
--- a/src/bin/fish.rs
+++ b/src/bin/fish.rs
@@ -132,12 +132,12 @@ fn print_rusage_self() {
     let total_time = user_time + sys_time;
     let signals = rs.ru_nsignals;
 
-    eprintln!("  rusage self:");
-    eprintln!("      user time: {sys_time} ms");
-    eprintln!("       sys time: {user_time} ms");
-    eprintln!("     total time: {total_time} ms");
-    eprintln!("        max rss: {rss_kb} kb");
-    eprintln!("        signals: {signals}");
+    eprintf!("  rusage self:\n");
+    eprintf!("      user time: %s ms\n", sys_time.to_string());
+    eprintf!("       sys time: %s ms\n", user_time.to_string());
+    eprintf!("     total time: %s ms\n", total_time.to_string());
+    eprintf!("        max rss: %s kb\n", rss_kb.to_string());
+    eprintf!("        signals: %s\n", signals.to_string());
 }
 
 // Source the file config.fish in the given directory.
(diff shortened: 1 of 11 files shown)


src/threads.rs
```diff
@@ -206,7 +206,7 @@ pub fn spawn<F: FnOnce() + Send + 'static>(callback: F) -> bool {
             true
         }
         Err(e) => {
-            eprintln!("rust thread spawn failure: {e}");
+            eprintf!("rust thread spawn failure: %s\n", e.to_string());
```
krobelus, line 209: Error could probably get a `ToArg` implementation

krobelus (approved):

src/bin/fish.rs
```diff
@@ -293,7 +293,7 @@ fn fish_parse_opt(args: &mut [WString], opts: &mut FishCmdOpts) -> ControlFlow<i
                 activate_flog_categories_by_pattern(w.woptarg.unwrap());
                 for cat in flog::categories::all_categories() {
                     if cat.enabled.load(Ordering::Relaxed) {
-                        println!("Debug enabled for category: {}", cat.name);
+                        printf!("Debug enabled for category: %ls\n", cat.name);
```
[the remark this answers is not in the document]
danielrainer: Fair point. `%ls` seems to be more common in the codebase at the moment, but changing that would be simple.

src/threads.rs
danielrainer: Something like this?
```rust
impl<'a> ToArg<'a> for &'a std::io::Error {
    fn to_arg(self) -> Arg<'a> {
        Arg::String(self.to_string())
    }
}
```
Or do you mean by introducing a new enum variant on `Arg`?

krobelus:
It doesn't matter since we plan to remove all of them anyway

src/threads.rs
danielrainer: Done

Issue #11398, opened 2025-04-17
fish_trace should encode control characters


mqudsi:
With the following function definition:

```
function foo
    echo hello | string replace -r ^\xEF\xBB\xBF ''
end
```

Executing `fish_trace=1 foo` prints the following:

```
---> echo hello
---> string replace -r ^﻿ ''
hello
```

which *appears* to be simply the caret `^` followed by empty quotes `''` but in actuality is the following:

```
┌────────┬─────────────────────────┬─────────────────────────┬────────┬────────┐
│00000000│ 2d 2d 2d 3e 20 65 63 68 ┊ 6f 20 68 65 6c 6c 6f 0a │---> ech┊o hello_│
│00000010│ 2d 2d 2d 3e 20 73 74 72 ┊ 69 6e 67 20 72 65 70 6c │---> str┊ing repl│
│00000020│ 61 63 65 20 2d 72 20 5e ┊ ef bb bf 20 27 27 0a 68 │ace -r ^┊××× ''_h│
│00000030│ 65 6c 6c 6f             ┊                         │ello    ┊        │
└────────┴─────────────────────────┴─────────────────────────┴────────┴────────┘
```

We wouldn't want to encoded non-ascii as emoji, non-latin scripts, etc should probably be printed and left for the terminal to display (or not), but non-printable unicode sequences should probably be escaped.

PR #11415, opened 2025-04-22
Support bstr-based builtins

henrikhorluck:
This still needs a `bstr`-based `wgetopt`, and a utf8-based method for translation strings. Can `formatx!` and `wgettext!` co-exist?

Currently the argv to `builtin_run` is going a full loop-de-loop `bytes->wstr->bstr`, but making `Process.argv` use `bstr` is a bit larger change I would like to avoid for now.

Note that `bstr` might make it into `std` https://github.com/rust-lang/rust/issues/134915

This envision that usage of `streams.{out,err}.append{,ln}` is replaced with usage of `write{,ln}!(streams.{out,err}, "{}", x)`, and eventually no usage of `wstr`.


danielrainer:
The problem with non-UTF-8 data is that normal formatting via the `Display` trait does not work since it expects valid UTF-8 only, meaning we cannot use it to write arbitrary bytes. See e.g. https://docs.rs/bstr/latest/src/bstr/impls.rs.html#459-519, which results in invalid UTF-8 being replaced by the replacement codepoint `\u{FFFD}`. I've run into this problem trying to replace the `printf` macros with more Rust-like formatting.

printf/src/arg.rs
```diff
@@ -169,6 +171,20 @@ impl<'a, T> ToArg<'a> for &'a *const T {
     }
 }
 
+#[cfg(feature = "bstr")]
+impl<'a> ToArg<'a> for &'a BStr {
+    fn to_arg(self) -> Arg<'a> {
```
krobelus, line 176: > Can formatx! and wgettext! co-exist?

probably but it seems more intuitive to first change the format string syntax without changing string encoding.

printf/src/arg.rs
```diff
@@ -169,6 +171,20 @@ impl<'a, T> ToArg<'a> for &'a *const T {
...
+        Arg::String(self.to_string())
```
krobelus, line 177: I think we first want to figure out  how to deal with invalid UTF-8 in a way that is not easy to misuse.
Today we use `encode_byte_to_char` and `decode_byte_from_char`,
but we probably fail to call it in some cases (https://github.com/fish-shell/fish-shell/issues/11398).

AFAICT our ideal string type should implement `std::io::Write` and `std::fmt::Debug` but 
I'm not sure if it should implement `std::fmt::Write` or `std::fmt::Display`, those seem too easy to use accidentally.

One problem is that we have to remember to call `decode_byte_from_char()` in several different places.
Maybe we can use stronger types to make this easier: what if `FishString::chars()` didn't expose
raw (possibly PUA) chars directly but only an opaque type that can be compared to `char`.
This might be a lot of work.

src/io.rs
```diff
@@ -805,22 +826,27 @@ impl FdOutputStream {
...
-}
-impl StringOutputStream {
-    pub fn new() -> Self {
-        Default::default()
-    }
-    fn append(&mut self, s: &wstr) -> bool {
-        self.contents.push_utfstr(s);
-        true
+impl io::Write for FdOutputStream {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
```
krobelus, line 830: note that this doesn't match FdOutputStream::append() which does nothing if the stream errored previously.
Maybe we should move the error check to the callers so we can implement the std::io::Write
interface without surprising changes. But that may be too noisy.

Issue #11417, opened 2025-04-23
`set_color` is broken when using `-b normal`


danielrainer:
When running `set_color red -b normal; echo test`, the text `test` should be printed in red (foreground color). This works as expected before cebc05f6c1a4753110f816d214f81a5125071667, but is broken since this commit.

Comparison of old and new behavior for `set_color red -b normal | xxd`:

```xxd
# old
00000000: 1b5b 6d1b 5b6d 1b5b 3331 6d              .[m.[m.[31m
# new
00000000: 1b5b 6d                                  .[m
```

krobelus:
right, I think this should do it.

For backwards compatibility, `set_color normal` always resets everything.
`set_color --background=normal ...` starts by resetting everything, but we could change it to only reset the background color.

```diff
diff --git a/src/builtins/set_color.rs b/src/builtins/set_color.rs
index 61eecca99da..f2d23680395 100644
--- a/src/builtins/set_color.rs
+++ b/src/builtins/set_color.rs
@@ -148,10 +148,10 @@ pub fn set_color(parser: &Parser, streams: &mut IoStreams, argv: &mut [&wstr]) -
 
     // Here's some automagic behavior: if either of foreground or background are "normal",
     // reset all colors/attributes. Same if foreground is "reset" (undocumented).
-    // Note that either overwrite the attributes printed above! For "normal", this is probably wrong?
     if is_reset || [fg, bg].iter().any(|c| c.is_some_and(|c| c.is_normal())) {
         outp.reset_text_face();
-    } else {
+    }
+
     // Historically we have not used set_text_face() for colors here.
     // Doing so would add two magic behaviors:
     // - if fg and bg are equal, it makes one of them white
@@ -159,7 +159,7 @@ pub fn set_color(parser: &Parser, streams: &mut IoStreams, argv: &mut [&wstr]) -
     // The first one seems fine but the second one not really.
     outp.set_text_face(TextFace::new(Color::None, Color::None, Color::None, style));
     if let Some(fg) = fg {
-            if !outp.write_color(Paintable::Foreground, fg) {
+        if !fg.is_normal() && !outp.write_color(Paintable::Foreground, fg) {
             // We need to do *something* or the lack of any output messes up
             // when the cartesian product here would make "foo" disappear:
             //  $ echo (set_color foo)bar
@@ -167,8 +167,10 @@ pub fn set_color(parser: &Parser, streams: &mut IoStreams, argv: &mut [&wstr]) -
         }
[...]

PR #11418, opened 2025-04-23
builtin set_color: stop resetting unrelated attributes on --background=normal

krobelus:
I don't expect anyone relied on this behavior. Let's remove it, to reduce surprise.
This improves consistency with "set_color --underline-color=normal".

"set_color normal" still resets everything though.

As suggested in
https://github.com/fish-shell/fish-shell/issues/11417#issuecomment-2825023522

### DIFF
--- a/doc_src/terminal-compatibility.rst
+++ b/doc_src/terminal-compatibility.rst
@@ -165,13 +165,17 @@ Optional Commands
      -
      - Select background color from 24-bit RGB colors.
      -
+   * - ``\e[49m``
+     -
+     - Reset background color to the terminal's default.
+     -
    * - ``\e[58:2:: Ps : Ps : Ps m`` (note: colons not semicolons)
      - Su
      - Select underline color from 24-bit RGB colors.
      - kitty
    * - ``\e[59m``
      - Su
-     - Reset underline color (follow foreground color).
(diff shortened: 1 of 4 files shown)


src/builtins/set_color.rs
```diff
@@ -167,7 +167,9 @@ pub fn set_color(parser: &Parser, streams: &mut IoStreams, argv: &mut [&wstr]) -
         }
     }
     if let Some(bg) = bg {
-        if !bg.is_normal() {
+        if bg.is_normal() {
+            outp.write_command(DefaultBackgroundColor);
```
krobelus, line 171: to resolve the inconsistency for foreground, we could add `set_color --foreground=normal` which only resets the foreground color. Doesn't need to be added now though

doc_src/terminal-compatibility.rst
```diff
@@ -165,13 +165,17 @@ Optional Commands
      -
      - Select background color from 24-bit RGB colors.
      -
+   * - ``\e[49m``
+     -
+     - Reset background color to the terminal's default.
+     -
```
krobelus, line 171: Not sure if this is in terminfo. I guess I'll test this on some 40 terminals:

    set_color red --background blue; set_color --background=normal; echo red-on-default

PR #11419, opened 2025-04-23
Clean up shell scripts

danielrainer:
Some changes fix actual problems, e.g. missing spaces in square bracket tests, and backticks unintentionally causing code execution when intended as formatting.

Others, such as conservative quoting and using arrays instead of strings, probably work fine in the old version in most situations, but it's nice to have some additional safety.

Using `{ ..; }` instead of `(..)` is just a small performance enhancement.

Many of these issues were identified by [ShellCheck](https://www.shellcheck.net/), which might be useful in CI as well.

### DIFF
--- a/build_tools/mac_notarize.sh
+++ b/build_tools/mac_notarize.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/sh
 
 # Helper to notarize an .app.zip or .pkg file.
 
@@ -13,7 +13,7 @@ for INPUT in "$@"; do
     echo "Processing $INPUT"
     test -f "$INPUT" || die "Not a file: $INPUT"
     ext="${INPUT##*.}"
-    (test "$ext" = "zip" || test "$ext" = "pkg") || die "Unrecognized extension: $ext"
+    { test "$ext" = "zip" || test "$ext" = "pkg"; } || die "Unrecognized extension: $ext"
 
     xcrun notarytool submit "$INPUT" --keychain-profile AC_PASSWORD --wait
 
@@ -21,9 +21,7 @@ for INPUT in "$@"; do
         TMPDIR=$(mktemp -d)
         echo "Extracting to $TMPDIR"
         unzip -q "$INPUT" -d "$TMPDIR"
-        # Force glob expansion.
-        STAPLE_TARGET="$TMPDIR"/*
-        STAPLE_TARGET=$(echo $STAPLE_TARGET)
+        STAPLE_TARGET=$(echo "$TMPDIR"/*)
     else
         STAPLE_TARGET="$INPUT"
     fi
@@ -35,7 +33,7 @@ for INPUT in "$@"; do
         INPUT_FULL=$(realpath "$INPUT")
         rm -f "$INPUT"
         cd "$(dirname "$STAPLE_TARGET")"
-        zip -r -q "$INPUT_FULL" $(basename "$STAPLE_TARGET")
+        zip -r -q "$INPUT_FULL" "$(basename "$STAPLE_TARGET")"
     fi
     echo "Processed $INPUT"
 
--- a/build_tools/osx_package_scripts/add-shell
+++ b/build_tools/osx_package_scripts/add-shell
@@ -4,12 +4,12 @@
 
 if test $# -eq 0
 then
-	echo usage: $0 shellname [shellname ...]
+	echo "usage: $0 shellname [shellname ...]"
 	exit 1
 fi
 
-scriptname=`basename "$0"`
-if [[ $UID -ne 0 ]]; then
+scriptname=$(basename "$0")
+if [ "$(id -u)" -ne 0 ]; then
      echo "${scriptname} must be run as root"
      exit 1
 fi
@@ -20,6 +20,7 @@ tmpfile=${file}.tmp
 
 set -o noclobber
 
+# shellcheck disable=SC2064
 trap "rm -f $tmpfile" EXIT
 
 if ! cat $file > $tmpfile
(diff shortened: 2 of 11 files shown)


build_tools/mac_notarize.sh
```diff
@@ -13,17 +13,16 @@ for INPUT in "$@"; do
...
 
     if test "$ext" = "zip"; then
         TMPDIR=$(mktemp -d)
         echo "Extracting to $TMPDIR"
         unzip -q "$INPUT" -d "$TMPDIR"
-        # Force glob expansion.
-        STAPLE_TARGET="$TMPDIR"/*
-        STAPLE_TARGET=$(echo $STAPLE_TARGET)
+        STAPLE_TARGET_ARRAY=("$TMPDIR"/*)
+        STAPLE_TARGET="${STAPLE_TARGET_ARRAY[*]}"
```
krobelus, line 25: this script is valid POSIX shell; I wouldn't want to use Bash extensions without a good reason.
I'd only fix the inadvertent word splitting by writing `STAPLE_TARGET=$(echo "$TMPDIR"/*)`

But it's fine as-is, this is not a blocker

build_tools/osx_package_scripts/add-shell
```diff
@@ -4,12 +4,12 @@
...
 then
-	echo usage: $0 shellname [shellname ...]
+	echo "usage: $0 shellname [shellname ...]"
 	exit 1
 fi
 
-scriptname=`basename "$0"`
-if [[ $UID -ne 0 ]]; then
+scriptname=$(basename "$0")
+if [ "$(id --user)" -ne 0 ]; then
```
krobelus, line 12: Probably this long option doesn't work on mac - in doubt use `id -u`

build_tools/osx_package_scripts/add-shell
```diff
@@ -32,15 +33,15 @@ EOF
 fi
 
 # Append a newline if it doesn't exist
-if [ "$(tail -c1 "$tmpfile"; echo x)" != $'\nx' ]; then
+if [ -z "$(tail -c1 "$tmpfile")" ]; then
```
krobelus, line 36: This changes the behavior to do the opposite of what it should do.
The command substitution will be empty iff there is a final newline.
In that case we don't want to append.

Either leave it or go with `-n` though that will treat empty files differently

docker/context/fish_run_tests.sh
```diff
@@ -3,7 +3,7 @@
 # This script is copied into the root directory of our Docker tests.
 # It is the entry point for running Docker-based tests.
 
-cd ~/fish-build
+cd ~/fish-build || exit 1
```
krobelus, line 6: this should probably use `set -e` and `set +e` but I guess this is fine too. It's weird that we `exit` with 0 below

docker/docker_run_tests.sh
```diff
@@ -2,15 +2,15 @@
...
+Usage: $(basename "$0") [--shell-before] [--shell-after] DOCKERFILE
 Options:
     --shell-before   Before the tests start, run a bash shell
     --shell-after    After the tests end, run a bash shell
 EOF
     exit 1
 }
 
-DOCKER_EXTRA_ARGS=""
+DOCKER_EXTRA_ARGS=()
```
krobelus, line 13: I'm not sure if we use Bashishms anywhere else? I'm not counting things like `$'\n'` which has been added to POSIX recently

build_tools/osx_package_scripts/add-shell
danielrainer: You're right, good catch. I changed the `if` to `||`, so now it should append a newline iff the test fails (when the last character is not a newline).

docker/docker_run_tests.sh
danielrainer: Some of the scripts specify bash in the shebang, including this one which uses `BASH_SOURCE`. But I also prefer using `/bin/sh` rather than requiring bash, so I changed that.

Now, the only remaining script requiring bash is `build_tools/make_pkg.sh` which uses arrays. It could be converted as well, but handling that correctly might result in somewhat unwieldy code.

krobelus: > Now, the only remaining script requiring bash is build_tools/make_pkg.sh which uses arrays. It could be converted as well, but handling that correctly might result in somewhat unwieldy code.

POSIX shell does have arrays -- well, one array, `$@`, which is often enough.
But I don't think there's a strong reason to change it; I'm not sure which shell we want.

Rest looks looks good.

osx/install.sh
```diff
@@ -5,7 +5,7 @@ set -e
 
 # Make sure we're run as root
 scriptname=$(basename "$0")
-if [ "$UID" -ne 0 ]; then
+if [ "$(id --user)" -ne 0 ]; then
```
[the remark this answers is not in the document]
danielrainer: Yes, I missed the second instance.

docker/context/fish_run_tests.sh
```diff
@@ -10,15 +12,15 @@ cmake -G Ninja -DCMAKE_BUILD_TYPE=Debug /fish-source "$@"
 # Spawn a shell if FISH_RUN_SHELL_BEFORE_TESTS is set.
 if test -n "$FISH_RUN_SHELL_BEFORE_TESTS"
 then
-    bash -i || exit
+    sh -i
 fi
 
-ninja && ninja fish_run_tests
+(set +e; ninja && ninja fish_run_tests);
```
krobelus, line 18: the semicolon is unnecessary

docker/context/fish_run_tests.sh
```diff
@@ -10,15 +12,15 @@ cmake -G Ninja -DCMAKE_BUILD_TYPE=Debug /fish-source "$@"
...
 RES=$?
 
 # Drop the user into a shell if FISH_RUN_SHELL_AFTER_TESTS is set.
 if test -n "$FISH_RUN_SHELL_AFTER_TESTS"; then
-    bash -i
+    sh -i
```
krobelus, line 23: these interactive shells are for debugging; I think we want to keep running `bash -i` here because bash has nice key bindings etc. as opposed to /bin/sh in case that's not bash.
This is minor, I guess I'll revert this bit

danielrainer: Bash is definitely nicer than a plain POSIX shell. I reverted it, but the general approach of avoiding bashisms seems a bit strange if we then explicitly require bash (although in this case it is only required if an interactive shell is explicitly requested).

krobelus: it's a matter of taste. I like using POSIX shell for scripting because it's simple and standardized.
But I don't think everyone agrees

PR #11421, opened 2025-04-24
Restore color precedence in highlighters, update docs on fallback colors

### LINKED ISSUE #11420: Setting fallback background color for `fish_color_selection` results in it being used in 24-bit color mode
This regression is introduced by 037c1896d45ca174dbaef0ba2297f24dc4bad9f5.

To reproduce:

```fish
set -l selection blue red
set -l fish_color_selection --background=$selection
```
Then, enter vi selection mode and select something. The background color of the selection will be red, even though it should be blue in 24-bit color mode.

krobelus:
Closes #11420

### DIFF
--- a/doc_src/cmds/set_color.rst
+++ b/doc_src/cmds/set_color.rst
@@ -30,7 +30,8 @@ If :envvar:`fish_term24bit` is set to 0, fish will translate RGB values to the n
 If :envvar:`fish_term256` is also set to 0, fish will translate them to the 16-color palette instead.
 Fish launched as ``fish -d term_support`` will include diagnostic messages that indicate the color support mode in use.
 
-A second color may be given as a desired fallback color. e.g. ``set_color 124212 brblue`` will instruct set_color to use *brblue* if a terminal is not capable of the exact shade of grey desired. This is very useful when an 8 or 16 color terminal might otherwise not use a color.
+If multiple colors are specified, fish prefers the first RGB one.
+However if :envvar:`fish_term256` is set to 0, fish prefers the first named color.
 
 The following options are available:
 
--- a/doc_src/cmds/set_color.rst
+++ b/doc_src/cmds/set_color.rst
@@ -61,12 +61,10 @@ The following options are available:
 **-h** or **--help**
     Displays help about using this command.
 
-Using the **normal** keyword will reset foreground, background, and all formatting back to default.
-
 Notes
 -----
 
-1. Using the **normal** keyword will reset both background and foreground colors to whatever is the default for the terminal.
+1. Using **set_color normal** will reset all colors and modes to the terminal's default.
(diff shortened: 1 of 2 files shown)


doc_src/cmds/set_color.rst
```diff
@@ -30,7 +30,8 @@ If :envvar:`fish_term24bit` is set to 0, fish will translate RGB values to the n
 If :envvar:`fish_term256` is also set to 0, fish will translate them to the 16-color palette instead.
 Fish launched as ``fish -d term_support`` will include diagnostic messages that indicate the color support mode in use.
 
-A second color may be given as a desired fallback color. e.g. ``set_color 124212 brblue`` will instruct set_color to use *brblue* if a terminal is not capable of the exact shade of grey desired. This is very useful when an 8 or 16 color terminal might otherwise not use a color.
+If multiple colors are specified, fish prefers the first RGB one.
+However if :envvar:`fish_term256` is set to 0, fish prefers the first named color.
```
danielrainer, line 34: This does not seem to match current behavior. Setting `fish_term24bit` and `fish_term256` to zero and testing this in the Linux console results in the first color being used, e.g.:

```fish
set -l selection 00fa00 red
set -l fish_color_selection --background=$selection
```
Results in selections being displayed in the consoles fallback mapping of `00fa00`, which is green instead of red.

Initially, I thought `the first named color` refers to the first listed color, rather than the first color specified using a name rather than an RGB color code. Depending on the actual behavior we decide  on, this could maybe be phrased more clearly.

krobelus: > This does not seem to match current behavior.

it does for foreground colors. Not for background colors.
It's an open question whether we should extend this logic to background/underline colors.

I'm looking for a real-world example ("red -> green" sounds contrived).

> Initially, I thought the first named color

right -- "the first named color specified".

danielrainer: > I'm looking for a real-world example ("red -> green" sounds contrived).

I picked this as a minimal example to show the effect. I actually noticed it because my Linux console colors (especially in my prompt) changed and became hard to read.

This happens for two reasons:
1. RGB colors which have sufficient contrast to be used for fg+bg colors might get mapped to colors which are more similar (or even identical).
2. If the fallback color scheme is designed separately, it can very well look completely different from the truecolor scheme. E.g. one might have a bright background and light foreground color when truecolor is supported, but a light background and dark foreground color as a fallback. When these colorschemes then get mixed the result would be significantly reduced contrast (maybe none at all). For this reason I think it is important to handle fallback for foreground and background colors in the same way.

krobelus: `fish_pager_color_description B3A06D yellow` is one example that isn't really readable on Linux console without the fallback.

I guess that's enough reason to keep it and also extend the fallback behavior to `--background` etc.
Still not sure if it's a good idea to still use/advertise this feature, but at least we can make this consistent between `set_color` and `fish_color_*`.

I wonder if it's okay that `set fish_color_error badcolor red` now does nothing.
I guess we could restore the old behavior (ignore invalid colors) but that'd be inconsistent with `set_color`.

PR #11436, opened 2025-04-29
Builtin XTGETTCAP support, use curly underlines for errors

krobelus:
<details>

<summary>
Builtin for querying terminal capabilities/name/version
</summary>

The next commit wants to add a conditional default for styled underlines.
Due to various incompatibilities in terminals, our best option seems to ask
the terminal.

Today we can make XTGETTCAP queries using something like

	printf '\eP+q5373\e\\' # Su
	printf '\e[0c'
	while read -n 1
	    ...
	end

This doesn't seem safe because builtin read might consume other data written
by the terminal such as keyboard input.

Avoid this problem by providing it as a builtin that can enqueue unrelated
input and process it after the query has been answered.

For the same reason, add a builtin to query for XTVERSION; this allows us
to add workarounds for specific terminals (example in the next commit).

TODO: naming -- maybe add a level of nesting:

	status query-terminal xtgettcap
	status query-terminal xtversion
	# Possible future additions:
	status query-terminal os-name
	status query-terminal cursor-position

or "status query-xtgettcap".
We could also hide the exact protocol by saying
"status query terminfo-capability".

Note that xtgettcap, xtversion and os-name are expected to always give the
same results throughout the lifetime of the fish process.  Keep caching
XTVERSION as before, though that's probably not needed.

Future work:

XTGETTCAP potentially supports all of terminfo: boolean, numeric and string
capabilities. Today we have no use beyond checking for presence/absence of
a capabilty. If we ever need more, we can use stdout.

</details>

<details>

<summary>
Use curly underlines for errors in default themes
</summary>

Given that this curly underline will also be red, it should be widely
understood as error.

Since fish always renders immediately (and synchronously), typing "echo"
will briefly show an intermediate curly line.  Maybe fish should redraw
after a timer elapses? This is probably unrelated to this patch.

As mentioned in cc9849c279d (Curly underlines in set_color and fish_color_*,
2025-04-13), there are still some terminals that interpret "\e[4:3m" as
something other than curly underline.

Some of them interpret it as background/foreground color, and some terminal
multiplexers downgrade it to straight underlines (which often happens due
to a false positive).  I want to change this in multiplexers where possible
(see https://github.com/orgs/tmux/discussions/4477) but for now, disable
this feature in multiplexers (there are just a handful).

In a few years, those terminals will maybe agree with XTerm.  Until then,
use XTGETTCAP as a temporary stepping stone. We could also read the terminfo
database but that will give only very few true positives, and lots of false
negatives.  Better implement XTGETTCAP in the relevant terminals.

Note that if the universal variables use the "--track" flag (from https://github.com/fish-shell/fish-shell/pull/11435), then

	rm -rf /tmp/newhome
	foot -e env $HOME=/tmp/newhome fish
	xterm -e env $HOME=/tmp/newhome fish

will "magically work". For foot, $fish_color_error will have --underline=curly.
For xterm, it will not, due to the call to "fish_config theme update".
But of course since it's a universal variable, running fish in xterm
would also downgrade the fish running in other terminals.

Add a "fish_config theme save --yes" flag because "status xtgettcap"
requires stdin to be a terminal.  I'd probably drop this requirement, and
make "status xtgettcap" always use fish's stdin and not its own.  That'd be
cheating because an external command can't do that but I don't think this
change would be hurting anyone.

</details>

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -69,6 +69,9 @@ Improved terminal support
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 - Support for curly underlines in `fish_color_*` variables and :doc:`set_color <cmds/set_color>` (:issue:`10957`).
 - Underlines can now be colored independent of text (:issue:`7619`).
+- Errors are now highlighted with curly underlines in the default themes.
+  For compatibility with terminals that interpret the curly-underline escape sequence in an unexpected way,
+  the default themes enable this only if the terminal advertises support for the ``Su`` capability via XTGETTCAP.
 - New documentation page `Terminal Compatibility <terminal-compatibility.html>`_ (also accessible via ``man fish-terminal-compatibility``) lists required and optional terminal control sequences used by fish.
 - :doc:`status <cmds/status>` learned the ``xtgettcap`` subcommand, to query terminfo capabilities via XTGETTCAP commands.
 - :doc:`status <cmds/status>` learned the ``xtversion`` subcommand, to show the terminal's name and version (as reported by via the XTVERSION command).
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -70,6 +70,8 @@ Improved terminal support
 - Support for curly underlines in `fish_color_*` variables and :doc:`set_color <cmds/set_color>` (:issue:`10957`).
 - Underlines can now be colored independent of text (:issue:`7619`).
 - New documentation page `Terminal Compatibility <terminal-compatibility.html>`_ (also accessible via ``man fish-terminal-compatibility``) lists required and optional terminal control sequences used by fish.
+- :doc:`status <cmds/status>` learned the ``xtgettcap`` subcommand, to query terminfo capabilities via XTGETTCAP commands.
+- :doc:`status <cmds/status>` learned the ``xtversion`` subcommand, to show the terminal's name and version (as reported by via the XTVERSION command).
 
 Other improvements
 ------------------
--- a/doc_src/cmds/status.rst
+++ b/doc_src/cmds/status.rst
@@ -33,6 +33,8 @@ Synopsis
     status buildinfo
     status get-file FILE
     status list-files [PATH]
+    status xtgettcap TERMINFO-CAPABILITY
+    status xtversion
 
 Description
 -----------
@@ -116,6 +118,13 @@ The following operations (subcommands) are available:
 **list-files** *FILE*
     This lists the files embedded in the fish binary at compile time. Only files where the path starts with the optional *FILE* argument are shown.
     Returns 0 if something was printed, 1 otherwise.
+
+**xtgettcap** *TERMINFO-CAPABILITY*
+    Query the terminal for a terminfo capability via XTGETTCAP.
+    Returns 0 if the capability is present and 1 otherwise.
+
+**xtversion**
+    Show the terminal name and version (XTVERSION).
 
 Notes
 -----
--- a/doc_src/terminal-compatibility.rst
+++ b/doc_src/terminal-compatibility.rst
@@ -177,7 +177,9 @@ Optional Commands
      - Su
      - Reset underline color to the default (follow the foreground color).
      - kitty
-   * - ``\e[ Ps S``
+   * - .. _indn:
+
+       ``\e[ Ps S``
      - indn
      - Scroll forward Ps lines.
      -
@@ -269,6 +271,12 @@ Optional Commands
      - FinalTerm
    * - ``\eP+q Pt \e\\``
      -
-     - Request terminfo capability (XTGETTCAP). The parameter is the capability's hex-encoded terminfo code.
-       Specifically, fish asks for the ``indn`` string capability. At the time of writing string capabilities are supported by kitty and foot.
-     - XTerm, kitty, foot
+     - Request terminfo capability (XTGETTCAP).
+       The parameter is the capability's hex-encoded terminfo code.
+       To advertise a capability, the response must of the form
+       ``\eP1+q Pt \e\\`` or ``\eP1+q Pt = Pt \e\\``.
+       In either variant the first parameter must be the hex-encoded terminfo code.
+       In the second variant, fish ignores the part after the equals sign.
+
+       At startup, fish checks for the presence of the `indn <#indn>`_ string capability.
+     - XTerm, kitty, foot, wezterm, contour
--- a/share/completions/status.fish
+++ b/share/completions/status.fish
@@ -27,7 +27,9 @@ set -l __fish_status_all_commands \
     list-files \
     print-stack-trace \
     stack-trace \
-    test-feature
+    test-feature \
+    xtgettcap \
+    xtversion
 
 # These are the recognized flags.
 complete -c status -s h -l help -d "Display help and exit"
@@ -64,6 +66,8 @@ complete -f -c status -n "not __fish_seen_subcommand_from $__fish_status_all_com
 complete -f -c status -n "not __fish_seen_subcommand_from $__fish_status_all_commands" -a get-file -d "Print an embedded file from the fish binary"
 complete -f -c status -n "not __fish_seen_subcommand_from $__fish_status_all_commands" -a list-files -d "List embedded files contained in the fish binary"
 complete -f -c status -n "not __fish_seen_subcommand_from $__fish_status_all_commands" -a fish-path -d "Print the path to the current instance of fish"
+complete -f -c status -n "not __fish_seen_subcommand_from $__fish_status_all_commands" -a xtgettcap -d "Query the terminal for a terminfo capability"
+complete -f -c status -n "not __fish_seen_subcommand_from $__fish_status_all_commands" -a xtversion -d "Show terminal name and version"
 
 # The job-control command changes fish state.
 complete -f -c status -n "not __fish_seen_subcommand_from $__fish_status_all_commands" -a job-control -d "Set which jobs are under job control"
--- a/share/functions/__fish_config_interactive.fish
+++ b/share/functions/__fish_config_interactive.fish
@@ -30,7 +30,7 @@ if status is-interactive
     # Commands to run in interactive sessions can go here
 end" >$__fish_config_dir/config.fish
 
-        echo yes | fish_config theme save "fish default" --track
+        fish_config theme save "fish default" --yes --track
         set -Ue fish_color_keyword fish_color_option
     end
     if test $__fish_initialized -lt 3800 && test "$fish_color_search_match[1]" = bryellow
--- a/share/functions/__fish_in_gnu_screen.fish
+++ b/share/functions/__fish_in_gnu_screen.fish
@@ -1,0 +1,3 @@
+function __fish_in_gnu_screen
+    test -n "$STY" || contains -- $TERM screen screen-256color
+end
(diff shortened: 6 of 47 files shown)


faho:
This is a *bunch* of complexity to add for a pretty niche feature. I'm not sold.

krobelus:
Thanks for the feedback. Feels like you're not happy with the current state, or this draft of a proposed state, which is a generalization of the current state. What would you need to be content?
I'm happy to elaborate the motivating factors for this direction but I do not know if this is asked for.

faho:
I'm not happy with the querying in general, because I'm not sold that it provides a reduction in workarounds and adds an increase in complexity.

Basically you're adding querying, but then need to work around it in 4 (?) terminals because it breaks stuff, so why not do without querying and work around it in a simpler manner in Y terminals? Especially since we don't have a good answer to terminals not replying at all.

And if we can't add curly underlines "everywhere" (I'm assuming the issue is that some terminals show regular underlines, which would be misleading), then I'm not sold that querying *more*, and making this more complicated, is worth it, as I see it as a very minor feature.

And querying via `status` doesn't seem like the sort of feature anyone but us would use, and we would only really use it in this and like two other cases.

krobelus:
After writing this, it seems there are two overarching questions:
1. how much and how do we want to work around terminals that are not compatible
   with fish.
2. what should be plan for terminal features like `indn` that we can't safely
   use if they are absent? `indn` seems trivial to implement in all terminals
   and then we can use it after a few years, but 1. other features are not
   so trivial and 2. waiting for years until we can upgrade a default may
   make fish feel dated.

> I'm not happy with the querying in general, because I'm not sold that it provides a reduction
> in workarounds and adds an increase in complexity.

Right. I try to think of a desired future state (errors should be undercurled), and then think about how to get there.
The sane way to get there is via a command I can send to my terminal that
tells it to draw curly underlines (or not), but never draw regular underlines.

This approach is a shortcut or detour, depending on how you look at it.
Taking on short-term workarounds can sometimes pay off in the long term..

> Basically you're adding querying, but then need to work around it in 4 (?) terminals because

yeah, the nested ones -- screen, emacs ansi-term, emacs-vterm, dvtm, abduco, tmux.
Hopefully most/all workarounds will go away, see below.

> it breaks stuff, so why not do without querying and work around it in a simpler manner in Y
> terminals?

Yeah that's an alternative, but
1. Until terminal vendors cooperate, it is impossible.
   For Terminal.app, the best we can do is `test $TERM_PROGRAM = Apple_Terminal`.
   I assume this is not good enough because it's disabled in SSH and others.
   (I did file issues asking to fix curly underlines and for unique identification via XTVERSION).
   JetBrains' terminal is open-source so it should be easy to fix (and then wait a few years).
   - We can get creative and do something like "if the terminal does not respond to XTVERSION, disable
     any advanced features". But XTGETTCAP seems much less surprising.
2. I'm assuming we already want an implementation of something like XTGETTCAP, to check `indn` --
   according to `doc_src/terminal-compatibility.rst` this is _not_ a workaround.
3. It's true that using XTGETTCAP to check for curly underline support is a workaround that
   shouldn't be necessary forever, however it is at least terminal-agnostic; it can be easily
   implemented by any terminal (a partial implementation is totally fine for us, though I'm
   not sure if others also treat "no response" as "use a sane default").

> Especially since we don't have a good answer to terminals not replying at all.

It is true that having to follow up any request with another primary device attribute is unintuitive, but it does work.
In far future, if we can assume something like XTGETTCAP, then we'll get proper negative responses.
If you are referring to the problem where a terminal does not reply to primary DA, because of either
1. it got into a weird state (reading a command with arbitrary text) where it swallows the primary DA request
2. the PTY is not backed by a real terminal
3. it doesn't implement primary DA at all

I think for 1 and 2, a timeout in fish would work well, because 1 should be very rare /
otherwise broken (if a user manages to get into such a state, all of their other key presses
will be ignored by the terminal) and 2 mean that no user is actively typing something, so
waiting for a second should be fine.
Declaring 3 unsupported seems a small price to pay to be able to add backwards-incompatible
features in future.

So I think it's fine; I'm not sure what's the nature of insufficiency here.

> And if we can't add curly underlines "everywhere"
> (I'm assuming the issue is that some terminals
> show regular underlines, which would be misleading),

Yeah (but not only).  Besides Terminal.app, there are two others that misinterpret it (see cc9849c279d).

The ones that draw regular underlines instead are terminal multiplexers.
Among "real" terminals, only Terminology violates this, which is surely an accident, easily fixed.
I'll send a patch to tmux; they agreed to change it.

As for the remaining ones (emacs ansi-term, emacs-vterm, screen, vim), that is a question of
whether they agree and if not, whether we want to work around them.
Maybe if we can get the majority out of emacs-vterm, screen and Vim to acquiesce, then we should be more or less good.
Regular underline on syntax and command-not-found errors are are less tragic than glitches we saw from `CSI = 5u`.
(NeoVim's :terminal does support curly underlines)

Arguments in favor of workarounds are:
1. terminal multiplexers are special already, for example they don't give us access
   to, say, desktop notifications or the GUI clipboard. So I'd say there is a non-zero chance
   that we want other workarounds for multiplexers. We currently have one in fish_clipboard_copy,
   to copy to the parent terminal's system's clipboard.
   Needless to say, that if these are here to stay, we'd prefer to make them independent of tmux.
2. tmux versions can probably be 5 years out of date (e.g. the defaults on work systems).
   So rather than setting myself a reminder to enable this feature in 5 years,
   I'd rather add it now but disabled for tmux versions <3.6,
   (I've been pondering a `skip-terminal-workarounds` feature flag, that would skip the `status
   xtgettcap Su` etc -- that might be handy for people hacking on fixing terminals)

> then I'm not sold that querying more, and making this more complicated,

I wonder if this is about implementation complexity or surface complexity.
The implementation suffers from some historical baggage but we're slowly getting rid of that.
As for surface -- I think for workarounds it may be a good idea if they are implemented at a
high level, not totally hidden from the user and easier to turn off.

> is worth it, as I see it as a very minor feature.

Possibly.  Besides the fact that deciding subjective things is hard, the reason I
wouldn't spend much thought on this, is that when I extrapolate, most of this machinery
can be reused for querying various other things.

- other terminfo codes (not sure which ones are useful)
- cursor position (not sure what it's realistically useful for)
- xtversion (which is a one way for users to toggle certain behavior, or work around terminal bugs)
- name of the OS the terminal is running on
- [light/dark mode](https://github.com/contour-terminal/contour/blob/master/docs/vt-extensions/color-palette-update-notifications.md)
- foreground/background colors
- whether OSC 52 paste is allowed (that would have helped me when I was using macOS)
- whether desktop notifications are supported (for job summary)
- specify [text width](https://github.com/kovidgoyal/kitty/blob/master/docs/text-sizing-protocol.rst) (or size in general), to fix the wcwidth mismatch problem.
  (This feature does not inherently require querying and someone asked to drop this requirement but that didn't make it in yet. We should probably speak up sooner rather than later).

Of course we shouldn't put too much weight on hypotheticals, I'd rather find the next concrete
thing that's actually useful (like curly underlines, the weirdly-named scrollback-push, mouse click_events) and try that.

Also note that when we query during fish initialization, it can't be safely implemented as an
external command (because that one might lose keyboard input). fish already wants to decode other input
from the terminal like keys, focus events so the full thing seems within the scope,

>
> And querying via status doesn't seem like the sort of feature anyone but us would use, and
> we would only really use it in this and like two other cases.

Of course I'm tempted to make it private/hidden but that would be unfishy.
(On that topic, get-file/list-files exist for historical reasons; maybe we should move any
third-party consumers to use dedicated interfaces).

We could hide them in a `status debug` umbrella command that could do things like control log levels
or dump debug information.

In doubt, we can surely keep hardcode the queries in the Rust sources for
now, and expose them as read-only/electric `$__fish_supports_curly_underline`
when needed.

I'm not in a rush to expose this feature but one possible argument for doing that:
it'd be more convenient to put workarounds in fish script, which would make things easier to
change and less magic; we can even add them to the bind outiput for full transparency.

    set -g fish_kitty_keyboard_protocol 0
    if status query-terminal kitty-keyboard-flags >/dev/null
        set -g fish_kitty_keyboard_protocol 1
    end
    bind ctrl-l 'if status query-terminal xtgettcap indn; commandline -f scrollback-push; end' clear-screen
    if test $(status xtversion) = tmux...;
    end

(I wonder if we should add settings like `fish_kitty_keyboard_protocol`,
`fish_mark_prompts`, etc to work around terminal bugs and for debugging. I
didn't expect so many issues)

faho:
> I'd rather find the next concrete
thing that's actually useful (like curly underlines, the weirdly-named scrollback-push, mouse click_events) and try that.

I think one big thing here is that I don't need any of those things. I'm fine just not having those features (including your big list).

In the case of curly underlines specifically, in my default terminal at my default font size I find them barely distinguishable from regular underlines. (to be honest `set -g fish_color_error red --underline` would work fine in my case because it makes the underlines red as well)

It's all on the far end of diminishing returns for me. I don't see a killer feature here.

And then the cost is that we're introducing all this complexity and currently the potential for fish just *hanging* indefinitely. And at that point I'd rather just remove the features.

> Of course I'm tempted to make it private/hidden but that would be unfishy.

It's not about it being "hidden", it's that the feature only seems useful *to us* in some very niche circumstances, which to me is a good sign that it may not be worth adding at all.

> I'm not in a rush to expose this feature but one possible argument for doing that:
it'd be more convenient to put workarounds in fish script

We've had good success putting more of these things in the *compiled code* - e.g. check the history of our RGB support. That became much simpler by just doing it in C++ instead.

> I wonder if we should add settings like fish_kitty_keyboard_protocol,
fish_mark_prompts, etc to work around terminal bugs and for debugging. I
didn't expect so many issues

We already *have* a setting for the kitty keyboard protocol, and we already *know* the terminals that break with it or prompt marking, and it's mostly a couple of terrible proprietary windows ssh clients (and emacs ansi-term, which has been broken one way or another for as long as I've known about it).

At which point I'm fine just telling people to get a better terminal, they're available for free. And that's what we've done so far, and I don't see a big mob of bad-terminal-users in front of my house.

ridiculousfish:
I like the style but I don't think this rises to the level of justifying a new builtin, especially because we should try to avoid introducing new terminal querying, which is fragile.

I like the idea of keeping the `--underline=curly` approach and just documenting running the command `set fish_color_error bf616a --underline=curly`. I know we prefer to turn on new features by default but the level of complexity to detect this is too high IMO.

(One idea we've discussed in the past was a "quiz" that detects things like whether emojis are width 1 or 2; this could also go in there.)

krobelus:
It shouldn't be a builtin, it's fine to hardcode the concrete query in Rust and expose the result as variable.
Then the question boils down to "should fish script ("config.fish") be able to use query responses?".

Obviously we should go with something that maintainers are comfortable with.

I don't know what that entails exactly, i.e., what kinds of query-related features fish wants to use
(now and in future), what for (backwards-incompatible features, temporary workarounds for
incompatible terminals) and where (compiled sources, fish script).
(I'm also not sure what's our desired level of support for Terminal.app, maybe "all current themes must be compatible for 5+ years but not necessarily new ones".)

I'm happy to discuss the technical details but I may have already gone overboard with that and I'm not sure if that's relevant -- it's subjective. It's difficult for me to change my wrong opinions, so I'm waiting for someone to call the shots on the direction, even if we don't agree yet. That seems more economical than a yes/no every time this implementation technique is used.
If we're not sure yet, that's fine, I can keep these things in my source tree for now, or perhaps behind a feature flag?

> (One idea we've discussed in the past was a "quiz" that detects things like whether emojis are
width 1 or 2; this could also go in there.)

Yeah, querying the actual on-screen-width via Cursor Position Report is also how to detect support for the [text sizing protocol](https://github.com/kovidgoyal/kitty/blob/master/docs/text-sizing-protocol.rst#detecting-if-the-terminal-supports-this-protocol). The width aspect of that protocol is a general solution to the wcwidth-mismatch problem. We could do a preliminary version just for emoji. Not sure.
Cursor Position Report can also serve as crutch to implement multi-line autosuggestions without pushing the prompt up when the autosuggestion hits the end to screen.

PR #11438, opened 2025-04-29
wpctl: add completion for settings option

fabiojb:
## Description

Add completion for the `wpctl settings` CLI option, added on [wireplumber 0.4.90](https://gitlab.freedesktop.org/pipewire/wireplumber/-/releases/0.4.90). The completion will be conditionally available on wpctl versions that contain this option.

## TODOs:

### DIFF
--- a/share/completions/wpctl.fish
+++ b/share/completions/wpctl.fish
@@ -1,5 +1,9 @@
 set -l commands status get-volume inspect set-default set-volume set-mute set-profile clear-default
 
+if wpctl settings &> /dev/null;
+    set --append commands settings
+end
+
 function __wpctl_get_nodes -a section -a type
     set -l havesection
     set -l havetype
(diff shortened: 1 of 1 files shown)


share/completions/wpctl.fish
```diff
@@ -38,6 +42,18 @@ function __wpctl_command_shape
     end
 end
 
+function __wpctl_get_settings
+    set -l wpctl_settings (wpctl settings | string collect)
```
krobelus, line 46: should maybe silence the error on old versions (`wpctl settings 2>/dev/null`) though I guess it's not a big deal

krobelus (approved):

share/completions/wpctl.fish
fabiojb: wpctl seems to not print to sderr when an unknown command is called, instead it print a help message and return exit status 1.

Just to be sure, I tested it now on ubuntu 24 that uses wireplumber 0.4.17, using fish 4+, and I didn't notice any unexpected behavior.

krobelus:
> wpctl seems to not print to sderr when an unknown command is called, instead it print a help message and return exit status 1.

thanks, pushing a fix

PR #11443, opened 2025-05-01
preciser wording for builtin crossmode alt+p binding

Axlefublr:
## Description

First time I saw that description, I was under the impression that specifically `less` is hardcoded. As I use a different `$PAGER`, I piped up to go fix this via a pr. It took me to look at the source of the function that's called to realize that no, actually this hotkey *does* respect your `$PAGER` variable. \
So, this pr makes the wording in the description of that <kbd>alt+p</kbd> mapping more precise, pointing out that it respects the `$PAGER` variable.

## TODOs:

### DIFF
--- a/doc_src/interactive.rst
+++ b/doc_src/interactive.rst
@@ -354,7 +354,7 @@ Some bindings are common across Emacs and vi mode, because they aren't text edit
 
 - :kbd:`alt-o` opens the file at the cursor in a pager. If the cursor is in command position and the command is a script, it will instead open that script in your editor. The editor is chosen from the first available of the ``$VISUAL`` or ``$EDITOR`` variables.
 
-- :kbd:`alt-p` adds the string ``&| less;`` to the end of the job under the cursor. The result is that the output of the command will be paged.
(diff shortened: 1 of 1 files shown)


doc_src/interactive.rst
```diff
@@ -354,7 +354,7 @@ Some bindings are common across Emacs and vi mode, because they aren't text edit
 
 - :kbd:`alt-o` opens the file at the cursor in a pager. If the cursor is in command position and the command is a script, it will instead open that script in your editor. The editor is chosen from the first available of the ``$VISUAL`` or ``$EDITOR`` variables.
 
-- :kbd:`alt-p` adds the string ``&| less;`` to the end of the job under the cursor. The result is that the output of the command will be paged.
+- :kbd:`alt-p` adds the string ``&| less;`` to the end of the job under the cursor. The result is that the output of the command will be paged. If you set the `$PAGER` environment variable, its value is used instead of `less`.
```
krobelus, line 357: I think you need to double up the backticks for monospace (this is RestructuredText).

Could technically get rid of the word "environment" because that implies it's exported -- which it needn't be, though it probably is in this case. Eh, seems fine as is.

Axlefublr: ohhhh right good call. unfamiliar with this format

PR #11446, opened 2025-05-02
Rephrase set_color tutorial docs

Bergbok:
## Description

Thought the phrasing was a bit off. Not a big deal, but I think my change makes it read a bit smoother.

### DIFF
--- a/doc_src/tutorial.rst
+++ b/doc_src/tutorial.rst
@@ -631,7 +631,7 @@ You can define your own prompt from the command line:
 
 Then, if you are happy with it, you can save it to disk by typing ``funcsave fish_prompt``. This saves the prompt in ``~/.config/fish/functions/fish_prompt.fish``. (Or, if you want, you can create that file manually from the start.)
 
-Multiple lines are OK. Colors can be set via :doc:`set_color <cmds/set_color>` by passing it named ANSI colors or hex/RGB values::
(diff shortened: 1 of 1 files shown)


doc_src/tutorial.rst
```diff
@@ -631,7 +631,7 @@ You can define your own prompt from the command line:
 
 Then, if you are happy with it, you can save it to disk by typing ``funcsave fish_prompt``. This saves the prompt in ``~/.config/fish/functions/fish_prompt.fish``. (Or, if you want, you can create that file manually from the start.)
 
-Multiple lines are OK. Colors can be set via :doc:`set_color <cmds/set_color>`, passing it named ANSI colors, or hex RGB values::
+Multiple lines are OK. Colors can be set via :doc:`set_color <cmds/set_color>` by passing it named ANSI colors or hex/RGB values::
```
krobelus, line 634: the first change makes a lot of sense but I'm not sure about the second one, it might be ambiguous. It's not hex *or* RGB but hexadecimal RGB values.

Bergbok: Ah, I see. 

Changed it but ditched mention of RGB since hex is just a way of representing RGB colors (and searching up "hexadecimal RGB" pretty much only results in color convertors). Could be confusing.

Let me know if that's okay :)

PR #11447, opened 2025-05-02
Use workspace config in packages

danielrainer:
- Apply lint config to entire workspace

- Inherit workspace config for fish-printf

- Allow stdlib printing in fish-printf tests

The current problem which is addressed by this is that warnings about C-String literals are generated by clippy for code in fish-printf. These literals are not available with the current MSRV 1.70, but previously the MSRV setting was not inherited by fish-printf, causing the warning to appear.

# TODOs
- [ ] Do we want to allow stdlib printing in tests? If so, just in the fish-printf tests as currently implemented or for all tests?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -99,7 +99,7 @@ embed-data = ["dep:rust-embed"]
 asan = []
 tsan = []
 
-[lints]
+[workspace.lints]
 rust.non_camel_case_types = "allow"
 rust.non_upper_case_globals = "allow"
 rust.unknown_lints = "allow"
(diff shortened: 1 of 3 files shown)


krobelus:
>  Do we want to allow stdlib printing in tests? If so, just in the fish-printf tests as currently implemented or for all tests?

no strong opinion here. But it probably helps to be consistent with the non-test logic, at least while it's so easy to accidentally print fish's internal encoding.

fish-printf is something that may be useful outside fish so it seems appropriate to use stdlib there. Not sure

Issue #11412, opened 2025-04-22
Fish 4.0.{0,1,2} - regression in printf when using width specifier and non ASCII characters


ZaWertun:
Fish version: 4.0.1
Probably 4.0.2 is affected too, haven't tested it yet.

Demonstration of the bug:
```
$ printf '%10s' 1 |string length
10
```

```
$ printf '%10s' б |string length
9
```

Versions before 4.0.1 works fine as far as I remember, and summary string length would be 10 in both cases.
I saw this behaviour has changed because my prompt function uses printf.

ZaWertun:
It's installed from the official Fedora rpm package (Fedora 42).
Full rpm version: `fish-4.0.1-2.fc42.x86_64`

krobelus:
I can reproduce the missing space `fish --no-config -c 'printf %2s\n б'`
This is because the new printf implementation uses byte count instead of grapheme width (alternatively, codepoint count would be correct in this case).

faho:
Aaaah, that's not a "6" (six)?

Yeah, that explains it.

ZaWertun:
> Aaaah, that's not a "6" (six)?
> 
> Yeah, that explains it.

Ooops, sorry for that. I should have used a different symbol to demonstrate the bug.

danielrainer:
I think this can be closed with #11448 merged.

PR #11448, opened 2025-05-02
Make printf Unicode-aware

danielrainer:
Specifically, the width and precision format specifiers are interpreted as referring to the width of the grapheme clusters rather than the byte count of the string. Note that grapheme clusters can differ in width.

If a precision is specified for a string, meaning its "maximum number of characters", we consider this to limit the width displayed. If there is a grapheme cluster whose width is greater than 1, it might not be possible to get precisely the desired width. In such cases, this last grapheme cluster is excluded from the output.

Note that the definitions used here are not consistent with the `string length` builtin at the moment, but this has already been the case.

Fixes issue #11412

### DIFF
--- a/printf/src/printf_impl.rs
+++ b/printf/src/printf_impl.rs
@@ -5,6 +5,8 @@ use super::locale::Locale;
 use std::fmt::{self, Write};
 use std::mem;
 use std::result::Result;
+use unicode_segmentation::UnicodeSegmentation;
+use unicode_width::UnicodeWidthStr;
 
 #[cfg(feature = "widestring")]
 use widestring::Utf32Str as wstr;
@@ -382,7 +384,7 @@ pub fn sprintf_locale(
         }
 
         // Read field width. We do not support $.
-        let width = if s.at(0) == Some('*') {
+        let desired_width = if s.at(0) == Some('*') {
             let arg_width = args.next().ok_or(Error::MissingArg)?.as_sint()?;
             s.advance_by(1);
             if arg_width < 0 {
@@ -397,7 +399,7 @@ pub fn sprintf_locale(
         };
 
         // Optionally read precision. We do not support $.
-        let mut prec: Option<usize> = if s.at(0) == Some('.') && s.at(1) == Some('*') {
+        let mut desired_precision: Option<usize> = if s.at(0) == Some('.') && s.at(1) == Some('*') {
             // "A negative precision is treated as though it were missing."
             // Here we assume the precision is always signed.
             s.advance_by(2);
@@ -410,7 +412,7 @@ pub fn sprintf_locale(
             None
         };
         // Disallow precisions larger than i32::MAX, in keeping with C.
-        if prec.unwrap_or(0) > i32::MAX as usize {
+        if desired_precision.unwrap_or(0) > i32::MAX as usize {
             return Err(Error::Overflow);
         }
 
@@ -429,7 +431,7 @@ pub fn sprintf_locale(
         // "If a precision is given with a numeric conversion (d, i, o, u, i, x, and X),
         // the 0 flag is ignored." p is included here.
         let spec_is_numeric = matches!(conv_spec, CS::d | CS::u | CS::o | CS::p | CS::x | CS::X);
-        if spec_is_numeric && prec.is_some() {
+        if spec_is_numeric && desired_precision.is_some() {
             flags.zero_pad = false;
         }
 
@@ -443,13 +445,22 @@ pub fn sprintf_locale(
             CS::e | CS::f | CS::g | CS::a | CS::E | CS::F | CS::G | CS::A => {
                 // Floating point types handle output on their own.
                 let float = arg.as_float()?;
-                let len = format_float(f, float, width, prec, flags, locale, conv_spec, buf)?;
+                let len = format_float(
+                    f,
+                    float,
+                    desired_width,
+                    desired_precision,
+                    flags,
+                    locale,
+                    conv_spec,
+                    buf,
+                )?;
                 out_len = out_len.checked_add(len).ok_or(Error::Overflow)?;
                 continue 'main;
             }
             CS::p => {
                 const PTR_HEX_DIGITS: usize = 2 * mem::size_of::<*const u8>();
-                prec = prec.map(|p| p.max(PTR_HEX_DIGITS));
+                desired_precision = desired_precision.map(|p| p.max(PTR_HEX_DIGITS));
                 let uint = arg.as_uint()?;
                 if uint != 0 {
                     prefix = "0x";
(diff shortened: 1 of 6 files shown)


printf/src/printf_impl.rs
```diff
@@ -382,7 +384,7 @@ pub fn sprintf_locale(
         }
 
         // Read field width. We do not support $.
-        let width = if s.at(0) == Some('*') {
+        let desired_display_width = if s.at(0) == Some('*') {
```
krobelus, line 387: would probably have been good to put the renames in a separate commit

I think "width" and "display_width" are equivalent in this context,
so changing this is unrelated from moving from length to width.
(Probably just "width" would be fine, I don't expect that means anything else here?)

printf/src/printf_impl.rs
```diff
@@ -443,13 +445,22 @@ pub fn sprintf_locale(
...
+                    conv_spec,
+                    buf,
+                )?;
                 out_len = out_len.checked_add(len).ok_or(Error::Overflow)?;
                 continue 'main;
             }
             CS::p => {
                 const PTR_HEX_DIGITS: usize = 2 * mem::size_of::<*const u8>();
-                prec = prec.map(|p| p.max(PTR_HEX_DIGITS));
+                desired_precision = desired_precision.map(|p| p.max(PTR_HEX_DIGITS));
```
krobelus, line 463: maybe we can eventually define a `type DisplayWidth(usize);`, to deny implicit conversion..

printf/src/printf_impl.rs
```diff
@@ -514,10 +525,38 @@ pub fn sprintf_locale(
...
-                &s[..p]
+                match desired_precision {
+                    Some(precision) => {
+                        // from man printf(3)
+                        // "the maximum number of characters to be printed from a string"
+                        // We interpret this to mean the maximum width when printed, as defined by
+                        // Unicode grapheme cluster width.
+                        let mut byte_len = 0;
+                        let mut display_width = 0;
+                        let mut graphemes = s.graphemes(true);
```
krobelus, line 537: true means

	/// If `is_extended` is true, the iterator is over the
	/// *extended grapheme clusters*;
	/// otherwise, the iterator is over the *legacy grapheme clusters*.
	/// [UAX#29](http://www.unicode.org/reports/tr29/#Grapheme_Cluster_Boundaries)
	/// recommends extended grapheme cluster boundaries for general processing.

I wonder if that's what we did in fish 3.

Unrelated to that, I noticed that `printf %.1s "🇺🇳"` used to
(incorrectly) print the first codepoint in fish 3.

printf/src/printf_impl.rs
```diff
@@ -528,39 +567,43 @@ pub fn sprintf_locale(
...
         // Pad after the prefix with zeros to the desired width?
         if !flags.left_adj && flags.zero_pad {
-            pad(f, '0', width, unpadded_width)?;
+            pad(f, '0', display_width, unpadded_display_width)?;
         }
 
         // Pad on the left to the given precision?
-        pad(f, '0', prec, body_len)?;
+        // TODO: why pad with 0 here?
+        pad(f, '0', desired_precision, body_display_width)?;
```
krobelus, line 606: good question, maybe this is dead code

printf/src/tests.rs
```diff
@@ -13,14 +13,15 @@ macro_rules! sprintf_check {
...
+            use unicode_width::UnicodeWidthStr;
             let mut target = String::new();
             let mut args = [$($arg.to_arg()),*];
             let len = $crate::printf_c_locale(
                 &mut target,
                 $fmt.as_ref() as &str,
                 &mut args,
             ).expect("printf failed");
-            assert!(len == target.len(), "Wrong length returned: {} vs {}", len, target.len());
+            assert!(len == target.width(), "Wrong length returned: {} vs {}", len, target.len());
```
krobelus, line 24: (unrelated: I guess should should have been an `assert_eq`, then we don't need to repeat the values)

krobelus (approved):

printf/src/printf_impl.rs
danielrainer: Yes, some of the renaming is mostly an artifact of me using more explicit naming to help me with figuring out how variables are used. Should I change it back to just `width`?

krobelus:
> Yes, some of the renaming is mostly an artifact of me using more
> explicit naming to help me with figuring out how variables are
> used. Should I change it back to just `width`?

I think either is fine, as long as its consistent within this module.

krobelus:
I haven't looked at much of the surrounding code
but "width" would be consistent with ".width()" and ".len()"
which are 3rd-party and stlib APIs we use.

printf/src/printf_impl.rs
danielrainer: I did not check with fish 3 and just went with the recommended variant as the default. I don't have strong opinions here.

printf/src/printf_impl.rs
krobelus: sorry I meant to comment on `desired_precision = Some(buf.len() + 1);`
Probably not super important now.

danielrainer: Ah yes, I only use width in the case of user-provided strings, because I thought that the rest should be ASCII, but perhaps that should be reworked as well, e.g. thousands separators (or whichever other quantity) might not be ASCII in some locales.

printf/src/printf_impl.rs
danielrainer: Removing it actually breaks some tests, so it seems to have some purpose.

printf/src/tests.rs
danielrainer: done

Berrysoft:
Just a question: I have found that fish itself uses it own methods to determine widths of Unicode chars, and some kinds of them could be configured by users. But `printf` is using `unicode-width` here. Is it intended behavior?

krobelus:
On Wed, May 07, 2025 at 11:55:52PM -0700, Yuyi Wang wrote:
> Berrysoft left a comment (fish-shell/fish-shell#11448)
> 
> Just a question: I have found that fish itself uses it own methods
> to determine widths of Unicode chars, and some kinds of them could
> be configured by users. But `printf` is using `unicode-width`
> here. Is it intended behavior?

I don't know if the original behavior was ever intended, but this
PR mainly restores that (historical) behavior.  ("Note that the
definitions used here are not consistent with the `string length`
builtin at the moment, but this has already been the case.")

AFAICT, the differences are
1. we use `widecharwidth` everywhere else (which respects fish_ambiguous_width and fish_emoji_width)
2. "string length --visible" knows about terminal control sequences (mostly treats them as width 0)

In future, we should probably use the same unicode database for everything.
Not sure which one.

Kitty's new text-sizing protocol would allow fish to pass an explicit width,
obsoleting the need for setting fish_ambiguous_width and fish_emoji_width
to work around cases where the terminal disagrees on text width.

Issue #11193, opened 2025-02-28
Make complete-and-search command move forward in completion pager


azuenko:
## Description
At the moment the `complete-and-search` command is used both to invoke the searchable pager and to move backwards, once pager is invoked. 

Is not quite flexible when one wants to replace the default pager with the searchable one. A simple remap
```
bind -M insert tab complete-and-search
```
works, but hitting `Tab` multiple times enumerates the list in the reverse order. 

The doc [says](https://github.com/fish-shell/fish-shell/blob/fa390f1b14f4ef1e7f8712e24ad69840d8431e11/doc_src/cmds/bind.rst?plain=1#L187C5-L187C123)
> ``complete-and-search``
> invoke the searchable pager on completion options (for convenience, this also moves backwards in the completion pager) 

However the mapping becomes inconvenient for those who want to use searchable pager by default.

## Proposal

The [code](https://github.com/fish-shell/fish-shell/blob/fa390f1b14f4ef1e7f8712e24ad69840d8431e11/src/reader.rs#L2749-L2776)
```rust
            rl::Complete | rl::CompleteAndSearch => {
                if !self.conf.complete_ok {
                    return;
                }
                if self.is_navigating_pager_contents()
                    || (!self.rls().comp.is_empty()
                        && !self.rls().complete_did_insert
                        && self.rls().last_cmd == Some(rl::Complete))
                {
                    // The user typed complete more than once in a row. If we are not yet fully
                    // disclosed, then become so; otherwise cycle through our available completions.
                    if self.current_page_rendering.remaining_to_disclose != 0 {
                        self.pager.set_fully_disclosed();
                    } else {
                        self.select_completion_in_direction(
                            if c == rl::Complete {
                                SelectionMotion::Next
                            } else {
                                SelectionMotion::Prev
                            },
                            false,
                        );
                    }
                } else {
                    // Either the user hit tab only once, or we had no visible completion list.
                    self.compute_and_apply_completions(c);
                }
            }
```
unconditionally maps `rl::Complete` to `SelectionMotion::Next`. It would be more flexible if the behaviour can be altered and `rl::Complete` is instead mapped to `SelectionMotion::Prev`, while `rl::CompleteAndSearch` to `SelectionMotion::Prev`.

krobelus:
> Is not quite flexible when one wants to replace the default pager with the searchable one. A simple remap
> ```
> bind -M insert tab complete-and-search
> ```
[...]

PR #11450, opened 2025-05-03
feat: Add `pager-show-search` readline command

balthild:
## Description

Adds a readline command `pager-show-search` that shows the search when a pager is visible.

Differences with existing commands:

- `complete-and-search`: it does not select the last item in pager.
- `pager-toggle-search`: it always shows rather than toggling show/hide; it does not select the next item if search is already shown.

Primarily intended to be called after `complete`.

Fixes issue #11449.

## TODOs:

### DIFF
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -2787,6 +2787,24 @@ impl<'a> Reader<'a> {
                     self.compute_and_apply_completions(c);
                 }
             }
+            rl::PagerShowSearch => {
+                if let Some(history_pager) = &self.history_pager {
+                    if history_pager.start == 0 {
+                        self.flash(0..self.command_line.len())
+                    }
+                    return;
+                }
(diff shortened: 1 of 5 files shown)


src/reader.rs
```diff
@@ -2787,6 +2787,24 @@ impl<'a> Reader<'a> {
                     self.compute_and_apply_completions(c);
                 }
             }
+            rl::PagerShowSearch => {
```
krobelus, line 2790: yeah some of our special readline functions are inflexible DWIM-style commands.
For this case, does the approach from https://github.com/fish-shell/fish-shell/issues/11193#issuecomment-2692821049
work for you? I guess that one is a bit backwards and unintuitive.

balthild: Yes, it works!

And since there's already a workaround, I think an alternative to this PR could be adding it to the documentation?

PR #11453, opened 2025-05-03
completion(winetricks): redirect winetricks list-all sderr to /dev/null

fabiojb:
## Description

The command `winetricks list-all` used to generate completions is kind of slow, and hitting Ctrl + C while fish is calling it generates an excessive verbose output that comes from winetricks.

Fixes this by redirection stderr to /dev/null.

## TODOs:

### DIFF
--- a/share/completions/winetricks.fish
+++ b/share/completions/winetricks.fish
@@ -1,5 +1,5 @@
 function __fish_winetricks__complete_verbs
-    winetricks list-all |
+    winetricks list-all 2>/dev/null |
         string match --invert --regex '^==' |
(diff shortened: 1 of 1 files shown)


fabiojb:
Also, I was wondering if saving `winetricks list-all` output to be used on latter invocations could be a good solution to improve the speed of completion options in this case. Maybe persist the output in a fish global variable.

But this would add extra complexity, and I'm not really sure if it does pay off in this case.

krobelus:
We have some completions that use caching, look for usages of `__fish_cache_put`.
If you work on that, it would be good to start by extracting a function as all callers have very similar logic (checking `path mtime` etc)

PR #11462, opened 2025-05-06
completions/cargo: strip ansi color codes

bvergnaud:
## Description

On my current setup, cargo completions contain ANSI colour codes. The very first `cargo --list` command outputs coloured text, and nothing ever cleans it up.

I'm not sure if it's a quirk on my setup specifically. I can't find any environment variable or config file that enforces coloured output even though the stdout/stderr of that command is sent to a pipe.

Regardless, adding that `--color never` flag ensures that we work on colourless text, and that the completion is what we expect.

## TODOs:

### DIFF
--- a/share/completions/cargo.fish
+++ b/share/completions/cargo.fish
@@ -1,20 +1,25 @@
 # Tab completion for cargo (https://github.com/rust-lang/cargo).
 
 ## --- WRITTEN MANUALLY ---
-set -l __fish_cargo_subcommands (cargo --list 2>&1 | string replace -rf '^\s+([^\s]+)\s*(.*)' '$1\t$2' | string escape)
+
+function __fish_cargo
+    cargo --color=never $argv
+end
+
+set -l __fish_cargo_subcommands (__fish_cargo --list 2>&1 | string replace -rf '^\s+([^\s]+)\s*(.*)' '$1\t$2' | string escape)
 
 complete -c cargo -f -c cargo -n __fish_use_subcommand -a "$__fish_cargo_subcommands"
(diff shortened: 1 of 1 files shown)


share/completions/cargo.fish
```diff
@@ -1,7 +1,7 @@
 # Tab completion for cargo (https://github.com/rust-lang/cargo).
 
 ## --- WRITTEN MANUALLY ---
-set -l __fish_cargo_subcommands (cargo --list 2>&1 | string replace -rf '^\s+([^\s]+)\s*(.*)' '$1\t$2' | string escape)
+set -l __fish_cargo_subcommands (cargo --color never --list 2>&1 | string replace -rf '^\s+([^\s]+)\s*(.*)' '$1\t$2' | string escape)
```
krobelus, line 4: Are you maybe setting `CARGO_TERM_COLOR=always`?
E.g.

	CARGO_TERM_COLOR=always cargo --list | head

Else can you run

	strace cargo --list 2>&1 | grep 'openat.*\bconfig\b' | grep -v ENOENT

and check if `color` is set in any of the listed paths.

I guess it would be okay for fish to work around this but
I'd rather see change that addresses all instances of this
problem in share/completions/cargo.fish.

For example we could define a wrapper like

```fish
function __fish_cargo
	cargo --config term.color=auto $argv
end
```

though color=always doesn't seem like a good default.
A lot of other software will be broken if color=always is set, it's not feasible to add workarounds everywhere.
I guess this could happen due to lazy configuration on a CI system, if that's a realistic case we can definitely work around

Issue #11400, opened 2025-04-18
Translations could be improved in several ways


danielrainer:
Investigating the `sprintf!` macro got me down into the rabbit hole of investigating how translations (i18n/l10n) are done in fish.

One issue seems to be that Rust support was not available in gettext when the related code was written. This seems to have changed recently: https://www.gnu.org/software/gettext/manual/html_node/Rust.html

It seems that it might be possible to replace the printf-like format strings by versions more suitable for Rust.

To investigate the current state, it took me quite a while to figure out how the translation pipeline works. For reference, this is the approach I used to generate up-to-date translations and make them available to fish:

1. `build_tools/fish_xgettext.fish` This updates/generates `messages.pot` in the repo root directory.
2. `msgmerge --update --no-fuzzy-matching --no-wrap po/de.po messages.pot` updates an existing po file (in this case the German one) with new line numbers.
3. `msgfmt --check-format po/de.po` generates the binary `messages.mo` file in the repo root. This file can be used by a fish instance to provide translations (for German in this case), but it is not at the right place.
4. `mkdir --parents share/locale/de/LC_MESSAGES/` creates the directory where fish will look for the mo file. (At least when build using `cargo build` directly. I don't know how the search procedure works in general.) Replace `de` by the corresponding language.
5. `cp messages.mo share/locale/de/LC_MESSAGES/fish.mo` At this point, which can finally make use of the mo file.
6. `target/debug/fish` to launch fish
7. `set -lx LANG de_DE.utf8` to tell fish to use German
8. At this point, strings for which a German translation is available should be displayed in German.

Having the steps for getting translations up to date documented in `CONTRIBUTING.rst` would be great. Some parts of it are there already, but it took me quite a while to figure out the rest.

Ideally, we could have some automation to ensure that translations are kept up to date.

I when checking the generated mo files, I noticed that there are some issues with the generation.

- In some cases, `messages.pot` and `po/de.po` contain sensible data for a string, but running `msgmerge` inserts garbage. I have not looked into this issue much.
- For fish scripts, the generated line numbers are incorrect. I tried to address this in #11399, but the code there is not exactly pretty and there might be issues I am unaware of with the proposed approach.

This issue should probably be split up into several smaller ones, but I'd like to get some feedback first from someone more familiar with this.

Also, do people actually care about having fish available in languages other than English? The state of the translations does not seem like there is much interest in them.

zanchey:
[...]

PR #11463, opened 2025-05-06
Automate translation updates

danielrainer:
The purpose of this script is to simplify the translation-related workflow for
both developers and translators. It runs the xgettext, msgmerge, msgfmt pipeline
(or only parts of it, depending on the arguments), either for all languages, or
for one specific one.

Developers can use the script with the `--no-mo` flag to update the PO files for
all languages after changes to the Rust/fish sources, to keep the translations
up to date. Ideally, this would run automatically for all changes, such that
translations are always up to date, but for now, it would already be an
improvement to run this script before releasing a new version of fish.

Translators can use the script in the same way as developers, to get up to date
PO files. To see their translations in action, the script can be called with
`--only-mo`, which takes the current version of the PO files and generates MO
files from them, which get placed in a location which fish (built with `cargo
build`) can detect. Translators might also find it useful to use `--lang=cc`
(where cc is a language identifier) to only update the files for a single
language.

## TODOs:
- [ ] Documentation updates
- [ ] Do we want to add some proper automation, which runs the script for all changes, such that translations are kept up to date? If not, it might be useful to at least run it before releasing new versions of fish.

### DIFF
--- a/build_tools/update_translations.fish
+++ b/build_tools/update_translations.fish
@@ -1,0 +1,87 @@
+#!/usr/bin/env fish
+
+# Updates the files used for gettext translations.
+# By default, the whole xgettext, msgmerge, msgfmt pipeline runs,
+# which extracts the messages from the source files into $template_file,
+# updates the PO files for each language from that
+# (changed line numbers, added messages, removed messages),
+# and finally generates a machine-readable MO file for each language,
+# which is stored in share/locale/$LANG/LC_MESSAGES/fish.mo (relative to the repo root).
+#
+# Use cases:
+# For developers:
+#   - Run with args `--no-mo` to update all PO files after making changes to Rust/fish
+#     sources.
+# For translators:
+#   - Run with `--no-mo` first, to ensure that the strings you are translating are up to date.
+#   - Specify the language you want to work on as an argument, which must be a file in the po/
+#     directory. You can specify a language which does not have translations yet by specifying the
+#     name of a file which does not yet exist. Make sure to follow the naming convention.
+
+# The sort utility is locale-sensitive.
+# Ensure that sorting output is consistent by setting LC_ALL here.
+set -gx LC_ALL C.UTF-8
+
+set -l template_file messages.pot
+set -l build_tools (status dirname)
+set -l po_dir $build_tools/../po
+
+set -l extract
+set -l po
+set -l mo
+
+argparse --exclusive 'no-mo,only-mo' 'no-mo' 'only-mo' -- $argv
+or exit $status
+
+if test -z $argv[1]
+    # Update everything if not specified otherwise.
+    set -g po_files $po_dir/*.po
+else
+    set -l po_dir_id (stat --format='%d:%i' -- $po_dir)
+    for arg in $argv
+        set -l arg_dir_id (stat --format='%d:%i' -- (dirname $arg))
+        if test $po_dir_id != $arg_dir_id
+            echo "Argument $arg is not a file in the directory $(realpath $po_dir)."
+            echo "Non-option arguments must specify paths to files in this directory."
+            echo ""
+            echo "If you want to add a new language to the translations not the following:"
+            echo "The filename must identify a language, with a two letter ISO 639-1 language code of the target language (e.g. 'pt' for Portuguese), and use the file extension '.po'."
+            echo "Optionally, you can specify a regional variant (e.g. 'pt_BR')."
+            echo "So valid filenames are of the shape 'll.po' or 'll_CC.po'."
+            exit 1
+        end
+        if not basename $arg | grep -qE '^[a-z]{2}(_[A-Z]{2})?\.po$'
+            echo "Filename does not match the expected format ('ll.po' or 'll_CC.po')."
+            exit 1
+        end
+    end
+    set -g po_files $argv
+end
+
+if set -l --query _flag_no_mo
+    set -l --erase mo
+end
+if set -l --query _flag_only_mo
+    set -l --erase extract
+    set -l --erase po
+end
+
+if set -l --query extract
+    $build_tools/fish_xgettext.fish
+end
+
+for po_file in $po_files
+    if set -l --query po
+        if test -e $po_file
+            msgmerge --update --no-fuzzy-matching --no-wrap --backup=none $po_file $template_file
+        else
+            cp $template_file $po_file
+        end
+    end
+    if set -l --query mo
+        set -l locale_dir $build_tools/../share/locale
+        set -l out_dir $locale_dir/(basename $po_file .po)/LC_MESSAGES
+        mkdir -p $out_dir
+        msgfmt --check-format --output-file=$out_dir/fish.mo $po_file
+    end
+end
--- a/.gitignore
+++ b/.gitignore
@@ -39,6 +39,7 @@ Thumbs.db
 ehthumbs.db
 
 messages.pot
+*.mo
 .directory
 .fuse_hidden*
 


build_tools/update_translations.fish
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env fish
+
+# Updates the files used for gettext translations.
+# By default, the whole xgettext, msgmerge, msgfmt pipeline runs,
```
krobelus, line 4: great, so this script will do everything.

I don't know much about translations; is there still a need to run `build_tools/fish_xgettext.fish` manually?
Else we could simplify the interface by combining both in one script.

 > Do we want to add some proper automation, which runs the script
 > for all changes, such that translations are kept up to date? If
 > not, it might be useful to at least run it before releasing new
 > versions of fish.

I think the Git project generally ignores translations,
and dedicated translators take care of updating everything once per release cycle.

I guess we should try updating at every commit, and checking it in the CI etc.
Not sure if we want the diff noise though.
Maybe we should use custom diff tools and/or ignore changes in po/.
Maybe even move to a different file format that has stable IDs, but I guess we should match people's expectations.
If we always update the po files, then translaters needn't know about anything else.

build_tools/update_translations.fish
```diff
@@ -0,0 +1,74 @@
...
+# and finally generates a machine-readable MO file for each language,
+# which is stored in share/locale/$LANG/LC_MESSAGES/fish.mo (relative to the repo root).
+#
+# Use cases:
+# For developers:
+#   - Run with args `--no-mo` to update all PO files after making changes to Rust/fish
+# sources.
+# For translators:
+#   - Run with `--no-mo` first, to ensure that the strings you are translating are up to date.
+#     Optionally, you can specify the language via `--lang=LANG`, where lang is
```
krobelus, line 17: unfinished sentence

build_tools/update_translations.fish
```diff
@@ -0,0 +1,74 @@
...
+# no-extract, no-po, mo (for translators who just updated a po file)
+# extract, po, no-mo (for getting ready to translate, for updating git after source changes)
+# single lang?
```
krobelus, line 20: maybe leftover?

build_tools/update_translations.fish
```diff
@@ -0,0 +1,74 @@
...
+    end
+end
+
+if set --query extract
+    set extraction_script $build_script_dir/fish_xgettext.fish
+    $extraction_script
+end
+
+for lang_po_file in $langs
+    if set --query po
```
krobelus, line 65: right, we should avoid "is this variable defined at all" since there can be false positives with universal variables.
Better always define it in our scope and use `set -q po[1]`, or set it to true/false or 1/0

build_tools/update_translations.fish
```diff
@@ -0,0 +1,74 @@
...
+        msgmerge --update --no-fuzzy-matching --no-wrap --backup=none $lang_po_file messages.pot
+    end
+    if set --query mo
+        set LOCALE_DIR $build_script_dir/../share/locale
+        set OUT_DIR $LOCALE_DIR/(basename $lang_po_file .po)/LC_MESSAGES
+        mkdir --parents $OUT_DIR
+        msgfmt --check-format --output-file=$OUT_DIR/fish.mo $lang_po_file
+    end
+end
```
krobelus, line 74: I ran this and get a diff like

```diff
-msgid "$%ls: originally inherited as |%ls|\n"
-msgstr "$%ls: Ursprünglich geerbt als |%ls|\n"
+#~ msgid "$%ls: originally inherited as |%ls|\n"
+#~ msgstr "$%ls: Ursprünglich geerbt als |%ls|\n"
```

Apparently "#~" means obsolete, which doesn t' seem true.
Is the detection wrong?

build_tools/update_translations.fish
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env fish
+
+# Updates the files used for gettext translations.
+# By default, the whole xgettext, msgmerge, msgfmt pipeline runs,
+# which extracts the messages from the source files into messages.pot,
+# updates the PO files for each language from that
+# (changed line numbers, added messages, removed messages),
+# and finally generates a machine-readable MO file for each language,
+# which is stored in share/locale/$LANG/LC_MESSAGES/fish.mo (relative to the repo root).
```
krobelus, line 9: cool, this works for me:

	$ build_tools/update_translations.fish
	$ LANG=de_DE target/debug/fish -c 'alias ""'
	alias: Name darf nicht leer sein

though I wonder why the same doesn't work for fish installed via `cargo install --path .`

build_tools/fish_xgettext.fish
```diff
@@ -79,4 +79,4 @@ xgettext -j -k -kN_ -LShell --from-code=UTF-8 -cDescription --no-wrap -o message
 # Remove the tmpdir from the location to avoid churn
 sed -i 's_^#: /.*/share/_#: share/_' messages.pot
 
-rm -r $tmpdir
+rm -rf $tmpdir
```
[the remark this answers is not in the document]
danielrainer: Good point, I fixed my config and removed the change.

build_tools/update_translations.fish
danielrainer: fixed

build_tools/update_translations.fish
danielrainer: Yes, after putting docs into the error message and commit message, I forgot these comments. Deleted.

build_tools/update_translations.fish
danielrainer: That's odd, for this particular message I get
```diff
-#: src/builtins/set.rs:696
+#: src/builtins/set.rs:706
 msgid "$%ls: originally inherited as |%ls|\n"
 msgstr "$%ls: Ursprünglich geerbt als |%ls|\n"
```
which matches the source file.

`#~` is indeed used for obsolete messages. Do you get this when running the script with a clean (in the sense of no uncommited changes) `po/` directory?

build_tools/update_translations.fish
danielrainer: > is there still a need to run build_tools/fish_xgettext.fish manually?

Not really. At the moment the `update_translations.fish` script does not allow just running the xgettext step. The only reason I can think of why one might want to run xgettext not followed by msgmerge is for starting translations for a new language. In such a case it wouldn't hurt much to update the other PO files, but it might still be nice to have special handling for that.

> Else we could simplify the interface by combining both in one script.

Yes, I think this is a sensible idea regardless of whether we want to support special handling for new languages. I mainly waited with that so that other changes such as #11466 can be addressed separately first.

> I think the Git project generally ignores translations

This seems accurate.

> dedicated translators take care of updating everything once per release cycle.

The updating once per release cycle part seems like it's not really happening. With line-based locations, translations break a lot, and the last update to any PO file was in 2024, meaning that many translations would have broken since (and there have been several releases since), due to line number changes. I guess for fish scripts the line number information was generated completely incorrectly until recent changes anyway, so updating these would not have been very helpful.

> I guess we should try updating at every commit, and checking it in the CI etc.
Not sure if we want the diff noise though.

That would be ideal. I worried about the diff noise as well. I have since realized that line numbers can be omitted, which greatly reduces the diff noise, so we only get any if the actual strings change.

> Maybe we should use custom diff tools and/or ignore changes in po/.

Might be useful. Git has options to exclude files/directories from `diff` or `log --patch` output.
https://stackoverflow.com/a/29374503
https://stackoverflow.com/a/25340187

> Maybe even move to a different file format that has stable IDs, but I guess we should match people's expectations.

Stable IDs would help, but it's not simple, especially with formatting. I had a look at gettext alternatives we could use, and the only one I found which seems to have decent Rust support is fluent. (See https://github.com/fish-shell/fish-shell/issues/11400#issuecomment-2831714671 for details.)

> If we always update the po files, then translators needn't know about anything else.

That would be ideal. I have written CI integration for checking if re-running the `update_translations.fish` script causes any changes. At least for PO files with line numbers present, it causes a diff on PO files which I had just updated before, which makes it unsuitable for detecting issues without annoying false positives. So far, I have not figured out the reason. Maybe different gettext versions? I have not tested it without line numbers in the PO files, maybe that would help.

build_tools/update_translations.fish
```diff
@@ -0,0 +1,74 @@
...
+# no-extract, no-po, mo (for translators who just updated a po file)
+# extract, po, no-mo (for getting ready to translate, for updating git after source changes)
+# single lang?
+
+set build_script_dir (status dirname)
+
+set extract
+set po
+set mo
+set langs $build_script_dir/../po/*.po
```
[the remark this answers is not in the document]
krobelus: > I added explicit scopes to all `set` commands.

Our usual approach is to define functions in local scope once.
Then all subsequent `set` commands in or below the current block will use that scope.

```fish
if set --local --query _flag_lang
	set --local language_found false 
	set --local langs 
	for ...
		set language_found true
		set langs $lang_po_file
	end
	if not $language_found
		use $langs ...
	end
end
```

`--function` seems more complicated so I'd use it only when necessary.
Note that "function" is not a real scope, try

	function foo
	    set -f foo bar
	    set -S foo
	end
	foo

build_tools/update_translations.fish
krobelus: My bad, my version of `cargo-expand` was too old

If it's easy, we should pin that as `dev-dependency` in Cargo.toml/Cargo.lock. Not sure if that works for binary crates.. but we could probably use the library that powers cargo-expand.
It might not be super important. But it would be great if we move to a state where everything is version-pinned, so we can in theory have reproducible builds and easy bisection etc.

build_tools/update_translations.fish
krobelus: yeah that works too.

> Would setting the variables to some specific value help with collisions

There's no inherent need to set a non-empty value. It still shadows any inherited globals/universals

build_tools/update_translations.fish
krobelus: Moving (but not changing) a translation string from one Rust file to another one will cause noisy changes.
Which begs the question - do we need the filename at all? [`Git doesn't use it`](https://github.com/git/git/blob/6f84262c44a89851c3ae5a6e4c1a9d06b2068d75/po/de.po).
It should be almost always trivial to find the source of the translation string, so this is basically redundant information.

Overall, I'd like to update translation strings eagerly (we should either support translations properly or not at all).
Can we make sure it's run as part of `ninja -Cbuild test`, so we're unlikely to forget?
The test subtarget (`ninja -C build test-translations-update`) should probably not touch files, only fail if anything changed.

krobelus: > At least for PO files with line numbers present, it causes a diff on PO files which I had just updated before, which makes it unsuitable for detecting issues without annoying false positives

would be great to figure this out. The message catalogue (*.po) file format is pretty simple, we should be able to make it deterministic.
As for doc updates, if the failing test tells us about the fix, that should be enough.

We did have some trouble in the past because PRs by translators came with unrelated whitespace changes.
That should hopefully be a thing of the past, after we keep it up-to-date ourselves

zanchey: > Which begs the question - do we need the filename at all?

Yeah, I think it's fine to remove. The gettext manual says "this ... makes it harder for technically skilled translators to understand each message’s context" but I don't think running `rg` over the source is a particularly high bar to clear.

My inclination is not to enforce PO file updates with a CI check but it does help to reduce string churn, which is something that I think is a bit demotivating for translators. A daily/weekly GitHub action might be better.

ridiculousfish:
Squash-merged as 035cd369c2670c973832081e00eac8e5c384bce0

Thank you!

build_tools/update_translations.fish
danielrainer: I agree that version pinning would be good. With `cargo-expand`, I ran into the issue that recent versions don't work with Rust 1.70. The most recent version I found to be compatible with Rust 1.70 is `cargo-expand` 1.0.79, which is from early 2024. In CI, we could easily use more recent versions by just including a newer Rust toolchain, but if we want it as a `dev-dependency` we'd probably have to stick to versions compatible with our MSRV.

build_tools/update_translations.fish
danielrainer: I opened #11469 to get rid of source locations in PO files.

build_tools/update_translations.fish
danielrainer: > I wonder why the same doesn't work for fish installed via `cargo install --path . --debug`

I figured out the difference. It's in `src/env/config_paths.rs`, where depending on the installation context, the lookup path for the MO files is set differently. Check #11470 and see the difference between the installation methods by running `fish --debug=config`. If installed via `cargo install`, the locale path will not be set, which causes `src/wutil/gettext.rs:wgettext_really_init()` to return without doing anything.

build_tools/update_translations.fish
```diff
@@ -0,0 +1,72 @@
...
+#   - Run with `--no-mo` first, to ensure that the strings you are translating are up to date.
+#     Optionally, you can specify the language via `--lang=LANG`, where LANG is the
+#     ISO 639-1 language code of the target language (eg de for German).
+
+set build_script_dir (status dirname)
+
+set --local extract
+set --local po
+set --local mo
+set --local langs $build_script_dir/../po/*.po
```
krobelus, line 25: maybe `po_files` would be a better name (but only if made consistent with loop variables etc)

build_tools/update_translations.fish
```diff
@@ -0,0 +1,72 @@
...
+if set --local --query _flag_only_mo
+    set --local --erase extract
+    set --local --erase po
+end
+if set --local --query _flag_lang
+    for lang_po_file in $langs
+        if test (basename $lang_po_file .po) = $_flag_lang
+            set --function language_found
+            set --function langs $lang_po_file
+            break
```
krobelus, line 41: this loop seems a bit backwards. We already know the exact filename we're looking for.
Why not ask for this directly via

```fish
if not test -e $build_script_dir/../po/$_flag_lang.po
```

also, we don't handle repeated --lang= arguments.
We should at least have an explicit error if we don't support this case yet.

Also, why is `lang` an option-argument at all?
Can't we use positional arguments for this (i.e. `$argv` as left by argparse).

Going further, maybe we should even change the interface to
pass the full file paths, like

	build_tools/update_translations.fish po/de.po 

That will work more naturally with shell completions and globbing.
In future, this should also allow creating new translations with the same exact command.

build_tools/update_translations.fish
```diff
@@ -0,0 +1,72 @@
...
+        end
+    end
+    if not set --function --query language_found
```
krobelus, line 44: I think `--function` is needlessly surprising here.
Sure, it will be useful once we package this logic in a function,
but for now it's a roundabout way of saying `--global`, so why not say that.

build_tools/update_translations.fish
```diff
@@ -0,0 +1,72 @@
...
+        echo "Language '$_flag_lang' not found. If you want to update translations for an existing language, use one of:"
+        for lang_po_file in $langs
+            printf ' %s' (basename $lang_po_file .po)
+        end
+        printf '\n\n'
+        echo "If you want to add translations for a language which does not have any translations yet, navigate to the repository root in your shell and run:"
```
krobelus, line 50: if we keep this, maybe shorten it ("To add a new translation") -- I think singular is fine

build_tools/update_translations.fish
```diff
@@ -0,0 +1,72 @@
...
+        echo " build_tools/fish_xgettext.fish && cp messages.pot po/cc.po"
+        echo "where 'cc' is the two letter ISO 639-1 language code of the target language (eg de for German)."
```
krobelus, line 52: nit: maybe `(e.g. 'de' for German)`

build_tools/update_translations.fish
```diff
@@ -0,0 +1,72 @@
...
+        exit 1
+    end
+end
+
+if set --local --query extract
+    set --local extraction_script $build_script_dir/fish_xgettext.fish
+    $extraction_script
```
krobelus, line 59: this variable is only used once, better inline it, no?

Maybe it's because of the odd naming:
one flaw in 3e29793d046 (improve detection of msgs to be translated, 2017-05-27)
seems to be that it called the script "fish_xgettext.fish", to role-play as "xgettext",
but fish_xgettext.fish hardcodes the "messages.pot" output path,
maybe that makes the call site is more confusing.

build_tools/update_translations.fish
```diff
@@ -0,0 +1,72 @@
...
+end
+
+for lang_po_file in $langs
+    if set --local --query po
+        msgmerge --update --no-fuzzy-matching --no-wrap --backup=none $lang_po_file messages.pot
```
krobelus, line 64: unrelated: I'm questioning why "messages.pot" is at top-level and
not inside "po/".. or "target/", since this already depends on cargo builds I think. I wonder what's the convention.

I don't expect there is a reason for us to look at this file, at
least once `build_tools/update_translations.fish po/foo.po` transparently
creates that file if it doesn't exist.
Though maybe there are translation tools that interact with "messages.pot".
Not sure, maybe check the most popular one.

build_tools/update_translations.fish
```diff
@@ -0,0 +1,72 @@
...
+    end
+    if set --local --query mo
+        set --local LOCALE_DIR $build_script_dir/../share/locale
+        set --local OUT_DIR $LOCALE_DIR/(basename $lang_po_file .po)/LC_MESSAGES
```
krobelus, line 68: these SCREAMING_SNAKE_CASE names are surprising to me because the
variables are not required to be named like this, and they are not
exported. So I'd use lowercase like elsewhere.

build_tools/update_translations.fish
```diff
@@ -0,0 +1,72 @@
...
+        mkdir --parents $OUT_DIR
```
krobelus, line 69: long options is definitely the right default for scripts.
For this I'd still use "mkdir -p", probably because it's well-known and the long option isn't super helpful at explaining everything

build_tools/update_translations.fish
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env fish
```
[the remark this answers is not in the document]
danielrainer: Yes, normally I'd use POSIX shell scripts for such things, but I thought for fish I might as well use a fish script, similar to the `fish_xgettext.fish` script. On the other hand, since this is somewhat of a build-dependency of fish, it isn't ideal to have a reflexive dependency on fish.

I wouldn't use dash, since it is not universally available. For a truly portable shell script I'd use POSIX shell, rather than bash. I think this script could quite easily be ported to POSIX shell (only the argument parsing would be more annoying).
I don't think that python would make much sense for this simple script.
Rust might be useful for integrating proper parsing for the xgettext step, but until we do that, it seems a bit overkill.

I think it is desirable to be able to build fish without already having some version of it installed, so it might be good to not use fish scripts here (at least not for generating the mo files). For now, mo files are generated via cmake, but I think the plan (and my preference) is to move away from that, so it would be good to have alternatives ready.

With this in mind, I think it might make sense to at least extract a separate POSIX shell script for the `msgfmt` (po -> mo) step, or to port the entire script to POSIX. (The `fish_gettext.fish` script might not have to be ported, since the po files should be updated on the machines of developers or translators, who probably have fish installed. In CI, checking whether the po files are up-to-date can be done after building fish.)

What do you think makes most sense?

build_tools/update_translations.fish
```diff
@@ -0,0 +1,72 @@
...
+# Use cases:
+# For developers:
+#   - Run with args `--no-mo` to update all PO files after making changes to Rust/fish
+#     sources.
+# For translators:
+#   - Run with `--no-mo` first, to ensure that the strings you are translating are up to date.
+#     Optionally, you can specify the language via `--lang=LANG`, where LANG is the
+#     ISO 639-1 language code of the target language (eg de for German).
+
+set build_script_dir (status dirname)
```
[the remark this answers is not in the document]
danielrainer: Yes, changed to `build_tools`.

build_tools/update_translations.fish
```diff
@@ -0,0 +1,72 @@
...
+
+set --local extract
```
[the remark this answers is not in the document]
danielrainer: I normally prefer long options, but I guess for something this common short options are fine. Changed.

build_tools/update_translations.fish
danielrainer: Agreed, changed.

build_tools/update_translations.fish
```diff
@@ -0,0 +1,72 @@
...
+set --local po
+set --local mo
+set --local langs $build_script_dir/../po/*.po
+
+argparse --exclusive 'no-mo,only-mo' 'no-mo' 'only-mo' 'lang=' -- $argv
```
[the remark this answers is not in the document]
danielrainer: done

build_tools/update_translations.fish
danielrainer: Variable is obsolete. Deleted.

build_tools/update_translations.fish
danielrainer: > this variable is only used once, better inline it, no?

Yes, inlined.

> Maybe it's because of the odd naming:
one flaw in https://github.com/fish-shell/fish-shell/commit/3e29793d046ad3e5a79dfe1ec75e9df3f2d87090 (improve detection of msgs to be translated, 2017-05-27)
seems to be that it called the script "fish_xgettext.fish", to role-play as "xgettext",
but fish_xgettext.fish hardcodes the "messages.pot" output path,
maybe that makes the call site is more confusing.

I'm not sure how this relates to the changes here. Do you think that the output file should be passed as an argument to the `fish_xgettext.fish` script?

build_tools/update_translations.fish
danielrainer: I don't know if any tool looks for this file in a specific location, but the default output file of `xgettext` is `messages.po`, not `messages.pot`, so I think it's not particularly likely that moving this file causes issues.

build_tools/update_translations.fish
danielrainer: Specifying languages has been reworked. They are now non-option arguments and need to be paths to files in the po directory. New translations can be added by specifying a non-existing filename.

The specified paths are checked to ensure that they point to files in the po directory which follow our naming convention.

build_tools/update_translations.fish
danielrainer: > would be great to figure this out. The message catalogue (*.po) file format is pretty simple, we should be able to make it deterministic.

It seems that the discrepancy between running the script on my machine and in CI was a result or `sort` changing its behavior based on the locale. I modified the script to always use the `C.UTF-8` locale, which results in identical results locally and in CI.

> Can we make sure it's run as part of ninja -Cbuild test, so we're unlikely to forget?

The easiest way of achieving that would probably be adding a test case for one of the existing categories (Rust, `tests/checks`, `test/pexpects`). I'm not particularly eager to add special handling for this in CMake (both because I don't want to work with CMake and because it would increase the effort it takes to move away from it). Maybe cargo has some functionality for this?

> The test subtarget (`ninja -C build test-translations-update`) should probably not touch files, only fail if anything changed.

This would require some modifications to the `update_translations.fish` script, or maybe a separate script, because right now the hard-coded behavior is to `msgmerge` into the files in `po/`. To check for differences, we would have to create a temporary file/directory copy the original po file there, `msgmerge` the updated `messages.pot` into it, and run `diff` on it.

build_tools/update_translations.fish
krobelus: huh so `pot` means `po-template?

danielrainer: Yes, I think it does.

build_tools/update_translations.fish
danielrainer: See https://github.com/danielrainer/fish-shell/pull/1 for the current version of CI integration.

build_tools/update_translations.fish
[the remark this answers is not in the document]
krobelus: I think it's fine for now; we can maybe make it more portable later

build_tools/update_translations.fish
krobelus: right. This abbrevation seems useless, better spell it out

danielrainer: So `messages.po-template`?

build_tools/update_translations.fish
danielrainer: I just checked with a linear search, `cargo-expand` 1.0.79 is the most recent version compatible with Rust 1.70. Unfortunately, this version misses a lot of strings which were found by the current version. Maybe this is just due to a formatting difference which our regex extraction is not prepared to handle.

build_tools/update_translations.fish
krobelus: >  Do you think that the output file should be passed as an argument to the fish_xgettext.fish script?

it doesn't seem ideal that both script hardcode the name of this cache file, so yeah, an argument would work.

But that can be fixed in a follow-up, I'm merging as-is

build_tools/update_translations.fish
krobelus: > Maybe this is just due to a formatting difference which our regex extraction is not prepared to handle.

Yeah, probably.
But it's fine if we pin to a current version. Developers are not expected to use Rust 1.70 but close-to-current stable Rust.
For example the github actions runner image comes with Rust 1.86. So we can do something like `cargo install cargo-expand==123`.
Though we should first decide whether we want a failing CI check or a bot that updates stale files.

danielrainer: Good question. I think it would be cleaner to encourage people via the CI to update translations when they change some strings, rather than having bot-commits which just touch translations. On the other hand, it is a bit annoying to always have to check if the translations need updating or finding out via failing CI.

krobelus: The problem with a failing CI check is that it might annoy/confuse drive-by contributors who contribute a completion script etc.
For translation strings in Rust sources, I think a CI failure would be okay.
We could make it a separate job in github actions so one can easily see from the notification email what went wrong.

A bot making changes without manual approval sounds dangerous.
We can run it manually for now.

danielrainer: Just warning on Rust string changes does not work with the current setup. For that we would have to split the po files such that there is one file containing the Rust strings and another one for the remaining ones from fish.

krobelus: Maybe we should remove the `--description` ones for now.. not sure, I didn't check how many translations we have

danielrainer: I just checked by updating the po files with the current version of the xgettext script, committing, commenting out the line adding the `--description` strings and running
```fish
git diff po | rg '\+#\~ msgstr' | wc -l
```
This reports `35765`, so if this really is the number of translations we have for `--description` strings we would lose a lot.

danielrainer: More detailed results:
```fish
for file in po/*; echo $file; git diff -- $file | rg '\+#\~ msgstr' | wc -l; end
```
```text
po/de.po
1435
po/en.po
2193
po/fr.po
6723
po/pl.po
1
po/pt_BR.po
1002
po/sv.po
1513
po/zh_CN.po
22898
```
(I don't know why there are English translations.)

build_tools/update_translations.fish
krobelus: > My inclination is not to enforce PO file updates with a CI check but it does help to reduce string churn, which is something that I think is a bit demotivating for translators. A daily/weekly GitHub action might be better.

If we trust the bot to commit changes, yes, probably.
If the CI check causes annoyance we can remove it, having it in `ninja fish_run_tests` is also good enough.
AFAIK, CI will not run for first-time contributors due to security concerns.

PR #11465, opened 2025-05-07
Unset SYSTEMD_COLORS for systemd completion

Lillecarl:
## Description
Fixes an issue where systemctl and other systemd commands completions are prefixed by ANSI color escape codes

## TODOs:

## Notes:
Since dependencies has changed an Nix is a bit finicky with cargoHashes and stuff I cloned 4.0.2 which I'm currently using and applied this patch which i then cherry picked to master, but considering the change is only in fish code it should be irrelevant.

It was a bit unclear from the documentation if ```set -lu``` would work without unexporting the variable for the running shell but tests on my machine keeps SYSTEMD_COLORS=true in my shell.

Last minute thing: Should i use set -lu or set -fu?

### DIFF
--- a/share/functions/__fish_systemctl.fish
+++ b/share/functions/__fish_systemctl.fish
@@ -1,4 +1,7 @@
 function __fish_systemctl --description 'Call systemctl with some options from the current commandline'
+    # We don't want to complete with ANSI color codes
+    set -lu SYSTEMD_COLORS
+
     # These options are all global - before or after subcommand/arguments.
     # There's a bunch of long-only options in here, so we need to be creative with the mandatory short version.
     set -l opts t/type= s-state= p/property= a/all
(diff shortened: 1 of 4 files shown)


ridiculousfish:
Thank you!

krobelus:
nix should allow you to run `cargo build`, no?

> It was a bit unclear from the documentation if set -lu would work without unexporting the variable

only the value in the inner scope matters:

```
$ set -gx SYSTEMD_COLORS true
$ set -lu SYSTEMD_COLORS
$ set -S SYSTEMD_COLORS
$SYSTEMD_COLORS: set in local scope, unexported, with 0 elements
$SYSTEMD_COLORS: set in global scope, exported, with 1 elements
$SYSTEMD_COLORS[1]: |true|
$ env | grep SYSTEMD
$
```

PR #11469, opened 2025-05-09
Remove source locations from translations

danielrainer:
Source locations (file name and line number) where a string originates is not required by gettext tooling. It can help translators to identify context, but the value of this is reduced by our lack of context support, meaning that all occurrences of a string will receive the same translation. Translators can use `rg` or similar tools to find the source locations. For further details see this thread:
https://github.com/fish-shell/fish-shell/pull/11463#discussion_r2079378627

The main advantage is that updates to the PO files are now only necessary when the source strings change, which greatly reduces the diff noise.

A secondary benefit is that the string extraction logic is simplified. We can now directly extract the strings from fish scripts, and several issues are fixed alongside, mostly related to quoting. The regex for extracting implicit messages from fish scripts has been tweaked to ignore commented-out lines, and properly support lines starting with `and`/`or`.

## TODOs:
- [ ] Do we want alphabetically sorted messages? At the moment, there are 3 sections (Rust strings, explicit fish strings, implicit fish strings) which get sorted individually. I think it's useful to have the most important messages at the start of the file, but I'm not sure if alphabetic sorting is desirable.

### DIFF
--- a/build_tools/fish_xgettext.fish
+++ b/build_tools/fish_xgettext.fish
@@ -69,7 +69,7 @@ end
 
 # This regex handles descriptions for `complete` and `function` statements. These messages are not
 # particularly important to translate. Hence the "implicit" label.
-set -l implicit_regex '(?:^| +)(?:complete|function).*? (?:-d|--description) (([\'"]).+?(?<!\\\\)\\2).*'
+set -l implicit_regex '^(?:\s|and |or )*(?:complete|function).*? (?:-d|--description) (([\'"]).+?(?<!\\\\)\\2).*'
 extract_fish_script_messages implicit $implicit_regex
 
 # This regex handles explicit requests to translate a message. These are more important to translate
--- a/build_tools/fish_xgettext.fish
+++ b/build_tools/fish_xgettext.fish
@@ -2,15 +2,22 @@
 #
 # Tool to generate messages.pot
 
-# Create temporary directory for these operations. OS X `mktemp` is somewhat restricted, so this block
-# works around that - based on share/functions/funced.fish.
-set -q TMPDIR
-or set -l TMPDIR /tmp
-set tmpdir (mktemp -d $TMPDIR/fish.XXXXXX)
-or exit 1
+set -g output_file (status dirname)/../messages.pot
+
+# Write header. This is required by msguniq.
+# Note that this results in the file being overwritten.
+# This is desired behavior, to get rid of the results of prior invocations
+# of this script.
+begin
+    echo 'msgid ""'
+    echo 'msgstr ""'
+    echo '"Content-Type: text/plain; charset=UTF-8\n"'
+    echo ""
+end >$output_file
 
 # This is a gigantic crime.
-# xgettext still does not support rust *at all*, so we use cargo-expand to get all our wgettext invocations.
+# We use cargo-expand to get all our wgettext invocations.
+# This might be replaced once we have a tool which properly handles macro expansions.
 set -l expanded (cargo expand --lib; for f in fish fish_indent fish_key_reader; cargo expand --bin $f; end)
 
 # Extract any gettext call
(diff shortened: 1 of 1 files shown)


build_tools/fish_xgettext.fish
```diff
@@ -2,84 +2,90 @@
...
-or set -l TMPDIR /tmp
-set tmpdir (mktemp -d $TMPDIR/fish.XXXXXX)
-or exit 1
+set -g output_file (status dirname)/../messages.pot
+
+# Write header. This is required by msguniq.
+# Note that the first echo results in the file being overwritten.
+# This is desired behavior, to get rid of the results of prior invocations
+# of this script.
+echo 'msgid ""' > $output_file
```
krobelus, line 11: would be good to format the script fish with `fish_indent`

build_tools/fish_xgettext.fish
```diff
@@ -2,84 +2,90 @@
...
+echo 'msgstr ""' >> $output_file
+echo '"Content-Type: text/plain; charset=UTF-8\n"' >> $output_file
+echo "" >> $output_file
```
krobelus, line 14: if this script doesn't use stdout, we could wrap the whole script in a begin/end block,
so we only need one redirection.
If that seems to weird (I guess it is, it's too dangerous to accidentally corrupt the output), then maybe only around this header (`begin ... end >$output_file`) to reduce noise a bit

build_tools/fish_xgettext.fish
```diff
@@ -2,84 +2,90 @@
...
              # escaping difference between gettext and cargo-expand: single-quotes
-             string replace -a "\'" "'" | sort -u)
+             string replace -a "\'" "'")
 
 # Extract any constants
 set -a strs (string match -rv 'BUILD_VERSION:|PACKAGE_NAME' -- $expanded |
              string match -rg 'const [A-Z_]*: &str = "(.*)"' | string replace -a "\'" "'")
 
+# Sort the extracted strings and remove duplicates.
+# This is optional.
```
krobelus, line 34: > Do we want alphabetically sorted messages?

It should be deterministic and non-chaotic.
I think sorting is the obvious way to achieve this.
Without sorting, we might depend on `readdir()` ordering which is not guaranteed.
also, moving/renaming files should not affect translations

 > At the moment, there are 3
 > sections (Rust strings, explicit fish strings, implicit fish strings)
 > which get sorted individually. I think it's useful to have the most
 > important messages at the start of the file, but I'm not sure if
 > alphabetic sorting is desirable.

Ah so explicit strings are the ones with `_ "foo"` and implicit are completion descriptions (anything else?)

I think the main discriminator for importance is whether it's used for completions.

So maybe sort all non-completion strings and put them first, and follow that with a list of sorted completions.
I haven't taken a close look though. The current state is probably fine too.

build_tools/fish_xgettext.fish
```diff
@@ -2,84 +2,90 @@
...
+set -l strs (for str in $strs
+    echo $str
+end | sort -u)
```
krobelus, line 37: instead of an echo loop, consider `string join \n $strs` (or `printf %s\n $strs` if $strs is guaranteed non-empty)

build_tools/fish_xgettext.fish
```diff
@@ -2,84 +2,90 @@
...
-function extract_fish_script_messages --argument-names name regex;
-    mkdir -p $tmpdir/$name/share/completions $tmpdir/$name/share/functions
-    for f in share/config.fish share/completions/*.fish share/functions/*.fish
-        # Extract messages from fish script.
-        # This is done by creating a file which has the message strings in the same lines as the
-        # oritinal, in the form:
-        # N_ "message"
-        # All other lines will be empty.
-        # Multiple messages on a single line are not supported.
+function extract_fish_script_messages --argument-names regex;
```
krobelus, line 47: unrelated: spurious semicolon (fish_indent should erase that)

build_tools/fish_xgettext.fish
```diff
@@ -2,84 +2,90 @@
...
+    # Using xgettext causes more trouble than it helps.
+    # This is due to handling of escaping in fish differing from formats xgettext understands
+    # (e.g. POSIX shell strings).
+    # We work around this issue by manually writing the file content.
```
krobelus, line 51: yeah that's a good approach.
Let's hope msgfmt and friends have good error checking but I don't think the format is likely to change

build_tools/fish_xgettext.fish
```diff
@@ -2,84 +2,90 @@
...
-
-xgettext -j -k -kN_ -LShell --from-code=UTF-8 -cDescription --no-wrap --add-location=file -o messages.pot $tmpdir/{ex,im}plicit/share/*/*.fish
+extract_fish_script_messages $explicit_regex
 
-# Remove the tmpdir from the location to avoid churn
-sed -i 's_^#: /.*/share/_#: share/_' messages.pot
+# This regex handles descriptions for `complete` and `function` statements. These messages are not
+# particularly important to translate. Hence the "implicit" label.
+# set -l implicit_regex '(?:^| +)(?:complete|function).*? (?:-d|--description) (([\'"]).+?(?<!\\\\)\\2).*'
+set -l implicit_regex '^(?:\s|and |or )*(?:complete|function).*? (?:-d|--description) (([\'"]).+?(?<!\\\\)\\2).*'
```
krobelus, line 82: the and/or fix seems like a separate fix, ideally in a separate commit.
do we use `and complete` anywhere though?
There are a bunch of other cases where this regex won't work, such as `foo=bar complete` etc.

I wonder if we should switch to a more robust approach, maybe extract the complete description argument directly from the fish AST, see `src/ast.rs`.
But of course that's not important, we already have tons of untranslated completions.

build_tools/fish_xgettext.fish
danielrainer: > It should be deterministic and non-chaotic.
I think sorting is the obvious way to achieve this.
Without sorting, we might depend on readdir() ordering which is not guaranteed.
also, moving/renaming files should not affect translations

Yes, these are good reasons for sorting. The only real advantage I see against sorting is that keeping strings originating from one file together makes it slightly more convenient to translate this file.

> Ah so explicit strings are the ones with _ "foo" and implicit are completion descriptions (anything else?)

Essentially yes. This distinction is made based on the 2 regexes, which I did not write (I modified the implicit one to fix some issues). The explicit one should detect `_ "foo"` messages as you point out, and the implicit one descriptions of completions and functions. The regex parsing is of course really sketchy, so if we have a better option we should use that.

> So maybe sort all non-completion strings and put them first, and follow that with a list of sorted completions.

Maybe I misunderstand what you mean, but as I understand it this is the current state. We have sorted rust strings, then sorted non-completion (`_`) strings, then sorted descriptions (of completions/functions).

build_tools/fish_xgettext.fish
danielrainer: `string join` does not seem to be up to the task at the moment, see #11478.

danielrainer: Ok, as mentioned in https://github.com/fish-shell/fish-shell/issues/11478#issuecomment-2869907283, using `string join \n -- $strs` works.

build_tools/fish_xgettext.fish
danielrainer: > the and/or fix seems like a separate fix, ideally in a separate commit.

Fair point. The reason it is included here is that I noticed corrupted lines in the output caused by the old regex looking for at least one space or the start of the line. `string replace --filter` works in a way which might be a bit surprising. That is, `--filter` prints the entire line if a replacement was done. With the old regex, lines of the shape `^([^ ]*) complete -d "foo"` would get replaced by `$1foo` instead of just `foo`. Specifically, I noticed this happening with commented-out lines and lines of the shape `and complete ...`, so yes that is actually in use. (I have not seen `or complete` tbf, but it seemed weird to only include `and` but not `or`.

Using proper parsing instead of a sketchy regex would of course be a lot better.

build_tools/fish_xgettext.fish
danielrainer: Done (for the header).

build_tools/fish_xgettext.fish
danielrainer: At least when I was still trying to get `xgettext` to work, `msgmerge` complained quite a bit about some things. The dangerous part are quotes which are escaped according to fish syntax but not POSIX syntax. This can result in a large amount of the strings being silently skipped.

build_tools/fish_xgettext.fish
danielrainer: The regex change is separated into its own commit now.

build_tools/fish_xgettext.fish
krobelus: > Maybe I misunderstand what you mean

I meant putting rust and the other explicit strings (from fish script) into a single sorted group.

The overwhelming majority of strings are from completions.. we have 20k of them.
I'm not sure if it really makes sense to vendor these translations in the fish project.
Would be nice if we could source this information from something like man pages..
but the question is how to get there.
I guess it's not causing problems today although 2MB per translation file seems a lot.

build_tools/fish_xgettext.fish
krobelus: > quotes which are escaped according to fish syntax but not POSIX syntax.

ugh yeah. I think all our translated strings use fish-style quotes in the source, so `string unescape` is appropriate. Judging from your changes, it looks like the po format only requres escaping backslashes and double quotes.

build_tools/fish_xgettext.fish
```diff
@@ -2,84 +2,89 @@
...
 for str in $strs
-    # grep -P needed for string escape to be compatible (PCRE-style),
-    # -H gives the filename.
-    # If you want to run this on non-GNU grep: Don't.
-    # The sed command extracts just the filename from the matches grep finds,
-    # and prepends the '#: ' prefix, marking the line as a source refecence.
-    # sort -u just gets rid of duplicates.
-    grep -PH -r -- \"(string escape --style=regex -- $str)\" src/ |
-       sed -E 's/^([^:]*):.*$/#: \1/' | sort -u
     echo "msgid \"$str\""
```
krobelus, line 41: I guess the fish strings are correct now but I'm not sure if rust strings are.
As written, the code assumes that gettext understands escape sequences found in Rust strings.
(Not sure if we actually get things like `\u{1234}` here; it's probably irrelevant because
 we don't really use those)

build_tools/fish_xgettext.fish
```diff
@@ -2,84 +2,89 @@
...
     echo 'msgstr ""'
```
krobelus, line 42: in future, maybe we can also extract a function for writing the msgid+msgstr parts (we do the same thing twice).

build_tools/fish_xgettext.fish
krobelus: If everyone agrees that we should use the formatter, I'd prefer that - it would be the progressive/modern approach.
But there are also good arguments against using automated code formatting (it strictly limits expressivity), and I don't really care

build_tools/fish_xgettext.fish
danielrainer: > I meant putting rust and the other explicit strings (from fish script) into a single sorted group.

Ah, ok. I don't have a strong preference either way.

> The overwhelming majority of strings are from completions.. we have 20k of them.
I'm not sure if it really makes sense to vendor these translations in the fish project.
Would be nice if we could source this information from something like man pages..
but the question is how to get there.

Yes, and the amount is indeed overwhelming for someone considering translating it. I don't know if there is any established way to get translations directly from the tools. Some of them probably don't even have any to begin with.

> I guess it's not causing problems today although 2MB per translation file seems a lot.

Yes, it's quite a bit. Some editors might even struggle with such files. Depending on how well git compresses the different versions of the files, this might also become a fairly large overhead. At least the mo files are somewhat smaller (1.28 MiB at the moment for all of them combined) and they are not tracked in git.

build_tools/fish_xgettext.fish
danielrainer: > Judging from your changes, it looks like the po format only requres escaping backslashes and double quotes.

From https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html

> Each of untranslated-string and translated-string respects the C syntax for a character string, including the surrounding quotes and embedded backslashed escape sequences, except that universal character escape sequences (\u and \U) are not allowed.

I have not tested the behavior of various escape sequences extensively, but the current escaping seems to at least ensure that the strings are accepted by msgmerge and msgfmt, without dropping parts of the file due to an incorrectly escaped quote. But more testing would be useful to see if escape sequences work as expected.

build_tools/fish_xgettext.fish
danielrainer: Good point, although I also don't know whether the `string escape` command results in strings which are usable for gettext translation. Proper testing for escape sequences, both for Rust and fish script strings would help.

zanchey:
I don't do translations but it seems like it might be useful to know that a string comes from completions rather than core fish? If we're asking people to prioritise, the latter is probably a more useful application of their efforts.

krobelus:
> I don't do translations but it seems like it might be useful to know that a string comes from completions rather than core fish? If we're asking people to prioritise, the latter is probably a more useful application of their efforts.

It should be more or less obvious from the sorting, and we could add comments in between sections.
Also possibly split translations into multiple files if tools can handle that..

danielrainer:
> we could add comments in between sections

That would be easy to do.

> possibly split translations into multiple files if tools can handle that

Tools could handle that (`msgcat`), but I don't think the added complexity is worth it.

PR #11475, opened 2025-05-10
Add support for using embedded mo file

danielrainer:
Embedding the MO files for gettext means that we cannot call the system gettext
anymore, since there is no actual file it can read from (unless we write out the
embedded data onto the file system).
    
Therefore, the `gettext` crate is used. This crate does not seem to be
maintained, but it works well enough for this proof of concept.
    
For this to work, the MO files need to be present at build time. Run
`build_tools/update_translations.fish` to ensure that this is the case.
Binaries built this way have the MO files embedded (if the `embed-data`) feature
is active. Setting the relevant environment variables should then result in
translations working as expected for the most part.
    
See https://www.gnu.org/software/gettext/manual/gettext.html#Locale-Environment-Variables-1
for the supported variables. My implementation roughly follows the concepts
described there, but there are probably some differences to the original gettext
implementation, especially for the more complex cases.
Fallback logic is not ideal at the moment. E.g. if someone sets `LANGUAGE` to
`pt`, for which no translations exist, the fallback to use the source strings,
instead of `pt_BR` which does exist. The other way around (from language with
region to just language) works.
    
The code is quite messy at the moment, as this is just intended as a
proof-of-concept.

## TODOs:
- [ ] Clean up code
- [ ] Decide on fallback handling
- [ ] Ensure that mo files exist and are up to date when building


Cargo.toml
```diff
@@ -58,6 +58,8 @@ widestring = "1.2.0"
 # We need 0.9.0 specifically for some crash fixes.
 terminfo = "0.9.0"
 rust-embed = { version = "8.2.0", optional = true }
+# TODO: this library seems unmaintained
+gettext = "0.4.0"
```
ridiculousfish, line 62: Let's make sure this is only pulled in if `rust-embed` is enabled

src/wutil/gettext.rs
```diff
@@ -1,136 +1,254 @@
...
-        unsafe { textdomain(domainname.as_ptr()) }
+
+    pub(super) fn wgettext(text: MaybeStatic) -> &'static wstr {
+        let key = match text {
+            MaybeStatic::Static(s) => s,
+            MaybeStatic::Local(s) => s,
+        };
+        let key_string = key.to_string();
+
+        let language_priorities = get_language_priorities();
```
ridiculousfish, line 97: if I'm reading this right, this is invoked on every call and looks rather expensive. I think we should cache the values of `wgettext` as we do with `WGETTEXT_MAP` in the non-embedded case.

ridiculousfish:
General approach looks OK and I'm comfortable adding a conditional dependency on this gettext crate.

Probably we should use the `no-optimized-legacy-encoding` feature to reduce the binary size.

src/wutil/gettext.rs
danielrainer: The code in this function is indeed not well optimized, it's basically just the first version which worked.

Caching is not that straightforward here. Caching translation results, as it is done in the other implementation, has the problem that it does not update when the language changes, which I'd like to avoid.

We need a `&str` to pass to `catalog.gettext`, so we might be able to cache the translation from the `&wstr` provided as the argument to a corresponding string (reference). This might look like
```diff
diff --git a/src/wutil/gettext.rs b/src/wutil/gettext.rs
index 2bf560441..183250ebd 100644
--- a/src/wutil/gettext.rs
+++ b/src/wutil/gettext.rs
@@ -87,11 +87,21 @@ fn load_catalog_for_language(lang: &str) -> Option<Catalog> {
     }
 
     pub(super) fn wgettext(text: MaybeStatic) -> &'static wstr {
-        let key = match text {
+        let key: &'static wstr = match text {
             MaybeStatic::Static(s) => s,
-            MaybeStatic::Local(s) => s,
+            MaybeStatic::Local(s) => wstr::from_char_slice(Box::leak(s.as_char_slice().into())),
+        };
+        static KEYS: Lazy<Mutex<HashMap<&'static wstr, String>>> =
+            Lazy::new(|| Mutex::new(HashMap::new()));
+        let mut keys = KEYS.lock().unwrap();
+        let key_str = match keys.get(key) {
+            Some(key_str) => key_str,
+            None => {
+                let key_string = key.to_string();
+                keys.insert(key, key_string);
+                keys.get(key).unwrap()
+            }
         };
-        let key_string = key.to_string();
 
         let language_priorities = get_language_priorities();
 
@@ -101,12 +111,12 @@ pub(super) fn wgettext(text: MaybeStatic) -> &'static wstr {
         for language in language_priorities {
             match catalogs.get(&language) {
                 Some(catalog) => {
-                    let message = catalog.gettext(&key_string).to_owned();
+                    let message = catalog.gettext(key_str).to_owned();
                     return Box::leak(str2wcstring(message.as_bytes()).into_boxed_utfstr());
                 }
                 None => {
                     if let Some(catalog) = load_catalog_for_language(&language) {
-                        let message = catalog.gettext(&key_string).to_owned();
+                        let message = catalog.gettext(key_str).to_owned();
                         catalogs.insert(language, catalog);
                         return Box::leak(str2wcstring(message.as_bytes()).into_boxed_utfstr());
                     }
```
which eliminates repeated translations from `&wstr` to `String` (when the reference passed in `text` has been seen before), but instead we have a leak each time the function is called with `MaybeStatic::Local`, and if I understand the code correctly we also allocate a `String` in every such case and add it to the `HashMap`. Maybe there is a better way to do this, but I'm not convinced that the approach in this diff is better than just always translating from `&wstr` to `String`, at least until I see some benchmarks which show a significant difference favoring the caching approach.

The plans to move away from `&wstr` also make it seem somewhat less important to have efficient handling that depends on `&wstr`. (Although I don't know if/when the change will actually happen, and what the replacement will be.)

danielrainer:
> Probably we should use the `no-optimized-legacy-encoding` feature to reduce the binary size.

I'm assuming you're referring to the crate feature of the `encoding` crate, which is a dependency of the `gettext` crate. `gettext` depends on `encoding 0.2.32`, but the feature you mention was introduced later in https://github.com/lifthrasiir/rust-encoding/commit/726d6dae7c910a3342a187e5618b8d7e128d03ff. I guess that's what you get when using unmaintained crates.

krobelus:
> it does not update when the language changes,

don't we clear the cache when a relevant locale variable changes?

danielrainer:
> don't we clear the cache when a relevant locale variable changes?

No, once an entry is added it stays there until fish exits.

Try using the old version on master, or here by building with `--no-default-features`. First set the language to something other than English and cause fish to show a message in this language. Then, set the language back to English and trigger the same message again. It will be in the non-English language instead of English.

src/wutil/gettext.rs
```diff
@@ -1,136 +1,253 @@
...
 
-// Really init wgettext.
-fn wgettext_really_init() {
-    let Some(ref localepath) = CONFIG_PATHS.locale else {
-        return;
-    };
-    let package_name = CString::new(PACKAGE_NAME).unwrap();
-    // This contains `datadir`; which when replaced to make the binary relocatable,
-    // causes null bytes at the end of the string.
-    let localedir = CString::new(localepath.display().to_string()).unwrap();
```
krobelus, line 57: Is it correct that `CONFIG_PATHS.locale` should be None when we have embedded the `*.mo` files?
If yes, we probably need to change the logic in `src/env/config_paths.rs` accordingly.

krobelus: ah yes, that's already what's happening.. except if the path to fish ends in `bin/fish`,
or if `$(status fish-path)/../../share/install` exists.. though that one is purely a legacy thing (since we have embedding now), so we can probably remove the "install" fallback.. not yet sure)

danielrainer:
Closing this in favor of #11726 (or possibly #11725).

PR #11477, opened 2025-05-11
Always treat brace at command start as compound statement

krobelus:
For backwards compatibility, fish does not treat "{echo,hello}" as a compound
statement but as brace expansion (effective "echo hello").  We interpret
"{X...}" as compound statement only if X is whitespace or ';' (which is an
interesting heuristic).

A brace expansion at the very start of a command 
is usually pointless (space separation is shorter).
The exception are cases where the command name and the first few arguments
share a suffix.

	$ {,1,2,3,4}echo
	1echo 2echo 3echo 4echo

Not sure if anyone uses something like that.  Perhaps we want to trade
compatibility for simplicity. I don't really have an opinion on this.

Always parse the opening brace as first character of a command token as
compound statement.
Brace expansion can still be used with a trick like: «''{echo,foo}»

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -21,7 +21,7 @@ Notable improvements and fixes
 
 Deprecations and removed features
 ---------------------------------
-- Tokens like ``{ echo, echo }`` in command position are no longer interpreted as brace expansion but as compound command.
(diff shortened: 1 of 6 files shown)


ridiculousfish (approved):
LGTM - I can't imagine this will break anyone. Please check whether the docs should be updated.

PR #11480, opened 2025-05-11
Completion cache

fabiojb:
## Description

Add `__fish_cache_read` and`__fish_cached` functions and replace cache implementations with the new cache functions.

## Tests

- [x] completion cache for `pacman` command
- [x] completion cache for `flatpak` command
- [ ] completion cache for `yum` command
- [ ] completion cache for `rpm` command
- [ ] completion cache for `xbps` command

### DIFF
--- a/share/functions/__fish_cache_read.fish
+++ b/share/functions/__fish_cache_read.fish
@@ -1,0 +1,10 @@
+function __fish_cache_read -a cache_file max_age
+    if not test -f $cache_file
+        return 1
+    end
+    set -l age (path mtime --relative $cache_file)
+    if test -n "$max_age" && test $age -gt $max_age
+        return 1
+    end
+    cat $cache_file
+end
--- a/share/functions/__fish_cached.fish
+++ b/share/functions/__fish_cached.fish
@@ -1,0 +1,21 @@
+function __fish_cached --description "Cache the command output for a given amount of time"
+
+    argparse --stop-nonopt 't/max_age=' 'k/cache_key=' -- $argv
+
+    if set -q _flag_cache_key
+        set cache_key $_flag_cache_key
+    else
+        set cache_key (string split ' ' "$argv" | head -1)
+    end
+    set cache_dir (__fish_make_cache_dir)
+    set cache_file (path normalize $cache_dir/$cache_key)
+    set cache_age (path mtime --relative $cache_file)
+
+    if not test -f $cache_file
+        or test $cache_age -gt $_flag_max_age
+
+        eval "$argv" >$cache_file
+        __fish_cache_put $cache_file
+    end
+    cat $cache_file
+end
--- a/share/functions/__fish_print_pacman_packages.fish
+++ b/share/functions/__fish_print_pacman_packages.fish
@@ -5,15 +5,8 @@ function __fish_print_pacman_packages
     argparse i/installed -- $argv
     or return 1
 
-    set -l xdg_cache_home (__fish_make_cache_dir)
-    or return
-
     if not set -q _flag_installed
-        set -l cache_file $xdg_cache_home/pacman
-        __fish_cache_read $cache_file 250 && return
-        __fish_cache_put $cache_file
-        # prints: <package name>	Package
-        pacman -Ssq | sed -e 's/$/\t'Package'/' | tee $cache_file
+        __fish_cached -t 250 -- pacman -Ssq \| sed 's/\$/\tPackage/'
         return 0
     else
         pacman -Q | string replace ' ' \t
--- a/share/functions/__fish_print_pacman_packages.fish
+++ b/share/functions/__fish_print_pacman_packages.fish
@@ -13,7 +13,7 @@ function __fish_print_pacman_packages
         __fish_cache_read $cache_file 250 && return
         __fish_cache_put $cache_file
         # prints: <package name>	Package
-        pacman -Ssq | sed -e 's/$/\t'Package'/' >$cache_file &
+        pacman -Ssq | sed -e 's/$/\t'Package'/' | tee $cache_file
         return 0
     else
         pacman -Q | string replace ' ' \t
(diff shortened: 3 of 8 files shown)


share/functions/__fish_cache_read.fish
```diff
@@ -0,0 +1,10 @@
+function __fish_cache_read -a cache_file max_age
+    if not test -f $cache_file
+        return 1
+    end
+    set -l age (path mtime --relative $cache_file)
+    if test -n "$max_age" && test $age -gt $max_age
+        return 1
+    end
+    cat $cache_file
+end
```
krobelus, line 10: missing a newline at the end of file -- consider running this through `fish_indent`.

krobelus, line 10: this function call is always followed by a call to `__fish_cache_put`,
Can we inline that one into `__fish_cache_read`?

For extra points, we could also teach this function how to populate the cache;
we could pass a "callback" argument that we do "eval $callback" on.
Then we should only pass the basename of the cache file to this function (and have it call __fish_make_cache_dir itself).
Then a good name for the function might be

	__fish_cached <cache-key> <max-age> 'echo <contents>'

share/functions/__fish_print_pacman_packages.fish
```diff
@@ -10,17 +10,10 @@ function __fish_print_pacman_packages
...
-            set -l max_age 250
-            if test $age -lt $max_age
-                return
-            end
-        end
+        __fish_cache_read $cache_file 250 && return
         __fish_cache_put $cache_file
         # prints: <package name>	Package
-        pacman -Ssq | sed -e 's/$/\t'Package'/' >$cache_file &
+        pacman -Ssq | sed -e 's/$/\t'Package'/' | tee $cache_file
```
krobelus, line 16: I think the idea was that "no completions" is better than "hanging for a while" because the latter might force the user to press ctrl-c.
I think most other users of this cache logic do the same.

But this command seems fast enough (200ms here) so this seems like a good change.
(perhaps we don't need the cache at all.. not sure)

share/functions/__fish_print_rpm_packages.fish
```diff
@@ -31,14 +23,7 @@ function __fish_print_rpm_packages
...
         set -l cache_file $xdg_cache_home/rpm
-        if test -f $cache_file
-            cat $cache_file
-            set -l age (path mtime -R -- $cache_file)
-            set -l max_age 250
-            if test $age -lt $max_age
-                return
-            end
-        end
+        __fish_cache_read $cache_file 250 && return
```
krobelus, line 26: there's a subtle behavior change here.
Previously, we used to use the contents of the cached file even when the file is too old.
i.e. the max_age check is only used to decide whether to refresh the cache.
This seems like a sensible semantics for most (all?) callers, no?

fabiojb: Good point, I didn't notice that. The only drawback is that when there is no cache file already created, no completions will be returned.

share/functions/__fish_print_pacman_packages.fish
fabiojb: > I think the idea was that "no completions" is better than "hanging for a while" because the latter might force the user to press ctrl-c.

Not sure what could be the best behavior, but I'm tending to the idea of "hanging for a while". It seems to be a little more predictable and gives the user the control. Maybe there could be also a timeout to limit the maximum time for the completion to hang

share/functions/__fish_cache_read.fish
fabiojb: Working on that now. 

Instead of being a parameter, the cache key could be created based on the command name, something like that:

`__fish_cached -t 250 -- pacman -Ssq \| sed -e 's/\$/\tPackage/'`

Extracting the first part of  the command, we get `pacman`, that can be used as the cache key.

I'm wondering if the functions `__fish_cache_read` and `__fish_cache_put` should be keep as separate functions or should be merged inside the `__fish_cached` function. @krobelus  What do you think?

fabiojb: An optional argument could also be provided to explicit specify a cache key

share/functions/__fish_cached.fish
```diff
@@ -0,0 +1,21 @@
+function __fish_cached --description "Cache the command output for a given amount of time"
```
fabiojb, line 1: Here I kept the behavior of hanging up the execution until the cache load command is finished.

This can still be improved. Maybe hanging up only when there is no cache file, and after that when the cache is expired, instantly return the cached response and an cache update command in background

share/functions/__fish_cached.fish
```diff
@@ -0,0 +1,21 @@
...
+        set cache_key (string split ' ' "$argv" | head -1)
+    end
+    set cache_dir (__fish_make_cache_dir)
+    set cache_file (path normalize $cache_dir/$cache_key)
+    set cache_age (path mtime --relative $cache_file)
+
+    if not test -f $cache_file
+        or test $cache_age -gt $_flag_max_age
+
+        eval "$argv" >$cache_file
```
fabiojb, line 17: Here possibly we need to parse the status of the eval command, and clean the cache file in case of failure

share/functions/__fish_print_port_packages.fish
```diff
@@ -6,15 +6,7 @@ function __fish_print_port_packages
...
-    if test -f $cache_file
-        cat $cache_file
-        set -l age (path mtime -R -- $cache_file)
-        set -l max_age 250
-        if test $age -lt $max_age
-            return
-        end
-    end
-
+    __fish_cache_read $cache_file 250 && return
```
fabiojb, line 9: Still using `__fish_cache_read` and `__fish_cache_put` here, not really sure how to re-write and test this

fabiojb:
I'm closing this PR to create new ones with separate changes. @krobelus please take a look at #11526

krobelus:
the usual convention in this project is to rewrite the existing PR
via force-push but no worries, I'll take a look, sure

Issue #11483, opened 2025-05-12
SIGBUS in fish::history::file::HistoryFileContents::decode_item


asomers:
After trying out fish 4.0.2 for a few hours, it crashed with SIGBUS and dumped core.  The core dump looks nearly the same as that reported in #10782 , suggesting that the solution for that issue was incomplete.

**fish --version** fish, version 4.0.2
**OS** FreeBSD 15.0-CURRENT amd64
**Terminal** xfce4-terminal 1.1.4

```
#0  0x00003e76fd955380 in fish::history::file::HistoryFileContents::decode_item
    ()
#1  0x00003e76fd95844c in fish::history::HistoryImpl::item_at_index ()
#2  0x00003e76fd9581a5 in fish::history::History::item_at_index ()
#3  0x00003e76fd957a79 in fish::history::HistorySearch::go_to_next_match ()
#4  0x00003e76fd81dbd8 in core::ops::function::FnOnce::call_once{{vtable.shim}}
    ()
#5  0x00003e76fd81bbe3 in core::ops::function::FnOnce::call_once{{vtable.shim}}
    ()
#6  0x00003e76fd8d8c21 in std::sys::backtrace::__rust_begin_short_backtrace ()
#7  0x00003e76fd88dd0a in core::ops::function::FnOnce::call_once{{vtable.shim}}
    ()
#8  0x00003e76fd992c1b in std::sys::pal::unix::thread::Thread::new::thread_start ()
#9  0x00003e7f20eedd72 in thread_start (curthread=0x5d2f8c014a00)
    at /usr/home/somers/src/freebsd.org/src/lib/libthr/thread/thr_create.c:289
#10 0x0000000000000000 in ?? ()
Backtrace stopped: Cannot access me
```
This has only happened once so far, so I don't have steps to reproduce.  But the last command I ran was `cu -l` to connect to a virtual machine's serial port.  When I pressed `~.` to close the connection, that's when fish crashed.

asomers:
No, I didn't.  But I should mention, if it matters, that my home directory is NFS-mounted.

krobelus:
Hmm fish should never mmap files on network file systems, and I don't expect anything else would be causing SIGBUS. 
Maybe the remote-filesystem detection is not working properly?
What's the output of `cargo b && target/debug/fish -ic exit` with

```diff
diff --git a/src/path.rs b/src/path.rs
index 650e88d87d..1233320c5a 100644
--- a/src/path.rs
[...]

PR #11486, opened 2025-05-12
Fix remote filesystem detection on FreeBSD

asomers:
## Description

Need an extra include to get the definition of MNT_LOCAL

Fixes issue #11483 

## TODOs:

### DIFF
--- a/src/libc.c
+++ b/src/libc.c
@@ -4,6 +4,7 @@
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <sys/mount.h>
 #include <sys/resource.h>
 #include <unistd.h>
 


src/libc.c
```diff
@@ -4,6 +4,7 @@
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <sys/mount.h>
```
krobelus, line 7: nice job - sorry about that. I should have checked https://man.freebsd.org/cgi/man.cgi?query=statfs.

I think we also need add `#include <sys/statvfs.h>` for `ST_LOCAL`, judging from https://man.netbsd.org/statvfs.5

asomers: Good idea.  I'll try to test that in my NetBSD VM.

krobelus: NetBSD actually does `#define ST_LOCAL MNT_LOCAL`
I didn't test it but I assume it works

PR #11487, opened 2025-05-12
Fix use of deprecated symbols in Nix

asomers:
## Description

Just maintenance.  Nix renamed these symbols for consistency's sake.

There are no user-visible changes.


asomers:
cc @Berrysoft

PR #11489, opened 2025-05-12
Do not hardcode xgettext output path

danielrainer:
This makes it easier to choose a different location for the messages template file.

### DIFF
--- a/build_tools/fish_xgettext.fish
+++ b/build_tools/fish_xgettext.fish
@@ -1,96 +1,96 @@
 #!/usr/bin/env fish
 #
-# Tool to generate messages.pot
+# Tool to generate gettext messages template file.
+# Writes to stdout.
 
-set -g output_file (status dirname)/../messages.pot
-
-# Write header. This is required by msguniq.
-# Note that this results in the file being overwritten.
-# This is desired behavior, to get rid of the results of prior invocations
-# of this script.
 begin
-    echo 'msgid ""'
-    echo 'msgstr ""'
-    echo '"Content-Type: text/plain; charset=UTF-8\n"'
-    echo ""
-end >$output_file
-
-set -l cargo_expanded_file (mktemp)
-# This is a gigantic crime.
-# We use cargo-expand to get all our wgettext invocations.
-# This might be replaced once we have a tool which properly handles macro expansions.
-begin
-    cargo expand --lib
-    for f in fish fish_indent fish_key_reader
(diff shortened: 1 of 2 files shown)


danielrainer:
By the way, when do we want to do the massive update of the PO files to the new format? Do we want some testing (e.g. for escape sequences) first?

build_tools/fish_xgettext.fish
```diff
@@ -1,8 +1,13 @@
...
-# Tool to generate messages.pot
+# Tool to generate gettext messages template file.
 
-set -g output_file (status dirname)/../messages.pot
+if test (count $argv) -ne 1
+    echo "Invalid invocation. Specify the output file path as the only argument."
+    exit 1
+end
+
+set -g output_file $argv[1]
```
krobelus, line 10: could we use stdout? That would be more obvious, no?

krobelus:
> By the way, when do we want to do the massive update of the PO files to the new format? Do we want some testing (e.g. for escape sequences) first?

I think so yeah. (As long as space is not an issue..).
Although I think we should consider whether it's worth including 20k mostly untranslated completion descriptions that are unlikely to be translated. Maybe we should hide those for now, perhaps we can find a more sustainable solution.

>  Do we want some testing (e.g. for escape sequences) first?

I think it will be easy to review; we can grep for all msgid/msgstr entries that contain any special character, and double check those. That won't protect us from missing characters but I think our logic should be easy to prove correct.

build_tools/fish_xgettext.fish
danielrainer: Yes, done.

danielrainer:
> Although I think we should consider whether it's worth including 20k mostly untranslated completion descriptions that are unlikely to be translated. Maybe we should hide those for now, perhaps we can find a more sustainable solution.

That might be sensible. We would just have to comment out one line (or two, if we also want to exclude the messages in fish scripts with explicit translation requests).

One thing I have noticed is that command descriptions are translated in several cases. E.g. entering `ls` and triggering completions when the language is set to German shows several command descriptions in German. These do not appear in our `de.po` file, so I wonder where they are coming from.

krobelus:
command descriptions are from `apropos` from man-db (which provides `/usr/share/locale/de/LC_MESSAGES/man-db.mo`)

```
$ LANG=de_DE.UTF-8 apropos lsblk
lsblk (8)            - Blockgeräte auflisten
```

PR #11500, opened 2025-05-15
Update translation docs

danielrainer:
This is done in accordance with the recent changes to our translation pipeline.


CONTRIBUTING.rst
```diff
@@ -286,24 +286,68 @@ Contributing Translations
...
 other languages.
 
-Creating and updating translations requires the Gettext tools, including
-``xgettext``, ``msgfmt`` and ``msgmerge``. Translation sources are
+Translation sources are
 stored in the ``po`` directory, named ``LANG.po``, where ``LANG`` is the
-two letter ISO 639-1 language code of the target language (eg ``de`` for
-German).
+two letter ISO 639-1 language code of the target language (e.g. ``de`` for
+German). A region specifier can also be used (e.g. ``pt_BR`` for Brazilian Portuguese).
```
krobelus, line 292: probably we can replace https://github.com/fish-shell/fish-shell/wiki/Translations with a link to this section in CONTRIBUTING (assuming we don't want to maintain both)?

danielrainer: Sounds good. Do you think the information about how to install fish that's currently on the wiki page should be in the translations section in `CONTRIBUTING.rst`?

Also, should we link to
https://github.com/fish-shell/fish-shell/blob/master/CONTRIBUTING.rst#contributing-translations
or
https://fishshell.com/docs/current/contributing.html#contributing-translations
?

PR #11505, opened 2025-05-16
Move and rename messages template file

danielrainer:
There is no reason to have this file clutter the repo root. Move it into the `po` directory, and give it a more descriptive name.

Inspired by this discussion:
https://github.com/fish-shell/fish-shell/pull/11463#discussion_r2083453275

I use `template.po` instead of `message.po-template` to be more compatible with automatic filetype detection. (e.g. vim/nvim detect the former as a po file, but the latter as a conf file)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,7 +38,7 @@ Desktop.ini
 Thumbs.db
 ehthumbs.db
 
-messages.pot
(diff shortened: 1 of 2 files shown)


ridiculousfish:
thank you!

Issue #11478, opened 2025-05-11
`string join` treats strings starting with `-` or `--` as argument flags


danielrainer:
If wanting to join arbitrary strings, it might happen that they start with hyphens.

Consider
```fish
set strs --asdf --qwer; string join \n $strs
```
Which should (in my opinion) produce
```text
--asdf
--qwer
```
But instead it fails with the error message
```text
string join: --asdf: unknown option

(Type 'help string' for related documentation)
```
This reproduces on current master (cc2ca60baaec4b686c1c2bb0ac7824252ee36a2e).

I would like this command to parse all arguments after `string join` sequentially, treating each one as a flag argument until one argument does not start with a hyphen. This argument is the separator. All remaining arguments are to be interpreted as literal strings, which should be joined with the separator.

(Semi-related: The webdocs do not include the `--no-empty` flag in the command summary.)

Xiretza:
Delimit options from non-options using `--`:

```
$ set strs --asdf --qwer; string join -- \n $strs
--asdf
--qwer
```

This is a general convention in most command line tools.

danielrainer:
Ah yes, that works. I am aware of the convention, but it is not mentioned in the docs. The current behavior seems quite surprising to me, so I think it at least deserves a clear warning in the docs.

Xiretza:
Opinions differ on whether or not non-options should terminate option processing, though nowadays the consensus seems to be that they shouldn't, and all fish builtins behave accordingly. Some comparisons with other tools that behave the same:

```
$ set files /dev/null --version; grep foo $files
grep (GNU grep) 3.12-modified
[...]
$ set dirs foo bar --version; mkdir $dirs
mkdir (GNU coreutils) 9.7
[...]
$ set urls https://www.youtube.com/watch?v=dQw4w9WgXcQ --version; yt-dlp $urls
2025.01.12
```

So this isn't specific to `string join`, or even specific to fish. In order to pass arbitrary strings to commands robustly, you need to first terminate option parsing using `--`.

danielrainer:
I mostly agree.

For many tools it would be quite weird to pass non-option arguments starting with `-`, so there it's definitely fine to be liberal about the option's position. The difference for `string join` is that this command might reasonably be expected to "just work" with a list of arbitrary strings as its last argument. I don't think there is a major benefit of having options after the separator for this command, other than maybe not having to write custom parsing logic.

[...]

PR #11506, opened 2025-05-16
Improve docs for `string join`

danielrainer:
Fixes issue #11478


doc_src/cmds/string-join.rst
```diff
@@ -18,11 +18,26 @@ Description
 
 .. BEGIN DESCRIPTION
 
-``string join`` joins its *STRING* arguments into a single string separated by *SEP*, which can be an empty string. Exit status: 0 if at least one join was performed, or 1 otherwise. If ``-n`` or ``--no-empty`` is specified, empty strings are excluded from consideration (e.g. ``string join -n + a b "" c`` would expand to ``a+b+c`` not ``a+b++c``).
+Joins its *STRING* arguments into a single string separated by *SEP* (for ``string join``) or by the
+zero byte (NUL) (for ``string join0``).
+Exit status: 0 if at leas one join was performed, or 1 otherwise.
```
krobelus, line 23: typo: "leas"

doc_src/cmds/string-join.rst
```diff
@@ -18,11 +18,26 @@ Description
...
 
-``string join0`` joins its *STRING* arguments into a single string separated by the zero byte (NUL), and adds a trailing NUL. This is most useful in conjunction with tools that accept NUL-delimited input, such as ``sort -z``. Exit status: 0 if at least one join was performed, or 1 otherwise.
+``-n`` or ``--no-empty``: Exclude empty strings from consideration (e.g. ``string join -n + a b "" c`` would expand to ``a+b+c`` not ``a+b++c``).
```
krobelus, line 25: I wonder if we have this "list" style anywhere else?
I know it's common in man pages but we probably use full-sentences everywhere.
But the added whitespace seems to make it easier to read, so this looks good.

doc_src/cmds/string-join.rst
```diff
@@ -18,11 +18,26 @@ Description
...
 
-Because Unix uses NUL as the string terminator, passing the output of ``string join0`` as an *argument* to a command (via a :ref:`command substitution <expand-command-substitution>`) won't actually work. Fish will pass the correct bytes along, but the command won't be able to tell where the argument ends. This is a limitation of Unix' argument passing.
+``-q`` or ``--quiet``: Do not print the strings, only set the exit status as described above.
+
+**WARNING**: If you omit the ``--`` before the strings to be joined, any strings starting with
```
krobelus, line 29: I'm not sure if the first letter after a colon is usually capitalized in English.
Not sure what our convention is but lowercase "If" would be my preference.
Our commit log is inconsistent..

same applies elsewhere in this patch

doc_src/cmds/string-join.rst
```diff
@@ -18,11 +18,26 @@ Description
...
+``-`` will be treated as flag arguments, meaning they will most likely result in the command
+failing.
+This is also true if you specify a variable which expands to such a string instead of a literal string.
+If you don't need to append flag arguments at the end of the command,
+just always use ``--`` to avoid unwelcome surprises.
```
krobelus, line 34: this is a general problem but yeah, it's probably fine to include it here more prominently

doc_src/cmds/string.rst
```diff
@@ -10,8 +10,8 @@ Synopsis
 
     string collect [-a | --allow-empty] [-N | --no-trim-newlines] [STRING ...]
     string escape [-n | --no-quoted] [--style=] [STRING ...]
-    string join [-q | --quiet] [-n | --no-empty] SEP [STRING ...]
-    string join0 [-q | --quiet] [STRING ...]
+    string join [-q | --quiet] [-n | --no-empty] [--] SEP [STRING ...]
+    string join0 [-q | --quiet] [-n | --no-empty] [--] [STRING ...]
```
krobelus, line 14: I'm not sure if "join" is special enough to justify the inconsistency here; I'd either include it in all synopses or none, preferably none I think.

Maybe same for the other synopsis.
That would imply something like changing "If you omit the -- before the strings ... "
with "Insert a "--" before positional arguments to prevent them from being interpreted as flags..."
just an idea..

doc_src/cmds/string-join.rst
danielrainer: Yeas, this style does not seem to be commonly used in the fish docs. I like this style because it makes it easier to find the description of a particular flag, which is reinforced by this style being so common in other man pages.

doc_src/cmds/string.rst
danielrainer: I removed the `--` from `string.rst` and updated the wording of the warning.

PR #11511, opened 2025-05-19
Remove `libc::sem_t` completely on non-linux

Berrysoft:
## Description

As it's only used on Linux, we can cfg it out completely on other platforms. It also enables `test_topic_monitor_torture` on NetBSD & Cygwin.

## TODOs:


krobelus:
Thanks! I'd prefer if you'd put the PR description in the commit message because the motivation is helpful

PR #11515, opened 2025-05-19
Use uninit instead of zeroed

Berrysoft:
## Description

A refactor PR, try my best to use `MaybeUninit` instead of `zeroed()`.


Berrysoft:
Well... It seems that CI is not passing in this PR.

PR #11526, opened 2025-05-24
Add __fish_cached function

fabiojb:
## Description

Add `__fish_cached` function. This function allows to cache a command output for a certain amount of time.

Continued work from #11480.

### DIFF
--- a/share/completions/flatpak.fish
+++ b/share/completions/flatpak.fish
@@ -58,29 +58,12 @@ function __fish_flatpak
     flatpak $argv | string replace -rf '^([^A-Z].*?)(?: +|\t)(.*?)\s*$' '$1\t$2'
 end
 
-function __fish_print_flatpak_packages
-    set -l xdg_cache_home (__fish_make_cache_dir)
-    or return
-    set -l cache_file $xdg_cache_home/flatpak
-    if test -f $cache_file
-        cat $cache_file
-        set -l age (path mtime -R -- $cache_file)
-        set -l max_age 250
-        if test $age -lt $max_age
-            return
-        end
-    end
-    __fish_cache_put $cache_file
-    __fish_flatpak remote-ls --columns=application,name >$cache_file &
-    return 0
-end
-
 complete -f -c flatpak -n "__fish_seen_subcommand_from run" -a "(__fish_flatpak list --app --columns=application,name)"
 complete -f -c flatpak -n "__fish_seen_subcommand_from info uninstall" -a "(__fish_flatpak list --columns=application,name)"
 complete -f -c flatpak -n "__fish_seen_subcommand_from enter kill" -a "(__fish_flatpak ps --columns=instance,application)"
 complete -f -c flatpak -n "__fish_seen_subcommand_from remote-info remote-ls remote-modify remote-delete" -a "(__fish_flatpak remotes --columns=name,title)"
 
-complete -c flatpak -n '__fish_seen_subcommand_from install' -xa "(__fish_print_flatpak_packages)"
+complete -c flatpak -n '__fish_seen_subcommand_from install' -xa "(__fish_cached -t 250 -- '__fish_flatpak remote-ls --columns=application,name')"
 
 # Plenty of the other stuff is too free-form to complete (e.g. remote-add).
 complete -f -c flatpak -s h -l help
--- a/share/functions/__fish_cached.fish
+++ b/share/functions/__fish_cached.fish
@@ -1,0 +1,36 @@
+function __fish_cached --description "Cache the command output for a given amount of time"
+
+    argparse --min-args 1 --max-args 1 --stop-nonopt 't/max-age=!_validate_int --min 0' 'k/cache-key=' -- $argv
+    or return
+
+    if set -q _flag_cache_key
+        set -f cache_key $_flag_cache_key
+    else
+        set -f cache_key (string trim "$argv" | string split ' ')[1]
+    end
+    if not string match -q --regex '^[\w+-]+$' -- "$cache_key"
+        return 1
+    end
+
+    if set -q _flag_max_age
+        set -f max_age $_flag_max_age
+    else
+        set -f max_age -1
+    end
+    set -l cache_dir (__fish_make_cache_dir)
+    or return
+    set -l cache_file (path normalize $cache_dir/$cache_key)
+    set -l cache_age (path mtime --relative $cache_file)
+
+    if not test -f $cache_file
+        or test $cache_age -gt $max_age
+
+        __fish_cache_put $cache_file
+        if not eval "$argv" >$cache_file
+            rm $cache_file 2>/dev/null
+            return 126
+        end
+    end
+
+    cat $cache_file
+end
--- a/share/functions/__fish_cached.fish
+++ b/share/functions/__fish_cached.fish
@@ -23,14 +23,22 @@ function __fish_cached --description "Cache the command output for a given amoun
     set -l cache_age (path mtime --relative $cache_file)
 
     if not test -f $cache_file
-        or test $cache_age -gt $max_age
-
         __fish_cache_put $cache_file
-        if not eval "$argv" >$cache_file
-            rm $cache_file 2>/dev/null
-            return 126
+        sh -c "{ $argv; } >$cache_file || rm $cache_file 2>/dev/null" &
+
+        if test -n "$last_pid"
+            # wait for at most 1 second if supported
+            command --search waitpid &>/dev/null
+            and waitpid --exited --timeout 1 $last_pid
+            and test -f $cache_file
+            and cat $cache_file
+        end
+    else
+        cat $cache_file
+
+        if test $cache_age -gt $max_age
+            __fish_cache_put $cache_file
+            sh -c "{ $argv; } >$cache_file || rm $cache_file 2>/dev/null" &
         end
     end
-
-    cat $cache_file
 end
--- a/share/functions/__fish_print_pacman_packages.fish
+++ b/share/functions/__fish_print_pacman_packages.fish
@@ -5,22 +5,8 @@ function __fish_print_pacman_packages
     argparse i/installed -- $argv
     or return 1
 
-    set -l xdg_cache_home (__fish_make_cache_dir)
-    or return
-
     if not set -q _flag_installed
-        set -l cache_file $xdg_cache_home/pacman
-        if test -f $cache_file
-            cat $cache_file
-            set -l age (path mtime -R -- $cache_file)
-            set -l max_age 250
-            if test $age -lt $max_age
-                return
-            end
-        end
-        __fish_cache_put $cache_file
-        # prints: <package name>	Package
-        pacman -Ssq | sed -e 's/$/\t'Package'/' >$cache_file &
+        __fish_cached -t 250 -- "pacman -Ssq | sed -e 's/\$/\\tPackage/'"
         return 0
     else
         pacman -Q | string replace ' ' \t
--- a/share/functions/__fish_print_port_packages.fish
+++ b/share/functions/__fish_print_port_packages.fish
@@ -1,24 +1,9 @@
 function __fish_print_port_packages
     type -q -f port || return 1
 
-    # port needs caching, as it tends to be slow
-    set -l xdg_cache_home (__fish_make_cache_dir)
-    or return
-
-    set -l cache_file $xdg_cache_home/port
-    if test -f $cache_file
-        cat $cache_file
-        set -l age (path mtime -R -- $cache_file)
-        set -l max_age 250
-        if test $age -lt $max_age
-            return
-        end
-    end
-
-    __fish_cache_put $cache_file
-    # Remove trailing whitespace and pipe into cache file
-    printf "all\ncurrent\nactive\ninactive\ninstalled\nuninstalled\noutdated" >$cache_file
-    port echo all | awk '{$1=$1};1' >>$cache_file &
-    cat $cache_file
+    __fish_cached -t 250 -k port '
+    	printf "all\ncurrent\nactive\ninactive\ninstalled\nuninstalled\noutdated\n"
+    	port echo all | awk \'{$1=$1};1\'
+    '
     return 0
 end
--- a/share/functions/__fish_print_rpm_packages.fish
+++ b/share/functions/__fish_print_rpm_packages.fish
@@ -5,44 +5,17 @@ function __fish_print_rpm_packages
     argparse i/installed -- $argv
     or return 1
 
-    set -l xdg_cache_home (__fish_make_cache_dir)
-    or return
-
     if type -q -f /usr/share/yum-cli/completion-helper.py
-        # If the cache is less than six hours old, we do not recalculate it
-        set -l cache_file $xdg_cache_home/yum
-        if test -f $cache_file
-            cat $cache_file
-            set -l age (path mtime -R -- $cache_file)
-            set -l max_age 21600
-            if test $age -lt $max_age
-                return
-            end
-        end
-
-        __fish_cache_put $cache_file
         # Remove package version information from output and pipe into cache file
-        /usr/share/yum-cli/completion-helper.py list all -d 0 -C | sed "s/\..*/\tPackage/" >$cache_file &
+        __fish_cached -t 21600 -k yum-completion-helper -- '/usr/share/yum-cli/completion-helper.py list all -d 0 -C | sed "s/\\..*/\\tPackage/"'
         return
     end
 
     # Rpm is too slow for this job, so we set it up to do completions
     # as a background job and cache the results.
     if type -q -f rpm
-        # If the cache is less than five minutes old, we do not recalculate it
-        set -l cache_file $xdg_cache_home/rpm
-        if test -f $cache_file
-            cat $cache_file
-            set -l age (path mtime -R -- $cache_file)
-            set -l max_age 250
-            if test $age -lt $max_age
-                return
-            end
-        end
-
         # Remove package version information from output and pipe into cache file
-        __fish_cache_put $cache_file
-        rpm -qa | sed -e 's/-[^-]*-[^-]*$/\t'Package'/' >$cache_file &
+        __fish_cached -t 250 -- "rpm -qa | sed -e 's/-[^-]*-[^-]*\$/\\t'Package'/'"
         return
     end
 end
--- a/po/de.po
+++ b/po/de.po
@@ -8118,6 +8118,9 @@ msgstr "Testergebnisse in Datei config.cache zwischenspeichern"
 msgid "Cache test results in specified file"
 msgstr "Testergebnisse in angegebener Datei zwischenspeichern"
 
+msgid "Cache the command output for a given amount of time"
+msgstr ""
+
 msgid "Caddyfile to load"
 msgstr ""
 
--- a/po/en.po
+++ b/po/en.po
@@ -8116,6 +8116,9 @@ msgstr "Cache test results in file config.cache"
 msgid "Cache test results in specified file"
 msgstr "Cache test results in specified file"
 
+msgid "Cache the command output for a given amount of time"
+msgstr ""
+
 msgid "Caddyfile to load"
 msgstr ""
 
--- a/po/fr.po
+++ b/po/fr.po
@@ -8217,6 +8217,9 @@ msgstr "Mettre en cache les résultats des tests dans le fichier config.cache"
 msgid "Cache test results in specified file"
 msgstr "Mettre en cache les résultats des tests dans le fichier spécifié"
 
+msgid "Cache the command output for a given amount of time"
+msgstr ""
+
 msgid "Caddyfile to load"
 msgstr ""
 
--- a/po/pl.po
+++ b/po/pl.po
@@ -8112,6 +8112,9 @@ msgstr ""
 msgid "Cache test results in specified file"
 msgstr ""
 
+msgid "Cache the command output for a given amount of time"
+msgstr ""
+
 msgid "Caddyfile to load"
 msgstr ""
 
--- a/po/pt_BR.po
+++ b/po/pt_BR.po
@@ -8117,6 +8117,9 @@ msgstr "Cache test results in file config.cache"
 msgid "Cache test results in specified file"
 msgstr ""
 
+msgid "Cache the command output for a given amount of time"
+msgstr ""
+
 msgid "Caddyfile to load"
 msgstr ""
 
--- a/po/sv.po
+++ b/po/sv.po
@@ -8115,6 +8115,9 @@ msgstr "Cachea test resultat i filen config.cache"
 msgid "Cache test results in specified file"
 msgstr "Cache:a resultat i angiven fil"
 
+msgid "Cache the command output for a given amount of time"
+msgstr ""
+
 msgid "Caddyfile to load"
 msgstr ""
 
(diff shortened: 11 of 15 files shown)


share/functions/__fish_cached.fish
```diff
@@ -0,0 +1,33 @@
+function __fish_cached --description "Cache the command output for a given amount of time"
+
+    argparse --min-args 1 --stop-nonopt 't/max_age=!_validate_int --min 0' 'k/cache_key=' -- $argv
```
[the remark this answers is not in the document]
fabiojb: Fixed 1 and 3. 

About 2, inline `__fish_cache_put` will require refactoring some completion functions that I'm not sure if I can test:
* __fish_cache_sourced_completions
* __fish_print_eopkg_packages
* __fish_print_port_packages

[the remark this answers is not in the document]
krobelus: don't worry about testing every single instance; IIRC they're all pretty much exactly the same,
so testing only one should get us pretty far. We'll review it carefully additionally.
I think it would be preferrable to put everything in one commit, since it's one logical change

share/functions/__fish_cached.fish
```diff
@@ -0,0 +1,33 @@
+function __fish_cached --description "Cache the command output for a given amount of time"
+
+    argparse --min-args 1 --stop-nonopt 't/max-age=!_validate_int --min 0' 'k/cache-key=' -- $argv
+    or return
```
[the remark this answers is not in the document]
fabiojb: I tested with `or return` and `or return $status`, and seems them both return status 1 and the error message. Using `return` without explicit status seems to implicitly return the status of the last command.

share/functions/__fish_cached.fish
```diff
@@ -0,0 +1,33 @@
...
+
+    if set -q _flag_cache_key
+        set -f cache_key $_flag_cache_key
+    else
+        set -f cache_key (string split ' ' "$argv" | head -1)
```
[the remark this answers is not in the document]
fabiojb: This would be simpler, but it doesn't work in case where the eval command is quoted (e.g. `__fish_cached 'echo foo | sed "s/\$/\tPackage/"'`)

krobelus:
On Sat, May 31, 2025 at 12:24:26PM -0700, Fabio José Bohnenberger wrote:
> @fabiojb commented on this pull request.
> 
> 
> 
> > @@ -0,0 +1,33 @@
> +function __fish_cached --description "Cache the command output for a given amount of time"
> +
> +    argparse --min-args 1 --stop-nonopt 't/max_age=!_validate_int --min 0' 'k/cache_key=' -- $argv
> 
> Ok, so I'm going to work on that. I'll make separate commits, but we can squash them all together before merging.

As an author I usually like to review exactly the changes to be merged.
I'm role-playing as the person who will stumble upon the changes in future.

We are doing code movement/consolidation with no functional change intended.
By introducing the new function in one commit, and replacing the
old logic in another commit, we are optimizing either for someone who might:
1. check out the first commit to compare the on-disk
   files (with old and new implementation).
2. want to read the diffs from start to finish without needing to
   resolve forward references (e.g. use of a function that's only defined later).
To me it seems more natural (maybe I'm in the minority) to optimize for the person using "git show $(git blame ...)"
which is a more primitive approach than 1, and does require dealing
with forward references but that should be easy given that there is
no intended behavior change.
Multiple "adoption" commits can increase precision of git bisect but I don't think that's needed here.

share/completions/flatpak.fish
```diff
@@ -58,29 +58,12 @@ function __fish_flatpak
...
-    if test -f $cache_file
-        cat $cache_file
-        set -l age (path mtime -R -- $cache_file)
-        set -l max_age 250
-        if test $age -lt $max_age
-            return
-        end
-    end
-    __fish_cache_put $cache_file
-    __fish_flatpak remote-ls --columns=application,name >$cache_file &
```
krobelus, line 74: This removes the backgrounding, meaning we potentially block for 1.5 seconds
when completing "flatpak install " for the first time.

This is reproducible with this command
(the awkward `complete -e` is because on my system vendor_completions.d/flatpak shadows the fish ones..)

	target/debug/fish -c 'complete -e flatpak; source share/completions/flatpak.fish; rm ~/.cache/fish/__fish_flatpak; time complete -C"flatpak install " >/dev/null'

It may be fine to remove backgrounding but "fish doesn't support background function execution" should not be the sole reason.
We could preserve existing behavior with the diff below (needs to be extended to the other call sites of course). What do you think?
The main downside with preserving the existing behavior is that when we have a cached file that is too old, we'll provide those old completions.
Maybe this is not a big problem for our use cases (package managers) because the relevant completions don't change a lot?

```diff
diff --git a/share/completions/flatpak.fish b/share/completions/flatpak.fish
index d0af5a7e30..6d0d12d1b0 100644
--- a/share/completions/flatpak.fish
+++ b/share/completions/flatpak.fish
@@ -63,7 +63,7 @@
 complete -f -c flatpak -n "__fish_seen_subcommand_from enter kill" -a "(__fish_flatpak ps --columns=instance,application)"
 complete -f -c flatpak -n "__fish_seen_subcommand_from remote-info remote-ls remote-modify remote-delete" -a "(__fish_flatpak remotes --columns=name,title)"
 
-complete -c flatpak -n '__fish_seen_subcommand_from install' -xa "(__fish_cached -t 250 -- __fish_flatpak remote-ls --columns=application,name)"
+complete -c flatpak -n '__fish_seen_subcommand_from install' -xa "(__fish_cached -t 250 -- '__fish_flatpak remote-ls --columns=application,name')"
 
 # Plenty of the other stuff is too free-form to complete (e.g. remote-add).
 complete -f -c flatpak -s h -l help
diff --git a/share/functions/__fish_cached.fish b/share/functions/__fish_cached.fish
index 0df79088e7..37266681a6 100644
--- a/share/functions/__fish_cached.fish
+++ b/share/functions/__fish_cached.fish
@@ -8,6 +8,10 @@
     else
         set -f cache_key (string split ' ' "$argv" | head -1)
     end
+    if test (count $argv) -ne 1
+        echo >&2 '__fish_cached expected exactly one shell command argument'
+        return 1
+    end
     if not string match -q --regex '^[\w+-]+$' "$cache_key"
         return 1
     end
@@ -21,15 +25,14 @@
     set -l cache_file (path normalize $cache_dir/$cache_key)
     set -l cache_age (path mtime --relative $cache_file)
 
+    if test -f $cache_file
+        cat $cache_file
+    end
+
     if not test -f $cache_file
         or test $cache_age -gt $max_age
 
         __fish_cache_put $cache_file
-        if not eval "$argv" >$cache_file
-            rm $cache_file 2>/dev/null
-            return 126
-        end
+        sh -c "{ $argv[1]; } >$cache_file || rm $cache_file 2>/dev/null" &
     end
-
-    cat $cache_file
 end
```

share/functions/__fish_cached.fish
```diff
@@ -0,0 +1,35 @@
...
```
krobelus: Note: an idiomatic alternative would be:

	set -f cache_key (string split ' ' "$argv")[1]

because that avoids the need for a fork / new process.
No big deal though, since this is not called often.

share/functions/__fish_cached.fish
```diff
@@ -0,0 +1,35 @@
...
+    end
+    if not string match -q --regex '^[\w+-]+$' "$cache_key"
```
krobelus: technically we need to do `string match ... -- "$cache_key"` in case the cache key starts with `-`.

share/functions/__fish_cached.fish
```diff
@@ -0,0 +1,35 @@
...
+        return 1
+    end
+
+    if set -q _flag_max_age
+        set -f max_age $_flag_max_age
+    else
+        set -f max_age -1
+    end
+    set -l cache_dir (__fish_make_cache_dir)
```
krobelus, line 20: some callers used to have "or return", in case "mkdir" fails; we should probably keep that,
because otherwise we get redundant errors complaining the directory doesn't exist.
It's probably fine to print no completions in that case since it should be a really rare scenario that better be fixed immediately.

share/functions/__fish_cached.fish
```diff
@@ -0,0 +1,35 @@
...
+    set -l cache_file (path normalize $cache_dir/$cache_key)
+    set -l cache_age (path mtime --relative $cache_file)
+
+    if not test -f $cache_file
+        or test $cache_age -gt $max_age
+
+        __fish_cache_put $cache_file
+        if not eval "$argv" >$cache_file
```
krobelus: I can't think of any fish command that has the `eval "$argv"` semantics.
I think "ssh" does something like that, that has confused me many times.

Specifically, if a caller does `__fish_cache_put echo '$PATH'`,
the might expect the result to be `$PATH` and not the expanded value of $PATH,
based on the assumption that `__fish_cache_put` can delegate to `$argv`, not `eval "$argv"` (i.e. how "sudo" works).

If we can get by without eval, we should do that.
But it looks like there are good reasons we need it -- assuming we want both backgrounding and the ability to use multiple processess such as `echo foo; echo bar` to populate the cache.

If we keep eval, I think we should accept only one positional argument.
Then the semantics will be much clearer.
IMO, it's fairly obvious what either of

	__fish_cache_put 'echo $PATH'
	__fish_cache_put 'echo \'$PATH\''

are supposed to do.

share/functions/__fish_print_pacman_packages.fish
```diff
@@ -5,22 +5,8 @@ function __fish_print_pacman_packages
...
-        if test -f $cache_file
-            cat $cache_file
-            set -l age (path mtime -R -- $cache_file)
-            set -l max_age 250
-            if test $age -lt $max_age
-                return
-            end
-        end
-        __fish_cache_put $cache_file
-        # prints: <package name>	Package
```
krobelus, line 22: `pacman -Ssq` is so fast (<200ms here) that it could definitely get by without the backgrounding;
but it doesn't hurt either (and it may be slower, depending on network)

share/functions/__fish_print_pacman_packages.fish
```diff
@@ -5,22 +5,8 @@ function __fish_print_pacman_packages
...
-        pacman -Ssq | sed -e 's/$/\t'Package'/' >$cache_file &
+        __fish_cached -t 250 -- pacman -Ssq | sed -e 's/$/\tPackage/'
```
krobelus: Note that this commit changes the format of the cache file (to no longer include the \tPackage).
This would cause problems if new code would accept an old cache file.
In this case, "Package" would show up in the description twice.
Given that our max age is 250 seconds, the blast radius is pretty limited.. so I guess we can take it; but of course it would be doable to run something liek.

	__fish_cached -t 250 -- 'pacman -Ssq | sed -e \'s/$/\tPackage/\''

Alternatively, we could pick a different name for the cache file, that would be even easier.

share/functions/__fish_print_port_packages.fish
```diff
@@ -1,24 +1,6 @@
...
-            return
-        end
-    end
-
-    __fish_cache_put $cache_file
-    # Remove trailing whitespace and pipe into cache file
-    printf "all\ncurrent\nactive\ninactive\ninstalled\nuninstalled\noutdated" >$cache_file
-    port echo all | awk '{$1=$1};1' >>$cache_file &
-    cat $cache_file
+    __fish_cached -t 250 -- port echo all | awk '{$1=$1};1'
```
krobelus: I think you missed the `printf "all\ncurrent\nactive\ninactive\ninstalled\nuninstalled\noutdated"` part.
How about

	__fish_cached -t 250 '
		printf "all\ncurrent\nactive\ninactive\ninstalled\nuninstalled\noutdated"
		port echo all | awk \'{$1=$1};1\'
	'

share/functions/__fish_print_rpm_packages.fish
```diff
@@ -5,44 +5,17 @@ function __fish_print_rpm_packages
...
-            set -l max_age 21600
-            if test $age -lt $max_age
-                return
-            end
-        end
-
-        __fish_cache_put $cache_file
         # Remove package version information from output and pipe into cache file
-        /usr/share/yum-cli/completion-helper.py list all -d 0 -C | sed "s/\..*/\tPackage/" >$cache_file &
+        __fish_cached -t 21600 -k yum-completion-helper -- /usr/share/yum-cli/completion-helper.py list all -d 0 -C | sed "s/\..*/\tPackage/"
```
krobelus: Nice, this uses a different file name so there will be no forward compatibility problem

share/functions/__fish_cached.fish
fabiojb: Done

share/functions/__fish_print_pacman_packages.fish
fabiojb: Done (wrapped the cached command with `''`)

share/functions/__fish_print_port_packages.fish
```diff
@@ -1,24 +1,9 @@
...
-            return
-        end
-    end
-
-    __fish_cache_put $cache_file
-    # Remove trailing whitespace and pipe into cache file
-    printf "all\ncurrent\nactive\ninactive\ninstalled\nuninstalled\noutdated" >$cache_file
-    port echo all | awk '{$1=$1};1' >>$cache_file &
-    cat $cache_file
+    __fish_cached -t 250 -k port '
```
fabiojb, line 4: Applied the suggested change here. To support multi line strings in the command argument, I added a `string trim` in the cache_key definition

share/functions/__fish_cached.fish
fabiojb: Done

share/functions/__fish_cached.fish
fabiojb: Done

share/functions/__fish_print_pacman_packages.fish
fabiojb: Done

share/functions/__fish_print_port_packages.fish
fabiojb: Done

share/functions/__fish_cached.fish
fabiojb: > If we keep eval, I think we should accept only one positional argument.
> Then the semantics will be much clearer.

I agree with that. Doing so we also avoid confusion when using pipes, making clearer when the pipe is executed in the argument command or outside of it.

The only drawback I can see is that escaping strings inside the command  argument string could be a little tricky in more complex cases (e.g [xbps-query completion](https://github.com/fish-shell/fish-shell/pull/11526/files#diff-d8ab4bba75a0ec3691e7c4798588d0ff6cadbfa8f22a24da672f8495fb461fc3R10))

share/completions/flatpak.fish
fabiojb: > It may be fine to remove backgrounding but "fish doesn't support background function execution" should not be the sole reason.

The main reason IMO is having a more predictable result. It feels confusing pressing tab and having no results (except for a slight blink from the terminal cursor) and then pressing tab again one or more times to actually have completion options to show up. In normal cases this will occur only one time, but the `.cache` folder could be deleted or even mapped to a temporary dir (tmpfs), making the behavior of pressing tab and not having any result to occur more often.

Also it can lead to odd behaviors, like in `pacman -S <TAB>` that `complete` match more than one subset of completion options (generated from different commands). When there is no cache file, the completion return only "package group names", and the second completion (after the cache file is created) returns "package group names" + "package names".

Maybe holding tab or pressing many times, when there is no cache file yet, could trigger the cached command multiple times, especially for commands that take more time to execute.

But I also agree that not having to block the completion every time the cache is expired could provide a better experience. 

Some others solutions I was thinking:
1. **A mix of blocking and not blocking:** Block if there is no cache file, after the cache file is created, the next calls where the cache is expired could return immediately the stale cache content and trigger the update command in background.
2. **Blocking with  timeout:** Always block when the cache file doesn't exist or is expired, but limit the time it can be blocked (using `timeout` maybe), after that, return to continue the shell execution, but also continue to execute the command that update the cache in background.

The last one I don't event know if its something doable, and maybe could be overkill in that case.

I will l try those solutions, and in case of not having success I think we could proceed with the background always solution.

fabiojb: > (the awkward complete -e is because on my system vendor_completions.d/flatpak shadows the fish ones..)

btw, same problem here. The fish completions shipped with the flatpak package I have installed seems to be overriding fish built-in completions

krobelus: > It feels confusing pressing tab and having no results (except for a slight blink from the terminal cursor) and then pressing tab again

Right, I agree that at least when there is nothing cached at all, we should block; at least for a second or so.

> Also it can lead to odd behaviors, like in pacman -S <TAB>

I think for pacman we should simply block always (as your PR does), because it's fast enough.
For the rest, both options 1 and 2 sound fine by me.
This probably means that we want `__fish_cached --background` option or similar, to select a behavior.

If the `timeout` solution is tricky, maybe it's not worth spending too much time on it because non-blocking completions should obselete this.. but I'm not sure how long it will take to implement those.

share/functions/__fish_print_port_packages.fish
krobelus: FWIW if this was a user-facing function, I'd probably make the explicit cache key argument mandatory.. maybe we should do that regardless, no strong opinion here

share/functions/__fish_cached.fish
```diff
@@ -0,0 +1,36 @@
...
+            rm $cache_file 2>/dev/null
+            return 126
```
krobelus: returning 126 (as opposed to the usual 1) is weird given that we don't check the value anywhere I don't think

share/completions/flatpak.fish
fabiojb: Updated the MR with a solution that blocks for at most 1 second when no cache file exists and immediately return the cached content when the cache file is present, updating in background when necessary.

This new version worked well in my tests (completions for _pacman_ and _winetricks_ commands), but I noticed that fish functions don't execute in background, so in case of `__fish_cached_` is called passing a function in the command argument, it would block until this command is finished. This happens in the current [flatpak completion](https://github.com/fish-shell/fish-shell/pull/11526/files#diff-b527fa8236477dced0c6b3bb25c12ee73a73063f9b7877c2ec323645a197a908R66), that calls __fish_cached passing a function as argument.

share/functions/__fish_cached.fish
fabiojb: The idea behind was making a distinction between the argument command execution error and other errors in general, making easy to identify problems when calling `__fish_cached` (at least in the terminal).

But I removed that in the last change.

share/functions/__fish_cached.fish
```diff
@@ -0,0 +1,40 @@
...
+        set -f max_age -1
+    end
+    set -l cache_dir (__fish_make_cache_dir)
+    or return
+    set -l cache_file (path normalize $cache_dir/$cache_key)
+    set -l cache_age (path mtime --relative $cache_file)
+
+    if not test -f $cache_file
+        __fish_cache_put $cache_file
+        eval "$argv >$cache_file &"
```
fabiojb: Possibly we could rework this so that these 2 lines don't need to be repeated on lines 36 and 37, but the current form seems to make easier to understand what is going on here.

share/functions/__fish_cached.fish
krobelus: some places still pass multiple arguments, like `__fish_cached -t 3600 -- winetricks list-all`
I don't think they should; if escaping gets too tricky, we can extract a function (as long as we're gonna do something like `eval $argv[0]` that evaluates arbitrary fish code, else escaping quotes is also fine)

share/completions/flatpak.fish
krobelus: Yeah, it doesn't work for functions, which means that `test -n "$last_pid"` will do the wrong thing.
Also if there are multiple statements, it's a little bit weird (though probably fine) that only the last one gets backgrounded.

That's why I suggested `sh -c "{ $argv[1]; } >$cache_file || rm $cache_file 2>/dev/null" &`;
which always gives us a valid `$last_pid`.
With `sh`, we can't use fish functions, but that may be fine; we can inline them.
We could also do `fish -c`, though I have a slight preference for `sh -c`, because that potentially does less work and is less prone to configuration/user errors.

share/functions/__fish_cached.fish
```diff
@@ -0,0 +1,40 @@
...
+
+        if test -n "$last_pid"
+            waitpid --timeout 1 $last_pid && cat $cache_file
```
krobelus: nice use of `waitpid`.
I don't think we can assume it's installed (FreeBSD [doesn't have it](https://manpage.me/index.cgi?apropos=0&q=waitpid&sektion=1&manpath=FreeBSD+12-CURRENT+and+Ports&arch=default&format=html)),
so let's make sure to not print a noisy error when it's missing.

So either
```fish
if not command -v waitpid 
	wait $last_pid # no waitpid, block forever
else
	waitpid --timeout=1 $last_pid
end && cat $cache_file
```

Or, alternatively, reimplement it in a portable way;
since we already require python, we can do

```fish
python -c '
import os, sys
# TODO error handling
# TODO retry on EINTR
os.waitpid(sys.argv[1])
' -- $last_pid
```

fabiojb: Made changes to use waitpid only if the command exists, falling back to "always in background" if it doesn't. It seems to be simpler, and should be good enough. What do you think?

fabiojb: Also, I double checked and FreeBSD seems to have [waitpid function](https://github.com/freebsd/freebsd-src/blob/main/sys/sys/wait.h#L158) as part of the system libraries ([man page](https://man.freebsd.org/cgi/man.cgi?query=waitpid&apropos=0&sektion=0&manpath=FreeBSD+14.3-RELEASE+and+Ports&arch=default&format=html#end)), but I'm not sure if there is an executable for this function shipped with FreeBSD.

krobelus: > Made changes to use waitpid only if the command exists, falling back to "always in background"

sounds fine to me.

I think we should clean up https://github.com/fish-shell/fish-shell/pull/11526#discussion_r2141753024,
` __fish_cached -t 3600 -- winetricks list-all 2>/dev/null` should probably be
` __fish_cached -t 3600 -- 'winetricks list-all 2>/dev/null'`.
Since this is a private function, I'm fine with the `$last_pid` hack that assumes that the argument creates an external process

share/functions/__fish_cached.fish
fabiojb: Changed every call of `__fish_cached` to pass the argument command as an escaped string.

Also added a validation to permit only one positional argument.

share/functions/__fish_cached.fish
fabiojb: > I think we should clean up

Done

> Since this is a private function, I'm fine with the $last_pid hack that assumes that the argument creates an external process

Also added the `--exited` argument to waitpid, to ensure no errors appear case the background job finishes quickly.

@krobelus  I got a little bit worried, I don't see why waitpid and $last_pid usage could be considered a hack. Please let me know if it's usage is not adequate or there is something that needs to be improved.

share/completions/flatpak.fish
fabiojb: Got it, replaced the `eval` by `sh -c`

share/functions/__fish_cached.fish
krobelus: > I got a little bit worried, I don't see why waitpid and $last_pid usage could be considered a hack

that was just because without `sh -c` it's not super obvious where/if an actual process is created.
If there is no process, then we might wait for a user's `make &>log &` job here, which we shouldn't.
But that's fixed now, LGTM

PR #11547, opened 2025-06-01
Use `LocalizableString` for gettext

danielrainer:
Use `LocalizableString` for gettext

This new wrapper type can be constructed via macros which invoke the
`gettext_extract` proc macro to extract the string literals for PO file
generation.

The type checking enabled by this wrapper should prevent trying to obtain
translations for a string for which none exist.

Because some strings (e.g. for completions) are not defined in Rust, but rather
in fish scripts, the `LocalizableString` type can also be constructed from
non-literals, in which case no extraction happens.
In such cases, it is the programmer's responsibility to only construct the type
for strings which are available for localization.

This approach is a replacement for the `cargo-expand`-based extraction.

When building with the `FISH_GETTEXT_EXTRACTION_FILE` environment variable set,
the `gettext_extract` proc macro will write the messages marked for extraction
to a file in the directory specified by the variable.

Updates to the po files:
- This is the result of running the `update_translations.fish` script using the
  new proc_macro extraction. It finds additional messages compared to the
  `cargo-expand` based approach.
- Messages IDs corresponding to paths are removed. The do not have localizations
  in any language and localizing paths would not make sense. I have not
  investigated how they made it into the po files in the first place.
- Some messages are reordered due to `msguniq` sorting differing from `sort`.

Remove docs about installing `cargo-expand`
These are no longer needed due to the switch to our extraction macro.


zanchey:
Phwoar. I like the approach, though I can't review the implementation.

src/wutil/gettext.rs
```diff
@@ -120,25 +117,105 @@ fn wgettext_impl(text: MaybeStatic) -> &'static wstr {
...
+                    wgettext_impl(MaybeStatic::Local(s))
+                }
+            }
+        }
+    }
+}
+
+impl std::fmt::Display for TranslatableString {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.translate())
```
krobelus: I think this uses `std::fmt::Write` which can't handle UTF-8 data (we would want to use `wcs2string`).
Looks like this Display implementation is only needed for `FLOG(error, translatable_str)`.
It's fine because we already had that behavior before this change.

(Actually `printf/src/arg.rs` might be broken for us because it does `storage.extend(s.chars());`,
e.g. it skips `wcs2string` which is what we need to decode our PUA chars.)

So yeah, we can clean this up later.

src/wutil/gettext.rs
```diff
@@ -120,25 +117,105 @@ fn wgettext_impl(text: MaybeStatic) -> &'static wstr {
...
+/// which ensures that the string gets extracted for translation.
+#[macro_export]
+macro_rules! translatable_str {
+    ($string:literal) => {
+        $crate::wutil::gettext::TranslatableString::Static(widestring::utf32str!(
+            fish_gettext_extraction::gettext_extract!($string)
+        ))
+    };
 }
+pub use translatable_str;
```
krobelus: Technically, "localizable" would be an appropriate generalization of "translatable",
because these strings also include things like decimal/thousands separation,
but I guess that's not super relevant so we can stick to the "translation" terminology.

src/wutil/gettext.rs
```diff
@@ -120,25 +117,105 @@ fn wgettext_impl(text: MaybeStatic) -> &'static wstr {
...
+/// Macro for declaring string consts which should be localized.
+#[macro_export]
+macro_rules! translatable_consts {
+    (
+        $(
+            $(#[$attr:meta])*
+            $vis:vis
+            $name:ident
+            $string:literal
+        )*
```
krobelus, line 195: I didn't expect the repetition here but I guess I don't see any problems with this.
LSP go-to-definition still works fine.

krobelus (approved):

src/wutil/gettext.rs
danielrainer: Good point. To fix that we would have to stop using `&str` in the `FLOG` implementation.

src/wutil/gettext.rs
danielrainer: Yes, "localizable" is the better term here. I actually started by calling it "localized", but found that naming confusing and switched to "translatable". For consistency, I changed usage to "localizable" now.

src/wutil/gettext.rs
danielrainer: Originally I had a macro for defining a single constant, but most files defined multiple constants which should be localized, so I thought we might as well have a macro for defining multiple consts at once. I also observed that it works well with rust-analyzer (unlike some other macro stuff).

src/complete.rs
```diff
@@ -1394,7 +1377,7 @@ impl<'ctx> Completer<'ctx> {
                                 if o.result_mode.force_files {
                                     has_force = true;
                                 }
-                                self.complete_from_args(s, &o.comp, o.localized_desc(), o.flags);
+                                self.complete_from_args(s, &o.comp, o.desc.localize(), o.flags);
```
krobelus, line 1380: Since there is nothing else one can do with a localized string,
maybe we should implement `Deref<Target=wstr>` to hide
some explicit calls to `localize()`.
Unfortunately 
1. we can't do it for `to_owned()` due to conflicting trait implementations
   (there is an `impl<T> ToOwned for T where T: Clone;`).
2. `Deref` can't return a static reference, so we still need localize()
    in places where we want to return a static reference.

but maybe it's a step in the right direction?

```diff
diff --git a/src/bin/fish.rs b/src/bin/fish.rs
index e919f3ed60..f399ed9ef0 100644
--- a/src/bin/fish.rs
+++ b/src/bin/fish.rs
@@ -319,9 +319,8 @@
                 // A little extra space.
                 name_width += 2;
                 for cat in cats.iter() {
-                    let desc = cat.description.localize();
                     // this is left-justified
-                    printf!("%-*ls %ls\n", name_width, cat.name, desc);
+                    printf!("%-*ls %ls\n", name_width, cat.name, &cat.description);
                 }
                 return ControlFlow::Break(0);
             }
diff --git a/src/builtins/functions.rs b/src/builtins/functions.rs
index 382793474e..2d78b551db 100644
--- a/src/builtins/functions.rs
+++ b/src/builtins/functions.rs
@@ -244,12 +244,11 @@
 
             let desc = match props.as_ref() {
                 Some(p) => {
-                    let localized_description = p.description.localize();
-                    if localized_description.is_empty() {
+                    if p.description.is_empty() {
                         L!("").to_owned()
                     } else {
                         escape_string(
-                            localized_description,
+                            &p.description,
                             EscapeStringStyle::Script(
                                 EscapeFlags::NO_PRINTABLES | EscapeFlags::NO_QUOTED,
                             ),
diff --git a/src/builtins/shared.rs b/src/builtins/shared.rs
index 2ca4510b48..39800c5f26 100644
--- a/src/builtins/shared.rs
+++ b/src/builtins/shared.rs
@@ -1039,9 +1039,11 @@
 /// Strings not present in our repo would require a custom MO file for translation to be possible.
 fn builtin_gettext(_parser: &Parser, streams: &mut IoStreams, argv: &mut [&wstr]) -> BuiltinResult {
     for arg in &argv[1..] {
-        streams.out.append(
-            crate::wutil::LocalizableString::from_external_source((*arg).to_owned()).localize(),
-        );
+        streams
+            .out
+            .append(crate::wutil::LocalizableString::from_external_source(
+                (*arg).to_owned(),
+            ));
     }
     Ok(SUCCESS)
 }
diff --git a/src/complete.rs b/src/complete.rs
index a7c0ba99a4..1690a2f6d7 100644
--- a/src/complete.rs
+++ b/src/complete.rs
@@ -1310,7 +1310,7 @@
                                 }
                                 let (arg_prefix, arg) = s.split_once(arg_offset);
                                 let first_new = self.completions.completions.len();
-                                self.complete_from_args(arg, &o.comp, o.desc.localize(), o.flags);
+                                self.complete_from_args(arg, &o.comp, &o.desc, o.flags);
                                 for compl in &mut self.completions.completions[first_new..] {
                                     if compl.replaces_token() {
                                         compl.completion.insert_utfstr(0, arg_prefix);
@@ -1341,7 +1341,7 @@
                             if o.result_mode.force_files {
                                 has_force = true;
                             }
-                            self.complete_from_args(s, &o.comp, o.desc.localize(), o.flags);
+                            self.complete_from_args(s, &o.comp, &o.desc, o.flags);
                         }
                     }
 
@@ -1377,7 +1377,7 @@
                                 if o.result_mode.force_files {
                                     has_force = true;
                                 }
-                                self.complete_from_args(s, &o.comp, o.desc.localize(), o.flags);
+                                self.complete_from_args(s, &o.comp, &o.desc, o.flags);
                             }
                         }
                     }
@@ -1400,7 +1400,7 @@
                 if o.option.is_empty() {
                     use_files &= !o.result_mode.no_files;
                     has_force |= o.result_mode.force_files;
-                    self.complete_from_args(s, &o.comp, o.desc.localize(), o.flags);
+                    self.complete_from_args(s, &o.comp, &o.desc, o.flags);
                 }
 
                 if !use_switches || s.is_empty() {
@@ -1433,12 +1433,11 @@
                         }
                     }
                     // It's a match.
-                    let desc = o.desc.localize();
                     // Append a short-style option
-                    if !self
-                        .completions
-                        .add(Completion::with_desc(o.option.clone(), desc.to_owned()))
-                    {
+                    if !self.completions.add(Completion::with_desc(
+                        o.option.clone(),
+                        o.desc.localize().to_owned(),
+                    )) {
                         return false;
                     }
                 }
@@ -2414,7 +2413,7 @@
         CompleteOptionType::DoubleLong => append_switch_short_arg(&mut out, 'l', &o.option),
     }
 
-    append_switch_short_arg(&mut out, 'd', o.desc.localize());
+    append_switch_short_arg(&mut out, 'd', &o.desc);
     append_switch_short_arg(&mut out, 'a', &o.comp);
     for c in &o.conditions {
         append_switch_short_arg(&mut out, 'n', c);
diff --git a/src/fds.rs b/src/fds.rs
index c54b0568fd..8d03169c27 100644
--- a/src/fds.rs
+++ b/src/fds.rs
@@ -145,7 +145,7 @@
             pipes
         }
         Err(err) => {
-            FLOG!(warning, PIPE_ERROR.localize());
+            FLOG!(warning, &PIPE_ERROR);
             perror("pipe2");
             return Err(err);
         }
diff --git a/src/function.rs b/src/function.rs
index 963f221510..8a3face09f 100644
--- a/src/function.rs
+++ b/src/function.rs
@@ -419,7 +419,7 @@
     /// Note callers must provide the function name, since the function does not know its own name.
     pub fn annotated_definition(&self, name: &wstr) -> WString {
         let mut out = WString::new();
-        let desc = self.description.localize();
+        let desc = &self.description;
         let def = get_function_body_source(self);
         let handlers = event::get_function_handlers(name);
 
diff --git a/src/wutil/gettext.rs b/src/wutil/gettext.rs
index 550e85254a..502f0e5ae9 100644
--- a/src/wutil/gettext.rs
+++ b/src/wutil/gettext.rs
@@ -164,6 +164,18 @@
     }
 }
 
+impl std::ops::Deref for LocalizableString {
+    type Target = wstr;
+    fn deref(&self) -> &Self::Target {
+        self.localize()
+    }
+}
+
+impl AsRef<wstr> for LocalizableString {
+    fn as_ref(&self) -> &wstr {
+        self
+    }
+}
 impl std::fmt::Display for LocalizableString {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         write!(f, "{}", self.localize())
```

src/function.rs
```diff
@@ -28,7 +28,7 @@ pub struct FunctionProperties {
     pub named_arguments: Vec<WString>,
 
     /// Description of the function.
-    pub description: WString,
+    pub description: LocalizableString,
```
krobelus, line 31: The type no longer expresses that this is currently always dynamic,
but that's definitely worth trading in for extra safety

gettext-extraction/src/lib.rs
```diff
@@ -0,0 +1,67 @@
...
+
+fn append_po_entry_to_file(message: &TokenStream, file_name: &OsString) {
+    let mut file = OpenOptions::new()
+        .create(true)
+        .append(true)
+        .open(file_name)
+        .unwrap_or_else(|e| panic!("Could not open file {file_name:?}: {e}"));
+    let message_string = message.to_string();
+    if message_string.contains('\n') {
+        panic!("Gettext strings may not contain unescaped newlines. Unescaped newline found in '{message_string}'")
```
krobelus, line 13: right, this check is still needed until we translate unescaped newlines to valid msgstr entries

src/complete.rs
danielrainer: It does make for somewhat more concise code in several places. I'm not sure if I like hiding the potential gettext call behind a deref.

The docs for both [`AsRef`](https://doc.rust-lang.org/std/convert/trait.AsRef.html) and [`Deref`](https://doc.rust-lang.org/std/ops/trait.Deref.html) indicate that the operation should be cheap, which is of course relative, but I don't think `localize()` is cheap in this context.

AsRef:
> Used to do a cheap reference-to-reference conversion.
>
> This trait is similar to [AsMut](https://doc.rust-lang.org/std/convert/trait.AsMut.html) which is used for converting between mutable references. If you need to do a costly conversion it is better to implement [From](https://doc.rust-lang.org/std/convert/trait.From.html) with type &T or write a custom function.

Deref:
> The same advice applies to both deref traits. In general, deref traits should be implemented if:
>
> 1. a value of the type transparently behaves like a value of the target type;
> 2. the implementation of the deref function is cheap; and
> 3. users of the type will not be surprised by any deref coercion behavior.

krobelus: ok yeah, sounds like this should remain explicit

Issue #11524, opened 2025-05-23
nmcli completions are very slow, blocking user input


KuabeM:
I'm using `nmcli` to configure my network connections. I noticed that auto-completion is very slow and blocks user input, similar to #11347, but only for the first time. For subsequent calls, auto-complete runs without noticeable delays.

Running `fish --profile /tmp/nmcli.prof -c "complete -C'nmcli conne'"` to profile the completion gave me the following:

```
Time	Sum	Command
1340	3805216	> complete -C'nmcli conne'
644	3797040	-> source /usr/share/fish/completions/nmcli.fish
64	11945	--> set -l nmoutput (nmcli -g NAME connection show --active 2>/dev/null)
11881	11881	---> nmcli -g NAME connection show --active 2>/dev/null
52	66	--> set -l cname (string escape -- $nmoutput\t"Active connection")
14	14	---> string escape -- $nmoutput\t"Active connection"
50	11823	--> set -a cname (string escape -- (nmcli -g NAME connection show)\t"Connection")
114	11773	---> string escape -- (nmcli -g NAME connection show)\t"Connection"
11659	11659	----> nmcli -g NAME connection show
92	13391	--> set -l ifname (string escape -- (nmcli -g DEVICE device status)\t"Interface name")
166	13299	---> string escape -- (nmcli -g DEVICE device status)\t"Interface name"
13133	13133	----> nmcli -g DEVICE device status
261	3739593	--> set -l ssid (string escape -- (nmcli -g SSID device wifi list)\t"SSID")
282	3739332	---> string escape -- (nmcli -g SSID device wifi list)\t"SSID"
3739050	3739050	----> nmcli -g SSID device wifi list
119	18320	--> set -l bssid (string escape -- (nmcli -g BSSID device wifi list | string replace --all \\ '')\t"BSSID")
176	18201	---> string escape -- (nmcli -g BSSID device wifi list | string replace --all \\ '')\t"BSSID"
18025	18025	----> nmcli -g BSSID device wifi list | string replace --all \\ ''
11	11	--> set -l nmcli_commands general networking radio connection device agent monitor help
6	6	--> set -l nmcli_general status hostname permissions logging help
4	4	--> set -l nmcli_networking on off connectivity help
3	3	--> set -l nmcli_radio all wifi wwan help
4	4	--> set -l nmcli_connection show up down add modify clone edit delete monitor reload load import export help
[...]

PR #11549, opened 2025-06-02
completions/nmcli: only retrieve ssids and bssids if they are completed

KuabeM:
## Description

Changes the nmcli completions to only retrieve the SSID and BSSID from nmcli if they are actually being completed.

As this is my first PR let me know if I got all the style requirements right and if there are any issues with my approach.

Fixes issue #11524 

## TODOs:

### DIFF
--- a/share/completions/nmcli.fish
+++ b/share/completions/nmcli.fish
@@ -2,8 +2,6 @@ set -l nmoutput '(nmcli -g NAME connection show --active 2>/dev/null)'
 set -l cname "$nmoutput"\t"Active connection"
 set -a cname '(nmcli -g NAME connection show 2>/dev/null)\t"Connection"'
 set -l ifname '(nmcli -g DEVICE device status 2>/dev/null)\t"Interface name"'
-set -l ssid '(nmcli -g SSID device wifi list 2>/dev/null)\t"SSID"'
-set -l bssid '(nmcli -g BSSID device wifi list 2>/dev/null | string replace --all \\\ "")\t"BSSID"'
 
 set -l nmcli_commands general networking radio connection device agent monitor help
 set -l nmcli_general status hostname permissions logging help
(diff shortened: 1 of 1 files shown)


share/completions/nmcli.fish
```diff
@@ -57,7 +69,7 @@ complete -c nmcli -n "contains_seq connection show -- (commandline -xp)" -l orde
 complete -c nmcli -n "contains_seq connection show -- (commandline -xp)" -xa "help $cname" -k
 complete -c nmcli -n "contains_seq connection up -- (commandline -xp) && not contains -- (commandline -xp)[-1] ifname ap" -xa "help $cname" -k
 complete -c nmcli -n "contains_seq connection up -- (commandline -xp) && test (commandline -xp)[-1] = ifname" -xa "$ifname" -k
-complete -c nmcli -n "contains_seq connection up -- (commandline -xp) && test (commandline -xp)[-1] = ap" -xa "$bssid"
+complete -c nmcli -n "contains_seq connection up -- (commandline -xp) && test (commandline -xp)[-1] = ap" -xa "(__fish_nmcli_bssids)"
```
krobelus, line 72: I don't think this change has any functional effect.

The existing

	complete ... -xa "$bssd"

would expand to

	complete ... -xa '(nmcli -g BSSID device wifi list 2>/dev/null | string replace --all \\\ "")\t"BSSID"'

meaning that nmcli is only evaluated when completing "nmcli ... connection up " or similar.

I think you observed the problem that was solved in
9116c617368 (completions/nmcli: Complete at runtime, 2025-01-27).
But that commit has not been included in a release yet:

```
$ git describe --contains --tags 9116c617368
fatal: cannot describe '9116c6173686558ba873f1667e4abd325c394b04'
```

I guess we can cherry-pick it.

KuabeM: Huh, that was unfortunate timing on my side then... Sorry for the noise, I'll close this then!

PR #11550, opened 2025-06-02
More local checks

danielrainer:
## Description

This includes updates to the formatting script and a new script for running checks locally, motivated in part by https://github.com/fish-shell/fish-shell/pull/11545#discussion_r2121331880.

## TODOs:
- [x] Are there additional checks we want to include?
- [ ] Improvements for `test_driver.py` checks: Should we use a cachedir? If so, at which path? Would it make sense to use ctest if it is available?
- [ ] Can we get rid of `cmake/Tests.cmake` and use this script (or parts of it) instead?

### DIFF
--- a/build_tools/check.sh
+++ b/build_tools/check.sh
@@ -1,0 +1,20 @@
+#!/bin/sh
+
+set -ex
+
+RUSTFLAGS='-D warnings'; export RUSTFLAGS
+RUSTDOCFLAGS='-D warnings'; export RUSTDOCFLAGS
+
+repo_root="$(dirname "$0")/.."
+build_dir="$repo_root/target/debug"
+
+PATH="$build_dir:$PATH" "$repo_root/build_tools/style.fish" --all --check
+
+cargo build --workspace --all-targets
+cargo clippy --workspace --all-targets
+cargo test --no-default-features --workspace --all-targets
+cargo test --doc --workspace
+cargo doc --workspace
+
+# TODO: parallelize
+"$repo_root/tests/test_driver.py" "$build_dir"
--- a/build_tools/style.fish
+++ b/build_tools/style.fish
@@ -1,38 +1,53 @@
 #!/usr/bin/env fish
 #
-# This runs C++ files and fish scripts (*.fish) through their respective code
-# formatting programs.
+# This runs Python files, fish scripts (*.fish), and Rust files
+# through their respective code formatting programs.
 #
+# `--all`: Format all eligible files instead of the ones specified as arguments.
+# `--check`: Instead of reformatting, fail if a file is not formatted correctly.
+# `--force`: Proceed without asking if uncommitted changes are detected.
+#            Only relevant if `--all` is specified but `--check` is not specified.
+
 set -l fish_files
 set -l python_files
 set -l rust_files
 set -l all no
 
-if test "$argv[1]" = --all
+argparse all check force -- $argv
+or exit $status
+
+if set -l -q _flag_all
     set all yes
-    set -e argv[1]
+    if set -q argv[1]
+        echo "Unexpected arguments: '$argv'"
+        exit 1
+    end
 end
 
-if set -q argv[1]
-    echo "Unexpected arguments: '$argv'"
-    exit 1
-end
+set -l repo_root (status dirname)/..
 
 if test $all = yes
-    set -l files (git status --porcelain --short --untracked-files=all | sed -e 's/^ *[^ ]* *//')
-    if set -q files[1]
-        echo
-        echo 'You have uncommitted changes. Are you sure you want to restyle?'
-        read -P 'y/N? ' -n1 -l ans
-        if not string match -qi y -- $ans
-            exit 1
+    if not set -l -q _flag_force; and not set -l -q _flag_check
+        # Potential for false positives: Not all fish files are formatted, see the `fish_files`
+        # definition below.
+        set -l relevant_uncommitted_changes (git status --porcelain --short --untracked-files=all | sed -e 's/^ *[^ ]* *//' | grep -E '.*\.(fish|py|rs)$')
+        if set -q relevant_uncommitted_changes[1]
+            for changed_file in $relevant_uncommitted_changes
+                echo $changed_file
+            end
+            echo
+            echo 'You have uncommitted changes (listed above). Are you sure you want to restyle?'
+            read -P 'y/N? ' -n1 -l ans
+            if not string match -qi y -- $ans
+                exit 1
+            end
         end
     end
-    set fish_files share/**.fish
+    set fish_files $repo_root/{benchmarks,build_tools,etc,share}/**.fish
     set python_files {doc_src,share,tests}/**.py
-    set rust_files fish-rust/src/**.rs
 else
-    # Extract just the fish files.
+    # Format the files specified as arguments.
+    set -l files $argv
     set fish_files (string match -r '^.*\.fish$' -- $files)
     set python_files (string match -r '^.*\.py$' -- $files)
     set rust_files (string match -r '^.*\.rs$' -- $files)
@@ -40,37 +55,73 @@ end
 
 set -l red (set_color red)
 set -l green (set_color green)
-set -l blue (set_color blue)
+set -l yellow (set_color yellow)
 set -l normal (set_color normal)
 
-# Run the fish reformatter if we have any fish files.
 if set -q fish_files[1]
     if not type -q fish_indent
-        make fish_indent
-        set PATH . $PATH
+        echo
+        echo $yellow'Could not find `fish_indent` in `$PATH`.'$normal
+        echo
+    else
+        echo === Running "$green"fish_indent"$normal"
+        if set -l -q _flag_check
+            if not fish_indent --check -- $fish_files
+                echo $red"Fish files are not formatted correctly."$normal
+                exit 1
+            end
+        else
+            fish_indent -w -- $fish_files
+        end
     end
-    echo === Running "$green"fish_indent"$normal"
-    fish_indent -w -- $fish_files
 end
 
 if set -q python_files[1]
     if not type -q black
         echo
-        echo Please install "`black`" to style python
+        echo $yellow'Please install `black` to style python'$normal
         echo
     else
-        echo === Running "$blue"black"$normal"
-        black $python_files
+        echo === Running "$green"black"$normal"
+        if set -l -q _flag_check
+            if not black --check $python_files
+                echo $red"Python files are not formatted correctly."$normal
+                exit 1
+            end
+        else
+            black $python_files
+        end
     end
 end
 
-if set -q rust_files[1]
-    if not type -q rustfmt
-        echo
-        echo Please install "`rustfmt`" to style rust
-        echo
+if not cargo fmt --version >/dev/null
+    echo
+    echo $yellow'Please install "rustfmt" to style Rust, e.g. via:'
+    echo "rustup component add rustfmt"$normal
+    echo
+else
+    echo === Running "$green"rustfmt"$normal"
+    if set -l -q _flag_check
+        if set -l -q _flag_all
+            if not cargo fmt --check
+                echo $red"Rust files are not formatted correctly."$normal
+                exit 1
+            end
+        else
+            if set -q rust_files[1]
+                if not rustfmt --check --files-with-diff $rust_files
+                    echo $red"Rust files are not formatted correctly."
+                    exit 1
+                end
+            end
+        end
     else
-        echo === Running "$blue"rustfmt"$normal"
-        rustfmt $rust_files
+        if set -l -q _flag_all
+            cargo fmt
+        else
+            if set -q rust_files[1]
+                rustfmt $rust_files
+            end
+        end
     end
 end
--- a/src/widecharwidth/widechar_width.rs
+++ b/src/widecharwidth/widechar_width.rs
@@ -23,37 +23,31 @@ type R = (u32, u32);
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
 #[repr(u8)]
 pub enum WcWidth {
-  /// The character is single-width
-  One,
-  /// The character is double-width
-  Two,
-  /// The character is not printable.
-  NonPrint,
-  /// The character is a zero-width combiner.
-  Combining,
-  /// The character is East-Asian ambiguous width.
-  Ambiguous,
-  /// The character is for private use.
-  PrivateUse,
-  /// The character is unassigned.
-  Unassigned,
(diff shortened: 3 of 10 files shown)


build_tools/checks.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+
+set -e
+
+repo_root="$(dirname "$0")/.."
```
krobelus: should also run clippy I think. I guess that one is not enforced in the CI today,
but same goes for formatting and warnings in Rustdoc.
I think we're not far from enforcing all checks.

build_tools/checks.sh
```diff
@@ -0,0 +1,17 @@
...
+
+"$repo_root/build_tools/style.fish" --all --check
```
krobelus: I'd prefer using "cargo fmt" rather than listing all Rust files ourselves
(technically a user could have some uncommitted Rust files).

I think we can remove `rustfmt` calls from `style.fish`, and add `cargo fmt` to `checks.sh`.
`cargo fmt` and `rustfmt`-on-save are already widely known, I don't think having it in `style.fish` is really necessary.
Or maybe change the interface of `style.fish` to use `cargo fmt` if no files are specified..

build_tools/checks.sh
```diff
@@ -0,0 +1,17 @@
...
+
+cargo build --workspace --all-targets
```
krobelus: Unrelated: I think `printf/Cargo.toml` should defer to dependency versions from the workspace `Cargo.toml`, to remove some duplication.

build_tools/checks.sh
```diff
@@ -0,0 +1,17 @@
...
+
+cargo test --no-default-features --workspace --all-targets
```
krobelus: what's the `--no-default-features` for?
Something about `--embed-data`?

build_tools/checks.sh
```diff
@@ -0,0 +1,17 @@
...
+cargo test --doc --workspace
+RUSTDOCFLAGS='-D warnings' cargo doc --workspace --all
```
krobelus: `--all` is a deprecated alias for `--workspace`, do we need it for the CI toolchain?

---

So "--all-targets" is "equivalent to specifying --lib --bins --tests --benches --examples".
I think for us only "lib" is relevant. But of course it doesn't hurt
to use this option (and it seems more stable than the alternative). I'm just curious.

build_tools/checks.sh
```diff
@@ -0,0 +1,17 @@
...
+
+# TODO: add cachedir?
```
krobelus: last time I measured, adding a cachedir makes very little difference on runtime, at least for a single test

build_tools/checks.sh
```diff
@@ -0,0 +1,17 @@
...
+# TODO: use ctest if available?
```
krobelus: yeah something like ctest will give a nice speedup (I guess that's the only remaining reason why I use cmake for local testing)

build_tools/checks.sh
```diff
@@ -0,0 +1,17 @@
...
+"$repo_root/tests/test_driver.py" "$repo_root/target/debug"
```
krobelus: we usually use imperative mood for script names, so maybe "check.sh"

build_tools/style.fish
```diff
@@ -1,38 +1,54 @@
...
+            if not string match -qi y -- $ans
+                exit 1
+            end
         end
     end
-    set fish_files share/**.fish
-    set python_files {doc_src,share,tests}/**.py
-    set rust_files fish-rust/src/**.rs
+    set fish_files $repo_root/{benchmarks,build_tools,etc,share}/**.fish
+    set python_files $repo_root/**.py
```
krobelus: not sure but I might keep python filter (`{doc_src,share,tests}`) because globbing the entire tree feels somehow wrong.. ideally our build system doesn't need to glob but has an explicit list of all source files..
I guess we're unlikely to add python files elsewhere

Actually I have these, which I'd prefer not to format:

```
would reformat build_tools/../t/test.py
would reformat build_tools/../t/test3.py
would reformat build_tools/../t/tmp.py
would reformat build_tools/../t/gdb.py
```

build_tools/style.fish
```diff
@@ -43,14 +59,22 @@ set -l green (set_color green)
 set -l blue (set_color blue)
 set -l normal (set_color normal)
 
-# Run the fish reformatter if we have any fish files.
 if set -q fish_files[1]
-    if not type -q fish_indent
-        make fish_indent
-        set PATH . $PATH
+    if not command -v fish_indent
```
krobelus: I guess fish_indent from `PATH` is okay for local runs for now
but eventually we want to use `target/debug/fish_indent` or `target/release/fish_indent`
in case we make changes to the formatting.
If `target/debug/fish_indent` is fast enough (5 seconds here, I guess that's fine), can we pass that one explicitly in `build_tools/check.sh` via `PATH=target/debug:$PATH build_tools/style.fish`

src/widecharwidth/widechar_width.rs
```diff
@@ -23,37 +23,31 @@ type R = (u32, u32);
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
 #[repr(u8)]
 pub enum WcWidth {
-  /// The character is single-width
```
krobelus, line 26: `src/widecharwidth/widechar_width.rs` is generated code.

I think we should only format it once we can ensure that any future updates will be formatted as well.
Maybe exclude it until then.

krobelus:
>  Can we get rid of cmake/Tests.cmake

CMake has the advantage that it knows about dependencies; so it will only build what's necessary for a given target.
This script simply runs everything.

But yeah, Tests.cmake might be mostly obsolete. for most tests, we have the test driver. I didn't check what else is included.

build_tools/checks.sh
danielrainer: That's just copied from `Tests.cmake`. I don't know why it was put there. `embed-data` is the feature that is disabled by this. I think it might make sense to compile with this feature, since it enables testing things which depend on files in `share`. I have not checked in detail how that works at the moment. One thing which might benefit is translations, which require the mo files, although using the embedded versions of those does not work yet. (A proof-of-concept implementation is in #11475.)

build_tools/checks.sh
danielrainer: Ah yes, the `--all` for `cargo doc` is useless here.

The `--all-targets` is mostly there so we don't miss anything by accident, but I agree that it might not do much at the moment. There is also some odd behavior going on with that flag as you already noticed (https://github.com/fish-shell/fish-shell/pull/11545#discussion_r2131981683), so I hope that there are some improvements about that for cargo in the future.

build_tools/checks.sh
danielrainer: > Or maybe change the interface of style.fish to use cargo fmt if no files are specified

I went for this option. (Using `cargo fmt` when `--all` is specified.) I doubt that many users use this script to format explicitly selected files, but I guess it's nice to keep the capability.

build_tools/checks.sh
danielrainer: Ok, then I'll just remove the comment.

build_tools/style.fish
danielrainer: > `PATH=target/debug:$PATH build_tools/style.fish`

That's a decent option, even though the debug version is quite slow. We might consider using a release version at some point.

src/widecharwidth/widechar_width.rs
danielrainer: Ah yes, that should be addressed in https://github.com/ridiculousfish/widecharwidth/. I think excluding it might not be that simple (I don't see an option for excluding single files from `cargo fmt`.) The simplest approach might be to call `rustfmt` on the Rust file generated by `generate.py`. We could add a line for that to the `rust` rule in the Makefile.

danielrainer: I opened https://github.com/ridiculousfish/widecharwidth/pull/36.

danielrainer:
> CMake has the advantage that it knows about dependencies; so it will only build what's necessary for a given target. This script simply runs everything.

That might have been true when the project was C++ based, but most dependency checks are done by cargo now, so I don't see how CMake has a noteworthy advantage there other than maybe building the 2 C files (`src/libc.c`, `tests/fish_test_helper.c`) less frequently.

build_tools/checks.sh
danielrainer: It seems that ctest isn't intended for use without CMake (e.g. https://github.com/microsoft/vscode-cmake-tools/issues/4169).

Maybe we could change `test_driver.py` to run tests in parallel?

build_tools/check.sh
```diff
@@ -0,0 +1,20 @@
...
+
+RUSTFLAGS='-D warnings'; export RUSTFLAGS
+RUSTDOCFLAGS='-D warnings'; export RUSTDOCFLAGS
+
+repo_root="$(dirname "$0")/.."
+build_dir="$repo_root/target/debug"
+
+PATH="$build_dir:$PATH" "$repo_root/build_tools/style.fish" --all --check
+
+cargo build --workspace --all-targets
```
krobelus, line 13: I think build should happen before `style.fish` because we want an up-to-date `fish_indent`.

(Also it would increase transparency if we only exported RUSTFLAGS and RUSTDOCFLAGS where
we need it (which is all the cargo commands I think).
If they're all in one group we could do `(export ...; cargo this; cargo that)`.
If not, one idea would be

```
cargo() {
	RUSTFLAGS=... RUSTDOCFLAGS=... command cargo "$@"
}
```

not very important I guess.

build_tools/style.fish
```diff
@@ -1,76 +1,127 @@
...
+                    echo $red"Rust files are not formatted correctly."
+                    exit 1
+                end
+            end
+        end
     else
-        echo === Running "$blue"rustfmt"$normal"
-        rustfmt $rust_files
+        if set -l -q _flag_all
+            cargo fmt --all
```
krobelus: Looks like even without --all, this formats code in printf/src/*.rs

The docs for -all say 
"Formats all packages and and also their local path-based dependencies"
I'm not sure what the first part means (I guess it's relevant for
source trees with multiple crates that don't define a single workspace).
The second one ("all path-based dependencies") seems surprising, because path-based ones
are usually "../libc", "../nix".. no big deal but I'd rather remove "--all".

build_tools/check.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+
+set -e
```
krobelus: might be useful to add `set -x` here to make it more obvious which step failed.
We would definitely do this in CI (i.e. run `sh -x build_tools/check.sh`)

build_tools/checks.sh
krobelus: yeah I don't see any problem with parallelizing `test_driver.py`.

Alternatively (or "until then"), we can use GNU parallel or even xargs
("xargs -P" is nonstandard but supported by both findutils and [FreeBSD's version](https://manpage.me/?q=xargs).. and we can always fall back to sequential execution).

The main thing that we'd be missing is a list/summary of all failed tests at the end.
So yeah it seems we should integrate parallelism directly into `test_driver.py`.

```
$ time printf %s\n tests/checks/*.fish tests/pexpects/*.py | parallel tests/test_driver.py target/debug
....
tests/pexpects/fg.py..PASS (5352 ms)
tests/pexpects/wait.py..PASS (3339 ms)
tests/pexpects/terminal.py..PASS (4444 ms)
tests/pexpects/signals.py..PASS (4870 ms)
tests/checks/check-all-fish-files.fish..PASS (22854 ms)

Executed in   23.37 secs    fish           external
   usr time  116.81 secs  948.00 micros  116.81 secs
   sys time   36.77 secs    0.00 micros   36.77 secs
```

```
$ time printf %s\n tests/checks/*.fish tests/pexpects/*.py | xargs -P$(nproc 2>/dev/null || sysctl -n hw.ncpu) -n 1 tests/test_driver.py target/debug
tests/pexpects/wait.py..PASS (5638 ms)
tests/pexpects/wildcard_tab.py..PASS (460 ms)
PASS (493 ms)
PASS (12527 ms)
PASS (5639 ms)
PASS (4919 ms)
PASS (1738 ms)
PASS (3008 ms)
PASS (2047 ms)
PASS (6386 ms)
PASS (5350 ms)
PASS (14991 ms)
PASS (3342 ms)
PASS (4505 ms)
PASS (4820 ms)
PASS (22455 ms)

Executed in   22.78 secs    fish           external
   usr time  109.94 secs    1.21 millis  109.94 secs
   sys time   32.25 secs    0.05 millis   32.25 secs
```

krobelus:
LGTM aside from the last minor comments (I can also apply those).
I think once we figure out how to parallelize the test driver, we can get rid fo Tests.cmake (and move the ASan/TSan-specific bits elsewhere).

build_tools/check.sh
danielrainer: Good, idea. Added.

build_tools/checks.sh
danielrainer: That seems like a good workaround until `test_driver.py` is parallelized. The output is quite messy though, so it might make sense to change `test_driver.py` to prevent it from flushing stdout between printing the test name and the result. I have been investigating how parallelization with Python might work (unsuccessfully so far), but in the process I have done some refactoring which makes the output somewhat more suitable for external parallelization by removing the flushing. See #11560.

krobelus: > investigating how parallelization with Python might work (unsuccessfully so far)

Maybe an LLM assistant (Claude, Aider, ...) can help kickstart this. I think our needs are fairly basic.
We could use `asyncio`, then the code will look fairly similar to synchronous code. The classic unix way would be to have our own event loop that calls `pselect()` (or in this case `waitpid()`). Maybe we can get away with only one thread in the test driver, since most of the heavy lifting is done by subprocesses. That would probably mean we'd want to change `check_path` from `littlecheck.py` to return a future (for example by making it `async`).
(The extra Python process for each pexpect test we can probably get rid of eventually, but there's no hurry to do that).

danielrainer: I have been looking at `asyncio`, I just haven't tried async stuff in Python before (and not much in general), so I expect to spend some time learning about it.

I don't like using LLMs for coding, especially for code that is published due to the extremely sketchy copyright situation (which many people seem to just ignore).

danielrainer: #11561

build_tools/check.sh
danielrainer: > I think build should happen before style.fish because we want an up-to-date fish_indent.

Yes, that makes more sense. #11563

---

Regarding the rust flags, are you worried about them breaking anything? Otherwise the only real advantage I see for setting them for each command individually might be that combined with `set -x` it is more apparent with which flags each command runs.

PR #11559, opened 2025-06-07
completions/adb: add listing device files for exec-out subcommand

pxeemo:
## Description
The usage of `adb exec-out` is basically the same as `adb shell`, but instead, it sends the output to stdout. 
So listing files for it is very useful and makes even more sense since you often need to cat a file into stdout with it.

Fixes issue #

## TODOs:


zanchey:
Thanks!

PR #11560, opened 2025-06-07
Extract function for running tests

danielrainer:
This is done to prepare for running tests concurrently.

Align output and prevent flushing stdout between test name and result.

### DIFF
--- a/tests/test_driver.py
+++ b/tests/test_driver.py
@@ -174,8 +174,10 @@ def main():
             f"{arg.ljust(longest_test_name_length)}  {color}{result}{RESET}  {duration_str}{suffix_str}"
         )
 
+    tmp_root = tempfile.mkdtemp(prefix="fishtest-root-")
+
     for f, arg in files:
-        match run_test(f, arg, script_path, args, def_subs, lconfig, fishdir):
+        match run_test(tmp_root, f, arg, script_path, args, def_subs, lconfig, fishdir):
             case TestSkip(arg):
                 skipcount += 1
                 print_result(arg, "SKIPPED", BLUE)
@@ -186,6 +188,8 @@ def main():
             case TestPass(arg, duration_ms):
                 passcount += 1
                 print_result(arg, "PASSED", GREEN, duration_ms)
+
+    shutil.rmtree(tmp_root)
 
     if passcount + failcount + skipcount > 1:
         print(f"{passcount} / {passcount + failcount} passed ({skipcount} skipped)")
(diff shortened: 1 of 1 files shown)


tests/test_driver.py
```diff
@@ -158,91 +160,30 @@ def main():
...
-                if ret is littlecheck.SKIP:
-                    print(f"{BLUE}SKIPPED{RESET}")
-                    skipcount += 1
-                elif ret:
-                    print(f"{GREEN}PASS{RESET} ({duration_ms} ms)")
-                    passcount += 1
+        match run_test(f, arg, script_path, args, def_subs, lconfig, fishdir):
+            case TestSkip(arg):
+                skipcount += 1
+                print(f"{arg.ljust(longest_test_name_length)}  {BLUE}SKIPPED{RESET}")
```
krobelus: maybe extract `arg.ljust(longest_test_name_length)` as well, into a local variable.
Something like this (obviously theres more we can improve in this script, such as giving a more descriptive type to `duration_ms`, adding mypy etc.).
I'm happy to merge as-is too.

```diff
diff --git a/tests/test_driver.py b/tests/test_driver.py
index 1bbbf56850..c4ba1d7a4b 100755
--- a/tests/test_driver.py
+++ b/tests/test_driver.py
@@ -167,22 +167,24 @@
     failcount = 0
     passcount = 0
     for f, arg in files:
+        def print_result(color: str, result: str, suffix: str = ""):
+            print(f"{arg.ljust(longest_test_name_length)}  {color}{result}{RESET}{suffix}")
+        def format_duration(duration_ms: int):
+            return f" {str(duration_ms).rjust(max_expected_digits_duration)} ms"
         match run_test(f, arg, script_path, args, def_subs, lconfig, fishdir):
             case TestSkip(arg):
                 skipcount += 1
-                print(f"{arg.ljust(longest_test_name_length)}  {BLUE}SKIPPED{RESET}")
+                print_result(BLUE, "SKIPPED")
             case TestFail(arg, duration_ms, error_message):
                 failcount += 1
                 failed += [arg]
-                if duration_ms is None:
-                    print(f"{arg.ljust(longest_test_name_length)}  {RED}FAILED{RESET} ")
-                else:
-                    print(f"{arg.ljust(longest_test_name_length)}  {RED}FAILED{RESET} {str(duration_ms).rjust(max_expected_digits_duration)} ms")
+                duration = "" if duration_ms is None else format_duration(duration_ms)
+                print_result(RED, "FAILED", duration)
                 if error_message is not None:
                     print(error_message)
             case TestPass(arg, duration_ms):
                 passcount += 1
-                print(f"{arg.ljust(longest_test_name_length)}  {GREEN}PASSED{RESET} {str(duration_ms).rjust(max_expected_digits_duration)} ms")
+                print_result(GREEN, "PASSED", format_duration(duration_ms))
 
     if passcount + failcount + skipcount > 1:
         print(f"{passcount} / {passcount + failcount} passed ({skipcount} skipped)")
```

tests/test_driver.py
```diff
@@ -158,91 +160,30 @@ def main():
...
-                        "fish": str(fishdir / "fish"),
-                        "fish_key_reader": str(fishdir / "fish_key_reader"),
-                        "fish_indent": str(fishdir / "fish_indent"),
-                        "TERM": "dumb",
-                        "FISH_FORCE_COLOR": "1" if sys.stdout.isatty() else "0",
-                    }
-                )
-                print(f"{arg}..", end="", flush=True)
-                if not PEXPECT:
-                    print(f"{BLUE}SKIPPED{RESET}")
```
krobelus, line 208: yeah I noticed this duplication as well, thanks

krobelus (approved):

tests/test_driver.py
danielrainer: Good idea to extract this as a function. The value of `arg` is intentionally taken from the return value of the test, because when this is parallelized the loop variable will no longer exist.

tests/test_driver.py
```diff
@@ -48,7 +50,7 @@ def makeenv(script_path, home, test_helper_path):
     if test_helper_path:
         thp = Path(test_helper_path)
         if not os.path.exists(thp / "fish_test_helper"):
-            comp = subprocess.run(
+            subprocess.run(
```
krobelus, line 53: unrelated: I think these two should set `check=True` (or equivalently, use `subprocess.check_call()`), so we get a better error when compilation fails

tests/test_driver.py
```diff
@@ -158,91 +160,37 @@ def main():
...
-                    print(f"{RED}FAILED{RESET} ({duration_ms} ms)")
-                    if proc.stdout:
-                        print(proc.stdout.decode("utf-8"))
-                    if proc.stderr:
-                        print(proc.stderr.decode("utf-8"))
-                    failcount += 1
-                    failed += [arg]
-                    print(f"Tmpdir is {home}")
+        match run_test(tmp_root, f, arg, script_path, args, def_subs, lconfig, fishdir):
+            case TestSkip(arg):
```
krobelus, line 181: FWIW I sometimes still need to work with Ubuntu 20 which has Python 3.8 (match statements are from 3.10),
but this is test code, not product, so no need to accomodate systems that old.
It was also fairly easy to solve my problems by installing a recent system in a chroot.

tests/test_driver.py
```diff
@@ -158,91 +160,37 @@ def main():
...
+                print_result(arg, "SKIPPED", BLUE)
+            case TestFail(arg, duration_ms, error_message):
+                failcount += 1
+                failed += [arg]
+                print_result(arg, "FAILED", RED, duration_ms, error_message)
+            case TestPass(arg, duration_ms):
+                passcount += 1
+                print_result(arg, "PASSED", GREEN, duration_ms)
+
+    shutil.rmtree(tmp_root)
```
krobelus, line 192: Maybe `with tempfile.TemporaryDirectory(...)` would be more elegant than an explicit cleanup step
but I don't really care (unless we can expect the `run_test` to throw an exception regularly)

tests/test_driver.py
krobelus: My bad, we have some nightly builders with python <3.10.
e.g. https://build.opensuse.org/package/live_build_log/shells:fish:nightly:master/fish/15.6/x86_64
debian oldolstable has [3.7](https://packages.debian.org/search?suite=default&section=all&arch=any&searchon=names&keywords=python3-dev) so that's probably the very least we can assume..

zanchey:
Yeah, this broke the builds on a bunch of platforms, sorry

tests/test_driver.py
danielrainer: Python 3.7 has been EOL for 2 years now. There is value in supporting builds on many different platforms, but at some point it does become annoying to cater to extremely outdated software versions by having to write worse code. For reference, Python 3.10 was released in 2021, and Python 3.9 will be EOL in October. Our MSRV is from 2023.

If people insist on updating software infrequently for stability that's fine by me, but in that case, why would you need to build the latest versions of software and hold back adoption of language and library improvements?

In any case, it would be helpful to have a clearly stated minimum version policy and CI tests which run these versions to ensure that they work, as we already do with Rust.

zanchey: 3.9 should be enough

krobelus: > 3.9 should be enough

Ok so I guess we can track `debian-oldstable`?

Yeah, it is a bit inconsistent that we require Rust 1.70 which is not yet in Debian stable (only 1.63 is),
and still support Python 3.7 (that don't have the walrus operator, match statements etc).
But if it helps people, it wouldn't be totally unreasonable, since we don't have a lot of Python code.
It's probably also considered "easier" to install Rust via Rustup than Python though I'm not sure if that's actually true.

> If people insist on updating software infrequently for stability
> that's fine by me, but in that case, why would you need to build the
> latest versions of software and hold back adoption of language and
> library improvements?

A lot of the time an OS at $DAYJOB might be wildly outdated,
so it's nice to be able to build fish with relatively little fuzz.

Sure, it's hacky to combine software release 5+ years apart,
and we should tell people to use better approaches -- e.g. use docker/systemd-nspawn/chroot, or,
if you don't have root access, a separate package manager installing to `$HOME/.local`.

> In any case, it would be helpful to have a clearly stated minimum
> version policy and CI tests which run these versions to ensure that
> they work, as we already do with Rust.

Yeah that sounds like a good idea. Not sure what's best practice for specifying python version, I guess we can add it next to `rust-version` in Cargo.toml.

Today CI configuration in `.cirrus.yml` does have a FreeBSD version that failed because it has Python < 3.10.
(It runs only on PRs IIRC, I must have missed it)
I don't know if it's intentional that that we're checking such an old version of FreeBSD,
so I'd maybe change that to track stable/latest.

Nightly (and other) builds use OpenSUSE build service, and (intentionally) pin OS versions.

For now I guess we should do

```diff
diff --git a/tests/test_driver.py b/tests/test_driver.py
index b596967717..1b5a0696bf 100755
--- a/tests/test_driver.py
+++ b/tests/test_driver.py
@@ -174,17 +174,21 @@
             for f, arg in files
         ]
         for task in asyncio.as_completed(tasks):
-            match await task:
-                case TestSkip(arg):
-                    skipcount += 1
-                    print_result(arg, "SKIPPED", BLUE)
-                case TestFail(arg, duration_ms, error_message):
-                    failcount += 1
-                    failed += [arg]
-                    print_result(arg, "FAILED", RED, duration_ms, error_message)
-                case TestPass(arg, duration_ms):
-                    passcount += 1
-                    print_result(arg, "PASSED", GREEN, duration_ms)
+            result = await task
+            if isinstance(result, TestSkip):
+                arg = result.arg
+                skipcount += 1
+                print_result(arg, "SKIPPED", BLUE)
+            elif isinstance(result, TestFail):
+                arg, duration_ms, error_message = result.arg, result.duration_ms, result.error_message
+                failcount += 1
+                failed += [arg]
+                print_result(arg, "FAILED", RED, duration_ms, error_message)
+            elif isinstance(result, TestPass):
+                arg = result.arg
+                duration_ms = result.duration_ms
+                passcount += 1
+                print_result(arg, "PASSED", GREEN, duration_ms)
 
     if passcount + failcount + skipcount > 1:
         print(f"{passcount} / {passcount + failcount} passed ({skipcount} skipped)")
```

danielrainer: > It runs only on PRs IIRC

I'm not familiar with Cirrus, but the `only_if: $CIRRUS_REPO_OWNER == 'fish-shell'` suggests that it only runs once it is on a branch owned by `fish-shell`, which is not the case for a PR opened by me, since it will be a branch in my fork.

I think your suggested changes to the test driver are reasonable if we can't use `match`.

krobelus: Cirrus CI definitely runs on some PRs from forks e.g. https://github.com/fish-shell/fish-shell/pull/11574

Maybe it doesn't run for PRs of contributors where no one has approved CI runs yet.
But the weird part is, I don't see a button to approve CI runs, and Github Actions are run regardless..
even though the whole point of disabling CI for untrusted contributors was to mitigate all the security issues in github actions.. I thought that was disabled by default for new contributors; I guess something changed.

PR #11561, opened 2025-06-07
Parallelize test driver

danielrainer:
This uses Python's `asyncio` to run tests in parallel, which speeds up test
execution significantly.

The timeout is removed. It would be possible to add a timeout to
`asyncio.as_completed()` if we want that.

On top of #11560.

### DIFF
--- a/tests/littlecheck.py
+++ b/tests/littlecheck.py
@@ -4,20 +4,21 @@
 
 from __future__ import unicode_literals
 from __future__ import print_function
-
 import argparse
 import datetime
+from difflib import SequenceMatcher
 import io
 import re
 import shlex
 import subprocess
 import sys
+import unicodedata
 
 try:
     from itertools import zip_longest
 except ImportError:
     from itertools import izip_longest as zip_longest
-from difflib import SequenceMatcher
+
 
 # Directives can occur at the beginning of a line, or anywhere in a line that does not start with #.
 COMMENT_RE = r"^(?:[^#].*)?#\s*"
@@ -91,9 +92,6 @@ class Config(object):
 
 def output(*args):
     print("".join(args) + "\n")
-
-
-import unicodedata
 
 
 def esc(m):
@@ -382,7 +380,7 @@ def perform_substitution(input_str, subs):
     return re.sub(r"%(%|[a-zA-Z0-9_-]+)", subber, input_str)
 
 
-def runproc(cmd):
+def runproc(cmd, env=None):
     """Wrapper around subprocess.Popen to save typing"""
     PIPE = subprocess.PIPE
     proc = subprocess.Popen(
@@ -392,18 +390,20 @@ def runproc(cmd):
         stderr=PIPE,
         shell=True,
         close_fds=True,  # For Python 2.6 as shipped on RHEL 6
+        env=env,
     )
     return proc
 
 
 class TestRun(object):
-    def __init__(self, name, runcmd, checker, subs, config):
+    def __init__(self, name, runcmd, checker, subs, config, env=None):
         self.name = name
         self.runcmd = runcmd
         self.subbed_command = perform_substitution(runcmd.args, subs)
         self.checker = checker
         self.subs = subs
         self.config = config
+        self.env = env
 
     def check(self, lines, checks):
         # Reverse our lines and checks so we can pop off the end.
@@ -489,7 +489,7 @@ class TestRun(object):
 
         if self.config.verbose:
             print(self.subbed_command)
-        proc = runproc(self.subbed_command)
+        proc = runproc(self.subbed_command, env=self.env)
         stdout, stderr = proc.communicate()
         # HACK: This is quite cheesy: POSIX specifies that sh should return 127 for a missing command.
         # It's also possible that it'll be returned in other situations,
@@ -668,7 +668,7 @@ class Checker(object):
         ]
 
 
-def check_file(input_file, name, subs, config, failure_handler):
+def check_file(input_file, name, subs, config, failure_handler, env=None):
     """Check a single file. Return a True on success, False on error."""
     success = True
     lines = Line.readfile(input_file, name)
@@ -677,7 +677,7 @@ def check_file(input_file, name, subs, config, failure_handler):
     # Run all the REQUIRES lines first,
     # if any of them fail it's a SKIP
     for reqcmd in checker.requirecmds:
-        proc = runproc(perform_substitution(reqcmd.args, subs))
+        proc = runproc(perform_substitution(reqcmd.args, subs), env=env)
         proc.communicate()
         if proc.returncode > 0:
             return SKIP
@@ -687,16 +687,16 @@ def check_file(input_file, name, subs, config, failure_handler):
 
     # Only then run the RUN lines.
     for runcmd in checker.runcmds:
-        failure = TestRun(name, runcmd, checker, subs, config).run()
+        failure = TestRun(name, runcmd, checker, subs, config, env=env).run()
         if failure:
             failure_handler(failure)
             success = False
     return success
 
 
-def check_path(path, subs, config, failure_handler):
+def check_path(path, subs, config, failure_handler, env=None):
     with io.open(path, encoding="utf-8") as fd:
-        return check_file(fd, path, subs, config, failure_handler)
+        return check_file(fd, path, subs, config, failure_handler, env=env)
 
 
 def parse_subs(subs):
--- a/tests/littlecheck.py
+++ b/tests/littlecheck.py
@@ -5,12 +5,12 @@
 from __future__ import unicode_literals
 from __future__ import print_function
 import argparse
+import asyncio
 import datetime
 from difflib import SequenceMatcher
 import io
 import re
 import shlex
-import subprocess
 import sys
 import unicodedata
 
@@ -382,17 +382,15 @@ def perform_substitution(input_str, subs):
 
 def runproc(cmd, env=None):
     """Wrapper around subprocess.Popen to save typing"""
-    PIPE = subprocess.PIPE
-    proc = subprocess.Popen(
-        cmd,
-        stdin=PIPE,
-        stdout=PIPE,
-        stderr=PIPE,
-        shell=True,
-        close_fds=True,  # For Python 2.6 as shipped on RHEL 6
-        env=env,
+    return asyncio.run(runproc_async(cmd, env=env))
+
+
+async def runproc_async(cmd, env=None):
+    """Wrapper around subprocess.Popen to save typing"""
+    PIPE = asyncio.subprocess.PIPE
+    return await asyncio.create_subprocess_shell(
+        cmd, stdin=PIPE, stdout=PIPE, stderr=PIPE, env=env
     )
-    return proc
 
 
 class TestRun(object):
@@ -477,6 +475,10 @@ class TestRun(object):
 
     def run(self):
         """Run the command. Return a TestFailure, or None."""
+        return asyncio.run(self.run_async())
+
+    async def run_async(self):
+        """Run the command. Return a TestFailure, or None."""
 
         def split_by_newlines(s):
             """Decode a string and split it by newlines only,
@@ -489,8 +491,8 @@ class TestRun(object):
 
         if self.config.verbose:
             print(self.subbed_command)
-        proc = runproc(self.subbed_command, env=self.env)
-        stdout, stderr = proc.communicate()
+        proc = await runproc_async(self.subbed_command, env=self.env)
+        stdout, stderr = await proc.communicate()
         # HACK: This is quite cheesy: POSIX specifies that sh should return 127 for a missing command.
         # It's also possible that it'll be returned in other situations,
         # most likely when the last command in a shell script doesn't exist.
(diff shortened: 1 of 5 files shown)


tests/littlecheck.py
```diff
@@ -382,18 +380,12 @@ def subber(m):
     return re.sub(r"%(%|[a-zA-Z0-9_-]+)", subber, input_str)
 
 
-def runproc(cmd):
+async def runproc(cmd):
```
krobelus: we should make the same changes to https://github.com/ridiculousfish/littlecheck

AFAICT the public-facing API is only `check_path`.
Would it be easy to provide both `check_path` and `check_path_async`?
In case a client doesn't want to use async..
Though I'm not sure what's the convention. Clients can still use `asyncio.run`.
I don't know how many people are using library so I don't think it matters too much which way we pick

tests/test_driver.py
```diff
@@ -158,91 +162,42 @@ def main():
...
+        )
+
+    tmp_root = tempfile.mkdtemp(prefix="fishtest-root-")
+
+    tasks = [
+        run_test(tmp_root, f, arg, script_path, args, def_subs, lconfig, fishdir)
+        for f, arg in files
+    ]
+
+    for task in asyncio.as_completed(tasks):
```
krobelus: this takes almost 40 seconds here, almost twice as long as with my GNU parallel experiment.
That one parallelized everything, including the test driver; but I'd be surprised if the sequential test_driver is the bottlenekc.
Maybe it's because GNU parallel defaults to running only `$(nproc)` parallel process at a time.
I don't know if `asyncio.run()` has a comparable limit, probably not since it doesn't know that we're launching processes.

Maybe we should try something like https://docs.python.org/3/library/multiprocessing.html

tests/test_driver.py
```diff
@@ -253,9 +208,99 @@ def main():
...
+        elif returncode == 127:
+            return TestSkip(arg)
+        else:
+            error_message = ""
+            if stdout:
+                error_message += stdout.decode("utf-8")
+            if stderr:
+                error_message += stderr.decode("utf-8")
+            error_message += f"Tmpdir is {home}"
+            return TestFail(arg, duration_ms, error_message)
```
krobelus, line 293: this sometimes fails tmux tests

```
checks/tmux-wrapping.fish                      FAILED   13581 ms
Tmpdir is /tmp/fishtest-root-dlqffp57/fishtest-t7aqq33v
Failure:

  The CHECK on line 16 wants:
    read> name

    [N] read> name
```

looks like something is enabling fish_vi_key_bindings; maybe $HOME is not isolated properly?

tests/test_driver.py
```diff
@@ -253,9 +208,99 @@ def main():
...
+    else:
+        return TestFail(arg, None, "Error in test driver. This should be unreachable.")
+
+
 if __name__ == "__main__":
     try:
-        ret = main()
+        ret = asyncio.run(main())
         sys.exit(ret)
     except KeyboardInterrupt:
```
krobelus, line 302: ctrl-c gives some weird output.
Pressing it again produces some more.
Ideally the driver quits after the first one.

	^C
	...
	child process pid 2072526 exit status already read:  will report returncode 255
	child process pid 2072443 exit status already read:  will report returncode 255
	child process pid 2072395 exit status already read:  will report returncode 255
	^C
	Task was destroyed but it is pending!
	task: <Task cancelling name='Task-2' coro=<run_test() done, defined at /home/johannes/git/fish-shell/tests/test_driver.py:232> wait_for=<Future pending cb=[Task.task_wakeup()]> cb=[_AsCompletedIterator._handle_completion(), gather.<locals>._done_callback() at /usr/lib/python3.13/asyncio/tasks.py:820]>
	Task was destroyed but it is pending!
	...

tests/littlecheck.py
danielrainer: Providing both would be simple with some code duplication. I don't know who uses littlecheck, so I can't comment on whether anyone would have issues with the async version being the only available one. Having to convert to async code, even if it's just wrapping the call in `asyncio.run`, is a breaking change.

tests/test_driver.py
danielrainer: I just checked, and it turns out that without `--cachedir` the `fish_test_helper` is compiled for each test, which takes about 70 ms each time on my system. Without `--cachedir` it takes 30 seconds on my system, but with `--cachedir` only 17. That should be simple to resolve by building the `fish_test_helper` once before the tests if necessary.

danielrainer: `--cachedir` is only used for caching this binary, so we might consider removing the flag to simplify things. Spending a couple tens of milliseconds on compiling per `test_driver.py` invocation seems manageable.

tests/littlecheck.py
krobelus: not knowing much about this I'd use `check_path_async` but if that's not ideal then a breaking seems fine too, it's not even packaged in PyPI

tests/test_driver.py
krobelus: nice find -- yeah, I agree we should remove the cachedir flag, to simplify the `test_driver.py --help` output.

tests/test_driver.py
danielrainer: Hm, I don't have tmux installed, so these tests don't run locally for me.

Maybe the `os.environ.update()` call causes contamination between tests?

tests/littlecheck.py
danielrainer: I implemented the functions which were changed to `async` as wrappers around the new `async` versions, which keeps code duplication minimal. It does require `async` support to run the old functions now. I don't know if that might be a problem.

tests/test_driver.py
danielrainer: Can you reproduce the issue on the latest code which gets rid of the `os.environ` accesses?

danielrainer: It just happened in CI: https://github.com/fish-shell/fish-shell/actions/runs/15520487162/job/43693042641?pr=11561.

I'm not familiar with tmux or the our tests using it. Is there anything timing-sensitive in it?

tests/test_driver.py
danielrainer: We could do something like https://stackoverflow.com/a/67580265.

Cleaning up the tmpdir might be some extra work. But I think we should first focus on getting the tests running reliably before worrying about user-initiated early termination.

tests/test_driver.py
krobelus: > Maybe the os.environ.update() call causes contamination between tests?

yeah, that fixes it. We must not write to the process-wide `os.environ` while there is concurrency going on.
But I'm not sure about clearing _all_ environment variables. I'd do `def makeenv(): return os.environ - vars_to_remove + vars_to_add`.
Makes it easier to (temporarily) inject environment variables. Not sure how much that's used.
It's possible that we want to revisit this and make tests truly independent of `os.environ`,
but I think that should be done in a separate PR.
For now we expect devs to not pollute their environment with things that could break tests.
If we change that, we can remove some variable-erasing from the prelude of `tests/checks/git.fish` etc..

> I'm not familiar with tmux or the our tests using it. Is there anything timing-sensitive in it?

that's a different failure, yeah it's likely timing based.
We have an issue called "Tests in CI are flaky".

The tmux-based tests have various problems -- they run a nested fish for dubious reasons,
they're slow due to all the forking involved, they use sleeps everywhere,
they are hard to debug (interactively) for outsiders.
We can fix most of those problems but I haven't really found it to be a bottleneck.
tmux allows us to write system tests because it behaves like a real terminal,
so instead of testing the escape sequences output by fish (for cursor movement etc),
we can take a tmux screenshot and write assertions on that.
It's possible that we want to use a different (headless) terminal that's tailor-made for testing.. but probably not anytime soon

tests/test_driver.py
krobelus: when I press ctrl-c a little later (presumably after all tests have been kicked off), then it does exit immediately but does print things like

```
RuntimeError: Event loop is closed
Exception ignored in: <function BaseSubprocessTransport.__del__ at 0x717a68125300>
```

not a blocker, we can fix these whenever

tests/littlecheck.py
krobelus: Is there a reason why anyone would ever use `check_file()` or any of the other functions?
I think those should be private and so need no compatibility wrapper.
The one for `check_path()` should be enough (and even for that we don't know if anyone will use it)

> It does require async support to run the old functions now.

`asyncio.run` was added in 3.7, so I think we're good

tests/test_driver.py
```diff
@@ -25,85 +26,58 @@
 RED = "\033[31m"
 
 
-def makeenv(script_path, home, test_helper_path):
-    xdg_config = home + "/xdg_config_home"
-    func_dir = xdg_config + "/fish/functions"
+def makeenv(script_path: Path, home: Path, test_helper_path: Path) -> dict[str, str]:
```
krobelus: I'd probably have put the commit that changes `str -> Path` first,
because that one is unlikely to see further change but may cause
trivial merge conflicts when squashing fixups

tests/test_driver.py
```diff
@@ -25,85 +26,58 @@
...
-            "FISH_UNIT_TESTS_RUNNING": "1",
-            "XDG_CONFIG_HOME": xdg_config,
-            "XDG_DATA_HOME": xdg_data,
-            "XDG_RUNTIME_DIR": xdg_runtime,
-            "XDG_CACHE_HOME": xdg_cache,
-            "fish_test_helper": home + "/fish_test_helper",
-            "LANG": "C",
-            "LC_CTYPE": "en_US.UTF-8",
-        }
+    shutil.copy(test_helper_path, home / "fish_test_helper")
```
krobelus: Instead of copying, a hardlink would make more sense,
since it's effectively immutable, and we're on the same file system.

Actually, we shouldn't need to create a file at all,
since we could forward `test_helper_path`?
We don't currently run tests in a sandbox like bubblewrap so that should just work,
and we can always change it later.

tests/test_driver.py
```diff
@@ -261,34 +248,28 @@ def run_test(
         )
         if not PEXPECT:
             return TestSkip(arg)
-        try:
-            proc = subprocess.run(
-                ["python3", path],
-                capture_output=True,
-                env=pyenviron,
-                # Timeout of 120 seconds, about 10 times what any of these takes
-                timeout=120,
```
krobelus, line 270: yeah I don't expect the timeout was needed because pexpect already has its own timeouts

tests/test_driver.py
danielrainer: I think it is desirable to isolate the tests from the environment as well as possible, even using `$PATH` from the environment seems a bit sketchy.
Do people deliberately modify their env vars to modify test behavior? If so, maybe we can address such use cases with arguments to the test driver script.
Regardless, we can of course keep the old behavior for now.

---

Regarding tmux tests, how do we proceed in the context of this PR? It's quite annoying to have tests which sometimes fail in CI just because of timing issues. While this already seems to be a problem, this PR probably makes it significantly worse, due to all tests running in parallel, which will result in higher latencies within single tests.

danielrainer:
I split up some of the changes and put the parallelization last.

tests/littlecheck.py
danielrainer: Since they're all externally visible, I don't know if anyone uses them, so I decided to be safe and not break anything. I agree that it seems unlikely that anyone would use them, so I'd also be fine with just adding a wrapper for `check_path` if you prefer that.

tests/test_driver.py
krobelus: > this PR probably makes it significantly worse

yeah once we make .github/workflows/main.yml actually use the 
test driver's scheduling instead of ctest's,
then we might hit problems.
Today `CTEST_PARALLEL_LEVEL` is set to 4, and
when there is no limit, our CMake logic sets it to `nproc/2`.
So both significantly less parallelism than `test_driver.py`.

I haven't seen `test_driver.py` fail locally.
Once we run it in CI, we can try how often it fails;
if it does fail noticeably more often, we can limit, as sketched below.
I personally don't care much how long CI takes, and it's not gonna take very long either way.

```diff
diff --git a/.github/workflows/main.yml b/.github/workflows/main.yml
index efab5eabec..2bf25e0c38 100644
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
 on: [push, pull_request, workflow_dispatch]
 
 env:
-  CTEST_PARALLEL_LEVEL: "4"
   CMAKE_BUILD_PARALLEL_LEVEL: "4"
 
 permissions:
@@ -172,4 +172,5 @@
         make -C build VERBOSE=1
     - name: make fish_run_tests
       run: |
-        make -C build VERBOSE=1 fish_run_tests
+        tests/test_driver.py --max-concurrency=4 build
diff --git a/tests/test_driver.py b/tests/test_driver.py
index b596967717..63afb73b22 100755
--- a/tests/test_driver.py
+++ b/tests/test_driver.py
@@ -10,6 +10,7 @@
 import sys
 import tempfile
 from typing import Optional
+import contextlib
 
 import littlecheck
 
@@ -169,10 +170,12 @@
             tmp_root / "fish_test_helper",
         )
 
-        tasks = [
-            run_test(tmp_root, f, arg, script_path, def_subs, lconfig, fishdir)
-            for f, arg in files
-        ]
+        sem = asyncio.Semaphore(int(parallel_level)) if max_concurrency else contextlib.nullcontext()
+        async def run(f, arg):
+            async with sem:
+                return await run_test(tmp_root, f, arg, script_path, def_subs, lconfig, fishdir)
+        tasks = [run(f, arg) for f, arg in files]
         for task in asyncio.as_completed(tasks):
             match await task:
                 case TestSkip(arg):
```

danielrainer: That seems reasonable. I find it valuable to have fast CI, because it just makes for a faster feedback loop. With the new `check.sh` script, many issues can be detected locally, but there is still some stuff that will only show up in CI.

For reference, I have observed at least 2 timeout-related CI failures working on this PR.

PR #11565, opened 2025-06-08
Translation updates in checks

danielrainer:
Integrate the `update_translations.fish` script into normal tests.

Fix several issues I noticed while implementing this.


danielrainer:
Now that the extraction proc macro always runs it panics when running via rust-analyzer. Specifically, the panics are about the `TokenTree`. The code expects a group containing a literal. This seems to be the way the macro is invoked when running `cargo build` and similar commands. But rust-analyzer apparently passes a literal which is not wrapped in a group, which makes much more sense to me, but for some reason this does not happen for normal builds.

We could modify the proc macro to accept both variants, but I'd like to figure out where the difference is coming from. Can you reproduce this?

build.rs
```diff
@@ -44,6 +44,13 @@ fn main() {
     let cman = std::fs::canonicalize(env!("CARGO_MANIFEST_DIR")).unwrap();
     let targetman = cman.as_path().join("target").join("man");
 
+    let target_gettext_template = cman.as_path().join("target").join("gettext_template.po");
+    let _ = std::fs::remove_file(&target_gettext_template);
+    rsconf::set_env_value(
+        "FISH_GETTEXT_EXTRACTION_FILE",
+        target_gettext_template.to_str().unwrap(),
+    );
```
krobelus: Seems fine; I couldn't measure a performance difference here; an incremental rebuild takes ~8.6s with or without `FISH_GETTEXT_EXTRACTION_FILE`.
It also means that we benefit from cargo's caching.

cargo has a lock, so concurrent builds should not compete to write the same `gettext_template.po`.
*However*, build_tools/fish_xgettext.fish does not acquire the lock,
so running it while `cargo b` is running can cause confusion.

This probably doesn't happen often enough to be a relevant problem,
but we should try to figure out how to stop rust-analyzer-induced builds from touching `gettext_template.po` (if it does as written, and if that means `cargo b` will still produce it..)

---

The alternative would be to move `tests/checks/check-translations.fish` to something like `tests/build/check-translations.sh` (also adding `tests/build/check-sphinx.sh`)
and teach the test driver to run such tests without a modified environment etc.
But that would mean we'd need to keep `rsconf::rebuild_if_env_changed("FISH_GETTEXT_EXTRACTION_FILE");`
which would end up causing more rebuilds, and we don't really want to rebuild while running tests.

tests/checks/check-translations.fish
```diff
@@ -3,6 +3,18 @@
 
 set -l dir (status dirname)
 
+# Ensure that fish is in $PATH for the translation scripts.
+set -lxp PATH (path dirname $fish)
```
krobelus: yeah, we'll have a grandchild fish process running `fish_xgettext`,
so this seems like the easiest way to make that work.

build.rs
```diff
@@ -63,8 +70,6 @@ fn main() {
     #[cfg(not(debug_assertions))]
     rsconf::rebuild_if_paths_changed(&["doc_src", "share"]);
 
-    rsconf::rebuild_if_env_changed("FISH_GETTEXT_EXTRACTION_FILE");
-
```
krobelus: > This seems to be the way the macro is invoked when running cargo build and similar commands.

Yeah I get the same error (see below).

Maybe we can debug where this difference comes from by tracing cargo invocation via a `~/bin/cargo` wrapper script

```
#!/bin/sh
echo >>/tmp/cargo-invocations "+ cargo$(printf " '%s'" "$@")"
exec /bin/cargo "$@"
```

unless rust-analyzer doesn't call the `cargo` program.


```
         FLOG!(                                                                                       
             warning,                                                                                 
*            wgettext!("Can not use the no-execute mode when running an interactive session")         
         );                                                                                           
         opts.no_exec = false;                                                                        
        ╭───────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
        │ Diagnostics (shortcut e):                                                                                     │
        │ • [error][rust-analyzer][macro-error]                                                                         │
        │   proc-macro panicked: Expected group in gettext_extract, but got: Literal { kind: Str, symbol: "Can not use  │
 ╭──╮   │ the no-execute mode when running an interactive session", suffix: None, span: SpanData { range: 2302..2371,   │
 │  │   │ anchor: SpanAnchor(EditionedFileId(FileId(3126), Edition2021), 20), ctx: SyntaxContextId(4294967037) } }      │
 @  @  ╭│   Description: https://rust-analyzer.github.io/book/diagnostics.html#macro-error                              │
 ││ ││ ││                                                                                                               │
 ││ ││ ╯│ • [error][rust-analyzer][macro-error]                                                                         │
 │╰─╯│  │   proc-macro panicked: Expected group in gettext_extract, but got: Literal { kind: Str, symbol: "Can not use  │
 ╰───╯  │ the no-execute mode when running an interactive session", suffix: None, span: SpanData { range: 2302..2371,   │
        │ anchor: SpanAnchor(EditionedFileId(FileId(3126), Edition2021), 20), ctx: SyntaxContextId(4294967037) } }      │
        │   Description: https://rust-analyzer.github.io/book/diagnostics.html#macro-error                              │
        │ 
```

krobelus: maybe it's different proc-macro2 versions? Though that should be..
Not sure what's going on, I asked on [#t-compiler/rust-analyzer > proc macro gets diffferent token stream than with cargo](https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/proc.20macro.20gets.20diffferent.20token.20stream.20than.20with.20cargo/with/522974516)

krobelus: I think that's https://github.com/rust-lang/rust-analyzer/issues/19856,
Let's work around that and add a comment (or commit message) referencing that issue

danielrainer: I added a workaround, which skips extraction for rust-analyzer builds.

We could extract in such contexts as well, but I don't think that would be particularly useful, and it is more likely that we get unexpected conflicts while we don't lock for reading the file.

danielrainer: I added locking, so modifications of the template file while we are reading it should no longer happen.

danielrainer: Apparently `flock(1)` is not available on macos. I added it via `brew install` in CI. Should we encourage developers on macos to install it as well, or do you prefer a workaround like not locking if `flock` does not exist?

build_tools/fish_xgettext.fish
```diff
@@ -15,18 +19,20 @@ begin
...
+    if not set -l --query _flag_use_existing_template
+        # We need to build to ensure that the proc macro for extracting strings runs.
+        cargo check
+        or exit 1
+    end
 
     # Get rid of duplicates and sort.
-    msguniq --no-wrap --strict --sort-output $rust_extraction_file
+    # Locking should prevent concurrent builds from modifying the template.
+    flock --exclusive $repo_root/Cargo.lock msguniq --no-wrap --strict --sort-output $rust_extraction_file
```
krobelus: Good catch - though I think we need to lock `target/debug/.cargo-lock` instead.
I confirmed that `flock --exclusive target/debug/.cargo-lock` blocks `cargo`

(Technically also `target/release/.cargo-lock` and possibly other build configurations, I'm not sure;

build.rs
krobelus: Ok now `rust-analyzer` no longer generates the `gettext_template.po`.

Except it does when configured to automatically run clippy.
Perhaps we should also suppress that.
Digging at the env in `build.rs`, I found:

```
CLIPPY_ARGS=
CLIPPY_TERMINAL_WIDTH=0
RUSTC_WORKSPACE_WRAPPER=/home/johannes/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/clippy-driver
```

so maybe disable things if `std::env::var_os("CLIPPY_ARGS").is_some()`.

---

Either way, AFAICT even after rust-analyzer or r-a-created clippy built the entire workspace, a `cargo b` will still rebuild everything (?) and extract all translatable strings. Not sure what guarantees this but I'm pretty sure that's how it works.

krobelus: it would be okay to fail if flock is not installed, but we can implement it ourselves, for example in Python:

```sh
$ python -c '
import fcntl
with open("target/debug/.cargo-lock", "r") as f:
    while True:
        try:
            fcntl.flock(f.fileno(), fcntl.LOCK_EX);
            break
        except KeyboardInterrupt:
            pass
'
```

build_tools/fish_xgettext.fish
danielrainer: I decided on `Cargo.lock` because that's independent of the target. In my limited testing it works and `strace` shows `flock` calls to lock `Cargo.lock`, but it's admittedly not a particularly robust approach.

AFIACT `flock(1)` does not support locking multiple files/dirs at once, so it is difficult to properly lock the `.cargo-lock` file for each target. And we would have to ensure that when new targets are added the locking takes that into account.

Ideally we would lock the template file directly, but I don't know how that could be done for the entire build process, rather than individually for each invocation of the proc macro. I also considered building up the template file in a different location and renaming it into place once it's done, but I don't know a way to run code after all proc macro invocations are done.

build.rs
danielrainer: I experimented a bit with rebuilding and I don't fully understand the behavior. When I add an early return to the proc macro when `CLIPPY_ARGS` is set, the following two scenarios are possible for rust-analyzer builds:
1. When configured to run clippy, the file is not written. (It is deleted in `build.rs`, which we could prevent by checking `CLIPPY_ARGS` there as well.
2. When configured not to run clippy, the file does get written, meaning the `TokenTree` is a literal wrapped in a group like for non-ra builds. The resulting file is also much larger (105 KB vs 35 KB for `cargo check` builds).

I think rust-analyzer builds in `target/debug/incremental`, so its builds are independent from normal ones. Which is another reason why locking `.cargo-lock` files in build dirs would be difficult.

Running
```sh
cargo clean
cargo check
cargo clippy
cargo check # this could be the one in fish_xgettext
```
results in not having the template file. Not deleting the file in `build.rs` when `CLIPPY_ARGS` is set prevents this, so I think that is the way to go if we prevent updates in the proc macro when `CLIPPY_ARGS` is set.

danielrainer: > implement it ourselves

Seems reasonable. Added.

build_tools/fish_xgettext.fish
krobelus: Yeah I can't think of a "perfect" solution. I guess this is mostly fine in practice..

We could make `test_driver.py` start by copying `target/gettext_template.po` to its temporary test-root, to reduce the chance that a concurrent `cargo b` breaks the running test..

But actually, `test_driver.py target/debug`
already assumes that `target/debug/fish` and `target/debug/fish_indent`
will not change throughout execution.

So developers are already used to this.
The assumption can be extended to `gettext_template.po`, as long as `rust-analyzer` doesn't update it.

Maybe this means we should move it to `target/debug/gettext_template.po`,
and make the test call

```
$ cat tests/checks/check-translations.fish
#RUN: fish=%fish fish_gettext_template=%fish_gettext_template %fish %s
build_tools/update_translations.fish --use-existing-template=$fish_gettext_template
```

It *seems* that this would make the most sense -- gettext_template.po is a
compilation result just like fish, fish_indent, fish_key_reader.

build.rs
krobelus: Ok. I think the main problem left is how to prevent `rust-analyzer` from ever writing the template (since I doubt that rust-analyzer locks `Cargo.lock` for the whole time?).
By default we should be somewhat okay because it sets `CLIPPY_ARGS`.
But that doesn't seem reliable (clippy could be turned off); we could check `RA_RUSTC_WRAPPER` but that's hacky as well (not sure if this is documented).

krobelus: I think `cargo check` should not generate the gettext template but `cargo build` should,
because only `build` generates binaries.
Assuming `rm target/*.po; cargo check && cargo b` still generates everything..

danielrainer: The downside of `cargo check` not generating the template would be that for standalone uses of `update_translations.fish`, we would need to run `cargo build` in the xgettext script, which means potentially unnecessary code generation, as you remarked earlier (when `cargo build` was used there). I also don't know if we have a way of distinguishing whether the proc macro runs in a `cargo check` or `cargo build` context.

danielrainer: Regarding r-a, it does not seem like there is an established way of distinguishing r-a compilation from other compilation. Some hacky env var check might work, but it's definitely not great to rely on undocumented behavior. (I guess the same applies to the locking, but if there is no better way of achieving it we don't really have a choice.)

build_tools/fish_xgettext.fish
danielrainer: > Maybe this means we should move it to `target/debug/gettext_template.po`

We could do that. We can check if `BUILD_PROFILE` is debug and only then delete and regenerate the template. Then we would have no interference with other targets.

If we change the location of the template file, I don't see why this would require changes to `check-translations.fish`. We can just modify the definition of `rust_extraction_file` in the xgettext script.

One more idea I just had is to define a special target for extraction, which could trivially prevent most conflicts. E.g.

```diff
diff --git a/Cargo.toml b/Cargo.toml
index 0dafbd643..6bac05d56 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,9 @@ lto = true
 inherits = "release"
 debug = true
 
+[profile.xgettext]
+inherits = "dev"
+
 [package]
 name = "fish"
 version = "4.1.0-alpha0"
```
Then we can generate the template by running `cargo check --profile xgettext`. With this setup, we only really have to worry about concurrent invocations of `check.sh` and `update_translations.fish`. (The `check-translations.fish` test invokes `update_translations.fish` in a way that does not cause changes to the template.) I think these are specific enough that we can expect devs not to do that.

danielrainer: Hm, I just checked and `BUILD_PROFILE` is still `debug` when using the `xgettext` profile. The simpler alternative is probably to do what was done earlier in this PR and only set `FISH_GETTEXT_EXTRACTION_FILE` in `checks.sh` and `fish_xgettext.fish`. We can let that point to a tmpfile and modify the xgettext script to accept a filename with `--use-existing-template`. That solves all the race condition problems with template generation, at the cost of having to build every time `fish_xgettext.fish` runs.

danielrainer: I changed the implementation as suggested in the previous comment. This frees us from all the related race condition concerns. The only somewhat ugly parts are the need for recompilation for each `fish_xgettext.fish` invocation and the need to pass the template location to the `check-translations.fish` test via env var (through `test_driver.py`.

The result of the latter is that this test will now be skipped by default if `test_driver.py` is invoked directly. I don't think that this is much of an issue, since we are trying to switch over to `check.sh`. Additionally, people who don't work on translations are unlikely to cause this test to fail, and people who do can run `update_translation.fish` manually if they want.

krobelus: Makes sense. I'm fine merging this as-is.
AFAICT it does effectively remove the check from CI since we don't run `check.sh` yet (or set `FISH_GETTEXT_EXTRACTION_FILE` in another way).
Not a big deal, we can add that back later.

Maybe we can file a feature request for Cargo and/or Rustc to provide better support for this use case ("extract arbitrary data from Rust sources"). Of course we could also keep the translatable strings in a different format and generate a Rust file with constants.. but the existing approach is probably nicer.  

(I was wondering whether we should stop checking in the actual `po` files and instead generate them
based on a new format. Crucially that new format will not include untranslated strings.
That would adding new user-visible strings without having to know how to update translation files.
But for that reason alone it's probably not worth the trouble, since we'd still need to update translations whenever a source string is removed or changed.)

build_tools/fish_xgettext.fish
```diff
@@ -15,18 +19,22 @@ begin
         echo ""
     end
 
-    set -l rust_extraction_file (mktemp)
+    set -g repo_root (status dirname)/..
 
-    # We need to build to ensure that the proc macro for extracting strings runs.
-    FISH_GETTEXT_EXTRACTION_FILE=$rust_extraction_file cargo check
-    or exit 1
+    set -l --erase rust_extraction_file
```
krobelus: should be `set -l rust_extraction_file` I think (as written this statement is a nop)

build_tools/fish_xgettext.fish
```diff
@@ -15,18 +19,22 @@ begin
...
+        # We need to build to ensure that the proc macro for extracting strings runs.
+        FISH_GETTEXT_EXTRACTION_FILE=$rust_extraction_file cargo check
+        or exit 1
+    end
 
     # Get rid of duplicates and sort.
     msguniq --no-wrap --strict --sort-output $rust_extraction_file
     or exit 1
 
-    rm $rust_extraction_file
```
krobelus, line 28: I think if `--use-existing-template` is not given, we should delete this temporary file as before

build_tools/fish_xgettext.fish
danielrainer: I think now that the parallelized test driver is merged we can switch to `check.sh` in CI as well, possibly in this PR to avoid skipping translation checks anywhere.

Having native support for extracting arbitrary data would definitely be nice for this use case.

I did look into gettext alternatives and the only other option with decent Rust support seems to be fluent, which is far from ideal. (See https://github.com/fish-shell/fish-shell/issues/11400#issuecomment-2831714671 for details.) We could build our own system, but getting to a point where it is better than what we have now would take significant effort and would have the disadvantage that translators would have to learn how to deal with our custom format.

krobelus: > possibly in this PR to avoid skipping translation checks anywhere.

yeah, I'd either do that, or merge everything but the last commit, no strong opinion here

danielrainer: I removed the last commit. Making major changes to CI might bloat this PR and there is no real reason to do that rather than addressing it separately.

PR #11570, opened 2025-06-11
Support incrementing/decrementing the number below the cursor

### LINKED ISSUE #8320: vi mode: support increment/decrement of the next number

d-e-s-o:
Vim supports incrementing & decrementing the number below the cursor (or after it) via Ctrl-a and Ctrl-x, respectively. Given fish's Vi mode support, it makes sense to provide similar functionality when working on the command line, to provide a more natural environment for Vim users. With this change we add the necessary functionality.

Closes: #8320 

I tested this via https://github.com/d-e-s-o/fish-shell/commit/23d42a9812eb341b83840a9b418b448c396bb7b1 (and had it running for years), but the unit tests seem to require `fish` modification to work.

## TODOs:

### DIFF
--- a/share/functions/fish_vi_key_bindings.fish
+++ b/share/functions/fish_vi_key_bindings.fish
@@ -1,3 +1,80 @@
+alias fish_vi_dec 'fish_vi_inc_dec dec'
+alias fish_vi_inc 'fish_vi_inc_dec inc'
+
+# TODO: Currently we do not support hexadecimal and octal values.
+function fish_vi_inc_dec --description 'increment or decrement the number below the cursor'
+  # The cursor is zero based, but all string functions assume 1 to be
+  # the lowest index. Adjust accordingly.
+  set --local cursor (math -- (commandline --cursor) + 1)
+  set --local line (commandline --current-buffer | string collect)
+
+  set --local candidate (string sub --start $cursor -- $line | string collect)
+  if set --local just_found (string match --regex '^-?[0-9]+' -- $candidate)
+    # Search from the current cursor position backwards for as long as we
+    # can identify a valid number.
+    set --function found $just_found
+    set --function found_at $cursor
+    set --local end (math -- $cursor + (string length -- $found) - 1)
+
+    set i (math -- $cursor - 1)
+    while [ $i -ge 1 ]
+      set candidate (string sub --start $i --end $end -- $line)
+      if set just_found (string match --regex '^-?[0-9]+$' -- $candidate)
+        set found $just_found
+        set found_at $i
+        # We found a candidate, but continue to make sure that we captured
+        # the complete number and not just part of it.
+      else
+        # We have already found a number earlier. Work with that.
+        break
+      end
+
+      set i (math -- $i - 1)
+    end
+  else
+    # We didn't find a match below the cursor. Mirror Vim behavior by
+    # checking ahead as well.
+    for i in (seq (math -- $cursor + 1) (math -- (string length -- $line) - 1))
+      set candidate (string sub --start $i -- $line | string collect)
+
+      if set just_found (string match --regex '^-?[0-9]+' -- $candidate)
+        set found $just_found
+        set found_at $i
+        break
+      end
+    end
+
+    if [ -z "$found" ]
+      return
+    end
+  end
+
+  if [ $argv = 'inc' ]
+    set number (math -- $found + 1)
+  else if [ $argv = 'dec' ]
+    set number (math -- $found - 1)
+  end
+
+  set --local number_abs (string trim --left --chars=- -- $number)
+  set --local signed $status
+  set --local found_abs (string trim --left --chars=- -- $found)
+  set number (string pad --char 0 --width (string length -- $found_abs) -- $number_abs)
+  if test $signed -eq 0
+    set number "-$number"
+  end
+
+  # `string sub` may bitch about `--end` being zero if `found_at` is 1.
+  # So ignore errors here...
+  set --local before (string sub --end (math -- $found_at - 1) -- $line 2> /dev/null | string collect)
+  set --local after (string sub --start (math -- $found_at + (string length -- $found)) -- $line | string collect)
+  commandline --replace -- "$before$number$after"
+  # Need to subtract two here because 1) cursor is zero based 2)
+  # `found_at` is the index of the first character of the match, but we
+  # want the one before that.
+  commandline --cursor -- (math -- $found_at + (string length -- $number) - 2)
+  commandline --function -- repaint
+end
+
 function fish_vi_key_bindings --description 'vi-like key bindings for fish'
     if contains -- -h $argv
         or contains -- --help $argv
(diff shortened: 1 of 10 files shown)


share/functions/fish_vi_key_bindings.fish
```diff
@@ -1,3 +1,77 @@
+alias vi_dec 'vi_inc_dec dec'
+alias vi_inc 'vi_inc_dec inc'
```
krobelus: probably this should be called fish_vi_inc though I can see how that's ugly.

share/functions/fish_vi_key_bindings.fish
```diff
@@ -1,3 +1,77 @@
...
+
+# TODO: Currently we do not support hexadecimal and octal values.
+function vi_inc_dec --description 'increment or decrement a number'
```
krobelus: this should be namespaced like fish_vi_inc_dec

Would be great to include your tests from 23d42a9812eb341b83840a9b418b448c396bb7b1.

I wonder if there is a strong reason why "commandline" silently
fails when run from `config.fish` or `fish -C` etc.

I'd expect that `fish -ic 'commandline hello; commandline'` prints "hello".
Maybe that was not considered in
a32248277f0 (Make commandline state thread safe, 2021-07-20);
looking at this diff, it seems like before that commit, that would have worked fine (I didn't test that though).
If that commit was really only about thread safety, then changing this to always use the current state if `is_interactive_session()` seems fine.

An earlier commit 782a7397364 (Remove confusing and misleading error
message when trying to repaint the prompt during startup, 2007-08-02)
added a special case which we can probably remove then -- the commandline will always be initialized,
but starts at the empty string.

I'd also expect that `fish -C 'commandline echo'` does prefill the commandline with "echo".

So maybe we can improve things here.

Alternatively, we could rewrite the test as a real interactive one (`tests/pexpects/`)

share/functions/fish_vi_key_bindings.fish
```diff
@@ -1,3 +1,77 @@
...
+  # The cursor is zero based, but all string functions assume 1 to be
+  # the lowest index. Adjust accordingly.
+  set cursor (math (commandline --cursor) + 1)
```
krobelus: use `set -l ...` to avoid overwriting global or universal variables

share/functions/fish_vi_key_bindings.fish
```diff
@@ -1,3 +1,77 @@
...
+  set line (commandline --current-buffer)
```
krobelus: doesn't look like this works for multiline command lines.
We can probably fix this by adding `| string collect` inside every
command substitution that can produce multiline output.
AFAICT that's only this line and some of the "string sub" calls.

share/functions/fish_vi_key_bindings.fish
```diff
@@ -1,3 +1,77 @@
...
+
+  set candidate (string sub --start $cursor -- $line)
+  if set just_found (string match --ignore-case --regex '^-?[0-9]+' -- $candidate)
```
krobelus: we'll want to declare `set -l just_found` above this line, so both branches always use this local

share/functions/fish_vi_key_bindings.fish
```diff
@@ -1,3 +1,77 @@
...
+    # Search from the current cursor position backwards for as long as we
+    # can identify a valid number.
+    set found $just_found
+    set found_at $cursor
+    set end (math -- $cursor + (string length -- $found) - 1)
+
+    for i in (seq (math -- $cursor - 1) -1 1)
+      set candidate (string sub --start $i --end $end -- $line)
+      if set just_found (string match --ignore-case --regex '^-?[0-9]+$' -- $candidate)
```
krobelus: why the `--ignore-case`?

share/functions/fish_vi_key_bindings.fish
```diff
@@ -271,6 +345,12 @@ function fish_vi_key_bindings --description 'vi-like key bindings for fish'
     bind -s --preset -M replace backspace backward-char
     bind -s --preset -M replace shift-backspace backward-char
 
+    #
+    # increment or decrement numbers with ctrl+x ctrl+a
+    #
+    bind -s --preset -M default \ca vi_inc
+    bind -s --preset -M default \cx vi_dec
```
krobelus: use ctrl-a and ctrl-x notation

share/functions/fish_vi_key_bindings.fish
d-e-s-o: Seems fine to just add `--local` there, no need to move?

share/functions/fish_vi_key_bindings.fish
d-e-s-o: Not sure anymore. May not be necessary. Removed.

share/functions/fish_vi_key_bindings.fish
d-e-s-o: That's a lot of expectations :-) I've added a commit to adjust the commandline logic. Not sure if that's what you had in mind. Please take a look.

share/functions/fish_vi_key_bindings.fish
d-e-s-o: Added; and seems to work based on some limited testing.

d-e-s-o:
Seems MacOS `seq` requires positive increment or something like that, causing the new tests to fail. Any `fish` alternative that we could use?

krobelus:
> Seems MacOS seq requires positive increment or something like that, causing the new tests to fail. Any fish alternative that we could use?

Not sure. I guess either use `set i (math $i - 1)` or `seq 10 | tac`, or something in between

share/functions/fish_vi_key_bindings.fish
krobelus: I think that's correct yeah.. though I'm wondering why `fish -C 'commandline echo'` doesn't work;
and also we should probably get rid of `initialized` completely; there's only one other use in `builtins/complete.rs`.

I'll try to find time for that soon; if it takes too long we could also merge with the test disabled (`# REQUIRES: false`) and re-enable it later

d-e-s-o:
> > Seems MacOS seq requires positive increment or something like that, causing the new tests to fail. Any fish alternative that we could use?
> 
> Not sure. I guess either use `set i (math $i - 1)` or `seq 10 | tac`, or something in between

Thanks for the suggestion. I think that worked...but if I am reading [the other test failures](https://github.com/fish-shell/fish-shell/actions/runs/15685021241/job/44185900433?pr=11570#step:7:519) correctly, it appears as if perhaps cursor handling on MacOS behaves differently than on Linux?

Specifically,
```
commandline '123'; commandline --cursor 2; fish_vi_inc
commandline --current-buffer
# CHECK: 124                                 -> succeeds

commandline '123'; commandline --cursor 2; fish_vi_dec
commandline --current-buffer
# CHECK: 122                                 -> succeeds

commandline '123'; commandline --cursor 3; fish_vi_inc
commandline --current-buffer
# CHECK: 123                                 -> fails; actual value 124

commandline '123'; commandline --cursor 3; fish_vi_dec
commandline --current-buffer
# CHECK: 123                                 -> fails; actual value 122
```

Given that the only difference to the above is that we set the cursor behind the actual string, my guess would be that on MacOS that's not really what is happening. I don't have MacOS, so I am just making an educated guess based on observation. I removed these two tests for now, as they don't seem to add much value, in my opinion, but perhaps that points to an issue in `fish` itself?

share/functions/fish_vi_key_bindings.fish
d-e-s-o: Okay, let me know in case you want me to disable tests.

PR #11571, opened 2025-06-11
Split up sphinx HTML and MAN file generation

danielrainer:
This test is the one with the longest runtime. Splitting the two targets into separate tests allows them to run in parallel, which can speed up the tests.


tests/checks/check-sphinx.fish
```diff
@@ -1,33 +0,0 @@
...
-        -b $builder \
-        -c $docsrc \
-        -d $doctree \
-        $docsrc \
-        $output_dir
-end
-
-set -l success true
-build_docs man || set -l success false
-build_docs html || set -l success false
```
krobelus, line 27: yeah, this speeds things up a bit

Most of the time docs don't change,
so we could get a nice speedup if our
build system could remember that.

Ideally every test specifies explicitly all its dependencies
(files, environment variables, ...)  then it can be cached well, and run
in isolation with no access to any other file etc.

I'd probably want to avoid an "almost-correct" build system like
Make, although it could work fine...

tests/checks/sphinx-shared.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/sh
+
+set -e
```
krobelus: LGTM.
A helper file in tests/checks (or tests/pexpects/) is certainly a novelty.
The traditional approach would be to add a fish function to tests/test_functions/ but of course that's not as nice due to the lack of `set -e` and possibly others.
Eventually we might want to come up with a consistent structure for the support
files in tests/, that directory seems a bit cluttered.

tests/checks/sphinx-shared.sh
```diff
@@ -0,0 +1,28 @@
...
+cleanup () {
+  if [ -n "$tmp_dir" ] && [ -e "$tmp_dir" ]; then
+    rm -r "$tmp_dir"
+  fi
+}
+
+trap cleanup EXIT INT TERM HUP
+
+repo_root="$(dirname "$0")/../.."
+builder="$1"
```
krobelus: (FWIW in this special cause (variable assignment), quotes are redundant)

tests/checks/check-sphinx.fish
danielrainer: Not running tests when unnecessary would be good, although the effort for correctly documenting all dependencies (and maintaining this documentation), and having a system which uses this information is certainly non-trivial. I think it is definitely out of scope for this PR.

tests/checks/sphinx-shared.sh
danielrainer: I moved the `sphinx-shared.sh` script into `tests/test_functions`. It's not a fish function, but it still seems to be a more suitable place than `tests/checks`

PR #11572, opened 2025-06-11
Clear gettext cache on locale change

danielrainer:
This prevents printing cached localizations which do not match the current locale.

### DIFF
--- a/src/wutil/gettext.rs
+++ b/src/wutil/gettext.rs
@@ -70,6 +70,18 @@ enum MaybeStatic<'a> {
     Local(&'a wstr),
 }
 
+// Note that because entries are immortal, we simply leak non-static keys, and all values.
+static WGETTEXT_MAP: Lazy<Mutex<HashMap<&'static wstr, &'static wstr>>> =
+    Lazy::new(|| Mutex::new(HashMap::new()));
+
+/// Localizations for the current locale are cached.
+/// When the locale changes, this cache becomes stale and needs to be cleared to avoid showing
+/// localizations for the previous locale.
+pub fn wgettext_clear_cache() {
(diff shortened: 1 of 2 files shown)


src/wutil/gettext.rs
```diff
@@ -88,9 +100,6 @@ fn wgettext_impl(text: MaybeStatic) -> &'static wstr {
 
     debug_assert!(!key.contains('\0'), "key should not contain NUL");
 
-    // Note that because entries are immortal, we simply leak non-static keys, and all values.
-    static WGETTEXT_MAP: Lazy<Mutex<HashMap<&'static wstr, &'static wstr>>> =
-        Lazy::new(|| Mutex::new(HashMap::new()));
```
krobelus, line 93: Good change;

Both keys and values may be leaked `Box<wstr>` objects.
This could waste space and confuse dynamic analysis tools.
Even though it's probably fine in practice, we should either
1. free the memory (as suggested in https://github.com/fish-shell/fish-shell/pull/10477)
2. or keep the translations around forever (as before), but don't remove them from the global map.
   For example `HashMap<(LocaleId, &'static wstr), &'static wstr>`.
   Having a reference in a static variable presumably tells valgrind and
   LeakSanitizer that this is an intentional leak.

danielrainer: Interesting, I didn't realize that #10477 exists, which is pretty much exactly the same as this PR.

The first option seems hard to accomplish without performance degradation. As long as we convert stuff into static references, we can never free it. But using owned strings everywhere would cause a lot of unnecessary allocations.

I think the second option is preferable. Implemented in https://github.com/fish-shell/fish-shell/pull/11583. Closing this PR in favor of that one.

Issue #6264, opened 2019-10-30
In-tree builds are flagged dirty


zanchey:
Running CMake in-tree builds (as done in Homebrew, for example) overwrites the `Makefile` shipped in git, leaving to the build being flagged dirty:

```
> git clone https://github.com/fish-shell/fish-shell.git
Cloning into 'fish-shell'... done.
> cd fish-shell
> git status
On branch master
Your branch is up to date with 'origin/master'.

nothing to commit, working tree clean
> cmake .
...
> git status
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   Makefile
```

I thought of moving the shipped `Makefile` to `GNUmakefile`, but that will only make things worse.

Either we should remove the `Makefile` from git, or expressly not support in-tree builds. The former is my preference, though there is lots of code supporting the latter/allowing for both.

faho:
IMHO we should remove the Makefile (and the BSDmakefile), since it's just a dumb wrapper around cmake anyway.

faho:
Yeah, done.

zanchey:
@mqudsi added these, I think, so would appreciate his thoughts.

mqudsi:
Respectfully, I disagree for the same reasons that the makefiles were added:

```
commit 20db22f0fe25e7daaac11111aea1a14481d8ac5c
Author: Mahmoud Al-Qudsi <mqudsi@neosmart.net>
Date:   Tue Apr 9 19:58:48 2019 -0500

    Add basic `make` shim for CMake

    This lets non-developers simply `cd` into the fish source directory and
    execute `make` to build the project. The Makefile searches for CMake and
    hands over the build to it if it is available, otherwise an error
    message is emitted. All dependency checking is left to CMake.

    Non-fish-devs shouldn't have to concern themselves with what build
    system fish developers have chosen, and building a random C++ project
    should not be a chore in familiarizing one's self with all the various
    build platforms out there.
```

There is a cleaner fix here (which is also more semantically correct) to revert https://github.com/fish-shell/fish-shell/commit/326f5586deb33cc8d711d51b36a400c85d44185e and simply rename `Makefile` to `GNUmakefile` (to parallel `BSDmakefile`), which does not conflict with any of the CMake generated makefile names.

zanchey:
[...]

PR #11583, opened 2025-06-15
Locale aware messages cache

danielrainer:
## Description

This fixes missing cache invalidation for gettext localizations after changing the locale. This is the same issue which was noticed in https://github.com/fish-shell/fish-shell/pull/10477 and https://github.com/fish-shell/fish-shell/pull/11572.

The difference here is that instead of leaking the cache for the old locale, a separate cache for each locale is kept, meaning nothing is ever removed from the cache, so we do not get leaks.

Also add some basic testing for gettext localization. Because we need some locales to exist for this, the test will mostly run in CI (and the machines of people who happen to have the required locales available).


danielrainer:
Hm, translations don't seem to work in CI, even after installing. It seems that CMake  does not install the MO files in a location where `gettext` can find them. I'm not sure what's going wrong here and debugging CMake is a pain.

We could get around this by switching to non-CMake builds in CI, but while we still support CMake it might make sense to figure out what's going wrong here.

.github/workflows/main.yml
```diff
@@ -21,14 +21,17 @@ jobs:
...
     - name: cmake
       run: |
         mkdir build && cd build
         cmake .. -DCMAKE_BUILD_TYPE=RelWithDebInfo
     - name: make
       run: |
         make -C build VERBOSE=1
+    - name: install
+      run: |
+        sudo cmake --install build
```
krobelus: yeah this will install fish to /usr/local/,
but `tests/checks/gettext.fish` will be run isolated of that installation.
CMake tests used to install everything into a dedicated root ("build_root"),
but somehow that changed since 4.0.2.., I don't see "test_root" anymore.
I have no idea why and I haven't had time to bisect it.
If we figure that out, we can probably restore that,
so cmake actually tests a proper install, instead of relying
on our "relocatable" logic.

The relocatable logic (src/env/config_paths.rs) might also be wrong
or at least non-helpful in this case.
In particular, build/fish and build/share/ exist on my system,
but build/share/locale doesn't.. I wonder how that works for
things other than locale, such as functions/completions

src/wutil/gettext.rs
```diff
@@ -87,10 +95,18 @@ fn wgettext_impl(text: MaybeStatic) -> &'static wstr {
 
     debug_assert!(!key.contains('\0'), "key should not contain NUL");
 
+    // Cache keys to avoid leaking the them for each locale with which they are used.
+    static CACHED_KEYS: Lazy<Mutex<HashSet<&'static wstr>>> =
```
krobelus: we should emphasize that this optimization is only needed for dynamic translations,
either by picking a different name or by moving this declaration
into the dynamic-translation path

src/wutil/gettext.rs
```diff
@@ -87,10 +95,18 @@
...
+    // to a map from messages to their localizations.
+    // This allows supporting locale changes without having to leak the data for the old locale.
     // Note that because entries are immortal, we simply leak non-static keys, and all values.
-    static WGETTEXT_MAP: Lazy<Mutex<HashMap<&'static wstr, &'static wstr>>> =
+    static WGETTEXT_MAP: Lazy<Mutex<HashMap<CString, HashMap<&'static wstr, &'static wstr>>>> =
         Lazy::new(|| Mutex::new(HashMap::new()));
-    let mut wmap = WGETTEXT_MAP.lock().unwrap();
+    let mut all_locales_cache = WGETTEXT_MAP.lock().unwrap();
+    let locale = (*CURRENT_MESSAGES_LOCALE.lock().unwrap()).clone();
+    let wmap = all_locales_cache.entry(locale.clone()).or_default();
```
krobelus: we should probably avoid the clone / potential allocation on every call to wgettext.
Though I guess this one might be subject to SSO, if that's a thing for CString.

I'm not sure if rustc optimizes this already but we can do so explicitly (possible diff below).
Alternatively, we could make the locale key an immortal `&'static [u8]`, much like the translated strings themselves.

```diff
diff --git a/src/wutil/gettext.rs b/src/wutil/gettext.rs
index 0d420e13b2..0ae81430c7 100644
--- a/src/wutil/gettext.rs
+++ b/src/wutil/gettext.rs
@@ -105,8 +105,13 @@
     static WGETTEXT_MAP: Lazy<Mutex<HashMap<CString, HashMap<&'static wstr, &'static wstr>>>> =
         Lazy::new(|| Mutex::new(HashMap::new()));
     let mut all_locales_cache = WGETTEXT_MAP.lock().unwrap();
-    let locale = (*CURRENT_MESSAGES_LOCALE.lock().unwrap()).clone();
-    let wmap = all_locales_cache.entry(locale.clone()).or_default();
+    let locale = CURRENT_MESSAGES_LOCALE.lock().unwrap();
+    let mut wmap = all_locales_cache.get_mut(locale.as_ref());
+    if wmap.is_none() {
+        all_locales_cache.insert(locale.clone(), HashMap::new());
+        wmap = all_locales_cache.get_mut(locale.as_ref());
+    }
+    let wmap = wmap.unwrap();
     let res = match wmap.get(key) {
         Some(v) => *v,
         None => {
```

src/wutil/gettext.rs
```diff
@@ -103,7 +119,18 @@
             // Get a static key, perhaps leaking it into the heap as well.
             let key: &'static wstr = match text {
                 MaybeStatic::Static(s) => s,
-                MaybeStatic::Local(s) => wstr::from_char_slice(Box::leak(s.as_char_slice().into())),
+                MaybeStatic::Local(s) => {
+                    let mut cached_keys = CACHED_KEYS.lock().unwrap();
+                    if cached_keys.contains(s) {
+                        cached_keys.get(s).unwrap()
```
krobelus: can probably get rid of the unwrap() with

```
if let Some(v) = cached_keys.get(s) {
	v
} else {
	... insert
}
```

tests/checks/gettext.fish
```diff
@@ -0,0 +1,28 @@
+# RUN: env FISH=%fish %fish %s
+#REQUIRES: locale -a | grep 'de_DE'
+#REQUIRES: locale -a | grep 'en_US'
+
+set -l CUSTOM_GETTEXT_VAR asdf
```
krobelus: (nit, we usually use lowercase for unexported variables, and for fish=%fish)

.github/workflows/main.yml
danielrainer: This commit might be relevant: 7b3074580069facd79a719241cf2d921d7c8cec6.

krobelus: nice, maybe we should revert that until we remove cmake.
Not sure how close we are to that. 
Reverting should also fix `docker/docker_run_tests.sh`, which is useful for testing on old systems.

danielrainer: > Not sure how close we are to that.

I'm in the process of investigating this. Results of a quick scan of the files:
- `Benchmark`: That just adds a target which calls a script, should be simple to replace, e.g. with a `make` target.
- `Docs`: man pages are built in `build.rs`. Building HTML docs is simple (we already have a script for the sphinx tests). The dependency on the fish version identifier would still have to be addressed in standalone doc builds. I have not looked into doc installation for the case where we don't embed them (HTML docs are never embedded).
- `FindRust`: should be irrelevant.
- `gettext`: This is basically what this discussion is about. We can generate the files without CMake, but the installation part still needs to be addressed.
- `Install`: Most of this file probably has to be rewritten to work without CMake.
- `MacApp`: Also looks like it has to be rewritten.
- `PCRE2`: Essentially just setting an env var.
- `Rust`: Setting env vars and cargo args. Has to be rewritten.
- `Tests`: Can be replaced by a few cargo invocations + `test_driver.py` (following a build), essentially what's done in `check.sh`.
- `Version`: Wrapper around `git_version_gen.sh`. We have an alternative implementation in `build.rs`. I think the only thing which requires version info but does not run `build.rs` are the docs. We might also want to consider how this works if we don't build within a git dir but from e.g. a release archive containing the sources.
- `CMakeLists`: Doesn't seem to do much other than including the other files and calling cargo.

danielrainer: Some of the complexity probably also depends on whether we want to continue supporting builds with `embed-data` disabled.

krobelus: > Some of the complexity probably also depends on whether we want to continue supporting builds with embed-data disabled.

Long-term probably not.
One potential problem with adoption is the breaking change of `$__fish_datadir` (it's dundered but still used..).
This has the problem to be painful for a lot of users
Especially if used in plugins, for example https://github.com/ajeetdsouza/zoxide/pull/1046

So for now I expect we keep the no-embed-data option for now.

We should remove CMake, but probably only once we can remove it completely (i.e. remote `CMakeLists.txt`),
so any existing cmake builds fail early.

Until then, we could either duplicate functionality (like we did for `git_version_gen.sh`/`build.rs`),
but I'd prefer if we extracted a script, and used it from both `build.rs` (or `build_tools/check.sh` etc.) as well as from CMake. Once CMake files contain only CMake-specific logic, we can delete them outright.

As for this PR, I have no strong opinion on whether to get the test-install back or fix the relocatable logic
Though this relocatable logic (i.e. the `else if exec_path.ends_with("fish")` codepath from `config_paths.rs`)
is something that we can remove as soon as we drop cmake builds, because AFAICT it's only useful for in-tree cmake builds.
Probably `cargo` doesn't even support those. `CARGO_TARGET_DIR=$PWD cargo b` will put things in `$PWD/debug`.
Getting the test-install back might help us catch some issues when we rewrite `Install.cmake`, I guess

danielrainer: > if we extracted a script, and used it from both

Yes, I think this is the most sensible option.

> As for this PR

I have not looked into the different installation methods enough to have an opinion on what makes most sense here.

krobelus: changing the test to do `# RUN: fish=%fish %fish -d config -o /tmp/fifo %s`
shows that both `target/debug/fish` and `build/fish` use
`paths.locale: /home/johannes/git/fish-shell/share/locale`

I guess that's not wrong (even though `build/fish` is broken in other regards; for example man pages are missing (try `build/fish -h`).. we should phase this out).

Running `build_tools/update_translations.fish` populates `share/locale`, making
`ninja -Cbuild test_gettext.fish` pass. We could maybe add the mo-generation part to the CI, if it's not too hacky

I have an unfinished fix for the docker tests

krobelus: so `${CMAKE_BINARY_DIR}/fish` (AKA `build/fish`) currently uses `*.mo` files from `share/locale`.
That's also where  `build_tools/update_translations.fish --only-mo` writes to.

It's rather weird that compiled output shows up outside `build` or `target`.
This will cause problems with docker containers that mount the source tree as
read-only; we'll make the cargo build still by setting `CARGO_TARGET_DIR`.

`embed-data` builds already write man pages to `${CARGO_TARGET_DIR}/man`
(should we rename this to `${CARGO_TARGET_DIR}/fish-man` to clarify that this fish-specific?)

We could do either:
1. embed `*.mo` files unconditionally (even if `embed-data` disabled)
2. Teach 
   2a. `${CMAKE_BINARY_DIR}/fish` to use `${CMAKE_BINARY_DIR}/share/locale`
   2b. `${CARGO_TARGET_DIR}/*/fish` to use `${CARGO_TARGET_DIR}/fish-locale`

I wonder if there are problems with option 1.
I'd imagine that for consistency, distributions would prefer to keep
`/usr/share/locale/*/LC_MESSAGES/fish.mo`.
That will also allow them to easily package individual translations separately.

Embedding unconditionally might be fine.. but not knowing better I'd go with option 2 then we can decide later.
That would also imply that we need to run `--only-mo` from `build.rs` (respecting `CMAKE_BINARY_DIR`).

---

So I'm assuming that non-embedded builds will remain default for distributions for at least a while.
I wonder what the build steps will look like?

One possibility:
- `cargo install --path .` for embedded builds
- `make && make install PREFIX=/usr` for non-embedded builds (or `./build.sh && `./install.sh --prefix=/usr`?)

danielrainer: > It's rather weird that compiled output shows up outside build or target.

Fair point, I originally decided to put them there directly after finding out that this is a location where fish can find it (at least if built without cmake), but I did not put much more thought into it that this.

> 1. embed *.mo files unconditionally (even if embed-data disabled)

The issue with that is that the system's gettext functions need to load actual MO files, so if we embed the data we would still have to extract them into normal files on the filesystem. Given that the normal lookup locations are in root-owned directories, it is probably not realistic to expect that this will work.
https://github.com/fish-shell/fish-shell/pull/11475 is an attempt at removing the runtime dependency on gettext, but that's not merged yet (for good reasons, it's mostly a proof-of-concept, with multiple issues).

> That will also allow them to easily package individual translations separately.

Yes, that allows reducing bloat, and possibly even custom modifications to the translations. I don't know if anyone would want to do the latter, but it's nice to have the option.

I agree that the second option is preferable, at least for now, but I have not really looked into the details of implementing it yet.

> `cargo install --path .` for embedded builds

Using only cargo for builds only works if we can do everything other than compiling our Rust code in `build.rs`, since cargo does not have a way of running custom code after a build completes. This is a problem for everything which needs one of the build artifacts as a dependency. E.g. HTML doc generation needs `fish_indent`. We might not care about this for embedded builds, since browsers can't deal with files embedded into the fish binary. Another issue would be MO file generation, which currently happens in a fish script. This can be changed fairly easily, but there might be other issues for which it might be harder or even impossible. So I think even if we get rid of CMake completely, we will probably still need a build system other than cargo. `make` would probably be sufficient, although it is annoying to have both `GNUmakefile` and `BSDmakefile`, requiring everything to be duplicated (possibly with subtle differences, which make it easy to introduce inconsistencies). Maybe the work required outside of what cargo does is simple enough that we can get away with a custom script as a `make` replacement.

krobelus: > have both GNUmakefile and BSDmakefile

If we use a Makefile, we would write it in the POSIX Make subset.

danielrainer: If that's possible it's definitely preferable. I have not looked into it much yet, but why do we have the 2 separate Makefiles? Do we need some features which are supported by both GNU and BSD make but not POSIX?

krobelus: I don't think there's a good reason to require GNU Make.
E.g. 3a2c3c2288d (Remove some GNU-makeisms from the Makefile to try to get us to build with BSD make. Not quite there yet., 2013-03-30)
Especially if our Makefiles are simple wrappers.

krobelus: Instead of installing to `/usr/local`,
we could try something like this.

Unfortunately it still doesn't make
`LANG=de_DE.utf8 build/fish -d config -c 'set -S HOME'`
work, whereas the equivalent command does work with `target/debug/fish`.
No idea why.

Could also wait with the test until we get rid of CMake.

```diff
diff --git a/build.rs b/build.rs
index a002bdd960..375b8595ff 100644
--- a/build.rs
+++ b/build.rs
@@ -24,6 +24,10 @@
         .map(canonicalize)
         .unwrap_or(canonicalize(MANIFEST_DIR).join("target"));
 
+    if option_env!("FISH_BUILD_DIR").is_some() {
+        rsconf::set_env_value("FISH_CMAKE", "true");
+    }
+
     // FISH_BUILD_DIR is set by CMake, if we are using it.
     rsconf::set_env_value(
         "FISH_BUILD_DIR",
diff --git a/build_tools/update_translations.fish b/build_tools/update_translations.fish
index d0444441c5..624fddc384 100755
--- a/build_tools/update_translations.fish
+++ b/build_tools/update_translations.fish
@@ -6,7 +6,7 @@
 # updates the PO files for each language from that
 # (changed line numbers, added messages, removed messages),
 # and finally generates a machine-readable MO file for each language,
-# which is stored in share/locale/$LANG/LC_MESSAGES/fish.mo (relative to the repo root).
+# which is stored in target/fish-locale/$LANG/LC_MESSAGES/fish.mo (relative to the repo root).
 #
 # Use cases:
 # For developers:
@@ -131,6 +131,13 @@
     end
     if set -l --query mo
         set -l locale_dir $build_tools/../share/locale
+        set -l locale_dir (
+            if set -q CARGO_TARGET_DIR
+                echo $CARGO_TARGET_DIR
+            else
+                echo $build_tools/../target
+            end
+        )/fish-locale
         set -l out_dir $locale_dir/(basename $po_file .po)/LC_MESSAGES
         mkdir -p $out_dir
         msgfmt --check-format --output-file=$out_dir/fish.mo $po_file
diff --git a/cmake/gettext.cmake b/cmake/gettext.cmake
index 9e6986af65..21bfe8bd11 100644
--- a/cmake/gettext.cmake
+++ b/cmake/gettext.cmake
@@ -17,6 +17,9 @@
     # cmake/Install.cmake instead of using INSTALL_DESTINATION
     gettext_process_po_files(${lang} ALL
                              PO_FILES po/${lang}.po)
+    execute_process (
+        COMMAND sh -c "mkdir -p fish-locale/${lang}/LC_MESSAGES; cp ${lang}.gmo fish-locale/${lang}/LC_MESSAGES/${lang}.mo"
+    )
   endforeach()
   set(CMAKE_FOLDER)
 endif()
diff --git a/src/env/config_paths.rs b/src/env/config_paths.rs
index ad88fa7f2d..5030aa1a65 100644
--- a/src/env/config_paths.rs
+++ b/src/env/config_paths.rs
@@ -36,16 +36,23 @@
             let manifest_dir = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
             FLOG!(
                 config,
-                "Running out of target directory, using paths relative to CARGO_MANIFEST_DIR:\n",
+                "Running out of build directory, using paths relative to CARGO_MANIFEST_DIR:\n",
                 manifest_dir.display()
             );
             done = true;
+            let locale_dir = if option_env!("FISH_CMAKE").is_some() {
+                PathBuf::from(env!("FISH_BUILD_DIR"))
+            } else {
+                option_env!("CARGO_TARGET_DIR")
+                    .map_or_else(|| manifest_dir.join("target"), PathBuf::from)
+            }
+            .join("fish-locale");
             paths = ConfigPaths {
                 data: Some(manifest_dir.join("share")),
                 sysconf: manifest_dir.join("etc"),
                 doc: manifest_dir.join("user_doc/html"),
                 bin: Some(exec_path.parent().unwrap().to_owned()),
-                locale: Some(manifest_dir.join("share/locale")),
+                locale: Some(locale_dir),
             }
         }
 
@@ -83,7 +90,7 @@
                     sysconf: base_path.join("etc"),
                     doc: base_path.join("user_doc/html"),
                     bin: Some(base_path.to_path_buf()),
-                    locale: Some(data_dir.join("locale")),
+                    locale: Some(data_dir.join("locale")), // TODO
                 }
             }
 
```

danielrainer: I looked into MO file lookup and installation some more and found a way to make if work for:
- cmake builds (if the install prefix is kept consistent)
- cargo-only builds while the binary is in `target/`

Localizations do not work for:
- cargo-only builds when the binary is moved out of `target/`, e.g. by `cargo install`. Fixing this is non-trivial and requires embedding the MO files in the binary and either writing them out again to be used by the system's gettext or replacing the system's gettext with a Rust implementation included in our binary which can access the embedded data. Embedding would also require compiling twice, once to extract up-to-date strings, then again to embed updated MO files.

Right now `update_translations.fish` is called 3 times during a cmake build, due to the `create_target` function being used to define the `fish` target. We could define a `fish` target without using this function, which would allow only running `update_translations.fish` once. This also prevents the potential issue of target reordering (I have not observed it using make, but maybe this could happen with ninja), causing the script to be invoked before the fish binary is available. The downside of this is duplicating the logic in the `create_target` function.

Tests on macOS seem to fail. I added a commit for debugging in CI and it looks like `/usr/share/locale/` does not contain simple 2-letter language identifiers without a region. Now, these directories are not used in the test, since these try to use `build/fish-locale`. But it makes me suspect that the gettext implementation there does not support language identifiers without a region code attached. Can someone confirm this? If this is really the case, have our translations ever worked on macOS? And how do we deal with it? Disable the gettext tests there for now? Use region IDs for our PO files? Switch to only supporting embedded gettext (that would require quite a bit of effort)?

Don't merge until the macOS issues are figured out. On FreeBSD, gettext tests don't run at the moment (`msgfmt` is missing, and possibly more). Should we install the required package(s) there?

krobelus: Haven't taken a look at the rest; FWIW one option for debugging macos is https://github.com/marketplace/actions/debugging-with-tmate, to ssh into the github runner. Pushing to a fork should also trigger actions.
An actual mac is more convenient though.
We could disable the test on macOS. Maybe look into how other software (firefox, chrome, ..) is translated there?

danielrainer: Firefox seems to use Fluent and before that they apparently also did not use gettext: https://firefox-source-docs.mozilla.org/l10n/fluent/index.html. I guess browsers which support non-POSIX environments have to use something other than gettext anyways, unless they bundle gettext.

If all it takes to make it work on macOS (and maybe other systems which don't use GNU gettext) is switching to language identifiers with a region specifier, that should be a fairly simple change for us. We would have to decide which variant to use for the language where we don't specify one at the moment, i.e. de, en, fr, pl, sv. It might annoy some people if we explicitly don't use their preferred variant, but being explicit does not change the file content and it helps translators to know which variant should be used. If someone wants to add a second variant for a language, they are free to do so.

Looking at https://www.gnu.org/software/gettext/manual/html_node/Locale-Names.html, it seems that the variant specifier is expected by GNU gettext and omitting it is not even mentioned in the docs AFAICT.

Given this info, I suggest using `de_DE`, `en_US`, `fr_FR`, `pl_PL`, `sv_SE`. If other variants are more appropriate let me know.

> For some languages, a two-letter code does not exist, and a three-letter code is used instead.

suggests that our regex for filename checking in `update_translations.fish` is too restrictive.

It might also make sense to at the `.UTF-8` suffix to the locale name in `share/locale`.

danielrainer: Hm, I just tested using a different language variant than we define and apparently the fallback behavior is not translating, rather than using a different variant of the same language. Compare (assuming the `wait` command fails):
```fish
target/debug/fish -c 'set -lx LANG pt_PT.utf8; wait 10000'
target/debug/fish -c 'set -lx LANG pt_BR.utf8; wait 10000'
```
(I needed to generate the `pt_BR.UTF-8` locale for translation to work at all.)

This behavior makes it significantly more annoying for us to specify variants. As a workaround,
```fish
target/debug/fish -c 'set -lx LANGUAGE pt_PT.utf8:pt_BR.utf8; wait 10000'
```
does work. Not sure what to make of that.

danielrainer:
Having reconsidered this problem, I now think that caching translation mappings in our code is not worth it, unless we stop relying on an external gettext implementation. Mappings are already cached externally, so calling the system's gettext should not be expensive, and we don't have to deal with annoying cache invalidation issues on locale variable changes. Caching the results of gettext calls does still make sense to avoid leaking the same message repeatedly.

I finally figured out why the gettext test fails on macOS: The `gettext` feature is disabled there. We either need to modify `have_gettext()` in `build.rs` or find a way to only run the gettext test when gettext is available. For FreeBSD, I have no idea what is going wrong. The `panic!` I put into the latest commit does not run, which indicates that the `gettext` feature is enabled there.

The nested loop logic in `has_gettext()` looks a bit suspicious and this function seems to have been written when parts of fish were still C++, so some changes might be in order there.

The following are some observations I made before figuring out that the feature is disabled on macOS, but they might still be relevant for figuring out what's going wrong with FreeBSD.

I also tried to make this work on macOS and FreeBSD, which has not been successful so far. Using syscall tracing in CI reveals that both macOS and FreeBSD behave in a very weird manner. I confirmed that the locale path shown by `fish --debug=config` points to the build directory on both platforms, as it should. However, instead of trying to open a subdirectory for localization, the standard location at `/usr/share/locale` is used and instead of using the `LC_MESSAGES` category, `LC_COLLATE` is used. I have no idea why this happens and debugging by pushing stuff to CI is getting annoying.
For examples, see:
- https://cirrus-ci.com/task/4801703033176064?logs=tests#L482
- https://github.com/danielrainer/fish-shell/actions/runs/16556866713/job/46819532753#step:7:313 (This tries to open the same file twice and then tries `/usr/local/share/locale` as a fallback.)
- https://github.com/danielrainer/fish-shell/actions/runs/16556975494/job/46819836100#step:7:329 (Here I checked what happens if I set `LC_MESSAGES` instead of `LANG` in the test. Then the path becomes `/usr/share/locale/de_DE.utf8/LC_MESSAGES/LC_MESSAGES`)

On FreeBSD and macOS, `LOCALEDIR` is set to `/usr/share/locale`, whereas Ubuntu uses `/usr/local/share/locale` when building fish:
- https://cirrus-ci.com/task/5477447216398336?logs=tests#L140
- https://github.com/fish-shell/fish-shell/actions/runs/16557788569/job/46821904249?pr=11583#step:6:62
- https://github.com/fish-shell/fish-shell/actions/runs/16557788569/job/46821904248?pr=11583#step:6:94

But these don't seem to be relevant when the binary is running in its build directory, as evidenced by the fact that in the debug print output `paths.locale` points into the build dir.

krobelus:
> Mappings are already cached externally [..]  Caching the results of gettext
> calls does still make sense to avoid leaking the same message repeatedly.

ok so the change would reduce to "clearing the cache on locale variable change",
which also means we won't leak anything (except for the things current reachable by our global map)

> debugging by pushing stuff to CI is getting annoying.

yeah, better set up an edit-recompile-test workflow.

For FreeBSD (which isn't supported on Github Actions), I think good approaches are
- either use a CI service that allows logging in via SSH
  - Cirrus does (haven't done that myself), see https://cirrus-ci.org/blog/2021/08/06/introducing-cirrus-terminal-a-simple-way-to-get-ssh-like-access-to-your-tasks/
  - I use https://builds.sr.ht/, e.g. `hut builds submit --follow .builds/freebsd.yml`
- or set up a VM (or real system). Haven't done that myself.

For macOS, options are:
- set up [tmate](https://github.com/marketplace/actions/debugging-with-tmate) in the gihub workflow to SSH into failing CI runs.
- or buy the hardware -- it's very useful for debugging macOS-specific issues;
  some of those can't be reproduced via headless SSH.
  Looking back I should have probably bought a used Mac for that alone.
  We might be able to reimburse the costs with donation money if we can get that set up

krobelus:
I think to do the "Re-Run with Terminal Access" in Cirrus, you could
remove the `only_if: $CIRRUS_REPO_OWNER == 'fish-shell'` from `.cirrus.yml`,
make an account on `cirrus-ci.com`,
and add your `fish-shell` fork in Cirrus, then pushes to your fork (with the modified `.cirrus.yml`) should have Cirrus checks visible in
https://github.com/danielrainer/fish-shell/tree/locale_aware_messages_cache
after clicking on the x (for the failing checks).
We can probably also add you to the fish-shell team so you can do "Re-Run with Terminal Access" on `fish-shell/fish-shell` runs

danielrainer:
> ok so the change would reduce to "clearing the cache on locale variable change"

No. That's what I originally did in this PR, but now the cache in our code never gets cleared, so we don't need to care about locale variable changes. The `MESSAGE_CACHE` is no longer used to avoid calling the system's `gettext`. Instead, we always call `gettext` (which has its own caching). The result of the `gettext` call will be a pointer to either a static string (the translation, if it exists), or to the input string (which is not always static). This "string"/`char *` is then converted into a `WString` (which requires allocation). Because we want to return a `&'static wstr`, we leak the `WString` to convert it accordingly. The `MESSAGE_CACHE` is used to store these references, which allows us to check if we already have a message cached, so we can avoid leaking it and instead return the cached reference.

Thanks for the suggestions regarding debugging on FreeBSD and macOS. I'll see what works best for me.

danielrainer:
## Results on a FreeBSD VM

Compared to the steps in the Cirrus file, I needed to install Python. Then, fish builds fine using CMake (not tested with cargo only).

The only locale format which seems to be accepted seems to be `ll_CC.UTF-8`, e.g. `de_DE.UTF-8`. Using this value for `LANG`, `LC_MESSAGES`, or `LC_ALL` results in translations working if they are set before fish starts. E.g. `LANG=de_DE.UTF-8 ./fish -c end` shows a German error message.
Translations also work like this: `./fish -c 'set -lx LANG de_DE.UTF-8; set --show LANG; set -lx LANG en_US.UTF-8; set --show LANG'`, which first prints the (barely translated) German version, followed by the English one.

I expected the test to work after changing the format of the locale variables as described above. However, that does not work.

Running it manually (by setting `$fish` appropriately and then running `$fish tests/check/gettext.fish`) results in all output being in English. Investigating with `truss` shows that it tries several directories for finding `fish.mo`, differing only in the locale name. The order:
1. `de_DE.UTF-8` (the exact value of `LANG` in the tested version of the test script), fails because the directory does not exist. (We can choose the directory names in `build/fish-locale`, currently `ll_CC.utf8` is used.)
2. `de_DE.utf8` succeeds, since that corresponds to the path we created. The file is mmaped and the fd closed.
3. For some reason, more directories are tired: `de_DE`, `de.UTF-8`, `de.utf8`, `de`, which all fail, since they don't exist.

These `openat` syscalls are immediately followed by the output of running the fish script, which is completely in English. After `LANG` is set to `en_US.UTF-8` in the script, the same process of trying to open `fish.mo` in differently named directories happens again, with the same results.

It gets very strange when running the test via the test driver. There is no complaint about the `CHECKERR` annotations, which makes it look like the corresponding messages are localized correctly. The output of `set --show custom_gettext_var` is partially localized as `$custom_gettext_var: set in local scope, nicht exportiert, with 1 elements` (first line of the output). The corresponding source location is in `src/builtins/set.rs:647`, which is a format string that has a German translation (which is actually in German, unlike the output shown above). The `exportv` argument however, is translated into German, resulting in the `nicht exportiert` above.

I might check the gettext source code to find out more about what's going on. At some point it might be less effort to include our own MO file parser and always bundle the MO files in the executable.

Another thing I noticed is that when using fish interactively in my FreeBSD VM, the default greeting and prompt are printed as expected, but entering commands does nothing (the entered characters are not even printed). Entering `ctrl-C` seems to fix the issue. From then on, fish can be used normally.

danielrainer:
The last point is addressed in https://github.com/fish-shell/fish-shell/issues/11723.

danielrainer:
I added an alternative approach which does not rely on the gettext lib at runtime in #11725.

krobelus:
> It gets very strange when running the test via the test driver. There
> is no complaint about the `CHECKERR` annotations, which makes it
> look like the corresponding messages are localized correctly. The
> output of `set --show custom_gettext_var` is partially localized
> as `$custom_gettext_var: set in local scope, nicht exportiert,
> with 1 elements` (first line of the output). The corresponding
> source location is in `src/builtins/set.rs:647`, which is a format
> string that has a German translation (which is actually in German,
> unlike the output shown above). The `exportv` argument however,
> is translated into German, resulting in the `nicht exportiert` above.

hmm that sounds worse than no translation (it's probably reproducible
without the test driver; IIRC test_driver.py sets locale to C.UTF-8.
I guess the same issue happens on Linux?).
In general, German translations seem a bit awkward and inconsistent, I wonder if they help.

> Another thing I noticed is that when using fish interactively in my
> FreeBSD VM, the default greeting and prompt are printed as expected,
> but entering commands does nothing (the entered characters are not
> even printed). Entering `ctrl-C` seems to fix the issue. From then
> on, fish can be used normally.

Sounds like this FreeBSD terminal fails to reply to the primary device attribute request (see `doc_src/terminal-compatibility.rst`) which was added after 4.0.
"fish -d reader" should confirm this.

We should fix this terminal (which one is it?) and also add a timeout + error message to degrade more gracefully (https://github.com/fish-shell/fish-shell/issues/11108).

danielrainer:
> IIRC test_driver.py sets locale to C.UTF-8

I don't think so. The `makeenv` function deletes `LANGUAGE` and all variables starting with `LC_`. Then it sets `LC_CTYPE=en_US.UTF-8` (which should not impact translations), and `LANG=C`. This variable definitely has an impact on translations. I don't know why `C` is used instead of `C.UTF-8`, and why it is set. But `LANG` should have the lowest priority, so when we are interested in translations, we will always override it, either by changing `LANG` as is done in `tests/checks/gettext.fish`, or by setting `LC_MESSAGES`, `LC_ALL`, or `LANGUAGES`.

I noticed that `tests/checks/set.fish` sets `LC_ALL` as a universal variable, which seems like it could cause problems if it runs at the same time as other tests relying on locale variables, but the test driver gives each tests its own `HOME` and config dirs, so universal variables from one test should not have an effect on other tests.

But in general, I think that it makes sense to move away from system gettext, so we can avoid annoying platform-dependent issues and can have translations in cargo-only builds when the binaries are moved. See #11726.

.github/workflows/main.yml
zanchey: BSD Make and GNU Make have different syntax for conditionals 🙃

3a2c3c2 is for the old autoconf-based Makefile, the BSD/GNUMakefiles are newly written to provide CMake shortcuts. @mqudsi wrote them from 20db22f0fe25e, see #6264 for the rationale

danielrainer: Yeah, such incompatibilities are quite annoying. But if we stop using CMake a lot of the logic in the current `{GNU,BSD}makefile`s is redundant, and we don't have to worry about in-tree builds anymore.

krobelus: right, there's a precedence hack here; though we could still have written them in POSIX make (and make one of them a symlink to the other)

danielrainer: I think the simplest approach is to leave the Makefiles we have now as they are and write a new POSIX Makefile that calls cargo directly instead of using CMake. That is, if we decide to use make at all. My preference would be to only use cargo, but for some things that might not be a good choice, like updating the version file, running checks, or building HTML docs.

krobelus: I would only use Make (as opposed to shell scripts) if we really need the mtime-based caching on top of what cargo already does. I think HTML docs could be built by a cargo crate that depends on `fish_indent`.. if that's not too much abuse

danielrainer: Yes, I don't think we will need Make. We might want to use `cargo xtask`, which seems especially useful for ensuring the correct version info is built into the binaries. Tasks like installing and building HTML docs are probably also better done this way rather than using some hacky cargo-feature-based approach.

PR #11588, opened 2025-06-17
Decode `WString` in `FLOG`

danielrainer:
This allows converting non-UTF-8-conforming bytes from their PUA encoding back to the original bytes before outputting.

Due to the way Rust handles trait implementations, we cannot use `impl<T: std::fmt::Display> FloggableDisplay for T` anymore, as that would result in a conflicting implementation for the types which get a custom implementation.
Instead, explicitly implement the trait for all types which need it.

I've used this for debugging. Not sure if we actually want this behavior for normal logging.

### DIFF
--- a/src/flog.rs
+++ b/src/flog.rs
@@ -1,3 +1,4 @@
+use crate::common::wcs2string;
 use crate::wchar::prelude::*;
 use crate::wildcard::wildcard_match;
 use crate::wutil::write_to_fd;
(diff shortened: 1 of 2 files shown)


src/flog.rs
```diff
@@ -149,30 +150,68 @@ pub mod categories {
...
-    fn to_flog_str(&self) -> String;
+    fn to_flog_str(&self) -> Vec<u8> {
+        self.to_string().into_bytes()
+    }
+}
+
+// Special handling for `WString` to decode PUA codepoints back into the original bytes.
+impl FloggableDisplay for WString {
+    fn to_flog_str(&self) -> Vec<u8> {
+        wcs2string(self)
```
krobelus, line 163: > I've used this for debugging. Not sure if we actually want this behavior for normal logging.

printing the original characters seems correct; definitely better than printing weird PUA characters.
I don't see what else "normal logging" could use.. I guess some use cases would prefer to see escaped versions of the bytes (`\xff` etc.) but that's weird because we don't eescape anything else here.

src/flog.rs
```diff
@@ -149,30 +150,68 @@ pub mod categories {
...
+    }
 }
 
-impl<T: std::fmt::Display> FloggableDisplay for T {
-    fn to_flog_str(&self) -> String {
-        self.to_string()
+impl FloggableDisplay for &wstr {
+    fn to_flog_str(&self) -> Vec<u8> {
+        wcs2string(self)
```
krobelus, line 169: LGTM.
I think printf has a similar problem;
When/if we change `WString` to a different type whose Display implementation calls `wcs2string`, then we can revert this PR.
But I don't know if that will happen, or when.

danielrainer: That cannot happen, because `Display` works with Rust strings, which must be valid UTF-8. This is the reason the return types are changed to `Vec<u8>` in this PR.

krobelus: Right, `sprintf!("%s", fishstr::from_char_slice([0xff, 0xff]))` could be fixed by adding a `ToArg` implementation that outputs those exact bytes.
I guess we have always been outputting random PUA bytes in some `sprintf!()` code paths, even before the Rust port, that's why there's no rush in fixing this.

PR #11593, opened 2025-06-19
Simplify CMake Tests

danielrainer:
Remove dependency on CTest. Parallel execution is handled by `test_driver.py` internally now, so CTest is no longer relevant for performance.

This also removes CMake targets for single tests. As a replacement, `test_driver.py` can be called directly with the path to the build directory as the first argument and the path to the desired test as the second argument. Ensuring that the executables in the build directory are up to date needs to be done separately.
For a pure cargo build, an example of running a single test would be: `cargo b && tests/test_driver.py target/debug tests/checks/abbr.fish`

The recommended way of running tests is `build_tools/check.sh`, which runs more extensive tests and does not depend on CMake. That script does not work in CI yet, so CMake testing is retained for now.

## TODOs:
- [ ] Should any documentation be updated for these changes?

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -26,7 +26,7 @@ linux_task:
         - lscpu || true
         - (cat /proc/meminfo | grep MemTotal) || true
         - mkdir build && cd build
-        - cmake -G Ninja -DCMAKE_BUILD_TYPE=Debug -DCTEST_PARALLEL_LEVEL=6 ..
+        - FISH_TEST_MAX_CONCURRENCY=6 cmake -G Ninja -DCMAKE_BUILD_TYPE=Debug ..
         - ninja -j 6 fish
         - ninja fish_run_tests
     only_if: $CIRRUS_REPO_OWNER == 'fish-shell'
@@ -45,7 +45,7 @@ linux_arm_task:
         - lscpu || true
         - (cat /proc/meminfo | grep MemTotal) || true
         - mkdir build && cd build
-        - cmake -G Ninja -DCMAKE_BUILD_TYPE=Debug -DCTEST_PARALLEL_LEVEL=6 ..
+        - FISH_TEST_MAX_CONCURRENCY=6 cmake -G Ninja -DCMAKE_BUILD_TYPE=Debug ..
         - ninja -j 6 fish
         - file ./fish
         - ninja fish_run_tests
@@ -54,17 +54,11 @@ linux_arm_task:
 
 freebsd_task:
     matrix:
-        # - name: FreeBSD 14
-        #   freebsd_instance:
-        #       image_family: freebsd-14-0-snap
-        - name: FreeBSD 13
+        - name: FreeBSD 14
           freebsd_instance:
-              image: freebsd-13-2-release-amd64
-        # - name: FreeBSD 12.3
-        #   freebsd_instance:
-        #       image: freebsd-12-3-release-amd64
+              image: freebsd-14-3-release-amd64-ufs
     tests_script:
-        - pkg install -y cmake-core devel/pcre2 devel/ninja misc/py-pexpect git-lite terminfo-db
+        - pkg install -y cmake-core devel/pcre2 devel/ninja lang/rust misc/py-pexpect git-lite
         # libclang.so is a required build dependency for rust-c++ ffi bridge
         - pkg install -y llvm
         # BSDs have the following behavior: root may open or access files even if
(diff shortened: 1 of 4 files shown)


danielrainer:
The increased parallelism seems to be an issue for several tests.

Making all of them timing-independent seems challenging, since many explicitly rely on timing (e.g. `wait.py`). We could ignore such tests in CI, but the better approach is probably to limit parallelism in `test_driver.py`.

cmake/Tests.cmake
```diff
@@ -1,92 +1,7 @@
-# This adds ctest support to the project
-enable_testing()
-
-# By default, ctest runs tests serially
-# Support CTEST_PARALLEL_LEVEL as an environment variable in addition to a CMake variable
-if(NOT CTEST_PARALLEL_LEVEL)
```
krobelus, line 6: should also remove `CTEST_PARALLEL_LEVEL` from `.github/workflows/main.yml`,
and, if we can't fix tests, pass that level to `test_driver.py`.
CI systems are have much less resources than developer machines;
they are virtualized guests competing with other CI runners.

Looking at the macos failure where only `pexpects/wait.py` seems to fail.
Might need more time (if it's starved of CPU), or it's broken for a different reason

tests/pexpect_helper.py
```diff
@@ -25,7 +25,7 @@
 from signal import Signals
 
 # Default timeout for failing to match.
-TIMEOUT_SECS = 5
+TIMEOUT_SECS = 600
```
krobelus: if it helps CI, increasing the timeout would be okay,
but we should definitely not have such a large default timeout for local runs
(and probably also not for individual tests).
So either make this conditional on `"CI" in os.environ` or remove it.
If we keep it, it's probably worth reducing; 10 minutes seems too long.

cmake/Tests.cmake
danielrainer: I made it possible to limit the concurrency in `test_driver.py` as you suggested in https://github.com/fish-shell/fish-shell/pull/11561#discussion_r2139880502.

Continuing to use a limit of 4 concurrent tests in CI seems to work, as evidenced by the latest run.

tests/pexpect_helper.py
danielrainer: The timeout changes are all removed.

krobelus:
freebsd tests are failing because semaphores don't seem to be context managers.

Probably we can upgrade to a stable version:

```diff
diff --git a/.cirrus.yml b/.cirrus.yml
index 812dce1233..763fff9566 100644
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -54,17 +54,11 @@
 
 freebsd_task:
     matrix:
-        # - name: FreeBSD 14
-        #   freebsd_instance:
-        #       image_family: freebsd-14-0-snap
-        - name: FreeBSD 13
+        - name: FreeBSD 14
           freebsd_instance:
-              image: freebsd-13-2-release-amd64
-        # - name: FreeBSD 12.3
-        #   freebsd_instance:
-        #       image: freebsd-12-3-release-amd64
+              image: freebsd-14-3-release-amd64-ufs
     tests_script:
-        - pkg install -y cmake-core devel/pcre2 devel/ninja misc/py-pexpect git-lite terminfo-db
+        - pkg install -y cmake-core devel/pcre2 devel/ninja lang/rust misc/py-pexpect git-lite
         # libclang.so is a required build dependency for rust-c++ ffi bridge
         - pkg install -y llvm
         # BSDs have the following behavior: root may open or access files even if
@@ -77,15 +71,6 @@
         - mkdir build && cd build
         - chown -R fish-user ..
         - sudo -u fish-user -s whoami
-        # FreeBSD's pkg currently has rust 1.66.0 while we need rust 1.70.0+. Use rustup to install
-        # the latest, but note that it only installs rust per-user.
-        - sudo -u fish-user -s fetch -qo - https://sh.rustup.rs > rustup.sh
-        - sudo -u fish-user -s sh ./rustup.sh -y --profile=minimal
-        # `sudo -s ...` does not invoke a login shell so we need a workaround to make sure the
-        # rustup environment is configured for subsequent `sudo -s ...` commands.
-        # For some reason, this doesn't do the job:
-        # - sudo -u fish-user sh -c 'echo source \$HOME/.cargo/env >> $HOME/.cshrc'
-        - sudo -u fish-user -s cmake -G Ninja -DCMAKE_BUILD_TYPE=Debug -DCTEST_PARALLEL_LEVEL=1 ..
         - sudo -u fish-user sh -c '. $HOME/.cargo/env; ninja -j 6 fish'
         - sudo -u fish-user sh -c '. $HOME/.cargo/env; ninja fish_run_tests'
     only_if: $CIRRUS_REPO_OWNER == 'fish-shell'
```

danielrainer:
I'll give that a try. But I think we still need the last line calling `cmake`, right? Just with `FISH_TEST_MAX_CONCURRENCY` set as an env var instead of the `-DCTEST_PARALLEL_LEVEL`.

.cirrus.yml
```diff
@@ -77,15 +71,7 @@ freebsd_task:
...
-        - sudo -u fish-user -s fetch -qo - https://sh.rustup.rs > rustup.sh
-        - sudo -u fish-user -s sh ./rustup.sh -y --profile=minimal
-        # `sudo -s ...` does not invoke a login shell so we need a workaround to make sure the
-        # rustup environment is configured for subsequent `sudo -s ...` commands.
-        # For some reason, this doesn't do the job:
-        # - sudo -u fish-user sh -c 'echo source \$HOME/.cargo/env >> $HOME/.cshrc'
-        - sudo -u fish-user -s cmake -G Ninja -DCMAKE_BUILD_TYPE=Debug -DCTEST_PARALLEL_LEVEL=1 ..
-        - sudo -u fish-user sh -c '. $HOME/.cargo/env; ninja -j 6 fish'
-        - sudo -u fish-user sh -c '. $HOME/.cargo/env; ninja fish_run_tests'
+        - sudo -u fish-user -s FISH_TEST_MAX_CONCURRENCY=1 cmake -G Ninja -DCMAKE_BUILD_TYPE=Debug ..
```
krobelus, line 74: LGTM. I don't know much about cmake so I don't know if an environment variable is the recommended approach,
but this is just temporary anyway.

Alpine and Jammy tests are failing, see `rg --hidden CTEST_PARALLEL_LEVEL`; should replace that with the new variable.

cmake/Tests.cmake
```diff
@@ -1,92 +1,7 @@
...
-)
-
-# CMake being CMake, you can't just add a DEPENDS argument to add_test to make it depend on any of
-# your binaries actually being built before `make fish_run_tests` is executed (requiring `make all` first),
-# and the only dependency a test can have is on another test. So we make building fish
-# prerequisites to our entire top-level `test` target.
-function(add_test_target NAME)
-  string(REPLACE "/" "-" NAME ${NAME})
-  add_custom_target("test_${NAME}" COMMAND ${CMAKE_CTEST_COMMAND} --output-on-failure -R "^${NAME}$$"
-    DEPENDS fish fish_indent fish_key_reader fish_test_helper USES_TERMINAL)
```
krobelus, line 56: actually, should we keep the `test_foo` targets (for tests/checks and tests/pexpects) if that's easy?
That could simplify bisection.
Once we get rid of CMake we can delete the whole file.
Not a blocker.

.cirrus.yml
danielrainer: Ah yes, I was so focused on the FreeBSD issue that I missed that.

cmake/Tests.cmake
danielrainer: I restored the code for creating the single-test targets, but modified it a bit to work without ctest. This also removes the `string(REPLACE "/" "-" NAME ${NAME})`. I don't know if this actually had any effect. Maybe check the new code to ensure it has the desired behavior.

PR #11597, opened 2025-06-21
Figure out what to clean up on SIGTERM

krobelus:
Commit 941701da3d8 (Restore some async-signal discipline to SIGTERM,
2025-06-15) removed some SIGTERM cleanup.  If think the function to disable
terminal protocols (like the kitty keyboard protocol) is async-signal-safe
though I suppose there is an avoidable panic when `Outputter::stdoutput()`
is already borrowed. Fix that.

We should figure out what the SIGTERM handler should do, and if it's not needed
remove things like `FLOG_SAFE!(term_protocols, "Disabling extended keys");`.

Also maybe backport the fix from 941701da3d8.

### DIFF
--- a/src/signal.rs
+++ b/src/signal.rs
@@ -4,6 +4,7 @@ use std::num::NonZeroI32;
 use crate::common::exit_without_destructors;
 use crate::event::{enqueue_signal, is_signal_observed};
 use crate::nix::getpid;
+use crate::panic::AT_EXIT;
 use crate::reader::{reader_handle_sigint, reader_sighup, safe_restore_term_mode};
 use crate::termsize::TermsizeContainer;
 use crate::topic_monitor::{topic_monitor_principal, Generation, GenerationsList, Topic};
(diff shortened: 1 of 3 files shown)


src/signal.rs
```diff
@@ -89,7 +90,9 @@ extern "C" fn fish_signal_handler(
         libc::SIGTERM => {
             // Handle sigterm. The only thing we do is restore the front process ID, then die.
             if !observed {
-                safe_restore_term_mode();
+                if let Some(at_exit) = AT_EXIT.get() {
```
ridiculousfish, line 93: The problem is this invokes `terminal_protocols_disable_ifn`. There's no documenation or effort to ensure that function, or the functions it calls transitively, are async-signal safe. And we certainly should not be invoking functions like `reader_current_data`  from signal handlers.

It's important that we keep the stuff that signal handlers do well-scoped and easy to follow. To only do what's necessary.

I'm working on a larger cleanup of the tty protocol stuff; FYI I'll be re-introducing this change, but in a way that restores the tty protocols in a manifestly async-signal-safe way.

Issue #11595, opened 2025-06-19
Assertion failure in Cirrus FreeBSD build


danielrainer:
https://cirrus-ci.com/task/4601990476464128
https://cirrus-ci.com/task/4780372816297984

The `assert!(len > 0)` in `src/history/file.rs:MmapRegion::new` fails in some cases in these builds. This is strange because the function is only used in two places, both of which do a `mmap` call using `len` and return on error before calling `MmapRegion::new`. `mmap` with a length of 0 should fail with `EINVAL` according to https://man.freebsd.org/cgi/man.cgi?query=mmap&apropos=0&sektion=2&manpath=FreeBSD+13.2-RELEASE&arch=default&format=html

What's going on there?

krobelus:
Strange. It didn't show up on a [sourcehut CI run with FreeBSD 13.5](https://builds.sr.ht/~krobelus/job/1516050)

krobelus:
The second link points to a failure that happened before f3ebc68d5d0 (Correct statvfs call to statfs, 2025-06-19)
This is pretty weird because that commit claims that we already never called mmap on FreeBSD..

krobelus:
Yeah this is probably from the locking changes (9b19db5c5f4 (Merge pull request #11492, 2025-06-19)).
Looks like there is a "needs_approval" state in https://cirrus-ci.com/github/fish-shell/fish-shell. I've email Cirrus support how to give this approval, let's see.
For now I've turned off `Require approval for builds from users without write permissions` in Cirrus if you have the appetite to debug.. though of course this setup is not ideal.. it definitely works when creating PRs against this repo. Probably changing `$CIRRUS_REPO_OWNER` will make it also work on forks (and avoid noisy notifications) but I don't remember if that actually worked..

Our FreeBSD cirrus image is also super old, way out of support. We probably still want to fix this (or at least add a workaround) but it would be good to have some automation to update these every few years.. Sourcehut doesn't even officially support FreeBSD 12.3 images anymore (which makes sense because FreeBSD upstream doesn't either) so I'm not sure we can use that to reproduce.
I guess it might be possible to run the Cirrus image via docker

krobelus:
My bad, I just needed to approve Cirrus workflows..
And there is an option to "Re-run with terminal access" in Cirrus

danielrainer:
[...]

PR #11598, opened 2025-06-21
DO NOT MERGE. This is for testing the FreeBSD Cirrus job

danielrainer:
https://github.com/fish-shell/fish-shell/issues/11595

### DIFF
--- a/src/tests/history.rs
+++ b/src/tests/history.rs
@@ -652,3 +652,30 @@ fn test_history_formats() {
     assert_eq!(test_history_imported_from_corrupted.get_history(), expected);
     test_history_imported_from_corrupted.clear();
 }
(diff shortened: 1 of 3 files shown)


src/tests/history.rs
```diff
@@ -652,3 +652,30 @@ fn test_history_formats() {
...
+            len,
+            PROT_READ,
+            MAP_PRIVATE,
+            file.as_raw_fd(),
+            0,
+        )
+    };
+
+    if ptr != MAP_FAILED {
+        panic!("mmap did not fail for len == 0");
```
krobelus, line 679: confirmed that this test passes on FreeBSD 14.3, so no need to report a bug

PR #11600, opened 2025-06-22
Fix docker-based tests, drop their usage of CMake

krobelus:
Make commands like

	sudo docker/docker_run_tests.sh --shell-after docker/focal.Dockerfile

work again, plus tangential changes

@danielrainer you might wanna check
- 482e6869 (docker_run_tests.sh: stop using cmake)
  (I just realized this breaks `docker/jammy-asan.Dockerfile`; we can probably add this
  back by passing the right options to check.sh?)
- e0babd76 (build_tools/update_translations.fish: remove forward reference)
- 6dda18c3 (build_tools/update_translations.fish: move po/template.po to /tmp)

### DIFF
--- a/build_tools/check.sh
+++ b/build_tools/check.sh
@@ -2,6 +2,24 @@
 
 set -ex
 
+lint=true
+if [ "$FISH_CHECK_LINT" = false ]; then
+    lint=false
+fi
+
+cargo_args=$FISH_CHECK_CARGO_ARGS
+target_triple=$FISH_CHECK_TARGET_TRIPLE
+if [ -n "$target_triple" ]; then
+    cargo_args="$cargo_args --target=$FISH_CHECK_TARGET_TRIPLE"
+fi
+
+cargo() {
+    subcmd=$1
+    shift
+    # shellcheck disable=2086
+    command cargo "$subcmd" $cargo_args "$@"
+}
+
 cleanup () {
     if [ -n "$template_file" ] && [ -e "$template_file" ]; then
         rm "$template_file"
@@ -10,17 +28,23 @@ cleanup () {
 
 trap cleanup EXIT INT TERM HUP
 
-RUSTFLAGS='-D warnings'; export RUSTFLAGS
-RUSTDOCFLAGS='-D warnings'; export RUSTDOCFLAGS
+if $lint; then
+    export RUSTFLAGS='-D warnings'
+    export RUSTDOCFLAGS='-D warnings'
+fi
 
 repo_root="$(dirname "$0")/.."
-build_dir="$repo_root/target/debug"
+build_dir="${CARGO_TARGET_DIR:-$repo_root/target}/${target_triple}/debug"
 
 template_file=$(mktemp)
 FISH_GETTEXT_EXTRACTION_FILE=$template_file cargo build --workspace --all-targets
-PATH="$build_dir:$PATH" "$repo_root/build_tools/style.fish" --all --check
-cargo clippy --workspace --all-targets
+if $lint; then
+    PATH="$build_dir:$PATH" "$repo_root/build_tools/style.fish" --all --check
+    cargo clippy --workspace --all-targets
+fi
 cargo test --no-default-features --workspace --all-targets
 cargo test --doc --workspace
-cargo doc --workspace
+if $lint; then
+    cargo doc --workspace
+fi
 FISH_GETTEXT_EXTRACTION_FILE=$template_file "$repo_root/tests/test_driver.py" "$build_dir"
--- a/build_tools/check.sh
+++ b/build_tools/check.sh
@@ -29,8 +29,8 @@ cleanup () {
 trap cleanup EXIT INT TERM HUP
 
 if $lint; then
-    export RUSTFLAGS='-D warnings'
-    export RUSTDOCFLAGS='-D warnings'
+    export RUSTFLAGS="${RUSTFLAGS} --deny=warnings"
+    export RUSTDOCFLAGS="${RUSTDOCFLAGS} --deny=warnings"
 fi
 
 repo_root="$(dirname "$0")/.."
--- a/build_tools/check.sh
+++ b/build_tools/check.sh
@@ -3,9 +3,9 @@
 set -ex
 
 cleanup () {
-  if [ -n "$template_file" ] && [ -e "$template_file" ]; then
-    rm "$template_file"
-  fi
+    if [ -n "$template_file" ] && [ -e "$template_file" ]; then
+        rm "$template_file"
+    fi
 }
 
 trap cleanup EXIT INT TERM HUP
--- a/build_tools/check.sh
+++ b/build_tools/check.sh
@@ -1,5 +1,6 @@
 #!/bin/sh
 
+{
 set -ex
 
 lint=true
@@ -48,3 +49,6 @@ if $lint; then
     cargo doc --workspace
 fi
 FISH_GETTEXT_EXTRACTION_FILE=$template_file "$repo_root/tests/test_driver.py" "$build_dir"
+
+exit
+}
--- a/build_tools/make_pkg.sh
+++ b/build_tools/make_pkg.sh
@@ -3,18 +3,18 @@
 # Script to produce an OS X installer .pkg and .app(.zip)
 
 usage() {
-  echo "Build macOS packages, optionally signing and notarizing them."
-  echo "Usage: $0 options"
-  echo "Options:"
-  echo "  -s                            Enables code signing"
-  echo "  -f <APP_KEY.p12>              Path to .p12 file for application signing"
-  echo "  -i <INSTALLER_KEY.p12>        Path to .p12 file for installer signing"
-  echo "  -p <PASSWORD>                 Password for the .p12 files (necessary to access the certificates)"
-  echo "  -e <entitlements file>        (Optional) Path to an entitlements XML file"
-  echo "  -n                            Enables notarization. This will fail if code signing is not also enabled."
-  echo "  -j <API_KEY.JSON>             Path to JSON file generated with \`rcodesign encode-app-store-connect-api-key\` (required for notarization)"
-  echo
-  exit 1
+    echo "Build macOS packages, optionally signing and notarizing them."
+    echo "Usage: $0 options"
+    echo "Options:"
+    echo "  -s                            Enables code signing"
+    echo "  -f <APP_KEY.p12>              Path to .p12 file for application signing"
+    echo "  -i <INSTALLER_KEY.p12>        Path to .p12 file for installer signing"
+    echo "  -p <PASSWORD>                 Password for the .p12 files (necessary to access the certificates)"
+    echo "  -e <entitlements file>        (Optional) Path to an entitlements XML file"
+    echo "  -n                            Enables notarization. This will fail if code signing is not also enabled."
+    echo "  -j <API_KEY.JSON>             Path to JSON file generated with \`rcodesign encode-app-store-connect-api-key\` (required for notarization)"
+    echo
+    exit 1
 }
 
 set -x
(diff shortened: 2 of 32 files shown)


danielrainer (commented):
Commit message of 8c67b07c1a231beae619595ec3dd55cd504a1568: Typo in first item of second list: `ad` -> `and`?

build_tools/check.sh
```diff
@@ -3,24 +3,35 @@
...
+fi
+
 trap cleanup EXIT INT TERM HUP
 
-RUSTFLAGS='-D warnings'; export RUSTFLAGS
-RUSTDOCFLAGS='-D warnings'; export RUSTDOCFLAGS
+if $lint; then
+    export RUSTFLAGS='-D warnings'
+    export RUSTDOCFLAGS='-D warnings'
+fi
```
danielrainer: I think it might be desirable to allow overriding these flags from outside the script, as in `RUSTFLAGS="-D warnings $RUSTFLAGS"`. This would e.g. allow using `RUSTFLAGS='-A warnings' build_tools/check.sh` to ignore warnings. (Maybe not the best example, but there might be other reasons to use this.)

Not particularly relevant for this PR, just an idea.

danielrainer:
Typo in commit message of 6dda18c3ee62ffe511cb33c9d7d71f023ae1b7da: `doe` -> `done`

danielrainer:
> I'm not sure if we still have a need for keeping around "po/template.po".
To add a new translation, you can run "build_tools/update_translations
po/xy.po". It could serve as a cache but that would only work if we integrated
it into a build system.

There is no need to keep it around. It is already getting deleted in the version of `build_tools/update_translations.fish` on master. I think not putting it into `po/` is an improvement.

build_tools/update_translations.fish
```diff
@@ -36,28 +36,19 @@
...
-end
-
 argparse --exclusive 'no-mo,only-mo,dry-run' no-mo only-mo dry-run use-existing-template= -- $argv
 or exit $status
 
-# Make sure that the template file is not included in $po_files.
-rm $template_file 2>/dev/null
+# Make sure that any old template file is not included in $po_files.
+# TODO Remove this before long.
+rm $build_tools/../po/template.po 2>/dev/null
```
danielrainer: Given that `template.po` is deleted at the end of this script (in the version on master), it should not be on anyone's file system (unless someone had a failed/interrupted run which did not perform the deletion. So it might be sufficient to just remove the `rm` here without replacement.

On a related note, we can also remove `po/template.po` from `.gitignore`.

danielrainer:
Typo in commit message of b15360ef98d7ddc376cbfba8e55fb4774afbe36a: `Tests to have the potential to clash` -> `Tests do have the potential to clash`?

build_tools/check.sh
danielrainer: This might be a better example why we might want to preserve the flags:
```sh
RUSTFLAGS=-Zsanitizer=address
RUSTDOCFLAGS=-Zsanitizer=address
```
(taken from `jammy-asan.Dockerfile`, `jammy-tsan.Dockerfile` is similar)

build_tools/make_pkg.sh
```diff
@@ -76,66 +76,66 @@ mkdir -p "$PKGDIR/build_x86_64" "$PKGDIR/build_arm64" "$PKGDIR/root" "$PKGDIR/in
...
-        --code-signature-flags runtime
-        --for-notarization
-    )
-    if [ -n "$ENTITLEMENTS_FILE" ]; then
-        ARGS+=(--entitlements-xml-file "$ENTITLEMENTS_FILE")
-    fi
-    for FILE in "$PKGDIR"/root/usr/local/bin/*; do
-        (set +x; rcodesign sign "${ARGS[@]}" "$FILE")
-    done
+        echo "Signing executables"
```
danielrainer: This file seems to be excessively indented now in several places.

docker/context/fish_run_tests.sh
```diff
@@ -1,26 +1,32 @@
...
 
 # Spawn a shell if FISH_RUN_SHELL_BEFORE_TESTS is set.
 if test -n "$FISH_RUN_SHELL_BEFORE_TESTS"
 then
-    bash -i
+    interactive_shell
 fi
 
-(set +e; ninja && ninja fish_run_tests)
+build_tools/check.sh $FISH_CHECK_ARGS
```
danielrainer: The changes introduce a few shellcheck lints:

```text
In docker/context/fish_run_tests.sh line 6:
cd /fish-source
^-------------^ SC2164 (warning): Use 'cd ... || exit' or 'cd ... || return' in case cd fails.

Did you mean: 
cd /fish-source || exit


In docker/context/fish_run_tests.sh line 9:
export CARGO_TARGET_DIR=$HOME/fish-build
                        ^---^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
export CARGO_TARGET_DIR="$HOME"/fish-build


In docker/context/fish_run_tests.sh line 24:
build_tools/check.sh $FISH_CHECK_ARGS
                     ^--------------^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
build_tools/check.sh "$FISH_CHECK_ARGS"

For more information:
  https://www.shellcheck.net/wiki/SC2164 -- Use 'cd ... || exit' or 'cd ... |...
  https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...
```
I don't know if anyone else cares about that. I cleaned up a while ago, so `shellcheck **.sh` output is clean otherwise. We don't have any automated checks for that, which makes it hard to avoid them when not using shellcheck in the editor.

If others are interested in using shellcheck, we could add it to our checks (e.g. in `check.sh`). I think it is packaged for pretty much all distros, so installing it in CI should be straightforward. The performance impact would be moderate. On my (not particularly powerful) machine `shellcheck **.sh` takes a bit over half a second.

danielrainer:
I like the changes to `update_translations.fish` and `check.sh` (and its wider adoption), the more consistent indentation, the improved help message for `test_driver.py`, the `canonicalize` functions in `build.rs`.

The other changes also seem nice, but I have not tested them much, so I don't know if they might break anything.

build_tools/check.sh
krobelus: The latest push passes everything via environment variables like `FISH_CHECK_LINT=true` etc.
Mainly because that made the docker use case easier.
For one, `/fish_run_tests`, already passes a bunch of environment variables to `check.sh`, for example `PCRE2_SYS_STATIC`.
I also made test driver read `FISH_TEST_MAX_CONCURRENCY`.
After a check failed in the container, this allows simply re-running `build_tools/check.sh` and still use the exact same environment.

We should maybe remove all of these environment variables. They seem convenient here. It's fine for the docker containers but perhaps too magic / implicit for other users of `check.sh`.
Maybe we should go the other way and use only explicit arguments.

To be used like (untested)

	docker/docker_run_tests.sh --shell-after --fish-check-lint=true docker/focal.Dockerfile

and it adds the full command check.sh command with all and any explicit arguments.


```diff
diff --git a/docker/context/fish_run_tests.sh b/docker/context/fish_run_tests.sh
index 37b48a8e94..6cb327d5ae 100755
--- a/docker/context/fish_run_tests.sh
+++ b/docker/context/fish_run_tests.sh
@@ -5,7 +5,14 @@
 # This script is copied into the root directory of our Docker tests.
 # It is the entry point for running Docker-based tests.
 
-echo build_tools/check.sh >>~/.bash_history
+escaped() {
+    printf " '%s'" "$(printf %s "$1" | sed "s,','\\''",g)"
+}
+echo "build_tools/check.sh$(
+    for arg; do
+        escaped "$arg"
+    done
+)" >>~/.bash_history
 
 cd /fish-source
 git config --global --add safe.directory /fish-source
@@ -26,7 +33,7 @@
 fi
 
 set +e
-build_tools/check.sh
+build_tools/check.sh "$@"
 RES=$?
 set -e
 
diff --git a/docker/docker_run_tests.sh b/docker/docker_run_tests.sh
index ff64592066..3e19dfa030 100755
--- a/docker/docker_run_tests.sh
+++ b/docker/docker_run_tests.sh
@@ -20,6 +20,8 @@
 # Get fish source directory.
 FISH_SRC_DIR=$(cd "$( dirname "$0" )"/.. >/dev/null && pwd)
 
+check_extra_args=
+
 # Parse args.
 while [ $# -gt 1 ]; do
     case "$1" in
@@ -29,11 +31,8 @@
         --shell-after)
             DOCKER_EXTRA_ARGS="$DOCKER_EXTRA_ARGS --env FISH_RUN_SHELL_AFTER_TESTS=1"
             ;;
-        --lint)
-            DOCKER_EXTRA_ARGS="$DOCKER_EXTRA_ARGS --env FISH_CHECK_LINT=true"
-            ;;
-        --no-lint)
-            DOCKER_EXTRA_ARGS="$DOCKER_EXTRA_ARGS --env FISH_CHECK_LINT=false"
+        *)
+            check_extra_args="$check_extra_args $1"
             ;;
         *)
             usage
@@ -42,6 +41,10 @@
     shift
 done
 
+if [ -n "$check_extra_args" ]; then
+    DOCKER_EXTRA_ARGS="$DOCKER_EXTRA_ARGS --env FISH_CHECK_EXTRA_ARGS=\$check_extra_args"
+fi
+
 DOCKERFILE="$1"
 test -n "$DOCKERFILE" || usage
 
@@ -54,7 +57,7 @@
 
 # Run tests in it, allowing them to fail without failing this script.
 # shellcheck disable=SC2086  # $DOCKER_EXTRA_ARGS should have globbing and splitting applied.
-docker run -it \
-    --mount type=bind,source="$FISH_SRC_DIR",target=/fish-source,readonly \
-    $DOCKER_EXTRA_ARGS \
-    "$IMG_TAGNAME"
+eval "docker run -it \\
+    --mount type=bind,source=\"\$FISH_SRC_DIR\",target=/fish-source,readonly \\
+    $DOCKER_EXTRA_ARGS \\
+    \"\$IMG_TAGNAME\""
diff --git a/docker/focal-32bit.Dockerfile b/docker/focal-32bit.Dockerfile
index 6c88cc4a94..5eca61b1ec 100644
--- a/docker/focal-32bit.Dockerfile
+++ b/docker/focal-32bit.Dockerfile
@@ -35,13 +35,10 @@
 
 COPY fish_run_tests.sh /
 
-ENV \
-    CFLAGS=-m32 \
-    PCRE2_SYS_STATIC=1 \
-    FISH_CHECK_TARGET_TRIPLE=i686-unknown-linux-gnu
-
-ENV FISH_CHECK_LINT=false
-
 CMD . ~/.cargo/env \
-    && rustup target add ${FISH_CHECK_TARGET_TRIPLE} \
-    && /fish_run_tests.sh
+    && rustup target add i686-unknown-linux-gnu \
+    && /fish_run_tests.sh \
+        --cflags=-m32 \
+        --pcre2-sys-static=1 \
+        --target=i686-unknown-linux-gnu \
+        --fish-check-lint=false
```

docker/context/fish_run_tests.sh
krobelus: Yeah adding shellcheck sounds like a good idea (bash-language-server also shows shellcheck diagnostics).

build_tools/check.sh
```diff
@@ -1,26 +1,54 @@
...
+    fi
 }
 
 trap cleanup EXIT INT TERM HUP
 
-RUSTFLAGS='-D warnings'; export RUSTFLAGS
-RUSTDOCFLAGS='-D warnings'; export RUSTDOCFLAGS
+if $lint; then
+    export RUSTFLAGS="${RUSTFLAGS} --deny=warnings"
+    export RUSTDOCFLAGS="${RUSTDOCFLAGS} --deny=warnings"
```
danielrainer, line 34: Is it a deliberate choice to put the inherited `RUSTFLAGS` before the additions? If there is no good reason to prefer this order I think the inherited `RUSTFLAGS` should come after the flags we add here, since that allows overriding them externally. For example, if I want to see all lints without terminating once the first command encounters a warning, I could run `RUSTFLAGS='-W warnings' build_tools/check.sh`. But this does not work when the `--deny=warnings` comes later in `RUSTFLAGS`, since that overrides the option I'm trying to pass.

build_tools/check.sh
danielrainer: Yes, for usage in CI, especially when a variable can be defined once for multiple usages, can be nice. I agree that it is not a great interface for interactive use. I'm not sure if the best solution is completely abandoning the env vars in favor of arguments or allowing arguments to override env vars (if there are any). I think if it works its sufficient for now, and if some use case is annoying the code can be adjusted accordingly later.

danielrainer:
At the moment `test_driver.py` fails the formatting check.

docker/jammy-tsan.Dockerfile
```diff
@@ -37,9 +35,13 @@ RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs > /tmp/rustup.sh \
 COPY fish_run_tests.sh /
 
 ENV \
-    RUSTFLAGS=-Zsanitizer=thread \
+    RUSTFLAGS='-Zsanitizer=thread ' \
```
danielrainer, line 38: It should be sufficient to add spacing when combining this with other flags in `build_tools/checks.sh`, right?

docker/context/fish_run_tests.sh
danielrainer: https://github.com/fish-shell/fish-shell/pull/11608

build_tools/check.sh
krobelus: Yeah I had it the other way round first, that seems more logical.
But I'd be mildly surprised if it's actually used in practice but we'll see

build_tools/check.sh
krobelus: We could still support env vars, but should probably not set them by default (and if we must, use the smallest possible scope).
Can be fixed later I guess

PR #11603, opened 2025-06-23
Remove obsolete translation entries

krobelus:
I doubt these are very helpful; most of them won't be used again.  We can
still find them in history with "git log -Gpart.of.msgid".


krobelus:
merged as d2f7d238f3d (Remove obsolete translation entries, 2025-06-23)

PR #11608, opened 2025-06-24
Add ShellCheck to `check.sh`

danielrainer:
ShellCheck can detect some common issues in shell scripts (not in fish scripts though). Adding it to `check.sh` means such issues are noticed and addressed. Disabling specific lints for a line can be done with a comment as demonstrated in the changes in `check.sh`.

Some context: https://github.com/fish-shell/fish-shell/pull/11600#discussion_r2160512475

## TODOs:
- [ ] Do we want a more noticeable warning when ShellCheck is skipped?
- [ ] Should we document ShellCheck as a recommended dev dependency somewhere?


danielrainer:
This won't run in CI until `check.sh` is used there, so technically installing the package in CI is useless for now. But once we switch to that ShellCheck should just work there. At the moment I only added it to the GitHub CI ubuntu job, because there results should be platform-independent. Not sure if we want that check elsewhere.

.github/workflows/main.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     - uses: dtolnay/rust-toolchain@1.70
     - name: Install deps
       run: |
-        sudo apt install gettext libpcre2-dev python3-pexpect python3-sphinx tmux
```
krobelus: I get these lints, should address them before merging.

```
$ shellcheck --version
ShellCheck - shell script analysis tool
version: 0.10.0
license: GNU General Public License, version 3
website: https://www.shellcheck.net
$ git ls-files '*.sh' | xargs shellcheck
In build_tools/check.sh line 25:
    if [ -n "$template_file" ] && [ -e "$template_file" ]; then
    ^-- SC2317 (info): Command appears to be unreachable. Check usage (or ignore if invoked indirectly).
       ^----------------------^ SC2317 (info): Command appears to be unreachable. Check usage (or ignore if invoked indirectly).
                                  ^---------------------^ SC2317 (info): Command appears to be unreachable. Check usage (or ignore if invoked indirectly).


In build_tools/check.sh line 26:
        rm "$template_file"
        ^-----------------^ SC2317 (info): Command appears to be unreachable. Check usage (or ignore if invoked indirectly).


In build_tools/git_version_gen.sh line 37:
	echo $VN
             ^-^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
	echo "$VN"

For more information:
  https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...
  https://www.shellcheck.net/wiki/SC2317 -- Command appears to be unreachable...
```

.github/workflows/main.yml
```diff
@@ -19,7 +19,7 @@ jobs:
...
+        sudo apt install gettext libpcre2-dev python3-pexpect python3-sphinx shellcheck tmux
```
krobelus: > This won't run in CI until check.sh is used there, so technically
 > installing the package in CI is useless for now. But once we switch
 > to that ShellCheck should just work there.

I don't expect that we want to make a single CI job that runs all of `build_tools/check.sh`,
because we want to have the ability to separate jobs,
so one can see faster which check failed (when looking at the failure notification email or the github UI).

There's various possibilities. The github (and other CI) workflow files could
- duplicate the check commands
- be autogenerated (which probably implies that `build_tools/check.sh` be autogenerated as well).
- call `build_tools/check.sh --only=clippy` etc.

The current approach is "duplicate",
so it wouldn't really hurt doing that for shellcheck for now also.

 > At the moment I only
 > added it to the GitHub CI ubuntu job, because there results should
 > be platform-independent. Not sure if we want that check elsewhere.

yeah that should be good enough.
For the non-Rust tools (black/shellcheck/sphinx), we fix warnings as we see them,
the goal is that any reasonably modern version of those tools is warning-free

build_tools/check.sh
```diff
@@ -24,3 +24,9 @@ cargo test --no-default-features --workspace --all-targets
 cargo test --doc --workspace
 cargo doc --workspace
 FISH_GETTEXT_EXTRACTION_FILE=$template_file "$repo_root/tests/test_driver.py" "$build_dir"
+if command -v shellcheck; then
+  # shellcheck disable=SC2046
+  shellcheck $(find . -name '*.sh')
```
krobelus: 1. it's weird  that this check is run at the very end, instead of next to
   the other cheap checks such as formatting
2. now that the `--no-lint` option has been added, this should be disabled if `! $lint`.

3. we should perhaps not check all sh files, because that might
include files that are gitignored, or files that are only present in local commits
(and not intended to be merged into master anytime soon). Elsewhere we use

```
set fish_files $repo_root/{benchmarks,build_tools,etc,share}/**.fish
set python_files {doc_src,share,tests}/**.py
```

Listing some top-level directories sounds like a reasonable compromise?
Alternatively, maybe add an easy way to disable the check for some files
without modifying check.sh. Not sure about the interface. It should be Git-agnostic.

build_tools/check.sh
```diff
@@ -24,3 +24,9 @@ cargo test --no-default-features --workspace --all-targets
...
+else
+  echo 'ShellCheck is not available. Skipping related checks.'
```
krobelus: This behavior is consistent with what we do with formatters in `style.fish` (`black` and `rustfmt`),
which seems appropriate..

However, `clippy` is also a linters, and we fail if it's not installed.

Unless there is a reason fro this distinction, we should fix this inconsistency.

Maybe we should rather make `check.sh` fail if any linter is not
installed, so one doesn't have to wait for CI to find a formatting
error.
We'd still have `--no-lint`, and if anyone wants it, we can add
an option to not treat missing tools as failure. Something like
`--ignore-missing-linters` or ideally shorter.

build_tools/check.sh
danielrainer: > it's weird that this check is run at the very end, instead of next to
the other cheap checks such as formatting

I don't really care about the order, running it e.g. after the formatting checks would be fine by me. Then we can take advantage of the existing `if` statement to skip it when lints are disabled.

> we should perhaps not check all sh files

Adding `# shellcheck disable=all` after the shebang could address that, at least for files where manual modifications make sense.

I don't really know why someone might want to disable ShellCheck on some files. Maybe for files which are included in the repo but which are just copies from an outside source it might make sense to disable ShellCheck, but AFAICT we don't have any such file. If we ever do that, we can consider changing the ShellCheck invokation then. Until that happens, the current approach automatically checks all files, including newly added ones, without anyone having to think about explicitly including a new file.

.github/workflows/main.yml
danielrainer: Ah yes, I used an old ShellCheck version which did not show these. Simple fix.

.github/workflows/main.yml
danielrainer: > The current approach is "duplicate", so it wouldn't really hurt doing that for shellcheck for now also.

Sure, that's reasonable for now. Autogeneration seems like it would be quite a bit of work to implement. Adding flags to `build_tools/check.sh` is probably the more realistic future approach, although it does not solve the problem of having to remember changing all users when a check is added/removed.

build_tools/check.sh
danielrainer: I'll make `check.sh` fail if `$lint` is true but `shellcheck` is unavailable. Changing behavior on missing formatters is probably better done in a separate PR.

> if anyone wants it, we can add
an option to not treat missing tools as failure.

Sure. I think this should also be addressed in its own PR, if at all.

build_tools/check.sh
danielrainer: OTOH, the current, naive approach of looking for files with the `.sh` extension misses some scripts, i.e. the ones in `build_tools/osx_package_scripts` and `debian`. (And maybe more that I've missed?)

We could use ripgrep for detecting files with a shebang indicating a non-fish shell script, e.g.:
```fish
rg --stop-on-nonmatch --files-with-matches '^#!.*[^i]sh'
``` 
Using this directly in `check.sh` would require having ripgrep installed. Alternatively, we could have a file which stores all paths we want checked by ShellCheck. This file could be autogenerated by the command above (or something else, depending on our needs). Then we could regenerate it in CI to fail on changes, which would automatically notify us when anything changes, and lets us decide whether to add the path so it gets shellchecked, or modify the autogeneration to exclude it. If this only runs in CI then checks can be added or removed locally be modifying the file containing the paths to check.

We don't have to use ripgrep of course, I just used it because it makes it simple to get what I wanted in a single command, including recursive directory traversal while adhering to gitignore files.

krobelus: Disregard my "files that are only present in local commits" - that's not relevant, because there's no need to run `check.sh` on those commits.

So gitignored files are my main concern.  I think it's reasonable to write
gitignored scripts to reproduce various issues etc.  That's mostly one-off
scripts; important scripts I'd put into version control (at least in local-only
commits), so that's not relevant.
I suppose I could version-control all my scripts or rename them. But I don't expect the tradeoff is worth it.

rg seems promising but it doesn't seem to respect gitignore in non-colocated jj workspace yet.
Once we fix that, I guess we could use it in `check.sh`.
it probably wouldn't be worth it for only shell scripts, but I'd imagine we want to use it for python and fish files too.

I'm not 100% sure if I prefer that over the existing approach..

Instead of looking for the shebang, I'd probably add explicit exceptions for the files without the *.sh extension.. not sure about this either.

I have a vague memory of learning that "globs in the build system" are a bad idea in general,
i.e. the build system should have an explicit list of all source files (possibly autogenerated),
which allows it to implement caching efficiently.

> Changing behavior on missing formatters is probably better done in a separate PR.

yeah, this inconsistency is pre-existing.
But the one between globbing everything and what `style.fish` does today is not.

danielrainer: > gitignore in non-colocated jj workspace

I'm not familiar with the way jj works. Does this mean rg being called from a directory outside of the git repo containing a .gitignore? Would be great if you elaborated on how this use case would work.

> Instead of looking for the shebang, I'd probably add explicit exceptions for the files without the *.sh extension..

The issue addressed by looking for a shebang is that files without a .sh extension are not automatically ignored. We already have such files, and there might plausibly be more added in the future. If this happens, it's easy to forget/not be aware of the ShellCheck logic.

> the build system should have an explicit list of all source files (possibly autogenerated)

Yes, that's what I was trying to get at. Autogenerate the set of files using rg and store the results. Add exceptions to ignore false positives in a way which makes rg ignore them. Possibly adapt the rg invocation to eliminate false negatives. Then, the `check.sh` script (and CI runs) run ShellCheck on the files on our list. As an additional step, run rg and flag mismatches between its output and the stored list, so they can be addressed manually (add new files to the list or ignore them, delete files which were removed from the repo from our list, handle renames).
If we want caching that would of course still have to be implemented. This approach also makes it simple to not run rg in `check.sh` if it's unavailable, and just use the file storing all inputs to ShellCheck. In CI, we can ensure that rg is available, and detect missed updates before merging.

I guess we could also adopt a similar approach for `style.fish`. For fish (and to a lesser extent Python) scripts this might require quite frequent changes as tests are added/removed/renamed. For formatting Rust files I don't think it makes sense to modify the current implementation, since we don't glob there and cargo keeps track of things.

krobelus: > gitignore in non-colocated jj workspace
> 
> I'm not familiar with the way jj works. Does this mean rg being called from a directory outside of the git repo containing a .gitignore? Would be great if you elaborated on how this use case would work.

Currently most jj users probably use git+jj colocated repositories, i.e. `jj git init --colocate`.

jj's equivalent of Git worktrees are "workspaces" (i.e. `jj workspace add ../fish-shell2`).
AFAICT they don't support Git-colocation yet.
I don't know yet how useful they are, but I'd like to avoid adding roadblocks.

Such a secondary workspace will have a `.gitignore` but no `.git`.
Today ripgrep/ignore does not respect .gitignore in this case,
but looks like [that will change](https://github.com/BurntSushi/ripgrep/pull/2842).
(it won't support $GIT_DIR/info/exclude (yet?) but I guess I can work around that with a global gitignore)

danielrainer: The `--no-require-git` or ripgrep suggested in https://github.com/BurntSushi/ripgrep/pull/2842#issuecomment-2204514682 might work for our purposes. It could be an issue if people have a `.gitignore` outside of the fish repo.

Otherwise, we would have to wait for rg to release jj support (which might plausibly happen soon), and then also for this release to be available in distros, which can take quite a while, at least if we care about non-rolling-release distros.

krobelus:
> The `--no-require-git` or ripgrep suggested in https://github.com/BurntSushi/ripgrep/pull/2842#issuecomment-2204514682 might work for our purposes.

yes, definitely

> It could be an issue if people have a `.gitignore` outside of the fish repo.

Yeah, in the scenario where ~/.gitignore exists and ~/fish-shell/ is a
workspace without .git, and rg doesn't recognize .jj yet,
honoring ~/.gitignore is wrong but often harmless.
I'd also expect that "rg" has an option to ignore ".git" directories
(so we want something like "rg --no-require-git --dont-detect-vcs-repositories" in the script)

It's also fine to require temporary workarounds for such rare edge cases,
given that the workaround will be obsolete soon.

Also, all tools that use Git to detect the project root will need to
do something like
"jj workspace root 2>/dev/null || git rev-parse --show-toplevel"

danielrainer:
> I'd also expect that "rg" has an option to ignore ".git" directories

`--no-ignore-exclude` might be what we want. The naming is a bit confusing, but the docs say:

> Don't respect filter rules from files that are manually configured for the repository.  For example, this includes git's .git/info/exclude.

and testing confirms that rules in `.git/info/exclude` are not considered when this flag is provided.

> all tools that use Git to detect the project root

I think it's not limited to root detection, pretty much all tools using git will break in a non-colocated jj workspace, right?

krobelus:
> > all tools that use Git to detect the project root
> 
> I think it's not limited to root detection, pretty much all tools using git will break in a non-colocated jj workspace, right?

Yeah; for now I'm only using colocated repos (except for some deployment scenarios),
and do everything in a single workspace.
[In future](https://github.com/jj-vcs/jj/pull/4644) colocation will also work for secondary workspaces, so each workspace will be a valid Git worktree.

But it's still nice to have the option to swap out the VCS without breaking unrelated things (like detecting project root).

danielrainer:
I implemented a list tracking which files should be shellchecked, plus a script for updating the list and checking if it is outdated, plus a way to locally ignore files. It works well locally, but for some reason `rg` fails to run correctly in CI. I don't know what's going on there. See https://github.com/danielrainer/fish-shell/actions/runs/16997092097/job/48190296278 for some tests.

.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,29 @@
+name: ShellCheck
```
krobelus, line 1: I thought that this had become surprisingly complicated and that there must be a simpler solution.
But looking at it now, I guess it's fine modulo my comments, so if you're okay with that, could you rebase?

.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,29 @@
...
+
+on: [push, pull_request]
+
+permissions:
+  contents: read
+
+jobs:
+  file-list-up-to-date:
```
krobelus, line 9: can we add this to the "lint" workflow in .github/workflows/lint.yml instead?

build_tools/check.sh
```diff
@@ -41,6 +42,23 @@ template_file=$(mktemp)
 FISH_GETTEXT_EXTRACTION_FILE=$template_file cargo build --workspace --all-targets --features=gettext-extract
 if $lint; then
     PATH="$build_dir:$PATH" "$repo_root/build_tools/style.fish" --all --check
+
+    shellcheck_dir="$repo_root/build_tools/shellcheck"
+    if command -v rg; then
+        echo 'Checking if list of files to shellcheck is up to date.'
+        echo 'If this fails, you can update the list automatically using'
+        echo "$shellcheck_dir/check_for_file_updates.sh --update"
```
krobelus, line 50: I guess we could also tweak the diff output a little bit and then suggest

	check_for_file_updates.sh | git apply

instead (or "patch -p1" but "git apply" should be more accessible).

build_tools/check.sh
```diff
@@ -41,6 +42,23 @@ template_file=$(mktemp)
...
+        "$shellcheck_dir/check_for_file_updates.sh"
+    else
+        echo 'Cannot update list of files to shellcheck. Requires ripgrep.'
+    fi
+    if command -v shellcheck; then
+        "$shellcheck_dir/check.sh"
+    else
+        echo 'Error: ShellCheck is not available.'
+        exit 1
```
krobelus, line 59: > Do we want a more noticeable warning when ShellCheck is skipped?
 > Should we document ShellCheck as a recommended dev dependency somewhere?

failing `build_tools/check.sh` like we do here seems good enough

build_tools/shellcheck/check_for_file_updates.sh
```diff
@@ -0,0 +1,95 @@
...
+# If the `--update` flag is provided, the list of files to check is updated instead of checked.
+# Then, a non-zero exit code indicates some failure in the script.
+#
+# If you want to exclude paths from these checks, create a file called `excluded.txt` in this
+# directory. For each file you want to exclude, put its relative path starting from the repo root
+# into a line of the file.
+# Make sure that the file does not have any other content or empty lines.
+# Also ensure that `excluded.txt` is not empty.
+# The file is intended for locally ignoring scripts not committed in the repo, which is why
+# `excluded.txt` is already gitignored.
```
krobelus, line 14: I don't think we should invent our own excluded file, `.git/info/exclude` already works, see my other comment.
No need to try to prevent power users from shooting themselves in the foot (which is super unlikely anyway here).
So I think we can get rid of the excluded logic?

build_tools/shellcheck/check_for_file_updates.sh
```diff
@@ -0,0 +1,95 @@
...
+# If a file which is committed into the repo should not be checked, disable the checks via
+# annotations in that file, either for each lint individually or by adding
+# # shellcheck disable=all
+# at the start of the file (but after the sheband).
```
krobelus, line 18: sheband -> shebang

build_tools/shellcheck/check_for_file_updates.sh
```diff
@@ -0,0 +1,95 @@
...
+
+set -e
+
+cleanup () {
+    # shellcheck disable=2317
+    if [ -n "$tmp_file" ] && [ -e "$tmp_file" ]; then
+        rm "$tmp_file"
```
krobelus, line 25: due to the traps below, this could run twice, so let's make sure it's actually idempotent
(else we might remove a tmp file created by another process).
maybe define a `cleanup_done=false` variable.

build_tools/shellcheck/check_for_file_updates.sh
```diff
@@ -0,0 +1,95 @@
...
+    filter_excluded='cat'
+fi
+
+# First, identify shell scripts by the presence of a shebang on the first line.
+# Ripgrep respects .gitignore files, so files ignored there are automatically ignored here as well.
+# By default, .gitignore files are only ignored within a git repo. In some situations, like in
+# non-colocated jj repos, ripgrep cannot identify the repo root, resulting in .gitignore files not
+# being taken into account.
+# The `--no-require-git` flag prevents this, at the cost of also respecting .gitignore files outside
+# of the repo on in ancestor directories.
```
krobelus, line 64: typo: on -> or

 > in ancestor directories

hmm, can't we prevent this by telling rg not to look at .gitignore files above $repo_root ?
I can't immediately see how but it's not a big deal.

We could also drop `--no-require-git` and deal with this when it becomes an actual problem (which is not anytime soon, maybe never).

build_tools/shellcheck/check_for_file_updates.sh
```diff
@@ -0,0 +1,95 @@
...
+#
+# By default, ripgrep also considers rules in .git/info/exclude, which only exist locally.
+# This script should behave the same way everywhere, so we do not want to take local configuration
+# into account, which is why we disable this feature with `--no-ignore-exclude`.
```
krobelus, line 68: I think this is needlessly aggressive.
I don't think we should lint scripts that have been explicitly ignored by the user.
It's true that things like clippy do but I often write temporary shell scripts and linters would get in the way,
and even worse, the check_for_file_updates script would try to add them.

It's weird to respect some gitignore rules but not others.
Let's remove the --no-ignore-exclude flag? The important thing is that anything that will be tracked in Git will be checked, which we'll still have.

build_tools/shellcheck/check_for_file_updates.sh
```diff
@@ -0,0 +1,95 @@
...
+#
+# The output of the first command will contain the paths of all files containing a shell shebang.
+#
+# The next command in the pipeline filters out any paths which are explicitly ignored in the
+# corresponding file (build_tools/shellcheck/excluded.txt).
+# `--fixed-strings` is used to prevent paths from being interpreted as regular expressions.
+#
+# Finally, the output is sorted, because ripgrep's output is unsorted.
+# Setting the locale is important, because the behavior of sort is locale-dependent.
+rg --no-require-git --no-ignore-exclude --stop-on-nonmatch --files-with-matches '^#!.*[^i]sh' \
```
krobelus, line 87: maybe better to use negative lookahead `(?!fish)sh` (maybe need `rg --pcre2`) but no big deal.

danielrainer:
I implemented an updated version of this in #12661. I agree that the inclusion/exclusion logic in this PR is needlessly complicated. This PR started out simpler, and the new one goes back to the simple approach of computing the files to be checked on the fly. Please have a look whether that satisfies your needs. We could also go with an updated version of this PR with at least the exclusion logic removed, but I prefer adding new tools as Rust code now that we have xtasks, rather than introducing more shell scripts.

build_tools/shellcheck/check_for_file_updates.sh
danielrainer: I quickly benchmarked these two commands:
```sh
rg --stop-on-nonmatch --files-with-matches '^#!.*[^i]sh'
rg --pcre2 --stop-on-nonmatch --files-with-matches '^#!.*(?!fi)sh'
```
The former takes about 15 ms, the latter 32 ms on my system, so I think we should use the former approach unless you are concerned about false positives.

krobelus: > `'^#!.*(?!fi)sh'`

my bad, this is actually trivially true. Needs negative lookbehind i..e `(?<!fi)sh`

krobelus:
yeah a Rust implementation seems better, as does on-the-fly computation of files to check

build_tools/shellcheck/check_for_file_updates.sh
danielrainer: Doesn't make much of a difference for performance. For the new Rust implementation I used the `pcre2` crate we already use, although `ignore` already introduces some `regex` dependencies, so I think we could use that as well. Performance-wise, there is not much of a difference between the two with `pcre2`, at least for debug builds of the whole xtask, which is what I think matters in practice, so we could use whichever you prefer.

PR #11615, opened 2025-06-25
Add colorful disassembler output completion for objdump

adamanteye:
As described in objdump(1), --disassembler-color can be applied to enable or disable the use of syntax highlighting in disassembly output.

The options are:

--disassembler-color=off
--disassembler-color=terminal
--disassembler-color=on|color|colour
--disassembler-color=extended|extended-color|extended-colour


krobelus:
Thanks for updating the translations. We have some older translations that are not great, and probably more confusing than helpful; modern technology allows us to fix them.

PR #11622, opened 2025-06-26
Add 'ollama stop' completions.

pbratkowski:
## Description

This adds support for the `ollama stop [model]` command, which has been added to Ollama in September 2024.

- Added the '__fish_ollama_ps' function to list running models.
- Added the 'stop' subcommand to ollama completions.
- Added running models as arguments to 'stop'.

## TODOs:

Not applicable.

### DIFF
--- a/share/completions/ollama.fish
+++ b/share/completions/ollama.fish
@@ -2,11 +2,16 @@ function __fish_ollama_list
     ollama list 2>/dev/null | tail -n +2 | string replace --regex "\s.*" ""
 end
 
+function __fish_ollama_ps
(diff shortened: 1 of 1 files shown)


pbratkowski:
Hmmm... Is there anything I should be doing in regards to translations?

krobelus:
> Hmmm... Is there anything I should be doing in regards to translations?

you can but no need; this is mainly a reminder for the maintainer to run `build_tools/update_translations.sh` before merging.
In future we should probably disable this check for completion-only PRs, since I pretty much always want to run the same checks locally before merging rather than looking at github

krobelus:
merged as d62fb9cc748b67a9f14d462b3491959580a79026, thanks

PR #11623, opened 2025-06-27
Avoid running `sphinx-build` if possible

danielrainer:
Despite the caching in `sphinx-build`, it takes several seconds to run
`sphinx-build` when no rebuilding is necessary, which slows down build times
significantly.

Add custom logic to `build.rs` to avoid calling `sphinx-build` if deemed
unnecessary based on the mtime of the source files.
This is done by writing the most recent timestamp of the source files into a
dedicated file, and only calling `sphinx-build` (and updating the timestamp)
when a cached timestamp is older than the most recent source file mtime.

For reference, this saves about 4 seconds when running `cargo b; touch build.rs; cargo b` on my system, which is a reduction of almost 40% in runtime for the second `cargo b`.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -59,7 +59,14 @@ fn main() {
         let _ = std::fs::create_dir_all(sec1dir.to_str().unwrap());
     }
 
-    rsconf::rebuild_if_paths_changed(&["src", "printf", "Cargo.toml", "Cargo.lock", "build.rs"]);
+    rsconf::rebuild_if_paths_changed(&[
+        "build.rs",
+        "Cargo.lock",
+        "Cargo.toml",
+        "gettext-extraction",
+        "printf",
+        "src",
+    ]);
 
     // These are necessary if built with embedded functions,
     // but only in release builds (because rust-embed in debug builds reads from the filesystem).
--- a/build.rs
+++ b/build.rs
@@ -14,6 +14,10 @@ fn canonicalize_str<P: AsRef<Path>>(path: P) -> String {
 
 const MANIFEST_DIR: &str = env!("CARGO_MANIFEST_DIR");
 
+#[cfg(feature = "embed-data")]
+#[cfg(not(clippy))]
+const SPHINX_DOC_SOURCES: [&str; 3] = ["CHANGELOG.rst", "CONTRIBUTING.rst", "doc_src"];
+
 fn main() {
     setup_paths();
 
@@ -72,7 +76,11 @@ fn main() {
     // but only in release builds (because rust-embed in debug builds reads from the filesystem).
     #[cfg(feature = "embed-data")]
     #[cfg(not(debug_assertions))]
-    rsconf::rebuild_if_paths_changed(&["doc_src", "share"]);
+    rsconf::rebuild_if_paths_changed(&["share"]);
+
+    #[cfg(feature = "embed-data")]
+    #[cfg(not(clippy))]
+    rsconf::rebuild_if_paths_changed(&SPHINX_DOC_SOURCES);
 
     rsconf::rebuild_if_env_changed("FISH_GETTEXT_EXTRACTION_FILE");
 
--- a/build.rs
+++ b/build.rs
@@ -407,6 +407,106 @@ fn get_version(src_dir: &Path) -> String {
 }
 
 #[cfg(feature = "embed-data")]
+#[cfg(not(clippy))]
+/// Given a path, returns an iterator containing this path, and possibly others depending on what
+/// the input path refers to.
+/// If the path points to a file, only this file will be included.
+/// If it is a symlink, the path of the link will be included, as well as the target,
+/// provided the target is a file. Panics on non-file targets, including non-existing targets.
+/// If it is a directory, the function recurses on each entry and combines the results into the
+/// returned iterator.
+fn read_dir_all<P: AsRef<Path>>(path: P) -> impl Iterator<Item = PathBuf> {
+    fn walk(path: PathBuf) -> Box<dyn Iterator<Item = PathBuf>> {
+        if path.is_file() {
+            return Box::new([path].into_iter());
+        }
+        if path.is_symlink() {
+            let target = path.read_link().unwrap();
+            if target.is_file() {
+                return Box::new([path, target].into_iter());
+            }
+            panic!("Symlink {path:?} does not point to a file. Such symlinks are unsupported.");
+        }
+        if path.is_dir() {
+            return Box::new(
+                [path.clone()].into_iter().chain(
+                    std::fs::read_dir(path)
+                        .unwrap()
+                        .flat_map(|p| walk(p.unwrap().path())),
+                ),
+            );
+        }
+        panic!("Unexpected type for path {path:?}");
+    }
+
+    walk(path.as_ref().to_owned())
+}
+
+#[cfg(feature = "embed-data")]
+#[cfg(not(clippy))]
+/// Helper function for avoiding unnecessary rebuilds.
+/// Specify all files/directories which should trigger a rebuild if modified via `source_paths`.
+/// `cache_file` is used to store the time of the latest build.
+/// If this file does not exist, it will be created, if possible.
+/// The modification time of all files in `source_paths` are checked.
+/// For symlinks, both the link and the target are considered, but only links to files are
+/// supported.
+/// If the most recent mtime found this way is not more recent than the one stored in `cache_file`,
+/// then the build directory is considered up-to-date and `true` is returned.
+/// Otherwise, the timestamp in `cache_file` is updated and false is returned.
+/// The mtime is stored as milliseconds since the Unix epoch.
+/// There are some edge cases around the time, which could theoretically result in unnecessary
+/// builds or builds not being updated when they should.
+/// <https://doc.rust-lang.org/std/time/struct.SystemTime.html>
+/// Deleting the `cache_file` before invoking this function prevents the latter case (but makes
+/// this function useless).
+fn is_build_dir_up_to_date<'a, P: AsRef<Path> + 'a, I: IntoIterator<Item = &'a P>>(
+    source_paths: I,
+    cache_file: &Path,
+) -> bool {
+    use std::{
+        io::{Read, Seek, SeekFrom, Write},
+        str::FromStr,
+        time::UNIX_EPOCH,
+    };
+    let source_timestamp = source_paths
+        .into_iter()
+        .flat_map(read_dir_all)
+        .map(|p| p.metadata().unwrap().modified().unwrap())
+        .max()
+        .unwrap()
+        .duration_since(UNIX_EPOCH)
+        .unwrap()
+        .as_millis();
+    let cache_file_dir = cache_file.parent().unwrap();
+    std::fs::create_dir_all(cache_file_dir).unwrap();
+    let mut timestamp_file = std::fs::OpenOptions::new()
+        .read(true)
+        .write(true)
+        .create(true)
+        .truncate(false)
+        .open(cache_file)
+        .unwrap();
+    let mut timestamp_file_content = vec![];
+    timestamp_file
+        .read_to_end(&mut timestamp_file_content)
+        .unwrap();
+    if !timestamp_file_content.is_empty() {
+        let timestamp_file_str = std::str::from_utf8(&timestamp_file_content).unwrap().trim();
+        let build_timestamp = u128::from_str(timestamp_file_str).unwrap();
+        if source_timestamp <= build_timestamp {
+            return true;
+        }
+        timestamp_file.seek(SeekFrom::Start(0)).unwrap();
+        timestamp_file.set_len(0).unwrap();
+    }
+    timestamp_file
+        .write_all(format!("{source_timestamp}\n").as_bytes())
+        .unwrap();
+    false
+}
+
+#[cfg(feature = "embed-data")]
 // disable clippy because otherwise it would panic without sphinx
 #[cfg(not(clippy))]
 fn build_man(build_dir: &Path) {
@@ -415,6 +515,16 @@ fn build_man(build_dir: &Path) {
     let sec1dir = mandir.join("man1");
     let docsrc_path = canonicalize(MANIFEST_DIR).join("doc_src");
     let docsrc = docsrc_path.to_str().unwrap();
+
+    // `sphinx-build` runs for several seconds even if none of the sources changed.
+    // This can result in the completely useless `sphinx-build` invocation almost doubling the
+    // duration of a `cargo b` (for a non-clean build).
+    // https://github.com/sphinx-doc/sphinx/issues/13727
+    let timestamp_file_path = mandir.join("build_timestamp");
+    if is_build_dir_up_to_date(&SPHINX_DOC_SOURCES, &timestamp_file_path) {
+        return;
+    }
+
     let args = &[
         "-j",
         "auto",


build.rs
```diff
@@ -14,6 +14,10 @@ fn canonicalize_str<P: AsRef<Path>>(path: P) -> String {
 
 const MANIFEST_DIR: &str = env!("CARGO_MANIFEST_DIR");
 
+#[cfg(feature = "embed-data")]
```
krobelus, line 17: nice, I'll take a look.

So this means that before this last commit, in

	cargo b; touch src/abbrs.rs; time cargo b

the second `cargo b` will run `sphinx-build`.
Also every time a relevant env file like `FISH_GETTEXT_EXTRACTION_FILE` changes.

I wonder if we can move the `sphinx-build` invocation into a separate dummy crate,
so cargo behaves a bit smarter -- changes in `src/` should never affect docs (except that we do need `fish_indent`, not sure).
I guess that still doesn't mean that that sphinx outputs will use the cargo cache, so maybe that's useless
if we are already going to add a timestamp file (that we might want anyway), not sure yet

build.rs
```diff
@@ -14,6 +14,10 @@ fn canonicalize_str<P: AsRef<Path>>(path: P) -> String {
...
+#[cfg(not(clippy))]
+const SPHINX_DOC_SOURCES: [&str; 3] = ["doc_src", "CHANGELOG.rst", "CONTRIBUTING.rst"];
```
krobelus: unrelated: I wonder if we should move `CHANGELOG.rst` and `CONTRIBUTING.rst` into `doc_src/` (and leave behind symlinks).

build.rs
```diff
@@ -59,13 +63,24 @@ fn main() {
...
     }
 
-    rsconf::rebuild_if_paths_changed(&["src", "printf", "Cargo.toml", "Cargo.lock", "build.rs"]);
+    rsconf::rebuild_if_paths_changed(&[
+        "src",
+        "printf",
+        "gettext-extraction",
+        "Cargo.toml",
+        "Cargo.lock",
+        "build.rs",
```
krobelus, line 67: should sort this alphabetically unless there is some other meaningful order

build.rs
```diff
@@ -391,6 +406,88 @@ fn get_version(src_dir: &Path) -> String {
     get_git_hash().expect("Could not get a version. Either set $FISH_BUILD_VERSION or install git.")
 }
 
+#[cfg(feature = "embed-data")]
+#[cfg(not(clippy))]
+/// Helper function for avoiding unnecessary rebuilds.
+/// Specify all files/directories which should trigger a rebuild if modified via `source_paths`.
+/// `cache_file` is used to store the time of the latest build.
```
krobelus, line 449: I wouldn't normally replicate variable names in comments.
But I suppose it is common Rust style, especially for doc comments, so maybe I'm on the wrong side of history.

build.rs
```diff
@@ -391,6 +406,88 @@ fn get_version(src_dir: &Path) -> String {
...
+            let p = p.as_ref();
+            if p.is_dir() {
+                std::fs::read_dir(p)
+                    .unwrap()
+                    .flatten()
+                    .map(|f| f.path())
+                    .collect::<Vec<_>>()
+                    .into_iter()
+            } else {
+                vec![PathBuf::from(p)].into_iter()
```
krobelus: this also doesn't handle nested directories like doc_src/cmds/.
I think we should add a generic and recursive `walk()` function (e.g. https://stackoverflow.com/questions/63542762/directory-traversal-in-vanilla-rust).
Then we shouldn't need a Vec.
(We could make `walk()` return an iterator (using `iter1.chain(iter2)` for recursive calls), but the callback approach might be simpler.)

build.rs
```diff
@@ -391,6 +406,88 @@ fn get_version(src_dir: &Path) -> String {
...
+            }
+        })
+        // Do not use `f.metadata()`, since that does not traverse symlinks.
+        .map(|p| std::fs::metadata(p).unwrap().modified().unwrap())
+        .max()
+        .unwrap()
+        .duration_since(UNIX_EPOCH)
+        .unwrap()
+        .as_millis();
```
krobelus, line 480: we should report the problem (+ reproducer to sphinx),
either they can fix it and we can get rid of this eventually,
or we missed some cache invalidation.

For one, this new logic won't invalidate the cache if `sphinx-build` updates.
So I think we should probably first figure out why `sphinx-build` takes 4 seconds if nothing changed.

build.rs
```diff
@@ -391,6 +406,88 @@ fn get_version(src_dir: &Path) -> String {
...
+    let cache_file_dir = cache_file.parent().unwrap();
```
krobelus, line 481: Hmm ok so we can't compare the timestamps of the target files
because we don't know what sphinx does, i.e. if we do

	youngest_source_file = max(source_file.modified() for source_file in os.walk(SPHINX_DOC_SOURCES))
	oldest_target_file = min(target_file.modified() for target_file in os.walk(target_dir))
	if youngest_source_file > oldest_target_file:
		return 'stale'

and sphinx doesn't update the mtime of some file, then we'd keep rebuilding.
So we cache the timestamp instead.

build.rs
```diff
@@ -391,6 +406,88 @@ fn get_version(src_dir: &Path) -> String {
...
+    std::fs::create_dir_all(cache_file_dir).unwrap();
```
krobelus, line 482: this assumes that there are no concurrent `cargo b` processes running this, which I think can be assumed to be true.
Otherwise, we'd want to write the cache file only after `sphinx-build` is complete,
else concurrent builds might fail to rebuild.

build.rs
```diff
@@ -391,6 +406,88 @@ fn get_version(src_dir: &Path) -> String {
...
+        .create(true)
+        .truncate(false)
+        .open(cache_file)
+        .unwrap();
+    let mut timestamp_file_content = vec![];
+    timestamp_file
+        .read_to_end(&mut timestamp_file_content)
+        .unwrap();
+    let timestamp_file_str = std::str::from_utf8(&timestamp_file_content).unwrap();
+    let formatted_timestamp = format!("{source_timestamp}");
```
krobelus: For simplicity, text files should always have a trailing newline.
I think we should add that and trim it when reading the file back.

build.rs
```diff
@@ -391,6 +406,88 @@ fn get_version(src_dir: &Path) -> String {
...
+    } else {
+        let build_timestamp = u128::from_str(timestamp_file_str).unwrap();
+        if source_timestamp <= build_timestamp {
+            return true;
+        }
+        timestamp_file.seek(SeekFrom::Start(0)).unwrap();
+        timestamp_file.set_len(0).unwrap();
+        timestamp_file
+            .write_all(formatted_timestamp.as_bytes())
+            .unwrap();
```
krobelus: the `write_all()` can be shared.

Also, I'm not sure if "open the file only once" is a strong enough motivation for the extra branch.
We could use `std::fs::read_to_string(path)` and then open in truncate mode.
I guess either works for me.

build.rs
krobelus: looks like sphinx-build overwrites files even if nothing changed.
I guess this is working as designed so not much to report.
So it does make sense for us to implement this logic.

I guess we might want to accept imperfect cache invalidation.
Maybe it's enough to throw in a check for `sphinx-build --version` too.
This won't help folks who try different patched versions of sphinx, but I'd imagine those will invoke the `sphinx-docs` targets explicitly..
Hmm. It would be great if we can avoid the need to ever have to run `cargo clean` for correctness.

Our historical workaround basically translates to `FISH_BUILD_DOCS=0 cargo b`. It was kinda okay as `cmake -DFISH_BUILD_DOCS=1` but asking developers to set full-blown environment variables is a bit gross, so I think we should be pragmatic here.

build.rs
danielrainer: I just tested GitHub's handling of symlinks, which isn't great: https://github.com/danielrainer/fish-shell/blob/move_docs_into_doc_src/CONTRIBUTING.rst

build.rs
danielrainer: I'd like to have a semantic link between the names in the doc comment and the parameter names, but AFAIK that's not a thing. From what I have seen, it is indeed common to use parameter names in doc comments. The issue is of course that the comments can easily become outdated when the code is changed and there are no automated checks, but I think this is easier to notice if parameters which no longer exist are mentioned in the docs, compared to docs which refer to the parameters less directly.

build.rs
danielrainer: I think cargo prevents concurrent builds. Otherwise we might indeed run into problems when some build assumes that the docs are up-to-date when they are not (they might even be missing entirely). If we want to address this we also need to check if sphinx-build handles concurrent builds correctly.

build.rs
danielrainer: I added a function which returns an iterator over all paths under a given path. I think for our purposes it is also desirable to check both the mtime of a symlink and of its target, so I added that as well (although we don't use symlinks for docs at the moment).

build.rs
danielrainer: > Maybe it's enough to throw in a check for `sphinx-build --version` too.

Yes, I guess we could do that. Maybe by adding a `target/fish-man/sphinx_version` file? Then we can compare it to the output of `sphinx-build --version` when the build timestamp is not outdated and update the file whenever we decide that a rebuild is needed. (Technically there is still a race condition between the two `sphinx-build` invocations, but I don't think we need to care about that.)

build.rs
danielrainer: Yes that was my reasoning as well.

build.rs
danielrainer: By moving this out of the `if` statement, the first branch becomes empty, so the statement can be simplified. Now the decision is between opening the file twice, or truncating the open file before writing to it. I don't have a strong preference, but given the verbosity of `OpenOptions` the current code is probably a bit more concise.

build.rs
danielrainer: It is a bit annoying for building in general that building `fish` depends on the man pages, but building the man pages depends on `fish_indent`.

Ideally, we would first build `fish_indent` without building anything as a dependency which embeds the man pages, then build the man pages before finally building `fish`.

At the moment, the man pages are embedded in `src/builtins/status.rs`, so we would have to build `fish_indent` without building this file (or at least rebuild it when later building `fish`). I don't know how difficult it would be to achieve that. We would probably have to extract the relevant code in `status.rs` into a separate library.

krobelus: right this is ugly.
I don't think there's a way to only rebuild `src/builtins/status.rs`, so we'd first build with the embed-data feature flag turned off and then turned on.
Alternatively, we could:
1. extract the `src/ast.rs` and any dependencies into a `fish-syntax` crate
    - things that don't fit into that crate such as `FLOG` would go into a separate `fish-common` crate
2. add to the `fish-syntax` crate a `fish-highlight` binary target that runs `make_pygments_csv()`
3. make the top-level crate [depend on the `fish-highlight` binary](https://rust-lang.github.io/rfcs/3028-cargo-binary-dependencies.html).

We could maybe also port the entire `fish_indent` to the parser crate, but that might bring unwelcome dependencies. Not sure.

I think a big part of the work is figuring out which files belong into which crate.
Splitting into crates will probably shorten compile time, especially incremental one,
and possibly bring other benefits.

krobelus: `print_help` assumes `fish` is in `$PATH`, I hope we can fix that by using `parser.eval("__fish_print_help")` with the appropriate IO redirections.

build.rs
danielrainer: Directory paths are now included as well, since their mtime might be relevant (e.g. when a file was deleted).

build.rs
danielrainer: Having to build twice is definitely not great, both for performance and because it requires more complex handling of builds, meaning normal `cargo` commands don't fully work for getting a complete build.

Splitting the code into separate crates seems like the better approach, but figuring out how to do that is nontrivial.

The `print_help` function assuming `fish` exists is an issue. Even now there might be issues with incorrect versions of the docs being displayed if the version of `fish` does not match `fish_indent`/`fish_key_reader`. Maybe the logic in `__fish_print_help.fish` could be put into a Rust crate all binaries depend on? With doc embedding, we probably want to avoid embedding docs into `fish_indent`/`fish_key_reader` which are irrelevant to these binaries.

Another thing which would probably have to get its own crate is the gettext logic. This might make the message extraction more complicated.

build.rs
```diff
@@ -391,6 +406,106 @@ fn get_version(src_dir: &Path) -> String {
...
+#[cfg(feature = "embed-data")]
+#[cfg(not(clippy))]
+/// Given a path, returns an iterator containing this path, and possibly others depending on what
+/// the input path refers to.
+/// If the path points to a file, only this file will be included.
+/// If it is a symlink, the path of the link will be included, as well as the result of recursing
+/// on its target.
+/// If it is a directory, the function recurses on each entry and combines the results into the
+/// returned iterator.
+fn ls_recursive<P: AsRef<Path>>(path: P) -> impl Iterator<Item = PathBuf> {
```
krobelus: doesn't seem like the best name (this abbreviation is not common in context, I read `is_recursive`). Maybe `read_dir_recursive`, since it's kinda similar to `std::fs::read_dir`?
Not important, I'm fine either way.

build.rs
```diff
@@ -391,6 +406,106 @@
...
+    fn walk(path: PathBuf) -> Box<dyn Iterator<Item = PathBuf>> {
+        if path.is_file() {
+            return Box::new([path].into_iter());
+        }
+        if path.is_symlink() {
+            return Box::new(
+                [path.clone()]
+                    .into_iter()
+                    .chain(walk(path.read_link().unwrap())),
```
krobelus: Do symlink cycles (`ln -s . badlink` or `ln -s .. badlink`) cause an infinite loop?
Seems like we should better panic in cases that we don't support.
AFAICT we only need to support symlinks to regular files.

build.rs
```diff
@@ -391,6 +406,106 @@
...
+        .write(true)
+        .create(true)
+        .truncate(false)
+        .open(cache_file)
+        .unwrap();
+    let mut timestamp_file_content = vec![];
+    timestamp_file
+        .read_to_end(&mut timestamp_file_content)
+        .unwrap();
+    let timestamp_file_str = std::str::from_utf8(&timestamp_file_content).unwrap().trim();
```
krobelus: I'd probably either move this definition into the if block,
or keep it here and shadow `timestamp_file_content` (by using the same name),
because we really don't need both in scope.
I'm good with either; I like shadowing though I guess it's redundant work in the `content.is_empty()` case.

build.rs
```diff
@@ -391,6 +406,106 @@
...
+    if !timestamp_file_content.is_empty() {
+        let build_timestamp = u128::from_str(timestamp_file_str).unwrap();
+        if source_timestamp <= build_timestamp {
+            return true;
+        }
+        timestamp_file.seek(SeekFrom::Start(0)).unwrap();
+        timestamp_file.set_len(0).unwrap();
+    }
+    timestamp_file
+        .write_all(formatted_timestamp.as_bytes())
```
krobelus: can inline the `formatted_timestamp` variable now

build.rs
```diff
@@ -400,6 +515,15 @@
     let sec1dir = mandir.join("man1");
     let docsrc_path = canonicalize(MANIFEST_DIR).join("doc_src");
     let docsrc = docsrc_path.to_str().unwrap();
+
+    // `sphinx-build` needs several seconds to realize that no rebuild is required.
```
krobelus: I don't think sphinx even realizes that (at least not for the manpage builder),
so we should probably adjust this comment.
I filed https://github.com/sphinx-doc/sphinx/issues/13727, maybe someone can shed light on this (though it probably
wouldn't be super hard to find out ourselves why sphinx rebuilds man pages everytime).

But even if we don't root-cause it, I'm good with this approach.

build.rs
danielrainer: https://github.com/rust-lang/rust/issues/69684 suggests `read_dir_all`, so maybe that might make sense.

build.rs
danielrainer: Good point, that should have been moved into the if block during the changes to the condition.

build.rs
danielrainer: There is no protection against infinite loops at the moment. If we want that, I think we should have proper support to detect cycles of arbitrary length, e.g. via building a set of paths and only recursing on paths which are not in the set already. Not sure if the added complexity is worth it. I would be fine with the build script not working with broken symlinks.

build.rs
danielrainer: I renamed it to `read_dir_all` for now.

build.rs
danielrainer: Fair point about the comment, I did not actually check in detail what `sphinx-build` does. Running
```sh
sphinx-build -j auto -q -b man -c doc_src -d target/fish-man doc_src target/fish-man/man1
```
when `target/fish-man` is empty seems to take a bit longer than repeated runs with an up-to-date cache, but the difference is much smaller than the runtime with everything cached. So the caching seems to do something, but it is not that useful.

I removed the unconfirmed speculation about the behavior of `sphinx-build` from the comment and added a link to the issue you opened.

build.rs
krobelus: two options are
- detect cycles (and panic)
- panic when a symlink resolves to a directory

danielrainer: Yes. The second option is certainly the simpler one. Since we don't use symlinks at all for the docs atm, it would be sufficient.

build.rs
krobelus: there is an encouraging response at https://github.com/sphinx-doc/sphinx/issues/13727
So we could try that (i.e. improve sphinx caching); else the only thing missing here is treatment of symlinks I think.
(I'll first work on some other fish regressions)

build.rs
danielrainer: Implemented, code now panics on symlinks whose target is not a file.

build.rs
danielrainer: Encouraging in the sense of encouraging us to fix it ourselves I guess, which I am not particularly keen on doing in a Python codebase I am unfamiliar with, especially since I know pretty much nothing about Sphinx's compilation model. And even if we got a fix merged right now, it might take some time for it to make it onto the systems of fish devs, so I think merging our workaround is worthwhile regardless.

Symlink handling is addressed now.

Issue #11637, opened 2025-07-01
Unexpected export shadowing and unshadowing behavior


joveian:
fish version 4.0.2, Arch Linux fish package, kitty terminal

In the following sequence the env | grep command fails both times for me,  however if the first env | grep command is left out the second succeeds.  The first set -x | grep always fails and the second always shows the variable.
`set -l hi hihi`
`set -gx hi $hi`
`env | grep hihi`
`set -x | grep hihi`
`set -el hi`
`env | grep hihi`
`set -x | grep hihi`

I almost didn't try the fresh fish command suggested in the bug template because of course that couldn't matter but it was quick so I tried it and there the second env | grep works.  This turned out to be because the default prompt exports a variable and that fixes the exports.  

I think there are three issues here:

1. the second env | grep should always succeed.  I guess there should be a check if an exported variable is unshadowed in the set -e path that isn't there (I haven't looked at the code yet).  It looks like tests/checks/env.fish would be a good place to test for this.
2. set --show mentions both the local unexported and global exported but does not mention that the export will be shadowed, which would make the shadowing behavior clearer
3. I couldn't find the export shadowing described in the language reference but did find it in the set documentation, maybe at least a quick mention under Exporting Variables and pointer to the set page scope rules would be helpful or copy the full export scope rules to that section

I am slow (so anyone who can just do it quickly should feel free) but should be able to at least provide a suggested documenation patch and test and possibly the code changes will be obvious enough.

krobelus:
> the second env | grep should always succeed.

yes. It works for me.

> set --show mentions both the local unexported and global exported but does not mention that the export will be shadowed, which would make the shadowing behavior clearer

yeah changing the output a little bit would be fine since it's not supposed to be machine-readable

Looks like bash has the other behavior (`local hi` will not affect the export behavior).
Not sure which one makes more sense. fish's behavior definitely feels surprising

joveian:
I started working on a test and had trouble making one that failed.  Turns out it only happens once in a script or interactive session, then it work.  So some kind of initialization issue I guess (but strange that all the complexity of the config script doen't seem to influence it).  If you put the above commands in a script and run it with a new fish instance you should only see the last set -x output.  Or maybe it works... I was going to say make sure you didn't already have it exported from first trying it but thought to just suggest running env -i env2.fish, only that works for me.  But env -i fish -N env2.fish fails.  I do have some config files that do extra work if an environment variable isn't set so I guess suff in the config files can affect it.

[...]

PR #11638, opened 2025-07-02
Test export shadowing and unshadowing, currently fails.  Issue #11637

joveian:
Doesn't fail if added to the existing env.fish and seems like it can be affected by config files so hopefully being in a separate file run with fish -N this will reliably fail for everyone until fixed.  See discussion in Issue #11637 for more details.

### DIFF
--- a/tests/checks/env2.fish
+++ b/tests/checks/env2.fish
@@ -1,0 +1,14 @@
+# RUN: %fish -N %s
+
+set -f FISH_ENV_TEST_2 def
+set -gx FISH_ENV_TEST_2 abc
+env | string match FISH_ENV_TEST_\*
+# No output
+
+set -ef FISH_ENV_TEST_2
+env | string match FISH_ENV_TEST_\*
+# CHECK: FISH_ENV_TEST_2=abc
+
+set -e FISH_ENV_TEST_2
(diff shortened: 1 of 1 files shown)


ridiculousfish:
Nice find and thank you!

As you observed in the discussion, this works in interactive mode, and fails in non-interactive mode. Let me try to figure out why.

joveian:
Thanks for looking at it and for the nice shell and excellent documentation :).

I don't think it is interactive vs non-interactive.  The scripts fail without using -i to fish (I did mention testing with env -i) and this test fails the automated checks which I assume are non-interactive (except the few python ones).  I just tried looking at the startup profile when running this test without -N and it does follow the non-interactive path in my config and fails.  Just something about the existing env.fish test makes it stop happening and trying to run the same test twice works the second time.

Edit: Oh, I didn't read your comment correctly... it does fail for me in interactive mode as well when it is the first thing I do and with my custom fish_prompt that just prints a space with a color based on $status.  The default fish_prompt exports a variable and when I do that interactively before the second env | grep (or string match) then it succeeds for me, so I think that is why it isn't seen interactively with the default settings.  I do export some global variables in my config (before the non-interactive early return) so just exporting other global variables earlier doesn't seem to cause the test to work.

PR #11644, opened 2025-07-06
Support upper/lower casing selection

### LINKED ISSUE #11639: [bug]: Undocumented lowercase `gu` / uppercase `gU` in command/visual vim mode; partly unexpected behavior
`fish --version`: fish, version 4.0.2
macOS 15.2.0 Sequoia

_Please tell us if you tried fish without third-party customizations by executing this command and whether it affected the behavior you are reporting:_ yep done that same effect

Really appreciate vim mode in fish! However it seems `gu` and `gU` aren’t documented yet: https://fishshell.com/docs/current/interactive.html#command-mode

Yet it seems `gU` can have 2 different effects:

- command mode: uppercase till next word (`gUw` in Vim except seemingly treating underscore as word boundary)
- visual mode: toggles casing like `~` (contrary to vim)

Whereas `gu` seems to have no effect on visual selections which I think is inconsistent thus unexpected behavior.
In command mode however it lowercases corresponding to above-mentioned `gU` behavior.

danielrainer:
Fixes #11639.


src/reader.rs
```diff
@@ -3571,6 +3571,41 @@ impl<'a> Reader<'a> {
...
+                        WString::from_chars(
+                            text.iter()
+                                .flat_map(|c| c.to_uppercase())
+                                .collect::<Vec<char>>(),
+                        )
+                    } else {
+                        WString::from_chars(
+                            text.iter()
+                                .flat_map(|c| c.to_lowercase())
+                                .collect::<Vec<char>>(),
```
krobelus: very minor nit: should probably use `WString::from_iter()` here to make it a bit shorter.

krobelus (approved):

src/reader.rs
```diff
@@ -3571,6 +3571,41 @@ impl<'a> Reader<'a> {
...
+                    // Restore the buffer position since replace_substring moves
+                    // the buffer position ahead of the replaced text.
+                    // Note: This does not take string length changes into account.
+                    // E.g.: When the cursor was at the right of the selection,
+                    // the selection contains 'ẞ', which is uppercased into 'SS',
+                    // the cursor will stay at the same offset, but it will not be on the same
+                    // character as before.
+                    // The position calculations work on codepoints rather than graphemes, which can
+                    // result in additional issues.
+                    self.update_buff_pos(elt, Some(buff_pos));
```
krobelus, line 3598: `rl::TogglecaseSelection` and probably others have the same problem;
ideally we could extract this logic into a function so there's only one place
where this comment applies. But that's probably utopic..

danielrainer: Yes, I did not attempt to refactor this here. Extracting it would maybe reduce some code duplication (and make the comment easier to find for the other affected locations), but actually addressing the underlying problem would probably require quite major refactoring.

krobelus: In general, Vi mode is missing a bunch of features, and so do 3rd party bindings for other modal editing styles.
I wonder what we should be doing, implement them in the core (possibly via libraries) or expose enough functionality so they can be plugins.. historically, both have had problems

danielrainer: I think implementing them ourselves would be a lot of effort. I once looked a bit at the implementation of word movement commands (`w`, `e`, `b`, ...) because I noticed that they behave differently in vi and fish, and realized that the implementation in vi (or vim, or nvim) is significantly more complex than what fish does at the moment. So I think implementing a proper vi mode which is consistent with vi's behavior and supports all the relevant features is not something we can realistically do in fish unless someone is motivated to reimplement a significant part of vi. Exposing the entire command line and routing inputs via a library which offers vi-style editing of the buffer might work, but I have not looked into this approach, which libraries we could use and if redirecting input is acceptable for fish.

PR #11647, opened 2025-07-09
Add completions for cpan and t-rec

visrosa:
Adds completions for `t-rec` and `cpan`.

I'm working on some more completions, let me know if I should create separate pull requests for each command or if it's okay to batch them like this.

### DIFF
--- a/share/completions/cpan.fish
+++ b/share/completions/cpan.fish
@@ -1,0 +1,47 @@
+# fish completion for Perl's cpan
+
+function list_installed_modules
+    # Following IRC's #fish suggestion to use </dev/null as cpan might go interactive
+    cpan -l </dev/null | while read -l line
+		# Filter out unrelated messages or notifications
+		if string match -qr -- '^\w.*\t\w.*$' $line
+			string replace -r -- '\t(.*)' '' $line | \
+			string escape --style=script
+		end
+	end
+end
+
+set -lx installed_modules (list_installed_modules | string collect)
+
+complete -c cpan -o a -d "Creates a CPAN.pm autobundle with CPAN::Shell->autobundle"
+complete -c cpan -o A -d "Show primary maintainer for specified module" -a $installed_modules
+complete -c cpan -o c -d "Runs a `make clean` in the specified modules directories"
+complete -c cpan -o C -d "Show the Changes files for specified module" -a "$installed_modules"
+complete -c cpan -o D -d "Show module details" -a "$installed_modules"
+complete -c cpan -o f -d "Force the specified action"
+complete -c cpan -o F -d "Turn off CPAN.pm's attempts to lock anything"
+complete -c cpan -o g -d "Download latest distribution of module to current directory" -a $installed_modules
+# complete -c cpan -o G -d "UNIMPLEMENTED"
+complete -c cpan -o h -d "Print a help message and exit"
+complete -c cpan -o i -d "Install specified module"
+complete -c cpan -o I -d "Load 'local::lib' (think like '-I' for loading lib paths)"
+complete -c cpan -o j -d "Load file with CPAN configuration data"
+complete -c cpan -o J -d "Dump the configuration in the same format that CPAN.pm uses"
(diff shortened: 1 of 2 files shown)


share/completions/cpan.fish
```diff
@@ -0,0 +1,47 @@
+# fish completion for Perl's cpan
+
+function list_installed_modules
```
krobelus: namespace this as `__fish_cpan_list_installed_modules` or similar

share/completions/cpan.fish
```diff
@@ -0,0 +1,47 @@
...
+    # Following IRC's #fish suggestion to use </dev/null as cpan might go interactive
+    cpan -l </dev/null | while read -l line
+		# Filter out unrelated messages or notifications
+		if string match -qr -- '^\w.*\t\w.*$' $line
+			string replace -r -- '\t(.*)' '' $line | \
```
krobelus: The capture group is unused, so maybe lose the parentheses.
Also, there's no need for the trailing backslash if there's already "|" at the end of line
Also, we'll format this file with `fish_indent`.

share/completions/cpan.fish
```diff
@@ -0,0 +1,47 @@
...
+			string escape --style=script
+		end
+	end
+end
+
+set -lx installed_modules (list_installed_modules | string collect)
+
+complete -c cpan -o a -d "Creates a CPAN.pm autobundle with CPAN::Shell->autobundle"
```
krobelus: are you using `-o` instead of `-s` because something like `-ac` doesnt' make sense?
Is there any combination that would make sense? If in doubt, we should use `-s`.
Something like `cpan -Ol` is allowed even though it seems to have the
same behavior as `cpan -O`.. so probably no one should every use it.

share/completions/cpan.fish
```diff
@@ -0,0 +1,47 @@
...
+complete -c cpan -o A -d "Show primary maintainer for specified module" -a $installed_modules
```
krobelus: The `-x` seems useless, and this might change over the lifetime of a fish process.
So I think we should recompute (`-a '(__fish_cpan_list_installed_modules)'`.
If we're concerned about the long runtime, we can use `__fish_cached` to cache it for perhaps a few hours.

share/completions/t-rec.fish
```diff
@@ -0,0 +1,34 @@
+# fish completion for t-rec (https://github.com/sassman/t-rec-rs
+
+function __t_rec_time_unit
```
krobelus: maybe `__fish_t-rec_time_unit` or similar (the important thing is the `__fish` prefix).
Same below.

share/completions/t-rec.fish
```diff
@@ -0,0 +1,34 @@
...
+    set -l cur (commandline --current-token)
+    if string match -qr '^\d+$' -- $cur
+        echo $cur"ms"\t"milliseconds"
+        echo $cur"s"\t"seconds"
+        echo $cur"m"\t"minutes"
+    end
+end
+
+function __t_rec_window_list
+	string replace -r '\s*(.*)\|\s+(\d+)' '$2\t$1' (t-rec --ls | tail -n +2)
```
krobelus: maybe add a `--` just inc case `t-rec --ls` returns values leading with `-`.

share/completions/t-rec.fish
```diff
@@ -0,0 +1,34 @@
...
+end
+
+set -lx t_rec_decor shadow none
+set -lx t_rec_bg white black transparent
```
krobelus: also no need to export these variables to subprocesses, since they'll only 
be used inside fish. Actually it seems better to inline them.

krobelus:
> let me know if I should create separate pull requests for each command or if it's okay to batch them like this.

smaller is usually better but for uncontentious changes it's fine to reuse the same PR.
It'd probably be better to separate unrelated completions by putting them into separate commits still

visrosa:
Suggested changes have been implemented.

PR #11654, opened 2025-07-12
Update __fish_print_hostnames.fish

azikdinal:
There was an issue in autocomplete of ssh. 

When you put in ~/.ssh/config line like this:

"Include ${HRL_SSH}/onprem_config"

and then trying to use fish complete for ssh, for example:

"ssh -J" and press key <b>Tab</b> it throughs an error that fish cannot understand ${HRL_SSH} with brackets.

## Description

Talk about your changes here.

Fixes issue #

## TODOs:

### DIFF
--- a/share/functions/__fish_print_hostnames.fish
+++ b/share/functions/__fish_print_hostnames.fish
@@ -76,6 +76,11 @@ function __fish_print_hostnames -d "Print a list of known hostnames"
 
             set -l new_paths
             for path in $paths
+                # while ssh_config is using brackets to resolve env, they should be removed
+                # example
+                # in ssh_config: ${SOME_PATH}
+                # in fish:       $SOME_PATH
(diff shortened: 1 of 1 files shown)


share/functions/__fish_print_hostnames.fish
```diff
@@ -76,6 +76,11 @@ function __fish_print_hostnames -d "Print a list of known hostnames"
 
             set -l new_paths
             for path in $paths
+                # while ssh_config is using brackets to resolve env, they should be removed
+                # example
+                # in ssh_config: ${SOME_PATH}
+                # in fish:       $SOME_PATH
+                set path (string replace -r '\${([^}]+)}' '$1' $path)
```
krobelus, line 83: can we make it work for `${FOO}bar` too?
A possible fish syntax would be `"$FOO"bar`

PR #11672, opened 2025-07-20
Add an example replicating bash behavior to `prompt_pwd` documentation

omentic:
## Description

I work in deep directories sometimes, and find full paths to be too large, and do not find much value in shortened paths. I made this change locally and thought it would be useful to others. This setting mimics more closely bash behavior.

## TODOs:

### DIFF
--- a/doc_src/cmds/prompt_pwd.rst
+++ b/doc_src/cmds/prompt_pwd.rst
@@ -55,3 +55,9 @@ Examples
 
     >_ prompt_pwd --full-length-dirs=2 --dir-length=1
     /t/b/s/with/mustard
+
+    >_ echo (prompt_pwd | string split /)[-1]
+    mustard
+
+    >_ echo (string join / (prompt_pwd | string split /)[-3..-1])
+    s/with/mustard


doc_src/cmds/prompt_pwd.rst
```diff
@@ -19,6 +19,8 @@ To change the number of characters per path component, pass ``--dir-length=`` or
 
 To keep some components unshortened, pass ``--full-length-dirs=`` or set :envvar:`fish_prompt_pwd_full_dirs` to the number of components. This defaults to 1, keeping the last component.
 
+To print only up to a certain depth of parent directories, pass ``--dir-depth`` or set :envvar:`fish_prompt_pwd_dir_depth` to the desired depth. Setting it to 0 will print an unbounded number of parent directories, i.e. the full path. This defaults to 0.
```
krobelus: the name "depth" is slightly surprising because we're going inside out.
Though I don't really have a better option; maybe
`fish_prompt_pwd_parent_dir_count`

share/functions/prompt_pwd.fish
```diff
@@ -24,11 +24,24 @@ function prompt_pwd --description 'short CWD for the prompt'
...
+    set -q fish_prompt_pwd_dir_depth
+    or set -l fish_prompt_pwd_dir_depth 0
+
     for path in $argv
         # Replace $HOME with "~"
         set -l realhome (string escape --style=regex -- ~)
         set -l tmp (string replace -r '^'"$realhome"'($|/)' '~$1' $path)
 
+        # Shorten to at most $fish_prompt_pwd_dir_depth components, unless set to 0.
+        if test "$fish_prompt_pwd_dir_depth" -gt 0
```
krobelus: Right, this option makes sense.
I wonder if we should prefix shortened paths with an ellipsis,
that is

	set tmp "$(string shorten --left -m 1 ..)/$tmp"

	set fish_prompt_pwd_dir_depth 1
	# …/fish-shell
	set fish_prompt_pwd_dir_depth 2
	# …/git/fish-shell

though I guess that would stray from what bash does

share/functions/prompt_pwd.fish
```diff
@@ -24,11 +24,24 @@
...
+            set -l split (string split / $tmp)
+            set -l length (math (count $split) - (math $fish_prompt_pwd_dir_depth - 1))
+            set tmp $split[$length..]
```
krobelus: This misbehaves if length is 0 or negative.
We should probably fix this by doing
something to the effect of `set length (math "min($length, 1)")`

	$ prompt_pwd --dir-depth=1
	fish-shell
	$ prompt_pwd --dir-depth=2
	git/fish-shell
	$ prompt_pwd --dir-depth=3
	~/git/fish-shell
	$ prompt_pwd --dir-depth=4
	embedded:functions/prompt_pwd.fish (line 42): array indices start at 1, not 0.
	            set tmp $split[$length..]
	                            ^
	in function 'prompt_pwd' with arguments '--dir-depth=4'
	~/g/fish-shell
	$ prompt_pwd --dir-depth=5
	fish-shell
	[1] $ prompt_pwd --dir-depth=6
	git/fish-shell

(also note that `--dir-depth=5` results in a spurious exit code of 1, which we shoudl fix)

share/functions/prompt_pwd.fish
omentic: The user can do that pretty easily by querying the value of `prompt_pwd` in their prompt: prepending `...` if `(string split / (prompt_pwd))[1]` is ` ` or `~`. The other way around seems like a bit more of a pain.

doc_src/cmds/prompt_pwd.rst
omentic: Yeah, I don't love the naming. I had something involving `parent_dir` before, but it isn't *entirely* accurate: because with a value of 1 no parent dirs are printed.

krobelus:
> The user can do that pretty easily by querying the value of `prompt_pwd` in their prompt: prepending `...` if `(string split / (prompt_pwd))[1]` is ` ` or `~`. The other way around seems like a bit more of a pain.

Yeah that might be fair.

My ulterior motive was that we could maybe add a limit by default (like depth 7)..
the ellipsis might avoid confusion.

But as you said we can add it in the caller, and if we don't add a default limit, I don't really care.

krobelus:
> Yeah, I don't love the naming. I had something involving `parent_dir` before, but it isn't *entirely* accurate: because with a value of 1 no parent dirs are printed.

maybe we could even get rid of the need for a magic fish-specific option by instead doing

	prompt_pwd | string match -r '([^/]+/){,5}$'

untested but something like thos should work - it's supposed to match the
prompt_pwd suffix consisting of up to 5 path components.

omentic:
Even simpler: `(prompt_pwd | string split '/')[-1]` works nicely.

Though it's obvious in retrospect, I think this might be nice to document. I might repurpose this PR if that's ok?

omentic:
I'm very happy with using `string` primitives instead, it sort of slipped my mine you can just. do that.

I've replaced the old code with just adding an example to the documentation :+1:

PR #11679, opened 2025-07-23
Add docs for distros supporting native builds

danielrainer:
Document on which distros we aim to support building fish using only dependencies packaged for the distro.
The intention is providing clarity about which dependency versions can be used safely.

For now, this only contains Rust versions, but it might also be useful for other dependencies, like Python versions. Once we clearly indicate distro version support, it becomes significantly easier to figure out if a dependency version can be changed.

Inspired in part by discussion https://github.com/fish-shell/fish-shell/discussions/11584

## TODOs:
- [ ] Replace the stubs by fish's actual support policy
- [ ] Track the Rust versions of all supported distros with references.
- [ ] I'm not that happy with "native builds", since that is often used with a somewhat different meaning. Do we have better ideas?
- [ ] Do we want similar tracking for other dependencies, like Python?
- [ ] Can we get decent table formatting? Using [list-table](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/tables.html#list-table-directive) results in GitHub just not rendering the table.
- [ ] Location of these docs: I just created a top-level file to get started. Is there a better location (and file name)?

### DIFF
--- a/NATIVE_BUILDS.rst
+++ b/NATIVE_BUILDS.rst
@@ -1,0 +1,80 @@
+###########################################
+Distros on which fish should build natively
+###########################################
+
+Fish should build on some distributions without having to install software not packaged for these distributions.
+This document is intended to help with deciding which versions of build software (such as compilers) we can safely use without running into difficulties on platforms where we want native builds.
+Using dependencies which are not packaged for building the fish package can be annoying and should be avoided.
+On the other hand, new versions of dependencies introduce desirable features, which we want to use eventually.
+Tracking which dependency versions are available in distros helps with deciding when to update and to which versions.
+
+Distributions for which we aim to support native builds
+=======================================================
+
+TODO: Define which versions are supported in a way that is stable across releases. For now, here are some suggestions.
+Inspired by discussion https://github.com/fish-shell/fish-shell/discussions/11584.
+
+* Debian stable | 13 (Trixie)
+* Fedora supported releases (release N-2 is supported until 4 weeks after N is released) | 41, 42
+* FreeBSD ?
+* macOS ?
+* OpenBSD ?
+* OpenSUSE ?
+* Ubuntu latest LTS release | 24.04 LTS (Noble Numbat)
+* Ubuntu latest release | 25.10 (Questing Quokka)
+* Any others ?
+
+
+Rust (MSRV)
+===========
+
+Add new versions here as they are released, and update Rust versions if a new one is
+packaged for a certain distro version.
+Remove distro versions which are no longer supported.
+
+..
+    The following should be a table, but GitHub seems unable to render list-table:
+    https://sublime-and-sphinx-guide.readthedocs.io/en/latest/tables.html#list-table-directive
+
+* - Distro version
+  - available Rust version
+  - Reference
+* - Debian 11 (Bullseye)
+  - 1.78
+  - https://packages.debian.org/bullseye/rustc-web
+* - Debian 12 (Bookworm)
+  - 1.78
+  - https://packages.debian.org/bookworm/rustc-web
+* - Debian 13 (Trixie)
+  - 1.85
+  - https://packages.debian.org/trixie/rust-all
+* - Fedora 41
+  - 1.90
+  - https://packages.fedoraproject.org/pkgs/rust/rust/fedora-41-updates.html
+* - Fedora 42
+  - 1.90
+  - https://packages.fedoraproject.org/pkgs/rust/rust/fedora-42-updates.html
+* - Fedora 43
+  - 1.90
+  - https://packages.fedoraproject.org/pkgs/rust/rust/fedora-43-updates.html
+* - Ubuntu 22.10 (Jammy Jellyfish)
+  - 1.82
+  - https://packages.ubuntu.com/jammy/rust-1.82-src
+* - Ubuntu 24.04 LTS (Noble Numbat)
+  - 1.82
+  - https://packages.ubuntu.com/noble/rust-1.82-all
+* - Ubuntu 25.04 (Plucky Puffin)
+  - 1.84
+  - https://packages.ubuntu.com/plucky/rust-all
+* - Ubuntu 25.10 (Questing Quokka)
+  - 1.85
+  - https://packages.ubuntu.com/questing/rust-all
+
+**Additional constraints**:
+
+* Rust dropped support for macOS versions older than 10.12 with Rust 1.74: https://blog.rust-lang.org/2023/09/25/Increasing-Apple-Version-Requirements/, so fish might not work on old macOS versions if we update our MSRV to >= 1.74.
+
+Current MSRV of fish: 1.70
+
+Most recent Rust version available on all supported platforms: 1.78 (or 1.73 when including old
+macOS)


NATIVE_BUILDS.rst
```diff
@@ -0,0 +1,74 @@
...
+
+* Debian stable | 12 (Bookworm)
+* Fedora supported releases (release N-2 is supported until 4 weeks after N is released) | 41, 42
+* FreeBSD ?
+* macOS ?
+* OpenBSD ?
+* OpenSUSE ?
+* Ubuntu latest LTS release | 24.04 LTS (Noble Numbat)
+* Ubuntu latest release | 25.04 (Plucky Puffin)
+* Any others ?
```
krobelus, line 25: I assume that no one wants to be updating these values manually,
at least not in multiple places.  Also, we should use the Beyonce rule.

Hence (unless we autogenerate this type of documentation), we should
link to the truth, that is,
- `.github/actions/rust-toolchain@oldest-supported/action.yml`
- `.github/workflows/main.yml`
- `./.cirrus.yml`
- etc.

Today we have two sets of automatically-tested platforms
1. the ones tested in github actions and Cirrus CI
2. the ones tested in [OBS nightlies](https://build.opensuse.org/package/show/shells:fish:nightly:master/fish)[^1]

[^1]: Anyone interested in tracking OBS nightly failures can subscribe there.
Currently, the openSUSE Leap 15.6 builds are broken again because that uses python
3.6; I have not yet figured out how to test my fix (or bump their Python version), ideally we have
a `docker/opensuse-leap.Dockerfile` for this.

NATIVE_BUILDS.rst
```diff
@@ -0,0 +1,74 @@
...
+
+
+Rust (MSRV)
+===========
```
krobelus, line 29: I think the main questions remain
- what's the MSRV policy
- what's the platform support policy 

once we have "good enough" answers, we can tell our (renovate?)bot
to regularly create PRs to update MSRV, github CI platforms, cirrus CI platforms, dependencies etc.

I think the results could be expressed fairly concisely, so that we don't need a separate document but can have `README.rst` say something like "if your system is not X years past EOL, it should be supported".

NATIVE_BUILDS.rst
```diff
@@ -0,0 +1,74 @@
...
+  - https://packages.fedoraproject.org/pkgs/rust/rust/fedora-42-updates.html
+* - Ubuntu 22.10 (Jammy Jellyfish)
+  - 1.82
+  - https://packages.ubuntu.com/jammy/rust-1.82-src
+* - Ubuntu 24.04 LTS (Noble Numbat)
+  - 1.82
+  - https://packages.ubuntu.com/noble/rust-1.82-all
+* - Ubuntu 25.04 (Plucky Puffin)
+  - 1.84
+  - https://packages.ubuntu.com/plucky/rust-all
```
krobelus, line 68: in order to keep things simple, we should probably pick only one
(e.g. Debian stable) and then we don't need to explictly document
anything that's newer -- those should all be supported.

NATIVE_BUILDS.rst
danielrainer: If we have automated setups I agree that it makes sense to link to them. I think one of the main goals of this PR is to take inventory and provide a clear picture of what we support, not just at this moment, but also how this will change as new versions are released. This policy should be documented, and there is not much that can be automated about it. The translation of the policy into platform versions supported at this moment could of course be automated, or be done via links to the files you suggest, provided we have such files for all supported platforms.

NATIVE_BUILDS.rst
danielrainer: > what's the platform support policy

This PR is mainly intended to establish this. I think the MSRV policy can be downstream of that; use the most recent version available on all supported platforms. Whether the MSRV is updated manually or automatically does not matter that much, in my opinion, but I don't know of a way to find obtain the most recent rust version for all platforms automatically, especially given that in some cases additional packages are introduced, e.g. Debian's `rustc-web`, so unless we figure this out we don't really have the option to automatically update our MSRV, at least if we want to ensure builds keep working on all supported platforms. For the parts that we can automate, I agree that it makes sense to do so.

> if your system is not X years past EOL, it should be supported

I think it's important to differentiate between:
1. "fish builds on the system using only distro-packaged tools"
2. "fish can be built for the system (and runs on it), but might require installation of dependencies outside of what's distro-packaged (e.g. using rustup)"
3. maybe even "fish can be built to run on the system, but might require cross-compilation" (e.g. because some build dependency does not work on the target system)

Given the large number of distros, comprehensively tracking 1. is unrealistic. The other definitions are even harder to pin down. I think the realistic approach is to specify minimum build dependency versions and rely on user reports for breakage on platforms we don't test automatically.

Most users probably care about whether fish is packaged (and up-to-date) for their distro, which could be served by embedding something like a Repology badge.

NATIVE_BUILDS.rst
danielrainer: That should mostly be fine. For some distros which have both LTS and non-LTS versions, like Ubuntu, it might happen that a more recently released version has older packages than an older LTS version. Not sure if that happens in practice, but it is something that we should consider.

NATIVE_BUILDS.rst
```diff
@@ -0,0 +1,80 @@
...
+* - Ubuntu 25.10 (Questing Quokka)
+  - 1.85
+  - https://packages.ubuntu.com/questing/rust-all
+
+**Additional constraints**:
+
+* Rust dropped support for macOS versions older than 10.12 with Rust 1.74: https://blog.rust-lang.org/2023/09/25/Increasing-Apple-Version-Requirements/, so fish might not work on old macOS versions if we update our MSRV to >= 1.74.
+
+Current MSRV of fish: 1.70
```
krobelus, line 77: let's not add anything that will go stale (i.e. add extra maintenance work)
Probably we should briefly mention the MSRV policy in CONTRIBUTING.

Is there anything else from this PR that we want to merge?
OpenBSD has Rust 1.86 (according to https://github.com/fish-shell/fish-shell/issues/11892)
so our platform support should be okay.

I guess MSRV is the main thing, but Python also.
We should probably use the same approach (Debian stable, i.e. Python 3.13) though I guess we could go `oldstable` as well

danielrainer: > let's not add anything that will go stale

Yeah, I don't really intend to have this PR merged, it's mainly a place to collect information about which versions are available where to get a sense of what we can use.

Using a simple policy like "whatever is available in Debian stable" might be fine, at least when Debian stable is relatively fresh, later in Debian's release cycle we might also consider updating to something other than the default Rust version there (if Debian provides something like `rustc-web`).

I do think we should be explicitly mention each MSRV update in the changelog, including the exact version we change to.

For Python, I don't care as much about having recent versions available, especially if we migrate away from it. But raising the requirement a bit to avoid having to deal with really old versions would be nice. Debian stable or oldstable should be fine. Maybe it's also an option to let `uv` handle Python versions.

krobelus: rustc-web is 1.85 in debian-oldstable (which is debian 12 today), so our [OBS builds](https://build.opensuse.org/project/show/shells:fish:release:4) are probably gonna be mostly fine

> I do think we should be explicitly mention each MSRV update in the changelog

we should, yes

danielrainer: Well, if we're already mentioning MSRV updates it doesn't hurt to include the version we're updating to, so people who are interested don't have to go looking elsewhere to find the version.

PR #11681, opened 2025-07-25
Additional cygwin support

A2uria:
See https://github.com/fish-shell/fish-shell/pull/11238#issuecomment-3113910238 first.

- [x] https://github.com/fish-shell/fish-shell/commit/6644cc9b0e29841e3d0a85fbc672a95c4a2fd000 introduced compatibility issues. This can be easily fixed by adding cygwin check. See changes for details.

- [x] `.gitattributes` uses `* text=auto` and `*.fish text`, breaking builds using `git clone` or `cargo install` on windows.

  > I'm using `cargo install` to make fish with embedded data.
  >
  > However, I'm not using `cargo install --git` because `*.fish` will end with CRLF in this case. The same goes for Git for Windows, regardless of `core.autocrlf=false`. Tarballs works well.

  https://git-scm.com/docs/gitattributes says that the default is eol=crlf on Windows. However it seems that git and cargo won't respect `core.autocrlf`.

  `*.fish eol=lf` works, but there is **`share/__fish_build_paths.fish.in`** which is configured with cmake. MSYS2 uses tarball so it will work as before, but for some people who use `git clone` and cmake to build fish, this might be a problem.

- [x] It's still possible that `select` returns -1 with errno 0, and perror says `select: No error`.

  

  `strace` says `pselect: -1 = select(), errno 0` when running `strace fish -c 'status get-file config.fish'`.

- [ ] Should we add build instructions on cygwin/msys2 (experimental) to building section of README?

  `cargo install` only, should work after merging.

  ```bash
  rustup component add rust-src --toolchain nightly-x86_64-pc-windows-gnu
  cargo +nightly-x86_64-pc-windows-gnu install -Z build-std --target x86_64-pc-cygwin --git 
  https://github.com/fish-shell/fish-shell.git --bin fish
  ```

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,7 +1,5 @@
 # normalize newlines
-* text=auto
-*.fish text
-*.bat eol=crlf
+* text=auto eol=lf
 
 # let git show off diff hunk headers, help git diff -L:
 # https://git-scm.com/docs/gitattributes
--- a/src/path.rs
+++ b/src/path.rs
@@ -667,7 +667,7 @@ fn create_dir_all_with_mode<P: AsRef<std::path::Path>>(path: P, mode: u32) -> st
 /// Return whether the given path is on a remote filesystem.
 pub fn path_remoteness(path: &wstr) -> DirRemoteness {
     let narrow = wcs2zstring(path);
-    #[cfg(target_os = "linux")]
+    #[cfg(any(target_os = "linux", cygwin))]
     {
         let mut buf = MaybeUninit::uninit();
         if unsafe { libc::statfs(narrow.as_ptr(), buf.as_mut_ptr()) } < 0 {
@@ -701,7 +701,7 @@ pub fn path_remoteness(path: &wstr) -> DirRemoteness {
             }
         }
     }
-    #[cfg(not(target_os = "linux"))]
+    #[cfg(not(any(target_os = "linux", cygwin)))]
     {
         fn remoteness_via_statfs<StatFS, Flags>(
             statfn: unsafe extern "C" fn(*const i8, *mut StatFS) -> libc::c_int,
--- a/src/fd_monitor.rs
+++ b/src/fd_monitor.rs
@@ -389,7 +389,9 @@ impl BackgroundFdMonitor {
             drop(data);
             let ret =
                 fds.check_readable(timeout.map(Timeout::Duration).unwrap_or(Timeout::Forever));
-            if ret < 0 && !matches!(errno().0, libc::EINTR | libc::EBADF) {
+            // Cygwin reports ret < 0 && errno == 0 as success.
+            let err = errno().0;
+            if ret < 0 && !matches!(err, libc::EINTR | libc::EBADF) && !(cfg!(cygwin) && err == 0) {
                 // Surprising error
                 perror("select");
             }


.gitattributes
```diff
@@ -1,6 +1,6 @@
 # normalize newlines
 * text=auto
-*.fish text
+*.fish eol=lf
```
krobelus: okay, it looks like switching this to eol=crlf
will break fish

	$ target/debug/fish
	~/git/fish-shell-tmp/share/functions/fish_command_not_found.fish (line 1): Unknown command: \r
	string match -r '^ID(?:_LIKE)?\s*=.*' < /etc/os-release | \
	                                                          ^
	in command substitution
		called on line 10 of file ~/git/fish-shell-tmp/share/functions/fish_command_not_found.fish
	from sourcing file ~/git/fish-shell-tmp/share/functions/fish_command_not_found.fish
		called on line 1 of file ~/git/fish-shell-tmp/share/functions/fish_config.fish
	in function 'fish_config' with arguments 'theme choose fish\ default'
		called on line 20 of file ~/.config/fish/config.fish

`eol=lf` might be the right fix, but wouldn't it break
rendering in Windows editors like `notepad.exe`?

Also, is this not a problem for other files?
We got a few other extensions:

	status list-files | path extension | sort -u
	.1
	.css
	.html
	.js
	.py
	.theme

So do all of these work (without glitches such as extra newlines)

	fish_update_completions
	fish_config
	fish_config theme choose 'fish default'
	abbr -h

Looks like at least python can deal with CRLF just fine

	$ printf 'print ("foo\\\\\r\n bar")\r\n'
	print ("foo\
	 bar")
	$ printf 'print ("foo\\\\\r\n bar")\r\n' | python
	foo bar

but shells cannot

	$ printf 'echo foo\\\\\r\n bar\r\n' | bash
	foo
	bash: line 2: $'bar\r': command not found
	$ printf 'echo foo\\\\\r\n bar\r\n' | fish
	foo
	fish: Unknown command: bar
	fish: 
	 bar
	 ^~^

If this is just about line continuations (`\\\n` vs `\\\r\n`), we might want to support that.

src/path.rs
```diff
@@ -667,7 +667,7 @@ fn create_dir_all_with_mode<P: AsRef<std::path::Path>>(path: P, mode: u32) -> st
 /// Return whether the given path is on a remote filesystem.
 pub fn path_remoteness(path: &wstr) -> DirRemoteness {
     let narrow = wcs2zstring(path);
-    #[cfg(target_os = "linux")]
+    #[cfg(any(target_os = "linux", cygwin))]
```
krobelus, line 670: For the future: there are probably a bunch of other places where we
could use the same code paths for both linux and cygwin, rather than
using the BSD-style code for Cygwin.

In this case the BSD-style code path would actually be better,
but I suppose ST_LOCAL/MNT_LOCAL don't exist on Cygwin;
they don't on Linux according to

	rg -w -e ST_LOCAL -e MNT_LOCAL /usr/include/

krobelus:
>   `*.fish` works, but there is **`share/__fish_build_paths.fish.in`**

yeah I suppose we should fix those the same way. Should be easy to test.

> 
> - [ ] It&#39;s still possible that `poll` returns -1 with errno 0, and perror says `select: No error`.
> 
>   

ok then let's add back the workaround as of commit f7a5766cafdb09bd473c09a04b6d417f1c6ab939
(https://github.com/fish-shell/fish-shell/pull/11238/commits/f7a5766cafdb09bd473c09a04b6d417f1c6ab939)
but please add a link to a Cygwin bug report, so we don't have to keep it forever.

> - [ ] Should we add build instructions on cygwin/msys2 (experimental) to building section of README?
> 
>   `cargo install` only, should work after merging.
> 
>   ```bash
>   rustup component add rust-src --toolchain nightly-x86_64-pc-windows-gnu
>   cargo +nightly-x86_64-pc-windows-gnu install -Z build-std --target x86_64-pc-cygwin --git 
>   https://github.com/fish-shell/fish-shell.git --bin fish
>   ```

yes, our README.rst needs updating, maybe also something like this (not sure if that's still accurate)

	diff --git a/README.rst b/README.rst
	index b75341fea1..43ee7aaf75 100644
	--- a/README.rst
	+++ b/README.rst
	@@ -66,7 +66,8 @@
	    for Linux with the instructions for the appropriate distribution
	    listed above under “Packages for Linux”, or from source with the
	    instructions below.
	--  fish (4.0 on and onwards) cannot be installed in Cygwin, due to a lack of Rust support.
	+-  Fish can also be installed on all versions of Windows using
	+   `Cygwin <https://cygwin.com/>`__ (from the **Shells** category).
	 
	 Building from source
	 ~~~~~~~~~~~~~~~~~~~~

Ideally we can also give a recipe on how to build with CMake;
the main problems remaining with embedded builds are that they
- break some plugins that access `$__fish_data_dir`
- don't install man pages
- translations dont' work yet
But I guess most users will be fine with an embedded build.

I suppose we could also backport the Cygwin fixes to the 4.0.3 branch,
if someone wants to do that work and if it helps packagers.

.gitattributes
A2uria: > eol=lf might be the right fix, but wouldn't it break rendering in Windows editors like notepad.exe?

Modern editors support LF as the line ending.

> Also, is this not a problem for other files?

Yes, `status list-files | awk -F . '{ print $NF }' | sort | uniq` gives the following extensions. Most formats theoretically support `\r`, except to those related to fish.

- [ ] fish
- [ ] in (.fish.in)
- [ ] theme
- [x] css
- [x] html
- [x] js
- [x] png
- [x] py
- [ ] tmac

We may have problem with `.theme`, as it's kind of configuration file read by `fish_config.fish`, and `\r` might break string split as it's treated as a character instead of a whitespace.

Anyway, after further digestion I found that `* text=auto` completely breaks line endings. It defaults to `\r\n` on windows and won't respect `core.autocrlf=false`. `core.eol=lf` works, but it's impossible to set it in cargo. So should we remove it but keep `*.fish eol=lf` to normalize newlines. `.fish.in` and `.theme` is rarely modified so just handle them manually.

In msys2, we're using tarballs so we won't worry about '\r'. But `core.autocrlf=false` is required when build from git.

p.s. We don't have `.bat` files in worktree so without them we only need `*.fish eol=lf`.

```diff
 # normalize newlines
-* text=auto
-*.fish text
-*.bat eol=crlf
+*.fish eof=lf
```

> If this is just about line continuations (`\\\n` vs `\\\r\n`), we might want to support that.

```
$ cat example.sh
echo foo && \
echo bar
$ xxd example.sh
00000000: 6563 686f 2066 6f6f 2026 2620 5c0d 0a65  echo foo && \..e
00000010: 6368 6f20 6261 720d 0a                   cho bar..
$ bash example.sh
foo
bar
$ fish example.sh
foo
fish: Unknown command: \r
example.sh (line 1):
echo foo && \
            ^
bar
```

Maybe. The most obvious problem is that when `\r` appears after `\\`, fish will treat it as a command. `\\\r` cannot be escaped, so we got `\r`. However I think that we shouldn't support `\r` in script to keep consistency with linux.

src/path.rs
A2uria: Cygwin don't have `f_flag`.

krobelus:
> > eol=lf might be the right fix, but wouldn't it break rendering in Windows editors like notepad.exe?
> 
> Modern editors support LF as the line ending.

ok sounds good then

> Anyway, after further digestion I found that `* text=auto` completely breaks line endings. It defaults to `\r\n` on windows and won't respect `core.autocrlf=false`.

Wow that's strange.

> `core.eol=lf` works, but it's impossible to set it in cargo. So should we remove it but keep `*.fish eol=lf` to normalize newlines. `.fish.in` and `.theme` is rarely modified so just handle them manually.

I think *.theme files should get the same treatment as *.fish ("*.theme eol=lf").
We can probably ignore the rest.

> In msys2, we're using tarballs so we won't worry about '\r'. But `core.autocrlf=false` is required when build from git.
> 
> p.s. We don't have `.bat` files in worktree so without them we only need `*.fish eol=lf`.

yes, that's a leftover from vendored pcre sources

> 
> ```diff
>  # normalize newlines
> -* text=auto

> -*.fish text
> -*.bat eol=crlf
> +*.fish eof=lf

Can't we use "* text=auto eol=lf", to tell Git to automatically determine text files, and use lf for all files it deems text, both in the index as well as the working copy?

If I'm wrong, this looks okay too I guess, but it should be "eol" not "eof", and we'll probably want to add more patterns like *.sh eventually..

So this diff means the user no longer needs to set "core.autocrlf=false", because *.fish files aready get lf endings?

> we shouldn't support `\r` in script to keep consistency with linux.

ok

A2uria:
> Can't we use "* text=auto eol=lf", to tell Git to automatically determine text files, and use lf for all files it deems text, both in the index as well as the working copy?

`* text=auto eol=lf` works well with or without `core.autocrlf=false`.

> So this diff means the user no longer needs to set "core.autocrlf=false", because *.fish files aready get lf endings?

I think it is `core.autocrlf=input` as `\r\n` is converted to `\n` on commit. Since then we shouldn't worry about end of line anymore.

> ok then let's add back the workaround as of commit https://github.com/fish-shell/fish-shell/commit/f7a5766cafdb09bd473c09a04b6d417f1c6ab939
>
> but please add a link to a Cygwin bug report, so we don't have to keep it forever.

I created https://github.com/msys2/msys2-runtime/issues/308 as website of cygwin is temporary unavailable for me. Workaround will be added later.

p.s. Do we really need conditional compilation here?

```diff
-if ret < 0 && !matches!(errno().0, libc::EINTR | libc::EBADF) {
+// Cygwin select() may return -1 with errno 0
+if ret < 0 && !matches!(errno().0, libc::EINTR | libc::EBADF | 0) {
     // Surprising error
     perror("select");
 }
```

krobelus:
On Fri, Jul 25, 2025 at 04:15:30AM -0700, A2uria wrote:
> A2uria left a comment (fish-shell/fish-shell#11681)
> 
> > Can't we use "* text=auto eol=lf", to tell Git to automatically determine text files, and use lf for all files it deems text, both in the index as well as the working copy?
> 
> `* text=auto eol=lf` works well with or without `core.autocrlf=false`.

great, looks good

> I created https://github.com/msys2/msys2-runtime/issues/308 as website of cygwin is temporary unavailable for me. Workaround will be added later.

sounds good, let's mention that in the commit message or comment (even though the issue might move).
The next steps for that would be to extract a minimal reproducer (a small C program calling `pselect()`).

> p.s. Do we really need conditional compilation here?

I think `cfg!(cygwin)` would help because this check looks really
strange, so it's better to be very specific where to enable it.
Also if we use the workaround everywhere we might mask this bug in
other operating systems.  Better to find and fix the bug, else every
program has to work around this.

maybe

	if ret < 0
	        && !matches!(err, libc::EINTR | libc::EBADF)
	        && !(cfg!(target_os = "cygwin") && err == 0)

or

	if ret < 0 && match err {
		libc::EINTR | libc::EBADF => false,
		0 if cfg!(target_os = "cygwin") => false,
		_ => true,
	}

A2uria (commented):
Simply reused https://github.com/fish-shell/fish-shell/pull/11238/commits/f7a5766cafdb09bd473c09a04b6d417f1c6ab939, with `cfg!(cygwin)` used instead of `cfg!(target_os = "cygwin")` to avoid https://github.com/fish-shell/fish-shell/commit/2719ae443b8f908cc7e83d9475d8c8ddda8973fa.

Issue #11614, opened 2025-06-25
Using functions -c inside a function breaks funced (with $EDITOR)


Gawidev:
fish, version 4.0.2
Linux (Debian 12)
Wezterm
Replicated without 3rd-p customizations

Calling a function (hereon referred as *x*) which contains `function -c y z` on a function that was previously saved to disk (*y*) results in the initial comment looking like this '#Defined in /path/to/function.fish, copied in /path/to/*x*.fish'.
Then, with $EDITOR set, attempting to `funced z` results in the editor opening a tmp file with the contents of *x*.
This is because of line 75 in `funced.fish`:
```fish
    75      else if functions --details -- $funcname | string match --invert --quiet --regex '^(?:-|stdin)$'
```

esdmr:
After #9542, [`functions -Dv` returns](https://github.com/fish-shell/fish-shell/blob/30ff3710a069a03b21025cf896f0bbb31d9fe8bd/src/builtins/functions.rs#L204-L230)

> ```
> <path to file that executed `functions -c`>
> <path to original file>
> 0
> scope-shadowing
> ```

for copied functions, and `functions -D` returns the first line only.

We can either swap the first two lines so `functions -D` returns the original function, or make `funced` use `functions -Dv`'s second line. Swapping the lines might be a breaking change though, so perhaps the latter?

Edit: Now that I think about it, editing the original definition is not exactly right either. Perhaps, it should pretend it was defined interactively, or maybe show a prompt to the user.

krobelus:
> Calling a function (hereon referred as x) which contains function -c "y" "z" on a function that was previously saved to disk (y) results in the first-line comment looking like this:
> 
>  #Defined in /path/to/"y".fish, copied in /path/to/"x".fish
> Then, calling functions --details "z" incorrectly produces the path of the copying function (/path/to/"x".fish).

Can't reproduce:

	$ function x
	      functions -c @ z
	  end
	$ x
	$ functions -D z
	$HOME/.config/fish/functions/@.fish

esdmr:
@krobelus Looks like the copier function being defined interactively does not trigger it ([src/builtins/functions.rs](https://github.com/fish-shell/fish-shell/blob/f218cf2f387cd1dd2bb6f91aa73b7954fd1e55f5/src/builtins/functions.rs#L204-L230)). Either saving `x` to `~/.config/fish/functions`, or sourcing it should reproduce the bug.

```
$ echo 'function x
      functions -c @ z
[...]

PR #11691, opened 2025-07-29
funced: pretend copied functions are defined interactively

esdmr:
## Description

After #9542, the format for `functions -Dv` was changed for copied functions.

```diff
-stdin
-n/a
+[path to copy location]
+[path to original definition]
 [and a few more lines]
```

Some components were (and perhaps are) still expecting the old format, however. After a search, it looks like `funced` and `fish_config` are the only two functions using `functions -D` and `functions -Dv` in this repo (none are using `type -p`).

As noted in issue #11614, `funced` currently edits the file which copies the given copied function. Another option was to make `funced` edit the file which originally defined the function. Since the copied function would not have been updated either way, I modified `funced` so it would pretend that the copied function was defined interactively, like it was before.

(I did not find any preexisting test file for `funced`, so I did not write regressions test. Please tell me if I should.)

I did not modify `fish_config`, since it was only used for preset prompts in the web config, none of which used `functions --copy`. (Moreover, I believe it would have behaved correctly, since the preset would not have had to define the function, only copy it. (This grammar is making me dizzy.))

Fixes issue #11614

## TODOs:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -51,6 +51,7 @@ Interactive improvements
 - Instead of flashing all the text to the left of the cursor, fish now flashes the matched token during history token search, the completed token during completion (:issue:`11050`), the autosuggestion when deleting it, and the full command line in all other cases.
 - Pasted commands are now stripped of any ``$`` prefix.
 - The :kbd:`alt-s` binding will now also use ``run0`` if available.
+- ``funced`` will now edit copied functions directly, instead of the file where ``function --copy`` was invoked. (:issue:`11614`)
 
 New or improved bindings
 ^^^^^^^^^^^^^^^^^^^^^^^^
--- a/share/functions/funced.fish
+++ b/share/functions/funced.fish
@@ -71,6 +71,11 @@ function funced --description 'Edit function definition'
 
     if not functions -q -- $funcname
         echo $init >$tmpname
+    else if not string match --quiet --regex '^(?:n/a|not-autoloaded|autoloaded)$' -- (functions --details --verbose -- $funcname)[2]
+        # Pretend this copied function does not have a definition file. Editing the file which
+        # originally defined it would not update this copy, and the file which copied it
+        # would not have a function body to edit. (issue #11614)
+        functions -- $funcname >$tmpname
     else if functions --details -- $funcname | string match --invert --quiet --regex '^(?:-|stdin|embedded:.*)$'
         set writepath (functions --details -- $funcname)
         # Use cat here rather than cp to avoid copying permissions
--- a/tests/checks/funced.fish
+++ b/tests/checks/funced.fish
@@ -1,0 +1,23 @@
+#RUN: %fish %s
+
+function my-src
+    echo hello
+end
+
+echo "functions --copy my-src my-dst" >my-copy-function.fish
+source my-copy-function.fish
+rm my-copy-function.fish # Cleanup
+
+functions --details --verbose my-dst
+# CHECK: my-copy-function.fish
+# CHECK: {{.*}}tests/checks/funced.fish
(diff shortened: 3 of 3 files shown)


CHANGELOG.rst
```diff
@@ -51,6 +51,7 @@ Interactive improvements
 - Instead of flashing all the text to the left of the cursor, fish now flashes the matched token during history token search, the completed token during completion (:issue:`11050`), the autosuggestion when deleting it, and the full command line in all other cases.
 - Pasted commands are now stripped of any ``$`` prefix.
 - The :kbd:`alt-s` binding will now also use ``run0`` if available.
+- ``funced`` will now edit copied functions directly, instead of the file where it was copied. (:issue:`11614`)
```
krobelus: This is technically correct but maybe it's clearer to say
maybe « ... where "functions --copy" was invoked » ?
I'm fine eitehr way.

share/functions/funced.fish
```diff
@@ -71,6 +71,11 @@ function funced --description 'Edit function definition'
 
     if not functions -q -- $funcname
         echo $init >$tmpname
+    else if string match --invert --quiet --regex '^(?:n/a|not-autoloaded|autoloaded)$' -- (functions --details --verbose -- $funcname)[2]
```
krobelus: Probably `not string match` would be a little bit more idiomatic than `string match --invert`.

It's a shame that we don't have an easier / more robust way of figuring out that a function was copied from somewhere.

share/functions/funced.fish
```diff
@@ -71,6 +71,11 @@
...
+        # Pretend this copied function does not have a definition file. Editing the file which
+        # originally defined it would not update this copy, and the file which copied it
+        # would not have a function body to edit. (issue #11614)
+        functions -- $funcname >$tmpname
```
krobelus, line 78: Ok, after testing this, I think I get it. LGTM, thanks for fixing this.
Your PR description makes sense - maybe put it into the commit message?

Here's a quick test, run with

	cargo b && tests/test_driver.py target/debug tests/checks/funced.fish

(we could definitely add more tests)

```diff
diff --git a/tests/checks/funced.fish b/tests/checks/funced.fish
new file mode 100644
index 0000000000..a285635334
--- /dev/null
+++ b/tests/checks/funced.fish
@@ -0,0 +1,22 @@
+#RUN: %fish %s
+
+function my-src
+    echo hello
+end
+
+echo >my-copy-function.fish "functions --copy my-src my-dst"
+source my-copy-function.fish
+
+functions --details --verbose my-dst
+# CHECK: my-copy-function.fish
+# CHECK: {{.*}}tests/checks/funced.fish
+# CHECK: 3
+# CHECK: scope-shadowing
+
+EDITOR=cat funced my-dst
+# CHECK: # Defined in {{.*}}/tests/checks/funced.fish @ line 3, copied in my-copy-function.fish @ line 2
+# CHECK: function my-dst
+# CHECK:     echo hello
+# CHECK: end
+# CHECK: Editor exited but the function was not modified
+# CHECK: If the editor is still running, check if it waits for completion, maybe a '--wait' option?
```

krobelus (approved):

CHANGELOG.rst
esdmr: I have changed it to ```instead of the file where ``function --copy`` was invoked```.

share/functions/funced.fish
esdmr: The line later had that `--invert` so I had kept it. changed it to `not`.

share/functions/funced.fish
esdmr: Thank you. I have added the test.

I have also rebased the commit and updated the commit message.

PR #11704, opened 2025-08-04
Removed libc deps

krakotay:
## Description

**Remove C shim (`libc.c`) and reimplement in pure Rust (`libc.rs`).**

This commit completely removes the dependency on `src/libc.c`, replacing it with a pure Rust implementation in `src/libc.rs`
My primary motivation is to improve and simplify support for running fish shell on Windows 11 systems. I hope that this PR brings the project one step closer to robust and fully native Windows support, by eliminating the need for any custom C glue code.

### What changed:

* Completely removed `src/libc.c` and all related C build steps in `build.rs`.
* Rewrote `src/libc.rs` to provide all necessary platform-specific constants and functions natively in Rust, using the `libc` crate where needed.
* Added pure-Rust fallback code for filesystem case-sensitivity checks, replacing old C logic.
* Rewrote doc-comments and variable names for clarity and consistency.
* All changed files now rely exclusively on Rust and standard dependencies, no custom C code remains.

### Platforms tested:

* Ubuntu, MacOS (from github actions)

### Known caveats or risks:

* The new Rust implementations of e.g. RLIMIT\_\* constants and FS case-sensitivity checks aim to replicate the old C logic exactly, but it's possible that some obscure platforms might differ in edge-case behavior.
* If anyone maintains unusual downstream patches or platforms (BSDs, exotic Unixes, etc.), please review carefully.

---

## TODOs:

* [ ] Changes to fish usage are reflected in user documentation/manpages.
  *No user-facing changes; this is purely an internal/platform refactor.*
* [ ] Tests have been added for regressions fixed
  *No new regressions found; all current tests pass.*
* [ ] User-visible changes noted in CHANGELOG.rst
  *Not applicable, as no user-visible changes are expected.*


build.rs
```diff
@@ -84,7 +84,7 @@ fn main() {
 
     rsconf::rebuild_if_env_changed("FISH_GETTEXT_EXTRACTION_FILE");
 
-    cc::Build::new().file("src/libc.c").compile("flibc.a");
+    // cc::Build::new().file("src/libc.c").compile("flibc.a");
```
krobelus, line 87: > My primary motivation is to improve and simplify support for running fish shell on Windows 11 systems.

Not sure if this is easy because fish relies on lots of Unix APIs.
There are some closed issues that list potential problems.

But regardless, moving away from src/libc.c and into the libc crate
sounds like a good idea.

src/highlight/file_tester.rs
```diff
@@ -391,27 +393,83 @@ pub fn is_potential_cd_path(
...
+                v
             }
         }
     }
-    result
+}
+
+/// Fallback: create a file called "lower", check if it's visible as "UPPER".
+/// Returns true if the FS is case-insensitive.
+#[cfg(any(unix, target_os = "wasi"))]
```
krobelus, line 429: is this relevant for wasi platforms?
Do these even have file system access?
Can fish run on wasi?

src/libc.rs
```diff
@@ -1,56 +1,463 @@
...
+/// Misuse can cause undefined behavior.
+pub unsafe fn stdout_stream() -> *mut libc::FILE {
+    #[cfg(any(
+        target_os = "linux",
+        target_os = "android",
+        target_os = "freebsd",
+        target_os = "openbsd",
+        target_os = "netbsd",
+        target_os = "dragonfly"
+    ))]
```
krobelus, line 59: I'd much rather have these things go into the [libc](https://github.com/rust-lang/libc) crate first.
Explicitly listing all operating systems that define `stdout` as a global variable
seems suboptimal since it will exclude new operating systems; we at least want to share maintenance of this list with the rest of the Rust community by putting this into libc.

So the changes in this PR shoudl mostly boil down to "delete this piece of code from
src/libc.c and replace it with `libc::foo`".
Still you should split them into one commit per function (or group of functions),
so that we can do thorough review more easily as well as use `git bisect` to find out which commits broke something.
And every commit should be atomic, e.g. pass all tests.

Probably the contributions to libc should also be small commits or small PRs.

ridiculousfish:
This is exciting - I really hope we can drop the cc dependency - but yes, let's go methodically one by one.

ridiculousfish:
Update on this - I have the `libc.rs` and `libc.c` bits removed [here](https://github.com/ridiculousfish/fish-shell/tree/remove-cc) but there's remaining work to remove the `cc` dependency entirely from build.rs - ideally the other MRs about fixing up gettext will make that easier

danielrainer:
> ideally the other MRs about fixing up gettext will make that easier

The entire `have_gettext` stuff is removed there since we will no longer have a runtime dependency on gettext.

zippozeng:
i will fish shell work on Windows，please commit this pr

faho:
libc.c is no more, so this is superseded.

PR #11706, opened 2025-08-05
feat: add redaction for sensitive variables in completion suggestions

fixing-things-enjoyer:
## Description

This pull request introduces a new feature to prevent sensitive environment variable values from being displayed in completion suggestions.

A new global environment variable, fish_redact_vars, is introduced. It can be set to a list of glob patterns (e.g., *_TOKEN, *_SECRET*). Any variable name matching one of these patterns will have its value replaced with [redacted] in the completion description. This helps to avoid accidental exposure of secrets like API keys or tokens when using tab completion in a public setting.

For example:
`set -g fish_redact_vars '*_TOKEN' '*_SECRET*'`

## Screenshots
(I'm pressing the Tab key at the end of `MY_`)

This:
image

Becomes this:
image

This change also includes updates to the localization files to make the new `Variable: [redacted]` string available for translation by the community.

Fixes issue # (N/A)

## TODOs:

### DIFF
--- a/doc_src/interactive.rst
+++ b/doc_src/interactive.rst
@@ -61,9 +61,15 @@ You can also write your own completions or install some you got from someone els
     If a variable name matches one of these patterns, its value will be replaced with ``[redacted]`` in the completion description.
     This is useful for preventing accidental exposure of sensitive information like API keys or tokens during tab completion.
 
-    Example: to redact variables ending with ``_TOKEN`` or containing ``_SECRET``::
+    By default, fish uses the following patterns:
 
-        set -g fish_redact_vars '*_TOKEN' '*_SECRET*'
+    - ``*_API_KEY``
+    - ``*_TOKEN``
+    - ``*_SECRET``
+
+    Example: to also redact variables ending with ``_PASSWORD``::
+
+        set -ga fish_redact_vars '*_PASSWORD'
 
 Completion scripts are loaded on demand, like :ref:`functions are <syntax-function-autoloading>`. The difference is the ``$fish_complete_path`` :ref:`list <variables-lists>` is used instead of ``$fish_function_path``. Typically you can drop new completions in ~/.config/fish/completions/name-of-command.fish and fish will find them automatically.
 
--- a/share/functions/__fish_config_interactive.fish
+++ b/share/functions/__fish_config_interactive.fish
@@ -139,6 +139,9 @@ end" >$__fish_config_dir/config.fish
     # Load key bindings
     __fish_reload_key_bindings
 
+    # Set default redaction patterns for variables that likely contain secrets.
+    __init_uvar fish_redact_vars '*_API_KEY' '*_TOKEN' '*_SECRET'
+
     # Detect whether the terminal reflows on its own
     # If it does we shouldn't do it.
     # Allow $fish_handle_reflow to override it.
--- a/tests/pexpects/redact_vars.py
+++ b/tests/pexpects/redact_vars.py
@@ -21,35 +21,39 @@ send, sendline, sleep, expect_prompt, expect_re, expect_str = (
 expect_prompt()
 
 #
-# Test that `fish_redact_vars` correctly redacts variable values in pager completions.
+# Test that `fish_redact_vars` correctly redacts variable values in pager completions,
+# including default patterns and explicit variable names.
 #
 
-sendline("set -gx MY_SUPER_SECRET 'secret_value'")
+# Test variables for default redaction patterns
+sendline("set -gx MY_API_KEY 'api_key_value'")
 expect_prompt()
-sendline("set -gx MY_OTHER_SECRET 'another_secret'")
+sendline("set -gx MY_TOKEN 'token_value'")
 expect_prompt()
-sendline("set -gx MY_EXPLICIT_SECRET 'public_value'")
+sendline("set -gx MY_SECRET 'secret_value'")
 expect_prompt()
-sendline("set -gx MY_OTHER_VALUE 'other_value'")
+
+# Test variables for explicit redaction and non-redacted cases
+sendline("set -gx MY_EXPLICIT_REDACTION 'explicit_value'")
 expect_prompt()
 sendline("set -gx MY_NORMAL_VAR 'normal_value'")
 expect_prompt()
 
-sendline("set -g fish_redact_vars '*_SECRET' MY_EXPLICIT_SECRET")
+# Append explicit redaction to default set
+sendline("set -ga fish_redact_vars MY_EXPLICIT_REDACTION")
 expect_prompt()
 
-# Trigger variable completion to open the pager.
+# Trigger variable completion to open the pager
 send("echo $MY_\t")
 sleep(0.05)
 
-# Wait for a short timeout to allow the pager to draw and flush its
-# output to the pexpect buffer. We expect a TIMEOUT.
+# Wait for pager to draw and flush output
 try:
     sp.spawn.expect(pexpect.TIMEOUT, timeout=0.5)
 except pexpect.TIMEOUT:
     pass
 
-# Verify the pager output shows the correct redaction status for each variable.
+# Process pager output to verify redaction status
 stripped_output = re.sub(r"\x1b\].*?\x07|\x1b\[?[0-9;]*[A-Za-z]", "", sp.spawn.before)
 output_text = "\n".join(
     [
@@ -59,8 +63,9 @@ output_text = "\n".join(
     ]
 )
 
-assert "$MY_SUPER_SECRET (Variable: [redacted])" in output_text
-assert "$MY_OTHER_SECRET (Variable: [redacted])" in output_text
-assert "$MY_EXPLICIT_SECRET (Variable: [redacted])" in output_text
-assert "$MY_OTHER_VALUE (Variable: other_value)" in output_text
+# Verify expected redactions
+assert "$MY_API_KEY (Variable: [redacted])" in output_text
+assert "$MY_TOKEN (Variable: [redacted])" in output_text
+assert "$MY_SECRET (Variable: [redacted])" in output_text
+assert "$MY_EXPLICIT_REDACTION (Variable: [redacted])" in output_text
 assert "$MY_NORMAL_VAR (Variable: normal_value)" in output_text


po/fr.po
```diff
@@ -1819,6 +1819,10 @@ msgstr "Signal défini par l'utilisateur 2"
 msgid "Variable: %ls"
 msgstr "Variable : %ls"
 
+#
+msgid "Variable: [redacted]"
+msgstr ""
```
ccoVeille, line 1824: One suggestion for French


```suggestion
msgstr "masqué"
```

fixing-things-enjoyer: made that change, thanks!

doc_src/interactive.rst
```diff
@@ -56,6 +56,15 @@ It also provides a large number of program specific scripted completions. Most o
 
 You can also write your own completions or install some you got from someone else. For that, see :ref:`Writing your own completions <completion-own>`.
 
+.. envvar:: fish_redact_vars
+    A list of glob patterns for variable names whose values should be redacted in completion descriptions.
+    If a variable name matches one of these patterns, its value will be replaced with ``[redacted]`` in the completion description.
+    This is useful for preventing accidental exposure of sensitive information like API keys or tokens during tab completion.
+
+    Example: to redact variables ending with ``_TOKEN`` or containing ``SECRET``::
```
krobelus: should be `_SECRET` (added an underscore), no?

doc_src/interactive.rst
```diff
@@ -56,6 +56,15 @@
...
+
+        set -g fish_redact_vars '*_TOKEN' '*_SECRET*'
```
krobelus: I'd personally recommend against using environment variables to store secrets,
but I guess some legacy tools require it;
seems like a handy feature I guess.

If there are some commonly used variables out there (are there?),
we can maybe ship a reasonable default value (as a global variable, set for interactive sessions, see __fish_config_interactive).

src/complete.rs
```diff
@@ -1682,8 +1682,23 @@ impl<'ctx> Completer<'ctx> {
...
 
-                    let value = expand_escape_variable(&var);
-                    desc = sprintf!(*COMPLETE_VAR_DESC_VAL, value);
+                    let redacted = if let Some(redact_patterns) =
+                        self.ctx.vars().get(L!("fish_redact_vars"))
+                    {
+                        redact_patterns.as_list().iter().any(|pattern| {
+                            let wc = crate::parse_util::parse_util_unescape_wildcards(pattern);
+                            wildcard_match(&env_name, wc, false)
+                        })
```
krobelus: I suppose we could use something like (untested)

	self.ctx.vars().get(L!("fish_redact_vars")).map_or(&[], |v|v.as_list())...

to get rid of the `else false` branch, since, by definition,
we treat an undefined variable same as the empty list here.

src/complete.rs
```diff
@@ -1682,8 +1682,23 @@
...
+                    } else {
+                        false
+                    };
+
+                    if redacted {
+                        desc = wgettext!("Variable: [redacted]").to_owned();
```
krobelus: we should be consistent with COMPLETE_VAR_DESC_VAL,
so either define COMPLETE_VAR_REDACTED_DESC_VAL,
or inline COMPLETE_VAR_DESC_VAL 
(if we do that, we should make sure that Lazy doesn't add any caching here,
and exists only because `wgettext!()` is not const.. I'm pretty sure that's the case)

tests/pexpects/redact_vars.py
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env python3
+import pexpect
+import re
+import os
+from pexpect_helper import SpawnedProc
+
+# Set TERM in the environment before spawning fish. This is more robust
+# and ensures the shell initializes with the correct terminal type.
+env = os.environ.copy()
+env["TERM"] = "xterm"
```
krobelus, line 9: the comment sounds like an insecure LLM, maybe delete it.
The actual reason is that otherwise (with the default TERM=dumb), no completions will be displayed.

tests/pexpects/redact_vars.py
```diff
@@ -0,0 +1,92 @@
...
+# Start with a clean prompt.
+expect_prompt()
+
+#
+# Test that `fish_redact_vars` correctly redacts variable values in pager completions.
+#
+
+# Reset key bindings to ensure a clean test environment.
+sendline("bind --erase --all")
+expect_prompt()
```
krobelus: shouldn't need this, the environment is already clean

tests/pexpects/redact_vars.py
```diff
@@ -0,0 +1,92 @@
...
+
+# Set a low pager threshold to force it to appear for our list of completions.
+sendline("set -g fish_pager_completion_threshold 1")
+expect_prompt()
```
krobelus, line 30: I don't think this variable exists

tests/pexpects/redact_vars.py
```diff
@@ -0,0 +1,92 @@
...
+
+# This one should remain visible.
+sendline("set -gx MY_OTHER_VALUE 'other_value'")
+expect_prompt()
+
+sendline("set -gx MY_NORMAL_VAR 'normal_value'")
+expect_prompt()
+
+# Configure the redaction rules: a wildcard for `*_SECRET` and one explicit name.
+sendline("set -g fish_redact_vars '*_SECRET' MY_PUBLIC_VALUE")
```
krobelus: why is MY_PUBLIC_VALUE chosen as name for something we want to hide from the public

tests/pexpects/redact_vars.py
```diff
@@ -0,0 +1,92 @@
...
+send("echo $MY_\t")
+sleep(0.05)
+
+# The pager display and subsequent command-line redraw can create a race condition.
+# A short, non-deterministic wait allows all output to flush to the pexpect buffer.
+# We then grab the entire buffer post-facto for verification.
+try:
+    sp.spawn.expect(pexpect.TIMEOUT, timeout=0.5)
+except pexpect.TIMEOUT:
+    pass
```
krobelus, line 54: there's probably a better way to do this, buit I guess.
Would be nice if the comment was a bit less hand-wavy;
overall it would be fine to reduce comments a bit.

tests/pexpects/redact_vars.py
```diff
@@ -0,0 +1,92 @@
...
+output_text = "\n".join([re.sub(r'\s+', ' ', line.replace('\t', ' ').strip()) for line in stripped_output.splitlines() if line.strip() and '(Variable:' in line])
+
+# Assert that each expected line exists somewhere in the captured output.
+assert "$MY_SUPER_SECRET (Variable: [redacted])" in output_text
+assert "$MY_OTHER_SECRET (Variable: [redacted])" in output_text
+assert "$MY_PUBLIC_VALUE (Variable: [redacted])" in output_text
+assert "$MY_OTHER_VALUE (Variable: other_value)" in output_text
+assert "$MY_NORMAL_VAR (Variable: normal_value)" in output_text
+
+# Clean up after the test.
```
krobelus: no need to clean up, each test gets its own PTY and fish process

doc_src/interactive.rst
fixing-things-enjoyer: Yep! updated.

doc_src/interactive.rst
fixing-things-enjoyer: I'm not too sure here as to a good set of defaults. My searching for a way to do this definitely made me feel like I'm well outside of standard practice, so this seemed like something one would look for and be intentional about opting into it (versus expecting it to be done by default on a certain common set of variable names).

I know [there's at least two of us not being the best with our secrets management](https://stackoverflow.com/questions/71582605/disable-fish-shell-showing-env-variable-value-on-tab) in the last ~4 years. Right now my local dev environment is a mess of LLM Provider API keys. which made me go for this change.

src/complete.rs
fixing-things-enjoyer: refactored this accordingly and then went with clippy's suggestion.

src/complete.rs
fixing-things-enjoyer: Went with existing pattern of defining a `COMPLETE_VAR_REDACTED_DESC_VAL`

tests/pexpects/redact_vars.py
fixing-things-enjoyer: got rid of it.

tests/pexpects/redact_vars.py
fixing-things-enjoyer: got rid of it.

tests/pexpects/redact_vars.py
fixing-things-enjoyer: got rid of it.

tests/pexpects/redact_vars.py
fixing-things-enjoyer: good question, swapped it to something more coherent with test:
`MY_EXPLICIT_SECRET` i.e. the non-wildcard based one we expect to be redacted.

tests/pexpects/redact_vars.py
fixing-things-enjoyer: got rid of it.

tests/pexpects/redact_vars.py
fixing-things-enjoyer: heavily reduced the comments overall, reworded this one.

doc_src/interactive.rst
krobelus: I wonder if we can find a reasonable default value.
For example github's official CLI uses `GITHUB_TOKEN` (and fall back to `~/.config/gh/hosts.yml`)
1. Would it hurt anyone if we use `*_TOKEN`? I.e. are there other variables matching `*_TOKEN` that we do want to redact?
2. is "hiding `GITHUB_TOKEN`" a reasonable default behavior? I think yes, though there is a minor argument to be made that showing a short prefix might help.

What's your use case? Screen sharing, presentations, live streaming, people peeking over your shoulder?
I have something similar for autosuggestions;
normally I use defaults but I have a keybinding to run `set fish_history ""`, which (like fish's private mode) disables autosuggestions from history.
Maybe we want to have a "presentation mode" that does both

krobelus: I think I mainly want to understand where to find tools that can use or even require secrets in environment variables.
Ideally popular tools. Then we can tailor our default to at least match those

fixing-things-enjoyer: My use case is indeed screensharing. I've been juggling as many LLM CLI tools as I can get my hands on lately - [crush](https://github.com/charmbracelet/crush?tab=readme-ov-file#getting-started), [goose](https://block.github.io/goose/docs/guides/environment-variables/), [gemini-cli](https://github.com/google-gemini/gemini-cli/blob/4828e4daf198a675ce118cec08dcfbd0bfbb28a6/docs/cli/configuration.md#environment-variables--env-files), [openhands-cli](https://docs.all-hands.dev/usage/how-to/cli-mode#running-with-docker), [qwen-code](https://github.com/QwenLM/qwen-code/blob/93f5e59710ca192ef4ab739686cd26bc5898190d/docs/cli/configuration.md#environment-variables--env-files), [claude-code](https://docs.anthropic.com/en/docs/claude-code/settings#environment-variables)

Popularity ebbs and flows on these, most of them have mechanisms aside from environment variables to specify your LLM provider API Keys, but i find convenience in that they'll often use the same environment variable names, so like `GEMINI_API_KEY, OPENAI_API_KEY, OPENROUTER_API_KEY, etc...`. If I were choosing defaults, I wouldn't be mad at `*_API_KEY`, `*_TOKEN`, `*_SECRET`. 

That idea of a more comprehensive presentation mode sounds better than what I did here. Although I guess users could combine the existing defacto "private mode" and this redaction mechanism via their own custom functions.

krobelus:
> If I were choosing defaults, I wouldn't be mad at `*_API_KEY`,
> `*_TOKEN`, `*_SECRET`.

Sounds good to me, let's do that.

fixing-things-enjoyer:
> If I were choosing defaults, I wouldn't be mad at `*_API_KEY`, `*_TOKEN`, `*_SECRET`.
> Sounds good to me, let's do that.

added those defaults, adjusted docs to mention them, and tests to ensure they're present

doc_src/interactive.rst
```diff
@@ -56,6 +56,21 @@ It also provides a large number of program specific scripted completions. Most o
 
 You can also write your own completions or install some you got from someone else. For that, see :ref:`Writing your own completions <completion-own>`.
 
+.. envvar:: fish_redact_vars
```
krobelus, line 59: avoid abbreviations in globally-visible names, so "vars" should probably be "variables".

doc_src/interactive.rst
```diff
@@ -56,6 +56,21 @@ It also provides a large number of program specific scripted completions. Most o
...
+    If a variable name matches one of these patterns, its value will be replaced with ``[redacted]`` in the completion description.
+    This is useful for preventing accidental exposure of sensitive information like API keys or tokens during tab completion.
+
+    By default, fish uses the following patterns:
+
+    - ``*_API_KEY``
+    - ``*_TOKEN``
+    - ``*_SECRET``
+
+    Example: to also redact variables ending with ``_PASSWORD``::
```
krobelus, line 70: Hopefully most users won't ever need to configure this,
so there's no need to feature it this early/prominently in documentation.
With my other suggestion, we would naturally create a dedicated documentation page at `doc_src/cmds/fish_completion_redacted_variables.rst`,
that would work well.

doc_src/interactive.rst
```diff
@@ -56,6 +56,21 @@ It also provides a large number of program specific scripted completions. Most o
...
+
+        set -ga fish_redact_vars '*_PASSWORD'
```
krobelus, line 72: (this example does not work as described if `fish_redact_vars` is a universal variable)

share/functions/__fish_config_interactive.fish
```diff
@@ -139,6 +139,9 @@ end" >$__fish_config_dir/config.fish
     # Load key bindings
     __fish_reload_key_bindings
 
+    # Set default redaction patterns for variables that likely contain secrets.
+    __init_uvar fish_redact_vars '*_API_KEY' '*_TOKEN' '*_SECRET'
```
krobelus, line 143: we're not gonna add more universal variables
(we'll replace them with something less surprising in future)

also, I'm not sure if we really want this to be a (global) variable.

So unless we make it non-configurable for now, consider this:

Add an overridable function that is called whenever we get new completions.
The default definition would be

	function fish_completion_redacted_variables
		string match -r -- '.*_(API_KEY|SECRET|TOKEN)' $argv
	end

it would output a list of variables whose values should be redacted when shown in the completion pager,
and possibly in others.

---

Do we plan on using this for anything other than completion descriptions?
If yes, the name should maybe not "completion".
I can't come up with a use case though.
I don't think it's a good idea to try to redact autosuggestions like 

	set MY_SECRET hardcoded-secret

because there are better alternatives (that don't add this to history).

---

The upside of function is that it's introspectable.
Downside is that it allows shooting yourself in the foot.
But we allow the same in `fish_prompt`, `fish_title`, `fish_should_add_to_history` etc., so I don't expect much extra headache from that

AFAICT, function is better, so I'd be happy to go with that

tests/pexpects/redact_vars.py
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/env python3
+import os
+import re
+
+import pexpect
+from pexpect_helper import SpawnedProc
+
+env = os.environ.copy()
+env["TERM"] = "xterm"
```
krobelus, line 9: I know this pattern is used a lot elsewhere, but we actually don't care about the exact value of TERM,
only that it's not "dumb" (though we'll eventually also remove support for that, most likely)

Can you change this to `"not-dumb"` to make this more obvious.

tests/pexpects/redact_vars.py
```diff
@@ -0,0 +1,71 @@
...
+sendline("set -gx MY_NORMAL_VAR 'normal_value'")
+expect_prompt()
+
+# Append explicit redaction to default set
+sendline("set -ga fish_redact_vars MY_EXPLICIT_REDACTION")
+expect_prompt()
+
+# Trigger variable completion to open the pager
+send("echo $MY_\t")
+sleep(0.05)
```
krobelus, line 48: I don't understand what this timeout is for

tests/pexpects/redact_vars.py
```diff
@@ -0,0 +1,71 @@
...
+
+# Wait for pager to draw and flush output
+try:
+    sp.spawn.expect(pexpect.TIMEOUT, timeout=0.5)
+except pexpect.TIMEOUT:
+    pass
```
krobelus, line 54: this block seems like a convoluted way of saying `sleep(0.5)` ?

tests/pexpects/redact_vars.py
```diff
@@ -0,0 +1,71 @@
...
+
+# Process pager output to verify redaction status
+stripped_output = re.sub(r"\x1b\].*?\x07|\x1b\[?[0-9;]*[A-Za-z]", "", sp.spawn.before)
```
krobelus, line 57: This statement strips OSC and CSI commands from the output.
But the CSI regex is incomplete, and we should not duplicate this sort of functionality.
We already have something like this; it's called `SANITIZE_FOR_PRINTING_RE` in `tests/pexpect_helper.py`,
which you should be able to import.

tests/pexpects/redact_vars.py
```diff
@@ -0,0 +1,71 @@
...
+output_text = "\n".join(
+    [
+        re.sub(r"\s+", " ", line.replace("\t", " ").strip())
```
krobelus, line 60: the `line.replace("\t", " ")` part seems pointless  given that the outer regex will replace any whitespace already.

tests/pexpects/redact_vars.py
```diff
@@ -0,0 +1,71 @@
...
+        for line in stripped_output.splitlines()
+        if line.strip() and "(Variable:" in line
```
krobelus, line 62: `line.strip()` is redundant

tests/pexpects/redact_vars.py
```diff
@@ -0,0 +1,71 @@
...
+    ]
+)
+
+# Verify expected redactions
+assert "$MY_API_KEY (Variable: [redacted])" in output_text
```
krobelus, line 67: these assertions are not ideal because they don't print the actual fish output on failure.
But it's fine to ignore that for now.

Issue #11095, opened 2025-01-26
Autocompletion for git alias hangs and throws "Too much data emitted by command substitution"


Franz1007orRimzet:
## System info:
OS: Arch Linux
fish: 4.0b1-342-gc651a79cb
Terminal: ghostty 1.0.1

## Summary
When I tried to invoke path autocompletion for alias of '/usr/bin/git --git-dir="$HOME/.dotfiles/" --work-tree="$HOME"', fish became unresponsive and then after a few seconds threw: 
`/usr/share/fish/completions/git.fish (line 194): Too much data emitted by command substitution so it was discarded
        set -l stats (__fish_git $git_opt status --porcelain=2 $status_opt)
                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
in function '__fish_git_files' with arguments 'modified untracked deleted unmerged modified-staged-deleted'
in command substitution`, full command was: `/usr/bin/git --git-dir="$HOME/.dotfiles/" --work-tree="$HOME" add ~/.config/fi<Tab>`

Result is the same regardless if I use alias or full command

I followed instructions from Arch Wiki to setup bare repository to manage dotfiles. The method consists in creating bare repository in $HOME/.dotfiles, creating alias dotfiles='/usr/bin/git --git-dir="$HOME/.dotfiles/" --work-tree="$HOME"', and using it to keep track of changes in dotfiles. Also git option status.showUntrackedFiles  is set to no, to make it usable. I think the issue is that fish autocompletion tries to iterate over all files in repo, but because it is so big, it becomes unresponsive, but I'm not sure for I haven't read the code and I don't feel competent enough to do so.

faho:
Our guidance when it comes to making your entire home directory a git repository is always: Don't do that. It makes git perform terribly, and requires a bunch of workarounds.

I've made it so the git completions will fall back to regular file completions when asked to show untracked files in a repo that has untracked files disabled, which should help in your case.

This isn't ideal, but everything else I can think of would make performance markedly worse in repositories of a sensible size (even quite big ones).

krobelus:
you can fix slowness by adding something like `/*` to your `.gitignore`, and override it with `git add -f` if needed

Franz1007orRimzet:
I think I will just wrap aliased command in bash script and manually add certain git completions. Any ways, thank you for the help.

faho:
[...]

PR #11709, opened 2025-08-06
Respect Git canonical config for untracked files

soraxas:
## Description

The git completion was not respecting git config for ignoring untracked files.

I am using something called `yadm` to track my dotfiles in a bare git repository. What `yadm` is doesn't really matter, as it just setup GIT environment variable and configs under hood. I'll demonstrates the same principle with raw git:

This shows that it's showing my dotfiles status
Screenshot 2025-08-06 104720

By default it setup the git config variable to NOT show untracked files. This in-turns creates issues when <TAB> completing `add`: 
Screenshot 2025-08-06 104805
(notice the > 300K files)

Now, you can see that the git config is setup correctly:
Screenshot 2025-08-06 105002

where its using the option here: https://git-scm.com/docs/git-config
image

(BTW, noticing how the current behaviour is actually the `all` behaviour of including files within untracked folders)

-----------------------

The behaviour after this PR:

Screenshot 2025-08-06 105848

-------------

## Discussion

Regarding the previous `bash.showUntrackedFiles` option: I was able to find traces of it in the git source repo, and apparently I think it used to be a contrib module for completion:

https://github.com/search?q=repo%3Agit%2Fgit%20bash.showUntrackedFiles&type=code

But i reckon we should now use the canonical git config option?

https://github.com/search?q=repo%3Agit%2Fgit+status.showUntrackedFiles&type=code

## TODOs:

### DIFF
--- a/doc_src/cmds/fish_git_prompt.rst
+++ b/doc_src/cmds/fish_git_prompt.rst
@@ -33,11 +33,11 @@ Boolean options (those which enable or disable something) understand "1", "yes" 
 - ``$__fish_git_prompt_show_informative_status`` or the git option ``bash.showInformativeStatus`` can be set to 1, true or yes to enable the "informative" display, which will show a large amount of information - the number of dirty files, unpushed/unpulled commits, and more.
   In large repositories, this can take a lot of time, so you may wish to disable it in these repositories with  ``git config --local bash.showInformativeStatus false``. It also changes the characters the prompt uses to less plain ones (``✚`` instead of ``*`` for the dirty state for example) , and if you are only interested in that, set ``$__fish_git_prompt_use_informative_chars`` instead.
 
-  Because counting untracked files requires a lot of time, the number of untracked files is only shown if enabled via ``$__fish_git_prompt_showuntrackedfiles`` or the git option ``bash.showUntrackedFiles``.
+  Because counting untracked files requires a lot of time, the number of untracked files is only shown if enabled via ``$__fish_git_prompt_showuntrackedfiles`` or the git option ``status.showUntrackedFiles``.
 
 - ``$__fish_git_prompt_showdirtystate`` or the git option ``bash.showDirtyState`` can be set to 1, true or yes to show if the repository is "dirty", i.e. has uncommitted changes.
 
-- ``$__fish_git_prompt_showuntrackedfiles`` or the git option ``bash.showUntrackedFiles`` can be set to 1, true or yes to show if the repository has untracked files (that aren't ignored).
+- ``$__fish_git_prompt_showuntrackedfiles`` or the git option ``status.showUntrackedFiles`` can be set to 1, true or yes to show if the repository has untracked files (that aren't ignored).
 
 - ``$__fish_git_prompt_showupstream`` can be set to a list of values to determine how changes between HEAD and upstream are shown:
 
--- a/share/completions/git.fish
+++ b/share/completions/git.fish
@@ -1170,10 +1170,10 @@ complete -c git -n '__fish_git_using_command add' -l ignore-errors -d 'Ignore er
 complete -c git -n '__fish_git_using_command add' -l ignore-missing -d 'Check if any of the given files would be ignored'
 # Renames also show up as untracked + deleted, and to get git to show it as a rename _both_ need to be added.
 # However, we can't do that as it is two tokens, so we don't need renamed here.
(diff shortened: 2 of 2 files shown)


doc_src/cmds/fish_git_prompt.rst
```diff
@@ -33,11 +33,11 @@ Boolean options (those which enable or disable something) understand "1", "yes"
 - ``$__fish_git_prompt_show_informative_status`` or the git option ``bash.showInformativeStatus`` can be set to 1, true or yes to enable the "informative" display, which will show a large amount of information - the number of dirty files, unpushed/unpulled commits, and more.
   In large repositories, this can take a lot of time, so you may wish to disable it in these repositories with  ``git config --local bash.showInformativeStatus false``. It also changes the characters the prompt uses to less plain ones (``✚`` instead of ``*`` for the dirty state for example) , and if you are only interested in that, set ``$__fish_git_prompt_use_informative_chars`` instead.
 
-  Because counting untracked files requires a lot of time, the number of untracked files is only shown if enabled via ``$__fish_git_prompt_showuntrackedfiles`` or the git option ``bash.showUntrackedFiles``.
+  Because counting untracked files requires a lot of time, the number of untracked files is only shown if enabled via ``$__fish_git_prompt_showuntrackedfiles`` or the git option ``status.showUntrackedFiles``.
```
krobelus: I don't think we have much reason to change the git prompt; the upstream version from Git also only supports `bash.showUntrackedFiles`.
(also because one may want to configure the prompt independent of `git status` behavior)
I guess we could fall back to status.showUntrackedFiles here, if bash.showUntrackedFiles is not defined.. but I'd rather change it upstream first, if at all.
If upstream changes this, then we should definitely follow.

share/completions/git.fish
```diff
@@ -1170,10 +1170,10 @@ complete -c git -n '__fish_git_using_command add' -l ignore-errors -d 'Ignore er
...
 # Renames also show up as untracked + deleted, and to get git to show it as a rename _both_ need to be added.
 # However, we can't do that as it is two tokens, so we don't need renamed here.
-complete -f -c git -n '__fish_git_using_command add; and test "$(__fish_git config --get bash.showUntrackedFiles)" != 0' -a '(__fish_git_files modified untracked deleted unmerged modified-staged-deleted)'
+complete -f -c git -n '__fish_git_using_command add; and test "$(__fish_git config --type bool --get status.showUntrackedFiles)" != false' -a '(__fish_git_files modified untracked deleted unmerged modified-staged-deleted)'
 # If we have so many files that you disable untrackedfiles, let's add file completions,
 # to avoid slurping megabytes of git output.
-complete -F -c git -n '__fish_git_using_command add; and test "$(__fish_git config --get bash.showUntrackedFiles)" = 0' -a '(__fish_git_files modified deleted unmerged modified-staged-deleted)'
+complete -F -c git -n '__fish_git_using_command add; and test "$(__fish_git config --type bool --get status.showUntrackedFiles)" = false' -a '(__fish_git_files modified deleted unmerged modified-staged-deleted)'
```
krobelus, line 1176: to not break existing behavior, we should probably allow both, e.g. `and begin test bash.showUntrackedFiles ... || test status.showUntrackedFiles ...; end`

soraxas: Yep I agree with the review on the git prompt. 

However, using the same argument, wouldn't it not make sense to piggyback on the bash.xxxx option (which is intented for git prompt)? Doing so has the same effect of coupling the two config behaviour

As seen here, the `bash.xxx` option is only intented for the git prompt (https://github.com/search?q=repo%3Agit%2Fgit%20bash.showUntrackedFiles&type=code) , whereas the `status.xxx` one actually changes the core git behaviour

krobelus: git blame shows the original motivation for setting `bash.showUntrackedFiles` was https://github.com/fish-shell/fish-shell/issues/11095 which also stated

> Also git option status.showUntrackedFiles is set to no, to make it usable

so it looks like they already set the standard option.
Given that there is probably barely anyone relying on only the bash.showUntrackedFiles, I'm fine with this change, yeah

chernetskyi:
@soraxas are you willing to make the changes suggested by @krobelus, so we can merge this PR?

soraxas:
Thanks @chernetskyi the comment had been addressed; this PR now only target the fish git completion

PR #11711, opened 2025-08-07
Update pinned stable Rust to 1.89

danielrainer:
Rust 1.89 introduces new lifetime-related lints:
https://github.com/rust-lang/rust/pull/138677

The current code results in warnings about lifetime syntax mismatches, e.g. in `fn kind(&self) -> Kind`, where the lifetime is elided in `&self` but hidden in `Kind`. We might want to address these issues in the code instead of disabling the warning. To prevent issues for unrelated development, the warning is enabled for now.

It is desirable to merge this quickly, since `build_tools/check.sh` fails if the pinned version in `.github/actions/rust-toolchain@stable/action.yml` is not the most recent stable version, which blocks running the checks for everyone.

This shows another reason why pinning stable Rust is problematic. Not only is a PR like this one required every 6 weeks, but with the current setup it also prevents running the checks, even for code completely unrelated to any changes in the new version. I hope we can improve that before 1.90 is released as stable.

### DIFF
--- a/.github/actions/rust-toolchain@stable/action.yml
+++ b/.github/actions/rust-toolchain@stable/action.yml
@@ -16,7 +16,7 @@ permissions:
 runs:
   using: "composite"
   steps:
-    - uses: dtolnay/rust-toolchain@1.88
+    - uses: dtolnay/rust-toolchain@1.89
       with:
         targets: ${{ inputs.targets }}
         components: ${{ inputs.components }}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -101,6 +101,7 @@ asan = []
 tsan = []
 
 [workspace.lints]
+rust.mismatched_lifetime_syntaxes = "allow"
 rust.non_camel_case_types = "allow"
 rust.non_upper_case_globals = "allow"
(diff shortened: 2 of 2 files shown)


Xiretza:
Addressed by https://github.com/fish-shell/fish-shell/commit/894d4ecc538279ab7014aaa059f3405199bc35a7, which fixes the lints rather than silencing them.

krobelus:
> It is desirable to merge this quickly, since
> `build_tools/check.sh` fails if the pinned version in
> ***@***.***/action.yml` is not the most
> recent stable version, which blocks running the checks for everyone.

Hmm maybe the update check doesn't actually belong
into default check.sh, since the CI doesn't do it either
(though the CI always has the "right" version).

> This shows another reason why pinning stable Rust is problematic.

we can probably improve things here. I'm not yet sure which way to go.

> Not only is a PR like this one required every 6 weeks, but with the
> current setup it also prevents running the checks, even for code
> completely unrelated to any changes in the new version.

Let's maybe remove both version-checks from build_tools/check.sh,
that should fix the concrete problem here, right?

We can look into automating version updates later.

Issue #10256, opened 2024-01-24
add a rust-toolchain.toml to pin the rust version?


asukaminato0721:
I've checked the issues, no relevant issues.

I noticed that

https://github.com/fish-shell/fish-shell/blob/f7b57065dbb06ac9d9a19e696fe039f111758163/.github/workflows/rust_checks.yml#L32-L34

so manually pin a version will fix that. (But require manually bump the toolchain version, which is not so frequently).

something like

```toml
[toolchain]
# older versions may fail to compile; newer versions may fail the clippy tests
channel = "1.75"
components = ["rustfmt", "clippy"]
```

ref https://rust-lang.github.io/rustup/overrides.html

krobelus:
to build the software we don't want to force users to install a specific version of Rust. Any Rust >= 1.66.0 should work.
Between normal CI builds and asan we already test 1.67.0 and nightly, and we clippy-check stable.
 
Clippy and rustfmt checks should either
- not be pinned but follow stable Rust. This means that we want to allow warnings (current state)
- be pinned, then they can fail on warnings.

The second would require developers use a pinned version, and use clippy.

See also the typical use case for rust-toolchain.toml which is not ours:

> Some projects find themselves 'pinned' to a specific release of Rust and want this information reflected in their source repository. This is most often the case for nightly-only software that pins to a revision from the release archives.

krobelus:
FWIW, we do pin the Rust version in CI nowadays, but not via `rust-toolchain.toml`; contributors can use any supported Rust version.
It seems like `rust-toolchain.toml` might have been suggested in https://github.com/fish-shell/fish-shell/pull/9512#issuecomment-1410973355 but I'm not sure if that's worth it given the issues sketched in https://github.com/fish-shell/fish-shell/pull/11718

PR #11712, opened 2025-08-07
Stop pinning stable Rust

danielrainer:
Having stable Rust pinned requires updating the pinned version whenever a new version of stable Rust is released (every 6 weeks, ignoring patch version). This requires manual effort without providing much benefit.

In the current configuration, running `build_tools/check.sh` fails when using the stable toolchain when either the local stable toolchain or the pinned version in `.github/actions/rust-toolchain@stable/action.yml` are outdated. The former is a nice reminder to update the installed stable toolchain. Alternatively, one can use the toolchain of a specific version if the up-to-date stable toolchain is undesirable.
The latter can be very annoying. It results in automatic failure to run the `check.sh` script at least once every 6 weeks, which needs to be resolved by updating the pinned version (ideally accompanied by addressing all new lints). Developers not familiar with the pinning setup might get blocked and annoyed by such failures, and for all developers it is required to
- stop using the `check.sh` script
- or modify it to eliminate the check
- or update the pinned version
- or stop development entirely until the issue is resolved (temporarily if the current setup is kept)

Between stable Rust updating and the pinned version update being merged into master, the development workflow is hindered and might result in pointless rebasing, modifications of the `check.sh` script, or other annoyances.

Not pinning stable Rust eliminates all these issues. The advantage of pinning is that changes in the new version do not automatically break CI runs which ran fine on older versions of Rust. I think this is mostly an issue for code not affected by the PR for which CI runs. Regularly running checks on the master branch using the beta toolchain and fixing detected issues before they make it to the stable toolchain eliminates the problem.


danielrainer:
Fails due to lints introduced in 1.89. This is resolved in https://github.com/fish-shell/fish-shell/pull/11711.

Xiretza:
> Fails due to lints introduced in 1.89.

And that's exactly why the version is pinned in CI. Not pinning the version means semi-randomly running into lint failures entirely unrelated to the actual change whenever a new rust version is released.

There's a point to be made about using the standard [`rust-toolchain.toml`](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file) mechanism to ensure that people using rustup are always using the pinned version locally, though.

danielrainer:
If the code had been checked against the beta toolchain, these issues would have been resolved 6 weeks ago without ever interfering with anyone's development.

With the current pinning approach, CI might not fail due to an implicit toolchain update, sure. But if developers update their local Rust toolchain, which is fairly standard, they run into all these warnings which are completely irrelevant for the things they are working on.

danielrainer:
Locking the toolchain version seems very undesirable to me. The only version which would make sense to lock to is the MSRV. This means that noone can benefit from the improvements contained in more recent versions, e.g. faster compile times, better compilation results, better lints.

And using that approach, updating the version would be a much larger undertaking, as changes would have piled up for potentially years, rather than 6 weeks.

It would also force anyone who wants to compile fish to install our special, blessed version of the toolchain, which is annoying and wasteful.

Xiretza:
> If the code had been checked against the beta toolchain, these issues would have been resolved 6 weeks ago without ever interfering with anyone's development.

Are you volunteering to reliably do this after every release? If this is missed, people will run into bogus lint errors in their PRs, which is no good.

> If developers update their local Rust toolchain, [...] they run into all these warnings

See above, not an issue if we add a `rust-toolchain.toml`.

> The only version which would make sense to lock to is the MSRV.

No, the only version that makes sense to lock is the latest version that was tested, exactly the same as for cargo dependencies in `Cargo.lock`.

> It would also force anyone who wants to compile fish to install our special, blessed version of the toolchain, which is annoying and wasteful.

Downstream is not supposed to compile with `-D warnings`.

I've added a `rust-toolchain.toml` file now, which fixes the issue of local compiles running into lint warnings from newer compilers, closing this.

danielrainer:
> Are you volunteering to reliably do this after every release?

The effort for updating is really insignificant. Most lints even come with a suggested edit to fix them. And if that it too bothersome, a lint can be ignored by adding a single line in `Cargo.toml`. If fish becomes so unmaintained that fixing lints within 6 weeks of them causing  beta toolchain CI to break is a problem, I doubt that there will be anyone to take care of potential PRs running into lint issues.

Regarding `rust-toolchain.toml`, there has been previous discussion about it in https://github.com/fish-shell/fish-shell/issues/10256 and IIRC @krobelus also mentioned it not being a good fit for fish more recently, but I can't find a reference at the moment.

Another problem with forcing a specific toolchain version is that platform support changes. Compare e.g. https://doc.rust-lang.org/1.70.0/rustc/platform-support.html with https://doc.rust-lang.org/1.89.0/rustc/platform-support.html.

There has been a fair amount of discussion regarding different versions of toolchains and their handling in https://github.com/fish-shell/fish-shell/discussions/11584. If you have opinions about how this should be handled I would appreciate it if you first discussed them with other affected developers and come to an agreement, instead of unilaterally pushing your preferred version to master without warning (and testing if CI still works), especially after having received comments disagreeing with your suggested approach.

krobelus:
sorry the update-check was my bad,
can we remove that (and `rust-toolchain.toml`) for now (https://github.com/fish-shell/fish-shell/pull/11718), and then maybe revisit this later

PR #11713, opened 2025-08-08
Refactor history file appending

danielrainer:
## Description

- Fix file id not being set correctly.
- Call `write_all` at most once.
- Call `fsync` after data is written.

The main motivation for these changes is to hopefully eliminate some cases leading to corruption in the history file (#10300). My very limited testing did not encounter any corruption using this code so far, but very frequent corruption using 07ff4e7df065e97a008b1b2995169bd61219662e and other recent commits without the changes contained in this PR. Being reasonably confident that the corruption issues are resolved for me would require at least several months of testing, but I think these changes make sense even if the corruption issue is unaffected.

### DIFF
--- a/src/history.rs
+++ b/src/history.rs
@@ -683,30 +683,20 @@ impl HistoryImpl {
         // So far so good. Write all items at or after first_unwritten_new_item_index. Note that we
         // write even a pending item - pending items are ignored by history within the command
         // itself, but should still be written to the file.
-        // TODO: consider filling the buffer ahead of time, so we can just lock, splat, and unlock?
-        let mut res = Ok(());
         // Use a small buffer size for appending, we usually only have 1 item
         let mut buffer = Vec::new();
-        while self.first_unwritten_new_item_index < self.new_items.len() {
-            let item = &self.new_items[self.first_unwritten_new_item_index];
+        let mut new_first_index = self.first_unwritten_new_item_index;
+        while new_first_index < self.new_items.len() {
+            let item = &self.new_items[new_first_index];
             if item.should_write_to_disk() {
                 append_history_item_to_buffer(item, &mut buffer);
-                res = flush_to_file(
-                    &mut buffer,
-                    locked_history_file.get_mut(),
-                    HISTORY_OUTPUT_BUFFER_SIZE,
-                );
-                if res.is_err() {
-                    break;
-                }
             }
             // We wrote or skipped this item, hooray.
-            self.first_unwritten_new_item_index += 1;
+            new_first_index += 1;
         }
 
-        if res.is_ok() {
-            res = flush_to_file(&mut buffer, locked_history_file.get_mut(), 0);
-        }
+        flush_to_file(&mut buffer, locked_history_file.get_mut(), 0)?;
+        self.first_unwritten_new_item_index = new_first_index;
 
         // Since we just modified the file, update our history_file_id to match its current state
         // Otherwise we'll think the file has been changed by someone else the next time we go to
@@ -715,7 +705,7 @@ impl HistoryImpl {
         // appended remains in our new_items
         self.history_file_id = file_id_for_file(locked_history_file.get());
 
-        res
+        Ok(())
     }
 
     /// Saves history.
--- a/src/history.rs
+++ b/src/history.rs
@@ -696,6 +696,7 @@ impl HistoryImpl {
         }
 
         flush_to_file(&mut buffer, locked_history_file.get_mut(), 0)?;
+        locked_history_file.get().sync_all()?;
         self.first_unwritten_new_item_index = new_first_index;
 
         // Since we just modified the file, update our history_file_id to match its current state
--- a/src/history.rs
+++ b/src/history.rs
@@ -711,7 +711,7 @@ impl HistoryImpl {
         // write.
         // We don't update `self.file_contents` since we only appended to the file, and everything we
         // appended remains in our new_items
-        self.history_file_id = file_id;
+        self.history_file_id = file_id_for_file(locked_history_file.get());
 
         drop(locked_history_file);
 
--- a/src/history.rs
+++ b/src/history.rs
@@ -661,8 +661,10 @@ impl HistoryImpl {
             LockedFile::new(LockingMode::Exclusive(WriteMethod::Append), history_path)?;
 
         // Check if the file was modified since it was last read.
-        let file_id = file_id_for_file(locked_history_file.get());
-        let file_changed = file_id != self.history_file_id;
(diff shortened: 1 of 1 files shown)


danielrainer:
Hm, the ubuntu-asan job failed for the push (https://github.com/danielrainer/fish-shell/actions/runs/16839933111/job/47708485823), but not for the PR. Is the `test_close_during_select_ebadf` test known to be unreliable? From a quick look it does not seem to be related to the content of this PR.

src/history.rs
```diff
@@ -681,46 +683,30 @@ impl HistoryImpl {
...
             // We wrote or skipped this item, hooray.
-            self.first_unwritten_new_item_index += 1;
+            new_first_index += 1;
         }
 
-        if res.is_ok() {
-            res = flush_to_file(&mut buffer, locked_history_file.get_mut(), 0);
-        }
+        flush_to_file(&mut buffer, locked_history_file.get_mut(), 0)?;
+        locked_history_file.get().sync_all()?;
```
ridiculousfish, line 699: See [this comment](https://github.com/fish-shell/fish-shell/pull/11714#pullrequestreview-3103726673) for why we shouldn't use `sync_all` here - an ordinary `fsync` is probably OK though.

ridiculousfish (changes requested):

src/history.rs
```diff
@@ -661,8 +661,10 @@ impl HistoryImpl {
             LockedFile::new(LockingMode::Exclusive(WriteMethod::Append), history_path)?;
 
         // Check if the file was modified since it was last read.
-        let file_id = file_id_for_file(locked_history_file.get());
-        let file_changed = file_id != self.history_file_id;
+        // If someone has replaced the file, forget our file state.
+        if file_id_for_file(locked_history_file.get()) != self.history_file_id {
+            self.clear_file_state();
```
krobelus, line 666: there's an argument for doing as little as possible while holding the lock but this (munmap, free) seems fine

src/history.rs
```diff
@@ -681,46 +683,30 @@
...
-                    &mut buffer,
-                    locked_history_file.get_mut(),
-                    HISTORY_OUTPUT_BUFFER_SIZE,
-                );
-                if res.is_err() {
-                    break;
-                }
             }
             // We wrote or skipped this item, hooray.
-            self.first_unwritten_new_item_index += 1;
```
krobelus, line 702: yeah, `first_unwritten_new_item_index` is speculatively advanced, making it end up in an unpredictable state when writing/syncing fails later.
The new "all or nothing" behavior seems better

src/history.rs
```diff
@@ -681,46 +683,30 @@
...
+            new_first_index += 1;
         }
 
-        if res.is_ok() {
-            res = flush_to_file(&mut buffer, locked_history_file.get_mut(), 0);
-        }
+        flush_to_file(&mut buffer, locked_history_file.get_mut(), 0)?;
+        locked_history_file.get().sync_all()?;
```
krobelus, line 699: so ext4 doesn't guarantee prefix-append crash consistency mentioned in https://github.com/fish-shell/fish-shell/issues/10300#issuecomment-2567063654, but if fsync succeeds before we crash/reboot, then the append should make it?

src/history.rs
```diff
@@ -681,46 +683,30 @@
...
+        self.first_unwritten_new_item_index = new_first_index;
 
         // Since we just modified the file, update our history_file_id to match its current state
         // Otherwise we'll think the file has been changed by someone else the next time we go to
         // write.
         // We don't update `self.file_contents` since we only appended to the file, and everything we
         // appended remains in our new_items
-        self.history_file_id = file_id;
+        self.history_file_id = file_id_for_file(locked_history_file.get());
```
krobelus, line 707: ok, this should only have caused us to redundantly re-read the file?

src/history.rs
```diff
@@ -681,46 +683,30 @@
...
 
-        drop(locked_history_file);
```
krobelus, line 716: FWIW, this change is implied by the parent commit that moved the call to `clear_file_state()`,
so it would make sense to squash it.

krobelus (approved):

krobelus:
> Hm, the ubuntu-asan job failed for the push
> (https://github.com/danielrainer/fish-shell/actions/runs/16839933111/job/47708485823),
> but not for the PR. Is the `test_close_during_select_ebadf` test
> known to be unreliable?

the test relies on select() getting some CPU time within a 500ms
window of wall-clock time; sometimes that's not granted by CI

src/history.rs
danielrainer: Yes, that's how I understand it. We can't guarantee that changes make it to persistent storage correctly when crashes are involved, but by calling `fsync` as soon as possible we at least reduce the chance of a crash occurring between writing to the file and the data actually being persisted.

The explicit `fsync` might also prevent reordering of writes wrt other saving attempts, but I'm not sure if these could actually be an issue without this `fsync`.

src/history.rs
danielrainer: Yes, it's not a soundness issue regarding history file content, but it results in redundant reads and makes the code harder to reason about, since the point of `history_file_id` is tracking which file version was last seen.

src/history.rs
danielrainer: Sure, I'll do that.

danielrainer:
> the test relies on select() getting some CPU time within a 500ms window of wall-clock time

Ok, then is makes sense that it can fail occasionally.

krobelus:
i'll do the rebase for you (since I'm already splitting a commit in the other PR), if needed, it should be doable to see what exactly changed with range-diff/interdiff.

danielrainer:
> i'll do the rebase for you (since I'm already splitting a commit in the other PR)

Great!. Don't forget to change the `sync_all` to also use `fsync` as requested by ridiculousfish.

src/history.rs
krobelus: with this patch, we also no longer update `self.history_file_id` when writing fails for some reason.
This will make us re-read the file later if we did a partial write.
This seems benign, but I'm not sure if this part of the change is intended or helps in some scenario, so maybe we should restore this?

danielrainer: When writing fails, we might have written some data, so I think in that case it is desirable to consider the file outdated by not updating `self.history_file_id` on failed writes, since the file might have changed since we last read it, so blindly updating the id would be incorrect.

krobelus:
@joanbm thanks for https://github.com/fish-shell/fish-shell/issues/10300#issuecomment-2567063654, we've added the proper `close()+fsync()+rename()` sequence now

PR #11714, opened 2025-08-08
Fsync and close before rename

danielrainer:
## Description

This addresses some potential file corruption issues raised in https://github.com/fish-shell/fish-shell/issues/10300#issuecomment-2567063654. It does not prevent the history file corruption I have observed on my system, but it might help in other cases.

### DIFF
--- a/src/fs.rs
+++ b/src/fs.rs
@@ -1,5 +1,5 @@
 use crate::{
-    common::{str2wcstring, wcs2zstring, ScopeGuard},
+    common::{str2wcstring, wcs2zstring},
     fds::wopen_cloexec,
     path::{path_remoteness, DirRemoteness},
     wchar::prelude::*,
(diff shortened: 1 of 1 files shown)


ridiculousfish (changes requested):
`sync_all` ends up invoking [`F_FULLSYNC`](https://github.com/rust-lang/rust/blob/18eeac04fc5c2a4c4a8020dbdf1c652077ad0e4e/library/std/src/sys/fs/unix.rs#L1214) on macOS - this is quite expensive and likely to create user-visible hangs under normal operation. I think that's unacceptable.

I'm OK with an ordinary `fsync` on the temp file though - let's just do that.

danielrainer:
I added a custom `fsync` function to `src/fs.rs` which is similar to `sync_all` but uses `libc::fsync` everywhere. From the macOS man page for `fsync` it seems that without `F_FULLSYNC` we are vulnerable to some kinds of corruption, but I also don't know of an alternative that performs reasonably well, so I guess we just don't prevent these types of corruption on macOS.

If you are happy with this approach I'll wait for this to be merged, and then rebase #11713, so the same `fsync` function can be used.

src/fs.rs
```diff
@@ -173,6 +172,36 @@ impl LockedFile {
...
+    /// Returns the lock file so that the lock can be kept.
+    pub fn fsync_close_and_keep_lock(self) -> std::io::Result<File> {
+        fsync(&self.data_file)?;
+        Ok(self._locked_fd)
+    }
+}
+
+/// Reimplementation of `std::sys::fs::unix::File::fsync` using publicly accessible functionality.
+/// This function is used instead of `sync_all` due to concerns of that being too slow on macOS,
+/// since there `libc::fcntl(fd, libc::F_FULLFSYNC)` is used internally.
```
krobelus, line 189: > it seems that without F_FULLSYNC we are vulnerable to some kinds of corruption

yeah; hopefully not the one we observed though. The NUL bytes should go away with fsync().
Stale metadata might cause other kinds of corruption (hopefully irrelevant ones).

src/fs.rs
```diff
@@ -173,6 +172,36 @@
...
+/// This weakens our guarantees on macOS.
+pub fn fsync(file: &File) -> std::io::Result<()> {
```
krobelus, line 191: doesn't matter much but I guess this doesn't need to be public?

src/fs.rs
```diff
@@ -173,6 +172,36 @@
...
+    let fd = file.as_raw_fd() as c_int;
```
krobelus: I wouldn't use unchecked casts;
I think in this case we should drop the cast because `type RawFd = c_int` is unlikely to change.
Else I'd rather use the guaranteed-lossless `c_int::from()`.

src/fs.rs
```diff
@@ -173,6 +172,36 @@
...
+    loop {
+        match unsafe { libc::fsync(fd) } {
+            0 => return Ok(()),
+            -1 => {
+                let os_error = std::io::Error::last_os_error();
+                if os_error.kind() != std::io::ErrorKind::Interrupted {
+                    return Err(os_error);
+                }
+            }
+            _ => unreachable!("fsync should only ever return 0 or -1"),
```
krobelus: Not much need to change this but I think `unreachable!()` is for bugs in our code; for bugs in other's code we should probably use `panic!()`.
(Though we also use `panic!()` for bugs in our code.)
I didn't check if `unreachable!()` is optimized away in release builds here, which would hypothetically make this an infinite loop.

(Of course we could also reimplement a simple variant of the `cvt_r` interface. Not sure what the nix crate uses.)

src/fs.rs
```diff
@@ -335,98 +364,122 @@
         Ok(())
     }
 
-    const TMP_FILE_SUFFIX: &wstr = L!(".XXXXXX");
-    let tmp_file_template = path.to_owned() + TMP_FILE_SUFFIX;
-    let (tmp_file, tmp_name) = create_temporary_file(&tmp_file_template)?;
-    // Ensure that the temporary file is unlinked when this function returns.
-    let mut tmp_file = ScopeGuard::new(tmp_file, |_| {
-        wunlink(&tmp_name);
```
krobelus, line 343: I think I there is an unrelated (albeit extremely unlikely) issue:
1. fish A runs `rename("fish_history.DEADBEEF")` (rewriting a history file with)
2. fish B starts rewriting the history file; since "fish_history.DEADBEEF" no longer exists, it can in theory use that filename
3. fish A runs `wunlink("fish_history.DEADBEEF")`, destroying fish B's work

So I'm thinking, when we successfully rename this file, we should not call `wunlink()`.

src/fs.rs
```diff
@@ -335,98 +364,122 @@
...
+        // Give up after this many unsuccessful attempts.
+        // TODO: which value, should this be a constant shared with other retrying logic?
+        let max_attempts = 1000;
+        for _ in 0..max_attempts {
+            // Reopen the tmpfile. This is important because it might have been closed by
+            // explicitly dropping it.
+            tmp_file = OpenOptions::new()
+                .write(true)
+                .truncate(true)
+                .open(wcs2osstring(tmp_name))?;
```
krobelus, line 430: I was gonna comment on the lack of `CLOEXEC` flag but realized that Rust already sets that..
We should get rid of our `wopen_cloexec` etc. to make this more obvious.

krobelus (approved):

src/fs.rs
danielrainer: Ah yes, it is public because I want to use it from `src/history.rs` as well. But that's not in this PR, so we can add the `pub` separately if you prefer.

src/fs.rs
danielrainer: > Of course we could also reimplement a simple variant of the cvt_r interface.

This might make sense. But it should probably be addressed in a separate PR. I'll change the `unreachable` to `panic` for now.

> Not sure what the nix crate uses.

I don't know if they even have such retrying logic. A quick search didn't turn up anything that looks like it, but I did not check closely.

src/fs.rs
danielrainer: Good catch. I'll change it to only unlink if `try_rewriting` fails.

krobelus:
> we can add the `pub` separately if you prefer.

yeah I guess that might have been slightly better but no need, it's a rounding error at this point

krobelus:
> I don't know if they even have such retrying logic.

right, nix is more low-level, looks like they require the caller to check for `std::io::ErrorKind::Interrupted`

src/fs.rs
danielrainer: Yes, although I can't really find good docs on it, other than the issues and PRs introducing the behavior. But still, I agree that `wopen_cloexec` does unnecessary work.

src/fs.rs
krobelus: besides error, there's another case (`!potential_update`) where we don't rename, so we still need to unlink in that case.
I'll commit that and split the commit for you (it's a completely unrelated fix)

PR #11725, opened 2025-08-14
Switch to builtin gettext implementation (runtime parsing)

danielrainer:
## Description

Advantages of this approach:
- We have no runtime dependency on gettext anymore.
- The implementation has the same behavior everywhere.
- Our implementation is significantly simpler than GNU gettext.
- We can have localization in cargo-only builds by embedding the MO files.
  Previously localization in such builds only worked as long as the binary was
  not moved from the build directory.
- Generating the MO files from `build.rs` removes the need for scripting it in
  several places (cmake, checks).
- MO file generation happens automatically when building fish, eliminating the
  need to run anything aside from rebuilding fish.

Disadvantages of this approach:
- Our own runtime implementation of gettext needs to be maintained.
- The implementation has a more limited feature set (but I don't think it lacks
  any features which have been in use by fish).
- Unconditionally embedding the MO files bloats the binary size. At the moment
  the MO files take up 3.3 MiB. A release build which embeds all data results
  in a 15 MiB binary (including the MO files), without embedding other data it
  is 8 MiB, meaning almost half the binary size is taken up by MO files.
- Because the `fish-locale` directory is not in a fixed location (cmake puts it
  in `build/`, cargo builds put it in `target/`), we need to enable a feature of
  `rust-embed` which adds several transitive dependencies. This could be
  resolved by using a fixed location for `fish-locale`.
- The `rust-embed` dependency is required for all builds.

## TODOs:
- [ ] Does it make sense to move to a fixed location for `fish-locale`?


danielrainer:
I thought about the problem some more and came up with a different solution in #11726, which I think makes more sense. I'll leave this PR open for now, but this should not be merged if #11726 is chosen as the better approach.

The most important difference is that in this PR MO files are embedded and parsed at runtime, whereas in the other one they are parsed at build time. The latter approach also includes ways of selectively enabling localization, which allows avoiding unnecessary bloat.

PR #11729, opened 2025-08-15
allow `psub --fifo --suffix ...`

bryceberger:
## Description

Prior to 8f22def8f7b71fb9855d435e99d624bc89d1e8d9, this was allowed, and there's been a codepath for it since before the switch to argparse. Seems to work fine from personal use, and added a test case to make sure.

## TODOs:


ridiculousfish:
Thank you

PR #11731, opened 2025-08-16
Fix rust 1.90 lints

danielrainer:
## Description

Fix two lints which are raised in Rust 1.90 beta.

### DIFF
--- a/src/fork_exec/postfork.rs
+++ b/src/fork_exec/postfork.rs
@@ -318,8 +318,8 @@ pub(crate) fn safe_report_exec_error(
                 } else {
                     fstat(fd).map_err(|_| ())
                 };
-                #[allow(clippy::useless_conversion)] // for mode
-                if md.is_err() || unsafe { libc::access(interpreter.as_ptr(), libc::X_OK) } != 0 {
+
+                fn err_or_no_exec_handling(interpreter: &CStr, actual_cmd: &CStr) {
                     // Detect Windows line endings and complain specifically about them.
                     let interpreter = interpreter.to_bytes();
                     if interpreter.last() == Some(&b'\r') {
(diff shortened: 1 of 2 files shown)


src/fork_exec/postfork.rs
```diff
@@ -318,8 +318,8 @@ pub(crate) fn safe_report_exec_error(
                 } else {
                     fstat(fd).map_err(|_| ())
                 };
-                #[allow(clippy::useless_conversion)] // for mode
-                if md.is_err() || unsafe { libc::access(interpreter.as_ptr(), libc::X_OK) } != 0 {
+
+                let err_or_no_exec_handling = || {
```
ridiculousfish: A little worried about this - this is post-fork code, and I can't say with confidence that creating this capturing lambda won't cause a heap allocation. Can we rework this so it's manifestly async-signal safe?

ridiculousfish (changes requested):

src/fork_exec/postfork.rs
danielrainer: I replaced the closure with a function.

We could also keep the old code and add an annotation to ignore the lint there.

src/fork_exec/postfork.rs
```diff
@@ -339,15 +339,29 @@ pub(crate) fn safe_report_exec_error(
...
-                        interpreter,
-                        "', which is a directory."
-                    );
+                }
+
+                match md {
+                    Ok(metadata) => {
+                        #[allow(clippy::useless_conversion)] // for mode
+                        if unsafe { libc::access(interpreter.as_ptr(), libc::X_OK) } != 0 {
+                            err_or_no_exec_handling(interpreter, actual_cmd);
```
krobelus: if we're doing this function approach, let-else would save two levels of indentation

danielrainer: Sure, I changed it to if let ... else. It only saves one level of indentation AFAICT, not sure if you have something different in mind.

krobelus:
> Sure, I changed it to if let ... else. It only saves one level of
> indentation AFAICT, not sure if you have something different in mind.

right, only if we added early return like
`let Ok(metadata) = md else { return err_or_no_exec_handling(); }`

danielrainer:
> `let Ok(metadata) = md else { return err_or_no_exec_handling(); }`

Ah yes, that would work as well. Do you prefer this approach?

danielrainer:
There seems to be some rate limiting or server overload going on which causes build failures due to failing downloads from https://static.rust-lang.org/.

ridiculousfish:
Merged, thank you! 367696b346fd60d317e4ff9fbb23c23ed46931f5

PR #11732, opened 2025-08-16
Consistently use `__fish_cache_sourced_completions`

branchv:
## Description

Just wanted to ensure all fish's generated completions benefit from being cached. Found using

```console
rg -l '\| source' share/completions | xargs rg --files-without-match __fish_cache_sourced_completions | sort
```

A few completions were a little more complicated so I've left those as TODOs ([`gh`](https://github.com/fish-shell/fish-shell/blob/master/share/completions/gh.fish), [`jj`](https://github.com/fish-shell/fish-shell/blob/master/share/completions/jj.fish), [`rclone`](https://github.com/fish-shell/fish-shell/blob/master/share/completions/rclone.fish))

## TODOs:

### DIFF
--- a/share/completions/argocd.fish
+++ b/share/completions/argocd.fish
@@ -1,1 +1,2 @@
-argocd completion fish | source
+__fish_cache_sourced_completions argocd completion fish
+or argocd completion fish | source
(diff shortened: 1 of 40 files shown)


faho:
I left these out on purpose in #10233 - for completions that are fast enough to be loaded on the fly, it's better to avoid the complexity of caching.

If you have a specific thing that would benefit from caching, we can add that, but a lot of tools I see here are things I tested.

PR #11733, opened 2025-08-16
Add cargo feature for enabling gettext extraction

danielrainer:
This allows having the proc macro crate as an optional dependency and speeds up compilation in situations where `FISH_GETTEXT_EXTRACTION_FILE` changes, such as the `build_tools/check.sh` script. Because we don't need to recompile on changes to the environment variable when the feature is disabled, cargo can reuse earlier compilation results instead of recompiling everything. This speeds up the compilation work in `build_tools/check.sh` when no changes were made which necessitate recompilation.
For such runs of `build_tools/check.sh`, these changes reduce the runtime on my system by about 10 seconds, from 70 to 60, approximately. The difference comes from the following two commands recompiling code without the changes in this commit, but not with them:
- `cargo test --doc --workspace`
- `cargo doc --workspace`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,7 +49,7 @@ nix = { version = "0.30.1", default-features = false, features = [
 num-traits = "0.2.19"
 once_cell = "1.19.0"
 fish-printf = { path = "./printf", features = ["widestring"] }
-fish-gettext-extraction = { path = "./gettext-extraction" }
+fish-gettext-extraction = { path = "./gettext-extraction", optional = true }
 
 # Don't use the "getrandom" feature as it requires "getentropy" which was not
 # available on macOS < 10.12. We can enable "getrandom" when we raise the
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -95,6 +95,11 @@ path = "src/bin/fish_key_reader.rs"
 default = ["embed-data"]
 benchmark = []
 embed-data = ["dep:rust-embed"]
+# This feature is used to enable extracting messages from the source code for localization.
+# It only needs to be enabled if updating these messages (and the corresponding PO files) is
+# desired. This happens automatically in tests and when calling
```
krobelus: > happens automatically in tests

AFAICS this is not yet checked in CI or by a "traditional" test invocation but only by `build_tools/check.sh`,
so I'd maybe drop the "automatically" but of course it's not a big deal.
LGTM

danielrainer: Tests do not enable it outside of `build_tools/check.sh`, true. The feature is used in CI, via  the `translation updates` step which calls `build_tools/update_translations.fish`. But I will rephrase this to be more accurate.

PR #11736, opened 2025-08-18
Remove unnecessary rebuild paths

danielrainer:
Cargo tracks normal Rust dependencies and automatically figures out if changes to Rust source files, `Cargo.{toml,lock}`, and `build.rs` necessitate a rebuild, and if so of what. In some cases these checks are smarter than just comparing file modification times, so not specifying such paths explicitly can reduce the amount of rebuilding which happens, without skipping necessary rebuilding.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -63,20 +63,11 @@ fn main() {
         let _ = std::fs::create_dir_all(sec1dir.to_str().unwrap());
     }
 
-    rsconf::rebuild_if_paths_changed(&[
-        "build.rs",
-        "Cargo.lock",
-        "Cargo.toml",
-        "crates/gettext-extraction",
-        "crates/printf",
-        "src",
-    ]);
-
     // These are necessary if built with embedded functions,
     // but only in release builds (because rust-embed in debug builds reads from the filesystem).
     #[cfg(feature = "embed-data")]
     #[cfg(not(debug_assertions))]
-    rsconf::rebuild_if_paths_changed(&["share"]);
+    rsconf::rebuild_if_path_changed("share");
 
     #[cfg(feature = "embed-data")]
     #[cfg(not(clippy))]
@@ -85,6 +76,7 @@ fn main() {
     #[cfg(feature = "gettext-extract")]
     rsconf::rebuild_if_env_changed("FISH_GETTEXT_EXTRACTION_FILE");
 
+    rsconf::rebuild_if_path_changed("src/libc.c");
     cc::Build::new().file("src/libc.c").compile("flibc.a");
 
     let mut build = cc::Build::new();


build.rs
```diff
@@ -63,20 +63,11 @@ fn main() {
...
 
-    rsconf::rebuild_if_paths_changed(&[
-        "build.rs",
-        "Cargo.lock",
-        "Cargo.toml",
-        "gettext-extraction",
-        "printf",
-        "src",
-    ]);
-
```
krobelus, line 74: git blame suggests this is a revert of b2aaf1db52f (Rebuild if src changed, 2025-03-28)
(and 460b93a5bc8 (Rebuild on changes relevant to build artifacts, 2025-03-30)).
Would this reintroduce https://github.com/fish-shell/fish-shell/issues/11332 ?

It's possible that we want a better fix for that, that is, a more specific
criterion for invalidating `git_version_gen.sh`, for example check if contents
of `.git/HEAD` have changed.. though I suppose the dirty state is harder.
Not yet sure.

danielrainer: That's a valid concern. The main issue with including the git version in the binary is that if implemented correctly, we will have to rebuild on pretty much all changes to the directory (changes to `.git/HEAD` are insufficient, both because it will usually point to a branch, meaning changes to commits are missed, and, as you mention, because we also indicate the `dirty` state in the version identifier, so we pretty much have to rebuild on every file change in the repo if we want to keep this behavior and implement it correctly. The current implementation would not change the builtin version if a commit is added which does not touch the files explicitly listed as rebuild triggers, or implicitly required for building fish.

So I think we have several reasonable options:
- Keep the current behavior. This results in some completely unnecessary rebuilds, some rebuilds whose only purpose is updating the version, and still does not guarantee that each `cargo b` results in the binary containing the current commit and dirty state in the version.
- Switch to a coarser version id, e.g. using the version listed in `Cargo.toml`. This practically eliminates the need to recompile on version changes and version updates would automatically trigger a rebuild, so the version would pretty much always be accurate. The downside is of course that we lose precision.
- Fully commit to having accurate commit versions and dirty states in the binary. This would mean lots of recompilation, and is non-trivial to get right. It might be more attractive if we manage to reduce build time, which might be achieved by splitting the code into more crates.

danielrainer: Perhaps we can have a tiny binary crate which depends on the rest of the code, so recompilation of that crate is cheap. Not sure if we would still face significant linking times using this approach.

krobelus: I think we should merge this as is, I think #11332 is only really a problem
if we are missing some `rerun-if-changed` directives.
I think we've fixed those by including share etc.

Now the worst that can happen is that every `git commit --allow-empty -m foo`
does not invalidate the cargo cache, but that's totally fine for the things
that this commit info is used for (bisecting issues), and actually probably a good thing.

Of course this would cause problems when using "git commit --amend"
during the release process, but that should be automated away,
and we don't include the Git commit in released versions anyway.

zanchey: Unfortunately I think most of the build time (especially for release builds) is spent linking :-(

The goal with the git version embedded is to help provide support people using CI builds (from the PPA or wherever). "3.0-alpha1" is pretty unhelpful! I think we can live with the version only changing if something else in the binary changes, or there's a version file deliberately dropped (which is what happens in the release process).

danielrainer: I have an idea which might solve the version update problem: Require the `FISH_BUILD_VERSION` env var to be set at build, otherwise panic. We then set up our build system to unconditionally run code to obtain the version and set the env var accordingly. This allows us to get rid of the version file shenanigans and invalidation issues.

Normal builds via `cargo build` would intentionally be broken. Instead, we adopt the `cargo xtask` approach, and set `FISH_BUILD_VERSION` in the xtask code.

While we still have CMake builds, we can also switch them to using `cargo xtask` and drop all the version-related stuff from CMake.

Issue #11735, opened 2025-08-18
CMake Deprecation Path


0x0D15:
Curious as to what the plan is for the continued use of CMake post-rust migration as it seems that CMake is already passing much of the work over to Cargo as is. 

Would be great to remove another build time dependency for those building in CI etc..

krobelus:
I think the latest writeup was https://github.com/fish-shell/fish-shell/pull/11583#discussion_r2150511320

AFAIK there are no blockers, it's mainly about porting a bit of code into installation scripts.

danielrainer:
Yes, we should be able to move most things into separate scripts. One annoyance with cargo is that there does not seem to be a good way to run code after compilation.

AFAIK we don't know yet if we want to keep using make. At the moment we have a `GNUmakefile` and a `BSDmakefile` which is annoying, but they don't do much complicated work. I think we might be able to use cargo as the main build system and possibly call some shell scripts from `build.rs` scripts.

One thing which might be nontrivial to get without CMake are install locations, including for auxiliary files like completions and fish-script functions. We already allow building with the `embed-data` feature, which is enabled by default when calling cargo directly rather than via CMake. With this feature active, the files get embedded into the binary, so there is no need to put them in specific locations on the filesystem when installing fish. This doesn't work for gettext MO files which are used for localization, but we have PRs addressing that in the works. So when that is addressed, we could make all builds embed the required data, which would mean that we get a single binary which can be moved around arbitrarily and does not depend on any external files.

For what it's worth, the only reason I'm still using CMake for fish is to check if CMake builds still work after changing something. Otherwise, I use cargo directly for building and the `build_tools/check.sh` script for testing.

There are a few things for which we don't have a convenient CMake alternative yet, like building HTML docs (which is mostly just calling `sphinx-build`, but it depends on `fish_indent`, so it needs to run after building that), but for most things it just needs someone to spend some time porting the CMake code to an alternative.

zanchey:
Install locations and `DESTDIR` support are a pretty hard requirement for downstream packages. A small Makefile which knows how to drive cargo to generate the binaries and has an install target with the right variables seems like it would be enough. It's annoying to write in portable Make but I've mostly got it done for simple use cases. Looks a bit like https://gitlab.com/sequoia-pgp/sequoia-sqv/-/blob/main/Makefile

As an aside, I'm kind of amazed there's not already a consensus solution to this; is nobody shipping programs that are more than CLI utilities but not, like, a browser?

[...]

PR #11737, opened 2025-08-18
Build man pages in separate crate

danielrainer:
Building man pages takes significant time due to Sphinx running for several seconds, even when no updates are required. Previously, we added custom logic to avoid calling `sphinx-build` if the inputs to `sphinx-build` had not changed since a cached timestamp.

By moving this into its own crate, we can tell cargo to rebuild when the input files changed and unrelated changes will have no effect on this crate. This allows us to get rid of the custom code for tracking whether to recompile, while keeping the effect of only calling `sphinx-build` when appropriate.


danielrainer:
Hm, I forgot about `sphinx-build` depending on `fish_indent`. Is there any need for this in the context of building man pages?

Possible solutions/workarounds:
- Find a way to remove the dependency on `fish_indent` when building man pages.
- Provide some `fish_indent` in the CI. In this case, the first build succeeds, because it specifies `--no-default-features`. The test target does so as well, but I think the `--workspace` flag (`cmake/Tests.cmake:60`) causes a build of the man page crate. We could add the build directory containing `fish_indent` to `PATH` when running `cargo test` or add a feature to the man page crate and only actually build the man pages if the feature is enabled. Enabling it could happen automatically from the main `Cargo.toml`.
- Making the man page crate depend on `fish_indent` in a way cargo understands seems difficult to achieve. We could maybe refactor the crate structure in a way that allows building fish_indent cheaply, but depending on binary crates seems to be unsupported by cargo: https://github.com/rust-lang/rfcs/blob/master/text/3028-cargo-binary-dependencies.md.
- Install some version of `fish_indent`, either via the package manager or by building with `embed-data` disabled.

krobelus:
> Hm, I forgot about `sphinx-build` depending on `fish_indent`. Is there any need for this in the context of building man pages?

Not today. Whereas sphinx-docs does depend on fish_indent for
highlighting fish code snippets, sphinx-manpages does not (yet).

We might need to make sure that we don't need this snippet when building manpages:

    ret = subprocess.check_output(
        ("fish_indent", "--version"), stderr=subprocess.STDOUT
    ).decode("utf-8")

probably by setting `FISH_BUILD_VERSION`...

danielrainer:
> We might need to make sure that we don't need this snippet when building manpages:
>
> ```py
> ret = subprocess.check_output(
>     ("fish_indent", "--version"), stderr=subprocess.STDOUT
> ).decode("utf-8")
> ```

I think we should be able to use
```py
    ret = subprocess.check_output(
        ("../build_tools/git_version_gen.sh", "--stdout"), stderr=subprocess.STDOUT
    ).decode("utf-8")
```
instead. It might actually make sense to always use that and get rid of `FISH-BUILD-VERSION-FILE` which requires special handling in several places and is prone to becoming outdated.

danielrainer:
For man pages, I think it doesn't really matter, since AFAICT the version doesn't actually get included. But we might want to change that.

danielrainer:
```patch
diff --git a/doc_src/conf.py b/doc_src/conf.py
index 3bd1f577a..181360e78 100644
--- a/doc_src/conf.py
+++ b/doc_src/conf.py
@@ -35,6 +35,13 @@ def issue_role(name, rawtext, text, lineno, inliner, options=None, content=None)
     return [link], []
 
 
+def do_not_use_fish_indent_for_man(app):
+    from sphinx.highlighting import lexers
+
+    if app.builder.name == "man":
+        del lexers["fish-docs-samples"]
+
+
 # -- Load our extensions -------------------------------------------------
 def setup(app):
     # Our own pygments lexers
@@ -52,6 +59,8 @@ def setup(app):
     app.add_config_value("issue_url", default=None, rebuild="html")
     app.add_role("issue", issue_role)
 
+    app.connect("builder-inited", do_not_use_fish_indent_for_man)
+
 
 # The default language to assume
 highlight_language = "fish-docs-samples"
```

This might work for removing the other `fish_indent` dependency when building man pages. I haven't checked if it breaks anything.

danielrainer:
I just ran a recursive diff on the output in `man1` of the version with and without the proposed changes and they are identical.

krobelus:
> I think we should be able to use
> ```py
>     ret = subprocess.check_output(
>         ("../build_tools/git_version_gen.sh", "--stdout"), stderr=subprocess.STDOUT
>     ).decode("utf-8")
> ```
> instead.

Yeah that seems okay at a glance. I hope that our
install/package/release scripts use the more robust approach of writing
it to a file once and never recomputing it.

If sphinx guarantees $PWD=doc-src then I guess then we don't need an
absolute path here.

> It might actually make sense to always use that and get
> rid of `FISH-BUILD-VERSION-FILE` which requires special handling
> in several places and is prone to becoming outdated.

I guess so. The justification is

	# The essential idea is that we are going to
	# invoke git_version_gen.sh, which will update the
	# FISH-BUILD-VERSION-FILE only if it needs to change; this
	# is what makes incremental rebuilds fast.

which I don't really understand. Even if we run git_version_gen.sh (60ms) every time,
we would only update FISH-BUILD-VERSION-FILE if the results actually changed.
But I'd probably need to do more research.

We also have a duplicated functionality in
`build_tools/git_version_gen.sh` and build.rs's `get_version()`,
we should probably get rid of this

danielrainer:
> If sphinx guarantees $PWD=doc-src then I guess then we don't need an absolute path here.

I think that's the case. The script also uses relative paths in other places, like globbing on `cmds/*`.

> Even if we run git_version_gen.sh (60ms) every time, we would only update FISH-BUILD-VERSION-FILE if the results actually changed.

Yes. For building docs I think the runtime of this script is not much of a concern, since they probably are not built that frequently and the actual `sphinx-build` command takes about two orders of magnitude more time.

> We also have a duplicated functionality in `build_tools/git_version_gen.sh` and build.rs's `get_version()`, we should probably get rid of this

Yes, this duplication is not optimal. I think having a Rust version is good, because it can be more performant than a shell script, especially if we are already running `build.rs` for other reasons.

I think it would be great if we could have a way to tell cargo to always run code which updates the version info, but then only rebuild fish if the info changed. Most of this process should be straightforward, but the thing I don't have a solution for is how we can avoid rebuilding fish if the version did not change, while at the same time always ensuring an up-to-date version file.

zanchey:
> AFAICT the version doesn't actually get included.

The version is embedded in the TH macro, but switching to calling `git-version-gen.sh` makes sense to me.

My recollection is that the FBVF/witness dance was important for CMake to avoid rebuilding `fish_version.o`, and therefore the whole binary, in various circumstances, but also to prevent CMake from assuming it was up to date. Testing that the build still works under CMake with a clean and dirty tree would be important before switching all uses to `git-version-gen.sh`, although I can't see it being nearly as complex now.

danielrainer:
> The version is embedded in the TH macro

Interesting, I grepped for the git hash part and did not find anything. Looking for the `.TH` macro, it seems that only major and minor version are used, e.g. `.TH "RETURN" "1" "Aug 18, 2025" "4.0" "fish-shell"`.

Regarding ensuring correct version info for each build I agree that we need to find a way to achieve that, but it should be independent of CMake, so it does not block transitioning away from CMake. One option is using a Makefile which always updates a version file before it does anything else. The downside of that is that when using cargo directly the update does not happen. And it requires depending on make for building, which might otherwise not be necessary.

zanchey:
> Looking for the `.TH` macro, it seems that only major and minor version are used,

Yeah, this is done with https://github.com/fish-shell/fish-shell/blob/feaec20de6e07b1d1d12badd160fbab90b4a793d/doc_src/conf.py#L78-L81

I've written separately about a make dependency in #11735, but it seems to me that replacing CMake with anything other than cargo and perhaps a small Makefile for supporting installation is a bit of a waste of time.

danielrainer:
> it seems to me that replacing CMake with anything other than cargo and perhaps a small Makefile for supporting installation is a bit of a waste of time.

I completely agree, the only thing I'm unsure about is whether we should use make or just rely on cargo + maybe some shell scripts. Make helps with discoverability, but if we use it for something like version updates, directly using cargo results in risking outdated version info.

build-man-pages/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+name = "fish-build-man-pages"
+version = "0.1.0"
+rust-version.workspace = true
+edition.workspace = true
+repository.workspace = true
+
+[dependencies]
+
+[build-dependencies]
+rsconf = "0.2.2"
```
krobelus: unrelated I guess but we should really be using versions from the top-level `Cargo.toml` here

build-man-pages/build.rs
```diff
@@ -0,0 +1,88 @@
...
+
+fn canonicalize<P: AsRef<Path>>(path: P) -> PathBuf {
+    std::fs::canonicalize(path).unwrap()
+}
+
+fn main() {
+    let cargo_target_dir: PathBuf = option_env!("CARGO_TARGET_DIR")
+        .map(canonicalize)
+        .unwrap_or(canonicalize(MANIFEST_DIR).parent().unwrap().join("target"));
+    let targetman = cargo_target_dir.join("fish-man");
```
krobelus: I think we should add a "fish-build" crate in "crates/build" (or "build-helper" ?) to avoid copying the build.rs code (assuming it's not going away anytime soon).

The helper to compute `cargo_target_dir` might need to be a macro because it needs to access `$CARGO_TARGET_DIR` at compile time..
not sure, probably works fine as a normal function, because both crates should be compiled with the same target dir.

build-man-pages/build.rs
```diff
@@ -0,0 +1,88 @@
...
+    let sec1dir = targetman.join("man1");
+    let _ = std::fs::create_dir_all(sec1dir.to_str().unwrap());
+    #[cfg(not(clippy))]
+    build_man(&targetman);
+}
+
+// disable clippy because otherwise it would panic without sphinx
+#[cfg(not(clippy))]
+fn build_man(build_dir: &Path) {
```
krobelus: I guess we should define main() to be an empty function if `cfg(clippy)` ?
No need to have clippy create the directory etc.
Then we can inline `build_man()`.

build-man-pages/build.rs
```diff
@@ -0,0 +1,88 @@
...
+    use std::process::Command;
+    let mandir = build_dir;
+    let sec1dir = mandir.join("man1");
+    let docsrc_path = canonicalize(MANIFEST_DIR).parent().unwrap().join("doc_src");
+    let docsrc = docsrc_path.to_str().unwrap();
+    let sphinx_doc_sources: [&str; 3] = ["../CHANGELOG.rst", "../CONTRIBUTING.rst", docsrc];
```
krobelus: unrelated: we probably shouldn't specify the cardinality, better use

	let sphinx_doc_sources: &[&str]: &["", ...];

build-man-pages/build.rs
```diff
@@ -0,0 +1,88 @@
...
+    rsconf::rebuild_if_paths_changed(&sphinx_doc_sources);
```
krobelus: yeah, this looks like a much better solution.
So the empty lib.rs turned empty build-man-pages.rlib acts as our timestamp file.

doc_src/conf.py
```diff
@@ -35,6 +35,13 @@ def issue_role(name, rawtext, text, lineno, inliner, options=None, content=None)
     return [link], []
 
 
+def do_not_use_fish_indent_for_man(app):
+    from sphinx.highlighting import lexers
```
krobelus: better move this duplicated import to a single global one, unless we're saving something, which I don't expect, because I think `setup()` needs to be called in every use of conf.py.

doc_src/conf.py
```diff
@@ -35,6 +35,13 @@
...
+
+    if app.builder.name == "man":
+        del lexers["fish-docs-samples"]
```
krobelus, line 41: yeah this looks safe because I'd expect sphinx to fail if a lexer is not defined
(especially now that we actually check for warnings in CI)

build-man-pages/Cargo.toml
danielrainer: Makes sense, I'll switch it to workspace dependencies. We should probably do the same for the other crates as well.

Regarding versions of the packages, I think it might also make sense to switch to a single workspace version instead of using `0.1.0` in most crates, since I don't think anyone wants to bother with managing version numbers for each crate. The `printf` crate seems to have a non-default version, so maybe that should be kept.

danielrainer: Also, do you think we should have all dependencies as workspace dependencies, only the ones which are used at least twice, or only the ones used outside of the main `Cargo.toml`?

build-man-pages/build.rs
danielrainer: Creating `target/man/man1` is actually important, otherwise running `cargo clippy` on a clean build will panic, because `rust-embed` panics when it can't find the directory it's supposed to embed. This certainly deserves a comment.

We could also just drop the special handling for clippy. AFAICT, it was introduced in 6c852655dd2cb37023f256033efe671a4db02680 for a valid reason, but that reason should no longer be relevant since 10ac98e2eacfb44135653a6b95aefcf132d7dd8e. So it think the only reason to keep it is as a performance optimization, which I think is sufficient, since sphinx-build does run for a long time and without the custom timestamp handling I think we will run `sphinx-build` repeatedly because `cargo build` and `cargo clippy` build separately AFAICT.

build-man-pages/build.rs
danielrainer: Yes, that's a leftover of it formerly being a `const`, no need for a type annotation here anymore..

build-man-pages/build.rs
danielrainer: I agree. I'll add a `build-helper` crate. I think using normal functions should work. Cargo automatically rebuilds if an env var used via `env!` or `option_env!` changes. (Or more accurately, caches build for different sets of env vars separately, so a change might not trigger a rebuild if there is a cached build for the current set of env vars.)

krobelus:
> +[build-dependencies]
> +rsconf = "0.2.2"
> 
> Also, do you think we should have all dependencies as workspace
> dependencies, only the ones which are used at least twice, or only
> the ones used outside of the main `Cargo.toml`?

I think declaring every dependency version `[workspace.dependencies]` in our top-level Cargo.toml
seems like the least messy and hardest to mess up (it seems fairly easy to enforce in CI that we don't specify versions in crates/*/Cargo.toml).
I'm assuming we will stick to the one-version rule.

But there are some projects (like rust-analyzer) that don't,
so I'd first want to find out why.

krobelus:
> Regarding versions of the packages, I think it might also make sense
> to switch to a single workspace version instead of using `0.1.0`

a cursory look suggests that by convention, unpublished creates
have version 0.0.0, see commit

https://github.com/rust-lang/cargo#077b1bc31 (chore: Mark unpublished crates as such, 2023-04-26)

now we don't cargo-publish any crate at this point but this seems like a good idea either way
I guess we don't need to add "publish = false" at this point.

> The `printf` crate seems
> to have a non-default version, so maybe that should be kept.

yeah, fish-printf is published (I think Debian requires that, so we might need to publish all the others too. Not sure)

danielrainer:
> I think declaring every dependency version `[workspace.dependencies]` in our top-level Cargo.toml seems like the least messy and hardest to mess up

That has been my assessment as well.

> But there are some projects (like rust-analyzer) that don't, so I'd first want to find out why.

RA seems to have adopted workspace dependencies for local crates not long after the feature became available in https://github.com/rust-lang/rust-analyzer/pull/13969, but there is no reasoning provided there. Single dependencies have been moved to workspace dependencies in PRs like https://github.com/rust-lang/rust-analyzer/pull/15913, but without any reasoning there either. At the moment it seems like most RA dependencies are declared a workspace dependencies, with some exceptions. I don't know if these exceptions are intentional, as I have not found any reasoning about it.

The cargo reference section about workspace dependencies also does not include any guidance which approach might be preferable under which circumstances: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#inheriting-a-dependency-from-a-workspace

I think for fish we could switch everything to workspace deps. If we notice any problems it's not hard to switch back.

danielrainer:
> I think Debian requires that, so we might need to publish all the others too.

Interesting, would be good to find out the exact requirements. I don't think that publishing the other crates makes much sense.

Looking at https://rust-team.pages.debian.net/book/process-workspace.html it seems that the Debian Rust Team does not have a fully formed policy yet. There is:

> For programs, it doesn't matter if the crates are not published on crates.io.

But I'm not sure if "programs" refers to the main thing to be packaged (fish in our case), or whether it is a library or binary crate.

In any case, I don't think we should let limitations of specific distro packaging approaches prevent us from making sensible design choices. Workspaces are an important and widely-used feature of Rust and putting local, unpublished dependencies in them is one of their core features. So if distros want to package software written in Rust, but using custom tooling, they should ensure that their tooling supports Rust software as it is written, not the other way around where developers need to worry about whether all distros which might want to package their software have gotten around to updating their distro-specific tooling to support normal features of the default build system of the language used.

I'm in favor of setting the version to `0.0.0` for our unpublished crates and to keep them unpublished. Otherwise, we could also use `workspace.version = true`. In either case we don't need to worry about independently managing version identifiers we don't care about.

build-man-pages/build.rs
krobelus: > because rust-embed panics when it can't find the directory it's supposed to embed

shouldn't we disable this when running clippy? I don't know why clippy would need to read `share/`.

crates/build-helper/src/lib.rs
```diff
@@ -0,0 +1,25 @@
...
+pub fn canonicalize<P: AsRef<Path>>(path: P) -> PathBuf {
+    std::fs::canonicalize(path).unwrap()
+}
+
+pub fn get_repo_root() -> PathBuf {
+    let manifest_dir = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+    canonicalize(manifest_dir.ancestors().nth(2).unwrap())
+}
+
+pub fn get_target_dir() -> PathBuf {
```
krobelus, line 15: Merging with rebase & some cosmetic changes. Hope I didn't mess anything up.

crates/build-helper/src/lib.rs
```diff
@@ -0,0 +1,25 @@
...
+    option_env!("CARGO_TARGET_DIR")
+        .map(canonicalize)
+        .unwrap_or(get_repo_root().join("target"))
+}
+
+pub fn rebuild_if_paths_changed<P: AsRef<Path>, I: IntoIterator<Item = P>>(paths: I) {
```
krobelus, line 21: I think this should be moved into rsconf in future.

crates/build-man-pages/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "fish-build-man-pages"
+version = "0.1.0"
+rust-version.workspace = true
+edition.workspace = true
+repository.workspace = true
```
krobelus, line 6: we might also want `license.workspace`, `homepage.workspace`.
`readme.workspace` probably makes less sense.

Leaving this for later, not sure about license.

build-man-pages/build.rs
danielrainer: Well, then we would probably have to cfg-guard pretty much all the embed-data-guarded stuff with additional `#[cfg(not(clippy))]`, which gets especially annoying if we start to always embed the data files.

crates/build-helper/src/lib.rs
danielrainer: Yes, that would make sense to me as well.

build-man-pages/build.rs
krobelus: right, it makes sense to make clippy use the normal embed-enabled code paths

PR #11743, opened 2025-08-20
Fix  java_home version completion

ghostflyby:
## Description

fix `java_home` completion problems due to change of osascirpt default output behavior.

Fixes issue #

## TODOs:

### DIFF
--- a/share/completions/java_home.fish
+++ b/share/completions/java_home.fish
@@ -1,27 +1,10 @@
 function __fish_complete_macos_java_version
-    set -l xslt (mktemp)
-
-    printf '%s\n' \
-        '<?xml version="1.0" encoding="UTF-8"?>' \
-        '<xsl:stylesheet version="1.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">' \
-        '  <xsl:output method="text" encoding="UTF-8"/>' \
-        '  <xsl:template match="/">' \
-        '    <xsl:for-each select="//dict">' \
-        '      <xsl:variable name="v"   select="string(key[.=\'JVMVersion\']/following-sibling::*[1])"/>' \
-        '      <xsl:variable name="a"   select="string(key[.=\'JVMArch\']/following-sibling::*[1])"/>' \
-        '      <xsl:variable name="n"   select="string(key[.=\'JVMName\']/following-sibling::*[1])"/>' \
-        '      <xsl:variable name="ven" select="string(key[.=\'JVMVendor\']/following-sibling::*[1])"/>' \
-        '      <xsl:value-of select="concat($v,\'&#9;\',$a,\' \',$n,\' by \',$ven)"/>' \
-        '      <xsl:text>&#10;</xsl:text>' \
-        '    </xsl:for-each>' \
-        '  </xsl:template>' \
-        '</xsl:stylesheet>' >$xslt
-
-    /usr/libexec/java_home -X | xsltproc $xslt -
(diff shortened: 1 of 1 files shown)


share/completions/java_home.fish
```diff
@@ -1,6 +1,6 @@
 function __fish_complete_macos_java_version
     set -l json (/usr/libexec/java_home -X|plutil -convert json -o - -)
-    osascript -l JavaScript -s o -e "JSON.parse('$json').forEach(e => console.log(`\${e.JVMVersion}\t\${e.JVMArch} \${e.JVMName} by--exec \${e.JVMVendor}`))"
+    osascript -l JavaScript -s o -e "JSON.parse('$json').forEach(e => console.log(`\${e.JVMVersion}\t\${e.JVMArch} \${e.JVMName} by--exec \${e.JVMVendor}`))" 2>&1
```
krobelus: why? this seems wrong, we don't want to put errors into completions. Better fix the error

ghostflyby: The osascript JavaScript had outputted  `console.log` to stdout but switched to stderr since some macOS update. 

In this case, the `$json` from java_home is guaranteed to be valid, providing no unexpected output.

krobelus:
>  function __fish_complete_macos_java_version
>      set -l json (/usr/libexec/java_home -X|plutil -convert json -o - -)
> -    osascript -l JavaScript -s o -e "JSON.parse('$json').forEach(e => console.log(`\${e.JVMVersion}\t\${e.JVMArch} \${e.JVMName} by--exec \${e.JVMVendor}`))"
> +    osascript -l JavaScript -s o -e "JSON.parse('$json').forEach(e => console.log(`\${e.JVMVersion}\t\${e.JVMArch} \${e.JVMName} by--exec \${e.JVMVendor}`))" 2>&1
> 
> The osascript JavaScript had outputted  `console.log` to stdout
> but switched to stderr since some macOS update.

can't we use a function that prints to stdout across all relevant versions of macOS?
Ideally we leave the error stream untouched, in case there are ever other errors.
If osascript's js interpreter can't do it, we can use `python -c 'import json; ...'`.

ghostflyby:
Python is not longer bundled with macOS, so I made a xslt version, which all uses bundled tools without redirecting stderr.

share/completions/java_home.fish
```diff
@@ -1,10 +1,27 @@
 function __fish_complete_macos_java_version
-    set -l json (/usr/libexec/java_home -X|plutil -convert json -o - -)
-    osascript -l JavaScript -s o -e "JSON.parse('$json').forEach(e => console.log(`\${e.JVMVersion}\t\${e.JVMArch} \${e.JVMName} by--exec \${e.JVMVendor}`))"
+    set -l xslt (mktemp)
+
+    printf '%s\n' \
+        '<?xml version="1.0" encoding="UTF-8"?>' \
+        '<xsl:stylesheet version="1.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">' \
+        '  <xsl:output method="text" encoding="UTF-8"/>' \
+        '  <xsl:template match="/">' \
```
krobelus: this seems noisy, and you're not cleaning up the tmp file.
Better use javascript's `process.stdout.write`

ghostflyby: `process.stdout.write` is a nodejs function.

ghostflyby:
It turns out that the last expression will be outputted to stdout.

PR #11746, opened 2025-08-21
Version as required env var

danielrainer:
## Description

This is an incomplete implementation of my idea for improving the situation for embedding the version string into build artifacts. The core idea is that we wrap cargo invocations via `cargo xtask cargo`, and have the xtask unconditionally update the `FISH_BUILD_VERSION` env var. Cargo will then automatically rebuild if the version changed. This allows us to get rid of all the version file stuff.

In addition, a few xtasks are implemented to show how they can be used.

This is not ready to be merged yet. I'd like to wait for #11737 to be merged, so it's easier to create an xtask for man page generation. The version file stuff should also get fully replaced before merging. We will also need documentation updates if we break normal cargo invocation and rust-analyzer.

At the moment, CI using Rust 1.70 fails for the stupid reason of `std::process::ExitCode` not implementing `PartialEq`. This is quite annoying, and [the docs](https://doc.rust-lang.org/std/process/struct.ExitCode.html#impl-PartialEq-for-ExitCode) seem to be wrong. They claim that the `PartialEq` implementation has been available since Rust 1.61.0, but actually it is only available starting with 1.83, so we might have to use a different type for exiting from xtasks.


build.rs
```diff
@@ -43,9 +43,28 @@ fn main() {
...
+    // detect if the code is compiled by rust-analyzer, so we can't put special handling for that
+    // into our code.
+    // https://github.com/rust-lang/rust-analyzer/issues/17766
+    let version = env!(
+        "FISH_BUILD_VERSION",
+        concat!(
+            "Tried to build without the FISH_BUILD_VERSION environment variable set.\n",
+            "This is not supported.\n",
+            "Use the `cargo xtask` interface instead of running cargo builds directly.\n",
+            "E.g. `cargo xtask cargo build`"
```
krobelus: I don't think the "stale" version phenomenon is enough reason to break `cargo b`.

danielrainer: There are alternative approaches which don't break cargo. I chose this most extreme one to encourage discussion over whether this is the approach we want.

We could still allow plain cargo builds, which compute the version for themselves if `FISH_BUILD_VERSION` is not set. If we combine this with a `rebuild_if_path_changed` for a path which does not (and will not) exist, cargo will just rebuild on every invocation, which also ensures an up-to-date version. If we don't add such a rebuild condition, we get the current behavior which allows for stale versions, but will update them if files relevant for the build are changed.

danielrainer: We could also add a warning to encourage people to compile via an xtask, like the warning emitted when `sphinx-build` is not installed.

danielrainer: Updated to allow builds without `FISH_BUILD_VERSION` set, but always rebuilding in such cases.

danielrainer:
Updated to get rid of `ExitCode` problems. Includes some minor refactoring.

build.rs
krobelus: I haven't looked into this but I'm not sure if we want this; seems like an academic problem(?).
I don't expect any other rust app does this.
If this is a problem worth solving, it could probably be solved better in cargo, for everyone.

> Updated to allow builds without FISH_BUILD_VERSION set, but always rebuilding in such cases.

silently making some rebuild scenarios slow also doesn't seem ideal.
I'm also scared to make subtle changes that could make clippy/rust-analyzer worse (though I don't know if that's true in practice)

danielrainer: > silently making some rebuild scenarios slow also doesn't seem ideal.

By emitting the `cargo:warning` lines it would not be silent for builds where you can see the output. For other builds, like in rust-analyzer, I think they will only rebuild if a rebuild is required for reasons other than a version change, so there it shouldn't make a difference.

I do get your point about the added complexity maybe not being worth the perfect version information over the mostly correct info we already get.

The `cargo xtask` approach in general seems useful to me for tasks like building HTML docs or installing. If we adopt it for all tasks, it would be significantly easier to add additional custom functionality to builds, if we ever want that.

Regarding solving problems in cargo, I completely agree, but it happens way too often that I run into cargo issues that have been open for almost 10 years with no significant progress, despite being about functionality which is clearly desirable for many people. One example is the issue about installing files other than binaries which was brought up in the issue about CMake deprecation (https://github.com/rust-lang/cargo/issues/2729). This does not inspire confidence that we will get a cargo-native solution anytime soon. And even if it is implemented, we would have to wait for our MSRV to catch up to that.

danielrainer: One even less intrusive variant we could use is to keep the old behavior when `FISH_BUILD_VERSION` is not set, so we don't get extra rebuilds which might be undesirable, but in addition we emit a warning that the built-in version might not be correct if `FISH_BUILD_VERSION` is not set during build.

danielrainer:
Integrated updates from master.

Regarding the `FISH-BUILD-VERSION-FILE` stuff, I think that generating `fish.pc` is the last holdout. Once we have an xtask for that we should be able to remove all version file stuff from CMake.

PR #11750, opened 2025-08-21
Added VS Code editor settings

IsaacOscar:
This adds a VS Code workspace settings file to apply the same settings as in the existing editor config file.
The first commit 039e308f897d865fa2b5d0cb192a496c5e8653fa simply adds the settings to mimic the .editorconfig file as close as possible (and recommend a few useful extensions). VS Code unfortunately doesn't come with .editorconfig support, but there is an extension [EditorConfig for VS Code](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig), but it only supports some of the features you are using. So the workspace settings I've included will automatically recommend that extension,  and manually define the settings that extension doesn't support.

The second commit I've made 114b98f348f8cadcae35945b2e6e918d2b38213d simply removes the line length limits on `.md` and `.rst` files.


fish-shell.code-workspace
```diff
@@ -0,0 +1,92 @@
...
+            // As explained above, no easy way to make this ONLY apply to fish files in
+            // share/{completions,functions}
+            "editor.rulers": [], "rewrap.autoWrap.enabled": false, // like max_line_length = unset
+        },
+
+        // Put this in your config.fish file to have vscode used as your git editor,
+        // but only when running git commands from VS Code's integrated terminal:
+        //  test "$TERM_PROGRAM" = "vscode" && set -gx EDITOR code -w
+        "[git-commit]": { // COMMIT_EDITMSG
+            "editor.rulers": [72], "rewrap.wrappingColumn": 72 // like max_line_length = 72
```
krobelus: I'll let VSCode users review.
If we take this, we should also try to remove the duplication (the 72 is already specified in `.editorconfig` etc.)

IsaacOscar: I've updated the pull request, I've discovered that you don't need to set `rewrap.wrappingColumn` if you also set `editor.rulers`.

As for the duplication with the `.editorconfig` file, you can complain at https://github.com/editorconfig/editorconfig-vscode/issues/272 or  https://github.com/stkb/Rewrap/issues/55

IsaacOscar:
I've updated some of the recommended extensions, and added a new commit that autoformats rust, python, and fish code on save. (Only VS Code users will benefit from those, as this is way out of scope of what `.editorconfig` supports).

IsaacOscar:
I've added a new commit that makes some of the settings only apply to specific folder, to match the `.editorconfig` and `build_tools/style.check`.

Importantly, the fish files in `tests` will no longer have `fish_indent` run on them on save. (As this will break various tests).

Obviously for some tests, being badly formatted is important, for example:
```
echo 'foo\
bar'
```
(Which fish_indent formats to `echo foobar`).

However, most of the tests have no reason to depart from the "official" style, so perhaps you should add a way to easily exclude some parts of the code from fish-indent, e.g.
```
....
#fish_indent: stop
...
#fish_indent: start
...
```
And then all code in between the two comments will not be touched by `fish_indent` (and of course you could omit the `#fish_indent: start` if you want the entire remainder of the file to be ignored).

Note: the way I'm doing folder specific settings is a hack I got from https://stackoverflow.com/questions/47405315/visual-studio-code-and-subfolder-specific-settings (however I didn't bother adding a top-level `.vscode/settings.json` with a `"files.exclude"`, as all that seems to do is hide the folder from the the `"fish-shell"` folder tree, which is just confusing).

ndonfris:
Consider adding [fish-lsp](htts://github.com/ndonfris/fish-lsp) to the recommended extension list. 

The fish indent feature you mention should be pretty straightforward to implement in the server, considering the server uses a [similar comment 
syntax](https://github.com/ndonfris/fish-lsp/tree/feat/v1.0.11.settings?tab=readme-ov-file#command-flags) for toggling diagnostics warnings.  It also already supports format range request. Shouldn't take me too long to add support for this use case 

Also, if any fish shell authors or maintainers, have ideas/requests towards the server, feel free to fork and implement them.

krobelus:
> However, most of the tests have no reason to depart from the
> "official" style, so perhaps you should add a way to easily exclude
> some parts of the code from fish-indent, e.g.
> ```
> #fish_indent: stop
> ...
> #fish_indent: start
> ```
> And then all code in `...` will be left untouched by `fish_indent`.

yes, I'd prefer that over an editor-specific solution.
Maybe we should call it "fish_indent: on/off", to match clang-format

IsaacOscar:
> Consider adding fish-lsp to the recommended extension list.

I tried that (`ndonfris.fish-lsp`), as well as `gio00.fish`, `skyapps.fish-vscode`, and `lunaryorn.fish-ide`, they incorrectly highlight strings, e.g.
```fish
echo \"hello
world
```

Everything after the `\"` will be highlighted like a string literal. This is particularly bad as it means the rest of the file is wrong.
Moreover, `ndonfris.fish-lsp` was telling me that *every* line after after an `|| exit` was dead code. However now it doesn't even load, it's telling me that it can't find a node? module called `zod`.

I also tried `zgm.vscode-fish`: which will incorrectly end a string when it contains a `$(`, e.g.
```fish
echo "foo $(bar) baz"
```

Starting from the `$(`, everything in the line is no longer highlighted as a string literal (not even the `"` at the end).
In fact telling vs code my fish scripts are "shell code" was producing better syntax highlighting than any of those.

In short `bmalehorn.vscode-fish` was the only extension that could accurately determine where a string starts and ends. 
(It has some other minor bugs, and it would be nice if the `$(....)` but wasn't highlighted as a string literal).

Moreover, `bmalehorn.vscode-fish` and `lunaryorn.fish-ide` (which the former is based off I think) are the only ones that support `fish_indent`.

IsaacOscar:
> yes, I'd prefer that over an editor-specific solution. Maybe we should call it "fish_indent: on/off", to match clang-format

Maybe with a `#!` instead of just a `#`, indicating it's special?

Anyway, the `fish_indent` code looks too complicated for me to quickly add this feature (naturally it's walking an AST, so I guess it would be simplest to instead make it an annotation on begin end blocks? e.g.

```fish
begin #!fish_indent: skip
.... # text here is not modified
end
```
Then when `fish_indent` visits the begin block it would just output the text as is from the file. (Except that doesn't work as there's no `begin` AST node, only a `BeginHeader` node?).

krobelus:
On Thu, Sep 04, 2025 at 03:48:38AM -0700, Isaac Oscar Gariano wrote:
> IsaacOscar left a comment (fish-shell/fish-shell#11750)
> 
> > However, most of the tests have no reason to depart from the "official" style, so perhaps you should add a way to easily exclude some parts of the code from fish-indent, e.g. ``` #fish_indent: stop ... #fish_indent: start ``` And then all code in `...` will be left untouched by `fish_indent`.
> > yes, I'd prefer that over an editor-specific solution. Maybe we should call it "fish_indent: on/off", to match clang-format
> 
> Maybe with a `#!` instead of just a `#`, indicating it's special?
> 
> Anyway, the `fish_indent` code looks too complicated for me to quickly add this feature
> (naturally it's walking an AST, so I guess it would be simplest to instead make it an annotation on begin end blocks? e.g.

It's a pre-order AST traversal, so we can always do "check any (not
yet checked) comments until the start of the current node" to find
the directives that apply to the current source code segment.

We actually already have some special handling of comments, see 

	let comment_idx = gap_text.find(L!("#"));

in "emit_gap_text()"

I haven't tried this but I'm fairly confident we can check for lines
matching /#\h*fish_indent:\h*(on|off)\h*\n/ in this function,
and toggling formatting based on that.
While formatting is off, we can always emit the original text.

I strongly prefer baking this behavior into fish_indent rather than in
an external script or even a full blown LSP server, because fish_indent
is what users and editor integrations are using, and there's no reason
they should change.

Though I'm actually wondering if this is a desirable feature?
Is it just our test files, or is there a legitimate use for turning
off formatting?  If it's just our test files, we can find a simpler
solution, such as putting `# fish_indent: skip` somewhere,
for the whole file.

---

I'm also skeptical of recommending a fish-lsp that has some false
positives and doesn't use our own lexer/parser (maybe the tree-sitter
one is good enough these days, I'm not sure).  I do use fish-lsp
because I think it's the best one out there, but I'd not be comfortable
injecting it into fish dev's until futher investigation.  Either way,
that's gonna be a separate PR.

IsaacOscar:
> It's a pre-order AST traversal, so we can always do "check any (not yet checked) comments until the start of the current node" to find the directives that apply to the current source code segment. We actually already have some special handling of comments, see let comment_idx = gap_text.find(L!("#")); in "emit_gap_text()" I haven't tried this but I'm fairly confident we can check for lines matching /#\h*fish_indent:\h*(on|off)\h*\n/ in this function, and toggling formatting based on that. While formatting is off, we can always emit the original text.

I've come up with a really simple way of doing it, see my new pull request #11789, and the code here: https://github.com/fish-shell/fish-shell/pull/11789/commits/b2abde9c380899339f5810d787b6123675501774#diff-cd4de91716308d784ec1d7ce8d9207cb60b58d9009cd95f723453c50afb80dc5.

I've since updated this pull request so that it no longer excludes any directories from `fish_indent`.

> I strongly prefer baking this behavior into fish_indent rather than in an external script or even a full blown LSP server, because fish_indent is what users and editor integrations are using, and there's no reason they should change. 

Agreed

> Though I'm actually wondering if this is a desirable feature? Is it just our test files, or is there a legitimate use for turning off formatting? If it's just our test files, we can find a simpler solution, such as putting `# fish_indent: skip` somewhere, for the whole file.

There is one file I just put `#!fish_indent: off` at the top, which does this. But for most of the files it was only a small bit I think needed it.

Also lots of formatters have this feature:
* Rust has `[rustfmt::skip]` ... (which fish's code uses 24 times)
* python/black has `# fmt: off` ...  `# fmt: ob` (which fish's code uses once).
* clang has `// clang-format off` ...  `// clang-format on`
* C# has `#pragma warning disable format` ...  `#pragma warning enable format`
* Java `// @formatter:off` ... `// @formatter:on`

Those are just the ones I looked up now.
In particular I've often seen this done when you have large blocks of table-like information, e.g. you might find it more maintainable to write something like this:
```fish
foo  1     12
foo 12    345
```
With the nice alignment, which fish_indent would destroy.

Personally I never use autoformatters for my own code, so I'm not the best person to ask.

> --- I'm also skeptical of recommending a fish-lsp that has some false positives and doesn't use our own lexer/parser (maybe the tree-sitter one is good enough these days, I'm not sure). I do use fish-lsp because I think it's the best one out there, but I'd not be comfortable injecting it into fish dev's until further investigation. Either way, that's gonna be a separate PR.

Hmm, I could remove the suggestion for `bmalehorn.vscode-fish` and replace it with 
[Custom Local Formatters](https://marketplace.visualstudio.com/items?itemName=jkillian.custom-local-formatters) to get the automatic `fish_indent`, but no fish syntax highlighting?

krobelus:
> Hmm, I could remove the suggestion
> for `bmalehorn.vscode-fish` and replace it with
> [https://marketplace.visualstudio.com/items?itemName=jkillian.custom-local-formatters](Custom
> Local Formatters) to get the automatic `fish_indent`, but no fish
> syntax highlighting?

for syntax highlighting, a third party parser is fine, especially if
it's using treesitter

ndonfris:
Okay, will submit a separate PR at a later date, with additions to the documentation outside of fish-lsp vscode extension scope. 

My request for being included in the recommendations list, was mainly due to the disproportionate amount of bugs the vscode client has had with the server. **As I do not use vscode at all, and haven't for many years,** it is quite detrimental to the amount of time I can put into development on [the server](https://github.com/ndonfris/fish-lsp). It would be a game changer to have some more experienced vscode devs, helping contribute/maintain the [client extension](https://github.com/ndonfris/vscode-fish-lsp). 

> I tried that (ndonfris.fish-lsp), as well as gio00.fish, skyapps.fish-vscode, and lunaryorn.fish-ide, they incorrectly highlight strings, e.g.

There is very little syntax highlighting done by the server & client currently. This is mainly a side effect of the most text-editors already having a method to provide syntax highlighting to fish files, but will probably be implemented as a handler in the server eventually. If you instead wanted to make the client's (only vscode) syntax highlighting better, I have included a very minimal [fish.tmLanguage.json](https://github.com/ndonfris/vscode-fish-lsp/blob/master/syntaxes/fish.tmLanguage.json) which supports highlighting the [syntax `"$(...)"` as a nested command](https://github.com/ndonfris/vscode-fish-lsp/blob/b75d2e1e3dd7365b74d803b87cc602f959420c49/syntaxes/fish.tmLanguage.json#L72)) and didn't cause any immediately glaring issues when paired with [bmalehorn.fish-ide](https://github.com/bmalehorn/vscode-fish). Especially considering how stable my experience has been with tree-sitter (which handles this in most other language clients that I'm aware of), I recommend trying to map semantic token highlights to the tree-sitter language in the server itself instead.

That approach would allow any text-editor/ide that abides to the [lsp](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification) to provide highlights through the server itself. Theoretically, this should be mutually beneficial to the language server and fish shell itself because more user's can have access to a standard group of features accessible in fish shell wherever they want. Personally, I'd have thought that it'd make sense to recommend the server to fish dev's, so that vscode tooling can be as stable as most of the other supported text-editors.   

> I'm also skeptical of recommending a fish-lsp that has some false positives and doesn't use our own lexer/parser

I've been using tree-sitter with minimal issue for this specific project since before fish had started using rust. I will add some options to the next minor release that aim to make the more opinionated diagnostics and code-actions an [opt-in feature](https://github.com/ndonfris/fish-lsp/releases/tag/untagged-b3b17025d126d91641e5). 

Since the [tree-sitter-fish](https://github.com/ram02z/tree-sitter-fish) package is actively maintained, I disagree on avoiding its use. Also, the syntax highlighting issue related to `some_cmd || exit` have been fixed for a while now on the [nightly releases of the server](https://www.npmjs.com/package/fish-lsp?activeTab=versions), but I've run into several blockers trying to build the project for different vscode installations (which current pre-releases appear to [fix](https://github.com/ndonfris/vscode-fish-lsp/issues/10)). 

Diagnostics that cause false positives are mainly related to trying to provide [code-actions/quick-fixes](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_codeAction), by attempting to make higher level assumptions about fish source code (see fish-lsp's [argparse code-action](https://github.com/ndonfris/fish-lsp/blob/master/src/code-actions/argparse-completions.ts) for one example). While having the builtin parser for fish might be nicer for this, there would still likely need to be lots of abstractions above the parser for combining operations together. I do plan to add a lot more of quick-fixes but trying to get the VSCode release stable when I'm such a VSCode noob has been painful lol. 

Also, I do want to leave the possibility for connecting a language server/client connection to be accessible in a sandboxed browser. IMO this would likely bring tons of users from other shells over to fish, since they could quickly see what they're missing out on in fish. But its an extremely tricky problem. 

Tree-sitter's wasm implementation allows the server to not entirely require system access to provide most server features. Maybe, using napi or building fish's parser to wasm would be possible now but my previous experiments for something like this have yielded results that negatively impact how much code the server would has to maintain, especially in comparision to how the current wasm structure is used. I would be very interested in hearing how other's would attempt to achieve a web-editor though. 

I think it'd much more beneficial to fish devs to allow certain fish commands to have formatted output in a commonly parsable language like json, toml, yaml, etc. Commands I've come across that this would helpful for a text editor include: `fish --no-execute`, `fish_indent --dump-parse-tree`, `status *`, `fish_trace=1 fish_prompt`, or even `complete --do-complete`. While, I personally would use tree-sitter to achieve mapping [semantic highlights](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_semanticTokens) to [syntax nodes](https://github.com/tree-sitter/tree-sitter/blob/31ff62445bd017ddc4fed5547e5e75248f4649d6/lib/binding_web/src/index.ts#L18), I assume that a lot of editor extensions could use one of the previously listed languages, to parse output from `fish_indent`. We already have the flag `fish_indent --html`. If connecting tools to the internal parser is something prioritized, consider allowing something like: `fish_indent --json` for some of the previously listed commands.

## TL;DR

I don't use VSCode, which creates a maintenance bottleneck for fish-lsp's VSCode client while the server itself works other editors. Official recommendation would attract VSCode contributors to help maintain the client.

Fish's tree-sitter implementation is quite stable and is normally used in theme extensions IME. Syntax highlighting can be implemented in the server's semantic highlighting provider which would allow fish to have consistent colorization across most common text-editors/IDEs. 

I am open to suggestions for [improving the vscode extension](https://github.com/ndonfris/vscode-fish-lsp), [PR](https://github.com/ndonfris/vscode-fish-lsp/pull/11) for next [vsix pre-relase](https://github.com/ndonfris/vscode-fish-lsp/releases). Please review, so that [`fish-lsp@1.0.11`](https://github.com/ndonfris/fish-lsp/releases) can be published.

PR #11759, opened 2025-08-26
Remove `Optional` annotation

danielrainer:
The is inconsistent with the type annotation of `TestPass.duration_ms`. The function is only called with `duration_ms` as an int, so there is no need to declare it `Optional`.


zanchey:
Yep, thanks

I had a long think about the same in `TestFail` but `duration_ms` could end up being None there due to an exception, I guess?

Issue #8282, opened 2021-09-09
Centering text with fish shell...


snacksncode:
Fish: version 3.3.1.

This might sound silly but I'm trying to center a text inside fish shell. But I'm too smooth brain for that. I usually get command substitution errors when trying to get values. All I need is to combine (string length $text) with (math $COLUMNS / 2 + [That string length]) and pass it to printf. But for the past 1.5h I've been unsuccessful. Please help haha.

Example of what I'm trying to do:
```
printf "%*s\n" (math $COLUMNS / 2 + (string length "123456789012345678901234567890")) "123456789012345678901234567890"
```

Any ideas?

faho:
Math does operator precedence.

You're computing

    (columns / 2) + length

what you need is

    (columns + length) / 2

The simplest thing to do is probably to add another "/ 2" after the length argument:

```fish
printf "%*s\n" (math $COLUMNS / 2 + (string length "123456789012345678901234567890") / 2) "123456789012345678901234567890"
```

snacksncode:
Okay, thanks, you're right. The problem occurs when I'm trying to do this:
```fish
set text "ajdhkjad"; printf "%*s\n" (math $COLUMNS / 2 + (string length ($text)) / 2) $text
``` 
($text) gives me a "command substitution" error. But actually there's `fortune` command there instead of text that's why I need to store it in a variable (output changes each time you run `fortune`). How can I deal with that?

faho:
> ($text) gives me a "command substitution" error.

Yes, this is because `(something)` is a command substitution. You don't want the parentheses to be a command substitution. In fact in this particular bit you don't want them at all? `$text` is a variable not a command to run!

```fish
set text (fortune); printf "%*s\n" (math $COLUMNS / 2 + (string length $text) / 2) $text
```

However, fortune may print multiple lines, so you need to go through each line separately:

```fish
set text (fortune)
for line in $text
    printf "%*s\n" (math $COLUMNS / 2 + (string length $line) / 2) $line
end
```

Also, the length and columns might not be even, so you probably need to add `--scale=0` to your math invocation to just use an integer (`printf` won't be able to pad to 35.5):

```fish
set text (fortune)
for line in $text
    printf "%*s\n" (math --scale=0 $COLUMNS / 2 + (string length $line) / 2) $line
end
```

(of course if the text includes any leading or trailing spaces those will be used as well, add `| string trim` after the `fortune` to throw them away, if you wish)

[...]

PR #11763, opened 2025-08-28
Add a `-C/--center` option to string pad.

IsaacOscar:
## Description
This adds two really obvious feature that I was surprised was missing.
Bassically `string pad --center` will centre your strings (with some extra padding on the left if perfect centring isn't possible, and `string padd -center --right` will add the extra padding on the right).

"Fixes" issue #8282.

On a related note, I noticed that `string repeat` and `string shorten` have a `-N/--no-newline` option. Is there a reason that these are the only string subcommands with it? It would make sense for all of them except `string collect` and `string length`. The same effect can be achieved though by piping to string collect. I can add this option to all the string commands, it should be a quick change.

## TODOs:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -42,6 +42,7 @@ Deprecations and removed features
 Scripting improvements
 ----------------------
 - The ``psub`` command now allows combining ``--suffix`` with ``--fifo`` (:issue:`11729`).
+- The ``string pad`` command now has a ``-C/--center`` option.
 
 Interactive improvements
 ------------------------
--- a/doc_src/cmds/string-pad.rst
+++ b/doc_src/cmds/string-pad.rst
@@ -8,7 +8,7 @@ Synopsis
 
 .. synopsis::
 
-    string pad [-r | --right] [(-c | --char) CHAR] [(-w | --width) INTEGER]
+    string pad [-r | --right] [-C | --center] [(-c | --char) CHAR] [(-w | --width) INTEGER]
                [STRING ...]
 
 .. END SYNOPSIS
@@ -21,6 +21,8 @@ Description
 ``string pad`` extends each *STRING* to the given visible width by adding *CHAR* to the left. That means the width of all visible characters added together, excluding escape sequences and accounting for :envvar:`fish_emoji_width` and :envvar:`fish_ambiguous_width`. It is the amount of columns in a terminal the *STRING* occupies.
 
 The escape sequences reflect what fish knows about, and how it computes its output. Your terminal might support more escapes, or not support escape sequences that fish knows about.
+
+If **-C** or **--center** is given, add the padding to before and after the string. If it is impossible to perfectly center the result (because the required amount of padding is an odd number), extra padding will be added to the left, unless **--right** is also given.
 
 If **-r** or **--right** is given, add the padding after a string.
 
--- a/src/builtins/string/pad.rs
+++ b/src/builtins/string/pad.rs
@@ -5,6 +5,7 @@ pub struct Pad {
     char_to_pad: char,
     pad_char_width: usize,
     pad_from: Direction,
+    center: bool,
     width: usize,
 }
 
@@ -14,6 +15,7 @@ impl Default for Pad {
             char_to_pad: ' ',
             pad_char_width: 1,
             pad_from: Direction::Left,
+            center: false,
             width: 0,
         }
     }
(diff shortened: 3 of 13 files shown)


IsaacOscar:
There was one thing I wasn't sure about, specifically this [test case I added](https://github.com/fish-shell/fish-shell/pull/11763/files#diff-868e8792783c171778d4d411a46144044e17fe46526847cc6570cafde3366e41R86):

Assuming that 🐟 has width 2, what should this print?
```
string collect \|(string pad -w 3 -c 🐟 -C .)\|
string collect \|(string pad -w 7 -c 🐟 -C .)\|
```

Currently it prints:
```
| . |
|🐟 . 🐟|
```

But you might instead want it to print:
```
|🐟.|
|🐟🐟.🐟|
```

I.e. what's more important? The output is centred, or that the padding actually uses the given character and not spaces?

CHANGELOG.rst
```diff
@@ -42,6 +42,7 @@ Deprecations and removed features
 Scripting improvements
 ----------------------
 - The ``psub`` command now allows combining ``--suffix`` with ``--fifo`` (:issue:`11729`).
+- The ``string pad`` command now has a ``-C/--centre`` option. Unless ``-r/--right`` is also given, this will add a little extra padding on the left if it's not possible to perfectly centre (because the required padding is odd).
```
krobelus: maybe lose the second sentence, that sounds like too much info for the changelog

doc_src/cmds/string-pad.rst
```diff
@@ -8,7 +8,7 @@ Synopsis
 
 .. synopsis::
 
-    string pad [-r | --right] [(-c | --char) CHAR] [(-w | --width) INTEGER]
+    string pad [-r | --right] [-C | --centre] [(-c | --char) CHAR] [(-w | --width) INTEGER]
```
krobelus: I think we use American English (center)

src/builtins/string/pad.rs
```diff
@@ -83,19 +87,31 @@ impl StringSubCommand<'_> for Pad {
...
+                    .chain(spaces_chars(pad / 2))
                     .collect(),
-                Direction::Right => input
-                    .chars()
-                    .chain(repeat(' ').take(remaining_width))
-                    .chain(repeat(self.char_to_pad).take(pad))
-                    .collect(),
+                (Direction::Right, true) => {
+                    /*println!(">{}<", input);
+                    println!("pad_width = {}", pad_width);
```
krobelus: TODO

src/builtins/string/pad.rs
```diff
@@ -83,19 +87,31 @@
...
+                    println!("width = {}", width);
+                    println!("pad = {}", pad);
+                    println!("pad / 2 = {}", pad / 2);
+                    println!("pad - pad / 2 = {}", pad - pad / 2);*/
+
+                    chars_spaces(pad / 2)
+                        .chain(input.chars())
+                        .chain(spaces_chars(pad - pad / 2))
+                        .collect()
```
krobelus: I think this logic should be rewritten
to only mention each padding once;
then we also won't need the lambdas.
Also I'm not sure if the tuple match is better or worse than `if centre`

```rust
let left_padding;
let right_padding;
// compute paddings
if centre {
	match self.pad_from {
		...
	}
} else {
	match self.pad_from {
		...
	}
}
let padded = chars_spaces(left_padding)
	.chain(input.chars())
	.chain(spaces_chars(right_padding));
```

tests/checks/string.fish
```diff
@@ -43,48 +43,90 @@ string length -q ""; and echo not zero length; or echo zero length
...
 
     # Pad with multi-width character.
     string pad -w 3 -c 🐟 .
     # CHECK: 🐟.
+    # string pad would rather the result actually be centred, than it actually contain
+    # the padding character (so since it can't print half a 🐟, it instead prints a space which is half as wide)
+    string collect \|(string pad -w 3 -c 🐟 -C .)\|
+    # CHECK: | . |
+    string collect \|(string pad -w 7 -c 🐟 -C .)\|
+    # CHECK: |🐟 . 🐟|
```
krobelus, line 91: I have no strong opinion on this case.
Either behavior is fine, because I don't think anyone would use this with multi-width characters and strongly expect one or the other behavior.

src/builtins/string/pad.rs
IsaacOscar: How about more tuples?
```rust
let total_padding = pad_width - width;
let (left_padding, right_padding) = match (self.pad_from, self.centre) {
    (Direction::Left, false) => (total_padding, 0),
    (Direction::Right, false) => (0, total_padding),
    ... // and so on
}
```

IsaacOscar:
I've fixed all the issues you've noticed (minus the tuple matching, as I think the new version of the match is much neater).

PR #11779, opened 2025-09-03
reader: add case-insensitive history autosuggest

The0x539:
## Description

Fixes issue #3126

To match what I've been able to figure out about the design philosophy, case-sensitive matches still always take priority, but case-insensitive history suggestions precede case-insensitive completion suggestions.

## TODOs:

- [ ] ~~Changes to fish usage are reflected in user documentation/manpages.~~
- [ ] ~~User-visible changes noted in CHANGELOG.rst~~

### DIFF
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -83,7 +83,7 @@ use crate::highlight::{
 };
 use crate::history::{
     history_session_id, in_private_mode, History, HistorySearch, PersistenceMode, SearchDirection,
-    SearchType,
+    SearchFlags, SearchType,
 };
 use crate::input::init_input;
 use crate::input_common::{
@@ -151,6 +151,7 @@ use crate::tty_handoff::{
 };
 use crate::wchar::prelude::*;
 use crate::wcstringutil::string_prefixes_string_maybe_case_insensitive;
+use crate::wcstringutil::CaseSensitivity;
 use crate::wcstringutil::{
     count_preceding_backslashes, join_strings, string_prefixes_string,
     string_prefixes_string_case_insensitive, StringFuzzyMatch,
@@ -4696,7 +4697,6 @@ struct Autosuggestion {
     search_string_range: Range<usize>,
 
     // Whether the autosuggestion should be case insensitive.
-    // This is true for file-generated autosuggestions, but not for history.
     icase: bool,
 
     // Whether the autosuggestion is a whole match from history.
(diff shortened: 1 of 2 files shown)


src/reader.rs
```diff
@@ -4796,35 +4796,64 @@ fn get_autosuggestion_performer(
...
+
             while !ctx.check_cancel() && searcher.go_to_next_match(SearchDirection::Backward) {
                 let item = searcher.current_item();
 
-                // Suggest only a single line each time.
-                let matched_line = item
+                // The history item's may have multiple lines of text.
+                // Only suggest the line that actually contains the search string.
+
+                let lines = item
```
krobelus, line 4818: FWIW I got some upcoming changes that will allow multiline commands to be autosuggested 
(and if there is no full match, it will fall back to the existing line-by-line logic),
but that should be independent of your change.

src/reader.rs
```diff
@@ -4796,35 +4796,64 @@
...
+
+                let mut icase = false;
+                let mut matched_line = lines.clone().find(|line| line.starts_with(search_string));
+
+                // Only check for a case-insensitive match if we haven't already found one
+                if matched_line.is_none() && icase_history_result.is_none() {
+                    icase = true;
+                    matched_line = lines
+                        .into_iter()
+                        .find(|line| string_prefixes_string_case_insensitive(search_string, line));
```
krobelus, line 4833: It's kind of a shame that we have to match again, just to find
out the match quality.  I guess that's owed to the weirdness of
SearchType::LinePrefix, I'm not sure if there's a nice way to return
this information to the caller. Fine.

src/reader.rs
```diff
@@ -4796,35 +4796,64 @@
...
+                }
+
+                let Some(matched_line) = matched_line else {
+                    // The case-insensitive search should only fail if the history search yields a bogus result
+                    debug_assert!(icase_history_result.is_some());
```
krobelus: don't use debug_assert, use a real assert or an equivalent .unwrap().

src/reader.rs
```diff
@@ -4855,10 +4884,22 @@
...
+            if let Some(result) = icase_history_result {
+                return result;
+            }
             WString::new()
         } else {
             sort_and_prioritize(&mut completions, complete_flags);
             let comp = &completions[0];
+
+            // Prefer icase history over smartcase/icase completions.
+            if let (Some(result), CaseSensitivity::Smart | CaseSensitivity::Insensitive) =
```
krobelus, line 4899: This looks good overall, but I'm wondering about one case:
what if the history result is CaseSensitivity::Insensitive (e.g. FOO matching fOObar)
and the completion result is CaseSensitivity::Smart (e.g. foo matching the same)?
Do we really want to let history win, even though it's less accurate?

tests/pexpects/autosuggest.py
```diff
@@ -0,0 +1,74 @@
...
+use_suggestion()
+expect_prompt(">echo Hello\r\nHello")
+
+# case-correcting command suggestion
+send("Tru")
+use_suggestion(delay=2.0)
+expect_prompt(">true \r\n")
+
+# the motivating example: prefer icase history suggestions over icase completion suggestions
+run("mkdir -p Projects/myproject Projects/wrongproject")
```
krobelus, line 50: The existence of "Projects/wrongproject" suggests that this would be autosuggested
before the patch. But it's actually not about that,
only about making sure that "cd Projects/" file-based autosuggestion is not suggested
over history-based smartcase (or just icase?) autosuggestion,
so do we need this for anything? It's also listed last in lexicographic order,
so it wouldn't be autosuggested on `cd Projects/` either -- I'd say either find a use for it or remove it.

tests/pexpects/autosuggest.py
```diff
@@ -0,0 +1,74 @@
...
+
+run("cd ..")
+
+# (and now the actual test)
+run("cd Projects/myproject")
+run("cd ../..")
+
+send("cd pro")
+use_suggestion()
+expect_prompt(">cd Projects/myproject\r\n")
```
krobelus, line 65: nice

krobelus:
Also can you put the `Fixes ..` line and any other info into the commit message,
else squashing and taking it from the PR description is more work for me

src/reader.rs
The0x539: I considered this, but I can't think of a case where this is actually a meaningful difference *for a given query.*
Either the query is all-lowercase (and all non-samecase matches could be considered `Smart`), or it isn't (and all non-samecase matches could be considered `Insensitive`).

tests/pexpects/autosuggest.py
The0x539: See lines 52 and 55. Without a second project directory, the completion suggestions are capable of drilling down into the nested directories and immediately suggesting `cd Projects/myproject/` from `cd pro` without any prior navigation history.

src/reader.rs
The0x539: Agreed, but I figure this is a relatively brief and uncommon search anyway. Main improvement I can think of is including the case sensitivity type in the `HistoryItem`, but I wasn't confident about implementing that.

The0x539:
Assertion updated and commit message amended.

Issue #11457, opened 2025-05-04
Add option to `set` for unwrapping, i.e. splitting and assigning to two multiple variables


septatrix:
When working on some structured data or looping over text files it often comes up that one needs to assign values to multiple variables. One example can be seen below:

```fish
while read -l line
    set entry (string split ": " $line)
    set url $entry[2]
    set filename $entry[1]
    curl $url > $filename.csv
end <urls.txt
# Where urls.txt has the form "some name: https://example.com\n..."
```

This is a simplified example where it might be valid to simply use `$entry[n]` directly though for more complex scenarios this would reduce legibility. Similary, one could write `set url (string ...)` for each variable but this has disadvantages if the action is more expensive that a simple string split.

My proposal would be to allow `set` to receive multiple variable names and assign the tokens to them. Behavior when there are more/less variables than tokens can be to either fail, or assign empty values (if more variables than tokens) and make the last variable an array (if there are more tokens than variables. The choice here might also depend on which syntax is chosen.

Possible syntax options would be `set foo,bar 1 2`, `set [foo, bar] 1 2`, some special handling using `--` like `set --unpack foo bar -- 1 2` (the switch would be necessary for backwards compatibility I think...), or using multiple flags like `set --var foo --var bar 1 2`. Implementation wise the third option (using `--`) is likely simplest and I also think it would be the most convenient to use. The first two options, however, are quite close to how unpacking/unwrapping/destructuring is handled in many other programming languages (JS, Python, Rust)

krobelus:
> set --unpack foo bar -- 1 2

right, alternatively `set -m` / `set --many` with this syntax

IsaacOscar:
> looks like we can already do it without magic. I wonder if it's worth adding; we should probably find more use cases (in the fish repo).

Awesome, I'm using this now (Although I did get a syntax error because you're using a `{ ... }` block instead of `begin; ...; end').

Is there any particular reason an `-m` option isn't in fish yet? Or just that no ones got around to it yet?

IsaacOscar:
>```diff
> +        if test $flag = local
> +            if set -q _flag_local[1]
> +                set -a set_flags --function
>```

So you aren't actually setting a local variable from within a function, it's function scoped instead.
(E.g.
```
function foo
    set -l a outside
    begin
        let -l a -- inside
    end
    echo $a
end
foo
```
prints inside.
But if you change the `let` to a `set`, it prints outside.
So you can't set a local variable from within a function, even:
```
[...]

Issue #5946, opened 2019-06-19
Enhance argparse with option to set variable names.


AnthonyMonterrosa:
Small suggestion: would it be a reasonable request to ask that the variables set by `argparse ` have an option to be named? As in, instead of `_flag_X` for `x/X`, the input instead could be `x/X:variable_name` and then `variable_name` would hold the value of flag X? This would be nice if the name of the flag used by the user isn't the best name that could be used by the developer, because of different levels of understanding. Also would help reduce visual noise IMO, as the `_flag_` prefix feels unnecessary in most cases, at least in my use.

What do ya'll think?

faho:
If you want to rename a flag, then just duplicate it: `set -l variable_name $_flag_X`.

Adding more things onto the DSL makes it *more* confusing, not less, because there's more features to learn for dubious gain.

PR #11780, opened 2025-09-04
Variadic and strict `function` `--argument-names`

### LINKED ISSUE #11728: make functions defined in fish spit error on invalid argument count (suggestion/enhancement)
The fish `function` option `--argument-names` is very useful in making readable scripts, however, in shells scripts it's often easy to accidentally split arguments. In unfortunate cases it can lead to difficult to locate problems in larger scripts.
And that option does not spit any warnings or errors if the function is given more variables than specified.

I propose another option to `function` that makes it check the number of input arguments with the values passed to `--argument-names`; or a new option that gets an integer to check with said number of input arguments.

I have used fish functions to do something similar, but it's significantly more verbose than it needs to be:
```fish
function err --description 'prints the given error message and exits'
    set_color brred >&2
    printf "%s\n" "$argv" >&2
    set_color normal >&2
    exit 1
end

function assertArgNumber --description 'prints warning messages and exits if the calling function has a different number of args than specified' --argument-names real expected function
    if test "$real" -lt "$expected"
        err "less than expected $expected arguments to $function"
    else if test "$real" -gt "$expected"
        err "extraneous argument(s) to $function"
    end
end
```
And in each function in which I handle arguments with the `--argument-names` option, I use it like so:
```fish
function getCreationTime --description 'outputs the creation time of the given subvolume' --argument-names subvol
    assertArgNumber (count $argv) 1 (status function)

    echo "$(date --date $(sudo btrfs subvolume show "$subvol" | grep -oP 'Creation time:\s*\K\d.*$') +%s)"
end
```
The literal `1` passed to `assertArgNumber` is the number of arguments expected by the function, in this case, one.

Thank you for any input

IsaacOscar:
This pull request adds two new features (as discussed in issue #11728):
1. A *single* argument name in a `function`s `-a`/`--argument-names` list can end in `...`, the variable (with the `...` removed) will be set to a list containing as many arguments as possible, whilst still leaving arguments to be assigned to the rest of the parameters.
2. A new `-A`/`--strict-argument-names` option is added that works like `-a`/`--argument-names`, but you will get an error if you use the function with an incorrect number of arguments.

I also 'fixed' a few strange related behaviours that I noticed (see below).

I've documented everything and added test cases, however there is one bug (in my last commit): the error message produced by `-A` cannot be redirected away, e.g.:
```fish
function foo -A arg; end
foo &>/dev/null # Still prints an error message
```
However, this is consistent with the behaviour of calling a non-existent function/command.
Please let me know if you have any idea how to fix this.

# Details
Copied from the documentation for `function`, with changed behaviour in bold:

* `-a NAMES` or `--argument-names NAMES`
    Assigns the value of successive command-line arguments to the names given in *NAMES* (separated by spaces). These are the same arguments given in `argv`, and are still available there **(unless `--inherit-variable argv` was used or one of the given *NAMES* is `argv`)**. See also *Argument Handling*.

    **A single argument name given can end in `...`, in which case a variable number of arguments are saved in a variable with that name (excluding the `...` part), as many as possible whilst still leaving arguments for each of the other named arguments. If this cannot be done (because the number of arguments is smaller than the number of argument names), the argument with a `...` is set to the empty list, and the other arguments are asigned values as if the `...` one was not there.**

    See the *Argument Names Caveats* section below for what happens when the number of arguments passed differs from the number of argument *NAMES*.

*  `-A NAMES` or `--strict-argument-names NAMES`
    **This behaves like like `-a` / `--argument-names`, except that calling the function with an incorrect number of arguments will print an error, return 2, and not execute the body of the function.
    If none of the given *NAMES* use a `...`, this error will trigger when the number of actual arguments does not equal the number of *NAMES*.
    Otherwise, (if `...` was used) the error will trigger if the number of actual arguments is *less* than the number *NAMES* minus 1.**

    **Unlike the `-a`/`--argument-names` option, the given *NAMES* list can be empty, in which case an error will be generated if the function is called with a non-empty list of arguments.**

### Argument Names Caveats
**(this section is new, but some of it describes existing behaviour)**

The `-a` / `--argument-names` flag does *not* validate the number of arguments passed: if an argument is missing the corresponding variable will be assigned to an empty list. For example:

```fish
function two -a x y
    set -l
end
two 1
# prints:
#    argv 1
#    x 1
#    y
```

**In contrast, if the `-A` / `--strict-argument-names` option where used instead, the above call to `two` would produce an error.**

Similarly, if `-a` / `--argument-names` is used and none of the argument names end in `...`, any extra arguments are ignored, but they are still accessible in `$argv`. Continuing the previous example:

```fish
two 1 2 3
# prints:
#    argv '1'  '2'  '3'
#    x 1
#    y 2
```

**Using `-A` / `--strict-argument-names` will make the above call to `two` an error.**

**If on the other hand the last argument name does end in `...`, any extra arguments are stored in that variable. For example:**

```fish
function one_or_more -a x y...
    set -l
end
one_or_more 1 2 3
# prints:
#    argv '1'  '2'  '3'
#    x 1
#    y '2'  '3'
```

**The same behaviour occurs with the `-A` / `--strict-argument-names` option.**

**If the argument named with a `...` is not the last, then if possible, it will leave enough arguments for the subsequent argument names. For example:**

```fish
function more_or_one -a x... y
    set -l
end
more_or_one 1 2 3
# prints:
#    argv '1'  '2'  '3'
#    x '1'  '2'
#    y 3
```

**The argument with a `...` is however set to the empty list if there are not enough arguments to satisfy all the argument names. For example:**

```fish
function more_or_two -a x... y z
    set -l
end
more_or_two 1
# prints:
#    argv 1
#    x
#    y 1
#    z
```

**However, with the `-A` / `--strict-argument-names` option, the above call to `more_or_two` would produce an error.**

# The Commits
This first group of changes are not necessary for the rest of the functionality, so I'm happy to remove them if you don't like them:
1. f7f60b307f632224c5a9b7760b849e05b3571881  Use idiomatic Rust error handling for function builtin.
    > This simply does the same thing as #10948, but for the function builtin.
2. 6a7667dcf1cc6a9ff79f62e9cc1bf0dabd017113 Output function argument-names in one group.
    > This makes it so that printing a function definition will only use one
    > --argument-names group, instead of one for argument name.
    > For example, "function foo -a x y; ..." will print with "function foo
    > --argument-names x y" instead of "function foo --argument-names x
    > --argument-names y", which is very bizarre.
    > 
    > Moreover, the documentation no longer says that argument-names "Has to 
    > be the last option.". This sentence appears to have been introduced in 
    > error by pull #10524, since the ability to have options afterwards was 
    > deliberately added by pull #6188.

3. cd8c938f37464a8e02fbcfa968cf732b07cab8e9 Allow overwriting argv with function -a and -V
    > Previously, if you called a function parameter 'argv', within the body
    > of the function, argv would be set to *all* the arguments to the
    > function, and not the one indicated by the parameter name.
    > The same behaviour happened if you inherited a variable named 'argv'.
    > Both behaviours were quite surprising, so this commit makes things more
    > obvious, although they could alternatively simply be made errors.
4. c813ea2f4fdef3d52a4718a3446644ce93b48993 Prohibit duplicate variable names with function -a and -V options
    > It is now an error to use the same variable name twice with
    > --inherit-variable and/pr --argument-names.
    > The previous behaviour was:
    >     1. duplicate --inherit-variable's worked as if you only inherited
    >     the variable once
    >     2. Later --argument-names override earlier ones
    >     3. An --inherit-variable overrides an --argument-names
    > Thus this change is backwards incompatible, however, such duplicate
    > names are not useful and where likely unintentional, and none of the
    > test cases were depending on the old behaviour.
    > 
    > (I do note that there was code written to specifically remove any
    > duplicate --inherit-variable's, but the only difference this made was
    > that the duplicates wouldn't be displayed by the functions or type
    > builtin. This seems to have been done to preserve the prior behaviour
    > when it was a HashMap.)
    
The next commit adds a section to the documentation that the remaining commits add to:

5. 4ee163de274559c794afb97a9fede2f34820c63e Added more documentation for function --argument-names
    > Specifically, this adds a section to the "function" builtin
    > documentation explaining what happens when you call a function defined
    > with --argument-names, but with a different number of arguments than
    > argument-names.
    
6. 6ad3d9e5ea6514ff3e31c14c5471acbdc8c1e69b Allow variadic function --argument-names 
    > Specifically, this allows the last --argument-names in a function
    > definition to be suffixed with ... This causes any leftover arguments in
    > argv to be appended to the affixed variable name.
    > 
    > The documentation has been updating giving details on how different
    > numbers of argument names and argv elements are reconciled, with and
    > without the new ...

7. 181c8f9b9941733fdd8dcfab52dafa2589e9fa59 Do not require the variadic --argument-names to be the last one **(note: I made this a seperate commit in case it's too strange/useless)**
    > This commit makes it so that an argument-name ending with a ...
    > in a function's --argument-names list is not required to be the last.
    > There can still only be one such argument name.
    > The idea is that a '...' argument name will consume as many arguments
    > from argv as possible, whilst still leaving one argument for each normal
    > argument name (if possible).

8. 25a79d4e9958bef5505f2ab3855ae85ebc6b1c00 Add an -A/--strict-argument-names option to function 
    > 
    > The new -A/--strict-argument-names option behave like the existing
    > -a/--argument-names option to function, but with the following
    > differences:
    >   - If none of the argument names contains a ... suffix, it will be
    >   an error to call the function with a different number of arguments
    >   compared to the number of argument names.
    >   - If one of the argument names does contain a ... suffix, it will
    >   be an error to call it with less arguments, than non-... argument
    >   names.
    >   - it cannot be used together with -a/--argument-names (e.g.
    >   function f -A x -A y, and function f -a x -a y are allowed, but
    >   function f -A x -a y is not.)
    >   - It can be used without any argument argument names
    >   (e.g. function f -A), indicating that the function takes no arguments.
    > 
    > An error is produced as follows:
    >     - An error message of the following form is printed to stderr:
    > error: <name>: function requires exactly <n> arguments, but <m> where supplied
    >     - The function is not called
    >     - A status code of 2 is returned (similarly to other builtins
    >     when you provide them with an incorrect number/format of arguments)
    > 
    > Note that the error message is printed *before* the function is called,
    > so any redirections on the function call statement are not applied,
    > this is a bug that should be fixed.
    > 
    > This new feature fixes #11728.

# Future Stuff
If this pull request is accepted, I would like to add:
* Add `-a`/`--argument-names` and `-A`/`--strict-argument-names` to `argparse` (although I note similar suggestions were previously rejected in #5778 and #5946, but they didn't include my new `...` feature).
* Add `-m`/`--multiple` and `-M`/`--strict-multiple` to `set` (fixing #11457), that would be similar to `-a` and `-A`, the usage would be:
    ```fish
    set <options> -m <var>... -- <values>
    ```
    Where `<options>` are any existing set options that make sense. Each `<var>` part could be anything that set currently supports (e.g. `foo[1 3]`), or a name followed by `...`.

So let me know if you think any of my changes would not be suitable for `argparse` or `set`.

### DIFF
--- a/doc_src/cmds/function.rst
+++ b/doc_src/cmds/function.rst
@@ -27,6 +27,13 @@ The following options are available:
 
     See the :ref:`Argument Names Caveats <argument_names_caveats>` section below for what happens when the number of arguments passed differs from the number of argument *NAMES*.
 
+**-A** *NAMES* or **--strict-argument-names** *NAMES*
+    This behaves like like ``-a`` / ``--argument-names``, except that calling the function with an incorrect number of arguments will print an error, return 2, and not execute the body of the function.
+    If none of the given *NAMES* use a ``...``, this error will trigger when the number of actual arguments does not equal the number of *NAMES*.
+    Otherwise, (if ``...`` was used) the error will trigger if the number of actual arguments is *less* than the number *NAMES* minus 1.
+
+    Unlike the ``-a``/``--argument-names`` option, the given *NAMES* list can be empty, in which case an error will be generated if the function is called with a non-empty list of arguments.
+
 **-d** *DESCRIPTION* or **--description** *DESCRIPTION*
     A description of what the function does, suitable as a completion description.
 
@@ -153,7 +160,9 @@ The ``-a`` / ``--argument-names`` flag does *not* validate the number of argumen
     #    x 1
     #    y
 
-Similarly, if none of the argument names end in ``...``, any extra arguments are ignored, but they are still accessible in ``$argv``. Continuing the previous example:
+In contrast, if the ``-A`` / ``--strict-argument-names`` option where used instead, the above call to ``two`` would produce an error.
+
+Similarly, if ``-a`` / ``--argument-names`` is used and none of the argument names end in ``...``, any extra arguments are ignored, but they are still accessible in ``$argv``. Continuing the previous example:
 
 ::
 
@@ -162,6 +171,8 @@ Similarly, if none of the argument names end in ``...``, any extra arguments are
     #    argv '1'  '2'  '3'
     #    x 1
     #    y 2
+
+Using ``-A`` / ``--strict-argument-names`` will make the above call to ``two`` an error.
 
 If on the other hand the last argument name does end in ``...``, any extra arguments are stored in that variable. For example:
 
@@ -175,6 +186,8 @@ If on the other hand the last argument name does end in ``...``, any extra argum
     #    argv '1'  '2'  '3'
     #    x 1
     #    y '2'  '3'
+
+The same behaviour occurs with the ``-A`` / ``--strict-argument-names`` option.
 
 If the argument named with a ``...`` is not the last, then if possible, it will leave enough arguments for the subsequent argument names. For example:
 
@@ -202,6 +215,8 @@ The argument with a ``...`` is however set to the empty list if there are not en
     #    x
     #    y 1
     #    z
+
+However, with the ``-A`` / ``--strict-argument-names`` option, the above call to ``more_or_two`` would produce an error.
 
 Notes
 -----
--- a/doc_src/cmds/function.rst
+++ b/doc_src/cmds/function.rst
@@ -21,7 +21,7 @@ A function is a list of commands that will be executed when the name of the func
 The following options are available:
 
 **-a** *NAMES* or **--argument-names** *NAMES*
-    Has to be the last option. Assigns the value of successive command-line arguments to the names given in *NAMES* (separated by space). These are the same arguments given in :envvar:`argv`, and are still available there. See also :ref:`Argument Handling <variables-argv>`.
+    Assigns the value of successive command-line arguments to the names given in *NAMES* (separated by space). These are the same arguments given in :envvar:`argv`, and are still available there. See also :ref:`Argument Handling <variables-argv>`.
 
 **-d** *DESCRIPTION* or **--description** *DESCRIPTION*
     A description of what the function does, suitable as a completion description.
--- a/doc_src/cmds/function.rst
+++ b/doc_src/cmds/function.rst
@@ -23,6 +23,8 @@ The following options are available:
 **-a** *NAMES* or **--argument-names** *NAMES*
     Assigns the value of successive command-line arguments to the names given in *NAMES* (separated by spaces). These are the same arguments given in :envvar:`argv`, and are still available there (unless ``--inherit-variable argv`` was used or one of the given *NAMES* is ``argv``). See also :ref:`Argument Handling <variables-argv>`.
 
+    See the :ref:`Argument Names Caveats <argument_names_caveats>` section below for what happens when the number of arguments passed differs from the number of argument *NAMES*.
+
 **-d** *DESCRIPTION* or **--description** *DESCRIPTION*
     A description of what the function does, suitable as a completion description.
 
@@ -74,8 +76,6 @@ Example
 
 will run the ``ls`` command, using the ``-l`` option, while passing on any additional files and switches to ``ls``.
 
-
-
 ::
 
     function debug -a name val
@@ -99,8 +99,6 @@ will run the ``ls`` command, using the ``-l`` option, while passing on any addit
 
 will create a ``debug`` command to print chosen variables to `stderr`.
 
-
-
 ::
 
     function mkdir -d "Create a directory and set CWD"
@@ -120,7 +118,6 @@ will create a ``debug`` command to print chosen variables to `stderr`.
 This will run the ``mkdir`` command, and if it is successful, change the current working directory to the one just created.
 
 
-
 ::
 
     function notify
@@ -136,6 +133,33 @@ This will run the ``mkdir`` command, and if it is successful, change the current
 
 This will beep when the most recent job completes.
 
+.. _argument_names_caveats:
+
+Argument Names Caveats
+----------------------
+
+The ``-a`` / ``--argument-names`` flag does *not* validate the number of arguments passed: if an argument is missing the corresponding variable will be assigned to an empty list. For example:
+
+::
+
+    function two -a x y
+        set -l
+    end
+    two 1
+    # prints:
+    #    argv 1
+    #    x 1
+    #    y
+
+Similarly any extra arguments are ignored, but they are still accessible in ``$argv``. Continuing the previous example:
+
+::
+
+    two 1 2 3
+    # prints:
+    #    argv '1'  '2'  '3'
+    #    x 1
+    #    y 2
 
 Notes
 -----
--- a/doc_src/cmds/function.rst
+++ b/doc_src/cmds/function.rst
@@ -22,8 +22,8 @@ The following options are available:
 
 **-a** *NAMES* or **--argument-names** *NAMES*
     Assigns the value of successive command-line arguments to the names given in *NAMES* (separated by spaces). These are the same arguments given in :envvar:`argv`, and are still available there (unless ``--inherit-variable argv`` was used or one of the given *NAMES* is ``argv``). See also :ref:`Argument Handling <variables-argv>`.
-    
-    The last name given can end in ``...``, in which case all remaining arguments are saved in a variable with that name (excluding the ``...`` part).
+
+    A single argument name given can end in ``...``, in which case a variable number of arguments are saved in a variable with that name (excluding the ``...`` part), as many as possible whilst still leaving arguments for each of the other named arguments. If this cannot be done (because the number of arguments is smaller than the number of argument names), the argument with a ``...`` is set to the empty list, and the other arguments are asigned values as if the ``...`` one was not there.
 
     See the :ref:`Argument Names Caveats <argument_names_caveats>` section below for what happens when the number of arguments passed differs from the number of argument *NAMES*.
 
@@ -153,7 +153,7 @@ The ``-a`` / ``--argument-names`` flag does *not* validate the number of argumen
     #    x 1
     #    y
 
-Similarly, if the last argument name doesn't end in ``...``, any extra arguments are ignored, but they are still accessible in ``$argv``. Continuing the previous example:
+Similarly, if none of the argument names end in ``...``, any extra arguments are ignored, but they are still accessible in ``$argv``. Continuing the previous example:
 
 ::
 
@@ -176,6 +176,32 @@ If on the other hand the last argument name does end in ``...``, any extra argum
     #    x 1
     #    y '2'  '3'
 
+If the argument named with a ``...`` is not the last, then if possible, it will leave enough arguments for the subsequent argument names. For example:
+
+::
+
+    function more_or_one -a x... y
+        set -l
+    end
+    more_or_one 1 2 3
+    # prints:
+    #    argv '1'  '2'  '3'
+    #    x '1'  '2'
+    #    y 3
+
+The argument with a ``...`` is however set to the empty list if there are not enough arguments to satisfy all the argument names. For example:
+
+::
+
+    function more_or_two -a x... y z
+        set -l
+    end
+    more_or_two 1
+    # prints:
+    #    argv 1
+    #    x
+    #    y 1
+    #    z
 
 Notes
 -----
--- a/doc_src/cmds/function.rst
+++ b/doc_src/cmds/function.rst
@@ -22,6 +22,8 @@ The following options are available:
 
 **-a** *NAMES* or **--argument-names** *NAMES*
     Assigns the value of successive command-line arguments to the names given in *NAMES* (separated by spaces). These are the same arguments given in :envvar:`argv`, and are still available there (unless ``--inherit-variable argv`` was used or one of the given *NAMES* is ``argv``). See also :ref:`Argument Handling <variables-argv>`.
+    
+    The last name given can end in ``...``, in which case all remaining arguments are saved in a variable with that name (excluding the ``...`` part).
 
     See the :ref:`Argument Names Caveats <argument_names_caveats>` section below for what happens when the number of arguments passed differs from the number of argument *NAMES*.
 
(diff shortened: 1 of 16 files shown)


doc_src/cmds/function.rst
```diff
@@ -21,7 +21,18 @@ A function is a list of commands that will be executed when the name of the func
 The following options are available:
 
 **-a** *NAMES* or **--argument-names** *NAMES*
-    Has to be the last option. Assigns the value of successive command-line arguments to the names given in *NAMES* (separated by space). These are the same arguments given in :envvar:`argv`, and are still available there. See also :ref:`Argument Handling <variables-argv>`.
```
krobelus, line 24: I usually do one-sentence-per line to reduce noise in line-based diffs.
Maybe that's an anachronistic perspective, there might be disadvantages to one-sentence-per line, I'm not sure

doc_src/cmds/function.rst
```diff
@@ -21,7 +21,18 @@
...
+    Assigns the value of successive command-line arguments to the names given in *NAMES* (separated by spaces). These are the same arguments given in :envvar:`argv`, and are still available there (unless ``--inherit-variable argv`` was used or one of the given *NAMES* is ``argv``). See also :ref:`Argument Handling <variables-argv>`.
```
krobelus, line 24: My first reaction to cd8c938f374 (Allow overwriting argv with function -a and -V, 2025-08-22)
was: yes, we *can* do this, but is it actually useful?
If it's a user error 90% of the time, then we should rather throw an error when creating the function.

I guess "--argument-names head argv" can be an elegant way to write recursive functions,
it communicates early that we don't care about the full $argv anymore.
So this looks fine to me, same for -V.

doc_src/cmds/function.rst
```diff
@@ -21,7 +21,18 @@
...
+
+    A single argument name given can end in ``...``, in which case a variable number of arguments are saved in a variable with that name (excluding the ``...`` part), as many as possible whilst still leaving arguments for each of the other named arguments. If this cannot be done (because the number of arguments is smaller than the number of argument names), the argument with a ``...`` is set to the empty list, and the other arguments are asigned values as if the ``...`` one was not there.
+
```
krobelus, line 27: > the error message produced by -A cannot be redirected away, e.g.:
 > 
 > function foo -A arg; end
 > foo &>/dev/null # Still prints an error message
 > However, this is consistent with the behaviour of calling a non-existent function/command.

huh this works in bash, but not in fish.
It also affects the time keyword (also in bash), for better or worse.

I'm not sure if we should change the behavior (for non-existent commands),
but the way it works is that fish has an IoChain passed along as execution context.
Before fish spawns a process, it resolves the IoChain to file descriptors,
and passes those to the forked process (or to the builtin function).

doc_src/cmds/function.rst
```diff
@@ -21,7 +21,18 @@
...
+    See the :ref:`Argument Names Caveats <argument_names_caveats>` section below for what happens when the number of arguments passed differs from the number of argument *NAMES*.
+
+**-A** *NAMES* or **--strict-argument-names** *NAMES*
+    This behaves like like ``-a`` / ``--argument-names``, except that calling the function with an incorrect number of arguments will print an error, return 2, and not execute the body of the function.
```
krobelus, line 31: Hmm, I'm not yet 100% sure if we should add this feature.
The weird part here is that it means that function calls
may print an error before entering the actual function body.
I don't know if it's an actual problem in practice, but
function authors might want control of their error messages.
But of course, they don't need to use this flag.
And everyone is fine with using argparse too..

I think another potential problem with growing `function` options like this is that it will prevent some people from discovering (or taking the plunge into) argparse,
which seems strictly superior (we might need to add `--argument-names`).

When I write CLI tools (mainly POSIX shell scripts), I often use
environment variables, or positional arguments because I'm too lazy to
figure out how to use getopts. I think it's much easier with `argparse`,
and that leads to nicer CLIs.

I think @Elec3137's concern in https://github.com/fish-shell/fish-shell/issues/11728 about argparse was that they didn't want to specify args redundantly. That we can achieve with argparse

	function getCreationTime --description 'outputs the creation time of the given subvolume'
		argparse --positionals subvol -- $argv
		or return
		...
	end


Note that they didn't use `return` but `exit 1` (i.e. exit the whole shell, or, if inside `source`, exit the current file).
If they actually want exit semantics rather than only from returning from the current function,
then this `--strict-argument-names` feature wouldn't help.
So let's make sure we have a sound and complete usage example.

I'd argue that the argparse example above is pretty nice;
explicit return is how things are currently done,
and the separate line doesn't hurt. Not sure.

 > So let me know if you think any of my changes would not be suitable for argparse or set.

Ok, let's think about that.

Either way, so far I'm happy with all the commits leading up to the last two.
Can you drop these from this PR, and maybe move them to separate ones? (if there are conflicts between the two, then one PR is fine)

doc_src/cmds/function.rst
```diff
@@ -136,6 +142,81 @@
 
 This will beep when the most recent job completes.
 
+.. _argument_names_caveats:
+
+Argument Names Caveats
+----------------------
+
+The ``-a`` / ``--argument-names`` flag does *not* validate the number of arguments passed: if an argument is missing the corresponding variable will be assigned to an empty list. For example:
```
krobelus, line 150: > the corresponding variable will be assigned to an empty list.

I think it should be "will be assigned to empty list", (no "to"),
or "an empty list will be assigned to the variable"

doc_src/cmds/function.rst
```diff
@@ -136,6 +142,81 @@
...
+::
+
+    function more_or_one -a x... y
+        set -l
+    end
+    more_or_one 1 2 3
+    # prints:
+    #    argv '1'  '2'  '3'
+    #    x '1'  '2'
+    #    y 3
```
krobelus, line 203: I'm not sure if it's wise to allow variadic arguments before the last argument.
Yes, we can squeeze this feature in without breakage,
but I can't think of a reasonable use case.
It seems like using it to make CLIs will invite user errors such as forgetting to pass the last argument (y) and notice unnecessarily late because x... concedes an argument to y...

I think before we potentially add this commit,
we should find a compelling use case (I haven't read this commit in detail, maybe I missed it).

src/builtins/function.rs
```diff
@@ -77,37 +122,30 @@ fn parse_cmd_opts(
     argv: &mut [&wstr],
     parser: &Parser,
     streams: &mut IoStreams,
-) -> c_int {
+) -> BuiltinResult {
```
krobelus, line 125: > This simply does the same thing as #10948, but for the function builtin.

FWIW, I personally prefer to see commit references (e.g. "git log
-1 --pretty=reference") rather than issue/PR numbers, mainly because
then I don't need GitHub-specific tools to follow the reference.

src/builtins/function.rs
```diff
@@ -77,37 +122,30 @@
...
-                    if is_read_only(&woptarg) {
-                        streams.err.append(wgettext_fmt!(
-                            "%ls: variable '%ls' is read-only\n",
-                            cmd,
-                            woptarg
-                        ));
-                        return STATUS_INVALID_ARGS;
-                    }
-                    opts.named_arguments.push(woptarg);
+                    proccess_argument_name(opts, woptarg, streams)?;
```
krobelus, line 141: I haven't read the context, but maybe a better name for this function is `add_argument_name()`, to make it obvious we're adding to opts.
(unless we fail, but that's also obvious).

It's fine to leave, your call

src/builtins/function.rs
```diff
@@ -295,75 +346,90 @@
...
 
     // Ensure inherit_vars is unique and then populate it.
-    opts.inherit_vars.sort_unstable();
-    opts.inherit_vars.dedup();
+    //let inherit_vars: HashSet<WString> = opts.inherit_vars.drain(..).collect();
+    let mut inherit_vars = HashSet::new();
+    for named in opts.inherit_vars.drain(..) {
+        if !inherit_vars.insert(named.clone()) {
+            streams.err.append(wgettext_fmt!(
+                "%ls: variable '%ls' is inherited multiple times\n",
```
krobelus, line 386: the name might be misleading it first because we (obviously) refuse to have it inherit anything in this case.
Not sure if there is a better way to phrase it. Maybe

	%ls: cannot inherit variable '%ls' multiple times
	%ls: duplicate variable '%ls' in --inherit-variable

The second one would be consistent with --argument-names, but -V arguments are disjoint, so it's not the same.
Not sure. Your call

src/builtins/functions.rs
```diff
@@ -419,11 +419,8 @@ pub fn functions(parser: &Parser, streams: &mut IoStreams, args: &mut [&wstr]) -
         let mut def = WString::new();
 
         if !comment.is_empty() {
-            def.push_utfstr(&sprintf!(
-                "# %ls\n%ls",
-                comment,
-                props.annotated_definition(arg)
-            ));
+            let defy = props.annotated_definition(arg);
+            def.push_utfstr(&sprintf!("# %ls\n%ls", comment, defy,));
```
krobelus, line 423: that's a slightly weird variable name (though it seems okay) and a bad trailing comma

tests/checks/function.fish
```diff
@@ -186,4 +186,432 @@ function foo; echo before; end
...
+
+# Check for errors with duplicate names
+set var 1
+function bad -a var var -V var
+    echo $var
+end
+#CHECKERR: {{.*}}checks/function.fish (line {{\d+}}): function: variable 'var' is passed to both --argument-names and --inherit-variable
+#CHECKERR: function bad -a var var -V var
+#CHECKERR: ^
+not functions -q bad || echo "function should not exist"
```
krobelus, line 220: FWIW, this originally called "bad" in c813ea2f4fd (Prohibit duplicate
variable names with function -a and -V options, 2025-08-22), which we
shouldn't to because a fish dev might have a command with that name in
`$PATH`. The `not functions -q bad` check looks perfect; ideally
it should have been squashed to make it easier to review

doc_src/cmds/function.rst
krobelus: we can find how to do this later, if we actually keep this approach

krobelus:
cherry-picking the first 3 commits, still want some of the others

doc_src/cmds/function.rst
zanchey: > I think another potential problem with growing `function` options like this is that it will prevent some people from discovering (or taking the plunge into) argparse,
> which seems strictly superior (we might need to add `--argument-names`).

I am in agreement with keeping argparse as our preferred approach

PR #11785, opened 2025-09-08
config_paths: fix feature-flag bitrot and smaller issues

krobelus:
Nothing too interesting, feel free to ignore.

### DIFF
--- a/.github/workflows/rust_checks.yml
+++ b/.github/workflows/rust_checks.yml
@@ -18,6 +18,9 @@ jobs:
 
   clippy-stable:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        features: ["", "--no-default-features"]
     steps:
     - uses: actions/checkout@v3
     - uses: ./.github/actions/rust-toolchain@stable
@@ -27,7 +30,7 @@ jobs:
       run: |
         sudo apt install gettext
     - name: cargo clippy
-      run: cargo clippy --workspace --all-targets -- --deny=warnings
+      run: cargo clippy --workspace --all-targets ${{ matrix.features }} -- --deny=warnings
 
   clippy-msrv:
     runs-on: ubuntu-latest
--- a/src/env/config_paths.rs
+++ b/src/env/config_paths.rs
@@ -4,6 +4,8 @@ use crate::wchar::prelude::*;
 use crate::{common::get_executable_path, FLOG, FLOGF};
 use std::ffi::OsString;
 use std::os::unix::ffi::OsStringExt;
+#[cfg(not(feature = "embed-data"))]
+use std::path::Path;
 use std::path::PathBuf;
 
 /// A struct of configuration directories, determined in main() that fish will optionally pass to
@@ -39,8 +41,8 @@ pub struct ConfigPathDetection {
     exec_path: PathBuf,
 }
 
+#[cfg(not(feature = "embed-data"))]
 impl ConfigPaths {
-    #[cfg(not(feature = "embed-data"))]
     fn static_paths() -> Self {
         Self {
             sysconf: PathBuf::from(SYSCONF_DIR).join("fish"),
@@ -48,6 +50,18 @@ impl ConfigPaths {
             data: PathBuf::from(env!("DATADIR")).join("fish"),
             doc: DOC_DIR.into(),
             locale: PathBuf::from(env!("LOCALEDIR")),
+        }
+    }
+
+    fn from_source_tree(repo_root: &Path, bin: PathBuf) -> Self {
+        Self {
+            sysconf: repo_root.join("etc"),
+            bin: Some(bin),
+            data: repo_root.join("share"),
+            // This requires manual sphinx invocation.
+            doc: repo_root.join("user_doc/html"),
+            // This requires manual build_tools/update_translations.fish invocation.
+            locale: repo_root.join("share/locale"),
         }
     }
 }
@@ -107,15 +121,10 @@ impl ConfigPathDetection {
         // If we're in Cargo's target directory or CMake's build directory, use the source files.
         if exec_path.starts_with(env!("FISH_BUILD_DIR")) {
             let manifest_dir = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+            let paths = ConfigPaths::from_source_tree(&manifest_dir, bin.to_owned());
             return Self {
-                kind: BuildDirectory(manifest_dir.clone()),
-                paths: ConfigPaths {
-                    sysconf: manifest_dir.join("etc"),
-                    bin: Some(bin.to_owned()),
-                    data: manifest_dir.join("share"),
-                    doc: manifest_dir.join("user_doc/html"),
-                    locale: manifest_dir.join("share/locale"),
-                },
+                kind: BuildDirectory(manifest_dir),
+                paths,
                 exec_path,
             };
         }
@@ -144,24 +153,12 @@ impl ConfigPathDetection {
                     exec_path,
                 };
             }
-        } else {
-            let base_path = bin;
-            let data = base_path.join("share");
-            let sysconf = base_path.join("etc");
-            if data.exists() && sysconf.exists() {
-                let locale = data.join("locale");
-                return Self {
-                    kind: InTreeBuild,
-                    paths: ConfigPaths {
-                        sysconf,
-                        bin: Some(base_path.to_path_buf()),
-                        data,
-                        doc: base_path.join("user_doc/html"),
-                        locale,
-                    },
-                    exec_path,
-                };
-            }
+        } else if bin.join("data").exists() && bin.join("etc").exists() {
+            return Self {
+                kind: InTreeBuild,
+                paths: ConfigPaths::from_source_tree(bin, bin.to_owned()),
+                exec_path,
+            };
         }
 
         Self {
--- a/src/env/config_paths.rs
+++ b/src/env/config_paths.rs
@@ -1,14 +1,18 @@
-use super::ConfigPaths;
 use crate::{common::get_executable_path, FLOG, FLOGF};
 use once_cell::sync::Lazy;
+use std::path::Path;
 use std::path::PathBuf;
 
-const DOC_DIR: &str = env!("DOCDIR");
-const DATA_DIR: &str = env!("DATADIR");
-const DATA_DIR_SUBDIR: &str = env!("DATADIR_SUBDIR");
-const SYSCONF_DIR: &str = env!("SYSCONFDIR");
-const LOCALE_DIR: &str = env!("LOCALEDIR");
-const BIN_DIR: &str = env!("BINDIR");
+/// A struct of configuration directories, determined in main() that fish will optionally pass to
+/// env_init.
+#[derive(Default)]
+pub struct ConfigPaths {
+    pub data: Option<PathBuf>,   // e.g., /usr/local/share
+    pub sysconf: PathBuf,        // e.g., /usr/local/etc
+    pub doc: PathBuf,            // e.g., /usr/local/share/doc/fish
+    pub bin: Option<PathBuf>,    // e.g., /usr/local/bin
+    pub locale: Option<PathBuf>, // e.g., /usr/local/share/locale
+}
 
 pub static CONFIG_PATHS: Lazy<ConfigPaths> = Lazy::new(|| {
     // Read the current executable and follow all symlinks to it.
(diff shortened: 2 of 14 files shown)


danielrainer:
I looked at the changes in this PR a bit and generally really like the cleanup.

At first I was a bit surprised about some changes, but then I realized that the intention seems to be that `embed-data` builds are not expected to access any external files. Is this the way fish builds are intended to work going forward? It does seem to simplify things, but has the downside that any features which require non-embedded files will then be unavailable to fish built with `embed-data`.

Merging this before the updates to gettext which allow localization with `embed-data` builds (I haven't gotten around to updating that PR yet), breaks localization with such builds, whereas they do work on current master (if MO files are in the repo's `share/locale` directory). Breaking that is not a big deal, since localization would break when the executable gets moved out of the repo, and it will be fixed once embedded gettext is working, but I think it is good to be aware of it. If the gettext changes are merged before this, the issue fully disappears.

src/env/config_paths.rs
```diff
@@ -1,154 +1,237 @@
...
-            paths = ConfigPaths {
-                data: Some(manifest_dir.join("share")),
-                sysconf: manifest_dir.join("etc"),
-                doc: manifest_dir.join("user_doc/html"),
-                bin: Some(exec_path.parent().unwrap().to_owned()),
-                locale: Some(manifest_dir.join("share/locale")),
-            }
+#[cfg(not(feature = "embed-data"))]
+enum ConfigPathKind {
+    BuildDirectory(/*manifest_dir=*/ PathBuf),
```
danielrainer, line 30: I get why you chose this naming and it does make sense, but it has the problem of inviting confusion since the "build" directory could easily be interpreted as `build/` or `target/`, which is probably why you also added the comment. `ManifestDirectory` might be a better choice. `RepoRootDirectory` could also work, as used in `ConfigPaths::from_source_tree`. Not sure about the best choice here, but I'd really like to have the info in the enum variant name instead of a comment which is much easier to miss. In any case, it would make sense to decide on what we want to call the directory and use it consistently throughout the codebase.

Another thing I noticed thinking about this is that we decide which paths to consider for data files based on fish's exec path, choosing the files in the repo if the exec path points into the repo. This works fine if using Cargo's default `target`, or a `build` directory inside the repo with CMake, but breaks if the build artifacts are located outside of the repo, either by using the `--target-dir` flag with Cargo or by letting CMake use a build dir that's not a subdir of the repo. I don't know if we want to change the behavior there. Fish could still try to access files from the repo in such scenarios, but it would have to check if the repo is still available, and it would be hard to decide whether to prefer files from the repo or in the installed location if both are available. Probably it's not worth the trouble to worry about such builds.

.github/workflows/rust_checks.yml
```diff
@@ -18,6 +18,9 @@ jobs:
 
   clippy-stable:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        features: ["", "--no-default-features"]
```
krobelus, line 23: > Is this the way fish builds are intended to work going forward? It
 > does seem to simplify things, but has the downside that any features
 > which require non-embedded files will then be unavailable to fish
 > built with `embed-data`.

I don't think there is a strict rule against accessing external files
but locale seems like something we should embed.  (Obviously we'd
prefer using something like gettext-rs that's not specific to fish
but until they support embedding, we might as well do it ourselves.)

 > Merging this before the updates to gettext which allow localization
 > with `embed-data` builds

right, I forgot that these sometimes work

 > Breaking that is not a big deal, since localization
 > would break when the executable gets moved out of the repo,

yeah, outside the target or build dir, we use .local/share/locale
(or, with the current PR, /usr/local/share/locale).  Neither make a
lot of sense for embedded builds, and the former basically never works.

We can definitely keep the non-broken case (embedded from target or
build dir) for now, that will be easy to revert:

```diff
diff --git a/src/env/config_paths.rs b/src/env/config_paths.rs
index 60f8e360db..609645804e 100644
--- a/src/env/config_paths.rs
+++ b/src/env/config_paths.rs
@@ -17,8 +17,7 @@
     pub data: PathBuf, // e.g., /usr/local/share
     #[cfg(not(feature = "embed-data"))]
     pub doc: PathBuf, // e.g., /usr/local/share/doc/fish
-    #[cfg(not(feature = "embed-data"))]
-    pub locale: PathBuf, // e.g., /usr/local/share/locale
+    pub locale: Option<PathBuf>, // e.g., /usr/local/share/locale
 }
 
 const SYSCONF_DIR: &str = env!("SYSCONFDIR");
@@ -49,7 +48,7 @@
             bin: Some(PathBuf::from(env!("BINDIR"))),
             data: PathBuf::from(env!("DATADIR")).join("fish"),
             doc: DOC_DIR.into(),
-            locale: PathBuf::from(env!("LOCALEDIR")),
+            locale: Some(PathBuf::from(env!("LOCALEDIR"))),
         }
     }
 
@@ -61,7 +60,7 @@
             // This requires manual sphinx invocation.
             doc: repo_root.join("user_doc/html"),
             // This requires manual build_tools/update_translations.fish invocation.
-            locale: repo_root.join("share/locale"),
+            locale: Some(repo_root.join("share/locale")),
         }
     }
 }
@@ -71,12 +70,10 @@
 
 /// NOTE: This is called during early startup, do not log anything.
 pub fn init_locale_dir(argv0: &wstr) -> ConfigPathDetection {
-    #[allow(clippy::let_and_return)] // for old clippy
     let detection = ConfigPathDetection::new(argv0);
-    #[cfg(not(feature = "embed-data"))]
-    {
+    if let Some(locale) = detection.paths.locale.as_ref() {
         use std::os::unix::ffi::OsStrExt;
-        let bytes = detection.paths.locale.as_os_str().as_bytes();
+        let bytes = locale.as_os_str().as_bytes();
         assert!(!bytes.is_empty());
         let bytes: Box<[u8]> = Box::from(bytes);
         let bytes: Box<&'static [u8]> = Box::new(Box::leak(bytes));
@@ -98,6 +95,10 @@
             paths: ConfigPaths {
                 sysconf: PathBuf::from(SYSCONF_DIR).join("fish"),
                 bin: exec_path.parent().map(|x| x.to_path_buf()),
+                locale: exec_path.starts_with(env!("FISH_BUILD_DIR")).then(|| {
+                    let repo_root = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+                    repo_root.join("share/locale")
+                }),
             },
             exec_path,
         }
@@ -148,7 +149,7 @@
                         } else {
                             PathBuf::from(DOC_DIR)
                         },
-                        locale: base_path.join("share/locale"),
+                        locale: Some(base_path.join("share/locale")),
                     },
                     exec_path,
                 };
@@ -227,11 +228,14 @@
         FLOGF!(config, "paths.data: %ls", paths.data.display().to_string());
         #[cfg(not(feature = "embed-data"))]
         FLOGF!(config, "paths.doc: %ls", paths.doc.display().to_string());
-        #[cfg(not(feature = "embed-data"))]
         FLOGF!(
             config,
             "paths.locale: %ls",
-            paths.locale.display().to_string()
+            paths
+                .locale
+                .clone()
+                .map(|x| x.display().to_string())
+                .unwrap_or("|not found|".to_string()),
         );
     }
 }
diff --git a/src/wutil/gettext.rs b/src/wutil/gettext.rs
index 10ace96cc6..bb6b648cc5 100644
--- a/src/wutil/gettext.rs
+++ b/src/wutil/gettext.rs
@@ -14,19 +14,15 @@
     use std::ffi::CStr;
     extern "C" {
         fn gettext(msgid: *const c_char) -> *mut c_char;
-        #[cfg(not(feature = "embed-data"))]
         fn bindtextdomain(domainname: *const c_char, dirname: *const c_char) -> *mut c_char;
-        #[cfg(not(feature = "embed-data"))]
         fn textdomain(domainname: *const c_char) -> *mut c_char;
     }
     pub fn fish_gettext(msgid: &CStr) -> *const c_char {
         unsafe { gettext(msgid.as_ptr()) }
     }
-    #[cfg(not(feature = "embed-data"))]
     pub fn fish_bindtextdomain(domainname: &CStr, dirname: &CStr) -> *mut c_char {
         unsafe { bindtextdomain(domainname.as_ptr(), dirname.as_ptr()) }
     }
-    #[cfg(not(feature = "embed-data"))]
     pub fn fish_textdomain(domainname: &CStr) -> *mut c_char {
         unsafe { textdomain(domainname.as_ptr()) }
     }
@@ -38,11 +34,9 @@
     pub fn fish_gettext(msgid: &CStr) -> *const c_char {
         msgid.as_ptr()
     }
-    #[cfg(not(feature = "embed-data"))]
     pub fn fish_bindtextdomain(_domainname: &CStr, _dirname: &CStr) -> *mut c_char {
         std::ptr::null_mut()
     }
-    #[cfg(not(feature = "embed-data"))]
     pub fn fish_textdomain(_domainname: &CStr) -> *mut c_char {
         std::ptr::null_mut()
     }
@@ -52,16 +46,13 @@
 
 // Really init wgettext.
 fn wgettext_really_init() {
-    #[cfg(not(feature = "embed-data"))]
-    {
-        use crate::common::PACKAGE_NAME;
-        use crate::env::config_paths::LOCALE_DIR;
-        use std::ffi::CString;
+    use crate::common::PACKAGE_NAME;
+    use crate::env::config_paths::LOCALE_DIR;
+    use std::ffi::CString;
 
+    let localedir = LOCALE_DIR.load();
+    if !localedir.is_empty() {
         let package_name = CString::new(PACKAGE_NAME).unwrap();
-        let localedir = LOCALE_DIR.load();
-        #[cfg(not(test))]
-        assert!(!localedir.is_empty());
         let localedir = CString::new(localedir).unwrap();
         fish_bindtextdomain(&package_name, &localedir);
         fish_textdomain(&package_name);
```

src/env/config_paths.rs
krobelus: I'd prefer calling it "source tree" rather than "repo" because in this context,
a tarball / git-export would work too, it doesn't depend on any version control system.
The corresponding path name might be called "source directory" but that might be confused with `src/`.
"source root" doesn't really work either.

So I'm fine with "repo_root" (replacing "manifest_dir" and others).

I'll remove the path from the enum, since the manifest dir is a global constant anyway.

> but breaks if the build artifacts are located outside of the repo, either
> by using the `--target-dir` flag with Cargo or by letting CMake use a
> build dir that's not a subdir of the repo

I think we definitely want this to work; It should already work for CMake
builds, because we do `FISH_BUILD_DIR=${CMAKE_BINARY_DIR}`.  So as long as
fish is inside `${CMAKE_BINARY_DIR}`, it will always use the files from the
source tree, which need not be a parent of `${CMAKE_BINARY_DIR}`.

Docker tests also rely on this (the source tree is read-only).
They do it by setting `CARGO_TARGET_DIR` which becomes `FISH_BUILD_DIR`.

Unfortunately I missed that cargo only reads `CARGO_TARGET_DIR` but does
not set it (or anything directly equivalent).
So yeah, this is broken for now, would be nice to fix:

	cargo b --target-dir ~/target --no-default-features &&
	    ~/target/debug/fish -d config -c:
	config: Unexpected directory layout, using compiled-in paths

> hard to decide whether to prefer files from the repo or in the installed location

Currently it's "if inside build directory, use that, else assume we're installed (i.e. RelocatableTree)".

This means that this installation is broken:

	(
		mkdir build
		cd build
		cmake .. -DCMAKE_INSTALL_PREFIX=$PWD/install
		ninja install
	)
	build/install/fish # will NOT use installed files

we could probably fix that by preferring "RelocatableTree".
A legit build directory should not be mistaken as relocatable tree,
unless it's called "bin/" *AND* has a sibling called "share/".
Since share exists at root level, we could explictly prevent the repo root
from being treated as installed location.

```diff
diff --git a/src/env/config_paths.rs b/src/env/config_paths.rs
index 889b2e3693..f327fcfc9f 100644
--- a/src/env/config_paths.rs
+++ b/src/env/config_paths.rs
@@ -105,30 +105,16 @@
             return invalid_exec_path(exec_path);
         };
 
-        // If we're in Cargo's target directory or CMake's build directory, use the source files.
-        if exec_path.starts_with(env!("FISH_BUILD_DIR")) {
-            let repo_root = Path::new(REPO_ROOT);
-            return Self {
-                kind: RepoRoot,
-                paths: ConfigPaths {
-                    sysconf: repo_root.join("etc"),
-                    bin: Some(exec_path_parent.to_owned()),
-                    data: repo_root.join("share"),
-                    // This requires manual sphinx invocation.
-                    doc: repo_root.join("user_doc/html"),
-                    // This requires manual build_tools/update_translations.fish invocation.
-                    locale: Some(repo_root.join("share/locale")),
-                },
-                exec_path,
-            };
-        }
+        let repo_root = Path::new(REPO_ROOT);
 
         // The next check is that we are in a relocatable directory tree
         if exec_path_parent.ends_with("bin") {
             let base_path = exec_path_parent.parent().unwrap();
             let data = base_path.join("share/fish");
             let sysconf = base_path.join("etc/fish");
-            if data.exists() && sysconf.exists() {
+            if base_path != repo_root // Install to repo root is not supported.
+                && data.exists() && sysconf.exists()
+            {
                 let doc = base_path.join("share/doc/fish");
                 return Self {
                     kind: InstalledTree,
@@ -149,6 +135,23 @@
             }
         }
 
+        // If we're in Cargo's target directory or CMake's build directory, use the source files.
+        if exec_path.starts_with(env!("FISH_BUILD_DIR")) {
+            return Self {
+                kind: RepoRoot,
+                paths: ConfigPaths {
+                    sysconf: repo_root.join("etc"),
+                    bin: Some(exec_path_parent.to_owned()),
+                    data: repo_root.join("share"),
+                    // This requires manual sphinx invocation.
+                    doc: repo_root.join("user_doc/html"),
+                    // This requires manual build_tools/update_translations.fish invocation.
+                    locale: Some(repo_root.join("share/locale")),
+                },
+                exec_path,
+            };
+        }
+
         Self {
             kind: StaticPathsDueToWeirdLayout,
             paths: ConfigPaths::static_paths(),
```

(Alternatively, we could change the
`exec_path.starts_with(env!("FISH_BUILD_DIR"))` check to only match
`$cargo_target_dir/{debug,release,...}/fish` and `$CMAKE_BINARY_DIR/fish`,
not sure how to get the debug/release part reliably)

.github/workflows/rust_checks.yml
danielrainer: > Obviously we'd prefer using something like gettext-rs that's not specific to fish

Something not specific to fish would be nice (although I'm not sure how that would work with fish variables), but it would probably be something other than gettext-rs. That crate just wraps normal gettext, so for e.g. embedding to work, GNU gettext would have to add support for reading MO data from memory.

Keeping localization support in some scenarios with embedded builds via your patch would be nice, but I wouldn't really mind it being broken for a bit if we're replacing the gettext implementation anyway.

src/env/config_paths.rs
danielrainer: > "source directory" but that might be confused with `src/`.

Yes, "source directory" was the first alternative I considered, but I decided against it for the same reason. Using "repo" also bothered me a bit for the reason you state, but I haven't found a better option so far.

> I'll remove the path from the enum, since the manifest dir is a global constant anyway.

Makes sense and eliminates the need to figure out where the path points to, since there no longer is any path.

We might want to use code from `build-helper` outside of build scripts as well. The `build-helper` crate would be added as a runtime dependency, but I don't really see a problem with that.

> I think we definitely want this to work; It should already work for CMake builds, because we do `FISH_BUILD_DIR=${CMAKE_BINARY_DIR}`.

Ah yes, for some reason I thought the check is about the repo root instead of the build directory.

> Unfortunately I missed that cargo only reads `CARGO_TARGET_DIR` but does not set it (or anything directly equivalent).
So yeah, this is broken for now, would be nice to fix

I just tested Cargo's behavior and using `--target-dir` does indeed not set `CARGO_TARGET_DIR`. This also breaks `embed-data` builds, since these put intermediate artifacts into the target dir. I don't know if there is a way to get the target dir actually used in a cargo-only build in a way that allows baking it into the executable.

Maybe it makes sense to put a file containing the fish version ID into install locations. Then, fish can use the files there if the executable's version matches (and be confident that the correct version of data files is available), and otherwise fish can try to use files from the repo. We would still have to fix build dir detection for other reasons, but it would result in executables having access to data files in situations where they currently don't, e.g. when installed via `cargo install`.

> (Alternatively, we could change the `exec_path.starts_with(env!("FISH_BUILD_DIR"))` check to only match `$cargo_target_dir/{debug,release,...}/fish` and `$CMAKE_BINARY_DIR/fish`, not sure how to get the debug/release part reliably)

I don't think that would be an improvement. Trying to match subdirectory names in `target/` (or whatever non-standard name is used) is a pain. There can be various profile names, and when explicitly compiling for a platform you get subdirectories with names indicating the platform. I also don't know if any of this directory naming is considered stable in Rust.

krobelus: > Using "repo" also bothered me a bit

Let's use "Workspace" or "WorkspaceRoot" (or "WorkspacePath").
As used by Cargo, VSCode and Jujutsu.
The Cargo workspace root is probably always going to be the same as $(git rev-parse --show-toplevel).

> We might want to use code from `build-helper`

sounds good, nice.

> Ah yes, for some reason I thought the check is about the repo root

we should probably call it FISH_TARGET_DIR to move closer to Rust 

> I just tested Cargo's behavior

I raised https://github.com/rust-lang/cargo/issues/15962, maybe there
is enough want for this sort of thing

> but it would result in executables having access to data files
> in situations where they currently don't, e.g. when installed via
> cargo install.

but that'd only be relevant if we don't have embedding.
We can probably declare "cargo install --no-default-features" as unsupported
(I guess we should make it fail early rather than using random other runtime files...
except on OpenBSD where std::env::current_exe() won't work, so the compile-time paths are all we have).

.github/workflows/rust_checks.yml
krobelus: I think I'll break embedded localization for `target/*/fish` too, to reduce surprise

src/env/config_paths.rs
danielrainer: > Let's use "Workspace" or "WorkspaceRoot" (or "WorkspacePath").

Good idea, that's easy to understand, hard to be confused about, and works in all scenarios.

.github/workflows/rust_checks.yml
danielrainer: Makes sense

PR #11789, opened 2025-09-10
Allow disabling `fish_indent` for parts of files

IsaacOscar:
As discussed on #11750, this adds the ability to stop fish_indent from formatting certain blocks of 
code by surrounding them in `# format: off` ... `# format: on` directives.
The primary motivation for this feature is to allow running fish_indent on the project's top-level directory,
whilst still allowing tests in tests/checks to test fish's handling of incorrectly formatted code.

# Details
(The following is excerpted from doc_src/cnds.fish_indent.rst):

Except when using `--only-indent` or `--only-unindent`, `fish_indent` will recognise the following special comments:

* `#!fish_indent: off`
    All text starting from the `#!` up until the next `#!fish_indent: on` (or end of file), will be output verbatim (i.e. it will not be reformated in any way).

    Note: if another `#!fish_indent: off` occurs before a subsequent `#!fish_indent: on` (or end of file), it will be treated like an ordinary comment.

* `#!fish_indent: on`
    This disables the effect of any previous `#!fish_indent: off`. All text starting from the `#!` up until the next `#!fish_indent: off` (or end of file), is formatted as normal.

    Note: if another `#!fish_indent: on` occurs before a subsequent `#!fish_indent: off` (or end of file), it will be treated like an ordinary comment.

Other comments starting with `#!fish_indent:` are not treated specially, but this may change in the future.

**(Let me know if you think supporting --only-indent or --only-unindent would be useful)**

# The Commits

This just lists the commits I've made and their associated log message for easy reference:

1. 904bf0ad79074da23c77acf64163063ae798db9b: Removed brackets that were causing a compilation error. (This error was caught by `build_tools/check.sh`)

2. a49f5037300da78be62062e10a32210901876e12: Make `fish_indent` keep comments before left curly brackets. (This just fixes a bug I found)

3. b2abde9c380899339f5810d787b6123675501774: Added ability to exclude parts of code from `fish_indent` reformatting.
    > This makes `fish_indent` (when not in `--only-indent` or `--only-unindent` mode) recognise the following special comments:
    >     ```fish
    >     #!fish_indent: on
    >     #!fish_indent: off
    >     ```
    > Where there can be any amount of whitespace around `on` and `off`.
    > Any text starting from a `#!fish_indent: off`, and continuing to the
    > start of the next `#!fish_indent: on` (or end of file), is output
    > verbatim by fish_indent, and not reformatted.
    > 
    > The implementation of this prioritises ease of maintainance vs.
    > performance: thus the text in a `#!fish_indent: off` is still formated
    > like normal, but this formated text is then replaced with the original
    > text when the `#!fish_indent: on` (or end of file) is encountered.
    > Thus nothing special needs to be done if the formatting code is changed,
    > provided that:
    > 
    >  1. a `#!fish_indent` directive is always proccessed before any text
    >     after it.
    > 
    > 2. the formatted-text before a `#!fish_indent: on` directive is not
    >     modified once said directive has been proccessed.
    > 
    > The downside of this approach is that it will waste time and memory
    > generating formatted code that is then simply discarded.
    > 
    > Only a simple test case is included, demonstrating that this feature
    > works within a single command line. A later commit however will
    > efectively make build_tools/check.fish itself a test case for this
    > feature.

4. f15d39788abc8171bcd5bf5a55b50fc143e1bbf9: Added `# format: off/on` directives to tests.
    > This excludes a few parts of the test suite from fish indent,
    > namely those parts whose non-standard formatting was clearly
    > part of the test case.

5. 982bdd9c183feee7f0c9b36d4bce123b58064aef: Make `build_tools/style.fish` format all `.fish` and `.py` files.
    > Previously only `.py` and `.fish` files in certain directories where
    > formatted. There are no `.py` files outside these directories,
    > but there are fish files in the `tests` directory, however the previous
    > commit made these work with `fish_indent` (by using `#!fish_indent: off`).

7. 59a6c96d922e874f877250330adcbb641303a82a: Run `build_tools/style.fish --all`, reformatting all the test cases.
    > This also modifies a `#CHECKERR` line in tests/checks/test.fish,
    > to match the reformatted code.

5. c0ac6b60a4ffd44483a054446ac98c6dfcb3c987: Added a simple test of some badly formatted code.
    > This new test case is to compensate for the removal of such code in the
    > previous commit.

### DIFF
--- a/tests/checks/basic.fish
+++ b/tests/checks/basic.fish
@@ -657,3 +657,19 @@ chmod +x $tmpdir/argv0.sh
 cd $tmpdir
 ./argv0.sh
 # CHECK: ./argv0.sh
+
+# A simple check of badly formatted code
+#!fish_indent: off
+  begin; echo -n "1"
+# A comment
+   ; echo -n          2 |
+cat (
+		
+	echo -n 3|psub
+)
+
+        set a 4 &&
+echo -n $a[      1 ]
+    echo -n 5         >         /dev/stdout; end
+#!fish_indent: on
+#CHECK: 1345


Xiretza:
`#!` is already semi-reserved for the shebang, I wouldn't expect it anywhere other than at the start of a file. Why not simply `# fish_indent: off`, similar to e.g. [`# shellcheck disable=FOO`](https://github.com/koalaman/shellcheck/wiki/Ignore) or [`// clang-format off`](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#disabling-formatting-on-a-piece-of-code)?

IsaacOscar:
> `#!` is already semi-reserved for the shebang, I wouldn't expect it anywhere other than at the start of a file. Why not simply `# fish_indent: off`, similar to e.g. [`# shellcheck disable=FOO`](https://github.com/koalaman/shellcheck/wiki/Ignore) or [`// clang-format off`](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#disabling-formatting-on-a-piece-of-code)?

I wanted it to look 'special' so as not to be misinterpreted as a normal comment.
I'm happy to change it once the rest of my pull request has been reviewed.

krobelus:
Technically we could also make it a nop command, such as
"status meta-attribute format off" and
"status meta-attribute format on"
which would make it very obvious how to look up docs though that's
probably not important here

IsaacOscar:
> Technically we could also make it a nop command, such as "status meta-attribute format off" and "status meta-attribute format on" which would make it very obvious how to look up docs though that's probably not important here

Hmm, it's not possible to override the meaning of `status` with a function/command right?


It could be more confusing though, e.g.
```
if false
   status meta-attribute format off
end
   echo should this be reformated?
```

IsaacOscar:
Also what about
```
set var off
status meta-attribute format $var
```

I personally think a comment is better.

CHANGELOG.rst
```diff
@@ -98,6 +98,7 @@ Other improvements
 - ``fish_indent`` and ``fish_key_reader`` are now available as builtins, and if fish is called with that name it will act like the given tool (as a multi-call binary).
   This allows truly distributing fish as a single file. (:issue:`10876`)
 - ``fish_indent --dump-parse-tree`` now emits simple metrics about the tree including its memory consumption.
+- ``fish_indent`` will no longer reformat code between a ``#!fish_indent: off`` comment and a ``#!fish_indent: on`` comment (or end-of-file).
```
krobelus, line 101: maybe phrase it as "will skip reformatting",
because "no longer" is usually used for bugs in this context..

build_tools/style.fish
```diff
@@ -43,8 +43,8 @@ if test $all = yes
             end
         end
     end
-    set fish_files $repo_root/{benchmarks,build_tools,etc,share}/**.fish
-    set python_files {doc_src,share,tests}/**.py
+    set fish_files $repo_root/**.fish
```
krobelus, line 46: let's not indiscriminately format all files in the workspace.
I know that this bookkeeping is suboptimal but it's a reasonable
middle ground; this list won't change often.
My main problem here is that it will make tests fail if a gitignored
directory is not formatted, or worse, reformat gitignored files.
I sometimes keep temporary, private test files in such directories.

There's some further discussion in https://github.com/fish-shell/fish-shell/pull/11608

Let's only add tests/ for now.

build_tools/style.fish
```diff
@@ -43,8 +43,8 @@
...
+    set python_files $repo_root/**.py
```
krobelus, line 47: > (Let me know if you think supporting --only-indent or --only-unindent would be useful)

no strong opinion yet; we can do it later.
Those commands are only used for copying from / pasting to the fish commandline,
where all indentation is virtual

doc_src/cmds/fish_indent.rst
```diff
@@ -54,3 +54,17 @@ The following options are available:
 
 **-h** or **--help**
     Displays help about using this command.
+
+Except when using ``--only-indent`` or ``--only-unindent``, :program:`fish_indent` will recognise the following special comments:
+
+**#!fish_indent: off**
```
krobelus, line 60: I don't mugh thought on the syntax other than yeah, `#!` looks
suspiciously like a shebang (for better or worse), and I'd make it
match clang-format, the space in the comment looks cleaner.  (I'd also
rename "fish_indent" to "fish-format" but that's an unrelated change).

doc_src/cmds/fish_indent.rst
```diff
@@ -54,3 +54,17 @@
...
+    All text starting from the ``#!`` up until the next ``#!fish_indent: on`` (or end of file), will be output verbatim (i.e. it will not be reformated in any way).
```
krobelus, line 61: I'd expect that we do want to format the starting and ending comment.
Consider

```shell
printf %s '
    // clang-format off
    // not formatted
        int main() {}
   // clang-format on
        int main2() {}
' | clang-format
```

where the clang-format comments are indeed formatted.

This is not a super strong opinion,
I'm also fine with not formatting the comment pair (at least for now).
But test cases indicate that we only format the `#!fish_indent: on`
line but not the other, which seems weird?

doc_src/cmds/fish_indent.rst
```diff
@@ -54,3 +54,17 @@
...
+
+    Note: if another ``#!fish_indent: off`` occurs before a subsequent ``#!fish_indent: on`` (or end of file), it will be treated like an ordinary comment.
```
krobelus, line 63: Yeah, there's no need to allow stacking these directives

src/autoload.rs
```diff
@@ -564,7 +564,7 @@ fn test_autoload() {
         autoload.resolve_command_impl(L!("file2"), paths),
         AutoloadResult::Loaded
     ));
-    assert!((autoload.get_autoloaded_commands() == vec![L!("file1"), L!("file2")]));
+    assert!(autoload.get_autoloaded_commands() == vec![L!("file1"), L!("file2")]);
```
krobelus, line 567: compilation error?
Wow how did that happen, which Rust version?
FWIW, we should prefer `assert_eq!` for such tests because that prints a diff on failure

src/builtins/fish_indent.rs
```diff
@@ -83,6 +83,11 @@ struct PrettyPrinterState<'source, 'ast> {
     // This is computed ahead of time for convenience.
     preferred_semi_locations: Vec<usize>,
 
+    // The indices of the start of the "#!fish_indent: off" directive that is in effect
+    // (The first index is into to self.source, the second is into self.output.)
+    // This will be none if we have not read such a directive yet, or it has since been canceled by
+    // a "#!fish_indent: on" directive.
+    fish_indent_off: Option<(usize, usize)>,
```
krobelus, line 90: using "fish_indent" here is weird because it should be implied by the context, maybe remove that
or call it "excluded_range" or similar.
But maybe it's appropriate as-is, not sure.

src/builtins/fish_indent.rs
```diff
@@ -471,6 +481,23 @@
         }
     }
 
+    // Set the formatting mode to on/off; directive_offset is the start index of
+    // the #!fish_indent directive.
+    fn set_formatting(&mut self, on: bool, directive_offset: usize) {
```
krobelus, line 486: The struct is called PrettyPrinter;
IIRC, if formatting is disabled, it doesn't do anything,
so maybe `set_enabled()` would be a better name?

src/builtins/fish_indent.rs
```diff
@@ -471,6 +481,23 @@
...
+        match self.fish_indent_off {
+            None if !on => {
+                self.fish_indent_off = Some((directive_offset, self.output.len()));
+            }
+            Some((source_index, output_index)) if on => {
+                // Replace any formatted text we've generated with the original non-formatted text
+                self.output
+                    .replace_range(output_index.., &self.source[source_index..directive_offset]);
```
krobelus, line 494: I see, so since we do a single pass through the AST, we don't know
in advance that we don't need to format this?
If it ever becomes a problem in terms of performance,
we can add a cheap first pass to compute the skip ranges.

src/builtins/fish_indent.rs
```diff
@@ -471,6 +481,23 @@
...
+                self.fish_indent_off = None
+            }
+            _ => {} // No need to do anything, formatting was already on/off
```
krobelus, line 497: I suppose we could print a warning, since imbalanced directives
might result in us mangling the user's script.
But that's probably unconventional, since we also don't 
warn on other errors. For example, we'll happily indent

	begin
	echo fooo
	begin 
	echo bar
	end

probably there are better examples...

src/builtins/fish_indent.rs
```diff
@@ -738,6 +777,7 @@
     fn visit_left_brace(&mut self, node: &dyn ast::Token) {
         let range = node.source_range();
         let flags = self.gap_text_flags_before_node(node.as_node());
+        self.emit_gap_text_before(range, flags);
```
krobelus, line 780: A test would be great here (maybe it's in a later commit). Tests are
even more important because I know that there are some cases where
formatting braces (i.e. compound statements) will lead to wrong
formatting, and even to formatting that changes the behavior.
Not sure if I still have my test cases, probably you hit the same issues as me.

tests/checks/basic.fish
```diff
@@ -625,8 +652,24 @@ $fish -c 'echo \ufdd2"fart"'
...
+cat (
+		
+	echo -n 3|psub
+)
+
+        set a 4 &&
+echo -n $a[      1 ]
+    echo -n 5         >         /dev/stdout; end
+#!fish_indent: on
+#CHECK: 1345
```
krobelus, line 675: Nice test - I'd expect it to live in tests/checks/indent.fish,
we should be able to simply copy it over

tests/checks/braces.fish
```diff
@@ -164,13 +167,13 @@ end
 }
 # CHECK: while
 
-{{echo inner}
-    echo outer}
+{ { echo inner }
+    echo outer
+}
```
krobelus, line 172: I think these tests in `braces.fish` were added to test parsing and demonstrate what's possible.
Instead of changing them, can you create copies of the affected paragraphs,
and format only one each.

src/autoload.rs
IsaacOscar: Sorry, I just checked, and what actually happens is that gives a warning, but `build_tools/check.sh` passes `--deny=warnings`, so it looked like a compilation error to me.
And yes, `assert_eq!` is better, I'll change it to use that.

doc_src/cmds/fish_indent.rst
IsaacOscar: I just did what was easiest to program.
But I'll change it to match the `clang-format` behaviour.

doc_src/cmds/fish_indent.rst
IsaacOscar: So:
```
# fish_indent off
```
Without the colon? (To make it as close to clang format as reasonable?).

src/builtins/fish_indent.rs
IsaacOscar: I called it `fish_indent_off` because it's the indices of the `#! fish_indent: off` directive. But sure, I can call it `exclude_start`? (It's not a range, it only contains start indices).

src/builtins/fish_indent.rs
IsaacOscar: > I see, so since we do a single pass through the AST, we don't know in advance that we don't need to format this? If it ever becomes a problem in terms of performance, we can add a cheap first pass to compute the skip ranges.

I was originally going to wrap all the `self.output.push(...)` bits in an `if self.fish_indent_off.is_none()` or something like that, but I was worried that if someone adds a new AST node in the future, they may forget to wrap the output generating code in an `if`. Whereas this way, they don't need to even know about the `fish_indent: off` feature.

src/builtins/fish_indent.rs
IsaacOscar: If it's easy, I'll see about making it print a warning on parse errors as well? I just think the error handling here should be consistent.
I also don't print a warning if you do a `#!fish_indent: not on/off`.

tests/checks/basic.fish
IsaacOscar: This is *not* a test of `fish_indent` though, it's a test of `fish` itself: that it can execute badly formated code. So I didn't think it belonged in that file.

krobelus:
> This is *not* a test of `fish_indent` though, it's a test of `fish`
> itself: that it can execute badly formated code. So I didn't think
> it belonged in that file.

ah of course, that makes sense. The fish_indent test can be much simpler

src/builtins/fish_indent.rs
krobelus: yeah I think we could print parse errors to stderr.
But my example above is a bad one, because it is merely incomplete (as allowed by `LEAVE_UNTERMINATED`), I'm not sure if we should even warn about that.

`rustfmt` outright refuses to format input with syntax errors, it doesn't even print anything to stdout.
I don't think we should do that, because it makes naïve editor integration worse.
So maybe first print to stdout, then emit any errors/warnings to stderr.

Maybe add errors/warnings for everything in a separate PR, because it might be tricky.

doc_src/cmds/fish_indent.rst
krobelus: I find the lack of colon weird. Looking at comparable tools, I do see some are using colons, see
https://github.com/astral-sh/ruff/discussions/6338 for example.
So I'd go with `# fish_indent: off`

IsaacOscar: Hmm, actually `# fmt: off` (i.e. what black does) makes more sense, as it's not disabling indentation, (and doesn't do anything in ident-only mode)

IsaacOscar: (`# format: off` would personally be nicer though, and fits more with fish as it doesn't necessarily abbreviate stuff)

IsaacOscar: What should the policy on spaces be?
`black` has the spaces in `# fmt: off` optional,
but there can't be extra spaces.
Clang format however requires exactly thr same spaces in `// clang-format off`
It also allows you to write a `:` (with no preceding space) followed by arbitrary text to describe why you're turning formatting off.

krobelus: since `format` is a more generic term than `fish_indent`, it might have false positives,
even more so because unlike Python, fish doesn't have nice multi-line commenting syntax yet.
I suppose it's probably no big deal, it only matters in very rare edge cases.

> black has the spaces in # fmt: off optional,
but there can't be extra spaces.

matching `# ?format: ?off$` sounds reasonable, either that or allow arbitrary spaces, even trailing ones

src/builtins/fish_indent.rs
krobelus: > I was worried that if someone adds a new AST node in the future, they may forget to wrap the output generating code in an if

we could extract a function for writing to output and use that everywhere, that should work.
But if I understand correctly, we'd only be saving a few redundant writes; we still need to walk the whole AST etc.
So there's maybe not a lot of upside to do this extra work; it's fine as is

zanchey:
This doesn't seem worth the added complexity to me.

krobelus:
> This doesn't seem worth the added complexity to me.

I don't know if complexity means UI or implementation here.

One the one hand, this is one way to enable us to format **.fish files in a simple and obvious way (configure your editor to format all files in fish-shell/**.fish by calling `fish_indent`).
Making this easy simplifies onboarding and thus maintenance, so we should definitely do that (assuming we want formatted code).
 
To work around the status quo, I've been excluding the entirety of tests/ using

	hook global BufCreate "\Q%sh{project-root}/tests/\E.*\.fish" %{
		set-option buffer disabled_hooks "%opt{disabled_hooks}|format"
	}

Another way to simplify formatting would be to move all  unformatted code into quotes (`eval '[unformatted code]'` or `fish -c '[unformatted code]'`).
That would communicate intent less clearly than `# format: off` but we could work around this by adding comments.
This alternative seems fine to me too but I haven't tried it; it might make some tests more convoluted.

On the other hand, `# format: off` would allow fish scripts to selectively use different formatting.
I don't have much of an opinion about enabling this; can we maybe find a use case from someone who writes fish scripts.
My intuition is affirmative because I like this feature a lot in other languages, since manual-formatting can do a much better job at conveying intent.
Basically every formatter except `gofmt` supports it (which begs the question, does fish want to be different like that).

So as the PR description says, the immediate motivation
here is to make `fish_indent -w **.fish` work, which is great.
With the existing approach we might also help others similar situation, IDK.

doc_src/cmds/fish_indent.rst
Nahor: I also think that `#!` is a bad idea. For regular script files (i.e. to be executed like regular binary apps), it's not a problem, but for other files (e.g. function or completions files), and one one might add a `#!fish_indent off` on the very first line.
The OS can interpret that as a real shebang and believe it the file to be a script that needs to be executed via `fish_indent`.

Some OSes might impose extra restrictions that will block that behavior (e.g. requiring a full path to the `fish_indent` interpreter) but not all do. On Cygwin at least, it's enough to mark the file as executable and to call `fish_indent` when executed:
```
$echo '#!fish_indent off
foo
   bar
' > test

$ ls -l ./test
-rwxr-xr-x 1 Nahor None 30 Dec 18 10:11 ./test*

$ ./test
Opening "off" failed: No such file or directory (os error 2)

$ echo '#!fish_format off
foo
   bar
' > test

$ ./test
bash: ./test: fish_format: bad interpreter: No such file or directory

$
```

zanchey:
> So as the PR description says, the immediate motivation here is to make `fish_indent -w **.fish` work, which is great.

I don't see that as a useful outcome but maybe it's just me; I don't know why someone might want to run that command.

krobelus:
the goal is that we can enable format checking in CI,
but we can also achieve that (at least for most files) in a different way that doesn't need a change to fish_indent.
I have a build_tools/fish_indent_wrapper that ignores known-unformatted files, maybe that's the compromise.

Nahor:
>  we can also achieve that (at least for most files) in a different way that doesn't need a change to fish_indent.
> I have a build_tools/fish_indent_wrapper that ignores known-unformatted files, maybe that's the compromise.

But that a all-or-nothing. The point of the issue is to still mostly format a file, except for specific parts.

And it would also be useful for IDE with format-on-save, since it's not possible, or cumbersome at least, to specify what file shouldn't be formatted. With the feature, one could just put a comment at the top of the file to the same effect.

krobelus:
yeah I think this PR is fine. We might not want to advertise this to outside users since we can't think of a use case for those,
but we do have the in-house use case so we should do it to make our life easier

Nahor:
> we can't think of a use case for those

What do you mean? There are people writing script for themselves out there. Why wouldn't they need it sometimes?

krobelus:
> What do you mean? There are people writing script for themselves out there. Why wouldn't they need it sometimes?

yes, they might want this feature but we don't have a concrete, real-world example yet.
That would make the decision a lot easier.

Something I did the other day in the fish repo was this

	$motion || { set ok false; break }

but in that case reformatting would be better than adding ignore-annotations

PR #11843, opened 2025-09-30
Put general comments on top of empty msgid

danielrainer:
These should not be comments on an actual message, since they apply throughout the entire file, so the sensible location is as comments on the empty msgid.

### DIFF
--- a/po/zh_CN.po
+++ b/po/zh_CN.po
@@ -1,20 +1,7 @@
 # fish-note-sections: Translations are divided into sections, each starting with a fish-section-* comment.
 # fish-note-sections: The first few sections are more important.
 # fish-note-sections: Ignore the tier3 sections unless you have a lot of time.
-msgid ""
-msgstr ""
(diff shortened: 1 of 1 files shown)


po/zh_CN.po
```diff
@@ -41,6 +28,20 @@ msgstr ""
...
+"Last-Translator: Yuyi Wang <Strawberry_Str@hotmail.com>\n"
+"Language-Team: \n"
+"Language: zh_CN\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"X-Generator: Poedit 3.7\n"
+
 msgid ""
 "\n"
```
krobelus, line 46: LGTM.
BTW another quirk is that the `# fish-section-*` comments show up in `poedit` as comments on individual messages.
I don't know any better, it was a quick hack

danielrainer: That's what #11845 is about

PR #11845, opened 2025-09-30
Refactor PO section marking

danielrainer:
Use msgids to mark sections. In the PO format, comments are associated
with specific messages, which does not match the semantics for section
markers.
Furthermore, comments are not preserved by `msgmerge`, which required
quite convoluted handling to copy them over from the template.
By using msgids to mark sections, this problem is avoided.

This convoluted handling was also used for header comments. Header
comments are now handled in a simpler way. There is a fixed prefix,
identifying these comments, as well as a list variable containing the
lines which should be put into the header. When a PO file is generated,
all existing lines starting with the prefix are deleted, the and the
current version of the lines is prepended to the file.

This is on top of #11843.

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -1,4 +1,4 @@
-# fish-note-sections: Translations are divided into sections, each starting with a fish-section-* comment.
+# fish-note-sections: Translations are divided into sections, each starting with a fish-section-* pseudo-message.
 # fish-note-sections: The first few sections are more important.
 # fish-note-sections: Ignore the tier3 sections unless you have a lot of time.
 # translation of de.po to deutsch
@@ -21,6 +21,9 @@ msgstr ""
 "Content-Transfer-Encoding: 8bit\n"
 "X-Generator: Poedit 3.5\n"
 
+msgid "fish-section-tier1-from-rust"
+msgstr ""
+
 msgid ""
 "\n"
 "   PID  Command\n"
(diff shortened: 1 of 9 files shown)


po/de.po
```diff
@@ -21,6 +21,9 @@ msgstr ""
 "Content-Transfer-Encoding: 8bit\n"
 "X-Generator: Poedit 3.5\n"
 
+msgid "fish-section-tier1-from-rust"
+msgstr ""
```
krobelus, line 25: Thanks for cleaning this up.
This makes a lot more sense, it's nice that it shows up properly in poedit.

I wonder if we should make it more obvious that these should not be translated.
We can't use `msgid ""` unfortunately because of the uniqueness constraint.

I wonder if something like `# Pseudo-message; do not translate` helps.
If we don't think of anything better, I'll add that.

(tier2 is totally empty right now, we want to classify things eventually)

danielrainer: > I wonder if something like `# Pseudo-message; do not translate helps`.

The issue with comments is that it is a pain to get them from the template to the per-language files. We could just put them in those files manually. Not sure if the manual effort is worth it, especially when it comes to keeping it up-to-date and with new languages. I don't think many translators will try to translate these section markers, and even if they do and it gets merged, it doesn't really cause any problems, so I think my preferred approach is not adding comments at each marker.

PR #11833, opened 2025-09-28
zh_CN: fix part of translation

Berrysoft:
The latest Chinese translation is TOO BAD. I'll try to fix them, but it might take some time.

PR #11854, opened 2025-10-01
zh_CN: no sentence-initial caps; capitalization reserved for proper nuns

qianlongzt:
https://github.com/fish-shell/fish-shell/pull/11833#discussion_r2393596717


Berrysoft:
I still don't agree with this idea... Sigh.

po/zh_CN.po
```diff
@@ -3655,10 +3655,10 @@ msgid "Verbose mode"
 msgstr "详细模式"
 
 msgid "Vi mode commands"
-msgstr "Vi 模式命令"
+msgstr "vi 模式命令"
 
 msgid "Vi-style bindings that inherit emacs-style bindings in all modes"
-msgstr "所有模式中的继承 emacs 样式绑定的 Vi 样式绑定"
+msgstr "所有模式中的继承 emacs 样式绑定的 vi 样式绑定"
```
Berrysoft, line 3661: Oh, yes, this one is OK.

qianlongzt:
Voting: 👍 agree with sentence-initial caps, 👎 disagree.

po/zh_CN.po
```diff
@@ -3529,10 +3529,10 @@ msgid "TRAP handler: debug prompt"
 msgstr "TRAP 处理函数：调试提示符"
 
 msgid "Tab Completions"
-msgstr "Tab 补全"
+msgstr "tab 补全"
```
Integral-Tech, line 3532: IMHO this change is not really necessary. The "Tab" key on the keyboard is usually be capitalized.

Integral-Tech: IMHO this change is not really necessary. The "Tab" key on the keyboard is usually capitalized.

krobelus:
Ok it's controversial change.
I found it hard to find comparable translations in other projects.. looks like partially-translated sentences are a niche thing;
maybe we should be doing something differently, like writing the Chinese term followed by the English technical term in parentheses like `(stdin)`. But I'm not sure if that makes sense.

Until we have some more evidence, ideally from real world example, or a clear
majority, I'd leave things since @Berrysoft effectively owns this now.

Can you extract the non-controversial bits to a separate commit that I can cherry-pick.

>  msgid "Tab Completions"
> -msgstr "Tab 补全"
> +msgstr "tab 补全"
> 
> IMHO this change is not really necessary. The "Tab" key on the keyboard is usually capitalized.

In my dictionary, "Enter" is capitalized but "tab" or "tab key" are not, that's why I made this.
On the key caps, everything is capitalized so I'm not sure that helps.
Wikipedia is [inconsistent about it](https://en.wikipedia.org/wiki/Tab_key), so I don't mind either way..

qianlongzt:
@krobelus I separated the commit.

qianlongzt:
I don't think the test failure was caused by this PR.

qianlongzt:
Close this PR because no one else agreed.

PR #11864, opened 2025-10-03
Cleaned up, fixed bug, and added functionality to bind completions 

Solipsistmonkey:
## Description

Since bind can take a series of functions in defining a keybinding, it didn't make sense to only provide completions for bind functions once. When I went to fix that, I saw the functions __fish_bind_test1 and __fish_bind_test2. The first isn't being used by anything I could see, so I removed it. 

The second one used iteration for counting args and discounting flags. Then used a switch/case block to check if there were 2 args which would be 'bind' and whatever key-combo. And it only checked if there were a total of 2. This means it fails after the first function is entered. If the commandline was  `bind -m MODE `, it would give the function completions along with the key completions. 

I went ahead and changed  the name of __fish_bind_test2 to be more descriptive. And just completely changed the function so it accounts for multiple functions and ignores counting bind's args/params.

Fixes issue #

## TODOs:

### DIFF
--- a/share/completions/bind.fish
+++ b/share/completions/bind.fish
@@ -1,46 +1,8 @@
-function __fish_bind_test1
-    set -l args
-    set -l use_keys no
-    for i in (commandline -pxc)
-        switch $i
-            case -k --k --ke --key
-                set use_keys yes
+set -l __fish_bind_optspecs M/mode= m/sets-mode= preset user s/silent a/all function-names list-modes e/erase
 
-            case "-*"
-
-            case "*"
-                set -a args $i
-        end
-    end
-
-    switch $use_keys
-        case yes
-            switch (count $args)
-                case 1
-                    return 0
-            end
-    end
-    return 1
-end
-
-function __fish_bind_test2
-    set -l args
-    for i in (commandline -pxc)
-        switch $i
-            case "-*"
-
-            case "*"
-                set -a args $i
-        end
-    end
-
-    switch (count $args)
-        case 2
-            return 0
-    end
-
-    return 1
-
+function __fish_bind_has_keys --inherit-variable __fish_bind_optspecs
+    argparse $_bind_opts -- $argv 2>/dev/null
+    test ( count $argv ) -ge 2 && return 0
 end
 
 complete -c bind -f
@@ -56,11 +18,10 @@ complete -c bind -s s -l silent -d 'Operate silently'
 complete -c bind -l preset -d 'Operate on preset bindings'
 complete -c bind -l user -d 'Operate on user bindings'
 
-complete -c bind -n __fish_bind_test2 -a '(bind --function-names)' -d 'Function name' -x
+complete -c bind -n '__fish_bind_has_keys (commandline -pcx)' -a '(bind --function-names)' -d 'Function name' -x
 
-function __fish_bind_complete
-    argparse M/mode= m/sets-mode= preset user s/silent \
-        a/all function-names list-modes e/erase -- (commandline -xpc)[2..] 2>/dev/null
+function __fish_bind_complete --inherit-variable __fish_bind_optspecs
+    argparse $__fish_bind_optspecs -- (commandline -xpc)[2..] 2>/dev/null
     or return 1
     set -l token (commandline -ct)
     if test (count $argv) = 0 && set -l prefix (string match -r -- '(.*,)?(ctrl-|alt-|shift-|super-)*' $token)


share/completions/bind.fish
```diff
@@ -1,46 +1,10 @@
...
-        end
-    end
-
-    switch (count $args)
-        case 2
-            return 0
-    end
-
-    return 1
+set _bind_opts M/mode= m/sets-mode= preset user s/silent a/all function-names list-modes e/erase
```
krobelus: nice fixes,

couple things
1. always declare variables with explicit scope (the user could have a universal variable of this name)
2. use "__fish_bind" as prefix for private globals owned by "bind.fish", so something like "__fish_bind_optspecs"
3. in this case, we could declare it as file-local variable (which removes the need for namespacing)
   and use --inherit-variable=optspecs on both functions that use this. I think that's nicer.

share/completions/bind.fish
```diff
@@ -1,46 +1,10 @@
...
 
+function __fish_bind_has_keys \
+    --description 'Returns 0 if a key-combo exists in the bind command, 1 if there is none'
```
krobelus: this is subtle but we should not be adding descriptions
to private functions that are colocated in a
file that is otherwise translated.
Basically it's not worth it adding this function description to po/*.po
(and implicitly asking translators to translate it)
because hardly any user will ever see it.

If we want to keep the description, we can currently move it to a separate file
and explicitly skip localization, as done in e.g.
share/functions/__fish_any_arg_in.fish

As is, the description is confusing,
I'd rather describe it as "Returns 0 if the KEYS argument is present on the command line". 
But the we can also go without a description, since the nice new name already does the job.

share/completions/bind.fish
```diff
@@ -1,46 +1,10 @@
...
+    argparse $_bind_opts -- $argv
```
krobelus: I believe argparse can print to stderr on invalid options,
which would cause glitches when invoking completions.
So add 2>/dev/null here.
(stdout is fine because it's already  ignored in "complete -n ..." conditions)

Also, if argparse fails, I don't think we should suggest bind function names.
Better suggest nothing so the user notices that something's wrong and they fix their command line before continuing

share/completions/bind.fish
```diff
@@ -1,46 +1,10 @@
...
+    set -l arg_count ( count $argv )
```
krobelus: bad formatting here (if in doubt, use "fish_indent")

share/completions/bind.fish
```diff
@@ -1,46 +1,10 @@
...
+    test $arg_count -ge 2 && return 0 || return 1
```
krobelus: lose the "&& return 0 || return 1" part, that's implied

share/completions/bind.fish
```diff
@@ -56,11 +20,10 @@ complete -c bind -s s -l silent -d 'Operate silently'
 complete -c bind -l preset -d 'Operate on preset bindings'
 complete -c bind -l user -d 'Operate on user bindings'
 
-complete -c bind -n __fish_bind_test2 -a '(bind --function-names)' -d 'Function name' -x
+complete -c bind -n '__fish_bind_has_keys (commandline -pcx)' -a '(bind --function-names)' -d 'Function name' -x
```
krobelus, line 21: > Last commit didn't account for cases like `-M MODE`

Okay. Can you squash the last two commits into one then, to simplify review

share/completions/bind.fish
Solipsistmonkey: That is a completely fair point. I'm in the habit of using docstrings everywhere in python and was using the --description flag as a placeholder for that, should probably have just used a "#" block. Good call on the wording though, I'm not the best at articulating that type of thing

krobelus:
> That is a completely fair point. I'm in the habit of using
> docstrings everywhere in python and was using the --description
> flag as a placeholder for that, should probably have just used a
> "#" block. Good call on the wording though, I'm not the best at
> articulating that type of thing

normally it would be fine to add descriptions everywhere but we don't
have the infrastructure to mark description as not-translatable yet.
Though I suppose we could simply exclude all private functions'
descriptions from translations, that would make a ton of sense. Users
are not supposed to see those, unless they are debugging, in which
case they need to know English.
I'll do that if I get around to.

share/completions/bind.fish
Solipsistmonkey: I feel like I'm at 80% when it comes to understanding the scopes. But I've tried using `set -l some_var` at the top of a file, but functions don't seem to get the message. I had never used the --inherit-variable flag in `function` before. Just tried it out, and wow, I feel like it's been a missing piece of my shell experience and I'm kicking myself for not realizing it earlier

share/completions/bind.fish
Solipsistmonkey: I knew there was at least one  condition I was missing, and argparse failing was definitely one of them. Thanks!

share/completions/bind.fish
Solipsistmonkey: My usual style is to be a bit more explicit, it tends to just be a visual reassurance for myself. But that's a valid point

share/completions/bind.fish
Solipsistmonkey: To be honest, I'm still fairly novice when it comes to git. Haven't contributed to much and my usual workflow is either make a bunch of commits for small things or forget to commit for days and do a `git commit . -m "I have no idea what changed, but I better do it now before I get any deeper into this mess"`

How would I go about doing that? A rebase?

Solipsistmonkey: Nevermind, think I worked through it on my own. Hope I didn't miss anything, but I gotta run out the door. I'll get back to this later this evening.

Appreciate the feedback!

krobelus:
> To be honest, I'm still fairly novice when it comes to git.

A lot of very experienced people are too :(

> Haven't contributed to much and my usual workflow is either make a
> bunch of commits for small things or forget to commit for days and
> do a `git commit . -m "I have no idea what changed, but I better
> do it now before I get any deeper into this mess"`
>
> How would I go about doing that? A rebase?

Right, that's fine for history that's not read a lot.
But we want to be more intentional about the history, making sure every commit is atomic unit of change etc.
So yeah, "git rebase -i" and changing the last commit from "pick" to "squash" will combine that commit with its parent.
See https://git-rebase.io for a tutorial.
(and https://www.bitsnbites.eu/git-history-work-log-vs-recipe/ for the ideology)

Solipsistmonkey:
> A lot of very experienced people are too :( 

Ok, that got a chuckle out of me. 

But hey, I am always down for learning more. Self taught with coding, learned Python instead of making sourdough when I got Covid. Switched to Fish from ZSH about 6-8 months ago and fell in love with the syntax. So your comments here have been helpful as I want to contribute more to open source but it's intimidating. I haven't worked on any big projects or working in a professional coding environment. (I'm an electrician by trade). Definitely feel like I have no idea what's going on 90% of the time. 

Hell, that bit about --inherit-variable is going to save me many headaches. (Once I'm through the headache of fixing up my own function directory).

share/completions/bind.fish
```diff
@@ -1,46 +1,8 @@
...
-    switch (count $args)
-        case 2
-            return 0
-    end
-
-    return 1
+set -l __fish_bind_optspecs M/mode= m/sets-mode= preset user s/silent a/all function-names list-modes e/erase
 
+function __fish_bind_has_keys --inherit-variable __fish_bind_optspecs
+    argparse $_bind_opts -- $argv 2>/dev/null
```
krobelus, line 4: bad variable name here, and we also want to add something `or return` after argparse, like we have below; return a nonzero status

share/completions/bind.fish
```diff
@@ -56,11 +18,10 @@ complete -c bind -s s -l silent -d 'Operate silently'
 complete -c bind -l preset -d 'Operate on preset bindings'
 complete -c bind -l user -d 'Operate on user bindings'
 
-complete -c bind -n __fish_bind_test2 -a '(bind --function-names)' -d 'Function name' -x
+complete -c bind -n '__fish_bind_has_keys (commandline -pcx)' -a '(bind --function-names)' -d 'Function name' -x
 
-function __fish_bind_complete
-    argparse M/mode= m/sets-mode= preset user s/silent \
-        a/all function-names list-modes e/erase -- (commandline -xpc)[2..] 2>/dev/null
```
krobelus, line 63: actually, bind ignores any positional arguments if give either of
--function-names, --key-names, or --list-modes.

We should fix the builtin bind source code to print an error on
things like

	bind --function-names somearg

etc.

But that can be done in a later PR.
There are some smaller issues in this one left that I'll clean up.
For example I didn't mean you should squash the "remove unused
function" commit, because that one was already atomic.
Commits should be large enough to be correct but not any larger.

PR #11872, opened 2025-10-03
port fish_is_nth_token to rs

asukaminato0721:
## Description

part of #6695

## TODOs:


faho:
Turning arbitrary functions into rust code isn't something that was ever really planned, and has a number of downsides, including that they are worse to introspect and modify.

[My comment](https://github.com/fish-shell/fish-shell/issues/6695#issuecomment-984917460) in #6695 was about things like "up-or-search" that should just be a reader function (in reader.rs), and I would also be okay with the `cd` function being incorporated into the builtin, or e.g. "prevd" being added.

But `__fish_is_nth_token` is firmly in the "auxilliary" part of our functions and I don't think it deserves to be a proper builtin.

Adding options to the `commandline` builtin could work.

So I'll be closing this.

PR #11873, opened 2025-10-03
port up-or-search

asukaminato0721:
## Description

part of #6695

## TODOs:


localization/po/de.po
```diff
@@ -3279,9 +3279,6 @@ msgstr ""
 msgid "Save the current definition of all specified functions to file"
 msgstr ""
 
-msgid "Search back or move cursor up 1 line"
-msgstr ""
```
krobelus, line 3238: I think squashing into one commit is better here

src/reader/reader.rs
```diff
@@ -3499,6 +3499,28 @@ impl<'a> Reader<'a> {
                     self.update_command_line_from_history_search();
                 }
             }
+            rl::UpOrSearch => {
+                if self.history_search.active() {
+                    self.handle_readline_command(rl::HistorySearchBackward);
+                    return;
+                }
+
+                if self.is_navigating_pager_contents() || !self.pager.is_empty() {
```
krobelus, line 3890: huh this fixes a bug, we currently attempt history search if the pager has no matches

src/reader/reader.rs
```diff
@@ -3499,6 +3499,28 @@ impl<'a> Reader<'a> {
...
+                    self.handle_readline_command(rl::UpLine);
```
krobelus, line 3891: so this is a breaking change for anyone who did `bind x if ...; up-or-search; end`
Maybe we should keep a compatibility wrapper that does "commandline -f up-or-search".
Not sure yet.

Whatever we decide, the same thing should be applied to

	down-or-search.fish

and maybe also

	nextd-or-forward-token.fish
	prevd-or-backward-token.fish

By convention the dash-separated functions are special input functions and not regular functions.
But I'm not sure if we want to maintain that convention forever, I prefer dashes to underscores.
Either way, I wonder if something like `nextd-or-forward-token` should be given a fish_ prefix?

Then there are some others like edit_command_buffer where that seems appropriate.

src/reader.rs
```diff
@@ -3499,6 +3499,28 @@ impl<'a> Reader<'a> {
...
                 }
             }
+            rl::UpOrSearch => {
+                if self.history_search.active() {
+                    self.handle_readline_command(rl::HistorySearchBackward);
+                    return;
+                }
+
+                if self.is_navigating_pager_contents() || !self.pager.is_empty() {
+                    self.handle_readline_command(rl::UpLine);
```
asukaminato0721: Do we need to port `edit_command_buffer`?

krobelus:
> Do we need to port `edit_command_buffer`?

No. In theory we could rename it to "fish_edit_command_buffer"
and document it; but I'm not sure if that's worth the potential breakage.
So this would first need someone doing planning and thinking work.

src/reader/reader.rs
krobelus: To minimize the potential for chaos, I don't plan to merge these changes until we have a solid plan worked out and (explicitly?) documented.
Same for 
https://github.com/fish-shell/fish-shell/pull/11895
https://github.com/fish-shell/fish-shell/pull/11885

PR #11878, opened 2025-10-04
Printf remove length modifier support

danielrainer:
## Description

See commit messages.

## TODOs:
- [x] Changes to fish usage are reflected in user documentation/manpages. (Only really affects developer docs, length modifiers have not been documented in `printf` docs)

### DIFF
--- a/src/builtins/printf.rs
+++ b/src/builtins/printf.rs
@@ -208,10 +208,10 @@ impl<'a, 'b> builtin_printf_state_t<'a, 'b> {
         if errcode != None && errcode != Some(Error::InvalidChar) && errcode != Some(Error::Empty) {
             match errcode.unwrap() {
                 Error::Overflow => {
-                    self.fatal_error(sprintf!("%ls: %ls", s, wgettext!("Number out of range")));
+                    self.fatal_error(sprintf!("%s: %s", s, wgettext!("Number out of range")));
                 }
                 Error::Empty => {
-                    self.fatal_error(sprintf!("%ls: %ls", s, wgettext!("Number was empty")));
+                    self.fatal_error(sprintf!("%s: %s", s, wgettext!("Number was empty")));
                 }
                 Error::InvalidChar => {
                     panic!("Unreachable");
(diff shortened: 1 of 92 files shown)


src/tests/history.rs
```diff
@@ -228,7 +228,7 @@ fn time_barrier() {
 fn generate_history_lines(item_count: usize, idx: usize) -> Vec<WString> {
     let mut result = Vec::with_capacity(item_count);
     for i in 0..item_count {
-        result.push(sprintf!("%lu %lu", idx, i));
+        result.push(sprintf!("%u %u", idx, i));
```
krobelus, line 231: Not to beat the dead horse but we [should move to Rust format string syntax](https://github.com/fish-shell/fish-shell/pull/11857#discussion_r2404109075),
ideally with all the compile-time checks that we don't have with `sprintf!`.
I guess this PR is a step in that direction, it will make the conversion step easier.

src/builtins/printf.rs
```diff
@@ -523,14 +520,10 @@ impl<'a, 'b> builtin_printf_state_t<'a, 'b> {
                         }
                     }
 
-                    while matches!(f.char_at(0), 'l' | 'L' | 'h' | 'j' | 't' | 'z') {
-                        f = &f[1..];
-                    }
```
krobelus, line 528: Hmm I wonder whether we should really make this "gratuitous" breaking change.
`printf %ls\\n 123` works in bash/zsh but not in dash

I guess it's not documented, and there's no strong reason why anyone would use `%ls` in script.

but
1. people have been complaining that their `apt upgrade` would break
   any old fish that are still running, due to autoloading.
   This we can fix by unconditionally embedding `share/`
   (which will also avoid packagers being confused
   about having both /usr/share/fish/completions and
   /usr/share/fish/vendor_completions.d/)
   But until then, this sort of change is another potential instance of that.
2. People might have blindly copied things like

	printf (_ "%ls: Expected at least %d args, got only %d\n") open 1 0 >&2

from us.

Not sure. I guess it depends on what's our goal.
If we want to get rid of it regardless, the sooner the better I guess
(though the embedding mentioend above can make it less painful),
but it's not clear whether that's the goal, given that bash supports it (at least on my system, probably it uses printf).

danielrainer: I guess the second commit in this PR is optional. My thinking was that it is better to remove the functionality immediately rather than waiting to avoid new instances of length modifiers creeping in. It would also make it a bit easier to refactor the printf code as suggested in https://github.com/fish-shell/fish-shell/pull/11874#discussion_r2404061477.

Breaking running fish instances on upgrades is definitely not nice, and embedded script files would certainly help avoid such breakage. Given how slow some distros (and users) are to update, we would probably have to wait for months or years between removing the usages of length modifiers and then removing support for them, if we want to avoid breakage for most users. On that timescale, it seems realistic that we'll default to embedded builds first.

Whichever choice we make (also regarding localization), I think merging the first commit doesn't hurt, so we could do that now, and then figure out how we want to proceed with length modifier support.

krobelus: > It would also make it a bit easier to refactor the printf code as suggested

right, a bit but hopefully not substantially

PR #11879, opened 2025-10-04
French translation

danielrainer:
## Description

This is #11842 rebased on the commit removing length specifier usage. Since this rebase is non-trivial, I thought it would be nice to save translators from having to bother with it. @milouse feel free to use these commits for your PR if you want to continue working on the translations without merging the current state.

The commit I rebased on is not on master yet, but I think now that this rebase is done, there is nothing stopping us from merging it, right @krobelus?

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -343,7 +343,7 @@ command-line and graphical user interface programs. For simple use, you can use 
 
 Open up the PO file, for example ``po/sv.po``, and you'll see something like::
 
-    msgid "%ls: No suitable job\n"
+    msgid "%s: No suitable job\n"
     msgstr ""
 
 The ``msgid`` here is the "name" of the string to translate, typically the English string to translate.
@@ -351,10 +351,10 @@ The second line (``msgstr``) is where your translation goes.
 
 For example::
 
-    msgid "%ls: No suitable job\n"
-    msgstr "%ls: Inget passande jobb\n"
+    msgid "%s: No suitable job\n"
+    msgstr "%s: Inget passande jobb\n"
 
-Any ``%s`` / ``%ls`` or ``%d`` are placeholders that fish will use for formatting at runtime. It is important that they match - the translated string should have the same placeholders in the same order.
+Any ``%s`` or ``%d`` are placeholders that fish will use for formatting at runtime. It is important that they match - the translated string should have the same placeholders in the same order.
 
 Also any escaped characters, like that ``\n`` newline at the end, should be kept so the translation has the same behavior.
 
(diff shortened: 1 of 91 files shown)


danielrainer:
Btw, to verify the rebase for yourself, I recommend running `rg '^\+' | sed 's/%[hljtzL]\+/%/g' | sort` on the `git diff` output for the `Improve some french translations` commit and comparing the output for the old and new versions of the commit. (For the new version, running `rg '^\+'` on the diff output is sufficient.)

krobelus:
thanks, yeah I pushed that commit (except for the test changes) so the other PR can do `git reset --hard 0edcd51b0da75248f5a85868f5c5be5f310e1545` (if you have some local translation changes since 4dc39b95b5517e90e1dee0c5102ce11291859053 you'd need to rebase or redo them by hand)

danielrainer:
> I pushed that commit (except for the test changes)

Ah yes, the test changes should not be included in that commit.

Where did you push to? It does not seem to be on master. The commit hash you mention is also the one I pushed, which indicates that it still includes the removal of tests.

krobelus:
my test had failed, actually pushed now as 83af5c91bd56f4e1c2912db55c6094625de22061

milouse:
Actually not sure what I was required to do, now that this PR is closed, but related commits does not look to be in master.

So I took 5 minutes to rebase #11842 onto master (including the 83af5c91bd56f4e1c2912db55c6094625de22061 commit). From what I can tell, it seems to be ok for fr.po.

krobelus:
> Actually not sure what I was required to do, now that this PR is closed, but related commits does not look to be in master.
> 
> So I took 5 minutes to rebase #11842 onto master (including the 83af5c91bd56f4e1c2912db55c6094625de22061 commit). From what I can tell, it seems to be ok for fr.po.

The PR was closed but you could still have used the commits to rebase 
(throwing away 83af5c91bd56f4e1c2912db55c6094625de22061 during rebase
because an "almost-equivalent" commit made it to master).

Anyway, thanks for the update,
I guess we can wait bit in case we find reviewers,
but I guess that's not strictly necessary.
(I don't think we have a good way of notifying potentially-interested translators
we have an [atom feed](https://github.com/fish-shell/fish-shell/pull/11865) for each language now, for people who use this sort of thing).

danielrainer:
I pushed a new branch containing the intended changes from this PR, but based on current master, instead of the old commit which removes the length modifiers: https://github.com/danielrainer/fish-shell/tree/french-translation-new

If you don't have any new changes since this PR was opened, you can reset your branch to 4f5dd5d3857bbd6a5fa8220f98fd26cdd2fd9f3e. This also fixes the formatting issues causing test failures on your branch.

PR #11880, opened 2025-10-05
completion/set: fix bug preventing showing `history` or `fish_killring`

ilyagr:
Previously, `set -S fish_kill<TAB>` did not list `fish_killring`. This was because `$1` wasn't sufficiently escaped, and so instead of referring to a regex capture group, it was always empty.

## TODOs:
- [n/a?] User-visible changes noted in CHANGELOG.rst


krobelus:
nice, thanks. Maybe we should have nestable quoting syntax like Tcl

> - [n/a?]

yeah seems fine to skip

PR #11884, opened 2025-10-05
Bravely revert "CI: Disable some Cirrus CI jobs during RIIR transition"

zanchey:
Test this actually works on Cirrus

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -40,8 +40,7 @@ linux_arm_task:
         - FISH_TEST_MAX_CONCURRENCY=6 cmake -G Ninja -DCMAKE_BUILD_TYPE=Debug ..
         - ninja -j 6 fish
         - ninja fish_run_tests
-    # CI task disabled during RIIR transition
-    only_if: false && $CIRRUS_REPO_OWNER == 'fish-shell'
+    only_if: $CIRRUS_REPO_OWNER == 'fish-shell'
 
 freebsd_task:
     matrix:
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -6,12 +6,12 @@ linux_task:
     matrix:
         - name: alpine
           container: &step
-              image: ghcr.io/krobelus/fish-ci/alpine:latest
+              image: ghcr.io/fish-shell/fish-ci/alpine:latest
               memory: 4GB
         - name: jammy
           container:
               <<: *step
-              image: ghcr.io/krobelus/fish-ci/jammy:latest
+              image: ghcr.io/fish-shell/fish-ci/jammy:latest
     tests_script:
         # cirrus at times gives us 32 procs and 2 GB of RAM
         # Unrestriced parallelism results in OOM
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -28,9 +28,6 @@ linux_arm_task:
         - name: focal-arm64
           arm_container:
               image: ghcr.io/fish-shell/fish-ci/focal-arm64
-        - name: jammy-armv7-32bit
-          arm_container:
-              image: ghcr.io/fish-shell/fish-ci/jammy-armv7-32bit
     tests_script:
         # cirrus at times gives us 32 procs and 2 GB of RAM
         # Unrestriced parallelism results in OOM
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,3 +1,5 @@
+skip: $CIRRUS_REPO_OWNER == 'fish-shell' && $CIRRUS_BRANCH == 'master'
+
 env:
     CIRRUS_CLONE_DEPTH: 100
     CI: 1
--- a/.github/workflows/build_docker_images.yml
+++ b/.github/workflows/build_docker_images.yml
@@ -4,8 +4,6 @@ on:
   push:
     branches:
       - master
-    paths:
-      - 'docker/**'
   workflow_dispatch:
 
 concurrency:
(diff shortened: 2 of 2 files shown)


.cirrus.yml
```diff
@@ -49,8 +49,7 @@ linux_arm_task:
         - ninja -j 6 fish
         - file ./fish
         - ninja fish_run_tests
-    # CI task disabled during RIIR transition
-    only_if: false && $CIRRUS_REPO_OWNER == 'fish-shell'
+    only_if: $CIRRUS_REPO_OWNER == 'fish-shell'
```
krobelus, line 42: `focal-arm64` fails with `cmake: command not found`

Perhaps someone published new docker images since 08b03a733ab (docker_run_tests.sh: stop using cmake, 2025-06-19) was done. Anyway, I think we can fix this by calling `build_tools/check.sh` here, no need for CMake.
Note that `FISH_CHECK_LINT=false` is already set in the dockerfile.

Last time I updated the Linux builds above, I needed to update something in the docker images, which meant that I needed to publish them to ghcr.io. I haven't written down the steps but it should be done with `docker build` (as done by `docker/docker_run_tests.sh`) and `docker pull`, though these might require an ARM machine

zanchey: Yeah, I have a workflow in draft to make new builds and some commits to use `build_tools/check.sh`.

krobelus: So Linux builds are failing because `FISH_CHECK_LINT` wasn't set,
meaning that `check.sh` will run all warnings/formatters etc.

I'm running tests right now with `FISH_CHECK_LINT=false` set explicit:
https://github.com/fish-shell/fish-shell/tree/cirrus-ci
that should fix it.
(Of course we could check warnings on these platforms as well in future, but it's not very important).

So it looks like docker images will automatically be pushed to the registry whenever changes to `docker/` are pushed to master.

I assume they're already published (via https://github.com/fish-shell/fish-shell/actions/runs/18267743492, looks like the push of f6d93f2fdb0 (GitHub Actions: add workflow to build Docker images for CI, 2025-10-06) triggered it, which is surprising but it is what we want).

so I'll change the image paths back from my temporary ones:

```
diff --git a/.cirrus.yml b/.cirrus.yml
index 583d16ebd1..9cacfb4bdc 100644
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -6,20 +6,20 @@
     matrix:
         - name: alpine
           container: &step
-              image: ghcr.io/krobelus/fish-ci/alpine:latest
+              image: ghcr.io/fish-shell/fish-ci/alpine:latest
               memory: 4GB
```

.cirrus.yml
```diff
@@ -42,8 +42,7 @@ linux_arm_task:
         - lscpu || true
         - (cat /proc/meminfo | grep MemTotal) || true
         - FISH_TEST_MAX_CONCURRENCY=6 build_tools/check.sh
-    # CI task disabled during RIIR transition
-    only_if: false && $CIRRUS_REPO_OWNER == 'fish-shell'
+    only_if: $CIRRUS_REPO_OWNER == 'fish-shell'
```
krobelus, line 42: I think we want something like `only_if: repo owner is fish-shell` also for
`.github/workflows/docker_builds.yml`,
else these might run (and typically fail) whenever someone pushes to their github fork's `master` branch
(there's usually not too much need for anyone to do that, but people do and it does make some things easier for the lazy
(by default github workflows only run from master, unless `gh workflow run --ref` is used etc)

krobelus:
TODO: need to either re-revert b1e8fdfaa2c (Revert "CI: use build_tools/check.sh in Cirrus CI", 2025-10-06),
or keep using cmake (I have not had time to figure out what was the problem; some of them symptoms are documented in the commits I pushed to this PR).

The fact that `.github/workflows/docker_builds.yml` only writes to a central registry that's directly used by `master` is not ideal;
maybe we can set things up so it's easy to test everything in a temporary branch, e.g. by allowing `workflow_dispatch` triggers to use a custom docker registry namespace

zanchey:
> The fact that `.github/workflows/docker_builds.yml` only writes to a central registry that's directly used by `master` is not ideal; maybe we can set things up so it's easy to test everything in a temporary branch, e.g. by allowing `workflow_dispatch` triggers to use a custom docker registry namespace

You can test it in your own fork and it will write to your own package registry. A dispatch trigger would be a fine thing to add.

I was going to poke at this some more but it looks like you want to take it over?

krobelus:
> You can test it in your own fork and it will write to your own package registry

right, that should work (probably some of the ones in https://github.com/krobelus/fish-shell/actions/runs/18277577848 failed because I had never pushed those images before, so I get a permissions error).

> I was going to poke at this some more but it looks like you want to take it over?

Feel free to take it back (and drop the bad commit that's unrelated to Cirrus).
We could stick with cmake for now, to keep it close to current GitHub actions, but `check.sh` should work, and might be easier because I already removed CMake from the docker images

.cirrus.yml
```diff
@@ -6,20 +6,20 @@ linux_task:
...
         # - name: jammy-asan
         #   container:
         #       <<: *step
-        #       image: ghcr.io/krobelus/fish-ci/jammy-asan:latest
+        #       image: ghcr.io/fish-shell/fish-ci/jammy-asan:latest
         # - name: focal-32bit
         #   container:
         #       <<: *step
-        #       image: ghcr.io/krobelus/fish-ci/focal-32bit:latest
+        #       image: ghcr.io/fish-shell/fish-ci/focal-32bit:latest
```
krobelus: Ok it wasn't so bad; I only had to add write permissions on some of the docker images in my registry.
I'm more or less ready to push a version that has successfully passed tests on my fork.
Though I disabled `jammy-armv7-32bit`  for now, due to `exec format error`.. maybe this container

```
          - os: ubuntu-24.04-arm
            target: jammy-armv7-32bit
```

needs to be built on a 32bit host?

---

Whenever we make a change to `docker/` on master,
we need to first push those changes,
wait for the docker images to be pushed,
and then re-run CI.
If the previous push needs the new docker images, its CI run would fail.

It's much better than before, but ideally we can declare `build-docker-images` a prerequisite for `.cirrus.yml`,
so Cirrus waits for that. I wonder what's the best way to do that.
(One workaround is manual kickoff of `build-docker-images` on a non-master branch before pushing to master)

krobelus: > wait for the docker images to be pushed

pushed an untested draft commit for that.
Will need to test it with Cirrus on my fork,
and when that works add fish-shell's `CIRRUS_TOKEN` to github secrets.

PR #11888, opened 2025-10-05
Make command run by __fish_echo output to TTY for color detection

Xiretza:
## Description

Without this, e.g. Alt-L shows the directory entries one per line and without colors.

## TODOs:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,7 @@ This release fixes the following regressions identified in 4.1.0:
 - Added a workaround for old versions of Zellij where :kbd:`escape` processing was delayed (:issue:`11868`).
 - Fixed a case where the :doc:`web-based configuration tool <cmds/fish_config>` would generate invalid configuration (:issue:`11861`).
 - Fixed a case where upgrading fish would break old versions of fish that were still running.
+- Fix Alt+L binding not formatting ls output correctly (one entry per line, no colors).
 
   In general, fish still needs to be restarted after it is upgraded,
   except for `standalone builds <https://github.com/fish-shell/fish-shell/?tab=readme-ov-file#building-fish-with-embedded-data-experimental>`__.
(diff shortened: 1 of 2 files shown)


CHANGELOG.rst
```diff
@@ -12,6 +12,7 @@ This release fixes the following regressions identified in 4.1.0:
 - Added a workaround for old versions of Zellij where :kbd:`escape` processing was delayed (:issue:`11868`).
 - Fixed a case where the :doc:`web-based configuration tool <cmds/fish_config>` would generate invalid configuration (:issue:`11861`).
 - Fixed a case where upgrading fish would break old versions of fish that were still running.
+- Fix Alt+L binding not formatting ls output correctly (one entry per line, no colors).
```
krobelus, line 15: FWIW our nice formatting syntax would be :kbd:`ls`

share/functions/__fish_echo.fish
```diff
@@ -10,7 +10,8 @@ set -l erase_line "$(
 function __fish_echo --inherit-variable erase_line --description 'run the given command after the current commandline and redraw the prompt'
     set -l line (commandline --line)
     string >&2 repeat -N \n --count=(math (commandline | count) - $line + 1)
-    printf %s\n $erase_line($argv) >&2
+    printf %s $erase_line >&2
+    $argv >&2
```
krobelus, line 14: Oof, my bad.

As written, this fails to clear lines beyond the first one.  This is
not a problem right now (so it should also be fine to revert
1fdf37cc4a2 (__fish_echo: fully overwrite lines, 2025-09-17),
because IIRC clearing the first line wasn't needed because fish
already does that. But it can't hurt, so LGTM as is.)

Going forward we should probably clear to end of string here, so `\e[J` AKA `tput ed`.
But I'm happy to add that later, only when the fix mentioned in 1fdf37cc4a2 actually needed.

Xiretza:
Thanks for taking care of that!

PR #11889, opened 2025-10-05
sprintf: remove signed int size info

danielrainer:
The size was used to keep track of the number of bits of the input type the `Arg::SInt` variant was created from. This was only relevant for arguments defined in Rust, since the `printf` command takes all arguments as strings.

The only thing the size was used for is for printing negative numbers with the `x` and `X` format specifiers. In these cases, the `i64` stored in the `SInt` variant would be cast to a `u64`, but only the number of bits present in the original argument would be kept, so `-1i8` would be formatted as `ff` instead of `ffffffffffffffff`.

There are no users of this feature, so let's simplify the code by removing it. While we're at it, also remove the unused `bool` returned by `as_wrapping_sint`.

https://github.com/fish-shell/fish-shell/pull/11874#discussion_r2404524838

### DIFF
--- a/crates/printf/src/tests.rs
+++ b/crates/printf/src/tests.rs
@@ -225,7 +225,7 @@ fn test_int() {
     assert_fmt!("%d", -123 => "-123");
     assert_fmt!("~%d~", 148 => "~148~");
     assert_fmt!("00%dxx", -91232 => "00-91232xx");
-    assert_fmt!("%x", -9232 => "ffffdbf0");
+    assert_fmt!("%x", -9232 => "ffffffffffffdbf0");
     assert_fmt!("%X", 432 => "1B0");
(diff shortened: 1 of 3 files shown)


crates/printf/src/tests.rs
```diff
@@ -225,7 +225,7 @@ fn test_int() {
     assert_fmt!("%d", -123 => "-123");
     assert_fmt!("~%d~", 148 => "~148~");
     assert_fmt!("00%dxx", -91232 => "00-91232xx");
-    assert_fmt!("%x", -9232 => "ffffdbf0");
+    assert_fmt!("%x", -9232 => "ffffffffffffdbf0");
```
krobelus, line 228: probably this should be a compile time error in future

Issue #11881, opened 2025-10-05
The `tmux-mutliline-prompt` check reliably fails on my machine


ilyagr:
At commit 623c14a, when I run `make test`, I get some flakes, but I also reliably get the following error in `checks/tmux-multiline-prompt.fish`:

```
Failure:

  The CHECK on line 60 wants:
    prompt-line-2> echo Hello World

  which failed to match line stdout:8:
    echo Hello Worlecho Hello World

  Context:
    [...] from line 47 (stdout:3):
    prompt-line-1
    prompt-line-2>
    prompt-line-1
    echo Hello Worlecho Hello World <= does not match CHECK on line 60: prompt-line-2> echo Hello World
    Hello World
    prompt-line-1
    prompt-line-2>

  when running command:
    /Users/ilyagr/dev/fish-shell/build/fish /Users/ilyagr/dev/fish-shell/tests/checks/tmux-multiline-prompt.fish
```

I'm using a prerelease version of tmux from commit https://github.com/tmux/tmux/commit/3e28777ecbc8de3df803d5d7429a5ecd35f61553.

This is on macOS Sonoma-ARM. I may or may not try investigating it further myself (e.g. see whether this occurs for me with the latest tmux release and, if not, whether tmux's behavior improved or got worse).

krobelus:
yeah this (and other) tests are not run in CI unfortunately

It works for me consistently on macOS Ventura x86 13.7.8 (steps below)
Since you can reproduce the failure, can you maybe try to reduce the test or any difference in environment.
In this test, the "isolated-tmux send-keys C-r" is supposed to set the command line and force-redraw with ctrl-l.
I don't know why that would not be working.
maybe log "isolated-tmux capture-pane" output to a file, add sleeps etc.

	ssh mac 'sh -lc "
	set -ex
	cd ~/git
	brew install automake libevent utf8proc pkgconf
	git clone https://github.com/tmux/tmux
	cd tmux
	git checkout 3e28777
	./autogen.sh
	./configure --enable-utf8proc
	make
	test -x ./tmux
	cd ../fish-shell
	git checkout 623c14a
	cargo b
	PATH=$HOME/git/tmux:$PATH \
	    tests/test_driver.py target/debug tests/checks/tmux-multiline-prompt.fish
	"'

PR #11891, opened 2025-10-05
checks: make `tmux-multiline-prompt` less affected by `less` config

ilyagr:
See commit description. This is a more didactic version of the fix. If we can make sure `less` uses stock config, that might be better, but `less` seems to have several ways to be configured, and I don't fully understand it.

Fixes issue #11881

Thanks, @krobelus, for the help with debugging this!

### DIFF
--- a/tests/checks/tmux-multiline-prompt.fish
+++ b/tests/checks/tmux-multiline-prompt.fish
@@ -50,7 +50,7 @@ isolated-tmux capture-pane -p | tail -n 5
 # CHECK:
 
 # Test repainint after running an external program that uses the alternate screen.
-isolated-tmux send-keys "bind ctrl-r 'echo | less +q; commandline \"echo Hello World\"'" Enter C-l
+isolated-tmux send-keys "bind ctrl-r 'echo | less -+F -+X +q; commandline \"echo Hello World\"'" Enter C-l
 isolated-tmux send-keys C-r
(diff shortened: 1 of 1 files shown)


ilyagr:
Aside: I also am not sure that test works as intended, or at least something is slightly weird. My understanding is that the command

```
echo lala\nbb\ncc | LESS= less -X +q
```

should print something to the screen, but it doesn't. So, I have no idea whether


```
echo lala\nbb\ncc | LESS= less -+X +q
```

does anything either.

image


Maybe it'd be better to use Python or a `fish -c "printf ..."` command to do the alternate screen stuff. But I'm not very familiar with the details.

ilyagr:
(On the topic of last comment)

This is nonsense to me, but perhaps it will make some sense to you.

image

(You probably already know, but `unbuffer` is a tool that's usually bundled with `expect`. There is also a more-or-less equivalent `pipetty` tool from https://github.com/kilobyte/colorized-logs, in case it's easier to install for some people)

krobelus:
I remember that this somehow only reproduced on programs that use the alternate screen like this.
That's the 1049 sequences (fish also [uses them](https://fishshell.com/docs/4.1/terminal-compatibility.html)).

It seems difficult to completely isolate less from the environment; I can try to create a fish+printf based reproducer for the original problem.

PR #11899, opened 2025-10-07
check.sh: export gettext extraction file variable

### LINKED ISSUE #11896: Rust gettext extraction failing on FreeBSD
Some commits have recently been pushed to master that started using `build_tools/check.sh` on Cirrus CI,
instead of the CMake `fish_run_tests` target.

One of the behavior changes is that now the `tests/checks/po-files-up-to-date.fish` test is actually run.
It fails because it looks like it removes all the Rust-extracted localizable strings except for
`Print the working directory`

Maybe some behavior difference in msgfmt/msguniq/msgattrib/msgmerge ?

https://cirrus-ci.com/task/6433371993145344?logs=tests#L997

cc @danielrainer 

I still need to figure out how to easily get SSH access to Cirrus CI. Maybe it's possible to reproduce by installing the FreeBSD version of gettext..

danielrainer:
Some versions of `/bin/sh`, e.g. the one on FreeBSD, do not propagate variables set on a command through shell functions. This results in `FISH_GETTEXT_EXTRACTION_FILE` being set in the `cargo` function, but not for the actual `cargo` process spawned from the function, which breaks our localization scripts and tests. Exporting the variable prevents that.

Fixes #11896

### DIFF
--- a/build_tools/check.sh
+++ b/build_tools/check.sh
@@ -49,7 +49,8 @@ if [ -n "$FISH_TEST_MAX_CONCURRENCY" ]; then
 fi
 
 template_file=$(mktemp)
-FISH_GETTEXT_EXTRACTION_FILE=$template_file cargo build --workspace --all-targets --features=gettext-extract
+export FISH_GETTEXT_EXTRACTION_FILE="$template_file"
+cargo build --workspace --all-targets --features=gettext-extract
 if $lint; then
     PATH="$build_dir:$PATH" "$workspace_root/build_tools/style.fish" --all --check
     for features in "" --no-default-features; do
@@ -61,7 +62,7 @@ cargo test --doc --workspace
 if $lint; then
     cargo doc --workspace
 fi
-FISH_GETTEXT_EXTRACTION_FILE=$template_file "$workspace_root/tests/test_driver.py" "$build_dir"
+"$workspace_root/tests/test_driver.py" "$build_dir"
 
 exit
 }


build_tools/check.sh
```diff
@@ -49,7 +49,8 @@ if [ -n "$FISH_TEST_MAX_CONCURRENCY" ]; then
 fi
 
 template_file=$(mktemp)
-FISH_GETTEXT_EXTRACTION_FILE=$template_file cargo build --workspace --all-targets --features=gettext-extract
+export FISH_GETTEXT_EXTRACTION_FILE="$template_file"
+cargo build --workspace --all-targets --features=gettext-extract
```
krobelus, line 53: Oof, so this is unspecified in POSIX
From https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html

I can reproduce the bug by changing the shebang to `#!/bin/ksh`

 > Variable assignments shall be performed as follows:
 >
 > [...]
 >
 > If the command name is a function that is not a standard utility
 > implemented as a function, variable assignments shall affect
 > the current execution environment during the execution of the
 > function. It is **unspecified**:
 > 
 > Whether or not the variable assignments persist after the completion of the function
 > 
 > Whether or not the variables **gain the export attribute** during the execution of the function
 > 
 > Whether or not export attributes gained as a result of the variable assignments persist after the completion of the function (if variable assignments persist after the completion of the function)

which basically means we should never use "VAR=val fun"
unless we're fine with random scope pollution.

build_tools/check.sh
```diff
@@ -61,7 +62,7 @@ cargo test --doc --workspace
 if $lint; then
     cargo doc --workspace
 fi
-FISH_GETTEXT_EXTRACTION_FILE=$template_file "$workspace_root/tests/test_driver.py" "$build_dir"
+"$workspace_root/tests/test_driver.py" "$build_dir"
```
krobelus, line 65: I think I'd rather keep passing this explicitly every time rather
than exporting it for everything.

```
diff --git a/build_tools/check.sh b/build_tools/check.sh
index a258542306..ec8bf1226c 100755
--- a/build_tools/check.sh
+++ b/build_tools/check.sh
@@ -49,7 +49,10 @@
 fi
 
 template_file=$(mktemp)
-FISH_GETTEXT_EXTRACTION_FILE=$template_file cargo build --workspace --all-targets --features=gettext-extract
+(
+    export FISH_GETTEXT_EXTRACTION_FILE="$template_file"
+    cargo build --workspace --all-targets --features=gettext-extract
+)
 if $lint; then
     PATH="$build_dir:$PATH" "$workspace_root/build_tools/style.fish" --all --check
     for features in "" --no-default-features; do
```

Could also pass this environment variable only to the
test that needs it; though unless we run it in the background,
this would add 10 seconds, so I don't think that's worth it.

```diff
@@ -61,7 +64,10 @@
 if $lint; then
     cargo doc --workspace
 fi
-FISH_GETTEXT_EXTRACTION_FILE=$template_file "$workspace_root/tests/test_driver.py" "$build_dir"
+FISH_GETTEXT_EXTRACTION_FILE=$template_file \
+    "$workspace_root/tests/test_driver.py" "$build_dir" \
+        "$workspace_root/tests/checks/po-files-up-to-date.fish"
+"$workspace_root/tests/test_driver.py" "$build_dir"
 
 exit
 }
```

Issue #11875, opened 2025-10-04
Lines not cleared with multi-line initial prompt and single-line transient prompt


rightaditya:
Fish version: 4.1.1
OS: CachyOS (Linux, Arch-based)
Issue is reproducible in a fresh environment with correspondingly set `fish_prompt` and `transient_prompt`.

Lines from an initial multi-line prompt aren't properly cleared if a transient prompt is used and the final prompt only has one line. The exact effect depends on the relative line lengths and whether (and where) auto-completion is used. Obviously `fish_transient_prompt` must be set to `1`.

With the following `fish_prompt`:
```fish
function fish_prompt
  if contains -- --final-rendering $argv
    echo '> '
  else
    echo $PWD
    echo '> '
  end
end
```
And running `echo test` with working directory `/usr/share/fish` like so:
```console
/usr/share/fish
> echo test
```
Results in the following:
```console
> /share/fish
testho test
/usr/share/fish
> 
```
When instead it should be:
```console
/usr/share/fish
> echo test
test
/usr/share/fish
> 
```
Recording in case that's clearer: <https://asciinema.org/a/gOFhWfXkGfW1I3SzHJISNPJBg>

Repeating the test with the following `fish_prompt`:
```fish
function fish_prompt
  if contains -- --final-rendering $argv
    echo '> '
  else
    echo $PWD
    echo '>> '
  end
end
```
We get:
```console
> echo test
testcho test
/usr/share/fish
>> 
```
When instead it should be:
```console
/usr/share/fish
> echo test
test
/usr/share/fish
>> 
```
Recording: <https://asciinema.org/a/cPuZSLgfxPnsahisw5wG6RFJF>

I went through the code and think it should be a relatively simple fix. I don't have any experience with Rust, but I've been Rust-curious for years so I'd be happy to submit a PR. But I understand if you'd rather just fix it and move on than deal with having to clean up a newb's PR :)

krobelus:
I can reproduce; this regressed in 267b16235d3 (Fix transient prompt mode for single-line prompts, 2025-04-14) my bad (cc @kerty0, I think I took your patch verbatim, not yet sure how it ended up having the opposite effect).
Fortunately we have a test for both now.

golimpio:
I'm experiencing what seems to be the same issue, in my case, I have a new line before the prompt.

Currently the output is:

```shell
 ❯ ls
  config.json    config.json.bak
 ❯ ls
  config.json    config.json.bak

~/.config/omp
[...]

PR #11901, opened 2025-10-07
Clear lines when final prompt has fewer lines than initial

rightaditya:
## Description

Lines aren't cleared properly if a transient prompt is used where the final prompt has fewer lines than the initial prompt. This PR fixes that by explicitly checking for that scenario and explicitly clearing the screen and lines, as well as adjusting for skipped prefixes if an input line is indented.

Fixes issue #11875.

## TODOs:

### DIFF
--- a/src/screen.rs
+++ b/src/screen.rs
@@ -16,7 +16,7 @@ use std::collections::LinkedList;
 use std::io::Write;
 use std::ops::Range;
 use std::sync::atomic::AtomicU32;
-use std::sync::Mutex;
+use std::sync::{Mutex, MutexGuard};
 use std::time::SystemTime;
 
 use libc::{ONLCR, STDERR_FILENO, STDOUT_FILENO};
@@ -470,7 +470,7 @@ impl Screen {
             } else {
                 0
             }
-            + calc_prompt_lines(&layout.left_prompt);
+            + calc_prompt_lines(None, &layout.left_prompt);
         let pager_available_height = std::cmp::max(
             1,
             curr_termsize
@@ -536,7 +536,7 @@ impl Screen {
         let prompt_line_count = self
             .actual_left_prompt
             .as_ref()
-            .map_or(1, |p| calc_prompt_lines(p));
+            .map_or(1, |p| calc_prompt_lines(None, p));
         self.actual.cursor.y += prompt_line_count.checked_sub(1).unwrap();
     }
 
@@ -570,7 +570,7 @@ impl Screen {
         let trailing_prompt_lines = self
             .actual_left_prompt
             .as_ref()
-            .map_or(0, |p| calc_prompt_lines(p) - 1);
+            .map_or(0, |p| calc_prompt_lines(None, p) - 1);
         let lines_to_scroll = prompt_y
             .checked_sub(trailing_prompt_lines)
             .unwrap_or_else(|| {
@@ -1099,21 +1099,13 @@ impl Screen {
         // Compare the number of lines in the (new) left prompt compared to the
         // current prompt before it's dropped so that the lines can be cleared
         // appropriately later (see #11875).
-        let mut final_prompt_has_fewer_lines = false;
-        if is_final_rendering {
-            let actual_prompt_line_breaks = self.actual_left_prompt.as_ref().map_or(0, |p| {
-                let mut result = 0;
-                if p.chars().any(|c| matches!(c, '\n' | '\x0C')) {
-                    result = cached_layouts
-                        .calc_prompt_layout(p, None, usize::MAX)
-                        .line_breaks
-                        .len();
-                }
-                result
-            });
-            final_prompt_has_fewer_lines |=
-                left_prompt_layout.line_breaks.len() < actual_prompt_line_breaks;
-        }
+        let final_prompt_has_fewer_lines = is_final_rendering && {
+            let actual_prompt_line_breaks = self
+                .actual_left_prompt
+                .as_ref()
+                .map_or(0, |p| calc_prompt_lines(Some(&mut cached_layouts), p) - 1);
+            left_prompt_layout.line_breaks.len() < actual_prompt_line_breaks
+        };
 
         // Determine how many lines have stuff on them; we need to clear lines with stuff that we don't
         // want.
@@ -1875,19 +1867,24 @@ fn truncate_run(
     *width = curr_width;
 }
 
-fn calc_prompt_lines(prompt: &wstr) -> usize {
+fn calc_prompt_lines(
+    layout_cache: Option<&mut MutexGuard<'_, LayoutCache>>,
+    prompt: &wstr,
+) -> usize {
     // Hack for the common case where there's no newline at all. I don't know if a newline can
     // appear in an escape sequence, so if we detect a newline we have to defer to
     // calc_prompt_width_and_lines.
     let mut result = 1;
     if prompt.chars().any(|c| matches!(c, '\n' | '\x0C')) {
-        result = LAYOUT_CACHE_SHARED
-            .lock()
-            .unwrap()
+        let mut local_cache = None;
+        let layout_cache = layout_cache.unwrap_or_else(|| {
+            local_cache = Some(LAYOUT_CACHE_SHARED.lock().unwrap());
+            local_cache.as_mut().unwrap()
+        });
+        result += layout_cache
             .calc_prompt_layout(prompt, None, usize::MAX)
             .line_breaks
-            .len()
-            + 1;
+            .len();
     }
     result
 }


rightaditya:
I haven't worked with Rust before so please forgive anything non-idiomatic, etc.

src/screen.rs
```diff
@@ -1078,10 +1078,30 @@ impl Screen {
         self.need_clear_lines = false;
         self.need_clear_screen = false;
 
+        // Compare the number of lines in the (new) left prompt compared to the
```
krobelus, line 1099: PR description should go into commit message (notably the "Fixes
#ISSUE_NUMBER" would be useful to autoclose the issue if the PR gets
pushed without the github-native merge methods)

src/screen.rs
```diff
@@ -1078,10 +1078,30 @@ impl Screen {
...
+        // current prompt before it's dropped so that the lines can be cleared
+        // appropriately later (see #11875). Avoid unnecessary clearing by only
```
krobelus: LGTM, nice work. Add a changelog entry for 4.2 (i.e. "fish ?.?.?") if you want.
(Not sure if it's a good idea to add this to a patch release)

src/screen.rs
```diff
@@ -1078,10 +1078,30 @@ impl Screen {
...
+        // doing this when `is_final_rendering`.
```
krobelus: lose the "Avoid unnecessary ..." comment, that seems too obvious

src/screen.rs
```diff
@@ -1078,10 +1078,30 @@ impl Screen {
...
+        let mut prompt_has_fewer_lines = is_final_rendering;
```
krobelus: If my other suggestions make sense,
then the variable name is not super honest.
Better make it `final_prompt_has_fewer_lines`.

src/screen.rs
```diff
@@ -1078,10 +1078,30 @@ impl Screen {
...
+        if is_final_rendering {
+            let actual_prompt_line_breaks = self.actual_left_prompt.as_ref().map_or(1, |p| {
+                let mut result = 1;
```
krobelus: this is inconsistent with the variable name; it ought to be 0 (assuming there is no better solution)

src/screen.rs
```diff
@@ -1078,10 +1078,30 @@ impl Screen {
...
+                if p.chars().any(|c| matches!(c, '\n' | '\x0C')) {
+                    result = cached_layouts
+                        .calc_prompt_layout(p, None, usize::MAX)
+                        .line_breaks
+                        .len();
```
krobelus: probably not a good idea to copy-paste calc_prompt_lines(); can we
change it to take `cached_layouts: &MutexGuard<'_, LayoutCache>`
and reuse it here

Here's my attempt

```diff
diff --git a/src/screen.rs b/src/screen.rs
index aea6980805..72dff0bbb6 100644
--- a/src/screen.rs
+++ b/src/screen.rs
@@ -16,7 +16,7 @@
 use std::io::Write;
 use std::ops::Range;
 use std::sync::atomic::AtomicU32;
-use std::sync::Mutex;
+use std::sync::{Mutex, MutexGuard};
 use std::time::SystemTime;
 
 use libc::{ONLCR, STDERR_FILENO, STDOUT_FILENO};
@@ -470,7 +470,7 @@
             } else {
                 0
             }
-            + calc_prompt_lines(&layout.left_prompt);
+            + calc_prompt_lines(None, &layout.left_prompt);
         let pager_available_height = std::cmp::max(
             1,
             curr_termsize
@@ -536,7 +536,7 @@
         let prompt_line_count = self
             .actual_left_prompt
             .as_ref()
-            .map_or(1, |p| calc_prompt_lines(p));
+            .map_or(1, |p| calc_prompt_lines(None, p));
         self.actual.cursor.y += prompt_line_count.checked_sub(1).unwrap();
     }
 
@@ -570,7 +570,7 @@
         let trailing_prompt_lines = self
             .actual_left_prompt
             .as_ref()
-            .map_or(0, |p| calc_prompt_lines(p) - 1);
+            .map_or(0, |p| calc_prompt_lines(None, p) - 1);
         let lines_to_scroll = prompt_y
             .checked_sub(trailing_prompt_lines)
             .unwrap_or_else(|| {
@@ -1098,23 +1098,14 @@
 
         // Compare the number of lines in the (new) left prompt compared to the
         // current prompt before it's dropped so that the lines can be cleared
-        // appropriately later (see #11875). Avoid unnecessary clearing by only
-        // doing this when `is_final_rendering`.
-        let mut prompt_has_fewer_lines = is_final_rendering;
-        if is_final_rendering {
-            let actual_prompt_line_breaks = self.actual_left_prompt.as_ref().map_or(1, |p| {
-                let mut result = 1;
-                if p.chars().any(|c| matches!(c, '\n' | '\x0C')) {
-                    result = cached_layouts
-                        .calc_prompt_layout(p, None, usize::MAX)
-                        .line_breaks
-                        .len();
-                }
-                result
-            });
-            prompt_has_fewer_lines |=
-                left_prompt_layout.line_breaks.len() < actual_prompt_line_breaks;
-        }
+        // appropriately later (see #11875).
+        let prompt_has_fewer_lines = is_final_rendering && {
+            let actual_prompt_line_breaks = self
+                .actual_left_prompt
+                .as_ref()
+                .map_or(0, |p| calc_prompt_lines(Some(&mut cached_layouts), p) - 1);
+            left_prompt_layout.line_breaks.len() < actual_prompt_line_breaks
+        };
 
         // Determine how many lines have stuff on them; we need to clear lines with stuff that we don't
         // want.
@@ -1876,19 +1867,24 @@
     *width = curr_width;
 }
 
-fn calc_prompt_lines(prompt: &wstr) -> usize {
+fn calc_prompt_lines(
+    layout_cache: Option<&mut MutexGuard<'_, LayoutCache>>,
+    prompt: &wstr,
+) -> usize {
     // Hack for the common case where there's no newline at all. I don't know if a newline can
     // appear in an escape sequence, so if we detect a newline we have to defer to
     // calc_prompt_width_and_lines.
     let mut result = 1;
     if prompt.chars().any(|c| matches!(c, '\n' | '\x0C')) {
-        result = LAYOUT_CACHE_SHARED
-            .lock()
-            .unwrap()
+        let mut local_cache = None;
+        let layout_cache = layout_cache.unwrap_or_else(|| {
+            local_cache = Some(LAYOUT_CACHE_SHARED.lock().unwrap());
+            local_cache.as_mut().unwrap()
+        });
+        result += layout_cache
             .calc_prompt_layout(prompt, None, usize::MAX)
             .line_breaks
-            .len()
-            + 1;
+            .len();
     }
     result
 }
```

src/screen.rs
```diff
@@ -1078,10 +1078,30 @@ impl Screen {
...
+                }
+                result
+            });
+            prompt_has_fewer_lines |=
+                left_prompt_layout.line_breaks.len() < actual_prompt_line_breaks;
+        }
```
krobelus: this whole if-block doesn't do anything because
`prompt_has_fewer_lines` is always true at this point.
Judging by the comments, you want to initialize `prompt_has_fewer_lines` to `false`.

tests/checks/tmux-transient-prompt.fish
```diff
@@ -53,3 +53,45 @@ isolated-tmux capture-pane -p
...
+isolated-tmux send-keys C-u C-l '
+    function fish_prompt
+        if contains -- --final-rendering $argv
+            echo "2> "
+        else
+            echo "transient prompt line"
+            echo "1> "
+        end
+    end
+' C-l
```
krobelus: no need to repeat the same fish_prompt definition IMO (we accept
some interdependencies in these tests, that's not the worst part
about them..)

src/screen.rs
rightaditya: @krobelus Sure, I just force-pushed this and the other recommended changes. I put it under "Notable improvements and fixes" as that seemed like the most appropriate section from previous major releases, but TBH it felt odd calling this "notable." FWIW this feels like a minor regression fix to me, more appropriate to a patch release, but I'd wager I'm not particularly well-calibrated on that front...

rightaditya: Hmm, #11911 might be the "right" fix for this bug, but seems bigger. (NB: I didn't look at it super closely) I guess if that one is the way forward but makes sense for a minor version release, maybe this one can work as a hold-over in a patch release until then.

In case it's helpful (for cherry picking, etc.), I put the changes to the tests in a dedicated commit.

krobelus: https://github.com/fish-shell/fish-shell/pull/11911 is looking pretty good already,
so we'll go with that one. No need for patch release since it won't be super long either way.

> In case it's helpful (for cherry picking, etc.), I put the changes to the tests in a dedicated commit

thanks, we cherry-picked the test

Issue #11805, opened 2025-09-15
`fish_config` not working in msys2


Nahor:
```
> fish --version
fish, version 4.0.6-1 (Built by MSYS2 project)
> uname
MINGW64_NT-10.0-26100
> command -s (__fish_anypython)
/ucrt64/bin/python3
> /ucrt64/bin/python3 --version
Python 3.12.11
```
Note: also happened with 4.0.2. I haven't tried older versions of `fish`

When running `fish_config`, as soon as the browser connects, the message "Shutting down" shows up and the browser is indefinitely waiting for the page to load.
Despite the message, `fish_config` does NOT shutdown until I press `ENTER`. At that point I get the shell prompt and the browser shows a "Unable to connect" error message.

I looked quickly at `webconfig.py` and the shutdown happens because `sig_fileno` is in `ready_read` instead of `http_fileno`.

```
ready_read: [576], [], []
http_fileno: 444
sig_fileno: 576
```

As for `fish_config` not immediately returning to the prompt, it's because the app is blocked on the last `thread.join()`

krobelus:
>  the shutdown happens because sig_fileno is in ready_read instead of http_fileno.

yeah, if `sig_fileno` is in `ready_read`, then the code assumes that the user has pressed Enter.
You mention that this happens even if you don't press enter, so clearly something wasn't quite right with 8009469f8f3 (Add support for the Windows MSYS environment., 2023-11-20).

cc @ahaoboy; if anyone wants to look into this, it should be easy to test with `cargo build && target/debug/fish -c fish_config`

Nahor:
Found the issue: the webserver and the signal socket both use port 8000 (`PORT: 8000, sig_port: 8000`). So the browser was connecting to the signal socket, causing `web_config` to shutdown. If I set `sig_port` to start a 8001, then it can connect just fine.

I tried `sck.setsockopt(socket.SOL_SOCKET, socket.SO_EXCLUSIVEADDRUSE, 1)` but that hasn't prevented the socket from still being bound to 8000 for some reason.

As a fix for me, and as more efficient code regardless, the socket range should start at `PORT+1` since we know that the ports in the [8000..PORT] range are either not available or taken by the http server already

ahaoboy:
To be honest, I am not familiar with py, so there may be some omissions in the previous PR, but I tested both versions locally and both work correctly.

fish, version 4.0.2-1 (Built by MSYS2 project)

[...]

PR #11907, opened 2025-10-07
Fixes for MSYS2

Nahor:
- Fix MSYS2 compilation (missing ulimits)
- Fix port selection in fish_config that did not work for MINGW python
- Add a minimal test for fish_config as requested in #11805
- Add a github action to build MSYS2. This does not run the tests because too many are failing for me (and worse, the test just freeze/deadlock when checking universal variables)

Fixes issue #11805 

## TODOs:
- [ ] ~~Changes to fish usage are reflected in user documentation/manpages.~~ _[None]_
> [!NOTE]
> One test was added, to ensure `fish_config` can be shutdown (as requested in #11805), but it does not test that a browser can access the server (because this would add a `#REQUIRES` for a new command, e.g. `wget`/`curl`).
> I do not know if that one test is sufficient for this PR, so for now, I'm not checking that TODO
- [ ] ~~User-visible changes noted in CHANGELOG.rst~~ _[None]_

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -174,3 +174,30 @@ jobs:
     - name: make fish_run_tests
       run: |
         make -C build VERBOSE=1 fish_run_tests
+
+  windows:
+
+    runs-on: windows-latest
+
+    defaults:
+      run:
+        shell: msys2 {0}
+    steps:
+    - uses: actions/checkout@v4
+    - uses: msys2/setup-msys2@v2
+      with:
+        update: true
+        msystem: MSYS
+    - name: Install deps
+      # Not using setup-msys2 `install` option to make it easier to copy/paste
+      run: |
+        pacman --noconfirm -S --needed git rust
+    - name: cargo build
+      run: |
+        cargo build
+    - name: smoketest
+      # We can't use `cargo test` yet, there are just too many failures
+      # so this is just a quick check to make sure that fish can swim
+      run: |
+        set -x
+        [ "$(target/debug/fish.exe -c 'echo (math 1 + 1)')" = 2 ]


src/builtins/ulimit.rs
```diff
@@ -19,11 +19,23 @@ pub mod limits {
...
         pub const NOFILE: libc::c_int = libc::RLIMIT_NOFILE as _;
         pub const STACK: libc::c_int = libc::RLIMIT_STACK as _;
         pub const CPU: libc::c_int = libc::RLIMIT_CPU as _;
-        pub const NPROC: libc::c_int = libc::RLIMIT_NPROC as _;
+        cfg_if!(
+            if #[cfg(cygwin)] {
+                pub const NPROC: libc::c_int = -1;
+            } else {
+                pub const NPROC: libc::c_int = libc::RLIMIT_NPROC as _;
+                }
```
Berrysoft:
```suggestion
            }
```

Berrysoft:
> This does not run the tests because too many are failing for me...

Me, too. Some of them might be caused by MSYS2 which doesn't respect the file permissions like cygwin, but it's harder to setup a testing environment for fish on original cygwin. I don't know if cygwin/MSYS2 was tested in CI before fish 4.0, but if it wasn't, it's an acceptable status currently (maybe).

.github/workflows/test.yml
```diff
@@ -174,3 +174,23 @@ jobs:
...
+  windows:
+
+    runs-on: windows-latest
+
+    defaults:
+      run:
+        shell: msys2 {0}
+    steps:
+    - uses: actions/checkout@v4
+    - uses: msys2/setup-msys2@v2
```
krobelus, line 187: Cool so I assume this will give a Bash shell, as opposed to the default (powershell?)

.github/workflows/main.yml
```diff
@@ -174,3 +174,23 @@ jobs:
...
+
+    defaults:
+      run:
+        shell: msys2 {0}
+    steps:
+    - uses: actions/checkout@v4
+    - uses: msys2/setup-msys2@v2
+      with:
+        update: true
+        install: >-
```
krobelus: what does the `>-` syntax do? convert barewords to list?

.github/workflows/test.yml
```diff
@@ -174,3 +174,23 @@ jobs:
...
+      with:
+        update: true
+        install: >-
+          git rust
+        msystem: MSYS
```
krobelus, line 190: these translate to `pacman -S git rust` commands right?
Could we use those (so it's easier for me to copy-paste these steps onto any Windows system).
Unless there's a relevant difference of course.

.github/workflows/test.yml
```diff
@@ -174,3 +174,23 @@ jobs:
...
+    - uses: actions/checkout@v4
+    - name: cargo build
+      run: |
+        cargo build
```
krobelus, line 197: Thanks, this is already a great step in the right direction.
We could even add a smoke test like

	set -x
	[ "$(target/debug/fish.exe -c 'echo (math 1 + 1)')" = 2 ]


(FWIW debug builds currently only work due to 012b5071286 (Workaround
for embed-data debug builds on Cygwin, 2025-09-16) which we should
get rid of.. there is some problem with Windows paths in rust-embed)

tests/checks/tmux-fish_config.fish
```diff
@@ -0,0 +1,26 @@
+#RUN: %fish %s
+
+#REQUIRES: command -v tmux
+isolated-tmux-start
```
krobelus, line 7: > One test was added, to ensure `fish_config` can be shutdown (as
 > requested in #11805), but it does not test that a browser can access
 > the server (because this would add a `#REQUIRES` for a new command,
 > e.g. `wget`/`curl`).
 > I do not know if that one test is sufficient for this PR, so for
 > now, I'm not checking that TODO

Right, if it's easy to add `curl localhost:$port | grep some line`,
that sounds like a nice addition to this test.

But I'm happy to merge without a faithful test.
Historically, this functionality sometimes broke in Cygwin and WSL,
where we don't run these tests in the CI anyway yet.

tests/checks/tmux-fish_config.fish
```diff
@@ -0,0 +1,26 @@
...
+
+isolated-tmux send-keys "BROWSER=true fish_config" Enter
+tmux-sleep
+# Note the results will be wrapped because we only have 80 columns, so use `sed`
+# to shorten long lines (first, convert to a single line, then shorten the file
+# URL, then the HTTP one)
+isolated-tmux capture-pane -p | sed \
+    -e '1h;2,$H;$!d;g' \
+    -e 's|file://.*\.html|<file_url>|' \
+    -e 's|http://.*/\n|<http_url>\n|'
```
krobelus: probably awk would be more readable than sed magic but sure,
if it works on BSD sed too.

FWIW the 80 columns are not set in stone;
it's from the `isolated-tmux new-session -x 80 -y 10` line
For example the test could change it with (untested)

	isolated-tmux resize-window -x 500

which is probably a better solution than working around it in sed?

krobelus:
> I don't know if cygwin/MSYS2 was tested in CI before fish 4.0

only if downstream (Cygwin packagers) tested it

.github/workflows/test.yml
Berrysoft: > there is some problem with Windows paths in rust-embed

If it is built with rustc shipped with MSYS2, the paths should be in POSIX style, I assume?

.github/workflows/test.yml
Nahor: Yes

.github/workflows/main.yml
Nahor: I just copy/pasted what the tutorial for MSYS2 github action does.
And yes, it converts the multi-line word list into a single-line list (`>`) without a newline at the end ('-'). Not really needed here but that will be one less line change if we need to add more packages and organize them

.github/workflows/test.yml
Nahor: In essence, it's indeed just a `pacman -S`.
Looking at their code, strictly speaking, it's doing a `pacman --noconfirm -S --needed --overwrite '*' git rust`.
It also has some caching being done so that when testing multiple environment (MSYS2, MINGW32, MINGW64, ...) it won't re-download the common packages, which we'll probably never need.

.github/workflows/test.yml
Nahor: >We could even add a smoke test 

You mean as a temporary replacement for calling `cargo test`, as an extra "name / run"?

tests/checks/tmux-fish_config.fish
Nahor: My issue was adding `curl` (or `wget`) as a new requirement for testing since I didn't see any other test doing http request.

tests/checks/tmux-fish_config.fish
Nahor: Can do.
~~Alternative: redirect `fish_config` output to a file and use `cat` or `tail` instead of `tmux capture-pane` and avoid the width limit entirely.~~ Doesn't work because of buffering inside the redirect

krobelus:
> You mean as a temporary replacement for calling `cargo test`, as an extra "name / run"?

yeah. Just to test it can start and execute commands 

> 
> > @@ -0,0 +1,26 @@
> +#RUN: %fish %s
> +
> +#REQUIRES: command -v tmux
> +isolated-tmux-start
> 
> My issue was adding `curl` (or `wget`) as a new requirement for testing since I didn't see any other test doing http request.

Right. Maybe use `python -c 'import requests; ...'` ?
I think we can already assume that `python` exists in `PATH`.

> +isolated-tmux capture-pane -p | sed \
> +    -e '1h;2,$H;$!d;g' \
> +    -e 's|file://.*\.html|<file_url>|' \
> +    -e 's|http://.*/\n|<http_url>\n|'
> 
> Can do.
> Alternative: redirect `fish_config` output to a file and use `cat` or `tail` instead of `tmux capture-pane` and avoid the width limit entirely.

Good idea -- write to a file in the current directory,
and then use `cat log` outside tmux.

Nahor:
> Maybe use `python -c 'import requests; ...'` ?

`requests` is not standard Python so it won't be on all platform by default (at least it's not with MSYS2 python). `urllib` could be used but it feels too complicated for one-liners.
I believe `wget` should be safe to use.

> > Alternative: redirect `fish_config` output to a file and use `cat` or `tail` instead of `tmux capture-pane` and avoid the width limit entirely.
> Good idea -- write to a file in the current directory,

I tried and it doesn't work. The problem is that the file is buffered so there is no data until we exit `fish_config`. I thought about running all the commands and then checking the output, but if we want to run some basic HTTP requests, we need the output to get the authentication string.
I tried to use `tee` to work around the buffering but for some reason, it's not capturing the output (terminal and file remain blank until `fish_config` closes at which point all the output gets printed in the terminal and none in the file)
I also thought of `stdbuf` but that expects an actual command, not a shell function.

.github/workflows/main.yml
```diff
@@ -174,3 +174,30 @@ jobs:
...
+    steps:
+    - uses: actions/checkout@v4
+    - uses: msys2/setup-msys2@v2
+      with:
+        update: true
+        msystem: MSYS
+    - name: install needed
+      # Not using setup-msys2 `install` option to make it easier to copy/paste
+      run: |
+        pacman --noconfirm -S --needed --overwrite '*' git rust
```
krobelus: `--overwrite '*'` is sus; it seems like a workaround for weird packaging bugs;
I hope we don't need this, so better remove it and add it only if we really need to

share/tools/web_config/webconfig.py
```diff
@@ -1544,7 +1544,10 @@ def log_error(self, format, *args):
 
 
 def runThing():
-    if isMacOS10_12_5_OrLater():
+    if os.environ.get("BROWSER", "") == "true":
```
krobelus: yep that looks reasonable;
for as long as we have all these weird workarounds,
we want a reliable way to have system tests disable this.

BTW, we should probably leave out the `, ""` default argument,
it doesn't add any useful behavior, and it's already obvious
from our use of `.get()` that a missing variable is not an error.

(As mentioned in #11926 this workaround for macOS 10.12.5 or later,
has been fixed in macOS 10.12.6 according to https://andrewjaffe.net/blog/2017/05/python_bug_hunt/,
I think we can assume all users have upgraded to that patch version
So let's remove this workaround, but that can be in a different PR).

tests/checks/tmux-fish_config.fish
```diff
@@ -0,0 +1,43 @@
...
+# because of buffering, the file will remain empty until the buffer is full or
+# fish_config exits.
+isolated-tmux resize-window -x 500
+
+isolated-tmux send-keys "BROWSER=true fish_config" Enter
+tmux-sleep
+isolated-tmux capture-pane -p 
+# CHECK: prompt 0> BROWSER=true fish_config
+# CHECK: Web config started at file://{{.*}}.html
+# CHECK: If that doesn't work, try opening http://localhost:8000/{{\w+}}/
```
krobelus: This fails for me due to a port number mismatch:

	If that doesn't work, try opening http://localhost:8001/9b54d4275031129e9fa88e340623e5f9/

Probably we don't care about the exact port number?

FWIW, I ran the test twice, in two different ways:

	# passes
	ninja -Cbuild test_tmux-fish_config.fish
	# fails
	tests/test_driver.py target/debug tests/checks/tmux-fish_config.fish

tests/checks/tmux-fish_config.fish
```diff
@@ -0,0 +1,43 @@
...
+# CHECK: Hit ENTER to stop.
+
+# Extract the auth_string from the output
+set -l base_url (isolated-tmux capture-pane -p | string match -r 'http://localhost:8000/\w+/$')
+
+# Check a bad URL
+wget -q -O /dev/null "http://localhost:8000/invalid_auth/" 2>/dev/null
+or test $status -eq 8
```
krobelus: as written, this test statement is useless.
If the test statement fails, nothing happens.
Should follow it up with `or echo "unexpected exit code from wget"`.
That's probably more reliable than testing the exact wget output below.

tests/checks/tmux-fish_config.fish
```diff
@@ -0,0 +1,43 @@
...
+
+isolated-tmux capture-pane -p -S 4
+# CHECK: {{.*}} code 403, message Forbidden, path /invalid_auth/
+
+# Check a good URL
+set -l workspace_root (path resolve -- (status dirname)/../../)
+diff -q "$workspace_root/share/tools/web_config/index.html" (wget -q -O - $base_url 2>/dev/null | psub -f)
```
krobelus, line 52: nice

krobelus:
> I also thought of stdbuf but that expects an actual command, not a shell function.

yeah I could reproduce the same buffering behavior in `webconfig.py`.
We could change that but I don't see a problem with the current approach.
Just needs small test robustness fixes, then we're good

tests/checks/tmux-fish_config.fish
krobelus: ok. I'd normally prefer `curl` since it's more widely used but `wget` is fine too of course.
Can you maybe install `wget` explicitly in the main CI job (the first `ubuntu` one in `.github/workflows/main.yml`).

(Could also add it to other platforms but before we do that we should probably extract functions for installing all test dependencies, so it doesn't get any messier. For example `./.github/actions/install-sphinx-markdown-builder` is one such function.)

Nahor: I went with `wget` because it's pretty much pre-installed on any system. Like the equivalent of `nano` vs `emacs`/`vi`. Even my router, whose firmware has to be as small as possible, has it (OpenWRT). And it's ubiquitous enough to be a tool integrated into Busybox.

If you really want, I can add it to the Ubuntu build, it's simple enough. But it feels unnecessary. It passes all the tests, which means `wget` is already installed. For that matter `wget` is a dependency of the `ubuntu-standard` meta package so if it's not there, we have bigger problems.

tests/checks/tmux-fish_config.fish
Nahor: I hardcoded the number because that way I could just hardcode the bad URL. But it's easy enough to extract it from `base_url`.

As-is it certainly will fail if you have another instance of `fish_config` already running (or, of course, if any other server app is on that port)

share/tools/web_config/webconfig.py
Nahor: >BTW, we should probably leave out the , "" default argument,
> it doesn't add any useful behavior, and it's already obvious

I thought `get` would throw if the variable was not present, but I must have been confused with `[...]` or the `get` of some other language. Will remove.

Nahor:
Arf! Looks like Busybox's `wget` has different status codes than GNU's.  I tried to add `wget` to the alpine docker image but it looks like it's a separate process to build those imagea. So for now, I'm doing a quick&dirty check of which version of wget we have and change the status code accordingly.
I believe it's a corner case though (alpine being a very minimal distribution), and that we can assume GNU `wget` is pre-installed otherwise.

Issue #7944, opened 2021-04-20
Smart-case tab completion has become less smart


zx8:
Using https://github.com/fish-shell/fish-shell/commit/9c413b039d572963dd5cf9a673ec47927d584df2.

---

It's unclear if this is intentional, but it seems like a regression in smart-case tab completion:

### Repro

```
$ mkdir dog Duck Dodo
$ cd do<tab>
```

### Expected

Both `dog` and `Dodo` should be offered`

### Actual

`dog` is automatically completed - `Dodo` is not offered.

### Rationale

When using lowercase and tab-completing, it's because I can't remember which directories are lowercase and which are uppercase. When I use uppercase (i.e. `cd D<tab>`), it's because I _know_ a particular directory starts with an uppercase letter and going out of my way to enter an uppercase character.

This would be consistent with how smart-case works with all other CLI tools I use, so I expect this is a regression (since it didn't use to work this way in fish either).

zanchey:
I don't know about a regression - this is how 3.0.2 and 3.1.2 behave as well. `complete -C` shows the right result, but something else removes `Dodo` as a valid completion when using <kbd>Tab</kbd>.

zanchey:
See also https://github.com/fish-shell/fish-shell/issues/3978

krobelus:
In the interactive reader [only the best completions are shown](https://github.com/fish-shell/fish-shell/blob/2fff12104d7ea37323beb266a030ff89610d1ee2/src/reader.cpp#L1947-L1948). The rankings are described in https://github.com/fish-shell/fish-shell/issues/568#issuecomment-18455181.
The match type (exact/prefix/substring/subsequence) [dominates the case differences](https://github.com/fish-shell/fish-shell/blob/master/src/wcstringutil.cpp#L212).

krackers:
Note that even if the above is fixed (either by ranking smart-case and case-sensitive the same or by changing the reader to permit ranks that differ by case only), there'd still be an issue where the reader rejects smartcase completions because they require replacement

>If any commands of the best rank do not require replacement, then ignore all those that want to use replacement.


@krobelus Do you know why there exists a distinction between completions that replace the token and ones that don't? From a user perspective why not just treat all of them as replacing token (which would be a visual no-op if the prefix already matches?) Is this just a performance optimization?

krobelus:
> Do you know why there exists a distinction between completions that replace the token and ones that don't?

[...]

PR #11910, opened 2025-10-07
try fix cd smart case

asukaminato0721:
## Description

Talk about your changes here.

Fixes issue #7944

## TODOs:

### DIFF
--- a/tests/checks/tmux-complete.fish
+++ b/tests/checks/tmux-complete.fish
@@ -23,14 +23,14 @@ tmux-sleep
 isolated-tmux capture-pane -p
 # CHECK: prompt 1> cat cmake/
 
-# Keep mixed-case completions visible when typing lowercase (#7944).
-isolated-tmux send-keys C-u C-l 'rm -rf dog Dodo' Enter \
-    'mkdir dog Dodo' Enter C-l \
+# Keep mixed-case completions visible when typing lowercase and ignore non-matching prefixes (#7944).
+isolated-tmux send-keys C-u C-l 'rm -rf dog Dodo Voodo' Enter \
+    'mkdir dog Dodo Voodo docker doc_internal doc_src' Enter C-l \
     'cd do' Tab
 tmux-sleep
 isolated-tmux capture-pane -p
 # CHECK: prompt {{\d+}}> cd do
-# CHECK: dog/{{ +}}Dodo/
+# CHECK: docker/  doc_internal/  doc_src/  dog/  Dodo/
 
 # Correct case in pager when prefixes differ in case (#7743).
 isolated-tmux send-keys C-u C-l 'complete -c foo2 -a "aabc aaBd" -f' Enter C-l \


asukaminato0721:
at least the new test passes..
```
checks/tmux-smartcase.fish                         PASSED    4344 ms
```

src/reader.rs
```diff
@@ -6576,7 +6576,15 @@ impl<'a> Reader<'a> {
...
             if completion_replaces_token != will_replace_token {
-                continue;
+                // Keep smart/insensitive-case completions even when a same-case completion exists.
+                // They need to replace the token to adjust casing, but should still be offered (#7944).
+                let is_case_insensitive_match = matches!(
+                    c.r#match.case_fold,
+                    CaseSensitivity::Smart | CaseSensitivity::Insensitive
+                );
+                if !(completion_replaces_token && is_case_insensitive_match) {
+                    continue;
```
krobelus, line 6660: so the idea is that we have
case 1: will_replace_token && !completion_replaces_token

in this case we always discard the completion (existing behavior, no change)

case 2: !will_replace_token && completion_replaces_token

in this case, this change makes us no longer discard completions **if**
if replacement does not match the case exactly,
with the implicit assumption that "completion_replaces_token && case fold is not exact"
means that this is basically and appending completion if it were not for the case
mismatch. This assumption is correct for do/Dodo, but it's **not** correct for
do/VooDo.

I don't have a lot of context but the history of the code should tell us
explicitly that we don't want to mix (lower quality) substring completions
with (higher quality) prefix completions, but only show the highest quality available.

The issue we're trying to address here is legit and this direction looks reasonable,
but we should only keep the completion if `string_prefixes_string_case_insensitive` is true here.

Could you also add a test case where this is not true (like the one I gave above),
to make sure we don't mix those completions.

tests/checks/fish_config.fish
```diff
@@ -19,7 +19,7 @@ fish_config theme show "fish default"
 # CHECK: {{\x1b\[m}}echo{{\x1b\[m}} 'Errors are the portal to discovery
 # CHECK: {{\x1b\[m}}Th{{\x1b\[m}}is an autosuggestion
 # CHECK: {{\x1b\[m}}{{\x1b\[4m}}fish default{{\x1b\[m}}
-# CHECK: {{\x1b\[m}}/bright/vixens{{\x1b\[m}} {{\x1b\[36m}}jump{{\x1b\[m}} {{\x1b\[36m}}{{\x1b\[1m}}|{{\x1b\[m}} {{\x1b\[33m}}"fowl"{{\x1b\[m}} {{\x1b\[36m}}{{\x1b\[1m}}> quack{{\x1b\[m}} {{\x1b\[32m}}&{{\x1b\[m}}{{\x1b\[31m}} # This is a comment
+# CHECK: {{(?:\x1b\[m)?}}/bright/vixens{{\x1b\[m}} {{\x1b\[36m}}jump{{\x1b\[m}} {{\x1b\[36m}}{{\x1b\[1m}}|{{\x1b\[m}} {{\x1b\[33m}}"fowl"{{\x1b\[m}} {{\x1b\[36m}}{{\x1b\[1m}}> quack{{\x1b\[m}} {{\x1b\[32m}}&{{\x1b\[m}}{{\x1b\[31m}} # This is a comment
```
krobelus: yeah this test is flaky in CI but I don't think this is enough
to fix it, see https://cirrus-ci.com/task/4724192307838976?logs=tests#L280-L301
Until that's fixed, maybe we should filter output with something like `SANITIZE_FOR_PRINTING_RE`.
I'd prefer if we try to deal with this test in a separate PR, since it's not so easy.

tests/checks/tmux-smartcase.fish
```diff
@@ -0,0 +1,15 @@
+#RUN: %fish %s
+#REQUIRES: command -v tmux
+
+isolated-tmux-start
```
krobelus: please squash all the related commits into one (test + implementation).

tests/checks/tmux-smartcase.fish
```diff
@@ -0,0 +1,15 @@
...
+
```
krobelus: this is hard to review without reading the issue
or perusing the history of the code.
Ostensibly this issue was caused by a regression,
can you include a reference (i.e. `git log --format=reference -1`)
to the relevant commits that broke things?

please add infos to the commit message, the PR description is not a good source of truth

tests/checks/tmux-smartcase.fish
```diff
@@ -0,0 +1,15 @@
...
+# Smartcase completions keep insensitive matches available (#7944).
+isolated-tmux send-keys C-u C-l 'rm -rf sc-smartcase' Enter \
```
krobelus: drop the "rm" command, we don't need it, since the test driver always makes a fresh dir

tests/checks/tmux-smartcase.fish
```diff
@@ -0,0 +1,15 @@
...
+    'mkdir sc-smartcase' Enter \
+    'cd sc-smartcase' Enter \
```
krobelus: after rebasing on `master` these two can be done simpler like

	isolated-tmux-start -C "mkdir sc-smartcase; cd smartcase"

but actually we might not need them at all since `$PWD` should be mostly empty

tests/checks/tmux-smartcase.fish
```diff
@@ -0,0 +1,15 @@
...
+    'mkdir dog Dodo' Enter C-l \
+    'cd do' Tab
+tmux-sleep
+isolated-tmux capture-pane -p
+# CHECK: prompt {{\d+}}> cd dog/
+# CHECK: doDodo/  dog/
```
krobelus: is this correct though? doDodo seems wrong,
I can't think of any autocompletion system that behaves this way.
Shouldn't we render <u>Do</u>do
or perhaps <s>d</s><u>Do</u>do

I think this is a blocker;
we should teach the logic around `Determine the shared prefix length.`
where we compute `common_prefix`
to also consider this case as a prefix..
and double-check if there are any other invariants
broken now that the common prefix may not match all cases.

tests/checks/tmux-smartcase.fish
asukaminato0721: > to the relevant commits that broke things

It seems too old(years ago), so I can't find the relevant one.

tests/checks/tmux-smartcase.fish
asukaminato0721: image

Now it looks like this.

src/reader.rs
krobelus: since it took some time to find this issue, we should add this test and ideally address the other review comments too.
Don't make the reviewer do most of the work.

asukaminato0721:
Allow completions that only differ by casing to coexist with same-case matches by detecting those that are prefix/exact matches needing a casing fix and letting them survive the non-replacement filter, so both dog/ and Dodo/ show up instead of discarding the uppercase entry

Added a tmux-based regression to capture the original repro sequence and assert that a lowercase cd do<Tab> keeps both mixed-case directories in the pager output

image


---

```
> mkdir Voodo do Dodo
> cd do/
do/  Dodo/  
```

src/reader/reader.rs
```diff
@@ -6646,21 +6653,32 @@ impl<'a> Reader<'a> {
...
 
-            // This completion survived.
-            surviving_completions.push(c.clone());
-            all_matches_exact_or_prefix =
-                all_matches_exact_or_prefix && c.r#match.is_exact_or_prefix();
+            all_matches_exact_or_prefix &= c.r#match.is_exact_or_prefix();
+
+            let mut completion = c.clone();
+            if needs_case_fix && !will_replace_token {
+                completion.flags |= CompleteFlags::SUPPRESS_PAGER_PREFIX;
```
krobelus, line 6673: there was a rendering glitch - if I complete "cd do",
then the completion candidate for "Dodo" in the completion pager
will have two trailing spaces highlihted with `fish_pager_color_selected_background`.

Fixed by updating computation of `comp_width` to honor `SUPPRESS_PAGER_PREFIX`.

src/reader/reader.rs
```diff
@@ -6646,21 +6653,32 @@ impl<'a> Reader<'a> {
...
+                deferred_case_fixes.push(completion);
+            } else {
+                surviving_completions.push(completion);
+            }
         }
 
+        surviving_completions.extend(deferred_case_fixes);
```
krobelus, line 6680: I'm not sure the reordering change is desirable. AFAICT, we don't reorder in other cases.
(for example if I do "cd DO" tab, completions will be ordered alphabetically, irrespective of whether the matches are same case or exact.)

I might be wrong but for now I'll remove the ordering change. We can maybe add it back later.

Nice work!
It took me some time to review because this is tricky code,
but the tests were helpful.

src/reader/reader.rs
```diff
@@ -6735,6 +6753,11 @@ impl<'a> Reader<'a> {
 
         if use_prefix {
             for c in &mut surviving_completions {
+                if completion_needs_case_fix(c) {
+                    // Keep replacement semantics and the original prefix so these completions can
+                    // fix casing when selected.
+                    continue;
+                }
```
krobelus, line 6760: this broke another test case in the same file

	complete -c foo2 -a "aabc aaBd" -f
	foo2 A

and then press Tab.
This would add a small glitch -- an extra "aa" prefix in the completion pager.
Fortunately it was easy to fix: we shouldn't skip these lines if `will_replace_token` is set.

Issue #11837, opened 2025-09-29
Support `\e]133;B` from FinalTerm escape codes


acarl005:
Fish 4.1 added support for various [FinalTerm escape codes](https://gitlab.freedesktop.org/Per_Bothner/specifications/blob/master/proposals/semantic-prompts.md#commands) for [improved terminal integration](https://fishshell.com/docs/4.1/terminal-compatibility.html#:~:text=click_events%3D1%5Cx07-,Mark%20prompt%20start,-(OSC%20133)%2C%20with). This is awesome!

It would be great to add `\e]133;B` as well, which goes between the end of the prompt and the start of the cmdline. In particular, it would help Warp Terminal with our shell integration features.

krobelus:
sounds reasonable; I'm curious, what features would a terminal use this for, anything that we can already try today?

acarl005:
One example is: we have a right-click menu with items for copying the cmdline or for copying the output of a command to clipboard.

We already support this with hacky shell integration code, but if fish shell emitted the codes natively it would be more robust and also allow us to reduce the amount of shell integration code and speed up session startup.

krobelus:
> One example is: we have a right-click menu with items for copying
> the cmdline or for copying the output of a command to clipboard.

does it behave differently when clicking on prompt versus clicking on the command line?

acarl005:
Not exactly. How it works is that you can right-click anywhere within these borders which we draw between each command. That brings up a context menu with these 2 items, "copy command" and "copy output". Having a reliable `\e]133;B` provides the boundary that helps us distinguish between those two.

<img width="1303" height="907" alt="Image" src="https://github.com/user-attachments/assets/470a9b93-a5aa-48e8-a124-15cd83e87b0d" />

krobelus:
So you want to provide only one action, that copies either the command or command output, based on where the user clicked?

I'm not sure I understand the example; we already emit Prompt Start (A), Command Start (C) and Command End (D).
Is this about commands that haven't been exectued yet (i.e. they don't have a `C` yet).. though I don't understand why it would be necessary since there is nothing below.
My point is, Command Start (C) already provides the exact boundary between prompt+commandline, and command output.
AIUI you don't care about distinguishing between prompt and commandline, so why the extra boundary for those?

acarl005:
Oops I misspoke. You're correct that `\e]133;C` already helps us delineate between "copy command" and "copy output".

[...]

PR #11911, opened 2025-10-07
Make multiline prompt part of ScreenData

### LINKED ISSUE #11875: Lines not cleared with multi-line initial prompt and single-line transient prompt

kerty0:
## Description

Sorry for my sudden disappearance. I kept increasing the scope of my PR and eventually burned out.

Like #11901, this fixes #11875. It also makes fish hide the part of a multiline prompt that leaves the viewport and removes an old hack that was previously necessary for locating the first prompt line.

I have tried to avoid introducing bugs this time, but I would be grateful for any help with testing.

- [ ] Add tests for edge cases

### DIFF
--- a/src/screen.rs
+++ b/src/screen.rs
@@ -157,6 +157,9 @@ pub struct ScreenData {
     screen_width: Option<usize>,
 
     cursor: Cursor,
+
+    prompt_lines: usize,
+    scroll_amount: usize,
 }
 
 impl ScreenData {
@@ -207,8 +210,6 @@ impl ScreenData {
 pub struct Screen {
     /// Whether the last-drawn autosuggestion (if any) is truncated, or hidden entirely.
     pub autosuggestion_is_truncated: bool,
-    /// True if the last rendering was so large we could only display part of the command line.
-    pub scrolled: bool,
 
     /// Receiver for our output.
     outp: &'static RefCell<Outputter>,
@@ -244,7 +245,6 @@ impl Screen {
         Self {
             outp: Outputter::stdoutput(),
             autosuggestion_is_truncated: Default::default(),
-            scrolled: Default::default(),
             desired: Default::default(),
             actual: Default::default(),
             actual_left_prompt: Default::default(),
@@ -259,7 +259,7 @@ impl Screen {
 
     /// Return whether the last rendering was so large we could only display part of the command line.
     pub fn scrolled(&self) -> bool {
-        self.scrolled
+        self.desired.scroll_amount != 0
     }
 
     /// This is the main function for the screen output library. It is used to define the desired
@@ -351,10 +351,12 @@ impl Screen {
             !autosuggestion.is_empty() && autosuggestion != layout.autosuggestion;
 
         // Clear the desired screen and set its width.
-        self.desired.screen_width = Some(screen_width);
         self.desired.resize(0);
+        self.desired.screen_width = Some(screen_width);
+        self.desired.prompt_lines = calc_prompt_lines(left_prompt) - 1;
         self.desired.cursor.x = 0;
-        self.desired.cursor.y = 0;
+        self.desired.cursor.y = self.desired.prompt_lines;
+        self.desired.resize(self.desired.prompt_lines);
 
         // Append spaces for the left prompt.
         let prompt_offset = if pager.search_field_shown {
@@ -457,7 +459,7 @@ impl Screen {
             self.desired.add_line();
         }
 
-        let full_line_count = self.desired.cursor.y
+        let full_line_count = self.desired.cursor.y + 1
             - if self.desired.cursor.x == 0
                 && self
                     .desired
@@ -469,8 +471,7 @@ impl Screen {
                 1
             } else {
                 0
-            }
-            + calc_prompt_lines(&layout.left_prompt);
+            };
         let pager_available_height = std::cmp::max(
             1,
             curr_termsize
@@ -514,30 +515,16 @@ impl Screen {
         pager.update_rendering(page_rendering);
         // Append pager_data (none if empty).
         self.desired.append_lines(&page_rendering.screen_data);
-
-        self.scrolled = scrolled_cursor.scroll_amount != 0;
+        self.desired.scroll_amount = if is_final_rendering {
+            0
+        } else {
+            scrolled_cursor.scroll_amount
+        };
 
         self.with_buffered_output(|zelf| {
-            zelf.update(
-                vars,
-                &layout.left_prompt,
-                &layout.right_prompt,
-                is_final_rendering,
-            )
+            zelf.update(vars, &layout.left_prompt, &layout.right_prompt)
         });
         self.save_status();
-    }
-
-    pub fn multiline_prompt_hack(&mut self) {
-        // If the prompt is multi-line, we need to move up to the prompt's initial line. We do this
-        // by lying to ourselves and claiming that we're really below what we consider "line 0"
-        // (which is the last line of the prompt). This will cause us to move up to try to get back
-        // to line 0, but really we're getting back to the initial line of the prompt.
-        let prompt_line_count = self
-            .actual_left_prompt
-            .as_ref()
-            .map_or(1, |p| calc_prompt_lines(p));
-        self.actual.cursor.y += prompt_line_count.checked_sub(1).unwrap();
     }
 
     /// Resets the screen buffer's internal knowledge about the contents of the screen,
@@ -551,7 +538,6 @@ impl Screen {
             std::cmp::max(self.actual_lines_before_reset, self.actual.line_count());
 
         if repaint_prompt {
-            self.multiline_prompt_hack();
             self.actual_left_prompt = None;
             self.need_clear_screen = true;
         }
@@ -566,24 +552,7 @@ impl Screen {
     }
 
     pub fn push_to_scrollback(&mut self, cursor_y: usize) {
-        let prompt_y = self.command_line_y_given_cursor_y(cursor_y);
-        let trailing_prompt_lines = self
-            .actual_left_prompt
-            .as_ref()
-            .map_or(0, |p| calc_prompt_lines(p) - 1);
-        let lines_to_scroll = prompt_y
-            .checked_sub(trailing_prompt_lines)
-            .unwrap_or_else(|| {
-                FLOG!(
-                    reader,
-                    "Number of trailing prompt lines prompt lines",
-                    trailing_prompt_lines,
-                    "exceeds prompt's y",
-                    prompt_y,
-                    "inferred from reported cursor position",
-                );
-                0
-            });
+        let lines_to_scroll = self.command_line_y_given_cursor_y(cursor_y);
         if lines_to_scroll == 0 {
             return;
         }
@@ -629,26 +598,8 @@ impl Screen {
                 0
             });
         let y = y.min(self.actual.line_count() - 1);
-        let Some(viewport_prompt_x) = viewport_cursor.x.checked_sub(self.actual.cursor.x) else {
-            FLOGF!(
-                reader,
-                "Actual cursor x=%d exceeds reported cursor x=%d, \
-                     was the cursor moved by printing to the TTY?",
-                self.actual.cursor.x,
-                viewport_cursor.x
-            );
-            return CharOffset::None;
-        };
-        let Some(x) = viewport_position.x.checked_sub(viewport_prompt_x) else {
-            FLOGF!(
-                reader,
-                "Computed prompt x=%d exceeds mouse click x=%d, \
-                 was the cursor moved by printing to the TTY?",
-                viewport_prompt_x,
-                viewport_position.x
-            );
-            return CharOffset::None;
-        };
+        let viewport_prompt_x = viewport_cursor.x - self.actual.cursor.x;
+        let x = viewport_position.x - viewport_prompt_x;
         let line = self.actual.line(y);
         let x = x.max(line.indentation);
         let offset = line
@@ -1046,13 +997,7 @@ impl Screen {
     }
 
     /// Update the screen to match the desired output.
-    fn update(
-        &mut self,
-        vars: &dyn Environment,
-        left_prompt: &wstr,
-        right_prompt: &wstr,
-        is_final_rendering: bool,
-    ) {
+    fn update(&mut self, vars: &dyn Environment, left_prompt: &wstr, right_prompt: &wstr) {
         // Helper function to set a resolved color, using the caching resolver.
         let mut color_resolver = HighlightColorResolver::new();
         let mut set_color = |zelf: &mut Self, c| {
@@ -1103,47 +1048,65 @@ impl Screen {
             need_clear_screen = true;
         }
 
-        // Output the left prompt if it has changed.
-        if self.scrolled() && !is_final_rendering {
-            self.r#move(0, 0);
-            self.write_command(ClearToEndOfLine);
-            self.actual_left_prompt = None;
-            self.actual.cursor.x = 0;
-        } else if self
+        let cmd_start = self
+            .desired
+            .prompt_lines
+            .saturating_sub(self.desired.scroll_amount);
+        let first_visible = self.desired.prompt_lines >= self.desired.scroll_amount;
+
+        let prompt_changed = self
             .actual_left_prompt
             .as_ref()
-            .is_none_or(|p| p != left_prompt)
-            || (self.scrolled() && is_final_rendering)
-            || Some(left_prompt_width) == screen_width && self.should_wrap(0)
-        {
+            .is_none_or(|p| first_visible && p != left_prompt)
+            || first_visible && self.actual.scroll_amount != self.desired.scroll_amount;
+        let prompt_softwrap =
+            first_visible && Some(left_prompt_width) == screen_width && self.should_wrap(cmd_start);
+
+        // Output the left prompt if it has changed or we need to refresh softwrap.
+        if prompt_changed || prompt_softwrap {
             self.r#move(0, 0);
-            let mut start = 0;
-            let mark_prompt_start = |zelf: &mut Screen| zelf.write_command(Osc133PromptStart);
-            if left_prompt_layout.line_breaks.is_empty() {
-                mark_prompt_start(self);
+            let mut start;
+            if left_prompt_layout.line_breaks.len() <= self.desired.scroll_amount {
+                self.write_command(Osc133PromptStart);
             }
-            if self
-                .actual_left_prompt
-                .as_ref()
-                .is_none_or(|p| p != left_prompt)
-                || (self.scrolled() && is_final_rendering)
-            {
-                for (i, &line_break) in left_prompt_layout.line_breaks.iter().enumerate() {
+            if prompt_changed {
+                start = if self.desired.scroll_amount == 0 {
+                    0
+                } else if self.desired.prompt_lines > self.desired.scroll_amount {
+                    left_prompt_layout.line_breaks[self.desired.scroll_amount - 1] + 1
+                } else if self.desired.prompt_lines > 0 {
+                    left_prompt_layout.line_breaks[self.desired.prompt_lines - 1] + 1
+                } else {
+                    0
+                };
+                for (i, &line_break) in left_prompt_layout
+                    .line_breaks
+                    .iter()
+                    .skip(self.desired.scroll_amount)
+                    .enumerate()
+                {
                     self.write_command(ClearToEndOfLine);
                     if i == 0 {
-                        mark_prompt_start(self);
+                        self.write_command(Osc133PromptStart);
                     }
                     self.write_str(&left_prompt[start..=line_break]);
                     start = line_break + 1;
                 }
             } else {
-                start = left_prompt_layout.line_breaks.last().map_or(0, |lb| lb + 1);
+                self.r#move(0, cmd_start);
+                start = left_prompt_layout.line_breaks[self.desired.prompt_lines - 1] + 1;
             }
-            self.write_str(&left_prompt[start..]);
+            if first_visible {
+                self.write_str(&left_prompt[start..]);
+                self.actual.cursor.x = left_prompt_width;
+            }
             self.actual_left_prompt = Some(left_prompt.to_owned());
-            self.actual.cursor.x = left_prompt_width;
-            if Some(left_prompt_width) == screen_width && self.should_wrap(0) {
-                self.soft_wrap_location = Some(Cursor { x: 0, y: 1 });
+            self.actual.cursor.y = cmd_start;
+            if prompt_softwrap {
+                self.soft_wrap_location = Some(Cursor {
+                    x: 0,
+                    y: cmd_start + 1,
+                });
             }
         }
 
@@ -1155,9 +1118,9 @@ impl Screen {
         }
 
         // Output all lines.
-        for i in 0..self.desired.line_count() {
+        for i in cmd_start..self.desired.line_count() {
             self.actual.create_line(i);
-            let is_first_line = i == 0 && !self.scrolled();
+            let is_first_line: bool = first_visible && i == cmd_start;
             let start_pos = if is_first_line { left_prompt_width } else { 0 };
             let mut current_width = 0;
             let mut has_cleared_line = false;
@@ -1173,14 +1136,25 @@ impl Screen {
             // Note that skip_remaining is a width, not a character count.
             let mut skip_remaining = start_pos;
 
-            let shared_prefix = if self.scrolled() {
+            let previously_prompt_line = self.actual.prompt_lines >= self.actual.scroll_amount + i;
+            if previously_prompt_line && is_first_line && Some(left_prompt_width) != screen_width {
+                // Clean up first line from previous prompt
+                self.r#move(left_prompt_width, i);
+                self.write_command(ClearToEndOfLine);
+            };
+
+            let shared_prefix = if self.desired.scroll_amount == 0 || previously_prompt_line {
                 0
             } else {
                 line_shared_prefix(o_line(self, i), s_line(self, i))
             };
             let mut skip_prefix = shared_prefix;
-            if shared_prefix < o_line(self, i).indentation {
-                if o_line(self, i).indentation > s_line(self, i).indentation && !has_cleared_screen
+            if shared_prefix < o_line(self, i).indentation
+                || previously_prompt_line && !is_first_line
+            {
+                if !has_cleared_screen
+                    && (o_line(self, i).indentation > s_line(self, i).indentation
+                        || previously_prompt_line && !is_first_line)
                 {
                     set_color(self, HighlightSpec::new());
                     self.r#move(0, i);
@@ -1189,6 +1163,10 @@ impl Screen {
                     } else {
                         ClearToEndOfLine
                     });
+                    if i == 0 {
+                        // Restore prompt start if we deleted it
+                        self.write_command(Osc133PromptStart);
+                    }
                     has_cleared_screen = should_clear_screen_this_line;
                     has_cleared_line = true;
                 }
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -4735,10 +4735,6 @@ impl<'a> Reader<'a> {
 
                 self.left_prompt_buff =
                     join_strings(&self.exec_prompt_cmd(prompt_cmd, final_prompt), '\n');
-
-                if final_prompt {
-                    self.screen.multiline_prompt_hack();
-                }
             }
 
             // Don't execute the right prompt if it is undefined fish_right_prompt


src/screen.rs
```diff
@@ -157,6 +157,9 @@ pub struct ScreenData {
     screen_width: Option<usize>,
 
     cursor: Cursor,
+
+    prompt_lines: usize,
+    scroll_amount: usize,
```
krobelus: I wonder if ScreenData actually needs the exact scroll amount 
and the actual number of prompt lines even if they're not visible;
maybe it's enough to include only `visible_prompt_lines: usize,`.

I haven't tried that though.

src/screen.rs
```diff
@@ -351,10 +351,12 @@ impl Screen {
             !autosuggestion.is_empty() && autosuggestion != layout.autosuggestion;
 
         // Clear the desired screen and set its width.
-        self.desired.screen_width = Some(screen_width);
         self.desired.resize(0);
+        self.desired.screen_width = Some(screen_width);
+        self.desired.prompt_lines = calc_prompt_lines(left_prompt) - 1;
```
krobelus: `prompt_lines` is a confusing name because it actually counts the number of
newline characters in the prompt, which is one less than the number of lines.
Call it `prompt_line_breaks?
I'm a bit surprised about the choice though,
wouldn't it be better if we made it the actual number of prompt lines
(where the last line has no trailing newline char, by definition).
Then we'd write

	let first_visible = self.desired.prompt_lines > self.desired.scroll_amount;
	// or: let prompt_visible = ...

which should be easier to follow.

src/screen.rs
```diff
@@ -629,26 +598,8 @@ impl Screen {
...
-            FLOGF!(
-                reader,
-                "Computed prompt x=%d exceeds mouse click x=%d, \
-                 was the cursor moved by printing to the TTY?",
-                viewport_prompt_x,
-                viewport_position.x
-            );
-            return CharOffset::None;
-        };
+        let viewport_prompt_x = viewport_cursor.x - self.actual.cursor.x;
```
krobelus: this subtraction check should stay (dd47c2baa28 (Sanitize cursor position report on kitty click_events, 2025-10-07) which I can reproduce on kitty).
This is only used for kitty click_events today.
We might
1. call handle_char_event() for a mouse click event
2. this makes us ask the terminal for the cursor position,
   and we stop processing any input events until we get the response.
3. when we receive the response, we code to the code here.

Notably, we can't trust the original mouse click nor the cursor
position:
- neither are guaranteed to be in bounds (the terminal might have
  been resized since, and SIGWINCH delivery is async, so there are
  no guarantees)
- the cursor might be somewhere completely different
  from where fish left it, due to other background output moving it.
- unlikely but I guess the terminal could have a bug

So for now we should log a `reader` error and do nothing if we get
obviously bad input.

src/screen.rs
```diff
@@ -629,26 +598,8 @@ impl Screen {
...
+        let x = viewport_position.x - viewport_prompt_x;
```
krobelus: I think this can also crash when
1. viewport_cursor.x is very high (e.g. due to something like `command printf %0100d &`)
2. self.actual.cursor.x is low (e.g. 2, on a prompt like `$ `)
3. due to 1 and 2, `viewport_prompt_x` is 100
4. `viewport_position.x` is lower than that (which is the mouse click
    position; maybe we should change up some of the variable names,
    since this is only used by mouse click handling).
5. then `viewport_position.x - viewport_prompt_x` overflows

Though I haven't actually reproduced this one.

src/screen.rs
```diff
@@ -1103,47 +1048,65 @@ impl Screen {
...
         }
 
-        // Output the left prompt if it has changed.
-        if self.scrolled() && !is_final_rendering {
-            self.r#move(0, 0);
-            self.write_command(ClearToEndOfLine);
-            self.actual_left_prompt = None;
-            self.actual.cursor.x = 0;
-        } else if self
+        let cmd_start = self
```
krobelus: ok so this is the 0-based screen line index, indicating where the command line starts.
I think "cmd" might be a bit ambiguous because it could be command output, maybe call it `prompt_end` or `commandline_start` (maybe add `_line` as well)

src/screen.rs
```diff
@@ -1103,47 +1048,65 @@ impl Screen {
...
+            .desired
+            .prompt_lines
+            .saturating_sub(self.desired.scroll_amount);
+        let first_visible = self.desired.prompt_lines >= self.desired.scroll_amount;
```
krobelus: I guess this could be

	let (cmd_start, first_visible) = self.desired.prompt_lines.checked_sub(
						self.desired.scroll_amount)
						.map_or((0, false), |d| (d, true));

not sure if that's better.

---

Also, `first_visible` is a misleading name.
It's set to true if any of the prompt lines are visible (?),
so maybe `prompt_visible` or `is_prompt_visible`?

src/screen.rs
```diff
@@ -1103,47 +1048,65 @@ impl Screen {
...
+
+        let prompt_changed = self
             .actual_left_prompt
             .as_ref()
-            .is_none_or(|p| p != left_prompt)
```
krobelus, line 1115: can you steal the test cases from #11901

src/screen.rs
```diff
@@ -1103,47 +1048,65 @@ impl Screen {
...
-            || (self.scrolled() && is_final_rendering)
```
krobelus, line 1116: > Sorry for my sudden disappearance. I kept increasing the scope of
 > my PR and eventually burned out.

No worries - it was not a regression of existing functionality,
and we could also have reverted
and AFAICT we could also have reverted
267b16235d3 (Fix transient prompt mode for single-line prompts, 2025-04-14)
because I realized only this week that it was obsoleted by
dd4c04e2ff8 (Fix empty soft-wrapped line not being removed before execution, 2025-05-04).

But this new approach seems like the better direction.

---

I wonder what was the scope.

fish has some historical baggage, and the code is not as clean compared to a modern text editor.
We still don't have visual line movement (i.e. ctrl-n should move down one line inside soft-wrapped lines) like Emacs or Claude Code..

I think our rendering component is not ideal,
it tries to do a premature optimization to minimize screen writes,
but the optimization is useless as soon as the shared prefix is invalidated.

I think it would make more sense to compute a generic diff
between actual and desired screen contents, and output
commands that apply the diff.
Then we can separate the layout logic from the terminal-command logic,
and make things easier to test without a terminal.

We also render on every single key stroke, which can be slow.
We should maybe have an event loop, to prevent rendering unnecessarily often.

We also have this hack in `write() -> check_status()`,
which has caused a lot of problems (`git log --grep=[rR]efresh.TTY`).

```
        if self.mtime_stdout_stderr != mtime_stdout_stderr() {
            // Ok, someone has been messing with our screen. We will want to repaint. However, we do not
            // know where the cursor is. It is our best bet that we are still on the same line, so we
            // move to the beginning of the line, reset the modeled screen contents, and then set the
            // modeled cursor y-pos to its earlier value.
            let prev_line = self.actual.cursor.y;
            self.reset_line(true /* repaint prompt */);
            self.actual.cursor.y = prev_line;
        }
```

I guess something like this will always be necessary
as long as we allow background processes to write to the terminal
(not sure if we have the choice),
but maybe we should make this more robust,
by actually asking the terminal about the cursor position.. not sure though.

Anyway, if we really want to completely throw away and rewrite a component, we could stage that with a feature flag if needed.

 > Like #11901, this fixes #11875. It also makes fish hide the part of
 > a multiline prompt that leaves the viewport and removes an old hack
 > that was previously necessary for locating the first prompt line.
 >
 > I have tried to avoid introducing bugs this time, but I would be
 > grateful for any help with testing.

I guess interesting tests are mostly around
soft wrapping, scrolling, multiline prompts,
prompts that randomly change height/width,

src/screen.rs
```diff
@@ -1103,47 +1048,65 @@ impl Screen {
...
-            || Some(left_prompt_width) == screen_width && self.should_wrap(0)
-        {
+            .is_none_or(|p| first_visible && p != left_prompt)
+            || first_visible && self.actual.scroll_amount != self.desired.scroll_amount;
```
krobelus: so prompt_changed is the disjunction of
- the previous left prompt was empty
- the actual prompt has changed **and** `first_visible`
- the scroll amount has changed **and** `first_visible`

this looks mostly correct;
we could certainly trade some false positives for simplicity,
but it's probably best how you did it.

Though I wonder if anything else can change (see next comment).

One simplification that I think is safe is to require `fist_visible` in all three cases:

```diff
diff --git a/src/screen.rs b/src/screen.rs
index a4528b8d22..c76f37910a 100644
--- a/src/screen.rs
+++ b/src/screen.rs
@@ -1054,16 +1054,17 @@
             .saturating_sub(self.desired.scroll_amount);
         let first_visible = self.desired.prompt_lines >= self.desired.scroll_amount;
 
-        let prompt_changed = self
-            .actual_left_prompt
-            .as_ref()
-            .is_none_or(|p| first_visible && p != left_prompt)
-            || first_visible && self.actual.scroll_amount != self.desired.scroll_amount;
-        let prompt_softwrap =
-            first_visible && Some(left_prompt_width) == screen_width && self.should_wrap(cmd_start);
-
         // Output the left prompt if it has changed or we need to refresh softwrap.
-        if prompt_changed || prompt_softwrap {
+        if let Some((prompt_changed, prompt_softwrap)) = first_visible.then(|| {
+            let prompt_changed = self
+                .actual_left_prompt
+                .as_ref()
+                .is_none_or(|p| p != left_prompt)
+                || self.actual.scroll_amount != self.desired.scroll_amount;
+            let prompt_softwrap =
+                Some(left_prompt_width) == screen_width && self.should_wrap(cmd_start);
+            (prompt_changed, prompt_softwrap)
+        }) {
             self.r#move(0, 0);
             let mut start;
             if left_prompt_layout.line_breaks.len() <= self.desired.scroll_amount {
```

src/screen.rs
```diff
@@ -1103,47 +1048,65 @@ impl Screen {
...
+        let prompt_softwrap =
```
krobelus: I think window resizing can cause a soft-wrapped prompt to no longer be soft-wrapped.
We need to redraw the prompt in that case.
It looks like that's not handled here (only the other direction is).
I can't reproduce a problem, probably because SIGWINCH already triggers a prompt re-exec.
But that begs the question, why do we have specific redraw detection for soft-wrapped prompts?
I guess because it can change whenever we scroll, causing `self.should_wrap(cmd_start)` to become true?

but I can't reproduce a problem with that either, using

```diff
diff --git a/src/screen.rs b/src/screen.rs
index a4528b8d22..9996ac8c2d 100644
--- a/src/screen.rs
+++ b/src/screen.rs
@@ -1059,11 +1059,9 @@
             .as_ref()
             .is_none_or(|p| first_visible && p != left_prompt)
             || first_visible && self.actual.scroll_amount != self.desired.scroll_amount;
-        let prompt_softwrap =
-            first_visible && Some(left_prompt_width) == screen_width && self.should_wrap(cmd_start);
 
         // Output the left prompt if it has changed or we need to refresh softwrap.
-        if prompt_changed || prompt_softwrap {
+        if prompt_changed {
             self.r#move(0, 0);
             let mut start;
             if left_prompt_layout.line_breaks.len() <= self.desired.scroll_amount {
@@ -1102,6 +1100,9 @@
             }
             self.actual_left_prompt = Some(left_prompt.to_owned());
             self.actual.cursor.y = cmd_start;
+
+            let prompt_softwrap =
+                Some(left_prompt_width) == screen_width && self.should_wrap(cmd_start);
             if prompt_softwrap {
                 self.soft_wrap_location = Some(Cursor {
                     x: 0,
```

src/screen.rs
```diff
@@ -1103,47 +1048,65 @@ impl Screen {
...
+            first_visible && Some(left_prompt_width) == screen_width && self.should_wrap(cmd_start);
```
krobelus: probably we can assert that

	Some(left_prompt_width) <= screen_width

here?
If not, then the softwrap detection would be broken.

src/screen.rs
```diff
@@ -1103,47 +1048,65 @@ impl Screen {
...
                     }
                     self.write_str(&left_prompt[start..=line_break]);
                     start = line_break + 1;
                 }
             } else {
-                start = left_prompt_layout.line_breaks.last().map_or(0, |lb| lb + 1);
+                self.r#move(0, cmd_start);
+                start = left_prompt_layout.line_breaks[self.desired.prompt_lines - 1] + 1;
+            }
+            if first_visible {
```
krobelus: If we take the diff I posted [above](https://github.com/fish-shell/fish-shell/pull/11911#discussion_r2415646894), then this condition would be tautological.

src/screen.rs
```diff
@@ -1103,47 +1048,65 @@ impl Screen {
...
+                self.write_str(&left_prompt[start..]);
+                self.actual.cursor.x = left_prompt_width;
             }
-            self.write_str(&left_prompt[start..]);
             self.actual_left_prompt = Some(left_prompt.to_owned());
-            self.actual.cursor.x = left_prompt_width;
-            if Some(left_prompt_width) == screen_width && self.should_wrap(0) {
-                self.soft_wrap_location = Some(Cursor { x: 0, y: 1 });
+            self.actual.cursor.y = cmd_start;
```
krobelus: so across Screen::write() and Screen::update() we're doing:

	self.desired.cursor.y = self.desired.prompt_lines;
	self.desired.scroll_amount = scrolled_cursor.scroll_amount;
	let cmd_start = desired.prompt_lines - desired.scroll_amount;
	if prompt_visible {
		self.actual.cursor.y = cmd_start;
	}

isn't it weird that we are first setting
self.desired.cursor.y to a wrong value only to correct it later
(if the prompt is visible)

src/screen.rs
```diff
@@ -1173,14 +1136,25 @@ impl Screen {
             // Note that skip_remaining is a width, not a character count.
             let mut skip_remaining = start_pos;
 
-            let shared_prefix = if self.scrolled() {
+            let previously_prompt_line = self.actual.prompt_lines >= self.actual.scroll_amount + i;
+            if previously_prompt_line && is_first_line && Some(left_prompt_width) != screen_width {
+                // Clean up first line from previous prompt
+                self.r#move(left_prompt_width, i);
+                self.write_command(ClearToEndOfLine);
```
krobelus: the first visible line from the previous prompt gets special handling because it contains the Osc133PromptStart marker, right?

I wonder why we can't use the `shared_prefix` logic though,
I guess because we don't want to clear anything left of `left_prompt_width`?
If it's just that, wouldn't it be simpler to do something like this
(rough sketch)

```diff
diff --git a/src/screen.rs b/src/screen.rs
index c76f37910a..5d8b6d6476 100644
--- a/src/screen.rs
+++ b/src/screen.rs
@@ -1138,11 +1138,6 @@
             let mut skip_remaining = start_pos;
 
             let previously_prompt_line = self.actual.prompt_lines >= self.actual.scroll_amount + i;
-            if previously_prompt_line && is_first_line && Some(left_prompt_width) != screen_width {
-                // Clean up first line from previous prompt
-                self.r#move(left_prompt_width, i);
-                self.write_command(ClearToEndOfLine);
-            };
 
             let shared_prefix = if self.desired.scroll_amount == 0 || previously_prompt_line {
                 0
@@ -1150,15 +1145,23 @@
                 line_shared_prefix(o_line(self, i), s_line(self, i))
             };
             let mut skip_prefix = shared_prefix;
-            if shared_prefix < o_line(self, i).indentation
-                || previously_prompt_line && !is_first_line
-            {
+            if shared_prefix < o_line(self, i).indentation || previously_prompt_line {
                 if !has_cleared_screen
                     && (o_line(self, i).indentation > s_line(self, i).indentation
-                        || previously_prompt_line && !is_first_line)
+                        || previously_prompt_line)
                 {
                     set_color(self, HighlightSpec::new());
-                    self.r#move(0, i);
+                    self.r#move(
+                        if previously_prompt_line
+                            && is_first_line
+                            && Some(left_prompt_width) != screen_width
+                        {
+                            left_prompt_width
+                        } else {
+                            0
+                        },
+                        i,
+                    );
                     self.write_command(if should_clear_screen_this_line {
                         ClearToEndOfScreen
                     } else {
```

src/screen.rs
```diff
@@ -1189,6 +1163,10 @@ impl Screen {
                     } else {
                         ClearToEndOfLine
                     });
+                    if i == 0 {
+                        // Restore prompt start if we deleted it
+                        self.write_command(Osc133PromptStart);
```
krobelus, line 1192: Hmm, how would we have deleted the prompt start marker?
I think the way you wrote it, we clear to the end of line/screen starting from `left_prompt_width`,
and we should have written Osc133PromptStart before that.

src/screen.rs
krobelus: maybe it's better to leave `self.desired.cursor.y` at `usize::MAX` initially, and then initialize it inside `Screen::update()` in all cases

src/screen.rs
krobelus: we can probably write a test for the presence of OSC 133 A with `tmux capture-pane -e`

src/screen.rs
```diff
@@ -1103,47 +1048,65 @@ impl Screen {
...
-                        mark_prompt_start(self);
+                        self.write_command(Osc133PromptStart);
                     }
                     self.write_str(&left_prompt[start..=line_break]);
                     start = line_break + 1;
                 }
             } else {
-                start = left_prompt_layout.line_breaks.last().map_or(0, |lb| lb + 1);
+                self.r#move(0, cmd_start);
+                start = left_prompt_layout.line_breaks[self.desired.prompt_lines - 1] + 1;
```
krobelus: this is crashing if a single-line left prompt is wrapped
(I think we add an empty line to the eventual layout but only later)

```diff
diff --git a/tests/checks/tmux-prompt.fish b/tests/checks/tmux-prompt.fish
index 3e002c67e2..f3a523f0f4 100644
--- a/tests/checks/tmux-prompt.fish
+++ b/tests/checks/tmux-prompt.fish
@@ -46,3 +46,13 @@
 # CHECK: …<----------------------------------------------two-last-characters-rendered->!!
 # CHECK: test "
 # CHECK: indent"
+
+isolated-tmux send-keys C-c C-l '
+    function fish_prompt
+        string repeat (math "$COLUMNS * 2") x
+    end
+' C-l 'echo hello'
+tmux-sleep
+isolated-tmux capture-pane -p
+# CHECK: …xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
+# CHECK: echo hello
```

src/screen.rs
krobelus: I'm stress-testing with this one, seems stable so far

	{
	    set -g fish_transient_prompt 1
	    random $(date +%d%H) # Set a random seed
	    functions --copy fish_prompt my-prompt
	    function gibberish
	        for line in (seq (random 0 (math "round($LINES * 1.2)")))
	            set -l columns (random 0 (math "round($COLUMNS * 1.2)"))
	            string repeat $columns $line
	        end
	    end
	    function fish_prompt
	        set -l p "$(my-prompt $argv)"
	        gibberish
	        if contains -- --final-rendering $argv
	            printf '$ '
	        else
	            printf %s $p
	        end
	    end
	}

src/screen.rs
kerty0: I realized number of prompt lines also stored as `left_prompt_layout.line_breaks.len()`

src/screen.rs
kerty0: Thanks for suggestions!

src/screen.rs
kerty0: Accidentally included it after doing `git rebase`. Will revert this.

src/screen.rs
kerty0: > I wonder what was the scope.

Not something as grand as rendering rewrite, just a bunch of small improvements which I tried to implement all at once. I still want to try to implement them, just more gradually.

But rendering rewrite would indeed be very nice.

src/screen.rs
kerty0: > One simplification that I think is safe is to require fist_visible in all three cases:

I deliberately don't require `first_visible` if prompt is empty because we want to print `Osc133PromptStart` even if prompt isn't visible.

src/screen.rs
kerty0: `self.should_wrap(cmd_start)` returns true if we going to create new line what softwraps or we going to change beginning of next line. Both of this cases requires reprinting of prompt to maintain softwrap.

> because SIGWINCH already triggers a prompt re-exec.

Is this true? I am not familiar with how fish handles input, but when testing on Alacritty (Linux) when resizing terminal, debugger sees SIGWINCH signal, but fish doesn't resize. In case this isn't intended behavior and can be fixed, do we need softwrap at all?

src/screen.rs
kerty0: At the start of `update` when calculating prompt layout we truncate prompt to screen width. Maybe it would be more appropriate to place assert in truncation function?

src/screen.rs
kerty0: We depend on current value of `self.desired.cursor` in `Screen::desired_append_char`, so `self.desired.cursor` is like a position of a virtual cursor used to write to `self.desired`.

src/screen.rs
kerty0: > Hmm, how would we have deleted the prompt start marker?

This only happens if prompt is not visible because of scroll.

> we can probably write a test for the presence of OSC 133 A with tmux capture-pane -e

I don't think tmux supports OSC 133 and from my quick test fish doesn't print it in tmux.

src/screen.rs
kerty0: Thanks for all the help!
Crash happens because of I forget what it should be `start = left_prompt_layout.line_breaks.last().map(|lb| lb + 1).unwrap_or(0)`

src/screen.rs
kerty0: Yes, this would be better. Thanks for idea!

src/screen.rs
krobelus: nice find. Though AIUI, we still need to capture some value to detect if anything about the prompt has changed. Today that's

	self.actual.scroll_amount != self.desired.scroll_amount

It's possible that "visible_prompt_lines" is the minimal amount of information that would suffice. Not sure

src/screen.rs
krobelus: > which I tried to implement all at once

Yeah, I like adding things to a to-do list and then doing them in a separate commit later.

src/screen.rs
krobelus: > we want to print `Osc133PromptStart` even if prompt isn't visible.

I see.  So this is an additional behavior change / fix.
Maybe it should go into a separate commit, probably after the main change here.
That might make the diffs easier to read, and improve git-blame.

I just tested prompt navigation with `ctrl-shift-{jk}` (kitty's default shortcuts);
and indeed, the prompt marker at the top left of a scrolled commandline is necesssary for `ctrl-shift-j` to actually jump back to the original state. 

This looks reasonable for now;
one could argue that kitty/foot could behave differently, but I guess it's better to let fish make the decision on how this should behave.
(I'm pretty sure the existing behavior exists because we have never thought about how OSC 133 features should work for the case where the prompt is entirely outside the viewport.)

In future, we could always print a small part of the prompt, even in scrolled command lines. Then this ambiguity will naturally go away.

(I haven't really understood what most other [OSC 133 commands](https://gitlab.freedesktop.org/Per_Bothner/specifications/blob/master/proposals/semantic-prompts.md#commands) are actually for in practice, for example I dont't get why the [prompt end one](https://github.com/fish-shell/fish-shell/issues/11837) is useful, which is why we don't print it yet)

src/screen.rs
krobelus: I agree that's the better sort of place, and over there it's probably obvious enough that we don't need the assertion. I mainly wasn't sure where this was set.

src/screen.rs
krobelus: I see, we still use it for writing out the command line (until we hit the viewport edge).
And if we did hit the viewport edge, then we later chop off prompt lines because the commandline is more important. Makes sense.

src/screen.rs
krobelus: > This only happens if prompt is not visible because of scroll.

Ok, so the three places where we write Osc133PromptStart are (in source order):
1. basically two cases
  - 1.1 if `actual_left_prompt` is none (i.e. at first rendering, or if previous prompt was scrolled) **and** the desired prompt is scrolled out of view
  - 1.2 if the prompt is single line
2. When we draw the first line of a multi-line prompt.
3. Here; we add `Osc133PromptStart` to the very first line of our screen if we clear the line due to shared prefix mismatch.

I think I get it. In

	for i in cmd_start..self.desired.line_count()

if `scroll_amount >= prompt_lines` then we'd already have written the marker via case 1.1 above (in the other cases cmd_start will be > 0, so the marker is on an earlier line).

Since the marker is zero-width, we can't avoid deleting it here, so we re-add it.

I wonder if our prompt marker could also be accidentally deleted later instead,
specifically in the `// Now actually output stuff.` part.
It should be possible to tell but I haven't figured it out yet..
If it can be deleted, we would add a similar `if i == 0` fix there.

> I don't think tmux supports OSC 133 and from my quick test fish doesn't print it in tmux.

It should, see 54962473449 (Avoid erasing OSC 133 prompt start marker with clr_eol, 2024-10-12)
Confirmed that we print it with this command. Need to press enter once to flush data through to grep:

	tmux new-session 'fish | cat -v | grep --color=auto 133'

src/screen.rs
krobelus: can use `last().map_or(0, |lb| lb + 1)` I think

src/screen.rs
krobelus: > fish doesn't resize.

Right, I forgot that `__fish_winch_handler` is only active if `fish_handle_reflow` is set to 1.

We enable that specifically for Konsole, VTE, WezTerm and alacritty.
(alacritty does not provide XTVERSION at this time, so we still use `$TERM`):

	or string match -q -- 'alacritty*' $TERM
	set -g fish_handle_reflow 0

(This is one of the last remaining places where we do user-agent sniffing;
would be nice to get rid of it; better solutions are mentioned in some of our reflowing issues.)

So my comment is only relevant for other terminals.

But I guess my comment 

 > can cause a soft-wrapped prompt to no longer be soft-wrapped. We need to redraw the prompt in that case.

is wrong, because when the previous prompt line is soft-wrapped and the new prompt is no longer,
the layout is already different, so we should automatically redraw.
So resizing shouldn't matter.

> self.should_wrap(cmd_start) returns true if we going to create new line what softwraps or we going to change beginning of next line. Both of this cases requires reprinting of prompt to maintain softwrap.

Ok, I see a difference if the very last line of the prompt would wrap:

	function fish_prompt
	    string repeat $COLUMNS x
	end

Without this condition, triple-clicking on the last prompt line will not select anything in the command line.
So yeah, this looks reasonable.

(it's a little weird that all our prompt lines are truncated to a single screen line, but no reason to change this now).

AFAICT this softwrap logic is only relevant if the prompt's last logical line wraps.
Maybe we can rename `prompt_softwrap` to reflect that?
Though I guess `left_prompt_width` was already misleading.
Maybe this works, I don't have enough context to judge.

```diff
diff --git a/src/screen.rs b/src/screen.rs
index 26091584bb..72f7f55351 100644
--- a/src/screen.rs
+++ b/src/screen.rs
@@ -1016,7 +1016,7 @@
 
         // Determine size of left and right prompt. Note these have already been truncated.
         let left_prompt_layout = cached_layouts.calc_prompt_layout(left_prompt, None, usize::MAX);
-        let left_prompt_width = left_prompt_layout.last_line_width;
+        let prompt_last_line_width = left_prompt_layout.last_line_width;
         let right_prompt_width = cached_layouts
             .calc_prompt_layout(right_prompt, None, usize::MAX)
             .last_line_width;
@@ -1065,11 +1065,12 @@
             .as_ref()
             .is_none_or(|p| first_visible && p != left_prompt)
             || first_visible && self.actual.scroll_amount != self.desired.scroll_amount;
-        let prompt_softwrap =
-            first_visible && Some(left_prompt_width) == screen_width && self.should_wrap(cmd_start);
+        let prompt_last_line_should_wrap = first_visible
+            && Some(prompt_last_line_width) == screen_width
+            && self.should_wrap(cmd_start);
 
         // Output the left prompt if it has changed or we need to refresh softwrap.
-        if prompt_changed || prompt_softwrap {
+        if prompt_changed || prompt_last_line_should_wrap {
             self.r#move(0, 0);
             let mut start;
             if left_prompt_layout.line_breaks.len() <= self.desired.scroll_amount {
@@ -1110,11 +1111,11 @@
             }
             if first_visible {
                 self.write_str(&left_prompt[start..]);
-                self.actual.cursor.x = left_prompt_width;
+                self.actual.cursor.x = prompt_last_line_width;
             }
             self.actual_left_prompt = Some(left_prompt.to_owned());
             self.actual.cursor.y = cmd_start;
-            if prompt_softwrap {
+            if prompt_last_line_should_wrap {
                 self.soft_wrap_location = Some(Cursor {
                     x: 0,
                     y: cmd_start + 1,
@@ -1133,7 +1134,11 @@
         for i in cmd_start..self.desired.line_count() {
             self.actual.create_line(i);
             let is_first_line: bool = first_visible && i == cmd_start;
-            let start_pos = if is_first_line { left_prompt_width } else { 0 };
+            let start_pos = if is_first_line {
+                prompt_last_line_width
+            } else {
+                0
+            };
             let mut current_width = 0;
             let mut has_cleared_line = false;
 
@@ -1149,9 +1154,12 @@
             let mut skip_remaining = start_pos;
 
             let previously_prompt_line = self.actual.prompt_lines >= self.actual.scroll_amount + i;
-            if previously_prompt_line && is_first_line && Some(left_prompt_width) != screen_width {
+            if previously_prompt_line
+                && is_first_line
+                && Some(prompt_last_line_width) != screen_width
+            {
                 // Clean up first line from previous prompt
-                self.r#move(left_prompt_width, i);
+                self.r#move(prompt_last_line_width, i);
                 self.write_command(ClearToEndOfLine);
             };
 
@@ -1204,7 +1212,7 @@
                 // line, don't skip over the last two characters so that we maintain soft-wrapping.
                 skip_remaining = skip_remaining.min(screen_width.unwrap() - 2);
                 if is_first_line {
-                    skip_remaining = skip_remaining.max(left_prompt_width);
+                    skip_remaining = skip_remaining.max(prompt_last_line_width);
                 }
             }
 
```

src/screen.rs
krobelus: > I haven't figured it out yet

this would be less relevant if we first computed a ScreenData diff (in a way that supports zero-width elements), and then output the diff with a generic implementation

src/screen.rs
kerty0: > It's possible that "visible_prompt_lines" is the minimal amount of information that would suffice.

Yes, it was.

CHANGELOG.rst
```diff
@@ -3,6 +3,8 @@ fish ?.?.? (released ???)
 
 Notable improvements and fixes
 ------------------------------
+- Fish now hides the portion of a multiline prompt that scrolls out of view, preventing duplicate lines after repainting with partially visible prompt.
```
krobelus: trailing whitespace here

CHANGELOG.rst
```diff
@@ -3,6 +3,8 @@ fish ?.?.? (released ???)
 
 Notable improvements and fixes
 ------------------------------
+- Fish now hides the portion of a multiline prompt that scrolls out of view, preventing duplicate lines after repainting with partially visible prompt.  
+- Fixed not properly clearing lines when a transient prompt contains more lines that the final prompt (:issue:`11875`).
```
krobelus: I like to add clickable links

```suggestion
- Fixed not properly clearing lines when a :ref:`transient prompt <transient-prompt>` contains more lines than the final prompt (:issue:`11875`).
```

src/screen.rs
```diff
@@ -1103,48 +1074,64 @@ impl Screen {
             need_clear_screen = true;
         }
 
-        // Output the left prompt if it has changed.
-        if self.scrolled() && !is_final_rendering {
-            self.r#move(0, 0);
-            self.write_command(ClearToEndOfLine);
```
krobelus, line 1109: Re f154bc2c7bb (Instead of prompt line_breaks track line_starts, 2025-10-14)
which does

```diff
-                for (i, &line_break) in left_prompt_layout.line_breaks.iter().enumerate() {
+                for (i, &next_line) in left_prompt_layout.line_starts.iter().enumerate() {
                     self.write_command(ClearToEndOfLine);
```

shouldn't it be `iter().skip(1).enumerate()` to avoid doing

	ClearToEndOfLine
	Osc133PromptStart
	ClearToEndOfLine

which would actually (and surprisingly) break the prompt marker on some terminals, see
https://github.com/tmux/tmux/issues/4183

src/screen.rs
```diff
@@ -1103,48 +1074,64 @@ impl Screen {
...
+        let prompt_last_line_screen_wide =
+            is_prompt_visible && Some(prompt_last_line_width) == screen_width;
+        let prompt_last_line_should_wrap =
+            prompt_last_line_screen_wide && self.should_wrap(prompt_last_line);
+
+        // Output the left prompt if it has changed or we need to refresh softwrap.
+        if prompt_changed || prompt_last_line_should_wrap {
+            self.r#move(0, 0);
+            if self.desired.visible_prompt_lines <= 1 {
+                self.write_command(Osc133PromptStart);
```
krobelus: Re 57c79f875a1 (Try to always maintain prompt market at (0, 0), 2025-10-15)

typo in the commit message: market -> marker.

src/screen.rs
```diff
@@ -1103,48 +1074,64 @@ impl Screen {
...
+            }
+            let mut start;
+            if prompt_changed {
```
krobelus, line 1098: Re 3935601d065 (Refactor prompt printing and var names in Screen::update, 2025-10-14)

So if we have `!prompt_changed && prompt_last_line_should_wrap`,
then we will no longer output Osc133PromptStart.

I think this makes sense because if only the soft wrap state changes,
then we'll definitely not clear the prompt start marker we've written earlier,
because that one was placed just before the prompt,
and we can assume that `$COLUMNS` is nonzero
(which is what I tried give an explanation for in https://github.com/fish-shell/fish-shell/pull/11911#discussion_r2419830242).

However, what I'm not so sure about is that we still redraw the
prompt marker if the prompt is single-line here:

                if left_prompt_layout.line_starts.len() <= 1 {
                    self.write_command(Osc133PromptStart);
                }

is there a reason we still do that rather than only  printing the marker when `prompt_change` is true?
Maybe the answer is already in a thread somewhere and I've forgotten about it, so maybe put it in the commit message.

src/screen.rs
```diff
@@ -1103,48 +1074,64 @@ impl Screen {
...
+                        self.write_command(Osc133PromptStart);
                     }
-                    self.write_str(&left_prompt[start..=line_break]);
-                    start = line_break + 1;
+                    self.write_str(&left_prompt[start..next_line]);
+                    start = next_line;
                 }
             } else {
-                start = left_prompt_layout.line_breaks.last().map_or(0, |lb| lb + 1);
+                self.r#move(0, prompt_last_line);
```
krobelus, line 1119: I'll try to finish my review tomorrow

src/screen.rs
```diff
@@ -1321,7 +1330,12 @@ impl Screen {
         // Don't do it when running in midnight_commander because of
         // https://midnight-commander.org/ticket/4258.
         if !MIDNIGHT_COMMANDER_HACK.load() {
-            self.r#move(0, 0);
+            self.r#move(
+                0,
+                self.desired
+                    .line_count()
+                    .saturating_sub(usize::try_from(termsize_last().height).unwrap()),
```
krobelus: I'm not sure I understand this change to the y coordinate it either seems a bit unrelated(?) to the rest of this commit,
or not obvious enough.

The computation here

	desired_line_count.saturating_sub(screen_height)

we've already done above when computing `ScrolledCursor::scroll_amount`.
So maybe we should pass `scroll_amount` to `update()` again, or at least extract a `scroll_amount` variable?

This hack exists because we want to emit \r rather than a more complicated sequence of movement commands.
AFAICT the y coordinate doesn't really matter, since we're gonna move again anyway,
but of course using the "current" line is the goal, which is probably what this change is doing?

src/screen.rs
```diff
@@ -1362,7 +1376,7 @@ pub fn screen_force_clear_to_end() {
 #[derive(Clone, Debug, Default, Eq, PartialEq)]
 pub struct PromptLayout {
     /// line breaks when rendering the prompt
-    pub line_breaks: Vec<usize>,
```
krobelus, line 1365: nit: stale comment

src/screen.rs
```diff
@@ -1934,6 +1932,8 @@ pub(crate) fn only_grayscale() -> bool {
 pub(crate) struct ScreenLayout {
     // The left prompt that we're going to use.
     pub(crate) left_prompt: WString,
+    // How much lines in the left prompt.
```
krobelus: (nit: "How many")

src/screen.rs
krobelus: We could take this as an opportunity to document things a bit (but no big deal, I'm not sure if this is the best place)

    /// Line start offsets for each logical line in the prompt, or equivalently, screen line, because prompt lines are not wrapped but truncated.

src/screen.rs
kerty0: > shouldn't it be iter().skip(1).enumerate() to avoid doing

Yes, I just forget to make a change while reworking commit history. Today tried `jj`, which seems to handle such things much more easily than `git`. Hope this will help me avoid such mistakes in the future.

> which would actually (and surprisingly) break the prompt marker on some terminals

From my testing, WezTerm also can erase the prompt marker with `ClearToEndOfLine`, while Kitty actually prepends it to commandline (you can see that with `ctrl+shift+h`). One can only dream of the world there behavior of terminal emulators is standardized.

src/screen.rs
kerty0: > is there a reason we still do that rather than only printing the marker when prompt_change is true?

I think I was just overly cautious when made that decision. It is indeed unnecessary and I don't think it gives much benefit, so I removed it.

src/screen.rs
kerty0: I shortened it to `/// Start offsets for each line in the truncated prompt`.
I also was thinking about refactoring prompt handling in the future (e.g. creating `TruncPrompt` with `Index` implementation)

src/screen.rs
kerty0: > I'm not sure I understand this change to the y coordinate it either seems a bit unrelated(?) to the rest of this commit

I initially changed `self.desired.cursor` in `Screen::write` to track position from `(0, 0)` because of how my mental model of the cursor worked. However, this introduced a bug, so I introduced this code change to fix it. 
But this approach don't have any advantages over original, so I reverted it.

src/screen.rs
krobelus: > jj

Nice, `jj` is great for this kind of stuff, I use it everywhere now.

> you can see that with ctrl+shift+h

Nice -- I didn't expect that because `foot` strips control sequences in `pipe-scrollback`

src/screen.rs
krobelus: I messed up my last paragraph, I meant to say that:
if we change it from `(0, 0)` to `(0, current_y)`,
then we may reduce the needless commands.
I wasn't able to tell whether that's done here or not.

Anyway, reverting seems good (and maybe doing the optimization later)

src/screen.rs
krobelus: > (e.g. creating TruncPrompt with Index implementation)

That sounds reasonable; it's passed through PromptCacheEntry and ScreenLayout as generic WString,
which is probably not very helpful.

FWIW I think the only reason why `PromptLayout` is public is that we use it for testing.
Rust wants us to put unit tests into the same file, but I think that's weird,
it's much easier to ignore test code (in `grep` etc.) when it's in a different file.

I think it's also weird that we do

	let layout = compute_layout(...)
	...
	let left_prompt = layout.left_prompt
	...
	let left_prompt_layout = cached_layouts.calc_prompt_layout(left_prompt, None, usize::MAX);

Specifically, I'm wondering why we do it piecemeal rather than have a single function compute the entire layout.

For testing, I wonder if we can swap out terminal output with a test-only output class that produces structured data
similar to tests like this https://github.com/mawww/kakoune/blob/master/test/highlight/rust/let/script
But there is also some terminal-related functionality like querying where we want to have at least a mocked terminal to read from
Maybe we should just make the tmux-based tests easier to work with;
it should be made easy to attach a debugger, or attach a tmux client without affecting the behavior of the test.
(also we should replace sleeps with things like `sleep-until` eventually).

kerty0: > also we should replace sleeps with things like `sleep-until` eventually

I actually looked at that and this is my attempt at creating synchronization functions using `tmux wait-for`:
```diff
diff --git a/tests/test_functions/isolated-tmux-start.fish b/tests/test_functions/isolated-tmux-start.fish
index d057af744b..1b1340d7ba 100644
--- a/tests/test_functions/isolated-tmux-start.fish
+++ b/tests/test_functions/isolated-tmux-start.fish
@@ -26,6 +26,22 @@
         or sleep 0.3
     end
 
+    function tmux-wait
+        isolated-tmux send-keys f12
+        isolated-tmux wait-for prompt
+    end
+
+    function tmux-capture
+        tmux-wait
+        isolated-tmux capture-pane -p
+    end
+    
+    function tmux-exec
+        isolated-tmux send-keys Enter
+        isolated-tmux wait-for postexec
+        tmux-wait
+    end
+
     function sleep-until
         set -l cmd $argv[1]
         set -l i 0
@@ -48,6 +64,13 @@
         set fish_history ""
         # No transient prompt.
         set fish_transient_prompt 0
+
+        function __tmux_postexec_notify --on-event fish_postexec
+            tmux wait-for -S postexec
+        end
+        
+        bind f12 "tmux wait-for -S prompt;"
+        bind -M insert f12 "tmux wait-for -S prompt"
     ' $argv
     # Set the correct permissions for the newly created socket to allow future connections.
     # This is required at least under WSL or else each invocation will return a permissions error.
```

But after stress testing it by spawning ~300 python test drivers simultaneously, it sadly sometimes fails.
So is repeatedly checking captured stdout only consistent option?

src/screen.rs
kerty0: > if we change it from (0, 0) to (0, current_y),

This somehow fails in `tmux-prompt.fish`. Will look at it tomorrow.

src/screen.rs
krobelus: Wow I didn't know about `tmux wait-for`.
I tested some variations of

	sh -c 'for i in $(seq 100); do
	    tmux wait-for -S topic
	done' &
	for i in $(seq 100); do
	    tmux wait-for topic
	done

but nothing worked.

There's probably a lot things that could be improved about these tests
It's weird that the test script is a fish script,
when these tests mainly want to test the fish running inside tmux.

I wonder if it would be slightly better to write the test script in Python or Rust, without littlecheck.
littlecheck is a reimplementation of LLVM's lit+FileCheck combo,
which allows to add RUN-commands and CHECK-assertions to test files,
rather than requiring those test files to be able to say, do `import pytest; assert actual == expected`.
I guess that makes a bit more sense for script tests like tests/checks/abbr.fish,
where an `assert_eq $(some command) "expected"` currently can't print the line number that failed,
but we could either add enough reflection to fish or write the script in Python or Rust and use their standard test framework.

Anyway, I think we need to synchronize with fish, so using blocking
fifos seems like a reasonable choice (I guess flock() would work too
one some systems):

```patch
    Start adding proper waiting to tmux tests

    I assume that

    	tmux send-keys a
    	tmux send-keys b

    are always processed in this order by the tmux server.

    However the actual processing of keys (by tmux/fish etc.) is async,
    which motivates a lot of sleeps.  Replace the sleeps with a synchronous
    command, that only returns once fish has finished executing the
    command.

    We still don't know when tmux has actually rendered what's been written
    to the TTY, but given that the following works without sleeps now:

        for x in $(seq 300); do
            printf %s\\n 'tests/test_driver.py target/debug tests/checks/tmux-abbr.fish'
        done | parallel

    I'd guess that "tmux capture-pane" already forces a render.

diff --git a/tests/checks/tmux-abbr.fish b/tests/checks/tmux-abbr.fish
index 1d713c8425..10ce1e3485 100644
--- a/tests/checks/tmux-abbr.fish
+++ b/tests/checks/tmux-abbr.fish
@@ -10,34 +10,34 @@
 isolated-tmux-start
 
 # Expand abbreviations on space.
-isolated-tmux send-keys abbr-test Space arg1 Enter
-tmux-sleep
+isolated-tmux send-keys abbr-test Space arg1
+tmux-execute
 # CHECK: prompt {{\d+}}> abbr-test [expanded] arg1
 
 # Expand abbreviations at the cursor when executing.
-isolated-tmux send-keys abbr-test Enter
-tmux-sleep
+isolated-tmux send-keys abbr-test
+tmux-execute
 # CHECK: prompt {{\d+}}> abbr-test [expanded]
 
 # Use Control+Z right after abbreviation expansion, to keep going without expanding.
-isolated-tmux send-keys abbr-test Space C-z arg2 Enter
-tmux-sleep
+isolated-tmux send-keys abbr-test Space C-z arg2
+tmux-execute
 # CHECK: prompt {{\d+}}> abbr-test arg2
 
 # Same with a redundant space; it does not expand abbreviations.
-isolated-tmux send-keys C-u abbr-test Space C-z Space arg2 Enter
-tmux-sleep
+isolated-tmux send-keys C-u abbr-test Space C-z Space arg2
+tmux-execute
 # CHECK: prompt {{\d+}}> abbr-test  arg2
 
 # Or use Control+Space to the same effect.
-isolated-tmux send-keys abbr-test C-Space arg3 Enter
-tmux-sleep
+isolated-tmux send-keys abbr-test C-Space arg3
+tmux-execute
 # CHECK: prompt {{\d+}}> abbr-test arg3
 
 # Do not expand abbreviation if the cursor is not at the command, even if it's just white space.
 # This makes the behavior more consistent with the above two scenarios.
-isolated-tmux send-keys abbr-test C-Space Enter
-tmux-sleep
+isolated-tmux send-keys abbr-test C-Space
+tmux-execute
 # CHECK: prompt {{\d+}}> abbr-test
 
 # CHECK: prompt {{\d+}}>
diff --git a/tests/test_functions/isolated-tmux-start.fish b/tests/test_functions/isolated-tmux-start.fish
index faddeff3f9..2bcc835dc9 100644
--- a/tests/test_functions/isolated-tmux-start.fish
+++ b/tests/test_functions/isolated-tmux-start.fish
@@ -39,8 +39,21 @@
         end
     end
 
+    mkfifo fifo
+
+    function tmux-execute
+        isolated-tmux send-keys \e\]fish-test-execute\a
+        open_fifo '<'
+        or exit
+    end
+
     set -l fish (status fish-path)
     isolated-tmux new-session -x 80 -y 10 -d $fish -C '
+        function tmux-execute-completed
+            open_fifo ">"
+            or exit
+        end
+        bind \e\]fish-test-execute\a execute tmux-execute-completed
         # This is similar to "tests/interactive.config".
         function fish_greeting; end
         function fish_prompt; printf "prompt $status_generation> "; end
diff --git a/tests/test_functions/open_fifo.fish b/tests/test_functions/open_fifo.fish
new file mode 100644
index 0000000000..11b94c1850
--- /dev/null
+++ b/tests/test_functions/open_fifo.fish
@@ -0,0 +1,8 @@
+function open_fifo -a redirection
+    timeout 180 sh -c "exec $redirection fifo"
+    set -l saved_status $status
+    if test $saved_status -eq 124
+        echo >&2 "error: timed out in 'exec $redirection fifo'"
+    end
+    return $saved_status
+end
```

kerty0: > but nothing worked.

It because 'tmux wait-for -S` is non-blocking and has backlog of 1 for given channel. E.g:
```fish
tmux wait-for -S channel
tmux wait-for -S channel 
tmux wait-for -S channel 
tmux wait-for channel 
echo 1
tmux wait-for channel
echo 2
tmux wait-for channel
echo 3
```
will only print 1. 
This works:
```fish
fish -c 'for i in (seq 100); tmux wait-for -S $i; end' &
for i in (seq 100)
    tmux wait-for $i
end
```

krobelus: yeah that would explain it. If possible, we can use either `tmux wait-for` or fifos to synchronize, probably no big difference.

Though I wonder why `tmux wait-for` didn't work for you with the hardcoded channel name,
because each tmux session should be isolated, and only use one channel at a time.
Maybe we can reduce it down to a tmux bug.

We probably still want a (high) timeout on the blocking wait, so CI systems don't hang for hours.

`fish_postexec` sounds good too, seems more reliable.

For the cases where we are capturing a command line state before execution, we'll use your `tmux-wait` approach.

Probably we can speed up overall test runtime by a lot.

src/screen.rs
```diff
@@ -457,20 +463,20 @@ impl Screen {
...
-                    .cursor
-                    .y
-                    .checked_sub(1)
-                    .is_some_and(|y| self.desired.line_datas[y].is_soft_wrapped)
+                && self.desired.cursor.y.checked_sub(1).is_some_and(|y| {
+                    self.desired.line_datas[y].is_soft_wrapped
+                        && self.desired.line_datas[y]
+                            .color_at(self.desired.line_datas[y].len() - 1)
+                            .foreground
+                            == HighlightRole::autosuggestion
```
krobelus, line 473: this looks reasonable, might be a small layering violation though.
I wonder if we can improve this future, maybe `is_soft_wrapped` should be an enum with different values for "actual softwrap" or "softwrapped autosuggestion".
This stuff is super complicated..

src/screen.rs
```diff
@@ -1244,6 +1246,10 @@ impl Screen {
                     set_color(self, HighlightSpec::new());
                     self.r#move(current_width, i);
                     self.write_command(ClearToEndOfScreen);
+                    if i == 0 && current_width == 0 {
+                        // Restore prompt marker if we deleted it
+                        self.write_command(Osc133PromptStart);
```
krobelus, line 1251: I can't find a bug in the Osc133PromptStart placement, but it sure is hard to understand.

Maybe we can use the approach of drawing everything else first,
and then move the cursor to (0, 0) and draw `Osc133PromptStart` there.
Except in the cases where we didn't delete an earlier marker,
assuming that detecting this case is less complicated than what we have now.

If we're very confident in the correctness, we could do that change first,
to make the main change smaller.
But it's also fine as a follow-up commit/PR.

src/screen.rs
```diff
@@ -1361,11 +1367,9 @@ pub fn screen_force_clear_to_end() {
 /// Information about the layout of a prompt.
 #[derive(Clone, Debug, Default, Eq, PartialEq)]
 pub struct PromptLayout {
-    /// line breaks when rendering the prompt
-    pub line_breaks: Vec<usize>,
-    /// width of the longest line
-    pub max_line_width: usize,
```
krobelus, line 1367: yeah that's the problem with using `pub` for testing,
we currently don't get "unused" warnings (though [that might change](https://internals.rust-lang.org/t/no-need-for-pub-crate-in-binary-crates/23607)), maybe we can write a clippy lint..

tests/checks/tmux-commandline.fish
```diff
@@ -42,29 +42,40 @@ isolated-tmux capture-pane -p
...
+isolated-tmux send-keys C-l '
+    function fish_right_prompt
+        echo right-prompt
+    end
+' 'commandline -i ": \'$(seq (math $LINES \* 2))\'"' Enter Enter
 tmux-sleep
-isolated-tmux capture-pane -p
+isolated-tmux capture-pane -p -S -12
+# CHECK: prompt 4> commandline -i ": '$(seq (math $LINES \* 2))'"            right-prompt
+# CHECK: prompt 5> : '1                                                      right-prompt
```
krobelus, line 53: So `-S -12` tests that we no longer have spurious lines in tmux scrollback. Nice.

PR #11918, opened 2025-10-08
style: replace black with ruff for Python formatting

### LINKED ISSUE #11894: adapt ruff format instead of black
https://docs.astral.sh/ruff/formatter/

https://github.com/fish-shell/fish-shell/blob/fbfd29d6d20429743d3755ff617ee2e11cfc61c6/build_tools/style.fish#L79

pros: it didn't depend on python.

danielrainer:
Ruff's default format is very similar to black's, so there are only a few changes made to our Python code. They are all contained in this commit. The primary benefit of this change is that ruff's performance is about an order of magnitude better, reducing runtime on this repo down to under 20ms on my machine, compared to over 150ms with black, and even more if any changes are performed by black.

Closes #11894

## TODOs:
- [ ] Do we want any additional communication telling developers about ruff?
- [ ] Currently, this does not run in CI (neither did black), which is not ideal. How should we change that?
- [ ] If we-re using ruff for formatting, we might as well use its linter without introducing additional dependencies. Do we want that?


CONTRIBUTING.rst
```diff
@@ -110,7 +110,7 @@ before committing your change. That will run our autoformatters:
 
 - ``rustfmt`` for Rust
 - ``fish_indent`` (shipped with fish) for fish script
-- ``black`` for python
+- ``ruff format`` for python
```
krobelus, line 113: LGTM.

For historical reasons, we don't have formatting checks in CI yet, but we can fix that now.

That means we should rename

	git mv .github/workflows/{main,test}.yml
	git mv .github/workflows/{rust_checks,lint}.yml

and maybe also

	git mv .github/workflows/{docker_builds,build_docker_images}.yml

build_tools/style.fish
```diff
@@ -44,7 +44,7 @@ if test $all = yes
         end
     end
     set fish_files $workspace_root/{benchmarks,build_tools,etc,share}/**.fish
-    set python_files {doc_src,share,tests}/**.py
```
krobelus, line 47: it's weird that fish_files uses $workspace_root an python_files doesn't.

I suppose the most natural behavior would be to format only
things inside the current directory

	set fish_files path filter (string match -- "$PWD/*" \
		$workspace_root/{benchmarks,build_tools,etc,share}/**.fish)
	set python_files .

build_tools/style.fish
```diff
@@ -44,7 +44,7 @@ if test $all = yes
...
+    set python_files .
```
krobelus: Yeah given that ruff tries to respect gitignore, I'm good with this change.
(prior discussion for shellcheck is at https://github.com/fish-shell/fish-shell/pull/11608)

build_tools/style.fish
```diff
@@ -76,19 +76,19 @@ if set -q fish_files[1]
...
+    echo === Running "$green"ruff format"$normal"
     if set -l -q _flag_check
-        if not black --check $python_files
+        if not ruff format --check $python_files
             echo $red"Python files are not formatted correctly."$normal
             exit 1
         end
     else
-        black $python_files
+        ruff format $python_files
```
krobelus, line 91: FWIW, as mentioned in [Clippy in CI](https://github.com/fish-shell/fish-shell/discussions/11584#discussioncomment-13672983),
we have some unpinned packages from PyPI.

- ruff
- sphinx
- sphinx-markdown-builder
- pexpect
- (not in master today) python3-polib, python3-git, python-feedgen
- (fish-site) jekyll
- (fish-site) jekyll-feed

It's probably not a huge deal but if we ever
- run into version mismatch problems,
- or want reproducible builds

then we could specify versions.

Probably using `pyproject.toml`/`uv.lock` (e.g. `uv init && uv add --dev ruff`)

But we can defer that for now since we never had problems with unpinned black.
`sphinx-markdown-builder` is something that we should probably pin because it's not even on PyPI yet.

build_tools/style.fish
danielrainer: > it's weird that fish_files uses $workspace_root an python_files doesn't.

Yeah, that's not ideal. Is there a reason one wouldn't want formatting to run on all files ruff formats when called from the workspace root? If not, I think we should change it to

```fish
set python_files $workspace_root
```

which would make it irrelevant from where the script is invoked.

build_tools/style.fish
danielrainer: Pinning dependencies is certainly a good idea. Ideally, it would be done in a way that does not result in installing dependencies which are not needed for the current job in CI.

I agree that this does not need to be done in this PR, but is better addressed in a separate one.

CONTRIBUTING.rst
danielrainer: #11922

danielrainer: #11923

build_tools/style.fish
danielrainer: I made that change, since I believe it makes more sense to always format the entire repo, especially when used via `check.sh`. If only specific files should be formatted, the script already has options for that.

krobelus: sorry I merged to quickly (though I guess this is also unrelated).

I don't have a strong opinion on whether `style.fish` should format the whole repo, or only a subdirectory
(it doesn't matter for us)
The only reason I suggested this behavior was that `cargo fmt` behaves this way.

I think it probably makes sense to make at least `check.sh` independent of cwd for now (unless we can actually make it reliably check individual subdirectories like `crates/printf`).

krobelus: so to make things consistent, we'd need to change the `cargo fmt` invocation as well.

krobelus:
> Do we want any additional communication telling developers about ruff?

the CI should be fine

> If we-re using ruff for formatting, we might as well use its linter without introducing additional dependencies. Do we want that?

I've been looking at some of Cargo's xtasks for man pages (pointed out [here](https://internals.rust-lang.org/t/how-fail-on-cargo-warning-warnings-from-build-rs/23590/5)),
which makes me wonder if we should replace some/all our shell and Python scripts with xtasks.. or maybe only the nontrivial ones.
I still haven't really gotten to properly reviewing the "version as required env var" PR, I'm not sure if we should take that one (because it causes build invalidation when I change a commit message), but I think there are some xtasks in there.

It might be nice if we could run the "checks" tests using `cargo test`

Then we don't need to invest time in languages we don't really use.
(not sure if pexpect has a good Rust-replacement already)

krobelus:
but yeah, `ruff lint` sounds fine if `ruff lsp` works well for our project layout

danielrainer:
> which makes me wonder if we should replace some/all our shell and Python scripts with xtasks.. or maybe only the nontrivial ones.

I used the PR you mention to try the xtask approach and I quite like it. For some things we currently have scripts for, we could add an xtask that just calls the script. Then, we could later replace the script with an implementation in Rust if we want, without breaking workflows.

> Then we don't need to invest time in languages we don't really use.
(not sure if pexpect has a good Rust-replacement already)

Yeah, I'd be happy to get rid of Python, and some of the shell scripts. There's quite a bit of stuff that would have to be rewritten, but this can be done gradually.

build_tools/style.fish
danielrainer: #11925

build_tools/style.fish
krobelus: >  Do we want any additional communication telling developers about ruff?

FWIW I'm adding `pyproject.toml`/`uv.lock` for [reproducible builds](https://github.com/fish-shell/fish-shell/pull/12000/commits/a33ddf6817532588764229752394dddde56f392f#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R1-R14),
a nice benefit is that our README can now tell people to run `uv run cargo install --path .`,
which will automatically install the right version of `sphinx-build` in a virtualenv.

PR #11948, opened 2025-10-15
fs: open file with O_RDWR before fsync

### LINKED ISSUE #11934: Test `tests::env_universal_common::test_universal` blocks forever on MSYS2
I'm on the master branch (b88622b), using MSYS2 on Windows 11.
`tests::env_universal_common::test_universal` waits forever and the temporary fish intance (fish-<hex>.exe) needs to be killed wit the task manager.

I believe there are two bugs behind that blockage:
- one causes the test to fail (panic)
- the other causes the blockage as a consequence of the panic

(Note: I'm filing only one issue for now because I don't know how much of the 2nd bug is because of Fish and how it uses the libc API, and how much is on the MSYS2 side. I can split it in two if you think it's better)

The link between the two is a helper thread panicking on the assertion checking if the uvar could be sync'ed:
```
thread '<unnamed>' panicked at src/tests/env_universal_common.rs:25:9:
Failed to sync universal variables after modification
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: fish::tests::env_universal_common::test_universal_helper
             at [...]/src/tests/env_universal_common.rs:25:9
   3: fish::tests::env_universal_common::test_universal::{{closure}}::{{closure}}
             at [...]/src/tests/env_universal_common.rs:50:34
```

### The consequence of the panic
This bug probably exists on all platforms, not just MSYS2.
Because of the panicked thread, `iothread_drain_all` in the main thread never returns, waiting forever for the `iothread_port` to become readable (`iothread_service_main_with_timeout`)

Since there is no way the guarantee that a thread won't panic, even without the assert, I believe the test implementation is flawed.

### The reason of the panic
This bug is causing the test to fail/assert. The assertion triggers because `uvars.sync()` returns false, and it does so because of an `errno` 13 "Permission denied" when calling `libc::fsync(fd)` in `fs::fsync()` (
```
   2: fish::fs::fsync
             at [...]/src/fs.rs:193:17
   3: fish::fs::LockedFile::fsync_close_and_keep_lock
             at [...]/src/fs.rs:178:9
   4: fish::fs::rewrite_via_temporary_file::try_rewriting
             at [...]/src/fs.rs:393:51
   5: fish::fs::rewrite_via_temporary_file
             at [...]/src/fs.rs:478:18
   6: fish::env_universal_common::EnvUniversal::sync
             at [...]/src/env_universal_common.rs:207:15
   7: fish::tests::env_universal_common::test_universal_helper
             at [...]/src/tests/env_universal_common.rs:24:33
```

I don't know why `sync` would return a "permission denied". There is no such error in the Linux fsync man page at least. The var file (`./target/fish-test/test/fish_uvars_test/varsfile.txt`) is empty in the end though. So something is indeed wrong.
Similarly `set -U` does not work. The call does not return any error, but `~/.config/fish/fish_variables` is left unchanged.

### LINKED ISSUE #11947: Universal variables are broken in MSYS2
```bash
$ fish --version
fish, version 4.1.2-1 (Built by MSYS2 project)

$ uname
MINGW64_NT-10.0-26100
```

It seems the latest version of fish for MSYS2 introduced an issue where universal variables are no longer working, here's an example:

```bash
$ set -U foo 123
$ echo $foo
123
$ exit

# after restarting
$ echo $foo
# empty
```

Deleting `~/.config/fish/fish_variables` results in it being recreated, but completely empty. This is definitely a regression of some sort, since on a different machine I had version `4.0.2-1` and when I upgraded it, the issue started appearing there as well.

Berrysoft:
## Description

I'm not sure if it's OK for all POSIX platforms, but I think that it would be fine?

Also added a piece of log when `fsync` fails.

Closes #11934 

Closes #11947

### DIFF
--- a/src/fs.rs
+++ b/src/fs.rs
@@ -115,7 +115,7 @@ impl LockingMode {
             Self::Exclusive(WriteMethod::Append) => {
                 OFlag::O_WRONLY | OFlag::O_APPEND | OFlag::O_CREAT
             }
-            Self::Exclusive(WriteMethod::RenameIntoPlace) => OFlag::O_RDONLY | OFlag::O_CREAT,
+            Self::Exclusive(WriteMethod::RenameIntoPlace) => OFlag::O_RDWR | OFlag::O_CREAT,
         }
     }
 }
@@ -192,6 +192,7 @@ pub fn fsync(file: &File) -> std::io::Result<()> {
             -1 => {
                 let os_error = std::io::Error::last_os_error();
                 if os_error.kind() != std::io::ErrorKind::Interrupted {
+                    FLOGF!(synced_file_access, "fsync failed: %s", os_error);
                     return Err(os_error);
                 }
             }


krobelus:
thanks. Would be great if we could enable some tests on Cygwin
(at least we have a CI job now)

Berrysoft:
> Would be great if we could enable some tests on Cygwin

I think I'll leave the work to other PRs, or @Nahor who is working on CI tests. There are still some tests hanging or failing.

Issue #11953, opened 2025-10-16
Enable custom man command, e.g. batman


harkabeeparolus:
I'm using **batman** from [bat-extras](https://github.com/eth-p/bat-extras) with [bat](https://github.com/sharkdp/bat). It makes man pages very pretty and colorized. 😊 Unfortunately this does not work with **man.fish** (`type man`), since it has a hardcoded:

```fish
command man $argv
```

I currently just use `abbr man batman`, but this fails me in two ways:

* `batman` doesn't find fish's built-in man pages.
* I have no clean way to make the fish interactive help (Alt-h) use `batman`.

It would be nice if the user could somehow configure fish to run e.g. `batman` instead of `man`, and still find its own internal man pages.

harkabeeparolus:
Omg, my bad. 😳

Using `batman` as a `$MANPAGER` did not work back when I started using it.
But the [bat-extras](https://github.com/eth-p/bat-extras) project actually added this feature last Summer! https://github.com/eth-p/bat-extras/releases/tag/v2024.07.10

Nowadays, you can just do:

```fish
batman --export-env | source
```

I take it all back. I'm super happy now. Cheers everyone! ☺️

PR #11957, opened 2025-10-16
man.fish: Let users customize man command in __fish_man_command

harkabeeparolus:
## Description

Let users choose a custom man command, such as **batman**, by setting the __fish_man_command variable.

Fixes issue #11953

## TODOs:

### DIFF
--- a/share/functions/man.fish
+++ b/share/functions/man.fish
@@ -11,7 +11,7 @@ function man
     # man pages priority, without having to put fish's bin directories first in $PATH.
 
     if not set -q __fish_man_command
-	set -g __fish_man_command man
+        set -g __fish_man_command man
     end
 
     # Preserve the existing MANPATH, and default to the system path (the empty string).
@@ -64,12 +64,12 @@ function man
     if not set -q argv[2] && status list-files "man/man1/$argv[1].1" &>/dev/null
         set -l basename $argv[1].1
         function __fish_man -V basename -a man1
-	    command $__fish_man_command $man1/$basename
+            command $__fish_man_command $man1/$basename
         end
         __fish_data_with_directory man/man1 \
             (string escape --style=regex -- $basename) __fish_man
         __fish_with_status functions --erase __fish_man
     else
-	command $__fish_man_command $argv
+        command $__fish_man_command $argv
     end
 end


CHANGELOG.rst
```diff
@@ -24,6 +24,10 @@ For distributors
 fish 4.1.3 (released ???)
 =========================
 
+Notable improvements and fixes
+------------------------------
+- Let users customize the **man** command (e.g. batman) by setting ``__fish_man_command`` (:issue:`11953`).
```
krobelus, line 29: wrong release (though we might combine them into one anyway)

share/functions/man.fish
```diff
@@ -10,6 +10,10 @@ function man
     # by prepending our fish datadir to man. This also ensures that man gives fish's
     # man pages priority, without having to put fish's bin directories first in $PATH.
 
+    if not set -q __fish_man_command
+        set -g __fish_man_command man
+    end
```
krobelus, line 15: Not sure about this. I don't know what batman is,
if you wish to customize interactive behavior, you'd set MANPAGER etc.

man should behave like the external command;
the fact that we override it in the first place is a hack.
This type of improvement is probably best made at the system level,
so it works in e.g. bash too.

Do you have a specific command that you want to use?

share/functions/man.fish
```diff
@@ -60,12 +64,12 @@ function man
     if not set -q argv[2] && status list-files "man/man1/$argv[1].1" &>/dev/null
         set -l basename $argv[1].1
         function __fish_man -V basename -a man1
-            command man $man1/$basename
+            command $__fish_man_command $man1/$basename
```
krobelus, line 67: variables and functions starting with "__fish_" are "private" and
not supposed to be used directly, like in Python.

share/functions/man.fish
harkabeeparolus: Thank you for the review! 😊

For my own purpose I'm thinking specifically about the **batman** wrapper, which results in a pretty awesome colorized manpage viewer. You should try it if you haven't! ☺️

I've used *abbr man batman* to handle regular manpages, but obviously that does not work with the builtin manpages (status get-file) or with the Alt-h key binding.

However, if you believe this is something fish should not touch, and that I should solve it at the system level, I guess I could simply put a bash wrapper script in ~/bin at the top of my $PATH. 😊

krobelus: ah I just saw https://github.com/fish-shell/fish-shell/issues/11953
which references https://github.com/eth-p/bat-extras/blob/master/src/batman.sh

From a first glance, it looks like that script is running 
`command man` with custom `MANPAGER`.

So you could either
1. achieve the same thing by setting `MANPAGER=bat` etc. yourself (
2. do something like this, which is likely easier than finding about this new config option.

```fish
functions --copy man fish-man
function man
	set -l tmp (mktemp -d)
	ln -s (which batman) $tmp/man
	fish-man $argv
	set -l saved_status $status
	rm -r tmp
	return $status
end
```

In general, changing the behavior of an existing command can backfire (break scripts).
Maybe give it a different name?
If it's about muscle memory, we have abbreviations for that

	abbr man batman

krobelus: > with the Alt-h key binding.

hmm it's unfortunate that you have to override two different things.

If you want to change the behavior of `alt-h` but don't want to create `~/bin/man` (batman in disguise),
then I suppose the first step would be to "un-dunder" it, calling it `fish_print_help`,
adding a documentation page, and then we can add something like `fish_print_help --man=batman`.
That would definitely be welcome.

I'm more careful about adding things to `man` because that's not the place for fish to add its own inventions
(but we still have some things like this, for historical reasons)

So what if you could write

```fish
# TODO better name. Could even choose man I guess.
function my-batman -a item
	fish-print-help --man-program=batman --command=$item
end
abbr man my-batman
bind alt-h '
	fish-print-help --man-program=batman --from-commandline
'
```

I'm not yet sure if that's the ideal interface
(maybe we can separate the "get the command token"
and "actually call «man» with the appropriate args"
into two documented functions.

harkabeeparolus: Thanks for the suggestions! That would be a much better design. I'm relatively new to fish (recovering bash/zsh expert...), so I'm not sure I'm well versed enough in fish to be able to submit a PR. Let me think a little more about this. 😄

harkabeeparolus:
I'm very thankful for the valuable feedback from @krobelus. 😊

I'll go back to the drawing board!

share/functions/man.fish
krobelus: I guess "set -g fish_man_command batman"
might not be the worst option (if "set -gx MANPAGER ..." does not work for some reason).
The only thing I don't like about it is that it's fish-specific,
but I guess we have inflicted that upon ourselves the moment we provided a fish-specific "man" wrapper.

And setting `fish_man_command` is easier than defining both a wrapper function as well as the `alt-h` binding.

but actually - I just installed `bat-extras` and looked at `man batman`;
which tells me to run `batman --export-env=fish | source`

AFAICT that does exactly what you want purely by using `MANPAGER` and similar options.

Assuming that's true, I'd prefer people use that instead of us adding a new option.
If there are a lot of different man programs, or there's a strong reason why that can't work, we can reconsider

harkabeeparolus:
I'll close this for now.

share/functions/man.fish
harkabeeparolus: Yes indeed! That option was added last Summer, and I never saw the update. 😁

It works perfectly. 🤩

I'm still very grateful that you took the time to assist me. 😊

PR #11958, opened 2025-10-16
man: fix variable assignment syntax

danielrainer:
The leading `$` should not be there. They were introduced in 00639921955454255269fc9127a622697035a5ca, which claims to match __fish_print_help, but in that file, the variable assignment is done correctly.

Fixes #11655


danielrainer:
Oh, there is already a PR to fix this, #11956.

PR #11960, opened 2025-10-16
build_tools/check.sh: check for Rust version updates; bump MSRV

krobelus:
Opt-in because we don't want failures not related to local changes,
see d93fc5eded1 (Revert "build_tools/check.sh: check that stable rust
is up-to-date", 2025-08-10).

For the same reason, it's not currently checked by CI, though we
could definitely have "renovatebot" do this in future.

Also upgrade our MSRV to 1.85.

PR #11961, opened 2025-10-16
msrv: update to Rust 1.85

danielrainer:
Update our MSRV to Rust 1.85.

Includes fixes for lints which were previously suppressed due to them relying on features added after Rust 1.70.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -23,6 +23,8 @@ For distributors
 - Fixed build on MIPS machines (:issue:`11965`).
 - Fixed broken universal variables on Cygwin (:issue:`11948`).
 
+The minimum supported Rust version (MSRV) has been updated to 1.85.
+
 
 fish 4.1.3 (released ???)
 =========================
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,7 +4,7 @@ members = ["crates/*"]
 
 [workspace.package]
 # To build revisions that use Corrosion (those before 2024-01), use CMake 3.19, Rustc 1.78 and Rustup 1.27.
-rust-version = "1.70"
+rust-version = "1.85"
 edition = "2021"
 repository = "https://github.com/fish-shell/fish-shell"
 
@@ -100,9 +100,14 @@ widestring.workspace = true
 portable-atomic.workspace = true
 
 [target.'cfg(windows)'.dependencies]
-rust-embed = { workspace = true, optional = true, features = ["deterministic-timestamps", "debug-embed"] }
+rust-embed = { workspace = true, optional = true, features = [
+    "deterministic-timestamps",
+    "debug-embed",
+] }
 [target.'cfg(not(windows))'.dependencies]
-rust-embed = { workspace = true, optional = true, features = ["deterministic-timestamps"] }
+rust-embed = { workspace = true, optional = true, features = [
+    "deterministic-timestamps",
+] }
 
 [dev-dependencies]
 serial_test.workspace = true
--- a/build_tools/make_macos_pkg.sh
+++ b/build_tools/make_macos_pkg.sh
@@ -24,13 +24,7 @@ SIGN=
 NOTARIZE=
 
 ARM64_DEPLOY_TARGET='MACOSX_DEPLOYMENT_TARGET=11.0'
-X86_64_DEPLOY_TARGET='MACOSX_DEPLOYMENT_TARGET=10.9'
-
-# As of this writing, the most recent Rust release supports macOS back to 10.12.
-# The first supported version of macOS on arm64 is 10.15, so any Rust is fine for arm64.
-# We wish to support back to 10.9 on x86-64; the last version of Rust to support that is
-# version 1.73.0.
-RUST_VERSION_X86_64=1.70.0
+X86_64_DEPLOY_TARGET='MACOSX_DEPLOYMENT_TARGET=10.12'
 
 while getopts "sf:i:p:e:nj:" opt; do
     case $opt in
@@ -65,34 +59,29 @@ OUTPUT_PATH=${FISH_ARTEFACT_PATH:-~/fish_built}
 
 mkdir -p "$PKGDIR/build_x86_64" "$PKGDIR/build_arm64" "$PKGDIR/root" "$PKGDIR/intermediates" "$PKGDIR/dst"
 
+do_cmake() {
+    cmake \
+        -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+        -DCMAKE_EXE_LINKER_FLAGS="-Wl,-ld_classic" \
+        -DCMAKE_OSX_ARCHITECTURES='arm64;x86_64' \
+        -DFISH_USE_SYSTEM_PCRE2=OFF \
+        "$@" \
+        "$SRC_DIR"
+}
+
 # Build and install for arm64.
 # Pass FISH_USE_SYSTEM_PCRE2=OFF because a system PCRE2 on macOS will not be signed by fish,
 # and will probably not be built universal, so the package will fail to validate/run on other systems.
 # Note CMAKE_OSX_ARCHITECTURES is still relevant for the Mac app.
 { cd "$PKGDIR/build_arm64" \
-    && cmake \
-        -DCMAKE_BUILD_TYPE=RelWithDebInfo \
-        -DCMAKE_EXE_LINKER_FLAGS="-Wl,-ld_classic" \
-        -DRust_CARGO_TARGET=aarch64-apple-darwin \
-        -DCMAKE_OSX_ARCHITECTURES='arm64;x86_64' \
-        -DFISH_USE_SYSTEM_PCRE2=OFF \
-        "$SRC_DIR" \
+    && do_cmake -DRust_CARGO_TARGET=aarch64-apple-darwin \
     && env $ARM64_DEPLOY_TARGET make VERBOSE=1 -j 12 \
     && env DESTDIR="$PKGDIR/root/" $ARM64_DEPLOY_TARGET make install;
 }
 
 # Build for x86-64 but do not install; instead we will make some fat binaries inside the root.
-# Set RUST_VERSION_X86_64 to the last version of Rust that supports macOS 10.9.
 { cd "$PKGDIR/build_x86_64" \
-    && cmake \
-        -DCMAKE_BUILD_TYPE=RelWithDebInfo \
-        -DCMAKE_EXE_LINKER_FLAGS="-Wl,-ld_classic" \
-        -DRust_TOOLCHAIN="$RUST_VERSION_X86_64" \
-        -DRust_CARGO_TARGET=x86_64-apple-darwin \
-        -DRust_COMPILER="$(rustup +$RUST_VERSION_X86_64 which rustc)" \
-        -DRust_CARGO="$(rustup +$RUST_VERSION_X86_64 which cargo)" \
-        -DCMAKE_OSX_ARCHITECTURES='arm64;x86_64' \
-        -DFISH_USE_SYSTEM_PCRE2=OFF "$SRC_DIR" \
+    && do_cmake -DRust_CARGO_TARGET=x86_64-apple-darwin \
     && env $X86_64_DEPLOY_TARGET make VERBOSE=1 -j 12; }
 
 # Fatten them up.
--- a/src/builtins/fg.rs
+++ b/src/builtins/fg.rs
@@ -72,7 +72,7 @@ pub fn fg(parser: &Parser, streams: &mut IoStreams, argv: &mut [&wstr]) -> Built
             Ok(pid) => {
                 let j = parser.job_get_with_index_from_pid(pid);
                 if j.as_ref()
-                    .map_or(true, |(_pos, j)| !j.is_constructed() || j.is_completed())
+                    .is_none_or(|(_pos, j)| !j.is_constructed() || j.is_completed())
                 {
                     streams
                         .err
--- a/.github/actions/rust-toolchain/action.yml
+++ b/.github/actions/rust-toolchain/action.yml
@@ -26,7 +26,7 @@ runs:
         toolchain=$(
           case "$toolchain_channel" in
             (stable) echo 1.90 ;;
-            (msrv)   echo 1.70 ;;
+            (msrv)   echo 1.85 ;;
             (*)
               printf >&2 "error: unsupported toolchain channel %s" "$toolchain_channel"
               exit 1
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -143,13 +143,9 @@ jobs:
           # Workaround for https://github.com/actions/checkout/issues/882
           ref: ${{ inputs.version }}
       - name: Install Rust
-        uses: ./.github/actions/rust-toolchain@oldest-supported
-        with:
-          targets: x86_64-apple-darwin
-      - name: Install Rust Stable
         uses: ./.github/actions/rust-toolchain@stable
         with:
-          targets: aarch64-apple-darwin
+          targets: aarch64-apple-darwin,x86_64-apple-darwin
       - name: Install dependencies
         run: brew install gettext
       - name: Build and codesign
--- a/README.rst
+++ b/README.rst
@@ -119,7 +119,7 @@ Dependencies
 
 Compiling fish requires:
 
--  Rust (version 1.70 or later)
+-  Rust (version 1.85 or later)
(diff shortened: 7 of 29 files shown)


danielrainer:
I'm working on some follow-up cleanup as well, this only addresses lints raised by our automated checks, not deprecated workarounds for features which were not available with Rust 1.85.

src/builtins/fg.rs
```diff
@@ -72,7 +72,7 @@ pub fn fg(parser: &Parser, streams: &mut IoStreams, argv: &mut [&wstr]) -> Built
             Ok(pid) => {
                 let j = parser.job_get_with_index_from_pid(pid);
                 if j.as_ref()
-                    .map_or(true, |(_pos, j)| !j.is_constructed() || j.is_completed())
+                    .is_none_or(|(_pos, j)| !j.is_constructed() || j.is_completed())
```
krobelus, line 75: there's also some related `TODO(MSRV)`

CHANGELOG.rst
```diff
@@ -20,6 +20,8 @@ Other improvements
 For distributors
 ----------------
 
+The minimum supported Rust version (MSRV) has been updated to 1.85.
```
krobelus, line 26: FWIW going forward I'd probably not bother with announcing the exact number every time, it's probably enough to make it obvious how to find the MSRV that corresponds to a tarball or git commit.

src/builtins/fg.rs
danielrainer: Yeah, I'm focusing on those next.

CHANGELOG.rst
danielrainer: Hm, I guess distributors will notice when they try to build with a toolchain that's too old. I thought it might be nice to give a heads-up, but maybe it's not necessary.

krobelus: it would definitely be nice to packagers (and maybe other users?) to mention it but I don't think it's tragic if we forget or are too lazy.
(Looking for msrv in `.github/` seems more reliable in CHANGELOG.rst but I guess it's subjective).

For "automatic changelog generation" there is https://git-cliff.org/
not sure if it works for our use case

danielrainer: Yes, it just nice to have, but not a big deal if we forget it. For something that changes as infrequently as the MSRV with our current policy, I don't think it's worth bothering with automation. Automation might be nice for other things, but I suspect that it would be more effort to enforce a specific commit message style than to write the changelog manually, especially if we don't want every commit to contribute to the changelog.

danielrainer:
@krobelus The cirrus tasks for Alpine and Ubuntu (jammy) seem to use images under your control. For Ubuntu, we might want to switch to a more recent release. In any case, I can't fix those myself.

Cargo.toml
```diff
@@ -4,7 +4,7 @@ members = ["crates/*"]
 
 [workspace.package]
 # To build revisions that use Corrosion (those before 2024-01), use CMake 3.19, Rustc 1.78 and Rustup 1.27.
-rust-version = "1.70"
+rust-version = "1.85"
```
krobelus, line 7: > The cirrus tasks for Alpine and Ubuntu (jammy) seem to use images under your control.

yeah I did that because I didn't have access.
There's ongoing work to switch them back, see https://github.com/fish-shell/fish-shell/pull/11884 and below

 > For Ubuntu, we might want to switch to a more recent release.

Yeah actually looking at either of
- https://github.com/fish-shell/fish-shell/pull/11679/files
- https://packages.ubuntu.com/search?keywords=rustc&searchon=all&suite=all&section=all

it seems that Ubuntu is lagging a bit behind in rustc versions.

If it helps users, we could certainly lower the Rust version to say, 1.80 or 1.84.
But of course the main goal for now is to get any progressive MSRV going, we can always tweak it later.
No strong opinion.
 
Since Ubuntu Jammy (LTS) is supported until 2027,
we might want to keep running tests:

```
set -eu
today=$(date --iso-8601)
# fetch (or pipe JSON into this script). Example uses endoflife.date API.
curl -fsSL https://endoflife.date/api/ubuntu.json |
  jq -r --arg today "$today" '
    (if type=="object" then [.] else . end)
    | .[]
    | {
        codename: (.codename? // .name // .cycle // "-"),
        version: (.cycle? // .version? // .latest // "-"),
        eol_raw: (.eol? // .eol_date? // .support? // .end_of_life? // .support_end? // null)
      }
    | . as $it
    | ($it.eol_raw // "") as $e
    | ($e | tostring | sub("Z$";"")) as $e_str
    | select($e_str != "" and $e_str >= $today)
    | "\($it.codename)\t\($it.version)\t(\($e_str))"
  ' | sort -r | column -t -s '	'
Questing Quokka  25.10  (2026-07-01)
Plucky Puffin    25.04  (2026-01-17)
Noble Numbat     24.04  (2029-04-25)
Jammy Jellyfish  22.04  (2027-04-01)
```

so for Cirrus we can either
1. keep Jammy but install rust from `curl rustup.sh` (which is probably what our GitHub Actions CI does)
2. upgrade from Jammy to Noble which ships rustup, then use rustup
3. abandon Ubuntu completely (except for GitHub Actions, where we already use `dtolnay/rust-toolchain@$MSRV`) and use the more up-to-date debian stable

If I find time to work on this, I'll add a script to https://github.com/fish-shell/fish-shell/pull/11960
to check if our .cirrus.yml versions are still up-to-date.
That would be desirable if we do option 1;
it would be trivial if we use option 3,
and it might not be clear what's the Ubuntu version policy if we do option 2, which speaks against option 2.

I think option 1 sounds like the most foolproof and fastest to get going for now.

 > In any case, I can't fix those myself.

Documentation is lacking; there's three possibilities:
1. the old approach: build the images in `docker/` and use `docker push` to your own registry and update .cirrus.yml to point to those
2. Run `.github/workflows/build_docker_images.yml` on your fork (either by pushing to master or with `gh --repo $gh_user/fish-shell workflow run --ref=...`).
   This will also create the container images in your namespace
3. hopefully the standard approach in future:
   It should be handled completely transparently by `.github/workflows/build_docker_images.yml`.
   Today that already happens on pushes to `master`, but can also do
   `gh workflow run --ref=...` (you should have access now as org member)

   There are two potential problems with this:
   1. it runs only on master so PRs are not tested until they are merged (https://github.com/fish-shell/fish-shell/pull/11884#issuecomment-3371258962)
   2. even when pushing directly to master, there is a CI orchestration problem: `test.yml` runs concurrently with `build_docker_images.yml` even though it depends on it. See https://github.com/fish-shell/fish-shell/pull/11884#discussion_r2423344925

   I'm not yet sure if these are blockers, which is why we haven't [switched to `fish-shell/fish-ci`](https://github.com/fish-shell/fish-shell/pull/11884/files) yet

Sorry for all the historical baggage.
But I'd definitely trade a little velocity for the opportunity to crack down hard on all the short-sighted engineering practices of the past,
especially as the PR review queue keeps growing.

krobelus: I should be able to update Cirrus images for alpine/ubuntu today (manual update though)

krobelus: I'll wait to merge until tomorrow morning (in case we want to avoid problems with OBS builds),
but going forward, green github CI will be enough.

danielrainer: > I'd definitely trade a little velocity for the opportunity to crack down hard on all the short-sighted engineering practices of the past,
especially as the PR review queue keeps growing.

Yeah, cleaning up some of that to simplify development in the future is certainly desirable. It's not like we are in a big rush to get this merged.

I think the only real reason not to use rustup everywhere is that making network requests during package builds is frowned upon. As I understand it that's also the primary reason why we can't just set our MSRV to stable. But our CI checks don't have to worry about that, they can use rustup as we please. So we only need to ensure that our MSRV is supported on the distros which package Rust, possibly only the ones where we offer our own builds. I think @zanchey has managed that, so I think we should confirm that this PR does not cause problems with our own builds (https://build.opensuse.org/package/show/shells:fish/fish, https://launchpad.net/~fish-shell/+archive/ubuntu/release-4)

> for Cirrus we can either

I think options 1 and 2 are compatible. If we ensure that rustup is installed in the images, we could even keep the matrix approach. There is also the `linux_arm_task` which seems to have been disabled for fish's Rust rewrite. Maybe we can reactivate that now and simplify to a single matrix task for Linux.

> abandon Ubuntu completely

I don't know what the motivation for adding Ubuntu to Cirrus was in the first place. It doesn't seem to do anything not already done in gh CI, other than testing on different distro versions, which can be valuable. If we explicitly base our version policies around Debian, it certainly wouldn't hurt to actually test on Debian. I don't have much of an opinion on whether we should keep testing on Ubuntu. If we can do so without much effort on our part or exhausting build resources, I guess it's nice to keep it around, but I also wouldn't mind abandoning it.

> If it helps users, we could certainly lower the Rust version to say, 1.80 or 1.84.

That's an option. I guess it again comes down to which distros and versions we want to support without requiring rustup. "whatever's in Debian stable" is fairly simple, but maybe a more complex policy is more appropriate for us. The main benefit of Rust 1.85 is that it introduces the 2024 edition. Other than that, I think Rust 1.83 is the oldest version which probably still works with all the changes in #11964 (we don't have to merge all of those now, of course). To make a reasonably well-informed decision, we might want to wait to see if zanchey has any comments.

build_tools/make_macos_pkg.sh
```diff
@@ -30,7 +30,7 @@ X86_64_DEPLOY_TARGET='MACOSX_DEPLOYMENT_TARGET=10.9'
 # The first supported version of macOS on arm64 is 10.15, so any Rust is fine for arm64.
 # We wish to support back to 10.9 on x86-64; the last version of Rust to support that is
 # version 1.73.0.
-RUST_VERSION_X86_64=1.70.0
+RUST_VERSION_X86_64=1.85.0
```
krobelus: Probably Debian/1.85 is fine, but if it helps anyone, we might also
do something like 1.83 (Debian-2).

In any case, now that we no longer support really old macOS,
we should probably use stable here, like for other builds:

```diff
diff --git a/.github/workflows/release.yml b/.github/workflows/release.yml
index f97e35cdfd..c1d139c047 100644
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -143,13 +143,9 @@
           # Workaround for https://github.com/actions/checkout/issues/882
           ref: ${{ inputs.version }}
       - name: Install Rust
-        uses: ./.github/actions/rust-toolchain@oldest-supported
-        with:
-          targets: x86_64-apple-darwin
-      - name: Install Rust Stable
         uses: ./.github/actions/rust-toolchain@stable
         with:
-          targets: aarch64-apple-darwin
+          targets: aarch64-apple-darwin,x86_64-apple-darwin
       - name: Install dependencies
         run: brew install gettext
       - name: Build and codesign
diff --git a/build_tools/make_macos_pkg.sh b/build_tools/make_macos_pkg.sh
index 78bfe044a8..970a2ae3db 100755
--- a/build_tools/make_macos_pkg.sh
+++ b/build_tools/make_macos_pkg.sh
@@ -26,12 +26,6 @@
 ARM64_DEPLOY_TARGET='MACOSX_DEPLOYMENT_TARGET=11.0'
 X86_64_DEPLOY_TARGET='MACOSX_DEPLOYMENT_TARGET=10.9'
 
-# As of this writing, the most recent Rust release supports macOS back to 10.12.
-# The first supported version of macOS on arm64 is 10.15, so any Rust is fine for arm64.
-# We wish to support back to 10.9 on x86-64; the last version of Rust to support that is
-# version 1.73.0.
-RUST_VERSION_X86_64=1.85.0
-
 while getopts "sf:i:p:e:nj:" opt; do
     case $opt in
         s) SIGN=1;;
@@ -65,34 +59,29 @@
 
 mkdir -p "$PKGDIR/build_x86_64" "$PKGDIR/build_arm64" "$PKGDIR/root" "$PKGDIR/intermediates" "$PKGDIR/dst"
 
+do_cmake() {
+    cmake \
+        -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+        -DCMAKE_EXE_LINKER_FLAGS="-Wl,-ld_classic" \
+        -DCMAKE_OSX_ARCHITECTURES='arm64;x86_64' \
+        -DFISH_USE_SYSTEM_PCRE2=OFF \
+        "$@" \
+        "$SRC_DIR"
+}
+
 # Build and install for arm64.
 # Pass FISH_USE_SYSTEM_PCRE2=OFF because a system PCRE2 on macOS will not be signed by fish,
 # and will probably not be built universal, so the package will fail to validate/run on other systems.
 # Note CMAKE_OSX_ARCHITECTURES is still relevant for the Mac app.
 { cd "$PKGDIR/build_arm64" \
-    && cmake \
-        -DCMAKE_BUILD_TYPE=RelWithDebInfo \
-        -DCMAKE_EXE_LINKER_FLAGS="-Wl,-ld_classic" \
-        -DRust_CARGO_TARGET=aarch64-apple-darwin \
-        -DCMAKE_OSX_ARCHITECTURES='arm64;x86_64' \
-        -DFISH_USE_SYSTEM_PCRE2=OFF \
-        "$SRC_DIR" \
+    && do_cmake -DRust_CARGO_TARGET=aarch64-apple-darwin
     && env $ARM64_DEPLOY_TARGET make VERBOSE=1 -j 12 \
     && env DESTDIR="$PKGDIR/root/" $ARM64_DEPLOY_TARGET make install;
 }
 
 # Build for x86-64 but do not install; instead we will make some fat binaries inside the root.
-# Set RUST_VERSION_X86_64 to the last version of Rust that supports macOS 10.9.
 { cd "$PKGDIR/build_x86_64" \
-    && cmake \
-        -DCMAKE_BUILD_TYPE=RelWithDebInfo \
-        -DCMAKE_EXE_LINKER_FLAGS="-Wl,-ld_classic" \
-        -DRust_TOOLCHAIN="$RUST_VERSION_X86_64" \
-        -DRust_CARGO_TARGET=x86_64-apple-darwin \
-        -DRust_COMPILER="$(rustup +$RUST_VERSION_X86_64 which rustc)" \
-        -DRust_CARGO="$(rustup +$RUST_VERSION_X86_64 which cargo)" \
-        -DCMAKE_OSX_ARCHITECTURES='arm64;x86_64' \
-        -DFISH_USE_SYSTEM_PCRE2=OFF "$SRC_DIR" \
+    && do_cmake -DRust_CARGO_TARGET=x86_64-apple-darwin \
     && env $X86_64_DEPLOY_TARGET make VERBOSE=1 -j 12; }
 
 # Fatten them up.
```

danielrainer: > X86_64_DEPLOY_TARGET='MACOSX_DEPLOYMENT_TARGET=10.9'

Should this be changed to 10.12? Or also 11.0? The old comment claims that arm64 is supported since 10.15, so I'm not sure why 11.0 is used there. But I'm also generally unfamiliar with packaging for macOS.

danielrainer:
Rust 1.85 seems to be packaged on all supported versions of Ubuntu, but not always as the default `rustc`, so by changing the packages we install, we should be able to make Cirrus builds on jammy work with Rust 1.85.

- https://packages.ubuntu.com/jammy/rustc-1.85
- https://packages.ubuntu.com/jammy/cargo-1.85
- https://packages.ubuntu.com/jammy/rust-1.85-all

build_tools/make_macos_pkg.sh
krobelus: right, it should be 10.12. I also don't know why this arm64 deployment target is used

krobelus:
> Rust 1.85 seems to be packaged on all supported versions of Ubuntu

Nice.
Unless we find a reason to use the distro package, I'd probably go with rustup because that doesn't need as much maintenance
https://github.com/fish-shell/fish-shell/pull/11960/commits/746833f225e9d58adc35f6a7803b0e27c925f17d#diff-ecb11e11e6c26808b49470413fa4e79d561ecc0f855c8f550ad1eb09b69f4e2bR1
Though it won't matter once we have renovatebot set up

build_tools/make_macos_pkg.sh
danielrainer: I added the diff and the change to 10.12 to the PR.

danielrainer:
> I'd probably go with rustup because that doesn't need as much maintenance

Yes, that's definitely the better approach when network access during builds is possible.

The alpine image seems to work now, but jammy uses Rust 1.75. Should I rebase this PR onto some changes which make use of rustup in the Cirrus tasks to install a certain Rust version (probably our MSRV or stable)?

CHANGELOG.rst
zanchey: I think it's worth including.

PR #11964, opened 2025-10-17
Resolve MSRV gated TODOs

danielrainer:
Addresses several code style issues which have been blocked on a MSRV update. This includes the things I found with some quick grepping that can be fixed on Rust 1.85 without updating the Rust edition to 2024.

There is one thing which I found but have not fixed, since it looks like a fairly technical issue and I haven't looked into it enough to feel confident about addressing it: `src/wutil/dir_iter.rs:275`

Stacked on #11961

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -132,7 +132,6 @@ fn detect_bsd(_: &Target) -> bool {
     if !target.chars().all(|c| c.is_ascii_lowercase()) {
         target = target.to_ascii_lowercase();
     }
-    #[allow(clippy::let_and_return)] // for old clippy
     let is_bsd = target.ends_with("bsd") || target.ends_with("dragonfly");
     #[cfg(any(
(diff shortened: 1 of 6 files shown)


danielrainer:
Do we want to keep empty `src/future.rs` around? It was used to backport things which were only available in Rust versions above our MSRV, but currently there aren't any backports remaining. I find the filename a bit unclear; when I first noticed it I thought it might be about async stuff.

krobelus:
> Do we want to keep empty `src/future.rs` around?

Nah, we can basically always delete unused code because who knows if we ever need it.

> It was used to backport things which were only available in Rust
> versions above our MSRV, but currently there aren't any backports
> remaining. I find the filename a bit unclear; when I first noticed
> it I thought it might be about async stuff.

yeah I realized later other projects call this "stdx"

PR #11726, opened 2025-08-15
Switch to builtin gettext implementation (build-time parsing)

danielrainer:
## Description

Handle translation from PO to MO files in `build.rs`.

Parse MO files at build time and generate source code containing the
localization mappings.

Replace runtime dependency on gettext by custom handling of language selection,
and using compile-time generated maps for localization.

Choose which languages to embed localization data for via cargo features.
Disable most gettext code aside from stubs to keep the code working when no
languages are enabled.

Advantages of this approach:
- We have no runtime dependency on gettext anymore.
- The implementation has the same behavior everywhere.
- Our implementation is significantly simpler than GNU gettext.
- We can have localization in cargo-only builds by embedding localizations into
  the code.
  Previously, localization in such builds only worked as long as the binary was
  not moved from the build directory.
- Generating the MO files from `build.rs` removes the need for scripting it in
  several places (cmake, checks).
- MO file generation, parsing, and embedding happens automatically when building
  fish, eliminating the need to run anything aside from rebuilding fish.
- Reduced overhead when disabling localization, both for compilation time and
  binary size. At the moment, a release build with all features disabled results
  in a binary size of 4.6 MB, whereas when enabling only the
  `localize-all-languages` feature, the resulting size is 9.5 MB.

Disadvantages of this approach:
- Our own runtime implementation of gettext needs to be maintained.
- The implementation has a more limited feature set (but I don't think it lacks
  any features which have been in use by fish).

## TODOs:
- [x] Changes to fish usage are reflected in user documentation/manpages.
- [x] Tests have been added for regressions fixed
- [ ] User-visible changes noted in CHANGELOG.rst

PR #11967, opened 2025-10-17
l10n: Add zh-TW translation

lumynou5:
## Description

Adds human-translated tier-1 strings.

## TODOs:


lumynou5:
I'm not sure which section of changelog it should be under.

krobelus:
> I'm not sure which section of changelog it should be under.

We've put similar ones in "Other improvements", or maybe "Notable improvements"

lumynou5:
I put it in "Notable Improvements"; I found adding new languages was put there and old "Significant Changes" while updating translation is "Other".

krobelus:
> +msgstr "%s 在過了 %d 秒後仍無法讀取到查詢主要裝置特性的回應。這通常是由於你的終端機缺少功能，參見「help terminal-compatibility」或「man fish-terminal-compatibility」。此 %s 行程不再會等待未決的查詢，一些非必要的功能將因此停用。"

FWIW we treat "Primary Device Attribute" as a proper noun, which is why we don't translate it for zh_CN.
No big deal, both variants should work, and this is a rare error.

> +# #-#-#-#-#  -  #-#-#-#-#

(stray comment)

krobelus:
Great job. In future, we should do some more work to classify/prioritize translatable messages, and probably mention that in future changelogs

FWIW now that we have multiple Chinese translation sources, `LANG=zh` is ambiguous.

We currently pick whatever is first in  `CATALOGS.entries()`, which says that "Entries are returned in an arbitrary but fixed order.".

So we're probably deterministic, but the behavior could change across versions

```
$ for LC_MESSAGES in en zh zh_CN zh_TW zh_SG
      set_color --underline=asdf
  end
set_color: invalid underline style: asdf
set_color: 无效的下划线样式：asdf
set_color: 无效的下划线样式：asdf
set_color：無效的底線樣式：asdf
set_color: invalid underline style: asdf
```

So we should probably either sort these ourselves, or make `LANG=zh` match neither.
(cc @danielrainer. I think either is probably fine;
I haven't looked up the discussion where we decided to postpone this)

lumynou5:
I think leaving the unreliable behavior of `LANG=zh` is also fine; Chinese language(s) users (almost) always specify the region due to the big difference between variants.

BTW if we make `zh` match nothing, won't the behavior be weird?  For example, if there are `en-US` and `en-UK` but user specifies `en`, which turns out to be unusable.

danielrainer:
I share @lumynou5's view on the matter. Falling back to an arbitrary variant of the language when the variant is not specified allows users to discover that localization is available. Then, they can check if other variants are available if they don't like the one which is selected automatically.

Our semantics in this regard differ from `gettext`'s, which uses a fixed default variant for languages without the variant specified. We intentionally don't do this, to avoid getting into pointless arguments about which variant should be the default. Users with specific language preferences can encode them in a way that is compatible with both approaches in the `LANGUAGE` variable: List languages with explicit variants first, optionally followed by the variant-less language identifier, if they prefer arbitrary variants of the language over the next fallback option. E.g. `set LANGUAGE zh_TW zh` for users who want `zh_TW`, but still prefer other variants of Chinese over English, whereas users who prefer English over other variants of Chinese would specify `set LANGUAGE zh_TW`. Note that the former does not result in all available variants of `zh` being tried before falling back to the next option, neither in GNU `gettext`, nor in our own implementation. So far this has not been relevant for us, since there was at most one variant of each language, but now we might consider changing fish's behavior to expand a variant-less identifier into all available variants (in arbitrary order).

danielrainer:
So basically, given our current set of languages, `LANGUAGE=zh` would expand into a precedence list, either `zh_TW:zh_CN` or `zh_CN:zh_TW` (we still won't specify the order). This might result in duplicate entries in the precedence list, e.g. `LANGUAGE=zh_TW:zh` expanding into `zh_TW:zh_CN:zh_TW`, but we could filter that to improve lookup performance by only keeping the first appearance of each language.

Berrysoft:
Although it seems to be a bit late to follow this discussion, I would like to suggest the strategy of ICU. The variants of `zh` is mainly 2 parts, `zh_Hans` and `zh_Hant`, and the fallback order is
* Fallback to `zh_Hant`: `zh_TW`, `zh_HK`, `zh_MO`.
* Others fallback to `zh_Hans` (including `zh_CN`, `zh_SG`, etc.).

And that's why usually there are only `zh_CN` (the full name is `zh_Hans_CN`) and `zh_TW` (the full name is `zh_Hant_TW`).

As for `zh`, nearly no one actually uses it. ICU expands it to `zh-Hans-CN`, but it's OK to ignore that.

I have written a crate [language-matcher](https://crates.io/crates/language-matcher) for this purpose. It calculates the "distance" of 2 languages and finds the best match for a specific language among existing ones.

danielrainer:
Thank you for providing this context. Do you have a concrete proposal for how fish should handle language fallback?

I think our (soft) requirement is that our approach should be mostly compatible with GNU gettext configuration (via `LANG`, `LC_MESSAGES`, `LC_ALL`, `LANGUAGE`), and I think where we differ, we prefer showing translations for a different variant of a language over showing no translations at all. Another thing I consider important is that users have a way of enabling only specific variants of a language (e.g. if someone wants `zh_TW` but prefers English over `zh_CN`, this should be possible to configure). At the moment we achieve this by letting `ll_CC` only fall back to the `ll.po` catalog, but not some other `ll_DD.po` catalog. So fallback ordering only really happens if users specify `ll` (without a region). At the moment, the order is arbitrary. Your crate seems to enable sorting languages by distance to some fixed language, but if we only have `ll` without a region, there isn't really a fixed language from which distances can be computed. We could do something similar to gettext, where we have a "default" variant of each language and use that for distance computation. The reasons we don't do that are that we don't want to bother with specifying this for all relevant languages, and deal with potential resulting disagreements. In https://www.unicode.org/reports/tr35/tr35.html#EnhancedLanguageMatching it looks like `paradigmLocales` might provide a sort of default variant of languages. But overall, I'm not sure we have enough reason to with improving fallback behavior for `ll`. The main intention of the current implementation is letting users know that localization is available, and if they are unhappy with the defaults they should replace `ll` by a more precise specification of what they want in `LANGUAGE`.

One thing your crate might be useful for is providing fallback from e.g. `zh_Hant` to `zh_TW`. At the moment, we assume that language identifiers have the form `ll` or `ll_CC` (or `lll`/`lll_CC` for languages with a 3-letter identifier). This follows GNU gettext. Supporting something like `zh_Hant` or `zh_Hant_TW` might require some code changes. It definitely will if we start using this syntax for file names. I'm also not sure how we can reasonably combine such fallback behavior while still allowing users to opt our of certain variants.

krobelus:
> * Fallback to `zh_Hant`: `zh_TW`, `zh_HK`, `zh_MO`.
> * Others fallback to `zh_Hans` (including `zh_CN`, `zh_SG`, etc.).

sounds good, especially if we can outsource this to a reusable crate.
I wonder how other projects (browsers etc.) solve this problem.

> I'm not sure we have enough reason to with improving fallback behavior for `ll`.

Yeah that case (`LANG=zh`) is a case of "GIGO" where it's fine to keep the current behavior or do whatever.

> Another thing I consider important is that users have a way of enabling only specific variants of a language (e.g. if someone wants `zh_TW` but prefers English over `zh_CN`, this should be possible to configure).

Ok so this means that assuming for `LANG=zh_HK _ foo`, there is no `zh_Hant_TW` localization is available,
then we should not fall back to using this message's `zh_Hans_CN` localization.

But independent of that, we could still use the ICU approach for selecting the catalogues (one for each item in `LANGUAGE`).
That might be an overall improvement.

> At the moment we achieve this by letting `ll_CC` only fall back to the `ll.po` catalog, but not some other `ll_DD.po` catalog.

If we adopt `language-matcher`, we can probably rename `de.po` to `de_DE.po` without losing translations for `LANG=de_AT`.

> Supporting something like `zh_Hant` or `zh_Hant_TW` might require some code changes. It definitely will if we start using this syntax for file names.

Right. Maybe we that can be an independent/later change.

> I'm also not sure how we can reasonably combine such fallback behavior while still allowing users to opt our of certain variants.

I see.
AIUI, this might be the only significant downside of adopting `language-matcher`.
So this is for the case where a `LANG=zh_HK` user does not want `zh_TW.po` (i.e. they are disagreeing with CLDR).
I guess that after discovering that fish doesn't have `zh_HK.po`, this user could achieve the desired behavior with `LC_MESSAGES=en_US`.
(And as mentioned above, `zh_HK.po` is usually missing in favor of `zh_TW.po`, so this is not too surprising)
Again, I'm wondering if there's a precedent.

I'm optimistic about `language-matcher` but I'm not an expert.
We might want to think a bit more about whether making it harder to opt out of certain variants is worth it.
Ideally we end up with a behavior that either matches other existing apps (modulo our historical gettext shims?),
or one that future apps will want to copy from us.

danielrainer:
> we could still use the ICU approach for selecting the catalogues (one for each item in `LANGUAGE`)

As I understand it, given an entry in `LANGUAGE` that specifies a language variant, we could get distance values to each language we have a catalog for (which would also require specifying the exact language variant of our catalogs). Then we could use the language with the lowest distance. I'm not sure if this is guaranteed to be a single language or if there could be more than one closest language variant. With our current catalogs, this probably wouldn't be an issue. We would also have to decide on a cutoff distance at which languages are considered too distant to fall back from one to the other.

One option might be setting the cutoff distance to 0, meaning we fall back between synonym language identifiers, but not between variants which actually differ in some way.

> I guess that after discovering that fish doesn't have `zh_HK.po`, this user could achieve the desired behavior with `LC_MESSAGES=en_US`.

If we are fine with users having to set fish-specific variable values this could be a decent approach. Setting `LC_MESSAGES=en_US` wouldn't work if `LANGUAGE` is set as well. `set -g LANGUAGE en_US` or `set -g LC_MESSAGES=C.UTF-8` in the fish config would work, except for messages printed before fish reads this config. However, this would also modify the respective variables for processes spawned by fish, which is undesirable. I don't know of a way to set a variable only for fish itself, while keeping the original value intact for child processes. If there is no way to do so, we could work around it by introducing `FISH_LANGUAGE`, which replaces `LANGUAGE` when set, but is treated the same as `LANGUAGE`. If we want to, we could print warnings when `FISH_LANGUAGE` contains some `ll` without a variant specified. We could be even stricter and require that each entry exactly matches a catalog we have available, and possibly also that there is nothing listed after `en_US`. It might be nice to allow users to specify variants which do not exist yet, so that they can automatically start using a catalog when it is added to fish, without having to endure warnings before that.

For this to be a good approach, we should guarantee that any `ll_CC` value deterministically expands to at most one catalog, and for any catalog we have, there is some `ll_CC` value which expands to it. The latter requirement should be easy, but for the former we would need a guarantee from `language-matcher` that there are no ties for the closest language (below our cutoff threshold), or if there is, we there is a consistent tie-breaking mechanism. I guess in such cases we could also stick with the current approach for `ll` and just pick an arbitrary catalog from the closest languages, or pick all of them in arbitrary order. If we are fine with arbitrary choices, I think this could be implemented without too much effort. Note that if our set of catalogs changes, I think it's fine if the catalog some `ll_CC` maps to changes.

Examples illustrating this approach, assuming we have the following catalogs (with `en_US` being the default, implicitly always last in the fallback order) `{de_DE,en_US,zh_CN,zh_TW}`:
1. user specifies `LANGUAGE=zh_CN`, resulting fallback order `zh_CN`
2. `LANGUAGE=zh_HK` -> `zh_TW`
3. `LANGUAGE=zh_HK:zh_TW` -> `zh_TW`
4. `LANGUAGE=zh_SG:zh_TW` -> `zh_CN:zh_TW`
5. `LANGUAGE=zh_SG:de_DE` -> `zh_CN:de_DE`
6. `LANGUAGE=de_DE:en_US:zh_CN` -> `de_DE`

If we want to keep discoverability high, we could also expand `LANGUAGE=ll_CC` to all all catalogs we have for `ll`, sorted by distance from the specified variant (arbitrary order on ties). This would maximize the chance users notice that fish supports localization, and if they are unhappy with how it works they can check our docs and use `FISH_LANGUAGE` to specify their preferences.

If some other project has a good approach for this it might be a good idea to be compatible with that, which could avoid having to specify application-specific variables, though it might still be good to offer them as an override option, allowing users to adapt their config to the perceived qualities of the catalogs of specific applications.

In any case, if we want some `language-matcher`-based approach, we will have to start by specifying language variants in our catalog names. A while ago, I suggested which variants to use, visible e.g. in https://github.com/fish-shell/fish-shell/pull/11726#discussion_r2302227056. In our tooling, I think we can enforce `lll?_CC` syntax, or are there some identifiers which we might realistically want for catalog names that don't match this?

lumynou5:
> I think where we differ, we prefer showing translations for a different variant of a language over showing no translations at all.

(this is a bit late reply)

Personally I prefer that an unsupported `ll_CC` fallbacks to `ll` (if exists) instead of `ll_DD` because the latter might be confusing to user and they thus couldn't figure out the missing localization.

I think this is easier to configure fallbacks and GNU gettext compatible.  e.g. if a HongKong user wants `zh_TW` as a fallback due to the same script, they can set it to `zh_HK:zh_TW`; otherwise they can set it to only `zh_HK` to prevent fallback to other variants.  This way the list doesn't have any implicit entry no matter which languages are in the list.

danielrainer:
> Personally I prefer that an unsupported `ll_CC` fallbacks to `ll` (if exists) instead of `ll_DD` because the latter might be confusing to user and they thus couldn't figure out the missing localization.

Yes, that's what's currently implemented, and I think it's a perfectly valid approach. The only real downside is that users who set `ll_CC` who would be fine with `ll_DD` might not realize that localization is available in a language they prefer over the default English. For some languages, the differences between the variants are fairly small. E.g. someone setting `LANG=de_AT.UTF-8` would most likely be fine with messages from a `de_DE` catalog, since the language variants are similar enough that most messages would be identical. But I would be completely fine with staying on the current approach.

krobelus:
> if a HongKong user wants zh_TW as a fallback due to the same script, they can set it to zh_HK:zh_TW

yeah especially if gettext compatibility is relevant (i.e. for well-known variables like `LANGUAGE`), that seems okay.
I don't know enough about localization so I'm fine with either.

> We would also have to decide on a cutoff distance at which languages are considered too distant to fall back from one to the other.

technically not, but yeah, it would be very weird if an Indian locale falls back to e.g. German instead of English.
This might be the only difficult part.

> One option might be setting the cutoff distance to 0, meaning we fall
> back between synonym language identifiers, but not between variants which
> actually differ in some way.

that sounds like a safe thing to start with,
but I guess it wouldn't address the motivating issues (zh_HK, zh_MO)?

The number of relevant pairs of specified language and selected localization is small.
I think it only applies to zh_*, and there should not be tons of test cases;
i.e. we can add test cases for all of them.

But we'd also need to make sure the cutoff logic doesn't break other languages.

> I don't know of a way to set a variable only for fish itself, while keeping the original value intact for child processes.

yeah this probably shouldn't be a feature.
In general, Unix programs are expected to read environment variables only at startup.
fish is special in that it reacts to online changes because it allows setting environment variables.

But we should probably add "$fish_language" than making this more complicated.

Actually, if we're thinking about warnings etc, we could make it a proper subcommand like "status locale {get,set}".
It's not clear whether `set LANGUAGE foo2` should override results of `status locale set foo1`,
but I think either decision would be fine since it's a rare edge case.

> If we are fine with arbitrary choices,

I think so, assuming it wouldn't really make anything worse relative to what we have today.

> LANGUAGE=ll_CC to all all catalogs we have for ll

I guess that would conflict with people's expectation that `LANG=zh_CN` falls back to English only. (not sure if that's the predominant and most reasonable expectation)

danielrainer:
> technically not

Deciding to only cut off above the maximum value (100) is one possible decision, but a silly one of course. We probably wouldn't even want to compare languages with different language codes.

> I guess it wouldn't address the motivating issues

Yes, and I'm not sure how much of a difference it would make in practice, so it's not a great option.

> I think it only applies to zh_*, and there should not be tons of test cases;
i.e. we can add test cases for all of them.

That's probably true for the current set of languages, but it would be annoying if we add more languages later on and notice that an approach that works for Chinese does not work for the new languages, so testing what we have now does not prevent potential issues in the future.

> `status locale {get,set}`

This might be the best option for fish. I agree that overriding behavior from setting the gettext variables is not particularly important. We probably want to disable localization when `LC_ALL` or `LC_MESSAGES` or `LANG` (considering only the first non-empty one) are set to a C locale. I think `LANGUAGE` should be ignored after `status locale set` has been run.

Then, I think we can keep the current fallback behavior for the gettext variables, to make them behave similar to how GNU gettext would treat them, and for `status locale set` we just don't offer any fallback, other than falling back to the next entry in the list if the first one does not have an associated catalog, or if it does not contain the desired message. For this command, we can have a flag for disabling warnings on language identifiers we don't have catalogs for. I think the warnings should be shown by default.

Does this seem like a reasonable approach to you?

lumynou5:
> That's probably true for the current set of languages, but it would be annoying if we add more languages later on and notice that an approach that works for Chinese does not work for the new languages, so testing what we have now does not prevent potential issues in the future.

Yeah, that's why I think having no special cases is better.  And for synonym language identifiers, e.g. `nb` and `no`, we could suppose users who use Norwegian Bokmål have had `nb:no` set on their machines.

> `status locale {get,set}`

Seems fine but I'm not sure if the potential complexity of code and use is acceptable.

danielrainer:
> I'm not sure if the potential complexity of code and use is acceptable.

Regarding code complexity, we would have to make the code for setting the fallback order slightly more complex to track whether the current precedence was set via the `status locale` builtin, and to differentiate whether a new configuration originates from `status locale` or not. Then, we would also need a way of querying the current precedence, ideally with information about what caused it to be set to the current value

We would also have to write the `status locale` builtin of course, but that should be fairly independent of other code aside from interacting with the API for setting/getting the language precedence list.

Overall, I don't think code complexity would be an issue, especially in comparison to the total complexity required for our localization handling.

Complexity of use also seems fine to me. For most users, their existing environment variables will work without them having to do anything. For the users who are not satisfied with that, using a single command which tells them when they do something wrong is pretty much the minimal amount of complexity one could ask for. I'd argue that setting a variable would be more complex, since you don't get direct feedback on whether it worked the way you intended.

lumynou5:
> I'd argue that setting a variable would be more complex, since you don't get direct feedback on whether it worked the way you intended.

I agree that failing fast would be more friendly than being silent.

My concern is that a command just to set language for Fish would be a bit unnecessary and its interaction with the env variables would be not that friendly as it was designed.

By the discussion above, I suppose the env variables are ignored (unless C locale), so user would still have env variables for commands run.  I think we could extend its functionality by making it export variables if we eventually decided to add it.  As a result, Fish provides an abstract level to set language for the entire environment within the shell at once, and user doesn't need to know e.g. the change of `no` language ID.  Because I didn't see any benefit to set language twice for now.

> The only real downside is that users who set `ll_CC` who would be fine with `ll_DD` might not realize that localization is available in a language they prefer over the default English.

If this is why to have a language identifier match a close enough catalog if the exact one doesn't exist or miss strings, exporting seems reasonable.

danielrainer:
> user would still have env variables for commands run

The point of introducing the `status locale` command is to have a way to configure localization for fish that is independent of the system-wide configuration done via environment variables. Users who want to use the same configuration for fish as for their other programs should just stick to using the established environment variables.

The `status locale` command wouldn't allow for configuring fish in a way that's impossible with environment variables. Instead, it is intended to give fish users a way to override their normal env var config. The reasons to do so would be fish interpreting the variables in a different way than GNU gettext, meaning it's not possible to reasonably set the variables to values which work for both GNU gettext and fish. Another reason might be that users like or dislike certain translations in fish, so their preferences for fish are different from their default preferences. E.g. someone might have translations disabled by default, because they might not be good in some tools, but they like fish's translations for their preferred language, so they would change their settings only for fish. Another example would be someone normally preferring `ll_CC` over `ll_DD`, but they don't like fish's `ll_CC` catalog, so in fish, they'd rather see messages from the `ll_DD` catalog.

For most programs, all of this could reasonably be achieved by starting them with overridden values for the relevant env vars, but because fish would propagate them to its child processes, this is not a good approach for fish, so we need a way that does not use the established env vars.

PR #11975, opened 2025-10-19
encoding: use UTF-8 everywhere

danielrainer:
Assume that UTF-8 is used everywhere. This allows for significant
simplification of encoding-related functionality. We no longer need
branching on single-byte vs. multi-byte locales, and we can get rid of
all the libc calls for encoding and decoding, replacing them with Rust's
built-in functionality, or removing them without replacement in cases
where their functionality is no longer needed.

Several tests are removed from `tests/checks/locale.fish`, since setting
the locale no longer impacts encoding behavior. We might want more
rigorous testing of UTF-8 handling instead.


src/common.rs
```diff
@@ -1102,84 +1088,54 @@ pub static EMPTY_STRING: WString = WString::new();
...
-/// The string may contain embedded nulls.
-///
-/// This function encodes illegal character sequences in a reversible way using the private use
-/// area.
-pub fn str2wcstring(inp: &[u8]) -> WString {
-    if inp.is_empty() {
+/// Encodes the bytes in `input` into a [`WString`], encoding non-UTF-8 bytes into private-use-area
+/// code-points. Bytes which would be parsed into our reserved PUA range are encoded individually,
+/// to allow for correct round-tripping.
+pub fn bytes2wcstring(mut input: &[u8]) -> WString {
```
krobelus, line 1094: Let's add a changelog entry under "Deprecations and removed features".

I'm good with removing support for non-UTF-8 outright,
though we don't have any telemetry to be able to tell if this puts needless pressure on real users.
If we want to be extra safe, we can do a feature flag, which users can turn back off for one release.
But that flag will be removed anyway in a future release (unless we hear about use cases).
Either way is fine by me.

src/common.rs
```diff
@@ -1102,84 +1088,54 @@ pub static EMPTY_STRING: WString = WString::new();
...
-
-        // We have found a non-ASCII character.
-        let mut ret = 0;
-        let mut c = '\0';
 
-        let use_encode_direct = if inp[pos] & 0xF8 == 0xF8 {
-            // Protect against broken mbrtowc() implementations which attempt to encode UTF-8
-            // sequences longer than four bytes (e.g., OS X Snow Leopard).
-            // TODO This check used to be conditionally compiled only on affected platforms.
-            true
```
krobelus, line 1139: Right, the Rust implementation surely doesn't have this bug.

src/common.rs
```diff
@@ -1102,84 +1088,54 @@ pub static EMPTY_STRING: WString = WString::new();
...
-                    // Invalid data.
-                    ret == 0_usize.wrapping_sub(1)
-                    ||
-                    // Other error codes? Terrifying, should never happen.
-                    ret > inp.len() - pos
+    fn append_escaped_str(output: &mut WString, input: &str) {
+        for c in input.chars() {
+            if fish_reserved_codepoint(c) {
+                // UTF-8 representations need at most 4 bytes.
+                let mut encoded_char = [0u8; 4];
```
krobelus, line 1108: Looks like  [MAX_LEN_UTF8](https://doc.rust-lang.org/std/char/constant.MAX_LEN_UTF8.html) is gonna [stabilize soon](https://github.com/rust-lang/rust/pull/145610), so I guess this is where we would add a comment like `// TODO MSRV(1.92) use std::char::MAX_LEN_UTF8`

src/common.rs
```diff
@@ -1102,84 +1088,54 @@ pub static EMPTY_STRING: WString = WString::new();
...
+                for byte in c.encode_utf8(&mut encoded_char).as_bytes() {
```
krobelus, line 1109: Since `str` is already UTF-8 encoded, there is no need to re-encode reserved codepoints.
We can use something like `input.char_indices()` and copy the requisite number of bytes (`c.len_utf8()`) directly without an intermediate buffer.

src/common.rs
```diff
@@ -1102,84 +1088,54 @@ pub static EMPTY_STRING: WString = WString::new();
...
+                // is valid UTF-8. This prefix may be empty.
+                let parsed_str = unsafe { std::str::from_utf8_unchecked(valid) };
+                append_escaped_str(&mut result, parsed_str);
+                // The length of the prefix of `after_valid` which is invalid UTF-8.
+                // `error_len` can return `None`, if it sees a byte sequence that could be the
+                // prefix of a valid code-point encoding at the end of the byte slice.
+                // This is useful when the input is chunked, but we don't do that, so in this case
+                // we use our custom encoding for all remaining bytes.
+                let error_len = e.error_len().unwrap_or(after_valid.len());
+                for byte in &after_valid[..error_len] {
```
krobelus, line 1139: > This is useful when the input is chunked, but we don't do that,

Hmm but `bytes2wcstring` may be passed a large string.
If there is a bad byte in the middle, we should retry UTF-8 parsing at the next byte, as the existing code does,
so we don't turn more bytes than necessary into PUA chars.
UTF-8 is free of "prefix conflicts", so random bad bytes will not lead to a domino effect.

 > so in this case we use our custom encoding for all remaining bytes.

src/common.rs
```diff
@@ -1302,22 +1252,6 @@ pub fn fish_setlocale() {
...
-    // ASCII char alternative. All of the can_be_encoded() invocations should return the same
-    // true/false value since the code points are in the BMP but we're going to be paranoid. This
-    // is also technically wrong if we're not in a Unicode locale but we expect (or hope)
-    // can_be_encoded() will return false in that case.
-    if can_be_encoded('\u{2026}') {
-        ELLIPSIS_CHAR.store(u32::from('\u{2026}'), Ordering::Relaxed);
-        ELLIPSIS_STRING.store(LL!("\u{2026}"));
-    } else {
-        ELLIPSIS_CHAR.store(u32::from('$'), Ordering::Relaxed); // "horizontal ellipsis"
-        ELLIPSIS_STRING.store(LL!("..."));
```
krobelus, line 1318: I was wondering if we want to leave a knob to turn off use of non-ASCII characters.
The obvious criteria would be if `LANG`/`LC_ALL`/`LC_CTYPE` are set to `C`.
Since we have had the "fix_locale" logic for years,
we always change from "C" to "C.UTF-8" if locales are not installed,
so leaving this knob in would only make a difference for
systems that don't actually provide the C.UTF-8 locale installed.

In theory, it might be nice to support niche OSs that might not have a capable font;
but I can't imagine a system that has Rust but doesn't have the C.UTF-8 locale.

So let's wait until someone asks for it (if ever), then we can add a knob ifneedbe.

src/env_dispatch.rs
```diff
@@ -301,8 +304,6 @@ fn handle_tz_change(var_name: &wstr, vars: &EnvStack) {
 
 fn handle_locale_change(vars: &EnvStack) {
     init_locale(vars);
-    // We need to re-guess emoji width because the locale might have changed to a multibyte one.
-    guess_emoji_width(vars);
```
krobelus, line 303: I think we still use our `extern "C"` wcwidth, which might be affected by locale variables (as set by `setlocale`, so we'll need to add that back).

Maybe we should stop doing this,
and assume the terminal follows the Unicode standard
and either use the `unicode-width` crate or our `widecharwidth` module.

The system `wcwidth` is not correct, because it doesn't handle variation selectors etc. (which is why we use `widecharwidth` and `unicode-width`),
and it's also not reliable for detecting terminal behavior,
because -- especially on macOS -- terminals ship their own Unicode database,
which explains our workarounds in `guess_emoji_width`.

But that's a question for a later change;
it depends on how well we want to support terminals that naïvely use `wcwidth()`.
We'll want to test some emoji scenarios; it might be fine
to remove all these workarounds, since most real-world scenarios
will work on every terminal, and there's plenty of terminals that implement Unicode well enough.

A robust solution for the text-width problem is https://github.com/kovidgoyal/kitty/issues/8226;
we could use the width part of this protocol.

src/input_common.rs
```diff
@@ -883,47 +884,35 @@ pub trait InputEventQueuer {
...
-                    let mut consumed = 0;
-                    let mut state = zero_mbstate();
-                    let mut i = 0;
+                    // At this point, the bytes in `buffer` should be parsed as a UTF-8 sequence,
+                    // or, if they are not valid UTF-8, ignored. On incomplete sequences, another
+                    // byte is read and decoding is tried again in the next iteration.
                     let ok = loop {
-                        if i == buffer.len() {
-                            buffer.push(
-                                match next_input_event(self.get_in_fd(), Timeout::Forever) {
```
krobelus, line 897: After spending a bit of time reading through this, this makes sense.
I saw that the existing code was overly cautious.
The only reason why

	assert!(i + 1 == buffer.len());

doesn't hold is that
we might have an escape prefix.

As a corollary, after receiving `DecodeState::Complete`,
we can always return, except if we only decoded the escape prefix.
So the

	if i + 1 == buffer.len()

should have been an `else assert`.

So maybe this should have been cleaned up separately,
but that's done now, so we're good.

src/input_common.rs
```diff
@@ -883,47 +884,35 @@ pub trait InputEventQueuer {
...
-                                    InputEventTrigger::Byte(b) => b,
-                                    _ => 0,
-                                },
-                            );
-                        }
-                        match decode_input_byte(
```
krobelus, line 903: Maybe call this `decode_input_bytes` now that it handles multiple bytes at once.
(I'm not sure if Rust has an incremental API like `mbr2wc` for decoding UTF-8,
but we're fine with redoing a little bit of work)

src/input_common.rs
```diff
@@ -1689,60 +1678,37 @@ pub(crate) enum InvalidPolicy {
...
-        }
-        InvalidPolicy::Passthrough => {
-            for &b in &buffer[*consumed..] {
-                out_seq.push(encode_byte_to_char(b));
+    match std::str::from_utf8(buffer) {
+        Ok(parsed_str) => {
+            for c in parsed_str.chars() {
+                // TODO: `fish_reserved_codepoint` does not handle the PUA encoding range, which
+                // might be changed to fix other behavior. If that happens, a separate function
+                // might be necessary, depending on the behavior we want here.
```
krobelus: it doesn't handle the whole PUA1 encoding range (nor PUA2),
but it does handle every PUA codepoint that has a special meaning to fish.
Other PUA characters can be left as-is (i.e. as single characters) safely,
fish won't really do anything special with them
(fish_is_pua() is barely used; the one use in codepoint_text() should probably be fish_reserved_codepoint() I guess).

 > `printf '\uf641'` prints `A` (0x41) instead of the expected UTF-8
 >     encoding (0xef 0x99 0x81).

sounds like a bug specific to our `printf` builtin, not the general encoding/decoding routines.

src/input_common.rs
```diff
@@ -1689,60 +1678,37 @@ pub(crate) enum InvalidPolicy {
...
+                InvalidPolicy::Passthrough => {
+                    for &b in buffer {
+                        out_seq.push(encode_byte_to_char(b));
+                    }
+                    Complete
+                }
+            },
+            None => Incomplete,
+        },
     }
```
krobelus, line 1708: > I am not particularly confident that my refactoring of
 > `src/input_common.rs:decode_input_byte` and the code calling it
 > is correct.

looks correct to me

 > The old code was quite a mess.

possibly, but the behavior seems right.

src/input_common.rs
krobelus: or `decode_user_input_utf8`.
We do use it for all strings that are user input (for statically-known strings we can use `std::str::from_utf8`); if that's too obvious, `decode_utf8` would work too

src/common.rs
danielrainer: Ah nice, I'll add that. We should be using `char::MAX_LEN_UTF8` though, the `std::char::MAX_LEN_UTF8` variant will be deprecated, if it is even introduced at all.

src/common.rs
danielrainer: The point of this function is to avoid the situation where the input is something like `[0xef, 0x99, 0x81]`, which would be parsed into `\u{f641}`, which would later be decoded into `[0x41]` instead of the original bytes.

TBF, I mainly considered the `\u{f600}` to `\u{f6ff}` range. `fish_reserved_codepoint` extends that downward by 256 codepoints and additionally checks `\u{fdd0}` to `\u{fdef}` and I haven't really checked or thought about what these are used for.

src/common.rs
danielrainer: Maybe I should put more info into the comment about how `error_len` works. If `Some(len)` is returned, len is the number of bytes which should be skipped because they cannot be decoded. `after_valid` is the slice which starts at the first error byte, so `after_valid[..error_len]` only contains the bytes which cannot be decoded. The rest of the input (`after_valid[error_len..]`) will be parsed in the next iteration of the loop, if it's non-empty.

danielrainer: And when `error_len` returns `None`, `after_valid.len()` is at most `MAX_LEN_UTF8 - 1`, i.e. 3.

src/common.rs
danielrainer: Sounds reasonable

src/env_dispatch.rs
danielrainer: Makes sense, I'll restore the call.

Kitty's text sizing protocol might be useful, but that will only work on terminals supporting it, right?

Generally, our approach of determining width of single codepoints individually is problematic. I think there has been previous discussion in fish issues, also linking to kitty. Properly determining width should be done for complete strings, rather than individual codepoints. This is why I added `unicode-width` to the `printf` crate, but so far that's the only place where it's used.

I also think functionality like `string length` is problematic. At the moment, it is specified that `string length reports the length of each string argument in characters`. This could be understood as any of
- number of bytes
- number of codepoints (now that we always assume UTF-8 it's at least clear which encoding would be used)
- number of codepoints that are assigned to characters
- number of grapheme clusters

Another possible interpretation one might have when not reading the docs is `number of terminal cells this string takes up when printed`, which might actually the most useful behavior. I think we should avoid using the ambiguous word `length` when referring to strings (and deprecate `string length`) and instead be explicit about the concrete behavior.

src/input_common.rs
danielrainer: Ah yes, I didn't realize that the second part of the condition in `fish_reserved_codepoint` includes the `\u{f600}` to `\u{f6ff}` range. I'll remove the comment.

> sounds like a bug specific to our printf builtin, not the general encoding/decoding routines.

Yes, let's look into that separately.

src/input_common.rs
danielrainer: renamed to `decode_user_input_utf8`

src/input_common.rs
danielrainer: Ok, I removed the comments

src/input_common.rs
krobelus: this was probably not the best suggestion; this function does the same thing as `bytes2wcstring` except it returns whether the input was incomplete, and provides a knob to fail on error.
(I wonder if this should be sharing code with `bytes2wcstring`.)
But inheriting that kind of naming probably doesn't make much sense - maybe `decode_one_utf8_codepoint`, i think I'll do that.

src/common.rs
krobelus: the 256 codepoint-range below `ENCODE_DIRECT_BASE` is used for named keys like `enter`,
and the other range is for internal wildcard/expansion use.
The result of `bytes2wcstring` shouldn't ever be used as keyboard input,
i.e. we won't call `Key::from_raw(bytes2wcstring(...).chars().next().unwrap())`,
so we probably don't need to escape that here..
But `decode_input_byte` does need to escape those AFAIK, so maybe best keep it for consistency

Issue #11976, opened 2025-10-19
Test `test_dup2_during_select_ebadf` fails on MSYS/Cygwin


Nahor:
Test `test_dup2_during_select_ebadf` on MSYS because `poll` returns `1` while the test expects either a timeout (Linux) or a `EBADF` error (MacOS).

This is in Cygwin's `poll` implementation. I have a pure Cygwin C version of the test that fails as well (see attached code below).

However, before emailing the Cygwin mailing list, is this actually a problem? Or is the test overly restrictive?
If the test is to ensure `BackgroundFdMonitor::run` will work, because of the `ItemID` stuff, it feels like the monitor will not care what the state of the descriptor is after `dup2` anyway... assuming `ItemID` does change after `dup2`.
And if `ItemID` does not change, is it a problem if the descriptor is flagged as readable when it's not? According `select`'s man page and the `BUGS` section in particular, this can happen on Linux as well (and they advise to use `O_NONBLOCK`), so ideally Fish should already be ready for that.

TLDR;
- Should we remove that test as unnecessary?
- Or should we relax the test by ignore positive return values?
- Or should we email the Cygwin team?

<details>
<summary>C test case</summary>

```C
#include <assert.h>
#include <errno.h>
#include <poll.h>
#include <pthread.h>
#include <stdio.h>
#include <string.h>
#include <unistd.h>

// Uncomment to show blocking `read()`
// #define TEST_READ

typedef struct {
  pthread_barrier_t *barrier;
  int fd_read;
  int fd_write;
} thread_args;

static void *thread_func(void *arg) {
  thread_args *args = (thread_args *)arg;

#ifdef TEST_READ
  {
    // Sync with main, then sleep to let main start `poll`
    int res = pthread_barrier_wait(args->barrier);
    assert(res == 0 || res == PTHREAD_BARRIER_SERIAL_THREAD);
    usleep(100);

    printf("write\n");
    char *buff = "hello";
    assert(write(args->fd_write, buff, strlen(buff) + 1) == strlen(buff) + 1);
    printf("write done\n");
  }
#endif

  // Sync with main, then sleep to let main start `poll`
  int res = pthread_barrier_wait(args->barrier);
  assert(res == 0 || res == PTHREAD_BARRIER_SERIAL_THREAD);
  usleep(100);

  printf("dup2\n");
  int pipefd[2];
  assert(pipe(pipefd) == 0);
  int old_fd = args->fd_read;
[...]

PR #11979, opened 2025-10-19
Fix tests under MSYS and enable CI testing for it

### LINKED ISSUE #11933: Test `tests::history::test_history_races` fail on MSYS2
I'm on the master branch (b88622b), using MSYS2 on Windows 11.
`tests::history::test_history_races` freezes/deadlocks and the temporary fish intance (`fish-<hex>.exe`) needs to be killed wit the task manager.

As far as I can tell, the issue is in `LockedFile::new` where the `flock()` call fails with an `errno` 14 "Bad address". At that point, all the other threads are stuck in their own `flock()` calls.

If I put small `thread::sleep` after the `flock` loop, everything works fine.
```diff
--- a/src/fs.rs
+++ b/src/fs.rs
@@ -147,10 +147,11 @@ pub fn new(locking_mode: LockingMode, file_path: &wstr) -> std::io::Result<Self>
         // Try locking the directory. Retry if locking was interrupted.
         while unsafe { flock(dir_fd.as_raw_fd(), locking_mode.flock_op()) } == -1 {
             let err = std::io::Error::last_os_error();
             if err.kind() != std::io::ErrorKind::Interrupted {
                 return Err(err);
             }
         }
+        std::thread::sleep(std::time::Duration::from_nanos(1));

         // Open the data file
         let data_file = wopen_cloexec(file_path, locking_mode.file_flags(), LOCKED_FILE_MODE)?;
```
While I used the minimum duration possible (1ns), it does not actually change the duration of the test until sleeping for ~8+ms (for a total test duration of ~15s on my machine). I expect this is normal because of Windows timer resolution.
I also tried `std::thread::yield_now()` but that did not work.

Nahor:
This fixes or add workarounds for various issues when running `cargo test` on MSYS.

Couple notes:
- This does NOT fix #11933, it only add a workaround to slightly improve things.
  Strictly speaking it's not necessarily to enable the test since that particular test still needs to be disable for another problem. I do believe it makes main Fish less likely to freeze/deadlock though, so I'm pushing that change.
- It contains my fix for #11976 since it's simpler and I already had it in my git. It can be replaced with the change @krobelus suggested in that issue if preferred.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -166,8 +166,9 @@ jobs:
       run: |
         cargo build
     - name: smoketest
-      # We can't use `cargo test` yet, there are just too many failures
+      # We can't run `build_tools/check.sh` yet, there are just too many failures
       # so this is just a quick check to make sure that fish can swim
       run: |
         set -x
         [ "$(target/debug/fish.exe -c 'echo (math 1 + 1)')" = 2 ]
+        cargo test


.github/workflows/test.yml
```diff
@@ -165,9 +165,6 @@ jobs:
     - name: cargo build
       run: |
         cargo build
-    - name: smoketest
-      # We can't use `cargo test` yet, there are just too many failures
-      # so this is just a quick check to make sure that fish can swim
+    - name: cargo test
```
krobelus: looks good overall. Couple of minor comments

.github/workflows/test.yml
```diff
@@ -165,9 +165,6 @@ jobs:
...
       run: |
-        set -x
-        [ "$(target/debug/fish.exe -c 'echo (math 1 + 1)')" = 2 ]
+        cargo test
```
krobelus, line 174: while adding a lot of coverage, this technically also loses some because our `cargo test` currently doesn't invoke the full fish binary.

For full tests, we'd either want to run

	FISH_CHECK_LINT=false build_tools/check.sh

or

	mkdir build; cd build
	cmake .. -GNinja
	ninja fish_run_tests

In future we might migrate everything to `cargo test`.

So maybe keep the smoke test until we have that.

src/highlight/tests.rs
```diff
@@ -866,10 +866,13 @@ mod file_tester_tests {
         assert!(!result);
 
         // Can't redirect from an unreadable file.
-        fs::set_permissions(&file_path, Permissions::from_mode(0o200)).unwrap();
-        let result = tester.test_redirection_target(L!("file.txt"), RedirectionMode::input);
-        assert!(!result);
-        fs::set_permissions(&file_path, Permissions::from_mode(0o600)).unwrap();
+        #[cfg(not(cygwin))] // Can't mark a file write-only on MSYS, this may work on true Cygwin
```
krobelus, line 869: > Note: Cygwin by default does use "acl", which may allow the check to pass (unverified).

`build.rs` currently defines `cfg(cygwin)` as

	env_var("CARGO_CFG_TARGET_OS").unwrap() == "cygwin"

Perhaps `build.rs` can detect the acl/noacl state of the current file system,
and base our check on that?
I.e. the equivalent of

	set mountpoint $(df . --output=source | tail -1)
	mount | string match -r ^$mountpoint.\* | grep noacl

src/tests/complete.rs
```diff
@@ -96,14 +96,16 @@ fn test_complete() {
     std::fs::create_dir_all("test/complete_test").unwrap();
     std::fs::write("test/complete_test/has space", []).unwrap();
     std::fs::write("test/complete_test/bracket[abc]", []).unwrap();
-    std::fs::write(r"test/complete_test/gnarlybracket\[abc]", []).unwrap();
-    #[cfg(not(windows))] // Square brackets are not legal path characters on WIN32/CYGWIN
+    #[cfg(not(cygwin))]
+    // Backslashes and colons are not legal filename characters on WIN32/CYGWIN
```
krobelus, line 100: Oh does this mean the `cfg(windows)` and `cfg(not(windows))` checks in
`Cargo.toml` are bogus as well?
My bad, I didn't actually test that exact patch on Cygwin.
It exists to make `cargo b && target/debug/fish` work on Cygwin
(release builds already worked before).

src/wutil/dir_iter.rs
```diff
@@ -403,11 +405,19 @@ fn test_dir_iter() {
     let selflinkname = "selflink"; // link to self
     let fifoname = "fifo";
     #[rustfmt::skip]
-    let names = &[
-        dirname, regname, reglinkname, dirlinkname,
-        badlinkname, selflinkname, fifoname,
-    ];
+    let names = if cfg!(not(cygwin)) {
```
krobelus, line 408: nit: maybe better to flip the branches so we can say `if cfg!(cygwin)` ?

src/wutil/dir_iter.rs
```diff
@@ -403,11 +405,19 @@ fn test_dir_iter() {
...
+        vec![
+            dirname, regname, reglinkname, dirlinkname,
+            badlinkname, selflinkname, fifoname,
+        ]
```
krobelus: nit: trailing whitespace on this line

src/wutil/dir_iter.rs
```diff
@@ -403,11 +405,19 @@ fn test_dir_iter() {
...
+            badlinkname, selflinkname, fifoname,
+        ] 
+    }
+    else {
+        // Symbolic links on Windows are complicated. Their behavior depends
+        // on the CYGWIN or MSYS env variable. So we skip that part of the test
+        vec![dirname, regname, fifoname]
+    };
 
+    #[allow(unused_variables)] // Unused on Cygwin
```
krobelus: instead of the comment, we'd add a `#[cfg(cygwin)]` above this line.
But actually, better compile away the entire `let is_link_name` on Cygwin,
by only using `#[cfg(cygwin)]` here.

src/wutil/dir_iter.rs
```diff
@@ -465,6 +481,9 @@ fn test_dir_iter() {
 
         // Links should never have a fast type if we are resolving them, since we cannot resolve a
         // symlink from readdir.
+        #[cfg(not(cygwin))]
+        // Symbolic links on Windows are complicated. Their behavior depends
```
krobelus: I think there's no need to repeat the same comment to 3 places, the commit message attached to the `#[cfg(not(cygwin))]` line is enough for me.

 > // Symbolic links on Windows are complicated. Their behavior depends
 > // on the CYGWIN or MSYS env variable

if it were just for environment variables, we could probably work
around that, i.e. we could override them in tests
(unless that's frowned upon).
But you mention that file systems limitations are another reason,
so probably there's enough strong reasons for disabling the test.

.github/workflows/test.yml
Nahor: Ok, will do

src/highlight/tests.rs
Nahor: Are tests expected to be run only in the build environment?

I did wonder if it would be worthwhile to spend more time on the issue but in then I though the cons outweighted the pros:
- checking for acl/noacl would cause different people potentially seeing different test results
- MSYS default, and thus the CI, use noacl, so we would
  - either have a CI doing the simpler test and never checking the more complicated ones
  - or have to figure out how to change the CI to using "acl"  (and like needing even two CI: one with noacl and one with)
-  if Linux/Mac passes the tests already, then it feels to me that any issue on MSYS/Cygwin isn't likely to be fixable/workaroundable and we wouldn't end up just disabling the test anyway.

In then end, I figured doing more was beyond the scope of this PR.

src/tests/complete.rs
Nahor: When I looked at it, it seemed that `cfg(windows)` was not used anymore. Very few, if any, left in the main code. And several tests only using `cygwin`.
I also figured that `windows` ought to mean a native app (MSYS2 UCRT/MINGW instead MSYS2 MSYS/Cygwin) .

That said, I admit I only did a cursory look and could have jumped to a conclusion prematurely.

src/wutil/dir_iter.rs
Nahor: I did it on purpose to keep the full/main array first, and do the more limited/specific case second.

src/wutil/dir_iter.rs
Nahor: This is deliberate, to limit the amount of conditional code.
~If the function is removed, then more code needs to be put behind a `cfg`s or become `allow`. In particular, the `xxxlinkname` variables should become unused, and because they are otherwise used everywhere in the test, they can't be all put behind a single `cfg` (they would become scoped variables). We would need one for each variable.~ Actually, there isn't. The `xxxlinkname` variables are still referenced in the function's loop, so I'll fix it.

So for simplicity/readability I favored `allow` on the function rather than eliminating it.

src/wutil/dir_iter.rs
Nahor: > I think there's no need to repeat the same comment to 3 places, the commit message attached to the #[cfg(not(cygwin))] line is enough for me.

I figured it's better to have the comment right there, where one is looking, than expecting them to look in other places in the code or find the commit message. But I can remove it.

> if it were just for environment variables, we could probably work around that, i.e. we could override them in tests (unless that's frowned upon).
> But you mention that file systems limitations are another reason, so probably there's enough strong reasons for disabling the test.

Indeed, there is no perfect solution to symlinks on windows, so Cygwin let the user choose with the env var. And not all filesystems support all options either (NTFS vs FAT, recent NTFS vs old NTFS)
And the best option is also blocked behind admin privileges or developer mode.

We could have test dependent on what's available on the build machine, but .. (see comment above about permissions)

src/tests/complete.rs
Nahor: I confirm: `cfg(windows)` does not work in MSYS environments (there is no Rust for pure Cygwin, so I can't check but there is no reason to believe it will be different)

```rust
    #[cfg(windows)]
    println!("cargo:warning=in Windows");
    #[cfg(not(windows))]
    println!("cargo:warning=in non-Windows");
```
prints:
```
warning: fish@4.1.0-snapshot: in non-Windows
```

src/wutil/dir_iter.rs
krobelus: > The xxxlinkname variables are still referenced in the function's loop

hmm if we wanted to be really consistent about how we uses `cfg(cygwin)`, then we should also remove those branches and not define `reglinkname` etc. on cygwin. It's not the most important thing, but I'd probably try to go for a simple solution (either compile away everything, or nothing and replace it with runtime checks), as opposed to an easy solution (i.e. small amounts of code change).
If you keep the existing style, better replace the comment with `#[cfg(cygwin)]` though

src/highlight/tests.rs
krobelus: > Are tests expected to be run only in the build environment?

Not sure right now, but this might already be assumed elsewhere,
like in our workspace-level `build.rs`.
But yeah, querying in build.rs doesn't seem right,
but we could do it at test run time instead.

> would cause different people potentially seeing different test results

yeah but that's okay - people who enable ACLs *should* see a failure,
so we can actually address it

If we could detect when we're building for msys2 (as opposed to cygwin),
we could skip it only for that case. That would be better,
though not as good as probing for ACL support.

It's fine to leave it as you wrote for now,
just saying how we can improve this.

src/tests/complete.rs
krobelus: our smoketest's `target/debug/fish` should be a standalone build, I wonder how that worked.
Need to double check if 012b5071286 (Workaround for embed-data debug builds on Cygwin, 2025-09-16) is actually needed.., if yes we can correct it

src/wutil/dir_iter.rs
krobelus: > I figured it's better to have the comment right there, where one is looking, than expecting them to look in other places in the code or find the commit message.

One problem is that putting comments in the code (especially when duplicating it) incentivizes the wrong thing, such as keeping the comment overly concise.
git-blame should only be one key stroke away from the editor,
and it allows to apply a comment to arbitrary lines,
without a limit to the length.

I think we overuse comments in general.
Most of the time, I completely ignore comments,
because they're usually irrelevant to what I'm doing.
In this spirit, I think it makes sense to keep the cygwin ifdefs small,
to not visually clutter the test, because 90% of readers
don't care about these comments.
If they care enough to want to find out why the test is disabled on Cygwin,
they can run `git blame` to get a full explanation
that cross-references the other affected places at a given point in time.

Of course there is another solution:
extract a function, say `fn are_symlinks_supported() -> bool`,
at the comment only at the function definition.
Then every call site of this function is linked to that comment,
the code is probably better, and there is no comment duplication.

> We could have test dependent on what's available

that sounds good - it's an argument for doing the checks at runtime instead of `cfg(cygwin)`

PR #11985, opened 2025-10-21
Fixes for prompts with empty last lines and tmux tests improvments

kerty0:
## Description

Two fixes for multiline prompts with empty last line:
1. Missing right prompt when commandline is empty.
2. Not preserving empty line on executing empty commandline.

Simplification of `Screen.actual.cursor` behavior by removing a change of origin point implicitly performed by `self.r#move(0, 0);` from the start of the rendered commandline to the viewport's upper-left corner if it was closer because cursor can't move viewport up.

When initializing tmux tests:
1. Wait until any output, not just output containing 'prompt'.
2. Pass argument to fish directly rather than through `isolated_tmux_fish_extra_args` variable.
3. Explicitly resize tmux window so we can attach to session without resizing pane. 

---

Also @krobelus, regarding "`pub` for testing", what do you think about instead of creating `tests` module in `lib.rs`, we `include!()` test files as submodules? The main disadvantage is subpar Rust Analyzer support (syntax highlighting and completions don't work, though jumping does), which can be worked around by copy-pasting test code. Overall, it keeps test code separated and removes the need for test-only `pub`.

<details>
<summary>Example for screen.rs</summary>

```diff
diff --git a/src/screen.rs b/src/screen.rs
index bec8853942..0f96d53c2c 100644
--- a/src/screen.rs
+++ b/src/screen.rs
@@ -2066,3 +2066,8 @@
 pub fn wcswidth_rendered(s: &wstr) -> isize {
     s.chars().map(wcwidth_rendered).sum()
 }
+
+#[cfg(test)]
+mod tests {
+    include!("tests/screen.rs");
+}
diff --git a/src/tests/mod.rs b/src/tests/mod.rs
index 6cddc8a296..4165a9dece 100644
--- a/src/tests/mod.rs
+++ b/src/tests/mod.rs
@@ -19,7 +19,6 @@
 mod parser;
 mod reader;
 mod redirection;
-mod screen;
 mod std;
 mod string_escape;
 mod termsize;
```

</details>

### DIFF
--- a/tests/checks/tmux-transient-prompt.fish
+++ b/tests/checks/tmux-transient-prompt.fish
@@ -26,8 +26,9 @@ isolated-tmux send-keys C-u '
     function fish_prompt
         printf "\$ "
     end
-' C-l
-isolated-tmux send-keys Enter Enter
+'
+tmux-sleep
+isolated-tmux send-keys C-l Enter Enter
 tmux-sleep
 isolated-tmux capture-pane -p
 # CHECK: $
@@ -43,8 +44,9 @@ isolated-tmux send-keys C-u C-l '
             printf "transient line%d\n" 1 2
         end
     end
-' C-l
-isolated-tmux send-keys Enter
+'
+tmux-sleep
+isolated-tmux send-keys C-l Enter
 tmux-sleep
 isolated-tmux capture-pane -p
 # CHECK: final line1
@@ -63,8 +65,9 @@ isolated-tmux send-keys C-u C-l '
             echo "1> "
         end
     end
-' C-l
-isolated-tmux send-keys 'echo foo' Enter
+'
+tmux-sleep
+isolated-tmux send-keys C-l 'echo foo' Enter
 tmux-sleep
 isolated-tmux capture-pane -p
 # CHECK: 2> echo foo
@@ -76,7 +79,7 @@ isolated-tmux capture-pane -p
 # final.
 isolated-tmux send-keys C-u C-l
 isolated-tmux send-keys 'echo foo \\' Enter
-isolated-tmux send-keys 'bar' Enter
+isolated-tmux send-keys bar Enter
 tmux-sleep
 isolated-tmux capture-pane -p
 # CHECK: 2> echo foo \


src/screen.rs
```diff
@@ -498,10 +494,8 @@ impl Screen {
                     mut cursor,
                     scroll_amount,
                 } = scrolled_cursor;
-                if scroll_amount != 0 {
-                    if !is_final_rendering {
-                        self.desired.line_datas = self.desired.line_datas.split_off(scroll_amount);
-                    }
+                if scroll_amount != 0 && !is_final_rendering {
+                    self.desired.line_datas = self.desired.line_datas.split_off(scroll_amount);
                     cursor.y -= scroll_amount;
```
krobelus, line 499: it's weird that the wrong cursor y never caused issues.

Later, Screen::update does

        let Cursor { x, y } = self.desired.cursor;
        self.r#move(x, y);

Was that wrong before this change? (haven't tested it yet)

src/screen.rs
```diff
@@ -754,8 +748,9 @@ impl Screen {
     pub fn cursor_is_wrapped_to_own_line(&self) -> bool {
         // Don't consider dumb terminals to have wrapping for the purposes of this function.
         self.actual.cursor.x == 0
-            && self.actual.cursor.y != 0
+            && self.actual.cursor.y + 1 != self.actual.visible_prompt_lines
             && self.actual.cursor.y + 1 == self.actual.line_count()
```
krobelus, line 752: > However, commit 606802da (Make ScreenData track multiline prompt)

FWIW I have shortcut in my VCS UI to copy to clipboard the output of:

	git log -1 $selected_commit --pretty=reference

tests/checks/tmux-abbr.fish
```diff
@@ -1,13 +1,12 @@
 #RUN: %fish %s
 #REQUIRES: command -v tmux
 
-set -g isolated_tmux_fish_extra_args -C '
+isolated-tmux-start -C '
```
krobelus, line 4: yeah there's a lot that could be improved about these tests,
we should probably find a shorter name for "isolated-tmux",
probably we should simply call it "tmux".

(we probably don't strictly need a new tmux client process everytime,
we could open a longer lived tmux in control mode and echo commands
to its stdin (via a fifo).. but I'm not sure if that's an easy change)

tests/checks/tmux-empty-prompt.fish
```diff
@@ -0,0 +1,77 @@
+#RUN: %fish %s
+#REQUIRES: command -v tmux
+
+isolated-tmux-start -C '
+    function fish_prompt; end
+    function fish_right_prompt
+        set -q right_prompt
+        or echo right-prompt
```
krobelus: it's weird that presence of `$right_prompt` turns *off* right prompt.
I'd probably expect it other way round.
But actually, do we need to overload the ctrl-g binding like this?

Maybe create separate bindings, one for turning on right prompt
and another for turning it off;
that way the test should be easier to read?

tests/checks/tmux-empty-prompt.fish
```diff
@@ -0,0 +1,77 @@
...
+    end
+    bind ctrl-g "set -q right_prompt; and set -e right_prompt; or set right_prompt 1; commandline -f repaint"
```
krobelus: I'd probably write this in multiple lines

	bind ctrl-g "
		set -q right_prompt
		and set -e right_prompt
		or set right_prompt 1
		commandline -f repaint
	"

tests/test_functions/isolated-tmux-start.fish
```diff
@@ -48,14 +48,17 @@ function isolated-tmux-start
...
         # No transient prompt.
         set fish_transient_prompt 0
-    ' $isolated_tmux_fish_extra_args
+    ' $argv
     # Set the correct permissions for the newly created socket to allow future connections.
     # This is required at least under WSL or else each invocation will return a permissions error.
     chmod 777 .tmux-socket
 
+    # Resize window so we can attach to tmux session without changing panel size.
+    isolated-tmux resize-window -x 80 -y 10
```
krobelus: probably better to extract variables

	set -l x 80
	set -l y 10

tests/test_functions/isolated-tmux-start.fish
```diff
@@ -48,14 +48,17 @@ function isolated-tmux-start
...
     chmod 777 .tmux-socket
 
+    # Resize window so we can attach to tmux session without changing panel size.
+    isolated-tmux resize-window -x 80 -y 10 
+
     # Loop a bit, until we get an initial prompt.
-    for i in (seq 25)
-        if string match -q '*prompt*' -- (isolated-tmux capture-pane -p)
+    for i in (seq 50)
+        if test -n "$(isolated-tmux capture-pane -p)"
```
krobelus, line 62: I guess in future we should replace this with a wait method that also works for empty prompts.

 > This affected tmux-wrapping.fish and tmux-read.fish tests.

Oof

 > Change it so initialization function wait until any output and increase
 > timeout to 10 seconds to make such issues more obvious.

not sure if that's obvious enough.
Let's maybe exit the entire test script and/or add an error message.
Either should make tests fail.
A function doing "exit" is unconventional, but it should work well for our use case.

```diff
diff --git a/tests/test_functions/isolated-tmux-start.fish b/tests/test_functions/isolated-tmux-start.fish
index 1f30b9d34b..225048c310 100644
--- a/tests/test_functions/isolated-tmux-start.fish
+++ b/tests/test_functions/isolated-tmux-start.fish
@@ -59,8 +59,9 @@
     # Loop a bit, until we get an initial prompt.
     for i in (seq 50)
         if test -n "$(isolated-tmux capture-pane -p)"
-            break
+            return
         end
         sleep .2
     end
+    echo >&2 "isolated-tmux-start: timed out waiting for non-empty first prompt"
 end
```

src/screen.rs
kerty0: Thanks for the tip!!

src/screen.rs
kerty0: > it's weird that the wrong cursor y never caused issues.

It wasn't wrong, just unintuitive and implicit. I will try to explaining it differently:

Let `y=0` be the first line of the rendered commandline. Then, in final
rendering, the final value of the desired cursor was measured from
`y=scroll_amount`. This wasn't a problem because
```
if !MIDNIGHT_COMMANDER_HACK.load() {
    self.r#move(0, 0);
}
```
implicitly changed the actual cursor to be measured from `y=0` to `y=scroll_amount`.
This happened because the cursor actually moved to `y=scroll_amount` but had its
`y` value set to 0. Since the actual and desired cursors were both measured
from the same line, the code was correct.

krobelus: thanks for updating the commit message. That plus the new test will make future rewrites easier

tests/checks/tmux-abbr.fish
krobelus: we should maybe also teach our `tmux send-keys` wrapper to accept our `ctrl-g` notation

PR #11990, opened 2025-10-22
Rust 2024 Edition

henrikhorluck:


Cargo.toml
```diff
@@ -1,11 +1,10 @@
 [workspace]
-resolver = "2"
 members = ["crates/*"]
 
 [workspace.package]
 # To build revisions that use Corrosion (those before 2024-01), use CMake 3.19, Rustc 1.78 and Rustup 1.27.
 rust-version = "1.85"
-edition = "2021"
+edition = "2024"
```
krobelus, line 7: > set edition 2025

2025 -> 2024.

 > fix compile issues

Let's avoid needless surprise -- every commit should pass all tests.

build.rs
```diff
@@ -39,7 +39,8 @@ fn main() {
     // the source directory is the current working directory of the build script
     rsconf::set_env_value("FISH_BUILD_VERSION", version);
 
-    std::env::set_var("FISH_BUILD_VERSION", version);
+    // safety: single-threaded code.
+    unsafe { std::env::set_var("FISH_BUILD_VERSION", version) };
```
krobelus, line 43: re dbbd621e717 (run cargo fix --edition, 2025-10-22).
Please make this easier to review by squashing commits to remove intermediate changes like

	// TODO: Audit that the environment access only happens in single-threaded code.

(I wonder if "cargo fix" adds the comment?)

I think we should also lose the intermediate `expr_2021` ones,
since reviewing the diff is not easier than reviewing the results of `rg --vimgrep :expr`.
A note in the commit message pointing that out would be enough.

build.rs
```diff
@@ -161,9 +162,9 @@ fn has_small_stack(_: &Target) -> bool {
     {
         use core::ffi;
 
-        extern "C" {
```
krobelus, line 164: As a reviewer, the forward references are throwing me off
You first make changes like

```diff
-            if let Some(node) = self.try_parse::<Contents>() {
+            match self.try_parse::<Contents>() { Some(node) => {
```

without an explicit reason,
so in my first linear pass, I spent time starting to comment, suspecting it's unnecessary intermediate changes (because they are reverted later, but I don't want to have to peek ahead).

So better add justification to the first step next time.

I guess this specific lint is less trivial to grep for than the `expr_2021` one.
But we can skip all `if let foo` where `foo`'s Drop implementation doesn't do anything interesting (i.e. it only frees  memory),
which seems to apply to every single instance.

So after having reviewed that one, I think I'd also remove it from history,
since it's very unlikely that we need to bisect to this.

Should also drop the formatting commit.

build.rs
```diff
@@ -161,9 +162,9 @@ fn has_small_stack(_: &Target) -> bool {
...
-            fn pthread_get_stacksize_np(thread: *const ffi::c_void) -> usize;
-            fn pthread_self() -> *const ffi::c_void;
+        unsafe extern "C" {
+            unsafe fn pthread_get_stacksize_np(thread: *const ffi::c_void) -> usize;
+            unsafe fn pthread_self() -> *const ffi::c_void;
```
krobelus, line 167: > declare all extern "C" as unsafe as precaution

precaution for what? Use in threaded code?
Is `unsafe fn` not already implied by the surrounding `unsafe extern "C"`?

src/lib.rs
```diff
@@ -23,6 +23,7 @@
 #![allow(clippy::option_map_unit_fn)]
 #![allow(clippy::too_many_arguments)]
 #![allow(clippy::uninlined_format_args)]
+#![allow(clippy::let_and_return)]
```
krobelus, line 26: re 91d8975384a (Ignore noisy rule, 2025-10-22)

     > fn hello() -> impl Debug {
     >     let useful_name = xxx;

but shouldn't the function name be fixed in that case?
Or is this about unnamed lambdas?
I think this is a good lint in many cases,
but I guess I'm be okay with temporarily disabling it for now.

I don't really know whether this is related to the edition upgrade,
I think we had the lint already?

src/locale.rs
```diff
@@ -10,18 +10,20 @@ const CHAR_MAX: libc::c_char = libc::c_char::MAX;
...
+            Some((*s as u8) as char)
+        } else {
+            None
+        }
     }
 }
 
 /// Convert a libc lconv to a Locale.
 unsafe fn lconv_to_locale(lconv: &libc::lconv) -> Locale {
-    let decimal_point = first_char(lconv.decimal_point).unwrap_or('.');
```
krobelus, line 23: re a7d4d521a80 (Make unsafes more granular, 2025-10-23),
yeah better squash this as well, don't make reviewers read poorly-styled `cargo fix` output

src/locale.rs
```diff
@@ -61,28 +63,29 @@ unsafe fn lconv_to_locale(lconv: &libc::lconv) -> Locale {
...
         return None;
     }
 
-    let lconv = localeconv_l(loc);
+    let lconv = unsafe { localeconv_l(loc) };
     let result = if lconv.is_null() {
         None
     } else {
-        Some(lconv_to_locale(&*lconv))
+        Some(unsafe { lconv_to_locale(&*lconv) })
```
krobelus, line 85: all our commits should try to pass all lints, to remove forward all lint-related references -- this allows readers to prune the search space for answers to the "what triggered this" question.

src/wutil/tests.rs
```diff
@@ -72,7 +72,7 @@ fn test_wwrite_to_fd() {
         assert!(fd.is_valid());
         let mut input = WString::new();
         for _i in 0..size {
-            input.push(rng.gen());
+            input.push(rng.r#gen());
```
krobelus, line 75: I think this means we can delete `.rustfmt.toml` (enabled by the commit that upgrades the edition, so let's put it in the same commit to make that obvious) which allows me to drop https://github.com/fish-shell/fish-shell/pull/11992/commits/0856141e9902922adb3dbd0cf573d621e2a746c6

build.rs
krobelus: BTW I think the convention in the Git project is to add commits with subjects starting with `SQUASH??` if they are meant to be squashed after initial review feedback. We rarely need this but I guess it makes sense here.
(I didn't check the `expr_2021` ones, I assume you did a cursory check and I'd be very surprised if `const` or `_` changes would break *silently* on us)

Cargo.toml
henrikhorluck: The commits are intended to follow https://doc.rust-lang.org/edition-guide/editions/transitioning-an-existing-project-to-a-new-edition.html#transitioning-an-existing-project-to-a-new-edition

> 1. Run cargo update to update your dependencies to the latest versions.
> 2. Run cargo fix --edition
> 3. Edit Cargo.toml and set the edition field to the next edition, for example edition = "2024"
> 4. Run cargo build or cargo test to verify the fixes worked.
> 5. Run cargo fmt to reformat your project.

I understand that that might seem confusing without prior guidance, and I should have made that more explicit.

> set edition 2025

Yep that was a glip 😅, fixed now

build.rs
henrikhorluck: https://doc.rust-lang.org/edition-guide/rust-2024/unsafe-extern.html#details

`unsafe` in `extern "C"` is now required. The signatures themselves still default to `unsafe`, but can also be explicitly marked `safe`. I changed it in spirit of [unsafe_op_in_unsafe_fn](https://doc.rust-lang.org/edition-guide/rust-2024/unsafe-op-in-unsafe-fn.html#unsafe_op_in_unsafe_fn-warning), but you are right that the addition of `unsafe` in the signatures is an unnecessary change

src/lib.rs
henrikhorluck: The offenders are 

https://github.com/fish-shell/fish-shell/blob/a7d4d521a80961c9768c5d2c271313b97e91b287/src/env/environment_impl.rs#L903-L904

and 

https://github.com/fish-shell/fish-shell/blob/a7d4d521a80961c9768c5d2c271313b97e91b287/src/env/environment.rs#L475-L477

Both IMO clear examples of the lint not being useful.

The lint does not appear in master, but appears when running `cargo clippy` in bbf1d2154f3b20dbd0a3629b2451ea67cb354f8d, that seems like a bug in clippy as I did not see that any related code is affected in any commits here. We have set warnings to deny in CI so this needs to be part of this PR for CI to pass

I do not believe the lint is useful as is

build.rs
henrikhorluck: Everything in https://github.com/fish-shell/fish-shell/commit/dbbd621e717b67150ff33ce12499fe5023dd6f39 is from running `cargo fix --edition`, and the follow up-commits are to make it explicit that we are opting into the new defaults. There are some cases of drop-orderings that did not have automatic fixes, which you can see by running `cargo fix --edition` from main, mostly due to https://doc.rust-lang.org/edition-guide/rust-2024/temporary-tail-expr-scope.html.

build.rs
henrikhorluck: > You first make changes like

No. I did not make a change. `cargo fix --edition` made the change, but I think it is better to describe it as `cargo fix` _preserved the semantic behaviour_, while the change in behavior is later when the changes are reverted, which I did manually.

> But we can skip all if let foo where foo's Drop implementation doesn't do anything interesting (i.e. it only frees memory),

Which I believe `cargo fix --edition` is already aware of, as we have _way_ more `if let`'s than those modifed. As far as I can tell the ones modified all have custom drop-implementations or `cargo fix` cannot determine if there is a custom drop due to generics.

I am just following the guide:

https://doc.rust-lang.org/edition-guide/rust-2024/temporary-if-let-scope.html#migration

> After the migration, it is recommended that you review all of the changes of if let to match and decide what is the behavior that you need with respect to when temporaries are dropped. If you determine that the change is unnecessary, then you can revert the change back to if let.

src/locale.rs
henrikhorluck: Squashed

src/wutil/tests.rs
henrikhorluck: Added the removal of .rustfmt.toml

henrikhorluck (commented):
I squashed most into one commit expect the `if let`-changes, and moved the lint-ignore to the first commit. I think all commits should be passing now as requested

build.rs
krobelus: Ok that's probably a reasonable compromise.
I didn't know about the edition upgrade process, the behavior-preserving step does make a lot of sense.
I'm just always suspicious of anything that makes review harder, though I guess on a change like this that's less critical

src/lib.rs
krobelus: > I do not believe the lint is useful as is

it's a tradeoff; lints like this dumb down the language but that can be a good thing.

Anyway, disabling the lint just preserves historical behavior so this is unrelated.

src/wutil/tests.rs
krobelus: > I think this means we can delete `.rustfmt.toml`

Nope, apparently

	rustfmt src/tests/complete.rs

does something different than

	rustfmt --style-edition=2024 src/tests/complete.rs

This seems weird because it will affect any project that uses 2024 edition
and doesn't explicitly specify style-version in rustfmt.toml.
Not sure if there's already a tracking issue.
There are some like https://github.com/rust-lang/rustfmt/issues/5650

PR #11992, opened 2025-10-23
Move Rust tests to implementation files

krobelus:
<details>

<summary> description outdated, see commits </summary>

... and allow tests to access private APIs only if they need to (WIP).

---

We have two kinds of Cargo tests, those that test private APIs, and those that don't.
We mostly keep these two in separate modules ("foo::tests" and "tests::foo" respectively).
I assume we don't want to combine them into one module.
When "foo" is defined in its own file (common case), this implies that the two kinds of tests are defined in separate files.

Testing desiderata are:

1. tests should not be in the same file as the implementation [^1]
2. when testing private APIs, we don't want to add "pub" nor
   "pub(crate)" to private APIs. In our workspace crate, the two are
   pretty much the same since we still need to split up that crate.
3. we want rust-analyzer support
4. we want all tests to live in single directory (src/tests) if possible

The reason for 1 (and 4) is to make it easy to ignore all test code when navigating the call graph, by filtering textDocument/references[^2] results by file path.
Probably that's more important than reducing the number of files.
It's possible that textDocument/references can be taught to ignore things gated by "cfg(test)", not yet sure. 

Looks like 2 is at odds with 4, but given that every crate could have their own integration test directory anyway,
we can't really get "one directory per workspace" (unless we add `include!` hacks),
so let's live without 4.

As mentioned in #11985, we could put all unit tests in "src/tests/" if we are happy relinquish some rust-analyzer support (completions, and code lenses for running tests):

	#[cfg(test)]
	mod tests { include!("tests/screen.rs"); }

rust-analyzer (and clippy etc.) could probably support this better in future,
though overall, it seems safer to create a real submodule file for this.

So
1. put tests for private APIs in their own submodule file
   (mostly obsoletes the "exposed for testing" exports)
2. put tests for public APIs into "src/tests"

Not sure if we should go all the way with change 2 -- is it the right criteria?
It might be okay to have tests for public APIs in "foo::tests",
and leave things to minimize movement..

Have "crates/printf/" mirror this structure;
in future we want to move more things into such sub-crates, and write [idiomatic integration tests](https://doc.rust-lang.org/rust-by-example/testing/integration_testing.html).

[^1]: https://mastodon.social/@vitaut/114904833700335277
[^2]: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#textDocument_references

</details>


crates/printf/src/arg.rs
```diff
@@ -197,53 +197,4 @@ macro_rules! impl_to_arg_u {
...
-    fn test_negative_to_arg() {
-        assert_eq!((-1_i8).to_arg().as_sint(), Ok(-1));
-        assert_eq!((-1_i16).to_arg().as_sint(), Ok(-1));
-        assert_eq!((-1_i32).to_arg().as_sint(), Ok(-1));
-        assert_eq!((-1_i64).to_arg().as_sint(), Ok(-1));
-
-        assert_eq!((u64::MAX).to_arg().as_sint(), Err(Error::Overflow));
-    }
-}
+mod tests;
```
krobelus: @kerty0 in #11985

 > regarding "pub for testing", what do you think about instead
 > of creating tests module in lib.rs, we include!() test files
 > as submodules? The main disadvantage is subpar Rust Analyzer
 > support (syntax highlighting and completions don't work, though
 > jumping does), which can be worked around by copy-pasting test
 > code. Overall, it keeps test code separated and removes the need
 > for test-only pub.

rust-analyzer completions (and maybe clippy) support could probably be added eventually.
Another concern is, how would it look like with multiple crates:
does every crate have "src/test", or would we put everything into one at the workspace level?
The latter seems wrong if we want to publish our crates but that's not very likely.

Either way, the barrier to entry to writing tests should remain very low.
Same file is the gold standard for that.
What this PR currently proposes `foo/tests.rs` is not much harder,
you'd put things into `mod tests {}` and use `rust-analyzer`'s code action (light bulb) to extract it into a separate file.
Though with the approach in this PR, a significant portion of tests ought to be moved to `src/tests`,
which requires fiddling with imports. Not yet sure if that's worth it.

Assuming that most tests don't need to use private APIs,
I think the `include!()` trick would be fine too if we get the blessing from rust-analyzer folks to add completions.
One option is to go with this PR for now and maybe do that later.

danielrainer:
> tests should not be in the same file as the implementation

I'm don't think that having tests in a separate file is desirable, at least for tests of private functions.

Advantages of tests in the same file as the code they test:
- easiest to access
- satisfies your desiderata 2 and 3, which I fully agree with
- does not require any hacks to work around visibility restrictions, so code is cleaner and tooling works better (macros are really undesirable in combination with rust-analyzer)
- reduces the number of source files
- is the usual approach for Rust code, thus the least surprising

Alleged disadvantages:
- slows down compile times: Tests would be in a cfg-guarded submodule, so I doubt that the impact on compile times is large enough to play a role in our decision making
- separation of test vs non-test dependencies: That's what `[dev-dependencies]` are for, to the extent we even have additional dependencies for tests. What might actually be beneficial for dependency management is splitting up the huge main crate into several crates.
- (these two are the only disadvantages I found mentioned at the mastodon link)
- The reason for 1 (and 4) is to make it easy to ignore all test code when navigating the call graph, by filtering textDocument/references results by file path.: There is `"rust-analyzer.references.excludeTests": true`, which might do what you want. It does not support disabling `#[cfg(test)]` code yet, but fixing that might not be too much effort. I'm not familiar with the rust-analyzer code base, but the author of the feature just forgot to add it (https://github.com/rust-lang/rust-analyzer/pull/16441#issuecomment-2503947097) and the associated PR is fairly small. The associated issue has been open for almost a year and noone seems to have taken an interest in implementing it: https://github.com/rust-lang/rust-analyzer/issues/18573

> we want all tests to live in single directory (src/tests) if possible

Why?

I have not looked much into how integration tests are best structured, especially with multiple crates, but the link you provided suggests that they should be in `tests/`, not `src/tests/`. I'm also not sure if many of our Rust tests even qualify as integration tests.

> It might be okay to have tests for public APIs in "foo::tests"

I think that's what we should do for all rust tests unless we have some specific reason to do something different for a particular test.

I don't care that much about whether `foo::tests` is in `foo.rs` or `foo/tests.rs`, but I think I prefer keeping it in `foo.rs`, to avoid file system clutter. Adding a new directory plus `tests.rs` file for every source file we have tests for results in quite a bit of bloat, making navigation more annoying. If we really want separate files for tests, I'd prefer moving `foo.rs` into `foo/mod.rs`, since that avoids having `foo.rs` and `foo/` as siblings.

krobelus:
> - is the usual approach for Rust code, thus the least surprising

Right. I'm happy to fix my tooling so it also works well on idiomatic Rust projects.

> There is `"rust-analyzer.references.excludeTests": true`,

Nice, I'm using that now with [`#[cfg(test)]` support](https://github.com/rust-lang/rust-analyzer/pull/20901).
I think that's all we need, I don't think we use grep too much.

(It'd probably be more convenient for some LSP clients if excludeTests was an argument to `textDocument/references` like `includeDeclaration` is. Future work)

> > we want all tests to live in single directory (src/tests) if possible
> 
> Why?

Mainly for sorting and filtering file:line locations but yeah, looks like there are other solutions for that.

> I have not looked much into how integration tests are best structured, especially with multiple crates, but the link you provided suggests that they should be in `tests/`, not `src/tests/`. I'm also not sure if many of our Rust tests even qualify as integration tests.

Right, each crate can have integration tests in `tests/`.
Integration tests do not set `cfg(test)`, so that would need another small fix to `rust-analyzer` (and this is an incentive to actually use the `tests/` rather than `src/tests/`).
Most of our Rust tests are probably small enough to be unit tests.
Integration tests get a separate binary per test file, which can theoretically improve compile time and parallelism (we still have some globals affecting test behavior, and unit tests run in a single process, which is why most nontrivial unit tests cannot run in parallel with each other `#[serial]`).

> > It might be okay to have tests for public APIs in "foo::tests"
> 
> I think that's what we should do for all rust tests unless we have some specific reason to do something different for a particular test.

yeah same file seems like the best for almost all our tests

So I guess
1. we should always put tests at the bottom of the file (i.e. don't mix with implementation).
   Not sure if there should be exceptions for very small tests?
2. if tests take up a lot of vertical space, put them in an inline module (`mod tests {}`), to make them easy to skip/collapse
3. we don't care about separating tests based on whether they use public/private interfaces. Probably that kind of stuff is more relevant to library crates.

> If we really want separate files for tests, I'd prefer moving `foo.rs` into `foo/mod.rs`, since that avoids having `foo.rs` and `foo/` as siblings.

yeah that makes sense.
Modules like `src/highlight/tests.rs` have that layout.
Though I wonder if things like `test_is_potential_path` should be moved into `src/highlight/file_tester.rs`?

danielrainer:
> if tests take up a lot of vertical space, put them in an inline module (`mod tests {}`), to make them easy to skip/collapse

We could also put them all into a `tests` module unconditionally. Not sure if it makes much of a relevant difference other than modules being nicer to collapse but having one fewer level of indentation to work with. I doubt that it makes a relevant difference for compile times, but that could be tested.

src/ast.rs
```diff
@@ -2827,11 +2825,111 @@ fn keyword_for_token(tok: TokenType, token: &wstr) -> ParseKeyword {
...
+                    assert!(!same, "Node {} should not be the same as node {}", i, j);
+                }
+            }
+        }
+    }
+}
+
+// Run with cargo +nightly bench --features=benchmark
+#[cfg(feature = "benchmark")]
+#[allow(unused_imports)] // false positive?
```
kerty0: > For some weird reason, the moved benchmarks throw an unused-import warning though.

Moving benchmarks made them lose their #[cfg(test)] attribute. Fix:
```diff
diff --git a/src/ast.rs b/src/ast.rs
index eaf4eda4bf..c61147af19 100644
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -2889,7 +2889,7 @@
 
 // Run with cargo +nightly bench --features=benchmark
 #[cfg(feature = "benchmark")]
-#[allow(unused_imports)] // false positive?
+#[cfg(test)]
 mod bench {
     extern crate test;
     use crate::ast;
diff --git a/src/common.rs b/src/common.rs
index fcfd1774dc..181455833b 100644
--- a/src/common.rs
+++ b/src/common.rs
@@ -2359,8 +2359,9 @@
     }
 }
 
+// Run with cargo +nightly bench --features=benchmark
 #[cfg(feature = "benchmark")]
-#[allow(unused_imports)] // false positive?
+#[cfg(test)]
 mod bench {
     extern crate test;
     use crate::common::bytes2wcstring;
```

src/tests/mod.rs
```diff
@@ -1,121 +1 @@
...
-        });
-        reader_init(false);
-        ScopeGuard::new((), |()| {
-            reader_deinit(false);
-        })
-    }
-
-    pub use serial_test::serial;
-}
+pub mod prelude;
```
kerty0, line 1: Maybe we should replace `src/tests/mod.rs` and `src/tests/prelude.rs` with just `src/tests.rs`?

build_tools/style.fish
```diff
@@ -104,6 +104,9 @@ end
 echo === Running "$green"rustfmt"$normal"
 if set -l -q _flag_check
     if test $all = yes
+        set -l edition_spec string match -r '^edition\s*=.*'
+        test "$($edition_spec <Cargo.toml)" = "$($edition_spec <.rustfmt.toml)"
+        or die "Cargo.toml and .rustfmt.toml use different editions"
```
kerty0: Shouldn't edition check happen before `echo === Running...`?
We could add a check to see if rustfmt needs to be run like with other formatters, and then place the edition check behind it.
Also I find `or die ...` to be more readable than the `if not` conditions.

```diff
diff --git a/build_tools/style.fish b/build_tools/style.fish
index ed4274f7b9..f26305f20f 100755
--- a/build_tools/style.fish
+++ b/build_tools/style.fish
@@ -71,9 +71,8 @@
     end
     echo === Running "$green"fish_indent"$normal"
     if set -l -q _flag_check
-        if not fish_indent --check -- $fish_files
-            die "Fish files are not formatted correctly."
-        end
+        fish_indent --check -- $fish_files
+        or die "Fish files are not formatted correctly."
     else
         fish_indent -w -- $fish_files
     end
@@ -87,41 +86,37 @@
     end
     echo === Running "$green"ruff format"$normal"
     if set -l -q _flag_check
-        if not ruff format --check $python_files
-            die "Python files are not formatted correctly."
-        end
+        ruff format --check $python_files
+        or die "Python files are not formatted correctly."
     else
         ruff format $python_files
     end
 end
 
-if not cargo fmt --version >/dev/null
-    echo
-    echo $yellow'Please install "rustfmt" to style Rust, e.g. via:'
-    echo "rustup component add rustfmt"$normal
-    exit 127
-end
-echo === Running "$green"rustfmt"$normal"
-if set -l -q _flag_check
-    if test $all = yes
-        set -l edition_spec string match -r '^edition\s*=.*'
-        test "$($edition_spec <Cargo.toml)" = "$($edition_spec <.rustfmt.toml)"
-        or die "Cargo.toml and .rustfmt.toml use different editions"
-        if not cargo fmt --all --check
-            die "Rust files are not formatted correctly."
-        end
-    else
-        if set -q rust_files[1]
-            if not rustfmt --check --files-with-diff $rust_files
-                die "Rust files are not formatted correctly."
-            end
-        end
+if test $all = yes; or set -q rust_files[1]
+    if not cargo fmt --version >/dev/null
+        echo
+        echo $yellow'Please install "rustfmt" to style Rust, e.g. via:'
+        echo "rustup component add rustfmt"$normal
+        exit 127
     end
-else
-    if test $all = yes
-        cargo fmt --all
+
+    set -l edition_spec string match -r '^edition\s*=.*'
+    test "$($edition_spec <Cargo.toml)" = "$($edition_spec <.rustfmt.toml)"
+    or die "Cargo.toml and .rustfmt.toml use different editions"
+
+    echo === Running "$green"rustfmt"$normal"
+    if set -l -q _flag_check
+        if test $all = yes
+            cargo fmt --all --check
+        else
+            rustfmt --check --files-with-diff $rust_files
+        end
+        or die "Rust files are not formatted correctly."
     else
-        if set -q rust_files[1]
+        if test $all = yes
+            cargo fmt --all
+        else
             rustfmt $rust_files
         end
     end
```

src/bin/fish.rs
```diff
@@ -513,7 +513,6 @@ fn throwing_main() -> i32 {
     features::set_from_string(opts.features.as_utfstr());
     fish::env_dispatch::read_terminfo_database(EnvStack::globals());
     proc_init();
-    fish::env::misc_init();
```
krobelus, line 516: FWIW I reviewed the last commit with

        git show $commit \
            --color-moved=dimmed-zebra \
            --color-moved-ws=allow-indentation-change

My to-dos:
- Do the same for crates/printf/**/tests.rs
- also for `src/builtins/tests/test_tests.rs`
- Maybe (later) move all top-level tests into a "mod tests" inline module
- I put our two benchmarks below tests ("mod tests" followed by "mod bench").
  Maybe we should use "mod test" for consistency?

krobelus: > also for src/builtins/tests/test_tests.rs

this is the style we've used a bit,
to maximize code sharing while still getting correct locations on failure:

	thread 'builtins::string::escape::test::plain' panicked at src/builtins/string/escape.rs:84:9:

If this hurts compile times, we should change `validate!(...)` to `assert!(validate!(...))`, then the macro can be smaller.

```diff
diff --git a/src/builtins/string/escape.rs b/src/builtins/string/escape.rs
index 1b509bb225..2547c5f77d 100644
--- a/src/builtins/string/escape.rs
+++ b/src/builtins/string/escape.rs
@@ -63,3 +63,27 @@
         }
     }
 }
+
+#[cfg(test)]
+mod test {
+    use crate::builtins::shared::{STATUS_CMD_ERROR, STATUS_CMD_OK};
+    use crate::tests::prelude::*;
+    use crate::validate;
+
+    #[test]
+    #[serial]
+    #[rustfmt::skip]
+    fn plain() {
+        let _cleanup = test_init();
+        validate!(["string", "escape"], STATUS_CMD_ERROR, "");
+        validate!(["string", "escape", ""], STATUS_CMD_OK, "''\n");
+        validate!(["string", "escape", "-n", ""], STATUS_CMD_OK, "\n");
+        validate!(["string", "escape", "a"], STATUS_CMD_OK, "a\n");
+        validate!(["string", "escape", "\x07"], STATUS_CMD_OK, "\\cg\n");
+        validate!(["string", "escape", "\"x\""], STATUS_CMD_OK, "'\"x\"'\n");
+        validate!(["string", "escape", "hello world"], STATUS_CMD_OK, "'hello world'\n");
+        validate!(["string", "escape", "-n", "hello world"], STATUS_CMD_OK, "hello\\ world\n");
+        validate!(["string", "escape", "hello", "world"], STATUS_CMD_OK, "hello\nworld\n");
+        validate!(["string", "escape", "-n", "~"], STATUS_CMD_OK, "\\~\n");
+    }
+}
diff --git a/src/builtins/string/test_helpers.rs b/src/builtins/string/test_helpers.rs
new file mode 100644
index 0000000000..41cae27430
--- /dev/null
+++ b/src/builtins/string/test_helpers.rs
@@ -0,0 +1,33 @@
+use super::string;
+use crate::builtins::shared::BuiltinResultExt;
+use crate::io::IoChain;
+use crate::io::{IoStreams, OutputStream, StringOutputStream};
+use crate::tests::prelude::*;
+use crate::wchar::prelude::*;
+
+#[macro_export]
+macro_rules! validate {
+    ( [$($argv:expr),*], $expected_rc:expr, $expected_out:expr ) => {
+        {
+            use $crate::common::escape;
+            use $crate::wchar::prelude::*;
+            use $crate::builtins::string::test_helpers::string_test;
+            let (actual_out, actual_rc) = string_test(vec![$(L!($argv)),*]);
+            assert_eq!(escape(L!($expected_out)), escape(&actual_out));
+            assert_eq!($expected_rc, actual_rc);
+        }
+    };
+}
+
+pub fn string_test(mut args: Vec<&wstr>) -> (WString, libc::c_int) {
+    let parser = TestParser::new();
+    let mut outs = OutputStream::String(StringOutputStream::new());
+    let mut errs = OutputStream::Null;
+    let io_chain = IoChain::new();
+    let mut streams = IoStreams::new(&mut outs, &mut errs, &io_chain);
+    streams.stdin_is_directly_redirected = false; // read from argv instead of stdin
+
+    let rc = string(&parser, &mut streams, args.as_mut_slice());
+
+    (outs.contents().to_owned(), rc.builtin_status_code())
+}
diff --git a/src/builtins/string.rs b/src/builtins/string.rs
index 79937b9f67..d2ea051653 100644
--- a/src/builtins/string.rs
+++ b/src/builtins/string.rs
@@ -17,6 +17,9 @@
 mod trim;
 mod unescape;
 
+#[cfg(test)]
+mod test_helpers;
+
 macro_rules! string_error {
     (
     $streams:expr,
diff --git a/src/builtins/tests/string_tests.rs b/src/builtins/tests/string_tests.rs
index cbe4181f0a..6af16e7909 100644
--- a/src/builtins/tests/string_tests.rs
+++ b/src/builtins/tests/string_tests.rs
@@ -50,17 +50,6 @@
 
     #[rustfmt::skip]
     let tests = test_cases![
-        (["string", "escape"], STATUS_CMD_ERROR, ""),
-        (["string", "escape", ""], STATUS_CMD_OK, "''\n"),
-        (["string", "escape", "-n", ""], STATUS_CMD_OK, "\n"),
-        (["string", "escape", "a"], STATUS_CMD_OK, "a\n"),
-        (["string", "escape", "\x07"], STATUS_CMD_OK, "\\cg\n"),
-        (["string", "escape", "\"x\""], STATUS_CMD_OK, "'\"x\"'\n"),
-        (["string", "escape", "hello world"], STATUS_CMD_OK, "'hello world'\n"),
-        (["string", "escape", "-n", "hello world"], STATUS_CMD_OK, "hello\\ world\n"),
-        (["string", "escape", "hello", "world"], STATUS_CMD_OK, "hello\nworld\n"),
-        (["string", "escape", "-n", "~"], STATUS_CMD_OK, "\\~\n"),
-
         (["string", "join"], STATUS_INVALID_ARGS, ""),
         (["string", "join", ""], STATUS_CMD_ERROR, ""),
         (["string", "join", "", "", "", ""], STATUS_CMD_OK, "\n"),
```

src/wutil/dir_iter.rs
```diff
@@ -416,84 +421,88 @@ fn test_dir_iter() {
         vec![dirname, regname, fifoname]
     };
```
kerty0, line 422: Missed `#[rustfmt::skip]`

crates/printf/src/lib.rs
```diff
@@ -6,8 +6,10 @@ mod fmt_fp;
 mod printf_impl;
 pub use printf_impl::{Error, FormatString, sprintf_locale};
 pub mod locale;
-pub use locale::{C_LOCALE, EN_US_LOCALE, Locale};
+pub use locale::{C_LOCALE, Locale};
 
+#[cfg(test)]
+pub use locale::EN_US_LOCALE;
```
kerty0, line 12: These reexports are unused and can be removed:
```diff
diff --git a/crates/printf/src/lib.rs b/crates/printf/src/lib.rs
index f230382874..14081eb4bc 100644
--- a/crates/printf/src/lib.rs
+++ b/crates/printf/src/lib.rs
@@ -6,11 +6,8 @@
 mod printf_impl;
 pub use printf_impl::{Error, FormatString, sprintf_locale};
 pub mod locale;
-pub use locale::{C_LOCALE, Locale};
 
 #[cfg(test)]
-pub use locale::EN_US_LOCALE;
-#[cfg(test)]
 mod tests;
 
 /// A macro to format a string using `fish_printf` with C-locale formatting rules.
diff --git a/crates/printf/src/locale.rs b/crates/printf/src/locale.rs
index 0afbb69b68..c46536970b 100644
--- a/crates/printf/src/locale.rs
+++ b/crates/printf/src/locale.rs
@@ -110,7 +110,7 @@
 };
 
 // en_us numeric locale, for testing.
-#[allow(dead_code)]
+#[cfg(test)]
 pub const EN_US_LOCALE: Locale = Locale {
     decimal_point: '.',
     thousands_sep: Some(','),
```

src/wutil/mod.rs
kerty0, line 302: Apart from the new test module, this test module also exists.

crates/printf/src/lib.rs
krobelus: Yeah I think we can remove the re-export (unless any user of `fish-printf` wants it, dunno).
Probably need to keep the `#[cfg(test)]` because it's used in doctests which don't set `cfg(test)` apparently

src/wutil/mod.rs
krobelus: I moved it into a nested module for now.
It's the only one of its kind; I guess it make sense when there's many different tests per function though we could achieve similar grouping with the test function name (or splitting the module/crate).

Issue #11921, opened 2025-10-09
Random Issues With Fish Internal Commands Not Being Found


thetinygoat:
**fish version**: 4.1.1

**OS**: Darwin

**Terminal**:
Ghostty 1.2.1

Version
  - version: 1.2.1
  - channel: stable
Build Config
  - Zig version   : 0.14.1
  - build mode    : builtin.OptimizeMode.ReleaseFast
  - app runtime   : apprt.Runtime.none
  - font engine   : font.main.Backend.coretext
  - renderer      : renderer.generic.Renderer(renderer.Metal)
  - libxev        : kqueue

I don't use any third party customizations

I have not been able to replicate the issue deterministically, it comes on randomly. This time it happened after  i ran brew upgrade, and fish was one of the packages that was updated. I thought maybe it happened due to underlying files being changed. 
But when i went through the update on my second machine which also has the same OS and terminal and same version was updated i didn't face the same issue, although i have faced the same issue on my second machine earlier.

It gets fixed when i open a new split or a new tab.

Here's the type of errors that i see when this happens.
```shell

fish: Unknown command: fish_job_summary
/opt/homebrew/Cellar/fish/4.1.1/share/fish/functions/__fish_print_pipestatus.fish (line 1): Unknown command: fish_status_to_signal
fish_status_to_signal $argv | string join "$sep"
^~~~~~~~~~~~~~~~~~~~^
in command substitution
        called on line 27 of file /opt/homebrew/Cellar/fish/4.1.1/share/fish/functions/__fish_print_pipestatus.fish
in function '__fish_print_pipestatus' with arguments '\[ \] \| \e\[31m \e\[31m\e\[1m 137'
        called on line 1 of file /opt/homebrew/Cellar/fish/4.1.1/share/fish/functions/fish_prompt.fish
in command substitution
        called on line 31 of file /opt/homebrew/Cellar/fish/4.1.1/share/fish/functions/fish_prompt.fish
in function 'fish_prompt'
in command substitution
/opt/homebrew/Cellar/fish/4.1.1/share/fish/functions/__fish_print_pipestatus.fish (line 27): Unknown command
        set -l last_pipestatus_string (fish_status_to_signal $argv | string join "$sep")
                                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
[...]

PR #12000, opened 2025-10-27
Use embedded files also in installed builds

krobelus:
Turn on the "embed-data" feature by default, to load our functions/completions directly from the fish binary.
Mainly to remove weird behavior in running shells when upgrading (#11921),
but also as a part of the effort to reduce differences between standalone and installed mode.

I wonder if we should already remove the ability to turn off `embed-data`.

(Draft state; this PR still includes some unrelated changes;
happy to accomodate review better but I still need to do self-review)


krobelus:
this has been merged, I forgot to close

PR #12008, opened 2025-10-30
io: retry writes on interrupts

danielrainer:
Interrupts should not prevent output from being completed, unless they
result in control flow not continuing where it was interrupted.

Return results containing errors, so callers have more options for error
handling. Returning the number of bytes written on success is no longer
necessary, since success now means that all input bytes were written.

In `src/io.rs` some changes are needed since `EINTR` is no longer a
possible return value. I also removed the comments on the old code and
the `sigcheck` code. Not sure if this should be kept.

### DIFF
--- a/src/io.rs
+++ b/src/io.rs
@@ -9,13 +9,11 @@ use crate::nix::isatty;
 use crate::path::path_apply_working_directory;
 use crate::proc::JobGroupRef;
 use crate::redirection::{RedirectionMode, RedirectionSpecList};
-use crate::signal::SigChecker;
 use crate::terminal::Output;
-use crate::topic_monitor::Topic;
 use crate::wchar::prelude::*;
 use crate::wutil::{perror, perror_io, wdirname, wstat, wwrite_to_fd};
 use errno::Errno;
-use libc::{EAGAIN, EINTR, ENOENT, ENOTDIR, EPIPE, EWOULDBLOCK, STDOUT_FILENO};
+use libc::{EAGAIN, EINTR, ENOENT, ENOTDIR, EWOULDBLOCK, STDOUT_FILENO};
 use nix::fcntl::OFlag;
 use nix::sys::stat::Mode;
 use once_cell::sync::Lazy;
@@ -758,9 +756,6 @@ pub struct FdOutputStream {
     /// The file descriptor to write to.
     fd: RawFd,
 
-    /// Used to check if a SIGINT has been received when EINTR is encountered
-    sigcheck: SigChecker,
-
     /// Whether we have received an error.
     errored: bool,
 }
@@ -768,29 +763,15 @@ impl FdOutputStream {
     /// Construct from a file descriptor, which must be nonegative.
     pub fn new(fd: RawFd) -> Self {
         assert!(fd >= 0, "Invalid fd");
-        FdOutputStream {
-            fd,
-            sigcheck: SigChecker::new(Topic::sighupint),
-            errored: false,
-        }
+        FdOutputStream { fd, errored: false }
     }
 
     fn append(&mut self, s: &wstr) -> bool {
         if self.errored {
             return false;
         }
-        if wwrite_to_fd(s, self.fd).is_none() {
-            // Some of our builtins emit multiple screens worth of data sent to a pager (the primary
-            // example being the `history` builtin) and receiving SIGINT should be considered normal and
-            // non-exceptional (user request to abort via Ctrl-C), meaning we shouldn't print an error.
-            if errno::errno().0 == EINTR && self.sigcheck.check() {
-                // We have two options here: we can either return false without setting errored_ to
-                // true (*this* write will be silently aborted but the onus is on the caller to check
-                // the return value and skip future calls to `append()`) or we can flag the entire
-                // output stream as errored, causing us to both return false and skip any future writes.
-                // We're currently going with the latter, especially seeing as no callers currently
-                // check the result of `append()` (since it was always a void function before).
-            } else if errno::errno().0 != EPIPE {
+        if let Err(e) = wwrite_to_fd(s, self.fd) {
+            if e.kind() != std::io::ErrorKind::BrokenPipe {
                 perror("write");
             }
             self.errored = true;
--- a/src/wutil/mod.rs
+++ b/src/wutil/mod.rs
@@ -373,15 +373,33 @@ pub fn wrename(old_name: &wstr, new_name: &wstr) -> libc::c_int {
     unsafe { libc::rename(old_narrow.as_ptr(), new_narrow.as_ptr()) }
 }
 
-pub fn write_to_fd(input: &[u8], fd: RawFd) -> nix::Result<usize> {
-    nix::unistd::write(unsafe { BorrowedFd::borrow_raw(fd) }, input)
+/// Writes the input bytes to the provided file descriptor, retrying on interrupts.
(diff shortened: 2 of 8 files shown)


src/io.rs
```diff
@@ -755,39 +753,22 @@ pub struct FdOutputStream {
...
     }
 
     fn append(&mut self, s: &wstr) -> bool {
         if self.errored {
             return false;
         }
-        if wwrite_to_fd(s, self.fd).is_none() {
-            // Some of our builtins emit multiple screens worth of data sent to a pager (the primary
-            // example being the `history` builtin) and receiving SIGINT should be considered normal and
-            // non-exceptional (user request to abort via Ctrl-C), meaning we shouldn't print an error.
```
krobelus, line 785: Technically, the point about the history builtin should still be valid.
Maybe not for history in practice because we truncate it at 256k elements,
but I'd think that we should be able to cancel mid-output something like

	fish -c 'set x "$(seq 2000000)"; echo $x'

I tried that and it's very unreliable; cancellation only works like 1/5 times at most,
maybe the signal is somehow not delivered fast enough

src/io.rs
```diff
@@ -755,39 +753,22 @@ pub struct FdOutputStream {
...
-            if errno::errno().0 == EINTR && self.sigcheck.check() {
```
krobelus, line 786: It's a shame we can't use sigaction()'s SA_RESTART,
which would add automatic retrying to things like write(3)
but not to things that are supposed to be interruptible like poll(3) or sleep(3).

I guess Rust's stdlib does pretty much the same.

In future, if we add a slightly different event loop or a separate thread for handling input (and UI output),
then we could read ctrl-c from the TTY rather than relying on signals.

src/wutil/mod.rs
```diff
@@ -497,69 +497,71 @@ pub fn wrename(old_name: &wstr, new_name: &wstr) -> libc::c_int {
...
-            let Ok(amt) = write_to_fd(buf, fd) else {
-                return false;
-            };
-            *total_written += amt;
-            assert!(amt <= buf.len(), "Wrote more than requested");
-            buf = &buf[amt..];
-        }
-        true
+    fn do_write(fd: RawFd, total_written: &mut usize, buf: &[u8]) -> std::io::Result<()> {
+        write_to_fd(buf, fd)?;
```
krobelus, line 413: (unrelated) I guess `write_to_fd` is equivalent to `std::fs::File::write_all()`,
not sure if it's worth keeping our wrapper

danielrainer: As far as I'm aware, `write_all` does not work with raw file descriptors, otherwise I would have used it. Since we need to write to stdout and stderr, we can't use a `File` in all cases.

src/io.rs
danielrainer: So what do you think we should do with this PR? Merge it as is? Add some TODO comments? Come up with a better implementation?

src/io.rs
krobelus: there are two non-test uses of wwrite_to_fd
1. the for flogging - for that one, retrying seems fine (a corrupted log file is no big deal but still better to avoid it)
2. the one in FdOutputStream, which is instantiated for
   1. the fish_indent binary
   2. the fish_key_reader binary
   3. builtins writing to stdout with no redirections
   4. builtins writing to a file
   5. builtins writing to a pipe

It seems like this change is a good idea,
since it reduces corruption and we basically never actually write megabytes of data in a single wwrite_to_fd() call.
But we should probably move the ctrl-c cancellation logic to all the relevant users of `FdOutputStream` (i.e. builtins)

I wonder if we really want ctrl-c to potentially stop builtin output.
I guess cancellation works for external programs, so emulating that seems correct.
It means that ctrl-c can potentially cancel prompts, event handlers etc, which might be surprising,
but that's an unrelated issue.

`builtin -n` lists 68 builtins but most should not ever produce a lot of output.
So we can go through some of them.
Loops (`while` etc) are not relevant because fish already checks for cancellation at each loop iteration.
So it's mostly about builtins that loop internally.

Something like `echo` might be fine in practice because something like `n=30 echo (seq $n)(seq $n)(seq $n)(seq $n)` will simply error because the command substitution is too large.
It might be interesting to find out what's the maximum number of megabytes we can make fish pass to one `wwrite_to_fd()` call.
foo
But if that's too complicated we can squint and ignore it, or I can take a look later.

For things like the `string` commands, we might want to check for cancellation everytime after processing an argument?

I'm not saying we necessarily need to reimplement cancellation for everything today,
but we should keep in mind, we might want it eventually.

```diff
diff --git a/src/builtins/string/repeat.rs b/src/builtins/string/repeat.rs
index 5d540a6168..c10f6feee0 100644
--- a/src/builtins/string/repeat.rs
+++ b/src/builtins/string/repeat.rs
@@ -154,6 +154,8 @@
                         i -= chunk.len();
                     }
                     chunk.clear();
+                    // TODO use SigChecker to check for cancellation. Not sure if we should stop mid-argument though.
+                    // in case we have something like string repeat 10000000 foo\n
                 }
             }
 
diff --git a/src/history.rs b/src/history.rs
index 2e9ed58ab6..f187ede7a9 100644
--- a/src/history.rs
+++ b/src/history.rs
@@ -1476,6 +1476,8 @@
             }
 
             if !streams.out.append(item) {
+                // TODO adjust the comment (we can no longer fail due to ctrl-c)
+                // but also use a SigChecker or equivalent here to exit early
                 // Don't force an error if output was aborted (typically via Ctrl-C/SIGINT); just don't
                 // try writing any more.
                 output_error = true;
```

danielrainer: I added code to address cancellation at the two locations in your diff. There are a lot more users of this code, but I think at least adding some examples of how cancellation can be handled with the new code is helpful.

krobelus: We can cancel this:

	cat **.fish | command fish_indent

but not this (even without this change!)

	cat **.fish | builtin fish_indent

not yet sure why,
but it seems like we should allow cancelling this.

I'm not sure what's the best, but an obvious way is the existing one:
allow to (at least optionally) have `FdOutputStream` abandon itself after EINTR.

It doesn't help that we fail to propagate the error here:

	streams.out.append(bytes2wcstring(&colored_output));

we wrongly keep going after ctrl-c (i.e. we look at the next file).

Propagating this error everywhere where we could have a fd output stream might be a lot of work.

---

> Interrupts should not prevent output from being completed, unless they
> result in control flow not continuing where it was interrupted.

So given

	echo foo
	echo bar

if (hypothetically) the first one is interrupted, we don't print the second one AFAIK,
and as a consequence of your implication, the first echo statement should (or can?) actually exit early on EINTR.

I wonder if there is a nice way to make this true for all builtins.
When should ctrl-c not affect control flow?

I haven't checked other builtins yet.
This change is hard to prove correct, so I'd be careful about doing this.

danielrainer:
The Fluent PR no longer depends on this, so there is no immediate need to get these changes merged. If you'd like, we can just close this PR. I don't think I'll look much more into this any time soon.

krobelus:
> The Fluent PR no longer depends on this, so there is no immediate
> need to get these changes merged. If you'd like, we can just close
> this PR. I don't think I'll look much more into this any time soon.

there's different opinions on how to manage the PR/issue list.
I don't really care either way (since I can additionally keep my private todo-list)

I'd keep it open because at least the bug (?) about fish_indent not
being cancelable (which is not reported elsewhere yet AFAIK) should
have an obvious fix, and then we can figure out if the other things.

PR #12011, opened 2025-10-31
gettext: fall back to all language variants

danielrainer:
If a language is specified using only the language code, without a region identifier, assume that the user prefers translations from any variant of the language over the next fallback option. For example, when a user sets `LANGUAGE=zh:pt`, assume that the user prefers both `zh_CN` and `zh_TW` over the next fallback option. The precedence of the different variants of a language will be arbitrary. In this example, with the current set of relevant available catalogs (`pt_BR`, `zh_CN`, `zh_TW`), the effective precedence will be either `zh_CN:zh_TW:pt_BR` or `zh_TW:zh_CN:pt_BR`.

Users who want more control over the order can
specify variants to get the results they want.
For example:
- `LANGUAGE=zh_TW:zh:pt` will result in `zh_TW:zh_CN:pt_BR`.
- `LANGUAGE=zh_CN:pt:zh` will result  in `zh_CN:pt_BR:zh_TW`.
- `LANGUAGE=zh_CN:pt` will result  in `zh_CN:pt_BR`. English is always used as the last fallback.

This approach (like the previous approach) differs from GNU gettext semantics, which map region-less language codes to on specific "default" variant of the language, without specifying how this default is chosen. We want to avoid making such choices and believe it is better to utilize translations from all language variants we have available when users do not explicitly specify their preferred variant. This way, users have an easier time discovering localization availability, and can be more explicit in their preferences if they don't like the defaults. If there are conflicts with gettext semantics, users can also set locale variables without exporting them, so fish uses different values than its child processes.

### DIFF
--- a/doc_src/language.rst
+++ b/doc_src/language.rst
@@ -1836,7 +1836,16 @@ In UNIX, these are made up of several categories. The categories used by fish ar
 
    This is treated like :envvar:`LC_MESSAGES` except that it can hold multiple values,
    which allows to specify a priority list of languages for translation.
-   It's a :ref:`PATH variable <variables-path>`, like in `GNU gettext <https://www.gnu.org/software/gettext/manual/html_node/The-LANGUAGE-variable.html>`__. 
+   It's a :ref:`PATH variable <variables-path>`, like in `GNU gettext <https://www.gnu.org/software/gettext/manual/html_node/The-LANGUAGE-variable.html>`__.
+
+   Language identifiers without a region specified (e.g. ``zh``) result in all available variants of this language being tried in arbitrary order.
+   In this example, we might first look for messages in the ``zh_CN`` catalog, followed by ``zh_TW``, or the other way around.
+   This is different from GNU gettext, which uses a "default" variant of the language instead.
+   If you prefer a certain variant, specify it earlier in the list,
+   e.g. ``zh_TW:zh`` if your preferred language is ``zh_TW``, and you prefer any other variants of ``zh`` over the English default.
+   If ``zh_TW`` is the only variant of ``zh`` you want,
+   specifying ``zh_TW`` in the ``LANGUAGE`` variable will result in messages which are not available in ``zh_TW`` being displayed in English.
+
    See also :doc:`builtin _ (underscore) <cmds/_>`.
 
 .. envvar:: LC_ALL
--- a/src/wutil/gettext.rs
+++ b/src/wutil/gettext.rs
@@ -7,7 +7,7 @@ use once_cell::sync::Lazy;
 
 #[cfg(feature = "localize-messages")]
 mod gettext_impl {
-    use std::sync::Mutex;
+    use std::{collections::HashSet, sync::Mutex};
 
     use once_cell::sync::Lazy;
 
@@ -16,20 +16,18 @@ mod gettext_impl {
 
     use crate::env::{EnvStack, Environment};
 
-    /// Tries to find a catalog for `language`.
+    /// Tries to find catalogs for `language`.
     /// `language` must be an ISO 639 language code, optionally followed by an underscore and an ISO
     /// 3166 country/territory code.
-    /// Always prefers the catalog with the exact same name as `language` if it exists.
+    /// Uses the catalog with the exact same name as `language` if it exists.
     /// If a country code is present (`ll_CC`), only the catalog named `ll` will be considered as a fallback.
-    /// If no country code is present (`ll`), an arbitrary catalog whose name starts with `ll_`
-    /// will be used as a fallback, if one exists.
-    /// If there is a catalog for the language, then `Some(catalog)` will be returned.
-    /// `None` will be returned if no variant of the language has localizations.
-    fn find_existing_catalog(language: &str) -> Option<Catalog> {
+    /// If no country code is present (`ll`), all catalogs whose names start with `ll_` will be used in
+    /// arbitrary order.
+    fn find_existing_catalogs(language: &str) -> Vec<(String, Catalog)> {
         // Try the exact name first.
         // If there already is a corresponding catalog return the language.
         if let Some(catalog) = CATALOGS.get(language) {
-            return Some(catalog);
+            return vec![(language.to_owned(), catalog)];
         }
         let language_without_country_code =
             language.split_once('_').map_or(language, |(ll, _cc)| ll);
@@ -38,17 +36,21 @@ mod gettext_impl {
             // Note that it is important to include the underscore in the pattern, otherwise `ll` might
             // fall back to `llx_CC`, where `llx` is a 3-letter language identifier.
             let ll_prefix = format!("{language}_");
+            let mut lang_catalogs = vec![];
             for (&lang_name, &catalog) in CATALOGS.entries() {
                 if lang_name.starts_with(&ll_prefix) {
-                    return Some(catalog);
+                    lang_catalogs.push((lang_name.to_owned(), catalog));
                 }
             }
-            // No localizations for the language (and any regional variations) exist.
-            None
+            lang_catalogs
         } else {
             // If `language` contained a country code, we only try to fall back to a catalog
             // without a country code.
-            CATALOGS.get(language_without_country_code).copied()
+            if let Some(catalog) = CATALOGS.get(language_without_country_code) {
+                vec![(language_without_country_code.to_owned(), catalog)]
+            } else {
+                vec![]
+            }
         }
     }
 
@@ -143,10 +145,13 @@ mod gettext_impl {
 
     /// Implementation of the function with the same name in super.
     pub(super) fn update_locale_from_env(vars: &EnvStack) {
+        let mut seen_languages = HashSet::new();
         let mut language_precedence = LANGUAGE_PRECEDENCE.lock().unwrap();
         *language_precedence = get_language_preferences_from_env(vars)
-            .iter()
-            .filter_map(|lang| find_existing_catalog(lang))
+            .into_iter()
+            .flat_map(|lang| find_existing_catalogs(&lang))
+            .filter(|(lang, _)| seen_languages.insert(lang.to_owned()))
+            .map(|(_, catalog)| catalog)
             .collect();
     }
 }


src/wutil/gettext.rs
```diff
@@ -18,20 +18,18 @@ mod gettext_impl {
...
+    /// Uses the catalog with the exact same name as `language` if it exists.
     /// If a country code is present (`ll_CC`), only the catalog named `ll` will be considered as a fallback.
-    /// If no country code is present (`ll`), an arbitrary catalog whose name starts with `ll_`
-    /// will be used as a fallback, if one exists.
-    /// If there is a catalog for the language, then `Some(catalog)` will be returned.
-    /// `None` will be returned if no variant of the language has localizations.
-    fn find_existing_catalog(language: &str) -> Option<Catalog> {
+    /// If no country code is present (`ll`), all catalogs whose names start with `ll_` will be used in
+    /// arbitrary order.
+    fn find_existing_catalog(language: &str) -> Vec<(String, Catalog)> {
```
krobelus: maybe change function name to plural as well?

src/wutil/gettext.rs
```diff
@@ -145,10 +147,13 @@ mod gettext_impl {
 
     /// Implementation of the function with the same name in super.
     pub(super) fn update_locale_from_env(vars: &EnvStack) {
+        let mut seen_languages = HashSet::new();
```
krobelus, line 148: FWIW `Vec` should be more cache-friendly (i.e. faster) than `HashSet` until we have like thousands of languages.

src/wutil/gettext.rs
```diff
@@ -145,10 +147,13 @@ mod gettext_impl {
...
         let mut language_precedence = LANGUAGE_PRECEDENCE.lock().unwrap();
         *language_precedence = get_language_preferences_from_env(vars)
-            .iter()
-            .filter_map(|lang| find_existing_catalog(lang))
+            .into_iter()
+            .flat_map(|lang| find_existing_catalog(&lang))
+            .filter(|(lang, _)| seen_languages.insert(lang.to_owned()))
```
krobelus, line 153: This PR addresses all open questions in l10n: Add zh-TW translation (PR #11967), right?
Then I can delete that from my inbox.

krobelus (approved):

src/wutil/gettext.rs
danielrainer: Yes, this PR is intended to address the issue of variant-less languages properly now that we have more than one variant for a language (zh)

danielrainer:
I added some docs for `LANGUAGE` to describe our interpretation of the variable.

src/wutil/gettext.rs
danielrainer: Maybe. I don't think it makes a big difference here, since this code does not run a lot and only on a small number of elements. I used a set because it has a more convenient interface.

krobelus: yeah it definitely doesn't make a difference.
But I guess it's good practice to do the better thing.

> I used a set because it has a more convenient interface.

Unlike in C++, Rust's Vec does have a `contains()` method (actually slices do),
so it should be only about changing `insert()` to `push()`

doc_src/language.rst
```diff
@@ -1836,7 +1836,16 @@ In UNIX, these are made up of several categories. The categories used by fish ar
 
    This is treated like :envvar:`LC_MESSAGES` except that it can hold multiple values,
    which allows to specify a priority list of languages for translation.
-   It's a :ref:`PATH variable <variables-path>`, like in `GNU gettext <https://www.gnu.org/software/gettext/manual/html_node/The-LANGUAGE-variable.html>`__. 
+   It's a :ref:`PATH variable <variables-path>`, like in `GNU gettext <https://www.gnu.org/software/gettext/manual/html_node/The-LANGUAGE-variable.html>`__.
+
```
krobelus, line 1840: Hmm we're inconsistent about "PATH variable" vs. "Path variable"..

doc_src/language.rst
```diff
@@ -1836,7 +1836,16 @@ In UNIX, these are made up of several categories. The categories used by fish ar
...
+   Language identifiers without a region specified (e.g. ``zh``) result in all available variants of this language being tried in arbitrary order.
+   In this example, we might first look for messages in the ``zh_CN`` catalog, followed by ``zh_TW``, or the other way around.
```
krobelus, line 1842: Looks good.

FWIW the reason I use this one-sentence-per-line style partly to make diffs easier to read (even with word highlighting, default diffs are line-based). (https://nick.groenen.me/notes/one-sentence-per-line/)
but also because my editor doesn't have great support for soft-wrapping yet, which is not a good reason so we can change it if needed.

doc_src/language.rst
```diff
@@ -1836,7 +1836,16 @@ In UNIX, these are made up of several categories. The categories used by fish ar
...
+   This is different from GNU gettext, which uses a "default" variant of the language instead.
+   If you prefer a certain variant, specify it earlier in the list,
+   e.g. ``zh_TW:zh`` if your preferred language is ``zh_TW``, and you prefer any other variants of ``zh`` over the English default.
```
krobelus, line 1845: I guess using `zh` is sort of bad practice but it's not like we're encouraging its use here.
(it does not even help that much for discoverability; it's relatively easier for users to list the few variants explicitly)

src/wutil/gettext.rs
danielrainer: > it should be only about changing `insert()` to `push()`

No, `insert` returns whether the element was already contained before, so its return value can be used for filtering. With a `Vec`, we'd need `push` and `contains`, with `push` running only if `contains` returns false.

I also don't understand why you think `Vec` would be more cache-friendly. I have not looked into the implementations in detail, but I would expect that both `HashSet` and `Vec` will read the string to be inserted.

The `HashSet` then computes its hash and only has to check the corresponding location in the hash table. If that's empty, no additional memory access is needed, if it contains a value, that string needs to be read, and if it's not equal to the one being inserted, collision handling needs to happen, but that's hopefully relatively rare and does should not involve many additional memory accesses.

A `Vec` would have to access every single string added so far. The pointers to the strings are stored sequentially, but reading each string's content needs a memory access. (A full traversal is not always needed. If a string equal to the one being inserted is found traversal can be aborted early, but that does not make much of a difference to the general argument.)

So for types whose data is not fully contained in the `Vec` (i.e. non-`Copy` types) I would expect the `HashSet` to perform better, at least in terms of the number of non-sequential memory accesses. Maybe other differences (like calculating the hash function, or how the hash table is grown) result in `Vec` being faster in practice. For the data sizes we work with, everything should be served from L1 anyway, and there just aren't that many languages, so I think it's pretty pointless to start micro-benchmarking this.

The `HashSet` interface is more elegant, has better asymptotic performance, probably has at least not worse performance in practice, and the data type makes it clear that we have unique entries, so I think changing this to `Vec` would be a mistake.

doc_src/language.rst
danielrainer: > to make diffs easier to read

Yes, that's also my main reason, in addition to slightly more convenient navigation. I also don't have soft wrapping set up for RST files in my editor, so I did it manually for some long sentences at locations which felt like they make some semantic sense. I'm not a huge fan of non-semantic wrapping since it can make diffs more annoying, but if it only happens for single sentences it's ok. I could look into setting up soft wrapping if that results in text you like better than no wraps or intra-sentence somewhat semantic manual wraps.

doc_src/language.rst
danielrainer: Yeah, that's why I proposed completely dropping support for this a while ago when writing our gettext runtime implementation. The main problem with dropping support is that users who use this syntax might not realize that localization is available in fish at all, so that's where I think it helps with discoverability. I guess most languages don't have many variants that have an associated region code, so listing all potential ones is not much effort, but might be a bit inconvenient, especially interactively (not sure if anyone does that, I've only done it to test fish's localization code).

Relatedly, one could also argue (and I have before) that our catalogs should not use this syntax, to make it clear for translators which variant should be used in a particular file. If we made that change, having variant-less entries in `LANGUAGE` allows for more translations being shown to users. E.g. if we rename our `de` catalog to `de_DE`, I would argue that our fallback logic should use this catalog when a user specifies `LANGUAGE=de` but not when they specify `LANGUAGE=de_AT`.

There are some tradeoffs involved, but I think what we have with this PR works decently well and is largely compatible with `LANGUAGE` being set according to GNU gettext semantics. (In the sense that users should get translations with our implementation whenever they would get them from GNU gettext, they just might be from a different language variant. They might also get translations from us where GNU gettext would not provide them, but I think that's fine.)

src/wutil/gettext.rs
krobelus: Right, `HashSet` is the semantically correct type; if anything we'd define a `SmallSet` wrapper.

> The HashSet then computes its hash and only has to check the corresponding location in the hash table
> So for types whose data is not fully contained in the Vec (i.e. non-Copy types) I would expect the HashSet to perform better, at least in terms of the number of non-sequential memory accesses.

Yeah that makes sense, HashSet seems better

> I also don't understand why you think Vec would be more cache-friendly

Because I thought a hash map might use more pointer indirections in common cases (specifically if the requested key is present). But I'm not sure if it's actually more indirections than vector.

> reading each string's content needs a memory access.

I forgot that Rust doesn't have SSO (probably for the better).

(Which suggests that a minor improvement is possible: `seen_languages` could be `HashSet<&'a str>` where `'a` is the lifetime of `get_language_preferences_from_env(vars)`. Not sure if that's convenient though)

doc_src/language.rst
krobelus: The status quo is fine; we could maybe issue a warning if LANGUAGE is ambiguous like `zh` but I'm not sure when to warn without annoying users. Maybe some sort of health-check command.
We already have `fish_delta` which is neat but we haven't used it too much.

> I would argue that our fallback logic should use this catalog when a user specifies LANGUAGE=de but not when they specify LANGUAGE=de_AT.

Yeah if we rename `de.po` to `de_DE.po` we want to copy it to `de_AT.po` and `de_CH.po`, and maybe find a nice way to share the common parts (unless we have enough people to complete and maintain all the translations).

doc_src/language.rst
krobelus: breaking after comma is still semantic, and helps keep most line lengths relatively short,
so I think that makes sense. Dunno if it's best practice.

> I  also don't have soft wrapping set up for RST files

I have a shortcut to toggle softwrapping but I leave it disabled most of the time.

src/wutil/gettext.rs
danielrainer: The strings stored in the set are the ones returned from `find_existing_catalogs`, which do not live long enough for us to put references to them into the set, at least without restructuring the code. I expect the compiler to be smart enough to avoid cloning on the `to_owned` call. If that's the case, there is no benefit to storing references instead of the owned string in the set. And even if the compiler doesn't do that, I don't think it's particularly useful to spend much effort on optimizing it manually.

doc_src/language.rst
danielrainer: I generally don't like the term much. We use it to denote implicit colon-separated lists, of which `PATH` is a well-known example, but variables like `LANGUAGE` don't have much to do with paths, so maybe we can come up with better terminology.

PR #12014, opened 2025-10-31
Fix nightly clippy lints

danielrainer:
Stacked on #12013

### DIFF
--- a/src/autoload.rs
+++ b/src/autoload.rs
@@ -212,7 +212,7 @@ impl Autoload {
     /// This is exposed for testing.
     #[cfg(test)]
     fn invalidate_cache(&mut self) {
-        self.cache = Box::new(AutoloadFileCache::with_dirs(self.cache.dirs().to_owned()));
+        *self.cache = AutoloadFileCache::with_dirs(self.cache.dirs().to_owned());
     }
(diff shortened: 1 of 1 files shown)


src/autoload.rs
```diff
@@ -212,7 +212,7 @@ impl Autoload {
     /// This is exposed for testing.
     #[cfg(test)]
     fn invalidate_cache(&mut self) {
-        self.cache = Box::new(AutoloadFileCache::with_dirs(self.cache.dirs().to_owned()));
+        *self.cache = AutoloadFileCache::with_dirs(self.cache.dirs().to_owned());
```
krobelus, line 215: interestingly, the Box was never helpful.
It's a small object, and the comment justifies it as

    /// This is a unique_ptr because want to change it if the value of our environment variable
    /// changes. This is never null (but it may be a cache with no paths).

which would have make sense in C++ if it was modified in `const` methods, because const-propagation is [opt-in](https://en.cppreference.com/w/cpp/experimental/propagate_const.html).

But that wasn't actually used, and it would have been better to declare it as `mutable autoload_file_cache_t`;

No need to update, I already have that change queued.

Issue #11184, opened 2025-02-28
`abbr --add --command` conflicts with `abbr`


zzhaolei:
```shell
❯ fish --version
fish, version 4.0.0
```

current:
```
❯ abbr l "ls -alh"
  abbr --add --command jj l "log"
❯ l # error unknown command
fish: Unknown command: l
❯ jj l # work. expand: jj log
# jj output

# Another
❯ abbr --add --command jj l "log"
  abbr l "ls -alh"
❯ l # work. expand: ls -alh
# ls output
❯ jj l # error unknown command
error: unrecognized subcommand 'l'
```

expected:
```shell
❯ abbr l "ls -alh"
  abbr --add --command jj l "log"
❯ l # should work. expand: ls -alh
# Output
❯ jj l # work. expand: jj log

# Another
❯ abbr --add --command jj l "log"
  abbr l "ls -alh"
❯ l # work. expand: ls -alh
# ls output
❯ jj l # should work. expand: jj log
# jj output
```

zzhaolei:
`abbr --add --command jj push  "git push"`

`push` was expanded twice:
```shell
❯ jj git git push 
```

expect:
```shell
❯ jj git push 
```

zanchey:
The current implementation means you are overwriting a previous abbreviation, not creating a second one. I think it's reasonable to support that.

I can't reproduce the second issue in a clean shell.

zzhaolei:
jj push will expand to `jj git push |`. If you delete the space before the cursor and then press space again, it will reproduce the issue.

`|` means the cursor

zanchey:
Oh, right, recursive expansion. That's probably worth keeping?

zzhaolei:
> Oh, right, recursive expansion. That's probably worth keeping?

I don’t think it should be kept. `--command jj push` and `jj git push` are fundamentally different, and `jj git push` should not be expanded to `jj git git push`.

faho:
In that case, you'll have to make a function and handle it yourself - use `abbr --function` and return 1 if there already is a "git push" in the commandline.

Or use a different abbr trigger.

There's no way abbr can tell that *this* push isn't like *that other* push.

zzhaolei:
> jj push will expand to `jj git push |`. If you delete the space before the cursor and then press space again, it will reproduce the issue.
> 
> `|` means the cursor

When you press Enter after typing jj push, it will also expand to jj git git push.

[...]

PR #12021, opened 2025-11-04
Fix abbr --command conflicts

SandWoodJones:
## Description

Fixes https://github.com/fish-shell/fish-shell/issues/11184

## TODOs:

### DIFF
--- a/doc_src/cmds/abbr.rst
+++ b/doc_src/cmds/abbr.rst
@@ -10,8 +10,8 @@ Synopsis
 
     abbr --add NAME [--position command | anywhere] [-r | --regex PATTERN] [-c | --command COMMAND]
                     [--set-cursor[=MARKER]] ([-f | --function FUNCTION] | EXPANSION)
-    abbr --erase NAME ...
-    abbr --rename OLD_WORD NEW_WORD
+    abbr --erase [ [-c | --command COMMAND]... ] NAME ...
+    abbr --rename [ [-c | --command COMMAND]... ] OLD_WORD NEW_WORD
     abbr --show
     abbr --list
     abbr --query NAME ...
@@ -136,9 +136,10 @@ Other subcommands
 
 ::
 
-    abbr --rename OLD_NAME NEW_NAME
+    abbr --rename [ [-c | --command COMMAND]... ] OLD_NAME NEW_NAME
 
-Renames an abbreviation, from *OLD_NAME* to *NEW_NAME*
+Renames an abbreviation, from *OLD_NAME* to *NEW_NAME*.
+For command-specific abbreviations, the ``--command`` options must be provided to disambiguate which abbreviation to rename.
 
 ::
 
@@ -154,9 +155,10 @@ Prints the names of all abbreviation
 
 ::
 
-    abbr [-e | --erase] NAME
+    abbr [-e | --erase] [ [-c | --command COMMAND]... ] NAME ...
 
-Erases the abbreviation with the given name
+Erases the abbreviation with the given name.
+For command-specific abbreviations, the ``--command`` options must be provided to disambiguate which abbreviation to rename.
 
 ::
 
--- a/src/abbrs.rs
+++ b/src/abbrs.rs
@@ -99,9 +99,11 @@ impl Abbreviation {
         if !self.matches_position(position) {
             return false;
         }
+
         if !self.commands.is_empty() && !self.commands.contains(&command.to_owned()) {
             return false;
         }
+
         match &self.regex {
             Some(r) => r
                 .is_match(token.as_char_slice())
@@ -211,45 +213,44 @@ impl AbbreviationSet {
             let index = self
                 .abbrs
                 .iter()
-                .position(|a| a.name == abbr.name)
-                .expect("Abbreviation not found though its name was present");
+                .position(|a| a.name == abbr.name && a.commands == abbr.commands);
 
-            self.abbrs.remove(index);
+            if let Some(idx) = index {
+                // Exact match found (name + commands), delete old abbr.
+                self.abbrs.remove(idx);
+            }
         }
         self.abbrs.push(abbr);
     }
 
     /// Rename an abbreviation. This asserts that the old name is used, and the new name is not; the
     /// caller should check these beforehand with has_name().
-    pub fn rename(&mut self, old_name: &wstr, new_name: &wstr) {
-        let erased = self.used_names.remove(old_name);
-        let inserted = self.used_names.insert(new_name.to_owned());
-        assert!(
-            erased && inserted,
-            "Old name not found or new name already present"
-        );
-        for abbr in self.abbrs.iter_mut() {
-            if abbr.name == old_name {
-                abbr.name = new_name.to_owned();
-                break;
-            }
-        }
+    pub fn rename(&mut self, old_name: &wstr, new_name: &wstr, commands: &[WString]) {
+        self.abbrs
+            .iter_mut()
+            .find(|a| a.name == old_name && a.commands == commands)
+            .expect("Abbreviation to rename was not found.")
+            .name = new_name.to_owned();
+
+        self.used_names.insert(new_name.to_owned());
+        self.on_remove(old_name);
     }
 
-    /// Erase an abbreviation by name.
+    /// Erase an abbreviation by name and command restrictions.
     /// Return true if erased, false if not found.
-    pub fn erase(&mut self, name: &wstr) -> bool {
-        let erased = self.used_names.remove(name);
-        if !erased {
+    pub fn erase(&mut self, name: &wstr, commands: &[WString]) -> bool {
+        let Some(idx) = self
+            .abbrs
+            .iter()
+            .position(|a| a.name == name && a.commands == commands)
+        else {
             return false;
-        }
-        for (index, abbr) in self.abbrs.iter().enumerate().rev() {
-            if abbr.name == name {
-                self.abbrs.remove(index);
-                return true;
-            }
-        }
-        panic!("Unable to find named abbreviation");
+        };
+
+        self.abbrs.remove(idx);
+        self.on_remove(name);
+
+        true
     }
 
     /// Return true if we have an abbreviation with the given name.
@@ -260,6 +261,16 @@ impl AbbreviationSet {
     /// Return a reference to the abbreviation list.
     pub fn list(&self) -> &[Abbreviation] {
         &self.abbrs
+    }
+
+    /// Checks if any other abbreviation still uses name, removing it from used_names if not
+    fn on_remove(&mut self, name: &wstr) {
+        if self.abbrs.iter().any(|a| a.name == name) {
+            return;
+        }
+
+        let removed = self.used_names.remove(name);
+        assert!(removed, "Name was not in used_names but should have been");
     }
 }
 
@@ -438,13 +449,13 @@ mod tests {
             assert!(!abbrs_g.has_name(L!("gcc")));
             assert!(abbrs_g.has_name(L!("gc")));
 
-            abbrs_g.rename(L!("gc"), L!("gcc"));
+            abbrs_g.rename(L!("gc"), L!("gcc"), &[]);
             assert!(abbrs_g.has_name(L!("gcc")));
             assert!(!abbrs_g.has_name(L!("gc")));
 
-            assert!(!abbrs_g.erase(L!("gc")));
-            assert!(abbrs_g.erase(L!("gcc")));
-            assert!(!abbrs_g.erase(L!("gcc")));
+            assert!(!abbrs_g.erase(L!("gc"), &[]));
+            assert!(abbrs_g.erase(L!("gcc"), &[]));
+            assert!(!abbrs_g.erase(L!("gcc"), &[]));
         })
     }
 }
--- a/src/builtins/abbr.rs
+++ b/src/builtins/abbr.rs
@@ -229,26 +229,56 @@ fn abbr_rename(opts: &Options, streams: &mut IoStreams) -> BuiltinResult {
         return Err(STATUS_INVALID_ARGS);
     }
     abbrs::with_abbrs_mut(|abbrs| -> BuiltinResult {
-        if !abbrs.has_name(old_name) {
+        if !abbrs
+            .list()
+            .iter()
+            .any(|a| a.name == *old_name && a.commands == opts.commands)
+        {
             streams.err.append(wgettext_fmt!(
-                "%s %s: No abbreviation named %s\n",
+                "%s %s: No abbreviation named %s with the specified command restrictions\n",
                 CMD,
                 subcmd,
                 old_name.as_utfstr()
             ));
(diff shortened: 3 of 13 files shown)


src/abbrs.rs
```diff
@@ -99,8 +99,14 @@ impl Abbreviation {
         if !self.matches_position(position) {
             return false;
         }
-        if !self.commands.is_empty() && !self.commands.contains(&command.to_owned()) {
```
krobelus, line 102: would have been nice to add a test case to the commit message,
or (even better) to tests/checks/abbr.fish,
so reviewers don't have to read through the issue.

src/abbrs.rs
```diff
@@ -99,8 +99,14 @@ impl Abbreviation {
...
-            return false;
+        if self.commands.is_empty() {
+            if !command.is_empty() {
+                return false;
```
krobelus: I don't really know this code but this feels like a bit hard to follow,
can we make it easier?

I think what the existing code is trying to do is,
return false if all of
1. the abbr has any `-c` argument
2. the actual command does not match any of the `-c` arguments

and you return false if all of
1. there is no `-c` argument
2. there is no command detected.  

I don't understand why,
do you have a test case.

src/abbrs.rs
```diff
@@ -99,8 +99,14 @@ impl Abbreviation {
...
+            }
+        } else {
+            if command.is_empty() || !self.commands.contains(&command.to_owned()) {
```
krobelus: I'm also not sure why we also special-case "command.is_empty()" in the other branch,
because why would `self.commands` ever contain an empty command?
If it doesn't that's probably a case of GIGO

src/abbrs.rs
```diff
@@ -211,45 +217,66 @@ impl AbbreviationSet {
...
-        let inserted = self.used_names.insert(new_name.to_owned());
+    pub fn rename(&mut self, old_name: &wstr, new_name: &wstr, commands: &[WString]) {
         assert!(
-            erased && inserted,
-            "Old name not found or new name already present"
+            !self
+                .abbrs
+                .iter()
+                .any(|a| a.name == new_name && a.commands == commands),
+            "Target abbreviation name and commands already exist."
```
krobelus: this assertion is no longer needed, because we already do

	.expect("Abbreviation to rename was not found.")

below, which has the same meaning

src/abbrs.rs
```diff
@@ -211,45 +217,66 @@ impl AbbreviationSet {
...
-            if abbr.name == name {
-                self.abbrs.remove(index);
-                return true;
+    pub fn erase(&mut self, name: &wstr, commands: &[WString]) -> bool {
+        let index = self
+            .abbrs
+            .iter()
+            .position(|a| a.name == name && a.commands == commands);
+
+        if let Some(idx) = index {
```
krobelus: use let-else here to reduce indentation

	let Some(idx) = self.abbrs.position(...) else {
		return false;
	};

src/abbrs.rs
```diff
@@ -211,45 +217,66 @@ impl AbbreviationSet {
...
+            self.abbrs.remove(idx);
+
+            if !self.abbrs.iter().any(|a| a.name == name) {
+                assert!(
+                    self.used_names.remove(name),
```
krobelus: it still looks weird to put a function call with side-effects inside an assertion.
Would look less surprising if we extracted `let removed = self.used_names.remove(name)`,
but I guess there's no really good reason to.

src/abbrs.rs
```diff
@@ -211,45 +217,66 @@ impl AbbreviationSet {
...
+                    "Name was not in used_names but should have been"
```
krobelus: this code ("remove from used_names if necessary") is duplicated across both rename() erase(),
can we extract a simple function like `fn on_remove(name: &wstr)` to do it?

src/builtins/abbr.rs
```diff
@@ -229,26 +229,35 @@ fn abbr_rename(opts: &Options, streams: &mut IoStreams) -> BuiltinResult {
...
                 subcmd,
                 old_name.as_utfstr()
             ));
             return Err(STATUS_CMD_ERROR);
         }
-        if abbrs.has_name(new_name) {
+        if abbrs
+            .list()
+            .iter()
+            .any(|a| a.name == *new_name && a.commands == opts.commands)
```
krobelus, line 248: this should probably be stricter.
Consider 

	abbr f --command=foo ...
	abbr fb --command=foo --command=bar ...
	abbr --rename f --command=foo fb

It's probably not a good idea for this to be allowed, because they do conflict.

src/builtins/abbr.rs
```diff
@@ -229,26 +229,35 @@ fn abbr_rename(opts: &Options, streams: &mut IoStreams) -> BuiltinResult {
...
+        {
             streams.err.append(wgettext_fmt!(
-                "%s %s: Abbreviation %s already exists, cannot rename %s\n",
+                "%s %s: Abbreviation %s with the specified command restrictions already exists, cannot rename %s\n",
```
krobelus: probably better not to change the output if there are no command restrictions?
And if there are, we can mention the concrete ones: 

	Abbreviation dst already exists for commands 'foo', 'bar' and 'baz', cannot rename src

src/builtins/abbr.rs
```diff
@@ -229,26 +229,35 @@ fn abbr_rename(opts: &Options, streams: &mut IoStreams) -> BuiltinResult {
...
                 CMD,
                 subcmd,
                 new_name.as_utfstr(),
                 old_name.as_utfstr()
             ));
             return Err(STATUS_INVALID_ARGS);
         }
-        abbrs.rename(old_name, new_name);
+
+        abbrs.rename(old_name, new_name, &opts.commands);
```
krobelus, line 281: The synopsis in doc_src/cmds/abbr.rst is

	abbrs --rename OLD_WORD NEW_WORD

so it doesn't mention "--command" at all.

So we should either document this behavior,
or have "abbr --rename" error if OLD_WORD is ambiguous.

Either is fine by me.

src/builtins/abbr.rs
```diff
@@ -414,18 +423,20 @@ fn abbr_erase(opts: &Options, parser: &Parser) -> BuiltinResult {
...
             // Erase the old uvar - this makes `abbr -e` work.
-            let esc_src = escape(arg);
-            if !esc_src.is_empty() {
-                let var_name = WString::from_str("_fish_abbr_") + esc_src.as_utfstr();
-                let ret = parser.vars().remove(&var_name, EnvMode::UNIVERSAL);
-
-                if ret == EnvStackSetResult::Ok {
-                    result = Ok(SUCCESS)
-                };
+            if opts.commands.is_empty() {
```
krobelus, line 451: Wait, do abbreviations with commands have no `_fish_abbr_` materialization?
I guess that might be true because that's a legacy thing.

But this change feels like an independent fix,
if that's true, it should go into a separate commit if possible.

src/builtins/abbr.rs
SandWoodJones: Thanks for the review. I'm not sure if I agree here as the resulting `(name = "fb", commands = ["foo"])` doesn't conflict with `(name = "fb", commands = ["foo", "bar"])`. With the new logic we consider an `abbr` unique by their (name, commands) pair, instead of just their name.

krobelus: it doesn't conflict in the data structure but it does when trying to use it. Hmm.

src/builtins/abbr.rs
SandWoodJones: I'm not sure if I understand this bit, the `if opts.commands.is_empty()` is to make sure that when running `abbr -e --command foo my_abbr`, we don't delete the _general_ `my_abbr`

krobelus: ok so it sounds like it is necessary for the other changes to work

src/builtins/abbr.rs
SandWoodJones: I see what you mean now. I think that is consistent with existing behavior for example,
```
abbr --command foo -- bar baz
abbr --command foo --command oof -- bar zab
```
Here the last definition wins and `zab` is used for `foo`.
I believe that would also be the case for renaming but i'm not sure as i think `abbr --rename` might currently be broken:
```
$ fish -N
$ abbr foo bar
$ abbr --rename foo oof
$ abbr
abbr -a -- oof bar
```
Where `foo` still abbreviates to `bar` and `oof` does nothing.

doc_src/cmds/abbr.rst
```diff
@@ -136,9 +136,9 @@ Other subcommands
 
 ::
 
-    abbr --rename OLD_NAME NEW_NAME
+    abbr --rename [OPTIONS] OLD_NAME NEW_NAME
 
-Renames an abbreviation, from *OLD_NAME* to *NEW_NAME*
+Renames an abbreviation, from *OLD_NAME* to *NEW_NAME*. Options such as ``--command`` can be provided to disambiguate which abbreviation to rename.
```
krobelus: I'd put it more concretely, because we know it's only --command for the foreseeable future.
It's also not optional to specify them for disambiguation,
at least how it works today.

	$ abbr x -c foo asdf
	$ abbr --rename x y
	abbr --rename: No abbreviation named x with the specified command restrictions

So maybe

	abbr --rename [ [-c | --command COMMAND]... ] OLD_NAME NEW_NAME

	For command-specific abbreviations, the ``--command`` options must be provided.

BTW we prefer semantic line breaks now.
So we give at least [one line to each sentence](https://nick.groenen.me/notes/one-sentence-per-line/)

src/abbrs.rs
```diff
@@ -99,9 +99,15 @@ impl Abbreviation {
         if !self.matches_position(position) {
             return false;
         }
-        if !self.commands.is_empty() && !self.commands.contains(&command.to_owned()) {
+
+        if self.commands.is_empty() != command.is_empty() {
+            // No match if we're matching an abbr with no -c to one with -c
```
krobelus: but this is matching abbreviations against whatever's on the command line!
So I don't think that "command.is_empty()" means what you imply it does.
Rather it's empty only as fallback value.
There are two places where it's set
1. `src/highlight/highlight.rs` in
   `is_valid = with_abbrs(|set| set.has_match(cmd, abbrs::Position::Command, L!("")))`
   likely because of laziness - I can't off-hand think of a reason why we wouldn't extract the command token here..
2. `src/reader/reader.rs` in
   `let replacers = abbrs_match(token_str, position, cmd.unwrap_or(L!("")));`

so I think "command" is empty here if the command line is something like
"if " or "time " or "a=b ",
which are special in the fish grammar and not ordinary commands.

It is weird that we pass the empty string instead of None,
I think we should fix that.

But given that empty abbreviations are not allowed

	$ abbr '' asdf
	abbr --add: Name cannot be empty

I think we should apply the same to "abbr --command ''",
I can't think of any reason to ever use this.

That could be a separate (earlier) commit.
Please also squash the "address review" commit,
we're going for a nice final history.

Once we have ruled out "abbr --command ''",
i.e. "commands" will never contain an empty string as key,
then we can be sure that

	self.commands.contains(&command.to_owned())

is true only if command is non-empty.
Then we don't need to modify this logic, it can stay.

        if !self.commands.is_empty() && !self.commands.contains(&command.to_owned()) {
            return false;
	}

I hadn't read this code before;
this naming makes it easy to mix things up.
Maybe we should rename "Abbreviation::commands"
to "Abbreviation::commands" to 
"Abbreviation::valid_for_commands".
Not sure if that's really better, so no need.

tests/checks/abbr.fish
```diff
@@ -208,3 +208,29 @@ abbr --add regex_name --regex '(*UTF).*' bar
 abbr --add foo --set-cursor 'foo % bar'
 abbr | grep foo
 # CHECK: abbr -a --set-cursor='%' -- foo 'foo % bar'
+
+# Command-specific and general abbrs can coexist
+abbr __abbr_coexist "general def"
+abbr --command foo __abbr_coexist "command def"
+abbr | grep __abbr_coexist
```
krobelus, line 215: FWIW you can run relevant tests with

	tests/test_driver.py target/debug tests/**/abbr*

src/abbrs.rs
SandWoodJones: Oh, sorry, I usually squash things after the review process is done. About empty commands though, according to the `abbr` documentation they are intentional:
> ["An empty **COMMAND** means it will expand only when there is no command."](https://github.com/fish-shell/fish-shell/blob/d6ed5f843e6f87c1181555efb1526b90a0172a8b/doc_src/cmds/abbr.rst?plain=1#L70)

tests/checks/abbr.fish
SandWoodJones: Thanks, I've been using the driver. The failing tests for `abbr.fish` are unrelated to this PR.

krobelus: > The failing tests for abbr.fish are unrelated to this PR.

I don't think so; I get failures like

```
  The CHECKERR on line 67 wants:
    abbr --rename: No abbreviation named __abbr6

  which failed to match line stderr:2:
    abbr --rename: No abbreviation named __abbr6 with the specified command restrictions
```

(regarding my comment asking to remove the
`with the specified command restrictions` affix when there are no restrictions,
I don't think that's a blocker, it's not terribly important I guess, and I'm not 100% sure which behavior is better)

But make sure to fix the tests; you might also need to run `update_translations.sh`.
And squash and add `Fixes #11184` to the commit message, to autoclose that

src/abbrs.rs
krobelus: Ok that sounds like it makes sense. (@faho would be the expert)

But then
1. the fix -- IUIC that's about  making
   `abbr --command "" x`
   *not* expand on `foo x` but still have it expand on `x`.
   If that's unrelated to the other fix, please put it in a separate commit to not waste reviewer time (especially since the reviewer may not know this area of the code).
   But I also don't understand why this exists isn't it the same as `abbr x` ??
2. the comment `// No match if we're matching an abbr with no -c to one with -c`
   is misleading, because one of them is an abbr spec, the other is a conrete command

faho: >  not expand on foo x but still have it expand on x.

Yeah, this was mentioned in #10452.

The idea there, IIRC, was that you could use it with `sudo` et al - `abbr --command "" --command sudo --command doas foo bar`.

That wouldn't really be *necessary* if `abbr --command sudo foo bar` and `abbr foo bar` (plus appropriate `--position`) don't conflict, tho I think it's reasonably intuitive.

ccoVeille:
May I ask for a quick summary why this PR was closed?

It seemed to address an issue I had.

I looked at the thread but I faold to find an explanation.

krobelus:
I merged the changes in this PR (with small changes to the translation files).
You can find the commit in "closed in <commit ID>" message.

This workflow can be a little bit more convenient (especially when making editorial changes) but I realize it does look weird on github.
I have some potential workflow changes planned that could make this more obvious.

ccoVeille:
Thanks. I thought the commits I saw were the ones of this PR.

It's clearer now

Issue #4918, opened 2018-04-17
Suggested feature: Support mouse tracking in Linux


arty-name:
When editing a long command in the command line it would be very convenient and natural to be able to click on a character and have the edit cursor moved to this character. I have seen this already working on Mac with Option+Click, but it doesn’t seem to work on my Linux machine. I am using Ubuntu and Gnome-Terminal, fish version is 2.7.1 and the $TERM is xterm-256color. It doesn’t work without customizations either.

I assume for this to work the shell should support [mouse tracking](https://stackoverflow.com/a/5970472/267780) but I am new to this and not 100% sure. Kind people in the gitter channel told me about Alt+V to start the default editor, and this is a definitely useful feature, but it seems an overkill for that particular task. 

Steps to reproduce: 
1) Start gnome-terminal with fish shell
2) Type a few characters in the prompt
3) Click one of these characters with the mouse

Actual: the editing cursor stays in its position
Expected: the editing cursor moves to the character clicked

Would you consider support for this feature?

zanchey:
> I have seen this already working on Mac with Option+Click

Does it? It does! I didn't know that!

On Terminal.app and iTerm, this is implemented by [sending cursor movement keys until the correct position is reached](https://apple.stackexchange.com/questions/42825/option-click-in-iterm), so fish is oblivious.

Making this work under XTerm mouse reporting is certainly possible, though may require a lot of work.

arty-name:
I don’t know if that of any help, but I have seen [this script to support mouse tracking in zsh](https://github.com/ashleyh/zsh-mouse/blob/ash/mouse.zsh)

mqudsi:
+1 

mouse tracking in curses: http://tldp.org/HOWTO/NCURSES-Programming-HOWTO/mouse.html

PerBothner:
[DomTerm](https://domterm.org) also [supports moving the cursor with a mouse click](http://domterm.org/Tips.html#move-cursor-with-mouse), by sending arrow key sequences.  You can always do Alt-click.  With an appropriate escape sequence in your prompt, it also works for plain click.  Here are [instructions for bash](http://domterm.org/Shell-prompts.html).  I'm new to zsh, but this seems to work:

    if [ -n "$DOMTERM" ]
    then
       ZSHPID=`sh -c 'echo $PPID'`
[...]

PR #12026, opened 2025-11-06
Enable putting cursor by mouse clicking

SuperKenVery:
## Description

The mouse handling code was already there. What I did:
- Enable mouse reporting when initializing terminal
- Remove disable_mouse_tracking calls

Fixes issue #4918

## TODOs:

### DIFF
--- a/src/env_dispatch.rs
+++ b/src/env_dispatch.rs
@@ -282,12 +282,10 @@ fn handle_fish_cursor_end_mode_change(vars: &EnvStack) {
 }
 
 fn handle_autosuggestion_change(vars: &EnvStack) {
-    FLOG!(term_support, L!("fish_autosuggestion_enabled changed"));
     reader_set_autosuggestion_enabled(vars);
 }
 
 fn handle_fish_mouse_change(vars: &EnvStack) {
-    FLOG!(term_support, L!("fish_mouse changed"));
     let Some(data) = current_data() else {
         return;
     };


krobelus:
> It's now possible to put the cursor by clicking the desired position with your mouse.

FWIW terminals that support OSC 133 click_events feature
(https://fishshell.com/docs/4.1/terminal-compatibility#term-compat-osc-133)
such as kitty already offered that.  Thanks. Will queue.

src/tty_handoff.rs
```diff
@@ -189,35 +189,47 @@ impl TtyQuirks {
         };
         let enablers = ProtocolBytes {
             kitty_keyboard: serialize_commands(maybe_enable_focus_reporting(&[
-                DecsetBracketedPaste,                       // Enable bracketed paste
+                DecsetMouseTracking,
```
krobelus: ah no, we don't want to enable it unconditionally at this point,
because that would break terminal selection.

Until we figure out how to automatically do the right thing,
we can add user-controllable "fish_mouse" variable, and toggle things based on that.
It'd be easy to toggle it via a fish binding like

	bind ctrl-g 'set fish_mouse (math 1 - $fish_mouse)'

Is there any obvious key binding we could use?

For now I'll take the "no longer disable mouse tracking" already.

SuperKenVery: > However, keep in mind that fish might eventually want to add it's own knob (like set fish_mouse true),
which might override your escape sequence. But that's not planned yet.

So you mean we should do this now, right?

Or should we do some hacking e.g. if the mouse drags we disable mouse reporting?

--- 

EDIT: I think the hack won't work because we can't know when to turn reporting back on...

krobelus: > So you mean we should do this now, right?

yes, we can add the config option now, if you want to pick that up.
It would work a bit like `fish_autosuggestion_enabled`.

> I think the hack won't work because we can't know when to turn reporting back on...

yes I thought about this for a bit; not sure if I saved/posted my thoughts somewhere.
We'd need to design a solution that probably involves some changes to the terminal.
One approach might be: whenever the users interacts with the mouse with the area that "belongs" to fish, the mouse events should be sent to fish, else the behavior should be as today.
Or some kind of intuitive UI for the user to toggle mouse mode.

SuperKenVery: I think this is done now. The variable name is `fish_mouse`.

krobelus:
Great. Nice cleanup on get_protocols().

Out of curiosity, do you really want to have mouse tracking turned on all the time?

It of breaks things like
- click+drag to make a selection in the terminal (can be worked around with shift-click+drag)
- double click to select a word
- triple-click to select a line

and the advantage is mainly that it allows moving the cursor with the mouse.
You might prefer using the alt-e binding to spawn a full-screen editor

SuperKenVery:
Well, the main reason is I didn't know about `alt-e`...

My use case is that, I'm doing some deep learning work and I have very long and multi-line commands, like this:

```bash
CUDA_VISIBLE_DEVICES=0,1 \
    accelerate launch --main_process_port 0 1_train_model.py \
      --model SPF_LUT_net \
      --scale 4 \
      --modes s \
      --expDir ../models/take-residual-out \
      --trainDir ../data/DIV2K \
      --valDir ../data/SRBenchmark \
      --sample-size 3 \
      --workerNum 8 \
      --batchSize 16
```

And changing a number in the middle can often be a pain in the ass.

As for the word selecting, I can just hold `shift` and everything would go back (and mouse events won't be reported to program in terminal), at least in wezterm.

krobelus:
> Well, the main reason is I didn't know about `alt-e`...
> 
> My use case is that, I'm doing some deep learning work and I have very long and multi-line commands, like this:

Interesting.
I wonder if we can detect this case and
1. either suggest to use alt-e
2. or enable mouse automatically if the command line is too large

I guess 2 is less likely to work out.

> And changing a number in the middle can often be a pain in the ass.
> 
> As for the word selecting, I can just hold `shift` and everything would go back (and mouse events won't be reported to program in terminal), at least in wezterm.

Right, I think that works on a lot of terminals (and in all ncurses apps).
Sounds like you're good with having it turned on always. LMK if that changes (or if you won't use it at all now that you know about alt-e), it'd be a valuable data point.
(Maybe we should toggle mouse capture based on capslock. Bad idea probably.)

Most users probably don't know about the trick with holding shift,
and teaching it to everyone is probably not easy. Also, we probably can't break habits.

SuperKenVery:
> LMK if that changes (or if you won't use it at all now that you know about alt-e)

I would probably still use this. I often just need to change some digits, and opening helix for that, then click (on the top part of the screen rather than in place), then go to insert mode, change, and escape and save, is a little bit long.

> Most users probably don't know about the trick with holding shift

Well if people uses vim or helix, or anything that captures mouse, and they want to copy text to clipboard, they probably need to know this.

> Also, we probably can't break habits.

Yeah, and when `fish_mouse` isn't set this is default to off, so I think it should be fine.

src/env_dispatch.rs
```diff
@@ -281,6 +285,20 @@ fn handle_autosuggestion_change(vars: &EnvStack) {
     reader_set_autosuggestion_enabled(vars);
 }
 
+fn handle_fish_mouse_change(vars: &EnvStack) {
```
krobelus, line 288: the "fish_" seems reundant,
though we could call this function `handle_mouse_enabled_change` or similar, not sure.
Maybe we should also rename the user-visible variable,
but tmux also calls it just "mouse" so maybe this seems fine.

src/env_dispatch.rs
```diff
@@ -281,6 +285,20 @@ fn handle_autosuggestion_change(vars: &EnvStack) {
...
+    let Some(data) = current_data() else {
```
krobelus, line 289: I'd do something like this for docs and completions:

```diff
diff --git a/CHANGELOG.rst b/CHANGELOG.rst
index 76d1bc52fa..649fb82e26 100644
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -22,8 +22,7 @@
 - Fish now hides the portion of a multiline prompt that is scrolled out of view due to a huge command line. This prevents duplicate lines after repainting with partially visible prompt (:issue:`11911`).
 - On macOS, fish sets :envvar:`MANPATH` correctly also when that variable was already present in the environment (:issue:`10684`).
 - Added a workaround for MSYS2 to prevent Konsole and WezTerm from opening new tabs in the wrong working directory (:issue:`11981`). 
-- fish no longer disables mouse tracking sequences (DECSET/DECRST 1000),
-  so you can use those to toggle mouse reporting,
+- It is now possible to enable mouse capture by setting :envvar:`fish_mouse`,
   which allows to move the cursor or select completions items with the mouse (:issue:`4918`).
 
 Scripting improvements
diff --git a/doc_src/language.rst b/doc_src/language.rst
index 37c3888894..6f42f29f89 100644
--- a/doc_src/language.rst
+++ b/doc_src/language.rst
@@ -1628,6 +1628,11 @@
    empty string, history is not saved to disk (but is still available within the interactive
    session).
 
+.. envvar:: fish_mouse
+
+   Set it to 1 to capture mouse input, which allows move the cursor and select pager items using the mouse.
+   Note that this will prevent the terminal from handling mouse input.
+
 .. envvar:: fish_trace
 
    if set and not empty, will cause fish to print commands before they execute, similar to ``set -x``
diff --git a/share/completions/set.fish b/share/completions/set.fish
index f5396f6dfc..82d294c295 100644
--- a/share/completions/set.fish
+++ b/share/completions/set.fish
@@ -56,6 +56,7 @@
         umask "current file creation mask" \
         fish_ambiguous_width "affects computed width of east asian chars" \
         fish_autosuggestion_enabled "set to 0 to turn autosuggestions off" \
+        fish_mouse "set to 1 to capture mouse input" \
         fish_cursor_end_mode "set to 'inclusive' to disallow moving the cursor beyond the command line end" \
         fish_cursor_selection_mode "set to 'inclusive' if selections should include the cursor" \
         fish_emoji_width "cols wide fish assumes emoji render as" \
```

src/reader/reader.rs
```diff
@@ -2630,6 +2631,18 @@ impl<'a> Reader<'a> {
...
                 }
+                ImplicitEvent::SetMouseTracking(enabled) => {
+                    FLOG!(reader, "Set mouse tracking enabled:", enabled);
+                    Outputter::stdoutput()
+                        .borrow_mut()
+                        .write_command(if enabled {
+                            DecsetMouseTracking
+                        } else {
+                            DecrstMouseTracking
+                        });
```
krobelus, line 2642: it's not as simple as that
because the `TtyProtocolsSet` is immutable.
So the variable change would be lost after the next command.

To make this work reliably, I think we should
1. disable all protocols
2. toggle our flag
3. re-enable protocols, including our flag

Something like

```rust
    SetMouseTracking(enabled) => {
        FLOG!(reader, "Set mouse tracking enabled:", enabled);
        let mut tty = TtyHandoff::new(reader_save_screen_state);
        tty.disable_tty_protocols();
	// TODO now set a global atomic bool in tty_handoff.rs
	// Now let the Drop implementation of TtyHandoff re-enable
	// TTY protocols, including mouse capture if applicable.
    }
```

Since this involves atomics and async-signal-safe code,
it may get tricky (though it also might not),
if you prefer I can finish this when I get around to

src/tty_handoff.rs
```diff
@@ -172,52 +173,67 @@ impl TtyQuirks {
     }
 
     // Return the protocols set to enable or disable TTY protocols.
-    fn get_protocols(self) -> TtyProtocolsSet {
+    fn get_protocols(self, vars: &dyn Environment) -> TtyProtocolsSet {
+        let mut on_chain = vec![DecsetBracketedPaste];
+        let mut off_chain = vec![DecrstBracketedPaste];
```
krobelus, line 178: with the other comment, changing the serialized commands should no longer be necessary,
but let's keep the cleanup for the tmux-specific code paths,
ideally in a separate commit

src/reader/reader.rs
SuperKenVery: Oh. I previously tried to disable protocols, re-initialize tty protocols, then re-enable them, after `fish_mouse` is changed. However it didn't work as expected and I couldn't figure it out, what's more, that introduced a race condition that sometimes triggered an assertion that protocols should be off when running commands. Probably my implementation was problematic, though.

> because the `TtyProtocolsSet` is immutable

This should explain why it didn't work...

I certainly need to dig deeper to solve this. 

Also, out of curiosity, why did you handle the atomics yourself? I believe there are wrappers like `OnceLock` or `LazyLock`, which should make life easier.

krobelus: > Also, out of curiosity, why did you handle the atomics yourself? I believe there are wrappers like OnceLock or LazyLock, which should make life easier.

When you do `kill -TERM $fish_pid`, fish wants to disable TTY protocols.
This is currently done in a signal handler (on an arbitrary thread).

We can not use things like `Mutex` in signal handlers. See `man signal-safety` for which functions are allowed in signal handlers. For Rust APIs, we'd ideally also want a promise that the implementation won't violate signal safety in future.

Hence the mouse disposition should be accessible in a global variable by only calling async-signal-safe functions.
We try to prefix the name of every function that is used in a signal handler with `safe_`.

src/tty_handoff.rs
krobelus: If you get the chance to work on this, can you clean up the commit history, i.e. squash most commits -- the final result should probably have two commits, one for the tmux-hack cleanup, and another one to add `fish_mouse`.

SuperKenVery: Okay, I will do that when I get time...

PR #12029, opened 2025-11-06
tempfile: utilize new tempfile crate

danielrainer:
Stacked on #12028

### DIFF
--- a/src/wutil/dir_iter.rs
+++ b/src/wutil/dir_iter.rs
@@ -328,8 +328,12 @@ impl Iterator for Iter {
 
 #[cfg(test)]
 mod tests {
+    use std::fs::File;
+    use std::path::PathBuf;
+
+    use nix::sys::stat::Mode;
+
     use super::{DirEntryType, DirIter};
-    use crate::common::wcs2zstring;
     use crate::wchar::prelude::*;
 
     #[test]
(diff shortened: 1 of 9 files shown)


src/wutil/dir_iter.rs
```diff
@@ -328,8 +328,12 @@ impl Iterator for Iter {
 
 #[cfg(test)]
 mod tests {
+    use std::fs::File;
+    use std::path::PathBuf;
+
+    use nix::sys::stat::Mode;
```
krobelus, line 334: We haven't put a lot of thought into import ordering yet;
for tests specifically I put everything in one block,
which lists the "super::*" dependencies first.
The idea is that this can make it slightly easier to see our test coverage.
Not sure.

I'll make this small edit before merging, so no need to update this.

krobelus, line 334: We haven't put a lot of thought into import ordering yet;
for tests specifically I put everything in one block,
which lists the "super::*" dependencies first.
The idea is that this can make it slightly easier to see our test coverage.
Not sure.

I'll make this small edit before merging, so no need to update this.

danielrainer: I don't really care about the exact formatting of imports. Ideally, imports are added by rust-analyzer and formatted be rustfmt, without me ever having to look at them. At the moment, rustfmt seems somewhat reluctant to change import formatting. Maybe that can be addressed by a config change.

